PR #15139, opened 2021-01-13
systemd: Rewrite shutdown modal to react and PF4

marusak:
Not much visual difference. But the date selector is fully localized, meaning it shows date in your locale and also the popup is localized (like first day of week, names of months...). Also I enhanced errors a bit more, meaning that  you get alert about wrong date right away and the submit button is disabled. Before you had to click submit to get info about wrong date or time.

Before:
beforeresart

Now:
newrestart

With inline error:
inlinerror


martinpitt:
Great work! Needs some more test case adjustments

marusak:
Almost there, it just seems that sometimes error message is not shown. I'll take a better look at this. In the meantime, ready for review.

pkg/systemd/shutdown.jsx
```diff
@@ -0,0 +1,198 @@
...
+                   </>}
+            >
+                <>
+                    <Form isHorizontal onSubmit={e => {
+                        // HACK: https://github.com/patternfly/patternfly-react/issues/5299
+                        e.preventDefault();
+                        return false;
+                    }}>
+                        <FormGroup fieldId="message" label={_("Message to logged in users")}>
+                            <TextArea id="message" value={this.state.message} onChange={v => this.setState({ message: v })} />
```
KKoukiou, line 173: Text area allows you to expand it horizontaly and vertically. When inside modals vertical works fine, the modal get's expanded as well, but the horizontal escapes outside of the modal. I have not reported it yet to PF, I need to check if it's related to any of our custom CSS.

KKoukiou (dismissed):
Very clean rewrite - I was only able to find one minor issue - we can fix it in a followup or mayb just report it upstream to pf  - it's nothing urgent.

KKoukiou:
> Very clean rewrite - I was only able to find one minor issue - we can fix it in a followup or mayb just report it upstream to pf - it's nothing urgent.

Ah and one more thing I noticed. When the scheduled time is single digit minute, we just display one digit, that's not common for times. so 14:01 is the expected time format not 14:1

Screen Shot 2021-01-14 at 14 05 46

pkg/systemd/shutdown.jsx
marusak: I've added `resize:vertical` so it can be resized only vertically.

marusak:
> Ah and one more thing I noticed. When the scheduled time is single digit minute, we just display one digit, that's not common for times. so 14:04 is the expected time format not 14:4.

hmm, I am thinking about this and the only way I can image this to be working is `onBlur` because if we would add the `0` during typing, someone would type `5`, we would add the zero to `05` but then the person would continue typing to `50` and it would just keep changing too much, wouldn't it? Or is there some best-practices doc about this?

marusak:
>  that's not common for times. so 14:04 is the expected time format not 14:4.

It was also regression from master, fixed it.

KKoukiou:
> > that's not common for times. so 14:04 is the expected time format not 14:4.
> 
> It was also regression from master, fixed it.

For that you could have also used type="time" on your text input, instead of two textInputs with this on Blur and so on.

KKoukiou:
> > > that's not common for times. so 14:04 is the expected time format not 14:4.
> > 
> > 
> > It was also regression from master, fixed it.
> 
> For that you could have also used type="time" on your text input, instead of two textInputs with this on Blur and so on.

Of course it's fine as is.

marusak:
> For that you could have also used type="time" on your text input, instead of two textInputs with this on Blur and so on.

Lets actually use https://www.patternfly.org/v4/components/time-picker but in some other PR

KKoukiou:
> > For that you could have also used type="time" on your text input, instead of two textInputs with this on Blur and so on.
> 
> Lets actually use https://www.patternfly.org/v4/components/time-picker but in some other PR

True true. I forgot about it.

KKoukiou (approved):

pkg/systemd/shutdown.scss
```diff
@@ -1,32 +1,17 @@
...
+#shutdown-minute {
+    width: 3em;
+    margin: 0 0.25em;
 }
 
-#shutdown-dialog input {
-    display: inline;
-    width: 10em;
+textarea {
+    resize: vertical;
```
KKoukiou, line 12: That one you could have done in react only https://www.patternfly.org/v4/components/text-area/#vertically-resizable-text-area

marusak: oh nice, I totally missed this option. I'll repush

KKoukiou: No worries from my side.

KKoukiou (approved):

PR #15120, opened 2021-01-11
po: Build translations with webpack

martinpitt:
Convert the `po2json` script into a webpack plugin, and integrate the
(now trivial) po.empty.js template.

This is another step for building the entire dist/ directory with
`npm run build` (i.e. a tool that web developers are familiar with),
and not having a split webpack+make toolchain any more.

Based on starter-kit commit 7f6ef51c12a75. cockpit-po-plugin.js was
extended to accept a `subdir` option for building a particular
pkg/<page>/ only, a `msggrep_options` option for additional `msggrep`
command line arguments, and a `wrapper` option for using a custom call
to poke the PO data into the page (which we need for the login page, as
that does not have cockpit.js). These are backwards compatible with thhe
starter-kit version, so in the future we can drop the latter copies and
replace them with a tagged git checkout.

Change how ws inserts translations into the login page: Previously, the
Makefile generated static/login.po.LL.html files *in addition* to the
usual po.LL.js files, which duplicated the translations again. Drop all
these, and make ws insert the po.*.js file directly instead. Add a new
marker so that this happens inside a `<script>` tag.

Note: It would be way more elegant to just make login.html include
`<script src="po.js"></script>`, but unfortunately that does not work
with ws (it does not pick the correct per-language po.LL.js file in the
login page code paths).

----

I compared translations against master with `for f in dist/*/po.de.js; do diff -u ../cockpit/$f $f; done|view -` and the diff looks reasonable. In fact this *adds* a few forgotten translations from manifest descriptions and keywords.

 - [x] Builds on top of PR #15119
 - [x] Builds on top of PR #15142 
 - [x] Fix login page translations and broken layout
 - [x] Adjust commit message, we also need to build manifests with webpack
 - [x] Fix handling po.js.gz (ws can't transparently serve static .gz files)
 - [x] Fix [check-pages regressions](https://logs.cockpit-project.org/logs/pull-15120-20210114-104422-ee2afe9e-fedora-33/log.html): PR #15149

PR #15142, opened 2021-01-14
playground: Drop extra.de.po

martinpitt:
The string does not exist in the source, we don't use it for anything,
and it requires extra overhead in the build system.


martinpitt:
Prerequisite for #15120

marusak (approved):
Thanks!
I thought that we use this for testing something, but seems not.

PR #15143, opened 2021-01-14
po: Don't double cd when building cockpit.appstream.pot

marusak:
Then the `find` could not find the file as it was one directory up as
expected.

I tried to unify all the .pot files to use the same pattern, but that
did not work out well, as each works ever so slightly differently.

### DIFF
--- a/po/Makefile.am
+++ b/po/Makefile.am
@@ -87,8 +87,8 @@ po/cockpit.manifest.pot:
 		$$(cd $(srcdir) && find $(TRANSLATE) -name 'manifest.json.in')
 
 po/cockpit.appstream.pot:
-	cd $(srcdir) && $(XGETTEXT) --output=$(abs_builddir)/$@ \
-		$$(cd $(srcdir) && find $(TRANSLATE) -name '*.appdata.xml.in')
+	cd $(srcdir) && GETTEXTDATADIRS=$(srcdir)/po $(XGETTEXT) --output=$(abs_builddir)/$@ \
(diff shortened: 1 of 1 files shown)


po/Makefile.am
```diff
@@ -87,8 +87,8 @@ po/cockpit.manifest.pot:
 		$$(cd $(srcdir) && find $(TRANSLATE) -name 'manifest.json.in')
 
 po/cockpit.appstream.pot:
-	cd $(srcdir) && $(XGETTEXT) --output=$(abs_builddir)/$@ \
-		$$(cd $(srcdir) && find $(TRANSLATE) -name '*.appdata.xml.in')
+	cd $(srcdir) && GETTEXTDATADIRS=$(srcdir)/po $(XGETTEXT) --output=$(abs_builddir)/$@ \
```
martinpitt, line 90: Why do we need `$GETTEXTDATADIRS` now? That's entirely new.

martinpitt:
btw, it's useful to push such branches to origin, then you can run the workflow directly with the real secrets.

po/Makefile.am
marusak: Probably we don't (at least it worked without it) but we use it in the following command, which this one is almost identical with.

martinpitt (approved):
Ack, thanks!

martinpitt:
Triggered the workflow, and it succeeded: https://github.com/cockpit-project/cockpit/actions/runs/485065875

It created PR #15144 

Thanks @marusak !

PR #15144, opened 2021-01-14

cockpituous:
- [x] fixed by PR #15120 
 
This can certainly be fixed without #15120 as well, but we need to investigate which msggrep exactly is acting up.


marusak:
hmm, this actually looks like there is some issue. The login page is not translated. It has been moved from `src/ws/login` to `pkg/static` and it is possible that now it is not correctly packaged.

martinpitt:
@marusak: Good news! This update works fine with PR #15120

martinpitt:
#15120 landed, but this is conflict-y now. I'll just re-run the workflow to get a fresh one.

PR #15145, opened 2021-01-14
test: automatic format/update test data to FMF and be able to schedule tests via TMT

jscotka:
It is able to convert and synchronize data about tests from python unittest format to FMF and set proper attributes to be able then run the testsuite via TMT.

It adds support for generic FMF decorators in syntax (supports https://github.com/cockpit-project/cockpit/pull/15145/files#diff-df67c587c2ba8687a9ea8bba616c38d2598c84ead7eb5af0635cacfddc81d760R101)
 - link to refecenced bug or issue via decorator `@FMF.link("bugzilla or issue link")`
 - generic tagging of tests `@FMF.tag("slow", "fast")`

## synchronize data to FMF for `verify` testsuite
```
PYTHONPATH=test/verify/ python test/common/fmf_metadata.py --path test/verify/ -f --file test/verify.fmf -u
```

## run one test via TMT via explicit TMT name
```
tmt --root test run -a provision -h local test --name /verify/check-testlib/TestRunTestListing/testBasic
```

## run all nondestructive tests
```
tmt -r test run -a provision -h local test --filter non_destructive:True
```

## verification of the convert tool
```
PYTHONPATH=test/common pytest -v -x -s test/common/fmf_test_formatting/test.py 
```


psss:
In general the format looks fine from the `tmt` point of view. The conversion of metadata stored directly in python code could be done in the future as a new [discover step](https://tmt.readthedocs.io/en/latest/spec/steps.html#discover) implementation. This would prevent the need to sync data to an `fmf` file and store it on disk before being able to explore/execute tests. The plan could then look like this:

```yaml
discover:
    how: python
execute:
    how: tmt
```

A couple of thoughts:

* The `summary` and `description` attributes could be synced from the docstring (first line and the rest) instead of using a dedicated `@fmf.summary` decorator.
* In order to run selected test I'd recommend using the unique fmf identifier, e.g. `tmt run test --name /check-examples/TestPinger/testExtend`.
* Instead of setting the `TEST_DIR` variable I'd suggest to make use of the [path](https://tmt.readthedocs.io/en/latest/spec/tests.html#path) attribute which seems to cover the same use case.
* The `skipBrowser`, `skipImage` and `skipPackage` attributes probably cover a similar use case as the relatively new [adjust](https://tmt.readthedocs.io/en/latest/spec/core.html#adjust) field.

test/common/run-tests
```diff
@@ -435,6 +435,15 @@ def main():
...
     os.environ["TEST_OS"] = image
 
+    # be able to pass test to scheduler also as ENV VALUE if not given as option
+    # compatibility with TMT scheduler
+    test_subject = os.getenv("TEST_SUBJECT")
+    if test_subject and not opts.tests:
+        opts.tests = [test_subject]
+    test_dir = os.getenv("TEST_DIR")
+    if test_dir:
+        opts.test_dir = test_dir
```
martinpitt, line 445: Please move that into the `add_argument()`, and say what the default is:
```py
    parser.add_argument('--test-dir', default=os.environ.get("TEST_DIR", testvm.TEST_DIR),
                        help="Directory in which to glob check-* files; default: %(default)s")
```

test/common/testlib.py
```diff
@@ -1409,25 +1409,45 @@ def jsquote(str):
     return json.dumps(str)
 
 
+def set_obj_attribute(testEntity, attribute, value, raise_text="", base_class=unittest.TestCase):
```
martinpitt: Do you mind to rename `raise_text` to `decorator_name` to clarify what it does?

test/verify/pytest_dynamic_loader.py
```diff
@@ -1,7 +1,10 @@
 import importlib
 import glob
 import os
-from inspect import isclass
```
martinpitt, line 4: I don't have a huge attachment to pytest_dynamic_loader.py myself, as we don't use it at all upstream nor in Fedora gating. However, it seems a bit weird to me to combine "FMF formatter" and "pytest runner" in one and the same file? Aren't these two wildly different purposes? Also, do you still actually need the pytest loader once you move to FMF?

test/verify/pytest_dynamic_loader.py
```diff
@@ -78,31 +78,25 @@ def identifier(text):
     return "/" + text
 
 
+def default_key(parent_dict, key, empty_obj):
```
martinpitt: Please squash this "lets simplify fmf formatter function" commit into the previous commit.

main.fmf
```diff
@@ -2489,7 +2675,8 @@
         TEST_SUBJECT: TestTuned.testBasic
       non_destructive: true
       skipImage:
+      - rhel-8-3-distropkg
       - fedora-coreos
       summary: check-tuned TestTuned testBasic
-duration: 10m
 test: test/common/run-tests
+duration: 20m
```
martinpitt: Duration is certainly not something that belongs into the .fmf file, as it depends on where you run it.

test/common/testlib.py
```diff
@@ -1462,6 +1477,18 @@ def nondestructive(testEntity):
     return set_obj_attribute(testEntity, "_testlib__non_destructive", True, raise_text="nondestructive", base_class=MachineCase)
 
 
+def addLink(*args):
```
martinpitt: Please drop the "add" from the two new decorators -- they always add something. `@link(url)` (or perhaps more specific, `@bug(url)`) and `@tag(name)` seem fine?  Please add docstrings to say what their purpose is, especially for `tag` where it's not obvious what to do with them.

martinpitt (changes requested):
I like the general direction, this just needs some fixes to the implementation. I only realized at the fourth commit ("rework structure") that the commits are not really independent, and that I should probably have looked at the entire diff at once. So my comments will probably be shown as obsolete, but they are still relevant. Can you please squash this, split out the independent commits, and drop the generated files? Then this will be much easier to review in the second round.

Thanks!

main.fmf
```diff
@@ -0,0 +1,2495 @@
+/check-accounts:
```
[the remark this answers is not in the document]
jscotka: > I like the general direction, this just needs some fixes to the implementation. I only realized at the fourth commit ("rework structure") that the commits are not really independent, and that I should probably have looked at the entire diff at once. So my comments will probably be shown as obsolete, but they are still relevant. Can you please squash this, split out the independent commits, and drop the generated files? Then this will be much easier to review in the second round.
> 
> Thanks!

Yes, I did not squashed commits, because I wanted to there way to be able to go back in the history in case some new solution will be worser, than new one :-)
and yes, this rework totally changed logic and structure and also your point why FMF and pytest wrapper are together.

test/common/testlib.py
jscotka: this is probably not relevant after rewrote to the more generic structure.

test/verify/pytest_dynamic_loader.py
jscotka: already reworked to separated file.

If I need this pytest loader, yes and no :-)

from one perspective, run-test does lots of magic inside, e.g. rescheduling, skipping logic, which is perfect but theoretically it maybe sometimes contra-productive and can hide potential issues (e.g. skip some test what is expected to be scheduled because it is important) also this rescheduling may lead to some strange situation. or may hide issue that it sometimes fails because of some error.

main.fmf
jscotka: you are right, but TMT does that in case duration is not set, so that it set to `5m` it could be probably rewritten in TMT plan, but not sure.
On the other hand, theoretically test itself should be more less the same time duration, what differs is that tests could also do provision part (what could be part of another tool and test just say which resources needs), so that when downloading images or starting VMs it can be much longer or I'm not right with this test time?

test/common/testlib.py
jscotka: It's already changed to suggested:
```
@FMF.link(url)
@FMF.tag("tag1", "tag2")
```
with some docstrings for these two, as probably the most important now.

martinpitt:
Two pycodestyle errors:
```
test/common/fmf_test_formatting/test.py:5:12: E225 missing whitespace around operator
test/common/fmf_test_formatting/test.py:8:1: E402 module level import not at top of file
```
You can run `tools/test-static-code` locally to reproduce/verify.

martinpitt (commented):
This looks very nice now, thanks! I like that the tests are self-testing :grin: Some remaining trivialities/questions, but let's land this for Wednesday's release, adding label.

test/main.fmf
```diff
@@ -0,0 +1,2 @@
+duration: 20m
```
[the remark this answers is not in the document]
jscotka: this duration was intended for each test, but that's true, that I can put there any value and modify for each test.

test/main.fmf
```diff
@@ -0,0 +1,2 @@
...
+test: common/run-tests
```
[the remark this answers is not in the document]
jscotka: I've fixed it to be able to run as it is. 
But you are right, this file could live also inside project root as I had it in the previous version, but it depends if we plan to use TMT `plans` right now. In case not, it is not so important to have it there, but if you wish I can move it there.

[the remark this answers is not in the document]
martinpitt: @jscotka: That's fine, let's leave the files where they are for now, as long as that works for you.

martinpitt (approved):
Thanks! I'll squash the fixup commit on landing.

jscotka:
> Thanks! I'll squash the fixup commit on landing.

Many thanks

jscotka:
@martinpitt issue already fixed.

martinpitt (approved):
Thanks! I expect that we'll have some further updates to that at some point, but at least we have the basic stuff in now.

PR #15149, opened 2021-01-15
machines: Fix test for different time formats

skobyda:
This would fix a failing test in case of different moment js formats
(e.g. "Today 11:11 PM" vs "Today at 11:11 PM")

Fixes a test fail at https://github.com/cockpit-project/cockpit/pull/15120


martinpitt (approved):
Wow, I was just going to start looking into this, and you already solved it. Many thanks!

PR #14913, opened 2020-11-17
Plotting with SVG

mvollmer:
This is a evolution of #14850.

- [x] Squash all the things.

PR #15151, opened 2021-01-15
Various cleanups

martinpitt:
These are each rather trivial, let's test them together for efficiency

### DIFF
--- a/pkg/lib/plot.css
+++ b/pkg/lib/plot.css
@@ -11,19 +11,6 @@
     color: black;
 }
 
-.flot-y-axis .flot-tick-label {
-    width: 2rem;
-    margin-right: 0.5rem;
-}
-
-.flot-x-axis .flot-tick-label {
-    margin-top: 0.25rem;
-}
-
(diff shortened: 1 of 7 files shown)


pkg/lib/plot.css
```diff
@@ -11,19 +11,6 @@
     color: black;
 }
 
-.flot-y-axis .flot-tick-label {
```
mvollmer, line 14: I think all of plot.css can be removed, actually.  We should have done that in #14913, but totally forgot about cleaning up properly in all the excitement...

mvollmer (changes requested):

pkg/lib/plot.css
martinpitt: I didn't dare to do that yet, as `git grep plot.css` shows that 5 other files import it still.

mvollmer: I see, let's do that in a separate PR then.

mvollmer (approved):

PR #15153, opened 2021-01-15
build: skip including webpack deps on NO_NPM=1

allisonkarlitskaya:
The `WEBPACK_STAMPS` are not the only way that the build directly
depends on the value defined in `WEBPACK_PACKAGES`.  Once the packages
have been built once, webpack-make will have generated the Makefile.deps
files for each of them, which will be included via `WEBPACK_DEPS`.
Those Makefile.deps files contain `TESTS +=` lines which add the
webpack-generated testcases directly to the list of tests.  This leads
to the wrong results (either extra tests running, or outright breakage)
when first building with npm and later transitioning to a no-npm build
via NO_NPM=1 or (respectively) removing the node_modules directory.

Conditionalise the inclusion of these "dependencies" Makefiles in the
same way we conditionalise other things.

A note about another approach: I first tried having the dependency
makefiles append the tests to the WEBPACK_TESTS variable, which is how
we handle everything else they generate (INPUTS, OUTPUTS, PO, etc.), and
conditionally including this new variable in TESTS, but this was causing
the tests not to run.  The reason for this is that automake treats
assignments to the TESTS variable in a special way and evaluates them
statically when creating Makefile.in.  `TESTS +=` appearing directly in
the Makefile.deps files was getting around this problem because these
files are never processed by automake.  Unfortunately, it needs to stay
that way.


martinpitt (approved):
Thanks!

FTR, I think we should move the browser unit tests to `npm run test`, and just call that from `make check` if `npm`/`node_modules/` are installed. I'd really like to remove all the npm stuff from the Makefiles, it's such a subtle and brittle entanglement..

PR #15155, opened 2021-01-15
Catching down testJournal.testBasic flake

marusak:
There is race that when stop() is called, it stops existing processes but
right after it a new process is created and thus it keeps running forever.
Normally this would not be a problem, but with this weird
combination of JS + React it can happen that messages from this service
are shown in the UI.

Honestly, the best solution would be to write it fully in React, I am really playing with the idea of doing that.

### DIFF
--- a/test/verify/check-journal
+++ b/test/verify/check-journal
@@ -245,7 +245,7 @@ ExecStart=/bin/sh -c 'for s in $(seq 10); do echo SLOW; sleep 0.1; done; sleep 1
         wait_log_present("JUST ERROR")
 
         # We cannot open this select with tests but we need to wait until it gets loaded
-        b._wait_present("#journal-service-menu option[value='check-journal']")
+        b._wait_present("#journal-service-menu option[value='just-a-service']")
         b.set_val("#journal-service-menu", "just-a-service")
         wait_log_not_present("JUST ERROR")
         wait_log_present("THE SERVICE")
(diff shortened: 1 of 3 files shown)


marusak:
On all OSes only once testJournal.testBasic was retried:
https://logs.cockpit-project.org/logs/pull-15155-20210115-140451-fc6cabaf-ubuntu-2004/log.html#307-1
But this is a completely new failure, but I might have an idea, so I'll take a look and push here.

marusak:
Ok, not a single `check-journal` failure nor retry in any OS, force triggering all tests once again.

test/verify/check-journal
```diff
@@ -136,7 +136,7 @@ ExecStart=/bin/sh -c '/usr/bin/seq 123; sleep 10'
 Description=Slowly log 10 identical lines
 
 [Service]
-ExecStart=/bin/sh -c 'for s in $(seq 10); do echo SLOW; sleep 0.1; done; sleep 10'
+ExecStart=/bin/sh -c 'sleep 5; for s in $(seq 10); do echo SLOW; sleep 0.1; done; sleep 10'
```
martinpitt, line 139: Ugh, that feels like a systemd journal bug. I don't really like sleeps, but it's harmless enough for now. Thanks!

martinpitt (approved):
Thanks for pinning this down!

marusak:
3 runs, in each all OSes passed check-journal 3 times. Seems to be working, but let's try it out once more :crossed_fingers:

test/verify/check-journal
marusak: It does a bit but it is super tight race, so almost not possible to create any reasonable reproducer except: run in loop until it hopefully fails.

marusak:
Green in 4 tries, each OS ran check-journal 12 times.

PR #15158, opened 2021-01-15
machines: jump back to the main VM page when the VM in whose Details page we are at get's removed from the state

KKoukiou:
Relevant to: rhbz#1916595

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -110,7 +110,7 @@ export class ListingTable extends React.Component {
         if (props.row.extraClasses)
             className = props.row.extraClasses.join(' ');
 
-        return <RowWrapper {...props} data-row-id={props.row.rowId} className={className} />;
+        return <RowWrapper {...props} {...props.row.props} data-row-id={props.row.rowId} className={className} />;
     }
 
     reformatColumns(columns, isExpandable) {
--- a/pkg/machines/app.jsx
+++ b/pkg/machines/app.jsx
@@ -18,10 +18,12 @@
  */
 import React from 'react';
 import PropTypes from 'prop-types';
-import { AlertGroup, Alert } from '@patternfly/react-core';
+import { AlertGroup, Alert, Button } from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { superuser } from "superuser.js";
 import cockpit from 'cockpit';
 
+import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
 import HostVmsList from "./components/vms/hostvmslist.jsx";
 import { StoragePoolList } from "./components/storagePools/storagePoolList.jsx";
 import { NetworkList } from "./components/networks/networkList.jsx";
@@ -34,6 +36,8 @@ import {
     usageStartPolling,
     usageStopPolling,
 } from "./actions/provider-actions.js";
+
+const _ = cockpit.gettext;
 
 superuser.reload_page_on_change();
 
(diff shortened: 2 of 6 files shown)


test/verify/check-machines
```diff
@@ -392,7 +392,15 @@ class TestMachines(MachineCase, StorageHelpers, NetworkHelpers):
         b.wait_in_text(".memory-usage-chart .pf-c-progress__status > .pf-c-progress__measure", "0 /")
         b.wait_in_text(".vcpu-usage-chart .pf-c-progress__status > .pf-c-progress__measure", "0%" )
 
-        self.goToMainPage()
+        # shut off of a transient VM will redirect us to the main page
+        m.execute("virsh dumpxml subVmTest1 > /tmp/subVmTest1.xml")
+        m.execute("virsh start {0} && virsh undefine {0}".format("subVmTest1"))
+        b.reload()
```
martinpitt, line 398: Why does it need a reload here? Isn't that kind of cheating? If it's because of some limitation of libvirt-dbus (no signal on transient shutoff?), please add a comment why it's necessary. Or maybe you meant `wait_page_load()`?

KKoukiou: We do get an event when a VM get's undefined while running, thus bacoming transient. And we properly update the UI.

But currently we don't have any indicator on the UI or the DOM tree which I can wait for till I am sure that the internal VM state object already updated according to the event, and we are ready to proceed with clicking the shut off button.

I can also add some info on the UI about transient VMs - that would benefit the user and we could rely on that property to wait until we proceed.

martinpitt: But in this case, wouldn't a reload() still be a race condition? Could it happen that the reload finishes before the transient VM stops? So maybe adding at least some internal DOM data-* attribute would be good here.

KKoukiou: Added the data-row-transient

martinpitt: Thanks, this looks great and more robust! I'm happy to :+1 after you hit "ready for review".

test/verify/check-machines
```diff
@@ -392,7 +392,16 @@ class TestMachines(MachineCase, StorageHelpers, NetworkHelpers):
...
 
+        # shut off of a transient VM will redirect us to the main page
         self.goToMainPage()
+        m.execute("virsh dumpxml subVmTest1 > /tmp/subVmTest1.xml")
+        m.execute("virsh start {0} && virsh undefine {0}".format("subVmTest1"))
+        b.wait_visible("tr[data-row-id=vm-subVmTest1-system][data-row-transient=true]")
+        self.goToVmPage("subVmTest1")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-forceOff")
+        b.wait_not_present("#vm-details")
```
martinpitt, line 403: This is a bit sloppy -- it would succeed if the page just stopped rendering anything, but it makes no assertion about that you really end up at the VM overview. Could you add a `wait_visible()` for the overview here?

If the page was really broken, then I figure the subsequent `b.click("#vm-subVmTest1-action-kebab button")` may fail, but such a clone action *might* also exist on the details page. So I think it's both less confusing and also more robust to explicitly say what we expect -- the "redirect us to the main page".

test/verify/check-machines
```diff
@@ -1380,16 +1389,17 @@ class TestMachines(MachineCase, StorageHelpers, NetworkHelpers):
         b.wait_in_text("#vm-{0}-state".format(name), "Paused")
         b.click("#vm-{0}-action-kebab button".format(name))
         b.click("#vm-{0}-delete".format(name))
-        self.browser.reload()
+        # Don't delete disks here because they will fail with 'storage volume does not exist'
```
martinpitt, line 1392: Please prefix with `HACK:` to keep these grep-able.

martinpitt (changes requested):
This looks good, thanks! I just have two remaining minor nitpicks.

martinpitt:
Another thing: Can you please change the commit message to say "Fixes https://bugzila..." to keep the links clickable? Also, it's really a fix, not just a side issue ("relevant to") for that bz.

KKoukiou:
@martinpitt let me check host tests behave and I will re-request your review after.

martinpitt (approved):
Cheers!

martinpitt:
Needs rebasing after #15182, and this isn't just mechanical, I'm afraid.

martinpitt (approved):

martinpitt:
Two qemu core dumps on [debian-testing](https://logs.cockpit-project.org/logs/pull-15158-20210120-195013-682a8303-debian-testing/log.html), which are known. Otherwise all green. :+1:

PR #15162, opened 2021-01-18

cockpituous:


martinpitt (approved):
Green, and the diff looks reasonable -- in particular, this does not bring back src/ws/ translation file paths, just pkg/static/ as it should be.

PR #15163, opened 2021-01-18
Fix check-metrics tests that use prepareArchive()

martinpitt:
This fixes the 7th [top flake](https://images-frontdoor.apps.ocp.ci.centos.org/failure-stats.html) ([example](https://logs.cockpit-project.org/logs/pull-15123-20210115-073403-44eb2a33-ubuntu-2004/log.html#81-2)), and another one that I saw during my local runs. I'm not 100% sure about the third commit yet, I'll either squash it or take it out. Locally tests work for me with debian-testing and ubuntu-2004, but let's get them a crunching through our bots.

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -55,6 +55,8 @@ def prepareArchive(machine, name, time):
     machine.upload(["verify/files/metrics-archives/{0}".format(name)], "/tmp/")
     machine.execute("""timedatectl set-ntp off
                        systemctl stop pmlogger
+                       # don't let NM set transient host names from DHCP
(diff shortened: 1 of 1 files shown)


martinpitt:
Dang, so close! All green except on [debian-stable](https://logs.cockpit-project.org/logs/pull-15163-20210118-081403-69acf9f2-debian-stable/log.html#311). So maybe that hack isn't obsolete yet, but the root cause was something else. Retrying without the big cleanup.

martinpitt:
Oui, ça marche !

marusak (approved):
A nice one! Thanks :)

PR #15164, opened 2021-01-18
test: Fix TestLogin.testExpired for pam_pwquality with retry=1

martinpitt:
This test is written with the assumption that trying to change the PAM
password to a bad one will have more than one retry. This isn't the case
on some images that use authselect, such as our centos-8-stream or
rhel-8-* images.

This was previously hacked around in the image *.install scripts [1],
but also hits us in Fedora/RHEL downstream gating tests, which is why
that test is currently skipped [2]. Adjust the PAM config in the test
itself, so that it can run in any testbed, and we can re-enable the test
downstream.

[1] https://github.com/cockpit-project/bots/pull/1555
[2] https://src.fedoraproject.org/rpms/cockpit/blob/917966b/f/tests/run-test.sh#_75


martinpitt:
Triggering additional centos8 test against https://github.com/cockpit-project/bots/pull/1555 . After this lands, I need to backport it to rhel-8.3 branch.

martinpitt:
Ah, [this is not a thing on Debian](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15164-20210118-090348-a2087907-debian-stable/log.html#304)

PR #15165, opened 2021-01-18
pkg/lib: Fixes for eslint

mvollmer:
These things turn up when running "npm run eslint" or "npm run eslint:fix".

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -1,5 +1,6 @@
 const path = require("path");
 const glob = require("glob");
+const fs = require("fs");
 const childProcess = require("child_process");
 const po2json = require('po2json');
 const Jed = require('jed');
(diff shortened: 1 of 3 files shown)


pkg/lib/cockpit-po-plugin.js
```diff
@@ -24,15 +25,17 @@ module.exports = class {
...
+        var statement;
+        var ret = null;
         const plurals = header["plural-forms"];
 
         if (plurals) {
             try {
                 /* Check that the plural forms isn't being sneaky since we build a function here */
                 Jed.PF.parse(plurals);
-            } catch(ex) {
-                fatal("bad plural forms: " + ex.message, 1);
```
mvollmer, line 35: I don't know where `fatal` would come from.  Is it a real thing that we can call?

marusak: It used to be defined in `po/po2json` as:
```
function fatal(message, code) {                                                 
    console.log((filename || "html2po") + ": " + message);                      
    process.exit(code || 1);                                                    
} 
```
See https://github.com/cockpit-project/cockpit/commit/72c180d78848946b939bbab23b063eee246336d8

martinpitt: Oops! Thanks for catching that!

martinpitt (approved):
Thanks for cleaning up after me! :-)

PR #15166, opened 2021-01-18
lib: Fix cockpit-po-plugin instantiation without options

martinpitt:
starter-kit and other projects that just have one webpack have no need
to specify any option. The plugin previously crashed with

    TypeError: Cannot read property 'subdir' of undefined

 in that case. Default to an empty object if options are undefined.


martinpitt:
Marking as release-blocker, so that we can unblock https://github.com/cockpit-project/starter-kit/pull/414 after this Wednesday.

marusak (approved):

PR #15173, opened 2021-01-19
machines: Disable boot order editing for running guests

KKoukiou:
It was implemented previously quite naively, resulting in wrong warnings
being shown in the UI when there was a mismatch between the active and
inactive XML bootable devices.

For example:
* The user would attach a transient disk/NIC to the running guest.
* That would cause a mismatch between the enumerated bootable devices
  between the active and the inactive XML configuration,
* The UI would show a warning, but there would be no way for the user to
  find out that the transient device they just added was the reason
  for the warning.

Doing a 1-1 comparison of the bootable devices between the active and
inactive XML is a real pain, since these devices don't have some unique
common identifier.

So let's remove this functionality of allowing the users to edit boot
order on running VMs to avoid more issues in the future.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1915765

### DIFF
--- a/pkg/machines/components/vm/overview/bootOrder.jsx
+++ b/pkg/machines/components/vm/overview/bootOrder.jsx
@@ -26,11 +26,9 @@ import {
     ListViewItem,
 } from 'patternfly-react';
 
-import { Button, Alert, Modal } from '@patternfly/react-core';
-import WarningInactive from '../../common/warningInactive.jsx';
+import { Button, Modal, Tooltip } from '@patternfly/react-core';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import {
-    getBootOrderDevices,
     findHostNodeDevice,
     getSortedBootOrderDevices,
     rephraseUI,
@@ -255,37 +253,6 @@ class BootOrderModal extends React.Component {
     render() {
         const { nodeDevices, vm } = this.props;
         const idPrefix = vmId(vm.name) + '-order-modal';
-
-        /**
-         * Returns whetever state of device represented in UI has changed
-         *
-         * @param {object} device
-         * @param {number} index order of device in list
-         * @returns {boolean}
-         */
-        function deviceStateHasChanged(device, index) {
-            // device was selected
-            if (device.checked && !device.initialOrder)
-                return true;
-
-            // device was unselected
-            if (!device.checked && device.initialOrder)
-                return true;
-
-            // device was moved in boot order list
-            if (device.initialOrder && device.initialOrder !== index + 1)
-                return true;
-
-            return false;
-        }
-
-        const showWarning = () => {
-            if (vm.state === "running" &&
-                this.state.devices.some((device, index) => deviceStateHasChanged(device, index))) {
-                return <Alert isInline variant='warning' id={`${idPrefix}-min-message`} title={_("Changes will take effect after shutting down the VM")} />;
-            }
-        };
-
         const defaultBody = (
             <div className="list-group dialog-list-ct">
                 <ListView className="boot-order-list-view">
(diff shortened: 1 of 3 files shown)


pkg/machines/components/vm/overview/bootOrder.jsx
```diff
@@ -360,26 +326,17 @@ function getBootOrder(vm) {
...
 
     return (
         <>
             {bootOrderShow && <BootOrderModal close={() => setBootOrderShow(false)} vm={vm} dispatch={dispatch} nodeDevices={nodeDevices} />}
             {getBootOrder(vm)}
-            { vm.persistent && vm.state === "running" && bootOrderChanged() && <WarningInactive iconId="boot-order-tooltip" tooltipId="tip-boot-order" /> }
-            <Button variant="link" className="edit-inline" isInline isDisabled={!vm.persistent} onClick={setBootOrderShow}>
-                {_("edit")}
-            </Button>
+            {vm.state == 'shut off' ? modalButton : <Tooltip content={_("Only editable when the guest is shut off")}>{modalButton}</Tooltip>}
```
skobyda, line 339: Isn't the only problem just with showing the warning? So wouldn't it solve the problem if we just get rid of the warning which does not work, but still allow to edit running guests?

KKoukiou: It's only the warning problematic indeed. What you suggest would be ideal, I just thought it's too much work. RIght now we compare devices betwen the inactive and active XML, according to the index (order) in which they are presented. That's wrong, because devices can be added, removed, then the order changes. So a solution would be indeed to have a device comparison function between the two XMLs. Is it something that you 're interested to work on?

(the only reason I removed live editing is because I was lazy (busy with other things) to implement the functionality properly.

skobyda: > (the only reason I removed live editing is because I was lazy (busy with other things) to implement the functionality properly.

But this functionality is implemented properly, no?
In pkg/machines/libvirt-xml-update.js there is a function updateBootOrder . When it wants to change a boot order of some device, it tries to find the proper device by various identifiers (for disks it's a target, for a network it's a MAC, ....). It does not just find the device by its index. 
And also transient devices from active XML should not bother us for live editing, because only devices visible in Boot Order Modal are the permanent devices (I tried to add transient disk to a running VM, it was not visible in boot order modal). And when boot order is updated, only inactive XML is updated.

So If I'm not wrong and above mentioned is correct, I propose:
- Let's keep the live editing. 
- Let's keep the warning in the boot order modal. Since we are dealing only with permanent devices and inactive XML, any change would need a VM reboot.
- Let's remove warning icon in overview card. That one is truly broken in some cases.

KKoukiou: All what you describe above is correct. I just thought it's weird for every other attribute to show the warning triangle in the overview, but not for the bootorder.

Then transient disks and networks which show up in the respective lists, will also not appear in the boot order dialog, which might cause the users to get confused.

skobyda: Okay, makes sense to keep consistent. 
Also considering that live edit doens't matter that much here, since user has to turn of the machine anyway to use the functionality of the boot order, so I'm not so keen on having the live edit functionality.

skobyda:
I see one machines failure, retrigerring
https://logs.cockpit-project.org/logs/pull-15173-20210128-063522-0c9c81ea-centos-8-stream/log.html#80

pkg/machines/components/vm/overview/bootOrder.jsx
KKoukiou: Right I was thinking the same thing.

skobyda (approved):
If the above mentioned test passes or it's a known flake, I'm okay with merging this

PR #15179, opened 2021-01-20
package.json: Lock @patternfly/react-styles version

martinpitt:
It's part of patternfly-react, so we should update it in a controlled
fashion. Yesterday's 4.7.25 release does not work with the older react-*
releases that we already pinned.

Let's update all these in lockstep after today's release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "@patternfly/patternfly": "4.70.2",
     "@patternfly/react-console": "4.2.7",
     "@patternfly/react-core": "4.84.4",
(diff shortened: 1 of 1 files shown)


martinpitt:
Super-urgent, as currently all our builds/tests [are broken](https://logs.cockpit-project.org/logs/pull-15154-20210120-043347-fedec0d8-debian-testing/log.html). This just brings us back to the status quo from last night, so only triggering a single fast test.

marusak (approved):
Thanks!!

PR #15101, opened 2021-01-06
networkmanager: start rewrite in react

KKoukiou:
In some cases the code was duplicated previously, thus this commit tried
to create some shared components:
see NetworkGraphs, FirewallSwitch

* [x] blocked on #14913
* [x] s/Configure/edit/
* [x] MAC address alignment

PR #15180, opened 2021-01-20
 networking: Fix closing of "Connection lost" dialog

mvollmer:


mvollmer:
This has regressed in #15101 but the tests didn't catch it.  Let's get it into master on its own to ensure that the test itself if correct.

mvollmer:
Oh, looks like it's already broken in master...

martinpitt (approved):
Thanks!

PR #15182, opened 2021-01-20
Split test/verify/check-machines into multiple files and try to robustify testAddDiskDirPool massive flake

KKoukiou:


martinpitt:
This looks great, thanks! [firefox failed early](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15182-20210120-151058-d07880b0-fedora-32-firefox/log.html), I retried that. [rhel-8-4](https://logs.cockpit-project.org/logs/pull-15182-20210120-151058-d07880b0-rhel-8-4/log.html) is an unrelated failure, and [debian-testing](https://logs.cockpit-project.org/logs/pull-15182-20210120-151058-d07880b0-debian-testing/log.html#302) is now an unexpected crash. At first I thought the test rearranging would make a naughty not apply any more, but we don't have any against qemu itself, and [this has happened before](https://images-frontdoor.apps.ocp.ci.centos.org/failure-stats.html). So I'd totally +1 this once firefox looks reasonable, as it at least makes tests easier to handle.

test/verify/check-machines
```diff
@@ -69,200 +67,7 @@ def release_target(used_targets, target):
 @skipImage("Atomic cannot run virtual machines", "fedora-coreos")
 @nondestructive
 @no_retry_when_changed
```
martinpitt: I think the `@decorators` on the parent class don't work. The individual test classes repeat them, so that's mostly harmless.

martinpitt (approved):
Please land when firefox finishes and is sufficiently green, this looks good!

marusak:
> Please land when firefox finishes and is sufficiently green, this looks good!

Can we get 3 full retries so we know we actually are fixing it and were not just lucky? :)

martinpitt:
I thought the point of this was mostly to make it easier to debug, but indeed the "user smaller images" part could help. Sure, we can do a full retry.

KKoukiou:
Ok I will force-trigger: the current run was 151058-d07880b0 in https://logs.cockpit-project.org/logs/

martinpitt:
[same qemu crash](https://logs.cockpit-project.org/logs/pull-15182-20210120-165853-d07880b0-debian-testing/log.html#281) on debian-testing again, let's fix this ASAP -- but it'll require a naughty, so separate PR. [fedora-32](https://logs.cockpit-project.org/logs/pull-15182-20210120-173356-d07880b0-fedora-32/log.html#292-2) failed `testCPUModel`, but this is also nothing new. By nature this PR is more prone to check-machines failures due to the `@no_retry_when_changed`.

Issue #15174, opened 2021-01-19
Remove custom CSS colors


garrett:
Originally, we needed a bunch of colors mapped because we were faking PF4 support. Additionally, some have been used across Cockpit for various reasons.

We should remove the unused ones and either replace usages of the custom colors with PatternFly ones directly and/or map some of the Cockpit-specific ones to PF as an alias. (The method of what to do depends on context.)

garrett:
Rebased & resolved conflicts

PR #15183, opened 2021-01-20
lib/page: Remove unused colors

garrett:
- [x] Remove color variables that are unused
- [x] Replace colors with official PF4 colors
- [x] Change references to PF4 across Cockpit
- [x] Remove custom fonts; switch to PF4 fonts across Cockpit
- [x] Move Nav colors to the Nav stylesheet

### DIFF
--- a/pkg/apps/application.css
+++ b/pkg/apps/application.css
@@ -4,10 +4,10 @@
 @import "../lib/cockpit-components-table.scss";
 
 .app-list-empty {
-    color: var(--color-subtle-copy);
+    color: var(--ct-color-subtle-copy);
     text-align: center;
-    border-top: 1px solid var(--color-border);
-    border-bottom: 1px solid var(--color-border);
+    border-top: 1px solid var(--ct-color-border);
+    border-bottom: 1px solid var(--ct-color-border);
 }
 
 .app-list-empty .spinner {
@@ -57,7 +57,7 @@
 
 .app td {
     border-top: 0px !important;
-    border-bottom: 1px solid var(--color-border);
+    border-bottom: 1px solid var(--ct-color-border);
 }
 
 .app td:nth-child(3) {
--- a/pkg/lib/cockpit-components-table.scss
+++ b/pkg/lib/cockpit-components-table.scss
@@ -162,14 +162,14 @@
 // Special handling for rows with errors
 .pf-c-table tbody tr:first-child.error {
     &, tbody.pf-m-expanded > & {
-        background-color: var(--color-ct-list-critical-bg) !important; /* keep red background when expanded */
-        border-top: 1px solid var(--color-ct-list-critical-border);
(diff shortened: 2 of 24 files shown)


garrett:
Fixes #15174

martinpitt:
@garrett: This looks great, thanks! Is this draft because there is more work that you want to do? (check boxes in the description). Or does it cause regressions in its current form? In the former case, we could land it and do the rest in separate PRs.

garrett:
Thanks! Yeah, it's getting pretty close to review state, but I want to do another pass over it before it's ready for others to review. Thursday evening came up quickly, as it often does. :wink:

garrett:
If https://github.com/cockpit-project/cockpit/pull/15360#discussion_r578254069 lands before this PR, the color variable in it will need updating.

garrett:
Rebased and resolved conflicts.

garrett:
We'll want to squash this down before merging. I'm leaving it separate commits right now for review.

martinpitt:
@garrett, are you ok with squashing everything into one commit (fine for me)? Or you want to keep several? (then it needs a force-push).

garrett:
Yeah, I think having everything eventually in one commit is fine. I got lazy with the last few commits and didn't prefix them, especially as they affected multiple parts of Cockpit. But if they're separate, they should be reworded (or merged with other related ones).

Do you think any should be batched together, or should I just have one single commit?

martinpitt:
single commit WFM, that's fine. We can still use some of the follow-up commit messages.

garrett:
Squashed everything to 1 commit, moved commit messages into the extra part of the message, and cleaned up the text a little.

martinpitt:
I wanted to confirm that our (s)css build toolchain recognizes undefined variables etc, and it *does not*. :exploding_head: So during review we need to make extra sure that none of the removed variables is still in use.

martinpitt (approved):
This looks great, thanks a lot for the cleanups!

garrett:
> I wanted to confirm that our (s)css build toolchain recognizes undefined variables etc, and it does not. :exploding_head: So during review we need to make extra sure that none of the removed variables is still in use.

The CSS variables can be set and used by anything. There's no way to know. That is: One file can optionally set it and another can optionally use it. As can the JavaScript (as it has full read/right access to CSS variables (aka: custom properties)).

Therefore, there's no way to _properly_ check in an automated way, unless it considers all the files and has an understanding about how the JS and CSS works. I guess the browsers themselves will know most of the time, unless the JavaScript modifies the variable and the codepath isn't executed (yet).

That said, I searched across all files for each variable in the page.scss file to see if it was fine to remove, a candidate for renaming, etc. And then, after reviewing each, I made decisions on what to do for each one.

As for the SCSS variables: They can be defined, but if they're not used, then they go away. SCSS variables are compiled out. Therefore, there's no real need to worry about them.

PR #15185, opened 2021-01-21
networking: Fix confusion between master and group

mvollmer:
The NM data structure says "master", our own model says "group".

### DIFF
--- a/pkg/networkmanager/interfaces.js
+++ b/pkg/networkmanager/interfaces.js
@@ -588,7 +588,7 @@ function NetworkManagerModel() {
             }
         };
 
-        if (!settings.connection.group) {
+        if (!settings.connection.master) {
             result.ipv4 = get_ip("ipv4", ip4_address_from_nm, ip4_route_from_nm, utils.ip4_to_text);
             result.ipv6 = get_ip("ipv6", ip6_address_from_nm, ip6_route_from_nm, utils.ip6_to_text);
         }
--- a/test/verify/check-networking-bond
+++ b/test/verify/check-networking-bond
@@ -82,6 +82,15 @@ class TestBonding(NetworkCase):
         b.click("#network-bond-settings-dialog button:contains('Cancel')")
         b.wait_not_present("network-bond-settings-dialog")
 
+        # Navigate to the first member, check it, navigate back
+        b.click("#network-interface-members tr[data-interface='%s'] td:first-child" % iface1)
+        b.wait_in_text("#network-interface-name", iface1)
+        b.wait_in_text("#network-interface-settings", "MTU")
+        b.wait_not_in_text("#network-interface-settings", "IPv4")
+        b.wait_not_in_text("#network-interface-settings", "IPv6")
+        b.click("#network-interface-settings a:contains(tbond)")
+        b.wait_in_text("#network-interface-name", "tbond")
+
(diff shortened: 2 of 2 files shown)


mvollmer:
A test, maybe, non?

mvollmer:
> A test, maybe, non?

Added.

martinpitt:
I'm a bit surprised that check-networking-bond tests are not run three times? Wasn't that the point of our "affected tests auto-retry"?

martinpitt (approved):
Thanks!

marusak:
ehm, what? When I did `tests-scan` on then just ran the whole command, but the commit I see there is the previous version and not the current one. That means, this did not test the new changes.

This is GH issue, I can reproduce like this:
```
$ git fetch upstream pull/15185/head:mvo-x
remote: Enumerating objects: 9, done.
remote: Counting objects: 100% (9/9), done.
remote: Compressing objects: 100% (1/1), done.
remote: Total 5 (delta 4), reused 5 (delta 4), pack-reused 0
Unpacking objects: 100% (5/5), 490 bytes | 21.00 KiB/s, done.
From github.com:cockpit-project/cockpit
 * [new ref]             refs/pull/15185/head -> mvo-x
 * 
$ git checkout mvo-x 
Switched to branch 'mvo-x'

$ git show
commit dadf8c69a889ef2e48dd00729e37020c2e1f407a (HEAD -> mvo-x)
Author: Marius Vollmer <mvollmer@redhat.com>
Date:   Thu Jan 21 13:34:36 2021 +0200

    networking: Fix confusion between master and group
    
    The NM data structure says "master", our own model says "group".

diff --git pkg/networkmanager/interfaces.js pkg/networkmanager/interfaces.js
index 877af0629..7b6be4924 100644
--- pkg/networkmanager/interfaces.js
+++ pkg/networkmanager/interfaces.js
@@ -588,7 +588,7 @@ function NetworkManagerModel() {
             }
         };
 
-        if (!settings.connection.group) {
+        if (!settings.connection.master) {
             result.ipv4 = get_ip("ipv4", ip4_address_from_nm, ip4_route_from_nm, utils.ip4_to_text);
             result.ipv6 = get_ip("ipv6", ip6_address_from_nm, ip6_route_from_nm, utils.ip6_to_text);
         }

```

So when we checkout head, we don't get the new changes, but we get the old version. This is a biggie as we cannot rely on our testing really. And it explains those rare cases when we landed something with green tests only to discover that those tests were broken.

But we have correct SHA, so we might need to use that to get changes and not `pull/<no>/head`?

marusak:
Please ignore `debian-testing@bots#1569` context. It will fail in one test I just was testing some changes and used this PR.

marusak:
```
Command: PRIORITY=0005 ./make-checkout --verbose --repo=cockpit-project/cockpit --rebase=master c32415c50aec062a1c21dbf31b53191c0f382d0c && cd make-checkout-workdir && TEST_OS=fedora-32 COCKPIT_BOTS_REF=master TEST_SCENARIO=container-bastion ../tests-invoke --pull-number 15185 --revision c32415c50aec062a1c21dbf31b53191c0f382d0c --repo cockpit-project/cockpit
```
seems to now pick the right revision and now affected tests should run 3 times.

marusak:
Fully green PR!

PR #15186, opened 2021-01-21
Fix TestSystemInfo.testCPUSecurityMitigationsDetect flake

martinpitt:
The first commit is good, and will fix the follow-up errors that we see often:
```
subprocess.CalledProcessError: Command 'if selinuxenabled 2>/dev/null; then chcon --reference /proc/cmdline /run/cmdline; fi && mount --bind /run/cmdline /proc/cmdline' returned non-zero exit status 32.
```

But the initial failure is in b.reload():
```
{"exceptionId":1,"text":"Uncaught (in promise)","lineNumber":0,"columnNumber":0,"scriptId":"1","url":"http://127.0.0.2:9191/cockpit/$91dd94251495964ea7683d4428bfbe9207ff1951456cd17b1642ef718eb9f1c5/system/hwinfo.html#/","exception":{"type":"object","className":"te","description":"te","objectId":"{\"injectedScriptId\":14,\"id\":1}","preview":{"type":"object","description":"te","overflow":false,"properties":[{"name":"problem","type":"string","value":"internal-error"},{"name":"exit_status","type":"number","value":"-1"},{"name":"exit_signal","type":"object","value":"null","subtype":"null"},{"name":"message","type":"string","value":"Internal error"},{"name":"toString","type":"function","value":""}]}},"executionContextId":14}
  File "/usr/lib64/python3.9/unittest/case.py", line 550, in _callTestMethod
    method()
  File "/work/cockpit-project/bots/make-checkout-workdir/test/verify/check-system-info", line 512, in testCPUSecurityMitigationsDetect
    spoof_threads(2, True, True, 'param1 param2 nosmt param3=value3')
  File "/work/cockpit-project/bots/make-checkout-workdir/test/verify/check-system-info", line 493, in spoof_threads
    b.reload()
  File "/work/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 139, in reload
    self.wait_js_cond("ph_select('iframe.container-frame').every(function (e) { return e.getAttribute('data-loaded'); })")
  File "/work/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 382, in wait_js_cond
    result = self.cdp.invoke("Runtime.evaluate",
  File "/work/cockpit-project/bots/make-checkout-workdir/test/common/cdp.py", line 109, in invoke
    res = self.command(cmd)
  File "/work/cockpit-project/bots/make-checkout-workdir/test/common/cdp.py", line 136, in command
RuntimeError: te
```

This is absolutely useless. I temporarily enabled debug builds for all OSes, and want to catch it in the act, to hopefully get a better exception. Running all the check-system-info tests three times here should certainly help. It only happens on centos-8-stream and rhel-8-4, apparently not anywhere else according to our flake stats.

### DIFF
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -479,9 +479,7 @@ class TestSystemInfo(MachineCase):
         m = self.machine
 
         self.restore_dir("/usr/local/bin")
-
-        self.login_and_go('/system/hwinfo')
-        b.wait_in_text("#hwinfo", "CPU")
+        m.start_cockpit()
 
         def spoof_threads(threads_per_core, expect_link_present, expect_smt_state=None, cmdline=None):
             m.write('/usr/local/bin/lscpu', lscpu.format(threads_per_core))
@@ -492,8 +490,7 @@ class TestSystemInfo(MachineCase):
                 m.execute('mount --bind /run/cmdline /proc/cmdline && rm /run/cmdline')
 
             try:
-                b.reload()
-                b.enter_page('/system/hwinfo')
+                b.login_and_go('/system/hwinfo')
 
                 if not expect_link_present:
                     b.wait_visible('#hwinfo th:contains(CPU)')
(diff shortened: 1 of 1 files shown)


martinpitt:
I retried several times, but I can't get it to crash in debug mode on the infra (of course!), and I can't get it to crash locally either. I tried building in debug, production mode, pre-running VM, freshly started VM, one or two parallel VMs, and manually hitting reload in my browser like a crazy person. So, plan B :grin:

martinpitt:
The check-system-info tests succeeded in all 45 cases.

marusak (approved):
nice, thanks for looking into it!

PR #15197, opened 2021-01-21
tests: remove no_retry_when_changed from  check-machines-storage-pools

KKoukiou:
Let's see how this behaves.

### DIFF
--- a/pkg/machines/components/storagePools/storagePool.jsx
+++ b/pkg/machines/components/storagePools/storagePool.jsx
@@ -64,26 +64,18 @@ export const getStoragePoolRow = ({ storagePool, vms, dispatch, onAddErrorNotifi
                    }))} />
     );
 
-    const overviewTabName = (
-        <div id={`${idPrefix}-overview`}>
-            {_("Overview")}
-        </div>
-    );
-    const storageVolsTabName = (
-        <div id={`${idPrefix}-storage-volumes`}>
-            {_("Storage volumes")}
-        </div>
-    );
     const tabRenderers = [
         {
-            name: overviewTabName,
+            name: _("Overview"),
             renderer: StoragePoolOverviewTab,
-            data: { storagePool }
+            data: { storagePool },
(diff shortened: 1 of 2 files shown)


martinpitt:
@KKoukiou : This doesn't work. We only do the "run tests three times" if ≤ 3 tests got changed in the PR. Otherwise we consider it a "PR that does a major test reorganization" and don't penalize it by requiring that basically *all* tests have to succeed three times in a row. You need to split this up into several smaller PRs. That makes sense anyway, to be able to investigate things in smaller bites.

KKoukiou:
> @KKoukiou : This doesn't work. We only do the "run tests three times" if ≤ 3 tests got changed in the PR. Otherwise we consider it a "PR that does a major test reorganization" and don't penalize it by requiring that basically _all_ tests have to succeed three times in a row. You need to split this up into several smaller PRs. That makes sense anyway, to be able to investigate things in smaller bites.

Oh did not know what, thanks for the clarification.

martinpitt:
@KKoukiou : FYI, that is done [here](https://github.com/cockpit-project/cockpit/blob/master/test/common/run-tests#L216)

marusak:
> We only do the "run tests three times" if ≤ 3 tests got changed in the PR.

3 files not tests, so you can do it in 3 files at once here.

KKoukiou:
> > We only do the "run tests three times" if ≤ 3 tests got changed in the PR.
> 
> 3 files not tests, so you can do it in 3 files at once here.

sure I understood it, I just actually think it makes sense to merge one by one the PRs for the removal of the decorators for the machines tests. I don't have high expectation for their stability.

martinpitt (approved):
Very nice! One down, 8 to go -- and it's good to start with the ones that are already stable, to avoid regressing those. Tests have not fully finished yet, but are looking good so far.

martinpitt (approved):
Epharisto!

PR #15198, opened 2021-01-21
a11y: Fix some general issues

marusak:
I guess tests would not hurt, would they :) (no-test while I write them, want to give @garrett chance to take a look)

### DIFF
--- a/pkg/shell/indexes.js
+++ b/pkg/shell/indexes.js
@@ -241,14 +241,7 @@ function MachinesIndex(index_options, machines, loader, mdialogs) {
             document.getElementById("main").style.removeProperty('--ct-color-host-accent');
         }
 
-        let component_manifest = state.component;
-        // If `state.component` is not known to any manifest, find where it comes from
-        if (compiled.items[state.component] == undefined) {
-            let s = state.component;
-            while (s && compiled.items[s] == undefined)
-                s = s.substring(0, s.lastIndexOf("/"));
-            component_manifest = s;
-        }
+        const component_manifest = find_component(state, compiled);
 
         // Filtering of navigation by term
         function keyword_filter(item, term) {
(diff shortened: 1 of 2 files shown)


marusak:
I found two other issues while I was writing tests, so fixing them here as well.

pkg/shell/indexes.js
```diff
@@ -550,7 +550,17 @@ function MachinesIndex(index_options, machines, loader, mdialogs) {
             $("#machine-spinner").toggle(frame && !$(frame).attr("data-ready"));
             $(frame).css('display', 'block');
             $(frame).attr('data-active', 'true');
-            item = compiled.items[state.component];
+
+            let component_manifest = state.component;
+            // If `state.component` is not known to any manifest, find where it comes from
+            if (compiled.items[state.component] == undefined) {
```
martinpitt, line 556: `== undefined` is rather sloppy, as e.g. `null == undefined` is true. If you really *only* want `undefined`, use `===`. If you just care about falsy (as I think is more appropriate here), use `!compiled.items[state.component]`.

pkg/shell/indexes.js
```diff
@@ -562,6 +562,7 @@ function MachinesIndex(index_options, machines, loader, mdialogs) {
 
             item = compiled.items[component_manifest];
             label = item ? item.label : "";
+            $(frame).attr('title', label);
```
martinpitt, line 565: Does it really make sense to explicitly set an empty title, instead of having no title at all?

martinpitt (commented):
Thanks! Some minor questions that I'd like to discuss a bit, but nothing too important. If you prefer, ok to land as-is.

pkg/shell/indexes.js
marusak: I copied this from 8155456.
It really just needs to check, that accessing of attribute was not successful. I want empty string to not satisfy this condition.
```
undefined == undefined
true
undefined == null
true
undefined == ""
false
```

But it is true, that accessing non-existing element returns undefined:
```
x={a:1}
Object { a: 1 }

x.foo
undefined
```

I'll update to use `===` in both places.
```

pkg/shell/indexes.js
marusak: I guess it does not, I'll update the logic.

martinpitt (approved):
Thanks! Resetting unrelated-test-failures label (we should have a workflow for that :grin: )

PR #15200, opened 2021-01-21
test: Only disable ntp if available and active

marusak:
In rhel-8.4 it is not available, see https://bugzilla.redhat.com/show_bug.cgi?id=1918340

As part of https://github.com/cockpit-project/bots/pull/1551
It still will have one test failing, but that one we need to naughtify


martinpitt:
That doesn't seem to work in [check-metrics](https://logs.cockpit-project.org/logs/pull-15200-20210121-201719-f0fabafc-rhel-8-4-bots-1551/log.html#302) yet?

marusak:
> That doesn't seem to work in check-metrics yet?

Ah, forgot to convert one case, fixed.

martinpitt (approved):
Nice, thanks!

PR #15201, opened 2021-01-22
test: Allow for asnychronous DOM changes in storagelib

mvollmer:
The code was careful not to click on non-existing elements, for
example, but it assumed a stable DOM so that there are no
check-before-use races.  This was true in the days of PhantomJS but is
no longer true with CDP.

Thus, we now allow a testlib.Error and retry.

### DIFF
--- a/test/verify/storagelib.py
+++ b/test/verify/storagelib.py
@@ -167,29 +167,41 @@ class StorageHelpers:
             tab = row + " .ct-listing-panel-body:nth-child(%d)" % (tab_index + 1)
             cell = tab + " dt:contains(%s) + *" % title
 
-            if not b.is_present(row + ".pf-m-expanded"):
-                if not b.is_present(row_item):
-                    return False
-                b.click(row_item)
+            # The DOM might change at any time while we are inspecting
+            # it, so we can't reliably test for a elements existence
+            # before clicking on it, for example.  Instead, we just
+            # click and catch the testlib.Error that happens when it
+            # is not there.  However, the click itself will wait for a
+            # timeout when the element is missing, so we check anyway
+            # before trying, just to speed things up.
+
+            try:
                 if not b.is_present(row + ".pf-m-expanded"):
+                    if not b.is_present(row_item):
+                        return False
+                    b.click(row_item)
+                    if not b.is_present(row + ".pf-m-expanded"):
+                        return False
+
+                if not b.is_present(tab) or not b.is_visible(tab):
+                    if not b.is_present(tab_btn):
+                        return False
+                    b.click(tab_btn)
+                    if not b.is_visible(tab):
+                        return False
+
+                if not b.is_present(cell) or not b.is_visible(cell):
                     return False
 
-            if not b.is_present(tab) or not b.is_visible(tab):
-                if not b.is_present(tab_btn):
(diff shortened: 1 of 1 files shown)


test/verify/storagelib.py
```diff
@@ -167,29 +167,42 @@ def check():
...
-            if not b.is_present(cell) or not b.is_visible(cell):
+                if val is not None and val in b.text(cell):
+                    return True
+                if alternate_val is not None and alternate_val in b.text(cell):
+                    return True
+                if cond is not None and cond(cell):
+                    return True
+                return False
+            except Error as ex:
+                print("RETRY", ex)
```
marusak, line 204: In the first run this storage test did not fail even once, but nor this RETRY was ever caught, lets try again. Force triggering it all.

marusak: I could reproduce it yesterday, so I'll try it locally and see, if I can hit that RETRY.

marusak: ok, I could reproduce it locally twice, once with
```
RETRY eval_js(ph_is_visible("#detail-content > article > div > table > tbody:nth-of-type(1) .ct-listing-panel-body:nth-child(2) dt:contains(Name) + *")): #detail-content > article > div > table > tbody:nth-of-type(1) .ct-listing-panel-body:nth-child(2) dt:contains(Name) + * not found
```
and once with
```
RETRY timeout
wait_js_cond(ph_is_visible("#detail-content > article > div > table > tbody:nth-of-type(1) .ct-listing-panel-body:nth-child(2) dt:contains(Name) + *")): {'exceptionId': 1, 'text': 'Uncaught (in promise) Error: condition did not become true', 'lineNumber': 233, 'columnNumber': 9, 'scriptId': '3', 'stackTrace': {'callFrames': [{'functionName': 'PhWaitCondTimeout', 'scriptId': '3', 'url': '', 'lineNumber': 232, 'columnNumber': 8}, {'functionName': '', 'scriptId': '3', 'url': '', 'lineNumber': 244, 'columnNumber': 23}]}, 'exception': {'type': 'object', 'subtype': 'error', 'className': 'PhWaitCondTimeout', 'description': 'Error: condition did not become true\n    at <anonymous>:245:24', 'objectId': '{"injectedScriptId":15,"id":2}'}}
cdp: {'source': 'network', 'level': 'error', 'text': 'Failed to load resource: the server responded with a status of 401 (Authentication failed)', 'timestamp': 1611315578506.5962, 'url': 'http://127.0.0.2:9091/cockpit/login', 'networkRequestId': '3436487.12'}
```
and in both cases the test eventually succeeded so it seems to be working.

marusak: Do you want to keep that print? Feels a bit pointless really.

mvollmer: > ok, I could reproduce it locally twice,

Cool, thanks!

mvollmer: > Do you want to keep that print? Feels a bit pointless really.

Yes, I agree, let's remove it.

mvollmer: > Yes, I agree, let's remove it.

Done.

marusak (approved):

PR #15202, opened 2021-01-22
machines: fix detection of the nic source when displaying the warning in the modal

KKoukiou:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1911657

### DIFF
--- a/test/verify/check-machines-nics
+++ b/test/verify/check-machines-nics
@@ -80,143 +80,30 @@ class TestMachinesNICs(VirtualMachinesCase):
 
         self.goToVmPage("subVmTest1")
 
-        class NICAddDialog(object):
-            def __init__(
-                # We have always have to specify mac and source_type to identify the device in xml and $virsh detach-interface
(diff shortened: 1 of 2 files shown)


KKoukiou:
Whoever reviews this I need an advice for how to use NICAddDialog in order that I avoid all this test_obj passing in order to access the self.browser from MachinesCase and the self.assertEqual from unittest

KKoukiou:
Gr why I can't use these methods from netlib :'( ?

test/verify/check-machines-nics
```diff
@@ -20,6 +20,7 @@
 import parent
 from machineslib import *
 from testlib import *
+from netlib import  *
```
KKoukiou, line 23: Oh this is redundant.

martinpitt (approved):
Thanks! The `self.NICAddDialog(self)` syntax is rather weird, and I think it ought to be possible to drop the first `self.`. But this is not important at all, the bug fix is.

PR #15203, opened 2021-01-22
test: Look for and prefer "firefox-nightly" executable

mvollmer:
So that people can easily have both the regular Firefox for their
normal use and Firefox Nightly for the tests.


marusak (approved):
Not that `firefox-nightly` is the name of executable from nightly (it is still `firefox`) but this makes it easy to have firefox side by side with normal firefox.

martinpitt (approved):
Simple enough, thank you!

PR #15204, opened 2021-01-22
tests: remove no_retry_when_changed from check-machines-snapshots

KKoukiou:


KKoukiou:
There are not failures or retries from check-machines-snapshots

marusak (approved):
Thanks!

Btw. I noticed today already twice that check-machines-console went wild, e.g. here https://logs.cockpit-project.org/logs/pull-15204-20210122-111733-326ec020-rhel-8-4-distropkg/log.html#79-2

PR #15206, opened 2021-01-22
shell: Only start "sudo" when there is a password

mvollmer:


martinpitt:
Do we have a chance to reproduce that in a test? Like, actually dropping the "incorrect password attempt." messages from `allow_authorize_journal_messages()`? this should then fail on -distropkg (and then we can add it back conditionally). Likewise, would be good to have a test case for passwordless sudo, that a single click really keeps working. Thanks!

mvollmer:
> Like, actually dropping the "incorrect password attempt." messages from `allow_authorize_journal_messages()`?

Good idea.

pkg/shell/machines/machines.js
```diff
@@ -42,12 +42,31 @@ export function host_superuser_storage_key(host) {
...
-    else
-        return null;
+        value = window.localStorage.getItem(key);
+
+    /* When connecting, we can optionally try to start a privileged
+     * bridge immediately.  However, it is quite likely that that
+     * needs a password and if we don't have one, it will likely fail.
+     * That would be okay, but sudo is very noisy about failures and
+     * might send nasty emails to your parents.  For that reason we
+     * always pass "init-superuser": "none" here.
```
mvollmer, line 56: This needs improvement: " For that reason we pass "init-superuser": "none" here when there is no password."

mvollmer:
Let's have test run with no allow_authorize_journal_messages at all.  I suspect that we no longer need most of them.  However, in many cases allow_authorize_jounral_messages is there to allow occasional messages from aborted logins due to asynchronous disconnections, and we would potentially introduce a lot of flakes.

I see if I can make some progress anyway in removing calls to allow_authorize_journal_messages anyway.

mvollmer:
> I see if I can make some progress anyway in removing calls to allow_authorize_journal_messages anyway.

I have done some spot tests and could remove a lot of calls.  Let's have the bots look at the rest.

mvollmer:
> Likewise, would be good to have a test case for passwordless sudo, that a single click really keeps working.

We have that, TestSuperuser.testNoPasswd

mvollmer:
> I have done some spot tests and could remove a lot of calls. Let's have the bots look at the rest.

There were no unexpected messages at all.  Let's try a couple of times to try and spot new flakes.

mvollmer:
One almost green run here: https://github.com/cockpit-project/cockpit/pull/15206/commits/6c1845f6cddaf6825d10bd746d072a37b7185c4a

martinpitt:
Big +1 on dropping allow_authorize_messages(). It has become that dumping ground of "my new test gets weird login failure messages, sprinkle enough of these until it shuts up". Thanks!

mvollmer:
Another run without unexpected journal messages: https://github.com/cockpit-project/cockpit/pull/15206/commits/1c1f616852d51e31d70f273758ca112180283c5c

mvollmer:
Another one: https://github.com/cockpit-project/cockpit/pull/15206/commits/56dc2fab69accec37c615f34872fa9e612dcb4e5

I am now confident enough that this doesn't actually re-introduce old flakes.  I'll polish this to be mergeable.

pkg/shell/machines/machines.js
mvollmer: Upps, forgot about this.  Fixed now.

pkg/shell/machines/machines.js
```diff
@@ -53,7 +53,7 @@ export function get_init_superuser_for_options(options) {
      * needs a password and if we don't have one, it will likely fail.
      * That would be okay, but sudo is very noisy about failures and
      * might send nasty emails to your parents.  For that reason we
-     * always pass "init-superuser": "none" here.
+     * pass "init-superuser": "none" here when there is no password.
```
martinpitt: Rebase error, this rather belongs in the previous commit. (But if that's the only thing, not a biggie)

martinpitt (dismissed):
Cheers! Really happy to drop all that "authorize messages" stuff (it was a bit of a `fix_my_test_I_dont_know_what_I_am_doing()`..)

pkg/shell/machines/machines.js
mvollmer: Ouch, let's fix it anyway.  More test runs, more better.

martinpitt (approved):
Cheers!

PR #15208, opened 2021-01-22
tests: some TestMachines flakes fixes

KKoukiou:


KKoukiou:
This needs work.

martinpitt:
@marusak: [check-machines-disks](https://logs.cockpit-project.org/logs/pull-15208-20210127-164037-dd3b8233-ubuntu-2004/log.html) was not retried three times. I think we landed the bots checkout fix more than 14 hours ago. Do you happen to have some idea in your mind already what happened here?

marusak:
> Do you happen to have some idea in your mind already what happened here?

The PR title is misleading, as it does not drop that :) See on line 40:
```
@no_retry_when_changed
class TestMachinesDisks(VirtualMachinesCase):
```

KKoukiou:
> > Do you happen to have some idea in your mind already what happened here?
> 
> The PR title is misleading, as it does not drop that :) See on line 40:
> 
> ```
> @no_retry_when_changed
> class TestMachinesDisks(VirtualMachinesCase):
> ```

Oh, I dropped the commit which removed the decorator by mistake, sorry for the confusion.

KKoukiou:
The no-retry label is not ready to be dropped - but let's move forward with this fixes anyway.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -41,8 +42,11 @@ import './cockpit-components-table.scss';
  * - actions: additional listing-wide actions (displayed next to the list's title)
  * - columns: { title: string, header: boolean, sortable: boolean }[] or string[]
  * - rows: {
- *      columns: (React.Node or string)[],
- *      props: { key: string, className: 'test', ...extraProps: object } - this property is mandatory and should contain a unique `key`, all additional properties are optional
+ *      columns: (React.Node or string or { title: string, key: string, ...extraProps: object}}[]
+                 Through extraProps the consumers can pass arbitary properties to the <td>
+ *      props: { key: string, ...extraProps: object }
```
martinpitt: Was it intended to drop `className` from the documentation again? It was just introduced in the previous commit (pulled out of extraProps), and I don't see a corresponding code change here which would handle its removal?

pkg/machines/components/vm/disks/vmDisksCard.jsx
```diff
@@ -234,7 +234,7 @@ export class VmDisksCard extends React.Component {
         }
 
         const rows = disks.map(disk => {
-            const idPrefixRow = `${idPrefix}-${disk.target || disk.device}`;
+            const idPrefixRow = `${idPrefix}-${(disk.target || disk.device)}`;
```
martinpitt, line 237: Second-guessing my JS fu: This is just purely stylistic, no behaviour change, right?

martinpitt (commented):
Great work, thank you! I'm glad that the flakes found actual bugs, so it wasn't just test robustification for a change. I have two questions, but this looks good by and large. (Let's see whether the bots agree).

pkg/lib/cockpit-components-table.jsx
KKoukiou: Hm, this was just cleaned up from the wrong commit, className is a normal property therefore encapsulated by extraProps, I though there is not need to document it seperately.

I can repost, to remove it in the commit that introduced this change 'className: 'test'' in the first place.

pkg/machines/components/vm/disks/vmDisksCard.jsx
KKoukiou: Indeed, this does not change anything -  sorry for the messy diff. I will revert to how it was if I will repost.

martinpitt (approved):
Cheers!

PR #15210, opened 2021-01-22
unit-tests: copy source into container explicitly

allisonkarlitskaya:
Our current approach to mounting the source directory as a volume inside of the container is problematic for two main reasons:                         
    
 - it doesn't work with selinux, due to incorrect labels 
    
 - it doesn't work with git worktrees, due to the .git file referring to a directory which doesn't get mounted into the container
    
We can fix both of these problems by being more explicit about injecting the source that we want to test into the container.                         
    
Replace our /source volume with a new /pack.tar file.  If it's present, /entrypoint will unpack it into the home directory and run an unpack.sh script from inside of it.  This is our mechanism for getting information into the container.                            
                       
Add a create-pack script on the outside for creating this archive.  It creates a git bundle for both the current HEAD, and for the node_modules in use by that HEAD.  It writes the script for unpacking those to its own stdout.  Each additional file which needs to be included in the .tar is written on a comment line.  Some sed/tar trickery in the start script makes sure everything gets packed up nicely.                                                
                                                                            
Moving the extraction steps to the entrypoint has a couple of nice effects: it means that when using a shell, you end up in an environment with everything already unpacked and ready to build.  It also means that the run.sh script is now "pure": you can run it directly on a source directory, outside of the container, and it will do the same things it does on the inside.
                                                                            
The source ends up in ~/cockpit now, instead of /tmp/source.

### DIFF
--- a/containers/unit-tests/Dockerfile
+++ b/containers/unit-tests/Dockerfile
@@ -13,7 +13,7 @@ WORKDIR /home/builder
 
 ENV LANG=C.UTF-8
 
-VOLUME /source
+VOLUME /pack.tar
 
 COPY entrypoint /
 ENTRYPOINT ["/entrypoint"]
--- a/containers/unit-tests/README.md
+++ b/containers/unit-tests/README.md
@@ -5,9 +5,11 @@ we want to exercise Cockpit with, mostly for `make distcheck` and `make
 check-memory`. This container runs on [GitHub](.github/workflows/unit-tests.yml),
 but can be easily run locally too.
 
-It assumes that the Cockpit source git checkout is available in `/source`. It
-will not modify that directory or take uncommitted changes into account, but it
-will re-use an already existing `node_modules/` directory.
+On startup, it will look for `/pack.tar`, extract it into the home directory,
+and run the contained unpack.sh.
+
+We use this mechanism to inject what we want to test.  See the create-pack
+script in this directory for details on how that works.
 
 The scripts can use either podman (preferred) or docker. If you use docker, you
 need to run all commands as root. With podman the containers work as either user
@@ -20,11 +22,6 @@ The `build` script will build the `cockpit/unit-tests` and
 build one variant, e.g. `build i386`.
 
 ## Running tests
-
-You need to disable SELinux with `sudo setenforce 0` for this. There is no
-other way for the container to access the files in your build tree (do *not*
-use the `--volume` `:Z` option, as that will destroy the file labels on the
-host).
 
 Tests in that container get started with the `start` script.  By default, this
 script runs the unit tests on amd64.  The script accepts a number of arguments
@@ -56,14 +53,11 @@ For interactive debugging, run a shell in the container:
 
     $ ./start shell     # start an interactive shell in default container
 
-You will find the cockpit source tree (from the host) mounted at `/source` in
-the container.
+You will find the cockpit source tree (from the host) unpacked in the home
+directory (~/cockpit) inside the container.
 
-`/source/containers/unit-tests/run.sh` will start the builds and test run, then
-you can investigate in the build tree at `/tmp/source/`.
-
-`/source/containers/unit-tests/run.sh` also includes a `build` mode which
-will checkout and build the source, but not run any tests.
+`containers/unit-tests/run.sh` will start the builds and test run, then
+you can investigate in the build tree.
 
 You can also attach to another container using the provided `exec` script.  For example:
 
--- a/containers/unit-tests/create-pack
+++ b/containers/unit-tests/create-pack
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+set -eu
+cd "$(realpath -m "$0"/../../..)"
+
+message() { printf "  %-8s %s\n" "$1" "$2" >&2; }
+
+# bundle the current state of HEAD
+(
+    head_sha="$(git rev-parse HEAD)"
+    head_pack="tmp/${head_sha}.pack"
+    if [ ! -f "${head_pack}" ]; then
+        message 'BUNDLE' "${head_pack}  [HEAD]"
+        # include the latest tag so that git-describe works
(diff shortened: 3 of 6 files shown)


martinpitt:
Agreed that it is rather gross to copy it twice, and moreover, having to copy the rather bulky `node_modules/`; can't we at least reuse that and let that be relabeled? It now goes away on `git clean` anyway, so a changed label shouldn't hurt?

One of these days we need to fix our build system to build everything in a separate build tree. `make-source` seems rather close to that already. Anyway, I can't say I like this much, but I already don't like the current setup, so not much of a difference :grin: This should at least update the documentation, though.

allisonkarlitskaya:
> Agreed that it is rather gross to copy it twice, and moreover, having to copy the rather bulky `node_modules/`; can't we at least reuse that and let that be relabeled? It now goes away on `git clean` anyway, so a changed label shouldn't hurt?

I'm going to try again on my idea of creating the worktree outside and having the inside work directly on it (ie: no further copying), which will reduce one of the copies.

Additionally: at least on btrfs, `cp` defaults to `--reflink=auto` and is therefore really fast and doesn't use very much space...

allisonkarlitskaya:
So I remembered what the problem was before: we don't build as root, and the directory on the outside is almost definitely not going to end up with the correct uid for the `builder` user on the inside...

allisonkarlitskaya:
The force-push just now fixes both of the original problems that this PR was opened for, and also represents a very large cleanup.

It also seems to be working correctly.

Generally, while developing this, it also started to feel like something that I might actually want to use locally.

martinpitt:
The first 8 commits are obvious no-brainers, and could land right away (in a separate PR).

I don't like the others, honestly, at least not from a quick pass-by review: This is a *lot* of non-straightforward shell code, copying, and hacking for something which should be done properly with volumes -- either `:ro` and a separate build tree, or with `:z` for writable directories and running the unit test container as root so that it uses the correct UID for the host user.

Copying is not only slow, it's also annoying because it breaks the "edit source on your host and re-run make in running container" development iteration use case.

So for these I'll need a lot more convincing -- let's talk in January?

allisonkarlitskaya:
> The first 8 commits are obvious no-brainers, and could land right away (in a separate PR).

→ #16749

allisonkarlitskaya:
> Copying is not only slow, it's also annoying because it breaks the "edit source on your host and re-run make in running container" development iteration use case.

But that doesn't work anyway: the existing setup already takes a clone of the `/source` repository into `/tmp/source` and builds from there...

This is more about replacing one form of copying with a different one...

martinpitt:
Notes for myself: I had a look at this again, on a fully standard F35 machine:
```sh
mkdir src; echo world > src/hello.txt
ls -lZ src/
# -rw-r--r--. 1 admin admin unconfined_u:object_r:user_home_t:s0 6 Jan  4 11:53 hello.txt

podman run -it --rm -v ./src:/src:ro quay.io/libpod/busybox sh -c 'ls -lr /src; grep -r . /src'
# ls: can't open '/src': Permission denied
# grep: /src: Permission denied
```

The same happens with `sudo`, so this is not a problem specific to user containers, and not even related to the UID inside the container -- the latter might impose an additional problem with regular Unix UIDs, but that one would be easy enough to work around -- just run the build inside the container as root (then it's equal to your user on the host).

So indeed you either need to switch off SELinux, or live with the changing container labels:
```
podman run -it --rm -v ./src:/src:ro,z quay.io/libpod/busybox sh -c 'ls -lr /src; grep -r . /src'
# -rw-r--r--    1 root     root             6 Jan  4 11:53 hello.txt
# /src/hello.txt:world

ls -lZ src/
# -rw-r--r--. 1 admin admin system_u:object_r:container_file_t:s0 6 Jan  4 11:53 hello.txt
```

I'm not sure how much of a problem that actually is, especially in a git tree which is not particularly precious. Unfortunately in https://github.com/cockpit-project/cockpit/commit/af1053ca00e2faa1b1dec30e8302108086acd43d I didn't say *why* it was a problem. :cry: 

This is not a bug in the sense that it works as designed -- SELinux forbids "exporting" files to containers unless they are typed `container_file_t`. As to whether this policy makes *sense* we can have a long discussion, but the current status quo won't change much.

So, in summary, we have three options: `:z`, `setenforce 0`, or copying the tree into the container via tar, or a temporary directory (which then can use `:z` just fine) -- this PR essentially does the latter.

martinpitt (changes requested):
During the holidays I just had a cursory glance at this, now a deeper look. I do agree about the two problems (SELinux and .git work trees), but I still find create-pack unacceptably and wildly over-engineered.

What we essentially need to do is to move the `git clone /source` out of the container into the host, so that we don't increase the amount of work and complexity.

This should essentially be:

    cp -a . $tmpdir
    podman run ... -v $tmpdir:/source:z

and then doing the build right in /source. This will work just fine if the container runs as root, but fail if it runs as user. I don't see a good reason to actually run it as non-root, do you?

This needs to be optimized a little: For sure the above `cp -a` is a little too naïve, as it will also copy any existing test/images/. That should certainly be skipped from the copying, as that takes long and is unnecessary. We want pretty much everything else from the build tree (node_modules, dist, autoconf files, etc.) -- if code is already built, there is little reason to not use them (this won't matter in CI, but is nice for local iteration).

But I propose to do this in smaller steps: For the first version, just move out the `git clone` and `node_modules` copying from the container to the host, and switch the unit-test container to root. That is a straight code move without any increasing complexity or performance penalty, and will fix the original problems (SELinux, worktrees).

Another commit can then add dist/ to the thing, which is simple enough.

For the pre-built C stuff and local changes, we can re-engineer this the other way around: not start from git clone, but from copying the whole tree and filter out test/common.

All of this should be < +10 lines and be obvious and understandable.

allisonkarlitskaya:
Your suggestions here are basically what I did in my first approach to solving this problem.  I'll come back to this later.

allisonkarlitskaya:
We landed #17304 instead.

PR #15184, opened 2021-01-21
package.json: Update @patternfly/react-styles package dependency

cockpituous:

PR #15177, opened 2021-01-20
package.json: Update @patternfly/react-console package dependency

cockpituous:

PR #15211, opened 2021-01-23
Update patternfly

marusak:
Fixes #15184
Fixes #15177


marusak:
I checked visually and haven't noticed any regression.

martinpitt (approved):
Thanks! Indeed the novnc dependency was added as a temporary workaround, it can go now.

PR #15214, opened 2021-01-25
shell: Never connect to remote machines automatically

mvollmer:
Release note:
----
### Shell: SSH connections to remote machines are only opened when necessary

Cockpit remembers all remote hosts of a session and used to immediately connect to all of them when starting the next session.  Now a SSH connection is only opened when navigating to the remote machine for the first time.

### DIFF
--- a/pkg/shell/base_index.js
+++ b/pkg/shell/base_index.js
@@ -793,7 +793,7 @@ function Index() {
             self.navigate(ev.state, true);
         });
 
-        self.navigate();
+        self.navigate(null, true);
         cockpit.translate();
         $("body").prop("hidden", false);
     };
--- a/test/verify/check-host-switching
+++ b/test/verify/check-host-switching
@@ -18,6 +18,7 @@
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
 import json
+import time
 
 import parent
 from testlib import *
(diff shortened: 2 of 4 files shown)


mvollmer:
The initial state of a remote machine looks like this in this draft:

image

Let's improve that by not making it look like an error.

mvollmer:
> Let's improve that by not making it look like an error.

This is no longer necessary.  Now we connect when navigating, so people will not see this curtain.

pkg/shell/base_index.js
```diff
@@ -790,10 +790,10 @@ function Index() {
 
     self.ready = function () {
         $(window).on("popstate", function(ev) {
-            self.navigate(ev.state, true);
+            self.navigate(ev.state);
```
mvollmer, line 793: This change is not necessary, I'd say.

mvollmer:
GREEEEEN!

test/verify/check-host-switching
```diff
@@ -389,6 +395,23 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
...
+        # And and connect to a second machine
+        b.click("#hosts-sel button")
+        self.add_machine(b, "10.111.113.2")
+        b.click("a[href='/@10.111.113.2']")
+        b.wait_visible("iframe.container-frame[name='cockpit1:10.111.113.2/system']")
+        self.assertIn("admin", m2.execute("loginctl"))
+        b.click("#hosts-sel button")
+        b.click("a[href='/@localhost']")
+        b.relogin()
+        time.sleep(60)
```
martinpitt, line 413: That's certainly a long time, but I see that we want to make really sure. :grin: But the test is destructive, so it doesn't  hold up the nd ones.

martinpitt (approved):
Thank you, amazing work! This was certainly not easy to figure out, but it's surprisingly clear to review. I had expected more noise.

PR #15218, opened 2021-01-26
test: Restart packagekit service after we do cleanup

marusak:
As needed for https://github.com/cockpit-project/bots/pull/1574

### DIFF
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -371,7 +371,7 @@ class TestUpdates(NoSubManCase):
         assertHistory("table.updates-history ul", ["secdeclare", "secparse", "sevcritical"])
 
         # stop PackageKit (e. g. idle timeout) to make sure the page survives that
-        m.execute("systemctl kill --signal=KILL packagekit.service")
+        m.execute("systemctl stop packagekit; systemctl reset-failed packagekit || true")
 
         # new security versions are now installed
         m.execute("test -f /stamp-secdeclare-3-4.b1 && test -f /stamp-secparse-4-2 && test -f /stamp-sevcritical-5-2")
--- a/test/verify/packagelib.py
+++ b/test/verify/packagelib.py
@@ -68,7 +68,9 @@ class PackageCase(MachineCase):
             self.machine.execute("rm -rf /etc/yum.repos.d/* /var/cache/yum/* /var/cache/dnf/*")
 
         # have PackageKit start from a clean slate
+        self.machine.execute("systemctl stop packagekit")
         self.machine.execute("systemctl kill --signal=SIGKILL packagekit; rm -rf /var/cache/PackageKit")
+        self.machine.execute("systemctl reset-failed packagekit || true")
         self.restore_file("/var/lib/PackageKit/transactions.db")
 
         if self.image in ["debian-stable", "debian-testing"]:


martinpitt:
Oh! That's a surprise. I thought it was the *killing* of packagekit that somehow messed up dbus-broker's brain. packagekit is supposed to be D-Bus activated, so it's really wrong to manually start it. Could it be that it perhaps goes into "failed" state, and it should rather do "systemctl stop && systemctl  reset-failed" after killing it?

marusak:
>  thought it was the killing of packagekit that somehow messed up dbus-broker's brain.

`systemctl kill --signal=SIGKILL packagekit` this is what messes up dbus-broker. The first request fails, we need to do two request, the second one and any other following will work, just the first request after killing ^ is not working.

> Could it be that it perhaps goes into "failed" state, and it should rather do "systemctl stop && systemctl reset-failed" after killing it?

It is in 'running' state after you kill it. Maybe I can just 'stop' it to tell dbus-broker it has changed?

marusak:
hmm, doing just 'stop' breaks `testPrivilegeChange` where it succeeds only on the second try. This but may be actuall issue on our side.
```
loading available updates failed: {"detail":"Failed to obtain authentication.","code":48}
```

martinpitt:
> It is in 'running' state after you kill it.

Ugh, *that* is a surprise indeed. Does it not get properly killed, or does it restart immediately? The latter would make sense, and then we actually need to `stop` it first, possibly with `--no-block` to avoid the 90s timeout. I.e. `stop; kill, reset-failed`, to be really robust?

marusak:
So I tried stopping it, before killing the process. It works in all cases but when we stop it and then log in as user with limited access it does not work and in logs for the service I see:
```
Jan 26 11:47:06 localhost.localdomain PackageKit[3318]: uid 1000 is trying to obtain org.freedesktop.packagekit.system-sources-refresh auth (only_trusted:0)
Jan 26 11:47:06 localhost.localdomain PackageKit[3318]: uid 1000 failed to obtain auth
Jan 26 11:47:37 localhost.localdomain PackageKit[3318]: uid 1000 is trying to obtain org.freedesktop.packagekit.system-sources-refresh auth (only_trusted:0)
Jan 26 11:47:37 localhost.localdomain PackageKit[3318]: uid 1000 failed to obtain auth
```
In this moment the unit is running and it was started by non authorized user trying to access updates page in cockpit.
But this must be a fallback from cleanup, as when I run this test for the first time, it works just fine. Also restarting seems fine. Just when it runs after some other packagekit test it has issues.

martinpitt:
Some tests now fail, presumably because previously all the tests assumed that packagekit wasn't running initially?

But either way, having to start it manually is just wrong. This really smells like some weird regression in dbus-broker. Let's look at a manual reproducer tomorrow morning that does not involve cockpit -- it seems you can set up this scenario with a couple of `pkcon` and `systemctl` commands?

marusak:
Aha, so the reason why `testPrivilegeChange` fails is because `pkcon refresh` is failing:
```
Transaction:	Refreshing cache
The daemon crashed mid-transaction!
```

But it does not happen in CI. I think this is very much so dependand on the test order.
`TEST_OS=fedora-testing ./test/verify/check-packagekit TestUpdates.testBasic TestAutoUpdates.testPrivilegeChange` fails, if I turn the order around it works.

In CI `TestUpdates.testInfoSecurity` is failing.

marusak:
> In CI `TestUpdates.testInfoSecurity` is failing.

Easy to reproduce locally. We kill packagekit and then try to keep using the page. Before it worked, now it fails with not finding the daemon. Refreshing page works. Tried just stopping the service instead of killing it, but that behaves the same.

test/verify/check-packagekit
```diff
@@ -371,7 +371,7 @@ class TestUpdates(NoSubManCase):
         assertHistory("table.updates-history ul", ["secdeclare", "secparse", "sevcritical"])
 
         # stop PackageKit (e. g. idle timeout) to make sure the page survives that
-        m.execute("systemctl kill --signal=KILL packagekit.service")
+        m.execute("systemctl stop packagekit && systemctl reset-failed packagekit")
```
martinpitt, line 374: This looks a little curious -- the `reset-failed` will *not* run if  `stop` fails, due to the shell's rather strange "false && true" behaviour (the compound command does not fail under `set -e`, but due to short-circuiting the `true` does not run)

marusak: Good point, fixed. The restart is actually needed at least for the new f-testing.

martinpitt (approved):
Dakujem!

dvdhrm:
This was actually a regression in upstream `dbus-broker`. This should be fixed with `dbus-broker-27-2` in Fedora. It already hit rawhide&F34, updates for F33 and F32 are pending.

Feel free to put us on CC for any issues you see with D-Bus on `dbus-broker`-enabled machines. Such behavioral changes are definitely not intentional.

martinpitt:
Nice, thanks @dvdhrm! I wasn't sure if that constituted as an actual bug, or just some random undefined behaviour change. I see 27-2 is in 33/34 stable already, so after our next image refresh we can try and revert this.

PR #15223, opened 2021-01-27
tools: Skip slow tests on hppa in Debian

martinpitt:
hppa needs almost an hour to run the test-tls-certfile unit test. Skip
slow tests on this architecture to avoid a build timeout.

https://bugs.debian.org/981127

### DIFF
--- a/tools/debian/rules
+++ b/tools/debian/rules
@@ -5,8 +5,10 @@ DEB_HOST_MULTIARCH ?= $(shell dpkg-architecture -qDEB_HOST_MULTIARCH)
 # only ship deprecated PatternFly API for stable releases (backports); Debian sid/testing has no VERSION_ID
 SHIP_PF_API = $(filter $(shell . /etc/os-release; echo $${VERSION_ID:-unstable}),10 20.04)
 
-# this is an emulated architecture for now, and too slow to run expensive unit tests
-ifeq ($(shell dpkg-architecture -qDEB_BUILD_ARCH),riscv64)
+# riscv is an emulated architecture for now, and too slow to run expensive unit tests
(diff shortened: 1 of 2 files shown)


tools/debian/rules
```diff
@@ -5,8 +5,10 @@ DEB_HOST_MULTIARCH ?= $(shell dpkg-architecture -qDEB_HOST_MULTIARCH)
 # only ship deprecated PatternFly API for stable releases (backports); Debian sid/testing has no VERSION_ID
 SHIP_PF_API = $(filter $(shell . /etc/os-release; echo $${VERSION_ID:-unstable}),10 20.04)
 
-# this is an emulated architecture for now, and too slow to run expensive unit tests
-ifeq ($(shell dpkg-architecture -qDEB_BUILD_ARCH),riscv64)
+# riscv is an emulated architecture for now, and too slow to run expensive unit tests
+# hppa's threading is absurdly slow (#981127)
+SLOW_ARCHES = $(filter $(shell dpkg-architecture -qDEB_BUILD_ARCH),riscv64 hppa)
```
allisonkarlitskaya, line 10: *squint*

Fine with this change in theory, but I wonder if you could find a more readable way to express it...

I guess we don't have any kind of `in` verb here unfortunately :/

allisonkarlitskaya (approved):

tools/debian/rules
martinpitt: Heh, yeah.. I modelled it after the other `ifneq()` bits we have (and previously had) in rules. Makefile syntax is so arcane that I usually resort to copy&paste of proven snippets than trying to invent something new :-/

PR #15225, opened 2021-01-27
web_socket_util_parse_headers: Move Insert after check

mehmooda:
It is possible for g_hash_table_insert to invalidate the strings. Move the ownership transfer to after the check to ensure we don't touch the name and value again after insert.

### DIFF
--- a/src/websocket/websocket.c
+++ b/src/websocket/websocket.c
@@ -324,8 +324,6 @@ web_socket_util_parse_headers (const gchar *data,
   gsize consumed = 0;
   gboolean end = FALSE;
   gsize line_len;
-  gchar *name;
-  gchar *value;
 
   parsed_headers = web_socket_util_new_headers ();
 
@@ -360,9 +358,9 @@ web_socket_util_parse_headers (const gchar *data,
               break;
             }
 
-          name = g_strndup (data, colon - data);
+          g_autofree gchar *name = g_strndup (data, colon - data);
           g_strstrip (name);
-          value = g_strndup (colon + 1, line - (colon + 1));
+          g_autofree gchar *value = g_strndup (colon + 1, line - (colon + 1));
           g_strstrip (value);
 
           if (!is_valid_line (name, -1) || !g_utf8_validate (value, -1, NULL))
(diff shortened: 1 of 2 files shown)


src/websocket/websocket.c
```diff
@@ -364,14 +364,14 @@ web_socket_util_parse_headers (const gchar *data,
           g_strstrip (name);
           value = g_strndup (colon + 1, line - (colon + 1));
           g_strstrip (value);
-          g_hash_table_insert (parsed_headers, name, value);
 
           if (!is_valid_line (name, -1) || !g_utf8_validate (value, -1, NULL))
             {
               g_debug ("received invalid header");
               consumed = -1;
               break;
```
allisonkarlitskaya, line 372: Exiting the loop without inserting the freshly-allocated `name` and `value` into the table is a leak.

A better approach is probably to move the declarations of the `name` and `value`  inside the loop, use `g_autofree` on them, and use `g_steal_pointer()` at the point of inserting them into the hash table.

allisonkarlitskaya (changes requested):

src/websocket/websocket.c
allisonkarlitskaya: Specifically: move the variable declaration to the site of the assignment.

mehmooda: I hope that I've done this correctly. Sorry it's my first time using g_autofree and g_steal_pointer

src/websocket/websocket.c
```diff
@@ -360,18 +358,18 @@ web_socket_util_parse_headers (const gchar *data,
...
             }
+          g_hash_table_insert (parsed_headers, g_steal_pointer(&name), g_steal_pointer(&value));
```
allisonkarlitskaya, line 372: This looks exactly correct, but a minor whitespace nag here.

```suggestion
          g_hash_table_insert (parsed_headers, g_steal_pointer (&name), g_steal_pointer (&value));
```

allisonkarlitskaya (commented):
I'd be happy to merge this by squashing into one commit and rephrasing the commit message if the bots agree.

mehmooda:
@allisonkarlitskaya Ready when you are.

allisonkarlitskaya (approved):
This is ready to land (with squashing).

Thanks for the test!

PR #15226, opened 2021-01-27
tools: Don't treat webpack warnings as errors

AdamMajer:
In some versions of webpack, irrelevant warnings are resulting
in failure to build from sources


martinpitt (approved):
Thanks, this makes sense. I'm triggering one integration test to make double-sure.

PR #15227, opened 2021-01-27
storage: Trigger udev after adding disks to a MDRaid device

mvollmer:
This should be triggered by mdraid itself, but doesn't seem to happen
reliably.  This is not a surprise.


martinpitt:
Yay! Two test run failures, retrying in our new spirit on "merge green PRs". There are also some [retried flakes](https://logs.cockpit-project.org/logs/pull-15227-20210127-141610-11d47a1e-ubuntu-stable/log.html#241), but that looks like something else?

martinpitt (approved):

PR #15228, opened 2021-01-27
 lib: follow the pattern for empty lists as seen on PF4 Table demos

KKoukiou:
Now it will look like:

Machines page:

Screen Shot 2021-01-27 at 14 39 23

Storage page:

Screen Shot 2021-01-27 at 15 46 54


martinpitt (approved):
Thank you! Nice cleanup, and great catch with the firefox vs. chrome special case!

PR #15229, opened 2021-01-27
services: use PF4 select for choosing unit type

### LINKED ISSUE #15196: services: Accessibility isssues
- [x] #services-dropdown is a Cockpit widget (not a PF4 one) and has an issue where it does not have an accessible name. That is, it's lacking a label. This can be fixed by adding a `<label>`, aria-label, or aria-labelledby (pointing to an ID). I'd suggest calling it "State" and consider using [the screen reader only class (`pf-u-screen-reader`)](https://www.patternfly.org/v4/utilities/accessibility#screen-reader-only) on a label.
- [ ] Cockpit logs have inappropriate usage of role="row" (see #15191, where this might be fixed instead)
- [ ] Description list has immediately children not of type `<dl>` or `<dt>`
- [x] List items are not used properly. `<li>` need to be children of a list item, such as  `<ul>` (or `<ol>`, which doesn't really apply here0. They cannot be free floating.

KKoukiou:
Fixes #15196

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -464,9 +464,11 @@ export class ServiceDetails extends React.Component {
                                 {relationships.map(rel =>
                                     rel.Units && rel.Units.length > 0 &&
                                         <DescriptionListGroup key={rel.Name.split().join("")}>
-                                            <DescriptionListTerm className="closer-lines">{rel.Name}</DescriptionListTerm>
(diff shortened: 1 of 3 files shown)


martinpitt:
Fails [TestServices.testBasic](https://logs.cockpit-project.org/logs/pull-15229-20210127-173030-38850013-fedora-32/log.html), moving to draft for now.

martinpitt:
Code change looks good, though! :+1:

martinpitt (approved):
Hooray!

PR #15230, opened 2021-01-27
tests: make testDiskEdit more stable by checking radio buttons value …

KKoukiou:
…before closing the modal

Radio buttons don't seem to expose a 'checked' attribute, therefore this
commits add the selected radio button value as an attribute of the
FormGroup.

Cherry-picked from https://github.com/cockpit-project/cockpit/pull/15208, if it really works in can be merged independently.


KKoukiou:
Nope did not help

PR #15232, opened 2021-01-28
test: Enable screenshots for Firefox

martinpitt:
The current nightlies now support the `Page.captureScreenshot()` API.

PR #15233, opened 2021-01-28
DONTMERGE test: Enable screenshots for Firefox + break a test

martinpitt:
PR #15232 now fixes screenshots for firefox. As our tests are now so robust :rofl: and it didn't happen to produce a flake yet, this is a quick known-broken PR to validate that everything works on the infra.


martinpitt:
Nevermind, https://logs.cockpit-project.org/logs/pull-15232-20210128-082558-a1779496-fedora-32-firefox/log.html does have some retries now.

PR #13694, opened 2020-03-10
build: force `npm install` after package.json changes

allisonkarlitskaya:
It's a reasonably common problem to pull, switch branches, or rebase on
top of new changes from master, and have changes to node module
dependencies which will cause the build to fail if `npm install` isn't
run again.  Even worse: sometimes the build succeeds, only to cause
runtime javascript errors during testing.

This is almost always indicated by a corresponding change in
package.json.

Make sure package-lock.json is newer than package.json, and if not,
display a warning about it, asking the user to fix it.  Use a ridiculous
amount of whitespace to make sure the message sticks out from the noise
and to make it easier to copy/paste the suggested command.

Also: add a new Makefile target `make npm` to explicitly touch
package-lock.json after a sucessful `npm install` (since npm won't do it
for itself if there are no changes).

PR #15235, opened 2021-01-28
build: add jump-start option to build without node

allisonkarlitskaya:
Add a pair of new makefile rules to support "jump-start" builds.

First, `make cockpit-jumpstart.tar.xz` can be used to create an archive
of the files needed to build the current version of cockpit without
requiring any of npm, node, or webpack.

Second, `make jumpstart` will use a `cockpit-jumpstart.tar.xz` found in
the current directory.

After a "jump-start", it's possible to build the C parts of cockpit and
run the full suite of unit tests, and prepare images for testing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -250,6 +250,31 @@ uninstall-local::
 dist-hook:: $(WEBPACK_INPUTS) $(WEBPACK_OUTPUTS)
 	$(V_TAR) tar -cf - $^ | tar -C $(distdir) -xf -
 
+# this makes it possible to `make check`
+JUMPSTART_EXTRAS = \
+	node_modules/chrome-remote-interface \
+	node_modules/ws \
+	$(NULL)
+
+# We extract with `tar -m` so the order of files equals the order of
+# timestamps.  Make sure the stamp files stay at the end so that they
+# are the "newest".
+JUMPSTART_CONTENTS = \
+	$(JUMPSTART_EXTRAS) \
+	$(WEBPACK_INPUTS) \
+	$(WEBPACK_OUTPUTS) \
+	$(WEBPACK_DEPS) \
+	$(WEBPACK_STAMPS)
+
+cockpit-jumpstart.tar: $(JUMPSTART_CONTENTS)
+	tar cf $@ $^
+
+cockpit-jumpstart.tar.xz: cockpit-jumpstart.tar
+	xz --extreme < $< > $@
+
+jumpstart:
+	tar xmf cockpit-jumpstart.tar.xz
+
 #
 
 ACLOCAL_AMFLAGS = -I tools ${ACLOCAL_FLAGS}
--- a/Makefile.am
+++ b/Makefile.am
@@ -66,6 +66,10 @@ V_GZIP = $(V_GZIP_$(V))
 V_GZIP_ = $(V_GZIP_$(AM_DEFAULT_VERBOSITY))
 V_GZIP_0 = @echo "  GZIP    " $@;
 
+V_XZ = $(V_XZ_$(V))
+V_XZ_ = $(V_XZ_$(AM_DEFAULT_VERBOSITY))
+V_XZ_0 = @echo "  XZ      " $@;
+
 V_COPY = $(V_COPY_$(V))
 V_COPY_ = $(V_COPY_$(AM_DEFAULT_VERBOSITY))
 V_COPY_0 = @echo "  COPY    " $@;
@@ -113,6 +117,14 @@ CAT_RULE = \
 GZ_RULE = \
 	$(V_GZIP) $(MKDIR_P) $(dir $@) && \
 	gzip -n -c $< > $@.tmp && $(MV) $@.tmp $@
+
+XZ_RULE = \
+	$(V_XZ) $(MKDIR_P) $(dir $@) && \
+	xz $(XZ_COMPRESS_FLAGS) -c $< > $@.tmp && $(MV) $@.tmp $@
+
+TAR_RULE = \
+	$(V_TAR) $(MKDIR_P) $(dir $@) && \
+	tar c $^ > $@.tmp && $(MV) $@.tmp $@
 
 SUBST_RULE = \
 	$(AM_V_GEN) $(MKDIR_P) $(dir $@) && sed \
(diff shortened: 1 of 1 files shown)


martinpitt:
Summary from IRC: I'd rather use the regular `make dist` tarballs. They are like 95% of what we need here already, and *they* are the product that we ship to all the distros. So IMHO it would make sense to have a workflow that builds a `make dist`, stores it as artifact, and test/image-prepare would then download that and start from there instead of re-building `make dist` from scratch. That would also be really convenient for us developers.

It would also simplify our downstream tests, which currently [have to install three extra NPM modules](https://src.fedoraproject.org/rpms/cockpit/blob/master/f/tests/run-test.sh#_26). Let's just include them into the release tarball to avoid the npm faff downstream, and ship with releases which have been validated. (I'll send a PR tomorrow).

WDYT?

allisonkarlitskaya:
I think if we do this on every single push/merge to master then I'm happy.  Personally, I've been using this setup all day here and I like it a lot.  I don't want to have to do a full `make dist` build to get what I need...

There's also the matter of filtering out the relevant parts of the `dist` in order to make sure we only get the needed parts (essentially, everything that's in the jumpstart file, and nothing else), at least at extraction time.  Finally, we need to make sure timestamps are treated in a sensible way.  It took quite some playing around today before I got that right.

fwiw, I was able to get `make check` working with only `chrome-remote-interface` and `ws` node modules.

allisonkarlitskaya:
Some comparisons:

`cockpit-jumpstart.tar.xz` is about 11MB and takes about 30s to build (from a built tree).

`cockpit-*.tar.xz` is 15M, `cockpit-cache-*.tar.xz` is 70M, together 85M.  They take in the ballpark of 7-8 minutes to build here, under the same conditions.

```
time (git clean -xdf && NO_NPM=1 ./autogen.sh && tar xm --strip-components=1 -f ../main/cockpit-cache-236.46.gcfde3190c.tar.xz cockpit-236.46.gcfde3190c/node_modules && tar xm --strip-components=1 -f ../main/cockpit-236.46.gcfde3190c.tar.xz cockpit-236.46.gcfde3190c/dist && make -j10)
``` works, and takes approximately the same amount of time as a jumpstart (~8 seconds more due to having to unpack larger archives).

allisonkarlitskaya:
After thinking about it a bit more, neither approach is really exactly what we want.  I ship the `WEBPACK_INPUTS` in my jumpstart tarball in order to keep the dependencies in the webpack `Makefile.deps` files happy.  I include those, in turn, because I want tests to work...

Shipping fragments of the `node_modules` directory is actually pretty pointless though.  It's not like we're gonna be able to build webpack stuff with only the files we have there ­— quite the opposite, it makes a later `npm install` more difficult.  We could probably skip that entirely and just touch the files that are needed, or even better, hack the Makefile.deps that we ship to not depend on them.  We just need to really hope nobody ever tries to make an actual tarball for actual distribution from this twisted world that we'll have then created...

martinpitt:
> I think if we do this on every single push/merge to master then I'm happy. 

Yes, that was the idea -- on `push` to master, and on `pull_request` for PRs, so that for everything we need to do there should be a tarball available.

> filtering out the relevant parts of the dist

How do you mean? What part would be unnecessary? If we drop any file there at unpack time, wouldn't the makefiles try to build them again?

> timestamps are treated in a sensible way

Right, that's going to be a bit more tricky. My gut feeling is that just unpacking the tarball with setting a current timestamp (instead of restoring the one in the tarball) should be fine?

> only chrome-remote-interface and ws

(You mean "sizzle", not "ws")

These do suffice for cockpit's unit tests and all our other projects (starter-kit etc.), but some of our cockpit integration tests use axe-core. However, the latter has been an unfinished lingering feature for ages, and is quite big (4.6 MB), so I'd rather look into changing our tests to gracefully handle an absent axe-core. We don't need this downstream, it's only a development feature.

I wouldn't build and archive cockpit-cache* for these runs -- IMHO it's fine to do that *only* at release time. This tarball has no other purpose than a legal one (make sure the source is available if/when npmjs.org is down?), and frankly I'd be fine to just drop it entirely. With only cockpit-N.tar.xz the build time and size is not that much bigger than with jumpstart, and the advantages (being simpler, faster on the whole, we make sure that it's complete, and we test what we actually ship) greatly outweigh, IMHO.

> Shipping fragments of the node_modules directory is actually pretty pointless though. It's not like we're gonna be able to build webpack stuff with only the files we have there ­— quite the opposite

Correct, you can't build from that. It's again just a legal thing (I figure -- this predated me) for providing the source of what went into the dist/ binary blob. But honestly we could just point to cockpit-cache for that. If we changed the disted node_modules/ to *only* chrome-remote-interface and sizzle, then we should be in the "jumpstart" ballpack (size and time wise), and would have something sensible.

martinpitt:
> I wouldn't build and archive cockpit-cache* for these runs -- IMHO it's fine to do that only at release time.

It seems that this is already the case? image-prepare uses [make-source](https://github.com/cockpit-project/bots/blob/master/make-source), which prefers `make dist-gzip`, which doesn't build -cache. So it seems we don't need to fiddle with that for now?

martinpitt:
I sent PR #15242 to do some bits of what was discussed here.

martinpitt:
> (You mean "sizzle", not "ws")

Actually, you mean both, sorry -- it's a peer dep of c-r-i. I'll update #15242 accordingly.

martinpitt:
Summarizing our lengthy IRC discussion. What we want is:
 1. `make` rebuilds affected webpacks on changes to pkg/ (which may *also* be pkg/lib/, so not *only* `pkg/<pack>/`
 2.  dist tarballs could ship with empty node_modules/ and `make` would still not try to build webpacks (in practice we *do* want to ship some node_modules that we need for running tests, like qunit, sizzle, and chrome-remote-interface -- but these are not related to `make`)
 3. everything gets rebuilt, or at least a message "clean up and start over" if `package-lock.json` gets updated (through `git pull` usually) -- see PR #13694 for an attempt

With these, we can provide:

 - fast and convenient developer testing (image-prepare could download pre-built dist tarballs from github artifacts)
 - lean dist tarballs
 - more robust and faster releases (don't rebuild all webpacks and ship an untested artifact, but download the prebuilt one from master that got tagged)

It's not yet fully clear exactly how to implement 1 and 2, as make is not powerful enough to express something like `build_target: any git-tracked file in pkg/pack/ and pkg/lib`. So we may still use webpack-make to generate Makefile.deps, but *remove* the depending files in node_modules/. Perhaps there is a more elegant and more intrusive better approach for this, but we could start with that.

For 3 we need to look at PR #13694 again, as that becomes a little more urgent. However, a strawpoll just showed that this whole "track dependencies in node_modules/" does not reliably work anyway. Everybody is accustomed to `git clean -fdx` and starting over after npm changes.

martinpitt:
PR #15313 implements the jump-start workflow for image-prepare and local integration test running fully. It also unpacks dist and node_modules, but that is still wonky and needs some more work (and teaching npm install to not mess up).

allisonkarlitskaya:
Dealing with this in #15900 now.

PR #15236, opened 2021-01-28
test: Retry deletion in StorageTestVDO

mvollmer:
The device is often still busy.

### DIFF
--- a/test/verify/check-storage-vdo
+++ b/test/verify/check-storage-vdo
@@ -110,7 +110,8 @@ class TestStorageVDO(StorageCase):
         # Delete
 
         b.click('.pf-c-card__header:contains("VDO") button:contains("Delete")')
-        self.confirm()
+        self.dialog_wait_open()
+        self.dialog_apply_with_retry(expected_errors=["Device or resource busy"])
         b.wait_visible("#storage")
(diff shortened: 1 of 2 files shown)


test/verify/check-storage-vdo
```diff
@@ -110,7 +110,7 @@ class TestStorageVDO(StorageCase):
         # Delete
 
         b.click('.pf-c-card__header:contains("VDO") button:contains("Delete")')
-        self.confirm()
+        self.confirm_with_retry()  # device might still be busy, retry a bit
```
KKoukiou, line 113: Instead of blindly retrying we could also maybe check if some process is keeping the device busy.
So:
wait("lambda: self.machine.execute("lsof | grep -v /whatever"), delay=3)
At least we will know if it's busy and timeouts, that it failed in this command. What do you think?

mvollmer: We could check the exact error message in the dialog and have a list of expected error messages.

KKoukiou: Sure that also makes sense!

mvollmer: Done!

KKoukiou (approved):
nice, the retry with the expected errors makes total sense.

PR #15237, opened 2021-01-28
Drop custom react components - Switch and Select - let's just use the ones provided but PF-react-core

KKoukiou:


martinpitt (approved):
Amazing work, thank you! Let's get this in quickly, as it's prone to get conflict-y fast.

PR #15239, opened 2021-01-28
test: fix an m.execute() invocation in check-realms

allisonkarlitskaya:
TestKerberos.testNegotiate() is using an undocumented feature of
m.execute() by passing the individual arguments of curl as a list rather
than a string.  This feature has long been broken: ssh just concatenates
the arguments back together anyway.  This is plainly illustrated by the
fact that this test has been calling m.execute() with

  '/usr/bin/curl -v -s'

as the first argument, and it's been working.

Split that out properly in preparation for an upcoming change to the
bots to make this function properly.

### DIFF
--- a/test/verify/check-realms
+++ b/test/verify/check-realms
@@ -810,7 +810,7 @@ ExecStart=/bin/true
         # Make sure negotiate auth is not offered first
         m.start_cockpit()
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
What a testament to our clean APIs that this *ever* worked :grinning:

PR #15242, opened 2021-01-29
node_modules/ improvements

martinpitt:
See individual commits for details. This does some parts of what is being discussed in PR #15235


martinpitt:
This is probably more @marusak reviewing matter as it mostly makes our lifes easier with Fedora/RHEL downstream testing. But it touches some aspects of the build simplification as well.

marusak (approved):
Thanks!

PR #15243, opened 2021-01-29
test: Fix check-realms curl invocation harder

martinpitt:
Commit 426bd42b31a1f was incompatible with
https://github.com/cockpit-project/bots/commit/13eff0d5743b
and also did not really make sense: `2>&1` is shell syntax, not an argv,
so instead run the whole command as a single shell string.

### DIFF
--- a/test/verify/check-realms
+++ b/test/verify/check-realms
@@ -810,9 +810,9 @@ ExecStart=/bin/true
         # Make sure negotiate auth is not offered first
         m.start_cockpit()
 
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Thanks!

but: Why no """?

martinpitt:
[test succeeded](https://logs.cockpit-project.org/logs/pull-15243-20210129-073314-484a2f21-debian-stable/log.html).

PR #15244, opened 2021-01-29
spec: Fix obsoleting of c-dashboard for also Fedora - FIX COMMIT MESSAGE ON MERGE!

marusak:
It was in condition for RHEL only.
Don't use 'Provides' as [1] states:
```
If a package supersedes/replaces an existing package without being a sufficiently
compatible replacement as defined above, use only the Obsoletes: line from the above example.
```

[1] https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages


martinpitt (approved):
Good catch, thank you!

martinpitt:
Let's please mention https://bugzilla.redhat.com/show_bug.cgi?id=1911825 in the commit message when merging.

PR #15252, opened 2021-01-31
test: Stop creating core dump artifacts

martinpitt:
They take a *lot* of space and upload bandwidth, and we use them so
rarely that it is not generally worth the trouble. We can re-enable them
on demand if we every run into crashes which can't easily be reproduced.


marusak (approved):
Thanks, indeed we rarely use those cores and if we would really need one, we can just enable it in a PR.
A bit weird to output message about downloading the core, but not saving it. I would comment out the first line as well, but not really  a nack.

martinpitt:
@marusak: The message is still correct for local runs. I *want* core dumps to get downloaded from the test VM to the host. I just don't want to attach them as log artifacts.

marusak:
ah I see, good point!

PR #15253, opened 2021-01-31
tests: Fix git help spewage when running from releases

martinpitt:
If there is no .git directory, i.e. when running from a release tarball,
then run-tests spits out an unsightly two-page long git error and help.
Check if the project is a git checkout before trying to determine
changed tests.


martinpitt:
Reproducer: Unpack a `make dist-gzip` or `make dist` tarball, and run `test/common/run-tests` (without arguments is fine, happens with any arguments). This does not actually break the tests, but ruins the test logs somewhat :grin:

marusak (approved):
Thanks!

PR #15254, opened 2021-01-31
tools: Drop cockpit-appstream split for RHEL/ELN ≥ 9

martinpitt:
Turns out that sources in BaseOS *can* ship binary RPMs in AppStream as
well. This got an explicit blessing, and will significantly simplify our
downstream packaging.

We may eventually reunite them for RHEL 8 as well, but for now let's fix
RHEL 9 and ELN.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1916251

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -66,10 +66,11 @@ Release:        1%{?dist}
 Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{version}/cockpit-%{version}.tar.xz
 %endif
 
-# in RHEL the source package is duplicated: cockpit (building basic packages like cockpit-{bridge,system})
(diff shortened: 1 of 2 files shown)


martinpitt:
This [debian-testing bond test failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15254-20210131-183925-a5af2acb-debian-testing/log.html) is our current top flake, hard to get green. Unrelated to here, but I'll take a look at that this morning.

KKoukiou (approved):

marusak:
So we just build it as in Fedora and RHEL tooling will ship it in the right repositories?

martinpitt:
@marusak : That's what Josh said, anyway. We may of course need to adjust the components of come packages if they slip into the wrong repository, but we can do that through tickets.

PR #15256, opened 2021-02-01
login: Always show the "Other options" header

### LINKED ISSUE #15255: Layout bug on login page re RequireHost
Cockpit version: 236
Page: Login

image

mvollmer:
Previously it would turn off when RequireHost was set to true in
cockpit.conf, but only when there wasn't any "/=HOST" element in the
URL already.

Let's just show it always, which should be less confusing and also
avoid a recently introduced layout bug.

Fixes #15255

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -564,7 +564,7 @@ if (window.NodeList && !NodeList.prototype.forEach)
         id("login-password-input").value = '';
 
         if (need_host()) {
-            hide("#option-group");
+            toggle_options(null, true);
             expanded = true;
         } else {
             hideToggle("#option-group", !connectable || form != "login");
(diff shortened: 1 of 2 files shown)


martinpitt:
[bastion host test](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15256-20210201-132108-2412407f-fedora-32-container-bastion/log.html) needs updating for the new behaviour.

mvollmer:
> [bastion host test](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15256-20210201-132108-2412407f-fedora-32-container-bastion/log.html) needs updating for the new behaviour.

Fixed.

mvollmer:
@garrett fixed the layout bug that this was a workaround for, so let's close this.

PR #15261, opened 2021-02-02
storage: use only PF4 components and stop including PF3 stylesheets

KKoukiou:
This commit:
* replaces icons with components from react-icons, also makes sure these
  are accessible
* replaces custom lists with DataLists component
* replaces all dialog fields like TextInputs, Checkboxes, Radio buttons with PF4
  components
* replaces custom dropdowns with Dropdown component

Todo:

- [x] The journal has lost the markup for repated entries, https://github.com/cockpit-project/cockpit/pull/15261#issuecomment-783273679
- [x] Spinner and progress text are too close together in dialogs, https://github.com/cockpit-project/cockpit/pull/15261#issuecomment-783277169
- [x] "Encryption options" dialog repeats itself, https://github.com/cockpit-project/cockpit/pull/15261#issuecomment-783286187

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -71,7 +71,7 @@ svg {
 
   > .pf-c-card__actions {
       flex-wrap: wrap;
-      button {
+      button:not(.pf-c-dropdown__toggle) {
           margin-bottom: 0.5rem;
       }
   }
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -215,11 +215,19 @@ import cockpit from "cockpit";
 
 import React, { useState } from "react";
 import {
-    Alert, Tooltip, TooltipPosition,
+    Alert,
     FormSelect, FormSelectOption,
+    Button,
+    Checkbox,
+    DataList, DataListItem, DataListCheck, DataListItemRow, DataListItemCells, DataListCell,
+    Form, FormGroup,
+    Radio,
     Select as TypeAheadSelect, SelectOption, SelectVariant,
-    DataList, DataListItem, DataListItemRow, DataListItemCells, DataListCell,
+    Spinner,
+    TextInput as TextInputPF4,
+    Tooltip, TooltipPosition,
 } from "@patternfly/react-core";
+import { InfoCircleIcon } from "@patternfly/react-icons";
 
 import { show_modal_dialog } from "cockpit-components-dialog.jsx";
 
@@ -230,48 +238,43 @@ import "form-layout.scss";
 
 const _ = cockpit.gettext;
 
-const Validated = ({ errors, error_key, explanation, children }) => {
-    var error = errors && errors[error_key];
-    var text = error || explanation;
-    // We need to always render the <div> for the has-error
-    // class so that the input field keeps the focus when
-    // errors are cleared.  Otherwise the DOM changes enough
-    // for the Browser to remove focus.
-    return (
-        <div className={error ? "ct-validation-wrapper has-error" : "ct-validation-wrapper"}>
-            { children }
-            { (text && text !== true) ? <span className="help-block dialog-error">{text}</span> : null }
-        </div>
-    );
-};
+const Row = ({ field, values, errors, onChange }) => {
+    const { tag, title, options } = field;
 
-const Row = ({ tag, title, errors, options, children }) => {
-    if (tag) {
-        const validated = (
-            <Validated errors={errors} error_key={tag} explanation={options.explanation}>
-                { children }
-            </Validated>
-        );
+    const error = errors && errors[tag];
+    const explanation = options && options.explanation;
+    const validated = (tag && errors && errors[tag]) ? 'error' : 'default';
 
-        if (title || title == "") {
-            if (options.widest_title)
-                title = (
-                    <>
-                        <div className="widest-title">{options.widest_title}</div>
-                        <div>{title}</div>
-                    </>
-                );
-            return (
+    function change(val) {
+        values[tag] = val;
+        onChange(tag);
+    }
+
+    const children = field.render(values[tag], change, validated);
+
+    if (title || title == "") {
+        let titleLabel = title;
+
+        if (options.widest_title)
+            titleLabel = (
                 <>
-                    <label className="control-label">{title}</label>
-                    <>{validated}</>
+                    <div className="widest-title">{options.widest_title}</div>
+                    <div>{title}</div>
                 </>
             );
-        } else {
-            return validated;
-        }
+        return (
+            <FormGroup label={titleLabel} validated={validated}
+                       helperTextInvalid={error} helperText={explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
+                { children }
+            </FormGroup>
+        );
     } else {
-        return children;
+        return (
+            <FormGroup validated={validated}
+                       helperTextInvalid={error || explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
+                { children }
+            </FormGroup>
+        );
     }
 };
 
@@ -279,22 +282,13 @@ function is_visible(field, values) {
     return !field.options || field.options.visible == undefined || field.options.visible(values);
 }
 
-const Body = ({ body, fields, values, errors, onChange }) => {
+const Body = ({ body, fields, values, errors, isFormHorizontal, onChange }) => {
     function make_row(field, index) {
         if (field.length !== undefined)
             return make_rows(field, index);
 
-        function change(val) {
-            values[field.tag] = val;
-            onChange(field.tag);
-        }
-
         if (is_visible(field, values))
-            return (
-                <Row key={index} tag={field.tag} title={field.title} errors={errors} options={field.options}>
-                    { field.render(values[field.tag], change) }
-                </Row>
-            );
+            return <Row key={index} field={field} values={values} errors={errors} onChange={onChange} />;
     }
 
     function make_rows(fields, index) {
@@ -302,9 +296,9 @@ const Body = ({ body, fields, values, errors, onChange }) => {
         if (rows.length === 0)
             return null;
         else if (index === undefined) // top-level
-            return <form className="ct-form">{ rows }</form>;
+            return <Form isHorizontal={isFormHorizontal !== false}>{ rows }</Form>;
         else // nested
-            return <div key={index} className="ct-form ct-form-box">{ rows }</div>;
+            return <FormGroup key={index}>{ rows }</FormGroup>;
     }
 
     return (
@@ -345,6 +339,7 @@ export const dialog_open = (def) => {
                         fields={nested_fields}
                         values={values}
                         errors={errors}
+                        isFormHorizontal={def.isFormHorizontal}
                         onChange={trigger => update(null, trigger)} />
         };
     };
@@ -400,7 +395,7 @@ export const dialog_open = (def) => {
             idle_message: (running_promise
                 ? <>
                     <span>{running_title}</span>
-                    <div className="spinner spinner-sm" />
+                    <Spinner className="dialog-wait-ct-spinner" isSVG size="md" />
                 </>
                 : null),
             extra_element: extra,
@@ -478,11 +473,13 @@ export const TextInput = (tag, title, options) => {
         options: options,
         initial_value: options.value || "",
 
-        render: (val, change) =>
-            <input data-field={tag} data-field-type="text-input"
-                   className="form-control" type="text" value={val}
-                   disabled={options.disabled}
-                   onChange={event => change(event.target.value)} />
+        render: (val, change, validated) =>
+            <TextInputPF4 data-field={tag} data-field-type="text-input"
+                          validated={validated}
+                          aria-label={title}
+                          value={val}
+                          isDisabled={options.disabled}
+                          onChange={change} />
     };
 };
 
@@ -493,10 +490,12 @@ export const PassInput = (tag, title, options) => {
         options: options,
         initial_value: options.value || "",
 
-        render: (val, change) =>
-            <input data-field={tag} data-field-type="text-input"
-                   className="form-control" type="password" value={val}
-                   onChange={event => change(event.target.value)} />
+        render: (val, change, validated) =>
+            <TextInputPF4 data-field={tag} data-field-type="text-input"
+                   validated={validated}
+                   aria-label={title}
+                   type="password" value={val}
+                   onChange={change} />
     };
 };
 
@@ -540,10 +539,12 @@ export const SelectOne = (tag, title, options) => {
         options: options,
         initial_value: options.value || options.choices[0].value,
 
-        render: (val, change) => {
+        render: (val, change, validated) => {
             return (
                 <div data-field={tag} data-field-type="select" data-value={val}>
-                    <FormSelect value={val} aria-label={tag} onChange={change}>
+                    <FormSelect value={val} aria-label={tag}
+                                validated={validated}
+                                onChange={change}>
                         { options.choices.map(c => <FormSelectOption value={c.value} isDisabled={c.disabled}
                                                                      key={c.title} label={c.title} />) }
                     </FormSelect>
@@ -562,14 +563,13 @@ export const SelectOneRadio = (tag, title, options) => {
 
         render: (val, change) => {
             return (
-                <span className="radio radio-horizontal" data-field={tag} data-field-type="select-radio">
+                <FormGroup isInline data-field={tag} data-field-type="select-radio">
                     { options.choices.map(c => (
-                        <label key={c.value}>
-                            <input type="radio" checked={val == c.value} data-data={c.value}
-                                     onChange={event => change(c.value)} />{c.title}
-                        </label>))
+                        <Radio key={c.value} isChecked={val == c.value} data-data={c.value}
+                            id={tag + '.' + c.value}
+                            onChange={event => change(c.value)} label={c.title} />))
                     }
-                </span>
+                </FormGroup>
             );
         }
     };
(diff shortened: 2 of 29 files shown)


pkg/storaged/things-panel.jsx
```diff
@@ -55,19 +55,19 @@ export class ThingsPanel extends React.Component {
...
         const actions = (
-            <StorageBarMenu id="devices-menu" label={_("Create devices")}>
-                { menu_item(null, _("Create RAID device"), () => create_mdraid(client)) }
-                { menu_item(lvm2_feature, _("Create volume group"), () => create_vgroup(client)) }
-                { menu_item(vdo_feature(client), _("Create VDO device"), () => create_vdo(client)) }
-            </StorageBarMenu>);
+            <StorageBarMenu id="devices-menu" label={_("Create devices")} menuItems={[
+                menu_item(null, _("Create RAID device"), () => create_mdraid(client)),
+                menu_item(lvm2_feature, _("Create volume group"), () => create_vgroup(client)),
+                menu_item(vdo_feature(client), _("Create VDO device"), () => create_vdo(client))].filter(item => item)} />
```
mvollmer, line 69: What does the `filter` do here?

KKoukiou: Some items here are sometimes null, this PF4 component get's really upset about null children.

pkg/storaged/dialog.jsx
```diff
@@ -471,32 +463,38 @@ export const dialog_open = (def) => {
 /* GENERIC FIELD TYPES
  */
 
-export const TextInput = (tag, title, options) => {
+export const TextInput = (tag, title, options, validated) => {
```
mvollmer, line 466: I don't understand the addition of `validated` here.  I don't think any caller actually passes in a value for it, and I dount that the PF4 TextInput has a `validated` property.

KKoukiou: Oh it does https://www.patternfly.org/v4/components/text-input#props

And I pass it to the children here https://github.com/cockpit-project/cockpit/pull/15261/files/7830bd747feb82212dad66d45976d56407b5c2d0#diff-5b932241940b997840c4960d1bda3e884d3ba162f9aa007b761920aa61fdd245R257

pkg/storaged/things-panel.jsx
mvollmer: Ah, makes sense. I had confused `filter` with `map`... d'oh.  What about writing `filter(item => item !== null)`?

KKoukiou: Sure I will replace on next repush.

pkg/storaged/dialog.jsx
mvollmer: > Oh it does https://www.patternfly.org/v4/components/text-input#props

Ohh!  Once again I wasn't able to find any proper documentation. Google send me here: https://www.patternfly.org/2020.04/documentation/react/components/textinput which has no "props" section.

mvollmer: > And I pass it to the children here

I think that doesn't actually go through the `TextInput` function, and will directly work on the result of the `render` function returned by `TextInput`.  I'll check this out more carefully.

(The things like `TextInput` are not React components, but I realize they look awefuly close.)

mvollmer: I have pushed a FIXUP with how I think it needs to be done.  In the case of `SelectOne`, the React.clone approach would stick the "validated" prop on the `div` instead of on the `FormSelect` where it needs to be.

KKoukiou:
@mvollmer, can I get another review on this one? There is also one test failing, TestStorageUnused.testUnused, which I can't figure out what wrong.

mvollmer:
There is now a gap between the dropdown button and the dropdown menu:

image

Previously:

image

KKoukiou:
> There is now a gap between the dropdown button and the dropdown menu:
> 
> image
> 
> Previously:
> 
> image

The issue here is that PF4 does not genuinely support buttons that expand dropdowns. I abused the Dropdown component to make this happen, so I don't expect that's ideal.

mvollmer:
RAID dialog:

image

Previously:

image

There is now a blue border all around a focused text input field. Is that intentional?

The disk selection should have the block device names right aligned.  The borders have changed.

mvollmer:
VDO dialog:

image

Previously:

image

The (i) bubbles are too far away from the text that they belong to.

mvollmer:
Text field of size sliders look differently now when focused:

image

Previously:

image

mvollmer:
Groups of checkboxes in dialogs look cramped and are not aligned with their label:

image

Previously:

image

mvollmer:
Dropdowns have a blue focus rings as well now:

image

but sometimes a blue focus line:

image

mvollmer:
Nested forms don't have a grey background:

image

Previously:

image

mvollmer:
Encryption keys table is not wide enough:

image

Previously:

image

mvollmer:
Field explanations have disappeared:

image

Previously:

image

mvollmer:
Checkbox groups without label are too far right:

image

Previously:

image

mvollmer:
Direct action buttons and the dropdown are too close together:

image

Previously:

image

KKoukiou:
> Dropdowns have a blue focus rings:
> 
> image
> 
> but sometimes a blue focus line:
> 
> image

The second variant is the expected. I will check you example.

mvollmer:
I found two issues that also exist on master.

The "widest title" is visible (Ouch!):

image

And the alert in "Edit NFS" has no margin:

image

I'll fix those separately.

mvollmer (changes requested):

pkg/storaged/storage-controls.jsx
```diff
@@ -208,21 +210,24 @@ export const StorageUsageBar = ({ stats, critical }) => {
...
 );
 
-export const StorageBarMenu = ({ label, children }) => {
-    const toggle = excuse => (
-        <Button variant="primary" isDisabled={!!excuse} data-toggle="dropdown" aria-label={label}>
-            <BarsIcon />
-        </Button>
-    );
+export const StorageBarMenu = ({ label, menuItems }) => {
+    const [isOpen, setIsOpen] = useState(null);
```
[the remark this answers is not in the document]
KKoukiou: Done

mvollmer:
> There is now a blue border all around a focused text input field. Is that intentional?
> [kkoukiou]: This is expected, all pF4 text inputs are blue when focused.

Indeed! (I don't like it, but *shrug*)

mvollmer:
> There is a bigger gap between the label and the input elements, which wastes some space.

The gap is still there, right?

mvollmer:
> Text field of size sliders look differently now when focused:
> 
> image
>  
> [kkoukiou]: Expected change

But they look differently than other text input fields, like the VDO name input. Do we haev two kinds now?

KKoukiou:
> There is a b

They are already compact, the spacing is what you see here are well if you specify isCompact, https://github.com/cockpit-project/cockpit/pull/15330

KKoukiou:
> > Text field of size sliders look differently now when focused:
> > image
> > [kkoukiou]: Expected change
> 
> But they look differently than other text input fields, like the VDO name input. Do we have two kinds now?

Ups right, that one I forgot to make PF4.

mvollmer:
> I don't think we should show grey background in nested forms. I even consider it a bug. If you disagree let's ask a designer.

It was designed with a grey background when we first introduced nested forms.  Personally, I think moving away from that because of technology limitations would be a bit weak.

Can we just keep using our old `ct-formt` layout system from `lib/form-layout.scss`?

KKoukiou:
> > I don't think we should show grey background in nested forms. I even consider it a bug. If you disagree let's ask a designer.
> 
> It was designed with a grey background when we first introduced nested forms. Personally, I think moving away from that because of technology limitations would be a bit weak.
> 
> Can we just keep using our old `ct-formt` layout system from `lib/form-layout.scss`?

I did not remove it because of technology limitations, I just  did not explicitly add it when I moved the dialogs to PF4 forms. In cockpit-machines we also have nested forms and we don't have or ever had grey background. form-layout is not meant to be used, was just a solution for when PF4 was not around. If we need some extra feature that PF4 does not offer we should report a RFE upstream and then it's ok to have a custom component. Since I see the design is a bit controversial here, let's just consult @garrett.

garrett:
> It was designed with a grey background when we first introduced nested forms. 

Andreas introduced that in ct-form.

The nested form should have labels on top, not labels on the side. This would probably help to resolve the need for having it styled with a grey background. (However, it might be a little weird on smaller screen sizes?)

I don't think PF designs for complex forms like this. Their progressive disclosure example doesn't even indent at all: https://www.patternfly.org/v4/components/form/design-guidelines/#use-progressive-disclosure

mvollmer:
> In cockpit-machines we also have nested forms and we don't have or ever had grey background.

I see, thanks! Nested forms with a greay background have a problem with disabled elements, which are also grey and thus blend into the background. So if removing the grey background here makes Cockpit more consistent, I am all for it.

garrett:
> removing the grey background here

Please note that more things need to be done than just removing the background. :wink:

mvollmer:
Checkboxes in a nested form have different row spacing than checkboxes in a normal form:

image

mvollmer:
> > There is a b
> 
> They are already compact, the spacing is what you see here are well if you specify isCompact, #15330

Ok, thanks!

KKoukiou:
> > removing the grey background here
> 
> Please note that more things need to be done than just removing the background. wink

Right but let's not make the other things part of this PR please.

KKoukiou:
> Checkboxes in a nested form have different row spacing than checkboxes in a normal form:
> 
> image

Right this is also expected. The first checkboxes, about encryption are form fields, and they don't assosiate to a label. The second group about mount options, are associated to a single label and belong to a group. 

See checkboxes in  https://www.patternfly.org/v4/components/form#horizontal vs https://www.patternfly.org/v4/components/form#horizontal-no-padding-top

garrett:
The labels should be on top in nested forms (as stated above). This can be done either with ct-form or in PF4.

The checkboxes should also have labels, as @KKoukiou says, as they're not part of the password confirmation, but encryption options.

KKoukiou:
> The labels should be on top in nested forms (as stated above). This can be done either with ct-form or in PF4.
> 
> The checkboxes should also have labels, as @KKoukiou says, as they're not part of the password confirmation, but encryption options.

PF4 decided if a form is horizontal or not by pf-m-horizontal class added to the `form` element. That being said, you can't really make some options horizontal and some vertical, because nested forms are not really nested `form` elements, since that would be wrong. So what you ask here, needs more thinking of how it can implemented, and actually I would rather get genuine support from it from PF4 upstream.

pkg/storaged/format-dialog.jsx
```diff
@@ -61,27 +61,28 @@ export function crypto_options_dialog_fields(options, visible, include_store_pas
...
-        CheckBoxes("crypto_options", "",
-                   {
-                       visible: visible,
-                       value: {
-                           auto: opt_auto,
-                           ro: opt_ro,
-                           extra: extra_options === "" ? false : extra_options
+        fields = [{ field: { title: _("Store passphrase"), tag: "store_passphrase" } }].concat(fields);
+
+    return fields.map(field => {
```
mvollmer, line 75: I don't understand this, I am afraid. The old code rturned just one `CheckBoxes` element, the new code returns multiple, one for each actual checkbox input.  Why does this need to be changed?

KKoukiou: its' the same difference you noticed here. https://github.com/cockpit-project/cockpit/pull/15261#issuecomment-779763836

Checkboxes can either belong to one label a a group, and in that case they should be rendered inside a `FromGroup` component. However, Checkboxes often are not part of one label, how the crypto options are, in this case they should be rendered one by one, each wrapped by a `FormGroup` element.

mvollmer: They belong to a group (except "Store passphrase"), but they didn't have a label. What about this:

image

mvollmer:
The little "danger" icon in a content row is now black:

image

pkg/storaged/format-dialog.jsx
KKoukiou: I am happy with this actually! Makes sense!

mvollmer:
The spinner in the format dialog does not appear anymore:

image

pkg/storaged/format-dialog.jsx
mvollmer: This change is also responsible for at least one test failure.

garrett:
Something more like this, perhaps?

storage-format-2021-02-16-1322

(I made this with Excalidraw; if you open the PNG with it, it'll let you edit it, as I embedded the data.)

KKoukiou:
> Something more like this, perhaps?
> 
> storage-format-2021-02-16-1322
> 
> (I made this with Excalidraw; if you open the PNG with it, it'll let you edit it, as I embedded the data. Or you can click this link to open it quickly: https://excalidraw.com/#json=6516990898339840,9-oKXT0E41-LgGi0cZG0uA (not sure how long it lasts).)

@garrett with the current PF4 compoent this is not really possible, I commented the reason above, https://github.com/cockpit-project/cockpit/pull/15261#issuecomment-779779979

I can take a look if this is possible to support upstream, but I think it should not block this PR.

garrett:
@KKoukiou: Oh, I missed that comment.

It's possible with ct-form, BTW. Also, it should be made possible in PF4 on a `pf-c-form__group` and/or `pf-c-input-group`.

Meanwhile, the form should collapse down to something like this on mobile view:

storage-format-mobile-2021-02-16-1322

(If PF4 isn't collapsing down, that's another issue.)

garrett:
Moved redesign to #15348.

If we can pick up any low-hanging fruit from the suggestions here, great. Otherwise, it should be done in a follow-up to address those issues.

mvollmer:
The hamburger should not be visible with limited access:

image

garrett:
What are labels aligned to? This is clearly a bug. 

Is it our bug or a PF4 one?

Screenshot from 2021-02-17 13-00-25

Screenshot from 2021-02-17 13-00-16

They're not aligned to the baseline, the middle, the bottom, or even the top of the text or widget.

I think this might be a PF4 bug in some recent version? I'll investigate.

garrett:
It looks like a PF4 bug with alignment and horizontal form elements... :frowning_face: 

Screenshot from 2021-02-17 13-08-19
Screenshot from 2021-02-17 13-07-27

I guess I'll craft an override and file a bug. I still need to finish our override for checkboxes (and probably radio buttons too) and file a bug for that as well. They really need to get their act together with regard to alignment. :disappointed:

garrett:
Filed https://github.com/patternfly/patternfly/issues/3868... I'm looking at a local override solution for us meanwhile.

mvollmer:
I'd say the triangle and spinner on the overview are now too small:

image

Let's make them medium:

image

mvollmer:
Jobs are no longer full width:

image

mvollmer:
The spinner in the content row is too small:

image

mvollmer:
The last column in the content row should be right aligned:

image

image

mvollmer:
How did I end up with two different sized triangles?

image

mvollmer:
> How did I end up with two different sized triangles?

There is a `flex-shrink: 1` active for that flex item, and the broswer will happily make the item smaller, and then for whatever reason the SVG scales itself down.

garrett:
> There is a flex-shrink: 1 active for that flex item, and the browser will happily make the item smaller, and then for whatever reason the SVG scales itself down.

If there's a flex shrinking in effect for that element, then it will shrink when it needs to. As SVG is scalable (that's the S in SVG), it'll rescale (in many circumstances, although that can be changed in some cases—but it's good for icons to be able to scale).

I'll take a look in a bit.

mvollmer:
> I'll take a look in a bit.

Please look at #5360 first, it changes this area drastically.

mvollmer:
The journal has lost the markup for repated entries:

image

Previously:

image

mvollmer:
Spinner and progress text are too close together in dialogs:

image

Previously:

image

mvollmer:
"Encryption options" dialog repeats itself:

image

garrett:
- padding on the right side of the action items in the header is incorrect (it should be smaller)
- padding all around the charts is huge; it should use similar padding to everything else
- padding around the cards' headers seems a bit large as well (most noticable at the top, but also underneith... to the left is OK however and to the right is not really visible due to the whitespace)
- we still have a problem with wrapping (but I think there's a long running PR about this, as well as an upstream PF issue, IIRC) — although I think we should probably just use gap in flex (which doesn't work for WebKit for the immediate future, but it'll finally be fixed in the next version https://caniuse.com/flexbox-gap)
- tables, especially those without headings should probably have a divider line
- filesystems table wraps on the size
- filesystems table has chunky size bars (they could also be made less wide)
- empty tables, such as NFS mounts in this screenshot, should not have table headings

Screenshot_2021-02-22 Storage - admin localhost localdomain(2)

 - many of the same issues above
 - what's "Create new logical volume" supposed to be? A button?
 - The layout falls apart at this size. I'm wondering if we really want to split logical and physical volumes to be side by side. We should definitely not around this size; it becomes debatable at larger widths.

Screenshot_2021-02-22 Storage - admin localhost localdomain(1)

- Same issues as above (makes sense, as this is the same screen at a smaller size)
- Having s smaller width causes lines to show up for the list in the logical volumes. The bottom one should not be there when used in a table like this.
- UUID and Capacity  have a lot of whitespace
Screenshot_2021-02-22 Storage - admin localhost localdomain

garrett:
What I was testing was this branch, but rebased, in a VM. I'm getting different results without it rebased and not in a VM. Hmm... odd. (Many of the same things still apply, but not all, oddly.)

KKoukiou:
> The journal has lost the markup for repated entries:
> 
> image
> 
> Previously:
> 
> image

Hm, so our current, pkg/lib/cockpit-component-logs.jsx needs PF3 to show proper styles. I will need to port it also to PF4. So, this is getting really big, I suggest I reimport PF3, to make the logs work, fix the last issues and we can remove the PF3 inclusion in followups.

garrett:
Many of the issues I pointed out must have been due to a mix and match of compilation of different branches, due to the build system issues with CSS. Sorry for the noise. Happy that the "issues" have been "resolved".

pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -71,7 +71,7 @@ svg {
 
   > .pf-c-card__actions {
       flex-wrap: wrap;
-      button {
+      button:not(.pf-c-dropdown__toggle){
```
garrett, line 74:
```suggestion
      button:not(.pf-c-dropdown__toggle) {
```

Suuuuuuper minor: This should have a space before the curly bracket.

garrett (approved):
Looks great now. :+1: 

I spotted a few minor issues that I need to address in other issues/PRs, but nothing with this PR anymore, now that I have it properly built. :blush: 

One very minor thing: There should be a space before the curly bracket. It doesn't do anything functionally nor should it change the CSS that gets compiled and sent to a browser. Therefore, I'm approving this and asking that you optionally consider the change, but no big deal otherwise.

garrett:
Should the NFS mounts have the empty state in the table? Or should it be in place of the table?

PF has an extra small empty state to be used in cards, like this: https://marvelapp.com/prototype/7c5cdhd/screen/72928590

Here's the important part of the mockup that I'm referencing: 
Screenshot_2021-02-23 x-small empty state(1)

Anyway, the point I'm making is that there isn't a set of length 0, which is what the in-table empty state is meant for, but it's something that's not defined. It's like `null` right now and we probably want something more like `undefined`. 

In friendlier words: It's the difference between doing a search (or filter) and finding nothing versus something not yet exiting.

This is related to the PR, but could be resolved in a follow-up, since it's just slight semantics and a little bit of style.

KKoukiou:
> Should the NFS mounts have the empty state in the table? Or should it be in place of the table?
> 
> PF has an extra small empty state to be used in cards, like this: https://marvelapp.com/prototype/7c5cdhd/screen/72928590
> 
> Here's the important part of the mockup that I'm referencing:
> Screenshot_2021-02-23 x-small empty state(1)
> 
> Anyway, the point I'm making is that there isn't a set of length 0, which is what the in-table empty state is meant for, but it's something that's not defined. It's like `null` right now and we probably want something more like `undefined`.
> 
> In friendlier words: It's the difference between doing a search (or filter) and finding nothing versus something not yet exiting.
> 
> This is related to the PR, but could be resolved in a follow-up, since it's just slight semantics and a little bit of style.

I agree, and this pattern exists in other places in cockpit. See machines snapshots. But it's not related to this PR. The NFS tables did not change from master at all here.

pkg/lib/patternfly/patternfly-4-overrides.scss
KKoukiou: Done

garrett:
> I agree, and this pattern exists in other places in cockpit. See machines snapshots. But it's not related to this PR. The NFS tables did not change from master at all here.

It didn't have a heading with an empty state before. If it's like that in master now, then I guess I missed previous commit at some point. Sigh. OK, that's a follow-up then.

garrett (dismissed):
This PR looks great to me!

mvollmer (dismissed):
Uff, thanks!

KKoukiou:
Oh the kdump failure is real, because I changed the spinner the PF4 in the dialog.jsx file. I will repost shortly.

mvollmer (approved):

martinpitt (approved):

PR #15262, opened 2021-02-02
machines: use toast notifications in the VM's details page

KKoukiou:
The VM details page can get quite long, and notifications were stored
before this commit on the top of the page. When the problematic operation was
initiated from somewhere in the bottom of the page, the notifications
panel was not visibile for usual screen sizes. Let's make the notifications
sticky as we do in the main page.

Previously:
Screen Shot 2021-02-02 at 10 12 19

Now:
Screen Shot 2021-02-02 at 10 11 30

### DIFF
--- a/pkg/machines/app.jsx
+++ b/pkg/machines/app.jsx
@@ -18,7 +18,7 @@
  */
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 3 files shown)


martinpitt:
Looks good to me, thank you!

skobyda (approved):
Looks good to me

PR #15263, opened 2021-02-02
lib: workaround missing dataLabel property in <Th> PF4 component

KKoukiou:
When we use `<Th>` as cell elements, the dataLabel property is not applied
correctly and we got a warning in the console. In addition in the mobile
mode the missing `data-label` was causing the cell labels to not exist.

Use the original `data-label` property to fix the visual regression until it's fixed upstream by:
https://github.com/patternfly/patternfly-react/pull/5406

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -333,7 +333,8 @@ const ComposableTableBasic = ({
 
                                     if (cells[cellIndex].header)
                                         return (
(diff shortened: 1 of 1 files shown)


KKoukiou:
@martinpitt this issue slipped my attention from that PR from yesterday. soorry

martinpitt (approved):
Ugh, `data-label` looks really weird in JS (like a subtraction :grin: ). Thanks for the upstream issue!

PR #15265, opened 2021-02-02
login: Never show the "> Other options" header when a host is required

mvollmer:


mvollmer:
The tests explicitly test for the previous behavior, so this isn't simply a bug.  Let's leave it as it is.

PR #15266, opened 2021-02-02
networkmanager: show an empty state if NM is not installed

### LINKED ISSUE #15245: network: show a  proper empty state when NM is not installed
Cockpit version: 236-1
OS: Arch Linux
Page: Network

Loading the network page, it remains completely empty.
This is a completely fresh installation of Arch. So it is entirely possible that just a package is missing.
But in contrast to other pages, I do not really get an error message I can understand.
image

## Journal:
```
-- Journal begins at Thu 2021-01-28 23:03:39 UTC, ends at Fri 2021-01-29 09:54:48 UTC. --
Jan 29 09:54:47 sec cockpit-tls[722]: cockpit-tls: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.
Jan 29 09:54:47 sec cockpit-tls[722]: gnutls_handshake failed: A TLS fatal alert has been received.
Jan 29 09:54:47 sec cockpit-tls[722]: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.
Jan 29 09:54:47 sec cockpit-tls[722]: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.
Jan 29 09:54:47 sec cockpit-tls[722]: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.
Jan 29 09:54:47 sec cockpit-tls[722]: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.
Jan 29 09:54:47 sec dbus-daemon[611]: [system] Activating via systemd: service name='org.freedesktop.hostname1' unit='dbus-org.freedesktop.hostname1.service' requested by ':1.37' (uid=0 pid=747 comm="cockpit-bridge ")
Jan 29 09:54:47 sec audit: BPF prog-id=31 op=LOAD
Jan 29 09:54:47 sec audit: BPF prog-id=32 op=LOAD
Jan 29 09:54:47 sec kernel: audit: type=1334 audit(1611914087.451:153): prog-id=31 op=LOAD
Jan 29 09:54:47 sec kernel: audit: type=1334 audit(1611914087.451:154): prog-id=32 op=LOAD
Jan 29 09:54:47 sec systemd[1]: Starting Hostname Service...
Jan 29 09:54:47 sec dbus-daemon[611]: [system] Successfully activated service 'org.freedesktop.hostname1'
Jan 29 09:54:47 sec systemd[1]: Started Hostname Service.
Jan 29 09:54:47 sec audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=systemd-hostnamed comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jan 29 09:54:47 sec kernel: audit: type=1130 audit(1611914087.495:155): pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=systemd-hostnamed comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jan 29 09:54:48 sec audit[1052]: USER_ACCT pid=1052 uid=0 auid=0 ses=1 msg='op=PAM:accounting grantors=pam_unix,pam_permit,pam_time acct="root" exe="/usr/bin/sudo" hostname=? addr=? terminal=? res=success'
Jan 29 09:54:48 sec kernel: audit: type=1101 audit(1611914088.228:156): pid=1052 uid=0 auid=0 ses=1 msg='op=PAM:accounting grantors=pam_unix,pam_permit,pam_time acct="root" exe="/usr/bin/sudo" hostname=? addr=? terminal=? res=success'
Jan 29 09:54:48 sec PackageKit[774]: get-updates transaction /14_bcabdeba from uid 0 finished with success after 105ms
```

## Steps to reproduce
It happens on every reload. Since it is likely a missing package, I can't tell how to reproduce it.

KKoukiou:
Previously the page was stuck with a spinner.

Fixes #15245

### DIFF
--- a/test/verify/check-networkmanager-basic
+++ b/test/verify/check-networkmanager-basic
@@ -128,6 +128,12 @@ class TestNetworkingBasic(NetworkCase):
                 b.wait_not_present("#networking-nm-crashed")
                 b.wait_visible("#networking-nm-disabled")
 
+        def assert_not_found():
+            # should hide graphs and actions and show the appropriate notification
+            b.wait_not_present("#networking-graphs")
+            b.wait_not_present("#networking-interfaces")
+            b.wait_visible("#networking-nm-not-found")
+
         self.login_and_go("/network")
         assert_running()
 
@@ -164,6 +170,14 @@ class TestNetworkingBasic(NetworkCase):
         b.click("#networking-nm-disabled button")
         assert_running()
         wait(lambda: m.execute("systemctl is-enabled NetworkManager"))
+
+        # /usr in coreos is readonly
+        if m.image not in ["fedora-coreos"]:
+        # remove the NM service file, test 'no-found' notification
+            # This works for ND tests since there is a self.addCleanup(m.execute, "systemctl enable --now NetworkManager") in the start of the test
+            self.restore_file('/usr/lib/systemd/system/NetworkManager.service')
+            m.execute("rm /usr/lib/systemd/system/NetworkManager.service && systemctl daemon-reload && systemctl stop NetworkManager")
+            assert_not_found()
 
         self.allow_journal_messages(".*org.freedesktop.NetworkManager.*")
 
--- a/pkg/networkmanager/app.jsx
+++ b/pkg/networkmanager/app.jsx
@@ -59,7 +59,7 @@ const App = () => {
     const plot_state_main = useObject(() => new PlotState(), null, []);
     const plot_state_iface = useObject(() => new PlotState(), null, []);
 
-    if (!model.ready || nmRunning_ref.current === undefined)
+    if (model.ready === undefined)
         return <EmptyStatePanel loading />;
 
     /* Show EmptyStatePanel when nm is not running */
@@ -78,6 +78,14 @@ const App = () => {
                                              {_("Troubleshoot…")}
                                          </Button>
                                      } />
+                </div>
+            );
+        } else if (!nmService.exists) {
+            return (
+                <div id="networking-nm-not-found">
+                    <EmptyStatePanel icon={ ExclamationCircleIcon }
+                                     title={ _("NetworkManager is not installed") } />
+
                 </div>
             );
         } else {
--- a/pkg/networkmanager/interfaces.js
+++ b/pkg/networkmanager/interfaces.js
@@ -205,6 +205,7 @@ export function NetworkManagerModel() {
     var objects = { };
 
     function complain() {
+        self.ready = false;
         console.warn.apply(console, arguments);
     }
 
@@ -1284,7 +1285,7 @@ export function NetworkManagerModel() {
     get_object("/org/freedesktop/NetworkManager", type_Manager);
     get_object("/org/freedesktop/NetworkManager/Settings", type_Settings);
 
-    self.ready = false;
+    self.ready = undefined;
     return self;
 }
 


mvollmer:
I have tested this by renaning the service and D-Bus name to "NetworkManager2" in the code. Works perfectly!

Would it be possible to have a integration test? (I.e., can we even open a browser without NM being installed?)

KKoukiou:
> I have tested this by renaning the service and D-Bus name to "NetworkManager2" in the code. Works perfectly!
> 
> Would it be possible to have a integration test? (I.e., can we even open a browser without NM being installed?)

I thought about it, I tested it by removing network-manager package from the test VM, then scp the cockpit dist code to /usr/share/cockpit. But I don't see how we can automate this in a test. Removing the package with dnf/apt will remove the cockpit-networkmanager as well.

mvollmer (dismissed):
Integration test would be awesome, but I would say it's optional since we rely on NM to get to a testable state.

martinpitt:
@KKoukiou : How about `systemctl mask NetworkManager && systemctl stop NetworkManager`?

KKoukiou:
> @KKoukiou : How about `systemctl mask NetworkManager && systemctl stop NetworkManager`?

But I really need the service to not exist to get to the 'not found' empty state. Mask && stop will just prevent the NM service from starting, it will not make it non-existent. 
(I did not try it out, maybe I missing something)

mvollmer:
> How about `systemctl mask NetworkManager && systemctl stop NetworkManager`?

We could fake a non-installed NM by removing it's systemd unit files and D-Bus config completely.  Maybe the best way to do that would be to erase it with rpm and the necessary --force options.

KKoukiou:
> > How about `systemctl mask NetworkManager && systemctl stop NetworkManager`?
> 
> We could fake a non-installed NM by removing it's systemd unit files and D-Bus config completely. Maybe the best way to do that would be to erase it with rpm and the necessary --force options.

Ok i will try, let's keep this one open, i will repost with a test.

mvollmer:
Here is what I came up with:

```
rm /usr/lib/systemd/system/NetworkManager.service && systemctl daemon-reload && systemctl stop NetworkManager
```

NM wont autolaunch via D-Bus, so we don't have to worry about that.

test/verify/check-networking-basic
```diff
@@ -165,6 +171,11 @@ class TestNetworkingBasic(NetworkCase):
         assert_running()
         wait(lambda: m.execute("systemctl is-enabled NetworkManager"))
 
+        # remove the DBus service, test 'no-found' notification
```
mvollmer, line 174: "DBus" is slightly confusing here, but not totally incorrect. :-)

test/verify/check-networking-basic
```diff
@@ -165,6 +171,11 @@ class TestNetworkingBasic(NetworkCase):
...
+        self.restore_file('/usr/lib/systemd/system/NetworkManager.service')
```
mvollmer, line 175: If this is a non-destructive test, we need to make sure that NM is started again, and that systemd is reloaded after the file is put back, no?

mvollmer (changes requested):
I think this needs more comments about how this works in a non-destructive test.  Why it is okay to stop NM without starting again, etc.

test/verify/check-networking-basic
KKoukiou: I guess you 're right, but I don't see any failing tests actually

KKoukiou: ah there is self.addCleanup(m.execute, "systemctl enable --now NetworkManager") in the start of the test

martinpitt: Correct, resolving.

martinpitt:
Dropping release-blocker; it's not a very important bug, let's not hold up the already delayed release. (If it still makes it in, fine of course)

KKoukiou:
@mvollmer the test is correct I think, not sure what I should do with this. Is it just the test comment? s/DBus/NetworkManager service/ ?

marusak:
Also needs rebase.

test/verify/check-networkmanager-basic
```diff
@@ -165,6 +171,11 @@ class TestNetworkingBasic(NetworkCase):
         assert_running()
         wait(lambda: m.execute("systemctl is-enabled NetworkManager"))
 
+        # remove the NM service file, test 'no-found' notification
+        self.restore_file('/usr/lib/systemd/system/NetworkManager.service')
+        m.execute("rm /usr/lib/systemd/system/NetworkManager.service && systemctl daemon-reload && systemctl stop NetworkManager")
```
martinpitt, line 176: This [does not work on coreos](https://logs.cockpit-project.org/logs/pull-15266-20210210-120426-ffc90af8-fedora-coreos/log.html#317).

I'd really rather this would not mess around in /usr. `systemctl mask --runtime` ought to work for most cases, you may need to mask some additional sockets or so (@mvollmer mentioned that masking was not enough, but I don't know how). Or move this into a separate destructive test and skip it on fedora-coreos.

martinpitt (changes requested):

test/verify/check-networkmanager-basic
mvollmer: > I'd really rather this would not mess around in /usr.

I am afraid we have to. We want to simulate a completely non-present NetworkManager.  One way would be to uninstall the package, but it seems that removing just the service file is enough.

> Or move this into a separate destructive test and skip it on fedora-coreos.

Skipping on fedora-coreos is appropriate.

mvollmer:
> the test is correct I think, not sure what I should do with this. Is it just the test comment?

Yes, just the comment.  It was surprising to both of us (I think) that this actually worked in a non-destructive test.  That should be explained, or maybe better, the test could be marked as destructive.

KKoukiou:
> > the test is correct I think, not sure what I should do with this. Is it just the test comment?
> 
> Yes, just the comment. It was surprising to both of us (I think) that this actually worked in a non-destructive test. That should be explained, or maybe better, the test could be marked as destructive.
 Actually we both had overseen the addCleanup that enables NM in the start of the test. I updated with comment to be more clear anyway.

mvollmer:
> Actually we both had overseen the addCleanup that enables NM in the start of the test. I updated with comment to be more clear anyway.

Ah, it's in the test itself! So someone reading the whole test (as I should have) would not be confused.  
But doe we not need `systemctl daemon-reload` as well?  Anyway, I am happy now. :-)

mvollmer (dismissed):

martinpitt:
Still needs to be skipped on coreos.

mvollmer (approved):

test/verify/check-networkmanager-basic
```diff
@@ -165,6 +171,14 @@ class TestNetworkingBasic(NetworkCase):
         assert_running()
         wait(lambda: m.execute("systemctl is-enabled NetworkManager"))
 
+        # /usr in coreos is readonly
+        if m.image not in ["fedora-coreos"]:
```
martinpitt, line 175: I would have used `if m.ostree_image:` or `m.needs_writable_usr()`, but we can clean this up if/when it gets in the way.

martinpitt (approved):
Cheers!

PR #15272, opened 2021-02-03
Fix accessibilty issues caused by missing properties in FormSelect

KKoukiou:


pkg/systemd/terminal.jsx
```diff
@@ -101,6 +101,7 @@ const _ = cockpit.gettext;
                                 </ToolbarItem>
                                 <ToolbarItem>
                                     <FormSelect onChange={this.onThemeChanged}
+                                                aria-label={_("Appearance")}
                                                 aria-labelledby="theme-select"
```
marusak, line 105: Isn't this "aria-labelledby" working?

KKoukiou: For accessibility it does, but we still get an error from PF-react because they don't count with it. https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/FormSelect/FormSelect.tsx#L42

Of course we can open an issue so that they also count with aria-labelledby but i thought it's not such big of a deal.

marusak: Then we might not need aria-labelledby? (I just think there is no need to say the same thing twice)

marusak (approved):
That being said, definitely not a nack :)

pkg/systemd/terminal.jsx
KKoukiou: Right, we don't need it, let me repost.

PR #15274, opened 2021-02-03
DEBUG: testNetworkingFirewall flake

KKoukiou:


mvollmer:
Firewalld seems to be stopped and disabled unexpectedly.

```
warning: {"problem":null,"name":"org.freedesktop.DBus.Error.NoReply","message":"Message recipient disconnected from message bus without replying"}
```

pkg/networkmanager/firewall-client.js
```diff
@@ -77,15 +77,17 @@ function initFirewalldDbus() {
...
         if (!firewall.enabled) {
             firewall.dispatchEvent('changed');
             return;
         }
 
         getZones()
                 .then(() => getServices())
-                .then(() => firewall.debouncedEvent('changed'))
-                .catch(error => console.warn(error));
+                .then(() => { console.info('debouncedEvent changed'); firewall.debouncedEvent('changed') })
```
mvollmer, line 89: This should probably be `return firewall.debounced...` now it always returns `undefined`.

mvollmer:
This looks interesting.

When we get a owner for the first time, there is a successful getZones call:

```
> info: initFirewalldDbus {"firewall":{"installed":true,"enabled":true, ..},"event":{"isTrusted":false},"owner":":1.59"}
> info: getZones
> info: getZones {"public":{"interfaces":["eth1"]}}
> info: getServices
> info: debouncedEvent changed
```

Then the firewall is disabled:
```
> info: initFirewalldDbus {"firewall":{"installed":true,"enabled":false, ...},"event":{"isTrusted":false},"owner":null}
> info: initFirewalldDbus {"firewall":{"installed":true,"enabled":false, ...},"event":{"isTrusted":false},"owner":null}
```

The it is enabled again, and immediately disabled. There should be a getZones call, but there is only a error message instead.
```
> info: initFirewalldDbus {"firewall":{"installed":true,"enabled":true, ...},"event":{"isTrusted":false},"owner":":1.81"}
> info: initFirewalldDbus {"firewall":{"installed":true,"enabled":false, ...},"event":{"isTrusted":false},"owner":null}
> warning: {"problem":null,"name":"org.freedesktop.DBus.Error.NoReply","message":"Message recipient disconnected from message bus without replying"}
```

martinpitt:
Could this be just something simple, like restarting firewalld.service often enough and running into systemd's restart limit? Placing some `systemctl reset-failed firewalld` in between would help with that. (But, I didn't look into the logs at all, just a random idea)

mvollmer:
> Could this be just something simple, like restarting firewalld.service often enough and running into systemd's restart limit?

No, that's not it.

mvollmer:
I have observed two things:

- The browser console log as captured by our test infrastructure is very very incomplete. I had to use TEST_SHOW_BROWSER=t and look at the console in the actual browser window to see all the expected console messages.  Only then did it all make sense.

- The "owner" signal of the dbus client is fired three to five times in a row for the same name.  The leads to a lot of concurrent initialization activities and might cause all kind of weirdness, I guess.

I'll add a commit here that suppresses superfluous calls to getZone et al. Let's see what happens.

I also leave all my ad-hoc debugging console messages in. Hopefully it becomes obvious that many are missing.

pkg/networkmanager/firewall-client.js
mvollmer: That was a red herring, debouncedEvent doesn't return anything important.

mvollmer:
> I'll add a commit here that suppresses superfluous calls to getZone et al. Let's see what happens.

That hasn't helped at all... *sniff*

mvollmer:
>     * The browser console log as captured by our test infrastructure is very very incomplete.

Aha, the cdp-driver explicitly ignores repeated console messages.  This is not useful during debugging, but I am not sure right now how a ood compromise would look. I'll just comment that out while inestigating here.

mvollmer:
> * The "owner" signal of the dbus client is fired three to five times in a row for the same name.

This doesn't actually seem to happen... I am sure I saw it when looking at the console log in the real browser... weird.

mvollmer:
I can pretty reliably reproduce this now locally, by swtching the firewall on and off in a tight loop until it fails. Let's see...

marusak:
> Aha, the cdp-driver explicitly ignores repeated console messages. This is not useful during debugging, but I am not sure right now how a ood compromise would look. I'll just comment that out while inestigating here.

Maybe we can do that if you are in trace mode, then we would show all messages. That would prevent logs getting too long, but you still could get into proper debugging mode with all the logs.

mvollmer:
> That would prevent logs getting too long

Do we actually have a problem with too many calls to console.log in third-party code? I know there are too many policy violation messages etc, but those don't come via console.log, no?  Maybe we should even have a white list of known annoying messages that we don't want to see more than once.

mvollmer:
Ok, I am pretty confident now that the flake is caused by too fast disabling/enabling of the firewall.  Our firewall-client doesn't seem to handle a disappearing D-Bus service too well when there are still method replies outstamding (although I haven't figured out the precise details).
One could argue that stopping firewalld explicitly via "systemctl stop firewalld" should not cause the process to exit until all method calls have been replied to. Or that our client needs to be robust against arbitrary disconnections. But I think that effort is not worth it.
Instead, let's just slow down the test a bit.

What happens in slfightly more detail:
- firewalld is started
- client notices a new owner
- client starts the getZones(), getServices() sequence
- as soon as getZones() comes up with the expected number of active zones, the test stops the firewalld process
- the getServices call gets an error

This doesn't fully explain everything we have seen, I think, but I would expect that waiting a bit after toggling the firewall to allow our client to settle down would avoid a lot of "unsupported" scenarios.

It's always unsatisfying to solve flakes with a time.sleep, but since we are talking about a UI here and not an API, I'd say it is justified.

mvollmer:
With 4b116ff, testStress has failed everywhere.  The logs show D-Bus errors, overlapping enable/disable actions, and also the multiple redundant owner change signals that I previously thought wouldn't actually happen.
Let's slow down testStress and have another go.

mvollmer:
> Let's slow down testStress and have another go.

That didn't change a whole lot, damn.  There was at least one run that seemed to have timed out, which is a good sign, but other runs have failed in the same way as before.

mvollmer:
(Just for the record I can run testStress locally successfully. So when it failed earlier it probably wasn't due to the same reason that causes the flake.)

mvollmer:
Ok, serializing enable and start seems to help. Now testStress fails either because of timeout, or because firewalld fails to load all expected zones. I didn't find any errors from our client anymore (but I might have missed them). testNetworkingPage still runs into errors, but that one isn't slowed down yet. So let's try that next.

mvollmer:
> but that one isn't slowed down yet. So let's try that next.

Seems to have helped.  I'll make a PR with all the relevant bits.

mvollmer:
>  I'll make a PR with all the relevant bits.

#15289

mvollmer:
@KKoukiou, I _think_ I am done here.  To summarize:

- The test were too fast in switching the firewalld service off after just turning it on. The client got into a bad state somehow, I think, but I don't have precise evidence for this, except observing some unexpected D-Bus errors.
- Enabling and starting the firewalld service was done in parallel, which seems to have confused some part of the stack. The first D-Bus call after getting a new owner failed sometimes with "not-found". That doesn't make much sense, but doing this sequentially has helped a lot. *shrug*
- We would get multiple "owner" signals for the same new owner, but we should only react to the first.

All of these things together seem to be responsible for the flaky behavior of the test.

I had a stress test going that did run reliable with all three issues fixed.  However, it did uncover a new one:

- The firewalld service itself does not always load all active zones after a start.

See for example https://logs.cockpit-project.org/logs/pull-15274-20210205-110347-3e20db41-fedora-32/log.html#6, where we expect 3 zones but only 2 are loaded.  All 3 zones had been loaded many times successfully in the stress test, until suddebly one was missing.
We could investigate this further to see whether the fault with our client or with firewalld.

mvollmer:
Here is a failure: https://ec2-54-89-13-31.compute-1.amazonaws.com/logs/pull-15274-20210208-114654-e411ec70-rhel-8-4/log.html They are pretty rare now.

martinpitt:
Eww, since when do we get such awful log URLs.. they are supposed to be "logs.cockpit-project.org", our proper DNS. I'm sure I broke that somehow this morning with my sink re-deployment.

mvollmer:
I think we are left with a D-Bus error resulting in an uncaught JavaScript exception, somewhere in fetchServiceInfos maybe. Let's catch it.

mvollmer:
We get NoReply errors here, but they are properly handled now: https://logs.cockpit-project.org/logs/pull-15274-20210209-115822-d1259572-rhel-8-4/log.html

martinpitt:
Fixed in #15289

PR #15275, opened 2021-02-03
login: Fix layout for "connect to" for requireHost

### LINKED ISSUE #15255: Layout bug on login page re RequireHost

garrett:
Fixes #15255

### DIFF
--- a/pkg/static/login.css
+++ b/pkg/static/login.css
@@ -144,16 +144,21 @@ p {
   margin-left: auto;
 }
 
-#option-group {
+/* Bump options & connection away from user/pass */
+#option-group,
+#option-group[hidden] + #server-group {
   margin-top: 2rem;
-  margin-bottom: 2rem;
 }
+
+#option-group {
+  margin-bottom: 1rem;
+}
+
 #server-group {
-  margin-top: -1rem;
   margin-bottom: 2rem;
 }
 
-.form-group:not([hidden]) + .form-group {
+.form-group:not(:first-child) {
   margin-top: 1rem;
 }
 


garrett:
I tested locally in-browser only, by modifying the CSS live in Firefox.

I tried using bind mounts to test this (how I usually test the login page), but got some weird error when trying to load the page. I guess the installed version of Cockpit and what's in tree aren't compatible?

Also, I could build a VM, but I don't know how to flip the RequireHost = true for a VM, so that line of testing is also not available to me.

Therefore, I've marked this as a draft. Please test in the normal mode and with RequireHost = true and share screenshots here. It _should_ work, but I really want to see it in action to be certain it does. Thanks!

mvollmer:
To test this, add this to /etc/cockpit/cockpit.conf:

```
[WebService]
RequireHost=true
```

and then open the login page for "/", i.e., without any "/=host" part already in it.

mvollmer:
> and then open the login page for "/", i.e., without any "/=host" part already in it.

You also need to restart Cockpit after changing cockpit.conf, `systemctl restart cockpit`.

mvollmer:
Looks like this for me:

image

There seems to be more space above "Connect to" than above "Password".

mvollmer (approved):

garrett:
> There seems to be more space above "Connect to" than above "Password".

That's 100% intentional. That's how it is by default too, with the expander.

User name & password are related. Connect to is not.

garrett:
> To test this, add this to /etc/cockpit/cockpit.conf:
...
> You also need to restart Cockpit after changing cockpit.conf, systemctl restart cockpit.

Right, I said I did that and it didn't work. I get an error.

Can you also post a default screen w/o the requireHost? I want to make sure I didn't break that too.

garrett:
Marking as ready for review, but I'd still like to see the default log in as a screenshot before this is merged.

mvollmer:
> Marking as ready for review, but I'd still like to see the default log in as a screenshot before this is merged.

Here:

image

image

garrett:
Awesome! Thanks! Looks good to me. :+1:  I'm glad I did this correctly. :grin:

PR #15276, opened 2021-02-03
login: Add margin under the error

### LINKED ISSUE #15255: Layout bug on login page re RequireHost

garrett:
Fixes #15255

### DIFF
--- a/pkg/static/login.css
+++ b/pkg/static/login.css
@@ -813,7 +813,7 @@ a:hover {
 }
 
 .dialog-error {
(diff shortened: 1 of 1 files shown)


garrett:
Error looks like this prior to the patch:
image

This PR adds 2rem of space between the error and the user name, which is the same amount of space between the password and the additional options / connect to host entry.

martinpitt (approved):
Thanks! This looks better indeed.

PR #15277, opened 2021-02-03
Add a global CSS rule for extending pages height and avoid double scrollbar

KKoukiou:
Remove all per page CSS for this purpose.

The single visual difference by this is that loading page spinner is in
the center of the page now, instead of the top that it was before.
This is the expected behavior.

Please review carefully :) Prone to :bug:

### DIFF
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -108,12 +108,6 @@ a.disabled:hover {
     padding-top: 4px;
 }
 
-/* HACK: Workaround for https://github.com/patternfly/patternfly/issues/174*/
-
-.page-ct {
-    margin-top: 20px;
-}
-
 .highlight-ct {
     background-color: var(--color-link-hover-bg);
 }
@@ -548,4 +542,14 @@ input[type=number]:not(.pf-c-form-control) {
     font-size: .875rem;
     line-height: 1.5;
     border-radius: .2rem;
+}
+
+// Let PF4 handle the scrolling through page component otherwise we might get double scrollbar
+html:not(.index-page) body {
+  overflow-y: hidden;
+
+  // Ensure UI fills the entire page (and does not run over)
+  .ct-page-fill {
+    height: 100% !important;
+  }
 }
--- a/pkg/systemd/terminal.scss
+++ b/pkg/systemd/terminal.scss
@@ -2,10 +2,7 @@
 @import "_global-variables.scss";
 @import "../../node_modules/@patternfly/patternfly/components/Button/button.css";
 @import "../../node_modules/@patternfly/patternfly/components/Toolbar/toolbar.scss";
-
-html, body {
-        height: 100%;
-}
+@import "../lib/page.scss";
 
 .console-ct-container {
     height: 100%;
@@ -55,10 +52,6 @@ html, body {
 
 .white-theme {
     background-color: #ffffff;
-}
-
-#terminal {
-    height: 100%;
 }
 
 .terminal-group {
--- a/pkg/apps/application.css
+++ b/pkg/apps/application.css
@@ -3,11 +3,6 @@
 @import "../lib/table.css";
 @import "../lib/cockpit-components-table.scss";
 
-.page-ct {
-    /* Remove margin above the header */
-    margin: 0;
-}
-
 .app-list-empty {
     color: var(--color-subtle-copy);
     text-align: center;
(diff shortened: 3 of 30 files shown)


garrett:
This is cool! Thanks for changes. It seems fine with a quick glance at the PR.

I'll build and test locally and look through the diff more closely.

pkg/lib/page.scss
```diff
@@ -549,3 +549,11 @@ input[type=number]:not(.pf-c-form-control) {
     line-height: 1.5;
     border-radius: .2rem;
 }
+
+html:not(.index_page) body {
+    // Ensure UI fills the entire page (and does not run over)
+    > div:first-child {
```
KKoukiou, line 555: This is wrong - will not affected pages which have other div before the main page div, which is very common. Removing it to see what happens.

pkg/systemd/terminal.scss
```diff
@@ -3,10 +3,6 @@
 @import "../../node_modules/@patternfly/patternfly/components/Button/button.css";
 @import "../../node_modules/@patternfly/patternfly/components/Toolbar/toolbar.scss";
 
-html, body {
-        height: 100%;
```
KKoukiou, line 7: terminal page broke - it does not include page.sccs apparently.

KKoukiou (commented):
Not all pages include page.scss - this needs to be moved in patternfly-4-overrides probably.

garrett:
## Network
- Spinner is at the top (might be related)
- Blank space at the bottom (I did control-minus to shrink the page): Screenshot_2021-02-03 Networking - garrett Rain

## Diagnostic Reports
- This page is weird... 
Screenshot_2021-02-03 Diagnostic Reports - garrett Rain

## Kernel Dump
- Page is centered vertically; it shouldn't be in this case. But it should be stretched still.
- Also, there's a vertical scrollbar, even with all this space!?
- Screenshot: 
Screenshot_2021-02-03 Kernel Dump - garrett Rain

## Software Updates, ostree edition

- The page doesn't stretch... but this is an external repo (so I'm not sure if it'll inherit from Cockpit... I think it doesn't directly)
   Screenshot_2021-02-03 Software Updates - garrett Rain

## Terminal
- It's broken, but you mentioned that you fixed it. I'm updating now. Perhaps some of the above is out-of-date too.

martinpitt:
Dropping release-blocker. This is too intrusive for a last-minute landing, let's instead just do a pinpointed fix on the updates page, and land this after the release.

pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -76,3 +76,13 @@ svg {
       }
   }
 }
+
+// Let PF4 handle the scrolling through page component otherwise we might get double scrollbar
+html:not(.index-page) body {
+  overflow-y: hidden;
+
+  // Ensure UI fills the entire page (and does not run over)
+  div.pf-u-h-100 {
```
garrett, line 85: 1. This shouldn't be tied to a `<div>` element. We really should not have CSS bound to elements _and_ classes (unless necessary... but it's not necessary ~99% of the time).
2. This doesn't look like a PF class, but a CT one? It's Cockpit-specific, right? Something like ct-page-fill might be better, as it's more descriptive and signifies that it's a Cockpit-specific utility class related to the page.

garrett (dismissed):
We shouldn't have Cockpit specific CSS masquerade as PF CSS.

I haven't done an in-Cockpit review; I just looked at the CSS diff on this new first-pass. I'll look in a bit.

pkg/lib/patternfly/patternfly-4-overrides.scss
KKoukiou: It's actually a PF class that I found through their website but I can't find it in the shipped code so I just copied their code. I can rename it of course.

garrett: Aha. Weird that it isn't shipped. Perhaps it's new and isn't in the release of PF we're using?

KKoukiou: They have some stylesheets only for their website, and I think it's part of these.

garrett (approved):
Works here! LGTM! :+1: 

I've looked through every page both on my machine and in a VM, and I zoomed way out to make things tiny to check.

The only issue is with cockpit-ostree, but I'm 92% sure that needs to be fixed in another repo. :wink:

PR #15278, opened 2021-02-03
updates: Get rid of the page header

skobyda:
Page header is usually no longer present on other cockpit pages


marusak (approved):
Nice, thanks!

PR #15282, opened 2021-02-04
Run affected tests three times on pkg/<page> changes

KKoukiou:


test/common/run-tests
```diff
@@ -209,15 +209,30 @@ def run(opts, image):
...
+        if len(changed_tests) > 3:
+            changed_tests = []
+
+        # Detect affected tests from changed pkg/* subdirectories
+        # If affected tests get detected from pkg/* changes, don't apply the
+        # "only do this for max. 3 check-* changes" (even if the PR also changes ≥ 3 check-*)
+        cmd = ["git", "diff", "--name-only", "HEAD", "origin/master", "pkg/"]
+        r = subprocess.run(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        if r.returncode == 0:
+            changed_pkgs = [pkg.decode("utf-8").split('/')[1] for pkg in r.stdout.strip().splitlines()]
```
martinpitt, line 229: I suggest to parse this into a set, not a list, as there are probably many duplicates.

test/common/run-tests
```diff
@@ -209,15 +209,30 @@ def run(opts, image):
...
+
+        cmd = ["ls", opts.test_dir]
+        r = subprocess.run(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        test_files = [os.path.join(opts.test_dir, test_file.decode("utf-8")) for test_file in r.stdout.strip().splitlines()]
```
martinpitt, line 233: Four lines down it already enumerates the test files with `glob.glob(os.path.join(opts.test_dir, "check-*"))`. Please move that up, save it into a var, and re-use it here.

martinpitt (changes requested):
Nice! I suggest to check this with machines and realms once this works in general. These are the "interesting" and expensive cases which may require us to bump the 2 hour timeout (and we should really do that -- even though testing will take long, it's even more painful to introduce new regressions).

test/common/run-tests
martinpitt: It's just being used in the `in changed_packages` further below, so not really important after all. Feel free to just keep as-is and resolve this.

test/common/run-tests
KKoukiou: Ah thanks I did not notice that one.

KKoukiou:
So retrying machines and realm tests did not seem to cause any timeout. https://github.com/cockpit-project/cockpit/commit/6e29e098c8dd85e2cab34fd525c0c5e2dcc4a4b9

martinpitt (approved):
Excellent, thank you! Let's land this, and bump timeouts if/when we run into failures.

PR #15286, opened 2021-02-05
test: Work around selenium Edge driver bug (scroll into view version)

martinpitt:
Commit 9113eb5f60 triggered the old Edge selenium driver bug again where
elements that are scrolled off-page too far are not "visible" to
Selenium any more. Scroll the udev service into view,
similar to commit b960b0fc.

### DIFF
--- a/test/selenium/selenium-base.py
+++ b/test/selenium/selenium-base.py
@@ -33,6 +33,9 @@ class BasicTestSuite(SeleniumTest):
         self.wait_frame("services")
         self.wait_id("services-list")
         self.click(self.wait_text("Socket", cond=clickable))
+        if os.environ.get("BROWSER") == 'edge':
+            # HACK: Edge does not see elements (kdump menu entry) which are below the visible window area
+            self.driver.execute_script('document.getElementById("systemd-udevd-control.socket").scrollIntoView()')
         self.wait_text("udev")
         self.wait_id("services-list")
         self.click(self.wait_text("Target", cond=clickable))


martinpitt:
I'd prefer this version if it works. If not, we can use #15287.

martinpitt:
Still failed, I dunno. I tested this selector in Firefox inspector, and it worked fine. But there's only so much energy that I want to pour into this (especially today on day of learning), so perhaps let's just use #15287?

test/selenium/selenium-base.py
```diff
@@ -33,6 +33,9 @@ def test30ChangeTabServices(self):
         self.wait_frame("services")
         self.wait_id("services-list")
         self.click(self.wait_text("Socket", cond=clickable))
+        if os.environ.get("BROWSER") == 'edge':
+            # HACK: Edge does not see elements (kdump menu entry) which are below the visible window area
+            self.driver.execute_script('document.getElementById("systemd-udevd.service").scrollIntoView()')
```
marusak, line 38: heh, from screenshots, the page is on `sockets` but this should be in `services`. So somehow this is on a wrong tab

marusak: oh, it is on a right page, but the id should be `systemd-udevd-kernel.socket` or `systemd-udevd-control.socket`

martinpitt: D'oh, thanks! Pushed

martinpitt:
This got a [little further](https://logs.cockpit-project.org/logs/pull-15286-20210205-110737-0eff0029-fedora-32-selenium-edge/log.html#3) -- but now we'd need to *scroll backwards* again to see the table headings again. :exploding_head:

marusak:
I mean, can we just go with https://github.com/cockpit-project/cockpit/pull/15287?
This is going to he HACK and HACK and we will probably drop it all together soon anyway.

martinpitt:
Yeah, let's. I don't really care.

PR #15289, opened 2021-02-05
networkmanager: Derace firewall toggling

mvollmer:


mvollmer:
Upps.commit message word salad.

test/verify/check-networking-firewall
```diff
@@ -83,11 +83,13 @@ class TestFirewall(NetworkCase):
         wait_unit_state(m, "firewalld", "active")
 
         b.wait_in_text("#networking-firewall-summary", "{} active zone".format(active_zones))
+        time.sleep(10) # give Cockpit time to settle
```
marusak, line 86: uh, 10s is looong time, that is something even user can hit. How does it propagate itself? Is something broken if you toggle it too quickly?

What does "settling" involve? I assume some API calls? If so, then we should either:
1. Wait in UI until those calls finish - I guess each call has some visible effect on the page, otherwise we would not do them
2. Show spinner/disable controls... while stuff is in progress

mvollmer:
> uh, 10s is looong time, that is something even user can hit.

Yes, the 10 seconds are supposed to account for loaded CI machines.  The actual settle time is probably sub-second.

test/verify/check-networking-firewall
mvollmer: > What does "settling" involve? I assume some API calls? If so, then we should either:
> 
>     1. Wait in UI until those calls finish - I guess each call has some visible effect on the page, otherwise we would not do them
> 
>     2. Show spinner/disable controls... while stuff is in progress

Ideally, yes.  However, there is no indication on the networking page except for the number of active zones and the activity is mostly triggered by signals from firewalld, I think.  So there is no well defined "we are done with startup" point I guess.

Of course, with more work we can improve this (and I don't really like sleeps like this at all, either).

We need to wait for the getServices calls to finish after getZones has returned the desired number of active zones.  The "change" signal is actually only emitted after the getServices is finished, hmm. The page is re-rendered independently of that apparently, and thus this flake is a consequence of the rewrite to React (which probably has changed the page updating dramatrically).

So, yeah. :-) Let's try to do better.

mvollmer: It turns out that we don't need to slow down the tests at all.  Our client is robust enough, except for one bug, which is now fixed.  Thanks for not letting me get away with this!

martinpitt (approved):
Amazing detective work, thanks! And I'm glad that for once it wasn't the tests which were broken :grin:  

Retrying, the debian-testing bonding failure is pretty annoying. But, our "only land green PRs regime.."

martinpitt:
BTW, thanks to @KKoukiou all the networking tests now ran three times :tada:

marusak:
[bond](https://logs.cockpit-project.org/logs/pull-15289-20210209-191839-e9b9b79e-debian-testing/log.html) again, but it it fails now again, lets just merge it red as other PRs are failing on tests this PR is fixing :)

martinpitt:
Meh, it's pretty nigh impossible to get the bond test to succeed three times in a row -- but that is unrelated, and this PR fixes another top-priority flake. So we have to compromise here still.

PR #15291, opened 2021-02-08
machines: add data-vm-transient attribute for the VM detail page and change data-row-transient to data-vm-transient

yunmingyang:


KKoukiou (changes requested):
Super, thanks, can you please also remove the back and forwards from https://github.com/cockpit-project/cockpit/blob/master/test/verify/check-machines-lifecycle#L140

and git grep data-row-transient, there is one more occurance of this string.

KKoukiou (approved):
Looks good, I don't think you need to escape the " in the wait_visible but it's not an issue.

For you information we stoped using Closes in PRs, github now allows us to get the PR that merged a commit by looking it in the commit itself.

Issue #14695, opened 2020-10-03
unable to elevate to administrative access when fingerprint auth is enabled


giorgiopiatti:
Cockpit version: 228
OS: Fedora 32

Hello

When I enable the option to authenticated by fingerprint in the OS  I cannot anymore elevate my access to administrative mode.
Steps to reproduce the issue:

1. Enable fingerprint auth
2. Login web console
3. click "turn on administrative access" 
4. click  "authenticate"

Following the above steps, I get "protocol-error".

If I disable instead the fingerprint authentication, and by following the same steps as above I can successfully authenticate by providing my user password. 

```
Oct 03 21:44:27 carbon systemd[1]: Started Cockpit Web Service http-redirect instance.
Oct 03 21:44:27 carbon audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=cockpit-wsinstance-http-redirect comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? termina>
Oct 03 21:44:28 carbon audit[14271]: USER_AUTH pid=14271 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:cockpit_session_t:s0 msg='op=PAM:authentication grantors=pam_usertype,pam_localuser,pam_unix acct="gio" exe="/usr/libexe>
Oct 03 21:44:28 carbon audit[14271]: USER_ACCT pid=14271 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:cockpit_session_t:s0 msg='op=PAM:accounting grantors=pam_unix,pam_localuser acct="gio" exe="/usr/libexec/cockpit-session>
Oct 03 21:44:28 carbon audit[14271]: CRED_ACQ pid=14271 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:cockpit_session_t:s0 msg='op=PAM:setcred grantors=pam_localuser,pam_unix acct="gio" exe="/usr/libexec/cockpit-session" ho>
Oct 03 21:44:28 carbon audit[14271]: USER_ROLE_CHANGE pid=14271 uid=0 auid=1000 ses=6 subj=system_u:system_r:cockpit_session_t:s0 msg='pam: default-context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 selected-context=unconfined>
[...]

PR #15293, opened 2021-02-08
shell: Genuine sudo errors

mvollmer:
- [x] #15666 
- [x] #15536
- [x] New screenshot for release notes
- [x] new test that expects the lecture
- [x] use "Switch to administrative access" as dialog title when there is a input field in it
- [x] move error to top in dialog
- [x] figure out `allow_failed_sudo_journal_messages`. Remove all calls?
- [ ] fix [race condition with reading lecture from sudo](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-centos-8-stream/log.html#234-2)

This avoids running sudo without explicit request. This means that we don't probe anymore with `sudo -vn` but instead show the actual error message from sudo.

Demo: https://www.youtube.com/watch?v=Z1Cz-TTPsUs

----
## Shell: No opportunistic sudo invocation

Cockpit will no longer try to run sudo on the very first login, and you have to explicitly turn on administrative access once for each browser profile that you use to access Cockpit. This prevents cockpit's use of sudo being noisy in the journal.

In addition, Cockpit now shows sudo's exact messages in the "Administrative access" dialog.

image

### DIFF
--- a/src/bridge/askpass.c
+++ b/src/bridge/askpass.c
@@ -142,7 +142,6 @@ main (int argc,
   gchar *user = NULL;
   gchar *cookie = NULL;
   gchar *challenge = NULL;
-  const gchar *prompt;
   gint ret = 1;
 
   static GOptionEntry entries[] = {
@@ -185,14 +184,10 @@ main (int argc,
   challenge = g_strdup_printf ("plain1:%s:", user);
   cookie = g_strdup_printf ("askpass%u%u", (unsigned int)getpid (), (unsigned int)time (NULL));
 
-  prompt = argv[1];
-  if (strstr (prompt, "[sudo] ") == prompt)
-    prompt = NULL;
-
   request = cockpit_transport_build_json ("command", "authorize",
                                           "challenge", challenge,
                                           "cookie", cookie,
-                                          "prompt", prompt,
+                                          "prompt", argv[1],
                                           NULL);
 
   /* Yes, we write to stdin which we expect to be a socketpair() */
--- a/src/bridge/cockpitpeer.c
+++ b/src/bridge/cockpitpeer.c
@@ -94,15 +94,35 @@ enum {
 
 G_DEFINE_TYPE (CockpitPeer, cockpit_peer, G_TYPE_OBJECT);
 
+static gchar *
+startup_take_stderr (CockpitPeer *self)
+{
+  if (!self->other)
+    return NULL;
+
+  CockpitPipe *pipe = cockpit_pipe_transport_get_pipe (COCKPIT_PIPE_TRANSPORT (self->other));
+  return cockpit_pipe_take_stderr_as_utf8 (pipe);
+}
+
 static void
 startup_done (CockpitPeer *self,
               const gchar *problem)
 {
+  gchar *stderr = startup_take_stderr (self);
+
+  if (self->other)
+    {
+      CockpitPipe *pipe = cockpit_pipe_transport_get_pipe (COCKPIT_PIPE_TRANSPORT (self->other));
+      cockpit_pipe_stop_stderr_capture (pipe);
+    }
+
   if (self->startup_done_function)
     {
-      self->startup_done_function (problem, self->startup_done_data);
+      self->startup_done_function (problem, stderr, self->startup_done_data);
       self->startup_done_function = NULL;
     }
+
+  g_free (stderr);
 }
 
 static void
@@ -280,16 +300,16 @@ on_other_control (CockpitTransport *transport,
         g_source_remove (self->authorize_values_timeout);
       self->authorize_values_timeout = g_timeout_add (2*60*1000, cockpit_peer_delete_authorize_values, self);
 
-      startup_done (self, problem);
-
       if (problem)
         {
+          startup_done (self, problem);
           cockpit_transport_close (transport, problem);
         }
       else
         {
           g_debug ("%s: received init message from peer bridge", self->name);
           self->inited = TRUE;
+          startup_done (self, NULL);
 
           if (!self->last_init)
             {
@@ -366,12 +386,15 @@ on_other_control (CockpitTransport *transport,
             {
               gchar *user_hex;
               gchar *user;
+              gchar *stderr;
               self->startup_auth_cookie = g_strdup (cookie);
               cockpit_authorize_subject (challenge, &user_hex);
               user = cockpit_hex_decode (user_hex, -1, NULL);
-              cockpit_router_prompt (self->router, user, prompt, on_answer, self);
+              stderr = startup_take_stderr (self);
+              cockpit_router_prompt (self->router, user, prompt, stderr, on_answer, self);
               g_free (user);
               g_free (user_hex);
+              g_free (stderr);
             }
         }
 
@@ -540,6 +563,8 @@ on_other_closed (CockpitTransport *transport,
         }
     }
 
+  startup_done (self, problem ? problem : startup_problem);
+
   g_signal_handler_disconnect (self->other, self->other_closed);
   g_signal_handler_disconnect (self->other, self->other_recv);
   g_signal_handler_disconnect (self->other, self->other_control);
@@ -547,8 +572,6 @@ on_other_closed (CockpitTransport *transport,
   self->other = NULL;
 
   self->closed = TRUE;
-
-  startup_done (self, problem ? problem : startup_problem);
 
   /* Handle any remaining open channels */
   channels = g_hash_table_get_values (self->channels);
@@ -842,7 +865,8 @@ spawn_setup (gpointer data)
 }
 
 static CockpitPipe *
-spawn_process_for_config (CockpitPeer *self)
+spawn_process_for_config (CockpitPeer *self,
+                          gboolean capture_stderr)
 {
   const gchar *default_argv[] = { "/bin/false", NULL };
   CockpitPipe *pipe = NULL;
@@ -853,6 +877,7 @@ spawn_process_for_config (CockpitPeer *self)
   GError *error = NULL;
   GPid pid = 0;
   int fds[2];
+  int stderr_fd = -1;
 
   if (socketpair (PF_LOCAL, SOCK_STREAM, 0, fds) < 0)
     {
@@ -872,7 +897,7 @@ spawn_process_for_config (CockpitPeer *self)
       g_spawn_async_with_pipes (directory, (gchar **)argv, env,
                                 G_SPAWN_DO_NOT_REAP_CHILD | G_SPAWN_SEARCH_PATH,
                                 spawn_setup, GINT_TO_POINTER (fds[0]),
-                                &pid, NULL, NULL, NULL, &error);
+                                &pid, NULL, NULL, capture_stderr ? &stderr_fd : NULL, &error);
 
       if (error)
         {
@@ -894,6 +919,7 @@ spawn_process_for_config (CockpitPeer *self)
                                "name", self->name,
                                "in-fd", fds[1],
                                "out-fd", fds[1],
+                               "err-fd", stderr_fd,
                                "pid", pid,
                                NULL);
           fds[1] = -1;
@@ -1073,7 +1099,7 @@ cockpit_peer_ensure_with_done (CockpitPeer *self,
       self->startup_done_function = done_function;
       self->startup_done_data = done_data;
 
-      pipe = spawn_process_for_config (self);
+      pipe = spawn_process_for_config (self, done_function != NULL);
       if (!pipe)
         {
           self->closed = TRUE;
@@ -1091,7 +1117,7 @@ cockpit_peer_ensure_with_done (CockpitPeer *self,
   else
     {
       if (done_function)
-        done_function (NULL, done_data);
+        done_function (NULL, NULL, done_data);
     }
 
   return self->other;
--- a/src/common/cockpitpipe.c
+++ b/src/common/cockpitpipe.c
@@ -23,6 +23,7 @@
 
 #include "cockpitflow.h"
 #include "cockpitunixfd.h"
+#include "cockpitunicode.h"
 
 #include <glib-unix.h>
 
@@ -32,6 +33,7 @@
 
 #include <dirent.h>
 #include <errno.h>
+#include <poll.h>
 #include <pty.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -104,6 +106,7 @@ typedef struct {
   gboolean err_done;
   GSource *err_source;
   GByteArray *err_buffer;
+  gboolean err_forward_to_log;
 
   gboolean is_user_fd;
 
@@ -372,6 +375,18 @@ dispatch_input (gint fd,
   return TRUE;
 }
 
+static void
+forward_error (CockpitPipe *self)
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  if (priv->err_buffer->len > 0)
+    {
+      g_warning ("%s: unexpected stderr output: %.*s", priv->name, priv->err_buffer->len, priv->err_buffer->data);
+      g_byte_array_set_size (priv->err_buffer, 0);
+    }
+}
+
 static gboolean
 dispatch_error (gint fd,
                 GIOCondition cond,
@@ -405,11 +420,18 @@ dispatch_error (gint fd,
               close_immediately (self, "internal-error");
               return FALSE;
             }
+
+          if (priv->err_forward_to_log)
+            forward_error (self);
+
           return TRUE;
         }
     }
 
   g_byte_array_set_size (priv->err_buffer, len + ret);
+
+  if (priv->err_forward_to_log)
+    forward_error (self);
 
   if (ret == 0)
     {
@@ -425,6 +447,22 @@ dispatch_error (gint fd,
 
   g_object_unref (self);
   return TRUE;
+}
+
+static gboolean
+fd_readable (int fd)
+{
+  struct pollfd pfd = { .fd = fd, .events = POLLIN };
+  return poll (&pfd, 1, 0) > 0 && (pfd.revents & POLLIN);
+}
+
+static void
+drain_error (CockpitPipe *self)
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  while (priv->err_source && fd_readable (priv->err_fd))
+    dispatch_error (priv->err_fd, G_IO_IN, self);
 }
 
 static void
@@ -1503,6 +1541,50 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
 
   g_return_val_if_fail (COCKPIT_IS_PIPE (self), NULL);
   return priv->err_buffer;
+}
+
+gchar *
+cockpit_pipe_take_stderr_as_utf8 (CockpitPipe *self)
+{
+  GByteArray *buffer;
+  GBytes *clean;
+  gchar *data;
+  gsize length;
+
+  drain_error (self);
+
+  buffer = cockpit_pipe_get_stderr (self);
+  if (!buffer)
+    return NULL;
+
+  /* A little more complicated to avoid big copies */
+  g_byte_array_ref (buffer);
+  g_byte_array_append (buffer, (guint8 *)"x", 1); /* place holder for null terminate */
+
+  {
+    g_autoptr(GBytes) bytes = g_byte_array_free_to_bytes (buffer);
+    clean = cockpit_unicode_force_utf8 (bytes);
+  }
+
+  data = g_bytes_unref_to_data (clean, &length);
+
+  /* Fill in null terminate, for x above */
+  g_assert (length > 0);
+  data[length - 1] = '\0';
+
+  return data;
+}
+
+void
+cockpit_pipe_stop_stderr_capture (CockpitPipe *self)
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  if (priv->err_buffer)
+    {
+      priv->err_forward_to_log = TRUE;
+      forward_error (self);
+    }
 }
 
 /**
--- a/src/ws/cockpitwebservice.c
+++ b/src/ws/cockpitwebservice.c
@@ -552,16 +552,6 @@ process_transport_authorize (CockpitWebService *self,
         {
           g_info ("received \"authorize\" %s \"challenge\", but for wrong user", type);
         }
-      else if (g_str_equal ("plain1", type) && g_str_equal (cockpit_creds_get_superuser (self->creds), "none"))
-        {
-          // Older versions of the bridge will send "plain1"
-          // challenges when trying to start a privileged sub-bridge,
-          // but not for anything else.  We use that fact to prevent
-          // them from starting the privileged sub-bridge when the
-          // user has selected that option on the login screen.
-
-          g_info ("received \"plain1\" authorize challenge, but superuser is \"none\"");
-        }
       else
         {
           password = g_bytes_get_data (data, NULL);
@@ -618,7 +608,6 @@ process_transport_init (CockpitWebService *self,
   JsonObject *object;
   JsonObject *capabilities;
   gboolean explicit_superuser_capability = FALSE;
-  const gchar *superuser;
   GBytes *payload;
   gint64 version;
 
@@ -657,19 +646,22 @@ process_transport_init (CockpitWebService *self,
       json_object_set_int_member (object, "version", 1);
       json_object_set_string_member (object, "host", "localhost");
 
-      superuser = getenv("COCKPIT_SUPERUSER") ?: cockpit_creds_get_superuser (self->creds);
-      if (superuser && *superuser && !g_str_equal (superuser, "none"))
+      if (explicit_superuser_capability)
         {
-          JsonObject *superuser_options;
+          const gchar *superuser = getenv("COCKPIT_SUPERUSER") ?: cockpit_creds_get_superuser (self->creds);
+          if (superuser && *superuser && !g_str_equal (superuser, "none"))
+            {
+              JsonObject *superuser_options;
 
-          superuser_options = json_object_new ();
-          json_object_set_string_member (superuser_options, "id", superuser);
-          json_object_set_object_member (object, "superuser", superuser_options);
-        }
-      else
-        {
-          json_object_set_boolean_member (object, "superuser", FALSE);
-          cockpit_creds_poison (self->creds);
+              superuser_options = json_object_new ();
+              json_object_set_string_member (superuser_options, "id", superuser);
+              json_object_set_object_member (object, "superuser", superuser_options);
+            }
+          else
+            {
+              json_object_set_boolean_member (object, "superuser", FALSE);
+              cockpit_creds_poison (self->creds);
+            }
         }
 
       payload = cockpit_json_write_bytes (object);
--- a/src/ws/test-auth.c
+++ b/src/ws/test-auth.c
@@ -152,10 +152,21 @@ test_application (Test *test,
   g_clear_pointer (&application, g_free);
 }
 
+typedef struct {
+  const gchar *superuser_mode;
+  gboolean expect_stored_password;
+} UserpassFixture;
+
+static const UserpassFixture fixture_superuser_any = {
+  .superuser_mode = "any",
+  .expect_stored_password = TRUE
+};
+
 static void
 test_userpass_cookie_check (Test *test,
                             gconstpointer data)
 {
+  const UserpassFixture *fix = data;
   GAsyncResult *result = NULL;
   CockpitWebService *service;
   CockpitWebService *prev_service;
@@ -166,6 +177,8 @@ test_userpass_cookie_check (Test *test,
   GHashTable *headers;
 
   headers = mock_auth_basic_header ("me", "this is the password");
+  if (fix && fix->superuser_mode)
+    g_hash_table_insert (headers, g_strdup ("X-Superuser"), g_strdup (fix->superuser_mode));
   cockpit_auth_login_async (test->auth, "/cockpit/", NULL, headers, on_ready_get_result, &result);
   g_hash_table_unref (headers);
 
@@ -187,7 +200,10 @@ test_userpass_cookie_check (Test *test,
   creds = cockpit_web_service_get_creds (service);
   g_assert_cmpstr ("me", ==, cockpit_creds_get_user (creds));
   g_assert_cmpstr ("cockpit", ==, cockpit_creds_get_application (creds));
-  g_assert_cmpstr ("this is the password", ==, g_bytes_get_data (cockpit_creds_get_password (creds), NULL));
+  if (fix && fix->expect_stored_password)
+    g_assert_cmpstr ("this is the password", ==, g_bytes_get_data (cockpit_creds_get_password (creds), NULL));
+  else
+    g_assert_null (cockpit_creds_get_password (creds));
 
   prev_service = service;
   g_object_unref (service);
@@ -205,7 +221,10 @@ test_userpass_cookie_check (Test *test,
   g_assert (prev_creds == creds);
 
   g_assert_cmpstr ("me", ==, cockpit_creds_get_user (creds));
-  g_assert_cmpstr ("this is the password", ==, g_bytes_get_data (cockpit_creds_get_password (creds), NULL));
+  if (fix && fix->expect_stored_password)
+    g_assert_cmpstr ("this is the password", ==, g_bytes_get_data (cockpit_creds_get_password (creds), NULL));
+  else
+    g_assert_null (cockpit_creds_get_password (creds));
 
   g_hash_table_destroy (headers);
   g_object_unref (service);
@@ -459,6 +478,7 @@ typedef struct {
   const gchar *user;
   const gchar *application;
   const gchar *cookie_name;
+  gboolean     expect_stored_password;
 } SuccessFixture;
 
 static void
@@ -549,7 +569,6 @@ test_custom_success (Test *test,
 
   headers = web_socket_util_new_headers ();
   g_hash_table_insert (headers, g_strdup ("Authorization"), g_strdup (fix->header));
-  g_hash_table_insert (headers, g_strdup ("X-Superuser"), g_strdup ("none"));
   cockpit_auth_login_async (test->auth, path, NULL, headers, on_ready_get_result, &result);
   g_hash_table_unref (headers);
 
@@ -568,7 +587,10 @@ test_custom_success (Test *test,
   service = cockpit_auth_check_cookie (test->auth, path, headers);
   creds = cockpit_web_service_get_creds (service);
   g_assert_cmpstr (application, ==, cockpit_creds_get_application (creds));
-  g_assert_null (cockpit_creds_get_password (creds));
+  if (fix->expect_stored_password)
+    g_assert_cmpstr ("this is the machine password", ==, g_bytes_get_data (cockpit_creds_get_password (creds), NULL));
+  else
+    g_assert_null (cockpit_creds_get_password (creds));
 
   login_data = cockpit_creds_get_login_data (creds);
   if (fix->data)
@@ -600,7 +622,8 @@ static const SuccessFixture fixture_ssh_remote_basic = {
   .path = "/cockpit+=machine",
   .user = "remote-user",
   .application = "cockpit+=machine",
-  .cookie_name = "machine-cockpit+machine"
+  .cookie_name = "machine-cockpit+machine",
+  .expect_stored_password = TRUE
 };
 
 static const SuccessFixture fixture_ssh_no_data = {
@@ -1172,6 +1195,7 @@ main (int argc,
 
   g_test_add ("/auth/application", Test, NULL, NULL, test_application, NULL);
   g_test_add ("/auth/userpass-header-check", Test, NULL, setup, test_userpass_cookie_check, teardown);
+  g_test_add ("/auth/userpass-store-check", Test, &fixture_superuser_any, setup, test_userpass_cookie_check, teardown);
   g_test_add ("/auth/userpass-bad", Test, NULL, setup, test_userpass_bad, teardown);
   g_test_add ("/auth/userpass-emptypass", Test, NULL, setup, test_userpass_emptypass, teardown);
   g_test_add ("/auth/headers-bad", Test, NULL, setup, test_headers_bad, teardown);
--- a/src/ws/test-handlers.c
+++ b/src/ws/test-handlers.c
@@ -304,7 +304,7 @@ test_login_accept (Test *test,
   g_assert (service != NULL);
   creds = cockpit_web_service_get_creds (service);
   g_assert_cmpstr (cockpit_creds_get_user (creds), ==, "me");
-  g_assert_cmpstr (g_bytes_get_data (cockpit_creds_get_password (creds), NULL), ==, PASSWORD);
+  g_assert_null (cockpit_creds_get_password (creds));
 
   token = cockpit_creds_get_csrf_token (creds);
   g_assert (strstr (output, token));
--- a/test/verify/check-shell-keys
+++ b/test/verify/check-shell-keys
@@ -36,7 +36,6 @@ class TestKeys(MachineCase):
         # Create a user without any role
         m.execute("useradd user -s /bin/bash -m -c 'User' || true")
         m.execute("echo user:foobar | chpasswd")
-        self.allow_failed_sudo_journal_messages()
 
         m.start_cockpit()
 
--- a/test/verify/check-shell-keys
+++ b/test/verify/check-shell-keys
@@ -106,7 +106,26 @@ class TestKeys(MachineCase):
         # Log in as admin
         login("admin", "foobar", "Administrator")
         b.go("#/user")
-        b.wait_in_text("#account-authorized-keys-list li.pf-c-data-list__item:first-child", "no authorized public keys")
+
+        if "ubuntu" not in m.image and "debian" not in m.image:
+            b.wait_in_text("#account-authorized-keys-list li.pf-c-data-list__item:first-child",
+                           "You do not have permission to view the authorized public keys for this account.")
+
+        # Get admin privs
+        b.leave_page()
+        b.wait_text("#super-user-indicator", "Limited access")
+        b.click("#super-user-indicator button")
+        b.wait_in_text(".pf-c-modal-box:contains('Administrative access')", "Please authenticate")
+        b.set_input_text(".pf-c-modal-box:contains('Administrative access') input", "foobar")
+        b.click(".pf-c-modal-box button:contains('Authenticate')")
+        # users page reloads on superuser change
+        b.expect_load_frame("cockpit1:localhost/users")
+        b.wait_not_present(".pf-c-modal-box:contains('Administrative access')")
+        b.wait_text("#super-user-indicator", "Administrative access")
+        b.enter_page("/users")
+
+        b.wait_in_text("#account-authorized-keys-list li.pf-c-data-list__item:first-child",
+                       "no authorized public keys")
         b.wait_js_func("ph_count_check", "#account-authorized-keys-list li.pf-c-data-list__item", 1)
 
         # Adding keys sets permissions properly
--- a/test/verify/check-shell-keys
+++ b/test/verify/check-shell-keys
@@ -115,12 +115,12 @@ class TestKeys(MachineCase):
         b.leave_page()
         b.wait_text("#super-user-indicator", "Limited access")
         b.click("#super-user-indicator button")
-        b.wait_in_text(".pf-c-modal-box:contains('Administrative access')", "Please authenticate")
-        b.set_input_text(".pf-c-modal-box:contains('Administrative access') input", "foobar")
+        b.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", "Password for admin:")
+        b.set_input_text(".pf-c-modal-box:contains('Switch to administrative access') input", "foobar")
         b.click(".pf-c-modal-box button:contains('Authenticate')")
         # users page reloads on superuser change
         b.expect_load_frame("cockpit1:localhost/users")
-        b.wait_not_present(".pf-c-modal-box:contains('Administrative access')")
+        b.wait_not_present(".pf-c-modal-box:contains('Switch to administrative access')")
         b.wait_text("#super-user-indicator", "Administrative access")
         b.enter_page("/users")
 
--- a/test/verify/check-system-shutdown-restart
+++ b/test/verify/check-system-shutdown-restart
@@ -43,17 +43,9 @@ class TestShutdownRestart(MachineCase):
         m.execute("useradd user -s /bin/bash -c 'User' || true")
         m.execute("echo user:foobar | chpasswd")
         m.execute("echo 'user ALL=(ALL) ALL' > /etc/sudoers.d/user")
+        self.password = "foobar"
 
-        m.start_cockpit()
-
-        # login
-        b.open("/system")
-        b.wait_visible("#login")
-        b.set_val("#login-user-input", "user")
-        b.set_val("#login-password-input", "foobar")
-        b.click('#login-button')
-        b.expect_load()
-        b.enter_page("/system")
+        self.login_and_go("/system", user="user")
 
         # shutdown button should be enabled and working
         b.click("#reboot-button")
--- a/pkg/shell/superuser.jsx
+++ b/pkg/shell/superuser.jsx
@@ -27,12 +27,6 @@ import "form-layout.scss";
 
 const _ = cockpit.gettext;
 
-export function can_do_sudo(host) {
-    return cockpit.spawn(["sudo", "-v", "-n"], { err: "out", environ: ["LC_ALL=C"], host: host })
-            .then(() => true,
-                  (err, out) => !(err.exit_status == 1 && out.match("Sorry, user .+ may not run sudo on .+\\.")));
-}
-
 class UnlockDialog extends React.Component {
     render() {
         const { state } = this.props;
@@ -221,15 +215,7 @@ export class SuperuserDialogs extends React.Component {
 
     unlock(error) {
         this.superuser.Stop().always(() => {
-            can_do_sudo(this.props.host).then(can_do => {
-                if (!can_do)
-                    this.set_unlock_state({
-                        message: _("You can not gain administrative access."),
-                        cancel: () => this.set_unlock_state({ closed: true })
-                    });
-                else
-                    this.start("sudo", error);
-            });
+            this.start("sudo", error);
         });
     }
 
@@ -248,13 +234,13 @@ export class SuperuserDialogs extends React.Component {
 
         let did_prompt = false;
 
-        const onprompt = (event, message, prompt, def, echo) => {
+        const onprompt = (event, message, prompt, def, echo, error) => {
             did_prompt = true;
             const p = { message: message, prompt: prompt, value: def, echo: echo };
             this.set_unlock_state({
                 prompt: p,
 
-                error: this.state.unlock_dialog_state.error,
+                error: error || this.state.unlock_dialog_state.error,
                 change: val => {
                     p.value = val;
                     this.update_unlock_state({ prompt: p });
@@ -290,14 +276,11 @@ export class SuperuserDialogs extends React.Component {
                     console.warn(err);
                     this.superuser.removeEventListener("Prompt", onprompt);
                     if (err && err.message != "cancelled") {
-                        if (did_prompt)
-                            this.unlock(_("This didn't work, please try again"));
-                        else
-                            this.set_unlock_state({
-                                message: _("Something went wrong"),
-                                error: err.toString(),
-                                cancel: () => this.set_unlock_state({ closed: true })
-                            });
+                        this.set_unlock_state({
+                            message: _("This did not work."),
+                            error: err.toString(),
+                            cancel: () => this.set_unlock_state({ closed: true })
+                        });
                     } else
                         this.set_unlock_state({ closed: true });
                 });
--- a/pkg/shell/superuser.jsx
+++ b/pkg/shell/superuser.jsx
@@ -19,7 +19,7 @@
 
 import cockpit from "cockpit";
 import React from "react";
-import { Button, FormSelect, FormSelectOption, Modal } from '@patternfly/react-core';
+import { Alert, Button, Modal } from '@patternfly/react-core';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import { host_superuser_storage_key } from './machines/machines';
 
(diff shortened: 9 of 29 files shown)


mvollmer:
(I'll make a video for this.)

src/bridge/cockpitpeer.c
```diff
@@ -279,16 +335,18 @@ on_other_control (CockpitTransport *transport,
         g_source_remove (self->authorize_values_timeout);
       self->authorize_values_timeout = g_timeout_add (2*60*1000, cockpit_peer_delete_authorize_values, self);
 
-      startup_done (self, problem);
-
       if (problem)
         {
           cockpit_transport_close (transport, problem);
+          startup_drain_stderr (self);
```
mvollmer, line 341: This doesn't drain everything... we should call g_io_channel_read_to_end, I guess.

martinpitt:
Still test failures and there's an outstanding TODO above. Marking as draft.

src/bridge/cockpitpeer.c
mvollmer: I put a loop around startup_stderr_drain.

martinpitt:
Thanks for the video! The thing that I find a bit irritating is that there is now no error message when you mis-type the password? In the current version we see "Sorry, try again", and that's also what sudo says on the command line.

There is also a few "XXX/TODO" left in the code, I suppose these should be cleaned up before landing?

martinpitt:
For an unprivileged user and a private browser window (i.e. no previous cookie), I now get exactly *one* sudo error:

     sudo[14561]:      joe : user NOT in sudoers ; PWD=/run/user/1002 ; USER=root ; COMMAND=/usr/bin/cockpit-bridge --privileged

The second time I log in, the browser cookie remembers non-admin, and does not try sudo any more.

That's about as much as we can do with keeping the current behaviour, I think. We *could* not do the initial "opportunistic sudo" attempt; that would be more like sudo works on the command line then, but break a long-standing behaviour. I am mildly in favor of actually doing this, as it's only a small inconvenience and the browser cookie remembers the state from the second time on. But this should be a follow-up.

So, :+1: from me for a behavioural aspect. @allisonkarlitskaya , do you have some time to look at the C parts?

mvollmer:
> In the current version we see "Sorry, try again", and that's also what sudo says on the command line.

We don't see "sorry, try again" after each failed attempt, we only see our own "This didn't work, please try again" message after three failed attempts.  (The "Sorry, try again" appears in the journal, but it is quite useless there, of course.)

However! Now we have the machinery here to include all the stderr output in the dialog that has appeared before the prompt. I think that's a splendid idea! :-)

mvollmer:
>  I am mildly in favor of actually [defaulting to limited access], as it's only a small inconvenience and the browser cookie remembers the state from the second time on. But this should be a follow-up.

I agree, I am also in favor of this now.

martinpitt:
Perfect, thanks @mvollmer! I'm fine with this as-is, we can ponder the "don't do initial sudo" thing separately. I'd still prefer @allisonkarlitskaya to review the glib-y part (mostly the first commit).

allisonkarlitskaya (changes requested):
I'm not 100% sure, but might it be a whole lot easier to use the `"err-fd"` property on `CockpitPipe` and `cockpit_pipe_get_stderr()`?

mvollmer:
> I'm not 100% sure, but might it be a whole lot easier to use the `"err-fd"` property on `CockpitPipe` and `cockpit_pipe_get_stderr()`?

Yes!  That should work, with a few additions to CockpitPipe.  I think we want stderr to be logged to the journal after initialization is done. Peer bridges usually run for the whole session and we don't wanto to just collect all their stderr into memory.  Also, we want to clear the accumulated messages when prompting.

src/bridge/cockpitpeer.c
```diff
@@ -95,15 +96,50 @@ G_DEFINE_TYPE (CockpitPeer, cockpit_peer, G_TYPE_OBJECT);
...
+  CockpitPipe *pipe;
+  GByteArray *buffer;
+  GBytes *bytes;
+  GBytes *clean;
+  gchar *data;
+  gsize length;
+
+  pipe = cockpit_pipe_transport_get_pipe (COCKPIT_PIPE_TRANSPORT (self->other));
+
+  buffer = cockpit_pipe_get_stderr (pipe);
```
mvollmer, line 121: This is a copy/paste, let's factor it out into a `cockpit_pipe_take_stderr_as_utf8` function.

src/bridge/cockpitpeer.c
```diff
@@ -279,16 +287,20 @@ on_other_control (CockpitTransport *transport,
...
+          gchar *stderr = startup_take_stderr (self);
+          startup_done (self, problem, stderr);
+          g_free (stderr);
           cockpit_transport_close (transport, problem);
         }
       else
         {
           g_debug ("%s: received init message from peer bridge", self->name);
           self->inited = TRUE;
+          CockpitPipe *pipe = cockpit_pipe_transport_get_pipe (COCKPIT_PIPE_TRANSPORT (transport));
```
allisonkarlitskaya, line 301: It seems like we might drop some stderr output this way.  Shouldn't we be taking the existing buffer before calling `stop()` and forwarding it?  We could build that functionality into `stop()` itself in fact, since you already add code in cockpitpipe for forwarding newly-arrived data *after* stop() is called, so you could then share....

src/common/cockpitpipe.c
```diff
@@ -382,6 +383,15 @@ dispatch_error (gint fd,
   gssize ret = 0;
   gsize len;
 
+  if (priv->err_buffer == NULL)
+    {
+      char buf[512];
+      ret = read (priv->err_fd, buf, 512);
+      if (ret > 0)
+        write (2, buf, ret);
```
allisonkarlitskaya, line 391: partial write?  blocking?  EINTR?

i guess you'll need to try a bit harder here....

it might even make sense to keep the buffer around and drain it out into fd 2 as it becomes writable...

or maybe it's possible to rethink if there's another way to do this

allisonkarlitskaya (changes requested):

src/common/cockpitpipe.c
mvollmer: I'd say blocking is fine here. We don't go to the trouble of avoiding to block when writing log messages anywhere else, no?

src/bridge/cockpitpeer.c
mvollmer: Yes, good point.

mvollmer: I have changed it so that we keep reading into the buffer (which ensures proper maintaining of the pipe's state). If a flag is set, the buffer is just dumped out to stderr again.

mvollmer:
`check-shell-multi-machine TestMultiMachine.testSshKeySetup` seems to be failing a lot here, for unrelated reasons (I think).  I'll use this opportunity to add some debugging output.

mvollmer:
> `check-shell-multi-machine TestMultiMachine.testSshKeySetup` seems to be failing a lot here, for unrelated reasons (I think). I'll use this opportunity to add some debugging output.

Happens in #15446.

src/common/cockpitpipe.c
```diff
@@ -1505,6 +1529,47 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
...
+}
+
+void
+cockpit_pipe_stop_stderr_capture (CockpitPipe *self)
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  if (priv->err_buffer)
+    {
+      priv->err_forward_to_log = TRUE;
```
allisonkarlitskaya, line 1568: This is a lot nicer.  Thanks :)

src/common/cockpitpipe.c
```diff
@@ -1505,6 +1529,47 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
   return priv->err_buffer;
 }
 
+gchar *
+cockpit_pipe_take_stderr_as_utf8 (CockpitPipe *self)
```
allisonkarlitskaya, line 1533: There's no `take` left here anymore, as far as I can tell.  This is just some extra UTF8 stuff...

src/common/cockpitpipe.c
```diff
@@ -1505,6 +1529,47 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
...
+  gchar *data;
+  gsize length;
+
+  buffer = cockpit_pipe_get_stderr (self);
+  if (!buffer)
+    return NULL;
+
+  /* A little more complicated to avoid big copies */
+  g_byte_array_ref (buffer);
+  g_byte_array_append (buffer, (guint8 *)"x", 1); /* place holder for null terminate */
```
allisonkarlitskaya, line 1547: ...you're modifying the internal buffer here

src/common/cockpitpipe.c
```diff
@@ -1505,6 +1529,47 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
...
+  bytes = g_byte_array_free_to_bytes (buffer);
```
allisonkarlitskaya, line 1548: ...and because of the extra ref, this will indeed take a copy.

src/common/cockpitpipe.c
```diff
@@ -1505,6 +1529,47 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
   return priv->err_buffer;
 }
 
+gchar *
+cockpit_pipe_take_stderr_as_utf8 (CockpitPipe *self)
+{
+  GByteArray *buffer;
```
allisonkarlitskaya, line 1535: New code should probably try to use `g_autoptr()` to the extent possible for stuff like this...

src/common/cockpitpipe.c
```diff
@@ -1505,6 +1529,47 @@ cockpit_pipe_get_stderr (CockpitPipe *self)
...
+  /* A little more complicated to avoid big copies */
+  g_byte_array_ref (buffer);
+  g_byte_array_append (buffer, (guint8 *)"x", 1); /* place holder for null terminate */
+  bytes = g_byte_array_free_to_bytes (buffer);
+  clean = cockpit_unicode_force_utf8 (bytes);
+  g_bytes_unref (bytes);
+
+  data = g_bytes_unref_to_data (clean, &length);
+
+  /* Fill in null terminate, for x above */
```
allisonkarlitskaya, line 1554: Probably this would all be a lot cleaner if you just passed the buffer, unmodified, into the force function, got the byte array back out, and added the nul after...

Are we expecting such large strings here?

src/common/cockpitpipe.c
```diff
@@ -372,6 +374,24 @@ dispatch_input (gint fd,
...
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  guint8 *data = priv->err_buffer->data;
+  guint len = priv->err_buffer->len;
+  while (len > 0)
+    {
+      ssize_t n = TEMP_FAILURE_RETRY (write (2, data, len));
+      if (n <= 0)
+        break;
```
allisonkarlitskaya, line 388: I guess at the very least, if we get an unhandled error then we should probably stop attempting to continue forwarding...

src/common/cockpitpipe.c
```diff
@@ -372,6 +374,24 @@ dispatch_input (gint fd,
...
+static void
+forward_error (CockpitPipe *self)
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  guint8 *data = priv->err_buffer->data;
+  guint len = priv->err_buffer->len;
+  while (len > 0)
+    {
+      ssize_t n = TEMP_FAILURE_RETRY (write (2, data, len));
```
allisonkarlitskaya, line 386: `STDERR_FILENO` would be nicer.

src/common/cockpitpipe.c
```diff
@@ -372,6 +374,24 @@ dispatch_input (gint fd,
...
+      if (n <= 0)
+        break;
+      len -= n;
```
allisonkarlitskaya, line 389: `g_byte_array_remove_range()` probably gives you a nicer way to implement this function

allisonkarlitskaya (changes requested):

src/common/cockpitpipe.c
mvollmer: It is supposed to reset the buffer to empty.  Does it not?  This is old code from cockpitpipechannel.c, read_stderr. I have factored it out into this common function.

src/common/cockpitpipe.c
mvollmer: `cockpit_unicode_force_utf8` returns a GBytes object which doesn't offer any way to append to it, no?  We could append a `\0` instead of a "x" and change cockpit_unicode_force_utf8 to accept it.

src/common/cockpitpipe.c
mvollmer: We might even kill the whole bridge, no?  Getting an error here means that stderr is hosed, and we shouldn't probably not continue at all with the session, no?  What does `g_log` do if there is an error writing to stderr?  Actually, we could just use `g_log` to handle all of this for us, no?

src/common/cockpitpipe.c
mvollmer: Fixed, thanks!

src/common/cockpitpipe.c
mvollmer: That would mean some copies, no?

Anyway, what about just using `fputs` and ignoring errors? :-) That's what glib does by default, no?

src/common/cockpitpipe.c
mvollmer: I don't think so... this ultimately calls `g_array_free (buffer, FALSE)`, which will always take the actual memory away from `buffer` but might preserve the `buffer` wrapper itself.  The `g_byte_array_ref` above makes sure that the wrapper stays intact.

src/common/cockpitpipe.c
mvollmer: I don't think this a problem, or is it?

src/common/cockpitpipe.c
mvollmer: I think we can only use it for `bytes`.  Done.

src/common/cockpitpipe.c
mvollmer: I have pushed a variant that uses fwrite.

src/common/cockpitpipe.c
allisonkarlitskaya: Ah.  You're right.  This also invalidates my above comments about how this isn't a `take` anymore.  It clearly is.

I read the docs incorrectly.

src/common/cockpitpipe.c
allisonkarlitskaya: It would only mean copies in case of partial writes, which we're not expecting.

src/common/cockpitpipe.c
allisonkarlitskaya: Even though it's not included in the length of the `GBytes`, the string from `cockpit_unicode_force_utf8()` comes from `GString` and is therefore already always nul terminated.

Maybe add a doc comment to the top of `cockpit_unicode_force_utf8()` mentioning this fact, and remove all the extra hacks here.

src/common/cockpitpipe.c
```diff
@@ -372,6 +374,15 @@ dispatch_input (gint fd,
   return TRUE;
 }
 
+static void
+forward_error (CockpitPipe *self)
+{
+  CockpitPipePrivate *priv = cockpit_pipe_get_instance_private (self);
+
+  fwrite (priv->err_buffer->data, priv->err_buffer->len, 1, stderr);
```
allisonkarlitskaya, line 382: uh... okay, whatever :rofl:

allisonkarlitskaya (changes requested):

src/common/cockpitpipe.c
mvollmer: > Even though it's not included in the length of the GBytes, the string from cockpit_unicode_force_utf8() comes from GString and is therefore already always nul terminated.

Not always.  When `g_utf8_validate` is happy with everything, the function returns the input GBytes unchanged (but with one more ref).

allisonkarlitskaya (dismissed):
I think we've bikeshedded enough.  It could maybe be more beautiful, but not substantially so...

martinpitt:
Tests are very red, and [rhel-8-4](https://logs.cockpit-project.org/logs/pull-15293-20210303-082642-a1483e62-rhel-8-4/log.html#218-1) looks horrible. But these may have been a victim of some bots corruption this morning, I restart them once.

allisonkarlitskaya:
...but let's wait until after the release.

src/common/cockpitpipe.c
mvollmer: I have pushed a FIXUP that works directly with nul characters instead of using the "x" as a placeholder.

mvollmer:
Upps, I accidentally pushed more stuff after the approval.  I have taken that away now, and we are back at a1483e6 which was approved by @allisonkarlitskaya.

martinpitt:
The test failures look as bad as before, and this time it isn't the bots...

mvollmer:
> The test failures look as bad as before, and this time it isn't the bots...

This looks like something in the infrastructure, and not the individual tests... For example inhttps://logs.cockpit-project.org/logs/pull-15293-20210303-103328-a1483e62-centos-8-stream/log.html all the second tries of changed tests fail with

```
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 970, in tearDown
    shutil.rmtree(self.tmpdir)
  File "/usr/lib64/python3.9/shutil.py", line 709, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/usr/lib64/python3.9/shutil.py", line 707, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpkuzgx62k'
```

Most if not all other failures are like this as well.

https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15293-20210303-103333-a1483e62-fedora-33/log.html#127-2 had a regular flake, but also just stopped in the middle.

I'll tall our pilot to have a look. :-)

mvollmer:
There is a suspicious JavaScript error:

RuntimeError: {"exceptionId":0,"text":"TypeError: can't access property \"default_host\", v is undefined","lineNumber":69,"columnNumber":131569,"url":"http://127.0.0.2:9391/cockpit/$94d951ea43463a7be7207e549849e88b635676a3664f0d6cc993a02e201719f2/shell/index.js"

mvollmer:
And, argh, the "didn't immediately connect on login" flake seems to be still there, which was supposed to be fixed by #15451...

mvollmer:
> There is a suspicious JavaScript error:
> 
> RuntimeError: {"exceptionId":0,"text":"TypeError: can't access property "default_host", v is undefined","lineNumber":69,"columnNumber":131569,"url":"http://127.0.0.2:9391/cockpit/$94d951ea43463a7be7207e549849e88b635676a3664f0d6cc993a02e201719f2/shell/index.js"

#15468

martinpitt:
The `default_host` fix landed, retrying tests. @mvollmer, is this otherwise ready?

martinpitt (changes requested):
I just tested this again in a fedora-33 VM in a private browser window (thus, no cookies and such), and it does not work as intended any more. As user "admin" I immediately get administrative privs, and as non-admin user it tries to run sudo:
```
m1.cockpit.lan sudo[2239]:      joe : user NOT in sudoers ; PWD=/run/user/1003 ; USER=root ; COMMAND=/usr/bin/cockpit-bridge --privileged
```
So the part with getting "genuine sudo errors" does work nicely now. But that's not (only) what the commit message and release note promise.

FTR: I fully agree to *not* doing sudo opportunistically on the very first usage of cockpit. After enabling it once, the cookie should of course remember and directly give you admin the next time.

I am fine with either of these: (1) restrict this to the "genuine sudo errors" part again, adjust commit message/release notes, and do the "no opportunistic" in a follow-up. Or (2) do it all in one go in this PR (even though it's two independent changes).

Thanks!

mvollmer:
Let me try to switch the default here. If there is too much fallout, I'll split it out.

mvollmer:
By changing the default to "none", we have lost one scenario: Directly logging into a remote host that runs a old version of Cockpit and getting sudo.  This fails now because cockpit-ws does not remember the password with "X-Superuser: none". We can make an exception for that and restore the scenario, I think.

mvollmer:
I want to split up the commits better...

mvollmer:
> We can make an exception for that and restore the scenario, I think.

Done. @martinpitt, ptal!

src/ws/cockpitwebservice.c
```diff
@@ -668,7 +658,6 @@ process_transport_init (CockpitWebService *self,
       else
         {
           json_object_set_boolean_member (object, "superuser", FALSE);
-          cockpit_creds_poison (self->creds);
```
mvollmer, line 671: We should still poisen the creds here for a bridge with "explicit_superuser_capability".

src/ws/test-handlers.c
```diff
@@ -306,7 +306,7 @@ test_login_accept (Test *test,
   g_assert (service != NULL);
   creds = cockpit_web_service_get_creds (service);
   g_assert_cmpstr (cockpit_creds_get_user (creds), ==, "me");
-  g_assert_cmpstr (g_bytes_get_data (cockpit_creds_get_password (creds), NULL), ==, PASSWORD);
+  g_assert (cockpit_creds_get_password (creds) == NULL);
```
mvollmer, line 309: g_assert_null?

src/ws/test-auth.c
```diff
@@ -184,6 +184,67 @@ test_userpass_cookie_check (Test *test,
...
+  g_assert_cmpstr ("me", ==, cockpit_creds_get_user (creds));
+  g_assert_null (cockpit_creds_get_password (creds));
+
+  g_hash_table_destroy (headers);
+  g_object_unref (service);
+  json_object_unref (response);
+}
+
+static void
+test_userpass_store_check (Test *test,
```
mvollmer, line 216: This is almost the same as "test_userpass_cookie_check", let's use a fixture to avoid code duplication.

test/verify/check-shell-keys
```diff
@@ -105,7 +105,22 @@ class TestKeys(MachineCase):
         # Log in as admin
         login("admin", "foobar", "Administrator")
         b.go("#/user")
-        b.wait_in_text("#account-authorized-keys-list li.pf-c-data-list__item:first-child", "no authorized public keys")
+        b.wait_in_text("#account-authorized-keys-list li.pf-c-data-list__item:first-child",
+                       "You do not have permission to view the authorized public keys for this account.")
+
+        # Get amdin privs
```
martinpitt, line 111: "amdin" typo. As you need to push anyway, would be nice to fix that, as we often grep the code for user names.

test/verify/check-system-shutdown-restart
```diff
@@ -44,17 +44,9 @@ class TestShutdownRestart(MachineCase):
         m.execute("useradd user -s /bin/bash -c 'User' || true")
         m.execute("echo user:foobar | chpasswd")
         m.execute("echo 'user ALL=(ALL) ALL' > /etc/sudoers.d/user")
+        self.password = "foobar"
 
-        m.start_cockpit()
-
-        # login
-        b.open("/system")
```
martinpitt, line 51: Ugh, thanks for the cleanup. I wonder why we did that in the past?

martinpitt (changes requested):
I tested this again, and it now works most excellently, :clap: Great work!

There are a few remaining test case failures, and the things you picked up in self-review.

I wonder if the second and third commits wouldn't better get squashed. Their commit messages say pretty much the same thing.

test/verify/check-system-shutdown-restart
mvollmer: It might one of the first tests which just never got ported to the new fangled login_and_go API...

src/ws/cockpitwebservice.c
mvollmer: Now we only add the "superuser" field to the "init" message for bridges with "explicit_superuser_capability".

src/ws/test-handlers.c
mvollmer: Done.

src/ws/test-auth.c
mvollmer: Done

test/verify/check-shell-keys
mvollmer: Fixed.

allisonkarlitskaya (changes requested):
I'm going to put me down as a -1.  Here's why:

Redirecting stderr to the pipe will have the unintended side-effect that all spawned processes (ie: the bridge running under sudo) will no longer see that stderr is connected to the journal, and will end up disabling structured logging.

The forwarding we're doing is not sufficient.

mvollmer:
> Redirecting stderr to the pipe will have the unintended side-effect that all spawned processes (ie: the bridge running under sudo) will no longer see that stderr is connected to the journal, and will end up disabling structured logging.

This is the reason for the `TestCurrentMetrics.testDiskIO` failure.

mvollmer:
> Redirecting stderr to the pipe will have the unintended side-effect that all spawned processes (ie: the bridge running under sudo) will no longer see that stderr is connected to the journal, and will end up disabling structured logging.

One option is to always reconnect the bridge to the journal, explicitly.  We do this currently, but only for bridges on remote machines.  I have pushed the "WIP - connect bridge stderr to the journal always" commit to make show this concretely.

garrett (changes requested):
Is the screenshot and video up to date?

If so, it has the following issues:

1. Heading should indicate that there's an error; something more like "Problem becoming administrator"
1. Error icon is missing from the heading
2. Inline error is styled incorrectly... the whole dialog is the notification (the docs linked below show how it should look)
3. It has two content bodies... it should only have one
4. The close button should be styled like a button

Please fix.

For reference, here's what error dialogs should look like: https://www.patternfly.org/v4/components/modal/design-guidelines/#error-dialogs

The button should be either primary or secondary. Link styled buttons should never be used, except for a cancel action (and when there's an action dialog).

In this case, if there's no action to accomplish something (like "retry"), then "close" itself is the default action (as it is here), and should be a button that looks like a button.

mvollmer:
> If so, it has the following issues:

This PR does not change the dialog, just the text in the error case.  Let's address these issues separately.

mvollmer:
> Redirecting stderr to the pipe will have the unintended side-effect that all spawned processes (ie: the bridge running under sudo) will no longer see that stderr is connected to the journal, and will end up disabling structured logging.

One more option (that I hadn't thought of before): Give sudo a controlling tty.  I think it will use that instead of stdin/stdout/stderr. This would also help with #14695, if possible.

garrett:
> This PR does not change the dialog, just the text in the error case. Let's address these issues separately.

The dialog is user-facing. If this is going in the release notes with a screenshot, the dialog needs to be fixed before this is mentioned. Therefore, it should be an immediate follow-up before the release... if this is going into this week's release.

It doesn't matter if this is changed in this PR or an immediate follow-up, but it needs to happen.

garrett:
Opened an issue @ https://github.com/cockpit-project/cockpit/issues/15536 to serve as the jump-off point for fixing the dialog. I've included a mockup and some example React which might serve as a place to start.

mvollmer:
> > If so, it has the following issues:
> 
> This PR does not change the dialog, just the text in the error case. Let's address these issues separately.

@garrett, I was half wrong about this, sorry.  Previously, Cockpit would immediately start the sudo conversation over and the error case would sometimes look like this:

 image

Does this change your assessment?

garrett:
That dialog is very odd and wrong too. Hmm.

garrett:
Quoting https://www.patternfly.org/v4/components/form/design-guidelines/#errors-and-validation, with added bolding on points where this new screenshot of the dialog has issues:

> When a form field submission results in an error, let users know as soon as possible. Always **present error states on the form using field level errors whenever possible**.
>
> Alternatively, you can also choose to emphasize errors using an **inline error at the top of the form**. Note that the **inline error should always be a supplemental component, used in addition to field-level errors**, and its message should generally **describe the problem**.
>
> When an error occurs, any placeholder text will no longer be visible as it will be replaced by the user’s input. Additionally, field level errors will replace any existing helper text until the error is fixed. Therefore, **it is important that your error message not only points out the error, but also explains what the user can do to fix it**.

Additionally, there's also https://www.patternfly.org/v4/components/form/design-guidelines/#validation-on-submission right below it, where it talks about validation on submission (whereas the above talks about it in general, also including on-the-fly error checking):

> **Validation on submission** is performed when the user attempts to submit a form. In this scenario, the user completes the form in its entirety **and the page is reloaded** with any errors detected.

This is for a form on a page, not in a modal. The page is reloaded, which is not really what happens here. But it is on submission.

> For validation on submission, use an inline alert along with field level errors. This **combination of error types should make it simple for the user to quickly identify areas that need to be fixed**. The inline alert should be placed below the form's header and description, and should take on the width of the form. Inline alerts should describe the general error, but refrain from going into great detail.

So here's the thing: This recommendation is based on a larger form with several inputs on a page. However, we're using a modal with one input, so it's obvious where the error is. If we double-up the errors here, it becomes redundant. So we would have to pick one.

This means we would _either_ have to have an inline notification at the top of the dialog _or_ have a contextual field-level error right under the input. But probably not both, despite the recommendation, due to how simple the form is here.

---

## Recommendations

1. Change the title text. It's not administrative access, it's asking for credentials for admin access. Perhaps: "Authenticate for administrative access"? It's too jargony for my liking, but it's better, at least.
2. Remove the body text description. The header, once renamed should handle this description.
3. If we're going for an inline notification, it should be at the top of the form. Otherwise, for password errors, they should be below the password input.
4. Change the text in the error. "This didn't work" is the problem here; it needs to be more specific. For example: "Password is not correct. Please try again."

Concerning the top inline alert versus the field-level error: I'm not sure at the moment which is best here. I'd have to see it. It probably also matters which issues can arise from authentication.

Here's what I'm sure of for errors:

- If it has issues not related to the password, such as not being able to contact the server (due to some network issue that shows up while the form is open, for example), that's definitely a top inline alert.
- If the user's password is correct but the user does not have access, then that's also a top-level inline alert. However, we _should_ know if the user can become admin or not before (speaking from a UX view, not an if-it's-possible-with-the-implementation view), and they shouldn't see this dialog. (Yeah, I think you said this isn't possible to know right now, if ever?)

So, even if we do have something at the field-level for the password-related errors, there would be cases where the inline alert at the top would be needed.

garrett:
We should show a counter, ideally counting down, when the wrong password is entered. Do we know if that number is always 3 times? Or do we know what it is set to, if configurable?

If we cannot, we could at least increment saying second attempt, third attempt, or such. Perhaps `Attempt #2`, `Attempt #3` or something like that, as strings for second, third, fourth, fifth, etc. would probably be more than a bit irksome to translate, especially if it's not always bound to 3 attempts, but could be like 50 or something silly.

mvollmer:
I'll try to give sudo a controlling tty, in a separate PR. Thus back to draft.

mvollmer:
> I'll try to give sudo a controlling tty, in a separate PR. Thus back to draft.

Bummer, sudo doesn't actually write errors to /dev/tty... so giving it a controlling tty won't help here.

allisonkarlitskaya:
I'll repeat my previous suggestion (from IRC, a week or two ago) here: add a new mode to cockpit-bridge which instructs it, on startup, to receive an fd from its stdin (`g_socket_new_from_fd()` → `g_socket_connection_factory_create_connection()` → `g_unix_connection_receive_fd()`) and `dup2()` it to stderr.

allisonkarlitskaya:
After some more chatting on IRC and in today's standup, it seems like we might land at a `--stderr-to-journal` switch to cockpit-bridge, which we set conditionally based on if the calling bridge's stderr is connected to the journal.  We'd then disable the stderr-capture code for sudo in the event that this isn't the case, effectively making this entire feature depend on cockpit running under systemd (as it usually is).

allisonkarlitskaya (changes requested):
Only reviewing ` bridge: Use stderr as D-Bus error message when starting peer bridges ` but the thing that jumps out at me is that it could all be substantially easier now: all of the stderr changes to cockpitpipe are probably no longer needed, and particularly, the code to do the forwarding should be removed: stderr will end up being closed directly after sudo by the `dup2()` in the peer bridge...

mvollmer:
> Only reviewing `bridge: Use stderr as D-Bus error message when starting peer bridges` but the thing that jumps out at me is that it could all be substantially easier now: all of the stderr changes to cockpitpipe are probably no longer needed, and particularly, the code to do the forwarding should be removed: stderr will end up being closed directly after sudo by the `dup2()` in the peer bridge...

I think we should keep `cockpit_pipe_take_stderr_as_utf8`. It's complicated enough to warrant its own function. But it doesn't need to be part of CockpitPipe. Would you rather see it be standalone in cockpitunicode.c, say?

As to `cockpit_pipe_stop_stderr_capture`... hmm. Right now I am thinking I would like to keep it as insurance. _If_ stderr in the peer bridge is closed immediately after startup, then the forwarding code wont run and we don't need it.

But what if for some reason things get messed up and stderr stays open? Without any changes to CockpitPipe, we would suck it all into a in-memory buffer, for the while life-time of the peer bridge, never to be seen again. I would rather dump it somewhere. This prevents unlimited growth of the in-memory buffer, and it doesn't hide potentially important messages. WDYT?

mvollmer:
>  I would rather dump it somewhere.

But it's an exceptional situation now, so we could log it as such: `g_warning("%s: unexpected stderr output: *.%s", ....)`

allisonkarlitskaya:
I had exactly the same concern as you about the stderr buffer, and was kinda ignoring it.  After thinking about it a bit more, though, I prefer this solution:

I think we should add a helper to `CockpitPipe`, something along the lines of `cockpit_pipe_is_error_done()` which returns the value of `priv->err_done`, possibly calling `dispatch_error()` first, non-blocking, in case that it's not done.  That would let us check, effectively, that stderr has indeed been closed by the remote, as we expect.

Then on receiving the init message from a privileged peer bridge, we'd call `cockpit_pipe_is_error_done()` to make sure.  In case stderr is not closed, then bring the whole peer down.

That all seems like a much better/more-direct way to deal with "make sure things are as we expect them to be" than writing a complicated piece of dead code that hopefully never runs.  It's also going to lead us to discovering any unexpected problems a lot quicker.

allisonkarlitskaya:
My approach is non-viable: sudo calls `fork()` and hangs around as a monitor process in order to perform PAM cleanup.  In doing so, it keeps stderr open (and may actually even end up writing to it).

mvollmer:
"needswork" for the journal messages.

src/bridge/askpass.c
```diff
@@ -186,8 +186,8 @@ main (int argc,
   cookie = g_strdup_printf ("askpass%u%u", (unsigned int)getpid (), (unsigned int)time (NULL));
 
   prompt = argv[1];
-  if (strstr (prompt, "[sudo] ") == prompt)
-    prompt = NULL;
+  /* if (strstr (prompt, "[sudo] ") == prompt) */
```
allisonkarlitskaya, line 189: ...erm?

src/common/cockpitpipe.c
```diff
@@ -405,12 +419,19 @@ dispatch_error (gint fd,
               close_immediately (self, "internal-error");
               return FALSE;
             }
+
+          if (priv->err_forward_to_log)
```
allisonkarlitskaya, line 423: Why don't we just check this condition at the start and log directly a `g_warning()` instead of messing around with the buffer?  I almost feel like we could git rid of the separate `err_forward_to_log` by nulling out the `err_buffer` pointer, in fact.

Alternatively, I kinda liked your idea to just close the stderr fd.......

allisonkarlitskaya (commented):
We're pretty close now, I think.

The points here are open questions, but really just questions.  I'm OK if you want to leave them this way.

src/bridge/askpass.c
mvollmer: Removed.

src/common/cockpitpipe.c
mvollmer: I think just continuing to pump stderr is the safest option. Otherwise we might lose important messages.  And with that sudo background watching process being around all the time, we can't actually be sure that nobody writes to stderr even if the bridge has successfully redirected it, no?

allisonkarlitskaya: Right.  We might well see a stderr message from a pam module running at session shutdown, for example...

src/bridge/cockpitpeer.c
```diff
@@ -96,15 +96,23 @@ G_DEFINE_TYPE (CockpitPeer, cockpit_peer, G_TYPE_OBJECT);
 
 static void
 startup_done (CockpitPeer *self,
-              const gchar *problem)
+              const gchar *problem,
+              const gchar *stderr)
```
allisonkarlitskaya, line 100: I feel like we could increase what this function is doing:

1) drop the `stderr` parameter again
2) if `problem != NULL` then call `take_stderr()` ourselves
3) always stop stderr here

That could probably remove some of the extra handling code that is happening at various call sites below.

src/bridge/cockpitpeer.c
```diff
@@ -366,12 +378,15 @@ on_other_control (CockpitTransport *transport,
             {
               gchar *user_hex;
               gchar *user;
+              gchar *stderr;
               self->startup_auth_cookie = g_strdup (cookie);
               cockpit_authorize_subject (challenge, &user_hex);
               user = cockpit_hex_decode (user_hex, -1, NULL);
-              cockpit_router_prompt (self->router, user, prompt, on_answer, self);
+              stderr = startup_take_stderr (self);
```
allisonkarlitskaya, line 385: This is pretty extremely pedantic, but it might happen that sudo prints some output to stderr, then calls askpass, which prints an auth control message to stdout.  On a busy system it might end up that cockpit-bridge doesn't get a change to run again until after that all happens (maybe it was in swap?) and then when it comes back it has a 50/50 chance to read from the socket or the stderr pipe first.  If it reads from the socket first, we'll see the control message before we have all of our stderr data from sudo.

We could fix this by manually polling stderr to check if there's outstanding data before we empty the buffer.  That's probably something that belongs in `CockpitPipe`.

Addressing this now, however unlikely we expect it to be in practice, might avoid some future flakes turning up because of this.  stdout/err ordering issues when they are not on the same fd/pty/whatever are really tricky to trace down.

allisonkarlitskaya (changes requested):

src/bridge/cockpitpeer.c
mvollmer: Good point!

src/bridge/cockpitpeer.c
mvollmer: There is one case where we destroy `self->other` before calling `startup_done` (and we need to get stderr before the destruction). I see if that can be made to work anyway.

src/bridge/cockpitpeer.c
mvollmer: I have added the `drain_error` function to read all that is available when getting the stderr output.

src/bridge/cockpitpeer.c
mvollmer: Seems to work, thanks for the suggestion!

martinpitt:
I took the liberty to rebase this, as this had a conflict in check-packagekit. Code-wise this looks good to me, but I'll leave the C part review to @allisonkarlitskaya as she has the most context about the stderr handling parts.

martinpitt (approved):
I also tested this interactively fow now. There was an incident, it got reported, not sure where.. but I am deeply sorry! :grinning: 

Thanks a lot! I love this feature, it's now really exposing what sudo is, in a tasteful way.

martinpitt:
@mvollmer: I retried [fedora-33/firefox three times](https://logs.cockpit-project.org/logs/pull-15293-20210428-052424-1879d7e4-fedora-33-firefox/log.html#232), and `test/verify/check-shell-keys TestKeys.testAuthorizedKeys` just won't budge. So I'm afraid this needs work still :cry:

martinpitt:
I can reproduce the failure fairly well locally, even through for me it happens two lines above the one in CI. With `-t` it looks like this:
```
-> wait: !ph_is_present(".pf-c-modal-box:contains('Switch to administrative access')")
<- {'type': 'undefined'}
-> wait: ph_is_present("#super-user-indicator")
<- {'type': 'undefined'}
-> wait: ph_is_visible("#super-user-indicator")
<- {'type': 'undefined'}
-> wait: ph_text_is("#super-user-indicator","Administrative access")
<- {'type': 'undefined'}
-> switch to frame None
-> wait: ph_is_present("iframe.container-frame[name='cockpit1:localhost/users'][data-loaded]")
<- {'type': 'undefined'}
-> wait: !ph_is_visible(".curtains-ct")
<- {'type': 'undefined'}
-> wait: ph_is_present("iframe.container-frame[name='cockpit1:localhost/users']")
<- {'type': 'undefined'}
-> wait: ph_is_visible("iframe.container-frame[name='cockpit1:localhost/users']")
<- {'type': 'undefined'}
-> switch to frame cockpit1:localhost/users
-> wait: ph_is_present("body")
CDP: executionContextDestroyed 12
CDP: frameNavigated {"frame":{"id":"2147483654","parentId":"19","loaderId":"127294240718954","url":"http://127.0.0.2:9091/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/users/index.html#/user","name":"cockpit1:localhost/users","securityOrigin":null,"mimeType":null}}
CDP: executionContextCreated {"context":{"id":13,"origin":"http://127.0.0.2:9091/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/users/index.html#/user","name":"","auxData":{"isDefault":true,"frameId":"2147483654","type":"default"}}}
<- raise {'request': {'method': 'Runtime.evaluate', 'params': {'expression': 'ph_wait_cond(() => ph_is_present("body"), 60000)', 'silent': False, 'awaitPromise': True, 'trace': 'wait: ph_is_present("body")', 'contextId': 12}}, 'response': {'message': 'Cannot find context with specified id', 'data': 'evaluate@chrome://remote/content/cdp/domains/content/Runtime.jsm:273:15\nexecute@chrome://remote/content/cdp/domains/DomainCache.jsm:101:25\nreceiveMessage@chrome://remote/content/cdp/sessions/ContentProcessSession.jsm:84:45\n'}}
  File "/usr/lib64/python3.9/unittest/case.py", line 550, in _callTestMethod
    method()
  File "/var/home/martin/upstream/c-sudo/test/verify/check-shell-keys", line 122, in testAuthorizedKeys
    b.enter_page("/users")
  File "/var/home/martin/upstream/c-sudo/test/common/testlib.py", line 522, in enter_page
    self._wait_present("body")
  File "/var/home/martin/upstream/c-sudo/test/common/testlib.py", line 395, in _wait_present
    self.wait_js_func('ph_is_present', selector)
  File "/var/home/martin/upstream/c-sudo/test/common/testlib.py", line 389, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
  File "/var/home/martin/upstream/c-sudo/test/common/testlib.py", line 381, in wait_js_cond
    result = self.cdp.invoke("Runtime.evaluate",
  File "/var/home/martin/upstream/c-sudo/test/common/cdp.py", line 109, in invoke
    res = self.command(cmd)
  File "/var/home/martin/upstream/c-sudo/test/common/cdp.py", line 136, in command
RuntimeError: {'request': {'method': 'Runtime.evaluate', 'params': {'expression': 'ph_wait_cond(() => ph_is_present("body"), 60000)', 'silent': False, 'awaitPromise': True, 'trace': 'wait: ph_is_present("body")', 'contextId': 12}}, 'response': {'message': 'Cannot find context with specified id', 'data': 'evaluate@chrome://remote/content/cdp/domains/content/Runtime.jsm:273:15\nexecute@chrome://remote/content/cdp/domains/DomainCache.jsm:101:25\nreceiveMessage@chrome://remote/content/cdp/sessions/ContentProcessSession.jsm:84:45\n'}}
CDP: {"source":"other","level":"warning","text":"Layout was forced before the page was fully loaded. If stylesheets are not yet loaded this may cause a flash of unstyled content.","timestamp":1619617756355,"url":"http://127.0.0.2:9091/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/users/index.html#/user","lineNumber":0}
```

So the "users" frame gets reloaded and ripped out underneath, so the "current" frame is the old users frame, which got reloaded in between. That explains why it's blank. For me the timing is just a little different, but same root cause.

martinpitt:
@mvollmer: I took the liberty to force-push the fix directly here, see the [interdiff](https://github.com/cockpit-project/cockpit/compare/1879d7e4404618a0b431eef0e5a87d71af5250c7..d5fe5495f8dd0e96c3eed41fe39f1e4b11927809). I deliberately did not rebase to keep that readable. With that I have this test running in two parallel loops for several minutes now, before roughly ever other run failed.

martinpitt (approved):

martinpitt:
Bah, now [testFrameReload](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-fedora-33-firefox/log.html#98-1) failed on firefox. I'll take a look..

martinpitt:
In fact there are tons of failures now, did something in that area change recently? (PatternFly?)

src/common/cockpitpipe.c
```diff
@@ -427,6 +448,31 @@ dispatch_error (gint fd,
...
+  fd_set rfds;
+  struct timeval tv;
+
+  FD_ZERO (&rfds);
+  FD_SET (fd, &rfds);
+
+  tv.tv_sec = 0;
+  tv.tv_usec = 0;
+
+  return select (1, &rfds, NULL, NULL, &tv) > 0;
```
allisonkarlitskaya, line 463:
```suggestion
  struct pollfd fd = { .fd = fd, .events = POLLIN };

  return poll (&fd, 1, 0) > 0;
```

allisonkarlitskaya (changes requested):

garrett:
Suggested mod for release note text:

---

## Shell: Improve sudo usage

Cockpit's use of sudo used to be noisy in the journal. This verbosity was due to checking if a logged in user had the ability to gain admin privileges, using `sudo -v -n` . Instead, Cockpit now shows sudo's exact messaging in the "Administrative access" dialog.

---

Also, why is that dialog screenshot _so very wide_? Should we constrain the width a little bit in this PR and then retake the screenshot? :wink:

```jsx
<Modal variant={ModalVariant.small}>
```

(Or add `pf-m-sm` to `pf-c-modal-box` in non-React.)

martinpitt:
> did something in that area change recently

Well, yes, sudo changed :grin:  This now causes the frame reloads (due to `superuser.reload_page_on_change()`) to happen at a different time, similarly to the testAuthorizedKeys() failure above. This is my local trace of testFrameReload(), this time it fails at exactly the same spot as on CI:

```
-> switch to frame None
-> wait: ph_is_present("iframe.container-frame[name='cockpit1:10.111.113.2/playground/test'][data-loaded]")
<- {'type': 'undefined'}
-> wait: !ph_is_visible(".curtains-ct")
CDP: executionContextDestroyed 8
CDP: frameNavigated {"frame":{"id":"2147483653","parentId":"19","loaderId":"78913581613126","url":"http://127.0.0.2:9191/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/test.html?i=1#/","name":"cockpit1:10.111.113.2/playground/test","securityOrigin":null,"mimeType":null}}
CDP: executionContextCreated {"context":{"id":10,"origin":"http://127.0.0.2:9191/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/test.html?i=1#/","name":"","auxData":{"isDefault":true,"frameId":"2147483653","type":"default"}}}
CDP: {"source":"other","level":"warning","text":"This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”.","timestamp":1619682601463,"url":"http://127.0.0.2:9191/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/test.html?i=1#/","lineNumber":0}
<- {'type': 'undefined'}
-> wait: ph_is_present("iframe.container-frame[name='cockpit1:10.111.113.2/playground/test']")
<- {'type': 'undefined'}
-> wait: ph_is_visible("iframe.container-frame[name='cockpit1:10.111.113.2/playground/test']")
<- {'type': 'undefined'}
-> switch to frame cockpit1:10.111.113.2/playground/test
-> wait: ph_is_present("body")
<- {'type': 'undefined'}
-> wait: ph_is_present("body")
CDP: executionContextDestroyed 10
CDP: frameNavigated {"frame":{"id":"2147483653","parentId":"19","loaderId":"78913581613130","url":"http://127.0.0.2:9191/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/test.html#/","name":"cockpit1:10.111.113.2/playground/test","securityOrigin":null,"mimeType":null}}
CDP: executionContextCreated {"context":{"id":11,"origin":"http://127.0.0.2:9191/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/test.html#/","name":"","auxData":{"isDefault":true,"frameId":"2147483653","type":"default"}}}
<- raise {'request': {'method': 'Runtime.evaluate', 'params': {'expression': 'ph_wait_cond(() => ph_is_present("body"), 60000)', 'silent': False, 'awaitPromise': True, 'trace': 'wait: ph_is_present("body")', 'contextId': 10}}, 'response': {'message': 'Cannot find context with specified id', 'data': 'evaluate@chrome://remote/content/cdp/domains/content/Runtime.jsm:273:15\nexecute@chrome://remote/content/cdp/domains/DomainCache.jsm:101:25\nreceiveMessage@chrome://remote/content/cdp/sessions/ContentProcessSession.jsm:84:45\n'}}
  File "/usr/lib64/python3.9/unittest/case.py", line 550, in _callTestMethod
    method()
  File "/var/home/martin/upstream/c-sudo/test/verify/check-shell-multi-machine", line 616, in testFrameReload
    b.enter_page("/playground/test", "10.111.113.2")
```

However, that was a good explanation for testAuthorizedKeys(), but not really for  the playground, that doesn't use `superuser.reload_page_on_change()`. The comment in the test says "go back to m2 and reload frame", it's just not clear to me right now what exactly triggers that.

martinpitt:
I have a local fix for testFrameReload(). Looking at the [testSshKeySetup()](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-fedora-coreos/log.html#99) failure now which [happened twice](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-ubuntu-stable/log.html#99-2). Also reproduces locally on fedora-33, it's not browser/OS specific:

The issue here is that sometimes it's [showing](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-fedora-coreos/TestMultiMachine-testSshKeySetup-fedora-coreos-127.0.0.2-2701-FAIL.png) "Reconnect" (`#machine-reconnect`) instead of "Login" (`#machine-troubleshoot`).

These mainly come from base_index.js' watchdog:
```js
    /* Watchdog for disconnect */
    var watchdog = cockpit.channel({ payload: "null" });
    $(watchdog).on("close", function(event, options) {
        var watchdog_problem = options.problem || "disconnected";
        console.warn("transport closed: " + watchdog_problem);
        $(self).triggerHandler("disconnect", watchdog_problem);
    });
```
But this is supposed to write that "transport closed:" warning which I don't see in the logs. So I think this one comes from pkg/shell/machines/machines.js (which makes sense, as we are connected to a remote machine). At this point:
```py
        m2.execute("rm /home/fred/.ssh/authorized_keys")
        b.relogin(None)
        b.wait_visible("#machine-troubleshoot")
```
it is very unlikely that the watchdog from the *new* session already fired, it rather smells like there is still some handler leftover from the previous session?

I diffed the trace log between a [successful](https://github.com/cockpit-project/cockpit/files/6398442/good.txt) and a [failed](https://github.com/cockpit-project/cockpit/files/6398444/bad.txt) run, and the only difference is that the bad one does not have this additional created/removed pair of the http://127.0.0.2:9091 shell frame.
```diff
 -> wait: ph_is_visible("#content")
-CDP: executionContextCreated {"context":{"id":15,"origin":"http://127.0.0.2:9091","name":"","uniqueId":"3543935714501317035.9136949532999889683","auxData":{"isDefault":true,"type":"default","frameId":"9CA2226CBE11F1597A1AA2BAE8697398"}}}
-CDP: executionContextCreated {"context":{"id":16,"origin":"http://127.0.0.2:9091","name":"","uniqueId":"-620074773911249754.-402270376999907816","auxData":{"isDefault":true,"type":"default","frameId":"AF17FEC1CF5FE60E1C32F9ABE685E8AE"}}}
-CDP: frameNavigated {"frame":{"id":"9CA2226CBE11F1597A1AA2BAE8697398","parentId":"3BE52124EDDC031099399F6FBC0B9A13","loaderId":"A099D7AA2DC2CCF169648D7C9BE011AB","name":"cockpit1:localhost/playground/preloaded","url":"http://127.0.0.2:9091/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/preloaded.html","urlFragment":"#/","domainAndRegistry":"","securityOrigin":"http://127.0.0.2:9091","mimeType":"text/html","adFrameType":"none","secureContextType":"SecureLocalhost","crossOriginIsolatedContextType":"NotIsolated","gatedAPIFeatures":["SharedArrayBuffers","SharedArrayBuffersTransferAllowed"]}}
-CDP: executionContextDestroyed 16
+CDP: executionContextCreated {"context":{"id":15,"origin":"http://127.0.0.2:9191","name":"","uniqueId":"-6771236617888697113.7154717187782556697","auxData":{"isDefault":true,"type":"default","frameId":"CCE11E50B51324901B2053EB3B5CA11B"}}}
+CDP: frameNavigated {"frame":{"id":"CCE11E50B51324901B2053EB3B5CA11B","parentId":"A09557CA2582DEE4E4021D741F15B751","loaderId":"4AF2B61DE57D3574C5DE7E60A05D315F","name":"cockpit1:localhost/playground/preloaded","url":"http://127.0.0.2:9191/cockpit/$c438832f0fdad2ce34af913243e536d2b2037e81244c9935812168bfc9e10f8d/playground/preloaded.html","urlFragment":"#/","domainAndRegistry":"","securityOrigin":"http://127.0.0.2:9191","mimeType":"text/html","adFrameType":"none","secureContextType":"SecureLocalhost","crossOriginIsolatedContextType":"NotIsolated","gatedAPIFeatures":["SharedArrayBuffers","SharedArrayBuffersTransferAllowed"]}}
 <- {'type': 'undefined'}
 -> wait: ph_is_present("#machine-troubleshoot")
```
But I don't think that's going to help us -- it rather smells like there is some cleanup action of logout() still happening.

martinpitt:
I am looking into [TestSuperuser.testBasic](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-centos-8-stream/log.html#234-2) failure. I can locally reproduce that on centos-8-stream, but no luck on fedora so far. Just like [on CI](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-centos-8-stream/log.html#234-2), this is (rightfully) expecting the lecture in the dialog, but it is [missing](https://logs.cockpit-project.org/logs/pull-15293-20210429-054124-d5fe5495-centos-8-stream/TestSuperuser-testBasic-centos-8-stream-127.0.0.2-2501-FAIL.png).

/var/db/sudo/lectured/ is empty when this happens, and when I  ssh into the VM after the failure as admin and use sudo, I *do* get lectured. So this smells like some race condition to capture sudo's stderr in some cases :cry: , i.e. precisely the thing that @allisonkarlitskaya and @mvollmer have spent weeks on trying to fix. When I use the browser interactively, I do get the lecture,  so it's not some persistently wrong state on the system.

What did you use to debug the sudo conversation while developing this?

martinpitt:
I pushed the fix for TestMultiMachine.testFrameReload(), and added another commit that *may* fix TestMultiMachine.testSshKeySetup(). This is a rather weak theory, but as I have serious trouble reproducing this, I want to give it a spin.

src/common/cockpitpipe.c
martinpitt: Pushed with two fixes (missing #include, and not duplicating the `fd` as both the function argument and the struct  variable name).

martinpitt:
Whee, all green! TestSuperuser.testBasic() failed three times on the above "missing lecture" issue, once on [rhel-8-5](https://logs.cockpit-project.org/logs/pull-15293-20210429-155243-3607f789-rhel-8-5/log.html#248) and twice on [centos-8-stream](https://logs.cockpit-project.org/logs/pull-15293-20210429-155243-3607f789-centos-8-stream/log.html#248). Honestly I'm inclined to investigate that in a follow-up, given how intrusive (bit-rot prone) this is, and that we are at the beginning of a release cycle, and that occasionally missing the lecture is not *that* much of a dealbreaker (given that so far we never shown it), but let's talk about that today.

I applied @allisonkarlitskaya 's select() → poll() simplification above, let's run this for another round.

I also looked at @garrett's suggestion to make the dialog less wide. With ModalVariant.small it looks like this:

image

I am not sure if the "Please authenticate..." should really be aligned with the input line or rather to the left with the label. I did not commit this yet, I suppose we could do that fine-tuning in a follow-up as well if we wanted to.

I also reworked the release note above with Garrett's suggestions. Both the original and the suggestion was a non-sequitur (the `sudo -v -n` probing was not actually the cause of not showing the exact messages, these are two independent changes). WDYT?

mvollmer:
From the propsed release notes:

> Cockpit no longer does that, so the very first time a user logs in they have to explicitly turn on administrative access.

These two things are actually separate. (Thing one: not running sudo -v -n; thing two: having to explicitly turn on administrative access.)

Cockpit used to run `sudo -v -n` when opening the "Switch to administrative access" dialog.  This was necessary because we couldn't distinguish the different failure reasons. Typing a wrong password looked the same as not being in the sudoers file. So Cockpit had to determine upfront whether sudo has any chance at all of succeeding.

Cockpit used to attempt sudo on the very first login.  This is no longer done, but it has nothing to do with running `sudo -v -n`.

Both changes were done to reduce noise in the journal, but are pretty independent.

So what about replacing that sentence with: "Additionally, Cockpit will no longer try to run sudo on the very first login, and you have to explicitly turn on administrative access once for each browser profile that you use to access Cockpit."

allisonkarlitskaya (approved):

martinpitt (dismissed):

martinpitt:
@mvollmer: We'd have two "Additionally" there, and just replacing the second sentence would still be confusing then. Maybe flip it around, and entirely drop the `sudo -v -n` detail?

> Cockpit will no longer try to run sudo on the very first login, and you have to explicitly turn on administrative access once for each browser profile that you use to access Cockpit. This prevents cockpit's use of sudo being noisy in the journal.
>
> In addition, Cockpit now shows sudo's exact messages in the "Administrative access" dialog.

mvollmer:
> Maybe flip it around, and entirely drop the `sudo -v -n` detail?

Sounds good!  Of course, being able to skip `sudo -v -n` took 95% of the work here :) but it is a implementation detail.

mvollmer:
@martinpitt, thanks for the test fixes! They look good to me, although I can't explain the TestMultiMachine.testSshKeySetup() flakiness either.  How does a new session get "disconnected" signals from the previous session?

A fix that might work as well is to wait for the current local session to disappear (instead of waiting for there to be no cockpit-ssh processes on a remote machine). That would be less specific and less magic. Should we try that?  (Using the code we use during teardown of nondestructive tests.)

martinpitt:
@mvollmer: I indeed wanted to start with the `loginctl` parsing loop, then thought "no, these are remote sessions", but then realized that it's enough to scan for cockpit-ssh locally (as we don't really care about lingering remote sessions). But of course all of cockpit-ssh once again should run in the primary local session. So yes, let's try that, I'd have a much better feeling about that.

mvollmer:
> So yes, let's try that, I'd have a much better feeling about that.

I'll give it a go.

mvollmer:
> > So yes, let's try that, I'd have a much better feeling about that.
> 
> I'll give it a go.

Ah, you were faster.

martinpitt:
@mvollmer: I pushed a simplified version, let's see how that works out.

martinpitt:
Meh, so that simplification [failed](https://logs.cockpit-project.org/logs/pull-15293-20210430-091524-ed559bdc-ubuntu-2004/log.html#260-1). It's still kind of inexplicable to me why, though.

martinpitt:
I reverted back to 42a453c3fa48d75b9baea634aab6a7397760ca73 which was green (now we still need to battle the heavily flaky unit test, but they were green before as well)

martinpitt (approved):

garrett:
> I am not sure if the "Please authenticate..." should really be aligned with the input line or rather to the left with the label. I did not commit this yet, I suppose we could do that fine-tuning in a follow-up as well if we wanted to.

Yeah, it shouldn't be aligned with the text entry, unless it would be below the text entry and discussing the text entry itself as helper text.

It's test at the top of the dialog and the entry is supposed to support it, not the other way around.

In other words: It should be left-aligned.

garrett:
> It should be left-aligned.

Any follow-up on this yet?

garrett:
I still see it...
Screenshot_2021-05-11 Services - garrett Rain


But I'm on a branch. However, I think this branch is newer than 11 days ago.

PR #15294, opened 2021-02-09
Build `make dist` artifact for every master commit and PR

martinpitt:
The npm install/webpack steps are very expensive, and annoying to do
locally for reviewing PRs or updating a local checkout. Introduce a new
"build-dist" workflow which creates a downloadable dist tarball for each
interesting commit (PRs and pushes to master for now).

Split out the building part from scenario-distcheck.sh into a new "dist"
scenario, so that we can finish this workflow in the least possible
amount of time, as we soon want to trigger the tests after this
workflow, to re-use the built dist tarball.

Skip building the cockpit-cache*.tar.xz for this workflow -- we don't
need it for CI and it's expensive to build.

------

Note: This does not yet use the built dist tarball in unit nor integration tests. I'd like to build this one step at a time. The next step would be to adjust image-prepare or perhaps make-source to try and download that tarball, and we can see how that works for us developers. After we settle this, and @allisonkarlitskaya is done with her "optimize dist tarball" work, we can take the next steps.

### DIFF
--- a/containers/unit-tests/scenario-dist.sh
+++ b/containers/unit-tests/scenario-dist.sh
@@ -1,0 +1,20 @@
+#!/bin/sh -eux
+
+if [ "${NO_NPM:-}" = "1" ]; then
+    # We can't 'make dist' with NO_NPM, so unset it and finish the build
+    unset NO_NPM
+    test ! -d /tmp/source/node_modules # this shouldn't be here
+    if [ -d /source/node_modules ]; then
+        cp -r /source/node_modules /tmp/source
+    fi
+    tools/npm-install
+    test -d /tmp/source/node_modules # this must surely be here now
+    make
+fi
+
+make NO_DIST_CACHE=1 dist
+
+# container has a writable /results/ in the "dist" scenario, but not in others; copy dist tarball there
+if [ -w /results ]; then
+    cp cockpit-*.tar.xz /results/
+fi
--- a/containers/unit-tests/scenario-distcheck.sh
+++ b/containers/unit-tests/scenario-distcheck.sh
@@ -1,21 +1,7 @@
 #!/bin/sh -eux
 
-if [ "${NO_NPM:-}" = "1" ]; then
-    # We can't 'make dist' with NO_NPM, so unset it and finish the build
-    unset NO_NPM
-    test ! -d /tmp/source/node_modules # this shouldn't be here
-    if [ -d /source/node_modules ]; then
-        cp -r /source/node_modules /tmp/source
-    fi
-    tools/npm-install
-    test -d /tmp/source/node_modules # this must surely be here now
-    make
-fi
-
-make XZ_COMPRESS_FLAGS='-0' V=0 distcheck 2>&1 || {
-  find -name test-suite.log | xargs cat
-  exit 1
-}
+# build dist tarball fist
+"$(dirname "$0")"/scenario-dist.sh
 
 # check translation build
 make po/cockpit.pot
--- a/containers/unit-tests/start
+++ b/containers/unit-tests/start
@@ -1,4 +1,5 @@
-#!/bin/sh -e
+#!/bin/sh
+set -eu
 
 top_srcdir=$(realpath -m "$0"/../../..)
 image=ghcr.io/cockpit-project/unit-tests:latest
@@ -6,6 +7,7 @@ no_npm=''
 scenario=''
 flags=''
 cmd=''
+results_vol=''
 
 while test $# -gt 0; do
   case "$1" in
@@ -16,6 +18,12 @@ while test $# -gt 0; do
     shell) flags=-it; cmd=/bin/bash;;
     :*) image=ghcr.io/cockpit-project/unit-tests"$1";;
     distcheck|check-memory) scenario="--env=TEST_SCENARIO=$1";;
+    dist)
+        scenario="--env=TEST_SCENARIO=$1"
+        results_vol="--volume=${top_srcdir}/tmp/dist:/results:Z"
+        mkdir -m 777 -p "${top_srcdir}/tmp/dist"
+        ;;
+
     *) echo "Unknown option '$1'"; exit 1;;
   esac
   shift
@@ -33,4 +41,5 @@ fi
 
 set -ex
 exec ${docker} run --shm-size=512M --volume "${top_srcdir}":/source:ro \
+       ${results_vol:+"$results_vol"} \
        ${ccarg:+"$ccarg"} $scenario $no_npm $flags -- "$image" $cmd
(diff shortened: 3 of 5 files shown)


martinpitt:
I [tested this on my fork](https://github.com/martinpitt/cockpit/actions/runs/550840855). In https://github.com/mpitt-test-org/actions-playground/pull/1 I also have a scheme how to run the "trigger integration tests" workflow after the "build dist artifact" workflow.

martinpitt:
Workflow [produced an artifact](https://github.com/cockpit-project/cockpit/actions/runs/550884163), and it's properly exposed on the API: https://api.github.com/repos/cockpit-project/cockpit/actions/artifacts

containers/unit-tests/scenario-dist.sh
```diff
@@ -0,0 +1,18 @@
...
+    fi
+    tools/npm-install
+    test -d /tmp/source/node_modules # this must surely be here now
+    make
+fi
+
+make XZ_COMPRESS_FLAGS='-0' NO_DIST_CACHE=1 dist
+
+# container has a writable /results/ in the "dist" scenario, but not in others; copy dist tarball there
+[ ! -w /results ] || cp cockpit-*.tar.xz /results/
```
allisonkarlitskaya, line 18: Can you write this as an `if`?

```suggestion
if [ -w /results ]; then
    cp cockpit-*.tar.xz /results/
fi
```

containers/unit-tests/scenario-dist.sh
```diff
@@ -0,0 +1,18 @@
...
+    test ! -d /tmp/source/node_modules # this shouldn't be here
+    if [ -d /source/node_modules ]; then
+        cp -r /source/node_modules /tmp/source
+    fi
+    tools/npm-install
+    test -d /tmp/source/node_modules # this must surely be here now
+    make
+fi
+
+make XZ_COMPRESS_FLAGS='-0' NO_DIST_CACHE=1 dist
```
allisonkarlitskaya, line 15: Why `-0`?  If we're not disting the cache, we're gonna save a lot of time on the compression anyway, and a smaller tarball might be nicer...

containers/unit-tests/scenario-distcheck.sh
```diff
@@ -1,21 +1,7 @@
...
-    test -d /tmp/source/node_modules # this must surely be here now
-    make
-fi
-
-make XZ_COMPRESS_FLAGS='-0' V=0 distcheck 2>&1 || {
-  find -name test-suite.log | xargs cat
-  exit 1
-}
+# build dist tarball fist
+$(dirname $0)/scenario-dist.sh
```
allisonkarlitskaya, line 4:
```suggestion
"$(dirname "$0")"/scenario-dist.sh
```

containers/unit-tests/start
```diff
@@ -16,6 +18,12 @@ while test $# -gt 0; do
     shell) flags=-it; cmd=/bin/bash;;
     :*) image=ghcr.io/cockpit-project/unit-tests"$1";;
     distcheck|check-memory) scenario="--env=TEST_SCENARIO=$1";;
+    dist)
+        scenario="--env=TEST_SCENARIO=$1"
+        results_vol="--volume ${top_srcdir}/tmp/dist:/results:Z"
+        mkdir -m 777 -p "${top_srcdir}/tmp/dist"
```
allisonkarlitskaya, line 24: Can't say I find this to be particularly wonderful...

I start to think we really need to change the way we invoke the container.  Our simple `run` approach is getting increasingly brittle.  I feel like we might split it into a more general create/copy/start/copy type of approach...

That can be saved for another day, though.

containers/unit-tests/start
```diff
@@ -16,6 +18,12 @@ while test $# -gt 0; do
     shell) flags=-it; cmd=/bin/bash;;
     :*) image=ghcr.io/cockpit-project/unit-tests"$1";;
     distcheck|check-memory) scenario="--env=TEST_SCENARIO=$1";;
+    dist)
+        scenario="--env=TEST_SCENARIO=$1"
+        results_vol="--volume ${top_srcdir}/tmp/dist:/results:Z"
```
allisonkarlitskaya, line 23: If you use the `--volume=` syntax here then you can properly quote this below.

```suggestion
        results_vol="--volume=${top_srcdir}/tmp/dist:/results:Z"
```

containers/unit-tests/start
```diff
@@ -32,5 +40,5 @@ if test -z "${docker:=$(which podman || which docker || true)}"; then
 fi
 
 set -ex
-exec ${docker} run --shm-size=512M --volume "${top_srcdir}":/source:ro \
+exec ${docker} run --shm-size=512M --volume "${top_srcdir}":/source:ro $results_vol \
```
allisonkarlitskaya, line 43: You can quote this like the `ccarg` if you use `=` above.

```suggestion
exec ${docker} run --shm-size=512M --volume "${top_srcdir}":/source:ro ${results_vol:+"$results_vol"} \
```

allisonkarlitskaya (changes requested):

containers/unit-tests/scenario-dist.sh
martinpitt: I mostly kept it as it was there before, and it does save a few minutes. But for this stage of the PR it's indeed not that relevant, it will become important only when integration tests get chained after this. Also note that the distcheck scenarios use this code as well. But I can take it out for now.

martinpitt:
Thanks @allisonkarlitskaya for the thorough review! I applied all you comments/suggestions. (Untested locally as I need to run out now, but if it breaks I'll debug it locally in an hour)

allisonkarlitskaya (approved):
:+1:

containers/unit-tests/scenario-distcheck.sh
```diff
@@ -1,21 +1,7 @@
...
-    test -d /tmp/source/node_modules # this must surely be here now
-    make
-fi
-
-make XZ_COMPRESS_FLAGS='-0' V=0 distcheck 2>&1 || {
-  find -name test-suite.log | xargs cat
-  exit 1
-}
+# build dist tarball fist
+"$(dirname "$0")"/scenario-dist.sh
```
martinpitt, line 4: @allisonkarlitskaya : As dropping the `XZ_COMPRESS_FLAGS='-0'` takes three extra minutes (for the unit tests), should we specify it here in exchange? So thaht at least our four unit test scenarios don't run slower?

martinpitt: Hmm, I figure an environment variable won't override the `XZ_COMPRESS_FLAGS = --extreme` declaration, only a `make VAR=value` does, right?

PR #15295, opened 2021-02-09
lib: add fileExists property to FileAutocomplete component

KKoukiou:
cockpit-certificates uses a custom version of that component that can be
dropped once FileAutocomplete supports creatable files.

### DIFF
--- a/pkg/lib/cockpit-components-file-autocomplete.jsx
+++ b/pkg/lib/cockpit-components-file-autocomplete.jsx
@@ -41,6 +41,7 @@ export class FileAutoComplete extends React.Component {
         this.finishUpdate = this.finishUpdate.bind(this);
         this.onToggle = this.onToggle.bind(this);
         this.clearSelection = this.clearSelection.bind(this);
+        this.onCreateOption = this.onCreateOption.bind(this);
 
         this.debouncedChange = debounce(300, (value) => {
             if (!value)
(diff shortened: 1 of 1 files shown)


pkg/lib/cockpit-components-file-autocomplete.jsx
```diff
@@ -184,8 +193,10 @@ FileAutoComplete.propTypes = {
     id: PropTypes.string,
     placeholder: PropTypes.string,
     superuser: PropTypes.string,
+    fileExists: PropTypes.bool,
     onChange: PropTypes.func,
 };
 FileAutoComplete.defaultProps = {
+    fileExists: true,
```
martinpitt, line 200: I find this property name a bit odd, but sure. ("mustExist" or invert it and "allowNotExist" ?)

martinpitt (approved):
Thanks!

pkg/lib/cockpit-components-file-autocomplete.jsx
KKoukiou: True, the property existed in the other component therefore I left it, but I will also would rather change it. I will repost.

pkg/lib/cockpit-components-file-autocomplete.jsx
```diff
@@ -80,6 +80,12 @@ export class FileAutoComplete extends React.Component {
         this.allowFilesUpdate = false;
     }
 
+    onCreateOption(newValue) {
```
KKoukiou, line 83: This also needs bind to this.

martinpitt (approved):
Cheers!

PR #15297, opened 2021-02-09
Make sure that cockpit.language is set also when that is english

KKoukiou:
cockpit.language was previously set inside cockpit.locale. That function
was only called for the languages which have a matching 'po' file.
English as the default does not have a po file and therefore
cockpit.language was not set for it. Similarly there was not cockpitLang
cookie when english was used.

### DIFF
--- a/pkg/base1/cockpit.js
+++ b/pkg/base1/cockpit.js
@@ -3905,10 +3905,10 @@ function factory() {
     var po_data = { };
     var po_plural;
 
-    cockpit.language = undefined;
+    cockpit.language = "en";
 
     cockpit.locale = function locale(po) {
-        var lang = cockpit.language || "en";
+        var lang = cockpit.language;
         var header;
 
         if (po) {


KKoukiou:
I think this change was introduced with the move to cockpit-po-plugin. This is also the reason behind our current image refreshes failures. See https://github.com/cockpit-project/bots/pull/1628

KKoukiou:
I am not sure if this the right fix, but I guess it should anyhow be a release blocker.

martinpitt:
Will review/think about this later, but please run tests already; especially against the refreshed image. On first sight I wouldn't know how this code would affect image refreshes? Or is it because the pre-installed cockpit on the image is being used by podman etc.?

marusak:
>  On first sight I wouldn't know how this code would affect image refreshes? Or is it because the pre-installed cockpit on the image is being used by podman etc.?

Yeah, new Fedora refresh has new cockpit version where this change is already present.

KKoukiou:
> > On first sight I wouldn't know how this code would affect image refreshes? Or is it because the pre-installed cockpit on the image is being used by podman etc.?
> 
> Yeah, new Fedora refresh has new cockpit version where this change is already present.

Right, the last fedora-33 update was https://github.com/cockpit-project/bots/pull/1556 which was two releases ago when this change was not present.

pkg/base1/cockpit.js
```diff
@@ -3905,7 +3905,7 @@ function factory() {
     var po_data = { };
     var po_plural;
 
-    cockpit.language = undefined;
+    cockpit.language = "en";
```
marusak, line 3908: It is being set on line 3928 (with default being picked on line 3911.
I am not sure if this is the right fix, as if you don't call `cockpit.locale()` (done in `pkg/lib/cockpit-po-plugin.js`) then you don't have any translation data really.

KKoukiou:
> Will review/think about this later, but please run tests already; especially against the refreshed image. On first sight I wouldn't know how this code would affect image refreshes? Or is it because the pre-installed cockpit on the image is being used by podman etc.?

I don't think we can test this really, apart from copying this code into the image that podman or certificates are using. And I don't think this can be done with tests-trigger.

pkg/base1/cockpit.js
KKoukiou: Correct, I also mention it the commit message. But we don't really need to call cockpit.locale for english since there is not any translation really

martinpitt:
> I don't think we can test this really,

Not against refreshed images indeed, but against our own tests at least.

martinpitt:
So here's what happens: Before https://github.com/cockpit-project/cockpit/commit/72c180d788 , we shipped a /usr/share/cockpit/base1/po.js.gz which is just a copy of po.empty.js (that contains an empty cockpit.locale() call for "en"). *that* is what provided a "po.en.js" (we never shipped that, but it's a fallback if cockpit-ws does not find a po.LL.js). After the po-plugin rework, we stopped shipping such a po.js, I didn't recognize that it was important.

pkg/base1/cockpit.js
martinpitt: So now that we know what's going on, we could ship a stub po.js again, or just take your fix, which seems a bit more elegant to me. But then please drop the default three lines below, to not duplicate it.

martinpitt (approved):
Thanks! I'd also appreciate @marusak 's review though.

marusak (approved):
Thanks, with the explanation how it used to work it makes sense.

PR #15298, opened 2021-02-09
Show busiest CPU cores

marusak:
Screenshot from 2021-02-09 13-55-02

Shows up to 16 cores ordered by their load. I picked 16 as it is a number that the UI manages just fine, is still human parsable (having list of 150 items is not) and standard user will see all their cores and on big servers it will be obvious that the list is shorter.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -27,6 +27,7 @@ import {
     Button,
     Card, CardTitle, CardBody, Gallery,
     DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
+    Flex, FlexItem,
     Page, PageSection,
     Progress, ProgressVariant,
     Select, SelectOption,
@@ -132,6 +133,9 @@ const CURRENT_METRICS = [
     { name: "network.interface.tx", units: "bytes", derive: "rate" },
     { name: "cgroup.cpu.usage", derive: "rate" },
     { name: "cgroup.memory.usage" },
+    { name: "cpu.core.user", derive: "rate" },
+    { name: "cpu.core.system", derive: "rate" },
+    { name: "cpu.core.nice", derive: "rate" },
 ];
 
 const HISTORY_METRICS = [
@@ -195,6 +199,7 @@ class CurrentMetrics extends React.Component {
             memUsed: 0, // GiB
             swapUsed: null, // GiB
             cpuUsed: 0, // percentage
+            cpuCoresUsed: [], // [ percentage ]
             loadAvg: null, // string
             disksRead: 0, // B/s
             disksWritten: 0, // B/s
@@ -323,6 +328,15 @@ class CurrentMetrics extends React.Component {
         newState.netInterfacesRx = this.samples[7];
         newState.netInterfacesTx = this.samples[8];
 
+        // Collect CPU cores
+        newState.cpuCoresUsed = [];
+        if (this.samples[11] && this.samples[11].length == this.samples[12].length && this.samples[12].length == this.samples[13].length) {
+            for (let i = 0; i < this.samples[11].length; i++) {
+                // CPU cores metrics are in ms/s; divide by 10 to get percentage
+                newState.cpuCoresUsed[i] = Math.round((this.samples[11][i] + this.samples[12][i] + this.samples[13][i]) / 10);
+            }
+        }
+
         // return [ { [key, value] } ] list of the biggest n values
         function n_biggest(names, values, n) {
             const merged = [];
@@ -390,20 +404,35 @@ class CurrentMetrics extends React.Component {
                 </Tooltip>);
         }
 
+        let cores = null;
+        if (this.state.cpuCoresUsed.length > 1) {
+            const top_cores = this.state.cpuCoresUsed.map((v, i) => [i, v]).sort((a, b) => b[1] - a[1])
+                    .slice(0, 16);
+            cores = top_cores.map(c => <Flex key={c[0]} justifyContent={{ default: 'justifyContentSpaceBetween' }}>
+                <FlexItem>{ cockpit.format(_("Core $0"), c[0]) }</FlexItem>
+                <FlexItem>{c[1]}%</FlexItem></Flex>
+            );
+        }
+
+        const cpu_usage = (
+            <Progress
+                id="current-cpu-usage"
+                value={this.state.cpuUsed}
+                className="pf-m-sm"
+                min={0} max={100}
+                variant={ this.state.cpuUsed > 90 ? ProgressVariant.danger : ProgressVariant.info }
+                title={ num_cpu_str }
+                label={ this.state.cpuUsed + '% ' } />);
+
         return (
             <Gallery className="current-metrics" hasGutter>
                 <Card id="current-metrics-card-cpu">
                     <CardTitle>{ _("CPU") }</CardTitle>
                     <CardBody>
                         <div className="progress-stack">
-                            <Progress
-                                id="current-cpu-usage"
-                                value={this.state.cpuUsed}
-                                className="pf-m-sm"
-                                min={0} max={100}
-                                variant={ this.state.cpuUsed > 90 ? ProgressVariant.danger : ProgressVariant.info }
-                                title={ num_cpu_str }
-                                label={ this.state.cpuUsed + '% ' } />
+                            {cores !== null ? <Tooltip content={ cores } position="bottom">
+                                {cpu_usage}
+                            </Tooltip> : cpu_usage }
                         </div>
 
                         { this.state.loadAvg &&
--- a/src/bridge/cockpitcpusamples.c
+++ b/src/bridge/cockpitcpusamples.c
@@ -24,6 +24,8 @@
 #include <stdio.h>
 #include <unistd.h>
 
+#define CPU_CORE_MAXLEN 8
+
 gint cockpit_cpu_user_hz = -1;
 
 static gint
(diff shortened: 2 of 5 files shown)


src/bridge/cockpitcpusamples.c
```diff
@@ -71,28 +71,45 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+      gchar cpu_core[8];
+      gchar core[5];
 
-      if (!(g_str_has_prefix (line, "cpu ")))
+      if (!(g_str_has_prefix (line, "cpu")))
         continue;
 
       #define FMT64 "%" G_GUINT64_FORMAT " "
-      if (sscanf (line + sizeof ("cpu ") - 1, FMT64 FMT64 FMT64 FMT64 FMT64,
+      if (sscanf (line, "%s " FMT64 FMT64 FMT64 FMT64 FMT64,
```
martinpitt, line 81: Let's please limit the string length, otherwise `cpu_core` could possibly overflow the stack. `%8s` should do the trick. To avoid duplicating the length, try something like
```c
#define CPU_CORE_MAXLEN 8
gchar cpu_core[CPU_CORE_MAXLEN]
sscanf(..., "%" #CPU_CORE_MAXLEN "s ..."
```

src/bridge/cockpitcpusamples.c
```diff
@@ -71,28 +71,45 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
 
       user_hz = ensure_user_hz ();
-      cockpit_samples_sample (samples, "cpu.basic.nice", NULL, nice*1000/user_hz);
-      cockpit_samples_sample (samples, "cpu.basic.user", NULL, user*1000/user_hz);
-      cockpit_samples_sample (samples, "cpu.basic.system", NULL, system*1000/user_hz);
-      cockpit_samples_sample (samples, "cpu.basic.iowait", NULL, iowait*1000/user_hz);
-      break;
+      if (strlen (cpu_core) > 3)
+        {
+          if (sscanf (cpu_core, "cpu%s", core) != 1)
```
martinpitt, line 96: You could keep the "cpu" prefix in the initial sscanf(), then you only have to deal with "empty string" or "a string number", and can drop the "3". Then you can case-differentiate with `!cpu_core[0]`. Then  this sscanf reduces to `atoi(cpu_core)`.

src/bridge/cockpitcpusamples.c
```diff
@@ -71,28 +71,45 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+          cockpit_samples_sample (samples, "cpu.core.user", core, user*1000/user_hz);
+          cockpit_samples_sample (samples, "cpu.core.system", core, system*1000/user_hz);
+          cockpit_samples_sample (samples, "cpu.core.iowait", core, iowait*1000/user_hz);
+        }
+      else
+        {
+          cockpit_samples_sample (samples, "cpu.basic.nice", NULL, nice*1000/user_hz);
+          cockpit_samples_sample (samples, "cpu.basic.user", NULL, user*1000/user_hz);
+          cockpit_samples_sample (samples, "cpu.basic.system", NULL, system*1000/user_hz);
+          cockpit_samples_sample (samples, "cpu.basic.iowait", NULL, iowait*1000/user_hz);
```
martinpitt, line 111: If you feel like it, please add a smoketest to src/bridge/test-metrics.c similar to `test_cgroup` -- you can check that the calculated rate is in a sensible range between 0 and 1000 (msec/s). This is usually a great helper for development, but that part is already done now. :grin: I'm also happy to send a PR for that as a follow-up (I really like writing these)

pkg/metrics/metrics.jsx
```diff
@@ -390,20 +404,33 @@ class CurrentMetrics extends React.Component {
...
                         <div className="progress-stack">
-                            <Progress
-                                id="current-cpu-usage"
-                                value={this.state.cpuUsed}
-                                className="pf-m-sm"
-                                min={0} max={100}
-                                variant={ this.state.cpuUsed > 90 ? ProgressVariant.danger : ProgressVariant.info }
-                                title={ num_cpu_str }
-                                label={ this.state.cpuUsed + '% ' } />
+                            {this.state.cpuCoresUsed.length
```
martinpitt, line 430: I'd change this to `> 1`, it doesn't make much sense to do that on 1-core machines (which are still very common for VMs). Then you  could also conditionalize the computation of `cores` and skip it on single-core machines.

test/verify/check-metrics
```diff
@@ -626,6 +626,15 @@ class TestMultiCPU(MachineCase):
...
         self.assertLessEqual(getMaximumSpike(b, "cpu", False, 1598968800000, 45), 1.0)
 
+        # Test current usage of cores
+        b.wait(lambda: progressValue(b, "#current-cpu-usage") < 20)
+        m.execute("systemd-run --collect --slice cockpittest -p CPUQuota=60% --unit cpu-hog dd if=/dev/urandom of=/dev/null")
+        m.execute("systemd-run --collect --slice cockpittest -p CPUQuota=30% --unit cpu-piglet dd if=/dev/urandom of=/dev/null")
+        b.wait(lambda: progressValue(b, "#current-cpu-usage") > 40)
+        b.mouse("#current-cpu-usage", "mouseenter")
+        b.wait(lambda: int(b.text(".pf-c-tooltip .pf-l-flex:first-child div:nth-child(2)")[:-1]) > 50)
+        b.wait(lambda: int(b.text(".pf-c-tooltip .pf-l-flex:first-child div:nth-child(2)")[:-1]) > 20)
```
martinpitt, line 636: ... and perhaps that the "total" CPU usage is > 40 and < 50? (average between the two cores)

martinpitt (changes requested):
Great job on figuring out the metrics! My usual round of nitpicks, but this looks like a great start.

I had a second progress bar in mind with just the most used individual core, for easier visual comparison and better discoverability. But we can add that later, too.

src/bridge/cockpitcpusamples.c
marusak: yeah, that was my initial idea but I was not able to write format string that would accept "possibly empty string".
The format is either "cpu  1 2 3" or "cpu14 1 2 3" and anything I came up with has then eaten the first value in the first case as core number. Any advice is welcomed :)

martinpitt: I see -- let's keep it as-is then. Thanks!

src/bridge/cockpitcpusamples.c
martinpitt: For my own future self: This can be done with [G_STRINGIFY](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-STRINGIFY:CAPS). The `#` operator only works inside macros.

allisonkarlitskaya:
```
#include <stdio.h>

int
main (void)
{
  char x[5] = "aaaa";

  sscanf ("abcd", "%3s", x);
  printf("%s\n", x);

  return 0;
}
```

```
toolbox:~$ ./x
abc
```

ie: `%3s` means "3 characters, plus nul" (ie: 4 characters).  Make sure to add 1 to the buffer to compensate.

allisonkarlitskaya (changes requested):

test/verify/check-metrics
marusak: I had `> 40` and I added `< 50` but the first run had some with `51%`. This smells like it might make this test flaky, but lets see.

marusak: nah, now the machine was somehow busy and it was 94%. I'll drop it.

martinpitt: Right, that was a thinko -- the CPU is probably busy with stuff other than these two synthetic CPU eater services.

src/bridge/cockpitcpusamples.c
```diff
@@ -71,28 +73,45 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
 
       user_hz = ensure_user_hz ();
-      cockpit_samples_sample (samples, "cpu.basic.nice", NULL, nice*1000/user_hz);
-      cockpit_samples_sample (samples, "cpu.basic.user", NULL, user*1000/user_hz);
-      cockpit_samples_sample (samples, "cpu.basic.system", NULL, system*1000/user_hz);
-      cockpit_samples_sample (samples, "cpu.basic.iowait", NULL, iowait*1000/user_hz);
-      break;
+      if (strlen (cpu_core) > 3)
+        {
+          if (sscanf (cpu_core, "cpu%s", core) != 1)
```
martinpitt, line 98: Oh, just now I realized that this doesn't parse it into an integer, but a string. If that was by intent (indexing the series by string names instead of ints), then we don't need that sscanf() at all -- just `core+3` is fine to chop off the "cpu" prefix.

pkg/metrics/metrics.jsx
```diff
@@ -390,20 +404,33 @@ class CurrentMetrics extends React.Component {
                 </Tooltip>);
         }
 
+        const top_cores = this.state.cpuCoresUsed.map((v, i) => [i, v]).sort((a, b) => b[1] - a[1])
+                .slice(0, 16);
+        const cores = top_cores.map(c => <Flex key={c[0]} justifyContent={{ default: 'justifyContentSpaceBetween' }}>
+            <FlexItem>{ cockpit.format(_("Core $0"), c[0]) }</FlexItem>
+            <FlexItem>{c[1]}%</FlexItem></Flex>
+        );
```
martinpitt, line 412: This could be conditionalized on `this.state.cpuCoresUsed.length > 1`, on single-core it's just wasted electrons.

test/verify/check-metrics
martinpitt, line 636: Is that some copy&paste error? don't you mean `nth-child(1)` in the first line? These selectors look exactly identical right now.

martinpitt (changes requested):
Thanks!

src/bridge/cockpitcpusamples.c
marusak: We need string as later on the name must be string and not int. And indeed, `+3` would do, I was playing too much with `sscanf` that I wanted to use it for everything :)

martinpitt (approved):
Cheers!

Issue #15283, opened 2021-02-04
Radio buttons in 237 not aligned with text (Automatic updates card)


AdamWill:
Cockpit version: 237
OS: Fedora
Page: Software Updates

In Fedora openQA testing, the cockpit updates functionality test fails on cockpit 237 because the radio buttons on the "Automatic updates" card are clearly misaligned with the text:
![offbuttons](https://user-images.githubusercontent.com/916551/106942018-7eb9b180-66d8-11eb-8ee5-936d597e6acb.png)
I can make the test pass by updating the reference screenshot, but this does seem like a real bug.

PR #15299, opened 2021-02-09
lib/pf: Unset PF3 margin for PF4 radio/checkbox

garrett:
This should hopefully keep both PF3 and PF4 in check and let us have the correct spacing for each.


KKoukiou (approved):
Fixes #15283

PR #15301, opened 2021-02-09
build system overhaul party PR

allisonkarlitskaya:
This is never going to land.  I'll start breaking it out into smaller bits tomorrow.

### DIFF
--- a/pkg/lib/stampfile-plugin.js
+++ b/pkg/lib/stampfile-plugin.js
@@ -1,0 +1,44 @@
+const fs = require('fs');
+const assert = require('assert');
+
+module.exports = class {
+    apply(compiler) {
+        compiler.hooks.afterEmit.tap('StampfilePlugin', compilation => {
+            const stamps = { };
+
+            for (const output in compilation.assets) {
+                const path = compilation.assets[output].existsAt;
+
+                /* `output` will be like
+                 *
+                 *   base1/cockpit.js
+                 *
+                 * `path` will be like
+                 *
+                 *   /build/dir/dist/base1/cockpit.js
+                 *
+                 * We want to get the full absolute path, up to and
+                 * including the first directory component of `output`.
+                 */
+                assert(path.startsWith("/"));
+                assert(path.endsWith(output));
+
+                const first_slash = output.indexOf('/');
+                const dir_length = path.length - output.length + first_slash;
+                const dir = path.slice(0, dir_length);
+
+                /* As per the above example, `dir` would be
+                 *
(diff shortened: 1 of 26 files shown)


pkg/lib/stampfile-plugin.js
```diff
@@ -0,0 +1,44 @@
...
+                 *   /build/dir/dist/base1
+                 *
+                 * Let's put a stamp there.
+                 */
+                stamps[dir + '/stamp'] = true;
+            }
+
+            for (const stampfile in stamps) {
+                fs.writeFileSync(stampfile, '');
+            }
```
martinpitt, line 41: Wouldn't it be simpler to directly call writeFileSync() in the first loop, and drop the `stamps` object?

martinpitt (commented):
I have some trouble to see where this is going. It feels like having a new plugin that just writes stamp files is a bit overkill? E.g. eventually we want to move building of manifest.json to webpack, and then the manifests would IMHO be a more appropriate "stamp", and we could just ditch the stamp files completely. Building manifests is probably also going to be a plugin, but I don't think we want two?

pkg/lib/stampfile-plugin.js
allisonkarlitskaya: The first loop is going to encounter very many files within this subdirectory, so we'd end up writing the stamp file over and over again...

allisonkarlitskaya: The comment about the manifest.json acting as a stamp is reasonable...  we used to use the Makefile.deps for this purpose, in fact...

allisonkarlitskaya: For the record about "where this is going": the idea is eventually that typing `webpack` will produce exactly the same output in `dist/` as if we typed `make`.

martinpitt: Loop> make sense, sorry for misreading that.

That's a good goal -- I rather meant whether the stamp file plugin is meant to be a temporary thing or a lasting (part of) the final solution.

martinpitt (approved):
OK, let's go with that for now, if it helps you to break the webpack cleanups into smaller steps.  I can still bicker about "stamp and manifest plugin are 90% identical" later on, and we can simplify things once everything is working. Thanks!

allisonkarlitskaya:
Most of the important bits of this landed elsewhere.  All that's really left are the trivial changes to `WEBPACK_INPUTS` and `WEBPACK_OUTPUTS`.  Those can be handled elsewhere, since this PR is a big and rather outdated mess.

PR #15302, opened 2021-02-09
build: Drop gzipped dist tarballs and make-source usage

martinpitt:
Since the introduction of `$XZ_COMPRESS_FLAGS` in commit 67831b9d2f and
supressing the build of cockpit-cache-* tarballs in commit 3482db3f5
there is no reason to build a .tar.gz dist any more, as `xz -0` is even
faster than gzip.

Drop the `dist-gzip` make target, adjust the "dump file name" target for
xz, and switch the .spec and image-prepare to using the tar.xz
targets/tarballs.

bots/make-source defaults to `dist-gzip`, is rather hard to work with
for dist-xz even with a custom target, and most of it is a big cockpit
special case (and not used in any other project). Let's phase it out,
and just do the few build steps ourselves -- these are project specific
anyway, so let's not pretend we can do an "one size fits it all" script.
While at it, replace make-source's static `-j8` with a dynamic CPU
count.

This will unblock building images straight from the pre-generated dist
tarballs in GitHub action artifacts.

 - [x] Fix bots debian.install to get along with tar.xz: https://github.com/cockpit-project/bots/pull/1658

FTR: I *may* want to split this back out into a tools/make-source (python,not shell) when I add the logic to try and download a pre-built dist tarball from GitHub. That is then probably more reusable for local development than as part of image-prepare. But I'd like to do that as part of the "download artifacts PR, not here.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -358,9 +358,9 @@ install-data-local:: dist/guide/html/index.html
 uninstall-local::
 	rm -rf $(DESTDIR)$(htmldir)
 
-# See bots/make-source
-dump-dist-gzip:
-	@echo "$(abs_builddir)/$(distdir).tar.gz"
+# See test/image-prepare
+dump-dist:
+	@echo "$(abs_builddir)/$(distdir).tar.xz"
 
 # Subdirectories to distribute everything that's committed to git
 COMMITTED_DIST = \
@@ -390,9 +390,6 @@ DIST_ARCHIVES = \
 # used from containers/unit-tests/run.sh
 XZ_COMPRESS_FLAGS = --extreme
 
-dist-gzip: distdir
-	$(DIST_TAR_MAIN) | GZIP=$(GZIP_ENV) gzip -c > $(distdir).tar.gz
-	find "$(distdir)" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "$(distdir)"
 dist-xz: distdir
 	$(DIST_TAR_MAIN) | xz $(XZ_COMPRESS_FLAGS) > $(distdir).tar.xz
 	[ -n "$$NO_DIST_CACHE" ] || $(DIST_TAR_CACHE) | xz $(XZ_COMPRESS_FLAGS) > cockpit-cache-$(VERSION).tar.xz
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -18,6 +18,7 @@
 
 import argparse
 import errno
+import multiprocessing
 import os
 import sys
 import subprocess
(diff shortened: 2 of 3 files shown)


martinpitt:
Hmm, two things: (1) this makes image-prepare much slower as it will now use `--extreme` by default. And (2) the whole make-source dance is unnecessarily complicated.

martinpitt:
Bwah, [image-prepare in an unbuilt tree](https://logs.cockpit-project.org/logs/pull-15302-20210209-163227-5368efaa-fedora-32/log.html) behaves differently than in a built tree. It'd perhaps be better to drop make-source altogether..

martinpitt:
OK, I took the extra bit of effort, the hackery wasn't worth it. Let's kill make-source!

martinpitt:
Current bots debian.install [assumes .tar.gz](https://logs.cockpit-project.org/logs/pull-15302-20210209-171641-5037d961-debian-stable/log.html), so I sent https://github.com/cockpit-project/bots/pull/1658 to fix that.

I trigger a run against that bots here to demonstrate that it works.

martinpitt:
@allisonkarlitskaya : In my [download-dist branch](https://github.com/martinpitt/cockpit/commit/0e858669fbc0f8b003936fb374769994cedcd3a0) (WIP) I split this out into test/make_dist.py. Please let me know if you'd rather have me squash that in here.

test/image-prepare
```diff
@@ -152,13 +153,23 @@ def build_and_install(build_image, build_results, skips, args):
                                  memory_mb=2048, cpus=4, maintain=True, networking=networking)
     completed = False
 
-    # While the machine is booting, make a tarball
+    # While the machine is booting, make a dist tarball
+
+    # we need to support completely clean git trees, unpacked release tarballs, and already configured trees
+    if not os.path.exists("Makefile"):
+        if os.path.exists('.tarball'):
```
allisonkarlitskaya, line 160: why not `./configure` if it exists, otherwise `./autogen.sh`?

Makefile.am
```diff
@@ -390,9 +390,6 @@ DIST_ARCHIVES = \
 # used from containers/unit-tests/run.sh
 XZ_COMPRESS_FLAGS = --extreme
 
-dist-gzip: distdir
```
allisonkarlitskaya, line 393: yay!!

test/image-prepare
```diff
@@ -152,13 +153,23 @@ def build_and_install(build_image, build_results, skips, args):
...
+
     # Do a development build on the latest Fedora image so that we can spot React and other dev-only warnings
     env = os.environ.copy()
     if os.path.basename(build_image) == TEST_OS_DEFAULT:
         env["NODE_ENV"] = "development"
-    source = subprocess.check_output([os.path.join(BOTS, "make-source")],
-                                     universal_newlines=True, env=env).strip()
+    # this is for a development build, not a release, so we care about speed, not best size
+    subprocess.check_call(["make", "--silent", "-j%i" % multiprocessing.cpu_count(),
+                           "NO_DIST_CACHE=1", "XZ_COMPRESS_FLAGS=-0", "dist"], env=env)
```
allisonkarlitskaya, line 171: i'm finding myself typing `make NO_DIST_CACHE=1 XZ_COMPRESS_FLAGS=-0` a lot lately.  I wonder if we want a separate rule for that...

On the other hand, I wonder if that's something we're gonna completely stop caring about once this all lands...

Let's wait :)

Makefile.am
```diff
@@ -358,9 +358,9 @@ install-data-local:: dist/guide/html/index.html
 uninstall-local::
 	rm -rf $(DESTDIR)$(htmldir)
 
-# See bots/make-source
-dump-dist-gzip:
-	@echo "$(abs_builddir)/$(distdir).tar.gz"
+# See test/image-prepare
+dump-dist-xz:
```
allisonkarlitskaya, line 362: Just call this `dump-dist`.

allisonkarlitskaya (changes requested):

test/image-prepare
martinpitt: I just copied that from bots/make-source, but indeed, why not.

Makefile.am
martinpitt: Done.

test/image-prepare
martinpitt: Done.

allisonkarlitskaya (approved):
Looks good :)

PR #15311, opened 2021-02-11
Replace 'restart' with 'reboot' when related to a system reboot

### LINKED ISSUE #15273: "restart" should specify what it is restarting; if it's restarting the computer, it should be "reboot"
As discussed in https://github.com/cockpit-project/cockpit/pull/14581:

Where we might say "restart", we should specify what we're restarting.

"Restart" (as we have it on the overview page) could mean "restart computer", "restart services", "restart the connection to Cockpit", "restart Cockpit", and "start over".

Just having the word "reboot" by itself is much clearer.

skobyda:
This should make clearer distinction between 'restart' (often used in relation to a service restart) and a 'reboot' (used in relation to a system reboot)

Fixes #15273

### DIFF
--- a/doc/guide/feature-systemd.xml
+++ b/doc/guide/feature-systemd.xml
@@ -55,7 +55,7 @@ Africa/Algiers
     <code>ntpd</code> can cause the <code>timedated</code> daemon to
     behave inconsistently with regards to time synchronization.</para>
 
-  <para>Cockpit restarts or powers down the machine by using the
+  <para>Cockpit reboots or powers down the machine by using the
     <ulink url="https://www.freedesktop.org/software/systemd/man/shutdown.html"><code>shutdown</code></ulink>
     command. To perform similar tasks from the command line, run it directly:</para>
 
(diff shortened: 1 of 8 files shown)


martinpitt (changes requested):
Thanks! Fortunately we already have "Reboot" as translated string, so it won't break too many translations. In favor of the change, but still needs a bunch of test adjustments.

skobyda:
Looking at tests, there are failures which seem unrelated (or I cannot think of how they relate to a simple string ammendation). @martinpitt have you seen these failures before?

martinpitt:
@skobyda : They mostly look familiar, yes. See our [weather report](https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&days=7). Retrying.

martinpitt (approved):
Thanks, looks good! Please land once everything is green.

martinpitt:
For some reason [rhel-8-4-distropkg](https://logs.cockpit-project.org/logs/pull-15311-20210216-083547-e538dadc-rhel-8-4-distropkg/log.html) is cursed on this PR. That's like the 5th time that it fails.

marusak:
There is probably no point in retrying this anymore. If we are sure it is unrelated error, lets just merge it as is.

martinpitt:
At least fedora-32 succeeded at last.

PR #15312, opened 2021-02-11
Delete COPYING

nbej:
Bbb


martinpitt (changes requested):
Is that some kind of prank? Or some weird "I need to collect fake PRs to get a T-Shirt" contest?

Not cool..

martinpitt:
https://github.com/nbej is clearly a spam account, opened two days ago and sent nothing but garbage. I reported the user to GitHub for abuse.

PR #15313, opened 2021-02-11
download GitHub artifact for image-prepare

martinpitt:
This happens if

 - there is a GitHub token (GitHub artifacts download needs to be authenticated)
 - the current git HEAD's SHA has an already built dist artifact, i.e. you checked out a revision from master or a fork/branch from a PR
 - you are requesting a production (not development) build.

That replaces a 10 minute npm install+webpack run with a few seconds download, which will make it much nicer to locally test a PR or current master version: You can just check out what you want to test and run `test/image-prepare -q` to get a test VM in 3 minutes.

This also gets us very close to just unpacking parts of the tarball into the checkout, so that you can do local development without having to initially rebuild all the webpacks. This still requires a few build system cleanups (which @allisonkarlitskaya is working on), but at least we have the VM testing use case sorted out.

 - [x] Requires dropping dist-gzip tarballs, as our artifacts build xz: PR #15302
 - [x] Possibly hard-code `GitHub(repo="cockpit-project/cockpit")` to only ever download artifacts from there, even when running in forks? Nicer for developers,  but a bit harder to test artifact changes on forks -- but that happens rarely. Could still respect an explicit `$GITHUB_BASE` to get both.
 - [x] Extract node_modules, so that tests can work
 - [x] Fix [selenium tests requiring axe](https://logs.cockpit-project.org/logs/pull-15313-20210211-160107-b9ff9069-fedora-32-selenium-chrome/log.html) (either by including it by default, or by dropping the dependency)


martinpitt:
## Initial round of tests
This happened in parallel with the "build" workflow, so there was no artifact available yet. The [debian-stable run](https://logs.cockpit-project.org/logs/pull-15313-20210211-065938-fa8c931d-debian-stable/log.html) correctly says this:

    make_dist: no download available for commit 34463bb482369d8cdbc7f51d26237e073f04394a

The [fedora-33](https://logs.cockpit-project.org/logs/pull-15313-20210211-065935-fa8c931d-fedora-33/log.html) run did not attempt a download, as that runs in devel mode.

## Test retry

I force-retried the tests, and [debian-stable](https://logs.cockpit-project.org/logs/pull-15313-20210211-075852-fa8c931d-debian-stable/log.html) still rebuilt itself because of the rebasing (another PR just landed). I think we want to keep this behaviour, correctness > speed.

So I need to re-do this exercise on a rebased branch. The dependency #15302 just landed, so I pushed the rebase.

martinpitt:
This is working well enough now that I think it's ready for review. It does not work too well now for local builds, due to some `npm install` weirdness (it seems to forget to install some files), but as developers are not yet used to this workflow anyway, I think that would be better as a follow-up. The main goal for now was to get from clean checkout to prepared image in as little time as possible.

This will also unblock triggering the tests from the build workflow instead of through the webhook, unblock creating a /livetest action for cockpit (hello @KKoukiou), and make it generally easier for us to review PRs. For the latter we certainly still need some convenience wrappers.

martinpitt:
Whoops, [selenium tests fail](https://logs.cockpit-project.org/logs/pull-15313-20210211-160107-b9ff9069-fedora-32-selenium-chrome/log.html), they need axe. I'll see to dropping this dependency, or including it into disted node_modules.

I would still appreciate an initial review from all of you how you generally like this workflow, and whether it works for you.

martinpitt:
First run was green (except for some flakes on rhel-8-4-distropkg). Doing a full re-run to test all contexts against existing dist tarball download.

martinpitt:
Argh! Something else landed on master again, so I need to rebase to really test the download workflow. This time I'll do that differently though, with just one run (no-test and tests-scan once the dist is published)

martinpitt:
Nice! All tests except fedora-33 now downloaded the dist tarball and jumpstarted. Ready to review for real!

test/make_dist.py
```diff
@@ -47,5 +50,104 @@ def make_dist(devel_mode):
...
+                sys.stdout.write("\r                             \r")
+
+    except urllib.error.HTTPError as e:
+        print("make_dist: Download failed:", e)
+        return None
+
+    with zipfile.ZipFile(zipio) as fzip:
+        names = fzip.namelist()
+        if len(names) != 1 or not names[0].endswith(".tar.xz"):
+            print("make_dist: expected zip artifact with exactly one tar.xz member")
```
KKoukiou, line 137: Hmm was this for your debugging or it can really happen?

KKoukiou (approved):
Perfect! I had only one question, but nothing blocking this - everything else is very clear. 

```
$ time test/image-prepare fedora-33
make_dist: Downloading dist tarball from https://api.github.com/repos/cockpit-project/cockpit/actions/artifacts/40631660/zip ...
...
real	3m7.926s
user	0m0.870s
sys	0m0.714s
```

test/make_dist.py
martinpitt: It Should Not Happen™ right now, but if we ever change our workflow to e.g. include cockpit-cache, this will then tell us why the download is not working (i.e. we need to adjust make_dist to the new tarball structure)

martinpitt:
@allisonkarlitskaya : I'd still like to get your review as well, as this directly touches what you are working on.

test/make_dist.py
```diff
@@ -135,7 +136,23 @@ def download_dist():
         if len(names) != 1 or not names[0].endswith(".tar.xz"):
             print("make_dist: expected zip artifact with exactly one tar.xz member")
             return None
-        return fzip.extract(names[0])
+        tar_path = fzip.extract(names[0])
+
+    # Extract node_modules and dist locally for speeding up the build and allowing integration tests to run
+    # unfortunately it is impossible with tar CLI and Python's tarfile to just extract that subset;
+    # so extract it all, pick what we want, and clean it up
+    unpack_dirs = [d for d in ["dist", "node_modules"] if not os.path.exists(d)]
```
allisonkarlitskaya, line 144: This is not necessary.  After playing around a bunch I was able to get this working with just plain `tar`.  You need to include the directory name of the archive, like `cockpit-222/dist/` and then it should work...

Also: `tar -m` is your friend.

test/image-prepare
```diff
@@ -46,6 +46,7 @@ def main():
     parser.add_argument('-q', '--quick', action='store_true', help='Build faster')
     parser.add_argument('-f', '--force', action='store_true', help='Force update of images')
     parser.add_argument('-b', '--build-image', action='store', help='Build in this image')
+    parser.add_argument('--devel-build-on-default-os', action='store_true', help='Do a development build on default TEST_OS')
```
allisonkarlitskaya, line 49: couldn't we just make this a (shorter named) flag which can work with any OS image?  this argument is practically screaming "i'm not meant to be used by any human", and i think humans might indeed sometimes want to do debug builds for local manual testing...

test/make_dist.py
```diff
@@ -47,5 +50,104 @@ def make_dist(devel_mode):
...
+    print(f"make_dist: Downloading dist tarball from {download_url} ...")
+    request = urllib.request.Request(download_url, headers={"Authorization": "token " + api.token})
+    zipio = io.BytesIO()
+    try:
+        with urllib.request.urlopen(request) as response:
+            if os.isatty(sys.stdout.fileno()):
+                total_size = 0
+            else:
+                total_size = None
+            MB = 2**20
```
allisonkarlitskaya, line 114: i think you meant `1e6` :)

allisonkarlitskaya (changes requested):

test/image-prepare
martinpitt: Exactly, it is *not* meant for humans. A human should use `NODE_ENV=development` or `./configure --enable-debug`. I don't see a reason why a human would use this half-automagic unintuitive thing. I just found that argparse supports `help=argparse.SUPPRESS` to not show the option at all, I think I'll use that.

test/make_dist.py
martinpitt: I couldn't make it work without already knowing the directory name in advance, as the pattern matching is too inflexible (`*` always matches across `/`). That would mean either to call `tar -tf` first (duplicating the xunzip, no thanks) or deriving it from the file name (which seems a bit brittle).

test/make_dist.py
martinpitt: That's a float. But I can give you 10**6 :grin: (downloading in MB or MiB blocks should make no practical difference)

test/image-prepare
martinpitt: I reworked this to put the logic entirely in test/run, where it IMHO fits best. This makes image-prepare and make_dist simpler also.

test/make_dist.py
martinpitt: I added `--touch` (I prefer long options in code for easier understanding), but I kept the unpack logic for now. We can discuss a better solution here or on IRC

allisonkarlitskaya: Deriving from the filename is completely legit.  automake literally has a variable called `$(distdir)` and it names the tarball based on the value of this variable.

martinpitt: Done, thanks!

test/make_dist.py
allisonkarlitskaya: It's entirely about what we show the user

allisonkarlitskaya (approved):

allisonkarlitskaya (approved):

PR #15320, opened 2021-02-12
test: Test output from failed tests

marusak:
This test checks fix introduced in 7a790f2.

Question: I would like to also test retries for changed test files and changed source files. My idea is to change for example `FITNESS FOR A PARTICULAR PURPOSE` string in copyright in one test file (`test/verify/check-example`) and in one source file (maybe users or so) and then check it is retried 3 times. Possibly also testing these retries for tests that fail or tests that don't want these retries. But this involves changing files on the machine that runs tests. Is that a good practice?

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -212,7 +212,7 @@ def run(opts, image):
     test_files = glob.glob(os.path.join(opts.test_dir, "check-*"))
     if os.path.exists(os.path.join(os.path.dirname(testvm.TEST_DIR), ".git")):
         # Detect affected tests from changed test files
-        cmd = ["git", "diff", "--name-only", "HEAD", "origin/master", opts.test_dir]
+        cmd = ["git", "diff", "--name-only", "origin/master", opts.test_dir]
         r = subprocess.run(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
         if r.returncode == 0:
             changed_tests = [test.decode("utf-8") for test in r.stdout.strip().splitlines()]
@@ -225,7 +225,7 @@ def run(opts, image):
         # Detect affected tests from changed pkg/* subdirectories
         # If affected tests get detected from pkg/* changes, don't apply the
         # "only do this for max. 3 check-* changes" (even if the PR also changes ≥ 3 check-*)
-        cmd = ["git", "diff", "--name-only", "HEAD", "origin/master", "pkg/"]
+        cmd = ["git", "diff", "--name-only", "origin/master", "pkg/"]
         r = subprocess.run(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
         if r.returncode == 0:
             changed_pkgs = set(pkg.decode("utf-8").split('/')[1] for pkg in r.stdout.strip().splitlines())
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -57,8 +57,8 @@ def flush_stdout():
             time.sleep(0.1)
 
 
-def print_test(test, print_tap=True):
-    for line in test.output.splitlines(keepends=True):
+def print_test(test, print_tap=True, retry_reason=""):
+    for line in test.output.strip().splitlines(keepends=True):
         while line:
             try:
                 sys.stdout.buffer.write(line)
@@ -69,17 +69,20 @@ def print_test(test, print_tap=True):
     flush_stdout()
 
     if not print_tap:
+        print(retry_reason)
         return
 
+    print()  # Tap needs to start on a separate line
     if test.process.returncode == 0:
-        print("ok " +  test_name(test))
+        print("ok " +  test_name(test) + retry_reason)
     elif test.process.returncode == 77 or b"# SKIP " in test.output:
         # If the test was skipped, add the last line (which contains the reason
         # for the skip) to the result
-        print("ok {0} {1}".format(test_name(test),
-                                  test.output.splitlines()[-1].strip().decode() if test.process.returncode == 77 else ""))
+        print("ok {0} {1}{2}".format(test_name(test),
+                                  test.output.splitlines()[-1].strip().decode() if test.process.returncode == 77 else "",
+                                  retry_reason))
     else:
-        print("not ok " + test_name(test))
+        print("not ok " + test_name(test) + retry_reason)
     flush_stdout()
 
 def finish_test(opts, test, affected_tests):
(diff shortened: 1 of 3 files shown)


martinpitt:
Wrt. your question, I think it'd be a bit unfriendly to change files on-disk. Our CI doesn't mind, but human developers do. But copying the original file to some check-examples-dynamictest and adding that to .gitignore would be fine IMHO. WDYT?

marusak:
> I think it'd be a bit unfriendly to change files on-disk. Our CI doesn't mind, but human developers do.

agree.

> But copying the original file to some check-examples-dynamictest and adding that to .gitignore would be fine IMHO. WDYT?

Thats a good idea! I'll try that out.

marusak:
> Thats a good idea! I'll try that out.

Ah, won't work as we do "git diff" to figure out which files have changed. Adding a new file (and especially if in .gitignore) will not show up.

So what about cleaning after itself? It is not very likely and often that someone would run this test locally. And if they would, it would cleanup after itself.

martinpitt:
@marusak : Ah, it requires `git diff`, I see. That's fine then, if you do that with `finally:`. But please note that you have to guard that entire thing on `os.path.exists(.git)`, so that it doesn't break in release tarballs.

test/verify/check-testlib
```diff
@@ -103,6 +103,23 @@ class TestRunTest(MachineCase):
...
+        env = os.environ.copy()
+        try:
+            env["TEST_FAILURES"] = "1"
+            del env["TEST_JOBS"]
+        except KeyError:
+            pass
+        process = subprocess.run([run_tests, "--test-dir", VERIFY_DIR, "TestExample.testFail", "TestExample.testSkip"],
+                                 env=env, capture_output=True)
+        stdout = process.stdout
+        self.assertRegex(stdout, b"\nnot ok 1 .*test\/verify\/check-example TestExample.testFail # RETRY 1 \(be robust against unstable tests\)\n")
```
marusak, line 116: strictly speaking, this should not work because this PR affects `test/verify/check-example` and therefore it should not be retried. I have fix for this. The only way we can get this in is by sending changes into `test/verify/check-example` as separate PR and then implementing these checks. I'll do that.

marusak: oh man, I can now send PR with those changes, then deliver this test and it will work. But if anyone in the future touches `test/verify/check-example` it won't work. I have one more idea, /ME goes to try it out.

marusak: oh no, I would have to do `git commit` on ones machine, which I really don't want to do. We actually have `git` in VMs, so I am now thinking about copying `test/` into VM where I create simple git repo for it and then can test it properly without being afraid of breaking ones files or commits.

marusak: ok, that would be massive task, it also needs bots, then we would need to create set of tests that don't need VMs...
Maybe we can just simply say to our run-tests to never consider `test-example` as affected file. We won't loose anything by doing that and we can then happily use it for testing. Then of course I need to change different file for testing of "changed files are retried more time" but that is not a problem, as if it was not changed in the PR, we will fake some change and if it was, then we don't even need to fake change (but we as well might to keep it simple).

This is not a Friday after lunch topic it seems :)

marusak:
No there yet, if I take this as tag it `origin/master` then it does not work, which means it would not work in any future PR. I'm looking, thought that I had this one figured out.

test/verify/check-testlib
```diff
@@ -105,6 +106,84 @@ class TestRunTest(MachineCase):
...
         self.assertNotIn(b"TestExample", out)
 
+    def testRetry(self):
+        # Don't test this if not in git repo as we use `git diff` for this logic
+        if not os.path.exists(os.path.join(ROOT_DIR, ".git")):
+            return
+
+        env = os.environ.copy()
+        try:
+            env["TEST_FAILURES"] = "1"
```
martinpitt, line 116: This should not be part of the `try:`

test/verify/check-testlib
```diff
@@ -105,6 +106,84 @@ class TestRunTest(MachineCase):
...
+            # Changed test is never retried
+            self.assertRegex(stdout, b"\nnot ok 2 .*test\/verify\/check-examples TestRetryExample.testFail\n")
+            self.assertNotRegex(stdout, b"testFail # RETRY")
+
+            # Using @no_retry_when_changed prevents this retry logic
+            self.assertRegex(stdout, b"\nok 3 .*test\/verify\/check-examples TestRetryExample.testNoRetry\n")
+            self.assertNotRegex(stdout, b"testNoRetry # RETRY")
+        finally:
+            with open(os.path.join(VERIFY_DIR, "check-examples"), 'w') as f:
+                f.write(original_source)
```
martinpitt, line 159: I recommend to do this with `self.addCleanup()` instead, that's a little more robust and avoids these huge retry blocks.

test/verify/check-testlib
```diff
@@ -105,6 +106,84 @@ class TestRunTest(MachineCase):
...
+
+        # Check retry logic for changed source
+        pkg = os.path.join(ROOT_DIR, "pkg", "shell", "hosts.jsx")
+        if not os.path.exists(pkg):
+            return;
```
martinpitt, line 164: stray semicolon; but moreover, I'd replace this with an assertion. If we ever reorganize our code that this file goes away, we want to adjust this test also, not silently skip that part from then on.

martinpitt (changes requested):
This is amazing, thanks for puzzling this together! Some questions, some remarks.

test/common/run-tests
```diff
@@ -90,7 +90,11 @@ def finish_test(opts, test, affected_tests):
     Return (retry_reason, exit_code). retry_reason can be None or a string.
     """
 
-    affected = test.command[0] in affected_tests
+    affected = False
+    for at in affected_tests:
+        if test.command[0].endswith(at):
+            affected = True
+            break
```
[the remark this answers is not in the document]
marusak: Good call with `any`.

And endswith should be enough. It compares `glob` and `git diff --name-only` that are split and striped.

test/common/run-tests
```diff
@@ -69,17 +69,20 @@ def print_test(test, print_tap=True):
     flush_stdout()
 
     if not print_tap:
+        print(retry_reason)
         return
 
+    print()  # Tap needs to start on a separate line
```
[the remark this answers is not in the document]
marusak: `if not print_tap:` does `return`, so no need for `else:`.

test/verify/check-testlib
marusak: good call. However I think this might break dowstream. But I can check if `/pkg` is present and then assert it.

martinpitt: We don't run check-testlib downstream, but even if we did: we expect to unpack the whole upstream dist tarball, not just parts of it.

martinpitt (commented):
Thanks! There is one remaining comment from the previous review round, plus the missing squashing. Otherwise LGTM

martinpitt (approved):
Cheers! LGTM (as usual, bots have the last word)

PR #15321, opened 2021-02-12
ws: Don't let remotectl touch permissions of admin-provided certificates

martinpitt:
Only call set_cert_attributes() for our own managaged 0-self-signed
certificate. Administrator provided ones are often shared with other
services, and mangling the permissions on every cockpit.service start
can actively damage these.

Add unit tests which cover both "our own" and "admin managed" cases.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1907254

### DIFF
--- a/src/ws/remotectl-certificate.c
+++ b/src/ws/remotectl-certificate.c
@@ -185,7 +185,13 @@ ensure_certificate (const gchar *user,
         g_warning ("Failed to remove %s: %m", path);
     }
 
-  ret = set_cert_attributes (path, user, group, selinux);
+  /* adjust permissions of the certificates that we manage automatically;
+   * don't touch admin-provided certs, they are often shared between multiple services */
+  if (g_str_has_suffix (path, "/0-self-signed.cert") ||
+      g_str_has_suffix (path, "/10-ipa.cert"))
+    ret = set_cert_attributes (path, user, group, selinux);
+  else
+    ret = 0;
 
 out:
   g_clear_object (&certificate);
--- a/src/ws/remotectl-certificate.c
+++ b/src/ws/remotectl-certificate.c
@@ -150,12 +150,10 @@ ensure_certificate (const gchar *user,
                     const gchar *selinux,
                     gboolean check_expired)
 {
-  GError *error = NULL;
-  GTlsCertificate *certificate = NULL;
-  gchar *path = NULL;
-  gint ret = 1;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GTlsCertificate) certificate = NULL;
 
-  path = cockpit_certificate_locate_gerror (&error);
+  g_autofree gchar *path = cockpit_certificate_locate_gerror (&error);
   if (path == NULL && g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
     {
       /* no certificate? create a self-signed one */
@@ -168,7 +166,7 @@ ensure_certificate (const gchar *user,
   if (!certificate)
     {
       g_message ("%s", error->message);
-      goto out;
+      return 1;
     }
 
   /* auto-renew our expired auto-created certificates */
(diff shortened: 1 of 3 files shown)


martinpitt:
I'm actually inclined to strenghten this a bit, and *requiring* in the tests that `getpwnam()` and `getgrgid()` succeed. Then you couldn't run this in a test env without a proper user group, but apparently we have done that for years without anybody complaining. This would get rid of all the `test->group ?` special cases (the "else" part of which we never test).

martinpitt:
Breaks two tests, I'll adjust them accordingly. While I am at it, I'll also make the above change.

src/ws/test-remotectlcertificate.c
```diff
@@ -130,11 +131,16 @@ setup (TestCase *tc,
...
 
-  gr = getgrnam (g_get_user_name ());
-  if (gr != NULL)
+  /* determine user's primary group */
+  pwd = getpwnam (g_get_user_name ());
+  if (pwd)
     {
-      g_ptr_array_add (ptr, "--group");
-      g_ptr_array_add (ptr, (gchar *) g_get_user_name ());
+      struct group *gr = getgrgid (pwd->pw_gid);
```
allisonkarlitskaya, line 139: This seems pretty sideways, particularly since we need to scan `/etc/group` to get the `pw_gid` value in the first place...

Why not just go directly to `getgrgid (getgid ())` ?

src/ws/test-remotectlcertificate.c
```diff
@@ -224,7 +234,8 @@ test_refresh_expired (TestCase *test,
   g_autofree gchar *oldpath = g_build_filename (test->cert_dir, "alice-expired.cert", NULL);
   g_autofree gchar *selfsigned_path = g_build_filename (test->cert_dir, "0-self-signed.cert", NULL);
   g_autoptr(GTlsCertificate) certificate = NULL;
-  char *argv[] = { "certificate", "--user", (gchar *) g_get_user_name (), "--ensure", NULL };
+  char *argv[] = { "certificate", "--ensure", "--user", (gchar *) g_get_user_name (),
+                   test->group ? "--group" : NULL, test->group, NULL };
```
allisonkarlitskaya, line 239: ...this made me raise an eyebrow... "gross, but okay, kinda creative... and it's just a test..."

src/ws/test-remotectlcertificate.c
```diff
@@ -243,7 +254,7 @@ test_refresh_expired (TestCase *test,
   g_assert_cmpint (g_tls_certificate_verify (certificate, NULL, certificate), ==, G_TLS_CERTIFICATE_EXPIRED);
 
   /* call with --ensure again, refreshes the cert */
-  ret = cockpit_remotectl_certificate (4, argv);
+  ret = cockpit_remotectl_certificate (test->group ? 6 : 4, argv);
```
allisonkarlitskaya, line 261: ...but this is a total no-go.  Can you replace that with `g_strv_length()`?

src/ws/remotectl-certificate.c
```diff
@@ -171,21 +171,30 @@ ensure_certificate (const gchar *user,
...
+            ensure_certificate (user, group, selinux, FALSE);  /* avoid infinite recursion */
+          else
+            g_warning ("Failed to remove %s: %m", path);
+        }
 
-  ret = set_cert_attributes (path, user, group, selinux);
+      ret = set_cert_attributes (path, user, group, selinux);
+    }
+  else
+    {
```
allisonkarlitskaya, line 194: This function is *begging* to be rewritten with autoptr, inversion of the "not us?" logic to be an early return, and avoiding the recursion...

allisonkarlitskaya (changes requested):

src/ws/test-remotectlcertificate.c
martinpitt: Right, this isn't anywhere near suid. Done, thanks for pointing out!

src/ws/test-remotectlcertificate.c
martinpitt: Dropped. As I wrote above, I made the unit test stricter. Nobody has complained in years that they want to build cockpit in an environment where the current user does not have a group name, so we can just as well make it official. :grin:

src/ws/test-remotectlcertificate.c
martinpitt: The condition got dropped, but g_strv_length() is nice anyway. Thanks! Done.

src/ws/remotectl-certificate.c
martinpitt: Added a new commit for g_auto. I had to change the structure a bit, the failed [realms test](https://logs.cockpit-project.org/logs/pull-15321-20210212-093531-65cde94f-fedora-33/log.html#289-2) pointed out that we still need to adjust the permissions for 10-ipa as well.

allisonkarlitskaya (approved):

PR #15323, opened 2021-02-12
test: FIx TestLogin.testTally for debian-testing

martinpitt:
Configure pam_faillock on debian-testing the same way as on all other
OSes, with deny=4. The default is 3, which caused the test to fail.

This was previously hidden by our naughty for the missing `faillock`
binary (https://github.com/cockpit-project/bots/issues/1585), but this
was fixed in pam 1.4.0-3.


KKoukiou (approved):

allisonkarlitskaya (approved):
Glad to see this making its way through :)

PR #15324, opened 2021-02-12
stop disting map files

allisonkarlitskaya:
...also make their generation conditional.


allisonkarlitskaya:
All RPM-based systems are failing due to RPM packaging issues caused by no `.map` files.

martinpitt (approved):
Looks great, thank you!

PR #15325, opened 2021-02-12
Make breadcrumbs sticky in all pages that can get 'tall' enough

### LINKED ISSUE #1523: Breadcrumb bar needs to be in fixed frame
The breadcrumb bar at the top of various pages needs to be in a fixed frame, so it doesn't scroll when you scroll the page. Having basic navigation scroll out of view is broken.

KKoukiou:
Fixes #1523

### DIFF
--- a/pkg/machines/components/networks/networkList.jsx
+++ b/pkg/machines/components/networks/networkList.jsx
@@ -48,15 +48,17 @@ export class NetworkList extends React.Component {
         const actions = (<CreateNetworkAction devices={devices} dispatch={dispatch} />);
 
         return (
-            <Page breadcrumb={
-                <Breadcrumb variant={PageSectionVariants.light} className='machines-listing-breadcrumb'>
-                    <BreadcrumbItem to='#'>
-                        {_("Virtual machines")}
-                    </BreadcrumbItem>
-                    <BreadcrumbItem isActive>
-                        {_("Networks")}
-                    </BreadcrumbItem>
-                </Breadcrumb>}>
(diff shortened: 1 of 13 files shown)


KKoukiou:
Oh I broke hwinfo tests, will repost.

marusak:
Nice work!
Firewall looks a bit weird though: (to my eyes at least)
weirdfirewall

KKoukiou:
> Nice work!
> Firewall looks a bit weird though: (to my eyes at least)
> weirdfirewall

Oh right, the firewall header should also be part of the sticky component. So I can either repost and split or post the firewall fix right after we merge this, since it's not too bad. Let me know what you prefer.

marusak:
> Oh right, the firewall header should also be part of the sticky component. So I can either repost and split or post the firewall fix right after we merge this, since it's not too bad. Let me know what you prefer.

In an ideal world I would like to see this PR updated :) But if it is realistically too hard to get green, then we can merge this and have follow up.

KKoukiou:
> > Oh right, the firewall header should also be part of the sticky component. So I can either repost and split or post the firewall fix right after we merge this, since it's not too bad. Let me know what you prefer.
> 
> In an ideal world I would like to see this PR updated :) But if it is realistically too hard to get green, then we can merge this and have follow up.

It is really hard. I retried already at least 3 times. Anyway, i see some failures that seem unrelated but are not common flakes so I would actually rather split it.

marusak (approved):
Thanks!
The firewall thingy will be a follow up.

PR #15326, opened 2021-02-12
Remove last redundant tabIndex attributes

### LINKED ISSUE #6057: tabbing order is not optimal: remove tabindex to use built-in browser order
Similar to what we did with the log in screen several weeks ago, we should rework `tabindex` across all of Cockpit, to help improve accessibility.

`tabindex` is generally a misfeature of HTML and gets in the way of natural order of tabbing through elements. (Generally one would want to remove all `tabindex` references, unless it is necessary to hardcode the tabbing order of elements which may be activated.)

See http://webaim.org/techniques/keyboard/tabindex for proper usage of `tabindex`.

---

Note: There are 105 remaining uses of `tabindex` within the Cockpit repository. However, simply removing these might break things (each would need to be inspected in context).

KKoukiou:
Fixes #6057

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -394,7 +394,6 @@ class CurrentMetrics extends React.Component {
                 <Tooltip content={ cockpit.format(_("$0 GiB total"), swapTotal) } position="bottom">
                     <Progress
                         id="current-swap-usage"
-                        tabIndex="0"
                         title={ _("Swap") }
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Cheers! I thought the progresses had to be tab-able to  get a proper screen reader output, but indeed that was bogus -- any other non-tabable element like labels needs to be covered by the screen reader anyway.

PR #15327, opened 2021-02-12
Fix italian translation

82ghost82:
Found several errors in italian translation, I tried to fix them.

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -69,7 +69,7 @@ msgstr "$0 Dispositivo a blocchi"
 
 #: pkg/storaged/mdraid-details.jsx:192
 msgid "$0 chunk size"
-msgstr "$0 Dimensione del pezzo"
+msgstr "$0 Dimensione del blocco"
 
 #: pkg/systemd/overview-cards/insights.jsx:155
 msgid "$0 critical hit"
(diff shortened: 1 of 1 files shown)


marusak:
Thank you!
We use https://translate.fedoraproject.org/projects/cockpit/master/ for translating cockpit. It is usually preferred if translators do translations there.
However, since you have already done so much work, I am happy to sync this into weblate in your stead. Please let me know if you are fine with that and I'll close this PR and push those translations into Weblate.

marusak:
Also if you give me your email that you use here in github, I can mark you as author of these changes.

82ghost82:
> We use https://translate.fedoraproject.org/projects/cockpit/master/ for translating cockpit. It is usually preferred if translators do translations there.

Oh sorry, I didn't know about that, I will take into account for future commits about translation.
There should be other, for sure: I did have only an afternoon to look at it, but some translated strings may still be not perfect, especially that with variables ($0, $1) which can have a "swapped" description/variable because I don't know exactly what are the variables without looking at the files to which the translations point or look at the gui directly.
However it should be a lot better now: the funniest was "informazioni sul pacchetto rinfrescante", which in english translates in something like "information about the cooling package" :D

> However, since you have already done so much work, I am happy to sync this into weblate in your stead. Please let me know if you are fine with that and I'll close this PR and push those translations into Weblate.

Yes, thank you, go for it. I'm quite new to github and pr/commits so I'm not very expert about github commands (after pushing the first pr which failed there was some panic to understand how to add the additional commit to the pr itself :D ), so if you can do it for me it's better, thank you.

>Also if you give me your email that you use here in github, I can mark you as author of these changes.

Thank you, crudo.daniele@gmail.com

marusak:
Ok, I've updated it in weblate.
You can see it in history: https://translate.fedoraproject.org/projects/cockpit/master/it/#history
Or also in this commit: https://github.com/cockpit-project/cockpit-weblate/commit/6ba792c8348171b28e85350e8d2187698680c332 (please mind, this `cockpit-project/cockpit-weblate` is read-only repository, it is just "backend" for weblate, no changes should be posted there manually).

Thank you for your contributions! If you are interested, there are still a lot of failing checks, strings needing editing or translating. You can see it here: https://translate.fedoraproject.org/projects/cockpit/master/it/#overview

PR #15328, opened 2021-02-12
manifest and stamp reform

allisonkarlitskaya:
Here's a few more patches on our way to cleaning up our build system...

I know the stamp plugin is contentious, but the motivation is basically this: I want to get to a place where running `webpack` directly is equivalent to what happens when we type `make`.  For that, I want to shrink the relevant make rule to almost nothing.  That means the `touch` needs to go.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,9 +15,6 @@ noinst_DATA =
 nodist_noinst_DATA =
 nodist_noinst_SCRIPTS =
 
-dist_systemdunit_DATA   =
-nodist_systemdunit_DATA =
-
 TESTS = $(NULL)
 
 CLEANFILES = \
@@ -48,12 +45,9 @@ SUFFIXES = \
 	.css .css.gz \
 	.html .html.gz \
 	.js .js.gz \
-	.json .json.in \
 	.jsx \
 	.map .map.gz \
 	.mo .po \
-	.service .service.in \
-	.socket .socket.in \
 	.svg .svg.gz \
 	.1 .8 .5 \
 	$(NULL)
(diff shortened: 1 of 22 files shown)


Makefile.am
```diff
@@ -186,7 +186,7 @@ dist/%/Makefile.deps:
 # This is the primary rule for running webpack; some pkgs like ssh or pcp only
 # have a manifest, no actual webpack, so don't call webpack-make for these
 dist/%/stamp: $(WEBPACK_CONFIG) $(srcdir)/tools/webpack-make $(srcdir)/node_modules/.stamp
-	$(V_WEBPACK) $(WEBPACK_MAKE) -d dist/$*/Makefile.deps -c $(WEBPACK_CONFIG) && touch $@
+	$(V_WEBPACK) $(WEBPACK_MAKE) -d dist/$*/Makefile.deps -c $(WEBPACK_CONFIG)
```
martinpitt: Honestly I still have a hard time accepting to trade a few characters here with 46 new lines of js code. It feels like overkill to having to write an entire webpack plugin with iterating etc. just to write one more stamp file. We don't write manifest.json from webpack for now, but included-files may very well take this role? It is under our own plugin's control, completely predictable, and would avoid making the webpack build even more involved. Also, the stamp files are completely uninteresting, they are just a crutch for hooking webpack into make (which you want to minimize as much as possible anyway).

That is not a hard veto, and we can clean this up later -- but  these things have a tendency to be sticky.

martinpitt (commented):
Thanks! I have some more questions before I can find the courage to +1 this.

Makefile.am
martinpitt: To clarify -- the stamp files are only for make's sake, not for webpack's, so it feels appropriate to me if the Makefiles create them. I guess the reason for this is that you can run "webpack" directly, then type "make" and not have stuff rebuilt? Then I'd really prefer to move the make dependencies to `included-files`.

allisonkarlitskaya: My basic idea is that if we run `webpack` independently, and then later, run `make`, then `make` shouldn't duplicate the work that `webpack` already did.  That means that whatever we use for dependency tracking needs to be created by both, in the same way.

After our 1:1 it seems like we may end up using something different as a stampfile, so I can back down from this for now and add the `touch` back to the `Makefile`...

martinpitt:
@allisonkarlitskaya : The first two commits are obsolete now. I think you recently mentioned that you want the manifests.json file to play the role of the stamp file now? Is that still the plan, or do you want to revive anything from this PR still? Thanks!

allisonkarlitskaya:
keine Kunst, kann weg

PR #15329, opened 2021-02-12
Remove webpack-make

allisonkarlitskaya:
Here's a first effort look at what it might look like to remove webpack-make from cockpit.


Makefile.am
```diff
@@ -166,33 +160,21 @@ WEBPACK_TEST_DEPS = \
...
-	  $(PRINT_COPY_MSG_$(V))$(MKDIR_P) $(dir $@) && \
-	  cp -p $(srcdir)/$@ $@.tmp && \
-	  mv $@.tmp $@)
-
-# This is the primary rule for running webpack; some pkgs like ssh or pcp only
-# have a manifest, no actual webpack, so don't call webpack-make for these
-dist/%/stamp: $(WEBPACK_CONFIG) $(srcdir)/tools/webpack-make $(srcdir)/node_modules/.stamp
-	$(V_WEBPACK) $(WEBPACK_MAKE) -d dist/$*/Makefile.deps -c $(WEBPACK_CONFIG)
+dist/%/stamp: $(WEBPACK_CONFIG) $(srcdir)/node_modules/.stamp
+	$(V_WEBPACK) NODE_ENV='$(NODE_ENV)' ONLYDIR='$*/' webpack
```
allisonkarlitskaya, line 171: This is the cool part.

allisonkarlitskaya:
Need to figure out what to do with test-*.html so that we included it in the testcases and don't install it....

allisonkarlitskaya:
I think this was misguided.  I had the feeling that it would be nice to invoke `webpack` directly, but there's actually no need for it.  Meanwhile, #15904 is showing that `webpack-make` doesn't necessarily imply `automake`...

We could definitely do with some extra cleanups, but those can come later.

PR #15331, opened 2021-02-14
image-prepare: various cleanups, add --overlay option

martinpitt:
I started this as preparation for replacing the fedora-testing image with a scenario. That will still need some discussion *where* to run these, similar to where to run the "nightly master" tests. I have an idea, but I'll discuss them in a meeting or on IRC.

In the meantime,  there is enough fodder here to warrant a full CI run. I also trigger the dnf-copr scenario, as this PR touches it.


marusak (approved):
Nice, thanks!

PR #15336, opened 2021-02-15
metrics: Don't show too much empty space for future events

marusak:
Show maximum of 5 empty minutes in the current hour. This does not affect missing data in general, only the current hour if cut short. See example: (this is extreme, takes two minutes after full hour)

Before:
1402old

Now:

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -575,7 +575,7 @@ const SvgGraph = ({ data, resource, have_sat }) => {
 };
 
 // data: type → SAMPLES_PER_H objects from startTime
-const MetricsHour = ({ startTime, data }) => {
+const MetricsHour = ({ startTime, data, clipLeading }) => {
     // compute graphs
     const graphs = [];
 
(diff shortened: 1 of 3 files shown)


marusak:
Broke one test. But in general ready for review and options.

pkg/metrics/metrics.jsx
```diff
@@ -575,7 +575,7 @@ const SvgGraph = ({ data, resource, have_sat }) => {
 };
 
 // data: type → SAMPLES_PER_H objects from startTime
-const MetricsHour = ({ startTime, data }) => {
+const MetricsHour = ({ startTime, data, squash }) => {
```
martinpitt, line 578: "squash" sounds a bit non-descriptive; maybe "clip" or "hideEmpty"?

pkg/metrics/metrics.jsx
```diff
@@ -617,7 +617,21 @@ const MetricsHour = ({ startTime, data }) => {
...
 
-    for (let minute = 59; minute >= 0; --minute) {
+    let minutes = 60;
+    if (squash) {
+        // When removing of empty minutes is allowed, find the highest valid minute and round to the next 5 minutes
+        let m = 59;
+        for (; m >= 0; --m) {
+            const dataOffset = m * SAMPLES_PER_MIN;
+            const dataSlice = normData.slice(dataOffset, dataOffset + SAMPLES_PER_MIN);
+            if (dataSlice.find(i => i !== null)) {
```
martinpitt, line 627: This is exactly the same iteratinon/test as the code below already does. But as we have to "rewind" to the previous full 5 minutes, I don't think we can fully avoid it. However, you could do that loop in 5 minute steps instead of 1 minute steps, avoid the arithmetic, and directly work on the `minutes` variable.

martinpitt (changes requested):
Nice! The approach looks fine to me, just some minor nagging.

martinpitt (approved):
Thanks!

PR #15340, opened 2021-02-15
storaged: Add a skip after the NFS editing top alert

### LINKED ISSUE #15339: storage: Alert in NFS edit dialog has no margin
image

mvollmer:
Fixes #15339


KKoukiou:
@mvollmer this is failing to run make

mvollmer:
> @mvollmer this is failing to run make

Thanks, fixed.

KKoukiou (approved):
I would rather handle spacing with CSS and in a more generic way but I can fix this in the big storage PR. 
So ideally that would be a CSS rule in page.scss always adding padding-bottom to alert when that's before a form.

PR #15342, opened 2021-02-15
tools: Link checker warns about redirections

marusak:


marusak:
Heh, needs tests adjusting

tools/urls-check
```diff
@@ -35,6 +35,9 @@ sys.dont_write_bytecode = True
 DAYS = 7
 TASK_NAME = "Validate all URLs"
 
+known_redirects = ["https://access.redhat.com/security/updates/classification/#",
+                   "https://firefox.com/", "https://www.microsoft.com/"]
```
martinpitt, line 39: Aren't these obsolete with the previous commit?

marusak: No they are not.
The first one is in condition, the other twos change to localized page, so I don't want to force there `en_us` or so.

martinpitt (approved):
Thanks! Looks good, assuming green tests.

PR #15346, opened 2021-02-16
machines: correct few details in storage page when pool's deactivated

skobyda:
It's impossible to create a volume when a storage pool is deactived.
But right now, user has to fill out the form only to find this out.
Also when storage pool is deactivated, the "No storage volumes defined
for this storage pool" string is shown in volumes list. That's incorrect.
There are no volumes because libvirt API cannot fetch any information
about volumes for deactivated storage pool. 
Therefore to fix both problems, just hide the storage volumes list if pool
is deactivated and instead show a warning telling to user to activate the
pool

### Before:
Screenshot from 2021-02-16 13-35-15
Screenshot from 2021-02-16 13-35-34

### After:
Screenshot from 2021-02-16 13-28-14

### DIFF
--- a/pkg/machines/components/storagePools/storagePoolVolumesTab.jsx
+++ b/pkg/machines/components/storagePools/storagePoolVolumesTab.jsx
@@ -112,7 +112,7 @@ export class StoragePoolVolumesTab extends React.Component {
                 <ListingTable variant='compact'
                     actions={actions}
                     aria-label={`Storage pool ${storagePool.name} Volumes`}
-                    emptyCaption={_("No storage volumes defined for this storage pool")}
+                    emptyCaption={storagePool.active ? _("No storage volumes defined for this storage pool") : _("Activate the storage pool to administer volumes")}
                     columns={columnTitles}
                     onSelect={this.onSelect}
                     rows={rows} />
--- a/pkg/machines/components/storagePools/storageVolumeCreate.jsx
+++ b/pkg/machines/components/storagePools/storageVolumeCreate.jsx
@@ -143,7 +143,7 @@ export class StorageVolumeCreate extends React.Component {
         const poolTypesNotSupportingVolumeCreation = ['iscsi', 'iscsi-direct', 'gluster', 'mpath'];
 
         const createButton = () => {
-            if (!poolTypesNotSupportingVolumeCreation.includes(this.props.storagePool.type)) {
+            if (!poolTypesNotSupportingVolumeCreation.includes(this.props.storagePool.type) && this.props.storagePool.active) {
                 return (
                     <Button id={`${idPrefix}-button`}
                         variant='secondary'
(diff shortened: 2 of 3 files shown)


pkg/machines/components/storagePools/storagePoolVolumesTab.jsx
```diff
@@ -109,13 +111,17 @@ export class StoragePoolVolumesTab extends React.Component {
...
-                    rows={rows} />
+                {storagePool.active
+                    ? <ListingTable variant='compact'
+                        actions={actions}
+                        aria-label={`Storage pool ${storagePool.name} Volumes`}
+                        emptyCaption={_("No storage volumes defined for this storage pool")}
+                        columns={columnTitles}
+                        onSelect={this.onSelect}
+                        rows={rows} />
+                    : <EmptyState>
```
KKoukiou, line 122: How about integrating this to the emptyCaption? 
```
emptyCaption={storagePool.active ?_("No storage volumes defined for this storage pool") : _("Activate the storage pool to administer volumes")}
```

KKoukiou (approved):
I would rather see the new empty message as part of the table but I leave it up to you. Or maybe let's just ask @garrett

pkg/machines/components/storagePools/storagePoolVolumesTab.jsx
skobyda: That would leave the table header with "Name", "UsedBy", "Size" present (I don't like it being present, I would have to add another condition to hide it). It would also leave the volume creation button present, which would require another condition to disable it. So I think it's simpler to hide the whole part about the volume (including the header and action buttons)

KKoukiou: From my understanding it's same for when the pool is active but there are not storage volumes. We still show the table headers. Therefore I would prefer to stick to that design, but I said I don't have strong feelings about it.

KKoukiou (dismissed):

garrett:
This is certainly better than before.

I think it would be better if we reworked the table to be a little more modern. As deactivation and activation are uncommon, as is deletion, it probably should go into a kebab menu. Additionally, we should try to avoid tabs in expanded tables. If we could show the metadata inline instead of in a key/value pairing (with badge like "labels"), we could have that at the top of an expanded list. And then we could have a secondary list below that.

Like this:

storage-pools-2021-02-16-1545(2)

And the deactivated state could use an empty state pattern when expanded:

storage-pools-2021-02-16-1545(3)

If this is too far out of scope, we could dial back the changes and just use the content of the empty state for this PR and move the rest to a separate PR.

garrett (approved):
I'm happy enough as-is; we could move the rest of what I suggested to a separate issue / PR and treat this as a quick fix.

KKoukiou:
@skobyda the tests fail.

skobyda:
Tests are passing now, there seem to be 2 unrelated failures, so retrying for the sake of green

KKoukiou (approved):

KKoukiou:
@skobyda we don't merge PRs even with unrelated failures anymore. We agreed on it recently in some meetings. Of course in extreme cases, after like 5 retries this is fine :)

PR #15349, opened 2021-02-16
overview: fix accessibility of the progress bar

### LINKED ISSUE #15187: overview: Usage bars lack ARIA labelling
On the overview page, the progress bar widgets we use lack ARIA labelling. They have `aria-labelledby` and point to an ID of an element that does not exist.

Meanwhile, the text next to the progress bars explains what's going on, so instead of `aria-labelledby`, I think we can just use `aria-hidden="true"` instead.

KKoukiou:
Fixes #15187
* [ ] Blocked on https://github.com/patternfly/patternfly-react/pull/5470

### DIFF
--- a/pkg/systemd/overview-cards/usageCard.jsx
+++ b/pkg/systemd/overview-cards/usageCard.jsx
@@ -118,23 +118,25 @@ export class UsageCard extends React.Component {
                     <table className="pf-c-table pf-m-grid-md pf-m-compact">
                         <tbody>
                             <tr>
-                                <th scope="row">{_("CPU")}</th>
+                                <th id="system-usage-cpu-progress" scope="row">{_("CPU")}</th>
                                 <td>
                                     <Progress value={this.state.cpuUsed}
                                         className="pf-m-sm"
                                         min={0} max={100}
                                         variant={ this.state.cpuUsed > 90 ? ProgressVariant.danger : ProgressVariant.info }
                                         label={ this.state.cpuUsed + '% ' + cores_str }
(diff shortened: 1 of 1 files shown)


marusak:
So I think this should be in pf-react-core 4.100.0 so once we update to that this can be unblocked. But at the same time, it is not strictly dependand on that, just that only once this version is available the a11y issue will disappear. So if you run tests on this I am happy to get this in.

pkg/systemd/overview-cards/usageCard.jsx
```diff
@@ -118,23 +118,25 @@ export class UsageCard extends React.Component {
...
                             </tr>
                             <tr>
-                                <th scope="row">{_("Memory")}</th>
+                                <th id="system-usage-memory-progress" scope="row">{_("Memory")}</th>
                                 <td>
                                     <Progress value={this.state.memUsed}
                                         className="pf-m-sm"
                                         min={0} max={Number(this.state.memTotal)}
                                         variant={fraction > 0.9 ? ProgressVariant.danger : ProgressVariant.info}
+                                        aria-labelledby="system-usaage-memory-progress"
```
marusak, line 139: `usaage` typo thus does not match the id.

marusak (changes requested):

pkg/systemd/overview-cards/usageCard.jsx
KKoukiou: fixed.

marusak (approved):
Thanks!

PR #15352, opened 2021-02-16
test: Hack around Chromium 88 crash with key input

martinpitt:
Chromium headless 88 crashes when trying to input Ctrl+A:
https://bugs.chromium.org/p/chromium/issues/detail?id=1170634

Work around that by pressing backspace as many times as the current
input length instead.

----

I just spent another hour on trying to fix this in a less gross way (see chromium bug), but unfortunately failed :crying_cat_face: Let's see how this works.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -323,7 +323,12 @@ class Browser:
     def set_input_text(self, selector, val, append=False, value_check=True):
         self.focus(selector)
         if not append:
-            self.key_press("a", 2) # Ctrl + a
+            if self.cdp.browser != "chromium":
+                self.key_press("a", 2) # Ctrl + a to select all
+            else:
+                # HACK: Chromium headless 88 crashes on the above: https://bugs.chromium.org/p/chromium/issues/detail?id=1170634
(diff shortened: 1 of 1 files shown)


marusak:
Hmm, which test is failing exactly?
I've now updated chromium and `ctrl+a` works just fine.
`Version 88.0.4324.150 (Developer Build) Fedora Project (64-bit)`

martinpitt:
@marusak: Any which uses `set_input_text()`, I am using `TestAccounts.testBasic` as guinea pig. Note that this only affects headless chromium. If you use the interactive one, it works. I noticed that [during a tasks container refresh](https://github.com/cockpit-project/cockpituous/pull/366) (which updated chromium from 87 to 88), see e.g. [this log](https://logs.cockpit-project.org/logs/pull-15210-20210125-065746-8d49c9b5-rhel-8-4/log.html). I have exactly the same package as you locally and can reproduce the crash just fine.

martinpitt:
Nevermind -- indeed installing `chromium` and uninstalling `chromium-headless` does the trick. I'll do that to downstreams instead.

PR #15356, opened 2021-02-17
test: Enable subscription-manager in dnf in TestUpdatesSubscriptions

martinpitt:
Most recent RHEL 8.4 cloud image does not enable it by default any more.

### DIFF
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -997,6 +997,8 @@ class TestUpdatesSubscriptions(PackageCase):
 
         # make sure that rhsm skips certificate checks for the server
         self.sed_file("s/insecure = 0/insecure = 1/g", "/etc/rhsm/rhsm.conf")
+        # enable subscriptions in dnf
+        self.sed_file("s/^enabled=0/enabled=1/", "/etc/yum/pluginconf.d/subscription-manager.conf")
 
         # Wait for the web service to be accessible
         m.execute(script=WAIT_SCRIPT % {"addr": "10.111.112.100"})


martinpitt:
Blocks https://github.com/cockpit-project/bots/pull/1684

test/verify/check-packagekit
```diff
@@ -997,6 +997,8 @@ class TestUpdatesSubscriptions(PackageCase):
 
         # make sure that rhsm skips certificate checks for the server
         self.sed_file("s/insecure = 0/insecure = 1/g", "/etc/rhsm/rhsm.conf")
+        # enable subscriptions in dnf
+        self.sed_file("/enabled/ s/0/1/", "/etc/yum/pluginconf.d/subscription-manager.conf")
```
KKoukiou, line 1001: That's a weird regexp but I am sure you know what you doing. But out of curiosity what's the difference between this and 's/enabled=0/enabled=1'?

KKoukiou (approved):

test/verify/check-packagekit
martinpitt: Not that much difference, just that one does not need to jump through hoops to specify the whole line. e.g. with arbitrary spaces in between, or which separator, etc. This says "in every line that contains "enabled", replace "0" with "1".

Maybe I should be a litte more precise though and say `/^enabled/`. I can re-post.

martinpitt:
The [metrics failure](https://logs.cockpit-project.org/logs/pull-15356-20210217-080731-1bcd38da-rhel-8-4-bots-1684/log.html) on the new image was expected. I just rebased that PR to include the new naughty, and I'll retry that test to not ruin our test statistics :grin: .

test/verify/check-packagekit
KKoukiou: I think it's fine. After all it's a test.

martinpitt: I made it stricter and in the form you suggested, if that's easier to read.

KKoukiou: It was more a question than suggestion :) but its' fine either way for me.

KKoukiou (approved):

PR #15360, opened 2021-02-17
storage: Improve side panel layout, again

mvollmer:
This uses Patternfly `Flex` components to do the layout, and CSS for font sizes and margins.

image

### DIFF
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -626,19 +626,17 @@ td.storage-action {
 
 /* Overview side panels */
 
-#storage .storage-sidebar table {
-    table-layout: fixed;
-}
-
 .sidepanel-row {
-    display: flex;
-    flex-direction: row;
-    justify-content: space-between;
-    gap: var(--pf-global--spacer--sm);
+    padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--md);
+    border-bottom: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
 }
 
-.sidepanel-row-detail {
-    margin-right: 0.5em;
+.sidepanel-row:last-child {
+    border-bottom: none;
+}
+
+.sidepanel-row:hover {
+    background-color: var(--color-link-hover-bg);
 }
 
 .sidepanel-row-name {
@@ -647,9 +645,6 @@ td.storage-action {
 
 .sidepanel-row-info {
     font-size: var(--pf-global--FontSize--xs);
-    display: flex;
-    flex-flow: row wrap;
-    justify-content: space-between;
 }
 
 .sidepanel-row .spinner {
--- a/pkg/storaged/side-panel.jsx
+++ b/pkg/storaged/side-panel.jsx
@@ -23,7 +23,7 @@ import React from "react";
 import { OptionalPanel } from "./optional-panel.jsx";
 import { get_block_link_parts, block_name } from "./utils.js";
 
-import { Spinner } from '@patternfly/react-core';
+import { Button, Spinner, Flex, FlexItem } from '@patternfly/react-core';
 
 const _ = cockpit.gettext;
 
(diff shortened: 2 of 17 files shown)


pkg/storaged/storage.scss
```diff
@@ -626,19 +626,20 @@ td.storage-action {
...
-    justify-content: space-between;
-    gap: var(--pf-global--spacer--sm);
+.sidepanel-row:last-child {
+    border-bottom: none;
 }
 
-.sidepanel-row-detail {
-    margin-right: 0.5em;
+.sidepanel-row:hover {
+    background-color: var(--color-link-hover-bg);
```
garrett, line 642: This variable will change soon in PR https://github.com/cockpit-project/cockpit/pull/15183 (which removes a _lot_ of Cockpit-specific CSS variables and standardizes the remaining ones with a prefix)

Specifically, to `--ct-color-link-hover-bg`
https://github.com/cockpit-project/cockpit/pull/15183/files#diff-7d0ccb2f6de8d81d3e61bc5c350c8b0461851c05480bee0f5d715d4ee8001909R309

I guess I need to finish up that PR. :wink:

pkg/storaged/storage.scss
```diff
@@ -626,19 +626,20 @@ td.storage-action {
 
 /* Overview side panels */
 
-#storage .storage-sidebar table {
-    table-layout: fixed;
+.sidepanel-row {
+    padding-left: var(--pf-global--spacer--md);
+    padding-right: var(--pf-global--spacer--md);
+    padding-top: var(--pf-global--spacer--sm);
+    padding-bottom: var(--pf-global--spacer--sm);
```
garrett, line 633: This is better with shorthand:

```suggestion
    padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--md);
```

garrett (changes requested):
We'll have to change the color at some point and should change the padding to shorthand. I still need to review this locally; right now, I just glanced at the diff (mainly focusing on the CSS).

pkg/storaged/storage.scss
garrett: Note: This shouldn't be updated until that lands. Else, it should stay the same and be updated in that other PR. (Whichever PR lands second of these two will need to change it, in other words.)

mvollmer: Can I use that prefix already here?

pkg/storaged/storage.scss
mvollmer: Done!

garrett: Explanation:

CSS has shorthand to combine things, it takes the form of one of the following:

- analogue clock directions: top right bottom left (this is why it has funky ordering... just think of the hands moving around a clock and you'll be fine)
- top/bottom left/right (what I'm using above)
- top sides bottom
- every-direction-at-once

It's only useful when defining all the padding-* border-* margin-* etc. at one time. (You can't skip a side, so re-declaring something will overwrite it.)

pkg/storaged/storage.scss
KKoukiou: > We'll have to change the color at some point and should change the padding to shorthand. I still need to review this locally; right now, I just glanced at the diff (mainly focusing on the CSS).

@garrett What does it mean we 'have to change the color'? You mean what table-hover used to do? If that's the case, isn't having links enough?

KKoukiou (approved):
I test it to me it looks good, and the code it much more PF4isch than before. 
I don't really understand what is the blocker here, I will wait for monday @garrett's feedback.

garrett:
I've been trying to figure out how to make warning icons and nothing I do seems to trip it up.

How can I reliably make warning icons appear for this page (in a VM)?

mvollmer:
> How can I reliably make warning icons appear for this page (in a VM)?

An easy way is to mount a filesystem in Cockpit, and then unmount it in the terminal.

garrett (approved):
LGTM

Notes:

- I've tried to figure out a way to recreate the issue pointed out in the other PR @ https://github.com/cockpit-project/cockpit/pull/15261#issuecomment-782085350 but cannot.

- It's a little weird to have an unnamed div in the panel for the decoration, but I guess it's there because of the flex. It would be great if it had a name like .sidepanel-row-deocration, but as it doesn't have any custom CSS (at least at the moment?) it's probably not really a big deal. I understand that it's there because its parent element is a flex container. So it's not really an issue.

pkg/storaged/storage.scss
garrett: @KKoukiou it means to change the name of the variable; it's changing in PR #15183

PR #15364, opened 2021-02-17
Restyle selinux to fit other pages

marusak:
No difference functionally, just looks.

Before:
beforeselinux

Now:
newandhsiny

### DIFF
--- a/pkg/selinux/setroubleshoot-view.jsx
+++ b/pkg/selinux/setroubleshoot-view.jsx
@@ -23,7 +23,7 @@ import React from "react";
 import {
     Alert, AlertActionCloseButton, Button,
     Page, PageSection, PageSectionVariants,
-    Switch
+    Switch, Stack, StackItem,
 } from "@patternfly/react-core";
 import { ExclamationCircleIcon, TrashIcon } from "@patternfly/react-icons";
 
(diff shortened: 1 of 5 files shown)


pkg/selinux/setroubleshoot-view.jsx
```diff
@@ -462,29 +470,37 @@ export class SETroubleshootPage extends React.Component {
...
 
-        var errorMessage;
+        let errorMessage;
         if (this.props.error) {
             errorMessage = (
-                <Alert isInline
-                    variant='danger' title={this.props.error}
-                    actionClose={<AlertActionCloseButton onClose={this.handleDismissError} />} />
+                <AlertGroup isToast>
+                    <Alert
```
KKoukiou, line 477: When using Alerts inside Toast you should use isLiveRegion property which adds the aria-live="polite"  This helps assistive technoligies somehow.

KKoukiou (changes requested):
only one tiny issue

pkg/selinux/setroubleshoot-view.jsx
KKoukiou: Btw opened just now https://github.com/patternfly/patternfly-react/issues/5477

marusak: thanks, fixed!

KKoukiou (approved):

Issue #6399, opened 2017-04-27
Allow browser to remember login


jniederm:
Please wrap the 'username' and 'password' input fields in login page to `form` tag to allow browser to remember credentials.

demo: https://jsfiddle.net/mdq538wm/1/

stefwalter:
Cockpit is a Linux session. I guess some other Linux sessions have auto-login too, but it's usually something that's explicitly enabled per user session. This needs some thought.

@andreasn @petervo What do you think about auto-login to Cockpit using browser password caching?

petervo:
From a linux session point of view, i don't really like it. But I'm not really able to articulate a clear reason why we shouldn't do it other than it makes me a bit uncomfortable. Unless someone else is I think it's sort of a default expectation people have in browsers these days and we should probably allow it to happen.

martinpitt:
TBH I don't see why we wouldn't do it - the browser keyring already has all sorts of credentials and is meant for that. It's usually protected by the user's password (in most cases) already anyway, so one has to enter it at least once.

rhwood:
FWIW, on my MacBook, running macOS High Sierra, all browsers except Firefox already offer to remember the login of cockpit running in a vagrant image (that's Google Chrome, Opera, and Safari that make the offer). Not sure if Firefox is not offering to save credentials only because the connection is not encrypted.

rhwood:
[NIST 800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html#-5112-memorized-secret-verifiers) specifically encourages supporting the use of password managers (10th paragraph).

Limero:
This would be very useful. Those not interested in saving the login can just choose "no" when prompted by their browser.

sergiomb2:
+1

sergiomb2:
+1

AndreasKieling:
+9001

martinpitt:
I tried this again, and I can't make Firefox remember the password for the life of me. The current web page already has the required input types (text and password) for user and password lines.

I tried converting into a proper `<form>` and even tried to change the login button into a "submit" input (which is wrong from a UI POV, but I wanted to check if it helps). Nothing.

```
--- src/ws/login.html
+++ src/ws/login.html
@@ -23,7 +23,7 @@
         </div><!--/.col-*-->
 
         <div id="login" class="col-sm-7 col-md-6 col-lg-5 login-area" style="visibility: hidden;">
-          <div role="form">
+          <form>
 
[...]

PR #15366, opened 2021-02-17
login: Use form on login page

### LINKED ISSUE #15335: Login page is not a proper HTML form
Cockpit version: 215
OS: Ubuntu 20.04
Page: -

The login form is not a html `<form>` element.
This leads to bad accessibility but also a degraded experience in general. For example my browser does not prompt to save the password.

Steps to reproduce

1. Open the cockpit website (without being logged in) in - you should see the login page
2. Inspect the DOM - there is no `<form>` element but only a `<div role="form">`
3. Login with your username/password - in Chrome at least you will not be prompted to save the password

I assume the autofill issue is also due to not using a proper form element. By looking at the code it seems that the login process is for some reason implemented mostly client side in JS for some reason unknown to me. However HTML5 also allows form elements with no action attribute which could be used instead of the current div.

marusak:
Fixes #15335

### DIFF
--- a/pkg/static/login.html
+++ b/pkg/static/login.html
@@ -43,7 +43,7 @@
     </div>
 
     <div id="login" class="login-area" hidden>
-      <div role="form">
+      <form onsubmit="return false">
 
         <div id="conversation-group" class="form-group" hidden>
           <div id="conversation-message"></div>
@@ -101,7 +101,7 @@
           </button>
           <a id="get-out-link" href="/" translate>Cancel</a>
         </div>
-      </div>
+      </form>
     </div>
 
     <div id="login-wait-validating" hidden>


martinpitt:
I think that's about the third PR that tries that, and they always have failed. Somehow it refuses to be that simple :cry:

martinpitt:
Interesting! Seems the integration tests are actually happy for the most part, and it's only selenium which needs updating. So apparently we got over whatever was breaking it in the past with some intermediary redesign. Yay!

marusak:
When I compare with your last try:  https://github.com/cockpit-project/cockpit/issues/6399#issuecomment-544405791
The difference mainly seems to be submit on button.

marusak:
Meh, why is firefox and edge unhappy? I'll try locally firefox.

marusak:
Indeed, firefox breaks

marusak:
With the new version also works on firefox. And now it even remembers user name, which it did not before (before it only remembered password).

marusak:
Works like a charm!

martinpitt (approved):
Thanks for figuring out the magic incantation!

PR #15368, opened 2021-02-17
logs: Add redirection to services page

marusak:
Both commits need tests but I would like to hear options about this first.

motdredirect

### DIFF
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -263,6 +263,12 @@ Unit=test.service
         b.go("/system/services#test.service")
         b.enter_page("/system/services")
 
+        b.click(".cockpit-logline:contains('WORKING')")
+        b.enter_page("/system/logs")
+        b.click("button:contains('Go to test.service')")
+        b.enter_page("/system/services")
+        b.wait_js_cond('window.location.hash === "#/test.service"')
+
         # Stop and disable and back again
         self.toggle_onoff()
         self.check_service_details(["Disabled"], ["Start", "Disallow running (mask)"], False)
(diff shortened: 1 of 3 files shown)


KKoukiou:
I am happy with this addition, I myself often have this services page -> logs -> services page path.

marusak:
I think I should filter out units we don't show. That involves:
1. `.scope`, `.mount`... This is easy, I'll just keep those 5 types we have
2. Units that are not present for any reason (maybe because it does not exists anymore). This is tricky. We would need to always load all units to check. I think we can just show warning that the unit was not found.

marusak:
Adding release-note label but not the most important or interesting thing. Mostly if we have not much for release notes this can be used.

marusak:
TestTimer.testCreate is impossible to get green on firefox. Since here it is affected test, it is 3 times as hard.

test/verify/check-system-services
```diff
@@ -263,6 +263,11 @@ Unit=test.service
         b.go("/system/services#test.service")
         b.enter_page("/system/services")
 
+        b.click(".cockpit-logline:contains('WORKING')")
+        b.enter_page("/system/logs")
+        b.click("button:contains('Go to test.service')")
+        b.enter_page("/system/services")
```
martinpitt, line 269: Could this check that it actually shows the test.service details, not the overview?

martinpitt (commented):
Wow, a two-digit issue! nice job :grin:

test/verify/check-system-services
martinpitt: ... asserting `window.location.hash` should work?

marusak: Good call, added

martinpitt (approved):
Thanks! I tested this, works great.

marusak:
The new test is not working. Seems weird as it is exactly how we check it a few lines above. Also it is on the right page. I'll check it out locally.

marusak:
Fixed. Not sure why it does not work the other way around - as it should wait as well.

martinpitt (approved):
Ah, makes sense! `enter_page()` has a side effect, it switches the "current frame" for the subsequent tests. So this couldn't work with the previous one.

PR #15370, opened 2021-02-18
Make all the things

martinpitt:
See individual commits for details. I tested all of make-{dist,srpm,rpm,deb} from a completely clean git checkout, with downloading and with self-building.

### DIFF
--- a/tools/make-debs
+++ b/tools/make-debs
@@ -1,0 +1,78 @@
+#!/bin/bash
+# This file is part of Cockpit.
+#
+# Copyright (C) 2021 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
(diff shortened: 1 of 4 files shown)


martinpitt:
@marusak: This should also fix your make-rpms woes.

tools/make-debs
```diff
@@ -0,0 +1,78 @@
...
+rm -rf "$builddir"
+mkdir -p "$builddir"
+olddir=$(pwd)
+cd "$builddir"
+tar --strip-components=1 -xf "$source"
+cp -r tools/debian .
+# plug in our version
+sed -i "1 s/(0-/(${version}-/" debian/changelog
+dpkg-buildpackage -us -uc -b -j$(nproc)
+cd "$olddir"
```
KKoukiou, line 74: `cd -` also works here, just to avoid storing olddir.

martinpitt: Done, thanks!

martinpitt:
I fixed up the messaging harder, and added another fix for a bug that @marusak just found - we shouldn't download a tarball if there are local uncommitted changes.

marusak:
Nice, thank you for fixing up both issues I've struggled with today morning!

KKoukiou (approved):
I have already tests the make-debs in my livetest branch btw, works as expected.

PR #15371, opened 2021-02-18
metrics: Don't show duplicate events

### LINKED ISSUE #14809: Duplicated events
At 3:38 PM, "net i/o spike" is shown twice, and at 3:53 PM same for "disk i/o":

image

Reproducer: [/var/log/pcp/pmlogger](https://github.com/martinpitt/performance-graphs/files/5369629/pcp.zip)  data to reproduce this. Select "Saturday, October 10" as the day.

Copied from https://github.com/martinpitt/performance-graphs/issues/81

marusak:
If one type of event happens in one minute block more then once, show it
just once.

Fixes #14809


martinpitt (commented):
Cheers! Any chance to turn your [reproducer](https://github.com/cockpit-project/cockpit/issues/14809#issuecomment-781138294) into a test?

marusak:
> Any chance to turn your reproducer into a test?

I might try. When I recreated that zip into tar.gz the timestamps were broken and pcp refused to read it. I hope there are some options how to preserve timestamp. Or then teach `archivePrepare` to work with zips as well.
Also the host name is different, but I can just teach `archivePrepare` to set the right hostname.

martinpitt:
Hmm, normally tar tries to restore original time stamps, you explicitly have to disable that with `--touch`. zip most likely can't do that. But anyway, if it's too complicated then nevermind.

marusak:
Ok, so unzipping and tarring was not working for me. Maybe some of those millions of options that these tools provide would do the trick, but meh, zip will work, right?

martinpitt:
That test indeed did [cause](https://logs.cockpit-project.org/logs/pull-15371-20210218-192910-c7405934-debian-stable/log.html#303) [failures](https://logs.cockpit-project.org/logs/pull-15371-20210218-192910-c7405934-ubuntu-2004/log.html#303) due to timestamps :/

martinpitt:
Marking release-blocker for 238.1 next Monday, this is good polishing matter.

marusak:
In ubuntu and debian `systemctl start pmlogger` hangs until it is eventually killed by timeout. It seems to be somewhat related to changing of hostname

martinpitt:
@KKoukiou ^ FYI -- that sounds like the same issue that killed your /livetest?

KKoukiou:
> @KKoukiou ^ FYI -- that sounds like the same issue that killed your /livetest?

Maybe :) I did not investigate that really.

marusak:
Meh, I don't know. That archive just does not want to play along with ubuntu/debian.
I've check other archives but neither has any double events.

martinpitt:
@marusak : How about just running the test on `TEST_OS_DEFAULT` then? should be sufficient. Not worth drilling down into this quirk..

martinpitt (approved):
Thanks! Weird that this requires zip, but :man_shrugging:

PR #15372, opened 2021-02-18
Revert "test: Enable subscription-manager in dnf in TestUpdatesSubscriptions

martinpitt:
This reverts commit 33d1a9bc6f5cf046a159466fd36ef52607480421.

This was really the wrong thing to do. Ideally rhsm would revert the
recent change that broke the default subscription-manager.conf, but if
that really has to stay, then our page needs another approach of
figuring out whether subscriptions are really required.

### DIFF
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -997,8 +997,6 @@ class TestUpdatesSubscriptions(PackageCase):
 
         # make sure that rhsm skips certificate checks for the server
         self.sed_file("s/insecure = 0/insecure = 1/g", "/etc/rhsm/rhsm.conf")
-        # enable subscriptions in dnf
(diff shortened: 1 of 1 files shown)


KKoukiou:
I thought we enabled it so that we can actually test it. I don't get why we redisable the subscriptions repository (or how it's called) now?

martinpitt:
I just filed https://bugzilla.redhat.com/show_bug.cgi?id=1930037 with the whole background. In short, this is not supposed to be disabled on the RHEL images, and like that we paper over this bug -- real users would still see this broken. I will most probably make a hack on the image itself, but that will then not apply to rhel-8-3 and other images.

marusak (approved):
Makes sense to test the page with tools that come by default. If user needs to go into this config and set this option in order that the page works, then we should deal with it in the UI.

PR #15373, opened 2021-02-18
Test tracer on rhel-8-4

martinpitt:
This is blocked still, but I mostly want to ensure that tracer works as expected on RHEL 8.4, so that we don't ship something broken.

 - [x] Add tracer to rhel-8-4 image: https://github.com/cockpit-project/bots/pull/1684


martinpitt:
@skobyda: Good news! [tracer works fine](https://logs.cockpit-project.org/logs/pull-15373-20210218-091053-2b4691be-rhel-8-4-bots-1684/log.html) on rhel-8-4. :tada:

marusak (approved):
Very nice, thanks!

PR #15375, opened 2021-02-18
metrics: Don't show swap column when no is being used

### LINKED ISSUE #14807: Don't show `swap` column when no swap used
When no swap was used in any minute of all rendered data just don't show this column but make it one-sided graph like networking and disk.

Copied from https://github.com/martinpitt/performance-graphs/issues/78

marusak:
Fixes #14807

nowswap


martinpitt (approved):
Sweet! Funny, I had the same issue open for looking at today :grinning:

PR #1705, opened 2015-01-21
bridge: Minor PCP metrics fixes

petervo:
* Don't assert >= 0 for unsigned integers since it will always be true
* use va_end in mock

Suggested by coverity

PR #15378, opened 2021-02-19
shell: Drop pkexec/sudo path hardcoding into manifest

### LINKED ISSUE #14347: administrative access impossible since cockpit 219 (migration from pkexec to sudo)
Refactoring administrative access on cockpit 219 included changes in witch sudo is used in replacement of pkexec. To test if user has access to sudo, a function spawn command "sudo -v -n".

In some O.S., calling this command return an exit code 1 with message "sudo: a password is required", even the user is in sudoers. This is caused by the sudo timestamp feature requiring the user to confirm his own password on 5 mins of sudo inactivity.

https://www.sudo.ws/man/1.8.24/sudoers_timestamp.man.html

This regression causes the user cannot longer get administrative access.

Tested on linux custom distribution based on Yocto Dunfell, (sudo v1.8.31).

Issue is present sin 219... tested until 223.

martinpitt:
There is little reason to do that. On any sensible system these programs
belong into `$PATH`, and as we don't build-depend on pkexec nor sudo,
the shipped manifests will always have /usr/bin/{sudo,pkexec} anyway.

This gets rid of one part of dynamic manifests, and thus makes them
more portable (~/.local/share/cockpit) and easier to generate by webpack.

Fixes #14347

-----

This is the easiest part of our attempt to not have to deal with *.manifest.in and having to invoke configure and make to build them. It'd be a lot nicer for manifests to be part of static dist and the release tarballs.

 - [x] Requires a truckload of testActivePages fixes: PR #15385

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -128,8 +128,6 @@ SUBST_RULE = \
 	-e 's,[@]prefix[@],$(prefix),g' \
 	-e 's,[@]PACKAGE[@],$(PACKAGE),g' \
 	-e 's,[@]VERSION[@],$(VERSION),g' \
-	-e 's,[@]PKEXEC[@],$(PKEXEC),g' \
-	-e 's,[@]SUDO[@],$(SUDO),g' \
 	-e 's,[@]user[@],$(COCKPIT_USER),g' \
 	-e 's,[@]group[@],$(COCKPIT_GROUP),g' \
 	-e 's,[@]wsinstanceuser[@],$(COCKPIT_WSINSTANCE_USER),g' \
--- a/configure.ac
+++ b/configure.ac
@@ -298,14 +298,6 @@ fi
 AC_PATH_PROG([CHPASSWD], [chpasswd], [/usr/sbin/chpasswd], [$PATH:/usr/local/sbin:/usr/sbin:/sbin])
 AC_DEFINE_UNQUOTED([PATH_CHPASSWD],["$CHPASSWD"],[Location of chpasswd binary])
 
-# pkexec
-AC_PATH_PROG([PKEXEC], [pkexec], [/usr/bin/pkexec], [$PATH:/usr/local/sbin:/usr/sbin:/sbin])
-AC_SUBST(PKEXEC)
-
-# sudo
-AC_PATH_PROG([SUDO], [sudo], [/usr/bin/sudo], [$PATH:/usr/local/sbin:/usr/sbin:/sbin])
-AC_SUBST(SUDO)
-
 # ssh-add
 AC_PATH_PROG([SSH_ADD], [ssh-add], [/usr/bin/ssh-add], [$PATH:/usr/local/sbin:/usr/sbin:/sbin])
 AC_DEFINE_UNQUOTED([PATH_SSH_ADD], ["$SSH_ADD"], [Location of ssh-add binary])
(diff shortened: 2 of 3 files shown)


marusak:
I guess fixes https://github.com/cockpit-project/cockpit/issues/14347 ?

martinpitt:
@marusak : Well spotted, yes! Added to description.

martinpitt:
Ugh, [breaks](https://logs.cockpit-project.org/logs/pull-15378-20210219-070706-d729d8e0-debian-stable/log.html#147) `test/verify/check-shell-active-pages TestActivePages.testBasic` on some OSes, moving to draft.

martinpitt:
This  change did not really *break* `TestActivePages` (*phew*, I would have a hard time to understand how that could happen), but the test had multiple defects all along.  This PR just exposed them because it runs all shell tests three times due to our new "affected" regime.

It was much worse locally, with full chromium it was pretty much *impossible* to get it to succeed. I now tried it several times with headless and visible chromium, and it works well for me. Let's see what the bots say.

martinpitt:
Still some flakes.

martinpitt:
I dropped the `timeout=5`, that was just for local debugging. I also fixed yet another race. Let's roll again!

martinpitt:
I can reproduce the failure on [debian/ubuntu](https://logs.cockpit-project.org/logs/pull-15378-20210219-144940-b0b23039-debian-testing/log.html#325). The `sleep` process just gets reparented to pid 1, so somehow the user session is configured slightly differently there.

martinpitt:
I folded in the test fixes from #15385, so that I get two results and can torture this test a bit more.

martinpitt:
@marusak: Since the  `# RETRY` rework, our HTML somehow [mis-counts the retries](https://logs.cockpit-project.org/logs/pull-15378-20210221-120746-b5b9c619-fedora-32-firefox/log.html). This [debian-stable log](https://logs.cockpit-project.org/logs/pull-15378-20210221-104759-b5b9c619-debian-stable/log.html) also has one retry which is not properly counted.

marusak:
> Since the # RETRY rework, our HTML somehow mis-counts the retries. This debian-stable log also has one retry which is not properly counted.

This has been the case for long time (ever since we started to retry successful runs). The fact that the progress bar doe not show colors is css issue that got broken in bootstrap somehow.

Completely unrelated to your comment I've posted https://github.com/cockpit-project/bots/pull/1705 and then noticed this comment.

Current firefox on F32:
currentfirefox

And with #1705:
newntfirefox


Currently debian-stable looks like this:
currentdebian

and with #1705 it looks like this:
newlogs

martinpitt:
@marusak : We should discuss that on the bots PR, but 36 and 42 affected retries seem excessive?

martinpitt:
Meh, [TestTimers.testCreate on firefox](https://logs.cockpit-project.org/logs/pull-15378-20210221-180019-b5b9c619-fedora-32-firefox/log.html#306-1) is awfully hard to get green -- it [fails in 3 out of 4 cases right now](https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&test=test%2Fverify%2Fcheck-system-services+TestTimers.testCreate)

mvollmer (approved):

PR #15380, opened 2021-02-19
make_dist: Changes in tests are not reason to skip download

marusak:
Common workflow is to fix tests in existing PR. One would adjust tests
and then prepare machine to validate these changes. But due to changes
in tests the downloading would be unnecessary skipped.


martinpitt (approved):
Nice tweak! :+1:

PR #15382, opened 2021-02-19
bridge: Clean password from memory

mvollmer:
- [x] grep for base64 encoded password

### DIFF
--- a/src/bridge/cockpitrouter.c
+++ b/src/bridge/cockpitrouter.c
@@ -32,6 +32,7 @@
 #include "common/cockpitpipetransport.h"
 #include "common/cockpittemplate.h"
 #include "common/cockpithex.h"
+#include "common/cockpitmemory.h"
 
 #include <stdlib.h>
 #include <string.h>
@@ -846,7 +847,12 @@ process_open (CockpitRouter *self,
         }
     }
   if (new_payload)
-    g_bytes_unref (new_payload);
+    {
+      gsize size;
+      gpointer data = g_bytes_unref_to_data (new_payload, &size);
+      cockpit_memory_clear (data, size);
+      g_free (data);
+    }
 }
 
 static void
--- a/test/verify/cracklib.py
+++ b/test/verify/cracklib.py
@@ -1,0 +1,36 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2021 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import re
+import base64
+import subprocess
+
+
+def assert_no_password_in_bridges(machine, user, passwd):
+    passwd_b64 = base64.b64encode(passwd.encode("utf-8")).decode("ascii")
+    basic_auth = base64.b64encode((user + ":" + passwd).encode("utf-8")).decode("ascii")
+
+    regex = '%s\|%s\|%s' % (re.escape(passwd),
+                            re.escape(passwd_b64),
+                            re.escape(basic_auth))
+
+    bridge_pids = machine.execute("pgrep cockpit-bridge").split()
+    for pid in bridge_pids:
+        try:
+            machine.execute("pid=%s; gcore $pid && ! grep '%s' core.$pid" % (pid, regex))
+        except subprocess.CalledProcessError:
+            print("Password found, but that's fine...")
--- a/test/verify/check-shell-multi-machine
+++ b/test/verify/check-shell-multi-machine
@@ -24,7 +24,7 @@ import time
 import parent
 from testlib import *
 from machine_core.constants import TEST_OS_DEFAULT
-
+from cracklib import assert_no_password_in_bridges
 
 def break_hostkey(m, address):
     filename = "/home/admin/.ssh/known_hosts"
@@ -891,6 +891,17 @@ class TestMultiMachine(MachineCase):
         b.wait_in_text('#troubleshoot-dialog', "The SSH key")
         b.wait_not_visible('.password-change-advice')
         b.wait_not_visible('.login-setup-auto')
+
+
+    def testCleanMemory(self):
+        b = self.browser
+        m1 = self.machine
+
+        self.login_and_go(None)
+        add_machine(b, "10.111.113.2", password="foobar")
+
+        assert_no_password_in_bridges(m1, "admin", "foobar")
+
 
 if __name__ == '__main__':
     test_main()
--- a/test/verify/check-superuser
+++ b/test/verify/check-superuser
@@ -19,6 +19,7 @@
 
 import parent
 from testlib import *
+from cracklib import assert_no_password_in_bridges
 
 @skipImage("Do not test BaseOS packages", "rhel-8-3-distropkg", "rhel-8-4-distropkg")
 class TestSuperuser(MachineCase):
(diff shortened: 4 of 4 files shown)


mvollmer:
There should be a failure on https://github.com/cockpit-project/cockpit/pull/15382/commits/c6cd18e3f05e98f9ef9dee42188b6b5d7cecfbc6 where we find the password in TestMultimachine. Once that has run, I'll push the fix.

mvollmer:
The password was also found in TestSuperuser, interesting.

martinpitt:
Still some [test failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15382-20210219-125335-30656a46-ubuntu-2004/log.html#316)

martinpitt:
That is a really clever test! The fix looks nice and in the right direction for sure, but apparently not sufficient?

It's also worth grepping for a basic auth encoded version, to spot  [this gem](https://github.com/cockpit-project/cockpit/blob/master/src/bridge/cockpitpeer.c#L978).

mvollmer:
> It's also worth grepping for a basic auth encoded version, to spot [this gem](https://github.com/cockpit-project/cockpit/blob/master/src/bridge/cockpitpeer.c#L978).

That's a good idea!

mvollmer:
So the tests are finding the password in various buffers where it was left behind when moving the Cockpit protocol messages around. I think it would be quite some work to make sure that all involved code reliably overwrites everything as soon as it is no longer needed.

One interesting alternative would be to use asymmetric encryption: cockpit-bridge would create a one-use private key when making the request for a secret and would send the public key along with that request. As soon as the reply has come back and has been decrypted, the private key is overwritten.

martinpitt:
Right, the protocol messages are a nuisance, as they are really hard to control. A more tightly controlled side channel or similar is quite a lot of work though, so I think let's postpone that part. We should at least land the first commit though, as that already improves things. (This was meant to be a "low-hanging fruit" and "fix regressions from the sudo rework" thing, not a complete architectural overhaul).

Do the tests work if you restrict them to just the plaintext password for now, and we comment out the base64 variants for the time being?

How would asymmetric encryption work in principle? The remote end that you send the password to needs to decrypt it, so you'd need to send the private key along and are back to square 1.

As long as both ends are on the same machine, there would be a simpler way: put the password into a sealed memfd and send its fd around instead of the password itself. Although I guess most of our streams are not Unix sockets, so perhaps a real file in XDG_RUNTIME_DIR (to guarantee a tmpfs).

mvollmer:
> Do the tests work if you restrict them to just the plaintext password for now, and we comment out the base64 variants for the time being?

I think not, but I am not sure.  Since this is about finding stuff in unused memory, we can't guarantee anything really.  It might be clean now, and full of passwords tomorrow.

> How would asymmetric encryption work in principle? The remote end that you send the password to needs to decrypt it, so you'd need to send the private key along and are back to square 1.

The encryption would only be about the channel between JS in the browser and the part of the bridge that want a password.  The bridge will decrypt the password and then use it as it does now.  That might mean encrypting it immediately again for passing it on in a "authorize" message to sudo or cockpit-ssh.  The point is that we this scheme, we know exactly where the password is in memory, and can clean it there.  (Actually, when forwarding "authorize" messages, we might be able to skip the decryption/reencryption in the intermediate bridge.)

> As long as both ends are on the same machine, there would be a simpler way: put the password into a sealed memfd and send its fd around instead of the password itself. Although I guess most of our streams are not Unix sockets, so perhaps a real file in XDG_RUNTIME_DIR (to guarantee a tmpfs).

This wont work for sudo on a remote host, no?

martinpitt:
> The encryption would only be about the channel between JS in the browser and the part of the bridge that want a password. The bridge will decrypt the password and then use it as it does now.

Right, that's how I understood it. But for decryption you need the private key, which again would need to be passed through the normal stream protocol, so I don't see what this would win.

> This wont work for sudo on a remote host, no?

Right.

So this convinced me that this is far from easy to fix -- fundamentally by design we need to transmit credentials through streams, and as long as they use gio and copy-happy glib data structures, they are hard to contain. So for now, I think let's just land the first commit, and shelve this?

mvollmer:
> > The encryption would only be about the channel between JS in the browser and the part of the bridge that want a password. The bridge will decrypt the password and then use it as it does now.
> 
> Right, that's how I understood it. But for decryption you need the private key, which again would need to be passed through the normal stream protocol, so I don't see what this would win.

No, we would only need to send the public key.  But, yeah, we shouldn't invent our own crypto protocols.

> So this convinced me that this is far from easy to fix -- fundamentally by design we need to transmit credentials through streams, and as long as they use gio and copy-happy glib data structures, they are hard to contain. So for now, I think let's just land the first commit, and shelve this?

As discussed, let's also land the test but ignore it when it fails.

test/verify/cracklib.py
```diff
@@ -0,0 +1,36 @@
+# This file is part of Cockpit.
```
allisonkarlitskaya, line 1: Does this have anything to do with the very well-known library of the same name?

test/verify/cracklib.py
```diff
@@ -0,0 +1,36 @@
...
+import re
+import base64
+import subprocess
+
+
+def assert_no_password_in_bridges(machine, user, passwd):
+    passwd_b64 = base64.b64encode(passwd.encode("utf-8")).decode("ascii")
+    basic_auth = base64.b64encode((user + ":" + passwd).encode("utf-8")).decode("ascii")
+
+    regex = '%s\|%s\|%s' % (re.escape(passwd),
```
allisonkarlitskaya, line 27: With all of the various different dialects of regexps, from a quick look here, I can't tell if this will be properly accepted by the grep invocation below.  Could we also add a test for being able to find something which we know is in memory, just to make sure this is working?

test/verify/cracklib.py
```diff
@@ -0,0 +1,36 @@
...
+                            re.escape(passwd_b64),
+                            re.escape(basic_auth))
+
+    bridge_pids = machine.execute("pgrep cockpit-bridge").split()
+    for pid in bridge_pids:
+        try:
+            machine.execute("pid=%s; gcore $pid && ! grep '%s' core.$pid" % (pid, regex))
+        except subprocess.CalledProcessError:
+            print("Password found, but that's fine...")
```
allisonkarlitskaya, line 36: Is there any way we could make this a bit "louder" without resulting in an outright fail?

src/bridge/cockpitrouter.c
```diff
@@ -846,7 +847,12 @@ process_open (CockpitRouter *self,
         }
     }
   if (new_payload)
-    g_bytes_unref (new_payload);
```
allisonkarlitskaya, line 849: That's coming from cockpitjson, which also makes a billion other copies of it as it recursively traverses the data... I guess you know that, though.

In any case, I'd prefer to turn the logic inside out here: `GBytes` allows specifying a function to be called to free the data in question: we could use a function which clears the memory before it frees it.  Then we could just do a normal unref.  I'd find that a lot cleaner than the approach here (which might actually end up making an extra copy of the data, if a ref was held for some reason).

allisonkarlitskaya (changes requested):

allisonkarlitskaya:
In general, I feel like a good approach here might be to create some new security-enhanced `GBytes` utility functions (possibly in `cockpitmemory.c`) which store data in a region that will be cleared on free.  Then we could do our best to ensure that no password ever gets copied outside of those, unless it's to/from the kernel.

This might require some enhancements to cockpitjson to enable a "this blob might contain passwords" mode.

src/bridge/cockpitrouter.c
martinpitt: Right -- we already do this in e.g. cockpitpeer.c:
```c

  self->authorize_values = g_hash_table_new_full (g_str_hash, g_str_equal,
                                                  g_free, clear_authorize_value);
```

And `clear_authorize_value()` is exactly such a wrapper which calls `cockpit_memory_clear()` and then `g_free()`.

More specifically with GBytes, this is already done in cockpitwebservice.c in process_socket_authorize():
```c
      bytes = g_bytes_new_with_free_func (password, strlen (password),
                                          clear_and_free_string, password);
```

Expanding that to cockpitjson requires some more thought, though: A lot of that goes through `cockpit_json_write_bytes()` which calls `g_bytes_new_take()` which has no "free func" argument. So we can't just apply this to `new_payload` straight away. We'll also need an API change to mark "secret" JSON to go via `cockpit_memory_clear()`, or just live with the (probably negligible?) overhead and just always clear it. But either way we need to make cockpitjson less copy trigger-happy, and work more with read-only references?

PR #15383, opened 2021-02-19
Fix TestMachinesSettings.testCPUModel flake

KKoukiou:


KKoukiou:
ok It happened again once, so it does not really fix it completely. Still fedora-32/firefox had 70% failure and here is succeeded straight away so I think it makes sense to keep these fixes.

pkg/machines/components/vm/overview/cpuTypeModal.jsx
```diff
@@ -56,7 +61,7 @@ export const CPUTypeModal = ({ vm, models, close }) => {
                footer={
                    <>
                        {error && error.dialogError && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.dialogErrorDetail} />}
-                       <Button variant='primary' id="cpu-config-dialog-apply" onClick={save}>
+                       <Button variant='primary' isDisabled={isLoading} isLoading={isLoading} id="cpu-config-dialog-apply" onClick={save}>
```
martinpitt, line 64: What does the `isLoading` attribute do? The test does not use that, it should be happy with disabling the button while loading. Just a leftover?

martinpitt (changes requested):
Yay, thanks! Just one little cleanup to do.

pkg/machines/components/vm/overview/cpuTypeModal.jsx
KKoukiou: isLoading, adds a spinner to the button https://www.patternfly.org/v4/components/button/#progress 

The test does not do that, because that API call is instant normally, would be impossible to capture that.

KKoukiou: In other words, not a big deal more like to increase consistency with other modals.

martinpitt: I see, so this is actually legit UI. Thanks!

martinpitt (approved):

martinpitt:
Some [ubuntu-2004 machines failures](https://logs.cockpit-project.org/logs/pull-15383-20210219-162916-f1f3dc51-ubuntu-2004/log.html), retrying for comparison

PR #15384, opened 2021-02-19
Introduce livetest workflow

KKoukiou:


KKoukiou:
Ups I broke make_dist

test/make_dist.py
```diff
@@ -156,16 +157,23 @@ def download_dist():
...
+            sys.exit(1)
     return source
 
 
 if __name__ == '__main__':
-    print(make_dist())
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-d', dest='download-only', action='store_true')
+    args = parser.parse_args()
+    print(make_dist(getattr(args, 'download-only')))
```
martinpitt, line 179: use `args.download_only` instead of that getattr hack. With the appropriate long option from above, argparse automatically translates dashes into underscores, as that is a very common use case.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+              return result.data
+            } catch (err) {
+              core.setFailed(`Request failed with error ${err}`)
+            }
+
+      - uses: actions/checkout@v2
+        with:
+          repository: ${{ fromJSON(steps.get-pr.outputs.result).head.repo.full_name }}
+          fetch-depth: 0
+          ref: ${{ fromJSON(steps.get-pr.outputs.result).head.ref }} # or .head.ref for branch name
```
martinpitt, line 69: That is an odd comment. starter-kit does not have that, I guess it needs to be retrofitted? Otherwise it would always checkout master, not the PR, right?

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Create a new test user
+        # WARNING: if you want to add this users to sudoers make sure to propertly delete the SINK_SSHRELAY_PRIVATE
+        run: sudo useradd --create-home livetest; echo livetest:foobar | sudo chpasswd
+
+      - name: Download dists tarball for current git SHA from GitHub
+        id: make_dist
+        continue-on-error: true
+        run: ./test/make_dist.py -d
```
martinpitt, line 78: Please use a long option for better readability.

test/make_dist.py
```diff
@@ -156,16 +157,23 @@ def download_dist():
...
+        else:
+            print("make_dist: Download failed: artifact does not exist")
+            sys.exit(1)
     return source
 
 
 if __name__ == '__main__':
-    print(make_dist())
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-d', dest='download-only', action='store_true')
```
martinpitt, line 177: Please add `--download-only` as a long option, and drop `dest=`.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+          fetch-depth: 0
+          ref: ${{ fromJSON(steps.get-pr.outputs.result).head.ref }} # or .head.ref for branch name
+
+      - name: Create a new test user
+        # WARNING: if you want to add this users to sudoers make sure to propertly delete the SINK_SSHRELAY_PRIVATE
+        run: sudo useradd --create-home livetest; echo livetest:foobar | sudo chpasswd
+
+      - name: Download dists tarball for current git SHA from GitHub
+        id: make_dist
+        continue-on-error: true
```
martinpitt, line 77: This is a bit unwieldy and error prone. Rather drop this, the next step that writes a comment will run on failure anyway, due to the `if`.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Download of the dists artifact failed: Retry again in 10 minutes'
+            })
+        if: steps.make_dist.outcome == 'failure'
+
+      - run: exit 1
+        if: steps.make_dist.outcome == 'failure'
```
martinpitt, line 94: You can drop this when dropping `continue-on-error` above.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Install dependencies
+        run: |
+          sudo sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list
+          sudo apt update
+          sudo apt-get -y build-dep cockpit
+          ./tools/make-debs --quick
+          # FIXME: pcp is just utterly slow to start makes the installation fail
+          rm -f cockpit-pcp*.deb
+          for deb in cockpit*.deb; do sudo apt -y install ./$deb; done
```
martinpitt, line 104: Please do `sudo apt-get install -y cockpit*.deb` rather, so that apt can resolve dependencies properly.

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Create an SSH private key from the secret
+        run: |
+          sudo touch /run/SINK_SSHRELAY_PRIVATE && sudo chown $(id -u):$(id -g) /run/SINK_SSHRELAY_PRIVATE
+          echo "${{ secrets.SINK_SSHRELAY_PRIVATE }}" > /run/SINK_SSHRELAY_PRIVATE
+          chmod 600 /run/SINK_SSHRELAY_PRIVATE
+          mkdir -p ~/.ssh &&  ssh-keyscan -H ${{ env.RELAY_HOST }} > ~/.ssh/known_hosts
+
+      - name: Keep ssh connection alive
+        run: sudo sed -i -r 's/#ClientAliveInterval.*/ClientAliveInterval 600/' /etc/ssh/sshd_config
```
martinpitt, line 139: Great job figuring that out! I suppose that should go back to starter-kit and podman?

.github/workflows/livetest.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Forward cockpit port 9090 to a server we can access
+        run: ssh -f -R ':${{ steps.remote_port.outputs.SSH_GATEWAY_PORT }}':localhost:9090 -i /run/SINK_SSHRELAY_PRIVATE -o ExitOnForwardFailure=yes sshgw@${{ env.RELAY_HOST }} 'sleep infinity'
+
+      - name: Block workflow from finishing
+        run: sleep 4h
+
+      - name: Cleanup ssh process
+        if: ${{ always() }}
```
martinpitt, line 148: Doesn't hurt, but the `if:` condition does not need the  `${{ }}`, it's always a condition. You didn't use it in the `if:`s above either.

martinpitt (changes requested):
Super-cool!  I see that you have [some runs on your fork](https://github.com/KKoukiou/cockpit/actions) already -- looking forward to trying this in action!

.github/workflows/livetest.yml
KKoukiou: So, this is not really fixed. I think we might need the relevant options on the client. So if I open the cockpit link, it does not really time out, just after the workflow VMs do, 6h?

But if I don't open the link within 20min it seems that somehow the SSH tunnel breaks and then cockpit is not accessible.

.github/workflows/livetest.yml
KKoukiou: Right, this needs to be fixed for starter-kit and podman, they are both wrong now.

KKoukiou: Removed comment.

test/make_dist.py
```diff
@@ -156,16 +157,23 @@ def download_dist():
...
+            source = build_dist()
+        else:
+            print("make_dist: Download failed: artifact does not exist")
+            sys.exit(1)
     return source
 
 
 if __name__ == '__main__':
-    print(make_dist())
+    parser = argparse.ArgumentParser()
```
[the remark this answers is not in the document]
martinpitt: You added this string as help to `--download-only`, but I really meant to add it to `ArgumentParser()`. It summarizes the purpose of the program in `--help`.

test/make_dist.py
```diff
@@ -156,16 +157,23 @@ def download_dist():
...
+    parser.add_argument('--download-only', action='store_true', help="Download or build release tarball")
```
martinpitt, line 177: This should be the help for the particular -d option, thus "Fail instead of build locally if download is not available" or something such.

martinpitt: Also, please also add back `-d`. Short options are more convenient to type. I just like to use long options in widespread code, as they are easier to understand when reading.

.github/workflows/livetest.yml
martinpitt: Maybe it would help to run a shell command that regularly prints something? I'd need to  open that up on the sink, but that would not be a problem.

martinpitt (changes requested):

.github/workflows/livetest.yml
KKoukiou: Sure, let's try it, so let's replace 'sleep infinity' with 'while true; echo 'I am alive'; sleep 60; done'

KKoukiou: Or something similar.

martinpitt (dismissed):
Triggering f-33 test as it's required. This LGTM now, thank you! Let's land that, and figure out the keepalive thing separately, then it's easier to play around with this. Thank you!

martinpitt:
Triggering a debian-stable test as well, so that we have one that actually downloads a tarball --  then we cover both cases in the tests.

KKoukiou:
> Triggering a debian-stable test as well, so that we have one that actually downloads a tarball -- then we cover both cases in the tests.

Ehm the for loop change for installing the debs caused a failure in the installation,

```
E: Couldn't find any package by regex 'cockpit_238.52.gf5b630c35-1_all.deb'
```

https://github.com/KKoukiou/cockpit/runs/1962533759?check_suite_focus=true 

I will look at it tomorrow.

martinpitt:
https://logs.cockpit-project.org:4625/ is currently active. I am trying to log in, but I keep getting back to the login screen, hmm.

martinpitt (approved):
Nice, thank you! This works really well now! :tada:

PR #15385, opened 2021-02-19
test: Fix TestActivePages races and bugs

martinpitt:
With full chromium instead of chromium-headless, the `focus(".terminal")`
focused *away from* the actual element that receives the input, so that
typing the sleep command went into Nirvana. Just drop that, it's
unnecessary -- we want the Terminal page to automatically focus the
input. (check-system-terminal also does not focus explicitly).

Drop the weird `line_text()` helper, and wait until the terminal shows a
full prompt (with '`$`), instead of just "anything". This does the same
as check-system-terminal (albeit a bit simplified).

Wait for the default selection of the "delete frame" dialog to happen.
Previously the clickery started too early before the dialog got fully
initialized, and then missed to actually stop the terminal frame.

Fix the waiting for "sleep command goes away"; before it waited for the
command to be present, thus was a no-op.

Replace the broken `ps | grep` with `pgrep -x`, which avoids catching
itself. Despite the effort with using a regexp, the grep command still
found itself. Also, `sleep` is a bad program to use, as on Debian/Ubuntu
that just gets reparented to pid 1 after its parent (the terminal shell)
dies, so it never gets killed (this was hidden by the previous bug where
the second wait loop was the wrong way around). Replace this with `cat`,
which will properly exit on SIGPIPE, and rename it to `kitten` to
ensure that our pgrep does not catch some unrelated cat.

Use a single shell commmand to the VM instead of a heavy polling loop.

--- 

Broken out from #15378, as this test case was so broken that it needs a *lot* of attention. It's still not 100% right.

### DIFF
--- a/test/verify/check-shell-active-pages
+++ b/test/verify/check-shell-active-pages
@@ -69,27 +69,25 @@ class TestActivePages(MachineCase):
         b.go("/system/terminal")
         b.enter_page("/system/terminal")
 
-        b.focus(".terminal")
-
         def line_sel(i):
             return '.terminal .xterm-accessibility-tree div:nth-child(%d)' % i
 
-        def line_text(t):
-            return t + u'\xa0' * (80 - len(t))
-
         # wait for prompt in first line
-        b.wait_text_not(line_sel(1), line_text(""))
+        b.wait_visible(".terminal .xterm-accessibility-tree")
+        b.wait_in_text(line_sel(1), '$')
 
-        # run the sleep command
-        b.key_press("sleep 999999\r")
+        # run a command that we can easily identify, and which will die with the terminal
+        b.key_press("bash -c 'exec -a kitten cat'\r")
 
-        # wait for it to be present in ps
-        wait(lambda: m.execute("ps ax | grep 'sleep 99.*999'"))
+        # wait for command to run
+        m.execute("until pgrep -afx kitten; do sleep 1; done")
 
         # now close the page
         showPagesAssertCount(3 + n_extra_preloaded)
 
-        # click on the other page as well
+        # terminal should be pre-checked by default
+        b.wait_visible('tr[data-row-id="cockpit1:localhost/system/terminal"] input[type=checkbox]:checked')
+        # click on the main page as well
         b.click('tr[data-row-id="cockpit1:localhost/system"] input[type=checkbox]')
         # wait until it's highlighted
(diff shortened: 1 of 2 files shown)


martinpitt:
[This](https://logs.cockpit-project.org/logs/pull-15378-20210219-144940-b0b23039-fedora-33/log.html#325) is the remaining flake that I'm hunting now. I can reproduce this locally reasonably well.

martinpitt:
I fixed another two race conditions. Now this test finally survived 2x5 parallel runs, once with debian-testing and once with fedora-33. Let's see whether the bots finally agree.

marusak:
[coreos](https://logs.cockpit-project.org/logs/pull-15385-20210219-183716-c5caf340-fedora-coreos/log.html#325) does not agree

martinpitt:
@marusak : Yep, I know, I just didn't get  to that yesterday any more. Should be fixed now, I ran the test several times locally on coreos.

test/verify/check-shell-active-pages
```diff
@@ -19,28 +19,23 @@
...
 
     def testBasic(self):
         b = self.browser
         m = self.machine
 
-        self.login_and_go("/system")
+        # disable preloads; they are different between OSes, make the counting hard, and happen asynchronously
+        # disable_preload(), sed_file() etc. need writable file system
+        for p in ["packagekit", "playground", "systemd"]:
+            self.restore_dir("/usr/share/cockpit/" + p)
```
martinpitt, line 35: I was just thinking, this should actually go into the implementation of `disable_preload()`. We don't use it anywhere on OSTree until now. I'll wait for a round of test results and then update that.

martinpitt: Previous round's results all succeeded everywhere, pushed this cleanup.

test/verify/check-shell-active-pages
```diff
@@ -69,36 +69,38 @@ class TestActivePages(MachineCase):
...
-
         # wait for prompt in first line
-        b.wait_text_not(line_sel(1), line_text(""))
+        b.wait_visible(".terminal .xterm-accessibility-tree")
+        b.wait_in_text(line_sel(1), '$')
 
-        # run the sleep command
-        b.key_press("sleep 999999\r")
+        # run a command that we can easily identify, and which will die with the terminal
+        b.key_press("bash -c 'exec -a kitten cat'\r")
```
marusak, line 72: kitten - cat :smile:

marusak (approved):
Very nice job! It simplifies it and also makes it more robust.

PR #15391, opened 2021-02-22
storaged: Replace --with-vdo-package and --with-nfs-client-package configure option with manifest maps

martinpitt:
This gets rid of one part of dynamic manifests, and thus makes them more
portable (~/.local/share/cockpit) and easier to generate by webpack. It
also avoids an "unbreak my build" configure argument for CentOS/RHEL.
The manifest is a much more "quirkable" text file which can be handled
through overrides by admins who know what they are doing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -136,7 +136,6 @@ SUBST_RULE = \
 	-e 's,[@]selinux_config_type[@],$(COCKPIT_SELINUX_CONFIG_TYPE),g' \
 	-e 's,[@]with_appstream_config_packages[@],$(with_appstream_config_packages),g' \
 	-e 's,[@]with_appstream_data_packages[@],$(with_appstream_data_packages),g' \
-	-e 's,[@]with_nfs_client_package[@],$(with_nfs_client_package),g' \
 	$< > $@.tmp && $(MV) $@.tmp $@ \
 	$(NULL)
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -137,7 +137,6 @@ SUBST_RULE = \
 	-e 's,[@]with_appstream_config_packages[@],$(with_appstream_config_packages),g' \
 	-e 's,[@]with_appstream_data_packages[@],$(with_appstream_data_packages),g' \
 	-e 's,[@]with_nfs_client_package[@],$(with_nfs_client_package),g' \
-	-e 's,[@]with_vdo_package[@],$(with_vdo_package),g' \
 	$< > $@.tmp && $(MV) $@.tmp $@ \
 	$(NULL)
 
--- a/configure.ac
+++ b/configure.ac
@@ -350,13 +350,6 @@ AC_ARG_WITH(nfs-client-package,
             [with_nfs_client_package='false'])
 AC_SUBST(with_nfs_client_package)
 
-AC_ARG_WITH(vdo-package,
-            AC_HELP_STRING([--with-vdo-package=JSON],
-                           [Package that is necessary to use VDO, or 'false']),
-            [],
-            [with_vdo_package='false'])
-AC_SUBST(with_vdo_package)
-
 # Address sanitizer
 
 AC_MSG_CHECKING([for asan flags])
--- a/configure.ac
+++ b/configure.ac
@@ -343,13 +343,6 @@ AC_ARG_WITH(appstream-data-packages,
             [with_appstream_data_packages='[[]]'])
 AC_SUBST(with_appstream_data_packages)
 
-AC_ARG_WITH(nfs-client-package,
-            AC_HELP_STRING([--with-nfs-client-package=JSON],
-                           [Package that is necessary for NFS clients, or 'false']),
-            [],
-            [with_nfs_client_package='false'])
-AC_SUBST(with_nfs_client_package)
-
 # Address sanitizer
 
 AC_MSG_CHECKING([for asan flags])
--- a/pkg/storaged/manifest.json.in
+++ b/pkg/storaged/manifest.json.in
@@ -52,7 +52,11 @@
     },
 
     "config": {
-        "nfs_client_package": @with_nfs_client_package@,
+        "nfs_client_package": {
+            "rhel": "nfs-utils", "fedora": "nfs-utils",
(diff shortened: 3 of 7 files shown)


pkg/storaged/vdos-panel.jsx
```diff
@@ -24,16 +24,25 @@ import { SidePanelRow } from "./side-panel.jsx";
 import {
     fmt_size, decode_filename,
     get_available_spaces, prepare_available_spaces,
-    get_config
 } from "./utils.js";
 import { dialog_open, TextInput, SelectSpace, SizeSlider, CheckBoxes } from "./dialog.jsx";
 
 const _ = cockpit.gettext;
 
+function vdo_package_name() {
```
mvollmer, line 32: I am not a fan of moving this logic into the code itself... can we have a `quirks` file that can be controlled easily by the packaging scripts (spec and debian/rules)?

martinpitt: I'm not much of a fan of it myself, but having it in the spec is not exactly great either. For quirks we could perhaps have some middle ground -- one *can* put quirks into the manifest, but if there are none the page would look at the OS and auto-detect features? So that we can distribute a manifest that works everywhere, without having to build and carefully configure?

mvollmer: Isn' the spec file the exact right place for this?  That's where we know exactly whether whether vdo support is available what the package is called, no?

mvollmer: > For quirks we could perhaps have some middle ground -- one can put quirks into the manifest, but if there are none the page would look at the OS and auto-detect features? 

I would put the quirks in their own file, completely separate from the manifest.  It could contain values per OS, and we could distribute a reasonable default version of it.  Something like this:

```
{
  "rhel": {
    "vdo_package": "vdo"
  },
  "centos": {
    "vdo_package": "vdo"
  }
}
```

mvollmer: (But of course, YAGNI and all.  Just hard coding what we support into the code is a good start until we really find the need to be more flexible.)

martinpitt: Such a distro → config map seems ok and rather generic; we could e.g. use a similar approach for the appstream metadata packages as well. But IMHO it's overkill to invent a second parallel hierarchy of data files and an API to access it from pages -- we have all that with manifests only. My concern is to fix manifests to be *portable*, not making them as small as possible.

I disagree that the spec file is the right place for that -- vdo may appear on Fedora or someone may install some COPR, or vdo may disappear from CentOS stream or CentOS 9, etc.

I much prefer runtime detection over build/configure time, but apparently we need to agree to disagree here.

So I see two ways, either of which I'd support:

 - Drop the os-release test in favor of actually asking the packaging system "do you know about the vdo package"? On our "blank /var/cache" test images that always incurs a 30 second package metadata download, but real-life systems usually have that cache, and thus it may not actually be so bad.

- Put this back into the manifest, but in a per-distro map.

WDYT?

martinpitt: FTR, I'm currently leaning towards the second option (per-distro map in manifest), which is fast and efficient, and "quirkable".

mvollmer: > I much prefer runtime detection over build/configure time, but apparently we need to agree to disagree here.

No, we agree.  Run-time detection is better, but not reliably possible here, as far as I can see. I am just looking for a way to change the value of "vdo_package" without having to change the actual code, which I am afraid might be to painful for those who actually want to change it.

mvollmer: > FTR, I'm currently leaning towards the second option (per-distro map in manifest), which is fast and efficient, and "quirkable".

Yes, that's a good idea. We can tell people to install a override.json file if they ever need to tweak this.

mvollmer: So what about keeping the `get_config` API and just changing it's implementation to work with the distro/config map?

martinpitt: @mvollmer: That's what I'm doing right now. Nice!

pkg/storaged/vdos-panel.jsx
```diff
@@ -31,9 +31,12 @@ import { dialog_open, TextInput, SelectSpace, SizeSlider, CheckBoxes } from "./d
 const _ = cockpit.gettext;
 
 export function vdo_feature(client) {
+    const os_vdo_map = get_config("vdo_package", false);
+    const os_release = JSON.parse(window.localStorage['os-release'] || "{}");
```
mvollmer, line 35: My idea was that `get_config` already does all the os-release lookup, and this code here would stay unchanged.  Then we can super cleanly use the same mechanism immediately for "nfs_client_package".

mvollmer: Something like:

```
export function get_config(name, def) {
    if (cockpit.manifests.storage && cockpit.manifests.storage.config) {
        var val = cockpit.manifests.storage.config[name];
        if (typeof val === 'object' && val !== null) {
            const os_release = JSON.parse(window.localStorage['os-release'] || "{}");
            val = val[os_release.ID];
        }
        return val !== undefined ? val : def;
    } else {
        return def;
    }
}
```

mvollmer: Wdyt?

mvollmer (approved):

pkg/storaged/vdos-panel.jsx
martinpitt: I see! I didn't dare to do that yet before looking at the specifics of NFS, but indeed it makes sense to handle both of them at the same time, to make sure the `get_config()` API works for both of them. Landing this PR isn't super-urgent, so I'll do that first and repost.

martinpitt: Done, and added the corresponding thing for NFS. I am still testing that locally, though, I'll ask for re-review when ready.

martinpitt:
I manually tested this on fedora-33 and debian-testing, works fine.

martinpitt:
Manually tested on rhel-8-4 as well.

mvollmer (approved):

PR #15392, opened 2021-02-22
storaged: Drop obsolete --with-storaged-iscsi-sessions

martinpitt:
This was fixed in storaged 5 years ago, and all supported OSes have a
newer udisks version. This configure flag has not been actually used in
our spec/debian files for a long time.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -136,7 +136,6 @@ SUBST_RULE = \
 	-e 's,[@]wsinstancegroup[@],$(COCKPIT_WSINSTANCE_GROUP),g' \
 	-e 's,[@]admin_group[@],$(admin_group),g' \
 	-e 's,[@]selinux_config_type[@],$(COCKPIT_SELINUX_CONFIG_TYPE),g' \
-	-e 's,[@]with_storaged_iscsi_sessions[@],$(with_storaged_iscsi_sessions),g' \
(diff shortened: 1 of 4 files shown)


martinpitt:
FTR, our spec has `Requires: udisks2 >= 2.6` which ensures that we have the runtime API.

mvollmer:
> This was fixed in storaged 5 years ago, and all supported OSes have a
> newer udisks version.

IIRC, this is for a missing feature in a library used by UDisks, and UDisks would adapt itself at compile time accordingly.  Thus, we can't look at the version of UDisks alone to figure out whether that feature is 
present at run-time or not.

> This configure flag has not been actually used in our spec/debian files for a long time.

But this should be enough to indicate that we can drop the option.

(We used it only on rhel iirc, I'll try to find the bug for it.)

mvollmer:
> > This was fixed in storaged 5 years ago, and all supported OSes have a
> > newer udisks version.
> 
> IIRC, this is for a missing feature in a library used by UDisks, and UDisks would adapt itself at compile time accordingly. Thus, we can't look at the version of UDisks alone to figure out whether that feature is
> present at run-time or not.

Ouch, new enough versions of UDisks _would_ tell us a t run-time whether that feature is compiled in or not, so we don't need our own compile time config... sorry, I was confused.

mvollmer (approved):

PR #15393, opened 2021-02-22
test: add checks when adding disk and no storage pools

yunmingyang:


KKoukiou (approved):
This looks good, I remember we had a bugzilla about it but I can't find it. Can you maybe post it here if it exists so that I can add it to the commit message on merge?

test/verify/check-machines-disks
```diff
@@ -210,6 +210,18 @@ class TestMachinesDisks(VirtualMachinesCase):
...
 
+        # Check when no storage pool and adding existed disk 
+        # delete the default pool
+        m.execute("virsh pool-destroy images && virsh pool-undefine images")
+        # Open "add disk" dialog
+        b.click("#vm-subVmTest1-disks-adddisk")
+        b.click("#vm-subVmTest1-disks-adddisk-useexisting")
+        # Check
+        b.wait_not_present("#navbar-oops")
+        b.wait_visible("#vm-subVmTest1-disks-adddisk-existing-select-pool:disabled")
```
KKoukiou, line 221: This is unneccessary, since the next line will fail if that does not exist.

PR #4964, opened 2016-09-05
bridge: Implement minimum package version checks

stefwalter:
This is so packages can check that the Cockpit bridge and base package are at least a given version. We expect the bridge and base package to be shipped together.

PR #15396, opened 2021-02-22
pkg: drop "version" from manifests

allisonkarlitskaya:
This field was introduced in #4964 as "purely informational for now",
and isn't even parsed by cockpit.  Meanwhile, it's one of the few things
preventing us from untangling our webpack build from autotools.

Even within cockpit, usage is inconsistent: most packages set "version"
to the string that @Version@ expands to, but machines had it set to the
integer 0, and pcp and static had no "version" field at all.

It seems nobody is going to miss this.  Just drop it.

The one place which did reference this was a test which directly read
the manifest.json.in file from the playground and does a direct
comparison on it.  That's been adjusted accordingly now.

 - [x] Drop obsolete remote login compat code, which reads base1 manifest version: PR #15423

### DIFF
--- a/pkg/apps/manifest.json.in
+++ b/pkg/apps/manifest.json.in
@@ -1,5 +1,4 @@
 {
-    "version": "@VERSION@",
     "requires": {
 	"cockpit": "122"
     },
--- a/pkg/base1/manifest.json.in
+++ b/pkg/base1/manifest.json.in
@@ -1,4 +1,5 @@
 {
-    "version": "@VERSION@",
+    "version": "219",
+    "version-note": "last API change: https://github.com/cockpit-project/cockpit/pull/13482",
     "priority": -1
 }
--- a/pkg/base1/test-http.js
+++ b/pkg/base1/test-http.js
@@ -27,7 +27,6 @@ QUnit.test("simple request", function (assert) {
     cockpit.http({ internal: "/test-server" }).get("/pkg/playground/manifest.json.in")
             .done(function(data) {
                 assert.deepEqual(JSON.parse(data), {
-                    version: "@VERSION@",
                     requires: {
                         cockpit: "122"
                     },
--- a/pkg/kdump/manifest.json.in
+++ b/pkg/kdump/manifest.json.in
@@ -1,5 +1,4 @@
 {
-    "version": "@VERSION@",
     "requires": {
         "cockpit": "122"
     },
--- a/pkg/machines/manifest.json.in
+++ b/pkg/machines/manifest.json.in
@@ -1,5 +1,4 @@
 {
-  "version": 0,
   "priority": 0,
   "requires": {
        "cockpit": "186"
--- a/pkg/metrics/manifest.json.in
+++ b/pkg/metrics/manifest.json.in
@@ -1,5 +1,4 @@
 {
-    "version": "@VERSION@",
     "requires": {
         "cockpit": "137"
     }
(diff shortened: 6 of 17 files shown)


martinpitt:
Several failures, moving to draft. I also fixed the PR description.

martinpitt:
The tests work again when I just revert pkg/base1/manifest.json.in , i.e. put back VERSION there. That seems to be the only relevant one. I can't immediately pinpoint code that would read it, though.

martinpitt:
[login page's machine_application_login_reload()](https://github.com/cockpit-project/cockpit/blob/master/pkg/static/login.js#L857) looks at the base1 manifest version. This is ancient code which we really don't need any more. I removed that code locally and at least the handful of tests that I ran locally work fine now.

martinpitt:
With [this patch](https://piware.de/tmp/0001-login-Drop-compat-code-for-version-120.patch) most of the failing tests should work again, but locally I still get a failure with `test/verify/check-superuser TestSuperuserOldWebserver.test`. But, incremental improvement..

martinpitt:
Force-pushed to include that commit, let's see what's left now. In the worst case, we just keep the version in base1 and teach `make dist` about substituting it?

martinpitt:
I split out dropping the compat code to #15423 to get it  tested separately.

martinpitt:
Looking at the [TestSuperuserOldWebserver.test failure](https://logs.cockpit-project.org/logs/pull-15396-20210225-084247-b6fbd834-fedora-33/log.html#265-2), some notes.

 - With a sit() at the top of the test and hacking version 238 back into /usr/share/cockpit/base1/manifest.json, the test works, so it's definitively a base1 version check.
 - Setting  version: 100 makes it fail, so it's not just about the field being present, the version matters.
 - PR #13482 introduced this check, but neither contains a relevant reference to "base1" nor to "version".
 - Commit 84b61958655f102e seems to not use a version number, but the `plain1` vs. `superuser` messages in the bridge init.

@mvollmer, do you know something off the top of your head what could require the bridge manifest's version here?

martinpitt:
This reproduces interactively. Clicking the menu on the remote machine on http://127.0.0.2:9091/cockpit+=10.111.113.2/@localhost/shell/index.html most often does not do anything; on some pages it works, but it always takes two or five clicks to eventually succeed. Also, the URL does not change at all.

My theory is that this is somehow related to 4511d83ba39ba673fd47b16fdde072f934e737dd -- while the new (remote) bridge does not have that any more, the old centos-7 bridge certainly would.

As such it seems to me that we'd first need to drop that backwards compat code from all the old bridges which are around, which given RHEL 7 will still be a while.

My proposal: Put back a static "version: "219" field into base1/manifest.json, but stop treating it as a package version, and consider it a legacy "API version" field for older releases, until RHEL/CentOS 7 go EOL. "219" because that's when the behaviour was last modified in PR #13482.

@allisonkarlitskaya , @mvollmer , WDYT?

mvollmer:
> WDYT?

I can't say anything about this, unfortunately.  Putting a static version just into base1/manifest certainly looks reasonable.

martinpitt:
Done, with this it ought to succeed now. Marking as ready for review. @mvollmer, want to give it a final review, as both Lis and I touched this now?

martinpitt:
[this is an unrelated failure](https://logs.cockpit-project.org/logs/pull-15396-20210225-130645-a57302ee-rhel-8-4/log.html#60). I [reported it to RHEL](https://bugzilla.redhat.com/show_bug.cgi?id=1933058) and [sent a naughty](https://github.com/cockpit-project/bots/pull/1723).

martinpitt:
This is about as green as tests can get right now, this looks good to land.@mvollmer, can you please review?

mvollmer (approved):

PR #15400, opened 2021-02-23
Some machines tests improvements

KKoukiou:


test/verify/check-machines-create
```diff
@@ -1060,7 +1060,7 @@ class TestMachinesCreate(VirtualMachinesCase):
                 wait(lambda: dialog.name in self.machine.execute("virsh list --persistent"), delay=3)
 
                 # unfinished install script runs indefinitelly, so we need to force it off
-                self.performAction(name, "forceOff")
+                self.test_obj.performAction(name, "forceOff", False)
```
martinpitt: That testObj change looks like it should have been part of the previous commit?

martinpitt (approved):
This does not yet address the [check-machines-lifecycle flake](https://logs.cockpit-project.org/logs/pull-15400-20210224-064403-62bfdc69-centos-8-stream/log.html#61) where it seems to have trouble clicking the "Clone" menu entry and get the dialog. That's exactly what this PR touches. However, this centralization does make sense, makes tests easier to read/write, and makes it easier to fix the option picking for all tests. Thanks!

test/verify/check-machines-create
KKoukiou: Ups yes, let's just squash on merge.

PR #15401, opened 2021-02-23
lib: Check installed Red Hat products instead of yum config file

martinpitt:
The `enabled=` option in subscription-manager.conf is not a reliable
indicator whether the installed system needs any subscriptions. On at
least the qcow image, the dnf plugins (both subscription-manager and
product-id) are disabled by default, and the first invocation of
subscription-manager auto-enables them.

Instead, ask RHSM if there are any installed Red Hat products on the
system. That list is empty on Fedora/CentOS. As before, if RHSM itself
is absent, quietly consider this as "system does not need subscription".

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1931429

### DIFF
--- a/pkg/lib/packagekit.js
+++ b/pkg/lib/packagekit.js
@@ -310,56 +310,51 @@ export function getSeverityIcon(info, secSeverity, iconSize = "fa-sm") {
         return `pficon pficon-enhancement ${iconSize}`;
 }
 
-const yum_plugin_enabled_re = /^\s*enabled\s*=\s*1\s*$/m;
-
 /**
  * Check Red Hat subscription-manager if if this is an unregistered RHEL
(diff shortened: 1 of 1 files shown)


martinpitt:
Triggering an additional test against the new rhel-8-4 image in https://github.com/cockpit-project/bots/pull/1710 , which has the enabled=1 hack reverted.

mvollmer (approved):

martinpitt:
[some regressions](https://logs.cockpit-project.org/logs/pull-15401-20210223-093850-0d8a68eb-rhel-8-4-bots-1710/log.html), but unrelated: check-machines-lifecycle is the top flake right now, and the journal failures are a regression in RHEL 8.4 which I will investigate in the bots refresh PR.

PR #15402, opened 2021-02-23
cockpituous: bodhi for Fedora 34, drop Fedora 32

martinpitt:
bodhi activation for Fedora 34 is today [1]. Drop Fedora 32 instead,
which is soon EOL.

[1] https://fedorapeople.org/groups/schedule/f-34/f-34-key-tasks.html


marusak (approved):
Time flies!

PR #15404, opened 2021-02-23
services: rewrite create timer dialog to react

KKoukiou:


martinpitt (approved):
This is amazing work, thank you! Last time I looked, firefox test was green three times in a row! Let's see how the re-run looks like.

martinpitt:
Oh, I see -- this is so hard to get green because it "affects" machines, due to the CSS cleanup. I suppose we'll have to land this with some red, and careful inspection.

PR #15407, opened 2021-02-23
machines: Live VM migration

skobyda:
Transport live VM to another host. Configurable options are:
- Destination URI
- Migrated VM's state: This defines how will VM look at the destination
  host. It can either be persistent and running, transient and running
  or it can be shut off. Note that if shut off option is chosen, then
  the VM on source host will be paused for the duration of the
  migration and "Copy storage" is disabled.
- Storage: User can either choose to copy all read-only disks, or they
  can choose that they have a shared storage setup where both source and
  destination hosts have access too. In the latter that case, no disks
  are transported.
- Undefine source: VM at the source host will be undefined after
  migration is finished. Not this option is enabled when state of
  Migrated VM is "transient".

Screenshot from 2021-02-23 14-52-59
Screenshot from 2021-02-23 14-53-32

And the things I would like to discuss: 
1. virt-manager offers also data transport type option, where user can choose if they want to use hypervisor native transport or tunnelled transport. But after looking at migration related topics on serverfault.com, stackoverflow.com and various blogs, I get the feeling it's not so much used options, so perhaps it's possible to omit it?
2. Perhaps "Undefine source" can be merged with some other option? Like apart from "Live transient", it could automatically undefine source VM once the migration is succesfully finished? Or is that too risky move?
3. There is also an "unsafe" flag. Is some cases, libvirt might decide that migration is too unsafe and may lead to data corruption. It's safe to migrate without data corruption when filesystems are cache-coherent (i.e. GFS2, OCFS or CEPH) and disks have cache option configured (disk XML property is `cache != 'none'`). If that's not the case, migration may lead to data corruption. What should be done about this situation? I can check the above mentioned conditions, and if they pass, I can either pass "unsafe" flag automatically + show warning that migration may lead to data corruption. Or I can display the "unsafe" option in Migrate dialog.
4. In this PR the migration is enabled only for running VM. But I'm not really sure if it should be disabled... Migrating shut off VM only takes xml from source host and moves it to the destination host and it cannot copy disks. So the same can be achieved by running:
```
virsh dumpxml FOO > foo.xml
scp foo.xml <targethost>
virsh define foo.xml
```

### DIFF
--- a/pkg/machines/components/vm/vmMigrateDialog.jsx
+++ b/pkg/machines/components/vm/vmMigrateDialog.jsx
@@ -1,0 +1,193 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from 'cockpit';
+import React, { useState } from 'react';
+import {
+    Button,
+    Checkbox,
+    Form,
+    FormGroup,
+    Modal,
+    Radio,
+    TextInput,
+    Tooltip
+} from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from "@patternfly/react-icons";
+
+import { migrateToUri } from '../../libvirt-dbus.js';
+import { isEmpty, isObjectEmpty } from '../../helpers.js';
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+
+const _ = cockpit.gettext;
+
+const DestUriRow = ({ validationFailed, destUri, setDestUri }) => {
+    return (
+        <FormGroup label={_("Destination URI")} fieldId="dest-uri"
+                   id="dest-uri"
+                   validated={validationFailed.destUri ? "error" : "default"}>
+            <TextInput id='dest-uri-input'
+                       validated={validationFailed.destUri ? "error" : "default"}
+                       value={destUri}
+                       placeholder={cockpit.format(_("Example, $0"), "qemu+ssh://192.0.2.16/system")}
+                       onChange={setDestUri} />
+        </FormGroup>
+    );
+};
+
+const OptionsRow = ({ temporary, live, setTemporary, setLive, setStorage }) => {
+    return (
+        <FormGroup label={_("Options")}
+                   fieldId="options"
+                   id="options"
+                   hasNoPaddingTop>
+            <Checkbox id="live"
+                      isChecked={live}
+                      label={_("Run after migration")}
+                      onChange={b => {
+                          setLive(b);
+                          if (!b)
+                              setStorage("shared");
+                      }} />
+            {live && <Checkbox id="temporary"
+                               isChecked={temporary}
+                               label={
+                                   <>
+                                       {_("Move temporarily")}
+                                       <Tooltip
+                                           position="top"
+                                           content={
+                                               <>
+                                                   <p>{_("When shut off on the destination, virtual machine will return to the original host in a shut off state.")}</p>
+                                                   <p>{_("This feature is often used for hardware maintanance or load-balacing.")}</p>
+                                               </>}>
+                                           <OutlinedQuestionCircleIcon id="migrate-tooltip" />
+                                       </Tooltip>
+                                   </>
+                               }
+                               onChange={setTemporary} />
+            }
+        </FormGroup>
+    );
+};
+
+const StorageRow = ({ storage, setStorage, live }) => {
+    const copyDisabled = !live;
+    let copyRadio = (
+        <Radio id="copy"
+               name="source"
+               label={_("Copy storage")}
+               isDisabled={copyDisabled}
+               isChecked={storage === "copy"}
+               onChange={() => setStorage("copy")} />
+    );
+
+    if (copyDisabled) {
+        copyRadio = (
+            <Tooltip id="storage-copy-tooltip"
+                     content={_("Offline migration doesn't allow copying storage")}>
+                {copyRadio}
+            </Tooltip>
+        );
+    }
+
+    return (
+        <FormGroup label={_("Storage")}
+                   fieldId="storage"
+                   id="storage"
+                   hasNoPaddingTop>
+            <Radio id="shared"
+                   name="source"
+                   label={_("I have shared storage set up")}
+                   isChecked={storage === "shared"}
+                   onChange={() => setStorage("shared")} />
+            {copyRadio}
+        </FormGroup>
+    );
+};
+
+export const MigrateDialog = ({ vmId, connectionName, toggleModal }) => {
+    const [destUri, setDestUri] = useState("");
+    const [error, setDialogError] = useState({});
+    const [inProgress, setInProgress] = useState(false);
+    const [storage, setStorage] = useState("shared");
+    const [temporary, setTemporary] = useState(false);
+    const [live, setLive] = useState(true);
+    const [validationFailed, setValidationFailed] = useState(false);
+
+    function validateParams() {
+        const validation = {};
+        if (isEmpty(destUri.trim()))
+            validation.destUri = _("Destination URI must not be empty");
+
+        return validation;
+    }
+
+    function onMigrate() {
+        if (!isObjectEmpty(validateParams())) {
+            setValidationFailed(true);
+            return;
+        }
+
+        setInProgress(true);
+        return migrateToUri(connectionName, vmId, destUri, storage, live, temporary)
+                .then(toggleModal, exc => {
+                    setInProgress(false);
+                    setDialogError({ dialogError: _("Migration failed"), message: exc.message });
+                });
+    }
+
+    const body = (
+        <Form isHorizontal>
+            <DestUriRow destUri={destUri}
+                        setDestUri={setDestUri}
+                        validationFailed={validationFailed} />
+            <StorageRow storage={storage}
+                        setStorage={setStorage}
+                        live={live} />
+            <OptionsRow temporary={temporary}
+                        live={live}
+                        setStorage={setStorage}
+                        setTemporary={setTemporary}
+                        setLive={setLive} />
+        </Form>
+    );
+
+    const footer = (
+        <>
+            {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.message} />}
+            <Button variant='primary'
+                    isLoading={inProgress}
+                    onClick={onMigrate}>
+                {_("Migrate")}
+            </Button>
+            <Button variant='link' onClick={toggleModal}>
+                {_("Cancel")}
+            </Button>
+        </>
+    );
+
+    return (
+        <Modal position="top" variant="medium" isOpen onClose={toggleModal}
+           title={_("Migrate VM to another host")}
+           footer={footer}>
+            {body}
+        </Modal>
+    );
+};
--- a/pkg/machines/libvirt-dbus.js
+++ b/pkg/machines/libvirt-dbus.js
@@ -165,6 +165,25 @@ const Enum = {
     VIR_NETWORK_EVENT_LAST: 4,
     // Keycodes
     VIR_KEYCODE_SET_LINUX: 0,
+    // Migrate
+    VIR_MIGRATE_LIVE: 1,
+    VIR_MIGRATE_PEER2PEER: 2,
+    VIR_MIGRATE_TUNNELLED: 4,
+    VIR_MIGRATE_PERSIST_DEST: 8,
+    VIR_MIGRATE_UNDEFINE_SOURCE: 16,
+    VIR_MIGRATE_PAUSED: 32,
+    VIR_MIGRATE_NON_SHARED_DISK: 64,
+    VIR_MIGRATE_NON_SHARED_INC: 128,
+    VIR_MIGRATE_CHANGE_PROTECTION: 256,
+    VIR_MIGRATE_UNSAFE: 512,
+    VIR_MIGRATE_OFFLINE: 1024,
+    VIR_MIGRATE_COMPRESSED: 2048,
+    VIR_MIGRATE_ABORT_ON_ERROR: 4096,
+    VIR_MIGRATE_AUTO_CONVERGE: 8192,
+    VIR_MIGRATE_RDMA_PIN_ALL: 16384,
+    VIR_MIGRATE_POSTCOPY: 32768,
+    VIR_MIGRATE_TLS: 65536,
+    VIR_MIGRATE_PARALLEL: 131072,
 };
 
 const LIBVIRT_DBUS_PROVIDER = {
@@ -1520,6 +1539,28 @@ function initResource(connectionName, method, updateOrAddMethod, flags) {
                 return Promise.all(objPaths[0].map(() => updateOrAddMethod({})));
             })
             .catch(ex => console.warn('initResource action failed:', ex.toString()));
+}
+
+export function migrateToUri(connectionName, objPath, destUri, storage, live, temporary) {
+    // direct migration is not supported by QEMU, so it's opposite, the P2P migration should always be used
+    let flags = Enum.VIR_MIGRATE_PEER2PEER;
+
+    if (!live) {
+        flags = flags | Enum.VIR_MIGRATE_OFFLINE | Enum.VIR_MIGRATE_PERSIST_DEST;
+    } else {
+        flags = flags | Enum.VIR_MIGRATE_LIVE;
+
+        if (!temporary)
+            flags = flags | Enum.VIR_MIGRATE_PERSIST_DEST;
+    }
+
+    if (storage === "copy")
+        flags = flags | Enum.VIR_MIGRATE_NON_SHARED_DISK;
+
+    if (temporary)
+        flags = flags | Enum.VIR_MIGRATE_UNDEFINE_SOURCE;
+
+    return call(connectionName, objPath, 'org.libvirt.Domain', 'MigrateToURI3', [destUri, {}, flags], { timeout, type: 'sa{sv}u' });
 }
 
 export function networkActivate(connectionName, objPath) {
--- a/test/verify/check-machines-lifecycle
+++ b/test/verify/check-machines-lifecycle
@@ -22,11 +22,15 @@ import subprocess
 import parent
 from machineslib import *
 from testlib import *
+from machine_core.constants import TEST_OS_DEFAULT
 
 @skipImage("Atomic cannot run virtual machines", "fedora-coreos")
-@nondestructive
 @no_retry_when_changed
 class TestMachinesLifecycle(VirtualMachinesCase):
+    provision = {
+        "machine1": {"address": "10.111.113.1/20", "dns": "10.111.112.100"},
+        "destination_host": {"image": TEST_OS_DEFAULT, "address": "10.111.113.2/20", "dns": "10.111.112.100"},
+    }
 
     def testBasic(self):
         self._testBasic()
@@ -98,30 +102,39 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         wait(lambda: "login as 'cirros' user." in self.machine.execute("cat {0}".format(args["logfile"])), delay=3)
 
         # Send Non-Maskable Interrupt (no change in VM state is expected)
-        self.performAction("subVmTest1", "sendNMI")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-sendNMI")
 
         b.wait(lambda: "NMI received" in self.machine.execute("cat {0}".format(args["logfile"])))
 
         # pause
-        self.performAction("subVmTest1", "pause")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-pause")
+        b.wait_in_text("#vm-subVmTest1-state", "Paused")
 
         # resume
-        self.performAction("subVmTest1", "resume")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-resume")
+        b.wait_in_text("#vm-subVmTest1-state", "Running")
 
         # reboot
         self.machine.execute("echo '' > {0}".format(args["logfile"]))
-        self.performAction("subVmTest1", "reboot")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-reboot")
         wait(lambda: "reboot: Power down" in self.machine.execute("cat {0}".format(args["logfile"])), delay=3)
         b.wait_in_text("#vm-subVmTest1-state", "Running")
 
         # force reboot
         self.machine.execute("echo '' > {0}".format(args["logfile"]))
-        self.performAction("subVmTest1", "forceReboot")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-forceReboot")
         wait(lambda: "Initializing cgroup subsys" in self.machine.execute("cat {0}".format(args["logfile"])), delay=3)
         b.wait_in_text("#vm-subVmTest1-state", "Running")
 
         # shut off
-        self.performAction("subVmTest1", "forceOff")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-forceOff")
+        b.wait_in_text("#vm-subVmTest1-state", "Shut off")
 
         # continue shut off validation - usage should drop to zero
         b.wait_in_text(".memory-usage-chart .pf-c-progress__status > .pf-c-progress__measure", "0 /")
@@ -131,12 +144,14 @@ class TestMachinesLifecycle(VirtualMachinesCase):
         m.execute("virsh dumpxml subVmTest1 > /tmp/subVmTest1.xml")
         m.execute("virsh start {0} && virsh undefine {0}".format("subVmTest1"))
         b.wait_visible("div[data-vm-transient=\"true\"]")
-        self.performAction("subVmTest1", "forceOff", False)
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-forceOff")
         b.wait_in_text("#virtual-machines-listing .pf-c-empty-state", "No VM is running")
         m.execute("virsh define --file /tmp/subVmTest1.xml")
 
         # clone
-        self.performAction("subVmTest1", "clone")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-clone a[aria-disabled=false]")
 
         b.wait_text(".pf-c-modal-box__title-text", "Create a clone VM based on subVmTest1")
         b.click("footer button.pf-m-primary")
@@ -166,7 +181,9 @@ class TestMachinesLifecycle(VirtualMachinesCase):
             b.wait_in_text("#vm-subVmTest2-state", "Running")
 
         # stop second VM, event handling should still work
-        self.performAction("subVmTest2", "forceOff")
+        b.click("#vm-subVmTest2-action-kebab button")
+        b.click("#vm-subVmTest2-forceOff")
+        b.wait_in_text("#vm-subVmTest2-state", "Shut off")
 
         b.click("#vm-subVmTest2-run")
 
@@ -330,7 +347,8 @@ class TestMachinesLifecycle(VirtualMachinesCase):
 
         addDisk(secondDiskVolName, poolName)
 
-        self.performAction(name, "delete")
+        b.click("#vm-{0}-action-kebab button".format(name))
+        b.click("#vm-{0}-delete".format(name))
 
         b.wait_visible("#vm-{0}-delete-modal-dialog .modal-body:contains(The VM is running)".format(name))
         b.wait_visible("#vm-{1}-delete-modal-dialog ul li:first-child #disk-source-file:contains({0})".format(img2, name))
@@ -358,7 +376,8 @@ class TestMachinesLifecycle(VirtualMachinesCase):
 
         self.machine.execute("virsh -c qemu:///system suspend {0}".format(name))
         b.wait_in_text("#vm-{0}-state".format(name), "Paused")
-        self.performAction(name, "delete")
+        b.click("#vm-{0}-action-kebab button".format(name))
+        b.click("#vm-{0}-delete".format(name))
         b.click("#vm-{0}-delete-modal-dialog button:contains(Delete)".format(name))
         self.waitVmRow(name, 'system', False)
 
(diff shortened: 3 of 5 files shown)


KKoukiou:
@skobyda I would remove some of the options I see above.

-- persistent flag should be hidden I believe. If the VM is transient on the host let's make it transient on destination. If it's persistent on the host let's make it persistent on the destination.

-- undefinesource flag which is the last option you expose is also a bit tricky. Using that will just undefine the VM on the source host, not stop it, which is a concept that we don't expose so far in the cockpit UI (transient VMs). I would say, we should not expose it, and if a user wants to remove it they can delete it manually.

How about exposing:

--suspend        do not restart the domain on the destination host. This I find a more useful usecase. I expect users to not want the same service VM to be running at both source and dest, we need to make it easy, that either the source it stopped when the dest is VM is running, or allow the to suspend the dest. The first option makes much more sense to me, but I would not find an option to do that.

-- copy storage should probably have sub options the list of disks the user would like to copy: --migrate-disks is in virsh. Otherwise what does it copy actually?

--unsafe I would not enable neither expose it to the user, If it fails the will see an error and they can fix it. If libivrt is too strict about errors let's consider how to expose that in the future.

--change-protection should be enabled I feel, users should not change the domain config while migration is happening

skobyda:
> @skobyda I would remove some of the options I see above.
> 
> -- persistent flag should be hidden I believe. If the VM is transient on the host let's make it transient on destination. If it's persistent on the host let's make it persistent on the destination.

After some thinking I agree. In 99% cases, users do not want transient VM, and as you mentioned it below, we don't really have that concept in cockpit (and have no issues about so maybe truly vast majority of users do not care about that). So I would now have 2 options there: Shutoff and Live persistent.
 
> -- undefinesource flag which is the last option you expose is also a bit tricky. Using that will just undefine the VM on the source host, not stop it, which is a concept that we don't expose so far in the cockpit UI (transient VMs). I would say, we should not expose it, and if a user wants to remove it they can delete it manually.

Oh. After migration is completed, the source VM is forced off, so undefining it will always make it disappear. So I guess my wording of "Undefine source", "VM on this host will cease to exist after shutdown" is confusing and probably wrong. I would keep that option, but I would reword it to "Undefine source" or "Delete source VM" and "VM on this host will cease to exist" . But I think this is fairly used option, so I would keep it.
 
> How about exposing:
> 
> --suspend do not restart the domain on the destination host. This I find a more useful usecase. I expect users to not want the same service VM to be running at both source and dest, we need to make it easy, that either the source it stopped when the dest is VM is running, or allow the to suspend the dest. The first option makes much more sense to me, but I would not find an option to do that.

That's not what --suspend does. The thing is that before the migration is succesfully finished, the VM is paused. And after migration is finnished, the VM is resumed (meaning the CPU is restarted). Manpages says that "do not restart the domain on the destination host", but by that it means that CPU is not restarted, so the VM stays in paused state. And that I think is not really a use case which is worth exposing.
So to assure you, the VM is not defaultly rebooted after migration. It's defaultly resumed and continues it's activity with minimal downtime.

> -- copy storage should probably have sub options the list of disks the user would like to copy: --migrate-disks is in virsh. Otherwise what does it copy actually?

I went through many serverfault and stackoverflow and related to libvirt migration, and never saw somebody use that option. I guess that was far from a proper market research to find out what functionality users want, but I still have feeling that --migrate-disks is rarely used. 
Also read-only disks are not transferable by libvirt, so only writeable disks are transferred. 
Since users want to smoothly migrate VM, and have that VM do the same functionality, they would need all of writeable disks which source VM had. 
 
> --unsafe I would not enable neither expose it to the user, If it fails the will see an error and they can fix it. If libivrt is too strict about errors let's consider how to expose that in the future.

Alright. I've found out more info about unsafe migration and I agree, I will not expose it.
Basically I see 2 general cases when migration can be considered unsafe and how to fix:
- User doesn't have a filesystem which supports cache coherent migration. They would have to set up a shared storage (perhaps that can be done thru storage page?)
- User has a disk which which has cache option != None. So I would add option to UI where user can edit that for a disk.

> --change-protection should be enabled I feel, users should not change the domain config while migration is happening

When using libvirt API, that flag is used by default: VIR_MIGRATE_CHANGE_PROTECTION - `This flag is used automatically when both sides support it. Explicitly setting this flag will cause migration to fail if either the source or the destination does not support it.` New libvirt versions definitely support, so we can rely on it being automatic.

KKoukiou (commented):
I agree with the comments  you sent, and by a far look the code makes sense as well. So let's do the cleanup, fix UI misalignments and issues, add tests and I will give it a proper review. Thanks! This looks great!

garrett:
Labels are vertically misaligned. @KKoukiou what did you add to other forms to fix this kind of error? Was it `isInline` on the `formGroup` element?

KKoukiou:
> Labels are vertically misaligned. @KKoukiou what did you add to other forms to fix this kind of error? Was it `isInline` on the `formGroup` element?

hasNoPaddingTop.

skobyda:
Fixed the padding: 
Screenshot from 2021-03-16 12-33-46

garrett:
> Fixed the padding:

???

Is that an old screenshot? It's not fixed.

Here's a zoomed-in illustration of the baselines:

Screenshot from 2021-03-16 12-43-44

While Destination URI and Delete source VM are due to a PF label alignment bug (and off by 1 to 2 pixels, based on the widget type), the others are way, way off and are a bug in the code in this PR.

There should only be 1 line where the baseline of the labels and the baseline of the first form element align. A baseline is the bottom part of the glyphs (letter shapes) of type (text), not including descenders (the parts that go down like in the letters: pgjq and even ones like e, s, and o a little etc.) That is, if you line up something with the bottom of letters like znmx (at least in most fonts), that should be the baseline.

Figma has a good article on baselines @ https://www.figma.com/dictionary/baseline/

For vertical alignment, free-form text should always be aligned to the baseline. For things encapsulated like buttons and inputs, it should be either baseline or vertically centered, depending on context (including the other types of widgets).

In this case, it's a bunch of freeform text in a form. This should all be aligned to the baseline.

garrett:
Here's what it should look like:

baseline-fixed

garrett:
Some of the text is a bit weird in this dialog. Perhaps it should be closer to something along these lines?

### Storage
- Duplicate storage on new host
- Use shared storage

Does it matter that shared storage is set up or not? What impact does it have here?

### Persistence
- Remove virtual machine from this host

skobyda:
> Some of the text is a bit weird in this dialog. Perhaps it should be closer to something along these lines?
> ### Storage
> 
>     * Copy storage to new host
> 
>     * Use shared storage
> 
> 
> Does it matter that shared storage is set up or not? What impact does it have here? Do we know if it's set up or not?

Shared storage in many cases is preferred option as it's safer. In some cases libvirt might refuse to to migrate without shared storage because disks are not on cache coherent filesystem.
We don't know, but somehow libvirt can figure out if shared storage is set up. I can look how they do it and can also figure it out in cockpit code too.

> ### Persistence
> 
>     * Remove virtual machine from this host

skobyda:
Oh I thought the problem was just with the "Delete source VM" being way off. This seems now closer, altough the radio buttons and label still do not seem perfectly alligned, but I don't know if that's still a PR bug or a PF bug
Screenshot from 2021-03-16 13-34-21

garrett:
That's a lot better! Thanks!

> This seems now closer, altough the radio buttons and label still do not seem perfectly alligned, but I don't know if that's still a PR bug or a PF bug

It's a PF bug. Their labels are off 1 to 2 px. I'm guessing it's based on rem, which is the root font size, but it gets rounded to 1 or 2 pixels depending on ???.

Here's what it looks like now, with the baselines illustrated (from this zoomed in crop of a screenshot of GIMP):

Screenshot from 2021-03-16 13-56-09


So your code is correct.

The most we can do is fix it in our override file and share our changes with PF. I'm 98% certain we have an issue open for this. Some projects only use top labelling. Others use side too. But I think we're the only one using side labelling and have a designer (now just me), so we're probably the only ones who are annoyed about the labels being off by a pixel or two?

garrett:
Screenshots need to be updated in the top post. They're out of date.

Should we use something based on the one at https://github.com/cockpit-project/cockpit/pull/15407#issuecomment-800224029 instead? Or take new ones?

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+
+import { migrateToUri } from '../../libvirt-dbus.js';
+import { isEmpty, isObjectEmpty } from '../../helpers.js';
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+
+const _ = cockpit.gettext;
+
+const DestUriRow = ({ validationFailed, destUri, setDestUri }) => {
+    return (
+        <FormGroup label={_("Destination URI")} fieldId="dest-uri"
```
KKoukiou, line 41: The fieldId should match the id of the component which is child.

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+                       placeholder={cockpit.format(_("Example, $0"), "qemu+ssh://192.0.2.16/system")}
+                       onChange={setDestUri} />
+        </FormGroup>
+    );
+};
+
+const TypeRow = ({ type, setType, setUndefineSource, setStorage }) => {
+    return (
+        <FormGroup label={_("Migrated VM state")}
+                   fieldId="type"
```
KKoukiou, line 56: Same, the fieldId should match the id. 'type'. And it would be ideal if you could choose some more specific id. Since ids should be unique, type has high changes being reused on the same dom somewhere.

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+
+const UndefineSourceRow = ({ undefineSource, setUndefineSource }) => {
+    return (
+        <FormGroup label={_("Delete source VM")}
+                   fieldId="undefine-source"
+                   id="undefine-source"
+                   hasNoPaddingTop>
+            <Checkbox id="undefine-source"
+                      isChecked={undefineSource}
+                      label={_("Remove virtual machine from this host")}
```
KKoukiou, line 84: @garrett does this need an article? aka "Remove the virtual machine ..."

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+        </FormGroup>
+    );
+};
+
+const StorageRow = ({ storage, setStorage, type }) => {
+    const copyDisabled = type === "offline";
+    let copyRadio = (
+        <Radio id="copy"
+               name="source"
+               label={_("Copy storage")}
```
KKoukiou, line 95: I think we need to add a explanatory popover here: "I checked all non-shared non-readonly images will be transferred."

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+            <UndefineSourceRow undefineSource={undefineSource}
+                               setUndefineSource={setUndefineSource} />
+        </Form>
+    );
+
+    const footer = (
+        <>
+            {!isObjectEmpty(error) && <ModalError dialogError={error.dialogError} dialogErrorDetail={error.message} />}
+            <Button variant='primary'
+                    isLoading={inProgress}
```
KKoukiou, line 177: We need to either disable the dialog closing while the migration is running or better close it and show a notification that the migration is in progress, and then once more when it's done. Similar approach to the create VM dialog, but we need to communicate it a bit better.

pkg/machines/libvirt-dbus.js
```diff
@@ -165,6 +165,25 @@ const Enum = {
...
+    VIR_MIGRATE_NON_SHARED_DISK: 64,
+    VIR_MIGRATE_NON_SHARED_INC: 128,
+    VIR_MIGRATE_CHANGE_PROTECTION: 256,
+    VIR_MIGRATE_UNSAFE: 512,
+    VIR_MIGRATE_OFFLINE: 1024,
+    VIR_MIGRATE_COMPRESSED: 2048,
+    VIR_MIGRATE_ABORT_ON_ERROR: 4096,
+    VIR_MIGRATE_AUTO_CONVERGE: 8192,
+    VIR_MIGRATE_RDMA_PIN_ALL: 16384,
+    VIR_MIGRATE_POSTCOPY: 32768,
```
KKoukiou, line 184: Can you remove all the flags that you don't use?

pkg/machines/libvirt-dbus.js
```diff
@@ -1522,6 +1541,28 @@ function initResource(connectionName, method, updateOrAddMethod, flags) {
             .catch(ex => console.warn('initResource action failed:', ex.toString()));
 }
 
+export function migrateToUri(connectionName, objPath, destUri, storage, type, undefineSource) {
+    // direct migration is not supported by QEMU, so it's opposite, the P2P migration should always be used
```
KKoukiou, line 1545: What do you mean, 'so it's opposite'?

test/verify/check-machines-lifecycle
```diff
@@ -98,30 +102,39 @@ class TestMachinesLifecycle(VirtualMachinesCase):
...
-        self.performAction("subVmTest1", "reboot")
+        b.click("#vm-subVmTest1-action-kebab button")
+        b.click("#vm-subVmTest1-reboot")
         wait(lambda: "reboot: Power down" in self.machine.execute("cat {0}".format(args["logfile"])), delay=3)
         b.wait_in_text("#vm-subVmTest1-state", "Running")
 
         # force reboot
         self.machine.execute("echo '' > {0}".format(args["logfile"]))
-        self.performAction("subVmTest1", "forceReboot")
+        b.click("#vm-subVmTest1-action-kebab button")
```
KKoukiou, line 129: This is clearly a rebase error :) And all the above test code, reverted a reverted commit.

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+        <FormGroup label={_("Migrated VM state")}
+                   fieldId="type"
+                   id="type"
+                   hasNoPaddingTop>
+            <Radio id="live"
+                   name="type"
+                   label={_("Live")}
+                   isChecked={type === "live"}
+                   onChange={() => setType("live")} />
+            <Radio id="offline"
```
KKoukiou, line 64: This has the following documentation on the libvirt.org

'Migrate a domain definition without starting the domain on the destination and without stopping it on the source host.'

If the source VM will not be stopped this needs to be communicated. You can just add a tooltip to both options explaining them.

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+
+const DestUriRow = ({ validationFailed, destUri, setDestUri }) => {
+    return (
+        <FormGroup label={_("Destination URI")} fieldId="dest-uri"
+                   id="dest-uri"
+                   validated={validationFailed.destUri ? "error" : "default"}>
+            <TextInput id='dest-uri-input'
+                       validated={validationFailed.destUri ? "error" : "default"}
+                       value={destUri}
+                       placeholder={cockpit.format(_("Example, $0"), "qemu+ssh://192.0.2.16/system")}
```
KKoukiou, line 47: @garrett do we need the 'Example, ' part here?

KKoukiou (changes requested):
It's in a very good state already, my biggest concern is that the migration dialog should not be blocking, since it might just take long. We should close it and with Toast notification notify the user when it's done.

pkg/machines/components/vm/vmMigrateDialog.jsx
```diff
@@ -0,0 +1,194 @@
...
+                       onChange={setDestUri} />
+        </FormGroup>
+    );
+};
+
+const TypeRow = ({ type, setType, setUndefineSource, setStorage }) => {
+    return (
+        <FormGroup label={_("Migrated VM state")}
```
KKoukiou, line 55: @garrett should we maybe be consistent and use inline radio groups as we do in most of the other dialogs?

KKoukiou (changes requested):

garrett:
> Shared storage in many cases is preferred option as it's safer. 

Then it should be the default, when available. And, by that logic, also first (also when available).

> In some cases libvirt might refuse to to migrate without shared storage because disks are not on cache coherent filesystem.

Is there a way to find out this issue before the migration attempt? Perhaps even before this dialog is shown?

> We don't know, but somehow libvirt can figure out if shared storage is set up. I can look how they do it and can also figure it out in cockpit code too.

It would be neat if there's a way to query libvirt about this, instead of having to maintain our own effectively duplicate code. (Unless it's trivial and won't change, then I guess it's possibly OK.)

> Alternatively there can be also just a checkbox with "Copy storage to new host", where if the checkbox is left unchecked, it will be assumed that shared storage is set up.

Copying and using shared storage are not opposites, however.

> But either storage must be coppied to a new host or shared storage has to be set up, otherwise migrated VM would not work on the other end.

I was thinking of only showing the applicable options at first, but if shared storage is preferred and safer, it should always be an option, even when it's not possible. In the cases where shared storage is not available, it would be good to message why and provide a hint on how to fix it.

While I think it would be a good idea to get this in as soon as possible, perhaps it could be a follow-up PR (ideally one that would land in the same release), unless you think it's suitable for this PR too (and then I'd defer to you).

pkg/machines/components/vm/vmMigrateDialog.jsx
garrett: **Quick summary**: UI text is often headline usage, where articles are dropped for simplicity.

This is sometimes pointed out in style guides, but often only implied.

---

UI text is often in classic* headline style of writing. Sometimes, this has included headline capitalization as well, but even in sentence capitalization, the text is still written like a headline.

_\* Of course wouldn't use the _clickbait_ headline style in a UI. :smirk: ("You wouldn't believe what this option does", "Six great tips for customizing virtual machines", "One sysadmin enabled shared storage and her story will astound you")_

In headline rules, you can — and should — drop additional words when it is clear what is being talked about. This conserves space and allows for more efficient scanning of the text.

Here's a good article about UI Copy (text in UIs) from Nielsen Norman Group (well-respected UI/UX designers who have written many books and articles):
https://www.nngroup.com/articles/ui-copy/

Relevant quote with emphasis:

> Keep command text short without sacrificing clarity. One-word, self-explanatory command text is easiest to scan, but obscure commands or those used infrequently will require more context. Use just enough text to accurately describe the command and include no more than 2–4 words.  Avoid overexplaining commands and **remove articles (“a”, “an”, and “the”) to speed up scannability, comprehension, and task completion**. While short commands are preferred, use enough text to explain the command sufficiently. 

For what it's worth, because we use headline style of text in UI, we often do drop punctuation, most commonly periods, at the end of a sentence... and often use phrases instead of sentences too.

For an example of a style guide that includes a suggestion on using this simplified type of text, please look at Google's Material design writing guidelines:
https://www.material.io/design/communication/writing.html

Most other places seem to imply this generally accepted UI text convention, including PatternFly. However, PatternFly does imply dropping unnecessary text and going for headline style text in many of its examples, such as [the writing error messages page](https://www.patternfly.org/v4/ux-writing/error-messages). (And [they do suggest simple text](https://www.patternfly.org/v4/ux-writing/ux-writing-best-practices#be-clear-and-concise)... however, they do not define exactly how it should be simple.)

Additionally, there's the concept of "[zero article](https://www.englishclub.com/grammar/determiners-zero-article.htm)", when articles such as "the" or "a" are often dropped from usage in English. It varies on UK vs. US English. However, headline rules and UI text go even beyond zero article or generic usage.

garrett: FWIW: I looked this up, as I wanted to make 100% sure what I was suggesting was fine and also to share with others here.

You're more correct for English grammar. Normally, you would need an article here, but headline and UI text conventions "throw that out the window" (get rid of it).

pkg/machines/components/vm/vmMigrateDialog.jsx
garrett: PatternFly recommends example:
https://www.patternfly.org/v4/components/form/design-guidelines/#user-help

> For accessibility reasons, you should always **precede placeholder text that is in example form with "Example," followed by a space** like "Example, (555) 555-5555" for a phone number example or "Example, 491" for a security code example. This will ensure that users with screen readers will know the difference between the form label and the placeholder text. Placeholder text that does not use an actual example, like “MM-DD-YYYY,” does not need to be preceded with “Example.”

pkg/machines/components/vm/vmMigrateDialog.jsx
garrett: Inline radios in the other forms are at the top and switch between different modes.

As this would be a toggle state of on/off, we could rework it to be a checkbox, however. Perhaps something like this:

☑ Run virtual machine after migration

After looking at the dialog again and thinking about it further, I think the order is wrong. Persistent stuff like storage should come first. Stateful things should come afterward.

So, this order, in other words:

1. Location URI
2. Storage
3. Delete original (also storage-related)
4. Run after migration

KKoukiou: virt-manager has the option * [ ] temporary move, which is a miz of our online/offline and undefine options. I have no preference over our way or what virt-manager exposed. I just thought that what they do, it's pretty nice and compact and good enough for first iteration. In addition, users are already used to that tool, so having similar dialog fields if they make sense is to the users' advantage I feel .

garrett:
I've grabbed a bunch of recommendations above and quickly whipped up a mockup in Excalidraw that shows:

1. Reordering
2. Grouping storage options together
3. Rewording
4. Changing state to a checkbox

machine-migration

Additionally, I made a variant that shows what we can do when there's no shared storage:

machine-migration-no_shared

I don't know what the hint text would be. It depends on how to solve the problem.

This is a suggestion based on the comment above that shared storage is preferable and safer, so it should be used whenever possible, versus copying, which is a fallback.

Would removing the original virtual machine only be an option for copying? Or also for shared storage? Does it delete original volumes or just remove the virtual machine definition?

If it _deletes_ the original storage volume when copying (for example), it should reflect that in the text. So it could change based on the radio button selection.

It's wordy, but something to this effect, if so:
"Remove original virtual machine and delete storage after copying"

KKoukiou:
> I've grabbed a bunch of recommendations above and quickly whipped up a mockup in Excalidraw that shows:
> 
>     1. Reordering
> 
>     2. Grouping storage options together
> 
>     3. Rewording
> 
>     4. Changing state to a checkbox
> 
> 
> machine-migration
> 
> Additionally, I made a variant that shows what we can do when there's no shared storage:
> 
> machine-migration-no_shared
> 
> I don't know what the hint text would be. It depends on how to solve the problem.
> 
> This is a suggestion based on the comment above that shared storage is preferable and safer, so it should be used whenever possible, versus copying, which is a fallback.
> 
> Would removing the original virtual machine only be an option for copying? Or also for shared storage? Does it delete original volumes or just remove the virtual machine definition?
> 
> If it _deletes_ the original storage volume when copying (for example), it should reflect that in the text. So it could change based on the radio button selection.
> 
> It's wordy, but something to this effect, if so:
> "Remove original virtual machine and delete storage after copying"

Why the remove original virtual machine is under storage label? If it's not shouldn't it have a label?

garrett:
> Why the remove original virtual machine is under storage label? If it's not shouldn't it have a label?

It's under storage, as it relates to storage (on the original machine), right?

KKoukiou:
> > Why the remove original virtual machine is under storage label? If it's not shouldn't it have a label?
> 
> It's under storage, as it relates to storage (on the original machine), right?

Not really. Remove the original machine refers to the VM definition. See man virsh, search for `--undefinesource undefines the domain on the source host`

garrett:
>Remove the original machine refers to the VM definition.

We could add a label in this case. The order still happens to be the same in this small change (location, storage, cleanup, action):

machine-migration(1)

What happens if someone selects copy and removes the original virtual machine? Is the storage on the old one cleaned up, or just a large file sitting around for no reason?

skobyda:
> After talking on IRC, @KKoukiou mentioned the temporary move checkbox in virt-manager.
> 
> Putting the emphasis on the migration being temporary (instead of removing the original versus keeping it) could look something like this:
> 
> machine-migration(2)

I would change the text "When shut off, virtual machine will return to the original host". With temporary migration, the original host will never leave the source host. It will just shut off on the source, and a transient running version of it will be created at the destination host. After the transient version is shut off at the destination host, it will cease to exist there and users still has to start up the original VM at the source manually.

garrett:
Oh, so if it's temporarily moved, then it would have to be manually started back up too?

When/how would someone use this?

garrett:
I guess we could note it with something like this?

machine-migration(3)

"When shut off on the destination, the virtual machine will return to the original host in a shut off state."

skobyda:
> Oh, so if it's temporarily moved, then it would have to be manually started back up too?
> 
> When/how would someone use this?

I found an sysadmin article that mentions that this option is useful for `hardware maintenance or emergency load-balancing` . So I guess in cases when you need to reboot the source host but want to minimize the downtime of the VM runinng on it, so you just move it temporarily.

garrett:
> useful for hardware maintenance or emergency load-balancing

Perhaps we should mention this in the tooltip?

> I guess in cases when you need to reboot the source host but want to minimize the downtime of the VM runinng on it, so you just move it temporarily.

Let's walk through the scenario:

1. Oh no! I just realized I need to reboot this machine!
2. Migrate important VM from the host to another
3. Reboot
4. Machine is back up, but the VM is running elsewhere... now what?

Does the origin host know the VM was temporary lent to another machine? Is there a way to request that it come back? Can Cockpit even show the state as migrated?

So manual intervention could get it back: Shut it down on the remote host. Start it up on the original host. But can this be automated somehow?

skobyda:
> > useful for hardware maintenance or emergency load-balancing
> 
> Perhaps we should mention this in the tooltip?
> 
> Here's what that could look like:
> 
> machine-migration(4)
> 
> > I guess in cases when you need to reboot the source host but want to minimize the downtime of the VM runinng on it, so you just move it temporarily.
> 
> Let's walk through the scenario:
> 
>     1. Oh no! I just realized I need to reboot this machine!
> 
>     2. Migrate important VM from the host to another
> 
>     3. Reboot
> 
>     4. Machine is back up, but the VM is running elsewhere... now what?

You can shut off the VM elsewhere and start it at the source to continue its function. If the VM has a shared storage set up, it's very smooth process where the source VM has all the data changes the destination might have done and can continue with the function the temporary VM was doing. If there is a shared storage set up, this might be more complicated (I will write libvirt guys to see if "Copy data" and "temporary move" together even make sense)

KKoukiou:
> > useful for hardware maintenance or emergency load-balancing
> 
> Perhaps we should mention this in the tooltip?
> 

I believe no. Once a user ends up in the VM migration dialog they really will know if they want to keep the VM running on the host or destination. It's basically the main reason they ended up there.

It's not an additional option that might be unclear.  In disks dialog, disk cache for example is not main option, it might affect performance somehow however. Therefore needs to be explained. The state of VM in a migration process it's really the main thing.

skobyda:
> > useful for hardware maintenance or emergency load-balancing
> 
> Perhaps we should mention this in the tooltip?
> 
> Here's what that could look like:
> 
> machine-migration(4)
> 
> > I guess in cases when you need to reboot the source host but want to minimize the downtime of the VM runinng on it, so you just move it temporarily.
> 
> Let's walk through the scenario:
> 
>     1. Oh no! I just realized I need to reboot this machine!
> 
>     2. Migrate important VM from the host to another
> 
>     3. Reboot
> 
>     4. Machine is back up, but the VM is running elsewhere... now what?
> 
> 
> Does the origin host know the VM was temporary lent to another machine? Is there a way to request that it come back? Can Cockpit even show the state as migrated?

I don't think there is a way for us to find out if there are migrated versions of some VM out there somewhere.

> So manual intervention could get it back: Shut it down on the remote host. Start it up on the original host. But can this be automated somehow?

In case of shared storage, this would consist of just shutting down the VM at the destination and starting it locally. You can do the remote connection thru virsh to connect to some remote qemu instance (by runinning `virsh -c [destination_host]`". So theoretically we could remember where the VM was migrated, and then just spawn "virsh -c [destination_host] destroy [vm_name]" followed by running "virsh start [vm_name]" locally. Altough I don't know how to put this concept into cockpit as we don't have remote management of other VM's yet in cockpit, altough libvirt supports it.

garrett:
So what happens if you migrate temporarily and start the VM again on the origin host while it's still running on the remote host? If Cockpit isn't aware of it, then this __will__ happen. What happens when it does?

KKoukiou:
> So what happens if you migrate temporarily and start the VM again on the origin host while it's still running on the remote host? If Cockpit isn't aware of it, then this **will** happen. What happens when it does?

Why this **will** happen?
Migrating VMs is not something do for fun just rendomly restarting VMs, people who will either know what they are doing or if they just play around they probably aware they might shoot themselves on the leg. So, I don't feel that we should have warnings protecting them from doing something wrong.

As Simon says, we can't know if a signel disk is used by VMs on other hosts, so I think there is nothing more to do there.

garrett:
There are tons of reasons why someone might inadvertently start a VM on a machine that's already been temporarily migrated:
- More than 1 person admins a machine.
- The VM might be a long running process. 
- An admin may need to tend to something else while the VM is running and come back and forget that they've migrated a VM.
- There might be similar VMs where some are migrated and some are not (but they're stopped).

I'm saying that it will definitely happen to someone for one of the above reasons (or countless others) if Cockpit doesn't show someone the migration status of a VM.

skobyda:
> So what happens if you migrate temporarily and start the VM again on the origin host while it's still running on the remote host? If Cockpit isn't aware of it, then this **will** happen. What happens when it does?

In case of shared storage, you can end up with 2 identical VMs using same data resources and can lead to data corruption.
I believe there can also networking conflicts.
In summary, you don't want both the original VM and the migrated VM running at the same time.
But I don't think we can also safely check it. I do not seem to find after googling if there is a way to see if any other VM is using the same shared storage (plus it would have to be checked that they are both the same VM).
The best solution I can think of is to warn user about this situation somehow.

garrett:
> In case of shared storage, you can end up with 2 identical VMs using same data resources and can lead to data corruption.
I believe there can also be networking conflicts.

Yikes. I thought shared storage was supposed to be "safer"? 😱

>  The best solution I can think of is to warn user about this situation somehow.

There's no kind of [lock](https://en.wikipedia.org/wiki/Lock_(computer_science)) to prevent this? Or some way to query to see if the machine is running somewhere, somehow? It seems like a rather huge oversight if so.

Unless there's some metadata that the VM is running elsewhere, the only place where we could warn someone is in the migration dialog. And this should actually be a warning, not an info, as there could be huge data loss and wonky networking.

Ugh, I'm not happy with introducing a warning on the dialog, but, unless there's some way to know about the problem, we kind of don't have a choice. 

---

The possibility of unseen, accidental data loss with no way to know makes this a critical problem. I guess if there really isn't a way to discover this information about the VM running elsewhere (even if it is running _somewhere_ else at all — not even including metadata about the host or anything else) then I guess we need to file an upstream bug for this to be added.

And we should be able to determine if querying is even possible, so we could not have a warning here, but have status information on the VM and a prevention of running the machine a second time. (If it's even possible, there should be a confirmation.)

---

Since this would only happen with a temporary migration and with a read/write volume, we should probably have the warning appear at the footer of the dialog in the form of an inline warning-class alert, but only when temporary migration is enabled.

It could be something like this:

⚠ Do not run multiple instances of this virtual machine with , as this will lead to data loss.

In other words: If it's a temporary move _and_ there's a read/write volume: Show the above warning.

skobyda:
> > In case of shared storage, you can end up with 2 identical VMs using same data resources and can lead to data corruption.
> > I believe there can also be networking conflicts.
> 
> Yikes. I thought shared storage was supposed to be "safer"? scream
> 
> > The best solution I can think of is to warn user about this situation somehow.
> 
> There's no kind of [lock](https://en.wikipedia.org/wiki/Lock_(computer_science)) to prevent this? Or some way to query to see if the machine is running somewhere, somehow? It seems like a rather huge oversight if so.
> 
> Unless there's some metadata that the VM is running elsewhere, the only place where we could warn someone is in the migration dialog. And this should actually be a warning, not an info, as there could be huge data loss and wonky networking.
> 
> Ugh, I'm not happy with introducing a warning on the dialog, but, unless there's some way to know about the problem, we kind of don't have a choice.
> 
> The possibility of unseen, accidental data loss with no way to know makes this a critical problem. I guess if there really isn't a way to discover this information about the VM running elsewhere (even if it is running _somewhere_ else at all — not even including metadata about the host or anything else) then I guess we need to file an upstream bug for this to be added.
> 
> And we should be able to determine if querying is even possible, so we could not have a warning here, but have status information on the VM and a prevention of running the machine a second time. (If it's even possible, there should be a confirmation.)
> 
> Since this would only happen with a temporary migration and with a read/write volume, we should probably have the warning appear at the footer of the dialog in the form of an inline warning-class alert, but only when temporary migration is enabled.
> 
> It could be something like this:
> 
> warning Do not run multiple instances of this virtual machine with , as this will lead to data loss.
> 
> In other words: If it's a temporary move _and_ there's a read/write volume: Show the above warning.

Yes, the concepot of lock exists for this purpose and in some cases it's used. What I described is the worst case scenario.
If user has a shared storage with a qcow disk, then there in no chance of data corruption, as QEMU will have a lock on the disk and it will not allow the second machine to start up if the other (i.e. migrated machine) machine is using the disk, so this case is covered.
However there are scenarios, like with iscsi block devices, there is no lock and there is a possibility of data corruption.
So since there is no metadata that keeps track about migrated machines, nor I think there is a way to see if 2 VMs which are accessing iscsi block devices are copies (migrated) version of each other, I think the safest solution would be:
To show the warning which you mentioned only in the case that user chooses Shared storage, they choose Temporary move (therefore the souce VM won't be undefined) and we find out from VM xml that it uses iscsi block device (or other disk device which doesn't have a lock).

KKoukiou:
> I agree. I would make the case even more specific: To show the warning which you mentioned only in the case that user chooses Shared storage, they choose Temporary move (therefore the souce VM won't be undefined) and there is aiscsi block device (or other disk device which doesn't have a lock).

I think we are overanalyzing this here. virt-manager does not show any warning and everyone is happy with that product. Most people use qcow2 disks anyway. I am not sure if we really want to add so much logic in cockpit, at least for the first iteration this is really not necessary I feel.

skobyda:
Okay, so far I updated the PR with what we agreed on.
Adding warning about dangers is still open to discussion. But in most cases, there is no big danger, as with "Copy Storage

I would personally put the "Run after migration" checkbox above "Temporary move". And make the "Temporary move" show only conditinally when "Run after migration" is checked. It doesn't make sense for combination of "Temporary move" but no "run after migration" to happen, as during temporary move the machine at the destination is transient, so it has to be running.
So rendering that checkbox will make prevent that combination from happening. WDTY @garrett ?
Screenshot from 2021-03-24 13-23-54
Screenshot from 2021-03-24 13-24-01

Screenshot from 2021-03-24 13-25-25
Also I'm still not sure about the sentence `virtual machine will return to the original host`. I think it could make user think that after VM is shutoff, another, reverse migration might happen where all the configuration changes which user could have done to the migrated VM will be migrated back. That would be no true. The machine at the original host will be the same as before the migration.
So I would perhaps rephrase it to "When shut off on the destination, migrated machine will cease to exist and the machine on the original host can safely resume."?

garrett:
> I am not sure if we really want to add so much logic in cockpit

This is why I suggested that upstream (libvirt in this case) should make sure there's a way to identify when a machine is temporarily checked out, if they haven't done this already.

And we should somehow show this, ideally with upstream support.

If it's not solved somehow, then people will blame Cockpit for corrupting their VM and losing their data. That's what I want for us to avoid.

If qcow2 shared storage volumes indeed support being locked, that's a way to prevent launching a second VM regardless of location. And something we could show in Cockpit. Something like this is exactly what I was talking about.

> But in most cases, there is no big danger, as with "Copy Storage" no data corruption can happen

Right, which is why I said the warning should only show up with shared storage with vulnerable volumes, such as read/write volumes that aren't locked.

> put the Run after migration checkbox above Temporary move

Is "Run after migration" checked by default? It's probably what most people would want if they're migrating a server. And then unchecking it would remove (or perhaps disable) the item instead of make another checkbox magically appear.

>  Also I'm still not sure about the sentence virtual machine will return to the original host. I think it could make user think that after VM is shutoff, another, reverse migration might happen where all the configuration changes which user could have done to the migrated VM will be migrated back. That would be no true. The machine at the original host will be the same as before the migration.
>
> So I would perhaps rephrase it to "When shut off on the destination, migrated machine will cease to exist and the machine on the original host can safely resume."?

Oh, I thought the shared storage would have the changes written back to it. Does it not then? What's the point of shared storage? Is it to just speed up the process?

And if the remote server isn't going to write into the VM, then is there harm in spinning up one on the original host at the same time? Or would that cause issues in the remote one as the contents on the disk might have changed from under it?

---

Overall, the changes are looking good. We just have a few more things to agree upon and address.

skobyda:
Moving this PR and the discussion here https://github.com/cockpit-project/cockpit-machines/pull/15

garrett:
Oh, why is copy data disabled? Shouldn't run be disabled instead? Perhaps with a notice why?

PR #14684, opened 2020-09-29
Add OpenSUSE Tumbleweed favicon

ledlamp:
Just installed cockpit on a tumbleweed system and was sad to see it had the default favicon so I found `/usr/share/pixmaps/distribution-logos/square-hicolor.svg` and converted it to a 48x48 favicon.ico.

PR #15408, opened 2021-02-23
branding: Remove Debian favicon

martinpitt:
This file is not actually installed. Instead, the debs contain a symlink
to the generic /usr/share/pixmaps/debian-logo.png, which works just
fine.
    
Related to https://bugs.debian.org/983218


martinpitt:
This icon is now 9 kB instead of the previous 1. I asked Philip in the Debian bug whether he has a better one, so let's keep this open for  a day or two.

marusak:
No chance that debian would ship favicon and we would just symlink it as we do anywhere else?

martinpitt:
@marusak: We do link the logo, but the browser favicon.ico must be in that spethial format. We have a version for default and ubuntu as well.

marusak:
I know, but the same way as openSuse managed to add it into distro: https://github.com/openSUSE/distribution-logos/issues/1 and then we could use it: https://github.com/cockpit-project/cockpit/pull/14684

If we could not ask debian developers to include favicon.

martinpitt:
I grepped my entire Debian server for *.ico, and right now there is none I'm afraid. Just /usr/share/pixmaps/debian-logo.png (which we already use in our branding as logo.png symlink).

However, I just realized that this is bogus -- our *source code* contains an icon, but our *binary packages* contain just a symlink:

    lrwxrwxrwx 1 root root 32 Feb  4 14:48 /usr/share/cockpit/branding/debian/favicon.ico -> ../../../pixmaps/debian-logo.png

That's what the Makefile says, too. So we don't need this at all, and I'll fix this instead to just remove the old icon from our source.

marusak (approved):
Ha, nice!

PR #15411, opened 2021-02-23
metrics: Going over 80% does not mean necessary event

marusak:
If the processor (for example) is loaded to around 80% all the time and just
fluxes one or two % up and down it is interpreted as event every minute.

Lets consider the rise over 80% to be at least somewhat significant.

I was working on this page while my CPU was busy and noticed I had waaay to many events, even though the graphs looked pretty stable. Loaded machine !== something bad is happening. Better ideas are welcomed. I was thinking about taking average from all points and then checking if the 80% mark is not close to average or so, but I though to start with something simple and see what others think.

Before:
justspikesandnothingelse

After:
werejustspikes

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -598,7 +598,7 @@ const MetricsHour = ({ startTime, data, clipLeading }) => {
                 return;
             const value = samples[type];
             // either high enough slope, or crossing the 80% threshold
-            if (prev_val !== null && (value - prev_val > 0.25 || (prev_val < 0.8 && value >= 0.8))) {
+            if (prev_val !== null && (value - prev_val > 0.25 || (prev_val < 0.75 && value >= 0.8))) {
                 const minute = Math.floor(i / SAMPLES_PER_MIN);
                 if (minute_events[minute] === undefined)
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
This certainly helps a lot already for this "around 80%" noise case, so fine to land. However, I think this will eventually need a cleaner approach. (1) it should be stateful: not repeat events for resource X after you already got one for X), and (2) have a hysteresis: clear the "just got event for X" state only after X usage drops significantly again.

E.g. You get one event after the CPU rises above 80%, and then won't get any further ones until the CPU dips below 40% once.

martinpitt (changes requested):
Whoops, breaks TestHistoryMetrics.testEvents

marusak:
> (1) it should be stateful: not repeat events for resource X after you already got one for X), and (2) have a hysteresis: clear the "just got event for X" state only after X usage drops significantly again.

Agree! That should be of course carried over minutes and hours so it needs some good thinking about it when we are implementing it.

martinpitt (approved):

marusak:
lifecycle is busted. This is like 5th retry. Once mroe I guess?

PR #15412, opened 2021-02-23
Clean up variable expansion in the router

allisonkarlitskaya:
This replaces the custom code for managing the argv/envp and variable expansions inside of them with a generic transform performed at the JSON level.  This already makes the code much easier to understand (since everything we do is eventually turned back into a JsonObject again anyway) and sets the stage for adding a second level of variable expansion without things getting wildly out of hand.

### DIFF
--- a/src/common/cockpitjson.c
+++ b/src/common/cockpitjson.c
@@ -967,3 +967,165 @@ cockpit_json_to_hash_table (JsonObject *object,
 
   return hash_table;
 }
+
+typedef const struct {
+  CockpitJsonWalkCallback callback;
+  gpointer                user_data;
+} WalkClosure;
+
+static JsonNode *
+cockpit_json_walk_node (JsonNode    *node,
+                        WalkClosure *closure);
+
+static JsonArray *
+cockpit_json_walk_array (JsonArray   *array,
+                         WalkClosure *closure)
+{
+  gint n = json_array_get_length (array);
+  JsonArray *copy = NULL;
+
+  for (gint i = 0; i < n; i++)
+    {
+      JsonNode *element = json_array_get_element (array, i);
+      JsonNode *node = cockpit_json_walk_node (element, closure);
+
+      if (node && !copy)
+        {
+          /* We've decided we need to copy now, so revisit everything up
+           * to this point and add it to the copy.
+           */
+          copy = json_array_sized_new (n);
+          for (gint j = 0; j < i; j++)
+            json_array_add_element (copy, json_array_dup_element (array, j));
+        }
+
+      if (copy)
+        json_array_add_element (copy, node ?: json_node_ref (element));
+    }
+
+  return copy;
+}
+
+static JsonObject *
+cockpit_json_walk_object (JsonObject  *object,
+                          WalkClosure *closure)
+{
+  JsonObject *copy = NULL;
+  JsonObjectIter iter;
+  const gchar *key;
+  JsonNode *value;
+
+  json_object_iter_init (&iter, object);
+  while (json_object_iter_next (&iter, &key, &value))
+    {
+      JsonNode *node = cockpit_json_walk_node (value, closure);
+
+      if (node && !copy)
+        {
+          const gchar *this_key = key;
+
+          /* We've decided we need to copy now, so revisit everything up
+           * to this point and add it to the copy.
+           */
+          copy = json_object_new ();
+          json_object_iter_init (&iter, object);
+          while (json_object_iter_next (&iter, &key, &value) && key != this_key)
+            json_object_set_member (copy, key, json_node_ref (value));
+
+          /* ...now key and value will be where they were before we entered */
+          g_assert (key == this_key);
+        }
+
+      if (copy)
+        json_object_set_member (copy, key, node ?: json_node_ref (value));
+    }
+
+  return copy;
+}
+
+static JsonNode *
+cockpit_json_walk_node (JsonNode    *node,
+                        WalkClosure *closure)
+{
+  JsonNode *new_node = NULL;
+
+  switch (json_node_get_node_type (node))
+    {
+    case JSON_NODE_ARRAY:
+      {
+        JsonArray *array = json_node_get_array (node);
+        JsonArray *copy = cockpit_json_walk_array (array, closure);
+
+        if (copy)
+          {
+            new_node = json_node_new (JSON_NODE_ARRAY);
+            json_node_take_array (new_node, copy ?: json_array_ref (array));
+          }
+
+        break;
+      }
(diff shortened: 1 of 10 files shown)


src/common/cockpitjson.c
```diff
@@ -967,3 +967,165 @@ cockpit_json_to_hash_table (JsonObject *object,
...
+        if (copy)
+          {
+            new_node = json_node_new (JSON_NODE_OBJECT);
+            json_node_take_object (new_node, copy ?: json_object_ref (object));
+          }
+
+        break;
+      }
+
+    default:
```
martinpitt, line 1082: Please make this `JSON_NODE_VALUE`; explicitly handle `JSON_NODE_NULL`, and make the  `default:` a `g_assert_not_reached()`. Please add a unit test for `null` values.

src/common/cockpitjson.c
```diff
@@ -967,3 +967,165 @@ cockpit_json_to_hash_table (JsonObject *object,
...
+        new_node = (* closure->callback) (node, closure->user_data);
+        break;
+      }
+    }
+
+  return new_node;
+}
+
+/**
+ * cockpit_json_walk:
```
martinpitt, line 1093: This API seems rather nice, and also very generic -- do you plan to submit that to json-glib?

mvollmer (approved):
Beautiful!

src/common/test-json.c
```diff
@@ -676,6 +676,155 @@ test_write_infinite_nan (void)
...
+
+  g_autoptr(JsonObject) source = cockpit_json_parse_object (patch_data, -1, &error);
+  g_assert_no_error (error);
+  json_object_seal (source);
+
+  /* for the nul case, we should get exactly the same object back */
+  {
+    int counter = 0;
+    g_autoptr(JsonObject) result = cockpit_json_walk (source, null_subst, &counter);
+    g_assert_cmpint (counter, ==, 13);
```
martinpitt, line 736: Meh magic numbers -- but I suppose there's no easy way to get the node count of a json tree, at least I didn't see any in the docs.

src/common/test-json.c
```diff
@@ -676,6 +676,155 @@ test_write_infinite_nan (void)
...
+{
+  int *counter = user_data;
+
+  (*counter)++;
+
+  return NULL;
+}
+
+static void
+test_walk (void)
```
martinpitt, line 724: This looks pretty good to me, and with valgrind this should ensure that the refcounting works. I think the only missing case is to cover JSON `null`s, as above.

martinpitt (changes requested):
This approach looks great to me, especially when you could propose that to json-glib. I have a few nitpicks, but nothing that would change the structure. Thanks!

src/common/cockpitjson.c
```diff
@@ -967,3 +967,165 @@ cockpit_json_to_hash_table (JsonObject *object,
...
+      if (node && !copy)
+        {
+          const gchar *this_key = key;
+
+          /* We've decided we need to copy now, so revisit everything up
+           * to this point and add it to the copy.
+           */
+          copy = json_object_new ();
+          json_object_iter_init (&iter, object);
+          while (json_object_iter_next (&iter, &key, &value) && key != this_key)
```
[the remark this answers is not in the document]
allisonkarlitskaya: I can't imagine that this wouldn't be the case, particularly since this object is immutable.

The underlying object is a hash table, of course, and this iterator is indeed backed by a `GHashTableIter`, which has no guarantees of any kind on order, but won't change if nothing in the table changes.

I have the same concerns for hashing and equality comparisons, but again...

src/common/cockpitjson.c
```diff
@@ -967,3 +967,165 @@ cockpit_json_to_hash_table (JsonObject *object,
...
+            json_object_set_member (copy, key, json_node_ref (value));
+
+          /* ...now key and value will be where they were before we entered */
+          g_assert (key == this_key);
```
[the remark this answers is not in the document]
allisonkarlitskaya: This is not a tautology.  The loop could also exit because we hit the end of the iter, and that's the case I wanted to guard against.

I actually previously kept a counter here (as with the array case) and iterated for that many counts, but I found it made the code hairy and hard to read.... I'd be happy to discuss this with you further, and maybe we can figure out a way to give you better reassurances about this...

src/common/cockpitjson.c
allisonkarlitskaya: It was my intend that `JSON_NODE_NULL` also be handed to the callback.  To be honest, if I could come up with a way to also handle the container cases with the callback, then I would... the limitation that we only call for value types is ... not great.

If we did do a change here, I would actually indeed try to expand the callback to be called for every type, with the following semantics for containers:

 - `NULL` return: do the iteration
 - non-`NULL` return: take this value in place
 - the callback might call back into `_walk()` on the subtree in order to do some combination of the above

src/common/cockpitjson.c
allisonkarlitskaya: I wasn't planning on it, but I could file a PR and see what they say, I guess.  I imagine they'd want an awful lot more testing than we do here...

src/common/test-json.c
allisonkarlitskaya: We do actually test that null is handled properly since the existing testcases for `cockpit_json_patch()` have a `null`, and we reuse that data...

src/common/cockpitjson.c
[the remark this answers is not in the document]
martinpitt: Argh, of course, I mis-read the condition. Thanks for considering, let's not make the code unnecessarily complicated then.

src/common/cockpitjson.c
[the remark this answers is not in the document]
martinpitt: Ack, thanks. Hopefully our tests would catch if this is ever not the case, but I agree it's rather unlikely to not be a reproducible order.

src/common/test-json.c
martinpitt: Was missing that, thanks!

src/common/cockpitjson.c
martinpitt: Right, I agree that the callback should also be called on `NULL`, I would just like this code to be explicit about it, and not do "anything else". It is arguably unlikely that there ever will be more JSON_NODE_* types, but I still find it easier to read if the cases are explicit.

martinpitt: I'm ok with that `default:` cleanup to happen in a follow-up. It's not actually changing functionality at this point.

martinpitt (approved):

PR #15416, opened 2021-02-24
common: parse into immutable JSON trees

allisonkarlitskaya:
We do most of our JSON parsing in cockpit via a per-thread shared parser
object.  It seems that none of the users of the parsing attempt to
modify the results, so modify this object to produce immutable JSON
trees.

### DIFF
--- a/src/common/cockpitjson.c
+++ b/src/common/cockpitjson.c
@@ -488,7 +488,7 @@ cockpit_json_int_equal (gconstpointer v1,
  * @length: length of @data or -1
  * @error: optional location to return an error
  *
- * Parses JSON into a JsonNode.
+ * Parses JSON into an immutable JsonNode.
  *
  * Returns: (transfer full): the parsed node or %NULL
  */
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Hm.  `override.json` is probably going to be an issue.  What's that for?

martinpitt:
@allisonkarlitskaya : Overrides are documented here: https://cockpit-project.org/guide/latest/packages.html#package-manifest . They need to stay working, I'm afraid.

allisonkarlitskaya:
It's a little bit worrying that the tests passed here...

It seems we make extensive use of overrides in the tests, but apparently only for performance reasons... and I guess nothing breaks if it doesn't work...

allisonkarlitskaya:
I'm going to take a different approach here.

martinpitt:
@allisonkarlitskaya : At least `TestStoragePackagesNFS.testNfsMissingPackages` ought to fail if override.json does not work. I don't see how it could pass. If I comment out the override.json creation in the test, it does fail.

PR #15420, opened 2021-02-24
Metrics journal integration

marusak:
Release note:

metrics: Journal integration

For every event on metrics page relevant journal logs are provided to help identify why some event has happened.

logsinmetrics


garrett:
This is super cool.

With some performance fixes to make it act more reasonable, it's good for an MVP. We could add in the tooltip type information on a per-minute average on a follow-up. (Having it over the minute means we might want to consider a different design, with average/min/max, perhaps.)

Preloading for error/warning could happen in another PR, with preloading in general for visible elements after other resources have loaded. Rendering should still happen on demand, even with pre-loading, otherwise the DOM will get super-chonky super-quick when scrolling and such. But having that information pre-cached without it being rendered shouldn't affect performance.

marusak:
So I had some fun evening looking into the performance and here are my find:
- I've created non-dev build. It is surely faster, but not fast enough even on my PC.
- I've of course started with react. I made sure only the one minute line is re-rendered. I even changed it so only the events part is re-rendered. Made no difference. The render was fast, it provides results immediatelly.
- I've looked into using of svgs. If I ever used only compressed minutes it made no difference neither.
- I've then jumped into profiling:

The pink is the whole update from click until it is visibly updated. The blue part is react doing render upon toggle. Those two parts are because at first the panel is opened with `Loading...` and then logs are being shown.
updateofgrid

So the obvious problem is `Layout` especially `Layout shift`. There is a lot of reading about it on the internet. In short it means that once we resize some element in the middle of the page, all elements must be recalculated to find their new position. This is especially bad with grid where the position finding is that more complicated. Also profiling told me that it had to calculate position of 16k elements! Not great. (12 hours, each has 60 minutes where each minute has at least 6 graph items + all the fun part around). I've tried to load just 2 hours and it was indeed faster.

Good quick reading about the problem is here: https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing

I have two ideas how we can proceed since we cannot just optimize the code - the problem is in browser.
1. Don't insert the new panel into graphs but do some kind of pop-over. It would not shift anything around. Illustration:
basepopover
2. Try to use flex (as the article suggest it should be much better). With the new layout I did in this PR elements are ordered correctly. There might be a problem with missing elements for empty data, but that is quick and easy fix. The question is, if we can use flex to make things align so nicely and if it would bring enough speed improvement.

martinpitt:
Honestly, I prefer the popover approach. Especially for resource spikes the log is prone to be very long, and I really doubt that we'll get by without a scrollbar. I wouldn't want to stretch the graph indefinitely, this just looks ridiculous.

marusak:
> Honestly, I prefer the popover approach. Especially for resource spikes the log is prone to be very long, and I really doubt that we'll get by without a scrollbar. I wouldn't want to stretch the graph indefinitely, this just looks ridiculous.

I still think we want to limit how many entries we show. Or don't we? Now I show up to 10. And then want to add just link to 'Go to logs...' as there can be thousands of entries.

martinpitt:
@marusak : Ah, I didn't see the Go to.. link -- that makes sense, yes. 10 isn't much though, but we can tweak this later.

garrett:
Have you used [will-change](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) and [contain](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) yet? They both exist for cases like this.

Browser support is good:

https://caniuse.com/will-change (supported everywhere)
https://caniuse.com/css-containment (not supported in WebKit yet)

garrett:
A popover approach doesn't easily let someone compare what happens between minutes, however, which is one of the goals of this.

garrett:
Additionally, if we look at the layout and ensure the widths of the graphs part are fixed, that should help out layout calculations too.\

Or `will-change: height` might already circumvent that anyway.

marusak:
> A popover approach doesn't easily let someone compare what happens between minutes, however, which is one of the goals of this.

I am not huge fan of popover either, but I would argue that comparing logs is a bit tricky. For that we would need to show all the logs, which easily can be in hundreds in the given minute. Now I show only 10 and want to have "go to all logs" button.

garrett:
At least in this VM, I have spikes with no journal content. I guess we shouldn't show that they're expandable unless there's actually something to show. And I guess that means at least some pre-loading. I suppose we could hide the expander until we know there's content? (But at that time, we'd probably know if there are warnings, etc.)

Screenshot_2021-02-25 admin localhost localdomain

marusak:
> When I close the journal on chrome it makes the block small again but on firefox it does not seem to roll it back up. No idea what is happening there yet. Also seems still slow to me. 

I just pushed commit which changes the one big grid into lines of flexes. It fixes both of these issues.

martinpitt:
I just gave this a quick test. With this [journal.txt](https://github.com/cockpit-project/cockpit/files/6044393/journal.txt), I see no journal entries at all:

image

17:22 was the "real" boot, the stuff before was image-prepare. This presumably needs to be a bit more adaptive.

I'll test/think about this more thoroughly next week, for now I was just curious.

marusak:
I was thinking about this and I think I want to try something like this:
1. We recognize events in 5s intervals. So lets try to limit this from when the first spike has happened until the last spike. (If only one event, then this will be just 5s interval)
2. In this limited range try with different priorities. If there are too many events raise the priority, if too little lower it.
3. If even `debug` cannot provide any data, make the interval bigger (probably +-5s - rounder to the whole minute)
4. Go to step 2 unless the previous interval == the current interval

I'll give this a shot.

marusak:
Ok, so I've implemented just that and now there are no empty minutes (or really hard to find such minute) while the logs being shown are as close to the spike as possible.

marusak:
New screenshot:

logsnewfull

I fixed a few issues, should work now better. I am writing tests now - a bit trickier than it might seem to make them robust. One idea would be to have journal in archive but that seems a bit ugh, so trying it with current metrics.

marusak:
Ready for review - both code and design. Spinning tests now. I am interested to see how the new test copes as it kind of expects some event to happen. There should be one (especially in nond environement) but only tests will show.

garrett:
Oh cool! Does this solve the ">" showing up when logs are empty? That was my biggest issue when trying it out last week. 

I'll build and look.

marusak:
> Does this solve the ">" showing up when logs are empty? That was my biggest issue when trying it out last week.

Not by preloading. But now it tries different ranges and priorities to find some useful logs. Empty items should be rare. But even if empty, it will provide "All logs" button which takes you to logs page with pre-filtered search box so you can try to edit it to find what you need. Of course, if you use VM with our archived pcp data then all items are going to be empty, as there is no journal for that given time.

garrett:
I've built it, but I'm having issues running the VM for some odd reason. I'll give up tonight and try again tomorrow. It's too late to deal with misbehaving computers. :wink:

martinpitt:
I just tested it again, and it now works a whole lot better. Amazing job, really! The page is super-snappy and clean. My only complaint is that I find the time selection in "All logs..." a little to confined -- only showing a 5 s range is not all that useful, I'd at least show the previous and subsequent minute so that you have some idea what *led* to the situation of high consumption. Also, we don't have a "load later entries" button to "expand upwards" (but that is definitively follow-up material).

So functionally I think this is good enough to land, and tweaks can be done as follow-ups. I'll do a code review tomorrow morning first thing.

garrett (approved):
Oh, right; I forgot to get back to this. It's quite good now.

My one concern with this right now is an empty journal. We need to solve it. Although, I guess it's really less likely in typical use... but it's still possible and still needs to be solved.

That said, I agree with @martinpitt; I think it's good enough for this PR to land and we can start working on follow-ups. WDYT?

martinpitt:
@marusak: FYI, [testcurrentmetrics failure](https://logs.cockpit-project.org/logs/pull-15420-20210302-173138-1a242e8b-centos-8-stream/log.html#195) on RHEL/CentOS 8, that seems relevant.

test/verify/check-metrics
```diff
@@ -442,6 +444,67 @@ class TestCurrentMetrics(MachineCase):
...
+                        minute = m.getAttribute("data-minute");
+                        hour = h.id.split("-")[2];
+                        return true;
+                    }
+                    return false;
+                });
+            });
+
+            return [hour, minute];
+        })""")
```
martinpitt, line 467: I'm afraid I have some trouble to understand what this does. It seems to search for the first available non-null data and returns its hour/minute. But how does that relate to a "CPU spike", and where/how is that being "given"?

Couldn't that be done in any simpler way? Like a CSS `:first-child` or `:first-of-type` etc, or a `querySelector(..)[0]`?

test/verify/check-metrics
```diff
@@ -442,6 +444,67 @@ class TestCurrentMetrics(MachineCase):
...
+
+            # Log timestamp needs to be in reasonable range
+            timestamp = int(b.text(".info-table-ct tr td:contains('__REALTIME_TIMESTAMP') + td")) / 1000
+            self.assertGreaterEqual(timestamp, hour + (minute - 1) * 60000) # -1 as if the event happens at X:00 we check X-1 minute as well
+            self.assertLessEqual(timestamp, hour + (minute + 1) * 60000)
+
+            # Check number of logs and remember location
+            b.click("a:contains('Logs')")
+
+            if (logs >= 10):
```
martinpitt, line 495: Same here -- tests should be assertions, not conditions. The test could submit some logs, or do something like 

    systemd-run --unit timeout 30 eatmycpu dd if=/dev/urandom of=/dev/null

then query the unit's start time with `sytemctl show`, and then ensure that you eventually see the systemd message in the journal in the expected hour.

martinpitt (changes requested):
Thanks! This PR is quite a mouthful, so I'm afraid I have quite a lot of review comments as well. I'm happy to help you with the tests, of course! Especially with figuring out some predictable journal logs and such.

pkg/metrics/metrics.jsx
```diff
@@ -815,9 +815,8 @@ class MetricsHour extends React.Component {
             for (; m >= 0; m = m - 5) {
                 const dataOffset = m * SAMPLES_PER_MIN;
                 const dataSlice = normData.slice(dataOffset, dataOffset + SAMPLES_PER_MIN * 5);
-                if (dataSlice.find(i => i !== null)) {
+                if (dataSlice.some(i => i !== null && i !== undefined))
```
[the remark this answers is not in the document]
marusak: They do, the difference is that one returns True/False and the other the first item it finds. I find True/False to be better for subsequent `if` condition.

marusak:
I've submited https://github.com/cockpit-project/cockpit/pull/15484 that splits unrelated changes.
Just now pushed fixed version - mostly with the logic about dynamic logs and also tests that use logs from archive.

marusak:
`Journal file /var/log/journal/6ce25324e91848b09ac5f144c608a7cc/journal.journal uses an unsupported feature, ignoring file.`
Uh oh, backward compatibility... I recorded on F-33 but seems ubuntu, debian and even fedora-32 finds this too progressive :/

@martinpitt WDYT - should I try and record on some old debian and hope newer fedoras will accept it, or is it enough to test this only on a few systems that can work with it?

martinpitt:
@marusak : Argh -- I'd go with restricting the test to only Fedora then, that should certainly suffice.

marusak:
green on the first try! https://github.com/cockpit-project/cockpit/pull/15484 does not want to get green, but here it went fine (I included those changes here)

marusak:
Rebased since #15484 landed + fixed one forgotten review comment.

marusak:
> Failed to add match '10:39:00': Invalid argument

Meh, same as https://github.com/cockpit-project/cockpit/commit/8ed2a7db06dd9d779f7ee7524625813cc881a554

pkg/metrics/metrics.jsx
```diff
@@ -20,7 +20,12 @@
 import cockpit from 'cockpit';
 import React from 'react';
 import moment from "moment";
+
+import '../lib/patternfly/patternfly-cockpit.scss'; // PF3 for logs panel
```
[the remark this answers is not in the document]
marusak: When https://github.com/cockpit-project/cockpit/pull/15499 lands I should be able to drop this. I'll check it out.

marusak:
https://github.com/cockpit-project/cockpit/pull/15488 Fixes that unrelated error. So I'll retest once that lands.

pkg/metrics/metrics.jsx
```diff
@@ -663,10 +722,38 @@ class MetricsMinute extends React.Component {
...
+
+            const entry = [{
+                props: { key: timestamp, 'data-row-id': timestamp },
+                columns: [{ title: desc }],
+                hasPadding: false,
+                expandedContent: body,
+            }];
+
+            events = <ListingTable aria-label={ _("Event logs") }
+                                   className="metrics-events"
```
KKoukiou: I think we want to add gridBreakPoint='' here to avoid collapsing into the mobile mode:

Currently is looks like this:
Screen Shot 2021-03-11 at 11 02 03


Removing the breakpoint makes it like this:
Screen Shot 2021-03-11 at 11 04 25

What do you think?

pkg/metrics/metrics.scss
```diff
@@ -284,23 +289,41 @@
...
-        align-items: baseline;
 
-        dt {
-            grid-column: time;
-            font-size: var(--pf-global--FontSize--sm);
-            grid-row: 1;
+        .description {
+            display: grid;
+            grid-template-columns: [info] 1fr [time] 1fr;
+            align-items: baseline;
```
KKoukiou, line 298: align items and justify items are not causing any visual difference neither in desktop neither in mobile sizes when removed. 
I can see that it was before there, but now it seems to apply on another element.

pkg/metrics/metrics.scss
```diff
@@ -284,23 +289,41 @@
...
+            justify-items: start;
+
+            .time {
+                grid-column: time;
+                font-size: var(--pf-global--FontSize--sm);
+                grid-row: 1;
+                display: flex;
+                justify-content: space-between;
```
KKoukiou, line 306: This is kinda working as expected but it's a confusing approach.
.time class wraps the time and the all-logs link. In this case we should either name it differently or just let the outer grid handle the spacing. And I prefer the latter.

So we can completely remove this class if we have the outer grid be live:

grid-template-columns: [info] 1fr [time] 1fr [link];
And let's add some spacing with grid-column-gap. Then we can mostly drop the .time class content.

pkg/metrics/metrics.scss
```diff
@@ -284,23 +289,41 @@
...
+            }
+
+            .type {
+                grid-column: info;
+            }
         }
 
-        dd {
-            grid-column: info;
+        .cockpit-log-panel {
```
KKoukiou, line 315: What's the purpose of the content of this rule? I removed it and I did not notice any change. If it's needed can we add some comment?

KKoukiou (commented):
JS looks good, just some style related comments, mostly for @garrett

test/verify/check-metrics
```diff
@@ -368,6 +368,38 @@ class TestHistoryMetrics(MachineCase):
...
 
+            b.logout()
+
+        #
+        # Journal logs
+        #
+
+        # Journal was recorded on Fedora 33 and when trying to use it with older systemd it fails with:
+        # `Journal file /var/log/journal/*/journal.journal uses an unsupported feature, ignoring file.`
+        if self.machine.image != "fedora-33":
```
martinpitt, line 379: This is a bit dangerous -- it will silently stop working once we drop fedora-33. Can we make it `TEST_OS_DEFAULT`, so that we'll learn once we try and switch to 34? It's not very clean, but it's the best idea that I have to avoid a time bomb.

Alternatively, add a list of OSes where the recording does not work, like debian-stable, ubuntu-2004, rhel-8 and so on. I seriously hope journal files are forward compatible.

martinpitt (changes requested):
Thanks! Just a minor test thingy. Re-dropping PF3 (which should be possible now) and addressing Katerina's concerns, then it's ready. Cheers!

pkg/metrics/metrics.jsx
marusak: awesome! I saw this before and was not sure what was happening and then forgot about it.
Happy to see that the fix is that simple!

pkg/metrics/metrics.scss
marusak: Actually there is a reason why I did this this way. When you toggle open one item, the time gets mis-aligned with other items:

Currently:
alignedtimes

With using ` [info] 1fr [time] 1fr [link];` (notice how times are not aligned anymore)
withgrid

I'll rename it to make it less confusing.

pkg/metrics/metrics.scss
marusak: To have "Loading..." and "No logs found" in the middle

test/verify/check-metrics
marusak: Good call, this would really be a great candidate to become forgotten. I think list of OSes where it does not work is the best way forward.

marusak:
@martinpitt @KKoukiou I've adjusted your comments and dropped PF3 import. Tests went green without any retry.

KKoukiou (approved):

test/verify/check-metrics
```diff
@@ -368,6 +368,38 @@ class TestHistoryMetrics(MachineCase):
...
 
+            b.logout()
+
+        #
+        # Journal logs
+        #
+
+        # Journal was recorded on Fedora 33 and when trying to use it with older systemd it fails with:
+        # `Journal file /var/log/journal/*/journal.journal uses an unsupported feature, ignoring file.`
+        if self.machine.image in ["centos-8-stream", "rhel-8-4", "ubuntu-2004", "ubuntu-stable", "debian-stable", "debian-testing"]:
```
martinpitt, line 379: Debian testing has systemd 247.3, ubuntu-stable has 246. These ought to be sufficiently new. But this is not that important indeed.

martinpitt (approved):
Thanks, this looks good to me now for a first public version. Amazing work!

garrett:
Which is the screenshot we should use for the release notes?

marusak:
> Which is the screenshot we should use for the release notes?

First comment, where "Release note" is written.

PR #15423, opened 2021-02-24
login: Drop compat code for version < 120

martinpitt:
cockpit-ws support for `/cockpit+` URLs got introduced in version 120
(October 2016). Our master branch has not supported such old OSes for a
long time, so drop the backwards compat code for remote servers which
don't support remote logins.

This gets rid of a place that parses base1's manifest version.


martinpitt:
Split from #15396 to test it separately.

mvollmer (approved):

PR #15427, opened 2021-02-25
machines: allow creation of non-root user for unattended installations

### LINKED ISSUE #15161: Unable to create VM without root password (UI issue?)
Cockpit version: 235
OS: Ubuntu 20.04 LTS
Page: Virtual Machines -> Create Virtual Machine

When creating a VM with the "Workstation" profile, a question mark popup says "Leave the password blank if you do not wish to have a root account created". However, the UI prohibits users from proceeding if the password field is blank.

Steps to reproduce

1. Click "Create VM" button in Virtual Machines tab.
2. Fill out everything (I used Ubuntu 20.04 LTS as my VM) **except the root password** and select Workstation as the profile. Notice that clicking on the question mark reveals a popup that says you do not need to enter a root password if you don't want a root account.
3. Click the Create button. The UI will not let you continue without entering a root password.

### LINKED ISSUE #15178: [machines] ubuntu20.04 requires the user-password to be set.
cockpit 235-1 on debian 11 bullseye (from apt-get install cockpit cockpit-machines)
osinfo-db is at version 0.20201218-1

When creating a VM with "download an os" choosing ubuntu 20.04, run unattended, profile workstation, entering a non empty root password the creation fails with:

ERROR ubuntu20.04 requires the user-password to be set. Domain installation does not appear to have been successful. If it was, you can restart your domain by running: virsh --connect qemu:///system start vmtestpolluxubuntudesktop otherwise, please restart your installation.

Note: same error with ubuntu 18.04 workstation.

KKoukiou:
Fixes #15161
  Fixes #15178

### DIFF
--- a/pkg/machines/components/create-vm-dialog/createVmDialog.jsx
+++ b/pkg/machines/components/create-vm-dialog/createVmDialog.jsx
@@ -457,13 +457,11 @@ const UnattendedRow = ({
     }, [userPassword, validationFailed]);
 
     let unattendedInstallationCheckbox = (
-        <FormGroup fieldId="unattended-installation">
-            <Checkbox id="unattended-installation"
-                      isChecked={unattendedInstallation}
-                      isDisabled={unattendedDisabled}
-                      onChange={checked => onValueChanged('unattendedInstallation', checked)}
-                      label={_("Run unattended installation")} />
-        </FormGroup>
+        <Checkbox id="unattended-installation"
+                  isChecked={unattendedInstallation}
+                  isDisabled={unattendedDisabled}
+                  onChange={checked => onValueChanged('unattendedInstallation', checked)}
+                  label={_("Run unattended installation")} />
     );
     if (unattendedDisabled) {
         unattendedInstallationCheckbox = (
@@ -474,7 +472,7 @@ const UnattendedRow = ({
     }
 
     return (
-        <>
+        <FormGroup fieldId="unattended-installation">
             {unattendedInstallationCheckbox}
             {!unattendedDisabled && unattendedInstallation && <>
                 {os.profiles.length > 0 &&
@@ -527,7 +525,7 @@ const UnattendedRow = ({
                                         change={(_, value) => onValueChanged('userPassword', value)} />
                 </>}
             </>}
-        </>
+        </FormGroup>
     );
 };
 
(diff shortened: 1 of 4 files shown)


pkg/machines/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -462,12 +501,31 @@ const UnattendedRow = ({ validationFailed, unattendedDisabled, unattendedInstall
...
+                                   validated={validationFailed.userLogin ? "error" : "default"}
+                                   value={userLogin || ''}
+                                   onChange={value => onValueChanged('userLogin', value)} />
+                    </FormGroup>
+                    <PasswordFormFields password={userPassword}
+                                        password_label={_("User password")}
+                                        password_strength={user_pwd_strength}
+                                        idPrefix="create-vm-dialog-user-password"
+                                        password_message={user_pwd_message}
+                                        password_label_info={profile == 'desktop' ? _("Leave the password blank if you do not wish to have a user account created") : ""}
```
KKoukiou, line 523: This has wrong condition.

test/verify/check-machines-create
```diff
@@ -122,11 +122,14 @@ class TestMachinesCreate(VirtualMachinesCase):
...
 
+        b = self.browser
+        m = self.machine
+
         self.login_and_go("/machines")
         self.browser.wait_in_text("body", "Virtual machines")
 
         # try to CREATE few machines
-        if self.machine.image in ['debian-stable', 'debian-testing', 'centos-8-stream']:
+        if m.image in ['debian-stable', 'debian-testing', 'centos-8-stream']:
```
martinpitt, line 132: FTW, it seems really weird that this hasn't worked on centos-8-stream for so long. Are we missing some package there? It should be even al little ahead of 8.4

martinpitt (approved):
Yay, thank you! That has been a wart for quite some time.

test/verify/check-machines-create
KKoukiou: The check is wrong, it will never be present this selector, we need to open the dialog before. I will post a fix separately, ok?

I am sure debian-testing has this option as well.

KKoukiou: (This PR does not touch affect this condition. Therefore I said follow up is fine)

martinpitt: Yes, I know it's not new, it just made me raise an eyebrow as that is suspiciuos. Follow-up is totally fine of course.

KKoukiou:
I retried already twice, at least machines tests are now green.

PR #15428, opened 2021-02-25
test: Fix run-tests changed_pkgs crash for non-cockpit projects

martinpitt:
Commit 2db63f19efd4f57 regressed run-tests for non-cockpit projects:
```
fatal: ambiguous argument 'pkg/': unknown revision or path not in the working tree.
[...]
  File "starter-kit/test/common/run-tests", line 240, in <listcomp>
    changed_tests.extend([test for test in test_files if any(pkg in test for pkg in changed_pkgs)])
NameError: free variable 'changed_pkgs' referenced before assignment in enclosing scope
```

Fix this by explicitly telling git that `pkg/` is a path, by prefixing
it with `--`. Also, move the `changed_tests` extension into the right
scope, in case the git command still fails.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -229,15 +229,15 @@ def run(opts, image):
             else:
                 changed_tests = []
 
-        # Detect affected tests from changed pkg/* subdirectories
+        # Detect affected tests from changed pkg/* subdirectories in cockpit
         # If affected tests get detected from pkg/* changes, don't apply the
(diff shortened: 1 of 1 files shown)


martinpitt:
Before it was [failing like this](https://logs.cockpit-project.org/logs/pull-428-20210225-070902-367fa387-fedora-33/log.html) when I tried to bump the test/common dep in starter-kit to 238. Now it works locally.

KKoukiou (approved):
Oh right, thanks for fixing this!

PR #15430, opened 2021-02-25
machines: Expose editing disk cache mode

skobyda:
Cache is a disk property which says if and how is I/O cached.
It is already exposed when adding a disk. It can be edited only
for a shutoff VM.

The motivation to implement this was https://github.com/cockpit-project/cockpit/pull/15407 , where migration could be considered unsafe and end up with an error `Migration may lead to data corruption if disks use cache other than none or directsync`. So it's good to allow user to change cache in order to fix the error.

Screenshot from 2021-02-25 12-22-27


KKoukiou (approved):

martinpitt:
Ignoring the other machines test failures, they are impossible to get green right now.

PR #15431, opened 2021-02-25
WIP: test: tmt test plan for upstream

jscotka:
First try how it could looks like (more less structure):

#### It is blocked by several issues:
- [x] https://github.com/psss/tmt/issues/566
- [ ] https://github.com/psss/tmt/issues/588
- [x] https://github.com/psss/tmt/issues/587

### DIFF
--- a/test/plans/upstream.fmf
+++ b/test/plans/upstream.fmf
@@ -1,0 +1,25 @@
+summary: Run browser integration tests on the host
+prepare:
+  - name: Install cockpit and test dependencies
+    how: install
+    package:
+    - cockpit
+    - cockpit-ws
+    - cockpit-system
+    - cockpit-bridge
+    - git
+    - libvirt-python3
+    - make
+    - npm
+    - python3
+  - name: Install Chromium
+    script: ./prepare-chromium.sh
+  - name: Create users for tests
+    script: ./prepare-users.sh
+  - name: Start cockpit service & generic hacks
+    script: ./prepare-generic.sh
+discover:
+  how: fmf
+  test: [/verify/check-testlib/TestRunTestListing/testBasic]
+execute:
+  how: tmt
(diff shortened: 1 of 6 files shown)


martinpitt:
@jscotka and I talked on IRC. Let's split test/run-verify-host.sh into these several smaller scripts, instead of duplicating them. This should not be blocked by any TMT features, and can land right now.

test/plans/upstream.fmf
```diff
@@ -0,0 +1,25 @@
...
+    - cockpit-ws
+    - cockpit-system
+    - cockpit-bridge
+    - git
+    - libvirt-python3
+    - make
+    - npm
+    - python3
+  - name: Install Chromium
+    script: ./prepare-chromium.sh
```
martinpitt, line 16: @jscotka: Does that work for you? I tried this locally with `tmt run`, and it is *entirely* ignored.

martinpitt: See https://github.com/martinpitt/cockpit/tree/tmt-split-setup

jscotka: Hi, no, there were issue, that it was not able to find this local path and had to be workarounded by some `../../` magic
I remember, we've done some issue and PR's exacly to this issue commected with this cockpit PR, because I've met it and reported.
But it should be fixed  in upstream, because there is the test what I remember: https://github.com/psss/tmt/tree/master/tests/prepare/shell/data
so maybe try to install/upgrade to new TMT, not sure if the release containing this feature is already shipped to channel what you use for your installation

jscotka: yep, your changes in https://github.com/martinpitt/cockpit/tree/tmt-split-setup seems very well. it is exactly What I've expected.

KKoukiou:
This is not touched 1.5 years now, please reopen if still interested.

PR #15432, opened 2021-02-25
machines: replace usage of ct-form-split with PF4 grid as suggested in the docs

KKoukiou:
See https://www.patternfly.org/v4/components/form/html-demos/#grid

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -102,7 +102,7 @@ svg {
 // Add some spacing to nested form groups - PF4 does not support these yet
 // https://github.com/patternfly/patternfly-design/issues/1012
 .pf-c-form__group-control {
-    > .pf-c-form__group, .pf-c-form__section {
+    .pf-c-form__group, .pf-c-form__section {
         padding-top: var(--pf-global--spacer--md);
     }
 }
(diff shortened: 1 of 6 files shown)


garrett (changes requested):
The label for the split feels too far away from the dropdown (or the dropdown could be too small... stretching it could be a fix)

Screenshot_2021-02-25 Virtual Machines - garrett Rain

Same issue for disk adding:

Screenshot_2021-02-25 Virtual Machines - garrett Rain(2)

KKoukiou:
@garrett the distance between the labels is the same in master I think. It did not really change from the looks at all.

garrett (approved):
Oh, I didn't check master.

I still think the labels are a little off, but if this PR doesn't change the look, I'm all for it!

:+1: 

(I guess next up, outside of this PR, is `pkg/storaged/dialog.jsx`, which is the last part of Cockpit where we're using ct-form-split? And perhaps then Cockpit-Podman, if it's using it?)

KKoukiou:
> Oh, I didn't check master.
> 
> I still think the labels are a little off, but if this PR doesn't change the look, I'm all for it!
> 
> +1
> 
> (I guess next up, outside of this PR, is `pkg/storaged/dialog.jsx`, which is the last part of Cockpit where we're using ct-form-split? And perhaps then Cockpit-Podman, if it's using it?)

Right, I did not touch storage to not have conflicts with the storage PR :)

martinpitt:
This is conflict-y now, needs rebasing.

KKoukiou:
@garrett can you please reapprove? I rebased and repushed because it was conflicty.

martinpitt (approved):
Interdiff looks clean, thanks! (Garrett is on PTO today)

PR #15436, opened 2021-02-28
Terminal: Add option to select font size

### LINKED ISSUE #11261: Settable size of font in terminal
Page:  Terminal

It is possible to set size of text in xterm.js
It would be nice to have, mostly as accessibility thing, where for some people size 12 of text can be rather small. It just needs design where this option should be placed.

Also see https://github.com/cockpit-project/cockpit/issues/11177

marusak:
Fixes #11261

Release note:

Terminal: Add option to choose font size

Users can change font size in terminal that fits their preferences.

fontsizeselector

### DIFF
--- a/pkg/lib/cockpit-components-terminal.jsx
+++ b/pkg/lib/cockpit-components-terminal.jsx
@@ -98,12 +98,12 @@ export class Terminal extends React.Component {
         super(props);
         this.onChannelMessage = this.onChannelMessage.bind(this);
         this.onChannelClose = this.onChannelClose.bind(this);
-        this.onWindowResize = this.onWindowResize.bind(this);
         this.connectChannel = this.connectChannel.bind(this);
         this.disconnectChannel = this.disconnectChannel.bind(this);
         this.reset = this.reset.bind(this);
         this.focus = this.focus.bind(this);
         this.onWindowResize = this.onWindowResize.bind(this);
+        this.resizeTerminal = this.resizeTerminal.bind(this);
         this.onFocusIn = this.onFocusIn.bind(this);
         this.onFocusOut = this.onFocusOut.bind(this);
         this.setText = this.setText.bind(this);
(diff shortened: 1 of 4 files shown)


martinpitt (changes requested):
Thanks, this is a nice improvement! However, that number input widget leaves lots to be desired:

 - Trying to input a number with the keyboard basically does not work at all. Try to select the number and clear the input, or try to overwrite the selection with something like "12", this all gets caught in overzealous input validation. The "1" gets auto-converted to a 6 (minimal size), and then you can't type the "2" any more.
 - With the - and + buttons you can exceed the boundaries.
 - Trying to click '-' again on font size 2 makes the browser spin something in the JS, and eventually shows the "a script is slowing down the page...". Then not even a forced reload works any more.

I suppose we should use a more field-tested and proven widget.

Please test the boundary conditions and direct input in the tests. 

I suppose we need to use a different widget for this

KKoukiou:
I agree with @martinpitt that the NumberInput component is far from optimal. how about using a <TextInput  type="number" /> instred?

marusak:
Yeah, I'll get into this once I am done with metrics. I'll check how this can be improved and report back to PF if there are some issues.

marusak:
Fixed all found issues. Martin mentioned on IRC that we should have non-linear steps but when I was looking into it I was not convinced.
It would save just a few clicks in the end, as between 6 to ~18 we want to have step size 1. And then between ~18 to 40 there is not that many numbers. Also it is very unlikely that someone would go out of range <10, 22> and starting at 16 you are a few clicks away. And this is mostly one-off action as you set it up and then keep using it. It is not like creating new container where every time you would need to click away. So for keeping it simple I just kept the +-1 steps.

KKoukiou (approved):

martinpitt (approved):
This looks nice now, thank you! Agreed about the exponential vs. linear increase -- if someone complains, we can refine this. Please coordinate landing with @allisonkarlitskaya , as this is affecting release notes.

PR #15438, opened 2021-02-28
ws: Fix remotectl --ensure for read-only /etc

martinpitt:
If the certificate already exists, and ownership and permissions are
already as expected, then don't call chmod()/chown(). This avoids
failing on a read-only /etc/ once the certificate is set up.

https://bugs.debian.org/983514

### DIFF
--- a/src/ws/remotectl-certificate.c
+++ b/src/ws/remotectl-certificate.c
@@ -73,6 +73,7 @@ set_cert_attributes (const gchar *path,
   struct passwd *pwd = NULL;
   struct group *gr = NULL;
   gint status = 0;
+  GStatBuf statbuf;
   mode_t mode;
   int ret = 1;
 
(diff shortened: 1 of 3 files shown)


martinpitt:
Whoops! Fixed the [crash](https://logs.cockpit-project.org/logs/pull-15438-20210228-104620-3abb99a2-fedora-33/log.html#314)

martinpitt:
New test [fails on coreos](https://logs.cockpit-project.org/logs/pull-15438-20210228-122735-1558239f-fedora-coreos/log.html#322)

src/ws/remotectl-certificate.c
```diff
@@ -106,16 +107,23 @@ set_cert_attributes (const gchar *path,
         }
     }
 
+  if (g_stat (path, &statbuf) < 0)
+    {
+      g_message ("could not stat certificate: %s: %s", path, g_strerror (errno));
+      goto out;
+    }
+
```
allisonkarlitskaya, line 115: this is getting out of hand....

Could we just set the correct permissions when creating the certificate and remove all of this extra logic?

If the admin went out of their way to change the permissions, who are we to say otherwise?

allisonkarlitskaya (changes requested):

src/ws/remotectl-certificate.c
martinpitt: Good point -- I'll look into that. Thanks!

martinpitt:
I discussed that with @allisonkarlitskaya . We rely on that "change permissions of existing certificate" logic for the 10-ipa one, as the JS code on the page does not know the correct group. This will require some larger refactoring, so maybe we want to land this as an intermediate fix to get rid of that bug?

allisonkarlitskaya (approved):
Yes.

Issue #2454, opened 2015-06-29
Don't show red bar for CD-ROM or non-writable partitions


stefwalter:
Don't show a red bar "storage full" bar for CD-ROM or non-writable mounts.

stefwalter:
![screenshot from 2015-06-29 13-27-33](https://cloud.githubusercontent.com/assets/795070/8407002/a0ecd670-1e62-11e5-94f3-f1be235dc732.png)

stefwalter:
In this case the red partition is an ISO written to a USB stick. But this should apply to other cases as well.

FireDrunk:
I want to fix this, but i don't know how to get the filesystem options from storaged backend.
I think we should add a utils.format_usage function which takes the device, checks if it's mounted readonly and if it's full, and returns the appropriate progress-bar-* class.

stefwalter:
@mvollmer would be the one to ask about specifics.

mvollmer:
One option is to look at org.storaged.Storaged.Block.ReadOnly.  This doesn't catch all read-only mounts, only genuinly read-only block devices.

I don't think Storaged can tell us much about the actual run-time mounts.  That might be worth fixing.

FireDrunk:
So do you want this to be fixed upstream? Or do we build something around the issue?

mvollmer:
> So do you want this to be fixed upstream? Or do we build something around the issue?

IMO, we should look at org.storaged.Storaged.Block.ReadOnly as a first step and not show red bars when it is true.

@stefwalter said he doesn't want red bars for read-only mounts in general, such as when you mount your rootfs read-only, as on Fedora Atomic.  I think Storaged doesn't readily give us this information, so as a second step, we would need to start looking at extending Storaged.

Are you up for both steps? That would be awesome.

martinpitt:
@mvollmer mentioned in #15434: We should filter out all devices with UDISKS_IGNORE set, and that snap should set UDISKS_IGNORE on its devices. See https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1714941

mvollmer:
> @mvollmer mentioned in #15434: We should filter out all devices with UDISKS_IGNORE set, and that snap should set UDISKS_IGNORE on its devices. See https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1714941

That would be for filtering out the actual loop devices from the "Other devices" tab.  I'd say this issue should be just about read-onlyness of mounts.

PR #15439, opened 2021-02-28
metrics: Hide read-only loop mounts

### LINKED ISSUE #15434: metrics: Don't show read-only mounts for Ubuntu snaps
Cockpit version: 238
OS: Ubuntu 20.04
Page: Metrics

This is useless and unnecessarily alarming:

image

Let's filter out /snap in particular, or possibly even all read-only mounts.

This can be seen in /livetest machines.

martinpitt:
These are often things like snaps or mounted ISO images, and are
usually 100% used. Hide them from the "Disks" card, as that is not an
alerting condition, and they take attention away from the "real" disks.

Fixes #15434

-----

This is similar to issue #2454, but on the metrics page. I'll send a PR for the storage page as well once this is in.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -240,15 +240,37 @@ class CurrentMetrics extends React.Component {
         }
     }
 
+    /* Return Set of mount points which should not be shown in Disks card */
+    hideMounts(procMounts) {
+        const result = new Set();
+        procMounts.trim().split("\n")
+                .forEach(line => {
+                    // looks like this: /dev/loop1 /var/mnt iso9660 ro,relatime,nojoliet,check=s,map=n,blocksize=2048 0 0
+                    const fields = line.split(' ');
+                    const options = fields[3].split(',');
+
+                    /* hide read-only loop mounts; these are often things like snaps or iso images
+                     * which are always at 100% capacity, but are uninteresting for disk usage alerts */
+                    if ((fields[0].indexOf("/loop") >= 0 && options.indexOf('ro') >= 0))
+                        result.add(fields[1]);
+                });
+        return result;
+    }
+
     updateMounts() {
-        /* df often exits with non-zero if it encounters any file system it can't read; but that's fine, get info about all the
-         * others */
-        cockpit.script("df --local --exclude-type=tmpfs --exclude-type=devtmpfs --block-size=1 --output=target,size,avail,pcent || true",
-                       { err: "message" })
-                .then(output => {
+        Promise.all([
+            /* df often exits with non-zero if it encounters any file system it can't read;
+               but that's fine, get info about all the others */
+            cockpit.script("df --local --exclude-type=tmpfs --exclude-type=devtmpfs --block-size=1 --output=target,size,avail,pcent || true",
+                           { err: "message" }),
+            cockpit.file("/proc/mounts").read()
+        ])
+                .then(([df_out, mounts_out]) => {
+                    const hide = this.hideMounts(mounts_out);
+
                     // skip first line with the headings
                     const mounts = [];
-                    output.trim()
+                    df_out.trim()
                             .split("\n")
                             .slice(1)
                             .forEach(s => {
@@ -257,6 +279,9 @@ class CurrentMetrics extends React.Component {
                                     console.warn("Invalid line in df:", s);
                                     return;
                                 }
+
+                                if (hide.has(fields[0]))
+                                    return;
                                 mounts.push({
                                     target: fields[0],
                                     size: Number(fields[1]),
@@ -272,7 +297,7 @@ class CurrentMetrics extends React.Component {
                     window.setTimeout(this.updateMounts, 10000);
                 })
                 .catch(ex => {
-                    console.warn("Failed to run df:", ex.toString());
+                    console.warn("Failed to run df or read /proc/mounts:", ex.toString());
                     this.setState({ mounts: [] });
                 });
     }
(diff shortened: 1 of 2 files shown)


martinpitt:
Still needs a test, but I'd like to try this in livetest on an Ubuntu image with several snaps installed.

martinpitt:
With master:

image


With this PR:

image

mvollmer:
CD-ROMs are not loop mounts, no?  Would it make sense to just skip all read-only mounts?

pkg/metrics/metrics.jsx
```diff
@@ -257,6 +278,9 @@ class CurrentMetrics extends React.Component {
                                     console.warn("Invalid line in df:", s);
                                     return;
                                 }
+
+                                if (hide.indexOf(fields[0]) >= 0)
```
mvollmer, line 282: This is quadratic, and the Enterprise will find a way to make that a problem. Let's use a hash table or Set for `hide`.

mvollmer (changes requested):

martinpitt:
snap mounts look like this:

    /var/lib/snapd/snaps/snapd_11036.snap on /snap/snapd/11036 type squashfs (ro,nodev,relatime,x-gdu.hide)

So checking for that flag would work as well. In fact, I think we should do both?

martinpitt:
> CD-ROMs are not loop mounts, no? 

Right, with "mounted CD-ROM images" I meant "mounted ISO files". I should clean up the commit message.

> Would it make sense to just skip all read-only mounts?

I was thinking about this, and I can't quite make up my mind. I think I would still like to see my read-only /usr partition. Hence I started a little more cautiously. loop + x-gdu-mount feels about right, WDYT?

pkg/metrics/metrics.jsx
martinpitt: Heh, for the few (there are not that many loop devices usually), the overhead of a Set will probably be bigger; but indeed it would look a bit cleaner, I'll do that.

mvollmer:
> > CD-ROMs are not loop mounts, no?
> 
> Right, with "mounted CD-ROM images" I meant "mounted ISO files". I should clean up the commit message.

Ah, makes sense, I missed the "images" part.
 
> > Would it make sense to just skip all read-only mounts?
> 
> I was thinking about this, and I can't quite make up my mind. I think I would still like to see my read-only /usr partition.

Because OS updates will actually write to it and might run out of space?  Yes, makes sense as well.

> Hence I started a little more cautiously. loop + x-gdu-mount feels about right, WDYT?

Sounds good.

martinpitt:
That is a bit awkward -- `x-gdu.hide` does not actually appear in /proc/mounts, but /run/mount/utab. Not even `findmnt` is parsing the latter. So we either need to parse that internal file (ugh), or call `mount` (much more expensive), but that does not show /dev/loop,  but the underlying file. So we'd even have to parse *both*.

pkg/metrics/metrics.jsx
martinpitt: Rewritten to use Set.

martinpitt:
Due to the above, I limited this to ro loop mounts for now. I propose to do x-gdu.hide in a follow-up, as it's rather difficult to test for.

Integration test case (just to avoid losing it):

```diff
--- test/verify/check-metrics
+++ test/verify/check-metrics
@@ -536,19 +536,22 @@ class TestCurrentMetrics(MachineCase):
 
         # Disk usage
 
-        # add 50 MB loopback disk; mount it once rw and once ro
+        # add 50 MB loopback disk; mount it once rw, once ro, and once with marking "ignore"
         m.execute("""set -e
                   F=$(mktemp /var/tmp/loop.XXXX)
                   dd if=/dev/zero of=$F bs=1M count=50
                   mkfs -t ext3 $F
-                  mkdir -p /var/cockpittest /var/cockpit-ro-test
+                  mkdir -p /var/cockpittest /var/cockpit-ro-test /var/cockpit-ignore-test
                   mount -o loop $F /var/cockpittest
                   RODEV=$(losetup -f --show $F)
                   mount -r $RODEV /var/cockpit-ro-test
                   losetup -d $RODEV
+                  IGNOREDEV=$(losetup -f --show $F)
+                  mount -o x-gdu.hide $IGNOREDEV /var/cockpit-ignore-test
+                  losetup -d $IGNOREDEV
                   rm $F
                   """)
-        self.addCleanup(m.execute, "umount /var/cockpittest /var/cockpit-ro-test")
+        self.addCleanup(m.execute, "umount /var/cockpittest /var/cockpit-ro-test /var/cockpit-ignore-test")
 
         self.assertLess(progressValue(b, ".pf-c-progress[data-disk-usage-target='/var/cockpittest']"), 5)
         progress_sel = ".pf-c-progress[data-disk-usage-target='/var/cockpittest'] .pf-c-progress__status"
@@ -562,6 +565,8 @@ class TestCurrentMetrics(MachineCase):
         b.mouse(progress_sel, "mouseleave")
         # read-only loop devices are not shown
         self.assertFalse(b.is_present(".pf-c-progress[data-disk-usage-target='/var/cockpit-ro-test']"))
+        # ignored mounts are not shown
+        self.assertFalse(b.is_present(".pf-c-progress[data-disk-usage-target='/var/cockpit-ignore-test']"))
 
         m.execute("dd if=/dev/zero of=/var/cockpittest/blob bs=1M count=40")
         b.wait(lambda: progressValue(b, ".pf-c-progress[data-disk-usage-target='/var/cockpittest']") >= 90)
```

martinpitt:
Wah wah [eslint](https://logs.cockpit-project.org/logs/pull-15439-20210301-093230-a5e5ed80-debian-stable/log.html) -- fixed.

mvollmer (approved):

PR #15442, opened 2021-03-01
test: Fix VMVirtualMachinesCaseHelpers.performAction() for disabled elements

### LINKED ISSUE #15441: TestMachinesLifecycle.testBasic fails for cloning
Page: Machines

This is our current top flake, and was breaking most PRs last week:
```
-> wait: ph_is_present(".pf-c-modal-box__title-text")
<- {'type': 'object', 'subtype': 'error', 'className': 'PhWaitCondTimeout', 'description': 'Error: condition did not become true\n    at <anonymous>:245:24', 'objectId': '6058255692305731313.5.1'}
  File "/var/home/martin/upstream/cockpit/test/verify/check-machines-lifecycle", line 141, in _testBasic
    b.wait_text(".pf-c-modal-box__title-text", "Create a clone VM based on subVmTest1")

testlib.Error: timeout
```
The [screenshot](https://logs.cockpit-project.org/logs/pull-15430-20210225-181217-907b9fef-rhel-8-4-distropkg/TestMachinesLifecycle-testBasic-rhel-8-4-distropkg-127.0.0.2-2301-FAIL.png) shows that the action menu gets opened and shows the "clone" action, but it never closes and the expected clone dialog never appears.

This reproduces locally just fine (I am using centos-8-stream). Let's use this issue to collect some debugging information.

martinpitt:
The kebab actions with disabled elements are rendered like this
(simplified):

    <li role="menuitem" id="vm-subVmTest1-clone">
        <a aria-disabled="true" class="pf-m-disabled pf-c-dropdown__menu-item">Clone</a>
    </li>

Thus clicking on the -clone ID directly happily clicked into the void
even if the contained `<a>` link was still disabled.

This fixes at least `TestMachinesLifecycle.testBasic` and possibly other
flakes as well.

Fixes #15441

### DIFF
--- a/test/verify/check-machines-settings
+++ b/test/verify/check-machines-settings
@@ -236,7 +236,7 @@ class TestMachinesSettings(VirtualMachinesCase):
         b.select_from_dropdown("#cpu-model-select-group select", "host-model")
         b.click("#cpu-config-dialog-apply")
         b.wait_in_text("#vm-subVmTest1-cpu-model .pf-c-description-list__text", "host")
-        self.performAction("subVmTest1", "run")
+        b.click("#vm-subVmTest1-run")
         b.wait_in_text("#vm-subVmTest1-cpu-model .pf-c-description-list__text", "custom")
         # In the test ENV libvirt does not properly set the CPU model so we see a tooltip https://bugzilla.redhat.com/show_bug.cgi?id=1913337
         # b.wait_not_present("#cpu-tooltip")
--- a/test/verify/machineslib.py
+++ b/test/verify/machineslib.py
@@ -28,7 +28,7 @@ class VirtualMachinesCaseHelpers:
         b = self.browser
         b.click("#vm-{0}-action-kebab button".format(vmName))
         b.wait_visible("#vm-{0}-action-kebab > .pf-c-dropdown__menu".format(vmName))
-        b.click("#vm-{0}-{1}".format(vmName, action))
+        b.click("#vm-{0}-{1} a".format(vmName, action))
 
         if not checkExpectedState:
             return


KKoukiou (approved):

martinpitt:
This seems to [regress](https://logs.cockpit-project.org/logs/pull-15442-20210301-075609-c783c373-fedora-33/log.html#84-2) for the "Run" button, which is presumably not an `<a>`.

KKoukiou:
> This seems to [regress](https://logs.cockpit-project.org/logs/pull-15442-20210301-075609-c783c373-fedora-33/log.html#84-2) for the "Run" button, which is presumably not an `<a>`.

Ups, I think the right way to fix it is to make everything button, by using component='button' property. After all performing the lifecycle operations does not belong to the spectrum of a link. Links are mostly for redirecting to other pages, anything that involves a  backend operation  should be implemented with a button.

martinpitt:
@KKoukiou : No worries, I have a separate commit for fixing that test case. I just wait for this round to complete before pushing, just in case there is another failure.

As to whether the kebab items should be a or button, I don't have a strong opinion. But IMHO this should be a follow-up, let's first fix the tests?

KKoukiou (approved):

martinpitt:
Wah wah [eslint](https://logs.cockpit-project.org/logs/pull-15439-20210301-093230-a5e5ed80-debian-stable/log.html) -- fixed.

martinpitt:
Not bad, not bad -- There are now just two Machines related failures. [TestMachinesSnapshots.testSnapshots](https://logs.cockpit-project.org/logs/pull-15442-20210301-092940-13120496-debian-testing/log.html) (which is a time diff flake, unrelated to React and such; retry worked); and [TestMachinesSettings.testCPUModel](https://logs.cockpit-project.org/logs/pull-15442-20210301-093457-13120496-fedora-32-firefox/log.html#322).

martinpitt:
The firefox test failed again, and at the very spot that I changed. I'll investigate this.

martinpitt:
It's indeed a different kind of flake. This turned green now, so I land this because it fixes our top flake.

PR #15443, opened 2021-03-01
build: major manifest cleanup

allisonkarlitskaya:
With previous commits having removed all compile-time variable content
from the manifests, we can now do our major cleanup:

 - rename all manifest.json.in to manifest.json in git

 - drop the relevant rules for performing variable substitution to
   create .json files from .json.in files

 - in place of the above, add a copy rule to webpack for copying the
   files from pkg/ to dist/

 - stop copying ssh/ and pcp/ into dist/, prefering to install their
   manifests directly out of $(top_srcdir)/pkg/

 - clean up the main webpack rule to remove the special cases for pcp
   and ssh.  Remove the mkdir -p: we always run webpack now, and webpack
   will create the directory for us.

 - update tools/min-base-version to look in pkg/ instead of dist/


allisonkarlitskaya:
```
+ grep -q pkg/systemd/manifest.json po/cockpit.pot
Error: Process completed with exit code 1.
```

I guess we need to regenerate this...

martinpitt (approved):
Ooh, this is nice. The hard work is paying off! :tada:

martinpitt:
Quick analysis of [firefox failures](https://logs.cockpit-project.org/logs/pull-15443-20210302-062758-9525688c-fedora-32-firefox/log.html):
 - `TestAccounts.testRootLogin`: affected test, two successes in three runs, Worked 4 times locally, and this really just smells like a race condition in the tests which log out too early. Either disabling the preloads or allowing the "not defined" message in logout() would both help, I'll send a PR for that.
 - `TestHostSwitching.testNoAutoconnect`: Similar root cause and situation as testRootLogin. 
 - `TestMultiMachineAdd.testBasic`: Succeeds locally. Not a regression, happened e.g. [here](https://logs.cockpit-project.org/logs/pull-15396-20210224-131745-da36b1b2-fedora-32-firefox/log.html#159) and in dozens of other runs as well. Curiously all on firefox.
 - `TestStorageResize.testResizeLuks`: Test works locally. Not a regression, happened e.g. [here](https://logs.cockpit-project.org/logs/pull-15396-20210224-131132-da36b1b2-ubuntu-stable/log.html#219) or [here](https://logs.cockpit-project.org/logs/pull-15334-20210215-101615-70823dec-fedora-32-firefox/log.html) as well. The screenshot says "storage cannot be managed on this system", which happens when pkg/storaged/client.js does not initialize.

Thus, I am now convinced that these are "just" unrelated bugs.

martinpitt:
Sent PR #15448 to hopefully address the first two flakes.

PR #15445, opened 2021-03-01
build: fix `make uninstall` if no debug is present

allisonkarlitskaya:
We have some custom rules in `make uninstall` to deal with removing the
installed files from dist/, including a separate rule for map files.
We don't always install map files anymore, and due to the changes in
96027bf8, don't even create the directory in that case.

This doesn't play nicely with our uninstall rule which assumes that the
directory will definitely exist.  Add a check.

Add a check also to the equivalent rule for the normal installation
files: this makes `make uninstall` nicely idempotent.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -250,10 +250,12 @@ install-data-local:: $(WEBPACK_INSTALL)
 install-data-local:: $(wildcard $(top_srcdir)/dist/*/*.map)
 	$(V_TAR) tar -c --transform="s@.*dist/@$(debugdir)$(pkgdatadir)/@" -T/dev/null $^ | tar --no-same-owner -C $(DESTDIR)/ -xv
 uninstall-local::
-	find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type f -exec rm -f {} \;
-	find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete
-	find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type f -delete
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -250,10 +250,12 @@ install-data-local:: $(WEBPACK_INSTALL)
 install-data-local:: $(wildcard $(top_srcdir)/dist/*/*.map)
 	$(V_TAR) tar -c --transform="s@.*dist/@$(debugdir)$(pkgdatadir)/@" -T/dev/null $^ | tar --no-same-owner -C $(DESTDIR)/ -xv
 uninstall-local::
-	find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type f -exec rm -f {} \;
-	find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete
-	find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type f -delete
-	find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete
+	test ! -d $(DESTDIR)$(pkgdatadir) || \
```
martinpitt, line 253: Feels a bit strange that we need this bit, as this was only about debug files.

martinpitt (approved):
Looks harmless enough, thanks!

Re-running the failed tests. I landed the fix for machines-lifecycle, and fedora-coreos did not start (timed out in image-prepare)

Makefile.am
allisonkarlitskaya: > Add a check also to the equivalent rule for the normal installation
files: this makes make uninstall nicely idempotent.

PR #15447, opened 2021-03-02
storage: remove <td> from the Jobs data list cell

KKoukiou:


martinpitt (approved):
Thanks for the cleanup!

PR #15448, opened 2021-03-02
tests: Avoid "ReferenceError: cockpit is not defined" failures

martinpitt:
These two tests do logout() quickly after logging in. This often causes
the asynchronously loading preloads to load too late, running into
"ReferenceError: cockpit is not defined.

Avoid that by disabling the preloads in these tests, they are not
relevant for them.

----

Seen in https://logs.cockpit-project.org/logs/pull-15443-20210302-062758-9525688c-fedora-32-firefox/log.html in PR #15443

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1329,7 +1329,7 @@ class MachineCase(unittest.TestCase):
         If the directory needs to survive reboot, `reboot_safe=True` needs to be specified; then this
         will just backup/restore the directory instead of bind-mounting, which is less robust.
         '''
-        if not self.is_nondestructive():
+        if not self.is_nondestructive() and not self.machine.ostree_image:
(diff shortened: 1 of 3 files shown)


martinpitt:
This is far from elegant, better ideas greatly appreciated!

martinpitt:
[breaks coreos](https://logs.cockpit-project.org/logs/pull-15448-20210302-091642-70f52d66-fedora-coreos/log.html#318). Meh, I thought I fixed that last week.

marusak (approved):
Yeah a bit "meh", but if it helps, then surely good enough

mvollmer:
I see this error every time I load the login page.  The PO machinery substitutes code into it that unconditionally calls `cockpit.locale`, but the login page itself does not load base1/cockpit.js.

PR #15449, opened 2021-03-02
test/make_dist: Don't download if ./dist/ already exists

martinpitt:
Prefer an existing local build/webpack, to make sure that local
developer settings are applied.

### DIFF
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -151,7 +151,6 @@ def make_dist(download_only=False, wait_download=False):
         if not download_only and not wait_download:
             source = build_dist()
         else:
-            print("make_dist: Download failed: pre-built dist tarball does not exist")
             sys.exit(1)
     return source
 
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -142,11 +142,12 @@ def download_dist(wait=False):
 
 
 def make_dist(download_only=False, wait_download=False):
-    # first try to download a pre-generated dist tarball; this is a lot faster
-    # but these tarballs are built for production NPM mode
     source = None
-    if os.getenv("NODE_ENV") != "development":
+    # on an unbuilt tree, try to download a pre-generated dist tarball; this is a lot faster
+    # these tarballs are built for production NPM mode
+    if not os.path.exists("dist") and os.getenv("NODE_ENV") != "development":
         source = download_dist(wait_download)
+
     if not source:
         if not download_only and not wait_download:
             source = build_dist()


test/make_dist.py
```diff
@@ -61,6 +61,10 @@ def download_dist():
     This can happen because the current directory is not a git checkout, or it is
     a SHA which is not pushed/PRed, or there is no ~/.config/github-token available.
     '''
+    if os.path.exists("dist"):
```
allisonkarlitskaya, line 64: I think it would be cleaner to rework the logic in `make_dist()` below to default to calling `build_dist()` if `dist/` already exists, and only call `download()` if we're sure that's what we want to do.

allisonkarlitskaya (changes requested):

test/make_dist.py
martinpitt: I was actually thinking about doing it the other way around: Right now most of the checks are inside download_dist() already, and only the NODE_ENV check is in the caller. Does that work for you as well?

martinpitt: .. of course I meant: ... so I'd rather move the NODE_ENV check into download_dist() as well. It's the same kind of conditional check as "local changes" or "no tarball available" or "no token"

martinpitt: Done.

martinpitt:
@allisonkarlitskaya: You mentioned yesterday that you still have some different goals and are unhappy with current make_dist. How about we collect a comprehensive list of use cases and write down what exactly should happen. Then we can see where we just need to adjust make_dist for some, and whether there really are some cases where "you" and "I" really want/need a different behaviour -- in other words, which new options we need to add. Please edit this comment, then we can iterate a bit.

## build modes:

 - **DLOAD**: download tarball, unpack dist/ and node_modules/
 - **INCR**: download tarball, unpack dist/ and node_modules; needs `tools/npm-install` to get full `node_modules`, then `tools/webpack-watch <page>` works with *just* rebuilding `<page>` webpack
 - **BUILD**: no download; autogen.sh and make (possibly with NO_NPM), or webpack; image-prepare does full autogen+make+webpack+make dist


## Developer who wants to test master or a PR:
 1. `NODE_ENV=development`, any tree state: **BUILD**; all other cases assume `NODE_ENV` empty or `production`
 2. pristine checkout: **DLOAD**. Then testing immediately works through the ~/.local/share/cockpit symlink, and `image-prepare` is really fast as it has the tarball already
 3. checkout without tracked changes (empty `git diff`), but already existing `dist/`: **BUILD**, let's not mess that up

## Developer who wants to change something:

 4. checkout with changes only to test/: **DLOAD**, as test/ does not influence the dist/VM. There may be more directories where this makes sense
 5. checkout with uncommitted changes to pkg (and possibly src): **INCR**
 6. checkout with local committed changes: **INCR**, as no pre-built tarball is available

We don't currently support **INCR** well -- you have to manually run `npm-install` and then `touch dist/*/stamp` to avoid rebuilding everything, and even that is a bit brittle. We need some tool to help us with that, and make sure in our unit tests that it stays working.

AFAICS, with this PR the behaviour is as specified when INCR degenerates to a full BUILD.

Looking forward to your expectations!

test/make_dist.py
```diff
@@ -159,15 +168,11 @@ def download_dist():
 
 def make_dist(download_only=False):
     # first try to download a pre-generated dist tarball; this is a lot faster
-    # but these tarballs are built for production NPM mode
-    source = None
-    if os.getenv("NODE_ENV") != "development":
-        source = download_dist()
+    source = download_dist()
```
allisonkarlitskaya, line 171: I think we lost the logic to check for `dist/` already existing here?

allisonkarlitskaya (changes requested):

test/make_dist.py
martinpitt: What do you mean? It's part of the checks that `download_dist()` does by itself.

allisonkarlitskaya (dismissed):
I'm approving this because it's not incorrect and because it unblocks you.  We still need to bikeshed a lot more on the big picture here, though :)

test/make_dist.py
allisonkarlitskaya: This thread is what i meant when I asked "where did the check go?"

At some point you moved it into `make_dist()` but now it's back here again...

allisonkarlitskaya (changes requested):

test/make_dist.py
martinpitt: Right, I moved the one remaining check *from* make_dist() *to* download_dist().

martinpitt: OK, moved to make_dist(). I can't say I like it, but I don't have a strong opinion about it :grin: Let's just fix this bug.

allisonkarlitskaya (approved):
Sorry for blocking this for so long.

Thanks :)

PR #15450, opened 2021-03-02
Reverting "tests: firefox sets the partition size slider to 25.0"

mvollmer:


martinpitt (approved):
Cheers! I was staring at that failure for a bit this morning already, and could not reproduce it locally (with firefox). So apparently for me it's indeed 25.0. Or maybe it is that for a short time only, and it changes to 25.1?

Anyway, assuming tests are green, +1. But please don't retry this -- if it's random between 25.0 and 25.1 we need to make the test a bit more liberal, or better yet, do better rounding in our code.

martinpitt:
All green, so let's get this in. (and it succeeded three times) Danke!

PR #15451, opened 2021-03-02
shell: Emit "ready" signal for machines at the right time

mvollmer:
It used to be emitted after the very first call to "refresh", which
would usually be the one that has all the data, but sometimes the
first call only has the overlays for localhost, or similar, and then
navigation would be done too soon.

The effect would be that Cockpit would not connect to the remote
machine immediately after log in, when there is a remote machine in
the initial URL, like "https://foo.org:9090/@bar.foo.org"

Now the Loader will explicitly trigger the "ready" signal when it has
done it's initial loading.

### DIFF
--- a/pkg/shell/machines/machines.js
+++ b/pkg/shell/machines/machines.js
@@ -122,9 +122,6 @@ function Machines() {
         if (!shared)
             return;
 
-        var emit_ready = !self.ready;
-
-        self.ready = true;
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice work! Gosh, all these bugs.. Glad we are taking some time to hunt them down!

PR #15350, opened 2021-02-16
Fedora 34

KKoukiou:
- [x] Move our fedora-33 image to cloud first, and fix tests: https://github.com/cockpit-project/bots/pull/1719 (as a lot of failures are due to that)
 - [x] `test/verify/check-kdump TestKdump.testBasic`: [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1933970), known issue https://github.com/cockpit-project/bots/issues/1734 , needs naughty pattern
 - [x] `test/verify/check-static-login TestLogin.testBasic` root bridge core dump -- *update*: pitti can't reproduce any more, likely fixed with the utmp binary fix below. Will re-open if CI disagrees.
 - [x]  test/verify/check-system-journal TestJournal.testBasic systemd changed log message https://github.com/systemd/systemd/commit/f5ec78e503212855eb2adf92628023148afcea6f
 - [x]  test/verify/check-shell-active-pages, check-systemd-terminal fail because of failed unit appearing on the first terminal line
 - [x]  test/verify/check-users TestAccounts.test* parsing utmp closes the cockpit channel - passing the binary: true option fixes the issue
 - [x] check-realms realm join failure for AD: [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1934124), known issue https://github.com/cockpit-project/bots/issues/1735 : https://github.com/cockpit-project/bots/pull/1740
 - [x] `TestStorageFormat.testFormatTypes` failure is https://github.com/cockpit-project/bots/issues/1733 , needs applying the naughty to fedora-34 as well (dosfstools-4.2-1.fc34.x86_64)
 - [x] `TestCurrentMetrics.testMemory`: fixed in PR #15454
 - [x]  check-storage-resize TestStorageResize.testResizeLuks: direct issue from https://bugzilla.redhat.com/show_bug.cgi?id=1934567, fixed locally
 - [x] fix watching binary channels: PR #15467

PR #15454, opened 2021-03-02
test: More robust memory hog

martinpitt:
The `VAR=hugestring` shell approach is rather brittle: bash seems to have
some kind of optimization or garbage collection that cleans up the
memory soon after allocation. In Fedora 34 this is so fast that it
breaks the test, but it eventually happens in other distros as well.
Also, the actual memory usage wiggles wildly between 600 and 0 MB for an
allocation of 200 MB.

Replace this with awk, which can also build arbitrarily long strings
(and thus memory usage). This is much faster, much more precise/stable,
and still works everywhere including Fedora CoreOS (which has neither
Python nor Perl).

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -467,8 +467,8 @@ class TestCurrentMetrics(MachineCase):
         # allocate a chunk of memory; this may cause other stuff to get unmapped,
         # thus not exact addition, but usage should go up
         size = 300 if have_swap else 200
-        m.execute("systemd-run --collect --slice cockpittest --unit mem-hog sh -ec "
-                  f"'MEMBLOB=$(yes | dd bs=1M count={size} iflag=fullblock); touch /tmp/hogged; sleep infinity'")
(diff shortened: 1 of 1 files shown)


martinpitt:
I explicitly tested that with fedora-34 in https://github.com/cockpit-project/cockpit/pull/15350 and https://github.com/cockpit-project/bots/pull/1719 and it works fine now. However, lots of other f-34 tests are still failing, so I'm not triggering it here.

martinpitt:
Affects downstream gating tests, release blocker.

martinpitt:
Still [fails on images with swap](https://logs.cockpit-project.org/logs/pull-15454-20210302-185237-9c18bf3a-fedora-32/log.html#320) -- I only tested on cloud-y images (I locally have a cloud based fedora-33). This was because the previous shell approach actually used *much* more than 300 MB of RAM. I adjusted the awk mem-hog2 to claim more memory, to ensure that it spills into swap.

marusak (approved):
Thanks!

PR #15455, opened 2021-03-02
test: Kerberized sudo might not need KRB5CCNAME in env_keep

Gundersanne:
According to the man pages (`man pam_sss_gss`), this depends on the sudo
version. It works in fedora-33 without the extra configuration of
env_keep at least.


martinpitt (approved):
Whee, nice! If that works and goes green, let's get it in. Dank je wel!

PR #15468, opened 2021-03-04
shell: Don't access undefined "source"

mvollmer:
The "source" variable might be undefined at this point in the code
when the window is closing.


martinpitt (approved):
Ooh, thanks! I've seen this failure a lot recently, nice to get this fixed again :+1:

PR #15470, opened 2021-03-04
Users fixes and restyling

### LINKED ISSUE #15195: accounts: Accessibility issues
- [ ] #accounts-list has children in its `<ul>` that are not `<li>` elements. Either this should change from a list item or we should remove the role="presentation" attribute from the `<li>` elements, which cause the element to be disregarded by the accessibility tree.
- [ ] The set password dialog's progress bar (used for password strength) has an invalid ARIA attribute called "aria-valuenow". Also, it is set to an empty string.

marusak:
newaccounts

Fixes #15195


KKoukiou (approved):

garrett:
I modified the screenshot above, re-cropping it and added a border.

mvollmer:
I don't think this warrants a release note.

PR #15474, opened 2021-03-04
SELinux restyling follow-ups

### LINKED ISSUE #15458: SELinux follow-ups
The SELinux restyling looks great overall!

I noticed there are a few small things, all of which can be seen in this screenshot:

Screenshot_2021-03-03 SELinux - garrett Rain

* [x] Too much space between toggle and its label (this may be a PF4 regression: https://github.com/patternfly/patternfly/issues/3880)
  kkoukiou: Indeed it's that bug, let's not track it here. It will be fixed automatically when we update.
* [x] Action to view automation script is not vertically aligned... it should be either vertically centered or baseline (and it's neither)
* [x] Action to view automation script should probably be a button; it should not look like a link, as it doesn't go somewhere, but shows a dialog
* [x] Responsive mode of the tables do not save any space here, but look ugly... as we can adjust the spacing, I'd suggest we don't collapse in this case (this probably applies even moreso to the bottom card's non-list "list")
* [x]  Empty state for the bottom card should not be in the list
* [ ]  Empty state in the bottom card should have an action to resolve, to install the package directly (although this may just be a consequence of looking at it on an rpm-ostree system, where that's not possible)

KKoukiou:
Fixes #15458

### DIFF
--- a/pkg/lib/cockpit-components-modifications.css
+++ b/pkg/lib/cockpit-components-modifications.css
@@ -15,3 +15,7 @@
 .green-icon {
     color: var(--pf-global--success-color--100);
 }
+
+.pf-c-card.modifications-table .pf-c-card__header {
+    justify-content: space-between;
+}
--- a/pkg/lib/cockpit-components-modifications.jsx
+++ b/pkg/lib/cockpit-components-modifications.jsx
@@ -20,7 +20,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import {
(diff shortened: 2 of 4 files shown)


martinpitt:
Looks much better now, thank you!

image

Note that I had to apply a pretty large hammer (`rpm -e --nodeps`) to uninstall setroubleshoot-server on Fedora, as cockpit-selinux depends on it.

martinpitt (approved):

PR #15475, opened 2021-03-04
machines: fix wrong usage of nested setState calls

KKoukiou:
This was resulting on console warnings:
Warning: An update (setState, replaceState, or forceUpdate) was scheduled from inside an update function

### DIFF
--- a/pkg/machines/components/vm/disks/diskAdd.jsx
+++ b/pkg/machines/components/vm/disks/diskAdd.jsx
@@ -410,46 +410,38 @@ export class AddDiskModalBody extends React.Component {
         case 'busType': {
             const existingTargets = Object.getOwnPropertyNames(this.props.vm.disks);
             const availableTarget = getNextAvailableTarget(existingTargets, value);
-            // use onValueChange instead of setState in order to perform subsequent state change logic
-            this.onValueChanged('target', availableTarget);
-            this.setState({ busType: value });
+            this.setState({ busType: value, target: availableTarget });
             break;
         }
         case 'file': {
-            stateDelta.file = value;
-            this.setState(prevState => {
-                if (value.endsWith(".iso")) {
-                    // use onValueChange instead of setState in order to perform subsequent state change logic
-                    this.onValueChanged("device", "cdrom");
-                }
-                return stateDelta;
-            });
+            if (value.endsWith(".iso")) {
+                // use onValueChange instead of setState in order to perform subsequent state change logic
+                this.onValueChanged("device", "cdrom");
+            }
+            this.setState({ file: value });
             break;
         }
         case 'device': {
-            stateDelta.device = value;
-            this.setState(prevState => {
-                let newBus;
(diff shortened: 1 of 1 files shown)


pkg/machines/components/vm/disks/diskAdd.jsx
```diff
@@ -439,16 +436,15 @@ export class AddDiskModalBody extends React.Component {
                 }
 
                 if (newBus)
-                    this.onValueChanged("busType", newBus);
+                    return ({ busType: newBus });
                 // Disk device "cdrom" and bus "virtio" are incompatible, see:
                 // https://listman.redhat.com/archives/libvir-list/2019-January/msg01104.html
                 else if (value === "cdrom" && prevState.busType === "virtio") {
                     // use onValueChange instead of setState in order to perform subsequent state change logic
```
skobyda, line 443: This comment here is no longer valid

pkg/machines/components/vm/disks/diskAdd.jsx
```diff
@@ -439,16 +436,15 @@ export class AddDiskModalBody extends React.Component {
                 }
 
                 if (newBus)
-                    this.onValueChanged("busType", newBus);
+                    return ({ busType: newBus });
```
skobyda, line 439: Leaving out the onValueChanged can cause the `target` property to not be correct. When busType is changed through `onValueChanged`, it also triggers a change of `target` (e.g. from vda to sda). If this is omitted, a disk with a devide type of "sata" can end up with a target "vdb" instead of "sdb", which would be wrong.

pkg/machines/components/vm/disks/diskAdd.jsx
```diff
@@ -439,16 +436,15 @@ export class AddDiskModalBody extends React.Component {
...
                 // Disk device "cdrom" and bus "virtio" are incompatible, see:
                 // https://listman.redhat.com/archives/libvir-list/2019-January/msg01104.html
                 else if (value === "cdrom" && prevState.busType === "virtio") {
                     // use onValueChange instead of setState in order to perform subsequent state change logic
                     // According to https://libvirt.org/formatdomain.html#hard-drives-floppy-disks-cdroms (section about 'target'),
                     // scsi is the default option for libvirt in this case too
-                    this.onValueChanged("busType", "scsi");
+                    return ({ busType: "scsi" });
```
skobyda, line 446: Same issue as above.

skobyda (dismissed):

pkg/machines/components/vm/disks/diskAdd.jsx
skobyda: Alternatively you can change `target` here together with busType.

KKoukiou: Got it! I will fix, thanks!

martinpitt (approved):
Thanks! Ignoring the stale fedora32/firefox test request, as we moved to f33.

PR #10968, opened 2019-01-14
storage: Always round dialog size slider input

mvollmer:
Otherwise we will very likely pass a non-integer number to the D-Bus
call, which will fail.

This affects the size slider for new partitions, and without this fix,
setting a size for a new partition with the slider will cause a
unhelpful error message.

PR #15477, opened 2021-03-05
storaged: ensure that the size is an integer when resizing volumes

### LINKED ISSUE #15409: Grow logical volume - Unexpected type 'double' in argument
Cockpit version: 237
OS: Fedora 32
Page: Storage/Volume group

Grow logical volume opens as:
image

I change the "8" to "9" and click "Grow":
image

Expected behaviour is that either decimals should be accepted, or it should populate the input field with an integer value to begin with (and have a more explanatory input validation error message instead of something that looks like a bug).

If it must be integer, probably TB is too big unit, it would feel too imprecise, and the dialog should instead select GB as unit when it's opened.

KKoukiou:
Fixes #15409

### DIFF
--- a/pkg/storaged/lvol-tabs.jsx
+++ b/pkg/storaged/lvol-tabs.jsx
@@ -50,11 +50,12 @@ function lvol_rename(lvol) {
     });
 }
 
-function lvol_and_fsys_resize(client, lvol, size, offline, passphrase) {
+function lvol_and_fsys_resize(client, lvol, targetSize, offline, passphrase) {
     var block, crypto, fsys;
     var crypto_overhead;
     var vdo;
     var orig_size = lvol.Size;
+    var size = parseInt(targetSize);
 
     block = client.lvols_block[lvol.path];
     if (!block)
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -542,7 +542,7 @@ export const SelectOne = (tag, title, options) => {
         render: (val, change, validated) => {
             return (
                 <div data-field={tag} data-field-type="select" data-value={val}>
-                    <FormSelect value={val} aria-label={tag}
+                    <FormSelect value={val} aria-label={title}
                                 validated={validated}
                                 onChange={change}>
                         { options.choices.map(c => <FormSelectOption value={c.value} isDisabled={c.disabled}
@@ -851,7 +851,7 @@ class SizeSliderElement extends React.Component {
     }
 
     render() {
-        const { val, max, round, onChange, tag } = this.props;
+        const { val, max, round, title, onChange } = this.props;
         const min = this.props.min || 0;
         const { unit } = this.state;
 
@@ -909,9 +909,10 @@ class SizeSliderElement extends React.Component {
             <div className="size-sliderx">
                 <StatelessSlider fraction={val / max} onChange={change_slider} />
                 <TextInputPF4 className="size-text"
-                       value={ val === "" || isNaN(val) ? val : cockpit.format_number(val / unit) }
-                       onChange={change_text} />
-                <FormSelect className="size-unit" value={unit} aria-label={tag} onChange={change_unit}>
+                              aria-label={title}
+                              value={ val === "" || isNaN(val) ? val : cockpit.format_number(val / unit) }
+                              onChange={change_text} />
+                <FormSelect className="size-unit" value={unit} aria-label={title} onChange={change_unit}>
                     { this.units.map(u => <FormSelectOption value={u.factor} key={u.name} label={u.name} />) }
                 </FormSelect>
             </div>
@@ -958,7 +959,7 @@ export const SizeSlider = (tag, title, options) => {
                                        max={all_options.max}
                                        min={all_options.min}
                                        round={all_options.round}
-                                       tag={tag}
+                                       title={title}
                                        onChange={change} />
                 </div>
             );


pkg/storaged/lvol-tabs.jsx
```diff
@@ -50,15 +50,16 @@ function lvol_rename(lvol) {
     });
 }
 
-function lvol_and_fsys_resize(client, lvol, size, offline, passphrase) {
+function lvol_and_fsys_resize(client, lvol, targetSize, offline, passphrase) {
     var block, crypto, fsys;
     var crypto_overhead;
     var vdo;
     var orig_size = lvol.Size;
+    var size = targetSize | 0;
```
martinpitt, line 58: That surely is the weirdest syntax for "convert this to an integer" that I have ever seen.. To me it looks like a "binary or with zero", which is in fact what it syntactically is.. it just seems to convert its arguments to ints as a side effect.

I would feel a little better if that would use `parseInt(targetSize)` instead, and handle the `null` and `undefined` cases separately (if we even need to handle that at all) -- e.g. `parseInt(targetSize || 0)`, or  just `parseInt(targetSize)` if we can already assume that we have a number.

KKoukiou: parseInt parses a string https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt  that's why I ended up using this syntax. But it will work as well. I will replace.

KKoukiou: It's a number, see `validate` function at SizeSlider

martinpitt: I see -- `Math.round()` then? That's probably also more correct than parseInt() or `| 0`, which always truncate. I.e. 3.9 would end up as 3, not 4.

KKoukiou: The size here is already in bytes and not the value we insert in the UI component. Rounding does not offer much.

martinpitt: I see - I was afraid it would be the "9.15" from the "Terabytes" input. Thanks!

martinpitt (approved):

pkg/storaged/lvol-tabs.jsx
```diff
@@ -50,11 +50,12 @@ function lvol_rename(lvol) {
     });
 }
 
-function lvol_and_fsys_resize(client, lvol, size, offline, passphrase) {
+function lvol_and_fsys_resize(client, lvol, targetSize, offline, passphrase) {
     var block, crypto, fsys;
     var crypto_overhead;
     var vdo;
     var orig_size = lvol.Size;
+    var size = parseInt(targetSize);
```
mvollmer, line 58: Do you know where the double comes from?  The size slider shouldn't produce any...

mvollmer: And, yeah, parseInt is a weird way to turn a double into an int, no?

KKoukiou: It's coming from the text input.

mvollmer (changes requested):
I would to block this until I had a closer look.  The size slider shouldn't be producing doubles, which is a bug that we have already fixed some time ago.

pkg/storaged/lvol-tabs.jsx
KKoukiou: The docs say parseInt takes a string as an argument, but passing double works as well because JS.
Tried:

parseInt(1234.5)
1234

mvollmer:
See #10968 for the original fix.

pkg/storaged/lvol-tabs.jsx
mvollmer: I see, then we should call Math.round() also in `change_text` of SizeSliderElement.

KKoukiou:
> See #10968 for the original fix.

Oh we probably need that round logic also for the textinput, now it applied only to the slider component.

pkg/storaged/lvol-tabs.jsx
mvollmer: And we need a test as well. :-)

mvollmer: > The docs say parseInt takes a string as an argument, but passing double works as well because JS.

Yes, it works, but it's still weird, imo.  Math.round() or Math.floor() is much cleaner.

KKoukiou: Right, just realized it as well.

Actually the user experience of the input field is pretty horrible. We change the input inline while the user types, therefore the user can't just use backspace and insert some new value when the input field is a double number at least.
I am even considering if we should disable that input component (make it readonly). Because we clearly cannot accept every value the user will insert.

KKoukiou:
One comment on this:

The text input should be readonly or removed. It's bad user experience if we try to round the input value while the user is inserting it. The slider should be enough by it self. However, the current slider is not usable with keyboard and lacking accessibility.

Let's just make the input readonly when we move this to PF4 slider. I tried to port it but it's still problematic, https://github.com/patternfly/patternfly-react/issues/5529

So closing this for now, and will fix it once Slider is available from PF4.

oskarb:
Well, I'm an outsider here but I'm confused by the overall approach.

There is #10968 " storage: Always round dialog size slider input #10968 ", because "Otherwise we will very likely pass a non-integer number to the D-Bus call, which will fail."

Why isn't it rounded closer to where it needs to be an integer, instead of handling it as part of a GUI component? (Note: I didn't check the actual code.) Though from other comments on that issue I get the feeling the slider always outputs _bytes_. If that is true, it would certainly make sense for the slider to always return an integer value.

But then that should really have nothing to do with the text input, because the text input doesn't operate on bytes. A fractional byte makes no sense, but 8.5 TB makes perfect sense. The text input value 8.5 should not be rounded to 8 or 9. It should be multiplied by the selected unit size to get bytes (and then that value should be rounded to whole bytes).

Alternatively, as I mentioned in #15409, if the current size is "8.15 TB", if it is decided that the text input value absolutely must be integer for reasons unknown to me, then it would not be sufficiently accurate (IMHO) to round this value to 8TB. Rounding e.g. 8.5TB to 9TB also feels too inaccurate.

So if it must be integer, then I think 8.15TB should instead be displayed as 8150 GB. E.g. values smaller than perhaps 10TB should be displayed as GB when the dialog opens, to allow for greater granularity. Though I still think that allowing decimal input would be nice.

I feel it would be less than ideal to make the text input readonly. In the general case, text input is more precise than sliders or similar, and it's good to have them when you know what value you want instead of having to fiddle with a slider.

martinpitt:
> I think text input is a worthwhile feature and I would be sad to see it go.

Big +1. Honestly I'd rather sacrifice the slider and keep the text input than the other way round.

KKoukiou:
@martinpitt @mvollmer The issue with the text input is that we can't just overwrite what the user tries to input, which is what we did till now.
We can add a warning under the field that the value will be rounded up onBlur, and then allow the user input whaterver.
We definitely can't do rounding and overwriting values while the user it typing which is what currently happens.

Currently for floating point inputs, the user can barely insert new value, or use backspace. This is a much worse user experience than using just the slider. (which has predifined allowd values)

martinpitt:
@KKoukiou : Undoubtedly we need to make the validation less aggressive. Of course the current behaviour is broken. I was just saying that I'd like to keep a (working) text input.

mvollmer:
> (and then that value should be rounded to whole bytes)

Yes, and this is missing.

PR #15478, opened 2021-03-05
Use manifest.json as the main build stamp for webpack

allisonkarlitskaya:
This branch gets rid of dist/*/stamp and uses the manifest for this purpose instead.

There are a bunch of other small cleanups.


martinpitt (approved):
Nice! Bots have the last word obviously, but this looks great.

martinpitt (changes requested):
Oops, tests just failed:

   make[2]: *** No rule to make target 'dist/base1/Makefile.deps', needed by 'distdir-am'.  Stop.

allisonkarlitskaya:
The fedora-33 fail is pretty interesting.  It's failing the `/web-server/local-address-only` case of `test-webserver` because this code expects that we ought *not* to be able to connect:

```
static void
test_address (TestCase *tc,
              gconstpointer data)
{
  gchar *resp = NULL;
  const TestFixture *fix = data;

  g_signal_connect (tc->web_server, "handle-resource", G_CALLBACK (on_shell_index_html), NULL);
  if (fix->local_only)
    {
      resp = perform_http_request (tc->localport, "GET /shell/index.html HTTP/1.0\r\nHost:test\r\n\r\n", NULL);
      cockpit_assert_strmatch (resp, "HTTP/* 200 *\r\n*");
      g_free (resp);
      resp = NULL;
    }
  else
    {
      /* If there is only one interface, then cockpit_web_server_new will get a NULL address and thus do listen on loopback */
      if (tc->hostport)
        assert_cannot_connect (tc->localport);              <==== here
    }
```

I guess that's a race that it's connecting to some old instance or something...

martinpitt (approved):
Nice, thank you!

PR #15479, opened 2021-03-05
test: Add support for pixel testing

mvollmer:
See https://github.com/cockpit-project/starter-kit/pull/433 for some actual action.

### DIFF
--- a/test/common/link-patterns.json
+++ b/test/common/link-patterns.json
@@ -1,0 +1,24 @@
+[
+    {
+        "label": "screenshot",
+        "pattern": "Wrote screenshot to ([A-Za-z0-9\\-\\.]+\\.png)$",
+        "url": "$1",
+        "icon": "bi bi-camera-fill"
+    },
+    {
+        "label": "new pixels",
+        "pattern": "New pixel test reference ([A-Za-z0-9\\-\\.]+\\.png)$",
+        "url": "$1"
+    },
+    {
+        "label": "journal",
+        "pattern": "Journal extracted to ([A-Za-z0-9\\-\\.]+\\.log(?:\\.[gx]z)?)$",
+        "url": "$1",
+        "icon": "bi bi-card-text"
+    },
+    {
+        "label": "changed pixels",
+        "pattern": "Differences in pixel test ([A-Za-z0-9\\-\\.]+)$",
+        "url": "pixeldiff.html#$1"
+    }
+]
--- a/test/common/pixel-tests
+++ b/test/common/pixel-tests
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+set -eu
+
+SUBDIR=test/reference
+REPO=pixel-test-reference
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/${REPO}"
+CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+PUSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
+
+message() {
+    [ "${V-}" != 0 ] || printf "  %-8s %s\n" "$1" "$2"
+}
+
+cmd_init() {
+    git submodule add -b empty "$CLONE_REMOTE" "$SUBDIR"
+}
+
+cmd_pull() {
+    git submodule update --init -- "$SUBDIR"
+}
+
+cmd_status() {
+    ( cd "$SUBDIR"
+      git rm --cached --quiet '*.png'
+      git add *.png
+      if git diff-index --name-status --cached --exit-code HEAD; then
+          echo No changes
+      fi
+    )
+}
+
+cmd_push() {
+    ( cd "$SUBDIR"
+      git rm --cached --quiet '*.png'
+      git add *.png
+      if ! git diff-index --name-status --cached --exit-code HEAD; then
+          git reset --soft empty
+          git commit --quiet -m "$(date)"
+          tag="sha-$(git rev-parse HEAD)"
+          git tag "$tag" HEAD
+          git push "$PUSH_REMOTE" "$tag"
+      else
+          echo No changes
+      fi
+    )
+    git add "$SUBDIR"
+}
+
+main() {
+    local cmd="${1-}"
+
+    if [ -z "${cmd}" ]; then
+        echo 'This command requires a subcommand: init pull status push'
+        exit 1
+    elif ! type -t "cmd_${cmd}" | grep -q function; then
+        echo "Unknown subcommand ${cmd}"
+        exit 1
+    fi
+
+    shift
+    [ "${V-0}" = 0 ] || set -x
+    "cmd_$cmd" "$@"
+}
+
+main "$@"
--- a/test/common/pixeldiff.html
+++ b/test/common/pixeldiff.html
@@ -1,0 +1,598 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Cockpit Integration Tests - Pixel diffs</title>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js" type="text/javascript"></script>
+        <style>
+/*
+MIT License
+
+Copyright (c) 2018 Diamant Haxhimusa
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+.pixelcompare-wrapper.pixelcompare-horizontal {
+  height: 100% !important;
+}
+.pixelcompare-horizontal .pixelcompare-handle:before, .pixelcompare-horizontal .pixelcompare-handle:after, .pixelcompare-vertical .pixelcompare-handle:before, .pixelcompare-vertical .pixelcompare-handle:after {
+  content: "";
+  display: block;
+  background: #fff;
+  position: absolute;
+  z-index: 30;
+}
+img.pixelcompare-before, img.pixelcompare-after {
+  height: 100%;
+  object-fit: cover;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -o-user-select: none;
+  user-select: none;
+}
+
+.pixelcompare-before-label, .pixelcompare-after-label, .pixelcompare-overlay {
+  position: absolute;
+  top: 0;
+  width: 100%;
+  height: 100%;
+}
+
+.pixelcompare-before-label, .pixelcompare-after-label, .pixelcompare-overlay {
+  transition-duration: 0.5s;
+}
+
+.pixelcompare-before-label, .pixelcompare-after-label {
+  transition-property: opacity;
+}
+
+.pixelcompare-horizontal .pixelcompare-before-label:before, .pixelcompare-horizontal .pixelcompare-after-label:before {
+  top: 50%;
+  margin-top: -19px;
+}
+
+.pixelcompare-vertical .pixelcompare-before-label:before, .pixelcompare-vertical .pixelcompare-after-label:before {
+  left: 50%;
+  margin-left: -45px;
+  text-align: center;
+  width: 90px;
+}
+
+.pixelcompare-left-arrow, .pixelcompare-right-arrow, .pixelcompare-up-arrow, .pixelcompare-down-arrow {
+  width: 0;
+  height: 0;
+  border: 6px inset transparent;
+  position: absolute;
+}
+
+.pixelcompare-left-arrow, .pixelcompare-right-arrow {
+  top: 50%;
+  margin-top: -6px;
+}
+
+.pixelcompare-up-arrow, .pixelcompare-down-arrow {
+  left: 50%;
+  margin-left: -6px;
+}
+
+.pixelcompare-container {
+  box-sizing: content-box;
+  z-index: 0;
+  height: 100%;
+  overflow: hidden;
+  position: relative;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -o-user-select: none;
+  user-select: none;
+}
+.pixelcompare-container img {
+  max-width: 100%;
+  position: absolute;
+  top: 0;
+  display: block;
+}
+/* .pixelcompare-container.active .pixelcompare-overlay,
+.pixelcompare-container.active :hover.pixelcompare-overlay {
+  background: transparent;
+} */
+.pixelcompare-container.active .pixelcompare-overlay .pixelcompare-before-label,
+.pixelcompare-container.active .pixelcompare-overlay .pixelcompare-after-label,
+.pixelcompare-container.active :hover.pixelcompare-overlay .pixelcompare-before-label,
+.pixelcompare-container.active :hover.pixelcompare-overlay .pixelcompare-after-label {
+  opacity: 0;
+}
+.pixelcompare-container * {
+  box-sizing: content-box;
+}
+
+.pixelcompare-before-label {
+  opacity: 0;
+}
+.pixelcompare-before-label:before {
+  content: attr(data-content);
+}
+
+.pixelcompare-after-label {
+  opacity: 0;
+}
+.pixelcompare-after-label:before {
+  content: attr(data-content);
+}
+
+.pixelcompare-horizontal .pixelcompare-before-label:before {
+  left: 10px;
+}
+
+.pixelcompare-horizontal .pixelcompare-after-label:before {
+  right: 10px;
+}
+
+.pixelcompare-vertical .pixelcompare-before-label:before {
+  top: 10px;
+}
+
+.pixelcompare-vertical .pixelcompare-after-label:before {
+  bottom: 10px;
+}
+
+.pixelcompare-overlay {
+  transition-property: background;
+  background: transparent;
+  z-index: 25;
+}
+.pixelcompare-overlay:hover,
+.pixelcompare-container.active .pixelcompare-overlay,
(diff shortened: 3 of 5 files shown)


test/common/pixeldiff.html
```diff
@@ -0,0 +1,599 @@
...
+        <script>
+$(function () {
+    var base = window.location.hash.replace("#", "");
+    $('#key').text(base);
+    $('#new').text(base + "-pixels.png").attr("href", base + "-pixels.png");
+    var img_now = $("<img>", { src: base + "-pixels.png", width: "100%" });
+    var img_ref = $("<img>", { src: base + "-reference.png", width: "100%" });
+    var img_delta = $("<img>", { src: base + "-delta.png", width: "100%" });
+    img_now.on("load", function () {
+        console.log(img_now[0].naturalHeight);
```
mvollmer, line 575: "Remove before flight"

mvollmer: Done.

mvollmer:
This is in use now with cockpit-machines, so we should probably merge this in its current state (or close to it), even though it is still very much under development. The "pixel-tests" tool in particular should only be used by well-trained professionals, under supervision of even better trained professionals.

test/common/link-patterns.json
```diff
@@ -0,0 +1,24 @@
+[
+    {
+        "label": "screenshot",
+        "pattern": "Wrote screenshots to ([A-Za-z0-9\\-\\.]+\\.png)$",
```
mvollmer, line 4: This must be "Wrote screenshot to"

martinpitt: This is not just cosmetical, right? It must match what testlib.py's snapshot() does, to actually recognize "normal" failure screenshots.

test/common/pixel-tests
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+set -eux
+
+SUBDIR=test/reference
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/pixel-test-reference"
+CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+PUSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
```
martinpitt, line 10: If you intend to use this from workflows, you'll have to push through https, with ~/.git-credentials. See e.g. po-refresh or publish-dist workflows how to set that up.

test/common/pixel-tests
```diff
@@ -0,0 +1,89 @@
...
+
+subgit() {
+    git -C "$SUBDIR" "$@"
+}
+
+setup() {
+    # Cases:
+    #
+    # - $SUBDIR doesn't exist in index => fail
+    # - $SUBDIR is no gitlink in index => fail
```
martinpitt, line 31: The code here in setup() does not even look at git, so these two comments should move to the place where the index is checked.

test/common/pixel-tests
```diff
@@ -0,0 +1,89 @@
...
+    # - $SUBDIR doesn't exist in fsys or is dir but $SUBDIR/.git doesn't exist => init repo
+    # - $SUBDIR is not git repo => fail
```
martinpitt, line 33: This is also not checked here.

test/common/pixel-tests
```diff
@@ -0,0 +1,89 @@
...
+    setup || exit 1
+
+    subgit status
+}
+
+cmd_push() {
+    setup || exit 1
+
+    subgit checkout --orphan temp
+    subgit commit -a -m "$(date)"
```
martinpitt, line 61: What does that actually commit? Nothing gets added or changed anywhere? 

Also, why the dance with the orphan checkout and such? It could just always init an empty repo, commit, and just push the tag, like in the [c-machines publish-dist repo](https://github.com/cockpit-project/cockpit-machines/blob/main/.github/workflows/publish-dist.yml#L33)

test/common/pixeldiff.html
```diff
@@ -0,0 +1,598 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Cockpit Integration Tests - Pixel diffs</title>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js" type="text/javascript"></script>
```
martinpitt, line 6: Let's eliminiate this in a follow-up. The operations here are simple enough to do them with the standard browser DOM API.

test/common/testlib.py
```diff
@@ -94,7 +99,7 @@ def attach(filename):
 
 
 class Browser:
-    def __init__(self, address, label, port=None):
+    def __init__(self, address, label, pixels_label, port=None):
```
martinpitt, line 102: It would be safe to let that default to None, AFAICS. That would disable pixel testing, right? The code below already relies on that anyway.

test/common/testlib.py
```diff
@@ -654,6 +661,63 @@ def snapshot(self, title, label=None):
...
+                        data_ref = ref.load()
+                        data_now = now.load()
+                        data_delta = delta.load()
+                        result = True
+                        width, height = ref.size
+                        for y in range(height):
+                            for x in range(width):
+                                if data_ref[x, y][3] == 255 and data_ref[x, y] != data_now[x, y]:
+                                    data_delta[x, y] = (255, 0, 0, 255)
+                                    result = False
```
martinpitt, line 701: Ugh.. hopefully our screenshots are small enough, this looks pretty horrible :-) (We use PIL anyway, doesn't that have such functionality? Or maybe imagemagick/graphicsmagick?)

martinpitt (changes requested):
A few minor nags, but nothing earth-shaking. If that is urgent (as c-machines already uses it), I'd be fine with landing this as-is, and do a followup. Thanks!

test/common/link-patterns.json
mvollmer: Correct.

mvollmer:
`needswork` for adding a feature to assert_pixels that allows it to ignore given selectors.

test/common/link-patterns.json
mvollmer: Fixed.

test/common/pixel-tests
mvollmer: Yes, this was way too naive.

test/common/pixeldiff.html
mvollmer: The whole pixeldiff.html file could be redone from scratch, tbh.

test/common/testlib.py
mvollmer: Yes, good point.

test/common/testlib.py
mvollmer: PIL can do some math, but I haven't taken the time yet to figure it all out.  This naive code needs a few seconds per image, which is very slow for what it does, but still acceptable in the big picture, I'd say.

mvollmer:
I have redone the pixel-tests tool to use genuine submodules. I'd expect this to be less confusing to other parts of our machinery, and to developers. It's still possible to ignore submodules and just use `pixel-tests`.

test/common/pixel-tests
mvollmer: Ok, I have made this even more complex now. The difference between this and publish-dist and node_modules is that test/references contains files that are directly worked on by the developer. So I didn't want to assume that we can just blow away the repository and start from scratch.

Also, I want `push` to be a no-op if nothing has actually changes, so we need to keep some history around. So I am not sure how much simpler this will get.

The one wart is the necessity to create the "temp" branch and the circus necessary to get rid of it again.  The pixel-test-reference repository has a branch called "empty", and we could use that as the parent of all commits... that would probably lead to simpler code.

test/common/pixel-tests
```diff
@@ -0,0 +1,65 @@
...
+      git rm --cached --quiet '*.png'
+      git add *.png
+      if ! git diff-index --name-status --cached --exit-code HEAD; then
+          git checkout --quiet --orphan temp
+          git commit --quiet -m "$(date)"
+          head=$(git rev-parse HEAD)
+          git -c advice.detachedHead=false checkout --quiet "$head"
+          git branch --quiet -d temp
+          git tag "sha-$head" HEAD
+          git push origin "sha-$head"
```
martinpitt, line 41: I still don't understand what's the deal with this -- can you please add a comment? If this is merely about having independent "branch-less" tags on the github repo? Then it should be enough to just push the tag, but not the branch (i.e. exactly what this git push already does). Or is it really important to *not* commit this to the local checkout/repo?

test/common/test-functions.js
```diff
@@ -259,3 +259,37 @@ function ph_wait_cond(cond, timeout) {
...
 }
+
+function currentFrameAbsolutePosition() {
+  let currentWindow = window;
+  let currentParentWindow;
+  let positions = [];
+  let rect;
+
+  while (currentWindow !== window.top) {
+    currentParentWindow = currentWindow.parent;
```
martinpitt, line 270: Would be nice to have a consistent 4-space indentation here.

test/common/testlib.py
```diff
@@ -654,6 +662,94 @@ def snapshot(self, title, label=None):
...
+                y1: int
+
+            def relative_clip(sel):
+                try:
+                    r = self.call_js_func('ph_element_clip', sel)
+                    return Rect(r['x'] - rect['x'],
+                                r['y'] - rect['y'],
+                                r['x'] - rect['x'] + r['width'],
+                                r['y'] - rect['y'] + r['height'])
+                except Error:
```
martinpitt, line 685: Should this log the exception as a warning at least, for debugging? This would usually fail if we specify an invalid CSS selector in `ignore`, which would then silently be, well, ignored.

test/common/testlib.py
```diff
@@ -654,6 +662,94 @@ def snapshot(self, title, label=None):
...
+                    return None
+
+            ignore_rects = list(map(relative_clip, ignore))
```
martinpitt, line 688: Doesn't need an explicit list() conversion, keeping this as iterator is fine.

test/common/testlib.py
```diff
@@ -654,6 +662,94 @@ def snapshot(self, title, label=None):
...
+            base = self.pixels_label + "-" + key
+            filename = base + "-pixels.png"
+            ref_filename = os.path.join(TEST_DIR, "reference", filename)
+            ret = self.cdp.invoke("Page.captureScreenshot", clip=rect, no_trace=True)
+            if "data" in ret:
```
martinpitt, line 693: Please turn that into an assertion. These all happen in "known good state" situations, they don't have to cope with unknown broken states like the handlers for failed tests. Id we add an invalid assert_pixels(), we want to fail the test, not silently ignore the error.

test/common/testlib.py
```diff
@@ -654,6 +662,94 @@ def snapshot(self, title, label=None):
...
+                        f.write(png_now)
+                    attach(filename)
+                    print("New pixel test reference " + filename)
+                    self.failed_pixel_tests += 1
+                else:
+                    img_now = Image.open(io.BytesIO(png_now)).convert("RGBA")
+                    img_ref = Image.open(io.BytesIO(png_ref)).convert("RGBA")
+                    img_delta = img_ref.copy()
+
+                    def ignoreable_coord(x, y):
```
martinpitt, line 707: Most imporant review comment of them all: Is it "ignorable" or "ignoreable"? Or "ignored" :grinning:

test/common/testlib.py
```diff
@@ -654,6 +662,94 @@ def snapshot(self, title, label=None):
...
+                        if ref.size != now.size:
+                            return False
+                        data_ref = ref.load()
+                        data_now = now.load()
+                        data_delta = delta.load()
+                        result = True
+                        width, height = ref.size
+                        for y in range(height):
+                            for x in range(width):
+                                if data_ref[x, y][3] == 255 and data_ref[x, y] != data_now[x, y]:
```
martinpitt, line 728: So this ignores everything with just a wee bit of alpha, right? Can you please add a comment about the high-level comparison strategy, so that we will still know a year from now?

martinpitt (changes requested):
Thanks! One more round, then this should be ready tomorrow.

test/common/pixel-tests
```diff
@@ -0,0 +1,65 @@
...
+
+SUBDIR=test/reference
+STORAGE_REPO=../pixel-test-reference
+
+message() {
+    [ "${V-}" != 0 ] || printf "  %-8s %s\n" "$1" "$2"
+}
+
+cmd_init() {
+    git submodule add -b empty "$STORAGE_REPO" test/reference
```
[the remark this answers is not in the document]
mvollmer: Oh, yes!

test/common/pixel-tests
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+set -eu
+
+SUBDIR=test/reference
+STORAGE_REPO=../pixel-test-reference
```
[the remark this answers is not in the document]
mvollmer: You are right. It's a clone url, relative to the clone url of the parent repo.  So if you check out the parent as `git@github.com:cockpit-project/cockpit-machines.git`, the storage repo will be `git@github.com:cockpit-project/pixel-test-reference`.

I'll add a comment.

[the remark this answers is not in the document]
mvollmer: Oh, I have to check the rules carefully... it might end up being relative to the a personal fork, not "origin"... hmm.

test/common/pixel-tests
```diff
@@ -0,0 +1,65 @@
...
+}
+
+cmd_pull() {
+    git submodule update --init -- "$SUBDIR"
+}
+
+cmd_status() {
+    ( cd "$SUBDIR"
+      git rm --cached --quiet '*.png'
+      git add *.png
```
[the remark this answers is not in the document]
mvollmer: Yeah, I had the same reservation... My current stance is that test/reference/ is special: it is defined to track exactly all the *.png files in the worktree, and the index is effectively not used for it.

`status` should be a preview of what `push` will do.  If there is a way to implement that without touching the index, that would be better.  I'll try  to figure something out.

test/common/pixel-tests
mvollmer: It's about having commits without parent.  But commits with the "empty"` branch as their parent are just as good, and might be much easier to produce.

test/common/test-functions.js
mvollmer: Yes, good point.  (This is obviously stackoverflowed...)

test/common/testlib.py
mvollmer: The only purpose of ignoring errors is to allow specifying a selector for elements that don't exist.  This helps with the specific cockpit-machines use of this. Sometimes there is a `#storage-size-helper`, sometimes not. But it's better to be explicit about which ones are expected to be there.
So I'll remove the error catching here and adapt cockpit-machines.

test/common/testlib.py
mvollmer: I had trouble without it... the `for` loop below would be empty, for some reason... I'll debug this.

test/common/testlib.py
mvollmer: Yes, good point.

test/common/testlib.py
mvollmer: it's "ignorable" :-)

test/common/testlib.py
mvollmer: Done.

test/common/pixel-tests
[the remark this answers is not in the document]
mvollmer: Ok, I have changed this back to use different URLs for fetching and pushing. Now pixel-tests uses "git submodule" for fetching, and explicitly pushes to a "git@github.com" url.

test/common/testlib.py
mvollmer: Done.

test/common/testlib.py
mvollmer: Done.

test/common/testlib.py
mvollmer: See https://github.com/cockpit-project/cockpit-machines/pull/152

test/common/pixel-tests
mvollmer: Done.

test/common/testlib.py
mvollmer: Yes, it's confirmed.  Apply this debugging instrumentation:

```
def ignorable_coord(x, y):
    print("IGN", x, y)
    for r in ignore_rects:
        print("on", r)
        if r and x >= r.x0 and x < r.x1 and y >= r.y0 and y < r.y1:
            return True
    return False
```

make a change to a reference image in a ingored part so that ignorable_coord is actually called, and you get this:

```
IGN 230 465
on Browser.assert_pixels.<locals>.Rect(x0=190, y0=459, x1=816, y1=480)
IGN 231 465
IGN 232 465
IGN 233 465
IGN 234 465
IGN 235 465
IGN 236 465
...
Differences in pixel test TestMachinesCreate-testCreateBasicValidation-auto
```

Thus, the first call to ignorable_coords works on the expected rectangle, and then the next calls don't do anything. It is as if the result of `map` can only be consumed once.

mvollmer: > It is as if the result of map can only be consumed once.

And that's exactly what's going on, and it is by design. Weird. *shrug*

martinpitt: Right, it's an iterator object. Makes sense then, thanks!

mvollmer:
"needswork" for not ignoring errors when working with the "ignore" param.

test/common/pixel-tests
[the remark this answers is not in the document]
martinpitt: Not urgent, this tool is not something that we would touch without being guided by some test failure/bot comment.

test/common/pixel-tests
martinpitt: This came back, so unresolving. This is *not* a blocker for landing this PR (as that is a bit urgent), but a workflow must be able to override this variable.

test/common/testlib.py
```diff
@@ -654,6 +662,112 @@ def snapshot(self, title, label=None):
             attach(filename)
             print("Wrote HTML dump to " + filename)
 
+    def assert_pixels(self, selector, key, ignore=[]):
+        """Compare the given element with its reference"""
+
+        if Image and self.pixels_label and self.cdp and self.cdp.valid:
+            rect = self.call_js_func('ph_element_clip', selector)
```
martinpitt, line 669: Nitpick: Invert the condition, return, and avoid the giant indentation of essentially the whole function.

test/common/testlib.py
```diff
@@ -654,6 +662,112 @@ def snapshot(self, title, label=None):
...
+                # following ways:
+                #
+                # - A pixel in the reference image might be
+                #   transparent.  These pixels are ignored.
+                #
+                # - The call to assert_pixels specifies a list of
+                #   rectangles (via CSS selectors).  Pixels within those
+                #   rectangles are ignored.
+                #
+                # - The RGB values of pixels need to differ by more than 2.
```
martinpitt, line 717: s/need to differ/must not differ/ -- this describes *ignored* (acceptable) changes.

martinpitt (changes requested):
Thanks! Just the "don't ignore exceptions" bit from the previous round and two trivial issues left.

test/common/pixel-tests
mvollmer: Yes, this very likely needs more changes.  $CLONE_REMOTE is only used to write the .gitmodules file, and $PUSH_REMOTE is used regardless of what is in .gitmodules...  Maybe we should just always push via https...

test/common/testlib.py
mvollmer: https://github.com/cockpit-project/cockpit-machines/pull/152 is merged, I have removed the exception catching.

test/common/testlib.py
mvollmer: Good point, done!

test/common/testlib.py
mvollmer: Good catch!

martinpitt (approved):
Cheers! packit still fails as this is not rebased, but that's ok. 'nuff bickering, let's land this and let evolution take its course.

Thanks, this is amazing!

Issue #14896, opened 2020-11-12
cockpit-tls gnutls_handshake failed: A TLS fatal alert has been received (with Chromium)


OlegBoul:
cockpit-tls [107658]: cockpit-tls gnutls_handshake failed: A TLS fatal alert has
when connecting

allisonkarlitskaya:
Can you please be more specific about what the problem is here?  Is that something in the journal?  Are you able to connect anyway?

OlegBoul:
yes, this message from 
journalctl -u cockpit

OlegBoul:
CentOS 8

OlegBoul:
and now the same text, but another number :
cockpit-tls [6846]

OlegBoul:
I think that is because of web server timeouts
when I use cockpit locally in my internal offce this message is rare enough,
but when I'm trying use cockpit remotely over VPN this alert almost continuous

allisonkarlitskaya:
Okay, but what's the problem?  Are you unable to connect?  Are you able to connect?  Is this just noise in the journal that's bothering you?

OlegBoul:
Yes, today I can't to connect remotely to me office over cockpit
and ought to get remoute session at office computer to use cockpit

how to encrese timeoutes for TLS handshakes?

OlegBoul:
it seems be like 
https://confluence.atlassian.com/bitbucketserverkb/error-gnutls_handshake-failed-a-tls-warning-alert-has-been-received-779171747.html

OlegBoul:
To increse timeoutes for cockpit TLS handshakes:
1) find cockpit.service (for me - /usr/lib/systemd/system/cockpit.service)
2) make drop-in for it: 
[Service]
ExecStart
for example:
cat > /etc/systemd/system/cockpit.service.d/CustomExecStart.conf
[Service]
ExecStart=
ExecStart=/usr/libexec/cockpit-tls --idle-timeout 180
----------------------
(90 is default for --idle-timeout, first  'ExecStart=' is because of RHEL bug to clear old value - https://bugzilla.redhat.com/show_bug.cgi?id=756787#c9)

But this doesn't solve a problen in daytime with huge Internet traffic::
Nov 14 19:43:40 localhost.localdomain cockpit-tls[4409]: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.

There are also in inet recomendations to update policycoreutils-python-utils and python3-policycoreutils
[...]

Issue #14704, opened 2020-10-06
warning: cockpit-ws: Failed to open certificate file /run/cockpit/tls/...: No such file or directory


RoyKas:
Cockpit version: 229
OS: Debian 10 (Buster)
Page: Any

This issue seems related to #13668. However, Cockpit is working fine. HTTPS is secured with a valid certificate.

But my syslog is filling-up with many of the below two lines:

>  cockpit-tls[8307]: cockpit-tls: gnutls_handshake failed: A TLS fatal alert has been received.
cockpit-ws[31121]: cockpit-ws: Failed to open certificate file /run/cockpit/tls/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855: No such file or directory

Folder /run/cockpit/tls is empty.

I've also found the following lines, which may be related to the issue:

> systemd[1]: cockpit-wsinstance-https@e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.service: Main process exited, code=killed, status=15/TERM
cockpit-session: pam_ssh_add: Failed adding some keys

Cockpit sessions typically end in:

>  cockpit-bridge[31193]: g_log_remove_handler: assertion 'handler_id > 0' failed

Either, my Cockpit installation is not working as intended, or unnecessary warnings are generated.

martinpitt:
>    cockpit-ws: Failed to open certificate file /run/cockpit/tls/...

Did you enable client certificate authentication on your system? I.e. anything in /etc/cockpit/cockpit.conf? I just did that, and I still cannot get that warning, though. I just checked the code, and it would only happen if your browser sends a client certificate for the cockpit page.

RoyKas:
There is no /etc/cockpit/cockpit.conf

Only configuration change I've made to the Debian Backport version (currently 237-1~bpo10+1) was to put a valid certificate in /etc/cockpit/ws-certs.d

martinpitt:
Nevermind, I'll fix that.

avnav88:
curious if this has been figured out...i'm having a similar issue:

```
cockpit-ws[989808]: cockpit-ws: Failed to open certificate file /run/cockpit/tls/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

martinpitt:
[...]

PR #15483, opened 2021-03-07
test: Check for unexpected info level journal messages

martinpitt:
service stdout/err output appear as journal priority "info" (6) instead
of "notice". This rendered a large portion of our
allow_journal_messages() inert (and explains why commit cf82ab8b132
"worked"), and more importantly, does not spot glib warnings that we
really want to see, such as in issue #14704 or #14896.

Allow cockpit-session's "last failed login" messages for every test for
now, they don't indicate a failure or unexpected condition in most cases
-- a lot of tests check logins which are supposed to fail.

Bring back a cleaner variant of commit cf82ab8b132 with a more specific
name `allow_failed_sudo_journal_messages()`, and use it in the tests
which explicitly test becoming administrator with unprivileged users.

Add specific expected messages to the remaining tests which produce
them.

### DIFF
--- a/src/ws/cockpitwsinstancecert.c
+++ b/src/ws/cockpitwsinstancecert.c
@@ -34,6 +34,8 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include "../tls/utils.h"
+
 #define CGROUP_REGEX         "^(0:|1:name=systemd):/system.slice/system-cockpithttps.slice/" \
                              "cockpit-wsinstance-https@([0-9a-f]{64}).service$"
 #define CGROUP_REGEX_FLAGS   (REG_EXTENDED | REG_NEWLINE)
@@ -132,6 +134,8 @@ https_instance_has_certificate_file (char   *contents,
   ssize_t r;
 
   if (https_instance == NULL) /* already warned */
+    goto out;
+  if (strcmp (https_instance, SHA256_NIL) == 0)
     goto out;
 
   dirfd = open ("/run/cockpit/tls", O_PATH | O_DIRECTORY | O_NOFOLLOW);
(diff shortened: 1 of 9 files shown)


martinpitt:
The [first run](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15483-20210307-195315-43a4da1a-fedora-33/log.html) has 25 failures, but they are all very similar.

@mvollmer: this also explains why commit cf82ab8b1322 seemingly "just worked" -- it was because collecting these messages does not work any more in the first place. Let's fix that!

martinpitt:
There is at least one more failure (`Test{AD,IPA}.testClientCertAuthentication` which I fixed locally, but I'll wait for this test run before pushing.

martinpitt:
@allisonkarlitskaya : Most of this is probably not your alley, but can you please have a look at the "ws: Don't warn about missing cert file for cert-less instance" commit?

src/ws/cockpitwsinstancecert.c
```diff
@@ -133,6 +135,8 @@ https_instance_has_certificate_file (char   *contents,
 
   if (https_instance == NULL) /* already warned */
     goto out;
+  if (strcmp (https_instance, SHA256_NIL) == 0)
```
allisonkarlitskaya, line 138: This is fine.

test/verify/check-system-journal
```diff
@@ -504,6 +504,8 @@ ExecStart=/bin/sh -c 'sleep 5; for s in $(seq 10); do echo SLOW; sleep 0.1; done
                         ["check-journal", "BEFORE BOOT", ""],
                         "Load earlier entries"
                         ])
+        # HACK: Figure out why this happens
+        self.allow_journal_messages(".*Failed to add match '00:03': Invalid argument.*")
```
marusak, line 508: ehm what? Is this happening on all OSes or just at some?

marusak (approved):
Thanks!
A few HACKs that should be looked at, but surely an improvement and those missing pieces seem like details, so no need to block this anymore on it.

martinpitt:
@marusak: Right, that's why I kept the hacks as separate commits, easier to revert like that.

test/verify/check-system-journal
marusak: Interesting. I am happy to take a look at this as I introduced this.

PR #15486, opened 2021-03-09
machines: add support for passing cloud-init metadata when creating VMs with cloud images

KKoukiou:
### Machines: support authentication against cloud images

Distro cloud images have locked login accounts by default. The user can now specify the root and user account login credentials when creating VMs based on cloud images. The cloud image itself will be used as the backing store for the VM's disk.

Screen Shot 2021-03-17 at 12 03 18

### DIFF
--- a/pkg/machines/components/create-vm-dialog/createVmDialog.jsx
+++ b/pkg/machines/components/create-vm-dialog/createVmDialog.jsx
@@ -298,9 +298,9 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, 
                     {cloudInitSupported ? <FormSelectOption value={CLOUD_IMAGE}
                                       label={_("Cloud base image")} /> : null}
                     <FormSelectOption value={LOCAL_INSTALL_MEDIA_SOURCE}
-                                      label={_("Local install media")} />
+                                      label={_("Local install media (ISO image or distro install tree)")} />
                     <FormSelectOption value={URL_SOURCE}
-                                      label={_("URL")} />
+                                      label={_("URL (ISO image or distro install tree)")} />
                     {connectionName == 'system' &&
                     <FormSelectOption value={PXE_SOURCE}
                                       label={_("Network boot (PXE)")} />}
--- a/pkg/machines/components/create-vm-dialog/createVmDialog.jsx
+++ b/pkg/machines/components/create-vm-dialog/createVmDialog.jsx
@@ -69,6 +69,7 @@ const _ = cockpit.gettext;
 
 const URL_SOURCE = 'url';
 const LOCAL_INSTALL_MEDIA_SOURCE = 'file';
+const CLOUD_IMAGE = 'cloud';
 const DOWNLOAD_AN_OS = 'os';
 const EXISTING_DISK_IMAGE_SOURCE = 'disk_image';
 const PXE_SOURCE = 'pxe';
@@ -145,6 +146,7 @@ function validateParams(vmParams) {
         case PXE_SOURCE:
             break;
         case LOCAL_INSTALL_MEDIA_SOURCE:
+        case CLOUD_IMAGE:
         case EXISTING_DISK_IMAGE_SOURCE:
             if (!vmParams.source.startsWith("/")) {
                 validationFailed.source = _("Invalid filename");
@@ -206,7 +208,7 @@ const NameRow = ({ vmName, onValueChanged, validationFailed }) => {
     );
 };
 
-const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, os, osInfoList, downloadOSSupported, onValueChanged, validationFailed }) => {
+const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, os, osInfoList, cloudInitSupported, downloadOSSupported, onValueChanged, validationFailed }) => {
     let installationSource;
     let installationSourceId;
     let installationSourceWarning;
@@ -218,6 +220,15 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, 
         installationSource = (
             <FileAutoComplete id={installationSourceId}
                 placeholder={_("Path to ISO file on host's file system")}
+                onChange={value => onValueChanged('source', value)}
+                superuser="try" />
+        );
+        break;
+    case CLOUD_IMAGE:
+        installationSourceId = "source-file";
+        installationSource = (
+            <FileAutoComplete id={installationSourceId}
+                placeholder={_("Path to cloud image file on host's file system")}
                 onChange={value => onValueChanged('source', value)}
                 superuser="try" />
         );
@@ -284,6 +295,8 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices, 
                     {downloadOSSupported
                         ? <FormSelectOption value={DOWNLOAD_AN_OS}
                                             label={_("Download an OS")} /> : null}
+                    {cloudInitSupported ? <FormSelectOption value={CLOUD_IMAGE}
+                                      label={_("Cloud base image")} /> : null}
                     <FormSelectOption value={LOCAL_INSTALL_MEDIA_SOURCE}
                                       label={_("Local install media")} />
                     <FormSelectOption value={URL_SOURCE}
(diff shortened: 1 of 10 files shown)


pkg/machines/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -529,6 +542,107 @@ const UnattendedRow = ({
     );
 };
 
+const CloudInitOptionsRow = ({
```
martinpitt: This copies a lot of code from UnattendedRow. Could you please generalize/parameterize the two into one component?

pkg/machines/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -529,6 +542,107 @@ const UnattendedRow = ({
...
+
+    const cloudInitCheckbox = (
+        <Checkbox id="cloud-init-checkbox"
+                  isChecked={useCloudInit}
+                  onChange={checked => onValueChanged('useCloudInit', checked)}
+                  label={_("Cloud init parameters")} />
+    );
+
+    return (
+        <FormGroup fieldId="unattended-installation">
```
martinpitt: I can't guess where you copied that from :grin:

pkg/machines/scripts/create_machine.sh
```diff
@@ -63,6 +65,32 @@ else
     UNATTENDED_PARAMS=""
 fi
 
+if [ "$USE_CLOUD_INIT" = "true" ]; then
+    if [ -z "$USER_PASSWORD" ]; then
+        USER_LINE=""
+        CREATE_USER_LINE=""
+    else
+        USER_LINE=$'\n'"    $USER_LOGIN:$USER_PASSWORD"
```
martinpitt, line 73: The `\n` here and writing `$ROOT_LINE$USER_LINE` is error prone. I assume you do this to avoid an empty line in the list, but if you have no root password, but an user one, the result will be
```
list: |
    
    user:pwd
```
anyway, so you may as well keep it simple? Or, if the userdata really doesn't get along with empty lines, fix it harder.

pkg/machines/components/create-vm-dialog/cloudInitOptionsRow.jsx
```diff
@@ -0,0 +1,131 @@
+/*
```
martinpitt, line 1: Please squash the "move cloud init related code to seperate file" commit (also, this should look different after factorizing)

martinpitt (changes requested):
Just a quick code review for now, I have not visually tested this yet. Thank you!

pkg/machines/scripts/create_machine.sh
KKoukiou: The reason I did it like this is indeed because cloud-init user-data file seems to be very sensitive to spaces and new lines.
Something like this does not work. 

```
#cloud-config
users:
  - name: test
chpasswd:
  list: |
    
    test:foobar
  expire: False
```
 
Thanks for catching this, I will repost with the fix.

KKoukiou: So we can't have new lines inside the list, but new lines between fields are fine. I tested the solution I am about to post with all scenarios. If you know some better templating that I can do through the script I am happy to improve it.

pkg/machines/components/create-vm-dialog/cloudInitOptionsRow.jsx
KKoukiou: I decided to do any code move in follow up, no need to be part of this PR.

pkg/machines/scripts/create_machine.sh
```diff
@@ -63,6 +65,32 @@ else
     UNATTENDED_PARAMS=""
 fi
 
+if [ "$USE_CLOUD_INIT" = "true" ]; then
+    if [ -z "$USER_PASSWORD" ]; then
+        USER_LINE=""
+        CREATE_USER_LINE=""
+    else
+        USER_LINE="    $USER_LOGIN:$USER_PASSWORD"
+        CREATE_USER_LINE="users:"$'\n'"  - name: $USER_LOGIN"
```
martinpitt, line 74: I'm actually surprised that this `'\n'` thingy works -- shell is usually rather picky/broken with nested spacing. But so much the better!

FTR, it'd be cleaner to build it piece-wise, like:
```sh
echo '#cloud-config' >> $USER_DATA_FILE
if [ -n "$VAR1" ]; then
    echo "    user=$VAR1" >> $USER_DATA_FILE
```
and so on. But not necessary to change it now.

martinpitt (changes requested):
Thanks for the refactor! There is still a test failure. Also, this doesn't seem to want to work for me in interactive testing, I'll send a separate comment.

martinpitt:
I built a VM with this branch, ran `bots/vm-run  -s cockpit.socket fedora-33`, and inside that, tried to create a cloud-init VM.

First, in the VM I got the current [debian 10 cloud image](https://cloud.debian.org/images/cloud/buster/daily/):
```
cd /tmp/
curl -O https://cloud.debian.org/images/cloud/buster/daily/20210312-574/debian-10-generic-amd64-daily-20210312-574.qcow2
```

Then I wanted to run it:

image


But the dialog has sat there for several minutes now. "Erstellen" (Create) is grayed out, and it's stuck. `virsh list --all` is empty, there is no process running, ~admin/.config/libvirt/qemu/ has nothing interesting, ~admin/.local/ does not even exist (so no local libvirt stuff).

I can cancel the dialog just fine, though. I wanted to try again on system libvirt, but a few seconds after closing the dialog somehow the machine completely locked up (ssh and cockpit). I came back for a few seconds, then I saw this in `top`:

```
     97 root      20   0       0      0      0 R  52.0   0.0   0:46.42 kswapd0                                                                                     
   2556 root      20   0  341828  13876   1588 R   6.3   1.2   0:00.46 python3                                                                                     
   2557 admin     20   0  362600  48316   2308 R   5.4   4.3   0:00.45 python3                                                                                     
   2596 admin     20   0  269852  13096    464 D   3.6   1.2   0:00.14 virt-install                                                                                
   2599 admin     20   0  269852  13228    588 D   3.6   1.2   0:00.14 virt-install                                                                                
   2601 admin     20   0  269856  13384    744 D   3.6   1.2   0:00.14 virt-install                                                                                
   2597 admin     20   0  269856  13284    640 R   3.1   1.2   0:00.13 virt-install                                                                                
   2598 admin     20   0  269856  13272    632 D   3.1   1.2   0:00.13 virt-install                                                                                
   2600 admin     20   0  269856  13104    464 D   3.1   1.2   0:00.13 virt-install 
```
Aha! So apparently it tried to do *something*, but as it ran out of RAM, it just went into some weird OOM crash loop. So I re-ran the VM with `-M 4096`. I also dialed down the disk to 1 GB only.

But same result -- the dialog is hanging. Nothing in `top`, not even after cancelling the dialog. But reloading the page suddenly makes several virt-installs appear and quickly disappear again.

Journal isn't particularly helpful:
```
Mar 12 13:24:35 m1.cockpit.lan audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=dbus-:1.4-org.libvirt@0 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Mar 12 13:27:27 m1.cockpit.lan libvirtd[2297]: libvirt version: 6.6.0, package: 5.fc33 (Fedora Project, 2020-12-07-17:58:21, )
Mar 12 13:27:27 m1.cockpit.lan libvirtd[2297]: hostname: m1.cockpit.lan
Mar 12 13:27:27 m1.cockpit.lan libvirtd[2297]: internal error: Unable to get DBus session bus connection: Unable to autolaunch a dbus-daemon without a $DISPLAY for X11
```
(not sure if that dbus-daemon message is fatal?)

I tried the same with qemu:///system, same result.

martinpitt:
To make sure I'm not going completely crazy, I tried to import it instead:

image

This failed at first:
image

Argh! Need to chown the downloaded qcow to admin (I downloaded it as root). And then, voilà:

image

With that,  I tried the cloud-init way again, but still the same error. It seems it didn't even get to the point where it stumbled over the "permission denied".

martinpitt:
Debugged on IRC -- I resized my test VM to be bigger (`bots/image-customize --resize 30G fedora-33`), and now the cloud-init works. I can log in as admin:foobar. I also tried to set a root password " foo bar" (with strategically placed spaces), and that worked also.

Thanks!

KKoukiou:
@martinpitt The difference bettween last revision and this, is that when I split the user credentials component, I forgot to conditionally display the plain user's fields. These are actually hidden when performing unattended installs with virt-install <= 3.0.0 because of some missing options. 
(Thus I broke tests on RHEL-8 and Centos 8 stream, virt-install there is still 2.1).

martinpitt (dismissed):
Thanks! LGTM once tests agree.

martinpitt (dismissed):

pkg/machines/components/create-vm-dialog/createVmDialog.jsx
```diff
@@ -309,9 +309,9 @@ const SourceRow = ({ connectionName, source, sourceType, networks, nodeDevices,
                     {cloudInitSupported ? <FormSelectOption value={CLOUD_IMAGE}
                                       label={_("Cloud base image")} /> : null}
                     <FormSelectOption value={LOCAL_INSTALL_MEDIA_SOURCE}
-                                      label={_("Local install media")} />
+                                      label={_("Local install media (ISO image or distro install tree)")} />
                     <FormSelectOption value={URL_SOURCE}
-                                      label={_("URL")} />
+                                      label={_("URL (ISO image or distro install tree)")} />
```
garrett, line 314: What's a distro install tree?

Is it an archive file with other files inside? Like a .tar / .tgz / or whatever.

Or is it a directory with files?

Or something else?

Will our users know what this means?

garrett:
Which screenshots are intended for the release notes?

It would be great if we could re-take them once #15515 lands, so they'd be current to the release.

pkg/machines/components/create-vm-dialog/createVmDialog.jsx
KKoukiou: It's a directory, https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/s1-kickstart2-install-tree

In case they don't know I guess they can google it? It's better than just having that value 'URL' string which does not contain much info.

KKoukiou:
> Which screenshots are intended for the release notes?
> 
> It would be great if we could re-take them once #15515 lands, so they'd be current to the release.

I have attached the screenshots in the first comment, plus the text for the release notes. I will retake them once the slider removal PR lands.

martinpitt (approved):
Delta review: The new test robustification looks good to me. For covering our bits it's enough to test the React UI → libvirt XML portion, and we can be a bit sloppy on the XML → runtime QEMU bits if it becomes too inconvenient.

garrett:
Note: Still needs a new screenshot, as this one has sliders and it's not what it will look like in the release.

martinpitt (approved):

PR #15488, opened 2021-03-09
logs: Fixup services loading

marusak:
The current approach had a bunch of issues:
- It was doing `sh -ec` instead of using `cockpit.script()`
- It was using `then()` and `done()` at the same time
- It would never fail (due to those `grep` and `sort`)
- It was doing grep which is ugly
- The `sort` was not necessary as we sort it anyway
- The `host` option was unused

But mainly it was not error prone since we introduced possible spaces in
qualifiers. We could now have `journalctl --since="2021-03-03 08:00"`
which would be then parsed in command as `["journalctl", "since=2021-03-03 08:00"]`.
That is fine because separate arguments are treated as such, but when
we did `join(" ")` it would generate `journalctl --since=2021-03-03 08:00` where `08:00`
would not be recognized anymore. This lead to not showing any services
on the logs page.

### DIFF
--- a/pkg/systemd/logs.js
+++ b/pkg/systemd/logs.js
@@ -175,17 +175,26 @@ $(function() {
         function load_service_filters(match, options) {
             loading_services = true;
             current_services = new Set();
+            // Ideally this would use `--output cat --output-fields SYSLOG_IDENTIFIER` and do
+            // without `sh -ec`, grep, sort, replaceAll and all of those ugly stuff
+            // For that we however need newer systemd that includes https://github.com/systemd/systemd/issues/13937
             const service_options = Object.assign({ output: "verbose" }, options);
-            let cmd = journal.build_cmd(match, service_options)[0].join(" ");
-            cmd += " | grep SYSLOG_IDENTIFIER= | sort -u";
-            cockpit.spawn(["sh", "-ec", cmd], { host: options.host, superuser: "try" })
+            let cmd = journal.build_cmd(match, service_options)[0];
+            cmd = cmd.map(i => i.replaceAll(" ", "\\ ")).join(" ");
+            cmd = "set -o pipefail; " + cmd + " | grep SYSLOG_IDENTIFIER= | sort -u";
+            cockpit.spawn(["/bin/bash", "-ec", cmd], { superuser: "try", err: "message" })
                     .then(function(entries) {
                         entries.split("\n").forEach(function(entry) {
                             if (entry)
                                 current_services.add(entry.substr(entry.indexOf('=') + 1));
                         });
                     })
-                    .done(function () {
+                    .catch(e => {
+                        // grep returns `1` when nothing to match, but in that case message is empty
+                        if (e.message)
+                            console.log("Failed to load services:", e.message);
+                    })
+                    .finally(() => {
                         loading_services = false;
                         show_service_filters();
                     });
--- a/test/verify/check-system-journal
+++ b/test/verify/check-system-journal
@@ -504,8 +504,8 @@ ExecStart=/bin/sh -c 'sleep 5; for s in $(seq 10); do echo SLOW; sleep 0.1; done
                         ["check-journal", "BEFORE BOOT", ""],
                         "Load earlier entries"
                         ])
-        # HACK: Figure out why this happens
-        self.allow_journal_messages(".*Failed to add match '00:03': Invalid argument.*")
+
+        b._wait_present("#journal-service-menu option[value='check-journal']")
 
         # New messages are not streamed when 'until' is in the past
         m.execute("logger -p user.err --tag check-journal RIGHT NOW")


pkg/systemd/logs.js
```diff
@@ -175,17 +175,18 @@ $(function() {
...
+            cockpit.spawn(cmd, { superuser: "try", err: "message" })
+                    .then(entries => {
+                        entries.split("\n").forEach(entry => {
                             if (entry)
-                                current_services.add(entry.substr(entry.indexOf('=') + 1));
+                                current_services.add(entry);
                         });
                     })
-                    .done(function () {
+                    .catch(e => console.log("Failed to load services:", e))
```
martinpitt, line 188: Not that important, but should it be `console.warn()`? I'm not sure if our tests would then fail if the warning is hit, or if we only do that on console.error(). I suppose that depends on if we actually expect this to happen sometimes, e.g. on invalid user input?

pkg/systemd/logs.js
```diff
@@ -175,17 +175,18 @@ $(function() {
...
             current_services = new Set();
-            const service_options = Object.assign({ output: "verbose" }, options);
-            let cmd = journal.build_cmd(match, service_options)[0].join(" ");
-            cmd += " | grep SYSLOG_IDENTIFIER= | sort -u";
-            cockpit.spawn(["sh", "-ec", cmd], { host: options.host, superuser: "try" })
-                    .then(function(entries) {
-                        entries.split("\n").forEach(function(entry) {
+            const service_options = Object.assign({ output: "cat" }, options);
+            const cmd = journal.build_cmd(match, service_options)[0];
+            cmd.splice(1, 0, "--output-fields=SYSLOG_IDENTIFIER");
```
martinpitt, line 180: This is still a bit ugly -- how about adding that as a recognized `options` to `build_cmd()` itself? That already does a lot of similar things, e.g. for `!options.count` → `--no-tail`.

martinpitt (changes requested):
Thanks for figuring this out! Note that the commit message says "not error prone", but it really was :grin:

pkg/systemd/logs.js
marusak: Yeah, I was thinking about it, but I though it was too specific usecase to introduced it into `journal`. But sure, I can add it, it surely won't hurt anything and will be a bit cleaner.

pkg/systemd/logs.js
marusak: > I suppose that depends on if we actually expect this to happen sometimes, e.g. on invalid user input?

It can indeed happen on invalid input, like incorrect timestamp: `since:2021--23` will log `Failed to parse timestamp: 2021--23`

marusak: I'll give it warning as users may actually notice they are doing something wrong

pkg/systemd/logs.js
martinpitt: You can make it a bit more general, like passing arbitrary value for `output-fields` or so? then again, easy to adjust once we actually need it

marusak:
well, this works only with newer systemd, from man pages for `--option cat` I see on older systemd:
```
generates a very terse output, only showing the actual message of each journal entry with no metadata, not even a timestamp.
```
while on newer ones:
```
generates a very terse output, only showing the actual message of each journal entry with no metadata, not even a timestamp. If combined with the --output-fields= option
               will output the listed fields for each log record, instead of the message.

```

Introduced in https://github.com/systemd/systemd/pull/15648

marusak:
`-o export` does not work due to https://github.com/systemd/systemd/issues/13937

marusak:
damn, this one is just too annoying. There seems not to be any solution, that would work everywhere.
The only robust approach (I can think of) is presented here. But that needs newer systemd, which would mean we would break at least ubuntu/debian.

I guess I'll try what we had before and just wrap arguments into `"` and if it contains `"` then I'll backslash those. Ugly, but should work.

marusak:
`warning: --: 1: set: Illegal option -o pipefail`
ubuntu and debian, hmm?

martinpitt:
@marusak: pipefail is a bashism, and Debian uses /bin/dash as default shell (/bin/sh). So use bash explicitly with `cockpit.spawn(["/bin/bash", "-ec", "...script"])` or restructure the code.

pkg/systemd/logs.js
```diff
@@ -175,17 +175,26 @@ $(function() {
...
+            // without `sh -ec`, grep, sort, replaceAll and all of those ugly stuff
+            // For that we however need newer systemd that includes https://github.com/systemd/systemd/issues/13937
             const service_options = Object.assign({ output: "verbose" }, options);
-            let cmd = journal.build_cmd(match, service_options)[0].join(" ");
-            cmd += " | grep SYSLOG_IDENTIFIER= | sort -u";
-            cockpit.spawn(["sh", "-ec", cmd], { host: options.host, superuser: "try" })
+            let cmd = journal.build_cmd(match, service_options)[0];
+            cmd = cmd.map(i => i.replaceAll(" ", "\\ ")).join(" ");
+            cmd = "set -o pipefail; " + cmd + " | grep SYSLOG_IDENTIFIER= | sort -u";
+            cockpit.spawn(["/bin/bash", "-ec", cmd], { superuser: "try", err: "message" })
```
martinpitt, line 185: Not important, but you can do this a bit more elegantly: `bash -o pipefail -ec ...`.

martinpitt (approved):
Thanks for figuring this out!

PR #15493, opened 2021-03-10
workflows: Upload built dist tarballs to -dist repository

martinpitt:
From there they can be downloaded without a GitHub token, and thus
become accessible to packit and other CI systems. Do this in a new
publish-dist workflow, as build-dist has no access to secrets and we
need our GitHub token for this.

As git's tarball names are not predictable, wrap them into a
`<sha>.tar` similar to the GitHub artifact.

Adjust make_dist.py to download the dist tarball from the -dist GitHub
repo. This gets rid of the bots API and token requirement.

----

This  is preparation for [running PackIt tests in upstream PRs](https://github.com/martinpitt/cockpit/pull/12). As this introduces a new `on_workflow:` workflow, this needs to be on master before it will actually run.

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -1,0 +1,50 @@
+name: publish-dist
+on:
+  workflow_run:
+    workflows: build-dist
+    types: [completed]
+
+jobs:
+  run:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download build-dist artifacts
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: build-dist
+          run_id: ${{ github.event.workflow_run.id }}
+
+      - name: Set up configuration and secrets
+        run: |
+          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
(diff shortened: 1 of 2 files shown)


martinpitt:
I tested this quite extensively on my fork in https://github.com/martinpitt/cockpit/pull/12 . I created https://github.com/martinpitt/cockpit-dist/commits/master where you can see the commits, and I also temporarily changed the "> 7 days old" to "> 4 hours old" to test the [drop old builds](https://github.com/martinpitt/cockpit-dist/commit/458eee9a) functionality.

This does *not* yet prune the git history. I [verified that this works](https://github.com/martinpitt/cockpit/pull/12#issuecomment-793977488), but I'd first like to get a bunch of real data and see if this approach works in general. If it's too slow after a week of tarballs (I don't want the build-dist workflow spend more than 1 minute on this, ideally much less), I'll revert back to [uploading dist tarballs to our image servers](https://github.com/martinpitt/cockpit/commit/58b7e5c070a2a580c554ffa067febf5e2755cdda). The latter is structurally much easier and race free, but needs our own infra.

martinpitt:
Ugh, interesting -- the build-dist workflow failed here because this is running the build_dist workflow as `pull_request`, and hence does not have access to the project secrets. I'll need to structure that differently.

marusak (approved):
Cool stuff!

PR #15496, opened 2021-03-10
selinux: remove last usages of PF3 and stop including it

KKoukiou:
This commit made the following changes:
* Replace icons with the React components from react-icons
* Replace `<pre>` with <TextArea>.` <pre> `in PF4 does not have a grey
  background
* Replace custom expandable section with ExpandableSection PF4
  component
* Replace list-group and list-group-items which are used for displaying
  solutions with Cards

### DIFF
--- a/pkg/lib/cockpit-components-modifications.jsx
+++ b/pkg/lib/cockpit-components-modifications.jsx
@@ -22,8 +22,9 @@ import React from 'react';
 import {
     Button, Card, CardHeader, CardTitle, CardBody,
     DataList, DataListItem, DataListItemRow, DataListCell, DataListItemCells,
-    Modal, Tabs, Tab, Text, TextVariants
+    Modal, Tabs, Tab, Text, TextArea, TextVariants
 } from '@patternfly/react-core';
+import { CheckIcon, CopyIcon, ExternalLinkAltIcon, OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 
 import cockpit from "cockpit";
 import './listing.scss';
@@ -71,9 +72,8 @@ class ModificationsExportDialog extends React.Component {
     render() {
         const footer = (
             <>
-                <Button variant='secondary' onClick={this.copyToClipboard}>
-                    { this.state.copied ? <span className="fa fa-check fa-xs green-icon" /> : <span className="fa fa-clipboard fa-xs" /> }
-                    <span>{ _("Copy to clipboard") }</span>
+                <Button variant='secondary' className="btn-clipboard" onClick={this.copyToClipboard} icon={this.state.copied ? <CheckIcon className="green-icon" /> : <CopyIcon />}>
+                    { _("Copy to clipboard") }
                 </Button>
                 <Button variant='secondary' className='btn-cancel' onClick={this.props.onClose}>
                     { _("Close") }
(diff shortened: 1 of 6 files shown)


KKoukiou:
Some tests are still failing - will repost.

pkg/lib/cockpit-components-modifications.jsx
```diff
@@ -88,22 +88,18 @@ class ModificationsExportDialog extends React.Component {
...
+                        <TextArea resizeOrientation='vertical' isReadOnly defaultValue={this.props.ansible.trim()} />
+                        <div className="ansible-docs-link">
+                            <QuestionCircleIcon size="xs" />
                             { _("Create new task file with this content.") }
-                            <a href="https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html"
-                                target="_blank" rel="noopener noreferrer">
-                                <i className="fa fa-external-link fa-xs" />
+                            <Button variant="link" component="a" href="https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html"
+                                    target="_blank"
+                                    icon={<ExternalLinkSquareAltIcon />}>
```
marusak, line 100: I find this icons a bit weird. We use `ExternalLinkAltIcon` elsewhere, I would go with that.

pkg/lib/cockpit-components-modifications.jsx
```diff
@@ -88,22 +88,18 @@ class ModificationsExportDialog extends React.Component {
...
                     </Tab>
                     <Tab eventKey="ansible" title={_("Ansible")}>
-                        <pre>
-                            {this.props.ansible.trim()}
-                        </pre>
-                        <div>
-                            <span className="fa fa-question-circle fa-xs" />
+                        <TextArea resizeOrientation='vertical' isReadOnly defaultValue={this.props.ansible.trim()} />
+                        <div className="ansible-docs-link">
+                            <QuestionCircleIcon size="xs" />
```
marusak, line 96: This black icon is just too punchy.
I would use `<OutlinedQuestionCircleIcon />` (without xs).

Before:
weirdicons

With icons I suggested:
newicons

pkg/selinux/setroubleshoot-view.jsx
```diff
@@ -161,44 +135,33 @@ class SELinuxEventDetails extends React.Component {
...
-                doState = <div>{caret} {detailsLink}</div>;
-                doElem = null;
-            }
             return (
-                <div className="list-group-item selinux-details" key={itm.analysisId + (itm.ifText || "") + (itm.doText || "")}>
-                    <div>
-                        <div>
-                            <span>{itm.ifText}</span>
-                        </div>
+                <Card isCompact key={itm.analysisId + (itm.ifText || "") + (itm.doText || "")}>
```
marusak, line 142: I am a bit on fence here. When I opened card that had just one solution I found it weird that it contains card. I came here to comment on it. But then I opened some other report that had two solutions and I saw why you picked it.

solutions

But I lean a bit more towards that Card is not the right component, it feels like it does not belong there. Have you thought about [DataList](https://www.patternfly.org/v4/components/data-list/)? Or just keeping it as `div`s and smashing [Divider](https://www.patternfly.org/v4/components/divider) inbetween?

marusak (changes requested):
Nice one! I have a few comments but mostly just about "i would like it more if it would be..." which you can ignore, if you feel like the current solution is better :)

pkg/lib/cockpit-components-modifications.jsx
KKoukiou: I am fine with both. For the record I just picked it because of the example here: https://www.patternfly.org/v4/components/button/#variations

pkg/lib/cockpit-components-modifications.jsx
```diff
@@ -88,22 +88,18 @@ class ModificationsExportDialog extends React.Component {
...
                             { _("Create new task file with this content.") }
-                            <a href="https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html"
-                                target="_blank" rel="noopener noreferrer">
```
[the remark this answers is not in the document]
KKoukiou: I thought it was something that was of no use really, but I just grepped out code and we use it for many external links I will add it back.

pkg/selinux/setroubleshoot-view.jsx
KKoukiou: Screen Shot 2021-03-11 at 10 35 04

pkg/lib/cockpit-components-modifications.jsx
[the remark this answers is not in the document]
marusak: https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable

It is security thingy and all urls in cockpit that take you to an external page should have it.

marusak (approved):
Nice one, thanks!

PR #15497, opened 2021-03-10
test: Don't try to use axe on Firefox

marusak:
And if it fails to find axe, write it to stderr and not to stdout.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1215,11 +1215,12 @@ class MachineCase(unittest.TestCase):
         will be appended to the file name, which is useful if you call this
         more than once within one test.
         """
-        if not checkRunAxe():
+
+        # Only test Axe on chromium browsers
+        if self.browser.cdp.browser != "chromium":
             return
-        # HACK: We cannot test axe on Firefox since `axe.run()` returns promise
-        # and Firefox CDP cannot wait for it to resolve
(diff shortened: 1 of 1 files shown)


test/common/testlib.py
```diff
@@ -1215,13 +1215,14 @@ def check_axe(self, label=None, suffix=""):
         will be appended to the file name, which is useful if you call this
         more than once within one test.
         """
-        if not checkRunAxe():
-            return
-        # HACK: We cannot test axe on Firefox since `axe.run()` returns promise
-        # and Firefox CDP cannot wait for it to resolve
+
+        # Only test Axe on chromium browsers
         if self.browser.cdp.browser == "firefox":
```
martinpitt, line 1220: Then let's actually make it do that with `!= "chromium`", just in case we get webkit or whatnot.

test/common/testlib.py
```diff
@@ -1474,7 +1475,7 @@ def checkRunAxe():
 
     # when running from release tarballs, module is not available
     if not os.path.exists(os.path.join(TEST_DIR, "common/axe.js")):
-        print('# enableAxe: axe is not installed, skipping')
+        sys.stderr.write('# enableAxe: axe is not installed, skipping')
```
martinpitt, line 1478: needs an extra `\n`

martinpitt (changes requested):

PR #15499, opened 2021-03-10
lib: make cockpit-logs-panels independent from PF3

KKoukiou:
The badges look different in bootstrap compared to PF4. Are we happy with the change?

Before:
Screen Shot 2021-03-10 at 14 59 31

After:
Screen Shot 2021-03-10 at 14 59 20

### DIFF
--- a/pkg/lib/cockpit-components-logs-panel.jsx
+++ b/pkg/lib/cockpit-components-logs-panel.jsx
@@ -20,7 +20,8 @@
 import cockpit from "cockpit";
 import React from "react";
 
-import { Card, CardHeader, CardActions, CardTitle, CardBody } from '@patternfly/react-core';
+import { Badge, Card, CardHeader, CardActions, CardTitle, CardBody } from '@patternfly/react-core';
+import { AngleRightIcon, ExclamationTriangleIcon, TimesCircleIcon } from '@patternfly/react-icons';
 
 import { journal } from "journal";
 import "journal.css";
@@ -68,11 +69,11 @@ export class JournalOutput {
                 onKeyPress={ev => this.onEvent(ev, entry.__CURSOR)}>
                 <div className="cockpit-log-warning" role="cell">
                     { warning
-                        ? <i className="fa fa-exclamation-triangle" />
+                        ? <ExclamationTriangleIcon className="ct-icon-exclamation-triangle" size="sm" />
                         : null
                     }
                     { problem
-                        ? <i className="fa fa-times-circle-o" />
+                        ? <TimesCircleIcon className="ct-icon-times-circle" size="sm" />
                         : null
                     }
                 </div>
@@ -82,7 +83,7 @@ export class JournalOutput {
                     count > 1
                         ? <div className="cockpit-log-service-container" role="cell">
                             <div className="cockpit-log-service-reduced" role="cell">{ident}</div>
-                            <span className="badge" role="cell">{count}&#160;<i className="fa fa-caret-right" /></span>
+                            <Badge isRead key={count} role="cell">{count}&#160;<AngleRightIcon /></Badge>
                         </div>
                         : <div className="cockpit-log-service" role="cell">{ident}</div>
                 }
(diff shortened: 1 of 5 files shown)


marusak:
> Are we happy with the change?

I am :) But I am not the one to make this decision

garrett:
Wow, this is much needed! Looks great in the screenshots!

I'm building it now and will approve it immediately if I don't find anything unexpected.

garrett (dismissed):
Looks good locally! :+1:

I'll check the livetest in a minute too, but based on screenshots and local testing, it's good!

martinpitt:
Sorry about the "Aborted without status", I had to restart our bots. But there are plenty of finished tests, and the [journal test still fails](https://logs.cockpit-project.org/logs/pull-15499-20210310-142035-7766ffdd-debian-stable/log.html#283-2)

pkg/lib/cockpit-components-logs-panel.jsx
```diff
@@ -82,7 +83,7 @@ export class JournalOutput {
                     count > 1
                         ? <div className="cockpit-log-service-container" role="cell">
                             <div className="cockpit-log-service-reduced" role="cell">{ident}</div>
-                            <span className="badge" role="cell">{count}&#160;<i className="fa fa-caret-right" /></span>
+                            <Badge isRead key={count} role="cell">{count}&#160;<AngleRightIcon /></Badge>
```
marusak, line 86: Do we need that `>`?
I never understood what it was trying to say.
For example we have the same logic of "squash same reports" in selinux and we don't show that icon there:
counter

KKoukiou: The difference between selinux and logs panel is that in the logs that badge is actually redirecting to the logs details when clicked on. I agree that's not obvious at all though that this arrow suggests that the badge contains a link.

KKoukiou: To be clear, it's not the badge that it's redirecting, the whole line is having the onClick handlers.

KKoukiou: Heh, we commented together. Shall I repost then removing the arrows and making the onClick be on the text?

marusak: @garrett WDYT
- is that arrow necessary? (This predates me, so I am not sure, if there has been some reason behind this)
- Would it be fine, if only text would redirect and not the whole line? Now on hover I think just the text changes color and cursor.

KKoukiou: @marusak How about I post a followup PR on this? I have the code change locally in fact, but since this is fine from @garrett and the tests seem happy I would rather merge and introduce the design change in a seperate PR.

marusak (approved):
Thank you!

pkg/lib/cockpit-components-logs-panel.jsx
marusak: Agree! Lets not block on this, especially since this blocks another PR.

KKoukiou: I see an issue with making the message the only clickable part of a row:

* If we make it a link or a button it will become blue by default (probably we don't want that)
* If we don't make it link or a button it will not be blue and the users will not have any idea where to click on the row to get redirected. Now indeed we only underline the message on hover, but at least clicking anywhere in the row will redirect.

garrett: I don't think the arrow is necessary. 

It doesn't do anything. From my understanding, the number is just the count of the number of the times it is repeated. I never understood why it had an arrow. (It predates me working on Cockpit too.)

If we all agree it isn't necessary, let's just drop it?

marusak: Agree! We will do this as follow up though as we want to move this forward.
@KKoukiou will you send PR for that?

PR #15504, opened 2021-03-11
Add FMF test, enable packit tests for Fedora 33

martinpitt:
This unifies our downstream and upstream testing. This is inspired by what we did in [starter-kit](https://github.com/cockpit-project/starter-kit/pull/432) and [podman](https://github.com/cockpit-project/cockpit-podman/pull/649), but for Cockpit it's a little more complicated.

I tested this extensively on https://github.com/martinpitt/cockpit/pull/12 , testing various different approaches, and this now works reasonably well. I propose to enable F33 first, and then 34 separately, as that still has a bunch of failures.

There is currently a big quirk: packit create-archive times out after 10 minutes, which most of the time is not enough to wait for the generated dist tarball to arrive. This is tracked in https://github.com/packit/packit-service/issues/1017 . The workaround is to send a `/packit build` comment after the "build" action is done. @TomasTomecek said he'll address that soon by bumping the timeout.

**Update**: The fix for packit just arrived: https://github.com/packit/sandcastle/pull/109


martinpitt:
I am a bit irritated about this (unrelated) [TestAD.testQualifiedUsers](https://logs.cockpit-project.org/logs/pull-15504-20210311-122809-4ac70864-fedora-34/log.html#285-2) test failure -- that's exactly https://github.com/cockpit-project/bots/issues/1735 and we have a naughty for that.

But indeed it also does not match locally -- I wonder if that's because of the 23089423423 messages that say

    warning: An accessible aria-label is required when using the progress component without a title.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/1781 to fix the fedora-34 failure above, everything else is green :four_leaf_clover:

test/run-verify-host.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh
```
marusak, line 1: `test/run-verify-host-user.sh` and ` test/run-verify-host.sh` names are a bit confusing. I expected one to run as `user` and the other as `root` doing the same or something like that. Which is kind of true, but still confusing at first to discover what does one script and what does the other one.

martinpitt: I'll gladly rename them to something more sensible. Any ideas? Maybe verify-host-{setup,run} ?

There's more work to be done for F34 (see https://github.com/martinpitt/cockpit/pull/13), and since with FMF the script name is not visible/relevant from outside any more, we can rename this at will.

martinpitt:
@marusak: Note that we run a loooot more Machines tests in RHEL, but they don't run well in the AWS TF instances. I hope in the future we'll get some more powerful instances, I was already talking about that with Miro.

PR #15515, opened 2021-03-12
Some fixes for machines page

KKoukiou:
The storage commit maks now the warning size smaller as expected:
With this PR:
Screen Shot 2021-03-15 at 14 51 12


garrett:
The progress bars also communicate the minimum and maximum, whereas that's not done in the text entries alone.

Additionally, there's a visual issue with the text entry butted up next to the dropdown. It might be PF related?

Screenshot from 2021-03-15 11-50-21 entry issue

And the options aren't memory-related, are they? (They should have their own label. Perhaps "Boot options"?)

Screenshot from 2021-03-15 11-50-21 memory options

garrett:
I'll continue looking at the other changes after lunch. I just wanted to share this initial impression first.

garrett:
This CSS will fix the focus ring truncation:

```css
.pf-c-input-group :focus {
  position: relative;
  z-index: 2;
}
```
Changing the position allows for z-index. Changing the z-index brings the focused element to the top, where the ring can be seen.

Confirmed that this is an upstream issue @ https://www.patternfly.org/v4/components/input-group/

I'll look into filing an issue with this fix. (Or finding one, if it exists, to suggest this.)

garrett:
Opened https://github.com/patternfly/patternfly/issues/3926 with the above CSS snippet as a suggested fix.

garrett:
Setting position messes with dropdowns. However, in testing, at least on Firefox, it's not needed. So the CSS override (until this is fixed upstream) would look like:

```css
.pf-c-input-group :focus {
  z-index: 2;
}
```

I've also updated the upstream issue.

KKoukiou:
> The progress bars also communicate the minimum and maximum, whereas that's not done in the text entries alone.

These boundaries were not helpful however in many scenarios. Specifically, the lower limits for memory and disk size, where the minumum suggested values from libosinfo, and these are always mor ethan necessary to spawn a VM. See the commit message for more info. The upper limits also were also problematic sometimes.  
> 
> Additionally, there's a visual issue with the text entry butted up next to the dropdown. It might be PF related?
> 
> Screenshot from 2021-03-15 11-50-21 entry issue
> 
> And the options aren't memory-related, are they? (They should have their own label. Perhaps "Boot options"?)

This used to be exactly like this before this pR as well. There are not boot options. Let's discuss it in a seperate issue how to group these.
> 
> Screenshot from 2021-03-15 11-50-21 memory options

garrett:
> This used to be exactly like this before this pR as well. 

> Let's discuss it in a seperate issue how to group these.

Sure. The title is "some fixes for the machine page, and this was some suggested fix. And there's no description in the PR.

But, okay, it can go elsewhere too.

> There are not boot options. 

It's literally what to do on first boot and when to boot (now, or specified later). But I guess, yeah, the term might be overloaded. They're not kernel parameters or options to pass to virsh or anything like that.

"Options" is kind of weird, as everything in the dialog is basically an option too. But we'll discuss in some other issue/PR.

garrett:
Here's the workaround for the label issue, as a separate PR: https://github.com/cockpit-project/cockpit/pull/15525

garrett:
Opened an issue for the missing options label @ https://github.com/cockpit-project/cockpit/issues/15526

martinpitt (approved):
Epharisto! I like how that works now -- it still gives good recommendations, but I can happily shoot myself into the foot now if I know what I'm doing. Code looks fine, and Garrett already reviewed the design (I didn't notice any thing odd there).

KKoukiou:
For the failing tests I already retried, but they are common flakes. I will merge with these failures.

PR #15520, opened 2021-03-14
workflows: Retry dist git commit several times

martinpitt:
When multiple parallel workflows race against each other, the final
`git push` may fail if origin/master has moved while trying to update it
locally. In that case, do the entire update construction again, as in
the meantime older tarballs may also have been cleaned up already.


martinpitt:
I noticed this in https://github.com/cockpit-project/cockpit/pull/15518 where the publish-dist tarball failed because of this race, and thus packit kept failing because there was no downloadable dist tarball.

I tested this [on my fork](https://github.com/martinpitt/cockpit/actions)/. Unfortunately I didn't manage to "win" the race (of course), both workflows succeeded at the first try. But this at least shows that I didn't make it worse. :grinning:

marusak (approved):
Danke

PR #15524, opened 2021-03-15
packagekit: fix alignment issues on smaller screens

### LINKED ISSUE #15523: Updates page is visually broken on smaller screens
When I make my window smaller the updates page gets wonky.
1. The 'Status' gets unnecessary wrapped
2. But mainly, nothing aligns in details.

updatesistragic

KKoukiou:
Fixes #15523

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -41,6 +41,7 @@ import {
     RedoIcon,
     ProcessAutomationIcon
 } from "@patternfly/react-icons";
+import { cellWidth, TableText } from "@patternfly/react-table";
 import { Remarkable } from "remarkable";
 
 import { AutoUpdates, AutoUpdatesBody } from "./autoupdates.jsx";
@@ -251,19 +252,18 @@ function updateItem(info, pkgNames, key) {
         type = (
             <>
                 <Tooltip id="tip-severity" content={ secSeverity || _("security") }>
-                    <span className={iconClasses} />
+                    <span className={iconClasses + ' severity-icon'} />
                     { (info.cve_urls && info.cve_urls.length > 0) ? info.cve_urls.length : "" }
                 </Tooltip>
             </>);
     } else {
         const tip = (info.severity >= PK.Enum.INFO_NORMAL) ? _("bug fix") : _("enhancement");
         type = (
-            <>
-                <Tooltip id="tip-severity" content={tip}>
-                    <span className={iconClasses} />
-                    { bugs ? info.bug_urls.length : "" }
-                </Tooltip>
-            </>);
+            <Tooltip id="tip-severity" content={tip}>
+                <span className={iconClasses + ' severity-icon'} />
+                { bugs ? info.bug_urls.length : "" }
+            </Tooltip>
+        );
     }
 
     const pkgList = pkgNames.map((n, index) => (
@@ -274,7 +274,7 @@ function updateItem(info, pkgNames, key) {
     const pkgs = pkgList;
     let pkgsTruncated = pkgs;
     if (pkgList.length > 4)
-        pkgsTruncated = pkgList.slice(0, 4).concat("…");
+        pkgsTruncated = pkgList.slice(0, 4).concat(<span>…</span>);
 
     let descriptionFirstLine = (info.description || "").trim();
     if (descriptionFirstLine.indexOf("\n") >= 0)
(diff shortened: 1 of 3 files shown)


marusak:
Needs tests adjustment

marusak:
So much better, thank you!

A few more issues:
dostsandnumer

- Notice those three dots - they should be in the list of packages and not as label
- The icon and number are too close and mis-aligned. Possibly unrelated but would be nice to fix as well :)

alwaysdots

Not sure if this is regression or not, but version number is always truncated. Feels a bit weird that on such big resolution not a single version is shown fully.

Possibly some of those are unrelated (the dots one is surely regression).

marusak (changes requested):
At least dots for truncated lists need to be correctly placed.

marusak:
Actually, both badge-number position and version truncation are regressions. Current master:

regressions

KKoukiou:
@marusak fixed all comments. with fixes the screenshots in different screensizes follow:
Screen Shot 2021-03-16 at 12 42 04
Screen Shot 2021-03-16 at 12 40 52
Screen Shot 2021-03-16 at 12 40 44

marusak:
The icon and number are still mis-aligned and too close to each other - in your screenshots it looks correct, but when I build it, it does not look like that...

KKoukiou:
> The icon and number are still mis-aligned and too close to each other - in your screenshots it looks correct, but when I build it, it does not look like that...

Hmm how did you build it? I just did rm -rf dist/ node_modules/ and test/make-dist and it is correctly spaced. I also see in the dev tools the CSS that actually adds the spacing.
```
.ct-table tr td.type .severity-icon {
    margin-right: var(--pf-global--spacer--xs);
}
```

Could you maybe open the dev tools as well and check that rule?

marusak:
I cloned it and `make`.
But also I don't see how your [force push](https://github.com/cockpit-project/cockpit/compare/0d14460b1962acafab5478ee5b03758ac74c5cb4..8173705ff1cb4a4b12f1fbffd031d90839b57a23) addressed it. Or haven't you seen this issue before either?

KKoukiou:
> I cloned it and `make`.
> But also I don't see how your [force push](https://github.com/cockpit-project/cockpit/compare/0d14460b1962acafab5478ee5b03758ac74c5cb4..8173705ff1cb4a4b12f1fbffd031d90839b57a23) addressed it. Or haven't you seen this issue before either?

It was supposed to be working before that force push.
In fact, when you first commented I was surprised because i was not seeing the issue. Wanna try the rm -rf dist/, make_dist approach? That should be identical code between us at least.

marusak:
Found it :) The problem is that you are seeing bug fix or enhancement, but I see this on security updates. And indeed, code agrees:
```js
    if (info.severity === PK.Enum.INFO_SECURITY) {                              
        if (secSeverityURL)                                                     
            secSeverityURL = <a rel="noopener noreferrer" target="_blank" href={secSeverityURL}>{secSeverity}</a>;
        type = (                                                                
            <>                                                                  
                <Tooltip id="tip-severity" content={ secSeverity || _("security") }>
--->                     <span className={iconClasses} />                            
                    { (info.cve_urls && info.cve_urls.length > 0) ? info.cve_urls.length : "" }
                </Tooltip>                                                      
            </>);                                                               
    } else {                                                                    
        const tip = (info.severity >= PK.Enum.INFO_NORMAL) ? _("bug fix") : _("enhancement");
        type = (                                                                
            <Tooltip id="tip-severity" content={tip}>                           
--->               <span aria-label={tip} className={iconClasses + ' severity-icon'} />
                { bugs ? info.bug_urls.length : "" }                            
            </Tooltip>                                                          
        );                                                                      
    }    

```

The first line needs to add ` ' severity-icon'` as well.

icons

pkg/packagekit/updates.scss
```diff
@@ -70,36 +70,11 @@ table.listing-ct thead th {
...
-                overflow: visible;
-
-                .truncating {
-                    background: linear-gradient(to left, rgba(237, 248, 255, 0), var(--ct-color-list-hover-bg) 3em);
-                    position: relative;
-                    padding: 0.5em 4em 0.5em 0;
-                }
-            }
+        .severity-icon {
+            margin-right: var(--pf-global--spacer--xs);
```
marusak, line 74: can this selector be made simpler?
like:
```
.ct-table tr {
    .severity-icon {
...
```
? I guess that the `td.type` is not necessary there, is it? And if is, it can be still squashed into one-line selector.
```

Comment (KKoukiou) [reply pkg/packagekit/updates.scss]: done

File: pkg/packagekit/updates.jsx [hunk 2]
```diff
@@ -251,19 +252,18 @@ function updateItem(info, pkgNames, key) {
         type = (
             <>
                 <Tooltip id="tip-severity" content={ secSeverity || _("security") }>
-                    <span className={iconClasses} />
+                    <span aria-label={secSeverity || _("security")} className={iconClasses + 'severity-icon'} />
```
marusak, line 255: you need to put there space before `severity-icon` as now it creates something like `class="fooseverity-icon"` and not `class="foo severity-icon"`.

marusak (changes requested):

marusak (approved):
Thanks!

PR #15528, opened 2021-03-15
kdump: redesign to match the styling of other pages

### LINKED ISSUE #15511: kdump: the label of the switch is misaligned
Spotted on: cockpit-239

misaligned-kdump

This happends because we use link as the switch label. This is not expected usage of the Switch component, and maybe we want to reapproach the design of this line.

KKoukiou:
Fixes #15511

Redesigned as requested in https://github.com/cockpit-project/cockpit/issues/15511#issuecomment-799502949

kdump-refresh

### DIFF
--- a/pkg/kdump/kdump-view.jsx
+++ b/pkg/kdump/kdump-view.jsx
@@ -17,18 +17,20 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
+import '../lib/patternfly/patternfly-4-cockpit.scss';
 import cockpit from "cockpit";
 
 import React from "react";
 import {
     Button, Checkbox,
+    Card, CardBody,
+    Flex,
     Form, FormGroup, FormSection,
     FormSelect, FormSelectOption,
     Page, PageSection, PageSectionVariants,
-    Flex,
     DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
     Spinner, Switch,
-    TextInput, Tooltip, TooltipPosition,
+    TextInput, Title, Tooltip, TooltipPosition,
 } from "@patternfly/react-core";
 import { OutlinedQuestionCircleIcon } from "@patternfly/react-icons";
 
@@ -444,8 +446,6 @@ export class KdumpPage extends React.Component {
                 serviceDescription = <span>{_("Service is starting")}</span>;
             else if (this.props.kdumpStatus.state == "stopping")
                 serviceDescription = <span>{_("Service is stopping")}</span>;
-            else
-                serviceDescription = <span>{_("More details")}</span>;
             if (this.props.reservedMemory == 0) {
                 const tooltip = _("No memory reserved. Append a crashkernel option to the kernel command line (e.g. in /etc/default/grub) to reserve memory at boot time. Example: crashkernel=512M");
                 serviceHint = (
@@ -456,8 +456,9 @@ export class KdumpPage extends React.Component {
             }
             kdumpServiceDetails = (
                 <>
-                    <Button variant="link" isInline className="service-link-ct-kdump" onClick={this.handleServiceDetailsClick}>{serviceDescription}</Button>
+                    {serviceDescription}
                     {serviceHint}
+                    <Button variant="link" isInline className="service-link-ct-kdump" onClick={this.handleServiceDetailsClick}>{_("more details")}</Button>
                 </>
             );
         } else if (this.props.kdumpStatus && !this.props.kdumpStatus.installed) {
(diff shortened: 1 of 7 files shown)


martinpitt:
Moving to draft (needs test adjustments)

garrett:
It's definitely better than what's merged in. However, I saw some oddness where the description list doesn't responsively change.

I'm happy with what's here compared to what's in Cockpit currently, but if it's not much effort to make the description list adjust itself on smaller sizes, that would be nice in this PR too.

## Normal
Screenshot_2021-03-16 Kernel Dump - garrett Rain

## Small desktop / Mobile
Wrapping is a bit odd.

Screenshot_2021-03-16 Kernel Dump - garrett Rain(1)

## Small mobile
Shouldn't this be responsive and move the labels above at small sizes?

Screenshot_2021-03-16 Kernel Dump - garrett Rain(2)

garrett:
Disclaimer: OK, so what I'm writing below is outside the scope of this PR, as I think this PR should only be to freshen up the paint by making it look like the rest of Cockpit.

---

Ideally, it would be nice if we could guide people what to do with this. Right now, it's not obvious what this is, what it does, or how to properly set it up.

I guess it's something to analyze kernel crashes... but that's all I know about it.

When I asked about it before, I was told it's some rare edge case thing that few use. I don't know if that's the case or what. Regardless, it could still use some redesign... but I'd need to know how it's used before suggesting anything other than the obvious.

However, that all said... the obvious things that jump out at me:

1. There's this huge tooltip with lots of text. If it's important, it should be on the page itself. If it's not so important, having it on a popover or a modal would be better. It could probably stand to be rewritten to be more concise as well.
2. We're trying to not have the links be actions, but locations. That goes for editing as well. There should be an edit button instead of a link.
3. The information assumes that it's set up properly. In a default state, this should probably use an empty state pattern and try to help set up things the first time.
4. Similarly, the test configuration button is disabled, as it _knows_ that this isn't set up.
5. Same for reserved memory... it's a label, but it isn't changable.
6. More details link goes to the system service. Once you're there, it's not obvious on what to do, and when you click back it doesn't work. (This back button problem is an open issue with the system journal and a few other places in Cockpit and should definitely be solved, but it's out of scope for kdump. But it does impact the UX of kdump.)

So, it should probably be something like:

1. Empty state pattern with a summary of what this is about. ("Kernel crash debugging is not active. Enable 'kdump' to help troubleshoot kernel crashes by capturing memory dumps." or something like this. I don't know.)
2. When enabling it for the first time, it should work without weird setups. (Have good defaults.)
3. Only show a lot of the things when the service is active and configured correctly. You can't test a configuration that doesn't exist.
4. RAM value should be [set based on whatever kdump docs suggest](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/kernel_administration_guide/kernel_crash_dump_guide#sect-kdump-memory-requirements). This depends on the system's architecture and available memory. But we know both of these things. It also says "Crash memory automatically reserved"... so do we even need to set it? Or are we just showing how much space is being used by kdump being active?

---

Anyway, I'm happy with this PR as is. Responsiveness would be a nice bonus. The rest can be thought about and worked on later.

garrett (dismissed):
This is fine enough on modern phone sizes. (There's a little bit of wrap, but it's not horrible.)

And I doubt anyone would (or should) be messing around with kdump from their phones?

So I'm approving it as-is. It's great. Thanks!

KKoukiou:
> This is fine enough on modern phone sizes. (There's a little bit of wrap, but it's not horrible.)
> 
> And I doubt anyone would (or should) be messing around with kdump from their phones?
> 
> So I'm approving it as-is. It's great. Thanks!

I think I can fix it, at least the description list should wrap. Let me try.

garrett:
I got kdump working in my F33 VM by running this in the terminal in Cockpit:

```
grubby --update-kernel=ALL --args="crashkernel=auto"
```

And then I rebooted the VM.

You can tell it specific kernels to enable it for or all. You can also tell it the amount of RAM to reserve, but auto does the right thing each time (so there really shouldn't be a need to manually set it).

I did also try crashing the kernel, which worked, but when I reconnected, I don't see anything about it in the page. It's as if I never clicked the button. That's confusing. However, there were items in `/var/crash/`, so I assume it worked as intended:

```sh
[admin@m1 ~]$ ls /var/crash
127.0.0.1-2021-03-16-10:16:06

[admin@m1 ~]$ ls /var/crash/127.0.0.1-2021-03-16-10\:16\:06/
kexec-dmesg.log  vmcore  vmcore-dmesg.txt
```

First step to improving the UI would be to automate the setup process (and not have it behind a switch). But again, it's low priority and outside of this PR.

garrett:
Here's what the page looks like when kdump is active, enabled, working, etc.:

Screenshot_2021-03-16 Kernel Dump - admin m1 cockpit lan(1)

Screenshot_2021-03-16 Kernel Dump - admin m1 cockpit lan(2)

garrett:
FWIW, grubby works [at least in RHEL7](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-Working_with_the_GRUB_2_Boot_Loader#sec-Making_Persistent_Changes_to_a_GRUB_2_Menu_Using_the_grubby_Tool) and up too.

So it's available in Fedora / RHEL / CentOS at least.

(Not sure about Debian / Ubuntu. It's apparently available openSUSE, but it looks like not by default. Not sure about other systems either.)

garrett:
I think I have a fix pending for the responsiveness @ https://github.com/cockpit-project/cockpit/pull/15535.

It's weird that the linked issue did not fix this by itself, and it's odd that the workaround didn't apply when it should have. :shrug: Perhaps it was a mistake, or perhaps I need more coffee. :coffee: Doesn't matter who/what/when/why or even how-much-coffee; the change in that PR should address the issue and keep it separate from this PR.

KKoukiou:
> I think I have a fix pending for the responsiveness @ #15535.
> 
> It's weird that the linked issue did not fix this by itself, and it's odd that the workaround didn't apply when it should have. shrug Perhaps it was a mistake, or perhaps I need more coffee. coffee Doesn't matter who/what/when/why or even how-much-coffee; the change in that PR should address the issue and keep it separate from this PR.

I fixed the responsiveness issue, with the pf-m-horizontal-on-sm class.

garrett:
> pf-m-horizontal-on-sm

Oh! In PF itself? Cool. I'll change that PR to remove the override then, if we don't need it.

KKoukiou:
> > pf-m-horizontal-on-sm
> 
> Oh! In PF itself? Cool. I'll change that PR to remove the override then, if we don't need it.

It's not that simple. The lists are not automatically responsive. They need the relevevant classes. See the issue in the override.

garrett:
I don't see anything about them not being responsive in the issue listed in the override. I just see about being able to specify a breakpoint, but nothing about having a default breakpoint (or not).

garrett (dismissed):
This looks good enough to me to bring it up to date.

I guess I should open a new issue with the complaints and suggestions, right? (And someone, at some time can address it all.)

KKoukiou:
@marusak it's design approved, I would like a code review as well though.

pkg/kdump/kdump-view.jsx
```diff
@@ -494,47 +494,54 @@ export class KdumpPage extends React.Component {
...
-                        <DescriptionList>
-                            <DescriptionListGroup>
-                                <DescriptionListTerm>{_("kdump status")}</DescriptionListTerm>
-                                <DescriptionListDescription>
-                                    <div role="group">
-                                        <Switch isChecked={!!serviceRunning}
-                                                onChange={this.props.onSetServiceState}
-                                                aria-label={_("kdump status")}
-                                                isDisabled={this.props.stateChanging} />
+                    <span className="kdump-heading-title-group">
```
marusak, line 497: Why you dropped `Flex` and instead wrote custom css? Is there something which `Flex` cannot do and we need our own css?

KKoukiou: I just copied the CSS from the firewall.css, I think I can do it from React as well. I can repost or send a followup and change it in the other pages as well, aka firewall, selinux, they all use the same flex for the headers.

marusak:
Oh right, there is no kdump in livetest machine :) But anyway, design is approved. Adding release note, @KKoukiou can you please post current screenshot?

pkg/kdump/kdump-view.jsx
marusak: Ah right, I see. Your call :) 
I would prefer slightly more doing it here as this PR introduces unnecessary css but I am fine with followup as well. I'll give you ack not to block this.

marusak (dismissed):
Thank!

pkg/kdump/kdump-view.jsx
KKoukiou: This PR, it's not really fixing anything, it's just redesign. So I am not in a hurry to merge it. I agree on your point on not introducing custom CSS, I am shooting at my own leg here basically :package: 

I will work on this PR. Moving it to draft.

garrett:
Modified screenshot to add a border and optimize. Added it to the top, as usual.

KKoukiou:
@marusak replaced custom CSS with pF4 flex in react, and changed the pattern also in the other pages that use it, as we discussed above.

marusak (approved):
Thank you!

Just double checking that the card is supposed to span across the whole screen, right?
spanns

garrett (dismissed):
Re-approving this from the design side. Looks like the screenshot.

One minor thing, which I guess might be solved by an additional PR:

Spinner doesn't have proper spacing between it and the text and is misaligned vertically.

kdump-spinner-alignment-spacing

kdump-spinner-alignment-spacing-3x

I've seen this before elsewhere. I'm not sure what the solution is. Perhaps there's something we can do to make sure it's proper using just PF4? :shrug:

KKoukiou:
> Thank you!
> 
> Just double checking that the card is supposed to span across the whole screen, right?
> spanns

Yes this is expected. I think it would be really weird to make the card not span to the end.

KKoukiou:
darn, I did not notice the spinner misalignment, I will repost @marusak @garrett

KKoukiou:
@garrett fixed the spacing issue, the spinner to SVG is handled in your PR.

garrett (approved):
@KKoukiou: Thanks for the spinner fixes!

LGTM :+1:

PR #15530, opened 2021-03-15
packagekit: lib: use TimePicker component from PF4

### LINKED ISSUE #1505: CockpitPipe closes itself prematurely
A CockpitPipe with in_fd == -1 and out_fd != -1 will close itself after the first cockpit_pipe_write.

This happens because the output queue will become empty which causes close_maybe to close the pipe.

### LINKED ISSUE #15055: Use PF4's new time picker
We're using a time selector in a few places in Cockpit. I suggest we consider PatternFly's new basic time picker:

https://github.com/patternfly/patternfly-org/blob/master/release-highlights/release-highlights-2020.16/release-highlights.md#basic-time-picker

1. The most obvious place I can think of is in the Software Updates automatic updates scheduling. 
2. Additionally, I think we have a scheduled time to reboot too (not the X minutes from now, but at a certain specific time).
3. This might be useful for the journal as well.
4. Performance page may also be a candidate for this widget.

Am I forgetting anything else where you need to select a specific time?

KKoukiou:
Fixes #15055

### DIFF
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -182,9 +182,11 @@ export class ShutdownModal extends React.Component {
                                 </Select>
                                 {this.state.selected === "x" && <>
                                     <DatePicker aria-label={_("Pick date")} locale={cockpit.language} dateFormat={d => moment(d).format('L')}
-                                        invalidFormatText="" dateParse={d => moment(d, 'L').toDate()}
-                                        value={moment(this.state.date).format('L')} onChange={(d, ds) => this.updateDatetime("date", ds)} />
+                                                className='shutdown-date-picker'
+                                                invalidFormatText="" dateParse={d => moment(d, 'L').toDate()}
+                                                value={moment(this.state.date).format('L')} onChange={(d, ds) => this.updateDatetime("date", ds)} />
                                     <TimePicker defaultTime={this.state.time} is24Hour
+                                                className='shutdown-time-picker'
                                                 id="shutdown-time"
                                                 invalidFormatErrorMessage=""
                                                 menuAppendTo={() => document.body}
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -174,7 +174,7 @@ export class ShutdownModal extends React.Component {
                         <FormGroup fieldId="delay" label={_("Delay")}
                                    helperTextInvalid={this.state.dateError}
                                    validated={this.state.dateError ? "error" : "default"}>
-                            <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                            <Flex className="shutdown-delay-group" alignItems={{ default: 'alignItemsCenter' }}>
                                 <Select toggleId="delay" isOpen={this.state.isOpen} selections={this.state.selected}
                                         className='shutdown-select-delay'
                                         onToggle={o => this.setState({ isOpen: o })} menuAppendTo="parent"
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -27,6 +27,7 @@ import {
 } from '@patternfly/react-core';
 
 import { ServerTime } from 'serverTime.js';
+import { validateTime } from 'timepicker-helpers.js';
 
 import "cockpit-components-shutdown.scss";
 
@@ -84,15 +85,7 @@ export class ShutdownModal extends React.Component {
             return;
         }
 
-        const h = parseInt(this.state.time.split(":")[0], 10);
-        const m = parseInt(this.state.time.split(":")[1], 10);
-
-        let time_error = false;
-        if (isNaN(h) || h < 0 || h > 23 ||
-            isNaN(m) || m < 0 || m > 59) {
-            time_error = true;
-        }
-
+        const time_error = !validateTime(this.state.time);
         const date = this.state.date;
 
         let date_error = false;
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -152,7 +152,7 @@ export class ShutdownModal extends React.Component {
             <SelectOption value="20" key="20">{_("20 minutes")}</SelectOption>,
             <SelectOption value="40" key="40">{_("40 minutes")}</SelectOption>,
             <SelectOption value="60" key="60">{_("60 minutes")}</SelectOption>,
-            <Divider key="divider" component="li" />,
+            <Divider key="divider-2" component="li" />,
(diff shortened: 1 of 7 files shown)


martinpitt:
Moving to draft due to test failures

KKoukiou:
> Moving to draft due to test failures

it was just flakiness. All affected tests ran 3 times successfully in a row now.

KKoukiou:
@martinpitt our 4st most common flake especially for centos - see https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?#

I have not managed to make it fail locally. So I am sure it's the 3 times force succeed to blame here. (it already succeeded the first time if yo see on centos here)

KKoukiou:
adding release blocker because it contains a bug fix.

martinpitt:
The [TestShutdownRestart.testBasic](https://logs.cockpit-project.org/logs/pull-15530-20210316-104940-2ea492e5-centos-8-stream/log.html#325-1) failure is  directly related. Obviously a flake, but the [screenshot](https://logs.cockpit-project.org/logs/pull-15530-20210316-104940-2ea492e5-centos-8-stream/TestShutdownRestart-testBasic-centos-8-stream-127.0.0.2-2501-FAIL.png) shows that the test is trying to set the input to "blah", but the input line does not respect that? Possibly the input validation resets that right away, and thus showing "blah" is not  a steady state? I.e. should this check be dropped?

KKoukiou:
> The [TestShutdownRestart.testBasic](https://logs.cockpit-project.org/logs/pull-15530-20210316-104940-2ea492e5-centos-8-stream/log.html#325-1) failure is directly related. Obviously a flake, but the [screenshot](https://logs.cockpit-project.org/logs/pull-15530-20210316-104940-2ea492e5-centos-8-stream/TestShutdownRestart-testBasic-centos-8-stream-127.0.0.2-2501-FAIL.png) shows that the test is trying to set the input to "blah", but the input line does not respect that? Possibly the input validation resets that right away, and thus showing "blah" is not a steady state? I.e. should this check be dropped?

It's failing here exactly the same way, just different component, because indeed I changed it. 
https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&test=test%2Fverify%2Fcheck-system-shutdown-restart+TestShutdownRestart.testBasic#

on centos-8 it has 35% failure rate, so I am not suprised this won't get green.

martinpitt:
I just tested that interactively, and I *can* input a bogus time, and it's a steady state. Perhaps our test is just missing to blur this properly, or what not?

image

So I think the test is still justified in general,  it's just broken. But independently from this PR, and we can/should fix that separately.

martinpitt (changes requested):
The above reboot dialog works fine, the "Reboot" button is disabled for invalid time formats. But that's not the case in the automatic updates:

image

I can click "Save changes" with that, and it happily accepts that. It even accepts it with "blah" and thus writes a broken config file:

image

KKoukiou:
> The above reboot dialog works fine, the "Reboot" button is disabled for invalid time formats. But that's not the case in the automatic updates:
> 
> image
> 
> I can click "Save changes" with that, and it happily accepts that. It even accepts it with "blah" and thus writes a broken config file:
> 
> image

Oh I know what this is. I will fix it tomorrow and try to figure out the flake in the shutdown dialog as well.

martinpitt:
The Reboot dialog now has a broken layout, even with very wide windows and English:

image

The time selector is much too wide now.

The automatic updates time dialog had that problem before (see screenshots above), but *that* layout is fine now:

image

However, it still does not properly disable "Save changes" on invalid times. I can click the button; nothing happens, and the dialog stays open, but I think it should be gray and light, and not clickable. This does not affect current master, as there is no free-form input and it is impossible to select an invalid time.

pkg/lib/cockpit-components-shutdown.scss
```diff
@@ -10,4 +10,6 @@
 
 .pf-c-date-picker {
     margin-left: 1em;
+    max-width: 13rem;
 }
+
```
martinpitt, line 14: trailing space

pkg/lib/cockpit-components-shutdown.jsx
```diff
@@ -32,6 +32,26 @@ import "cockpit-components-shutdown.scss";
 
 const _ = cockpit.gettext;
 
+// FIXME: https://github.com/patternfly/patternfly-react/issues/5564
+// Copied from: https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/TimePicker/TimePickerUtils.tsx
+// and slightly adjusted
+const validateTime = (time) => {
```
martinpitt, line 38: There are now two code copies of this around -- do you think that will be done in PF-react proper, and these can go away entirely in the forseeable future? Or should we move this into a cockpit-component-timepicker?

martinpitt (changes requested):
Code looks good otherwise, just a question and a tiny nitpick. Thanks!

pkg/lib/cockpit-components-shutdown.jsx
KKoukiou: I can work on the PF4 repo so that we will not need it:
It's because we cant really know in our code if the inserted value is valid: https://github.com/patternfly/patternfly-react/issues/5564

Also their validator is always applied regardless of the input so I had to make sure it's the same.
See: https://github.com/patternfly/patternfly-react/issues/5565

So, I will move this to lib, it does not hurt and work on it soon regardless.

KKoukiou:
> The Reboot dialog now has a broken layout, even with very wide windows and English:
> 
> image
> 
> The time selector is much too wide now.
> 
> The automatic updates time dialog had that problem before (see screenshots above), but _that_ layout is fine now:
> 
> image
> 
> However, it still does not properly disable "Save changes" on invalid times. I can click the button; nothing happens, and the dialog stays open, but I think it should be gray and light, and not clickable. This does not affect current master, as there is no free-form input and it is impossible to select an invalid time.

This has not changed from master actually. But I can defiintely set a smaller max-width.

KKoukiou:
@martinpitt ok I think I covered all your concerns above:
Here are screenshots from the dialog in multiple screensizes:


Screen Shot 2021-03-24 at 11 01 30
Screen Shot 2021-03-24 at 11 01 18


Screen Shot 2021-03-24 at 11 04 03
Screen Shot 2021-03-24 at 10 55 58

martinpitt (approved):
Nice work on the validation factoring! Indeed the layout is fine now. The auto-updates button still does not get disabled properly on invalid times, but that's "just" a minor cosmetical bug -- you can't actually configure wrong times any more, the button is inert then. We can fix that in a follow-up. Thank you!

KKoukiou:
@martinpitt and also we fixed on major flake here. https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&days=7&test=test%2Fverify%2Fcheck-system-shutdown-restart+TestShutdownRestart.testBasic

Issue #14840, opened 2020-10-30
services page got much slower to load between cockpit 217 and 218


AdamWill:
Cockpit version: 220
OS: Fedora
Page: Services

Sorry this report is late, but I was looking into Fedora openQA test failures on aarch64 and realized that the Cockpit test is quite often failing because the Services page takes too long to load. Looking at the test history, this seems to have started back in April when cockpit 218 landed.

When the test enters the page, a Loading spinner shows for a long time. Even on x86_64 this sometimes takes 20 seconds or more, but on aarch64 it often keeps showing for around two minutes, at which point the test times out. I just tested on my own x86_64 desktop and saw the spinner for several seconds when loading the page, looking at the JS console doesn't seem to show anything useful. System logs don't show anything relevant during the time the spinner is showing.

AdamWill:
I'm guessing this is related to the several changes @KKoukiou made to how services are parsed between 217 and 218.

AdamWill:
@paulwhalen

paulwhalen:
FWIW, on hardware (Raspberry Pi 4-4GB), the delay is about 4 seconds.

AdamWill:
Huh, odd that it takes so long in the VM. Did you test from a Server DVD install? That's how the openQA test runs.

paulwhalen:
This was a network installation.

AdamWill:
What package set? I'm thinking the behaviour may depend on how many / which services are present...

paulwhalen:
Just the default  "@^server-product-environment"

KKoukiou:
@AdamWill thanks for reposting this. Previously we were displaying all units as they were fetched from the API, now we wait until all are fetched and show the spinner in the meantime, then we display the list.
This justifies your comment that after 217 this page became slower, in terms of we show the spinner instead the list being updated.
However I really never encountered such long loading times as you mention. I will keep this issue as a reference, and try to find a way to make this page faster again.

AdamWill:
some logging might help debug it - for instance, it could log each service as it parses it, with a timestamp, so we could see if there is one or a few that are taking much longer. I could not find anything along these lines when I looked. It'd be ideal if it could be logged to the system journal (that makes it much easier to get out of an openQA test) but I could probably work with it being logged to the javascript console too.

martinpitt:
I guess this is still relevant (Services page has become slower), so keeping this open for now.

AdamWill:
[...]

PR #15531, opened 2021-03-16
Fix FMF tests for Fedora 34, enable in packit

martinpitt:
This was developed against my fork in https://github.com/martinpitt/cockpit/pull/13, with lots of intermediate/failing results (which justify the commits here).

### DIFF
--- a/test/run-verify-host.sh
+++ b/test/run-verify-host.sh
@@ -29,6 +29,11 @@ if rpm -q setroubleshoot-server; then
     dnf remove -y setroubleshoot-server
 fi
 
+if grep -q 'ID=.*fedora' /etc/os-release; then
+    # required by TestLogin.testBasic, but tcsh is not available in CentOS/RHEL
+    dnf install -y tcsh
+fi
+
 # make libpwquality less aggressive, so that our "foobar" password works
 printf 'dictcheck = 0\nminlen = 6\n' >> /etc/security/pwquality.conf
 
--- a/test/run-verify-host.sh
+++ b/test/run-verify-host.sh
@@ -23,6 +23,12 @@ if ! rpm -q chromium; then
     dnf install -y chromium
 fi
 
+# HACK: setroubleshoot-server crashes/times out randomly (breaking TestServices),
+# and is hard to disable as it does not use systemd
+if rpm -q setroubleshoot-server; then
+    dnf remove -y setroubleshoot-server
+fi
+
 # make libpwquality less aggressive, so that our "foobar" password works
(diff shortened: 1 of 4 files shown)


test/run-verify-host.sh
```diff
@@ -23,6 +23,12 @@ if ! rpm -q chromium; then
     dnf install -y chromium
 fi
 
+# HACK: setroubleshoot-server crashes/times out randomly (breaking TestServices),
+# and is hard to disable as it does not use systemd
+if rpm -q setroubleshoot-server; then
+    dnf remove -y setroubleshoot-server
```
martinpitt, line 34: I want to keep this for posterity, in case we revisit that in the future: I first tried a more pinpointed approach:

```diff
--- test/verify/check-system-services
+++ test/verify/check-system-services
@@ -39,6 +39,11 @@ class TestServices(MachineCase):
         # were removed and daemon reloaded.
         self.restore_dir("/etc/systemd/system", "systemctl daemon-reload && systemctl reset-failed", True)
 
+        # setroubleshoot-server crashes/times out randomly
+        self.machine.execute("systemctl mask --runtime dbus-:1.4-org.fedoraproject.Setroubleshootd@.service")
+        self.addCleanup(self.machine.execute, "systemctl unmask --runtime dbus-:1.4-org.fedoraproject.Setroubleshootd@.service")
+        self.machine.execute("systemctl stop dbus-:1.4-org.fedoraproject.Setroubleshootd@*.service")
+
     def pick_tab(self, n):
         self.browser.click('#services-filter li:nth-child({0}) a'.format(n))
```

but that did not work, these setroubleshoot instances still work. These are not "real" systemd units, they are made as transient units by dbus-broker on the fly. I suppose that just does not check for masked units and such.

martinpitt:
Nice, all green on the first try :+1:

test/verify/check-system-services
```diff
@@ -82,7 +82,8 @@ class TestServices(MachineCase):
         self.browser.wait_in_text(self.svc_sel(service), state_new)
 
     def wait_page_load(self):
-        self.browser.wait_not_present(".pf-c-empty-state .pf-c-spinner[aria-valuetext='Loading...']")
+        with self.browser.wait_timeout(180):
```
marusak, line 85: https://github.com/cockpit-project/cockpit/issues/14840

maybe?

marusak (approved):
Very nice, thanks!

test/verify/check-system-services
martinpitt: Could be, but hard to say. It's just consistently happening on F34, but fine on F33 and earlier.

PR #15532, opened 2021-03-16
test: add the check which is the dialog can be closed correctly

yunmingyang:
Related with: https://bugzilla.redhat.com/show_bug.cgi?id=1792379

### DIFF
--- a/test/verify/check-machines-disks
+++ b/test/verify/check-machines-disks
@@ -847,6 +847,34 @@ class TestMachinesDisks(VirtualMachinesCase):
         b.wait_not_present("#vm-subVmTest1-disks-vdd-device")
         b.wait_not_present(".pf-c-modal-box")
 
+        # Test detaching several disks and the deletion dialog can be closed correctly
+        m.execute("virsh vol-create-as myPoolOne diskVirtio --capacity 1M --format qcow2")
+        m.execute("virsh vol-create-as myPoolOne diskSata --capacity 1M --format qcow2")
+        wait(lambda: "diskVirtio" in m.execute("virsh vol-list myPoolOne"))
+        wait(lambda: "diskSata" in m.execute("virsh vol-list myPoolOne"))
+        m.execute("virsh attach-disk --domain subVmTest1 --source {0}/diskVirtio --target vdf --targetbus virtio --persistent".format(p1))
+        m.execute("virsh attach-disk --domain subVmTest1 --source {0}/diskSata --target sda --targetbus sata --persistent".format(p1))
+
+        # Need to refresh when attaching disk for shutoff VM
+        b.reload()
+        b.enter_page('/machines')
+        b.wait_in_text("body", "Virtual machines")
+        b.wait_visible("#vm-subVmTest1-disks-sda-device")
+        b.wait_visible("#vm-subVmTest1-disks-vdf-device")
+
+        def removeDiskAndChecks(target):
+            b.click("#delete-subVmTest1-disk-{}".format(target))
+            b.wait_in_text("div[role=dialog]:contains(\"Remove Disk\")", target)
+            b.click("div[role=dialog] button:contains(Remove)")
+            # The deletion dialog should be closed
+            b.wait_not_present("div[role=dialog]:contains(\"Remove Disk\")")
+            # No error shown
+            b.wait_not_present("div[aria-label=\"Danger Alert\"]")
+            b.wait_not_present("#vm-subVmTest1-disks-{}-device".format(target))
+
+        removeDiskAndChecks("sda")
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
Is there some fixed bug that encourages you to add this test? Can you add a link if so. 

And please make the selector a bit more readable. Ideally id without anything else or if itdoes not exist don't add styling related classes from pf4 in the selector.

test/verify/check-machines-disks
```diff
@@ -847,6 +847,35 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+        b.wait_visible("#vm-subVmTest1-disks-sda-device")
+        b.wait_visible("#vm-subVmTest1-disks-vdf-device")
+
+        def removeDiskAndChecks(target):
+            b.click("#delete-subVmTest1-disk-{}".format(target))
+            b.wait_visible("#app + div div.pf-c-modal-box.pf-m-align-top.pf-m-md")
+            b.wait_in_text("#app + div span.pf-c-modal-box__title-text", target)
+            b.click("#app + div button.pf-c-button.pf-m-danger.pf-m-progress")
+            # The deletion dialog should be closed
+            b.wait_not_present("#app + div div.pf-c-modal-box.pf-m-align-top.pf-m-md")
```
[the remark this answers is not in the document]
yunmingyang: Sure, In the issue, the dialog will not be closed, so in this case, I think we need to check whether the dialog is closed. So this selector is to check the deletion dialog. Though the dialog has a id, it is dynamic. So I choose "#app + div" for indicating the element we want to select is in the first <div> which is placed immediately after #app. And for div.pf-c-modal-box.pf-m-align-top.pf-m-md, I think I can change it to div[role="dialog"].

test/verify/check-machines-disks
```diff
@@ -847,6 +847,35 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+        b.enter_page('/machines')
+        b.wait_in_text("body", "Virtual machines")
+        b.wait_visible("#vm-subVmTest1-disks-sda-device")
+        b.wait_visible("#vm-subVmTest1-disks-vdf-device")
+
+        def removeDiskAndChecks(target):
+            b.click("#delete-subVmTest1-disk-{}".format(target))
+            b.wait_visible("#app + div div.pf-c-modal-box.pf-m-align-top.pf-m-md")
+            b.wait_in_text("#app + div span.pf-c-modal-box__title-text", target)
+            b.click("#app + div button.pf-c-button.pf-m-danger.pf-m-progress")
```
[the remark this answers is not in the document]
yunmingyang: #app + div div[role=dialog] button:contains(Remove)

yunmingyang:
@KKoukiou Sure, the issue link is https://bugzilla.redhat.com/show_bug.cgi?id=1792379.

test/verify/check-machines-disks
```diff
@@ -847,6 +847,35 @@ class TestMachinesDisks(VirtualMachinesCase):
...
+        b.reload()
+        b.enter_page('/machines')
+        b.wait_in_text("body", "Virtual machines")
+        b.wait_visible("#vm-subVmTest1-disks-sda-device")
+        b.wait_visible("#vm-subVmTest1-disks-vdf-device")
+
+        def removeDiskAndChecks(target):
+            b.click("#delete-subVmTest1-disk-{}".format(target))
+            b.wait_visible("#app + div div[role=dialog]")
+            b.wait_in_text("#app + div span", target)
```
KKoukiou, line 868: Please avoid using such selectors that contain 0 information for the developer about what element it's actually selecting. 
They have two main issue:

* really hard to read the test code
* if I change the code slightly to have another div this selector will break.

In this case you should rather add an #id in the js code to the item so that you can use that one. Same for the dialog element and the buttons.

KKoukiou: Sorry for not making it obvious in the first review round. I would really prefer to get this in a more readable state before merging.

yunmingyang: OK, but I would like to try to change the selector firstly. How do the new selectors look like?

KKoukiou: We will need to introduce something like this https://github.com/cockpit-project/cockpit/blob/master/pkg/machines/components/vm/disks/vmDisksCard.jsx#L289 for the resource modal component.

Then we can use  id={`delete-${objectId}-button`} for the remove button and  id={`delete-${objectId}-modal`} for the modal selector. 

I can also do it and push on your branch if you prefer.

KKoukiou (approved):
@yunmingyang you can send the changes we discussed on IRC as a separate PR. With the last push it look good enough from my side.

PR #15533, opened 2021-03-16
packit: Enable Fedora Rawhide

martinpitt:


martinpitt:
Erk, that was meant for my fork.

PR #15534, opened 2021-03-16
storage: Improve size slider text input

### LINKED ISSUE #15409: Grow logical volume - Unexpected type 'double' in argument

mvollmer:
We no longer convert the text input into a number immediately on every
change.  That did interfere severely with the user input.  For
example, it was impossible to type a decimal point "." at the end of
the text because it would immediately be removed again when converting
the text to a number and back again.

Instead, textual input is converted to a final number in the validate
function.

Fixes #15409

### DIFF
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -109,6 +109,9 @@
    The second argument has all values of all fields, in case you need
    to look at more than one field.
 
+   It is permissible to overwrite fields of "vals" to change the final
+   value of a field.
+
    The validate function can also return a promise which resolves to
    null or an error message.  If that promise is rejected, that error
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):

KKoukiou:
So much better from usability perspective!

PR #15535, opened 2021-03-16
pf4: Remove description list responsive override

garrett:
TODO:

- [ ] Find description lists in Cockpit that need a responsive override and add a pf-m-horizontal-on-sm (or similar) class

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -33,13 +33,6 @@ svg {
     vertical-align: -0.125em;
 }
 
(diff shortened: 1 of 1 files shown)


garrett:
Oh, we should use a standard PF variable, not 640px. I'm looking and will change this in a force-push.

Additionally, I think the summary I wrote here should be in the git commit itself.

KKoukiou (changes requested):
There is already an official PF way to do this. I think we should fix it my using their classes now.

garrett:
Updated PR to remove the section and add a note about using a class instead. 

Also: I converted it to draft, while we round up the parts of Cockpit that need to be adapted.

garrett:
I guess someone, someday will work on this. Perhaps.

garrett:
(But we have some other stuff to do right now...)

PR #15539, opened 2021-03-16
machines: fix alignment of radio buttons in 'add disk' dialog

KKoukiou:
Remove isInline from the "Persistence" group. This only makes sense to
use when more than one items are part of the group.


garrett:
Where do I go to test this? If it's the radio buttons that are visible in this screenshot, then it looks good to me. Otherwise, I'm not sure what else to test.

Screenshot_2021-03-16 Virtual Machines - garrett Rain

garrett (approved):
Looks good to me! Thanks!

Here's what it looks like when the machine is running, so the persistence option is visible:

Screenshot_2021-03-16 Virtual Machines - garrett Rain(1)

PR #15541, opened 2021-03-16

cockpituous:


po/LINGUAS
```diff
@@ -7,6 +7,7 @@ he
 it
 ja
 ko
+nb_NO
```
marusak, line 10: Hold your horses bots, this is not enough, it needs to be added into `pkg/shell/manifest.json` as well. Of course this locale does not exist in `po/language_map.txt`

marusak (changes requested):
Needs a bit of manual intervention. I'll take a look

po/LINGUAS
marusak: Fixed. I want to check preview once it is build as well though.

marusak:
I though we could mention new languages in release notes. And here we can also thank the author directly https://github.com/stabasm . Also a little sales pitch for weblate won't hurt I guess so more people know about it and go and improve translations :) 

Release note:

Cockpit is now also available in Norwegian Bokmål localization. Thank you Stein-Aksel Basma for this contribution!
Cockpit is being translated in [weblate](https://translate.fedoraproject.org/projects/cockpit/master/) where anyone can help to improve Cockpit.

marusak (approved):

PR #15545, opened 2021-03-17
fmf: Increase plan timeout and duration for Fedora

martinpitt:
In the European evenings/American mornings, the tests on Testing Farm
are really slow (supposedly, AWS spot instances with lots of noisy
neighbors?), have a hard time finishing in two hours, and lots of tests
fail because expensive actions (such as loading the Services page or the
Terminal torture test) actually take more than one minute.

Introduce a `$TEST_TIMEOUT_FACTOR` environment variable to globally
scale the timeouts, use it for FMF tests on Fedora, and bump the global
plan timeout to four hours.

[example from yesterday evening](http://artifacts.dev.testing-farm.io/b0212fd2-d876-457f-bdfd-d0cfded5b77a/)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -103,6 +103,9 @@ You can set these environment variables to configure the test suite:
     TEST_SHOW_BROWSER  Set to run browser interactively. When not specified,
                        browser is run in headless mode.
 
+    TEST_TIMEOUT_FACTOR Scale normal timeouts by given integer. Useful for
+                        slow/busy testbeds or architectures.
+
 ## Test machines and their images
 
 The code under test is executed in one or more dedicated virtual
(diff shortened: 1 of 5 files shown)


martinpitt:
Tested on https://github.com/martinpitt/cockpit/pull/14 where it now succeeded even for Rawhide.

martinpitt:
[provisioning error](http://artifacts.dev.testing-farm.io/097cfa09-b2e6-4747-9b0a-4f6a09500804/pipeline.log) for F34, [f33 worked fine](http://artifacts.dev.testing-farm.io/b8c09a6c-62c9-4ab4-8464-62e82c2f2ba7/). Restarting.

martinpitt:
LOL, seems we now broke TF -- since this morning, all tests fail with " Guest couldn't be provisioned: Condition 'ip_ready' failed to pass within given time".

marusak (approved):
uh oh, up to 3 times slower?
But looks good, thanks! Lets see if that helps :crossed_fingers:

martinpitt:
@marusak: Well, it's at least "more than two", as it usually takes less than an hour, and was timing out after two. I think there's also something wrong with F34 -- F33 is regularly "fast", while F34 tests run much longer. But let's not make that break the tests, we can collect some more data first.

PR #15553, opened 2021-03-18
Fix coverity issues

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=1938693

 - [x] do a [final pass in coverity](https://cov01.lab.eng.brq.redhat.com/covscanhub/task/209064/)

### DIFF
--- a/src/bridge/cockpitconnect.c
+++ b/src/bridge/cockpitconnect.c
@@ -395,35 +395,27 @@ GSocketAddress *
 cockpit_connect_parse_address (CockpitChannel *channel,
                                gchar **possible_name)
 {
-  GSocketConnectable *connectable;
-  GSocketAddressEnumerator *enumerator;
   GSocketAddress *address;
-  GError *error = NULL;
-  gchar *name = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree gchar *name = NULL;
 
-  connectable = parse_address (channel, &name, NULL);
+  g_autoptr(GSocketConnectable)  connectable = parse_address (channel, &name, NULL);
   if (!connectable)
     return NULL;
 
   /* This is sync, but realistically, it doesn't matter for current use cases */
-  enumerator = g_socket_connectable_enumerate (connectable);
-  g_object_unref (connectable);
+  g_autoptr(GSocketAddressEnumerator) enumerator = g_socket_connectable_enumerate (connectable);
 
   address = g_socket_address_enumerator_next (enumerator, NULL, &error);
-  g_object_unref (enumerator);
 
   if (error != NULL)
     {
       cockpit_channel_fail (channel, "not-found", "couldn't find address: %s: %s", name, error->message);
-      g_error_free (error);
-      g_free (name);
       return NULL;
     }
 
   if (possible_name)
-    *possible_name = name;
-  else
-    g_free (name);
+    *possible_name = g_steal_pointer (&name);
 
   return address;
 }
@@ -734,7 +726,10 @@ cockpit_connect_parse_stream (CockpitChannel *channel)
 
   address = parse_address (channel, &name, &local);
   if (!address)
-    return NULL;
+    {
+      g_free (name);
+      return NULL;
+    }
 
   connectable = g_new0 (CockpitConnectable, 1);
   connectable->address = address;
--- a/src/common/test-jsonfds.c
+++ b/src/common/test-jsonfds.c
@@ -422,6 +422,7 @@ test_memfd_error_cases (void)
 
       content = cockpit_memfd_read (fd, &error);
       cockpit_assert_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_INVAL, "*not memfd?*");
+      g_assert (content == NULL);
       g_clear_error (&error);
       close (fd);
 
--- a/src/ws/cockpitcertificate.c
+++ b/src/ws/cockpitcertificate.c
@@ -56,11 +56,10 @@ static gchar *
 get_machine_id (void)
 {
   static const char HEX[] = "0123456789abcdef";
-  gchar *content;
   gchar *machine_id = NULL;
 
-  if (g_file_get_contents ("/etc/machine-id", &content, NULL, NULL))
-    machine_id = g_strstrip (g_strcanon (content, HEX, ' '));
+  if (g_file_get_contents ("/etc/machine-id", &machine_id, NULL, NULL))
+    g_strstrip (g_strcanon (machine_id, HEX, ' ')); /* both functions modify machine_id in-place */
 
   return machine_id;
 }
--- a/src/ws/cockpitwebservice.c
+++ b/src/ws/cockpitwebservice.c
@@ -1093,21 +1093,18 @@ on_web_socket_message (WebSocketConnection *connection,
                        CockpitWebService *self)
 {
   CockpitSocket *socket;
-  GBytes *payload;
-  gchar *channel;
+  g_autofree gchar *channel = NULL;
 
   socket = cockpit_socket_lookup_by_connection (&self->sockets, connection);
   g_return_if_fail (socket != NULL);
 
-  payload = cockpit_transport_parse_frame (message, &channel);
+  g_autoptr(GBytes) payload = cockpit_transport_parse_frame (message, &channel);
   if (!payload)
     return;
 
   /* A control channel command */
   if (!channel)
-    {
-      dispatch_inbound_command (self, socket, payload);
-    }
+    dispatch_inbound_command (self, socket, payload);
 
   /* An actual payload message */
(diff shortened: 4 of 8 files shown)


martinpitt:
I now learned how to run coverity scans on a local srpm, and I have some further adjustments to make.

martinpitt:
Ah, sad: gcc in our unit-tests container does not yet know about this:

    error: unknown warning option '-Werror=misleading-indentation'; did you mean '-Werror=binding-in-condition'? [-Werror,-Wunknown-warning-option]

so reverting this bit.

martinpitt:
Coverity run finished, see https://bugzilla.redhat.com/show_bug.cgi?id=1938693#c7 -- this is about as clean as we can reasonably get it now.

src/bridge/cockpitconnect.c
```diff
@@ -403,7 +403,10 @@ cockpit_connect_parse_address (CockpitChannel *channel,
 
   connectable = parse_address (channel, &name, NULL);
   if (!connectable)
-    return NULL;
+    {
+      g_free (name);
```
allisonkarlitskaya, line 407: ...or drop the other three `free()` in this function and make it a `g_autofree`.  Could do the same with the `GError` nearby...

src/ws/test-kerberos.c
```diff
@@ -403,9 +403,11 @@ mock_kdc_start (void)
       ret = read (out_fd, input->str + len, 256);
       if (ret < 0)
         {
-          if (errno != EAGAIN && errno != EINTR)
-            g_error ("couldn't read from mock-kdc: %s", g_strerror (errno));
-            break;
```
allisonkarlitskaya, line 408: Oof.  That's ugly.

src/ws/test-kerberos.c
```diff
@@ -403,9 +403,11 @@ mock_kdc_start (void)
...
+          if (errno == EAGAIN || errno == EINTR)
+            continue;
```
allisonkarlitskaya, line 407: I don't like this restructuring.  Please either keep the original structure (with added braces), or do your approach, but then remove the non-error case from the `else` (and unindent it).  There's no sense in having continue; and break; before you effectively hit the end of the body of the loop (because you skip the else)...

src/ws/session-utils.c
```diff
@@ -659,7 +659,10 @@ btmp_log (const char *username,
     .ut_type = LOGIN_PROCESS,
   };
 
+  /* see utmp(5), it is ok to not null-terminate these if they have maximum size */
+  /* coverity[buffer_size_warning : FALSE] */
```
allisonkarlitskaya, line 663: This is an obnoxious bug in Coverity.  I'd expect to see this for lastlog, but not for utmp.  Or maybe the glibc on the testing machine is just old?

See also https://sourceware.org/bugzilla/show_bug.cgi?id=24899 and https://sourceware.org/bugzilla/show_bug.cgi?id=25844

src/ws/cockpitwebservice.c
```diff
@@ -1093,31 +1093,25 @@ on_web_socket_message (WebSocketConnection *connection,
...
   CockpitSocket *socket;
-  GBytes *payload;
-  gchar *channel;
+  g_autofree gchar *channel = NULL;
 
   socket = cockpit_socket_lookup_by_connection (&self->sockets, connection);
   g_return_if_fail (socket != NULL);
 
-  payload = cockpit_transport_parse_frame (message, &channel);
+  g_autoptr(GBytes) payload = cockpit_transport_parse_frame (message, &channel);
```
allisonkarlitskaya, line 1101: go go go!

src/ws/cockpitcertificate.c
```diff
@@ -56,11 +56,10 @@ static gchar *
 get_machine_id (void)
 {
   static const char HEX[] = "0123456789abcdef";
-  gchar *content;
   gchar *machine_id = NULL;
 
-  if (g_file_get_contents ("/etc/machine-id", &content, NULL, NULL))
-    machine_id = g_strstrip (g_strcanon (content, HEX, ' '));
+  if (g_file_get_contents ("/etc/machine-id", &machine_id, NULL, NULL))
+    g_strstrip (g_strcanon (machine_id, HEX, ' '));
```
allisonkarlitskaya, line 62: I hate these functions.  Maybe add a comment to the effect of "modifies and moves in-place"?

src/common/test-jsonfds.c
```diff
@@ -410,7 +410,7 @@ static void
 test_memfd_error_cases (void)
 {
   g_autoptr(GError) error = NULL;
-  g_autofree gchar *content = NULL;
```
allisonkarlitskaya, line 413: I don't like this one.

It popped up last time I tested this code and I ignored it because I consider it to be conceptually wrong.  By static typing, this is a string which should be freed!  Consider if someone copies this code as an example...

I wonder if your added assert will be enough to satisfy Coverity....  I'd prefer to keep the `g_autofree` in place.

src/common/test-jsonfds.c
```diff
@@ -422,10 +422,10 @@ test_memfd_error_cases (void)
 
       content = cockpit_memfd_read (fd, &error);
       cockpit_assert_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_INVAL, "*not memfd?*");
+      g_assert (content == NULL);
       g_clear_error (&error);
       close (fd);
 
-
```
allisonkarlitskaya, line 428: I was using the extra space to separate conceptually separate independent "cases".

allisonkarlitskaya (changes requested):

src/bridge/cockpitconnect.c
martinpitt: That was the first thing I thought of, of course. But that doesn't work, as `name` is passed to the caller as output argument through `gchar **possible_name` in the success case. The alternative would be to sacrifice efficiency for readability, use g_auto, and dup the string.

src/ws/test-kerberos.c
martinpitt: The original structure (with added braces) made no sense. It would terminate the read loop on EAGAIN/EINTR, which is exactly what these error codes *don't* as you to do.

Indeed the `break` is redundant as `g_error()` is fatal, but I suppose that's for the sake of readability or so? But indeed I agree to dropping the else, that's easier to read.

src/ws/session-utils.c
martinpitt: They are running on RHEL 8, with glibc 2.28. The fix only landed in 2.30. Thanks for the reference, I added it as a comment.

src/ws/cockpitcertificate.c
martinpitt: good idea, done.

src/common/test-jsonfds.c
martinpitt: Ack, I put it back, and I'll do another coverity run to validate it. If it complains again, I can add an annotation.

src/common/test-jsonfds.c
martinpitt: ok, I put them back.

src/bridge/cockpitconnect.c
allisonkarlitskaya: That's exactly one of the `g_steal_pointer()` example cases outlined here: https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-steal-pointer

src/ws/test-kerberos.c
allisonkarlitskaya: Looks nicer, thanks.

The further handling of the `ret == 0` case below is a bit off-putting, but let's not go too wild here...

allisonkarlitskaya (changes requested):
Everything looks good now except the `g_steal_pointer()` thing.

src/bridge/cockpitconnect.c
martinpitt: TIL! I rewrote the function to use g_automagic for everything.

src/bridge/cockpitconnect.c
```diff
@@ -395,35 +395,27 @@ GSocketAddress *
 cockpit_connect_parse_address (CockpitChannel *channel,
                                gchar **possible_name)
 {
-  GSocketConnectable *connectable;
```
allisonkarlitskaya, line 398: I see an awful lot of red and I like it.  :+1:

allisonkarlitskaya (approved):

src/common/test-jsonfds.c
martinpitt: No more test-jsonfds.c complaints in the coverity run with this approved PR, so this is good.

martinpitt:
Coverity finished, and looks good, unblocking. Still waiting for tests.

PR #15557, opened 2021-03-18
lib/console: Force width to containment size

### LINKED ISSUE #15472: VM console does not fit onto screen
Cockpit version: cockpit-238.1-1.fc32.s390x
OS: Fedrao 32
Page: Virtual machines

Hello, when I launch a new VM, overview pane changes a bit and displays "CPU Type" field which is longer than usual and t his pushes the Console pane to the right so it's not visible on a reasonably-sized browser window. I tested on 1440x900.

I have Czech locale and chances are the translation is simply too wide, I wanted to test this with English but I am unable to switch to English - when I select the language and confirm and logout/login, I still see Czech for some reason. Maybe a bug?

Anyway, this is S390x where processor type is also long string: "z14ZR1-base", in total it does not fit. Small issue.

Cockpit S390x

garrett:
The console likes to hardcode its width, causing layouts to break.

Forcing it to 100% makes it render within the boundaries.

Fixes #15472

### DIFF
--- a/pkg/lib/console.css
+++ b/pkg/lib/console.css
@@ -55,3 +55,8 @@
         background: var(--ct-color-text);
     }
 }
+
+/* Ensure the console fits to its container (and doesn't attempt to go beyond the limits) */
+.xterm-screen {
+    width: 100% !important;
+}


garrett:
I cannot properly test this locally. Please ensure it works with normal text in the terminal without side effects.

Also, if you'd like to fix this in another way, please do so. This is a bit of a dirty hack since it uses `!important` to override what JavaScript specifies for the width (in pixels), rather than digging into the code to more properly fix it.

martinpitt:
FTR, the packit/fedora-34 failure here is because garrett's branch is slightly behind master and does not yet contain commit b6734a025f2e5ab281c47dd0b746289830c408af . It's ok to ignore here.

KKoukiou (approved):
Works for me on all screen sizes. Unless @marusak has something against it I am happy.

marusak:
I don't :) My solution would be something on the same line so lets not dwell on it if it works.

PR #15560, opened 2021-03-19
workflows: Run npm-update twice a week only

martinpitt:
This is what we used to do when we still used npm-trigger. Some modules
like react-console are updated very often, and validating these
becomes too expensive/annoying.


marusak (approved):
Thanks!

PR #15561, opened 2021-03-19
test-static-code: test more python files

allisonkarlitskaya:
The recent changes to FMF (which included adjusting the exclude pattern
for the python static code checks) triggered an interesting (yet
very understandable) limitation in git: if you create the pre-push hook
in the recommended way:

  $ ln -s ../../tools/test-static-code .git/hooks/pre-push

and then use worktrees, then git will always run the check out of the
primary worktree, instead of the one you're working in.

That led to a bunch of newly-introduced files not being properly ignored
during the pycodestyle checks and a bunch of errors being thrown.

Upon looking into this a bit closer, I noticed that the way we're
invoking pycodestyle will cause it to check all the files immediately in
test/*, plus all files in subdiretcories with filenames of *.py, but to
skip everything else, including all of the test/verify/ scripts.

Here's a proposal for how we can improve the detection logic for which
scripts to scan, but because we went so long without checking anything
in test/verify/* and tools/*, there is an awful lot of cleanup to do
before we can merge this.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -125,11 +125,11 @@ def finish_test(opts, test, affected_tests):
             return None, 0
 
     # do we get a specific retry reason from tests-policy?
-    m = re.search(b"\s*# RETRY (.*)$", test.output, re.MULTILINE)
+    m = re.search(rb"\s*# RETRY (.*)$", test.output, re.MULTILINE)
     if m:
         retry_reason = m.group(1)
         # remove it from test output; we must not print it after the 3rd time, and going to print it separately
-        test.output = re.sub(b"\s*# RETRY .*\\n", b"", test.output)
+        test.output = re.sub(rb"\s*# RETRY .*\n", b"", test.output)
     elif affected: # Don't retry affected failed tests
         retry_reason = None
     else:
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -130,7 +130,7 @@ def finish_test(opts, test, affected_tests):
         retry_reason = m.group(1)
         # remove it from test output; we must not print it after the 3rd time, and going to print it separately
         test.output = re.sub(rb"\s*# RETRY .*\n", b"", test.output)
-    elif affected: # Don't retry affected failed tests
+    elif affected:  # Don't retry affected failed tests
         retry_reason = None
     else:
         # HACK: many tests are unstable, always retry them 3 times
@@ -318,7 +318,7 @@ def run(opts, image):
                 web_address = "{0}:{1}".format(m.machine.web_address,
                                                m.machine.web_port)
 
-                if i == opts.batches - 1: # Last machine needs to resolve the rest
+                if i == opts.batches - 1:  # Last machine needs to resolve the rest
                     batch = serial_tests[batch_size * i:]
                 else:
                     batch = serial_tests[batch_size * i: batch_size * i + batch_size]
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -41,11 +41,11 @@ class Test:
 
 def test_name(test):
     return "{0} {1} {2}{3}".format(
-            test.test_id,
-            test.command[0],
-            test.command[-1],
-            " [ND@{0}]".format(test.serial_machine) if test.nondestructive else "",
-            )
+        test.test_id,
+        test.command[0],
+        test.command[-1],
+        " [ND@{0}]".format(test.serial_machine) if test.nondestructive else "",
+    )
 
 
 def flush_stdout():
@@ -74,16 +74,17 @@ def print_test(test, print_tap=True, retry_reason=""):
 
     print()  # Tap needs to start on a separate line
     if test.process.returncode == 0:
-        print("ok " +  test_name(test) + retry_reason)
+        print("ok " + test_name(test) + retry_reason)
     elif test.process.returncode == 77 or b"# SKIP " in test.output:
         # If the test was skipped, add the last line (which contains the reason
         # for the skip) to the result
         print("ok {0} {1}{2}".format(test_name(test),
-                                  test.output.splitlines()[-1].strip().decode() if test.process.returncode == 77 else "",
-                                  retry_reason))
+                                     test.output.splitlines()[-1].strip().decode() if test.process.returncode == 77 else "",
+                                     retry_reason))
     else:
         print("not ok " + test_name(test) + retry_reason)
     flush_stdout()
+
 
 def finish_test(opts, test, affected_tests):
     """Returns if a test should retry or not
@@ -145,12 +146,14 @@ def finish_test(opts, test, affected_tests):
         print_test(test, print_tap=opts.thorough)
     return None, 1
 
+
 def check_valid(filename):
     name = os.path.basename(filename)
     allowed = string.ascii_letters + string.digits + '-_'
     if not all(c in allowed for c in name):
         return None
     return name.replace("-", "_")
+
 
 def build_command(filename, test, opts):
     cmd = [filename]
@@ -164,6 +167,7 @@ def build_command(filename, test, opts):
         cmd.append("-l")
     cmd.append(test)
     return cmd
+
 
 class GlobalMachine:
     def __init__(self, restrict=True, cpus=None, memory_mb=None):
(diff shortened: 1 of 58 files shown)


martinpitt:
Ugh, nice finding! :+1:  on this approach, but indeed a lot of fallout - hopefully most of it can be fixed automatically.

test/containers/check-bastion
```diff
@@ -18,17 +18,17 @@
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
 
+from testvm import VirtMachine
+from testlib import *
+from verify import parent
 import os
```
martinpitt: I don't like this (the whole "python files: fix header include order" commit). I really prefer builtin modules at the top, then third-party modules, and finally the project-local module imports.

Also, this order does not really make sense -- if you could already import testlib and verify here, then there's no need to fiddle with sys.path later and  importing `parent`.

test/verify/check-machines-create
```diff
@@ -689,7 +689,7 @@ class TestMachinesCreate(VirtualMachinesCase):
                 query_result = '{0}'.format(self.os_name)
                 # throws exception if grep fails
                 self.machine.execute(
-                    "osinfo-query os --fields=name | tail -n +3 | sed -e 's/\s*|\s*/|/g; s/^\s*//g; s/\s*$//g' | grep '{0}'".format(
+                    "osinfo-query os --fields=name | tail -n +3 | sed -e 's/\\s*|\\s*/|/g; s/^\\s*//g; s/\\s*$//g' | grep '{0}'".format(
```
martinpitt, line 692: Let's also use a raw string here instead of these unsightly `\\`.

test/verify/check-machines-create
```diff
@@ -755,7 +755,7 @@ class TestMachinesCreate(VirtualMachinesCase):
             self.browser.wait_text("#vm-{}-disks-sda-device".format(self.name), "cdrom")
             self.goToMainPage()
 
-            virt_install_cmd = "ps aux | grep 'virt\-install\ \-\-connect'"
+            virt_install_cmd = "ps aux | grep '[v]irt-install --connect'"
```
martinpitt, line 758: That `[v]` needs an explanation. How is that any different from just `v`?

test/verify/check-system-s4u-ssh
```diff
@@ -108,7 +108,7 @@ session    include      system-auth
 """)
 
         sshd_machine.execute(script="""#!/bin/sh
-        sed -i '/\[domain\/cockpit.lan\]/ a pam_gssapi_services = sudo, sudo-i' /etc/sssd/sssd.conf
+        sed -i '/\\[domain\\/cockpit.lan\\]/ a pam_gssapi_services = sudo, sudo-i' /etc/sssd/sssd.conf
```
martinpitt, line 111: Let's use a raw string here.  This is also not consistent, 5 lines down in the `foobarfoo\nfoobarfoo\n..` are still single quotes.

test/verify/check-system-terminal
```diff
@@ -34,8 +34,8 @@ class TestTerminal(MachineCase):
 
         # Make sure we get what we expect
         self.write_file("/home/admin/.bashrc", """
-PS1="\\u@\\h \\W]\$ "
-PROMPT_COMMAND='printf "\\033]0;%s@%s:%s\\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"'
+PS1="\\u@\\h \\W]\\$ "
+PROMPT_COMMAND='printf "\\033]0;%s@%s:%s\\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\\~}"'
```
martinpitt, line 38: This would also look much nicer with a raw string.

test/verify/check-testlib
```diff
@@ -174,12 +174,12 @@ class TestRunTest(MachineCase):
         process = subprocess.run([run_tests, "--test-dir", VERIFY_DIR, "--list", "TestHostEditing.testLimits",
                                   "TestHostSwitching.testEdit"], env=env, capture_output=True)
         stdout = process.stdout
-        self.assertRegex(stdout, b"\nTestHostEditing.testLimits # RETRY 1 \(test affected tests 3 times\)\n")
-        self.assertRegex(stdout, b"\nTestHostEditing.testLimits # RETRY 2 \(test affected tests 3 times\)\n")
-        self.assertRegex(stdout, b"\nTestHostEditing.testLimits\n")
-        self.assertNotRegex(stdout, b"RETRY 3")
-        self.assertRegex(stdout, b"\nTestHostSwitching.testEdit\n")
-        self.assertNotRegex(stdout, b"TestHostSwitching.testEdit # RETRY")
+        self.assertIn(stdout, b"\nTestHostEditing.testLimits # RETRY 1 (test affected tests 3 times)\n")
```
martinpitt, line 177: Wrong argument order -- assertIn has parameters (needle, haystack), while assertRegex is the other way round.

I'm actually rather shocked that the test succeeded anyway -- how can that be??

test/verify/machineslib.py
```diff
@@ -98,7 +98,7 @@ def startLibvirt(self):
         # Wait for the network 'default' to become active
         m.execute("virsh net-define /etc/libvirt/qemu/networks/default.xml || true")
         m.execute("virsh net-start default || true")
-        m.execute("until virsh net-info default | grep 'Active:\s*yes'; do sleep 1; done")
+        m.execute("until virsh net-info default | grep 'Active:\\s*yes'; do sleep 1; done")
```
martinpitt, line 101: raw string instead of \\

tools/test-static-code
```diff
@@ -124,7 +124,7 @@ else
     #   - executable matching ^#!/usr/bin/python3
     PYFILES="$(find test/ tools/ -type f '(' -name '*.py' -o -executable -execdir grep -q '^#!/usr/bin/python3' '{}' ';' ')'  -print)"
     # FIXME: Fix code for the warnings and re-enable them
-    out=$(python3 -m "$PYTHON_STYLE_CHECKER" --ignore E501,E261,W504 $PYFILES) || true
+    out=$(python3 -m "$PYTHON_STYLE_CHECKER" --max-line-length=415 --ignore E261,W504 $PYFILES) || true
```
martinpitt: FTR, I agree to moving to `--max-line-length`, this makes more sense. We can then try to reduce this to at least 150 or so (in a follow-up)

test/containers/check-bastion
```diff
@@ -18,17 +18,17 @@
...
-from testvm import VirtMachine
-from testlib import *
-from verify import parent
 import os
 import sys
 
 test_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 sys.path.append(test_dir)
 
+from verify import parent
```
martinpitt, line 27: ... looking at "Revert "python files: fix header include order"", that now explains how tests could actually work. :grin: Please eliminate this commit pair.

martinpitt (changes requested):
Thanks! Let's get this in ASAP, as this will conflict with pretty much every other landed PR.

test/verify/check-machines-create
allisonkarlitskaya: I thought this would have fallen under the "usual tricks" category: this is to avoid `grep` finding itself in the output of `ps`.  The `[v]` approach is even the more idiomatic one...

allisonkarlitskaya: I'll add a comment

test/verify/check-machines-create
allisonkarlitskaya: I have an irrational tendency to use raw strings only for regexps, (I used to think the "r" stood for regexp :rofl:) but there's actually no reason not to use them here as well...

test/verify/check-system-s4u-ssh
allisonkarlitskaya: This is kinda funny: we have echo -ne to get the `\n` expanded, but they actually never even made it that far, since Python did it for us....

Using a raw string now, and it will do what was clearly originally intended here.

martinpitt:
```
FAIL: tools/test-static-code 7 pep8 test
test/containers/run-tests:37:1: E302 expected 2 blank lines, found 1
ERROR: tools/test-static-code process failed: 1
```

martinpitt (approved):
LGTM now, thanks!

test/common/run-tests
```diff
@@ -125,11 +125,11 @@ def finish_test(opts, test, affected_tests):
             return None, 0
 
     # do we get a specific retry reason from tests-policy?
-    m = re.search(b"\s*# RETRY (.*)$", test.output, re.MULTILINE)
+    m = re.search(rb"\s*# RETRY (.*)$", test.output, re.MULTILINE)
     if m:
         retry_reason = m.group(1)
         # remove it from test output; we must not print it after the 3rd time, and going to print it separately
-        test.output = re.sub(b"\s*# RETRY .*\\n", b"", test.output)
+        test.output = re.sub(rb"\s*# RETRY .*\\n", b"", test.output)
```
allisonkarlitskaya, line 132: `\\n` → `\n`

martinpitt (approved):

PR #15562, opened 2021-03-19
systemd: split out ipa certificate script

allisonkarlitskaya:
Stop building a giant blob of shell from Javascript.  Install the script
in /usr/libexec instead, and execute it from there.

Right now, this is a straight refactor, but it's the start of what will
become a more generally-useful utility for manipulating certificates in
cockpit.

RFC?


martinpitt:
I generally like the approach, it goes into the direction of cleaning up our messy certificate handling. This introduces a dependency of cockpit-system to cockpit-ws on the same machine for this functionality. It needs careful checking if that would break the use case of joining a remote host. My initial feeling is "no" -- if that remote host only has cockpit-{system,bridge} installed, but not -ws, then you could join the machine to a realm, but not set up the -ws certificate from IPA (as there is no ws). But I suppose we should create a test case for this first to make sure we don't introduce subtle breakage.

Thanks!

allisonkarlitskaya:
> This introduces a dependency of cockpit-system to cockpit-ws on the same machine for this functionality. It needs careful checking if that would break the use case of joining a remote host. My initial feeling is "no" -- if that remote host only has cockpit-{system,bridge} installed, but not -ws, then you could join the machine to a realm, but not set up the -ws certificate from IPA (as there is no ws). But I suppose we should create a test case for this first to make sure we don't introduce subtle breakage.


@pitti: right at the top of this function:

```
    function install_ws_credentials() {
        // skip this on remote ssh hosts, only set up ws hosts
        if (cockpit.transport.host !== "localhost")
            return cockpit.resolve();
```

src/tls/cockpit-certificate-helper
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh -eu
```
martinpitt, line 1: nitpick: Please use `set -eu` on the second line, so that this stays around if you call the script through e.g. `sh -x helper` (which is fairly common for development)

pkg/systemd/overview-cards/realmd-operation.js
```diff
@@ -399,46 +399,10 @@ function instance(realmd, mode, realm, state) {
...
-                  '    cat /run/cockpit/ipa.key  >> /etc/cockpit/ws-certs.d/10-ipa.cert; ' +
-                  '    rm -f /run/cockpit/ipa.key; ' +
-                  'fi; ';
-
-        // use a temporary keytab to avoid interfering with the system one
-        var proc = cockpit.script(script, [], {
-            superuser: "require", err: "message",
-            environ: ["KRB5CCNAME=/run/cockpit/keytab-setup"]
-        });
+        var proc = cockpit.spawn([helper, "ipa", kerberos.RealmName, user],
```
martinpitt, line 404: I wonder if it should be `ipa create` or `ipa-create` right away? We probably want to treat this as a top-level mode, and add local certificate generation as well soon?

src/tls/cockpit-certificate-helper
```diff
@@ -0,0 +1,43 @@
...
+klist -k /etc/cockpit/krb5.keytab | grep -qF "${SERVICE}" || ipa-getkeytab -p "HTTP/${HOST}" -k /etc/cockpit/krb5.keytab
+
+# request an SSL certificate; be sure to not leave traces of the .key on disk or
+# get race conditions with file permissions; also, ipa-getcert
+# cannot directly write into /etc/cockpit due to SELinux
+if ipa-getcert request -f /run/cockpit/ipa.crt -k /run/cockpit/ipa.key -K "HTTP/${HOST}" -w -v; then
+    mv /run/cockpit/ipa.crt /etc/cockpit/ws-certs.d/10-ipa.cert
+    cat /run/cockpit/ipa.key  >> /etc/cockpit/ws-certs.d/10-ipa.cert
+    rm -f /run/cockpit/ipa.key
+fi
```
martinpitt, line 43: Bah, I'm so looking forward to dropping this crap :grin:

martinpitt (changes requested):
Thanks! This test case failure in check-testlib looks really suspicious..

pkg/systemd/overview-cards/realmd-operation.js
allisonkarlitskaya: My feeling was to let `ipa` be the "verb".  If we want a verb that's actually a verb, I guess I'd prefer to model it after `ipa-getcert request`, which is the main thing going on here... maybe `ipa-request` or something (with the "cert" now in the name of the binary)?

martinpitt (approved):
Danke!

PR #15565, opened 2021-03-21
tools: Don't delete manifest.json in webpack watch mode

martinpitt:
Commit b82b22abef679 caused webpack-watch to always delete
manifest.json, which broke efficient local development pretty hard.

Leave the manifest.json alone, they actually *are* good enough for
building RPMs now, as these are just static file copes. Still keep the
cleanup of Makefile.deps though, until webpack-make gets cleaned up more
thoroughly.

### DIFF
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -89,11 +89,8 @@ function process_result(err, stats) {
             // Force "make" to re-create everything.  The results of
             // incremental building are not good enough for building
             // RPMs, for example.
-            var stamp = makefile.split("/").slice(0, -1).concat(["manifest.json"]).join("/");
             if (fs.existsSync(makefile))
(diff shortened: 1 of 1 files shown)


martinpitt:
@garrett : You stumbled over this last week. This should fix your (and everyone's really!) woes.

allisonkarlitskaya (approved):
This fix is very obviously correct, so let's land it.

Are we sure that we need to remove the Makefile as well?  I think we've already arrived at the spot where the results are good enough.  Does anyone know otherwise?

martinpitt:
@allisonkarlitskaya : I'm not at all sure about this -- it could very well be that Makefile.deps is alright now; I just wanted this to be minimally invasive to not overly disturb your open "clean up build system" PRs.

PR #15567, opened 2021-03-22
workflows: Regularly prune -dist repository

martinpitt:
This avoids unbounded growth of the -dist repository. This force-pushes,
so is prone to race with parallel publish-dist jobs. So run this at
Sunday night, where the probability of human PRs is very low; no other
workflows run at this time either.

-----

I tested this on my fork. Before:
```
$ git clone https://github.com/martinpitt/cockpit-dist c
$ du -hs c/.git; du -hs c
314M	c/.git
562M	c
```

Then in [ran the workflow](https://github.com/martinpitt/cockpit/actions/runs/675122930), and now it [shows the truncated history](https://github.com/martinpitt/cockpit-dist/commits/master) and is smaller:
```
$ git clone https://github.com/martinpitt/cockpit-dist c2
$ du -hs c2/.git; du -hs c2
281M	c2/.git
529M	c2
```
I expect that the effect will be a lot more dramatic on the real cockpit repo.

### DIFF
--- a/.github/workflows/prune-dist.yml
+++ b/.github/workflows/prune-dist.yml
@@ -1,0 +1,39 @@
+# truncate the cockpit-dist history every Sunday night, to avoid unbounded growth
+name: prune-dist
+on:
+  schedule:
+    - cron: '0 1 * * 0'
+  # can be run manually on https://github.com/cockpit-project/cockpit/actions
+  workflow_dispatch:
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up configuration and secrets
+        run: |
+          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
+          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
+          # we push to -dist repo via https://github.com, that needs our cockpituous token
+          git config --global credential.helper store
+          echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
+
+      - name: Clone -dist repo
+        run: |
+          git clone https://github.com/${{ github.repository }}-dist.git dist-repo
+
+      - name: Truncate history of -dist repo
+        run: |
+          set -ex
+          cd dist-repo
+          # oldest commit that applies to any still present tarball
+          REF=$(git log --pretty=format:%H * | tail -n1)
+
+          git checkout --orphan temp $REF
+          git commit -m "Truncated history"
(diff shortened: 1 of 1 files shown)


.github/workflows/prune-dist.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
+          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
+          # we push to -dist repo via https://github.com, that needs our cockpituous token
+          git config --global credential.helper store
+          echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
+
+      - name: Clone -dist repo
+        run: |
+          git clone https://github.com/${{ github.repository }}-dist.git dist-repo
```
marusak, line 23: Ehm, why don't we run this directly in `https://github.com/${{ github.repository }}-dist.git dist-repo`?
We don't want to set up hooks there?

martinpitt: I suppose we could, but I thought it'd be nice to be able to blow away the repo and start from scratch. The publish-dist workflow can deal with an empty repo. But I'm happy to move it to -dist as well.

martinpitt: I.e. this is particularly helpful if you work on your cockpit fork -- then you don't have to actually fork the cockpit-dist repo as well (which is usually huge), and just start your own forked -dist from scratch.

.github/workflows/prune-dist.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Truncate history of -dist repo
+        run: |
+          set -ex
+          cd dist-repo
+          # oldest commit that applies to any still present tarball
+          REF=$(git log --pretty=format:%H * | tail -n1)
```
marusak, line 30: I don't think that `*` is necessary there or is it?

marusak (approved):
Makes sense, thanks!

.github/workflows/prune-dist.yml
martinpitt: It is -- without it, you'd just get the *entire* log, and thus REF would always be the first-ever commit, and you'd not prune anything. The `*` expands to all currently "visible" tar files in the branch -- publish-dist already regularly `git rm`'s the ones older than 7 days. So if you do `git log 1234.tar | tail`, you'll get the oldest commit that introduced 1234; and if you do it for all tarballs with `*`, you'll get the oldest commit of all the current tarballs.

PR #15569, opened 2021-03-22
Kpatch integration

marusak:
Note about commits:
Includes a few unrelated commits, I'll split them into separate PR once I get test runs on them. Almost certain at least one of them breaks tests but too many tests might be affected so I'll let bots do the dirty work.

Notes:
- I would really like to test this on real patches. To my knowledge there are no patches yet for 8.4, but I'll check it out again.
- Tests are not as I would like them - I would prefer to also test kernel update with patches for this new kernel - but I struggled to make it work so I just check that `dnf kpatch` is set to `auto` and hope it works as it is supposed to. I would also prefer to test `kpatch list` but making own patches that would be real was just impractical to do, so I ended up faking `kpatch` utility.
- We possibly could watch /sys/kernel/livepatch to recognize when patches are loaded, but seems rather corner case. When using our UI to install patches it will be updated after package updates
- There is corner case when for some reason patches are installed but not loaded. This would happen only if user would mangle with it in some way (like disabling service and rebooting machine). We could show `Install` button but I haven't implemented that yet.
- I don't show any changelog, follow up?

How it looks like: (sorry no screencast, computer does not want to collaborate)
(No patches installed nor loaded, but available):
packgesandneabled

The edit button (that can be `Install` when kpatch is not installed but available) opens this dialog. When user toggles switches a note is shown on bottom of what will happen if they press 'Apply'.
willdostuff

When some patches are also loaded (it will show also when some are only installed).
withactivepatches

### DIFF
--- a/pkg/packagekit/autoupdates.jsx
+++ b/pkg/packagekit/autoupdates.jsx
@@ -25,7 +25,9 @@ import {
     FormSelect, FormSelectOption,
     Modal, Radio, Text, TextVariants,
     TimePicker,
+    Split, SplitItem,
 } from '@patternfly/react-core';
+import { InfoIcon } from "@patternfly/react-icons";
 
 import { install_dialog } from "cockpit-components-install-dialog.jsx";
 import { validateTime } from "timepicker-helpers.js";
@@ -253,7 +255,7 @@ function getBackend(forceReinit) {
  * Properties:
  * onInitialized(enabled): (optional): callback once backend knowsn whether automatic updates are enabled
  */
-export const AutoUpdatesBody = (props) => {
+export const AutoUpdatesStatus = (props) => {
     const { enabled, type, day, time } = props;
     const days = {
         "": "every day",
@@ -274,7 +276,16 @@ export const AutoUpdatesBody = (props) => {
         str = _("Automatic updates are not set up");
     }
 
-    return (<Text component={TextVariants.p}>{str}</Text>);
+    return (
+        <Split hasGutter>
+            <SplitItem>
+                <InfoIcon />
+            </SplitItem>
+            <SplitItem isFilled>
+                <Text component={TextVariants.p}>{str}</Text>
+            </SplitItem>
+        </Split>
+    );
 };
 
 /**
@@ -283,7 +294,7 @@ export const AutoUpdatesBody = (props) => {
  * Properties:
  * onInitialized(enabled): (optional): callback once backend knowsn whether automatic updates are enabled
  */
-export class AutoUpdates extends React.Component {
+export class AutoUpdatesSettings extends React.Component {
     constructor() {
         super();
         this.state = {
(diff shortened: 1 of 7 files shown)


pkg/packagekit/autoupdates.jsx
```diff
@@ -397,21 +407,32 @@ export class AutoUpdates extends React.Component {
...
-                                        this.setState({ showModal: true });
-                                    }, () => null);
-                        } else {
-                            this.setState({ showModal: true });
-                        }
-                    }}>
-                {_("Edit")}
-            </Button>
+            <Split hasGutter>
+                <SplitItem>
```
KKoukiou, line 411: So here you use the split item to nicely position a label - value pair. 
I suggest you use DescritionList for such usecases. If for some reason you can't, (I see there are some buttons on the right) let's make sure that the generated html is also accessible for screen readers. For that you will need an aria-labelledby  on the second split item.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,319 @@
...
+                        loaded: true,
+                        unavailable: d.unavailable_names || [],
+                        missing: d.missing_names || [],
+                    });
+
+                    if ((d.unavailable_names || []).length === 0 && (d.missing_names || []).length === 0)
+                        this.checkSetup();
+                })
+                .catch(e => {
+                    console.log("Failed to query info about availablity of kpatch");
```
KKoukiou, line 78: This is more like: kpatch not available on the system, no? In any case it's just a console message, I don't mind.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,319 @@
...
+
+        if (this.state.auto !== this.state.futureAuto)
+            promises.push(this.installPatches(this.state.futureAuto));
+
+        if (this.state.enabled !== this.state.futureEnabled)
+            promises.push(this.handleService(this.state.futureEnabled));
+
+        Promise.all(promises)
+                .then(() => this.setState({ showModal: false }))
+                .catch(console.log)
```
KKoukiou, line 135: Let's JSON.stringify the exception at least? Otherwise we don't see it in the test logs.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,319 @@
...
+                <SplitItem>
+                    <Text component={TextVariants.h4}>{_("Kernel live patching")}</Text>
+                </SplitItem>
+                <SplitItem isFilled>
+                    <Text component={TextVariants.small}>{state}</Text>
+                </SplitItem>
+                <SplitItem>
+                    <Button variant="secondary"
+                            isSmall
+                            disabled={!this.props.privileged}
```
KKoukiou, line 208: s/disabled/isDisabled

KKoukiou (changes requested):

pkg/packagekit/updates.jsx
```diff
@@ -831,12 +832,18 @@ class CardsPage extends React.Component {
...
+            <StackItem>
+                <KpatchStatus />
+            </StackItem>
         </Stack>;
 
         const settingsContent = <Stack hasGutter>
             <StackItem>
                 <AutoUpdatesSettings onInitialized={newState => this.setState(newState)} privileged={this.props.privileged} />
             </StackItem>
+            <StackItem>
```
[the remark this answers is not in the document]
marusak: thanks! I was thinking about it whether it is actually necessary. I'll cleanup after myself for sure

marusak:
Note to myself - Commit "packagekit: Restyle UI to have status and settings cards" includes:
```
 
+    const patches = pkgNames.some(p => p.name.startsWith("kpatch-patch"));
+
+    const pkgTitle = <>
+        <span>{pkgsTruncated}</span>
+        {patches && <>{" "}<Badge>{_("patches")}</Badge></>}
+    </>;
+
     let descriptionFirstLine = (info.description || "").trim();
     if (descriptionFirstLine.indexOf("\n") >= 0)
         descriptionFirstLine = descriptionFirstLine.slice(0, descriptionFirstLine.indexOf("\n"));
(3/13) Stage this hunk [y,n,q,a,d,K,j,J,g,/,e,?]? n
@@ -318,7 +327,7 @@ function updateItem(info, pkgNames, key) {
 
     return {
         columns: [
-            { title: pkgsTruncated },
+            { title: pkgTitle },
```
which needs to be part of commit "packagekit: Introduce kpatch integration".

marusak:
Closing for time being. I want to test this on real system with real patches but that will take a while until such environment is available.

PR #15571, opened 2021-03-22
verify: revert recent changes to check-testlib

allisonkarlitskaya:
The regexp changes here broke things, but the breakage was being
temporarily covered up by a separate bug, so they didn't show in the
initial round of tests on the PR.

At the same time, things are very much broken right now, so revert the
changes in check-testlib.

We need to add an exception to the static code checks again until we fix
this properly.

### DIFF
--- a/test/verify/check-testlib
+++ b/test/verify/check-testlib
@@ -128,12 +128,12 @@ class TestRunTest(MachineCase):
         process = subprocess.run([run_tests, "--test-dir", VERIFY_DIR, "TestExample.testFail", "TestExample.testSkip"],
                                  env=env, capture_output=True)
         stdout = process.stdout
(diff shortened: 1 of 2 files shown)


marusak (approved):
Thanks!

test/verify/check-testlib
```diff
@@ -128,12 +128,12 @@ class TestRunTest(MachineCase):
         process = subprocess.run([run_tests, "--test-dir", VERIFY_DIR, "TestExample.testFail", "TestExample.testSkip"],
                                  env=env, capture_output=True)
         stdout = process.stdout
-        self.assertRegex(stdout, rb"^not ok 1 .*test/verify/check-example TestExample.testFail # RETRY 1 \(be robust against unstable tests\)$")
```
martinpitt, line 131: My first instinct is that this can't work without `re.M` multi-line matching. This has always been intended as a substring match.

PR #15574, opened 2021-03-23
pf4: Switch spinners to SVG

### LINKED ISSUE #15573: Spinners are going the wrong way and are glitchy
Several pages have spinners going counter-clockwise / anti-clockwise and this version has artifacts, especially in Firefox.

PatternFly has solved these issues by introducing an SVG based spinner. However, it's not the default.

The pages which have issues are:

- machines
- networking
- logs
- software updates
- selinux
- kdump

And it might be a problem elsewhere too.

We do have some spinners that do spin clockwise, as they should, but these most likely are legacy spinners anyway. Having a mix is bad; all should spin clockwise (legacy or PF4). And non should look glitchy, of course. :wink:

Suggestion for a fix: Switch all PF4 spinners to the new SVG based spinner.

garrett:
Fixes #15573

### DIFF
--- a/pkg/apps/utils.jsx
+++ b/pkg/apps/utils.jsx
@@ -62,7 +62,7 @@ export const ProgressBar = ({ title, data }) => {
                 {_("Waiting for other programs to finish using the package manager...")}
             </SplitItem>
             <SplitItem>
-                <Spinner size="md" />
+                <Spinner isSVG size="md" />
             </SplitItem>
         </Split>);
     } else {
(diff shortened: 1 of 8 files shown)


KKoukiou (approved):
If tests agree ( but I don't see how this could break tests actually)

KKoukiou (commented):
This affects cockpit-machines as well. Before we do the code deletion from here, we should resync it.

PR #15575, opened 2021-03-23
users: Fix lastlog parsing

### LINKED ISSUE #15452: users: parsing of dates for lastlog in user details page is wrong
If the inspected user has an entry in last log and it's not currently connected it always displays invalid date.

Screen Shot 2021-03-02 at 16 45 27

martinpitt:
get_last_login() made the assumption that the date is separated by three
spaces. But that's not the case in reality, as the "From:" column can
become very wide. As there is no reliable column size or separator that
sets apart the date, just take the last five words from the line
instead, which should be the date.

With the previous code, this tried to parse "web cons ::ffff:172.27.0.
Tue Mar..." as a date, which returned an invalid `Date` object. This
confused the `!details.logged.last` check -- it was a proper object (and
thus not falsy), but stringifying it yielded `null`, and thus moment
spat out "Invalid date". Avoid that by explicitly testing whether
getTime() is an actual number -- for invalid dates this will be `NaN`.

This was not caught by the tests because they just asserted that the
message changed away from "Logged in", thus they were happy with
"Invalid date". Instead, make sure the line contains something date-ish
which at least contains the current month name and year.

Fixes #15452

### DIFF
--- a/pkg/users/account-details.js
+++ b/pkg/users/account-details.js
@@ -63,14 +63,18 @@ function get_logged(name) {
 
 function get_last_login(name) {
     function parse_last_login(data) {
-        data = data.split('\n')[1]; // throw away header
-        if (data.length === 0) return null;
-        data = data.split('   '); // get last column - separated by spaces
-
-        if (data[data.length - 1].indexOf('**Never logged in**') > -1)
+        const line = data.split('\n')[1]; // throw away header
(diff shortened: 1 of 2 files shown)


martinpitt:
Argh, forgot eslint with webpack-watch..

KKoukiou (approved):
Taking the last 5 fields here looks indeed more robust. Thanks :)

PR #15576, opened 2021-03-23
build: Fix polkit detection

### LINKED ISSUE #14052: polkit detection is automagic
Cockpit version: 218
OS: Linux

https://github.com/cockpit-project/cockpit/pull/13963 states that it implements a `--disable-polkit` configure switch, in 218 however the polkit detection is automagic, there's no such configure switch to disable polkit integration.

`configure: WARNING: unrecognized options:  --disable-polkit`
`./configure --help | grep polkit`

Please add such a configure switch so distributions are able to properly package a cockpit version without polkit integration.

martinpitt:
Fixes #14052

I tested this with all six combinations -- auto (no argument), `--enable-polkit`, `--disable-polkit`, with and without polkit-devel installed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,7 +84,6 @@ GIO_CFLAGS="$GIO_CFLAGS -DGLIB_VERSION_MAX_ALLOWED=$GLIB_VERSION_DEF"
 
 PKG_CHECK_MODULES(LIBSYSTEMD, [$LIBSYSTEMD_REQUIREMENT])
 PKG_CHECK_MODULES(JSON_GLIB, [$JSON_GLIB_REQUIREMENT])
-PKG_CHECK_MODULES(POLKIT, [$POLKIT_REQUIREMENT],autodetect_polkit="yes",autodetect_polkit="no")
 PKG_CHECK_MODULES(GNUTLS, [$GNUTLS_REQUIREMENT])
 PKG_CHECK_MODULES(KRB5, [$KRB5_REQUIREMENT])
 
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (changes requested):
If we're fixing bugs anyway...

I'd prefer we make polkit support mandatory by default (ie: remove all autodetection) and then add an option to `--disable-polkit`.  Never autodetected.

Prevents accidentally building cockpit with missing features, and leaves us with less "cases" to test, by half.

allisonkarlitskaya:
ps: I'd like to make that a general thing: as part of our recent push to clean up the build system, let's remove all automagic detections, entirely.  polkit just volunteered to be candidate #1.

martinpitt:
@allisonkarlitskaya Sure, makes sense! Updated, it's now much simpler.

configure.ac
```diff
@@ -94,19 +93,14 @@ AC_SUBST(COCKPIT_CFLAGS)
 AC_SUBST(COCKPIT_LIBS)
 
 # bridge with optional polkit
-AC_ARG_ENABLE(polkit, AC_HELP_STRING([--disable-polkit], [Disable usage of polkit]),,enable_polkit=$autodetect_polkit)
-AC_MSG_CHECKING([whether to build with polkit])
+AC_ARG_ENABLE(polkit, AC_HELP_STRING([--disable-polkit], [Disable usage of polkit]),,enable_polkit=yes)
```
allisonkarlitskaya, line 96: Some small bikeshedding:

`enable_polkit` will be set to
  - `yes` if `--enable-polkit`
  - `no` if `--disable-polkit`
  - not set, if neither

so it maybe makes sense to simplify this a bit and handle the default by dropping this extra assignment here and inverting the check below to test `"$enable_polkit" != "no"` (ie: "unless explicitly disabled...")

On the other hand, that will result in an undefined variable expansion (which is not an error in configure scripts, but which you might find to be undesirable).

Take it or leave it :)

configure.ac
```diff
@@ -94,19 +93,14 @@ AC_SUBST(COCKPIT_CFLAGS)
...
-    AC_MSG_ERROR([Couldn't find polkit. Try installing polkit])
-  fi
+  PKG_CHECK_MODULES(POLKIT, [$POLKIT_REQUIREMENT])
   COCKPIT_BRIDGE_CFLAGS="$COCKPIT_CFLAGS $POLKIT_CFLAGS"
   COCKPIT_BRIDGE_LIBS="$COCKPIT_LIBS $POLKIT_LIBS"
   AC_DEFINE_UNQUOTED([WITH_POLKIT], [], [Build with polkit])
 else
-  enable_polkit="no"
-  AC_MSG_RESULT($enable_polkit)
   COCKPIT_BRIDGE_CFLAGS="$COCKPIT_CFLAGS"
```
allisonkarlitskaya, line 104: It seems a bit yucky to define this in a branch of a check that's polkit-specific, but that's something we can iron out later on...

configure.ac
```diff
@@ -579,6 +573,7 @@ echo "
         With coverage:              ${enable_coverage}
         With address sanitizer:     ${asan_status}
         With PCP:                   ${enable_pcp}
+        With polkit:                ${enable_polkit}
```
allisonkarlitskaya, line 576: Okay.  Maybe, for this alone, it makes sense to keep it as you have it above :)

allisonkarlitskaya (approved):

PR #15577, opened 2021-03-23
lib: replace onHover custom function in plots with an inline tooltip

### LINKED ISSUE #4006: storage: highlighted drive not apparent when it is not on screen
If you hover your mouse over a particular graph portion in the Read or the Write graph the corresponding drive will be highlighted. This is quite nice, but ineffective if you have enough drives that they don't all appear on the screen. This is somewhat related to #3934, because if the drives were ordered above the RAID devices and VGs, it might be more likely to see the ones being selected. But, with a few more drives the problem comes back.

This is just an observation, everything works as intended, it just doesn't always have the effect intended, i.e., to identify the drive of interest.

I do not know what would be best to unambiguously identify the drive that shows the behavior. The highlighting has the advantage that it skips the problem of names. One could imagine a tooltip with the name of the drive that would appear when the cursor hovered over the particular graph segment, but all this might cause ugly display issues and if it is a WWN or serial number combination still be kind of hard for the user to cope with.

KKoukiou:
Previously when hovering on the resources, we highlighted them in the
relevant part of the screen. This has two main issues:

* The lists can get long so nothing guarantees that the resources of interest
are in the visible part of the screen

* The highlighted section was just changing background color which is
not helping people with limited vision.

This commit replaces this highlighted sections on hover with an inline
tootip that appears when the plot is focused.

Note: title is also read by screen reader software therefore is an accessibility
enhancement from our previous state.

Fixes #4006

### DIFF
--- a/pkg/lib/cockpit-components-plot.jsx
+++ b/pkg/lib/cockpit-components-plot.jsx
@@ -308,7 +308,7 @@ const useLayoutSize = (init_width, init_height) => {
     return [ref, size];
 };
 
-export const SvgPlot = ({ title, config, plot_state, plot_id, onHover, className }) => {
+export const SvgPlot = ({ title, config, plot_state, plot_id, className }) => {
     const [container_ref, container_size] = useLayoutSize(0, 0);
     const [measure_ref, measure_size] = useLayoutSize(36, 20);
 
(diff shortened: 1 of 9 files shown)


garrett (approved):
I totally agree with you about the list highlighting being problematic.

And moving something that was done with JavaScript detecting hovering into something that's native to SVG is a good approach.

Thanks for the PR. LGTM! :+1:

garrett:
BTW: Our charts aren't accessible and there's no real way to make a live updating chart accessible for screen readers. We treat them as hidden images as to not overwhelm someone with a flood of stats that are effectively useless. So the title attribute's value is lost a11y-wise, as it's not just a child of a `role=presentation` element, but also of one that's a `role=image` with `aria-hidden=true` (intentionally set).

How we could solve the a11y issue for presenting stats, which is _way outside the scope of this PR, of course_: Show exceptional events.

It is something I would like for us to do for the overview page... and then bring the relevant alert to the top of each page where it makes sense. (For example: Exceedingly high disk warnings (being full, high IO, more than 0 bad blocks, etc.) being shown on overview and also on the storage page.) The first step is handling disk full events, and that's a card in our roadmap. So we'll get there eventually.

marusak (approved):
Thanks!
Design was approved, code and tests look fine.

PR #15579, opened 2021-03-24
[r79] test: Use predictable host names in check-multi-machine

martinpitt:
This makes the test stricter, and also makes them work on OSes that
don't set a host name by default (like fedora-coreos). The new Overview
page doesn't render `#system_information_hostname_text` at all without a
host name.

Cherry-picked from master commit 1cfa5787b6be1

----

This should unbreak the rhel-atomic refresh in https://github.com/cockpit-project/bots/pull/1822


marusak (approved):
Thanks, maybe also testing `@bots@1822` would be useful

martinpitt:
@marusak: I tested it locally; I first need to rebase that bots PR, I'll do that now after landing this.

PR #15580, opened 2021-03-24
test: Only parse last line for RETRY and SKIP indicators

marusak:
Reverting to broken commit so I expect tests to fail.
Also I want to test this as picking up the last line might be tricky with all the troubles we had recently with having and not having new lines.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -75,7 +75,7 @@ def print_test(test, print_tap=True, retry_reason=""):
     print()  # Tap needs to start on a separate line
     if test.process.returncode == 0:
         print("ok " + test_name(test) + retry_reason)
-    elif test.process.returncode == 77 or b"# SKIP " in test.output:
+    elif test.process.returncode == 77 or b"# SKIP " in test.output.splitlines()[-1]:
         # If the test was skipped, add the last line (which contains the reason
         # for the skip) to the result
         print("ok {0} {1}{2}".format(test_name(test),
@@ -95,6 +95,7 @@ def finish_test(opts, test, affected_tests):
     """
 
     affected = any([test.command[0].endswith(t) for t in affected_tests])
+    result = test.output.splitlines()[-1]
 
     # Try affected tests 3 times
     if test.process.returncode == 0 and affected and test.retry_when_affected and test.retries < 2:
@@ -116,20 +117,24 @@ def finish_test(opts, test, affected_tests):
                 if test.output != changed:
                     changed += b"\n"
                 test.output = changed
+                result = test.output.splitlines()[-1]
         except OSError as ex:
             if ex.errno != errno.ENOENT:
                 sys.stderr.write("\nCouldn't run tests-policy: {0}\n".format(str(ex)))
 
-        if b"# SKIP" in test.output:
+        if b"# SKIP" in result:
             print_test(test, print_tap=False)
             return None, 0
 
     # do we get a specific retry reason from tests-policy?
-    m = re.search(rb"\s*# RETRY (.*)$", test.output, re.MULTILINE)
+    m = re.search(rb"\s*# RETRY (.*)", result)
     if m:
         retry_reason = m.group(1)
         # remove it from test output; we must not print it after the 3rd time, and going to print it separately
-        test.output = re.sub(rb"\s*# RETRY .*\n", b"", test.output)
+        new_result = re.sub(rb"\s*# RETRY .*\n", b"", result)
+        output = test.output.splitlines()[:-1]
+        output.append(new_result)
+        test.output = "\n".join(output)
     elif affected:  # Don't retry affected failed tests
         retry_reason = None
     else:


marusak:
@martinpitt @allisonkarlitskaya this is why we considered that test to be OK even if it failed - the body of the test contained "# SKIP" thus we skipped it. But the body was something like "Regex foo does not match "xyz #SKIP..." and the real result did not contain any skip.

marusak:
Ok, [fails](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15580-20210324-071454-fda3f094-fedora-coreos/log.html) now as it should.

marusak:
> allisonkarlitskaya marked this pull request as draft 5 minutes ago

Why? I would say ready for review as I just removed that one commit.

allisonkarlitskaya:
> > allisonkarlitskaya marked this pull request as draft 5 minutes ago
> 
> Why? I would say ready for review as I just removed that one commit.

Bad timing :)

I did it in response to



> Ok, [fails](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15580-20210324-071454-fda3f094-fedora-coreos/log.html) now as it should.

test/common/run-tests
```diff
@@ -75,7 +75,7 @@ def print_test(test, print_tap=True, retry_reason=""):
     print()  # Tap needs to start on a separate line
     if test.process.returncode == 0:
         print("ok " + test_name(test) + retry_reason)
-    elif test.process.returncode == 77 or b"# SKIP " in test.output:
+    elif test.process.returncode == 77 or b"# SKIP " in test.output.splitlines()[-1]:
```
allisonkarlitskaya, line 78: I don't like the approach.  There's no guarantee in TAP that #skip will always come on the last line.

Why not rather check for something like a line looking like "^ok \d+ # skip" plus/minus noise/case?  That should always be a skip (since other messages should get `#` prefixed)

allisonkarlitskaya (changes requested):

test/common/run-tests
marusak: >  There's no guarantee in TAP that #skip will always come on the last line.

Isn't there? Then thats bummer

> Why not rather check for something like a line looking like "^ok \d+ # skip" plus/minus noise/case? 

That `ok \d+...` is added by `test/common/run-tests` and does not come from tap. We need to recognize it in tap and then this script adds that message (see `print_test()`)

marusak: Hmm, actually it seems that it is guaranteed that #SKIP is on the last line, see lines 79 and 82.

test/common/run-tests
```diff
@@ -113,23 +114,25 @@ def finish_test(opts, test, affected_tests):
             proc = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
             changed = proc.communicate(test.output)[0]
             if proc.returncode == 0:
-                if test.output != changed:
-                    changed += b"\n"
```
marusak, line 117: Ha,  I can put this back in, because `breaklines` behaves a bit differently than `split("\n")` in the metter of trailing newlines:
```py
>>> x = "foo\nbar\n"
>>> print(x.splitlines()[-1])
bar
>>> print(x.split("\n")[-1])

>>> 

```

marusak: Done

test/common/run-tests
allisonkarlitskaya: I think this is actually the wrong place to look.  This script isn't interpreting its own output.  It's interpreting the output from testlib.  For that, the relevant line is 1578:

```python3
        # Return 77 if all tests were skipped
        if len(skips) == test_count:
            sys.stdout.write("# SKIP {0}\n".format(", ".join(["{0} {1}".format(str(s[0]), s[1]) for s in skips])))
            return 77
        if failures:
            sys.stdout.write("# {0} TEST{1} FAILED {2}\n".format(failures, "S" if failures > 1 else "", details))
            return 1
        else:
            sys.stdout.write("# {0} TEST{1} PASSED {2}\n".format(test_count, "S" if test_count > 1 else "", details))
            return 0
```

and indeed, the last line will have `# SKIP`.  But then that means that the only place we are expecting to see `# SKIP` will also always return 77.  So, do we need this extra check at all, actually?

marusak: Also `bots/tests-policy` can add `# SKIP` and then returncode is not 77.
Also `bots/tests-policy` can add `# RETRY`.

I think that if we teach test-policy to return 77 and 1 we can simplify this.

marusak: I __think__ that this logic is triple complicated because tests-policy works also with `not ok ...` part which complicates parsing. If we change it, we will surely break tests as every project would need to use new `test/common/run-tests` immediately. But if we would create copy of tests-policy then we can switch easily and once none is using current tests-policy we can just drop it.

marusak:
Closing in favor of #15585

PR #15582, opened 2021-03-24
webpack-make: drop a "watch mode" special case

allisonkarlitskaya:
Stop removing Makefile.deps in watch mode.  There's no good reason for
it anymore, and it's causing problems.

Also, drop conditionalising the generation of the dependencies on the -m
argument having been passed.  This argument is mandatory.


martinpitt (approved):
Works really well, thanks! I'm glad to lose this hack.

PR #15583, opened 2021-03-24
lib: some improvements on the empty state patterns in tables

### LINKED ISSUE #15417: storage: Use empty state pattern at the card body instead of at the table level
When there are no NFS mounts, show an empty state pattern instead of a table.

It's basically the difference between a list of 0 elements versus an undefined list. It's fine to show an empty state inside the table for filtered tables, but this is not that.

End result should look approximately the same, except without the table header rows.

### LINKED ISSUE #15419: storage: Devices & iSCSI cards need to improve empty state style
Screenshot_2021-02-24 Storage - garrett Rain(2)

It should match the empty state of https://github.com/cockpit-project/cockpit/issues/15417. (That is: Both need to be adjusted and both should look similar to the PF empty state for cards.)

Reference for the PF Card empty state:

https://marvelapp.com/prototype/7c5cdhd/screen/72928590

Excerpt: 

Screenshot_2021-02-23 x-small empty state(1)

Since the card already has a heading, we should probably remove the second heading and just use the secondary text, which would resemble something like this modification I did on the above crop of the PF mockup:

Screenshot_2021-02-23 x-small empty state(1)-mod

KKoukiou:
Fixes #15417
Fixes #15419

* [x] Blocked on #15634

### DIFF
--- a/pkg/lib/ct-card.scss
+++ b/pkg/lib/ct-card.scss
@@ -34,6 +34,19 @@
   font-size: var(--pf-global--FontSize--2xl);
 }
 
+// Remove redundant padding from embedded toolbars (handled by header)
+.ct-card.pf-c-card .pf-c-toolbar {
+  padding: 0;
+}
+
+// Remove excess top padding from top-level empty state in cards,
+// as card headers already add enough space
+.ct-card > .pf-c-card__body > .pf-c-empty-state {
+    --pf-c-empty-state__body--MarginTop: 0;
+    padding-top: 0;
+    padding-bottom: var(--pf-global--spacer--md);
+}
+
 .ct-cards-grid {
     --ct-grid-columns: 2;
     --pf-l-gallery--GridTemplateColumns: repeat(var(--ct-grid-columns), 1fr);
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -30,7 +30,10 @@ import {
     sortable,
     expandable,
 } from '@patternfly/react-table';
-import { EmptyState, EmptyStateBody, Title } from '@patternfly/react-core';
+import {
+    EmptyState, EmptyStateBody, EmptyStateSecondaryActions,
+    Text, TextContent, TextVariants,
+} from '@patternfly/react-core';
 
 import './cockpit-components-table.scss';
 
@@ -52,6 +55,7 @@ import './cockpit-components-table.scss';
  *   }[]
  * - emptyCaption: header caption to show if list is empty
  * - emptyCaptionDetail: extra details to show after emptyCaption if list is empty
+ * - isEmptyStateInTable: if empty state is result of a filter function this should be set, otherwise false
  * - variant: For compact tables pass 'compact'
  * - gridBreakPoint: Specifies the grid breakpoints ('', 'grid' | 'grid-md' | 'grid-lg' | 'grid-xl' | 'grid-2xl')
  * - sortBy: { index: Number, direction: SortByDirection }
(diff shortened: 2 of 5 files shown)


garrett:
So this is what I'm seeing:

Screenshot_2021-03-25 Storage - admin m1 cockpit lan

And the rendered HTML for the sections is:

For the main column: 

```html
<div class="pf-c-empty-state storage-card-empty-state">
  <div class="pf-c-empty-state__content">
     <h5 class="pf-c-title pf-m-md">No NFS mounts set up</h5>
   </div>
</div>
```

```html
<div class="pf-c-empty-state__content">
  <div class="pf-c-empty-state__body">No iSCSI targets set up</div>
</div>
```

The main difference between the two is the `pf-c-empty-state` outer element in the main column that's missing in the side columns. And the main column has the message in a header. (But there's no content under it, so it might be weird. I guess the PF empty state generally has heading, content, and actions, and each section is optional.)

garrett:
Looking in the code, I don't see much of a difference, aside from using a title instead of a body element and the size of xs, which might be causing the difference in padding. Perhaps it should be sm instead. I'll look and see if I can sort it out in the React code, as it'd be a small change if so.

garrett:
BTW, technically, if we were to use a heading where the h5 is, it should be an h3, as we shouldn't skip levels of headings.

However, as the card has am h2 at the top, I think it's better not being a heading. I'll play around and post something if it's good enough.

garrett:
I started fixing this, then things that touched it (ct-card included), then checked other pages where it's used, then started fixing the other pages, and so on and so on. It's all related.

Currently trying to split it all into separate commits. Debating if I should make a bunch of PRs or just re-scope this one to update the cards (with existing commits for empty state + the card ones). I moved the empty state handling from your commit into PF cards too, BTW.

Sorry, it's a quite a bit more than originally intended, but I have fixed a bunch of spacing and alignment problems across Cockpit. Whoops.

Anyway, keeping related things as separate commits, so they're more easy to review and debug (in the future, if needed) and possibly even split apart into separate PRs.

garrett:
I managed to split off one commit to #15597, as it was completely unrelated to the page.

The rest of my commits affect the spacing of the cards on this page, including the empty states. Some also affect other pages too (Networking, Machines, and anywhere else cards are used), as I tried to keep most of the changes generalized for all pages.

pkg/storaged/nfs-panel.jsx
```diff
@@ -106,10 +106,10 @@ export class NFSPanel extends React.Component {
                                   { title:  _("Size"), transforms: [cellWidth(40)] }
                               ]}
                               rows={mounts} />
-                    : <EmptyState className='storage-card-empty-state'>
-                        <Title headingLevel="h5" size="md">
```
KKoukiou: This is the empty state we display from pkg/lib/cockpit-component-table.jsx as well.
I feel if we change it here we need to change it there as well.

pkg/lib/ct-card.scss
```diff
@@ -29,6 +29,23 @@
...
+// Remove redundant padding from embedded toolbars (handled by header)
+.ct-card.pf-c-card .pf-c-toolbar {
+  padding: 0;
+}
+
+// Remove excess top padding from empty state in cards,
+// as card headers already add enough space
+.ct-card.pf-c-card .pf-c-empty-state {
+    --pf-c-empty-state__body--MarginTop: 0;
+    padding-top: 0;
```
KKoukiou, line 46: This looks wrong in another pages now.
Screen Shot 2021-03-29 at 17 03 57
Screen Shot 2021-03-29 at 17 03 26

pkg/storaged/nfs-panel.jsx
garrett: Ah, yeah, we might need to.

pkg/lib/ct-card.scss
garrett: Ugh. This is broken... but...

- Those pages should be using an empty state directly instead of a list with an empty state.
- List with empty states are supposed to be used for filters on a list, and generally contain reset the list as a pattern.

I guess we could change the CSS to affect direct empty states instead of all empty states. (And fix the lists in another PR as they're out of scope here.)

Thanks for pointing this out. I'll fix.

KKoukiou: I think we can do it in this PR, it's literally one small place to change it.

KKoukiou: And then I don't need to handle explicitely the NFS case.

garrett: Sounds good. I'll still make the padding/margin changes only apply to a toplevel empty state in the CSS meanwhile.

Having empty states in lists still makes sense for when they're being filtered. I'd hate to break that for any lists in cards that are filtered.

garrett:
Heading off for the evening. I'll rebase my commit tomorrow and share it.

Meanwhile, it's just this bit of CSS:

```scss
// Remove excess top padding from top-level empty state in cards,
// as card headers already add enough space
.ct-card > .pf-c-card__body > .pf-c-empty-state {
    --pf-c-empty-state__body--MarginTop: 0;
    padding-top: 0;
}
```

Emphasis on `.ct-card > .pf-c-card__body > .pf-c-empty-state` changing. (The comment changed slightly too.)

KKoukiou:
@garrett rebased and removed conflicts. Can you please give a review here?

garrett:
Better!

It seems weird that NFS and Devices are using different font sizes. Is that intentional? I think this is the last thing that jumps out at me on this PR.

Screenshot_2021-04-12 Storage - garrett Rain

KKoukiou:
@garrett it's because I use pf-m-xs for the side panel empty state. Which of the two versions should I keep? The XS or the normal? Expect that the XS causes smaller paddings as well.

garrett:
Ah, the XS is too small, probably. I guess increasing the font size makes extra padding that looks weird too. Not sure what's best, but XS is hard to read; it's meant to be used sparingly overall, I think. The text is probably more correctly SM?

KKoukiou:
@garrett Fixed the inconsistent font-sizes.

garrett (approved):
Looks great. Thanks! :tada:

PR #15584, opened 2021-03-24
 File filter...  lib: packagekit: update progress bars by using the PK backend provided value

### LINKED ISSUE #13953: apps: Progressbars go backwards
As spotted in #13945, progress bars in the Applications page sometimes goes backwards.

The code for the Applications page should ensure that progress bars only move forward. 

If there's jitter, perhaps it should adapt the number updates a bit. Otherwise, it should at least check previous with current value and only update the presented percentage with a higher value, not a lower one.

[xkcd comic 612](https://xkcd.com/612/)

If the jumps are a bit too quick, we probably should make sure the width is adjusted with CSS using transition to smoothly display the visual growth (so it quickly animates instead of immediately jumping). The progress bar may already have this baked in. (It was too quick to check.)

One possible naïve approach, assuming the spikes are only returned once in a while per sequential call (so this might not work), may be to always show the lower of the last two progress percentages and only advance when it is higher. Something simple like this may work if it goes from 4% to 50% to 10% to 35% etc. Then the 50% would be dropped as noise and the progress bar would show more accurate, always forward moving behavior. Tied with CSS transition smoothing (which _might_ be baked-in or might not — but it's just 1 line to add) would make the progress bar _feel_ much more accurate too (so you wouldn't really notice the skip in that iteration of the loop).

I was testing the the `debian-stable` VM, in case that matters.

KKoukiou:
progress_reporter (see pkg/apps/packagekit.js) is combining progress
from multiple operations by using the base/range logic.

The only way for this concept to work is to only have this callback get
called when the percentage has changed.

Otherwise, we just keep adding progress without any actual change
happening in the backend.

Example:

method = InstallPackages
base = 1, range: 99

PK backend: Properties changed: { Percentage: 0 }
progress_reporter: data_percentage = 1 + 0 * 100/99 = 1

PK backend: Properties changed: { Percentage: 1 }
progress_reporter: data_percentage = 1 + 1 * 100/99 ~= 2.01

PK backend: Properties changed: { Status: 2 }
progress_reporter: data_percentage = 1 + 2.01 * 100/99 <--- here the
percentage increased but it should not! Percentage property did not change

Fixes #13953

### DIFF
--- a/pkg/lib/packagekit.js
+++ b/pkg/lib/packagekit.js
@@ -242,7 +242,7 @@ export function cancellableTransaction(method, arglist, progress_cb, signalHandl
         let allow_wait_status = false;
         const progress_data = {
             waiting: false,
-            percentage: 0,
+            absolute_percentage: 0,
             cancel: null
         };
 
@@ -256,10 +256,10 @@ export function cancellableTransaction(method, arglist, progress_cb, signalHandl
                 if ("Status" in props)
                     status = props.Status;
                 progress_data.waiting = allow_wait_status && (status === Enum.STATUS_WAIT || status === Enum.STATUS_WAITING_FOR_LOCK);
-                if ("Percentage" in props && props.Percentage <= 100)
-                    progress_data.percentage = props.Percentage;
                 if ("AllowCancel" in props)
                     progress_data.cancel = props.AllowCancel ? cancel : null;
+                if ("Percentage" in props && props.Percentage <= 100)
+                    progress_data.absolute_percentage = props.Percentage;
 
                 progress_cb(progress_data);
             }
--- a/pkg/apps/packagekit.js
+++ b/pkg/apps/packagekit.js
@@ -23,8 +23,8 @@ import * as PK from "packagekit.js";
 function progress_reporter(base, range, callback) {
     if (callback) {
         return function (data) {
-            if (data.percentage >= 0)
-                data.percentage = base + data.percentage / 100 * range;
+            if (data.absolute_percentage >= 0)
+                data.percentage = base + data.absolute_percentage / 100 * range;
             callback(data);
         };
     }


pkg/lib/packagekit.js
```diff
@@ -256,12 +256,12 @@ export function cancellableTransaction(method, arglist, progress_cb, signalHandl
                 if ("Status" in props)
                     status = props.Status;
                 progress_data.waiting = allow_wait_status && (status === Enum.STATUS_WAIT || status === Enum.STATUS_WAITING_FOR_LOCK);
-                if ("Percentage" in props && props.Percentage <= 100)
-                    progress_data.percentage = props.Percentage;
                 if ("AllowCancel" in props)
                     progress_data.cancel = props.AllowCancel ? cancel : null;
```
marusak, line 260: But it will not now notify on changes like `progress_data.cancel` and such, won't it?

KKoukiou: Correct, but  really enabling/disabling the cancel operation 1 second before or after does not cause a big problem right?

Let me think about this a bit. Maybe there is a better solution.

KKoukiou: Ok I found  a cleaner fix.

marusak:
sorry I don't follow. It seems like you only renamed the member from `percentage` to `absolute_percentage`. Am I missing here something?

KKoukiou:
> sorry I don't follow. It seems like you only renamed the member from `percentage` to `absolute_percentage`. Am I missing here something?

Yes, if you look closer, I changed the name of the backend provided value to absolute_percentage and then I build the percentage value based on this, instead of itself.

https://github.com/cockpit-project/cockpit/pull/15584/commits/2569077dbd7ff2c96c16f777bb524b661c9d706b#diff-2b92bf3cd700b56dfbebfd6729f731df72b929746b87e4489fb798be15c1416aR27

garrett (approved):
Oh, this is so much better! :+1:

martinpitt (approved):
Ugh, what a brown paperbag bug! Thanks!

PR #15585, opened 2021-03-24
tests: Don't try to parse skip and retry reasons from output

marusak:
This was prone to unexpected behavior if output would contain any of
these key words.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -8,7 +8,6 @@ import sys
 import unittest
 import errno
 import time
-import re
 import socket
 import tempfile
 import binascii
@@ -57,7 +56,7 @@ def flush_stdout():
             time.sleep(0.1)
 
 
-def print_test(test, print_tap=True, retry_reason=""):
+def print_test(test, print_tap=True, retry_reason="", skip_reason=""):
     for line in test.output.strip().splitlines(keepends=True):
         while line:
             try:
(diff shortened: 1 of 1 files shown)


marusak:
This needs https://github.com/cockpit-project/bots/pull/1840 so now testing it.
Thinks I am looking at:

- [ ] test/verify/check-testlib
- [ ] skipped test due to @skipImage
- [ ] skipped test due to known issue
- [ ] retried test due to harness
- [ ] retried affected test

marusak:
```
2021-03-24T14:52:44.6095116Z test/common/run-tests:75:5: E303 too many blank lines (2)
2021-03-24T14:52:44.6096254Z ERROR: tools/test-static-code process failed: 1
```
gotcha, will fix on next force push

marusak:
Known issue needs a bit of fix:
```
ok 224 test/verify/check-storage-resize TestStorageResize.testGrowShrinkEncryptedHelp # SKIP b'Known issue #1739'
```

marusak:
It did not reproduce retry 'due to harness' but otherwise looks good. One code style and one bytes vs str mismatch. I'll fix those on force push, but ready for review.

martinpitt:
```
FAIL: tools/test-static-code 7 pep8 test
test/common/run-tests:75:5: E303 too many blank lines (2)
```

test/common/run-tests
```diff
@@ -67,83 +66,85 @@ def print_test(test, print_tap=True, retry_reason=""):
...
 
+    if retry_reason:
+        retry_reason = " " + retry_reason
+    if skip_reason:
+        skip_reason = " " + skip_reason
+
+
     if not print_tap:
-        print(retry_reason)
+        print((retry_reason or "") + (skip_reason or ""))
```
martinpitt, line 76: Why the `or ""`? The default is already `""`, so this would only fail if you explicitly passed `None`? But, I see the code below in finish_test() sets them to `None`, so perhaps just make the default args None.

test/common/run-tests
```diff
@@ -67,83 +66,85 @@ def print_test(test, print_tap=True, retry_reason=""):
...
-        print("not ok " + test_name(test) + retry_reason)
+        print("not ok {0}{1}".format(test_name(test), retry_reason))
     flush_stdout()
 
 
 def finish_test(opts, test, affected_tests):
     """Returns if a test should retry or not
 
-    Call test-policy on the test's output, print if needed.
+    Call tests-validate on the test's output, print if needed.
```
martinpitt, line 93: That'll need to be renamed. (Some more instances below).

test/common/run-tests
```diff
@@ -67,83 +66,85 @@ def print_test(test, print_tap=True, retry_reason=""):
...
     print()  # Tap needs to start on a separate line
     if test.process.returncode == 0:
-        print("ok " + test_name(test) + retry_reason)
-    elif test.process.returncode == 77 or b"# SKIP " in test.output:
-        # If the test was skipped, add the last line (which contains the reason
-        # for the skip) to the result
-        print("ok {0} {1}{2}".format(test_name(test),
-                                     test.output.splitlines()[-1].strip().decode() if test.process.returncode == 77 else "",
-                                     retry_reason))
+        print("ok {0}{1}".format(test_name(test), retry_reason))
```
martinpitt, line 82: Proposal for cleanup: do the ` # RETRY <reason>` formatting here instead of in teh caller. Then the whole line formatting is in one place, and you don't have to duplicate it any more. Same for skip, of course.

martinpitt (changes requested):
Thanks! Please give some details in the commit message, like what it got replaced with (point to the bots PR and that it's now using special exit codes).

test/common/run-tests
marusak: right, because when I am parsing these reasons I default to None thus I can pass None directly. This is thinko. I'll init reasons to "" and then should be fine.

test/common/run-tests
marusak: The problem is with SKIP as we can get it from `test/common/testlib.py`. That means we have it once pre-formatted. Retry could be formatted here, but then it would be treated differently from skip.

marusak:
I'll trigger all the tests with force push

marusak:
/packit build

(because failed with `Guest couldn't be provisioned: Condition 'ip_ready' failed to pass with given time` - [log](http://artifacts.dev.testing-farm.io/631ab64d-135b-4b43-9248-c2b468df488b/))

test/common/run-tests
```diff
@@ -67,83 +66,84 @@ def print_test(test, print_tap=True, retry_reason=""):
...
+        print_test(test, not opts.list, "# RETRY {0} ({1})".format(test.retries, retry_reason))
         return retry_reason, 0
-    elif test.process.returncode in [0, 77]:
-        print_test(test, not opts.list)
+
+    # If test is being skipped pick up the reason
+    if test.process.returncode == 77:
+        lines = test.output.splitlines()
+        skip_reason = lines[-1].strip().decode("utf-8")
+        test.output = b"\n".join(lines[:-1])
```
martinpitt, line 112: I thought we now wouldn't have to do this split/parse/join dance any more -- the entire output should be the reason, no?

martinpitt (changes requested):
Looks good, thanks!  Simpler and more robust that way. Just some minor cleanup left.

test/common/run-tests
marusak: Well yeah, this one is special case: https://github.com/cockpit-project/cockpit/blob/master/test/common/testlib.py#L1579

But it is just "last line when returned 77" which is rather safe compered to previous regexes.

martinpitt: I see, thanks!

martinpitt (approved):

PR #15589, opened 2021-03-25
test: Drop run-tests --base default branch name

martinpitt:
This was inelegant (hardcoding) and often wrong: cockpit has stable
rhel-8.x branches, and cockpit-machines' default branch is "main".  E.g.
a local run-tests invocation in cockpit-machines or other project which
has a "main" instead of a "master" branch previous spat out this error:

    fatal: ambiguous argument 'origin/master': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
    fatal: bad revision 'origin/master'

It is also conceptually dubious: The "retry affected tests" logic is
intended for pull requests, which always have an explicit target branch.
It is not intended for downstream gating (where we previously had a hack
to disable it), nor particularly desirable for local test runs (where
developers usually want to investigate/re-run a test in a more
controlled fashion).

Change run-tests to require explicitly enabling the "retry affected
tests" mechanism by specifying a base branch, either through
`$BASE_BRANCH` or with `--base`. Once that is given, *require* that git
is available, the checkout is a git repo, and `git diff` succeeds. This
will prevent silently breaking this feature in our CI.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -215,14 +215,12 @@ def run(opts, image):
     # If file from `test-dir` was changed in the PR, all tests from it are considered affected
     changed_tests = []
     test_files = glob.glob(os.path.join(opts.test_dir, "check-*"))
-    if os.path.exists(os.path.join(os.path.dirname(testvm.TEST_DIR), ".git")):
+    if opts.base:
         # Detect affected tests from changed test files
-        cmd = ["git", "diff", "--name-only", "origin/" + opts.base, opts.test_dir]
-        r = subprocess.run(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
-        if r.returncode == 0:
-            # Never consider 'test/verify/check-example' to be affected - our tests for tests count on that
-            # This file provides only examples, there is no place for it being flaky, no need to retry
-            changed_tests = [test.decode("utf-8") for test in r.stdout.strip().splitlines() if not test.endswith(b"check-example")]
+        diff_out = subprocess.check_output(["git", "diff", "--name-only", "origin/" + opts.base, opts.test_dir])
+        # Never consider 'test/verify/check-example' to be affected - our tests for tests count on that
+        # This file provides only examples, there is no place for it being flaky, no need to retry
+        changed_tests = [test.decode("utf-8") for test in diff_out.strip().splitlines() if not test.endswith(b"check-example")]
 
         # If more than 3 test files were changed don't consider any of them as affected
         # as it might be a PR that changes more unrelated things.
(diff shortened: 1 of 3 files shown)


martinpitt:
With this, `test/common/run-tests -l` is now clean, and `test/common/run-tests -l --base main` in my [special-cases branch](https://github.com/cockpit-project/cockpit-machines/pull/17) finds the retries:
```
TestMachinesSnapshots.testSnapshotCreate # RETRY 1 (test affected tests 3 times)
TestMachinesSnapshots.testSnapshotCreate # RETRY 2 (test affected tests 3 times)
TestMachinesSnapshots.testSnapshotCreate
TestMachinesSnapshots.testSnapshotRevert # RETRY 1 (test affected tests 3 times)
TestMachinesSnapshots.testSnapshotRevert # RETRY 2 (test affected tests 3 times)
TestMachinesSnapshots.testSnapshotRevert
TestMachinesSnapshots.testSnapshots # RETRY 1 (test affected tests 3 times)
TestMachinesSnapshots.testSnapshots # RETRY 2 (test affected tests 3 times)
TestMachinesSnapshots.testSnapshots
```

test/common/run-tests
```diff
@@ -215,14 +215,12 @@ def run(opts, image):
...
-    if os.path.exists(os.path.join(os.path.dirname(testvm.TEST_DIR), ".git")):
+    if opts.base:
         # Detect affected tests from changed test files
-        cmd = ["git", "diff", "--name-only", "origin/" + opts.base, opts.test_dir]
-        r = subprocess.run(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
-        if r.returncode == 0:
-            # Never consider 'test/verify/check-example' to be affected - our tests for tests count on that
-            # This file provides only examples, there is no place for it being flaky, no need to retry
-            changed_tests = [test.decode("utf-8") for test in r.stdout.strip().splitlines() if not test.endswith(b"check-example")]
+        changed_files = subprocess.check_output(["git", "diff", "--name-only", "origin/" + opts.base, opts.test_dir])
```
marusak, line 220: on lines 238-240 we have the same pattern, want to change that one as well to `check_output`?

martinpitt: I confirmed that non-existing paths (`pkg/` only exist in cockpit) does not make it fail, so I'll do that.

martinpitt: Done. I renamed the variable to `diff_out`, as otherwise with that schema we would have `changed_pkg` and `changed_pkgs`  below, which is confusing.

marusak (approved):
Thanks!

PR #15593, opened 2021-03-25
bridge: Always connect stderr to journal when not being interactive

mvollmer:


mvollmer:
@allisonkarlitskaya, you will not like this, but I just wanted to make one of the options in the sudo/stderr/journal saga explicit.

I think this increases consistency, actually.

This will cause the session prcoesses (cockpit-bridge, cockpit-pcp, all scripts spawned by cockpit-bridge) to always log to the journal, even if the session is started by a cockpit-ws process that is not logging to the journal. Such as when running cockpit-ws in a terminal. Personally, I am fine with that, mostly because I could never really figure out who logs where under which conditions anyway.

allisonkarlitskaya:
Seems to break tests :(

allisonkarlitskaya (changes requested):
.

mvollmer:
> Seems to break tests :(

This is because `check_journal_messages` forgot to also look at messages with the "cockpit/ssh" identifier. Which means we have ignored all messages on secondary machines since, like, forever. :-(

This PR changes the syslog identifier for SSH sessikons to "cockpit-bridge" as well, and the tests found one message from sudo that should have been allowed.

mvollmer:
> Which means we have ignored all messages on secondary machines since, like, forever. :-(

And I only now realize that we have ignored all g_log message as well, since they don't have SYSLOG_IDENTIFIER set at all... ouch.

mvollmer:
See #15599 for the sorting out of the journal messages.

mvollmer:
Ouch, I messed this up somehow and removed the actual change when pushing 4225056...

mvollmer:
Rebased onto #15599.

src/bridge/bridge.c
```diff
@@ -746,10 +725,34 @@ main (int argc,
...
+        }
+
+      /* Connect stderr to the local journal explicitly.  This makes
+       * sure that we log to the machine where this process is
+       * running, and that glib always does structured logging even if
+       * our stderr was redirected to a pipe to capture errors from
+       * any wrapper that has started us (like sudo).
+       */
+
+      int fd = sd_journal_stream_fd ("cockpit-bridge", LOG_WARNING, 0);
```
allisonkarlitskaya, line 743: In any case, please set this to something other than "cockpit-bridge": cockpit-bridge will end up with the glib-reopened channel and name itself correct "cockpit-bridge".  Everything else that ends up in here will be something which *isn't* cockpit-bridge (ie: random stderr output from the things we spawn).  I set this to `cockpit/ssh` before it make it clear that it wasn't a program name.  Maybe `cockpit/session` now or so?

allisonkarlitskaya (changes requested):
This makes me unhappy.

But I think it makes you more happy than it makes me unhappy.

Assuming you can fix up all of the fallout from this change, I'm not gonna keep blocking you.  And it makes me happy that it will dramatically simplify the sudo patch.

But why not just add the commandline argument, as we discussed?

src/bridge/bridge.c
mvollmer: Ok, done!

src/bridge/bridge.c
```diff
@@ -746,10 +725,34 @@ main (int argc,
...
+       */
+
+      int fd = sd_journal_stream_fd ("cockpit/session", LOG_WARNING, 0);
+
+      /* If it didn't work, then there's no journal.  That's OK: we'll
+       * just send the output back to the client after all.
+       *
+       * If it did work, rename the fd to 2 (stderr).
+       */
+      if (fd >= 0)
```
allisonkarlitskaya, line 750: If we get a failure to connect to the journal, and we're relying on this redirection to prevent misinterpreting sudo's stderr messages, then this is gonna be a problem...

allisonkarlitskaya (changes requested):

src/bridge/bridge.c
mvollmer: The main point of connecting to journald here is to convince glib to do the usual structured logging, no?  (And in the ssh case to keep the messages on the same host as the processes.)

So if we can't connect to journald, then glib probably can't either and will fall back to unstructured logging. I think that's just fine.

For the ssh case, we could try connecting to the old school syslog thing. How is that done, actually?

allisonkarlitskaya:
I'm working on an alternate proposal.

mvollmer:
Blocked on @allisonkarlitskaya:s proposal

mvollmer:
We are going with #15666

PR #15596, opened 2021-03-25
overview: buttons in the health card are used inline and should be using the isInline property

### LINKED ISSUE #15591: overview: Odd link padding in the health card
When running a Debian VM for testing, I saw an oddly padded link in the overview, for a failed service (not sure if it affects more linked health issues):

health-padding-crop

It's a regression.

The element is marked as a `pf-c-button pf-m-link`, but it's using button padding, when it should not. This adds an interior 1rem, which is fine, except for the left side, as this moves it too far away from the icon. (Compare to the package update check passing.)

KKoukiou:
This removes the unwanted padding added.

Fixes #15591


martinpitt (approved):
Cheers!

Issue #15547, opened 2021-03-17
machines: storage pools list is broken on mobile mode


KKoukiou:
![Screen Shot 2021-03-17 at 10 02 33](https://user-images.githubusercontent.com/14921356/111441798-f6cbbc00-8707-11eb-8bea-1d8488a29ceb.png)

KKoukiou:
Name for Networks list.

PR #15597, opened 2021-03-25
machines: Fix label in mobile mode

garrett:
Labels ("Shut off") do no longer fill the maximum area, as seen in this screenshot:

Screenshot_2021-03-25 Virtual Machines - garrett Rain

(There are a couple of other fixes here in this screenshot as well, but those are a part of a tangled web of commits intended to affect spacing for our card usage in general and empty state patterns specifically, soon to be landing in #15583.)

### DIFF
--- a/pkg/machines/components/vms/hostvmslist.scss
+++ b/pkg/machines/components/vms/hostvmslist.scss
@@ -15,6 +15,13 @@
     --pf-c-table--cell--Width: 1%;
 }
 
+@media screen and (max-width: $pf-global--breakpoint--lg) {
+    // Fix labels to not fully stretch when in responsive mode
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
This is not only related to machines, it affects networks and storage as well.

See https://github.com/cockpit-project/cockpit/issues/15547

KKoukiou:
I will open the PR in the cockpit-machines project.

garrett:
> Also can you please open machines PRs against the new repo @cockpit-project/cockpit-machines

Whoops! Sorry.

> I will open the PR in the cockpit-machines project.

Thanks!

> This is not only related to machines, it affects networks and storage as well.
>
> See #15547

I wonder if we need a workaround or some Cockpit-specific CSS, like our ct-card stuff. (Or we just override it locally in both places.)

Issue #15544, opened 2021-03-17
TestLogin.testConversation sometimes invokes sudo


martinpitt:
Recently I saw several instances of [this test failure](https://logs.cockpit-project.org/logs/pull-15543-20210317-032630-7a4e80e4-debian-testing/log.html#197):
```
testlib.Error: FAIL: Test completed, but found unexpected journal messages:
sudo: 3 incorrect password attempts
```
The [journal](https://logs.cockpit-project.org/logs/pull-15543-20210317-032630-7a4e80e4-debian-testing/TestLogin-testConversation-debian-testing-127.0.0.2-2301-FAIL.log.gz) confirms this:
```
Mar 17 03:52:05 debian root[31264]: COCKPITTEST: start TestLogin.testConversation
Mar 17 03:52:05 debian sudo[31188]:    admin : 3 incorrect password attempts ; PWD=/run/user/1002 ; USER=root ; COMMAND=/usr/bin/cockpit-bridge --privileged
Mar 17 03:52:05 debian cockpit-ws[31188]: sudo: 3 incorrect password attempts
Mar 17 03:52:05 debian sudo[31199]:    admin : a password is required ; PWD=/run/user/1002 ; USER=root ; COMMAND=validate
Mar 17 03:52:05 debian systemd[1]: systemd-timedated.service: Succeeded.
Mar 17 03:52:05 debian systemd[1]: session-132.scope: Succeeded.
Mar 17 03:52:05 debian systemd[1]: Stopped Session 132 of user admin.
```
This also happened on CentOS 8 and other OSes, it's not OS specific. 

This is a bit hard to reproduce locally. But I don't understand it conceptually -- this test is not *meant* to attempt sudo authentication. It just checks the initial login page.

*If* this test legitimately does superuser escalation, then fixing it is trivial (`self.allow_failed_sudo_journal_messages()`), but I don't see a good justification for it.

@mvollmer, do you have any idea about this?

martinpitt:
[still happens](https://logs.cockpit-project.org/logs/pull-15668-20210408-142703-164c3689-debian-testing/log.html#149)

mvollmer:
> But I don't understand it conceptually -- this test is not _meant_ to attempt sudo authentication. It just checks the initial login page.

The test does a full login.  The `Browser.try_login` method will by default request to run sudo. But that should always succeed. So the question is why sudo sometimes fail.

martinpitt:
[...]

PR #15599, opened 2021-03-26
test: Look at more journal messages

mvollmer:
This adds the "cockpit/ssh" syslog identifier, which is used for
messages written to stderr on remote machines, and adds all the
GLIB_DOMAINS that are used by core Cockpit.

### DIFF
--- a/src/bridge/cockpitpcpmetrics.c
+++ b/src/bridge/cockpitpcpmetrics.c
@@ -177,7 +177,7 @@ build_meta (CockpitPcpMetrics *self,
               rc = pmNameInDom (self->metrics[i].desc.indom, vs->vlist[j].inst, &instance);
               if (rc != 0)
                 {
-                  g_warning ("%s: instance name lookup failed: %s", self->name, pmErrStr (rc));
+                  g_warning ("%s: instance name lookup failed: %s.%d: %s", self->name, self->metrics[i].name, vs->vlist[j].inst, pmErrStr (rc));
                   instance = NULL;
                 }
 
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1161,8 +1161,8 @@ class MachineCase(unittest.TestCase):
             # Debian images don't have any non-C locales (mostly deliberate, to test this scenario somewhere)
             self.allowed_messages.append("invalid or unusable locale: .*")
 
-        if self.image.startswith('fedora'):
-            # Fedora switched to dbus-broker
+        if self.image.startswith('fedora') or self.image.startswith('rhel-9'):
+            # Fedora and RHEL 9 have switched to dbus-broker
             self.allowed_messages.append("dbus-daemon didn't send us a dbus address; not installed?.*")
 
         if self.image in ['fedora-34']:
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1082,6 +1082,8 @@ class MachineCase(unittest.TestCase):
     def allow_hostkey_messages(self):
         self.allow_journal_messages('.*: .* host key for server is not known: .*',
                                     '.*: refusing to connect to unknown host: .*',
+                                    '.*: .* host key for server has changed to: .*',
+                                    '.*: host key for this server changed key type: .*',
                                     '.*: failed to retrieve resource: hostkey-unknown')
 
     def allow_restart_journal_messages(self):
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1061,6 +1061,11 @@ class MachineCase(unittest.TestCase):
         r"#1\) Respect the privacy of others.",
         r"#2\) Think before you type.",
         r"#3\) With great power comes great responsibility.",
+
+        # Something unknown sometimes goes wrong with PCP, see #15625
+        "pcp-archive: instance name lookup failed: .*",
+        "direct: instance name lookup failed: .*",
+        "pcp-archive: no such metric: .*"
     ]
 
     default_allowed_messages += os.environ.get("TEST_ALLOW_JOURNAL_MESSAGES", "").split(",")
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1125,12 +1125,28 @@ class MachineCase(unittest.TestCase):
         machine = machine or self.machine
         # on main machine, only consider journal entries since test case start
         cursor = (machine == self.machine) and self.journal_start or None
-        syslog_ids = ["cockpit-ws", "cockpit-bridge"]
+
+        # Journald does not always set trusted fields like
+        # _SYSTEMD_UNIT or _EXE correctly for the last few messages of
+        # a dying process, so we filter by the untrusted but reliable
+        # SYSLOG_IDENTIFIER instead.
+
+        matches = [
+            "SYSLOG_IDENTIFIER=cockpit-ws",
+            "SYSLOG_IDENTIFIER=cockpit-bridge",
+            "SYSLOG_IDENTIFIER=cockpit/ssh",
+            "GLIB_DOMAIN=cockpit-ws",
+            "GLIB_DOMAIN=cockpit-bridge",
+            "GLIB_DOMAIN=cockpit-ssh",
+            "GLIB_DOMAIN=cockpit-pcp"
+        ]
+
         if not self.allow_core_dumps:
-            syslog_ids += ["systemd-coredump"]
+            matches += ["SYSLOG_IDENTIFIER=systemd-coredump"]
             self.allowed_messages.append("Resource limits disable core dumping for process.*")
 
-        messages = machine.journal_messages(syslog_ids, 6, cursor=cursor)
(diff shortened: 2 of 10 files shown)


mvollmer:
Here is one newly found message:
```
pcp-archive: instance name lookup failed: Instance identifier not defined in the PCP archive log
```
This might be harmless situation, and we might want to not emit this warning in the first place.

mvollmer:
Other messages are about problems with ssh host keys, like
```
cockpit-ssh 10.111.113.2: refusing to connect to unknown host: 10.111.113.2:22
cockpit-ssh 10.111.113.2: ssh-ed25519 host key for server has changed to: SHA256:8sAqALpPEgWV5aBSjVsmNQ9xgVVtxp1n0esG+zSQ6bI
```

If these errors are also shown in the UI, I think we can omit from the journal.

mvollmer:
> Here is one newly found message:
> 
> ```
> pcp-archive: instance name lookup failed: Instance identifier not defined in the PCP archive log
> ```
> 
> This might be harmless situation, and we might want to not emit this warning in the first place.

No, this happens when PCP gives us a set of values for the instances of a metric, but then fails to retrieve the name of the instance. I'll investigate a bit.

mvollmer:
Another one:

```
org.freedesktop.NetworkManager: couldn't get property org.freedesktop.systemd1.Service ExecStartPre at /org/freedesktop/systemd1/unit/NetworkManager_2dwait_2donline_2eservice: GDBus.Error:org.freedesktop.DBus.Error.NoReply: Remote peer disconnected
```

The disconnection is expected, but why is the D-Bus channel for o.fd.NM trying to get systemd properties of systemd pathnames?

mvollmer:
> The disconnection is expected, but why is the D-Bus channel for o.fd.NM trying to get systemd properties of systemd pathnames?

#15617

test/common/testlib.py
```diff
@@ -1123,10 +1123,34 @@ def check_journal_messages(self, machine=None):
...
+            glib_domain("cockpit-pcp")
+        ]
+
         if not self.allow_core_dumps:
-            syslog_ids += ["systemd-coredump"]
-        messages = machine.journal_messages(syslog_ids, 6, cursor=cursor)
+            matches += [syslog_id("systemd-coredump")]
+
+        messages = machine.journal_messages(matches, 6, cursor=cursor)
+        print("M", messages)
```
martinpitt, line 1156: debugging, remove before landing

test/common/testlib.py
```diff
@@ -1123,10 +1123,34 @@ def check_journal_messages(self, machine=None):
...
+        def syslog_id(v):
+            return "SYSLOG_IDENTIFIER=" + v
+
+        def glib_domain(v):
+            return "GLIB_DOMAIN=" + v
+
+        matches = [
+            syslog_id("cockpit-ws"),
+            syslog_id("cockpit-bridge"),
+            syslog_id("cockpit/ssh"),
```
martinpitt, line 1145: Ugh, our naming consistency really rocks! :crying_cat_face:

src/bridge/cockpitpcpmetrics.c
```diff
@@ -177,7 +177,7 @@ build_meta (CockpitPcpMetrics *self,
               rc = pmNameInDom (self->metrics[i].desc.indom, vs->vlist[j].inst, &instance);
               if (rc != 0)
                 {
-                  g_warning ("%s: instance name lookup failed: %s", self->name, pmErrStr (rc));
+                  g_warning ("%s: instance name lookup failed: %s.%d: %s", self->name, self->metrics[i].name, vs->vlist[j].inst, pmErrStr (rc));
```
martinpitt, line 180: I don't see a reason to not just keep this. Can you please just clean up the commit message?

pkg/networkmanager/interfaces.js
```diff
@@ -400,8 +400,8 @@ export function NetworkManagerModel() {
     var watch;
 
     self.preinit.then(() => {
-        subscription = client.subscribe({ }, signal_emitted);
-        watch = client.watch({ });
+        subscription = client.subscribe({ path_namespace: "/org/freedesktop/NetworkManager" }, signal_emitted);
+        watch = client.watch({ path_namespace: "/org/freedesktop/NetworkManager" });
```
martinpitt, line 404: Is that still necessary after #15617? It certainly can't hurt, but does it still make a difference now?

test/verify/check-static-login
```diff
@@ -861,6 +861,8 @@ matchrule = <SUBJECT>^DC=LAN,DC=COCKPIT,CN=alice$
         # by IPv6 address and port
         check_server("[::1]:22", True)
 
+        self.allow_failed_sudo_journal_messages()
```
martinpitt, line 864: This reminds me of #15544 -- it's tempting to just slap on this "ignore sudo", but why? Since we stopped invoking sudo by default, nothing here *should* try to raise privileges?

It also doesn't match the commit message -- this is about sudo, while cockpit/ssh is, well, ssh.

test/common/testlib.py
```diff
@@ -1082,6 +1082,8 @@ def allow_journal_messages(self, *patterns):
     def allow_hostkey_messages(self):
         self.allow_journal_messages('.*: .* host key for server is not known: .*',
                                     '.*: refusing to connect to unknown host: .*',
+                                    '.*: .* host key for server has changed to: .*',
+                                    '.*: host key for this server changed key type: .*',
```
martinpitt, line 1086: Would it be too coarse to just allow `.*host key.*server.*`?

martinpitt (changes requested):
Thanks!

test/common/testlib.py
mvollmer: I'd say so...

test/common/testlib.py
mvollmer: I leave it in for a bit for sanity checking, it is so easy to mess up the match expressions and end up with no journal messages at all...

src/bridge/cockpitpcpmetrics.c
mvollmer: Done.

pkg/networkmanager/interfaces.js
mvollmer: Not necessary anymore, removed.

test/verify/check-static-login
mvollmer: > Since we stopped invoking sudo by default, nothing here should try to raise privileges?

This change in default comes only with #15293, which is still open.

mvollmer: > It also doesn't match the commit message -- this is about sudo, while cockpit/ssh is, well, ssh.

Hmm, the commit says: `test: Expect sudo failures on remote login`.  Isn't that what you are asking for?  The messages are from sudo, but they happen during remote login.

martinpitt: The subject yes, but it seems odd that adding `SYSLOG_IDENTIFIER=cockpit/ssh` would suddenly show sudo errors? So these are sudo messages from the remote machine, which are being forwarded through ssh?

mvollmer: No, they are in the journal of the remote machine. They have always been there, but `check_journal_messages` did not retrieve them.

test/common/testlib.py
mvollmer: Removed

martinpitt:
Still [several](https://logs.cockpit-project.org/logs/pull-15599-20210406-082844-ee873948-fedora-34/log.html#14) [unexpected](https://logs.cockpit-project.org/logs/pull-15599-20210406-082844-ee873948-ubuntu-2004/log.html#131) messages, plus an annoying amount of "WTH?" unrelated flakes.

martinpitt (commented):
Looks great now, thanks! So with these additional two or three messages silenced, this should be good to go.

martinpitt:
[Two](https://logs.cockpit-project.org/logs/pull-15599-20210406-124750-7da1536b-fedora-34/log.html#301) [more](https://logs.cockpit-project.org/logs/pull-15599-20210406-124750-7da1536b-debian-testing/log.html).. I figure we'll get some more of these in the near future, but let's at least address the known ones. Sorry for the whack-a-mole!

test/verify/check-connection
```diff
@@ -737,6 +737,7 @@ until pgrep -f cockpit-bridge.*--privileged; do sleep 1; done
         self.allow_journal_messages("Refusing to render service to dead parents.")
         self.allow_journal_messages(".*No authentication agent found.*")
         self.allow_journal_messages(".*Peer failed to perform TLS handshake.*")
+        self.allow_journal_messages("cannot reauthorize identity\\(s\\): unix-user:admin unix-user:fedora")
```
martinpitt, line 740: FYI: this is easier written with raw strings: `r"foo\(..."`. (No need to push again, of course)

martinpitt (approved):
Cheers! Ignoring that pesky valgrind hang, it succeded before and is unrelated.

PR #15600, opened 2021-03-26
Remove unused slider code from lib/patternfly and shell.scss

KKoukiou:
PF4 offers a Slider component and that should be used if/when needed.


martinpitt (approved):
Danke!

PR #15601, opened 2021-03-26
tls: split out certificate handling code

allisonkarlitskaya:
Create a new file — certificate.c — and split the code for loading the certificate file into it.

The main benefit of this change is that we will be able to use the same code to help check for the existance of the certificate from our new remotectl replacement.

Use the chance to add a wrapper structure around the certifcate data with support for refcounting.  This will come in handy later.


src/tls/connection.c
```diff
@@ -624,6 +669,7 @@ connection_handshake (Connection *self)
           return false;
         }
 
+      gnutls_session_set_verify_function (self->tls, verify_peer_certificate);
```
martinpitt, line 674: Nice! I wonder why I didn't do that back then .. perhaps we still had to support a too old GnuTLS version?

martinpitt (approved):
That by itself doesn't do much, but I see the follow-up PRs -- so obviously it makes sense for them. Thanks for the auxiliary fixes, too!

PR #15602, opened 2021-03-26
tls: set permissions of 10-ipa.cert on install

allisonkarlitskaya:
This was never being explicitly done, under the assumption that the
certificate file, as created by IPA, would be world-readable by default.
It's not, so let's do it ourselves.

The reason this is working is because remotectl is coming along after
the fact and setting the permissions, incorrectly (under the assumption
that there was also a private key in there), but in a way that would
allow cockpit-tls to load it, at least.

Remove that code from remotectl now as well: we will soon stop running
remotectl in the normal case.

release blocker because: this is a bug that we should fix before the new ipa-certificate-creating script appears in its first release.

### DIFF
--- a/src/ws/remotectl-certificate.c
+++ b/src/ws/remotectl-certificate.c
@@ -193,8 +193,7 @@ ensure_certificate (const gchar *user,
 
   /* adjust permissions of the certificates that we manage automatically;
    * don't touch admin-provided certs, they are often shared between multiple services */
-  if (g_str_has_suffix (path, "/0-self-signed.cert") ||
-      g_str_has_suffix (path, "/10-ipa.cert"))
+  if (g_str_has_suffix (path, "/0-self-signed.cert"))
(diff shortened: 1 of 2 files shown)


src/ws/remotectl-certificate.c
```diff
@@ -193,8 +193,7 @@ ensure_certificate (const gchar *user,
 
   /* adjust permissions of the certificates that we manage automatically;
    * don't touch admin-provided certs, they are often shared between multiple services */
-  if (g_str_has_suffix (path, "/0-self-signed.cert") ||
-      g_str_has_suffix (path, "/10-ipa.cert"))
+  if (g_str_has_suffix (path, "/0-self-signed.cert"))
```
martinpitt, line 196: I'm glad to see this go -- and here's to getting rid of the other half, too :)

martinpitt (approved):
Danke!

PR #15604, opened 2021-03-26
shell: let's use the NavGroup && NavItem components

KKoukiou:


KKoukiou:
Ok I realized now why @marusak did not use the original. It's because of these nested buttons inside NavItem in the host switcher. This is not ready to be ported to the upstream components.

PR #15611, opened 2021-03-29
test: Centralize test decorator for distropkg

martinpitt:
It is annoying and noisy having to bump the same thing in umpteen tests
each time we add or drop a new -distropkg image. Introduce a new
@skipDistroPackage decorator and use that everywhere.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -60,6 +60,7 @@ __all__ = (
     'nondestructive',
     'no_retry_when_changed',
     'skipImage',
+    'skipDistroPackage',
     'skipBrowser',
     'skipPackage',
     'enableAxe',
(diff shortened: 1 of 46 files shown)


martinpitt:
FTR, if anyone knows how to make this a non-functional decorator (`@skipDistroPackage` instead of `skipDistroPackage()`), please tell me. I've tried for half an hour, and I am just too stupid for it... It works just fine for the "run" case, but the "skip" one always fails with `TypeError: decorator() missing 1 required positional argument: 'test_item'` deep in unittest's loader.

marusak:
> if anyone knows how to make this a non-functional decorator (@skipDistroPackage instead of skipDistroPackage()), please tell me

hmm, not me. I've tried as well and does not work what I though that should. But this is completely fine so no need to dwell on that.

marusak:
I've triggered `rhel-8-5-distropkg` so that we actually run some distropkg context to verify this.

marusak (approved):
Thanks!
Please lets make sure distropkg context works. Also would be nice if we would test this in test/verify/check-testlib

martinpitt:
@marusak : Oops, thanks for triggering the test -- indeed it seems I sent this before the bots PR landed which enabled them by default.

PR #15612, opened 2021-03-29
test: Reset allowed_journal_messages between test cases

mvollmer:
Appending to the list would modify the class variable, which is shared
between all instances of MachineCase.  Thus the allow_journal_messages
call from one test would still be active in the next test.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -818,6 +818,8 @@ class MachineCase(unittest.TestCase):
         return int(v[0]) < version
 
     def setUp(self, restrict=True):
+        self.allowed_messages = self.default_allowed_messages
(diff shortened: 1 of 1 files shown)


martinpitt:
[These failures](https://logs.cockpit-project.org/logs/pull-15612-20210329-080815-a57db54d-fedora-33-firefox/log.html#25-2) are our current top flakes, retrying.

martinpitt (approved):
Whoops, well spotted! This may uncover some new test flakes due to "new" unexpected messages, but we'll have to find and fix these. Danke!

mvollmer:
Interesting, the luks firefox failure is not actually a timed out wait for the expected UI element, but: `Debugger: expected Debugger.Object, got Proxy`.

PR #15614, opened 2021-03-29
test: Add special cases for rhel-9-0

martinpitt:
Image added in https://github.com/cockpit-project/bots/pull/1859

 - [x] Builds on top of #15616
 - [ ] Fix missing dhclient and naughties: https://github.com/cockpit-project/bots/pull/1860

### DIFF
--- a/test/verify/check-storage-vdo
+++ b/test/verify/check-storage-vdo
@@ -27,6 +27,7 @@ from testlib import *
 SUPPORTED_OS = ["centos-8-stream", "rhel-8-4", "rhel-8-4-distropkg", "rhel-8-5", "rhel-8-5-distropkg"]
 
 
+@skipImage("no vdo package yet on RHEL 9, https://bugzilla.redhat.com/show_bug.cgi?id=1916216", "rhel-9-0", "rhel-9-0-distropkg")
 class TestStorageVDO(StorageCase):
 
     def testVdo(self):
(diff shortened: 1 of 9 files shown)


martinpitt:
I did a first pass through the failures, sent a bots PR to fix the image/naughties, and pushed some fixes here.

martinpitt:
I went through the previous round's 4 failures, and locally everything succeeded for me. [TestLogin.testSELinuxRestrictedUser ](https://logs.cockpit-project.org/logs/pull-15614-20210330-051315-637dcdf0-rhel-9-0-bots-1860/log.html#293) still fails on CI.

I tried to run all check-static-login tests in a local persistent machine, that worked also.

Can we debug this separately, and land this as incremental improvement?

test/verify/check-storage-vdo
```diff
@@ -27,6 +27,7 @@ from testlib import *
 SUPPORTED_OS = ["centos-8-stream", "rhel-8-4", "rhel-8-4-distropkg", "rhel-8-5", "rhel-8-5-distropkg"]
 
 
+@skipImage("no vdo package yet on RHEL 9, https://bugzilla.redhat.com/show_bug.cgi?id=1916216", "rhel-9-0", "rhel-9-0-distropkg")
```
marusak, line 30: Hmm, This would be better as naughty I guess as we we almost certain to forget this here and not test even when vdo will be available.

marusak: Ah right, just read the BZ. This is fine and when we get reply that it will be build then we should move it to naughty.

marusak (approved):
Nice progress, thanks!
Sure we can land this and fix up those remaining two separately.

martinpitt:
@marusak: The TestIPA failure is "just" a normal flake, #15618 helps with that.

PR #15615, opened 2021-03-29
Restyle updates to have status and settings cards

marusak:
newstuffpackagekit

notavailable

notinstalledpckkit

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -30,6 +30,7 @@ import {
     DescriptionList, DescriptionListTerm, DescriptionListGroup, DescriptionListDescription,
     Flex, FlexItem,
     Spinner,
+    Stack, StackItem,
     Page, PageSection, PageSectionVariants,
     Text, TextContent, TextListItem, TextList, TextVariants,
 } from '@patternfly/react-core';
@@ -44,7 +45,7 @@ import {
 import { cellWidth, TableText } from "@patternfly/react-table";
 import { Remarkable } from "remarkable";
 
-import { AutoUpdates, AutoUpdatesBody } from "./autoupdates.jsx";
+import { AutoUpdates, getBackend } from "./autoupdates.jsx";
 import { History, PackageList } from "./history.jsx";
 import { page_status } from "notifications";
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
@@ -703,7 +704,7 @@ const UpdateSuccess = ({ onIgnore, openServiceRestartDialog, openRebootDialog, r
     </>);
 };
 
-const StatusCard = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages, onValueChanged }) => {
+const UpdatesStatus = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages, onValueChanged }) => {
     const numUpdates = Object.keys(updates).length;
     const numSecurity = count_security_updates(updates);
     const numRestartServices = tracerPackages.daemons.length;
@@ -720,7 +721,7 @@ const StatusCard = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages
             notifications.push({
                 id: "security-updates-available",
                 stateStr: cockpit.format(stateStr, numSecurity),
-                icon: <span id="icon" className={PK.getSeverityIcon(highestSeverity, undefined, "fa-lg")} />,
+                icon: <span id="icon" className={PK.getSeverityIcon(highestSeverity)} />,
                 secondary: <Text id="last-checked" component={TextVariants.small}>{lastChecked}</Text>
             });
         } else {
@@ -730,7 +731,7 @@ const StatusCard = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages
             notifications.push({
                 id: "updates-available",
                 stateStr: cockpit.format(stateStr, numUpdates, numSecurity),
-                icon: <span id="icon" className={PK.getSeverityIcon(highestSeverity, undefined, "fa-lg")} />,
+                icon: <span id="icon" className={PK.getSeverityIcon(highestSeverity)} />,
                 secondary: <Text id="last-checked" component={TextVariants.small}>{lastChecked}</Text>
             });
         }
@@ -738,7 +739,7 @@ const StatusCard = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages
         notifications.push({
             id: "system-up-to-date",
             stateStr: STATE_HEADINGS.uptodate,
-            icon: <CheckIcon color="green" size="md" />,
+            icon: <CheckIcon color="green" />,
             secondary: <Text id="last-checked" component={TextVariants.small}>{lastChecked}</Text>
         });
     }
@@ -748,8 +749,8 @@ const StatusCard = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages
         notifications.push({
             id: "packages-need-reboot",
             stateStr: cockpit.format(stateStr, numRebootPackages),
-            icon: <RebootingIcon id="icon" size="md" />,
-            secondary: <Button variant="danger" onClick={() => onValueChanged("showRebootSystemDialog", true)} isInline>
+            icon: <RebootingIcon />,
+            secondary: <Button variant="danger" onClick={() => onValueChanged("showRebootSystemDialog", true)}>
                 {_("Reboot system...")}
             </Button>
         });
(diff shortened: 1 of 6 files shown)


marusak:
Pushed one more commit. Now empty card is not shown.

pkg/packagekit/updates.jsx
```diff
@@ -771,24 +773,22 @@ const StatusCard = ({ updates, highestSeverity, timeSinceRefresh, tracerPackages
...
     return (<>
         { notifications.map(notification => (
-            <Flex direction={{ default: 'column', sm: 'row' }} key={notification.id} id={notification.id} className="status-notification">
-                <FlexItem>{notification.icon}</FlexItem>
-                <Flex flex={{ default: 'flex_1' }} direction={{ default: 'column' }}>
-                    <FlexItem>
-                        <Text id="state" component={TextVariants.p}>{notification.stateStr}</Text>
-                    </FlexItem>
-                    <FlexItem>
+            <Split key={notification.id} id={notification.id} hasGutter>
```
KKoukiou, line 784: This is breaking sometimes:
Screen Shot 2021-03-29 at 16 32 30

Try:
`<Flex flexWrap={{ default: 'nowrap' }}>`
Seems to work.

KKoukiou (changes requested):
I am not 100% happy with the status card being juts 500% width and aligned left in case the autoupdates is missing, but also it's too little info to expand it to the fill the page horizontally, so well.
Ane comment basically here.

garrett:
1. What's that weird `i` icon? 
2. And why is the automatic updates not being set up in the status card in the first place? It's already in the enabled/disabled part of the settings. Not everyone wants automatic updates. (That's fine if they do not. There are good reasons for only doing manual updates.)

garrett (changes requested):
1. What's that weird `i` icon? 
2. And why is the automatic updates not being set up in the status card in the first place? It's already in the enabled/disabled part of the settings. Not everyone wants automatic updates. (That's fine if they do not. There are good reasons for only doing manual updates.)

garrett:
Alignment of the edit button is... unknown. It's not centered and not even baseline; not sure what it is. It should be vertically centered.

Screenshot from 2021-03-29 18-20-59

garrett:
Heading and (what should be) the description list label aren't aligned left. Or if they are, the label has margin/padding it should not have.

Screenshot from 2021-03-29 18-27-47

Perhaps this is related to the label / text size, if it's using a different widget?

marusak:
> What's that weird i icon?

InfoIcon. Found it less disruptive than the [proposed blue icon](https://github.com/cockpit-project/cockpit-design/blob/559f08327c6b17d353f914ff86e70b718937dab7/software-updates/software-updates-kpatch.png)

> And why is the automatic updates not being set up in the status card in the first place

Right, so when  automatic updates are not set up or are disabled, then no status should be shown. I'll change it.

> Why is the disabled/enabled text small?

No reason in particular :) Would it be preferable to have the name bold and then status as normal text?

> Alignment of the edit button is... unknown. 

Button is the height of the parent and the text is top aligned (with padding). I'll align the text to be in the center. (Hope we can do this with PF)

pkg/packagekit/updates.jsx
marusak: How does it break?

marusak:
> I am not 100% happy with the status card being just 50% width and aligned left in case the autoupdates are missing

Fair point, now looks like this:
Screenshot from 2021-03-30 08-08-53

marusak:
> Heading and (what should be) the description list label aren't aligned left. Or if they are, the label has margin/padding it should not have.

Well, it is defined [here](https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/ct-card.scss#L28).
I am not sure why we have it there but it causes this behavior (also on selinux, users... pages)

marusak:
> Well, it is defined [here](https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/ct-card.scss#L28).
I am not sure why we have it there but it causes this behavior (also on selinux, users... pages)

So I looked why we have it and it seems it is for cases when cards contain list. Then there is no padding for content. Now it looks like this:
withpadding

And if I remove this rule it looks like this:
withhouthepadding

And on cards that have normal body it looks now like this:
withthepaddingdf

And when I remove it:
withouthepaddingsf


So here is my thinking:
1. Do we really need it? It is one more custom css rule that is causing troubles.
2. If it is better this way and we are convinced about it, we should send PR upstream to PF. It is not breaking feature that we cannot release without it. And we can see why PF designed it like this.
3. If we have to have it and we cannot ask PF to do the same, then we should either limit it only when the card contains list or also edit card body padding.

@garrett and @KKoukiou WDYT?

garrett:
Removing it fixes the left side, but breaks the top, right, and bottom padding. PatternFly is incorrect and inconsistent here with the interior spacing. And just setting it all to md is also wrong, as that breaks the left side. The padding really should be `padding: var(--pf-global--spacer--md) var(--pf-global--spacer--md) var(--pf-global--spacer--md) var(--pf-global--spacer--lg)` so that the space on all sides is correct with respect to standard widget spacing in PF basically everywhere else.

We should:
1. Add an override to fix this.
2. Report upstream.

As it changes the spacing, they'll probably even consider it a "breaking change", which means it'll take a while longer to get into PatternFly, hence us needing to have a local override.

marusak:
> What's that weird i icon?

Better?
iicaptain

Or this one?
InfoCircleIcon

Or whatever from https://patternfly-react.surge.sh/icons/?

marusak:
Some updates:

With almost everything there can be:
busybusy

When `dnf-automatic` is not installed:
notsetup

When automatic updates are not supported: (twice)
alonewithstiff
Screenshot from 2021-03-30 14-21-56

When disabled:
notbeing used

marusak:
> Security notice about frequency should be in the configuration card. It's not a status, but a summary of what the configuration will do.

Right, so it should always be in `Settings` and instead on `Enabled` it would say this string? Or can I get design for it?

garrett:
The updates information should be like the mockup that redesigned the Software page:

auto-updates

Except, that when designing kpatch, we realized the configuration area would also have to have kpatch settings, which means that it needs to turn into a description list.

Other than that, the area should have a summary of what is going to happen, or it should say "Disabled" or "Not set up" (when dependencies are not installed).

garrett:
When Andreas and I were working on the kpatch wireframe, we glossed over the settings area; what's there was thought to be a placeholder, not a regression in design. Apologies.

I thought I had a kpatch design that combined it all, but I don't see it locally... at least in the design repo I can mash them up and make another if needed.

garrett:
Cart with a title; the body has a description list, with the description titles titling each section and the descriptions showing a summary of what will be done. The edit buttons pop up dialogs for editing the automatic updates and kpatch settings.

software-updates-overview-settings

This is a medium-fidelity mockup; the padding, style, and spacing is not perfect. PatternFly would ideally do the right thing with the correct widgets.

garrett:
Oh, an additional note: When a kernel patch is applied, it should still show up in the status area, as it's affecting the system. Whereas the auto-updates are _not_ actually a status, but something that is scheduled to happen.

garrett:
Here's an updated software overview to show the context:

software-updates-overview

marusak:
@garrett 

notsetup11

disabled

setuoped

garrett:
Looks great!

I ran out of time to test this today (so I'll have to test it tomorrow), but I like what I see in the screenshots.

pkg/packagekit/updates.jsx
KKoukiou: Ah sorry that I did not answer, missed the notification. I just meant the alignment of the icon with the text on the right, I guess you fixed it already according to the screenshots.

KKoukiou (commented):
There are still some test failures on the page, https://logs.cockpit-project.org/logs/pull-15615-20210412-123303-63220073-debian-stable/log.html#266-1 

but other than that I am happy with the code changes.

marusak:
> /livetest

no autoupdates on debian

garrett:
> no autoupdates on debian

Thanks for letting me know. I'll keep checking it in Fedora/Debian VMs locally then. :grin:

marusak:
Hmm, in tests the page does not want to get loaded. Locally it works just fine.

garrett (approved):
LGTM! :+1: 

Here's what I saw:

Fedora: 

Screenshot_2021-04-14 Software Updates - admin fedora(1)

CentOS 8 Stream (which doesn't have descriptions, apparently):

Screenshot_2021-04-14 Software Updates - admin localhost(1)

Debian Testing:

Screenshot_2021-04-14 Software Updates - admin debian

marusak:
Thank you @garrett for the review! @KKoukiou acked code changes so now I just need to figure out why that one test mos-behaves and this is good to go.

marusak:
Ok, tests are now happy. I've retried F34 but that was unrelated [flake](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15615-20210415-073529-c9b6b345-fedora-34/log.html#157-2)
@KKoukiou may I get review/ack?

marusak:
Green, but thanks to LGTM I noticed one bug I introduced. I'll send fix in a minute.

marusak:
One common [flake](https://logs.cockpit-project.org/logs/pull-15615-20210415-093133-2e16ea8e-debian-stable/log.html#203-2), retrying.

marusak:
Green!

KKoukiou (approved):

PR #15616, opened 2021-03-29
test: Generalize some OS special cases

martinpitt:
Use name prefix matching for tests which clearly apply to all
RHEL/CentOS images by design choice (as opposed to "specific bugs").

Invert the feature/bug OS lists in check-machines-create, so that it
captures the old behaviour and go to zero, as opposed to having to be
extended with each new OS (which is error prone).

----

Prerequisite for fixing tests for RHEL 9 (see #15614)

### DIFF
--- a/test/verify/check-machines-create
+++ b/test/verify/check-machines-create
@@ -1383,7 +1383,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         m.execute("virsh attach-interface --persistent VmNotInstalled bridge virbr0")
 
         # Change the os boot firmware configuration
-        supports_firmware_config = m.image in ['fedora-33', 'fedora-34', 'fedora-testing', 'centos-8-stream', 'rhel-8-4', 'rhel-8-4-distropkg', "rhel-8-5", "rhel-8-5-distropkg", 'debian-testing', 'ubuntu-stable', 'ubuntu-2004']
+        supports_firmware_config = m.image not in ['debian-stable']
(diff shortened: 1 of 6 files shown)


martinpitt:
I accidentally triggered a rhel-9-0 test here (I meant to trigger it in #15614), so we need to ignore that failure here.

test/verify/check-machines-create
```diff
@@ -1383,7 +1383,7 @@ class TestMachinesCreate(VirtualMachinesCase):
         m.execute("virsh attach-interface --persistent VmNotInstalled bridge virbr0")
 
         # Change the os boot firmware configuration
-        supports_firmware_config = m.image in ['fedora-33', 'fedora-34', 'fedora-testing', 'centos-8-stream', 'rhel-8-4', 'rhel-8-4-distropkg', "rhel-8-5", "rhel-8-5-distropkg", 'debian-testing', 'ubuntu-stable', 'ubuntu-2004']
+        supports_firmware_config = m.image not in ['debian-stable']
```
KKoukiou, line 1386: This needs to be synced back to the other repo.

KKoukiou (approved):

martinpitt:
[common flake](https://logs.cockpit-project.org/logs/pull-15616-20210329-131042-5b6473d4-fedora-33-firefox/log.html#325-2), but retrying to good measure.

martinpitt:
Bored retrying the firefox luks failure..

PR #15617, opened 2021-03-29
bridge: Explicitly filter DBus signals by sender when cache watching

mvollmer:
This matters when the watch rules are not sufficiently narrow.  For
example, when there is no rule at all, a cache for o.fd.NetworkManager
would react to PropertiesChanged notifications from o.fd.systemd1.

It would cache and distribute values for systemd1 properties (which is
wasteful but mostly harmless), and would ask o.fd.NetworkManager for
values of systemd properties that have been invalidated (which is
clearly wrong).

### DIFF
--- a/src/bridge/cockpitdbuscache.c
+++ b/src/bridge/cockpitdbuscache.c
@@ -61,6 +61,7 @@ struct _CockpitDBusCache {
   /* The readable DBus name, and actual unique owner */
   gchar *logname;
   gchar *name;
+  gchar *name_owner;
 
   /* Introspection stuff */
   GHashTable *introspected;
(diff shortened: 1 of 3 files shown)


mvollmer:
This gets triggered by `client.watch({ })` in pkg/networkmanager/.

src/bridge/cockpitdbuscache.c
```diff
@@ -1040,6 +1041,12 @@ on_properties_signal (GDBusConnection *connection,
   g_debug ("%s: signal PropertiesChanged at %s", self->logname, path);
   g_variant_get (body, "(&s@a{sv}@as)", &interface, NULL, NULL);
 
+  /* We have to filter the sender ourselves; glib doesn't do that for
+     well-known names.
+   */
+  if (self->name_owner && g_strcmp0 (sender, self->name_owner) != 0)
```
martinpitt, line 1047: FTR, the whole point of strcmp0 is that you don't have to do the NULL check explicitly. (Harmless of course, and hopefully the compiler just optimizes it away)

martinpitt (approved):
Thanks!

src/bridge/cockpitdbuscache.c
mvollmer: It's necessary here since we want to skip the return when name_owner is NULL (very unlikely in practice, but possible).

martinpitt: Of course, totally mis-read that :blush

PR #15621, opened 2021-03-30
Some custom CSS cleanup act

KKoukiou:
Related issue: https://github.com/patternfly/patternfly/issues/3959

### DIFF
--- a/pkg/lib/cockpit-components-table.scss
+++ b/pkg/lib/cockpit-components-table.scss
@@ -3,28 +3,6 @@
 @import "@patternfly/patternfly/components/Table/table-grid.scss";
 
 .ct-table {
-    // Reverse out the default padding for table bodies,
-    // so lists can have more horizontal room
-    &-wrapper {
-        // Container padding is defined higher up; it cascades here
-        // (so this can always be in sync to reverse out the actual padding)
-        --outer-padding: var(--continer-padding-x);
-        --padding-offset-border: 1px;
-        --padding-offset-x: calc(var(--pf-global--spacer--lg) + (var(--padding-offset-border) * 2));
-        --padding-offset-y: 1rem;
-        max-width: calc(100vw - (var(--outer-padding) * 2));
-        overflow: auto;
-        clear: both;
-        padding-bottom: var(--padding-offset-y);
-        margin-bottom: calc(var(--padding-offset-y) * -1);
-        margin: calc(var(--padding-offset-y) * -0.5) calc(var(--padding-offset-x) * -1) calc(var(--padding-offset-y) * -1);
-
-        // Apply standard padding to the sides of the header
-        > .ct-table-header {
-            padding: 0 var(--pf-global--spacer--md);
-        }
-    }
-
     &.pf-m-compact {
         > thead, > tbody {
             > tr:not(.pf-c-table__expandable-row) {
--- a/pkg/lib/cockpit-components-table.scss
+++ b/pkg/lib/cockpit-components-table.scss
@@ -26,11 +26,6 @@
     }
 
     &.pf-m-compact {
-        // Remove excess padding from compact tables toggles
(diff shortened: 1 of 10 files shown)


KKoukiou:
This is not ready and it breaks alignment in mostly all cards. I opened an issue to PF4 reporting the reason (IMO) we use the override here. Let's fix it properly in PF4 overrides file and make sure to remove the override once it's fixed upstream.

garrett:
Note: Right now, this doesn't fix the spacing around the header. There's too much on the top, right of the button, and below.

Top: Inconsistent, also no correct for the button
Right: Not correct for the button
Bottom: Duplicate space, as the body adds space too
Left: Correct; the text aligns with text in the body

KKoukiou:
> Note: Right now, this doesn't fix the spacing around the header. There's too much on the top, right of the button, and below.

The two much space should be fixed with using isCompact property on cards. It's not really about too much space here, it's about misaligned elements.

> 
> Top: Inconsistent, also no correct for the button
> Right: Not correct for the button
> Bottom: Duplicate space, as the body adds space too
> Left: Correct; the text aligns with text in the body

garrett:
> It's not really about too much space here, it's about misaligned elements.

Too much space also causes misaligned elements, so it's about both.

It's misaligned to the vertical rhythm and, more obviously, horizontally misaligned to the content below it. (Again: What we have now is misaligned to the left, and this fixes it, but ignores the vertical issues and also the horizontal issue for the button.)

KKoukiou:
I stole the toolbar in card header fix from https://github.com/cockpit-project/cockpit/pull/15583, just to avoid confusion that something is broken in machines page because of the changes here.

KKoukiou:
This is ready for review, @marusak @garrett please take a look when you have time.

marusak:
It i indeed better in general I found a few cases though. They might be unrelated, or it might be that we are using something wrongly there. I haven't investigate it yet.

Side panels seem wrong. While Filesystem and NFS mounts are fine, Devices and Drives have content too left (or heading too right)
sidepanel

System modifications are fine but content in occurrences is too left.


Here it seems that content is too right.
tooleft

KKoukiou:
@marusak fixed your comments. Then went forwards and deleted some more custom things.

marusak:
Awesome!
Nice cleanup indeed \o/
I haven't noticed any issue but I would like @garrett to take a look as well.
Commits could do with a bit of squashing and of course we should run tests.

KKoukiou:
> Awesome!
> Nice cleanup indeed \o/
> I haven't noticed any issue but I would like @garrett to take a look as well.
> Commits could do with a bit of squashing and of course we should run tests.

Cool, then I will squash and leave it here for next week @garrett to review as well.

garrett (dismissed):
Storage tables do not line up with the header. However, they do line up in Networking, Firewall, and Machines.

Screenshot from 2021-04-08 18-19-26

Interestingly enough, on Storage, for the side cards, it does line up:

Screenshot from 2021-04-08 18-20-00

I think that's all I have at this point. Fix this, and I think it's good to merge.

Congrats on the massive amount of now-unneeded legacy code removed! :partying_face:

KKoukiou:
@marusak this is ready, can you take another look and then we can land, since it's blocking one more PR. :)

marusak (approved):
Thanks!

KKoukiou:
The tracer failure is unrelated, I retried already once. The test succeded in other distros, see here https://logs.cockpit-project.org/logs/pull-15621-20210409-100022-834fc173-rhel-8-5/log.html

marusak:
> The tracer failure is unrelated, I retried already once. The test succeded in other distros, see here https://logs.cockpit-project.org/logs/pull-15621-20210409-100022-834fc173-rhel-8-5/log.html

I am looking into source and there should be easy fix for that. I'll try to reproduce this.

PR #15622, opened 2021-03-30
test: Centos-8-stream now has tracer

skobyda:
Fixes https://github.com/cockpit-project/bots/pull/1861

### DIFF
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -37,7 +37,7 @@ done
 """
 
 OSesWithoutTracer = ["debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable", "ubuntu-testing",
-                     "fedora-coreos", "centos-8-stream"]
+                     "fedora-coreos"]
 
 
 class NoSubManCase(PackageCase):


skobyda:
Trigerred `centos-8-stream@bots#1861`. Let's see now if `testBasic` and `testSecurityInfo` will fail on `centos-8-stream` but pass on `centos-8-stream@bots#1861`, as expected.

martinpitt:
@skobyda : You did not actually trigger that centos-8-stream@bots#1861 task. Did that throw any error? Happy to do it for you of course, but we should sort that out for you?

skobyda:
That's weird. So the tracer is available on centos-8-stream, but from the failures it looks like it doesn't return any results? Gotta investigate futher

skobyda:
So I try to reproduce what we are doing in the tests, and also connect the VM to network and reinstall some packages independetly (like firewalld), which should cause tracer to report something. So it's very like a bug.
So first thing I will do in the morning is create a bug report a ping a tracer guy since this seems like a big thing.

marusak:
Sent upstream patch https://github.com/FrostyX/tracer/pull/165
Now I'll think what we want to actually do on our side - if some hack or just naughty.

marusak:
`centos-8-stream` will fails. `centos-8-stream@bots#1861` has unrelated machines failure. I'll retry just for hack of it. [log](https://logs.cockpit-project.org/logs/pull-15622-20210406-145737-29f23575-centos-8-stream-bots-1861/log.html)
We will also need to backport this into 8.4 branch but I'll do that once this lands.
It needs to land in lockstep with https://github.com/cockpit-project/bots/pull/1861 which makes it that bit more awkward.

marusak (approved):

martinpitt (approved):
Danke! Test is still running, but mostly done.

Issue #14010, opened 2020-05-04
The times displayed by Cockpit are confusing re timezones


mvollmer:
This comes from https://bugzilla.redhat.com/show_bug.cgi?id=1824144

There is potential for confusion when Cockpit displays absolute times, such as with systemd timer units. It is not clear relative to what timezone the display should be interpreted.

Cockpit could help by explicitly displaying the timezone. People have also asked for being able to set the timezone that Cockpit uses to display times.

![image](https://user-images.githubusercontent.com/3228183/80940156-8796f300-8de7-11ea-8ceb-9b5b1ae1d275.png)

![image](https://user-images.githubusercontent.com/3228183/80940219-c0cf6300-8de7-11ea-8dcc-28457848f4e2.png)

![image](https://user-images.githubusercontent.com/3228183/80940252-e197b880-8de7-11ea-9cae-b8d5f508be97.png)

mvollmer:
@andreasn @garrett what do you think?

martinpitt:
Indeed there's a possible mix between the browser's time zone (using moment) and the server's time zone -- I'm not sure if all of our time stamps are formatted client  side using moment, or if we get some pre-formatted stamps from the server.

It feels a bit like clutter to me to repeat the time zone for every time that we display, but we could at least have a PoC how it would look like.

garrett:
## Discussion

The more we can reduce the clutter, the better:
- We're showing seconds where it doesn't really matter. 
- In many cases, we should probably not show the year either (unless it's different from current).

I'd suggest a standard date formatter that makes dates as short as possible, but provide a hover+focus state where there's a way to get the complete date, including timezones. (I guess we're using moment for this, but might be nice to wrap it to have a consistent, simple API for everything to use.)

If we had a formatter like this, we could specify that timezones are wanted (for some cases, like failed log in attempts). It _feels_ like we want timezones for login times, as that's from a user and can be done anywhere. 

**Example**: If someone is logging in on my account at a weird time, I'd like to know. If it says 11:42 AM, and that's plausable in my timezone, it was probably a typo. If it says 11:42 AM (PST), I might think a moment and realize that's 20:42 here — and it wasn't me (unless I was working late). If Cockpit shows me 20:42 (CEST), then I _know_ it wasn't me without even having to think about it.

**So**: I think showing in browser's local time is almost always what we want to do. And clarify the timezone for logins.

## Text "mockups"

Examples:

> There was 1 failed login attempt since the last successful login. 
[...]

PR #15626, opened 2021-03-31
test: Run browser in a different timezone

### LINKED ISSUE #1957: Run avocado phantomjs process in different timezone
In order to catch the same bugs that the old tests used to, the phantomjs should be run in a different timezone than cockpit and the rest of the server.

This is important for any test that touches the timezone or any date value.

marusak:
Using Kolkata as that is 5 hours and 30 minutes ahead of UTC thus will
provide unexpected dates if we use something incorrectly.

Fixes #1957

When I run chromium with TZ it produces: (Notice computer time, system time in cockpit and last login time which normally would be the same as current time).
timeZ

Running tests to see if and what this breaks.

### DIFF
--- a/test/common/cdp.py
+++ b/test/common/cdp.py
@@ -269,6 +269,8 @@ class CDP:
             environ = os.environ.copy()
             environ["HOME"] = self._browser_home
             environ["LC_ALL"] = "C.UTF-8"
+            # Run in Kolkota time zone that is shifted by +5:30 to GMT
+            environ["TZ"] = "Asia/Kolkata"
             # this might be set for the tests themselves, but we must isolate caching between tests
             try:
                 del environ["XDG_CACHE_HOME"]
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -169,7 +169,7 @@ class TestHistoryMetrics(MachineCase):
         # VM just started, we don't have 12 hours of data
         b.wait_in_text(".metrics .pf-c-alert", "No data available between")
         # initial data gap is < 24 hours, does not show date
-        year = m.execute("date +%Y").strip()
+        year = m.execute("TZ='Asia/Kolkata' date +%Y").strip()
         self.assertNotIn(year, b.text(".metrics .pf-c-alert"))
 
         # can try to load earlier data; only updates "no data" alert as there is no data
--- a/test/verify/check-system-journal
+++ b/test/verify/check-system-journal
@@ -28,7 +28,7 @@ sleep_crash_list_sel = "#journal-box .cockpit-logline .cockpit-log-message:conta
 class TestJournal(MachineCase):
     def setUp(self):
         super().setUp()
-        self.test_start_time = self.machine.execute("date +%s").strip()
+        self.test_start_time = self.machine.execute("TZ='Asia/Kolkata' date +%s").strip()
 
     def injectExtras(self):
         self.browser.inject_js("""
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -277,8 +277,8 @@ Unit=test.service
         b.wait_text(self.svc_sel('test-onboot.timer') + ' .service-unit-triggers', '')
         run_cmd(f"systemctl {params} start test-onboot.timer")
         # Check the next run. Since it triggers 200mins after the boot, it might be today or tomorrow (after 20:40)
-        today_stamp = int(m.execute("date +%s").strip())
-        today_plus_200min = m.execute(f"date --date=@{today_stamp + 200 * 60} '+%b %-d, %Y'").strip()
+        today_stamp = int(m.execute("TZ='Asia/Kolkata' date +%s").strip())
+        today_plus_200min = m.execute(f"TZ='Asia/Kolkata' date --date=@{today_stamp + 200 * 60} '+%b %-d, %Y'").strip()
         b.wait_in_text(self.svc_sel('test-onboot.timer') + ' .service-unit-next-trigger', today_plus_200min)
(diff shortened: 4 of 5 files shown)


marusak:
Vm snapshots testing failed with some comparing of time. Lets see if this is wrong in Cockpit or just the test is not flexible enough.

martinpitt:
@marusak: That `checkTimeDiff(time1, time2, 60))` fails all the time on master as well, just perhaps less often.

marusak:
Ha, does it? Here it failed in 4 different OSes. I'll take this as opportunity to take a look and hopefully make it more robust then.

marusak:
Still failed once, lets debug this https://logs.cockpit-project.org/logs/pull-15626-20210331-095121-b5666e0a-debian-testing/log.html#327-1

martinpitt:
Looks fine, thanks! But this somehow seems to break `TestAccounts.testBasic`?

marusak:
Unrelated. I'll repush though once https://github.com/cockpit-project/cockpit/pull/15636 lands so we don't have confused weather.

martinpitt (approved):
Force-pushed a clean rebase to master to re-run the tests. Thanks!

martinpitt:
It's a bit suspicious that [TestShutdownRestart](https://logs.cockpit-project.org/logs/pull-15626-20210401-184427-f75e8553-fedora-33-firefox/log.html#306-2) failed on both fedora-33 scenarios in the same way. Maybe because they ran at the same time?

marusak:
It actually might be an issue as right before that it works with date and time. I have an idea what might be an issue. The theory is that it works in morning and until ~19:00 UTC and then does not work.
Right now it passes locally. I'll try in the afternoon as well (as I am going afk right now anyway).

Theory:
In the [logs](https://logs.cockpit-project.org/logs/pull-15626-20210401-184427-f75e8553-fedora-33-firefox/log.html#306-2) I see `Shutdown scheduled for Fri 2021-04-02 19:31:23 UTC, use 'shutdown -c' to cancel.` but it ran on `2021-04-01`. So when we show current date we show it with current timezone but when we parse it pack from the datetime widget we consider it as UTC. (just the date part it seems).

marusak:
It indeed is broken in the evening.
The problem is that we call `date` in `pkg/lib/cockpit-components-shutdown.jsx` and I guess it expects that server and browser are in the same timezone.

marusak:
Time in reboot dialog does not make sense. Here are facts:
- Date is shown in locale
- Time is shown in system time
- When we get time form UI (partially in localized and partially in not localized form) we give to system to tell is diff with current system time

That does not make any sense. There are two options (as I see it):
1. Always show system date and time
2. Always show (both date and time) in current user locale

In both cases we can just compare initial datetime with new datetime and get offset. We don't need to call `date` utility on the system.

@garrett WDYT is the best for users? If my system is in UTC+1 and I am in UTC+5 should I say that I want the system to be rebooted at midnight of my time or system time?

garrett:
I suppose in a lot of cases, especially for passive things such as logs it doesn't make too much of a difference, as long as it's consistent and understandable at which time the event occurred.

However, for active things like scheduling a reboot (including updates & rebooting), someone may want a VM in a different timezone to be synchronized to their own timezone.

Simple example:  A business in Finland has a data center in Germany. They want all their machines to reboot relative to their own local time (not the server's local time).

But... crontabs are local time to the machine.

What if we expand the Finland/Germany story a little? The same Finnish company also have some of their admins in Germany who are in charge of the German data center and a more local Finnish one. They want to reboot the machines relative to Finnish time too. Since they're in charge of machines in both timezones, they want to make sure they're rebooting them when they think they actually are doing so.

So something that acts upon the machine should probably have the time clearly stated in both timezones, when the local and remote timezones do not match. (When they match, it's easy: Just show one.)

But, again, for passive things that have happened, such as logs, just having the machine's time (with a timezone shown when local and remote timezones do not match) probably works.

Does this make sense? WDYT?

garrett:
Oh last login is also tricky. It's from the perspective of the user, but it's also reported as local time on the system too, right? :thinking:

marusak:
> Oh last login is also tricky. It's from the perspective of the user, but it's also reported as local time on the system too, right?

In general I think we show system time only on overview under `System time` and everywhere else we show localized time (I think).

> So something that acts upon the machine should probably have the time clearly stated in both timezones, when the local and remote timezones do not match. (When they match, it's easy: Just show one.)

Hmm, can I have some design for this? (For this reboot dialog)

martinpitt:
Reboot schedule, log events etc. should be shown/specified according to the target system's time. That's what the system API actually does, and these actions apply to the machine, not your browser. It's irrelevant where your browser is. So showing both times (if they differ) may help with the disambiguation, but on places where we show only one time, it should be the target system's time IMHO.

marusak:
I've found https://github.com/cockpit-project/cockpit/issues/14010 which is exactly about this.
This seems that it is going to be much bigger fish to fry that anticipated.

garrett:
I generally agree with Martin. However, a time offset indicator might make sense in some places.

For the journal, it could look like this on the top header area:

journal-tz

And then if you click the label thingy, it could show actual times on both machines (remote and local):

journal-tz-popover

This would reduce any confusion and would allow us to not duplicate the time everywhere in the journal.

martinpitt:
I sent PR #16393 to fix TestShutdownRestart which was one of the breakers here. I also fixed a few similar issues recently, like commit 206e8d76b8b1 . So after #16393 lands, it may be worth giving this another try?

marusak:
I don't plan on working on this any time now so closing. This needs some more extensive work with introducing new components for time.

PR #15627, opened 2021-03-31
test: Drop journal save/restore hacks in TestJournal

martinpitt:
Commits c27aa2601338b and 40f8e0df6d40d built an ever-increasing series
of delicate hacks to restore the journal to the original state, but this
again breaks on RHEL 9: it confuses the audit subsystem, which causes
several system service failures (such as hostnamed, homed, and the
systemd@user service) and eventually breaks testSELinuxRestrictedUser.

Revert that hackery, make testBasic() destructive, and fix the other
nondestructive tests to only consider the journal since the test start
time, so that they can still run multiple times on the same running VM.

-----

### Original description from debugging phase:

`TestLogin.testSELinuxRestrictedUser` was the one remaining [failure](https://logs.cockpit-project.org/logs/pull-15614-20210330-051315-637dcdf0-rhel-9-0-bots-1860/log.html#293) from #15614 . I can't reproduce this locally, and I don't know what's breaking the second login. The first one apparently worked fine, and the second one times out. The journal does not give any hint about it. So, on to some blackbox debugging..

```
  File "/work/bots/make-checkout-workdir/test/verify/check-static-login", line 459, in testSELinuxRestrictedUser
    b.login_and_go("/system", user="unpriv")
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 578, in login_and_go
    self.expect_load()
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 150, in expect_load
    self.cdp.command('expectLoad(%i)' % (self.cdp.timeout * self.timeout_factor * 1000))
  File "/work/bots/make-checkout-workdir/test/common/cdp.py", line 136, in command
RuntimeError: timed out waiting for page load
```

### DIFF
--- a/test/verify/check-system-journal
+++ b/test/verify/check-system-journal
@@ -28,28 +28,7 @@ sleep_crash_list_sel = "#journal-box .cockpit-logline .cockpit-log-message:conta
 class TestJournal(MachineCase):
     def setUp(self):
         super().setUp()
-
-        if self.is_nondestructive():
-            # we need persistent journal for the rebooting tests
-            self.addCleanup(self.machine.execute, "systemctl try-restart systemd-journald")
-            self.restore_dir("/run/log/journal")
-            self.restore_dir("/var/log/journal")
-            self.machine.execute("""
-            UNITS=""
-            for u in systemd-journald.socket systemd-journald-dev-log.socket systemd-journald.service; do
-                if systemctl --quiet is-active $u; then
-                    UNITS="$UNITS $u"
-                fi
-            done
-            systemctl stop $UNITS
-            # avoid running into startup limit
-            systemctl reset-failed $UNITS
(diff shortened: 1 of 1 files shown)


martinpitt:
Interesting, this [succeeded now](https://logs.cockpit-project.org/logs/pull-15627-20210331-070106-8c15296d-rhel-9-0/log.html). Putting back all the nondestructive tests now, to check if it's about timing (noisy neighbors) or about subtly breaking the VM from previous tests.

martinpitt:
Now it's just our old friend [testStorageResize](https://logs.cockpit-project.org/logs/pull-15627-20210331-154518-3386065e-rhel-9-0/log.html), retrying.

marusak (approved):
Nice, thank you!

PR #15630, opened 2021-03-31
test: Fix dnf-copr upgrade scenario

martinpitt:
Don't let dnf install Recommends on upgrade. We don't want that to
install cockpit's recommendations like c-packagekit, as this will make
the `rpm -e` commands from the *.install script fail.

----

Blocks https://github.com/cockpit-project/bots/pull/1828


marusak (approved):
Thanks!

PR #15633, opened 2021-04-01
shell: User needs to relogin to change language

### LINKED ISSUE #8160: language change has no effect in chromium on our test images
While debugging failed tests for the PhantomJS → Chome transition, the `testBasic (check_pages.TestPages)` failure is actually real:

 * Prepare a fedora-27 image (haven't tested others), `test/vm-run fedora-27`, log in, `systemctl start cockpit.socket`
 * In chromium, log into that VM's cockpit at https://127.0.0.2:9091/
 * Select "Display Language" in the menu bar and switch to German
 * UI reloads, but stays in English; showing `cockpit.language` in the JS console still shows "en"

The weird thing is that it *does* work when logging into my laptop's host cockpit (https://localhost:9090), and it *also* works when doing this in Firefox. So the problem isn't unanimously in the test VM or in the browser.

marusak:
Fixes #8160
https://bugzilla.redhat.com/show_bug.cgi?id=1666310

### DIFF
--- a/src/ws/cockpitchannelresponse.c
+++ b/src/ws/cockpitchannelresponse.c
@@ -33,6 +33,7 @@ typedef struct {
   CockpitWebService *service;
   gchar *base_path;
   gchar *host;
+  gchar *lang;
 } CockpitChannelInject;
 
 static void
@@ -46,6 +47,7 @@ cockpit_channel_inject_free (gpointer data)
         g_object_remove_weak_pointer (G_OBJECT (inject->service), (gpointer *)&inject->service);
       g_free (inject->base_path);
       g_free (inject->host);
+      g_free (inject->lang);
       g_free (inject);
     }
 }
@@ -53,6 +55,7 @@ cockpit_channel_inject_free (gpointer data)
 static CockpitChannelInject *
 cockpit_channel_inject_new (CockpitWebService *service,
                             const gchar *path,
+                            const gchar *lang,
                             const gchar *host)
 {
   CockpitChannelInject *inject = g_new (CockpitChannelInject, 1);
@@ -60,6 +63,7 @@ cockpit_channel_inject_new (CockpitWebService *service,
   g_object_add_weak_pointer (G_OBJECT (inject->service), (gpointer *)&inject->service);
   inject->base_path = g_strdup (path);
   inject->host = g_strdup (host);
+  inject->lang = g_strdup (lang);
   return inject;
 }
 
@@ -125,6 +129,21 @@ cockpit_channel_inject_perform (CockpitChannelInject *inject,
   cockpit_web_response_add_filter (response, filter);
   g_object_unref (filter);
   g_free (prefixed_application);
+
+  // Lang hack
+  if (inject->lang)
+    {
+      gchar *po_lang;
+      GBytes *po_inject;
+      po_lang = g_strdup_printf ("?lang=%s", inject->lang);
+      po_inject = g_bytes_new_take (po_lang, strlen (po_lang) );
+      CockpitWebFilter *po_filter = cockpit_web_inject_new ("/po.js", po_inject, 1);
+
+      cockpit_web_response_add_filter (response, po_filter);
+
+      g_bytes_unref (po_inject);
+      g_object_unref (po_filter);
+    }
 }
 
 #define COCKPIT_TYPE_CHANNEL_RESPONSE  (cockpit_channel_response_get_type ())
@@ -516,12 +535,6 @@ parse_host_and_etag (CockpitWebService *service,
   const gchar *accept = NULL;
   gchar **languages = NULL;
   gboolean translatable;
-  gchar *language;
-
-  /* Parse the language out of the CockpitLang cookie and set Accept-Language */
-  language = cockpit_web_server_parse_cookie (headers, "CockpitLang");
-  if (language)
-    g_hash_table_replace (headers, g_strdup ("Accept-Language"), language);
 
   if (!where)
     {
@@ -576,6 +589,7 @@ cockpit_channel_response_serve (CockpitWebService *service,
   const gchar *host = NULL;
   const gchar *pragma;
   gchar *quoted_etag = NULL;
+  gchar *lang = NULL;
   GHashTable *out_headers = NULL;
   gchar *val = NULL;
   gboolean handled = FALSE;
@@ -592,6 +606,11 @@ cockpit_channel_response_serve (CockpitWebService *service,
   g_return_if_fail (in_headers != NULL);
   g_return_if_fail (COCKPIT_IS_WEB_RESPONSE (response));
   g_return_if_fail (path != NULL);
+
+  /* Parse the language out of the CockpitLang cookie and set Accept-Language */
+  lang = cockpit_web_server_parse_cookie (in_headers, "CockpitLang");
+  if (lang)
+    g_hash_table_replace (in_headers, g_strdup ("Accept-Language"), lang);
 
   /* Where might be NULL, but that's still valid */
   if (!parse_host_and_etag (service, in_headers, where, path, &host, &quoted_etag))
@@ -701,7 +720,7 @@ cockpit_channel_response_serve (CockpitWebService *service,
   self = cockpit_channel_response_new (service, response, transport,
                                        out_headers, object);
 
-  self->inject = cockpit_channel_inject_new (service, injecting_base_path, host);
+  self->inject = cockpit_channel_inject_new (service, injecting_base_path, lang, host);
   handled = TRUE;
 
   /* Unref when the channel closes */
--- a/pkg/shell/base_index.js
+++ b/pkg/shell/base_index.js
@@ -185,7 +185,7 @@ function Frames(index, setupIdleResetTimers) {
             frame.url = base + "/" + component;
             if (component.indexOf("/") === -1)
                 frame.url += "/index";
-            frame.url += ".html";
+            frame.url += ".html?lang=" + language;
         }
 
         if (!hash)


martinpitt:
It may be enough to simply show the alert with the button? If the user does not want to log out right now, they at least know why the language did not change right away.

garrett:
1. This isn't a problem in Firefox
2. This doesn't seem to be a problem in Chrome, at least locally
3. The UI is all wrong for this

This sounds like a bad workaround for a browser bug. And if it's not a browser bug, then we're doing something wrong in the code.

garrett:
Is the reported problem due to a proxy?

garrett:
If the problem is with caching on Chrome on a remote server under some mysterious circumstances, it would not be correct to tell someone to log out and back in if they are unaffected.

You could cache a translated string or some strings that would definitely be translated and check to see if they're the same before and after the language switch. If they are the same, then show a message that the system needs to be restarted. But this would have to be outside of the dialog, unless switching languages is a hot setting which changes on selection (and it is not; it needs confirmation).

So, to make it clear:

1. Cache active string of something that will 100% be different when translated
2. User switches languages, confirms in the dialog
3. Check the new active string versus the old cached active string
4. Show a _new modal_ saying that the user must log out to see changes (preferably in the newly selected language _somehow_, regardless of caching), with a log out button.

This stops us from blanket-banning all browsers that _actually work properly_ and it also stops us from browser sniffing (which should _always_ be avoided).

But, as that's still a workaround, better yet:

1. Fix the issue for Chrome :wink:

garrett:
Further Chrome-specific stuff:

If Chrome is being told to reload from the server and it checks the server and gets a 304 modified and gives up, then that's a bug in Chrome. A workaround for Chrome to always load the page is to modify the HTTP headers, I guess?

The findings block @ https://stackoverflow.com/questions/10719505/force-a-reload-of-page-in-chrome-using-javascript-no-cache suggests this:

> After looking into this I have found that this issue is HTTP Protocol handling;
> 
> 1.  Chrome sends a request with `Pragma: no-cache` HTTP field
> 2.  Server responds with `Last-Modified: DATE1` field
> 3.  JS uses `location.reload(true)` to force a reload from server not cache
> 4.  Chrome sends a request with `If-Modified-Since: DATE1` field
> 5.  Server responds with `HTTP Status 304 Not Modified`
> 
> The server application is at fault for not noticing the state change in the dynamic page content, and thus not returning a `200`. However, Chrome/WebKit is the only browser that sends a `If-Modified-Since` field when the JS `location.reload(true)` is called.

But, really, telling the browser to avoid the cache and the browser ignores that directive and fetches from the cache is just broken, if this is what's happening.

martinpitt:
We got an RPL? request in https://bugzilla.redhat.com/show_bug.cgi?id=1666310 for this.

I can't repoduce it with interactive chromium on my system either. But we still see it in the [starter-kit test suite](https://github.com/cockpit-project/starter-kit/blob/master/test/check-application#L51), and have the same workaround in cockpit's check-pages. There it seems a force-reload without cache fixes the issue, so I still think that a less intrusive hack may be something like "if (browser == chrome) force_reload"?

garrett:
>  There it seems a force-reload without cache fixes the issue, so I still think that a less intrusive hack may be something like "if (browser == chrome) force_reload"?

No, that's bad. We should _never_ browser-sniff. Detecting for features and detecting when things do not work is the way to go.

What's Chrome? Is it just Google's? Chromium? Edge? Opera? Brave? All of these could _probably_ fail in _some_ version. It starts to become problematic when checking for browser+version combos.

Why not just use what I suggested above instead?

Basically: Compare a string from before switching languages to after switching languages and if they're the same, then force a logout. That would catch when this is a problem and allow browsers where it's not a problem to just work as expected.

Also: The problem is with force-reload. Some versions of Chrome (and friends?) simply do not force a refresh when you tell them to do so, _causing_ this bug, requiring logging out (or manual shift-reload, possibly).

martinpitt:
> We should never browser-sniff.

OK, fair enough.

> Cache active string of something that will 100% be different when translated

It's this bit which is problematic.

> The problem is with force-reload

You mean when done through the API? As when I do Ctrl+Shift+R in chromium, that totally works.

garrett:
> You mean when done through the API? As when I do Ctrl+Shift+R in chromium, that totally works.

Right, but that's not the problem here. The problem is refreshing the page programmatically, in the browser. We cannot tell people to do shift+ctrl+r or shift+f5 as a workaround.

I did a little more research and found that passing true in `location.reload(true);` has been deprecated (just passing true; the API is otherwise the same and still supported). I guess that's the difference between Chrome and Firefox here. Chrome apparently doesn't respect bypassing the cache, whereas Firefox seemingly still does.

I've seen other places that recommend using `window.location.replace(window.location.href);` (specifically using _replace_ as to not add to the history stack). I know that browsers will still attempt to fetch the location, but am not sure if Chrome (or others) would bypass cache. (Browsers _may_ still try to grab the resources, get a 304 _not modified_ and then use cache instead. We'd have to test.)

The surefire way is to either have the webserver return something other than a 304 or use cache-busting techniques, like appending a query string (even when unneeded).

We _could_ possibly have a language-based query string that's otherwise ignored, I suppose. That would look something like `window.location.replace(window.location.href + '?lang=de');` — but this is an extremely naive approach, just to simply show what would happen. The more correct answer would really be something like this:

```js
let params = new URLSearchParams(window.location.search);

// Set will add or override existing lang values in a query string
params.set('lang', 'de');

// window.location.search always starts with '?' or is an empty string; both cases work here
let newUrl = window.location.href.replace(window.location.search, '?' + params.toString());

// Basically reload the same URL, except with lang sent to the server (which is otherwise ignored)
window.location.replace(newURL);
```

marusak:
> I've seen other places that recommend using `window.location.replace(window.location.href);`

That does not work either.

I'll look into using lang string in query now.

marusak:
> That would look something like window.location.replace(window.location.href + '?lang=de'); — but this is an extremely naive approach, just to simply show what would happen.

Nah, doesn't work either. I see this new query in url, it reloads the page yet is stays in original locale.

garrett:
> Nah, doesn't work either. I see this new query in URL, it reloads the page yet is stays in original locale.

It's literally a different page resource though. It should work.

garrett:
> Nah, doesn't work either. I see this new query in url, it reloads the page yet is stays in original locale.

Is the problem that the language file is not in the main resource URL, but another one?

marusak:
I think that po.js being cached may be very much the issue. I'm looking now how to tell browser to reload this file. In practice it looks simple but does not work just like that. I think we need a bit more logic with all the machinery we have around how page actually ends up translated.

garrett:
Is there a way to have something like a session start time always in the URL for the language file and then bump it at the start of a new session and also a language switch?

It's currently:
```html
<script src="../*/po.js"></script>
```

I'm suggesting something like:
```html
<script src="../*/po.js?sessionStart=2021-04-21T14:10:37.540Z"></script>
```

This is using the output of
```js
new Date().toJSON()
```
...but could be really anything, as long as it changes each time. Having a datetime in there with seconds just ensures the change. It could be `?sessionRev=0` on initlal load, then increment to 1 on a language switch, then 2 on another language switch or whatever as well. However, then there's the possibility of caching issues. Hence the suggestion of making it a time based one so there's no risk of aggressive caching.

The server doesn't have to care about this at all; it's just for the sake of the browser, to let it know there's "another URL" and it shouldn't use the cached version.

martinpitt:
Doesn't that break the actual caching of the file? Like, each page switch/update would have to re-download it?

garrett:
> Doesn't that break the actual caching of the file? Like, each page switch/update would have to re-download it?

That's why I suggestion each session, not each page load. (And bumping the session or whatever indicator for this each time the language changes too.)

It's ~140 kb uncompressed. Downloading it once per session or language switch is OK, I think.

martinpitt:
Ah, I didn't know that the browser would ever cache replies which have GET parameters -- it seems dangerous (and in many/most cases wrong) to assume that parameterized queries are idempotent. (http://what.is.the.time.com?tz=UTC)

garrett:
Browsers totally do cache query strings when possible in a GET request. POST is a little different\*.

https://tools.ietf.org/html/rfc2616#section-9.3:
>    The response to a GET request is cacheable if and only if it meets
>    the requirements for HTTP caching described in [section 13](https://tools.ietf.org/html/rfc2616#section-13).

https://tools.ietf.org/html/rfc2616#section-13.9:

>    Unless the origin server explicitly prohibits the caching of their
>    responses, the application of **GET and HEAD methods to any resources
>    SHOULD NOT have side effects that would lead to erroneous behavior if
>    these responses are taken from a cache**. They MAY still have side
>    effects, but a **cache is not required to consider such side effects in
>    its caching decisions**. Caches are always expected to observe an
>    origin server's explicit restrictions on caching.
> 
>    We note one exception to this rule: since some applications have
>    traditionally used GETs and HEADs with query URLs (those containing a
>    "?" in the rel_path part) to perform operations with significant side
>    effects, **caches MUST NOT treat responses to such URIs as fresh unless
>    the server provides an explicit expiration time**. This specifically
>    means that responses from HTTP/1.0 servers for such URIs SHOULD NOT
>    be taken from a cache. See [section 9.1.1](https://tools.ietf.org/html/rfc2616#section-9.1.1) for related information.

(That is HTTP/1.0 _explicitly_. > 1.0 is different, as there are explicit expiration times that can be given.)

Here's some in-depth guide from Amazon about caching with GET query strings, where they explicitly state it and even use language parameters in a query string as an example of a cached resource.
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/QueryStringParameters.html

\* Note: Even [POST is cachable under certain circumstances](https://tools.ietf.org/html/rfc2616#section-9.5)...
> Responses to this method are not cacheable, _**unless** the response_
>    _includes appropriate Cache-Control or Expires header fields_. However,
>    the 303 (See Other) response can be used to direct the user agent to
>    retrieve a cacheable resource.

---

**TL;DR**: It happens. Browsers can and do cache GET requests with query strings under various circumstances. And it's fairly common to use query strings in a GET request for assets for cache controlling both JavaScript (in this case) and also CSS and images too. (Often this is bumped on build time, but not always.)

And it's common to request languages and cache the results with query strings. (Even if the query does nothing server side.)

src/ws/cockpitchannelresponse.c
```diff
@@ -125,6 +129,21 @@ cockpit_channel_inject_perform (CockpitChannelInject *inject,
   cockpit_web_response_add_filter (response, filter);
   g_object_unref (filter);
   g_free (prefixed_application);
+
+  // Lang hack
+  if (inject->lang)
```
marusak, line 134: Problem with this is that this only works in shell. I have not been able to figure out where and how .html pages for iframes are being handled and server so I could edit their content as well.

@mvollmer do you have some idea?

marusak:
I pushed here what I had working - that being shell being translated when language changes. I still was not able to find how this works for frames themselves.

martinpitt:
#17092 landed which fixes the bug, arguably with a too large :hammer: . @marusak, do you think there's something good here which we can salvage?

marusak:
Nah, I think your solution is good enough, thanks!

Neustradamus:
@martinpitt Thanks for your PR about this old bug.
I hope it is the good one!

Thanks @marusak too.

Issue #15473, opened 2021-03-04
ct-card action alignment does not work properly with links


garrett:
In both of these screenshots, the action does not properly align with the title.

![log-link-to-service](https://user-images.githubusercontent.com/10246/109982604-5b4b4c00-7d02-11eb-84d5-50fb0f342eec.png)

![selinux-restyled](https://user-images.githubusercontent.com/10246/109982609-5be3e280-7d02-11eb-9e98-4598f0e5c51d.png)

If it were a button (and it's not), it should be vertically aligned to the center of the heading. As it does not have the outer border, the baselines should align.

Here are some illustrations showing where the baselines of each item are located:

![Screenshot from 2021-03-04 16-02-32](https://user-images.githubusercontent.com/10246/109984098-bd588100-7d03-11eb-920a-e388e2757f3b.png)

![Screenshot from 2021-03-04 16-05-10](https://user-images.githubusercontent.com/10246/109984100-bdf11780-7d03-11eb-819e-7fd50cee6e87.png)

In each case, the text on the right should use the bottom baseline, not the top.

With a corrected baseline, the screens should look like this:

![log-link-to-service (copy)](https://user-images.githubusercontent.com/10246/109984630-4079d700-7d04-11eb-90a6-5ea9b18304cc.png)

![selinux-restyled (copy)](https://user-images.githubusercontent.com/10246/109984625-3fe14080-7d04-11eb-9eb5-e5ac266b3787.png)

garrett:
Anyway, this isn't necessarily page specific, I think, but more of a problem of ct-card, if it is actually even used on these pages. (We override a few things that PF provides to make the card layout a bit more optimal for us. I think there's alignment for buttons, but not for links, as we've only ever used buttons in the action area so far. I _think_ ct-card is the right place to fix this, possibly with an `align-self` for whatever the `pf-c-link` is called.)

KKoukiou:
This is not related only to ct-card. Having a button link directly on the PF4 documentation will produce the same result;
From the dev tools seems its directly because of:

```
.pf-c-card__actions {
...
    margin: var(--pf-c-card__header-toggle--MarginTop) var(--pf-c-card__header-toggle--MarginRight) var(--pf-c-card__header-toggle--MarginBottom) auto;

``` 
I am not sure what they are trying to do there and they have this weird margins (only the margin-left: auto makes sense to me)

Maybe it makes sense to fix it in a generic way upstream.

garrett:
OK!

[...]

PR #15634, opened 2021-04-01
Remove custom link-button class - use Button variant="link" instead

KKoukiou:


pkg/lib/cockpit-components-logs-panel.jsx
```diff
@@ -158,7 +158,7 @@ export class LogsPanel extends React.Component {
             <Card className="cockpit-log-panel">
                 <CardHeader>
                     <CardTitle>{this.props.title}</CardTitle>
-                    { this.props.goto_url && <CardActions><button className="link-button" role="link" onClick={e => cockpit.jump(this.props.goto_url)}>{_("All logs")}</button></CardActions>}
+                    { this.props.goto_url && <CardActions><Button variant="link" onClick={e => cockpit.jump(this.props.goto_url)}>{_("All logs")}</Button></CardActions>}
```
marusak, line 161: This one seems misaligned. Not sure if that was the case before or not, but i think we can fix it now.
Screenshot from 2021-04-08 10-35-18

marusak (changes requested):
Just one alignment issue.

pkg/lib/cockpit-components-logs-panel.jsx
KKoukiou: All links in card actions are misaglined https://github.com/cockpit-project/cockpit/issues/15473 

i think it would be some CSS override, which will affect so many pages, are you sure you want it part of this?

marusak: Aha, since it is known and there is issue for it, then this is OK. Lets not fix it here.

KKoukiou: (it's the same in master of course)

marusak (approved):
My concern is a known issue. Thanks!

marusak:
`check-shell-multi-machine TestMultiMachine.testSshKeySetup ` does not want to get green. It is affected test thus not retried. Almost certain it is unrelated though

PR #15636, opened 2021-04-01
users: Fix lastlog parsing - April the 1st edition

marusak:
This parsing has been changed in 299e6f6f51 on 23.3.2021.
Therefore we only tested it on dates with double digit dates which
failed to notice that when date is single digit it does not work correctly.

We split the string on space, which produces:
`Tue Mar 23 14:49:04 +0000 2021` => ["Thu", "Mar", "23", "14:49:04", "+0000", "2021"]
`Thu Apr  1 08:58:51 +0000 2021` => ["Thu", "Apr", "", "1", "08:58:51", "+0000", "2021"]

When we then take last 5 fields in the latter case we don't get month
which breaks subsequent parsing.

### DIFF
--- a/pkg/users/account-details.js
+++ b/pkg/users/account-details.js
@@ -68,7 +68,8 @@ function get_last_login(name) {
             return null;
 
         // line looks like this: admin            web cons ::ffff:172.27.0. Tue Mar 23 14:49:04 +0000 2021
-        const date_fields = line.split(' ').slice(-5);
+        // or like this:         admin            web cons ::ffff:172.27.0. Thu Apr  1 08:58:51 +0000 2021
+        const date_fields = line.split(/ +/).slice(-5);
         const d = new Date(date_fields.join(' '));
         if (d.getTime() > 0)
             return d;


marusak:
Current master is broken like [this](https://logs.cockpit-project.org/logs/pull-15626-20210401-052825-994dd582-centos-8-stream/log.html#324)

KKoukiou (approved):
Hey Edge users, sorry for this :) 
Looks good!

pkg/users/account-details.js
```diff
@@ -68,8 +68,9 @@ function get_last_login(name) {
             return null;
 
         // line looks like this: admin            web cons ::ffff:172.27.0. Tue Mar 23 14:49:04 +0000 2021
-        const date_fields = line.split(' ').slice(-5);
-        const d = new Date(date_fields.join(' '));
+        // or like this:         admin            web cons ::ffff:172.27.0. Thu Apr  1 08:58:51 +0000 2021
+        const date_fields = line.replaceAll("  ", " ").split(' ');
```
martinpitt, line 72: How about `line.split(/ +/).slice(-5)` ?

martinpitt (commented):
D'oh! It is infuriating how utterly non-machine parseable the `lastlog` output is. Proposing a more robust fix, but feel free to land this as a quickfix to unbreak other PRs.

pkg/users/account-details.js
KKoukiou: Ah nice, with this we don't need to use replaceAll which is not really supported in older Edge.

marusak: Good call, updated

martinpitt (approved):

PR #15638, opened 2021-04-01
overview: remove some custom styling from our superuser alert 

KKoukiou:
Before:
Screen Shot 2021-04-01 at 14 06 22

After:
Screen Shot 2021-04-01 at 14 18 23

### DIFF
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -138,7 +138,7 @@ class LoginMessages extends React.Component {
             <Alert id='login-messages'
                    variant={this.state.messages.type}
                    isInline={!fail_count_message}
-                   className={!fail_count_message ? "no-title" : ""}
+                   className={!fail_count_message ? "login-messages" : ""}
                    actionClose={<AlertActionCloseButton onClose={this.close} />}
                    title={fail_count_message || last_log_item}
             >
@@ -262,8 +262,8 @@ class OverviewPage extends React.Component {
                         </div>
                     </PageSection>
                     <PageSection variant={PageSectionVariants.default}>
-                        <LoginMessages />
                         <Gallery className='ct-system-overview' hasGutter>
+                            <LoginMessages />
                             <MotdCard />
                             <HealthCard />
                             <UsageCard />
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -243,7 +243,9 @@ class OverviewPage extends React.Component {
                 {this.state.rebootModal && <ShutdownModal onClose={() => this.setState({ rebootModal: false })} />}
                 {this.state.shutdownModal && <ShutdownModal shutdown onClose={() => this.setState({ shutdownModal: false })} />}
                 <Page>
-                    <SuperuserAlert />
+                    <PageSection variant={PageSectionVariants.light} padding={{ default: 'noPadding' }}>
+                        <SuperuserAlert />
+                    </PageSection>
                     <PageSection className='ct-overview-header' variant={PageSectionVariants.light}>
                         <div className='ct-overview-header-hostname'>
                             <h1>
--- a/pkg/systemd/overview.scss
+++ b/pkg/systemd/overview.scss
@@ -5,36 +5,10 @@
 @import "../lib/table.css";
 
 .ct-limited-access-alert {
-  align-items: center;
-  background: var(--pf-global--active-color--200);
-  margin-bottom: 0;
-  padding: 1rem 1.5rem;
-
-  // Override PF4 spacing to better align to the overview page
-  // and rely on ct-limited-access-alert for padding & alignment
-  &.pf-c-alert.pf-m-info.pf-m-inline {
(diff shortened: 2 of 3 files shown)


garrett:
When we made the overrides, the inline alerts looked quite different.

Also, it's more like a banner... https://www.patternfly.org/v4/components/banner/react-demos/basic-sticky-banner/ but PF banners are too small and don't support buttons.

The last login banner clashes with this too. But That should probably be in the system information card under uptime. (IIRC: The login notification was added without consulting either Andreas or myself. It just appeared one day.)

KKoukiou:
> When we made the overrides, the inline alerts looked quite different.

This PR just removes the custom CSS to make the superuser not look different from other alerts at least. 

> 
> Also, it's more like a banner... https://www.patternfly.org/v4/components/banner/react-demos/basic-sticky-banner/ but PF banners are too small and don't support buttons.
>

So let's keep it an info alert.
 
> The last login banner clashes with this too. But That should probably be in the system information card under uptime. (IIRC: The login notification was added without consulting either Andreas or myself. It just appeared one day.)

I agree, and I can move it but separately from this.

garrett:
>> login notification
> I agree, and I can move it but separately from this.

Yep, totally! I didn't mean in this PR; just pointing out that the two notifications clash. Happy you agree!

> This PR just removes the custom CSS

Happy about dropping all the custom stuff that is no longer needed.

---

OK, so I like this PR in general, but I do have the following issues with the results at the moment:

The sides should line up. This would require interior padding on the left and right of the banner/alert. As it doesn't have padding outside (nor should it), we need special casing here. It's on both sides.

Screenshot from 2021-04-08 16-22-54

Here's the right side:

Screenshot from 2021-04-08 16-23-08

Also the vertical alignment is odd. There's 48px above the bottom and 60px below. It should be equal. Some of this is due to the line at the top, which is unnecessary in this case (as it blends in too much with the dark background of shell above). When the line is counted, there is 55px above the button. But with the line, even with equal padding above and below, it may look visually "off". 

This vertical spacing issue is a problem with the text on the left too, but it's easier to see it with the button (and the text and the button should be vertically aligned to one another already).

(There are also some vertical alignment issues with the icons by a few pixels. I think we still have a longstanding issue open in the PF repos about this, so I'm ignoring it here... and it also affects the icons outside the notification too, so it's definitely not specific to this PR.)

garrett:
A fix for the padding on the sides would be to use the PF utilitity classes for padding: `pf-u-pl-lg` `pf-u-pr-lg`, as seen on https://www.patternfly.org/v4/utilities/spacing/

It might need to be handled differently for responsive modes too. But there's (the same) + `-on-lg` for that too, etc.

KKoukiou:
* @garrett In the current master the button on the right is far from aligned as well. It did not break here. Here is a screenshot from master.
Screenshot from 2021-04-09 12-03-24

I will try to fix the right side alignment issue however.

* The icons are indeed some pixes of vertically, this is broken everywhere on master since ever and I have created an issue upstream as well. https://github.com/patternfly/patternfly/issues/3959

* Regarding the left side it looks correct to me with the current PR, the center of the icons should be used to alignment base . https://www.patternfly.org/v4/guidelines/icons/
Previously we had the left side aligned from the left of the icons, which is wrong.

garrett:
> In the current master the button on the right is far from aligned as well. It did not break here. Here is a screenshot from master.

Right, it's a regression. It wasn't like this before some point in time when PF was updated. Just bringing it up as it impacts this PR somewhat, as the overrides had assumptions that are no longer true and this is affecting the overrides.

> I will try to fix the right side alignment issue however.

Thanks!

> The icons are indeed some pixels of vertically, this is broken everywhere on master since ever and I have created an issue upstream as well.

Ah, this issue is actually for the buttons, not the icons. But button vertical alignment in cards is a problem too. Thanks for opening that issue!

For reference, alignment issues:

- SVG: https://github.com/patternfly/patternfly/issues/3628
- pficon: https://github.com/patternfly/patternfly/issues/3536

garrett (approved):
(Apologies for taking a while to respond on this one. I had it in an open tab and lost it between all the other PRs. Whoops.)

Looks good at normal (and even mobile) sizes. Thanks!

Screenshot_2021-04-14 Overview - garrett Rain

However, when it's quite wide (at >= 1440px viewport, but this includes the shell), the margin at the left isn't right. (The right side is right though.)

Screenshot_2021-04-14 Overview - garrett Rain(1)

Anyway, it's an improvement overall. If you can easily fix the left side in larger sizes, please include that in this PR too. But it might be tricky, with icon padding or something else mysteriuous. If so, a follow-up is OK.

Marking as "approve", as my guess is that it's trickier (as the right side is doing the right thing and the left includes an icon and such).

KKoukiou:
@garrett I have no clear idea for fixing about the left spacing but I can take a look when I have time as a followup.

garrett:
Yeah, it wasn't clear to me either, but I didn't look at it closely. I'm happy to work together to figure it out. At least it's not a big deal and doesn't show up under many circumstances.

PR #15643, opened 2021-04-02
first round of LGTM fixes

allisonkarlitskaya:
Let's play around a bit with LGTM over the weekend...

https://lgtm.com/projects/g/cockpit-project/cockpit/

### DIFF
--- a/pkg/base1/cockpit.js
+++ b/pkg/base1/cockpit.js
@@ -3630,8 +3630,8 @@ function factory() {
          * close function to only work when a problem is
          * indicated.
          */
-        var old_close;
         if (shared) {
+            const old_close = client.close;
             client.close = function() {
                 if (arguments.length > 0)
                     old_close.apply(client, arguments);
--- a/tools/test-static-code
+++ b/tools/test-static-code
@@ -119,10 +119,10 @@ fi
 if [ -z "${PYTHON_STYLE_CHECKER-}" ]; then
     echo "ok 7 pycodestyle test # SKIP pycodestyle not installed"
 else
-    # all normal files in test/ and tools/ which are either:
+    # all normal files which are either:
     #   - filename is *.py
     #   - executable matching ^#!/usr/bin/python3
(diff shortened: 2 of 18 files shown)


tools/test-static-code
```diff
@@ -122,7 +122,7 @@ else
     # all normal files in test/ and tools/ which are either:
     #   - filename is *.py
     #   - executable matching ^#!/usr/bin/python3
-    PYFILES="$(find test/ tools/ -type f '(' -name '*.py' -o -executable -execdir grep -q '^#!/usr/bin/python3' '{}' ';' ')'  -print)"
+    PYFILES="$(find . -type f '(' -name '*.py' -o -executable -execdir grep -q '^#!/usr/bin/python3' '{}' ';' ')'  -print)"
```
martinpitt, line 125: This currently lives in the "various: pycodestyle changes" commit, but is meant to be in the "test-static-code: check all python files" commit. Please either move it, or squash the two commits together.

pkg/base1/cockpit.js
```diff
@@ -3630,8 +3630,8 @@ function factory() {
          * close function to only work when a problem is
          * indicated.
          */
-        var old_close;
         if (shared) {
+            var old_close = client.close;
```
martinpitt, line 3634: Please use `const` here -- scoping of `var` is really broken in JS, it does not do what you think it does; it always applies to the scope of the containing function, not just the `{}` block

martinpitt (changes requested):
Mostly LGTM :grin: Just two trivial things. (and of course your "LGMT" typos in the commit messages) Thanks!

pkg/base1/cockpit.js
allisonkarlitskaya: done, thanks.

I explicitly avoided changing this for exactly this reason.  I knew it was different, and I didn't know which one I wanted :rofl:

tools/test-static-code
allisonkarlitskaya: done, thanks

martinpitt (approved):
Thank you! As you already enabled LGTM and this is now red on all other PRs, landing this quickly.

PR #15650, opened 2021-04-04
more pycodestyle fixes

allisonkarlitskaya:
The first one of these is pretty important: the test-static-code script, when run in a directory with nodes_modules/ checked out will try to scan all the python it finds in there, and reject an awful lot of it, which (if you have the git hook setup) will pretty much prevent you from ever pushing anything.

### DIFF
--- a/tools/test-static-code
+++ b/tools/test-static-code
@@ -118,11 +118,13 @@ fi
 
 if [ -z "${PYTHON_STYLE_CHECKER-}" ]; then
     echo "ok 7 pycodestyle test # SKIP pycodestyle not installed"
+elif [ ! -e .git ]; then
+    echo "ok 7 pycodestyle test # SKIP not a git tree"
 else
     # all normal files which are either:
     #   - filename is *.py
     #   - executable matching ^#!/usr/bin/python3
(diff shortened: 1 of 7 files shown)


tools/test-static-code
```diff
@@ -122,7 +122,7 @@ else
     # all normal files which are either:
     #   - filename is *.py
     #   - executable matching ^#!/usr/bin/python3
-    PYFILES="$(find . -type f '(' -name '*.py' -o -executable -execdir grep -q '^#!/usr/bin/python3' '{}' ';' ')'  -print)"
+    PYFILES="$(git grep -lI '^#!.*python3') $(git ls-files "*.py")"
```
martinpitt, line 126: This ought to be guarded by `if [ -e .git ]` like in "valid JSON files"`, as we run test-static-code during `make check`. It is entirely plausible to build a released cockpit in a minimal build environment without `git` being installed.

martinpitt (changes requested):
Nice, thank you!

tools/test-static-code
allisonkarlitskaya: Good idea.  Thanks!

martinpitt (approved):
Cheers!

marusak (approved):
Than you!

PR #15652, opened 2021-04-05
test: Use unittest assert methods

marusak:
These provide better log when they fail.
For example `assert (size > 300000)` provides:
```
    assert (size > 300000)
AssertionError
```

while `assertGreater(size, 300000) provides:
```
    assert (size > 300000)
AssertionError: 1 is not greater than 300000
```

I only kept `assert False, <reason>`.

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -14,7 +14,7 @@ from testlib import *
 from machine_core.constants import TEST_OS_DEFAULT
 
 
-def getMaximumSpike(browser, g_type, saturation, hour, minute):
+def getMaximumSpike(test, g_type, saturation, hour, minute):
     # only for minutes with events, which have SVG graphs
     sel = "#metrics-hour-{0} div.metrics-minute[data-minute={2}] div.metrics-data-{1} polygon".format(hour, g_type, minute)
     if saturation:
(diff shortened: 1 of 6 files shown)


test/verify/check-metrics
```diff
@@ -23,7 +23,7 @@ def getMaximumSpike(browser, g_type, saturation, hour, minute):
         sel += ":first-child"
 
     points = browser.attr(sel, "points")
-    assert "NaN" not in points, points
+    self.assertNotIn("NaN", points)
```
marusak, line 26: self is not defined

test/verify/check-metrics
```diff
@@ -14,43 +14,43 @@ from testlib import *
 from machine_core.constants import TEST_OS_DEFAULT
 
 
-def getMaximumSpike(browser, g_type, saturation, hour, minute):
+def getMaximumSpike(self, g_type, saturation, hour, minute):
```
martinpitt, line 17: This is weird for a non-method. Better call this `test` or so.

martinpitt (dismissed):
Thanks! Especially looking forward to getting better debugging for storage-resize.

test/verify/check-metrics
marusak: You meant like this? (pushed)

martinpitt: Right, thanks!

martinpitt (approved):

PR #15654, opened 2021-04-06
test: Ignore errors when removing self.tmpdir during teardown

mvollmer:
The directory sometimes seems to disappear asynchronously, leading to
occasional errors like:

    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp4n4loeh6'

This shouldn't happen, but life is too short to fix all the bugs in it.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -973,7 +973,7 @@ class MachineCase(unittest.TestCase):
         if self.checkSuccess() and self.machine.ssh_reachable:
             self.check_journal_messages()
             self.check_browser_errors()
-        shutil.rmtree(self.tmpdir)
+        shutil.rmtree(self.tmpdir, ignore_errors=True)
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
I've seen this occasionally. So far I attributed it to docker messing up on our RHEL-7 e2e machines, they often seem to get entire swaths of file system being pulled away underneath the tests. But this little hack is harmless enough, so grudgingly accepting :grin: Danke!

mvollmer:
Also, `self.tmpdir` isn't really used, or is it?  (I found one use in `check-packagekit`)

martinpitt:
@mvollmer : Indeed -- I guess we had more instances of that in the past, but got rid of them. I suppose we could also just eliminate it completely and adjust check-packagekit to just download to `./tmp/`. Most tests use `vm_tmpdir` inside the testbed, it's rare that tests need a tmpdir on the host.

PR #15655, opened 2021-04-06
Remove cockpit-machines bits

KKoukiou:
The package was moved to its own repository at
https://github.com/cockpit-project/cockpit-machines/.


package.json
```diff
@@ -25,8 +25,6 @@
     "qunit-tap": "1.5.1",
     "react": "16.14.0",
     "react-dom": "16.14.0",
-    "redux": "4.0.5",
```
martinpitt, line 28: I think we can drop react-console as well.

tools/cockpit.spec
```diff
@@ -282,8 +282,6 @@ done
 rm -r %{buildroot}/%{_prefix}/%{__lib}/cockpit-test-assets
 # files from -pcp
 rm -r %{buildroot}/%{_libexecdir}/cockpit-pcp %{buildroot}/%{_localstatedir}/lib/pcp/
-# files from -machines
```
martinpitt, line 285: There are more instances of "machines" in the file, such as in the comment in line 66, and building machines.list (which is the command that causes the rpm build failure).

martinpitt (changes requested):
Yay!

tools/cockpit.spec
KKoukiou: Thanks for the hint!

martinpitt:
Remaining vestiges:
 - tools/debian/source/lintian-overrides
 - .github/ISSUE_TEMPLATE/bug_report.md and .github/ISSUE_TEMPLATE/enhancement.md (use a different example page name)
 - pkg/lib/page.scss (comment)
 - test/run-verify-host-user.sh (cirros image download and some TestMachines excludes)

None of these are critical though, so we can clean these up in a follow-up also.

KKoukiou:
> Remaining vestiges:
> 
>     * tools/debian/source/lintian-overrides
> 
>     * .github/ISSUE_TEMPLATE/bug_report.md and .github/ISSUE_TEMPLATE/enhancement.md (use a different example page name)
> 
>     * pkg/lib/page.scss (comment)

Comments in lib should remain since we use it also in external projects. 


> 
>     * test/run-verify-host-user.sh (cirros image download and some TestMachines excludes)
> 
> 
> None of these are critical though, so we can clean these up in a follow-up also.


Fixed all apart from page.scss

test/run-verify-host-user.sh
```diff
@@ -45,12 +45,7 @@ TESTS=""
 EXCLUDES=""
 RC=0
 if [ -n "$test_optional" ]; then
-    # pre-download cirros image for Machines tests
-    bots/image-download cirros
-
     # triggers SELinux violation
```
martinpitt, line 48: that comment line should then go as well.

martinpitt (approved):
Cheers! As above, fine to clean up the page.scss and fmf test comments in a follow-up, to not block other PRs.

martinpitt:
Some weird [TestSOS failures on centos](https://logs.cockpit-project.org/logs/pull-15655-20210406-092210-eabe2f4a-centos-8-stream/log.html#143-2), let's retry and compare.

martinpitt:
`TestTeam.testBasic` failed in [both](http://artifacts.dev.testing-farm.io/d655c217-08b1-48de-bca6-0dae693d6d3b/) [packit tests](http://artifacts.dev.testing-farm.io/0d9db52d-14b2-4f4a-9496-8e88d4a77a5d/) (plus two more on F34), which is a bit suspicious.  The [screenshot](http://artifacts.dev.testing-farm.io/d655c217-08b1-48de-bca6-0dae693d6d3b/work-allwtBvMP/plans/all/discover/default/tests/TestTeam-testBasic-fedora-33-localhost-22-FAIL.png) shows "connection type 'veth' is not valid", which suggests that dropping cockpit-machines now does not pull in some dependency any more which the test relies on.

Running locally to investigate.

martinpitt:
I squashed the fixups, added a separate commit to fix `testNfsMissingPackages`, and just added back the libvirt test dependencies for TestFirewall and TestTeam. I'm not happy about that, as libvirtd is *really* heavy, but this is good enough for now -- after all, the tests have installed these all the time through c-machines dependencies.

martinpitt:
TestSOS [failed again](https://logs.cockpit-project.org/logs/pull-15655-20210406-131758-c252853d-centos-8-stream/log.html#148-2). I locally ran TestSOS on centos-8-stream, and can't reproduce this. Unfortunately the log is absolutely useless; there is some machine restarting involved, so supposedly an earlier test breaks it somehow.

There is only a handful of tests which ran in that ND machine before:
```
ok 34 test/verify/check-example TestNondestructiveExample.testOne [ND@1]
ok 35 test/verify/check-example TestNondestructiveExample.testTwo [ND@1]
ok 248 test/verify/check-system-realms TestPackageInstall.testDialogTransition [ND@1]
ok 105 test/verify/check-packages TestPackages.testBasic [ND@1]
ok 106 test/verify/check-pages TestPages.testBasic [ND@1]
ok 107 test/verify/check-pages TestPages.testFrameReload [ND@1]
ok 108 test/verify/check-pages TestPages.testMenuSearch [ND@1]
ok 109 test/verify/check-pages TestPages.testPageStatus [ND@1]
ok 110 test/verify/check-pages TestPages.testPlots [ND@1]
ok 111 test/verify/check-pages TestPages.testPtBRLocale [ND@1]
ok 112 test/verify/check-pages TestPages.testReactPatterns [ND@1]
ok 113 test/verify/check-pages TestPages.testShellPreload [ND@1]
ok 114 test/verify/check-pages TestPages.testShellReload [ND@1]
ok 147 test/verify/check-sosreport TestSOS.testAppStream [ND@1]
```

I ran all of them and then TestSOS in the same machine, and it still works.

martinpitt:
I ran this cockpit-7 with this branch, to see what on earth happens to that poor VM:

    git clone -b delete-machines https://github.com/KKoukiou/cockpit
    cd cockpit
    tools/make-bots
    TEST_OS=centos-8-stream test/image-prepare -q
    bots/vm-run centos-8-stream
    TEST_OS=centos-8-stream test/verify/check-sosreport -tv --machine 127.0.0.2:2201 --browser 127.0.0.2:9091 TestSOS.testBasic

This works fine, just as it does locally. Then I ran

    TEST_OS=centos-8-stream test/common/run-tests --test-dir test/verify -n --jobs 1 TestSOS

here I noticed that packagekitd+dnf go fairly crazy, using 50% CPU each all the time. dnf is from sosreport, but maybe packagekit gets triggered from the previous `TestSOS.testAppStream`? I now see the eventual `Restarting global machine 0`, I suppose it was running out of memory or something such.

There seems to be something subtly different between `vm-run` + `--machine` and letting run-tests start the VMs by itself. The same command now reproduces the long hang locally. I did not have the patience to wait for the machine restart, I aborted the running test after 8 minutes.

I also noticed that sos generally collects a *lot* of data -- we have some outdated thingy in the test:
```
        if m.image in ["centos-8-stream", "debian-stable"]:  # Still has sos in version 3.x
```
which is not true any more -- centos-8-stream now has sos-4.0-10.el8.noarch just like RHEL 8. When I fix that, it seems to behave normally again.

I pushed that fix, let's see CI's opinion.

martinpitt:
LGTM now, but as I changed this, I'd like @marusak to have the final word.

marusak (approved):
Nice, thank you!
If I would have to guess I would say that we will definitely keep some things behind that are not necessary anymore, but lets do such cleanups when we discover them.

martinpitt:
Packit f33 failed with a TestBonding flake, but I'm ok with ignoring this for now.

PR #15656, opened 2021-04-06
pycodestyle: E402 fixups (import at top of file)

allisonkarlitskaya:
We have two classes of of cases where our imports are not at the top of
the file:

 - various tricks to manipulate the path before importing other modules

 - gi.require_version()

For gi.require_version() we could probably just drop it, but let's leave
it there and add a #NOQA comment.

For the path manipulation, let's up our game a bit.  We avoid this issue
in test/verify/ by symlinking path.py from test/common/.  Do the same
thing for test/container/ and tools/.

Fix a bug in path.py: it uses abspath() instead of realpath() when
determining the value of TEST_DIR, which means that it's not resolving
the symlink.  That's worked up to this point only because path.py has
only ever been linked to from the same depth as its original location in
test/common/.  Linking it from tools/ breaks that assumption.

Make some minor adjustment for not having access to the variable defined
as a side effect of the path manipulations (but we can still get them
from path.py).

Remove also a spurious dont_write_bytecode = True from tools/urls-check.
We're not doing this at all consistently from other places that import
parent.py.

Adjust tools/test-static-code to remove the E402 exception.

### DIFF
--- a/test/containers/run-tests
+++ b/test/containers/run-tests
@@ -22,17 +22,11 @@ import imp
 import os
 import string
 import unittest
+from parent import TEST_DIR, BASE_DIR
+import testlib
+
 
 sys.dont_write_bytecode = True
-
-base_dir = os.path.dirname(os.path.realpath(__file__))
-testdir = os.path.dirname(base_dir)
-sys.path.append(testdir)
-
-from verify import parent
-parent  # pyflakes
-
-from common import testlib
 
 
 def check_valid(filename):
@@ -47,7 +41,7 @@ def run(opts):
     # Now actually load the tests, any modules that start with "check-*"
     loader = unittest.TestLoader()
     suite = unittest.TestSuite()
-    for filename in glob.glob(os.path.join(base_dir,
+    for filename in glob.glob(os.path.join(TEST_DIR, "containers",
                                            "check-{0}*".format(opts.container))):
         name = check_valid(filename)
         if not name or not os.path.isfile(filename):
@@ -68,7 +62,7 @@ def main():
     if not opts.container:
         opts.container = 'bastion'
 
-    if not os.path.isdir(os.path.abspath(os.path.join(testdir, "..", "containers", opts.container))):
+    if not os.path.isdir(os.path.abspath(os.path.join(BASE_DIR, "containers", opts.container))):
         sys.stderr.write("Unable to run tests for unknown container {0}.\n".format(opts.container))
         exit(1)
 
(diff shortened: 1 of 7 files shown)


martinpitt:
Looks good, thanks! But let's queue this after #15655, as that's the bulky one that will collide with anything else that touches pkg/machines/.

I'm fairly sure that we can't drop `gi.require_version()` -- a few years ago it was made mandatory by pygobject, and it would not work at all any more without it. You are welcome to try, of course.

allisonkarlitskaya:
Fixed up a couple of minor issues spotted by LGTM (didn't remove enough unused `import` after the changes).

allisonkarlitskaya:
Significantly reworked parent.py and rebased.  It'll be easier to drop the machines stuff after the other PR lands.

tools/urls-check
```diff
@@ -18,20 +18,15 @@
...
 import argparse
 import subprocess
 import urllib
 from urllib.request import urlopen, Request
-
-sys.path.append(os.path.join(dirname(dirname(abspath(__file__))), "bots"))
+from parent import BASE_DIR
 import task
 
-sys.dont_write_bytecode = True
```
martinpitt, line 33: This is fine, but should be in a separate commit.

test/containers/run-tests
```diff
@@ -22,17 +22,11 @@ import imp
...
-
-base_dir = os.path.dirname(os.path.realpath(__file__))
-testdir = os.path.dirname(base_dir)
-sys.path.append(testdir)
 
-from verify import parent
-parent  # pyflakes
-
-from common import testlib
+sys.dont_write_bytecode = True
```
martinpitt, line 29: Let's drop that as well for consistency, or keep it in urls-check as well? (But of course better if we can just drop it)

martinpitt (changes requested):
Nice cleanup! Let's do the next round after #15655, with the rebase and dropped commit. After that it should be good to land.

test/containers/run-tests
allisonkarlitskaya: I left this one (along with `test/common/run-tests`) because they import the individual testcases, which puts them in a bit of a different category.

tools/urls-check
allisonkarlitskaya: OK.

test/containers/run-tests
martinpitt: Ack.

martinpitt (approved):
Danke!

PR #15657, opened 2021-04-06
test: Remount encrypted file system

marusak:
Upon growing volume that contains encrypted file system this fs is
unmounted. Then `df` does not show from it what we expect.

I am not sure if this is expected behavior or if it is issue in cockpit or udisks.

See https://github.com/cockpit-project/bots/pull/1877

### DIFF
--- a/test/verify/check-storage-resize
+++ b/test/verify/check-storage-resize
@@ -88,6 +88,10 @@ class TestStorageResize(StorageCase):
                 self.content_head_action(fsys_row, "Mount")
                 self.confirm()
                 self.wait_mounted(fsys_row, fsys_tab)
+            if crypto:  # Encrypted FS is not remounted upon resizing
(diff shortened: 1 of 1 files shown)


martinpitt:
@marusak : The "unexpected unmounting" is https://github.com/cockpit-project/bots/issues/1739 -- so maybe that's just another variant how this can fail?

martinpitt:
And indeed, the test failures on debian-stable etc. corroborate that -- the fs is mounted there already.

marusak:
Does seem that it is the case only on newer systems. F34 and rhel-9 do have this behaviour, F33, debian, rhel-85 don't.

marusak:
Aha, thanks pitti! Good catch.

marusak:
It is indeed different pattern. I'll send bots PR and explain it there.

PR #15661, opened 2021-04-07
deduplicate a lot of socketpair() wrappers

allisonkarlitskaya:


martinpitt (approved):
Thanks! (Still fun that it is a net gain in lines of code, but there's headers and such..)

PR #15664, opened 2021-04-08
Increase use of GSocket in CockpitPipe

allisonkarlitskaya:
This lets us avoid some systems-level programming, a `GSocketAddress` translation shim, and adds a nice new `"socket"` property on `CockpitPipe`, which we use to help tidy up some code in `CockpitPeer`.

I'm not yet sure if it's worth it...

Note: this is a net increase in lines of code, but a decrease in complexity, I'd say.  Most of the increase is GObject property faff.

Random note: this removes the last use of raw `socketpair()` outside of the tests.

### DIFF
--- a/src/bridge/cockpitpeer.c
+++ b/src/bridge/cockpitpeer.c
@@ -23,13 +23,15 @@
 #include "cockpitrouter.h"
 
 #include "common/cockpitauthorize.h"
+#include "common/cockpithex.h"
 #include "common/cockpitjson.h"
 #include "common/cockpitmemory.h"
-#include "common/cockpittransport.h"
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
I did this in preparation for the peer-stderr-forwarding trickery and ended up not using it at all, for a couple of reasons:
 - `GSocket` setting the fd non-blocking was breaking expectations of existing code
 - using `GSocket` from early in `main()` in the peer bridge was broken since it broke the "don't enter GLib" rule

All in all, this refactor, on its own, isn't worth it.  Maybe some day we'll need this functionality, but until that time, let's shelve this.

PR #15665, opened 2021-04-08
containers: Pull fedora image from registry.fedoraproject.org

marusak:
Otherwise it pulls from docker.io where we can hit pull rate limit.


martinpitt (approved):
Dakujem!

PR #15666, opened 2021-04-08
proof of concept: pass stderr via socket to the privilged helper

allisonkarlitskaya:
So far this is pretty useless, but it will allow us to use the stderr fd to reliably communicate with sudo, without side-effects.

### DIFF
--- a/src/bridge/bridge.c
+++ b/src/bridge/bridge.c
@@ -39,6 +39,7 @@
 
 #include "common/cockpitassets.h"
 #include "common/cockpitchannel.h"
+#include "common/cockpitfdpassing.h"
 #include "common/cockpithacks-glib.h"
 #include "common/cockpitjson.h"
 #include "common/cockpitpipetransport.h"
@@ -682,12 +683,45 @@ main (int argc,
   signal (SIGABRT, cockpit_test_signal_backtrace);
   signal (SIGSEGV, cockpit_test_signal_backtrace);
 
-  /* In case we are run via sshd and we have journald, make sure all
-   * logging output ends up in the journal on *this* machine, not sent
-   * back to the client.
-   */
-  if (g_getenv ("SSH_CONNECTION") && !g_log_writer_is_journald (2) && ! isatty(2))
+  if (g_strv_contains ((const gchar **) argv, "--privileged"))
     {
+      /* We are being spawned, under sudo or pkexec, by the user's copy
+       * of the bridge.  In that case, the first thing that will happen,
+       * if we receive our stderr via the socket that is our stdin.
+       */
+      const char *msg = "\n{\"command\": \"send-stderr\"}";
+      g_print ("%zu\n%s", strlen (msg), msg);
+
+      int parent_stderr_fd;
+      int r = cockpit_socket_receive_fd (STDIN_FILENO, &parent_stderr_fd);
+      if (r == 0)
+        {
+          /* on EOF, just silently exit */
+          return 0;
+        }
+      else if (r == -1)
+        {
+          g_printerr ("cockpit-bridge: recvmsg(stdin) failed: %m\n");
+          return 1;
+        }
+      else if (parent_stderr_fd == -1)
+        {
+          g_printerr ("cockpit-bridge: message from stdin contains no fd\n");
+          return 1;
+        }
+      else
+        {
+          r = dup2 (parent_stderr_fd, STDERR_FILENO);
+          g_assert (r == STDERR_FILENO); /* that should really never fail */
+          close (parent_stderr_fd);
+        }
+    }
+  else if (g_getenv ("SSH_CONNECTION") && !g_log_writer_is_journald (2) && ! isatty(2))
+    {
+      /* In case we are run via sshd and we have journald, make sure all
+       * logging output ends up in the journal on *this* machine, not sent
+       * back to the client.
+       */
       int fd = sd_journal_stream_fd ("cockpit/ssh", LOG_WARNING, 0);
 
       /* If it didn't work, then there's no journal.  That's OK: we'll
--- a/src/bridge/cockpitpeer.c
+++ b/src/bridge/cockpitpeer.c
@@ -23,12 +23,13 @@
 #include "cockpitrouter.h"
 
 #include "common/cockpitauthorize.h"
+#include "common/cockpitfdpassing.h"
+#include "common/cockpithex.h"
 #include "common/cockpitjson.h"
 #include "common/cockpitmemory.h"
-#include "common/cockpittransport.h"
 #include "common/cockpitpipe.h"
 #include "common/cockpitpipetransport.h"
-#include "common/cockpithex.h"
+#include "common/cockpittransport.h"
 
 #include <sys/prctl.h>
 #include <sys/socket.h>
@@ -400,7 +401,22 @@ on_other_control (CockpitTransport *transport,
           cockpit_transport_send (transport, NULL, reply);
           g_bytes_unref (reply);
         }
+    }
 
+  /* cockpit-bridge --privileged expects to receive a copy of our stderr */
+  else if (g_str_equal (command, "send-stderr") &&
+           cockpit_json_get_bool (self->config, "privileged", FALSE, &privileged) && privileged)
+    {
+      CockpitPipe *pipe = cockpit_pipe_transport_get_pipe (COCKPIT_PIPE_TRANSPORT (transport));
+
+      int out_fd;
+      g_object_get (pipe, "out-fd", &out_fd, NULL);
+
+      if (!cockpit_socket_send_fd (out_fd, STDERR_FILENO))
+        {
+          g_critical ("sendmsg() with stderr fd failed: %m");
+          cockpit_transport_close (transport, "internal-error");
+        }
     }
 
   /* Otherwise we need an init message first */
--- a/src/tls/connection.c
+++ b/src/tls/connection.c
@@ -46,6 +46,7 @@
 #include <gnutls/gnutls.h>
 #include <gnutls/x509.h>
 
+#include <common/cockpitfdpassing.h>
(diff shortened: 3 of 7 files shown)


mvollmer:
I like this! The `send-stderr` command is a very cool approach to this, imo.

src/bridge/bridge.c
```diff
@@ -682,12 +685,33 @@ main (int argc,
   signal (SIGABRT, cockpit_test_signal_backtrace);
   signal (SIGSEGV, cockpit_test_signal_backtrace);
 
-  /* In case we are run via sshd and we have journald, make sure all
-   * logging output ends up in the journal on *this* machine, not sent
-   * back to the client.
-   */
-  if (g_getenv ("SSH_CONNECTION") && !g_log_writer_is_journald (2) && ! isatty(2))
+  if (g_strv_contains ((const gchar **) argv, "--privileged"))
```
mvollmer, line 688: Thinking out loud: why would we do this only for privileged peers? Why not do it always? One answer: because not all things that start cockpit-bridge understand the send-stderr command.

src/bridge/bridge.c
```diff
@@ -682,12 +685,33 @@ main (int argc,
...
+      g_autoptr(GUnixConnection) connection = cockpit_socket_connection_new_dup_fd (G_TYPE_UNIX_CONNECTION,
+                                                                                    STDIN_FILENO, &error);
+      if (!connection)
+        errx (EXIT_FAILURE, "couldn't create socket from stdin: %s", error->message);
+
+      int fd = g_unix_connection_receive_fd (connection, NULL, &error);
+      if (fd == -1)
+        errx (EXIT_FAILURE, "couldn't receive stderr socket from stdin: %s", error->message);
+
+      int r = dup2 (fd, 2);
```
mvollmer, line 706: The commit message says "The first thing we do (before calling into any part of GLib) is to
request and receive the stderr file descriptor of the bridge which is
spawning us." but there is a lot of glib stuff going on here before calling `dup2`, no?

src/bridge/bridge.c
allisonkarlitskaya: So I think about the potential implications of a weird remote host telling us to send it stderr over ssh, for example.  It just doesn't make sense, so I fence it in for the time being to what we actually need it for.

If we want to expand it later, that will be easy.

src/bridge/bridge.c
allisonkarlitskaya: That's a pretty good point...

allisonkarlitskaya: So basically, glib doesn't really consider that stderr might be replaced while the program is running.  It contains a static variable which is initialised at the first time that a message is printed (note: this does *not* exclude debug messages not shown) which records if stderr is the journal.  After that gets set, it's impossible to change it.

So yes, if we ran this with debugging output enabled, I guess we'd get in trouble.  But we'd also end up with a lot of debugging output in the sudo error message :)

allisonkarlitskaya:
I'm currently rewriting the fd-send/receive stuff to be "pure C".  Will probably push the next version tomorrow.

src/bridge/bridge.c
mvollmer: Makes sense.

src/bridge/bridge.c
allisonkarlitskaya: I addressed this by rewriting in pure C.

mvollmer:
LGTM! I'll put #15293 on top of this to verify that it all actually works out.

allisonkarlitskaya:
This is ready, but we shouldn't land it until we are sure the sudo stuff can use this and will definitely go in.

allisonkarlitskaya:
Been trying to make sense of the debian/ubuntu `TestKerberos.testNegotiate` fail for a while.  The big sudo-related `# HACK` sicks out like a sore thumb, but I'm really not sure how it could be related, to be honest...  very confusing.

allisonkarlitskaya:
Okay.  I think it is indeed, unrelated.  The issue is probably as follows: if we shut down the main bridge before we have a chance to properly complete bringing up the root bridge, then the root bridge is going to get EOF when trying to receive the stderr fd, and throw the error message `cockpit-bridge: failed to receive stderr fd: Invalid argument`.

That `EINVAL` is from us: it just means that the syscall didn't fail but we have no fd inside the received data.

Previously, getting EOF before the bridge did anything meaningful was probably just met with the bridge exiting without warning.

I think we need to modify the stderr-receiving code to exit silently in case we get EOF while reading to read the fd.

allisonkarlitskaya:
Following changes in the last commit:
 - remove the dubious `receive fd into` thing and the synthetic `EINVAL` errno.  Let's just open-code that in `main()` from the bridge
 - add support for better error detection, and particularly reporting of EOF on receive
 - proper API docs

src/bridge/bridge.c
```diff
@@ -58,6 +59,7 @@
 #include <glib-unix.h>
 #include <glib/gstdio.h>
 #include <gio/gunixsocketaddress.h>
+#include <gio/gunixconnection.h>
```
allisonkarlitskaya, line 62: Need to drop that.

src/bridge/bridge.c
```diff
@@ -682,12 +684,45 @@ main (int argc,
   signal (SIGABRT, cockpit_test_signal_backtrace);
   signal (SIGSEGV, cockpit_test_signal_backtrace);
 
-  /* In case we are run via sshd and we have journald, make sure all
-   * logging output ends up in the journal on *this* machine, not sent
-   * back to the client.
-   */
-  if (g_getenv ("SSH_CONNECTION") && !g_log_writer_is_journald (2) && ! isatty(2))
+  if (g_strv_contains ((const gchar **) argv, "--privileged"))
```
allisonkarlitskaya, line 687: Strictly speaking, we shouldn't use this corner of GLib either.....

src/bridge/cockpitpeer.c
```diff
@@ -400,7 +401,19 @@ on_other_control (CockpitTransport *transport,
...
+  else if (g_str_equal (command, "send-stderr") &&
+           cockpit_json_get_bool (self->config, "privileged", FALSE, &privileged) && privileged)
+    {
+      CockpitPipe *pipe = cockpit_pipe_transport_get_pipe (COCKPIT_PIPE_TRANSPORT (transport));
+
+      int out_fd;
+      g_object_get (pipe, "out-fd", &out_fd, NULL);
 
+      bool success = cockpit_socket_send_fd (out_fd, STDERR_FILENO);
+      g_assert (success);
```
allisonkarlitskaya, line 416: Maybe this should be less brutal.  I can't imagine why it would ever fail, but if it does, probably g_critical() would be nicer.  In that case, we should also definitely close the connection as well.

src/tls/connection.c
```diff
@@ -267,15 +230,22 @@ buffer_write_to_fd (Buffer *self,
...
       do
         s = sendmsg (fd, &msg, MSG_NOSIGNAL | MSG_DONTWAIT);
       while (s == -1 && errno == EINTR);
 
       debug (BUFFER, "  sendmsg returns %zi %s", s, (s == -1) ? strerror (errno) : "");
 
-      fd_send_complete (fd_to_send, s);
+      if (fd_to_send && *fd_to_send != -1 && s != -1)
+        {
+          close (*fd_to_send);
```
allisonkarlitskaya, line 246: ... too bad there's not a nicer way to write that.

src/bridge/bridge.c
allisonkarlitskaya: We already call other random little "harmless" bits of GLib for the ssh case, though.  This is probably fine, and better than the alternative.

mvollmer:
> This is ready, but we shouldn't land it until we are sure the sudo stuff can use this and will definitely go in.

I have tested it manually, and it works beautifully!

src/bridge/bridge.c
```diff
@@ -682,12 +683,45 @@ main (int argc,
...
     {
+      /* We are being spawned, under sudo or pkexec, by the user's copy
+       * of the bridge.  In that case, the first thing that will happen,
+       * if we receive our stderr via the socket that is our stdin.
+       */
+      const char *msg = "\n{\"command\": \"send-stderr\"}";
```
mvollmer, line 692: Intuitively, I would have expected the "\n" to be at the end here...

src/bridge/cockpitpeer.c
```diff
@@ -400,7 +401,22 @@ on_other_control (CockpitTransport *transport,
           cockpit_transport_send (transport, NULL, reply);
           g_bytes_unref (reply);
         }
+    }
 
+  /* cockpit-bridge --privileged expects to receive a copy of our stderr */
+  else if (g_str_equal (command, "send-stderr") &&
+           cockpit_json_get_bool (self->config, "privileged", FALSE, &privileged) && privileged)
```
mvollmer, line 408: Why the extra check for `"privileged"` here?

src/bridge/bridge.c
allisonkarlitskaya: This is part of the cockpit protocol.  First comes the channel name, then a newline, then the payload.  In the case of control messages, the channel name is empty (ie: newline at the start).

src/bridge/cockpitpeer.c
allisonkarlitskaya: I think we covered this one before: it would be weird to get this message from something that's not a privileged bridge, and it might cause us to engage in weird behaviour (like trying to send an fd to ssh).  Better to treat that as unexpected than to attempt to fulfill the request.  We can expand it later if we want...

src/bridge/bridge.c
mvollmer: Ah, of course, thanks!

mvollmer (approved):

mvollmer:
Tests in #15293 are also happy. Shall we just merge this?

PR #15667, opened 2021-04-08
shell: use pf4 Menu component for the display language list

KKoukiou:
New:

Screen Shot 2021-04-09 at 13 42 09

Screen Shot 2021-04-09 at 13 48 31

Before:

Screen Shot 2021-04-09 at 12 27 55

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -52,7 +52,6 @@
 
         "jsx-a11y/anchor-is-valid": "off",
         "jsx-a11y/no-autofocus": "off",
-        "jsx-a11y/no-onchange": "off",
 
         "eqeqeq": "off",
         "import/no-webpack-loader-syntax": "off",
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -305,20 +305,6 @@ input[type=number]:not(.pf-c-form-control) {
 
 [hidden] { display: none !important; }
 
-/* btn-lg, btn-sm don't exist in PF4 */
-.pf-c-button.btn-lg {
-    padding: .5rem 1rem;
-    font-size: 1.25rem;
-    line-height: 1.5;
-    border-radius: .3rem;
-}
-.pf-c-button.btn-sm {
-    padding: .25rem .75rem;
-    font-size: .875rem;
-    line-height: 1.5;
-    border-radius: .2rem;
-}
-
 // Let PF4 handle the scrolling through page component otherwise we might get double scrollbar
 html:not(.index-page) body {
   overflow-y: hidden;
--- a/pkg/networkmanager/interfaces.js
+++ b/pkg/networkmanager/interfaces.js
@@ -1845,7 +1845,7 @@ PageNetworkIpSettings.prototype = {
                         $('<strong>').text(title),
                         $('<div class="pull-right">').append(
                             header_buttons,
-                            add_btn = $('<button class="pf-c-button pf-m-secondary btn-sm">')
+                            add_btn = $('<button class="pf-c-button pf-m-secondary pf-m-small">')
                                     .append('<span class="fa fa-plus">')
                                     .css("margin-left", "10px")
                                     .click(add()))),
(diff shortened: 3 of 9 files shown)


KKoukiou:
One note:  this blue background that exists in selectable lists in cockpit in many places is not something that patternfly offers. I actually like the blue selected lines, but I would stick to what's available from the pf lib.

marusak:
(distropkg)[https://logs.cockpit-project.org/logs/pull-15667-20210409-093453-df3b9417-rhel-8-5-distropkg/log.html#268] needs some attention

KKoukiou:
> (distropkg)[https://logs.cockpit-project.org/logs/pull-15667-20210409-093453-df3b9417-rhel-8-5-distropkg/log.html#268] needs some attention

Good catch.
if "distropkg" in m.image: is fine? 
I saw we have three distropkg images now.

marusak:
> if "distropkg" in m.image: is fine?

directly specifying 8.5 would be better I think. 8.4 is only in 8.4 branch, that we don't care about anymore. 9.0 is not yet enabled. I am not sure about this one exactly, since it should be "trickle down" kind of image.

marusak:
Other than that it all looks fine, thanks!

marusak (approved):
Thanks! (assuming green tests)

garrett:
Oh, I thought this was a bug.

The shadows look really quite odd and there's an issue of spacing between the language and the widgets.

KKoukiou:
> Oh, I thought this was a bug.
> 
> The shadows look really quite odd and there's an issue of spacing between the language and the widgets.

Hmm, the shadow is the default PF4 shadow for list items, and regarding the spacing I am not sure what you mean. Can you post maybe a screenshot with the expected spacing?

garrett:
Sure thing, I'll look to see what PF expects and see what we can / should do.

The functionality is great though!

garrett:
Oh, I guess this is a menu, but menus should only show up as a popup... and this shouldn't be a popup.

Here's the widget (or something that looks a lot like it):
https://www.patternfly.org/v4/components/menu#filtering-with-text-input

But this widget is really just meant as a special filterable menu:
https://www.patternfly.org/v4/components/menu/design-guidelines

There's a similar widget over at the tree view:
https://www.patternfly.org/v4/components/tree-view/#with-search

However, this isn't really a tree and that is missing the outline. So it's not right either. But same general idea with a filterbox and a scrollable, selectable list below.

If this were HTML, I'd suggest using an input above and a multselect `<select multiple="true">` below. 

I don't see a multiple select element in PF4 though? I looked at form select and you can add multiple="true", but the CSS doesn't allow it to get tall enough and the padding gets all weird and such, so that doesn't work.

Menu is definitely the wrong widget, however. It's only supposed to be used as a popup (and that's why there's a shadow around this, to make it look like it's a popup, as that's what it is indended to be).

I guess the tree view version is actually closer in intention than the menu, even if both aren't quite right. I suppose the tree view supports non-expandable leaf items in the top level? And then we'd just need to have a box around it and have overflow. But these should really be built-in features. (And they don't seem to be?)

Anyway, the end result should look like this:

Screenshot from 2021-04-26 16-04-55


The SCSS override I made for this (the menu widget, which is not the right widget... _but what is?_) looks something this:

```scss
.display-language-modal {
  .pf-c-menu {
    box-shadow: none;
    margin: var(--pf-global--spacer--sm) 0 0;

    .pf-c-divider {
      display: none;
    }

    &__content {
      border: 1px solid var(--pf-global--BorderColor--100);
      margin-top: var(--pf-global--spacer--sm);
    }

    &__search {
      padding: 0;
    }
  }
}

#display-language-list {
  /* Remove margin; move this CSS to .pf-c-menu override above */
}
```

We could apply the custom CSS locally for this use and try to make sure PF4 grows a proper multi-select widget in the meantime, complete with a searchable variation. (Unless I'm missing one, but I spent a while searching for something suitable.)

PR #15668, opened 2021-04-08
lib: Adjust cockpit-po-plugin for webpack v5

martinpitt:
The plugin works with webpack v5, but throws a deprecation warning:

> [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets
> will be frozen in future, all modifications are deprecated.
> BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
> Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
> Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.

When building with webpack v5, use the more fine-grained `processAssets`
hook, which is meant for adding extra assets. This API is not available
in v4 yet, so until we move all our projects to v5 we have to support
both cases.

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -4,6 +4,7 @@ const fs = require("fs");
 const childProcess = require("child_process");
 const po2json = require('po2json');
 const Jed = require('jed');
+const webpack = require('webpack');
 
 const srcdir = process.env.SRCDIR || path.resolve(__dirname, '..', '..');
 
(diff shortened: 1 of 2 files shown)


martinpitt:
I noticed that in https://github.com/cockpit-project/starter-kit/pull/445 . Not strictly a blocker, but let's sort that out before webpack actually stops supporting the old way.

martinpitt:
Some random test failures: [this one](https://logs.cockpit-project.org/logs/pull-15668-20210408-142703-164c3689-debian-testing/log.html#149) is issue #15544 and I don't understand it.

[this one](https://logs.cockpit-project.org/logs/pull-15668-20210408-141850-164c3689-rhel-9-0/log.html#159) is a simple test adjustment, https://bugzilla.redhat.com/show_bug.cgi?id=1836182 also affects rhel-9-0. I pushed a commit to fix that.

KKoukiou (approved):
Would not it make sense to actually change this when we move to webpack 5 in package.json? Is it something more blocking us from that?

I have no issue introducing this now, just curious what's the plan here.

martinpitt:
@KKoukiou : I wanted to start with porting starter-kit and podman etc. to webpack 5 first, as these are easier (and I can build up the necessary experience). pkg/lib is shared between all projects, so it needs to be compatible with both for a while. I do want to drop the v4 fallbacks once everything moved over, of course.

PR #15670, opened 2021-04-09
containers: Use correct $container type

marusak:
In commit e87ad98d4 base image was changed which meant that now `$container` is `oci` and not `podman`.


martinpitt:
Interesting! If that's it, then that suggests that the `label INSTALL` somehow doesn't work on the quay image?

marusak:
Aha!
https://github.com/cockpit-project/cockpit-container/blob/master/atomic-install#L47
here `container` is `oci` and not `podman`.

marusak:
Ok, actual fix and not just "I-have-no-idea-what-I-am-doing" hack

martinpitt (approved):
Huh, how obscure -- thanks for figuring that out!

PR #15671, opened 2021-04-09
lib: add patternfly override for misaligned card actions

### LINKED ISSUE #15473: ct-card action alignment does not work properly with links

KKoukiou:
Fixes #15473

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -65,20 +65,24 @@ svg {
     }
 }
 
-.pf-c-card__header {
-  // upstream fix (pending): https://github.com/patternfly/patternfly/pull/3714
-  flex-wrap: wrap;
-  row-gap: var(--pf-global--spacer--sm);
-  justify-content: space-between;
+.pf-c-card {
+  // https://github.com/patternfly/patternfly/issues/3959
+  --pf-c-card__header-toggle--MarginTop: 0;
+  .pf-c-card__header {
+    // upstream fix (pending): https://github.com/patternfly/patternfly/pull/3714
+    flex-wrap: wrap;
+    row-gap: var(--pf-global--spacer--sm);
+    justify-content: space-between;
 
-  > .pf-c-card__actions {
-      flex-wrap: wrap;
-      row-gap: var(--pf-global--spacer--sm);
+    > .pf-c-card__actions {
+        flex-wrap: wrap;
+        row-gap: var(--pf-global--spacer--sm);
 
-      // PF4 CardActions act up when using buttons while the title is large of font
-      // https://github.com/patternfly/patternfly/issues/3713
-      margin: unset;
-      padding-left: unset;
+        // PF4 CardActions act up when using buttons while the title is large of font
+        // https://github.com/patternfly/patternfly/issues/3713
+        margin: unset;
+        padding-left: unset;
+    }
   }
 }
 


marusak (dismissed):
In #15473 @garrett shown pictures where it was aligned to bottom, not center. I prefer it more like it is here, but lets hear what @garrett thinks as well.

Looks like this:
currentdffsdf

garrett:
What's aligned here? I don't see any alignment except horizontally centered for the "No log entries" text. If you're talking about the header and the action, they're simply not aligned.

Either the "all logs" action should be a (secondary) button and it would be center aligned or it needs to be link and baseline aligned. One or the other. It should not be aligned to the center if a link, and it should also not be aligned to the bottom under any circumstances.

Also, "All logs" is not an action. "View all logs" is.
https://www.patternfly.org/v4/components/button/design-guidelines#content

garrett:
(Yeah, this might need to be fixed outside of this PR. But it's not obvious how this PR interacts with the logs card if that's not doing things properly. Or we might need to have alignment be different for the links inside this PR. :shrug:)

KKoukiou:
@garrett I don't know how I can fix this the way you expect it to work, im fact I just did what the patternfly designer commented on the linked issue.

garrett:
If it's a button in a flex, it can be fixed with `align-items: center` on the flex container or `align-self: center` on the button itself. If it's a link, then it should be `baseline`, not `center`.

That's the CSS solution (above).

For JSX, there's something like this for the PF flex element:
```
    <Flex direction={{ default: 'column' }} alignSelf={{ default: 'alignSelfBaseline' }}>
```

And there are classes like this for all sorts of widgets:
https://www.patternfly.org/v4/utilities/flex/#aligned-items
https://www.patternfly.org/v4/utilities/flex/#aligned-self

But both the class and the JSX use the CSS solution when you get down to the implementation. So it doesn't matter. The idea is the same, regardless of how it's coded in Cockpit.

The best approach is probably with an `align-items: baseline` on the flexing element and then use `align-self: center` on the buttons, as this will work with normal text and also links and have the special-casing for the buttons. :shrug: If PF aready applies the baseline alignment for the flex, then that's one less thing to set. (Note: baseline alignment should be the default in HTML+CSS, IIRC.)

However, if PF applies a `center` vertical alignment for the items, then do the reverse and add an align-self to the links and link-styled-buttons.

We don't want to do this on a component-by-component level. What would most like be best: have 1 simple line of CSS as an override and submit an issue/patch upstream (and eventually remove the override).

KKoukiou:
@garrett the button and the header title, have both div wrappers, and these have the same parent. Therefore I am not sure the solutions above work. In anycase, I overestimated how much CSS I can write with this, if you want to take over feel free, else I will close it and you can handle it when you have time.

garrett:
I see what you mean. Yeah, after digging in the code, this isn't as obvious as it appears to be. What I said above applies in concept, but (like you said) the wrappers get in the way, in practice.

We could consider using the secondary style for the buttons instead of the link style. That's specific to the implementations however. And then the object should be vertically centered. (It wouldn't change position, but it wouldn't have arbitrary blank space around it — there would be the button's border framing the space.)

And for any we do want to consider keeping as links... well, we might have to compute an offset between the button height and text height. Unless we'd figure out how to get it to align despite being in wrappers, which would be better.

garrett:
Ah wait, this still fixes the alignment, and I think we might need it for the buttons. Re-opening. Whoops.

martinpitt:
Conflicty now, needs rebasing.

KKoukiou:
> Ah wait, this still fixes the alignment, and I think we might need it for the buttons. Re-opening. Whoops.

Right, this makes action alignment, when these are a button. perfect. The center alignment is wrong for links as @marusak mentions, but it's better that what we have now. I think a real solution is not straightforward, so I would suggest to get this partial fix in for start, unless we actually manage to make it work globally.

garrett:
BTW: This looked good to me for the buttons when reviewing it; I will happily approve it once it is rebased.

(Then we can figure out what to do about links and link-styled buttons at a later time, in some future follow-up.)

KKoukiou:
@garrett so shall we get this in or not?

garrett (approved):
> so shall we get this in or not?

Yep, we should.

PR #15672, opened 2021-04-10
Update ko.po

simmon-nplob:
update translation

### DIFF
--- a/po/ko.po
+++ b/po/ko.po
@@ -1,12 +1,3 @@
-# MinWoo Joh <igtzhsou@naver.com>, 2015. #zanata
-# MinWoo Joh <igtzhsou@naver.com>, 2016. #zanata
-# Eun-Ju Kim <eukim@redhat.com>, 2018. #zanata
-# Ludek Janda <ljanda@redhat.com>, 2018. #zanata, 2020.
-# cockpit <cockpituous@gmail.com>, 2018. #zanata
-# Eun-Ju Kim <eukim@redhat.com>, 2019. #zanata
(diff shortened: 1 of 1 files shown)


marusak:
Is this serious change? If so, then there is no reason to remove authors.
Anyway, if you are interested in translating Cockpit you can do so in https://translate.fedoraproject.org/projects/cockpit/master/

martinpitt:
@marusak : We've had several spam PRs with bogus changes recently. That account is very new, and the only other "contribution" was https://github.com/fedora-infra/asknot-ng/pull/138 with a similarly bogus change. My hunch is that this is some kind of "collecting fake credits" for $whatever purpose..

simmon-nplob:
I have translated cockpit, and I want to do an optimized translation. Since there was no previous translation, I want to correct it. 
It won't be big for that project, but it makes sense for me. Thanks you co-operation..by simmon

simmon-nplob:
fedora-infra/asknot-ng#138.... No! The web part needs to be applied, and the translation part needs to be applied. That part is not spam. This is the part I requested so that the web applied in the middle of the translation will be operated and the part will be applied. If the service is serviced while the translation has not been completed, it will be in a strange state.

marusak:
> I have translated cockpit

Right, I [see](https://translate.fedoraproject.org/user/simmon/#changes) that you are active translator, thanks for that!

> I want to do an optimized translation. Since there was no previous translation, I want to correct it.

What does optimized mean in this case? And how does removing mention of previous translators fixes anything?

PR #15674, opened 2021-04-11
test: Relax expected messages in TestConnection.testCockpitDesktop

martinpitt:
Sometimes it looks like this:

    cannot reauthorize identity(s): unix-user:admin

Spotted in https://github.com/cockpit-project/bots/pull/1912 where it caused [this random failure](https://logs.cockpit-project.org/logs/pull-1912-20210411-111351-7ebd59aa-ubuntu-stable-cockpit-project-cockpit/log.html#14)


martinpitt:
This also appears on the [weather report](https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html), it happens outside of that bots PR as well.

marusak (approved):

PR #15675, opened 2021-04-11
test: Drop usage of Machine.atomic_image

martinpitt:
This was dropped in
https://github.com/cockpit-project/bots/commit/ac9b124ee1a7

Drop cases which were really Atomic specific, and replace most others
with `ostree_image`.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -109,29 +109,28 @@ class TestConnection(MachineCase):
             self.assertTrue(cookie["httpOnly"])
         self.assertFalse(cookie["secure"])
 
-        if not m.atomic_image:  # cannot write to /usr on Atomics, and cockpit-session is in a container
-            # damage cockpit-session permissions (Fedora-ish and Debian-ish path), expect generic error message
-            m.execute("chmod g-x /usr/libexec/cockpit-session 2>/dev/null || chmod g-x /usr/lib/cockpit/cockpit-session")
-            b.open("/system")
-            b.wait_in_text('#login-fatal-message', "Internal error in login process")
-            m.execute("chmod g+x /usr/libexec/cockpit-session 2>/dev/null || chmod g+x /usr/lib/cockpit/cockpit-session")
+        # damage cockpit-session permissions (Fedora-ish and Debian-ish path), expect generic error message
+        m.execute("chmod g-x /usr/libexec/cockpit-session 2>/dev/null || chmod g-x /usr/lib/cockpit/cockpit-session")
+        b.open("/system")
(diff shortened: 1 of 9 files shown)


test/verify/check-kdump
```diff
@@ -32,12 +32,7 @@ class TestKdump(MachineCase):
 
     def enableKdump(self):
         # we need to make sure that the kernel command line options include our crashkernel parameter
-        if self.machine.atomic_image:
-            # on atomic we have a crashkernel option already, but for auto
-            contents = self.machine.execute(command="cat /boot/grub2/grub.cfg", quiet=True)
-            self.assertIn("crashkernel=auto", contents)
-            self.machine.write("/boot/grub2/grub.cfg", contents.replace("crashkernel=auto", "crashkernel=256M"))
-        elif self.machine.image in ["rhel-8-0", "rhel-8-0-distropkg", "rhel-8-1"]:
+        if self.machine.image in ["rhel-8-0", "rhel-8-0-distropkg", "rhel-8-1"]:
```
martinpitt, line 35: I could have cleaned this up also.. I'll do that if I forgot something else, otherwise :shrug:

test/verify/check-connection
```diff
@@ -109,7 +109,7 @@ class TestConnection(MachineCase):
             self.assertTrue(cookie["httpOnly"])
         self.assertFalse(cookie["secure"])
 
-        if not m.atomic_image:  # cannot write to /usr on Atomics, and cockpit-session is in a container
```
marusak, line 112: You changed all of these from `if [not] m.atomic_image` to `if [not] m.ostree_image` but it does not make really sense.

`atomic_image` was `m.image == "rhel-atomic"`. Now you converted it into `ostree_image` which is `m.image == "fedora-coreos"`.

I think all of these conditions can be dropped.

martinpitt: I tried to keep the underlying reason for those. Atomic was OSTree as well, and most of these special cases were about "non-writable /usr" or "local login uses cockpit-ssh" and such. As such, I found these changes less intrusive (while still conceptually correct) than dropping them all.

But of course we won't actually start testing the rhel-7 branch on CoreOS, so we may just as well remove all the special cases. That's going to be a bigger diff noise.

test/verify/check-kdump
martinpitt: Done.

test/verify/check-connection
martinpitt: OK, dropped all Atomic special cases now.

marusak (approved):
Danke!

PR #15679, opened 2021-04-12
cockpiterror.c: Use glib macro instead of our own implementation

marusak:
I've just updated to F34 and cannot `make` cockpit due to:
```
In file included from /usr/include/glib-2.0/glib/gthread.h:32,
                 from /usr/include/glib-2.0/glib/gasyncqueue.h:32,
                 from /usr/include/glib-2.0/glib.h:32,
                 from src/common/cockpiterror.h:23,
                 from src/common/cockpiterror.c:22:
src/common/cockpiterror.c: In function ‘cockpit_error_quark’:
/usr/include/glib-2.0/glib/gatomic.h:117:5: error: argument 2 of ‘__atomic_load’ discards ‘volatile’ qualifier [-Werror=incompatible-pointer-types]
  117 |     __atomic_load (gapg_temp_atomic, &gapg_temp_newval, __ATOMIC_SEQ_CST); \
      |     ^~~~~~~~~~~~~
/usr/include/glib-2.0/glib/gthread.h:260:7: note: in expansion of macro ‘g_atomic_pointer_get’
  260 |     (!g_atomic_pointer_get (location) &&                             \
      |       ^~~~~~~~~~~~~~~~~~~~
src/common/cockpiterror.c:38:7: note: in expansion of macro ‘g_once_init_enter’
   38 |   if (g_once_init_enter (&quark_volatile)) {
      |       ^~~~~~~~~~~~~~~~~
```

This is quick fix. @allisonkarlitskaya does that make sense?

### DIFF
--- a/src/common/cockpiterror.c
+++ b/src/common/cockpiterror.c
@@ -29,16 +29,4 @@
  * Error codes.
  */
 
-GQuark
-cockpit_error_quark (void)
-{
-  static GQuark domain = 0;
-  static volatile gsize quark_volatile = 0;
-
-  if (g_once_init_enter (&quark_volatile)) {
-      domain = g_quark_from_static_string ("cockpit-error");
-      g_once_init_leave (&quark_volatile, 1);
-  }
-
-  return domain;
-}
+G_DEFINE_QUARK(cockpit-error, cockpit_error)


allisonkarlitskaya (changes requested):
This code is completely ridiculous.

1) volatile isn't necessary with `g_once_init_enter()`/`leave()`.  In that sense, this commit is fine, but:

2) `g_once_init_enter()`/`leave()` isn't even necessary here, because `g_quark_from_static_string()` is gonna return the same value every time anyway, so you don't need to worry about concurrent invocation (`GQuark` does its own locking internally).

3) just like `G_DEFINE_QUARK()` instead please?

marusak:
Thanks @allisonkarlitskaya, updated. (Hopefully correctly :) )

allisonkarlitskaya (approved):
Thanks!

PR #15680, opened 2021-04-12
test: Wait for session termination before the next non-destructive test

### LINKED ISSUE #15544: TestLogin.testConversation sometimes invokes sudo

mvollmer:
Otherwise the last few journal messages of the current test might leak
into the next and wrongly be considered as part of it.

Fixes #15544

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -965,9 +965,16 @@ class MachineCase(unittest.TestCase):
                         "done; until rmmod scsi_debug; do sleep 1; done")
 
         # Terminate all lingering Cockpit sessions
-        self.addCleanup(self.machine.execute,
-                        "loginctl --no-legend list-sessions | awk '/web console/ { print $1 }' | "
-                        "xargs --no-run-if-empty -L1 loginctl terminate-session")
+        def terminate_sessions():
+            sessions = self.machine.execute("loginctl --no-legend list-sessions | awk '/web console/ { print $1 }'").strip().split()
+            for s in sessions:
+                self.machine.execute("loginctl terminate-session %s" % s)
+                # Wait for it to be no longer active. Sometimes
+                # sessions are permanently stuck in state "closing",
+                # but that's fine since they won't do any harm.
+                m.execute("while loginctl show-session %s | grep -q 'State=active'; do sleep 1; done" % s)
+
+        self.addCleanup(terminate_sessions)
 
     def tearDown(self):
         if self.checkSuccess() and self.machine.ssh_reachable:


martinpitt (approved):
Many thanks for tracking this down!

martinpitt (changes requested):
Oops! that breaks $OMGmany tests. Should we start applying this to testExpire only for now?

mvollmer:
> Oops! that breaks $OMGmany tests. Should we start applying this to testExpire only for now?

I do some spot tests to see why the sessions don't terminate...

mvollmer:
I couldn't get it to fail locally, so I put some debug output into testlib. Let's see. But the plan is to wait only a limited time, and then just continue instead of failing. The idea is that lingering sessions are blocked on something and wont produce any more output, and the waiting period takes care of slow things like sudo.

mvollmer:
> But the plan is to wait only a limited time, and then just continue instead of failing.

Better plan: declare the global machine to be broken and restart it.

mvollmer:
The only sessions that keep stuck are ones for "alice" on fedora-33... so that might be deterministic and fixable...

mvollmer:
`test/verify/check-static-login TestLogin.testClientCertAuthentication` seems to produce the stuck session.

mvollmer:
> test/verify/check-static-login TestLogin.testClientCertAuthentication seems to produce the stuck session.

This is a known thing and the test itself ignores stuck sessions, let's do the same in testlib.py.

But, isn't this the thing that Linux, cgroups, and systemd are supposed to be good at? *shrug*

martinpitt:
@mvollmer: Sure,  but we don't enable logind's `KillUserProcesses=` option. That *would* do the cleanup, but distros have never enabled it by default as it breaks screen, tmux and the like. Without that, login keeps sessions around as long as they still have processes.

mvollmer:
> Without that, login keeps sessions around as long as they still have processes.

The stuck sessions here have no processes. This must be some bug somewhere, similar to https://bugs.freedesktop.org/show_bug.cgi?id=89024

martinpitt:
Current [fedora-33 run](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15680-20210419-065427-0bb1666b-fedora-33/log.html#87) failed with an unexpected message in `TestHostSwitching.testNoAutoconnect`: "/system/overview.css: external channel failed: terminated". This is already on the weather report, and thus unrelated. Thanks for figuring this out!

martinpitt (approved):

martinpitt:
I sent PR #15715 to fix the above flake.

PR #15682, opened 2021-04-12
network: flip direction of the MAC dropdown in add bond dialog

KKoukiou:
Relevant to: https://bugzilla.redhat.com/show_bug.cgi?id=1946877

This is far from optimal but I was not able to find a way to let it
escape the PF4 dialog.
It needs to be ported to PF4 anyway, which will properly solve the issue.


martinpitt (approved):
Thanks! Good enough for the PF4+bootstrap hack :grin:

PR #15683, opened 2021-04-13
test: Remove unused files

marusak:
Some of them were left behind after machines were moved out and about
rest I have no idea what they were used for.


martinpitt (approved):
Danke! These are even some leftovers from the cockpit-kubernetes removal :grin:

PR #15685, opened 2021-04-14
networkmanager: Fix device connection button

martinpitt:
Commit 5c7b284a082 broke the "connect device" button. Fix the initial
condition in connect() to what we had before that commit: we can connect
if there is either a MainConnection that we can activate, or a device
with ghost settings. (Thanks to Marius Vollmer for this).

Also fix the modify call, `self.dev` is not a thing.

https://bugzilla.redhat.com/show_bug.cgi?id=1946874

### DIFF
--- a/pkg/networkmanager/network-interface.jsx
+++ b/pkg/networkmanager/network-interface.jsx
@@ -123,7 +123,7 @@ export const NetworkInterfacePage = ({
     }
 
     function connect() {
-        if (iface.MainConnection && !(dev && ghostSettings))
+        if (!(iface.MainConnection || (dev && ghostSettings)))
             return;
 
         function fail(error) {
@@ -140,7 +140,7 @@ export const NetworkInterfacePage = ({
 
         with_checkpoint(model, modify,
                         {
-                            devices: self.dev ? [self.dev] : [],
+                            devices: dev ? [dev] : [],
                             fail_text: cockpit.format(_("Switching on <b>$0</b> will break the connection to the server, and will make the administration UI unavailable."), dev_name),
                             anyway_text: cockpit.format(_("Switch on $0"), dev_name)
                         });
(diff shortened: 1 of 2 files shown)


pkg/networkmanager/network-interface.jsx
```diff
@@ -123,7 +123,7 @@ export const NetworkInterfacePage = ({
     }
 
     function connect() {
-        if (iface.MainConnection && !(dev && ghostSettings))
+        if (! (iface.MainConnection || (dev && ghostSettings)))
```
martinpitt, line 126: [before the react rewrite](https://github.com/cockpit-project/cockpit/blob/842d312dd90261046a1fef69477ee92dac254568/pkg/networkmanager/interfaces.js#L2332) the condition was:

    if (!self.main_connection && !(self.dev && self.ghost_settings)) {

So this effectively goes back to that.

martinpitt:
wah wah whitespace, fixed

pkg/networkmanager/network-interface.jsx
KKoukiou: Was it? Expand pkg/networkmanager/interfaces.js  to see it: https://github.com/cockpit-project/cockpit/commit/5c7b284a082edc4c9bd66a1efd662bcfdd2b1dc7#diff-e6968ba34dad066826484f30b54426b4891fbd5e72d45297d295a8f46086e3c8L2332

KKoukiou: Ah right, I just saw the missing leading exclamationmark.

KKoukiou (approved):

martinpitt:
The [rhel-9-0 failure](https://logs.cockpit-project.org/logs/pull-15685-20210414-063126-2ba8de7b-rhel-9-0/log.html#194-2) is fixed by https://github.com/cockpit-project/bots/pull/1924 and is unrelated. I'm afraid this somehow broke in master with the recent bots adjustments.

martinpitt:
Retrying after the bots fix landed.

mvollmer (approved):

PR #15687, opened 2021-04-14
Two coverity fixes

martinpitt:
RHEL's Coverity scanner got fixed to get along with glib ref counting, so the scan results are down to just two unit test issues. These are both false positives, but let's address them to have a clean slate.

https://bugzilla.redhat.com/show_bug.cgi?id=1938693

### DIFF
--- a/src/tls/test-server.c
+++ b/src/tls/test-server.c
@@ -375,6 +375,7 @@ teardown (TestCase *tc, gconstpointer data)
   g_assert_cmpint (wait (NULL), ==, -1);
   g_assert_cmpint (errno, ==, ECHILD);
   /* connection should fail */
+  /* coverity[leaked_handle : FALSE] */
   g_assert_cmpint (do_connect (tc), ==, -ECONNREFUSED);
   g_unsetenv ("COCKPIT_WS_PROCESS_IDLE");
 
--- a/src/ws/test-auth.c
+++ b/src/ws/test-auth.c
@@ -899,7 +899,7 @@ test_multi_step_fail (Test *test,
 {
   GHashTable *headers = NULL;
   gint spot = 0;
-  gchar *id = NULL;
+  g_autofree gchar *id = NULL;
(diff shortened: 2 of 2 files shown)


src/tls/test-server.c
```diff
@@ -375,6 +375,7 @@ teardown (TestCase *tc, gconstpointer data)
   g_assert_cmpint (wait (NULL), ==, -1);
   g_assert_cmpint (errno, ==, ECHILD);
   /* connection should fail */
+  /* coverity[leaked_handle : FALSE] */
   g_assert_cmpint (do_connect (tc), ==, -ECONNREFUSED);
```
allisonkarlitskaya, line 379: Annoying that it still doesn't understand this pattern :(

allisonkarlitskaya (approved):
Thanks!

src/tls/test-server.c
martinpitt: I'm not really sure how to rewrite that in a different way. Just storing the result in a variable and then asserting it doesn't fundamentally change the data flow, I suppose it just doesn't know about the intention of test-* code not making sense with assertions being compiled out?

kdudka: Yes, it is annoying.  I tried to extend Coverity's model of `connect()` to set `errno` in the failing case but it did not make this false positive go away.

kdudka: We already model `g_assertion_message()` as `__coverity_panic__()`.  The problem here seems to be that Coverity fails to infer that `fd` is closed if and only if `do_connect()` returns a negative value, which `-ECONNREFUSED` is for sure.

martinpitt:
@allisonkarlitskaya : Is there some way to restart LGTM? I did log into the website, but I didn't find a retry button. Perhaps you have one, as you registered it?

allisonkarlitskaya:
I don't have this button :(

martinpitt:
OK, just landing then.

PR #15689, opened 2021-04-14
base1: Add option to disable file reading for cockpit.file().watch()

### LINKED ISSUE #14188: Add method to prevent retrieving contents with File.watch
Currently File.watch() automatically retrieves the content of a file when it changes and passes that to the callback.  In some cases (ie: watching an sqlite database or tracking the size of a file) the content of the file may be unimportant or unreadable, but a change in the file should trigger an action.  This is especially important when calling watch() on large files, where a too-large error may be thrown.

A flag could be added to the File options object or watch() function call to control the behavior.
Alternately, the signature of the callback passed to watch() could be inspected to determine if it has enough arguments for the contents and/or tag, and ignore the read() if there aren't enough arguments.

martinpitt:
This is useful when watching large files or non-readable devices, i.e.
when the caller is only interested in the actual change notification.

Add a `read` option to watch(), which can be set to `false` to disable
reading. If not present, keep the original default to not break
backwards compatibility.

Fixes #14188

### DIFF
--- a/doc/guide/cockpit-file.xml
+++ b/doc/guide/cockpit-file.xml
@@ -36,7 +36,7 @@ promise
     .then((new_content, new_tag) => { ... })
     .catch(error => { ... })
 
-file.watch((content, tag, [error]) => { })
+file.watch((content, tag, [error]) => { }, [ { read: boolean } ])
 
 file.close()
 </programlisting>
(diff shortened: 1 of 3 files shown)


marusak (approved):
Thanks :)

PR #15692, opened 2021-04-14
storaged: Apply a dialog when its form is submitted

mvollmer:
Previously, it would reload the page.

* [x] Blocked on https://github.com/patternfly/patternfly-react/issues/5680

### DIFF
--- a/pkg/lib/cockpit-components-dialog.jsx
+++ b/pkg/lib/cockpit-components-dialog.jsx
@@ -327,3 +327,21 @@ export function show_modal_dialog(props, footerProps) {
 
     return dialogObj;
 }
+
+export function apply_modal_dialog(event) {
+    const dialog = event && event.target && event.target.closest("[role=dialog]");
+    const button = dialog && dialog.querySelector("button.apply");
+
+    if (button) {
+        const event = new MouseEvent('click', {
+            view: window,
+            bubbles: true,
+            cancelable: true,
+            button: 0
+        });
+        button.dispatchEvent(event);
+    }
+
+    event.preventDefault();
+    return false;
+}
--- a/pkg/packagekit/autoupdates.jsx
+++ b/pkg/packagekit/autoupdates.jsx
@@ -255,7 +255,7 @@ export class AutoUpdates extends React.Component {
         this.handleChange = this.handleChange.bind(this);
     }
 
-    handleChange() {
+    handleChange(event) {
         const { enabled, type, day, time } = this.state;
 
         if (!validateTime(time))
@@ -266,6 +266,10 @@ export class AutoUpdates extends React.Component {
                 .always(() => {
                     this.setState({ pending: false, showModal: false });
                 });
+
+        if (event)
+            event.preventDefault();
+        return false;
     }
 
     render() {
@@ -274,7 +278,7 @@ export class AutoUpdates extends React.Component {
 
         const enabled = !this.state.pending && this.props.privileged;
         const body = (
-            <Form isHorizontal>
+            <Form isHorizontal onSubmit={this.handleChange}>
                 <FormGroup fieldId="type" label={_("Type")} hasNoPaddingTop>
                     <Radio isChecked={!this.state.enabled}
                            onChange={e => this.setState({ enabled: false, type: null })}
--- a/pkg/users/password-dialogs.js
+++ b/pkg/users/password-dialogs.js
@@ -23,7 +23,7 @@ import { superuser } from "superuser";
 import { Form, FormGroup, TextInput } from '@patternfly/react-core';
 
 import { has_errors } from "./dialog-utils.js";
-import { show_modal_dialog } from "cockpit-components-dialog.jsx";
+import { show_modal_dialog, apply_modal_dialog } from "cockpit-components-dialog.jsx";
 import { password_quality, PasswordFormFields } from "cockpit-components-password.jsx";
 
 const _ = cockpit.gettext;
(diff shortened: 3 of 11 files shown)


KKoukiou (commented):
I think this is not related only to storage page only or pages that use this dialog component. See https://bugzilla.redhat.com/show_bug.cgi?id=1934323 

Can fix fix it in the same PR in all affected places? Maybe PF4 should handle it actually, but it's not clear to me how.

mvollmer:
So we would need to add onSubmit handlers to all `<form>` and `<Form>` elements. Let me try...

mvollmer:
For reference:

```
$ git grep -i '<form '
kdump/kdump-view.jsx:            <Form isHorizontal>
lib/cockpit-components-shutdown.jsx:                    <Form isHorizontal>
lib/form-layout.scss:// a <form class="ct-form"> and layout magic happens.
networkmanager/firewall.jsx:                <Form isHorizontal>
networkmanager/firewall.jsx:                <Form isHorizontal>
networkmanager/index.html:    <form class="ct-form">
networkmanager/index.html:    <form class="ct-form">
networkmanager/index.html:    <form class="ct-form">
networkmanager/index.html:    <form class="ct-form">
networkmanager/index.html:    <form class="ct-form">
networkmanager/index.html:    <form class="ct-form">
networkmanager/index.html:    <form class="ct-form">
packagekit/autoupdates.jsx:            <Form isHorizontal>
playground/react-demo-dialog.jsx:            <form className="ct-form">
shell/machines/machine-add.html:    <form class="ct-form pf-c-form">
shell/machines/machine-change-auth.html:    <form class="ct-form">
shell/machines/machine-change-port.html:    <form class="ct-form">
shell/superuser.jsx:            body = <form className="ct-form" onSubmit={state.apply}>
shell/superuser.jsx:            body = <form className="ct-form" onSubmit={state.apply}>
static/login.html:      <form onsubmit="return false">
storaged/dialog.jsx:            return <Form isHorizontal={isFormHorizontal !== false}>{ rows }</Form>;
systemd/overview-cards/configurationCard.jsx:                <Form isHorizontal>
systemd/overview-cards/realmd-operation.html:                        <form class="ct-form">
systemd/overview-cards/realmd-operation.html:                      <form class="ct-form realms-op-join-only">
systemd/services/timer-dialog.jsx:            <Form isHorizontal>
users/account-create-dialog.js:        <Form isHorizontal>
users/account-details.js:                            <Form isHorizontal>
users/expiration-dialogs.js:        <Form className="expiration-modal">
users/expiration-dialogs.js:        <Form className="expiration-modal">
users/password-dialogs.js:        <Form isHorizontal>
```

martinpitt:
Moving to draft while this is blocked.

mvollmer:
Answered here: https://github.com/patternfly/patternfly-react/issues/5680#issuecomment-831538214

Summary: If a dialog has a single input, pressing ENTER in that input should apply the dialog.

The code here in this PR relies on the browser to submit the form under exactly the right conditions. It is the browser that counts how many input fields are in the form and emits the "submit" event when there is only one and the user presses ENTER in it. Our code intercepts the "submit" event and turns it into a "click" event on the "Apply" button.

I think dividing the responsibilities like that is fine.

So the remaining work here is to intercept the "submit" event on all forms and turn it into a click event on the relevant button.

KKoukiou:
And we still need to handle the submit event for the multi field forms, and in that case ignore it. ev.preventDefault(). Yeap, that will touch a lot of files here :)

mvollmer:
> And we still need to handle the submit event for the multi field forms, and in that case ignore it. 

Hmm, but we can never test this code, right? (Because browsers will not emit the "submit" event for multifield dialogs.)

pkg/packagekit/autoupdates.jsx
```diff
@@ -266,6 +266,10 @@ export class AutoUpdates extends React.Component {
                 .always(() => {
                     this.setState({ pending: false, showModal: false });
                 });
+
```
KKoukiou, line 269: That will also be executed when the button will be clicked, not onky if the form is submitted with onSubmit. I don't see how this would hurt us, that we preventDefault for normal footer buttons. Since they don't have some desired default behavior? So I am actually fine with it.

pkg/users/password-dialogs.js
```diff
@@ -136,7 +136,7 @@ function SetPasswordDialogBody({ state, errors, change }) {
     } = state;
 
     return (
-        <Form isHorizontal>
+        <Form isHorizontal onSubmit={apply_modal_dialog}>
```
KKoukiou, line 139: This is here just for the case where 'Old password' is missing I guess?

pkg/lib/cockpit-components-dialog.jsx
```diff
@@ -327,3 +327,21 @@ export function show_modal_dialog(props, footerProps) {
 
     return dialogObj;
 }
+
+export function apply_modal_dialog(event) {
```
KKoukiou, line 331: That's not so much intuitive but I am happy you found a solution for dialog_open, can't think of anything simpler.

KKoukiou (approved):

pkg/users/password-dialogs.js
mvollmer: I just added it to all <Form>s that I could find, regardless of whether they have a single input or not.

pkg/lib/cockpit-components-dialog.jsx
mvollmer: Would it be better to call this `apply_current_modal_dialog`?

KKoukiou: I did not mean the function name :) I meant that's weird way of triggering the hander. aka trying to find the button with apply class and then clicking it virtually etc.

mvollmer: Yeah, thatw as the easiest way to get cockpit-component-dialog to run a action callback in the right context, and it doesn't seem super wrong.

PR #15693, opened 2021-04-14
lib/pf4: Use alert font style for alert titles

### LINKED ISSUE #15639: Alerts in cockpit use bigger font-size compared to the Alerts in PF4 docs
These are some alerts from our overview page:
Screen Shot 2021-04-01 at 14 06 22

And these are from the PF4 documentation:
Screen Shot 2021-04-01 at 15 25 50

This is some fallout from the PF4-PF3 mix.

garrett:
Fixes #15639

PF4 uses H4 for titles. However, H4 is often set with different font family and
size in other frameworks; notably, but not limited to PatternFly 3.

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -118,3 +118,12 @@ svg {
 .pf-c-input-group :focus {
     z-index: 2;
 }
+
+// Alerts use elements that have fonts set in other frameworks (including PF3);
+// generally, this is an H4 that often has a font size and sometimes family set.
+// Therefore, it should inherit from the alert font set at the pf-c-alert level.
+// https://github.com/patternfly/patternfly/issues/4004
+.pf-c-alert__title {
+    font-size: inherit;
+    font-family: inherit;
+}


pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -118,3 +118,11 @@ svg {
 .pf-c-input-group :focus {
     z-index: 2;
 }
+
+// Alerts use elements that have fonts set in other frameworks (including PF3);
+// generally, this is an H4 that often has a font size and sometimes family set.
+// Therefore, it should inherit from the alert font set at the pf-c-alert level.
+// https://github.com/patternfly/patternfly/issues/4004
+.pf-c-alert__title {
+    font: inherit;
```
KKoukiou, line 127: I think you should put font-size: inherit; font-family: inherit
The font-weight is set propertly from the alert CSS and we need that.

With your changes the alerts will be:
Screen Shot 2021-04-16 at 12 29 39

But they should be like this:
Screen Shot 2021-04-16 at 12 29 28

KKoukiou (changes requested):

pkg/lib/patternfly/patternfly-4-overrides.scss
garrett: Huh. Weird. Font is shorthand for those and _should_ cover it, but being explicit is fine too. And you showed there is a difference in this case.

_Thanks for catching this!_

I guess this means it needs to be adjusted on the upstream issue too...

KKoukiou:
@garrett friendly reminder, do you want to update this one?

garrett:
@KKoukiou: Thanks for the reminder!

I reworked the font to font-size and font-family, rebased, and force pushed.

KKoukiou (approved):
Thanks works as expected now.

PR #15696, opened 2021-04-15
storage: Deal with race on storage initialization

marusak:


marusak:
Ok, I see this "Storage cannot be managed" issue [here](https://logs.cockpit-project.org/logs/pull-15696-20210415-114423-7d25c81e-fedora-34/log.html#170) and also [here](https://logs.cockpit-project.org/logs/pull-15696-20210415-114423-7d25c81e-rhel-9-0/log.html#169)

In both cases `client.features` is `false` in [this condition](https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/devices.jsx#L81).

I don't see `log 1` thus it means that at least one proxy in both init methods are stuck. I'll add some further debug infos.

marusak:
Oh no, `client.features` is false, that means that `init_manager()` failed.

marusak:
Ok, so `org.freedesktop.UDisks2.Manager` proxy fails to be created. [here](https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/client.js#L841)

marusak:
OK, great! So now there isn't any failed storage test that would fail with 'Can not manage storage' while in all of those 10 previous pushes at least a few always failed.
I can see in log that it still would hit this but it can now react when the proxy is finally available.
I'll cleanup after myself and ask for review.

pkg/storaged/client.js
```diff
@@ -393,13 +382,12 @@ function init_model(callback) {
                 var defer = cockpit.defer();
                 client.manager_lvm2 = proxy("Manager.LVM2", "Manager");
                 client.manager_iscsi = proxy("Manager.ISCSI.Initiator", "Manager");
-                wait_all([client.manager_lvm2, client.manager_iscsi],
-                         function () {
-                             client.features.lvm2 = client.manager_lvm2.valid;
-                             client.features.iscsi = (client.manager_iscsi.valid &&
+                Promise.allSettled([client.manager_lvm2.wait(), client.manager_iscsi.wait()]).then(() => {
```
martinpitt, line 385: Checked that we can [use allSettled](https://caniuse.com/?search=Promise.allSettled) with our current babel browser settings.

martinpitt (approved):
Awesome job tracking this down! :100:

marusak:
Ok, not a single storage test again failed on that common flake.
A few oopses `TypeError: Cannot read property 'xfs' of undefined` like on [fedora-34](https://logs.cockpit-project.org/logs/pull-15696-20210420-105420-8af90ee0-fedora-34/log.html#198), [rhel-9-0](https://logs.cockpit-project.org/logs/pull-15696-20210420-105420-8af90ee0-rhel-9-0/log.html#184) or [ubuntu-2004](https://logs.cockpit-project.org/logs/pull-15696-20210420-105420-8af90ee0-ubuntu-2004/log.html#177).

I am thinking if it might be related as I don't recall seeing this problem before.

martinpitt:
[Two](https://logs.cockpit-project.org/logs/pull-15696-20210420-121419-e754af59-fedora-33/log.html#193) [storage](https://logs.cockpit-project.org/logs/pull-15696-20210420-121419-e754af59-rhel-9-0/log.html#182) retries, but they look unrelated and familiar. Thanks!

martinpitt (approved):

PR #15697, opened 2021-04-15
lib: Remove some unneeded code from cockpit-component-dialog

KKoukiou:
* onClick on buttons is fired only with the primary key anyway - no need
for extra check

* modal-in class is not used


marusak (approved):
Thanks :)

PR #1923, opened 2015-03-16
test-avocado: Continously update OS in Cockpit guests.

mvollmer:

PR #15698, opened 2021-04-15
test: Enable TLS for mock insights server

martinpitt:
Latest insights-client dropped the `insecure_connection` option and only
accepts TLS. Change mock-insights to serve TLS, and re-use our "alice"
certificate for it.

Disable certificate verification, as apparently there is no way to
convince the later stages of insights-client to accept a locally added
CA.

### DIFF
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -686,19 +686,20 @@ fi
         m.write("/usr/share/cockpit/subscription-manager/override.json", '{ "features": { "insights": true } }')
 
         # Run a mock version of the Insights API locally and configure
-        # insights-client to access it.
-
-        m.upload(["verify/files/mock-insights"], "/var/tmp")
+        # insights-client to access it. That requires a good enough
+        # TLS mock insights server certificate
+        m.upload(["verify/files/mock-insights", "../src/tls/ca/alice.key", "../src/tls/ca/alice.pem"], "/var/tmp")
         m.spawn("/var/tmp/mock-insights", "mock-insights")
         m.write("/etc/insights-client/insights-client.conf",
                 """
 [insights-client]
 gpg=False
 auto_config=False
-base_url=127.0.0.1:8888/r/insights
+base_url=localhost:8888/r/insights
+# HACK: despite the good insights.cert above, some later part of insights *still* does not respect that
+cert_verify=False
 username=admin
 password=foobar
-insecure_connection=True
 """)
 
         # Initially we are not registered
--- a/test/verify/files/mock-insights
+++ b/test/verify/files/mock-insights
@@ -20,6 +20,7 @@
 from http.server import *
 import json
 import re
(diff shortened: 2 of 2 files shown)


martinpitt:
This blocks https://github.com/cockpit-project/bots/pull/1921 and #1923. Triggering extra test against that new 8-5 image. I only test ed it locally against the latter.

marusak:
Triggering also `rhel-8-4@bots#1923`

test/verify/files/mock-insights
```diff
@@ -137,7 +138,9 @@ class handler(BaseHTTPRequestHandler):
 
 
 def insights_server(port):
-    HTTPServer(('', port), handler).serve_forever()
+    httpd = HTTPServer(('', port), handler)
```
marusak, line 141: I got this far :)

test/verify/check-system-info
```diff
@@ -686,19 +686,32 @@ fi
...
         m.spawn("/var/tmp/mock-insights", "mock-insights")
         m.write("/etc/insights-client/insights-client.conf",
                 """
 [insights-client]
 gpg=False
 auto_config=False
-base_url=127.0.0.1:8888/r/insights
+base_url=localhost:8888/r/insights
+# HACK: despite the good insights.cert above, some later part of insights *still* does not respect that
+cert_verify=False
```
marusak, line 712: Do we need all the cert fun if we just then ignore it?

test/verify/files/mock-insights
martinpitt: Some random dude wrote [this blog post](https://www.piware.de/2011/01/creating-an-https-server-in-python/) a decade ago, which was useful :grin:

test/verify/check-system-info
martinpitt: The server still needs to load a TLS certificate which makes the initial connection not hiccup. I started with that, and waded through all the follow-up errors before I gave up there. I can try to simplify it a bit.

martinpitt: See e.g. [this first attempt](https://github.com/cockpit-project/bots/pull/1921#issuecomment-820242174) with a simple certificate, which was already doomed. It really doesn't want to be that easy :cry:

marusak (approved):
Looks good, works, lets not waste way too much time on it. Thanks!

test/verify/check-system-info
martinpitt: Indeed it works with the plain "alice" certificate, yay! Much simpler. Dang, I did this the wrong way around..

marusak:
`rhel-8-4@bots#1921 ` does not say much

marusak (approved):
sooo much simpler :)

martinpitt:
Heh, yes, triggered the wrong one first, but  also the right one afterwards :)

PR #15699, opened 2021-04-15
login: Add check for String.prototype.replaceAll

garrett:
As we're using `"".replaceAll()` in Cockpit already, we should have a check for it.

This bumps up our minimum supported browser versions to mid-2020.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -212,6 +212,7 @@ if (window.NodeList && !NodeList.prototype.forEach)
                req("console", window) &&
                req("pushState", window.history) &&
                req("textContent", document) &&
+               req("replaceAll", String.prototype) &&
                req("supports", window.CSS) &&
                css("display", "flex") &&
                css("display", "grid");


garrett:
This is based on a conversation on IRC with @marusak where he said we're using this and not pollyfilling it. (Or won't be soon? Not sure.)

garrett:
Companion part of this for communication via the website is @ https://github.com/cockpit-project/cockpit-project.github.io/pull/402

marusak:
We should then also bump it up here https://github.com/cockpit-project/cockpit/blob/master/.babelrc.json before we teach it to auto recognize it.

martinpitt (approved):
Thanks!

I'm frankly a bit shocked -- why do we have all that babel stuff if it doesn't care about such stuff? I see some replaceAll() wrangling in the webpacks, but that just seems to be jQuery, not a polyfill. Hmm. So let's land this, the website update, and do a follow-up for bumping the browser versions in babelrc.

PR #15700, opened 2021-04-15
node_modules: manage as a git submodule

allisonkarlitskaya:
- [x]  #15719


.github/workflows/unit-tests-refresh.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: sudo containers/unit-tests/start CC=clang distcheck
 
       - name: Run i386 gcc distcheck test
-        run: sudo containers/unit-tests/start :i386 NO_NPM=1 distcheck
+        run: sudo containers/unit-tests/start :i386 distcheck
```
martinpitt, line 34: Shouldn't this use `WEBPACK_PACKAGES=''` then, to also explicitly test that  mode?

.github/workflows/unit-tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
          - check-memory
          - CC=clang :i386 check-memory
          - CC=clang distcheck
-         - :i386 NO_NPM=1 distcheck
+         - :i386 distcheck
```
martinpitt, line 12: Dito

HACKING.md
```diff
@@ -80,14 +80,6 @@ to build everything.  Cockpit has a single non-recursive Makefile.  You can
 only run `make` from the top-level and it will always rebuild the whole
 project.
 
-When building from git, you can optionally set the `NO_NPM` environment
-variable to `1` to prevent `autogen.sh` from invoking `npm`.  As a result,
```
martinpitt, line 84: ... and this could be changed to document `$WEBPACK_PACKAGES` then, either for building only one (a common web developer case), or nothing at all (C developer case)

.gitignore
```diff
@@ -55,7 +55,6 @@ depcomp
 /containers/*/rpms
 /cockpit.service
 /cockpit-wrapper
-/node_modules
```
martinpitt, line 58: Does it hurt/conflict with a submodule to leave that in? I wonder what happens if you don't pull the submodule, and just run `npm install` directly manually. You may not be able to push that anywhere, but a local build still ought to work?

Makefile.am
```diff
@@ -220,15 +193,13 @@ clean-local::
...
-node_modules/.stamp: package.json
-	@echo ''
-	@echo '    package.json changed'
-	@echo ''
-	@echo '    Please run:'
-	@echo ''
-	@echo '        tools/npm-install'
-	@echo ''
-	@false
+FORCE:
```
martinpitt, line 196: Is that some leftover from experimentation? If that's supposed to land, it needs a comment at least. There is no other reference than the dependency of package-lock.json, and it's neither phony nor ever built.

autogen.sh
```diff
@@ -39,8 +39,6 @@ if [ -z "$(which glib-gettextize)" ]; then
     exit 1
 fi
 
-test "${NO_NPM:-}" != "1" && tools/npm-install
```
martinpitt, line 42: This actually helps me with cleaning up the -dist stuff as well. But this still feels incomplete - autogen.sh should ensure in *some way* that node_modules is available. Maybe with a simple `[ -e package-lock.json  ] || tools/node-modules`?

containers/unit-tests/start
```diff
@@ -13,7 +13,7 @@ while test $# -gt 0; do
   case "$1" in
     --docker) docker="docker";;
     --podman) docker="podman";;
-    NO_NPM=1) no_npm='--env=NO_NPM=1';;
+    NO_NPM=1) no_npm='--env=NO_NPM=1';; # noop
```
martinpitt, line 16: Why do we still need that then? Nothing looks at it any more.

martinpitt (changes requested):
Initial review of the `NO_NPM` → `WEBPACK_PACKAGES` transition. This feels like it can and should land separately, it shouldn't be mangled into the node_modules/ handling. Can you please split it out into a new PR, with some fixes? Thanks!

Makefile.am
allisonkarlitskaya: This is a farly standard idiom: https://www.gnu.org/software/make/manual/html_node/Force-Targets.html

I'll add a comment, though.

autogen.sh
allisonkarlitskaya: I disagree with this.  The point here is that `node_modules` should be automatically created by make on an as-needed ("as" = "if and when") basis.  If we do a `WEBPACK_PACKAGES=''` build, for example, then `node_modules` will never be checked out.

containers/unit-tests/start
allisonkarlitskaya: I left it there just in case it would get invoked from a workflow or something, but I guess not, actually...

Makefile.am
martinpitt: I still don't see it. package-lock.json is not phony conceptually -- it is a real file that make should build. What's wrong with just `package-lock.json:` ?

tools/node-modules
```diff
@@ -0,0 +1,182 @@
...
+    fi
+
+    # double check that the tag is what it claims to be
+    test "$(git --git-dir "${CACHE_DIR}" rev-parse "${tag}")" = "${sha}"
+
+    # verify that our package.json is equal to the one the cached node_modules
+    # was created with, unless --force is given
+    if [ "${1-}" != "--force" ]; then
+        local our_package_json_sha="$(cat package.json | sha1sum)"
+        local their_package_json_sha="$(git --git-dir "${CACHE_DIR}" cat-file -p "${sha}":.package.json | sha1sum)"
```
martinpitt, line 45: Can we use sha256sum here? We still get the [occasional audit](https://bugzilla.redhat.com/show_bug.cgi?id=1934937) for the long-broken SHA-1, so let's please avoid that noise in `git grep`.

HACKING.md
```diff
@@ -247,6 +244,28 @@ In a Debian/Ubuntu build environment you can build debs with
...
+automatically unpacked from a cache kept in a separate git repository.  You can
+force the unpack to occur using the `tools/node-modules checkout` command, but
+this shouldn't be necessary.  In the event that you need to modify
+`package.json` (to install a new module, for example) then you'll need to run
+`tools/node-modules rebuild` to create a new cache from the result of running
+`npm install` on your new `package.json`.
+
+Before landing changes to the main cockpit project (or proposing a PR), this
+rebuild will need to be pushed to the cache repository.  `tools/node-modules
+push` will do this for you.
```
martinpitt, line 259: This feels conceptually wrong to me. A PR can get rejected, superseded, rebased, and contributors don't even have the permission to push to the cache. We will somehow need to find a way to do this on the CI side.

HACKING.md
```diff
@@ -247,6 +244,28 @@ In a Debian/Ubuntu build environment you can build debs with
...
+
+The `tools/node-modules` script inspects the `GITHUB_BASE` environment variable
+to determine the correct repository to use when fetching and pushing.  It will
+strip the repository name (leaving the project- or username) and use the
+`node-cache.git` repository in that namespace.  If `GITHUB_BASE` is unset, it
+will default to `cockpit-project/node-cache.git`.
+
+A local cache is maintained in `~/.cache/cockpit-dev/node-cache.git`.
```
martinpitt, line 267: Remember to adjust that according to the "needs github org in the path" comment from above.

tools/node-modules
```diff
@@ -0,0 +1,182 @@
...
+
+  - git add node_modules && git commit
+
+EOF
+}
+
+cmd_push() {
+    local tag="sha-$(git -C node_modules rev-parse HEAD)"
+    git -C node_modules tag -f "${tag}" HEAD
+    git -C node_modules push git@github.com:cockpit-project/cockpit-node_modules "${tag}"
```
martinpitt, line 106: hardcoded repo

martinpitt (changes requested):
Some initial review of the actual node_modules handling. Overall this is exciting!

My main gripe is with developers pushing to the cache before a PR lands, or is approved, or even is created. (Annoying for cockpit-project members, and creating an assymetry for contributors). We probably already talked about that, but it feels like cache handling ought to be more automated. Otherwise there are just a bunch of local nitpicks.

tools/node-modules
```diff
@@ -0,0 +1,185 @@
...
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/cache.git"
+CLONE_URL="https://github.com/cockpit-project/cockpit-node_modules"
+
+cmd_remove() {
+    # if we did this for ourselves the rm is enough, but it might be the case
+    # that someone actually used git-submodule to fetch this, so clean up after
+    # that as well.  NB: deinit nicely recreates the empty directory for us.
+    rm -rf node_modules
+    git submodule deinit node_modules
+    rm -rf "$(git rev-parse --absolute-git-dir)/modules/node_modules"
```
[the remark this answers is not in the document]
allisonkarlitskaya: If you do normal use of git submodules, git will setup a .git file (in the same way as it does for worktrees) which points to the gitdir for the submodule at that location.

This can also happen as a result of `git submodule absorbgitdirs`.  Here's the docs for that command:

```
absorbgitdirs
   If a git directory of a submodule is inside the
   submodule, move the git directory of the submodule into
   its superproject’s $GIT_DIR/modules path and then connect 
   the git directory and its working directory by setting 
   the core.worktree and adding a .git file pointing to the
   git directory embedded in the superprojects git
   directory.
```

Annoyingly, this isn't cleaned up on deinit.

This is basically just an attempt to make sure we're really squeaky clean.  It's part of the "so clean up after that as well..." mentioned in the comment above.

tools/node-modules
```diff
@@ -0,0 +1,185 @@
...
+}
+
+cmd_checkout() {
+    # we check out the node_modules corresponding to the gitlink in the index
+    local sha="$(git submodule status --cached node_modules | cut -c2-41)"
+    local tag="sha-${sha}"
+
+    if [ ! -d "${CACHE_DIR}" ]; then
+        [ "${V-}" != 0 ] || echo "  INIT     ${CACHE_DIR}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's not a bad idea.  I was trying to minimise the size of the output in the case of `$V=0`, but I guess I can keep the `$V` check the way it is.

tools/node-modules
allisonkarlitskaya: Weird.  At some point I thought I had changed that, because ya, we have no need to match git here.  Will do.

HACKING.md
allisonkarlitskaya: I disagree.  I keep comparing this to image refreshes, because I think it's a very good comparison.  Bots (or even humans) push images to the mirrors, and then we update a symlink (in this case gitlink) in the cockpits repo and propose a PR and do review there.  We can reject broken vm images, and we can reject broken node_modules updates in exactly the same way.  Also note that the node_modules commits are completely independent of each other:  There is no history, only tags to single parentless commits...

The only thing missing here is a prune.

I guess you imagine some kind of a procedure on a PR to have a bot come along and turn package.json into a commit in our `node-cache` repository?  This might be helpful, but it's also annoying: it means that if you're hacking on something that requires a change to package.json, you'd first need to create a PR, push your package.json change, ask the bot to do the build, pull the bot's commit into your local repo, and then continue working.... :/

Makefile.am
```diff
@@ -220,15 +193,13 @@ clean-local::
...
+
+V_NODE_MODULES = $(V_NODE_MODULES_$(V))
+V_NODE_MODULES_ = $(V_NODE_MODULES_$(AM_DEFAULT_VERBOSITY))
+V_NODE_MODULES_0 = @V=0
+package-lock.json: FORCE
+	$(V_NODE_MODULES) tools/node-modules make_package_lock_json
```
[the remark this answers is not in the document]
allisonkarlitskaya: I think we'll need to discuss this point face-to-face, since from my viewpoint, the thing you're complaining about is almost the entire benefit of this change.

tools/node-modules
```diff
@@ -0,0 +1,185 @@
+#!/bin/sh
+
+set -eu
+
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/cache.git"
+CLONE_URL="https://github.com/cockpit-project/cockpit-node_modules"
```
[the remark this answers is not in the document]
allisonkarlitskaya: I had accidentally failed to push some earlier changes that cleaned this up.  Did that now.

tools/node-modules
```diff
@@ -0,0 +1,185 @@
+#!/bin/sh
+
+set -eu
+
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/cache.git"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Are you suggesting a different cache directory per organisation?  Why?

.gitignore
allisonkarlitskaya: I'm pretty sure this would conflict with the "always make sure `node_modules` matches the gitlink" logic that gets run every time we type `make`.  That particular piece of logic is *very* aggressive: it would delete your `node_modules` and replace it with the downloaded one.

Makefile.am
allisonkarlitskaya: `package-lock.json:` without any dependencies is a rule for producing `package-lock.json` if it doesn't exist (it can never be out of date, because it has no dependencies).

`package-lock.json: FORCE` is a rule that will run every single time someone asks for package-lock.json to be updated.  Our script is run every time in order to check that `node_modules` is correctly in sync with the gitlink.

tools/node-modules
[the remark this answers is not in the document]
martinpitt: I meant that `log()` should check `$V` by itself.

HACKING.md
martinpitt: I had assumed that the classic `npm install` would still work locally. If that's not the case, then we can try this, but I still don't like that it divides PR contributors into two classes. (this is why I really like the dist caching -- it doesn't get in the way and doesn't make workflows more complicated). I think on IRC you mentioned something like that, like computing a hash over package-lock, and using that as lookup value.

Anyway, I don't want this to be a hard veto -- we can try it, and if it doesn't work for us, go back.

martinpitt: image refreshes  are actually a good comparison, right -- I can change the scripts and run image-create locally (the equivalent of `npm install`), and then can develop against this right away. There is no need to first push this somewhere and let the bot commit anything. But AFAICS this would work here, right? For locally running webpack/image-prepare etc. I wouldn't *have* to push, I can first test this locally if it's even something that I want, right?

tools/node-modules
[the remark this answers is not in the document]
martinpitt: Wouldn't multiple origins in the same cache clone hopelessly collide? It seems much cleaner to have one cache repo per cloned remote repo rather than switching `origin` around every time I set `$GITHUB_BASE`.

Makefile.am
martinpitt: Ah, I see, thanks! (So this actually *is* like a phony target)

Makefile.am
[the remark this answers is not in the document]
martinpitt: So you suggest we have two cases? (1) developer builds from git could involve internet-y things during `make`, and (2) builds from release tarballs can't ever change package.json and `make` will be guaranteed to work offline. I can certainly live with that. This PR does (1) now, and we need to check what happens in the (2) case -- i.e. `make dist`, unpack the tarball, touch package.json, and see what explodes, and if that error message needs to be improved somehow.

tools/node-modules
```diff
@@ -0,0 +1,173 @@
...
+
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/node-cache.git"
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/node-cache"
+CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+PUSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
+
+v0() {
+    printf "  %-8s %s\n" "$1" "$2"
```
martinpitt, line 13: Please move the `${V}` check here and call this `log()` or `message()` or something such, as suggested in the earlier review round.

tools/node-modules
```diff
@@ -0,0 +1,173 @@
...
+    # package-lock.json is used as the stamp file for all things that use
+    # node_modules (mostly webpack), so this is the main bit of glue that
+    # drives the entire process
+
+    # We try our best not to touch package-lock.json unless it actually changes
+
+    # This isn't going to work for a tarball, but as long as
+    # package-lock.json is already there, we're OK
+    if [ ! -e .git -a -e package-lock.json ]; then
+        exit 0
```
martinpitt, line 147: This means that if someone is trying to work on an unpacked release tarball, this will silently succeed and ignore attempts to change package.json. ISTM that this should have a proper error message and exit 1 if package.json is newer.

HACKING.md
```diff
@@ -56,6 +56,11 @@ Creating VM images locally (not necessary for running tests) needs the following
 
     $ sudo dnf install virt-install
 
+Updating the `node_modules` (in case you need to modify `package.json`)
```
martinpitt, line 59: please squash the "document t/n-m" commit into the previous one that introduces it.

HACKING.md
martinpitt: ... does this even need to be documented that explicitly? Perhaps it's enough to just say `~/.cache/cockpit-dev/` here? It's not like we want people to start making assumptions about it, it should remain a private implementation detail IMHO.

martinpitt (changes requested):
Thanks! The first three commits seem fine to me, just in case  you want to land them separately (but there's no hurry to it, and I have a mild preference in landing it all together).

Makefile.am
allisonkarlitskaya: Not exactly.  The timestamp of the file is still considered in this case: in the event that we replace the file with a new version, the targets depending on it will get rebuilt.

Pure phony targets never stat the underlying "file".

Makefile.am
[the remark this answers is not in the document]
allisonkarlitskaya: I'll add some extra checks to make sure that this stuff really absolutely never runs from a tarball.

tools/node-modules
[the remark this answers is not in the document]
allisonkarlitskaya: Right.  Now I understand this.  Of course.  Will fix :)

HACKING.md
allisonkarlitskaya: Exactly correct.

But as is the case with image refreshes, if you want *other* people to be able to use it (and in any case, before landing a PR) you'll need to publish it.

tools/node-modules
allisonkarlitskaya: I wanted to keep the `${V}` check out of here to prevent seeing the info output in `set +x` mode in case `V` != `0`.  But maybe it's not so awful...

tools/node-modules
allisonkarlitskaya: Will add.

HACKING.md
allisonkarlitskaya: Agreed.

allisonkarlitskaya:
The last unaddressed bit is the `NO_NPM` vs `WEBPACK_PACKAGES=''` stuff.  I'm not 100% how to proceed here, particularly, because `WEBPACK_PACKAGES=''` currently won't stop `node_modules` from being checked out (due to some font files used in the guide).

martinpitt:
@allisonkarlitskaya : You mean testing `WEBPACK_PACKAGES=''` in unit tests? I'm ok with not covering this for now, as it's a rarely used feature. Honestly, I'd much rather go into a direction where you can clone, `npm install` (or its moral equivalent), and `tools/webpack-watch kdump` (or again, it's future replacement).

tools/node-modules
```diff
@@ -0,0 +1,186 @@
...
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/node-cache"
+CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+PUSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
+
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/${GITHUB_REPOSITORY}.git"
+
+message() {
+    [ "${V-}" != 0 ] || printf "  %-8s %s\n" "$1" "$2"
```
martinpitt, line 13: This is the wrong way around. I want the messages to appear with `make V=1`, right?

Not sure if make exports `V=0` by default, or if it's empty. Probably best to say `[ "${V-}" != 1 ] || printf` and ignore the possibility of `V=2` and similar?

martinpitt (dismissed):

tools/node-modules
martinpitt: Nevermind -- I see the `V_NODE_MODULES_0` bit now, so this is some "inverse" switch.

martinpitt:
I played around with this locally. This works very well, by and large.

Some bugs:
 - [ ] Calling `tools/node_modules` enables `set -x` by default, which is too noisy. 
 
 - [ ] Changing package.json and running `make` is a silent successful no-op. This either needs to invoke "rebuild node_modules", or at least tell me what to do.

    starter-kit [does the former](https://github.com/cockpit-project/starter-kit/blob/master/Makefile#L162) with a `package-lock.json: package.json` rule, which automatically rebuilds everything.

    Current cockpit master fails the `make` immediately with "package.json changed. Please run..." which is fine as well.


For the record, in yesterday's meetings the main issue where we agreed to disagree is the general architecture. We need some way to link a current cockpit commit to some ref in the node-cache.git, but we disagree how to do that. My counter-proposal is to make package.json more precise (avoid `^` and pin down devDependencies), and use sha256sum(package.json) as an intrinsic property for a `sha-<checksum>` tag lookup in the node-cache.git.

 - Lis prefers a submodule. That prevents an attack scenario where a malicious developer pushes a malicious set of node_modules/ to the node-cache.git and force-pushes an existing tag to point to that malicious code.

 - Martin thinks that this is not an useful attack scenario, as then an attacker can as well push cockpit.git itself and 0wn our systems that way. The price that we have to pay for submodules is just too high: making it hard to work without the cache, making it hard for external contributors to submit PRs that work, having inconsistencies due to the redundancy between package-lock.json and the reflink, and making it harder for ourselves as we have to use a special tool instead of well-known workflows.

As a result, I'd like @mvollmer or @mmarusak to do the final review.

allisonkarlitskaya:
Reply to above:

 - I like the `set -x` because it shows exactly what's happening.  It makes this all feel a bit less "magic", particularly for the "rebuild" and "push" cases.  I go to great care to turn it off for the case where we run from the Makefile

 - the other thing is bug :)

As for the architectural disagreement, I'd frame my stance as less to do with security (although this is a natural following of the main point), but more to do with wanting to define a build of cockpit as something decided, deterministically, by what's in the cockpit repository (and, specifically, the hash of the commit that we're building).  Right now, we can have builds of a given PR failing from one to the next just because some node module changed.  Our releases are non-deterministic in a similar way.  Having a cache won't substantially change that logic: our build could just start failing depending on if we rebuilt the cache or not.  Having a gitlink that specifies an exact set of `node_modules` gives us absolute certainty about what we're building.

Another way of looking at this: we were talking about committing package-lock.json to our repository a while ago to get these same benefits (even though it wouldn't have given us the same byte-for-byte guarantees that git does).  It would have almost exactly the same set of drawbacks as the approach here, and it would trade the "external contributor PR" one where one of us needs to rebuild the node_modules for them with one where they throw a giant blob of a package-lock.json patch at us and we just have to trust it (since nobody's gonna read it anyway).

martinpitt:
We both agree that we don't want to commit package-lock.json. It seemed promising at first sight, but doesn't live up to its promise and doesn't help with making builds faster.

> define a build of cockpit as something decided, deterministically, by what's in the cockpit repository

We fully agree on that goal. It has long bugged me that our builds can spontaneously break, and our releases are effectively totally untested.

Our disagreement is that you want to have an extrinsic (and redundant) submodule reflink, and want an intrinsic `sha(package.json)` as the definitive answer to "what do we want node_modules/ to be". Everything else (having a tool to manage node_modules, storing them in git, using tags as "keepalive anchors", etc.) we agree on.

martinpitt:
Something is wrong on debian-stable: It failed with "Error: Cannot find module 'chrome-remote-interface'" twice in a row now. Something that runs inside the VM build that somehow breaks it? Some incompatibility with older git version?

martinpitt:
Thanks for the fixes! This now looks and works well. I can't retract my nack while this is still in draft, but once you mark it ready for review I will. I.e. no veto, and I'm fine with actually running this for a while, and seeing how it works for us.

There are still some missing pieces here:
 1. Make npm-update workflow/bot deal with refreshes that don't change package.json, i.e. send regular PRs for them
 2. Making these PRs more explicit by dropping all `^` from package.json and pin down the microversions instead, so that package.json reflects what's inside the node_modules reflink
 3. Regularly clean up old tags in node-cache.git

1 and 3 are relatively urgent, and need to be quick follow-ups. I realize 2 is more contentious, and we probably need to have a team vote -- but it's also not urgent.

Cheers!

martinpitt:
> Having a gitlink that specifies an exact set of node_modules gives us absolute certainty about what we're building.

Your statement is correct about "certainty" in the sense of "reproducibility", but certainly not about "certainty" being "trustworthy": How would you audit a "use latest eslint" PR that bumps package.json and includes a reflink? Right now we only have the former, which is obvious to look at. But reflink commits are practically unauditable. I trust our own bots to not do something weird, but if your attack model is "someone steals a github token/cookie", then that means that we really need to review/audit PRs from our own team as well. We usually do, but now that we have to commit reflinks instead of having them built from well-defined and small text content, which weakens our reviews *a lot*.

marusak:
I am not convinced. I feel like I am missing the real point here. Committing package-lock would be way more transparent - it is json file where are defined versions of all npm modules that one uses. Compared to some blob (or how it is going to be stored?) somewhere (and where? both https://github.com/cockpit-project/cockpit-node_modules and https://github.com/cockpit-project/node-cache are empty).

> The biggest benefit here is that normal users won't need to run npm anymore.

Whats problematic about running npm?

Is this opt-in and automated? I mean, if I send PR where I bump/add/drop some npm module, it will recognize that there is no blob for it in git submodule, will build a new one and publish it into appropriate repo? Or do I need to generate this /node-modules directory, commit is somehow somewhere and then send PR? (if latter how does it work when two PRs are doing changes to npm modules?)

martinpitt:
> Committing package-lock would be way more transparent

I'm with Lis on that one -- package-lock.json is too large, diffs are too noisy to meanigfully audit/understand.

It *is* reasonably reproducible: Starting npm-install twice from a clean git tree produces *identical* package-lock.json. The bit that is weird is running it twice in a row:
```
git clean -ffdx
tools/npm-install
mv package-lock.json{,.prev}
tools/npm-install
```

after that, the diff is 3746  lines, with lots of:
```diff
@@ -9169,7 +9169,8 @@
       "dependencies": {
         "JSONStream": {
           "version": "1.3.5",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.5.tgz",
+          "integrity": "sha512-E+iruNOY8VV9s4JEbe1aNEm6MiszPRr/UfcHMz0TQh1BXSxHK+ASV1R6W4HpjBhSeS+54PIsAMCBmwD06LLsqQ==",
           "optional": true,
           "requires": {
             "jsonparse": "^1.2.0",
```
That may be fixable in npm-install, of course, so that it becomes idempotent.

Nevertheless, I feel the sweet spot is pinning down our top-level devDependencies in package.json and refreshing these every week, like we already do for runtime depenencies. That's an understood cut-and-dried process, produces auditable PRs, and probably comes closest to "what is the definition of what we expect node_modules/ to be". 

Of course caching node_modules/ has its own advantages -- checking this out from git is much faster and reliable than running npm install. I welcome this aspect as an improvement. I would not veto committing package-lock.json, but I think such a caching solution is nicer overall (I really like the dist/ caching from cockpit-machines -- to me this feels exactly right. Thanks @allisonkarlitskaya for coming up with that idea!).

allisonkarlitskaya:
I don't like this proposal anymore, at least not in its current form.  All of this "security thinking" caused me to realise that generating the `node_modules` in a completely non-transparent way (since it's just a giant block of code, and indeed, none of us will review that) in a completely uncontrolled environment on a random developer machine who happened to last touch package.json?  ehhh...

So if my machine is pwned, then we're all gonna be pwned after that.

This needs a bit of rethinking.

marusak:
> after that, the diff is 3746 lines, with lots of:

That sounds wrong. It is advised to commit this file so I guess this is some fluke? We should update lock file only when we change package.json.
If I understand correctly you removed the lock and generated it once again, right? Does the same happen if you commit it, then clean up tree and then update npm?

> Nevertheless, I feel the sweet spot is pinning down our top-level devDependencies in package.json and refreshing these every week, like we already do for runtime depenencies.

That would not prevent from transitive broken deps (which happen every now and then). Also it does not provide the same tree for everyone.

> Of course caching node_modules/ has its own advantages -- checking this out from git is much faster and reliable than running npm install. I welcome this aspect as an improvement. 

Right, so it is about speed. Not sure about how much reliable is it compared to npm with package-lock though. Nothing against making things faster, but I don't feel like current implementation cons outweigh the speed improvement it brings.

martinpitt:
> If I understand correctly you removed the lock and generated it once again, right? 

I tested two cases: two npm-installs from a completely clean tree -- this is reproducible. And then clean tree and running npm-install twice -- that produces that s/bundled/integrity/ diff. (And I'm sure that's fixable somehow).

> That would not prevent from transitive broken deps (which happen every now and then).

Does it? So far we usually fix random build breakages by explicitly installing the previous version with a fixed version number. I propose to just always do that, so that package.json reflects reality more accurately, and diff in PRs will show us what actually changed.

marusak:
> Does it? So far we usually fix random build breakages by explicitly installing the previous version with a fixed version number. I propose to just always do that, so that package.json reflects reality more accurately, and diff in PRs will show us what actually changed.

If you would specify version for every single npm package that we have in node_modules you would in fact create package-lock. So when you do npm-install with lock it will install always the same package versions for every single npm, no matter how far in deps chain they are. That surely solves issue when suddenly all PRs fail because there was update of some dep of dep of dep... that we don't lock (nor does any other package).

martinpitt:
> If you would specify version for every single npm package that we have in node_modules you would in fact create package-lock.

Right, that was not my intention. I would merely pin down the top-level packages that we currently have in package.json, and ignore transitive dependencies. This is a compromise between "some noise" and "have auditable package.json diffs".

tools/node-modules
```diff
@@ -0,0 +1,206 @@
...
+            exit 1
+        fi
+    done
+
+    cmd_remove
+
+    npm install
+
+    cp package.json node_modules/.package.json
+    echo /.cache >> node_modules/.gitignore
```
mvollmer, line 102: What is this for?  Does `npm install` create this directory?

tools/node-modules
```diff
@@ -0,0 +1,206 @@
...
+    git -C node_modules -c gc.autoDetach=false commit -a -m "$(sha256sum package.json)"
+
+    cat <<EOF
+Next steps:
+
+  - tools/node-modules push
+
+  - git add node_modules && git commit
+
+EOF
```
mvollmer, line 115: `cmd_rebuild` doesn't touch the repo in $CACHE_DIR, right?

mvollmer:
autogen.sh still calls npm-install.

tools/node-modules
```diff
@@ -0,0 +1,223 @@
...
+        fi
+
+        echo "*** Can't update node modules unless running from git" >&2
+        exit 1
+    fi
+
+    # Otherwise, our main goal is to ensure that the node_modules from
+    # the index is the one that we actually have.
+    local sha="$(get_index_gitlink node_modules)"
+    if [ ! -e node_modules/.git ]; then
```
mvollmer, line 166: Is this always here? I got the impression it might also be in `.git/modules/node_modules`. Or do we sometimes have both?

tools/node-modules
```diff
@@ -0,0 +1,223 @@
...
+***
+***    tools/node-modules rebuild
+***
+*** and add the result to the index.
+EOF
+        exit 1
+    fi
+
+    # Only copy the package-lock.json if it differs from the one we have
+    if ! cmp -s node_modules/.package-lock.json package-lock.json; then
```
mvollmer, line 189: I don't see where `node_modules/.package-lock.json` is actually created... hmm.

allisonkarlitskaya: `npm` creates it.

tools/node-modules
```diff
@@ -0,0 +1,223 @@
...
+        git --git-dir "${CACHE_DIR}" init ${quiet}
+        git --git-dir "${CACHE_DIR}" remote add origin "${CLONE_REMOTE}"
+    fi
+
+    # fetch the tag if we don't already have it
+    if ! git --git-dir "${CACHE_DIR}" rev-parse --verify --quiet "${tag}" >/dev/null; then
+        message FETCH "node_modules  [sha: ${sha}]"
+        # fetch by sha to prevent us from downloading something we don't want
+        # the tag should come automatically with it
+        git --git-dir "${CACHE_DIR}" fetch ${quiet} --tags origin "${sha}"
```
martinpitt, line 50: `--tags` fetches *all* tags, not just the one you want. I run into the same trouble in #15748, which is why I started this with `FETCH_HEAD` (which is of course a bit racy)

martinpitt: As always, [stackoverflow to the rescue](https://stackoverflow.com/questions/45338495/fetch-a-single-tag-from-remote-repository): `git fetch origin tag "$tag" --no-tags`

.github/workflows/node_modules-pull_request.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Read package.json
+        id: package_json
+        uses: actions/github-script@v3
+        with:
+          script: return require('fs').readFileSync('package.json');
+          result-encoding: string
```
martinpitt, line 19: Any reason why this is better than `run: cat package.json` ?

.github/workflows/node_modules-pull_request.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Resolve and verify node_modules gitlink
+        id: verify
+        run: |
+            set -ux
+            node_modules="$(git rev-parse '${{ github.sha }}:node_modules')"
+            test -n "${node_modules}"
+            git fetch "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY%/*}/node-cache" "${node_modules}"
+            git cat-file blob "${node_modules}:.package.json" > node_modules-package.json
+            diff -u node_modules-package.json package.json
```
martinpitt, line 29: Could this re-use tools/node-modules with some test-static-code or similar command?

.github/workflows/node_modules-pull_request.yml
```diff
@@ -0,0 +1,67 @@
...
+  rebuild:
+    needs: check
+    if: ${{ failure() && needs.check.outputs.package_json != '' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Write package.json
+        uses: actions/github-script@v3
+        with:
+          script: require('fs').writeFileSync("package.json", ${{ toJSON(needs.check.outputs.package_json) }});
```
martinpitt, line 40: Similar to above, how is that better than writing it in shell with `echo -n` or `printf`?

.github/workflows/node_modules-publish.yml
```diff
@@ -0,0 +1,111 @@
+name: node_modules publish
+on:
+  workflow_run:
+    workflows: node_modules
+    types: [completed]
+
+jobs:
+  push:
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
```
martinpitt, line 9: FTR, you don't need the `${{` wrapper in `if:`, that's always a condition.

.github/workflows/node_modules-publish.yml
```diff
@@ -0,0 +1,111 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    env:
+      GIT_DIR: git-dir.git
+
+    steps:
+      - name: Debug
+        run: echo '${{ toJSON(github.event) }}'
+
```
martinpitt, line 18: This needs to go from production. I found this echo construct to be very brittle (try it with the full `github` object and watch your workflow explode). A more robust way is to set an env variable to it and echo that, like the [documentation suggests](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-printing-context-information-to-the-log-file). But still, this should not be in production.

martinpitt (changes requested):
Thank you! Some hopefully smaller issues still, plus of course all tests are red because the cache is missing a current build?

.github/workflows/node_modules-publish.yml
allisonkarlitskaya: I think we talked about this in previous review rounds: I still prefer it because of some weird comments in the documentation that it's required for complex expressions.

.github/workflows/node_modules-pull_request.yml
allisonkarlitskaya: Mostly because I know that `toJSON()` is going to get the quoting exactly right for javascript...

.github/workflows/node_modules-pull_request.yml
allisonkarlitskaya: I'm trying to keep the workflow completely independent from any file other than `package.json`.

.github/workflows/node_modules-pull_request.yml
allisonkarlitskaya: The goal here is to get the content of `package.json` into the `.result` variable for the step.

tools/node-modules
```diff
@@ -0,0 +1,215 @@
...
+            echo "Please delete ${file} from ${top_srcdir} and start again"
+            exit 1
+        fi
+    done
+
+    cmd_remove
+
+    npm install
+
+    cp package.json node_modules/.package.json
```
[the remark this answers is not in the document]
allisonkarlitskaya: This is missing.  Some other version of `npm` that I tested against did this, but I discovered after more testing that it's definitely not universal.  That needs to be fixed.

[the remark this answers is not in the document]
allisonkarlitskaya: `--tags` doesn't do what I thought: it fetches *all* tags.  oops!  I thought that it turns on fetching the tags associated with the given commits, which would be the opposite of `--no-tags`:

```
       -n, --no-tags
           By default, tags that point at objects that are downloaded from the remote
           repository are fetched and stored locally. This option disables this automatic
           tag following. The default behavior for a remote may be specified with the
           remote.<name>.tagOpt setting. See git-config(1).
```

Weird that those two aren't opposites.  I'll just remove this.

tools/node-modules
```diff
@@ -0,0 +1,215 @@
...
+
+cmd_checkout() {
+    # we default to check out the node_modules corresponding to the gitlink in the index
+    local sha="${1-$(get_index_gitlink node_modules)}"
+    local tag="sha-${sha}"
+
+    if [ ! -d "${CACHE_DIR}" ]; then
+        message INIT "${CACHE_DIR}"
+        mkdir -p "${CACHE_DIR}"
+        git --git-dir "${CACHE_DIR}" init ${quiet}
```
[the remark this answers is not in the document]
allisonkarlitskaya: Right.

tools/node-modules
```diff
@@ -0,0 +1,215 @@
+#!/bin/sh
+
+set -eu
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/node-cache"
+CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+PUSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Dropping it since we don't expecting pushing from users of this script anymore.  Thanks for the spot.

tools/node-modules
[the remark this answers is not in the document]
allisonkarlitskaya: Also: not sure why I didn't make this `--bare`...

allisonkarlitskaya:
I reworked the tags things by excluding the tags and manually constructing a tag *locally* for the commit that was just fetched.  This gives us some extra protection against tags being created that don't match the given sha, and I actually like it better than what was there before. :)

Seems to be mostly working now.  Not sure what the LGTM story is.....

allisonkarlitskaya:
Okay, cool: LGMT just proved two things:

 - we are really operating completely independently from the `.gitmodules` file (because it was wrong)
 - LGMT does a recursive clone, which actually *does* use this file (as I speculated at some point might be useful for external tools).  nice that we have this to make sure this this approach is also working :)

.github/workflows/node_modules-pull_request.yml
martinpitt: ah, and you don't want to use the `::set-output` stuff, ack.

tools/node-modules
```diff
@@ -132,6 +132,23 @@ EOF
     fi
 }
 
+# called from tools/test-static-code
+cmd_test_static_code() {
+    node_modules_update=$(git log --format=%H -n 1 node_modules)
```
martinpitt, line 137: This commit "tools/test-static-code: add node_modules check" adds the same `cmd_test_static_code()` *again*, it is already present.

martinpitt (commented):
Thanks for the fixes! There's a cosmetical issue left (I'm fine for fixing this in a follow-up). I walked through a manual update again, some things that we may want to improve in the future:

 - `tools/node-modules rebuild` could do the `git add && git commit` with a standard message by itself.
 - That command does not commit to the local cache, it seems to be entirely local.

Tests are apparently on their way to green.

The next (rather urgent) step is then to adjust the npm-update bot to deal with this new approach. In particular we need to either (1) pin down devDependencies so that the weekly PR diffs are actually meaningful (my prefered solution, but we've been through this), or teach it to do updates without corresponding package.json changes. Do you already have some  rough PoC for this? I mean, a set of five shell commands what *should* happen is perfectly okay, it doesn't need to be a ready-made bot update yet.

Thanks!

martinpitt (approved):

allisonkarlitskaya:
> teach it to do updates without corresponding package.json changes. Do you already have some rough PoC for this? I mean, a set of five shell commands what _should_ happen is perfectly okay, it doesn't need to be a ready-made bot update yet.


I was thinking of adding a "please do a `node_modules` update" command or label to force it, for the case where we don't have a `package.json` change.

I also think we might want to discuss the timing of our updates.  I think it might be nice, for stability reasons, if we get our "latest versions" at the start of each cycle, immediately after releasing the previous version.  That gives us the full cycle to shake out any issues that pop up.

I have a vague list of long-term goals enabled by this change, at the top of which is to do away with the -cache tarball and replace it with a tag in the node-cache git repository...

Also: I start to think that maybe we want to never prune the repository.....

martinpitt:
For now I think I'm ok with not updating the cache at all without some explicit package.json change. We'll probably fall behind a bit with babel etc, and changing these will come as a pure surprise instead of an explicit action (I really really *really* want to drop these `^`!), but that doesn't hurt us much short-term. So as a minimum we just need to see that the weekly npm-update PRs will actually work. That's easy enough to test with reverting a recent upgrade, rebuilding the cache, and then letting the bot run locally again.

Issue #14321, opened 2020-07-06
services: Extend service details pages with Firewall information


garrett:
In several usability tests, people have looked at (and filtered) the Services page to find a service to open a port.

Example: "Apache isn't working. Oh, the port is blocked." Goes to Services, finds httpd details page... and is sometimes lost at what to do next.

Possible solution: Show port status Service details.

Something like:
- "Port 443/TCP is open."
- "Port 445/TCP is blocked by the firewall."

And then have a button to block or unblock the port, and a link for viewing all firewall rules (which would then go to the Firewall page).

marusak:
Is the idea here that we would just check if service === `httpd.service` and then show this firewall info?
As there are many web servers implementations.

garrett:
The idea is to _somehow_ know the exposed port(s) for a service, for Cockpit to check and show the status on the service details page, and then allow the user to act upon it somehow.

So, for example: they're at Caddy's service page (which is an HTTP server, but specifically not Apache) and see that it's running but the port is blocked. And then they could unblock it (probably selecting a zone, etc.)... and that would affect the firewall rules.

While you and I know that someone should go to the Network page, then Firewall, then add a rule to allow the port... an overwhelming amount of people did not do it that way, hence this issue asking for some integration of Firewall functionality in the Service details page.

KKoukiou:
For socket activated services we can get this information for ListenStream, in the socket unit file. 
And then we can check if the firewall is running, and if that port is blocked or not. 

So for every service, we can check if it has a socket file as well. If so we can get the port from there and continue with the firewall check. 
It looks doable just too much logic maybe for the services page.

KKoukiou:
The part of exposingthe port a service listens to in the service details was straightforward to implement.
The result is now:

![Screen Shot 2021-04-17 at 17 37 29](https://user-images.githubusercontent.com/14921356/115049090-a009fb80-9eda-11eb-8bf8-a6aec19da9a5.png)

The firewall - service page integration however looks not really possible, in a clean way at least.

I faced the following problems while I was doing some effort to move forward:

* Services IDs in the firewall zones service list are not an exact match with the actual service ids, see ssh in firewall API vs sshd.{service, socket} systemd units
* It's hard to know if a port/service is blocked by firewalld, in firewall page we use the oppossite logic, that means, we add services to zones in order to allow traffic for some service, not block it
[...]

PR #15701, opened 2021-04-15
Hide unimportant details from service details and show some important ones instead

KKoukiou:
So I was thinking how I can solve https://github.com/cockpit-project/cockpit/issues/14321 
Went to service details and I realized that so much information there are not useful for most scenarios, or at least they were never useful for me till now. 
So I though I should hide them by default, and display indead some more useful information, like the port a socket activated service is exposed at, for example.

Having the above mentioned property exposed we can move forward in the future with linking the service details of service sockets to the firewall page. 

What do you think? @marusak ?

With my changes cockpit.socket looks like this:
Screen Shot 2021-04-16 at 18 50 18

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -524,7 +524,6 @@ export class ServiceDetails extends React.Component {
                                         { status }
                                     </DescriptionListDescription>
                                 </DescriptionListGroup>
-                                <hr />
                                 <DescriptionListGroup>
                                     <DescriptionListTerm>{ _("Path") }</DescriptionListTerm>
                                     <DescriptionListDescription id="path">{this.props.unit.FragmentPath}</DescriptionListDescription>
@@ -539,7 +538,6 @@ export class ServiceDetails extends React.Component {
                                         {cockpit.format("$0 ($1)", this.props.unit.Listen[0][1], this.props.unit.Listen[0][0])}
                                     </DescriptionListDescription>
                                 </DescriptionListGroup>}
-                                <hr />
                                 { notMetConditions.length > 0 &&
                                     <DescriptionListGroup>
                                         <DescriptionListTerm className="failed">{ _("Condition failed") }</DescriptionListTerm>
@@ -548,7 +546,6 @@ export class ServiceDetails extends React.Component {
                                         </DescriptionListDescription>
                                     </DescriptionListGroup>
                                 }
-                                <hr />
                                 {triggerRelationshipsList}
                             </DescriptionList>
                             {extraRelationshipsList}
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -245,13 +245,11 @@ export class ServiceDetails extends React.Component {
             clearInterval(this.interval);
     }
 
-    shouldComponentUpdate(nextProps, nextState) {
-        // Don't update when only actions resolved, wait until API triggers some redrawing
-        if ((nextState.waitsAction === false && this.state.waitsAction === true) ||
-            (nextState.waitsFileAction === false && this.state.waitsFileAction === true) ||
-            nextProps.loadingUnits)
-            return false;
-        return true;
+    static getDerivedStateFromProps(nextProps, prevState) {
+        return {
+            waitsAction: nextProps.loadingUnits,
+            waitsFileAction: nextProps.loadingUnits,
+        };
     }
 
     doMemoryCurrentPolling() {
@@ -290,8 +288,7 @@ export class ServiceDetails extends React.Component {
             extra_args = ["fail"];
         this.setState({ waitsAction: true });
         systemd_client.call(SD_OBJ, SD_MANAGER, method, [this.props.unit.Names[0]].concat(extra_args))
-                .catch(error => this.setState({ error: error.toString() }))
-                .finally(() => this.setState({ waitsAction: false }));
+                .catch(error => this.setState({ error: error.toString(), waitsAction: false }));
     }
 
     unitFileAction(method, force) {
@@ -306,8 +303,7 @@ export class ServiceDetails extends React.Component {
                     /* Executing daemon reload after file operations is necessary -
                      * see https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl.c [enable_unit function]
                      */
-                    systemd_client.call(SD_OBJ, SD_MANAGER, "Reload", null)
-                            .then(() => this.setState({ waitsFileAction: false }));
+                    systemd_client.call(SD_OBJ, SD_MANAGER, "Reload", null);
                 })
                 .catch(error => {
                     this.setState({
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -514,14 +514,14 @@ export class ServiceDetails extends React.Component {
                                     <DescriptionListGroup>
                                         <DescriptionListTerm className="failed">{ _("Condition failed") }</DescriptionListTerm>
                                         <DescriptionListDescription id="condition">
-                                            {notMetConditions.map(cond => <div key={cond.split(' ').join('')}>{cond}</div>)}
+                                            {notMetConditions.map(cond => <div key={cond}>{cond}</div>)}
                                         </DescriptionListDescription>
                                     </DescriptionListGroup>
                                 }
                                 <hr />
                                 {relationships.map(rel =>
                                     rel.Units && rel.Units.length > 0 &&
-                                        <DescriptionListGroup key={rel.Name.split().join("")}>
+                                        <DescriptionListGroup key={rel.Name}>
                                             <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
                                             <DescriptionListDescription id={rel.Name.split(" ").join("")}>
                                                 <ul className="comma-list">
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -24,6 +24,7 @@ import {
     Alert, Button,
     DescriptionList, DescriptionListTerm, DescriptionListGroup, DescriptionListDescription,
     Dropdown, DropdownItem, DropdownSeparator, KebabToggle,
+    ExpandableSection,
     Tooltip, TooltipPosition,
     Card, CardBody, CardTitle, Text, TextVariants,
     Modal, Switch
@@ -425,6 +426,13 @@ export class ServiceDetails extends React.Component {
         const tooltipMessage = enabled ? _("Stop and disable") : _("Start and enable");
         const hasLoadError = this.props.unit.LoadState !== "loaded" && this.props.unit.LoadState !== "masked";
         const loadError = this.props.unit.LoadError ? this.props.unit.LoadError[1] : null;
+
+        // These are relevant for socket and timer activated services
+        const triggerRelationships = [
+            { Name: _("Triggers"), Units: this.props.unit.Triggers },
+            { Name: _("Triggered by"), Units: this.props.unit.TriggeredBy },
+        ];
+
         const relationships = [
             { Name: _("Requires"), Units: this.props.unit.Requires },
             { Name: _("Requisite"), Units: this.props.unit.Requisite },
@@ -441,12 +449,34 @@ export class ServiceDetails extends React.Component {
             { Name: _("Before"), Units: this.props.unit.Before },
             { Name: _("After"), Units: this.props.unit.After },
             { Name: _("On failure"), Units: this.props.unit.OnFailure },
-            { Name: _("Triggers"), Units: this.props.unit.Triggers },
-            { Name: _("Triggered by"), Units: this.props.unit.TriggeredBy },
             { Name: _("Propagates reload to"), Units: this.props.unit.PropagatesReloadTo },
             { Name: _("Reload propagated from"), Units: this.props.unit.ReloadPropagatedFrom },
             { Name: _("Joins namespace of"), Units: this.props.unit.JoinsNamespaceOf }
         ];
+
+        const relationshipsToList = rels => {
+            return rels.filter(rel => rel.Units && rel.Units.length > 0)
+                    .map(rel =>
+                        <DescriptionListGroup key={rel.Name}>
+                            <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
+                            <DescriptionListDescription id={rel.Name.split(" ").join("")}>
+                                <ul className="comma-list">
+                                    {rel.Units.map(unit => <li key={unit}><a href={"#/" + unit} className={this.props.isValid(unit) ? "" : "disabled"}>{unit}</a></li>)}
+                                </ul>
+                            </DescriptionListDescription>
+                        </DescriptionListGroup>
+                    );
+        };
+
+        const triggerRelationshipsList = relationshipsToList(triggerRelationships);
+
+        const extraRelationshipsList = (
+            <ExpandableSection id="service-details-show-relationships" toggleText={triggerRelationshipsList.length ? _("Show more relationships") : _("Show relationships")}>
+                <DescriptionList>
+                    {relationshipsToList(relationships)}
+                </DescriptionList>
+            </ExpandableSection>
+        );
 
         const conditions = this.props.unit.Conditions;
         const notMetConditions = [];
@@ -519,18 +549,9 @@ export class ServiceDetails extends React.Component {
                                     </DescriptionListGroup>
                                 }
                                 <hr />
-                                {relationships.map(rel =>
-                                    rel.Units && rel.Units.length > 0 &&
-                                        <DescriptionListGroup key={rel.Name}>
-                                            <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
-                                            <DescriptionListDescription id={rel.Name.split(" ").join("")}>
-                                                <ul className="comma-list">
-                                                    {rel.Units.map(unit => <li key={unit}><a href={"#/" + unit} className={this.props.isValid(unit) ? "" : "disabled"}>{unit}</a></li>)}
-                                                </ul>
-                                            </DescriptionListDescription>
-                                        </DescriptionListGroup>
-                                )}
+                                {triggerRelationshipsList}
                             </DescriptionList>
+                            {extraRelationshipsList}
                         </CardBody>
                     </>
                 }
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -140,7 +140,7 @@ class ServiceActions extends React.Component {
 
             if (actions.length > 0) {
                 actions.push(
-                    <DropdownSeparator key="divider" divider />
+                    <DropdownSeparator key="divider" />
                 );
             }
 
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -472,7 +472,7 @@ export class ServiceDetails extends React.Component {
 
         const extraRelationshipsList = (
             <ExpandableSection id="service-details-show-relationships" toggleText={triggerRelationshipsList.length ? _("Show more relationships") : _("Show relationships")}>
-                <DescriptionList>
+                <DescriptionList isHorizontal>
                     {relationshipsToList(relationships)}
                 </DescriptionList>
             </ExpandableSection>
@@ -517,7 +517,7 @@ export class ServiceDetails extends React.Component {
                             }
                         </CardTitle>
                         <CardBody>
-                            <DescriptionList>
+                            <DescriptionList isHorizontal>
                                 <DescriptionListGroup>
                                     <DescriptionListTerm>{ _("Status") }</DescriptionListTerm>
                                     <DescriptionListDescription id="statuses">
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -457,7 +457,7 @@ export class ServiceDetails extends React.Component {
                             <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
                             <DescriptionListDescription id={rel.Name.split(" ").join("")}>
                                 <ul className="comma-list">
-                                    {rel.Units.map(unit => <li key={unit}><a href={"#/" + unit} className={this.props.isValid(unit) ? "" : "disabled"}>{unit}</a></li>)}
+                                    {rel.Units.map(unit => <li key={unit}><Button isInline variant="link" component="a" href={"#/" + unit} isDisabled={!this.props.isValid(unit)}>{unit}</Button></li>)}
                                 </ul>
                             </DescriptionListDescription>
                         </DescriptionListGroup>
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -507,6 +507,7 @@ export class ServiceDetails extends React.Component {
                                         <Tooltip id="switch-unit-state" content={tooltipMessage} position={TooltipPosition.right}>
                                             <span>
                                                 <Switch isChecked={enabled}
+                                                        aria-label={tooltipMessage}
                                                         isDisabled={this.state.waitsAction || this.state.waitsFileAction}
                                                         onChange={this.onOnOffSwitch} />
                                             </span>
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -35,6 +35,7 @@ import { systemd_client, SD_MANAGER, SD_OBJ } from "./services.jsx";
 import './service-details.scss';
 
 const _ = cockpit.gettext;
+const METRICS_POLL_DELAY = 30000; // 30s
 
 /*
  * React template for showing basic dialog for confirming action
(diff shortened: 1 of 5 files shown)


garrett:
Ooooo, I :heart: this! You've got my approval for this PR when it's time for it.

Great work!

pkg/systemd/services/service-details.jsx
```diff
@@ -452,6 +452,16 @@ export class ServiceDetails extends React.Component {
                                     <DescriptionListTerm>{ _("Path") }</DescriptionListTerm>
                                     <DescriptionListDescription id="path">{this.props.unit.FragmentPath}</DescriptionListDescription>
                                 </DescriptionListGroup>
+                                {this.props.unit.MemoryCurrent && <DescriptionListGroup>
```
marusak, line 486: Is this how much it currently consumes? If so, is it dynamical changing over time?

pkg/systemd/services/service-details.jsx
```diff
@@ -494,16 +493,14 @@ export class ServiceDetails extends React.Component {
                                         {cockpit.format("$0 ($1)", this.props.unit.Listen[0][1], this.props.unit.Listen[0][0])}
                                     </DescriptionListDescription>
                                 </DescriptionListGroup>}
-                                <hr />
```
marusak, line 455: This commit makes previous one about now showing this item obsolete.

marusak:
I like it a lot! Especially since this is the whole screen on my laptop screen:
screenbefore
And this is with this PR:
afterrel

The space is used just soo much better. Also having `Listen` is very useful.

It shows some random `0` though, can you please take a look at that?
justzero


Unrelated but noticed that there is not bottom padding, I cannot scroll any further than:
anyfrther
It is something we should take a look at.

pkg/systemd/services/services.jsx
```diff
@@ -607,6 +631,18 @@ class ServicesPage extends React.Component {
             }
         }
 
+        if (unitNew.Id.slice(-6) == "socket") {
+            unitNew.is_timer = true;
```
martinpitt, line 635: This looks like a copy&paste error.

martinpitt (changes requested):
So cool, thanks @KKoukiou !

pkg/systemd/services/service-details.jsx
KKoukiou: It is but it's also not updated asynchronously right now, let me see how I can get an event for that.

Very nice catch ;)

KKoukiou: So this seems a bit hard, according to systemd documentation this property never emits a signal when changed. https://man7.org/linux/man-pages/man5/org.freedesktop.systemd1.5.html

martinpitt: Then I suppose it could update itself every 30s or so through a timer?

pkg/systemd/services/services.jsx
KKoukiou: Fixed.

pkg/systemd/services/service-details.jsx
KKoukiou: Removed.

pkg/systemd/services/service-details.jsx
KKoukiou: Ok polling it is.

pkg/systemd/services/service-details.jsx
```diff
@@ -208,11 +208,40 @@ export class ServiceDetails extends React.Component {
             waitsFileAction: false,
             note: "",
             error: "",
+            delay: 30000, // 30s
```
martinpitt, line 212: This is a bit weird as state -- should rather be a const above the class? `METRICS_POLL_DELAY` or so?

pkg/systemd/services/service-details.jsx
```diff
@@ -208,11 +208,40 @@ export class ServiceDetails extends React.Component {
...
+            unit_properties: {},
         };
 
         this.onOnOffSwitch = this.onOnOffSwitch.bind(this);
         this.unitAction = this.unitAction.bind(this);
         this.unitFileAction = this.unitFileAction.bind(this);
+        this.addUnitProperties = this.addUnitProperties.bind(this);
+        this.doPropertyPolling = this.doPropertyPolling.bind(this);
```
martinpitt, line 220: Not necessary -- these are always called directly, not as a "function pointer" to component properties.

pkg/systemd/services/service-details.jsx
```diff
@@ -208,11 +208,40 @@ export class ServiceDetails extends React.Component {
...
+
+        this.unitType = props.unit.Id.split('.').slice(-1)[0];
+        this.unitTypeCapitalized = this.unitType.charAt(0).toUpperCase() + this.unitType.slice(1);
+
+        if (props.unit.ActiveState == "active") {
+            this.doPropertyPolling();
+            this.interval = setInterval(this.doPropertyPolling, this.state.delay);
+        }
```
martinpitt, line 228: Does this really have to repeat the code in componentDidUpdate()? I thought the latter would be called on initial mounting as well?

pkg/systemd/services/service-details.jsx
```diff
@@ -208,11 +208,40 @@ export class ServiceDetails extends React.Component {
...
+    }
+
+    componentDidUpdate(prevProps) {
+        // If unit became active start property polling and if got inactive stop
+        if (this.props.unit.ActiveState === 'active' && !this.interval) {
+            this.doPropertyPolling();
+            clearInterval(this.interval);
```
martinpitt, line 235: At this point this.interval is false-y, so this isn't necessary (apparently it does not crash, though)

pkg/systemd/services/service-details.jsx
```diff
@@ -208,11 +208,40 @@ export class ServiceDetails extends React.Component {
...
+            this.interval = setInterval(this.doPropertyPolling, this.state.delay);
+        }
+        if (this.props.unit.ActiveState === 'inactive' && this.interval) {
+            this.doPropertyPolling();
+            clearInterval(this.interval);
+        }
+    }
+
+    componentWillUnmount() {
+        clearInterval(this.interval);
```
martinpitt, line 245: Does this cause any error message when this.interval is invalid? To be on the safe side, this could explicitly check.

pkg/systemd/services/service-details.jsx
```diff
@@ -224,6 +253,29 @@ export class ServiceDetails extends React.Component {
         return true;
     }
 
+    doPropertyPolling() {
+        systemd_client.call(this.props.unit.path,
+                            "org.freedesktop.DBus.Properties", "GetAll",
+                            [cockpit.format("org.freedesktop.systemd1.$0", this.unitTypeCapitalized)])
```
martinpitt, line 258: cockpit.format is a bit overkill for this -- simple `+` will do.

pkg/systemd/services/service-details.jsx
```diff
@@ -224,6 +253,29 @@ export class ServiceDetails extends React.Component {
...
+                .then(result => {
+                    this.addUnitProperties({
+                        MemoryCurrent: (result[0].MemoryCurrent && result[0].MemoryCurrent.v > 0) ? result[0].MemoryCurrent.v : null,
+                        Listen: result[0].Listen ? result[0].Listen.v : null
```
martinpitt, line 262: It makes sense that `MemoryCurrent` does not generate PropertiesChanged signals, but I'm a bit surprised that changes to `Listen` don't? Is it even realistic that `Listen` changes *while* a service is active? If we only need to poll for MemoryCurrent, then GetAll could be reduced to a much cheaper direct Get.

pkg/systemd/services/service-details.jsx
```diff
@@ -224,6 +253,29 @@ export class ServiceDetails extends React.Component {
...
+                    });
+                });
```
martinpitt, line 264: Let's please handle errors from GetAll here (with a simple console.log) to avoid a page Oops in weird cases.

pkg/systemd/services/service-details.jsx
```diff
@@ -224,6 +253,29 @@ export class ServiceDetails extends React.Component {
...
+    }
+
+    addUnitProperties(unit) {
+        let needsUpdate = false;
+
+        if (this.state.unit_properties.Listen !== unit.Listen ||
+            this.state.unit_properties.MemoryCurrent !== unit.MemoryCurrent)
+            needsUpdate = true;
```
martinpitt, line 272: This feels a bit too verbose, but harmless enough.

test/verify/check-system-services
```diff
@@ -269,6 +269,10 @@ Unit=test.service
         b.enter_page("/system/services")
         b.wait_js_cond('window.location.hash === "#/test.service"')
 
+        # Check that current memory is shown
+        b.wait_visible("#memory")
+        self.assertGreater(int(b.text("#memory").split(" ")[0]), 0)
```
martinpitt, line 274: Please 
 - check that there is no Listen
 - stopping the unit makes Memory go away, starting it again brings it back
 - ideally it should also test that the regular update works. Schema: Let test.service wait for a /tmp/do-memhog file, then have it allocate a big chunk (see check-metrics for existing code); the test then gets to this point, creates do-memhog, and ensures that the shown value actually increases (it should take at most 30s, so our standard timeouts seem fine)

test/verify/check-system-services
```diff
@@ -329,6 +333,15 @@ Unit=test.service
 
         b.go("#/")
 
+        # Check that listen property is shown for socket units
+        self.pick_tab(3)
+        b.click(self.svc_sel("dbus.socket"))
+        b.wait_visible("#listen")
+        b.click(".pf-c-breadcrumb a:contains('Services')")
```
martinpitt, line 340: ... and does not show Memory

pkg/systemd/services/service-details.jsx
```diff
@@ -447,13 +455,37 @@ export class ServiceDetails extends React.Component {
...
-            { Name: _("Triggered by"), Units: this.props.unit.TriggeredBy },
             { Name: _("Propagates reload to"), Units: this.props.unit.PropagatesReloadTo },
             { Name: _("Reload propagated from"), Units: this.props.unit.ReloadPropagatedFrom },
             { Name: _("Joins namespace of"), Units: this.props.unit.JoinsNamespaceOf }
         ];
 
+        const relationshipsToList = rels => {
+            return rels.filter(rel => rel.Units && rel.Units.length > 0)
+                    .map(rel => {
+                        return (
```
martinpitt, line 462: Style nitpick: replace `=> { return (..) }` with a more direct `=> (..)` (this is what lambda functions are really meant to be..)

pkg/systemd/services/service-details.jsx
```diff
@@ -447,13 +455,37 @@ export class ServiceDetails extends React.Component {
...
+                            <DescriptionListGroup key={rel.Name.split().join("")}>
```
martinpitt, line 463: why the split/join dance? keys can contain spaces

pkg/systemd/services/service-details.jsx
```diff
@@ -447,13 +455,37 @@ export class ServiceDetails extends React.Component {
...
+                                <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
+                                <DescriptionListDescription id={rel.Name.split(" ").join("")}>
+                                    <ul className="comma-list">
+                                        {rel.Units.map(unit => <li key={unit}><a href={"#/" + unit} className={this.props.isValid(unit) ? "" : "disabled"}>{unit}</a></li>)}
```
martinpitt, line 467: I thought we moved away from `disabled` classes and use the proper `disabled` attribute. I know this is just moving existing code, but it is now the *only* place in cockpit that uses a "disabled" class. If there's a good reason for that, we can look at that in a follow-up, otherwise this may be a good time to clean this up?

test/verify/check-system-services
```diff
@@ -530,28 +530,33 @@ ExecStart=/bin/sh -c "while true; do sleep 5; done"
 
         # service a
         b.wait_js_cond('window.location.hash === "#/test-a.service"')
+        b.click("#service-details-show-relationships  button")
```
martinpitt, line 533: double space (Sanne would be *so* proud! :grin:)

test/verify/check-system-services
```diff
@@ -530,28 +530,33 @@ ExecStart=/bin/sh -c "while true; do sleep 5; done"
...
         b.wait_visible(rel_sel("Before", "test-b.service"))
         b.wait_visible(rel_sel("Conflicts", "test-c.service"))
         b.click(rel_sel("Before", "test-b.service"))
 
         # service b
         b.wait_js_cond('window.location.hash === "#/test-b.service"')
+        b.click("#service-details-show-relationships  button")
```
martinpitt, line 540: Do these tests cover the trigger relationships somewhere, to make sure they work as intended?

martinpitt (changes requested):
This works and looks really well :clap: Many thanks for this!

pkg/systemd/services/service-details.jsx
martinpitt: In other words, the Listen property is defined by the unit's `ListenStream=` definition. So changing that would require a daemon-reload and a service restart.

pkg/systemd/services/service-details.jsx
KKoukiou: componentDidUpdate method is not called for the initial render. So when we show the page first time we need to start the polling. Then everytime the page updates, this might mean that the unit changed state, so we might need to start or stop the polling.

martinpitt: Ack, thanks; I suppose I was thinking of `useEffect()` as the thing that runs both on initial mount and on update. (That may still be the more elegant way, but let's not block on this).

test/verify/check-system-services
KKoukiou: There are some relationships test, but I will check the trigger relationships explicitly in addition.

pkg/systemd/services/service-details.jsx
KKoukiou: Done

pkg/systemd/services/service-details.jsx
KKoukiou: Removed.

pkg/systemd/services/service-details.jsx
KKoukiou: Seems, not but guarded it anyway.

pkg/systemd/services/service-details.jsx
KKoukiou: It was there before, but it's indeed useless.

pkg/systemd/services/service-details.jsx
KKoukiou: Sure I will add a commit on top/

test/verify/check-system-services
KKoukiou: Fixed

pkg/systemd/services/service-details.jsx
KKoukiou: Fixed

pkg/systemd/services/service-details.jsx
```diff
@@ -208,11 +209,71 @@ export class ServiceDetails extends React.Component {
...
+        // Subscribe to PropertiesChanged - follow changes on Listen property
+        if (this.unitType == 'socket') {
+            systemd_client.subscribe(
+                {
+                    path: props.unit.path,
+                    interface: "org.freedesktop.DBus.Properties",
+                    member: "PropertiesChanged"
+                },
+                (path, iface, signal, args) => {
+                    if (args[1].Listen) {
```
martinpitt, line 233: It still amazes me how a socket unit could change its `ListenStream` property while it's running. Did you observe that in practice? How would one do that?

pkg/systemd/services/service-details.jsx
```diff
@@ -224,6 +285,31 @@ export class ServiceDetails extends React.Component {
         return true;
     }
 
+    doMemoryCurrentPolling() {
+        systemd_client.call(this.props.unit.path,
+                            "org.freedesktop.DBus.Properties", "Get",
+                            [cockpit.format("org.freedesktop.systemd1.$0", this.unitTypeCapitalized), 'MemoryCurrent'])
```
martinpitt, line 290: (copying from previous review round, still relevant) cockpit.format is a bit overkill for this -- simple + will do.

pkg/systemd/services/service-details.jsx
```diff
@@ -561,7 +561,7 @@ export class ServiceDetails extends React.Component {
                                 <hr />
                                 {relationships.map(rel =>
                                     rel.Units && rel.Units.length > 0 &&
-                                        <DescriptionListGroup key={rel.Name.split().join("")}>
+                                        <DescriptionListGroup key={rel.Name.split(" ").join("")}>
```
martinpitt: but... why? What is wrong with `key={rel.Name}`? The spaces don't hurt at all in a key. And as split() is a no-op here, it obviously worked before :grin:

martinpitt (changes requested):
Thanks! There are still quite a number of issues from the previous review round, I marked the others as resolved. Many of the middle ones are unhelpfully hidden by github, you have to expand them :crying_cat_face:

pkg/systemd/services/service-details.jsx
KKoukiou: No I did not, I just saw it in the API documentation that it emits signal when changed. Also in your previous review I thought you wanted me to write a event listener to follow changes for this property, but I guess I misunderstood the comment. Shall I just remove it then? I also feel it's useless to have an even listener for this.

pkg/systemd/services/service-details.jsx
KKoukiou: This is implemented now: services: replace custom disabled class in service relationships comma list

test/verify/check-system-services
KKoukiou: there is already a test for it: see string:

self.assertIn(b.text("#Triggers .pf-c-description-list__text"), ["dbus.service", "dbus-broker.service"])

pkg/systemd/services/service-details.jsx
KKoukiou: So I can remove the event listener I have, but in that case, I still need to call the Get API for Listen in componentDidUpdate or else move the fetching of Listen in the main services.jsx file, which actually refetches some unit properties on daemon reload or unit events. I would rather do the later.

pkg/systemd/services/service-details.jsx
martinpitt: Sorry for the misunderstanding! Yes, please remove it all; it can't change while a .socket unit is running.

test/verify/check-system-services
KKoukiou: Socket units also have CurrentMemory shown in the UI

martinpitt: Oh wow -- that shouldn't ever be something else than "trivial" :grin:

KKoukiou: Indeed. I could also hide it, but systemctl status shows it as well, so..

KKoukiou:
Still some test failures, marking as draft till these are resolved.

pkg/systemd/services/service-details.jsx
KKoukiou: So I moved the 'LIsten' property fetching one component up, because where it was we would not receive updates even if the unit restarted.

martinpitt:
test-static-code [fails](https://github.com/cockpit-project/cockpit/pull/15701/checks?check_run_id=2417239742):
```
test/verify/check-system-services:22: 'time' imported but unused
test/verify/check-system-services:810: redefinition of unused 'time' from line 22
```

KKoukiou:
> test-static-code [fails](https://github.com/cockpit-project/cockpit/pull/15701/checks?check_run_id=2417239742):
> 
> ```
> test/verify/check-system-services:22: 'time' imported but unused
> test/verify/check-system-services:810: redefinition of unused 'time' from line 22
> ```

Ups sorry, reposted.

pkg/systemd/services/services.jsx
```diff
@@ -596,6 +615,17 @@ class ServicesPage extends React.Component {
 
         this.updateComputedProperties(unitNew);
 
+        if (unitNew.Id.slice(-6) == "socket") {
```
martinpitt, line 618: `.endsWith("socket")` works without magic numbers.

pkg/systemd/services/service-details.jsx
```diff
@@ -224,6 +253,23 @@ export class ServiceDetails extends React.Component {
...
+    doMemoryCurrentPolling() {
+        systemd_client.call(this.props.unit.path,
+                            "org.freedesktop.DBus.Properties", "Get",
+                            ["org.freedesktop.systemd1." + this.unitTypeCapitalized, 'MemoryCurrent'])
+                .then(result => {
+                    this.addUnitProperties(
+                        "MemoryCurrent",
+                        result[0] && result[0].v > 0 ? result[0].v : null,
+                    );
+                }, ex => console.log("Failed to get MemoryCurrent unit property for unit " + this.props.unit.Id));
```
martinpitt, line 264: Would be nice to log the actual exception.

test/verify/check-system-services
```diff
@@ -329,6 +329,15 @@ Unit=test.service
 
         b.go("#/")
 
+        # Check that listen property is shown for socket units
+        self.pick_tab(3)
+        b.click(self.svc_sel("dbus.socket"))
+        b.wait_visible("#listen")
```
martinpitt: `wait_in_text("#listen", "/run/dbus/")` ? to make sure it doesn't show something empty or bogus?

test/verify/check-system-services
```diff
@@ -382,6 +391,39 @@ Unit=test.service
...
+        self.write_file("/etc/systemd/system/mem-hog.service",
+                        """
+[Unit]
+Description=Mem Hog Service
+
+[Service]
+ExecStart=/bin/awk -f /tmp/mem-hog.awk
+
+[Install]
+WantedBy=default.target
```
martinpitt, line 407: No need for this bit

test/verify/check-system-services
```diff
@@ -382,6 +391,39 @@ Unit=test.service
...
+
+        # Check that MemoryCurrent is shown
+        b.click(self.svc_sel("mem-hog.service"))
+        b.wait_not_present("#memory")
+        m.execute("systemctl start mem-hog.service")
+        initial_memory = float(b.text("#memory .pf-c-description-list__text").split(" ")[0])
+        self.assertGreater(initial_memory, 1)
+        m.execute("touch /tmp/continue")
+        # MemoryCurrent auto-updates every 30s - when it updates the memory used should be ~50MiB
+        with b.wait_timeout(30):
```
martinpitt, line 421: That would fail if the JS code's interval timer happened to start at the same time as the test -- which is actually fairly likely. Let's just use the standard timeout here.

martinpitt (changes requested):
This works really nicely now :+1: Also thanks for the nice dynamic memory test case!

A few remaining easy nitpicks.

test/verify/check-system-services
KKoukiou: I have tihs self.assertIn("Stream", b.text("#listen .pf-c-description-list__text")), what you commented on seems outdated version.

KKoukiou: I can add the whole thing, /run/dbus (Stream) though, I was just trying to avoid distribution specific differences.

martinpitt: I reviewed the commits one by one indeed; `/run/` isn't distribution specific, that should be perfectly safe. "dbus" certainly ought to appear somewhere as well.

pkg/systemd/services/service-details.jsx
```diff
@@ -208,20 +210,63 @@ export class ServiceDetails extends React.Component {
...
+    doMemoryCurrentPolling() {
+        systemd_client.call(this.props.unit.path,
+                            "org.freedesktop.DBus.Properties", "Get",
+                            ["org.freedesktop.systemd1." + this.unitTypeCapitalized, 'MemoryCurrent'])
+                .then(result => {
+                    this.addUnitProperties(
+                        "MemoryCurrent",
+                        result[0] && result[0].v > 0 ? result[0].v : null,
+                    );
+                }, ex => console.log(ex.message));
```
martinpitt, line 264: FWIW, the previous "Failed to.." string was fine, I meant to add `ex.toString()` as a second argument. But good enough.

martinpitt (approved):
Thanks! :+1: assuming the tests agree.

martinpitt:
@marusak: [rhel-9-0](https://logs.cockpit-project.org/logs/pull-15701-20210423-124323-ceca3b95-rhel-9-0/log.html) failed in an interesting way:
```
+ rpm -U --force build-results/cockpit-debuginfo-242.63.g64ad0d211-1.wip.el9.x86_64.rpm build-results/cockpit-system-242.63.g64ad0d211-1.wip.el9.noarch.rpm build-results/cockpit-debugsource-242.63.g64ad0d211-1.wip.el9.x86_64.rpm build-results/cockpit-ws-242.63.g64ad0d211-1.wip.el9.x86_64.rpm build-results/cockpit-tests-242.63.g64ad0d211-1.wip.el9.x86_64.rpm build-results/cockpit-bridge-242.63.g64ad0d211-1.wip.el9.x86_64.rpm build-results/cockpit-packagekit-242.63.g64ad0d211-1.wip.el9.noarch.rpm build-results/cockpit-storaged-242.63.g64ad0d211-1.wip.el9.noarch.rpm build-results/cockpit-pcp-242.63.g64ad0d211-1.wip.el9.x86_64.rpm build-results/cockpit-242.63.g64ad0d211-1.wip.el9.x86_64.rpm
error: Failed dependencies:
	(selinux-policy >= 34.1.2 if selinux-policy-targeted) is needed by cockpit-ws-242.63.g64ad0d211-1.wip.el9.x86_64
```

This is some weird collision with your PR? that just landed

martinpitt:
Or perhaps some interaction between https://github.com/cockpit-project/cockpit/pull/15707 and the image refresh in https://github.com/cockpit-project/bots/pull/1955 ? They landed in parallel

martinpitt:
@garrett: I made new screenshots (the original one was way too wide) and added a release note. Can you please have a look?

garrett:
@martinpitt Cool!

I do remember a screenshot showing a port; these don't though?

garrett:
Here's a cropped screenshot from the above thread that shows a port:

114979796-c81e3e00-9e8b-11eb-9b3f-681fcb9035cb-3

martinpitt:
@garrett Right, I picked an Unix socket instead of a TCP one, but a TCP socket is fine as well. I just didn't want to use Cockpit's own service, as that felt a little too self-referential. But fine for me, of course :) Do you want to do the corresponding service one as well, or want me to re-do mine?

PR #15704, opened 2021-04-16
test: Drop obsolete expected message filters

martinpitt:


martinpitt:
One [unexpected message on fedora-33](https://logs.cockpit-project.org/logs/pull-15704-20210416-051103-6611fa08-fedora-33/log.html#189):

    pcp-archive: couldn't create pcp archive context for /var/log/pcp/pmlogger/localhost/20210416.05.20: Permission denied (-13)

Curiously that's not one affected by this PR, but I'll add that to the existing pcp list as a separate entry.

KKoukiou (approved):

PR #15707, opened 2021-04-16
Add custom selinux policy

marusak:
This policy will be shipped in cockpit-ws package on rhel-9 onward and fedora-33 onward.
This policy is taken from Fedora rawhide.

We don't plan to backport to rhel-8 and centos-8. Debian/suse may come later.

### DIFF
--- a/selinux/cockpit.fc
+++ b/selinux/cockpit.fc
@@ -1,0 +1,16 @@
+/usr/lib/systemd/system/cockpit.*		--	gen_context(system_u:object_r:cockpit_unit_file_t,s0)
+/etc/systemd/system/cockpit.*	--	gen_context(system_u:object_r:cockpit_unit_file_t,s0)
+
+/usr/libexec/cockpit-ws		--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-tls	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-wsinstance-factory	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+
+/usr/libexec/cockpit-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+/usr/libexec/cockpit-ssh	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+
+/usr/share/cockpit/motd/update-motd    -- gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+
+/var/lib/cockpit(/.*)?      gen_context(system_u:object_r:cockpit_var_lib_t,s0)
+
+/var/run/cockpit(/.*)?   gen_context(system_u:object_r:cockpit_var_run_t,s0)
+/var/run/cockpit-ws(/.*)?   gen_context(system_u:object_r:cockpit_var_run_t,s0)
--- a/selinux/cockpit_session_selinux.8
+++ b/selinux/cockpit_session_selinux.8
@@ -1,0 +1,255 @@
+.TH  "cockpit_session_selinux"  "8"  "21-04-16" "cockpit_session" "SELinux Policy cockpit_session"
+.SH "NAME"
+cockpit_session_selinux \- Security Enhanced Linux Policy for the cockpit_session processes
+.SH "DESCRIPTION"
+
+Security-Enhanced Linux secures the cockpit_session processes via flexible mandatory access control.
+
+The cockpit_session processes execute with the cockpit_session_t SELinux type. You can check if you have these processes running by executing the \fBps\fP command with the \fB\-Z\fP qualifier.
+
+For example:
+
+.B ps -eZ | grep cockpit_session_t
+
+
+.SH "ENTRYPOINTS"
+
+The cockpit_session_t SELinux type can be entered via the \fBcockpit_session_exec_t\fP file type.
+
+The default entrypoint paths for the cockpit_session_t domain are the following:
+
+/usr/libexec/cockpit-ssh, /usr/libexec/cockpit-session
+.SH PROCESS TYPES
+SELinux defines process types (domains) for each process running on the system
+.PP
+You can see the context of a process using the \fB\-Z\fP option to \fBps\bP
+.PP
+Policy governs the access confined processes have to files.
+SELinux cockpit_session policy is very flexible allowing users to setup their cockpit_session processes in as secure a method as possible.
+.PP
+The following process types are defined for cockpit_session:
+
+.EX
+.B cockpit_session_t
+.EE
+.PP
+Note:
+.B semanage permissive -a cockpit_session_t
+can be used to make the process type cockpit_session_t permissive. SELinux does not deny access to permissive process types, but the AVC (SELinux denials) messages are still generated.
+
+.SH BOOLEANS
+SELinux policy is customizable based on least access required.  cockpit_session policy is extremely flexible and has several booleans that allow you to manipulate the policy and run cockpit_session with the tightest access possible.
+
+
+.PP
+If you want to allow all domains to execute in fips_mode, you must turn on the fips_mode boolean. Enabled by default.
+
+.EX
+.B setsebool -P fips_mode 1
+
+.EE
+
+.PP
+If you want to allow confined applications to run with kerberos, you must turn on the kerberos_enabled boolean. Enabled by default.
+
+.EX
+.B setsebool -P kerberos_enabled 1
+
+.EE
+
+.PP
+If you want to allow system to run with NIS, you must turn on the nis_enabled boolean. Disabled by default.
+
+.EX
+.B setsebool -P nis_enabled 1
+
+.EE
+
+.PP
+If you want to enable polyinstantiated directory support, you must turn on the polyinstantiation_enabled boolean. Disabled by default.
+
+.EX
+.B setsebool -P polyinstantiation_enabled 1
+
+.EE
+
+.SH "MANAGED FILES"
+
+The SELinux process type cockpit_session_t can manage files labeled with the following file types.  The paths listed are the default paths for these file types.  Note the processes UID still need to have DAC permissions.
+
+.br
+.B auth_cache_t
+
+	/var/cache/coolkey(/.*)?
+.br
+
+.br
+.B auth_home_t
+
+	/root/\.yubico(/.*)?
+.br
+	/root/\.config/Yubico(/.*)?
+.br
+	/root/\.google_authenticator
+.br
+	/root/\.google_authenticator~
+.br
+	/home/[^/]+/\.yubico(/.*)?
+.br
+	/home/[^/]+/\.config/Yubico(/.*)?
+.br
+	/home/[^/]+/\.google_authenticator
+.br
+	/home/[^/]+/\.google_authenticator~
+.br
+
+.br
+.B faillog_t
+
+	/var/log/btmp.*
+.br
+	/var/log/faillog.*
+.br
+	/var/log/tallylog.*
+.br
+	/var/run/faillock(/.*)?
+.br
+
+.br
+.B initrc_var_run_t
+
+	/var/run/utmp
+.br
+	/var/run/random-seed
+.br
+	/var/run/runlevel\.dir
+.br
+	/var/run/setmixer_flag
+.br
+
+.br
+.B lastlog_t
+
+	/var/log/lastlog.*
+.br
+
+.br
+.B pam_var_run_t
+
+	/var/(db|adm)/sudo(/.*)?
+.br
+	/var/lib/sudo(/.*)?
+.br
+	/var/run/sudo(/.*)?
+.br
+	/var/run/pam_ssh(/.*)?
+.br
+	/var/run/sepermit(/.*)?
+.br
+	/var/run/pam_mount(/.*)?
+.br
+	/var/run/pam_timestamp(/.*)?
+.br
+
+.br
+.B security_t
+
+	/selinux
+.br
+
+.br
+.B shadow_t
+
+	/etc/shadow.*
+.br
+	/etc/gshadow.*
+.br
+	/etc/nshadow.*
+.br
+	/var/db/shadow.*
+.br
+	/etc/security/opasswd
+.br
+	/etc/security/opasswd\.old
+.br
+
+.br
+.B var_auth_t
+
+	/var/ace(/.*)?
+.br
+	/var/rsa(/.*)?
+.br
+	/var/lib/abl(/.*)?
+.br
+	/var/lib/rsa(/.*)?
+.br
+	/var/lib/pam_ssh(/.*)?
+.br
+	/var/lib/pam_shield(/.*)?
+.br
+	/var/opt/quest/vas/vasd(/.*)?
+.br
+	/var/lib/google-authenticator(/.*)?
+.br
+
+.br
+.B wtmp_t
+
+	/var/log/wtmp.*
+.br
+
+.SH FILE CONTEXTS
+SELinux requires files to have an extended attribute to define the file type.
+.PP
+You can see the context of a file using the \fB\-Z\fP option to \fBls\bP
+.PP
+Policy governs the access confined processes have to these files.
+SELinux cockpit_session policy is very flexible allowing users to setup their cockpit_session processes in as secure a method as possible.
+.PP
+
+.I The following file types are defined for cockpit_session:
+
+
+.EX
+.PP
+.B cockpit_session_exec_t
+.EE
+
+- Set files with the cockpit_session_exec_t type, if you want to transition an executable to the cockpit_session_t domain.
+
+.br
+.TP 5
+Paths:
+/usr/libexec/cockpit-ssh, /usr/libexec/cockpit-session
+
+.PP
+Note: File context can be temporarily modified with the chcon command.  If you want to permanently change the file context you need to use the
+.B semanage fcontext
+command.  This will modify the SELinux labeling database.  You will need to use
+.B restorecon
+to apply the labels.
+
+.SH "COMMANDS"
+.B semanage fcontext
+can also be used to manipulate default file context mappings.
+.PP
+.B semanage permissive
+can also be used to manipulate whether or not a process type is permissive.
+.PP
+.B semanage module
+can also be used to enable/disable/install/remove policy modules.
+
+.B semanage boolean
+can also be used to manipulate the booleans
+
+.PP
+.B system-config-selinux
+is a GUI tool available to customize SELinux policy settings.
+
+.SH AUTHOR
+This manual page was auto-generated using
+.B "sepolicy manpage".
+
+.SH "SEE ALSO"
+selinux(8), cockpit_session(8), semanage(8), restorecon(8), chcon(1), sepolicy(8), setsebool(8)
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -81,6 +81,13 @@ Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{v
 %define build_optional 1
 %endif
 
+# Ship custom SELinux policy only in Fedora and RHEL-9 onward
+%if 0%{?rhel} >= 9 || 0%{?fedora}
+%define selinuxtype targeted
+%define with_selinux 1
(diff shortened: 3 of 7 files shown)


marusak:
> DEBUG: make: /usr/share/selinux/devel/Makefile: No such file or directory

Right, needs `selinux-policy-devel` package.

martinpitt:
Summary from IRC discussion: Let's discuss with SELinux upstream why we need that silly extra package. I see zero advantages of it. IMHO cockpit-ws should have something like `Requires: (extra-selinux-dep if selinux-policy)` and on systems without SELinux the three unused files don't really matter. With a separate package, cockpit-ws would still have to depend on cockpit-selinux-policy to make installation work on SELinux systems, so this can't in principle help to avoid dependencies.

martinpitt:
> IMHO cockpit-ws should have something like Requires: (extra-selinux-dep if selinux-policy)

FTR, I don't think we'll actually need that -- everything that `%{?selinux_requires}` expands to is already present on a system which has SELinux enabled. This is just for the case where it would really have to pull in something extra.

The main work with that would be to make the `%post` scripts (and friends) conditional to "is selinux installed/available". But getting rid of these project-selinux-polixy packages (which at scale are a significant archive and local install clutter) seems worth that?

vmojzis:
> Summary from IRC discussion: Let's discuss with SELinux upstream why we need that silly extra package. I see zero advantages of it. IMHO cockpit-ws should have something like `Requires: (extra-selinux-dep if selinux-policy)` and on systems without SELinux the three unused files don't really matter. With a separate package, cockpit-ws would still have to depend on cockpit-selinux-policy to make installation work on SELinux systems, so this can't in principle help to avoid dependencies.

Well, you don't. The only issue I can see would occur when installing on systems without SELinux. /usr/share/selinux/packages belongs to selinux-policy package and without it there is no owner.
Also you'd need some custom code for selinux-policy dependency since %selinux_requires adds a dependency on the exact version of selinux-policy that was used for compilation (necessary to ensure compatibility).

The sub-package is really only for separating your code from SELinux.

tools/cockpit.spec
```diff
@@ -499,6 +514,34 @@ test -f %{_bindir}/firewall-cmd && firewall-cmd --reload --quiet || true
 %postun ws
 %systemd_postun_with_restart cockpit.socket cockpit.service
 
+%package selinux-policy
+Summary: SELinux policies
+Requires: cockpit-ws = %{version}-%{release}
```
vmojzis, line 519: If you decide to stick with cockpit-selinux-policy, it would be best to reverse this dependency (cockpit-ws requires cockpit-selinux-policy) so that the policy is installed first. Also, it could cause circular dependency in the future.

tools/cockpit.spec
```diff
@@ -499,6 +514,34 @@ test -f %{_bindir}/firewall-cmd && firewall-cmd --reload --quiet || true
...
+%post selinux-policy
+%selinux_modules_install -s %{selinuxtype} %{_datadir}/selinux/packages/%{selinuxtype}/%{name}.pp.bz2
+
+%postun selinux-policy
+if [ $1 -eq 0 ]; then
+    %selinux_modules_uninstall -s %{selinuxtype} %{name}
+fi
+
+%posttrans selinux-policy
+%selinux_relabel_post -s %{selinuxtype}
```
vmojzis, line 543: Since we're dealing with a service, we should move selinux_relabel_post to %post/%postun and add service restart after relabeling... see: https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Example_spec_file_changes_to_incorporate_-selinux_subpackage (version for daemons)

marusak:
Ok, I rewrote it and now it is being shipped in cockpit-ws package. `pre`, `post` and `postrun` scripts run only when `/usr/sbin/selinuxenabled` is present. I've tried installing cockpit-ws on fedora-34 with selinux and it applied the new policy and also tried installing it into F34 container where no selinux is present and it worked just fine.

tools/cockpit.spec
```diff
@@ -481,23 +495,43 @@ authentication via sssd/FreeIPA.
...
+    %selinux_modules_install -s %{selinuxtype} %{_datadir}/selinux/packages/%{selinuxtype}/%{name}.pp.bz2
+    %selinux_relabel_post -s %{selinuxtype}
+fi
 
 %preun ws
 %systemd_preun cockpit.socket cockpit.service
 
 %postun ws
 %systemd_postun_with_restart cockpit.socket cockpit.service
+# SELinux
```
marusak, line 530: I guess this should be first and then we should restart.

marusak:
I am running one test against bots#1942 and it successfully built rpms.

tools/cockpit.spec
```diff
@@ -481,22 +506,44 @@ authentication via sssd/FreeIPA.
...
+    %{_mandir}/man8/%{name}_ws_selinux.8.*
+    %ghost %{_sharedstatedir}/selinux/%{selinuxtype}/active/modules/200/%{name}
+%endif
+
 %pre ws
 getent group cockpit-ws >/dev/null || groupadd -r cockpit-ws
 getent passwd cockpit-ws >/dev/null || useradd -r -g cockpit-ws -d /nonexisting -s /sbin/nologin -c "User for cockpit web service" cockpit-ws
 getent group cockpit-wsinstance >/dev/null || groupadd -r cockpit-wsinstance
 getent passwd cockpit-wsinstance >/dev/null || useradd -r -g cockpit-wsinstance -d /nonexisting -s /sbin/nologin -c "User for cockpit-ws instances" cockpit-wsinstance
+# selinux
```
marusak, line 522: So these three script should run only when cockpit-ws actually ships those files (see `with_selinux` macro). It will fail especially in `%post` (`%pre` and `%postrun` will be just pointless operation (I think)).
Is there a good way of doing this? I think I cannot just use `with_selinux` macro as this happens in a different time. I could check if `%{_datadir}/selinux/packages/%{selinuxtype}/%{name}.pp.bz2` is present. Is there a better way of doing this?

tools/cockpit.spec
```diff
@@ -83,6 +83,11 @@ Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{v
 %define build_optional 1
 %endif
 
+# Ship custom SELinux policy only in Fedora and RHEL-9
+%if 0%{?rhel} == 9 || 0%{?fedora}
```
martinpitt, line 87: >= 9, please (Fedora ELN is already targetting 10)

selinux/cockpit_session_selinux.8
```diff
@@ -0,0 +1,255 @@
+.TH  "cockpit_session_selinux"  "8"  "21-04-16" "cockpit_session" "SELinux Policy cockpit_session"
```
martinpitt, line 1: FTR, it'd be better to integrate that into our normal docbook manpage flow, but this should be a follow-up.

tools/cockpit.spec
```diff
@@ -165,6 +178,11 @@ exec 2>&1
 
 make -j4 %{?extra_flags} all
 
+%if 0%{?with_selinux}
+    make -f /usr/share/selinux/devel/Makefile cockpit.pp
```
martinpitt, line 182: Interesting, that looks into the selinux/ folder by default and takes all files there? That seems quite implicit.

tools/cockpit.spec
```diff
@@ -189,6 +207,13 @@ install -p -m 644 tools/cockpit.pam $RPM_BUILD_ROOT%{_sysconfdir}/pam.d/cockpit
 rm -f %{buildroot}/%{_libdir}/cockpit/*.so
 install -D -p -m 644 AUTHORS COPYING README.md %{buildroot}%{_docdir}/cockpit/
 
+# selinx
```
martinpitt, line 210: typo, but also, the `%if` makes that clear enough anyway.

tools/cockpit.spec
```diff
@@ -189,6 +207,13 @@ install -p -m 644 tools/cockpit.pam $RPM_BUILD_ROOT%{_sysconfdir}/pam.d/cockpit
...
+%if 0%{?with_selinux}
+    install -D -m 644 %{name}.pp.bz2 %{buildroot}%{_datadir}/selinux/packages/%{selinuxtype}/%{name}.pp.bz2
```
martinpitt, line 212: It feels a bit weird having to install these from the spec only, instead of from `make install`. But good enough for now, especially while we don't ship these in Debian and don't build the manpage from XML/md

tools/cockpit.spec
martinpitt: Why wouldn't wrapping this into `with_selinux` work? That's processed very early in the spec file, and is a purely textual operation. The post-processed `%post` textual contents would then just get copied verbatim. Did you try it? If it doesn't work, then the runtime `test -e` seems ok, but it'd IMHO be more elegant at rpm build time.

martinpitt (changes requested):
Some nitpicks, but this looks really good, thank you!

tools/cockpit.spec
marusak: > that looks into the selinux/ folder by default and takes all files there? 

Indeed.

selinux/cockpit.fc
```diff
@@ -0,0 +1,18 @@
+# cockpit stuff
```
marusak, line 1: If I force push again I'll drop this comment

marusak:
@vmojzis I've rewritten it to be a part of cockpit-ws package. Could you please give it a one more look? Thanks!

martinpitt:
The [packit failures](http://artifacts.dev.testing-farm.io/8d1070eb-f7bb-4e94-9ccb-1b3004c69436/) look familiar, and unrelated. Retrying.

tools/cockpit.spec
```diff
@@ -481,22 +503,48 @@ authentication via sssd/FreeIPA.
...
+%endif
+
 %pre ws
 getent group cockpit-ws >/dev/null || groupadd -r cockpit-ws
 getent passwd cockpit-ws >/dev/null || useradd -r -g cockpit-ws -d /nonexisting -s /sbin/nologin -c "User for cockpit web service" cockpit-ws
 getent group cockpit-wsinstance >/dev/null || groupadd -r cockpit-wsinstance
 getent passwd cockpit-wsinstance >/dev/null || useradd -r -g cockpit-wsinstance -d /nonexisting -s /sbin/nologin -c "User for cockpit-ws instances" cockpit-wsinstance
 
+%if 0%{?with_selinux}
+if [ -s %{_sbindir}/selinuxenabled ]; then
```
martinpitt, line 520: This looks strange; What's the point of checking if an executable is non-empty? Wouldn't it make more sense to do something like `if %{_sbindir}/selinuxenabled 2>/dev/null; then` ?

martinpitt (changes requested):
Thanks! This by and large looks great to me, aside from the runtime condition.

One question: How does this make sure that cockpit-ws' policy overrides the one shipped by selinux-policy? I think I already asked that question in the tech talk back then, but unfortunately forgot the answer. Is there some kind of priority somewhere?

tools/cockpit.spec
martinpitt: (the `2>&/dev/null` to avoid noise if libselinux is not installed, like in containers)

tools/cockpit.spec
```diff
@@ -481,22 +503,48 @@ authentication via sssd/FreeIPA.
 %attr(4750, root, cockpit-wsinstance) %{_libexecdir}/cockpit-session
 %{_datadir}/cockpit/branding
 
+%if 0%{?with_selinux}
+    %{_datadir}/selinux/packages/%{selinuxtype}/%{name}.pp.bz2
+    %{_mandir}/man8/%{name}_session_selinux.8.*
+    %{_mandir}/man8/%{name}_ws_selinux.8.*
+    %ghost %{_sharedstatedir}/selinux/%{selinuxtype}/active/modules/200/%{name}
```
marusak, line 510: This `200` gives the priority. 100 is default priority for system policy, thus this one has higher priority.

vmojzis:
> @vmojzis I've rewritten it to be a part of cockpit-ws package. Could you please give it a one more look? Thanks!

As I wrote above, you need some custom code for selinux-policy dependency. The policy can only be safely installed on a system with the version of selinux-policy that was used for compilation (or higher). We normally use the %selinux_requires macro for that. Conditional require will probably do the trick.

marusak:
That macro is defined here: https://github.com/fedora-selinux/selinux-policy-macros/blob/master/macros.selinux-policy#L32 so I'll get inspired

martinpitt:
@maruak: Ah nice! So it seems that we at least need

    Conflicts: selinux-policy < %{_selinux_policy_version}

and *possibly* (if our %post scripts use anything from that)

    Requires: (policycoreutils if selinux-policy)

libselinux-utils is already a dependency of policycoreutils, so IMHO we don't need to mention that explicitly.

marusak:
```
Conflicts: selinux-policy < %{_selinux_policy_version}
```

Why turning it around? I was thinking of
```
Requires: (selinux-policy >= %{_selinux_policy_version} if selinux-policy-%{selinuxtype})
```
It is a bit weird one but this one will update `selinux-policy` if you are updating `cockpit-ws` and need newer policy. I am not 100% sure what dnf does if you have older version of selinux-policy that new cockpit-ws conflicts. (I think you would need to use `--allowerasing` to be able to update it).

I guess we also should require adequate version of `selinux-policy-base`, thus adding:
```
Requires(post): selinux-policy-base >= %{_selinux_policy_version}
```
Then these three are remaining. While you are correctly mentioning that these should already be present, I think it cannot hurt to mention them. Feels more correct that just assuming it will be pulled by something else.
```
Requires(post): libselinux-utils
Requires(post): policycoreutils
Requires(post): policycoreutils-python-utils
```

martinpitt:
> `Requires: (selinux-policy >= %{_selinux_policy_version} if selinux-policy-%{selinuxtype})`

Is that a thing? I don't see selinux-policy-%{selinuxtype} anywhere in 
[here](https://github.com/fedora-selinux/selinux-policy-macros/blob/master/macros.selinux-policy#L32). But conceptually,

    Requires: (selinux-policy >= %{_selinux_policy_version} if selinux-policy)

sounds very similar to `Conflicts: selinux-policy < %{_selinux_policy_version}`. I got brought up with Debian, where that conditional Requires does not exist, and I'd use the Conflicts:. But if that Requires: works, that's fine of course.

For the three remaining ones: Indeed you should explicitly mention dependencies that cockpit-ws calls directly in its %post and related scripts. The others should be the job of transitive dependencies IMHO.

marusak:
> Is that a thing?

It is. It is that package that is advised to use for checks if you should install your own selinux-policy package:
https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Adding_dependency_to_the_spec_file_of_corresponding_package

marusak:
```
rpm -q --requires cockpit-ws
(policycoreutils if selinux-policy-targeted)
(selinux-policy >= 34.3 if selinux-policy-targeted)
...
```

Installs fine and runs script on machine with selinux. Also installs fine and has no noise when installing into container.

marusak:
> Indeed you should explicitly mention dependencies that cockpit-ws calls directly in its %post and related scripts. The others should be the job of transitive dependencies IMHO.

Agree. I checked those macros we use in pre/post/postrun and all binaries come from `policycoreutils`.

marusak:
aha, packit is unhappy:
```
"error: line 451: Unknown rich dependency op 'selinux-policy': Requires: (selinux-policy >= package selinux-policy is not installed if selinux-policy-targeted)"]
```

Message: Preparation of the repository for creation of an SRPM failed: Failed to parse SPEC file: ["warning: line 414: It's not recommended to have unversioned Obsoletes: Obsoletes: cockpit-dashboard", "error: line 451: Unknown rich dependency op 'selinux-policy': Requires: (selinux-policy >= package selinux-policy is not installed if selinux-policy-targeted)"]
Exception: PackitSRPMException('Preparation of the repository for creation of an SRPM failed: Failed to parse SPEC file: ["warning: line 414: It\'s not recommended to have unversioned Obsoletes: Obsoletes: cockpit-dashboard", "error: line 451: Unknown rich dependency op \'selinux-policy\': Requires: (selinux-policy >= package selinux-policy is not installed if selinux-policy-targeted)"]'
```

File: tools/cockpit.spec [hunk 13]
```diff
@@ -81,6 +81,13 @@ Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{v
 %define build_optional 1
 %endif
 
+# Ship custom SELinux policy only in Fedora and RHEL-9 onward
+%if 0%{?rhel} >= 9 || 0%{?fedora}
+%define selinuxtype targeted
+%define with_selinux 1
+%define selinux_policy_version %(rpm -q --queryformat "%{V}" selinux-policy)
```
marusak, line 88: We cannot build this if there is no selinux-policy installed.
Of course, that is brought by BuildRequires. So maybe having this not as `%define` but directly in condition or something? Or that would still be too early as the rpm is parsed before? So just dummy version?

martinpitt: Dummy version seems like a good enough workaround for packit builds indeed. Thanks!

tools/cockpit.spec
```diff
@@ -81,6 +81,13 @@ Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{v
 %define build_optional 1
 %endif
 
+# Ship custom SELinux policy only in Fedora and RHEL-9 onward
+%if 0%{?rhel} >= 9 || 0%{?fedora}
+%define selinuxtype targeted
+%define with_selinux 1
+%define selinux_policy_version %(rpm --quiet -q selinux-policy && rpm -q --queryformat "%{V}" selinux-policy || echo 1)
```
martinpitt, line 88: Just in case you have to push again, please add a comment that the fallback is for packit.

martinpitt (approved):
Cool, this looks really good now! And mostly clean, aside from the little packit hack. :+1:

tools/cockpit.spec
```diff
@@ -481,22 +506,48 @@ authentication via sssd/FreeIPA.
...
 %post ws
 %tmpfiles_create cockpit-tempfiles.conf
 %systemd_post cockpit.socket cockpit.service
 # firewalld only partially picks up changes to its services files without this
 test -f %{_bindir}/firewall-cmd && firewall-cmd --reload --quiet || true
 
+%if 0%{?with_selinux}
+if %{_sbindir}/selinuxenabled 2>/dev/null; then
+    %selinux_modules_install -s %{selinuxtype} %{_datadir}/selinux/packages/%{selinuxtype}/%{name}.pp.bz2
+    %selinux_relabel_post -s %{selinuxtype}
```
vmojzis, line 537: I'd move this before "%systemd_post" to ensure the service is executed with the proper label.

vmojzis:
Other than my comment in the %post section, it looks great. Thanks :)

tools/cockpit.spec
vmojzis: Just realised that this does not contain release number. The version  can easily stay the same for 1/2 a year in this case. Please use both version and release (as is the case in %selinux_requires macro).

lnussel:
Is moving policies to packages the general direction in Fedora?
Wondering whether it would make sense to create file triggers to avoid adding scriptlets to all kinds of packages.

martinpitt:
@lnussel : Yes, the selinux team even gave some [presentation](https://docs.google.com/presentation/d/1A6i7dQ8v-Xs0UQduyJipptDDhofzRp5gX4vc7L3dQz4/edit#slide=id.g89895455fc_0_20) a while ago (not sure if you can see it, though) and asked around for packages adopting their own policy.

marusak:
@vmojzis when there is installed `selinux-policy-doc` package on system installing new cockpit-ws fails with:
```
Error: Transaction test error:
  file /usr/share/man/man8/cockpit_session_selinux.8.gz from install of cockpit-ws-242.81.g6c7b0a0d9-1.wip.fc34.x86_64 conflicts with file from package selinux-policy-doc-34.3-1.fc34.noarch
  file /usr/share/man/man8/cockpit_ws_selinux.8.gz from install of cockpit-ws-242.81.g6c7b0a0d9-1.wip.fc34.x86_64 conflicts with file from package selinux-policy-doc-34.3-1.fc34.noarch
```

Is there a best practice how to deal with this kind of situation? (We want to make it work and not just tell users to use `--replacefiles`)

martinpitt:
@marusak: Let's not block this forever, and just name them *.8cockpit.gz for the time being. We can rename them back once they get dropped from selinux-policy, and then add a versioned Conflicts.

PR #15713, opened 2021-04-19
test: Disable rhsm.service through systemd

martinpitt:
subscription-manager uses systemd now. This avoids a hanging unit start
when systemd tries to D-Bus activate it.

---

This happened in the downstream RHEL 8.5 gating tests. I hacked around it in the STI scripts there.


marusak (approved):
Makes sense and is much cleaner approach, thanks!

PR #15714, opened 2021-04-19
test: Select first pool from candlepin

martinpitt:
The POOL_COMMAND on recent candlepin versions now finds two pools with
almost exactly the same properties. Ensure that we only pick the first
matching one, to avoid building a broken multi-line command.


marusak (approved):
Makes sense, thanks!
Assuming green tests.

PR #15716, opened 2021-04-19
packagekit: Accept arbitary minutes in the OnCalendar unit property validator

### LINKED ISSUE #15712: Message in web UI - Automatic updates are not set up
Cockpit version: cockpit-242-1.fc33.x86_64
OS: Fedora 33
Page: Updates

Fresh install of Fedora 33 Server, enabled automatic updates, set to reboot every day, after one reboot - the page shows just a message 

Automatic updates
Automatic updates are not set up

And there is no button to set it up.

Steps to reproduce

1. Install Fedora 33 server
2. Set up automatic updates via cockpit web UI
3. After a reboot/update that part doesn't work any more

image

KKoukiou:
Previously when the user would have had arbitary minutes already set up,
we would set supported=false and hide the autoupdates card from the UI.

Fixes #15712

### DIFF
--- a/pkg/packagekit/autoupdates.jsx
+++ b/pkg/packagekit/autoupdates.jsx
@@ -99,7 +99,7 @@ class DnfImpl extends ImplBase {
                             if (output.indexOf("InactiveSec=1d\n") >= 0)
                                 this.day = this.time = "";
                             else
-                                this.supported = false;
+                                this.parsing_failed = true;
                         }
 
                         debug(`dnf getConfig: supported ${this.supported}, enabled ${this.enabled}, type ${this.type}, day ${this.day}, time ${this.time}, installed ${this.installed}; raw response '${output}'`);
@@ -135,7 +135,7 @@ class DnfImpl extends ImplBase {
         if (words.length == 1 && validTime.test(words[0]))
             this.time = words[0].replace(/^0+/, "");
         else
-            this.supported = false;
+            this.parsing_failed = true;
     }
 
     setConfig(enabled, type, day, time) {
@@ -359,6 +359,11 @@ export class AutoUpdates extends React.Component {
 
         return (<>
             <div id="autoupdates-settings">
+                {enabled && this.state.backend.parsing_failed &&
+                <Alert isInline
+                       variant="info"
+                       className="autoupdates-card-error"
+                       title={_("Failed to parse unit files for dnf-automatic.timer or dnf-automatic-install.timer. Please remove custom overrides to enable this feature.")} />}
                 <Flex alignItems={{ default: 'alignItemsCenter' }}>
                     <Flex grow={{ default: 'grow' }} alignItems={{ default: 'alignItemsBaseline' }}>
                         <FlexItem>
--- a/pkg/packagekit/autoupdates.jsx
+++ b/pkg/packagekit/autoupdates.jsx
@@ -116,8 +116,7 @@ class DnfImpl extends ImplBase {
     parseCalendar(spec) {
         // see systemd.time(7); we only support what we write, otherwise we treat it as custom config and "unsupported"
         const daysOfWeek = ["mon", "tue", "wed", "thu", "fri", "sat", "sun"];
-        // TODO: allow arbitrary minutes once we support that in the UI widget
-        const validTime = /^((|0|1)[0-9]|2[0-3]):00$/;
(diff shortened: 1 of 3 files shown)


pkg/packagekit/autoupdates.jsx
```diff
@@ -116,8 +116,7 @@ class DnfImpl extends ImplBase {
     parseCalendar(spec) {
         // see systemd.time(7); we only support what we write, otherwise we treat it as custom config and "unsupported"
         const daysOfWeek = ["mon", "tue", "wed", "thu", "fri", "sat", "sun"];
-        // TODO: allow arbitrary minutes once we support that in the UI widget
-        const validTime = /^((|0|1)[0-9]|2[0-3]):00$/;
+        const validTime = /^((|0|1)[0-9]|2[0-3]):[0-5][0-9]$/;
```
marusak, line 119: Do we support than in UI widget though? :)

marusak:
Was this the reason why it would not show it up for you?

pkg/packagekit/autoupdates.jsx
KKoukiou: Yes
Screen Shot 2021-04-20 at 16 51 51

KKoukiou:
> Was this the reason why it would not show it up for you?

Yes

marusak:
> Yes

Could we add `console.log` when we fail to parse the time and set `supported` to false to say something like `Could not understand automatic updates format: <what_it_was_trying_to_parse>` so that the next time we would see that in console directly?

pkg/packagekit/autoupdates.jsx
marusak: Can you also please post screenshot when you save changes?

KKoukiou:
> > Yes
> 
> Could we add `console.log` when we fail to parse the time and set `supported` to false to say something like `Could not understand automatic updates format: <what_it_was_trying_to_parse>` so that the next time we would see that in console directly?

Sure, I will update.

pkg/packagekit/autoupdates.jsx
```diff
@@ -359,6 +359,11 @@ export class AutoUpdates extends React.Component {
 
         return (<>
             <div id="autoupdates-settings">
+                {enabled && this.state.backend.parsing_failed &&
+                <Alert isInline
+                       variant="danger"
```
martinpitt, line 364: danger seems inflated: neither is there anything actually wrong on the system, nor can the user do any dangerous action here. info seems enough?

pkg/packagekit/autoupdates.jsx
```diff
@@ -359,6 +359,11 @@ export class AutoUpdates extends React.Component {
...
+                       className="autoupdates-card-error"
+                       title={_("Failed to parse unit files for dnf-automatic.timer or dnf-automatic-install.timer. Please remove custom overrides to enable this feature.")} />}
```
martinpitt, line 366: Rather technical, but accurate and helpful for someone who knows what this is about. I first thought "let's add a link to the services page", but (1) that can't arbitrarily edit overrides either, and (2) we don't precisely know here *which* unit. So let's keep it as-is.

martinpitt (changes requested):
Thanks for looking into this bug! I'm glad to see one of my old TODOs being addressed.

pkg/packagekit/autoupdates.jsx
KKoukiou: This will alert will emerge from the following scenarios:
*  invalid onCalendar 
* invalid OnUnitInactiveSec
* usage of none of these two

The first two scenarios are actually a problematic configuration, having some typo in the config file or so. 
The last one, is weird, it's a timer unit file without repeating pattern?

@martinpitt therefore i thought it's a danger. I am happy to revert it to info, but am I missing something here?

martinpitt: I thought the normal case was just that the custom timer uses a format that we can't parse. Like the `OnCalendar=*-02-* 11:00` that we have in the test suite, which is a perfectly legal `systemd.time(7)` specification, but too complicated for our parser.

KKoukiou: Indeed. Just noticed.

martinpitt:
Sorry for the packit failures -- I had to reinitialize our -dist repo as it became too big (I'm working on fixing that).

martinpitt (dismissed):
Thanks!

martinpitt (approved):

PR #15717, opened 2021-04-19
terminal: show an alert if navigator.clipboard.readText API is not supported

### LINKED ISSUE #12939: Terminal: paste action doesn't work if using the menu to paste
Cockpit version: 203
OS: Fedora
Page: Terminal

I tried to copy and paste in the terminal. For some reason paste doesn't work if you use the right click menu. If I use the shortcut (shift+insert) it works as expected.

I get this message in the console of the web browser:

```
Text could not be pasted, use Shift+Insert: TypeError: navigator.clipboard.readText is not a function 5 cockpit-components-terminal.jsx:200:20
    setText cockpit-components-terminal.jsx:200
    React 9
    unstable_runWithPriority scheduler.production.min.js:20
    React 3
```

KKoukiou:
Note: The patternfly import needs to be done before the react-core imports,
otherwise it resets the styles. (resulting is false modal background)

Fixes #12939

The alert modal looks like this:
Screen Shot 2021-04-20 at 18 56 14

### DIFF
--- a/pkg/lib/cockpit-components-terminal.jsx
+++ b/pkg/lib/cockpit-components-terminal.jsx
@@ -19,9 +19,15 @@
 
 import React from "react";
 import PropTypes from "prop-types";
+import { Modal, Button } from "@patternfly/react-core";
 import { Terminal as Term } from "xterm";
+
 import { ContextMenu } from "cockpit-components-context-menu.jsx";
+import cockpit from "cockpit";
+
 import "console.css";
+
+const _ = cockpit.gettext;
 
 const theme_core = {
     yellow: "#b58900",
@@ -116,7 +122,8 @@ export class Terminal extends React.Component {
             cursorBlink: true,
             fontSize: props.fontSize || 16,
             fontFamily: 'Menlo, Monaco, Consolas, monospace',
-            screenReaderMode: true
+            screenReaderMode: true,
+            showPastingModal: false,
         });
 
         this.terminalRef = React.createRef();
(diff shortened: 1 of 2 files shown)


pkg/lib/cockpit-components-terminal.jsx
```diff
@@ -191,6 +198,18 @@ export class Terminal extends React.Component {
     render() {
         return (
             <>
+                <Modal title={_("Pasting in cockpit terminal")}
```
martinpitt, line 201: Please no "cockpit". Maybe "Paste error"?

pkg/lib/cockpit-components-terminal.jsx
```diff
@@ -191,6 +198,18 @@ export class Terminal extends React.Component {
...
+                       position="top"
+                       variant="small"
+                       isOpen={this.state.showPastingModal}
+                       onClose={() => this.setState({ showPastingModal: false })}
+                       actions={[
+                           <Button key="cancel" variant="secondary" onClick={() => this.setState({ showPastingModal: false })}>
+                               {_("Close")}
+                           </Button>
+                       ]}>
+                    {_("This action is unavailable using the context menu, but you can still use Shift+Insert for paste.")}
```
martinpitt, line 211: Like this it sounds a bit confusing, like we papered over our own bug. Maybe "Your browser does not allow paste from the context menu. You can use Shift+Insert."

martinpitt (changes requested):
Thanks for fixing this! It's the little things that make a polished  product :+1: Just some verbiage nitpicking.

pkg/lib/cockpit-components-terminal.jsx
KKoukiou: Simple and good. Yes, I always forget that cockpit is not a term not RHEL.

pkg/lib/cockpit-components-terminal.jsx
KKoukiou: I like the suggestion, replaced.

martinpitt (approved):
Epharisto!

PR #15718, opened 2021-04-20
workflows: Reduce published dist tarball retention to 3 days

martinpitt:
The repository gets too big with 7 days of builds, and makes prune-dist
fail due to too big pushes.


martinpitt:
After this lands, I'll reset the dist cache to clean up. Hopefully we'll soon replace it with a much less wasteful approach (like we use in c-machines), but that still requires some work on our build system.

KKoukiou (approved):

PR #15719, opened 2021-04-20
Package lock changes

allisonkarlitskaya:
Here's the first 3 commits from #15700 which seem uncontroversial, and which block other stuff from landing.


martinpitt (approved):
Cheers!

PR #15720, opened 2021-04-20
Fix debian/copyright failure during `make dist` with pre-built dist/

martinpitt:
Our current [cockpit-dist](https://github.com/cockpit-project/cockpit-dist/) approach is broken: it becomes too big too quickly, and causes [prune-dist](https://github.com/cockpit-project/cockpit/actions/workflows/prune-dist.yml) to just error out on `git push -f`. In cockpit-machines we recently [reworked the dist storing](https://github.com/cockpit-project/cockpit-machines/pull/84) to keep the literal files in git, not the full tarballs.

For that to work in cockpit, we need to have `make dist` *not* do npm/webpack after unpacking pre-built dist/ and package-lock.json. It should be fast (well under 1 minute) to retain the "fast local PR testing/development" behaviour that we have right now.

The main thing that is currently broken is the generated debian/copyright file.

With the fixes here, this sequence works:

    git clean -ffdx; test/make_dist.py && time sh -c './autogen.sh && make --silent NO_DIST_CACHE=1 XZ_COMPRESS_FLAGS=-0 dist'

The first make_dist downloads the tarball, and the `make` just does an actual local tarball build to validate that this works. This has to run autogen.sh, configure, and the `make dist` part, which takes about 14s in total. This is still bearable IMHO.

 - [x] #15719
 - [x] #15727

### DIFF
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -131,12 +131,15 @@ def download_dist(wait=False):
         tar_path = os.path.realpath(names[0])
 
     # Extract npm/webpack related files locally for speeding up the build and allowing integration tests to run
-    unpack_paths = [d for d in ["dist", "node_modules", "package-lock.json"] if not os.path.exists(d)]
+    unpack_paths = [d for d in ["dist", "node_modules", "package-lock.json", "tools/debian/copyright"] if not os.path.exists(d)]
     if unpack_paths:
         message("make_dist: Extracting from tarball:", ' '.join(unpack_paths))
         prefix = os.path.basename(tar_path).split('.tar')[0] + '/'
         prefixed_unpack_paths = [prefix + d for d in unpack_paths]
         subprocess.check_call(["tar", "--touch", "--strip-components=1", "-xf", tar_path] + prefixed_unpack_paths)
+        # after the above, dist manifests may be newer than copyright due to tar touching order
+        # restore file time stamps as they would be after a build
+        subprocess.check_call(["touch", "tools/debian/copyright"])
 
     return tar_path
 
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 *.policy
 *.so
 *.mo
+*.tmp
 test-*.log
 test-*.trs
 *~
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,7 @@ EXTRA_DIST = \
 	package.json \
 	package-lock.json \
 	README.md \
+	tools/debian/copyright \
 	$(NULL)
 
 clean-local::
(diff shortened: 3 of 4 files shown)


test/make_dist.py
```diff
@@ -131,9 +131,9 @@ def download_dist(wait=False):
         tar_path = os.path.realpath(names[0])
 
     # Extract node_modules and dist locally for speeding up the build and allowing integration tests to run
```
KKoukiou, line 133: Do you want to add package-lock.json in the comment here as well? 
And maybe rename the unpack_dirs to unpack_items? or something else. Jist nitpicking here I don't care so much about it.

test/make_dist.py
```diff
@@ -131,12 +131,14 @@ def download_dist(wait=False):
...
-    unpack_dirs = [d for d in ["dist", "node_modules", "package-lock.json"] if not os.path.exists(d)]
+    unpack_dirs = [d for d in ["dist", "node_modules", "package-lock.json", "tools/debian/copyright"] if not os.path.exists(d)]
     if unpack_dirs:
         message("make_dist: Extracting from tarball:", ' '.join(unpack_dirs))
         prefix = os.path.basename(tar_path).split('.tar')[0] + '/'
         prefixed_unpack_dirs = [prefix + d for d in unpack_dirs]
         subprocess.check_call(["tar", "--touch", "--strip-components=1", "-xf", tar_path] + prefixed_unpack_dirs)
+        # after the above, dist manifests may be newer than copyright due to tar touching order, avoid rebuild
```
KKoukiou, line 140: It does not even do it in alphabetical order? Not need to answer, that is a weird one though :P

KKoukiou (approved):

allisonkarlitskaya:
I feel like we're starting to push up against the edge of the limitations that we get from using `make dist` to generate the thing that we commit to git.  The debian copyright thing is a good example of that.  If we allow ourselves to dist a proper cache of dist/, we could simply include the correct information there for allowing `make dist` to build this file as normal (assuming we have node_modules as wel, which should be really really easy, once #15700 lands).

I'm also against using the dist cache as a mechanism for official release tarballs, as we discussed a lot on IRC this morning.  We need quite a lot more guarantees about consistency/security/etc. before I'd be happy with that.  I also don't see a big advantage here, other than saving a few minutes, which isn't the highest priority on the top of my head when we're releasing.

I feel like we should wait a little bit here....

test/make_dist.py
martinpitt: I strongly suspect this depends on the order of files in the tar -- this is a stream processor by nature, so I guess it iterates over the files in the tar, filters according to the command line, and writes them out in that order.

martinpitt:
I don't mind moving the debian copyright file to dist, then this would be a bit simpler and indeed also more elegant, as we wouldn't have to put it into the "source" tree.

As I wrote in the description, the urgent goal here is to unbreak the current dist caching for PRs, packit, and local developers. Committing tarballs clearly does not work well. Releases are much less of a concern, and with the upcoming fixed `node_modules` pinning and your experimentation result that webpack is reproducible, this will fix the most pressing issue.

test/make_dist.py
martinpitt: Right. I did that in the split out PR #15727.

martinpitt:
> I don't mind moving the debian copyright file to dist

On second thought I actually do -- we currently have the nice situation that most of dist (all except dist/guide) is built by webpack. It feels wrong to put more make-generated bits into it.

martinpitt:
Pushed some cleanups, but this is blocked on #15727 right now.

allisonkarlitskaya (approved):

allisonkarlitskaya (approved):
Hilarious that `tools/debian/copyright` is already in `.gitignore`.  This is clearly not the first time we've messed around with this logic...

PR #15721, opened 2021-04-20
metrics: Show active menu item and add documentation to pcp

marusak:
Currently when one  would navigate to `/metrics` (through "View details and history") no menu item would be shown as active. Also it was not possible to define documentation items for metrics.

Problem with this component is that it is being treated as separate component but it does not have separate menu entry. There are a few options how this could be fixed:
- Add 'Metrics' menu item
- Serve this on `/system/metrics` and move `pkg/metrics` source into `pkg/systemd/source`
- Do what I have done in this PR
- Some other hack/workaround I haven't though of (?)

Before I document this new option, write tests and  cleanup commits I would like to hear what others think. Is this step in a good direction or would you solve this differently?

### DIFF
--- a/doc/guide/packages.xml
+++ b/doc/guide/packages.xml
@@ -136,8 +136,9 @@ $ cockpit-bridge --packages
       <varlistentry>
         <term>parent</term>
         <listitem><para>This option is used when module does not have its own menu item but is
-          a part of a different module. This is described by JSON object with property <code>component</code>
-          which takes name of the superordinate component.</para></listitem>
+          a part of a different module. This is described by JSON object with properties <code>component</code>
+          which takes name of the superordinate component and <code>docs</code> with list of documentation URLs
+          for the given page. See below for structure of <code>docs</code> property.</para></listitem>
       </varlistentry>
     </variablelist>
 
--- a/doc/guide/packages.xml
+++ b/doc/guide/packages.xml
@@ -133,6 +133,12 @@ $ cockpit-bridge --packages
         explained below.
         </para></listitem>
       </varlistentry>
+      <varlistentry>
+        <term>parent</term>
+        <listitem><para>This option is used when module does not have its own menu item but is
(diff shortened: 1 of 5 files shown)


martinpitt:
I find this PR quite clever actually (moving to draft though, for the TODOs). The one thing that I wouldn't like is to move the code into pkg/systemd -- I specifically wanted it to be a separate webpack as systemd is already large enough as-is, and thus annoying for development.

As for adding an extra "Metrics" menu item,  I have no strong opinion. The menu is already quite  large, but OTOH Metrics may also fit well as a top-level category like Storage or Accounts, or  at least as a tool (although that'd fit less well, I'd say).

More opinions?

KKoukiou:
@marusak @martinpitt Aside from the menu being long alteady, another issue i see about having the metrics as a separate menu item, is that for people who have disabled PCP that page does not have much content. Therefore I would rather keep it under overview as a parent component. 

No strong opinion here either however, happy with either case.

marusak:
I want to get this in before I record video about metrics page.
I've added documentation and test. Ready for review.

martinpitt (dismissed):
I like this! It feels natural, unsurprising, and unintrusive.

marusak:
Hmm, 3 times in a row failed [check-shell-multi-machine-key TestMultiMachineKeyAuth.testBasic](https://logs.cockpit-project.org/logs/pull-15721-20210422-144043-5f15f97e-fedora-coreos/log.html#129-2) on coreos. I don't think this PR might break it though.

marusak:
Bummer, it tries to run this test 3 times and I don't think it is stable enough to survive that. Maybe a good opportunity to improve this test.

martinpitt:
@marusask: I'd be okay with landing as-is, as this was blocking you. It's certainly good enough.

marusak:
> as this was blocking you

I can still prepare environment with this patch even when not merged

martinpitt (approved):
Niice!

PR #15722, opened 2021-04-20
firewall: call zone remove API method when deleting a zone from the UI

### LINKED ISSUE #12279: firewall: Interfaces managed by networkmanager results in weird behaviour on the firewall page
Some interfaces are managed by networkmanager. These can be removed from the runtime configuration but not from the permanent one. Aftre a reload the zone attached to this interface would pop right back.

Maybe disable removing zones that are attached to interfaces managed by networkmanager?

KKoukiou:
Update: so with @Gundersanne with agreed to block this PR, and make this change together with adding the ability to remove interfaces and sources from the zone, so that we don't actually remove a feature right now from the cockpit firewall UI.

Added blocked because this now needs design, to add the new features.

<hr />

Previously the delete operation of the UI did not actually remove the
zone from the firewalld configuration. Instead it emptied it, by
removing all the interfaces and sources from it.

That's especially problematic with predefined (builtin) zones, which get redefined
when the firewalld gets restarted. In other words, if the user deleted
from the UI a predefined zone and then reloaded the firewall, all their
changes would be reverted.

Let's simplify the logic here, by just calling remove method instead of our
logic of emptying the zone. In addition we now show an error in the UI if the zone
failed to get removed (in the case of built it zones for example).
This is much more useful instead of faking the deletion, and then
suprising the users by reappearing deleted zones.

This change lastly follows the idea that all changes performed from
firewall UI are persistent.

Fixes #12279

### DIFF
--- a/pkg/networkmanager/firewall-client.js
+++ b/pkg/networkmanager/firewall-client.js
@@ -460,46 +460,16 @@ firewall.activateZone = (zone, interfaces, sources) => {
     return p;
 };
 
-/*
- * A zone is considered deactivated when it has no interfaces or sources.
- */
-firewall.deactiveateZone = (zone) => {
-    const zoneObject = firewall.zones[zone];
-    let promises = zoneObject.interfaces.map(i => firewalld_dbus.call('/org/fedoraproject/FirewallD1',
-                                                                      'org.fedoraproject.FirewallD1.zone',
-                                                                      'removeInterface', [zone, i]));
-    promises = promises.concat(zoneObject.source.map(s => firewalld_dbus.call('/org/fedoraproject/FirewallD1',
-                                                                              'org.fedoraproject.FirewallD1.zone',
-                                                                              'removeSource', [zone, s])));
-    let p = Promise.all(promises).then(() => firewalld_dbus.call('/org/fedoraproject/FirewallD1/config',
-                                                                 'org.fedoraproject.FirewallD1.config',
-                                                                 'getZoneByName', [zone]));
-    p = p.then(path => {
-        /* Once this signal is received, it's safe to actually emit the changed
-         * signal and thus update the UI */
-        const subscription = firewalld_dbus.subscribe({
-            interface: 'org.fedoraproject.FirewallD1.config.zone',
-            path: path[0],
-            member: 'Updated'
-        }, (path, iface, signal, args) => {
-            firewall.activeZones.delete(args[0]);
-            getZones().then(() => getServices());
-            subscription.remove();
-        });
-
+firewall.removeZone = (zone) => {
+    const p = firewalld_dbus.call('/org/fedoraproject/FirewallD1/config',
+                                  'org.fedoraproject.FirewallD1.config',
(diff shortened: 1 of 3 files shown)


KKoukiou:
So after investigating the API half day already today, I really believe that we should expose the zones properties better in the UI, interfaces and sources should be editable as well, so that the user can actually remove them and do what this 'delete' operation previously did.

martinpitt:
Makes sense to me at first sight, thanks! @Gundersanne, do you still remember why it was written that way before? Did the `.remove()` method simply not exist back then, or was there some reason to not actually delete the zones? Dank je wel!

KKoukiou:
@martinpitt so I think the reason it was done like this was that it's not allowed to call the remove() API on built in zones so @Gundersanne we tried to workaround it.
Therefore the previous implementation was deactivating them instead of removing, by removing all the sources and interfaces from them. This would hide them from the UI but only till next firewalld reload, when the interfaces would be added back automatically.

So I decided to instead expose the API as the CLI would do that, but also offer the ability in next commit to delete interfaces and sources, so the users who really want the previous effect can still achieve it.

Gundersanne:
So to recap my thoughts from irc: I indeed didn't use `remove()` because when used on built-in zones it will fail. And the way to 'deactivate' a builtin zone is to simply remove all interfaces/sources from it, so it doesn't actually apply to anything. This happens to both permanent and runtime configurations. Built-in zones are the most common.

There are certain zones that are added at runtime, for instance if libvirt and firewalld are running, libvirt will automatically add the libvirt zone to the firewall at runtime. Removing the libvirt zone and then restarting (I think both services need to be restarted), results in the libvirt zone being added again.

And I think NetworkManager can manage zones the same way (not 100% sure), and that is the 'weird' behaviour. Where you delete a zone but NM will recreate it at runtime. But I'm not sure.

KKoukiou:
So I just checked the behavior in the firewall-config and it does not allow one to delete the build-in zones. The 'Remove' button is disabled for the permanent operations, and missing completely for the runtime.

I think this is already an improvement, but as a followup I would do that they do there, that means always disable the button (and show a tooltip) for the builtin zones - since we only do permanent changes here.

Gundersanne:
Hmm, I'm getting the 'builtin' error as well with libvirt

image


Also how do I 'deactivate' a builtin zone now?

KKoukiou:
@Gundersanne is it maybe a builtin zone as well ? I mean the error is API generated.
I think we need to add an extra feature for the user to be able to remove interfaces and sources from the UI. So if they remove these, the zone would not appear anymore, (as the previous delete operation did).

KKoukiou:
I am also happy to block this PR, and add a commit for the new feature, aka deleting interfaces and sources from a zone. But for that I would need a design.

KKoukiou:
@garrett we have agreed to block this on a new design, right?It's not urgent feature of course just want to make sure it's clear.
Do you need a separate issue?

KKoukiou:
I am closing these - it needs design and it's not on our high prio list.

Issue #11904, opened 2019-05-23
Add IPs to firewall without port


menathor:
I see there has been some discussion about adding IPs with ports in the firewall. It would be great if we could also add IPs without having to specify a service or port port, i.e. the equivalent of:

firewall-cmd --permanent --zone=public --add-source=xxx.xxx.xxx.xxx/32

There are times when we need to whitelist an IP for all ports

martinpitt:
@Gundersanne, I thought your most recent work allows that now? I. e. through defining a new zone with IP masks, and adding services to that zone?

Gundersanne:
Yes but right now we only show/add/remove zones which have the `default` (= reject) target. So these zones function as a whitelist for the defined services.

But we don't support the most trusted zones, where added services function as a blacklist (and a source address range would function as a whitelist).

So sadly we don't support this (yet).

Also the `public` zone has a `default` target so would never work to whitelist an entire address range.

marusak:
I think this still makes sense, but it would need some good design.

Gundersanne:
We'd need to distinguish between blacklist and whitelist zones.


![image](https://user-images.githubusercontent.com/11140201/67847014-6c7abc00-fb02-11e9-8517-2962a9cdd7bf.png)

martinpitt:
The firewall page has been able to define zones with IP ranges for a long time:

![image](https://user-images.githubusercontent.com/200109/213634789-13b4216e-05ea-4ccb-a7b7-d732a04d8c17.png)

I retitle this issue as it's really confusing otherwise.

PR #15724, opened 2021-04-20
[WIP]: firewall: (in my opinion) simplify zones card

KKoukiou:
So today I spent a short time in firewall page, I understood what the whole zones concept is about and I thought we can expose it better. The previous implementation was:
* quite complicated to understand - especially the Add services/Port dialog
* not possible to extend - see the need to add / remove interfaces and resources

This is far from being a ready implementation I just want some thumbs up or thumbs down, so that I know if it makes sense to look more into this direction.

Here are screenshots from the all the affected UI parts:

The dialog where split to 'Add ports' and 'Add services':
Screen Shot 2021-04-21 at 18 05 45
Screen Shot 2021-04-21 at 18 05 30

The different zone part were converted into tabs: 
Screen Shot 2021-04-21 at 18 07 46
Screen Shot 2021-04-21 at 18 07 41
Screen Shot 2021-04-21 at 18 07 38
Screen Shot 2021-04-21 at 18 07 36

@garrett you mentioned that you were part of the initial design, do you have something against this direction? As I say, take this PR only as a proof of concept, not a ready to review bit.

### DIFF
--- a/pkg/networkmanager/firewall-client.js
+++ b/pkg/networkmanager/firewall-client.js
@@ -472,4 +472,27 @@ firewall.removeZone = (zone) => {
     });
 };
 
+/*
+ * Add port into the specified zone
+ *
+ * Returns a promise that resolves when the port is added.
+ */
+firewall.addPort = (zone, port) => {
+    return firewalld_dbus.call('/org/fedoraproject/FirewallD1/config',
+                               'org.fedoraproject.FirewallD1.config',
+                               'getZoneByName', [zone])
+            .then(path => firewalld_dbus.call(path[0], 'org.fedoraproject.FirewallD1.config.zone',
+                                              'addPort', [port[0], port[1]]));
+};
+
+/*
+ * Like addPort(), but adds multiple ports at once
+ * to the specified zones.
+ *
+ * Returns a promise that resolves when all ports are added.
+ */
+firewall.addPorts = (zone, ports) =>
+    Promise.all(ports.map(p => firewall.addPort(zone, p)))
+            .then(() => firewall.reload());
+
 export default firewall;
--- a/pkg/networkmanager/firewall-client.js
+++ b/pkg/networkmanager/firewall-client.js
@@ -460,46 +460,16 @@ firewall.activateZone = (zone, interfaces, sources) => {
     return p;
 };
 
-/*
- * A zone is considered deactivated when it has no interfaces or sources.
- */
-firewall.deactiveateZone = (zone) => {
-    const zoneObject = firewall.zones[zone];
-    let promises = zoneObject.interfaces.map(i => firewalld_dbus.call('/org/fedoraproject/FirewallD1',
-                                                                      'org.fedoraproject.FirewallD1.zone',
-                                                                      'removeInterface', [zone, i]));
-    promises = promises.concat(zoneObject.source.map(s => firewalld_dbus.call('/org/fedoraproject/FirewallD1',
-                                                                              'org.fedoraproject.FirewallD1.zone',
-                                                                              'removeSource', [zone, s])));
-    let p = Promise.all(promises).then(() => firewalld_dbus.call('/org/fedoraproject/FirewallD1/config',
-                                                                 'org.fedoraproject.FirewallD1.config',
-                                                                 'getZoneByName', [zone]));
-    p = p.then(path => {
-        /* Once this signal is received, it's safe to actually emit the changed
-         * signal and thus update the UI */
-        const subscription = firewalld_dbus.subscribe({
-            interface: 'org.fedoraproject.FirewallD1.config.zone',
-            path: path[0],
-            member: 'Updated'
-        }, (path, iface, signal, args) => {
-            firewall.activeZones.delete(args[0]);
-            getZones().then(() => getServices());
-            subscription.remove();
-        });
-
+firewall.removeZone = (zone) => {
+    const p = firewalld_dbus.call('/org/fedoraproject/FirewallD1/config',
+                                  'org.fedoraproject.FirewallD1.config',
+                                  'getZoneByName', [zone]);
+    return p.then(path => {
         return firewalld_dbus.call(path[0],
                                    'org.fedoraproject.FirewallD1.config.zone',
-                                   'getSettings', [])
-                .then(settings => {
-                    settings[0][10] = [];
-                    settings[0][11] = [];
-                    return firewalld_dbus.call(path[0],
-                                               'org.fedoraproject.FirewallD1.config.zone',
-                                               'update', [settings[0]]);
-                });
+                                   'remove', [])
+                .then(() => firewall.reload());
     });
-
-    return p.catch(error => console.warn(error));
 };
 
 export default firewall;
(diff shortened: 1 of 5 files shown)


KKoukiou:
So now we can really straightforward implement feature requests like https://github.com/cockpit-project/cockpit/issues/11904 without squeezing our brains for an appropriate design.

KKoukiou:
@marusak some first opinion here? nothing for review yet, just fetching some feedback on this re-approaching.

garrett:
Firstly: Thanks for the experimentation!

## Ports

The idea with ports versus services is that the services should be the primary way of poking holes in the firewall. Ports should be a secondary thing, used when there's not a service. A lot of firewalls do not have the concept of services, so people have to manually add ports.

Therefore, if we're doing something like this, we might want to consider calling the tab "custom ports" instead of just "ports".

## Tabs vs. firewall-at-a-glance

This redesign also obfuscates a lot of the information about the firewall; it's all hidden behind tabs in this PR. The current implementation is meant to give someone a quick overview of what's going on. This version breaks that concept by forcing someone to click through all the tabs in every zone to understand their firewall and then remember each tab for each zone. There's no more "here's your firewall at a glance".

Perhaps we can come up with a hybrid approach that allows someone to see their firewall at a glance and still have it a bit more modular too.

KKoukiou:
@garrett thanks for the feedback. I understand that it's nice to have al information in sight, instead of hidden behind tabs, but if that means combining operations into one dialog (like we did before for 'Add services' and 'Add ports') or making our life harder to add new features (removing and adding interfaces and sources) then I feel it generates more problems than the benefits it offers.

So from this PR:
* I think it's really important to spit the dialog into 'Add ports' and 'Add services', the idea of creating a service before adding ports which exists now, is really not intuitive.
* Regarding the tabular approach I am really open to other ideas, we could also move the content of zones into their separate pages, but this also harms the full overview of the firewall status.
* I believe with this design we have so much more space and we have the ability to add also more inline documentation to clarify the concepts. Firewalld page from the usability studies has proven to be a pain point for users, and I think the only way we can solve it it by seperating better the concepts, (Zones, Services, Ports, Interfaces, Sources ... ) and then adding inline documentation for each of them.

garrett:
The description text is overwhelming here (and is a bit absurd being very verbose, hardcoded English in other languages... but we're pulling this from the descriptions; this is just an aside :wink:) and it causes the action button to wrap:

Screenshot_2021-04-21 Netzwerk - garrett Rain(2)

I think it'd be better if we had a summary of important info not on the first tab instead, and move the description elsewhere (perhaps in a tooltip behind an 🛈). 

What could go there instead? For example, the network interface. And if you click it, it'd take you to the interface tab. And perhaps info that there's a source override (not necessarily what, as that'd get verbose — just that there's # of them), and you could click it and it'd take you its respective tab. Doing something like this would make it easier to understand what's going on and we could have the best of both UIs.

Additionally, we could have "services and ports" as the first tab, with services at the top and ports below. There could be two buttons at the top, one for "Add services" and another for "Add custom ports". This would allow for having tabs and separate dialogs... _and_ also help with a glanceable overview of the firewall.

---

Also, mobile has issues, but I think that's expected at this stage of development. (Fixing the text, as suggested above, would help. And doing something with the tabs so they're not using a scrolly overflow would also help. But we can look at this later, once we have the large desktop UI sorted out.)

Screenshot_2021-04-21 Netzwerk - garrett Rain(3)

garrett:
> then adding inline documentation for each of them

:100: Yeah, totally.

For example: It's not clear whatsoever what "sources" is for example. Having inline doc would help a lot:

Screenshot_2021-04-21 Netzwerk - garrett Rain(4)

KKoukiou:
@garrett sorry for the description it was not supposed to be there, I was just trying things out and forgot to revert that commit. Pretend it's not here, and I will remove on the next push.

marusak:
While splitting it into tabs makes sense, I agree that it makes much harder to actually see what any given zone does. For example I want to know if I have opened port 90. I'll need to go through 4 tabs to know if it is not opened.
Also do we assume users are fully aware of diferences between services, ports, interfaces ans sources? So if someone is asked to "Open port 1234 for the foo app". Would they try to look for `foo` in services? Would they just open 1234 port manually in `ports`? Or would they click on all 4 buttons and look if there is no `foo` option first?

KKoukiou:
> While splitting it into tabs makes sense, I agree that it makes much harder to actually see what any given zone does. 

Indeed, I agree. But it's hard to extend the page with more features with the current design. Zones have even more properties that we don't display in the UI right now, like rich rules, port forwarding and others.

>  For example I want to know if I have opened port 90. I'll need to go through 4 tabs to know if it is not opened.

> Also do we assume users are fully aware of diferences between services, ports, interfaces ans sources? So if someone is asked to "Open port 1234 for the foo app". Would they try to look for `foo` in services? Would they just open 1234 port manually in `ports`? Or would they click on all 4 buttons and look if there is no `foo` option first?

This is exactly the question I had in mind when writing this. Someone tells you 'Open port 1234 for the foo app'. Why would you ever click on the 'Add service' button? If foo is a systemd service, you might get this right, but I think for custom apps this will be for the users a bit of a confusion. 

Well in the end it does not really matter if they end up opening the port of the sevice or adding the service itself. The latter is IMO nicer, but a user who goes with the first option would not get any harm either.

Also don't forget that for checking if a port 1234 is open, one needs to check all zones that are active. It's already not a trivial on blick thing. Maybe a global page seach for service or port would be helpful in this case. 

After all the discussion here, I feel that indeed the current design has some benefits. But the issue with it not being extendable still exists. Users can't edit custom ports, can't edit interfaces and addresses. If we aggree that all of the our out of our future supported features then let's keep the current design, maybe with some button renames. Otherwise, any feature addition with require redesign of this page.

garrett:
> Someone tells you 'Open port 1234 for the foo app'. Why would you ever click on the 'Add service' button?

That's why they have been in the same dialog, with service as default, to promote people adding a service instead of a one-off arbitrary port.

The UI can certainly change, but the high-level idea of guiding someone to use a service instead of a custom port should remain.... somehow, even if it looks completely different.

> it does not really matter if they end up opening the port of the sevice or adding the service itself

Yes and no. Technically, it's the same end effect; this is true.

However, adding the service future-proofs it in case the service needs more ports or needs to be a different port (it does happen sometimes, albeit rarely). And, more importantly, you can easily tell _why_ something is open in the firewall view. This is especially important for teams of admins _or_ forgetful admins (which everyone is at some point, to some degree).

> After all the discussion here, I feel that indeed the current design has some benefits. But the issue with it not being extendable still exists.

I really appreciate all the thought and effort you put into this PR, @KKoukiou! It looks nice and you're solving problems with the current UI that do definitely exist. It makes me really happy to see your initiative and design work here! :sparkles: :tada:  :sparkles: 

I think we can blend what's in this PR and what we currently have to have a hybrid that's better. There are definitely shortcomings and strengths of both.

I wonder if some of these pieces can be split out to immediately improve what we have in Cockpit.

septatrix:
I would really like the ability to add interfaces and IP ranges to zones which was added here. By the looks of it there does not seem to be any follow up PR or any issues tracking some of the improvements brought forward here (except maybe #11904) thought he improvement still seem worthwhile. So should I create a new issue for adding interfaces and IP ranges to existing zones?

septatrix:
x-ref #18018

PR #15727, opened 2021-04-21
Restore package-lock.json from cached dist tarballs

martinpitt:
Commit d3f51710dda moved from node_modules/.stamp to package-lock.json
as our canonical stamp file, which regressed make_dist. This fixes
running autogen/configure/make dist with pre-built dist again, as it
does not pointlessly rebuild everything again.

In autogen.sh, don't run npm install if package-lock.json already
exists.

----

These two are the harmless/obvious changes from PR #15720.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -39,7 +39,7 @@ if [ -z "$(which glib-gettextize)" ]; then
     exit 1
 fi
 
-tools/npm-install
+[ -e package-lock.json ] || tools/npm-install
 
 rm -rf autom4te.cache
 
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -68,7 +68,7 @@ def download_dist(wait=False):
         message("make_dist: not a git repository")
         return None
 
-    if subprocess.call(["git", "diff", "--quiet", "--", ":^test"]) > 0:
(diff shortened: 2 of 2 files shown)


martinpitt:
Nope, happens in #15729 as well [log](https://logs.cockpit-project.org/logs/pull-15729-20210421-060458-631578b1-debian-stable/log.html).

martinpitt:
No-change force push (just rebase to current master) after reverting centos-7 image, to not completely ruin our test statistics.

allisonkarlitskaya:
Probably what you actually want to do here is to `touch package-lock.json`?  The `package-lock.json` that we restore out of the cache will be full of entries for stuff that's not actually in `node_modules`...

We may also just want to rethink using this file as the stamp.

allisonkarlitskaya (approved):

martinpitt:
@allisonkarlitskaya : I'm not sure about that, TBH. Right now this is still how we want the real release dist tarballs to look like. They should contain a package-lock.json for reproducibility/license reasons, and a reduced node_modules/ for running tests without having to npm install. It's a bit awkward indeed. I hope that will somewhat resolve itself once we stop actually committing/using dist *tarballs* for the "quick CI/local testing" purpose, which will be my next PR after #15720.

PR #15729, opened 2021-04-21
build: Fix `make install` failure with --disable-doc

### LINKED ISSUE #15710: make install fails with --disable-doc: No rule to make target 'dist/guide/html/index.html', needed by 'install-data-local'
Cockpit version: xxx
OS: openSUSE Leap 15.3

Hi, I am trying to build cockpit for oS Leap 15.3 (because it is not in the repos).
It appears that makes cannot install that 'dist/guide/html/index.html' file for some reason;
I have tried multiple different --prefix but the same issue occur

Steps to reproduce

1. Install openSUSE leap 15.3
2. Install the toolchain / tools needed to build cockpit
3. Configure it, compile it and try to install it to any directory prefix

image
image

martinpitt:
Move the custom dist-hook bits into `if ENABLE_DOC`, as these files
don't exist with `--disable-doc`.

Fixes #15710

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -295,13 +295,6 @@ include doc/man/Makefile-man.am
 dist-doc-hook:
 	@true
 
-else
-
-dist-doc-hook:
-	@echo "*** doc must be enabled (ie: --enable-doc) in order to make dist or distcheck"
-
-endif
-
 dist-hook:: dist-doc-hook
 	@true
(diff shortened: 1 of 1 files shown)


martinpitt:
No-change force push after reverting centos-7 image, to not completely ruin our test statistics.

marusak (approved):
Danke!

PR #15732, opened 2021-04-22
Fix two coverity errors

martinpitt:
Spotted by coverity:

    src/common/cockpitjson.c:1073: cond_notnull: Condition "copy", taking true branch. Now the value of "copy" is not "NULL".
    src/common/cockpitjson.c:1076: notnull: At condition "copy", the value of "copy" cannot be "NULL".
    src/common/cockpitjson.c:1076: dead_error_condition: The condition "copy" must be true.
    src/common/cockpitjson.c:1076: dead_error_line: Execution cannot reach the expression "json_object_ref(object)" inside this statement: "json_node_take_object(new_n...".

### DIFF
--- a/src/common/cockpitjson.c
+++ b/src/common/cockpitjson.c
@@ -1059,7 +1059,7 @@ cockpit_json_walk_node (JsonNode    *node,
         if (copy)
           {
             new_node = json_node_new (JSON_NODE_ARRAY);
-            json_node_take_array (new_node, copy ?: json_array_ref (array));
+            json_node_take_array (new_node, copy);
           }
 
         break;
@@ -1073,7 +1073,7 @@ cockpit_json_walk_node (JsonNode    *node,
         if (copy)
           {
             new_node = json_node_new (JSON_NODE_OBJECT);
-            json_node_take_object (new_node, copy ?: json_object_ref (object));
+            json_node_take_object (new_node, copy);
           }
 
         break;
--- a/src/common/cockpitwebcertificate.c
+++ b/src/common/cockpitwebcertificate.c
@@ -99,6 +99,7 @@ cockpit_certificate_locate (bool missing_ok,
 {
   const char * const *dirs = cockpit_conf_get_dirs ();
 
+  assert (error != NULL);
   assert (*error == NULL);
 
   for (int i = 0; dirs[i]; i++)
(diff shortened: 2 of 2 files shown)


martinpitt:
Ugh, [GitHub workflows](https://github.com/cockpit-project/cockpit/pull/15732/checks?check_run_id=2408260405) just broke, they can't start our unit-tests container any more. That's going to be some "fun"..

src/common/cockpitjson.c
```diff
@@ -1059,7 +1059,7 @@ cockpit_json_walk_node (JsonNode    *node,
         if (copy)
           {
             new_node = json_node_new (JSON_NODE_ARRAY);
-            json_node_take_array (new_node, copy ?: json_array_ref (array));
+            json_node_take_array (new_node, copy);
```
allisonkarlitskaya, line 1062: weird!  I have a memory of having already fixed this.......

src/common/cockpitwebcertificate.c
```diff
@@ -99,6 +99,7 @@ cockpit_certificate_locate (bool missing_ok,
 {
   const char * const *dirs = cockpit_conf_get_dirs ();
 
+  assert (error != NULL);
```
allisonkarlitskaya, line 102: sure...

src/common/cockpitwebcertificate.c
```diff
@@ -117,7 +118,7 @@ cockpit_certificate_locate (bool missing_ok,
         return cert_path;
     }
 
-  if (error && !missing_ok)
```
allisonkarlitskaya, line 120: okay.  This one is really wrong, though.  We just added an assert for `error != NULL` above.  Surely when we re-run the scan, it'll be happy enough now?

allisonkarlitskaya (changes requested):

src/common/cockpitwebcertificate.c
martinpitt: I'm afraid I don't understand -- coverity's objection was that `error &&` is always true here, so it's a dead condition.

allisonkarlitskaya: Uh.  `EBRAIN`.

allisonkarlitskaya (approved):

PR #15733, opened 2021-04-22
workflows: Run unit-tests container as root

martinpitt:
On the latest GitHub 20.04 image, podman can't pull our unit-tests
container any more:

    Error committing the finished image: error adding layer with blob "sha256:...":
    Error processing tar file(exit status 1): operation not permitted

Let's run the containers as root. This does not matter much for the
host, and until podman is officially supported in Ubuntu, system podman
has fewer potential permission problems.

----

These spontaneously started to fail today, like [build-dist here](https://github.com/cockpit-project/cockpit/pull/15732/checks?check_run_id=2408260405) or [unit-tests here](https://github.com/cockpit-project/cockpit/pull/15732/checks?check_run_id=2408145586).

[tested build-dist on my fork](https://github.com/martinpitt/cockpit/actions/runs/773717557), but not unit-tests. Let's do that here.

### DIFF
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -19,7 +19,7 @@ jobs:
           ref: "${{ github.event.pull_request.head.sha || github.sha }}"
 
       - name: Run unit-tests container
-        run: containers/unit-tests/start dist
+        run: sudo containers/unit-tests/start dist
 
       - name: Create dist tarball artifact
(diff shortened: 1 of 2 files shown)


marusak (approved):
Thanks!
This smells like regression though, maybe worth reporting?

martinpitt:
@marusak : Yes, I'll look at this more closely, I just wanted to quickly unbreak our CI.

PR #15735, opened 2021-04-22
metrics: sticky header for day selector should be after the sticky breadcrumb section

### LINKED ISSUE #15435: Bar with day selector and labels is not sticky on metrics page
The bar containing day selector and names of columns should be sticky when scrolling down on the page.
It might be similar to #15333 but naively applying the same fix did not help.

There is css for it, but does not work: (from `pkg/metrics/metrics.scss`)
```
        &-sticky {                                                              
            position: sticky;                                                   
            top: 0;                                                             
            z-index: 99;                                                        
        }         
```

KKoukiou:
Closes #15435

I left some space between the two sticky bars, thought it makes sense, but I am not sure.
Screen Shot 2021-04-23 at 17 18 29

### DIFF
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -353,7 +353,7 @@
 
         &-sticky {
             position: sticky;
-            top: 0;
+            top: 3.75rem;
             z-index: 99;
         }
     }


marusak:
That number seems magic. Isn't it some PF variable or something?
And the gap looks weird. I think there should not be any gap, possibly just thin 1px line (like border).

KKoukiou:
> That number seems magic. Isn't it some PF variable or something?
> And the gap looks weird. I think there should not be any gap, possibly just thin 1px line (like border).

Well it's try and error. The thing is that top just takes the distance of the sticky header from the top. If you set top: 0, it will have distance 0 from the top, and will be hidden behind the other sticky item, in this case the breadcrumb, because the latter has bigger zindex. So we just need to add some distance from the top, for not a gap we need the height of the element of the breadcrumb header. (Which I have no idea how to get).
Let's allow @garrett to solve this then.

KKoukiou:
@garrett I need your help here to get the right top variable. Danke.

garrett:
Oh, we definitely should not hardcode the offset. Perhaps the document order/nesting is wrong?

garrett:
(BTW: In this case, definitely is "soft", as it's something we should _not_ do, but sticky is hard to deal with and sometimes there isn't a correct way to do it, so sometimes a hack is needed, which is the hardcoded offset. I'll see what I can do to avoid an offset, but if we have to do it, then we have to do it. But we can try to figure out what it is (instead of a magic number) and document it if so.)

garrett:
Oh, I couldn't get the day line sticking too. I thought that's what it was about for a while. It would be nice to get that to stick, but it's super hard without data... so I have to build it, wait for a while until there's a header, and try again. And then build another VM, wait forever again, and so on. :facepalm: 

Anyway, I'll look at this again, to get the actual two sticky bars aligned properly, if possible.

marusak:
You can use our VMs. Any test that uses `prepareArchive()` will have some historic data.

garrett:
> You can use our VMs. 

I have been. That's the problem. I can't see it locally (even with PCP installed with `rpm-ostree install`), so I have to use VMs. And the VMs don't have a history yet, as they're freshly created.

> Any test that uses prepareArchive() will have some historic data.

Yeah, I have to figure that out again, somehow. Edit some file, add that somewhere in the file, and run it somehow. And then it opens it in Chromium, so then I have to figure out how to debug sticky in Chrome's devtools.

marusak:
> Edit some file, add that somewhere in the file, and run it somehow.

Just add `sit()` in test where you expect it to be set up.

> And then it opens it in Chromium, so then I have to figure out how to debug sticky in Chrome's devtools.

Use `TEST_BROWSER=firefox` to open firefox and not chrome. (Requires firefox nightly)

KKoukiou:
Closing this as i have no idea how to solve it better. @garrett you have the issue open, feel free to pick it up when you feel like.

PR #15740, opened 2021-04-23
overview: Don't render empty elements

### LINKED ISSUE #15738: Overview cards are wrongly arranged
If there is no login message or you just close it, cards are incorrectly arranged.
cardsaremisaligned
The empty space is used for `<div id="login-messages" empty="yes"></div>`
This is fallback from f3e0873c introduced in https://github.com/cockpit-project/cockpit/pull/15638

We should fix this before next release

marusak:
Since this is part of grid it would create empty slots.

Fixes #15738

### DIFF
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -129,7 +129,7 @@ class LoginMessages extends React.Component {
         }
 
         if (!last_login_message && !fail_count_message && !last_fail_message)
-            return (<div id='login-messages' empty='yes' />); // for testing
+            return null;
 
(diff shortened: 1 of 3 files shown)


KKoukiou (changes requested):
In overview.scss there is the:

```
131     &__footer {                                                                 
132       &:empty {                                                                 
133         display: none;                                                          
134       }                                                                         
135     }                                                                           
136   }
```

which is not deprecated.

marusak:
> In overview.scss there is the:

Aha, let me take a look if we can remove this safely.

marusak:
Removed. Thanks for catching this!

KKoukiou (approved):

PR #15745, opened 2021-04-24
specfile: Change order of selinux post action

marusak:
Moving above to ensure that service is executed with the proper label.

See https://github.com/cockpit-project/cockpit/pull/15707#discussion_r619292665

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -526,17 +526,17 @@ fi
 %endif
 
 %post ws
-%tmpfiles_create cockpit-tempfiles.conf
-%systemd_post cockpit.socket cockpit.service
-# firewalld only partially picks up changes to its services files without this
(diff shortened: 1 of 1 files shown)


martinpitt (approved):

marusak:
`make_dist: https://github.com/cockpit-project/cockpit-dist/raw/master/a49d8c62c7020bc3648789104d97943b24056292.tar not yet available, waiting...`
packit does not see archive. I'll land, does not seem related.

martinpitt:
Indeed, [publish-dist failed](https://github.com/cockpit-project/cockpit/runs/2426372450?check_suite_focus=true). Some weird race condition with multiple runs. But I'm reworking this anyway, I'm about 80% on my way to use the (much better and collision resistant) cockpit-machines approach for cockpit.

PR #15748, opened 2021-04-26
Rework pre-built dist upload/download

martinpitt:
In publish-dist, directly commit dist/ and related files to the -dist
repo (the ones which make_dist actually needs), and tag the commit with
the SHA, so that they are still easy to look up.  Drop the "old tarball"
removal. This uses git in a much more efficient way, as it now only
stores the deltas. In particular, for commits/PRs which don't change
src/, that will just add an additional tag to the previous build and
store no extra files at all. This avoids regularly overflowing the -dist
repo and making `git push` hang.

Simplify build-dist to do away with the intermediate tarball, as there
are not so many files to make upload-artifact inefficient.

Push every build as an individual tag only, without having them on a
branch. That avoids the need for force-pushing and history rewriting,
which fixes the race of publish-dist against prune-dist.  There is also
no more race condition between parallel publish-dist workflows.

Change prune-dist to keep the tags of the last 7 days plus current
master HEAD, and let garbage collection take care of cleaning up older
ones.

Rework make_dist to get the dist repo in ~/.cache/cockpit-dev/, so that
it persists across project checkouts, and shallow-fetch the desired SHA
build. Unpacking requires some care: Unfortunately our dist tarball
generation does not deal with sub-second file mtime differences, which
can invert the relative mtime of package-lock.json vs.
dist/*/manifest.json; to avoid this, only touch the webpack-built files,
but leave the `npm install` ones alone.

Taken from cockpit-machines:
https://github.com/cockpit-project/cockpit-machines/commit/612bb6aa8d1f
https://github.com/cockpit-project/cockpit-machines/commit/ae57e6249397

-----

I tested this on my fork. The [dist repo](https://github.com/martinpitt/cockpit-dist) now has two tags, and otherwise just an empty master branch. [build-dist](https://github.com/martinpitt/cockpit/actions/runs/784970139) and [publish-dist](https://github.com/martinpitt/cockpit/actions/runs/784999140) look as expected. I also ran [prune-dist](https://github.com/martinpitt/cockpit/runs/2436773139?check_suite_focus=true) -- of course there are not yet any tags older than a week, but it shows that it correctly recognized current master HEAD and the two tags.

The difference to the cockpit-machines workflows is now minimal. I will send a follow-up workflow to c-machines to further reduce it.

Downloading also works fine. I tested this with:

    git clean -ffdx; ln -s ../bots; GITHUB_BASE=martinpitt/cockpit test/image-prepare -qv fedora-34

```
[...]
From https://github.com/martinpitt/cockpit-dist
 * tag               sha-8b214e16d7ea3088131209567212d7aa29af5f59 -> FETCH_HEAD
make_dist: Extracting cached node_modules...
make_dist: Extracting cached package-lock.json...
make_dist: Extracting cached dist...
make_dist: Extracting cached tools/debian/copyright...
```
This uses the cached download, and then making the tarball is rather fast (~ 10 s).

### DIFF
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -15,12 +15,18 @@ jobs:
           # need this to also fetch tags
           fetch-depth: 0
 
-      - name: Run unit-tests container
-        run: sudo containers/unit-tests/start dist
+      - name: Run unit-tests container to build dist
+        run: |
+          sudo containers/unit-tests/start dist
+          tar --strip-components=1 -xf tmp/dist/cockpit-*.tar.xz
 
-      - name: Create dist tarball artifact
+      - name: Create dist artifact
         uses: actions/upload-artifact@v2
         with:
           name: dist
-          path: tmp/dist/cockpit-*.tar.xz
-          retention-days: 7
+          path: |
+            dist/
+            node_modules/
+            package-lock.json
+            tools/debian/copyright
+          retention-days: 1
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -14,9 +14,6 @@ jobs:
         with:
           # need this to also fetch tags
           fetch-depth: 0
-          # avoid checking out the default pull/NNN/head ref, as that gives
-          # mismatching SHAs in generated tarballs
-          ref: "${{ github.event.pull_request.head.sha || github.sha }}"
 
       - name: Run unit-tests container
         run: sudo containers/unit-tests/start dist
@@ -24,8 +21,6 @@ jobs:
       - name: Create dist tarball artifact
         uses: actions/upload-artifact@v2
         with:
-          # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions
-          # for pull_requests, use HEAD of the proposed branch, for pushes to origin the current SHA
-          name: "dist-${{ github.event.pull_request.head.sha || github.sha }}"
+          name: dist
           path: tmp/dist/cockpit-*.tar.xz
           retention-days: 7
--- a/.github/workflows/prune-dist.yml
+++ b/.github/workflows/prune-dist.yml
@@ -1,4 +1,4 @@
-# truncate the cockpit-dist history every Sunday night, to avoid unbounded growth
+# truncate the -dist history every Sunday night, to avoid unbounded growth
 name: prune-dist
 on:
   schedule:
@@ -13,7 +13,6 @@ jobs:
       - name: Set up configuration and secrets
         run: |
           printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
-          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
           # we push to -dist repo via https://github.com, that needs our cockpituous token
           git config --global credential.helper store
           echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
@@ -22,18 +21,23 @@ jobs:
         run: |
           git clone https://github.com/${{ github.repository }}-dist.git dist-repo
 
-      - name: Truncate history of -dist repo
+      - name: Delete old tags
         run: |
           set -ex
+
+          # head commit SHA of default branch
+          HEAD=$(git ls-remote 'https://github.com/${{ github.repository }}' main master | cut -f1 | head -n1)
+
           cd dist-repo
-          # oldest commit that applies to any still present tarball
-          REF=$(git log --pretty=format:%H * | tail -n1)
-
-          git checkout --orphan temp $REF
-          git commit -m "Truncated history"
-          git rebase --onto temp $REF master
-          git branch -D temp
-          git reflog expire --expire=now --all
-
-      - name: Force-push -dist repo
-        run: git -C dist-repo push -f
+          now="$(date +%s)"
+          for tag in $(git tag -l); do
+              tag_time="$(git show -s --format=%at $tag)"
+              if [ "$tag" = "sha-${HEAD}" ]; then
+                  echo "$tag refers to current project default branch HEAD, keeping"
+              elif [ $((now - tag_time)) -ge 604800 ]; then
+                  echo "$tag is older than 7 days, deleting..."
+                  git push origin ":$tag"
+              else
+                  echo "$tag is younger than 7 days, keeping"
+              fi
+          done
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -8,6 +8,7 @@ jobs:
   run:
     if: github.event.workflow_run.conclusion == 'success'
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Download build-dist artifacts
         # 2.14.0; if you update this, audit the diff and ensure that it does not leak/abuse secrets
@@ -16,52 +17,23 @@ jobs:
           name: dist
           workflow: build-dist
           run_id: ${{ github.event.workflow_run.id }}
-          path: dist
+          path: dist-repo
 
       - name: Set up configuration and secrets
         run: |
           printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
-          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
           # we push to -dist repo via https://github.com, that needs our cockpituous token
           git config --global credential.helper store
           echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
 
-      - name: Commit dist tarball dist repo
+      - name: Commit to dist repo
         run: |
           set -ex
+          cd dist-repo
+          git init
+          git add .
+          git commit -m "Build for ${{ github.event.workflow_run.head_sha }}"
+          tag='sha-${{ github.event.workflow_run.head_sha }}'
+          git tag "$tag"
 
-          # we need a fully predictable name/URL, and git introduces these
-          # additional numbers into the version; so wrap it in a predictable named tar
-          sha='${{ github.event.workflow_run.head_sha }}'
-
-          # multiple parallel workflows race against each other, so the final
-          # `git push` may fail
-          rc=1
-          for retry in $(seq 5); do
-              git clone https://github.com/${{ github.repository }}-dist.git dist-repo
-              tar -cvf "dist-repo/${sha}.tar" -C dist/
-
-              # freshly created empty repo?
-              cd dist-repo
-              git rev-parse HEAD >/dev/null 2>&1 || git init
-              git add "${sha}.tar"
-              git commit -m "Build for $sha"
-
-              # remove tarballs older than 3 days
-              now=$(date +%s)
-              for f in *.tar; do
-                  fmtime=$(git log --pretty=%at -n1 -- $f)
-                  [ $(($now - $fmtime)) -lt 259200 ] || git rm $f
-              done
-              [ -z "$(git status --short)" ] || git commit -m 'Drop old builds'
-
-              if git push; then
-                  rc=0
-                  break
-              else
-                  echo "ERROR: conflict? Retrying git commit"
-                  rm -rf dist-repo
-              fi
-          done
-          exit $rc
-
+          git push https://github.com/${{ github.repository }}-dist.git "$tag"
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -10,7 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Download build-dist artifacts
-        uses: dawidd6/action-download-artifact@v2
+        # 2.14.0; if you update this, audit the diff and ensure that it does not leak/abuse secrets
+        uses: dawidd6/action-download-artifact@b9571484721e8187f1fd08147b497129f8972c74
         with:
           workflow: build-dist
           run_id: ${{ github.event.workflow_run.id }}
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -13,8 +13,10 @@ jobs:
         # 2.14.0; if you update this, audit the diff and ensure that it does not leak/abuse secrets
         uses: dawidd6/action-download-artifact@b9571484721e8187f1fd08147b497129f8972c74
         with:
+          name: dist
           workflow: build-dist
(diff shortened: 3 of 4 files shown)


martinpitt:
Note that this will slightly regress the time `make_dist` takes, as the extra autogen.sh/make dist run now takes an additional 14 seconds. But the current "store full tarballs in git" approach is not viable, I don't want to regularly recreate the -dist repo because git push times out.

But that's still much better than the ~ 10 minutes that npm install and webpack would take locally. As a little consolation, downloading the cache should now generally be faster than before.

martinpitt:
packit can't succeed here, as the publish-dist changes go together with the build-dist changes, and the former only runs from master.

.github/workflows/publish-dist.yml
```diff
@@ -8,57 +8,34 @@ jobs:
...
-          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
           # we push to -dist repo via https://github.com, that needs our cockpituous token
           git config --global credential.helper store
           echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' >> ~/.git-credentials
 
-      - name: Commit dist tarball dist repo
+      - name: Commit to dist repo
         run: |
           set -ex
+          SHA="$(cat dist-repo/SHA)"
```
allisonkarlitskaya, line 31: This is insecure.

The SHA came from a workflow which is under complete control of a potentially-hostile pull request.  Someone could open a pull request against cockpit, populate our current master branch SHA into this file, and then we'd pick it up here and tag their malicious payload as a valid cache of the current state of our master branch.

This is what I've been spending my week worrying about... I think we need to do better here.

One approach might be to upload things to git only after stuff lands on master, and use some form of artifact-based downloading for testing pull requests...

Another approach might be to use `pull_request_target` and try to come up with a way to harden the build.  For the `node_modules` stuff I'm running in a separate docker container, for example...

allisonkarlitskaya (changes requested):

.github/workflows/publish-dist.yml
martinpitt: Of course it is untrusted, that's what the current master workflow has as well. But that's fine -- us project mainainers must review a PR. It's just a cache, it's exactly the same as locally checking out and building an unreviewed PR.

martinpitt: Discussed in meeting. Will rework this, in a separate PR. Either with:
 * Not storing the SHA in the workflow, but instead querying the github.event.workflow for the originally proposed repo/branch and determining the SHA that way.
 * If that somehow does not work out, restrict publish-dist to PRs whose owners are in cockpit-project. We'd lose packit tests for "outsider" PRs, but that does not hurt that much. (At least until we figure out a better way)

.github/workflows/prune-dist.yml
```diff
@@ -10,10 +10,12 @@ jobs:
   run:
     runs-on: ubuntu-latest
     steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
```
allisonkarlitskaya, line 14: Is this just to get `git rev-parse HEAD` below?  I feel like you could get that a lot easier than cloning everything...

.github/workflows/publish-dist.yml
```diff
@@ -8,57 +8,34 @@ jobs:
   run:
     if: github.event.workflow_run.conclusion == 'success'
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Download build-dist artifacts
         uses: dawidd6/action-download-artifact@v2
```
allisonkarlitskaya, line 14: I'm really going to insist on this point: please pin this down to a specific sha, at the very least.

.github/workflows/publish-dist.yml
```diff
@@ -8,57 +8,34 @@ jobs:
...
-          done
-          exit $rc
+          cd dist-repo
+          git init
+          git add .
+          git commit -m "Build for $SHA"
+          tag="sha-$SHA"
+          git tag "$tag"
 
+          git push https://github.com/${{ github.repository }}-dist.git $tag
```
allisonkarlitskaya, line 41: Usual quoting complaint here:  `$tag` is coming from `$SHA` which is fed to us from an untrusted workflow... this could do some serious damage.

test/make_dist.py
```diff
@@ -52,115 +51,78 @@ def build_dist():
...
-    dists = glob.glob(f"cockpit-*{sha[:8]}*.tar.xz")
-    if dists:
-        message("make_dist: already downloaded", dists[0])
-        return os.path.abspath(dists[0])
+    dist_git_checkout = os.path.join(os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache")), "cockpit-dev", CACHE_REPO)
+
+    if not os.path.exists(dist_git_checkout):
+        message(f"make_dist: Creating dist cache {dist_git_checkout}")
+        os.makedirs(dist_git_checkout)
+        subprocess.check_call(["git", "-C", dist_git_checkout, "init", "--bare", "--quiet"])
```
allisonkarlitskaya, line 79: It probably makes more sense to use `--git-dir` instead of `-C`.  Also: consider passing the directory as an argument to `init` rather.

test/make_dist.py
```diff
@@ -52,115 +51,78 @@ def build_dist():
...
     if subprocess.call(["git", "diff", "--quiet", "--", ":^test", ":^packit.yaml", ":^.github"]) > 0:
         message("make_dist: uncommitted local changes, skipping download")
-        return None
+        return False
 
-    dists = glob.glob(f"cockpit-*{sha[:8]}*.tar.xz")
-    if dists:
-        message("make_dist: already downloaded", dists[0])
-        return os.path.abspath(dists[0])
+    dist_git_checkout = os.path.join(os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache")), "cockpit-dev", CACHE_REPO)
```
allisonkarlitskaya, line 74: I'd expect `+ ".git"` here.

Also: i feel like the entire thing would read a lot easier as an f-string.  `os.path.join()` is a bit silly in places like this and really harms readability.

Also, very very minor gripe: I hate that we call `expanduser` even if `XDG_CACHE_HOME` is set...

test/make_dist.py
```diff
@@ -52,115 +51,78 @@ def build_dist():
...
+    # built files need to have current time, not commit time, to satisfy make dependencies
+    tar_cmd = { False: ["tar", "-x"], True: ["tar", "--touch", "-x"] }
+    for (unpack_path, touch) in [("node_modules", False),
+                                 ("package-lock.json", False),
+                                 ("dist", True),
+                                 ("tools/debian/copyright", True)]:
+        if os.path.exists(unpack_path):
+            continue
+        message(f"make_dist: Extracting cached {unpack_path}...")
+        p_git = subprocess.Popen(["git", "-C", dist_git_checkout, "archive", "FETCH_HEAD", unpack_path],
```
allisonkarlitskaya, line 106: `FETCH_HEAD` is unsafe here: you might be building more than one thing on the same system, sharing the cache directory.  Use the tag.  Also, more `--git-dir`.

test/make_dist.py
```diff
@@ -52,115 +51,78 @@ def build_dist():
...
-        message("make_dist: Extracting from tarball:", ' '.join(unpack_paths))
-        prefix = os.path.basename(tar_path).split('.tar')[0] + '/'
-        prefixed_unpack_paths = [prefix + d for d in unpack_paths]
-        subprocess.check_call(["tar", "--touch", "--strip-components=1", "-xf", tar_path] + prefixed_unpack_paths)
-        # after the above, dist manifests may be newer than copyright due to tar touching order
-        # restore file time stamps as they would be after a build
-        subprocess.check_call(["touch", "tools/debian/copyright"])
-
-    return tar_path
+    # built files need to have current time, not commit time, to satisfy make dependencies
```
allisonkarlitskaya, line 97: I feel like with the order that we unpack things, we could probably *always* touch.  That would be a nice cleanup to some otherwise kinda-evil code.

Even if not: please just write this as a normal conditional inside the loop.

test/make_dist.py
```diff
@@ -52,115 +51,78 @@ def build_dist():
...
+    for (unpack_path, touch) in [("node_modules", False),
+                                 ("package-lock.json", False),
+                                 ("dist", True),
+                                 ("tools/debian/copyright", True)]:
+        if os.path.exists(unpack_path):
+            continue
+        message(f"make_dist: Extracting cached {unpack_path}...")
+        p_git = subprocess.Popen(["git", "-C", dist_git_checkout, "archive", "FETCH_HEAD", unpack_path],
+                                 stdout=subprocess.PIPE)
+        subprocess.check_call(tar_cmd[touch], stdin=p_git.stdout)
```
allisonkarlitskaya, line 108: almost feels like using `check_call` with `shell=True` might be easier.  Your call.

test/make_dist.py
```diff
@@ -52,115 +51,78 @@ def build_dist():
...
+        assert p_git.wait() == 0
```
allisonkarlitskaya, line 109: Does python guarantee that the expression after an `assert` is executed?  We need this to reap the zombie...

allisonkarlitskaya (changes requested):

.github/workflows/build-dist.yml
```diff
@@ -15,17 +15,25 @@ jobs:
...
           # avoid checking out the default pull/NNN/head ref, as that gives
-          # mismatching SHAs in generated tarballs
+          # mismatching SHAs in generated artifacts
           ref: "${{ github.event.pull_request.head.sha || github.sha }}"
 
-      - name: Run unit-tests container
-        run: sudo containers/unit-tests/start dist
+      - name: Run unit-tests container to build dist
+        run: |
+          sudo containers/unit-tests/start dist
```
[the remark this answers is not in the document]
martinpitt: Unfortunately no, see commit 8917ad9c77a1c6016d1463785af96c5dfd3a1c67

test/make_dist.py
martinpitt: I did start out with "always touch", but that really gets in the way. `make` is not good with files that have sub-second resolution time differences, this leads to strange and not always 100% reliably reproducible failures, at least on my system. This does not happen during a "real" build, as building a webpack or npm install take way more than one second. The only alternative that I found was to add `time.sleep(1)` in between, but I found that way more ugly.

test/make_dist.py
martinpitt: I really don't like running dynamically built shell commands, especially not if they are constructed from user-settable variables such as `$GITHUB_BASE`, or may have paths with spaces.

test/make_dist.py
martinpitt: I had some trouble with the tag, as `git fetch origin <tagname>` does not quite fetch the actual tag, you get "not a valid object name: sha-ba9...". Indeed refs/tags/ is empty. And `git fetch --tag origin <tagname>` fetches *all* tags. Your PR #15700 has exactly the same problem. I'll do some googling/experimentation.

.github/workflows/publish-dist.yml
martinpitt: Fixed in separate commit, similar to https://github.com/cockpit-project/cockpit-machines/pull/143

.github/workflows/prune-dist.yml
martinpitt: Indeed. Replaced with `git ls-remote`.

.github/workflows/publish-dist.yml
martinpitt: Fixed.

test/make_dist.py
martinpitt: Thanks! Done.

test/make_dist.py
martinpitt: Added .git. I kept os.path.join, I prefer explicit path components over the more brittle string interpolation, but I agree that this is just a personal whim.

test/make_dist.py
martinpitt: I'm not sure -- I split it, to be on the safe side.

test/make_dist.py
martinpitt: I replaced the fetching with the magic incantation to *actually* fetch the tag, now it works fine. I also followed up to your PR to do the same there.

martinpitt:
Re-tested on my fork: [publish-dist succeeded](https://github.com/martinpitt/cockpit/runs/2490029929?check_suite_focus=true), it [pushed the new tag correctly](https://github.com/martinpitt/cockpit-dist/tags), [prune-dist](https://github.com/martinpitt/cockpit/runs/2490046703?check_suite_focus=true) correctly recognized current master HEAD, and `GITHUB_BASE=martinpitt/cockpit test/make_dist.py -d` downloads the cache and builds a tar in ~ 10 seconds.

test/make_dist.py
martinpitt: BTW, the commit message does explain that.

martinpitt: I also added a comment right here.

.github/workflows/publish-dist.yml
martinpitt: Done in a separate commit, like in cockpit-machines in https://github.com/cockpit-project/cockpit-machines/pull/143

test/make_dist.py
```diff
@@ -52,115 +51,82 @@ def build_dist():
...
-                    tario.write(block)
-
-                # clear the download progress in tty mode
-                if total_size is not None:
-                    sys.stderr.write("\r                             \r")
-
-                break
-
-        except urllib.error.HTTPError as e:
+            subprocess.check_call(["git", "--git-dir", dist_git_checkout, "fetch", "--no-tags", "--depth=1", "origin", "tag", tag])
```
allisonkarlitskaya, line 86: Didn't know about `tag <tag>`.  cool.

.github/workflows/publish-dist.yml
```diff
@@ -8,57 +8,32 @@ jobs:
...
-                  rm -rf dist-repo
-              fi
-          done
-          exit $rc
-
+          cd dist-repo
+          git init
+          git add .
+          git commit -m "Build for ${{ github.event.workflow_run.head_sha }}"
+          tag='sha-${{ github.event.workflow_run.head_sha }}'
```
allisonkarlitskaya, line 36: :grinning: 

Same comment about single quotes as for c-machines, if you care.

allisonkarlitskaya (approved):
There are so many wins here...

Thanks for dealing with all the bickering :)

.github/workflows/publish-dist.yml
martinpitt: Yup, just sent https://github.com/cockpit-project/cockpit-machines/pull/145 to synchronize/cleanup the workflows.

martinpitt:
I deleted and re-initialized https://github.com/cockpit-project/cockpit-dist , so that we get rid of all the old cruft.

PR #15749, opened 2021-04-26
Restyle language menu to select

garrett:
The language selection dialog now looks like this in the PR:
Screenshot_2021-04-26 Esittely - garrett Rain

I've worked on interactions a little too. Check out background override with the `:focus-within` selector. It improves both keyboard and mouse usage, keeping the background highlight on the selected item even when something else is being selected or when the focus is outside of the list.

### DIFF
--- a/pkg/shell/shell.scss
+++ b/pkg/shell/shell.scss
@@ -127,10 +127,46 @@ html.index-page body {
         max-height: 20rem;
         overflow: auto;
     }
-}
 
-#display-language-list {
-    margin-bottom: var(--pf-global--spacer--sm);
+    p:not(:last-child) {
+        // Since we're using PF3 and it redefines several base elements, reset it to what PF4 does by default
+        margin-bottom: var(--pf-global--spacer--md);
+    }
+
+    // PF overrides to fake a multiselect widget (as one does not currently exist in PF4).
+    // A menu gives us the interaction we want, but the styling is a bit off.
+    // Therefore, we're changing the visuals here locally.
+    // PF4 upstream request for multi-select @ https://github.com/patternfly/patternfly/issues/4027
+    .pf-c-menu {
+        // It's  not really a menu, so it shouldn't have a shadow
+        box-shadow: none;
+
+        // Divider is silly between the widgets in this context
+        .pf-c-divider {
+            display: none;
+        }
+
+        &__content {
+            // An overflow multi-select widget needs an outline
+            border: 1px solid var(--pf-global--BorderColor--100);
+            // There should be minimal space between the widgets (replacing the divider)
+            margin-top: var(--pf-global--spacer--sm);
+        }
+
+        // Search should not be inset when there's no border containing it
+        &__search {
+            padding: 0;
+        }
+
+        // Keep the background on a selected item even when it doesn't have
+        // focus, allowing keyboard control to have the only background color
+        // when active but also keep the background color when the list loses
+        // focus (such as when the keyboard or mouse navigates outside,
+        // including initial rendering of the list.
+        &__list:not(:focus-within) .pf-m-selected {
+            background-color: var(--pf-c-menu__list-item--hover--BackgroundColor);
+        }
+      }
 }
 
 iframe.container-frame {
--- a/pkg/shell/shell.scss
+++ b/pkg/shell/shell.scss
@@ -128,11 +128,6 @@ html.index-page body {
         overflow: auto;
     }
 
-    p:not(:last-child) {
-        // Since we're using PF3 and it redefines several base elements, reset it to what PF4 does by default
-        margin-bottom: var(--pf-global--spacer--md);
-    }
-
     // PF overrides to fake a multiselect widget (as one does not currently exist in PF4).
     // A menu gives us the interaction we want, but the styling is a bit off.
     // Therefore, we're changing the visuals here locally.
--- a/pkg/shell/shell.scss
+++ b/pkg/shell/shell.scss
@@ -127,41 +127,6 @@ html.index-page body {
         max-height: 20rem;
         overflow: auto;
     }
-
-    // PF overrides to fake a multiselect widget (as one does not currently exist in PF4).
-    // A menu gives us the interaction we want, but the styling is a bit off.
-    // Therefore, we're changing the visuals here locally.
-    // PF4 upstream request for multi-select @ https://github.com/patternfly/patternfly/issues/4027
-    .pf-c-menu {
-        // It's  not really a menu, so it shouldn't have a shadow
-        box-shadow: none;
-
-        // Divider is silly between the widgets in this context
-        .pf-c-divider {
-            display: none;
-        }
-
-        &__content {
-            // An overflow multi-select widget needs an outline
-            border: 1px solid var(--pf-global--BorderColor--100);
-            // There should be minimal space between the widgets (replacing the divider)
-            margin-top: var(--pf-global--spacer--sm);
-        }
-
-        // Search should not be inset when there's no border containing it
-        &__search {
-            padding: 0;
-        }
-
-        // Keep the background on a selected item even when it doesn't have
-        // focus, allowing keyboard control to have the only background color
-        // when active but also keep the background color when the list loses
-        // focus (such as when the keyboard or mouse navigates outside,
-        // including initial rendering of the list.
-        &__list:not(:focus-within) .pf-m-selected {
-            background-color: var(--pf-c-menu__list-item--hover--BackgroundColor);
-        }
-      }
 }
 
 iframe.container-frame {
--- a/pkg/lib/menu-select-widget.scss
+++ b/pkg/lib/menu-select-widget.scss
@@ -1,0 +1,35 @@
+// FIXME: Remove this custom implementation once a component exists upstream.
+// PF overrides to fake a multiselect widget (as one does not currently exist in PF4).
+// A menu gives us the interaction we want, but the styling is a bit off.
+// Therefore, we're changing the visuals here locally.
+// PF4 upstream request for multi-select @ https://github.com/patternfly/patternfly/issues/4027
+.ct-menu-select-widget.pf-c-menu {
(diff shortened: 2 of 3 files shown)


garrett:
Upstream issue for this local override is at:
https://github.com/patternfly/patternfly/issues/4027

We could move this override to be at the PF4 override level if you prefer, but as the widget doesn't exist in PF, this isn't overriding a PF widget, but making a new widget by faking the style on top of the wrong widget. And since it's such a hack and we're only using this on the shell's dialog level, I kept it here.

Otherwise, we could transplant the code under `lib/` and use `ct-listselect` instead of `display-language-modal`. What do you think?

garrett:
(BTW: The screenshot is not in French. It's in Finnish, switching to French. :rofl:)

martinpitt:
Current master for comparison:
image

martinpitt (dismissed):
Thanks @garrett , this LGTM! I can't find any keyboard a11y improvement -- both current master and this PR are equally broken in current firefox 88 at least -- I can't change the selection with cursor keys, space, or return, I can only scroll around. But indeed it looks better now, taking less wasted space.

pkg/shell/shell.scss
```diff
@@ -127,10 +127,46 @@ html.index-page body {
...
-    margin-bottom: var(--pf-global--spacer--sm);
+    .pf-c-menu {
+        // The widget needs a little spacing between the info text and itself
+        margin-top: var(--pf-global--spacer--sm);
+    }
+
+    // PF overrides to fake a multiselect widget (as one does not currently exist in PF4).
+    // A menu gives us the interaction we want, but the styling is a bit off.
+    // Therefore, we're changing the visuals here locally.
+    // PF4 upstream request for multi-select @ https://github.com/patternfly/patternfly/issues/4027
```
marusak, line 139: Shouldn't this be in PF overwrites file where we keep all PF css hacks? We want to keep it as easy as possible to manage these hacks so having them on one place make is more transparent.

pkg/shell/shell.scss
```diff
@@ -127,10 +127,46 @@ html.index-page body {
         max-height: 20rem;
         overflow: auto;
     }
-}
 
-#display-language-list {
-    margin-bottom: var(--pf-global--spacer--sm);
+    .pf-c-menu {
+        // The widget needs a little spacing between the info text and itself
+        margin-top: var(--pf-global--spacer--sm);
```
marusak, line 133: I think this could be fixed in some other way using PF directly than adding separate css rule. For example using `<Text>` component instead of `<p>` or possibly using `<Stack>` or so.

KKoukiou:
> Thanks @garrett , this LGTM! I can't find any keyboard a11y improvement -- both current master and this PR are equally broken in current firefox 88 at least -- I can't change the selection with cursor keys, space, or return, I can only scroll around. But indeed it looks better now, taking less wasted space.

@martinpitt can you elaborate what's the expected behavior here? I am using https://www.patternfly.org/v4/components/menu/#option-single-select and the keys work like this:
* enter/space is selecting an item
* arrow keys can be used to move between items
* tab moves to the dialog apply button which it the first selectable element after the menu in the dom

pkg/shell/shell.scss
garrett: I explained it above.

But I'll explain it here too: 

This isn't overriding PF. We're using one widget out of context, in a completely wrong way and we're doing a hack to make it look like a widget that doesn't exist.

pkg/shell/shell.scss
garrett: True. I have it here because I wanted to minimize the changeset and also because I have to change the margin and padding anyway as the _menu_ does its own margin and padding and border (which I turn off) because it thinks it is a menu and not a list selection.

garrett:
> this LGTM! I can't find any keyboard a11y improvement -- both current master and this PR are equally broken in current firefox 88 at least

It works here in Firefox 88 with keyboard navigation. :shrug: 

The point was that the CSS doesn't _break_ keyboard a11y by showing two things highlighted at once. Without my fix, the highlight would be lost when the entry itself doesn't have focus. Your screenshot comparison even shows this. I fixed that, without the duplicated highlight in keyboard focus mode.

pkg/shell/shell.scss
```diff
@@ -127,10 +127,46 @@ html.index-page body {
...
+            display: none;
+        }
+
+        &__content {
+            // An overflow multi-select widget needs an outline
+            border: 1px solid var(--pf-global--BorderColor--100);
+            // There should be minimal space between the widgets (replacing the divider)
+            margin-top: var(--pf-global--spacer--sm);
+        }
+
```
KKoukiou, line 155: There is a tab here which github does not show. If you repost pls remove No need to repost only for this.

garrett: Whoops. I'll fix.

pkg/shell/shell.scss
garrett: It's already in a `<p>` and `<Text>` is just a very roundabout way to make a `<p>` anyway. (Look at the rendered source in PF's examples.)

The problem is that we're including CSS that redefines what `<p>` means. This would affect `<Text>` just the same, as it's just `<p>`.

Look at this screenshot crop from Firefox's dev tools pointed at the `<p>`:

p-problem

There's a reset in the _variables file (is this PF3?) and a margin defined in the glyphicons (again, is this from PF3?)

martinpitt:
> The point was that the CSS doesn't break keyboard a11y by showing two things highlighted at once. 

Yes,  I did understand that, this is why I approved -- it does not make things worse.

@KKoukiou : I figured out what I experienced this morning: If I first use the *mouse* to click into the language list, then I can use the cursor up/down keys to move the gray bar, and space to select the item (with the check mark). However, if you open the dialog and press Tab twice, then it looks like it's focusing the list. However, it's really focusing the "list container", and all that the cursors keys do is to operate the scroll bar. If I press Tab again, *then* it focuses the actual list, and cursor up/down/space works.

This probably makes sense somewhat, that the scrollable area/scrollbar is focusable independently.

martinpitt (approved):
@garrett 's changes from 20 mins ago was just a rebase, still LGTM. Thank you!

garrett:
I have one more minor tweak, to switch the padding from the widget to the `<p>` above. I can't really use PF4 overrides to do this, as it will most definitely change assumptions across the site, and the problem is with PF3 (with it overwriting things), not PF4. The solution is to drop PF3 whenever we can, but that will take time.

Meanwhile, this fix would still work with PF3 dropped if it's redefining what PF4 does natively anyway.

It stinks, but this is the least bad solution until we drop PF3.

garrett:
My latest little tweak was to move the margin-top of the pf-c-menu to be a margin-bottom of p:not(:last-child), which is what PF4 does. As it doesn't have access to pf-c-*, I changed it to use pf-global spacer of md, which is what it resolves to in PF4 anyway.

https://github.com/cockpit-project/cockpit/pull/15749/commits/029d96b351b4d5e87dfe96b3d779af3c54df8906#diff-afd101e8a0fe38bd176fb79bb31c3c02021231050251a934e1db390d1bc3f9cbR131-R134:
```scss
    p:not(:last-child) {
        // Since we're using PF3 and it redefines several base elements, reset it to what PF4 does by default
        margin-bottom: var(--pf-global--spacer--md);
    }
```

...instead of:
```scss
     .pf-c-menu {
        // The widget needs a little spacing between the info text and itself
        margin-top: var(--pf-global--spacer--sm); 
    }
```

In other words, I'm redefining what PF4 does when it isn't stomped on by PF3.

It's not a more global Cockpit fix as it would _definitely_ affect other pages and I want to contain the scope of it to this PR.

End result is a touch more spacing between the paragraph and the element (md instead of sm), but it still is mainly the same:

Screenshot_2021-04-28 Overview - garrett Rain

garrett:
My point above with @marusak is that even using another PF4 widget (which resolves to the same thing anyway, so let's not) wouldn't even help here. This is yet another PF3 + PF4 issue. They're supposed to work together fine enough, but just do not (in some small ways). It's mainly related to how PF3 resets and redefines several elements that PF4 just does not do, and PF4 builds assumptions on top of them not being modified.

It's a bunch of minor stuff, but it does hit us from time to time.

pkg/shell/shell.scss
KKoukiou: The issue I see with this not being in overrides, is that we can't reuse this component in other pages, since this CSS file is not shared.

Indeed we used to have till now overrides only for bug overrides, but missing components can also go in some global file? At least some CSS in pkg/lib? We can also move it in a follow up.

pkg/shell/shell.scss
KKoukiou: This is not the problem here however. The Text component from PF4 would override the styles seen here from PF4 if used. However there is not bottom margin for the last p child, which is the case here.
The PF4 rule looks like this:

```
.pf-c-content p:not(:last-child) {
    margin-bottom: var(--pf-c-content--MarginBottom);
}
```

Flex would work however.

KKoukiou: Replaced with Flex.

pkg/shell/shell.scss
KKoukiou: Moved it to pkg/lib.

KKoukiou:
@marusak I fixed you comments and sent fix ups, could you take one more look?

pkg/shell/shell.scss
garrett: Sure, in lib/ and we can call it ct-listselect or whatever else you prefer.

KKoukiou (approved):
please squash on merge or else I will squash everything but for the first commit and repush after review.

marusak (approved):
> I fixed you comments and sent fix ups, could you take one more look?

Code wise looks good, thanks!
Do you want to a design review, @garrett ?

garrett:
I'll recheck. Building now.

garrett:
Looks good to me! :+1: 
Screenshot_2021-05-04 Overview - garrett Rain

PR #15752, opened 2021-04-27
storaged: Check for udisks manager proxy failure

martinpitt:
If the client.manager.wait() is rejected, the following initializations
then can't detect any features, which sets client.features = {}. This
tricks the UI into thinking that initialization is complete, but nothing
really works: There are no known features or file systems.

To make this less hilariously hard to debug, do the same as
client.init() does on a wait failure and return `client.features =
false`. This keeps the overview page in the "initialization" empty
state.

If there is a "changed" signal from the manager proxy after that, it
will then re-trigger the initialization properly. Otherwise, the page
remains in "Storage can not be managed on this system", which is more
honest than offering broken dialogs everywhere.

----

Spotted in https://github.com/cockpit-project/bots/pull/1952 where various storage tests fail very often. See
the debugging odyssey [in the ubuntu-stable image refresh](https://github.com/cockpit-project/bots/pull/1952#issuecomment-827310787) and below for details.

The root cause of this was fixed in PR #15761 , but this is still a nice robustification IMHO.

### DIFF
--- a/pkg/storaged/client.js
+++ b/pkg/storaged/client.js
@@ -505,7 +505,15 @@ function init_model(callback) {
     Promise.allSettled([client.manager.wait(),
         client.mdraids.wait(), client.vgroups.wait(), client.drives.wait(),
         client.blocks.wait(), client.blocks_ptable.wait(), client.blocks_lvm2.wait(), client.blocks_fsys.wait()
-    ]).then(() => {
+    ]).then(results => {
+        // we at least need the manager object; if it doesn't exist, wait for the next proxy onchanged
+        if (results[0].status !== 'fulfilled') {
+            console.warn("init_model(): udisks manager proxy failed:", JSON.stringify(results[0].reason));
+            client.features = false;
+            callback();
+            return;
+        }
+
         pull_time().then(function() {
             enable_features().then(function() {
                 query_fsys_info().then(function(fsys_info) {
@@ -841,7 +849,8 @@ client.init = function init_storaged(callback) {
                                         "/org/freedesktop/UDisks2/Manager", { watch: true });
 
     udisks_manager.wait().then(() => init_client(udisks_manager, callback))
-            .catch(() => {
+            .catch(ex => {
+                console.warn("client.init(): udisks manager proxy failed:", JSON.stringify(ex));
                 client.features = false;
                 callback();
             });


martinpitt:
Triggering an additional test against the ubuntu-stable refresh in https://github.com/cockpit-project/bots/pull/1952 , as that is very prone to these errors.

martinpitt:
With this fix, I ran `TestStorageMounting.testMountOptions` in two parallel loops, and got through maybe 50 iterations, much longer than it ever held up without the fix. I still got one failure eventually where there was  no "changed" event any more after a failed proxy initialization, which then resulted in a "Storage cannot be managed on this system" empty state. This is at least much more explicit than showing a half-broken/empty page, and easier for creating naughties.

martinpitt:
On CI it looks [much worse](https://logs.cockpit-project.org/logs/pull-15752-20210427-082756-e7814bcd-ubuntu-stable-bots-1952/log.html) -- indeed a lot of tests now fail with "udisks manager proxy not ready yet", apparently without ever getting a "changed" signal.

I still like the more explicit failure mode here, but of course it doesn't really help.

I can't find anything wrong with a CLI loop like

    for i in `seq 100`; do systemctl stop udisks2; udisksctl dump |head; sleep 0.1; done

martinpitt:
I enabled dbus debugging around the init part of the test. Interesting parts:

```
error: XXX waiting for udisks failed: [] 
log: XXX inited, features false
log: XXX StoragePage.render; inited true client.features false
```
That's from `udisks_manager.wait().catch()` directly in `client.init()`. This is already unexpected, but our code still handles that gracefully. There are *no* udisks related D-Bus messages before this at all, so the .wait() fails immediately. This is certainly wrong?

After that, a whole lot of D-bus flurry happens for all kinds of services, I filtered these a bit:
```
debug: dbus: {"meta":{"org.freedesktop.UDisks2.Manager":{"methods":{"CanFormat":{"in":["s"],"out":["(bs)"]},"CanResize":{"in":["s"],"out":["(bts)"]},"CanCheck":{"in":["s"],"out":["(bs)"]},"CanRepair":{"in":["s"],"out":["(bs)"]},"LoopSetup":{"in":["h","a{sv}"],"out":["o"]},"MDRaidCreate":{"in":["ao","s","s","t","a{sv}"],"out":["o"]},"EnableModules":{"in":["b"],"out":[]},"EnableModule":{"in":["s","b"],"out":[]},"GetBlockDevices":{"in":["a{sv}"],"out":["ao"]},"ResolveDevice":{"in":["a{sv}","a{sv}"],"out":["ao"]}},"properties":{"Version":{"flags":"r","type":"s"},"SupportedFilesystems":{"flags":"r","type":"as"},"SupportedEncryptionTypes":{"flags":"r","type":"as"},"DefaultEncryptionType":{"flags":"r","type":"s"}},"signals":{}}}}
debug: dbus: {"notify":{"/org/freedesktop/UDisks2/Manager":{"org.freedesktop.UDisks2.Manager":{"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}}}}
debug: dbus: {"notify":{"/org/freedesktop/timedate1":{"org.freedesktop.timedate1":{"TimeUSec":1619514526301948}}}}
log: XXX udisks_manager.changed, reinit
debug: dbus: {"watch":{"path_namespace":"/org/freedesktop/UDisks2"},"id":"3"}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.MDRaid","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.VolumeGroup","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.LogicalVolume","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Drive","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Drive.Ata","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Block","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.PartitionTable","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Partition","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Block.LVM2","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.PhysicalVolume","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Filesystem","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Encrypted","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Swapspace","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.ISCSI.Session","path_namespace":"/org/freedesktop/UDisks2"}}
debug: dbus: {"add-match":{"interface":"org.freedesktop.UDisks2.Job","path_namespace":"/org/freedesktop/UDisks2"}}
log: XXX init_model begin, calling allSettled
> log: XXX init_model allSettled finished [{"status":"rejected","reason":[]},{"status":"fulfilled"},{"status":"fulfilled"},{"status":"fulfilled"},{"status":"fulfilled"},{"status":"fulfilled"},{"status":"fulfilled"},{"status":"fulfilled"}]
> warning: udisks manager proxy not ready yet
```

So again the proxy.wait() is broken here.

martinpitt:
I now have a reasonably useful debug log which illustrates what's happening:
```
> warning: XXX DBusProxy ctor org.freedesktop.UDisks2.Manager
> warning: XXX DBusProxy.update org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid  props·
> warning: XXX DBusProxy.update org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager stays invalid
> warning: XXX DBusClient.watch {"path":"/org/freedesktop/UDisks2/Manager","interface":"org.freedesktop.UDisks2.Manager"} created call id 2
> warning: XXX called DBusProxy.wait org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager func·
[...]
-> wait: ph_is_present("#drives")
> warning: XXX DBusProxy.update org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid  props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}
> warning: XXX DBusProxy.waited org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid true ex []
> warning: XXX DBusProxy.update org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid true props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}
> warning: Traceback (most recent call last):
[...]
-> wait: ph_is_present("#drives")
> warning: XXX DBusProxy.update org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid  props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}
> warning: XXX DBusProxy.waited org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid true ex []
> warning: XXX DBusProxy.update org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid true props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}

-> wait: ph_is_visible("#drives")
<- {'type': 'undefined'}
XXXXXXXXXX iter 0
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
<- {'type': 'undefined'}
XXXXXXXXXX iter 1
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
<- {'type': 'undefined'}
[...]
XXXXXXXXXX iter 21
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
> warning: XXX DBusProxy.waited org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid  ex []
<- {'type': 'object', 'value': []}
```

The most interesting bit is the *absence* of any DBusProxy/DBusClient calls during the iterations -- the caching in cockpit.dbus is pretty aggressive, so the iterations don't actually cause any new  actual D-Bus activity, until in iteration 21 when the proxie's  .wait() finally finishes.

The code makes sure that at no point in time the proxy's `.valid` moves from true to false.

There are some oddities which may point to the problem:
 * Why are there *three* calls (log lines) of DBusProxy.waited() ? This should all be cached, so I'd expect one, or 25 (if it's not cached).
 * Why do the three calls have different results? In particular, why is the third call `!valid` even though the first two are `valid` and nothing *sets* `valid` to false again?
 * The `[]` in the first two calls can only realistically come  from `on_message()`, which does `dfd.resolve(msg.reply[0] || [], options)`. Going to debug that next.

Code/test patch:
```diff
diff --git pkg/base1/cockpit.js pkg/base1/cockpit.js
index 22209c661..755442704 100644
--- pkg/base1/cockpit.js
+++ pkg/base1/cockpit.js
@@ -3004,6 +3004,9 @@ function factory() {
         var defined = false;
         var waits = cockpit.defer();
 
+        if (iface.endsWith("UDisks2.Manager") > 0)
+            console.warn("XXX DBusProxy ctor", iface);
+
         /* No enumeration on these properties */
         Object.defineProperties(self, {
             client: { value: client, enumerable: false, writable: false },
@@ -3013,6 +3016,8 @@ function factory() {
             wait: {
  enumerable: false, writable: false,
                 value: function(func) {
+                    if (iface.endsWith("UDisks2.Manager") > 0)
+                        console.warn("XXX called DBusProxy.wait", iface, path, "func", !!func);
                     if (func)
                         waits.promise.always(func);
                     return waits.promise;
@@ -3086,12 +3091,20 @@ function factory() {
         }
 
         function update(props) {
+            if (iface.endsWith("UDisks2.Manager") > 0)
+                console.warn("XXX DBusProxy.update", iface, path, "valid", valid, "props", JSON.stringify(props));
             if (props) {
                 extend(self.data, props);
                 if (!defined)
                     define();
                 valid = true;
             } else {
+                if (iface.endsWith("UDisks2.Manager") > 0) {
+                    if (valid)
+                        console.warn("XXXXXXXXXXXXXXXXXX DBusProxy.update", iface, path, "became invalid");
+                    else
+                        console.warn("XXX DBusProxy.update", iface, path, "stays invalid");
+                }
                 valid = false;
             }
             self.dispatchEvent("changed", props);
@@ -3112,6 +3125,8 @@ function factory() {
         client.subscribe({ path: path, interface: iface }, signal, options.subscribe !== false);
 
         function waited(ex) {
+            if (iface.endsWith("UDisks2.Manager") > 0)
+                console.warn("XXX DBusProxy.waited", iface, path, "valid", valid, "ex", JSON.stringify(ex));
             if (valid)
                 waits.resolve();
             else
@@ -3473,10 +3488,14 @@ function factory() {
             var dfd = cockpit.defer();
 
             var msg = JSON.stringify({ watch: match, id: id });
-            if (send(msg))
+            if (send(msg)) {
                 calls[id] = dfd;
-            else
+                if (path.interface && path.interface.endsWith("UDisks2.Manager"))
+                    console.warn("XXX DBusClient.watch", JSON.stringify(path), "created call id", id);
+            } else {
+                console.warn("XXXXXXXXXXXXXXXXXXXXXXX DBusClient.watch.send failed");
                 dfd.reject(new DBusError(closed));
+            }
 
             var ret = dfd.promise;
             ret.remove = function remove() {
@@ -3572,6 +3591,7 @@ function factory() {
         };
 
         self.proxy = function proxy(iface, path, options) {
+            // console.warn("XXX self.proxy", iface, path);
             if (!iface)
                 iface = name;
             iface = String(iface);
@@ -3580,6 +3600,8 @@ function factory() {
             var Constructor = self.constructors[iface];
             if (!Constructor)
                 Constructor = self.constructors["*"];
+            else
+                console.error("XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX used custom constructor for", iface);
             if (!options)
                 options = { };
             ensure_cache();
diff --git test/verify/check-storage-mounting test/verify/check-storage-mounting
index 8d6dd418a..8bf2dfbd8 100755
--- test/verify/check-storage-mounting
+++ test/verify/check-storage-mounting
@@ -119,13 +119,22 @@ class TestStorageMounting(StorageCase):
         m = self.machine
         b = self.browser
 
+        import subprocess
+        subprocess.check_call(["scp", "-r", "dist/base1/cockpit.js", "c:/usr/share/cockpit/base1/"])
+
         self.login_and_go("/storage")
+        b.wait_visible("#drives")
 
-        # Add a disk
-        m.add_disk("50M", serial="MYDISK")
-        b.wait_in_text("#drives", "MYDISK")
-        b.click('#drives .sidepanel-row:contains("MYDISK")')
-        b.wait_visible('#storage-detail')
+        # b.eval_js("window.debugging = 'dbus'")
+
+        for i in range(25):
+            print("XXXXXXXXXX iter", i)
+
+            # Add a disk
+            d = m.add_disk("50M", serial="MYDISK")
+            b.eval_js('cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()')
+            m.rem_disk(d)
+        return
 
         # Open format dialog and play with the checkboxes
```

martinpitt:
Finally, progress! I was already wondering about why I saw so few messages -- it seems chromium swallows/groups identical console.log messages. I now gave every `DBusProxy` instances an unique ID, and all of a sudden they appear:
```
XXXXXXXXXX iter 0
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
> warning: XXX DBusProxy id 30 ctor org.freedesktop.UDisks2.Manager
> warning: XXX DBusProxy.update 30 org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager stays invalid; props·
> warning: XXX called DBusProxy.wait id 30 org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager func·
> warning: XXX DBusProxy.update 30 valid org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager prev valid  props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}
> warning: XXX DBusProxy.waited 30 org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid true props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"} ex []
> warning: XXX DBusProxy.update 30 valid org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager prev valid true props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}
<- {'type': 'undefined'}

XXXXXXXXXX iter 12
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
> warning: XXX DBusProxy id 68 ctor org.freedesktop.UDisks2.Manager
> warning: XXX DBusProxy.update 68 org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager stays invalid; props·
> warning: XXX called DBusProxy.wait id 68 org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager func·
> warning: XXX DBusProxy.waited 68 org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager valid  props {} ex []
> warning: XXX DBusProxy.update 68 valid org.freedesktop.UDisks2.Manager /org/freedesktop/UDisks2/Manager prev valid  props {"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}
<- {'type': 'object', 'value': []}
```

This finally makes sense: 
 - In most cases, a proxy gets a property update (which turns it valid) before the client.watch() fires. But in some cases, the client.watch() fires first, so the proxy is not valid yet.
 - I think it makes sense to postpone the wait() result until the proxy becomes valid, as that's it's main raison d'être and documented that way.
 - The `[]` is simply the successful reply without arguments from the `add-match` call -- this has no return arguments. We should not treat it as an exception, but we can't tell this apart in waited() any more as this is called through `.always()`

```diff
diff --git pkg/base1/cockpit.js pkg/base1/cockpit.js
index 22209c661..b3afbbc7f 100644
--- pkg/base1/cockpit.js
+++ pkg/base1/cockpit.js
@@ -2996,14 +2996,20 @@ function factory() {
         };
     }
 
+    var proxy_id = 1;
+
     function DBusProxy(client, cache, iface, path, options) {
         var self = this;
+        self._id = proxy_id++;
         event_mixin(self, { });
 
         var valid = false;
         var defined = false;
         var waits = cockpit.defer();
 
+        if (iface.indexOf("UDisks2.Manager") > 0)
+            console.warn("XXX DBusProxy id", self._id, "ctor", iface);
+
         /* No enumeration on these properties */
         Object.defineProperties(self, {
             client: { value: client, enumerable: false, writable: false },
@@ -3013,6 +3019,8 @@ function factory() {
             wait: {
  enumerable: false, writable: false,
                 value: function(func) {
+                    if (iface.indexOf("UDisks2.Manager") > 0)
+                        console.warn("XXX called DBusProxy.wait id", self._id, iface, path, "func", !!func);
                     if (func)
                         waits.promise.always(func);
                     return waits.promise;
@@ -3087,11 +3095,19 @@ function factory() {
 
         function update(props) {
             if (props) {
+                if (iface.indexOf("UDisks2.Manager") > 0)
+                    console.warn("XXX DBusProxy.update", self._id, "valid", iface, path, "prev valid", valid, "props", JSON.stringify(props));
                 extend(self.data, props);
                 if (!defined)
                     define();
                 valid = true;
             } else {
+                if (iface.indexOf("UDisks2.Manager") > 0) {
+                    if (valid)
+                        console.warn("XXXXXXXXXXXXXXXXXX DBusProxy.update", self._id, iface, path, "became invalid; props", JSON.stringify(props));
+                    else
+                        console.warn("XXX DBusProxy.update", self._id, iface, path, "stays invalid; props", JSON.stringify(props));
+                }
                 valid = false;
             }
             self.dispatchEvent("changed", props);
@@ -3112,6 +3128,8 @@ function factory() {
         client.subscribe({ path: path, interface: iface }, signal, options.subscribe !== false);
 
         function waited(ex) {
+            if (iface.indexOf("UDisks2.Manager") > 0)
+                console.warn("XXX DBusProxy.waited", self._id, iface, path, "valid", valid, "props", JSON.stringify(self.data), "ex", JSON.stringify(ex));
             if (valid)
                 waits.resolve();
             else
@@ -3121,8 +3139,11 @@ function factory() {
         /* If watching then do a proper watch, otherwise object is done */
         if (options.watch !== false)
             client.watch({ path: path, interface: iface }).always(waited);
-        else
+        else {
+            if (iface.indexOf("UDisks2.Manager") > 0)
+                console.warn("XXXXXXXXXXXXXXX DBusProxy ctor", self._id, ": no watching, calling waited() immediately");
             waited();
+        }
     }
 
     function DBusProxies(client, cache, iface, path_namespace, options) {
@@ -3232,6 +3253,8 @@ function factory() {
         function send(payload) {
             if (channel && channel.valid) {
                 dbus_debug("dbus:", payload);
+                if (payload.indexOf('UDisks2.Manager"') > 0)
+                    console.warn("XXX DBusClient.send", payload);
                 channel.send(payload);
                 return true;
             }
@@ -3264,6 +3287,8 @@ function factory() {
                 channel.close({ problem: "protocol-error" });
                 return;
             }
+            if (msg.id == 2) // udisks2.Manager DBusClient.watch() call always seems to get 2
+                console.warn("XXX on_message for call id 2", payload);
             var dfd, options;
             if (msg.id !== undefined)
                 dfd = calls[msg.id];
@@ -3274,6 +3299,8 @@ function factory() {
                         options.type = msg.type;
                     if (msg.flags)
                         options.flags = msg.flags;
+                    if (!msg.reply[0])
+                        console.warn("XXX on_message with empty reply", payload);
                     dfd.resolve(msg.reply[0] || [], options);
                     delete calls[msg.id];
                 }
@@ -3412,9 +3439,9 @@ function factory() {
             });
 
             var msg = JSON.stringify(method_call);
-            if (send(msg))
+            if (send(msg)) {
                 calls[id] = dfd;
-            else
+            } else
                 dfd.reject(new DBusError(closed));
 
             return dfd.promise;
@@ -3473,10 +3500,14 @@ function factory() {
             var dfd = cockpit.defer();
 
             var msg = JSON.stringify({ watch: match, id: id });
-            if (send(msg))
+            if (send(msg)) {
                 calls[id] = dfd;
-            else
+                if (path.interface && path.interface.indexOf("UDisks2.Manager") > 0)
+                    console.warn("XXX DBusClient.watch", JSON.stringify(path), "created call id", id);
+            } else {
+                console.warn("XXXXXXXXXXXXXXXXXXXXXXX DBusClient.watch.send failed");
                 dfd.reject(new DBusError(closed));
+            }
 
             var ret = dfd.promise;
             ret.remove = function remove() {
@@ -3572,6 +3603,7 @@ function factory() {
         };
 
         self.proxy = function proxy(iface, path, options) {
+            // console.warn("XXX self.proxy", iface, path);
             if (!iface)
                 iface = name;
             iface = String(iface);
@@ -3580,6 +3612,8 @@ function factory() {
             var Constructor = self.constructors[iface];
             if (!Constructor)
                 Constructor = self.constructors["*"];
+            else
+                console.error("XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX used custom constructor for", iface);
             if (!options)
                 options = { };
             ensure_cache();
diff --git test/verify/check-storage-mounting test/verify/check-storage-mounting
index 8d6dd418a..8bf2dfbd8 100755
--- test/verify/check-storage-mounting
+++ test/verify/check-storage-mounting
@@ -119,13 +119,22 @@ class TestStorageMounting(StorageCase):
         m = self.machine
         b = self.browser
 
+        import subprocess
+        subprocess.check_call(["scp", "-r", "dist/base1/cockpit.js", "c:/usr/share/cockpit/base1/"])
+
         self.login_and_go("/storage")
+        b.wait_visible("#drives")
 
-        # Add a disk
-        m.add_disk("50M", serial="MYDISK")
-        b.wait_in_text("#drives", "MYDISK")
-        b.click('#drives .sidepanel-row:contains("MYDISK")')
-        b.wait_visible('#storage-detail')
+        # b.eval_js("window.debugging = 'dbus'")
+
+        for i in range(25):
+            print("XXXXXXXXXX iter", i)
+
+            # Add a disk
+            d = m.add_disk("50M", serial="MYDISK")
+            b.eval_js('cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try" }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()')
+            m.rem_disk(d)
+        return
 
         # Open format dialog and play with the checkboxes
```

martinpitt:
> I think it makes sense to postpone the wait() result until the proxy becomes valid, as that's it's main raison d'être and documented that way.

I did try that, but it turns out that we don't actually use it that way in the code. For example, the storage page does

                Promise.allSettled([client.manager_lvm2.wait(), client.manager_iscsi.wait()]).then(() => {

or
```js
    Promise.allSettled([client.manager.wait(),
        client.mdraids.wait(), client.vgroups.wait(), client.drives.wait(),
        client.blocks.wait(), client.blocks_ptable.wait(), client.blocks_lvm2.wait(), client.blocks_fsys.wait()
```
where e.g. lvm2 is *not* guaranteed to ever become "valid". As the current `.wait()` behaves "wrongly", I'll keep that behaviour for now and just fix the misinterpretation of the `[]` as error.

martinpitt:
Finally! I think this is good enough in principle for now, and already a big improvement. Fortunately the actual patch is quite small, it was just a ridiculously hard journey for me to get there..

martinpitt:
Tests look reasonably good. I realized that we now don't need the storaged commit any more, as the proxy wait should now be reliable.

martinpitt:
Changed my mind again -- I keep the storaged robustification for the next time when some D-Bus error happens. Getting to the root cause of that bug was too hard to do it all over again in the future.

martinpitt:
[testResizeLuks](https://logs.cockpit-project.org/logs/pull-15752-20210428-052504-63de6efc-ubuntu-stable-bots-1952/log.html#196-2) is still broken on the refreshed ubuntu-stable, but I'll handle that in the image refresh or in a separate PR. This is an entirely different problem, and one which we had on other OSes for a while, I think it's https://github.com/cockpit-project/bots/issues/1739 and thus needs the naughty copied.

marusak (approved):
thank you for looking into this! This `wait()` behavior is really confusing.

martinpitt:
After a quick discussion with @mvollmer we both agree that what we would *like* is that after calling proxy.wait(), proxy.valid is definitively true or false. My attempt at this is below -- if `client.wait()` can resolve the `waits` promise already (because the proxy sent an update before it finished), then it's all good, and basically how the code behaves today. But if the update() happens later, then resolving or rejecting `waits` is postponed to update():

```diff
--- pkg/base1/cockpit.js
+++ pkg/base1/cockpit.js
@@ -3093,36 +3093,38 @@ function factory() {
             });
         }
 
-        function update(props) {
+        function update(props, _nofail=false) {
             if (props) {
                 if (iface.indexOf("UDisks2.Manager") > 0)
-                    console.warn("XXX DBusProxy.update", self._id, "valid", iface, path, "prev valid", valid, "waits", waits.promise.state, "props", JSON.stringify(props));
+                    console.warn("XXX DBusProxy.update", self._id, "valid", "_nofail", _nofail, iface, path, "prev valid", valid, "waits", waits.promise.state, "props", JSON.stringify(props));
                 extend(self.data, props);
                 if (!defined)
                     define();
                 if (!valid) {
                     if (iface.indexOf("UDisks2.Manager") > 0)
                         console.warn("XXX DBusProxy.update", self._id, "turned valid, resolving");
-                    // waits.resolve();
+                    waits.resolve();
+                    valid = true;
                 } else {
                     if (iface.indexOf("UDisks2.Manager") > 0)
                         console.warn("XXX DBusProxy.update", self._id, "was already valid");
                 }
-                valid = true;
             } else {
                 if (iface.indexOf("UDisks2.Manager") > 0) {
                     if (valid)
                         console.warn("XXXXXXXXXXXXXXXXXX DBusProxy.update", self._id, iface, path, "became invalid; props", JSON.stringify(props));
                     else
-                        console.warn("XXX DBusProxy.update", self._id, iface, path, "stays invalid; props", JSON.stringify(props));
+                        console.warn("XXX DBusProxy.update", self._id, "_nofail", _nofail, iface, path, "stays invalid; props", JSON.stringify(props));
                 }
+                if (!_nofail)
+                    waits.reject(new DBusError("interface has no properties"));
                 valid = false;
             }
             self.dispatchEvent("changed", props);
         }
 
         cache.connect(path, iface, update, true);
-        update(cache.lookup(path, iface));
+        update(cache.lookup(path, iface), true);
 
         function signal(path, iface, name, args) {
             self.dispatchEvent("signal", name, args);
@@ -3138,7 +3140,14 @@ function factory() {
         // If watching then do a proper watch, otherwise the object is done
         if (options.watch !== false) {
             client.watch({ path: path, interface: iface })
-                .then(() => waits.resolve())
+                .then(() => {
+                    if (valid) {
+                        console.warn("XXX DBusProxy ctor", self._id, "already valid, resolving waits");
+                        waits.resolve();
+                    } else {
+                        console.warn("XXX DBusProxy ctor", self._id, "not valid yet, deferring to update()");
+                    }
+                })
                 .catch(ex => waits.reject(ex));
         } else {
             waits.resolve();
diff --git test/common/cdp.py test/common/cdp.py
index 9244546f7..2c992f95f 100644
--- test/common/cdp.py
+++ test/common/cdp.py
@@ -68,7 +68,7 @@ def jsquote(str):
 class CDP:
     def __init__(self, lang=None, verbose=False, trace=False, inject_helpers=[]):
         self.lang = lang
-        self.timeout = 60
+        self.timeout = 5
         self.valid = False
         self.verbose = verbose
         self.trace = trace
diff --git test/verify/check-storage-mounting test/verify/check-storage-mounting
index 8bf2dfbd8..ddcc9dc7f 100755
--- test/verify/check-storage-mounting
+++ test/verify/check-storage-mounting
@@ -121,6 +121,7 @@ class TestStorageMounting(StorageCase):
 
         import subprocess
         subprocess.check_call(["scp", "-r", "dist/base1/cockpit.js", "c:/usr/share/cockpit/base1/"])
+        subprocess.check_call(["scp", "-r", "dist/storaged", "c:/usr/share/cockpit/"])
 
         self.login_and_go("/storage")
         b.wait_visible("#drives")
```

But that causes a hang with the [iscsi proxy](https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/client.js#L385): That simply does not exist in Debian/Ubuntu, so:

   * There only ever is one `update(undefined)` call from the attempt to fish out an existing proxy from the cache here:

            update(cache.lookup(path, iface), true);

     We don't want that to reject the promise already, as it's perfectly valid for the first time a proxy is instantiated to not be in the cache..

 * When client.watch() resolves, the proxy is not valid yet. But as per the desire above, we don't want to reject the waits promise yet, as the next update() could make it valid, and thus it gets defered.

 * update() will *never* be called, as that is not a valid proxy. As a result, the proxy(ISCSI).wait() promise will hang forever, and with it the Storage page.

So AFAICS, the only way out of that would be to make DBusClient.wait() stronger -- it needs to resolve only when there is a definitive answer. I don't see how to fix this in DBusProxy in a robust way.

mvollmer:
I have reproduced the bug with @martinpitt's test case, and added my own log messages.

When successful:
 
```
XXXXXXXXXX iter 0
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try", debug: true }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
> log: SEND {"add-match":{"path":"/org/freedesktop/UDisks2/Manager","interface":"org.freedesktop.UDisks2.Manager"}}
> log: SEND {"watch":{"path":"/org/freedesktop/UDisks2/Manager","interface":"org.freedesktop.UDisks2.Manager"},"id":"2"}
> log: MSG {"owner":":1.5"}
> log: MSG {"meta":{"org.freedesktop.UDisks2.Manager":{"methods":{"CanFormat":{"in":["s"],"out":["(bs)"]},"CanResize":{"in":["s"],"out":["(bts)"]},"CanCheck":{"in":["s"],"out":["(bs)"]},"CanRepair":{"in":["s"],"out":["(bs)"]},"LoopSetup":{"in":["h","a{sv}"],"out":["o"]},"MDRaidCreate":{"in":["ao","s","s","t","a{sv}"],"out":["o"]},"EnableModules":{"in":["b"],"out":[]},"EnableModule":{"in":["s","b"],"out":[]},"GetBlockDevices":{"in":["a{sv}"],"out":["ao"]},"ResolveDevice":{"in":["a{sv}","a{sv}"],"out":["ao"]}},"properties":{"Version":{"flags":"r","type":"s"},"SupportedFilesystems":{"flags":"r","type":"as"},"SupportedEncryptionTypes":{"flags":"r","type":"as"},"DefaultEncryptionType":{"flags":"r","type":"s"}},"signals":{}}}}
> log: MSG {"notify":{"/org/freedesktop/UDisks2/Manager":{"org.freedesktop.UDisks2.Manager":{"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}}}}
> log: MSG {"reply":[],"id":"2"}
> log: MSG {"notify":{"/org/freedesktop/UDisks2/Manager":{"org.freedesktop.UDisks2.Manager":{}}}}
<- {'type': 'undefined'}
```

There is a `meta` and `notify` message before the `reply` for the watch call, as expected.  There is also a unexpected empty `notify` message afterwards. No idea what that is about.

When it fails:
```
XXXXXXXXXX iter 1
-> cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try", debug: true }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()
> log: SEND {"add-match":{"path":"/org/freedesktop/UDisks2/Manager","interface":"org.freedesktop.UDisks2.Manager"}}
> log: SEND {"watch":{"path":"/org/freedesktop/UDisks2/Manager","interface":"org.freedesktop.UDisks2.Manager"},"id":"2"}
> log: MSG {"owner":":1.5"}
> log: MSG {"reply":[],"id":"2"}
> log: MSG {"meta":{"org.freedesktop.UDisks2.Manager":{"methods":{"CanFormat":{"in":["s"],"out":["(bs)"]},"CanResize":{"in":["s"],"out":["(bts)"]},"CanCheck":{"in":["s"],"out":["(bs)"]},"CanRepair":{"in":["s"],"out":["(bs)"]},"LoopSetup":{"in":["h","a{sv}"],"out":["o"]},"MDRaidCreate":{"in":["ao","s","s","t","a{sv}"],"out":["o"]},"EnableModules":{"in":["b"],"out":[]},"EnableModule":{"in":["s","b"],"out":[]},"GetBlockDevices":{"in":["a{sv}"],"out":["ao"]},"ResolveDevice":{"in":["a{sv}","a{sv}"],"out":["ao"]}},"properties":{"Version":{"flags":"r","type":"s"},"SupportedFilesystems":{"flags":"r","type":"as"},"SupportedEncryptionTypes":{"flags":"r","type":"as"},"DefaultEncryptionType":{"flags":"r","type":"s"}},"signals":{}}}}
> log: MSG {"notify":{"/org/freedesktop/UDisks2/Manager":{"org.freedesktop.UDisks2.Manager":{"SupportedFilesystems":["ext2","ext3","ext4","vfat","ntfs","exfat","xfs","reiserfs","nilfs2","btrfs","minix","udf","f2fs","swap"],"DefaultEncryptionType":"luks1","SupportedEncryptionTypes":["luks1","luks2"],"Version":"2.9.2"}}}}
<- {'type': 'object', 'value': []}
```

The `meta` and `notify` messages come after the reply, violating the ordering guarantees of the protocol.  The weird empty `notify` message is absent here.

So, this looks like a bug in cockpit-bridge, and the next step would be to dig into that. I don't think we should try to change cockpit.js to cope with a bug in cockpit-bridge.

mvollmer:
My logging changes, for reference:

```diff
diff --git a/pkg/base1/cockpit.js b/pkg/base1/cockpit.js
index 22209c661..c345b888c 100644
--- a/pkg/base1/cockpit.js
+++ b/pkg/base1/cockpit.js
@@ -3231,6 +3231,8 @@ function factory() {
 
         function send(payload) {
             if (channel && channel.valid) {
+                if (args.debug)
+                    console.log("SEND", payload);
                 dbus_debug("dbus:", payload);
                 channel.send(payload);
                 return true;
@@ -3253,6 +3255,8 @@ function factory() {
         }
 
         function on_message(event, payload) {
+            if (args.debug)
+                console.log("MSG", payload);
             dbus_debug("dbus:", payload);
             var msg;
             try {
diff --git a/test/common/chromium-cdp-driver.js b/test/common/chromium-cdp-driver.js
index e2507f9fb..08f8fde93 100755
--- a/test/common/chromium-cdp-driver.js
+++ b/test/common/chromium-cdp-driver.js
@@ -66,7 +66,7 @@ var messages = [];
 var logPromiseResolver;
 var nReportedLogMessages = 0;
 var unhandledExceptions = [];
-var shownMessages = []; // Show every message just once, keep here seen messages
+var shownMessages = [ "One of aria-label or aria-labelledby properties should be passed when using the progress component without a title." ]; // Show every message just once, keep here seen messages
 
 function clearExceptions() {
     unhandledExceptions.length = 0;
@@ -81,7 +81,7 @@ function setupLogging(client) {
 
         messages.push([ info.type, msg ]);
         if (shownMessages.indexOf(msg) == -1) {
-            shownMessages.push(msg);
+            // shownMessages.push(msg);
             process.stderr.write("> " + info.type + ": " + msg + "\n")
         }
 
@@ -128,7 +128,7 @@ function setupLogging(client) {
             delete msg.args;
             const msgstr = JSON.stringify(msg);
             if (shownMessages.indexOf(msgstr) == -1) {
-                shownMessages.push(msgstr);
+                // shownMessages.push(msgstr);
                 process.stderr.write("CDP: " + JSON.stringify(orig) + "\n");
             }
         }
diff --git a/test/verify/check-storage-mounting b/test/verify/check-storage-mounting
index 8d6dd418a..8215f2909 100755
--- a/test/verify/check-storage-mounting
+++ b/test/verify/check-storage-mounting
@@ -119,13 +119,22 @@ class TestStorageMounting(StorageCase):
         m = self.machine
         b = self.browser
 
+        m.upload([ "/home/mvo/work/cockpit/dist/base1/cockpit.js" ], "/usr/share/cockpit/base1")
+
         self.login_and_go("/storage")
+        b.wait_visible('#drives')
 
-        # Add a disk
-        m.add_disk("50M", serial="MYDISK")
-        b.wait_in_text("#drives", "MYDISK")
-        b.click('#drives .sidepanel-row:contains("MYDISK")')
-        b.wait_visible('#storage-detail')
+        # b.eval_js("window.debugging = 'dbus'")
+
+        for i in range(25):
+            print("XXXXXXXXXX iter", i)
+
+            # Add a disk
+            d = m.add_disk("50M", serial="MYDISK")
+            b.eval_js('cockpit.dbus("org.freedesktop.UDisks2", { superuser: "try", debug: true }).proxy("org.freedesktop.UDisks2.Manager", "/org/freedesktop/UDisks2/Manager", { watch: true }).wait()')
+            m.rem_disk(d)
+
+        return
 
         # Open format dialog and play with the checkboxes
 
```

martinpitt:
The root cause of this was fixed in PR #15761 , but the storage page change is still a nice robustification IMHO. So I cleaned this up.

marusak (approved):

martinpitt:
Hmm, not sure why the [test-dbus-framed](https://github.com/cockpit-project/cockpit/pull/15752/checks?check_run_id=2473320067) unit test on i386 hates me so much in this PR -- this is obviously unrelated, but happened three times in a row now.

pkg/storaged/client.js
```diff
@@ -505,7 +505,15 @@ function init_model(callback) {
     Promise.allSettled([client.manager.wait(),
         client.mdraids.wait(), client.vgroups.wait(), client.drives.wait(),
         client.blocks.wait(), client.blocks_ptable.wait(), client.blocks_lvm2.wait(), client.blocks_fsys.wait()
-    ]).then(() => {
+    ]).then(results => {
+        // we at least need the manager object; if it doesn't exist, wait for the next proxy onchanged
+        if (results[0].status !== 'fulfilled') {
+            console.warn("init_model(): udisks manager proxy failed:", JSON.stringify(results[0].reason));
+            client.features = false;
```
mvollmer, line 512: This could just check `client.manager.valid`, which actually happens in enable_udisks_features.

pkg/storaged/client.js
```diff
@@ -505,7 +505,15 @@ function init_model(callback) {
     Promise.allSettled([client.manager.wait(),
         client.mdraids.wait(), client.vgroups.wait(), client.drives.wait(),
         client.blocks.wait(), client.blocks_ptable.wait(), client.blocks_lvm2.wait(), client.blocks_fsys.wait()
-    ]).then(() => {
+    ]).then(results => {
+        // we at least need the manager object; if it doesn't exist, wait for the next proxy onchanged
```
mvollmer, line 509: A "onchanged" event doesn't rerun feature detection, no?  The only way to get out of "features = false" is to reload the page.

mvollmer:
Hmm, I am inclined to vote against this... The code in question shouldn't ever run with `client.manager.valid == false` because this is already caught in `init_storaged`.

The whole way how the storage page is initialized is bonkers and doesn't deal at all with UDisk2 coming and going.  For example, we should be re-running EnableModules when udisksd is restarted.

So, I'd say this change makes the code even more complex, without a definite gain. Is that fair?

pkg/storaged/client.js
marusak: It does. It calls `init_client()`.

pkg/storaged/client.js
martinpitt: Right, it could. Not sure if it's more robust to rely on that or on the .wait() promise. I'm happy to change this.

pkg/storaged/client.js
martinpitt: Right. That's in fact what has hidden the B-Bus race in many cases.

martinpitt:
@mvollmer: I don't think it adds complexity, it just adds more error checking. Given how utterly hard it was for me (4.5 hours) to trace the reason for the occasional "this dialog has broken value" to "the Manager proxy has failed", and that `Promise.allSettled()` has no built-in error-checking *at all* (that's what it is for, you are supposed to check errors yourself), my gut feeling is that it's still worth it. But I don't have a strong opinion about it, and that dbus unit test hates me anyway -- so there's no urgency about this at all.

pkg/storaged/client.js
mvollmer: > It does. It calls `init_client()`.

Oh, your are right. That changes everything, sorry for missing it.

mvollmer:
> The whole way how the storage page is initialized is bonkers

It's less bonkers than I thought, and my arguments are incorrect: init_client does run during the lifetime of the page.

mvollmer (approved):

PR #15761, opened 2021-04-29
bridge: Don't stop introspecting immediately when a ObjectManager is detected

mvollmer:


martinpitt (approved):
Thanks so much for figuring this out! :100: 

The rhel-8-5 test was stuck, the sink crashed with a "broken pipe" error. I restarted it.

PR #15762, opened 2021-04-29
lib: Drop redundant error message in alert modals

garrett:
Alerts are already visibly errors and provide ARIA to screen readers.

Additionally, `<kbd>` is not passed to screen readers in any special way.
By default, `<kbd>` just unnessarily changes the font to mono.

### DIFF
--- a/pkg/lib/cockpit-components-inline-notification.jsx
+++ b/pkg/lib/cockpit-components-inline-notification.jsx
@@ -89,7 +89,7 @@ InlineNotification.propTypes = {
 export const ModalError = ({ dialogError, dialogErrorDetail }) => {
     return (
         <Alert variant='danger' isInline title={dialogError}>
-            { dialogErrorDetail && <p> {_("Error message")}: <samp>{dialogErrorDetail}</samp> </p> }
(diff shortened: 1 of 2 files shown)


KKoukiou (dismissed):
If tests agree :)

garrett:
> If tests agree :)

"works on my machine!"

martinpitt:
Unfortunately the tests do [need to be adjusted](https://logs.cockpit-project.org/logs/pull-15762-20210429-150401-dc4a4498-debian-stable/log.html#71-2). Indeed it actually [looks better now](https://logs.cockpit-project.org/logs/pull-15762-20210429-150401-dc4a4498-debian-stable/TestFirewall-testMultipleZones-debian-stable-127.0.0.2-2301-FAIL.png).

According to `git grep 'Error message' test/verify` this was the only place which made that assumption. @garrett , want to update that yourself, or want me to FP over your branch?

martinpitt:
As @garrett is out for the weekend now, I force-pushed that to fix the test, and the commit message; it was `<samp>`, not `<kbd>`.

KKoukiou (approved):

garrett:
Thanks for seeing this through! :+1:

PR #15763, opened 2021-04-29
test: Fix initial expected memory in TestServices.testBasic

martinpitt:
This often starts out with being exactly "1.0 MB", which fails the
previous condition. There is no particular expectation how much memory
this initially uses, so just reduce it.

----

I've seen this at least twice today, most recently [here](https://logs.cockpit-project.org/logs/pull-15761-20210429-145258-8dad8c4e-rhel-8-5/log.html#252)


marusak (approved):
Thanks :)
But since we have no idea what should be there maybe just check for it existing would be enough. Or maybe `> 0` as is reasonable to think the value should be non-zero.
But this works for me as well.

PR #15765, opened 2021-04-30
Drop cockpit.all()

martinpitt:
cockpit.all() was deprecated over two years ago in commit f83c353d8697c, and was never documented API.

Port the last two remaining users to Promise.all(), and clean it up.


martinpitt:
Bah, eslint, always complaining :grinning:

marusak (approved):
whooho \o/

PR #15766, opened 2021-04-30
lib: Remove duplicate css rule

marusak:
This exact same rule is defined a few lines below once again.


marusak:
Actually, do we even use `.ct-card` anywhere?

KKoukiou (approved):

PR #15770, opened 2021-04-30
base1: Fix D-Bus watch tests

martinpitt:
Similarly to commit dc2aaff1ab82c26898, the other two D-Bus watch tests
need a deep update of the cache as well. After commit 4e5a309f0af the
order of the messages changed (often), and we could get the "empty"
state updates after the "real" one.


martinpitt:
They started to fail the unit tests of pretty much every PR now, e.g. https://github.com/cockpit-project/cockpit/actions/runs/799015719 So this is rather urgent.

mvollmer (approved):

PR #15771, opened 2021-04-30
.github: add workflow to spot .github/ changes

allisonkarlitskaya:
Ensure that we explicitly ack changes to .github/ by way of setting a
label.

### DIFF
--- a/.github/workflows/reposchutz.yml
+++ b/.github/workflows/reposchutz.yml
@@ -1,0 +1,41 @@
+name: repository
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+
+jobs:
+  check:
+    name: Protection checks
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    env:
+      GIT_DIR: git-dir.git
+    steps:
+      - name: Clear .github-changes label
+        if: ${{ !endsWith(github.event.action, 'labeled') }}
+        uses: actions/github-script@v3
+        with:
+          script: |
+            try {
+              const { repo: { owner, repo }, number: issue_number } = context;
+              await github.issues.removeLabel({ owner, repo, issue_number, name: '.github-changes' });
+            } catch (e) {
+              if (e.name == 'HttpError' && e.status == 404) {
+                /* expected: 404 if label isn't set */
+              } else {
+                throw e;
+              }
+            }
+
+      - name: Check for .github changes
+        # We want to run this check any time the .github-changes label is not
+        # set, which needs to include the case where we just unset it above.
+        if: ${{ !endsWith(github.event.action, 'labeled') ||
+                !contains(github.event.pull_request.labels.*.name, '.github-changes') }}
+        run: |
+          set -x
+          git init -b main
+          git remote add origin "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}"
+          git fetch --depth=1 origin '${{ github.sha }}'
+          git fetch --depth=1 origin '${{ github.event.pull_request.head.sha }}'
+          git diff --exit-code '${{ github.sha }}' '${{ github.event.pull_request.head.sha }}' -- .github


marusak:
OOI what is point of this?
And do I read it correctly that some workflow just fails when you touch .github? (and you cannot make it green)

.github/workflows/github-metadata.yml
```diff
@@ -0,0 +1,23 @@
+name: github_metadata
+on:
+  pull_request_target:
+    types: [opened, reopened, labeled, unlabeled, synchronize]
+
+jobs:
+  check:
+    name: protect github metadata
```
martinpitt, line 8: Bikeshedding: "Protect GitHub workflow changes"

.github/workflows/github-metadata.yml
```diff
@@ -0,0 +1,23 @@
...
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'github-metadata-change') }}
```
martinpitt, line 9: Maybe "approved-workflow-change"?

.github/workflows/github-metadata.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    env:
+      GIT_DIR: git-dir.git
```
martinpitt, line 13: What does this do?

.github/workflows/github-metadata.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Fetch PR
+        run: |
+          set -x
+          git init -b main
+          git remote add origin "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}"
+          git fetch origin '${{ github.event.pull_request.head.sha }}'
```
martinpitt, line 20: Hmmm.. You have a really strict timeout. Could this perhaps just do something like

    curl --silent --show-error -L https://github.com/cockpit-project/cockpit/pull/15771.diff | grep '^+++ .*/\.github/'

(like for [this PR](https://github.com/cockpit-project/cockpit/pull/15771.diff)) instead of cloning the whole repo just for this?

martinpitt (changes requested):
Thanks! Some bikeshedding. I'm not 100% sure yet whether we  want this or this is just additional bureaucracy -- we review PRs before landing anyway, and PR workflows don't even run for first-time contributors automatically (there is an explicit approval button now). But it's an interesting idea.

martinpitt:
If we want to put any kind of reliance on this, there is an important part missing, though: Every PR update must automatically remove the label again. Otherwise you could send a PR, get it reviewed, we set the label, then you force-push and add something malicious. This could even trick us into *not* carefully reviewing the force-push again.

.github/workflows/github-metadata.yml
allisonkarlitskaya: There's other things (like codeowners and so) that can live in `.github/`.  It's not just workflows.

.github/workflows/github-metadata.yml
allisonkarlitskaya: ditto on workflow vs metadata bikeshed.

.github/workflows/github-metadata.yml
allisonkarlitskaya: This makes it so the `git init` below knows where to write to.  It's basically an alternative to separately passing `--git-dir` to every single git command.

.github/workflows/github-metadata.yml
allisonkarlitskaya: That seems a bit racy....  I'd prefer to run on the object we've been handed to run on....

allisonkarlitskaya:
> If we want to put any kind of reliance on this, there is an important part missing, though: Every PR update must automatically remove the label again. Otherwise you could send a PR, get it reviewed, we set the label, then you force-push and add something malicious. This could even trick us into _not_ carefully reviewing the force-push again.

This part I absolutely agree with.  I'll add that.

.github/workflows/github-metadata.yml
martinpitt: OK, fair enough. Calling it "meta"data is still a bit weird though. And our other "name:"s are capitalized.

.github/workflows/github-metadata.yml
martinpitt: Can you then please at least do shallow fetches of the two things you want to compare? And bump the timeout a bit?

martinpitt:
Let's discuss that in tomorrow's team meeting, I put that on the agenda.

allisonkarlitskaya:
We discussed this in the meeting, and although some people had some reservations about if we *really* need this (since we should catch these kinds of changes in reviews), nobody thought that it would really hurt anything to have it.

Additionally, there is at least one scenario where this is helpful: when you review a large patch in-depth the first time, and then request changes, it's not uncommon to just go to the changed section to make sure that you got what you asked for.  That's maybe not the best practice for dealing with external contributions, but we're all probably guilty of it from time to time.  Additionally, if the submitter rebased their patch in the meanwhile, the github-generated "forcepushed" diff is basically unreadable.  That's a window where someone could sneak in nasty changes, and this workflow would catch that.

Also: the potential cost of someone landing malicious changes to our `.github` directory is incredibly high.

allisonkarlitskaya:
Changed: the label is called `.github-changes` now, which is pretty neutral in terms of terminology squabbling, and the label now gets unset on any code changes.

I've also generalised the name of the workflow so that we could add additional "safety checks" here if we wanted to.

.github/workflows/github-metadata.yml
martinpitt: Repeating in new review, as this is labeled as outdated.

.github/workflows/reposchutz.yml
```diff
@@ -0,0 +1,40 @@
...
+        # We want to run this check any time the .github-changes label is not
+        # set, which needs to include the case where we just unset it above.
+        if: ${{ !endsWith(github.event.action, 'labeled') ||
+                !contains(github.event.pull_request.labels.*.name, '.github-changes') }}
+        run: |
+          set -x
+          git init -b main
+          git remote add origin "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}"
+          git fetch origin '${{ github.event.pull_request.head.sha }}'
+          git diff --exit-code '${{ github.sha }}' '${{ github.event.pull_request.head.sha }}' -- .github
```
martinpitt, line 40: So the intention here is to not fetch the whole repo (as a naïve `git clone` would do), but only the bits we need. I wonder how this can work, though, as it only fetches one half, and not `github.sha`. Either this only works if the PR is fully rebased on current master (and thus `github.sha` is part of the history of `github.event.pull_request.head.sha`), or somehow the fetch pulls all the heads/branches anyway (in which case there's nothing we can do).

Did you try this with an "outdated" branch?

.github/workflows/reposchutz.yml
```diff
@@ -0,0 +1,40 @@
+name: repository
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+
+jobs:
+  check:
+    name: Protection checks
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
```
martinpitt, line 10: There is a little too much going on here for this timeout -- let's make it 5 mins?

.github/workflows/reposchutz.yml
```diff
@@ -0,0 +1,40 @@
...
+              } else {
+                throw e;
+              }
+            }
+
+      - name: Check for .github changes
+        # We want to run this check any time the .github-changes label is not
+        # set, which needs to include the case where we just unset it above.
+        if: ${{ !endsWith(github.event.action, 'labeled') ||
+                !contains(github.event.pull_request.labels.*.name, '.github-changes') }}
```
martinpitt, line 34: I find a label with a dot a bit weird -- can we drop that?

martinpitt (changes requested):
Thanks!

.github/workflows/reposchutz.yml
allisonkarlitskaya: The `github.sha` is according to the docs, the base commit of the PR, which is also one of the parent commits of the merge commit that github makes for the PR (as I understand it).

But it also means I fetch the full history, which is not necessary.  Probably it makes sense to replace this with two separate fetches, of the SHAs that we are interested in, and `--depth=1`.

.github/workflows/reposchutz.yml
allisonkarlitskaya: Sure

.github/workflows/reposchutz.yml
allisonkarlitskaya: Why?  The github docs even mention that label names could happily contain emoji.  Do you have a suggestion for an alternative?

martinpitt: It's not related to what the API allows, this was a purely aesthetical remark. Just "github-changes"? But this is of course not that important, not a strong opinion.

.github/workflows/reposchutz.yml
martinpitt: Sounds good!

.github/workflows/reposchutz.yml
allisonkarlitskaya: I'd prefer to leave it as is, in that case.  ".github" is the thing that's changing, so it makes sense to me to name it as such :)

martinpitt (approved):
reposchutz! @major would be so proud :grin:

major:
@martinpitt

PR #15778, opened 2021-05-03
make dist fixes, allow building in a minimal environment, re-enable packit

martinpitt:
This re-enables packit tests after commit c4908fe050dda.

This is a hack, but at least we have all our tests to tell us when we break this again.

- [ ] ~add some missing build dependencies to sandcastle container: https://github.com/packit/sandcastle/pull/122~

### DIFF
--- a/tools/mock-build-env/README.md
+++ b/tools/mock-build-env/README.md
@@ -1,0 +1,5 @@
+These are some stub pkg-config and include files that allow us to run
+`make dist` in a restricted build environment without the regular build
+dependencies, in particular [Packit](https://packit.dev/)'s [sandcastle container](https://github.com/packit/sandcastle).
+
+These are used in [packit.yaml](./packit.yaml).
--- a/Makefile.am
+++ b/Makefile.am
@@ -290,10 +290,6 @@ recheck-memory: valgrind-suppressions
 
 if ENABLE_DOC
 
-include doc/Makefile-doc.am
-include doc/guide/Makefile-guide.am
-include doc/man/Makefile-man.am
-
 dist-doc-hook:
 	@true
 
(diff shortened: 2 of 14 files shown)


allisonkarlitskaya (commented):
As we discussed on IRC: I'd prefer to try to construct an independent mock environment that makes `./configure` happy without making changes to our build system.  I'm confident that this is possible.

martinpitt:
I moved that over to a bunch of mock .pc files and a single empty include. This isn't too bad IMHO, and at least nicely separated from our actual build system. The other commits are valid fixes IMHO.

tools/mock-build-env/README.md
```diff
@@ -0,0 +1 @@
+These files are ...
```
martinpitt, line 1: ... still to be written, d'oh!

martinpitt: Fixed.

allisonkarlitskaya (approved):
I'm of two minds about your approach here:
 - on one hand, I feel like it would be better to mock up ssh/pcp/etc stuff and remove all the `--disable` in order to avoid running into a situation where we're testing a non-canonical build (since you noticed that we produced different results depending on our configure flags and such differences might slip in again in the future)
 - on the other hand, testing different scenarios is always good, and indeed, led you to find new bugs here.

So maybe a TODO item instead: you can merge this as is (or take my suggestion to add more "mocking" and less "disabling" and we do another round of review) and then later on we can add an explicit test scenario that tries to build cockpit with as much `--disable` as we can reasonably throw at it and checks to make sure we still get a viable result  which properly distchecks and so on...

martinpitt:
For posterity, we discussed that on IRC:

    lis | but it's actually a good point that packit is basically doing exactly the testing i proposed

PR #15779, opened 2021-05-03
test: Quick-fix dist tarball unpacking

martinpitt:
The tar timestamp workaround from commit cf04588de554 with only touching
webpacks, but not node_modules/ and package-lock.json was completely
flawed: This would only work on an already existing cockpit.git
checkout, but not on any clone that happened after building the dist
cache.

So we have to touch all the files, which also simplifies the code quite
a bit.

To work around the `tar --touch` bug, add a sleep to ensure that the
unpacked files are at least one second apart. This is a giant hack of
course, but let's unbreak our CI and investigate this in an orderly
fashion.

-----

Fixes failures like [this](https://logs.cockpit-project.org/logs/pull-1980-20210503-111056-aa04ab54-ubuntu-stable-cockpit-project-cockpit/log.html).

I'm terribly sorry for the thinko!

I deliberately make this no-test, so that I can trigger some integration tests *after* the dist cache is built. Starting them right away would not use any caching, and thus not exercise this code path. I only start fedora-33 right away (as that doesn't use caching anyway).

### DIFF
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -95,20 +95,17 @@ def download_cache(wait=False):
             message(f"make_dist: pre-built dist for {sha} not yet available, waiting...")
             time.sleep(30)
 
-    # Unfortunately our dist tarball generation does not deal with sub-second file mtime differences,
-    # which can invert the relative mtime of package-lock.json vs.  dist/*/manifest.json
-    # to avoid this, only touch the webpack-built files (to satisfy make dependencies),
-    # but leave the `npm install` ones alone.
-    for (unpack_path, touch) in [("node_modules", False),
-                                 ("package-lock.json", False),
-                                 ("dist", True),
-                                 ("tools/debian/copyright", True)]:
+    for unpack_path in ["node_modules", "package-lock.json", "dist", "tools/debian/copyright"]:
         if os.path.exists(unpack_path):
             continue
         message(f"make_dist: Extracting cached {unpack_path}...")
         p_git = subprocess.Popen(["git", "--git-dir", dist_git_checkout, "archive", tag, unpack_path],
                                  stdout=subprocess.PIPE)
-        subprocess.check_call(["tar", "-x"] + (["--touch"] if touch else []), stdin=p_git.stdout)
+        # HACK: tar --touch does not deal with sub-second file mtimes, which can invert the relative mtime
+        # of package-lock.json vs. dist/*/manifest.json. Until this gets investigated/fixed properly, make sure that
+        # the files are at least one second apart
+        time.sleep(1)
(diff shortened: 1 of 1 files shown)


martinpitt:
publish-dist finished, triggering two more tests.

marusak (approved):
Honestly I have no idea why this was there. But if it unbreaks current bots, I am happy with this - less hacky - approach.

martinpitt:
Yep, [what I feared](https://logs.cockpit-project.org/logs/pull-15779-20210503-120039-507471f3-rhel-8-5-distropkg/log.html), this is the original bug. I'll add a blatant `sleep(1)` hack for now to unbreak this, and then investigate it without being in a fire.

martinpitt:
OK, that's better, both tests are past `make dist`. PTAL?

test/make_dist.py
```diff
@@ -95,20 +95,17 @@ def download_cache(wait=False):
...
         if os.path.exists(unpack_path):
             continue
         message(f"make_dist: Extracting cached {unpack_path}...")
         p_git = subprocess.Popen(["git", "--git-dir", dist_git_checkout, "archive", tag, unpack_path],
                                  stdout=subprocess.PIPE)
-        subprocess.check_call(["tar", "-x"] + (["--touch"] if touch else []), stdin=p_git.stdout)
+        # HACK: tar --touch does not deal with sub-second file mtimes, which can invert the relative mtime
+        # of package-lock.json vs. dist/*/manifest.json. Until this gets investigated/fixed properly, make sure that
+        # the files are at least one second apart
+        time.sleep(1)
```
marusak, line 107: ugh

marusak (approved):
Lets unblock ourselves, thanks!

test/make_dist.py
martinpitt: Sorry, now I remember. `tar --touch` is fine, it is the `make dist` *afterwards* which is broken -- *that* only puts truncated seconds timestamps into the newly built tarball, so that it fails in image-prepare afterwards. I'll adjust at least the commit message on landing.

martinpitt:
```
git clean -ffdx; test/make_dist.py -d; tar --full-time -tvf cockpit-*.tar.xz | egrep 'dist/.*/manifest.json|package-lock.json'
```
gives
```
-rw-r--r-- martin/martin    1296 2021-05-03 14:45:50           cockpit-243.25.g162fdb504/dist/shell/manifest.json
-rw-r--r-- martin/martin  674911 2021-05-03 14:45:50.600263599 cockpit-243.25.g162fdb504/package-lock.json
```

which is an inverted order -- manifest.json was probably something like "50.7..", it got truncated, and thus is now older than package-lock.json.

martinpitt:
Fixed properly in #15780

PR #15780, opened 2021-05-03
build: Fix `make dist` to preserve time stamps

martinpitt:
When using cached dist/ files, `make dist` is very fast, and files are
often much less than one second apart. Copying files to the temporary
dist tree with plain `tar` dropped the sub-second mtimes, which often
leads to inverted mtime ordering. E.g. while the ordering is correct in
the build tree:

    $ stat -c %y package-lock.json  dist/shell/manifest.json
    2021-05-03 15:58:09.164707531 +0200
    2021-05-03 15:58:09.256708795 +0200

the manifest.json's mtime got stored as "15:58:09" in the generated
`make dist` tarball, and suddenly become older than package-lock.json.

This caused build failures in environments without npm/node (like during
RPM package build).

Consistently use the POSIX tar format when copying file trees around, so
that all files are copied with exact mtimes. This replaces the
`sleep(1)` hack from commit 62eeb7e001.

-----

Triggering tests manually to wait for publish-dist and actually use the dist cache.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -207,9 +207,9 @@ package-lock.json: package.json
 
 install-data-local:: $(WEBPACK_INSTALL) $(MANIFESTS)
 	$(MKDIR_P) $(DESTDIR)$(pkgdatadir)
-	$(V_TAR) tar -cf - $^ | tar --no-same-owner -C $(DESTDIR)$(pkgdatadir) --strip-components=1 -xvf -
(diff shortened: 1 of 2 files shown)


martinpitt:
See https://github.com/cockpit-project/cockpit/pull/15779#issuecomment-831237312 for analysis, debugging notes, and reproducers.

marusak (approved):
\o/

PR #15781, opened 2021-05-03
workflows: Go back to tarball artifact output in build-dist

martinpitt:
While putting the individual files into the artifact zip works (done in
commit cf04588de554fa6), the artifact becomes much bigger (8.5 MB → 39
MB), much slower to create/upload (1 s → 40s [sic]), and also slower to
download/unpack (1 s → 13s).

So go back to storing the entire dist tarball as artifact (we already
have it anyway), and move the file selection to publish-dist.

-----

Tested on my fork. Compare the step times and artifact sizes of [build-dist](https://github.com/martinpitt/cockpit/runs/2493417211?check_suite_focus=true) and [publish-dist](https://github.com/martinpitt/cockpit/runs/2493492640?check_suite_focus=true) against a [build-dist](https://github.com/cockpit-project/cockpit/runs/2493135975?check_suite_focus=true) and [publish-dist](https://github.com/cockpit-project/cockpit/runs/2493208898?check_suite_focus=true) run on master.

I tested `GITHUB_BASE=martinpitt/cockpit test/image-prepare -q fedora-34` in a clean tree, and that correctly downloads and uses the cache, and successfully builds an image.

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -17,7 +17,7 @@ jobs:
           name: dist
           workflow: build-dist
           run_id: ${{ github.event.workflow_run.id }}
-          path: dist-repo
+          path: dist-unpack
 
       - name: Set up configuration and secrets
         run: |
@@ -29,9 +29,24 @@ jobs:
       - name: Commit to dist repo
         run: |
           set -ex
+          mkdir dist-repo
           cd dist-repo
+          TARS=(../dist-unpack/cockpit-*.tar.xz)
+          if [ "${#TARS[@]}" -ne 1 ]; then
+              echo "Expected exactly one tarball" >&2
+              exit 1
+          fi
+          tar --strip-components=1 -xf "${TARS[0]}"
+
+          # prevent any sneaky/buggy stuff from build-dist
+          danger="$(find -maxdepth 1 -name '.git*')"
+          if [ -n "$danger" ]; then
+              printf "Unexpected files in dist artifact:\n%s\n" "$danger" >&2
+              exit 1
+          fi
+
           git init
-          git add .
+          git add dist/ node_modules/ package-lock.json tools/debian/copyright
           git commit -m "Build for ${{ github.event.workflow_run.head_sha }}"
           tag='sha-${{ github.event.workflow_run.head_sha }}'
           git tag "$tag"
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -18,15 +18,10 @@ jobs:
       - name: Run unit-tests container to build dist
         run: |
           sudo containers/unit-tests/start dist
-          tar --strip-components=1 -xf tmp/dist/cockpit-*.tar.xz
 
       - name: Create dist artifact
         uses: actions/upload-artifact@v2
         with:
           name: dist
-          path: |
-            dist/
-            node_modules/
-            package-lock.json
-            tools/debian/copyright
+          path: tmp/dist/cockpit-*.tar.xz
           retention-days: 1


martinpitt:
After that I promise I'll be done meddling with dist/ caching for the time being :grin: Well, I still want to make build-dist faster (not build all the C stuff, not use xz extreme compression), but that can wait a bit.

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +28,11 @@ jobs:
       - name: Commit to dist repo
         run: |
           set -ex
+          mkdir dist-repo
           cd dist-repo
+          tar --strip-components=1 -xf ../cockpit-*.tar.xz
```
allisonkarlitskaya, line 33: Interestingly, it would be possible for the builder to have sent us more than one tarball, in which case this would expand to the name of the first one as a parameter to `-f` and the name of the following ones as parameters understood to be the list of files to extract.

I don't *think* there's a potential attack there, but it's certainly a bit unexpected.

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +28,11 @@ jobs:
...
           git init
-          git add .
+          git add dist/ node_modules/ package-lock.json tools/debian/copyright
```
allisonkarlitskaya, line 35: An additional hardening step I did in my `node_modules` workflow: please run `find -name '.git*' -delete` through here to make sure we aren't getting anything weird.  You're protected from the really bad stuff in the root directory, since you explicitly specify the filenames to unpack, but you could still get some unexpected things like ignore and attribute files in the subdirs.

allisonkarlitskaya (changes requested):

.github/workflows/publish-dist.yml
martinpitt: Good point! It needs to be a bit stronger though, as `-delete` does not delete directories ("find: cannot delete '...’: Directory not empty"). So something like

     find -name '.git*' -print0 | xargs -0 --no-run-if-empty rm -r --verbose

But then I realized that we can just let tar not unpack them in the first place, with `--exclude`. I did that now.

.github/workflows/publish-dist.yml
martinpitt: I added a validation that there is exactly one tarball.

martinpitt:
Updated according to @allisonkarlitskaya 's review above. [publish-dist run on my fork](https://github.com/martinpitt/cockpit/runs/2498934374?check_suite_focus=true) looks good, and I re-tested `GITHUB_BASE=martinpitt/cockpit test/image-prepare -q fedora-34`.

.github/workflows/publish-dist.yml
allisonkarlitskaya: In my case, `find` failing to delete directories returned with an error status and aborted the workflow, which was also fine.  Your approach is also very nice, though. :) I might steal it :rofl:

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +28,17 @@ jobs:
       - name: Commit to dist repo
         run: |
           set -ex
+          mkdir dist-repo
           cd dist-repo
+          TAR="$(ls ../cockpit-*.tar.xz)"
+          if [ $(echo "$TAR" | wc -l) -ne 1 ]; then
```
allisonkarlitskaya, line 34: Does this do what you think it does?  I read that as printing the list of tarballs on one line and then asking how many lines it is.

github workflows run with bash: why not just do the glob inside an array assignment and see what you get from it?

```
var=(*glob*)
echo "${#var[@]}"
```

Alternatively: is there some way we can get ourselves a bit more involved in the artifact unpack logic?

allisonkarlitskaya: One more concern while we're at it: we unpack the tarball from the zip file, openly and then create a subdirectory named `dist-repo` and do work inside of that.  Could the `.zip` file itself contain a `dist-repo/` subdirectory filled with extra goodies to trip us up  (like an initialised hostile git repo, for example?).  You give a path on creation, but not on unpacking...

allisonkarlitskaya (changes requested):

.github/workflows/publish-dist.yml
martinpitt: That's a good point - in fact, let's not even delete, but just fail the workflow if there is any ".git*" stuff inside. Because that's either a trap, or a bug of our's.

.github/workflows/publish-dist.yml
martinpitt: The .zip can contain any crap that it wants. If it contains a dist-repo, then the `mkdir dist-repo` will fail. If the built .tar contains dist-repo/, then we'll end up having dist-repo/dist-repo/, which will just be ignored.

.github/workflows/publish-dist.yml
martinpitt: Updated to hard-fail on any top-level `.git*` stuff. We *may* need to dial that back at some point if needed, but until then, better safe than sorry?

martinpitt:
Reworked again for some feedback on IRC. [publish-dist run on my fork](https://github.com/martinpitt/cockpit/runs/2499693965?check_suite_focus=true), and re-tested a local build with my forked dist cache.

.github/workflows/publish-dist.yml
martinpitt: I also changed the workflow to unpack it into a "dist-unpack/" dir, and only grab the tarball out of it.

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +29,25 @@ jobs:
...
           cd dist-repo
+          TARS=(../dist-unpack/cockpit-*.tar.xz)
+          if [ "${#TARS[@]}" -ne 1 ]; then
+              echo "Expected exactly one tarball" >&2
+              exit 1;
+          fi
+          tar --strip-components=1 -xf "${TARS[0]}"
+
+          # prevent any sneaky/buggy stuff from build-dist
+          danger="$(find -maxdepth 1 -name '.git*')"
```
allisonkarlitskaya, line 42: If one of the `node_modules` that we happen to ship starts including a `.github` directory one day this is going to trigger.  Hopefully we stop having `node_modules` here by then, though.

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +29,25 @@ jobs:
...
         run: |
           set -ex
+          mkdir dist-repo
           cd dist-repo
+          TARS=(../dist-unpack/cockpit-*.tar.xz)
+          if [ "${#TARS[@]}" -ne 1 ]; then
+              echo "Expected exactly one tarball" >&2
+              exit 1;
+          fi
+          tar --strip-components=1 -xf "${TARS[0]}"
```
allisonkarlitskaya, line 39: I actually preferred when you explicitly listed the files to extract here...

allisonkarlitskaya (approved):

.github/workflows/publish-dist.yml
martinpitt: That's what the `-maxdepth 1` is for. It now ignores somedir/.git* stuff.

.github/workflows/publish-dist.yml
martinpitt: I never did actually, as that's so darn hard with tar. (I had a prediction of the stripped path prefix in test/make_dist.py at some point, maybe you are thinking of that?)

.github/workflows/publish-dist.yml
allisonkarlitskaya: Still not a huge fan of limiting this instead of just excluding all `.git` stuff.  There were some annoying `.gitattributes` files in some npm modules that it was nice to strip out...

.github/workflows/publish-dist.yml
allisonkarlitskaya: Indeed, I'm not sure where I saw this anymore.  Too many rounds of looking at similar-but-different patches here.

I think I mentioned before: armed with the tarball name, you can efficiently and reliably determine the distdir, if you wanted to...

If we were going to make any change, though, I'd put back the `--exclude '.git*'`.  That was a really nice thing, and it would work here again, since we remove the `.zip` danger again by having a separate subdir.

You could also go full-paranoid (as I do in node_modules) and keep the gitdir separate, but that's really a matter of taste.

.github/workflows/publish-dist.yml
martinpitt: We have for example tools/.gitignore which is regularly shipped in the tarball (that's what broke [this run](https://github.com/martinpitt/cockpit/runs/2499583978?check_suite_focus=true.) where I had the "super-strict" mode). I am not actually sure why (and we can clean this up), but indeed some node_modules/somemodule/.github/ seems totally fine to me?

allisonkarlitskaya (changes requested):

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +29,25 @@ jobs:
       - name: Commit to dist repo
         run: |
           set -ex
+          mkdir dist-repo
           cd dist-repo
+          TARS=(../dist-unpack/cockpit-*.tar.xz)
+          if [ "${#TARS[@]}" -ne 1 ]; then
+              echo "Expected exactly one tarball" >&2
+              exit 1;
```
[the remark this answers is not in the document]
martinpitt: :fearful: *phew*, glad you caught that :wink: Fixed.

.github/workflows/publish-dist.yml
```diff
@@ -29,9 +29,25 @@ jobs:
...
+          fi
+          tar --strip-components=1 -xf "${TARS[0]}"
+
+          # prevent any sneaky/buggy stuff from build-dist
+          danger="$(find -maxdepth 1 -name '.git*')"
+          if [ -n "$danger" ]; then
+              echo "Unexpected files in dist artifact:" >&2
+              echo "$danger" >&2
```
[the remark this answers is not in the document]
martinpitt: Good call, thanks! Fixed.

martinpitt:
Updated to Lis' latest review, [publish-dist](https://github.com/martinpitt/cockpit/runs/2500764322?check_suite_focus=true), tested locally, etc. pp. … all good.

allisonkarlitskaya (approved):

PR #15783, opened 2021-05-03
containers: Adjust label commands in cockpit/ws Dockerfile to follow man page example

KKoukiou:
The label commands where adjusted to contain IMAGE and NAME as env variables.
With newer podman, these are apparently just passed into the environment of the
docker run call itself, instead of being string-substituted on the command line.

See containers/podman#10192

### DIFF
--- a/containers/ws/Dockerfile
+++ b/containers/ws/Dockerfile
@@ -11,9 +11,9 @@ RUN echo -e '[group_cockpit-cockpit-preview]\nname=Copr repo for cockpit-preview
 # Again see above ... we do our branching in shell script
 RUN /container/install-package.sh && /container/prep-container.sh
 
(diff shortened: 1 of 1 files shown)


KKoukiou:
I built the container locally and ran the runlabel command:
```
$ podman container runlabel install cockpit/ws
cockpit-ws must not be installed in the host
Error: `/proc/self/exe run --rm --privileged -v /:/host -e IMAGE=quay.io/cockpit/ws:latest quay.io/cockpit/ws:latest /container/atomic-install quay.io/cockpit/ws:latest` failed: exit status 1

```
Seems the the IMAGE, NAME are replaced correctly.

martinpitt (approved):
Thanks! Makes me wonder if we could replace the /usr/bin/docker with something that works for podman as well? It seems it already substitutes it with /proc/self/exe, but it still looks odd.

PR #13826, opened 2020-03-31
tools: spec file adjustments for openSUSE

lnussel:
I don't know if you want to litter your spec file with our stuff but worth a try I guess :)
Apart from the usual package name differences there are also some files packages twice or not owned directories caught.
The debug trick doesn't work, the macro has different content on openSUSE. Also, openSUSE doesn't like stale symlinks.

PR #15784, opened 2021-05-04
tools: Dynamically manage motd/issue symlinks in package scripts

martinpitt:
Users commonly want to just remove /etc/motd.d/cockpit or
/etc/issue.d/cockpit.issue, instead of symlinking them to /dev/null.
However, rpms' %config handling does not respect that, and brings back
the files on upgrade. This also raises an alarm in `rpm --verify`.

Likewise, dpkg's conffile logic only handles plain file content changes
or removals, not symlinks.

Stop packaging the symlinks, and create/remove them in the package
scrips instead.

https://bugzilla.redhat.com/show_bug.cgi?id=1876848

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -537,6 +537,57 @@ class TestConnection(MachineCase):
 
         self.allow_journal_messages(".*Peer failed to perform TLS handshake")
 
+    @skipImage("Can't remove/upgrade packages on OSTree", "fedora-coreos")
+    def testWsPackage(self):
+        m = self.machine
+
+        if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+            # clean up debug symbols, they get in the way of upgrading
+            m.execute("dpkg --purge cockpit-ws-dbgsym")
+        elif m.image.startswith("rhel"):
+            # subscription-manager-cockpit depends on cockpit-ws
+            m.execute("rpm --erase subscription-manager-cockpit")
+
+        def install():
+            if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+                m.execute("dpkg --install /var/tmp/build-results/cockpit-ws_*.deb")
+            else:
+                # FIXME: current package complains about mode change on /run/cockpit and selinux
+                m.execute("if rpm -q cockpit-ws; then rpm -V --nomode cockpit-ws; fi")
+                m.execute("rpm --upgrade --force /var/tmp/build-results/cockpit-ws-*.rpm")
+                m.execute("rpm -V --nomode cockpit-ws")
+
+        def remove():
+            if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+                m.execute("dpkg --purge cockpit cockpit-ws")
+            else:
+                m.execute("rpm --erase cockpit cockpit-ws")
+
+        # upgrade from distro version (our images have cockpit-ws preinstalled) sets up dynamic motd/issue symlink
+        self.assertEqual(m.execute("readlink /etc/motd.d/cockpit").strip(), "/run/cockpit/motd")
+        self.assertEqual(m.execute("readlink /etc/issue.d/cockpit.issue").strip(), "/run/cockpit/motd")
+
+        # package upgrade keeps them
+        install()
+        self.assertEqual(m.execute("readlink /etc/motd.d/cockpit").strip(), "/run/cockpit/motd")
+        self.assertEqual(m.execute("readlink /etc/issue.d/cockpit.issue").strip(), "/run/cockpit/motd")
+
+        # manual change/removal is respected on upgrade
+        m.execute("ln -sf /dev/null /etc/motd.d/cockpit; rm /etc/issue.d/cockpit.issue")
+        install()
+        self.assertEqual(m.execute("readlink /etc/motd.d/cockpit").strip(), "/dev/null")
+        m.execute("test ! -e /etc/issue.d/cockpit.issue")
+
+        # removing the package cleans up the links
+        remove()
+        m.execute("test ! -e /etc/motd.d/cockpit")
+        m.execute("test ! -e /etc/issue.d/cockpit.issue")
+
+        # fresh install (most of our test images have cockpit-ws preinstalled, so the first test above does not cover that)
+        install()
+        self.assertEqual(m.execute("readlink /etc/motd.d/cockpit").strip(), "/run/cockpit/motd")
+        self.assertEqual(m.execute("readlink /etc/issue.d/cockpit.issue").strip(), "/run/cockpit/motd")
+
     @skipImage("OSTree doesn't have cockpit-ws", "fedora-coreos")
     def testCommandline(self):
         m = self.machine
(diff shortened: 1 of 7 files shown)


martinpitt:
Oops, on [RHEL](https://logs.cockpit-project.org/logs/pull-15784-20210504-080143-7dba87c7-rhel-8-5/log.html#277) subscription-manager-cockpit depends on cockpit-ws. It should not do that (I'll file a PR), but we can simply work around that in our tests.

test/verify/check-connection
```diff
@@ -537,6 +537,57 @@ class TestConnection(MachineCase):
...
+            m.execute("dpkg --purge cockpit-ws-dbgsym")
+        elif m.image.startswith("rhel"):
+            # subscription-manager-cockpit depends on cockpit-ws
+            m.execute("rpm --erase subscription-manager-cockpit")
+
+        def install():
+            if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+                m.execute("dpkg --install /var/tmp/build-results/cockpit-ws_*.deb")
+            else:
+                # FIXME: current package complains about mode change on /run/cockpit and selinux
```
marusak, line 555: What is it complaining about? Is this related to these changes?

martinpitt: No, unrelated:
```
# rpm -V cockpit-ws
.M.......  g /run/cockpit
.M.......  g /var/lib/selinux/targeted/active/modules/200/cockpit
```
The first one has been there forever (/run/cockpit is just a %ghost), and I figure the other one was introduced with adopting our SELinux policy. We should fix these, of course, but for now this makes sure that  we don't introduce other complains than 'M'ode changes.

marusak (dismissed):
Thanks!

martinpitt:
@KKoukiou Do you still want to review this one also? (happy to explain packaging details, if you want to learn)

KKoukiou:
> @KKoukiou Do you still want to review this one also? (happy to explain packaging details, if you want to learn)

Yes, will do it right now. Matej was very fast :)

martinpitt:
I sent https://github.com/candlepin/subscription-manager/pull/2617 to drop the cockpit-ws dependency from subscription-manager-cockpit.

tools/cockpit.spec
```diff
@@ -551,6 +559,11 @@ if %{_sbindir}/selinuxenabled 2>/dev/null; then
 fi
 %endif
 %systemd_postun_with_restart cockpit.socket cockpit.service
+# clean up dynamic motd/issue symlinks on removal
```
KKoukiou, line 562: Is this really needed? I read in http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html that ghost files are also removed when the package is.

martinpitt: Excellent point -- I'll test this! (easy enough with the integration tests now). If it does, I'll drop that bit. Setting needswork for now.

martinpitt:
Nice! I locally tested this on Fedora 34, and it does clean up the links on removal without the `%postrm` stuff. Pushed, let's validate it everywhere.

KKoukiou (approved):

martinpitt:
[unrelated flake](https://logs.cockpit-project.org/logs/pull-15784-20210505-154339-e5975891-fedora-33-firefox/log.html#106); retrying, I'll look at that separately tomorrow.

martinpitt:
Doesn't want to budge -- I suppose this is related to switching TEST_OS_DEFAULT to f34 today, but firefox test is still f33? So check-shell-host-switching does not disable preloads on the auxiliary machines any more, which changed the timing.

martinpitt:
I added some more fixes which will hopefully fix the above failure.

test/verify/check-connection
martinpitt: @lnussel : You added this `%ghost %dir /run/cockpit` in  PR #13826. What was the purpose of that? It  breaks `rpm -V` validation, and this should really not be a directory that rpm should be concerned about. It's managed by systemd and cockpit's units entirely, not by rpm. Would it be ok to drop that again?

martinpitt:
Whee, seems I tamed check-shell-host-switching at last :partying_face:

marusak (approved):
Good call with removing of ghosts files, thanks :)

test/verify/check-connection
lnussel: Well you could remove it, it's not mandatory. But shouldn't hurt either. For now it mostly serves the purpose you see the, ie rpm -V. Why is there a disagreement about the mode of the directory?

martinpitt: I have no idea.. On the actual system it's just standard root:root 0755. But as the rpm does not actually ship that dir, and `%ghost %dir` does not declare any explicit mode, I'm not sure what the expected mode even is.. But maybe it's not *really* about the mode, but it's complaining about the "g"?

     g %ghost file (i.e. the file contents are not included in the package payload).

then it seems pretty stupid to complain about this, because not shipping a file in the package is *exactly* what %ghost is for..

lnussel: The rpm does include a mode for the ghost file as it's actually created in %install. So from what I can tell it looks like doing the right thing here.

```
# rpm -qvl cockpit-ws|grep /run/cockpit$
drwxr-xr-x    2 root     root                        0 Apr 22 06:57 /run/cockpit
# rpm -Vv cockpit-ws|grep /run/cockpit$
.........  g /run/cockpit
# chmod 700 /run/cockpit
# rpm -V cockpit-ws
.M.......  g /run/cockpit
# rpm --restore cockpit-ws
# l -d /run/cockpit
drwxr-xr-x 2 root root 100  7. Mai 05:33 /run/cockpit/
# rpm -V cockpit
#
```

martinpitt: I just checked the .spec again, and indeed that bit happens *only*  on SUSE:
```sh
# need this in SUSE as post build checks dislike stale symlinks
install -m 644 -D /dev/null %{buildroot}/run/cockpit/motd
```
So I suppose I'll conditionalize the `%ghost` as well, as the Fedora/RHEL packages don't ship bits in /run (and they really *really* shouldn't! Packaging /run makes me cringe. Are you sure you want that in OpenSUSE?)

martinpitt: I believe this OpenSUSE-specific `install` hack should be obsolete since commit 5705709d416954c169e9af442ea3460ad48b6e9c . I'll send a PR after landing my `rpm -V` fix, and ask @lnussel for confirmation.

lnussel: %ghost in /run doesn't really package anything :-) Anyway feel free to remove, I have to maintain some openSUSE specific patches anyway (license, permission macro etc).
Installing links to /run in /etc makes me cringe btw :-) agetty, login, pam etc do support reading motd and issue from /run. Looks like just openssh doesn't.

martinpitt: Right, but it keeps information about the expected modes/properties of the ghosted file/dir. Full ack about the installing links to /etc.. We still have some distros to support which don't yet support reading from /run/motd.d/ directly (that's fixed in most recent PAM, though). One of these days this will go away.

PR #15787, opened 2021-05-04
test: Don't fail when a session has closed before we could terminate it

mvollmer:
Example failure: https://logs.cockpit-project.org/logs/pull-152-20210504-131857-cb98179e-fedora-34/log.html#43-1


marusak (approved):
Thanks!

PR #15790, opened 2021-05-05
storage: Luks modal body fixes

### LINKED ISSUE #15773: LUKS key slot removal confirmation removes the wrong key slot
Cockpit version: 242
OS: Fedora 33
Page: Storage

When deleting a LUKS key slot with passphrase, Cockpit asks to enter a passphrase to confirm. If I enter a passphrase from another slot in this form, instead of deleting the selected passphrase slot, the one with passphrase I've entered is deleted.

Steps to reproduce

0. Create a new encrypted volume
1. Set key slot 0 passphrase to `passphrase0`
2. Add key slot 1 with `passphrase1`
3. Remove key slot 1, upon confirmation enter `passphrase0`
4. Observe key slot 0 removed

marusak:
- The first commit changes logic of how we remove keys. Fixes #15773

- The second commit changes UI:
Before it looked like this:
removepassphrase
Now it looks like this:
afternewkeyslot
forceremoval
uponerror

-The third commit fixes layout in `Add key` modal
Before it looked like this:
lastcommitbefore
Now it looks like this:
lastcommitfun

@garrett WDYT? One thing is about 'Remaining passphrase' helper text. We probably want to mention that user should type in different passphrase, not the one that is removing. Should we just change label?

Needs tests fixes, working on them now.

### DIFF
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -124,7 +124,7 @@ export function clevis_recover_passphrase(block) {
             .then(output => output.trim());
 }
 
-/* Passphrase operations
+/* Passphrase and slot operations
  */
 
 function passphrase_add(block, new_passphrase, old_passphrase) {
@@ -139,19 +139,20 @@ function passphrase_change(block, key, new_passphrase, old_passphrase) {
                          { superuser: true, err: "message" }).input(old_passphrase + "\n" + new_passphrase + "\n");
 }
 
-function passphrase_remove(block, passphrase) {
-    var dev = decode_filename(block.Device);
-    return cockpit.spawn(["cryptsetup", "luksRemoveKey", dev],
-                         { superuser: true, err: "message" }).input(passphrase);
-}
+function slot_remove(block, slot, passphrase) {
+    const dev = decode_filename(block.Device);
+    const opts = { superuser: true, err: "message" };
+    const cmd = ["cryptsetup", "luksKillSlot", dev, slot.toString()];
+    if (passphrase === false) {
+        cmd.splice(2, 0, "-q");
+        opts.pty = true;
+    }
 
-/* Generic slot operations
- */
+    const spawn = cockpit.spawn(cmd, opts);
+    if (passphrase !== false)
+        spawn.input(passphrase + "\n");
 
-function slot_remove(block, slot) {
-    var dev = decode_filename(block.Device);
-    return cockpit.spawn(["cryptsetup", "luksKillSlot", "-q", dev, slot.toString()],
-                         { superuser: true, err: "message", pty: true });
+    return spawn;
 }
 
 /* Dialogs
@@ -409,10 +410,7 @@ function remove_passphrase_dialog(block, key) {
             DangerButton: true,
             Title: _("Remove"),
             action: function (vals) {
-                if (vals.passphrase === false)
-                    return slot_remove(block, key.slot);
-                else
-                    return passphrase_remove(block, vals.passphrase);
+                return slot_remove(block, key.slot, vals.passphrase);
             }
         }
     });
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -21,9 +21,10 @@ import cockpit from "cockpit";
 import React from "react";
 
 import {
-    Card, CardBody, CardTitle, CardHeader, CardActions,
+    Card, CardBody, CardTitle, CardHeader, CardActions, Checkbox,
+    Form, FormGroup,
     DataListItem, DataListItemRow, DataListItemCells, DataListCell, DataList,
-    Text, TextVariants
+    Text, TextVariants, TextInput as TextInputPF, Stack,
 } from "@patternfly/react-core";
 import { EditIcon, MinusIcon, PlusIcon, ExclamationTriangleIcon } from "@patternfly/react-icons";
 
@@ -357,44 +358,38 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
 }
 
 const RemovePassphraseField = (tag, key, dev) => {
+    function validate(val) {
+        if (val === "")
+            return _("Passphrase can not be empty");
+    }
+
     return {
         tag: tag,
-        title: <ExclamationTriangleIcon className="ct-icon-exclamation-triangle" size="lg" />,
-        options: { },
+        title: null,
+        options: { validate: validate },
         initial_value: "",
+        bare: true,
 
-        render: (val, change) => {
-            const key_slot = cockpit.format(_("key slot $0"), key.slot);
+        render: (val, change, validated, error) => {
             return (
-                <div data-field={tag}>
-                    <h3>
-                        { fmt_to_fragments(_("Remove passphrase in $0?"), <b>{key_slot}</b>) }
-                    </h3>
+                <Stack hasGutter>
                     <p>{ fmt_to_fragments(_("Passphrase removal may prevent unlocking $0."), <b>{dev}</b>) }</p>
-
-                    <div name="remove-passphrase" className="progressive-disclosure ct-form">
-                        <div className="form-group">
-                            <label>
-                                <input type="radio" checked={val !== false}
-                                       autoFocus
-                                       onChange={event => change("")} />
-                                {_("Confirm removal with passphrase")}
-                            </label>
-                            <input className="form-control" type="password" hidden={val === false}
-                                   value={val} onChange={event => change(event.target.value)} />
-                        </div>
-                        <div className="form-group">
-                            <label>
-                                <input type="radio" checked={val === false}
-                                       onChange={event => change(false)} />
-                                { fmt_to_fragments(_("Force remove passphrase in $0"), <b>{key_slot}</b>) }
-                            </label>
-                            <p className="slot-warning" hidden={val !== false}>
-                                Removing a passphrase without confirmation can be dangerous.
-                            </p>
-                        </div>
-                    </div>
-                </div>
+                    <Form>
+                        <Checkbox id="force-remove-passphrase"
+                                  isChecked={val !== false}
+                                  label={_("Confirm removal with an alternate passphrase")}
+                                  onChange={checked => change(checked ? "" : false)}
+                                  body={val === false
+                                      ? <p className="slot-warning">
+                                          {_("Removing a passphrase without confirmation of another passphrase may prevent unlocking or key management, if other passphrases are forgotten or lost.")}
+                                      </p>
+                                      : <FormGroup label={_("Passphrase from any other key slot")} fieldId="remove-passphrase">
+                                          <TextInputPF id="remove-passphrase" type="password" value={val} onChange={value => change(value)} />
+                                      </FormGroup>
+                                  }
+                        />
+                    </Form>
+                </Stack>
             );
         }
     };
@@ -402,10 +397,11 @@ const RemovePassphraseField = (tag, key, dev) => {
 
 function remove_passphrase_dialog(block, key) {
     dialog_open({
-        Title: _("Remove passphrase"),
+        Title: <><ExclamationTriangleIcon className="ct-icon-exclamation-triangle" /> {cockpit.format(_("Remove passphrase in key slot $0"), key.slot)}</>,
         Fields: [
             RemovePassphraseField("passphrase", key, block_name(block))
         ],
+        isFormHorizontal: false,
         Action: {
             DangerButton: true,
             Title: _("Remove"),
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -415,15 +415,16 @@ function remove_passphrase_dialog(block, key) {
 const RemoveClevisField = (tag, key, dev) => {
     return {
         tag: tag,
-        title: <ExclamationTriangleIcon className="ct-icon-exclamation-triangle" size="lg" />,
+        title: null,
         options: { },
         initial_value: "",
+        bare: true,
 
         render: (val, change) => {
             return (
                 <div data-field={tag}>
-                    <h3>{ fmt_to_fragments(_("Remove $0?"), <b>{key.url}</b>) }</h3>
-                    <p>{ fmt_to_fragments(_("Keyserver removal may prevent unlocking $0."), <b>{dev}</b>) }</p>
+                    <p>{ fmt_to_fragments(_("Remove $0?"), <b>{key.url}</b>) }</p>
+                    <p className="slot-warning">{ fmt_to_fragments(_("Keyserver removal may prevent unlocking $0."), <b>{dev}</b>) }</p>
                 </div>
             );
         }
@@ -432,7 +433,7 @@ const RemoveClevisField = (tag, key, dev) => {
 
 function remove_clevis_dialog(client, block, key) {
     dialog_open({
-        Title: _("Remove Tang keyserver"),
+        Title: <><ExclamationTriangleIcon className="ct-icon-exclamation-triangle" /> {_("Remove Tang keyserver")}</>,
         Fields: [
             RemoveClevisField("keyserver", key, block_name(block))
         ],
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -253,7 +253,7 @@ const Row = ({ field, values, errors, onChange }) => {
         onChange(tag);
     }
 
-    const children = field.render(values[tag], change, validated);
+    const children = field.render(values[tag], change, validated, error);
 
     if (title || title == "") {
         let titleLabel = title;
@@ -271,14 +271,15 @@ const Row = ({ field, values, errors, onChange }) => {
                 { children }
             </FormGroup>
         );
-    } else {
+    } else if (!field.bare) {
         return (
             <FormGroup validated={validated}
                        helperTextInvalid={error} helperText={explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
                 { children }
             </FormGroup>
         );
-    }
+    } else
+        return children;
 };
 
 function is_visible(field, values) {
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -226,7 +226,7 @@ import {
     Form, FormGroup,
     Radio,
     Select as TypeAheadSelect, SelectOption, SelectVariant,
-    Spinner,
+    Spinner, Split,
     TextInput as TextInputPF4,
     Tooltip, TooltipPosition,
 } from "@patternfly/react-core";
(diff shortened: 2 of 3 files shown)


pkg/storaged/crypto-keyslots.jsx
```diff
@@ -353,64 +342,67 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+                title={cockpit.format(_("Remove passphrase in key slot $0"), this.props.slot)}
+                titleIconVariant="warning"
+                isOpen
+                onClose={this.props.onClose}
+                footer={<>
+                    {
+                        this.state.error && <ModalError dialogError={this.state.error} />
+                    }
+                    <Button key="remove" variant="danger" isLoading={this.state.pending} isDisabled={this.state.pending} onClick={this.onRemove}>{_("Remove")}</Button>
+                    <Button key="cancel" variant="link" isDisabled={this.state.pending} onClick={this.props.onClose}>{_("Cancel")}</Button>
```
KKoukiou, line 381: Disabling the cancel only is partial, you need also to disable Esc close if you want to be accurate.
grep this onEscapePress={() => undefined}

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -21,9 +21,10 @@ import cockpit from "cockpit";
 import React from "react";
 
 import {
-    Card, CardBody, CardTitle, CardHeader, CardActions,
+    Button, Card, CardBody, CardTitle, CardHeader, CardActions, Checkbox,
+    Form, FormGroup,
     DataListItem, DataListItemRow, DataListItemCells, DataListCell, DataList,
-    Text, TextVariants
+    Text, TextVariants, Modal, TextInput as TextInputPF, Stack,
```
KKoukiou, line 27: Why you renamed the import?

marusak: Because TextInput is imported also from `storagelib.js` a few lines below

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -139,19 +139,17 @@ function passphrase_change(block, key, new_passphrase, old_passphrase) {
...
-}
+    const cmd = ["cryptsetup", "luksKillSlot", dev, slot.toString()];
+    if (passphrase === false)
+        cmd.splice(2, 0, "-q");
 
-/* Generic slot operations
- */
+    const spawn = cockpit.spawn(cmd, { superuser: true, err: "message" });
+    if (passphrase !== false)
+        spawn.input(passphrase + "\n");
```
mvollmer: Note that luksKillSlot will succeed if `passphrase` is any valid passphrase it doesn't need to be the one in the given slot.
So we still have the issue that people might accidentally remove the wrong slot.

mvollmer: Man page: " Except running  in  batch-mode (-q) a remaining passphrase must be supplied"  Does that maybe even mean that one must provide a passphrase that is in some other slot than the one that is about to be removed, as a proof that you still know how to unlock this thing after removing the slot?

marusak: So the thing is: Now you are removing slot by typing its passphrase. If you click on 'Remove slot 3' but write down passphrase for slot 2 it will remove slot 2. See  #15773

With this change we will remove only the slot that you select from UI and you need to write _any other_ passphrase (not the one that you are removing).

marusak: >  Does that maybe even mean that one must provide a passphrase that is in some other slot than the one that is about to be removed, as a proof that you still know how to unlock this thing after removing the slot?

Exactly.

garrett (changes requested):
### Incorrect alert placement

When the password is submitted and is incorrect, the error should be at the top, not the bottom. See https://www.patternfly.org/v4/components/form/design-guidelines#error-validation-on-submission:

> For validation on submission, supplement field level errors with an inline alert to make it simple for the user to quickly identify and address areas that need attention. The **inline alert should be placed below the form’s header and description**, and it should equal the total width of the form.

(Added emphasis)

### Odd phrase?

What does "Use remaining passphrase" mean?
 
### Alignment and spacing issues

Alignment of "Key source" radios is not correct. The radios should be aligned with the baseline of the label. There's some option for this in React/CSS (...but I forget what it's called... was it something to do with the first item or inline or something else?)

It feels like there's too much vertical spacing in the dialogs. I guess there needs to be extra spacing (or some other separator) between the configuration and the confirmation/unlock password, but not really the rest.

### Clarification? (Perhaps a later PR?)

Also, I think "Removing a passphrase without confirmation may be dangerous." might need some rewording. It's what we currently use, what's in the mockup, and what's in this PR. But it doesn't really specify how it could be dangerous or what it may do. (But this could be solved in a follow-up.) 

From my understanding, it's possible to remove the wrong passphrase and a disk may not mount. This could result in data not being able to be accessed or even the computer to no longer boot. We might want to _explicitly_ state this. (Again, it's fine to defer this to another PR, as all the rest are clear improvements.)

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -355,67 +342,67 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+                    <p>{ fmt_to_fragments(_("Passphrase removal may prevent unlocking $0."), <b>{block_name(this.props.block)}</b>) }</p>
+                    <Checkbox id="run-image-dialog-tty"
+                              isChecked={!this.state.force}
+                              label={_("Confirm removal")}
+                              onChange={checked => this.setState({ force: !checked })}
+                              body={this.state.force
+                                  ? <p className="slot-warning">
+                                      {_("Removing a passphrase without confirmation can be dangerous.")}
+                                  </p>
+                                  : <Form isHorizontal>
```
KKoukiou, line 395: Conceptually the Form should contain also the Checbox. It's part of the form right :) ?

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -355,67 +342,67 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+                        this.state.error && <ModalError dialogError={this.state.error} />
+                    }
+                    <Button key="remove" variant="danger" isLoading={this.state.pending} isDisabled={this.state.pending} onClick={this.onRemove}>{_("Remove")}</Button>
+                    <Button key="cancel" variant="link" isDisabled={this.state.pending} onClick={this.props.onClose}>{_("Cancel")}</Button>
+                </>}
+                position="top" variant="medium"
+            >
+                <Stack hasGutter>
+                    <p>{ fmt_to_fragments(_("Passphrase removal may prevent unlocking $0."), <b>{block_name(this.props.block)}</b>) }</p>
+                    <Checkbox id="run-image-dialog-tty"
```
KKoukiou, line 387: This should be in the Form and then inside a FormGroup.

KKoukiou: You wil need hasNoPaddingTop property in the FormGroup.

KKoukiou (commented):
So in all pages where we have forms in dialogs, the Form is encapsulating all elements or form fields. For nested fields we use nested form groups.
If this approach visually works I don't have an issue, but also it's nice to have a more consistent approach.

marusak:
> What does "Use remaining passphrase" mean?

Please read the first comment till the end, I was asking about it.

> Also, I think "Removing a passphrase without confirmation may be dangerous." might need some rewording...

Right, the point being if you remove the last passphrase you know you won't be able to unlock it again. Thus you are asked to supply some different passphrase as confirmation that you still have access. If you have idea for better wording please let me know.

pkg/storaged/crypto-keyslots.jsx
marusak: > This should be in the Form and then inside a FormGroup.

it already is. It wraps this into `Form` and `FormGroup`. And this was adding additional `FormGroup` into `FormGroup`.

marusak: Good call with ` hasNoPaddingTop`, with that it looks like this:
padding

pkg/storaged/crypto-keyslots.jsx
marusak: Not really - this form is part of checkbox body

pkg/storaged/crypto-keyslots.jsx
mvollmer: > With this change we will remove only the slot that you select from UI and you need to write any other passphrase (not the one that you are removing).

Oh, I always understood the dialog the other way around: "please type the passphrase you want to remove".

garrett:
So, what do you think of something like these two states?

---

☑ Confirm slot passphrase removal with an alternate passphrase
      Passphrase from any other key slot: [_______________]

---

☐ Confirm slot passphrase removal with an alternate passphrase

Removing a passphrase without confirmation of another passphrase may prevent unlocking or key management, if other passphrases are forgotten or lost.

marusak:
This new label breaks to two lines:
toolongtext
@KKoukiou do you have idea how to tell `FormGroup` not to do that? Or do you think it is fine this way?

pkg/storaged/crypto-keyslots.jsx
KKoukiou: Right, but what I saw is that it's conceptually wrong to leave some fields outside of the form. In this case the checkbox. 

Copying from https://www.patternfly.org/v4/components/form/design-guidelines: 
A form is a group of elements used to collect information from a user in a variety of contexts including in a modal

Whenever we use forms in modals we put all fields of the modals inside the form. I don't understand why this is special.

marusak:
Wrapped it in Form as @KKoukiou wanted me to :) It would not be possible with `isHorizontal` but I think we might not want it anyway as it wraps the label unnecessarily. Looks like this now:
witherrorform
forcenewstring

marusak:
Ready for review

pkg/storaged/crypto-keyslots.jsx
KKoukiou: This is broken all over cockpit, let's not blick this PR on this.

KKoukiou (dismissed):

pkg/storaged/crypto-keyslots.jsx
marusak: Oh I forgot about this

mvollmer:
So I am thinking Cockpit should still have a way to run luksRemoveKey (in addition to luksKillSlot), for the case where you know the passphrase that you want to remove, but don't know what slot it is. We could have a "global" button for this, next to the button for adding a new passphrase.

Also, if inputting a passphrase is now a way to proof to yourself that you are still able to unlock the thing after killing a slot, we should also ask for it when removing non-passphrase slots, like the clevis ones.  (Maybe this happens, I should run the code I guess... :-)

mvollmer:
> Also, if inputting a passphrase is now a way to proof to yourself that you are still able to unlock the thing after killing a slot,

Also, this could take into account whether there is a clevis slot and offer to test them for you...

mvollmer:
Removing a tang server looks like this:

image

I think it should use the same new style with the warning icon in the title etc.

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -355,67 +342,68 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+            spawn.input(this.state.passphrase + "\n");
+
+        spawn.then(this.props.onClose)
+                .catch(e => this.setState({ error: e.message }))
+                .always(() => this.setState({ pending: false }));
+    }
+
+    render() {
+        return (
+            <Modal
```
mvollmer, line 374: I don't think we should avoid the dialog abstraction in dialog.jsx unless it makes things really really hard. Rather, we should extend the abstraction to make totally custom layout possible. I'll give that a shot.

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -355,67 +342,68 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+                title={cockpit.format(_("Remove passphrase in key slot $0"), this.props.slot)}
+                titleIconVariant="warning"
+                isOpen
+                onClose={this.props.onClose}
+                actions={[
+                    <Button key="remove" variant="danger" isLoading={this.state.pending} isDisabled={this.state.pending} onClick={this.onRemove}>{_("Remove")}</Button>,
```
mvollmer, line 380: The rest of the storage dialogs don't put the spinner inside the button, so let's not do it here either.

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -355,67 +342,68 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+                    <Button key="cancel" variant="link" isDisabled={this.state.pending} onClick={this.props.onClose}>{_("Cancel")}</Button>
+                ]}
+                position="top" variant="medium"
+            >
+                <Stack hasGutter>
+                    { this.state.error && <ModalError dialogError={this.state.error} /> }
```
mvollmer, line 386: The rest of the storage dialogs put the errors resulting from an action into the footer. Let's keep doing this here as well.

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -355,67 +342,68 @@ function edit_tang_adv(client, block, key, url, adv, passphrase) {
...
+                        <Checkbox id="force-remove-passphrase"
+                                  isChecked={!this.state.force}
+                                  label={_("Confirm removal with an alternate passphrase")}
+                                  onChange={checked => this.setState({ force: !checked })}
+                                  body={this.state.force
+                                      ? <p className="slot-warning">
+                                          {_("Removing a passphrase without confirmation of another passphrase may prevent unlocking or key management, if other passphrases are forgotten or lost.")}
+                                      </p>
+                                      : <FormGroup label={_("Passphrase from any other key slot")} fieldId="remove-passphrase">
+                                          <TextInputPF id="remove-passphrase" type="password" value={this.state.passphrase} onChange={value => this.setState({ passphrase: value })} />
```
mvollmer, line 398: I think we should reject empty passphrases.

mvollmer (changes requested):

pkg/storaged/crypto-keyslots.jsx
mvollmer: Here is what I have so far: https://github.com/cockpit-project/cockpit/compare/master...mvollmer:storage-new-remove-key-layout-experiment

What do you think?

garrett:
>  I am thinking Cockpit should still have a way to run luksRemoveKey (in addition to luksKillSlot), for the case where you know the passphrase that you want to remove, but don't know what slot it is.

If you do remove the passphrase and don't know any others, then it'll prevent unlocking, right? (So this is a possibly dangerous action?)

> I think it should use the same new style with the warning icon in the title etc.

Agreed. It should be "Remove Tang keyserver at 127.0.0.1" for the title.

mvollmer:
> If you do remove the passphrase and don't know any others, then it'll prevent unlocking, right? (So this is a possibly dangerous action?)

Yes, so we might want to have the same protection there: "Passphrase from any other slot".

However, the `luksRemoveKey` command will not ask for a remaining passphrase.  Only `luksKillSlot` will do this, so there is no convenient way to implement this, I think.

garrett:
> However, the luksRemoveKey command will not ask for a remaining passphrase. Only luksKillSlot will do this, so there is no convenient way to implement this, I think.

_If_ we implement it, then it should have a "dangerous" status. (https://www.patternfly.org/v4/components/modal/html/#danger-alert) And we should let someone know why it's dangerous too.

But I think it's time that we should discuss it in a separate issue, and it should get its own PR if we do want to implement it.

pkg/storaged/crypto-keyslots.jsx
marusak: Good call! I've picked the idea and pushed it here. Content of dialogs looks the same, just the dialog itself looks as rest of the dialogs.
Thanks!

pkg/storaged/crypto-keyslots.jsx
marusak: I've went back to using generic dialog, so now all dialogs looks the same.

marusak:
@mvollmer can you please take a look now?
Also fixed the tang remove modal dialog:
remotang

garrett:
>  Also fixed the tang remove modal dialog:

That's not really fixed though?

It should look like this:

storage-tang-removal

This isn't pixel perfect, but it's approximately right. Use default PF4 spacing (which should look close to this).

The width is `sm`; the text is normal. We use other warning text in red like this elsewhere. The removal button is the correct shade of red (whereas your screenshot is darker).

It's mainly following https://www.patternfly.org/v4/components/modal/design-guidelines/#confirm-a-destructive-action, except with the red danger text at the bottom.

KKoukiou:
@garrett the mockup looks a bit weird, since it's a 'warning' modal dialog with a 'danger' alert message. As if we can't decide if it's a error or a warning.

marusak:
> The width is `sm`

out of scope of this PR. This PR does not change the dialog itself at the end, only the content

> The removal button is the correct shade of red (whereas your screenshot is darker).

ditto

garrett:
It's a warning dialog with something that's _possibly_ dangerous. We could just shift it fully to danger, but PF usually uses the danger icon for errors.

garrett:
> out of scope of this PR

It says Luks modal fixes, and this is a Luks modal, right?

Feel free to address it in an immediate follow up if you'd rather.

KKoukiou:
> It's a warning dialog with something that's _possibly_ dangerous. We could just shift it fully to danger, but PF usually uses the danger icon for errors.

Ok correct, PF quite super clear [guidelines for this ](https://www.patternfly.org/v4/components/modal/design-guidelines/#confirm-a-destructive-action).

```
If an action is difficult or impossible to undo, add a warning icon beside the headline.
```

So we should use warning modal. But the red text is not used at all in their examples. And afaik there is not any component which offers danger colored text apart from the FormHelperText, https://www.patternfly.org/v4/components/form/#invalid-with-form-alert 
But this is not a form :)

marusak:
> The removal button is the correct shade of red (whereas your screenshot is darker).

Screenshot had focus on that button thats why it was darker :)

Looks now like this:
redredred

marusak:
@KKoukiou about the red text - fair point that we don't do that except when field does not validate.
Here are screenshots how it looks without the red color: (not yet pushed)

twoohnered
redredred


oneohnered
forcenewstring

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -432,7 +433,7 @@ const RemoveClevisField = (tag, key, dev) => {
 
 function remove_clevis_dialog(client, block, key) {
     dialog_open({
-        Title: _("Remove Tang keyserver"),
+        Title: <><ExclamationTriangleIcon className="ct-icon-exclamation-triangle" /> {_("Remove Tang keyserver")}</>,
```
KKoukiou, line 436: To make modals with warnings you are supposed to pass the titleIconVariant="warning" to the Modal component directly.
In this case you need to add the a new property to the Dialog in cockpit-components-dialog. 

In anycase, we can do it in followup.

KKoukiou (approved):

pkg/storaged/crypto-keyslots.jsx
marusak: Yeah, I wanted to only focus on the body of dialogs in this PR as there are a lot of things we should do with modals themselves - like this alert, size of them, spinner position when action is happening...

garrett (approved):
LGTM :+1:

marusak:
@mvollmer wanna take a look as well?

mvollmer (approved):
Code looks good.

I think "Remove slot and not passphrase" and "Fix layout of key slot removal dialog" could be combined into a single commit, since the latter has the  necessary UI string changes for the new functionality of the former.

marusak:
> I think "Remove slot and not passphrase" and "Fix layout of key slot removal dialog" could be combined into a single commit, since the latter has the necessary UI string changes for the new functionality of the former.

ah right, might be. But don't want to re-run all the test and collect acks so I'll merge it as is. In any case, the old UI was not clear anyway :)

PR #15793, opened 2021-05-05
packit: Disable tests

martinpitt:
Tests got broken with the reorganization of the dist cache in commit
cf04588de554. To fix them, the sandcastle container needs to get a few
more packages pre-installed [1], this needs to be rolled out to
production, and then we need to adjust packit.yaml [2].

Until then, disable the tests as they immediately fail to build the SRPM
anyway.

[1] https://github.com/packit/sandcastle/pull/122
[2] https://github.com/cockpit-project/cockpit/pull/15778


martinpitt:
I do this with a heavy heart, but right now all our PRs are red, and the tests don't run anyway.

KKoukiou (approved):

PR #15795, opened 2021-05-06
test: Test Grafana integration via CLI

martinpitt:
Demonstrate how to export PCP metrics from a client machine into an
already existing Grafana installation. This uses the recently added
Grafana container on the services image [1], which is completely
unconfigured other than having the two required plugins (PCP and Redis)
installed (as our tests can't access the internet).

So far, Cockpit is not involved in this at all. Over time, we can build
UI for this and replace the CLI commands.

Document in the feature page how to expose PCP metrics manually.

[1] https://github.com/cockpit-project/bots/pull/1990

-----
 - [x] Add grafana to services image: https://github.com/cockpit-project/bots/pull/1990
 - [x] Add redis to Fedora 34/testing images: https://github.com/cockpit-project/bots/pull/1991
 - [x] (optional) Add redis to Debian/Ubuntu images: https://github.com/cockpit-project/bots/pull/1994
 - [x] (optional) Add redis to RHEL/CentoS images: https://github.com/cockpit-project/bots/pull/1995
 - [x] (optional) Add redis to fedora-33 https://github.com/cockpit-project/bots/issues/1999
 - [x] Generate Grafana browser screenshots on failures

### DIFF
--- a/doc/guide/feature-pcp.xml
+++ b/doc/guide/feature-pcp.xml
@@ -2,19 +2,20 @@
 <!DOCTYPE chapter PUBLIC "-//OASIS//DTD DocBook XML V4.3//EN"
 	"http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd">
 <chapter id="feature-pcp">
-  <title>PCP</title>
+  <title>PCP Metrics</title>
 
   <para>If available, Cockpit uses the
     <ulink url="https://pcp.io/">Performance Co-Pilot</ulink> framework to
-    gather metrics data about the system. This data is used to display graphs.
+    gather metrics data about the system. This data is used to display the history
+    graphs on the "Performance Metrics" page.
     Cockpit can use the PCP logging feature to display archived data about the
     system from a different point in time. If PCP is not available, then Cockpit
     gathers the metrics data itself, but archival features are not available.</para>
 
   <para>Whether or not metrics are archived depends on whether the
     <code>pmlogger.service</code> and <code>pmcd.service</code> systemd units
-    are running or not.  The "Store Performance Data" button in Cockpit
-    will enable and start these services.</para>
+    are running or not.  The "Store metrics" switch in the "Configuration" card on
+    the "Overview" page will enable and start these services.</para>
 
   <para>To see similar metrics data from the command line, you can use tools like
     <ulink url="https://pcp.io/books/PCP_UAG/html/LE38515-PARENT.html#LE91266-PARENT"><code>pmstat</code></ulink>
--- a/doc/guide/feature-pcp.xml
+++ b/doc/guide/feature-pcp.xml
@@ -31,4 +31,19 @@ $ <command>pmstat</command>
 ...
 </programlisting>
 
+  <para>These metrics can also be exposed to other machines on a TCP port with
+  <ulink url="https://linux.die.net/man/1/pmproxy">pmproxy</ulink>
+  and <ulink url="https://redis.io/">Redis</ulink>:</para>
+
+<programlisting>
+systemctl enable --now redis pmproxy
+# if you use firewalld, open port 44322:
+firewall-cmd --permanent --add-service pmproxy
+firewall-cmd --reload
+</programlisting>
+
+<para>This allows you to gather and visualize PCP metrics from multiple machines with
+  <ulink url="https://grafana.com/">Grafana</ulink> and the
+  <ulink url="https://grafana-pcp.readthedocs.io">PCP Grafana plugin</ulink>.</para>
+
 </chapter>
--- a/doc/guide/feature-pcp.xml
+++ b/doc/guide/feature-pcp.xml
@@ -13,8 +13,9 @@
     gathers the metrics data itself, but archival features are not available.</para>
 
   <para>Whether or not metrics are archived depends on whether the
-    <code>pmlogger.service</code> and <code>pmcd.service</code> systemd units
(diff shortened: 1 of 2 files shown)


test/verify/check-metrics
```diff
@@ -731,5 +731,81 @@ class TestMultiCPU(MachineCase):
...
+        m.execute("until curl http://10.111.112.100:3000; do sleep 1; done")
+        # enable PCP plugin; like on Cog (Configuration) menu → Plugins → Performance Co-Pilot → Enable
+        mg.execute("curl --silent --show-error -u admin:foobar -d '' 'http://127.0.0.1:3000/api/plugins/performancecopilot-pcp-app/settings?enabled=true'")
+
+        # enable redis+pmproxy on supervised machine (should become an UI element)
+        if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+            redis = "redis-server"
+        else:
+            redis = "redis"
+        m.execute(f"systemctl enable --now {redis} pmproxy")
```
martinpitt, line 760: Let's just start it, not enable -- that works on Debian as well through `Alias=redis.service`. It just can't be enabled through an alias. We don't need this to reboot.

martinpitt:
The image rebuilds both landed, so I enabled the test everywhere except for f-33 (which was not rebuilt yet).

test/verify/check-metrics
martinpitt: Done.

martinpitt: Nack, a non-running service does not respect Alias= either, this is strictly an `enable` thing. So going back to the original solution.

martinpitt:
Dang, I had that working on debian-testing locally yesterday, but now the redis/pmproxy queries hang forever on all Debian/Ubuntus. I disabled the new test on these for now, I'll debug that separately. I robustified the test and avoided the polling/reload loop, which was rather ugly.

martinpitt:
On centos-8-stream the PCP redis source does respond, but never gets any data. That also needs debugging. Disabling there too for now, let's iterate separately on that.

martinpitt:
@KKoukiou : Want to have a first look whether this roughly matches what you did?

martinpitt:
Still [broken on debian-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15795-20210507-062818-e8444356-debian-stable/log.html#276):
```
# curl http://localhost:44322/series/ping
curl: (56) Recv failure: Connection reset by peer
```
No new log output at all in /var/log/pcp/. Since this works in Debian testing, I just skipped the test there and not bother.

I also added generation of Grafana browser screenshots on failure.

----

pmlogger.service fails on [ubuntu-2004](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15795-20210507-062819-e8444356-ubuntu-2004/log.html#276). This is because we try to enable/restart it before booting is fully done, and it is already running. Fixed.

Finally, f-33 refresh finished/landed, so enabling there as well.

martinpitt:
A [single failure](https://logs.cockpit-project.org/logs/pull-15795-20210507-075328-ef824faf-ubuntu-stable/log.html#276-1), and [screenshots work](https://logs.cockpit-project.org/logs/pull-15795-20210507-075328-ef824faf-ubuntu-stable/TestGrafanaClient-testBasic-services-127.0.0.2-2602-FAIL-grafana.png). Seems I was a bit too optimistic about timeouts on our CI, I bumped them.

test/verify/check-metrics
```diff
@@ -731,5 +731,86 @@ class TestMultiCPU(MachineCase):
...
+        mg.execute("curl --silent --show-error -u admin:foobar -d '' 'http://127.0.0.1:3000/api/plugins/performancecopilot-pcp-app/settings?enabled=true'")
+
+        # start redis and pmproxy on supervised machine (should become an UI element)
+        if m.image.startswith("debian") or m.image.startswith("ubuntu"):
+            m.execute("systemctl enable --now redis-server")
+            # pmproxy is runnning by default, re-start it to pick up redis
+            m.execute("systemctl restart pmproxy")
+        else:
+            # start PCP data collection (pmlogger), should become an UI element instead of just saying "not running"
+            m.execute("systemctl enable --now pmlogger")
```
martinpitt, line 762: I implemented that in PR #15803

doc/guide/feature-pcp.xml
```diff
@@ -2,19 +2,20 @@
...
+    graphs on the "Performance Metrics" page.
     Cockpit can use the PCP logging feature to display archived data about the
     system from a different point in time. If PCP is not available, then Cockpit
     gathers the metrics data itself, but archival features are not available.</para>
 
   <para>Whether or not metrics are archived depends on whether the
     <code>pmlogger.service</code> and <code>pmcd.service</code> systemd units
-    are running or not.  The "Store Performance Data" button in Cockpit
-    will enable and start these services.</para>
+    are running or not.  The "Store metrics" switch in the "Configuration" overview
```
KKoukiou: "Configuration card" in the overview page is maybe more straightforward? For me it's clear, I just wonder if people would search this card in the metrics page.

KKoukiou (commented):
Only one comment for discussion, applies to two places. Amazing :)

doc/guide/feature-pcp.xml
martinpitt: Good point, thanks! Updated docs accordingly.

PR #15802, opened 2021-05-07
services: when a transient service fails remove it from the list

### LINKED ISSUE #15801: services: transient units when they become inactive they should be removed from the list
Cockpit version: 243
Page: Services

When a transient systemd unit start there is a JobNew/JobRemoved signal which let's us be aware of it, and we add it to the list we show.
When the unit becomes inactive however, (or just fails to start after some delay) we update the 'State' but we don't remove it from the list.
<!---

Result can be something like this:
117149032-1baee800-adb7-11eb-98c1-0261a4cd9c44

These will be removed on page refresh, but these should not have been there in the first place.

KKoukiou:
The list of units we present is fetched at the initial page load. After
that we just update it by listening to signals.

In this case we were updating the properties of a transient unit that
failed and still showed it in the list. This however is wrong, since
when transient units stop they seize to exist.

Closes #15801


KKoukiou:
@marusak can you test if that fixes your issue? If yes I will try to add a test.

martinpitt:
@KKoukiou : Easy enough to test transient units:

    sudo systemd-run -u aaasuccess.service sleep 3
    sudo systemd-run -u aaafail.service sh -c 'sleep 3; false'

Normally failed transient units stay loaded, and thus get shown by `systemctl --failed`. That behaviour actually makes sense, IMHO. Only successful units should be cleaned up (and are currently not). Only an explicit `sudo systemctl reset-failed aaafail` drops them from "systemd's memory" (i.e. cause it to get unloaded).

You can use `systemd-run --collect` to force a cleanup even after failure.

marusak:
> Normally failed transient units stay loaded, and thus get shown by `systemctl --failed`.

UI looks like this (without patch in this PR, current master):
loogs

and `systemctl --failed` reports:
```
$ systemctl --failed
  UNIT                               LOAD   ACTIVE SUB    DESCRIPTION                
● NetworkManager-wait-online.service loaded failed failed Network Manager Wait Online

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.
1 loaded units listed.
```

marusak:
This PR indeed helps with my issue :)

KKoukiou:
> This PR indeed helps with my issue :)

Indeed but this will hide the failed units from the example that Martin mentioned above. Which are actually supposed to be listed. I literally can't think of any difference between the transient fail unit from the example Martin sent and the one the instance unit org.fedoraproject.Setroubleshootd that fails on your system. (it fails on mine as well since I upgraded to f34)

KKoukiou:
I think I found the difference:

```
CollectMode=

    Tweaks the "garbage collection" algorithm for this unit. Takes one of inactive or inactive-or-failed. If set to inactive the unit will be unloaded if it is in the inactive state and is not referenced by clients, jobs or other units — however it is not unloaded if it is in the failed state. In failed mode, failed units are not unloaded until the user invoked systemctl reset-failed on them to reset the failed state, or an equivalent command. This behaviour is altered if this option is set to inactive-or-failed: in this case the unit is unloaded even if the unit is in a failed state, and thus an explicitly resetting of the failed state is not necessary. Note that if this mode is used unit results (such as exit codes, exit signals, consumed resources, …) are flushed out immediately after the unit completed, except for what is stored in the logging subsystem. Defaults to inactive.
```

test/verify/check-system-services
```diff
@@ -789,6 +789,24 @@ Where=/fail
...
+
+        self.machine.execute("systemd-run --collect --unit test-autocollect@1.service sh -c 'sleep 3; false'")
+        self.machine.execute("systemd-run --unit test-manual-collect@1.service sh -c 'sleep 3; false'")
+
+        self.wait_service_present("test-autocollect@1.service")
+        self.wait_service_present("test-manual-collect@1.service")
+
+        self.wait_service_in_panel("test-manual-collect@1.service", "Failed to start")
+
+        self.wait_service_not_present("test-autocollect@1.service")
```
martinpitt, line 804: This is a race condition -- it would be more robust to replace the `sleep 3` with something like `until test -e /tmp/autocollect-continue; do sleep 0.5; done` and touch the file here. But tests seem happy, we can still do that when necessary.

martinpitt (approved):
Thanks! Feels a bit sad to hardcode so many property checks, but as we maintain that cache ourselves, we also have to mirror what systemd does.

PR #15803, opened 2021-05-07
metrics: Add button to enable pmlogger.service

martinpitt:
If there is no data and pmlogger.service is not running, give a more
specific message, and offer to enable/start it if the user is
privileged.

Split the test case into two cases: pmlogger is stopped (then UI for
restarting it), or it's failing or hanging (then old troubleshoot link
UI).

-----

Limited access mode, not running:
image

Superuser access, not running -- this is the common case:
image

Unit failed (permissions  don't matter):
image

And finally, if there's something "weird" going on, i. e. pmlogger is running, but it does not collect any data (hopefully most users will never see this):
image

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -38,6 +38,8 @@ import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import cockpit from 'cockpit';
 import * as machine_info from "../lib/machine-info.js";
 import * as packagekit from "packagekit.js";
+import * as service from "service";
+import { superuser } from "superuser";
 import { journal } from "journal";
 
 import { EmptyStatePanel } from "../lib/cockpit-components-empty-state.jsx";
@@ -881,6 +883,8 @@ class MetricsHistory extends React.Component {
             hours: [], // available hours for rendering in descending order
             loading: true, // show loading indicator
             metricsAvailable: true,
+            pmLoggerState: null,
+            pmLoggerOpPending: false,
             error: null,
             isDatepickerOpened: false,
             selectedDate: null,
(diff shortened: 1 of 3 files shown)


martinpitt:
[testNoDataEnable failure on debian-testing](https://logs.cockpit-project.org/logs/pull-15803-20210507-142309-d9ddecc7-debian-testing/log.html#278) and [debian-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15803-20210507-142309-d9ddecc7-debian-stable/log.html#278) (didn't reproduce locally yet). But the [journal](https://logs.cockpit-project.org/logs/pull-15803-20210507-142309-d9ddecc7-debian-testing/TestHistoryMetrics-testNoDataEnable-debian-testing-127.0.0.2-2301-FAIL.log.gz) shows that it failed due to the restart limit, presumably because all tests are run three times

martinpitt:
@garrett: This is fairly trivial, just giving you a chance to ponder the new button string. Thanks!

pkg/metrics/metrics.jsx
```diff
@@ -893,21 +896,31 @@ class MetricsHistory extends React.Component {
...
-                        this.today_midnight = new Date(current_hour).setHours(0, 0, 0, 0);
-                        this.setState({
-                            selectedDate: this.today_midnight,
-                        });
-                    })
-                    .catch(ex => this.setState({ error: ex.toString() }));
+        /* supervise pmlogger.service, to diagnose missing history */
+        this.pmlogger_service = service.proxy("pmlogger.service");
+        this.pmlogger_service.addEventListener("changed", () => {
+            if (this.pmlogger_service.state !== this.state.pmLoggerState)
```
marusak, line 902: Is this check necessary?

marusak (commented):
Nice! Just one question

pkg/metrics/metrics.jsx
martinpitt: Not strictly -- just because this page is rather expensive to render, I thought I'd rather go through some extra effort to not call setState() unnecessarily.

martinpitt: I mean, even calling render() itself is expensive (not the actual screen update, as shadow DOM would of course optimize that away)

martinpitt: This is also the reason why I'm not using our fancy new `useEvent(proxy, "changed")`. There are about 4 successive events with mostly redundant data (for the particular `state` property that we are interested in).

marusak: right, makes sense and cannot hurt :)

marusak (approved):

pkg/metrics/metrics.jsx
```diff
@@ -1075,12 +1088,38 @@ class MetricsHistory extends React.Component {
...
+                                    .then(() => this.initialLoadData())
+                                    .catch(ex => console.error("starting pmlogger.service failed:", JSON.stringify(ex)));
+                            if (!this.pmlogger_service.enabled)
+                                this.pmlogger_service.enable();
+                        } }>{_("Enable PCP metrics collector")}</Button>);
+                }
+                // when not superuser, don't show any action
+            } else {
+                if (this.pmlogger_service.state === 'failed')
+                    paragraph = _("pmlogger.service failed");
```
garrett, line 1111:
```suggestion
                    paragraph = _("pmlogger.service has failed");
```

pkg/metrics/metrics.jsx
```diff
@@ -1075,12 +1088,38 @@ class MetricsHistory extends React.Component {
...
+                else /* running, or initialization hangs */
+                    paragraph = _("pmlogger.service fails to collect data");
```
garrett, line 1113:
```suggestion
                    paragraph = _("pmlogger.service is failing to collect data");
```

garrett (changes requested):
Looks great!

Minor tense nitpicks:

- pmlogger.service has failed
- pmlogger.service is failing to collect data

I've added these as suggestions in this review. :wink:

pkg/metrics/metrics.jsx
```diff
@@ -1075,12 +1088,38 @@ class MetricsHistory extends React.Component {
...
+        if (!this.state.metricsAvailable) {
+            let action;
+            let paragraph;
+
+            if (this.pmlogger_service.state === 'stopped') {
+                paragraph = _("pmlogger.service is not running");
+
+                if (superuser.allowed) {
+                    action = (
+                        <Button variant="primary" onClick={() => {
```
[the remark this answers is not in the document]
martinpitt: Normally the service should switch from "stopped" (the button is shown only then) to "activating" pretty much instantly. "activating" can take long indeed. But I added a new `pmLoggerOpPending` bool state to be completely correct here. Thanks for spotting!

garrett (approved):
This works for me. Thanks for adding this feature!

marusak (approved):
Thanks :)

PR #15805, opened 2021-05-10
containers: Add explicit docker.io registry for unit-tests

martinpitt:
Most recent podman resolves the short "debian" name to quay.io. But
https://quay.io/repository/i386/debian is not a thing, thus building the
i386 container fails now with

    Error: error creating build container: Error initializing source docker://quay.io/i386/debian:stable:
    Error reading manifest stable in quay.io/i386/debian: unauthorized: access to the requested resource is not authorized

Explicitly pull from docker.io to fix that.

---

This fixes the [unit-tests-refresh workflow failures](https://github.com/cockpit-project/cockpit/actions/workflows/unit-tests-refresh.yml) that have started two weeks ago. Submitting this as an origin branch so that I can manually run the workflow against this branch.


martinpitt:
Heh, the "build unit tests container if it changed" step [fails here](https://github.com/cockpit-project/cockpit/pull/15805/checks?check_run_id=2542299931) for the same reason, unsurprisingly. Once the manual refresh worked, these should also get green again.

martinpitt:
Yay, all green! I linked my manual unit-tests-refresh workflow in the description, thus the checks plus that covered all changes here.

allisonkarlitskaya (approved):

PR #15806, opened 2021-05-10
test: Allow restart messages in TestSuperuser.testBasic

martinpitt:
The test calls Browser.relogin().

---

Fixes [this failure](https://logs.cockpit-project.org/logs/pull-15803-20210507-143509-d9ddecc7-fedora-33-firefox/log.html#198)


martinpitt:
Our own tests are green. Ignoring the unit-tests failure, it was mostly green first, and now it fails because of that very recent npm update.

PR #15807, opened 2021-05-10
Fix rpm validation errors

martinpitt:
Follow-up from https://github.com/cockpit-project/cockpit/pull/15784#pullrequestreview-651965021


marusak (approved):
Danke :)

PR #15809, opened 2021-05-10
storage: Use FormGroup helperText consistently in dialogs

mvollmer:
The two uses of FormGroup in Row looked inconsistent to me re helperText, but I didn't investigate this further

### DIFF
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -274,7 +274,7 @@ const Row = ({ field, values, errors, onChange }) => {
     } else {
         return (
             <FormGroup validated={validated}
-                       helperTextInvalid={error || explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
+                       helperTextInvalid={error} helperText={explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
                 { children }
             </FormGroup>
         );


KKoukiou (commented):
helperText is for information after the form field. Is explanation an information?

marusak (approved):
It will show `Saving a new passphrase requires unlocking the disk. Please provide a current disk passphrase.`
This makes sense as it would show this message only if you validation would fail. I believe it makes more sense to show this always.

marusak:
Wait or maybe the current wording does not fit there ideally. I think it should just say "Current disk passphrase to unlock the disk." or something like that?

mvollmer:
> helperText is for information after the form field. Is explanation an information?

Yes, "explanation" should always be displayed (unless there is an error, maybe).

For this PR, I was just looking at the code in the Row function, though:

```
    if (title || title == "") {
        let titleLabel = title;

        if (options.widest_title)
            titleLabel = (
                <>
                    <div className="widest-title">{options.widest_title}</div>
                    <div>{title}</div>
                </>
            );
        return (
            <FormGroup label={titleLabel} validated={validated}
                       helperTextInvalid={error} helperText={explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
                { children }
            </FormGroup>
        );
    } else if (!field.bare) {
        return (
            <FormGroup validated={validated}
                       helperTextInvalid={error || explanation} hasNoPaddingTop={field.hasNoPaddingTop}>
                { children }
            </FormGroup>
        );
```

Note how the first call to FormGroup is different from the second.  The first one made more sense to me.

KKoukiou:
> It will show `Saving a new passphrase requires unlocking the disk. Please provide a current disk passphrase.`
> This makes sense as it would show this message only if you validation would fail. I believe it makes more sense to show this always.

Only one of helperText and helperTextInvalid is shown at once and that depends on the validation state. (validated  property).

KKoukiou:
@marusak you talk about this dialog.
Screen Shot 2021-05-13 at 10 02 34


This helper text is always shown now, and I think it's correct. Do you disagree with that?

marusak:
> This helper text is always shown now, and I think it's correct. Do you disagree with that?

I do think it is correct to always show it. Just that the wording could be somewhat different for helper text. In general we don't say in helper texts things like "Please provide" for example. So I would change label from 'Disk passphrase' to 'Current disk passphrase' and then kept helper text as "Saving a new... the disk." if it is even necessary to have this helper text.

mvollmer:
> This helper text is always shown now, 

This change here only affects fields that don't have a label to the left of them, so this dialog should not have changed at all.

KKoukiou:
@mvollmer @marusak let's merge it then and adjust the 'Please provide' helper text, which is actually present in many dialogs, none of which are affected by this PR, in another PR.

Issue #14803, opened 2020-10-26
Option to turn off PCP


marusak:
This page replaced the "enable pcp" button on Overview. The empty state provides the "on", but there is no "off" button.

Perhaps we could have a settings section like this? Where it's shoved under the shortest card on the right.

![metrics-config](https://user-images.githubusercontent.com/10246/92609170-90786380-f2b6-11ea-99f7-579c9dd1311b.png)

Since the toggle is probably too visible and too easy to turn off, we could have something like the CPU mitigations @ https://localhost:9090/system/hwinfo — where there's a settings link that brings up a dialog with a toggle and a confirmation button. (This would only be for disabling, as it would presumably disable the service _and_ remove packages.)

Copied from https://github.com/martinpitt/performance-graphs/issues/18

martinpitt:
I don't quite like to squeeze it into the Network card:

 - This card row are the *current* metrics, and this switch refers to PCP (metrics history).
 - The network card does not have a fixed height, there can (and often are) many network interfaces.
 - There will be more PCP config for Grafana integration (PR #15810)

My current thought is to replace the "Install cockpit-pcp" and "Enable PCP metrics collector" buttons in the empty states:

![metrics-no-pcp-master](https://user-images.githubusercontent.com/200109/117935078-a5087200-b303-11eb-84ad-983db1154caa.png)

... with a new "PCP config" card:

![image](https://user-images.githubusercontent.com/200109/117934774-4cd17000-b303-11eb-9780-578f7e4bd144.png)

(This obviously has a lot of font/spacing/etc. issues, it's just a draft). This also simplifies the workflow: Enabling the switch would install cockpit-pcp (if necessary), and enable/start the service, there is no need any more to click *two* buttons.

Then the only case for an empty state would be if pmlogger.service failed (with the "Troubleshoot.." link), or runs but somehow does not collect data.

@garrett,  @marusak , WDYT?

marusak:
I like how the separate card looks like, but it feels to prominent for actions that you probably want to do once.

> we could certainly also place the card below the graph

You mean under history graphs? I think none would ever find it there.

An idea that comes to my mind would be to have kebab menu somewhere and it would open pop-up containing this settings. Surely not super discoverable, but as I said, this is very much a corner case for someone who will be looking for it. We don't have ideal place for this menu as we don't have any heading (as for example on firewall page) but maybe on the right in the breadcrumb panel?
[...]

PR #15810, opened 2021-05-10
Drop PCP configuration from Overview, add pcp config dialog to Metrics page

martinpitt:
nDrop installation/enablement of pmlogger/pmcd. This functionality moved
to the Performance Metrics page, where it is better motivated and
contextual.

 - [x] Add UI for re-disabling pmlogger to metrics page: #14803
 - [x] test that pmlogger.service gets enabled/disabled (not just stopped/started)
 - [x] `TestHistoryMetrics.testBasic` flake variant [one (4 times)](https://logs.cockpit-project.org/logs/pull-15810-20210517-134329-7d702193-centos-8-stream/log.html#277), [two](https://logs.cockpit-project.org/logs/pull-15810-20210517-134329-7d702193-ubuntu-2004/log.html#268), [three](https://logs.cockpit-project.org/logs/pull-15810-20210517-134329-7d702193-fedora-34/log.html#277-2)
 - [x] Dialog design updates from Garrett

## Metrics: PCP configuration dialog

The Metrics page header has a new "Metrics settings" button:

metrics-menu

This allows you to enable or disable the [Performance Co-Pilot](https://pcp.io/) (PCP) metrics collector:

metrics-settings

Related options will be added in the near future.

## Other screenshots for PR review
Off setting. I can't say I like the gray "Collect metrics" text, it makes the switch look disabled. But I suppose we should change that globally if we care

image

Mobile view:

image

The same button/dialog (except for primary instead of secondary) is also available from the empty state, for a better mental/logical connection. It subsumes the previous "Enable PCP metrics collector" button from that empty state, and the on/off switch on the Overview configuration card:

image

This provides a space where to add more PCP settings. The one that will follow immediately is pmproxy, for Grafana integration (PR #15812), which could look somewhat like this:

image

But we could also configure pmcd, pmie, etc. there in the future.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import moment from "moment";
 
 import {
@@ -27,13 +27,16 @@ import {
     Card, CardTitle, CardBody, Gallery,
     DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
     Flex, FlexItem,
+    Modal,
     Page, PageSection, PageSectionVariants,
     Progress, ProgressVariant,
     Select, SelectOption,
+    Switch,
+    Text, TextContent, TextVariants,
     Tooltip,
 } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody, TableGridBreakpoint, TableVariant, TableText, RowWrapper, cellWidth } from '@patternfly/react-table';
-import { ExclamationCircleIcon } from '@patternfly/react-icons';
+import { ExclamationCircleIcon, CogIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 import cockpit from 'cockpit';
 import * as machine_info from "../lib/machine-info.js";
@@ -41,11 +44,13 @@ import * as packagekit from "packagekit.js";
 import * as service from "service";
 import { superuser } from "superuser";
 import { journal } from "journal";
+import { useObject, useEvent } from "hooks.js";
 
 import { EmptyStatePanel } from "../lib/cockpit-components-empty-state.jsx";
 import { ListingTable } from "cockpit-components-table.jsx";
 import { JournalOutput } from "cockpit-components-logs-panel.jsx";
 import { install_dialog } from "cockpit-components-install-dialog.jsx";
+import { ModalError } from "cockpit-components-inline-notification.jsx";
 import "journal.css";
 
 const MSEC_PER_H = 3600000;
@@ -867,6 +872,92 @@ class MetricsHour extends React.Component {
     }
 }
 
+const persistentServiceState = proxy => ['running', 'stopped', 'failed', undefined].indexOf(proxy.state) >= 0;
+
+const PCPConfig = ({ buttonVariant }) => {
+    const [dialogVisible, setDialogVisible] = useState(false);
+    const [dialogError, setDialogError] = useState(null);
+    const [dialogLoggerValue, setDialogLoggerValue] = useState(false);
+    const [pending, setPending] = useState(false);
+    const s_pmlogger = useObject(() => service.proxy("pmlogger.service"), null, []);
+
+    debug("PCPConfig s_pmlogger.state", s_pmlogger.state);
+
+    useEvent(s_pmlogger, "changed");
+    useEvent(superuser, "changed");
+
+    if (!superuser.allowed)
+        return null;
+
+    const handleSave = () => {
+        setPending(true);
+        if (dialogLoggerValue) {
+            Promise.all([s_pmlogger.start(), s_pmlogger.enable()])
+                    .then(() => setDialogVisible(false))
+                    .catch(err => setDialogError(err.toString()))
+                    .finally(() => setPending(false));
+        } else {
+            Promise.all([s_pmlogger.stop(), s_pmlogger.disable()])
+                    .then(() => setDialogVisible(false))
+                    .catch(err => setDialogError(err.toString()))
+                    .finally(() => setPending(false));
+        }
+    };
+
+    return (
+        <>
+            <Button variant={buttonVariant} icon={<CogIcon />}
+                    isDisabled={ !persistentServiceState(s_pmlogger) }
+                    onClick={ () => {
+                        setDialogLoggerValue(s_pmlogger.state === 'running');
+                        setDialogVisible(true);
+                    } }>
+                { _("Metrics settings") }
+            </Button>
+
+            {dialogVisible &&
+            <Modal position="top" variant="small" isOpen
+                   id="pcp-settings-modal"
+                   onClose={ () => setDialogVisible(false) }
+                   title={ _("Metrics settings") }
+                   description={
+                       <div className="pcp-settings-modal-text">
+                           { _("Performance Co-Pilot collects and analyzes performance metrics from your system.") }
+
+                           <Button component="a" variant="link" href="https://cockpit-project.org/guide/latest/feature-pcp.html"
+                                isInline
+                                target="_blank" rel="noopener noreferrer"
+                                icon={<ExternalLinkAltIcon />}>
+                               { _("Read more...") }
+                           </Button>
+                       </div>
+                   }
+                   footer={<>
+                       { dialogError && <ModalError dialogError={ _("Failed to configure PCP") } dialogErrorDetail={dialogError} /> }
+
+                       <Button variant='primary' onClick={handleSave} isDisabled={pending}>
+                           { _("Save") }
+                       </Button>
+                       <Button variant='link' className='btn-cancel' onClick={ () => setDialogVisible(false) }>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+                   }>
+                <Switch id="switch-pmlogger"
+                        isChecked={dialogLoggerValue}
+                        label={
+                            <Flex spaceItems={{ modifier: 'spaceItemsXl' }}>
+                                <FlexItem>{ _("Collect metrics") }</FlexItem>
+                                <TextContent>
+                                    <Text component={TextVariants.small}>(pmlogger.service)</Text>
+                                </TextContent>
+                            </Flex>
+                        }
+                        onChange={enable => setDialogLoggerValue(enable)} />
+            </Modal>}
+        </>);
+};
+
 class MetricsHistory extends React.Component {
     constructor(props) {
         super(props);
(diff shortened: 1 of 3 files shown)


marusak:
https://github.com/cockpit-project/cockpit/issues/14803

martinpitt:
I pushed a first version of the config dialog. It is functionally correct now, but there is zero layout in the dialog, thus it is very broken:

image

(also needs tests adjustments)

martinpitt:
@garrett : I still want to refine the tests, but I'd appreciate a first look at the visuals. I made sceenshots of the new things, they ought to suffice for the first round? I have a gut feeling that we should explain things a little better perhaps?

garrett:
Dialog should use `sm` size. The offsite link needs more spacing and the text should not break (it should not wrap, unless absolutely necessary; the link should be considered its own widget). Labels should be to the right of switches; we don't need to do a form layout here with labels on the left (or above). I don't think the .service names are necessary in this dialog. We might want to include more information about the port and/or how to contact the service, however.

metrics-settings

martinpitt:
Thanks @garrett! Good points, I'll rework this. My [initial attempt](https://github.com/cockpit-project/cockpit/pull/15810#issuecomment-842243498) was indeed without a form, but that looked rather sad -- I mostly want this for the alignment, not for the "form"-i-ness. But I'll revisit this, in other places we e.g. use a DataList, not a form.

> I don't think the .service names are necessary in this dialog.

I thought it would be nice to give a hint that it's a systemd unit. Perhaps "Collect metrics (`pmlogger.service`)"?

garrett:
> I thought it would be nice to give a hint that it's a systemd unit. Perhaps "Collect metrics (pmlogger.service)"?

I actually had this as a mockup too. Here's what it looked like:

metrics-2

And had a stacked version as well, trying to get rid of the ()

metrics-3

Ultimately, I thought it didn't provide much value and dropped the service names altogether.

(At one point, I did even color them blue and was thinking of having them serve as links... but then going to another page within Cockpit via a link _on a modal_ is weird.)

martinpitt:
I fixed the testBasic regression. In exchange I broke `TestHistoryMetrics.testNoDataEnable` (I think), but that's a puzzle for tomorrow.

martinpitt:
I updated the dialog design as per @garrett's proposals above, I moved from a `<Form>` to `<Flex>`. I'm not 100% sure if that is the right choice, but I didn't find anything more appropriate in PF. (Aside from using a classic plain `<table>`). I updated the screenshots in the description.

I also updated/fixed the tests to what I had in mind, so this is ready for another review round now.

pkg/metrics/metrics.jsx
```diff
@@ -1213,14 +1213,17 @@ class MetricsHistory extends React.Component {
...
-    <Page groupProps={{ sticky: 'top' }}
-          isBreadcrumbGrouped
-          breadcrumb={
-              <Breadcrumb>
-                  <BreadcrumbItem onClick={() => cockpit.jump("/system")} to="#">{_("Overview")}</BreadcrumbItem>
-                  <BreadcrumbItem isActive>{_("Performance Metrics")}</BreadcrumbItem>
-              </Breadcrumb>
-          }>
+    <Page>
+        <PageSection variant={PageSectionVariants.light}>
```
KKoukiou: Breadcrumb should be passed like this https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/details.jsx#L129
Or you can see also here https://www.patternfly.org/v4/components/page/react-demos but there is so much other junk.

KKoukiou (changes requested):

pkg/metrics/metrics.jsx
```diff
@@ -867,6 +872,91 @@ class MetricsHour extends React.Component {
...
+                       <Button variant='link' className='btn-cancel' onClick={ () => setDialogVisible(false) }>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+                   }>
+
+                <Text component={TextVariants.p} className="pcp-settings-modal-text">
+                    { _("Performance Co-Pilot collects and analyzes performance metrics from your system.") }
+
+                    <a href="https://pcp.io/" target="_blank" rel="noopener noreferrer">
```
KKoukiou, line 938: You can also use PF4's `<Button component='a' variant='link icon=/...>` I don't think it will change the style though.

KKoukiou (commented):
I want to double check the the custom CSS introduced can't be avoided. Will check that locally shortly, pls keep the PR open.

pkg/metrics/metrics.jsx
```diff
@@ -867,6 +872,91 @@ class MetricsHour extends React.Component {
...
+                        <ExternalLinkAltIcon />&nbsp;{ _("Read more...") }
+                    </a>
+                </Text>
+
+                <Flex direction={{ default: 'column' }}>
```
KKoukiou, line 943: Using the flex here results in misalignments.
Screen Shot 2021-05-19 at 12 44 51
 

I will try to find a better apporach.

KKoukiou (changes requested):

pkg/metrics/metrics.jsx
KKoukiou: In addition the Switch should have a label or an aria label to be accessible. Clicking on the label should also enable/disable the switch.

pkg/metrics/metrics.jsx
martinpitt: That's exactly how it was before, [this commit](https://github.com/cockpit-project/cockpit/pull/15810/commits/c56aa0f42d945256f4b4aeac2ce7dd7b750af258) moves away from it. See the [discussion in #14803](https://github.com/cockpit-project/cockpit/issues/14803#issuecomment-839703508) for details. Reason 1: This makes the header bar exactly as high as the Overview, so it's more consistent; Reason 2: The Page.breadcrump attribute has no way to place an additional action into the header bar.

pkg/metrics/metrics.jsx
martinpitt: I could also use the "builtin" label of a switch, which will most probably fix the alignment, and certainly fix the "linking" of the two. The main downside is that the label becomes light grey when the button is off, and thus the widget looks disabled when it's not. But that's a general problem of the switches, so let's rather fix that with CSS if we care. Then I just need to have a way to add the (unitname) bit.

I'll make a version with that, and re-post.

KKoukiou: Right, that should be fixed globally is patternfly overrides file if we decide it's wrong.

pkg/metrics/metrics.jsx
KKoukiou: I see, makes sense as a change then.

pkg/metrics/metrics.jsx
martinpitt: Thanks! I applied your fixup.

pkg/metrics/metrics.jsx
martinpitt: Thanks for your fixup! applied.

martinpitt:
Thanks @KKoukiou for your help! PR and screenshots updated.

pkg/metrics/metrics.jsx
KKoukiou: I see two drawbacks with this change still:
* breadcrumb is not longer sticky
* we dont use the default paddings, indeed overview and kdump use the same approach, but all other pages not. 

Definitely not blockers.

KKoukiou (approved):

KKoukiou (commented):
Now that I rethink about it the undoing of the sticky header, is a small regression. What do you think?

pkg/metrics/metrics.jsx
KKoukiou: https://paste.centos.org/view/6c0c0993
This change will keep the header sticky.

martinpitt: Nice, thank you! Admittedly I didn't even notice the non-stickyness. Applied.

KKoukiou (dismissed):

garrett:
FWIW, here's a screenshot:

Screenshot from 2021-05-19 09-50-52

I only see one option; I guess that's intentional for now?

The read more link is super odd. It links to the PCP website? Is that a placeholder? I expected to read more about PCP works in Cockpit, not just go to the PCP website.

martinpitt:
@garrett : Yes, it's intentional. See the description, the "pmproxy" stuff is a separate PR. That was just a preview how the dialog would extend to more options in the future.

The link goes to https://pcp.io/ . It's similar to the other external links that we have, like https://access.redhat.com/security/vulnerabilities/L1TF on the hwinfo side. I thought it'd be nice to link to the actual documentation of the technology that provides all these metrics (the learning Linux/extend to other machines aspect). Neither pmlogger nor pmproxy/Grafana are cockpit specific, it just helps you to set these up. (In the same way as you'd format a drive through udisks on the Storage page, or start a libvirt VM on Machines page).

An alternative link target would be https://cockpit-project.org/guide/latest/feature-pcp.html , which would be one level of indirection -- but maybe that's preferable?

garrett:
"Read more" states the page will be more about the text it's talking about. The PCP website isn't that. Our page is.

So we could link our site with more information, adjust the text to make it clear it's just linking to PCP, change the link to be the "Performance Co-Pilot (with the external link icon), or drop the link.

martinpitt:
> "Read more" states the page will be more about the text it's talking about. 

Interesting, that's exactly what I had in mind -- the sentence before explains what PCP is, but that dialog is way too small to have all the details about it.

> So we could link our site with more information

OK, I'll link to our feature page then, that seems like the best compromise. I really want some doc link in there.

Thanks!

martinpitt:
> I'll link to our feature page then, that seems like the best compromise. I really want some doc link in there.

Done. No visual difference, just the link target changed. @garrett, PTAL?

martinpitt:
Pushed a rebase to get past the new "protection checks".

pkg/metrics/metrics.jsx
```diff
@@ -867,6 +872,92 @@ class MetricsHour extends React.Component {
...
+                                icon={<ExternalLinkAltIcon />}>
+                               { _("Read more...") }
+                           </Button>
+                       </div>
+                   }
+                   footer={<>
+                       { dialogError && <ModalError dialogError={ _("Failed to configure PCP") } dialogErrorDetail={dialogError} /> }
+
+                       <Button variant='primary' onClick={handleApply} isDisabled={pending}>
+                           { _("Apply") }
```
garrett, line 939:
```suggestion
                           { _("Save") }
```

This should be "Save", not "Apply".

garrett (changes requested):
Looks good!

The action button should say "Save" though. Aside from that, it's ready.

garrett (approved):
Cool! I see you've changed it to save. :+1:

PR #15812, opened 2021-05-10
metrics: Add switch for setting up pmproxy

martinpitt:
This is the next (and a rather major) step towards setting up a machine in Grafana.
    
 - [x] PR #15810 
 - [x] PR #15877

TODO:
 - [x] Use in TestGrafana
 - [x] don't disable redis with pmproxy; See https://github.com/performancecopilot/pcp/issues/1296
 - [x] consider a redis requires.d/ symlink instead of statically enabling
 - [x] pmproxy needs pmlogger, so disable button if pmlogger is off
 - [x] add link to firewall page if firewalld is running
 - [x] release note
 - [x] [fix testPmProxySettings flake](https://logs.cockpit-project.org/logs/pull-15812-20210601-120726-e08ecc5d-rhel-8-5/log.html#284) on rhel-8-5
 - [x] Add zone selector to alert
 - [x] Test pmproxy firewalld error
 - [x] Change TestGrafana to use the alert selector instead of the firewall page
 - [x] Add box shadow to alert
 - [x] Reduce alert/pagesection padding 
 - [ ] Fix [test flake](https://github.com/cockpit-project/cockpit/pull/15812#issuecomment-849514020)
    
## Metrics: Grafana client setup

The "Metrics settings" button introduced in [Cockpit 245](https://cockpit-project.org/blog/cockpit-245.html) has a new switch to set up exporting metrics to the network through the [PCP pmproxy service](https://linux.die.net/man/1/pmproxy) and the [Redis database](https://grafana-pcp.readthedocs.io/en/latest/datasources/redis.html):

Screen Shot 2021-05-32 at 11 38 16

If the machine uses firewalld, you can allow pmproxy in a particular zone afterwards:

fw-sel

After that, a remote [Grafana](https://grafana.com/) instance can read the machine's metrics through the [PCP Grafana plugin](https://grafana-pcp.readthedocs.io/), as described on the [PCP metrics feature page](https://cockpit-project.org/guide/latest/feature-pcp.html).

### DIFF
--- a/pkg/lib/cockpit-components-firewalld-request.scss
+++ b/pkg/lib/cockpit-components-firewalld-request.scss
@@ -1,0 +1,11 @@
+@import "@patternfly/patternfly/utilities/BoxShadow/box-shadow.css";
+
+.pf-c-page__main-section.ct-no-bottom-padding {
+    --pf-c-page__main-section--PaddingBottom: 0;
+}
+
+// <Toolbar> embedded into an <Alert>
+.pf-c-toolbar.ct-alert-toolbar {
+    --pf-c-toolbar--BackgroundColor: transparent;
+    --pf-c-toolbar--PaddingBottom: 0;
+}
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -938,7 +938,7 @@ const PCPConfig = ({ buttonVariant }) => {
                    footer={<>
                        { dialogError && <ModalError dialogError={ _("Failed to configure PCP") } dialogErrorDetail={dialogError} /> }
 
-                       <Button variant='primary' onClick={handleSave} isDisabled={pending}>
+                       <Button variant='primary' onClick={handleSave} isDisabled={pending} isLoading={pending}>
                            { _("Save") }
                        </Button>
                        <Button variant='link' className='btn-cancel' onClick={ () => setDialogVisible(false) }>
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -51,6 +51,7 @@ import { ListingTable } from "cockpit-components-table.jsx";
 import { JournalOutput } from "cockpit-components-logs-panel.jsx";
 import { install_dialog } from "cockpit-components-install-dialog.jsx";
 import { ModalError } from "cockpit-components-inline-notification.jsx";
+import { FirewalldRequest } from "cockpit-components-firewalld-request.jsx";
 import "journal.css";
 
 const MSEC_PER_H = 3600000;
@@ -875,11 +876,12 @@ class MetricsHour extends React.Component {
 const persistentServiceState = proxy => ['running', 'stopped', 'failed', undefined].indexOf(proxy.state) >= 0;
 const validServiceState = proxy => ['running', 'stopped'].indexOf(proxy.state) >= 0;
 
-const PCPConfig = ({ buttonVariant }) => {
+const PCPConfig = ({ buttonVariant, firewalldRequest }) => {
     const [dialogVisible, setDialogVisible] = useState(false);
     const [dialogError, setDialogError] = useState(null);
     const [dialogLoggerValue, setDialogLoggerValue] = useState(false);
     const [dialogProxyValue, setDialogProxyValue] = useState(null);
+    const [dialogInitialProxyValue, setDialogInitialProxyValue] = useState(null);
     const [pending, setPending] = useState(false);
 
     const s_pmlogger = useObject(() => service.proxy("pmlogger.service"), null, []);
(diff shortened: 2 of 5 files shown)


martinpitt:
Much better! Just [one flake](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15812-20210527-083118-ac689531-fedora-34/log.html#284), all other metrics tests are green now. In that failure, the pmproxy switch is disabled, so pmproxy.service or redis.service were not in a stable "running" or "stopped" state. The test already waits for pmproxy to finish starting, and the dialog even auto-updates to changes in the services while its open. Thus, the state must be some stable one other than "stopped" or "running".

The [journal](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15812-20210527-083118-ac689531-fedora-34/TestHistoryMetrics-testPmProxySettings-fedora-34-127.0.0.2-2301-FAIL.log.gz) shows that redis is stopped and pmproxy is started, so no hint from there.

This does not reproduce locally, so in the next round I'll enable debug messages and try to catch it.

martinpitt:
There is still a pesky test flake on RHEL 8.5 that I need to track down, but other than that I'm now happy with this initial functionality. I'd like to do the on-demand installation of redis in a follow-up. @KKoukiou , @garrett , any comments about the design? The screenshots in the description are up to date.

garrett:
It's weird to have an alert in Cockpit to go somewhere else and do ... something ... rather than fixing it immediately.

Could we have a zone selector with an explicit service enable instead? And have a fallback to visit the firewall? I would imagine it would look like actions separated with a vertical line like a standard toolbar https://www.patternfly.org/v4/components/toolbar/#items (except this isn't really a toolbar; just showing that it's in PF's design language) in the alert's action area.

Drop the close button (as dismissing it would cause a silent error). And don't show the alert if the firewall is not running or if the rule exists somewhere in the firewall already.

metrics-pmproxy-notification-2021-06-01-1705(2)

martinpitt:
THanks @garrett for reviewing!

> Could we have a zone selector with an explicit service enable instead?

That's what I started to work on yesterday, and then discarded it again, although that was inside the dialog itself. It's plausible that it needs to go in more than one zone, so it would need to be a checkbox or multi-select. My biggest gripe though is what this field should show if pmproxy was already enabled in some zone. I.e. this would quickly escalate to what the full firewall page was showing.

However, with this being in an alert, and only shown when enabling pmproxy (not for showing the current state in the dialog), this would be more realistic for the simple cases, yes.

> And don't show the alert if the firewall is not running

That already happens

> or if the rule exists somewhere in the firewall already.

That I'm not so sure about, and it is quite expensive to check, but I can do that.

Would you say that either of these are a "first version" requirement?

garrett:
>>    And don't show the alert if the firewall is not running
>
> That already happens

Great! That's what I suspected, but I was multi-tasking and didn't find the time to try it out.

>>     or if the rule exists somewhere in the firewall already.
>
> That I'm not so sure about, and it is quite expensive to check, but I can do that.

It would be a bit weird to enable the setting and be asked to add it to the firewall when it already exists. :neutral_face: 

> Would you say that either of these are a "first version" requirement?

I would consider Cockpit guiding someone on what to do a first version requirement. If it's in another, follow-up PR _that lands in the same release_ is splitting hairs. It should definitely land in the same release though.

If it's handled in a follow-up, then the absolute minimum would be to:

"Add the **pmproxy** service in the firewall to share metrics."
"Visit Firewall"

But I strongly feel that Cockpit should enable it in the same UI, else someone could easily forget what to do or might not find out how to add it due to the _huge_ context switch.

martinpitt:
Ack, I'll work on that tomorrow then (and probably until next week)

martinpitt:
How's this? This is an actual toolbar inside now. It just needs one CSS tweak with "make the background color transparent" (otherwise it appears as white), but otherwise seems to be quite close to your proposal:

image

It's a dummy still, but it does read out available zone names from firewalld, and skips the alert if pmproxy is enabled in any zone already. Wiring up the rest should be fairly standard plumbing.

garrett:
> make the background color transparent

For the dropdown? It's OK as white, but we can look into that. If it were transparent, we'd want to darken the border too.

It also seems like there's a little too much padding under the toolbar; I think it's a combination of the widgets in a way that PF hasn't tried.

But otherwise, it looks good, aside from text tweaks.

martinpitt:
@garrett: No, for the toolbar itself. The CSS is already applied. The dropdown is fine, IMHO. Agreed about the "too much padding", I'll tweak this when I have stuff wired up and the bots can chew on the tests.

martinpitt:
Running out of time today, so pushing what I have. I added some TODOs to the description for my Monday self. Locally I ran into a crash of the firewall page when stopping `firewalld.service`; I rearranged the test a bit to avoid that and made a todo item on my personal list for that. Let's see what the tests say.

martinpitt:
All the [TestService](http://artifacts.dev.testing-farm.io/1554c14e-8b81-4d26-97bc-f7e02eccb9f1/) checks now suddenly fail in packit. This PR doesn't touch the systemd page at all, and the metrics checks don't even run in packit, so at first sight this smells unrelated. OTOH the other recent PRs are fine. I mostly want to record the test URL for now. Most probably I just need to rebase to pick up PR #15818.

The more concerning issue is that this pesky [missing "started" systemd dbus proxy issue is still missing occasionaly](https://logs.cockpit-project.org/logs/pull-15812-20210531-092100-aeb4f086-rhel-8-5/log.html#284-2). I pushed an extra-verbose debugging version.

pkg/lib/cockpit-components-firewalld-request.scss
```diff
@@ -0,0 +1,11 @@
+@import "@patternfly/patternfly/utilities/BoxShadow/box-shadow.css";
+
+.pf-c-page__main-section.ct-no-bottom-margin {
```
KKoukiou, line 3: This removes padding - not margin - which makes the class name misleading. No big deal however.

pkg/metrics/metrics.jsx
```diff
@@ -873,46 +874,114 @@ class MetricsHour extends React.Component {
...
+        redis_name = "redis.service";
+    }
+
+    debug("PCPConfig s_pmlogger.state", s_pmlogger.state);
+    debug("PCPConfig s_pmproxy state", s_pmproxy.state, "redis exists", s_redis.exists, "state", s_redis.state, "redis-server exists", s_redis_server.exists, "state", s_redis_server.state);
 
     if (!superuser.allowed)
         return null;
 
     const handleSave = () => {
```
KKoukiou, line 914: Ideal would be to use isLoading Button property while the dialog is processing the save. And of course disable it. Because it takes usually a few seconds till the dialog is closed.

KKoukiou (approved):
Two tiny nitpicks - this is amazing work! We can fix these two small issues in followup.

pkg/metrics/metrics.jsx
martinpitt: Disabling should already happen through `isDisabled={pending}`. But good point about the `isLoading`, I'll try that.

martinpitt: Done, in a separate commit as that dialog already exist in master. Thanks for the suggestion!

pkg/lib/cockpit-components-firewalld-request.scss
martinpitt: Thanks for spotting! Fixed while I was re-pushing anyway.

KKoukiou (approved):

PR #15813, opened 2021-05-11
lib: Only offer configuring specific NTP servers with enabled systemd-timesyncd

martinpitt:
Checking for the service unit existence is not sufficient: On Fedora,
systemd-timesyncd.service does exist, but is disabled and never actually
used, as chronyd.service is running. Thus setting any custom NTP servers
has no effect, and is thus highly misleading.

Run `TestSystemInfo.testTimeServers` on all OSes, and ensure that the
"specific NTP servers" option is disabled when timesyncd is not in use.

----

I noticed that while investigating our [long-standing testTimeServers unstable test](https://logs.cockpit-project.org/logs/pull-15810-20210510-174243-2cab9769-fedora-34/log.html#280), and thought "WTH? That doesn't make any sense"..

### DIFF
--- a/pkg/lib/serverTime.js
+++ b/pkg/lib/serverTime.js
@@ -265,7 +265,7 @@ export function ServerTime() {
          * - systemd-timedated is answering for
          *   org.freedesktop.timedate1 as opposed to, say, timedatex.
          *
-         * - systemd-timesyncd is actually available.
+         * - systemd-timesyncd is enabled (false if chrony is being used)
          *
          * The better alternative would be to have an API in
          * o.fd.timedate1 for managing the list of NTP server
@@ -282,8 +282,8 @@ export function ServerTime() {
             return Promise.resolve(result);
         }
 
-        if (!timesyncd_service.exists) {
-            console.log("systemd-timesyncd not available, ntp server configuration not supported");
+        if (!timesyncd_service.enabled) {
+            console.log("systemd-timesyncd not enabled, ntp server configuration not supported");
             return Promise.resolve(result);
         }
 
--- a/pkg/lib/serverTime.js
+++ b/pkg/lib/serverTime.js
@@ -640,30 +640,16 @@ function change_systime_dialog(server_time, timezone) {
                                                                     state.manual_hours,
                                                                     state.manual_minutes));
                     } else {
-                    /* HACK - https://bugzilla.redhat.com/show_bug.cgi?id=1272085
-                     *
-                     * Switch off NTP, bump the clock by one microsecond to
-                     * clear the NTPSynchronized status, write the config
-                     * file, and switch NTP back on.
-                     *
-                     */
+                        // Switch off NTP, write the config file, and switch NTP back on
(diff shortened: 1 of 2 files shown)


marusak (approved):
Thanks!

martinpitt:
Still [fails on ubuntu-2004](https://logs.cockpit-project.org/logs/pull-15813-20210511-063109-b7c51f06-ubuntu-2004/log.html#281)

martinpitt:
Halleluja! testTimeServers succeeded everywhere three times on first try. The only red was [testMotd on debian-testing](https://logs.cockpit-project.org/logs/pull-15813-20210512-044851-1ea26ef8-debian-testing/log.html#279-1), which did not succeed three times in a row (affected test), but that is a matter for another debugathon. Retried that one.

marusak:
[Ubuntu](https://logs.cockpit-project.org/logs/pull-15813-20210512-053610-1ea26ef8-ubuntu-2004/log.html#282-1) still fails. Is this not the same issue?

martinpitt:
*sigh* this is impossible to  please. Honestly I'm almost at the point of saying "screw it, we toss out custom NTP servers. We have been doing it wrong for a long time, we are *still* doing it wrong even with this PR (e.g. we only parse that cockpit-specific include file, not the whole config), and it doesn't conceptually work on Fedora, RHEL, and CentOS anyway. The races and bugs in systemd make this almost impossible to get right -- there just simply is no timedatectl API which works for chrony and timesyncd.

I still can't see a sensible and robust definition of "are custom NTP servers supported".

@marusak, @mvollmer, do you have an opinion about this?

martinpitt:
@keszybz just explained in https://bugzilla.redhat.com/show_bug.cgi?id=1272085 how the "synchronized: yes" is *not* a bug. So I dropped that hack.

I also added a new commit with a different approach to avoiding that race condition. It's ugly to do that in the test instead of the code, but as I wrote above this whole approach is just too broken to be 100% race free :cry:

martinpitt:
That seems to have helped -- [testTimeServers on ubuntu-2004](https://logs.cockpit-project.org/logs/pull-15813-20210512-084211-57a51077-ubuntu-2004/log.html) succeeded three times, just that `testMotd` flake hit again. Retrying.

thrix:
@martinpitt F34 should be back, but I had to roll back to released image 1.2 :( nightly is like dead, does not boot at all ...

martinpitt:
[failed again](https://logs.cockpit-project.org/logs/pull-15813-20210512-125925-57a51077-ubuntu-2004/log.html#282) :hammer_and_wrench: :exploding_head:

martinpitt:
testTimeServers succeeded three times everywhehre! There's just that [unrelated testMotd failure on d-stable](https://logs.cockpit-project.org/logs/pull-15813-20210517-061115-c3775bdf-debian-stable/log.html), which I'll retry.

marusak (approved):
Phew, hopefully that solved that :crossed_fingers: 

Thanks! :)

PR #15814, opened 2021-05-11
packit: Fix mock-build-env setup

martinpitt:
Pass `$PKG_CONFIG_PATH` through the environment instead of as a
./configure argument, so that the mock pkg-config files are actually
being used.

----

This fixes the [packit srpm build failures](https://dashboard.packit.dev/results/srpm-builds/23021) that have happened since yesterday (a sandcastle container refresh happened recently). I'm not actually sure *what exactly* changed, but `./configure --help` and the error message both clearly say that this is an environment variable.


martinpitt:
Meh,  [still failed](https://dashboard.packit.dev/results/srpm-builds/23073) -- I could perfectly reproduce this previously in `podman run -ti --rm --memory 768MB -v $PWD:/src -w /src quay.io/packit/sandcastle:prod bash`, but not any more now..

marusak (approved):
Thanks for looking at it!

PR #15815, opened 2021-05-11
storage: More robust testing of content listing tabs

mvollmer:
The tests used to identify tabs by their position in the DOM, but that
isn't reliable since the ListingPanel instantiates tabs dynamically.
Thus, while the tests might look for the third .ct-listing-panel-body
for the third tab, only two tabs might have been instantiated and the
.ct-listing-panel-body for the third tab is actually the second one in
the DOM.

To make this reliable, let's put the actual tab index into the DOM as
an attribute and have the tests look for that.

We never actually got bitten by this, but we will in the future when
we have more tabs than two.


KKoukiou (approved):

mvollmer:
There is one more case that needs fixing.

KKoukiou (approved):

PR #15816, opened 2021-05-11
containers: Speed up dist unit-tests scenario

martinpitt:
Skip the full `make` build and xz extreme compression in the "start
dist" scenario.

### DIFF
--- a/containers/unit-tests/build.sh
+++ b/containers/unit-tests/build.sh
@@ -1,4 +1,4 @@
 #!/bin/sh -eux
 
 ./autogen.sh --prefix=/usr --enable-strict
-make
+[ "${TEST_SCENARIO:-}" = "dist" ] || make
--- a/containers/unit-tests/scenario-dist.sh
+++ b/containers/unit-tests/scenario-dist.sh
@@ -1,6 +1,6 @@
 #!/bin/sh -eux
 
-make NO_DIST_CACHE=1 dist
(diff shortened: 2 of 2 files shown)


.github/workflows/build-dist.yml
```diff
@@ -12,16 +12,19 @@ jobs:
...
           fetch-depth: 0
 
-      - name: Run unit-tests container to build dist
-        run: |
-          sudo containers/unit-tests/start dist
+      - name: Build dist
+        run:  |
+          # container runs as different user
+          chmod -R a+w source
+          sudo podman run -i -v ./source:/source -w /source ghcr.io/cockpit-project/unit-tests test/make_dist.py
```
allisonkarlitskaya, line 23: Why not support that officially in the `start` script?

allisonkarlitskaya (changes requested):

.github/workflows/build-dist.yml
allisonkarlitskaya: Testing "`make dist` without `make`" is semi-legit as a scenario...

martinpitt:
@allisonkarlitskaya : Reworked. build-dist  run here took 8:32 minutes, which is significantly faster than the [previous recent runs](https://github.com/cockpit-project/cockpit/actions/workflows/build-dist.yml) (9:30 to 11:30 minutes).

allisonkarlitskaya (approved):
That's a lot better!  Thanks!!

PR #15818, opened 2021-05-11
SPUR: services: Improve UI in the main services page

### LINKED ISSUE #15796: services: Translate all states and make sure they are capitalized
Came up in the UI review.

### LINKED ISSUE #15797: services: Port main services list to Table component
Came up in the UI review.

* keep name and description in one column which will wrap in smaller screen
* disable PF4 defaut mobile mode
*  ensure that table columns with long texts are wrapped or truncated
* Intergrate the filtering

### LINKED ISSUE #15798: services: Units without description end up showing the unit name in both ‘columns’
Hide these in this case.

Came up in the UI review.

KKoukiou:
Fixes #15797
Fixes #15798
Fixes #15796

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -41,6 +41,7 @@ import './cockpit-components-table.scss';
  * See https://www.patternfly.org/v4/components/table/
  * Properties (all optional unless specified otherwise):
  * - caption
+ * - id: optional identifier
  * - className: additional classes added to the Table
  * - actions: additional listing-wide actions (displayed next to the list's title)
  * - columns: { title: string, header: boolean, sortable: boolean }[] or string[]
@@ -203,6 +204,9 @@ export class ListingTable extends React.Component {
         if (this.props.rows.length == 0)
             tableProps.className += ' ct-table-empty';
 
+        if (this.props.id)
+            tableProps.id = this.props.id;
+
         if (this.props.variant)
             tableProps.variant = this.props.variant;
 
@@ -312,6 +316,7 @@ ListingTable.defaultProps = {
     showHeader: true,
 };
 ListingTable.propTypes = {
+    id: PropTypes.string,
     caption: PropTypes.string,
     emptyCaption: PropTypes.node,
     emptyCaptionDetail: PropTypes.node,
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -1,4 +1,6 @@
 /*** PF4 overrides ***/
+// Pull in variables (for breakpoints)
+@import "_global-variables.scss";
 
 /* WORKAROUND: Navigation problems with Tertiary Nav widget on mobile */
 /* See: https://github.com/patternfly/patternfly-design/issues/840 */
@@ -40,7 +42,7 @@ svg {
     }
 }
 
-// The default gap between the rows in horizonal lists is too large
+// The default gap between the rows in horizontal lists is too large
 .pf-c-description-list.pf-m-horizontal {
     --pf-c-description-list--RowGap: 1rem;
 }
@@ -130,4 +132,33 @@ svg {
 .pf-c-alert__title {
     font-size: inherit;
     font-family: inherit;
+}
+
+.pf-c-toolbar {
+    // There should be an opt-in to make toolbars sticky.
+    // For now, we'll use a ct-sticky class while we wait for an official solution.
+    // https://github.com/patternfly/patternfly/issues/4113
+    &.ct-sticky {
+        position: sticky;
+        top: 0;
+        z-index: 2;
+    }
+
+    // Make summary content use the same vertical space as the filter toggle,
+    // when possible.
+    // https://github.com/patternfly/patternfly/issues/4115
+    &.ct-compact {
+        @media screen and (max-width: $pf-global--breakpoint--lg - 1) {
+            display: flex;
+            flex-wrap: wrap;
+
+            > .pf-c-toolbar__content:first-child {
+                flex: auto;
+            }
+
+            .pf-c-toolbar__content-section {
+                width: auto;
+            }
+        }
+    }
 }
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -151,6 +151,50 @@ class ServicesPage extends React.Component {
             tabErrors: {},
             isFullyLoaded: false,
         };
+
+        // Possible LoadState values: stub, loaded, not-found, bad-setting, error, merged, masked
+        // See: typedef enum UnitLoadStateState https://github.com/systemd/systemd/blob/main/src/basic/unit-def.h
+        this.loadState = {
+            stub: _("Stub"),
+            loaded: "",
+            "not-found": _("Not found"),
+            "bad-setting": _("Bad setting"),
+            error: _("Error"),
+            merged: _("Merged"),
+            masked: "", // We present the masked from the unitFileState
+        };
+
+        // Possible ActiveState values: active, reloading, inactive, failed, activating, deactivating, maintainance
+        // See: typedef enum UnitActiveState https://github.com/systemd/systemd/blob/main/src/basic/unit-def.h
+        this.activeState = {
+            active: _("Running"),
+            reloading: _("Reloading"),
+            inactive: _("Not running"),
+            failed: _("Failed to start"),
+            activating: _("Running"),
+            deactivating: _("Not running"),
+            maintainance: _("Maintanance"),
+        };
+
+        // Possible UnitFileState values: enabled, enabled-runtime, linked, linked-runtime, alias, masked, masked-runtime, static, disabled, invalid, indirect, generated, transient, bad
+        // See: typedef enum UnitFileState https://github.com/systemd/systemd/blob/main/src/basic/unit-file.h
+        this.unitFileState = {
+            enabled: _("Enabled"),
+            "enabled-runtime": _("Enabled"),
+            disabled: _("Disabled"),
+            linked: _("Linked"),
+            "linked-runtime": _("Linked"),
+            alias: _("Alias"),
+            masked: _("Masked"),
+            "masked-runtime": _("Masked"),
+            static: _("Static"),
+            invalid: _("Invalid"),
+            indirect: _("Indirect"),
+            generated: _("Generated"),
+            transient: _("Transient"),
+            bad: _("Bad"),
+        };
+
         /* Functions for controlling the toolbar's components */
         this.onClearAllFilters = this.onClearAllFilters.bind(this);
         this.onFileStateSelect = this.onFileStateSelect.bind(this);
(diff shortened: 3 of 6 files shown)


KKoukiou:
@marusak wanna give it a first look and give me some feedback? I need to write some extra tests, did not do it yet.

KKoukiou:
Sending some screenshots from the last iteration:
Screen Shot 2021-05-13 at 12 24 45
Screen Shot 2021-05-13 at 12 24 21
Screen Shot 2021-05-13 at 12 24 13
Screen Shot 2021-05-13 at 12 23 30
Screen Shot 2021-05-13 at 12 23 11

KKoukiou:
I see there is a test failing in two distros, I will figure out today. But keeping this open for a review anyway.

marusak:
There are no search labels when in mobile mode?
nolabel

KKoukiou:
> There are no search labels when in mobile mode?
> nolabel

No it's designed like this. I guess to save space.

KKoukiou:
@marusak do you consider the missing labels in the mobile mode a blocker? In the general case I think it's totally fine, in this specific case however, we have the same generic text 'All', which makes it indeed confusing.

I see two approaches here:
* Rename the default text to something like 'Choose active state', 'Choose unit file state' like here https://www.patternfly.org/v4/components/select/#single . This will solve the problem with the missing labels
* Stop using the mobile mode of the toolbar, but that will need a lot of custom CSS to make it wrap properly, and keeping the labels with their fields always in the same line. 


Any other thoughts on this?

garrett:
> Rename the default text to something like 'Choose active state', 'Choose unit file state'

Oof. Not a fan.

> Stop using the mobile mode of the toolbar, but that will need a lot of custom CSS to make it wrap properly, and keeping the labels with their fields always in the same line.

I know you're usually against custom CSS when possible — and I agree — but we might need it here. Depending on how it's implemented, it might be as simple as `flex-direction: column`. However, these seemingly "simple" things are usually not very simple as they appear. :grin: 

(I'd have to look. I've been juggling a lot today, but would be happy to look at this tomorrow.)

KKoukiou:
> > Rename the default text to something like 'Choose active state', 'Choose unit file state'
> 
> Oof. Not a fan.
> 
> > Stop using the mobile mode of the toolbar, but that will need a lot of custom CSS to make it wrap properly, and keeping the labels with their fields always in the same line.
> 
> I know you're usually against custom CSS when possible — and I agree — but we might need it here. Depending on how it's implemented, it might be as simple as `flex-direction: column`. However, these seemingly "simple" things are usually not very simple as they appear. grin
> 
> (I'd have to look. I've been juggling a lot today, but would be happy to look at this tomorrow.)

I already tried, the problem there is that the Toolbar items will just wrap anywhere, also betwen labels. So one mind end up with label in seperate line of the dedicated toolbar item.

garrett:
@KKoukiou: OK! :+1: 

(As you've already tried option # 2, we could merge this PR in with your suggestion # 1, provided there's aria text too, as in the example's source.)

martinpitt:
One [services related failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15818-20210518-103920-fc5c98a6-fedora-34/log.html#281), but I suppose it's not generally broken, just a flake.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -41,6 +41,7 @@ import './cockpit-components-table.scss';
  * See https://www.patternfly.org/v4/components/table/
  * Properties (all optional unless specified otherwise):
  * - caption
+ * - id: optional identifier
```
martinpitt, line 44: Please add this to `ListingTable.propTypes`. One of these days we really need to figure out how to validate propTypes during build or tests..

martinpitt (commented):
Thanks, really nice work! Code-wise this looks ok to me, just one nitpick. I'll leave the remaining  design discussions between you, Matej, and Garrett, I don't want to add more noise.

marusak:
> @marusak do you consider the missing labels in the mobile mode a blocker?

No, was just asking in general

KKoukiou:
@garrett @marusak ok one more iteration, IMO this is the  best so far. Tried to utilize select with checkboxes and chip groups in the toolbar.
Screen Shot 2021-05-27 at 14 26 15
Screen Shot 2021-05-27 at 14 29 22
Screen Shot 2021-05-27 at 14 29 32

garrett:
> IMO this is the best so far

:partying_face: 

It looks good in the screenshots. I'm going to build it here now.

KKoukiou:
> > IMO this is the best so far
> 
> partying_face
> 
> It looks good in the screenshots. I'm going to build it here now.

Don't hesitate to do any extra picky UI review here. Don't forget this was SPUR :)

marusak:
> This page needs filtering to be useful, so hiding it — especially on mobile — probably doesn't make sense. Hiding filtering behind an unlabeled icon makes it too easy to miss, and someone could try to find a service by scrolling and not realize how huge the list can be.
> (So I think we should drop the filter toggle and always show filtering, even on mobile.)

I am a bit on fence here. When filters are opened it can take almost half of the screen on smaller mobile devices. So hiding it makes sense. But I agree that only showing "2 filters applied" might be less useful. How about not showing "2 filters applied" but showing which filters are applied? (I mean those grey boxes).

KKoukiou:
> > This page needs filtering to be useful, so hiding it — especially on mobile — probably doesn't make sense. Hiding filtering behind an unlabeled icon makes it too easy to miss, and someone could try to find a service by scrolling and not realize how huge the list can be.
> > (So I think we should drop the filter toggle and always show filtering, even on mobile.)
> 
> I am a bit on fence here. When filters are opened it can take almost half of the screen on smaller mobile devices. So hiding it makes sense. But I agree that only showing "2 filters applied" might be less useful. How about not showing "2 filters applied" but showing which filters are applied? (I mean those grey boxes).

The '2 filters applied' string and generally the layout change in mobile is coming from the PF component internally, I can't control it. What I can do it to remove the toggle, in that case the mobile mode will have all filters visible and will show the badges.
The combination show badges but hide filters in mobile mode is not how the PF component is implemented unfortunately.

garrett:
> When filters are opened it can take almost half of the screen on smaller mobile devices.

People using mobile devices are used to scrolling. As long as filtering doesn't take so much of the screen that you can't see any of the list, it's OK.

Smaller devices, like an iPhone SE, have other issues too...

Screen Shot 2021-05-28 at 09 46 48

1. Tab overflow at the top is awful
2. Columns overlap (perhaps nowrap, which looks like it's being used here, should only be used at larger sizes)
3. Arranging the information in a different way on small screens may make sense?

Meanwhile, back to filtering: If we drop the collapse/expand method, that actually frees up some vertical space (where the icon itself is used):

Screenshot from 2021-05-27 09-49-29

The icon is 1/4 of the height of the filtering.

garrett:
I have a bunch of "Bad" services. Is this expected? And half of the "bad" ones are running, so they can't be that "bad"?

Screenshot_2021-05-27 Services - garrett Rain

garrett:
There's no border between the filtering and the list, and there probably should be.

pkg/systemd/services/services.scss
```diff
@@ -4,6 +4,21 @@
 @import "./timer.scss";
 @import "../../lib/table.css";
 
+// Wrap the filter info
+.services-toolbar {
+    display: flex;
+    flex-wrap: wrap;
+    border-bottom: 1px solid var(--pf-global--BorderColor--100);
```
garrett, line 11: Oh the border showed up in this commit too. Whoops! I'll probably be squashed down anyway... :grin:

garrett:
I turned on some performance "hacks" in d1f4399. These are valid to use, especially in cases like these. 

However, when profiling the page, I noticed there are like a _billion_ DOM changes (which then trigger recalculations and layout changes, which I limited with the CSS)... we should get to the heart of the problem and fix the tons of DOM changes. Ideally, we would render the items into an off-screen DOM fragment and then slam the entire DOM fragment into the page in one go to start, so everything's just calculated once.

Right now, it seems and feels like the page is being updated several times per row... I haven't looked at the code (yet) though; I'm just basing this on how slow everything feels and the performance graph.

At least, for now, the CSS helps and makes you not want to throw your computer out the window when trying to edit the page. :wink:

garrett:
I didn't attempt to fix the layout issues as seen in this screenshot (from above) yet:

Screen Shot 2021-05-28 at 09 46 48

I wanted to address the other issues first and ran out of time to address this one today.

pkg/systemd/services/services.scss
KKoukiou: Sure, please remember to add a comment

pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -146,15 +146,17 @@ svg {
     // when possible.
     // https://github.com/patternfly/patternfly/issues/4115
     &.ct-compact {
-        display: flex;
-        flex-wrap: wrap;
+        @media screen and (max-width: $pf-global--breakpoint--lg - 1) {
```
KKoukiou, line 149: That causes a compilation failure:
 Error: Undefined variable: "$pf-global--breakpoint--lg".
        on line 149 of pkg/lib/patternfly/patternfly-4-overrides.scss
>>         @media screen and (max-width: $pf-global--breakpoint--lg - 1) {

garrett: It's a SASS variable from pkg/lib/_global-variables.scss. We have to use it.

garrett:
I think at this point, we could consider the PR functionally done and split the following off into follow-ups:

1. Performance-impacting code issues.
2. The layout issue in small modes.

Hopefully we can address both before the release and have this PR and the follow-ups in the next release.

@KKoukiou: WDYT?

KKoukiou:
This is ready for another review.

marusak:
I like the design! Nice work :) I will now check the commits.
One think that I noticed but not sure if related - some units have empty file state. It might be related that we don't define translation for it?
Screenshot from 2021-06-02 12-12-29emtpies

marusak:
> One think that I noticed but not sure if related - some units have empty file state. It might be related that we don't define translation for it?

nvm, it is the same on master

pkg/systemd/services/services.jsx
```diff
@@ -151,6 +151,50 @@ class ServicesPage extends React.Component {
             tabErrors: {},
             isFullyLoaded: false,
         };
+
+        // Possible LoadState values: stub, loaded, not-found, bad-setting, error, merged, masked
+        // See: typedef enum UnitLoadStateState https://github.com/systemd/systemd/blob/main/src/basic/unit-def.h
+        this.loadState = {
+            stub: _("Stub"),
+            loaded: "",
```
marusak, line 162: Why we don't represent this one?

KKoukiou: Firstly, this did not change. 
If you ask generally - we don't specify the difference in the UI between loaded and unloaded units. We just show everything.

test/verify/check-system-services
```diff
@@ -84,6 +84,16 @@ class TestServices(MachineCase):
 
         self.browser.wait_in_text(self.svc_sel(service), state_new)
 
+    def select_file_state(self, state):
+        self.browser.click("#services-dropdown-file-state")
+        self.browser.set_checked("fieldset label:contains('{0}') input".format(state), True)
+        self.browser.click("#services-dropdown-file-state")
```
marusak, line 90: These two functions should also wait for the "chip" (is that the right name?) to appear.

test/verify/check-system-services
```diff
@@ -369,31 +382,34 @@ Unit=test.service
...
 
         # Select only failed services
-        b.select_from_dropdown("#services-dropdown-active-state", "failed")
+        init_filter_state()
+        self.select_active_state("Failed to start")
         self.wait_service_not_present("test.service")
         self.wait_service_present("test-fail.service")
 
         # Check filtering and selecting together - empty state
+        init_filter_state()
```
marusak, line 410: Can you please also replace one of these to check that those `x` for removing it from UI work?

marusak: Ideally in both ways - X to remove all of one group and X to remove just one specific item.

pkg/systemd/services/services.jsx
```diff
@@ -33,7 +33,6 @@ import {
     ToolbarContent,
     ToolbarItem,
```
marusak, line 34: There are two commits that "move create timer button". Can we squash those?

pkg/systemd/services/services.jsx
marusak: Right, I also noticed that we have checked for `loaded` in render method.

marusak (changes requested):
Nice one!
Can you please write more tests for those multi-choice selects?

pkg/systemd/services/services.jsx
KKoukiou: Done

test/verify/check-system-services
KKoukiou: Done

test/verify/check-system-services
KKoukiou: Done

KKoukiou:
@marusak @garrett I just realized something. We made the toolbar sticky but the header, which contains the tabs is not sticky. In fact I would expect that the header would be sticky as well. What do you think?

marusak:
> We made the toolbar sticky but the header, which contains the tabs is not sticky. In fact I would expect that the header would be sticky as well. What do you think?

I have noticed that as well but I thought it was conscious decision made by you and Garrett.

Otherwise all looks good to me!
Tests are not yet passing on all OS though.

KKoukiou:
> > We made the toolbar sticky but the header, which contains the tabs is not sticky. In fact I would expect that the header would be sticky as well. What do you think?
> 
> I have noticed that as well but I thought it was conscious decision made by you and Garrett.
> 
> Otherwise all looks good to me!
> Tests are not yet passing on all OS though.

yes seems the new tests are not happy in all distros. Will fix and ping you again .

KKoukiou:
filrefox somehow runs in smaller window and renders mobile mode. Therefore the test is failing - need to adjust

KKoukiou:
@marusak tests are ready now.

garrett:
>> We made the toolbar sticky but the header, which contains the tabs is not sticky. In fact I would expect that the header would be sticky as well. What do you think?
>
>  I have noticed that as well but I thought it was conscious decision made by you and Garrett.

I can't remember if that's intentional, but I do know that all of us were concerned about the filter height on smaller screens, and this would bump that up a bit.

And most people only really care about the services, right? And even when an admin might care about something else, they'd usually switch first, right?

So I think it's OK to not have the tabs sticky anyway, whether or not it's intentional.

That said: If we do want the tabs sticky too, then that should probably be in a follow-up.

marusak (approved):
Thanks!

PR #15819, opened 2021-05-11
systemd: Fix insights URL

martinpitt:
Drop the extra /insights suffix, which is not valid.

https://bugzilla.redhat.com/show_bug.cgi?id=1958379

### DIFF
--- a/pkg/systemd/overview-cards/insights.jsx
+++ b/pkg/systemd/overview-cards/insights.jsx
@@ -132,7 +132,7 @@ export class InsightsStatus extends React.Component {
 
         let url;
         if (this.state.id)
-            url = "https://cloud.redhat.com/insights/inventory/" + this.state.id + "/insights";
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):
Random that these two are in the same PR, but otherwise OK :)

martinpitt:
@marusak: Ideally this really belongs into subscription-manager as a notification. That already computes the URL. I thought we already do that,  but apparently not :sob: I'm afraid we have no end-to-end way to validate that, as we don't have our own insights instance.

The second commit just needed to go "somewhere". It was part of an earlier PR that I thought would be straightforward, but that got stuck.

martinpitt:
@marusak : I'm not entirely sure if that actually ever worked. [here](https://github.com/candlepin/subscription-manager/blob/master/cockpit/src/insights.jsx#L484) is the corresponding URL from sub-man, and this has never changed -- it's the original code that @mvollmer wrote over a year ago.

PR #15823, opened 2021-05-12
storage: Improved passphrase handling

mvollmer:
Whenever a passphrase is needed, we now also try the stored passphrase
in addition to one recovered from Clevis.

But while a passphrase recovered from Clevis can be assumed to work, a
stored passphrase can easily be wrong.  Therefore, if a operation that
requires a passphrase fails, we now give the use the opportunity to
provide a different one.

This also fixes the flow when adding/changing Clevis keys.  Earlier,
this was done with two dialogs where the first asks for the passphrase
and the second will show the error if that passphrase was wrong,
without giving the user a opportunity to correct the passphrase.  Now
this is still done in two steps, but the second one will show a
passphrase input if the operation failed because of a wrong
passphrase.

Unlocking a disk is one of the operations that the above applies to.
It is however a bit special: It is possible to unlock a disk with a
stored passphrase or a Clevis passphrase without having the passphrase
travel through the Browser.  So we do this.

Besides improving passphrase handling for the existing operations,
this also lays the groundwork for a upcoming combined Unlock-and-Mount
operation.  That operation needs to automatically use a stored
passphrase, not let the passphrase travel through the Browser, fall
back to asking for a passphrase if the recovered one fails, and open a
dialog without doing any opportunistic unlocking just to see what will
work.

And if we anyway need all these features, why not use them in all
places that deal with passphrases.  Hence this change.

### DIFF
--- a/pkg/storaged/clevis-luks-passphrase.sh
+++ b/pkg/storaged/clevis-luks-passphrase.sh
@@ -2,8 +2,25 @@
 
 set -eu
 
+# clevis-luks-passphrase [--type] DEV
+#
+# Try to recover a passphrase via clevis that will open DEV, and
+# output it on stdout.  Such a passphrase can be used for things like
+# resizing of LUKSv2 containers, or adding/changing keys.
+#
+# If "--type" is given, this tool writes just "clevis" to stdout when
+# a passphrase is found, instead of the actual passphrase.  This is
+# useful to limit exposure of the passphrase when all you want to know
+# is whether "clevis luks unlock" can be expected to succeed.
+
+opt_type=no
+if [ $# -gt 1 ] && [ "$1" = "--type" ]; then
+    opt_type=yes
+    shift
+fi
+
 if [ $# -ne 1 ]; then
-    echo "usage: $0 DEV" >&2
+    echo "usage: $0 [--type] DEV" >&2
     exit 1
 fi
 
@@ -21,7 +38,11 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
     luksmeta show -d "$DEV" | while read slot state uuid; do
         if [ "$state" = "active" -a "$uuid" = "$CLEVIS_UUID" ]; then
             if pp=$(luksmeta load -d "$DEV" -s "$slot" | clevis decrypt); then
-                printf '%s\n' "$pp"
+                if [ "$opt_type" = yes ]; then
+                    echo clevis
+                else
+                    printf '%s\n' "$pp"
+                fi
                 break
             fi
         fi
@@ -32,8 +53,12 @@ elif cryptsetup isLuks --type luks2 "$DEV"; then
         tok=`cryptsetup token export --token-id "$id" "$DEV"`
         jwe=`printf '%s\n' "$tok" | jose fmt -j- -Og jwe -o- | jose jwe fmt -i- -c`
 
-        if pt=`echo -n "$jwe" | clevis decrypt`; then
-            printf '%s\n' "$pt"
+        if pt=`printf '%s' "$jwe" | clevis decrypt`; then
+            if [ "$opt_type" = yes ]; then
+                echo clevis
+            else
+                printf '%s\n' "$pt"
+            fi
             break
         fi
     done
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -36,6 +36,7 @@ import { job_progress_wrapper } from "./jobs-panel.jsx";
 
 import { FilesystemTab, is_mounted, mounting_dialog } from "./fsys-tab.jsx";
 import { CryptoTab } from "./crypto-tab.jsx";
+import { get_existing_passphrase } from "./crypto-keyslots.jsx";
 import { BlockVolTab, PoolVolTab } from "./lvol-tabs.jsx";
 import { PVolTab, MDRaidMemberTab, VDOBackingTab } from "./pvol-tabs.jsx";
 import { PartitionTab } from "./part-tab.jsx";
@@ -199,15 +200,20 @@ function create_tabs(client, target, is_partition) {
     }
 
     function unlock() {
-        if (!client.features.clevis)
-            return unlock_with_passphrase();
-        else {
-            return clevis_unlock()
-                    .then(null,
-                          function () {
-                              return unlock_with_passphrase();
-                          });
-        }
+        var crypto = client.blocks_crypto[block.path];
+        if (!crypto)
+            return;
+
+        return get_existing_passphrase(block, true).then(type => {
+            if (type == "stored") {
+                return (crypto.Unlock("", {})
+                        .catch(() => unlock_with_passphrase()));
+            } else if (type == "clevis") {
+                return (clevis_unlock()
+                        .catch(() => unlock_with_passphrase()));
+            } else
+                unlock_with_passphrase();
+        });
     }
 
     function unlock_with_passphrase() {
@@ -215,33 +221,17 @@ function create_tabs(client, target, is_partition) {
         if (!crypto)
             return;
 
-        /* If there is a stored passphrase, the Unlock method will
-         * use it unconditionally.  So we don't ask for one in
-         * that case.
-         *
-         * http://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Block.html#gdbus-method-org-freedesktop-UDisks2-Block.GetSecretConfiguration
-         */
-        return block.GetSecretConfiguration({}).then(function (items) {
-            for (var i = 0; i < items.length; i++) {
-                if (items[i][0] == 'crypttab' &&
-                    items[i][1]['passphrase-contents'] &&
-                    utils.decode_filename(items[i][1]['passphrase-contents'].v)) {
-                    return crypto.Unlock("", { });
+        dialog_open({
+            Title: _("Unlock"),
+            Fields: [
+                PassInput("passphrase", _("Passphrase"), {})
+            ],
+            Action: {
+                Title: _("Unlock"),
+                action: function (vals) {
+                    return crypto.Unlock(vals.passphrase, {});
                 }
             }
-
-            dialog_open({
-                Title: _("Unlock"),
-                Fields: [
-                    PassInput("passphrase", _("Passphrase"), {})
-                ],
-                Action: {
-                    Title: _("Unlock"),
-                    action: function (vals) {
-                        return crypto.Unlock(vals.passphrase, {});
-                    }
-                }
-            });
         });
     }
 
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -38,7 +38,7 @@ import {
     dialog_open,
     SelectOneRadio, TextInput, PassInput, Skip
 } from "./dialog.jsx";
-import { decode_filename, block_name } from "./utils.js";
+import { array_find, decode_filename, block_name } from "./utils.js";
 import { fmt_to_fragments } from "./utilsx.jsx";
 import { StorageButton } from "./storage-controls.jsx";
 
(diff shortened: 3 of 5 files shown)


pkg/storaged/crypto-keyslots.jsx
```diff
@@ -169,20 +181,53 @@ export function existing_passphrase_fields(explanation) {
...
 
-export function get_existing_passphrase(dlg, block) {
-    const prom = clevis_recover_passphrase(block).then(passphrase => {
-        if (passphrase == "") {
-            dlg.set_values({ needs_explicit_passphrase: true });
-            return null;
-        } else {
-            return passphrase;
+function get_stored_passphrase(block, just_type) {
+    // XXX - With just_type, don't call GetSecretConfiguration
```
mvollmer, line 185: This needs doing.

mvollmer: Done

KKoukiou (commented):
I am having a bit hard time to understand the changes exactly because I am not familiar with Clevis keys etc. But I want to give it a try however.


The main question I have is why do we want a second dialog opening when unlocking the disk failed? I would expect the same dialog remaining open, and invaliding the 'Password' field.

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -1,7 +1,13 @@
-#! /bin/bash
+#! /bin/sh
 
 set -eu
 
+type=no
+if [ $# -gt 1 ] && [ "$1" = "--type" ]; then
```
martinpitt, line 6: This is fine. Just because you were asking about shell yesterday, another common idiom is to use variable defaults `"[ ${1:-}" = "--type" ]`. This gets along with `set -u` without having to explicitly check for existence.

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -1,7 +1,13 @@
...
+    type=yes
+    shift;
```
martinpitt, line 8: unnecessary semicolon

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -21,7 +27,11 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
     luksmeta show -d "$DEV" | while read slot state uuid; do
         if [ "$state" == "active" -a "$uuid" == "$CLEVIS_UUID" ]; then
             if pp=$(luksmeta load -d "$DEV" -s "$slot" | clevis decrypt); then
-                echo $pp
+                if [ "$type" = yes ]; then
+                    echo clevis
```
martinpitt, line 31: In this case the just computed `$pp` is entirely ignored. Does this actually have to call luksmeta|clevis in that  case, i.e. does that have a side effect?

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -1,7 +1,13 @@
-#! /bin/bash
+#! /bin/sh
 
 set -eu
 
+type=no
```
martinpitt, line 5: Please add a comment what this means. As the usual noun ("data type") a yes/no makes no sense. As a verb ("hacking on the keyboard") it does, but it'd be nice to document the behaviour.

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -30,10 +40,14 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
 elif cryptsetup isLuks --type luks2 "$DEV"; then
     for id in `cryptsetup luksDump "$DEV" | sed -rn 's|^\s+([0-9]+): clevis|\1|p'`; do
         tok=`cryptsetup token export --token-id "$id" "$DEV"`
-        jwe=`jose fmt -j- -Og jwe -o- <<<"$tok" | jose jwe fmt -i- -c`
```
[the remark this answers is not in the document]
martinpitt: `<<<` is a bashism, this replaces it with POSIX shell. This goes together with changing the hashbang "bash" to "sh". This could be its own independent commit to make it clearer. (FTR: There is no problem with using bash, this is just a stylistic decision)

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -30,10 +40,14 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
...
+        jwe=`printf '%s\n' "$tok" | jose fmt -j- -Og jwe -o- | jose jwe fmt -i- -c`
 
         if pt=`echo -n "$jwe" | clevis decrypt`; then
-            echo $pt
+            if [ "$type" = yes ]; then
+                echo clevis
```
martinpitt, line 47: Same here, `$pt` is ignored. It feels like the whole script could short-circuit all the program calls and directly exit with "echo clevis" if `--type` is given. Which then bears the question how much of the script actually remains to justify calling it in the first place in the `--type` case?

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -21,7 +27,11 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
     luksmeta show -d "$DEV" | while read slot state uuid; do
         if [ "$state" == "active" -a "$uuid" == "$CLEVIS_UUID" ]; then
             if pp=$(luksmeta load -d "$DEV" -s "$slot" | clevis decrypt); then
-                echo $pp
+                if [ "$type" = yes ]; then
+                    echo clevis
+                else
+                    echo $pp
```
martinpitt, line 33: This is an user-defined passphrase, isn't it? Does this have any chance of containing *any* unrestricted characters, like `-` or `\n`?  If this can't be 100% guaranteed to be only e.g. hex characters, this ought to at least be quoted, and better use `printf` to avoid stumbling over leading `-`.

martinpitt (changes requested):
Thanks! Some picking on the shell script.

mvollmer:
> The main question I have is why do we want a second dialog opening when unlocking the disk failed? I would expect the same dialog remaining open, and invaliding the 'Password' field.

This is just how adding a clevis key was designed: the first dialog asks for the parameters (including a passphrase if necessary), then we take those parameter and retrieve the key, and the second dialog asks the user to confirm the fingerprint of the key. Only then is the passphrase actually used and the operation might fail because of it.

Thus, hmm, should we maybe always ask for the passphrase in the second dialog?

pkg/storaged/content-views.jsx
```diff
@@ -206,49 +207,42 @@ function create_tabs(client, target, is_partition) {
...
-                          });
-        }
+        var crypto = client.blocks_crypto[block.path];
+        if (!crypto)
+            return;
+
+        return get_existing_passphrase(block, true).then(type => {
+            if (type == "stored") {
+                return (crypto.Unlock("", {})
+                        .catch(() => {
```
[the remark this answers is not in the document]
mvollmer: The difference is in the return value. As written, this returns `undefined`, with a single-line expression, it would "accidentally" return the return value of `unlock_with_passphrase()`. I like to make it expicit that the return value of `unlock_with_passphrase() is not actually used. But, yeah, you people keep telling me otherwise, so *shrug*. :-)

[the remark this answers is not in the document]
mvollmer: Done.

pkg/storaged/content-views.jsx
```diff
@@ -206,49 +207,42 @@ function create_tabs(client, target, is_partition) {
...
+                            unlock_with_passphrase();
+                        }));
+            } else if (type == "clevis") {
+                return (clevis_unlock()
+                        .catch(() => {
+                            unlock_with_passphrase();
```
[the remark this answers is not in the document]
mvollmer: Done.

pkg/storaged/clevis-luks-passphrase.sh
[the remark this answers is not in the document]
mvollmer: This is used with `cockpit.script` which always uses `/bin/sh`. Previously, the script would not run at all on Debian (because there is no `clevis` binary there). Now it runs always and thus the old version would fail on Debian.

There is no strong reason to run it also on Debian, but it also does no harm. And since I was changing it anyway, why not make it work with plain /bin/sh.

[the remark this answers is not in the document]
mvollmer: There actually is a `clevis` package in Debian, let's test it!

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: Fixed

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: Yeah, I was asking myself the same... the main point of the `--type` option is to avoid sending the actual passphrase back to the browser and smear all over the JavaScript heap.

The guarantee we need is that if this script says "clevis", we can assume that `clevis luks unlock` will succeed. This doesn't only need a configured key slot, but we also need to be in an environment where it will work (tang server reachable, etc).

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: Would `opt_type` be enough?

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: See above. This script should only output "clevis" if there is at least one clevis slot that actually works right now. This is so that the UI can adjust nicely before trying to run `clevis luks unlock`.

(Hmm, `clevis luks unlock` has a -t option for testing a slot... we could use that in a loop over all slots. But, hmm, hmm, that uses cryptsetup --test-passphrase, which is suuuper slow...)

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: > This is an user-defined passphrase, isn't it?

Not really, it's a random string produced by clevis itself. But your points are still valid, let's use printf.

pkg/storaged/clevis-luks-passphrase.sh
martinpitt: To me it was quite obvious that it is an option, but not what it actually *means*. So I'd prefer a comment (of course you can rename it to opt_ also)

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: Done.

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: Done.

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: I think we need to call `clevis decrypt`, `clevis luks unlock -t SLT` does the same. So I wouldn't change anything here.

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: So, again, I wouldn't make any further changes.

pkg/storaged/clevis-luks-passphrase.sh
mvollmer: Ah, I see.  I have added a full help text.

pkg/storaged/clevis-luks-passphrase.sh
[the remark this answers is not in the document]
mvollmer: See #15835. I have moved the /bin/sh fixes to that PR.

mvollmer:
The testClevisTang failure on debian-stable is due to a bug in its clevis-luks-bind script: It will exit 0 when given a wrong passphrase. We didn't test this case before so #15835 didn't catch it. -> naughty pattern

This is fixed in debian-testing and everywhere else.

I'll push some more cleanups that remove workarounds for old versions of clevis.

mvollmer:
> I'll push some more cleanups that remove workarounds for old versions of clevis.

Did that separately: #15850

mvollmer:
`debian-stable@bots#2035` was wrong, 2035 is the known issue, not the PR that adds it as a naughty pattern.

mvollmer:
Alright, now that removing a passphrase asks for any other passphrase, we should also cover it here. 

This PR is about operations on a crypto thing that require a passphrase, like resizing, adding keys, and unlocking. Now removing a key is also one of those.

mvollmer:
>  so maybe we leave this for later then.

#15853

martinpitt (approved):
Danke! LGTM. (Again, not signing off the design)

KKoukiou (approved):

PR #15824, opened 2021-05-12
packit: Run on CentOS stream

martinpitt:
This should allow us to catch test bugs before releases (like the recent
introduction of `dataclass` into testlib, which isn't available in
RHEL/CentOS 8 yet).

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -3,7 +3,8 @@ actions:
   create-archive:
     # the upstream git spec file has unresolved macros which are normally put
     # in at `make dist` time; they are not important for CI
-    - sh -c 'sed "s/%{npm-version:.*}/0/" tools/cockpit.spec > cockpit.spec'
+    # also, build optional packages for CentOS 8
+    - sh -c 'sed "s/%{npm-version:.*}/0/; s/build_optional 0/build_optional 1/" tools/cockpit.spec > cockpit.spec'
     # this is being triggered immediately on a pull_request event; wait for
     # build-dist.yml to generate the dist tarball
     - env AUTOGEN_CONFIGURE_ARGS='--disable-polkit --disable-ssh --disable-pcp --disable-doc --with-systemdunitdir=/invalid CPPFLAGS=-Itools/mock-build-env' PKG_CONFIG_PATH=tools/mock-build-env test/make_dist.py --wait
@@ -14,6 +15,6 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      # once this works, move to fedora-all
       - fedora-33
       - fedora-34
+      - centos-stream-8-x86_64


martinpitt:
Hmm, that [doesn't work](http://artifacts.dev.testing-farm.io/175b83b1-00d6-47a4-9d9d-54df9fa391f8/), a lot of basic packages like firewalld or make are not available, and it also seems to not find the packages from copr. @thrix, @TomasTomecek is that expected? Do you happen to know any existing project which runs packit tests on CentOS 8 stream?

thrix:
So I am looking into this now, seems repos are there ...

```
# yum repolist
repo id                                                                                          repo name
appstream                                                                                        CentOS Stream 8 - AppStream
baseos                                                                                           CentOS Stream 8 - BaseOS
extras       
```

At least the basic ones ...

thrix:
@martinpitt so your problem is `cockpit-tests` not available ....

The output is a bit misleading, `tmt` does `rpm -q package` I believe, all repos are there ... PowerTools should be enabled ... but the package `cockpit-tests` is not available seems ...

martinpitt:
Hah, indeed, thanks @thrix and sorry for the noise! Indeed the [centos 8 build](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-15824/centos-stream-8-x86_64/02183190-cockpit/) only builds the baseos packages and skips the appstream parts. That's what we want in general for *real* centos, but not for packit. (Hopefully) fixed.

martinpitt:
Cool, this worked! Triggering the mandatory f-33 test.

packit.yaml
```diff
@@ -14,6 +15,6 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      # once this works, move to fedora-all
       - fedora-33
       - fedora-34
+      - centos-stream-x86_64
```
TomasTomecek, line 20:
```suggestion
      - centos-stream-8-x86_64
```

to make this explicit for 8; the target you provided won't be working in future since won't be clear if you want 8 or 9 :)

TomasTomecek:
> Hmm, that [doesn't work](http://artifacts.dev.testing-farm.io/175b83b1-00d6-47a4-9d9d-54df9fa391f8/), a lot of basic packages like firewalld or make are not available, and it also seems to not find the packages from copr. @thrix, @TomasTomecek is that expected? Do you happen to know any existing project which runs packit tests on CentOS 8 stream?

Catching up with last week's notifications; glad you gentlemen solved this out :) happy to see the stream 8 test suite pass!

packit.yaml
martinpitt: Thanks! Applied. (And indeed in the  future I really want 9 as well :grin: )

packit.yaml
```diff
@@ -3,7 +3,8 @@ actions:
   create-archive:
     # the upstream git spec file has unresolved macros which are normally put
     # in at `make dist` time; they are not important for CI
-    - sh -c 'sed "s/%{npm-version:.*}/0/" tools/cockpit.spec > cockpit.spec'
+    # also, build optional packages for CentOS 8
```
marusak, line 6: Heh, was wondering how would you convince that logic to build all of it - now I see :)

marusak (approved):
Thanks!

PR #15826, opened 2021-05-12
storage: use SHA256 for Tang fingerprints

allisonkarlitskaya:
Fix for https://github.com/cockpit-project/bots/pull/2006.

For some time now, `tang-show-keys` has been showing the user key
fingerprints based on SHA256, while we've been computing them in Cockpit
using SHA1.  Our advice for the user to compare these two values isn't
very useful in that case.

This was exposed by a change to jose to default to using SHA256 instead
of SHA1 if no hash algorithm is specified.  We invoke jose directly from
the integration tests.

Switch cockpit over to computing hashes using SHA256, and adjust our
invocation of jose to explicitly specify the algorithm as SHA256 as
well, so that we remain compatible also with older versions.

The longer fingerprint causes the text to wrap in a somewhat ugly way.  Tagging @garrett for review and suggested changes there.


allisonkarlitskaya:
Here's the screenshot of the problem that I describe: 

image

garrett:
We might want to make the modal dialog wider to display the string all in one line. 

The string should probably be in a (relatively new) clipboard copy widget, with the assumption that someone might want to copy to verify.

It could be one of two variants:
https://www.patternfly.org/v4/components/clipboard-copy#read-only (if this fits or if it allows wrapping)
or
https://www.patternfly.org/v4/components/clipboard-copy#inline-compact-in-sentence (long copy string in block, specifically)

I think it's still useful to show the string in a larger monospaced font to make manual comparisons easier, so we probably still want to do some custom styling even if we do use this other widget.

allisonkarlitskaya:
This is green against cockpit-project/bots#2006 (which it blocks) so subject to approval @mvollmer it might be nice to land this now (and handle the UI issues in a separate PR).  @garrett OK?

martinpitt (approved):
Nice that this works against older jose versions as well! Cheers!

PR #15828, opened 2021-05-14
lib: Remove unused parameters in callbacks

marusak:
This gets reported when this library is being used in other projects
with different set of eslint rules.


martinpitt (approved):
Danke!

Issue #15821, opened 2021-05-11
users: `lock` and `expire` are two different operations and the wording should not be used interchangeably


KKoukiou:
This is the dialog for the account 'expiration'. See the wording for 'locking' and 'expiring' is used interchangeably as if they are the same operation.

![117258725-8f033900-ae4d-11eb-8fd3-51c8b2901a60](https://user-images.githubusercontent.com/14921356/117870956-1b25be00-b29d-11eb-853f-189f371730a8.png)

Looking at this state I was super confused! 'Lock account' is checked, and then there is a text saying, 'Never lock account'. Let's decide :) 
![Screen Shot 2021-05-12 at 21 18 32](https://user-images.githubusercontent.com/14921356/117872104-9340b380-b29e-11eb-8415-392601a0eb00.png)

In addition the buttons for choosing the expiration date (also for the password expiration one line bellow) are not using verbs but the information text instead. 

All in all I believe the final UI should be something close to this and eliminate the vague wording.
![Screen Shot 2021-05-12 at 21 06 10](https://user-images.githubusercontent.com/14921356/117871164-5fb15980-b29d-11eb-8f9b-b2ea84a6515e.png)

Issue #15794, opened 2021-05-06
users: Misaligned radio button fields in lock account dialog


KKoukiou:
![Screen Shot 2021-05-07 at 09 29 23](https://user-images.githubusercontent.com/14921356/117258725-8f033900-ae4d-11eb-8fd3-51c8b2901a60.png)

KKoukiou:
In addition that form field which is supposed to be a date field is using TextInput. This should be  also switched to Datepicker.

PR #15832, opened 2021-05-17
users: Improve account lock and expire dialogs

KKoukiou:
Fixes https://github.com/cockpit-project/cockpit/issues/15821
Fixes https://github.com/cockpit-project/cockpit/issues/15794

### DIFF
--- a/pkg/lib/cockpit-components-dialog.jsx
+++ b/pkg/lib/cockpit-components-dialog.jsx
@@ -236,7 +236,7 @@ export class Dialog extends React.Component {
             </Popover>;
 
         return (
-            <Modal position="top" variant="medium"
+            <Modal position="top" variant={this.props.variant || "medium"}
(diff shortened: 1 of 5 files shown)


martinpitt (approved):
This looks much better, thank you! During testing I noticed that the "Lock account" checkbox is broken (it always immediately unchecks again), but this is unrelated.

KKoukiou:
> This looks much better, thank you! During testing I noticed that the "Lock account" checkbox is broken (it always immediately unchecks again), but this is unrelated.

Can you reproduce on master as well? Indeed I don't see how I could have broken this, since I did not touch the Checkbox. However, i just tried it on master and it works as expected.

martinpitt:
> Can you reproduce on master as well?

Yes, I did compare it against current master during my review (hence the "not a regression from this PR")

PR #15833, opened 2021-05-17
packagekit: fix some react warnings

### LINKED ISSUE #15768: React key warning in OsUpdates component
I noticed in browser console the following warning:
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `BodyRow`. It was passed a child from UpdatesList. See https://fb.me/react-warning-keys for more information.
    in span (created by UpdatesList)
    in BodyRow (created by BaseBody)
    in tbody (created by TbodyBase)
    in TbodyBase (created by Tbody)
    in Tbody (created by BodyWrapper)
    in BodyWrapper (created by BaseBody)
    in BaseBody (created by Context.Consumer)
    in Body (created by ContextBody)
    in ContextBody (created by Context.Consumer)
    in TableBody (created by ListingTable)
    in table (created by TableComposableBase)
    in TableComposableBase (created by TableComposable)
    in TableComposable (created by Provider)
    in Provider (created by Table)
    in Table (created by ListingTable)
    in ListingTable (created by UpdatesList)
    in UpdatesList (created by CardsPage)
    in div (created by CardBody)
    in CardBody (created by CardsPage)
    in article (created by Card)
    in Card (created by CardsPage)
    in CardsPage (created by OsUpdates)
    in div (created by Gallery)
    in Gallery (created by OsUpdates)
    in section (created by PageSection)
    in PageSection (created by OsUpdates)
    in main (created by Page)
    in div (created by Page)
    in Page (created by OsUpdates)
    in OsUpdates
```

KKoukiou:
Fixes #15768

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -480,6 +480,7 @@ class ApplyUpdates extends React.Component {
     }
 
     componentDidMount() {
+        this._mounted = true;
         var transactionPath = this.props.transaction;
 
         PK.watchTransaction(transactionPath, {
@@ -489,6 +490,9 @@ class ApplyUpdates extends React.Component {
                 // small timeout to avoid excessive overlaps from the next PackageKit progress signal
                 PK.call(transactionPath, "org.freedesktop.DBus.Properties", "GetAll", [PK.transactionInterface], { timeout: 500 })
                         .done(reply => {
+                            if (this._mounted === false)
+                                return;
+
                             const percent = reply[0].Percentage.v;
                             let remain = -1;
                             if ("RemainingTime" in reply[0])
@@ -516,6 +520,10 @@ class ApplyUpdates extends React.Component {
                         });
             },
         });
+    }
+
+    componentWillUnmount() {
+        this._mounted = false;
     }
 
     render() {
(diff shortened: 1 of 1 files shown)


pkg/packagekit/updates.jsx
```diff
@@ -489,6 +489,9 @@ class ApplyUpdates extends React.Component {
                 // small timeout to avoid excessive overlaps from the next PackageKit progress signal
                 PK.call(transactionPath, "org.freedesktop.DBus.Properties", "GetAll", [PK.transactionInterface], { timeout: 500 })
                         .done(reply => {
+                            if (this._mounted === false)
```
martinpitt, line 492: `this._mounted` is never set to true -- you want to do that at the start of componentDidMount(). `_mounted` seems to be a common convention (judging by the many stackoverflow and other articles), but it's not a React builtin variable.

Also, please replace `mounted === false` with `!mounted`, it's a simple boolean.

martinpitt (changes requested):
Thanks for fixing that!

pkg/packagekit/updates.jsx
KKoukiou: This works however because of the === which compares also types apart from value. do undefined !== false. Since you find it confusing I will adjust.

martinpitt: Right, I know that distinction -- but AFAICS we never want to explicitly set it to `undefined` or `null`, we are only interested in in true/false here.

martinpitt (dismissed):
The `= true` fix landed, thanks! So this looks good to me, thanks! (the `=== false` thing is just stylistic, feel free to push or not)

KKoukiou:
> The `= true` fix landed, thanks! So this looks good to me, thanks! (the `=== false` thing is just stylistic, feel free to push or not)

Repushed with the condition adjustment as you suggested.

martinpitt (approved):

PR #15835, opened 2021-05-18
test: Run TestStorageLuks.testClevis also on the Debians

mvollmer:
- [x] #15844

### DIFF
--- a/pkg/storaged/clevis-luks-passphrase.sh
+++ b/pkg/storaged/clevis-luks-passphrase.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#! /bin/sh
 
 set -eu
 
@@ -19,9 +19,9 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
     luksmeta test -d "$DEV" 2>/dev/null || exit 0
 
     luksmeta show -d "$DEV" | while read slot state uuid; do
-        if [ "$state" == "active" -a "$uuid" == "$CLEVIS_UUID" ]; then
+        if [ "$state" = "active" -a "$uuid" = "$CLEVIS_UUID" ]; then
             if pp=$(luksmeta load -d "$DEV" -s "$slot" | clevis decrypt); then
-                echo $pp
+                printf '%s\n' "$pp"
                 break
             fi
         fi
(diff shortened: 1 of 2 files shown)


mvollmer:
This needs the bash->sh fixes from #15823, I'll move it over to here, as suggested.

pkg/storaged/clevis-luks-passphrase.sh
```diff
@@ -21,7 +21,7 @@ if cryptsetup isLuks --type luks1 "$DEV"; then
     luksmeta show -d "$DEV" | while read slot state uuid; do
         if [ "$state" == "active" -a "$uuid" == "$CLEVIS_UUID" ]; then
             if pp=$(luksmeta load -d "$DEV" -s "$slot" | clevis decrypt); then
-                echo $pp
+                printf '%s\n' "$pp"
```
martinpitt, line 24: I think you should then also change the hashbang to /bin/sh to indicate that. There are several test failures anyway, so you can slip that into the next push. Thanks!

mvollmer: Ah, good point.

mvollmer: Done, and fixed the rest as well.

martinpitt (dismissed):
Hah, more bashisms. LGTM, good to go once tests go green. Danke!

mvollmer:
ubuntu-2004 seems to have a old version of tang that still suffers from startup/key creation races.  We had workarounds for those for some time but removed them when the issues got fixed in the Fedoras.

Instead of bringing back the workarounds, let's avoid the races by not creating our own keys: #15844

martinpitt (approved):
All Debian/Ubuntu tests passed, yay!

test/verify/check-storage-luks
```diff
@@ -139,11 +139,14 @@ class TestStorageLuks(StorageCase):
...
         wait(lambda: m.execute("(loginctl list-users | grep admin) || true") == "")
 
-    @skipImage("No clevis/tang", "debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable")
     def testClevisTang(self):
         m = self.machine
         b = self.browser
 
+        if m.image == "ubuntu-stable":
+            # ubuntu-stable needs this workaround.
+            m.execute("mkdir -p /var/db/tang")
```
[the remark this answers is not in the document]
mvollmer: Not that I know of but tang-8-2 in debian-testing doesn't have this problem while tang-8-1 in ubuntu-stable does. So I am guessing the bug is known and fixed and the fix will come to ubuntu eventually...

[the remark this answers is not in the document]
martinpitt: Thanks! Indeed the [changelog for 8-2](https://tracker.debian.org/news/1239254/accepted-tang-8-2-source-into-unstable/) mentions exactly this.

PR #15837, opened 2021-05-18
testlib: add locale message exception for fedora-coreos

allisonkarlitskaya:
testlib: add locale message exception for fedora-coreos
    
    fedora-coreos recently dropped the glibc langauge packs, meaning that
    we're now getting the "unknown locale" warning there as well (since the
    browser sends 'en-us').
    
    We have an exception in place for debian-*, so extend that to
    fedora-coreos.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1251,8 +1251,8 @@ class MachineCase(unittest.TestCase):
         if "TEST_AUDIT_NO_SELINUX" not in os.environ:
             messages += machine.audit_messages("14", cursor=cursor)  # 14xx is selinux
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
:woman_pilot: :compass:  :muscle:

martinpitt:
The [remaining six failures](https://logs.cockpit-project.org/logs/pull-15837-20210518-110048-a1255a1d-fedora-coreos-bots-2008/log.html#138) are a new SELinux violation which needs to be handled as a naughty in https://github.com/cockpit-project/bots/pull/2008 .

PR #15841, opened 2021-05-19
test: Use basic.target instead of sockets.target in TestServices

### LINKED ISSUE #15838: tests: services: remove dependency from sockets.target
In rhel-8-5.0 tests for services page seem to fails on sockets.target not running. There is not need to check this specific target, let's create a robust dummy one.

```
  File "test/verify/check-system-services", line 187, in testBasic
    self.wait_service_state("sockets.target", "active")
  File "test/verify/check-system-services", line 82, in wait_service_state
```

martinpitt:
In some test environments sockets.target is (inexplicably) not running,
even though basic.target `Wants=` it. Check `basic.target` instead,
which is in the critical path for a successful boot, and also
alphabetically earlier so that we can see it in screenshots.

Fixes #15838

### DIFF
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -184,7 +184,7 @@ Unit=test.service
         # Selects Targets tab
         self.pick_tab(2)
         b.wait_not_in_text("#services-list", "test")
-        self.wait_service_state("sockets.target", "active")
+        self.wait_service_state("basic.target", "active")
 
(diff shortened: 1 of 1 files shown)


martinpitt:
I sent https://github.com/cockpit-project/bots/pull/2033 to fix FCOS. It's unrelated, but failing on most PRs now.

KKoukiou (approved):

PR #15843, opened 2021-05-19
improvements for node_modules publish workflow

allisonkarlitskaya:
- do compares against the `base.sha` of the PR (similar to #15842)
 - add a PR: ${URL} line to the top of the commit message

### DIFF
--- a/.github/workflows/node_modules-publish.yml
+++ b/.github/workflows/node_modules-publish.yml
@@ -13,6 +13,7 @@ jobs:
       GIT_DIR: git-dir.git
       HEAD_SHA: ${{ github.event.workflow_run.pull_requests[0].head.sha }}
       BASE_SHA: ${{ github.event.workflow_run.pull_requests[0].base.sha }}
+      PULL_URL: ${{ github.event.workflow_run.pull_requests[0].url }}
 
     steps:
       - name: Set up work area
@@ -65,14 +66,19 @@ jobs:
       - name: Stage git commit
         run: |
           set -ux
+          cat - artifact/npm-install.log > commit-message <<EOF
+          PR: $(sed 's@//api.@//@; s@/repos/@/@; s@/pulls/@/pull/@;' <<< "${PULL_URL}")
+
+          EOF
+          rm artifact/npm-install.log
           mv artifact/package.json node_modules/.package.json
           mv artifact/package-lock.json node_modules/.package-lock.json
+          rmdir artifact
+
           echo /.cache >> node_modules/.gitignore
           git --work-tree node_modules add --all
-          git --work-tree node_modules commit --quiet -F - < artifact/npm-install.log
-          rm artifact/npm-install.log
+          git --work-tree node_modules commit --quiet -F - < commit-message
           rm -rf node_modules
-          rmdir artifact
 
       - name: Configure cockpituous credentials
         run: echo 'https://token:${{ secrets.COCKPITUOUS_TOKEN }}@github.com' > ~/.git-credentials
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
https://github.com/allisonkarlitskaya/cockpit/pull/7 for a demo

allisonkarlitskaya:
There's a major problem with the new workflows: the github-actions user seems not to be able to modify our PRs:
 - in the case of `node_modules`, the comment at the end can't be posted
 - reposchutz can't clear the .github-changes label

.github/workflows/node_modules-publish.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     timeout-minutes: 5
     env:
       GIT_DIR: git-dir.git
+      HEAD_SHA: ${{ github.event.workflow_run.pull_requests[0].head.sha }}
```
martinpitt, line 14: Ugh, I've never seen this to be a list. `github.event.workflow_run.head_sha` seemed quite fine? But then again, maybe `pull_requests[]` is just what I was looking for for *reliably* getting the original information of the triggering `pull_request`. This would make a lot of things simpler, or even possible again (see https://github.com/cockpit-project/cockpit-machines/pull/143). So I'll definitively look at this again.

.github/workflows/node_modules-publish.yml
```diff
@@ -11,6 +11,8 @@ jobs:
...
+      BASE_SHA: ${{ github.event.workflow_run.pull_requests[0].base.sha }}
```
martinpitt, line 15: `github.event.workflow_run.base_sha` isn't a thing then?

.github/workflows/node_modules-publish.yml
```diff
@@ -65,14 +66,19 @@ jobs:
       - name: Stage git commit
         run: |
           set -ux
+          cat - artifact/npm-install.log > commit-message <<EOF
+          PR: $(sed 's@//api.@//@; s@/repos/@/@; s@/pulls/@/pull/@;' <<< "${PULL_URL}")
```
martinpitt, line 70: Can you prefix this with `#` to make it clickable when viewing in github?

What's with the double `//`? If this is supposed to just show the number, why not just take the `.number` property instead of `.url`, or use `basename` ?

martinpitt (changes requested):
Seems by and large fine, but blocked on sorting out the bot permissions. Some nitpicking included. Thanks!

.github/workflows/node_modules-publish.yml
allisonkarlitskaya: This sed is about turning the https://api... URL into the normal human HTML URL.

.github/workflows/node_modules-publish.yml
allisonkarlitskaya: No.  The head sha gets promoted but the base sha doesn't, unfortunately.  And then I use the head sha for consistency.  Although I'm starting to wonder if that gives us the correct result when people push out-of-date branches.  I'm going to need to do a bit more research here unfortunately :(

allisonkarlitskaya: Result of research is here:

https://gist.github.com/allisonkarlitskaya/32a48ef569d58d43c44c8d0bbee0e144

The long story short is that we don't have access to the PR base SHA from `workflow_run` events.  We can make up for it by using `github.sha` which is the current state of the branch, which is close enough when combined with the tricks I'm using in #15854.

Meanwhile, my approach here is also bogus: for some reason, the `pull_requests[]` array is only populated for pull requests opened from a branch of the same repository, which is kinda useless for us.

It also explains why no comments were being posted: we iterate over an empty array.

I think I'm going to have to send the PR number along with the artifacts from the build workflow.  We can be sure that workflow hasn't been modified by using the `git log` tricks.

This stuff sure isn't easy to get right...

allisonkarlitskaya:
→ #15880

Issue #15830, opened 2021-05-17
storage: enhance SHA256 Tang fingerprints modal


garrett:
Follow-up enhancement issue to #15826.

As stated in https://github.com/cockpit-project/cockpit/pull/15826#issuecomment-839964280:

> We might want to make the modal dialog wider to display the string all in one line. 
> 
> The string should probably be in a (relatively new) clipboard copy widget, with the assumption that someone might want to copy to verify.
> 
> It could be one of two variants:
> https://www.patternfly.org/v4/components/clipboard-copy#read-only (if this fits or if it allows wrapping)
> or
> https://www.patternfly.org/v4/components/clipboard-copy#inline-compact-in-sentence (long copy string in block, specifically)
> 
> I think it's still useful to show the string in a larger monospaced font to make manual comparisons easier, so we probably still want to do some custom styling even if we do use this other widget.

So, here it is in a TODO list format:

- [ ] Change modal size (slightly wider?)
- [ ] Use clipboard widget (and keep larger, monospaced formatting)

PR #15844, opened 2021-05-19
storage: Tang key hash improvements

mvollmer:
tang-show-keys is now widely enough available to use it in our tests instead of jose.

However, different versions of tang-show-keys use different hashing algorithms, so we need to account for all possibilities in our dialog.

This PR also removes from the dialog the alternative command that people might use if they don't have tang-show-keys. The assumption again is that everybody has it now, and having this in the dialog just adds to the perceived complexity of the task.

### DIFF
--- a/test/verify/check-storage-luks
+++ b/test/verify/check-storage-luks
@@ -144,14 +144,6 @@ class TestStorageLuks(StorageCase):
         m = self.machine
         b = self.browser
 
-        # We generate the keys and cache explicitly before starting
-        # the socket.  This allows us to control their names, which
-        # helps later on.
-
-        m.execute("rm -rf /var/db/tang/*")
-        m.execute("jose jwk gen -i '{\"alg\":\"ES512\"}' >/var/db/tang/sig1.jwk")
-        m.execute("jose jwk gen -i '{\"alg\":\"ECMR\"}' -o /var/db/tang/enc1.jwk")
-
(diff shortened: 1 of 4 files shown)


mvollmer:
This related to #15830.  Instead of making the dialog wider, I have made the thumbprint text smaller... I'll check out the clipboard stuff laterish.

martinpitt (approved):
Thanks! I did not test this visually, but code changes make sense.

test/verify/check-storage-luks
```diff
@@ -193,8 +185,8 @@ class TestStorageLuks(StorageCase):
         self.dialog_set_val("tang_url", "127.0.0.1")
         self.dialog_set_val("passphrase", "vainu-reku-toma-rolle-kaja")
         self.dialog_apply()
-        b.wait_in_text("#dialog", "Make sure the key hash from the Tang server matches:")
-        b.wait_in_text("#dialog", m.execute("jose jwk thp -a S256 -i /var/db/tang/sig1.jwk").strip())
+        b.wait_in_text("#dialog", "Make sure the key hash from the Tang server matches")
+        b.wait_in_text("#dialog", m.execute("tang-show-keys").strip())
```
allisonkarlitskaya, line 189: :heart:

allisonkarlitskaya (approved):

garrett:
What's up with the scrollbar and weird border?

PR #15847, opened 2021-05-20
packit: Enable Rawhide

martinpitt:
Even though this is prone to instability, we'd still rather like to learn about OS regressions earlier than later.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -17,4 +17,5 @@ jobs:
       targets:
       - fedora-33
       - fedora-34
+      - fedora-rawhide
       - centos-stream-8-x86_64


martinpitt:
Hmm, "interesting" -- [centos 8 run](http://artifacts.dev.testing-farm.io/a844b21d-ff1e-42c2-8161-3d3fc187cc28/) now did get chromium 90, but still fails on this crash. In my Fedora 33 toolbox this works now. Perhaps different build settings or some underlying library? Anyway, this is not so important -- I'll wait for the Fedora results, and then just keep the "enable rawhide" part.

marusak (approved):
worked, nice!

PR #15848, opened 2021-05-20
users: Limit maximum length of a new password

### LINKED ISSUE #15775: No error for invalid new password
Cockpit version: 243
OS: ubuntu 20.4.2
Page: set password

I set my password to a 3363 char length password with multi-lines,
no error showed 
I'm logged out and I can't log in,

Steps to reproduce

1. open the profile page 
2. copy multiline 3363 char string 
3. paste the string 
4. setting the password

marusak:
Fixes #15775

passwordtoolong


test/verify/check-users
```diff
@@ -121,7 +121,17 @@ class TestAccounts(MachineCase):
         b.click('#accounts-create-dialog button.apply')
         b.wait_in_text("#accounts-create-dialog .pf-c-form__helper-text.pf-m-error", "The passwords do not match")
 
+        # too long password
+        long_password = "2a02-x!h4a" * 30
+        b.set_input_text('#accounts-create-password-pw1', long_password)
+        b.set_input_text('#accounts-create-password-pw2', long_password)
+        print(long_password)
+        sit()
```
martinpitt, line 129: debugging leftover

pkg/users/password-dialogs.js
```diff
@@ -207,6 +207,9 @@ export function set_password_dialog(account, current_user) {
         if (state.password != state.password_confirm)
             errors.password_confirm = _("The passwords do not match");
 
+        if (state.password.length > 256)
+            errors.password = _("Password is longer than 256 characters");
+
```
martinpitt, line 212: This gets overwritten by a bad quality check below. Also, given that the test failure somehow lands on the overview page, this seems to be wired up wrongly? That error should appear inside the dialog, right?

martinpitt (changes requested):

pkg/users/password-dialogs.js
marusak: The fact it would get overwritten is fine in my option.
This is very corner case check and also I consider quality check to be of higher quality.

And I just realized that we have two separate `validate` methods, one for changing password and one for validating a new password when creating a new account.

martinpitt (approved):
Nice, thanks!

PR #15850, opened 2021-05-20
storage: Remove two obsolete clevis workarounds

mvollmer:


mvollmer:
> Failed almost everywhere.. still applies after all?

One bit is still needed I think: `pty: true`. I guess clevis-luks-unbind calls cryptsetup in the same way as our workaround did, which for some reason requires a tty on stdin. *shrug*

mvollmer:
Oi, missed the "-f" flag.

martinpitt (approved):
Nice, thanks! Tests looked good, just retried debian-stable for some usual suspects.

PR #15854, opened 2021-05-21
reposchutz: improve change detection logic

allisonkarlitskaya:
The workflow has a fairly simple strategy at present: if there are
changes to the `.github` directory (as per `git diff`) then fail.  That
has two problems:

 - the changes might have come from an unrelated PR that got merged in
   the meantime; and

 - the PR can "get away with" individual commits that make changes, as
   long as there are future commits that revert those changes.
   Depending on the merge/rebase strategy used when landing, thre are
   some creative way to sneak changes through anyway.

Replace the previous logic with a new one: list all commits which are:

 - reachable from the PR;

 - not already on the target branch; and

 - modify `.github`

and fail if the result is non-empty.

Factor out the context expressions for the relevant SHAs into
environment variables to improve readability.

Finally, since our new approach depends on more than just looking at the
two head commits, a shallow fetch isn't sufficient anymore.  `git fetch`
has a (somewhat new) relevant feature `--shallow-exclude=` which would
be very helpful to us here, but the GitHub server seems not to support
it, so just fetch a full history for the time being.

### DIFF
--- a/.github/workflows/reposchutz.yml
+++ b/.github/workflows/reposchutz.yml
@@ -10,6 +10,9 @@ jobs:
     timeout-minutes: 5
     env:
       GIT_DIR: git-dir.git
+      HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+      BASE_SHA: ${{ github.event.pull_request.base.sha }}
+
     steps:
       - name: Clear .github-changes label
         if: ${{ !endsWith(github.event.action, 'labeled') }}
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
An example: https://github.com/allisoninc/playground/pull/1

martinpitt (approved):
Nice, thanks for the analysis! Now of course the old reposchutz is lamenting as well, so I'll add the label now. This should automatically re-run and pacify it, right?

allisonkarlitskaya:
> Nice, thanks for the analysis! Now of course the old reposchutz is lamenting as well, so I'll add the label now. This should automatically re-run and pacify it, right?

In this case, the "lament" is correct: we're introducing changes to `.github` and we should absolutely be setting the label.

PR #15857, opened 2021-05-24
storage: Reload systemd before mounting after formatting

mvollmer:
Otherwise systemd might have outdated UUID / mount point mappings and
might unmount our new filesystem right away again.

Here are the details.  Start with this in /etc/fstab:

    UUID=<uuid1> /run/foo ...

This causes systemd to associate uuid1 with /run/foo.  Now uuid1 is
unmounted, the block device is formatted, and a new fstab is written
with the new uuid2 but with the same mount point:

    UUID=<uuid2> /run/foo ...

At this point, systemd sees that uuid1 has disappeared and will make
sure that /run/foo is not mounted, because it still associates
/run/foo with uuid1.  At the same time, Cockpit will mount the new
filesystem at /run/foo.  This is a race and sometimes Cockpit does the
mount, and then systemd does its cleanup and unmounts.  (If systemd
does the cleanup first, it will find /run/foo already unmounted and
not do anything.)

The race is avoided by reloading systemd before having Cockpit do the
mount.  This way, systemd will forget the uuid1 -> /run/foo
association.

- [ ] #15859

### DIFF
--- a/pkg/storaged/format-dialog.jsx
+++ b/pkg/storaged/format-dialog.jsx
@@ -356,8 +356,8 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
                 return utils.teardown_active_usage(client, usage)
                         .then(utils.reload_systemd)
                         .then(format)
-                        .then(maybe_mount)
-                        .then(utils.reload_systemd);
+                        .then(utils.reload_systemd)
+                        .then(maybe_mount);
             }
         }
     });


martinpitt:
This [fsck failure](https://logs.cockpit-project.org/logs/pull-15857-20210524-105153-3820eb83-fedora-34/log.html#200-2) looks familiar, retrying.

pkg/storaged/format-dialog.jsx
```diff
@@ -356,8 +356,8 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
                 return utils.teardown_active_usage(client, usage)
                         .then(utils.reload_systemd)
                         .then(format)
-                        .then(maybe_mount)
-                        .then(utils.reload_systemd);
+                        .then(utils.reload_systemd)
+                        .then(maybe_mount);
```
martinpitt, line 360: Is this synchronous enough? I was fairly sure that Reload() was async. You may need to wait for the `Reloading(False)` signal here?

martinpitt (approved):
Makes sense in general, I just wonder if it's strong enough. Can you reproduce this failure in some semi-reliable way?

pkg/storaged/format-dialog.jsx
mvollmer: It certainly slows down everything enormously... I'll check to make sure.

mvollmer: The "method_return" for the call to Reload comes after the signal that informs about "Reloading=false":

```
‣ Type=method_call  Endian=l  Flags=4  Version=1 Cookie=2  Timestamp="Mon 2021-05-24 12:02:28.097747 UTC"
  Sender=:1.40  Destination=org.freedesktop.systemd1  Path=/org/freedesktop/systemd1  Interface=org.freedesktop.systemd1.Manager  Member=Reload
  UniqueName=:1.40
  MESSAGE "" {
  };
[...]
‣ Type=signal  Endian=l  Flags=1  Version=1 Cookie=7132  Timestamp="Mon 2021-05-24 12:02:28.650280 UTC"
  Sender=:1.2  Path=/org/freedesktop/systemd1  Interface=org.freedesktop.systemd1.Manager  Member=Reloading
  UniqueName=:1.2
  MESSAGE "b" {
          BOOLEAN false;
  };

‣ Type=method_return  Endian=l  Flags=1  Version=1 Cookie=7133  ReplyCookie=2  Timestamp="Mon 2021-05-24 12:02:28.650307 UTC"
  Sender=:1.2  Destination=:1.40
  UniqueName=:1.2
  MESSAGE "" {
  };
```

That's enough proof for me.

mvollmer:
> Can you reproduce this failure in some semi-reliable way?

It started happening reliably when I was extending TestStorageLuks.testNoFsys in #13192.

mvollmer:
> This [fsck failure](https://logs.cockpit-project.org/logs/pull-15857-20210524-105153-3820eb83-fedora-34/log.html#200-2) looks familiar, retrying.

Seems to happen a lot here. I see if I can figure it out...

mvollmer:
> > This [fsck failure](https://logs.cockpit-project.org/logs/pull-15857-20210524-105153-3820eb83-fedora-34/log.html#200-2) looks familiar, retrying.
> 
> Seems to happen a lot here. I see if I can figure it out...

This seems to be another version of https://github.com/cockpit-project/bots/issues/1739.

During this test, Cockpit first grows the LUKS container (which due to the bug results in the fsys being unmounted), and then grows the fsys.  Growing a mounted fsys needs different code from growing a unmounted fsys, and Cockpit uses the wrong code here because it hasn't yet figured out that the fsys got unmounted. Normally, the timing seems to be so that Cockpit already knows that the fsys has been unmounted when starting to grow it. Hurray for asynchronous everything!

I think it would make sense for Cockpit to be a bit more robust here, and maybe wait for the effects of the luks resizing to have rippled out. Or probably more realistically, do a synchronous check whether or/not the fsys is mounted instead of believing its asynchronous version of reality.

mvollmer:
> I think it would make sense for Cockpit to be a bit more robust here

#15857

pkg/storaged/format-dialog.jsx
martinpitt: Thanks for checking! The "slows down enormously" is both good and bad, I figure :grin: 

Good to know that Reload() is synchronous, that's indeed much more helpful for API users.

mvollmer: > The "slows down enormously" is both good and bad, I figure.

It's not so bad during formatting (although we should say in the dialog that it is happeneing, "Reloading system configuration", maybe).  It is a bit annoying during unmounting, which would otherwise be instant.  Also, it doesn't help that we need to do it twice...

PR #15859, opened 2021-05-24
storage: Synchronously check for mountedness during resizing

mvollmer:


mvollmer:
The test will still fail because the unmount triggered by resizing the LUKS container is still unexpected (and wrong). However, this change here will make it so that the user is left with a fully resized (but unexpectedly unmounted filesystem), instead of a half resized one.

mvollmer:
Grr, is findmnt not generally available?

mvollmer:
> Grr, is findmnt not generally available?

No, the code was checking the wrong block device for mountedness.

martinpitt (approved):
Nasty hack, but :man_shrugging: what can you do.. it doesn't affect the on-disk end result, so we can live with this well. Thanks for tracking this down!

PR #15861, opened 2021-05-25
packagekit: Translate automatic updates frequency

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=1964177

----

"every day" still had some old translations, so this works now:

image

The others don't yet have msgids, and thus are not translated yet, but at least can be with this fix.


marusak (approved):
Thanks!

PR #15864, opened 2021-05-26
tls: Fix quoting of locale check

martinpitt:
Spotted by coverity: "warning[SC2046]: Quote this to prevent word splitting."


marusak (approved):
Thanks!

PR #15869, opened 2021-05-26
lib: Don't reload systemd after enabling/disabling a service

martinpitt:
There is no good reason for doing that -- systemd will automatically
update the unit's state with these API calls. This is causing lockups of
services when e.g. pmlogger.services gets started and enabled at the
same time, as the parallel reload messes up the "activating"
transaction.

This was introduced in the original code in commit 6cbbcec8c2 without a
particular justification, so probably it was just working around an
early systemd bug back then?

 - [x] wait for after the release

### DIFF
--- a/pkg/lib/service.js
+++ b/pkg/lib/service.js
@@ -249,6 +249,8 @@ export function proxy(name, kind) {
     systemd_manager.addEventListener("JobNew", on_job_new_removed_refresh);
     systemd_manager.addEventListener("JobRemoved", on_job_new_removed_refresh);
 
+    systemd_manager.addEventListener("UnitFilesChanged", refresh);
+
     function wait(callback) {
         wait_callbacks.promise.then(callback);
     }
--- a/pkg/lib/service.js
+++ b/pkg/lib/service.js
@@ -292,25 +292,6 @@ export function proxy(name, kind) {
         return dfd.promise();
     }
 
-    function call_manager_with_reload(method, args) {
-        return call_manager(method, args).then(function () {
-            var dfd = cockpit.defer();
-            call_manager("Reload", [])
-                    .done(function () { dfd.resolve() })
-                    .fail(function (error) {
-                    // HACK: https://bugzilla.redhat.com/show_bug.cgi?id=1560549
-                    // some systemd versions disconnect too fast from the bus
-                        if (error.name === "org.freedesktop.DBus.Error.NoReply") {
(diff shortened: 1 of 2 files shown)


martinpitt:
I fixed reacting to systemd API enable/disable. The [tuned failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15869-20210526-115841-b26f3bb9-fedora-34/log.html#268-2) still happens, will debug that tomorrow. But re-running CI to confirm that everything else is fine now.

mvollmer:
> This was introduced in the original code in commit 6cbbcec8c2 without a
> particular justification, so probably it was just working around an
> early systemd bug back then?

No, I remember explicitly checking this, and iirc this was documented behavior: calling EnableUnitFiles would just modify the filesystem and not the internal state of systemd. (I can't find that documentation anymore right now.) At the time the code for "systemctl enable" also did a explicit Reload. But now `systemctl` doesn't seem to call EnableUnitFiles at all anymore... what's going on there?

So, yeah, I guess calling Reload is no longer necessary when calling EnableUnitFiles. Nice!

(I'll try to find the relevant change in systemd.)

martinpitt:
Meh, the tuned failure is a dealbreaker -- it seems calling systemd's `EnableUnitFiles()` updates/reloads the unit state *sometimes*, but not in this case. That's probably the reason why `systemctl enable` does a full daemon reload, even though conceptually for enable/disable this is totally unnecessary. *sigh*, so back to debugging why pmlogger.service start hangs with a parallel daemon-reload..

mvollmer:
> That's probably the reason why systemctl enable does a full daemon reload

Yes, it still does, I got confused by the install_client_side branch of it.

> it seems calling systemd's EnableUnitFiles() updates/reloads the unit state sometimes,

My systemd fu is rusty, but the "sometimes" might be explained by loaded vs unloaded units.  If a unit is already loaded when you enable it, systemd will not pick up changes to symlinks, but if it is not loaded, systemd will pick up the current syslinks when it is loaded.

martinpitt:
@mvollmer : Right, that seems to be it. Unfortunately there is no flag or API to tell systemd to reload that one unit already (or, it could just do this by default..) Enablement links don't change the runtime state or transactions at all, so this would be harmless.

Either way, I found an alternative solution for pmlogger, so let's not waste more time on this :grin: Thanks for looking!

PR #15870, opened 2021-05-26
storage: Fix scrollbar issue; change pre display

garrett:
Fixes the scrollbar issue introduced in https://github.com/cockpit-project/cockpit/pull/15844

PR #15871, opened 2021-05-26
storage: Use semantic HTML in verify key modal

garrett:
Follow-up to #15844, to fix the HTML.

Would've been in #15870, but I wanted the previous fix to be implemented quickly. This changes a lot more.

### DIFF
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -296,14 +296,15 @@ td.storage-action {
     margin-right: 1em;
 }
 
+.sigkey-heading {
+    color: var(--pf-global--Color--dark-100);
+    font-size: var(--pf-global--FontSize--sm);
+    margin: var(--pf-global--spacer--md) 0 0;
+}
+
 .sigkey-hash {
     font-family: monospace;
     font-size: 140%;
-}
-
-.inline-pre {
-    display: inline-block;
-    padding: 4px;
 }
 
 .pf-c-modal-box h3 {
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -21,7 +21,8 @@ import cockpit from "cockpit";
 import React from "react";
 
 import {
-    Card, CardBody, CardTitle, CardHeader, CardActions, Checkbox,
+    Card, CardBody, CardTitle, CardHeader, CardActions,
+    Checkbox, ClipboardCopy,
     Form, FormGroup,
     DataListItem, DataListItemRow, DataListItemCells, DataListCell, DataList,
     Text, TextVariants, TextInput as TextInputPF, Stack,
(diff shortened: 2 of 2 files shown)


garrett:
It doesn't look _so much_ different, but anyone using a screen reader would care.

Screenshot from 2021-05-26 16-52-00

allisonkarlitskaya (commented):
Can you add a copy widget to those while you're at it?

garrett:
> Can you add a copy widget to those while you're at it?

I'd love that.

Specifically inline compact code: https://www.patternfly.org/v4/components/clipboard-copy#inline-compact-code

mvollmer (approved):

mvollmer:
Thanks!

KKoukiou (commented):
This PR also add some style fixes which are not mentioned in the commit message. Especially the 'pre' customization I would like to check before we merge this.

pkg/storaged/storage.scss
```diff
@@ -400,14 +400,22 @@ td.storage-action {
...
+    margin: var(--pf-global--spacer--md) 0 0;
+}
+
 .sigkey-hash {
     font-family: monospace;
     font-size: 140%;
 }
 
 .inline-pre {
+    background: var(--pf-global--BorderColor--300);
```
KKoukiou, line 415: OK Patternfly does not style ` <pre> ` with grey background etc.
Instead for code blocks or text areas they provide their own components - see https://www.patternfly.org/v4/components/code-block 

As this is overriding default PF styles I think this is not correct do here - as we 're trying to move away from custom stylesheets.

KKoukiou (changes requested):

garrett:
I've changed the pre to an inline code copy from PF @ 8895962.

However, it looks like this:

Screenshot from 2021-05-27 18-35-33

And it should be inline, with code, and have a grey background like this:

Screenshot_2021-05-27 PatternFly 4 • Clipboard copy

From https://www.patternfly.org/v4/components/clipboard-copy/#inline-compact-code

Do we need to update PF to get the new variants perhaps?

KKoukiou:
For the inline copy clipbord component we need https://github.com/patternfly/patternfly-react/pull/5629
which is included in @patternfly/react-core@4.108.0
We currently have lower version than this so it's expected that you can't use it with this patternfly version @garrett .

garrett:
Added a blocked tag, as we're blocked on a PF upgrade. Hopefully we can upgrade soon and get this merged in shortly after.

KKoukiou (approved):

martinpitt:
Whoops, this slipped through the cracks, sorry! The PF update landed a month ago, and this was already approved. I pushed a rebase and I'll see to get this into today's release.

martinpitt:
I tested the rebase. The dialog looks fine to me, and copy&paste works as expected.

image

martinpitt (approved):

mvollmer (approved):

PR #15874, opened 2021-05-27
services: network: storage: Rename ‘All logs’ button

### LINKED ISSUE #15799: services: network: storage: Rename ‘All logs’ button
Actions should contain verbs (Show would imply that they’re seen inline.)

marusak:
Fixes #15799

@garrett I ma not sure about the right wording, do you a better idea?


garrett:
We talked about "View all logs" in the call.

garrett (changes requested):
We talked about "View all logs" in the call and how it should either have proper alignment or switched to a secondary button style (which would fix the alignment too).

marusak:
Updated:
Screenshot from 2021-05-27 20-23-25alllogsexist

When there are no logs, the button is hidden:
Screenshot from 2021-05-27 20-22-56nobutton

And it is still link on metrics page:
Screenshot from 2021-05-27 20-26-38alllogs2

(as I prefer it to secondary button in this case):
Screenshot from 2021-05-27 20-24-53event1

garrett (approved):
> And it is still link on metrics page
> (as I prefer it to secondary button in this case)

Yeah, me too. It's too heavyweight otherwise.

LGTM! :+1:

PR #15875, opened 2021-05-27
sosreport: Provide form to control sos command options

TurboTurtle:
Adds a form to allow users to set several options commonly used when
generating an sos report.

Includes updating the called command to `sos` when generating the command
to run, in line with the 'binary' change for sos-4.0 and later.

Additionally, set the sosreport image to the project's logo.

Signed-off-by: Jake Hunsaker <jhunsake@redhat.com>


marusak:
Thank you for this contribution!
I'll address at first visuals.
1. I am not super sure about this huge logo:
Screenshot from 2021-05-31 09-42-55hugelogo
currently looks like this:
Screenshot from 2021-05-31 09-45-30oldlogo

2. I though that this page was already rewritten into PF4 and react, but seems it is not the case - thus the dialog looks somewhat different to other pages. I propose not to be bothered by this and we can migrate it after this lands. I am happy to do it. But regarding the dialog:
Screenshot from 2021-05-31 09-43-46plugins

All the strings should be sentence case (Only the first letter capital, also `ALL` should be just `all`).

3. It now has somewhat weird workflow - open one dialog, set options, click on submit button, it closes dialog, opens different dialog... I think it should be one dialog. You select options, then submit it. After that at the bottom a progress bar appears, all controls are disabled.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,124 @@
...
+    <div hidden id="sos-options-menu">
+        <div class="pf-c-backdrop">
+            <div class="pf-l-bullseye">
+                <div class="pf-c-modal-box pf-m-md pf-m-align-top" role="dialog" aria-modal="true">
+                    <button class="pf-c-button pf-m-plain" type="button" aria-label="Close" id="sos-options-close-button">
+                        <i class="fa fa-times" aria-hidden="true"></i>
+                    </button>
+                    <header class="pf-c-modal-box__header">
+                        <h1 class="pf-c-modal-box__title" translate="yes">Set SoS Report Options</h4>
+                        <small>All settings are optional, if in doubt leave default values and click 'Run Report'</small>
```
marusak, line 56: I don't think we need to specify this. If we do, then maybe it should be behind (?) or even in documentation. I think it is rather clear that those options are optional.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,124 @@
...
+                    </header>
+                    <div class="pf-c-modal-box__body">
+                        <div id="sos-settings-body">
+                            <form class="ct-form" id="sos-opts-form">
+                                <label class="control-label" for="sos-label">Report Label</label>
+                                <input id="sos-label" class="form-control" type="text" title="Append a label/name to the final report archive name. No spaces.">
```
marusak, line 62: all the labels and titles need to be translated. You can do `translate="title"` for titles and `translate="yes"` for labels.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,124 @@
...
+                                </label>
+
+                                <label class="control-label pf-c-switch" for="sos-verify">Verify Packages &nbsp;&nbsp;
+                                    <span class="pficon pficon-help" title="Have the package manager verify all installed packages"></span>
+                                </label>
+                                <label class="checkbox-inline">
+                                    <input id="sos-verify" type="checkbox">
+                                </label>
+
+                                <label class="control-label" for="sos-verbose">Verbose Logging &nbsp;&nbsp;
```
marusak, line 117: those ` &nbsp;&nbsp;` will definitely get broken with translations. Please use css for adding space between components.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,124 @@
...
+                                <label class="control-label" for="sos-skip-plugins">Skip These Plugins</label>
+                                <input id="sos-skip-plugins" class="form-control" title="List of plugin names to skip, comma delimited">
+
+                                <label class="control-label" for="sos-only-plugins">Only Run These Plugins</label>
+                                <input id="sos-only-plugins" class="form-control" title="List of plugin names to ONLY run, comma delimited">
+
+                                <label class="control-label" for="sos-plugin-option">Plugin Options</label>
+                                <input id="sos-plugin-option" class="form-control" title="Set plugin options, comma delimited list. Options take the form <plugin>.<option>=<value>">
+
+                                <label class="control-label" for="sos-allow-system-changes">Allow System Changes &nbsp;&nbsp;
```
marusak, line 89: those `&nbsp;&nbsp;` will definitely get broken with translations. Please use css for adding space between components.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,124 @@
...
+                                </select>
+
+                                <label class="control-label" for="sos-skip-plugins">Skip These Plugins</label>
+                                <input id="sos-skip-plugins" class="form-control" title="List of plugin names to skip, comma delimited">
+
+                                <label class="control-label" for="sos-only-plugins">Only Run These Plugins</label>
+                                <input id="sos-only-plugins" class="form-control" title="List of plugin names to ONLY run, comma delimited">
+
+                                <label class="control-label" for="sos-plugin-option">Plugin Options</label>
+                                <input id="sos-plugin-option" class="form-control" title="Set plugin options, comma delimited list. Options take the form <plugin>.<option>=<value>">
```
marusak, line 87: I don't see this helper text in the UI. Nor for 'Only run these plugins' and for 'Skip these plugins'.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,124 @@
...
+                                    <option value=2>2</option>
+                                    <option value=3>3</option>
+                                    <option value=4 selected>4</option>
+                                </select>
+
+                                <label class="control-label" for="sos-skip-plugins">Skip These Plugins</label>
+                                <input id="sos-skip-plugins" class="form-control" title="List of plugin names to skip, comma delimited">
+
+                                <label class="control-label" for="sos-only-plugins">Only Run These Plugins</label>
+                                <input id="sos-only-plugins" class="form-control" title="List of plugin names to ONLY run, comma delimited">
```
marusak, line 84: These two are almost unusable. Can we get list of all plugins and then show it in the UI?
Possibly [multi select](https://www.patternfly.org/v4/components/select#multiple) with pre-selected plugins that are auto enabled.

marusak: Actually, only one at a time makes sense, doesn't it?
If you pick some "to only run" then "skip these" doesn't make a lot of sense.
So only having one field with the auto enabled ones and making it easy to deselect all to make 'only run these' easy to do.

marusak (changes requested):

pkg/sosreport/index.html
martinpitt: You need to mark all human readable strings as translatable. For tag contents, `translate="yes"`, for the input titles `translate="title"`.

(And ugh, yes, this really ought to be React..)

pkg/sosreport/index.html
TurboTurtle: So right now there are over 300 plugins in sos. Even on super sparse distributions you're looking at ~50 default enabled plugins, so I think any form of select here will be rather clunky/unwieldy. The form inputs match the CLI option syntax - a comma delimited string.

At first glance yes, only having "skip" or "only" available makes sense, however there are use cases where both can be used at the same time. For example, imagine that `/etc/sos/sos.conf` specifies to always run the `foo` profile, which sets it's own defined set of plugins. However, the user knows that they want to modify the plugin list slightly. If they use `skip` they can then disable plugins that were otherwise enabled by the `foo` profile, and similarly if they use the `only` field they can add additional plugins on top of what the profile enables.

pkg/sosreport/index.html
TurboTurtle: Ack to both

pkg/sosreport/index.html
TurboTurtle: Ack, however I'm noticing that for the labels that include the "?" <span>, using `translate="yes"` causes those spans to not render. Setting `translate="title"` however (in either the label or the span) does, but I'm not sure if that's sufficient?

pkg/sosreport/index.html
martinpitt: I'm not convinced about this either. This really isn't something where a human writes down five plugins with precise names without error. If you need such fine-grained control, IMHO you are better off with actually *using* the command line -- this UI isn't meant to be (basically) a GUI widget for typing a command line.

IMHO a good middle ground would be to only list available profiles, such as `security`, 'system`, `hardware`. There's 22 of them, which is reasonable. WDYT?

TurboTurtle: A profile dropdown would be handy, and could be easily populated by scraping `sos report --list-profiles` (if that's acceptable? Or would this need to be a curated select?).

My concern with not having the skip option exposed at all is more historical from generating a report from CLI. There's a long history of certain plugins (e.g. 'logs') exploding archive size due to some other issue, or when combined with other options like `--all-logs`. E.G. the user is thinking "I need all logs for X, Y, and Z plugins, but I want to skip the logs plugin because I don't need to entire system journal for this system". We've made several safeguards against hung plugins in the past few releases, but combining `--all-logs` with "give me the system journal" is still something that can surprise users.

That being said, if this should just be targeting the very simple "give me a default report" use case, and anything else should defer to using the CLI directly, then that makes sense and I could drop off a number of the other options in this PR too. Just have the basics - label, case number, default log size, and the `--clean` toggle.

Does that sound better than providing any of the plugin/profile/preset controls?

martinpitt: I think showing `sos report --list-profiles` would be acceptable (is there some summary which the page could parse and show?) Note that by default programs are being called with the selected browser locale, so if list-profiles shows i18n'ed explanation strings, they'd even be shown in the UI. It is not maintainable/desirable to have such a "profile name" → "human description" mapping in the cockpit plugin. But even with just the English short names these should already be helpful.

As to the desired use cases, you are a much better person to tell than us Cockpit developers. We are mostly trying to give some guidance/guard rails as to usability and such, so please don't interpret it as "we don't want this" -- we do! FWIW, if you think it's useful to discuss this with high bandwidth, let's schedule a call.

In its current form I certainly find the dialog overwhelming (it would certainly be the biggest dialog in Cockpit by far), so it does make sense to think about some balance between flexibility and comprehension. Clearly the current page is too simple -- you made this PR for a reason, and I also saw a similar request in bugzilla in the past.

My personal gut feeling is that 22 pre-defined profiles with reasonably understandable names are about the right granularity, whereas free-form "computer text" input for 300+ plugins is too much. You make a convincing case for an option to include (or not) the full journal (even though I think that's by far too much -- it's usually many gigabytes going back a year or even more!), but as you say that could have a sensible cut-off point (like "500 MB" or "1 week"). Report label and case number also are very sensible.

OTOH options like "allow system changes", "verify packages" or "skip post-processing" sound obscure enough to me as an "outsider" that they should have reasonable defaults. "threads" is also over-configuration -- `nproc` is a fine default there for an interactive UI where you are waiting for the result.

So maybe let's play this "you are the expert" vs. "we are the trial subjects and give design feedback" for a few more rounds, until we are both happy?

Thank you!

TurboTurtle: > FWIW, if you think it's useful to discuss this with high bandwidth, let's schedule a call.

That'd probably be a great step. I'll shoot you an email to get something arranged.

TurboTurtle:
Sorry for the long delay on this - got caught up on other stuff and it slipped my mind for a while.

Anyways, I've updated the PR to drop most of the added form options. Now reduced to report label, case number, selecting if the archive should be encrypted, and selecting if it should be obfuscated. The latter two are handled internally by sos.

I opted to not include profile selection, as after review we (the sos project) don't really see all that much use of that feature. I also did not include a passthru for the native "upload to a defined vendor" functionality.

I've still got some other ideas for this plugin, but I think those are better left until a react rewrite as they are mostly UI-centric ideas - e.g. listing existing archives on the system and the ability to retrieve/delete those from cockpit, ability to check for and update the local sos install, etc...

martinpitt:
The react rewrite landed now, should be simpler to add things to the dialog now. This also already included the `sosreport` → `sos` command change. I'll also clean up the `archive_files` bits (for which you did the .md5 → .sha256 change) and update the logo; I'll send a separate PR for that, then we can concentrate on adding the new fields. If you want to work on that, I'll be happy to guide you, otherwise I can send a PR (or force-push to this one).

martinpitt:
>  I'll also clean up the archive_files bits 

Done in PR #16557 

> and update the logo

@TurboTurtle , your PR here [adds a new logo](https://github.com/cockpit-project/cockpit/blob/e3db1cde5ff4506c9b05cb68bc99f87c98ef65d6/pkg/sosreport/sosreport.png). Where does it come from? What's the license of it? It is not anywhere in https://github.com/sosreport/sos . We also need a 64x64 px version for the AppStream metadata; I'm happy to scale it down myself, but (1) perhaps we can find the original SVG for this (your version is very blurry), and (2) we can't put it into our repo without copyright/license information.

If that is not an "official" logo, then perhaps we could also use the "Life Ring" [PatternFly icon](https://patternfly-react.surge.sh/icons/)?

TurboTurtle:
> @TurboTurtle , your PR here [adds a new logo](https://github.com/cockpit-project/cockpit/blob/e3db1cde5ff4506c9b05cb68bc99f87c98ef65d6/pkg/sosreport/sosreport.png). Where does it come from? What's the license of it? It is not anywhere in https://github.com/sosreport/sos . We also need a 64x64 px version for the AppStream metadata; I'm happy to scale it down myself, but (1) perhaps we can find the original SVG for this (your version is very blurry), and (2) we can't put it into our repo without copyright/license information.
> 

It's actually from here - https://openclipart.org/detail/22163/lifebelt, so it's under the Creative Commons Zero 1.0 Public Domain License.


>  then we can concentrate on adding the new fields. If you want to work on that, I'll be happy to guide you, otherwise I can send a PR (or force-push to this one).

I had some other, what I thought would be longer-term, ideas floating around in my head - but it may be best to just knock them out together. Let me get those written down and I'll follow up with you on those to see if they're actually viable

martinpitt:
@TurboTurtle : So that rescue ring is the designated sosreport logo now? If so, I'd rather include the original SVG version (or re-render it at higher res)

garrett:
> I am not super sure about this huge logo:

We don't show logos of anything else within Cockpit. I'm not sure why SoS would need a logo or any branding, aside from its own project website.

pkg/sosreport/index.html
```diff
@@ -28,20 +28,60 @@
...
+                        <div id="sos-settings-body">
+                            <form class="ct-form" id="sos-opts-form">
+                                <label class="control-label" for="sos-label" translate="yes">Report label</label>
+                                <input id="sos-label" class="form-control" type="text" translate="title" title="Append a label/name to the final report archive name. No spaces.">
+
+                                <label class="control-label" for="sos-case-id" translate="yes">Case number</label>
+                                <input id="sos-case-id" class="form-control" translate="title" title="If you have a support case open with a vendor, enter that case id here">
+
+                                <label class="control-label" for="sos-encrypt-pass" translate="yes">Encrypt archive with passphrase</label>
+                                <input id="sos-encrypt-pass" class="form-control" translate="title" title="Encrypt the archive with this passphrase" placeholder="Leave empty to not encrypt">
```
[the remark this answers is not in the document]
garrett: Additionally: This is not an appropriate placeholder.

That text should be "helper text", not a placeholder.

Please consult:
- Helper text guidelines: https://www.patternfly.org/v4/components/form/design-guidelines/#helper-text
- Form React examples (with helper text even in the first form example): https://www.patternfly.org/v4/components/form 
- Placeholder text guidelines (optionally, just to see when placeholder text should be used... and here it shouldn't be): https://www.patternfly.org/v4/components/form/design-guidelines/#helper-text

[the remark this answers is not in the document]
garrett: Also: Helper text still would need to be translated. :wink:

garrett:
The UI for this dialog is odd:



The label plus labeless checkbox is bad for accessibility and usability reasons. These should all be in one or two options groups with a checkbox and then a label. (And the (?) icons should be outside the label, so clicking on them will show a popover, not toggle the checkbox.)

None of the inputs give any hints about what contents go inside. Is it a number, a string, or something else? Also, how many of these items are optional and which are needed? Are they all really necessary? Can we show the necessary fields and hide others under an advanced options [expandable section](https://www.patternfly.org/v4/components/expandable-section)?

After you click a "Generate report" ("Run" doesn't really make sense here), the dialog should disappear and the progress should be shown in-page with a progress bar (or at least a spinner, if indeterminate or if it takes a short time).

---

Moreover:

I'd suggest rewriting the page in such a way where it lists past reports and have a way to create new ones. You can download the latest report (or older reports).

If the latest report is the only one available, then show details about it, allow for downloading, and have a way to create a new one.

Don't hide away everything in modal dialogs. It's fine to have a modal to create a report when more information is needed, but that is all. Everything else should happen in-page.

But depending on the scope of this PR, this part can be deferred to a follow-up.

garrett:
Here's a first pass at an idea of how this could work... Everything's subject to change and this probably needs more design iterations. That said, it's a start, so we can discuss further how this could work.

## Default state of the page, as a list

SOS report list

The menu could have items like:

- Download
- Delete report

Perhaps we could have a download action be the default action on the right?

The main reports section could have:

- Download all reports
- Delete all reports

## Creating a new report

Run report modal

The dialog could be something like this.

I don't know what a case number is or what would go there, so I added a (?) icon button to have a popover to describe it, as our users won't know what the case number is for either. As I'm assuming it's a number, I'm sizing the field to indicate that it's not supposed to be extremely long.

Should "report label" actually be the name for the report? (We generally use "name" across cockpit, not "label".)

## Empty state

SOS report empty

This uses a basic PatternFly empty state. It's probably better to have an empty state for the page instead of a card with a list with a list empty state.

garrett:
> Out of curiosity, what toggles the "problems found" status?

Whoops! Yeah, I guess this is just debugging information, not matching against rules. Fixed in the new mockups.

---

I've modified the mockups based on your feedback. Additionally, I've added download buttons to the actions area... what do you think? 

SOS report list

Run report modal

TurboTurtle:
:+1: 

That looks great!

marusak:
@TurboTurtle are you interested in implementing of this new design. Do you need some help or do you want us to do implement it? Just asking so we don't wait on each other or don't do duplicate work.

TurboTurtle:
In the interest of time, I'd say I'll defer to you guys to move forward on this. I'd likely need a whole lot of hand holding for one, and I've also got a lot of stuff on the sos and rig projects to get through and don't want to hold this up and further.

martinpitt:
Replaced with PR #17240. @TurboTurtle already commented on that, so he's aware of that.

PR #15879, opened 2021-06-01
test: Fix PackageCase NetworkManager setup

martinpitt:
The libvirtd code path was incomplete: It's not enough to start
libvirtd, it would also need to ensure that the `default` network is
actually up.

However, we don't actually need this any more -- just adding a dummy
connection works on all Debian/Ubuntus, so let's simplify this. On the
Fedora family this is not necessary at all, as NM already manages the
main interface there.

----

This should fix a common [flake on debian-stable](https://logs.cockpit-project.org/logs/pull-15812-20210531-080229-aeb4f086-debian-stable/log.html). I am not 100% sure yet if we also need this on Testing Farm machines, hence starting out as draft.

### DIFF
--- a/test/verify/packagelib.py
+++ b/test/verify/packagelib.py
@@ -43,18 +43,10 @@ class PackageCase(MachineCase):
         else:
             raise NotImplementedError("unknown image " + self.machine.image)
 
-        # PackageKit refuses to work when offline
-        if "ubuntu" in self.image:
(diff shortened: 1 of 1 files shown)


martinpitt:
[packit F34](http://artifacts.dev.testing-farm.io/1d9f7da0-474a-42aa-9acd-2fa7ed37101b/) had unrelated failures, rawhide and CentOS 8 succeeded. Trying again.

marusak (approved):
Thanks, nice cleanup as well :)

PR #15880, opened 2021-06-01
.github: Change node_modules approach

allisonkarlitskaya:
The hand-off between a pull_request workflow and a workflow_run workflow
depending on it is ridiculously difficult to get right.  Stop trying.

The current approach of "implicitly rebuild node_modules every time you
push something" is also not really what we want.

To this end, introduce a "/npm install" command that can be invoked on
pull requests by contributors to the team.  This mostly does the same
thing, without the trouble of figuring out which pull request we're
running against and if we can trust the workflows running there.

We approach the permissions issue by explicitly managing the abilities
of the GITHUB_TOKENs we generate and we isolate `npm install` by:

 - using `--ignore-scripts`;

 - running it inside a container; and

 - running it in a separate job with no permissions.

We also introduce the use of a "node-cache upload" environment to
manage the deploy key used for pushing to the cache.  This is the first
use of deploy keys and environments.  In general, we plan to introduce
more moves to explicit use of environments in the future, in order to
prevent exposing secrets to workflows that don't need them and to move
away from using the cockpituous token for things that could be done with
a much-better-scoped credential.

Meanwhile, the check for "is node_modules okay?" is improved (using the
latest "git log" tricks) and moved to reposchutz.

 -  [x] Document the setup in  tools/README.node_modules


allisonkarlitskaya:
Not shown here: we need to do the following actions to make this work:
 - generate an ssh key
 - upload the public half of it as a writable deploy key on the `node-cache` repository
 - create a `node-cache upload` environment
 - upload the private half of it as the `NODE_CACHE_DEPLOY_KEY` secret in that environment
 - destroy the local copy

Might also be nice to document that process somewhere.

.github/workflows/node_modules.yml
```diff
@@ -0,0 +1,258 @@
...
+    steps:
+      - name: Interpret comment
+        id: interpret
+        uses: actions/github-script@v4
+        with:
+          result-encoding: string
+          script: |
+            // Make sure we're being spoken to
+            const { action, issue, comment: { body, id: comment_id } } = context.payload;
+            if (!(action === 'created' && body === '/npm install' && 'pull_request' in issue)) {
```
martinpitt, line 26: Why does this have to be code? Couldn't the whole job be guarded by

    if: github.event.issue.pull_request && startsWith(github.event.comment.body, '/livetest')

? (action == created is already a top-level workflow condition). That would prevent spawning a VM for each and every PR comment,  also more declarative, and nicer because it's a proper skip instead of the code-induced "fail".

.github/workflows/node_modules.yml
```diff
@@ -0,0 +1,258 @@
...
+            });
+            const permission = permission_response.data.permission;
+            if (permission !== 'admin' && permission !== 'write') {
+                await github.reactions.createForIssueComment({
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    comment_id: context.payload.comment.id,
+                    content: 'confused'
+                });
+                await github.issues.createComment({
```
martinpitt, line 47: Any chance this could edit the existing comment? it seems all too easy to create an infinite loop here, as this workflow triggers itself right now.

(As discussed, I'd *really* like this to be a task in the description, or a label instead, which would be structurally much simpler -- with a label you can even skip the whole permission check)

martinpitt:
> Might also be nice to document that process somewhere.

tools/README.node_modules perhaps? I added a task to the description.

.github/workflows/node_modules.yml
allisonkarlitskaya: I thought about the label thing too and even preferred it (and for the same reason) but it means that the entire workflow would end up running any time any label is set, including reporting a full set of irrelevant statuses to the PR.

.github/workflows/node_modules.yml
allisonkarlitskaya: For the reason stated in the comment: there's no real way to "exit" a workflow.  The best you can do is to fail it.  Otherwise you keep needing `if:` conditions on everything that follows.

To be honest, maybe that's worth removing the `/* HACK */`, saving the VM run, and seeing a bit less red on our actions page.  I'll take another look.

martinpitt: If the first job gets skipped, all  subsequent jobs which `need:` it will get skipped too, no?

.github/workflows/node_modules.yml
martinpitt: ooh, because it's `on: pull_request_target` type `labelled`? I was hoping we could use the [`on: issues`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#issues) trigger with type "labeled", but now I remember that this literally only works on issues. (I fell into that trap before already -- this is where workflow triggers and webhook events differ).

So let's ignore this for now.

.github/workflows/node_modules.yml
allisonkarlitskaya: Worth looking into, actually.

For steps, this is certainly not the case.  Skipping one step means the ones after it run still.  Even the word "skip" itself suggests going on to the next thing...

I'm happy to do it with conditions otherwise, anyway.

allisonkarlitskaya:
Another TODO, before I forget to write it down: I was trying to make a script (as part of `reposchutz`) to read all of the comments from the PR and build a list of commit IDs that came from the bot, but I got stuck in the world of pagination/iterators/async/etc.

.github/workflows/node_modules.yml
allisonkarlitskaya: Managed to get it working with explicit conditions on each job, which isn't so bad.

martinpitt (approved):
Thanks! This seems good enough to try it in practice now.

This now introduces a completely new structure, while for other tasks we use the "scan PR description for tasks" + "bots label" approach, so we'll need to see how this works for us in practice in terms of reproducibility, and comprehending what's going on. But let's give this a try!

martinpitt:
> The current approach of "implicitly rebuild node_modules every time you push something" is also not really what we want.

For the record, I disagree -- it is what *I* want. This is supposed to be a cache/optimization, and mostly happen behind the curtains. To me, package.json is still the single source of truth. But as I said, let's see this in practice, we can always change this.

martinpitt:
@allisonkarlitskaya : This complicates the npm-update bot: It will now autogenerate broken PRs, as these need a manual `/npm install` command and manual push now. PR #15895 already needed one human interaction, now they will need two. Compared to "everything just works automatically" that we had before, that's quite a regression..

allisonkarlitskaya:
> @allisonkarlitskaya : This complicates the npm-update bot: It will now autogenerate broken PRs, as these need a manual `/npm install` command and manual push now.

cockpit-project/bots#2088 should help a bit.

I also have a TODO item that the `npm install`-runner might forcepush an update on the original PR, but that seemed a bit too much.  Maybe it makes sense for npm-update PRs, though?

martinpitt:
@allisonkarlitskaya Thanks for 2088! My gut feeling is to leave it be for a while, and revisit that if it annoys us too much. Otherwise your head will explode at some point :exploding_head: ..

PR #15883, opened 2021-06-03
Fix missing borders in Selects

KKoukiou:
Disable mergeLongHand in cssnano.

The official fix has already landed upstream but we unfortunately can't
update cssnano without updating also CssMinimizerPlugin.
An update of CssMinimizerPlugin will not work without webpack-5.

Therefore let's keep the WA for now.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -433,7 +433,15 @@ module.exports = {
 
     optimization: {
         minimize: production,
-        minimizer: [new TerserJSPlugin({ extractComments : false }), new CssMinimizerPlugin()],
+        minimizer: [
+           new TerserJSPlugin({ extractComments : false }),
+           new CssMinimizerPlugin({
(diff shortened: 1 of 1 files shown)


marusak:
Sorry, I am bit out of context here. Can you please add link to "The official fix" and as well post before and after pictures?

KKoukiou:
> Sorry, I am bit out of context here. Can you please add link to "The official fix" and as well post before and after pictures?

ah sorry - forgot to link the fix - we can update the commit message on merge https://github.com/cssnano/cssnano/pull/1057

KKoukiou:
With the fix:

Screen Shot 2021-06-04 at 11 59 50

Before the fix:
Screen Shot 2021-06-05 at 15 15 34

marusak (approved):
Thanks!

PR #15887, opened 2021-06-04
lib/patternfly: Remove :focus-visible outline from some elements

mvollmer:
- [ ] https://github.com/patternfly/patternfly/issues/4121

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -123,6 +123,25 @@ svg {
     z-index: 2;
 }
 
+// Remove the default :focus-visible outline from elements that
+// receive the :focus-visible class also when focused via the mouse
+// and already have a visible :focus style.
+//
+// The default :focus-visible rendering differs from browser to
+// browser, changes over time, doesn't necessarily fit into the
+// Patternfly color scheme, looks ugly when used together with the
+// Patternfly :focus styling, and serves no purpose when :focus is
+// already visible.
+//
+// This would all be acceptable if these elements wouldn't receive
+// :focus-visible when focused with the mouse.
+//
+// https://github.com/patternfly/patternfly/issues/4121
+.pf-c-form-control:focus-visible {
+  outline: none;
+}
+
+
 // Alerts use elements that have fonts set in other frameworks (including PF3);
 // generally, this is an H4 that often has a font size and sometimes family set.
 // Therefore, it should inherit from the alert font set at the pf-c-alert level.


pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -123,6 +123,25 @@ svg {
...
+// The default :focus-visible rendering differs from browser to
+// browser, changes over time, doesn't necessarily fit into the
+// Patternfly color scheme, looks ugly when used together with the
+// Patternfly :focus styling, and serves no purpose when :focus is
+// already visible.
+//
+// This would all be acceptable if these elements wouldn't receive
+// :focus-visible when focused with the mouse.
+//
+// https://github.com/patternfly/patternfly/issues/4121
```
KKoukiou, line 139: Removing the ring is very intrusive IMO, I would rather wait for the linked issue to get an answer!

mvollmer:
If I would be king for a day, this is what I would do. :-)

Right now focusing a text input with the mouse looks like this:

Firefox:
image

Chrome:
image

GNOME Web:
image

The thick blue bottom border comes from Patternfly. Firefox adds a thin blue outline, Chrome adds a thick black outline, GNOME Web adds a faint dotted outline.

With the change in this PR, focussing a text input with the mouse looks like this:

Firefox:
image

Chrome:
image

GNOME Web:
image

GNOME Web doesn't seem to support :focus-visible but also has a suitably non-annoying focus ring.

mvollmer:
This PR makes tabbing through a dialog visually less consistent on Chrome: some elements get styled according to Patternfly, some elements get the default thick black outline. It would be good to not interfere with :focus-visible styling for this reason, but the damage the default does to mouse-focused text inputs is just too high, IMO. Thus, if anything, we should probably take more control over :focus-visible and define it for all elements.

pkg/lib/patternfly/patternfly-4-overrides.scss
mvollmer: Yes, but I have to get this out of my system. :-)

garrett (changes requested):
While the "before" screenshots are all wildly inconsistent, the "after" screenshots do not look focused.

Chrome changed its focused outlines recently,
Firefox changed its default widgets in 89 (last week),
WebKit's is heavily influenced by macOS (and GTK's port might not reflect intent).

And, of course, all other browsers are basically Chrome.

But first rule of web development with accessibility in mind: Don't break focus states. (Often, this is in the form of `outline`.) The after ones do not look focused, therefore this needs work.

PR #15888, opened 2021-06-04
HACKING: add dnf (sub)command install

smwltr:
dnf commands are not optional - probably just a typo.


marusak (dismissed):
Thanks!

marusak:
2 commits are fine. However they should have "Fixes #" lines. But I see that now you closed both of those issues manually.
https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

HACKING.md
```diff
@@ -37,15 +37,13 @@ Note that `tools/cockpit.spec` is a template filled in by
 `tools/gen-spec-dependencies`, and cannot be directly parsed by dnf.
 The following should work in a fresh Git clone:
 
-    $ sudo dnf install dnf-utils
-    $ TEMPFILE=$(mktemp -u --suffix=.spec) && \
-      sed 's/%{npm-version:.*}/0/' tools/cockpit.spec >$TEMPFILE && \
-      sudo dnf builddep --spec $TEMPFILE && \
-      rm $TEMPFILE
+    $ sudo dnf install dnf-utils python-srpm-macros
```
marusak, line 40: The previous approach would not actually change any files in checkout. When I do your approach then `git diff` contains bunch of changes.
Since one would do this once, it is fine I guess. By any chance can you explain why the previous approach did not work for you and you had to use `tools/gen-spec-dependencies`?

smwltr: I thought I should run those commands, but now I see the JS is not needed for building. Right, I think this confused me:
> To actually build the Cockpit binaries themselves from source (including to run the integration tests locally), you will need additional header files and other components. **Check tools/cockpit.spec for the concrete Fedora build dependencies.**

> Note that tools/cockpit.spec is a template filled in by tools/gen-spec-dependencies, and cannot be directly parsed by dnf. The following should work in a fresh Git clone:

May I suggest a wording change:

To actually build the Cockpit binaries themselves from source, including to run the integration tests locally, you will need additional header files and other components. tools/cockpit.spec contains the concrete Fedora build dependencies. However, it is a template, and so cannot be directly parsed by dnf. It is modified by tools/gen-spec-dependencies when making the dist target (is that true?). As the JS/node/npm/etc/? dependencies are only used for packaging (or whatever the true reason is), to install Fedora build dependencies, the following should work in a fresh Git clone:

The original sounds like it wants me actually check/look at the contents of the file. Then straight away mentions using gen-spec-dependencies. The above might be more clear. I think also the code formatting could be dropped around tools/gen-spec-dependencies and tools/cockpit.spec, as it put it's on par with the other commands in this document that one should run.

smwltr: Never mind about the code formatting (backticks). I should wait till read more docs before making such a suggestion. I did look over these `https://github.com/cockpit-project/cockpit/wiki/Cockpit-Coding-Guidelines`, but maybe it's just an unspoken convention.

marusak: > It is modified by tools/gen-spec-dependencies when making the dist target (is that true?)

It is. Or you can just say "when doing official builds".

> As the JS/node/npm/etc/? dependencies are only used for packaging (or whatever the true reason is)

There is no "JS" per se. Node and npm are used only to put all the things together and the result that is produced does not need these anymore.

> I think also the code formatting could be dropped around tools/gen-spec-dependencies and tools/cockpit.spec, as it put it's on par with the other commands in this document that one should run.

Lets not do that. It is a good practice to format files and commands this way. It makes it much easier to read as you know what you need to parse.

smwltr: I hope that is clearer now. There were also some more packages that needed to be installed for the tests. It took a while to understand what my problem was. I think it may have been IOMMU related, but not sure, as I have no problems with other nested OSes on the same hardware. So I'm not sure if it IOMMU is the problem. I'd like to improve the docs in this regard also.

HACKING.md
```diff
@@ -25,34 +25,35 @@ On Fedora:
...
 
-    $ sudo dnf chromium-headless dbus-daemon
+    $ sudo dnf install chromium-headless dbus-daemon
 
 For running integration tests, the following dependencies are required:
 
     $ sudo dnf install curl expect xz rpm-build chromium-headless \
-        libvirt-daemon-kvm libvirt-client python3-libvirt
+        libvirt-daemon-kvm libvirt-client python3-libvirt python3-pyflakes \
+        python3-pycodestyle python3-pyyaml
```
martinpitt, line 53: pyflakes and pycodestyle are not necessary for the integration tests. They are just being used by `make check`, and even that just skips these two if the packages are not installed. Doesn't do much harm, though.

martinpitt (approved):
Thank you!

smwltr:
On 6/14/21 2:38 PM, Martin Pitt wrote:
> ***@***.**** approved this pull request.
> 
> Thank you!
> 
> ------------------------------------------------------------------------
> 
> In HACKING.md 
> <https://github.com/cockpit-project/cockpit/pull/15888#discussion_r650658717>:
> 
>> +        libvirt-daemon-kvm libvirt-client python3-libvirt python3-pyflakes \
> +        python3-pycodestyle python3-pyyaml
> 
> pyflakes and pycodestyle are not necessary for the integration tests. 
> They are just being used by |make check|, and even that just skips these 
> two if the packages are not installed. Doesn't do much harm, though.

Indeed. I suppose that they could have their own '$ sudo dnf install 
...' line before the '$ make check' line.

Also, from what I can tell, the pyYaml package is only needed to run the 
integration tests:
***@***.*** cockpit]$ ./test/common/run-tests
   File "/home/smwltr/cockpit/./test/common/run-tests", line 19, in <module>
     import testlib
   File "/home/smwltr/cockpit/test/common/testlib.py", line 43, in <module>
     from fmf_metadata.base import set_obj_attribute, is_test_function, 
generic_metadata_setter
   File "/home/smwltr/cockpit/test/common/fmf_metadata/__init__.py", 
line 1, in <module>
     from fmf_metadata.base import FMF
   File "/home/smwltr/cockpit/test/common/fmf_metadata/base.py", line 5, 
in <module>
     import yaml
ModuleNotFoundError: No module named 'yaml'
***@***.*** cockpit]$

May it should be mentioned in 'test/README.md' instead?

I was thinking of expanding the developer docs for Debian, if that's 
something you feel is useful. I don't know which distros are supported 
for development. I think developer documentation for those distros would 
expose cockpit to more developers.

martinpitt:
@smwltr : Moving integration-test only dependencies to test/README.md would indeed be better IMHO. HACKING already refers to that in the "Running the integration test suite" section, and otherwise only talks about building and the unit tests.

Expanding the developer docs for Debian is much appreciated. E. g. @allisonkarlitskaya had developed Cockpit on Debian stable for quite a while, and our docs do have some "On debian, run apt.." instructions, but they are probably not very consistent.

Thanks!

PR #10119, opened 2018-09-20
tools: Stop hardcoding the Python 3 package name

martinpitt:
Consistently use the RPM macro instead and depend on the interpreter
executable directly.

Since the macro does not yet exist in current Fedora 28/29 (it will
soon), provide a fallback with the current value.

Thanks to Tomas Orsava for the original patch!

https://bugzilla.redhat.com/show_bug.cgi?id=1631174

PR #15890, opened 2021-06-04
cockpit.spec: python3 Requires is malformed

smwltr:
gen-spec-dependencies does not replace this line with something that dnf builddep can use. python3-dbus follows shortly after. So I assume that hardcoding python3 is also fine.

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -648,7 +648,7 @@ Recommends: udisks2-lvm2 >= 2.6
 Recommends: udisks2-iscsi >= 2.6
 Recommends: device-mapper-multipath
 Recommends: clevis-luks
-Requires: %{__python3}
+Requires: python3
 %if 0%{?suse_version}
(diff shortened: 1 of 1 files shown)


marusak:
This is valid macro, see https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_macros

Is it causing an issue for you? The reason why we use this and not just `python3` is historical - when python2 was being replaced with python3 different distros packaged it differently, there were different alternatives and the whole system was (still is?) a mess. These macros are supposed to shield packagers from it. It is possible that today `python3` directly would work for eveyrhing we support from master, but unless we have to change it, I would prefer to keep this macro.

marusak:
> Is it causing an issue for you?

I see answer in https://github.com/cockpit-project/cockpit/issues/15891, I will follow up there

marusak:
FTR https://github.com/cockpit-project/cockpit/pull/10119 (also related BZ)

PR #15892, opened 2021-06-04
livetest: stop spinning up a vm every single time anyone writes any comment on any PR or issue

KKoukiou:
If we put a `if:` on each `job:` then the workflow will dispatch, but at
least it won't start any VMs.

### DIFF
--- a/.github/workflows/livetest.yml
+++ b/.github/workflows/livetest.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   pr-info:
+    if: startsWith(github.event.comment.body, '/livetest')
     runs-on: ubuntu-20.04
     steps:
       - name: Query author repository permissions


allisonkarlitskaya (changes requested):
As written, it seems like you really only need one job here.  The job can have the static checks based on the event blob and the permissions checking can just fail (`core.setFailed()` or something like that) in case access is denied.

If you want to add/remove additional permissions to the github token, for example, then it might make sense to keep things separate...

But in any case, now that I mention it: please introduce `permissions:` to scope the credential to what is actually required here.

KKoukiou:
Works as expected: tested on https://github.com/KKoukiou/cockpit/pull/3 
First normal run: https://github.com/KKoukiou/cockpit/actions/runs/906460733
And run which should not spawn a VM: https://github.com/KKoukiou/cockpit/actions/runs/906462637

KKoukiou:
> As written, it seems like you really only need one job here. The job can have the static checks based on the event blob and the permissions checking can just fail (`core.setFailed()` or something like that) in case access is denied.
> 
> If you want to add/remove additional permissions to the github token, for example, then it might make sense to keep things separate...
> 
> But in any case, now that I mention it: please introduce `permissions:` to scope the credential to what is actually required here.

Sure, this workflow is broken anyway. Mind getting this tiny improvement in so that we don't spawn VMs like that and I will take a look in improvements another time.

allisonkarlitskaya (approved):
Approved because, strictly speaking, it's an improvement :)

Looking forward to the follow-up.

martinpitt (approved):
This could have been no-test, but prodding the tests a bit to not ruin statistics. (Also, the reposchutz one needed the label). Danke!

PR #15896, opened 2021-06-07
test: Allow selinux denials regarding "dma_heap"

mvollmer:
They don't seem to affect operations significantly but are blocking
image refreshes.


martinpitt (approved):
Danke! To make double-sure, feel free to trigger a run against the refreshed rhel-9-0 image (or f-34).

Note that you'll probably need one more for rhel-8-5, the SELinux violations there look different.

martinpitt:
The reposchutz fix landed, doing the label dance to re-run it.

mvollmer:
The three ´rhel-9-0#2086` will behandled in that PR. The selinux denial ignoring is working here as planned.

PR #15897, opened 2021-06-07
reposchutz: drop superfluous check for tag

allisonkarlitskaya:
`git fetch` isn't fetching the named tag, but rather just populating
`FETCH_HEAD`.  That's OK.  We already check that the fetch was
successful and we got the thing we expected, so checking that the tag is
also equal to the thing we expect is just superfluous.  We never use it
anyway.

### DIFF
--- a/.github/workflows/reposchutz.yml
+++ b/.github/workflows/reposchutz.yml
@@ -62,7 +62,6 @@ jobs:
                   exit 1
               fi
               [ "$(git rev-parse FETCH_HEAD)" = "${node_modules}" ]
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! (BTW, if you do more workflow changes today, please use no-test; we have much reduced CI capacity due to e2e being down)

martinpitt:
The [reposchutz failure](https://github.com/cockpit-project/cockpit/pull/15897/checks?check_run_id=2762022484) here is still from the old version, this PR can't self-validate.

PR #15899, opened 2021-06-07
packagekit: Introduce kpatch integration

marusak:
Now there are patches available, I tested this on real rhel 8.4 machine.
@garrett and @martinpitt WDYT?

kpatch is available, but not installed:
Screenshot from 2021-06-07 12-48-03Install

kpatch settings dialog:
image

Kpatch is installed, enabled and all the fun stuff. Also patches are available (see in the list)
Screenshot from 2021-06-07 12-35-49k1

There are some patches applied:
Screenshot from 2021-06-07 12-36-42k2

This one is probably something we don't want. I run different kernel, but I have patches for different kernel installed. I guess I should not list it? (the other kernel is still available, I just don't use it right now). Should it list only installed patches for the current kernel?
Screenshot from 2021-06-07 12-42-06installed

Release note:

## Software Updates: Introduce basic kpatch support

On operating systems with support for live kernel patching Cockpit now offers to enable this feature. Software updates page shows if `kpatch` is supported, if there are any patches active as well as supports enabling subscribing to patches for future kernel versions.  

kpatch1

### DIFF
--- a/pkg/packagekit/kpatch.jsx
+++ b/pkg/packagekit/kpatch.jsx
@@ -1,0 +1,372 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from "react";
+import PropTypes from "prop-types";
+import {
+    Alert, Button, Checkbox,
+    Flex, FlexItem,
+    Modal, Radio,
+    Split, SplitItem,
+    Stack,
+    Text, TextVariants
+} from "@patternfly/react-core";
+import { InfoIcon } from "@patternfly/react-icons";
+
+import cockpit from "cockpit";
+import { proxy as serviceProxy } from "service";
+import { check_missing_packages } from "packagekit.js";
+import { install_dialog } from "cockpit-components-install-dialog.jsx";
+
+const _ = cockpit.gettext;
+
+export class KpatchSettings extends React.Component {
+    constructor() {
+        super();
+
+        this.state = {
+            loaded: false,
+            auto: null, // `dnf kpatch` is set to `auto`
+            enabled: null, // kpatch.service is enabled
+            missing: [], // missing packages from `kpatch`, `kpatch-dnf`
+            unavailable: [], // unavailable packages from `kpatch`, `kpatch-dnf`
+
+            // Modal states
+            error: "",
+            updating: false,
+            showModal: false,
+            applyCheckbox: false, // state of the checkbox
+            justCurrent: null, // radio state
+
+            kernelName: "", // uname -r
+            patchName: null, // kpatch-patch name
+            patchInstalled: null, // kpatch-patch installed
+            patchUnavailable: null, // kpatch-patch available
+        };
+
+        this.kpatchService = serviceProxy("kpatch");
+
+        this.checkSetup = this.checkSetup.bind(this);
+        this.handleChange = this.handleChange.bind(this);
+        this.onClose = this.onClose.bind(this);
+        this.handleInstall = this.handleInstall.bind(this);
+    }
+
+    // Only current patches or also future ones
+    current(enabled, installed, unavailable) {
+        return enabled && (installed || unavailable);
+    }
+
+    componentDidMount() {
+        check_missing_packages(["kpatch", "kpatch-dnf"])
+                .then(d =>
+                    this.checkSetup().then(() =>
+                        this.setState({
+                            loaded: true,
+                            unavailable: d.unavailable_names || [],
+                            missing: d.missing_names || [],
+                        })
+                    )
+                )
+                .catch(e => console.log("Could not determine kpatch availability:", JSON.stringify(e)));
+
+        this.kpatchService.addEventListener('changed', () => {
+            this.setState(state => {
+                const current = this.current(this.kpatchService.enabled, state.patchInstalled, state.patchUnavailable);
+                return ({
+                    enabled: this.kpatchService.enabled,
+                    justCurrent: current && !state.auto,
+                    applyCheckbox: current,
+                });
+            });
+        });
+    }
+
+    checkSetup() {
+        // TODO - replace both with `dnf kpatch status` once https://github.com/dynup/kpatch-dnf/pull/8 lands
+        const kpatch_promise = cockpit.file("/etc/dnf/plugins/kpatch.conf").read()
+                .then(data => {
+                    if (data) {
+                        const auto = /autoupdate\s*=\s*True/i.test(data);
+                        this.setState((state, _) => {
+                            const current = this.current(state.enabled, state.patchInstalled, state.patchUnavailable);
+                            return ({
+                                auto: !!auto,
+                                justCurrent: current && !auto,
+                                applyCheckbox: current,
+                            });
+                        });
+                    }
+                })
+                .catch(() => true); // Ignore errors, most likely just does not exist
+
+        const uname_promise = cockpit.spawn(["uname", "-r"])
+                .then(data => {
+                    const fields = data.split("-");
+                    const kpp_kernel_version = fields[0].replaceAll(".", "_");
+                    let release = fields[1].split(".");
+                    release = release.slice(0, release.length - 2); // remove el8.x86_64
+                    const kpp_kernel_release = release.join("_");
+                    const patch_name = ["kpatch-patch", kpp_kernel_version, kpp_kernel_release].join("-");
+                    return check_missing_packages([patch_name])
+                            .then(d =>
+                                this.setState((state, _) => {
+                                    const installed = (d.unavailable_names || []).length === 0 && (d.missing_names || []).length === 0;
+                                    const unavailable = (d.unavailable_names || []).length > 0;
+                                    const current = this.current(state.enabled, installed, unavailable);
+                                    return ({
+                                        kernelName: data,
+                                        patchName: patch_name,
+                                        patchInstalled: installed,
+                                        patchUnavailable: unavailable,
+                                        justCurrent: current && !state.auto,
+                                        applyCheckbox: current,
+                                    });
+                                })
+                            );
+                })
+                .catch(console.error);
+
+        return Promise.allSettled([kpatch_promise, uname_promise]);
+    }
+
+    handleInstall() {
+        this.setState({ updating: true });
+        install_dialog(this.state.missing)
+                .then(() => this.setState({ missing: [], updating: false }))
+                .catch(() => this.setState({ updating: false }));
+    }
+
+    onClose() {
+        this.setState((state, _) => {
+            const current = this.current(state.enabled, state.patchInstalled, state.patchUnavailable);
+            return ({
+                justCurrent: current && !state.auto,
+                applyCheckbox: current,
+                showModal: false,
+                error: "",
+            });
+        });
+    }
+
+    handleChange() {
+        this.setState({ updating: true });
+
+        if (this.state.applyCheckbox) {
+            let install = Promise.resolve(true);
+            if (this.state.justCurrent) {
+                install = new Promise((resolve, reject) => {
+                    cockpit.spawn(["dnf", "-y", "kpatch", "manual"], { superuser: "require", err: "message" })
+                            .then(() => {
+                                if (!this.state.patchUnavailable && !this.state.patchInstalled)
+                                    // TODO - replace with `dnf kpatch install` once https://github.com/dynup/kpatch-dnf/pull/8 lands
+                                    cockpit.spawn(["dnf", "-y", "install", this.state.patchName], { superuser: "require", err: "message" }).then(resolve)
+                                            .catch(reject);
+                                else
+                                    resolve();
+                            })
+                            .catch(reject);
+                });
+            } else {
+                install = cockpit.spawn(["dnf", "-y", "kpatch", "auto"], { superuser: "require", err: "message" });
+            }
+            install
+                    .then(() =>
+                        this.kpatchService.enable().then(() =>
+                            this.kpatchService.start().then(() =>
+                                this.setState({ showModal: false, error: "" })
+                            )
+                        )
+                    )
+                    .catch(e => this.setState({ error: e.toString() }))
+                    .finally(() => this.checkSetup().then(() => this.setState({ updating: false })));
+        } else {
+            cockpit.spawn(["dnf", "-y", "kpatch", "manual"], { superuser: "require", err: "message" })
+                    .then(() =>
+                        this.kpatchService.disable().then(() =>
+                            this.kpatchService.stop().then(() =>
+                                this.setState({ showModal: false, error: "" })
+                            )
+                        )
+                    )
+                    .catch(e => this.setState({ error: e.toString() }))
+                    .finally(() => this.checkSetup().then(() => this.setState({ updating: false })));
+        }
+    }
+
+    render() {
+        // Not yet recognized
+        if (this.state.loaded === false || this.state.patchName === null)
+            return null;
+
+        // Not supported on this system
+        if (this.state.unavailable.length > 0)
+            return null;
+
+        let state = _("Enabled");
+        let actionText = _("Edit");
+        let action = () => this.setState({ showModal: true });
+
+        if (this.state.missing.length > 0) {
+            state = _("Not installed");
+            actionText = _("Install");
+            action = this.handleInstall;
+        } else if (!this.state.enabled) {
+            state = _("Disabled");
+            actionText = _("Enable");
+        }
+
+        const kernel_name = this.state.kernelName ? " (" + this.state.kernelName + ")" : "";
+        const error = this.state.error ? <Alert variant='danger' isInline title={this.state.error} /> : null;
+
+        const body = <Checkbox id="apply-kpatch"
+                               isChecked={this.state.applyCheckbox}
+                               label={_("Apply kernel patches")}
+                               onChange={checked => this.setState({ applyCheckbox: checked })}
+                               body={<>
+                                   <Radio id="current-future"
+                                          label={_("for current and future kernels")}
+                                          onChange={() => this.setState({ justCurrent: false })}
+                                          isDisabled={!this.state.applyCheckbox}
+                                          isChecked={!this.state.justCurrent} />
+                                   <Radio id="current-only"
+                                          label={_("for current kernel only") + kernel_name}
+                                          onChange={() => this.setState({ justCurrent: true })}
+                                          isDisabled={!this.state.applyCheckbox}
+                                          isChecked={this.state.justCurrent} />
+                               </>}
+        />;
+
+        return (<>
+            <div id="kpatch-settings">
+                <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                    <Flex grow={{ default: 'grow' }} alignItems={{ default: 'alignItemsBaseline' }}>
+                        <FlexItem>
+                            <b>{_("Kernel patching")}</b>
+                        </FlexItem>
+                        <FlexItem>
+                            {state}
+                        </FlexItem>
+                    </Flex>
+                    <Flex>
+                        <Button variant="secondary"
+                                isSmall
+                                isDisabled={!this.props.privileged || this.state.updating}
+                                onClick={action}>
+                            {actionText}
+                        </Button>
+                    </Flex>
+                </Flex>
+            </div>
+            <Modal position="top" variant="small" id="kpatch-setup" isOpen={this.state.showModal}
+                title={_("Kernel patch settings")}
+                onClose={ this.onClose }
+                footer={
+                    <>
+                        <Button variant="primary"
+                                isLoading={ this.state.updating }
+                                isDisabled={ this.state.updating }
+                                onClick={ this.handleChange }>
+                            {_("Apply")}
+                        </Button>
+                        <Button variant="link"
+                                isDisabled={ this.state.updating }
+                                onClick={ this.onClose }>
+                            {_("Cancel")}
+                        </Button>
+                    </>
+                }>
+                <>
+                    {error}
+                    {body}
+                </>
+            </Modal>
+        </>);
+    }
+}
+
+KpatchSettings.propTypes = {
+    privileged: PropTypes.bool.isRequired,
+};
+
+export class KpatchStatus extends React.Component {
+    constructor() {
+        super();
+
+        this.state = {
+            loaded: [],
+            installed: [],
+            changelog: null, // FIXME - load changelog
+        };
+    }
+
+    componentDidMount() {
+        cockpit.spawn(["kpatch", "list"], { superuser: "try", err: "ignore", environ: ["LC_MESSAGES=C"] })
+                .then(m => {
+                    const parts = m.trim().split("\n\n");
+                    if (parts.length !== 2 ||
+                        !parts[0].startsWith("Loaded patch modules:") ||
+                        !parts[1].startsWith("Installed patch modules:")) {
+                        console.warn("Unexpected output from `kpatch list`", m);
+                        return;
+                    }
+
+                    const loaded = parts[0].split("\n")
+                            .slice(1)
+                            .map(i => i.split(" ")[0]);
+                    const installed = parts[1].split("\n")
+                            .slice(1)
+                            .map(i => i.split(" ")[0]);
+                    this.setState({ loaded, installed });
+                })
+                .catch(() => true); // Ignore errors
+    }
+
+    render() {
+        let text = [];
+        text = this.state.loaded.map(i =>
+            <Text key={i} component={TextVariants.p}>
+                { cockpit.format(_("Kernel patch $0 is active"), i) }
+            </Text>
+        );
+
+        if (text.length === 0)
+            text = this.state.installed.map(i =>
+                <Text key={i} component={TextVariants.p}>
+                    { cockpit.format(_("Kernel patch $0 is installed"), i) }
+                </Text>
+            );
+
+        if (text.length > 0)
+            return (
+                <Split hasGutter>
+                    <SplitItem>
+                        <InfoIcon />
+                    </SplitItem>
+                    <SplitItem isFilled>
+                        <Stack>
+                            {text}
+                        </Stack>
+                    </SplitItem>
+                </Split>
+            );
+
+        return null;
+    }
+}
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -37,6 +37,8 @@ done
 """
 
 OSesWithoutTracer = ["debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable", "fedora-coreos"]
+OSesWithoutKpatch = ["debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable",
+                     "fedora-coreos", "fedora-33", "fedora-34", "centos-8-stream"]
 
 
 class NoSubManCase(PackageCase):
@@ -136,6 +138,131 @@ class TestUpdates(NoSubManCase):
         else:
             self.fail("Timed out waiting for updates spinner to go away")
 
+    @skipImage("kpatch is not available", *OSesWithoutKpatch)
+    def testKpatch(self):
+        b = self.browser
+        m = self.machine
+
+        kernel_ver_arch = m.execute("uname -r").strip()
+        fields = kernel_ver_arch.split("-")
+        kpp_kernel_version = fields[0].replace(".", "_")
+        release = fields[1].split(".")[:-2] # remove el8.x86_64
+        kpp_kernel_release = "_".join(release)
+
+        sanitized_kernel_ver = "_".join(kernel_ver_arch.split(".")[:-2])
+        self.createPackage("-".join(["kpatch-patch", kpp_kernel_version, kpp_kernel_release]), "0", "0", arch=self.secondary_arch,
+                           provides="kpatch-patch = {0}".format(kernel_ver_arch))
+        self.enableRepo()
+
+        self.restore_file("/etc/dnf/plugins/kpatch.conf")
+        m.execute("systemctl disable --now kpatch")
+
+        m.start_cockpit()
+        b.login_and_go("/updates")
+
+        # Enable and install kpatch
+        b.wait_in_text("#kpatch-settings", "Disabled")
+        b.click("#kpatch-settings button:contains('Enable')")
+
+        # Apply kernel patches for current and future kernels
+        b.click("#apply-kpatch")
+        b.wait_visible("#apply-kpatch:checked")
+        b.wait_visible("#current-future:checked")
+
+        b.click("button:contains('Apply')")
+        b.wait_not_present("#kpatch-setup")
+
+        self.assertIn("True", m.execute("grep autoupdate /etc/dnf/plugins/kpatch.conf"))
+        self.assertEqual(m.execute("systemctl is-enabled kpatch").strip(), "enabled")
+        self.assertEqual(m.execute("systemctl is-active kpatch").strip(), "active")
+        # Patches should be installed
+        m.execute("rpm -q kpatch-patch-" + sanitized_kernel_ver)
+
+        # Faking real patches is really hard, so fake `kpatch list` command
+        # To make sure it has the same structure as we expect it to be first check the real command
+        self.assertEqual(m.execute("kpatch list"), "Loaded patch modules:\n\nInstalled patch modules:\n")
+
+        kpatch = """
+#!/bin/bash
+echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled patch modules:\nkpatch_3_10_0_1062_1_1 (3.10.0-1062.el7.x86_64)"
+"""
+
+        self.createPackage("kpatch-patch-" + sanitized_kernel_ver, "1", "0", arch=self.secondary_arch,
+                           provides="kpatch-patch = {0}".format(kernel_ver_arch),
+                           content={"/usr/local/bin/kpatch": kpatch},
+                           postinst="chmod +x /usr/local/bin/kpatch; mount -o bind /usr/local/bin/kpatch /usr/sbin/kpatch")
+        self.enableRepo()
+        self.addCleanup(m.execute, "umount /usr/sbin/kpatch")
+
+        b.click("#status .pf-c-card__actions button")
+        b.wait_visible(".pf-c-badge:contains('patches')")
+        b.click("button:contains('Install all updates')")
+        b.click("button:contains('Continue')")
+        b.wait_in_text("#status", "Kernel patch kpatch_3_10_0_1062_1_1 is active")
+
+        # Switch to 'for current kernel only'
+        b.click("#kpatch-settings button:contains('Edit')")
+        b.click("#current-only")
+        b.wait_visible("#current-future:not(:checked)")
+        b.wait_visible("#current-only:checked")
+
+        b.click("button:contains('Apply')")
+        b.wait_not_present("#kpatch-setup")
+
+        self.assertIn("False", m.execute("grep autoupdate /etc/dnf/plugins/kpatch.conf"))
+        self.assertEqual(m.execute("systemctl is-enabled kpatch").strip(), "enabled")
+        self.assertEqual(m.execute("systemctl is-active kpatch").strip(), "active")
+        # Patches should be installed
+        m.execute("rpm -q kpatch-patch-" + sanitized_kernel_ver)
+
+        # Test disabling applying patches
+        b.click("#kpatch-settings button:contains('Edit')")
+        b.wait_visible("#apply-kpatch:checked")
+        b.click("#apply-kpatch")
+        b.wait_visible("#apply-kpatch:not(:checked)")
+
+        b.click("button:contains('Apply')")
+        b.wait_not_present("#kpatch-setup")
+
+        b.wait_in_text("#kpatch-settings", "Disabled")
+        b.click("#kpatch-settings button:contains('Enable')")
+
+        # Test closing of the dialog and resetting of changes
+        b.wait_visible("#apply-kpatch:not(:checked)")
+        b.click("#apply-kpatch")
+        b.wait_visible("#apply-kpatch:checked")
+        b.click("button:contains('Cancel')")
+        b.wait_not_present("#kpatch-setup")
+        b.click("#kpatch-settings button:contains('Enable')")
+        b.wait_visible("#apply-kpatch:not(:checked)")
+
+        # Test 'for current kernel only' from clean state
+        m.execute("umount /usr/sbin/kpatch")
+        m.execute("systemctl restart kpatch")
+        m.execute("dnf -y remove kpatch-patch-" + sanitized_kernel_ver)
+        b.reload() # Not listening on patches being removed
+        b.enter_page("/updates")
+
+        b.wait_in_text("#kpatch-settings", "Disabled")
+        b.click("#kpatch-settings button:contains('Enable')")
+
+        # Apply kernel patches for current and future kernels
+        b.click("#apply-kpatch")
+        b.wait_visible("#apply-kpatch:checked")
+        b.click("#current-only")
+        b.wait_visible("#current-future:not(:checked)")
+        b.wait_visible("#current-only:checked")
+
+        b.click("button:contains('Apply')")
+        b.wait_not_present("#kpatch-setup")
+
+        self.assertIn("False", m.execute("grep autoupdate /etc/dnf/plugins/kpatch.conf"))
+        self.assertEqual(m.execute("systemctl is-enabled kpatch").strip(), "enabled")
+        self.assertEqual(m.execute("systemctl is-active kpatch").strip(), "active")
+        # Patches should be installed
+        m.execute("rpm -q kpatch-patch-" + sanitized_kernel_ver)
+
+
     def testBasic(self):
         # no security updates, no changelogs
         b = self.browser
@@ -153,7 +280,6 @@ class TestUpdates(NoSubManCase):
         # no updates on the Software Updates page
         b.go("/updates")
         b.enter_page("/updates")
-        b.wait_not_present("#status")
         b.wait_in_text("#status", "System is up to date")
         # PK starts from a blank state, thus should force refresh and set the "time since" to 0
         b.wait_in_text("#last-checked", "Last checked: less than a minute ago")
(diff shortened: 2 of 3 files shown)


pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,325 @@
...
+};
+
+export class KpatchStatus extends React.Component {
+    constructor() {
+        super();
+
+        this.state = {
+            loaded: [],
+            installed: [],
+            changelog: null, // FIXME - load changelog
```
marusak, line 254: Right, I forgot about this. I'll take a look at it now. But I guess even a follow up will do?

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        const details = [];
+        if (this.state.enabled !== this.state.futureEnabled) {
+            if (this.state.futureEnabled)
+                details.push(_("kpatch.service will be enabled. After each reboot patches will be applied."));
+            else
+                details.push(_("kpatch.service will be disabled. After reboot no patches will be applied."));
+        }
+        // FIXME - should we uninstall current patches as well?
```
marusak, line 172: Should we?

pkg/packagekit/kpatch.jsx
martinpitt: Is that treated differently than a "normal" package changelog? But yes, fine for me to do that in a follow-up.

pkg/packagekit/kpatch.jsx
martinpitt: I'm inclined to say that this should not do any magic beyond `pkcon update`/`dnf update`. Cleanup of old kernels (and corresponding kpatches) belongs to the packaging system, not cockpit. It's IMHO not acceptable to pile up either of them with automatic updates or not paying attention.

martinpitt:
> Should it list only installed patches for the current kernel?

I think yes, otherwise it's misleading, as kpatches don't really apply to available, but not-booted kernels. Once you reboot into the "other" kernel, it will still not apply the kpatch.

pkg/packagekit/kpatch.jsx
marusak: no, it is the same changelog. Maybe a bit more parsing and pointing to CVEs...

pkg/packagekit/kpatch.jsx
marusak: Agree.

marusak:
> Once you reboot into the "other" kernel, it will still not apply the kpatch.

It will apply the patch for the other kernel if you "enable" kpatch.

martinpitt:
What I meant is, kpatches are only relevant for the running kernel, *ever*. Any kpatch should also be delivered as a proper kernel package update, and on next boot, the fixed kernel will be loaded straight away.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,325 @@
...
+                        loaded: true,
+                        unavailable: d.unavailable_names || [],
+                        missing: d.missing_names || [],
+                    });
+
+                    if ((d.unavailable_names || []).length === 0 && (d.missing_names || []).length === 0)
+                        this.checkSetup();
+                })
+                .catch(e => {
+                    console.log("Failed to query info about availablity of kpatch");
```
garrett, line 79:
```suggestion
                    console.log("Could not determine kpatch availability.");
```

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,325 @@
...
+            state = _("Disabled");
+            actionText = _("Enable");
+        }
+
+        const details = [];
+        if (this.state.enabled !== this.state.futureEnabled) {
+            if (this.state.futureEnabled)
+                details.push(_("kpatch.service will be enabled. After each reboot patches will be applied."));
+            else
+                details.push(_("kpatch.service will be disabled. After reboot no patches will be applied."));
```
garrett, line 170:
```suggestion
                details.push(_("The kpatch service will be enabled. Patches will be applied during every boot."));
            else
                details.push(_("The kpatch service will be disabled. No patches will be applied after a reboot."));
```

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,325 @@
...
+        }
+        // FIXME - should we uninstall current patches as well?
+        if (this.state.auto !== this.state.futureAuto) {
+            if (this.state.futureAuto)
+                details.push(_("Current kernel patches will be installed. Also when a new kernel is available, patches will be automatically installed as well."));
+            else
+                details.push(_("For future kernel versions patches won't be installed."));
```
garrett, line 177:
```suggestion
                details.push(_("Available kernel patches will be applied. Patches for new kernel versions will automatically applied."));
            else
                details.push(_("Patches for future kernel versions will not be applied."));
```

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,325 @@
...
+                                isSmall
+                                disabled={!this.props.privileged}
+                                onClick={action}>
+                            {actionText}
+                        </Button>
+                    </Flex>
+                </Flex>
+            </div>
+            <Modal position="top" variant="small" id="kpatch-setup" isOpen={this.state.showModal}
+                title={_("Kernel live patching settings")}
```
garrett, line 220:
```suggestion
                title={_("Kernel live patch settings")}
```

```suggestion
                title={_("Kernel patch settings")}
```

What are we calling this? Kernel patches, kernel live patches, or what? We should be consistent.

pkg/packagekit/kpatch.jsx
garrett: There's a middle ground: We could add a note in the settings modal that the package is still installed, but not used.

garrett (changes requested):
What's this weird line?

121004308-d0f60680-c78e-11eb-979c-dcf6f3ae36ba-weirdline

I'm not a fan of the bullet, nor of the text.

121004308-d0f60680-c78e-11eb-979c-dcf6f3ae36ba-bullet

I'll suggest changes in-line.

pkg/packagekit/kpatch.jsx
garrett: There's a middle ground: We could add text to the modal about the package being installed, but unused.

marusak:
> What's this weird line?

My browser acting up when taking screenshot :)

> I'm not a fan of the bullet, nor of the text.

Should there be no bullets? There might be between 0 up to two separate notes that will likely wrap across two lines.

garrett:
> Should there be no bullets? There might be between 0 up to two separate notes that will likely wrap across two lines.

Why not just `<p>`?

pkg/packagekit/kpatch.jsx
marusak: "Available kernel patches will be applied. Patches for new kernel versions will automatically applied." - doesn't that miss a verb? "will automatically apply" or "will automatically be applied"?

pkg/packagekit/kpatch.jsx
marusak: Ok, lets drop "live" from everywhere.

marusak:
> Why not just `<p>`?

Then looks like this: 
Screenshot from 2021-06-08 11-59-03changes

To me it looks better with points, but :man_shrugging:

garrett:
Looking at those two sentences... would helper text make sense? It's used for other form elements. However, I don't see a switch example with it.

marusak:
> Looking at those two sentences... would helper text make sense? It's used for other form elements. However, I don't see a switch example with it.

You mean to put the explanation underneath each switch? The issue is, that if you don't toggle the switch, it is not present, so after toggling the first switch, the second one would jump a bit lower.

garrett:
> The issue is, that if you don't toggle the switch, it is not present, so after toggling the first switch, the second one would jump a bit lower.

The second one doesn't exist if the first isn't toggled, right?

BTW: "Subscribe to" not "Subscribe**d** to". And we don't want to have a switch say "is enabled" (first one). So we'll need to rewrite those accordingly. Apologies for not catching that earlier.

Also, the text is quite redundant with the switches. The switch itself says if something is enabled or not. We don't need to restate that, right?

So, I'm saying we either need to cut way back on the (redundant) text, like this:

kernel-patch-modal-simple-disabled


kernel-patch-modal-simple

Or split it into sections like so:

kernel-patch-modal-sections-disabled

kernel-patch-modal-sections

---

Excalidraw source: **[🠟 kernel-patch-modal.zip](https://github.com/cockpit-project/cockpit/files/6617147/kernel-patch-modal.zip)**

marusak:
> The second one doesn't exist if the first isn't toggled, right?

No, they are not dependent. One is `systemctl enable kpatch` the other one is `dnf kpatch auto`. One is about installing patches, one about applying them.

Right, the new design makes sense, my concern is that when you toggle switch, it is not obvious if it actually already happened or if you need to click the submit button for it to happen. Nor it shows the current state (once you start toggling them on and off).

Right now there is a list of actions, that will happen if you click on submit button.
Screenshot from 2021-06-09 10-05-12default

click on first and on bottom it will say what is going to happen when you submit it:
Screenshot from 2021-06-09 10-05-28ff

toggle second one and you get also another note:
Screenshot from 2021-06-09 10-05-45ff

If you then get them back to their current state, the messages will disappear. Equally if currently would be both on and then I would open the dialog and toggle both of them off I get different messages:
Screenshot from 2021-06-09 10-12-51bopthoff

garrett:
> No, they are not dependent. One is systemctl enable kpatch the other one is dnf kpatch auto. One is about installing patches, one about applying them.

I don't understand.

1. What happens when you have the kernel patch service enabled and you don't subscribe?
2. What happens when the kernel patch service is disabled and you subscribe?

And when would you want to do either scenario # 1 or 2?

marusak:
> I don't understand.

If you enable service, you are saying "When you boot this machine also load patches for the kernel that you are booting into". If you subscribe, you are saying "When there is a new kernel and you `dnf` install it, also check if there are patches for this kernel and install them as well". So two unrelated things.

> And when would you want to do either scenario # 1 or 2?

There might be specific scenarios, like that you don't want to install patches for new kernels, because you only care about this current one. Or possibly that you don't want to auto load patches when booting this kernel, but you want them to be installed for future ones, as you might need them for that new kernel. And then you would also enable auto loading. (we also might want to have 'Load' button, not yet implemented).

garrett:
So... basically...

1. Apply kernel patches on boot
2. Download and apply new patches during updates

...is this right?

marusak:
> Apply kernel patches on boot

yes

> Download and apply new patches during updates

yes. It is mainly for new kernel patches, but once you toggle this on, you will get for the current kernel patches updates as well.

garrett:
Are these settings applied immediately or when someone clicks "Save" on the dialog? If the latter, then they should probably be checkboxes.

https://www.patternfly.org/v4/components/switch/design-guidelines#when-to-use-switches-versus-checkboxes:
> Use checkboxes when options do not save automatically and require the user to perform an additional action to save changes (in this case, pressing the “Save changes” button).

So the dialog could look like this:

---

### Kernel patch settings
☑ Apply kernel patches on boot
☑ Download and apply new patches

Kernel patches will be applied now, during updates, and at boot.

(Save) Cancel

---

So a summary string could be like this:
- Kernel patches will not be applied
- Kernel patches will be applied now, during updates, and at boot.
- Kernel patches will be applied now and at boot.
- Kernel patches will be applied at boot.
- Kernel patches will be applied now.

I wish we already had helpertext; but it's coming soon:
https://patternfly-pr-4140.surge.sh/components/helper-text#icon

garrett:
@marusak and I talked in IRC about this further. I think we're leaning toward something like the following?

---

### Kernel patch settings
☑ Apply kernel patches for current kernel (5.12.9-300.x86_64)
☑ Automatically apply kernel patches for future kernels

(Save) Cancel

---

And the second option would depend on the first. (That is, it's only possible to enable when the first is.)

Alternatively, this might be better communicated as:

---

### Kernel patch settings
☑ Apply kernel patches
     🞅 for current kernel only (5.12.9-300.x86_64)
     🞊 for current and future kernels

(Save) Cancel

---

marusak:
We agreed on the "alternative" approach.
Implementation note - "for current and future kernels" would do "dnf kpatch auto" but we need to figure patches package name for the first radio button on our own (or do `auto` and once it finishes switch it to `manual`?). And in both cases do `systemctl enable kpatch`. Also when un-checking the checkbox, we need let user know, that it will only be applicable from another boot, that we don't unload currently loaded patches (if there are any currently loaded).

martinpitt:
This was rather stale and had lots of red, so I pushed a clean rebase (no conflicts).

martinpitt:
I tested the current PR interactively, and I had much the same thoughts as @garrett above -- the two current options don't make too much sense right now. I still have one conceptual difficulty beyond this, though:

The current PR says "The kpatch service will be enabled. Patches will be applied during every boot.". I know that the text is being changed, and the "( ) current only" vs. "( ) now and all future kernels" makes much more sense. But conceptually, the whole point of a kpatch is that it does **not** happen during boot, but right away. If you *can* afford a boot, there would not be any purpose -- you just install the updated regular kernel package, and you're done.

I also favor the second alternative (radio buttons) over the first (check boxes) -- IMHO it's too confusing to say "enable this for all future kernels, but not the current one".

Admittedly I have some serious doubts about the concept of `dnf kpatch auto` -- this ought to have been a metapackage install just like `kernel` itself, and just *be* a normal package (with appropriate tight dependencies to the actual kernel-*version* package). I suppose this ship has sailed, but we don't need to push that into people's faces indeed.

The current test is [not idempotent](https://logs.cockpit-project.org/logs/pull-15899-20210616-074558-7c80e048-rhel-8-5/log.html#275-1). That's easy to fix, and I'll push a partial fix. However, this fails to actually enable kpatch.service, and the test is broken (I'll mention that in the detailed code review).

I'm happy to pick this up after my PTO next week.

test/verify/check-packagekit
```diff
@@ -136,6 +138,66 @@ class TestUpdates(NoSubManCase):
...
+                           provides="kpatch-patch = {0}".format(kernel_ver_arch),
+                           content={"/usr/local/bin/kpatch": kpatch},
+                           postinst="chmod +x /usr/local/bin/kpatch; rm /usr/sbin/kpatch; ln -s /usr/local/bin/kpatch /usr/sbin/kpatch")
+        self.enableRepo()
+
+        b.click("#status .pf-c-card__actions button")
+        b.wait_visible(".pf-c-badge:contains('patches')")
+        b.click("button:contains('Install all updates')")
+        b.click("button:contains('Continue')")
+        b.wait_in_text("#status", "Kernel patch kpatch_3_10_0_1062_1_1 is active")
```
martinpitt, line 199: This should also check disabling, and also individually enabling just one option. The design of that changed a lot, so probably makes sense to reimplement the design first.

test/verify/check-packagekit
```diff
@@ -136,6 +138,66 @@ class TestUpdates(NoSubManCase):
...
+
```
martinpitt, line 200: Needs another (destructive) test case to check installing kpatch package on demand.

martinpitt (changes requested):
Some initial coarse review, mostly for my future self in a week.

test/verify/check-packagekit
```diff
@@ -136,6 +138,66 @@ class TestUpdates(NoSubManCase):
...
+        sanitized_kernel_ver = "_".join(kernel_ver_arch.split(".")[:-2])
+        self.createPackage("kpatch-patch-" + sanitized_kernel_ver, "0", "0", arch=self.secondary_arch,
+                           provides="kpatch-patch = {0}".format(kernel_ver_arch))
+        self.enableRepo()
+
+        m.start_cockpit()
+        b.login_and_go("/updates")
+
+        # Enable and install kpatch
+        b.wait_in_text("#kpatch-settings", "Disabled")
```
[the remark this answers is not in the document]
martinpitt: Fixed.

test/verify/check-packagekit
```diff
@@ -136,6 +138,66 @@ class TestUpdates(NoSubManCase):
...
+        b.click(".pf-l-stack__item:nth-child(2) .pf-c-switch__input")
+        b.wait_in_text("#kpatch-setup", "Available kernel patches will be applied")
+        b.wait_in_text("#kpatch-setup", "The kpatch service will be enabled")
+        b.wait_visible(".pf-l-stack__item:nth-child(2) .pf-c-switch__input:checked")
+
+        b.click("button:contains('Apply')")
+        b.wait_not_present("#kpatch-setup")
+
+        b.wait(lambda: "True" in m.execute("grep autoupdate /etc/dnf/plugins/kpatch.conf"))
+        b.wait(lambda: "active" in m.execute("systemctl is-active kpatch || true"))
```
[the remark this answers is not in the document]
martinpitt: Fixed.

martinpitt:
Fixed a few easy things. The tests will fail now.

garrett:
Here's the latest mockup:

kpatch-settings

It's basically what I wrote above, but I swapped the two options, so kpatch is applied for current and future kernels by default, with the second item being only for the current kernel.

When "apply kernel patches" is unchecked, it disables the options.

kpatch-settings-disabled

Source: [**🠫 kpatch-settings.excalidraw.zip**](https://github.com/cockpit-project/cockpit/files/6701632/kpatch-settings.excalidraw.zip)

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,323 @@
...
+        this.kpatchService.addEventListener('changed', () => {
+            this.setState({ enabled: this.kpatchService.enabled, futureEnabled: this.kpatchService.enabled });
+        });
+    }
+
+    handleService(enable) {
+        if (enable)
+            return this.kpatchService.enable();
+        else
+            return this.kpatchService.disable();
```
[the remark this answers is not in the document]
marusak: should it? If the service is running does not do much of a difference. It is about enabling/disabling so the next boot behaves as expected.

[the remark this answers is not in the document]
martinpitt: If it's currently running and you disable it, stopping it is mostly a cleanup for being consistent (and validating that nothing more depends on that service). It's more important for the "enable" case, when you want to use it right away, don't you?

marusak:
There is one tricky part - when unchecking "Apply kernel patches" one would expect that they would stop being applied for the current kernel as well. But unloading of patches is not yet supported. We should somehow tell the user, that they will be unloaded only after reboot. Of course, we can disable and stop the service. I am now thinking about alert that would state that unloading is not supported.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                    this.setState({
+                        loaded: true,
+                        unavailable: d.unavailable_names || [],
+                        missing: d.missing_names || [],
+                    });
+
+                    this.checkSetup();
+                })
+                .catch(e => {
+                    console.log("Could not determine kpatch availability.");
```
martinpitt, line 80: Let's print the error message as well for easier debugging. Also you can drop the { } wrapper.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                });
+            });
+        });
+    }
+
+    checkSetup() {
+        cockpit.file("/etc/dnf/plugins/kpatch.conf").read()
+                .then(data => {
+                    if (data) {
+                        const auto = /autoupdate\s*=\s*True/.test(data);
```
martinpitt, line 99: is `true` allowed as well? (use the `i` option?)

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                    });
+
+                    this.checkSetup();
+                })
+                .catch(e => {
+                    console.log("Could not determine kpatch availability.");
+                });
+
+        this.kpatchService.addEventListener('changed', () => {
+            this.setState((state, _) => {
```
martinpitt, line 84: `(state, _)` → `state`

I'm curious: What does the functional setState() do differently than the plain one? (You also have access to the current state there)

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                    }
+                })
+                .catch(() => true); // Ignore errors, most likely just does not exist
+
+        cockpit.spawn(["uname", "-r"])
+                .then(data => {
+                    const fields = data.split(".");
+                    const sanitized_kernel_ver = fields.slice(0, fields.length - 2)
+                            .join("_");
+                    const patch_name = "kpatch-patch-" + sanitized_kernel_ver;
```
martinpitt, line 117: it's not that the full version is "insane", it just includes the distro revision. I suggest to first split it by `-` (upstream vs. downstream), and then handle the upstream version. I am rather surprised that kpatch-patch-* package names don't include the distro revision, though -- these often have hundreds of patches which are relevant to the binary diffs.

Is that naming pattern something which we can rely on? That feels like a pretty big assumption. This is something which I complained about in the earliest design phase already -- there should be some metapackage which pulls in the kpatch-patches-version ones, or the `dnf kpatch` command should do that. We don't want admins having to fish around for such package version numbers. Cockpit may be able to figure these out (until they change the naming pattern), but humans?

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                                        patchUnavailable: unavailable,
+                                        justCurrent: current && !state.auto,
+                                        applyCheckbox: current,
+                                        kernelName: data,
+                                    });
+                                });
+                            })
+                            .catch(() => true);
+                })
+                .catch(() => true);
```
martinpitt, line 136: Let's do a console.error() here. If uname -r fails, we really want to know.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    handleInstall() {
+        this.setState({ updating: true });
+        install_dialog(this.state.missing)
+                .then(() => {
+                    this.setState({ missing: [], updating: false });
+                })
```
martinpitt, line 144: drop braces

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                .catch(() => this.setState({ updating: false }));
```
martinpitt, line 145: This is mostly for the "cancel button" case IIRC, right? Package install failure errors are handled/displayed internally by install_dialog().

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    onClose() {
+        this.setState((state, _) => {
+            const current = state.enabled && (state.patchInstalled || state.patchUnavailable);
```
martinpitt, line 150: I've seen this "enabled && (installed || unavailable)" pattern at least four times now. Little helper property?

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    handleChange() {
+        this.setState({ updating: true });
+
+        if (this.state.applyCheckbox) {
+            let install = Promise.resolve(true);
+            if (this.state.justCurrent) {
+                install = new Promise((resolve, reject) => {
+                    cockpit.spawn(["dnf", "-y", "kpatch", "manual"], { superuser: "require", err: "message" })
```
martinpitt, line 167: That's what I meant above: Why can't the `dnf kpatch` command find and (un)install the relevant kpatch-unintelligiblename rpm? Is there a bugzilla for this? This is really unfriendly for humans and brittle for cockpit.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                            .then(() => {
+                                if (!this.state.patchUnavailable && !this.state.patchInstalled)
+                                    cockpit.spawn(["dnf", "-y", "install", this.state.patchName], { superuser: "require", err: "message" }).then(resolve)
+                                            .catch(reject);
```
martinpitt, line 171: Do we really need this `new Promise()` wrapper? simply chaining the dnf kpatch and dnf install promises should DTRT?

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                install = cockpit.spawn(["dnf", "-y", "kpatch", "auto"], { superuser: "require", err: "message" });
+            }
+            install.then(() => {
+                this.kpatchService.enable().then(() => {
+                    this.kpatchService.start().then(() => {
+                        this.setState({ showModal: false, error: "" });
+                    });
+                });
+            })
+                    .catch(e => this.setState({ error: e.toString() }))
```
martinpitt, line 187: Please put the .then in line 180 on the next line and indent it; otherwise this is really odd to read and associate to the corresponding then.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                                else
+                                    resolve();
+                            })
+                            .catch(reject);
+                });
+            } else {
+                install = cockpit.spawn(["dnf", "-y", "kpatch", "auto"], { superuser: "require", err: "message" });
+            }
+            install.then(() => {
+                this.kpatchService.enable().then(() => {
```
martinpitt, line 181: Please drop the { } wrappers or `return` the next promise. Like this it ignores failures of the inner promise as they are not properly chained. (Same below in the "else" branch)

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+
+        // Not supported on this system
+        if (this.state.unavailable.length > 0)
+            return null;
+
+        let state = _("Enabled");
+        let actionText = _("Edit");
+        let action = () => this.setState({ showModal: true });
+
+        if (this.state.missing.length > 0 && this.state.unavailable.length === 0) {
```
martinpitt, line 221: At this point we already know that this.state.unavailable.length == 0. I think it makes sense to exit early (as in line 214).

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+            actionText = _("Install");
+            action = this.handleInstall;
+        } else if (!this.state.enabled) {
+            state = _("Disabled");
+            actionText = _("Enable");
+        }
+
+        let kernel_name = "";
+        if (this.state.kernelName)
+            kernel_name = " (" + this.state.kernelName + ")";
```
martinpitt, line 232: General code style suggestion: Make this const and use ?: . Similar to `error` below.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+        if (this.state.error)
+            error = <Alert variant='danger' isInline title={this.state.error} />;
+
+        const body = <Checkbox id="apply-kpatch"
+                               isChecked={this.state.applyCheckbox}
+                               label={_("Apply kernel patches")}
+                               onChange={checked => this.setState({ applyCheckbox: checked })}
+                               body={<>
+                                   <Radio id="current-future" label={_("for current and future kernels")} onChange={() => this.setState({ justCurrent: false })} isDisabled={!this.state.applyCheckbox} isChecked={!this.state.justCurrent} />
+                                   <Radio id="current-only" label={_("for current kernel only") + kernel_name} onChange={() => this.setState({ justCurrent: true })} isDisabled={!this.state.applyCheckbox} isChecked={this.state.justCurrent} />
```
martinpitt, line 244: Do you mind breaking these long lines and put every property  on a single line? Easier to read.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                            <b>{_("Kernel patching")}</b>
+                        </FlexItem>
+                        <FlexItem>
+                            {state}
+                        </FlexItem>
+                    </Flex>
+                    <Flex>
+                        <Button variant="secondary"
+                                isSmall
+                                disabled={!this.props.privileged}
```
martinpitt, line 262: I thought `Button` requires `isDisabled` -- the button does not actually get disabled when dropping privvs. Also, in that case shouldn't this be wrapped into a Tooltip, to explain why it is disabled?

I tried to open the page in limited access mode (or reload after dropping privs), and just get a "log: Could not determine kpatch availability.", and the button does not appear at all. That is okay I think, it's just inconsistent with how the page looks like after dropping privs. (However, not a biggie).

The "Automatic updates" option behaves correctly.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                                onClick={action}>
+                            {actionText}
+                        </Button>
+                    </Flex>
+                </Flex>
+            </div>
+            <Modal position="top" variant="small" id="kpatch-setup" isOpen={this.state.showModal}
```
martinpitt, line 269: Does it make sense to even render the modal on `!showModal`? I thought our usual pattern was `{ show && <Modal isOpen ...> }`. Like this it would be okay as well, I just seem to remember that sometimes we ran into weird interactions with form submits.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+        this.checkSetup = this.checkSetup.bind(this);
+    }
+
+    checkSetup() {
+        cockpit.spawn(["kpatch", "list"], { superuser: "try", err: "ignore" })
+                .then(m => {
+                    const parts = m.trim().split("\n\n");
+                    if (parts.length !== 2 ||
+                        !parts[0].startsWith("Loaded patch modules:") ||
+                        !parts[1].startsWith("Installed patch modules:")) {
```
martinpitt, line 320: Ugh.. Let's please at least call kpatch with `LC_MESSAGES=C.UTF-8` then, to not stumble over translations. Do our integration tests catch this? (As a lot of them use the mocked `kpatch list`)

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                        console.warn("Unexpected output from `kpatch list`");
```
martinpitt, line 321: Let's show the output then, to make debugging easier.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+                .catch(() => true); // Ignore errors
+    }
+
+    componentDidMount() {
+        this.checkSetup();
+    }
+
+    componentWillUnmount() {
+        if (this.watch)
+            this.watch.close();
```
martinpitt, line 342: Reset `this.watch = null` here?

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+
+export class KpatchStatus extends React.Component {
+    constructor() {
+        super();
+
+        this.state = {
+            loaded: [],
+            installed: [],
+            changelog: null, // FIXME - load changelog
+            opened: false,
```
martinpitt, line 308: Not used by anything.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,376 @@
...
+        };
+
+        this.checkSetup = this.checkSetup.bind(this);
```
martinpitt, line 311: Not necessary. You could even just move the whole contents into componentDidMount() for simplicity.

test/verify/check-packagekit
```diff
@@ -136,6 +138,126 @@ class TestUpdates(NoSubManCase):
         else:
             self.fail("Timed out waiting for updates spinner to go away")
 
+    @skipImage("kpatch is not available", *OSesWithoutKpatch)
+    def testKpatch(self):
+        b = self.browser
+        m = self.machine
+
+        kernel_ver_arch = m.execute("uname -r").strip()
+        sanitized_kernel_ver = "_".join(kernel_ver_arch.split(".")[:-2])
```
martinpitt, line 147: This mirrors the code, see my comment about it above.

test/verify/check-packagekit
```diff
@@ -136,6 +138,126 @@ class TestUpdates(NoSubManCase):
...
+        self.createPackage("kpatch-patch-" + sanitized_kernel_ver, "0", "0", arch=self.secondary_arch,
+                           provides="kpatch-patch = {0}".format(kernel_ver_arch))
```
martinpitt, line 149: Could this also add the Requires: that the real package has, to ensure this has accurate dependencies?

test/verify/check-packagekit
```diff
@@ -136,6 +138,126 @@ class TestUpdates(NoSubManCase):
...
+
+        self.assertIn("True", m.execute("grep autoupdate /etc/dnf/plugins/kpatch.conf"))
+        self.assertEqual(m.execute("systemctl is-enabled kpatch").strip(), "enabled")
+        self.assertEqual(m.execute("systemctl is-active kpatch").strip(), "active")
+        # Patches should be installed
+        m.execute("rpm -q kpatch-patch-" + sanitized_kernel_ver)
+
+        # Faking real patches is really hard, so fake `kpatch list` command
+        # To make sure it has the same structure as we expect it to be first check the real command
+        self.assertEqual(m.execute("kpatch list"), "Loaded patch modules:\n\nInstalled patch modules:\n")
```
martinpitt, line 178: Ah good, so this will tell us when the strings change.

test/verify/check-packagekit
```diff
@@ -136,6 +138,126 @@ class TestUpdates(NoSubManCase):
...
+        b.wait_visible("#apply-kpatch:not(:checked)")
+
+        b.click("button:contains('Apply')")
+        b.wait_not_present("#kpatch-setup")
+
+        b.wait_in_text("#kpatch-settings", "Disabled")
+        b.click("#kpatch-settings button:contains('Enable')")
+
+        # Test closing of the dialog and resetting of changes
+        time.sleep(1) # Re-reading conf takes a second and bots are too fast
```
martinpitt, line 225: What does this actually wait for? This needs to go before landing.

test/verify/check-packagekit
```diff
@@ -153,8 +275,8 @@ class TestUpdates(NoSubManCase):
         # no updates on the Software Updates page
         b.go("/updates")
         b.enter_page("/updates")
-        b.wait_not_present("#status")
-        b.wait_in_text("#status", "System is up to date")
+        with b.wait_timeout(300):
+            b.wait_in_text("#status", "System is up to date")
```
martinpitt, line 279: Is that a related change? What suddenly takes so long here, all the package updates should be mocked and thus pkcon refresh etc. should be super fast?

martinpitt (changes requested):
Thanks! Sorry for the large number of comments, some of them are just questions or don't need an actual change. I must say, I'm not impressed with how kpatches are packaged.. Using them on the CLI is .. not really intuitive and robust for admins.

pkg/packagekit/kpatch.jsx
marusak: It is safer when you work with the current state in `setState` - " If the next state depends on the current state, we recommend using the updater function form" from https://reactjs.org/docs/react-component.html#setstate

So I take it as a rule of thumb, that if I use value from state in setState, then I use this updater callback form.

pkg/packagekit/kpatch.jsx
marusak: https://github.com/dynup/kpatch-dnf/blob/master/kpatch.py#L50 - we can use this same pattern then.

> or the dnf kpatch command should do that

It does it, but only when you do `dnf kpatch auto`. But that also enables future updates. This was the biggest change with the new design. Maybe if we could teach `dnf kpatch` `install` command that would only install patch but not subscribe to patches for future kernels it would simplify things a lot. Also that should be a simple patch for kpatch-dnf.

pkg/packagekit/kpatch.jsx
marusak: mostly for parity with any other .then/.catch patterns. Cannot hurt I guess? But errors are indeed displayed directly in the dialog.

martinpitt: I just checked pkg/lib/cockpit-components-install-dialog.jsx and the promise just rejects on cancel, this is correct.

pkg/packagekit/kpatch.jsx
marusak: https://github.com/dynup/kpatch-dnf/pull/8

marusak: Actually, if we are going to delegate more things to kpatch-dnf, lets also move there logic for recognizing if there is something to be installed in the first place.

pkg/packagekit/kpatch.jsx
marusak: This is part of https://github.com/dynup/kpatch-dnf/pull/8 - adding comment into code

pkg/packagekit/kpatch.jsx
marusak: > At this point we already know that this.state.unavailable.length == 0

correct, I can just drop the second part of the check.

> I think it makes sense to exit early (as in line 214).

Exit early? We need to render the "Install" button.

pkg/packagekit/kpatch.jsx
marusak: We are rather inconsistent about this. I am not aware of any issue when using this pattern and it feels a bit cleaner than having `{this.state.showModal && <Modal` pattern.

pkg/packagekit/kpatch.jsx
martinpitt: I meant that we keep the condition in 214 (the "early exit" code path).

test/verify/check-packagekit
marusak:
```
Requires:	kpatch >= 0.6.1-1
Requires:	kernel = %{kernel_ver}
```
The real package has these two requires. Not sure if that helps us in any way.

pkg/packagekit/kpatch.jsx
marusak: in `then()` there might or might not be a next promise - see the if/else. Honestly not sure how it would behave in the else-case, but this approach seems more clear to me.

pkg/packagekit/kpatch.jsx
marusak: Right, so the thing seems to be:
1. If you just drop privileges, the button is disabled
2. If you re-load page, it fails to read if needed packages are installed. Using `check_missing_packages` to also know, if these packages are even available.

marusak:
I think that [ubuntu-2004](https://logs.cockpit-project.org/logs/pull-15899-20210729-105755-41149761-ubuntu-2004/log.html#273) is unrelated, but looking.

martinpitt:
@marusak: The [resolved crash](https://logs.cockpit-project.org/logs/pull-15899-20210729-122621-41149761-ubuntu-2004/log.html#47) is certainly unrelated. I botched the fix yesterday, and sent https://github.com/cockpit-project/bots/pull/2256 to fix it. The first two look like weird infra flakes, the TestUpdates one smells familiar -- I fixed something similar two days ago in #16151. That happened because the initial state was constructed so that it would not immediately show the empty state, but there was a split second where it would show the main screen. Maybe that PR reintroduces that? I can have a closer look tomorrow.

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,371 @@
...
+                .catch(() => true); // Ignore errors, most likely just does not exist
+
+        const uname_promise = cockpit.spawn(["uname", "-r"])
+                .then(data => {
+                    const fields = data.split("-");
+                    const kpp_kernel_version = fields[0].replaceAll(".", "_");
+                    let release = fields[1].split(".");
+                    release = release.slice(0, release.length - 2); // remove el8.x86_64
+                    const kpp_kernel_release = release.join("_");
+                    const patch_name = ["kpatch-patch", kpp_kernel_version, kpp_kernel_release].join("-");
```
martinpitt, line 126: To confirm, this will also go away in favor of `dnf kpatch status`, right?

pkg/packagekit/kpatch.jsx
```diff
@@ -0,0 +1,371 @@
...
+                                        patchName: patch_name,
+                                        patchInstalled: installed,
+                                        patchUnavailable: unavailable,
+                                        justCurrent: current && !state.auto,
+                                        applyCheckbox: current,
+                                    });
+                                })
+                            );
+                })
+                .catch(console.error);
```
martinpitt, line 144: Does that actually work? I.e. do you get a readable string instead of '[Object]'? Might be better to drop this and leave it to the catch handler in componentDidMount to log it?

test/verify/check-packagekit
martinpitt: Mostly as a validation that the name computation is correct, and that we are not  trying to uninstall `kpatch` while there are still some kpatches installed.

test/verify/check-packagekit
```diff
@@ -136,6 +138,131 @@ class TestUpdates(NoSubManCase):
...
+        # Faking real patches is really hard, so fake `kpatch list` command
+        # To make sure it has the same structure as we expect it to be first check the real command
+        self.assertEqual(m.execute("kpatch list"), "Loaded patch modules:\n\nInstalled patch modules:\n")
+
+        kpatch = """
+#!/bin/bash
+echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled patch modules:\nkpatch_3_10_0_1062_1_1 (3.10.0-1062.el7.x86_64)"
+"""
+
+        self.createPackage("kpatch-patch-" + sanitized_kernel_ver, "1", "0", arch=self.secondary_arch,
```
martinpitt, line 190: Can the test also create a non-current kpatch-patch-* version, and ensure that it does *not* get listed? This is an outstanding question from the description: "This one is probably something we don't want. I run different kernel, but I have patches for different kernel installed. I guess I should not list it?"

test/verify/check-packagekit
```diff
@@ -949,6 +1075,35 @@ class TestWsUpdate(NoSubManCase):
 
         self.allow_restart_journal_messages()
 
+@skipImage("Image uses OSTree", *OSesWithoutKpatch)
```
martinpitt, line 1078: Not quite :grin:

test/verify/check-packagekit
```diff
@@ -949,6 +1075,35 @@ class TestWsUpdate(NoSubManCase):
...
+        self.createPackage("kpatch-dnf", "999", "1", content={"/etc/dnf/plugins/kpatch.conf": ""})
+        self.enableRepo()
+
+        b.reload()
+        b.enter_page("/updates")
+
+        b.wait_in_text("#kpatch-settings", "Not installed")
+        b.click("#kpatch-settings button:contains('Install')")
+        b.wait_in_text("#dialog", "kpatch, kpatch-dnf will be installed")
+        b.click("#dialog button:contains('Install')")
```
martinpitt, line 1103: Check that the packages actually got installed

martinpitt (changes requested):
Thank you! There are still 3 outstanding issues from the previous round, I marked the others as resolved. Some remaining minor stuff here -- I think the biggest issue is settling and testing the " I run different kernel, but I have patches for different kernel installed. I guess I should not list it?" question.

pkg/packagekit/kpatch.jsx
marusak: I moved this already into `componentDidMount`

pkg/packagekit/kpatch.jsx
marusak: Indeed, the whole `checkSetup()` should be just call to `dnf kpatch status`.

pkg/packagekit/kpatch.jsx
marusak: > Does that actually work? I.e. do you get a readable string instead of '[Object]'? 

Of course, you get an object, that you can expand.

> Might be better to drop this and leave it to the catch handler in componentDidMount to log it?

Not really. It is used in different places, not just in `componentDidMount`, but mainly it is used in `Promise.allSettled` thus it will never reject in `componentDidMount`.

test/verify/check-packagekit
marusak: > Mostly as a validation that the name computation is correct

How would that help? It is just `Requires: kernel = $(uname -r)` and hardcoded version of kpatch.

test/verify/check-packagekit
marusak: Good call. Just note that `dnf kpatch auto` installs patches for all available kernels, not just for the currently booted one. But indeed, we want to show just the current one - which we do, as we get it from `kpatch list` (and you cannot load patches for different kernel). The test is not great at this as it fakes output of this command, so installing another package is not gonna change anything.

test/verify/check-packagekit
martinpitt: The code doesn't offer to uninstall kpatch, so not important.

martinpitt (approved):
Thanks! LGTM now. Not sure if @garrett wants to do a final sign-off as well?

garrett:
Looks great overall, juding from these screenshots. (I'm not sure how to see it on my system otherwise, so I think your detailed view in the top comment is probably enough. Other than that, a screencast video could help.)

> Kpatch is installed, enabled and all the fun stuff. Also patches are available (see in the list)

Kernel patches are supposed to be the first item(s) at the top of the update list.

> This one is probably something we don't want. I run different kernel, but I have patches for different kernel installed. I guess I should not list it? (the other kernel is still available, I just don't use it right now). Should it list only installed patches for the current kernel?

Right; the status should show only active kpatch(es) for the current running kernel, not those that are installed.

garrett (changes requested):
Looks great overall.

Note: Going from the screenshots only, as I don't know how to run this on my system. (A screencast could possibly help? But this is _probably_ enough.)

Two things:

1. kernel patches should always be the first item in the updates list
2. only show active patches in the status area (that is, applies to the current running kernel)

martinpitt:
> kernel patches should always be the first item in the updates list

OOI, why? Packages should be sorted by severity first, then alphabetically. They should not be *that* special?

marusak:
> Going from the screenshots only, as I don't know how to run this on my system. (A screencast could possibly help? But this is probably enough.)

I cannot run this on my system either, I test it on rhel-8-5 machine

> kernel patches should always be the first item in the updates list
> > OOI, why? Packages should be sorted by severity first, then alphabetically. They should not be that special?

Same question. Not super convinced it should be placed on top of the list. Should we then move kernel updates as well so it is obvious there is also kernel update?

> only show active patches in the status area (that is, applies to the current running kernel)

That is the case already.

garrett:
For kernel packages:

I thought it made sense at the top as kernel patches are kind of special compared to the rest of the software on the system. (FWIW: I meant at the top of the security when they're security related and at the top of the non-security list when they're not related to security.)

But if you both think kernel patches shouldn't show at the top, then I guess we can just interleave them as-is.

As for "only show active patches in the status area (that is, applies to the current running kernel)", I was responding to "This one is probably something we don't want. I run different kernel, but I have patches for different kernel installed. I guess I should not list it? (the other kernel is still available, I just don't use it right now). Should it list only installed patches for the current kernel?"

garrett (approved):
As my points have been addressed and they're small(ish) anyway, I'm approving this.

martinpitt:
I updated the release note text and optimized the screenshots.

PR #15900, opened 2021-06-07
jumpstart / git-caching

allisonkarlitskaya:
The summary:
 - minor changes to the copyright scanning script to avoid a mess on stderr
 - add a new `git-utils.sh` which contains a common implementation of the logic for maintaining and using a cache of a git repository of a given name relative to `GITHUB_BASE`, with utilities for checking it out to various different places in the cockpit tree
 - use that from `tools/node-modules`
 - use that from `tools/make-bots`
 - ~~use that from `test/common/pixel-tests`~~
 - use that from a new script `tools/jumpstart`
 - use that last one from `test/make_dist.py`
 - only commit `dist/` and `package-lock,json` to `cockpit-dist`

### DIFF
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -51,60 +51,19 @@ def build_dist():
 
 
 def download_cache(wait=False):
-    '''Download pre-built webpack for current git SHA from GitHub
-
-    These are produced by .github/workflows/build-dist.yml for every PR and push.
-    This is a lot faster than having to npm install and run webpack.
-
-    Returns True when successful, or False if the download isn't available.
-    This can happen because dist/ already exists, or the current directory is not a git checkout,
-    or it is a SHA which is not pushed/PRed.
-    '''
-    try:
-        sha = subprocess.check_output(["git", "rev-parse", "HEAD"], stderr=subprocess.DEVNULL).decode().strip()
-    except subprocess.CalledProcessError:
-        message("make_dist: not a git repository")
-        return False
-
-    if subprocess.call(["git", "diff", "--quiet", "--", ":^test", ":^packit.yaml", ":^.github"]) > 0:
-        message("make_dist: uncommitted local changes, skipping download")
-        return False
-
-    dist_git_checkout = os.path.join(os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache")), "cockpit-dev", CACHE_REPO + ".git")
-
-    if not os.path.exists(dist_git_checkout):
-        message(f"make_dist: Creating dist cache {dist_git_checkout}")
-        subprocess.check_call(["git", "init", "--bare", "--quiet", dist_git_checkout])
-        subprocess.check_call(["git", "--git-dir", dist_git_checkout, "remote", "add", "origin", "https://github.com/" + CACHE_REPO])
-
-    tag = "sha-" + sha
-
-    retries = 50 if wait else 1  # 25 minutes, once every 30s
-    while retries > 0:
+    tries = 50 if wait else 1  # 25 minutes, once every 30s
+    for retry in range(tries):
         try:
-            subprocess.check_call(["git", "--git-dir", dist_git_checkout, "fetch", "--no-tags", "--depth=1", "origin", "tag", tag])
-            break
-        except subprocess.CalledProcessError:
-            retries -= 1
-
-            if retries == 0:
-                message(f"make_dist: Downloading pre-built dist for SHA {sha} failed")
-                return False
-
-            message(f"make_dist: pre-built dist for {sha} not yet available, waiting...")
+            subprocess.check_call(["tools/webpack-jumpstart"])
+            return True
+        except subprocess.CalledProcessError as e:
+            if e.returncode != 2 or not wait:
+                break
+            message("make_dist: pre-built dist not yet available, waiting...")
             time.sleep(30)
 
-    for unpack_path in ["node_modules", "package-lock.json", "dist", "tools/debian/copyright"]:
-        if os.path.exists(unpack_path):
-            continue
-        message(f"make_dist: Extracting cached {unpack_path}...")
-        p_git = subprocess.Popen(["git", "--git-dir", dist_git_checkout, "archive", tag, unpack_path],
-                                 stdout=subprocess.PIPE)
-        subprocess.check_call(["tar", "-x", "--touch"], stdin=p_git.stdout)
-        result = p_git.wait()
-        assert result == 0
-
-    return True
+    message("make_dist: Downloading pre-built dist failed")
+    return False
 
 
 def make_dist(download_only=False, wait_download=False):
--- a/tools/git-utils.sh
+++ b/tools/git-utils.sh
@@ -1,0 +1,110 @@
+# doesn't do anything on its own.  must be sourced.
+
+# The script which sources this script must set the following variables:
+#   GITHUB_REPO = the relative repo name of the submodule on github
+#   SUBDIR = the location in the working tree where the submodule goes
+# We also expect `set -eu`.
+[ -n "${GITHUB_REPO}" ]
+[ -n "${SUBDIR}" ]
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/${GITHUB_REPO}"
+HTTPS_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/${GITHUB_REPOSITORY}.git"
+
+if [ "${V-}" = 0 ]; then
+    message() { printf "  %-8s %s\n" "$1" "$2"; }
+    quiet='--quiet'
+else
+    message() { :; }
+    quiet=''
+fi
+
+# runs a git command on the cache dir
+git_cache() {
+    git --git-dir "${CACHE_DIR}" "$@"
+}
+
+# reads the named gitlink from the current state of the index
+# returns (ie: prints) a 40-character commit ID
+get_index_gitlink() {
+    if ! git ls-files -s "$1" | egrep -o '\<[[:xdigit:]]{40}\>'; then
+        echo "*** couldn't read gitlink for file $1 from the index" >&2
+        exit 1
+    fi
+}
+
+init_cache() {
+    if [ ! -d "${CACHE_DIR}" ]; then
+        message INIT "${CACHE_DIR}"
+        mkdir -p "${CACHE_DIR}"
+        git init --bare --template='' ${quiet} "${CACHE_DIR}"
+        git_cache remote add origin "${HTTPS_REMOTE}"
+    fi
+}
+
+# This checks if the given argument "$1" (already) exists in the repository
+# we use git fsck to to avoid problems with incomplete fetches: we want to make
+# sure the complete commit is there
+check_ref() {
+    git_cache fsck --no-dangling --connectivity-only "$1" 2>/dev/null
+}
+
+# Fetch a specific commit ID into the cache
+# Either we have this commit available locally (in which case this function
+# does nothing), or we need to fetch it.  There's no chance that the object
+# changed on the server, because we define it by its checksum.
+fetch_sha_to_cache() {
+    sha="$1"
+
+    init_cache
+    # No "offline mode" here: we either have the commit, or we don't
+    if ! check_ref "${sha}"; then
+        message FETCH "${SUBDIR}  [ref: ${sha}]"
+        git_cache fetch --no-tags ${quiet} origin "${sha}"
+        # tag it to keep it from being GC'd.
+        git_cache tag "sha-${sha}" "${sha}"
+    fi
+}
+
+# General purpose "fetch" function to be used with tags, refs, or nothing at
+# all (to fetch everything).  This checks the server for updates, because all
+# of those things might change at any given time.  Supports an "offline" mode
+# to skip the fetch and use the possibly-stale local version, if we have it.
+fetch_to_cache() {
+    # We're fetching a named ref (or all refs), which means:
+    #  - we should always do the fetch because it might have changed. but
+    #  - we might be able to skip updating in case we already have it
+    init_cache
+    if [ -z "${OFFLINE-}" ]; then
+        message FETCH "${SUBDIR}  ${1+[ref: $*]}"
+        git_cache fetch --prune ${quiet} origin "$@"
+    fi
+}
+
+# Consistency checking: for a given cache commit "$1", check if it contains a
+# file "$2" which is equal to the file "$3" present in the working tree.
+cmp_from_cache() {
+    git_cache cat-file blob "$1:$2" | cmp "$3"
+}
+
+# Like `git clone` except that it uses the original origin url and supports
+# checking out commit IDs as detached heads.  The target directory must either
+# be empty, or not exist.
+clone_from_cache() {
+    message CLONE "${SUBDIR}  [ref: $1]"
+    [ ! -e "${SUBDIR}" ] || rmdir "${SUBDIR}"
+    mkdir "${SUBDIR}"
+    cp -a --reflink=auto "${CACHE_DIR}" "${SUBDIR}/.git"
+    git --git-dir "${SUBDIR}/.git" config --unset core.bare
+    git -c advice.detachedHead=false -C "${SUBDIR}" checkout ${quiet} "$1"
+}
+
+# This copies the files without setting up the git repository.  The copied
+# files are expected to be in a same-named subdirectory inside the cache
+# repository.
+unpack_from_cache() {
+    message "UNPACK" "${SUBDIR}  [ref: $1]"
+    git_cache archive "$1" "${SUBDIR}" | tar -x --touch "${SUBDIR}"
+}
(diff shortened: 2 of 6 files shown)


tools/jumpstart
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+# Download pre-built webpack for current git SHA from GitHub
```
martinpitt, line 3: I know you like that name, but can we pretty please name it to something descriptive? download-dist, or maybe cached-webpack? (dist seems more appropriate, as that's the directory it uses).

This current split between make_dist.py and this shell script bothers me -- the remaining Python script does not do anything any more that is worth writing in Python. Is there any use case for even having two separate scripts? I'd rather just completely port make_dist to a shell script.

tools/jumpstart
```diff
@@ -0,0 +1,45 @@
...
+tag="sha-$(git rev-parse HEAD)"
+if ! fetch_to_cache tag "${tag}"; then
+    echo "There is no cache entry ${tag}" >&2
+    exit 2
+fi
+
+if ! cmp_from_cache "${tag}" "package-lock.json" "package-lock.json"; then
+    echo "The cached package-lock.json doesn't match our own" >&2
+    exit 1
+fi
```
martinpitt, line 43: This also needs the `$NODE_ENV != development` check.

martinpitt (changes requested):
Thanks! Aside from the naming question (hardest of all!) my biggest worry is about the split. The rest is small local issues which should hopefully be easy to fix.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
+# doesn't do anything on its own.  must be sourced.
+
+GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
+GITHUB_REPOSITORY="${GITHUB_BASE%/*}/${GITHUB_REPO}"
+HTTPS_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+SSH_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Thanks!  Removed.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
...
+
+CACHE_DIR="${XDG_CACHE_HOME-${HOME}/.cache}/cockpit-dev/${GITHUB_REPOSITORY}.git"
+
+if [ "${V-}" = 0 ]; then
+    message() { printf "  %-8s %s\n" "$1" "$2"; }
```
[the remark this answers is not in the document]
allisonkarlitskaya: These are the " CC  xyz.c" type messages.  Those are exactly what we want in the `V=0` case.

In the other case, we can `set +x` to get more detailed output.  Each script is handling that differently, admittedly for somewhat historical reasons.

 - node-modules: default to `+x`
 - bots: default to no output at all
 - jumpstart: default to `V=0` output

Maybe I'll make bots `V=0` by default as well, and add `set +x` for the other case....

tools/node-modules
```diff
@@ -1,16 +1,11 @@
...
+CLONE_TARGET='node_modules'
 
-message() {
-    [ "${V-}" != 0 ] || printf "  %-8s %s\n" "$1" "$2"
-}
+set -eu
+cd "$(realpath -m "$0"/../..)"
+. tools/git-utils.sh
 
 get_index_gitlink() {
```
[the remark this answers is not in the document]
allisonkarlitskaya: Thanks.  Removed.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
+# doesn't do anything on its own.  must be sourced.
```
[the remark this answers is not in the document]
allisonkarlitskaya: I did this and renamed `CLONE_TARGET` to `SUBDIR`.  This is the name that the pixel tests were using and I like it better because we're not always cloning (eg: in the dist/ case).

tools/make-bots
```diff
@@ -3,14 +3,16 @@
...
+. tools/git-utils.sh
 
 if [ ! -d bots ]; then
-    git clone --quiet --reference-if-able ${XDG_CACHE_HOME:-$HOME/.cache}/cockpit-project/bots https://github.com/cockpit-project/bots.git
-    if [ -n "$COCKPIT_BOTS_REF" ]; then
-        git -C bots fetch --quiet --depth=1 origin $COCKPIT_BOTS_REF
-        git -C bots checkout --quiet FETCH_HEAD
-    fi
+    fetch_to_cache # it's small, so keep everything cached
+    clone_from_cache "${COCKPIT_BOTS_REF-master}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html § 2.6.2:

> `${parameter:-[word]}`
>    Use Default Values. If parameter is unset or null, the expansion of word (or an empty string if word is omitted) shall be substituted; otherwise, the value of parameter shall be substituted.

...and just below...

> In the parameter expansions shown previously, use of the `<colon>` in the format shall result in a test for a parameter that is unset or null; omission of the `<colon>` shall result in a test for a parameter that is only unset.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
...
+        exit 1
+    fi
+}
+
+init_cache() {
+    if [ ! -d "${CACHE_DIR}" ]; then
+        message INIT "${CACHE_DIR}"
+        mkdir -p "${CACHE_DIR}"
+        git init --bare --template='' ${quiet} "${CACHE_DIR}"
+        git --git-dir "${CACHE_DIR}" remote add origin "${HTTPS_REMOTE}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Added a `git_cache` function.  It's a bit less easy to read, imho, but it jives nicely with my "DRY" feelings.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
...
+    if ! check_ref "${sha}"; then
+        init_cache
+        message FETCH "${CLONE_TARGET}  [ref: ${sha}]"
+        git --git-dir "${CACHE_DIR}" fetch --no-tags ${quiet} origin "${sha}"
+        # tag it to keep it from being GC'd.
+        git --git-dir "${CACHE_DIR}" tag "sha-${sha}" "${sha}"
+    fi
+}
+
+fetch_to_cache() {
```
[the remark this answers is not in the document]
allisonkarlitskaya: I added some comments to better explain the difference, but I prefer to keep the split as it is.  It makes sense to me to treat these two cases separately.  Originally, I tried to implement that as one common function, but the two cases are really different.

I want to prevent GC because in the case where we use this function (fetching commit IDs) we currently don't have plans to ever remove items from the upstream cache repository.  If we do want to implement pruning, we should do that separately.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
...
+    #  - we should always do the fetch because it might have changed. but
+    #  - we might be able to skip updating in case we already have it
+    if [ -z "${OFFLINE-}" ]; then
+        init_cache
+        message FETCH "${CLONE_TARGET}  ${1+[ref: $*]}"
+        git --git-dir "${CACHE_DIR}" fetch --prune ${quiet} origin "$@"
+    fi
+}
+
+cmp_from_cache() {
```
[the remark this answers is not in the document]
allisonkarlitskaya: Done.

tools/git-utils.sh
```diff
@@ -0,0 +1,80 @@
...
+    git --git-dir "${CACHE_DIR}" cat-file blob "$1:$2" | cmp "$3"
+}
+
+# Like `git clone` except that it uses the original origin url
+# and supports checking out commit IDs as detached heads
+clone_from_cache() {
+    message CLONE "${CLONE_TARGET}  [ref: $1]"
+    [ ! -e "${CLONE_TARGET}" ] || rmdir "${CLONE_TARGET}"
+    mkdir "${CLONE_TARGET}"
+    cp -a --reflink=auto "${CACHE_DIR}" "${CLONE_TARGET}/.git"
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's really just about the default case of the directory existing, but empty, which is what git does for submodules.

tools/jumpstart
allisonkarlitskaya: I'm trying to capture the essence of "get this worktree quickly up and running and ready for new hacking".  I don't actually care what's happening to make that happen.  I just want the first time I type "make" to be fast.

I actually also consider to fetch the bots and/or call `./autogen.sh` for you (possibly using `"$@"` for extra arguments).

allisonkarlitskaya: Let's chat about this on IRC :)

tools/jumpstart
allisonkarlitskaya: OK.

tools/git-utils.sh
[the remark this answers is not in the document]
martinpitt: Ah, I see; Sorry, I think I fell into the exact same trap in the previous version.

test/common/pixel-tests
```diff
@@ -1,19 +1,14 @@
...
 
-set -eu
+GITHUB_REPO='pixel-test-reference'
+SUBDIR='test/reference'
 
-SUBDIR=test/reference
-REPO=pixel-test-reference
+set -eu
+cd "$(realpath -m "$0"/../../..)"
+. tools/git-utils.sh
```
mvollmer, line 8: The `pixel-tests` tool is also used by external projects, and they only have `test/common`, not `tools`.  So this here will fail, no?

test/common/pixel-tests
```diff
@@ -49,20 +44,4 @@ cmd_push() {
     git add "$SUBDIR"
 }
 
-main() {
```
mvollmer, line 52: `main` is still needed, no?

test/common/pixel-tests
```diff
@@ -1,19 +1,14 @@
...
 
-SUBDIR=test/reference
-REPO=pixel-test-reference
+set -eu
+cd "$(realpath -m "$0"/../../..)"
+. tools/git-utils.sh
 
-GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
-GITHUB_REPOSITORY="${GITHUB_BASE%/*}/${REPO}"
-CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
```
mvollmer, line 10: `CLONE_REMOTE` is still used below.  It's `HTTPS_REMOTE` now, no?

test/common/pixel-tests
allisonkarlitskaya: I wasn't aware of that.  I'll drop this commit for the time being.

allisonkarlitskaya:
the pixel tests commit was a disaster, so I've dropped it.

tools/git-utils.sh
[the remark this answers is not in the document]
martinpitt: Sure, I wasn't talking about the repo on github, but the local ~/.cache repo. Fetching would/should never actually push any tag to origin, so that tagging here should at most influence the local repo?

martinpitt:
packit fails with

   tools/jumpstart: line 29: NODE_ENV: unbound variable

I'm ack'ing github-changes here, mostly want to see it being removed by the next force-push again :grin:

allisonkarlitskaya:
> I'm ack'ing github-changes here, mostly want to see it being removed by the next force-push again grin

Uh.  Oops?

martinpitt (dismissed):
The fine-grained stuff is fine now, thanks for the fixes! Now we just need to talk about the outstanding issue above (not having two scripts which do almost the same thing, and defining its purpose). "script that does something" is not good enough, sorry :grin:

allisonkarlitskaya:
So there are two separate scripts here, that actually don't have a lot to do with each other, in my opinion.

One script takes a fresh repository and pre-seeds it with some nice stuff to get you up and running as fast as possible.  It's very logically called "jumpstart", because that's what it does.

The other script builds a tarball from a repository, using the best means available (possibly including #1, but not always) and tells you the path to the tarball.  It's very logically called "make dist", because that's what it does.

Let's chat tomorrow :)

martinpitt:
How about tools/webpack-jumpstart? That aligns with tools/webpack-{make,watch} and at least says *what* it jumpstarts. Maybe that's the sort of good compromise that makes us about equally (but hopefully not too much) grumpy :grin:

allisonkarlitskaya:
I'm fine with the name, if it needs to be that way (although it pushes against my plans to also add a bots checkout).

The only real issue remaining here is what to do with `make_dist.py`.  You mentioned that packit is using that (and some quick grepping also turns up `livetest.yml` and `make-debs`/`make-srpm`), so I'm afraid we can't just merge it into `image-prepare`. I'm happy to port it to shell if you think that's nicer, and maybe move it to `tools/`.  I'm also happy to move `--wait` to the jumpstart script...

Thoughts?

allisonkarlitskaya:
As discussed, the latest changes:
 - script is now called `webpack-jumpstart`
 - drop the publish workflow changes.  will proposed those in a separate PR along with a change to use the newly-added `pkg/build` (and no container) to drive the "build" side.

martinpitt:
Adding blocked, as @allisonkarlitskaya told me on IRC to hold off on this. Also, this needs integration test runs.

tools/git-utils.sh
```diff
@@ -0,0 +1,108 @@
...
+fetch_sha_to_cache() {
+    sha="$1"
+
+    # No "offline mode" here: we either have the commit, or we don't
+    if ! check_ref "${sha}"; then
+        init_cache
+        message FETCH "${SUBDIR}  [ref: ${sha}]"
+        git_cache fetch --no-tags ${quiet} origin "${sha}"
+        # tag it to keep it from being GC'd.
+        git_cache tag "sha-${sha}" "${sha}"
```
martinpitt, line 65: Leftover from  previous conversation: It's not clear to me at all why we need this tag. If git wants to gc the local cache repo from time to time, that's actually desirable -- I see no point in actively preventing that. In fact, we should probably do something about not growing the local caches indefinitely.

Naturally on the github.com repo we want (and have) tags, and do garbage collection in a more controlled fashion. But AFAICS these local tags never get pushed to origin, they already exist there at this point?

test/make_dist.py
```diff
@@ -51,60 +51,19 @@ def build_dist():
...
-            break
-        except subprocess.CalledProcessError:
-            retries -= 1
-
-            if retries == 0:
-                message(f"make_dist: Downloading pre-built dist for SHA {sha} failed")
-                return False
-
-            message(f"make_dist: pre-built dist for {sha} not yet available, waiting...")
+            subprocess.check_call(["tools/webpack-jumpstart"])
```
martinpitt, line 57: Possible gotcha: image-prepare imports this as Python module, so I'm not sure if we can/should assume that cwd==project root here. could this compute the path relative to `__file__`?

tools/webpack-jumpstart
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
+# Download pre-built webpack for current git SHA from GitHub
```
martinpitt, line 3: I think this should say dist/. As you noticed today, there's dist/guide which isn't a webpack, and the corresponding workflow is also called build-dist, not build-webpack. I wouldn't mind to rename the script accordingly as well (dist-jumpstart, or download-dist), but I feel for this PR I've already used up my quota of name bikeshedding; I can certainly live with the current one :grin:

martinpitt (commented):
Thanks! Found a few more little issues, but nothing important -- ok for me to land as-is once you are happy. I'll trigger a few tests over night to get some results.

martinpitt:
The other test failures are known flakes, but [this one looks relevant](https://logs.cockpit-project.org/logs/pull-15900-20210609-221634-ed183349-ubuntu-2004/log.html):
```
fatal: Not a valid object name 380ae06319b3e0ebeba0b46e8117e137e3f66315:.package.json
cmp: EOF on - which is empty

*** node_modules 380ae06319b3e0ebeba0b46e8117e137e3f66315 doesn't match our package.json
*** refusing to automatically check out node_modules
```

allisonkarlitskaya:
> The other test failures are known flakes, but [this one looks relevant](https://logs.cockpit-project.org/logs/pull-15900-20210609-221634-ed183349-ubuntu-2004/log.html):

I have absolutely no idea why that would be the case, particularly considering the build, running in the exact same tasks container, is working for all of the other images.

Maybe some weird race?  Ugh.

test/make_dist.py
allisonkarlitskaya: I don't think that makes sense: in other places we do, for example
```
            subprocess.check_call('./autogen.sh')
```

tools/git-utils.sh
allisonkarlitskaya: I want to add an explicit prune mechanism at a later date.  I'm too worried about git dropping commits between commands, or if a separate parallel command is running for another workdir and it accidentally triggers a GC, for example...

allisonkarlitskaya:
> The other test failures are known flakes, but [this one looks relevant](https://logs.cockpit-

Pretty sure I figured it out, and it was cool: an interrupted (or concurrently running) `git fetch` can leave us in a state where we have the commit object present in the repository, but not the tree and file objects that it points to.

Performing a simple lookup with `rev-parse`, as we do to avoid fetching, would return "true" in that case, but we wouldn't actually be able to access the files.

I moved the check over to a full connectivity-checking fsck call instead.  That should catch this issue.  In that case we'll redo the fetch.

martinpitt (approved):
Thanks! Green, and the misleading comment above is not that important. I'm sure that we'll touch this stuff more :grin:

PR #15901, opened 2021-06-07
Move to webpack 5

KKoukiou:


webpack.config.js
```diff
@@ -345,7 +345,7 @@ function get_msggrep_options () {
 }
 
 const plugins = [
-    new IncludedModulesPlugin((section || "") + "included-modules"),
+    production ? new IncludedModulesPlugin((section || "") + "included-modules") : undefined,
```
martinpitt, line 348: I'm not sure about that -- Isn't that required for developers for correct `make` dependencies? Or does this *only* affect dependencies to `node_modules`, not to e.g. pkg/lib/?

Can you please add a rationale to the commit message?

martinpitt:
This is now a case where we need the bots to rebuild node_modules, I'll trigger it with the magic comment. After that, please squash that in (the bot sends a comment how), and re-push. Thanks @KKoukiou !

webpack.config.js
KKoukiou: Let me move this commit outside of this PR as it's not so relevant.

martinpitt:
There are a bunch of [aborted without status](https://logs.cockpit-project.org/logs/pull-15901-20210607-145145-c9a24c3a-rhel-8-5/log) test runs. The log only says

    podman[2125153]: client_loop: send disconnect: Broken pipe

I think we are just regularly breaking podman with 12 parallel workers on that machine :cry: I'll retry them.

martinpitt:
Oops, there's something else also:

    fatal: update_ref failed for ref 'ORIG_HEAD': cannot lock ref 'ORIG_HEAD': Unable to create '/work/bots/.git/ORIG_HEAD.lock': No space left on device

Indeed the 200 GB root dir is 95% full. Nearly all of that is /var/lib/containers/. There's not much else that I can do quickly than just tear down two runners. Tomorrow I'll attach some external storage to that thing.

allisonkarlitskaya:
Heads up with #15922.  If it lands first, you'll want to update the `sass-loader` version to the one that works with Webpack 5 before merging this.

martinpitt:
#15922 landed, so this now requires bumping sass-loader.

KKoukiou:
There is some drama on the debian based distro still

martinpitt:
Failure:
```
NODE_ENV=production SRCDIR=/build/cockpit-246.53.gc8a9aea95 ./tools/webpack-make -d dist/users/Makefile.deps -c ./webpack.config.js
/usr/bin/env: 'node': No such file or directory
```

This usually means that the generated tarball contains some absolute file paths. In particular, /build is only correct in the tasks container, but not in the prepared VM image (on Debian, it's /var/lib/pbuilder/... or /var/tmp). Our tarballs (in particular, Makefile.deps and friends) should never contain absolute file paths. Thus this smells like more fallout from https://github.com/webpack/webpack/issues/11971

KKoukiou:
> Failure:
> 
> ```
> NODE_ENV=production SRCDIR=/build/cockpit-246.53.gc8a9aea95 ./tools/webpack-make -d dist/users/Makefile.deps -c ./webpack.config.js
> /usr/bin/env: 'node': No such file or directory
> ```
> 
> This usually means that the generated tarball contains some absolute file paths. In particular, /build is only correct in the tasks container, but not in the prepared VM image (on Debian, it's /var/lib/pbuilder/... or /var/tmp). Our tarballs (in particular, Makefile.deps and friends) should never contain absolute file paths. Thus this smells like more fallout from [webpack/webpack#11971](https://github.com/webpack/webpack/issues/11971)

Thanks that was super useful!

martinpitt (approved):
Code changes look good, thanks for figuring out the subdirs workaround!

I think we should still do a visual inspection to make sure there are no CSS regressions.

KKoukiou:
> Code changes look good, thanks for figuring out the subdirs workaround!
> 
> I think we should still do a visual inspection to make sure there are no CSS regressions.

I will do it later today!

KKoukiou:
Went over all pages with the prebuilt dist artifact from this PR - so production build - and there are no visual regressions.

Issue #15886, opened 2021-06-03
Layout: Disks with long serial id mess up the page layout


ne20002:
Simply a screenshot:

![grafik](https://user-images.githubusercontent.com/39007279/120669758-07323e00-c490-11eb-82f4-5c6f0e011626.png)

Would it be an option to limit width of serial id?

marusak:
I think we want to just limit it with css.
When I try it out, the whole panel gets out of screen:
![Screenshot from 2021-06-04 14-36-52outofscreen](https://user-images.githubusercontent.com/12330670/120802219-78cac480-c542-11eb-9c47-15fc30802e5d.png)

ne20002:
Yes, and with the panel getting out of the screen on the right side, the warning icon as seen in my screenshot also gets out of sight.
Limitting the field displaying the serial with css should be enough. The other information in the boxes is more relevant than the serial and should not be moved outside the screen.

PR #15731, opened 2021-04-22
package.json: Update @patternfly/patternfly, @patternfly/react-styles, @patternfly/react-core, @patternfly/react-table

cockpituous:

PR #15903, opened 2021-06-07
storage: Layout fixes

garrett:
This is a bunch of little fixes of small, unrelated things for the storage page, including a fix for #15886.

I tried to use PF4 when possible, but fell back on some custom CSS changes in several places, as:
1. we're not using the latest PF4 (and cannot use the utility classes yet, so I crafted some fallbacks that should cause no harm — as they're simple and the same as PF4's classes — but the specific CSS still should be removed once we upgrade PF4)
2. in the storage page, we're also still using custom widgets for listing

### DIFF
--- a/pkg/storaged/nfs-details.jsx
+++ b/pkg/storaged/nfs-details.jsx
@@ -337,7 +337,7 @@ export class NFSDetails extends React.Component {
 
                         <DescriptionListGroup>
                             <DescriptionListTerm className="control-DescriptionListTerm">{_("Size")}</DescriptionListTerm>
-                            <DescriptionListDescription>
+                            <DescriptionListDescription className="pf-u-align-self-center">
                                 { entry.mounted
                                     ? <StorageUsageBar stats={fsys_size} critical={0.95} block={entry.fields[1]} />
                                     : "--"
--- a/pkg/storaged/side-panel.jsx
+++ b/pkg/storaged/side-panel.jsx
@@ -120,12 +120,12 @@ export class SidePanelRow extends React.Component {
                       onKeyPress={this.props.go ? go : null}
                       onClick={this.props.go ? go : null}>
                 <Flex flexWrap={{ default: 'nowrap' }}>
-                    <FlexItem grow={{ default: 'grow' }} className="sidepanel-row-name">{this.props.name}</FlexItem>
+                    <FlexItem grow={{ default: 'grow' }} className="sidepanel-row-name pf-u-text-break-word">{this.props.name}</FlexItem>
                     <FlexItem>{decoration}</FlexItem>
                 </Flex>
                 <Flex className="sidepanel-row-info">
                     <FlexItem grow={{ default: 'grow' }} className="sidepanel-row-detail">{this.props.detail}</FlexItem>
-                    <FlexItem className="sidepanel-row-devname">{this.props.devname}</FlexItem>
+                    <FlexItem className="sidepanel-row-devname pf-u-text-break-word">{this.props.devname}</FlexItem>
                 </Flex>
             </FlexItem>
         );
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -216,6 +216,10 @@ div.progress {
     min-width: 250px;
 }
 
+#storage-detail .ct-table {
+    border-top: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
+}
+
 td.usage-text {
     text-align: right;
 }
@@ -473,11 +477,7 @@ td.storage-action {
 
 .sidepanel-row {
     padding: var(--pf-global--spacer--sm) var(--pf-global--spacer--lg);
-    border-bottom: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
-}
-
-.sidepanel-row:last-child {
-    border-bottom: none;
+    border-top: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100);
 }
 
 .sidepanel-row.pf-c-empty-state {
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -28,15 +28,9 @@
     --pf-l-gallery--GridTemplateColumns: 1fr;
 }
 
-#main-storage.pf-c-page,
 #storage-detail .pf-c-page__main-section,
 #storage-detail .pf-c-page__main-nav {
-    padding: var(--pf-global--gutter);
     overflow: auto;
-}
-
-#main-storage > .pf-c-page__main {
-    padding-bottom: var(--pf-global--spacer--md);
 }
 
 #storage .create-storage {
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -19,6 +19,7 @@
 @import "../lib/page.scss";
 @import "../../node_modules/@patternfly/patternfly/components/Card/card.css";
 @import "../lib/ct-card.scss";
+@import "@patternfly/patternfly/utilities/Flex/flex.scss";
 @import "@patternfly/patternfly/utilities/Text/text.scss";
 
 #storage .pf-c-card {
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -491,7 +491,7 @@ td.storage-action {
 }
 
 .sidepanel-row-name {
-    font-size: var(--pf-global--FontSize--md);
+    font-size: var(--pf-global--FontSize--sm);
 }
 
 .sidepanel-row-info {
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -19,6 +19,7 @@
 @import "../lib/page.scss";
 @import "../../node_modules/@patternfly/patternfly/components/Card/card.css";
 @import "../lib/ct-card.scss";
+@import "@patternfly/patternfly/utilities/Text/text.scss";
 
 #storage .pf-c-card {
     @extend .ct-card;
--- a/pkg/storaged/overview.jsx
+++ b/pkg/storaged/overview.jsx
@@ -19,7 +19,7 @@
 
 import React from "react";
 
-import { Page, Grid, GridItem, Card, CardBody, Gallery } from "@patternfly/react-core";
+import { Page, PageSection, Grid, GridItem, Card, CardBody, Gallery } from "@patternfly/react-core";
 
 import { StoragePlots } from "./plot.jsx";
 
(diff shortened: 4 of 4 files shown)


pkg/storaged/nfs-details.jsx
```diff
@@ -337,7 +337,7 @@ export class NFSDetails extends React.Component {
 
                         <DescriptionListGroup>
                             <DescriptionListTerm className="control-DescriptionListTerm">{_("Size")}</DescriptionListTerm>
-                            <DescriptionListDescription>
+                            <DescriptionListDescription className="storage-align-self-center pf-u-align-self-center">
```
KKoukiou, line 340: pf-u-align-self-center is exactly the same like the storage-align-self-center class. Why do you add both? The PF class seems to exist upstream.

pkg/storaged/storage.scss
```diff
@@ -31,12 +31,11 @@
 #main-storage.pf-c-page,
 #storage-detail .pf-c-page__main-section,
 #storage-detail .pf-c-page__main-nav {
-    padding: var(--pf-global--gutter);
     overflow: auto;
 }
 
 #main-storage > .pf-c-page__main {
-    padding-bottom: var(--pf-global--spacer--md);
+    padding: var(--pf-global--gutter);
```
KKoukiou, line 38: This should be fixed by adding the PageSection component around the grid. In pkg/storaged/overview.jsx This will also fix the padding to be the same like all other pages, now it's smaller.

KKoukiou: <details>
  <summary>Click to expand!</summary>
  
``` diff
$ git diff
diff --git pkg/storaged/overview.jsx pkg/storaged/overview.jsx
index 65a8e91a3..70ef05ccb 100644
--- pkg/storaged/overview.jsx
+++ pkg/storaged/overview.jsx
@@ -19,7 +19,7 @@
 
 import React from "react";
 
-import { Page, Grid, GridItem, Card, CardBody, Gallery } from "@patternfly/react-core";
+import { Page, PageSection, Grid, GridItem, Card, CardBody, Gallery } from "@patternfly/react-core";
 
 import { StoragePlots } from "./plot.jsx";
 
@@ -37,30 +37,32 @@ import { StorageLogsPanel } from "./logs-panel.jsx";
 export const Overview = ({ client, plot_state }) => {
     return (
         <Page id="main-storage">
-            <Grid hasGutter>
-                <GridItem md={8} lg={9}>
-                    <Gallery hasGutter>
-                        <Card>
-                            <CardBody>
-                                <StoragePlots plot_state={plot_state} />
-                            </CardBody>
-                        </Card>
-                        <FilesystemsPanel client={client} />
-                        <LockedCryptoPanel client={client} />
-                        <NFSPanel client={client} />
-                        <JobsPanel client={client} />
-                        <StorageLogsPanel />
-                    </Gallery>
-                </GridItem>
-                <GridItem md={4} lg={3} className="storage-sidebar">
-                    <Gallery hasGutter>
-                        <ThingsPanel client={client} />
-                        <DrivesPanel client={client} />
-                        <IscsiPanel client={client} />
-                        <OthersPanel client={client} />
-                    </Gallery>
-                </GridItem>
-            </Grid>
+            <PageSection>
+                <Grid hasGutter>
+                    <GridItem md={8} lg={9}>
+                        <Gallery hasGutter>
+                            <Card>
+                                <CardBody>
+                                    <StoragePlots plot_state={plot_state} />
+                                </CardBody>
+                            </Card>
+                            <FilesystemsPanel client={client} />
+                            <LockedCryptoPanel client={client} />
+                            <NFSPanel client={client} />
+                            <JobsPanel client={client} />
+                            <StorageLogsPanel />
+                        </Gallery>
+                    </GridItem>
+                    <GridItem md={4} lg={3} className="storage-sidebar">
+                        <Gallery hasGutter>
+                            <ThingsPanel client={client} />
+                            <DrivesPanel client={client} />
+                            <IscsiPanel client={client} />
+                            <OthersPanel client={client} />
+                        </Gallery>
+                    </GridItem>
+                </Grid>
+            </PageSection>
         </Page>
     );
 };
diff --git pkg/storaged/storage.scss pkg/storaged/storage.scss
index b80be8273..36225c06e 100644
--- pkg/storaged/storage.scss
+++ pkg/storaged/storage.scss
@@ -34,10 +34,6 @@
     overflow: auto;
 }
 
-#main-storage > .pf-c-page__main {
-    padding: var(--pf-global--gutter);
-}
-
 #storage .create-storage {
     margin-bottom: 10px;
 }
```
</details>

pkg/storaged/storage.scss
```diff
@@ -219,6 +219,10 @@ div.progress {
     min-width: 250px;
 }
```
KKoukiou, line 221: When we don't show the headers then the first item does not have a top border. I don't really consider it a bug. How do you think we should override the default behavior here?

pkg/storaged/side-panel.jsx
```diff
@@ -120,12 +120,12 @@ export class SidePanelRow extends React.Component {
...
                       onClick={this.props.go ? go : null}>
                 <Flex flexWrap={{ default: 'nowrap' }}>
-                    <FlexItem grow={{ default: 'grow' }} className="sidepanel-row-name">{this.props.name}</FlexItem>
+                    <FlexItem grow={{ default: 'grow' }} className="sidepanel-row-name pf-u-text-break-word">{this.props.name}</FlexItem>
                     <FlexItem>{decoration}</FlexItem>
                 </Flex>
                 <Flex className="sidepanel-row-info">
                     <FlexItem grow={{ default: 'grow' }} className="sidepanel-row-detail">{this.props.detail}</FlexItem>
-                    <FlexItem className="sidepanel-row-devname">{this.props.devname}</FlexItem>
+                    <FlexItem className="sidepanel-row-devname pf-u-text-break-word">{this.props.devname}</FlexItem>
```
KKoukiou, line 128: pf-u-text-break-word is in @patternfly/patternfly": "4.96.2", that we consume now in cockpit

KKoukiou (changes requested):
I did not check yet the last commit reg hover.

pkg/storaged/storage.scss
```diff
@@ -521,3 +521,8 @@ td.storage-action {
 .crypto-keyslots-list {
     border: none;
 }
+
+// Holdover until PF is upgraded and we get pf-u-align-self-center and drop this
```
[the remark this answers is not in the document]
garrett: We need the new PF. It's https://github.com/cockpit-project/cockpit/pull/15731

pkg/storaged/nfs-details.jsx
garrett: So we could actually upgrade to PF and that would take effect, then drop the custom CSS. Blocked on https://github.com/cockpit-project/cockpit/pull/15731

pkg/storaged/storage.scss
garrett: Ah cool.

pkg/storaged/side-panel.jsx
garrett: Nope, just like the other utility classes needed for this PR, it's not in Cockpit yet. Waiting on https://github.com/cockpit-project/cockpit/pull/15731.

garrett: Note: I tested these new utility classes and none worked, hence the class for the proper method and the duplication to fix it for now, until we actually can use them.

KKoukiou: This is not accurate. So, I git described the commit which introduced these classes and they are there since 4.96.2. So definitely not blocked. 

Note: When adding such comments in the code please always link the exact version that we are blocked on, rather than 'Our version of PF4 is out of date at the moment', otherwise these workaround would be in the code for eternity.

The reason you can't use them is that you probably did not import the files you need in order to consume them. Import @patternfly/patternfly/utilities/Text/text.scss for this case.

pkg/storaged/nfs-details.jsx
KKoukiou: Same the other comment. These are not blocked, they are merged in fact two years ago in PF upstream. \o/

You just need to import the stylesheet. src/patternfly/utilities/Flex/flex.scss

pkg/storaged/storage.scss
```diff
@@ -500,6 +500,34 @@ td.storage-action {
     font-size: var(--pf-global--FontSize--xs);
 }
 
+.sidepanel-row:not(.pf-c-empty-state):hover {
```
KKoukiou, line 503: This is basically making the content of the side table rows look like a link. Instead of making it look like one, can we just make it a link and avoid these overrides?

pkg/storaged/storage.scss
```diff
@@ -500,6 +500,34 @@ td.storage-action {
...
+    cursor: pointer;
+}
+
+/* Add hover styling to ct-tables and sidepanels */
+
+.sidepanel-row:not(.pf-c-empty-state):hover,
+.storage-mounts .ct-table tbody tr:hover {
+    box-shadow: var(--pf-global--BoxShadow--sm);
```
KKoukiou, line 511: Is this really needed? we have tables with rows that are links and redirect in other pages in may pages, like network, machines etc. None of them have box-shadow in rows.

pkg/storaged/storage.scss
```diff
@@ -500,6 +500,34 @@ td.storage-action {
...
+.storage-mounts .ct-table {
+    // Reset color style
+    --ct-color-list-hover-text: all;
+}
+
+.sidepanel-row:hover .sidepanel-row-name,
+#others .sidepanel-row:hover .sidepanel-row-devname,
+.storage-mounts tr:hover td[data-label=Name],
+.storage-mounts tr:hover td[data-label=Server] {
+    // ct color variable is for the ct-table-based widgets
```
KKoukiou, line 523: These can go away if you use `<Button variant="link" component="link"` for the rows content.

KKoukiou (changes requested):

pkg/storaged/storage.scss
garrett: All the table demos have headings; I guess non-heading isn't used as much?
https://www.patternfly.org/v4/components/table

Data list demos have lines at the top
https://www.patternfly.org/v4/components/data-list

I think it's a bug or an oversight.

We should probably fix it elsewhere though? Perhaps in PF4 overrides and file an issue upstream? WDYT?

KKoukiou: No idea if it's a bug. Sounds like an easy however.

pkg/storaged/storage.scss
garrett: I was copying PF4's table hover. But, yeah, I guess we don't do that anyway. I'll drop it.

pkg/storaged/storage.scss
garrett: Probably. I'd need help from someone to have this happen.

pkg/storaged/side-panel.jsx
garrett: It doesn't work.

So I added these at the top:

```jsx
import "@patternfly/patternfly/utilities/Flex/flex.scss";
import "@patternfly/patternfly/utilities/Text/text.scss";
```

And got this:

```
Error: no mixin named pf-utility-builder
        on line 31:10 of node_modules/@patternfly/patternfly/utilities/Text/text.scss
>> @include pf-utility-builder($pf-u-font-size-options, $pf-global--breakpoint-
   ---------^
Error: no mixin named pf-utility-builder
        on line 11:10 of node_modules/@patternfly/patternfly/utilities/Flex/flex.scss
>> @include pf-utility-builder($pf-u-flex-direction-options, $pf-global--breakp
   ---------^
```

So I added this above the other imports:

```jsx
import "@patternfly/patternfly/sass-utilities/mixins.scss";
```

And same errors.

What should we do here instead?

garrett: There aren't any other examples in Cockpit that I could look at, sadly.

KKoukiou: Can you import "../../node_modules/@patternfly/patternfly/utilities/Flex/flex.scss"; on the pkg/lib/storage.scss file? Just tried it and it works fine there.

pkg/storaged/storage.scss
KKoukiou: Something like https://github.com/cockpit-project/cockpit/commit/adbce4f0a4c017c64de674af349438a7f30eee6b

KKoukiou: And this one to use hrefs instread of onClick handlers on the links, https://github.com/KKoukiou/cockpit/commit/7fe59a41bbcbe86b3f57ebb92123574df486d883

pkg/storaged/side-panel.jsx
garrett: Oh, this was on the JSX file, as that's where it'll be used. I see; importing on SCSS makes sense, as the utilities are just CSS. :facepalm: Right, this makes sense. Thanks!

pkg/storaged/storage.scss
garrett: Oh, but we're not applying it to PF4 widgets here; this is all legacy code. But I think it's also a problem with PF4 too.

KKoukiou: We use actually directly PF4 flex.

KKoukiou (commented):
All the changes look good apart from last commit: storage: Make table hovering consistent 

This needs to be ported to proper link button, instead of trying to make it look like one :)

garrett:
I'm working on the last one now, porting things over. Of course, adjusting and shifting around the other things broke your proposed changes, so I'm adapting it.

I think I'm nearly done making it work and will push when ready.

garrett:
Oh, things are broken... :cry:

garrett:
I've dropped the hovering and your patches on top and moved those to https://github.com/cockpit-project/cockpit/pull/15914.

As you can see, it's quite broken. The whole card is a link for some reason, and it only affects the sidepanels, not the other tables, so everything's inconsistent.

I've left the commits separate, so we could debug it. But the rest of the changes should go in ASAP, hopefully for the release?

KKoukiou (approved):

KKoukiou:
@garrett Please don't merge till we trigger the tests here.

mvollmer (approved):

mvollmer:
I call this green. fedora-33/firefox is missing, but heh.

PR #15904, opened 2021-06-07
pkg/build: new script for standalone webpack

allisonkarlitskaya:
Split the necessary variables and rules for building dist/ out of the
primary Makefile.am and into pkg/build.  Include that back into the main
Makefile.am.

The only changes are to the invocation of webpack-make:

 - don't set the BUILDDIR.  Nothing uses that anymore.

 - don't call webpack-make via tools/missing.  This won't work without
   autogen, and it never made sense anyway, mostly leading to non-sense
   error messages.

 - stop using timeout.  We don't seem to get stuck builds all that often
   anymore, and besides, we have better ways of dealing with those these days.
   Meanwhile, using this tool interferes with signal delivery (which is why we
   often see node running long after terminating a build).

This file is a makefile, but has `#!/usr/bin/make -p` on top, and is
marked as executable, so it can be invoked as a script, like so:

```sh
pkg/build -j$(nproc)
```

from a clean checkout.  This allows building dist/ without first running
./autogen.sh.

Meanwhile, from the standpoint of a normal build, nothing has changed.

### DIFF
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -16,25 +16,19 @@ var ops = stdio.getopt({
     watch: { key: "w", args: 0, description: "Enable webpack watch mode" },
 });
 
-var srcdir = (process.env.SRCDIR || ".").replace(/\/$/, '');
-var makefile = ops.deps;
-var prefix = "packages";
-var npm = { "dependencies": { } };
+const makefile = ops.deps;
+const prefix = makefile.split("/").slice(-2, -1)[0];
+process.env["ONLYDIR"] = prefix + "/";
 
-if (makefile) {
-    prefix = makefile.split("/").slice(-2, -1)[0];
-    process.env["ONLYDIR"] = prefix + "/";
-    npm = JSON.parse(fs.readFileSync(path.join(srcdir, "package.json"), "utf8"));
-}
-
-var cwd = process.cwd();
-var config_path = path.resolve(cwd, ops.config);
-var config = require(config_path);
+const srcdir = (process.env.SRCDIR || ".").replace(/\/$/, '');
+const cwd = process.cwd();
+const config_path = path.resolve(cwd, ops.config);
+const config = require(config_path);
 
 // The latest input file time updated and used below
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
I have a feeling (completely unconfirmed) that the removal of `timeout` in this PR will fix the issue we see in the tests container where the workdir can't be removed because it's non-empty.

tools/webpack-make
```diff
@@ -27,7 +27,7 @@ var ops = stdio.getopt({
     watch: { key: "w", args: 0, description: "Enable webpack watch mode" },
 });
 
-var srcdir = process.env.SRCDIR || ".";
+var srcdir = process.env.SRCDIR || "./";
```
martinpitt, line 30: would be nice to make the subject more descriptive: s/minor bugfix/fix default srcdir separator/ or so?

martinpitt (dismissed):
Thanks! CI broke on ENOSPC yesterday, retrying tests.

tools/webpack-make
allisonkarlitskaya: The description of the problem in this commit message is actually incorrect.  I'll look into it again, but I'm also happy to just drop this commit.

allisonkarlitskaya: The problem is actually kinda funny: we iterate over input files and we check if filenames begin with the srcdir, and then remove the srcdir, plus one extra character (for the `/` that got inserted by path joining) from the start of the filename.

If srcdir is a bare `.` then this also matches `.babelrc` for example, and removes the `.` plus one extra character, landing us with `abelrc`.

I guess the real fix is actually to check if the string starts with `srcdir + '/'`.

allisonkarlitskaya: Okay.  Fixing this led me to fix a bunch of other things in `webpack-make`.  This PR is a big larger in scope now.

martinpitt (approved):
Thanks, looks great! Triggering a few representative tests.

PR #15906, opened 2021-06-08
test: Call self.machine.allowed_messages in check_journal_messages

mvollmer:
instead of having all the image specific quirks listed explicitly.


martinpitt (approved):
Nice! I landed the bots PR, and triggered the required fedora-34 status, so this ought to get green now.

martinpitt (approved):

PR #15908, opened 2021-06-08
test: Use correct selectors for listing toggles in check-selinux

mvollmer:
Looks like we always had a spiffy clean SELinux record.  But a new
rhel-9-0 image now needs this code to work again.


KKoukiou (approved):

martinpitt:
Half of the tests finished. To unblock stuff a little bit, I land this now, this is straightforward enough.

PR #15909, opened 2021-06-08
lib: remove listing-ct-toggle unused rules from listing.scss

KKoukiou:
Also drop jquery-patterns which use deprecated classes and of course
the library we 're so hard trying to get rid of.


martinpitt (approved):
Niiiice!

martinpitt:
Feel free to land this with just half of the tests; I just had to restart the unit tests, otherwise I'd have landed.

PR #15910, opened 2021-06-08
test: Only remove a loopback device file once its done

mvollmer:
Removing it early breaks newer LVM2 versions, see
https://bugzilla.redhat.com/show_bug.cgi?id=1969408

### DIFF
--- a/test/verify/storagelib.py
+++ b/test/verify/storagelib.py
@@ -66,12 +66,14 @@ class StorageHelpers:
 
         Return the device name.
         '''
+        # HACK: https://bugzilla.redhat.com/show_bug.cgi?id=1969408
+        # It would be nicer to remove $F immediately after the call to
+        # losetup, but that will break some versions of lvm2.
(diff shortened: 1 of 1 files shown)


mvollmer:
Needed for the rhel-9-0 image refresh.

martinpitt (dismissed):
Yeah, well, hmmkay. But thanks for tracking down the cause and bz!

mvollmer:
https://bugzilla.redhat.com/show_bug.cgi?id=1969408 has suggestions for workarounds, but I think it is best for us not to modify the lvm2 configuration. We actually want to find bugs like this.

But I'll mark this as a HACK so that we might remove it laterish.

martinpitt:
packit tests [failed](http://artifacts.dev.testing-farm.io/0b55015b-1ed8-482e-8d02-3647833115c6/) with

    ‘/root/ARTIFACTS/work-all8oa21B/plans/all/discover/default/tests/test/run-verify-host-user.sh’: Permission denied

This.. is new..

martinpitt (approved):

PR #15911, opened 2021-06-08
insights: Correctly align Insights status

marusak:
Before:
Screenshot from 2021-06-08 13-48-39currentinsights

After:
Screenshot from 2021-06-08 14-05-40aftyerinsights

### DIFF
--- a/pkg/systemd/overview-cards/insights.jsx
+++ b/pkg/systemd/overview-cards/insights.jsx
@@ -19,6 +19,7 @@
 
 import React from 'react';
 import { Button } from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 import cockpit from "cockpit";
 import * as service from "service.js";
@@ -182,7 +183,12 @@ export class InsightsStatus extends React.Component {
         return (
             <li className="system-health-insights">
                 {icon}
-                <Button variant="link" component='a' href={url}>{_("Insights: ")} {text}</Button>
+                <Button isInline variant="link" component='a' href={url}
+                        target="_blank" rel="noopener noreferrer"
+                        icon={<ExternalLinkAltIcon />}
+                        iconPosition="right">
+                    {_("Insights: ")} {text}
+                </Button>
             </li>);
     }
 }


KKoukiou (dismissed):

marusak:
Should this also have icon for external link? And also `target="_blank" rel="noopener noreferrer"` for security reasons?

KKoukiou:
> Should this also have icon for external link? And also `target="_blank" rel="noopener noreferrer"` for security reasons?

Both true. So, ExternalLinkSquareAltIcon. Feel free to repost.

marusak:
Screenshot from 2021-06-08 14-20-04redirection

marusak:
Isn't that icon too disruptive?

KKoukiou:
> Isn't that icon too disruptive?

it is actually :( how about you remove the description ou of the link. so '2 hits, including moderate' and then next two it add a link 'View insights', maybe with the external link icon.

marusak:
> it is actually :( how about you remove the description ou of the link. so '2 hits, including moderate' and then next two it add a link 'View insights', maybe with the external link icon.

That could work, but it would be too much for a single line... Maybe @garrett has a good idea about this?

garrett:
ExternalLinkSquareAltIcon can be used at the end of links too, BTW: https://www.patternfly.org/v4/components/button/design-guidelines/#link-button

If that alone doesn't help, we could do something like:

```
🗷 Insights: 2 hits, including moderate
   View details ⍈
```

Or perhaps:
```
🗷 Moderate hits; 2 total
   View Insights ⍈
```

We could even try to make that fit on one line?

> 🗷 Moderate hits; 2 total [View Insights ⍈](#)

And if there's just one, it could be like this?

> 🗷 Moderate hit [View Insights ⍈](#)

marusak:
Screenshot from 2021-06-09 09-42-03includingimportnat

Having it at the end seems fine? @KKoukiou @garrett WDYT?

KKoukiou:
> Screenshot from 2021-06-09 09-42-03includingimportnat
> 
> Having it at the end seems fine? @KKoukiou @garrett WDYT?

Looks fine to me. Make sure to use it like in the docs:
```
    <Button variant="link" icon={<ExternalLinkSquareAltIcon />} iconPosition="right">
      Link
    </Button>
```
In order to get proper spacing.

marusak:
> Looks fine to me. Make sure to use it like in the docs:

Thats how I used it here. I havn't yet pushed as I want to get ack first.

garrett (approved):
Works for me. Thanks!

KKoukiou (approved):

martinpitt:
packit failure is most likely because the proposed branch is 45 commits behind master, so let's ignore that here.

martinpitt:
Today's tasks container refresh (inevitable due to moving webhook) brought a new firefox which is now [failing two metrics tests](https://logs.cockpit-project.org/logs/pull-15911-20210609-134050-0295530f-fedora-33-firefox/log.html#271-2). We need to investigate them, but no need to block PRs.

PR #15913, opened 2021-06-08
storage: take into consideration x-systemd.automount when showing warning in fs tab

### LINKED ISSUE #15902: storage: Cockpit does not understand systemd automount
On my Pi 4 server, I'm running Cockpit and I get warnings that partitions won't be automatically mounted on next boot.

However, I'm using x-systemd.automount:

An excerpt from my `/etc/fstab`:
```
UUID=67fc30f3-5ec8-4aba-840d-5ceb1fd0f72d /var/srv ext4 nofail,noauto,x-systemd.automount,x-systemd.device-timeout=1ms,x-systemd.idle-timeout=5min 0 0
LABEL=Butter /var/mnt/raid btrfs nofail,noauto,rw,x-systemd.automount,x-systemd.idle-timeout=5min,relatime,compress-force=zstd:3,space_cache=v2,subvolid=5,subvol=/,x-systemd.device-timeout=1ms 0 0
```

The difference between x-systemd.automount and normal auto: 
- `auto` will automount and block the system's boot. This is especially bad for devices that can be physically disconnected.
- `x-systemd.automount` will boot as much as possible without blocking on the device, until it's needed.

And here's a warning in Cockpit:

Screenshot 2021-06-07 at 11-24-45 Storage - garrett Snow

There are two ways to support this:

1. Simply have Cockpit understand that this option is equivalent to `auto` and don't display the warning.
2. Have a UI to set the flag. (Full support)

I'm suggesting # 1 (simple support) in this issue.

KKoukiou:
When the FS is currently mounted and x-systemd.automount is used
together with noauto, do not show a warning about inconsistent mount
point.

The FS will be mounted, will just not block early boot.

Closes #15902

### DIFF
--- a/pkg/storaged/fsys-tab.jsx
+++ b/pkg/storaged/fsys-tab.jsx
@@ -118,6 +118,7 @@ export function check_mismounted_fsys(client, path, enter_warning) {
     const [, dir, opts] = get_fstab_config(block);
     const split_options = parse_options(opts);
     const opt_noauto = extract_option(split_options, "noauto");
+    const opt_systemd_automount = split_options.indexOf("x-systemd.automount") >= 0;
     const is_mounted = mounted_at.indexOf(dir) >= 0;
     const other_mounts = mounted_at.filter(m => m != dir);
     const crypto_backing_noauto = get_cryptobacking_noauto(client, block);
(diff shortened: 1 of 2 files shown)


KKoukiou:
test infra is down - sending in with no-test anyway - to get some code review and I will run the tests when I can. Tests pass locally.

pkg/storaged/fsys-tab.jsx
```diff
@@ -118,6 +118,7 @@ export function check_mismounted_fsys(client, path, enter_warning) {
     const [, dir, opts] = get_fstab_config(block);
     const split_options = parse_options(opts);
     const opt_noauto = extract_option(split_options, "noauto");
+    const opt_systemd_automount = extract_option(split_options, "x-systemd.automount");
```
mvollmer, line 121: This will omit "x-systemd.automount" also from the UI. (It will be removed from the `split_options` array.) We should just check whether it is in it, `split_options.indexOf("x-systemd.automount") != -1`

mvollmer (changes requested):

pkg/storaged/fsys-tab.jsx
KKoukiou: Oh good catch - did not realize it.

mvollmer (approved):

PR #15914, opened 2021-06-08
storage: Fix links and hovering

garrett:
Follow-up to https://github.com/cockpit-project/cockpit/pull/15903

### DIFF
--- a/pkg/storaged/drives-panel.jsx
+++ b/pkg/storaged/drives-panel.jsx
@@ -86,7 +86,7 @@ export class DrivesPanel extends React.Component {
                               name={name}
                               devname={block_name(block)}
                               detail={desc}
-                              go={() => cockpit.location.go([dev])}
+                              location={"#/" + dev}
                               job_path={path}
                               key={path} />
             );
--- a/pkg/storaged/mdraids-panel.jsx
+++ b/pkg/storaged/mdraids-panel.jsx
@@ -39,7 +39,7 @@ const MDRaidRow = ({ client, path }) => {
                       name={mdraid_name(mdraid)}
                       devname={block && block_name(block)}
                       detail={fmt_size(mdraid.Size) + " " + _("RAID device")}
-                      go={() => cockpit.location.go(["mdraid", mdraid.UUID])}
+                      location={"#/vg/mdraid/" + mdraid.UUID}
                       job_path={path} />
     );
 };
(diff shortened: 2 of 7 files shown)


garrett:
Everything but the last three commits are handled in the other PR. Once that lands and this is rebased, those should drop out. (However, the commits depend on the others, so they're here as well.)

garrett:
We chatted about this in IRC today. The main points:

- The goal of this PR was to make different imeplemtations of similar widgets act similarly. Right now, there are multiple versions of how this is shown in the Storage page.
- PatternFly's hover state is lacking. It doesn't look clickable. It's too subtle and in many cases (bad laptop screen, poor room lighting, multiple sight-related disabilities), the light grey might not even be visible.
- The list hover action in Cockpit was created to solve issues found in UX studies, to make them look more clickable. The hover with a link solved the issues when re-testing.
- Relying on color alone is problematic due to some forms of colorblindness.
- Making every item look like a link is overkill and becomes noisy.
- Perhaps using a PF4 list with links on the titles of each item (expanded to the cell to increase the clickable target) might be the best solution. It's either this or making the whole row clickable and show link style on hover, like in the journal and in this PR in commit 9e6b09f.

garrett:
Rebased.

pkg/storaged/drives-panel.jsx
```diff
@@ -86,7 +86,7 @@ export class DrivesPanel extends React.Component {
                               name={name}
                               devname={block_name(block)}
                               detail={desc}
-                              go={() => cockpit.location.go([dev])}
+                              location={"#/" + dev}
```
mvollmer, line 89: We should pass the original argument here: `location={[dev]}` and use `cockpit.location.go` in SidePanelRow below, to get the encoding right.  If we really need to use `href` of the Button, let's use `cockpit.location.encode` to encode the abstract location into a URL fragment.

mvollmer (changes requested):

pkg/storaged/drives-panel.jsx
KKoukiou: Thanks for taking a look. button links should use href for accessibility purposes  indeed - but here we haven't finalized yet if we should actually use links. But thanks for the pointer!

garrett:
> if we should actually use links

We probably should. And if they're block (which expand to the parent cell), they would be easier to click/tap than inline. (This is what we do on other pages, like Machines and Services top-level.

Other devices would need to be fixed to have a title, I guess. (Not sure what the title would be though.)

garrett:
Closed, in preference to #16033.

We'll eventually want to upgrade the lists on the page, but later.

PR #15916, opened 2021-06-09
test: More robust pixel-tests pushing

mvollmer:
Use --force to really blow away the index.  Git otherwise refuses to
remove things from it that are neither in the HEAD commit nor in the
working tree.

Fetch the "empty" branch before using it.

Always push to the remote, even when there are no changes.  The previous push might have failed and this is an easy way to try again.

### DIFF
--- a/test/common/pixel-tests
+++ b/test/common/pixel-tests
@@ -24,7 +24,7 @@ cmd_pull() {
 
 cmd_status() {
     ( cd "$SUBDIR"
-      git rm --cached --quiet '*.png'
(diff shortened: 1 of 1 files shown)


mvollmer:
> Always push to the remote, even when there are no changes. The previous push might have failed and this is an easy way to try again.

This needs to be implemented.

KKoukiou (approved):
These are the exact changes I did manually with @mvollmer guidance earlier today and worked. 
+1

I wonder if I like the pixel-tests wrapper or if we should all learn submodules, but not discussion for here

PR #15917, opened 2021-06-09
shell: Make keys scrollable instead of modal body

### LINKED ISSUE #15769: shell: dropdown cannot escape out of the SSH keys dialog
Cockpit version: 243
Page: Shell

Screen Shot 2021-04-31 at 14 12 50

garrett:
This allows the select to pop above the list and outside of the dialog.

Fixes #15769

### DIFF
--- a/pkg/shell/shell.scss
+++ b/pkg/shell/shell.scss
@@ -141,17 +141,25 @@ iframe.container-frame {
 }
 
 #credentials-dialog {
-    div.table-scrollable {
-        max-height: 500px;
-        overflow-y: auto;
-        margin-right: 7px;
-        padding-right: 8px;
-        margin-top: 15px;
-        padding-top: 0px;
+    .pf-c-modal-box__body {
+        // Turn off modal scrolling
+        overflow: visible;
+    }
+
+    .modal-body {
+        display: grid;
+        grid-template-rows: auto 1fr;
+        // Full screen minus the button, heading, and padding
+        max-height: calc(100vh - (var(--pf-global--spacer--md) * 11));
+        grid-gap: var(--pf-global--spacer--md);
+    }
+
+    .table-scrollable {
+        // Scroll the table content instead
+        overflow: auto;
     }
 
     table {
-        margin-top: 0px;
         width: 100%;
 
         thead tr.load-custom-key {


pkg/shell/shell.scss
```diff
@@ -141,17 +141,25 @@ iframe.container-frame {
...
+        // Turn off modal scrolling
+        overflow: visible;
+    }
+
+    .table-scrollable {
+        margin-top: var(--pf-global--spacer--md);
+        // Hacky approximation of content outside the list
+        // (we shouldn't have to do this if the layout would be flex or grid)
+        --content-approximation: 18rem;
+        --screen-minus-content: 100vh - var(--content-approximation);
```
garrett, line 154: I'm not a fan of using a magic number, but I couldn't get the crufty code to behave, so here's a hack that works in the extreme smallest size all the way through desktop. As mobile is completely hosed anyway, this unbreaks it to be a little less bad than it already is, in the worst case scenario.

garrett: I figured out how to make this less hacky. I'll change and force push.

pkg/shell/shell.scss
```diff
@@ -141,17 +141,25 @@ iframe.container-frame {
...
+    .pf-c-modal-box__body {
+        // Turn off modal scrolling
+        overflow: visible;
+    }
+
+    .modal-body {
+        display: grid;
+        grid-template-rows: auto 1fr;
+        // Full screen minus the button, heading, and padding
+        max-height: calc(100vh - (var(--pf-global--spacer--md) * 11));
```
garrett, line 153: This is still a "magic number" in a way, but it's based on the standard PatternFly unit and only applies to the heading, button height, and the padding. All of the content is handled inside of this specific grid and the sizing of the list wrapper (table-scrollable) is whatever is left over after the top of the dialog.

(Previous iteration used a magic number that also guesstimated the text at the top and sizing was done in the wrapper. This is much better. Both are better than what we have in Cockpit currently, though.)

KKoukiou (approved):

martinpitt:
@KKoukiou : No point about retrying firefox, it's broken since today's tasks container update (inevitable due to moving webhook to AWS, sorry). We need to fix for firefox 89 or roll back the container on the tasks runner.

PR #15921, opened 2021-06-10
cockpit-po-plugin: use gettext-parser

allisonkarlitskaya:
The first 5 commits are small commits that make obvious changes.  The following two are large, but are pure refactors that don't impact the output in any way.

Improvements:
 - port from po2json to direct use of `gettext-parser`, dropping the extra node module in the process
 - don't use `msggrep` anymore

Changes:
 - `cockpit.js` never looks at the first item in the translations array, so hardcode it null
 - drop the `x-generator` header
 - use a lambda instead of an anonymous function, and trim it down
 - indentation fixes

Random and unrelated:
 - stop outputting the timestamp in `Makefile.deps`, in the name of build determinism.  The webpack build is now fully deterministic as a result.


pkg/lib/cockpit-po-plugin.js
```diff
@@ -105,24 +78,36 @@ module.exports = class {
...
+                ' "": {\n',
+                `  "plural-forms": ${this.get_plural_expr(parsed.headers['plural-forms'])},\n`,
+                `  "language": "${parsed.headers.language}"\n`,
+                ' }'
+            ];
+            for (const [msgctxt, context] of Object.entries(parsed.translations)) {
+                const context_prefix = msgctxt ? msgctxt + '\u0004' : ''; /* for cockpit.ngettext */
+
+                for (const [msgid, translation] of Object.entries(context)) {
+                    if (translation.comments.flag === 'fuzzy')
```
martinpitt, line 110: Is `===` correct? there could be more flags, doesn't that need to be a substring match?

martinpitt (approved):
Nice, thanks! One question, but LGTM. You wanted to land yourself, so I'm not doing this.

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: It didn't seem to change anything for our current set of `.po` files.  I'll do a follow-up.

PR #15922, opened 2021-06-10
webpack: bring back sass-loader, using Dart Sass

allisonkarlitskaya:
This partially reverts commit 4063e3df28fb877adc3336c1b4526c02ffec3b84, dropping our custom loader to call the sassc commandline tool, and bringing back sass-loader, and then makes further changes.

We are now using Dart Sass (which we already had in our package.json for some reason).  Pin the version to strictly less than 1.33 to avoid an annoying warning about division (which we don't have a lot of control over, since most cases of its are coming from PatternFly, who are currently deciding what to do about this issue, themselves).

See https://github.com/patternfly/patternfly/issues/4096 for some more discussion.

 - [x] #15941


webpack.config.js
```diff
@@ -545,7 +547,15 @@ module.exports = {
...
                         },
                     },
-                    'sassc-loader',
+                    {
+                        loader: 'sass-loader',
+                        options: {
+                            sassOptions: {
+                                outputStyle: 'compressed',
+                            },
+                            sourceMap: true,
```
KKoukiou, line 556: sourceMap: !production

webpack.config.js
```diff
@@ -560,7 +570,15 @@ module.exports = {
                             url: false
                         }
                     },
-                    'sassc-loader',
+                    {
+                        loader: 'sass-loader',
+                        options: {
+                            sourceMap: true,
```
KKoukiou, line 576: sourceMap: !production

webpack.config.js
```diff
@@ -560,7 +570,15 @@ module.exports = {
...
+                            sassOptions: {
+                                outputStyle: 'compressed',
```
KKoukiou, line 578: compressed should be only in production mode I believe.

KKoukiou (changes requested):

allisonkarlitskaya:
There's a minor slowdown here on the order of a couple of seconds.  Apparently it's down to async call overhead and we could reverse that by using the `fibers` package to do some kind of weird magic, but that's due to stop working in Node version 16...

Let's look at that separately.

allisonkarlitskaya:
the problem is that the word `truetype` is appearing in our CSS files suddenly, but only in `NODE_ENV=development`.  There are a lot of cases of this:

```css
dist/storaged/storage.css:  src: url("patternfly-icons-fake-path/pficon.eot?#iefix") format("eot"), url("patternfly-icons-fake-path/pficon.woff2") format("woff2"), url("patternfly-icons-fake-path/pficon.woff") format("woff"), url("patternfly-icons-fake-path/pficon.ttf") format("truetype"), url("patternfly-icons-fake-path/pficon.svg#pficon") format("svg");
```

which is failing the check:

```sh
elif grep "\.\./fonts/OpenSans\|fonts/.*eot\|truetype" ${cssfiles} >&2; then
    echo "ERROR: invalid patternfly fonts paths found" >&2
```

martinpitt:
@allisonkarlitskaya These are supposed to be fixed up by the `string-replace-loader` bits in webpack.config.js. At first sight this is because the pattern matches on `src:url..`, but dart-sass compiles it as `src: url..` (with a space in between). Try adjusting the string-replace pattern to `src: *url` ?

allisonkarlitskaya:
Just merged the parallel commit for the po plugin to make sure this would be handled correctly here.

image

Cool!

martinpitt (approved):
This looks great to me, and we all seemed to agree on IRC that this is the right way to go. Thanks!

martinpitt:
@allisonkarlitskaya : starter-kit (and derived projects) also use this. Do you want to send corresponding PRs for these as well? I cana give you a hand with that, of course.

allisonkarlitskaya:
> @allisonkarlitskaya : starter-kit (and derived projects) also use this. Do you want to send corresponding PRs for these as well? I cana give you a hand with that, of course.

gern :)

garrett:
> Pin the version to strictly less than 1.33 to avoid an annoying warning about division (which we don't have a lot of control over, since most cases of its are coming from PatternFly, who are currently deciding what to do about this issue, themselves).

[PatternFly 4.115.0 fixes Dart Sass compatibility](https://github.com/patternfly/patternfly/pull/4164), so once we upgrade PF to that version (or higher), we can even adjust our version of Dart Sass accordingly (that is: unlock the version requirements).

This was a _lot_ faster and less painful than what I thought it would be. :partying_face:

allisonkarlitskaya:
> [PatternFly 4.115.0 fixes Dart Sass compatibility](https://github.com/patternfly/patternfly/pull/4164), so once we upgrade PF to that version (or higher), we can even adjust our version of Dart Sass accordingly (that is: unlock the version requirements).

@garrett And PF3?

KKoukiou:
> > [PatternFly 4.115.0 fixes Dart Sass compatibility](https://github.com/patternfly/patternfly/pull/4164), so once we upgrade PF to that version (or higher), we can even adjust our version of Dart Sass accordingly (that is: unlock the version requirements).
> 
> @garrett And PF3?

I don't think that PF4 is anyhow maintained or supported.

allisonkarlitskaya:
> > > [PatternFly 4.115.0 fixes Dart Sass compatibility](https://github.com/patternfly/patternfly/pull/4164), so once we upgrade PF to that version (or higher), we can even adjust our version of Dart Sass accordingly (that is: unlock the version requirements).
> > 
> > 
> > @garrett And PF3?
> 
> I don't think that PF3 is anyhow maintained or supported.

This is kinda my point: I think we're gonna be stuck on the old Dart Sass version until we can stop using PF3.

PR #15923, opened 2021-06-10
check-metrics: use raw-string to avoid invalid escape warnings

allisonkarlitskaya:


allisonkarlitskaya:
firefox tests say things like

`(firefox:71921): GLib-ERROR **: 17:32:49.717: creating thread 'gmain': Error creating thread: Resource temporarily unavailable`

Seems like our CI nightmare continues...

martinpitt:
@allisonkarlitskaya : This is similar to the "resource not available" failures on fork(). Sometimes the CentOS CI runners fail like that, when there's not enough available processes. I  thought containers would shield you from that kind of stuff. This never happens on e2e or aws. I just retried the test.

martinpitt:
Actually no, that gmain_thread just looks like a broken firefox. I temporarily reverted our tasks container on the bots. One fire at a time..

PR #15924, opened 2021-06-10
storage: Offer choice of LUKS version 1 or 2 when formatting

mvollmer:
This is a sneaky attempt to solve a UI issue in #13192: By using a dropdown for switching on Encryption, we have a obvious place for adding the "Re-use old format and keys" option.

### DIFF
--- a/pkg/storaged/format-dialog.jsx
+++ b/pkg/storaged/format-dialog.jsx
@@ -145,10 +145,6 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
         return vals.type != "empty" && vals.type != "dos-extended";
     }
 
-    function is_encrypted(vals) {
-        return vals.crypto.on;
-    }
-
     function add_fsys(storaged_name, entry) {
         if (storaged_name === true ||
             (client.fsys_info && client.fsys_info[storaged_name] && client.fsys_info[storaged_name].can_format)) {
@@ -157,13 +153,31 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
     }
 
     var filesystem_options = [];
-    add_fsys("xfs", { value: "xfs", title: "XFS - " + _("Recommended default") });
+    add_fsys("xfs", { value: "xfs", title: "XFS " + _("(recommended)") });
     add_fsys("ext4", { value: "ext4", title: "EXT4" });
     add_fsys("vfat", { value: "vfat", title: "VFAT" });
     add_fsys("ntfs", { value: "ntfs", title: "NTFS" });
     add_fsys(true, { value: "empty", title: _("No filesystem") });
     if (create_partition && enable_dos_extended)
         add_fsys(true, { value: "dos-extended", title: _("Extended partition") });
+
+    function is_encrypted(vals) {
+        return vals.crypto !== "none";
+    }
+
+    function add_crypto_type(value, title, recommended) {
+        if ((client.manager.SupportedEncryptionTypes && client.manager.SupportedEncryptionTypes.indexOf(value) != -1) ||
+            value == "luks1") {
+            crypto_types.push({
+                value: value,
+                title: title + (recommended ? " " + _("(recommended)") : "")
+            });
+        }
+    }
+
+    var crypto_types = [{ value: "none", title: _("No encryption") }];
+    add_crypto_type("luks1", "LUKS1", false);
+    add_crypto_type("luks2", "LUKS2", true);
 
     var usage = utils.get_active_usage(client, create_partition ? null : path);
 
@@ -211,12 +225,8 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
                           validate: (name, vals) => utils.validate_fsys_label(name, vals.type),
                           visible: is_filesystem
                       }),
-            CheckBoxes("crypto", "",
-                       {
-                           fields: [
-                               { tag: "on", title: _("Encrypt data") }
-                           ]
-                       }),
+            SelectOne("crypto", _("Encryption"),
+                      { choices: crypto_types }),
             [
                 PassInput("passphrase", _("Passphrase"),
                           {
@@ -284,6 +294,7 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
                 var config_items = [];
                 if (is_encrypted(vals)) {
                     options["encrypt.passphrase"] = { t: 's', v: vals.passphrase };
+                    options["encrypt.type"] = { t: 's', v: vals.crypto };
 
                     var item = {
                         options: { t: 'ay', v: utils.encode_filename(crypto_options_dialog_options(vals)) },
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -32,8 +32,6 @@ import sha1 from "js-sha1";
 import sha256 from "js-sha256";
 import stable_stringify from "json-stable-stringify-without-jsonify";
 
-import * as python from "python.js";
-
 import {
     dialog_open,
     SelectOneRadio, TextInput, PassInput, Skip
@@ -42,7 +40,6 @@ import { array_find, decode_filename, block_name } from "./utils.js";
 import { fmt_to_fragments } from "./utilsx.jsx";
 import { StorageButton } from "./storage-controls.jsx";
 
-import luksmeta_monitor_hack_py from "raw-loader!./luksmeta-monitor-hack.py";
 import clevis_luks_passphrase_sh from "raw-loader!./clevis-luks-passphrase.sh";
 
 const _ = cockpit.gettext;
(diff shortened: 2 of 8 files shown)


mvollmer:
Looks like this:

image

mvollmer:
At least debian-stable doesn't have support for LUKS v2, yet. Cockpit will just omit it from the dropdown there.

pkg/storaged/format-dialog.jsx
```diff
@@ -165,6 +161,26 @@ export function format_dialog(client, path, start, size, enable_dos_extended) {
...
+        return vals.crypto !== "none";
+    }
+
+    function add_crypto_type(value, title) {
+        if ((client.manager.SupportedEncryptionTypes && client.manager.SupportedEncryptionTypes.indexOf(value) != -1) ||
+            value == "luks1") {
+            crypto_types.push({
+                value: value,
+                title: title + (value == client.manager.DefaultEncryptionType
+                    ? " - " + _("Recommended default")
```
KKoukiou, line 174: I would add it as a helperText bellow the select, but this can gladly be a followup or not at all.

KKoukiou (approved):
Codewise is good - most of the code is just moved anyway. Exposing the encryption type is fine from my side as well.

garrett:
1. Can we communicate what the original format is when it is selected?

2. Are cypher types going to be selectable as well? 

   - I happened upon this page recently about reencyption of LUKS devices: https://asalor.blogspot.com/2012/08/re-encryption-of-luks-device-cryptsetup.html 
   - and also an LWN article about Adiantum from 2019 @ https://lwn.net/Articles/776721/ which also had an example in the comments such as ` cryptsetup luksFormat --type luks2 --sector-size 4096 -c xchacha12,aes-adiantum-plain64 -s 256 <device>` 0 
   - ...and this got me wondering about these options. The custom encryption options would (probably?) work as a passthrough to specify some of this stuff _if you know exactly what you're doing_, but it's a bit weird if it's common. (This is, _if_ any of this is common.)

3. What do the options in the dropdown look like?
   - Other issue had the following:
     ```
     Encryption:  None                | v |
                  -------------------------
                  No encryption
                  Re-use old keys and format 
                  LUKSv1
                  LUKSv2 (recommended)
                  Truecrypt
      ```
   - We should probably say v1 is legacy or not recommended... I'd suggest just leaving it out, unless v2 isn't possible? But that might be going too far.
   - Shouldn't TrueCrypt be replaced with VeraCrypt? TrueCrypt had issues and was discontinued 7 years ago. (Or is this TrueCrypt format, like TCW?) Perhaps we should really use select groups instead of dividers, and have supported and legacy groups, with TrueCrypt in legacy, along with LUKSv1?

This can all be addressed in other PRs though. I guess this PR really is to combine the encryption off/on toggle with the selection of existing types.

mvollmer:
>     2. Are cypher types going to be selectable as well?

No, this is all limited by the API in UDisks2.  UDisks2 only offers "luks1" and "luks2".  UDisks2 also has a default, which is determined by the OS provider.  Cockpit will show the default as "Recommended". I think Fedora actually still defaults to "luks1", but I might be wrong.

>     3. What do the options in the dropdown look like?

In this PR, we will only get these possibilities:

   No Encryption
   LUKS v1 - Recommened
   LUKS v2

or

   No Encryption
   LUKS v1
   LUKS v2 - Recommened

or

   No Encryption
   LUKS v1

>        * We should probably say v1 is legacy or not recommended... I'd suggest just leaving it out, unless v2 isn't possible? But that might be going too far.

I think LUKS v1 is still the default on many OSes. All our supported OSes except debian-stable support LUKS v2, but only RHEL and Centos actually default to it, I think.  But I have to check to be certain.

>        * Shouldn't TrueCrypt be replaced with VeraCrypt?

I guess.. this was the only other non-LUKS thing that I could think of.  UDisks2 doesn't seem to support it, nor VeraScrypt...

garrett:
> I think Fedora actually still defaults to "luks1", but I might be wrong.

Fedora switched to LUKS 2 by default as of Fedora 30, >2 years ago.
https://pagure.io/fedora-docs/release-notes/issue/282
https://fedoraproject.org/wiki/Changes/SwitchCryptsetupDefaultToLUKS2

So I'd suggest:

- No Encryption
- LUKS1
- LUKS2 (recommened)

And when LUKS2 isn't available, we wouldn't have a recommended option. (We're not going to recommend LUKS1.)

RHEL documentation uses "LUKS1" and "LUKS2" when talking about LUKS version numbers, and says "LUKS" when referring to the concept outside of version numbers:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/encrypting-block-devices-using-luks_security-hardening#luks-versions-in-rhel-8_encrypting-block-devices-using-luks:

> In RHEL 8, the default format for LUKS encryption is LUKS2. The legacy LUKS1 format remains fully supported and it is provided as a format compatible with earlier RHEL releases.

mvollmer:
> Fedora switched to LUKS 2 by default as of Fedora 30, >2 years ago.

Right, I thought that was the case, but our tests disagree. They only expect LUKS2 on rhel and centos...

The default format used by Cockpit is determined by /etc/udisks2/udisks2.conf. Running `cryptsetup luksFormat` has its own default. Anaconda might just use the `cryptsetup` default, or have its own.

On our fedora-34 image:

```
[udisks2]
# Comma separated list of modules to load.
# Use asterisk to load all the modules.
modules=*
# Valid options are 'ondemand' or 'onstartup'.
modules_load_preference=ondemand

[defaults]
# Valid options are 'luks1' or 'luks2'
encryption=luks1
```

On our rhel-8-5 image:

```
[udisks2]
# Comma separated list of modules to load.
# Use asterisk to load all the modules.
modules=*
# Valid options are 'ondemand' or 'onstartup'.
modules_load_preference=ondemand

[defaults]
# Valid options are 'luks1' or 'luks2'
encryption=luks2
```

I'll ask the UDisks people what's going on. It doesn't look like a simple oversight, since someone has changed the default on RHEL but not on Fedora, hmm...

mvollmer:
> So I'd suggest:
> 
>     * No Encryption
> 
>     * ————————
> 
>     * LUKS1
> 
>     * LUKS2 (recommened)
> 
> 
> And when LUKS2 isn't available, we wouldn't have a recommended option. (We're not going to recommend LUKS1.)

Roger, makes sense to me.

pkg/storaged/format-dialog.jsx
mvollmer: This follows the same style we have for the filesystem type.

mvollmer:
Looks like this:

image

Note that we say "XFS - Recommended default" for the filesystem type.  Maybe we should make that consistent. I vote for just "Recommended" in both cases.

KKoukiou:
@mvollmer I would also stick to only 'Recommended' and not use 'by default'. I would also however add the 'Recommended' string to the option description as seen here https://www.patternfly.org/v4/components/select/#single-with-description

mvollmer:
> However this the the Select, I guess you ;re using FormSelect which does not have 'description' option. So in that case inline is fine of course.

Correct. I wouldn't know when to use one and when the other, though.

garrett:
FWIW: Using Form Select is more robust. It will always work overflows and it will show the native selector on mobile devices. However, the popup is not able to be styled and it does not support additional sub-widgets, as it's treated like (or in some cases is) a native widget.

---
About the content of the dropdowns:

We should use a "(recommended)", not "- Recommended". 

The minus sign is not a divider, but often stands in for a hypen. The em dash (—) is used to separate extra information or mark a break in a sentence. The en dash (–) is used to mark ranges, and sometimes like an em dash but with spaces on the sides. See: https://www.scribbr.com/language-rules/dashes/ & https://www.patternfly.org/v4/ux-writing/punctuation/#em-dash-en-dash-hyphen

As "recommended" does not start a sentence, it should be lowercase. As it isn't an actual attribute of the item, I'm suggesting parenthesis instead of a form of dash.

Additionally, "Type" (which should probably be "Filesystem type" or just "Filesystem") uses "- Recommended default". All of the "default" strings in this same part of the UI should be consistent.

garrett:
Oh, additionally: if we do use the rich select that @KKoukiou suggests, we can have proper dividers. Normal HTML Form Selects do not support dividers. (You have to leave them out or do a hack with a bunch of dashes set to disabled, or something similar.)

And, if we used that, we wouldn't have a dash or parenthesis in the description. And the other items probably shouldn't get a description.

garrett:
However, using a dropdown for 3 items is a bit wacky. We could split it to 3 (inline) radio buttons. :shrug:

mvollmer:
> However, using a dropdown for 3 items is a bit wacky. We could split it to 3 (inline) radio buttons. shrug

There will be four with #13192.

garrett:
> There will be four with #13192.

OK, keep previous suggestions. :wink:

mvollmer:
Looks like this now with "(recommended)":

image

KKoukiou (approved):
Approving again the interdiff

garrett (approved):
LGTM :+1:

PR #15925, opened 2021-06-10
storage: Re-order format dialog fields

mvollmer:


mvollmer:
Looks like this now:

image

image

martinpitt (approved):
Looks fine to me, thanks! But also requesting review from @garrett , in case he has some  opinion there.

mvollmer:
> But also requesting review from @garrett , in case he has some opinion there.

Yes, and there are many more fixes coming, which we probably should all be in the same release, so that the Format dialog doesn't change a little bit in each of the next five releases.

mvollmer:
Part of #15937 now.

PR #15929, opened 2021-06-10
shel: Make our SCSS Dart Sass friendly

garrett:
If we divide by multiplying fractions, it works without any odd Sass functions and imports.


allisonkarlitskaya (approved):
I mean, that seems OK to me :)

It's not strictly necessary at this point since we will be porting to an older sass version, but something like this will probably need to happen eventually (unless they decide to change course).

PR #15930, opened 2021-06-10
metrics: Change "Disks" history label to "Disk I/O"

martinpitt:
During user testing it was reported several times that the "Disks"
history column is ambiguous wrt. whether it shows capacity usage or I/O.
It's the latter, so update the label. Keep it in the current metrics
though, as that shows both kinds of metrics (and further clarifies them
with labels).

Also keep "Network" instead of "Network I/O", as there is no ambiguity
there, and the label would get too wide.

https://bugzilla.redhat.com/show_bug.cgi?id=1969904


KKoukiou (approved):

garrett (approved):
Thanks! Makes sense, looks good to me. :+1: 

(I just happened to be browsing GitHub on my day off and knew exactly what this change would be. I meant to send up a PR yesterday; it would've been exactly the same.)

PR #15931, opened 2021-06-10
More css cleanups + HWinfo to PF4 description list

KKoukiou:


pkg/base1/cockpit.js
```diff
@@ -24,6 +24,7 @@
 import "../lib/patternfly/patternfly-cockpit.scss";
 import "../lib/page.scss";
 import "@patternfly/patternfly/components/Button/button.css";
+import "../lib/table.css";
```
martinpitt, line 27: Now I see what triggered your question from this morning. As I said, I'd be happy to just leave the examples "ugly" (they already are, anyway), and not support cockpit.css any further. But fine for now.

KKoukiou:
Horizontal lists spacing between rows was previously overwritetten by https://github.com/cockpit-project/cockpit/blob/master/pkg/lib/patternfly/patternfly-4-overrides.scss#L45 and now this does not apply any more. 
I am happy to adjust it to apply to all description lists again.

Can do it as part of https://github.com/cockpit-project/cockpit/pull/15935

KKoukiou:
> Code changes LGTM, but see above layout thingy. If that's how PF4 wants lists to look like, then no veto from me. Danke!

They indeed want lists to work like this: https://www.patternfly.org/v4/components/description-list/#horizontal-2-col

But we have agreed to remove spacing between rows in horizontal lists. I will adjust PF4 overrides in another PR.

PR #15933, opened 2021-06-11
cockpit-po-plugin: improve fuzzy flag matching

allisonkarlitskaya:
We might have more than one flag, so look for the word "fuzzy" in the
entire flags string.

We use \b because JS RE doesn't seem to support < and > or \< and \>.

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -107,7 +107,7 @@ module.exports = class {
                     if (!this.check_reference_patterns(patterns, references))
                         continue;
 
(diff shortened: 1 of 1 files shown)


martinpitt (dismissed):
Thanks! (I would have been happy with indexOf(), there's not that many flags -- but yay precision!)

allisonkarlitskaya:
`''.matches()` vs `"".match()` oops.

martinpitt (approved):

allisonkarlitskaya:
Did a pair of local builds, and this one also preserves the "doesn't change the output" property that I made note of in the other commits.  I'd like to add that note here as well, when merging, via squash.

PR #15935, opened 2021-06-11
Make description lists collapse to vertical mode when on xs screen sizes

KKoukiou:
We used to have an local override for this in
patternfly-4-overrides.scss which however stopped working at some point.

PF4 exposed a class to do that now, so use that class for this purpose.

See https://github.com/patternfly/patternfly/issues/3630

Note: There are some cases were we might still want horizontal lists
even on xs screen sizes, and that when both label and content take very
small space. See metrics -> disk load.

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -35,14 +35,8 @@ svg {
     vertical-align: -0.125em;
 }
 
-// Patternfly horizontal lists should convert to vertical in small screens https://github.com/patternfly/patternfly/issues/3630
-@media (min-width: 640px) {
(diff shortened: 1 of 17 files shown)


KKoukiou:
Example

After:

Screen Shot 2021-06-12 at 11 29 41

Before:
Screen Shot 2021-06-12 at 11 29 20

garrett:
First looks good.

Why is the second description list (logical volumes) not collapsing?

KKoukiou:
> First looks good.
> 
> Why is the second description list (logical volumes) not collapsing?

We talked over it on IRC, it's because it's different widget - namely Table.

garrett (approved):
Nice to see support for this upstream! LGTM :+1:

Issue #1733, opened 2015-01-27
Trying to compile Cockpit under Ubuntu


FireDrunk:
As talked about during the meeting, i started compiling Cockpit on an Ubuntu 14.04.1 Virtual Machine.

After some initial dependency installations i got stock on this error:

(started with ./autogen.sh --prefix=/usr --enable-maintenance-mode --enable-debug)
```
configure: error: Package requirements (gudev-1.0 >= 165
                                   udisks2 >= 2.0.91
                                   accountsservice >= 0.6.35) were not met:

No package 'accountsservice' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables COCKPIT_DAEMON_CFLAGS
and COCKPIT_DAEMON_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```
The following dependancies are already installed:

```
root@cockpit-dev:/data/cockpit/tools# dpkg -l | grep udev
ii  gir1.2-gudev-1.0                    1:204-5ubuntu20.9                     amd64        libgudev-1.0 introspection data
ii  libgudev-1.0-0:amd64                1:204-5ubuntu20.9                     amd64        GObject-based wrapper library for libudev
ii  libgudev-1.0-dev                    1:204-5ubuntu20.9                     amd64        libgudev-1.0 development files
ii  libudev1:amd64                      204-5ubuntu20.9                       amd64        libudev shared library
ii  udev                                204-5ubuntu20.9                       amd64        /dev/ and hotplug management daemon
root@cockpit-dev:/data/cockpit/tools# dpkg -l | grep udisks
ii  gir1.2-udisks-2.0                   2.1.3-1                               amd64        GObject based library to access udisks2 - introspection data
ii  libudisks2-0:amd64                  2.1.3-1                               amd64        GObject based library to access udisks2
ii  libudisks2-dev                      2.1.3-1                               amd64        GObject based library to access udisks2 - development files
ii  udisks2                             2.1.3-1                               amd64        D-BUS service to access and manipulate storage devices
root@cockpit-dev:/data/cockpit/tools# dpkg -l | grep accountsservice
ii  accountsservice                     0.6.35-0ubuntu7.1                     amd64        query and manipulate user account information
ii  libaccountsservice0:amd64           0.6.35-0ubuntu7.1                     amd64        query and manipulate user account information - shared libraries
```
[...]

PR #15937, opened 2021-06-11
storage: Polish for the Format dialog

mvollmer:
Release note:

### Storage: the Format dialog has been polished

Fields have been re-ordered and warnings and errors are now consistent with Cockpit's common styling.

image

### DIFF
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -230,7 +230,7 @@ import {
     TextInput as TextInputPF4,
     Tooltip, TooltipPosition,
 } from "@patternfly/react-core";
-import { InfoCircleIcon } from "@patternfly/react-icons";
+import { InfoCircleIcon, ExclamationTriangleIcon } from "@patternfly/react-icons";
 
 import { show_modal_dialog, apply_modal_dialog } from "cockpit-components-dialog.jsx";
 
@@ -238,6 +238,7 @@ import { fmt_size, block_name, format_size_and_text } from "./utils.js";
 import client from "./client.js";
 
 import "form-layout.scss";
+import "@patternfly/patternfly/components/HelperText/helper-text.css";
 
 const _ = cockpit.gettext;
 
@@ -318,6 +319,16 @@ function flatten(arr1) {
     return arr1.reduce((acc, val) => Array.isArray(val) ? acc.concat(flatten(val)) : acc.concat(val), []);
 }
 
+const HelperTextWarning = ({ text }) =>
+    <div className="pf-c-helper-text">
+        <div className="pf-c-helper-text__item pf-m-warning">
+            <span className="pf-c-helper-text__item-icon">
+                <ExclamationTriangleIcon className="ct-icon-exclamation-triangle" />
+            </span>
+            <span className="pf-c-helper-text__item-text">{text}</span>
+        </div>
+    </div>;
+
 export const dialog_open = (def) => {
     const nested_fields = def.Fields || [];
     const fields = flatten(nested_fields);
@@ -393,7 +404,7 @@ export const dialog_open = (def) => {
 
         const extra = <div>
             { def.Footer }
-            { def.Action && def.Action.Danger ? <Alert isInline variant='danger' title={def.Action.Danger} /> : null }
+            { def.Action && def.Action.Danger ? <HelperTextWarning text={def.Action.Danger} /> : null }
         </div>;
 
         return {
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -288,6 +288,14 @@ function is_visible(field, values) {
 }
 
 const Body = ({ body, fields, values, errors, isFormHorizontal, onChange }) => {
+    let error_alert = null;
+
+    if (errors && errors.toString() != "[object Object]") {
+        // This is a global error from a failed action
+        error_alert = <Alert variant='danger' isInline title={errors.toString()} />;
+        errors = null;
+    }
+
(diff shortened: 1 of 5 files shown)


mvollmer:
@garrett, I'll work on the storage polish here. Let's try to get it all done for a single release.

garrett:
Our longstanding PatternFly update PR merged, so please also migrate the "alert" warning at the bottom of the dialog to use [helper text](https://www.patternfly.org/v4/components/helper-text) instead.

mvollmer:
About the kebab menu:

image

This looks messy to me, and would be cleaned up if everything would be behind the Kebab menu, including "Create partition".

The reason we have some actions in their own button is to make them more prominent as the "next obvious action".  If "Create partition" would be in a menu, people might struggle to figure out
how to proceed after creating a partition table, for example.

Another example: we make it easy to mount a filesystem by having a prominent "Mount" button, but we make it slightly harder to unmount it by only having "Unmount" in the menu.

garrett:
Good news! There are PatternFly components for these things now.

Quick action buttons should still have a space to the side even if there are no actions.
- composable actions (table variant): https://www.patternfly.org/v4/components/table#composable-actions

Also, if we want to do the indentation, there's a widget for that: 
- tree table (a table variant): https://www.patternfly.org/v4/components/table#tree-table
- composable tree table (another table variant): https://www.patternfly.org/v4/components/table#composable-tree-table
- https://www.patternfly.org/v4/components/tree-view/

In this case, the tree table with composable actions or a composable tree table with actions is what we want.

mvollmer:
Split the content stuff out into #15968

mvollmer:
With HelperText, Alert at top, and the hint that a unmount will happen first:

image

garrett:
> With HelperText, Alert at top, and the hint that a unmount will happen first:

Cool, it's already a lot better!

We should style up the currently in use thing... somehow. (I'm not sure exactly how, but it needs something.) But the rest is already tons better! :+1:

mvollmer:
> We should style up the currently in use thing... somehow.

Yes.  Here is a more complex example:

image

This is for erasing a disk that has multiple partitions, two of which have mounted filesystems, and one is part of a volume group.

garrett:
I wonder if we could use an info helper text for the currently used messages, and ensure there's a bit more spacing between each item (if PF won't add it by default). If using Helper text doesn't add the spacing itself, [additional spacing can be added with PF helper classes](https://www.patternfly.org/v4/utilities/spacing) once they're imported in the SCSS file, such as `pf-u-mt-md` (meaning: PatternFly Utility Margin Top Medium).

mvollmer:
The Format dialog with alignment lines for encryption options.

image

And for the Mount options:

image

garrett:
@KKoukiou pointed out it's the issue I reported @ https://github.com/patternfly/patternfly/issues/3868

mvollmer:
> @KKoukiou pointed out it's the issue I reported @ [patternfly/patternfly#3868](https://github.com/patternfly/patternfly/issues/3868)

Ok, then we don't need to change anything about checkbox alignments here.  Ready for review!

mvollmer:
Oh, wait, the teardown messages.

garrett:
I _think_ I'm on the branch and this is what I'm seeing locally.

Screenshot 2021-06-21 at 14-24-31 Storage - garrett Slate

Screenshot 2021-06-21 at 14-25-38 Storage - garrett Slate

Why is the slider the first thing? Shouldn't it be the name? And then the size should go between the type and erase option?

mvollmer:
> Why is the slider the first thing? Shouldn't it be the name? And then the size should go between the type and erase option?

I'll do that.

mvollmer:
TestStorageFormat.testFormatTypes fails on debina-testing reliably.

mvollmer:
> TestStorageFormat.testFormatTypes fails on debina-testing reliably.

This also fails on master, hmm.

mvollmer:
> > TestStorageFormat.testFormatTypes fails on debina-testing reliably.
> 
> This also fails on master, hmm.

Ahh, this is known issue #1733.

garrett (approved):
It's definitely getting better:

Screenshot 2021-07-01 at 10-58-37 Storage - garrett Thunder

I'm not a fan of how the "currently in use" block looks. But I guess that would require some design work. Unless you have an idea and want to follow-up with that part of the dialog on this PR, I'll approve for the rest. We can talk about and optimize that later (and should probably open up an issue for that specifically).

pkg/storaged/dialog.jsx
```diff
@@ -318,6 +319,16 @@ function flatten(arr1) {
     return arr1.reduce((acc, val) => Array.isArray(val) ? acc.concat(flatten(val)) : acc.concat(val), []);
 }
 
+const HelperTextWarning = ({ text }) =>
```
martinpitt, line 331: Any reason why this does not use the [HelperText](https://www.patternfly.org/v4/components/helper-text/) React component? That would also avoid the manual CSS import.

martinpitt (approved):
Danke! One question, but good enough as-is. Can you please write a little release note and add a current screenshot?

KKoukiou (approved):
Two comments:
* As far as I know we show dialog errors in the bottom of the dialogs in all other pages
* HelperText component should be used as @martinpitt pointed out

But it's really fine as is, we can post a followup if we agree on the two points above.

pkg/storaged/dialog.jsx
mvollmer: Oh, I somehow thought it doesn't exist...

PR #15938, opened 2021-06-11
streamline dist workflow

allisonkarlitskaya:
In summary:
 - use pkg/build instead of automake
 - run directly on the github actions runner

 - [x] #15922

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -14,10 +14,10 @@ jobs:
         # 2.14.0; if you update this, audit the diff and ensure that it does not leak/abuse secrets
         uses: dawidd6/action-download-artifact@b9571484721e8187f1fd08147b497129f8972c74
         with:
-          name: dist
+          name: webpack-jumpstart
           workflow: build-dist
           run_id: ${{ github.event.workflow_run.id }}
-          path: dist-unpack
+          path: artifact
 
       - name: Set up configuration and secrets
         run: |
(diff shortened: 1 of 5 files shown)


allisonkarlitskaya:
https://github.com/allisonkarlitskaya/cockpit/pull/10/checks?check_run_id=2802265419

allisonkarlitskaya:
packit fails are caused by the `workflow_run` workflow on the master branch not matching the `pull_request` workflow on this one.

Makefile.am
```diff
@@ -60,10 +60,6 @@ V_COPY = $(V_COPY_$(V))
 V_COPY_ = $(V_COPY_$(AM_DEFAULT_VERBOSITY))
 V_COPY_0 = @echo "  COPY    " $@;
 
-V_TAR = $(V_TAR_$(V))
```
martinpitt, line 63: Dropping this here looks wrong -- it's still being used by `install-data-local` and `dist-hook`.

.github/workflows/publish-dist.yml
```diff
@@ -14,10 +14,10 @@ jobs:
         # 2.14.0; if you update this, audit the diff and ensure that it does not leak/abuse secrets
         uses: dawidd6/action-download-artifact@b9571484721e8187f1fd08147b497129f8972c74
         with:
-          name: dist
+          name: webpack-jumpstart
           workflow: build-dist
           run_id: ${{ github.event.workflow_run.id }}
-          path: dist-unpack
+          path: artifact
```
martinpitt, line 20: naming nitpick: please set the path to webpack-jumpstart to reflect the tar and artifact name.

martinpitt (changes requested):
Thanks! Needs rebasing now, and two minor things, but this looks nice!

Makefile.am
allisonkarlitskaya: `Makefile.am` includes `pkg/build` and uses many things that are defined there, including the `V_COPY`, just above as well..

.github/workflows/publish-dist.yml
allisonkarlitskaya: I prefer the name artifact, in fact.  I find it to be absolutely unambiguous as a directory name in a job that downloads an artifact.  Too many things have the same name, otherwise, and it's harder to read and understand what's what.

I used the same thing in the `/npm install` workflow.

martinpitt: I see -- this is a directory *containing* webpack-jumpstart.tar.

martinpitt (approved):
LGTM, thank you!

allisonkarlitskaya:
Landing now because LGTM is going to fail anyway (see above).  Will push a test PR to make sure it's working properly, after that.

PR #15939, opened 2021-06-11
[r84] test: Call self.machine.allowed_messages in check_journal_messages

martinpitt:
Instead of having all the image specific quirks listed explicitly.

Cherry-picked from master commit f754e5f72dc.

 - [x] Fix unit-tests workflow target branch: PR #15940

PR #15940, opened 2021-06-11
workflows: Fix unit-tests' "changed" detection to compare against correct target branch

martinpitt:
Hardcoding "master" does not work for PRs which target e.g. rhel-8.4.

---

As noticed in #15939

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Build unit test container if it changed
         run: |
-          changes=$(git diff --name-only origin/master..HEAD -- containers/unit-tests/)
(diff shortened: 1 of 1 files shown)


martinpitt:
For master, the [workflow output](https://github.com/cockpit-project/cockpit/pull/15940/checks?check_run_id=2803677806#step:3:1) proves that it's a no-change. It still expands to "origin/master".

martinpitt:
... and in #15939 the unit-tests runs show that it works as expected there as well.

allisonkarlitskaya (approved):

allisonkarlitskaya (approved):
Looks fine, and I also confirmed that the two scenarios you cited are working as expected.

I've moved away from using `git diff` in workflows in this way in favour of `git log HEAD --not BASE` to look for matching commits.  That's a matter of taste, of course.

PR #15942, opened 2021-06-12
Weird whitespace

smwltr:
I think that file got included by accident.


martinpitt:
Thanks for spotting that, indeed that was a bad commit. However, that whole file is unused. It should be removed, together with doc/guide/feature-virtualmachines.xml and its mentioning in tools/test-static-code. Do you want to do this, or want me to send a PR?

smwltr:
I removed the file. It looks like 
`doc/guide/feature-virtualmachines.xml` is already removed.

martinpitt (approved):
Danke!

PR #15943, opened 2021-06-12
image-prepare: symlink only when image complete

### LINKED ISSUE #12781: `image-prepare` should use a tmpfile and rename it when done
Right now, `image-prepare fedora-30` will create `test/images/fedora-30.qcow2` straight away and link it into place.  That means that if the process is interrupted, you end up with a `fedora-30` link in the `test/images` directory, and the tests will try to run against that.

We should use a tempfile instead and rename it into place only when everything has run to the end successfully.

Alternatively: we could put `fedora-30.qcow2` in place, but not symlink it until the end.

smwltr:
Fixes #12781

I am not 100% sure this is the best improvement. I felt like rewriting much of it, but as I don't yet know all the ways in which the tests images can be used, this is probably the least intrusive and does what #12781 requests: creates the symlink to the image file only after building it has been completed.

### DIFF
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -121,7 +121,6 @@ def prepare_install_image(base_image, overlay=False):
                                qcow2_image])
         if os.path.lexists(install_image):
             os.unlink(install_image)
-        # os.symlink(os.path.basename(qcow2_image), install_image)
     return install_image, qcow2_image
 
 
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -121,8 +121,8 @@ def prepare_install_image(base_image, overlay=False):
                                qcow2_image])
         if os.path.lexists(install_image):
             os.unlink(install_image)
-        os.symlink(os.path.basename(qcow2_image), install_image)
-    return install_image
+        # os.symlink(os.path.basename(qcow2_image), install_image)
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! I had probably moved it to the end with `if completed:`, instead of calling it twice. But that's total bikeshedding, and we can do that if we touch the code again. Thank you!

martinpitt:
Whoops, this caused a regression, see https://github.com/cockpit-project/bots/pull/2114 . I'll send a PR.

smwltr:
On 6/15/21 12:59 PM, Martin Pitt wrote:
> Whoops, this caused a regression, see cockpit-project/bots#2114 
> <https://github.com/cockpit-project/bots/pull/2114> . I'll send a PR.
> 

Ah, ya. I see the problem. I'll have another go at it.

smwltr:
On 6/15/21 12:59 PM, Martin Pitt wrote:
> Whoops, this caused a regression, see cockpit-project/bots#2114 
> <https://github.com/cockpit-project/bots/pull/2114> . I'll send a PR.

I see your change with the new finalize_image function. So please 
disregard my last comment. Thanks!

PR #15948, opened 2021-06-14
packit.yaml: workaround to fix docs build

allisonkarlitskaya:
We don't get docs built by the build workflow anymore, so try building
them in the sandcastle.

update: That didn't work.  Let's add a nasty workaround, instead.

### DIFF
--- a/test/make_dist.py
+++ b/test/make_dist.py
@@ -20,7 +20,6 @@ import multiprocessing
 import os
 import subprocess
 import sys
-import time
 import argparse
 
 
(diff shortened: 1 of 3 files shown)


test/make_dist.py
```diff
@@ -51,19 +50,16 @@ def build_dist():
...
-            return True
-        except subprocess.CalledProcessError as e:
-            if e.returncode != 2 or not wait:
-                break
-            message("make_dist: pre-built dist not yet available, waiting...")
-            time.sleep(30)
-
-    message("make_dist: Downloading pre-built dist failed")
-    return False
+    cmd = 'tools/webpack-jumpstart'
```
martinpitt, line 53: This needs to be an array, otherwise the `.append()` below will crash. check_call() may get along with a single string (maybe implying shell, but hopefully assuming just a single argument-less exec name).

martinpitt (changes requested):
Yay green packit! Sorry for not spotting the bug below in the first round.

test/make_dist.py
allisonkarlitskaya: thanks!

i wrote this section twice, and as usual, had it right in the first time, and messed it up in the rewrite :)

allisonkarlitskaya:
I actually tested it this time and it works now.

```
⬢[lis@f34 packit-build-docs]$ rm -rf dist
⬢[lis@f34 packit-build-docs]$ test/make_dist.py --wait
  FETCH    dist  [ref: tag sha-9f88a677819a9624c65fa3ec32f4159f7cc238c0]
fatal: couldn't find remote ref refs/tags/sha-9f88a677819a9624c65fa3ec32f4159f7cc238c0
  WAIT     30s
```

martinpitt (approved):
Cheers! Tests are still running, but assuming green this looks good.

PR #15949, opened 2021-06-14
package.json: add fibers package

allisonkarlitskaya:
Add the "fibers" package to our node_modules/

sass-loader will detect if this package is present and inject it into
Dart Sass, enabling faster builds by avoiding async call overhead.


allisonkarlitskaya:
I'm closing this because, although this was a modest performance improvement:
 - it's 3M of extra `node_modules/`
 - ...including arch-specific binaries
 - this is going to stop working with Node 16, anyway.

PR #15951, opened 2021-06-14
test: check out the bots on an as-needed basis

allisonkarlitskaya:
Most python things that need to use the bots first import parent.py to
set up the path to enable importing things from the bots directory.

Use this as a nice place for a hook to check out the bots, if they're
not already there.

With this in place, we no longer need the blanket call in autogen.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -57,7 +57,6 @@ if test -z "${NOCONFIGURE:-}"; then
     fi
 fi
 
-tools/make-bots
 test/common/pixel-tests pull
 
 if test -n "${NOCONFIGURE:-}"; then
--- a/test/common/parent.py
+++ b/test/common/parent.py
@@ -1,4 +1,6 @@
+import importlib
 import os
+import subprocess
 import sys
 
 BASE_DIR = os.path.realpath(f'{__file__}/../../..')
@@ -8,3 +10,9 @@ BOTS_DIR = f'{BASE_DIR}/bots'
 sys.path.append(BOTS_DIR)
 sys.path.append(f'{TEST_DIR}/common')
 sys.path.append(f'{BOTS_DIR}/machine')
+
+
+def ensure_bots():
+    if not os.path.isdir(BOTS_DIR):
+        subprocess.check_call([f'{BASE_DIR}/tools/make-bots'])
+        importlib.invalidate_caches()
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -16,6 +16,7 @@ import importlib.machinery
 import importlib.util
 
 import parent
+parent.ensure_bots()  # NOQA: testvm lives in bots/
 import testlib
 import testvm
 
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -24,14 +24,11 @@ import subprocess
 
 import make_dist
 
-from common.parent import ensure_bots
+from common.parent import BASE_DIR, TEST_DIR, BOTS_DIR, ensure_bots
 ensure_bots()  # NOQA: testvm lives in bots/
 import testvm
 
-TEST = os.path.abspath(os.path.dirname(__file__))
-BASE = os.path.normpath(os.path.join(TEST, ".."))
-BOTS = os.path.join(BASE, "bots")
-os.environ["PATH"] = "{0}:{1}".format(os.environ.get("PATH"), BOTS)
+os.environ["PATH"] = "{0}:{1}".format(os.environ.get("PATH"), BOTS_DIR)
 
 networking = None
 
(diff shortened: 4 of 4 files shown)


allisonkarlitskaya:
...the other obvious place is `test/run` and it contains an explicit `tools/make-bots` call.

allisonkarlitskaya:
One possible draw-back: our own muscle memory of thinking that we can do things like `bots/tests-trigger` after autogen is done.

martinpitt:
I can't say that I like this. Automagically cloning stuff from the interwebs by importing a Python module feels like way too much magic to me. What scenario is this fixing? Perhaps there is a more specific place where this could be hooked into, like test/image-prepare?

allisonkarlitskaya:
```
⬢[lis@f34 main]$ git grep 'import parent' | wc -l
69
```

ie: every single `test/verify/check-*` script, plus:
 - `run-tests`
 - `image-prepare`

martinpitt:
Neither test/verify/check-* nor run-test do anything useful without a prepared image, so it's really just that one place.

allisonkarlitskaya:
I agree that side-effecting on import is to be avoided, and that this is the mother of all side-effects (and non-deterministic to boot, since we don't pin bots revision with a gitlink).  Normally this sort of thing would make my skin crawl, but for some reason in this case I feel pretty OK about it.

The bigger drawback in my opinion is that `bots/` isn't there, and `tools/make-bots` is difficult to type.  However: in (practically) any case where you already did an autogen, `make bots` will also be available, and that's actually really easy to type.

I'm good with treating this as an unnecessary optimisation and declining to merge this, but in terms of implementation, I think we won't do better, unfortunately...

allisonkarlitskaya:
> Neither test/verify/check-* nor run-test do anything useful without a prepared image, so it's really just that one place.

To be honest, I consider this a bug that we should fix.  Those either need to error out, or to build the image for themselves first.

`vm-run` against a non-prepared image makes sense.  tests never do.

martinpitt:
run-tests and check-* support `--machine` and `--browser` to run against an already existing testbed. That's what we use in packit, downstream gating, and even locally to do quick iterations on an already running vm-run VM. Of course these can be handled specially. Aside from that, I agree that it wouldn't hurt to check this in check-*. run-tests OTOH is being shared with starter-kit and friends, and there it may actually make sense to run this against a pristine VM image -- IMHO we should not unnecessarily restrict that.

allisonkarlitskaya:
The "did image-prepare get run?" question is #15954.

I had a thought to do this from `test_main()` which would actually be a great place for it.  The only problem is that before we get to `test_main()`, we already import a lot of stuff from `bots/`.  This could maybe be munged about a bit, but it's probably not gonna be pretty...

martinpitt:
Relaying from IRC: replacing it with a descriptive assertion works for me, yes; that may be more useful than "testvm: no such module"

allisonkarlitskaya:
I actually don't like this solution.  Because of the mentioned problem with `testvm` being imported so early in the file, we basically have to settle for one of the following:

 - subprocess call right at the top of the file before the `import` lines in `image-prepare` and `run-tests`; or
 - `image-prepare` failing on fresh checkout (which is the status quo, but would be cool to fix)

martinpitt:
@allisonkarlitskaya "This" solution == what this PR does? Or the proposal to print a meaningful "use make-bots, Luke!" message instead of an ImportError?

allisonkarlitskaya:
> @allisonkarlitskaya "This" solution == what this PR does? Or the proposal to print a meaningful "use make-bots, Luke!" message instead of an ImportError?

I mean that I don't like the error message idea.

I'm not really totally sure where to go from here, to be honest.  I find myself increasingly developing node features without running autogen.  `webpack-jumpstart` and `pkg/build` are seriously useful/powerful.  The thing that bothers me most about this new lifestyle is that `bots/` is always missing :)

martinpitt:
I'm afraid I still don't understand/know in what particular situation this gets in the way. Jumpstarting, (re)building dist/, doing C or JS development should not need bots/ -- if it does, let's patch it up with intercepting `ImportError` and disabling some features.

For running tests against a cockpit CI VM you need image-prepare (which could do the make-bots call). The only case where this is not the case is packit and gating, which runs our tests against an existing VM *outside* of our bots VM machinery -- that certainly does need to grab bots/, but at the same time these have  long scripts which can (and do) care about that.

Can you please describe the steps that you want to do starting from a clean checkout? Thanks!

allisonkarlitskaya:
> I'm afraid I still don't understand/know in what particular situation this gets in the way. Jumpstarting, (re)building dist/, doing C or JS development should not need bots/ -- if it does, let's patch it up with intercepting `ImportError` and disabling some features.

It's all about manual usage of the scripts in `bots/` to do stuff like `tests-trigger` and so on...  In fairness, my last comment was pretty confusing and orthogonal to what I'm doing in this PR...



> For running tests against a cockpit CI VM you need image-prepare (which could do the make-bots call).

This is exactly the problem: it can't.  `image-prepare` imports `testvm`, so `bots/` needs to already be there at the time the `import`s are run.  adding a call at the top of `main()` is already too late.

imho `parent.py` is a pretty magical script which is already doing some Very Evil things.  I don't think this particular addition is too much of a stretch, and I really can't imagine a better way... unless you want the subprocess call right at the top of image-prepare, before the imports...

martinpitt:
Thanks for the clarification! Indeed for calling bots/foo without bots/  there is not much that we can do, as there is nothing to hook into. (Of course you can just call tests-trigger and friends from the bots checkout, but that then needs `--repo`).

>  unless you want the subprocess call right at the top of image-prepare, before the imports...

Yes, I think that'd be much less evil than doing this in parent.py.

allisonkarlitskaya:
Thanks for nagging on this.  I think I'm pretty happy with the compromise solution that I just pushed, and I imagine you will be too.

martinpitt (approved):
This looks really nice now, great solution! And thanks for cleaning up that very stale import code in image-prepare.

martinpitt:
Bah:

    ModuleNotFoundError: No module named 'testvm'

martinpitt (changes requested):
breaks tests

martinpitt (approved):
Danke!

allisonkarlitskaya:
ps: my official solution to the "I like to run scripts out of `bots/`" problem was to set up some shell aliases to run them out of my bots main branch checkout instead.  That's even nicer than it was before, now.

PR #15952, opened 2021-06-15
test: Re-fix image-prepare with --overlay

martinpitt:
Commit 779d7f08936ba caused three regressions:

 * Crash with `UnboundLocalError: local variable 'qcow2_image' referenced before assignment`
   with `--overlay`, if the target image was already existing (as it
   happens with preparing the `dnf-copr` scenario). In that case
   prepare_install_image() does not create any new overlay, so return
   None instead.

 * Crash due to the target symlinking already existing with `--overlay`,
   as the previous os.unlink() was not moved along with the os.symlink()
   move.

 * `VirtMachine` needs to be called on the .qcow2 overlay file
   instead of the target symlink, as the latter does not exist yet
   without `--overlay`.

Factorize the logic into a new finalize_image().

Also rename the return vaules of prepare_install_image() to something
more descriptive.

-----

I tested this with three scenarios:

 * No existing test/images/, standard `test/image-prepare fedora-34`
 * Install-only with pre-built rpms: `test/image-prepare -vI fedora-34`
 * dnf-copr-like scenario with existing image and `--overlay`:

       bots/image-customize -v --run-command 'touch /pwned' fedora-34
       test/image-prepare -qv --overlay fedora-34

These now all behave as expected. I will also trigger an additional dnf-copr scenario here, which is the one that broke in https://github.com/cockpit-project/bots/pull/2114

### DIFF
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -119,9 +119,9 @@ def prepare_install_image(base_image, overlay=False):
         subprocess.check_call(["qemu-img", "create", "-q", "-f", "qcow2",
                                "-o", "backing_file={0},backing_fmt=qcow2".format(base_image),
                                qcow2_image])
-        if os.path.lexists(install_image):
-            os.unlink(install_image)
-    return install_image, qcow2_image
+        return install_image, qcow2_image
+    else:
+        return install_image, None
 
 
 def run_install_script(machine, do_build, do_install, skips, arg, args):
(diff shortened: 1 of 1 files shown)


martinpitt:
@smwltr FYI in case you want to try this.

KKoukiou (approved):

PR #15954, opened 2021-06-15
testlib: force use of prepared images

allisonkarlitskaya:
For images not explicitly specified by a provision, assume that we want
the image to have been created by test/image-prepare.

Throw an early exception with a useful message, for ourselves, if it
doesn't exist.

Note that passing a pathname as the image name is explicitly supported,
and the basename (without extension) will be used for setting the
Machine.image variable.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -853,15 +853,17 @@ class MachineCase(unittest.TestCase):
 
     def new_machine(self, image=None, forward={}, restrict=True, cleanup=True, **kwargs):
         machine_class = self.machine_class
-        if image is None:
-            image = self.image
         if opts.address:
             if machine_class or forward:
                 raise unittest.SkipTest("Cannot run this test when specific machine address is specified")
-            machine = testvm.Machine(address=opts.address, image=image, verbose=opts.trace, browser=opts.browser)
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice guidance, thanks!

martinpitt (changes requested):
Oops, no, this check needs to be moved down when it actually instantiates a virt machine. This must work with `--address`/`--browser` (see packit test failure)

test/common/testlib.py
```diff
@@ -854,7 +854,9 @@ def label(self):
     def new_machine(self, image=None, forward={}, restrict=True, cleanup=True, **kwargs):
         machine_class = self.machine_class
         if image is None:
-            image = self.image
+            image = os.path.join(TEST_DIR, "images", self.image)
+            if not os.path.exists(image):
+                raise FileNotFoundError("Can't run tests without a prepared image; use test/image-prepare")
```
martinpitt, line 859: Move this after the "else:" of `if opts.address`, then it should be good.

allisonkarlitskaya: Thanks.  I added back the `image or self.image` logic for the `--machine` case.

martinpitt (approved):
Danke! This looks better.

allisonkarlitskaya:
The rhel packit fails are test fails, not testing framework fails :)

PR #15955, opened 2021-06-15
networking: Add first pixel tests :rocket: 

KKoukiou:
And init test/reference external repository.

### DIFF
--- a/test/run
+++ b/test/run
@@ -6,6 +6,7 @@
 set -eu
 
 tools/make-bots
+test/common/pixel-tests pull
 
 TEST_SCENARIO=${TEST_SCENARIO:-verify}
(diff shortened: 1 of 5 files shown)


test/run
```diff
@@ -6,6 +6,7 @@
 set -eu
 
 tools/make-bots
+test/common/pixel-tests pull
```
martinpitt, line 9: Local developers would need that as well, right? We don't ever use test/run locally, and this is easy to forget. In c-machines we have a Makefile rule for it and depend on it by `check`. For cockpit I think it could either be done by autogen.sh (which already installs bots) or test/image-prepare.

Would also be good to add a pointer to test/README.md.

KKoukiou: Good point!

mvollmer (approved):

PR #15956, opened 2021-06-15
lib: table: parameterize expandId and contentId with the table id

### LINKED ISSUE #15193: network: Duplicate ids
On the Network page, these a11y issues should be fixed:

- [x] Logs have role-related issues, MM - this will be fixed as part of #15191 as it uses the same renderer
- [x] "Receiving" `<td>`s have `role="cell&quot;"`. It's odd that they have a role of "cell", but no matter what, shouldn't have the `&quot;`
- [x] Firewall switch lacks a label. Either it should have a `<label>` (with a for, or wrapping the widget), aria-label, or aria-labelledby. Additionally, it's our custom switch widget (`onoff-ct`) and should be ported to PF4's switch widget.

Firewall section:
- [ ] duplicate IDs — an ID should be unique on a page. Examples: expandable-toggle0, expandable-toggle2, etc. and expanded-content1, expanded-content3, etc. See https://github.com/patternfly/patternfly-react/issues/5690 https://github.com/cockpit-project/cockpit/pull/15956
    - example generated DOM: `<button aria-labelledby="simple-node0 expandable-toggle0" id="expandable-toggle0" aria-expanded="false" aria-disabled="false" aria-label="Details" class="pf-c-button pf-m-plain" type="button" data-ouia-component-type="PF4/Button" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-Button-plain-1">`
    - example DOM, part 2: `<td class="" colspan="4" data-key="0" id="expanded-content1">`
- [x] Firewall switch lacks a label. Either it should have a `<label>` (with a for, or wrapping the widget), aria-label, or aria-labelledby. Additionally, it's our custom switch widget (`onoff-ct`) and should be ported to PF4's switch widget. (Same as overview.)

KKoukiou:
This fixes the duplicate Ids issues for expandable rows in networking firewall page.

Fixes #15193

### DIFF
--- a/pkg/networkmanager/firewall.jsx
+++ b/pkg/networkmanager/firewall.jsx
@@ -184,6 +184,7 @@ function ZoneSection(props) {
         {props.zone.services.length > 0 &&
         <CardBody className="contains-list">
             <ListingTable columns={[{ title: _("Service"), props: { width: 40 } }, { title: _("TCP"), props: { width: 30 } }, { title: _("UDP"), props: { width: 30 } }]}
+                          id={props.zone.id}
                           aria-label={props.zone.id}
                           variant="compact"
                           emptyCaption={_("There are no active services in this zone")}
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -204,8 +204,11 @@ export class ListingTable extends React.Component {
         if (this.props.rows.length == 0)
             tableProps.className += ' ct-table-empty';
 
-        if (this.props.id)
+        if (this.props.id) {
             tableProps.id = this.props.id;
+            tableProps.expandId = this.props.id + '-expandable-toggle';
+            tableProps.contentId = this.props.id + '-expanded-content';
+        }
 
         if (this.props.variant)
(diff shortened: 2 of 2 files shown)


pkg/networkmanager/firewall.jsx
```diff
@@ -184,6 +184,7 @@ function ZoneSection(props) {
         {props.zone.services.length > 0 &&
         <CardBody className="contains-list">
             <ListingTable columns={[{ title: _("Service"), props: { width: 40 } }, { title: _("TCP"), props: { width: 30 } }, { title: _("UDP"), props: { width: 30 } }]}
+                          id={props.zone.id}
```
martinpitt, line 187: This looks curious at first sight, as `id=` is already a standard HTML attribute. But it does end up as actual `id` of the generated `<table>`, so seems ok.

martinpitt (approved):
Danke! I wonder what went wrong with packit (missing publish-dist sha), I'll retry once.

martinpitt:
packit: "The cached package-lock.json doesn't match our own".  It's not wrong, master advanced to webpack 5 in the meantime. However, ok to ignore here.

Issue #15927, opened 2021-06-10
Create timer modal improvements (SPUR)


marusak:
- [x] `After specific boot` needs validation (try putting there "`-x` seconds")
- [x] `Repeat hourly` should be better explained (MM did not get it takes minutes, maybe specify somewhere that minutes are needed)
- [x] `Repeat hourly` wraps +- buttons to different line when it does not validate [1]
- [x] In `Repeat yearly` height of date picker is taller than of hour picker - they don’t align [2] (not for all, for MM they alight, for KK they don't, check development builds on different browsers) (kk: can't reproduce any more)
- [x] +- button don’t fit onto one line on iphone SE [4] 
- [x] +- button do not align with selectors [3] (#16618)
- [ ] On small screens the ends of date and hour dropdowns are partly hidden behind shell [5] (Reported upstream, I don't see how we can fix it locally https://github.com/patternfly/patternfly-react/issues/6616)

Questions:
- [x] `Repeat monthly` and pick `31` - would that work for months with 30 days? (or is that just out of our scope?) 
- [ ] Icons in +- buttons are not perfectly centered? (we were not exactly sure, lets check it out - if it is an issue move to the list above) (kk: can't see the problem here)

[1]
![Screenshot from 2021-06-10 15-22-5670](https://user-images.githubusercontent.com/12330670/121532570-c2faec80-c9ff-11eb-9aae-86798e71b494.png)

[2] 
![Screen Shot 2021-06-11 at 14 24 27](https://user-images.githubusercontent.com/12330670/121533647-c478e480-ca00-11eb-8714-72228395ae71.png)

[3]
![Screenshot from 2021-06-10 15-29-02](https://user-images.githubusercontent.com/12330670/121533826-f12cfc00-ca00-11eb-898a-d2057f5322d8.png)

[4]
![Screen Shot 2021-06-11 at 15 33 35](https://user-images.githubusercontent.com/12330670/121534165-42d58680-ca01-11eb-9996-91d2bf86a9e7.png)

[5]
![Screen Shot 2021-06-11 at 15 33 44](https://user-images.githubusercontent.com/12330670/121534380-74e6e880-ca01-11eb-85a0-914fc748c6ee.png)

KKoukiou:
Regarding the last point:
> Icons in +- buttons are not perfectly centered? (we were not exactly sure, lets check it out - if it is an issue move to the list above)

[...]

PR #15958, opened 2021-06-16
Some create timer modal improvements (SPUR)

KKoukiou:
Partially addresses https://github.com/cockpit-project/cockpit/issues/15927


garrett:
Dropdowns are center aligned... and I'm not sure why.

Screenshot from 2021-06-16 13-24-30

Hourly timers are still non-obvious. (I guess that's minute offset?)

image

- We're creating a single timer in the UI, so the dialog should be "Create timer", not "Create timers".
- Is there a limit or a recommended length on service names? If so, we should show that in the UI. 
- Are there only ever two options for "Run"? It should probably be renamed "When" as it's after system boot or at a specific time.
- After system boot is a delay, right? It should probably be called "Delay"?

Some sketches:

service-timer-modal(1)

service-timer-modal(2)

service-timer-modal(4)

garrett:
I'm fine if this PR just adjusts the list above, including the simle stuff like alignment and the text strings. UI changes in this case are mainly superfluous and can go in a follow-up or just be ignored, as what we have isn't so bad (except for the lack of communication of what that text entry is for hourly mode and the "Run"/"When" dropdown for just two items).

KKoukiou:
> I'm fine if this PR just adjusts the list above, including the simle stuff like alignment and the text strings. UI changes in this case are mainly superfluous and can go in a follow-up or just be ignored, as what we have isn't so bad (except for the lack of communication of what that text entry is for hourly mode and the "Run"/"When" dropdown for just two items).

I can fix your suggestions in this PR - it's about getting the perfect from this anyway. When we are satisfied I will also add a pixel test to make sure we don't break it.

KKoukiou:
@garrett there are still some unsolved issue from https://github.com/cockpit-project/cockpit/issues/15927#issuecomment-862108330 which are not so trivial.

Can you please review this again anyway? and maybe we can split the enhancements in multiple PRs.

garrett (approved):
I like the changes and I'm happy with splitting off the rest to another PR.

There is a little problem with the layout in super-small sizes (iPhone SE), but everything else is fine.

Screen Shot 2021-06-17 at 16 48 01


One somewhat notable thing: When adding new items, they're below the scroll. We should probably instruct the browser to scroll down to the bottom when adding them. When there's no overflow, it shouldn't do anything. This mainly affects smaller screens like mobile.

But, overall, happy to see a lot of this custom code go away, and the rest can be addressed in follow-up PR(s).

garrett (changes requested):
Oh, I just noticed the radio button labels aren't clickable. This means it's also broken for accessibility.

Screenshot from 2021-06-17 17-01-01

The label exists, but it doesn't have a for="" attribute pointing to the ID of the `<input>`... and the input also doesn't have a generated unique ID.

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -159,153 +164,189 @@ const CreateTimerDialogBody = ({ setIsOpen }) => {
...
+                        <Radio value="system-boot"
+                               name="boot-or-specific-time"
+                               onChange={() => setDelay("system-boot")}
+                               isChecked={delay == "system-boot"}
+                               label={_("After system boot")} />
+                        <Radio value="specific-time"
+                               name="boot-or-specific-time"
+                               onChange={() => setDelay("specific-time")}
+                               isChecked={delay == "specific-time"}
+                               label={_("At specific time")} />
```
garrett, line 178: I guess leaving out an ID doesn't make the radios get hooked up properly in PatternFly, so each needs its own unique ID.

This really should be a compile time error for PF, or PF should wrap the label around the radio. Either-or, but right now, neither happens, so we need to manually ensure there's an ID.

garrett (dismissed):
1. Dropdowns look fine (that was my fault, and on my machine only)
   Screenshot from 2021-06-21 16-09-06
2. Labels are now fixed for a11y and are clickable!

LGTM! :+1:

garrett (approved):

PR #15959, opened 2021-06-16
    test/run: add a "/devel" scenario

allisonkarlitskaya:
Remove the logic of setting `NODE_ENV=development` for default
    `TEST_OS`, and set it according to the addition of a new test scenario
    called `/devel`.
    
    Add also the option for `/firefox-devel` to be able to test a devel
    build against Firefox.

### DIFF
--- a/test/run
+++ b/test/run
@@ -11,11 +11,8 @@ TEST_SCENARIO=${TEST_SCENARIO:-verify}
 TEST_OS_DEFAULT=$(PYTHONPATH=bots python3 -c 'from machine.machine_core.constants import TEST_OS_DEFAULT; print(TEST_OS_DEFAULT)')
 
 case $TEST_SCENARIO in
-    verify|firefox)
-        # do a development build on the latest Fedora image so that we can spot React and other dev-only warnings
-        if [ "$TEST_OS" = "$TEST_OS_DEFAULT" ]; then
-            export NODE_ENV=development
-        fi
+    verify|devel|firefox|firefox-devel)
+        [ "${TEST_SCENARIO}" = "${TEST_SCENARIO%%devel}" ] || export NODE_ENV=development
         [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##firefox}" ] || export TEST_BROWSER=firefox
         test/image-prepare --verbose $TEST_OS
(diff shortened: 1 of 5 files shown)


martinpitt:
I manually triggered the fedora-34/devel context here with https://github.com/cockpit-project/bots/pull/2120, to validate this.

martinpitt:
packit fails with "The cached package-lock.json doesn't match our own", presumably this branch is out of date? But let's wait for the integration test runs, they are a bit non-trivial to start.

test/run
```diff
@@ -11,11 +11,8 @@ TEST_SCENARIO=${TEST_SCENARIO:-verify}
...
 
 case $TEST_SCENARIO in
-    verify|firefox)
-        # do a development build on the latest Fedora image so that we can spot React and other dev-only warnings
-        if [ "$TEST_OS" = "$TEST_OS_DEFAULT" ]; then
-            export NODE_ENV=development
-        fi
+    verify|devel|firefox|devel+firefox)
+        [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##devel}" ] || export NODE_ENV=development
         [ "${TEST_SCENARIO}" = "${TEST_SCENARIO%%firefox}" ] || export TEST_BROWSER=firefox
```
martinpitt: FTR, you can be a bit more liberal here, and use `${TEST_SCENARIO##*devel}` to support any ordering, or extend this to a third scenario in the future. But not important right now.

martinpitt (dismissed):
Thanks! LGTM, if it works.

allisonkarlitskaya:
There's a bunch of places around the tests that hardcode the assumption that the default OS is a devel build.  I need to clean those up still.

martinpitt:
FTR, let's avoid `+` as part of context names. `+` is the URL encoding of a space, and our sink logic isn't clever enough to quote this appropriately. "firefox-devel" or "devel-firefox" seems quite fine.

allisonkarlitskaya:
`AttributeError: 'VirtMachine' object has no attribute 'disable_preload'`

uh, oops

KKoukiou (approved):

allisonkarlitskaya:
packit is unhappy, but all of the new scenarios are working correctly.

PR #15967, opened 2021-06-17
webpack-make refactors

allisonkarlitskaya:
Here's a few commits to tidy up `webpack-make` and teach it a more modern dialect of JS.

These are pretty much exclusively focused around the `Makefile.deps` generation code.

This PR doesn't change the result.  It's pure cleanup, with no side-effects:

```
⬢[lis@f34 cockpit]$ diff -ur master/dist webpack-make-refactors/dist
⬢[lis@f34 cockpit]$ 
```

### DIFF
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -76,9 +76,17 @@ function generateDeps(makefile, stats) {
     var pkgdir = path.dirname(makefile);
     var stampfile = pkgdir + '/manifest.json';
 
-    stats.compilation.fileDependencies.forEach(function(file) {
-        maybePushInput(inputs, file);
-    });
+    for (const file of stats.compilation.fileDependencies) {
+        // node modules  are handled by the dependency on package-lock.json
+        if (file.includes('/node_modules/'))
+            continue;
+
+        // Webpack 5 includes directories: https://github.com/webpack/webpack/issues/11971
+        if (fs.lstatSync(file).isDirectory())
+            continue;
+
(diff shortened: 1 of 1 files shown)


tools/webpack-make
```diff
@@ -100,38 +100,35 @@ function generateDeps(makefile, stats) {
...
         if (!output.endsWith("/manifest.json") && !output.endsWith(".map"))
             outputs[output] = output;
 
-	if (output.indexOf("/test-") !== -1 && output.endsWith(".html")) {
-            tests[output] = output;
+        if (output.includes("/test-")) {
+            if (output.endsWith(".html")) {
+                tests[output] = output;
+            }
             continue;
```
martinpitt: This is a logic change. Now it continues for any test-* file which does not end with .html. I did not think about whether the old or new version is correct.

martinpitt (approved):
Danke, this looks nice! Please diff the built rpms to make sure that we don't include new files like test-foo.js, due to the changed filtering logic (see below).

tools/webpack-make
allisonkarlitskaya: It `continue;`s for all `test-*` files.  That's intended: everything below that point is about dealing with install paths.

This really is a straight refactor: there are no changes to the output here.

allisonkarlitskaya: (I was so careful about testing it, because changing this logic has burned us in the past).

PR #15969, opened 2021-06-17
build: handle .gz file output explicitly

allisonkarlitskaya:
Using make to implicitly generate .gz output at `make install` time has
    two major problems:
    
     1) in srcdir!=builddir builds, these are the only files that land in
        the builddir/dist/ directory, since all of the webpack outputs come
        already shipped with the tarball
    
     2) since `make install` is often run as root, these files end up
        in your build tree, owned by root
    
    At the same time, we don't want to integrate that into the webpack
    process, because it would slow things down during development.
    
    Move the .gz outputs to a separate variable in webpack-make, and deal
    with them in a more explicit fashion from the Makefile, from an install
    hook.

 - [x] #15967

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,31 +36,18 @@ clean-local::
 	find $(srcdir) -name '*.pyc' -delete
 
 SUFFIXES = \
-	.css .css.gz \
-	.html .html.gz \
-	.js .js.gz \
 	.jsx \
-	.map .map.gz \
 	.mo .po \
 	.service .service.in \
 	.socket .socket.in \
-	.svg .svg.gz \
 	.1 .8 .5 \
 	$(NULL)
-
-V_GZIP = $(V_GZIP_$(V))
-V_GZIP_ = $(V_GZIP_$(AM_DEFAULT_VERBOSITY))
-V_GZIP_0 = @echo "  GZIP    " $@;
 
 MV = mv -f
 
 CAT_RULE = \
         $(AM_V_GEN) $(MKDIR_P) $(dir $@) && \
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -64,24 +64,16 @@ GZ_RULE = \
...
-	-e 's,[@]sysconfdir[@],$(sysconfdir),g' \
-	-e 's,[@]libdir[@],$(libdir),g' \
-	-e 's,[@]includedir[@],$(includedir),g' \
-	-e 's,[@]bindir[@],$(bindir),g' \
-	-e 's,[@]sbindir[@],$(sbindir),g' \
-	-e 's,[@]prefix[@],$(prefix),g' \
-	-e 's,[@]exec_prefix[@],$(exec_prefix),g' \
 	-e 's,[@]prefix[@],$(prefix),g' \
-	-e 's,[@]PACKAGE[@],$(PACKAGE),g' \
-	-e 's,[@]VERSION[@],$(VERSION),g' \
```
martinpitt, line 90: doc/guide/version.in uses `@VERSION`, and it IMHO makes sense there. Please keep this. I checked the others, and they are indeed unused.

martinpitt (changes requested):
Thanks, nice cleanup! Just one gotcha.

Makefile.am
allisonkarlitskaya: `doc/guide/version.in` is substituted from `configure.ac`, not the `$(SUBST_RULE)`.

martinpitt (approved):

PR #15971, opened 2021-06-18
webpack-watch: rework path handling a bit

allisonkarlitskaya:
webpack-watch currently hands the following option to webpack-make:

  -d "$mydir/../dist/$1/Makefile.deps"

which results in "tools/../dist/" appearing in several places in the
generated Makefile.deps, eg:

  static: tools/../dist/static/manifest.json

Use the usual $(realpath -m) tricks to change our working directory to
the top_srcdir and call webpack from there with a fixed relative path,
the same way as pkg/build does.

### DIFF
--- a/tools/webpack-watch
+++ b/tools/webpack-watch
@@ -2,7 +2,7 @@
 # Calls webpack-make in watch mode for the given pkg/<page>
 
 set -eu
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Danke! Triggering required f34 test.

PR #15978, opened 2021-06-22
github: Add pixeltest-review workflow

mvollmer:


allisonkarlitskaya (changes requested):
This seems very complicated.

Probably you can do the same with a single shell job which fetches the head and the base references, and uses something like reposchutz does:

```yaml
    env:
      GIT_DIR: git-dir.git
      HEAD_SHA: ${{ github.event.pull_request.head.sha }}
      BASE_SHA: ${{ github.event.pull_request.base.sha }}

    steps:
        run: |
          git init -b main
          git remote add origin "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}"
          git fetch origin "${BASE_SHA}" "${HEAD_SHA}"
          git log --full-history --exit-code --patch "${HEAD_SHA}" --not "${BASE_SHA}" -- test/reference
```

See https://github.com/cockpit-project/cockpit/blob/master/.github/workflows/reposchutz.yml#L52

There is then a slight mismatch between the fact that you have this result in the shell world but might like to post a comment from Javascript world, but there are many ways to bridge this gap.  Let's discuss on IRC.

Finally: it may make sense to integrate this into reposchutz.  The check is very much of the same flavour as the checks being performed there.  It would be a good fit, and would save spinning up yet another VM.

allisonkarlitskaya:
One more thing I forgot to mention: writing a comment every time anything happens in the PR is probably too much.  Even writing a comment on every force push might be too much.  We might want to figure out a way to decrease the frequency somehow.

mvollmer:
> One more thing I forgot to mention: writing a comment every time anything happens in the PR is probably too much.

Yes, we could inspect the old comments on the PR and only add a new one if there are real changes. Or we might let the workflow edit the description, and add a line at the end that points to the test/reference diff.

mvollmer:
> Finally: it may make sense to integrate this into reposchutz. The check is very much of the same flavour as the checks being performed there. It would be a good fit, and would save spinning up yet another VM.

I'll check this out.  But note that we also want these comments in external projects, like cockpit-machines. Not all of them might have or want reposchutz.

mvollmer:
> There is then a slight mismatch between the fact that you have this result in the shell world but might like to post a comment from Javascript world, but there are many ways to bridge this gap.

Also, the comment needs the values of head.sha and base.sha to construct the diff URL.

allisonkarlitskaya:
> Yes, we could inspect the old comments on the PR and only add a new one if there are real changes. Or we might let the workflow edit the description, and add a line at the end that points to the test/reference diff.

I like this idea.

It should also be possible to post a status with a link to the diff.  I've never tried that, but the API is there for it (think: similar to what `tests-trigger` does).

allisonkarlitskaya:
> Also, the comment needs the values of head.sha and base.sha to construct the diff URL.

[assuming you mean the gitlink value from head.sha and base.sha respectively]

I didn't account for this in my review, indeed.

martinpitt:
@mvollmer , @KKoukiou, @allisonkarlitskaya : Given that this has been stalled for so long, and by now we all figured out how to look at the ref image changes directly on GitHub (even though this is a bit fiddly, due to having to replace `...` with `..` in the URLs), should we just close this? (+1 from me)

KKoukiou:
> @mvollmer , @KKoukiou, @allisonkarlitskaya : Given that this has been stalled for so long, and by now we all figured out how to look at the ref image changes directly on GitHub (even though this is a bit fiddly, due to having to replace `...` with `..` in the URLs), should we just close this? (+1 from me)

Yes let's close this. I will do the removal from cockpit-machines for being consistent in this case.

PR #15984, opened 2021-06-23
packit.yaml: stop testing on rawhide

allisonkarlitskaya:
This has been pretty consistently red lately, for reasons outside of our
control.  Take it out of the list for now so that we don't get too used
to merging red PRs.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -19,5 +19,4 @@ jobs:
       targets:
       - fedora-33
       - fedora-34
-      - fedora-rawhide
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Agreed. FTR, I added this back then as an experimental thing, and it used to be green for quite a while. But indeed we already spend so much time on chasing regressions in "stable" Fedoras, looking even further ahead has diminishing returns for us. We need to fix that on the distro level (gating tests in Fedora and such).

PR #15991, opened 2021-06-24
test: Disable font hinting with Chromium

mvollmer:
Hinting seems to involve some randomness and might produce slightly
different glyphs on different runs.  This is trouble for the pixel
tests.  Let's switch it off.


KKoukiou:
@mvollmer I bet you need to update the existing screenshots in this repo within this PR, otherwise they will fail.

KKoukiou (approved):

PR #15994, opened 2021-06-24
lib/page.css: drop custom ct-select in favor of the PF4 solution

KKoukiou:


martinpitt:
Lots of networking related test flakes, as this triggers the "affected must succeed 3 times" logic. One round of retries.

martinpitt:
Logs page on master:

image
image

Logs page on this PR looks exactly the same to me:

image

Thanks!

martinpitt (approved):

martinpitt:
Two flakes on [ubuntu-2004](https://logs.cockpit-project.org/logs/pull-15994-20210624-164302-0e815c48-ubuntu-2004/log.html), unrelated. Good enough, these are hard to get to green with 3x runs, and no need to block this.

PR #15995, opened 2021-06-25
move pixel-tests to use git-utils.sh, including caching

allisonkarlitskaya:


Makefile.am
```diff
@@ -28,6 +28,7 @@ EXTRA_DIST = \
 	package.json \
 	package-lock.json \
 	README.md \
+	pkg/lib/git-utils.sh \
```
martinpitt, line 31: Is it actually important to ship this now, when we haven't so far? (FTR, I'm fine with it, just wondering if that is coincidental, or deliberate)

pkg/lib/git-utils.sh
```diff
@@ -10,6 +10,7 @@
 GITHUB_BASE="${GITHUB_BASE:-cockpit-project/cockpit}"
 GITHUB_REPOSITORY="${GITHUB_BASE%/*}/${GITHUB_REPO}"
 HTTPS_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
+SSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
```
martinpitt, line 13: FTR, :+1 for calling this SSH_REMOTE instead of PUSH_REMOTE, as pushing may happen over https as well.

test/common/pixel-tests
```diff
@@ -1,25 +1,35 @@
...
-GITHUB_REPOSITORY="${GITHUB_BASE%/*}/${REPO}"
-CLONE_REMOTE="https://github.com/${GITHUB_REPOSITORY}"
-PUSH_REMOTE="git@github.com:${GITHUB_REPOSITORY}"
+GITHUB_REPO='pixel-test-reference'
+SUBDIR='test/reference'
 
-message() {
-    [ "${V-}" != 0 ] || printf "  %-8s %s\n" "$1" "$2"
-}
+set -eu
```
martinpitt, line 6: Tiny style nitpick: This ought to stay the first line in this (and every) script. Presumably it just got moved by accident?

test/common/pixel-tests
```diff
@@ -1,25 +1,35 @@
...
-    git submodule update --init -- "$SUBDIR"
+    # Our main goal is to ensure that the test/reference from
+    # the index is the one that we actually have.
+    local sha="$(get_index_gitlink test/reference)"
+    if [ ! -e test/reference/.git ]; then
+        # nothing there yet...
+        cmd_checkout "${sha}"
+    elif [ "$(git -C test/reference rev-parse HEAD)" != "${sha}" ]; then
+        # wrong thing there...
+        cmd_checkout "${sha}"
```
martinpitt, line 31: Looks a bit odd to have two branches with exactly the same code. This could just `||` the condition.

martinpitt (commented):
Thanks! A few tiny style nitpicks. As this adds a whole lot of new code, the commit message could do with some rationale. Supposedly this improves/introdudes the local caching, whereas the original script always pulled it fresh from the internet?

Let's give @mvollmer a chance to review this. If he doesn't get to it before PTO, I think I feel reasonably qualified to ack/land this :grin:

test/common/pixel-tests
```diff
@@ -1,25 +1,35 @@
...
 
 cmd_init() {
     git submodule add -b empty "$CLONE_REMOTE" "$SUBDIR"
 }
 
+cmd_checkout() {
+    local sha="${1-$(get_index_gitlink node_modules)}"
+
+    fetch_sha_to_cache "${sha}"
+    cmd_remove
```
mvollmer, line 18: I am not sure I like how `pixel-tests pull` might now delete test/reference completely. This will remove untracked files, while `git checkout` and `git submodule update` leave those alone.

On the one hand, removing untracked files will make it impossible for the `pull` to fail because of conflicts, on the other hand, it might throw files away that the developer cares about.

mvollmer (approved):
Approved, since I mostly care about the `status` and `push` sub-commands. They are unique to the pixel-tests workflow, I'd say, and if we make `pull` or `checkout` more similar to the rest of our tools, I can't disagree.

I was taking the position that if you know submodules, test/common/pixel-tests should offer no surprises for you. This might no longer be true with this change, but I can't say I have a strong opinion. We should update the blog post if necessary.

test/common/testlib.py
```diff
@@ -662,6 +662,9 @@ def assert_pixels(self, selector, key, ignore=[]):
         if not (Image and self.pixels_label and self.cdp and self.cdp.valid):
             return
 
+        # Make sure everything is up to date.  This is normally very fast.
+        subprocess.check_call([f'{TEST_DIR}/common/pixel-tests', 'pull'])
```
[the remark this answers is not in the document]
allisonkarlitskaya: This is one of the reasons I wanted your review here.  What do you want to do about this?  Refuse to delete the old tree if it's not clean?

test/common/pixel-tests
allisonkarlitskaya: See comment above.  I'd like your input to decide what to do here :)

test/common/testlib.py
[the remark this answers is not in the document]
mvollmer: > Refuse to delete the old tree if it's not clean?

Yes, that sounds good. Also, have a `--force` option for `pull` to delete it anyway.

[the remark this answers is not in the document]
allisonkarlitskaya: I'll implement this.  Thanks.

allisonkarlitskaya:
We can bring this back later at some point, if we want.

PR #15997, opened 2021-06-25
build: don't "dist" test/reference

allisonkarlitskaya:
This is being accidentally included via the "commited dist" mechanism,
based on its presence in the git.  It doesn't make sense to do that,
though: this directory might contain reference images, be empty, or
might not exist at all.

So we probably actually need to discuss this: does it make sense to dist this, after all?  Does anyone do integration tests based on a tarball *on the host*?  What about the checks that say we only run the pixel tests on fedora-34?  Is this meaningful for packit?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -214,7 +214,7 @@ tools/debian/copyright: tools/debian/copyright.template $(MANIFESTS)
 # also automatically update minimum base dependency in RPM spec file
 dist-hook:: $(MANIFESTS)
 	( cd $(srcdir); git ls-tree HEAD --name-only -r $(COMMITTED_DIST) || (echo $(COMMITTED_DIST) | tr ' ' '\n' ) ) | \
-		tar --format=posix -C $(srcdir) -cf - -T - | tar -C $(distdir) -xf -
+		tar --format=posix -C $(srcdir) --exclude test/reference -cf - -T - | tar -C $(distdir) -xf -
 	echo $(VERSION) > $(distdir)/.tarball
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
The more I think about this, the more I think it's right:
 * if we start adding a lot of reference images then disting them is going to get increasingly painful
 * in particular, we don't seem to have the plan to have a separate repository per-project, so each project would then dist all images

We probably need to figure out a way for tarball releases to get the correct set of reference images, if they're interested in doing that.  Maybe we tag off the reference images, in the same way we do for `node_modules`?

martinpitt (approved):
At least for the time being I have a rather strong opinion that pixel tests should be done upstream only, and not shipped in releases. They are prone to fail when testing against newer browsers, and in general they should depend on code changes or npm modules, not on changes in OS packages (other than browsers).

This is a long-winded way to say "I fully agree", thank you!

mvollmer:
> So we probably actually need to discuss this: does it make sense to dist this, after all?

I think it makes sense, if the tests themselves are also in the dist tarball.  The tests are referencing the reference images, and are arguably broken without them.  And since the dist tarball has no indication of where to get the reference images from (it doesn't include .gitmodules and the test/reference gitink), the only option is to include test/reference.

If that's not a convincing argument, then we should make sure that assert_pixels just does nothing when test/reference as a whole does not exist. (But still "complain" about a new reference image when test/reference exists, but not the exact expected image in it.)

mvollmer:
> I think it makes sense,

But I actually added "--exclude test/reference" to cockpit-machines/Makefile when introducing the very first pixel test there... so much for consistency with myself... hmm.

PR #15998, opened 2021-06-25
metrics: Fix enabling pmlogger in Metrics Settings if pcp is not installed

martinpitt:
Previously this just failed with "unit pmlogger.service does not exist".
Instead, when enabling pmlogger, offer to install the package. Just like
with the button in the empty state, request a relogin afterwards to pick
up the new manifest.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1255,7 +1255,7 @@ class MetricsHistory extends React.Component {
         if (this.state.needsLogout)
             return <EmptyStatePanel
                         icon={ExclamationCircleIcon}
-                        title={_("You need to relogin to be able to see metrics")}
+                        title={_("You need to relogin to be able to see metrics history")}
                         action={<Button onClick={() => cockpit.logout(true)}>{_("Log out")}</Button>} />;
 
         if (cockpit.manifests && !cockpit.manifests.pcp)
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1059,7 +1059,6 @@ class MetricsHistory extends React.Component {
             isDatepickerOpened: false,
             selectedDate: null,
             packagekitExists: false,
-            needsLogout: false,
         };
 
         this.handleMoreData = this.handleMoreData.bind(this);
@@ -1130,9 +1129,7 @@ class MetricsHistory extends React.Component {
 
     handleInstall() {
         install_dialog("cockpit-pcp")
-                .then(() => {
-                    this.setState({ needsLogout: true });
-                })
+                .then(() => this.props.setNeedsLogout(true))
                 .catch(() => null); // ignore cancel
     }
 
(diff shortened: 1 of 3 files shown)


martinpitt:
In the integration test we only have a fake package. But I interactively tested that with the real Fedora packages, and it DTRT: After removing pcp, enabling pmlogger in cockpit, logging out and back in, history metrics appear automatically.

pkg/metrics/metrics.jsx
```diff
@@ -905,13 +905,27 @@ const PCPConfig = ({ buttonVariant, firewalldRequest }) => {
...
+        if (dialogLoggerValue && s_pmlogger.state === undefined && !needsLogout) {
+            debug("PCPConfig: request to enable pmlogger, but missing pmlogger.servicie, offering install; needs logout", needsLogout);
+            setDialogVisible(false);
+            install_dialog("cockpit-pcp")
+                    .then(() => {
+                        debug("PCPConfig: cockpit-pcp installation successsful");
+                        setNeedsLogout(true);
+                        handleSave();
+                    })
+                    .catch(() => null); // ignore cancel
```
KKoukiou, line 925: This will however ignore all exceptions not just cancelled promise, right?

martinpitt: Right, but the dialog already handles package install failures etc. internally (as inline alerts). `install_dialog()` itself is documented accordingly:

> The install_dialog function returns a promise that is fulfilled when the dialog closes after
> a successful installation.  The promise is rejected when the user cancels the dialog.

We use it the same way in pkg/packagekit/autoupdates.jsx or pkg/storaged/optional-panel.jsx.

But indeed I just saw that we don't handle cancel() in handleInstall() on the metrics page -- this causes an unhandled exception that appears in the log. I'll fix that in a separate commit.

KKoukiou: ack thanks

martinpitt:
There's some test failures (and in addition, that confusing situation where two tests run in parallel!), the missing cancel handling of the already existing install dialog, and a "servicie" typo in a comment, back to draft.

martinpitt:
Indeed the logic to not recursively trigger the install dialog [was broken](https://logs.cockpit-project.org/logs/pull-15998-20210625-125357-183ec17b-debian-testing/log.html#272). Fixed now, and fixed another race condition in the test.

martinpitt:
There was still a race condition in the Install → apply Save handover, which I noticed in the on-demand redis installation. This ought to work better, and extend to other packages.

martinpitt:
Oops, this [causes some unexpected fallout](https://logs.cockpit-project.org/logs/pull-15998-20210628-105155-dd17683f-centos-8-stream/log.html#147), I suppose the test should better stop pmlogger in the end.

martinpitt:
@KKoukiou : Ready to review for real now.

KKoukiou (approved):

PR #15999, opened 2021-06-25
cockpit-session: drop PAM module

allisonkarlitskaya:
Stop the split-out of the client certificate authentication to a PAM
module.  The original reasons for doing so are no longer valid, and the
module design complicates passing additional information into the
authentication step.

The bulk of the logic in the PAM module is moved to a new file,
cockpit-session-client-certificate.[ch].  The PAM module is left around
as a stub, for those who still include it in their configs, but it is
removed from our PAM configs, and the manpage has been dropped.

Some adjustments are made: we use the debug() macro from session-utils
and generally move over to logging to stderr instead of syslog.  We also
replace pam status codes with success bools.

TODO:
 - [x] packaging support for migrating existing configs
 - [x] add back the docs somewhere appropriate

Release note:

## PAM: Deprecate `pam_cockpit_cert` module

The `pam_cockpit_cert` module is no longer required for client certificate authentication, and has been deprecated.  This release continues to ship a stub module which returns `PAM_IGNORE`, but this stub will be dropped in a future release.  If you modified  `/etc/pam.d/cockpit`, the package upgrade will not automatically update the file -- please remove the relevant entry manually in this case.


martinpitt:
The "cockpit-session: No matching user for certificate" failure looks like a missing allow_journal_message, or possibly a leftover debugging? Not sure whether these should be logged regularly, but with the PAM module the test doesn't suppress this kind of message.

src/tls/README.md
```diff
@@ -87,14 +87,21 @@ see the [guide](../../doc/guide/cert-authentication.xml) and
 [manpage](../../doc/man/cockpit.conf.xml). Then cockpit-tls will ask the
 browser for a client certificate.
 
-If cockpit-ws sees that cockpit-tls exports a certificate for its connection
-(by checking its cgroup instance name, which is the certificate fingerprint,
-and checking /run/cockpit/tls for it), then it will request the `tls-cert`
-authentication schema from cockpit-session, instead of the usual `basic` or
-`gssapi`. cockpit-session then does *not* set a PAM user name, and
```
martinpitt, line 94: I think this is still relevant -- you may still want to control/deviate authentication schemes in cockpit.conf.

allisonkarlitskaya:
> The "cockpit-session: No matching user for certificate" failure looks like a missing allow_journal_message, or possibly a leftover debugging? Not sure whether these should be logged regularly, but with the PAM module the test doesn't suppress this kind of message.

I cleaned that up in my last push.  It's a result of the change from syslog to journal-via-stderr logging mentioned in the commit message.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -0,0 +1,141 @@
...
+
+out:
+  sd_bus_error_free (&err);
+  sd_bus_message_unref (user_obj_msg);
+  sd_bus_unref (bus);
+  return result;
+}
+
+char *
+cockpit_session_client_certificate_check (void)
```
martinpitt, line 124: Bikeshedding: `cockpit_session_map_client_certificate_user()`? (Feel free to just resolve, but a _check() function sounds like it should return a bool or do an assert)

src/ws/session.c
```diff
@@ -550,17 +551,18 @@ perform_tlscert (const char *rhost)
 
   debug ("start tls-cert authentication for cockpit-ws %u", getppid ());
 
-  /* pam_cockpit_cert sets the user name from the certificate */
-  res = pam_start ("cockpit", NULL, &conv, &pamh);
+  char *username = cockpit_session_client_certificate_check ();
+  if (username == NULL)
+    exit_init_problem (PAM_AUTH_ERR);
+
+  res = pam_start ("cockpit", username, &conv, &pamh);
```
martinpitt, line 558: Here would be a good place to free `username` again. I suppose our valgrind check doesn't go across this suid binary, and it's short-lived, so maybe we can just ignore the leak.

tools/cockpit.spec
```diff
@@ -471,7 +471,6 @@ authentication via sssd/FreeIPA.
 %doc %{_mandir}/man8/cockpit-ws.8.gz
 %doc %{_mandir}/man8/cockpit-tls.8.gz
 %doc %{_mandir}/man8/remotectl.8.gz
-%doc %{_mandir}/man8/pam_cockpit_cert.8.gz
 %doc %{_mandir}/man8/pam_ssh_add.8.gz
 %dir %{_sysconfdir}/cockpit
 %config(noreplace) %{_sysconfdir}/cockpit/ws-certs.d
```
martinpitt, line 476: The `%config(noreplace) %{_sysconfdir}/pam.d/cockpit` in the next line bites a bit here. (See [docs](https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html)). It behaves similar to dpkg's conffiles. So at some point when we drop the PAM stub we need to apply some seddery in the `%post`/.postinst script to fix up the PAM file.

But for now I think we should add a grep to the post scripts that writes a deprecation warning if the PAM file still contains `pam_cockpit_cert`.

src/ws/pam_cockpit_cert.c
```diff
@@ -22,179 +22,15 @@
 
 #include "config.h"
 
-#include <assert.h>
```
martinpitt, line 25: Can you please add a comment to the file that it is a stub for old/modified /etc/pam.d/cockpit files, is deprecated, and will be removed eventually? Just to unconfuse our future selves.

martinpitt (changes requested):
Thanks! I'm happy with doing this in general -- we did this originally to fit the ssh/sudo use cases, but we won't use PAM but kerberos for that.

src/ws/cockpit-session-client-certificate.c
allisonkarlitskaya: `cockpit_session_client_certificate` is namespace.  `check()` is indeed, a bit ... suboptimal.  I'll change it to `map_user()`.

src/ws/session.c
allisonkarlitskaya: This was an oversight.

allisonkarlitskaya:
More reviews welcome, but please don't merge.  I'd like to self-review after I've had a chance to sleep on it — this has high "could accidentally break important stuff" potential.

martinpitt (approved):
Thanks! :+1: I hope my review is meaningful while in draft state, but if not, happy to re-review after your self-review.

martinpitt (changes requested):
Oops, seems the message has changed, tests now fail on unexpected

    cockpit-session: Failed to map certificate to user: [org.freedesktop.sssd.Error.NotFound] User not found

martinpitt:
[this unexpected message](https://logs.cockpit-project.org/logs/pull-15999-20210625-165215-e7fab7d8-fedora-34/log.html#263) is different and also interesting

allisonkarlitskaya:
> [this unexpected message](https://logs.cockpit-project.org/logs/pull-15999-20210625-165215-e7fab7d8-fedora-34/log.html#263) is different and also interesting

That'll do it:

```
         # sssd-dbus not available
         m.execute("systemctl mask sssd-ifp && systemctl stop sssd-ifp")
```

More "now in journal instead of pam log" redux.  Fixed.

martinpitt (dismissed):
Cheers! Tests are still running, but approval-on-green.

allisonkarlitskaya:
> Cheers! Tests are still running, but approval-on-green.

I still haven't done my self-review, so please no merging yet :)

martinpitt (approved):
Re-approving test message update. I adjusted the release note a bit, in most cases a package update ought to do the right thing already. But please adjust back to good English :grin:

PR #16000, opened 2021-06-25
cockpit-tls client certificate cleanup

allisonkarlitskaya:
- [x] #15999


src/tls/connection.c
```diff
@@ -734,7 +734,8 @@ connection_thread_loop (Connection *self)
 }
```
martinpitt, line 734: "we can the presence of the property" in commit message -- this sentence no verb

src/ws/cockpitauth.c
```diff
@@ -1469,14 +1468,20 @@ cockpit_auth_login_async (CockpitAuth *self,
...
    * definitive login type, and don't just silently fall back to other types */
-  if (https_instance_has_certificate_file (NULL, 0) != -1)
+  const gchar *client_certificate;
+  if (metadata && cockpit_json_get_string (metadata, "client-certificate", NULL, &client_certificate))
     {
       g_debug ("TLS connection has peer certificate, using tls-cert auth type");
       type = g_strdup ("tls-cert");
-      /* don't send any actual authorization here; we don't want to put any trust in data sent from cockpit-ws */
-      authorization = g_strdup ("tls-cert");
+      authorization = g_strdup (client_certificate);
```
martinpitt, line 1484: What does that buy us eventually? c-session does not read that at the moment (as the commit message points out, "continued to be ignored"), but the original comment still stands -- cockpit-session can't trust anything from cockpit-ws. Client certs are public, and thus their fingerprints are predictable, so a hacked cockpit-ws could send any value to cockpit-session.

I'd feel much better about connection_create_metadata() just sending `client-certificate: true`, so that it can tell ws to use tls-cert mode. But that should not be passed on to c-session.

allisonkarlitskaya: maybe the name is misleading: this is not the cert, but rather the basename of the file in /run/cockpit/tls which contains the cert.

martinpitt: Of course, but same difference -- as I mentioned, the fingerprint (aka cert path basename) is also predictable. As that information passed to c-session is untrustworthy, we shouldn't pass it so that we can be sure c-session won't ever start to use it.

martinpitt:
Unblocked, the depending PR landed.

src/ws/cockpitauth.c
martinpitt: FTR: it's similar to a password, c-session will still validate it via cgroup membership and content comparison.

martinpitt:
testClientCertAuthentication in [packit centos-8](http://artifacts.dev.testing-farm.io/8777b4a9-4b27-47a9-8598-7b173cbd0a6a/) looks relevant.

martinpitt:
I triggered our own centos-8-stream test here, which will hopefully replicate the regression.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -43,65 +43,50 @@
...
+      if (strncmp (buffer, "0::", 3) == 0)
+        {
+          /* cgroupsv2 (or hybrid) case.  Return the entire line. */
+          result = strdupx (buffer);
+          break;
+        }
+      else if (strncmp (buffer, "1:name=systemd:", 15) == 0)
+        {
+          /* cgroupsv1.  Rewrite to what we'd expect from cgroupsv2. */
+          asprintfx (&result, "0::/%s", buffer + 15);
```
martinpitt, line 79: Ugh.. but okay, this is still easier than accepting both cases later on.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -43,65 +43,50 @@
...
+          break;
+        }
     }
 
-  buf[pmatch[CGROUP_REGEX_MATCH].rm_eo] = '\0';
+  fclose (fp);
 
-  return buf + pmatch[CGROUP_REGEX_MATCH].rm_so;
+  if (out_length)
+    *out_length = strlen (result);
```
martinpitt, line 87: This will crash if `result` is NULL, i.e. if scanning the cgroup file had no match. This Should Not Happen™, but there has been wild stuff out there, like old LXC emulating /proc and /sys for unprivileged containers in the past, or maybe even custom systemd configuration. Better add an `&& result` here.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -333,8 +308,23 @@ cockpit_session_client_certificate_map_user (const char *client_certificate_file
...
     }
 
+  size_t my_cgroup_length;
+  char *my_cgroup = read_proc_self_cgroup (&my_cgroup_length);
+  if (my_cgroup == NULL)
+    {
+      warnx ("Could not determine cgroup of this process");
+      return NULL;
+    }
+  if (strncmp (cert_pem, my_cgroup, my_cgroup_length) != 0)
```
martinpitt, line 318: Thinking aloud: This would also accept sub-cgroups of the expected one. However, (1) the additional suffix would then be part of `sssd_map_certificate()` and fail there (not even decodable as PEM), and (2) that should technically not even matter, as (hopefully) a cgroup child of the target ws instance can't get any more privileges than the parent. So LGTM.

martinpitt (changes requested):
I had a first deeper look. I like the direction!

src/ws/cockpit-session-client-certificate.c
allisonkarlitskaya: We include the newline we read from `/proc/self/cgroup` and include it in the comparison.  That could use some documentation.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -43,65 +43,50 @@
...
+   * Once we no longer need this support, we can drop the loop, switch
+   * to fread(), and just return the entire content of the file.
+   */
+  char buffer[1024];
+  char *result = NULL;
+  while (fgets (buffer, sizeof buffer, fp))
     {
-      warnx ("Not running in a template cgroup, unable to parse systemd unit instance.\n\n/proc/self/cgroups content follows:\n%s\n", buf);
-      return NULL;
+      assert (strchr (buffer, '\n')); /* ...or else, kernel bug. */
```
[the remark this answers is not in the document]
allisonkarlitskaya: It could catch two cases that we should never see:
 1. very long lines (ie: couldn't fit the entire line into the buffer, so we got a partial read)
 2. no newline at end of file

I don't expect to see either of those cases here, but the newline is important to the security properties here: we use it as the terminator between the cgroup header and the certificate in the file.  It's important we make sure to have it.

I think I might make this a more explicit check at the end, just before we return, where we calculate the `strlen()` anyway, that `strlen() - 1` is the `'\n'`.

src/ws/cockpit-session-client-certificate.c
allisonkarlitskaya: added as an assert.  I don't want to deal with this function returning NULL.

martinpitt:
The "internal error in login process" discovered by the tests seems to be [this](https://logs.cockpit-project.org/logs/pull-16000-20210630-072418-96d5050b-fedora-34/TestAD-testClientCertAuthentication-fedora-34-127.0.0.2-2801-FAIL.log.gz):

    cockpit-session: src/ws/cockpit-session-client-certificate.c:90: read_proc_self_cgroup: Assertion `*out_length >= 5' failed.

src/tls/client-certificate.c
```diff
@@ -0,0 +1,419 @@
...
+  snprintf(path, PATH_MAX,  "/proc/self/fd/%d", fd);
+
+  if (linkat (AT_FDCWD, path, dirfd, filename, AT_SYMLINK_FOLLOW) == 0)
+    {
+      *out_filename = filename;
+      return true;
+    }
+  else
+    {
+      warn ("Unable to link client certificate file to /run/cockpit/tls/%s", filename);
```
martinpitt, line 298: This /run/cockpit/tls/ path is not strictly part of the API of this function. Not a biggie, as it's a single-use helper.

src/tls/client-certificate.c
```diff
@@ -0,0 +1,419 @@
...
+ * client_certificate_create_tmpfile:
+ * @dirfd: the directory that will eventually contain the file
+ * @out_fd: a write-mode fd for the created file
+ */
+static bool
+client_certificate_create_tmpfile (int  dirfd,
+                                   int *out_fd)
+{
+  /* create with no permissions: only root will read this */
+  int fd = openat (dirfd, ".", O_TMPFILE | O_WRONLY, 0);
```
martinpitt, line 314: Still, what's wrong with 0400, so that cockpit-tls itself can at least read it? Doesn't 0 cause troubles with unit tests?

src/ws/cockpit-session-client-certificate.c
```diff
@@ -121,7 +121,7 @@ sssd_map_certificate (const char *certificate, char** username)
 }
 
 char *
-cockpit_session_client_certificate_map_user (void)
+cockpit_session_client_certificate_map_user (const char *client_certificate_filename)
```
martinpitt, line 303: Having this API change in this `tls: add a "client-certificate" field to metadata` commit is a little weird, as this functino does not use the new argument in any way. I see that a later commit does, though, so maybe this will just get cleaned up during rebasing.

src/tls/connection.c
```diff
@@ -805,6 +805,9 @@ connection_create_metadata (Connection *self)
   cockpit_json_print_string_property (stream, "origin-ip", ip, -1);
   cockpit_json_print_integer_property (stream, "origin-port", port);
 
+  if (self->certfile_fd != -1)
+    cockpit_json_print_string_property (stream, "client-certificate", self->fingerprint.str, -1);
```
martinpitt: Maybe add an assert that fingerprint is set? We expect certfile_fd and fingerprint to be set together, so Should Not Happen™ (but that's what assertions are for)

src/ws/session.c
```diff
@@ -522,15 +522,22 @@ pam_conv_func_dummy (int num_msg,
...
 {
   struct pam_conv conv = { pam_conv_func_dummy, };
+  const char *client_certificate;
   pam_handle_t *pamh;
   int res;
 
   debug ("start tls-cert authentication for cockpit-ws %u", getppid ());
 
-  char *username = cockpit_session_client_certificate_map_user ();
+  if (strncmp (authorization, "tls-cert ", 9) == 0)
```
martinpitt, line 535: This is already done by `cockpit_authorize_type()`, so this could even be turned into an assertion, and a hard crash.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -138,69 +153,79 @@ https_instance_has_certificate_file (char   *contents,
   if (https_instance == NULL) /* already warned */
     goto out;
 
+  /* No tricky stuff, please */
+  if (!valid_256_bit_hex_string (filename))
+    {
+      warnx ("tls-cert authentication token is invalid");
+      goto out;
+    }
```
martinpitt, line 147: I wonder if that extra effort with the helper function is really justified. The check immediately below already validates that it has the only possible value -- so any "funny" stuff would be caught there anyway.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -138,69 +153,79 @@ https_instance_has_certificate_file (char   *contents,
...
     }
 
   if (buf.st_size == 0)
     {
-      warnx ("Could not read certificate: /run/cockpit/tls/%s is empty", https_instance);
+      warnx ("Could not read certificate: /run/cockpit/tls/%s is empty", filename);
       goto out;
     }
 
-  if (contents != NULL)
```
martinpitt: Add an assertion instead?

martinpitt (changes requested):
I re-reviewed this up to "session: use the provided client certificate filename" (timeout, sorry). A few really minor thoughts/issues, but so far this looks really good! I also wouldn't mind if you sent the first few commits as a separate PR, so that we can land this piece-wise.

src/tls/test-server.c
```diff
@@ -120,6 +120,42 @@ block_sigchld (void)
...
+  while ((name = g_dir_read_name (dir)))
+    {
+      g_autofree char *filename = g_build_filename (tc->runtime_dir, name, NULL);
+
+      g_autofree char *contents = NULL;
+      g_autoptr(GError) error = NULL;
+      if (!g_file_get_contents (filename, &contents, NULL, &error))
+        {
+          /* files are flying around all the time: this might reasonably fail */
+          if (g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
```
martinpitt, line 141: Do they? I don't understand how this can be anything but predictable in the unit tests -- each test gets its own temporary directory, and there is no paralellism. The only case that I can imagine is that this gets called if there is no file at all, but then this function already cleanly returns `false`.

martinpitt (changes requested):
Finished the review. I found a few more small issues since the last round, but this by and large looks good. Nice work!

The biggest outstanding issues are of course that this doe not actually work yet (unit tests and integration tests fail).

src/tls/client-certificate.c
allisonkarlitskaya: I changed this to `0400`.  It turned out to be necessary for testing.

src/ws/cockpit-session-client-certificate.c
allisonkarlitskaya: But this changes in future commits...

src/ws/cockpit-session-client-certificate.c
allisonkarlitskaya: That also changes in future commits.  I don't want to get too caught up on the in-between stuff.

src/tls/client-certificate.c
```diff
@@ -0,0 +1,418 @@
...
+ * program.
+ *
+ * The return value of this function needs to be free()d.
+ */
+static char *
+client_certificate_get_wsinstance (const gnutls_datum_t *certificate)
+{
+  unsigned char digest_data[256 / 8];
+  size_t digest_size = sizeof digest_data;
+  int r = gnutls_fingerprint (GNUTLS_DIG_SHA256, certificate, digest_data, &digest_size);
```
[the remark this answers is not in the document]
allisonkarlitskaya: I strongly agree there.  We're going to land this before making further changes.

src/tls/test-server.c
allisonkarlitskaya: We loop, waiting for the file to disappear during teardown, for example.  In that case we'd possibly see the directory entry but be unable to open the file, if the deletion happens in between.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -43,65 +43,55 @@
...
+  if (fp == NULL)
+    return NULL;
+
+  /* Support cgroups v1 by looping.
+   * Once we no longer need this support, we can drop the loop, switch
+   * to fread(), and just return the entire content of the file.
+   */
+  char buffer[1024];
+  char *result = NULL;
+  while (fgets (buffer, sizeof buffer, fp))
```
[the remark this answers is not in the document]
allisonkarlitskaya: Will add a comment there.

src/tls/client-certificate.c
```diff
@@ -0,0 +1,418 @@
...
+    {
+      warnx ("Verifying TLS peer failed: %s", gnutls_strerror (ret));
+      return ret;
+    }
+
+  return GNUTLS_E_SUCCESS;
+}
+
+/**
+ * to_hex:
```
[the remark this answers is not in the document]
allisonkarlitskaya: OK.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -43,65 +43,55 @@
...
-  static char buf[1024];
-  regmatch_t pmatch[CGROUP_REGEX_GROUPS];
-  regex_t preg;
-  int fd;
-
-  /* read /proc/self/cgroup */
-  fd = open ("/proc/self/cgroup", O_RDONLY | O_CLOEXEC | O_NOFOLLOW);
-  if (fd < 0)
-    {
-      warn ("Failed to open /proc/self/cgroup");
```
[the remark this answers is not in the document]
allisonkarlitskaya: OK.

src/ws/session.c
```diff
@@ -544,10 +544,9 @@ perform_tlscert (const char *rhost,
 
   debug ("start tls-cert authentication for cockpit-ws %u", getppid ());
 
-  if (strncmp (authorization, "tls-cert ", 9) == 0)
-    client_certificate = authorization + 9;
-  else
-    exit_init_problem (PAM_AUTH_ERR);
+  /* True, otherwise we wouldn't be here. */
+  assert (strncmp (authorization, "tls-cert ", 9));
```
martinpitt, line 548: Before you wonder about all sorts of failures: Missing `== 0`

src/ws/session.c
```diff
@@ -544,10 +544,9 @@ perform_tlscert (const char *rhost,
...
+  client_certificate = authorization + 9;
```
martinpitt, line 549: This is the file name on disk which contains the cert, not the cert itself, right? Might be worth renaming this to avoid confusion.

martinpitt (changes requested):

src/ws/session.c
allisonkarlitskaya: Thanks lol.

src/ws/session.c
allisonkarlitskaya: It's also called "client-certificate" in the metadata blob...

allisonkarlitskaya: I changed the name.

allisonkarlitskaya:
The unit tests are failing with a whole lot of `dbus-daemon --print-address --session` processes running.  That seems related to #16157 but why wouldn't it have been failing before now?

martinpitt:
Quick summary  from local unit tests run (in container), as a reminder for myself next week: I see a lot of leaked `dbus-daemon --print-address --session` processes and a zombie `sh` process. Maybe this is interacting badly with the [GSubprocess conversion](https://github.com/cockpit-project/cockpit/commit/c0d52ddac0eefbe68fde27a229ba544910ca7939)? Might make sense to back that out (my PR even had a version where the fdwalk() porting happened *before* GSubprocess) and see if it changes things. Likewise, it might be worth to replace cockpit_close_range() with a `return 0` stub -- this only breaks one unit test, which should be easy enough to disable.

martinpitt:
I reproruced that in a tree with jumpstart and `NOCONFIGURE=1 ./autogen.sh` and 

    podman run -it --rm -v $PWD:/source -u root ghcr.io/cockpit-project/unit-tests bash

so that iterating is much faster. Inside I created /tmp/b with `/source/configure && make`.

This still reproduces the [original hang in CI](https://github.com/cockpit-project/cockpit/pull/16000/checks?check_run_id=3203462391#step:4:3130):
```
# ./test-tls-server 
/server/mixed-protocols: OK
/server/run-idle: OK
/server/no-tls/single-request: OK
/server/no-tls/many-serial: OK
/server/no-tls/many-parallel: OK
/server/no-tls/redirect: OK
/server/tls/no-client-cert: OK
/server/tls/client-cert: test-tls-server: Couldn't create temporary file for client certificate: Operation not supported
test-tls-server: Disconnecting client due to above failure.
**
ERROR:/source/src/tls/test-server.c:295:assert_https_outcome: assertion failed (len >= 100): (-54 >= 100)
**
ERROR:/source/src/tls/test-server.c:339:assert_https_outcome: assertion failed (status == 0): (134 == 0)
```

So it is [this openat() call](https://github.com/cockpit-project/cockpit/pull/16000/commits/063c55e39f622750e00b63a38966e2056325a4b1#diff-f331442b6ca76d073b9211f551f6518879d06200cbb09139ee7fe5819cd2665eR279) in client_certificate_create_tmpfile() which fails in the container. Unfortunately strace does not work in the container. Note that this happens without valgrind.

martinpitt:
@allisonkarlitskaya : I added a FIXUP which avoids `O_TMPFILE`, and added a separate commit which fixes a little regression from commit 11059a88195c268. I also rebased to main, for that I force-pushed. But I did *not* change any of your commits.

allisonkarlitskaya:
> @allisonkarlitskaya : I added a FIXUP which avoids `O_TMPFILE`, and added a separate commit which fixes a little regression from commit [11059a8](https://github.com/cockpit-project/cockpit/commit/11059a88195c268bd549501148b8f4e367144020). I also rebased to main, for that I force-pushed. But I did _not_ change any of your commits.

I don't like this.  Normally (not testing), we'll only be writing this file to `/run` and that ought to always be tmpfs.  We rather just need to figure out a way to fix our testing container so that this isn't a problem.

martinpitt:
@allisonkarlitskaya : OK, good point; we'll need to revisit this once we actually make cockpit-tls work in the cockpit/ws container, as we don't have a /run there by default. But until then we can uses some working variant of commit 075cf9ead then (not sure why this does this [weird breakage](https://github.com/cockpit-project/cockpit/pull/16000/checks?check_run_id=3212862000#step:4:36), I didn't investigate this too much today -- it *does* work locally.

allisonkarlitskaya:
> @allisonkarlitskaya : OK, good point; we'll need to revisit this once we actually make cockpit-tls work in the cockpit/ws container, as we don't have a /run there by default. But until then we can uses some working variant of commit [075cf9e](https://github.com/cockpit-project/cockpit/commit/075cf9ead17e60d9607c8625e7048751261e15c4) then (not sure why this does this [weird breakage](https://github.com/cockpit-project/cockpit/pull/16000/checks?check_run_id=3212862000#step:4:36), I didn't investigate this too much today -- it _does_ work locally.

I think even with `cockpit-tls` in the container this won't be an issue: we can't do cert auth in that case, and we now only write cert files in the case that cert auth is enabled (and the client actually sent one).

allisonkarlitskaya:
One more idea: maybe we could modify the test to write the temporary runtime directory as a subdir of the XDG runtime directory instead of in `/tmp`.  We must surely have a reliable tmpfs *somewhere* in our testing container...

martinpitt (changes requested):
Nice! You (accidentally, I suppose) dropped my "tls: Fix build failure on nonexisting fingerprint with DEBUG" commit, and there's two trivial comment issues. I'd like to block landing this until packit comes back to make sure that /dev/shm/ actually works in koji builders. This hopefully happens today, otherwise I'll test it manually in COPR land it anyway.

I'll push the trivial fix-ups now.

src/tls/client-certificate.c
```diff
@@ -0,0 +1,384 @@
...
+
+  return GNUTLS_E_SUCCESS;
+}
+
+/**
+ * client_certificate_get_wsinstance:
+ * @certificate: the certificate presented by the peer
+ *
+ * Determines the correct cockpit-ws instance for handling connections
+ * for this @peer_certificate.
```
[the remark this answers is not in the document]
martinpitt: Fixed.

src/tls/client-certificate.c
```diff
@@ -0,0 +1,384 @@
...
+    warnx ("Partial write of %s to certificate file: %zu of %zu", description, s, size);
+
+  return false;
+}
+
+static bool
+client_certificate_write_cgroup_header (int         fd,
+                                        const char *wsinstance)
+{
+  /* First write the expected cgroup of the wsinstance */
```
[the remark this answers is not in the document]
martinpitt: Dropped it -- client_certificate_accept() already documents the structure of the file.

martinpitt (approved):

martinpitt:
FTR, packit was fixed and the [COPR builds](https://copr.fedorainfracloud.org/coprs/packit/cockpit-project-cockpit-16000/build/2356689/) are happy. https://github.com/cockpit-project/bots/pull/2267 also landed, so this should hopefully have been the final retry round.

PR #16001, opened 2021-06-25
tests: add some new pixel tests in network page

KKoukiou:


test/verify/check-networkmanager-basic
```diff
@@ -41,6 +41,17 @@ class TestNetworkingBasic(NetworkCase):
...
 
+        b.assert_pixels(
+            "#networking", "network-main",
+            ignore=[
+                "#networking-graphs .pf-l-grid",
+                "tr[data-interface=cockpit1] td[data-label=Sending]", "tr[data-interface=cockpit1] td[data-label=Receiving]",
+                "tr[data-interface=eth0] td[data-label=Sending]", "tr[data-interface=eth0] td[data-label=Receiving]",
+                "tr[data-interface=v_cockpit1] td[data-label=Sending]", "tr[data-interface=v_cockpit1] td[data-label=Receiving]",
+                ".cockpit-log-panel"
+            ]
```
martinpitt, line 52: My first thought was "this is capturing and ignoring too much", but it actually makes sense -- we are testing the page layout (like spacing between cards) as well, which can't hurt. So let's try this, and we can dial it back if it gets in the way somehow. I only hope that we don't get bitten by different scrolling position.

In general, quite a number of these screenshots contain scroll bars -- so whenever firewalld gets a new profile or drops one, or the "Network logs" contain more or fewer messages, the scroll bar slider size will change. It may be better to ignore the scrollbar in the images as well?

martinpitt (approved):
I think this is too strict, but I'm not opposed to starting with that. For sure it's good to have more of these, thanks!

martinpitt (changes requested):
Nevermind, there [are some pixel failures](https://logs.cockpit-project.org/logs/pull-16001-20210625-161217-7eed4f02-fedora-34/log.html#285-2).

KKoukiou:
> Nevermind, there [are some pixel failures](https://logs.cockpit-project.org/logs/pull-16001-20210625-161217-7eed4f02-fedora-34/log.html#285-2).

Forgot to push one image indeed.

test/verify/check-networkmanager-basic
KKoukiou: Indeed, we should be carefull with content that might change with OS updates etc. Which the firewall images indeed are capturing.

martinpitt:
[test failed](https://logs.cockpit-project.org/logs/pull-16001-20210625-181908-4f8d4c6d-fedora-34/log.html#271-2) on "1 zone" vs. "2 zones". This is a race condition as the libvirt zone gets initialized asynchronously. But for the purposes of layout testing this is not important, so I suggest to just ignore that selector as well.

I also think we should ignore the IP addresses at least, and for sure we need to ignore the date in the logs view. Perhaps ignore the entire contents of the logs card, and this only tests its presence, size, and placement?

KKoukiou:
> [test failed](https://logs.cockpit-project.org/logs/pull-16001-20210625-181908-4f8d4c6d-fedora-34/log.html#271-2) on "1 zone" vs. "2 zones". This is a race condition as the libvirt zone gets initialized asynchronously. But for the purposes of layout testing this is not important, so I suggest to just ignore that selector as well.
> 
> I also think we should ignore the IP addresses at least, and for sure we need to ignore the date in the logs view. Perhaps ignore the entire contents of the logs card, and this only tests its presence, size, and placement?

I am actually already ignoring the button in the firewall card, I am quite suprised the ignore did not take effect since the selector is correct. The content of logs card is ignore completely as well. 
The ip addresses I did not ignore yet, good point! :)

test/verify/check-networkmanager-basic
```diff
@@ -41,6 +41,22 @@ class TestNetworkingBasic(NetworkCase):
         b.wait_in_text("#networking-graphs", "Transmitting")
         b.wait_in_text("#networking-graphs", "Receiving")
 
+        ifaces = list(filter(None, m.execute("ip r | grep -o -P '(?<=dev\ ).*(?=\ proto)'").split('\n')))
```
martinpitt, line 44: What does the `filter()` do?

Listing /sys/class/net may be much easier?
```py
# python3 -c 'import os; print(os.listdir("/sys/class/net"))'
['eth1', 'virbr0-nic', 'lo', 'eth0', 'virbr0']
```


And you may want to filter out `lo` (with either implementation). 

If you don't like /sys/class/net: `ip -brief link show | awk '{print $1}'`

KKoukiou: > What does the `filter()` do?

There was an empty string in the list, just removes it. 
> 
> Listing /sys/class/net may be much easier?
> 
> ```python
> # python3 -c 'import os; print(os.listdir("/sys/class/net"))'
> ['eth1', 'virbr0-nic', 'lo', 'eth0', 'virbr0']
> ```
> 
> And you may want to filter out `lo` (with either implementation).
> 
> If you don't like /sys/class/net: `ip -brief link show | awk '{print $1}'`

Thanks for making this nice :)

martinpitt (approved):
Danke! I trigger another test, just to make sure it works there also.

Issue #15979, opened 2021-06-22
remove Moment.js


garrett:
Moment.js [has been deprecated](https://momentjs.com/docs/#/-project-status/). We should switch to something else. See Moment's suggested list @ https://momentjs.com/docs/#/-project-status/recommendations/

Ideally, we could just use native JavaScript date & time functions. They're much better now (and supported in every browser) than when moment.js was first made. If not native, we should use something that's a small wrapper and has parts we can pick and choose (date-fns for example).

Absolute times (convert to `Intl.DateFormat`):
 - [x] Accounts page: #16002 
 - [x] Metrics page: #16015
 - [x] Common library, Overview/Logs pages: #16019
 - [x] Hardware Info: #16029
 - [x] Shutdown dialog (non-trivial, as current date picker is broken), blocked on #16053
 - [x] Plots, Services: PR #16028 
 - [x] SELinux: PR #16036

Time durations (not covered by `Intl.DateFormat`):
- [ ] Storage: Blocked on #15937 and #15968: #16097
- [x] Uptime on Overview: #16084
- [x] Updates: #16090

martinpitt:
Indeed the `Intl` class looks quite sufficient for absolute date/time display:

```js
new Intl.DateTimeFormat('de', { dateStyle: "short", timeStyle: "short"}).format(d)
"26.06.21, 10:45"
new Intl.DateTimeFormat('en', { dateStyle: "short", timeStyle: "short"}).format(d)
"6/26/21, 10:45 AM"
new Intl.DateTimeFormat('zh-CN', { dateStyle: "short", timeStyle: "short"}).format(d)
"2021/6/26 上午10:45"
```

So we can at least start with replacing these. Maybe we need to have a little wrapper in pkg/lib/ which will use `cockpit.language` and provide a collection of default options so that we retain some consistency. We currently call `moment.format(...)` with a wildly differing set of options, maybe we can clean this up a little bit. Although on first sight this may actually be justified..

We use durations in 4 places in our code (packagekit, 2x storaged, overview), `Intl` has no counterpart for this. But that would at least reduce moment usage quite a bit, and we can look for a small replacement lib for these cases.

martinpitt:
I ported the accounts page in #16002, comments appreciated.

PR #16002, opened 2021-06-26
users: Replace moment with Intl.DateTimeFormat

martinpitt:
moment is deprecated, and the users page only uses absolute dates/times.

See #15979

-----

This is the first page for trying this, it's easy to convert. Let's discuss general style/coding issues here, then I'm happy to port some more pages.

The difference is negligible, with Intl it even looks a tiny bit better in "last login", and it's identical for "Access":

Master, English:
user-master-en

This PR, English:
image

Master, German:
user-master-de

This PR, German:
image

### DIFF
--- a/pkg/users/account-details.js
+++ b/pkg/users/account-details.js
@@ -19,7 +19,6 @@
 
 import cockpit from 'cockpit';
 import React, { useState, useEffect } from 'react';
-import moment from "moment";
 import { superuser } from "superuser";
 import { apply_modal_dialog } from "cockpit-components-dialog.jsx";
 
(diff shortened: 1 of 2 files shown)


martinpitt:
f34 tests were happy, so let's run them all.

martinpitt:
Oof, all tests fail now. My branch is up to date, so node_modules is not actually outdated. Supposedly the master branch renaming still needs some adjustments? That's something for tomorrow then.

allisonkarlitskaya (changes requested):
The failure is caused by `node_modules` being updated by this commit.

martinpitt:
Thanks. No idea how that went in, repushed with cleaning this up.

martinpitt:
This now mysteriously [fails](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16002-20210627-122048-006c2362-debian-stable/log.html#275-2) `test/verify/check-testlib TestRunTest.testRetry`. It didn't before, so somehow we broke something unrelated. My hunch is the `env["BASE_BRANCH"] = "master"` which needs to be adjusted for the rename to "main".

martinpitt:
Nevermind, Allison already did it in #16006.

KKoukiou (approved):

KKoukiou (approved):
I am fine with the new module and also how you use it here.

PR #16003, opened 2021-06-27
workflows: change 'master' to 'main' in a workflow

allisonkarlitskaya:
This seems to be the only thing which needs to be changed immediately.
Everything else is URLs in documentation, and GitHub will automatically
forward those.

### DIFF
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -2,7 +2,7 @@ name: build-dist
 on:
   pull_request:
   push:
-    branches: [master]
+    branches: [main]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks! Triggering required f34 test. In #16002 everything just went up in flames, and it's not that bit, but let's figure that out separately.

martinpitt:
@allisonkarlitskaya : We also immediately need to rename/change the -weblate branch, and possibly adjust something in weblate UI (I've never done this so far).

martinpitt:
Also, testmap in bots

allisonkarlitskaya:
I just renamed -weblate to `main` but I couldn't figure out how to update it in the weblate UI.  Also: we have the branch set to `master` in the Makefile, so maybe that's OK for now?

Gundersanne (approved):

PR #16004, opened 2021-06-27
build: add .sha256sums file to jumpstart tarball

allisonkarlitskaya:


.github/workflows/publish-dist.yml
```diff
@@ -31,9 +31,10 @@ jobs:
           set -ex
           mkdir dist-repo
           cd dist-repo
-          tar -xf ../artifact/webpack-jumpstart.tar dist package-lock.json
+          tar -xf ../artifact/webpack-jumpstart.tar dist package-lock.json .sha256sums || true
```
martinpitt, line 34: Do you already have the follow-up for this in the pipeline? I don't want that alternative truth in production for any extended time (> 1 day).

pkg/build
```diff
@@ -29,15 +29,20 @@ MANIFESTS = \
...
 V_TAR = $(V_TAR_$(V))
 V_TAR_ = $(V_TAR_$(AM_DEFAULT_VERBOSITY))
 V_TAR_0 = @echo "  TAR     " $@;
 
+ls-webpack-inputs:
+	ls -1 package-lock.json $(WEBPACK_INPUTS) | uniq
+
 webpack-jumpstart.tar: $(MANIFESTS) package-lock.json pkg/build
-	$(V_TAR) tar cf webpack-jumpstart.tar dist package-lock.json
+	@pkg/build -s ls-webpack-inputs | xargs sha256sum > .sha256sums
```
martinpitt, line 43: What's the advantage of doing this complicated way (sub-make and new internal phony target) as opposed to just putting the `ls` here? Also, Aren't `$WEBPACK_INPUTS` already single files? Why do they need to go through `ls` again, as opposed to `sha256sum package-lock.json $(WEBPACK_INPUTS)`?

martinpitt (commented):
What's wrong with keeping package-lock.json in the dist cache? It seems very useful both for quickly inspecting/diffing two trees, as well as being able to reconstruct `node_modules` with plain `npm install`. I wouldn't want to completely break that workflow for developers.

.github/workflows/publish-dist.yml
allisonkarlitskaya: #16005

pkg/build
allisonkarlitskaya: This part is annoying and complicated, for two reasons:
 - in the first `make` run, `WEBPACK_INPUTS` is empty, because the `Makefile.deps` weren't created yet.  The second `make` run is to give it a chance to read those files in after webpack runs
 - I use `ls` instead of `echo` because there is quite a lot of duplication in `$(WEBPACK_INPUTS)`, and I don't know a better way than `uniq` (which requires separate, sorted, lines — as `ls` produces).  Probably I could drop the `xargs` use here, though, in favour of a `$()`.  Ideas welcome!

.github/workflows/publish-dist.yml
allisonkarlitskaya: For the record, I think this "alternative truth" is safe.

martinpitt (changes requested):
I just reviewed #16005. Let's first discuss the structure that we want eventually, before landing this half of it. So blocking for now.

pkg/build
martinpitt: Amen to "annoying and complicated" :grin: -- see #16005, is there really no simpler way to do this?

PR #16005, opened 2021-06-27
jumpstart: use sha256sums file for consistency checking

### LINKED ISSUE #15960: dist/ caching does the wrong thing in non-rebased PRs
Any time someone proposes a non-rebased PR, GitHub generates a merge commit vs. the master branch and starts running the `build-dist` workflow out of that.  We end up a `dist/` corresponding to the PR head branch, as if it had been rebased onto the current target branch.  That's what we want: testing changes as they would look if they landed on the master branch now.

The problem starts, however, when the `publish-dist` workflow starts running and names the result of that build according to the sha of the PR head commit.  We do that for two main reasons:
 - this is the easier thing to check from the `workflow_run` event that triggers that workflow
 - we want to get the cache result for non-rebased checkouts of the PR head branch (for example, for testing locally, or also for packit builds)

The problem is made complete by running `tools/webpack-jumpstart` in that PR branch.  You are working with a non-rebased `src/`, `pkg/`, `node_modules`, `package-lock.json`, and everything else, but your `dist/` was built from a rebased version of the tree.  This is most readily apparent when there have been changes to the `node_modules/` on the master branch, because now your `package-lock.json` will be out of sync.

The current situation also prevents the dist cache from being useful from our integration test runs (which rebase the branch, ending up with a different commit ID) and will also be a problem in the future when packit also does a merge (and surely ends up with another distinct result).

This problem has been making my skin crawl for quite some time, because we basically have a cache that introduces corrupt results into the tree: the `dist/` that you get might very well be different than the `dist/` you would get if you ran the build yourself locally.

I'm also afraid that there are not any good solutions here.

 - we could disable `build-dist` for PRs and use it only on `push`.  This is the case where it's the most useful, anyway.
 - we could query the PR from the workflow_run (this is harder than it sounds) and tag using the merge commit ID, and then figure out a way to get the merge commit onto our own machines, packit, etc. in order to make sure everything associated with a PR is testing from the exact same set of files
 - we could abandon naming by SHA for the PR merge case and, for example, name the branch according to the PR number or something like that
 - we could try to figure out a way to tie the cache to the PR itself (as it was when we had artifacts), but it needs to be done in a way that packit can access it
 - ???

allisonkarlitskaya:
- [ ] #16004 

This is going to fail until the new `publish-dist` workflow lands in #16004 

Merging this is also going to break packit builds for any branch that's not completely up-to-date with main at the time of merging.

This technically fixes #15960, but the "strictly correct" behaviour here will regress things a bit in the short term.  We might want to wait on packit/ogr#584 in order to not make a complete mess.

### DIFF
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -31,10 +31,9 @@ jobs:
           set -ex
           mkdir dist-repo
           cd dist-repo
-          tar -xf ../artifact/webpack-jumpstart.tar dist package-lock.json .sha256sums || true
+          tar -xf ../artifact/webpack-jumpstart.tar dist .sha256sums
           git init
           git add dist .sha256sums
-          [ ! -e package-lock.json ] || git add package-lock.json
           git commit -m "Build for ${{ github.event.workflow_run.head_sha }}"
           tag='sha-${{ github.event.workflow_run.head_sha }}'
           git tag "$tag"
--- a/.github/workflows/publish-dist.yml
+++ b/.github/workflows/publish-dist.yml
@@ -31,9 +31,10 @@ jobs:
           set -ex
           mkdir dist-repo
           cd dist-repo
-          tar -xf ../artifact/webpack-jumpstart.tar dist package-lock.json
+          tar -xf ../artifact/webpack-jumpstart.tar dist package-lock.json .sha256sums || true
           git init
-          git add dist package-lock.json
+          git add dist .sha256sums
+          [ ! -e package-lock.json ] || git add package-lock.json
           git commit -m "Build for ${{ github.event.workflow_run.head_sha }}"
           tag='sha-${{ github.event.workflow_run.head_sha }}'
           git tag "$tag"
--- a/tools/git-utils.sh
+++ b/tools/git-utils.sh
@@ -83,10 +83,22 @@ fetch_to_cache() {
     fi
 }
 
+# Given cache commit "$1", print content of file "$2"
+cat_from_cache() {
+    git_cache cat-file blob "$1:$2"
+}
+
 # Consistency checking: for a given cache commit "$1", check if it contains a
 # file "$2" which is equal to the file "$3" present in the working tree.
 cmp_from_cache() {
-    git_cache cat-file blob "$1:$2" | cmp "$3"
+    cat_from_cache "$1" "$2" | cmp "$3"
+}
+
+# Consistency checking: for a given cache commit "$1", check if it contains a
+# file ".sha256sums" which is a list of sha256 checksums which must match the
+# working tree.
+sha256sum_from_cache() {
+    cat_from_cache "$1" ".sha256sums" | sha256sum --check --quiet --strict
 }
 
 # Like `git clone` except that it uses the original origin url and supports
--- a/.github/workflows/build-dist.yml
+++ b/.github/workflows/build-dist.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Build jumpstart tarball
         run: pkg/build -j$(nproc) webpack-jumpstart.tar
 
-      - name: Create dist artifact
+      - name: Create artifact
         uses: actions/upload-artifact@v2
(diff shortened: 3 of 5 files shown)


.github/workflows/publish-dist.yml
```diff
@@ -31,10 +31,9 @@ jobs:
           set -ex
           mkdir dist-repo
           cd dist-repo
-          tar -xf ../artifact/webpack-jumpstart.tar dist package-lock.json .sha256sums || true
+          tar -xf ../artifact/webpack-jumpstart.tar dist .sha256sums
```
martinpitt: We may not use package-lock.json in workflows any more, but it's still an useful file to look at/diff as a developer, and ensures that "regular" web developers not familiar with our cache magic can work with the tree. So I'd rather leave this out.

tools/git-utils.sh
```diff
@@ -83,10 +83,22 @@ fetch_to_cache() {
...
 cmp_from_cache() {
-    git_cache cat-file blob "$1:$2" | cmp "$3"
+    cat_from_cache "$1" "$2" | cmp "$3"
+}
+
+# Consistency checking: for a given cache commit "$1", check if it contains a
+# file ".sha256sums" which is a list of sha256 checksums which must match the
+# working tree.
+sha256sum_from_cache() {
+    cat_from_cache "$1" ".sha256sums" | sha256sum --check --quiet --strict
```
martinpitt, line 101: This is quite expensive, isn't it?

allisonkarlitskaya:
The part not mentioned here is what this will facilitate in the future: I have an idea to move from using the exact commit ID number, and rather doing `git ls-remote` on the cache, and selecting the first commit from the history of the current branch for which we have a cache object.  The idea is then to download the cache object, and use it if the checksums match.

That would solve this usecase of "I have a bunch of C-only commits and still want the cache" in a very elegant way.

allisonkarlitskaya:
It also means we don't need any "dirty working tree" type of checks.  Either the working tree changes would have invalidated the cache, or they wouldn't.

martinpitt (changes requested):
Admittedly I'm not a big fan of this. It's rather convoluted and expensive, and it's IMHO checking at the wrong place. We currently use the current work tree SHA to what we get from the cache, which should already precisely identify what we want -- that plus the check for modified files should already be unambiguous on the local dev side.

I thought your objection is that the build-dist tagged cache entries themselves are a bit of a lie, as they don't *really* get built from the advertised SHA, but from a rebase against current master. So if we want to become that strict, then we should either drop the rebasing (but that would greatly reduce the faith in test results) or include the current target branch head (on top of which the rebase happened) in the dist cache tags. This should achieve the same effect without all that local computation, be much more obvious, and in principle we could then also rebuild the same PR head commit against multiple versions of a rebase.

martinpitt:
I see -- so you want to do this as essentially a way to extend the "these local changes are harmless" checks to already commited local changes. This would then kind of fix both that case, as well as webpack-related changes introduced in rebases. That at least makes sense as a use case.

Honestly my main concern is that the whole dist/node_modules handling has already become really complex (to the point that not even most of our team understands it any more), and this is making it even worse. This is of course because it's a fundamentally complex problem, not because it's implemented badly, it's just something that bothers me a lot.

tools/git-utils.sh
allisonkarlitskaya:
```
⬢[lis@f34 jumpstart-checksum]$ time sha256sum --check --quiet --strict < .sha256sums 

real	0m0.019s
user	0m0.010s
sys	0m0.009s
```

no.

.github/workflows/publish-dist.yml
allisonkarlitskaya: We discussed that this was a misunderstanding: `package-lock.json` remains in the `node_modules` directory, but this is the `dist/` cache we discuss here.

martinpitt: Indeed, sorry for the thinko!

PR #16006, opened 2021-06-27
various: replace 'master' with 'main' in urls

allisonkarlitskaya:
We have a lot of links to our GitHub repository on the web littered
around in our source tree.  Change them all from master to main.


martinpitt:
Tests are going to [fail](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16002-20210627-122048-006c2362-debian-stable/log.html#275-2) on another "master" value in tests. Want to add this right here, so that we can land this green?

martinpitt (approved):
Some tests are already green, and I have absolutely no idea why it would work with check-testlib's "master" here, but not in #16002. Anyway, *if* it works, feel free to land, and we can fix check-testlib separately. If not, I'm happy to reapprove.

martinpitt:
Oh, it was already part of a different PR, nevermind.

PR #16008, opened 2021-06-28
login: Move forEach patch & noscript inline

### LINKED ISSUE #15992: login: JavaScript error message appears on high latency and/or low bandwidth connections
When there's a lot of latency and/or low bandwidth, the login page loads and renders before the JavaScript is downloaded. This makes it look like JavaScript is not enabled.

Solution would be to inline the JavaScript needed to hide the message and swap out the message to indicate the page is loading.

garrett:
Fixes #15992

The `forEach` patch is at the top so the fix for IE (which we still need to "support", to tell IE users we don't support them). Hiding the `noscript` elements is at the bottom so it runs immediately after the browser constructs the DOM. (It's like `document`'s `ready` event, without event binding.)

### DIFF
--- a/pkg/static/login.html
+++ b/pkg/static/login.html
@@ -7,6 +7,11 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta name="robots" content="noindex">
   <meta insert_dynamic_content_here>
+  <script type="text/javascript">
+    /* Patch IE to support forEach on NodeLists, used in show/hide */
+    if (window.NodeList && !NodeList.prototype.forEach)
+        NodeList.prototype.forEach = Array.prototype.forEach;
+  </script>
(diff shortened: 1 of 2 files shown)


garrett:
I shift-reloaded the pages with 2G throttle on.

### Before

https://user-images.githubusercontent.com/10246/123609044-9b799180-d7ff-11eb-90b9-3eb1938301e1.mp4

### After


https://user-images.githubusercontent.com/10246/123609159-b64c0600-d7ff-11eb-80d1-153d904a8311.mp4

garrett:
Note: Throttling is found in Firefox's mobile view (control+shift+m). Chrome has something similar too.

martinpitt (approved):
Thank you! This is spot-on!

PR #16009, opened 2021-06-28
metrics: Install redis package on demand

martinpitt:
Always show the "Export to network" switch. Install the redis package if
redis.service does not exist.

Change the pmlogger existence check from `.state === undefined` to
`.exists`, as that is easier to understand, and we were already using
the latter for pmproxy.

Rename TestCockpitPcp to TestMetricsPackages, as this covers more
packages than just cockpit-pcp now.

-----

I interactively tested that with

      systemctl stop pmlogger pmproxy redis; rpm -e --verbose cockpit-pcp pcp redis; systemctl daemon-reload

and then enabling pmlogger and pmproxy either one by one, or both together. After pmlogger and relogin (as advertised), history appears, and after pmproxy, querying metrics with

    curl 'http://localhost:44322/series/labels?names=hostname'

works. Our integration tests can only test package install and actual service configuration separately.

 - [x] Builds on top of #15998
 - [x] Builds on top of #16018

There isn't any thing visually/CSS-y new here, but screenshots for reference.

## Metrics: Install packages on demand

[Cockpit 247](https://cockpit-project.org/blog/cockpit-247.html) introduced setting up a machine for Grafana metrics. However, this only worked if the `pcp` and `redis` packages were already installed. With this Cockpit version, the options will always work, and missing packages get installed on demand.

image

image

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -915,18 +915,27 @@ const PCPConfig = ({ buttonVariant, firewalldRequest, needsLogout, setNeedsLogou
         return null;
 
     const handleSave = () => {
-        // when enabling pmlogger, install cockpit-pcp/pcp on demand
-        if (!deferSaveOnInstall && dialogLoggerValue && s_pmlogger.state === undefined) {
-            debug("PCPConfig: request to enable pmlogger, but missing pmlogger.service, offering install");
-            setDialogVisible(false);
-            install_dialog("cockpit-pcp")
-                    .then(() => {
-                        debug("PCPConfig: cockpit-pcp installation successful");
-                        setNeedsLogout(true);
-                        setDeferSaveOnInstall(true); // avoid recursive install_dialog
-                    })
-                    .catch(() => null); // ignore cancel
-            return;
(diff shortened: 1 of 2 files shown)


martinpitt:
Still chasing [this test flake](https://logs.cockpit-project.org/logs/pull-16009-20210628-143757-68e7709e-debian-testing/log.html#282-1) on debian-testing/ubuntu-stable. It seems that right after start and logging in, `real_redis.exists` is not true for a brief period of time. I tried hard to reproduce that locally, and I also don't get this in a local debug log. Thus I [enabled debugging on CI](https://github.com/cockpit-project/cockpit/pull/16009/commits/c6f37d4d6e1e7e2aad5856448fee0399bbd7d14f) and [got a log](https://logs.cockpit-project.org/logs/pull-16009-20210629-053857-c6f37d4d-debian-testing/backup.cgs8yd82/log.html#267-1). Indeed that shows that the redis-server proxy has not been initialized for a long time, until after closing the dialog.

This is a bit tricky -- we can't just  generally wait for it, as it will never start to exist if the package is actually not installed. I have one theory -- the proxy is not valid until `exists !== null`. I pushed a refined log to show that, plus a fix to disable the button in that state (that was just an oversight, and needs to happen anyway).

KKoukiou (approved):
Thanks! Nice to see all the bits gathering to a complete feature!

PR #16015, opened 2021-06-29
metrics: Replace moment with Intl.DateTimeFormat

martinpitt:
moment is deprecated, and the metrics page only uses absolute dates/times.

Clean up the generation of the "last 14 days" selector to avoid the two
useless back-and-forth conversions, and handle the "Today" case in a
more symmetric fashion.

Avoid any locale based formatting altogether with the journal timestamp
computations, as they expect a fixed format anyway. (We could also do
this with `@unixtimestamp`, but this does not generate human friendly
values in the Logs page search box.)

See #15979

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -18,7 +18,6 @@
  */
 
 import React, { useState } from 'react';
-import moment from "moment";
 
 import {
     Alert,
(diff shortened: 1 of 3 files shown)


martinpitt:
Off, many thanks @garrett for spotting this! This is  definitively a blocker.

martinpitt:
Indeed:
```js
new Intl.DateTimeFormat('pt_BR', { dateStyle: "full" }).format()
Uncaught RangeError: Incorrect locale information provided
    at new DateTimeFormat (<anonymous>)
```
It needs to "pt-BR" or "pt-br". Same for zh-CN/zh-TW, it works fine with a dash.

martinpitt:
I added a fix, and fixed the same crash on the users page (from PR #16002 ). Unfortunately the users and metrics pages have non-overlapping date/time formats, but that locale handling is now the kind of commonality which we should move into a library. I'll look at that during the next page porting.

martinpitt:
With that crash in mind, I think I'll port the Overview page next -- then check-pages will validate it in all languages automatically.

martinpitt:
release blocker due to the users page crash fix, which is already on main.

KKoukiou (approved):
I wonder if we could better renamed the cockpit.language instead of that underscore hack here. But this is good as is.

martinpitt:
@KKoukiou : Don't worry, that `dateFormatLang` thing won't stay around for long. I already moved/renamed it in PR #16019

PR #16018, opened 2021-06-29
Two urgent Metrics page fixes

martinpitt:
This was the root cause of my woes in https://github.com/cockpit-project/cockpit/pull/15812#issuecomment-849514020 and  https://github.com/cockpit-project/cockpit/pull/16009#issuecomment-870290908  . These are part of #16009, but that may still need some discussions/design iterations -- let's fix these, as they also cause test flakes.


KKoukiou (approved):
Great to see the test clean up as well

PR #16019, opened 2021-06-30
Create time format helper library, port Overview and Journal pages from moment to that

martinpitt:
This is now the third page that moves from moment.js to Intl.DateTimeFormat, and we have enough commonality to factorize this. The Overview page is also part of check-page's "iterate through all known languages", so we get some nice test coverage with this.

 - [x] Builds on top of PR #16015

This changes the format a bit, but it's now consistent between users, metrics, and overview pages:

German on main:
main-systemtime
main-lastlogin

German on this branch:
image
image

English on main:
image
image

English on this branch:
image
image

Journal details page header on main:
image

Journal details page header on this branch:
image

### DIFF
--- a/pkg/lib/journal.js
+++ b/pkg/lib/journal.js
@@ -18,9 +18,7 @@
  */
 
 import cockpit from "cockpit";
-import moment from "moment";
-
-moment.locale(cockpit.language);
+import * as timeformat from "timeformat";
 
 const _ = cockpit.gettext;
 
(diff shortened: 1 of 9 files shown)


martinpitt:
The [publish-dist run failed](https://github.com/cockpit-project/cockpit/runs/2952349177?check_suite_focus=true) on "URI expired", even though [build-dist](https://github.com/cockpit-project/cockpit/pull/16019/checks?check_run_id=2952302723) just ran a few minutes earlier. Huh?

@allisonkarlitskaya : That may or may not get fixed with merging the workflows. Let's see :grin: For now, I suppose not much reason to panic, this happens seldom enough.

martinpitt:
Bah, almost green, but I forgot to update the XXX in the commit message. Done.

KKoukiou (approved):

PR #16021, opened 2021-06-30
autogen.sh: minimalistic rewrite

allisonkarlitskaya:
Take the axe to autogen.sh and replace it with a clean minimalistic rewrite.


martinpitt (approved):
Very cool! 5-line autogen.sh (with 50% empty lines in between :grin: )

PR #16022, opened 2021-06-30
lgtm.yml: don't webpack when checking C code

allisonkarlitskaya:
Add a lgtm.yml file to override the default command for the C build: we
gain nothing from running the full webpack build, so skip it.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,5 @@
+extraction:
+  cpp:
+    index:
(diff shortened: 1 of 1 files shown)


martinpitt (approved):

allisonkarlitskaya:
Seems to drop the C++ analysis time from 25 minutes to about 12.  Not sure why it takes so long, still...

PR #16025, opened 2021-06-30
build: stop using gresource

allisonkarlitskaya:
We use glib-compile-resources to get a single html file bundled.  That
seems like overkill.

Ideally, we'd use #embed for this, but that seems like it'll take a
while longer, so use a small Python script to convert the .html into C,
and compile the resulting file in the usual way.

Formerly, cockpitwebresponse would default to serving a minimalist
internally-hardcoded document for errors, unless the resource was
provided.  Since this resource was provided from both -bridge and -ws,
the internal document was only used for testing.  That was for good
reason: the full document would be a pain to test on, and we'd have to
update the tests any time someone modified the html.

Let's turn the logic around there and default to the real document and
give the tests a mechanism to inject their minimal one.  We use a
pointer instead of a constant, allowing it to be directly overridden
from the tests, and simplifying the code in cockpitwebresponse a bit.

We tried another approach with the usual `ld -b binary` tricks to
convert the file directly to a linkable symbol, but this was causing
strange crashes on rhel and fedora.  Going via C is a sure thing.

### DIFF
--- a/src/bridge/test-packages.c
+++ b/src/bridge/test-packages.c
@@ -1209,6 +1209,10 @@ main (int argc,
 
   cockpit_test_init (&argc, &argv);
 
+  extern const gchar *cockpit_webresponse_fail_html_text;
+  cockpit_webresponse_fail_html_text =
+    "<html><head><title>@@message@@</title></head><body>@@message@@</body></html>\n";
+
   g_test_add ("/packages/simple", TestCase, &fixture_simple,
               setup, test_simple, teardown);
   g_test_add ("/packages/forwarded", TestCase, &fixture_forwarded,
--- a/src/common/cockpitwebresponse.c
+++ b/src/common/cockpitwebresponse.c
@@ -39,14 +39,6 @@
 #include <string.h>
 
 /**
- * Certain processes may want to have a non-default error page.
- */
-const gchar *cockpit_web_failure_resource = NULL;
-
-static const gchar default_failure_template[] =
-  "<html><head><title>@@message@@</title></head><body>@@message@@</body></html>\n";
-
-/**
  * CockpitWebResponse:
  *
  * A response sent back to an HTTP client. You can use the high level one
@@ -1121,9 +1113,7 @@ cockpit_web_response_error (CockpitWebResponse *self,
   gchar *reason = NULL;
   gchar *escaped = NULL;
   const gchar *message;
-  GBytes *input = NULL;
   GList *output, *l;
-  GError *error = NULL;
 
   g_return_if_fail (COCKPIT_IS_WEB_RESPONSE (self));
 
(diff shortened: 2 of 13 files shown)


src/common/cockpitwebresponse.c
```diff
@@ -1172,18 +1162,10 @@ cockpit_web_response_error (CockpitWebResponse *self,
...
-        {
-          g_critical ("couldn't load: %s: %s", cockpit_web_failure_resource, error->message);
-          g_error_free (error);
-        }
-    }
-
-  if (!input)
-    input = g_bytes_new_static (default_failure_template, strlen (default_failure_template));
+  extern char _binary_src_common_fail_html_start[];
+  extern char _binary_src_common_fail_html_end[];
```
martinpitt, line 1166: These are magic symbols created by `ld -b` then?

My gut feeling is to keep it simple, and just put fail.html as a static string here. (defining a string with `#include` is neat, but allegedly only works for C++; and `xxd` is not available in a basic build system environment).

src/bridge/test-packages.c
```diff
@@ -1221,10 +1221,10 @@ main (int argc,
               setup, test_localized_prefer_region, teardown);
   g_test_add ("/packages/localized-fallback", TestCase, &fixture_fallback,
               setup, test_localized_fallback, teardown);
-  g_test_add ("/packages/incompatible/version", TestCase, &fixture_version,
-              setup, test_incompatible_version, teardown);
-  g_test_add ("/packages/incompatible/requires", TestCase, &fixture_requires,
-              setup, test_incompatible_requires, teardown);
+  //g_test_add ("/packages/incompatible/version", TestCase, &fixture_version,
```
martinpitt, line 1224: That's just temporary because the tests need to be updated? That's a bit surprising, this merely changes the way that the file gets included, but not its actual contents?

src/common/cockpitwebresponse.c
allisonkarlitskaya: I went through all the possibilities:
 - `xxd` would be an annoying dependency and it wouldn't advance my goal to evict stuff from `BUILT_SOURCES`
 - static string is really hard when you want to have newlines and stuff in it, without making the entire file hard to read:
```
"<html>\n"
"  <body>\n"
"    <p>\n"
...
```
 - C++11 has "raw strings" which start with `R"(`, end with `)"` and include newlines, and `gcc` even supports them in `C`, but clang doesn't
 - `#embed` is coming, but probably not for another year or two

src/bridge/test-packages.c
allisonkarlitskaya: The problem is that the gresource machinery was so heavy-handed that we used it only from the real executables (`-bridge`, `-ws`) and a handful of — but not all — tests.

When the resource isn't available, the code would fall back on an internal simplified version, and that's what the tests are expecting to see.

src/common/cockpitwebresponse.c
martinpitt: Right, it's fine, just wanted to explore this a bit.

allisonkarlitskaya:
I wanted to make sure the `ld` hacks used here were supported on other architectures so I did a koji build, results here: https://koji.fedoraproject.org/koji/taskinfo?taskID=79472362

martinpitt (approved):
Beautiful, thanks!

martinpitt:
Whoops, we forgot to land this. Rebased to resolve conflict, and will land on green.

martinpitt (dismissed):

martinpitt:
Testing farm results look disastrous -- but they look exactly like in PR #16681, so presumably unrelated. We need to sort this out first *sigh*

martinpitt:
Debian/Ubuntu/arch are fine, but completely breaks Fedora/RHEL. Moving back to draft.

martinpitt (approved):
Nice and clean, thanks!

martinpitt:
This [c8s failure](http://artifacts.dev.testing-farm.io/f49e6129-2232-41b1-8dc1-1a231bf8e383/) is interesting -- several completely unrelated tests got a message `couldn't create runtime dir: /run/user/1001: Permission denied`. This is unrelated, though.

PR #16030, opened 2021-07-01
clean up dist rules

allisonkarlitskaya:
build: clean up distdir rules
    
    I've always wondered why we split our dist-hook up so randomly: it seems
    like some group of things is in one place and another group in another.
    Then there's the one for displaying the error message for the docs.
    
    Replace this with a combined override rule for `distdir`.  We're now in
    control of the process: of course, we call distdir-am first, but after
    that, we can do everything we need, and from one place instead of three
    different ones.
    
    In addition to cleanup, this has two positive impacts:
    
     - `make dist` after `--disable-doc` will now fail immediately, not
       after a bunch of work is already done
    
     - we no longer depend on `$(BUILT_SOURCES)` here, which avoids
       needlessly building some files which won't be disted anyway

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -217,6 +217,12 @@ tools/debian/copyright: tools/debian/copyright.template $(MANIFESTS)
 	$(AM_V_GEN) $(srcdir)/tools/build-debian-copyright > $@.tmp
 	$(MV) $@.tmp $@
 
+# This is the stock rule with $(BUILT_SOURCES) dependency removed.
+# In the future, we can do this via "no-dist-built-sources" automake option:
+# https://git.savannah.gnu.org/cgit/automake.git/commit/?id=13659a7385
+distdir:
+	$(MAKE) $(AM_MAKEFLAGS) distdir-am
+
 # Build up the distribution using $COMMITTED_DIST and include node_modules licenses
 # also automatically update minimum base dependency in RPM spec file
 dist-hook:: $(MANIFESTS)
--- a/src/common/Makefile-common.am
+++ b/src/common/Makefile-common.am
@@ -228,6 +228,8 @@ test_version_LDADD = $(libcockpit_common_a_LIBS)
 test_webresponse_SOURCES = \
 	src/common/test-webresponse.c \
 	src/common/mock-io-stream.c src/common/mock-io-stream.h \
+	$(NULL)
+nodist_test_webresponse_SOURCES = \
 	$(COCKPIT_ASSETS) \
 	$(NULL)
 test_webresponse_CFLAGS = $(libcockpit_common_a_CFLAGS)


allisonkarlitskaya:
With the other PRs merged, this now gets us to a place where `make dist` doesn't do any unnecessary work: it only builds the guide, and the `debian/copyright` file, and we do intend to dist those.

Makefile.am
```diff
@@ -212,13 +199,19 @@ tools/debian/copyright: tools/debian/copyright.template $(MANIFESTS)
...
 
-# Build up the distribution using $COMMITTED_DIST and include node_modules licenses
-# also automatically update minimum base dependency in RPM spec file
-dist-hook:: $(MANIFESTS)
+distdir: $(WEBPACK_INPUTS) $(WEBPACK_OUTPUTS) $(WEBPACK_TEST_DEPS) $(MANIFESTS)
+if !ENABLE_DOC
+	@echo "*** doc must be enabled (ie: --enable-doc) in order to make dist or distcheck"
+	@false
+endif
+	$(MAKE) $(AM_MAKEFLAGS) distdir-am
```
martinpitt, line 207: Random comment from the peanut gallery: I find `dist-hook::` much more elegant than a recursive `make` call with a magic internal target. However, I do agree to merging all the `dist-hook::`s into one.

allisonkarlitskaya: Funny: the magical make invocation with the internal target needs to stay, but `dist-hook:` is back anyway, specifically because we need the recursive make invocation to get the correct values of `$(WEBPACK_INPUTS)` and so after the build completes.

Makefile.am
```diff
@@ -212,6 +204,13 @@ tools/debian/copyright: tools/debian/copyright.template $(MANIFESTS)
 	$(AM_V_GEN) $(srcdir)/tools/build-debian-copyright > $@.tmp
 	$(MV) $@.tmp $@
 
+distdir:
```
martinpitt, line 207: Hmm, do we now need to use `distdir:` and that recursive make exclusively to run the `ENABLE_DOC` check as the first thing? TBH I'd rather pay the price of doing some unnecessary stuff before failing (how often is that going to happen anyway) and keep using the official API/hooks instead (and avoid recursively calling `make` -- this is such a trap)

allisonkarlitskaya: No.  This is really about the fact that the default `distdir:` rule depends (inexplicably) on `$(BUILT_SOURCES)`.  We could come up with ways to get stuff out of `BUILT_SOURCES` (cf. #16025 and some even more exotic efforts for gdbus-codegen) or we could just avoid the problem entirely, which is now my preferred solution.

allisonkarlitskaya: We could also just decide we don't care about this.  I'm using this "clean up `make dist`" approach as a way to seek out legitimate bugs in our build system, and have already gotten quite a few of them (including the first commit of this PR, which absolutely should go in) but avoiding extra work for the sake of it is maybe not the most important thing in the world...

martinpitt: Ah I see, I wasn't aware of implicit `BUILT_SOURCES` dependency. I guess it's a bit philosophical whether these should be in the dist tarball, but `gdbus-codegen` and `glib-compile-resources` (the only two cases of BUILT_SOURCES, AFAICS) are already covered by our build deps.

So, I don't veto this, and don't have a strong opinion, but this seems to introduce hacks instead of getting rid of them. There's some tension between "minimal dist" and "simple Makefile.am" here :sob:

allisonkarlitskaya: I agree, completely.  I'm happy to drop the second commit and only do the first one.

Might also file a bug against automake while I'm at it.  This is a strange design decision.  Probably they'd advise me that if I don't like it, I should just override the `distdir:` target :rofl:

allisonkarlitskaya: I filed https://debbugs.gnu.org/cgi/bugreport.cgi?bug=49317

Let's just go with the first commit here.  The extra building is silly, bit nothing to lose sleep over.

allisonkarlitskaya: automake seems to agree that this is a bug and proposed adding an option to fix it, and asked me to write a patch.  Cool.  I love perl.  Uh...

I feel better about the hack, to be honest, if I can add a comment pointing to an upstream commit showing that it will eventually be removed.

allisonkarlitskaya: The patch went upstream.  I added a comment referencing it and removed the docs stuff in order to keep it purely related to what upstream accepted.

allisonkarlitskaya:
The packit fail is interesting here: there seems to be an annoying race: when the `synchronize` event triggers `pull_request_target`, the `merge_commit_sha` in the event blob references the *old* merge commit from before the force-push.  That was weeks ago, and the mismatch is causing a failure to unpack the `dist/`.  Ugh.

Unrelated failure, to be sure.

martinpitt (approved):
Thanks! Nice work on automake!

Needs some test retries.

PR #16031, opened 2021-07-01
make-bots: support symlinking to another bots dir

allisonkarlitskaya:
We probably all have a checkout of `bots/` somewhere already.  Martin
mentioned that he likes to setup a symlink to it when creating a new
worktree.  Let's automate that process in an elegant way.

Modify `tools/make-bots` to check for the presence of a symlink at
`~/.config/cockpit-dev/bots`.  If one is found, use the directory that
it points to as a target for a symlink from `bots`, as an alternative
to checking out our own copy.

A small matter of taste: we print the (relative) directory name of the
specified bots directory in the log message, but we actually link to the
`~/.config/cockpit-dev/bots` symlink itself.  That way you can update the
symlink as-needed (if working on another branch, for example) and all
existing cockpit repository worktrees will automatically switch over.

### DIFF
--- a/tools/make-bots
+++ b/tools/make-bots
@@ -12,11 +12,16 @@ set -eu
 cd "$(realpath -m "$0"/../..)"
 . tools/git-utils.sh
 
-if [ ! -d bots ]; then
+if [ ! -e bots ]; then
     [ -n "${quiet}" ] || set -x
-    # it's small, so keep everything cached
-    fetch_to_cache ${COCKPIT_BOTS_REF+"${COCKPIT_BOTS_REF}"}
-    clone_from_cache "${COCKPIT_BOTS_REF-master}"
+    if [ -h ~/.config/cockpit-dev/bots ]; then
+        message SYMLINK "bots → $(realpath --relative-to=. ~/.config/cockpit-dev/bots)"
+        ln -sfT "$(realpath --relative-to=. ~/.config/cockpit-dev)/bots" bots
+    else
+        # it's small, so keep everything cached
+        fetch_to_cache ${COCKPIT_BOTS_REF+"${COCKPIT_BOTS_REF}"}
+        clone_from_cache "${COCKPIT_BOTS_REF-master}"
+    fi
 else
     echo "bots/ already exists, skipping"
 fi


tools/make-bots
```diff
@@ -12,11 +12,16 @@ set -eu
...
 
-if [ ! -d bots ]; then
+if [ ! -e bots ]; then
     [ -n "${quiet}" ] || set -x
-    # it's small, so keep everything cached
-    fetch_to_cache ${COCKPIT_BOTS_REF+"${COCKPIT_BOTS_REF}"}
-    clone_from_cache "${COCKPIT_BOTS_REF-master}"
+    if [ -h ~/.config/cockpit-dev/bots ]; then
+        message SYMLINK "bots → $(realpath --relative-to=. ~/.config/cockpit-dev/bots)"
+        ln -sfT "$(realpath --no-symlinks --relative-to=. ~/.config/cockpit-dev/bots)" bots
```
martinpitt, line 19: This causes a strange symlink:

    bots -> ../../../../../home/martin/.config/cockpit-dev/bots

i.e. it's relative to my checkout dir, which is conceptually wrong -- the checkout dir should not be part of the definition of "where is bots". Also, going all the way back to / makes a relative symlink rather pointless?

*either* `--relative-to=. ~/.config/cockpit-dev/bots` *or* `--no-symlinks` both work fine -- in the first case I get `../bots`, in the second I get /home/martin/.config/cockpit-dev/bots . The second is slightly more correct, but I don't have a strong opinion there.

martinpitt (changes requested):
This is a really nice one! As a developer I can now set this once and for all.

tools/make-bots
allisonkarlitskaya: I guess this is related to your adventures with your home directory not being sure if it lives in /var/home or /home?

I'm also using ostree and I don't get this problem.  IIRC you had played around with something there, though?

martinpitt: Yes, passwd says `/home/martin`, but in OSTree there's `/home → /var/home`, that may explain the confusion. Interesting to hear that it works for you, as you also use OSTree? But still, why the `--relative-to` conceptually? This is begging for such weird results.

allisonkarlitskaya: My `$HOME` is `/var/home/lis`.  I think your problem is that have a non-canonical path as your homedir.

martinpitt: Right, impedance mismatch between systemd-homed and rpm-ostree..

allisonkarlitskaya: Right.  So the issue here is that the `--no-symlinks` flag that was intended to prevent the resolution of the `bots` symlink itself also prevented the resolution of the home directory symlink in the case that `$HOME` was set to a non-canonical path.

I fixed that by moving the "/bots" component of the path outside of the `realpath` call, and removing `--no-symlinks`.  That means we get a properly canonicalised (and relativized) path for the segment leading up to ~/.config, but we keep the bots link itself.

martinpitt (approved):
`bots -> ../../.config/cockpit-dev/bots` :tada: 

Danke!

PR #16032, opened 2021-07-01
npm-install workflow: ignore *.pyc files

allisonkarlitskaya:
These are being created/deleted semi-randomly, resulting in the
submodule showing up with a "dirty" diff in the parent project.
    
Ignore the noise.

bonus: fix up a bug in the error handling code of the workflow that I caught while testing it.

The working case can be observed here:
 - https://github.com/allisonkarlitskaya/cockpit/pull/11#issuecomment-872198505
 - https://github.com/allisonkarlitskaya/node-cache/blob/d981c33c850313d78f0ff81bd1650b24c7f9ef8d/.gitignore

The functional failure handling can be seen here (in response to a deliberately introduced syntax error in `package.json`, for testing purposes):
 - https://github.com/allisonkarlitskaya/cockpit/pull/11#issuecomment-872202367

### DIFF
--- a/.github/workflows/npm-install.yml
+++ b/.github/workflows/npm-install.yml
@@ -257,5 +257,5 @@ jobs:
                 owner: context.repo.owner,
                 repo: context.repo.repo,
                 issue_number: context.issue.number,
-                body: `> /npm install\n\nThe workflow failed.  See here for more information: https://github.com/${owner}/${repo}/actions/runs/${context.runId}`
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Danke!

allisonkarlitskaya:
Now that this is landed, it will be used next time someone does a `/npm install` and merges the resulting change.

PR #16043, opened 2021-07-05
storaged: Check transient hostname as well for RAID name

martinpitt:
Some hosts don't have a static host name in /etc/hostname, or it is set
to "localhost" (which some OSes also count as "not really set"). These
get their host names via DHCP, which hostnamed considers "transient".

Consider the transient host name as well to decide whether a RAID is
local or not.

-----

This was spotted in https://github.com/cockpit-project/bots/pull/2168 , see that PR for details. I trigger an additional run against that PR to make sure it works now.

### DIFF
--- a/pkg/storaged/utils.js
+++ b/pkg/storaged/utils.js
@@ -172,15 +172,21 @@ export function mdraid_name(mdraid) {
     if (!mdraid.Name)
         return "";
 
-    var parts = mdraid.Name.split(":");
+    const parts = mdraid.Name.split(":");
 
     if (parts.length != 2)
         return mdraid.Name;
 
-    /* if we call hostnamed too early, before the dbus.proxy() promise is fulfilled,
-     * it will not be valid yet; it's too inconvenient to make this
-     * function asynchronous, so just don't show the host name in this case */
-    if (hostnamed.StaticHostname === undefined || parts[0] == hostnamed.StaticHostname)
+    /* Check the static (from /etc/hostname) and transient (acquired from DHCP server via
+     * NetworkManager → hostnamed, may not exist) host name -- if either one matches, we
+     * consider the RAID a local one and just show the device name.
+     * Otherwise it's a remote one, and include the host in the name.
+     *
+     * However: if we call hostnamed too early, before the dbus.proxy() promise is
+     * fulfilled, it will not be valid yet (hostnamed properties are undefined);
+     * it's too inconvenient to make this function asynchronous, so just don't
+     * show the host name in this case. */
+    if (hostnamed.StaticHostname === undefined || parts[0] == hostnamed.StaticHostname || parts[0] == hostnamed.Hostname)
         return parts[1];
     else
         return cockpit.format(_("$name (from $host)"),
--- a/pkg/storaged/test-util.js
+++ b/pkg/storaged/test-util.js
@@ -67,9 +67,15 @@ QUnit.test("mdraid_name_remote", function (assert) {
     utils.mock_hostnamed(null);
 });
 
-QUnit.test("mdraid_name_local", function (assert) {
+QUnit.test("mdraid_name_local_static", function (assert) {
     utils.mock_hostnamed({ StaticHostname: "sweethome" });
-    assert.strictEqual(utils.mdraid_name({ Name: "sweethome:mydev" }), "mydev", "expected name for local host");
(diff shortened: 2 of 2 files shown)


martinpitt:
I'm stumped about [this rhel-8-5@bots#2168 test failure](https://logs.cockpit-project.org/logs/pull-16043-20210705-051730-53f5d8db-rhel-8-5-bots-2168/log.html). It acts like it runs cockpit from main instead of this PR, it's exactly the same test failure as the original.  Locally the test succeeds. And yet, cockpit-storaged-247.71.g53f5d8db4-1 is installed.

martinpitt:
Maybe that's a race condition with initializing the hostnamed proxy. I temporarily added a debug commit.

martinpitt:
Of course it [worked fine with the extra debugging](https://logs.cockpit-project.org/logs/pull-16043-20210705-065109-debfe566-rhel-8-5-bots-2168/log.html), yay Heisenbug.

pkg/storaged/utils.js
```diff
@@ -172,15 +172,15 @@ export function mdraid_name(mdraid) {
...
+    const parts = mdraid.Name.split(":");
 
     if (parts.length != 2)
         return mdraid.Name;
 
     /* if we call hostnamed too early, before the dbus.proxy() promise is fulfilled,
      * it will not be valid yet; it's too inconvenient to make this
      * function asynchronous, so just don't show the host name in this case */
-    if (hostnamed.StaticHostname === undefined || parts[0] == hostnamed.StaticHostname)
+    if (hostnamed.StaticHostname === undefined || parts[0] == hostnamed.StaticHostname || parts[0] == hostnamed.Hostname)
```
allisonkarlitskaya, line 183: There's an awful lot going on in that line now, and some of the things are of a different nature.

Is it possible that `hostnamed.Hostname` will also be undefined?

At the very least, this comment should probably be expanded from explaining what the first item is attempting to do to explaining what all 3 items are attempting to do, and what the expected result (ie: showing only the name) is.

allisonkarlitskaya (changes requested):

pkg/storaged/utils.js
martinpitt: Yes, it's possible that hostnamed.Hostname is undefined, during the time when hostname proxy is not initialized yet (as the comment says). In that case it wants to treat all detected RAIDs as local. That behaviour does not change, it just changes the "proxy is initialized" case. If Hostname is undefined, the new comparison is always false, and thus it's a no-op.

martinpitt: I.e. do you want me to add `hostnamed.Hostname === undefined`? I'm not sure if that's a good idea, a machine might legitimately not have a transient host name and never get one, this might break the actual name detection.

@allisonkarlitskaya, do you have a suggestion how to rewrite this? I'm honestly not sure.

allisonkarlitskaya: I think I'm mostly looking for a bigger comment, which explains:

 - what this does (ie: volume name without hostname, mostly for the "on this machine" case)
 - workaround for not having the hostname yet (`undefined`)
 - the other two cases being the static and transient name, and what that means.  i had no idea this was even possible.

martinpitt: Makes sense. How about this?

    /* if we call hostnamed too early, before the dbus.proxy() promise is
     * fulfilled, it will not be valid yet (hostnamed properties are undefined);
     * it's too inconvenient to make this function asynchronous, so just don't
     * show the host name in this case. Otherwise we check the static (from
     * /etc/hostname) and transient (acquired from DHCP server via
     * NetworkManager → hostnamed, may not exist) host name -- if either one matches, we
     * consider the RAID a local one and just show the device name.
     * Otherwise it's a remote one, and include the host in the name */

martinpitt: Version 2:

    /* Check the static (from /etc/hostname) and transient (acquired from DHCP server via
     * NetworkManager → hostnamed, may not exist) host name -- if either one matches, we
     * consider the RAID a local one and just show the device name.
     * Otherwise it's a remote one, and include the host in the name.
     *
     * However: if we call hostnamed too early, before the dbus.proxy() promise is
     * fulfilled, it will not be valid yet (hostnamed properties are undefined);
     * it's too inconvenient to make this function asynchronous, so just don't
     * show the host name in this case. */

allisonkarlitskaya: much nicer, thanks

allisonkarlitskaya (approved):

PR #16047, opened 2021-07-05
workflows: Reduce unit test timeout to 20 minutes

martinpitt:
Sometimes they hang eternally, and it's annoying to be told only that
late. They normally take 13 minutes, and we don't want them to take
longer than 20.


allisonkarlitskaya (approved):
I almost did the same recently.

PR #16049, opened 2021-07-05
test: Ensure core dump limit in TestConnection.testBasic

martinpitt:
Most recent RHEL 8 re-disables core dumps by default by setting
`DefaultLimitCORE=0` (as in RHEL 7). Instead of making assumptions about
the test environment, ensure that core dumping is enabled for hostnamed.

----

Spotted in https://github.com/cockpit-project/bots/pull/2156 , running an extra check against that image refresh.


KKoukiou (approved):

allisonkarlitskaya (approved):
I didn't know you could use newlines to separate commands in `sh -c`.  But of course you can, because why wouldn't you be able to?

Thanks!

PR #16051, opened 2021-07-06
containers/unit-tests: Put back sassc

martinpitt:
This was dropped prematurely in commit 012483503e53, our rhel-8.4 branch
still uses it and backporting the sass changes is too intrusive for a
stable branch.

Also, drop our `COCKPITUOUS_GHCR_TOKEN` from the unit-tests-refresh workflow. The only other consumer of that is cockpituous, which I dropped in https://github.com/cockpit-project/cockpituous/pull/426 . After that, we can eliminate that secret.

 - [x] [Rebuild unit-tests containers](https://github.com/cockpit-project/cockpit/actions/runs/1003182461): Succeeded and [pushed new images](https://github.com/cockpit-project/cockpit/pkgs/container/unit-tests)

### DIFF
--- a/.github/workflows/unit-tests-refresh.yml
+++ b/.github/workflows/unit-tests-refresh.yml
@@ -9,6 +9,9 @@ jobs:
   # we do both builds and all tests in a single run, so that we only upload the containers on success
   refresh:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):
Very happy to see the token cleanup.

PR #16052, opened 2021-07-06
test: Adjust TestGrafanaClient for Grafana 8

martinpitt:
- The login form now requires actual typing to trigger the handlers.

- Dashboard CSS classes changed a bit. Let's avoid any assumption and be
  satisfied with the "Welcome" string appearing anywhere on the page.

- "Save & Test" button is now called "Save & test"

- The "Dashboard" tab's aria label moved from the containing `<li>` into
  the `<a>`. That makes a label-based selector awkward, so instead
  select by link target in the tabs section.

- The "Delete dashboard" button in the data source import does not have
  any useful properties any more, and that check is not important. Drop
  it.

- The dashboard panel classes changed in a non-overlapping way. Replace
  the `.panel-wrapper` selector with a generic `<div>` one.

With these changes the test works for both Grafana 7.5 and 8.0.

----

This blocks the services image refresh in https://github.com/cockpit-project/bots/pull/2183 . This happens to pick up a newer Grafana version.

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -1071,6 +1071,12 @@ class TestGrafanaClient(MachineCase):
         b = self.browser
         mg = self.machines['services']
 
+        # avoid dynamic host name changes during PCP data collection, and start from clean slate
+        m.execute("""systemctl stop pmlogger
+                     systemctl reset-failed pmlogger
+                     rm -rf /var/log/pcp/pmlogger
+                     hostnamectl set-hostname grafana-client""")
+
         # start Grafana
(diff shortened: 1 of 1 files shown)


martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-16052-20210706-120331-5f2f1855-fedora-33/log.html#276) is relevant, looking.

marusak (approved):
Thanks!
Happy to see how it nicely integrates and how we can also test the whole chain including grafana UI itself :)

PR #16053, opened 2021-07-06
lib: fix validation of date and time manual insertion in shutdown/restart dialog

KKoukiou:
The validation is now done when the field loses focus. Even if the
inserted value is invalid, the component now does not overwrite the user
inputted value, rather it leaves it and shows the invalid helper text.

Also translate the aria label for the date picker button.


martinpitt:
[integration tests still fail](https://logs.cockpit-project.org/logs/pull-16053-20210706-185713-6cd1fe49-centos-8-stream/log.html#283)

unit tests fail on pyflakes:

    test/verify/check-system-shutdown-restart:21: 'datetime.timedelta' imported but unused

martinpitt:
Many thanks! This works a whole lot better now, it even accepts both ISO and localized format in English. In German there is some regression now, though: Without changing anything except selecting "Bestimmte Zeit", it immediately starts with "invalid time format":

image

Even though 07.07.2021 is perfectly valid in German. I also can't type any other German date, only ISO works (like 2021-07-12).

That sounds like some missing moment initialization somewhere? I want to change this away from moment.js anyway as the next step, I just wouldn't want this regression to land in the release. I'll take a quick look.

I also suspect integration tests to still fail, as the previous force-push did not change anything functional.

martinpitt:
This gets us closer:
```diff
--- pkg/lib/cockpit-components-shutdown.jsx
+++ pkg/lib/cockpit-components-shutdown.jsx
@@ -86,13 +86,15 @@ export class ShutdownModal extends React.Component {
         }
 
         const time_error = !validateTime(this.state.time);
-        const date = new Date(this.state.date);
+        const date = moment(this.state.date, 'L').toDate();
 
```
The change here turned `this.state.date` from an already parsed moment.toDate() object (from DatePicker's `dateParse=` property) to the raw string. This gets us the correct parsing for the current locale format, i.e. it accepts the same format as it shows.

However, this loses the ability to parse ISO format, and the placeholder in the input box only gives that ("yyyy-MM-dd"). So supposedly we should fall back to `Date` parsing when moment fails?

Showing ISO format as the placeholder seems acceptable, as long as it works. It doesn't seem too easy to pry the locale's date format template out of moment. This may work:
```
> moment.locale('en')
'en'
> moment.localeData()._longDateFormat.L
'MM/DD/YYYY'
> moment.locale('de')
'de'
> moment.localeData()._longDateFormat.L
'DD.MM.YYYY'
```
But it's prying into moment's internal structures, and we'll move away from moment.js RSN anyway. So for now I suppose let's just continue to use the ISO format as placeholder, and make sure that parsing that works. WDYT @KKoukiou ?

martinpitt:
@KKoukiou : I pushed a fix that falls back to ISO date format parsing. I threw all sorts of valid and invalid dates in English and German at it, and also tried to not change focus after input. Everything seems to work correctly now. WDYT?

This still leaves the integration test, looking now.

martinpitt (dismissed):
Indeed, tests seem happy now as well. Thanks @KKoukiou ! As I dabbled with this, do you mind re-reviewing and landing?

test/verify/check-system-shutdown-restart
```diff
@@ -152,11 +153,31 @@ class TestShutdownRestart(MachineCase):
         b.wait_not_val("#shutdown-time-input", "")
         old = b.val("#shutdown-time-input")
         b.set_input_text("#shutdown-time-input", "blah")
+        b.wait_not_present("#delay-helper")
```
martinpitt, line 156: Meh, this is still [causing a race condition](https://logs.cockpit-project.org/logs/pull-16053-20210707-072202-f3bae295-rhel-8-5/backup.6fl6fuzc/log.html#283). It's not that bad, and the parallel test run that got triggered may just get green. In the light of this being release-critical, this can be a follow-up.

martinpitt: It [failed also on firefox](https://logs.cockpit-project.org/logs/pull-16053-20210707-072202-f3bae295-fedora-34-firefox/log.html#283-1), so hard to get green. I can reproduce this locally. It means that the date validation function did not run correctly, as 07/07/2021 (as in the screenshot) is clearly correct. So my suspicion is that the delay-helper was there all along, not triggered by the "blah" input.

martinpitt: Pushed a fixup which *may* help. I've run the test a dozen times locally, and don't get the failure any more (but may just be good/bad luck)

martinpitt:
[this failure](https://logs.cockpit-project.org/logs/pull-16053-20210707-102518-4c41803e-fedora-33/log.html#283) proves that the validation error already happens before typing the wrong time. Unfortunuately there is no screenshot attached

martinpitt:
This is still racy, so let's not block today's release on this.

KKoukiou:
Racy test is exactly the same still there

KKoukiou:
Ok seems the 'lib: disable fields until the initialization for the dialog is finished' commit indeed stabilized the flakiness in the dialog. Let me know if you prefer that I split date-dns commit out of this PR.

martinpitt:
For English this now looks pretty much perfect:

image
image

For German, the date format is a little bit weird (other places say "11.07.", not "11.7"), but certainly good enough:
image

The date pattern "dd.MM.y" is a bit weird, though:

image

Also, it seems one now can't input ISO date format any more?

But -- the perfect is the enemy of the good, and this already works *so* much better than before! I restarted the unrelated test failures, and will review code now.

pkg/lib/cockpit-components-shutdown.jsx
```diff
@@ -109,7 +104,7 @@ export class ShutdownModal extends React.Component {
             return;
         }
 
-        const cmd = ["date", "--date=" + moment(date).format('YYYY-MM-DD') + " " + this.state.time, "+%s"];
+        const cmd = ["date", "--date=" + (new Intl.DateTimeFormat().format(this.state.dateObject)) + " " + this.state.time, "+%s"];
```
martinpitt, line 110: That DateTimeFormat() will output US format, as no locale is given. I confirmed that `date --date` does accept that in non-English locales as well.

pkg/lib/timeformat.js
```diff
@@ -16,9 +23,24 @@ export const time = t => formatter({ timeStyle: "short" }).format(t);
 export const timeSeconds = t => formatter({ timeStyle: "medium" }).format(t);
 // June 30, 2021
 export const date = t => formatter({ dateStyle: "long" }).format(t);
+// 2021-06-30
+export const dateShort = t => formatter().format(t);
```
martinpitt, line 27: The comment is misleading:
```
> new Intl.DateTimeFormat('en').format()
'7/11/2021'
```
This makes sense, as that *is* the US short date format. YYYY-MM-DD is ISO format. We may use that somewhere some day, but right now we are standardizing around localized date formats.

I.e. the comment should be "06/30/2021"

pkg/lib/timeformat.js
```diff
@@ -16,9 +23,24 @@ export const time = t => formatter({ timeStyle: "short" }).format(t);
...
 // Wednesday, June 30, 2021
 export const weekdayDate = t => formatter({ dateStyle: "full" }).format(t);
+
+// The following options are helpful for placeholders
+// yyyy/mm/dd
+export const dateShortFormat = () => locales[dateFormatLangDateFns()].formatLong.date({ width: 'short' });
+
+// Parse a string localized date like 30.06.21 to a Date Object
+export function parseShortDate(dateStr) {
+    const parsed = parse(dateStr, dateShortFormat(), new Date());
```
martinpitt, line 41: I'm a bit surprised that this does not need a locale specification? This tries to parse the date for any existing locale?

martinpitt (approved):
Very nice job figuring out the race condition and converting from moment! There's a tiny comment error and one question, but if that goes green I'm more than happy to land.

martinpitt:
`TestStoragePartitions-testSizeSlider` has become awfully hard to get green, but we need to figure this out separately.

PR #16069, opened 2021-07-08
prune-dist.yml: run in cockpit-dist upload

allisonkarlitskaya:
Successful test run is here: https://github.com/cockpit-project/cockpit/runs/3018284994?check_suite_focus=true


martinpitt (approved):
Nice one! This seems to be a good pattern for workflows that write to repos.

PR #16070, opened 2021-07-08
Try to not step on PF4 when adjusting headings

garrett:
Attempt to solve https://github.com/patternfly/patternfly/issues/4206 by:
- remove our heading override
- change our remaining overrides to not apply to PF4 (anything a "pf-" in the class attribute)
- fix anything that looks wrong afterward

### DIFF
--- a/pkg/apps/application-list.jsx
+++ b/pkg/apps/application-list.jsx
@@ -180,7 +180,7 @@ export class ApplicationList extends React.Component {
             <Page>
                 <PageSection variant={PageSectionVariants.light}>
                     <Flex alignItems={{ default: 'alignItemsCenter' }}>
-                        <h2>{_("Applications")}</h2>
+                        <h2 className="pf-u-font-size-3xl">{_("Applications")}</h2>
                         <FlexItem align={{ default: 'alignRight' }}>
                             <Flex>
                                 {refresh_progress}
--- a/pkg/apps/application.css
+++ b/pkg/apps/application.css
@@ -1,3 +1,5 @@
+@import "@patternfly/patternfly/utilities/Text/text.css";
+
 /* Application list */
 
 .app-list-empty {
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -5,24 +5,28 @@
 
 /* Globally resize headings */
 h1 {
-  font-size: var(--pf-global--FontSize--4xl);
+  --ct-heading-font-size: var(--pf-global--FontSize--4xl);
 }
 
 h2 {
-  font-size: var(--pf-global--FontSize--3xl);
+  --ct-heading-font-size: var(--pf-global--FontSize--3xl);
 }
 
 h3 {
-  font-size: var(--pf-global--FontSize--2xl);
+  --ct-heading-font-size: var(--pf-global--FontSize--2xl);
 }
 
 h4 {
-  font-size: var(--pf-global--FontSize--xl);
+  --ct-heading-font-size: var(--pf-global--FontSize--lg);
 }
 
-h4 {
-  font-size: var(--pf-global--FontSize--lg);
+// Only apply a custom font size when a heading does NOT have a PF4 class
+h1, h2, h3, h4 {
+    &:not([class*=pf-]) {
+        font-size: var(--ct-heading-font-size);
+    }
 }
+
 /* End of headings resize */
 
 a {
--- a/pkg/lib/patternfly/patternfly-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-overrides.scss
@@ -5,8 +5,14 @@ $font-family-mono: SFMono-Regular, menlo, monaco, consolas, "Liberation Mono", C
 
 // Resize headings
 h1, h2, h3, h4, h5, h6 {
-  line-height: var(--pf-global--LineHeight--sm);
-  font-family: var(--pf-global--FontFamily--redhatfont--heading--sans-serif);
+  // Reset to default after PF3+Bootstrap; prepare for PF4
+  font: unset;
+
+  // Only adjust font properties for non-PF4 elements
+  &:not([class*=pf-]) {
+    line-height: var(--pf-global--LineHeight--sm);
+    font-family: var(--pf-global--FontFamily--redhatfont--heading--sans-serif);
+  }
 
   .breadcrumb + & {
     // Vertically align contents of headings directly following breadcrumbs


garrett:
TODO:

- [ ] Fix login alert alignment issue (there should not be a p in the heading, but a span is ok... but this doesn't fix it yet)

garrett:
`node_modules/bootstrap/less/type.less` has a lot of classless h1 - h6; in addition to the changes here in this PR, I think this is affecting Cockpit.

garrett:
OK, I think this might actually "properly" fix the issue within Cockpit until we remove PF3+Bootstrap (which is the real, proper fix).

pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -147,3 +147,9 @@ svg {
         }
     }
 }
+
+// Description sometimes shows up when empty, but this adds space to a grid,
+// causing extra space at the bottom
+.pf-c-alert__description:empty {
+    display: none;
+}
```
garrett, line 155: After changing the fonts, I noticed the log in alert on the overview page was misaligned, as it had an empty description. I couldn't see a description being specified, so I was wondering if this is a PF4 issue or not. Unsure.

Ideally, if we could fix this in the JSX (`pkg/systemd/overview.jsx`, around line 140), that would be great. However, if this is a bug, we need to get it fixed upstream (in the JS). Meanwhile, this CSS will remove it from the layout if it's empty (so there won't be a gap, padding, etc.).

pkg/apps/application-list.jsx
```diff
@@ -180,7 +180,7 @@ export class ApplicationList extends React.Component {
             <Page>
                 <PageSection variant={PageSectionVariants.light}>
                     <Flex alignItems={{ default: 'alignItemsCenter' }}>
-                        <h2>{_("Applications")}</h2>
+                        <h2 className="pf-u-font-size-3xl">{_("Applications")}</h2>
```
garrett, line 183: Changing the heading CSS dropped the size of this heading, so I imported the PF4 utility classes and use the appropriate font size adjustment as a class.

pkg/systemd/overview.jsx
```diff
@@ -133,8 +133,8 @@ class LoginMessages extends React.Component {
         if (!last_login_message && !fail_count_message && !last_fail_message)
             return null;
 
-        const last_log_item = <p id='last-login'><b>{_("Last login:")}</b> {last_login_message}</p>;
-        const last_fail_item = <p id='last-failed-login'><b>{_("Last failed login:")}</b> {last_fail_message}</p>;
+        const last_log_item = <span id='last-login'><strong>{_("Last login:")}</strong> {last_login_message}</span>;
+        const last_fail_item = <span id='last-failed-login'><strong>{_("Last failed login:")}</strong> {last_fail_message}</span>;
```
garrett, line 137: We should _not_ have a paragraph tag `<p>` inside a heading. An inline element like a `<span>` is more appropriate for this. Additionally, we should use semantic elements like `<strong>` instead of meaningless presentation-only elements like `<b>`. (It looks the same, but has meaning.)

This change was an attempt to fix the alignment of the alert information, but it didn't fix it. I adjusted the CSS after and _that_ worked. However, the `<p>` in an `<h4>` is still an issue, so I included the fix here as it would (probably) still impact the alignment, and it's still definitely wrong.

pkg/lib/patternfly/patternfly-4-overrides.scss
KKoukiou: Can you specify the conditions where there was not content in the alert apart from the titlle? I took a look to our code and I was not able to identify such case :( 
Btw it's not an upstream issue, not providing the Alert body will not render the pf-c-alert__descriotion.

garrett: It looks like this:

image

See how the title is misaligned due to having an empty body for the alert?

garrett: Here's what the DOM looks like:

Screenshot from 2021-07-20 15-38-56

I have the offending description body selected to highlight it.

KKoukiou: This needs to be fixed in JS, we currenty past the last_login_message as an empty string to the Alert component, therefore you see this.

pkg/systemd/overview.jsx
garrett: No, that's still needed. H4 is bold by default and in some alerts, it's not normal color text (such as a warning). We have a `<strong>` inside an `<h4>` for the log in alert, for example. By removing this block, it makes the entire string emboldened and the heading uses RedHatDisplay instead of RedHatText. (in others, such as alerts, it affects the color too. It doesn't in this case.)

I just tested this.

Without the block:

Screenshot 2021-09-06 at 10-45-06 Overview - garrett Bolt

With the block:

Screenshot 2021-09-06 at 10-45-59 Overview - garrett Bolt

garrett: Oh, looking at the screenshots, the color override is also in effect here too. (Dark blue to black.)

pkg/lib/patternfly/patternfly-4-overrides.scss
garrett: @KKoukiou: Do you think the empty `last_login_message` needs to be fixed within this PR?

garrett:
@KKoukiou: What's needed for this PR?

1. The CSS block you suggested removing actually still does something and is unrelated to this PR.
2. The JS you suggested fixing seems like it might be out of scope for this PR. (And I have a workaround in CSS to hide an empty grid cell in this PR to fix the alignment issue, so it would not make any visual or screenreader difference, even if the JS fix is the more proper way.) I'd need some help from someone to get the JavaScript changes implemented.

Anything else?

KKoukiou:
@garrett then it's fine, can you please rebase?

garrett:
Rebased, force-pushed

pkg/lib/patternfly/patternfly-4-overrides.scss
KKoukiou: As I mention in the comment above this is the way to fix this. 
Firstly: PF4 overrides always should have a upstream issue which is missing here
Secondly: This looks like that needs to be fixed in JS instead.

garrett: We talked in IRC; this should be moved to the local CSS, as it probably is only triggered by the JS locally. Ideally, the JavaScript would be fixed, but it's a bit out of scope of this PR and I can't do that... I'd need someone else to more properly fix it.

KKoukiou: @garrett this can be properly fixed with:
``` diff
diff --git pkg/systemd/overview.jsx pkg/systemd/overview.jsx
index 1aef9178d..53bdff620 100644
--- pkg/systemd/overview.jsx
+++ pkg/systemd/overview.jsx
@@ -143,6 +143,11 @@ class LoginMessages extends React.Component {
 
         const last_log_item = <span id='last-login'><strong>{_("Last login:")}</strong> {last_login_message}</span>;
         const last_fail_item = <span id='last-failed-login'><strong>{_("Last failed login:")}</strong> {last_fail_message}</span>;
+        const child = [];
+        if (last_login_message && fail_count_message);
+            child.append(last_log_item);
+        if (last_fail_message)
+            child.append(last_fail_item);
 
         return (
             <Alert id='login-messages'
@@ -152,8 +157,7 @@ class LoginMessages extends React.Component {
                    actionClose={<AlertActionCloseButton onClose={this.close} />}
                    title={fail_count_message || last_log_item}
             >
-                {last_login_message && fail_count_message && last_log_item}
-                {last_fail_message && last_fail_item}
+                {child.length ? child : null}
             </Alert>
         );
     }
```

but it will conflict with Jelles PR so let's just remove and let him fix it.

garrett:
Dropped PF4 override; squashed; rebased

KKoukiou (approved):

garrett:
> pixel tests failed - in superuser dialog. Can you take a look?

It looks fine.

Screenshot 2021-09-07 at 13-29-33 Overview - garrett Bolt

Screenshot 2021-09-07 at 13-29-21 Overview - garrett Bolt

Pixel tests will fail if they include headings, as sizes are different in places. That's actually the intent of this PR (to fix the sizing, especially for PF4 headings, and the superuser dialog uses PF4, as evidenced by the "pf-c-modal-box__title-text" class).

KKoukiou (approved):

KKoukiou:
SRPM build failed is probably due to this branch not being rebased. This PR is otherwise harmless.

PR #16073, opened 2021-07-09
fmf: Really force EPEL 8 on RHEL 9

martinpitt:
Do the same as in cockpit-machines:
https://github.com/cockpit-project/cockpit-machines/blob/main/test/browser/browser.sh#L23


martinpitt:
I'm frankly quite stunned how this ever worked without this sed in RHEL 9.. Testing this in https://src.osci.redhat.com/rpms/cockpit/pull-request/71#

marusak (approved):
Thanks!
Have you seen it failing? Because if it works without it, then possibly we don't need the whole special epel case?

martinpitt:
@marusak : Yes, I did. See the linked OSCI PR.

martinpitt:
Even though the test results are a disaster now, installing chromium at least went fine :grin:

PR #16076, opened 2021-07-09
workflows: rename environments and secrets

allisonkarlitskaya:
After a bit of a trial period, the names `cockpit-dist upload` and
`COCKPIT_DIST_DEPLOY_KEY` seem a bit awkward.  Rename them to just
`cockpit-dist` and `DEPLOY_KEY`.

Ditto `node-cache upload` and `NODE_CACHE_DEPLOY_KEY`.

The new environments and secrets have already been deployed.  The old
ones can be pulled down after we land this.

### DIFF
--- a/.github/workflows/npm-install.yml
+++ b/.github/workflows/npm-install.yml
@@ -139,7 +139,7 @@ jobs:
     needs: [start, build]
     if: ${{ needs.build.result == 'success' }}
     runs-on: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


martinpitt (approved):

allisonkarlitskaya:
`jumpstart / publish` was an expected fail since the public half of the deploy key from the old environment was already replaced in the target repo.

PR #16077, opened 2021-07-09
po-refresh: eliminate COCKPITUOUS_TOKEN

allisonkarlitskaya:
Replace the use of the COCKPITUOUS_TOKEN in our po-refresh GitHub
workflow with a mixture of two new tightly-scoped credentials:

  - for pushing to the `cockpit-weblate` repo we use a deploy key (and
    switch to pushing via ssh)

  - for pushing to the `cockpit` repo and opening the PR, we use the
    GITHUB_TOKEN from github-actions itself (still via https).  We add a
    `permissions:` line to ensure we have only the required permissions.

Did a test run here: https://github.com/cockpit-project/cockpit/runs/3027172789?check_suite_focus=true
Successfully pushed this: https://github.com/cockpit-project/cockpit-weblate/commit/a8ac4ccb48b6c12a2f92f759ec07ef037315448b

...but didn't open a PR :(

### DIFF
--- a/.github/workflows/po-refresh.yml
+++ b/.github/workflows/po-refresh.yml
@@ -9,6 +9,11 @@ on:
 jobs:
   po-refresh:
     runs-on: ubuntu-20.04
+    permissions:
+      contents: write
+      pull-requests: write
+      statuses: write
+    environment: cockpit-weblate
     steps:
       - name: Set up dependencies
         run: |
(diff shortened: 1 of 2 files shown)


marusak:
> ...but didn't open a PR :(

Now it should. Can you please ru-run?

.github/workflows/po-refresh.yml
```diff
@@ -20,21 +24,26 @@ jobs:
...
-          # the default GITHUB_TOKEN would override our ~/.git-credentials from above
-          token: '${{ secrets.COCKPITUOUS_TOKEN }}'
 
       - name: Run po-refresh bot
         run: |
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.DEPLOY_KEY }}'
+
+          mkdir -p ~/.config/cockpit-dev
+          echo '${{ secrets.GITHUB_TOKEN }}' > ~/.config/cockpit-dev/github-token
```
martinpitt, line 43: Why does that need to happen again? Does something overwrite it? It's already done in "Set up.." step above.

allisonkarlitskaya: Oops!

allisonkarlitskaya:
→ #16422

Issue #14752, opened 2020-10-15
Validate all URLs


cockpituous:
Checked 117 URLs out of which 1 is/are invalid:
https://www.patternfly.org/v4/documentation/react/components/table

PR #16080, opened 2021-07-09
pkg: Update pf url

marusak:
See #14752

I mostly just wanted to open a PR since it has been a while :)
But also I am wondering, if we want to check redirections for PF repo as they seem to be transferring issues a lot.


martinpitt (approved):
Danke!

PR #16087, opened 2021-07-12
build cleanups for cockpit.pot generation

allisonkarlitskaya:
Here's a bunch of changes to `po/Makefile.am` to make it possible to build `cockpit.pot` without needing to do the whole automake thing.

You can see it in use from a new workflow here: 
 - https://github.com/allisoninc/cockpit-weblate/blob/main/.github/workflows/sync-pot.yml
 - https://github.com/allisoninc/cockpit-weblate/runs/3048500342?check_suite_focus=true

Together with the new workflow this solves one half of the "what do we do about weblate?" cockpituous token puzzle: this part ends up requiring no credentials at all, since it runs out of the -weblate repository.

The other half still open, of course, is getting the .po files from the weblate back into cockpit and opening a PR and running tests.  That's gonna be a bit of work still....

The changes here are more or less harmless on their own and can go in already, I think.  Then I'd probably try to land the `cockpit-weblate` `sync-pot` workflow at the same time as disabling the synchronisation from the `cockpit` repo side.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -235,7 +235,7 @@ bots:
 	tools/make-bots
 
 multi-word: po/cockpit.pot
-	$(MSGGREP) --no-wrap --msgid -e '\S\+\s' po/cockpit.pot | grep "^msgid " | cut -d " " -f2- | sort --reverse > titles
+	msggrep --no-wrap --msgid -e '\S\+\s' po/cockpit.pot | grep "^msgid " | cut -d " " -f2- | sort --reverse > titles
 
 title2sentence: multi-word
 	tools/title2sentence.py -i titles -o t2s.sh
--- a/po/Makefile.am
+++ b/po/Makefile.am
@@ -82,12 +82,12 @@ po/cockpit.manifest.pot: package-lock.json
 
 po/cockpit.appstream.pot:
 	$(AM_V_GEN) $(MKDIR_P) $(dir $@) && \
-	cd $(srcdir) && GETTEXTDATADIRS=$(srcdir)/po xgettext --output=$(abs_builddir)/$@ \
+	cd $(srcdir) && GETTEXTDATADIRS=$(srcdir)/po xgettext --output=$@ \
 		$$(find $(TRANSLATE) -name '*.appdata.xml.in')
 
 po/cockpit.polkit.pot:
 	$(AM_V_GEN) $(MKDIR_P) $(dir $@) && \
-	cd $(srcdir) && GETTEXTDATADIRS=$(srcdir)/po xgettext --output=$(abs_builddir)/$@ \
+	cd $(srcdir) && GETTEXTDATADIRS=$(srcdir)/po xgettext --output=$@ \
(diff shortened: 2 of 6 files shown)


martinpitt:
A few unrelated test failures; [f34/devel](https://logs.cockpit-project.org/logs/pull-16087-20210712-163822-6a95f0da-fedora-34-devel/log.html#188) is an unspecific cockpit-pcp crash without a traceback. The [failed bonding pixel test](https://logs.cockpit-project.org/logs/pull-16087-20210712-163822-6a95f0da-fedora-34/log.html#64-2) is rather alarming, though -- the close button is missing. 

In PF4 this is controlled by the `showClose=` property, but we don't set that anywhere other than pkg/lib/cockpit-components-dialog.jsx (but there it's hardcoded to false), and pkg/networkmanager does not use that component. It's still PF3/jQuery.

In that particular dialog, the close button has ID `#network-bond-settings-close-button`, but it's only being used to attach a click handler, not show/hide it. Does that ring a bell, @KKoukiou ? I can send a PR to just wait for the close button to appear, but as it's static HTML, how on earth could it not be visible?

martinpitt:
Same with the not yet visible expand triangle for the MAC. But at least this is dynamically built in PageNetworkBondSettings.update(). So maybe let's wait for that.

Makefile.am
```diff
@@ -235,7 +235,7 @@ bots:
 	tools/make-bots
 
 multi-word: po/cockpit.pot
-	$(MSGGREP) --no-wrap --msgid -e '\S\+\s' po/cockpit.pot | grep "^msgid " | cut -d " " -f2- | sort --reverse > titles
+	msggrep --no-wrap --msgid -e '\S\+\s' po/cockpit.pot | grep "^msgid " | cut -d " " -f2- | sort --reverse > titles
```
martinpitt, line 238: FTR, I'm not actually sure why the previous configure.ac did that "static path" detection. There was no way to override it with some `--with` option or so. :+1: on that simplification.

martinpitt (approved):
Thanks! Indeed this looks completely compatible with our current po-refresh bot, so this can go in. Just retrying the failures for good measure (and I also want to see if the pixel test failure is a race or if we broke anything yesterday)

po/Makefile.am
```diff
@@ -67,7 +67,7 @@ po/cockpit.c.pot:
 # Extract translate attribute, Glade style, angular-gettext HTML translations
 po/cockpit.html.pot: package-lock.json
 	$(AM_V_GEN) $(MKDIR_P) $(dir $@) && \
-	$(srcdir)/tools/missing $(srcdir)/po/html2po -d $(srcdir) -o $@ \
```
marusak, line 62: Do we use this script anywhere else or we can drop it as well?

martinpitt: Good point, these two are the last usages.

martinpitt: Actually, tools/missing is generated dynamically by autoconf, not tracked in our git.

PR #16088, opened 2021-07-12
po: Replace `upload-pot` rule with @true

allisonkarlitskaya:
We no longer need to upload cockpit.pot from our side, since
cockpit-weblate is pulling it for itself.

We need to keep the rule, however, since bots/po-refresh assumes that it
exists.  We'll do a cleanup round later.

 - [x] #16087 
 - [ ] cockpit-project/cockpit-weblate#3 (land simultaneously)


martinpitt (changes requested):

allisonkarlitskaya:
→ #16422

PR #16089, opened 2021-07-13
test: Wait for networking dialog initialization before pixel test

martinpitt:
Otherwise it could happen that the dialog is not fully rendered yet.
Observed in the wild with the bonding dialog's MAC dropdown triangle
(presumably as the list was still empty).

----

[Failed test example](https://logs.cockpit-project.org/logs/pull-16087-20210712-163822-6a95f0da-fedora-34/log.html#64-2)


martinpitt:
[too specific name matching](https://logs.cockpit-project.org/logs/pull-16089-20210713-055659-f9e48faf-fedora-coreos/log.html#274-2), we only do the actual pixel test on Fedora 34. It can just look at the VLAN ID instead.

PR #16084, opened 2021-07-12
systemd: Move system uptime from moment to date-fns

martinpitt:
Introduce a new timeformat.distanceToNow() wrapper for this.

Drop moment initialization from systemd page, as this was the last usage
of moment.

----

This looks a bit different now, date-fns adds the "about.." as it is rounded:

image

image

PR #16090, opened 2021-07-13
packagekit: Port from moment to timeformat

martinpitt:
This changes the "last checked" slightly for short time intervals (as
moment's and date-fns' durations are a bit different), and moves from
ISO to our standard localized format for the update history.

 - [x] builds on top of #16084 

"Last checked" on main:

main-en
main-de

With this PR:
pr-en
pr-de

History on main (all langs):

history-main-en

History with this PR (standard localized format that we now use everywhere):
history-pr-en

### DIFF
--- a/pkg/packagekit/history.jsx
+++ b/pkg/packagekit/history.jsx
@@ -17,12 +17,12 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-import moment from 'moment';
 import React from "react";
 import PropTypes from "prop-types";
 
 import { Tooltip } from "@patternfly/react-core";
 import { ListingTable } from "cockpit-components-table.jsx";
+import * as timeformat from "timeformat.js";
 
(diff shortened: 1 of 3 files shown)


martinpitt:
In [one case](https://logs.cockpit-project.org/logs/pull-16090-20210713-074000-5445d489-fedora-34/log.html#272) the 'last checked' still said "a few seconds ago". This is not [documented](https://date-fns.org/v2.22.1/docs/formatDistanceToNow), and neither can I find this in the date-fns code. So at the moment I have no explanation/fix for this.

martinpitt:
argh, [pixel failures](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16090-20210713-195218-7348f853-fedora-34/log.html#148-2) - indeed a test/reference change slipped in through rebasing. Fixed.

KKoukiou (approved):

Issue #15191, opened 2021-01-21
logs: Accessibility issues


garrett:
On the logs page, there are the following accessibility issues (as detected by "Accessibility Insights for Web"):

- [ ] #logs-help-toggle icon is missing an accessible name; _one_ of the following should be addressed:
    - the content inside the `<a></a>` does not describe the link
    - icon (inside the link) does not have fallback text (which would fix the above)
    - there's no aria-label or aria-labelledby
- [ ] #start-box is described by a role of "rowgroup", yet it does not have children element(s) with role="group"
- [ ] .cockpit-log-service-reduced has a role of "cell", but it does not have a parent element with a role of "row"
- [ ] .badge has a role of "cell", but it does not have a parent element with a role of "row"
- [ ] .journal-start has a role of "rowgroup", yet it does not have a parent with a role of grid, table, or treegrid (in this case: table is probably most accurage)

And, in other issues:
- the logs page also is missing a title attribute or equivalent (but this is addressed in another isusue (#15188)
- the logs pace also lacks lang on html (#15189)

marusak:
This is most likely going to be implemented as part of https://github.com/cockpit-project/cockpit/issues/15170

PR #16091, opened 2021-07-13
Port logs page to React + PF4

KKoukiou:
## Logs: ...
Screenshot from 2021-08-09 13-11-59


KKoukiou:
Rebased for the tests to run.

garrett:
Wow, this looks great so far!

---

## Date selector

I wonder if we could use a date selector for the since and until? I guess it could be considered an optimization, probably as a follow-up, and not a part of this PR if so.

image

The two could be part of the same field, like @ https://www.patternfly.org/v4/components/date-picker/react-demos#date-and-time-range-picker.

image


The downside would be relative dates and date strings like "today" and "yesterday", unless the text entry happened to support those too. (It wouldn't look obvious. However, the widget does show the current day by default on popup, so someone could specify a specific day for yesterday rather easily anyway?)



---

## Helper text

The helper text is so jargony and dense. It totally works for a first draft as they do explain the expected values in the fields, but, wow...

I guess some of this has to be there, to help people who know systemd? However, I wonder if we could make it less obtuse for people who don't know or care about systemd.

Even breaking it up into paragraphs should help, as it does look like a wall of "scary" text. And we could probably use language that's easier to parse too. (It's definitely proper English, but hard to read and understand.) Perhaps we could also use a bulleted list with examples?

Do these strings come from the systemd man page? (Doing a search confirms: Yes, this is from the systemd man page.)

As we have a bunch of rich widgets from PatternFly instead of having to resort to command-line flags, we should use more than just text entries with obscure command-line attributes. Using PF widgets would let us drop the (?) helper text entirely for some and replace it with friendlier text for others. 

Some examples of the systemd text popups and suggestions on what widget to replace them with are below. (I already covered the time-based ones above, and the freeform text one has to be a text entry by nature of the content.)

image

Could this be a dropdown select with human-friendly priority levels?

image

Could this be a dropdown with "Current boot", "Previous boot", "X boots ago" (or something similar) up to the highest X number of boots available in the journal?

image

Could this be something like a chip group in a dropdown, specifically a select that's a "Multiple with Render Custom Chip Group"? https://www.patternfly.org/v4/components/select#multiple-with-render-custom-chip-group

garrett:
Here's what a detailed search could look like. As it uses richer widgets instead of just text entries, it's able to get away without all the blocks of text explaining the details to type into a command line. :wink:

logs - search details - compact

Again, this could be in a follow-up; just posting it here to continue the conversation.

KKoukiou:
> Here's what a detailed search could look like. As it uses richer widgets instead of just text entries, it's able to get away without all the blocks of text explaining the details to type into a command line. wink
> 
> logs - search details - compact
> 
> Again, this could be in a follow-up; just posting it here to continue the conversation.
> 
> I'm not sure if we can easily use the existing PF4 widgets to do this (it might be too convoluted?) or would need to coordinate with upstream.

I am a bit worried that this implementation can't be upstreamed into PF4, but also that this does not 100% make sense. This component relies on the fact that whatever you type in the free text can be auto filled in the text fields bellow and vice versa. When some of the text fields are selects or date pickers as above, the current behavior that the text search box is synced with every keyup with the fields is actually not possible.

garrett:
It should be possible to sync the sets of values, as for the dropdown fields only have a specific set of values that journalctl takes. And just having a bunch of inputs is definitely horrible for UX, as evidenced by the necessity of the verbose text, which is quite hard to comprehend. (This is a cognitive accessibility issue too, in addition to a usability one, as a result.) 

There's already string parsing to split the values into the various fields. It could be the case where if it matches (a valid option), then set the string and fallback to treat it like plain text otherwise.

"Since" and "until" are both using dates/times, but if it's not validated as dates, people could still type in "today", "yesterday", and so on. And we could parse that and fill the fields with the same values just as well. The difference between this and the fields is basically to allow someone to input dates and times with graphical widgets.

But, anyway, with this mockup, we have something to show the PF team and get their feedback. The fields we're using are definitely Cockpit-specific, but perhaps they could have some insight on how to make the implementation easier for us.

garrett:
The journal identifier menu's select items do not wrap and are not truncated, so they trigger an overflow in the menu and also overlap with the check icon for selected items:

Screenshot from 2021-07-26 13-15-44

Setting white-space to normal fixes this by allowing a user to see the full string (in many cases, this may be the important and unique part of a system name) and shows the checkmark for selected items:

Screenshot from 2021-07-26 13-17-33

I've pushed the following diff in commit f9d702d.

```diff
diff --git pkg/systemd/logs.scss pkg/systemd/logs.scss
index 8e84963e8..cf462fb47 100644
--- pkg/systemd/logs.scss
+++ pkg/systemd/logs.scss
@@ -107,3 +107,8 @@
        }
    }
 }
+
+// Long names without spaces do not wrap; triggers overflow and overlaps with check icon
+#journal-identifier-menu .pf-c-select__menu-item {
+    white-space: normal;
+}
```

I'm not sure if this is specific to the page or should go upstream. PatternFly is intentionally setting white-space to nowrap; there's probably a reason. At least in this insistence, we need a fix.

As far as the unit names, I guess the foo(bar) pattern is actually part of the unit's name? Here's a screenshot:

Screenshot from 2021-07-26 13-27-21

(I'm not sure why this is a warning, but I guess it's set that way by the service for whatever reason, so it's not a Cockpit issue, I think.)

test/common/run-tests
```diff
@@ -96,6 +96,7 @@ def finish_test(opts, test, affected_tests):
     """
 
     affected = any([test.command[0].endswith(t) for t in affected_tests])
+    affected = True
```
marusak, line 99: argh :/ and I was wondering how the hack I managed to break those tests so badly... All tests were retried. (that goes for local hacks and `git add -u`)

marusak:
@martinpitt want to take a look since I touch it a lot as well?
I'll prepare release notes in the meantime.

test/verify/check-system-services
```diff
@@ -497,6 +456,81 @@ ExecStart=/bin/awk -f /tmp/mem-hog.awk
         # Check that listen is not shown for .service units
         b.wait_not_present("#listen")
 
+    def testLogs(self):
+        m = self.machine
+        b = self.browser
+
+        self.write_file("/etc/systemd/system/test.service",
```
martinpitt, line 463: How about factoring this and bin/test-service into a helper method, so that you don't have to repeat it here.

pkg/systemd/logs.jsx
```diff
@@ -0,0 +1,395 @@
...
+                                                textFilter={textFilter} onTextFilterChange={onTextFilterChange} />
+                                </ToolbarItem>
+                                <ToolbarItem className="text-help">
+                                    <Popover id="logs-help-popover"
+                                             hasAutoWidth
+                                             bodyContent={
+                                                 <div className="logs-help-menu">
+                                                     <span className="help-links">
+                                                         <Button variant="link" component="a" isInline
+                                                                 href="https://www.freedesktop.org/software/systemd/man/journalctl.html"
```
martinpitt, line 205: Did you deliberately drop the [RHEL documentation link](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/reviewing-logs_system-management-using-the-rhel-8-web-console#filtering-logs-in-the-web-console_reviewing-logshere?) I thought it was quite nice. Or does this not apply to the new page any more?

test/verify/check-system-journal
```diff
@@ -419,7 +414,8 @@ ExecStart=/bin/sh -c 'sleep 5; for s in $(seq 10); do echo SLOW; sleep 0.1; done
         m.execute("systemctl start systemd-journald.service")
         m.execute("logger -p user.err --tag check-journal BEFORE BOOT")
         b.go("#/?tag=check-journal")
-        wait_log_lines([expected_date, ["check-journal", "BEFORE BOOT", ""]])
+        # FIXME
+        #  wait_log_lines([expected_date, ["check-journal", "BEFORE BOOT", ""]])
```
martinpitt, line 418: This is still outstanding?

martinpitt (changes requested):
Thanks! The React code and structure looks sane and clean, I didn't go through it in that much detail. But I did find some UI/design issues during interactive testing, I'll follow up with a comment about them.

pkg/systemd/logs.jsx
marusak: Yes. The inline help explains all what is possible and I think it is even more clear now that through the rhel docs.

martinpitt:
- [ ] The top panel reactive behaviour is confusing. On wide displays, it looks fine:

image

once you make it a little smaller, the button wraps around. Doesn't look great, but fair enough, stuff needs to go somewhere:

image

It could at least have some space between the rows, though?

 - [ ] But when making it narrower, it all of a sudden looks like this:

image

There's just the funnel icon, which I only saw on second sight, and presumably also is not accessible. Could it (1) have some text as well, and (2) kick in only on much narrower widths? With the panel using two lines, there's plenty of space to fit everything (the filter input could wrap around), and the funnel expander is then only for mobile mode.

On current main, it looks better:

image

This both has spacing between the rows, as well as actually uses the rows to show everything.

 - [ ] The help icon popup also needs some work: it gets cut off at the right, and generally feels too narrow -- you can hardly read anything in that text. Can we make the arrow get attached to the right side of the box instead of the center, and make the box wider?
 
image

 - [ ] Can we make it so that hitting Enter in the filter bar updates the logs? Right now you have to find (not that obvious!) and click the → button. Enter works in main right now, so this is a regression.

 - [ ] as `priority: err` was the default, I tried `priority: warn` and got no results. Ah, it needs to be `warning`. As there is just a few fixed possible values for this, could the expanded version be a proper select instead? (Garrett already mentioned this above)
 
image

 - [ ] Dropping the Priority filter and merging it into the text box was a conscious decision, or an oversight? I am not sure how many people use this, but at least I certainly do.

marusak:
> The help icon popup also needs some work

This is actually annoying. If I place it to the left, it gets cut:
Screenshot from 2021-08-07 09-09-34left

if placed down:
Screenshot from 2021-08-07 09-21-30noplace

In ideal case, this would not even exist, and it would be part of the filter dropdown.
@KKoukiou already implemented that https://github.com/patternfly/patternfly-react/pull/6043 I'll check if this is already released.

marusak:
Ok, it works, I'll just use that.
Screenshot from 2021-08-07 09-46-12ok

martinpitt:
@marusak : Oh, that is lovely!

marusak:
> Can we make it so that hitting Enter in the filter bar updates the logs? Right now you have to find (not that obvious!) and click the → button. Enter works in main right now, so this is a regression.

Right, I was complaining about this as well. This is not supported in the component itself. Let me see, what we can do about it.

>  As there is just a few fixed possible values for this, could the expanded version be a proper select instead? 

We cannot, it accepts only text fields.

> Dropping the Priority filter and merging it into the text box was a conscious decision, or an oversight? I am not sure how many people use this, but at least I certainly do.

It was conscious decision from my side. But if people use it and it is hard to use in the text form, then I can move it back outside again.

marusak:
> Right, I was complaining about this as well. This is not supported in the component itself. Let me see, what we can do about it.

@KKoukiou Implemented this in PF, https://github.com/patternfly/patternfly-react/pull/6145

test/verify/check-system-services
marusak: done

test/verify/check-system-journal
marusak: fixed

marusak:
All comments were addressed @martinpitt 

Looks like this:
Screenshot from 2021-08-09 13-05-16

When it wraps: (the search bar is too long, but otherwise would be too short on one line - kind of compromise)
Screenshot from 2021-08-09 13-04-55

This is right before it jumps into the collapsed state:
Screenshot from 2021-08-09 13-04-10

Which looks like this:
Screenshot from 2021-08-09 13-05-41

marusak:
I cannot ask for review from @KKoukiou but would like to get her feedback as well

marusak:
Added one more commit. Somewhat unrelated, should not be squashed but it fixes the last accessibility issue as per #15191 thus makes sense to fix here as well.

pkg/systemd/logs.jsx
```diff
@@ -0,0 +1,429 @@
...
+
+        cockpit.location.go([], Object.assign(options, { [newTimeFilter.key]: newTimeFilter.value }));
+    };
+
+    return (
+        <Page>
+            <PageSection id="journal" padding={{ default: 'noPadding' }}>
+                <Toolbar>
+                    <ToolbarContent>
+                        <ToolbarToggleGroup toggleIcon={<><span className="pf-c-button__icon pf-m-start"><FilterIcon /></span>{_("Toggle filters")}</>} breakpoint="md">
```
KKoukiou, line 190: When inserting some text in the searchinput, it grows and wraps to next line.
Screen Shot 2021-08-10 at 12 35 08
Screen Shot 2021-08-10 at 12 33 02

marusak: I cannot reproduce this. Does this happen on some specific browser, some specific width...?
I tried chrome and firefox, different width, typed very long strings and it never wrapped.

allisonkarlitskaya: Can't reproduce that on Firefox here, but the styling looks off: 
Screenshot from 2021-08-11 11-12-55

In particular, the [X] having a blue underline seems wrong.

martinpitt:
I ticked off most issues in my [previous UI review](https://github.com/cockpit-project/cockpit/pull/16091#issuecomment-893574457), now only the "press enter in 'Filters' input line" issue is left. 

There's one more issue:

 - [ ] In narrow/mobile mode, the "filters" toggle is missing the labels:

image

marusak:
> now only the "press enter in 'Filters' input line" issue is left.

This will be fixed by PF, see https://github.com/patternfly/patternfly-react/pull/6145
I propose not to block on this, it will start working soon-ish.

marusak:
> In narrow/mobile mode, the "filters" toggle is missing the labels:

Hmm, patterfly has this rule which hides those labels. I'll look into PF and see why this was  introduced, seems like it was done on purpose.
```
.pf-c-toolbar__expandable-content .pf-m-label {
    display: none;
    visibility: hidden;
}
```

martinpitt:
Thanks @marusak : I don't think they are crucial, no veto from me -- I just wanted to make sure it's not an unintended bug.

marusak:
> Hmm, patterfly has this rule which hides those labels. I'll look into PF and see why this was introduced, seems like it was done on purpose.

Opened https://github.com/patternfly/patternfly/issues/4284

I would say that we don't necessarily need to have workaround for this. @KKoukiou WDYT?

KKoukiou:
> > Hmm, patterfly has this rule which hides those labels. I'll look into PF and see why this was introduced, seems like it was done on purpose.
> 
> Opened [patternfly/patternfly#4284](https://github.com/patternfly/patternfly/issues/4284)
> 
> I would say that we don't necessarily need to have workaround for this. @KKoukiou WDYT?

I don't think think we need a workaround for it. If PF decides to not implement the requested change, we can always use placeholder select entries in the selects instead of labels.

PR #16096, opened 2021-07-14
Fix current RHEL 9 round of coverity complaints

martinpitt:


src/websocket/test-websocket.c
```diff
@@ -1284,9 +1284,9 @@ main (int argc,
   };
 
   signal (SIGPIPE, SIG_IGN);
-  g_setenv ("GSETTINGS_BACKEND", "memory", TRUE);
-  g_setenv ("GIO_USE_PROXY_RESOLVER", "dummy", TRUE);
-  g_setenv ("GIO_USE_VFS", "local", TRUE);
+  g_assert (g_setenv ("GSETTINGS_BACKEND", "memory", TRUE));
```
martinpitt, line 1287: FTR, I know that this pattern is frowned upon -- but this is an unit test only, where compiling without assertions makes no sense.

marusak (approved):
Thanks!

PR #16097, opened 2021-07-14
storaged: Port from moment to timeformat

### LINKED ISSUE #15979: remove Moment.js

martinpitt:
This was the last page to be converted. Drop moment npm dependency, it's
deprecated and we don't want to grow new usage of it.

Fixes #15979

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "js-sha1": "0.6.0",
     "js-sha256": "0.9.0",
     "json-stable-stringify-without-jsonify": "1.0.1",
-    "moment": "2.29.1",
     "mustache": "4.2.0",
     "patternfly": "3.59.5",
     "patternfly-react": "2.39.17",
--- a/pkg/storaged/crypto-tab.jsx
+++ b/pkg/storaged/crypto-tab.jsx
@@ -25,7 +25,6 @@ import {
     Flex, FlexItem,
 } from "@patternfly/react-core";
 import cockpit from "cockpit";
-import moment from "moment";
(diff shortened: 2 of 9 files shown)


martinpitt:
I'm confused by this [webpack-jumpstart failure](https://github.com/cockpit-project/cockpit/pull/16097/checks?check_run_id=3066615376):
```
*** node_modules ec5d06198a2ce17b9826ca829b439b6337ea1c73 doesn't match our package.json
```

Well yes, that's why this PR changes node_modules from that SHA to 970222f9da to pick up the rebuild..

KKoukiou (approved):
Heh you made it all

PR #16102, opened 2021-07-15
common: Add Content-Type for wasm

### LINKED ISSUE #16099: WebAssembly streaming compilation doesn't work due to missing Content-Type HTTP Header for .wasm files
Cockpit version: 215-1
OS: Ubuntu 20.04.2 LTS
Page: n/a (new module)

Cockpit doesn't set `Content-Type` HTTP Response Header (MIME type, Media Type) for `.wasm` files (WebAssembly binary modules). For that reason WebAssembly streaming compilation doesn't work and falls back to not optimal "ArrayBuffer instantiation". MIME type should be `application/wasm` in order to allow effective streaming compilation of WebAssembly modules.

Regarding my particular case, I'm getting next warning in Chrome Debug Console from my Framework, when I run my module under Cockpit:
> Streaming compilation failed. Falling back to ArrayBuffer instantiation.  TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.

Additionally I've tested my module under `nginx/1.21.1` and found that it adds `Content-Type: application/wasm` Header for `.wasm` files by default.

Regarding `application/wasm` MIME (Media) type:
1. `application/wasm` is recommended MIME Type for files containing WebAssembly modules in binary format (`.wasm` files) according WebAssembly Specification ([https://www.w3.org/TR/wasm-core/](https://www.w3.org/TR/wasm-core/)).
2. `application/wasm` MIME type is required for effective WebAssembly streaming compilation according WebAssembly Specification ([https://www.w3.org/TR/wasm-web-api/](https://www.w3.org/TR/wasm-web-api/)). Here's simpler description with examples: [https://developers.google.com/web/updates/2018/04/loading-wasm](https://developers.google.com/web/updates/2018/04/loading-wasm).
3. `application/wasm` is officially registered MIME type ([https://www.iana.org/assignments/media-types/media-types.xhtml](https://www.iana.org/assignments/media-types/media-types.xhtml)).
4. `application/wasm` MIME type already has been added to a very popular NGINX web server ([https://trac.nginx.org/nginx/ticket/1606](https://trac.nginx.org/nginx/ticket/1606)).

### Steps to reproduce

Execute some code with `WebAssembly.instantiateStreaming` in a Cockpit Module, example can be found here: [https://developers.google.com/web/updates/2018/04/loading-wasm](https://developers.google.com/web/updates/2018/04/loading-wasm).

Additionally I could provide a sample Cockpit Module if it would be required.

martinpitt:
Fixes #16099


marusak:
By any chance could we have unit test for this?

martinpitt:
We could -- I was about to, I looked at test_content_type() in src/common/test-webresponse.c. That checks one (text/html) case in that declarative list, and none of the others are covered. As they all work the same, I found it not worth the effort of copying the test at first. But as you asked as well, I can look into it.

martinpitt:
I tried with [something like this](https://gist.github.com/martinpitt/7cab3f57f66cda8d145634c2ba481906), but that doesn't work -- the unit test uses the actual files in pkg/shell, and we don't have any *.wasm files in our tree. I can easily extend it to at least check the content-type for *.png and *.js, though.

martinpitt:
Test case added.

martinpitt:
@marusak: PTAL?

marusak (approved):
Danke!

PR #16104, opened 2021-07-15
ws: Avoid running ws+browser as root in cockpit-desktop

### LINKED ISSUE #15862: cockpit-desktop: Don't run the browser as root
If cockpit-desktop is invoked as the root user, having a root cockpit desktop inside the browser is fine. But it should not start the browser itself (webkit, firefox, etc.)  as root. Instead, detect this and run the browser as an unprivileged user (perhaps `adm`?) with a temporary home directory.

martinpitt:
When running cockpit-desktop as root, getting a root user cockpit
session inside the browser is fine, and is usually desired.

However, we don't need to run the browser and cockpit-ws themselves as
root for that -- the desktop team actively discourages running GTK
programs as root.

Fixes #15862

---

 - [ ] Fix access to /tmp/.X11-unix/ in container
 - [ ] Test this in an actual Fedora desktop

### DIFF
--- a/src/ws/cockpit-desktop.in
+++ b/src/ws/cockpit-desktop.in
@@ -198,4 +198,10 @@ fi
 
 HOME="$BROWSER_HOME" '$BROWSER' http://127.0.0.90'"$URL_PATH"'
 '
-unshare --user --map-root-user --net /bin/bash -c "$SCRIPT" <&${COPROC[0]} >&${COPROC[1]}
+
+# avoid running ws and browser as root, drop to bin user if available
+if [ $(id -u) -eq 0 ] && id bin >/dev/null 2>/dev/null; then
+    DROPPRIVS='setpriv --reuid=bin --regid=bin --clear-groups --no-new-privs --bounding-set=-all'
+fi
+
+${DROPPRIVS:-} unshare --user --map-root-user --net /bin/bash -c "$SCRIPT" <&${COPROC[0]} >&${COPROC[1]}
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -769,6 +769,28 @@ G_MESSAGES_DEBUG=all XDG_CONFIG_DIRS=/usr/local %s -p 9999 -a 127.0.0.90 --local
                 # should clean up processes
                 self.assertEqual(m.execute("! pgrep -a cockpit-ws && ! pgrep -a cockpit-bridge"), "")
 
+        # browser runs as the same user when calling as user
+        m.write("/tmp/browser.sh", """#!/bin/sh -e
+                                      echo $$ > /tmp/browser-pid.txt
+                                      while [ -e /tmp/browser-pid.txt ]; do sleep 0.1; done""")
+        m.execute("chmod 755 /tmp/browser.sh")
+        m.spawn("su - -c 'BROWSER=/tmp/browser.sh %s system' admin" % cockpit_desktop, "cockpit-desktop-user.log")
+        ugid = m.execute("""until [ -e /tmp/browser-pid.txt ]; do sleep 0.1; done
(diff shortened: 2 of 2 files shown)


martinpitt:
I developed this against the integration test with the mocked browser, and made sure that on my own laptop OS (with firefox, no webkit installed) running cockpit-desktop as my user still works as expected. I can't run it as root through sudo, as that process can't access my wayland or X.org server. This is something to watch out for when dropping privileges.

I now tested this on a clean Fedora 34 desktop live system. I installed cockpit-{ws,system}, then `/usr/libexec/cockpit-desktop /` works as before and gives me a nice webkit UI.  I set a root password, logged in as root, and first confirmed that cockpit-desktop works as root (webkit browser runs as root). This updated version indeed fails with "cannot open display: :0", so this needs to be fixed to plumb through XAUTHORITY and/or the ACLs.

This can be reproduced in a simpler manner with `setpriv --reuid=bin --regid=bin --clear-groups gnome-calculator`. @mcatanzaro , do you have a suggestion what do do there? Something along the lines of this:

```sh
cp $AUTHORITY /tmp/xauth
chown bin:bin /tmp/xauth
setpriv --reuid=bin --regid=bin --clear-groups env XAUTHORITY=/tmp/xauth gnome-calculator
```
Of course the real script wouldn't use /tmp/xauth but copy that into the temporary home dir. But this at least seems to work -- is that kosher?

mcatanzaro:
> Of course the real script wouldn't use /tmp/xauth but copy that into the temporary home dir. But this at least seems to work -- is that kosher?

I have no clue tbh, sorry. Trying to avoid running as root is laudable, but I've never seen an attempt to do this within a root session before.

mcatanzaro:
> Trying to avoid running as root is laudable

Also I'm a little surprised that WebKit works at all when run as root. Do you have the web process sandbox enabled? You should enable that with webkit_web_context_set_sandbox_enabled() (which should become default in the future). It will probably crash immediately if run as root.

martinpitt:
@mcatanzaro : Yes, cockpit-desktop [does enable sandboxing](https://github.com/cockpit-project/cockpit/blob/main/src/ws/cockpit-desktop.in#L56). `ps` confirms that the process is running as root, not sure if it perhaps drops some capabilities, seccomp, or similar.

martinpitt:
I'm still fighting with getting this to work through `unshare --net`.

This works (after the above preparation of /tmp/xauth):
```
setpriv --reuid=bin --regid=bin --clear-groups --no-new-privs --bounding-set=-all --reset-env /bin/bash -ec 'export DISPLAY=:0; export XAUTHORITY=/tmp/xauth; gnome-calculator'
setpriv --reuid=bin --regid=bin --clear-groups --no-new-privs --bounding-set=-all --reset-env unshare --user --map-root-user /bin/bash -ec 'export DISPLAY=:0; export XAUTHORITY=/tmp/xauth; gnome-calculator'
```
But once you isolate the network namespace, it gives "connection refused":
```
setpriv --reuid=bin --regid=bin --clear-groups --no-new-privs --bounding-set=-all --reset-env unshare --user --map-root-user --net /bin/bash -ec 'export DISPLAY=:0; export XAUTHORITY=/tmp/xauth; gnome-calculator'
```
Presumably because it can't access the socket in /tmp/.X11-unix/ any more? But network isolation is the central measure by how cockpit-desktop works, so this is not negotiable. Curiously, just `unshare --user --map-root-user --net xwininfo` does work, so it has some other way of talking to X.org.

I admittedly don't understand the X.org protocol/authentication very well, so this will need some further research.

martinpitt:
This is super hard and stale. Moving to my "someday/maybe" TODO list instead.

Issue #14491, opened 2020-08-19
UX: Typeahead entry clears when dropdown dismissed


garrett:
Duplicating https://github.com/patternfly/patternfly/issues/3378 here, as I think we should have a local fix until PF upstream fixes it and we pull the upstream fix back in.

> Typing (including pasting with the keyboard) works. However, if an entry is not selected either with the mouse or via the keyboard (by pressing down and hitting enter, for example), then the entry field is cleared.
>
> When someone is working on the rest of the form and has submitted it, it is easy to overlook a form they've already entered (and didn't expect to get cleared).
>
> Cockpit BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1866225

> ## How to trigger this issue (on PF4's demo)
> 
> 1. Copy this string: **Florida**
> 2. Visit https://www.patternfly.org/v4/documentation/react/components/select#typeahead
> 3. Enter the typeahead with a click (or by tabbing to it)
> 4. Paste (or type) the string Florida
> 5. Important: **_Do not_** select the dropdown's "Florida" entry. Do not click it or use the keyboard to select it.
> 6. Click elsewhere or tab away to another widget to dismiss the dropdown
> 7. Notice how the text that was entered via the keyboard is gone!
> 
> ## Expected results
> 
> Any text that was typed or pasted into the entry should persist. It should not be cleared.

> Note: This is also an accessibility issue, as tabbing away from the entry also clears it. For someone relying on a screen reader, this means there's probably no way to know that the entry was changed from what they entered.

As mentioned in https://github.com/cockpit-project/cockpit/pull/14475#pullrequestreview-468274246 & https://github.com/cockpit-project/cockpit/pull/14475#pullrequestreview-469458753

martinpitt:
Still current as of today. Just the URL is now https://www.patternfly.org/v4/components/select

garrett:
This is being worked on @ https://github.com/patternfly/patternfly-design/issues/1003

When the PatternFly issue is resolved and it lands, we can revisit it here and adjust accordingly. (It may require us to do something on our end too, if it's not handled by default.)

PR #16105, opened 2021-07-15
lib: Retaining keyboard-entered value in file autocompletion typeahead text field when focus leaves input away

KKoukiou:
This will allow us to fix https://bugzilla.redhat.com/show_bug.cgi?id=1866225

Fixes https://github.com/cockpit-project/cockpit/issues/14491

See more info: https://github.com/patternfly/patternfly-react/issues/5502


martinpitt (approved):
Does what it says on the tin. Danke!

PR #16106, opened 2021-07-15
test: Don't delete pixeldiff source code through attach()

martinpitt:
Normally, attach() is being called for dynamically generated files, and
it is (moderately) desirable to move them into the attachment directory
instead of copying them. But this must not happen for static source code
files, otherwise the second call will fail bitterly if two tests run in
parallel. This currently breaks cockpit-podman a lot.

Switch attach() to copy files by default, which is the safe thing to do.
Add a `move` flag to opt into the optimization, and use it for
known-safe dynamically generated files like screenshots and journals. In
particular, this stops moving common/pixeldiff.html and similar git
tracked files out of the source tree.

----

I am not actually sure how we didn't notice that before -- right now it's absolutely impossible for me to get https://github.com/cockpit-project/cockpit-podman/pull/737 to green, as it keeps failing on this bug. Examples:

  - https://logs.cockpit-project.org/logs/pull-737-20210715-155154-b9c1ea35-ubuntu-stable/log
  - https://logs.cockpit-project.org/logs/pull-737-20210715-155154-b9c1ea35-rhel-9-0/log
  - https://logs.cockpit-project.org/logs/pull-737-20210715-151741-7b19e782-rhel-8-5/log

Note that these don't appear as failures in the HTML log, just as incomplete test, as the failure happens so early. We might just have discarded these as "infrastructure quirk" and retried?

It could of course also be that I misunderstood the reason for that crash -- but in any case, I think this fix (or just simply always copying instead of moving) is the right thing to do.

 - [x] change to opt-into move():
 - [x] test with a failing pixeldiff: [log](https://logs.cockpit-project.org/logs/pull-16106-20210716-045708-0d290599-fedora-34/log.html)

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -91,12 +91,20 @@ opts.jobs = 1
 opts.fetch = True
 
 
-def attach(filename):
+def attach(filename, move=False):
+    '''Put a file into the attachments directory
+
+    By default the file gets copied. You can set move=True for dynamically generated
+    files which are not touched by parallel tests.
+    '''
     if not opts.attachments:
         return
     dest = os.path.join(opts.attachments, os.path.basename(filename))
     if os.path.exists(filename) and not os.path.exists(dest):
-        shutil.move(filename, dest)
+        if move:
+            shutil.move(filename, dest)
+        else:
+            shutil.copy(filename, dest)
 
 
 class Browser:
@@ -645,7 +653,7 @@ class Browser:
             if "data" in ret:
                 with open(filename, 'wb') as f:
                     f.write(base64.standard_b64decode(ret["data"]))
-                attach(filename)
+                attach(filename, move=True)
                 print("Wrote screenshot to " + filename)
             else:
                 print("Screenshot not available")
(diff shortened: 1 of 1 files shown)


martinpitt:
I think I changed my mind -- let's make the safe behaviour (copy) the default, and only move known-big ones like core dumps and journals. Opt into the unsafe (but efficient) behavior.

I should also test this by breaking a pixel test intentionally, to make double-sure that attachments still work.

martinpitt:
All green ([example](https://logs.cockpit-project.org/logs/pull-16106-20210715-162340-736343cd-fedora-34/log.html)). I updated the API to do the safe thing by default, and added an extra commit which deliberately breaks a pixel test, to prove that attachments still work.

martinpitt:
[Failed pixel diff test](https://logs.cockpit-project.org/logs/pull-16106-20210716-045708-0d290599-fedora-34/log.html) succeeded, the [results dir](https://logs.cockpit-project.org/logs/pull-16106-20210716-045708-0d290599-fedora-34/) also has some other attachments. So this all looks good. Pushing the real thing now.

marusak (approved):
Makes sense, thanks!

PR #16109, opened 2021-07-16
test: Adjust to new systemd capitalized description

marusak:
See https://github.com/systemd/systemd/commit/0a59216

### DIFF
--- a/test/verify/check-storage-nfs
+++ b/test/verify/check-storage-nfs
@@ -215,7 +215,7 @@ class TestStorageNfs(StorageCase):
         b.click("#detail-header button:contains(Unmount)")
         self.dialog_wait_open()
         b.wait_in_text("#dialog", "The filesystem is in use")
-        b.wait_in_text("#dialog", "of user root")
+        self.assertRegex(b.text(".units-table td:first-child"), r"Session \d+ of [uU]ser root")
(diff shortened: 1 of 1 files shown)


test/verify/check-storage-nfs
```diff
@@ -215,7 +215,8 @@ class TestStorageNfs(StorageCase):
         b.click("#detail-header button:contains(Unmount)")
         self.dialog_wait_open()
         b.wait_in_text("#dialog", "The filesystem is in use")
-        b.wait_in_text("#dialog", "of user root")
+        desc = b.text(".units-table td:first-child")
+        self.assertTrue(re.match(r"Session \d+ of [uU]ser root", desc))
```
martinpitt, line 219: Please no assertTrue() -- failures are really useless. Use [assertRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex) instead. Unfortunately that doesn't accept flags such as `re.I`, so the string is fine I think.

martinpitt (changes requested):

test/verify/check-storage-nfs
marusak: Updated

martinpitt (approved):
Danke!

PR #16110, opened 2021-07-16
metrics: Adjust history metrics graph labels to avoid overlapping

### LINKED ISSUE #16016: metrics: Layout issues with metrics labels
On the metrics page, there are various issues in different languages.

In German, for example, the text overlaps poorly:

Screenshot from 2021-06-29 13-59-59

Same for Turkish...
Screenshot from 2021-06-29 14-06-21

And Russian...
Screenshot from 2021-06-29 14-04-10

And so on...

---

In English, the I/O wraps:

Screenshot from 2021-06-29 14-03-06

martinpitt:
metrics: Adjust history metrics graph labels to avoid overlapping

 * Increase the width of half-columns from 5% to 8%, as otherwise there
   is not enough space for the label even in English on wide pages. It
   does not hurt to use the space that we have to provide a little more
   detail on the half-column then.

 * Reduce the top label width for narrow browsers in two steps, as
   otherwise they begin to overlap.

 * Use absolute font sizes instead of relative ones, as otherwise the
   sublabels become too small (as they are relative to the parent label,
   not to the global default), and we can also assume that the PF
   standard sizes are sane.

Addresses half of #16016; very narrow sizes/mobile mode still needs to
be fixed (probably through rotating the labels).

### English, 1920p wide:

main:
main-en-1920

This PR:
image

### English, 960p wide:
main:
main-en-960

This PR:
image

### German, 1920p wide:
main:
main-de-1920

This PR:
image

### German, 960p wide:
main:
main-de-960

image

### Mobile

main:

Screen Shot 2021-07-16 at 15 14 25

This PR: A tad better, but still broken. See below:

Screen Shot 2021-07-16 at 15 32 48

Old version of PR with minimum width (regresses mobile mode):
Screen Shot 2021-07-16 at 15 13 46

### DIFF
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -232,7 +232,7 @@
 
 .metrics {
     --column-size: 10vw;
-    --half-column-size: 5vw;
+    --half-column-size: 8vw;
     --data-min-height: 5px;
 
     &-heading,
@@ -265,7 +265,7 @@
     }
 
     &-sublabels {
-        font-size: 65%;
+        font-size: var(--pf-global--FontSize--xs);
         color: rgba(0,0,0,0.7);
         display: flex;
     }
@@ -473,6 +473,18 @@
 
     .pf-c-card__body:first-child {
         padding-top: 0;
+    }
+}
+
+// Graph column labels are the widest part; scale down the labels for narrow widths
+@media (min-width: 800px) and (max-width: 1200px) {
+    .metrics-label-graph {
+        font-size: var(--pf-global--FontSize--s);
+    }
+}
+@media (max-width: 800px) {
+    .metrics-label-graph {
+        font-size: var(--pf-global--FontSize--xs);
     }
 }
 


martinpitt:
So this clearly helps "rather narrow" to "wide" screens, but breaks mobile mode due to the minimum width. For that/very narrow windows, I think we need to let go of the minimum width again, and instead use [vertical text](https://developer.mozilla.org/en-US/docs/Web/CSS/text-orientation) for the labels.

With this first approach:
```diff
--- pkg/metrics/metrics.scss
+++ pkg/metrics/metrics.scss
@@ -231,8 +231,8 @@
 }
 
 .metrics {
-    --column-size: max(8ex, 10vw);
-    --half-column-size: max(8ex, 8vw);
+    --column-size: 10vw;
+    --half-column-size: 8vw;
     --data-min-height: 5px;
 
     &-heading,
@@ -484,7 +484,9 @@
 }
 @media (max-width: 800px) {
     .metrics-label-graph {
-        font-size: var(--pf-global--FontSize--xs);
+        font-size: var(--pf-global--FontSize--s);
+        writing-mode: vertical-rl;
+        text-orientation: mixed;
     }
 }
```

.. the general layout is free of overlaps, but the sublabel positions and and alignments are all wrong:

image

So I backed out the minimum width part and updated the screenshots. Like that it's an incremental improvement and avoids the regression. We can then look into how the labels should look in vertical mode separately.

garrett:
@martinpitt: We had similar issues on the website's download page's distro table @ https://cockpit-project.org/running.html and I solved it by rotating the text with CSS (`transform: rotate(-30deg);` with a small page width).

martinpitt:
@garrett : I'm happy to play around with the angled text on Thursday. Marking as "ready for review" as I think this is already a big improvement without regressions, so we may want to land it for tomorrow's release?

garrett:
When there are two labels to indicate different columns, they don't match up at all, and it's hard to understand what they mean, if you don't already know.

garrett (approved):
LGTM :+1: 

I like the vertical idea; we can work on that as a follow-up. What's in the PR right now is an improvement on sizing.

PR #16112, opened 2021-07-17
spec: Fix syntax error in PAM grep

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=1983077

----

Validated  with

    test/image-prepare -qB fedora-34
    rpm -q --scripts -p tmp/build-results/cockpit-ws-*.rpm

Now it is correct:
```
# check for deprecated PAM config
if grep --color=auto pam_cockpit_cert /etc/pam.d/cockpit; then
```


marusak (approved):
Danke!

PR #16114, opened 2021-07-17
tools: Add --rsync option to webpack-{watch,make}

martinpitt:
This allows developers to rapidly iterate on code that gets tested in
our standard VMs without having to manually copy the code. This
minimizes the cycle to "save" -- "wait a bit" -- "press reload in the
browser".

webpack-watch has no actual option parsing, so relax the positional
argument checking and pass on any additional arguments to webpack-make.
This requires to put the options *after* the page name, but we'll get
used to that quickly.

----

As my laptop doesn't have libvirt, setroubleshoot etc. installed, I spend the majority of my development iteration with manually calling `scp -r` after each successful build. This gets rid of this annoyance, especially as I occasionally forget the scp and then get confused why my changes don't apply. It's not exactly a pretty implementation in webpack-make, but it gets the job done.. :man_shrugging: 

WDYT?

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -178,6 +178,15 @@ as
 
 Then reload cockpit in your browser after building the page.
 
+Often you need to test your code in a VM. `webpack-make` has an `-r`/`--rsync`
+option for copying the built webpack into the given SSH target's
+/usr/share/cockpit/ directory. If you use cockpit's own test VMs and set up the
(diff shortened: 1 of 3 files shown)


KKoukiou (approved):
Makes sense.

martinpitt:
@allisonkarlitskaya, @marusak: Do you use a workflow like this? Apparently Katerina doesn't really. Does it make sense to you?

PR #16119, opened 2021-07-20
webpack-jumpstart.yml: add separate "head" build

allisonkarlitskaya:
webpack-jumpstart has been telling a lie for the past while: we've been
building the "merge" commit but labelling it in the cache with the
"head" sha.  This occasionally results in some errors about
inconsistencies, particularly in the case where package-lock.json has
changed, and is preventing us from introducing more robust consistency
checking in #16005.

The reason we do this is in large part a crutch to get around the fact
that packit only knows how to test the head commit.  Our current "lies"
allow us to get some kind of "partial merge" into packit for testing.

Meanwhile, however, packit is working on switching over to building the
merge commit instead of the head commit, and it seems like this will not
be configurable.  When that happens, our current setup will break,
because packit will go looking in our cache based on the merge commit,
but it won't find it.

Let's stop labelling merge commit build results with the head commit
sha.  Instead, we'll label them with the correct sha: the sha of the
merge commit.  That way, when packit is changed, it will continue to
work for us.

We also add an additional build step to explicitly build the "head"
commit and correctly label it as the head commit.  This temporary
measure is gated by an `if:` to not occur in `push` events, but only in
pull requests (where we always have separate merge and head commits,
even if the branch is fully rebased).  This will allow packit to
continue to work with its current practice of checking out the head
commit, and will allow for a smooth transition when they flip the switch
over to using the merge commit.  These extra steps can be removed at
that point.

 - test run for push to main: https://github.com/allisoninc/cockpit/actions/runs/1048186668
 - test PR: https://github.com/allisoninc/cockpit/pull/2
 - test run from PR: https://github.com/allisoninc/cockpit/actions/runs/1048187526

### DIFF
--- a/.github/workflows/webpack-jumpstart.yml
+++ b/.github/workflows/webpack-jumpstart.yml
@@ -34,8 +34,41 @@ jobs:
           path: webpack-jumpstart.tar
           retention-days: 1
 
+  build-head:
+    if: ${{ github.event.pull_request.merge_commit_sha != github.event.pull_request.head.sha }}
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: none
+    timeout-minutes: 20
+    env:
+      HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+    steps:
+      - name: Build jumpstart tarball
+        run: |
+          id="$(docker container create -u node -w /home/node node /bin/sh -c \
+            'git init cockpit &&
+             cd cockpit &&
+             git fetch --depth 1 https://github.com/cockpit-project/cockpit "'"${HEAD_SHA}"'":build &&
+             git checkout build &&
+             pkg/build webpack-jumpstart.tar -j$(nproc)'
+          )"
+
+          docker container start --attach "${id}" >&2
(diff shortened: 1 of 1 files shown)


martinpitt:
Summary from IRC: I'm not sure it makes sense to build the "merge" branch, as that SHA is mostly not reproducible.  The *identification* of what you want to build/test makes sense to be the HEAD of the proposed branch -- the merge SHA is rather useless for that. That *may* require concatenation with the SHA that it was built against, but honestly that doesn't make much difference in practice -- both a local dev who wants to test a branch, as well as CI would only ever use the "tag-HEAD" build anyway. This piece of non-well-definedness needs to live somewhere, as that's something that we explicitly want (be able to develop and land more than one PR in parallel).

So I think the thing to fix once packit switches over from refs/N/head to refs/N/merge is to use some git query to figure out the original HEAD of the merged branch, instead of literally `git rev-parse HEAD` (that's the bit that breaks once packit checks out the merge commit). Something along the lines of "search for download HEAD if HEAD is an actual commit, and the first commit's HEAD from the merged branch if the topmost commit is a merge". I.e. what you see in `git log` as the first commit after the merge.

PR #16120, opened 2021-07-20
test: Test sudo on cockpit machine in check-system-s4u-ssh

martinpitt:
This is the more common case, without a remote ssh host in between. Not
too many surprises, just that this uncovers that the current approach of
setting up the delegation rule does not scale. Filed as
https://pagure.io/freeipa/issue/8927


Gundersanne (approved):
Nice :) Don't think you can get much closer to cockpit's usecase :D

PR #16124, opened 2021-07-21
Replace broken links with linux.die.net links

jpall12:
Update broken `pcpstat` and `pcpinfo` links with working linxu.die.net links. On PCP Metrics page

If you feel that the updated links are not the best choice feel free to edit or drop them below.


martinpitt (approved):
Thank you!

PR #16130, opened 2021-07-22
test: Drop unnecessary kinit admin@COCKPIT.LAN in check-system-s4u-ssh

martinpitt:
Creating the impersonating S4U ticket for the user will eventually
happen in cockpit-session. This must not require the admin
password/kinit. Conceptually, it is supposed to only use the cockpit
krb5.keytab.

Drop the call, it works fine without it.


martinpitt:
Pushed a second commit that gets rid of another awkward thing which we don't want to use in cockpit-session (calling `kinit`).

Gundersanne (approved):
Nice :)

PR #16131, opened 2021-07-22
cleanup: stop disting the guide

allisonkarlitskaya:
- stop doing the versions dance when updating the website on release
 - stop including the built guide in the tarball

 * [ ] clean up the website itself for our new plan of always having only `latest/`


allisonkarlitskaya:
```
-rw-r--r--. 1 8845548 main/cockpit-249.4.g2473e8954.tar.xz
-rw-r--r--. 1 8719628 stop-disting-the-guide/cockpit-249.9.g4d1cf648f.tar.xz
```
~125kB less, ~1.5%.  That's nice I guess?

tools/release-guide
```diff
@@ -89,16 +89,23 @@ prepare()
 
     cd ..
 
+    trace "Building the docs"
+    mkdir build
+    cd build
+    "../extract/${subdir}/configure"
+    make "$DIRECTORY/html"
+    cd ..
```
martinpitt, line 97: This feels wasteful -- we already build git during the release process, and thus at this point cwd is a configured tree (as at least `release-source` calls `make dist`). Are you concerned about being able to do these steps separately? If so, we need to add a doc build to cockpit's distcheck, to make sure it works out of a tarball, instead of just a git clone (doing that is a good idea anyway, of course)

martinpitt (commented):
This looks great to me, thanks! One comment.

tools/release-guide
martinpitt: As per IRC discussion, testing this `make "$DIRECTORY/html"` during unit tests can be a follow-up.

martinpitt (approved):

PR #16135, opened 2021-07-22
metrics: Rotate labels  in small sizes

garrett:
This is a challenge. I have something that rotates the text. It works in English. Switch to German, and... well... "Auslagerungsspeicher" breaks everything.

### DIFF
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -486,8 +486,65 @@
     }
 }
 @media (max-width: 800px) {
+    // Drop extra padding in mobile sizes (graphs need all the available space)
+    main > .pf-c-page__main-section {
+        padding-left: 0;
+        padding-right: 0;
+
+        &:not(:first-child) {
+            padding-bottom: 1rem;
+        }
+
+        &:last-child {
+            padding-top: 0;
+            // Fix stacking order for stickiness
+            z-index: 999;
+        }
+    }
+
+    // Rotate labels
     .metrics-label-graph {
-        font-size: var(--pf-global--FontSize--xs);
+        // Use an in-between font size (half step from xs to s)
+        font-size: calc(var(--pf-global--FontSize--s) + var(--pf-global--FontSize--xs) / 2);
+        display: grid;
+        grid-template: 1fr / 1fr;
+
+        > span:not(.metrics-sublabels) {
+            // "Outdent" the main label by indenting with a negative medium spacer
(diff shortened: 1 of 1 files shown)


martinpitt:
I just [updated the translation of that](https://translate.fedoraproject.org/translate/cockpit/main/de/?checksum=4424e50f14a31a34) back to "Swap". The UI uses that original word in a couple of places, and it's by and large a technical proper noun in German anyway.

garrett:
In English:

Screen Shot 2021-07-22 at 17 03 15

Even The space-constrained SE size works OK:

Screen Shot 2021-07-22 at 17 04 35

In German...


Screen Shot 2021-07-22 at 17 05 47

Screen Shot 2021-07-22 at 17 05 54

Screen Shot 2021-07-22 at 17 05 26

garrett:
Oh, I wonder if https://developer.mozilla.org/en-US/docs/Web/CSS/transform-box might help.

KKoukiou (commented):
Please add comments in the CSS

garrett:
Added comments, rebased, force-pushed

martinpitt:
I just re-discovered this -- sorry, it slipped through the cracks across my holidays. I just tested this again, and I'm on the fence -- the rotated labels certainly help width-wise, and avoid painting into the right off-screen side. But the overflow into the previous card feels like a regression:

image

I feel that the status quo of having a small font which overlaps is the lesser evil:

image

To fix this, we need to increase the height of the table header so that it fits the rotated word -- I'm not sure if CSS is strong enough for that?

garrett:
> we need to increase the height of the table header so that it fits the rotated word

CSS can, but rotating moves it out of the precomputed space for the label, so then at that point, nothing (HTML, CSS, JS) knows about the space it takes. I was looking for a solution or an alternate approach for that but didn't find when I was working on this. 

Meanwhile, I went on vacation and have had a bunch of other stuff to do.

We'll get this working one way or another sooner than later.

garrett:
Rebased; no conflicts. I guess it probably still works for some definition of "works". There's a reason why this one's a draft. (It's a bit buggy and mainly an experiment.)

KKoukiou:
There is an open issue to track this and this is a very subiotimal solution, on the edge of being a regression. I prefer closing this, keeping the upstream issue for tracking, https://github.com/cockpit-project/cockpit/issues/16016, and reattempting the solution when there is capacity.

PR #16144, opened 2021-07-23
lib: Move po conversion scripts to lib

martinpitt:
starter-kit and friends currently contain copies of these. Let's move
them to pkg/lib so that they can be shared, and evolve along with
cockpit-po-plugin.

Moving to pkg/lib triggers test-static-code's "invalid translatable=
attribute" check. Drop the "Glade compatible form" comment to avoid
this, we haven't used this form in years and don't want to promote it
any more.


martinpitt:
The [ubuntu-stable](https://logs.cockpit-project.org/logs/pull-16144-20210723-070617-49bf02fb-ubuntu-stable/log.html) flake was fixed by https://github.com/cockpit-project/bots/pull/2234 . Re-pushing to reset test stats. This is my guinea pig for "land stuff without retries"

martinpitt:
Argh, my previous bot commit was right, but it didn't match the current common flake. Sent https://github.com/cockpit-project/bots/pull/2235

allisonkarlitskaya (approved):
Ya,  that's fine...

Are we really completely unstable to do with with xgettext or some kind of pre-existing code that we could get from npm, though?

martinpitt:
@allisonkarlitskaya : That requires some research. xgettext can't read JSON (manifests) nor HTML/XML, and either way it's doing some rather elaborate parsing. It also needs to have the knowledge which particular manifest keys/HTML nodes/attributes to translate. So presumably we'll need to keep some kind of tool around, but of course it's entirely possible that we can simplify them.

PR #16149, opened 2021-07-26
ws: Move to session specific kerberos keyring

martinpitt:
Storing delegated kerberos tickets into the default (user-)global
default keyring is a bit problematic: It changes the available
credentials to all other currently running sessions of that user, and
prevents being able to clean up the ccache after the session ends. The
global default keyring is really aimed at users doing `kinit` explicitly
(or the initial desktop session login).

Move to what SSH does [1], create a session specific file ccache in the
user's XDG runtime directory, and export it through `$KRB5CCNAME`.
Delete the ccache at session end. logind will clean up the runtime
directory for us after a while as well, which is a nice second line of
defense.

The user can always copy the ticket somewhere else while the session is
running, so this does not need to be a 100% robust enforced measure. But
it may help to avoid leaking secrets and keeping them around
unnecessarily long, in particular as neither kernel KEYRING: nor KCM:
caches get cleaned up automatically when PAM sessions end (in particular
not by `PAM_DELETE_CRED`).

Defining the ccache location for ourselves also drops the last direct
krb5 API usage, cockpit-session now only uses the plain GSSAPI.

[1] https://github.com/openssh/openssh-portable/blob/57ed647ee07b/auth-krb5.c#L72

-----

While I was working on the S4U delegation stuff (for sudo/ssh with certificate authentication) I was rather surprised that this cleanup does not currently happen. I always had assumed that this was the very purpose of [pam_setcred()](https://linux.die.net/man/3/pam_setcred), as it explicitly advertises kerberos tickets.

But I am not sure if this is really a good idea, or if it could interfere with something else like parallel ssh sessions. Right now, starting a cockpit session (through GSSAPI auth) replaces an existing ticket in the default keyring, also for existing sessions (as the keyring is per user, not per session). With this PR, the ticket then gets removed, and the parallel running ssh session ends up having no ticket at all (and needs to `kinit` again).

Should cockpit-session perhaps not *ever* write into the default ccache, but use a classic cache file and set `$KRB5CCNAME` inside the session?

@simo5, @dpal, do you have some guidance/opinion here? Thanks!

### DIFF
--- a/src/ws/session.c
+++ b/src/ws/session.c
@@ -28,7 +28,6 @@
 #include <gssapi/gssapi.h>
 #include <gssapi/gssapi_generic.h>
 #include <gssapi/gssapi_krb5.h>
-#include <krb5/krb5.h>
 #include <fcntl.h>
 
 static char *last_txt_msg = NULL;
@@ -553,9 +552,15 @@ perform_tlscert (const char *rhost)
 }
 
 
-static void
+/* Return path of ccache file (including FILE: prefix), clean this up at session end */
+static char *
 store_krb_credentials (gss_cred_id_t creds, uid_t uid, gid_t gid)
 {
+  gss_key_value_set_desc store;
+  struct gss_key_value_element_struct element;
+  OM_uint32 major, minor;
+  char *ccache;
+
   assert (creds != GSS_C_NO_CREDENTIAL);
 
   bool was_root = getuid() == 0;
@@ -568,35 +573,35 @@ store_krb_credentials (gss_cred_id_t creds, uid_t uid, gid_t gid)
 
   assert (geteuid () == uid && getegid() == gid);
 
-  krb5_context k5;
-  int res = krb5_init_context (&k5);
-  if (res == 0)
-    {
-      gss_key_value_set_desc store;
-      struct gss_key_value_element_struct element;
-      OM_uint32 major, minor;
+  /* The ccache path needs to be unique per cockpit session; as cockpit-session runs throughout the
+   * lifetime of sessions, our pid is unique. We expect the cache to be cleaned up at the end, but
+   * if not, and the PID gets recycled, this just overwrites the old obsolete one, which is good. */
+  asprintfx (&ccache, "FILE:/run/user/%u/cockpit-session-%u.ccache", uid, getpid ());
+  debug ("storing kerberos credentials in session: %s", ccache);
 
-      store.count = 1;
-      store.elements = &element;
-      element.key = "ccache";
-      element.value = krb5_cc_default_name (k5);
+  store.count = 1;
+  store.elements = &element;
+  element.key = "ccache";
+  element.value = ccache;
 
-      debug ("storing kerberos credentials in session: %s", element.value);
-
-      major = gss_store_cred_into (&minor, creds, GSS_C_INITIATE, GSS_C_NULL_OID, 1, 1, &store, NULL, NULL);
-      if (GSS_ERROR (major))
-        warnx ("couldn't store gssapi credentials: %s", gssapi_strerror (GSS_C_NO_OID, major, minor));
-
-      krb5_free_context (k5);
-    }
-  else
-    {
-      warnx ("couldn't initialize kerberos context: %s", krb5_get_error_message (NULL, res));
-    }
-
+  major = gss_store_cred_into (&minor, creds, GSS_C_INITIATE, GSS_C_NULL_OID, 1, 1, &store, NULL, NULL);
+  if (GSS_ERROR (major))
+    warnx ("couldn't store gssapi credentials: %s", gssapi_strerror (GSS_C_NO_OID, major, minor));
 
   if (was_root && (setresuid (0, 0, 0) != 0 || setresgid (0, 0, 0) != 0))
     err (127, "Unable to restore permissions after storing gss credentials");
+
+  return ccache;
+}
+
+static void
+release_krb_credentials (char *ccache)
+{
+  /* strip off FILE: prefix for deleting */
+  assert (strncmp (ccache, "FILE:", 5) == 0);
+  if (unlink (ccache + 5) != 0)
+    warn ("couldn't clean up kerberos ticket cache %s", ccache);
+  free (ccache);
 }
 
 int
@@ -609,6 +614,7 @@ main (int argc,
   char *authorization;
   char *type = NULL;
   const char **env;
+  char *ccache = NULL;
   int status;
   int res;
   int i;
@@ -679,10 +685,21 @@ main (int argc,
   for (i = 0; env_saved[i] != NULL; i++)
     pam_putenv (pamh, env_saved[i]);
 
-  if (want_session) /* no session → no login messages → no memfd */
+  if (want_session) /* no session → no login messages or XDG_RUNTIME_DIR → no memfd or session ccache */
     {
       if (pam_putenv (pamh, "COCKPIT_LOGIN_MESSAGES_MEMFD=3") != PAM_SUCCESS)
         errx (EX, "Failed to set COCKPIT_LOGIN_MESSAGES_MEMFD=3 in PAM environment");
+
+      if (creds != GSS_C_NO_CREDENTIAL)
+        {
+          ccache = store_krb_credentials (creds, pwd->pw_uid, pwd->pw_gid);
+          char *ccache_env = NULL;
+          asprintfx (&ccache_env, "KRB5CCNAME=%s", ccache);
+
+          if (pam_putenv (pamh, ccache_env) != PAM_SUCCESS)
+            errx (EX, "Failed to set KRB5CCNAME in PAM environment");
+          free (ccache_env);
+        }
     }
 
   env = (const char **) pam_getenvlist (pamh);
@@ -711,9 +728,6 @@ main (int argc,
 
       int login_messages_fd = cockpit_json_print_finish_memfd (&login_messages);
 
-      if (creds != GSS_C_NO_CREDENTIAL)
-        store_krb_credentials (creds, pwd->pw_uid, pwd->pw_gid);
-
       const int remap_fds[] = { -1, -1, -1, login_messages_fd };
       status = spawn_and_wait (bridge_argv, env, remap_fds, 4, pwd->pw_uid, pwd->pw_gid);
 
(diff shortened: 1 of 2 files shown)


martinpitt:
While I see that this is somewhat optional for GSSAPI "single sign on" authentication, it seems to me that for delegated S4U tickets from TLS cert/smart card auth we have to clean them up once the session ends. Otherwise we'll leak credentials to sudo like crazy.

simo5:
On Mon, 2021-07-26 at 00:25 -0700, Martin Pitt wrote:
> While I see that this is somewhat optional for GSSAPI "single sign on" authentication, it seems to me that for delegated S4U tickets from TLS cert/smart card auth we have to clean them up once the session ends. Otherwise we'll leak credentials to sudo like crazy.
> 
You could create a "session" (instead of "user") keyring in kernel, or
a private FILE type ccaches for this usage. I think it is a better idea
than using default credentials caches, as with default creds there is
too much a risk of interfering with other processes (esp when logging
in as root).

HTH,
Simo.

-- 
Simo Sorce
RHEL Crypto Team
Red Hat, Inc

martinpitt:
I see that [MIT krb docs](https://web.mit.edu/kerberos/krb5-latest/doc/basic/ccache_def.html) show support for session keyring -- this seems ideal, better than putting some random file keytab somewhere (although that would still seem okay).

Now I realize that this approach here by itself is bogus -- on Fedora, the default keyring from krb5-libs is `KEYRING:persistent:<uid>`, and that is explicitly meant to survive logouts. The sssd-kcm package overwrites that to `KCM:` to use the global caching daemon, but I guess that's supposed to have a similar lifetime independent from sessions.

I have some trouble with using `KEYRING:session`, but I'll keep looking at that. `FILE:` certainly works fine.

simo5:
On Mon, 2021-07-26 at 05:01 -0700, Martin Pitt wrote:
> I see that [MIT krb docs](https://web.mit.edu/kerberos/krb5-latest/doc/basic/ccache_def.html) show support for session keyring -- this seems ideal, better than putting some random file keytab somewhere (although that would still seem okay).
> 
> Now I realize that this approach here by itself is bogus -- on Fedora, the default keyring from krb5-libs is `KEYRING:persistent:<uid>`, and that is explicitly meant to survive logouts. The sssd-kcm package overwrites that to `KCM:` to use the global caching daemon, but I guess that's supposed to have a similar lifetime independent from sessions.
> 
> I have some trouble with using `KEYRING:session`, but I'll keep looking at that. `FILE:` certainly works fine.
> 

'session' is tricky and not super well tested.
I do not know if it work properly *before* a login process that
establish a session :/

Using FILEs is fine for temporary stuff, the only downside is cleaning
up the file on failures, but you could always create the file on a
private tmp on tmpfs so that worst case reboot, or destruction of the
private tmp mount will take care of it.

Simo.

-- 
Simo Sorce
RHEL Crypto Team
Red Hat, Inc

martinpitt:
@simo5 : Thanks for your feedback! I looked at what SSH does (always a good idea :grin:), and reworked this to put a FILE: ccache into the user's XDG_RUNTIME_DIR (i.e. tmpfs, automatic eventual cleanup), and clean it up when the session ends. This now stops interference with parallel sessions, and I think is overall nicer. I'd appreciate a quick look from you if my rationale in the commit message and the overall approach makes sense? Thank you!

@allisonkarlitskaya : There's some string manipulation/allocation stuff here which might trigger some raised eyebrows from you -- looking forward to the cleanups!

src/ws/session.c
```diff
@@ -553,10 +552,19 @@ perform_tlscert (const char *rhost)
...
 {
+  gss_key_value_set_desc store;
+  struct gss_key_value_element_struct element;
+  OM_uint32 major, minor;
+  static char ccache[PATH_MAX];
+  int res;
+
   assert (creds != GSS_C_NO_CREDENTIAL);
+  /* can only be called once */
+  assert (ccache[0] == '\0');
```
simo5, line 567: Why not have an actual if check? assert calls are eliminated in optimized builds, normally ...

simo5 (commented):
LGTM, aside from comments

src/ws/session.c
```diff
@@ -568,35 +576,26 @@ store_krb_credentials (gss_cred_id_t creds, uid_t uid, gid_t gid)
...
-      OM_uint32 major, minor;
-
-      store.count = 1;
-      store.elements = &element;
-      element.key = "ccache";
-      element.value = krb5_cc_default_name (k5);
+  /* The ccache path needs to be unique per cockpit session; as cockpit-session runs throughout the
+   * lifetime of sessions, our pid is unique. We expect the cache to be cleaned up at the end, but
+   * if not, and the PID gets recycled, this just overwrites the old obsolete one, which is good. */
+  res = snprintf (ccache, sizeof ccache, "FILE:/run/user/%u/cockpit-session-%u.ccache", uid, getpid ());
```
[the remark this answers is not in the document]
martinpitt: @simo5: Indeed it's not very relevant here as `PATH_MAX` is more than enough, but I'm interested in general. I use this pattern a lot, and my reason for snprintf is that its size argument *includes* the terminating NUL. So if your buffer is too short, you get a truncated string (which may be a vulnerability on itself in some cases, of course), but never a buffer overflow.

But why "- 2"? Nothing appends anything to that string.

src/ws/session.c
martinpitt: Calling it twice would be an internal programming error, so I thought an assertion would be appropriate (we do run tests with a devel build as well, and even our upstream production builds don't compile out assertions). It's similar to the creds assertion above it. But I don't mind much to turn these into runtime checks. I'll wait for @allisonkarlitskaya 's review (I'm sure she finds some other stuff) and can then update this.

Thanks for your review!

martinpitt:
[Some](https://logs.cockpit-project.org/logs/pull-16149-20210726-143723-19b4f899-debian-stable/log.html) [older](https://logs.cockpit-project.org/logs/pull-16149-20210726-143723-19b4f899-ubuntu-2004/log.html#282) [OSes](https://logs.cockpit-project.org/logs/pull-16149-20210726-143723-19b4f899-ubuntu-stable/log.html#282) take way too much time to get their logind sessions cleaned up, I need to check more precisely, or just check for cockpit-bridge.

src/ws/session.c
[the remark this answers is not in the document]
simo5: -2 is a typo 😂

src/ws/session.c
```diff
@@ -28,7 +28,6 @@
 #include <gssapi/gssapi.h>
 #include <gssapi/gssapi_generic.h>
 #include <gssapi/gssapi_krb5.h>
-#include <krb5/krb5.h>
```
allisonkarlitskaya, line 31: Cool :)

src/ws/session.c
allisonkarlitskaya: My opinion is that we shy away from memory allocation too much when we write C.

Just allocate a string?  If you don't like it because you can't return it and later pass it to `unlink()` then return the entire string and add a new trivial function for `delete_krb_credentials()` and do the check for `FILE:` and `+5` there instead, and bonus `free()`?

Bonus is you can use our `asprintfx()` helper and not argue about if you need to add or remove one to the length :)

allisonkarlitskaya (changes requested):
I like it!  Some nags/ideas about string handling.

src/ws/session.c
martinpitt: You are right that in general I am a bit allergic against dynamic allocations, and doubly so in suid root stuff. But in this case the assertion is not just a crutch for the static memory, but an actual first-class intended thing. We should *never* store credentials twice, as the second one would clobber the first, and that'd be an internal program logic error.

If you don't like that style, I'm also okay with having a `static char *ccache` and ensuring that it's `NULL`, and using asprintfx().

allisonkarlitskaya: In general I'm also hoping to dramatically reduce the use of globals and static locals.  I have a hard time imagining a scenario under which we accidentally call this function twice.

If you insist, though, I accept the compromise. :)

martinpitt:
@allisonkarlitskaya : Changed according to your review; it's a dynamic string now, moves the FILE: prefix handling to the cleanup side (I mildly disagree, but also, waaay below the level that I actually care :grin: ). I cleaned up two other small things, so in total it's just 4 more lines now.

martinpitt:
> we create the krb token before we launch the session

That yes, but we only *write* it way after `pam_open_session()`. The writing is defered to `store_krb_credentials()` which is called in main(). I find that structure a bit confusing as well, but the code has been like that for years, and I don't want to change it around without understanding the historical reasons. But it actually works quite nicely in our favor for this case.

> but the XDG runtime dir should be created by logind as part of starting the session, no

Correct. I.e. I think `pam_open_session()` triggers this. And if somehow it doesn't (because of broken PAM config or so), it will just not write the ticket and warn in journal, but not generally break the cockpit login.

allisonkarlitskaya:
@martinpitt looks like you rebased without making any changes?

martinpitt:
Bwah, git pilot error, sorry.. Pushed for real now.

src/ws/session.c
```diff
@@ -568,35 +575,35 @@ store_krb_credentials (gss_cred_id_t creds, uid_t uid, gid_t gid)
...
+}
+
+static void
+release_krb_credentials (char *ccache)
+{
+  /* strip off FILE: prefix for deleting */
+  assert (strncmp (ccache, "FILE:", 5) == 0);
+  if (unlink (ccache + 5) != 0)
+    warn ("couldn't clean up kerberos ticket cache %s", ccache);
+  free (ccache);
```
allisonkarlitskaya, line 606: Since you store the string in a static, the create function should probably return `const` and you should definitely not free it here while there's still a pointer around for it.

... or you could just remove the `static` ;)

allisonkarlitskaya (changes requested):
one more little gripe, sorry

src/ws/session.c
martinpitt: Having a `const` return and not freeing it was my original approach which you didn't like, so I suppose that's a no-go. You really want me to lose that assertion, don't you :grin: (FAOD, I actually *want* this to be a singleton)

martinpitt: OK, pushed -- it now does the same thing as before, except with less performance, more ways to fail, and fewer assertions :man_shrugging:

martinpitt:
There was a [race condition in the test](https://logs.cockpit-project.org/logs/pull-16149-20210727-181001-ff653b6a-ubuntu-2004/log.html#283), fixed.

src/ws/session.c
allisonkarlitskaya: ... these are a few of my favorite things! 🎶

allisonkarlitskaya (approved):

PR #16150, opened 2021-07-26
ws: Use cockpitmemory allocation functions

martinpitt:
The previous dup_string() copied strings containing NUL bytes (and an
explicit length). We don't require that for user or display names, as
they are not allowed to contain NUL bytes, so normal strndup() behaviour
is fine for these.

### DIFF
--- a/src/ws/session.c
+++ b/src/ws/session.c
@@ -20,6 +20,7 @@
 #include "config.h"
 
 #include "common/cockpitjsonprint.h"
+#include "common/cockpitmemory.h"
 
 #include "cockpit-session-client-certificate.h"
 #include "session-utils.h"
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (changes requested):
Note that `dup_string()` isn't exactly `strndup()`, but something more like memdup+x.  I think it's fine in this case, but it deserves a mention in the commit message at least.

martinpitt:
@allisonkarlitskaya : I'll gladly update the commit message, I'm just afraid I don't know where the difference is between strndup() and memcpy() plus extra NUL?

allisonkarlitskaya:
> @allisonkarlitskaya : I'll gladly update the commit message, I'm just afraid I don't know where the difference is between strndup() and memcpy() plus extra NUL?

`strncpy()` copies the given length, or up to the first nul, whatever is less, plus 1 (for nul).  `memdup()` does the whole thing, regardless of any nuls in the input.  So if you wanted to use the given length parameter to index into the string after the copy, that would be fine with `memdup()`, but an error with `strndup()`.

martinpitt:
Ah, of course -- I am aware of memcpu() copying null bytes, but somehow I thought that dup_string() called strlen() itself -- it doesn't. Added a comment and updated the commit message.

allisonkarlitskaya (approved):
Would have been okay without the code # comments, but doesn't hurt with, either.

PR #16153, opened 2021-07-26
webpack-jumpstart.yml: perform merge for ourselves

allisonkarlitskaya:
Using the .merge_commit_sha from GitHub is seriously dangerous (it's
    usually not up to date when the workflow gets launched) and kinda
    pointless (it's completely transient, and will change next time someone
    pushes main).
    
    Perform the merge for ourselves to dodge the first problem and provide
    some information to help reproduce the merge in order to solve the
    second problem.

 - https://github.com/allisoninc/cockpit/pull/2
 - (main push) https://github.com/allisoninc/cockpit/runs/3162944269?check_suite_focus=true
 - (PR) https://github.com/allisoninc/cockpit/runs/3162948761?check_suite_focus=true

### DIFF
--- a/.github/workflows/webpack-jumpstart.yml
+++ b/.github/workflows/webpack-jumpstart.yml
@@ -11,17 +11,21 @@ jobs:
       pull-requests: none
     timeout-minutes: 20
     env:
-      MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha || github.sha }}
+      BASE_SHA: ${{ github.sha }}
+      HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
     steps:
       - name: Build jumpstart tarball
         run: |
-          id="$(docker container create -u node -w /home/node node /bin/sh -c \
-            'git init cockpit &&
-             cd cockpit &&
-             git fetch --depth 1 https://github.com/cockpit-project/cockpit "'"${MERGE_SHA}"'":build &&
-             git checkout build &&
-             pkg/build webpack-jumpstart.tar -j$(nproc)'
-          )"
+          id="$(docker container create -u node -w /home/node node /bin/sh -ec \
+            "git init cockpit
+             cd cockpit
+             git config user.email '<>'
+             git config user.name 'nobody'
+             git remote add origin ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}
+             git fetch --no-tags origin ${BASE_SHA}:base ${HEAD_SHA}:head
+             git checkout --detach base
+             git merge --no-edit head
+             pkg/build webpack-jumpstart.tar -j$(nproc)")"
 
           docker container start --attach "${id}" >&2
           docker container cp "${id}":/home/node/cockpit/webpack-jumpstart.tar .
@@ -43,7 +47,7 @@ jobs:
     timeout-minutes: 5
     env:
       GIT_DIR: git-dir.git
-      MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha || github.sha }}
+      BASE_SHA: ${{ github.sha }}
       HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
     steps:
       - name: Set up work area
@@ -67,9 +71,10 @@ jobs:
         run: |
           set -ux
           mkdir worktree
-          tar -C worktree -x --exclude '.git*' dist package-lock.json < artifact/webpack-jumpstart.tar
-          git --work-tree worktree add dist package-lock.json
-          git --work-tree worktree commit --quiet -m "Build for ${MERGE_SHA}"
+          tar -C worktree -x --exclude '.git*' dist package-lock.json tree < artifact/webpack-jumpstart.tar
+          echo "${BASE_SHA}" > worktree/merge-base
+          git --work-tree worktree add dist package-lock.json merge-base tree
+          git --work-tree worktree commit --quiet -m "Build for ${HEAD_SHA}"
           rm -rf worktree
 
           git tag "sha-${HEAD_SHA}"
--- a/pkg/build
+++ b/pkg/build
@@ -36,8 +36,8 @@ V_TAR = $(V_TAR_$(V))
 V_TAR_ = $(V_TAR_$(AM_DEFAULT_VERBOSITY))
 V_TAR_0 = @echo "  TAR     " $@;
 
-webpack-jumpstart.tar: $(MANIFESTS) package-lock.json
-	$(V_TAR) tar cf webpack-jumpstart.tar dist package-lock.json
+webpack-jumpstart.tar: $(MANIFESTS) package-lock.json tree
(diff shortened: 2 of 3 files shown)


pkg/build
```diff
@@ -75,3 +75,15 @@ V_NODE_MODULES_0 = @V=0
...
+# Make sure the tree is clean, first.  Gets included in the jumpstart tar.
+tree: FORCE
+	@test -z "$$(git status --porcelain)" || (git status && false)
+	@git rev-parse HEAD^{tree} > tree.tmp && \
+	if cmp -s tree tree.tmp; then \
+		rm tree.tmp; \
+	else \
+		printf "  %-8s %s\n" GEN "tree  [$$(cat tree.tmp)]"; \
+		mv tree.tmp tree; \
+	fi
```
martinpitt, line 89: That feels about five times more effort than it's worth, but sure :grin:

.github/workflows/webpack-jumpstart.yml
```diff
@@ -11,16 +11,21 @@ jobs:
...
+            "set -x &&
+             git config --global user.email '<>' &&
+             git config --global user.name 'nobody' &&
+             git init cockpit &&
              cd cockpit &&
-             git fetch --depth 1 https://github.com/cockpit-project/cockpit "'"${MERGE_SHA}"'":build &&
-             git checkout build &&
-             pkg/build webpack-jumpstart.tar -j$(nproc)'
+             git fetch --no-tags https://github.com/cockpit-project/cockpit ${BASE_SHA}:base ${HEAD_SHA}:head &&
+             git checkout --detach ${BASE_SHA} &&
```
martinpitt, line 26: Stylistic: You defined the "base" alias for BASE_SHA, so you may just as well use it? (I'd actually be fine with just directly using the macros, and drop the aliases)

.github/workflows/webpack-jumpstart.yml
```diff
@@ -11,16 +11,21 @@ jobs:
...
+             git merge --no-edit head &&
```
martinpitt, line 27: We rebase in both integration tests (`make-checkout`) and for final landing. It would IMHO be nicer to consistently use rebase. AFAIK you don't really use the merge commit anywhere? That would also save you the `git config` dance.

.github/workflows/webpack-jumpstart.yml
```diff
@@ -67,9 +72,10 @@ jobs:
         run: |
           set -ux
           mkdir worktree
-          tar -C worktree -x --exclude '.git*' dist package-lock.json < artifact/webpack-jumpstart.tar
-          git --work-tree worktree add dist package-lock.json
-          git --work-tree worktree commit --quiet -m "Build for ${MERGE_SHA}"
+          tar -C worktree -x --exclude '.git*' dist package-lock.json tree < artifact/webpack-jumpstart.tar
+          echo "${BASE_SHA}" > worktree/merge-base
```
martinpitt, line 76: I'm a bit confused here -- if we can get the original BASE_SHA in this `on_workflow`, why even bother with building the `tree` file in the `build` job? At this point `BASE_SHA == tree`, no?

martinpitt: Sorry, that was a thinko of course -- a tree ID != commit SHA. But still, couldn't this do `git rev-parse` the BASE_SHA to a tree object here?

.github/workflows/webpack-jumpstart.yml
allisonkarlitskaya: I wanted to use the sha directly so the merge doesn't go on the "base" branch.  I only added `--detach` later to suppress the warning about it, but in fact, with `--detach`, I can use the branch name and still get a detached head.  I'll do this.

.github/workflows/webpack-jumpstart.yml
allisonkarlitskaya: Rebase would even be nice because you can do it as a one-liner.

I use merge because GitHub uses merge and so will packit. 

Also, I still need the dance for rebase because of the committer email (and no option I could find to disable the rewriting).

I have a (probably irrational) concern about merges and rebasing producing different results.

martinpitt: "GitHub uses merge" -- in what way? not in any that concerns us at least. Assuming that merging really could produce a different tree than rebasing, then we *really* want to rebase here, to be consistent with what happens in reality (CI and landing)

allisonkarlitskaya: GitHub refuses to rebase and merge if rebasing would produce a different result from merging.

I actually consider the fact that I intend to implement this via rebasing on the other side a feature: that way we'll find out if something changed.

I can try to implement this as rebase, but I also don't see any good reason in favour, and most of my "gut feeling" points to merge.

allisonkarlitskaya: ..
and of course, I mean GitHub uses merges for running workflows.

martinpitt: I just validated that `git merge --no-edit` fails with exit code 1 on conflicts. It leaves the tree in a complete mess (which we don't ever want to build -- if it succeeds, because the conflict is in the C part, we lose). So to make this work (regardless of rebase vs. merge), the container command needs to be invoked with the shell's `-e` option.

.github/workflows/webpack-jumpstart.yml
```diff
@@ -11,16 +11,21 @@ jobs:
...
     env:
-      MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha || github.sha }}
+      BASE_SHA: ${{ github.sha }}
+      HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
     steps:
       - name: Build jumpstart tarball
         run: |
           id="$(docker container create -u node -w /home/node node /bin/sh -c \
-            'git init cockpit &&
+            "set -x &&
```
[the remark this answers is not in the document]
martinpitt: .. I was missing that all of them are chained with `&&`, so that's not important. (Again stylistics: `set -e` and a multi-line string would look a bit cleaner?)

.github/workflows/webpack-jumpstart.yml
allisonkarlitskaya: Correct.  That will be used to do some magic in the future with the jumpstart script, possibly via a new `--rebase` argument.  I wanted to leave that for later, though.

.github/workflows/webpack-jumpstart.yml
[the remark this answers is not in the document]
allisonkarlitskaya: interesting proposal.  I'll give it a try :)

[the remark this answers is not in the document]
allisonkarlitskaya: I moved to your idea and added an `origin` remote to shorten the long line.

.github/workflows/webpack-jumpstart.yml
allisonkarlitskaya: done.

martinpitt (approved):
Looks great, thank you!

PR #16156, opened 2021-07-27
Split out cockpit-session into new directory, fdwalk/close_range code cleanup

martinpitt:
First round of cleanups for cockpit-session.

 - [ ] Builds on top of 16149


martinpitt:
The fdwalk cleanup caused [a regression](https://logs.cockpit-project.org/logs/pull-16156-20210727-065650-55c4e379-fedora-34/log.html#162):
   
    couldn't get agent pid from ssh-agent output:

I'll investigate this in a follow-up, I backed out the last commit.

allisonkarlitskaya:
Please wait on this.  I said I wanted to do it *after* the client cert stuff lands 😅

martinpitt:
@allisonkarlitskaya : ah sorry for the misunderstanding, I thought you wanted me to do this *before*. As the client cert rework is still some way out, I suppose I'll rework my S4U bits and the fdfwalk cleanup (#16157) to not depend on this PR then?

martinpitt:
Moving to draft and blocking on #16000 then.

martinpitt:
I rebased this to current main plus #16168, this was quite some work to catch up with the recent cockpit-tls cleanup. But now I don't expect this to change much any more.

allisonkarlitskaya (changes requested):
Misses changing the name of the cross-file reference in `src/tls/client-certificate.c`.

Not crazy about the new shorter names... I guess we can bikeshed that a bit.

martinpitt:
> Misses changing the name of the cross-file reference in src/tls/client-certificate.c.

Thanks! Fixed locally, will push after #16168 lands.

> Not crazy about the new shorter names... I guess we can bikeshed that a bit.

I seemed to me that it fell into the session-utils.h pattern, and we could use `session-*authtype*.c`. The long name now has too much no scoping for me, and too long overall. But if you'd rather keep that, I can back out the rename.

martinpitt:
Rebased, and fixed the file ref in the comment. Happy to bikeshed this further -- @allisonkarlitskaya , please feel free to force-push this during my PTO. I don't have any particular urgency about that myself.

martinpitt:
@allisonkarlitskaya : Did the renaming as discussed in our 1-on-1.

allisonkarlitskaya (changes requested):
Sorry for the more-bikeshed, but although I agree with dropping the prefix from the filename (since the directory strongly implies it now) I don't like renaming the functions.

martinpitt:
@allisonkarlitskaya OK, renamed the function back.

allisonkarlitskaya (approved):

PR #16157, opened 2021-07-27
Clean up cockpitunixfd.[hc], move to Linux close_range API, various simplifications

martinpitt:
Linux 5.9 finally added a close_range() call [1]. This is not yet
exposed by glibc, but let's move our code to be ready to use it once it
is available. Add a new common cockpit_close_range() function which
offers the same API, and use it everywhere.

Remove cockpit_unix_fd_close_*() from cockpitunixfd.h. It does not fit
there, as (1) that file otherwise contains backports of [2] and will
eventually disappear, and (2) we also want to use close_range() from
programs that don't include glib.

This also gets rid of some fdwalk() implementation duplication.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=278a5fbaed89dacd04e9d
[2] https://developer.gnome.org/glib/stable/glib-UNIX-specific-utilities-and-integration.html


martinpitt:
As Lis wants to postpone #16156, I rebased this to current main. This should not interfere any more with anything else now.

martinpitt:
I think it's worth landing this now:

 * We remove the triplication of code (I diffed the implementations to make sure they do the same thing)
 * We avoid introducing more code that uses the old API (in particular with the `except` thing, which `close_range()` does not support)
 * This gives us a central place to introduce `close_range()`, with the `#ifdef HAVE_CLOSE_RANGE` (for old glibcs) and handling `ENOSYS` (for old kernels). We'll need these fallbacks for the next few years for at least RHEL/CentOS 8 and older Debians/Ubuntus, but I don't want to wait that long for using `close_range()` once it becomes available.

src/bridge/bridge.c
```diff
@@ -177,7 +177,8 @@ static void
 setup_dbus_daemon (gpointer addrfd)
```
allisonkarlitskaya, line 177: This is begging for a rewrite: hardcode `--print-address=3` and `dup2()` the fd over, and just close everything from 4 and up.

Even better: use `GSubproces` instead, with `g_subprocess_launcher_take_fd()`, and drop the child setup function.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,110 @@
...
+  return res;
+}
+
+#endif /* HAVE_FDWALK */
+
+
+int
+cockpit_close_range (int from, int max_fd, int flags)
+{
+  assert (flags == 0);
```
allisonkarlitskaya, line 107: Just drop the flags parameter, if you're going to do that.  This is an internal API after all.

Drop the `max_fd` paramter while you're at it, since there is one user of it that should be refactored anyway.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,110 @@
...
+            break;
+          warnx ("couldn't close fd: %m");
+          return -1;
+        }
+    }
+
+  return 0;
+}
+
+#ifndef HAVE_FDWALK
```
allisonkarlitskaya, line 38: I have no idea on what kind of Linux system we're supposed to have `fdwalk()`, but it's not a part of glibc.  Please drop the conditionalization of this function and the corresponding check for it in `configure.ac`.

Meanwhile, the callback structure here is actually pointless and has always bothered me: merge the struct typedef plus all three functions in this file into a single simple function, with a single parameter, which does one thing: closes all fds above a certain number.

allisonkarlitskaya (changes requested):
I like the -339, but I think we can do better than +137 :)

src/pam-ssh-add/Makefile.am
```diff
@@ -1,17 +1,19 @@
...
 
-libpam_ssh_add_a_CFLAGS = $(NULL)
+libpam_ssh_add_a_CFLAGS = -fPIC
 
 libpam_ssh_add_a_LIBS = \
 	libpam_ssh_add.a \
 	$(PAM_LIBS)
 
-pam_ssh_add_FILES = src/pam-ssh-add/pam-ssh-add.c src/pam-ssh-add/pam-ssh-add.c
+pam_ssh_add_FILES = src/pam-ssh-add/pam-ssh-add.c
```
allisonkarlitskaya, line 16: Well, that's nice I guess?

src/pam-ssh-add/Makefile.am
```diff
@@ -1,17 +1,19 @@
...
 
 libpam_ssh_add_a_SOURCES = \
+	src/common/cockpitcloserange.h \
+	src/common/cockpitcloserange.c \
 	src/pam-ssh-add/pam-ssh-add.c \
 	src/pam-ssh-add/pam-ssh-add.h \
 	$(NULL)
 
-libpam_ssh_add_a_CFLAGS = $(NULL)
+libpam_ssh_add_a_CFLAGS = -fPIC
```
allisonkarlitskaya, line 10: ...and what's this?

allisonkarlitskaya (changes requested):
These either slipped in by accident, or belong to a separate commit.  Either way, more explanation please :)

src/common/cockpitcloserange.c
martinpitt: Nack -- the *point* was to move the `close_range()` API, which has this very API. As the next step this will call the actual `close_range()` once it's available in glibc, and keep the current code as a fallback, and eventually we can entirely drop it.

src/common/cockpitcloserange.c
martinpitt: I can add simplification of fdwalk to my TODO list, but I don't want to do it in this PR -- this cleanly moves code around withouth changing behaviour or implementation.

src/pam-ssh-add/Makefile.am
martinpitt: Right, this is unrelated cleanup. Split out into a separate commit.

src/pam-ssh-add/Makefile.am
martinpitt: It was a bug in the Makefile which didn't surface until trying to add a second .c file. Fixed in a separate commit.

martinpitt:
> I like the -339, but I think we can do better than +137 :)

Have a look at PR #16164 which is queued after this one -- it removes even more :grin:

src/common/cockpitcloserange.c
martinpitt: I did add a commit which drops the configure check, this is easy enough and safe.

martinpitt: OK, on second thought I also added a commit right here which simplifies the function as you mentioned.

src/bridge/bridge.c
martinpitt: Done. Tested with:
```
❱❱❱ make && env -u DBUS_SESSION_BUS_ADDRESS G_MESSAGES_DEBUG=all ./cockpit-bridge | cat
[...]
(cockpit-bridge:232955): cockpit-bridge-DEBUG: 08:49:54.609: launched dbus-daemon: 232957
(cockpit-bridge:232955): cockpit-bridge-DEBUG: 08:49:54.613: session bus address: unix:abstract=/tmp/dbus-uolitsRRHL,guid=1fe30868a2766df409b2654961024f92
```
... and I can see the process properly in `ps`. This results in 31 fewer lines of code, and looks quite nice indeed. Thanks for pointing out, I didn't know that API yet.

src/bridge/bridge.c
```diff
@@ -173,66 +173,41 @@ send_init_command (CockpitTransport *transport,
...
-
-  g_spawn_async_with_pipes (NULL, dbus_argv, NULL, flags,
-                            setup_dbus_daemon, GINT_TO_POINTER (addrfd[1]),
-                            &pid, NULL, NULL, NULL, &error);
-
-  close (addrfd[1]);
+  g_autoptr(GSubprocessLauncher) dbus_daemon = g_subprocess_launcher_new (G_SUBPROCESS_FLAGS_STDOUT_SILENCE |
+                                                                          G_SUBPROCESS_FLAGS_STDERR_SILENCE);
+  g_subprocess_launcher_unsetenv (dbus_daemon, "G_DEBUG");
+  g_subprocess_launcher_take_fd (dbus_daemon, addrfd[1], 4);
```
allisonkarlitskaya, line 196: Why 4?

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,92 @@
...
+
+          fd = (int) l;
+
+          if ((long) fd != l)
+              continue;
+
+          if (fd == dirfd (d))
+              continue;
+
+          if ((res = closefd (from, max_fd, fd)) != 0)
```
allisonkarlitskaya, line 69: Just put the condition here and call `close()` directly?

All of the error checking above is a bit weird imho.  `close()` is always successful. Any error you get there is delayed from a previously failed operation (and the fd is still closed).  The obvious exception is EBADF, but even in that case you end up with a closed fd.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,92 @@
...
+   * rlimit trick */
+
+#endif
+
+  if (getrlimit (RLIMIT_NOFILE, &rl) == 0 && rl.rlim_max != RLIM_INFINITY)
+      open_max = rl.rlim_max;
+  else
+      open_max = sysconf (_SC_OPEN_MAX);
+
+  for (fd = 0; fd < open_max; fd++)
```
allisonkarlitskaya, line 87: And here, just start the loop from the minimum fd and then you don't need any extra logic.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,92 @@
...
+cockpit_close_range (int from, int max_fd, int flags)
+{
+  int open_max;
+  int fd;
+  int res = 0;
+
+  struct rlimit rl;
+
+  /* we want to keep the API compatible to glibc's upcoming close_range(), but don't implement flags */
+  assert (flags == 0);
```
allisonkarlitskaya, line 41: May as well assert that the upper range is -1 as well.  It lets us simplify logic some.

src/common/cockpitpipe.c
```diff
@@ -1407,7 +1408,7 @@ cockpit_pipe_pty (const gchar **argv,
   pid = forkpty (&fd, NULL, NULL, &winsz);
   if (pid == 0)
     {
-      if (cockpit_unix_fd_close_all (3, -1) < 0)
+      if (cockpit_close_range (3, G_MAXINT, 0) < 0)
```
allisonkarlitskaya, line 1411: It's a bit weird to see the glib macro used with a "pure C" call.  Maybe that's because you'd need another header for the libc one?

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,92 @@
+#include "cockpitcloserange.h"
```
allisonkarlitskaya, line 1: Copyright header please, and #include "config.h" as the first line.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,92 @@
...
+  int open_max;
+  int fd;
+  int res = 0;
+
+  struct rlimit rl;
+
+  /* we want to keep the API compatible to glibc's upcoming close_range(), but don't implement flags */
+  assert (flags == 0);
+
+#ifdef __linux__
```
allisonkarlitskaya, line 43: Cockpit is always on Linux....

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,92 @@
...
+  DIR *d;
+
+  if ((d = opendir ("/proc/self/fd"))) {
+      struct dirent *de;
+
+      while ((de = readdir (d))) {
+          long l;
+          char *e = NULL;
+
+          if (de->d_name[0] == '.')
```
allisonkarlitskaya, line 53: That seems redundant with the error check after the int parsing.

allisonkarlitskaya (changes requested):

martinpitt:
Fixed a [memory leak](https://github.com/cockpit-project/cockpit/pull/16157/checks?check_run_id=3189233294)

src/bridge/bridge.c
martinpitt: Right, should have taken 42! (I looked at the number in the original code, and it was 4, that must have stuck in  my brain). Changed to 3.

src/common/cockpitcloserange.c
martinpitt: Dropped.

src/common/cockpitcloserange.c
martinpitt: Done.

src/common/cockpitcloserange.c
martinpitt: Done, thanks for pointing out that close() is always successful in Linux -- that helps indeed!

src/common/cockpitcloserange.c
martinpitt: Done, and also added the missing `#pragma once`.

src/common/cockpitpipe.c
martinpitt: Replaced with `INT_MAX`.

src/common/cockpitcloserange.c
martinpitt: I validated that this works fine, strtol()'s end pointer is indeed at the `.` then (it's not trying funny float parsing stuff). Dropped.

martinpitt:
@allisonkarlitskaya : Addressed all your comments, thanks for your great review!

src/bridge/bridge.c
```diff
@@ -467,7 +436,7 @@ run_bridge (const gchar *interactive,
   gboolean closed = FALSE;
   const gchar *directory;
   struct passwd *pwd;
-  GPid daemon_pid = 0;
+  g_autoptr (GSubprocess) dbus_daemon_process = NULL;
   GPid agent_pid = 0;
```
martinpitt, line 440: FTR: once this gets a +1 and lands, I'm happy to rewrite the SSH agent setup as well, as right now it's asymmetric. But I wanted to get one right first.

src/bridge/bridge.c
```diff
@@ -173,66 +173,41 @@ send_init_command (CockpitTransport *transport,
...
 {
-  GError *error = NULL;
-  GString *address = NULL;
+  g_autoptr(GError) error = NULL;
   gchar *line;
   gsize len;
   gssize ret;
-  GPid pid = 0;
-  gchar *print_address = NULL;
+  GSubprocess *process = NULL;
```
allisonkarlitskaya, line 183: if you're gonna be juggling this all around anyway, can you move it to the style of declaration the variable at its site of first use?

src/bridge/bridge.c
```diff
@@ -173,66 +173,41 @@ send_init_command (CockpitTransport *transport,
...
     {
       g_warning ("pipe failed to allocate fds: %m");
-      goto out;
+      return NULL;
     }
 
-  print_address = g_strdup_printf ("--print-address=%d", addrfd[1]);
-  dbus_argv[1] = print_address;
-
   /* The DBus daemon produces useless messages on stderr mixed in */
```
allisonkarlitskaya, line 192: also stdout?  I wonder if we couldn't use a stdout pipe: that would be a lot easier than all of this messing about with launchers and pipes and read loops...

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,88 @@
...
+#include <limits.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/resource.h>
+
+int
+cockpit_close_range (int from, int max_fd, int flags)
+{
+  int open_max;
```
allisonkarlitskaya, line 36: move variables down to where we use them.  in the case of `fd` you'll have it twice, but that's good.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,88 @@
...
+          if (fd == dirfd (d))
+              continue;
+
+          /* don't bother about error checking; On Linux, EINTR still closes the fd, and with EBADF we already have a closed fd */
+          if (fd >= from)
+            close (fd);
+        }
+
+      closedir (d);
+      return 0;
```
allisonkarlitskaya, line 73: You always return 0 now.  Make this void and drop the error handling from the callers.

allisonkarlitskaya (changes requested):

src/bridge/bridge.c
martinpitt: Done.

src/bridge/bridge.c
martinpitt: The current `dbus-daemon` that I have works ok with `--print-address` (which goes to stdout). I'm a bit afraid of dbus-daemon writing anything else there, but given that it is advertised for exactly this use case, let's do this. I do this as a separate commit, just in case.

src/common/cockpitcloserange.c
martinpitt: Done.

src/common/cockpitcloserange.c
martinpitt: close_range() returns an int, I'd like to keep that API.

martinpitt:
I already retried packit builds once, it seems [centos-8-stream COPR builds](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-16157/centos-stream-8-x86_64/02352532-cockpit/builder-live.log.gz) have some ongoing networking problems. This is unrelated though, and our own [centos-8-stream](https://logs.cockpit-project.org/logs/pull-16157-20210729-085640-9a2f130b-centos-8-stream/log.html) build shows that there are no API incompatibilities.

src/bridge/bridge.c
```diff
@@ -173,79 +173,39 @@ send_init_command (CockpitTransport *transport,
...
 
-  address = g_string_new ("");
+  g_autoptr(GString) address = g_string_new ("");
   for (;;)
     {
-      len = address->len;
+      gsize len = address->len;
       g_string_set_size (address, len + 256);
-      ret = read (addrfd[0], address->str + len, 256);
+      gssize ret = g_input_stream_read (g_subprocess_get_stdout_pipe (process), address->str + len, 256, NULL, &error);
```
allisonkarlitskaya, line 203: `read_all()`

allisonkarlitskaya (changes requested):

src/bridge/bridge.c
martinpitt: Can't -- dbus-daemon does not close its stdout after writing the address, so read_all() will hang forever (I did try this). I'm afraid we have to read the full line. I was a bit disappointed that there's no `g_input_stream_read_line()` as an fgets() counterpart, that'd be really handy.

allisonkarlitskaya: `g_data_input_stream_read_line ()`

martinpitt: Hah, thanks! So much nicer now. Pushed.

martinpitt:
This is just a bit short of `g_spawn_subprocess_and_read_one_line_from_it()` :grin:

allisonkarlitskaya (changes requested):
Code looks good, but after all of that back and forth this needs a bit of rebasing: let's do the GSubprocess cleanups first and then do the implementation of the close_range() call in a single commit, in its current form.  There's no sense doing the more complicated implementation only to simplify it a couple of commits later.

Also: one of your commit messages talks about "stout" instead of "stdout".  Time for a bear? :)

martinpitt:
@allisonkarlitskaya : Rebased in the suggested order, and validated that it was a net-zero change. As this is such a grab-bag of cleanups now, I also added the removal of cockpitunixfd.[hc], as that's fairly simple now.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,85 @@
...
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "config.h"
+#include "cockpitcloserange.h"
```
allisonkarlitskaya, line 21: Usually a newline between `config.h` and the other includes.

src/common/cockpitcloserange.c
```diff
@@ -0,0 +1,85 @@
...
+
+  if ((d = opendir ("/proc/self/fd"))) {
+      struct dirent *de;
+
+      while ((de = readdir (d))) {
+          long l;
+          char *e = NULL;
+
+          errno = 0;
+          l = strtol (de->d_name, &e, 10);
```
allisonkarlitskaya, line 50:
```suggestion
          long l = strtol (de->d_name, &e, 10);
```

allisonkarlitskaya (changes requested):
trivial

src/common/cockpitcloserange.c
allisonkarlitskaya: also: pretty sure you don't need to initialize `e`.

src/common/cockpitcloserange.c
martinpitt: Done.

src/common/cockpitcloserange.c
martinpitt: Done both.

allisonkarlitskaya (approved):
tja.

^5.

PR #16158, opened 2021-07-27
implement `tools/webpack-jumpstart --rebase`

allisonkarlitskaya:
- a `--rebase` option will allow rebasing the local tree to get the same tree we built the cache against
 - update packit to use `--rebase`

### DIFF
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -31,8 +31,6 @@ done
 
 [ -n "${quiet}" ] || set -x
 
-tools/node-modules make_package_lock_json
-
 if [ -e dist ]; then
     echo "jumpstart: dist/ already exists, skipping" >&2
     exit 1
@@ -60,6 +58,8 @@ for try in $(seq 50 -1 0); do
     message WAIT 30s
     sleep 30s
 done
+
+tools/node-modules make_package_lock_json
 
 if ! cmp_from_cache "${tag}" "package-lock.json" "package-lock.json"; then
     echo "The cached package-lock.json doesn't match our own" >&2
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -17,6 +17,18 @@ set -eu
 cd "$(realpath -m "$0"/../..)"
 . tools/git-utils.sh
 
+wait=''
+while [ $# != 0 ] ; do
+    case "$1" in
+        --wait)
+            wait=1;;
+        *)
+            echo "usage: $0 [--wait]" >&2
+            exit 1
+    esac
+    shift
+done
+
 [ -n "${quiet}" ] || set -x
 
 tools/node-modules make_package_lock_json
@@ -41,7 +53,7 @@ for try in $(seq 50 -1 0); do
     if fetch_to_cache tag "${tag}"; then
         break
     fi
-    if [ "${1-}" != '--wait' -o "$try" = '0' ]; then
+    if [ -z "${wait}" -o "$try" = '0' ]; then
         echo "There is no cache entry ${tag}" >&2
         exit 1
     fi
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -18,12 +18,15 @@ cd "$(realpath -m "$0"/../..)"
 . tools/git-utils.sh
 
 wait=''
+rebase=''
 while [ $# != 0 ] ; do
     case "$1" in
         --wait)
             wait=1;;
+        --rebase)
+            rebase=1;;
         *)
-            echo "usage: $0 [--wait]" >&2
+            echo "usage: $0 [--rebase] [--wait]" >&2
             exit 1
     esac
     shift
@@ -58,6 +61,21 @@ for try in $(seq 50 -1 0); do
     message WAIT 30s
     sleep 30s
 done
+
+if [ -n "${rebase}" ]; then
+    if [ -n "$(git status --porcelain)" ]; then
+        echo 'Refusing to rebase tree with local changes' >&2
+        git status >&2
+        exit 1
+    fi
+    merge_base="$(cat_from_cache "${tag}" merge-base)"
+    # If we don't already have that commit, we'll need to go fetch it
+    if ! git fsck --no-dangling --connectivity-only "${merge_base}"; then
+        message FETCH ".  [${merge_base}]"
+        git fetch --no-write-fetch-head origin "${merge_base}"
+    fi
+    git rebase -- "${merge_base}"
+fi
 
 tools/node-modules make_package_lock_json
 
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -77,11 +77,16 @@ if [ -n "${rebase}" ]; then
     git rebase -- "${merge_base}"
 fi
 
-tools/node-modules make_package_lock_json
-
-if ! cmp_from_cache "${tag}" "package-lock.json" "package-lock.json"; then
-    echo "The cached package-lock.json doesn't match our own" >&2
-    exit 1
+target_tree="$(cat_from_cache "${tag}" tree)"
+if [ "$(git rev-parse HEAD^{tree})" != "${target_tree}" ]; then
+    if [ -n "${rebase-}" ]; then
+        echo "Internal error: even after rebase, don't have the correct tree" >&2
+        exit 1
+    else
+        echo "The current working tree needs to be rebased: try --rebase" >&2
+        exit 1
+    fi
 fi
 
+tools/node-modules make_package_lock_json
 unpack_from_cache "${tag}"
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
I did some testing in context of https://github.com/cockpit-project/cockpit/pull/16160

`--rebase` seems to be working properly when I locally test (manually copying over the script).  Even the combination with `--best` is working as expected (I did a commit with the script update, to test).

Of course, packit builds in that PR are broken because of `package-lock.json` inconsistency and the fact that it doesn't contain the new webpack-jumpstart script yet... but that's kinda the point :)

martinpitt:
Triggering [fedora-34 test](https://logs.cockpit-project.org/logs/pull-16158-20210727-105605-2e4d24d5-fedora-34/log.html) after jumpstart build completed, to make sure it works alright with the cached download.

That somehow did not work:
```
  FETCH    dist  [ref: tag sha-7f04ffb57872baeacac3f4352bef3e918ffaa00f]
fatal: couldn't find remote ref refs/tags/sha-7f04ffb57872baeacac3f4352bef3e918ffaa00f
There is no cache entry sha-7f04ffb57872baeacac3f4352bef3e918ffaa00f: try --best?
```

allisonkarlitskaya:
> Triggering [fedora-34 test](https://logs.cockpit-project.org/logs/pull-16158-20210727-105605-2e4d24d5-fedora-34/log.html) after jumpstart build completed, to make sure it works alright with the cached download.
> 
> That somehow did not work:
> 
> ```
>   FETCH    dist  [ref: tag sha-7f04ffb57872baeacac3f4352bef3e918ffaa00f]
> fatal: couldn't find remote ref refs/tags/sha-7f04ffb57872baeacac3f4352bef3e918ffaa00f
> There is no cache entry sha-7f04ffb57872baeacac3f4352bef3e918ffaa00f: try --best?
> ```

That makes perfect sense: the CI builds rebase the branch against main (which has received new commits in the meantime).  The commit that they have would therefore not be one that we have a cache for.  This was always what happened in that case.

tools/webpack-jumpstart
```diff
@@ -17,6 +17,18 @@ set -eu
 cd "$(realpath -m "$0"/../..)"
 . tools/git-utils.sh
 
+wait=''
+while [ $# != 0 ] ; do
```
martinpitt, line 23: Let's see when we move to `getopt(1)` :grin:

tools/webpack-jumpstart
```diff
@@ -41,7 +53,7 @@ for try in $(seq 50 -1 0); do
     if fetch_to_cache tag "${tag}"; then
         break
     fi
-    if [ "${1-}" != '--wait' -o "$try" = '0' ]; then
+    if [ -z "${wait-}" -o "$try" = '0' ]; then
```
martinpitt, line 71: I'd prefer you wouldn't use the `-` fallback for undefined variables here, to better spot typos. You explicitly initialize the known options to empty strings (also in the subsequent commit).

tools/webpack-jumpstart
```diff
@@ -19,14 +19,17 @@ cd "$(realpath -m "$0"/../..)"
...
     case "$1" in
         --wait)
             wait=1;;
         --rebase)
             rebase=1;;
+        --best)
+            best=1;;
         *)
-            echo "usage: $0 [--rebase] [--wait]" >&2
+            echo "usage: $0 [--best] [--rebase] [--wait]" >&2
```
martinpitt, line 32: Given that us puny humans are rather sensitive to antimatter, perhaps disallow specifying both options at the same time, at least for the time being?

tools/webpack-jumpstart
```diff
@@ -85,3 +104,8 @@ fi
 
 tools/node-modules make_package_lock_json
 unpack_from_cache "${tag}"
+
+for file in $(git diff --name-only "${target_tree}"); do
+    message TOUCH "${file}"
+    touch "${file}"
+done
```
martinpitt, line 111: I don't have a concrete scenario where this would interfere, but somehow it feels wrong/too imprecise to do that. It would render all built object/binary files etc. as out of date as well (I haven't thought much about the implications of touching package.json and friends in the wrong order). I'd much rather do this with `--rebase` only, then at least the damage is more controlled.

tools/webpack-jumpstart
```diff
@@ -44,18 +47,29 @@ if [ "${NODE_ENV-}" = "development" ]; then
     exit 1
 fi
 
-if ! git diff --quiet -- ':^test' ':^packit.yaml' ':^.github'; then
-    echo 'jumpstart: uncommitted local changes, skipping download' >&2
-    exit 1
```
martinpitt, line 36: I really don't like dropping this check unconditionally. Now downloading the cache would claim to work even with local modifications, but then do that in vain and rebuild stuff anyway. To me as a developer this is too mushy and unpredictable -- either the cache applies to my tree or it doesn't. In the latter case, rebuilding webpacks is not really the end of the world.

martinpitt (commented):
The first six commits are straightforward and obviously nice (my minor style nags are not a blocker). However, the --best thing is just horrible IMHO. If you find someone in the team who likes it, I don't veto it, but I can't get it over me to +1 this myself..

tools/webpack-jumpstart
allisonkarlitskaya: Good catch.  I didn't do the initialization before, but then I thought about how they might sneak in via the environment, so I added it.  Then I forgot to remove this.

tools/webpack-jumpstart
allisonkarlitskaya: Honestly, it doesn't do a whole lot...

tools/webpack-jumpstart
allisonkarlitskaya: I actually tried it and it works as I expected: it rebases the PR, and then touches the files from your extra commits.  I'll adjust the commit message.

tools/webpack-jumpstart
allisonkarlitskaya: This actually has more to do with `--best` than with `--rebase` (which should always end up with the identical tree).  In case you don't do `--best` then this will only touch files that you've modified locally and haven't yet committed.

I was having the same thoughts as you, though and went through a couple of different phases there:
 1. what if we're touching files that have some other build dependency order between them?  that could get weird if we touch them in the wrong order
 2. but, it's probably OK, because the built files are all ignored by git
 3. but, regardless, avoiding rebuilding stuff would be nice

What I landed at in terms of the actual proper approach here is to scan the dist/*/Makefile.deps for the changed files and delete the build results for files that have changed.  That's the purest thing we can do.  I was going to save that for some future date, but since you mention it now, I can give it a try :)

tools/webpack-jumpstart
allisonkarlitskaya: See the above comment about my alternate proposal.  We could maybe include this behaviour in `--best`: having some dirty working tree is kinda similar to having some commits on top of the most recent cache object?

tools/webpack-jumpstart
martinpitt: Sorry, of course I meant `--best` above, that was a typo (fixed). Scanning Makefile.deps and touching the "correct" files in correct order honestly is far beyond what I'd tolerate as "reasonable effort" -- this is rebuilding make rules and `make`. So I'd rather dial this *down* than *up*, and pay the price of occasional webpack build for keeping this reasonably simple and predictable.

allisonkarlitskaya: Dropping this for this PR.  Will bring it back later, with improvements as discussed.

martinpitt (approved):
Thanks! Indeed, it's much easier to land this in two steps, and get the rebasing out of the way.

martinpitt:
Argh:

    /sandcastle/quay-io-packit-sandcastle-prod-20210727-125102788368/tools/webpack-jumpstart: line 71: cat_from_cache: command not found

martinpitt (changes requested):
packit build regression

allisonkarlitskaya:
Looks like I dropped one too many commits.  Moved the utility function to before the commit which uses it and pushed again.

martinpitt (dismissed):
Danke! (Assuming green again), I triggered the f34 test.

allisonkarlitskaya:
```
  FETCH    dist  [ref: tag sha-8923cef249182f7dd7f7f88953170ea4530734c9]
Rebasing (1/6)
Committer identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'sandcastle@quay-io-packit-sandcastle-prod-20210727-130055405074.(none)')
```

fml :(

allisonkarlitskaya:
packit is happy with non-rebased.  Now let's try pushing a fully-rebased one.

martinpitt (approved):
Vielen Dank! :1st_place_medal:

PR #16159, opened 2021-07-27
users: Don't allow locking of the current account

### LINKED ISSUE #16152: accounts: Don't let someone lock their currently used account
_Originally posted by @Verhoeckx in https://github.com/cockpit-project/cockpit/discussions/16133#discussioncomment-1037900:_

> I wonder however if it's possible to lock all accounts (so you won't be able to log in anymore in Cockpit)?
> Because I still have the option to also lock my own user account :fearful: !

It's basically a "break my system" feature that we shouldn't expose.

marusak:
Locking of not-current account is already tested.

Fixes #16152


martinpitt (approved):
Dakujem! Agreed this makes sense. On a single-user machine you basically never want to do this anyway.

PR #16162, opened 2021-07-27
lib: python.js: detect python path and then spawn the python script seperately

KKoukiou:
Previously python.js was passing the python command through the shell,
resulting in the args with special chars breaking formatting.

### DIFF
--- a/pkg/lib/python.js
+++ b/pkg/lib/python.js
@@ -20,8 +20,6 @@
 import cockpit from "cockpit";
 
 // FIXME: eventually convert all images to python 3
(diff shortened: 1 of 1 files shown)


martinpitt:
I'm not so sure that I like that -- it will turn *every* python.spawn() in cockpit from one into two calls, even though all the current ones don't need that, and we can certainly assume that the value of the `which` doesn't change during the lifetime of a cockpit session. I was thinking of putting that logic into cockpit-machines itself, and not use python.spawn() there at all -- i.e. at page load, do the `which` thing and remember the result, and then on invoking the virt-install helpers, directly call the remembered command line program.

PR #16164, opened 2021-07-28
common: Drop cockpitunixfd in favor of official glib API

martinpitt:
These are backports of the official glib API [1]. This has existed in
glib since 2.30, and configure.ac requires ≥ 2.44.

[1] https://developer.gnome.org/glib/stable/glib-UNIX-specific-utilities-and-integration.html

 - [ ] Builds on top of #16157


martinpitt:
Folded into PR #16157

PR #16165, opened 2021-07-28
partial jumpstart

allisonkarlitskaya:
purely for discussion purposes

### DIFF
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -106,7 +106,7 @@ done
 if [ -n "${rebase}" ]; then
     merge_base="$(cat_from_cache "${tag}" merge-base)"
     # If we don't already have that commit, we'll need to go fetch it
-    if ! git fsck --no-dangling --connectivity-only "${merge_base}"; then
+    if ! git rev-list --quiet --objects "${merge_base}"; then
         message FETCH ".  [${merge_base}]"
         git fetch --no-write-fetch-head origin "${merge_base}"
     fi
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -50,14 +50,12 @@ if [ -n "$(git status --porcelain)" ]; then
     git status >&2
     exit 1
 fi
-
-# If we are rebasing and the HEAD commit is a merge, use the merged
-# commit ID to find the cached data, instead of the merge itself.
+# If we are rebasing and the HEAD commit is a merge, reset to the
+# merged commit before proceeding.
 if test -n "${rebase}" && git rev-parse --verify HEAD^2 >/dev/null; then
-    tag="sha-$(git rev-parse HEAD^2)"
-else
-    tag="sha-$(git rev-parse HEAD)"
+    git reset --hard HEAD^2
 fi
+tag="sha-$(git rev-parse HEAD)"
 
 for try in $(seq 50 -1 0); do
     if fetch_to_cache tag "${tag}"; then
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -44,6 +44,40 @@ if [ "${NODE_ENV-}" = "development" ]; then
     exit 1
 fi
 
+# Here we find "base commit": the commit in our local tree which is equal to
+# the HEAD of the PR at the time the jumpstart was built.  This might not be
+# the commit that the jumpstart was actually built against (in case the PR
+# wasn't fully-rebased and GitHub merged some commits from main).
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
I think there's basically three "simple" approaches here, with different degrees of complexity:

 1. touch the files that changed in working tree vs. the cache build tree
 2. do as this PR, and delete the dist/ dirs that depend on changed files
 3. do the opposite: only unpack the dirs after we know they didn't depend on changed files

I think 3 is definitely more beautiful, but much harder to implement.  Maybe it's worth it.  I also thought the `touch` approach was actually kinda OK, since it's unlikely we end up touching built files (since they're ignored by git), but I have to admit that it's not very "clean".

tools/webpack-jumpstart
```diff
@@ -44,18 +47,29 @@ if [ "${NODE_ENV-}" = "development" ]; then
     exit 1
 fi
 
-if ! git diff --quiet -- ':^test' ':^packit.yaml' ':^.github'; then
-    echo 'jumpstart: uncommitted local changes, skipping download' >&2
-    exit 1
```
martinpitt, line 49: (Copied from #16158)

I really don't like dropping this check unconditionally. Now downloading the cache would claim to work even with local modifications, but then do that in vain and rebuild stuff anyway. To me as a developer this is too mushy and unpredictable -- either the cache applies to my tree or it doesn't. In the latter case, rebuilding webpacks is not really the end of the world.

I see two possible compromises:

 - Move this into the `else` branch  for the non-`--best` case
 - Drop the hardcoded "known safe" list (which I suppose is the one that annoys you?) and replace it with some `--ignore-local` or `--force` option. That makes stuff a wee bit harder for devs, but not unreasonably so.

tools/webpack-jumpstart
```diff
@@ -107,10 +107,16 @@ if [ "$(git rev-parse HEAD^{tree})" != "${target_tree}" ]; then
...
+# At this point, the only way that the working tree is different from
+# ${target_tree} is if --best was used.
+
 tools/node-modules make_package_lock_json
 unpack_from_cache "${tag}"
 
 for file in $(git diff --name-only "${target_tree}"); do
-    message TOUCH "${file}"
-    touch "${file}"
+    for match in $(grep -slP "^\t${file} \\\\$" dist/*/Makefile.deps); do
```
martinpitt, line 117: This is some serious duplication of `make` rules and logic. I'd rather not want this to happen without `--best`. Can you please conditionalize the whole paragraph on `[ -n `${best} ]`?

tools/webpack-jumpstart
allisonkarlitskaya: From this I understand that you're trying to say:

> After running jumpstart, without --best, typing `make` should never ever run webpack.  In the case that running webpack would be required, I'd rather have the jumpstart fail.

I think that's a reasonable property, and you're right that it's not currently the case.  I had intended to add a blanket check for changes in the local tree to prevent that unless `--best` was given (but your option 2), but I have another idea for a third option:

 - do everything, and in the case that any `dist/` directory would be missing, then fail.

That gives us a very very granular control over exactly which files the user is allowed to have modified (only the exact files which wouldn't result in rebuilding) and removes any arbitrary "this should be okay" type list-making.

It's also a pain in the butt to implement.  I'll let it simmer on the back burner...

allisonkarlitskaya: btw: in accordance with the above, I consider that maybe `--best` should be called `--partial` or so.  fwiw, it was originally called `--scan`.  I'm open to suggestions.

martinpitt: I like `--partial`, I think it denotes the intention pretty well.

tools/webpack-jumpstart
allisonkarlitskaya: I got rid of this approach: we just touch all the files now.

tools/webpack-jumpstart
allisonkarlitskaya: I changed the name to `--partial`.

allisonkarlitskaya: I also made the check *stronger*.  It will reject any changes at all to the local tree, unless `--partial` is given.

martinpitt (approved):
Nice work! The first commit that refuses jumpstarting with *any* local change feels like a step backwards -- a common use case (at least for me) is to have some local changes to test/ . But first of all, stashing them is easy enough, and the `--partial` mode certainly makes up for it :wink:

PR #16166, opened 2021-07-28
.gitignore: ignore /bots even if it's a symlink

allisonkarlitskaya:
Having a / at the end of the ignore pattern causes a symlink not to be
ignored.  For some time now, tools/make-bots can put a symlink there, so
ignore that too.


martinpitt (approved):
Nice catch, thanks!

PR #16167, opened 2021-07-28
Some small cockpit-session kerberos fixes

martinpitt:
They are prerequisites for the S4U stuff, but apply to the already existing GSS authentication as well. I want to test them separately, and they should be straightforward to land.

### DIFF
--- a/src/ws/session.c
+++ b/src/ws/session.c
@@ -366,8 +366,9 @@ static bool
 acquire_service_credentials (gss_OID mech_type, gss_cred_usage_t usage, gss_cred_id_t *cred)
 {
   /* custom credential store with our cockpit keytab */
-  static gss_key_value_element_desc store_elements[] = { { .key = "keytab", .value = COCKPIT_KTAB } };
-  static const gss_key_value_set_desc cockpit_ktab_store = { .count = 1, .elements = store_elements };
+  static gss_key_value_element_desc store_elements[] = { { .key = "keytab", .value = COCKPIT_KTAB, },
+                                                         { .key = "ccache", .value = "MEMORY:", } };
+  static const gss_key_value_set_desc cockpit_ktab_store = { .count = 2, .elements = store_elements };
   OM_uint32 major, minor;
 
   debug ("acquiring cockpit service credentials");
--- a/src/ws/session.c
+++ b/src/ws/session.c
@@ -378,9 +378,9 @@ acquire_service_credentials (gss_OID mech_type, gss_cred_usage_t usage, gss_cred
 
   if (GSS_ERROR (major))
     {
-      /* This is a routine error message, don't litter */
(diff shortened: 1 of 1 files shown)


src/ws/session.c
```diff
@@ -378,9 +378,9 @@ acquire_service_credentials (gss_OID mech_type, gss_cred_usage_t usage, gss_cred
 
   if (GSS_ERROR (major))
     {
-      /* This is a routine error message, don't litter */
       const char *msg = gssapi_strerror (mech_type, major, minor);
-      if (!strstr (msg, "nonexistent or empty"))
+      /* don't litter journal with expected error messages if keytab was not set up */
+      if (major != GSS_S_UNAVAILABLE && major != GSS_S_NO_CRED)
```
martinpitt, line 383: Actually I only observed GSS_S_NO_CRED "in the wild" when /etc/cockpit/krb5.keytab does not exist. So maybe we should restrict it to the latter. Indeed [the docs](https://docs.oracle.com/cd/E19683-01/816-1331/reference-4/index.html) say that `GSS_S_UNAVAILABLE` means "The operation or option is unavailable" which is certainly unexpected.

martinpitt: Fixed.

martinpitt: Bwah, the original string [still appears in our tests on older OSes](https://logs.cockpit-project.org/logs/pull-16167-20210728-084740-be421b0d-ubuntu-stable/log.html#154). But there it is `GSS_S_FAILURE`, with a totally incomprehensible "minor" code of 0o22661635265. I'll put back the string check for the old releases.

src/ws/session.c
```diff
@@ -362,6 +362,30 @@ map_gssapi_to_local (gss_name_t name,
...
+  debug ("acquiring cockpit service credentials");
+  major = gss_acquire_cred_from (&minor, GSS_C_NO_NAME, GSS_C_INDEFINITE, GSS_C_NO_OID_SET, usage,
+          (!getenv ("COCKPIT_TEST_KEEP_KTAB") && access (COCKPIT_KTAB, F_OK) == 0) ? &cockpit_ktab_store : NULL,
+          cred, NULL, NULL);
+
+  if (GSS_ERROR (major))
+    {
+      /* This is a routine error message, don't litter */
+      const char *msg = gssapi_strerror (mech_type, major, minor);
+      if (!strstr (msg, "nonexistent or empty"))
```
Gundersanne: I wonder if there's a better way to filter this error, considering gssapi has error codes. But it's just for debug messages so whatever.

Gundersanne (approved):
LGTM!

src/ws/session.c
Gundersanne: Sorry, you did this in the last commit. Marking as resolved.

PR #16168, opened 2021-07-28
ws: Create S4UProxy user ticket on tlscert authentication

martinpitt:
Turn the Proof of Concept in check-system-s4u-ssh into cockpit-session
code to create an S4U ticket for the user in tlscert authentication
mode. This will be used in the future to further authenticate to sudo or
ssh, but this requires further configuration. In the default FreeIPA
configuration, these tickets are not trusted by anything, so it is
harmless to create them by default.

The testClientCertAuthentication in TestIPA covers the successful case,
in TestAD the "not available" case (no unexpected error messages), and
in TestLogin the case where the machine is not joined to a domain.

Adjust check-system-s4u-ssh to acquire the S4U ticket from cockpit
through a TLS certificate login, instead of the custom Python script.

-----

Documentation preview is on https://piware.de/tmp/guide-s4u . In particular, look at the [SSO server setup](https://piware.de/tmp/guide-s4u/sso.html#sso-server) and the [Authentication to other services](https://piware.de/tmp/guide-s4u/cert-authentication.html#certauth-forwarding) parts.

 - [x] Builds on top of PR #16167 
 - [x] Builds on top of PR #16172
 - [x] Test authentication to sudo
 - [x] Test authentication to ssh
 - [x] Add documentation

## Certificate/smart card authentication to ssh and sudo

Once you logged into Cockpit with a certificate, you likely need to switch to administative mode, or connect to remote machines through SSH. If your machine is part of a FreeIPA domain, and your user account does not have a password, you can now set up [constrained delegation rules](https://www.freeipa.org/page/V4/Service_Constraint_Delegation) so that the Cockpit certificate authentication gets forwarded to sudo and/or ssh. Please see the [Client certification documentation](https://cockpit-project.org/guide/latest/cert-authentication.html) for details.

### DIFF
--- a/doc/guide/cert-authentication.xml
+++ b/doc/guide/cert-authentication.xml
@@ -202,4 +202,83 @@ CPUQuota=30%
 
   </section>
 
+  <section id="certauth-forwarding">
+    <title>Authentication to other services like sudo and ssh</title>
+
+    <para>Once you logged into Cockpit with a certificate, you likely need to switch to administative mode
+      (root privileges through sudo), or connect to remote machines through SSH. If your user account has a password,
+      that can be used for authenticating to sudo or ssh as usual.</para>
+
+    <para><emphasis>Supported with FreeIPA only:</emphasis> As an alternative to password authentication, you can
+      also declare the initial Cockpit certificate authentication as trusted for authenticating to SSH,
+      sudo, or other services. For that purpose, Cockpit automatically creates an
+      <ulink url="https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-sfu/bde93b0e-f3c9-4ddf-9f44-e1453be7af5a">S4U2Proxy Kerberos ticket</ulink>
+      in the user session:
+    </para>
+
+<programlisting>
+$ klist
+Ticket cache: FILE:/run/user/1894000001/cockpit-session-3692.ccache
+Default principal: user@EXAMPLE.COM
+
+Valid starting     Expires            Service principal
+07/30/21 09:19:06  07/31/21 09:19:06  HTTP/myhost.example.com@EXAMPLE.COM
+07/30/21 09:19:06  07/31/21 09:19:06  krbtgt/EXAMPLE.COM@EXAMPLE.COM
+	for client HTTP/myhost.example.com@EXAMPLE.COM
+</programlisting>
+
+    <para>You can set up <ulink url="https://www.freeipa.org/page/V4/Service_Constraint_Delegation">constrained delegation rules</ulink>
+      to enumerate which hosts (including its own) that ticket is trusted to access. For example, if the cockpit session runs on host
+      <code>myhost.example.com</code> and should be trusted to access its own host (through sudo) and another host
+      <code>remote.example.com</code> (through ssh), create a delegation like this:
+    </para>
+
+<programlisting>
+# a list of target machines which can be accessed by a particular rule
+ipa servicedelegationtarget-add cockpit-target
+ipa servicedelegationtarget-add-member cockpit-target \
+  --principals=host/myhost.example.com@EXAMPLE.COM \
+  --principals=host/remote.example.com@EXAMPLE.COM
+
+# allow cockpit sessions (HTTP/ principal) to access that host list
+ipa servicedelegationrule-add cockpit-delegation
+ipa servicedelegationrule-add-member cockpit-delegation \
+  --principals=HTTP/myhost.example.com@EXAMPLE.COM
+ipa servicedelegationrule-add-target cockpit-delegation \
+  --servicedelegationtargets=cockpit-target
+</programlisting>
+
+    <para>In addition, you need to enable GSS (Kerberos) authentication in the corresponding services.</para>
+
+    <itemizedlist>
+      <listitem>
+        <para>For SSH, enable <code>GSSAPIAuthentication yes</code> in
+          <ulink url="https://linux.die.net/man/5/sshd_config">/etc/ssh/sshd_config</ulink>.</para>
+      </listitem>
+
+      <listitem>
+        <para>For sudo, enable <code>pam_sss_gss</code> as described in the
+          <ulink url="https://www.mankier.com/8/pam_sss_gss">manpage</ulink>:
+          In <code>/etc/sssd/sssd.conf</code>: Add an entry for your domain:</para>
+
+<programlisting>
+[domain/example.com]
+pam_gssapi_services = sudo, sudo-i
+</programlisting>
+
+        <para>In <code>/etc/pam.d/sudo</code>, enable the module in the first line:</para>
+
+<programlisting>
+auth sufficient pam_sss_gss.so
+</programlisting>
+
+      </listitem>
+    </itemizedlist>
+
+    <para><emphasis>Caveat:</emphasis> The delegated S4U ticket is not yet forwarded to remote SSH
+    hosts when connecting to them from Cockpit, so authenticating to sudo on the remote host with
+    that ticket does not work. This will be provided in a future version.</para>
+
+  </section>
+
 </chapter>
--- a/src/ws/session.c
+++ b/src/ws/session.c
@@ -366,9 +366,11 @@ static bool
 acquire_service_credentials (gss_OID mech_type, gss_cred_usage_t usage, gss_cred_id_t *cred)
 {
   /* custom credential store with our cockpit keytab */
-  static gss_key_value_element_desc store_elements[] = { { .key = "keytab", .value = COCKPIT_KTAB, },
-                                                         { .key = "ccache", .value = "MEMORY:", } };
-  static const gss_key_value_set_desc cockpit_ktab_store = { .count = 2, .elements = store_elements };
+  gss_key_value_element_desc store_elements[] = {
+      { .key = (usage == GSS_C_INITIATE) ? "client_keytab" : "keytab", .value = COCKPIT_KTAB, },
+      { .key = "ccache", .value = "MEMORY:", }
+  };
+  const gss_key_value_set_desc cockpit_ktab_store = { .count = 2, .elements = store_elements };
   OM_uint32 major, minor;
 
   debug ("acquiring cockpit service credentials");
@@ -380,8 +382,10 @@ acquire_service_credentials (gss_OID mech_type, gss_cred_usage_t usage, gss_cred
     {
       const char *msg = gssapi_strerror (mech_type, major, minor);
       /* don't litter journal with error message if keytab was not set up, as that's expected */
-      if (major != GSS_S_NO_CRED && !strstr (msg, "nonexistent or empty"))
-        warnx ("couldn't acquire server credentials: %s", msg);
+      /* older krb versions hide the interesting bits behind the generic GSS_S_FAILURE and an uncomparable
+       * minor code, so do string comparison for these */
+      if (major != GSS_S_NO_CRED && !strstr (msg, "nonexistent or empty") && !strstr (msg, "No Kerberos credentials available"))
+        warnx ("couldn't acquire server credentials: %o %s", major, msg);
       return false;
     }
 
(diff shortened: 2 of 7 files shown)


src/ws/session.c
```diff
@@ -366,9 +366,11 @@ static bool
 acquire_service_credentials (gss_OID mech_type, gss_cred_usage_t usage, gss_cred_id_t *cred)
 {
   /* custom credential store with our cockpit keytab */
-  static gss_key_value_element_desc store_elements[] = { { .key = "keytab", .value = COCKPIT_KTAB, },
-                                                         { .key = "ccache", .value = "MEMORY:", } };
-  static const gss_key_value_set_desc cockpit_ktab_store = { .count = 2, .elements = store_elements };
+  gss_key_value_element_desc store_elements[] = {
+      { .key = (usage == GSS_C_INITIATE) ? "client_keytab" : "keytab", .value = COCKPIT_KTAB, },
```
martinpitt, line 370: I'm not fully happy with this. Either let's turn `usage` into a `bool is_client`, or explicitly pass the keytab type as an argument.

test/verify/check-system-realms
```diff
@@ -621,6 +633,35 @@ ipa user-add-cert alice --certificate="%(cert)s"
...
+        self.assertIn("csrf-token", out)
+        ccache_env = m.execute("xargs -0n1 < /proc/$(pgrep cockpit-bridge)/environ | grep KRB5CCNAME=").strip()
+        out = m.execute(f"su -c '{ccache_env} klist' alice")
+        self.assertIn("Ticket cache: FILE:/run/user/", out)
+        self.assertIn("Default principal: alice@COCKPIT.LAN", out)
+        self.assertIn("for client HTTP/x0.cockpit.lan@COCKPIT.LAN", out)
+
+        # does not interfere with persistent ticket in other sessions
+        self.assertEqual(m.execute("su -c klist alice"), persistent_ticket)
+        # S4U proxy ticket gets cleaned up on logout
```
martinpitt, line 657: add new line before this

test/verify/check-system-s4u-ssh
```diff
@@ -88,7 +88,20 @@ class TestS4USsh(MachineCase):
         """
         ipa_machine.execute("podman exec freeipa bash -ec '%s'" % script)
 
-        # ssh client, get keytab for authenticating and make ccache using gssapi
+        # enable gssapiauth for ssh
+        sshd_machine.execute("sed -ri 's/#GSSAPIAuthentication (yes|no)/GSSAPIAuthentication yes/' /etc/ssh/sshd_config")
+        sshd_machine.execute("systemctl daemon-reload && systemctl restart sshd")
```
martinpitt, line 99: Doesn't need daemon-reload (was already in the old code, but let's clean it up)

martinpitt (commented):
self-review

src/ws/session.c
martinpitt: On second thought, a boolean is hard to read in calls, and it's not actually bad to centralize the knowledge of how to map usages to config keys here. Let's see what other reviewers say.

src/ws/session.c
```diff
@@ -533,6 +537,56 @@ pam_conv_func_dummy (int num_msg,
...
+  gss_buffer_desc user_buf = { .length = strlen (username), .value = (char *) username };
+  gss_cred_id_t server_cred = GSS_C_NO_CREDENTIAL;
+  gss_name_t impersonee = GSS_C_NO_NAME;
+
+  res = getpwnam_r (username, &pwd_buf, pwd_string_buf, sizeof pwd_string_buf, &pwd);
+  if (pwd == NULL)
+    {
+      warnx ("couldn't load user info for: %s: %s", username, res == 0 ? "not found" : strerror (res));
+      goto out;
+    }
```
allisonkarlitskaya, line 557: Why?  If we're convinced already that we want to issue a ticket for this username then surely we already know that it exists... It's not like you use any of the return values for anything...

src/ws/session.c
```diff
@@ -556,6 +609,10 @@ perform_tlscert (const char *rhost)
 
   res = open_session (pamh);
 
+  create_s4u_ticket (username);
```
allisonkarlitskaya, line 612: I think we need to talk about this a bit more explicitly: why not enable this also for password authentication or even Kerberos authentication?

allisonkarlitskaya (changes requested):
Two immediate comments after quickly skimming.

src/ws/session.c
martinpitt: Oops, yes -- that's a leftover from an ealier approach, will remove.

src/ws/session.c
martinpitt: For Kerberos, no -- we already create a TGT there, which is much more useful (and *much* easier to use/configure) than the S4U one. For password auth, we could do this (at least if the machine is in a FreeIPA domain). I didn't do it here for the principle of least surprise, as we already use passwords to auth to sudo (and promote SSH keys), but it wouldn't hurt I guess.

martinpitt: Sorry, I remember now: For password auth we already (implicitly) create a TGT, that's just how domain-managed users/PAM work:

image

So an S4U one would again be a downgrade, and thus pointless IMHO.

src/ws/session.c
martinpitt: Dropped.

Gundersanne (dismissed):
Sad that the error filtering is quirky, but lgtm :)

martinpitt:
I reworked this quite a bit to also test sudo, and figure out a way to get an indefinitively long-lived session with curl.

martinpitt:
SSH [fails on debian-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16168-20210802-164116-7fc21784-debian-stable/log.html#279) and [ubuntu-2004](https://logs.cockpit-project.org/logs/pull-16168-20210802-164116-7fc21784-ubuntu-2004/log.html#279) because older SSH versions apparently fail to map the `x0` name to the fully qualified `host/x0.cockpit.lan` principal. Giving the full host name helps. Pushing that fix (locally tested on debian-stable only).

The [fedora-33 failure](https://logs.cockpit-project.org/logs/pull-16168-20210802-164116-7fc21784-fedora-33/log.html#279) is something else, I'll investigate that separately.

martinpitt:
I reworked the testing to be at *least* 100% cooler now :sunglasses: It's now running a real cockpit session with the real browser with cert authentication -- at this point, another bucket of :heart: to whoever implemented `socat`!

test/verify/check-system-realms
```diff
@@ -609,7 +622,7 @@ ExecStart=/bin/true
         alice_cert = '\n'.join([line for line in alice_cert.splitlines() if not line.startswith("----")])
         # set up an IPA user with a TLS certificate; can't use "admin" due to https://pagure.io/freeipa/issue/6683
         ipa_machine.execute(r"""podman exec -i freeipa sh -exc '
-ipa user-add --first=Alice --last="Developer" alice
+ipa user-add --first=Alice --last="Developer" --shell=/bin/bash alice
```
martinpitt, line 628: FTR, this is because the Terminal check below waits for an "alice" to appear in the prompt. With the default /bin/sh shell, the prompt is just `$`. I think it's more realistic to verify this with bash.

src/ws/session.c
allisonkarlitskaya: I guess we aren't super interested in the scenario where the user has a different password for krb than they do on the local machine.  That's kinda missing the point of what this is all supposed to be about...

This is cool: `KCM:256000500:68970`.  Any way we can steal that for our own cache instead of the tmpfile business?

allisonkarlitskaya:
The krb5 PAM module goes out of its way to check the krb config to see where it creates the ticket cache, and puts it in the `KCM:` if that's been configured.  Our code currently just always writes a file in the runtime dir.  That's not new, though: we introduced that in #16149 (which is where we might have had this discussion earlier), so it's not directly related to this PR.  We've agreed to fix this in a followup PR.

doc/guide/cert-authentication.xml
```diff
@@ -202,4 +202,83 @@ CPUQuota=30%
...
   </section>
 
+  <section id="certauth-forwarding">
+    <title>Authentication to other services like sudo and ssh</title>
+
+    <para>Once you logged into Cockpit with a certificate, you likely need to switch to administative mode
+      (root privileges through sudo), or connect to remote machines through SSH. If your user account has a password,
+      that can be used for authenticating to sudo or ssh as usual.</para>
+
+    <para><emphasis>Supported with FreeIPA only:</emphasis> If the user account does not have a
```
allisonkarlitskaya, line 212: You can also do this even if the account does have a password, no?  Maybe just write it like "As an alternative to password authentication, you can also..."

src/ws/session.c
```diff
@@ -533,6 +537,45 @@ pam_conv_func_dummy (int num_msg,
   return PAM_CONV_ERR;
 }
 
+static void
+create_s4u_ticket (const char *username)
+{
+  OM_uint32 major, minor;
+  gss_buffer_desc user_buf = { .length = strlen (username), .value = (char *) username };
```
allisonkarlitskaya, line 544: Can you move this down to where it's actually used?  I find it difficult to understand the flow of information when it's all at the top like this...  Also: could this be `const`?

src/ws/session.c
```diff
@@ -533,6 +537,45 @@ pam_conv_func_dummy (int num_msg,
...
+  if (GSS_ERROR (major))
+    {
+      warnx ("Failed to import user name %s: %s", username, gssapi_strerror (GSS_C_NO_OID, major, minor));
+      goto out;
+    }
+
+  /* store credentials into global creds; they will be put into the session and cleaned up in main() */
+  major = gss_acquire_cred_impersonate_name (&minor, server_cred, impersonee,
+                                             GSS_C_INDEFINITE, GSS_C_NO_OID_SET, GSS_C_INITIATE,
+                                             &creds, NULL, NULL);
```
allisonkarlitskaya, line 563: Another nag for a future cleanup: this global `creds` thing needs to die.  We should be passing that around explicitly.

allisonkarlitskaya (approved):
Seems pretty reasonable, except for a bunch of existing issues.

I would have hoped to see everything krb-related moved to a separate file by now, but that can come later...

I'm actually happy to deal with everything else in a followup.

doc/guide/cert-authentication.xml
martinpitt: Right. Thanks for the suggestion, applied.

src/ws/session.c
martinpitt: Moved down the three declarations.

I'm afraid it can't be const:
```
src/ws/session.c:553:36: error: passing argument 2 of ‘gss_import_name’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  553 |   major = gss_import_name (&minor, &user_buf, GSS_KRB5_NT_PRINCIPAL_NAME, &impersonee);
      |                                    ^~~~~~~~~
In file included from src/ws/session.c:28:
/usr/include/gssapi/gssapi.h:565:5: note: expected ‘gss_buffer_t’ {aka ‘struct gss_buffer_desc_struct *’} but argument is of type ‘const gss_buffer_desc *’ {aka ‘const struct gss_buffer_desc_struct *’}
  565 |     gss_buffer_t,       /* input_name_buffer */
      |     ^~~~~~~~~~~~
```

src/ws/session.c
martinpitt: Agreed, this is a good follow-up thing. added to my TODO list.

src/ws/session.c
martinpitt: Ah, moving all three down [fails](https://github.com/cockpit-project/cockpit/pull/16168/checks?check_run_id=3299309971) -- takes a moment to realize why, but it makes sense. So I only moved down `user_buf`.

allisonkarlitskaya (approved):

Issue #13882, opened 2020-04-09
Don't show successful last login in alert but in system information


garrett:
I don't think it should be a dismissable notification. You're not being notified of anything. It is a piece of static information about the state of the system (in this case, limited to a specific user).

Right now, as a notification, it competes with MOTD, which is also a dismissable notification:

![Screenshot_2020-04-09 Overview - Rain-2](https://user-images.githubusercontent.com/10246/78902573-41fe3900-7a7a-11ea-80ae-5d9023a282ff.png)

And, it's also quite weird paired with the sudo branch too:

![Screenshot_2020-04-08 Overview - localhost localdomain](https://user-images.githubusercontent.com/10246/78902706-6d812380-7a7a-11ea-80c5-9ca1f1e76a61.png)

(Although this has been changed in the sudo branch.)

It may make sense to move it to system information and show the session and session type as well.

Perhaps something like this:

**Previous log in**: garrett 19 Apr 12:34 PM on web console

- Drop year, if the same as current.
- Drop seconds.
- Use locale.
- Say "previous" instead of "last", as this isn't necessarily the last log in to the system, it's the previous time the same user was logged in... right?
- The previous log in can be linked with a popup or page that shows the last log in times (multiple), similar to running "last" (or a similar command) in a terminal.

garrett:
This all said, if there were _failed_ logins, those should be shown with a notification. That's a possibly important thing to show as a notification, but not a normal login.

KKoukiou:
@garrett @marusak so to understand this properly:

* we always show the last login in the system card
* we show when there are failed login attempts above in an alert card

Since these two information are quite connected, shouldn't we keep them together? I guess in the system card, but I feel we should not split them. Of course having a alert icon or something for failed login attemts is important. 

In matters of "Previous' vs "Last" wording, I think both are problematic, but previous is even more problematic.

Previous is contradictory with the fact that the user can be still logged in. In this case, of a currently connected user, the 'previous' is a misleading term.
We can use 'Latest' which is the term that lastlog is using as well, but that's the same as "Last" actually. So basically we are refering to the latest login for this specific user, not for the whole system. And for this reason 'Last" is not really wrong.

marusak:
> Target here is:

Yes. The thinking is that last successful login is not really important enough to be in alert box. Therefore we can just drop it into system card. But failed attempts are something that you should be notified, thus it should be shown in alert box.

Whether just icon would do, I'll keep that for Garrett.

[...]

PR #16169, opened 2021-07-28
Make switching to admin access more visible

### LINKED ISSUE #14305: navigation: permission switcher has low discoverability rate
Cockpit version: 220
OS: RHEL
Page: Navigation

During the usability testing in June with 5 participants, 2 participants failed the task, and 2 needed assistance to complete the task of putting the system in limited access mode.

Screenshot from 2020-06-30 18-02-27

garrett:
Fixes #14305.

Changes to the shell and overview alert.

### DIFF
--- a/pkg/systemd/superuser-alert.jsx
+++ b/pkg/systemd/superuser-alert.jsx
@@ -24,6 +24,8 @@ import {
     Alert, Button
 } from '@patternfly/react-core';
 
+import { LockIcon } from '@patternfly/react-icons';
+
 import { SuperuserDialogs } from "../shell/superuser.jsx";
 
 const _ = cockpit.gettext;
@@ -50,7 +52,8 @@ export class SuperuserAlert extends React.Component {
             <>
                 <Alert className="ct-limited-access-alert"
                        hidden={this.superuser.Current != "none"}
-                       variant="info" isInline
+                       variant="warning" isInline
+                       customIcon={<LockIcon />}
                        actionClose={actions}
                        title={_("Web console is running in limited access mode.")} />
             </>);
--- a/pkg/shell/nav.scss
+++ b/pkg/shell/nav.scss
@@ -271,6 +271,18 @@ $desktop: $phone + 1px;
 
     .dropdown-menu {
         width: auto !important;
+
+        .pf-c-button {
+            // Set position relative to the menu item
+            position: relative;
+
+            // Align icons midway in the left gutter
+            svg {
+                position: absolute;
+                left: calc((var(--mobile-menu-padding-width) - var(--pf-global--icon--FontSize--sm)) / 2);
+                top: calc((100% - var(--pf-global--icon--FontSize--md))/2);
+            }
+        }
     }
 
     .area-ct-subnav,
@@ -323,9 +335,10 @@ $desktop: $phone + 1px;
         .dropdown-menu > li > a,
         .dropdown-menu > li > button,
         .dropdown-menu > li > span {
+            --mobile-menu-padding-width: var(--pf-global--spacer--2xl);
             /* Make menu items a consistent, touch-friendly size */
-            padding: 0 40px;
-            line-height: 54px;
+            padding: 0 var(--mobile-menu-padding-width);
+            line-height: var(--pf-global--spacer--2xl);
             color: #333;
         }
     }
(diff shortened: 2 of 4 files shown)


garrett:
I goofed up the breakpoint stuff when I dropped some commits and squashed otehrs. I'll fixup, force-push, and mark as ready for review after. Aside from that, it's good enough to look at. I'll post screenshots when I have this fixed too. 

It looks a lot like the mockup, but I decided switching the alert border is probably a little too much. I kept everything else I mocked up, however.

garrett:
Screenshot:

Screenshot 2021-07-28 at 15-27-16 Overview - garrett Bolt

Whoops. I forgot to re-check mobile:

Screenshot 2021-07-28 at 15-28-38 Virtual Machines - garrett Bolt

garrett:
Shell's mobile menu is now fixed, with icons midway in the gutter area:

Screenshot 2021-07-28 at 16-08-26 Virtual Machines - garrett Bolt

pkg/systemd/superuser-alert.jsx
```diff
@@ -50,7 +52,8 @@ export class SuperuserAlert extends React.Component {
             <>
                 <Alert className="ct-limited-access-alert"
                        hidden={this.superuser.Current != "none"}
-                       variant="info" isInline
+                       variant="warning" isInline
```
KKoukiou, line 55: I did not take a look in the CSS changes yet, but I am not convinced that it's correct to show a 'warning' alert for not privileged logins..

martinpitt: I'm a bit torn here. On one hand being not root is the expected mode for SSH/desktop sessions, and some users are not even allowed to become root (there is no way to predict that). OTOH being root is the main use case for cockpit..

But even if we dial this back to info, I find all the other changes really excellent! Thanks @garrett!

garrett: It doesn't have a warning icon, but a custom one, however. 

The blue alert with another blue alert (login, which should _not_ be an alert, but currently is) is an excellent way to completely ignore the alert, which is part of the problem.

The "warning" is that Cockpit will not work the way you expect, if you generally use Cockpit as an admin, which most people do. It can be ignored by going to other pages.

Right now, Cockpit is quite broken in several pages when you're not the admin. (Pages don't let you know and may be blank or partially blank without any fallback message.) So it is indeed a warning and important to know to switch (if you can). We should also fix these pages, but it's still important regardless, as it will forever be important to be an admin to perform admin tasks.

garrett:
Note: I did consider a PF banner component (which is relatively new), but it even states:

> For use cases requiring more than a single line of content or requiring important actions for the user to take, the flow should make use of the [inline alert](https://www.patternfly.org/v4/components/alert/design-guidelines#inline-alerts) or [hint pattern](https://www.patternfly.org/v4/components/hint/design-guidelines) instead, as they allow for more flexibility around these use cases.

And then I considered hint:

> A hint is in-app messaging that provides a one-step reminder, explanation, or call to action for a page or modal.

> Use hints to share information about an interaction or prerequisite step that might not be immediately obvious to the user.

A hint sounds about right from the description, but doesn't really fit our overview page modal. It would have to show up where the _system_ alerts and MOTD ([and login info](https://github.com/cockpit-project/cockpit/issues/13882)?!???!) shows up. Everything in that area should be system-related, not session.

Hence, I kept it as an undismissable alert at the top as-is. It lets us change the color to indicate something is important, use an icon to grab attention and reinforce the top-right action in the shell (using the same lock icon and color when it's applicable), and make the action obvious with a button. These are all things that banner and hint components don't let us do in one way or another.

pkg/systemd/superuser-alert.jsx
garrett: I'd like to point out that we had people in our UX study from a while back who missed not just the top shell area, but also the notification. It wasn't eye-catching enough. Part of that is the understated blue color (including its repetition with the log in "alert").

KKoukiou: Thanks for the detailed answer. Since we only borrowed the 'warning' color and not the icon as well, I think indeed we are walking on the middle ground of raising attention but not really marking it as a warning. I like it, thanks for this improvement.

KKoukiou (approved):

martinpitt:
@garrett : I updated the screenshot for the "Session" icon change, and cropped it. I also updated it to include the "Session" menu as well, as it's the same screenshot, and both affect the shell panel. I also added crediting @Verhoeckx. Do you mind proofreading this?

PR #16174, opened 2021-07-30
non-controverial tls refactoring

allisonkarlitskaya:
Here's the first few commits from #16000, as requested.  Those should be easy to land.


martinpitt (approved):
Thanks! (I had expected even more commits, as they also shouldn't break anything, but indeed let's spoon-feed this to the bots).

PR #16175, opened 2021-07-30
mehr tls Gedöns

allisonkarlitskaya:
More prep towards #16000.

These ones might need a bit more discussion


src/ws/session.c
```diff
@@ -534,15 +534,22 @@ pam_conv_func_dummy (int num_msg,
...
 {
   struct pam_conv conv = { pam_conv_func_dummy, };
+  const char *client_certificate;
   pam_handle_t *pamh;
   int res;
 
   debug ("start tls-cert authentication for cockpit-ws %u", getppid ());
 
-  char *username = cockpit_session_client_certificate_map_user ();
+  if (strncmp (authorization, "tls-cert ", 9) == 0)
```
martinpitt, line 547: This is already done by cockpit_authorize_type(), so this could even be turned into an assertion, and a hard crash.

src/ws/cockpit-session-client-certificate.c
```diff
@@ -333,8 +319,23 @@ cockpit_session_client_certificate_map_user (const char *client_certificate_file
...
     }
 
+  size_t my_cgroup_length;
+  char *my_cgroup = read_proc_self_cgroup (&my_cgroup_length);
+  if (my_cgroup == NULL)
+    {
+      warnx ("Could not determine cgroup of this process");
+      return NULL;
+    }
+  if (strncmp (cert_pem, my_cgroup, my_cgroup_length) != 0)
```
martinpitt, line 329: This works because both strings are expected to include the final `\n`, and because cgroups can't contain newlines. A comment for that wouldn't hurt here.

martinpitt (approved):
Thanks! Most of these already got two review rounds in #16000, so nothing serious any more. I'm fine with my two comments being addressed in a follow-up.

PR #16180, opened 2021-08-01
move webpack jumpstart build to `node:14`

allisonkarlitskaya:
From the sunday-trainrides-with-nothing-to-do department:

`node:14` is a pre-cached image on GitHub, so it's faster.  I had to fix a few problems with old git versions, but it seems to work now.

See
 - https://github.com/allisoninc/cockpit/pull/3
 - https://github.com/allisoninc/cockpit/pull/3/checks?check_run_id=3213790366

### DIFF
--- a/.github/workflows/webpack-jumpstart.yml
+++ b/.github/workflows/webpack-jumpstart.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - name: Build jumpstart tarball
         run: |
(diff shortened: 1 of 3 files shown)


martinpitt:
Restarted timed-out unit test. Packit is currently broken for unrelated reasons (persistent k8s volume running out of space), we can ignore that for this PR.

martinpitt (approved):
Thanks! That shaves off 1.5 minutes from the build, nice!

PR #16183, opened 2021-08-02
abrt: Update url to gnome API

### LINKED ISSUE #16179: Validate all URLs
Checked 137 URLs out of which 1 is/are invalid:
https://developer.gnome.org/gio/stable/gio-GDBusError.html#g-dbus-error-encode-gerror

marusak:
Fixes #16179

### DIFF
--- a/pkg/systemd/reporting.jsx
+++ b/pkg/systemd/reporting.jsx
@@ -295,7 +295,7 @@ class BusWorkflowRow extends React.Component {
                         /* GLib encodes errors for transport over the wire,
                          * but we don’t have a good way of decoding them without calling into GIO.
                          *
-                         * https://developer.gnome.org/gio/stable/gio-GDBusError.html#g-dbus-error-encode-gerror
+                         * https://developer-old.gnome.org/gio/stable/gio-GDBusError.html#g-dbus-error-encode-gerror
                          *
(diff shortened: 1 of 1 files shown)


pkg/systemd/reporting.jsx
```diff
@@ -295,7 +295,7 @@ class BusWorkflowRow extends React.Component {
                         /* GLib encodes errors for transport over the wire,
                          * but we don’t have a good way of decoding them without calling into GIO.
                          *
-                         * https://developer.gnome.org/gio/stable/gio-GDBusError.html#g-dbus-error-encode-gerror
+                         * https://developer-old.gnome.org/gio/stable/gio-GDBusError.html#g-dbus-error-encode-gerror
```
martinpitt, line 298: Ugh, how poor -- the link is broken on https://www.gnome.org/technologies/ as well, I was hoping for some better replacement than this (which looks like it's going to be decomissioned soon)

martinpitt (approved):
Danke!

pkg/systemd/reporting.jsx
marusak: Right, I am also not convinced this is the best solution as it seems it is not going to stick around for too long. But that is where you get if you just navigate from https://developer.gnome.org

PR #16184, opened 2021-08-02
test: Adjust for PackageKit 1.2.4 severity parsing support

martinpitt:
PackageKit 1.2.4 finally fixed a long standing bug [1] about mis-parsing
the "severity" field for an update. The current version now finally
returns correct results. Adjust the tests to factorize the "supports
severities" into a `self.supports_severity` flag, and make it dynamic
based on the PackageKit version.

Also adjust the old bugs.freedesktop.org reference to the current one
(the bug tracker got migrated to GitHub a while ago).

[1] https://github.com/PackageKit/PackageKit/issues/268

----

Blocks https://github.com/cockpit-project/bots/pull/2265

### DIFF
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -61,9 +61,13 @@ class TestUpdates(NoSubManCase):
     def setUp(self):
         super().setUp()
 
-        # only the yum backend properly recognizes "enhancement" severity; apt
-        # does not have that metadata and PackageKit-dnf does not parse it
-        if self.backend == "yum":
+        pkcon_version = self.machine.execute("pkcon --version").strip()
+
+        # only the yum and PackageKit ≥ 1.2.4 dnf (https://github.com/PackageKit/PackageKit/issues/268) backends
+        # properly recognize "enhancement" severity; apt does not have that metadata
+        self.supports_severity = not (self.backend == "apt" or pkcon_version < "1.2.4")
+
+        if self.supports_severity:
             self.enhancement_severity = "enhancement"
         else:
             self.enhancement_severity = "bug"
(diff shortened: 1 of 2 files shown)


martinpitt:
[This needs fixing](https://logs.cockpit-project.org/logs/pull-16184-20210802-083004-f50ffa0f-debian-stable/log.html#276). I have the fix locally, but I won't push it before https://github.com/cockpit-project/bots/pull/2267 lands.

martinpitt:
Today, COPR has [some intermittent problems with RPM signing](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-16184/fedora-34-x86_64/02356688-cockpit/backend.log.gz). 33 and centos8 are running, I'll retry if they also fail, otherwise ignore.

test/verify/check-packagekit
```diff
@@ -61,9 +61,13 @@ class TestUpdates(NoSubManCase):
...
         super().setUp()
 
-        # only the yum backend properly recognizes "enhancement" severity; apt
-        # does not have that metadata and PackageKit-dnf does not parse it
-        if self.backend == "yum":
+        pkcon_version = self.machine.execute("pkcon --version").strip()
+
+        # only the yum and PackageKit ≥ 1.2.4 dnf (https://github.com/PackageKit/PackageKit/issues/268) backends
+        # properly recognize "enhancement" severity; apt does not have that metadata
+        self.supports_severity = not (self.backend == "apt" or pkcon_version < "1.2.4")
```
marusak, line 68: Not the best way how to compare versions - but since this is just test, that will do :)

martinpitt: Right, I know -- my hope is that by the time we have an 1.2.10 (like in 7 years from now? :grin: ) we can just drop this special case completely.

marusak (approved):

PR #16185, opened 2021-08-02
shell: Change user icon in the top menu

### LINKED ISSUE #16173: shell: Session menu UI change
We could change the top-right to look like this:

cockpit-overview-session (copy)

I noticed that PF now suggests a filled circle for the heading, wheras the line version is for in-page help icons. This is new. We should update. I've updated it in this mockup. With the outline version that we're currently using, the cog looks out of place, but it fits right in with the solid version.

Screenshot 2021-07-29 at 14-18-13 PatternFly 4 • Icons

In context of the rest of the page, it would look like this:

cockpit-overview-session

---

One problem with the cog: We're using it in the metrics settings. This would mean two cogs right above each other. They'd each have different words, but it'd probably still be a little odd.

Here's a screenshot:

Screenshot 2021-07-29 at 14-19-39 Overview - garrett Bolt

And here's a modification to make it show what it would look like even with the stacking:

Screenshot 2021-07-29 at 14-19-39 Overview - garrett Bolt-mod

Additionally, someone might think the cog is a flower. But that's fine. Whatever, it's the session and settings flower then. :white_flower: :grin: 

If you have a suggestion for something else, PatternFly's icon library is browsable at: https://www.patternfly.org/v4/guidelines/icons/#all-icons

_Originally posted by @garrett in https://github.com/cockpit-project/cockpit/discussions/16154#discussioncomment-1086819_

marusak:
It is about 'Session' and not about user.

Fixes #16173

Screenshot from 2021-08-02 12-52-18s1

Screen Shot 2021-08-02 at 12 52 33

### DIFF
--- a/pkg/shell/index.html
+++ b/pkg/shell/index.html
@@ -51,7 +51,7 @@
           </div>
           <div id="navbar-docs" class="dropdown ct-header-item ct-header-docs">
             <a role="link" tabindex="0" id="navbar-docs-dropdown" class="ct-nav-toggle dropdown-toggle" data-toggle="dropdown">
-              <span class="pficon pficon-help"></span>
+              <span class="icon-custom pficon pficon-help"></span>
               <span translate>Help</span>
               <span class="fa fa-caret-down pf-c-dropdown__toggle-icon"></span>
             </a>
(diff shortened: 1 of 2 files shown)


garrett:
First glance: Looks great in the screenshots! I'm looking forward to testing this out on my machine. :grin:

garrett (approved):
LGTM :+1: 

I pushed a minor change with an admin and force push that removed the inline style inside the SVG that wasn't applied anyway.

martinpitt:
The release note is part of PR #16169

PR #16190, opened 2021-08-03
test: Add IdentitiesOnly to the ssh config example

jelly:
With the IdentitiesOnly option, ssh only offers the configured
IdentitiyFile. This option is useful for situations where the ssh agent
offers too many different identities causing an authentication failure.


martinpitt (approved):
Thank you! I added that to my own configuration, and works as expected.

I label this with "no-test", as this is a README-only fix which is not covered by integration tests, so the minimal set (fedora-34 only) is sufficient here, and we don't need the full blast of "all OSes/browsers". I still suggest you to look at the suite of tests that we run, to get some feeling what kind of testing we do.

PR #16197, opened 2021-08-04
test: Reset failed units in TestKerberos.testNegotiate terminal test

martinpitt:
Similarly to commit 7a6b68a8afc, reset failed systemd units. Otherwise
they appear as the first few lines in the terminal (through a bash
profile plugin) and break the test assumption of the prompt being in the
first line.

----

Seen in https://github.com/cockpit-project/bots/pull/2274 in [this failure](https://logs.cockpit-project.org/logs/pull-2274-20210803-145127-82795814-fedora-33-cockpit-project-cockpit/log.html#254-2)


marusak (approved):
Thanks!

mvollmer (approved):

PR #16198, opened 2021-08-04
Generalize ignoring test output files

jelly:
Ignore all Test file extensions in a general manner so for example
Test*.gz is not missing.


marusak (approved):
Thanks!

martinpitt:
Let's add a "gitignore:" commit message prefix with squash&merge.

PR #16199, opened 2021-08-04
HACKING: Link the mailing list address

jelly:
Linkify the cockpit-devel mailing list address to the mailman overview
page.

I had to search for the page to subscribe, so I think it's worth adding a direct link :)


martinpitt (approved):
Thanks!

PR #16204, opened 2021-08-05
Fix version 250 Coverity regressions, fix pycodestyle regressions

martinpitt:
It seems we really need to update our unit-tests container to pull in a recent pycodestyle.. I'm doing that in PR #16207


marusak (approved):
Thanks!

PR #16205, opened 2021-08-05
containers: Install static Python checkers from PyPI

martinpitt:
Debian stable's pyflakes and pycodestyle are too old, and are missing a
lot of issues that our local versions (rightfully) complain about.
Install them from PyPI instead.

-----

This ought to fail right now, as our current code has some pycodestyle errors. I fixed them in #16204. But let's validate that it does.

### DIFF
--- a/containers/unit-tests/setup.sh
+++ b/containers/unit-tests/setup.sh
@@ -39,9 +39,8 @@ dependencies="\
     npm \
     nodejs:amd64 \
     pkg-config \
-    pyflakes3 \
     python3 \
-    python3-pep8 \
+    python3-pip \
     sassc \
     ssh \
     strace \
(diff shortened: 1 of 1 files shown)


martinpitt:
So this [unexpectedly succeeded](https://github.com/cockpit-project/cockpit/pull/16205/checks?check_run_id=3249459729) (aside from a timeout). Which is weird, locally it fails:
```
❱❱❱ containers/unit-tests/start shell
[...]
builder@9ef75b0d60fe:/$ cd /source/
builder@9ef75b0d60fe:/source$ tools/test-static-code 
1..8
ok 1 pyflakes pkg tools
ok 2 pyflakes test
ok 3 js-translatable-strings
ok 4 unsafe-security-policy
ok 5 patternfly-font-paths # skip dist/ not built
ok 6 json-verify
test/verify/check-packagekit:153:44: E261 at least two spaces before inline comment
test/verify/check-packagekit:247:19: E261 at least two spaces before inline comment
test/verify/check-packagekit:270:5: E303 too many blank lines (2)
test/verify/check-packagekit:1084:1: E302 expected 2 blank lines, found 1
test/verify/check-packagekit:1116:1: E302 expected 2 blank lines, found 1
not ok 7 pycodestyle test
ok 8 node_modules freshness
builder@9ef75b0d60fe:/source$ echo $?
1
```

Moreover, I also get these errors with the *current* container, so it is *not* about pyflakes/pycodestyle being too old.

There is something deeply wrong here, investigating..

martinpitt:
Got it, will send a new PR for that.

PR #16206, opened 2021-08-05
doc: Document paths in cockpit.file

marusak:
We got in the last weeks two users who struggled with `cockpit.file()`
as they were trying to resolve files in relative paths.


martinpitt (approved):
Dakujem!

PR #16207, opened 2021-08-05
containers: Properly run test-static-code during unit tests

martinpitt:
Since commit 4077dee3d2f65b our unit test runs did not run pycodestyle
any more, as that started to skip the pycodestyle check in a non-git
tree (such as the unpacked tree that `distcheck` runs in). As
test-static-code does not run in `check-memory` either, none of the four
scenarios ended up running it.
    
So explicitly call it in the distcheck scenario, and relatively early so
that failures (which happen quite often) are easier to spot in the log.

----

Spotted in PR #16204 , which also has a commit to actually fix the current pycodestyle errors. Until that lands, this PR should fail.

Also fix the timeout in the workflow.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -11,7 +11,7 @@ jobs:
          - CC=clang distcheck
          - :i386 distcheck
       fail-fast: false
-    timeout-minutes: 20
+    timeout-minutes: 60
     env:
       # current podman backport falls over as non-root, and completely breaks down with :i386 image
       docker: docker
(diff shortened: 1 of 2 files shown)


martinpitt:
Very good, [now it fails](https://github.com/cockpit-project/cockpit/pull/16207/checks?check_run_id=3249849053) as expected. As soon as #16204 will land, I'll rebase this and it should become green again.

martinpitt:
#16204 landed, so this ought to work now. Rebased.

marusak (approved):
Danke!

Issue #15594, opened 2021-03-25
Progress bars jump when percentages are the label


garrett:
On at least the software page, possibly on other places, the progressbar width shifts as the numbers increment.

The _proper_ solution would be to use tabular numbers, but our font does not support that (yet). Related: https://github.com/cockpit-project/cockpit/issues/15090 & https://github.com/RedHatOfficial/RedHatFont/issues/27

Therefore, a work-around that we could do right now is to set `min-width: 3ch` on the label. As this would affect anything with a progress bar, we should probably have this as an override.

jelly:
For the "Software Updates" case I reproduced the issue using the `testUpdates.testBasic` test case and extending it with:

```python
+        for i in range(1, 35):
+            pkgname = "test{}".format(i)
+            self.createPackage(pkgname, "1.0", "1", install=True)
+            self.createPackage(pkgname, "1.0", "2", postinst="sleep 1;")
         # create two updates; force installing chocolate before vanilla
```

This makes the issue visible for me.

I also noticed the issue with the CPU/Memory graphs on the `Overview`

PR #16208, opened 2021-08-05
packagekit: don't make the progress bar jump

jelly:
The Red Hat font numbers are not all the same size which makes the
progress bar "jump" when the progress changes. As a workaround, the
label now reserves enough space so the progress bar now doesn't jump
around.

### DIFF
--- a/pkg/packagekit/updates.scss
+++ b/pkg/packagekit/updates.scss
@@ -137,6 +137,11 @@ div.changelog {
     margin: 10ex auto 0;
 }
 
+/* workaround font not supporting tabular numbers yet https://github.com/cockpit-project/cockpit/issues/15090 */
+.pf-c-progress__status {
+    min-width: 3ch;
+}
+
 /* don't let the install progress bar get too wide */
 .progress-cancel {
     display: flex;


jelly:
Should I add a note / reference to the general issue we have https://github.com/cockpit-project/cockpit/issues/15594

martinpitt:
Thanks! The way to make a link/note to #15994 is to add a "Fixes #" to the commit message. I just triggered the tests, so no need to push just for that -- we can do that on squash & merge. But please edit this PR's description to add the Fixes # -- this will already be enough to link the pages, and close the issue once this PR lands.

pkg/packagekit/updates.scss
```diff
@@ -137,6 +137,11 @@ div.changelog {
     margin: 10ex auto 0;
 }
 
+/* workaround font not supporting tabular numbers yet https://github.com/cockpit-project/cockpit/issues/15090 */
+.pf-c-progress__status {
```
martinpitt, line 141: At first I thought this needs to be more specific -- but we have two progress bars on the Updates page, and they both show percentages. So it's fine to apply this to both of them.

martinpitt (approved):

jelly:
> Thanks! The way to make a link/note to #15994 is to add a "Fixes #" to the commit message. I just triggered the tests, so no need to push just for that -- we can do that on squash & merge. But please edit this PR's description to add the Fixes # -- this will already be enough to link the pages, and close the issue once this PR lands.

Sure, I didn't add it as there are other instances where the "progress bar jumps", so wasn't sure if the ticket should be auto-closed.

PR #14618, opened 2020-09-14
maint: Stop including PF API files for fedora > 33, RHEL > 8 && debian testing

KKoukiou:
This also need to be done for debian - 2 (in adjust-for-release-file?).
Also there is cockpit.css API which is also not supposed to be used. Should we drop it now as well?

PR #16212, opened 2021-08-06
examples: Drop form-table-ct and cockpit.css usage

martinpitt:
Rewrite our examples to use a guttered Stack layout [1] instead of
the obsolete form-table-ct. This is not completely ideal, but good
enough for the examples -- and we really don't want to promote
cockpit.css usage any more.

cockpit.css does not contain this, but that by itself has been obsolete
for a long time. Replace that with a direct inclusion of PatternFly. As
v4 does not offer a bundled CDN download any more, add symlinks to the
copy in node_modules -- the examples are usually run out of the cockpit
tree anyway.

Fix long-running-process to use a PF4 button.

[1] https://www.patternfly.org/v4/layouts/stack


martinpitt:
pinger:

pinger

simple-package:
simple-package

long-running-process in main:
longrunning-main

long-running-process in this PR:
longrunning-pr

martinpitt:
After this lands, I'll send a follow-up to remove cockpit.css. I'll make double-sure that no external project uses this any more (we deprecated it years ago and never documented it).

martinpitt:
Wrt cockpit.css: We already deprecated this in PR #14618, and only ship it on Debian 10, Fedora ≤ 33, and Ubuntu ≤ 20.04. So I'd say, let's drop that for good once we stop supporting Fedora 33 (I hear Fedora 35 gets branched off next week, and Debian 11 releases this month, so not that long any more). I put it onto my todo list.

marusak (approved):
Thanks!

PR #16213, opened 2021-08-06
systemd: Correct typo HostName -> Hostname

### LINKED ISSUE #9909: transient host name not shown in the host switcher
When a server is using a transient host name, the host name is displayed correctly on the System page but on the top of the sidebar it just says localhost. You can reproduce this by unsetting the host name with ``hostnamectl set-hostname ''``. Attached is a screenshot of the problem.

screenshot

jelly:
The systemd hostnamed D-Bus API returns Hostname for the configured
system hostname. This issue is only visible when no static hostname is
configured as PrettyHostname and StaticHostname are then empty.

Fixes #9909

### DIFF
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -193,6 +193,10 @@ class TestSystemInfo(MachineCase):
         b.wait_in_text('#system_information_hostname_text', "Adventure Box (host1.cockpit.lan)")
         self.assertEqual(m.execute("hostname").strip(), "host1.cockpit.lan")
 
+        m.execute("hostnamectl set-hostname ''")
+        m.execute("hostnamectl set-hostname --transient 'mydhcpname'")
+        b.wait_in_text('#system_information_hostname_text', 'mydhcpname')
+
         b.logout()
         m.execute("chattr -i /etc/os-release && rm /etc/os-release")
         m.execute("rm /usr/lib/os-release || true")
--- a/pkg/shell/machines/machines.js
+++ b/pkg/shell/machines/machines.js
@@ -493,7 +493,7 @@ function Loader(machines, session_only) {
         if (!machine.color)
             overlay.color = machines.unused_color();
 
-        var label = props.PrettyHostname || props.StaticHostname;
+        var label = props.PrettyHostname || props.StaticHostname || props.Hostname;
         if (label && label !== machine.label)
             overlay.label = label;
 
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -182,7 +182,7 @@ class OverviewPage extends React.Component {
 
         const pretty_hostname = this.state.hostnameData.PrettyHostname;
         const static_hostname = this.state.hostnameData.StaticHostname;
-        let str = this.state.hostnameData.HostName;
+        let str = this.state.hostnameData.Hostname;
 
         if (pretty_hostname && static_hostname && static_hostname != pretty_hostname)
             str = pretty_hostname + " (" + static_hostname + ")";
@@ -250,7 +250,7 @@ class OverviewPage extends React.Component {
                     <PageSection className='ct-overview-header' variant={PageSectionVariants.light}>
                         <div className='ct-overview-header-hostname'>
                             <h1>
-                                {this.hostname_text() || ""}
+                                {this.hostname_text()}
                             </h1>
                             {this.state.hostnameData &&
                              this.state.hostnameData.OperatingSystemPrettyName &&
--- a/test/verify/check-shell-host-switching
+++ b/test/verify/check-shell-host-switching
@@ -135,6 +135,7 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
 
     def testBasic(self):
         b = self.browser
+        m1 = self.machines["machine1"]
         m2 = self.machines["machine2"]
         m3 = self.machines["machine3"]
 
@@ -148,6 +149,12 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
         self.wait_host_addresses(b, ["localhost"])
 
         b.wait_not_present("button:contains('Edit hosts')")
+
+        # Test that transient hostname shows up
+        m1.execute("hostnamectl set-hostname ''")
+        m1.execute("hostnamectl set-hostname --transient 'mydhcpname'")
+        b.wait_in_text("#nav-hosts .nav-item a", "mydhcpname")
+        m1.execute("hostnamectl set-hostname 'localhost'")
 
         self.add_machine(b, "10.111.113.2")
         self.wait_host_addresses(b, ["localhost", "10.111.113.2"])


jelly:
Not 100% sure if the transient hostname is correctly obtained correctly in the VM I start with `./bots/vm-run -s cockpit fedora-34` it's `fedora`.

But the code in cockpits bots does something else it seems https://github.com/cockpit-project/bots/blob/9450dad0b0f588c705d576165c3f42c7072a347d/machine/machine_core/machine_virtual.py#L349

marusak:
If you need to repush, then please fix commit message, it is `Fixes: #9909` (missing `#`). If not, it can be fixed on "Squash & Merge"

martinpitt:
Yes, this will definitively need a repush. I'll give it a more detailed review yesterday.

test/verify/check-system-info
```diff
@@ -193,6 +193,12 @@ class TestSystemInfo(MachineCase):
         b.wait_in_text('#system_information_hostname_text', "Adventure Box (host1.cockpit.lan)")
         self.assertEqual(m.execute("hostname").strip(), "host1.cockpit.lan")
 
+        transient_hostname = m.image.split('-')[0]
```
martinpitt, line 196: Let's not make this assumption. As the tests show, at least on Debian and RHEL 8 the host name ends up as "localhost". I suggest to call `hostnamectl --transient` to detect the current transient host name. However, this would still assume that the transient host name is actually different from the static one, which may not be the case on all images. Hence, let's drop this altogether. See below.

test/verify/check-system-info
```diff
@@ -193,6 +193,12 @@ class TestSystemInfo(MachineCase):
...
+        m.execute("hostnamectl set-hostname ''")
```
martinpitt, line 197: This sets the static *and* transient name, which makes the test pointless. Instead I suggest to do `set-hostname --static ''` and `set-hostname --transient mydhcpname`, so that you pin down the names at that point. Then the actual check can wait for 'mydhcpname'.

test/verify/check-system-info
```diff
@@ -193,6 +193,12 @@ class TestSystemInfo(MachineCase):
...
+        b.wait_in_text('#system_information_hostname_text', transient_hostname)
+        # Allow error: Failed to create "/run/systemd/default-hostname": Permission denied
+        self.allow_journal_messages(".*scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=dir permissive=0")
```
martinpitt, line 200: This is an actual OS bug, it even got reported and fixed in RHEL 9 a month ago here: https://bugzilla.redhat.com/show_bug.cgi?id=1966492 . It's not *quite* the same, as that concerned file removal, but it looks close enough. Can you please clone this for Fedora 34 (the little blue icon with two overlapping squares on the top right -- not that easy to see!) with mentioning the current systemd/selinux-policy package names?

We handle such unexpected messages in a slightly more delicate manner: On our bots project we maintain a [set of OS bugs labelled with "knownissue"](https://github.com/cockpit-project/bots/issues?q=is%3Aissue+is%3Aopen+label%3Aknownissue), e.g. https://github.com/cockpit-project/bots/issues/2162 for a recent SELinux bug. Then in https://github.com/cockpit-project/bots/tree/main/naughty we have a set of "failure patterns", like https://github.com/cockpit-project/bots/blob/main/naughty/fedora-34/2162-selinux-nft-pipe for the previous example (the file name is "issue number - short description"). Our test machinery tracks when and where it has acutally seen these issues, turns the "FAIL"s into a "SKIP" (i.e. a success with a warning), and makes a note in the knownissue.

Once it sees that a pattern is not used any more, it auto-files a PR like https://github.com/cockpit-project/bots/pull/2239 that cleans up the pattern -- and we can then close the downstream bug, or do another appropriate action.

So please: duplicate the bugzilla for  F34, create a bots knownissue for it, send a bots PR to add the pattern, and drop this line.

martinpitt (changes requested):
Thanks! The fix looks good, the test needs work. I invited you to the contributors team, so that tests will start automatically for you.

test/verify/check-system-info
jelly: I've filed a bugzilla issue, https://github.com/cockpit-project/bots/issues/2291 and created a pattern, I am just not 100% sure if it's correct.

jelly:
> If you need to repush, then please fix commit message, it is `Fixes: #9909` (missing `#`). If not, it can be fixed on "Squash & Merge"

Thanks, fixed!

martinpitt (approved):
Thanks! A bit curious that this actually succeeded *without* https://github.com/cockpit-project/bots/pull/2292 -- the test does not actually seem to trigger this SELinux denial? The test succeeded three times everywhere.

martinpitt:
Do you mind if we retitle this "systemd: Show transient hostname"? That's what it effectively does.

test/verify/check-system-info
```diff
@@ -193,6 +193,10 @@ class TestSystemInfo(MachineCase):
         b.wait_in_text('#system_information_hostname_text', "Adventure Box (host1.cockpit.lan)")
         self.assertEqual(m.execute("hostname").strip(), "host1.cockpit.lan")
 
+        m.execute("hostnamectl set-hostname --static ''")
+        m.execute("hostnamectl set-hostname 'mydhcpname'")
```
martinpitt, line 197: Oh wait, that's cheating -- `set-hostname` also writes /etc/hostname. Please do `--transient`.

martinpitt (changes requested):
So when you force-push again, do you mind updating the commit message yourself? Thanks!

test/verify/check-system-info
jelly: Yup.. ha! Setting --transient after only static to empty does not work either, as the `Pretty Hostname` is still set. So I've first emptied everything and then set the transient hostname in the test.

jelly:
I've included the fix for the hosts selector, if this needs to be a separate PR I can make that happen.

martinpitt (dismissed):
Thanks! Would be nice for the shell fix to get an integration test as well, but it's at least the same approach as in systemd.

marusak:
> Thanks! Would be nice for the shell fix to get an integration test as well, but it's at least the same approach as in systemd.

@jelly Wanna add that, or can I merge this one?

jelly:
> > Thanks! Would be nice for the shell fix to get an integration test as well, but it's at least the same approach as in systemd.
> 
> @jelly Wanna add that, or can I merge this one?

Oh sure, I can add it, can also add it in a follow up PR as I'm not sure yet in which test file it should be added.

martinpitt (approved):
Thank you! Test looks nice, bots have the last word as always.

marusak:
@jelly [this](https://logs.cockpit-project.org/logs/pull-16213-20210811-102834-7b4744c3-fedora-34/log.html#274-2) failure seems familiar, doesn't it? :)

I'll reopen both the issue and PR.

martinpitt:
Sorry, I accidentally retried the failed test, after landing #16227. But this failure here was different.

marusak (approved):
Thanks!

marusak:
@jelly the pattern needs to be exptended also for [fedora-coreos](https://logs.cockpit-project.org/logs/pull-16213-20210811-153533-2fd39fb1-fedora-coreos/log.html#279) and [rhel-9](https://logs.cockpit-project.org/logs/pull-16213-20210811-153533-2fd39fb1-rhel-9-0/log.html#279-2)

Do you want me to do that?

jelly:
> @jelly the pattern needs to be exptended also for [fedora-coreos](https://logs.cockpit-project.org/logs/pull-16213-20210811-153533-2fd39fb1-fedora-coreos/log.html#279) and [rhel-9](https://logs.cockpit-project.org/logs/pull-16213-20210811-153533-2fd39fb1-rhel-9-0/log.html#279-2)
> 
> Do you want me to do that?

Sounds good to me!

marusak:
Sent https://github.com/cockpit-project/bots/pull/2302

PR #16215, opened 2021-08-08
lib: Some page.scss cleanups

### LINKED ISSUE #15425: Move PF3 overrides from page.scss to patternfly-overrides.scss
When working on the CSS variable reduction @ #15183, I noticed that there are apparently still quite a few PatternFly 3 overrides in page.scss.

Anything that can be moved to the override file for PF3 should be.

It doesn't make sense to pull in PF3 overrides for all pages — especially for PF4-only pages.

martinpitt:
`.blank-slate-pf` is a PF3 (only) class, and the p + p margin hack only
happens on pages with PF3+PF4. Move these two to
patternfly-overrides.scss, so that they get cleaned up together with PF3
removal.

Similarly, the form group and modal box padding are PF4 specific fixes,
move them to patternfly-4-overrides.scss.

`.modal-content` is a PF3-only class which is only being used in the
shell -- move it to shell.scss.

Now page.scss only has `.ct-*` classes and changes which are genuinely
Cockpit specific (such as heading size changes).

Fixes #15425

### DIFF
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -84,12 +84,6 @@ a.disabled:hover {
     max-height: 90vh;
 }
 
-/* Allow kebab menus to have a ^ with overflowing */
-/* Note: This means kebab menus cannot be _too_ long */
-.dropdown-kebab-pf > .dropdown-menu {
-    overflow: visible;
-}
-
 .highlight-ct {
     background-color: var(--ct-color-link-hover-bg);
 }
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -101,13 +101,6 @@ a.disabled:hover {
     position: fixed;
 }
 
-.panel .well {
-    margin-bottom: 0px;
-    border: none;
-    border-radius: 0px;
-    background-color: var(--pf-global--palette--black-200);
-}
-
 /* Dialog patterns */
 
 .dialog-wait-ct {
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -95,12 +95,6 @@ a.disabled:hover {
     overflow: visible;
 }
 
-/* Align these buttons more nicely */
-.btn.fa-minus,
-.btn.fa-plus {
-    padding-top: 4px;
-}
-
 .highlight-ct {
     background-color: var(--ct-color-link-hover-bg);
 }
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -29,12 +29,6 @@ a {
     cursor: pointer;
 }
 
-p + p {
-    // The mix of PF3 and PF4 removes margin from paragraphs.
-    // We want successive paragraphs to have spaces between each other.
-    margin-top: var(--pf-global--spacer--md);
-}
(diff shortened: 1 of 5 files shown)


martinpitt:
This still needs proper visual testing, and discussing some remaining bits in page.scss, in particular:
```css
/* HACK: https://github.com/patternfly/patternfly/issues/255 */
input[type=number]:not(.pf-c-form-control) {
  padding: 0 0 0 5px;
}
```

I also wonder about this one:
```css
.disabled {
    pointer-events: auto;
}
```
(there's a couple of related rules)

garrett:
I don't think the correct PF issue is linked. The issue for for backdrops and this clearly adjusts non-PF4 input numbers, adjusting the padding.

```
/* HACK: https://github.com/patternfly/patternfly/issues/255 */
input[type=number]:not(.pf-c-form-control) {
  padding: 0 0 0 5px;
}
```

However, if I add `-3` to the URL, the issue looks related:

https://github.com/patternfly/patternfly-3/issues/255

garrett:
> I also wonder about this one:
> 
> .disabled {
>     pointer-events: auto;
> }

We do use `disabled` as a class in places still, instead of using the `disabled` attribute (which is the correct way of doing things). :shrug: 

As for pointer events, that really shouldn't be needed unless one of the following is true:
- turn off mouse-related events for the element, so it passes through to any element behind (`pointer-events: none`)
- overwriting some other pointer-events for whatever reason (if PF or Bootstrap or some other CSS is turning it off for whatever reason)

This particular snippet is saying "don't let someone click (or select) a disabled item, like what looks like a disabled button", which is faking proper UX (can't click disabled buttons), but not the correct way to make something disabled.

I guess it's important for something in Cockpit (since I see we're still using it in JS, including `cockpit.js`), but could be implemented more correctly.  Since it's in `cockpit.js`, we probably need to keep it in `page.scss` for now.

KKoukiou:
@garrett did you find any occurance where we still use disabled class? I grepped and found nothing.

garrett:
> @garrett did you find any occurance where we still use disabled class? I grepped and found nothing.

pkg/base1/cockpit.js:

```js
    document.addEventListener("click", function(ev) {
        if (ev.target.classList && in_array(ev.target.classList, 'disabled'))
          ev.stopPropagation();
    }, true);
```
(This is checking for its existence however.)

pkg/shell/indexes.js:

```js
        $("#navbar-dropdown").addClass("disabled");
```

This is actually adding the class, probably for styling.

These are the last remnants that I see across our codebase, at least for Cockpit itself. I haven't checked the sub-projects yet, like Machines, Podman, OStree, etc. (Removing the CSS here would get synced to those projects eventually and they could break if they do indeed use a class for a disabled state.)

KKoukiou:
@garrett added back then the disabled class stuff and we can deal with that in a new PR.

garrett:
> added back then the disabled class stuff and we can deal with that in a new PR.

Yeah, agreed 💯: best not to deal with that here (and delay the rest further).

KKoukiou (approved):

PR #16216, opened 2021-08-08
test: Make unexpected PCP error messages more specific

### LINKED ISSUE #15625: Integration tests produce PCP warnings that should be explained
We see these warnings from `cockpit-pcp` in our integration tests. We only discovered them when we cast a wider net for journal messages (in #15599), so it is unknown when they have started.  It would be excellent if we could figure out why they happen.

        pcp-archive: no such metric: blah

This might be harmless and expected since an archive might be from a time when pmlogger didn't log that particular metric that we try to find in it.  Cockpit has its own pmlogconf snippet in /var/lib/pcp/config/pmlogconf/tools/cockpit (which we probably have neglected), so the pmlogger configuration might change when cockpit is installed.

        pcp-archive: instance name lookup failed: blah

This doesn't look so harmless.  This message appears when cockpit-pcp has read a sample from an archive, and that sample contains a list of instance indices, but then pcp can't convert the index into a name. This looks like a corrupted archive.

martinpitt:
Only two of these messagess still actually appear, drop the other two.

The two "no such metric: kernel.all.cpu.nice: Unknown metric name"
instances in check-metrics happen because these two tests start with
an empty /var/log/pcp and pmlogger.service stopped, so that indeed the
metrics channel will not be able to find any metric (and
`kernel.all.cpu.nice` is just the first one that it tries). So these are
legitimate, ignore these in the specific tests with a comment.

Move the networking/storage ones into the particular test classes and
make them more specific, to avoid introducing more of them.

See #15625

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1185,12 +1185,6 @@ class MachineCase(unittest.TestCase):
         r"#1\) Respect the privacy of others.",
         r"#2\) Think before you type.",
         r"#3\) With great power comes great responsibility.",
-
-        # Something unknown sometimes goes wrong with PCP, see #15625
-        "pcp-archive: instance name lookup failed: .*",
-        "direct: instance name lookup failed: .*",
-        "pcp-archive: couldn't create pcp archive context for.*",
-        "pcp-archive: no such metric: .*"
     ]
 
     default_allowed_messages += os.environ.get("TEST_ALLOW_JOURNAL_MESSAGES", "").split(",")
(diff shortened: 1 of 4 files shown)


martinpitt:
This round went fine, aside from a pycodestyle issue and an unrelated flake. Trying another round to be sure.

martinpitt:
@mvollmer This doesn't solve the issue completely, but it's a worthwhile step IMHO.

mvollmer (approved):

Issue #7476, opened 2017-08-09
DMI Info Fallback


dcrdev:
In instances where the mainboard manufacturer has not populated the DMI tables within the firmware, can a fallback be introduced for the "Hardware" section on the "System" tab?

As a fallback could this information be sourced from /proc/cpuinfo ?

Just built a new server and the firmware has empty DMI tables, resulting in Cockpit just displaying "To Be Filled By O.E.M." several times over.

dcrdev:
Ok I didn't fully understand what was happening here, but now I can shed some light on the problem as I was going to submit a pr but my workaround now seems like the wrong solution.

The problem is that my board has an ipmi chip and that is being listed by dmidecode before the main board. So the second entry accurately shows information about my board, the first entry has no dmi information written.

monreal:
Similar or same in my case. Please get the information from "Base Board Information".

> Handle 0x0001, DMI type 1, 27 bytes
> System Information
> 	Manufacturer: To Be Filled By O.E.M.
> 	Product Name: To Be Filled By O.E.M.
> 	Version: To Be Filled By O.E.M.
> 	Serial Number: To Be Filled By O.E.M.
> 	UUID: 71c28570-1177-0000-0000-000000000000
> 	Wake-up Type: Power Switch
> 	SKU Number: To Be Filled By O.E.M.
> 	Family: To Be Filled By O.E.M.
> 
> Handle 0x0002, DMI type 2, 15 bytes
> Base Board Information
> 	Manufacturer: ASRock
> 	Product Name: J4105-ITX
> 	Version:                       
> 	Serial Number: M80-B1004200114
> 	Asset Tag:                       
> 	Features:
> 		Board is a hosting board
> 		Board is replaceable
> 	Location In Chassis:                       
> 	Chassis Handle: 0x0003
> 	Type: Motherboard
> 	Contained Object Handles: 0
> 
> Handle 0x0003, DMI type 3, 22 bytes
> Chassis Information
> 	Manufacturer: To Be Filled By O.E.M.
> 	Type: Desktop
> 	Lock: Not Present
> 	Version: To Be Filled By O.E.M.
> 	Serial Number: To Be Filled By O.E.M.
> 	Asset Tag: To Be Filled By O.E.M.
> 	Boot-up State: Safe
> 	Power Supply State: Safe
> 	Thermal State: Safe
> 	Security Status: None
> 	OEM Information: 0x00000000
> 	Height: Unspecified
[...]

PR #16217, opened 2021-08-09
systemd: Read hardware info from DeviceTree

martinpitt:
If DMI is not available, fall back to reading the model and serial
number from /proc/device-tree/

----

On my RasPi 3  on main, Overview and hwinfo:

raspi-overview-main

raspi-hwinfo-main

On this PR:

raspi-overview-pr

raspi-hwinfo-pr

As I [mentioned here](https://github.com/cockpit-project/cockpit/issues/7476#issuecomment-894980354), I tried hard to fake a /proc/device-tree in our tests, but did not find a way to do this. So for now I tested this on my actual RasPi.

### DIFF
--- a/pkg/lib/machine-info.js
+++ b/pkg/lib/machine-info.js
@@ -131,6 +131,24 @@ export function dmi_info(address) {
     return pr;
 }
 
+// decode a binary Uint8Array with a trailing null byte
+function decode_proc_str(s) {
+    return cockpit.utf8_decoder().decode(s.slice(0, -1));
+}
+
+export function devicetree_info() {
+    let model, serial;
+
+    return Promise.all([
+        // these succeed with content === null if files are absent
+        cockpit.file("/proc/device-tree/model", { binary: true }).read()
+                .then(content => { model = content ? decode_proc_str(content) : null }),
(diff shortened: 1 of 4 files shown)


martinpitt:
[One failure](https://logs.cockpit-project.org/logs/pull-16217-20210809-081100-faea4110-rhel-9-0/log.html#218-2). Not sure what's wrong, the HTML dump seems okay. Trying to reproduce locally

marusak:
> Not sure what's wrong, the HTML dump seems okay.

It renders `DIMM0` not `DIMM 0`.

martinpitt:
Ah, thanks for spotting. I ran the test several times locally, and can't reproduce this. However, this is the real system DIMM information, not the fake one. So apparently in some cases, the VM's emulated DMI info says "DIMM0" instead of "DIMM 0"? I guess I'll send a PR to just relax that string, and retry this test here.

martinpitt:
Sent PR #16218 for the "DIMM0" vs "DIMM 0" issue.

martinpitt:
Green now.. :man_shrugging:

marusak (approved):
Looks good, thanks!

PR #16220, opened 2021-08-09
bridge: Fix PCP archives on 32 bit architectures

### LINKED ISSUE #16083: pcp integration does not work on armhf based platform
Cockpit version: 248
OS: ubuntu impish (developement branch) armhf

this does not happen on x64

Steps to reproduce

1. install and enable pcp metrics
2. wait some time until data should be recorded

the result is that it keeps saying "No data available" even tough pmval -f 3 disk.partitions.write -a /var/log/pcp/pmlogger/... clearly shows that there is data.
see bug 12747

martinpitt:
`label.ll_start.tv_sec` is a 32 bit type on armhf and other platforms.
Multiplying with 1000 causes an integer overflow, which completely
messes up the time stamp. Fix this by casting to a 64 bit type first.

Fixes #16083

------

I validated the bug and fix on my RasPi 3. The fix is good, but I'd like to create a unit test for this.

 - [x] Add unit test

### DIFF
--- a/src/bridge/cockpitpcpmetrics.c
+++ b/src/bridge/cockpitpcpmetrics.c
@@ -121,7 +121,7 @@ result_meta_equal (CockpitPcpMetrics *self,
 static gint64
 timestamp_from_timeval (struct timeval *tv)
 {
-  return tv->tv_sec * 1000 + tv->tv_usec / 1000;
+  return (gint64) tv->tv_sec * 1000 + tv->tv_usec / 1000;
 }
 
 static JsonObject *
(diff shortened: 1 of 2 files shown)


src/bridge/cockpitpcpmetrics.c
```diff
@@ -695,7 +695,7 @@ add_archive (CockpitPcpMetrics *self,
       return;
     }
 
-  info->start = label.ll_start.tv_sec * 1000 + label.ll_start.tv_usec / 1000;
+  info->start = (guint64) label.ll_start.tv_sec * 1000 + label.ll_start.tv_usec / 1000;
```
martinpitt, line 698: `info->start` is a `gint64`, so let's move to that right away.

allisonkarlitskaya (approved):
Some extra brackets would be nice to avoid having to lookup the order of operations, but this looks good, thanks.

PR #16223, opened 2021-08-10
selinux: allow login to read motd file

lnussel:
If pam_motd is enabled, login needs to be able to read cockpit's file

### DIFF
--- a/selinux/cockpit.te
+++ b/selinux/cockpit.te
@@ -180,3 +180,11 @@ optional_policy(`
 optional_policy(`
 	unconfined_domtrans(cockpit_session_t)
 ')
+
+# login may read motd file through pam
+optional_policy(`
+    gen_require(`
+        type local_login_t;
+    ')
+    cockpit_read_pid_files(local_login_t)
+')


lnussel:
A similar rule is in ssh upstream: https://github.com/fedora-selinux/selinux-policy/blob/c0b4e9408b41cb33de0424ce9ba7aead1a991490/policy/modules/services/ssh.te#L356

I suppose if the cockpit module is meant to be moved here instead all of those bits have to be moved as well!?

marusak:
> I suppose if the cockpit module is meant to be moved here instead all of those bits have to be moved as well!?

I am still not sure what is the right place. If it should be shipped by Cockpit or by related packages. For example, shouldn't this PR be part of https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/system/locallogin.te with something like:
```
optional_policy(`
    gen_require(`                                                               
        type cockpit_var_run_t;                                                 
    ')   
    
    read_files_pattern(local_login_t, cockpit_var_run_t, cockpit_var_run_t)
    read_lnk_files_pattern(local_login_t, cockpit_var_run_t, cockpit_var_run_t)
')
```

lnussel:
I don't know, you tell me :-) Wouldn't that still be some kind of a layering violation resp circular dependency if the cockpit module itself gets outsourced?

The situation is a bit confusing to me. So far cockpit_read_pid_files() is used in the other files that refer to cockpit.
Then there is also https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/files.fc#L312

        /run/cockpit/motd	--	gen_context(system_u:object_r:etc_t,s0)

Which seems to suggest the idea was to make cockpit's motd have the same context as the ones in /etc. Looks like that doesn't work?

martinpitt (approved):
I'm not 100% sure either, I think this is just one of these "in between" things which don't have a clear solution. But I lean towards the producer declaring which consumers are allowed, and as cockpit is optional, but login_t is always there, this PR feels more robust than adding it to selinux-policy.

Danke!

Requeued two stuck tests, @marusak can have the last word and land or move to selinux-policy.

marusak:
:man_shrugging: 
Maybe lets wait until someone from selinux team will respond in https://github.com/fedora-selinux/selinux-policy/pull/757

Issue #16219, opened 2021-08-09
network settings plus button misaligned


jelly:
Cockpit version: 250
OS: Fedora 34
Page: Network

As can be seen in the screenshot the "plus button" should be aligned right next to the "Addresses" selection menu.

![image](https://user-images.githubusercontent.com/67428/128720557-4ff19771-8cc7-4a25-bb0c-477ae46eaa1e.png)

PR #16224, opened 2021-08-10
networkmanager: Align add button next to adresses selection

jelly:
Use display flex to allow space for the "add button", so it appears on
the same line as the select element. To keep the minus buttons button
aligned with the other buttons the margin-right has to be explictly
removed.

Fixes: #16219

### DIFF
--- a/pkg/networkmanager/networking.scss
+++ b/pkg/networkmanager/networking.scss
@@ -93,7 +93,7 @@
     }
 
     tr td .pf-m-secondary {
-        margin-right: 4px;
+        margin-right: 0px;
     }
 }
 
@@ -453,6 +453,10 @@ th {
 #add-zone-services-readonly legend {
     padding: 0;
     line-height: 1;
+}
+
+.network-ip-settings-row .pull-right {
+    display: flex;
 }
 
 // Animation for Firewall's add service dialog,


pkg/networkmanager/networking.scss
```diff
@@ -455,6 +455,10 @@ th {
     line-height: 1;
 }
 
+.network-ip-settings-row .pull-right {
```
KKoukiou, line 458: As we talked over on IRC this could better be done like:
``` diff
diff --git pkg/networkmanager/interfaces.js pkg/networkmanager/interfaces.js
index 11676a0c9..33cbc5678 100644
--- pkg/networkmanager/interfaces.js
+++ pkg/networkmanager/interfaces.js
@@ -1843,7 +1843,7 @@ PageNetworkIpSettings.prototype = {
                 $('<div class="network-ip-settings-row">').append(
                     $('<div>').append(
                         $('<strong>').text(title),
-                        $('<div class="pull-right">').append(
+                        $('<div class="pull-right pf-l-flex pf-m-nowrap">').append(
                             header_buttons,
                             add_btn = $('<button class="pf-c-button pf-m-secondary pf-m-small">')
                                     .append('<span class="fa fa-plus">')
```

So that we use the PF4 classes, and in the future this can be ported over to react PF4 component.

KKoukiou (approved):
As this dialog is not using PF4 anyway, I am fine with this fix as is :) Thanks @jelly !

martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-16224-20210810-144638-675c82b8-fedora-33/log.html#65-2) is a recently introduced flake, sorry about that. I filed PR #16227 to fix that. Retrying.

pkg/networkmanager/networking.scss
garrett: @jelly: FWIW, just to let you know:

We shouldn't use `pull-right` in Cockpit. The class uses `float: right`, which is almost always wrong for many reasons:

- floats are wacky and not always predictable
- floats need one or another type of containing hack (adjusting overflow, using clear, etc.), which can have other side effects
- floats cause the content to be out of order (which negatively impacts all accessibility... not just screen readers, but also keyboard navigation and sometimes even mouse interaction)

Floats should only be used when something is a pulled element (such as an image) and content should wrap around it (which is more for an article and does not happen in UI like in Cockpit). Floats should never be used for general layout.

Instead of floats, flex or grid should be used (depending on what would work best in the context).

But, I agree with @KKoukiou that as this is legacy code that isn't using PF4 anyway, _this is fine for now_. 

When it's re-written to _not_ use jQuery and to use PF4, it should definitely not use `pull-right`. There are other issues with the code in this block, like a classless `<div>` used as a wrapper, but this, like the pull-right, is due to a lot of legacy. 

**To be clear**: I'm not suggesting any more changes here at the moment (aside from it being rewritten in the future). :grin:

PR #16225, opened 2021-08-10
tools: Update user-agent in urls-check script

### LINKED ISSUE #16221: Validate all URLs
Checked 139 URLs out of which 1 is/are invalid:
https://bugzilla.mozilla.org/show_bug.cgi?id=135309

marusak:
It seems that just `Mozilla/5.0` was not enough for https://bugzilla.mozilla.org
as it was returning `400 - Bad request`. Putting there something more
real makes it work again.

Fixes #16221

### DIFF
--- a/tools/urls-check
+++ b/tools/urls-check
@@ -99,7 +99,7 @@ def check_urls(verbose):
 
         try:
             # Specify agent as some websites otherwise block requests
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! Wow, didn't know that bz was that picky about the agent string..

PR #16226, opened 2021-08-11
selinux:  Allow cockpit_ws_t get attributes of fs_t filesystems

marusak:
This permission is required by a cockpit.socket ExecStartPost script
which executes
"systemctl show --property Listen cockpit.socket"
in the cockpit-ws caller domain.

Resolves: rhbz#197918
Picked from https://github.com/fedora-selinux/selinux-policy/commit/95b6d266399


martinpitt (approved):
Dakujem!

PR #16227, opened 2021-08-11
test: Re-ignore variant of PCP error message

martinpitt:
Commit 8bd081d8c1712c made the error message more specific, but it turns
out that a variant sometimes happens as well:

    pcp-archive: instance name lookup failed: network.interface.in.bytes.5: Instance identifier not defined in the PCP archive log

So change the pattern back to what we had before that commit.

----

[example](https://logs.cockpit-project.org/logs/pull-16224-20210810-144638-675c82b8-fedora-33/log.html#65-2)


marusak (approved):

martinpitt:
Meh, [one more](https://logs.cockpit-project.org/logs/pull-16227-20210811-100839-2af7dc77-debian-stable/log.html#181)

marusak (approved):
Seems that the pattern was almost correct back then :)

PR #16229, opened 2021-08-12
tests: Fix pyflakes warnings

marusak:
test/verify/check-system-realms:711:18 f-string is missing placeholders
test/verify/check-system-realms:675:32: W605 invalid escape sequence '\['

### DIFF
--- a/test/verify/check-system-realms
+++ b/test/verify/check-system-realms
@@ -672,7 +672,7 @@ ipa-advise enable-admins-sudo | sh -ex
         m.execute(f'''! selinuxenabled || semanage port -m -t websm_port_t -p tcp 443
                       mkdir -p /run/systemd/system/cockpit.socket.d/
                       printf "[Socket]\nListenStream=\nListenStream=443" > /run/systemd/system/cockpit.socket.d/listen.conf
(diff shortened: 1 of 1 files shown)


test/verify/check-system-realms
```diff
@@ -672,7 +672,7 @@ ipa-advise enable-admins-sudo | sh -ex
         m.execute(f'''! selinuxenabled || semanage port -m -t websm_port_t -p tcp 443
                       mkdir -p /run/systemd/system/cockpit.socket.d/
                       printf "[Socket]\nListenStream=\nListenStream=443" > /run/systemd/system/cockpit.socket.d/listen.conf
-                      sed -i '/\[WebService/ aOrigins = http://{m.web_address}:{m.web_port}' /etc/cockpit/cockpit.conf''')
+                      sed -i '/[WebService/ aOrigins = http://{m.web_address}:{m.web_port}' /etc/cockpit/cockpit.conf''')
```
mvollmer, line 675: Hmm, isn't the intention here to pass `\[` to sed? (so that it matches a literal `[` instead of starting a character class.)

mvollmer: Yes, `f'''\['''` produces `\[`, and to silence the warning without changing the result, this needs to be `f'''\\['''`

mvollmer (approved):

PR #16235, opened 2021-08-16
test: "pcp: Unknown metric" also happens in TestCurrentMetrics

mvollmer:
For example here: https://logs.cockpit-project.org/logs/pull-16234-20210816-101305-cf949df2-fedora-34-devel/log.html#47


marusak (approved):
Thanks!

PR #16236, opened 2021-08-17
test: change match string in selinux modifications table

allisonkarlitskaya:
The message about "Allow zebra..." is changing in some distros and seems
likely to change again.  Instead of trying to build a regexp around
that, let's just match on a different item from the table.


marusak (approved):
Thanks, this one seems like a good compromise, of course tests have the last word.

PR #16237, opened 2021-08-17
more selinux string fixes

allisonkarlitskaya:
...and a nice regexp helper.

### DIFF
--- a/test/verify/check-selinux
+++ b/test/verify/check-selinux
@@ -95,7 +95,7 @@ class TestSelinux(MachineCase):
         self.machine.execute(script=FIX_AUDITD)
 
         # wait for Modifications table to initialize
-        b.wait_in_text(".modifications-table", "Allow zebra")
+        b.wait_text_matches(".modifications-table", "Allow zebra.*write.*config")
 
         # httpd_read_user_content should be off by default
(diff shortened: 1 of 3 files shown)


test/verify/check-selinux
```diff
@@ -192,7 +193,7 @@ class TestSelinux(MachineCase):
             # fallback for older Chrome releases
             b.grant_permissions("clipboardRead", "clipboardWrite")
 
-        b.wait_visible(".pf-c-data-list__cell:contains(Allow zebra to write config)")
+        b.wait_visible(".pf-c-data-list__cell:contains(Allow zebra)")
```
mvollmer, line 196: Should this also be "Allow zebra.*write.*config" like above for .modifications-table?

allisonkarlitskaya: Because we use the `:contains()` selector, we can't put a regexp here.

mvollmer (approved):

test/verify/check-selinux
mvollmer: Ouch, obviously.

allisonkarlitskaya:
rhel-9-0 is clean against cockpit-project/bots#2306 so I'll land that one straight away.  fedora-33 on cockpit-project/bots#2313 has a (probable) flake so I'll rerun that batch, though.

PR #16246, opened 2021-08-19
Fixes for new Fedora 35 image

mvollmer:


mvollmer:
Still a lot of failures with the fedora-25 image, but this PR fixes the tests that it is supposed to fix: TestPages.testBasic, TestTerminal.testBasic, TestUpdates.testKpatch.

mvollmer:
Let's try to get them all fixed here -> draft.

mvollmer:
Green!

mvollmer:
> Green!

Hmm, not anymore.  TestSystemInfo.testOverview is flaky now, I'll check.

test/verify/check-system-terminal
```diff
@@ -107,6 +107,8 @@ PROMPT_COMMAND='printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/
         # Check that when we `exit` we can still reconnect with the 'Reset' button
         b.key_press("exit\r")
         b.wait_in_text(".terminal .xterm-accessibility-tree", "disconnected")
+        # Remove failed units which will show up in the first terminal line
```
marusak, line 110: Why would this be needed? We call `systemctl reset-failed` in `setUp`. Is this race, that this service fails between setUp and this call?

mvollmer: Yes, apparently.  rhsm.service is started on-demand, probably by our own "Software Updates" page.

But:
- why is rhsm.service in the image at all?
- the crash disrupts other things and needs to be fixed obviously. So, naughty? wdyt?

marusak: Do we even need `rhsm.service` for any Fedora-35 test? How about masking it in `.prepare` script?

mvollmer: >     * why is rhsm.service in the image at all?

Because we explicitly install it in `fedora.setup`, so that is not a mistake.

mvollmer: I think we need it for the tests. People can install it from the repositories, and then our Software page interacts with it, so we should test that.

I'll make a naughty pattern for this, we already have two for the tests that interact with rhsm.service.

mvollmer: Done.

mvollmer:
I can't reproduce the TestNetworkingMTU.testMtu failure on debian-testing. I have never seen the journal message before:

```
pcp-archive: couldn't create pcp archive context for /var/log/pcp/pmlogger/debian/20210824.13.21: Permission denied (-13)
```

marusak (approved):
Thanks!

PR #16252, opened 2021-08-24
fmf: Add hack for stuck chpasswd

marusak:
Also see https://github.com/cockpit-project/bots/commit/c7ea7746b3

This is needed for downstream gating

### DIFF
--- a/test/run-verify-host.sh
+++ b/test/run-verify-host.sh
@@ -43,6 +43,9 @@ if grep -q 'ID=.*fedora' /etc/os-release; then
     dnf install -y tcsh
 fi
 
+#HACK: unbreak rhel-9-0's default choice of 999999999 rounds, see https://bugzilla.redhat.com/show_bug.cgi?id=1993919
+sed -ie 's/#SHA_CRYPT_MAX_ROUNDS 5000/SHA_CRYPT_MAX_ROUNDS 5000/' /etc/login.defs
+
 # make libpwquality less aggressive, so that our "foobar" password works
 printf 'dictcheck = 0\nminlen = 6\n' >> /etc/security/pwquality.conf
 
--- a/test/run-verify-host.sh
+++ b/test/run-verify-host.sh
@@ -18,19 +18,11 @@ nproc
 free -h
 rpm -q cockpit-system
 
-# install browser; on RHEL, use chromium from epel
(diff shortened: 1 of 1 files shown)


marusak:
More like `fml:...` - testing this in downstream now.

marusak:
packit is failing because `semanage: command not found`

marusak:
Firefox and chpasswd hacks worked in downstream :tada: 
Let also make ti green in upstream.

marusak:
oh, komm schon:
```
+ dnf remove -y setroubleshoot-server
Dependencies resolved.
================================================================================
 Package                       Arch    Version          Repository         Size
================================================================================
Removing:
 setroubleshoot-server         x86_64  3.3.24-1.el8     @koji-override-1  1.3 M
Removing unused dependencies:
 policycoreutils-python-utils  noarch  2.9-15.el8       @baseos           138 k
```

test/run-verify-host.sh
```diff
@@ -18,19 +18,11 @@ nproc
...
-        rm -f /etc/yum.repos.d/epel.repo
-        dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
-        sed -i 's/$releasever/8/' /etc/yum.repos.d/epel*.repo
-        dnf config-manager --enable epel
-    fi
-    dnf install -y chromium
-fi
+# install browser; on RHEL, use firefox-nightly as chromium broke CDP on rhel-9
+# Install firefox to pull in all the deps
+dnf -y install firefox
```
mvollmer, line 23: Why do we install firefox from the repos and then install it again from mozilla.org?

marusak: "Install firefox to pull in all the deps"

We need some libraries to be present on the system - therefore I install firefox from repos so it pulls it in. But we need nightly as that only has CDP.

test/run-verify-host.sh
```diff
@@ -18,31 +18,26 @@ nproc
...
```
allisonkarlitskaya, line 23: That goes better in the `require:` list in `test/verify.fmf`, no?

marusak: Not really, it is part of a hack. I hope that one day we can do this in some better way, so keeping it at one place feel cleaner to me.

test/run-verify-host.sh
mvollmer: Ah, makes sense!

mvollmer (approved):

PR #16255, opened 2021-08-24
systemd: Don't show the current year for logins

jelly:
As part of #13882 improvements, when the succesful/failed login date is
in the same year as the current year drop the year in the date time
output. Intl.DateTimeFormat has no standard preset without a year in it.


marusak (approved):
Nice, thanks!

PR #16258, opened 2021-08-25
Po plugin dependencies

allisonkarlitskaya:


martinpitt:
unit tests fail:
```
/tmp/source/pkg/lib/cockpit-po-plugin.js
  2:7  error  'glob' is assigned a value but never used  no-unused-vars
```

pkg/lib/cockpit-po-plugin.js
```diff
@@ -23,7 +23,12 @@ module.exports = class {
                     name: 'CockpitPoPlugin',
                     stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL,
                 },
-                () => Promise.all(glob.sync(path.resolve(srcdir, 'po/*.po')).map(f => this.buildFile(f, compilation)))
+                () => {
+                    const linguas_file = path.resolve(srcdir, "po/LINGUAS");
```
martinpitt, line 27: starter-kit and friends don't have/use LINGUAS, so the glob needs to stay at least as a fallback, or the other projects need to be converted first. Right now the following readFileSync() would crash with these.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -25,8 +25,10 @@ module.exports = class {
                 },
                 () => {
                     const linguas_file = path.resolve(srcdir, "po/LINGUAS");
+                    compilation.fileDependencies.add(linguas_file);
                     const linguas = fs.readFileSync(linguas_file, 'utf8').match(/\S+/g);
                     const po_files = linguas.map(lang => path.resolve(srcdir, 'po', lang + '.po'));
+                    po_files.forEach(po_file => compilation.fileDependencies.add(po_file));
```
martinpitt, line 31: It seems equally simple and effective to just add all *.po files as dependencies.

martinpitt (changes requested):
Honestly I'd rather just get rid of the LINGUAS file in cockpit.git, and make it work like starter-kit and friends. It has always been this error-prone redundant piece of information..

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: The reason for this is that a glob result doesn't have a timestamp: we can't teach make to check if the glob result changed.  `LINGUAS` has exactly this purpose, and we should use it.

We could potentially add `LINGUAS` to the `Makefile` dependencies and still use the glob, but it wouldn't bring as much: it would just move the crash from `webpack-make` to the `Makefile` telling us that it doesn't know how to get the `LINGUAS` file.

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: We're quite explicit about the fact that we don't promise any API here.  We'd clearly need to add LINGUAS into the other projects before we bump the version of `pkg/lib` that they pull in.

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: It's worth mentioning: I came across this as an actual bug: I pulled the cockpit repo after a .po refresh and was surprised to find that nothing got rebuilt.

martinpitt: But LINGUAS only gets updated rarely, when a new language gets added. It does not change on po file updates. So how does its timestamp help us in any way?

martinpitt (changes requested):
At the very least the static code tests need fixing, so that unit tests pass.

pkg/lib/cockpit-po-plugin.js
martinpitt: My point is rather that it is not really useful for starter-kit and friends: They use webpack, not make, for building all the po stuff, and webpack-watch isn't powerful enough for rebuilding on newly added files anyway. It just simply doesn't have that kind of "add dependency list" mechanism.

As discussed on IRC: I think it'd be fine to add LINGUAS to dependencies *if* it exists (i.e. for cockpit.git), and otherwise keep the globbing as-is.

martinpitt (approved):
Cheers, this looks nice now! Bots have the last word as usual, but the approach is good now for both cockpit and starter-kit.

PR #16259, opened 2021-08-25
test: Make TestUpdates.testBasic more robust

skobyda:
In downstream gating, a package can get installed to a test VM, which
could affect the status on updates page. Use tracer to make
the test robust and find out the correct status to be expected.

### DIFF
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -271,6 +271,22 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
         b = self.browser
         m = self.machine
 
+        def check_status():
+            if m.image in OSesWithoutTracer:
+                b.wait_in_text("#status p", "System is up to date")
+                # PK starts from a blank state, thus should force refresh and set the "time since" to 0
+                b.wait_in_text("#last-checked", "Last checked: less than a minute ago")
+            else:
+                code = m.execute("tracer > /dev/null 2>&1; echo $?").rstrip()
+                if code == "104":
+                    b.wait_in_text("#status p", "a system reboot")
+                elif code == "102" or code == "101":
+                    b.wait_in_text("#status p", "to be restarted")
+                else:
+                    b.wait_in_text("#status p", "System is up to date")
+                    # PK starts from a blank state, thus should force refresh and set the "time since" to 0
+                    b.wait_in_text("#last-checked", "Last checked: less than a minute ago")
+
         self.enable_preload("packagekit", "index")
 
         # Refresh cache so cockpit does not try to reload page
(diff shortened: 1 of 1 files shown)


test/verify/check-packagekit
```diff
@@ -271,6 +271,22 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
...
+            else:
+                code = m.execute("tracer > /dev/null 2>&1; echo $?").rstrip()
+                if code == "104":
+                    b.wait_in_text("#status p", "a system reboot")
+                elif code == "102" or code == "101":
+                    b.wait_in_text("#status p", "to be restarted")
+                elif code == "0":
+                    b.wait_in_text("#status p", "System is up to date")
+                    # PK starts from a blank state, thus should force refresh and set the "time since" to 0
+                    b.wait_in_text("#last-checked", "Last checked: less than a minute ago")
```
marusak, line 288: Should this be checked only when tracer returns 0? I guess we want to check it every time, so it can go outside of all the conditions.

test/verify/check-packagekit
```diff
@@ -271,6 +271,22 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
         b = self.browser
         m = self.machine
 
+        def check_status(b, m):
```
martinpitt, line 274: b and m are already defined in this scope, so no need to pass them as arguments here.

martinpitt (changes requested):
Thanks! Just a small thing to clean up when you factor out the "last checked" bit.

test/verify/check-packagekit
skobyda: Outside of all conditions? That would cause that in some cases, like when code 104 is returned, the test would check for `wait_in_text("#status p", "a system reboot")` and `wait_in_text("#status p", "System is up to date")`
So I think it should be in `else` condition

martinpitt: I think @marusak just refered to teh last line the "last checked". That should be common for all scenarios.

marusak: Indeed, I just meant the `Last checked: less than a minute ago` one.

skobyda: But the "last checked" line is not present for other scenarios (e.g. system needs reboot, some services need restart). It's only present for "System is up to date" and "# updates available" states.
And that's correct. Because if you have the latter two states, you may want to rerun RefreshCache to check if there are any new updates to the system.
But if the state is tracer telling you that your system needs reboot, then rerunning RefreshCache won't change a thing, and "last checked" is not really needed to be shown.

martinpitt: I see, thanks for the explanation!

martinpitt (approved):

martinpitt:
packit test failures are unrelated, so landing.

Issue #15786, opened 2021-05-04
When there is motd the login message takes just half screen size on big screens


KKoukiou:
Cockpit version: 243
Page: Overview

![Screen Shot 2021-05-05 at 13 03 00](https://user-images.githubusercontent.com/14921356/116994834-8b4ea580-acd9-11eb-9437-27e657bef7bc.png)

It would make sense that modt and login messages would just be splitting one line for small screens.

KKoukiou:
If it's not a bug just close, I was not sure.

garrett:
This is definitely a bug. This shouldn't happen.

marusak:
So there are 4 possible situations:
- Neither motd nor the last login are present (right now last login is always present, but one can close it or with https://github.com/cockpit-project/cockpit/issues/13882 it could happen)
- Only motd is present
- Only last-login is present
- Neither is present

When neither or both are present, then I think we want to span it across the whole screen.
If both are present then we want to split the screen in half? Or just have two 1/3 cards to match cards below?

@garrett WDYT?

KKoukiou:
So currently motd and login messages are both alerts; so according to @garrett these should span fully grid-column: 1/-1

However I clearly remember many discussions that login messages should not be an alert, but part of the system card.


Here is how it looks fully spanned. I feel it's too much wasted space.

![Screen Shot 2021-06-11 at 09 02 44](https://user-images.githubusercontent.com/14921356/121479982-dfc8fd00-c9ca-11eb-9960-f64853e9da2d.png)

@garrett do you think this worths a bit of a design guidance? In that case let's block it on some mockup.

marusak:
> However I clearly remember many discussions that login messages should not be an alert, but part of the system card.

It should be part of the card, when it is just info (just last login, no failed logins). When there are failed logins, we want to highlight it, thus it is fine to take a lot of space.

https://github.com/cockpit-project/cockpit/issues/13882

garrett:
> However I clearly remember many discussions that login messages should not be an alert, but part of the system card.
> Here is how it looks fully spanned. I feel it's too much wasted space.

Yep, agreed on all parts.

It should not be an alert. However, alerts should still span across. So until it's fixed to not be an alert (unless there's an actual problem), it should span across. But it should still be properly fixed to not alert for something that's perfectly normal.

> When there are failed logins, we want to highlight it

Correct: Failed logins are alerts. Successful logins are not alerts.

Issue #28, opened 2013-11-04
Show all user sessions and history


stefwalter:
Not just whether someone is currently logged or when the last time was.

martinpitt:
This would fit well on the Accounts page. A single account already has a "Terminate Session" button, so the overview could at least show all running sessions, or perhaps better, change the card appearance to indicate if that user is logged in or not.

marusak:
Related [design](https://raw.githubusercontent.com/cockpit-project/cockpit-design/9a65b44ffa9cf60a6eaf02044e797d0046b77d6b/users/login-warning.png). (Last two screen).

marusak:
`last --time-format iso mmarusak` can provide this list.

PR #16260, opened 2021-08-25
Move last login to system `Health` card

jelly:
Always show the last login information under system information and only
show failed logins as a notification.

Closes: #13882

---

## Overview: move last login to Health Card

Previously, last successful or failed logins were shown in a dismissable alert, which took a lot of space. That login information was moved to the Health card.

image

A failed login:

image

### DIFF
--- a/pkg/systemd/overview-cards/lastLogin.jsx
+++ b/pkg/systemd/overview-cards/lastLogin.jsx
@@ -1,0 +1,134 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React, { useState, useEffect } from 'react';
+import { Flex, FlexItem } from '@patternfly/react-core';
+import { ExclamationTriangleIcon, UserIcon } from "@patternfly/react-icons";
+
+import * as timeformat from "timeformat";
+
+import cockpit from "cockpit";
+
+import './lastLogin.scss';
+
+const _ = cockpit.gettext;
+
+// Do the full combinatorial thing to improve translatability
+const generate_line = (host, line) => {
+    let message = "";
+    if (host && line) {
+        message = cockpit.format(_("from <host> on <terminal>", "from $0 on $1"), host, line);
+    } else if (host) {
+        message = cockpit.format(_("from <host>", "from $0"), host);
+    } else if (line) {
+        message = cockpit.format(_("on <terminal>", "on $0"), line);
+    }
+    return message;
+};
+
+const getFormattedDateTime = (time) => {
+    const now = new Date();
+    const date = new Date(time);
+    if (date.getFullYear() == now.getFullYear()) {
+        return timeformat.dateTimeNoYear(date);
+    }
+    return timeformat.dateTime(date);
+};
+
+const LastLogin = () => {
+    const [messages, setLoginMessages] = useState(null);
+
+    useEffect(() => {
+        if (messages === null) {
+            const bridge = cockpit.dbus(null, { bus: "internal" });
+            bridge.call("/LoginMessages", "cockpit.LoginMessages", "Get", [])
+                    .then(reply => {
+                        const obj = JSON.parse(reply[0]);
+                        if (obj.version == 1) {
+                            setLoginMessages(obj);
+                        } else {
+                        // empty reply is okay -- older bridges just don't send that information
+                            if (obj.version !== undefined)
+                                console.error("unknown login-messages:", reply[0]);
+                        }
+                    })
+                    .catch(error => {
+                        console.error("failed to fetch login messages:", error);
+                    });
+        }
+    }, [messages]);
+
+    if (messages === null || !messages['last-login-time']) {
+        return null;
+    }
+
+    let icon = null;
+    let headerText = null;
+    let underlineText = null;
+    let headerClass = "pf-u-text-break-word";
+    let underlineClass = "pf-u-text-break-word";
+    const lastLoginText = _("Last successful login:") + " " + getFormattedDateTime(messages['last-login-time'] * 1000);
+    const failedLogins = messages['fail-count'];
+
+    if (failedLogins) {
+        let iconClass = "system-information-failed-login-warning-icon";
+        if (failedLogins > 5) {
+            iconClass = "system-information-failed-login-danger-icon";
+            headerClass += " system-information-failed-login-danger";
+        } else {
+            headerClass += " system-information-failed-login-warning";
+        }
+
+        icon = <ExclamationTriangleIcon className={iconClass} size="sm" />;
+        headerText = cockpit.format(cockpit.ngettext("$0 failed login attempt", "$0 failed login attempts", failedLogins), failedLogins);
+        underlineText = getFormattedDateTime(messages['last-login-time'] * 1000) + " " + generate_line(messages['last-login-host'], messages['last-login-line']);
+    } else {
+        icon = <UserIcon className="system-information-last-login-icon" size="sm" />;
+        headerText = lastLoginText;
+        underlineClass += " ct-grey-text pf-u-font-size-sm";
+        underlineText = generate_line(messages['last-login-host'], messages['last-login-line']);
+    }
+
+    return (
+        <li className="last-login" id="page_status_last_login">
+            <Flex flexWrap={{ default: 'nowrap' }}>
+                <FlexItem spacer={{ default: 'spacerSm' }} className="last-login-icon">{icon}</FlexItem>
+                <Flex direction={{ default: 'column' }}>
+                    <FlexItem id="system_last_login"
+                            className={headerClass}
+                            spacer={{ default: 'spacerNone' }}
+                    >
+                        {headerText}
+                    </FlexItem>
+                    <FlexItem id="system_last_login_from"
+                              className={underlineClass}
+                    >
+                        {underlineText}
+                    </FlexItem>
+                    {failedLogins &&
+                    <FlexItem id="system_last_login_success" className="pf-u-text-break-word">
+                        {lastLoginText}
+                    </FlexItem>
+                    }
+                </Flex>
+            </Flex>
+        </li>
+    );
+};
+
+export default LastLogin;
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -24,14 +24,12 @@ import cockpit from "cockpit";
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {
-    Alert, AlertActionCloseButton,
     Page, PageSection, PageSectionVariants,
     Gallery,
     Dropdown, DropdownItem, DropdownToggle, DropdownToggleAction,
 } from '@patternfly/react-core';
 
 import { superuser } from "superuser";
-import * as timeformat from "timeformat";
 
 import { SystemInfomationCard } from './overview-cards/systemInformationCard.jsx';
 import { ConfigurationCard } from './overview-cards/configurationCard.jsx';
@@ -43,121 +41,6 @@ import { SuperuserIndicator } from "../shell/superuser.jsx";
 import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
 
 const _ = cockpit.gettext;
-
-class LoginMessages extends React.Component {
-    constructor() {
-        super();
-        this.state = { messages: {} };
-
-        this.close = this.close.bind(this);
-
-        const bridge = cockpit.dbus(null, { bus: "internal" });
-        bridge.call("/LoginMessages", "cockpit.LoginMessages", "Get", [])
-                .then(reply => {
-                    const obj = JSON.parse(reply[0]);
-                    if (obj.version == 1) {
-                        if (obj['fail-count'] > 5) {
-                            obj.type = "danger";
-                        } else if (obj['fail-count'] > 0) {
-                            obj.type = "warning";
-                        } else {
-                            obj.type = "info";
-                        }
-
-                        this.setState({ messages: obj });
-                    } else {
-                        // empty reply is okay -- older bridges just don't send that information
-                        if (obj.version !== undefined)
-                            console.error("unknown login-messages:", reply[0]);
-                    }
-                })
-                .catch(error => {
-                    console.error("failed to fetch login messages:", error);
-                });
-    }
-
-    close() {
-        const bridge = cockpit.dbus(null, { bus: "internal" });
-        bridge.call("/LoginMessages", "cockpit.LoginMessages", "Dismiss", [])
-                .catch(error => {
-                    console.error("failed to dismiss login messages:", error);
-                });
-
-        this.setState({ messages: {} });
-    }
-
-    render() {
-        const messages = this.state.messages;
-
-        // Do the full combinatorial thing to improve translatability
-        function generate_line(host, line, datetime) {
-            let message = "";
-            if (host && line) {
-                message = cockpit.format(_("<date> from <host> on <terminal>", "$0 from $1 on $2"), datetime, host, line);
-            } else if (host) {
-                message = cockpit.format(_("<date> from <host>", "$0 from $1"), datetime, host);
-            } else if (line) {
-                message = cockpit.format(_("<date> on <terminal>", "$0 on $1"), datetime, line);
-            } else {
-                message = datetime;
-            }
-            return message;
-        }
-
-        function getFormattedDateTime(time) {
-            const now = new Date();
-            const date = new Date(time);
-            if (date.getFullYear() == now.getFullYear()) {
-                return timeformat.dateTimeNoYear(date);
-            }
-            return timeformat.dateTime(date);
-        }
-
-        let last_login_message;
-        if (messages['last-login-time']) {
-            // time is a Unix time stamp, convert to ms since epoch
-            const datetime = getFormattedDateTime(messages['last-login-time'] * 1000);
-            const host = messages['last-login-host'];
-            const line = messages['last-login-line'];
-            last_login_message = generate_line(host, line, datetime);
-        }
-
-        let last_fail_message;
-        if (messages['last-fail-time']) {
-            const datetime = getFormattedDateTime(messages['last-fail-time'] * 1000);
-            const host = messages['last-fail-host'];
-            const line = messages['last-fail-line'];
-            last_fail_message = generate_line(host, line, datetime);
-        }
-
-        let fail_count_message = null;
-        if (messages['fail-count']) {
-            fail_count_message = cockpit.format(cockpit.ngettext(
-                "There was $0 failed login attempt since the last successful login.",
-                "There were $0 failed login attempts since the last successful login.",
-                messages['fail-count']), messages['fail-count']);
-        }
-
-        if (!last_login_message && !fail_count_message && !last_fail_message)
-            return null;
-
-        const last_log_item = <p id='last-login'><b>{_("Last login:")}</b> {last_login_message}</p>;
-        const last_fail_item = <p id='last-failed-login'><b>{_("Last failed login:")}</b> {last_fail_message}</p>;
-
-        return (
-            <Alert id='login-messages'
-                   variant={this.state.messages.type}
-                   isInline={!fail_count_message}
-                   className={!fail_count_message ? "login-messages" : ""}
-                   actionClose={<AlertActionCloseButton onClose={this.close} />}
-                   title={fail_count_message || last_log_item}
-            >
-                {last_login_message && fail_count_message && last_log_item}
-                {last_fail_message && last_fail_item}
-            </Alert>
-        );
-    }
-}
 
 class OverviewPage extends React.Component {
     constructor(props) {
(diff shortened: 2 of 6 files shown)


marusak (dismissed):
The approach seems good, thanks!
Tests needs fixing and also it needs to span across, not just one small card.

jelly:
> It should always span across
> 
> Screenshot from 2021-08-26 14-53-57shouldspan

Resolved by adjusting the scss.

marusak:
> Resolved by adjusting the scss.

Nice, looks good!
Can you please add `Fixes #15786` as well then, thanks

jelly:
> > Resolved by adjusting the scss.
> 
> Nice, looks good!
> Can you please add `Fixes #15786` as well then, thanks

Added, if you did not notice yet.

allisonkarlitskaya (commented):
Sorry for coming late to this party, but I dislike this change.

The reason that this information is shown at login is that the last *valid* login might not have been *you*.

A valid login with an unexpected date or source IP address is actually a much more serious problem than a rejected login, which is why always showing this information is good security practice.

martinpitt:
I agree that hiding it in "system information" is a little too undiscoverable -- in particular, "system information" are static properties about your machine, whereas this is user specific and security-relevant. So perhaps it would at least make more sense in the "Health" card? That one has dynamic information, so at least users are not trained to completely ignore this card the third time they've seen Cockpit.

I'll check the [common criteria](https://www.commoncriteriaportal.org/files/ppfiles/CPP_ND_V2.2E.pdf) whether there is any specific requirement how prominent this information needs to be -- I agree that for the most part humans are just not good enough to spot a  "weird" date/time/IP address in the last login, so from that POV it doesn't have to be that prominent.

marusak:
Fair enough. I think that having it in Heath card with blue (i) icon will do the same, as having it the alert. Honestly, users just learn to ignore the banner as well after couple of times and it takes some significant space off the page. So moving it to Health card will serve to users who are actually interested just fine.

jelly:
> Fair enough. I think that having it in Heath card with blue (i) icon will do the same, as having it the alert. Honestly, users just learn to ignore the banner as well after couple of times and it takes some significant space off the page. So moving it to Health card will serve to users who are actually interested just fine.

Implemented it locally but it seems to me that there is too much information to show or can we do something about "Last login"?

image

marusak:
It does not have to be one line. You can try to style it somehwhat differently:
Random idea:
```
Last login             Sep 02, 05:04PM
                    127.0.0.2 on pts/0
```

pkg/systemd/loginMessage.jsx
```diff
@@ -0,0 +1,50 @@
+import cockpit from "cockpit";
```
jelly, line 1: As this is a new file, it probably requires a copyright notice

marusak: yes, just copy it from other files, just adjust year to 2021

jelly:
The last login information is now moved to the health card and is shown as below:
image

pkg/systemd/overview-cards/lastLogin.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+        if (loginMessages && !('last-login-host' in loginMessages)) {
+            return null;
+        }
+
+        return (
+            <li className="last-login" id="page_status_last_login">
+                <Flex flexWrap={{ default: 'nowrap' }}>
+                    <FlexItem spacer={{ default: 'spacerSm' }}><InfoCircleIcon size="sm" /></FlexItem>
+                    <FlexItem spacer={{ default: 'spacerMd' }} className="system-information-last-login pf-u-text-nowrap">{_("Last login")}</FlexItem>
```
KKoukiou, line 42: This classes need the corresponding CSS files to be imported.

``` diff
diff --git pkg/systemd/overview-cards/lastLogin.scss pkg/systemd/overview-cards/lastLogin.scss
index 43736ac20..a027d1673 100644
--- pkg/systemd/overview-cards/lastLogin.scss
+++ pkg/systemd/overview-cards/lastLogin.scss
@@ -1,3 +1,6 @@
+@import "_global-variables.scss";
+@import '@patternfly/patternfly/utilities/Text/text.scss';
+
 .system-information-last-login {
     font-weight: var(--pf-global--FontWeight--semi-bold);
 }
```

Add this to lastLogin.scss

KKoukiou (changes requested):

KKoukiou:
@garrett @jelly Do we really want the (i) icon on the left of the Last login label? I find it redundant and just taking space:

Without icon:
Screen Shot 2021-09-07 at 11 01 58

With:
Screen Shot 2021-09-07 at 11 01 43

pkg/systemd/overview-cards/lastLogin.jsx
```diff
@@ -0,0 +1,50 @@
...
+import { InfoCircleIcon } from "@patternfly/react-icons";
+
+import cockpit from "cockpit";
+import { LoginMessage } from "../loginMessage.jsx";
+
+import './lastLogin.scss';
+
+const _ = cockpit.gettext;
+
+export class LastLogin extends React.PureComponent {
```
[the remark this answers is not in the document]
jelly: It's not that hard, so I converted it to a functional component.

garrett:
Health is really about the system health though, not about user information. But I guess one could say the same about _system_ information.

Anyone consider usage yet? :laughing: (I mean, it's about the machine's use by the user, I guess?)

The health card does not use a label + info pattern, unlike the system info and usage cards (and configuration card), which all do.

jelly:
> Health is really about the system health though, not about user information. But I guess one could say the same about _system_ information.

The health card was chosen as it is more visible to the user then the system information.

> Anyone consider usage yet?  (I mean, it's about the machine's use by the user, I guess?)
> 
> The health card does not use a label + info pattern, unlike the system info and usage cards (and configuration card), which all do.

I understand that it indeed looks inconsistent with for example the `security updates` row. But I don't think showing it with a clock makes it clear enough for the user that it's the last logged in time.

Is the usage card the way to go?

marusak:
> Health is really about the system health though, not about user information. But I guess one could say the same about system information.

We discussed this already at least 3 times :) We all are aware it is not the ideal, but seems like the best we have to keep it reasonably visible (as a security requirement) and it not being the most important thing on the page. System card is too static for this info for users to notice it.

> Is the usage card the way to go?

I would not think so - it is about "resources usage" and I think having there last login would be even more inconsistent.

> Do we really want the (i) icon on the left of the Last login label? 

I would make the icon blue (as is next to `System time`) to keep it consistent - all other items have icon in this card. But I would not higlight `Last login`, I would just make it normal string "Last login on September 6th..." the same way we don't do "<b>Updates</b>: Security updates available".

garrett:
> We discussed this already at least 3 times

While I was away... Sorry. :wink: 

The icon could be `fa-user` too, perhaps, since it's user information about the last time the user account logged in.

Also, this same area then changes to a warning (with an orange warning icon) in the same place, not an alert, right? If it's under the system health, then there shouldn't be a different warning.

Is this the previous login? Having the last login is probably mostly useless (as you're logged in to see it, so it's assumed to be correct).

Additionally, perhaps we should show the user name there and even link it to the user's account page.

It would look something like this (covering both normal and warning states):

system-overview-health-login

And then, at some point in the future (definitely outside the scope of this PR), we should add in the last several log in times and attempts on the user account page directly. It would basically be a formatted version of `last $USER`.

marusak:
> The icon could be `fa-user` too, perhaps, since it's user information about the last time the user account logged in.

good call! lets do that

> Also, this same area then changes to a warning (with an orange warning icon) in the same place, not an alert, right? If it's under the system health, then there shouldn't be a different warning.

For now it stays as alert. But if you think it would be fine this way, I am all for it - seems nice and visible in the mockups for me.

> And then, at some point in the future (definitely outside the scope of this PR), we should add in the last several log in times and attempts on the user account page directly. It would basically be a formatted version of last $USER.

See our oldest issue - https://github.com/cockpit-project/cockpit/issues/28 :)

garrett:
> For now it stays as alert. But if you think it would be fine this way, I am all for it - seems nice and visible in the mockups for me.

It's better this way. If the previous log in is a health item, then it's a health item... and should not be split between two places. Having it in just the health card presents it in a consistent way. 

And there are already health items that are an alert (which is actually the whole point of the card, right?), so there's definitely precedence.

Please note that I also changed the strings a bit, including the warning title. I did have it the same as it currently is initially, but found that to be too verbose without adding anything of value:

g40178

The long string is just too long here in this context and it doesn't fit in with the rest of the summary strings in the card, which are all short on purpose. Additionally, the wrapping makes it 4 lines instead of 2 like the normal log in info. The title competes with the actual information.

And this will become even more useful once we have the login info on the accounts page.

pkg/systemd/overview-cards/loginMessage.jsx
```diff
@@ -52,14 +53,20 @@ export class LoginMessage extends React.PureComponent {
     }
 
     render() {
-        const { time, host, line } = this.props;
+        const { time, host, line, text, username } = this.props;
         const datetime = this.getFormattedDateTime(time * 1000);
-        return this.generate_line(host, line, datetime);
+        return (
+            <>
+                <Button variant="link" isInline component="span" onClick={() => cockpit.jump(`/users#/${username}`)}>{username}</Button> {text} {this.generate_line(host, line, datetime)}
```
KKoukiou, line 60: Why did you add component='span' here? feels weird to have a span act like a button.

KKoukiou: Could you break the line? It's very long! I have to scroll right to see it on my laptop screen.

pkg/systemd/overview.scss
```diff
@@ -94,7 +94,7 @@
   --card-width: 22rem;
   --pf-l-gallery--GridTemplateColumns: repeat(auto-fill, minmax(var(--card-width), 1fr));
 
-  .motd-box, .login-messages {
+  .motd-box, #login-alert {
```
KKoukiou, line 97: There is not component with id login-alert now

KKoukiou (changes requested):

pkg/systemd/overview-cards/loginMessage.jsx
jelly: > Why did you add component='span' here? feels weird to have a span act like a button.

I was going off the patternfly inline example.

KKoukiou: Which one? They have tons of examples just to show the possible use cases.

pkg/systemd/loginMessage.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import * as timeformat from "timeformat";
+
+const _ = cockpit.gettext;
+
+export class LoginMessage extends React.PureComponent {
```
martinpitt: Using `React.PureComponent` is a new style in cockpit. We usually use functional style, which would seem fine here. You can keep the two helpers as subfunctions inside of it.

pkg/systemd/overview-cards/lastLogin.jsx
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React from 'react';
+import { Flex, FlexItem } from '@patternfly/react-core';
+import { InfoCircleIcon } from "@patternfly/react-icons";
+
+import cockpit from "cockpit";
+import { LoginMessage } from "../loginMessage.jsx";
```
martinpitt: Feels a bit like over-splitting to me -- moving the loginMessage here, as helper function or full component, would keep things simpler and easier to read IMHO.

pkg/systemd/overview.jsx
```diff
@@ -40,46 +39,16 @@ import { UsageCard } from './overview-cards/usageCard.jsx';
 import { SuperuserAlert } from './superuser-alert.jsx';
 import { SuperuserIndicator } from "../shell/superuser.jsx";
 import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
-import { LoginMessage } from "./loginMessage.jsx";
 
 const _ = cockpit.gettext;
 
-class LoginAlert extends React.Component {
```
martinpitt: I was rather confused when I read the first commit and saw it introduce this, only to see that the second commit removes it again. let's squash these two on landling.

martinpitt (commented):
Thanks, this looks nice! I only have some stylistic comments, but no blockers. I leave the final review to @marusak and @KKoukiou, who both still have a -1 on this.

Can you please crop the second screenshot to only show the health card, and balance the border of the first one? So that they look a little nicer in the relnotes? I'll massage the text a little bit, but of course feel free to adjust further.

pkg/systemd/overview.jsx
jelly: I kept it as two comments and want to squash it later, as I wasn't 100% sure everyone agrees on the new behaviour.

pkg/systemd/overview-cards/lastLogin.jsx
jelly: I'll add it as a helper then, I was also a bit on the fence about leaving this separate.

pkg/systemd/overview-cards/lastLogin.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+    if (messages === null || !messages['last-login-time'] || username === null) {
+        return null;
+    }
+
+    let icon = null;
+    let text = null;
+    const failedLogins = messages['fail-count'];
+
+    if (failedLogins) {
```
marusak, line 73: So we no longer have orange and red versions?

pkg/systemd/overview-cards/lastLogin.jsx
```diff
@@ -0,0 +1,105 @@
...
+        icon = <ExclamationTriangleIcon className="system-information-last-login-warning-icon" size="sm" />;
+        text = (
+            <FlexItem id="system_last_login"
+                    className="pf-u-text-break-word system-information-failed-login"
+                    spacer={{ default: 'spacerSm' }}>
+                {cockpit.format(cockpit.ngettext("$0 failed login attempt", "$0 failed login attempts", failedLogins), failedLogins)}
+            </FlexItem>
+        );
+    } else {
+        icon = <UserIcon className="system-information-last-login-icon" size="sm" />;
```
marusak, line 83: Last login is shown only when there is no failed one? I thought that we would always show last login and if tehre is failed one, then also show that one

pkg/systemd/overview-cards/loginMessage.jsx
```diff
@@ -0,0 +1,74 @@
+/*
```
marusak, line 1: Could files `pkg/systemd/overview-cards/loginMessage.jsx` and `pkg/systemd/overview-cards/lastLogin.jsx ` be unified, or is there a good reason why they are two separate files?

pkg/systemd/overview-cards/lastLogin.jsx
KKoukiou: I also asked about this - seems that the design needed the oppositite. @garrett can you clarify this?

garrett: The design assumes that there would be a list of logins on the user page, but we can add both (at least for now, possibly even after) if that would make everyone happier.

garrett:
Alternate design that attempts to:

- Unify the information presented (user information is dropped, so the same string is used in both success and failed states)
- Show successful login info with the failed login state
- Make the link to the account page where login details would be presented more obvious (of course, leave this out until login details are presented on that page)
- Balance everything without showing too much information (This should not be an information dump.)

It does drop the from/on for the successful login (and makes the text smaller) when there are failed login attempts, with the assumption that the failed login attempts should have the focus and that someone who cares about this will look at the details on the user page.

Health-card-login-states

Note: We can proceed with what's already in this PR if everyone agrees it's fine enough. This mockup could be a follow-up, if it's acceptable. I can adjust it further if you 'd like too.

WDYT?

pkg/systemd/overview-cards/lastLogin.jsx
jelly: No, this is no longer in the design made by @garrett shown below.

jelly:
> Alternate design that attempts to:
> 
> * Unify the information presented (user information is dropped, so the same string is used in both success and failed states)
> * Show successful login info with the failed login state
> * Make the link to the account page where login details would be presented more obvious (of course, leave this out until login details are presented on that page)
> * Balance everything without showing too much information (This should not be an information dump.)
> 
> It does drop the from/on for the successful login (and makes the text smaller) when there are failed login attempts, with the assumption that the failed login attempts should have the focus and that someone who cares about this will look at the details on the user page.
> 
> Health-card-login-states(1)
> 
> Note: We can proceed with what's already in this PR if everyone agrees it's fine enough. This mockup could be a follow-up, if it's acceptable. I can adjust it further if you 'd like too.
> 
> WDYT?

I like it! I have updated the PR with the latest design, there will be a follow up required to show the "View login history".

There is one small deviation I have made, of which I am not sure if you like it, I've kept the behaviour where on more then 5 failed logins the icon turns red.

image

What do you think? I am not sure if the text should be this dark red. Following the [patternfly guide](https://www.patternfly.org/v4/guidelines/colors/#primary-colors) I'm not sure what's normal here.

pkg/systemd/overview-cards/lastLogin.jsx
jelly: Implemented, that show a red icon when failed logins > 5.

pkg/systemd/overview.jsx
jelly: Resolved.

pkg/systemd/overview-cards/lastLogin.jsx
jelly: Moved into one file.

martinpitt (dismissed):
Thanks! Very clean code now, and this is a nice self-contained change now. Indeed let's move the "link to accounts page" part to a follow-up. Can you please update the release note text (there is no accounts page link now) and the screenshots (please make sure to capture a consistent frame around the health card)

marusak:
Nice work!
One thing I noticed is that it does not align with other items in the health card. Notice how icons are not centered and also the text does not start at the same place.

notaligned1

KKoukiou (commented):
There is an inconsistency between the fail and normal last login, 
Screen Shot 2021-09-20 at 09 25 14
Screen Shot 2021-09-20 at 09 24 18


The `from` part is missing from the normal last login when there is also a failed login. Secondly the `from` part is small and greyed out in the failed scenario but not in the normal scenario.

I find these inconstencies not really problematic - just unnecessary.
Are these done on purpose?

garrett:
>  Are these done on purpose?

There's way too much information, especially when there's a failed login. The idea is to highlight the most necessary information first, show secondary information in a de-emphasized manner, and have detailed information on the user page.

Here's an adjusted mockup that does not drop the text, but it does change the following:

- Makes all the login information on the success state just normal text
- Removes "successful", as a login is really considered successful, right? And an attempt is not successful.
- Deemphasizes (successful) login text with a lighter secondary grey in the case of attempts, instead of dropping information, similar to "...and X more"

Health-card-login-states-revised(2)


Similar to the previous mockup, there's an "...and more", as without that it makes no sense to say 7 and only show 1. At that point, we might as well not even show the latest, as it's just confusing (unless it says there's more). Also similar to the previous one, in this mockup the secondary information is in a smaller text.

Does this address your concerns?

KKoukiou:
> > Are these done on purpose?
> 
> There's way too much information, especially when there's a failed login. The idea is to highlight the most necessary information first, show secondary information in a de-emphasized manner, and have detailed information on the user page.
> 
> Here's an adjusted mockup that does not drop the text, but it does change the following:
> 
>     * Makes all the login information on the success state just normal text
> 
>     * Removes "successful", as a login is really considered successful, right? And an attempt is not successful.
> 
>     * Deemphasizes (successful) login text with a lighter secondary grey in the case of attempts, instead of dropping information, similar to "...and X more"
> 
> 
> Health-card-login-states-revised(2)
> 
> Similar to the previous mockup, there's an "...and more", as without that it makes no sense to say 7 and only show 1. At that point, we might as well not even show the latest, as it's just confusing (unless it says there's more). Also similar to the previous one, in this mockup the secondary information is in a smaller text.
> 
> Does this address your concerns?

Indeed I like this version more - but let's consider it as a followup - to not delay this pr more.

garrett:
> let's consider it as a followup - to not delay this pr more

Sure, sounds fine.

I think we probably have time to land this PR, make a few fixups to make it more like the latest mockup, and add the login history on the user account page all before the next release.

KKoukiou:
@marusak you still have a nack here, is it obsolete?

marusak:
https://github.com/cockpit-project/cockpit/pull/16260#issuecomment-922690729

I would say this should be looked at. Or you think it is fine?

KKoukiou:
> [#16260 (comment)](https://github.com/cockpit-project/cockpit/pull/16260#issuecomment-922690729)
> 
> I would say this should be looked at. Or you think it is fine?

Oh right, let's fix it. :)

KKoukiou:
> Nice work!
> One thing I noticed is that it does not align with other items in the health card. Notice how icons are not centered and also the text does not start at the same place.
> 
> notaligned1

@jelle adding a flex-basis: 1.25rem; to the FlexItem of the system-information-last-login-icon class solves this issue. Of course this is not the nicest solution - but moving everything to PF4 is out of the scope if this PR.

jelly:
> > Nice work!
> > One thing I noticed is that it does not align with other items in the health card. Notice how icons are not centered and also the text does not start at the same place.
> > notaligned1
> 
> @jelle adding a flex-basis: 1.25rem; to the FlexItem of the system-information-last-login-icon class solves this issue. Of course this is not the nicest solution - but moving everything to PF4 is out of the scope if this PR.

Thanks, pushed a fix

KKoukiou (approved):

KKoukiou (approved):

garrett:
Are the follow-ups implemented yet? It would be good to get them in the release, so the UI doesn't change needlessly... and we should update the screenshots in this PR accordingly.

jelly:
> Are the follow-ups implemented yet? It would be good to get them in the release, so the UI doesn't change needlessly... and we should update the screenshots in this PR accordingly.

Showing the last succesful logins is shown in the accounts details page now in this PR https://github.com/cockpit-project/cockpit/pull/16402#partial-pull-merging

garrett:
> Showing the last succesful logins is shown in the accounts details page now in this PR #16402 (comment)

What about the other follow-ups including making the strings similar in the different states and the link to the accounts page?

mockup

garrett:
Quick reply: Ah, whoops. I didn't see the link on the overview on that PR. I see it's there. Nice.

PR #16273, opened 2021-08-30
webpack-watch: update webpack-make invocation

jelly:
In 8d3a015dcc7c0cf1b the -d option was removed in favor of positional
arguments.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -184,7 +184,7 @@ option for copying the built webpack into the given SSH target's
 SSH `c` alias as described in [test/README.md](./test/README.md), you can use
 one of these commands:
 
-    tools/webpack-make -d dist/kdump/Makefile.deps -r c
+    tools/webpack-make kdump -r c
     tools/webpack-watch kdump -r c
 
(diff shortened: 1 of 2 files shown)


martinpitt:
@jelly : Thanks! There is one wart with "no-test": The fedora-34 context is required, so you need to trigger that one manually.

martinpitt (approved):
Cheers!

allisonkarlitskaya (approved):
This is obviously correct.

Bonus points to anyone who:
 - turns the eslint thing into a `--no-eslint` flag on the argument list of `webpack-make`
 - makes `webpack-watch` into a symlink to `webpack-make`
 - makes `webpack-make` default to `--no-eslint` and `--watch` if its `argv[0]` is `webpack-watch`

martinpitt:
@jelly : Please run the fedora-34 test *after* PR #16277 lands (as that currently blocks everything). Sorry for the mess..

martinpitt:
Fix landed, triggering the test.

PR #16276, opened 2021-08-30
containers: Bump bastion and ws containers to Fedora 34

martinpitt:
Tested with
```
podman build -t cockpit/ws containers/ws
podman build -t cockpit/bastion containers/bastion

podman run -it --rm -p 9999:9090 localhost/cockpit/bastion
# ... and logging into my piware.de server
```

Our CI does not really cover this `FROM:`, as it replaces it with the pre-built `cockpit/base` container on our images. I.e. that already tests on Fedora 34. This only affects the official production container that we push during release.

### DIFF
--- a/containers/bastion/Dockerfile
+++ b/containers/bastion/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:33
+FROM registry.fedoraproject.org/fedora:34
 
 ARG VERSION
 ARG INSTALLER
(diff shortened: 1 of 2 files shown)


martinpitt:
I tested ws by copying the locally built cockpit/ws into a fedora-coreos VM:

    podman save localhost/cockpit/ws | ssh c podman load

and then inside:
```
podman container runlabel INSTALL localhost/cockpit/ws
podman container runlabel --name cockpit-ws RUN localhost/cockpit/ws
```

marusak (approved):
Thanks!

PR #16279, opened 2021-08-30
storage: Wrap edit_config in a mutex

mvollmer:


mvollmer:
Theory so far:

- Updating crypttab (and fstab) via the UDisks2 API is inherently racy: Changing anything about via the API is not reflected immediately in the D-Bus properties that represent its contents. They will only be updated asynchronously via a inotify trigger inside udisksd.
- #13192 has added a couple of background changes to crypttab; we now maintain the "noauto" flag in it autoamtically. These will likely race with the changes triggered in the tests.
- _Maybe_ UDisks2 isn't racy like this anymore in more recent versions, and that's why we see this on RHEL 8 and Centos 8 but not anywhere else.

mvollmer:
>  * Updating crypttab (and fstab) via the UDisks2 API is inherently racy

This is actually false for how Cockpit changes crypttab.  It uses GetSecretConfiguration, which will read crypttab (and fstab) synchronously before changing it.  However, the tests seem to trigger changes too fast and concurrently. They should wait for the expected changes to show up in fstab and crypttab before carrying on.

mvollmer:
https://logs.cockpit-project.org/logs/pull-16279-20210830-135934-fc5c6cc1-rhel-8-5-distropkg/log.html#288-1

mvollmer:
Current plan: serialize calls to edit_config with a Mutex.

mvollmer:
Damn, still fails in the same way... how?

mvollmer:
I can't seem to get clear cur experimental evidence of what is actually going on here... There _might_ have been a failure with the mutex, but I can't reproduce it a second time.  Now I can't even provoke the failure without the mutex.

But the Mutex is the right thing, theoretically, so I say let's put it in.

martinpitt (approved):
Thanks for figuring this out! Let's run with that for a while, and see whether it helps.

However, this just papers over a bug in udisks -- that would be the proper place to serialize file changes, to avoid race conditions.

PR #16280, opened 2021-08-31
systemd: Ignore "To Be Filled By O.E.M." DMI values

### LINKED ISSUE #7476: DMI Info Fallback

martinpitt:
Cheap or self-built machines often have template DMI information which
was not set by the vendor. Ignore such fields and treat them like
empty/absent values. This will trigger the fallback to the board
name/vendor, which in many cases is present and valid; if that does not
work either, just don't show any hardware name/version at all.

Fixes #7476

### DIFF
--- a/pkg/lib/machine-info.js
+++ b/pkg/lib/machine-info.js
@@ -101,7 +101,12 @@ function parseDMIFields(text) {
             return;
         const file = line.slice(0, sep);
         const key = file.slice(file.lastIndexOf('/') + 1);
-        const value = line.slice(sep + 1);
+        let value = line.slice(sep + 1);
+
+        // clean up after lazy OEMs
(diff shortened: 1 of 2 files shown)


martinpitt:
[This](https://logs.cockpit-project.org/logs/pull-16280-20210831-071636-4c080105-fedora-34-firefox/log.html#182) is an unrelated SELinux flake. Handled by https://github.com/cockpit-project/bots/pull/2360 , I will retry once that landed.

marusak (approved):
Danke!

PR #16281, opened 2021-08-31
remove remotectl

allisonkarlitskaya:
## webserver: Drop remotectl utility

[Cockpit 241](https://cockpit-project.org/blog/cockpit-241.html) already stopped using the deprecated `remotectl` utility for creating certificates for the main `cockpit.service`, in favor of the much simpler `/usr/libexec/cockpit-certificate-ensure`. This version moves the remaining usage (containers and documentation) to that helper, and drops `remotectl`.

If you use scripts, playbooks, or similar direct invocations of `remotectl`, please consider removing that: Cockpit creates or refreshes a self-signed certificate on startup automatically if required. If you must pre-create it, please move your scripts to call
`/usr/libexec/cockpit-certificate-ensure`. However, please note that this script is *not* a stable and supported public interface.

PR #16282, opened 2021-08-31
tests: use machine.libexecdir

allisonkarlitskaya:
Remove some hard-coded special cases and replace them with m.libexecdir.


allisonkarlitskaya:
→ #16281

PR #16283, opened 2021-08-31
base1: Fix long shutdown hang of test-dbus{,-address}

martinpitt:
The `NeverReturn()` call causes that unresponded method call to sit in
dbus-daemon's brain, so that shutting down the GTestDBus waits some 25s
for it to time out. This often causes the 60s timeout to be hit
especially during the i386 valgrind runs.

Avoid this by remembering and queueing up all these calls, and complete
them on shutdown. This happens after a unit test is done with all of its
assertions, so is a race free way of avoiding the timeout.

---

I've seen plenty of timed out i386 valgrind tests today, which is annoying. I can't link to an example, as retrying the test re-uses the run ID, and thus the logs are gone. But it looked like this:
```
ERROR: dist/base1/test-dbus-address.html process failed: 124
```

`no-test` as this code only runs in unit tests.


allisonkarlitskaya (changes requested):
Summary from our discussion: we need to find a way to get this message out of "dbus's brain" without using a timeout approach (which could, itself, become a different race).

 - if the service is a separate process, then kill it: that'll cause dbus-daemon to drop the outstanding message object
 - or kill the client
 - if it's in the same process, we can try to disconnect the process from the bus
 - or: queue the method invocation and send the reply just before we tear down the test bus
 - we could try to kill dbus-daemon, but that's maybe a bit ugly
 - maybe we could try to send the message with the "no reply expected" flag set, but I guess that's a bit harder to do without hacks in actual code

martinpitt:
The first idea of closing the client connection:
```diff
                 .always(function() {
                     assert.equal(this.state(), "rejected", "call timed out");
+                    dbus.close();
                     done();
                 });
     });
```
... does not work; it should have been implicitly closed before anyway, as the `var dbus` went out of scope. Apparently the server side does not clean up pending invocations on closing connections, supposedly because it would be actively  hard to terminate pending function calls. So I'll try the next suggestion of remembering and mopping up pending responses before shutdown.

martinpitt:
I tried to actually respond to the NeverReturn() calls in the end, and queue its invocation. I know that it eventually must be a list, but as there only ever is one call for each object, this intermediate state is supposed to work:
```diff
diff --git pkg/base1/test-dbus-common.js pkg/base1/test-dbus-common.js
index a9c038766..ac91dbd45 100644
--- pkg/base1/test-dbus-common.js
+++ pkg/base1/test-dbus-common.js
@@ -50,7 +50,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("call method with timeout", function (assert) {
+    QUnit.only("call method with timeout", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
diff --git src/ws/mock-service.c src/ws/mock-service.c
index ea18e3a18..0652c2165 100644
--- src/ws/mock-service.c
+++ src/ws/mock-service.c
@@ -28,9 +28,11 @@
 #include <string.h>
 
 typedef struct {
+  TestFrobber *frobber;
   GDBusConnection *connection;
   GDBusObjectManagerServer *object_manager;
   GHashTable *other_names;
+  GDBusMethodInvocation *never_return_invocation;
 } MockData;
 
 /* ---------------------------------------------------------------------------------------------------- */
@@ -51,9 +53,11 @@ on_handle_hello_world (TestFrobber *object,
 static gboolean
 on_handle_never_return (TestFrobber *object,
                         GDBusMethodInvocation *invocation,
-                        const gchar *greeting,
                         gpointer user_data)
 {
+  MockData *data = user_data;
+  data->never_return_invocation = g_object_ref (invocation);
+  g_message("XXX on_handle_never_return: queued method invocation %p", data);
   return TRUE;
 }
 
@@ -712,8 +716,15 @@ static void
 mock_data_free (gpointer data)
 {
   MockData *mock_data = data;
+  if (mock_data->never_return_invocation)
+    {
+      g_message ("XXX mock_data_free: finishing pending invoc %p", mock_data->never_return_invocation);
+      test_frobber_complete_never_return (mock_data->frobber, mock_data->never_return_invocation);
+    }
+
   g_object_unref (mock_data->connection);
   g_hash_table_destroy (mock_data->other_names);
+  g_object_unref (mock_data->frobber);
   g_free (mock_data);
 }
 
@@ -728,10 +739,6 @@ mock_service_create_and_export (GDBusConnection *connection,
   MockData *mock_data;
   gchar *path;
 
-  mock_data = g_new0 (MockData, 1);
-  mock_data->connection = g_object_ref (connection);
-  mock_data->other_names = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
-
   /* Test that we can export an object using the generated
    * TestFrobberSkeleton subclass. Notes:
    *
@@ -751,6 +758,12 @@ mock_service_create_and_export (GDBusConnection *connection,
   test_frobber_set_writeonly_property (exported_frobber, "Mr. Burns");
   test_frobber_set_readonly_property (exported_frobber, "blah");
 
+  mock_data = g_new0 (MockData, 1);
+  mock_data->frobber = g_object_ref (exported_frobber);
+  mock_data->connection = g_object_ref (connection);
+  mock_data->other_names = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
+
   object_manager = g_dbus_object_manager_server_new (object_manager_path);
   mock_data->object_manager = object_manager;
   g_object_set_data_full (G_OBJECT (object_manager), "mock-data", mock_data, mock_data_free);
@@ -769,7 +782,7 @@ mock_service_create_and_export (GDBusConnection *connection,
   g_signal_connect (exported_frobber, "handle-hello-world",
                     G_CALLBACK (on_handle_hello_world), NULL);
   g_signal_connect (exported_frobber, "handle-never-return",
-                    G_CALLBACK (on_handle_never_return), NULL);
+                    G_CALLBACK (on_handle_never_return), mock_data);
   g_signal_connect (exported_frobber,
                     "handle-test-primitive-types",
                     G_CALLBACK (on_handle_test_primitive_types), NULL);
diff --git src/ws/test-server.c src/ws/test-server.c
index 27784232c..e3c7f6652 100644
--- src/ws/test-server.c
+++ src/ws/test-server.c
@@ -953,7 +953,9 @@ main (int argc,
   g_main_loop_unref (loop);
 
   g_strfreev (server_roots);
+  g_message("XXX test-server shutting down test bus...");
   g_test_dbus_down (bus);
+  g_message("XXX test-server shut down test bus");
   g_object_unref (bus);
   g_free (bridge_argv);
   g_free (guid);

```

However, it doesn't:
```
$ make check TESTS=./dist/base1/test-dbus-address.html
test-server-Message: 16:42:07.692: XXX on_handle_never_return: queued method invocation 0xc6ba20
> log: ok 1
> log: ok 2 - call method with timeout: call timed out
> log: # test: call method with timeout 
test-server-Message: 16:42:07.732: XXX on_handle_never_return: queued method invocation 0xc8d660
> log: ok 3
> log: ok 4 - call method with timeout : call timed out
> log: 1..4
> log: cockpittest-tap-done
test-server-Message: 16:42:07.751: XXX mock_data_free: finishing pending invoc 0x7f875c00b0b0
test-server-Message: 16:42:07.751: XXX mock_data_free: finishing pending invoc 0x7f875c00b040
test-server-Message: 16:42:07.751: XXX test-server shutting down test bus...

# XXXXXXXXXXX there is a 25s pause here

(test-server:50790): GLib-GIO-WARNING **: 16:42:37.753: Weak notify timeout, object ref_count=1
PASS: dist/base1/test-dbus-address.html 1 
PASS: dist/base1/test-dbus-address.html 2 - call method with timeout: call timed out
PASS: dist/base1/test-dbus-address.html 3 
PASS: dist/base1/test-dbus-address.html 4 - call method with timeout : call timed out
```

src/ws/mock-service.c
```diff
@@ -28,9 +28,11 @@
 #include <string.h>
 
 typedef struct {
+  TestFrobber *frobber;
```
allisonkarlitskaya, line 31: Can't you just directly submit the invocation without needing to keep this around?

allisonkarlitskaya (changes requested):

src/ws/mock-service.c
martinpitt: I'm afraid I don't know what you mean -- `test_frobber_complete_never_return()` needs both the object (TestFrobber) and the invocation data. The alternative would be to keep the individual `object` from `on_handle_never_return()` as well, and store a new struct in the GList (which contains object and invocation); but (1) that would make the whole bookeeping more complicated, and (2) it's always the same object anyway, so I found this a lot simpler.

martinpitt: So I tested
```diff
--- src/ws/mock-service.c
+++ src/ws/mock-service.c
@@ -720,7 +720,7 @@ mock_data_free (gpointer data)
 
   /* answer all NeverReturn() calls now, to avoid service shutdown hang */
   for (GList *elem = mock_data->never_return_invocations; elem; elem = elem->next)
-    test_frobber_complete_never_return (mock_data->frobber, elem->data);
+    test_frobber_complete_never_return (NULL, elem->data);
   g_list_free_full (mock_data->never_return_invocations, g_object_unref);
 
   g_object_unref (mock_data->connection);
```

and this actually seems to work -- is that kosher? I.e. can I *rely* on these (auto-generated) handlers to not require `object`? If so, then I don't need to keep around `frobber`, and the patch would be simpler. But that feels a bit like pushing luck?

martinpitt: Oh, I see the gdbus-codegen autogenerated code:
```
void
test_frobber_complete_never_return (
    TestFrobber *object G_GNUC_UNUSED,
    GDBusMethodInvocation *invocation)
```

and that seems to be the case for all generated test_frobber_complete_*() functions. So if you tell me that this is reliable and safe, I'll be happy to drop it!

allisonkarlitskaya: I mean that you can just call `g_dbus_method_invocation_return_value()`.

martinpitt: Aah, I see -- that works fine, thanks! Updated.

allisonkarlitskaya (approved):
This is really great work, all around: 
  - taking on an annoying flake that has been bugging us forever
  - correctly diagnosing a totally non-obvious problem
  - really nice implementation of the fix

Thanks!  and, of course: :+1:

PR #16284, opened 2021-08-31
test: Allow another variant of pcp-archive messages

martinpitt:
Sometimes `TestPages.testBasic` produces this after looking at network
related pages. Put it back into generally allowed patterns (partial
revert from commit 8bd081d8c).

----

[example](https://logs.cockpit-project.org/logs/pull-16281-20210831-121051-a153938e-debian-testing/log.html#93)


allisonkarlitskaya (approved):
OK.

PR #16287, opened 2021-09-01
test: Drop obsolete libvirt restart hack from check-networkmanager-firewall

martinpitt:
The bug was fixed long ago, and this now gets in the way on Fedora 35
where libvirtd moved to socket activation.

### DIFF
--- a/test/verify/check-networkmanager-firewall
+++ b/test/verify/check-networkmanager-firewall
@@ -62,7 +62,6 @@ class TestFirewall(NetworkCase):
         # many tests expect the "libvirt" zone; on images with libvirt, wait until libvirt adds its zone
         # (older versions don't do that yet)
         if m.image not in ["ubuntu-2004", "ubuntu-stable"]:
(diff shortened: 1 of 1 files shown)


martinpitt:
This should hopefully help for failures like [this (our CI)](https://logs.cockpit-project.org/logs/pull-16281-20210831-184227-dbe34f3f-fedora-35/log.html#56-2) and [this (testing farm)](http://artifacts.dev.testing-farm.io/e76f2ea5-8429-42ad-91dc-26e8e07e37e6/)

martinpitt:
So this cleanup is good -- it succeeded three times (affected test) in our CI on all OSes.

However, on [fedora-33 packit](http://artifacts.dev.testing-farm.io/c5497888-4853-46af-a72d-7c1a51132d72/) I still got the same old race condition, so this does not yet help to actually fix that test flake. Retrying to make sure it wasn't broken by this change.

martinpitt:
#16288 works, closing.

PR #16275, opened 2021-08-30
Enable Fedora 35 in packit, drop/clean up Fedora 33 packit/test/doc

martinpitt:
- [x] PR #16288
 - [x] PR #16289
 - [x] disable fedora-33 (already disabled in release/CI)
 - [x] https://github.com/cockpit-project/bots/pull/2368

PR #16288, opened 2021-09-01
test: Drop libvirt startup in check-networkmanager-firewall

martinpitt:
It is not actually being used any more by tests, it has loads of race
conditions due to being a dynamically added zone, and requires expensive
libvirt test dependencies in FMF/gating tests.

Only keep waiting for the libvirt zone in firewalld if libvirt is
actually running and the default network exists. Otherwise this could
happen during tests and confuse the expected zone count.

---

This is a more intrusive, but better (if it works) alternative to PR #16287.

This should hopefully help for failures like [this (our CI)](https://logs.cockpit-project.org/logs/pull-16281-20210831-184227-dbe34f3f-fedora-35/log.html#56-2) and [this (testing farm)](http://artifacts.dev.testing-farm.io/e76f2ea5-8429-42ad-91dc-26e8e07e37e6/). This is our second and third biggest flake right now.

### DIFF
--- a/test/verify.fmf
+++ b/test/verify.fmf
@@ -25,9 +25,6 @@ require:
   - targetcli
   - tlog
   - tuned
-  # only required for check-networkmanager-firewall, very expensive; TODO: make tests work without these
-  - libvirt-daemon-kvm
-  - libvirt-client
 
 test: ./run-verify-host.sh
 duration: 4h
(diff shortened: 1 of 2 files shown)


martinpitt:
A [single unrelated flake](https://logs.cockpit-project.org/logs/pull-16288-20210901-082742-3b5e484e-fedora-35/log.html#133)  which I fixed in PR #16291. So this works everywhere else *including* packit fedora-35 (in PR #16275), so let's go with this and close #16287.

KKoukiou (approved):
This is reasonable. These tests should not rely on such preconfigured zones as the libvirt one, rather create their own ones.

PR #16290, opened 2021-09-01
selinux: allow bulk deletion of selinux notifications

### LINKED ISSUE #14033: bulk dismiss selinux alerts
It would be nice to have the ability to bulk-dismiss SELinux alerts, this would be particularly helpful when building policies and wanting to wipe the cached entries clean to see what new ones appear.

KKoukiou:
Fixes #14033

image

### DIFF
--- a/pkg/selinux/setroubleshoot-view.jsx
+++ b/pkg/selinux/setroubleshoot-view.jsx
@@ -23,13 +23,13 @@ import React from "react";
 import {
     Alert, AlertGroup, AlertActionCloseButton, Badge, Button,
     Divider,
-    Card, CardHeader, CardTitle, CardBody,
+    Card, CardActions, CardHeader, CardTitle, CardBody,
     ExpandableSection,
     Flex,
     Page, PageSection, PageSectionVariants,
     Switch, Stack, StackItem, Text, TextArea, TextVariants,
 } from "@patternfly/react-core";
-import { ExclamationCircleIcon, ExclamationTriangleIcon, InfoCircleIcon, TrashIcon } from "@patternfly/react-icons";
+import { ExclamationCircleIcon, ExclamationTriangleIcon, InfoCircleIcon } from "@patternfly/react-icons";
 
 import { Modifications } from "cockpit-components-modifications.jsx";
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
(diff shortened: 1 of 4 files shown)


pkg/selinux/setroubleshoot-view.jsx
```diff
@@ -386,8 +386,6 @@ export class SETroubleshootPage extends React.Component {
                     { title: itm.description }
                 ];
                 if (itm.count > 1) {
-                    title = cockpit.format(cockpit.ngettext("$0 occurrence", "$0 occurrences", itm.count),
-                                           itm.count);
```
martinpitt, line 390: It's being used in line 406 below in the `<CardTitle>`. Indeed it makes not much sense to change it to *this* string, as the default is "SELinux access control errors". But I feel this requires a little more research what this was really meant to do?

martinpitt (commented):
I'll do a proper review tomorrow, just one quick remark

KKoukiou:
@martinpitt thanks for the thoughts and insight on the design! I totally agree with your remarks. Your suggestion sounds good solution indeed!  Let me give it a try :)

pkg/selinux/setroubleshoot-view.jsx
KKoukiou: Explained it in the commit message what went wrong - found the falty commit in the git history.

martinpitt:
Yay, this is much nicer IMHO:

image

I'll make a screenshot with two selections for the release notes, put it into the description. Thanks!

pkg/selinux/setroubleshoot-view.jsx
```diff
@@ -318,6 +310,12 @@ export class SETroubleshootPage extends React.Component {
         e.stopPropagation();
     }
 
+    onSelect(_, isSelected, rowId) {
+        const selected = Object.assign(this.state.selected);
+        selected[this.props.entries[rowId / 2].key] = isSelected;
```
martinpitt, line 315: Divisions in an array index, and doing arithmetic with an ID both look weird..  That could at least use a comment. There is somehow two rows for an entry? Can we be sure the ID is even? Can we even be sure that it contains a number? (type-wise, it's a string, no?) Could it be `//` to make sure it uses integer division, or perhaps avoid arithmetic at all?

KKoukiou: There is a row also for the expandable items. Therefore it's tr = number of rows * 2.

onSelect returns an index, not an DOM id. That's a confusing name for this variable indeed. 

For avoiding arithmetic at all I need to rewrite cockpit-components-table.jsx to use the new PF4 table (this uses an deprecated PF4 table) which I want to do - but out of the scope of this PR.

martinpitt: Thanks for the explanation -- so this will disappear eventually, good.

martinpitt (approved):

PR #16292, opened 2021-09-01
make webpack-watch a symlink of webpack-make

jelly:
This removes the wrapper script `webpack-watch` and makes `webpack-make` if it's called as `webpack-watch`.

### DIFF
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -15,8 +15,13 @@ const parser = argparse.ArgumentParser()
 parser.add_argument('-c', '--config', { help: "Path to webpack.config.js", default: "webpack.config.js" })
 parser.add_argument('-r', '--rsync', { help: "rsync webpack to ssh target after build", metavar: "HOST" })
 parser.add_argument('-w', '--watch', { action: 'store_true', help: "Enable webpack watch mode" })
+parser.add_argument('-e', '--no-eslint', { action: 'store_true', help: "Disable eslint linting" })
 parser.add_argument('prefix', { help: "The directory to build (eg. base1, shell, ...)", metavar: "DIRECTORY" })
 args = parser.parse_args()
+
+if (args.no_eslint) {
+    process.env["ESLINT"] = "0";
+}
 
 if (args.prefix.includes('/')) {
     parser.error("Directory must not contain '/'")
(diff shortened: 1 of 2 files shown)


jelly:
I've left out keeping the symlinks in the dist tarball as I'm not 100% sure what the implications are if we don't pass `-h` for the symlinks in the `test` folder.

tools/webpack-make
```diff
@@ -11,11 +11,14 @@ const argparse = require("argparse");
...
+// argv0 is node
+const webpack_watch = process.argv[1].includes('webpack-watch');
+
 const parser = argparse.ArgumentParser()
 parser.add_argument('-c', '--config', { help: "Path to webpack.config.js", default: "webpack.config.js" })
 parser.add_argument('-r', '--rsync', { help: "rsync webpack to ssh target after build", metavar: "HOST" })
-parser.add_argument('-w', '--watch', { action: 'store_true', help: "Enable webpack watch mode" })
-parser.add_argument('--no-eslint', { action: 'store_true', help: "Disable eslint linting" })
+parser.add_argument('-w', '--watch', { action: 'store_true', help: "Enable webpack watch mode", default: webpack_watch })
+parser.add_argument('--no-eslint', { action: 'store_true', help: "Disable eslint linting", default: webpack_watch })
```
martinpitt, line 21: I wonder if we should enable eslint by default for watching as well. Personally I find it sometimes useful, sometimes it gets in the way -- it's not really about the time (incremental builds with eslint don't take much longer), but more about whether you are hacking on something where "unused function" etc. is annoying. For consistency I'd generally enable it, and provide an option to disable it. @KKoukiou , @mvollmer , @marusak , do you have an opinion about changing the default?

Also, could this get a short option, like `-e`, as it is rather common?

martinpitt (changes requested):
Thanks! I'm really glad to see that nasty shell wrapper gone. Supposedly this gets duplicated in the shipped tarball? (Doesn't matter much, it's not a tool you'd use on release tarballs anyway)

jelly:
> Thanks! I'm really glad to see that nasty shell wrapper gone. Supposedly this gets duplicated in the shipped tarball? (Doesn't matter much, it's not a tool you'd use on release tarballs anyway)

Yes, it will get duplicated in the tarball, which we can fix but I am not 100% sure about the implications for the other symlinks in our tarball.

martinpitt (approved):
Thanks! No feedback so far, so let's land this (keeping the current behaviour). Triggering required f34 test.

PR #16294, opened 2021-09-02
package.json: Move argparse to devDependencies

martinpitt:
It is only being used by the build-time tools pkg/lib/*2po and
webpack-make, none of which are used during runtime.


martinpitt:
As this is trivial, I added a second commit to fix [another test flake](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16275-20210902-050457-45544122-debian-stable/log.html#167)

KKoukiou (approved):

jelly (approved):

Issue #14978, opened 2020-11-27
track package-lock.json in git


lnussel:
AFAICT so far releases are done by tagging a commit in git. Then a ci job takes that, modifies the spec file to include the tag as version, runs npm etc to produce the minified files in dist/ as well as `package-lock.json`.

Would it be possible to commit back the spec file and the `package-lock.json` to git? Could be a separate release-XXX tag for example. That would allow to recreate the dist content and build a package in a separate build system such as OBS just by fetching sources from upstream git.

KKoukiou:
@martinpitt with make_dist @lnussel can do exactly what he needs right?

martinpitt:
We have been talking about tracking package-lock.json in git in general, to avoid some pitfalls when some random npm modules update and regress. That would solve your issue as well.

@KKoukiou : make_dist is unrelated -- our release tarballs have included package-lock.json for a long while alreay, so if your build system is ok with downloading anything, then it's better to download the official release tarball rather than some git tag.

But if it helps in the meantime, we can publish the package-lock.json as a separate release artifact on GitHub, so that you only have to download that instead of the 8 MB full tarball.

martinpitt:
We actually have that now: Our node_modules is [tracked in git](https://github.com/cockpit-project/node-cache/). So if you are building from cockpit.git, `tools/node_modules checkout` will fetch the correct tag from the node cache. This includes package-lock.json, but also all the actual modules right away. Is that suitable for you @lnussel , or is there still a step missing?

KKoukiou:
Closing as there is not reply from the reporter and in my opinion the reported issue is already addressed.

lnussel:
ah sorry, I missed the question. No it's not really what I had mind. Also doesn't include the spec file modifications :-)

martinpitt:
> No it's not really what I had mind.

What is missing then?

lnussel:
something like this https://github.com/cockpit-project/cockpit/compare/251...lnussel:251-dist?expand=1

You don't have to. No is a valid answer :-)

martinpitt:
@lnussel : You can get package-lock.json from several places: Our node-cache has both the full `node_modules/` as git tree, as well as https://github.com/cockpit-project/node-cache/blob/252/.package-lock.json. It's also contained in the release tarballs.

[...]

PR #16296, opened 2021-09-02
workflows: Tag node-cache on releases

martinpitt:
We want our node-cache to have tags corresponding to cockpit releases
[1], so that we get a proper replacement for the old manually built
cockpit-cache-XXX.tar.xz artifacts. These are useful for having a
straightforward way for taking a release tarball, getting a matching
node_modules, and thus being able to patch/build cockpit from a release
(i.e. what distributions do all the time).

So far we have manually tagged node-cache for a few releases, and it's
part of our release process [2], but we have forgotten a few times --
and this is perfectly automatable.

Add this to our standard release workflow. It must run in a separate
job, as it uses the deploy key from the "node-cache" environment; this
secret is (deliberately) not present in the "release" environment for
the container. Delete the tag on the GitHub repo first, so that this
stays idempotent -- we want to be able to re-run a failed release.

[1] https://github.com/cockpit-project/node-cache/tags
[2] https://github.com/cockpit-project/cockpit/wiki/ReleaseProcess

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,3 +33,35 @@ jobs:
           export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
           cd /build
           release-runner -r https://github.com/$GITHUB_REPOSITORY -t $(basename $GITHUB_REF) tools/cockpituous-release
+
+  release-node-cache:
+    runs-on: ubuntu-latest
+    environment: node-cache
+    # done via deploy key, token needs no write permissions at all
+    permissions:
+      contents: read
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set up git
+        run: |
+            git config user.name "GitHub Workflow"
+            git config user.email "cockpituous@cockpit-project.org"
+
+      - name: Tag node-cache
+        run: |
+          set -eux
+          TAG="$(basename $GITHUB_REF)"
+
+          tools/node-modules checkout
+          cd node_modules
+          git tag "$TAG"
+
+          git remote add cache "ssh://git@github.com/${GITHUB_REPOSITORY%/*}/node-cache"
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.DEPLOY_KEY }}'
+          # make this idempotent: delete an existing tag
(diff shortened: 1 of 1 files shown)


martinpitt:
For testing this, I [copied this new job to the po-refresh workflow](https://github.com/cockpit-project/cockpit/commit/de8eb19f83a3d7ee0c299ab0a1554c97c40f7e84), so that I could trigger it from https://github.com/cockpit-project/cockpit/actions/workflows/po-refresh.yml (workflow_run actions must be *present* on main for them to appear there). The only difference to the real workflow here is that I replaced GITHUB_REF with a hardcoded "250", as there is no way to manually trigger from a branch and simulate a "on: tag" event.

The [workflow succeeded](https://github.com/cockpit-project/cockpit/runs/3494087007?check_suite_focus=true) and created https://github.com/cockpit-project/node-cache/releases/tag/250 .

Now I can do this:
```
curl -L -O https://github.com/cockpit-project/cockpit/releases/download/250/cockpit-250.tar.xz
tar xf cockpit-250.tar.xz
cd cockpit-250/
curl -L https://github.com/cockpit-project/node-cache/archive/refs/tags/250.tar.gz | tar -C node_modules --strip-components=1 -xz
```

The last command is not exactly obvious, but it is robust and predictable, and does not need any knowledge of git links or SHAs (which are not a thing from release tarballs).

With these I can patch stuff, and rebuild:

   ESLINT=0  tools/webpack-make -d dist/sosreport/Makefile.deps

(need to disable ESLINT as we fail to ship .eslint* in tarballs -- we may want to fix that).

So, to make this classy we could introduce a tools/ script to fetch node_modules for an unpacked release tarball -- but at least it's possible now. This lays the ground work for fixing #14978 and https://bugzilla.redhat.com/show_bug.cgi?id=1969450 properly.

martinpitt:
After this lands, I'll update https://github.com/cockpit-project/cockpit/wiki/ReleaseProcess to remove the tagging step.

KKoukiou (approved):
Not feeling super good to review this one. However the workflow is very descriptive and simple. Keys cleanups are also done in the end, and since you already tested it I can't see what else can go wrong here.

PR #16298, opened 2021-09-02
[r85] containers: Fix SELinux context of SSH private key

martinpitt:
In Fedora ≥ 34, the container is not able to read the passed key any
more due to SELinux restrictions. Use the designated docker/podman
volume flag `:Z` to fix that.

Backported from main commit 5954eedaae0cf18

### DIFF
--- a/containers/bastion/README.md
+++ b/containers/bastion/README.md
@@ -7,7 +7,7 @@ By default on login cockpit will attempt to establish a ssh connection to the gi
 
 You can also mount an encrypted private key inside the container and set the environment variable `COCKPIT_SSH_KEY_PATH` to point to it.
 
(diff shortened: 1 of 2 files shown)


martinpitt:
@jelly : If you have a déjà vu on that one: We have a set of stable branches like rhel-8.X which get requested bug fixes, and sometimes also some necessary test updates to keep up with changes in test images and our CI. This is such a case which blocks https://github.com/cockpit-project/bots/pull/2377 . So the rule is: "Everything that goes into stable branches first lands in master", unless of course it's something truly specific like commit 6d92f4bdce8e9. We label these as "backport".

Content/patch-wise, these are completely trivial. The thing to review for stable branches is policy (does this make sense to get backported?), not patch content.

jelly (approved):

PR #16299, opened 2021-09-02
Introduce tests on mobile screen sizes

KKoukiou:
Tested locally with: TEST_MOBILE=true TEST_OS=fedora-34 test/verify/check-networkmanager-basic -st TestNetworkingBasic.testBasic

It created a: TestNetworkingBasic-testBasic-network-main-mobile-pixels.png file as expected.

### DIFF
--- a/test/common/cdp.py
+++ b/test/common/cdp.py
@@ -75,12 +75,19 @@ class CDP:
         self.inject_helpers = inject_helpers
         self.browser = os.environ.get("TEST_BROWSER", "chromium")
         self.show_browser = bool(os.environ.get("TEST_SHOW_BROWSER", ""))
+        self.mobile = bool(os.environ.get("TEST_MOBILE", ""))
         self.download_dir = tempfile.mkdtemp()
         self._driver = None
         self._browser = None
         self._browser_home = None
         self._browser_path = None
         self._cdp_port_lockfile = None
+        if not self.mobile:
+            self.window_width = "1920"
+            self.window_height = "1200"
+        else:
+            self.window_width = "414"
+            self.window_height = "1920"
 
     def invoke(self, fn, **kwargs):
         """Call a particular CDP method such as Runtime.evaluate
@@ -175,7 +182,7 @@ class CDP:
                     "--disable-namespace-sandbox", "--disable-seccomp-filter-sandbox",
                     "--disable-sandbox-denial-logging", "--disable-pushstate-throttle",
                     "--font-render-hinting=none",
-                    "--v=0", "--window-size=1920x1200", "--remote-debugging-port=%i" % cdp_port, "about:blank"]
+                    "--v=0", f"--window-size={self.window_width}x{self.window_height}", f"--remote-debugging-port={cdp_port}", "about:blank"]
         elif self.browser == "firefox":
             subprocess.Popen([exe, "--headless", "--no-remote", "-CreateProfile", "blank"], env=env).communicate()
             profile = glob.glob(os.path.join(self._browser_home, ".mozilla/firefox/*.blank"))[0]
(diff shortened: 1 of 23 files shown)


test/common/cdp.py
```diff
@@ -170,12 +171,15 @@ def browser_cmd(self, cdp_port, env):
         if not exe:
             raise SystemError(self.browser + " is not installed")
 
+        ws = "1920x1200"
```
marusak, line 174: on line 204 we set size of firefox browser. I would say that setting `TEST_MOBILE` env should work for both (even if we don't have context for it right now)

test/run
```diff
@@ -14,6 +14,7 @@ case $TEST_SCENARIO in
     verify|devel|firefox|firefox-devel)
         [ "${TEST_SCENARIO}" = "${TEST_SCENARIO%%devel}" ] || export NODE_ENV=development
         [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##firefox}" ] || export TEST_BROWSER=firefox
+        [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##mobile}" ] || export TEST_MOBILE=""
```
martinpitt, line 17: This needs to be set to an actual value.

martinpitt (changes requested):

martinpitt:
So this PR would mean that we would immediately need to provide -mobile versions of all 14 pixel tests, right? That seems fine to me.

KKoukiou:
> So this PR would mean that we would immediately need to provide -mobile versions of all 14 pixel tests, right? That seems fine to me.

Correct.

test/common/cdp.py
KKoukiou: Fixed.

test/run
KKoukiou: Fixed.

KKoukiou:
@marusak @martinpitt What do you think about increasing the vertical size (a lot) in order that we get more content in each mobile screenshot? 

Now they look like:
TestNetworkingBasic-testBasic-network-main-mobile-pixels


Increasing the vertical size only want make us lose something from the expected behavior of the CSS etc so I think it's a win - win.

martinpitt:
@KKoukiou : Agreed -- I think all our CSS only looks at the width. Talller screens are more useful to visually debug test failures, and we can still imagine how a real screen would look like.

test/common/cdp.py
```diff
@@ -197,7 +201,10 @@ def browser_cmd(self, cdp_port, env):
             with open(os.path.join(profile, "handlers.json"), "w") as f:
                 f.write('{"defaultHandlersVersion":{"en-US":4},"mimeTypes":{"application/xz":{"action":0,"extensions":["xz"]}}}')
 
-            cmd = [exe, "-P", "blank", "--window-size=1920,1200", "--remote-debugging-port=%i" % cdp_port, "--no-remote", "localhost"]
+            ws = "1920,1200"
+            if self.mobile:
+                ws = "414,736"  # iPhone 6/7/8 Plus
```
martinpitt, line 206: Please don't repeat the numbers -- can you set self.width and self.height in the constructor and just assemble the strings here? (comma vs 'x')

test/verify/check-superuser
```diff
@@ -245,6 +261,23 @@ class TestSuperuserOldShell(MachineCase):
         "machine2": {"address": "10.111.113.2/20", "image": "centos-7"},
     }
 
+    def checkSuperuserIndicator(self, access):
```
martinpitt, line 264: These two helpers seem common enough to move to the `Browser` class in testlib.py, as they would often even apply to third-party projects. Even now they are already copied four times.

martinpitt (approved):
Very cool! This is conflict prone and will get some follow-ups anyway, so let's defer the helper factorization and get this in ASAP. (Just waiting for packit)

Thanks a lot for working on this! :heart:

test/verify/check-superuser
KKoukiou: Sure- will move it right after.

martinpitt:
packit failures are due to non-rebased branch. The test does run in general.

PR #16300, opened 2021-09-03
services: fix missing icons from the services details UI

KKoukiou:
pficon classes seem to not work any more, at least for the services
page.

Start using the icons from @patternfly/react-icons, which will make the
icons reappear.

This also enables us to stop including PF3 in this page.


martinpitt (dismissed):
Awesome, thank you! Yay less PF3.

martinpitt (approved):
Hah, nice that you found another one! :+1:

jelly (approved):

PR #16303, opened 2021-09-05
lib: patternfly: remove custom workarounds which are already fixed in PF upstream

KKoukiou:


martinpitt:
The [failing pixel tests](https://logs.cockpit-project.org/logs/pull-16303-20210905-184914-a34fdab9-fedora-34/log.html#60) show some wiggling for the button positions, but I didn't see some substantial error. So I think they merely need to get updated?

KKoukiou:
Oh this needs indeed pixel tests updated. I will let https://github.com/cockpit-project/cockpit/pull/16299 get in first then - pixel tests rebases are cumbersome.

KKoukiou:
Just rebased - and re-run pixel tests to check what tests need update.

martinpitt (approved):
Nice!

PR #16306, opened 2021-09-06
test: Port pixeldiff.html from jQuery to standard DOM API

martinpitt:
I tested this with

- downloading a few pixel diff pngs from https://logs.cockpit-project.org/logs/pull-16303-20210905-184914-a34fdab9-fedora-34/ 
 - run `python3 -m http.server` in test/common
 - open http://localhost:8000/pixeldiff.html#TestStorageLuks-testLuks-tab in browser, and reload after every change

It now looks and behaves identical to the original.


martinpitt:
I want to do a few things away from jQuery in cockpit, and this was a nice warm-up exercise.

KKoukiou (approved):

PR #16308, opened 2021-09-06
systemd: Port system time change from bootstrap to PF4 date picker

martinpitt:
This makes manual date picking/setting actually work. The bootstrap one
showed a wildly wrong current date, and did not actually set an updated
time.

------

On main: Notice the completely garbled date (I did not change anything, just clicked once into the date input):

systime-datepicker-main

On this PR:

systime-datepicker-pr


KKoukiou (approved):
Nice thanks :) This would deserve a pixel tests to not allow us to break it again but let's add it in a follow up :)

PR #16310, opened 2021-09-06
playground: Port from jQuery to DOM/JS standard API

martinpitt:
This shaves off 1.3 MB from that webpack.


marusak (approved):
Nice, thanks :)

PR #16312, opened 2021-09-06
tools: Provide eslint mode for webpack-watch

martinpitt:
Since commit 92acdea1e1ee3 it is not possible any more to run
webpack-watch with eslint, as there is no option to enable it, and the
program overwrites `ESLINT=1` from the environment.

Drop the weird inconsistency between webpack-{make,watch}), and default
to running eslint in both modes. Then `--no-eslint` can be used to
disable it.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -171,10 +171,8 @@ which reduces the build time to less than a third. E. g.
 
     $ tools/webpack-watch systemd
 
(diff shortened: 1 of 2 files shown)


jelly:
I guess the overhead from running eslint in watch mode isn't that high so it should be fine to enable by default.

jelly (approved):

PR #16318, opened 2021-09-07
shell: credentials code cleanup and partial de-jquery

martinpitt:


martinpitt:
I tried to port pkg/shell/credentials.js away from jQuery, but it uses too much magic from it. IMHO it's better to just rewrite this into React.

[WIP patch](https://github.com/cockpit-project/cockpit/files/7121811/cockpit-WIP-shell-Port-credentials-from-jQuery-to-standard-D.patch.txt). to avoid losing it in case it is still useful for something

test/verify/check-shell-multi-machine-key
```diff
@@ -70,17 +70,17 @@ class TestMultiMachineKeyAuth(MachineCase):
         self.browser.eval_js("loaded = false")
         self.browser.eval_js("""
             load = function (user) {{
-                var proc = cockpit.spawn([ 'ssh-add', '{0}' ], {{ pty: true, directory: user.home + '/.ssh' }})
-                    .stream(function(data) {{
+                const proc = cockpit.spawn([ 'ssh-add', '{0}' ], {{ pty: true, directory: user.home + '/.ssh' }});
```
marusak, line 73: No need to store this in variable, but does not really matter.

marusak (approved):
Thank you!

test/verify/check-shell-multi-machine-key
martinpitt: How do you mean? It's being used in the `.stream` handler, it refers to the original `cockpit.spawn()` return object. What you *can't* do (I learned from debugging) is

    const proc = cockpit.spawn().stream().then().catch()

and use *that* value -- at that time it is a transformed promise, which does not have an `.input()` function any more.

marusak: Oh, it is being used for `input()` - missed that, nevermind :)

PR #16320, opened 2021-09-07
Fix some selinux page UI issues 

### LINKED ISSUE #16311: selinux: Layout has issues, both horizontally and vertically
This UI needs to be redesigned and reimplemented so it's easier to understand and perform actions. (These actions are mutually exclusive, but not presented as such in the UI. And they're currently badly worded.) But redoing the UI will probably take a while.

In the meantime, there are some issues with the current implementation's layout, making it hard to read on both the desktop and especially on mobile. These should _hopefully_ be relatively easy to fix.

## Horizontal

SELinux solutions have a `colspan="4"` when there are 5 columns in the row, causing the content to not stretch to the right edge, so there's a gap in both desktop and mobile views. (Clicking on the image should hopefully show it with a different background from white, so it's easier to see the issue.)

Illustration:
Screenshot from 2021-09-06 17-39-53

The blue content should stretch to the edge. Setting the `colspan` to 5 instead fixes this layout issue.

## Vertical

Additionally, there's no vertical spacing between elements, so it's just a bunch of text and items jumbled together. It's like a "wall of text" without spacing to show what elements belong to others.

Here's an example of what I'm talking about:

Screenshot from 2021-09-06 17-50-38

And when in mobile mode, these two issues make the text extremely hard to read:

Screenshot 2021-09-06 at 17-47-50 SELinux - garrett Bolt

## Tabs

Additionally, the tabs are wrapped weirdly in mobile mode. (See screenshot above.)

---

So there are 3 layout issues here with this current UI:

1. Horizontal, cause by a colspan that's off-by-one
2. Vertical, caused by using either unsemantic elements, the wrong elements, or some CSS that is overdefined or not currently defined. (I haven't looked.)
3. Tabs wrap oddly.

I can solve the vertical issues and the tabs, but I need someone else to look into the horizontal `colspan` issue, as `colspan` doesn't seem to be specified in the code, oddly.

KKoukiou:
Fixes #16311

### DIFF
--- a/pkg/lib/cockpit-components-listing-panel.scss
+++ b/pkg/lib/cockpit-components-listing-panel.scss
@@ -7,6 +7,7 @@
     );
     display: flex;
     align-items: flex-end;
+    padding-right: var(--pf-global--spacer--lg);
 
     .pf-c-nav {
       margin-left: var(--pf-global--spacer--xs);
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -148,7 +148,7 @@ export class ListingTable extends React.Component {
         return res;
     }
 
-    reformatRows(rows) {
+    reformatRows(rows, colSpan) {
         let rowIndex = 0;
         return rows.reduce((total, currentValue) => {
             const rowFormatted = {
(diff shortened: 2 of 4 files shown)


garrett:
Looks great overall!

Switching tabs makes columns shift around. They change size for whatever reason (tables with colspan are finicky, really). You can easily see this in action, as the badge changes positions as it's not right-aligned.

Here are screenshots where you can see the inspector show different widths for the same element after switching tabs:


Screenshot from 2021-09-07 18-08-31

Screenshot from 2021-09-07 18-08-35

But aside from that, the layout is already much, much better!

marusak (approved):
Nice!
The issue that this PR fixes still talks about wrapping tabs, which is still the case. I think the line with tabs and date just needs to be flex so it wraps and using standard PF tabs so on small screens they behave as expected. Of course, this is fine for another PR as well.

garrett:
Oh. I was working on these issues, as mentioned in the standup, and just pushed up some bugfix changes. :neutral_face:

Issue #16327, opened 2021-09-08
Fix checkboxes without labels


garrett:
As mentioned in https://github.com/cockpit-project/cockpit-podman/pull/750#issuecomment-914105290:

> We should never have checkboxes without labels. They're inaccessible and the click target is much smaller (as the label itself is part of the click target when done in a proper, accessible way) and that benefits everyone, even people with full sight using a mouse.
> 
> Clarifications: 
> 1. There needs to be a left-side label for the group of checkbox options, so they are distinct from the options above. This is different from the checkbox's label.
> 2. Yes, this issue exists elsewhere in Cockpit, such as Virtual Machines. It should also be fixed.

First step: Identifying where the problems are.

Here's a list:

- Podman's create container dialog
- Machine's create new VM dialog
- Network's bridge settings (Spanning tree protocol (STP))

I went through all of Cockpit (at least what's currently installed on my system), and those are the only places I saw. Please add to it if you find something else.

garrett:
We fixed the network dialog. We're working on the Container redesign. Machines has an aria-labelled-by at least.

PR #16328, opened 2021-09-08
network: Change STP label and group options

garrett:
Related: #16327

### DIFF
--- a/pkg/networkmanager/index.html
+++ b/pkg/networkmanager/index.html
@@ -104,27 +104,30 @@
         {{! member interfaces will be rendered into here as a list of checkboxes }}
       </div>
 
-      <label class="control-label" for="network-bridge-settings-stp-enabled-input" translate="yes">Spanning tree protocol (STP)</label>
+      <label class="control-label" for="network-bridge-settings-stp-enabled-input" translate="yes">Options</label>
       <label class="checkbox-inline">
         <input id="network-bridge-settings-stp-enabled-input" type="checkbox"
           {{#stp_checked}}checked{{/stp_checked}}/>
+        <span translate="yes">Spanning tree protocol (STP)</span>
       </label>
 
-      <label class="control-label" for="network-bridge-settings-stp-priority-input" translate="yes">STP priority</label>
-      <input id="network-bridge-settings-stp-priority-input" class="form-control" type="text"
-             value="{{stp_priority}}"/>
+      <div role="group" class="ct-form" id="network-bridge-settings-stp">
+        <label class="control-label" for="network-bridge-settings-stp-priority-input" translate="yes">STP priority</label>
+        <input id="network-bridge-settings-stp-priority-input" class="form-control" type="text"
+              value="{{stp_priority}}"/>
 
-      <label class="control-label" for="network-bridge-settings-stp-forward-delay-input" translate="yes">STP forward delay</label>
-      <input id="network-bridge-settings-stp-forward-delay-input" class="form-control" type="text"
-             value="{{stp_forward_delay}}"/>
+        <label class="control-label" for="network-bridge-settings-stp-forward-delay-input" translate="yes">STP forward delay</label>
+        <input id="network-bridge-settings-stp-forward-delay-input" class="form-control" type="text"
+              value="{{stp_forward_delay}}"/>
 
-      <label class="control-label" for="network-bridge-settings-stp-hello-time-input" translate="yes">STP hello time</label>
-      <input id="network-bridge-settings-stp-hello-time-input" class="form-control" type="text"
-             value="{{stp_hello_time}}"/>
+        <label class="control-label" for="network-bridge-settings-stp-hello-time-input" translate="yes">STP hello time</label>
+        <input id="network-bridge-settings-stp-hello-time-input" class="form-control" type="text"
+              value="{{stp_hello_time}}"/>
 
-      <label class="control-label" for="network-bridge-settings-stp-max-age-input" translate="yes">STP maximum message age</label>
-      <input id="network-bridge-settings-stp-max-age-input" class="form-control" type="text"
-             value="{{stp_max_age}}"/>
+        <label class="control-label" for="network-bridge-settings-stp-max-age-input" translate="yes">STP maximum message age</label>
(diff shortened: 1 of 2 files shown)


garrett:
## Before

Screenshot 2021-09-08 at 11-17-43 Networking - garrett Bolt

Screenshot 2021-09-08 at 11-17-57 Networking - garrett Bolt


## After

Screenshot 2021-09-08 at 11-11-58 Networking - garrett Bolt

Screenshot 2021-09-08 at 11-20-33 Networking - garrett Bolt

pkg/networkmanager/index.html
```diff
@@ -104,27 +104,30 @@ <h1 id="testing-connection-curtain-restoring" translate="yes">Restoring connecti
         {{! member interfaces will be rendered into here as a list of checkboxes }}
       </div>
 
-      <label class="control-label" for="network-bridge-settings-stp-enabled-input" translate="yes">Spanning tree protocol (STP)</label>
+      <label class="control-label" for="network-bridge-settings-stp-enabled-input" translate="yes">Options</label>
       <label class="checkbox-inline">
         <input id="network-bridge-settings-stp-enabled-input" type="checkbox"
           {{#stp_checked}}checked{{/stp_checked}}/>
+        <span translate="yes">Spanning tree protocol (STP)</span>
```
KKoukiou, line 111: This STP checkbox is about enabling / disabling STP, and then it exposes some options. Therefore I think this needs different handling

How about labeling the checkbox that enables STP and exposes these options like: `Use spanning tree mode` and then have the Options bellow?

Otherwise this can also use a Switch instead of a checkbox.

KKoukiou (changes requested):

pkg/networkmanager/index.html
KKoukiou: In other words - I don't really like nesting the main checkbox for STP under generic Options label. It feels off there especially because all options then are related to STP.

garrett: > I don't really like nesting the main checkbox for STP under generic Options label

You're right. Options is too generic here, especially as we only have one. STP is a topology option, so I'm changing the label to something more specific: Topology, which is what STP manages.

There's also the newer OSPF (Open Shortest Path First) and EIGRP (Enhanced Interior Gateway Routing Protocol) in the same family of solutions to provide alternate paths, but we only support STP.

---

> Otherwise this can also use a Switch instead of a checkbox.

Nope, this shouldn't be a switch. It doesn't direct apply, as it needs a save state. Therefore, it's a checkbox, not a switch.

From https://www.patternfly.org/v4/components/switch/design-guidelines/#when-to-use-switches-versus-checkboxes:
> A switch changes an option and saves it simultaneously, while checkboxes require a separate action to save the selection, such as pressing a “Submit” or “Save” button.

garrett:
Changed label to "Topology":

Screenshot 2021-09-08 at 13-08-14 Networking - garrett Bolt

pkg/networkmanager/index.html
```diff
@@ -104,27 +104,30 @@ <h1 id="testing-connection-curtain-restoring" translate="yes">Restoring connecti
         {{! member interfaces will be rendered into here as a list of checkboxes }}
       </div>
 
-      <label class="control-label" for="network-bridge-settings-stp-enabled-input" translate="yes">Spanning tree protocol (STP)</label>
+      <label class="control-label" for="network-bridge-settings-stp-enabled-input" translate="yes">Topology</label>
```
KKoukiou, line 107: I think STP is a protocol - not a Network topology. But please ask someone else on that.

KKoukiou (commented):
Please ask @mvollmer for this 'Topology' label. I think it should be 'Protocol'. Otherwise approved.

pkg/networkmanager/index.html
garrett: Docs I've read about it say it's related to network topology. :shrug: Protocol is also fine.

KKoukiou: Marius was on the opinion that all these terms are too raw and we should rather leave it Options. I also proposed 'Additional options'. Let's pick up one of the two, I trust marius on this.

garrett: Here are some examples:

https://www.tutorialspoint.com/spanning-tree-protocol

> Spanning Tree Protocol (STP) is a communication protocol operating at data link layer the OSI model to prevent bridge loops and the resulting broadcast storms. It creates a loop − free topology for Ethernet networks.

https://en.wikipedia.org/wiki/Spanning_Tree_Protocol

> The Spanning Tree Protocol (STP) is a network protocol that builds a loop-free logical topology for Ethernet networks.

https://techacad.net/introduction-to-spanning-tree-protocol/

> Spanning Tree Protocol was invented to create a loop free topology by blocking interfaces.

---

But we could call it protocol or whatever too. I was trying to use a word that more specifically explains the intent of STP, rather than going with something less specific like "protocol" (which is also redundant, as that's what the P stands for) or vague like "options", which we already decided against at the beginning of this PR.

However, the goal is to move the label to actually be a label for the checkbox. The side label's value is less important, so as long as you're happy with the term, I'm happy.

mvollmer: What about:

```
Spanning tree protocol:  [ ] Enable
```

garrett: We could also consider "Fault tolerance" :shrug:

mvollmer: Which turns into this when enabled:

```
Spanning tree protocol:  [x] Enable
                         Priority: ____
                         Forward delay: ___
                         ...
```

garrett: @mvollmer: No.

Firstly, that defeats the entire goal of this PR.

Secondly, that's like saying "enable/disable enable", which makes no sense.

See this recent twitter thread:
https://twitter.com/hbons/status/1434810758974648324

Screenshot if you don't want to visit twitter:

image

mvollmer: Then leave "Options".

KKoukiou (dismissed):

KKoukiou (approved):

PR #16331, opened 2021-09-09
[rhel-9.0] backport FMF hacks

martinpitt:
We should have https://github.com/cockpit-project/cockpit/pull/16252 on the rhel-9.0 branch to reflect reality, and get a clean fmf `ref:`.

### DIFF
--- a/test/run-verify-host.sh
+++ b/test/run-verify-host.sh
@@ -43,6 +43,9 @@ if grep -q 'ID=.*fedora' /etc/os-release; then
     dnf install -y tcsh
 fi
 
+#HACK: unbreak rhel-9-0's default choice of 999999999 rounds, see https://bugzilla.redhat.com/show_bug.cgi?id=1993919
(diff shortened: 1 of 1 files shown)


marusak (approved):
Thanks!

martinpitt:
Ugh, err, let's not do this -- I'll revert the packit f33 change, it's not important.

martinpitt:
> I'll revert the packit f33 change, it's not important.

To clarify, I'll submit it in a new PR, together with enabling c9s, once that works (testing in PR #16632)

martinpitt:
@marusak: FYI, the failing jumpstart publish is because I added that additional commit for packit f33, and dropped it again, so it was back at the original SHA and thus the dist cache commit already existed. We can just ignore this (we may fix it for main, but too much of a corner case for stable branch to bother)

marusak (approved):
Thanks!

PR #16332, opened 2021-09-09
packit: Enable CentOS Stream 9

martinpitt:


martinpitt:
@thrix, @TomasTomecek : I noticed in https://api.dev.testing-farm.io/v0.1/composes that TF now generally has CentOS 9 stream available, and tried to enable it here. But the [pipeline.log](http://artifacts.dev.testing-farm.io/9a854037-a54f-458b-b906-92266fa7318b/pipeline.log) reports a failure to boot/connect to the VM instance. Is that expected, and the images just don't work yet? Or a bug in the current image, and it's generally supposed to work? Thanks!

thrix:
@martinpitt sorry, there is an issue we need to resolve, the images use different default user and our tooling cannot work with that yet. Let me try to resolve it ASAP.

martinpitt:
Rebased to get a current result.

martinpitt:
Still the [same failure](http://artifacts.dev.testing-farm.io/0d5a05cd-d48c-42f4-b194-3e23435c3f3f/pipeline.log)

martinpitt:
Miro said the TF should support C9 at last -- next round!

martinpitt:
Yay, it works, thanks @thrix ! (Just triggering required fedora-34 test now)

PR #16335, opened 2021-09-10
More mobile test fixes

martinpitt:
At some point there should probably be some factorization for "check something in the system menu", but it's not yet clear to me how this could look like. I suggest we first fix all tests, and then look at all the mobile special cases.

Remaining `@skipMobile`s:
 - `check-metrics TestGrafanaClient`: Let's not bother testing Grafana in mobile mode
 - `check-storage-partitions testSizeSlider`: Mobile mode has a too narrow slider that can't make precise enough settings, so the numbers are off; we may think about how to fix that, but I'd keep it out of this PR.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -677,6 +677,18 @@ class Browser:
 
         self.switch_to_frame(cur_frame)
 
+    def click_system_menu(self, path):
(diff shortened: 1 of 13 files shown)


KKoukiou (approved):
Whoa great progress ! 2 more remaining only !

PR #16342, opened 2021-09-13
common: Restrict frame embedding to same origin

### LINKED ISSUE #16122: Is cockpit vulnerable to clickjacking? [CVE-2021-3660]
Through a security scan, I was notified:

> _The remote web server does not set an X-Frame-Options response header or a Content-Security-Policy 'frame-ancestors' response header in all content responses. This could potentially expose the site to a clickjacking or UI redress attack, in which an attacker can trick a user into clicking an area of the vulnerable page that is different than what the user perceives the page to be. This can result in a user performing fraudulent or malicious transactions._

Suggested solution is: 

> Return the X-Frame-Options or Content-Security-Policy (with the 'frame-ancestors' directive) HTTP header with the page's response.

Is there a way to do this through the config in cockpit?   or is there some other reason it is not subject to clickjacking?

Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1980688 (plus a few dependent ones for the individual packages in Fedora and RHEL 8/9)

martinpitt:
Declare `X-Frame-Options: sameorigin` [1] so that cockpit frames can
only be embedded into pages coming from the same origin. This is similar
to setting CORP in commit 2b38b8d (which applies to `<script>`,
``, etc.).

The main use case for embedding is to run cockpit-ws behind a reverse
proxy, while also serving other pages. Cross-origin embedding is
discouraged these days to prevent "clickjacking".

Cross-origin embedding already did not work in practice, due to our
cookie restrictions: Frames would always just show the login page.
However, this looks confusing and is unclean. With X-Frame-Options, the
browser instead shows an explanatory error page.

Mention the same origin requirement in the embedding documentation.

Fixes #16122
https://bugzilla.redhat.com/show_bug.cgi?id=1980688
CVE-2021-3660

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

-----

 - [x] Get working reproducer from [security bz reporter](https://bugzilla.redhat.com/show_bug.cgi?id=1980688)  that this was not actually exploitable with cross-origins from different IPs/host names
 - [x] Update [documentation](https://cockpit-project.org/guide/latest/embedding.html) that embedding only works for same-origin  frames
 - [x] Set `X-Frame-Options: sameorigin` as a third line of defence
 - [x] builds on top of #16346
 - [x] fix `test-httpstream` to be robust against header hash ordering: PR #16348

## webserver: Restrict frame embedding to same origin

Cockpit's web server now sets the [X-Frame-Options header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) to explicitly disallow frame embedding into a different origin. Due to cookie policy restrictions, this already did not work before in most cases: embedded frames would always just show the login page. But with this new option, the browser forbids this more directly, and shows an explanatory error page instead. (CVE-2021-3660)

image

### DIFF
--- a/test/verify/check-embed
+++ b/test/verify/check-embed
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+import subprocess
+
 import parent
 from testlib import *
 
@@ -71,6 +73,41 @@ Shell=/shell/index.html
         b.wait_visible("iframe[name='embed-auth'][loaded]")
         b.switch_to_frame("embed-auth")
         b.wait_visible("#login-user-input")
+
+    @skipBrowser("Chromium cannot inspect cross-origin frames", "chromium")
+    def testCrossOrigin(self):
+        b = self.browser
+        m = self.machine
+
+        pyhttpd = subprocess.Popen(['python3', '-m', 'http.server', '--bind=localhost',
+                                    '--directory=test/verify/files/embed-cockpit', '12346'])
+
+        def clean_pyhttpd():
+            pyhttpd.terminate()
+            pyhttpd.wait()
+
+        self.addCleanup(clean_pyhttpd)
+
+        # log in normally, to get the auth cookie into the browser and thus maximize possible cross-domain exposure
+        self.login_and_go()
+
+        b.open("http://localhost:12346/index.html")
+        b.set_val("#embed-address", "http://{0}:{1}".format(m.web_address, m.web_port))
+        b.click("#embed-full")
+        # FIXME (#16122): we should not even get that far, frame loading should already be blocked here
+        b.wait_visible("iframe[name='embed-full'][loaded]")
+        b.switch_to_frame("embed-full")
+
+        # second line of defense: existing login cookie does not work (default browser protection)
+        b.wait_visible("#login")
+        b.set_val("#login-user-input", "admin")
+        b.set_val("#login-password-input", "foobar")
+        b.click('#login-button')
+        b.expect_load_frame("embed-full")
+
+        # third line of defense: login succeeds and creates a PAM session; but loading session UI does not
+        # (again, due to default browser protection)
+        b.wait_visible("#login")
 
 
 if __name__ == '__main__':
--- a/test/verify/check-embed
+++ b/test/verify/check-embed
@@ -94,20 +94,13 @@ Shell=/shell/index.html
         b.open("http://localhost:12346/index.html")
         b.set_val("#embed-address", "http://{0}:{1}".format(m.web_address, m.web_port))
         b.click("#embed-full")
-        # FIXME (#16122): we should not even get that far, frame loading should already be blocked here
         b.wait_visible("iframe[name='embed-full'][loaded]")
         b.switch_to_frame("embed-full")
 
-        # second line of defense: existing login cookie does not work (default browser protection)
-        b.wait_visible("#login")
-        b.set_val("#login-user-input", "admin")
-        b.set_val("#login-password-input", "foobar")
-        b.click('#login-button')
-        b.expect_load_frame("embed-full")
-
-        # third line of defense: login succeeds and creates a PAM session; but loading session UI does not
-        # (again, due to default browser protection)
-        b.wait_visible("#login")
+        # X-Frame-Options sameorigin blocks frame
+        if b.cdp.browser == "firefox":
+            b.wait_visible("body.neterror")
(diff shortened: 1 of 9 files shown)


martinpitt:
Darn, when writing the `STATIC_HEADERS_CHUNKED` hack I already had a hunch that assuming glib hash ordering would backfire -- and [it did](https://logs.cockpit-project.org/logs/pull-16342-20210914-072625-784e2518-rhel-8-5/log.html).

I filed PR #16346 to land the test/example fixes, and will rework test-httpstream to be more robust against header changes.

martinpitt:
I tried very hard again today to actually exploit this, but still did not get very far. I now left the description sufficiently vague, and I think this is ready to land despite not having fully confirmed the original scenario. See https://bugzilla.redhat.com/show_bug.cgi?id=1980688#c19 for details.

@mvollmer, want to have a look?

test/verify/check-embed
```diff
@@ -72,6 +74,37 @@ Shell=/shell/index.html
...
 
+    @skipBrowser("Chromium cannot inspect cross-origin frames", "chromium")
+    def testCrossOrigin(self):
+        b = self.browser
+        m = self.machine
+
+        pyhttpd = subprocess.Popen(['python3', '-m', 'http.server', '--bind=localhost',
+                                    '--directory=test/verify/files/embed-cockpit', '12346'])
+        self.addCleanup(pyhttpd.wait)
+        self.addCleanup(pyhttpd.terminate)
```
mvollmer, line 85: Should these be the other way around? terminate() then wait()?

mvollmer: Ahh, cleanup are executed in reverse order!  I propose to make a single cleanup that does everything related to pyhttpd.

test/verify/check-embed
```diff
@@ -72,6 +74,37 @@ Shell=/shell/index.html
...
+        # second line of defense: existing login cookie does not work (default browser protection)
+        b.wait_visible("#login")
+        b.set_val("#login-user-input", "admin")
+        b.set_val("#login-password-input", "foobar")
+        b.click('#login-button')
+        b.expect_load_frame("embed-full")
+
+        # third line of defense: login succeeds and creates a PAM session; but loading session UI does not
+        # (again, due to default browser protection)
+        b.wait_visible("#login")
```
mvollmer: Is there a race here? ('#login' might still be visible here and the real page might load later if the login actually worked.)
Can we wait for #login-user-input to be empty again, for example?

mvollmer (approved):
One question and one potential code improvement...

test/verify/check-embed
martinpitt: It's a race condition either way, I'm afraid -- if we wait for `#login` to *not* be present, then the failure and re-draw of login might happen faster than the browser and our JS/Python polling machinery to pick it up. My hope here was that waiting for the frame to load in line 102) would already be correct, and that after that there can't be any "old" contents.

Note that the next commit removes all that code anyway.

test/verify/check-embed
martinpitt: Correct -- I can do a single one with a `lambda:` instead, which does both things.

test/verify/check-embed
mvollmer: Yeah, waiting for something to disappear and then reappear doesn't work.  Does expect_load_frame do anything when the frame is already loaded?

One idea would be to add something to the DOM artificially in the test before clicking #login, and then wait for it to disappear (which only happens when the page is reloaded.)

> Note that the next commit removes all that code anyway.

Heh, true.

martinpitt: > Does expect_load_frame do anything when the frame is already loaded?

yes, it literally waits for a "frame loaded" event. If there is none, it will hang and time out, and fail the test. It's edge triggered (unlike our `wait_present()` and friends, which is level-triggered)

martinpitt: I ran this test just once in CI (successfully), and lots of times locally. This was mostly to play around with the original behaviour, what the CVE complained about.

test/verify/check-embed
martinpitt: Done, I added a helper function as `lambda: terminate() or wait()` would look a little too hackish.

test/verify/check-embed
martinpitt: I'm not changing this -- I am reasonably convinced that the explicit `expect_load_frame()` DTRT here.

mvollmer: Agreed.

mvollmer (approved):

garrett:
Updated title and comment for release notes

PR #16346, opened 2021-09-14
various fixes/improvements to testing embedding and reverse proxying

martinpitt:
Split out from #16342, as the full thing still needs reporter feedback and some rework of the unit test.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -917,23 +917,23 @@ ProtocolHeader = X-Forwarded-Proto
         m.execute("setsebool -P httpd_can_network_connect on")
         self.allow_journal_messages("audit.*bool=httpd_can_network_connect.*val=1.*")
 
-    def checkCockpitOnProxy(self):
+    def checkCockpitOnProxy(self, urlroot=""):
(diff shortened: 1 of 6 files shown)


test/verify/check-connection
```diff
@@ -986,6 +987,47 @@ server {
...
+             f"https://alice:{https_port}/embed-cockpit/embed.css"],
+            stderr=subprocess.STDOUT)
+        self.assertIn(b"HTTP/1.1 200 OK", out)
+        self.assertIn(b"#embed-links", out)
+
+        # embedding
+        b = self.browser
+        b.ignore_ssl_certificate_errors(True)
+        b.open(f"https://{https_host}:{https_port}/embed-cockpit/index.html")
+        b.set_val("#embed-address", f"https://{m.web_address}:{https_port}/cockpit-root")
```
martinpitt: This should be `https_host` instead of `m.web_address`. It's the same value, but looks confusing.

martinpitt: Fixed.

marusak (approved):
Thanks!

PR #16347, opened 2021-09-14
bridge: Filter more DBus signals by sender when cache watching

mvollmer:
Manager signals also need to be filtered, just as we did for property
signals in 11bea587f9a9385d262bb896155afe7aa30747be.


mvollmer:
Example: https://logs.cockpit-project.org/logs/pull-14109-20210913-133846-715e389c-centos-8-stream/log.html#257

martinpitt (approved):
Danke!

PR #16348, opened 2021-09-14
bridge: Robust JSON object comparison in test-httpstream.c

martinpitt:
Comparing the received control message JSON against the expected value
in string form has always been brittle: It assumes a
predictable/reproducible GLib dictionary serialization. This will break
across different GLib versions once we add a new static header.

So change the tests to compare the control message with
`cockpit_assert_json_eq()` (which will do parsing and deep object
comparison) instead of plain string comparison. Introduce a new
assert_channel_response() helper for this.

### DIFF
--- a/src/bridge/test-httpstream.c
+++ b/src/bridge/test-httpstream.c
@@ -34,7 +34,7 @@
 extern gboolean cockpit_webserver_want_certificate;
 
 /* JSON dict snippet for headers that are present in every request */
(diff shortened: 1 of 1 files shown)


martinpitt:
This blocks PR #16342, where naïvely expanding `STATIC_HEADERS` [breaks across OSes](https://logs.cockpit-project.org/logs/pull-16342-20210914-072625-784e2518-rhel-8-5/log.html)

KKoukiou (approved):

Issue #15085, opened 2021-01-02
Selecting a service in the logs ui gets slower as the amount of logs grow


xhalo32:
Cockpit version: 215
OS: `Linux beta 5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

Going to the logs ui and selecting the service causes high CPU usage and a significantly long loading period when a large amount of system logs (1.2GB in my case) are present. The selected time span does not make a difference to the time it takes to fetch all the logs. The command which is using all of the cpu when clicking the service dropdown menu is:
```shell
journalctl -q --no-tail --output=verbose --reverse -- PRIORITY=0 PRIORITY=1 PRIORITY=2 PRIORITY=3 PRIORITY=4 PRIORITY=5 PRIORITY=6 PRIORITY=7
```

Steps to reproduce

1. Set up a cockpit server and wait for the server's logs to grow to a few GBs.
2. Go to the logs part of the cockpit ui (`/system/logs`)
3. Click on the service drop down menu and observe that it takes a while to finish loading

Screeshot of the dropdown menu when it's loading:
![image](https://user-images.githubusercontent.com/15152190/103467826-c35dd000-4d5b-11eb-8429-c0eca97441b3.png)

martinpitt:
I'm afraid that we have no control over how long `journalctl` itself takes to find logs. The journal has "traditionally" been quite slow, as it just takes time to sift through gigabytes of logs, particularly on rotating disks. However, it really should get better if you select "current boot" or "last  24 hours" in the top left "time" selector. Can you confirm that?

@marusak, @mvollmer: IMHO our default to show so much available logs is a bit misfortunate. It also takes a while here, makes the page slow and bulky. Today I get logs up to December 17, so it roughly looks like we are showing 14 days by default? My on-disk journal goes back to Oct 27, so we at least don't show *all* logs by default. But maybe defaulting to 24 hours is enough, and we have the "Load earlier entries"/time selector as well?

martinpitt:
@xhalo32 : Note that also cockpit 215 is quite old, and the Services page was much improved since then.

marusak:
Ok, let me explain.
We show only 5000 logs by default. Then you can click `Load earlier entries` to load 1000 more. If your log files are > 1GB, you most likely have there hundred thousands or millions of entries.
[...]

PR #16349, opened 2021-09-14
ssystemd: show last three days of logs by default -- FIX COMMIT MESSAGE ON MERGE: systemd: show 24 hours of logs by default

jelly:
To circumvent loading a lot of logs by default which can be fairly slow
set the default preset to showing the last three days of logs.

Fixes: #15085

### DIFF
--- a/pkg/systemd/logs.jsx
+++ b/pkg/systemd/logs.jsx
@@ -56,7 +56,7 @@ const _ = cockpit.gettext;
 const timeFilterOptions = [
     { key: "boot", value: 0, toString: () => _("Current boot"), },
     { key: "boot", value: "-1", toString: () => _("Previous boot") },
-    { key: "since", value: "-24hours", toString: () => _("Last 24 hours") },
+    { key: "since", value: "-24hours", toString: () => _("Last 24 hours"), default: true },
     { key: "since", value: "-7days", toString: () => _("Last 7 days") },
 ];
 
@@ -83,8 +83,7 @@ const getTimeFilterOption = options => {
         return timeFilterOptions.find(option => option.key == 'boot' && option.value == options.boot);
     else if (options.since)
         return timeFilterOptions.find(option => option.key == 'since' && option.value == options.since);
-    else
-        return undefined;
+    return timeFilterOptions.find(option => 'default' in option); // Use the default key
 };
 
 export const LogsPage = () => {
@@ -273,6 +272,7 @@ export const LogsPage = () => {
                          variant={PageSectionVariants.light}
                          id="journal-box">
                 <JournalBox dataFollowing={dataFollowing}
+                            defaultSince={timeFilter ? timeFilter.value : getTimeFilterOption({}).value}
                             setCurrentIdentifiers={setCurrentIdentifiers}
                             setFilteredQuery={setFilteredQuery}
                             updateIdentifiersList={updateIdentifiersList}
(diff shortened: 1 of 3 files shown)


pkg/systemd/logs.jsx
```diff
@@ -273,6 +274,7 @@ export const LogsPage = () => {
                          variant={PageSectionVariants.light}
                          id="journal-box">
                 <JournalBox dataFollowing={dataFollowing}
+                            defaultSince={timeFilter.value}
```
martinpitt, line 277: This must be where the [tests crash](https://logs.cockpit-project.org/logs/pull-16349-20210914-141315-c1a05c4f-fedora-35/log.html#240-2):

    RuntimeError: TypeError: Cannot read property 'value' of undefined

pkg/systemd/logs.jsx
```diff
@@ -57,6 +57,7 @@ const timeFilterOptions = [
     { key: "boot", value: 0, toString: () => _("Current boot"), },
     { key: "boot", value: "-1", toString: () => _("Previous boot") },
     { key: "since", value: "-24hours", toString: () => _("Last 24 hours") },
+    { key: "since", value: "-3days", toString: () => _("Last 3 days"), default: true },
```
martinpitt, line 60: Was there some particular reasoning about 3 days? ISTM that keeping it simple and showing the last day is a bit more intuitive -- Usually, when you look at logs, you want to investigate something recent. But bikeshedding, of course

martinpitt (changes requested):

pkg/systemd/logs.jsx
jelly: I was under the impression it was decided here https://github.com/cockpit-project/cockpit/issues/15085#issuecomment-891630590 I think the reasoning for selecting 3 days is that one day doesn't give enough information.

pkg/systemd/logs.jsx
jelly: Ah, that's why I initially thought about passing `timeFilter`, will fix.

jelly: It's due to the test giving since=@123123213 which doesn't match anything and returns undefined and I was somehow thinking the code would always fallback to my default else.

pkg/systemd/logs.jsx
jelly: After some more thought I've changed it to the last 24 hours.

martinpitt (approved):
WFM, thanks! @marusak, are you d'accord with this?

The commit message is obviously wrong now, but we can fix that on squash&merge -- no need to re-run CI.

marusak:
Cool, while I have nothing against last 24 hours, I miss a way, how to get more logs.
There is no button `Load previous logs` on the bottom (which is correct, since there were not more than 5000 logs in last 24 hours). There is no option in the first dropdown to switch to `All` or to `Current` but mainly the filter has option `Since` empty - it should be `-24hours`.

I think we just need to pre-fill the since option as well - that way users know, what is limiting them and then can for example delete it or extend it.

marusak:
Another issue is that this should only apply for the Logs page. If you go for example to services page, pick up service that has some logs, but that did not happen in the last 24 hours and click on `View all logs` - the select states `Last 24 hours` while it shows logs from couple of days ago.

martinpitt:
FTR, with the time filter set to "last 24 hours", the "Load earlier entries" button *does* load earlier days for me. I clicked on it twice, and now it goes back all the way to Aug 25 for me.

PR #16352, opened 2021-09-15
storage: Wait for the watch during init, not the unwatched proxies

mvollmer:
Calling wait() on a proxy with { watch: false } is pointless, it will
resolve immediately.  Instead, waiting on the explicit watch() call
will correctly delay us until all proxies in scope of that watch call
have been initialized.

The manager object has been waited on and checked independently
already before, so we don't have to worry about it being valid in
init_model.

### DIFF
--- a/pkg/storaged/client.js
+++ b/pkg/storaged/client.js
@@ -154,8 +154,6 @@ client.call = function call(path, iface, method, args, options) {
 };
 
 function init_proxies () {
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Oooh, this is a nice simplification and robustification!

mvollmer:
> Oooh, this is a nice simplification and robustification!

Yeah, and after all these years... you can always learn something new. :-)

PR #16354, opened 2021-09-15
test: check-kdump cleanup

martinpitt:
- Clicking "Apply" again with a failed NFS validation is bogus -- it's
   not going to change the validation error and not going to close the
   dialog. Afterwards, the test clicked on the settings link again,
   which is not possible in practice. Instead, close the dialog and
   ensure that it goes away.

 - Address the settings link via ID and explicitly check its expected
   text.  Selecting it via :contains() is weird.

 - Use button classes instead of strings for the dialog, and scope the
   buttons for the dialog.

 - Use f-strings instead of .format()

 - Drop unused rreplace() method

 - Fold rebootMachine() into enableKdump()

-----

This was some preparation for PR #16356

### DIFF
--- a/test/verify/check-kdump
+++ b/test/verify/check-kdump
@@ -27,27 +27,20 @@ from testlib import *
 @skipDistroPackage()
 class TestKdump(MachineCase):
 
-    def rreplace(self, s, old, new, count):
-        li = s.rsplit(old, count)
-        return new.join(li)
-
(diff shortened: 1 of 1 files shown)


martinpitt:
Rebased to fix packit, our own tests were all green.

marusak (approved):
Thanks!

Issue #16353, opened 2021-09-15
Reduce excess padding around cards in space-constrained layouts to optimize for content


garrett:
Goal: remove excess padding that only gets in the way of actual content on space constrained layouts, for all pages.

---

The Overview page has a one-off that reduces card padding in mobile mode. I am pretty sure we handled this across pages at one point. Perhaps pages have been rewritten since and this feature fell away? (Overview was the first page using PF4, IIRC, so that's probably why it's local, and why I seem to remember other (pre-PF4) pages having this feature too.)

Here's what I'm talking about... it only takes effect in super-small sizes, to shave off a little excess padding... compare:

![Screen Shot 2021-09-15 at 10 48 12](https://user-images.githubusercontent.com/10246/133402023-08e72257-d8b8-443f-b0cc-65564a9fde2e.png)
![Screen Shot 2021-09-15 at 10 49 19](https://user-images.githubusercontent.com/10246/133402109-661be782-6e4a-40eb-8739-bd153c9d58c7.png)

- This could be done in a more global, generic manner.
- It could drop the extra padding for all mobile, not just extreme sizes. (To be determined, but I think it might help.)
- We could consider it either as an opt-in for Cockpit or as a PF4 override. :shrug: 

Dropping the padding and stretching cards to the width for overview is @ https://github.com/cockpit-project/cockpit/blob/main/pkg/systemd/overview.scss#L196-L218

KKoukiou:
This should be done the PF4 way. https://github.com/cockpit-project/cockpit/pull/16355

garrett:
@KKoukiou and I hashed it out in #16355 and I suggested some opt-in CSS to go into `page.scss` in https://github.com/cockpit-project/cockpit/pull/16355/files#r710164707, and then we just add the class to the various pages where we want to drop the edges.

I can do this pretty easily, so I'll assign it to myself (and eventually get to it).

PR #16355, opened 2021-09-15
overview: remove page padding on mobile the PF4 way

KKoukiou:


pkg/systemd/overview.jsx
```diff
@@ -271,7 +271,7 @@ class OverviewPage extends React.Component {
                             { headerActions }
                         </div>
                     </PageSection>
-                    <PageSection variant={PageSectionVariants.default}>
+                    <PageSection variant={PageSectionVariants.default} padding={{ default: 'noPadding', sm: 'padding' }}>
```
martinpitt, line 274: This looks the wrong way around, but does work perfectly. Nice!

martinpitt (approved):
Epharisto!

martinpitt:
FTR: centos-8-stream had a flake, retried that. Want to wait for that. The branch is out of date, ignoring packit.

pkg/systemd/overview.jsx
garrett: I guess it's "mobile first" saying no padding by default, but add padding for size `sm` and up?

garrett (changes requested):
This merges the first card (or alert, depending) with the top info area, as there's no gap. 

The original code has a padding of 0.5rem (= `--pf-global--spacer--sm`, but we were super-new to PF4 when this page was ported, so those spacing variables weren't used here... but it's equivalent).

To fix, please re-add the `sm` spacer to the top and bottom, but not the sides. (The top is more important; the bottom indicates that it's a card and not a footer, so I'd like it too.)

garrett:
Related: Ideally, we'd eventually want to make this concept work for all pages. (In pre-PF4 Cockpit, we had something like this across many/most pages.) I don't know if there's a way to enable it across all of Cockpit or if we just have to change it a page at a time.

Reducing the padding on the sides across all pages is #16353, which is related, but a different scope than this PR.

pkg/lib/page.scss
```diff
@@ -189,3 +189,8 @@ html:not(.index-page) body {
 .ct-icon-times-circle {
     color: var(--pf-global--danger-color--100);
 }
+
+.ct-page-add-top-bottom-padding.pf-c-page__main-section.pf-m-no-padding.pf-m-padding-on-sm {
+    padding-top: var(--pf-c-page__main-section--PaddingTop);
+    padding-bottom: var(--pf-c-page__main-section--PaddingBottom);
+}
```
garrett, line 196: At this point, should we consider an opt-in ct-* class with a media query that just sets the padding?

```css
@media screen and (max-width: $pf-global--breakpoint--sm) {
  .ct-page-mobile-padding {
    padding: padding-top: var(--pf-c-page__main-section--PaddingTop) 0 var(--pf-c-page__main-section--PaddingBottom);
  }
}
```
(Which hardcodes the top and bottom, like what's in this PR at the moment.)

or

```css
@media screen and (max-width: $pf-global--breakpoint--sm) {
  .ct-page-mobile-padding {
    padding-left: 0;
    padding-right: 0;
  }
}
```
(Which just nullifies the left and right margins, leaving the top and bottom untouched.)

It would be less CSS and less React one className to add (`.ct-page-mobile-padding`).

Meanwhile, we could also ask PF to add a paddingWidth option (or something like that).

KKoukiou:
Let's do it globally for all pages with an PF4 override then. Closing this - as I am not super interested to work on whole cockpit rewrite on this at this point.

PR #16357, opened 2021-09-16
storage: Read /etc/os-release from current host

mvollmer:
The data in localStorage is always for the login host.


martinpitt (approved):
Whoops, nice catch! Feels a little weird having to implement yet another parser for os-release, but indeed I didn't see a quick better way either. Thanks!

PR #16361, opened 2021-09-16
tools: Move ownership of /usr/share/cockpit to cockpit-bridge.rpm

martinpitt:
Owning it by the meta-package is wrong: Page packages like
cockpit-{storage,podman,machines} should only depend on c-bridge, or at
most -system (which in turn depends on -bridge), not on the full
metapackage. Avoid an unowned directory in such situations.

Spotted by Ben Beasley in https://bugzilla.redhat.com/show_bug.cgi?id=196945

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -220,7 +220,9 @@ install -D -p -m 644 AUTHORS COPYING README.md %{buildroot}%{_docdir}/cockpit/
 %endif
 
 # Build the package lists for resource packages
-echo '%dir %{_datadir}/cockpit/base1' > base.list
+# cockpit-bridge is the basic dependency for all cockpit-* packages, so centrally own the page directory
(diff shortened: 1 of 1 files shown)


martinpitt:
```
# rpm -qf /usr/share/cockpit
cockpit-bridge-253.9.g693188eee-1.wip.fc34.x86_64
```
I also tested an upgrade:
```
# rpm -q cockpit cockpit-bridge
cockpit-250-1.fc34.x86_64
cockpit-bridge-250-1.fc34.x86_64

# rpm -U --verbose /tmp/cockpit-*.rpm
# rpm -qf /usr/share/cockpit
cockpit-bridge-253.9.g693188eee-1.wip.fc34.x86_64
```
So the ownership transfer went alright.

marusak (approved):
Makes sense, thanks!

PR #16368, opened 2021-09-20
storage: Replace all "var" keywords with "let" or "const"

mvollmer:
So that we don't mix old and new styles.

### DIFF
--- a/pkg/storaged/block-details.jsx
+++ b/pkg/storaged/block-details.jsx
@@ -37,9 +37,9 @@ const _ = cockpit.gettext;
 
 export class BlockDetails extends React.Component {
     render() {
-        var block = this.props.block;
+        const block = this.props.block;
 
-        var header = (
+        const header = (
             <Card>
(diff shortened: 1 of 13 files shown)


mvollmer:
This was done with sed and eslint:fix.

martinpitt:
@mvollmer : No point in `packit build` -- your branch is out of date, so with packit's current "rebase to latest origin/main" behaviour there just won't be a matching cached dist build. @allisonkarlitskaya and I need to talk about what to do there (I think she had a plan already), but until then it's "rebase or ignore" (ignore seems fine here)

martinpitt:
This PR includes an update to test/reference, why? Is that just a git rebase/pilot error?

mvollmer:
> This PR includes an update to test/reference, why? Is that just a git rebase/pilot error?

Oh, yes!

martinpitt (approved):
Clever approach! Feels like we should do this throughout cockpit, and then add some eslint rule that yells on using `var`.

martinpitt:
BTW, `"no-var": "warn",` in .eslintrc.json does just that, and works fine.

PR #16369, opened 2021-09-20
lib: Eliminate `var` keyword

martinpitt:
This has broken scoping and is hard to audit. For consistency, use
`const` where possible, and `let` where necessary.

-----

Tested with temporarily adding `"no-var": "warn",` to .eslintrc.json. Done with replacing all `var` with `const`, and working through the compiler errors.


martinpitt:
After this gets in, I'm happy to port our pages, and permanently enable the eslint rule.

mvollmer:
>  Done with replacing all `var` with `const`, and working through the compiler errors.

Just fyi, what I did was `sed -i 's/var /let /' *; npm run eslint:fix`.  That catches almost everything, except things like `let x; a(); x = 12; b()` where `x` should be const but it requires moving the declaration like this `a(); const x = 12; b()`

martinpitt:
@mvollmer : Right, I went the other way around, as often with some very simple rearrangement and fixing scope, `let` can be moved to `const`.

pkg/packagekit/updates.jsx
```diff
@@ -63,54 +63,48 @@ import callTracerScript from 'raw-loader!./callTracer.py';
...
-let PK_STATUS_STRINGS = {};
-let PK_STATUS_LOG_STRINGS = {};
+const STATE_HEADINGS = {};
+const PK_STATUS_STRINGS = {};
+const PK_STATUS_LOG_STRINGS = {};
 const packageSummaries = {};
 
 function init() {
-    STATE_HEADINGS = {
-        loading: _("Loading available updates, please wait..."),
```
mvollmer, line 73: This can actually be done in the global scope, outside of any function.  I.e., just have `const STATE_HEADINGS = { loading: _("...") };` above.

mvollmer (changes requested):

pkg/packagekit/updates.jsx
martinpitt: No, it can't -- `_()` aka gettext() is a dynamic function imported from cockpit, and the page needs to set the locale first. I validated that again with

```diff
--- pkg/packagekit/updates.jsx
+++ pkg/packagekit/updates.jsx
@@ -63,13 +63,14 @@ import callTracerScript from 'raw-loader!./callTracer.py';
 const _ = cockpit.gettext;
 
 // "available" heading is built dynamically
-const STATE_HEADINGS = {};
+const STATE_HEADINGS = {
+    loading: _("Loading available updates, please wait..."),
+};
 const PK_STATUS_STRINGS = {};
 const PK_STATUS_LOG_STRINGS = {};
 const packageSummaries = {};
 
 function init() {
-    STATE_HEADINGS.loading = _("Loading available updates, please wait...");
     STATE_HEADINGS.locked = _("Some other program is currently using the package manager, please wait...");
     STATE_HEADINGS.refreshing = _("Refreshing package information");
     STATE_HEADINGS.uptodate = _("System is up to date");
```

and on a German UI this now does not translate the "Loading available updates..." any more, as at the "static" scope the locale is still "C".

mvollmer: Ah, damn.

mvollmer (approved):

pkg/packagekit/updates.jsx
marusak: Yes, this hit us many times before. If cockpit is not yet loaded and initialized, then `cockpit.language` is not yet set and therefore any of these strings would not be translated.

marusak (approved):
Thanks!

pkg/packagekit/updates.jsx
martinpitt: As discussed on IRC, I fixed the HTML to make cockpit.gettext() work in the global scope, and eliminated the init() function. Looks much nicer now.

martinpitt: @marusak has some doubts that this might be a race condition. I reverted to the previous (approved) version, and made a "someday" todo item to revisit this.

marusak (approved):

mvollmer (approved):

PR #16340, opened 2021-09-13
package.json: Update xterm

cockpituous:

PR #16370, opened 2021-09-20
terminal: Adjust spacing (above & left)

garrett:
Follow-up to https://github.com/cockpit-project/cockpit/pull/16340, to fix the overlap with the scrollbar.


KKoukiou:
A test is failing https://logs.cockpit-project.org/logs/pull-16370-20210920-163509-260a90d3-fedora-34-devel/log.html#275-2 and also pixel tests need update.

garrett:
I haven't touched anything but CSS, so I don't know how a test would fail. 

As far as pixel testing, I changed the way the vertical alignment is handled, so it splits the difference between the top and bottom, to align the content in the middle, instead of not having any space above and all the space below. So, yes, this would change a screenshot.

KKoukiou:
> I haven't touched anything but CSS, so I don't know how a test would fail.
> 
> As far as pixel testing, I changed the way the vertical alignment is handled, so it splits the difference between the top and bottom, to align the content in the middle, instead of not having any space above and all the space below. So, yes, this would change a screenshot.

The terminal test check for specific lines in the terminal their contents. Since you changes the terminal width, some characters changed position / lines. I can take a look in the tests. For the screenshots you can follow the process https://cockpit-project.org/blog/pixel-testing.html , it's straightforward.

garrett:
> For the screenshots you can follow the process https://cockpit-project.org/blog/pixel-testing.html , it's straightforward

It's not though.

Take some file from somewhere with a failed test that's somewhere and put it in a place? Where are all of these things?

```sh
$ ./test/common/pixel-tests init
'test/reference' already exists in the index
$ ./test/common/pixel-tests pull
$ ./test/common/pixel-tests status
No changes
```

Now what? The instructions don't say anything else to do. I'm assuming the tests didn't run. It doesn't say how to run them, or where the results are.

KKoukiou (dismissed):

marusak:
It does not copy the correct text on mobile. Looking a bit more, possibly due to font-size on mobile.

marusak:
Oh also we count with padding when we recalculate size. That probably also needs adjusting.

marusak:
> /* Add a touch of space to the left, but not enough to overlap the scrollbar */
>    padding-left: var(--pf-global--spacer--xs);

@garrett this does almost nothing with regards of the scrollbar overlapping. For that we need to adjust js which counts how many characters we can fit on one line. So I think it is safe to assume that we want `--pf-global--spacer--xs` on right and left as well?

marusak:
https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/cockpit-components-terminal.jsx#L292

So we take the width we have available, subtract 22 and that is width we use. When you limited the padding it was just "luck" that `--pf-global--spacer--xs + <width of scrollbar> < 22px`. Actually it is around `15px` so we are leaving sometimes unnecessary big gap of the left.

pkg/lib/console.css
```diff
@@ -57,6 +58,6 @@
 }
 
 /* Ensure the console fits to its container (and doesn't attempt to go beyond the limits) */
-.xterm-screen, .xterm-viewport {
```
marusak, line 60: @garrett why you dropped `.xterm-screen` from here?

marusak:
This is an issue if the scrollbar width is more than we expect - on Chrome it is wider.

pkg/lib/console.css
garrett: It didn't seem needed?

garrett:
> So I think it is safe to assume that we want `--pf-global--spacer--xs` on right and left as well?

No, the right and bottom of the terminal are based on character widths. For the vertical alignment, I tell it to split the extra space. We can't do that horizontally too, as there's a scrollbar.

> This is an issue if the scrollbar width is more than we expect - on Chrome it is wider.

Which is exactly what this PR addresses (horizontally).

marusak:
For some reason, I did not get bots saying that refs has changed. But the pictures got updated, you can see how it looks like now.

KKoukiou:
> For some reason, I did not get bots saying that refs has changed. But the pictures got updated, you can see how it looks like now.

We don't have pixel tests workflow in cockpit. That's only merged in c-machines repo so far.

PR #16371, opened 2021-09-21
realmd: React on privileges change

marusak:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2006028

### DIFF
--- a/pkg/systemd/overview-cards/realmd-operation.js
+++ b/pkg/systemd/overview-cards/realmd-operation.js
@@ -1,5 +1,6 @@
 import jQuery from "jquery";
 import cockpit from "cockpit";
+import { superuser } from "superuser";
 import * as packagekit from "packagekit.js";
 import { install_dialog } from "cockpit-components-install-dialog.jsx";
 import "patterns";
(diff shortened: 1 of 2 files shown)


test/verify/check-system-realms
```diff
@@ -83,7 +83,12 @@ class CommonTests:
         # Tell realmd to enable domain-qualified logins; unqualified ones are covered in testUnqualifiedUsers
         m.write("/etc/realmd.conf", "[cockpit.lan]\nfully-qualified-names = yes\n", append=True)
 
-        self.login_and_go("/system")
+        # Test that we reconnect on privileges change
```
marusak, line 86: This is a bit awkward test case - I used this as reproducer. @martinpitt WDYT?

martinpitt: Seems fine to me -- certainly better than joining yet another time, as that's quite expensive.

martinpitt (approved):
Nice, thank you!

marusak:
meh - `TypeError: Cannot read properties of null (reading 'close')`

pkg/systemd/overview-cards/realmd-operation.js
```diff
@@ -602,7 +603,8 @@ export function setup() {
                 self.dispatchEvent("changed");
             }
             $(realmd).off();
-            realmd.close();
+            if (realmd)
+                realmd.close();
```
martinpitt, line 607: At this point in the code the `realmd` variable should never be `null`. I figure what happens is that calling setup_realms_proxy() a second time does not clean up the initial `"close"` event handler, so that will still run, set `realmd = null`, and crash the second handler.

So I think a better fix would be to call `$(realmd).off()` at the start of the function if `realmd` is already defined, to clean up the old event handler.

martinpitt (changes requested):

martinpitt (approved):
Dakujem!

martinpitt:
@marusak : Rebuilding packit won't help, the branch is out of date.

PR #16376, opened 2021-09-21
lib/patternfly: Fix "form select" padding

garrett:
PatternFly's form select does not use similar spacing compared to PF's select.

This causes issues when the form select is not stretched. Seen @:
https://github.com/cockpit-project/cockpit-podman/issues/755

Upstream issue:
https://github.com/patternfly/patternfly/issues/4387

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -28,6 +28,21 @@ ul.pf-c-select__menu {
     overflow-y: auto;
 }
 
+/* Adjust padding on form selects to resemble PF non-form selects */
+/* (This can be seen when the longest text is selected on a non-stretched select) */
+/* Upstream: https://github.com/patternfly/patternfly/issues/4387 */
+/* Cockpit-Podman: https://github.com/cockpit-project/cockpit-podman/issues/755 */
+select.pf-c-form-control {
+    --pf-c-form-control--PaddingRight: 41px;
+    --pf-c-form-control--PaddingLeft: 8px;
+
+    // Firefox's select text has additional padding (4px)
+    @-moz-document url-prefix() {
+        --pf-c-form-control--PaddingRight: 37px;
+        --pf-c-form-control--PaddingLeft: 4px;
+    }
+}
+
 /* All SVGs used in PF4 have some inline style to align them
  * https://github.com/patternfly/patternfly-react/issues/4767
  */


pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -28,6 +28,21 @@ ul.pf-c-select__menu {
     overflow-y: auto;
 }
 
+/* Adjust padding on form selects to resemble PF non-form selects */
+/* (This can be seen when the longest text is selected on a non-stretched select) */
+/* Upstream: https://github.com/patternfly/patternfly/issues/4387 */
+/* Cockpit-Podman: https://github.com/cockpit-project/cockpit-podman/issues/755 */
+select.pf-c-form-control {
+    --pf-c-form-control--PaddingRight: 41px;
+    --pf-c-form-control--PaddingLeft: 6px;
```
garrett, line 37: We could possibly reverse out the padding to the normal selects, if they apply the values on the `pf-c-form-control` and use that minus some values (possibly spacing related, but as this doesn't divide cleanly, we would probably still have to add magic `px` values, especially to the Firefox compensation below).

In other words, I think we're stuck with magic values in `px` one way or the other, to make form selects look more like custom selects.

KKoukiou (approved):
Since there is an upstream PF4 issue we can leave even with the magic numbers wa. thanks :)

garrett:
The visual tests that failed pointed me to a location that has form selects in the same place as inputs, so it's mixing and matching elements. I've used that as reference to correct the left padding (which was off by 2px in all browsers in this PR prior, as I only tested based on isolated examples from the PF demo instead of in Cockpit itself with other elements nearby).

I force pushed and I guess I'll have to wait for the visual tests to see what happens and then correct them when they get built.

garrett:
For the non-Firefox values, the left padding looks like it's 0.5 of spacer, the right seems like 2.5 of spacer might be about right (I'd have to re-check with both in-page). For Firefox, we could use the same values in a calc() and remove 4px to make it pixel perfect with the rest.

KKoukiou (changes requested):
Failing pixel tests.

garrett:
> Failing pixel tests.

Not surprising, as changing this PR will change the visuals, which would cause the so-called "pixel" tests to fail. But we need them to fail first to get the new PNGs to download, then copy, then push.

garrett:
It's only failing mobile, I guess? Fedora 34 doesn't even show the "pixel" tests.

garrett:
I updated the images _again_, and just got:

```
$ ./test/common/pixel-tests push
No changes
Everything up-to-date
```

:shrug:

garrett:
@KKoukiou: It looks like the pixel tests passed this time? I see pakit failures, but everything else seems fine?

I guess I could try to change the values to PF variables and it should still pass afterward if I do it successfully (as they should resolve to the same pixel values).

KKoukiou (approved):

PR #16379, opened 2021-09-22
Adjust the sides of pages in mobile sizes

### LINKED ISSUE #16353: Reduce excess padding around cards in space-constrained layouts to optimize for content

garrett:
Fixes #16353

### DIFF
--- a/pkg/networkmanager/firewall.jsx
+++ b/pkg/networkmanager/firewall.jsx
@@ -943,7 +943,7 @@ export class Firewall extends React.Component {
                           <BreadcrumbItem isActive>{_("Firewall")}</BreadcrumbItem>
                       </Breadcrumb>}
                   additionalGroupedContent={
-                      <PageSection id="firewall-heading" className="firewall-heading" variant={PageSectionVariants.light}>
+                      <PageSection id="firewall-heading" variant={PageSectionVariants.light} className="firewall-heading">
                           <Flex alignItems={{ default: 'alignItemsCenter' }} justifyContent={{ default: 'justifyContentSpaceBetween' }}>
                               <Flex alignItems={{ default: 'alignItemsCenter' }} id="firewall-heading-title-group">
                                   <Title headingLevel="h2" size="3xl">
@@ -954,7 +954,7 @@ export class Firewall extends React.Component {
                               { enabled && !firewall.readonly && <span className="btn-group">{addZoneAction}</span> }
                           </Flex>
                       </PageSection>}>
-                <PageSection id="zones-listing">
+                <PageSection id="zones-listing" className="ct-pagesection-mobile">
                     { enabled && <Stack hasGutter>
                         {
                             zones.map(z => <ZoneSection key={z.id}
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -138,10 +138,10 @@ class OverviewPage extends React.Component {
                 {this.state.rebootModal && <ShutdownModal onClose={() => this.setState({ rebootModal: false })} />}
                 {this.state.shutdownModal && <ShutdownModal shutdown onClose={() => this.setState({ shutdownModal: false })} />}
                 <Page>
-                    <PageSection variant={PageSectionVariants.light} padding={{ default: 'noPadding' }}>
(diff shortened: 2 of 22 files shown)


garrett:
I haven't checked this yet; I needed to get this branch out of the way to switch to other things, so I figured I'd share what I have as a draft PR.

It will break visual testing for mobile sizes.

garrett:
Rebased. (A few commits could still be squashed together though.)

I think it's ready for review.

It'd be nice to land this soon. It will definitely need the pixel tests updated, but they have to run again first.

martinpitt:
[This run](https://logs.cockpit-project.org/logs/pull-16379-20220214-165550-10c76b55-fedora-35/log.html) has the pixel updates. These look really nice, thank you! Do you want to update them yourselves, or want me to push these?

pkg/networkmanager/firewall.jsx
```diff
@@ -943,7 +943,7 @@ export class Firewall extends React.Component {
                           <BreadcrumbItem isActive>{_("Firewall")}</BreadcrumbItem>
                       </Breadcrumb>}
                   additionalGroupedContent={
-                      <PageSection id="firewall-heading" className="firewall-heading" variant={PageSectionVariants.light}>
+                      <PageSection id="firewall-heading" variant={PageSectionVariants.light} className="firewall-heading">
```
marusak, line 946: Shouldn't this one also have `ct-pagesection-mobile`? If not, then we can revert this change?

pkg/systemd/overview.jsx
```diff
@@ -138,10 +138,10 @@ class OverviewPage extends React.Component {
                 {this.state.rebootModal && <ShutdownModal onClose={() => this.setState({ rebootModal: false })} />}
                 {this.state.shutdownModal && <ShutdownModal shutdown onClose={() => this.setState({ shutdownModal: false })} />}
                 <Page>
-                    <PageSection variant={PageSectionVariants.light} padding={{ default: 'noPadding' }}>
+                    <PageSection variant={PageSectionVariants.light} padding={{ default: 'noPadding' }} className="ct-pagesection-mobile">
                         <SuperuserAlert />
                     </PageSection>
-                    <PageSection className='ct-overview-header' variant={PageSectionVariants.light}>
+                    <PageSection variant={PageSectionVariants.light} className='ct-overview-header'>
```
marusak, line 144: Same?

pkg/networkmanager/firewall.jsx
garrett: No, I think it needs padding as it has breadcrumbs, IIRC.

pkg/systemd/overview.jsx
garrett: Probably another one with some content inside that isn't holding a card view. (This time, the heading with system name and reboot.)

marusak:
- [arch](https://logs.cockpit-project.org/logs/pull-16379-20220215-094728-1630e9de-arch/log.html) is busted. Being fixed in https://github.com/cockpit-project/bots/pull/2944
- [debian-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16379-20220215-115331-1630e9de-debian-stable/log.html#70) and [debian-testing](https://logs.cockpit-project.org/logs/pull-16379-20220215-115328-1630e9de-debian-testing/log.html#70) are unhappy with firewall. Being fixed in https://github.com/cockpit-project/cockpit/pull/17004

PR #16380, opened 2021-09-22
lib: Pass Python scripts via "-c" instead of stdin

mvollmer:
This frees up stdin for use by the actual program.

Also, this fixes quoting of $@ to allow spaces in the arguments.


martinpitt (approved):
Wow, nice that this works! While this looks sad in `ps`, fixing python.invoke() for inputting passwords and other runtime parameterization is well worth it.

PR #14109, opened 2020-05-19
storage: Stratis PoC

mvollmer:
- There are no signals coming from the D-Bus API, that's why the code
  polls the ObjectManager.  Because of that, Cockpit's dbus channel
  doesn't do it's usual automatic Introspection, and that's why the
  code needs to use pool.data.Foo all over the place.

- When creating a Stratis filesystem, we should include mount point
  etc in the dialog.

- We should somehow indicate more clearly which filesystems share a
  pool on the Overview page.  I have an idea...

Demo: https://www.youtube.com/watch?v=Q0ZFYz-crNE&feature=youtu.be

PR #16381, opened 2021-09-22
bridge: Switch off CockpitDBusCache while it has no name owner

mvollmer:
Otherwise a cache for the default name would still react to signals
from foreign sources when the default name can't be activated.

### DIFF
--- a/src/bridge/cockpitdbusjson.c
+++ b/src/bridge/cockpitdbusjson.c
@@ -2703,7 +2703,13 @@ on_name_vanished (GDBusConnection *connection,
 
   CockpitDBusPeer *peer = g_hash_table_lookup (self->peers, name);
   if (peer)
-    cockpit_dbus_cache_set_name_owner (peer->cache, NULL);
+    {
+      // Using the empty string as the name owner means that the cache
+      // will not process any signals until we set a real name owner
+      // again in on_name_appeared.
+      //
+      cockpit_dbus_cache_set_name_owner (peer->cache, "");
+    }
 
   if (!G_IS_DBUS_CONNECTION (connection) || g_dbus_connection_is_closed (connection))
     cockpit_channel_close (channel, "disconnected");


mvollmer:
Let's see what the tests think.

This is still too hackish; each cache should watch its name owner independently...

src/bridge/cockpitdbusjson.c
```diff
@@ -2703,7 +2703,13 @@ on_name_vanished (GDBusConnection *connection,
 
   CockpitDBusPeer *peer = g_hash_table_lookup (self->peers, name);
   if (peer)
-    cockpit_dbus_cache_set_name_owner (peer->cache, NULL);
+    {
+      // Using the empty string as the name owner means that the cache
```
martinpitt, line 2707: I was going to complain about C++ style comments, but we already have a few in our code.

src/bridge/cockpitdbusjson.c
```diff
@@ -2703,7 +2703,13 @@ on_name_vanished (GDBusConnection *connection,
...
+      // will not process any signals until we set a real name owner
+      // again in on_name_appeared.
+      //
+      cockpit_dbus_cache_set_name_owner (peer->cache, "");
```
martinpitt, line 2711: This triggers this branch, right?
```
  if (self->name_owner && g_strcmp0 (sender, self->name_owner) != 0)
    return;
```

Would it be a more robust fix to add `!self->name_owner` to `on_properties_signal()` perhaps?

martinpitt (commented):
OK for me if you want to land this draft, just one question.

src/bridge/cockpitdbusjson.c
mvollmer: What do you mean exactly?  Turn the condition into this instead of the change in this PR?
```
if (!self->name_owner || g_strcmp0 (sender, self->name_owner) != 0)
  return;
```
This would certainly be more straightforward, and I had considered this actually.  However, a given CockpitDBusJson instance might create any number of CockpitDBusCache instances, but it will only ever set the name_owner of the cache instance for the default_name. Thus, all caches for non-default names will not ever process any signals.

That's what I referred to above, with "each cache should watch its name owner independently", and it would be fair to reject this PR here and go the whole way.

(It's not clear to me whether we actually make use of non-default names for a CockpitDBusJson instance ever, so the condition above might just work. If I am not mistaken, that would then mean that we are breaking a feature that we spend a lot of complexity on for its implementation, but that we don't use... would be worth figuring out, actually.)

martinpitt: My thinking yesterday was that `name_owner` has only two cases: it starts out as `NULL`, so CockpitDBusCache would (1) process signals for all unowned proxies; and (2) once the name appears, it will ignore signals for unexpected/weird senders. I admittedly don't understand all the details here -- in particular, why (1) would even be a thing -- if you try to connect to org.fd.foo, and that service isn't owned, how can it be possibly sending signals? and (2) when would a signal appear from a name that doesn't match the owner? That might be the same question as your

> It's not clear to me whether we actually make use of non-default names for a CockpitDBusJson instance ever

This PR now explicitly introduces a synthetic variant of (2) with the magic empty string. But is there really a point in treating them differently than the "unowned" (aka "service does not exist") variant?

So anyway, this isn't a rejection -- if you think this is an improvement, I'm ok to land. I just can't claim that I understand at all what's happening :sob:

martinpitt: I.e. in my simplistic world view, a service you connect to always has *one* name, doesn't exist otherwise, and if it sends a signal without a name, something is completely broken. I. e. if we can get away with
```c
if (!self->name_owner || g_strcmp0 (sender, self->name_owner) != 0)
  {
    g_warning("HUH, WTF? %s", sender);
    return;
  }
```

(slight string change suggested), then that would certainly simplify stuff a lot. :grin: Or can it happen that a D-Bus watch receives signals from "foreign" services somehow?

mvollmer: > Or can it happen that a D-Bus watch receives signals from "foreign" services somehow?

Yes, that's the "bug" we try to fix.

If you subscribe to a signal with a well-known name, like so
```
g_dbus_connection_signal_subscribe (connection,
                                    "org.freedesktop.UDisks2",
                                    "org.freedesktop.DBus.Properties",
                                    "PropertiesChanged",
                                    NULL, /* object_path */
                                    NULL, /* arg0 */
                                    G_DBUS_SIGNAL_FLAGS_NONE,
                                    on_properties_signal,
                                    self, NULL)
```
and the connection has match rules that allow signals from other sources, such as "org.storage.stratis2", then this subscription here will call on_properties_signal also for signals from "org.storage.stratis2".

This happens because glib can not filter based on a well known name, see here: https://github.com/ImageMagick/glib/blob/main/gio/gdbusconnection.c#L3539

(Upps, I wandered into the ImageMagick repo somehow, well good enough.)

So, ways to fix this "bug" are to only subscribe with the unique name, or to ignore signals that don't come from the expected unique name.  For both we need to track the owner of each name that we have a cache for.  Right now we only track the owner of the default name, and I was hesitant to change that.

With this incomplete fix, we have three cases for a cache:

1) no name tracking is done, allow all signals (thus subject to damage from foreign signals).
2) name tracking is done, name has no owner, ignore all signals
3) name tracking is done, name has a owner, ignore all signals not from owner

With this PR, we distinguish all three cases and set the name_owner of a cache to:

1) NULL
2) ""
3) "34:" or whatever

Without this PR, case 2) would behave like 1), i.e. the cache for the default name would be subject to damage from foreign signals while the name has no owner.


Yeah, so, I guess it would almost be easier to get rid of 1) right away instead of living for a while longer with the complicated mess that are the three cases...

martinpitt: Thanks for the details! This makes much more sense now.

To me it seems rather asymmetric to me that an unowned service should be different than a service which came and went again (owned/unowned), it feels like name_vanish should reset it back to the initial state. So if that works, and you agree that this would be a cleaner (although a bit more risky) fix, then we can at least do a CI run with that, and see how it fares?

Otherwise I trust your judgement here, and I'm happy to land this.

mvollmer: > To me it seems rather asymmetric to me that an unowned service should be different than a service which came and went again (owned/unowned)

It is not, one of on_name_appeared and on_name_disappeared will be called "immediately", so the cache for the default name will be initialized as either case 2 or 3.

mvollmer:
This fix here matters for #14109, where there is a test that disables org.storage.stratis2, and we might occasionally run into the described scenario.

mvollmer:
Ok, I have opened #16387, but right now I think this is all we need...

src/bridge/cockpitdbusjson.c
martinpitt: Ack, thanks for clarifying!

martinpitt (approved):

PR #16393, opened 2021-09-24
test: Fix TestShutdownRestart.testBasic for different OS timezones

martinpitt:
This test queried the host's date to schedule a shutdown for the VM.
Our RHEL images run in the EDT time zone, so after 20:00 EDT the host
already moved to the next day, and the test would schedule the shutdown
for the wrong day. Avoid this by querying the VM's date instead.

----

This has broken pretty much every automatic RHEL/CentOS image refresh in a long time, as they keep running in the early UTC morning while it's still "yesterday" on these images. [example log](https://logs.cockpit-project.org/logs/pull-2450-20210924-023137-2306c660-rhel-8-5-cockpit-project-cockpit/log.html#273-2) from https://github.com/cockpit-project/bots/pull/2450

### DIFF
--- a/test/verify/check-system-shutdown-restart
+++ b/test/verify/check-system-shutdown-restart
@@ -18,7 +18,6 @@
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
 import parent
-from datetime import date
 from testlib import *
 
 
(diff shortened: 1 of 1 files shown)


martinpitt:
```
This is reproducible by calling this at the top, and running on rhel-8-5 image (which has EDT timezone):

    m.execute("timedatectl set-ntp off; sleep 3; timedatectl set-time 'Thu 2021-09-23 23:19:23'")
```

marusak (approved):
Thanks!
While this fixes the test, it still is case of https://github.com/cockpit-project/cockpit/pull/15626 - that users might actually be confused when doing such actions through Cockpit.
But solution to ^ is non-trivial, so this is of course good for now :)

martinpitt:
Retrying unrelated flake, I sent a fix to https://github.com/cockpit-project/bots/pull/2451

PR #16394, opened 2021-09-24
test: Remove old storage hack

marusak:
I am only testing this if it still fails.


martinpitt (approved):
Worked three times in a row, so I suppose :+1: Thanks!

marusak:
> Worked three times in a row

I retried 3 times before, so at least 12 times.

PR #16397, opened 2021-09-24
two trivial build fixes

allisonkarlitskaya:


configure.ac
```diff
@@ -490,9 +490,7 @@ else
 fi
 AC_MSG_RESULT($enable_strict)
 
-# Python version; prefer python3
-m4_define_default([_AM_PYTHON_INTERPRETER_LIST], [python3 python2 python])
-AM_PATH_PYTHON([2.7])
+AM_PATH_PYTHON([3.8])
```
martinpitt, line 493: This [breaks on RHEL 8](https://logs.cockpit-project.org/logs/pull-16397-20210924-132311-36ca5839-rhel-8-6/log.html), it only has Python 3.6.

martinpitt (changes requested):
One character change away from looking good :grin:

martinpitt (approved):
Cheers!

PR #16400, opened 2021-09-27
common: Use sigabbrev_np() instead of our own signal string table

martinpitt:
This has been available in glibc for a long time. All of our table
entries are present in glibc [1], except for these three aliases:
SIGCLD (same as SIGCHLD), SIGIO (same as SIGPOLL), and SIGUNUSED
(same as SIGSYS).

Still keep cockpit_strsignal() as such, as it also builds a name for
runtime signals and returns "UNKNOWN" instead of `NULL` for an unknown
signal.

[1] https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/generic/siglist.h

### DIFF
--- a/src/common/cockpitunixsignal.c
+++ b/src/common/cockpitunixsignal.c
@@ -40,72 +40,20 @@
  * Copyright (C) 2014 Karel Zak <kzak@redhat.com>
  */
 
+/* for sigabbrev_np() */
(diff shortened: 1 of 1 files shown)


martinpitt:
Ah dang, `sigabbrev_np()` is glibc ≥ 2.32 only

PR #16402, opened 2021-09-27
users: Add an list of recent logins to the account details

jelly:
Show a list of the 15 recent logins on the account details page, the last
login health card now includes a link to the account detail page.

Closes: #28

---

# Users: Login history

The user account details page now includes a list of your 15 most recent logins.

users-login-history

### DIFF
--- a/pkg/systemd/overview-cards/lastLogin.jsx
+++ b/pkg/systemd/overview-cards/lastLogin.jsx
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 import React, { useState, useEffect } from 'react';
-import { Flex, FlexItem } from '@patternfly/react-core';
+import { Button, Flex, FlexItem } from '@patternfly/react-core';
 import { ExclamationTriangleIcon, UserIcon } from "@patternfly/react-icons";
 
 import * as timeformat from "timeformat";
@@ -52,6 +52,7 @@ const getFormattedDateTime = (time) => {
 
 const LastLogin = () => {
     const [messages, setLoginMessages] = useState(null);
+    const [name, setName] = useState(null);
 
     useEffect(() => {
         if (messages === null) {
@@ -71,7 +72,10 @@ const LastLogin = () => {
                         console.error("failed to fetch login messages:", error);
                     });
         }
-    }, [messages]);
+        if (name === null) {
+            cockpit.user().then(user => setName(user.name));
+        }
+    }, [messages, name]);
 
     if (messages === null || !messages['last-login-time']) {
         return null;
@@ -123,6 +127,15 @@ const LastLogin = () => {
                     {failedLogins &&
                     <FlexItem id="system_last_login_success" className="pf-u-text-break-word">
                         {lastLoginText}
+                    </FlexItem>
+                    }
+                    {name &&
+                    <FlexItem>
+                        <Button variant="link" isInline
+                                className="pf-u-font-size-sm"
+                                onClick={() => cockpit.jump("/users#/" + name)}>
+                            {_("View login history")}
+                        </Button>
                     </FlexItem>
                     }
                 </Flex>
--- a/pkg/users/account-logs-panel.jsx
+++ b/pkg/users/account-logs-panel.jsx
@@ -1,0 +1,92 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2021 Red Hat, Inc.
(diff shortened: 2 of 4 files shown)


jelly:
> Show a list of the 5 recent logins on the account details page, the last
> login health card now includes a link to the account detail page.
> 
> Closes: #28

That said, I'm not 100% sure it closes #28

pkg/users/account-logs-panel.jsx
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }, [logins, name]);
+
+    return (
+        <Card id="account-logs">
+            <CardTitle>
+                <Text component="h2">{_("Account logs")}</Text>
+            </CardTitle>
+            <CardBody>
+                <DataList aria-label="Account log list" isCompact className="account-logs-list">
```
KKoukiou, line 81: aria-labels also need to be translated.

pkg/users/account-logs-panel.jsx
```diff
@@ -0,0 +1,123 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from 'cockpit';
+import React, { useState, useEffect } from 'react';
+
+import { Card, CardTitle, CardBody, DataList, DataListItem, DataListItemRow, DataListItemCells, DataListCell, Text } from '@patternfly/react-core';
```
KKoukiou, line 23: Do you have some mockup for this? 
The issue is that I am afriad you 're using the wrong component here.
If you want that header lines (table header) to be present you can't use data list, you need to use table, or our implementation of it, aka ListingTable.


When using this the mobile mode is broken, as headers appear as a cell. 
Screen Shot 2021-09-27 at 15 50 21

KKoukiou (changes requested):

pkg/users/account-logs-panel.jsx
jelly: > Do you have some mockup for this?

The mockup is here https://raw.githubusercontent.com/cockpit-project/cockpit-design/9a65b44ffa9cf60a6eaf02044e797d0046b77d6b/users/login-warning.png

> The issue is that I am afraid you 're using the wrong component here.
> If you want that header lines (table header) to be present you can't use data list, you need to use table, or our implementation of it, aka ListingTable.

Aha!

KKoukiou: It's a ListingTable with a contains-list class (we already use it) on the table body.

pkg/users/account-logs-panel.jsx
jelly: done!

garrett (changes requested):
Looks good overall!

I request these small tweaks:
- Heading should be "Login history"
- Table should be compact, show a few more entries if possible (possibly 10, 15)
- Alignment between page sections is not consistent, but it should be

Screenshot from 2021-09-28 16-12-05

garrett:
Previous: 8690496
My changes: 3cd066a76

I squashed it into the original commit.

garrett (dismissed):
With my little tweaks (@ 3cd066a76), I'm happy with this PR.

garrett:
I've updated the screenshot and description as well.

pkg/systemd/overview-cards/lastLogin.jsx
```diff
@@ -125,6 +129,15 @@ const LastLogin = () => {
                         {lastLoginText}
                     </FlexItem>
                     }
+                    {name &&
+                    <FlexItem>
+                        <Button variant="link" isInline
```
KKoukiou, line 134: Generally we should use `<a>` component when we get redirected to other pages, also inside cockpit. That said component='a' here with an href. 

However it's wrong all over cockpit, one more won't make a difference. We can fix it in a big batch in the future.

KKoukiou (approved):

KKoukiou (commented):
there is one failure in https://logs.cockpit-project.org/logs/pull-16402-20210928-154922-172490ee-fedora-coreos/log.html#289

garrett:
Aha! The spacing was fixed in wide modes, but it's collapsing oddly in narrower modes and I screenshot the narrower version. That's why I thought it was fixed (it is, in larger sizes) but the screenshot isn't.

I'll adjust the screenshot. We can fix things in follow-ups.

marusak (approved):

PR #16407, opened 2021-09-28
autogen.sh: Download webpack cache during release

martinpitt:
Our dist/ is very reliable now, and gets built in exactly the same
GitHub workflow environment as the release. So use it instead of
rebuilding everything again.

As cockpituous' release-source locally clones the checkout again, we
can't just put the webpack-jumpstart call into
tools/cockpituous-release. Do it in autogen.sh instead when releasing,
i.e. if `$RELEASE_SOURCE` is set.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -6,4 +6,9 @@ srcdir="$(realpath -m "$0"/..)"
 
 (cd "${srcdir}" && autoreconf -is --warnings obsolete)
 
+# When calling this as part of cockpituous release-source, download cached dist/
+if [ -n "${RELEASE_SOURCE:-}" ] && [ ! -d dist ]; then
+    tools/webpack-jumpstart --wait
+fi
+
 [ -n "${NOCONFIGURE:-}" ] || exec "${srcdir}/configure" "$@"


martinpitt:
I tested this with a crippled down cockpituous-release, a local test tag, in a `ghcr.io/cockpit-project/release` container.

martinpitt:
@allisonkarlitskaya : This is not exactly elegant, maybe you have a better idea? We could of course add this knowledge to https://github.com/cockpit-project/cockpituous/blob/main/release/release-source#L99 but that doesn't feel right either.

allisonkarlitskaya:
My biggest question: what do we gain with this?

Also: instead of the magic envvar, I'd prefer to modify the release script itself to check if the package in question contains a jumpstart script and call it before autogen.  But again: why?

autogen.sh
```diff
@@ -6,4 +6,9 @@ srcdir="$(realpath -m "$0"/..)"
 
 (cd "${srcdir}" && autoreconf -is --warnings obsolete)
 
+# When calling this as part of cockpituous release-source, download cached dist/
+if [ -n "${RELEASE_SOURCE:-}" ] && [ ! -d dist ]; then
+    tools/webpack-jumpstart --wait
```
allisonkarlitskaya, line 11: In any case, you'd want to run that in `${srcdir}`.

allisonkarlitskaya (changes requested):

allisonkarlitskaya:
Just chatted a bit more on IRC and here's another one:

Right now, the tarball we release is a deterministic product of the build environment (presumably clean, from GitHub Actions) and the signed SHA of the release tag.  Everything else that we pull in (npm modules for example) is cryptographically verified (via the gitref link, for example).  It took a lot of work to get to this point.

In theory, the dist cache is always the same, but it's also a potential weakness: there's absolutely no verification of what we download.  That could be a way for an attacker to get something into a tarball which otherwise looks like it was generated from a signed commit.  I think I'd rather just spend the extra 5 minutes twice a week and generate it for ourselves.

martinpitt:
> I'd prefer to modify the release script itself to check if the package in question contains a jumpstart script and call it before autogen.

I don't, honestly -- this hardcodes too much internal knowledge into the release scripts, which aim to be generic. Ideally this would stay in tools/cockpituous-release, but for that we need to add some extra support into release-source. I'm happy to do that *if* we agree to do it at all.

> But again: why?

Aside from the koji builds (which I have on my TODO list separately, mostly to parallelize them), the webpack is the second longest chunk of the release process, and seemed like an easy win.

You seem to have some reservations with doing this at all (clean build, using the dist cache as trojan horse vector), and I don't have a strong opinion about it -- I just wanted to get this off my todo list. At least it's now in GitHub's brain in case we want to dig this out in the future.

PR #16409, opened 2021-09-28
systemd: Set auto-detected realm domain/admin as proper input values

### LINKED ISSUE #16390: Cockpit should pre-populate some of the domain-join fields
Page: Overview

When joining a domain, there are two fields that could be auto-filled by Cockpit under some circumstances.

I propose the following:

1. Domain Field - If the system hostname is a FQDN, Cockpit should call `realm discover <FQDN>` and if it discovers a matching domain, pre-populate the field (with the option to edit it, of course).
2. Domain Administrator Name - Once the FreeIPA or AD domain is identified, this field should be pre-populated with the default administrative user for that domain type. For FreeIPA, the default administrator is `admin`; for Active Directory it is `Administrator`.

martinpitt:
Setting them as placeholders was confusing:
 - They got prefixed with "e.g.", quoted, and greyed out, and thus did
   not look like proper values.
 - Leaving the "domain" field empty used the auto-detected value; but
   leaving the "admin" field empty literally tried to authenticate as
   empty user name, so that field had to be filled.

Bring the dialog in line with the standard behaviour, use static fixed
placeholders (in case there is no auto-detected domain), and set
auto-detected realm/user names as actual values. With that, joining a
domain only requires to fill the admin password, everything else has
sane defaults.

Update the tests to expect pre-filled autodetected domain/admin names.
In TestPackageInstall.testDialogTransition, check that they remain empty
if there is no realm available. Introduce a `data-discover` attribute so
that the test can know when discovery finished and found no realms.

Fixes #16390

----

I also added a second commit which fixes the "Join" button spinner to actually appear. Right now, it's just a static gray misaligned button.

It looks like this now right after clicking the "Join domain.." link in the Configuration card:

image

And like this if there is no detected domain:

image

There are more things wrong with this dialog: The "Join" button is not disabled while there is incomplete input, and the dropdown triangle  is oddly placed. I plan to address these when rewriting this dialog in React/PF4.

### DIFF
--- a/pkg/systemd/overview-cards/realmd-operation.js
+++ b/pkg/systemd/overview-cards/realmd-operation.js
@@ -113,13 +113,11 @@ function instance(realmd, mode, realm, state) {
         auth_changed($(this));
     });
 
-    let title, label;
+    let title;
     if (mode == 'join') {
         title = _("page-title", _("Join a domain"));
-        label = _("Join");
         $(".realms-op-join-only").show();
         $(".realms-op-leave-only-row").hide();
-        $(".realms-op-apply").text(label);
         check("");
     } else {
         title = _("page-title", _("Domain"));
--- a/pkg/systemd/overview-cards/realmd-operation.js
+++ b/pkg/systemd/overview-cards/realmd-operation.js
@@ -195,6 +195,9 @@ function instance(realmd, mode, realm, state) {
                             $("#realms-op-address").parent()
                                     .removeClass("has-success")
                                     .addClass("has-error");
+                            $("#realms-op-address").attr("data-discover", "failed");
+                        } else {
+                            $("#realms-op-address").attr("data-discover", "empty");
                         }
 
                         realm = null;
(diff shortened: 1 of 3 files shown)


martinpitt:
This fails on Ubuntu 20.04 and RHEL 8.5, needs some work.

pkg/systemd/overview-cards/realmd-operation.js
```diff
@@ -211,6 +214,12 @@ function instance(realmd, mode, realm, state) {
                         realm = realmd.proxy(REALM, path);
                         $(realm).on("changed", update);
                         realm.wait(function() {
+                            // fill in unset realm/user with autodetected values
```
KKoukiou, line 217: It would make sense to have the input disabled while auto-discovery is running so that we avoid stepping over inserted text from the users.

So:
* disable fields
* run autodiscovery
* insert values and remove disabled attribute

KKoukiou (commented):
Just one comment. Is realm domain discovery something instant?

KKoukiou (approved):

PR #16411, opened 2021-09-28
po-refresh.yml: split out pot upload to weblate

allisonkarlitskaya:
The current po-refresh job does two things:

  - generates an updated cockpit.pot and commits it to the weblate repo

  - copies .po files from the weblate repo and opens a PR with them

These two things are actually completely independent from each other, so
let's separate them out to allow for more fine-grained use of
credentials.

Introduce a update-pot job to the existing po-refresh workflow which
takes care of updating cockpit.pot and committing the results to the
weblate repository.  This job requires very few permissions: it only
needs to write to the weblate repository and it does that by way of a
deploy key.

Meanwhile, remove the commands from the `update-pot` rule in
po/Makefile.am to prevent this job from being done twice.  We need to
keep the target itself because bots/po-refresh expects it to exist.

 * [x] successful test run: https://github.com/cockpit-project/cockpit/runs/3733559193?check_suite_focus=true → https://github.com/cockpit-project/cockpit-weblate/commit/b7e48b2344968dd10854efb84a647446e92b7ea6

### DIFF
--- a/.github/workflows/po-refresh.yml
+++ b/.github/workflows/po-refresh.yml
@@ -7,6 +7,43 @@ on:
   # can be run manually on https://github.com/cockpit-project/cockpit/actions
   workflow_dispatch:
 jobs:
+  pot-upload:
+    environment: cockpit-weblate
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set up dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y --no-install-recommends gettext
+
+      - name: Clone source repository
+        uses: actions/checkout@v2
+        with:
+          path: src
+
+      - name: Install node_modules
+        run: make -C src -f pkg/build package-lock.json
+
+      - name: Generate .pot file
+        run: make -C src -f po/Makefile.am po/cockpit.pot
+
+      - name: Clone weblate repository
+        uses: actions/checkout@v2
+        with:
+          path: weblate
+          repository: ${{ github.repository }}-weblate
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
+
+      - name: Commit .pot to weblate repo
+        run: |
+          cp src/po/cockpit.pot weblate/cockpit.pot
+          git config --global user.name "GitHub Workflow"
+          git config --global user.email "cockpituous@cockpit-project.org"
+          git -C weblate commit -m "Update source file" -- cockpit.pot
+          git -C weblate push
+
   po-refresh:
     runs-on: ubuntu-20.04
     steps:
--- a/po/Makefile.am
+++ b/po/Makefile.am
@@ -112,10 +112,8 @@ update-po: po/cockpit.pot
 $(WEBLATE_REPO):
 	git clone --depth=1 -b $(WEBLATE_REPO_BRANCH) $(WEBLATE_REPO_URL) $(WEBLATE_REPO)
 
-upload-pot: po/cockpit.pot $(WEBLATE_REPO)
-	cp $(builddir)/po/cockpit.pot $(WEBLATE_REPO)
-	git -C $(WEBLATE_REPO) commit -m "Update source file" -- cockpit.pot
-	git -C $(WEBLATE_REPO) push
+upload-pot:
+	@true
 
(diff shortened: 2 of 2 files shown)


po/Makefile.am
```diff
@@ -112,10 +112,8 @@ update-po: po/cockpit.pot
 $(WEBLATE_REPO):
 	git clone --depth=1 -b $(WEBLATE_REPO_BRANCH) $(WEBLATE_REPO_URL) $(WEBLATE_REPO)
 
-upload-pot: po/cockpit.pot $(WEBLATE_REPO)
-	cp $(builddir)/po/cockpit.pot $(WEBLATE_REPO)
-	git -C $(WEBLATE_REPO) commit -m "Update source file" -- cockpit.pot
-	git -C $(WEBLATE_REPO) push
+upload-pot:
```
martinpitt, line 115: This is a hack until we adjust bots/po-refresh to not expect the target any more, right? And consequently fix podman/machines to separate POT uploads as well. I don't think we should keep a broken rule permanently.

.github/workflows/po-refresh.yml
```diff
@@ -7,6 +7,43 @@ on:
   # can be run manually on https://github.com/cockpit-project/cockpit/actions
   workflow_dispatch:
 jobs:
+  pot-upload:
```
martinpitt, line 10: Fine for me as an incremental step to separate the privileges. Eventually I'd prefer this as a standalone pot-refresh.yml workflow, so that it can be run independently, scheduled daily, and more clearly structures what's going on.

martinpitt (commented):
Thanks @allisonkarlitskaya ! I like this even more than the previous suggestion of moving that workflow to the -weblate repo. It requires an extra deploy key, but by now we know how to handle these, and it avoids any magic in the weblate repo.

I'd like to discuss how the final solution will look like, to make sure we agree where we want to go. From my POV:

 - Split out a pot-refresh.yml action, and put it into c-{machines,podman}; generate deploy keys for these three
 - Stop bots/po-refresh to do `upload-pot`.
 - Add a bots/pot-refresh which receives (most of) the code of that, to minimize duplication of logic in the workflow, and make it easier to run/debug manually.
 - And most importantly: With this we should be able to drop `secrets.COCKPITUOUS_TOKEN` from the po-refresh step, and just use the default GitHub token.

@allisonkarlitskaya , @marusak WDYT? I'm happy to do a quick gmeet about this for faster turnaround.

po/Makefile.am
allisonkarlitskaya: This is a compatibility shim for the other projects, yes.  I'm not ready to rip out the invocation of that rule from `bots` until the other projects also don't need it.

.github/workflows/po-refresh.yml
allisonkarlitskaya: I actually think it makes sense for both of them to run as separate jobs inside of the same workflow, triggered at the same time, but I'm happy to discuss that further.  The manual trigger argument might be a good one, for example.

po/Makefile.am
martinpitt: Yes, of course. I do agree we need to keep it for a bit, I just wanted to ensure it's not going to stay like this forever.

.github/workflows/po-refresh.yml
martinpitt: For translators it makes sense to get more frequent POT updates, otherwise they'd have up to a week delay between us landing a new string and them being able to translate it. OTOH we don't want daily PO refresh PRs, that's just too much noise. For that reason, and because they do independent things, I think two separate workflows would be cleaner.

allisonkarlitskaya:
I think the order of the steps might be a bit different than what I had planned.  Next, I'd like to work on getting the cockpituous token out of the workflow for updating the .po files in cockpit.

Ultimately I also want to remove the process of copying *.po between the repos from the build systems and from the bots script as well: that will fall to the workflow, and the only thing the bots script will do is create commits, push them, and open the PR.  As an intermediate step, I plan to use an environment variable to switch off the building step inside the bots script (since we hardly need a full automake in order to copy *.po and commit it).

martinpitt (approved):

PR #16413, opened 2021-09-28
storage: Layout tweaks

mvollmer:
- Usage bars don't span the whole page width any more

- Description lists follow standard Patternfly layout, except when
  there are very long labels.

- There is less padding at the start of a nested form in a dialog.

- The teardown footer message is placed at the bottom of the dialog
  body, to get better padding, especially when there is nothing else
  in the body.


mvollmer:
>  * There is less padding at the start of a nested form in a dialog.

This isn't specific to storage, actually, and might be better fixed in our patternfly overrides maybe.

mvollmer:
>   * The teardown footer message is placed at the bottom of the dialog
>     body, to get better padding, especially when there is nothing else
>     in the body.

This is arguably wrong... this will cause the tear down message to scroll with the rest of the body.

mvollmer:
I split this into separate commits and try to figure out how to keep the teardown stuff in the footer.

mvollmer:
> keep the teardown stuff in the footer.

Actually, the teardown stuff can be large, so it might be better to keep it in the scrolling part of the dialog. We should instead say "Unmount and ..." in the apply button so that people are aware that there is some teardown going on even if the full info about it is scrolled out of view.

mvollmer:
This tries to fix the following comments from https://github.com/cockpit-project/cockpit/pull/14109#issuecomment-927732119.

 - Way too much space between encrypt data and the passphrase
 - Progress bar is too long; the usage information should be closer to its label
 - Too much space between label and values [in DescriptionList]
 - It's weird that it even mentions unmounting when it's going to be deleted. [About dialogs that do some teardown before their main action. Not addressed here yet.]
 - There's too much space between the heading and the content and not enough space between currently in use line and the list.
 - The list has too much space between the items; perhaps they should both be left aligned? [Not yet done]

mvollmer:
A typical "confirm deletion" dialog looks like this now:

image

@garrett, should this say "Unmount and delete", or should we remove "Proceeding will unmount all filesystems on it" and keep just "Delete" in the apply button?  We recently changed the "Format" dialog to explicitly say "Unmount and format", so now I am not super sure which way to go.

(Also, the "/mnt" and "/dev/sda1" texts need to be close together.)

mvollmer:
We could redesign the teardown info to be list instead of the awkward table:

```
This device is currently in use.
- /dev/sda1 is mounted on /run/data. It will be unmounted.
- /dev/sda2 is mounted on /run/data2. It will be unmounted.
- /dev/sda3 is a member of the LVM2 volume group foo. It will be removed.
```

garrett:
> should this say "Unmount and delete", or should we remove "Proceeding will unmount all filesystems on it" and keep just "Delete" in the apply button? 

At first, I thought unmount and delete makes some sense, but unmounting is kind of a detail here... would someone expect something to be deleted and still be mounted (somehow)?

If it's unclear without it, unmount and delete works though.

> We recently changed the "Format" dialog to explicitly say "Unmount and format", so now I am not super sure which way to go.

Well, we should be consistent. I guess unmount and delete works.

> Should we use "/!\ Permanently delete /dev/sda1" as the dialog title?

Yes, I think I mentioned it somewhere?

> We could redesign the teardown info to be list instead of the awkward table:
>
> This device is currently in use.
> - /dev/sda1 is mounted on /run/data. It will be unmounted.
> - /dev/sda2 is mounted on /run/data2. It will be unmounted.
> - /dev/sda3 is a member of the LVM2 volume group foo. It will be removed.

It's verbose, but probably much clearer. I think it makes sense.

mvollmer:
Here is a dialog with big teardown info. (I add the cleaned up version shortly.)

image

It shows all three types of uses that Cockpit can currently tear down.  Each kind of use can theoretically have arbitrary many entries in its table.  Practically, there might be as many as you have logical volumes in a volume group, or partitions on a disk.

mvollmer:
Here it is with a list:

image

I used `TextContent, `TextList` and `TextListItem` for this...

garrett:
Both dialogs are too verbose and have spacing issues. I guess I should recreate this as a mockup and play around with it until it looks right to me, then we could recreate the mockup in the code.

garrett:
How's this? I'm basically using a compact table to list the partitions, the associated location with every partition, and the action that will be performed. I've also reworked the heading, the helper text at the bottom, and changed the "Erase" select to a "Secure erase" option checkbox, making it more similar to the format partition dialog.

image

Wait. Shouldn't this be "Initialize disk", not "Format"? It's setting up the disk, whereas formatting is more for the partitions?

garrett:
BTW: I don't have the table stretch all the way, to keep the items near each other. We could stretch it though. The lines help to associate the fields anyway. Not a big change, either way.

mvollmer:
>  and changed the "Erase" select to a "Secure erase" option checkbox

The "Format" dialog still calls this "Erase" and not "Secure erase". I think we should not say "Secure" since it is debatable whether simply overwriting a block device with zeros is actually secure, because of all the caching and wear leveling redirection that drives do.

mvollmer:
> How's this?

Nice!  So it should just be "Format" in the apply button, not "Unmount and format"?

mvollmer:
This is how it looks now:

image

garrett:
> The "Format" dialog still calls this "Erase" and not "Secure erase". 

Whatever it's called, it should be consistent.

> I think we should not say "Secure" since it is debatable whether simply overwriting a block device with zeros is actually secure, because of all the caching and wear leveling redirection that drives do.

It's often called "Secure erase" elsewhere, but, yeah, it might give someone a false sense of security.

We shouldn't call it "Erase" as it's erasing by default. This makes it look like it's not erasing if we just call it "erase" and the option is unchecked.

What do you think of "Full erase" or "Complete erase" or "Thorough erase" or "Overwrite erase" or just "Overwrite"?

> This is how it looks now:

Isn't that misleading, if it has "unmount" as an action? It would need to show that it's getting unmounted and being erased, right?

Also, shouldn't it be "Partition" instead of "Device" in the list? Everything there is a partition, not actually a device.

(And seeing it full-width shows that it's fine, as there are lines to help guide someone's eye to show the relationship between the partition and action. :wink:)

garrett:
Here's what GNOME Disks does for the erase option:

Screenshot from 2021-10-04 12-04-16

And Windows 10 (which seems to invert it and it might overwrite if it's unselected, but it's not clear):

format-hardrive-windows-10_2020_ L1 _ L1 _ L1

And macOS (it has a sub-sub-dialog with a slider):

Disk-Utility-Erase

Security-Options


I guess it's fine enough as-is. We might want to change the string to add a "(slower)" like this: "Overwrite existing data with zeros (slower)"... what do you think? And/or choose one of the labels in my previous comment, but I guess it's optional really if it works like this here. I'm probably overthinking it. :wink:

mvollmer:
> Isn't that misleading, if it has "unmount" as an action? It would need to show that it's getting unmounted and being erased, right?

Ah, I think I see where you want to go.  The table would be a summary of everything that will happen, including deletion of partitions, formatting of block devices, etc.  Right?

Currently we only summarize certain possibly unexpected extra things that will have to be done before carrying out the main operation (like unmounting a filesystem before formatting the block device).

We could argue whether to say anything about these extra preparations at all, actually.  Maybe the simplest way out is to just drop them altogether.  After all, if you initialize a disk, all its former filesystems need to be unmounted first, and Cockpit might just silently do it.

But then again, I want to eventually go further than just unmounting a filesystem, and let the user kill sysessions and services that are using a filesystem that needs to be unmounted. And those actions should be confirmed, I'd say.

Here are more examples:

image

Would you show the "This device is in use" table also when those filesystems would not be mounted?

Here is the simplest case, formatting a mounted filesystem.

image

mvollmer:
Here is what happens with NFS, which can already kill sessions and services:

image

garrett:
> Ah, I think I see where you want to go. The table would be a summary of everything that will happen, including deletion of partitions, formatting of block devices, etc. Right?

Yep.

> Would you show the "This device is in use" table also when those filesystems would not be mounted?

No, that's already covered by the dialog's messaging itself. The table shows the actions to everything effected, and if the format/initialize options aren't included, it feels like they won't happen to that partition (as another option is shown instead). 

Does that make sense?

> Here is what happens with NFS, which can already kill sessions and services:

Huh. Interesting.

The messaging is wrong, as a _session_ isn't stopping the unmounting, but a _process_ is. (It belongs to a session, but it's not just the session.)

The padding / alignment looks wrong. It's probably because lines are missing. Either it needs lines or the alignment needs fixing.
Vertical space also feels a bit off for whatever reason.

garrett:
So here's the thing: The changes table really should be used for > 1 change, to get an overview on what will be different on the system. For changing something that only affects the intended target, it should really be integrated into the dialog.

What I mean, in plainer text is that the unable to mount the filesystem should show which filesystem and it should be force unmount. Does it kill the processes or does it unmount despite the processes?

 In other words, from https://www.systutorials.com/force-linux-unmount-filesystem-reporting-device-busy/, is it option 1 (force unmount) or 3 (fuser -k, then unmount)? (I guess we're using option 2 (lsof) or 3 (fuser w/o -k) to get the processes?

And also: For the formatting a single partition when it already exists, the dialog itself communicates this, so why should we also say it will be unmounted? That's useless information, when everything else says it will be destroyed. Having the unmount... and only unmount... as the option makes it look like it will only be unmounted. But it's still redundant even if it had just one entry saying it will be unmounted and reformatted.

garrett:
So I've revised the mockup and split it into multiple conditions, as I outlined above. Hopefully this helps.

I've added the two helpertext messages. We might need to override to as inline variant. If it doesn't support it directly (such as setting 'component' to 'span', which isn't shown as supported, but might work), we could use one of the PatternFly helpers (and attach the appropriate "className" in the JSX): https://www.patternfly.org/v4/utilities/display

To summarize the differences:

- Only show the changes when there are > 1 changes
- Show the device currently in use message when it's currently in use
- Show the erases all data message when there's existing content
- Currently in use and erases data helper texts messages will usually both be shown in most cases (it's possible to have unmounted disks that have data, so they'd show the data message even if not mounted)
- I changed the labels
- I revised the disk-based mockups to be "initialize" (format is for partitions, initialize for disks)

## Single, unused 

image

## Single, used

image

## Multiple changes, unused

(This doesn't happen, right? :wink:)

## Multiple changes, used

image

mvollmer:
> So I've revised the mockup

Let's continue in #16431 with the teardown stuff, it starts to look like a lot more than tweaking whitespace. :-)  I have copied your comment over already.

mvollmer:
Here are screenshots for all the changes in this PR.

## Header panels of detail pages

Before:
image

After:
image

Less space between labels and values, shorter usage bars

## Dialogs

Before:
image

After:
image

Icon in title but not danger message, danger message in red, better margins for the teardown table, no stretching of its columns, table can scroll out of view, apply button is more verbose (in case teardown table scrolls out of view)

Before:
image

After:
image

"Format disk" is now "Initialize", "Erase" is no longer the first field and its a checkbox.

mvollmer:
> We might want to change the string to add a "(slower)" like this: "Overwrite existing data with zeros (slower)"... what do you think?

Good idea!

> What do you think of "Full erase" or "Complete erase" or "Thorough erase" or "Overwrite erase" or just "Overwrite"?

I think "Overwrite" is best.

garrett:
The tables still aren't tables. Tables have headings and are aligned properly. (Your initialize after shows that the mount points aren't aligned.) Additionally, the text above the table needs space between it and the table.

 But I guess we'll work on that in the other PR (#16431)?

mvollmer:
> The tables still aren't tables. Tables have headings and are aligned properly.

True, ouch, I missed the misalignment.  The code uses Patternfly DataList, which is indeed not a table. We can get some alignment out of it, but not perfectly I guess.  I change it up to use relative column widths of 2, 1, and 4.

> But I guess we'll work on that in the other PR (#16431)?

Yes, but we shouldn't make it worse here, I'd say.

garrett:
>  The code uses Patternfly DataList, which is indeed not a table. We can get some alignment out of it, but not perfectly I guess

Yeah, the DataList shouldn't really be used except for simple lists. It doesn't align properly.

https://www.patternfly.org/v4/components/data-list/design-guidelines/#data-list-elements:
>
> PatternFly offers two components for displaying large data sets: data lists and [tables](https://www.patternfly.org/v4/components/table/design-guidelines). While they satisfy similar use cases, choosing the correct component to use in your design will be dependent on the type of data you need to display.
> 
> Use data lists when:
> 
> - A flexible layout is more important than arranging information in a grid.
> - You want to include active content like a chart.
> - Content displayed may vary between rows.

mvollmer:
> I change it up to use relative column widths of 2, 1, and 4.

That doesn't work nicely with long device names, so let's just keep it as it is for now.

garrett:
>>    I change it up to use relative column widths of 2, 1, and 4.
>
> That doesn't work nicely with long device names, so let's just keep it as it is for now.

The number of items doesn't really matter and an HTML table-based widget, such as the PatternFly table should stretch to accomodate the content of the cells.

Plus, without headers, which tables provide, it's not really obvious what each bit of data is by context.

pkg/storaged/content-views.jsx
```diff
@@ -636,8 +636,11 @@ const BlockContent = ({ client, block, allow_partitions }) => {
                           })
             ],
             Action: {
-                Title: teardown_and_format_title(usage),
-                Danger: _("Formatting a disk will erase all data on it."),
+                Title: teardown_and_apply_title(usage,
```
mvollmer, line 637: This change is in the wrong commit...

mvollmer: Moved.

pkg/storaged/storage.scss
```diff
@@ -408,3 +408,7 @@ td.storage-action {
 .pf-c-description-list .pf-c-progress {
     max-width: 30ch;
 }
+
```
KKoukiou, line 414: this essentially overrides https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/patternfly/patternfly-4-overrides.scss#L128 

We need to fix that rule instead.

KKoukiou (changes requested):

pkg/storaged/storage.scss
KKoukiou: The way nested forms are implemented in the storage page within dialog.jsx is wrong in the first place. The nested element should just be children of the FormGroup that they belong to.
Like here: https://codesandbox.io/s/f3gfq

PR #16427, opened 2021-10-04
systemd: Drop computer-ou input field from realm join dialog

martinpitt:
This has never worked -- literally the first-ever commit (87325b96bc343)
got this wrong (the code queried `.realms-join-computer-ou`, while the
HTML defines `#realms-op-computer-ou`). This never had an integration
test, testing it does not work with our Samba server (as it does not
define any "Computers" OU and `adcli show-computer` does not show the OU
either). Also, that feature is incomplete (there are many more fields
that one can set [1]), and nobody complained in all the years.

So let's just remove this, and only bring it back in the future when
there is some actual demand. This also gets rid of the
`.realm-active-directory-only` class special case.

[1] https://freedesktop.org/software/realmd/docs/gdbus-org.freedesktop.realmd.KerberosMembership.html

### DIFF
--- a/pkg/systemd/overview-cards/configurationCard.jsx
+++ b/pkg/systemd/overview-cards/configurationCard.jsx
@@ -39,7 +39,6 @@ export class ConfigurationCard extends React.Component {
         super(props);
 
         this.state = {
-            serverTime: '',
             hostEditModal: false,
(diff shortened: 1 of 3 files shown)


marusak (approved):
:) Thanks!

PR #16428, opened 2021-10-04
networkmanager: Port dialogs to React

KKoukiou:


mvollmer:
Some manual testing notes:

- The STP checkbox in "Add bridge" doesn't work right. Clicking reveals the STP options but it is not checked. Clicking again does nothing.
- Changing the ID in "Add VLAN" doesn't update the "Name" field.

Also on "main":

- Renaming a team, bond, or bridge makes it lose its members 
- Bridge row is shown as "0edit" (see screenshot below)
- Renaming vlan results in "Unexpected error", "vlan1 failed to create resources: Failed to create VLAN interface 'vlan1' for 'vlan1': File exists", but renaming succeeds apparently, interface is inactive afterwards.
- Changing the parent of a vlan always changes the name as well

image

mvollmer:
Out of scope: Runtime detection of team support should now be possible via D-Bus. We should check whether that is really true.

garrett:
> We used to have a warning icon in the "Connection will be lost" dialog

The warning icon should exist in the title area, as per PF guidelines:

https://www.patternfly.org/v4/components/modal/design-guidelines/#confirm-a-destructive-action

KKoukiou:
> The "Apply" buttons should be "Save".
> 
> Reason: "Save" always means "make the changes and close this dialog", whereas "Apply" sometimes does this too but more often means "Just make the changes now and keep the dialog open", famously in Windows, where most of our users are. Making the options "Save" and "Cancel" is much clearer.
> 
> (We probably have this problem elsewhere in Cockpit, and we probably inherited it here too, but it should be fixed.)
> 
> FWIW: PatternFly has "Save" in terminology yet has no "Apply". https://www.patternfly.org/v4/ux-writing/terminology/

Thanks for explaining, lets fix this in another PR as it's like this on master. This was meant to be a pure rewrite, of course I needed to squeeze some design changes in. It's however already 25 commits, so any improvements extra could better wait for a followup I feel.

garrett:
The new layout works fine even in small screen sizes (although when scrollbars are hidden, some features will be missed, but what can we do about that, really?)
Screen Shot 2021-10-14 at 11 49 41

It has issues in mobile. Here's a very large mobile phone:

Screen Shot 2021-10-14 at 11 50 26

And here's a small mobile phone:

Screen Shot 2021-10-14 at 11 50 42
Screen Shot 2021-10-14 at 11 50 05

We could solve this in a follow-up, unless you feel like tackling it here too (especially if there's something obvious we could do in PF).

PR #16429, opened 2021-10-04
systemd: Rewrite realmd configuration in React and standard PF4

martinpitt:
The dialog is split into separate join and a leave components/modals
now, to make them easier to understand and maintain. For consistency
with other pages, split the `realms-op` dialog ID into
`realms-{join,leave}-dialog`.

Fix a few usability problems along the way, such as disabling the inputs
and Join button for missing/invalid inputs.

Don't implement the `.Cancel()` API, as this was never wired up in the
UI -- the cancel button gets disabled as soon as the operation starts
(and the tests explicitly test that).

Move the realmd specific CSS out of overview.scss into
overview-cards/realmd.scss.

----

This does *not* port the "Authentication" selector. That one is completely broken right now -- it never opens the menu, and once you hack the DOM in the inspector to show the selections, they are still not applied correctly. At least [one-time password](https://fedoraproject.org/wiki/QA:Testcase_realmd_join_otp?rd=QA:Testcase_Active_Directory_realmd_join_otp) fails with a "Cannot authenticate as Administrator@" error (i.e. it defaulted back to password/administrator). I will add this back as a follow-up, and write tests.

 - [x] Fix unaligned layout in leave dialog
 - [x] Fix join error details pre to grow tastefully
 - [x] ~Pressing Enter should behave like clicking Join~ -- no, we don't do that
 - [x] Add integration tests for disabled join button while fields are incomplete, and disabled inputs while joining
 - [x] fix [TestAD.testQualifiedUsers](https://logs.cockpit-project.org/logs/pull-16429-20211006-072021-a8fefc0a-rhel-8-6/log.html#284) on RHEL/CentOS 8
 - [x] Add naughty for leave crash on Ubuntu: https://github.com/cockpit-project/bots/pull/2486

## Join dialog

The error details don't resize at all with the surrounding alert; missing spacing between alert and buttons; non-functional Authentication select; broken `<pre>` layout, and the `<pre>` has ugly horizontal scrolling:

On main:
realm-join-error
realm-join-error-details

On this PR: All the above issues are fixed:

Screen Shot 2021-10-06 at 09 24 54

Screen Shot 2021-10-06 at 14 26 03

## Leave dialog (expanded)

On main:

realm-info-main

On this PR:

Screen Shot 2021-10-07 at 10 06 21


martinpitt:
This is now functionally complete, except for pressing Enter to DTRT in the join dialog (see TODO item). Now I'll need some help with some layout/CSS issues (again, see TODO)

martinpitt:
@garrett: With Katerina's and Matej's help I fixed all outstanding layout/UX issues that I noticed. But as our last word on design, do you want to have a final look? I updated the main vs. this PR screenshots in the description. Of course you can also test it interactively:
```diff
--- test/verify/check-system-realms
+++ test/verify/check-system-realms
@@ -111,6 +112,7 @@ class CommonTests:
                 b.wait_val(self.op_address, "cockpit.lan")
 
         # Join cockpit.lan
+        sit()
         b.click(self.domain_sel)
         b.wait_popup("realms-join-dialog")
         set_address()
```

and then

    test/verify/check-system-realms TestAD.testQualifiedUsers -stv

garrett (changes requested):
It's definitely a lot better. Thanks!

I noticed the padding is not correct, and inline alerts should not have such long titles.

It should look like this in PF4, with proper spacing and using a title:

Screenshot 2021-10-06 at 16-58-47 PatternFly 4 • Alert

martinpitt:
I investigated the [leave failure on Ubuntu](https://logs.cockpit-project.org/logs/pull-16429-20211006-124217-64832238-ubuntu-stable/log.html#288). This is a long-standing bug in the Ubuntu package. It only surfaces now because the old code entirely ignored errors from the `.Deconfigure()` call:

    call = cleanup_ws_credentials().then(function() { realm.Deconfigure(options) });

(missing `return`). I'll file an Ubuntu bug and naughty it, as I really don't want to keep that bug.

martinpitt:
https://github.com/cockpit-project/bots/pull/2496 landed, retrying.

PR #16430, opened 2021-10-04
.github: run npm-update workflow with deploy key

allisonkarlitskaya:
Drop the final use of COCKPITUOUS_TOKEN from the cockpit repo and use
the `DEPLOY_KEY` in the `self` environment instead.  The PR opening gets
done with the scope-constrained GitHub Actions token.

Note: this is going to break the `/npm install` that `bots/npm-update` writes into the PR: it will be ignored, since it comes from a GitHub Actions token.  The agreed upon plan to sort that out involves triggering the `npm-install` workflow via `workflow_run` on the `npm-update` workflow and scanning the open PRs to find the one that was just created, and treat the `/npm install` comment there as the one that triggered the rebuild.

### DIFF
--- a/.github/workflows/npm-update.yml
+++ b/.github/workflows/npm-update.yml
@@ -6,27 +6,26 @@ on:
   workflow_dispatch:
 jobs:
   npm-update:
+    environment: self
+    permissions:
+      pull-requests: write
     runs-on: ubuntu-20.04
     steps:
       - name: Set up dependencies
         run: |
           sudo apt update
-          sudo apt install -y npm make
-
-      - name: Set up configuration and secrets
-        run: |
-          printf '[user]\n\tname = Cockpit Project\n\temail=cockpituous@gmail.com\n' > ~/.gitconfig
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
That's going to be a bit confusing in this intermediate state, as the bot will still produce an `/npm install` comment which doesn't do anything. But right now, cockpit npm-update PRs already require human action to actually push the updated gitlink. Ideally this should happen all from one workflow, with several jobs (or containerization) for the privilege separation. Then we wouldn't need all this "bots tell the other bots some magic comment" indirection. But of course not duplicating code is important, too.

Scanning open PRs sounds hackish, but `on: workflow` events normally have at least some minimal information about the event that triggered the workflow -- does that not include the issue number, or at least a SHA?

Either way, none of these problems affect cockpit-{podman,machines}, as they don't have a node_modules cache. I think with our current arms cabinet we can fully eliminate COCKPITUOUS_TOKEN from these two without any regression.

Thanks!

martinpitt (approved):
No-change force-push to "enable" no-test.

PR #16431, opened 2021-10-05
storage: Improve action summary for dialogs

mvollmer:
New pixels: https://github.com/cockpit-project/pixel-test-reference/compare/792d490ff3cfc7dedf9c1dddb0020c6d44c419ec..1534f778d70a5a75d2dd3167e6c6334568ae0763

Demo: https://www.youtube.com/watch?v=yG_uFDtaRYM

### DIFF
--- a/pkg/storaged/crypto-keyslots.jsx
+++ b/pkg/storaged/crypto-keyslots.jsx
@@ -224,17 +224,6 @@ export function get_existing_passphrase(block, just_type) {
     });
 }
 
-export function get_existing_passphrase_for_dialog(dlg, block, just_type) {
-    const prom = get_existing_passphrase(block, just_type).then(passphrase => {
-        if (!passphrase)
-            dlg.set_values({ needs_explicit_passphrase: true });
-        return passphrase;
-    });
-
-    dlg.run(_("Unlocking disk..."), prom);
-    return prom;
-}
-
 export function request_passphrase_on_error_handler(dlg, vals, recovered_passphrase, block) {
     return function (error) {
         if (vals.passphrase === undefined) {
@@ -246,6 +235,21 @@ export function request_passphrase_on_error_handler(dlg, vals, recovered_passphr
                     }));
         } else
             return Promise.reject(error);
+    };
+}
+
+export function init_existing_passphrase(block, just_type, callback) {
+    return {
+        title: _("Unlocking disk"),
+        func: dlg => {
+            return get_existing_passphrase(block, just_type).then(passphrase => {
+                if (!passphrase)
+                    dlg.set_values({ needs_explicit_passphrase: true });
+                if (callback)
+                    callback(passphrase);
+                return passphrase;
+            });
+        }
     };
 }
 
@@ -272,7 +276,7 @@ function validate_url(url) {
 function add_dialog(client, block) {
     let recovered_passphrase;
 
-    const dlg = dialog_open({
+    dialog_open({
         Title: _("Add key"),
         Fields: [
             SelectOneRadio("type", _("Key source"),
@@ -319,10 +323,11 @@ function add_dialog(client, block) {
                     });
                 }
             }
-        }
+        },
+        Inits: [
+            init_existing_passphrase(block, false, pp => { recovered_passphrase = pp })
+        ]
     });
-
-    get_existing_passphrase_for_dialog(dlg, block).then(pp => { recovered_passphrase = pp });
 }
 
 function edit_passphrase_dialog(block, key) {
@@ -347,7 +352,7 @@ function edit_passphrase_dialog(block, key) {
 function edit_clevis_dialog(client, block, key) {
     let recovered_passphrase;
 
-    const dlg = dialog_open({
+    dialog_open({
         Title: _("Edit Tang keyserver"),
         Fields: [
             TextInput("tang_url", _("Keyserver address"),
@@ -364,10 +369,11 @@ function edit_clevis_dialog(client, block, key) {
                     edit_tang_adv(client, block, key, vals.tang_url, adv, existing_passphrase);
                 });
             }
-        }
+        },
+        Inits: [
+            init_existing_passphrase(block, false, pp => { recovered_passphrase = pp })
+        ]
     });
-
-    get_existing_passphrase_for_dialog(dlg, block).then(pp => { recovered_passphrase = pp });
 }
 
 function edit_tang_adv(client, block, key, url, adv, passphrase) {
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -231,13 +231,16 @@ import {
     Spinner, Split,
     TextInput as TextInputPF4,
     Popover,
-    HelperText, HelperTextItem
+    HelperText, HelperTextItem,
+    List, ListItem
 } from "@patternfly/react-core";
 import { ExclamationTriangleIcon, InfoIcon, HelpIcon } from "@patternfly/react-icons";
 
 import { show_modal_dialog, apply_modal_dialog } from "cockpit-components-dialog.jsx";
+import { ListingTable } from "cockpit-components-table.jsx";
 
-import { fmt_size, block_name, format_size_and_text } from "./utils.js";
+import { fmt_size, block_name, format_size_and_text, format_delay, for_each_async, decode_filename } from "./utils.js";
+import { fmt_to_fragments } from "utils.jsx";
 import client from "./client.js";
 
 import "@patternfly/patternfly/components/HelperText/helper-text.css";
@@ -484,10 +487,30 @@ export const dialog_open = (def) => {
             });
         },
 
+        set_attribute: (name, value) => {
+            def[name] = value;
+            update(null, null);
+        },
+
+        add_danger: (danger) => {
+            def.Action.Danger = <>{def.Action.Danger} {danger}</>;
+            update(null, null);
+        },
+
         close: () => {
             dlg.footerProps.dialog_done();
         }
     };
+
+    for_each_async(def.Inits || [],
+                   init => {
+                       if (init) {
+                           const promise = init.func(self);
+                           self.run(init.title, promise);
+                           return promise;
+                       } else
+                           return Promise.resolve();
+                   });
 
     return self;
 };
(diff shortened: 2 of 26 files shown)


mvollmer:
From @garrett:

So I've revised the mockup and split it into multiple conditions, as I outlined above. Hopefully this helps.

I've added the two helpertext messages. We might need to override to as inline variant. If it doesn't support it directly (such as setting 'component' to 'span', which isn't shown as supported, but might work), we could use one of the PatternFly helpers (and attach the appropriate "className" in the JSX): https://www.patternfly.org/v4/utilities/display

To summarize the differences:

- Only show the changes when there are > 1 changes
- Show the device currently in use message when it's currently in use
- Show the erases all data message when there's existing content
- Currently in use and erases data helper texts messages will usually both be shown in most cases (it's possible to have unmounted disks that have data, so they'd show the data message even if not mounted)
- I changed the labels
- I revised the disk-based mockups to be "initialize" (format is for partitions, initialize for disks)

## Single, unused 

image

## Single, used

image

## Multiple changes, unused

(This doesn't happen, right? :wink:)

## Multiple changes, used

image

garrett:
How's this for unmount when there's a process blocking it?

unmount-force

pkg/storaged/dialog.jsx
```diff
@@ -1032,18 +1035,32 @@ export const TeardownMessage = (usage) => {
...
 
-    add_usage_message(parts, teardown.MDRaidMembers,
-                      _("This device is currently used for RAID devices. Proceeding will remove it from its RAID devices."),
+    add_usage_message(uses, teardown.MDRaidMembers,
+                      _("remove"),
                       "Name", "MDRaid");
 
+    if (uses.length > 0)
+        parts.push(
+            <div>
```
garrett, line 1053: Why `<div>`? Should this be `<>` (shorthand for `<React.Fragement>` https://reactjs.org/docs/fragments.html)

mvollmer:
> We might want to say that services and processes will be stopped before formatting or something too. I didn't add that in the mockup (ran out of time), but we probably want to add it in.

Yes, I think that is important.

mvollmer:
@garrett, I am not sure how to represent services that keep a mount point busy during unmount.

For teardown (formatting, deleting, ...), we have two lists in the popover:

image

How should we do this for the Unmount dialog?  The current code in this PR just puts services in the same table as processes:

image

garrett:
Yeah, I guess this is inconsistent. Part of the issue is that the amount of information in the table is a lot more than in the simple list. I wonder if we should adjust what's shown for both.

Is the user a bit odd for the service? Would the process ever have a description, or is that only for services?

Also, the warning in the dialog of the first screenshot (of your last comment) shouldn't be "Using processes will be..." but perhaps "Associated processes..." or something like that.

mvollmer:
> Is the user a bit odd for the service? Would the process ever have a description, or is that only for services?

Processes don't have descriptions right now, but maybe we can come up with something useful.  Maybe the full command line?

Services mostly runs as root, but you can configure them to run as some other user. But those are non "real" user accounts that would log in, but system accounts like "nobody", "apache", "mail", or "cockpit-ws-instance".  So, yeah, that kind of information is probably better shown on the details page for the actual service.

mvollmer:
> but maybe we can come up with something useful. Maybe the full command line?

Oh, the "Process" column is already the full command line.

mvollmer:
> Also, the warning in the dialog of the first screenshot (of your last comment) shouldn't be "Using processes will be..." but perhaps "Associated processes..." or something like that.

Yes.  While the information is gathered, we say "Checking related processes...."  So "Related processes will be killed."?  Do we need to mention services?  "Related processes and services will be killed."

garrett:
> Yes. While the information is gathered, we say "Checking related processes...." So "Related processes will be forcefully stopped."? Do we need to mention services? "Related processes and services will be forcefully stopped."

Yeah, that makes sense, as long as services exist, right?

mvollmer:
> How should we do this for the Unmount dialog? The current code in this PR just puts services in the same table as processes:

I changed that to two tables:

image

garrett:
Having two tables is messy... but it's especially messy if they don't line up. They would be better like this:

image

This looks less chaotic.

I guess we could specify the size of the cells width of (100% / number of items) = 25%... if it's implemented as an HTML table, it should have some give so the content should be flexible, with 25% just being the recommended default.

To more clearly illustrate what I'm talking about, here's the modified screenshot with guidelines I used to shove pixels around:

Screenshot from 2021-12-14 18-06-47

mvollmer:
> Having two tables is messy...

Yeah... I wonder how often this happens? (both individual processes and services)

mvollmer:
> Having two tables is messy... but it's especially messy if they don't line up. They would be better like this:

Looks like this now, with 25% width for each column:

image

garrett (dismissed):
I think this is good to land.

The sessions versus services thing is probably a corner case, but the tables look a lot better with the column alignment adjustment.

Whatever issues that may exist (I haven't checked mobile lately, for example), can be addressed in follow-ups. I don't want to block this PR from a design perspective. This PR is a big improvement.

martinpitt:
Moving to draft to clean up the PR review list. Tests still need updating, and there's a TODO for more pixel tests.

test/common/testlib.py
```diff
@@ -783,18 +783,18 @@ def assert_pixels(self, selector, key, ignore=[]):
...
+                                       r['x'] - rect['x'] + r['width'],
+                                       r['y'] - rect['y'] + r['height']),
+                            self.call_js_func('ph_selector_clips', sels)))
 
         reference_dir = os.path.join(TEST_DIR, 'reference')
         if not os.path.exists(os.path.join(reference_dir, '.git')):
             subprocess.check_call([f'{TEST_DIR}/common/pixel-tests', 'pull'])
 
-        ignore_rects = list(map(relative_clip, map(lambda item: selector + " " + item, ignore)))
+        ignore_rects = relative_clips(list(map(lambda item: selector + " " + item, ignore)))
```
martinpitt, line 797: You shouldn't need the `list()` here.

pkg/storaged/crypto-keyslots.jsx
```diff
@@ -136,6 +136,7 @@ function clevis_unlock(block) {
 }
 
 export function unlock_with_type(client, block, passphrase, passphrase_type) {
+    console.log("UN", passphrase, passphrase_type);
```
martinpitt, line 139: Debugging leftover, and a blocker as it logs passphrases.

pkg/storaged/dialog.jsx
```diff
@@ -998,129 +1021,229 @@ export const SizeSlider = (tag, title, options) => {
...
-    if (has_sessions && has_services)
-        add_para(parts, _("The filesystem is in use by login sessions and system services. Proceeding will stop these."));
-    else if (has_sessions)
-        add_para(parts, _("The filesystem is in use by login sessions. Proceeding will stop these."));
-    else if (has_services)
-        add_para(parts, _("The filesystem is in use by system services. Proceeding will stop these."));
+    const service_rows = users.filter(u => u.unit).map(u => {
+        return {
+            columns: [
+                u.unit.replace(/\.service$/, "").substr(0, 40),
```
martinpitt, line 1183: It's plausible to have services longer than 40 characters. Would it be possible to limit these via CSS instead? max-width and `text-overflow: ellipsis` or so? (and thus showing the full name in a tooltip)

Not a blocker though, fine to do this as a follow-up.

pkg/storaged/dialog.jsx
```diff
@@ -998,129 +1021,229 @@ export const SizeSlider = (tag, title, options) => {
...
 };
 
 export function teardown_and_apply_title(usage, plain_title, unmount_title, remove_title) {
-    if (usage.Teardown && usage.Teardown.Mounts)
-        return unmount_title;
-    else if (usage.Teardown && (usage.Teardown.PhysicalVolumes || usage.Teardown.MDRaidMembers))
-        return remove_title;
-    else
-        return plain_title;
+    return plain_title; // XXX - correct?
```
martinpitt, line 1248: Unresolved todo

test/verify/check-storage-mounting
```diff
@@ -69,9 +69,27 @@ class TestStorageMounting(StorageCase):
...
+Description=Test Service
+
+[Service]
+WorkingDirectory={mount_point_foo}
+ExecStart=/usr/bin/sleep infinity
+""")
+        m.execute("systemctl start keep-mnt-busy")
+
         self.content_dropdown_action(1, "Unmount")
+        b.wait_in_text("#dialog", str(sleep_pid))
```
martinpitt, line 86: Please also assert that keep-mnt-busy service is mentioned in the dialog

martinpitt (changes requested):
Very cool, thanks! Structurally this looks fine, I just have a few nitpicks.

test/common/testlib.py
mvollmer: A "map iterator" can not be passed to json.dumps, which is used by `self.call_js_func('ph_selector_clips', sels)` to convert the arguments to JS.  We could do `self.call_js_func('ph_selector_clips', list(sels))` but then `relative_clips` would have the side effect of consuming its argument, which I find odd.

mvollmer: And just to flog this dead horse some more, consider this:

```
>>> json.dumps([1, 2, 3])
'[1, 2, 3]'
>>> json.dumps(map(lambda x: 2*x, [1, 2, 3]))
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib64/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type map is not JSON serializable
```

I find this bizarre, from a "Functional Programming" point of view.  I would expect `map` to only return an iterator when mapping over iterators. If this is supposed to emulate lazy evaluation, then an iterator should not be consumed when iterating over it.

Anyway, `list(map(...))` it is from now on, and always! :-)

pkg/storaged/crypto-keyslots.jsx
mvollmer: Ouch.

pkg/storaged/dialog.jsx
mvollmer: Fxied.

pkg/storaged/dialog.jsx
mvollmer: I give it a try...

test/verify/check-storage-mounting
mvollmer: Done.

test/common/testlib.py
martinpitt: Hehe, thanks for the research!

pkg/storaged/dialog.jsx
mvollmer: Oh, Patternfly has nice "modifiers" for Table cells.  `modifier="truncate"` would show ellipses, but let's use "breakWord" so that nothing gets hidden but long words will be broken in their middle.  (And let's use "nowrap" for the "Runtime" column.)

Super long command line (one of Cockpits embedded python scripts that we now load via the command line):

image

Super long service name and description (cockpits default https instance):

image

When we force the table columns to be 25/25/25/25 percent:

image
image

mvollmer: Fixed.

martinpitt: Superb, thanks!

martinpitt (approved):

martinpitt:
@mvollmer Thanks! Can you please still add a release note?

PR #16441, opened 2021-10-07
storaged: this page does not use jquery

KKoukiou:
Stop importing it.


martinpitt:
Nice, that seems to work! :+1: @KKoukiou can you please drop no-test, push a rebase (for packit), and run the whole test saga?

PR #16443, opened 2021-10-08
test: Add pixel tests for realm join/leave, fix test flake

martinpitt:
Plus a small code cleanup that I forgot in PR #16429.

### DIFF
--- a/test/verify/check-system-realms
+++ b/test/verify/check-system-realms
@@ -118,6 +118,7 @@ class CommonTests:
         # admin gets auto-detected
         b.wait_val(self.op_admin, self.admin_user)
         b.set_input_text(self.op_admin_password, self.admin_password)
+        b.assert_pixels("#realms-join-dialog", "realm-join")
         b.click(f"#realms-join-dialog button{self.primary_btn_class}")
         # running operation cannot be cancelled any more
         b.wait_visible("#realms-join-dialog button.pf-m-link:disabled")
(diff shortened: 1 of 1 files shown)


KKoukiou:
> There is a [very subtle diff](https://logs.cockpit-project.org/logs/pull-16443-20211008-145138-e7ac22b5-fedora-34-mobile/pixeldiff.html#TestAD-testQualifiedUsers-realm-leave-mobile) in the caret expander. Not sure yet what to do about that -- @mvollmer , @KKoukiou , do you have some guidance/experience here?

Right, as the expander has animation it's hard to capture it correct. I already ignore it in cockpit-machines. https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-disks#L323

martinpitt:
[this failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16443-20211009-074411-f3fafb3a-fedora-34-devel/log.html#284) is unrelated to the pixel tests, but a regression from my react PR. I added a fix here.

mvollmer:
Shouldn't the triangle point downwards here?

https://raw.githubusercontent.com/cockpit-project/pixel-test-reference/605e2236c4d61ed0a5c384aaec4fe5dfb13b9d64/TestAD-testQualifiedUsers-realm-leave-mobile-pixels.png

PR #16451, opened 2021-10-12
Fix building with npm 7, drop obsolete `npm run eslint` hack

martinpitt:
Tested in a Fedora 35 toolbox. Similar to cockpit-project/starter-kit#501, but requires a little more work.

With current main, pretty much all pages fail to build like this:
```
ERROR in ./node_modules/@patternfly/react-core/dist/esm/components/FileUpload/FileUpload.js 3:0-38
Module not found: Error: Can't resolve 'react-dropzone' in '/var/home/martin/upstream/cockpit/node_modules/@patternfly/react-core/dist/esm/components/FileUpload'
resolve 'react-dropzone' in '/var/home/martin/upstream/cockpit/node_modules/@patternfly/react-core/dist/esm/components/FileUpload'
  Parsed request is a module
  using description file: /var/home/martin/upstream/cockpit/node_modules/@patternfly/react-core/package.json (relative path: ./dist/esm/components/FileUpload)
```

For most of us who are on F35 already (@marusak and @garrett at least?) this was not yet visible because we usually just download/use the node_modules cache, instead of running `npm install` locally.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -254,7 +254,7 @@ const CockpitPoPlugin = require("./pkg/lib/cockpit-po-plugin");
 const srcdir = process.env.SRCDIR || __dirname;
 const builddir = process.env.BUILDDIR || __dirname;
 const libdir = path.resolve(srcdir, "pkg" + path.sep + "lib");
-const nodedir = path.resolve(srcdir, "node_modules");
+const nodedir = path.relative(process.cwd(), path.resolve(srcdir, "node_modules"));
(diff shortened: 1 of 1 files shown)


martinpitt:
@marusak: I'd just ignore the protection checks -- there are two commits here which change package.json, and I didn't want to rebuild it twice. The first one has zero impact on node_modules/ (it just changes scripts/), it just triggers the overzealous check. But let me know if you want me to split this.

PR #16452, opened 2021-10-12
[r79] Fix build with npm 7

martinpitt:


martinpitt:
Got it working now. I also avoided the relative path stuff, I've seen in c-machines pixel tests that these can have an impact.

PR #16453, opened 2021-10-12
tls: Fix SELinux policy and documentation for certmonger certificates, fix auto-renewal for FreeIPA issued ws certificates

### LINKED ISSUE #16450: documented certmonger integration does not work
Reported [on the mailing list](https://lists.fedorahosted.org/archives/list/cockpit-devel@lists.fedorahosted.org/thread/I5ZCA72PY5G7ZINVTZUJCT6PBZBXQXI4/):
```
$ getcert request -f ${CERT_FILE} -k ${KEY_FILE} -D $(hostname --fqdn) -C \
"sed -n w/etc/cockpit/ws-certs.d/50-from-certmonger.cert ${CERT_FILE} \
${KEY_FILE}"
```
reportedly failed with
```
type=AVC msg=audit(1633548839.586:8895): avc:  denied  { create } for
  pid=836449 comm="sed" name="50-from-certmonger.cert"
scontext=system_u:
system_r:certmonger_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
```

/etc/cockpit/ws-certs.d should be `certs_t` instead of `etc_t` to make this work.

Let's fix the documentation at least, but possibly also the default permissions of the certs dir. We don't test this in CI, let's add a test to make sure this keeps working.

After that, the file permissions need updating so that cockpit-ws can read that. This is a separate issue though, we should generally stop relying on file perms and just copy the cert to /run in a pre-start command.

martinpitt:
Certmonger is not allowed to write into /etc/cockpit/ws-certs.d/, as
it currently has the default type `etc_t`, but certmonger can only write
into `cert_t` directories. Change our SELinux policy accordingly.

Update the documentation to drop the seddery, cockpit-tls has supported
a split certificate for some time now. The old command generated a file
which was inaccessible to cockpit-tls, add the missing options to
configure proper permissions.

Add an integration test to make sure that this actually works.

Fixes #16450

## FreeIPA-issued webserver certificates get auto-renewed

When you join a [FreeIPA](https://www.freeipa.org) realm with Cockpit, it automatically requests a TLS certificate for its webserver from the FreeIPA domain controller. Browsers usually trust these certificates much more than the default self-signed ones. Cockpit now calls [ipa-getcert](https://linux.die.net/man/1/ipa-getcert) in such a way that the generated `/etc/cockpit/ws-certs.d/10-ipa.{cert,key}` certificate gets renewed automatically.

### DIFF
--- a/src/tls/cockpit-certificate-helper.in
+++ b/src/tls/cockpit-certificate-helper.in
@@ -108,23 +108,30 @@ cmd_ipa_request() {
     klist -k "${KEYTAB}" | grep -qF "${SERVICE}" || \
         ipa-getkeytab -p "HTTP/${HOST}" -k "${KEYTAB}"
 
-    # request the certificate (into the working directory)
-    ipa-getcert request -f 10-ipa.cert -k 10-ipa.key -K "HTTP/${HOST}" -w -v
-
-    # install it into /etc
-    install_cert 10-ipa.cert
-    install_key 10-ipa.key
+    # request the certificate and put it into our certificate directory
+    if ! selinuxenabled || [ "$(stat -c %C /etc/cockpit/ws-certs.d/ | cut -f3 -d:)" = cert_t ]; then
+        # the prefered way is to let certmonger write the certificate directly, so that auto-refresh works
+        ipa-getcert request -f "${COCKPIT_WS_CERTS_D}/10-ipa.cert" -k "${COCKPIT_WS_CERTS_D}/10-ipa.key" -K "HTTP/${HOST}" -m 640 -o root:${COCKPIT_GROUP:-root} -M 644 -w -v
+    else
+        # ... but that does not work on RHEL 8 or customized SELinux policies, so copy it ourselves
+        # request the certificate (into the working directory)
+        ipa-getcert request -f 10-ipa.cert -k 10-ipa.key -K "HTTP/${HOST}" -w -v
+        # install it into /etc
+        install_cert 10-ipa.cert
+        install_key 10-ipa.key
+    fi
 }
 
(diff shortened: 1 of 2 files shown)


jh23453:
I've tried the change on Debian 11/bookworm with cockpit from backports:

ii  cockpit        254-1~bpo10+1 all          Web Console for Linux servers

I've used the following command:
`ipa-getcert request --wait --keyfile=/etc/cockpit/ws-certs.d/50-certmonger.key --key-owner=root:cockpit-ws --key-perms=600 --certfile=/etc/cockpit/ws-certs.d/50-certmonger.crt --cert-owner=root --cert-perms=644 -N CN=echidna.example.org -D echidna.example.org -K cockpit/echidna.example.org
`
I got these files:
`root@echidna:/etc/cockpit/ws-certs.d# ls -l
insgesamt 12
-rw-r----- 1 root cockpit-ws 2914  5. Apr 2016  0-self-signed.cert
-rw-r--r-- 1 root root       1684 12. Okt 17:42 50-certmonger.cert
-rw------- 1 root cockpit-ws 1704 12. Okt 17:42 50-certmonger.key
`
Cockpit fails with:
`Okt 12 17:45:33 echidna cockpit-tls[2434996]: cockpit-tls: Invalid server certificate+key file /etc/cockpit/ws-certs.d/50-certmonger.cert: Required key not available
`
Any ideas what might be missing on my system?

jh23453:
> > --key-owner=root:cockpit-ws --key-perms=600
> 
> This needs to be 0640, so that the group can actually read the file.

That did the trick - I really had the permissions mixed up!
Thanks - I really like the change and hope to see it in a release.

Issue #16435, opened 2021-10-06
metrics: Optimize for lower resolution laptop/desktop screens


garrett:
Two big issues:

1. The cards start to overflow horizontally
2. Historical metrics cannot be seen without scrolling

![Screenshot from 2021-10-06 17-13-20](https://user-images.githubusercontent.com/10246/136236998-39680a1a-245a-4164-b30b-ad93f5e4b18c.png)
(I've blanked out anything that might possibly be personally identifying.)

As this was via screen sharing, I don't know the actual screen resolution. But it looks like it might be around 73% of this image, which would be 1280 x 720.

Here's the same screenshot scaled 1:1 (click to view at size):

![1280x720](https://user-images.githubusercontent.com/10246/136239877-25b645a3-1877-4fc3-9c7e-1492991a1df6.png)

Usable document real estate is:

- 1280x578 for all of Cockpit, after removing the browser UI (tabs, nav, bookmarks) and the Windows task bar.
- 1026x514 for the usable page area, after removing the above, Chrome's scrollbars on Windows, Cockpit's header, Cockpit's navigation 

And I guess we need to make sure to test the following resolutions:

- 1366×768 (second most common resolution in 2021, according to some random website that was the first search hit)
- 1280x720 (seventh most common resolution from the same list)

Corporate environments would hopefully have higher resolution screens, but we can't just wave our hands at that, as a lot of people are still working from home in the pandemic and a lot will be working from home after the pandemic too.

martinpitt:
@garrett: What we can do easily is to do an one-off run with those resolutions, and we'll get failed pixel tests (with new images) for these two resolutions. We can't easily do "all of cockpit", as one needs to define at which point they need to happen.

garrett:
@martinpitt: That would be a good start. I guess combining that with manual testing and asking everyone to check these sizes (as we go forward) and file issues would be helpful.

garrett:
You can add the tiny laptop resolution of 1280 x 578 (to simulate the reduction of Chrome UI, bookmarks, and Windows taskbar) in Firefox's mobile mode and add it as a preset:

![Screenshot from 2021-10-12 14-56-38](https://user-images.githubusercontent.com/10246/136960434-ec7363f2-b458-4c4a-8ea1-62e357488336.png)

Here's what the metrics page looks like:

[...]

PR #16465, opened 2021-10-13
metrics: Make better use of space

garrett:
Addressing https://github.com/cockpit-project/cockpit/issues/16435

Currently a work in progress. I'm trying to keep each change separate, at least for now, in case any are controversial.

The goal is to optimize in several places to make the page better overall on smaller screens (and more optimal on larger screens too).

### DIFF
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -170,6 +170,15 @@ html:not(.index-page) body {
     color: var(--pf-global--danger-color--100);
 }
 
+// Action buttons in headers add extra space. Offset that with a negative margin
+// to compensate, so headings are always the same height regardless of action
+// buttons or not.
+.pf-c-page__main-breadcrumb .pf-c-button {
+    --offset: calc(-1 * var(--pf-global--spacer--sm));
+    margin-top: var(--offset);
+    margin-bottom: var(--offset);
+}
+
 // Drop side padding in mobile mode,
 // intended mainly for PF PageSection elements (pf-c-page__main-section).
 // It's similar to adding padding={{ default: 'noPadding', sm: 'padding' }},
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1421,7 +1421,7 @@ export const Application = () => {
     const [needsLogout, setNeedsLogout] = useState(false);
 
     return <Page additionalGroupedContent={
-        <PageSection id="metrics-header-section" variant={PageSectionVariants.light}>
+        <PageSection id="metrics-header-section" variant={PageSectionVariants.light} type='breadcrumb'>
             <Flex>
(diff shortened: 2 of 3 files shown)


marusak:
It seems all good to me, thanks! Are you still working on this?

garrett:
Yep, I'm still working on it. Oh! I even have a few local commits that I didn't push related to the top header breadcrumb bar.

garrett:
I did some number crunching on some semi-reliable stats, and came up with a visualization of the likely screen availabilty of our users (as we don't track, this is pretty wild speculation, but it's based on assumptions and available statistics from non-Cockpit sources):

screen-resolution-heatmap

View this at 1:1 (open in new tab and click to expand to 1:1) to see how it compares to the screen you're using. This is definitely an empathy check to feel how little our users can see at one time versus what we can see at one time. Quite a difference, yeah?

This is an auto-levels-adjusted overlay of all the screen sizes. I think the vertical oddness is from iPad usage? Otherwise obvious mobile sizes have been excluded. This visualization is a useful anonymized way of showing the stats. It aligns with what we're seeing elsewhere, although we may have users with slightly higher resolutions than other sites (but not by much... certainly not enough to count).

It's aligned to the top-left. This isn't a direct comparison of resolutions, but a visualization of available screen pixel real estate available to our users. Also, this doesn't take into account browser UI (tabs, bookmarks, etc.) or persistent operating system UI (Windows taskbar, GNOME top bar, macOS menubar, etc.)... so the actual content area is lower.

The darkest landscape rectangle is 1980x720 ("720p" in media terms).

The top 7 most popular screen resolutions are, in order:

Resolution | Marketing term
-- | --
1920 x 1080 | FHD / 1080p
1600 x 864 |  ??? probably some popular laptops and/or screens ???
1366 x 768 | FWXGA
1280 x 720 | HD / WXGA / 720p
2560 x 1440 | (W)QHD
1920 x 1200 | WUXGA
1680 x 1050 | WSXGA+

This is similar to other resolution charts for 2021 on the Internet, but a little bit better resolution wise.

So, sadly, we need to make sure 1280 x 720 is properly supported everywhere, as it's the 4th most common resolution. (And third place is not that far off!) 

But even these larger resolutions are useful, as they are target resolutions to check from time to time to make sure everyone sees optimized content when possible.

garrett:
Rebased

garrett:
Note: When the flex workaround (#17043) lands, it'll fix up the button in the header, so it'll look like this:

Screen Shot 2022-02-24 at 10 52 17

Otherwise, PF's `<Flex>` is problematic for wrapped items. But this flex change affects all of Cockpit (in a good way), not just Metrics.

martinpitt (approved):
Thanks! I tested this in various situations, and it behaves better now. Especially for the sticky historical header!

PR #16470, opened 2021-10-14
Port credentials dialogs to React

KKoukiou:
main:
image

this PR:
image


martinpitt:
[this failure](https://logs.cockpit-project.org/logs/pull-16470-20211014-154638-19ae429e-fedora-coreos/log.html#294) looks relevant? Supposedly coreos is a special snowflake here as it always authenticates through ssh?

KKoukiou:
> Thanks! This generally works very well now. I updated the screenshot in the description. The only thing that's broken now is the "x" close button in the modal, it's apparently not wired up? That should be simple to fix?

Fixed. Thanks for noticing!

PR #16471, opened 2021-10-14
test: refactor chromium/firefox browser handling

ptoscano:
The test class & module `CDP` have various code bits to handle the differences between chromium and firefox browsers scattered in different places. To better isolate these differences, create a small `Browser` class hierarchy with class variables and methods for the browser-specific behaviours, such as: executable names, CDP .js files, arguments.

Additionally, `shutil.which` is used instead of spawning `which` executables.
Additionally, the following changes were done:
- use shutil.which instead of spawning which executables
- use `subprocess.check_call()` instead of `subprocess.Popen()` for a simple process invocation
- add "chromium-freeworld" as executable for chromium

The only browser-specific code left in `CDP` is when injecting scripts on browser startup; isolating that properly would require a bit more of work, and it is not worth yet.

### DIFF
--- a/test/common/cdp.py
+++ b/test/common/cdp.py
@@ -63,7 +63,7 @@ class Browser(abc.ABC):
 
 class Chromium(Browser):
     NAME = "chromium"
-    EXECUTABLES = ["chromium-browser", "chromium", "google-chrome"]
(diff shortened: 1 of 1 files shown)


marusak:
A load of ` SystemError: chromium is not installed`, see for example [here](https://logs.cockpit-project.org/logs/pull-16471-20211014-111939-65fc5ed1-rhel-8-6/log.html)

PR #16473, opened 2021-10-14
networking: Remove duplicated CSS imports

garrett:
When debugging the networking page, I kept coming across a lot of duplicated CSS. I've found these lines to be the culprit. Removing them seems to fix it with no visual regressions I have been able to find.

(The remaining imports do have regressions when removed.)

This _may_ change when https://github.com/cockpit-project/cockpit/pull/16428/ lands.


marusak:
Closing as it is being done in https://github.com/cockpit-project/cockpit/pull/16428 and we don't need this that urgently, so lets not cause conflicts

PR #16475, opened 2021-10-14
all: Rename occurances of "Apply" to "Save", fix racy TestNetworkingBasic pixel test

garrett:
"Apply" buttons should be "Save".

Reason: "Save" always means "make the changes and close this dialog", whereas "Apply" sometimes does this too but more often means "Just make the changes now and keep the dialog open", famously in Windows, where most of our users are. Making the options "Save" and "Cancel" is much clearer.

### DIFF
--- a/pkg/kdump/kdump-view.jsx
+++ b/pkg/kdump/kdump-view.jsx
@@ -181,7 +181,7 @@ class KdumpTargetBody extends React.Component {
  * kdumpActive       kdump service status
  * onSetServiceState called when the OnOff state is toggled (for kdumpActive), parameter: desired state
  * stateChanging     whether we're currently waiting for our last change to take effect
- * onApplySettings   called with current dialog settings when the user clicks apply
+ * onSaveSettings   called with current dialog settings when the user clicks Save
  * kdumpStatus       object as described in kdump-client
  * reservedMemory    memory reserved at boot time for kdump use
  * onCrashKernel     callback to crash the kernel via kdumpClient, expects a promise
@@ -279,11 +279,11 @@ export class KdumpPage extends React.Component {
         this.state.dialogObj.render();
     }
 
-    handleApplyClick() {
+    handleSaveClick() {
         return new Promise((resolve, reject) => {
-            this.props.onApplySettings(this.state.dialogSettings)
+            this.props.onSaveSettings(this.state.dialogSettings)
                     .then(resolve)
-                    .catch(error => reject(cockpit.format(_("Unable to apply settings: $0"), String(error))));
+                    .catch(error => reject(cockpit.format(_("Unable to save settings: $0"), String(error))));
         });
     }
 
@@ -356,8 +356,8 @@ export class KdumpPage extends React.Component {
         const footerProps = {
             actions: [
                 {
-                    clicked: this.handleApplyClick.bind(this),
-                    caption: _("Apply"),
+                    clicked: this.handleSaveClick.bind(this),
+                    caption: _("Save"),
                     style: 'primary',
                 },
             ],
--- a/pkg/kdump/kdump.js
+++ b/pkg/kdump/kdump.js
@@ -37,7 +37,7 @@ const initStore = function(rootElement) {
 
     dataStore.kdumpClient = new kdumpClient.KdumpClient();
 
-    dataStore.applySettings = settings =>
+    dataStore.saveSettings = settings =>
         dataStore.kdumpClient.validateSettings(settings)
                 .then(() => dataStore.kdumpClient.writeSettings(settings));
 
@@ -63,7 +63,7 @@ const initStore = function(rootElement) {
             stateChanging: dataStore.stateChanging,
             reservedMemory: dataStore.kdumpMemory,
             kdumpStatus: dataStore.kdumpStatus,
-            onApplySettings: dataStore.applySettings,
+            onSaveSettings: dataStore.saveSettings,
             onCrashKernel: dataStore.kdumpClient.crashKernel,
         }), rootElement);
     };
--- a/pkg/networkmanager/bond.jsx
+++ b/pkg/networkmanager/bond.jsx
@@ -28,7 +28,7 @@ import {
 } from '@patternfly/react-core';
 import { ExternalLinkSquareAltIcon, HelpIcon } from '@patternfly/react-icons';
 
-import { MacMenu, MemberInterfaceChoices, NetworkModal, Name, dialogApply } from './dialogs-common.jsx';
+import { MacMenu, MemberInterfaceChoices, NetworkModal, Name, dialogSave } from './dialogs-common.jsx';
 import { ModelContext } from './model-context.jsx';
 
 import { v4 as uuidv4 } from 'uuid';
@@ -110,7 +110,7 @@ export const BondDialog = ({ connection, dev, setIsOpen, settings }) => {
             }
         });
 
-        dialogApply({
+        dialogSave({
             model,
             dev,
             connection,
--- a/pkg/networkmanager/bridge.jsx
+++ b/pkg/networkmanager/bridge.jsx
@@ -26,7 +26,7 @@ import {
     TextInput,
 } from '@patternfly/react-core';
 
-import { MemberInterfaceChoices, NetworkModal, Name, dialogApply } from './dialogs-common.jsx';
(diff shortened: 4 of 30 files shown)


garrett:
This probably changes a little too much (most likely in the storage page and perhaps in the software updates page). I wanted to share it as-is as a draft to let the tests run and hit the existing pixel testing to see what happens.

Meanwhile, I'll manually review all the changes too.

garrett:
> Hmm, no strong option here. In these dialogs you make changes and then want to apply them. To me Apply seems like better word for that kind of change, but really cannot say if Save is better for user man_shrugging

Consistency is good. It's saving the changes.

But, yeah, this whole PR is kind of low priority and there might be unintended string changes.

mvollmer:
For Storage:
- "Save" for "Stored Pasphrase" makes sense, it just writes to a file in /etc/luks-leys.
- "Save" for "Encryption Options" makes sense, it just writes to /etc/crypttab.
- "Save" for the "Mount" dialog is not so clear to me. It will also unmount and re-mount the filesystem. Do we want to include changes to the actual run-time in "Save"?

In general, I would rather follow GNOME conventions than Windows conventions.

GNOME Disks has a "Ok" button when changing fstab, and a "Change" button when changing the filesystem label.  However, GNOME Disks doesn't feel very polished or up-to-date these days.

GNOME Settings uses "Apply" when changing network settings.

But then again, I couldn't find anything about dialogs in the GNOME HIG...

garrett:
> In general, I would rather follow GNOME conventions than Windows conventions.

This is about Cockpit and PatternFly conventions, not GNOME or Windows. That said, using Save is more like GNOME; Apply is more like Windows.

https://www.howtogeek.com/howto/15185/hey-you-stop-using-the-apply-button-and-just-click-ok-geek-rants/

"Apply" does not close a dialog, especially on Windows, whereas "Save" does, which is the point of renaming the buttons in this PR... as the actions we have in modals all intentionally close the dialog in Cockpit.

As a large amount of our users actually come from Windows, we should respect that on Windows, "Apply" does not close dialogs, so therefore, it should not be used on modals when actions would close the dialogs.

> But then again, I couldn't find anything about dialogs in the GNOME HIG...

GNOME usually uses Save or Update, not Apply. It's not called out explicitly in the HIG, but it is hinted at:
https://developer.gnome.org/hig/patterns/controls/buttons.html?highlight=save

---

As far as PatternFly is concerned, it has "Save" in the terminology (with its counterpart opposite of "Cancel" mentioned), but does not even define "Apply".

https://www.patternfly.org/v4/ux-writing/terminology/

---

> "Save" for the "Mount" dialog is not so clear to me. It will also unmount and re-mount the filesystem. Do we want to include changes to the actual run-time configuration in "Save"?

We could use "Update" (which is a bit generic and might be odd, given the context of the rest of the modals having save) or "Save and remount" (which is specific and consistent using the word "save", with a second action being included in the button).

At first, I thought "perhaps we should hyphenate re-mount", but it's a word in the MW dictionary. https://www.merriam-webster.com/dictionary/remount ... and it means:

> re·​mount | \ ˈrē-ˌmau̇nt \\ (noun)
> : a fresh horse to replace one no longer available

But it also means "to mount (something) again", which is more appropriate for Cockpit. 🏇

martinpitt:
> Reason: "Save" always means "make the changes and close this dialog", whereas "Apply" sometimes does this too but more often means "Just make the changes now and keep the dialog open

That's not really part of the definition of the words, right? Just common practice from Windows? E.g. "Save" in Word doesn't close the document either.

garrett:
> That's not really part of the definition of the words, right? Just common practice from Windows? E.g. "Save" in Word doesn't close the document either.

Correct, but you have to think about context and what people using Cockpit (and especially Windows) expect.

Save on the save document modal dialog (common save dialog) in Word _closes the save dialog window_.

Apply on modals in Windows do not close the dialogs, but save always does.

And this isn't just about people who use Windows either. On Linux, KDE is rather popular, and they use the same OK / Apply / Cancel paradigm as Windows. (Where "apply" does not close the modal dialog.)

Here's a screenshot I just took from the latest version of KDE, with the file manager preferences dialog showing OK (which applies and closes the window), Apply (which only applies and doesn't close the window), and Cancel (which closes the window without applying). Just like on Windows, Apply is redundant as OK applies and closes the window. 

Screenshot from 2021-11-08 13-39-49

Clicking on Apply and Cancel would actually do the same thing as clicking on OK. 

I've actually seen lots of people in usability studies in the past as well as just watching people use Windows and KDE over the years click the Apply button first and then OK button. Either you're going to apply the settings or not and then you're going to close the window. All of those tasks are reflected in the OK (or Save or any other action) and Cancel buttons.

But the point I'm making here is that it's common for Windows and KDE users to expect an Apply button in modal dialogs to not close the window, therefore, it's a loaded term and if we close modals with the "Apply" button, it will diverge from their expectations. Hence this PR.

garrett:
Windows has done a lot to get rid of the Apply button in Settings (as it's all in an instant-apply app now), but it still exists when you get to modal dialogs (especially older, untouched ones, which still show up in Windows 10 and 11). 

Here's a device dialog in Windows 10:

Screenshot from 2021-11-08 13-47-38

Apply doesn't close the dialog (and is disabled until changes are made, just like in KDE), but the other buttons (in this case OK and Cancel) do.

I'm assuming that Microsoft saw the same thing I had over the years and realized that Apply was an anti-pattern and tried to move a lot of their most common settings away from it in the Windows 8 & 10 era. (But it still even persists in Windows 11, as Microsoft never truly fully changes their UI, they just bury the old and rework the most common user-facing stuff.)

martinpitt:
@garrett : Thanks! Sorry for the misunderstanding -- I didn't want to say "I disagree", just wanted to confirm that I understood it in the right way, as I haven't used Windows nor KDE in a long time.

In the interest of easing translator work, we should of course replace all buttons with just :+1: and :x: ! *cough*

garrett:
Rebased

garrett:
Rebased; also added network and kdump dialogs to the apply → save change.

martinpitt:
I adjusted the tests for the renaming and [updated the pixel tests](https://github.com/cockpit-project/pixel-test-reference/compare/e37891c2b526d1de366e25ae4f57ef314990a45c..6579d45fef970177293dc1c666530fb150476109).

martinpitt:
That [failure happened](https://logs.cockpit-project.org/logs/pull-16475-20220318-112949-5b57a2a4-fedora-35/log.html#61-1) because tests run multiple times. I added another commit which fixes that, and at the same time makes the pixel test a lot simpler.

As I now meddled too much with this, @marusak, can you please do the final review?

martinpitt:
[Some](https://logs.cockpit-project.org/logs/pull-16475-20220318-172950-02d7d82c-arch/log.html#265) [unrelated](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16475-20220318-172950-02d7d82c-debian-stable/log.html) flakes, I sent PR #17152 to fix them.

martinpitt:
Ready for review. Can we please land this soon, as it is rather invasive? (pixel tests, lots of changed places)

pkg/kdump/kdump-view.jsx
```diff
@@ -357,7 +357,7 @@ export class KdumpPage extends React.Component {
             actions: [
                 {
                     clicked: this.handleApplyClick.bind(this),
```
marusak, line 359: So this leaves behind `handleApplyClick`. In `pkg/networkmanager/dialogs-common.jsx` it leaves behind `id=...-apply`. These are just names so does not matter, just maybe would be a bit cleaner for future ourselves when we wonder why we call it "apply" :) `git grep "[aA]pply" pkg/` shows a lot of such cases. Not necessary to do anything, just pointing it out.

marusak (commented):
There is one new occurrence in `pkg/systemd/overview-cards/cryptoPolicies.jsx` - It has `Apply and reboot` or `Apply only` buttons. I wonder if these should not also be `Save and reboot`/`Save only` :thinking:

pkg/kdump/kdump-view.jsx
martinpitt: Good point! I renamed these, at least all that I could find which immediately make sense.

I also adjusted cryptoPolicies, that landed recently and we didn't have it on the radar in this PR.

marusak (approved):
Thank you!

PR #16483, opened 2021-10-15
metrics: Initialize top most timestamp to something usable

### LINKED ISSUE #16462: metrics: When current historical metrics do not load, _all_ older metrics load and overwhelm the browser
When I load the metrics page on my work laptop, it fails to load the historical metrics and then auto-loads all the historical metrics it can find.

This makes the browser (Firefox specifically) unresponsive and makes debugging in the developer tools difficult to impossible.

The metrics page should always limit the number of data points by default, even if it goes into a fallback mode.

marusak:
When the initial load of the last 12 hours does not return any data
then after one minute when we try to update the last minute we load all
the data there is.
The updating logic calls `load_data` with `this.most_recent`. This variable
has initial value `0` and is updated when some data are parsed.
Calling `load_data(0)` translates to loading all the data it can find.

Fixes #16462

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1119,6 +1119,7 @@ class MetricsHistory extends React.Component {
                 .then(out => {
                     const now = parseInt(out.trim()) * 1000;
                     const current_hour = Math.floor(now / MSEC_PER_H) * MSEC_PER_H;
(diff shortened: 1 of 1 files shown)


martinpitt:
The unit test failures here just got fixed, can you please rebase? The two test failures are unrelated, flakes due to extra "3x" strain.

PR #16497, opened 2021-10-19
lib: port ListingPanel component away from the deprecated tertiary Nav to Tabs

KKoukiou:
Presence property was dropped as it's not used anyway.

The previous default 'loadOnDemand' behavior is now achieved using the
'mountOnEnter' Tabs component property.

Pixel tests were updated accordingly.

### DIFF
--- a/pkg/lib/cockpit-components-listing-panel.jsx
+++ b/pkg/lib/cockpit-components-listing-panel.jsx
@@ -19,92 +19,32 @@
 
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Flex, Nav, NavItem, NavList } from '@patternfly/react-core';
+import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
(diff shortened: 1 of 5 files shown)


martinpitt:
Ouch @ the broken buttons on the [storage page](https://github.com/cockpit-project/pixel-test-reference/blob/a50ed28cd6e7e6f7a8d50829a2b988b9a2edb323/TestStorageBasic-testBasic-partition-mobile-pixels.png) in mobile mode! But this is not a regression, it already happens on main, and even in medium-width (~ 900px) desktop mode:

image

So this can/should be fixed independently, which will then require updating the pixel tests -- but I think that's actually a feature, it makes reviewing/validating much more explicit.

I tested this and confirm that this brings back keyboard navigation on both the SSH dialog and the Storage page. Many thanks!

(FTR, keyboard navigation is still broken on at least the Storage page's kebab action menu)

PR #16505, opened 2021-10-21
storage: Allow mounting all NFS directories

mvollmer:
and not only those that are in the drop down list.  The list is not
always accurate and the user might have a better idea of what will
work.

### DIFF
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -537,6 +537,8 @@ const TypeAheadSelectElement = ({ options, change }) => {
     return (
         <TypeAheadSelect
             variant={SelectVariant.typeahead}
+            isCreatable
+            createText={_("Use")}
             id="nfs-path-on-server"
             isOpen={isOpen}
             selections={value}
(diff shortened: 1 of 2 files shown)


pkg/storaged/dialog.jsx
```diff
@@ -537,6 +537,8 @@ const TypeAheadSelectElement = ({ options, change }) => {
     return (
         <TypeAheadSelect
             variant={SelectVariant.typeahead}
+            isCreatable
```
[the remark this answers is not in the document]
KKoukiou: Int this case yes let's keep the isCreatetable property. Sorry for the late answer, missed the notification.

test/verify/check-storage-nfs
```diff
@@ -170,6 +182,35 @@ class TestStorageNfs(StorageCase):
...
+
+        m.execute("mkdir /home/foo /home/bar")
+        m.write("/etc/exports", "/home/foo 127.0.0.0/24(rw)\n/home/bar 127.0.0.0/24(rw)\n")
+        m.execute("systemctl restart nfs-server")
+
+        # Break showmount.  Cockpit uses showmount to list exported
+        # directories, but that relies on a properly configured
+        # firewall etc.  Even if showmount doesn't work, Cockpit
+        # should allow people to mount arbitrary directories.
+        m.execute("chmod a-x /usr/sbin/showmount")
```
KKoukiou, line 199: This needs to be cleaned up in teardown, so an addCleanup  is necessary here.

mvollmer: This is not a `@nondestructive` test, though.

KKoukiou (approved):

test/verify/check-storage-nfs
KKoukiou: Aha! Did not realize, thanks!

PR #16516, opened 2021-10-26
tuned: Port dialog to react

KKoukiou:
This is current main:

image

This PR:

image

- [x] https://github.com/cockpit-project/bots/pull/2568


garrett:
The list should be scrollable, not the dialog. Opening an issue about this now...

PR #16518, opened 2021-10-26
storage: Redesign "content rows" everywhere

mvollmer:
This is a step towards https://github.com/cockpit-project/cockpit/discussions/16440

Screenshot (up-to-date): https://github.com/cockpit-project/cockpit/pull/16518#issuecomment-972712100
Demo (not pixel-perfect, but still valid): https://www.youtube.com/watch?v=xLfxxm-VKmU

------

### Storage: More information in table rows

The rows of the various tables now show more information about filesystems and other things.  You can now immediately see where a filesystem is mounted, there are links to the pages for volume groups etc, and the usage level is shown.

image

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -154,9 +154,9 @@ export const ListingTable = ({
             <React.Fragment key={rowKey + "-inner-row"}>
                 <Tr {...rowProps}>
                     {isExpandable
-                        ? (row.expandedContent
-                            ? <Td expand={
-                                {
+                        ? <Td expand={
+                            row.expandedContent
+                                ? {
                                     rowKey,
                                     isExpanded,
                                     onToggle: () => {
@@ -164,9 +164,8 @@ export const ListingTable = ({
                                             afterToggle(!expanded[rowKey]);
                                         setExpanded({ ...expanded, [rowKey]: !expanded[rowKey] });
                                     }
-                                }} />
-                            : <Td />
-                        ) : null}
+                                } : null} />
+                        : null}
                     {onSelect &&
                         <Td select={{
                             rowIndex,
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -420,7 +420,7 @@ function create_tabs(client, target, is_partition, is_extended) {
 }
 
 function block_description(client, block) {
-    let type, used_for, link, size;
+    let type, used_for, link, size, critical_size;
     const block_stratis_blockdev = client.blocks_stratis_blockdev[block.path];
     const block_stratis_locked_pool = client.blocks_stratis_locked_pool[block.path];
     const vdo = client.vdo_overlay.find_by_backing_block(block);
@@ -458,6 +458,7 @@ function block_description(client, block) {
             used_for = vgroup.Name;
             link = ["vg", used_for];
             size = [block_pvol.Size - block_pvol.FreeSize, block_pvol.Size];
+            critical_size = 1;
         } else if (client.mdraids[block.MDRaidMember]) {
             const mdraid = client.mdraids[block.MDRaidMember];
             type = _("RAID member");
@@ -498,7 +499,8 @@ function block_description(client, block) {
         type: type,
         used_for: used_for,
         link: link,
-        size: size
+        size: size,
+        critical_size: critical_size
     };
 }
 
@@ -543,7 +545,7 @@ function append_row(client, rows, level, key, name, desc, tabs, job_object) {
         { title: desc.link ? <StorageLink onClick={() => cockpit.location.go(desc.link)}>{desc.used_for}</StorageLink> : desc.used_for },
         {
             title: desc.size.length
-                ? <StorageUsageBar stats={desc.size} critical={0.95} block={name} />
+                ? <StorageUsageBar stats={desc.size} critical={desc.critical_size || 0.95} block={name} />
                 : utils.fmt_size(desc.size),
             props: { className: "ct-text-align-right" }
         },
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -33,7 +33,7 @@ import { ExclamationTriangleIcon } from "@patternfly/react-icons";
 
 import { ListingTable } from "cockpit-components-table.jsx";
 import { ListingPanel } from 'cockpit-components-listing-panel.jsx';
-import { StorageButton, StorageBarMenu, StorageMenuItem } from "./storage-controls.jsx";
+import { StorageButton, StorageLink, StorageBarMenu, StorageMenuItem, StorageUsageBar } from "./storage-controls.jsx";
 import * as PK from "packagekit.js";
 import {
     format_dialog, parse_options, extract_option, unparse_options
@@ -44,7 +44,6 @@ import { FilesystemTab, is_mounted, mounting_dialog, get_fstab_config } from "./
 import { CryptoTab, edit_config } from "./crypto-tab.jsx";
 import { get_existing_passphrase, unlock_with_type } from "./crypto-keyslots.jsx";
 import { BlockVolTab, PoolVolTab, VDOPoolTab } from "./lvol-tabs.jsx";
-import { PVolTab, MDRaidMemberTab, VDOBackingTab, StratisBlockdevTab } from "./pvol-tabs.jsx";
 import { PartitionTab } from "./part-tab.jsx";
 import { SwapTab } from "./swap-tab.jsx";
 import { UnrecognizedTab } from "./unrecognized-tab.jsx";
@@ -106,7 +105,6 @@ function create_tabs(client, target, is_partition, is_extended) {
 
     const block_fsys = content_block && client.blocks_fsys[content_block.path];
     const block_lvm2 = block && client.blocks_lvm2[block.path];
-    const block_pvol = content_block && client.blocks_pvol[content_block.path];
     const block_swap = content_block && client.blocks_swap[content_block.path];
 
     const block_stratis_blockdev = block && client.blocks_stratis_blockdev[block.path];
@@ -118,8 +116,8 @@ function create_tabs(client, target, is_partition, is_extended) {
 
     const is_filesystem = (content_block && content_block.IdUsage == 'filesystem');
     const is_stratis = ((content_block && content_block.IdUsage == "raid" && content_block.IdType == "stratis") ||
-                      (block_stratis_blockdev && client.stratis_pools[block_stratis_blockdev.Pool]) ||
-                      block_stratis_locked_pool);
+                        (block_stratis_blockdev && client.stratis_pools[block_stratis_blockdev.Pool]) ||
+                        block_stratis_locked_pool);
 
     // Adjust for encryption leaking out of Stratis
     if (is_crypto && is_stratis)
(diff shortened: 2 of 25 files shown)


mvollmer:
Changes to pixel test references:
https://github.com/cockpit-project/pixel-test-reference/compare/fc9ba8ef81f5a6a7b688761faad2e3cb5f015a70..b4b12b145e376c178001f38fff2242fd415f6510

mvollmer:
Growing thin pools is somehow broken here.

mvollmer:
Empty partitioned disks look misaligned:

image

Once there is something in them, they look better:

image

garrett:
The alignment is a bit off. Here are the same screenshots with things changed a bit:

image

---

The alignment is sub-optimal here still, since the button is just floating:

image

This is better, as the button is aligned, but the thing is... we actually want to align the action buttons when they exist, so I guess the above one is better within the context of everything?

image

---

Here are guidelines to clearly show the alignment. To achieve this, we'd need to have the left align / right align and also the appropriate padding on the edges. PatternFly usually handles a lot of this for us, but sometimes it doesn't.

image

image

There are utility classes for alignment (https://www.patternfly.org/v4/utilities/alignment) and spacing (padding/margin, https://www.patternfly.org/v4/utilities/spacing). Sometimes the units aren't right and we need custom CSS. But often, these are appropriate. But you have to include the CSS utilities in to use them.

mvollmer:
Now this uses the official PF4 table actions, which (finally) make actions look ok for mobile.

image

image

"Create partition" is out of place in mobile because this doesn't use the PF4 table actions yet. A single button without menu entries still produces a menu, but with a empty popdown.

mvollmer:
I have moved the free space label into the first column, which is "Name". I think this looks better in desktop layout, even if one can argue that is is wrong in mobile.

Alignment is still a bit weird.  When very wide, it is aligned with the panel header:

image

When it gets a bit narrower, they get misaligned:

image

garrett:
Why are the buttons so bad when they're in the rows?

image

image

We don't have that issue elsewhere. They're lacking vertical spacing. There's no margin around them or padding in the table rows Should they be using the smaller variant too when in the rows too? (Unsure about the sizing. Am sure about the spacing though.)

mvollmer:
> Why are the buttons so bad when they're in the rows?

Good question. Patternfly explicitly switches off padding for them, but their example still has some padding in the rows anyway. I'll figure it out.

mvollmer:
> > Why are the buttons so bad when they're in the rows?
> 
> Good question. Patternfly explicitly switches off padding for them, but their example still has some padding in the rows anyway. I'll figure it out.

There is no padding because we explicitly use `variant="compact"` for the storage tables.  Here is how they look in the default variant:

image

mvollmer:
Okay, it turns out that we can keep our old action widgets (with all their extra features compared to regular buttons and menu items), and just need to put the `pf-c-table__action` class on the last column to get a proper mobile layout.

mvollmer:
Ouch, there is trouble with narrow windows:

image

The table gets stuck at a certain maximum width and the side panel will overlap it.  This is already happening on master, but at a narrower width and is thus much less noticable.

mvollmer:
> Ouch, there is trouble with narrow windows:

I 'fixed' that with a Flex...

mvollmer:
Okay, I think I am happy with this now.

Drive with empty partition table:

image

Drive with various partitions:

image

Stratis:

image

"Worst case" volume group (long type label, big button):

image

Same in the most narrow two-column layout:

image

Same in the widest one-column layout:

image

Same in most narrow one-column layout:

image

Same in mobile layout:

image

garrett (changes requested):
The /dev/sda1 should not be under the remove button. It's too far away from the rest of the information.

image

Free space is still not aligned to the left. (It should be.) I mentioned this above: https://github.com/cockpit-project/cockpit/pull/16518#issuecomment-963958780

It's looking pretty great otherwise.

pkg/storaged/stratis-details.jsx
```diff
@@ -656,9 +652,8 @@ export const StratisPoolDetails = ({ client, pool }) => {
             <CardBody className="contains-list">
                 <ListingTable emptyCaption={_("No filesystems")}
                               aria-label={_("Filesystems")}
-                              columns={[_("Name"), _("Used for"), _("Size"), _("Actions"), _("Menu")]}
+                              columns={[_("Name"), _("Used for"), _("Size")]}
                               showHeader={false}
-                              variant="compact"
```
KKoukiou, line 649: Hmm are you sure about this? It looks quite bulky now.

Now:
Screen Shot 2021-11-18 at 15 50 58

Before:
Screen Shot 2021-11-18 at 15 50 48

pkg/storaged/fsys-panel.jsx
```diff
@@ -182,11 +168,12 @@ export class FilesystemsPanel extends React.Component {
                     className={mounts.length ? 'table-hover' : ''}
                     onRowClick={onRowClick}
                     columns={[
-                        { title: _("Name"), transforms: [cellWidth(30)], sortable: true },
-                        { title: _("Mount point"), transforms: [cellWidth(30)], sortable: true },
-                        { title:  _("Size"), transforms: [cellWidth(40)] }
+                        { title: _("Source"), transforms: [cellWidth(25)], sortable: true },
```
KKoukiou, line 171: I think these transforms don't work anymore with the new implementation. I forgot to port them to the 'new' way when I ported pkg/lib/cockpit-components-table.

pkg/storaged/details.jsx
```diff
@@ -43,14 +43,19 @@ export class StdDetailsLayout extends React.Component {
...
                     </GridItem>
-                    <GridItem md={8} lg={9}>
-                        <div id="detail-content">
-                            { this.props.content }
-                        </div>
-                        <JobsPanel client={this.props.client} />
-                    </GridItem>
-                    <GridItem id="detail-sidebar" md={4} lg={3}>
-                        { this.props.sidebar }
+                    <GridItem span={12}>
```
KKoukiou, line 46: This Grid wrapper is not needed anymore then. Removing it should have 0 effect.

KKoukiou (commented):
So i really like this Pr, I left a few comments, but I think we get somewhere better.

pkg/storaged/storage.scss
```diff
@@ -166,7 +166,7 @@ td.job-action {
 
 .content-action {
     text-align: right;
```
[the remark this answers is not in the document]
mvollmer: Yes, it's still needed.  `pf-c-table__action` does not include right aligning, PF must do that some other way.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -165,7 +165,7 @@ export const ListingTable = ({
                                         setExpanded({ ...expanded, [rowKey]: !expanded[rowKey] });
                                     }
                                 }} />
-                            : <Td />
+                            : <Td className="pf-c-table__toggle" />
```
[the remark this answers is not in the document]
mvollmer: Nice, this is better!

pkg/storaged/stratis-details.jsx
mvollmer: @garrett?

pkg/storaged/fsys-panel.jsx
mvollmer: Right, they don't seem to make a difference.  Removed.

pkg/storaged/details.jsx
mvollmer: The whole Grid still has an effect, it produces the gutter between header and content.  It might be a bit overkill for that now, though...

KKoukiou:
@mvollmer needs some tests adjustments now :)

mvollmer:
> The /dev/sda1 should not be under the remove button. It's too far away from the rest of the information.
> 
> image

I agree, but where should we place it instead?  We can give the row a max-width, which would look like this:

image

> Free space is still not aligned to the left. (It should be.)

Fixed!

> It's looking pretty great otherwise.

Nice, thanks!

garrett:
> I agree, but where should we place it instead? We can give the row a max-width, which would look like this:

We can also use gap in flex nowadays, as that's supported in all browsers for a little while now. (WebKit was the longest holdout, until the end of April 2021... and it has had a few releases on macOS and iOS since. FWIW: Chrome added it in 2020 and Firefox in 2018.) https://caniuse.com/flexbox-gap Using a gap could help to ensure the items are always separated by a minimum of space (which I think was part of the original problem).

So we might want to reconsider the layout anyway. But this would probably be for another PR; this one is rather large enough already. :wink:

garrett:
> We can give the row a max-width

BTW: A max-width is the wrong thing to do here. What happens when the strings get too long? It might look OK in the screenshot with your current content, but it could break the UI a bit with other strings (especially the long device string names based on GUIDs and such).

mvollmer:
> What happens when the strings get too long?

The row will start wrapping like this:

image

I think this is okay.

mvollmer:
> So we might want to reconsider the layout anyway. But this would probably be for another PR; this one is rather large enough already. wink

I agree.  With the big redesign, we wont have any of these sidebar boxes anymore at all.

garrett (approved):
Overall: Approve! :+1: 

Notes: 

1. The padding feels a bit much, but I guess I'm used to our old tables, where we had padding in-between PF's large and compact sizes. Compact is definitely wrong; default is fine (although spacious).
2. Speaking of padding: It would be good to make the main listing in the storage overview page not be compact, to match the details page. That can be done in this PR or a follow-up.
3. The device label still irks me, but let's handle that in a follow up. It'll change in the redesign anyway, so it's not really a big deal.

The second point is the only actionable item, and it makes sense more as a follow-up. It's probably not big really, but I don't want to stall this PR. :grin:

pkg/storaged/content-views.jsx
```diff
@@ -724,15 +745,18 @@ export const Block = ({ client, block, allow_partitions }) => {
 function append_logical_volume_block(client, rows, level, block, lvol) {
     let tabs, desc;
     if (client.blocks_ptable[block.path]) {
+        // XXX - just make this a link to the details page for the block device?
```
martinpitt, line 747: Does this still need to be resolved here, or is that a todo for later?

martinpitt (commented):
Thanks! The code changes look very reasonable. I interactively tested this, and really like it! Especially showing the mount points in the overview, which is so much more useful.

This is conflicty and thus needs a rebase (probably due to Katerina's recent conversion to f-strings), but otherwise let's land that soon.

martinpitt:
@mvollmer : ... and of course you need to update the pixel tests

pkg/storaged/content-views.jsx
mvollmer: Ah, good catch. I think we need to resolve this. This is for partition tables on logical volumes, which probably nobody is doing and Cockpit offers no way to do it, so it is super rare. It might not even be supported by LVM2. But we need to make sure nothing explodes. (I should add a test.)

mvollmer:
> 2\. It would be good to make the main listing in the storage overview page not be compact, to match the details page.

Good point! I do it here.

mvollmer:
> 3\. The device label still irks me, but let's handle that in a follow up. It'll change in the redesign anyway, so it's not really a big deal.

Yeah, agreed.

pkg/storaged/content-views.jsx
mvollmer: This now links to the details page for the block.  Puting partition tables on device mapper devices seems to be broken in all kinds of ways.  It is possible to do it and get the corresponding block devices like `/dev/mapper/vgroup0-lvol0p1`, but formatting doesn't really work, and removing the partition table does not remove the device mapper devices...

martinpitt (approved):
Yay, thanks! I added release-note, can you please add one to the PR description?

garrett (approved):
:+1: LGTM!

I'm not used to the non-compact tables. It's like this:

a little cat on a big chonky cat

I'd prefer something in the middle, like a `pf-m-compact-sorta` or something.

PR #16522, opened 2021-10-26
Remove jQuery

marusak:
I haven't removed everything yet. This is WIP, I want to get a base run now to catch what I broke. Then I need to test one thing where it is being used. And finally I will try to convert rest of the cases.

- [ ] followup - remove jquery and bootstrap from `package.json`


marusak:
Oh no, conflicts with `node_modules` :/

Getting there, needs some css love. Looks like this:
Screenshot from 2021-11-02 08-01-08

I need to remove those white borders around dropdowns (`border: none` does not seem to work) and mobile navigation is broken.

garrett:
These outlines are made with a `::before` pseudo element in the PatternFly CSS.

A simple approach via pure CSS, using PF selectors:

```scss
.pf-c-dropdown__toggle::before,
.pf-c-dropdown__toggle.pf-m-action .pf-c-dropdown__toggle-button::before {
  border: none;
}
```

This can also be changed via PF variables:

```scss
.pf-c-dropdown__toggle::before,
.pf-c-dropdown__toggle.pf-m-action .pf-c-dropdown__toggle-button::before {
  --pf-c-dropdown__toggle--before--BorderRightColor: transparent;
  --pf-c-dropdown__toggle--before--BorderLeftColor: transparent;
  --pf-c-dropdown__toggle--before--BorderTopColor: transparent;
  --pf-c-dropdown__toggle--before--BorderBottomColor: transparent;
}
```

These selectors come from PF's `dropdown.scss`. We'd want to probably adjust the selectors to make them specific based on the heading and/or show that they're Cockpit-specific. And because the variables cascade, we probably want to set them as high up as possible (probably the same place where PF does).

So we'd _really_ probably want to do something in the local SCSS like _this_ (instead of the above):

```scss

.ct-header-item {
  // Drop PF borders for the dropdowns
  --pf-c-dropdown__toggle--before--BorderRightColor: transparent;
  --pf-c-dropdown__toggle--before--BorderLeftColor: transparent;
  --pf-c-dropdown__toggle--before--BorderTopColor: transparent;
  --pf-c-dropdown__toggle--before--BorderBottomColor: transparent;
}
```

Since we should already have a block for `ct-header-item` in the file, we can just add these four variables and the comment above them to the existing block.

garrett:
We probably need `--pf-c-dropdown__toggle-icon--MarginRight: 0;` at the same level as well, as there's a double-spacing issue with the button padding and the icon's margin.

I guess I should get my hands dirty in this PR and make a bunch of CSS changes, right? :wink:  I do see a few other small issues.

pkg/shell/topnav.jsx
```diff
@@ -116,6 +116,7 @@ export class TopNav extends React.Component {
             <DropdownItem key="creds" id="sshkeys" component="button" onClick={() => this.setState({ credentialsDialogOpened: true })}>
                 {_("SSH keys")}
             </DropdownItem>,
+            <DropdownSeparator key="separator" />,
```
marusak, line 122: The `key` needs to be unique. On line 102 this array already had such key.

marusak: (This was mostly note for myself to fix this on squashing, no need to repush)

garrett:
Button styling is only on the text; the icon is outside of the bounds. However, I believe I added this as a custom CSS mod. I think we just need to change it to the parent element. I'll look into this.

Screenshot 2021-11-03 at 14-28-55 Overview - garrett Bolt

---

We have the same icon for both of these. Perhaps we should only have an icon when locked (similar to desktop mode. The blue link style might be odd? But I think this is on main too.

Screenshot from 2021-11-03 14-31-08
Screenshot from 2021-11-03 14-30-48

---

The spinner is in the wrong place. (It should be to the left of the host menu in mobile mode.) However, it's like this on main too, so it's not a regression, but still a bug. If you can fix it in this PR, that would be awesome. Otherwise, it would be nice to fix this in a follow-up.

image

---

Also, the menus don't dismiss when you click anywhere else (especially other menus, especially in mobile). It's kind of weird. But It's probably not a regression? (I haven't checked.)

garrett:
Opening a modal from the shell in mobile mode causes extra whitespace to persist under the navigation bar. I don't know why yet. I'll have to finish debugging it tomorrow.

martinpitt:
Some observations about [TestActivePages.testBasic failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16522-20211103-172255-112e018c-debian-testing/log.html#299):
 * The active page now immediately reloads, whereas on main you just get a blank page, but it apparently stays the active one. You need to switch to some other frame and then back.
 * This is rather unimportant -- we can just easily disallow killing the currently active frame, or accept that the current page reloads after killing. This feature is very well  hidden, and more of a developer thing anyway.
 * I wanted to test "shell: Rewrite top navigation to React" in isolation, as the "shell: Remove jQuery" commit breaks the top menus entirely. However, with the following hack it becomes possible:
```diff
--- pkg/shell/index.html
+++ pkg/shell/index.html
@@ -59,6 +59,7 @@
             </ul>
           </div>
           <div class="dropdown ct-header-item ct-header-menu">
+            <a role="link" tabindex="0" id="active-pages" translate="yes" class="navbar-advanced">Active pages</a>
             <a role="link" tabindex="0" id="navbar-dropdown" class="ct-nav-toggle dropdown-toggle" data-toggle="dropdown">
               <span class="icon-custom pficon pficon-cog"></span>
               <span translate>Session</span>
@@ -72,7 +73,6 @@
               </li>
               <li class="divider display-language-menu"></li>
               <li>
-                <a role="link" tabindex="0" id="active-pages" translate="yes" class="navbar-advanced">Active pages</a>
               </li>
               <li>
                   <a role="link" tabindex="0" id="credentials-item" translate="yes">SSH keys</a>
```
    I confirm that this *does* keep the old behaviour. So the behaviour change really is in the "shell: Rewrite top navigation to React" bug.

marusak:
Huh, so on mobile switching to administrative mode does not remember it after relogin.  This should be completely unrelated to changing host switcher. Possibly something related to it being in `<Dropdown>`?

marusak:
aha! In desktop (and in previous version) it was never unmounted. Here every time the dropdown is closed it is unmounted and that somehow breaks the behaviour.

pkg/shell/machines/machines.js
```diff
@@ -647,7 +647,7 @@ function Loader(machines, session_only) {
             };
             channel.addEventListener("message", channels_listeners_message[host]);
 
-            channels_listeners_close[host] = () => {
```
marusak: This belongs to different commit.

marusak:
Created a few standalone commits - each is working on its own.
Tests are almost there, one more problem with mobile privileges switcher - needs investigation. https://logs.cockpit-project.org/logs/pull-16522-20211109-101108-a2f58599-fedora-34-mobile/log.html

I would say this is ready for review while I try to figure out the last failing tests.

pkg/shell/base_index.js
```diff
@@ -46,8 +45,7 @@ function Frames(index, setupIdleResetTimers) {
     self.iframes = { };
 
     function remove_frame(frame) {
-        $(frame.contentWindow).off();
```
martinpitt, line 49: I can't actually see which event handlers were attached to contentWindow. Was that just a leftover, or does/did it happen through aliasing? The [docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/remove) don't say anything about auto-cleaning event handlers on removal.

pkg/shell/base_index.js
```diff
@@ -76,7 +74,8 @@ function Frames(index, setupIdleResetTimers) {
         frame.timer = null;
 
         try {
-            ready = $("body", frame.contentWindow.document).is(":visible");
+            if (frame.contentWindow.document && frame.contentWindow.document.body)
+                ready = frame.contentWindow.document.body.offsetWidth > 0 && frame.contentWindow.document.body.offsetHeight > 0;
```
martinpitt, line 76: Is non-zero offset what jQuery defined as ":visible"? I.e. that's not the same as the `hidden` attribute or `display: none` style, but more like "finished loading"?

pkg/shell/base_index.js
```diff
@@ -409,8 +408,8 @@ function Router(index) {
                     source = register(child);
                 }
                 if (source) {
-                    const reply = $.extend({ }, cockpit.transport.options,
-                                           { command: "init", host: source.default_host, "channel-seed": source.channel_seed }
+                    const reply = cockpit.extend({ }, cockpit.transport.options,
+                                                 { command: "init", host: source.default_host, "channel-seed": source.channel_seed }
```
martinpitt, line 410: OK, but `Object.assign()` should do the same, and of course the l33t ES6 way is to use the spread operator :grin:

pkg/shell/base_index.js
```diff
@@ -575,20 +576,10 @@ function Index() {
...
-    });
-
-    /* Handles an href link as seen below */
-    $(document).on("click", "a[href]", function(ev) {
-        const a = this;
-        if (!a.host || window.location.host === a.host) {
-            self.jump(a.getAttribute('href'));
-            ev.preventDefault();
-            ev.stopImmediatePropagation();
-        }
```
martinpitt, line 591: So this translated standard clicks into cockpit.jump() calls to track their history? This seems to have gone without replacement -- does that change anything wrt. history/back button handling?

pkg/shell/base_index.js
```diff
@@ -854,8 +846,10 @@ function CompiledComponents() {
     self.items = {};
 
     self.load = function(manifests, section) {
-        $.each(manifests || { }, function(name, manifest) {
-            $.each(manifest[section] || { }, function(prop, info) {
+        Object.entries(manifests || { }).forEach(m => {
+            const [name, manifest] = m;
```
martinpitt, line 787: Note that you can use syntactical destructuring to write this in a nicer way: `.forEach(([name, manifest]) => { ...`

martinpitt (changes requested):
Review of the first commit (I need to do this in several smaller batches). I mostly have some questions, feel free to ignore/resolve the stylistic nitpicking. Nice work!

pkg/shell/active-pages-modal.jsx
```diff
@@ -0,0 +1,113 @@
...
+        this.state.frames.forEach(element => {
+            if (element.selected)
+                this.props.frames.remove(element.host, element.component);
+        });
+        this.props.onClose();
+    }
+
+    render() {
+        const self = this;
+        const frames = self.state.frames.map(frame => {
```
martinpitt, line 69: No need for the "self", just use `this.state` here. (not used anywhere else)

pkg/shell/active-pages-modal.jsx
```diff
@@ -0,0 +1,113 @@
...
+        for (const address in props.frames.iframes) {
+            for (const component in props.frames.iframes[address]) {
+                const iframe = props.frames.iframes[address][component];
+                frames.push({
+                    frame: iframe,
+                    component: component,
+                    address: address,
+                    name: iframe.getAttribute("name"),
+                    visible: iframe.getAttribute("data-active") === 'true',
+                    selected: iframe.getAttribute("data-active") === 'true',
```
martinpitt, line 42: That looks weird -- there's only ever one active frame, isn't it? Why are all the others initialized as non-visible? I may misunderstand what data-active means of course, and the old code initialized `visible` like it does here. Just the `selected` is new.

pkg/shell/active-pages-modal.jsx
```diff
@@ -0,0 +1,113 @@
...
+    render() {
+        const self = this;
+        const frames = self.state.frames.map(frame => {
+            const columns = [{
+                title: <Split>
+                    <SplitItem isFilled>
+                        {frame.displayName}
+                    </SplitItem>
+                    <SplitItem>
+                        {frame.visible && <Label color="blue">{_("active")}</Label>}
```
martinpitt, line 76: So this maps `data-active` to frame.visible and back to label "active". Could this be called frame.active then to reduce confusion?

martinpitt (changes requested):
Review of second and third commit.

marusak:
My theory is that when the auth switcher dialog changes to additional info, it does not happen in one modal, but the modal is closed for brief moment and then reopened with different content. That does not work greatly with dropdown, as it is closed when action is selected (when modal is opened, once modal is closed, the dropdown is closed). Since for a moment the modal is closed, the dropdown gets closed as well and then when new modal should be reopened, it is already unmounted.

garrett:
> You can also remove bootstrap at this point!

exploding head emoji

(I need to get back to this PR and re-review it again.)

pkg/shell/base_index.js
marusak: [Since jquery 1.3.2](https://blog.jquery.com/2009/02/20/jquery-1-3-2-released/):

> an element is visible if its browser-reported offsetWidth or offsetHeight is greater than 0.

marusak: And yes, this is until page shows nothing in the frame - for that duration we show spinner in the top row. Once page shows something (usually spinner), then we remove the spinner from top navigation and let frame itself handle loading.

pkg/shell/base_index.js
marusak: yes, I am aware of all three options, but went with this one for no good reason. But I think we might want to replace all `cockpit.extend` with something directly from JS. (in followup)

pkg/shell/base_index.js
marusak: I believe that this used to be used for the main navigation. But that now is properly handled by components themselves, e.g. https://github.com/cockpit-project/cockpit/blob/main/pkg/shell/indexes.js#L341

I think it was forgotten in https://github.com/cockpit-project/cockpit/commit/377054a3f1189bb26122ed8c8b798cea8137729d

pkg/shell/active-pages-modal.jsx
marusak: `s/visible/active` I guess. I just moved it, but I can rename it. Selected means if you checked that frame in the modal to be closed. By default active page is selected.

pkg/shell/active-pages-modal.jsx
marusak: yes, will do

marusak:
> You can also remove bootstrap at this point!

Indeed. But I'll do a followup with rebuilding npm.

pkg/shell/base_index.js
martinpitt: Thanks! I added it to my "later" todo list, let's not block this PR on such neglibilities.

marusak:
I call this green!

garrett (approved):
It seems to work fine. The main way I can tell that I'm actually using this PR and not main is that the PF menus in mobile are not dark, but light.

I tried adding `pf-m-dark` to them in the DOM, but it didn't seem to change things. I guess PatternFly probably doesn't have dark menus in PF4 at the moment.

Anyway, LGTM.

If you know of an easy way to fix the left menus to make them dark, like the main menu and host switcher, that would be great... and it could be a follow-up as it's a minor cosmetic thing for mobile only. Otherwise, this is superb!

martinpitt (approved):
Looks great to me, dakujem! :heart: 

I have [one outstanding question](https://github.com/cockpit-project/cockpit/pull/16522#discussion_r745578147) from the previous review round, but that's just for curiousity -- this obviously works.

pkg/shell/base_index.js
marusak: We actually do setup some event listeners:
https://github.com/cockpit-project/cockpit/blob/main/pkg/shell/base_index.js#L97
and https://github.com/cockpit-project/cockpit/blob/main/pkg/shell/base_index.js#L552
But `element.remove()` should also disconnect all listeners. (Not that you would get any of these events on element that was removed from the tree anyway).

marusak:
> It seems to work fine. The main way I can tell that I'm actually using this PR and not main is that the PF menus in mobile are not dark, but light.

> I tried adding pf-m-dark to them in the DOM, but it didn't seem to change things. I guess PatternFly probably doesn't have dark menus in PF4 at the moment.

Indeed, this is basic PR dropdown and it is not provided a dark themed one. The color of the dropdown is defined through `--pf-c-dropdown__menu--BackgroundColor` variable, so it can be changed. But that of course then needs changing color of text, separators... So If we want this, PF should provide the styling on its own and we would just add some class or something.

PR #16532, opened 2021-10-28
storage: Support Stratis API version 3.0 

mvollmer:
- [x] Handle missing "Used" and "TotalPhysicalUsed" properties.
- [x] Handle missing "KeyDescription" in the LockedPools prop.
- [x] Support both 3.0 and 2.4 at the same time.

Developed and tested with https://github.com/stratis-storage/stratisd/pull/2817.


mvollmer:
This should now be "complete", but not yet "done".

mvollmer:
I did an ad-hoc integration test run with a locally compiled version of stratisd from https://github.com/stratis-storage/stratisd/pull/2817/commits/2a9dc259370b8d23eecb8c5b10971226ea02d153.

All tests in check-storage-stratis pass, except TestStorageStratis.testCli. That fails because I didn't update `/usr/bin/stratis` to a version that works with stratisd 3.

This is the script that I used to install the locally compiled stratisd:

```
STRATIS=/home/mvo/work/rust/stratisd

./bots/image-customize -u $STRATIS/target/debug/stratisd:/var/tmp \
                       -u $STRATIS/stratisd.conf:/usr/share/dbus-1/system.d \
                       -u $STRATIS/systemd/stratisd.service:/usr/lib/systemd/system \
                       -r "mv /var/tmp/stratisd /usr/libexec/; restorecon /usr/libexec/stratisd" \
                       -v fedora-34
```

I also fixed the "missing --prompt" and "KillMode=process" bugs manually.

jbaublitz:
@mvollmer That's great! The prompt parameter was changed simply because it has been a dead parameter for a while now, but we left it around for compatibility until we could make breaking changes. The behavior doesn't change at all.

@mulkieran Should we put out one more patch release with the fix for the systemd unit file if @mvollmer is going to support stratisd 2.4.x?

pkg/storaged/client.js
```diff
@@ -900,21 +903,122 @@ client.vdo_overlay = vdo_overlay();
 /* Stratis */
 
 client.stratis_start = () => {
+    return stratis2_start()
+            .catch(error => {
+                if (error.problem == "not-found")
+                    return stratis3_start();
```
martinpitt, line 909: At some point we want to switch this around, but fine for now.

pkg/storaged/client.js
```diff
@@ -900,21 +903,122 @@ client.vdo_overlay = vdo_overlay();
...
     client.stratis_pools = { };
     client.stratis_blockdevs = { };
     client.stratis_filesystems = { };
-    client.stratis_manager.data.LockedPoolsWithDevs = [];
-    client.stratis_manager.data.KeyList = [];
+    client.stratis_manager.LockedPools = {};
 
     return client.stratis_manager.wait()
             .then(() => {
+                client.features.stratis_found = true;
```
martinpitt, line 986: You may want to move this a bit below, where it sets `client.features.stratis = true`. That matches the stratis3 variant then. I was rather confused why this method doesn't set the flag.

pkg/storaged/client.js
```diff
@@ -969,42 +1074,56 @@ client.stratis_start = () => {
...
                 });
             });
-};
+}
 
 function stratis_fetch_properties(proxy, props) {
     const stratis = client.stratis_manager.client;
 
-    stratis.call(proxy.path, "org.storage.stratis2.FetchProperties.r4", "GetProperties", [props])
+    return stratis.call(proxy.path, "org.storage.stratis2.FetchProperties.r4", "GetProperties", [props])
```
martinpitt, line 1082: This most certainly doesn't work with stratis3, and https://github.com/stratis-storage/stratisd/pull/2817 seems to even get rid of this custom property API (yay!). How can this not check for stratis3? This function is used in a lot of places, are they all called with stratis2 only, and stratis3 somehow does not need all this shuffling? If so, please rename them to `stratis2_*`.

pkg/storaged/client.js
```diff
@@ -969,42 +1074,56 @@ client.stratis_start = () => {
...
-                        }
-                    }
+                    if (result[p] && result[p][0])
+                        values[p] = result[p][1].v;
                 }
-                // Be careful not to add properties to a proxy that
-                // doesn't exist anymore.  That would bring it back.
-                if (changed && (proxy == client.stratis_manager ||
-                                client.stratis_pools[proxy.path] ||
-                                client.stratis_filesystems[proxy.path]))
```
martinpitt, line 994: The logic here became significantly simpler, and the commit log doesn't explain why this "be careful.." is not required any more. I mostly just want to make sure that you did not forget to put this back in while working on stratis3 support?

martinpitt (commented):
Thanks! That's quite a lot of data structure shuffling in client.js, but by my uninitiated first look it almost feels "too easy"? I have a few questions.

Nevertheless, nice work! Supporting two rather very different APIs is looking a lot less horrible than I feared :grin:

mvollmer:
> That's quite a lot of data structure shuffling in client.js, but by my uninitiated first look it almost feels "too easy"?

The main change from 2.4 to 3.0 for us is the replacement of the FetchProperties interface with regular D-Bus properties.  The data structure shuffling you see is the code that translates the old FetchProperties to the new D-Bus properties. That way, the actual "business logic" code can be written as if it is already talking to the 3.0 API.

So, yes, it's actually quite easy to translate from 2.4 to 3.0. Devil is in the details, but conceptually this is pretty clean.

pkg/storaged/client.js
mvollmer: Yes, agreed.

pkg/storaged/client.js
mvollmer: stratis3 and stratis2 are not the same in this regard.  Stratis2 might set `feature.stratis_found` to true but never set `feature.stratis` (when the version check fails).

We don't really need `stratis_found` at all, probably. It only matters when people install a stratisd version < 2.4 on a OS where we do on-demand installation of Stratis. All these OSes (fedora and centos) have stratisd 2.4.2, so one really has to get creative to cause `stratis_found` to make a difference.

It was important in #16525, where it would make life sane for people that have Stratis 3 installed.

But it seems to be too complex now, so I'll remove it, ok?

pkg/storaged/client.js
mvollmer: > If so, please rename them to stratis2_*.

Yes, that's a good point.  All this code is for Stratis 2 only. It will all go away with Stratis 3, and that's the reason why I have advocated to get rid of FetchProperties. I am very happy that it is happening. :-)

pkg/storaged/client.js
mvollmer: It is not forgotten, and thanks for bringing this up.  We needed to be careful because we were using dbus.notify.  That function will create a new proxy when it gets properties for a path that isn't in the cache (as it should).

The new code just writes to the JavaScript objects directly and triggers the client update process directly. If we do that with a object that is not longer in the cache, it's a pointless activity, but it wont add it back to the cache.

I'll add this to the commit message.

pkg/storaged/client.js
martinpitt: Yay, happy to hear! So yes, let's rename all of these to `stratis2_*` then

pkg/storaged/client.js
mvollmer: Done.

pkg/storaged/client.js
mvollmer: Done.

pkg/storaged/client.js
mvollmer: Done.

martinpitt (approved):
Nice, this is really rather elegant now! :clap:

PR #16541, opened 2021-10-29
systemd: Add "performance" as metrics keyword

jelly:


martinpitt:
@mvollmer: FYI, [another instance](https://logs.cockpit-project.org/logs/pull-16541-20211029-152146-75810ec5-ubuntu-stable/log.html#186) of that d-bus signal confusion. It's not as rare as we hoped :cry:

PR #16545, opened 2021-11-01
services: show user owned systemd units in the services page

KKoukiou:
## services: Show user owned systemd units

Most users are commonly familiar with the system systemd units. However, systemd offers the ability for unprivileged users to control and configure per-user units. Such units can now be managed from web console UI, the same way system-wide units are. 

Screen Shot 2021-11-05 at 09 51 35

### DIFF
--- a/pkg/systemd/services/service-tabs.jsx
+++ b/pkg/systemd/services/service-tabs.jsx
@@ -36,7 +36,7 @@ export const service_tabs_suffixes = ["service", "target", "socket", "timer", "p
  */
 export function ServiceTabs({ onChange, activeTab, tabErrors }) {
     const service_tabs = {
-        service: _("System services"),
+        service: _("Services"),
         target: _("Targets"),
         socket: _("Sockets"),
         timer: _("Timers"),
--- a/pkg/systemd/services/service.jsx
+++ b/pkg/systemd/services/service.jsx
@@ -58,6 +58,7 @@ export class Service extends React.Component {
             return null;
 
         const serviceDetails = <ServiceDetails unit={this.props.unit}
+                                owner={this.props.owner}
                                 permitted={superuser.allowed}
                                 loadingUnits={this.props.loadingUnits}
                                 isValid={this.props.unitIsValid} />;
@@ -76,7 +77,7 @@ export class Service extends React.Component {
                   id="service-details"
                   breadcrumb={
                       <Breadcrumb>
-                          <BreadcrumbItem to='#'>{_("Services")}</BreadcrumbItem>
+                          <BreadcrumbItem to={"#" + cockpit.location.href.replace(cockpit.location.path[0], '')}>{_("Services")}</BreadcrumbItem>
                           <BreadcrumbItem isActive>
                               {this.props.unit.Id}
                           </BreadcrumbItem>
@@ -84,7 +85,7 @@ export class Service extends React.Component {
                 <PageSection>
                     <Gallery hasGutter>
                         <GalleryItem>{serviceDetails}</GalleryItem>
-                        {(this.props.unit.LoadState === "loaded" || this.props.unit.LoadState === "masked") &&
+                        {((this.props.unit.LoadState === "loaded" || this.props.unit.LoadState === "masked") && this.props.owner == "system") &&
                         <GalleryItem>
                             <LogsPanel title={_("Service logs")} match={match} emptyMessage={_("No log entries")} max={10} goto_url={url} search_options={{ prio: "debug", service: cur_unit_id }} />
                         </GalleryItem>}
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -20,15 +20,17 @@
 import '../../lib/patternfly/patternfly-4-cockpit.scss';
 import 'polyfills'; // once per application
 
-import React, { useState } from "react";
+import React, { useState, useEffect, useCallback } from "react";
 import ReactDOM from 'react-dom';
 import {
     Button,
     Bullseye,
+    Flex, FlexItem,
     Select, SelectVariant, SelectOption,
     Page, PageSection, PageSectionVariants,
     Card,
     SearchInput,
+    ToggleGroup, ToggleGroupItem,
     Toolbar,
     ToolbarContent,
     ToolbarItem,
@@ -55,7 +57,10 @@ const _ = cockpit.gettext;
 //
 superuser.reload_page_on_change();
 
-export const systemd_client = cockpit.dbus("org.freedesktop.systemd1", { superuser: "try" });
+export const systemd_client = {
+    system: cockpit.dbus("org.freedesktop.systemd1", { bus: "system", superuser: "try" }),
+    user: cockpit.dbus("org.freedesktop.systemd1", { bus: "session" }),
+};
 const timedate_client = cockpit.dbus('org.freedesktop.timedate1');
 export let clock_realtime_now;
 export let clock_monotonic_now;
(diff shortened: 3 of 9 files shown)


martinpitt:
Wow, nice one! I tested this on my system, and I see my user units, can start them, etc.

But I found two issues which I believe are blockers:

 - None of my running user units seem to have logs, e.g.:
image
   However, `journalctl --user -u syncthing -b` does have a lot of entries.

 - When not having admin mode, I still can't control user units:
image

    The `superuser` check needs to be disabled for these.

martinpitt:
The first two commits look nice, and should hopefully not break anything -- feel free to submit them as a separate PR to get them tested and out of the way.

martinpitt:
>     * None of my running user units seem to have logs, e.g.:

I now saw that you explicitly mentioned that in the commit log. Then let's just hide the Logs panel entirely, instead of showing "No logs".

pkg/systemd/services/service-tabs.jsx
```diff
@@ -36,7 +36,7 @@ export const service_tabs_suffixes = ["service", "target", "socket", "timer", "p
  */
 export function ServiceTabs({ onChange, activeTab, tabErrors }) {
     const service_tabs = {
-        service: _("System services"),
+        service: _("Services"),
```
martinpitt, line 39: :+1: , what were we thinking :grin:

pkg/systemd/services/service.jsx
```diff
@@ -76,15 +77,15 @@ export class Service extends React.Component {
...
+                          <BreadcrumbItem to={"#" + cockpit.location.href.replace(cockpit.location.path[0], '')}>{_("Services")}</BreadcrumbItem>
                           <BreadcrumbItem isActive>
                               {this.props.unit.Id}
                           </BreadcrumbItem>
                       </Breadcrumb>}>
                 <PageSection>
                     <Gallery hasGutter>
                         <GalleryItem>{serviceDetails}</GalleryItem>
-                        {(this.props.unit.LoadState === "loaded" || this.props.unit.LoadState === "masked") &&
+                        {(this.props.unit.LoadState === "loaded" || this.props.unit.LoadState === "masked" || this.props.owner == "system") &&
```
martinpitt, line 88: This system check needs to become a `&&` to hide it for user units, I suppose?

test/verify/check-system-services
```diff
@@ -40,6 +40,7 @@ class TestServices(MachineCase):
         # Also `reset-failed` as failed units can be still listed with `systemctl` even when all files
         # were removed and daemon reloaded.
         self.restore_dir("/etc/systemd/system", "systemctl daemon-reload && systemctl reset-failed", True)
+        self.restore_dir("/etc/systemd/user", "su - admin -c 'export XDG_RUNTIME_DIR=/run/user/$(id -u admin); systemctl --user daemon-reload && systemctl reset-failed'", True)
```
martinpitt, line 43: The `reset-failed` is missing `--user`, that's what's breaking the tests.

pkg/systemd/services/services.jsx
```diff
@@ -50,12 +52,10 @@ import { useEvent, usePageLocation } from "hooks";
...
 
-// As long as we have long-running superuser channels, we need to
-// reload the page when the access level changes.
-//
-superuser.reload_page_on_change();
-
-export const systemd_client = cockpit.dbus("org.freedesktop.systemd1", { superuser: "try" });
+export const systemd_client = {
+    system: cockpit.dbus("org.freedesktop.systemd1", { bus: "system", superuser: "try" }),
+    session: cockpit.dbus("org.freedesktop.systemd1", { bus: "session" }),
```
martinpitt, line 57: Let's please call it `user`. These days it's really the user's bus/systemd instance, it gets shared across all the user's sessions. This also reflects `systemctl --user`.

test/verify/check-system-services
```diff
@@ -187,29 +196,38 @@ OnBootSec=200min
...
+        def run_cmd(cmd):
+            if user:
+                m.execute(f"su - admin -c 'export XDG_RUNTIME_DIR=/run/user/$(id -u admin); {cmd}'")
+            else:
+                m.execute(cmd)
+
+        self.make_test_service(path)
+        self.machine.execute(f"systemctl {params} start default.target")
+
+        url = "/system/services#/?owner=session" if user else "/system/services"
```
martinpitt, line 210: This should be `owner=user` (I made the same comment for the code part now)

test/verify/check-system-services
```diff
@@ -218,13 +236,14 @@ Unit=test.service
...
         self.wait_service_state("basic.target", "active")
 
-        # Make sure that symlinked services also appear in the list
-        self.wait_service_state("reboot.target", "inactive")
-        self.wait_service_state("ctrl-alt-del.target", "inactive")
-        self.wait_service_state("runlevel6.target", "inactive")
-        self.goto_service("runlevel6.target")
-        b.wait_in_text(".service-name", "Reboot")
-        b.click("#services-page .pf-c-breadcrumb__link")
+        if not user:
```
martinpitt, line 239: Let's also test some user units -- `dbus.service` should be a pretty safe bet. Also, the user unit has systemd's standard targets, like `basic.target` and `default.target`. The latter two also exist on the system instance, so they can be tested for both values of `user`.

test/verify/check-system-services
```diff
@@ -218,13 +236,14 @@ Unit=test.service
...
+            # Make sure that symlinked services also appear in the list
+            self.wait_service_state("reboot.target", "inactive")
+            self.wait_service_state("ctrl-alt-del.target", "inactive")
+            self.wait_service_state("runlevel6.target", "inactive")
+            self.goto_service("runlevel6.target")
```
martinpitt, line 244: FWIW, let's remove them. They are an ancient SysV init compatibility shim, and nobody cares about these any more.

martinpitt (changes requested):
Cheers!

test/verify/check-system-services
KKoukiou: oh awesome you saw this, I was really wondering.

garrett:
Oh, this is a welcome surprise! It's super neat to see it!

test/verify/check-system-services
KKoukiou: These checks where about to check symlinked units. These three uints are essentially the same file. I was not able to find symlinked units for system.

martinpitt:
I found one regression:
 - Login as unprivileged user, or drop privileges and force-reload the page
 - Raise privileges
 - Try to start (or enable/disable) a unit → "permission denied" error. This works on main, and apparently we don't have a test for privilege changes on the Services page.

pkg/systemd/services/services.jsx
```diff
@@ -50,12 +52,10 @@ import { useEvent, usePageLocation } from "hooks";
 
 const _ = cockpit.gettext;
 
-// As long as we have long-running superuser channels, we need to
-// reload the page when the access level changes.
-//
-superuser.reload_page_on_change();
```
martinpitt, line 55: Let's put that back for now, to fix the regression I mentioned above.

martinpitt (changes requested):
Thanks! LGTM, other than the superuser change.

pkg/systemd/services/services.jsx
martinpitt: Ah, you pushed before I finished reviewing :+1:

martinpitt (approved):

KKoukiou:
Sorry - I managed to make a typo in the single added line :)

martinpitt (approved):
Hehe - sorry, I didn't spot that either.

martinpitt:
@KKoukiou : I updated the release note a bit, let me know what you think.

PR #16547, opened 2021-11-02
tools: Add Arch Linux packaging files

jelly:


jelly:
One issue is that shadow fails:
```
Nov 03 11:48:55 archlinux sh[241]: user 'builder': directory '/home/builder' does not exist
Nov 03 11:48:55 archlinux sh[241]: pwck: no changes
```

tools/arch/cockpit-wsinstance.sysuser.conf
```diff
@@ -0,0 +1 @@
+u cockpit-wsinstance - "User for cockpit-ws instances"
```
martinpitt, line 1: Is there some Arch policy that they have to be separate files? The systemd-sysusers file format allows multiple lines, and these certainly belong together.

jelly: There is no policy for those files, but I guess the maintainer found it more clear to have a separate file per user.

test/verify/check-system-services
```diff
@@ -221,8 +221,10 @@ Unit=test.service
         # Make sure that symlinked services also appear in the list
         self.wait_service_state("reboot.target", "inactive")
         self.wait_service_state("ctrl-alt-del.target", "inactive")
-        self.wait_service_state("runlevel6.target", "inactive")
-        self.goto_service("runlevel6.target")
+        # Arch Linux has no SYSV_COMPAT runlevels https://github.com/systemd/systemd/blob/eaaa02bffe3b49504b58fad1980ac22eef3002cd/units/meson.build#L3
+        runlevel = "reboot.target" if m.image == "arch" else "runlevel6.target"
+        self.wait_service_state(runlevel, "inactive")
+        self.goto_service(runlevel)
```
martinpitt, line 227: Incidentally, @KKoukiou just cleaned that up in PR #16545 . Funny how things play out, I said "let's remove it, although it won't get in the way anytime soon".. and here we are!

jelly: Cool, I rebased and dropped it.

test/common/testlib.py
```diff
@@ -1022,7 +1022,8 @@ def enable_preload(self, package, *pages):
     def system_before(self, version):
         try:
             v = self.machine.execute("""rpm -q --qf '%{V}' cockpit-system ||
-                                        dpkg-query -W -f '${source:Upstream-Version}' cockpit-system
+                                        dpkg-query -W -f '${source:Upstream-Version}' cockpit-system ||
+                                        pacman -Q cockpit | awk '{ print $2 }'
```
martinpitt, line 1026: This is not quite right -- you need to strip off the distro version ("-1") as well. I independently did that this morning in https://github.com/cockpit-project/cockpit/pull/16662 as this is blocking image refreshes. After that lands, you can remove this commit.

martinpitt:
Looking at the [TestAccounts.testUnprivileged failure](https://logs.cockpit-project.org/logs/pull-16547-20211117-164214-28aaf144-arch-bots-2619/log.html#294-2): I can't reproduce this locally with just a naïve
```
test/image-prepare -q arch
bots/vm-run arch
TEST_OS=arch test/verify/check-users TestAccounts.testUnprivileged $RUNC -stv
```
So either this got fixed in the meantime (the recent image refresh?) or some earlier test has a lasting side effect on the VM.

I also tried all 5 tests in

    TEST_OS=arch test/common/run-tests --test-dir=test/verify -n TestAccounts

This fails testBasic due to `self.assertEqual(shell, '/bin/true')` (shell is /bin/bash), but that's rather obvious; CI [fails the same way](https://logs.cockpit-project.org/logs/pull-16547-20211202-120251-a012745c-arch/log.html#252)

Then `testRootLogin` fails with a page load timeout, again same as in CI. After that it's just stuck for me, it doesn't get any further.

And indeed I can reproduce this without the run-tests driver just fine:
```
TEST_OS=arch test/verify/check-users TestAccounts.testBasic $RUNC
# this fails as above
TEST_OS=arch test/verify/check-users TestAccounts.testUnprivileged $RUNC
# this fails with "permission denied", and hangs
```
So testBasic is destroying the VM.

test/verify/check-static-login
```diff
@@ -414,14 +414,19 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
...
+            http_auth_denied_code = 403
+
         self.assertEqual(self.curl_auth_code('/cockpit/login', ''), 401)
         self.assertEqual(self.curl_auth_code('/cockpit/login', 'foo:'), 401)
         self.assertEqual(self.curl_auth_code('/cockpit/login', 'foo:bar\n'), 401)
         self.assertEqual(self.curl_auth_code('/cockpit/login', 'foo:bar:baz'), 401)
         self.assertEqual(self.curl_auth_code('/cockpit/login', ':\n\n'), 401)
-        self.assertEqual(self.curl_auth_code('/cockpit/login', 'admin:bar'), 401)
+        self.assertEqual(self.curl_auth_code('/cockpit/login', 'admin:bar'), http_auth_denied_code)
         self.assertEqual(self.curl_auth_code('/cockpit/login', 'foo:bar'), 401)
```
martinpitt, line 436: Similarly to the previous comment on an earlier commit, using `assertIn(..., [401, 403])` seems perfectly fine here. Also, these two changes should be in the same commit.

test/verify/check-storage-resize
```diff
@@ -159,7 +159,7 @@ class TestStorageResize(StorageCase):
         # and unmount the fs explicitly.  But then we also need to
         # check it explicitly.
         #
-        self.machine.execute(f"umount '{fs_dev}' && fsadm -y check '{fs_dev}' && fsadm -y resize '{fs_dev}' '{size}'")
+        self.machine.execute(f"umount '{fs_dev}' || true && fsadm -y check '{fs_dev}' && fsadm -y resize '{fs_dev}' '{size}'")
```
martinpitt, line 163: I don't like `|| true` in tests, it makes bugs too hard to see/debug. Please rather use `(! mountpoint {fs_dev} || umount {fs_dev}) && ...` so that we don't cover up a *failed* unmount.

test/verify/check-metrics
```diff
@@ -516,6 +516,7 @@ class TestHistoryMetrics(MachineCase):
         self.assertEqual(m.execute("systemctl is-enabled pmlogger").strip(), "enabled")
 
     @nondestructive
+    @skipImage("pm proxy alert doesn't show on Arch Linux", "arch")
```
martinpitt, line 519: missing TODO: - let's keep these grep'able

martinpitt (changes requested):
Thanks! A bunch of small cleanups, but let's land this ASAP.

test/verify/check-pages
```diff
@@ -118,8 +118,10 @@ OnCalendar=daily
         b.switch_to_top()
         self.checkDocs(["Managing services"])
         b.click("#toggle-docs")
-        b.wait_visible('#toggle-docs + ul a:contains("Managing services")[href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/managing-services-in-the-web-console_system-management-using-the-rhel-8-web-console"]')
-        b.wait_visible('#toggle-docs + ul a:contains("Web Console")[href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/index"]')
+        b.wait_visible(
+            '#toggle-docs + ul a:contains("Managing services")[href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/managing-services-in-the-web-console_system-management-using-the-rhel-8-web-console"]')
+        b.wait_visible(
+            '#toggle-docs + ul a:contains("Web Console")[href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/index"]')
```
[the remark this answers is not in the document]
jelly: sorry, that's my editor which does autopep8

tools/arch/PKGBUILD
```diff
@@ -0,0 +1,119 @@
+# Maintainer: Massimiliano Torromeo <massimiliano.torromeo@gmail.com>
+# Contributor: Iwan Timmer <irtimmer@gmail.com>
+# Contributor: Mark Constable <markc@renta.net>
+# Contributor: Anatol Pomozov <anatol.pomozov@gmail.com>
+
+pkgbase=cockpit
+pkgname=(cockpit cockpit-pcp cockpit-test)
+pkgver=VERSION
+pkgrel=1
+pkgdesc='A systemd web based user interface for Linux servers'
```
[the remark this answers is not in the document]
jelly: Ok, noted.

tools/arch/PKGBUILD
```diff
@@ -0,0 +1,119 @@
...
+        "cockpit-wsinstance.sysuser.conf")
+sha256sums=('SKIP'
+            'a979e236681c6a06906937cf0f012e976347af5d6d7e7ae04a11acb01cc2689d'
+            '1ad9dad75858264778bd94799b60c651f7cc1c7f7fa1c54622174303e639287a'
+            '46ee8ecad7bc97ba588ab9471dde76e41c00daf40658902425626c3a1938b438')
+
+prepare() {
+  cd cockpit-$pkgver
+  # TODO: disable buggy package-lock check
+  sed -r '/^cmd_make_package_lock_json\b/ a exit 0' -i tools/node-modules
```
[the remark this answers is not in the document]
jelly: Not sure, it comes from the upstream packaging, we can just drop it.

tools/arch/PKGBUILD
```diff
@@ -0,0 +1,119 @@
...
+              "packagekit: manage packages")
+
+  cd cockpit-$pkgver
+  make DESTDIR="$pkgdir" install
+  rm -rf "$pkgdir"/usr/{src,lib/firewalld}
+  install -Dm644 "$srcdir"/cockpit.pam "$pkgdir"/etc/pam.d/cockpit
+  install -Dm644 "$srcdir"/cockpit-ws.sysuser.conf "$pkgdir"/usr/lib/sysusers.d/cockpit-ws.conf
+  install -Dm644 "$srcdir"/cockpit-wsinstance.sysuser.conf "$pkgdir"/usr/lib/sysusers.d/cockpit-wsinstance.conf
+
+  echo "z /usr/lib/cockpit/cockpit-session - - cockpit-wsinstance -" >> "$pkgdir"/usr/lib/tmpfiles.d/cockpit-tempfiles.conf
```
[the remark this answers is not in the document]
jelly: `ccockpit-ws` is indeed a sysuser so it's uid is not known upfront, so that's likely why it's like it. Shall we convert these comments into an upstream ticket?

tools/arch/PKGBUILD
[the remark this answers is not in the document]
martinpitt: No, please keep the PKGBUILD exactly as it is downstream -- these were indeed comments that apply to arch downstream, and should be handled there. Of course, it is right if we temporarily try to drop these hacks and see what breaks (yay CI :grin: ), and then fix it *properly* upstream.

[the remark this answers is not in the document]
martinpitt: In other words, these were just unrelated comments, these obviously don't block this PR.

tools/arch/PKGBUILD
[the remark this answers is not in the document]
martinpitt: With my Debian and now Fedora packaging background this looks really strange -- but I am not familiar with Arch packaging guidelines/best practices -- perhaps that just is the sanctioned approach. So I'll leave that up to you.

test/verify/check-connection
```diff
@@ -700,7 +706,11 @@ class TestConnection(MachineCase):
         # login handler: wrong password
         headers = m.execute("curl -s --head --header 'Authorization: Basic {}' http://172.27.0.15:9090/cockpit/login".format(
             base64.b64encode(b"admin:hahawrong").decode()))
-        self.assertIn("HTTP/1.1 401 Authentication failed\r\n", headers)
+        # On Arch Linux PAM errors are treated a bit different.
+        if m.image == "arch":
+            self.assertIn("HTTP/1.1 403 Permission denied\r\n", headers)
+        else:
+            self.assertIn("HTTP/1.1 401 Authentication failed\r\n", headers)
```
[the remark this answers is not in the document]
jelly: Ok, splitting and fixing.

test/verify/check-packagekit
```diff
@@ -577,6 +581,7 @@ ExecStart=/usr/local/bin/{self.packageName}
 
         self.allow_restart_journal_messages()
 
+    @skipImage("Arch Linux does not start services by default", "arch")
```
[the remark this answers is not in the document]
jelly: Arch actually has tracer as a package and it is added on the image, there was some issue with the test.

test/verify/check-storage-nfs
```diff
@@ -312,7 +312,7 @@ class TestStoragePackagesNFS(PackageCase, StorageCase, StorageHelpers):
 
         # HACK
         # Packagekit on Arch Linux does not deal well with detecting new repositories
-        if "arch" == m.image:
+        if m.image == "arch":
```
martinpitt: Rebase error -- this belongs into the first commit.

martinpitt (changes requested):
Thanks! I triggered 3 tests before the review.

There's just two tiny issues left. But mainly, this is intrusive enough so that we have to run the full test suite. Please drop no-test, rebase against main, and let the tests run everywhere.

test/verify/check-metrics
```diff
@@ -687,7 +687,8 @@ class TestCurrentMetrics(MachineCase):
     def setUp(self):
         super().setUp()
         # packagekit/dnf often eats a lot of CPU; silence it to have better control over CPU usage
-        self.machine.execute("systemctl mask packagekit && killall -9 /usr/libexec/packagekitd && killall -9 dnf || true")
+        packagekitd = "/usr/lib/packagekitd" if self.machine.image == "arch" else "/usr/libexec/packagekitd"
```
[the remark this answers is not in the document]
jelly: This is how it is:
```
[jelle@t14s][~/projects/cockpit]%pacman -Ql packagekit| grep packagekitd
packagekit /usr/lib/packagekitd
```

martinpitt:
I dropped no-test, and force-pushed a rebase (no changes), to run all tests.

martinpitt (dismissed):
Thanks! LGTM now, let's land when green.

martinpitt:
This [breaks all apt package tests](https://logs.cockpit-project.org/logs/pull-16547-20211208-185123-b8bf7bd3-ubuntu-2004/log.html#287-2)

test/verify/packagelib.py
```diff
@@ -264,6 +375,20 @@ def enableRepo(self):
             # pid will not be present for rebooting tests
             self.addCleanup(self.machine.execute, "kill %i || true" % pid)
             self.machine.wait_for_cockpit_running(port=12345)  # wait for changelog HTTP server to start up
+        if self.backend == "alpm":
```
martinpitt, line 378: This needs to be "elif", otherwise the "else" branch (yum/dnf) is run for apt.

martinpitt: Fixed.

martinpitt (approved):

PR #16549, opened 2021-11-02
test: Fix unstable TestKeys.testPrivateKeys pixel test

martinpitt:
Unfocus the input line, to avoid the blinking cursor and the slightly
unpredictable rounded corners of the focus ring.

----

[example failure](https://logs.cockpit-project.org/logs/pull-16538-20211102-104210-427b29b1-fedora-34-mobile/pixeldiff.html#TestKeys-testPrivateKeys-ssh-keys-dialog-mobile), this has plagued us for a while

### DIFF
--- a/test/verify/check-shell-keys
+++ b/test/verify/check-shell-keys
@@ -302,6 +302,8 @@ class TestKeys(MachineCase):
         b.set_input_text("#id_dsa-new-password", "foobar")
         b.set_input_text("#id_dsa-confirm-password", "foobar")
 
+        # avoid blinking cursor and blurry focus ring
+        b.blur("#id_dsa-confirm-password")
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):

mvollmer:
Random idea: Put the call to `blur` into `set_input_text`.  That function is only using focus as an implementation detail and nobody should rely on it leaving focus on the DOM element. If some tests need that, we should change them to explicitly set the focus.

martinpitt:
@mvollmer Great idea! This also simplifies most of our tests. I just found two places which rely on set_input_text() *not* blurring, but that's easy enough to customize. I'll send a PR ASAP.

martinpitt:
@mvollmer: Done in #16551, let's see what breaks.

PR #16550, opened 2021-11-02
lib: use new PF4 table (ComposableTable) implementation for all Table variants

KKoukiou:
The ListingTable component has already been partially ported to the non legacy PF4
table implementation, move it now completely.

This change is just an rewrite of the ListingTable component and should
not anyhow impact the consumers, as the properties and functionality
were not affected.

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -17,18 +17,11 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useState } from 'react';
 import {
+    ExpandableRowContent,
     TableComposable, Thead, Tbody, Tr, Th, Td,
-    Table,
-    TableHeader,
-    TableBody,
-    headerCol,
-    RowWrapper,
     SortByDirection,
-    sortable,
-    expandable,
 } from '@patternfly/react-table';
 import {
     EmptyState, EmptyStateBody, EmptyStateSecondaryActions,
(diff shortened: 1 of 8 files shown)


pkg/lib/cockpit-components-table.jsx
```diff
@@ -63,325 +56,190 @@ import './cockpit-components-table.scss';
...
-}) => {
     return (
         <>
             {header}
-            <TableComposable {...props}>
+            <TableComposable {...extraProps} {...tableProps}>
                 {showHeader && <Thead>
                     <Tr>
+                        {isExpandable && <Th />}
+                        {onSelect && <Th />}
```
martinpitt, line 216: Are these empty table headers just "filling" to get the expected number of columns? But they are class-less, how do they actually render the expander sign/select checkbox?

martinpitt (approved):
This works great, thank you! Tab headers are finally tab-able now.

I have one question, but that's just for understanding.

I suggest to squash this on landing, as especially the third commit is really just a fixup for the first (which already updates pixel refs)

pkg/lib/cockpit-components-table.jsx
KKoukiou: When the table has checkboxes or expanders on rows, (first cell of row occupied) the header columns need to move 1 or 2 positions (when both expander and checkbox exist) right accordingly. I checked how they do it in the PF4 examples, and they just render empty th as we do here.

Similarly, if one row is expandable and the next row is not, (we do that in the storage page in content views), we need to move the first column of the non - expandable row one position right, so that it aligns with the first expandable row.

Hope this makes sense.

martinpitt: Oh right, these are the expanded rows. Thanks!

PR #16551, opened 2021-11-02
test: Generalize set_input_text() blurring

martinpitt:
Generalize the blurring of the input element after typing keys in
set_input_text(). We keep running into the same problems with pixel
tests, and if tests need a particular item to be focussed, they should
be explicit in that.

The time/date validation checks in check-system-shutdown-restart *are*
dependent on keeping the focus, so provide a flag to disable the blurring.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -345,7 +345,7 @@ class Browser:
         else:
             self.wait_text(f"{selector} .pf-c-select__toggle-text", value)
 
-    def set_input_text(self, selector, val, append=False, value_check=True):
+    def set_input_text(self, selector, val, append=False, value_check=True, blur=True):
         self.focus(selector)
         if not append:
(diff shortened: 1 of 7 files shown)


martinpitt:
The pixeldiff errors look reasonable IMHO, as expected they remove the input line focus ring and cursor. I'll update the references.

jelly:
Once this is merged, I think we can also resolve the blur in the cockpit-podman tests https://github.com/cockpit-project/cockpit-podman/blob/main/test/check-application#L1207

mvollmer:
Pixel test diffs: https://github.com/cockpit-project/pixel-test-reference/compare/0d7ddd9b33943a3eb47f4f3a1fa2e0ea30a91f80..b7232801b22b175854215dc0019497ca70df1c07

mvollmer (approved):

PR #16558, opened 2021-11-03
storage: Assorted fixes

mvollmer:


martinpitt (approved):
Cheers!

martinpitt:
[Glitch here](https://logs.cockpit-project.org/logs/pull-16558-20211104-072329-f65a7194-centos-8-stream/log.html#256-1), "module scsi_debug is in use". This is from running the affected tests thrice -- if that's an obvious fix, maybe let's add it here?

mvollmer:
> [Glitch here](https://logs.cockpit-project.org/logs/pull-16558-20211104-072329-f65a7194-centos-8-stream/log.html#256-1), "module scsi_debug is in use". This is from running the affected tests thrice -- if that's an obvious fix, maybe let's add it here?

Hmm, the journal doesn't make a lot of sense... the test manages to create /dev/sda when it runs the second time, but then the journal just stops and there is no hint of running vgvreate...

martinpitt:
@mvollmer : That's usually a sign of some OOM condition, but then most often ssh also stops working, and the machine gets restarted. I see no hint of that in the log, so perhaps ssh actually survived. But this still wouldn't match the "scsi_debug already in use" error? Anyway, thanks for having a look! Apparently not something quick then.

mvollmer:
The testSnapshots failure has something to do with LVM2 and UDisks2. Sometimes UDisks2 misses the creation of the TEST volume group.  `udevadm trigger` puts things back in order.

mvollmer:
When it fails, information in udev is wrong (ID_FS_* props are missing). Running `udevadm trigger` will correct that, and will also put some messages into the journal from lvm itself, as if the vg had only now been created.  Hmm.  Running `lvcreate` will also fix the situation, probably because it triggers udev.

So once again udev turns out to be a unreliable source. I start wondering whether it is actually a supported technology.

Let's just ram another `udevadm trigger` into our tests and cross the fingers.

martinpitt:
@mvollmer: udev as in the narrow "userspace daemon" sense is actually fairly reliable -- I haven't ever observed it to just drop events. You can check with comparing kernel and udev events in `udevadm monitor`.

That sounds like a bug in the kernel's LVM driver then which is missing to send out the original kernel uevent, or possibly in its supporting helper programs (which probably occasionally also synthesize events from user space). It [would](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f6f09c15a767fad1206068bc09ec62d3cf273460) [not be](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9ec491447b90ad6a4056a9656b13f0b3a1e83043) the [first](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=75ab6afacda01a6bd2d3ecd4cb8485f7c8fa2fdb) [time](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c01a21b77722db0474bbcc4eafc8c4e0d8fed6d8)..

test/verify/check-storage-lvm2
```diff
@@ -283,6 +283,12 @@ class TestStorageLvm2(StorageCase):
         b.wait_in_text("#drives", disk)
 
         m.execute("vgcreate TEST %s" % disk)
+
+        # HACK - Sometimes the vgcreate call above doesn't update
+        #        udev.  Make sure it happens.
+        #
+        m.execute("udevadm trigger")
```
martinpitt, line 290: That is a really big and blunt hammer, and I wouldn't be suprised if that messes up other things in the VM. `--name-match {disk}` or at least `--subsystem-match block` would be more targetted. But I see that we commit this sin a lot in the tests, so it doesn't matter that much if we clean up 5 or 6 places.

martinpitt (approved):

test/verify/check-storage-lvm2
mvollmer: If we are worried about that, we should only trigger /dev/sda via `echo change >/sys/block/sda/uevent` or similar I guess.

mvollmer:
> udev as in the narrow "userspace daemon" sense is actually fairly reliable -- I haven't ever observed it to just drop events.

I meant "udev as a source of information" regardless of which component is ultimately to blame for the unreliability.

And it _does_ drop events when they come too quickly, and it doesn't follow its own locking protocol for device mapper devices. We just went through this for Stratis... :-)

PR #16559, opened 2021-11-03
storage: Unmount via dialog

mvollmer:
Unmounting takes a long time now because we need to reload
systemd (because we modify fstab).  And while a unmount operation is
running, we should block other operations that modify fstab.

### DIFF
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -166,7 +166,7 @@ td.job-action {
 
 .content-action {
     text-align: right;
-    width: 1% !important;
+    white-space: nowrap !important;
 }
 
 .dialog-item-tooltip {
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -195,7 +195,7 @@ tr[class*=content-level-] {
 
     // Add space for the button and offset
     > .pf-c-table__toggle + td {
-        padding-left: var(--offset);
+        padding-left: calc(var(--offset) + var(--pf-c-table--cell--PaddingLeft));
     }
 }
 
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -410,7 +410,7 @@ td.storage-action {
     max-width: 30ch;
 }
 
-#mounts .pf-c-progress__measure, #nfs-mounts .pf-c-progress__measure {
+#mounts .pf-c-progress__measure, #nfs-mounts .pf-c-progress__measure, #detail-content .pf-c-progress__measure {
     min-width: 7rem;
 }
 
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -243,6 +243,7 @@ td.storage-action {
 .units-table > tbody > tr > td:nth-child(1),
 .units-table > thead > tr > th:nth-child(1) {
     border-right: none;
+    padding-left: 0px;
 }
 
 .units-table > tbody > tr > td:nth-child(2),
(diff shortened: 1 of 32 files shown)


martinpitt:
How can I test this? I created a LUKS 2 file system on the `modprobe scsi_debug dev_size_mb=200` device, get it mounted, click "Unmount" in the kebab menu, and don't see a dialog.

mvollmer:
> How can I test this? I created a LUKS 2 file system on the `modprobe scsi_debug dev_size_mb=200` device, get it mounted, click "Unmount" in the kebab menu, and don't see a dialog.

That should do it... hmm.

mvollmer:
The dialog is pretty bare:

image

We could add a sentence that explains what will happen.  "Unmounting a filesystem makes its content unavailable to services and applications."

Eventually, there will be "teardown" information in there about services and sessions that keep this filesystem busy.

garrett:
Which filesystem is being unmounted? What device or RAID or whatever is it coming from? Is it currently in use? Is it dangerous to unmount the filesystem?

These are all questions I have that the dialog is not answering whatsoever.

garrett (changes requested):
Oh, related to the previous comment: Please answer these questions in the dialog.

(I forgot to comment about that with a request changes. Whoops!)

Ideally, the modal wouldn't block, but other operations wouldn't be available until the unmounting is done. But having a confirmation dialog on an unmount probably makes sense anyway, I guess... at least if it's currently in use. So having it pause for a few seconds as the unmount finishes isn't the worst thing.

mvollmer:
> Which filesystem is being unmounted?

Let's add that to the title.

> What device or RAID or whatever is it coming from?

I think this is clear from where the "Unmount" menu entry is placed, no?

> Is it currently in use? Is it dangerous to unmount the filesystem?

We don't have that information yet, unfortunately.  If it is in use, the unmount will fail with an error message.

garrett:
>>    What device or RAID or whatever is it coming from?
>
> I think this is clear from where the "Unmount" menu entry is placed, no?

It's not clear in the dialog and people do misclick, get interrupted in the middle of tasks, etc. Having the necessary information to perform a task self-contained within a dialog is a good thing. Even if someone remembers what they were doing, it's a bit of reassurance to make sure they'd doing the right thing, especially on a task that could cause issues.



>>    Is it currently in use? Is it dangerous to unmount the filesystem?
>
> We don't have that information yet, unfortunately. If it is in use, the unmount will fail with an error message.

If we don't know if it's hazardous, then it could be. 

It would be nice to check ahead of time and present that information, instead of resulting with an error.

mvollmer:
> It would be nice to check ahead of time and present that information, instead of resulting with an error.

Alright! :-)  I have started working on this, and while it needs some more polish code-wise, this was pretty easy to get working, actually.  Looks like this (exactly like the NFS unmount dialog):

image

garrett:
You need to have singular and plural wording (plus, I adjusted the text a little):

1. This filesystem is currently in use. Unmounting will stop all related processes.
2. This filesystem is currently in use. Unmounting will stop the related process.

Also, shouldn't process come first, before the session?

pkg/storaged/storage.scss
```diff
@@ -232,15 +232,17 @@ td.storage-action {
     overflow: visible !important;
 }
 
+.units-table {
+    width: 100%;
+}
```
garrett, line 237: We could use PF utils instead.

https://www.patternfly.org/v4/utilities/sizing

The class `pf-u-w-100` will make something 100% width. It also supports breakpoints and other sizing, but this is a 1:1 change.

To use it: Just make sure the utils CSS is imported and add the class to the element.

mvollmer:
This now also stops processes when doing teardown, and is getting a bit out of hand... I'll merge this with #16431.

PR #16563, opened 2021-11-04
networkmanager: Fix JavaScript error when applying settings

jelly:
show_modal_dialog expects an action to be a function which returns a
promise, as it didn't you would get an error that undefined has no
method 'then'. The issue was introduced in 3899b9b5d79df.

### DIFF
--- a/pkg/networkmanager/interfaces.js
+++ b/pkg/networkmanager/interfaces.js
@@ -1268,7 +1268,7 @@ export function syn_click(model, fun) {
     return function() {
         const self = this;
         const self_args = arguments;
-        model.synchronize().then(function() {
+        return model.synchronize().then(function() {
             fun.apply(self, self_args);
         });
     };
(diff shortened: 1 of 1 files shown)


pkg/networkmanager/interfaces.js
```diff
@@ -1556,11 +1556,9 @@ export function with_checkpoint(model, modify, options) {
...
                                             show_breaking_change_dialog({
                                                 ...options,
-                                                action: () => {
-                                                    syn_click(model, function () {
-                                                        modify();
-                                                    });
-                                                }
+                                                action: syn_click(model, function () {
+                                                    modify();
+                                                }),
```
martinpitt, line 1561: Just leaves me curious -- why not go all the way and do `action: syn_click(model, modify)`?

martinpitt (dismissed):
Nice catch!

Would be nice if you could rebase and also let the packit tests run, but not very important.

pkg/networkmanager/interfaces.js
jelly: Ooo, good point done!

martinpitt (approved):
Thanks! Our networking tests are notoriously unstable, retrying (3x due to affected here, hard to get green)

PR #16573, opened 2021-11-08
base1: tweak error message check in test-dbus.js

allisonkarlitskaya:
Fedora 35 now gives an error message about "Object does not exist" vs.
previous versions saying "No such interface".  Accept both messages.

We don't actually run unit tests in Fedora 35 at the moment (although perhaps we should) but I can confirm that this fixes `make check` for me locally.


martinpitt (approved):
Thanks! I can reproduce the failure and confirm it fixes it.

PR #16571, opened 2021-11-08
test: Ignore wrong cross-owner D-Bus errors

martinpitt:
Even after commit 11bea587f9 we still sometimes get error messages from
UDisks2 on the stratisd D-Bus proxy. This breaks tests randomly due to
unexpected messages:

    org.storage.stratis2: couldn't get managed objects at /org/freedesktop/UDisks2:
    GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable

----

This has become too annoying.. I see this once or twice a day, [today's instance](https://logs.cockpit-project.org/logs/pull-16570-20211108-054727-2feaa5d5-fedora-34-devel/log.html#213).

 - [x] See if PR #16574 fixes this properly, and if so, land that instead.

PR #16574, opened 2021-11-08
bridge: Block signals initially for the default CockpitDBusCache

mvollmer:
The CockpitDBusCache for the default name of a dbus channel is already
careful not to process signals from other sources than the default
name, but it starts doing that only as soon as the unique name is
known (or known to not exist).  Before that, it will accept all
signals, which might still lead to journal messages like

    org.storage.stratis3: couldn't get managed objects at /org/freedesktop/UDisks2: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.storage.stratis3 was not provided by any .service files

Thus, we block signals initially as well, until the unique name is
known.

### DIFF
--- a/src/bridge/cockpitdbusjson.c
+++ b/src/bridge/cockpitdbusjson.c
@@ -2792,6 +2792,13 @@ process_connection (CockpitDBusJson *self,
                                                                 self, NULL);
           self->default_watched = TRUE;
           subscribe_and_cache (self);
+
+          /* Stop the cache from processing signals until we know its
+             bus name.
+           */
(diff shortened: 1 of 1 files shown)


martinpitt:
If that works, it's certainly preferable to PR #16571, so let's wait for results here first. Thanks!

mvollmer:
> If that works,

Yeah, my main concern is that this breaks something...

martinpitt (approved):
Sweet, thank you! Let's run the tests one more round and then land.

martinpitt:
Green twice in a row :clap:

PR #16575, opened 2021-11-08
tools: Use our own SELinux policy on RHEL/CentOS 8 as well

martinpitt:
tools: Use our own SELinux policy on RHEL/CentOS 8 as well

This will fix the SELinux type of /etc/cockpit/ws-certs.d (from commit
6ec488185), and is going to fix systemd's cleanup of /run/cockpit in
PR #16572.

Drop the distribution test from the .spec, as this is actually also
being used in OpenSUSE. As this now builds on RHEL 8,  add a `name == "cockpit"`
guard, as we don't want to build this for cockpit-appstream.

This allows us to drop the fallback from cockpit-certificate-helper, as
certmonger can now finally write directly into ws-certs.d/ everywhere.
This will fix certificate auto-refresh on RHEL 8. Drop the corresponding
hacks/special cases from the integration tests.

-----

I asked @zpytela and @vmojzis  and got their blessing for this. It will also help with getting rid of that hack in linux-system-roles eventually: https://github.com/linux-system-roles/cockpit/#certificate-setup

 - [x] https://github.com/cockpit-project/bots/pull/2611

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -81,8 +81,8 @@ Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{v
 %define build_optional 1
 %endif
 
-# Ship custom SELinux policy only in Fedora and RHEL-9 onward
-%if 0%{?rhel} >= 9 || 0%{?fedora}
+# Ship custom SELinux policy (but not for cockpit-appstream)
+%if "%{name}" == "cockpit"
 %define selinuxtype targeted
 %define with_selinux 1
 %define selinux_policy_version %(rpm --quiet -q selinux-policy && rpm -q --queryformat "%{V}-%{R}" selinux-policy || echo 1)
--- a/src/tls/cockpit-certificate-helper.in
+++ b/src/tls/cockpit-certificate-helper.in
@@ -108,18 +108,8 @@ cmd_ipa_request() {
     klist -k "${KEYTAB}" | grep -qF "${SERVICE}" || \
         ipa-getkeytab -p "HTTP/${HOST}" -k "${KEYTAB}"
 
-    # request the certificate and put it into our certificate directory
-    if ! selinuxenabled || [ "$(stat -c %C /etc/cockpit/ws-certs.d/ | cut -f3 -d:)" = cert_t ]; then
-        # the prefered way is to let certmonger write the certificate directly, so that auto-refresh works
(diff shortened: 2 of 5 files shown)


tools/cockpit.spec
```diff
@@ -81,8 +81,8 @@ Source0:        https://github.com/cockpit-project/cockpit/releases/download/%{v
 %define build_optional 1
 %endif
 
-# Ship custom SELinux policy only in Fedora and RHEL-9 onward
-%if 0%{?rhel} >= 9 || 0%{?fedora}
+# Ship custom SELinux policy on Fedora and RHEL/CentOS
+%if 0%{?rhel} || 0%{?fedora}
```
allisonkarlitskaya, line 85: ...so everywhere, then?

Can we drop the `%if` entirely?  Or are you thinking about other distros?

martinpitt: The main distro which we support with our .spec is OpenSUSE. I have zero ideas how SELinux is being managed there, and we don't have it in CI so I don't dare to do this big change to it without at least consulting @nussel

lnussel: Thanks for the consideration! You really don't have to do that though. Just don't worry. If it breaks I'll fix it. As a matter of fact I've patched that %if condition to also be true on openSUSE. I couldn't send the patch yet as IIRC futher down the spec file expects setroubleshoot which is still stuck in the review queue.

martinpitt:
[This](https://logs.cockpit-project.org/logs/pull-16575-20211108-170318-869d8246-rhel-8-6-bots-2611/log.html#24-2) looks much better! :+1: These are expected failures, as the helper can now actually write directly into ws-certs.d/, and we can clean up that hack. So I'll get the image refreshes landed, and then clean this up. I suppose the image refreshes will also stumble over this, though.

tools/cockpit.spec
martinpitt: Thanks @lnussel, so much the better! Then I'll just drop the whole `%{with_selinux}` then, which should make the spec simpler.

martinpitt: Eh sorry, no -- we still have the cockpit-appstream hack for RHEL 8, so `%{with_selinux}` will have to stay for a while, I'm afraid. But at least the marked condition above can go.

martinpitt: Dropped.

martinpitt:
[This failure on Debian stable/testing](https://logs.cockpit-project.org/logs/pull-16575-20211109-061610-d536685e-debian-testing/log.html#283-2) comes rather out of the blue -- I can't reproduce it locally, and I don't know right now how it is related to the changes here. Back to Draft.. *sigh*

marusak (approved):
Nice, thank you!

PR #16576, opened 2021-11-08
apps: packagekit with apt backend reports backward percentages for th…

### LINKED ISSUE #16564: applications: Progress bar goes backwards
When checking for applications, progress bars bounce around. They should only move forwards.

tempsnip

_Originally posted by @csjjpm in https://github.com/cockpit-project/cockpit/issues/16484#issuecomment-947632500_

KKoukiou:
…e refresh operation

Guard these wrong values that the packagekit API returns by making sure
these are corresponding to an increasing percentage value.

Fixes #16564

### DIFF
--- a/pkg/apps/packagekit.js
+++ b/pkg/apps/packagekit.js
@@ -23,8 +23,11 @@ import * as PK from "packagekit.js";
 function progress_reporter(base, range, callback) {
     if (callback) {
         return function (data) {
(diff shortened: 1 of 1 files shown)


KKoukiou:
I think the while approach of writing back to the API provided value structure is wrong, I want to check it a bit more.

PR #16578, opened 2021-11-09
apps: fix packagekit progress bar backword percentages

### LINKED ISSUE #16564: applications: Progress bar goes backwards

KKoukiou:
Packagekit with apt backend reports backward percentages for the refresh
operation.

Guard these wrong values that the packagekit API returns by making sure
these are corresponding to an increasing percentage value.

Rewrite progress_reporter into a class and store the new percentage
value there as an attribute, instead of overwriting the object that the API returns, as
that it not public attribute.

Upstream bug report: https://github.com/PackageKit/PackageKit/issues/516

Fixes #16564

### DIFF
--- a/pkg/apps/packagekit.js
+++ b/pkg/apps/packagekit.js
@@ -20,13 +20,23 @@
 import cockpit from "cockpit";
 import * as PK from "packagekit.js";
 
-function progress_reporter(base, range, callback) {
-    if (callback) {
-        return function (data) {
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice, thank you!

PR #16580, opened 2021-11-09
storaged: Properly represent LVM VDO pools

martinpitt:
The `vdo cli` API and standalone "vdo" block device type has been deprecated a while ago, in favor of the LVM/udisks-lvm2 API. The latter is the officially documented one [1].

The official VDO API is a particular "VDO" type of LVM volume, similar
to a thin pool volume or a normal LVM block LV. These consist of two
LVs:

 (1) The physical "pool" volume which holds the compressed data; from
     the user's point of view these are an internal implementation detail
     and have no sensible actions other than growing. In particular,
     these should not be (de)activated [2]. The `lvcreate` CLI also does
     not require specifying them, it will just make up a new name.

 (2) The virtual block volume which gets formatted, mounted, and used.

Stop showing type (1) as standalone top-level LV volumes.  Instead, add a
"VDO Pool" tab to the type (2) volumes, show the pool size (including
Grow button) there, and also show information and toggles from the
`org.freedesktop.UDisks2.VDOVolume` interface.

The Storage page cannot yet create this kind of VDO volume, this only
fixes the presentation of existing ones.

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deduplicating_and_compressing_logical_volumes_on_rhel/creating-a-deduplicated-and-compressed-logical-volume_deduplicating-and-compressing-logical-volumes

[2] udisks currently also represents them wrongly: https://github.com/storaged-project/udisks/issues/934

----

On current main, these "new-style" VDO devices look like this; the pool is a top-level thing with a broken (or even dangerous?) Activate/Deactivate button (it times out eventually, but does change the status in LVM), it has no logical connection to the actually interesting volume, and does not show any interesting properties other than the size:

main

Now the pool is a tab inside the vdo0 volume ("Volume" tab is the same as before):

pr-vdo

The "Grow" buttons on both tabs work. I did not yet add a "Rename" button for the pool; it works, but it causes a page crash, and it's also not interesting. I don't even plan to expose the pool name on UI creation, it can just be `<volname>_pool`.

The "Filesystem" tab is also unchanged, but for reference:

pr-fs

 - [ ] Agree on design
 - [x] Use `lvcreate` instead of raw D-Bus call
 - [x] Show correct/precise physical size
 - [x] Add On/Off buttons for compression and deduplication
 - [x] Add missing tests for the actions (grow volume, grow pool)

Follow-ups:
 - [ ] Show overhead (Requires polling VDOVolume.GetStatistics)
 - [ ] Add creation dialog
 - [ ] Drop creation of old-style VDO volumes

### DIFF
--- a/pkg/storaged/client.js
+++ b/pkg/storaged/client.js
@@ -171,6 +171,7 @@ function init_proxies () {
     client.blocks_crypto = proxies("Encrypted");
     client.blocks_swap = proxies("Swapspace");
     client.iscsi_sessions = proxies("ISCSI.Session");
+    client.vdo_vols = proxies("VDOVolume");
     client.jobs = proxies("Job");
 
     return client.storaged_client.watch({ path_namespace: "/org/freedesktop/UDisks2" });
--- a/pkg/storaged/lvol-tabs.jsx
+++ b/pkg/storaged/lvol-tabs.jsx
@@ -29,7 +29,7 @@ import {
     DescriptionListDescription,
     Flex, FlexItem,
 } from "@patternfly/react-core";
-import { StorageButton, StorageLink } from "./storage-controls.jsx";
+import { StorageButton, StorageLink, StorageOnOff } from "./storage-controls.jsx";
 import {
     existing_passphrase_fields, get_existing_passphrase_for_dialog,
     request_passphrase_on_error_handler
@@ -457,13 +457,13 @@ export class BlockVolTab extends React.Component {
     }
 }
 
+function perc(ratio) {
+    return (ratio * 100).toFixed(0) + "%";
+}
+
 export class PoolVolTab extends React.Component {
     render() {
         const self = this;
-
-        function perc(ratio) {
-            return (ratio * 100).toFixed(0) + "%";
-        }
 
         function rename() {
             lvol_rename(self.props.lvol);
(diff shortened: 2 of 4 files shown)


martinpitt:
@mvollmer : I am interested in a first UX review from you. At first I thought we should make these very similar to the LVM Thin Pool thingies, but behaviourally they are really quite different -- a thin pool can have arbitrarily many (including zero) volumes, whereas a VDO volume always has *exactly* one backing pool, so that it's just an implementation detail. Of course I'll flesh this out a bit, but I'd like to have some agreement on the general style first.

After this lands, I'll work on creating these "new-style" VDO devices (what the D-Bus call in the test does now), and finally drop all the old-style VDO bits (that's going to include the Python notification helper etc.)

pkg/storaged/client.js
```diff
@@ -171,6 +171,7 @@ function init_proxies () {
     client.blocks_crypto = proxies("Encrypted");
     client.blocks_swap = proxies("Swapspace");
     client.iscsi_sessions = proxies("ISCSI.Session");
+    client.vdo_vol = proxies("VDOVolume");
```
mvollmer, line 174: The convention is to use plural here: `vdo_vols`.  Or `lvols_vdo` if this is an extension interface to LogicalVolume (and this should be moved next to `client.lvols = proxies("LogocalVolume")`.

mvollmer:
The old VDO support had some additional statistics and features:

image

(There are two files on the filesystem: 500 MiB zeroes and 500 MiB random data.)

It would tell you how much space was saved due to compression/deduplication.

It would tell you about the index memory.

Compression and deduplication could be switched on and off.

Maybe we want to add them back here already?

martinpitt:
> The old VDO support had some additional statistics and features:
> 
> (There are two files on the filesystem: 500 MiB zeroes and 500 MiB random data.)

```
mkfs.xfs /dev/vdo_vgroup/vdo0
mount /dev/vdo_vgroup/vdo0 /mnt/
dd if=/dev/zero of=/mnt/empty bs=1M count=2000
dd if=/dev/urandom of=/mnt/garbage bs=1M count=2000
```

> It would tell you how much space was saved due to compression/deduplication.

I have some trouble with this. Initially I get these numbers on the vdo0 LV:

```
org.freedesktop.UDisks2.LogicalVolume interface -         -            
.DataAllocatedRatio                   property  d         0                                        emits-change
.MetadataAllocatedRatio               property  d         0                                        emits-change
.Size                                 property  t         10003415040                              emits-change

org.freedesktop.UDisks2.VDOVolume     interface -  
.UsedSize                             property  t         3856158720                               emits-change

```

After the "empty" files above, the numbers don't change at all. After the "garbabe" file I get these changes:
```
.DataAllocatedRatio                   property  d         0.2                                      emits-change
.UsedSize                             property  t         5939347456                               emits-change
```

I've never seen `MetadataAllocatedRatio` be non-zero. `DataAllocatedRatio` and `.UsedSize` grow sensibly. But I can't tell how much was the "overhead" and which from the actual data blocks. But in the end that probably also doesn't matter?

So what we can show is the physically used size including overhead (VDOVolume.UsedSize) and the relative (LogicalVolume.DataAllocatedRatio) and absolute (LogicalVolume.DataAllocatedRatio * LogicalVolume.Size) sizes. Certainly also a percentage between the two, but it will look odd if there is relatively little data, and thus the overhead of the 4 GB metadata will greatly outweigh the saving. So the "50% saving" in the current info is also quite a cheat :grin: 

Until we make up our minds how to present that sensibly, how about I just show the two numbers above?

On the vdo0_pool device there are no size related properties other than the static `.Size 6002049024` (the physical total pool size).

`lvdisplay` has some info as well:
```
  LV Path                /dev/vdo_vgroup/vdo0_pool
  LV VDO Pool usage      99.29%
  LV VDO Pool saving     50.12%
  LV VDO Used size       5.55 GiB

 LV Path                /dev/vdo_vgroup/vdo0
  LV Size                <9.32 GiB
```
The "pool saving" is correct - half of the data in my two files are zero, which compress to nothing. The "pool usage 99%" does not make much sense to show, though -- as there is an 1:1 relationship between the physical pool an the virtual LV, that is always supposed to be ~ 100%.

> It would tell you about the index memory.

No idea what that is, there is zero information about it in either `lvdisplay` or the udisks properties. So I'd just leave it out for now?

> 
> Compression and deduplication could be switched on and off.

This works:
```
busctl call org.freedesktop.UDisks2 /org/freedesktop/UDisks2/lvm/vdo_vgroup/vdo0 org.freedesktop.UDisks2.VDOVolume EnableCompression 'ba{sv}' false 0
```
same for `true` and same for `EnableDeduplication` -- the values change instantly in the UI.

> Maybe we want to add them back here already?

Yes, I'm happy to do that -- I just wanted to post an initial sketch to know whether this is even the right direction.

pkg/storaged/client.js
martinpitt: Renamed to vdo_vols.

mvollmer:
> I have some trouble with [percent saved].

Yeah, fair enough.  The old code gets it directly from VDO.  If LVM2 doesn't give it directly, let's not bother trying to compute it ourselves.

martinpitt:
Huh, [how did I break the legacy test on centos-8-stream](https://logs.cockpit-project.org/logs/pull-16580-20211116-124018-ee39cd51-centos-8-stream/log.html#299) only.. Gotta investigate.

martinpitt:
Added a separate commit to fix the independent test flake from above. It hit so hard here because of "3x affected" testing.

pkg/storaged/lvol-tabs.jsx
```diff
@@ -508,3 +508,77 @@ export class PoolVolTab extends React.Component {
...
+    function grow() {
+        lvol_grow(client, vdo_pool_vol, { });
+    }
+
+    function toggle_compression() {
+        return new Promise((resolve, reject) => {
+            const new_state = !vdo_iface.Compression;
+            vdo_iface.EnableCompression(new_state, {})
+                    .catch(reject);
+            client.wait_for(() => vdo_iface.Compression === new_state)
```
mvollmer, line 525: What about chaining the waiting with the method call?

```
return vdo_iface.EnableCompression(new_state, {})
   .then(() => client.wait_for(() => vdo_iface.Compression === new_state));
```

client.wait_for returns a promise (that delivers the return value of the condition callback), so we can just chain these up.

pkg/storaged/lvol-tabs.jsx
```diff
@@ -508,3 +508,77 @@ export class PoolVolTab extends React.Component {
...
+                    .then(resolve);
+        });
+    }
+
+    function toggle_deduplication() {
+        return new Promise((resolve, reject) => {
+            const new_state = !vdo_iface.Deduplication;
+            vdo_iface.EnableDeduplication(new_state, {})
```
mvollmer, line 533: Same here.

mvollmer (changes requested):

pkg/storaged/lvol-tabs.jsx
martinpitt: I thought of that initially, but then was worried about race conditions -- that the property *might* already be correct after EnableCompression() returns, and then there would not be a direct property update signal afterwards any more.

However, that does look more elegant for sure, so I did that now. I ran it locally a few times (where I could hardly reproduce the original error that CI caught), and it seems fine. Let's see what the bots think.

martinpitt: Seems they are all happy :green_heart:

mvollmer: > I thought of that initially, but then was worried about race conditions -- that the property _might_ already be correct after EnableCompression() returns, and then there would not be a direct property update signal afterwards any more.

`wait_for` should handle that.  It checks the condition once immediately, and then on every change notification.

mvollmer (approved):

PR #16581, opened 2021-11-09
'Close' button in dialogs should be styled as secondary button

KKoukiou:


martinpitt (approved):
Thanks! I visually tested that with the "Keep connection" button on Networking, it really makes more sense that way.

PR #16587, opened 2021-11-10
Fix DatePicker's start of week for most cases

martinpitt:
PF's DatePicker defaults to Sunday as the start of the week instead of
the given locale's convention. There is currently no official JS API
which exposes that piece of locale information [1], and
`cockpit.language` does not even contain the country in most cases. So
add a `timefirmat.firstDayOfWeek()` approximation by language, which at
least gets it right for most of our supported languages *except*
English-speaking countries in Europe (in particular, en-GB and en-IE).

Add missing aria and localization to the timer creation and user account
expiration dialogs, so that their month names are properly translated.

[1] https://github.com/tc39/ecma402/issues/6

---

English looks as before:

date-picker-en

In German, the week now starts on Monday:

date-picker-de

### DIFF
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -197,6 +197,7 @@ export class ShutdownModal extends React.Component {
                                                 invalidFormatText=""
                                                 isDisabled={!this.state.formFilled}
                                                 locale={cockpit.language}
+                                                weekStart={timeformat.firstDayOfWeek()}
                                                 onBlur={this.calculate}
                                                 onChange={(d, ds) => this.updateDate(d, ds)}
                                                 placeholder={timeformat.dateShortFormat()}
--- a/pkg/lib/serverTime.js
+++ b/pkg/lib/serverTime.js
@@ -520,6 +520,7 @@ function ChangeSystimeBody({ state, errors, change }) {
                                     dateParse={timeformat.parseShortDate}
                                     invalidFormatText=""
                                     locale={cockpit.language}
+                                    weekStart={timeformat.firstDayOfWeek()}
                                     placeholder={timeformat.dateShortFormat()}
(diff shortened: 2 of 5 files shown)


garrett:
Is there a way to get the local computer's timezone? That would fix the approximation, right?

(The timezone in the screenshot is the remote system's timezone, of course.)

KKoukiou (approved):
I feel that weekStart should not be exposed property but rather set by PF according to the locale provided. However since it's missing from intl we can have this solution.

garrett:
For most of English and much of Spanish: if it's in the Americas and Austrailia, then it's Sundays first, if in Europe, it's Monday.

(English in South Africa is Sunday first.)

garrett:
FWIW, I'm consulting the map @ 
image
https://en.wikipedia.org/wiki/Week#/media/File:First_Day_of_Week_World_Map.svg

It's not perfect, but we could check offset for less than / greater than and make an exception for South Africa (SAST/EAT) if we know the timezone (instead of just the offset).

garrett:
Alternatively:

We could export the server's locale information.

And if browser time is the same as the system time (or peraps +- 3 timezones off... think servers in Pacific time and living in Eastern time in the US, or servers in France but living in Finland), use that to instruct the weekday start.

Any solution is a bit hacky though, if we don't have access to this information from the browser itself. (Unless we add this as a preference in the language selection dialog too. But we'd still want something that generally solves it without a setting, naturally.)

garrett (approved):
_If_ we decide to have a little more logic to get it closer to correct and/or add a preference in the language dialog, it would build upon this. So I'm happy that this PR makes it more correct for more people. It's definitely better than what we currently have.

Also, the first day of the week is not just Sunday or Monday, so this is incomplete. (But I guess for the languages we currently officially support, it's probably correct enough, except for European English and American Spanish.)

PR #16597, opened 2021-11-11
users: Clear password errors when password is changed

jelly:
When a user types a bad password and tries to set a weak password
errors are shown underneath the password field. Changing the password
then again does not clear the password.


martinpitt:
Nice, thanks! We already have a "password mismatch" test. Can you please make it type another character there and assert that the error message goes away?

martinpitt (dismissed):
Thanks!

martinpitt (approved):

PR #16600, opened 2021-11-12
test: Use add_loopback_disk() to create loopback disks

marusak:
This method has proper cleanup defined.

This is easily reproducible:
```diff
diff --git test/verify/check-storage-partitions test/verify/check-storage-partitions
index b0c783dd4..c58789744 100755
--- test/verify/check-storage-partitions
+++ test/verify/check-storage-partitions
@@ -58,6 +58,7 @@ class TestStoragePartitions(StorageCase):
         self.content_row_wait_in_col(1, 1, "ext4 file system")
 
         self.content_dropdown_action(1, "Delete")
+        b.wait_visible("#foobar")
         self.confirm()
         self.content_row_wait_in_col(1, 0, "Free space")
```

and then run `./test/common/run-tests --test-dir test/verify TestStoragePartitions.testPartitions`. Without this patch it fails for the first time on this new `wait_visible` and then quickly fails on `dev = m.execute("set -e; dd if=/dev/zero of=/var/tmp/disk1 bs=1M count=10; "` as seen [here](https://logs.cockpit-project.org/logs/pull-16572-20211110-131547-9d5ddabf-ubuntu-2004/log.html#186). With this patch it fails every time on the new `wait_visible`. Also works just fine 3 successful times in a row.

### DIFF
--- a/test/verify/check-storage-partitions
+++ b/test/verify/check-storage-partitions
@@ -26,7 +26,6 @@ from testlib import *
 class TestStoragePartitions(StorageCase):
 
     def testPartitions(self):
-        m = self.machine
         b = self.browser
 
         self.login_and_go("/storage")
@@ -41,10 +40,7 @@ class TestStoragePartitions(StorageCase):
         #
         # https://github.com/storaged-project/storaged/issues/97
 
-        dev = m.execute("set -e; dd if=/dev/zero of=/var/tmp/disk1 bs=1M count=10; "
-                        "losetup --show loop12 /var/tmp/disk1; "
-                        "rm /var/tmp/disk1").strip()
-        self.addCleanup(m.execute, "losetup -d " + dev)
+        dev = self.add_loopback_disk(10, "loop12")
 
         b.wait_visible('.sidepanel-row:contains("%s")' % dev)
         b.click('.sidepanel-row:contains("%s")' % dev)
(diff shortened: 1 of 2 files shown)


martinpitt (dismissed):
Hah, nice!

marusak:
[different flake](https://logs.cockpit-project.org/logs/pull-16600-20211112-081531-2c3f24ce-ubuntu-2004/log.html#299) I might take a look at this one as well

martinpitt:
Curious -- why is that test dependent on the "loop12" name? I don't see that name anywhere else. Did `--find` fail for some reason ?

marusak:
> Curious -- why is that test dependent on the "loop12" name? I don't see that name anywhere else. Did --find fail for some reason ?

Honestly, no idea. But there is this comment just above:
```py
        # A loopback device ends with a digit and partitions have               
        # names like "/dev/loop0p1".  Check that the storage stack has          
        # no difficulties with that.                                            
        #                                                                       
        # We are especially careful to use a device name that doesn't           
        # end in all zeros, because that would be too easy and                  
        # wouldn't trigger this bug:                                            
        #                                                                       
        # https://github.com/storaged-project/storaged/issues/97 
```

so maybe related? Now it went green just fine.

Now only [TestKdumpNFS.testBasic](https://logs.cockpit-project.org/logs/pull-16600-20211112-125650-5c03652c-centos-8-stream/log.html#48-2) flaked, which is one of or biggest flakes according to weather report. Unrelated of course.

martinpitt (approved):
Ack, thanks! That'll be a mystery for another day then :grin:

PR #16603, opened 2021-11-15
tests: fedora-coreos is now based on fedora-35 - adjust conditions accordingly

KKoukiou:


martinpitt (approved):
Thanks! This needs to land in lockstep with the image refresh.

PR #16604, opened 2021-11-15
login: Add toggle button to view password

garrett:
Based on https://www.patternfly.org/v4/components/login-page

This adds a toggle to switch text input types from "password" to "text",
to make it cleartext for viewing and back. SVG icons are inlined and
chosen based on the state of the input. Minimal JavaScript is used to
swap the attribute.

Allowing one to choose to swap to clear text lets people use complex
passwords and not have to second guess what they typed. It's a usability
and accessibility feature. (And it shouldn't be used when someone is
actively looking at their computer, but otherwise should be fine.)

The form still defaults to hiding the password by default.

### DIFF
--- a/pkg/static/login.css
+++ b/pkg/static/login.css
@@ -189,6 +189,7 @@ form .control-label {
   color: #151515;
 }
 
+.pf-c-button.pf-m-control,
 .form-control[type=password],
 .form-control[type=text] {
   display: block;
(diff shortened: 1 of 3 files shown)


KKoukiou (approved):
Tried this and works as expected. 
The UI is identical to the login form from the PF4 docs.

PR #16608, opened 2021-11-16
Eliminate obsolete cockpit.extend()

martinpitt:
This was an ages old compatibility shim for a time when Object.assign()
was not yet a standard browser API. Replace it with Object.assign() for
updating existing objects, and the spread operator for creating new
ones.

Drop the documentation as well. Keep cockpit.extend() as alias for
Object.assign() to keep the current API.


marusak (approved):
Thank you!

PR #16611, opened 2021-11-16
bridge: modernise ssh-agent spawning code

allisonkarlitskaya:
Port the code for spawning ssh-agent to GSubprocess, merging it with the
(now-very-similar) code for spawning dbus-daemon.

We spawn ssh-agent with -D so that we have better control over its
lifetime.  In particular, the `prctl()` that we were previously doing
had no effect, because without `-D` ssh-agent `fork()`s immediately.
Closing all file descriptors also had no effect because this was already
the default behaviour of `g_spawn_sync()` and is also the default
behaviour of `GSubprocess`.

### DIFF
--- a/src/bridge/bridge.c
+++ b/src/bridge/bridge.c
@@ -174,145 +174,73 @@ send_init_command (CockpitTransport *transport,
 }
 
 static GSubprocess *
-start_dbus_daemon (void)
+start_helper_process (const gchar * const  *argv,
+                      const gchar          *socket_pattern,
+                      const gchar          *socket_envvar)
 {
-  g_autoptr(GError) error = NULL;
+  {
+    const gchar *env = g_getenv (socket_envvar);
+    if (env && env[0])
+      return NULL;
+  }
 
+  g_autoptr(GError) error = NULL;
   /* The DBus daemon produces useless messages on stderr mixed in */
-  g_autoptr(GSubprocessLauncher) dbus_daemon = g_subprocess_launcher_new (G_SUBPROCESS_FLAGS_STDOUT_PIPE |
-                                                                          G_SUBPROCESS_FLAGS_STDERR_SILENCE);
-  g_subprocess_launcher_unsetenv (dbus_daemon, "G_DEBUG");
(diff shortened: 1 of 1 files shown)


src/bridge/bridge.c
```diff
@@ -174,145 +174,71 @@ send_init_command (CockpitTransport *transport,
...
+  g_assert_no_error (error);
 
-  if (!g_spawn_check_exit_status (status, &error))
+  g_autoptr(GMatchInfo) info = NULL;
+  if (!g_regex_match (regex, first_line, 0, &info))
     {
-      g_warning ("couldn't start %s: %s: %s", agent_argv[0],
-                 error->message, agent_error);
-      goto out;
+      g_warning ("output from %s didn't match expected pattern %s", argv[0], socket_pattern);
```
martinpitt, line 220: This is certainly a Does Not Happen™ condition, but still -- would be nice to terminate and wait the process here.

src/bridge/bridge.c
```diff
@@ -174,145 +174,71 @@ send_init_command (CockpitTransport *transport,
...
 
-static gboolean
-have_env (const gchar *name)
+static GSubprocess *
+start_ssh_agent (void)
 {
-  const gchar *env = g_getenv (name);
-  return env && env[0];
+  const char * const cmd[] = { "ssh-agent", "-s", "-D", NULL };
+  return start_helper_process (cmd, "SSH_AUTH_SOCK=([^;]*);", "SSH_AUTH_SOCK");
```
martinpitt, line 241: I'm quite sure you need to escape the () here as well, just like you did in start_dbus_daemon().

martinpitt (changes requested):
Nice one!

src/bridge/bridge.c
allisonkarlitskaya: i don't use `\` anywhere, actually.

src/bridge/bridge.c
allisonkarlitskaya: OK.  I did a force exit, and same for the similar case just above.

martinpitt (approved):
g_subprocess_force_exit(), oh my!

PR #16613, opened 2021-11-16
common: drop mock-io-stream.[ch]

allisonkarlitskaya:
This is basically just implementing GSimpleIOStream.  Use that instead.


martinpitt (approved):
:heart:

PR #16614, opened 2021-11-16
Port shell header to masthead component

KKoukiou:


garrett (changes requested):
I think we need to do the following in this PR:

- [ ] Add the highlight line back to desktop and mobile views

I guess I should look into this? I'm happy to do so.

---

We should fix this in this PR, but it could also be a follow-up:

- Fix residual highlight after closing the menu (might be a PF4 bug?)
  - This would possibly use https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

These should most definitely be follow-ups and we'll have open questions and will need to make decisions on how to best solve them:

- Make Admin access similar, perhaps?
- Port Host switcher to a similar dropdown? (We talked about this. It should be in another PR. It will require more work.)
  - (I think you're right where we can make this a more traditional dropdown/popover with a modal to add/edit/remove. The dropdown/popover would be able to have filtering inside still.)

garrett:
I didn't get to this today; sorry.

I'll try to get it done on Monday when everyone else is traveling.

garrett:
I've added a workaround for the masthead menu mouse dismissal issue and filed an upstream bug @ https://github.com/patternfly/patternfly/issues/4534

KKoukiou:
@garrett are you fine with keeping the suggestion for future improvements? I think the current code is already an improvement, and not regressing anything. 

I would like to ficus on my anaconda related work, thus wanted to get this in and not have pending cockpit work.

garrett:
> @garrett are you fine with keeping the suggestion for future improvements? I think the current code is already an improvement, and not regressing anything.

Sure! I can open up issue(s) for it. The only things meant for this PR directly was adding the highlight line back and fixing the residual line when something was activated by the mouse and is no longer active. I fixed those two things in commits here.

> I would like to ficus on my anaconda related work, thus wanted to get this in and not have pending cockpit work.

Yep, makes total sense!

I just built this PR again with the latest changes and will look at it right now.

garrett (approved):
Looks good!

The only thing I have a bit of a concern about now is how we've reverted the UX change for the highlighting of the menus at the top. (We did this as a response to a UX test from a while back.) But this is how PF has it, and there is a little bit of action when hovering, so perhaps it's actually fine.

Plus, the top-right looks different now when not hovering compared to when we ran those UX tests a while back, so it's most likely fine anyway.

garrett:
OK, I've spun out the follow-ups to separate issues.

KKoukiou:
The packit tests fail because I closed and reopend the PR by mistake.

martinpitt (approved):
Thank you! This LGTM, but I'd really appreciate if we could squash these commits. They all belong together, and the second commit reverts a part that the first commit introduces.

I added it to my TODO list to add a pixel test for the shell.

PR #16618, opened 2021-11-17
services: make minus/plus icons in create timer dialog align with the selectors

### LINKED ISSUE #15927: Create timer modal improvements (SPUR)

KKoukiou:
Fixes #15927

### DIFF
--- a/pkg/systemd/services/timer-dialog.jsx
+++ b/pkg/systemd/services/timer-dialog.jsx
@@ -326,13 +326,11 @@ const CreateTimerDialogBody = ({ setIsOpen, owner }) => {
                                                 <Button aria-label={_("Remove")}
                                                         variant="secondary"
                                                         isDisabled={repeatPatterns.length == 1}
-                                                        isSmall
                                                         onClick={() => setRepeatPatterns(repeatPatterns.filter((item, item_idx) => idx != item_idx))}>
                                                     <MinusIcon />
                                                 </Button>
                                                 <Button aria-label={_("Add")}
                                                         variant="secondary"
-                                                        isSmall
                                                         onClick={() => {
                                                             if (repeat == "hourly")
                                                                 setRepeatPatterns([...repeatPatterns, { key: repeatPatterns.length, minute: "0" }]);


pkg/systemd/services/timer-dialog.jsx
```diff
@@ -250,6 +252,11 @@ const CreateTimerDialogBody = ({ setIsOpen, owner }) => {
                             const min = repeatPatterns[idx].minute;
                             const validationFailedMinute = !(/^[0-9]+$/.test(min) && min <= 59 && min >= 0);
 
+                            if (repeat == 'monthly' && repeatPatterns.some(p => p.day > 28)) {
+                                validated = "warning";
+                                helperText = _("Choosing a day after the 28th limits the execution of the timer to the months which contain it");
```
martinpitt, line 257: :thinking: that feels rather obvious to me.. did some user actually complain about this? 

Can we trim the string a bit, so that it's easier to translate and won't explode in length in other languages? How about "Will not run every month"?

martinpitt (changes requested):
Thanks! Sorry for the bikeshed, but I'm always a bit sensitive/picky when it comes to long translatable strings.

pkg/systemd/services/timer-dialog.jsx
KKoukiou: I also find it obvious and I am happy to drop this commit. I think it was just a question in the first SPUR.

KKoukiou: See the issue attached, there is a discussion about it.

KKoukiou: I removed the commit, I actually agree that it's not necessary.

martinpitt (approved):
OK, thanks! I didn't mean to completely talk you out of this, but indeed it's a bit on the edge.

PR #16619, opened 2021-11-17
packagekit: Show "Web Console restart" for Arch and derivates

jelly:
On Arch Linux and derivates cockpit-ws is included in the cockpit
package which makes users miss out on the "Web Console restart" warning
message. As these distriubtions all use the "alpm" backend, the
component now on load fetches the configured backend (can only be one)
to determine which package to check. Using packagekit was chosen over
parsing /etc/os-release or checking if pacman is installed on the
system as we already interact with packagekit over Dbus.

### DIFF
--- a/pkg/lib/packagekit.js
+++ b/pkg/lib/packagekit.js
@@ -501,3 +501,11 @@ export function install_missing_packages(data, progress_cb) {
                                       }
                                   });
 }
+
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Nice one! Retrying the flakes.

PR #16620, opened 2021-11-17
Remove info-table-ct class which is used in log detail page

### LINKED ISSUE #15170: Port log detail table to PF4
Page: Logs / detail view (click on a log entry)

The page uses an HTML table with "info-table-ct" and the "labels" are actually `<td>`s instead of `<th>`s, making this inaccessible.

We _could_ fix the table... or outright port it to PF4 (which should also fix the a11y issues). Of course, it makes sense to do the latter and port to [PF4's description list (horizontal variant)](https://www.patternfly.org/v4/components/description-list#horizontal).

---

Once this is done, we might want to consider de-emphasizing the **ALL_UPPERCASE** labels, perhaps with CSS like the following:

```scss
.log-details .pf-c-description-list__term {
  --pf-c-description-list__term--FontWeight: var(--pf-global--FontWeight--normal);
  color: var(--pf-global--Color--200);
}
```

PF4's bolded labels in the description list wasn't designed with the intent of having all description list terms to be uppercase. They're yelling.

In a previous commit a while back, I wanted to to make these lowercase (via CSS, so copy/paste would preserve the uppercase), but was told that it's important to maintain parity with journalctl's commandline. (Is it actually case-sensitive though?) So we'd have to figure out another way to de-emhpasize the characters.

Our main font, RedHatText does not support tabular numbers, so I'd be surprised if it would support a small capitals option. Therefore, I am suggesting the above snippet (or something like it) as a one-off.

KKoukiou:
Fixes #15170


KKoukiou:
New:
Screen Shot 2021-11-17 at 17 28 21
Screen Shot 2021-11-17 at 17 28 16
Screen Shot 2021-11-17 at 17 27 31
Screen Shot 2021-11-17 at 17 27 20
Screen Shot 2021-11-17 at 17 27 11
Screen Shot 2021-11-17 at 17 27 05
Screen Shot 2021-11-17 at 17 26 54
Screen Shot 2021-11-17 at 17 26 46
Screen Shot 2021-11-17 at 17 26 37

garrett:
I've added a few commits that makes the page look like this, at least on my system:

Screenshot 2021-11-17 at 18-11-33 Logs - garrett Bolt

It might look differently on different entries, though, so just treat these as suggestions.

I also did add some margin and padding, and we could probably use PF utils instead for the same thing, if applied in the right place. However, I'm not aware of how to add a border via utils.

And I did override the PF header and action components to move the link-styled-button closer. It might be able to be done with PF utilities too (at least the alignment part; not sure about the gap), now that I'm looking at it.

If you think these suggestions should be implemented differently, please feel free to drop them and re-implement how you'd like.

pkg/systemd/logs.scss
```diff
@@ -35,6 +35,16 @@
 }
 
 #log-details {
+    .pf-c-card__header {
```
KKoukiou, line 38: This should be done from JS, the link can be put in the CardTitle, instead of the CardAction and this is not needed.

Also this breaks other views like:
Screen Shot 2021-11-17 at 18 33 07

KKoukiou: So I removed this comimit, as in this PR, I also changed that card title to be the unit name, and not the syslog idendifier. So the initial issue is solved.

garrett: Yeah, this works. We should make the link more obvious too, by not having it so far away. People had issues finding it. But it can be handled in another PR. What's here is a huge improvement and I'm happy you did this (and so quickly)!

garrett (dismissed):
LGTM!

pkg/systemd/logs.scss
```diff
@@ -47,7 +47,6 @@
     .multiline {
         word-break: break-all;
         white-space: pre-wrap !important;
-        font-family: monospace, monospace;
```
martinpitt, line 56: This applies to both pf-c-card__title (where it's justified and matches the intention) and .multiline, which we use in various ABRT logs. It should probaly stay for .multiline, i.e. this rule needs to be broken up?

martinpitt (changes requested):
Question/possible fixup for @garrett , otherwise this looks nice!

pkg/systemd/logs.scss
```diff
@@ -56,6 +56,9 @@
 
     .pf-c-description-list {
         --pf-c-description-list--m-horizontal__term--width: 25ch;
+        border-top: 1px solid var(--pf-global--BorderColor--100);
```
KKoukiou, line 59: This adds a weird border to the description lists when it's under tabs. 
Screen Shot 2021-11-18 at 16 27 34

KKoukiou: I suggest let's revert this commit, as it's actualy changing the design from before, but also breaking the abrt logs details.

garrett: I was afraid of that. I couldn't find some entry to trigger those tabs.

Could we make it conditional to not having tabs? I guess add a `:first-child` to it, perhaps, if the tabs are at the same level? Or have it be a class the JSX adds? The styling of the border and margin/padding should be in the same style.

garrett: It's fine if we even drop the commit to merge this in. But I'd like a little round of polishing up afterward if that's the case. (But it is simple enough that we might be able to include it with a minor modification in this PR?)

KKoukiou: I actually think that you don't need border in the list but a horizontal ruler with spacers, and that only in that simple cards, (not abrt).

pkg/systemd/logs.scss
KKoukiou: I brought it back for the multine class.

martinpitt (approved):
Excellent, thank you! This looks good as-is, and the extra ruler (discussion from above) can be discussed/changed separately.

PR #16621, opened 2021-11-17
login: Use × icon as input clear; show when needed

garrett:
This changes the red emoji &#x274c; to an actual SVG icon from PatternFly. It also only shows when there's something typed into the field, thanks to the (empty) placeholder attribute.

Screenshot 2021-11-17 at 17-29-04 bolt
Screenshot 2021-11-17 at 17-28-53 bolt

### DIFF
--- a/pkg/static/login.css
+++ b/pkg/static/login.css
@@ -75,13 +75,13 @@ img {
 
 #option-group,
 .pf-c-button,
-.cross,
+.input-clear,
 button {
   cursor: pointer;
 }
 
 #option-group svg,
-.cross {
+.input-clear {
   opacity: .7;
 }
 
@@ -520,7 +520,7 @@ label.checkbox {
   text-decoration: underline;
 }
 
-.cross,
+.input-clear,
 .inline .container .help-block {
   color: #000;
 }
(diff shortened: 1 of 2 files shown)


garrett:
The gibberish on the second screenshot is to demonstrate what happens when one types too much in the input. The background of the × is preserved (instead of floating on top of the text).

allisonkarlitskaya (commented):
The visual changes are really positive and help out with some really ugly 15 pixels of UI in the start screen of cockpit client.  So +1 there.

I'm not one to review these kinds of changes though.  Probably @KKoukiou or so?  In any case, I kinda dislike the svg path being copied into the source here.  Can't we get webpack to do that for us?

martinpitt (approved):
Excellent, thank you! Works as expected.

@allisonkarlitskaya : PF icons are a bit .. difficult. For now I think this is fine. If and when we rewrite the login page in "proper" React-PF4, these copies will go away, of course.

garrett:
We already include other SVG icons inline in the page. It doesn't make sense to have them as external files just to be included once and compiled in just for this one use.

I did optimize them to be as small as possible tweaking the paths a little in Inkscape and then using SVGO (technically the web app UI @ https://jakearchibald.github.io/svgomg/).

> If and when we rewrite the login page in "proper" React-PF4, these copies will go away, of course.

Yes, I'd agree... but we do have special considerations for the login page. It needs to work in browsers that PF4 does not support, to properly say we don't support the browsers. :wink: 

And that's only part of the reason it's special and its own thing.

PR #16623, opened 2021-11-18
storaged: Create VDO devices as LVM LV type

martinpitt:
This is the officially documented and supported API [1], both on the
command line and in UDisks. Add them as a new type to the "Create
logical volume" dialog and drop the standalone one from the things
panel.

Drop the "Use 512 byte emulation", this is not a thing with LVM VDO.

Also drop the "index memory" slider. This is a bit dubious, not even
exposed by `lvcreate` or the documentation, and can only be done via
changing the config file. The UDisks2 D-Bus API does support setting it,
but this is probably not something that the admin should customize
lightly (and if so, they can still set it in the config file). Also,
neither `lvdisplay` nor UDisks2 have an API to get the current value.

The pool name is rather uninteresting -- in fact, lvcreate does not even
need it. udisks2 does not currently support automatic name creation
(https://github.com/storaged-project/udisks/issues/939), so do the same
thing as lvcreate and use the next available `vpoolN`.

Rework the on-demand package installation: As it is only necessary when
selecting "vdo" in the "create LV" dialog, it can't happen before
opening the dialog. Also, we don't want (or even can) have the standard
install dialog on top of the create dialog; instead, just show an info
alert that the vdo package is going to be installed, and build the
progress and installation into the "Create" button. The vdo+kmod-kvdo
packages are small enough that this won't take an inordinate amount of
time anyway. We also don't expect any package removals there, so abort
if they would happen.

Adjust the tests to use the new dialog to create the VDO. But also
validate that the API properly reacts to CLI creation/removal.

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deduplicating_and_compressing_logical_volumes_on_rhel/creating-a-deduplicated-and-compressed-logical-volume_deduplicating-and-compressing-logical-volumes

----

New choice in the dialog if `vdo` package is installed:

vdo-create

If it is not installed, but available, you get an info alert, and the progress status. The "Cancel" link works until packagekit is into installing the package. I tested that manually.

vdo-create-installing

The whole process takes just a few seconds even with the real vdo package (not just our test suite mock), so I think that's okay with just the spinner.

### DIFF
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -20,7 +20,7 @@
 import cockpit from "cockpit";
 import {
     dialog_open, TextInput, PassInput, SelectOne, SizeSlider, CheckBoxes,
-    BlockingMessage, TeardownMessage, teardown_and_apply_title
+    BlockingMessage, TeardownMessage, Message, teardown_and_apply_title
 } from "./dialog.jsx";
 import * as utils from "./utils.js";
 
@@ -34,6 +34,7 @@ import { ExclamationTriangleIcon } from "@patternfly/react-icons";
 import { ListingTable } from "cockpit-components-table.jsx";
 import { ListingPanel } from 'cockpit-components-listing-panel.jsx';
 import { StorageButton, StorageBarMenu, StorageMenuItem } from "./storage-controls.jsx";
+import * as PK from "packagekit.js";
 import {
     format_dialog, parse_options, extract_option, unparse_options
 } from "./format-dialog.jsx";
@@ -52,7 +53,7 @@ const _ = cockpit.gettext;
 
 const C_ = cockpit.gettext;
 
-function next_default_logical_volume_name(client, vgroup) {
+function next_default_logical_volume_name(client, vgroup, prefix) {
     function find_lvol(name) {
         const lvols = client.vgroups_lvols[vgroup.path];
         for (let i = 0; i < lvols.length; i++) {
@@ -64,7 +65,7 @@ function next_default_logical_volume_name(client, vgroup) {
 
     let name;
     for (let i = 0; i < 1000; i++) {
-        name = "lvol" + i.toFixed();
+        name = prefix + i.toFixed();
         if (!find_lvol(name))
             break;
     }
@@ -160,7 +161,7 @@ function create_tabs(client, target, is_partition, is_extended) {
             Fields: [
                 TextInput("name", _("Name"),
                           {
-                              value: next_default_logical_volume_name(client, vgroup),
+                              value: next_default_logical_volume_name(client, vgroup, "lvol"),
                               validate: utils.validate_lvm2_name
                           }),
                 SizeSlider("size", _("Size"),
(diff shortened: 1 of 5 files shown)


martinpitt:
@garrett: This is a new workflow, where on-demand package installation is built into the creation dialog. We wanted to avoid putting a dialog on top of a dialog, and the involved packages are really small. WDYT? The screenshots are accurate right now.

pkg/storaged/content-views.jsx
```diff
@@ -842,17 +872,68 @@ export class VGroup extends React.Component {
...
                             return vgroup.CreatePlainVolume(vals.name, vals.size, { });
                         else if (vals.purpose == "pool")
                             return vgroup.CreateThinPoolVolume(vals.name, vals.size, { });
+                        else if (vals.purpose == "vdo") {
+                            return (need_vdo_install ? install_package(vdo_package, progress) : Promise.resolve())
+                                    .then(() => {
+                                        progress(_("Creating VDO device")); // not cancellable any more
+                                        return vgroup.CreateVDOVolume(
+                                            // HACK: hardcode pool name; https://github.com/storaged-project/udisks/issues/939
+                                            vals.name, vals.name + "_vdopool",
```
mvollmer, line 930: We could re-implement the loguic in lvcreate here and choose the first 'vpool<N>' name that is still free, similar to how we create default names for new logical volumes.

mvollmer:
@martinpitt, what about some pixel tests?

martinpitt:
@mvollmer : I'm happy about pixel tests if you agree on the general design. I'll add them after @garrett gave his blessing.

pkg/storaged/content-views.jsx
martinpitt: Good idea, I'll look into that.

mvollmer:
LGTM otherwise!

pkg/storaged/content-views.jsx
martinpitt: Conveniently there already is next_default_logical_volume_name(), so I just had to generalize this a bit. Done!

martinpitt:
@mvollmer: Ah, now I remember again -- vdo doesn't work on Fedora, so we can't have pixel tests.

mvollmer:
> @mvollmer: Ah, now I remember again -- vdo doesn't work on Fedora, so we can't have pixel tests.

Ouch, that's too obvious!

mvollmer (approved):

garrett:
Overall: Modal dialogs should not use an inline alert, except as an actual alert for an error or warning, and in that case, it should be at the top. (I remember PatternFly having a design guideline about this, but I cannot find it. Perhaps they dropped this when they added alert modals?) But even then, when there's an error, it should be in the same context of where the error comes from whenever possible.

But _**this** dialog_ should not have an alert for an information string. This should be helper text. In this case, with an information icon, it should be a helper text widget, I guess. Otherwise, there's the built-in form input helper text.

https://www.patternfly.org/v4/components/helper-text#static-with-custom-icons
https://www.patternfly.org/v4/components/form/#form-group-with-additional-label-info

garrett:
Since this is merged, please address it in a quick follow-up.

garrett:
This shouldn't be a warning, but should be an info type of a notice.

martinpitt:
I sent PR #16644 for this, with info and our standard info icon.

PR #16624, opened 2021-11-19
Fix title case strings

marusak:


pkg/storaged/stratis-details.jsx
```diff
@@ -364,17 +364,17 @@ export const StratisPoolDetails = ({ client, pool }) => {
 
     function create_fs() {
         dialog_open({
-            Title: _("Create Filesystem"),
+            Title: _("Create filesystem"),
```
martinpitt, line 367: If we break translatable strings, then let's please fix them properly, and move to "file system".

martinpitt (changes requested):
Always a little sad to break so many translations -- but I suppose we can't update our po/* files, we'd have to do that in weblate now?

pkg/storaged/stratis-details.jsx
marusak: we use "file system" once in the whole Cockpit and "filesystem" dozens of times. Are you sure I should change this one occurrence to "file system"?

martinpitt: IMHO *if* we touch a translatable string, we should better just fix it at the same time.

marusak: Done

allisonkarlitskaya (changes requested):
"filesystem" is very consistently used through (for example) `man mount(8)`

allisonkarlitskaya (dismissed):
...actually, I changed my mind.  I prefer "file system" now!!! :rofl:

In seriousness, though, you'll need to rebase this

Long story short: I didn't notice that Martin was checking out the head commit when I removed his use of a rebasing jumpstart (and this was before we properly supported unmerging).  I'll fix that in a separate issue.  It basically means that we use the merged workflow with the head tree, and in your case, that doesn't include the `test` script.  So ya... rebase please :/

marusak:
Re/plant-based

martinpitt:
Still two pixel test errors, sorry

martinpitt:
Still five more pixel tests to update, I force-pushed them now. Thanks!

martinpitt (approved):

martinpitt (approved):

PR #16626, opened 2021-11-19
ws: only spawn cockpit-session on known auth types

allisonkarlitskaya:
If we receive requests for exotic authentication types from the client like:
         
``` 
Authorization: Random xyz
```
    
and those types are not explicitly disabled in cockpit.conf with a stanza like:
                                                                      
```             
[Random]
action = none
```
   
then by default we'll try to spawn cockpit-session to respond to them.

This doesn't make a lot of sense, as cockpit-session doesn't support any types other than "basic", "negotiate" and "tls-cert" (which is only ever used internally, and already blocked when received from clients).                                                 
                                                                           
Modify the check to only spawn cockpit-session for the recognised types.  In case another type is specified, then the command to handle that type needs to be explicitly specified:
                                                                            
```                                                                        
[Random]                             
command = /path/to/my/handler
```
   
as we already have for several cases in the unit tests.

### DIFF
--- a/src/ws/cockpitauth.c
+++ b/src/ws/cockpitauth.c
@@ -1121,7 +1121,6 @@ cockpit_session_launch (CockpitAuth *self,
   const gchar *action;
   const gchar *command;
   const gchar *section;
-  const gchar *program_default;
 
   gchar **env = g_get_environ ();
 
@@ -1153,6 +1152,7 @@ cockpit_session_launch (CockpitAuth *self,
   else
     section = type;
 
+  const gchar *program_default = NULL;
   if (g_strcmp0 (section, COCKPIT_CONF_SSH_SECTION) == 0)
     {
       if (!host)
@@ -1164,12 +1164,21 @@ cockpit_session_launch (CockpitAuth *self,
       capture_stderr = cockpit_conf_bool ("WebService", "X-For-CockpitClient", FALSE);
       program_default = cockpit_ws_ssh_program;
     }
-  else
+  else if (type && (g_str_equal (type, "basic") ||
+                    g_str_equal (type, "negotiate") ||
+                    g_str_equal (type, "tls-cert")))
     {
       program_default = cockpit_ws_session_program;
     }
 
   command = type_option (section, "command", program_default);
+
+  if (!command)
+    {
+      g_set_error (error, COCKPIT_ERROR, COCKPIT_ERROR_AUTHENTICATION_FAILED,
+                   "Authentication disabled");
+      goto out;
+    }
 
   if (cockpit_creds_get_rhost (creds))
     {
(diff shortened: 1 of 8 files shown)


src/ws/test-auth.c
```diff
@@ -749,10 +749,22 @@ static const ErrorFixture fixture_non_ascii = {
...
+static const SuccessFixture fixture_bearer = {
+  .warning = NULL,
+  .data = NULL,
+  .header = "Bearer correcttoken"
+};
+
+static const ErrorFixture fixture_invalid_bearer = {
+  .error_code = COCKPIT_ERROR_AUTHENTICATION_FAILED,
+  .error_message = "Unsupported Authorization type",
+  .header = "Bearer ocrrecttoken"
```
martinpitt, line 765: Can you please use `wrongtoken` here, to better convey that this is not a typo, but a deliberately wrong password for testing?

martinpitt (changes requested):
Thanks! We still have some negotiating to do *how* to fix that, but I fully agree that this needs to be robustified.

src/ws/cockpitauth.c
```diff
@@ -1164,13 +1164,22 @@ cockpit_session_launch (CockpitAuth *self,
...
       program_default = cockpit_ws_session_program;
     }
 
   command = type_option (section, "command", program_default);
 
+  if (!command)
+    {
+      g_set_error (error, COCKPIT_ERROR, COCKPIT_ERROR_AUTHENTICATION_FAILED,
+                   "Authentication disabled");
+      goto out;
```
martinpitt, line 1180: Just to make sure I fully understand this: Disabling an auth method with `action = none` already happens further up in line 1136. This code would only apply to a custom auth type that has no `command =`, so it's less about "disabled" and more about "misconfigured", right? (No need to change the string)

martinpitt (approved):
This is really nice, clean, and simple. Thank you!

src/ws/cockpitauth.c
allisonkarlitskaya: > Just to make sure I fully understand this: Disabling an auth method with `action = none` already happens further up in line 1136. This code would only apply to a custom auth type that has no `command =`, so it's less about "disabled" and more about "misconfigured", right? (No need to change the string)

It's more about the case we discussed: random unexpected incoming Authorization headers.  The default "localhost" action will still be taken for those, but without a command configured, they will now fail.  That's slightly different than the "action" approach we discussed but it has the advantage of not breaking existing users of command= for strange auth types, including many in the unit tests.  If the user explicitly configured a command then they probably expect it to work.

martinpitt: Ah right, due to the NULL program_default for unknown auth types, type_option() would return NULL for that as well. `</thinking_aloud>`. Thanks!

PR #16628, opened 2021-11-22
tools: Fix SELinux module installation in non-SELinux environments

martinpitt:
`%selinux_modules_{,un}install` works fine in environments like guestfs,
which don't run with SELinux enabled. Adjust the condition so that
module installation happens there as well. The relabeling still won't
work there, but that has its own guard with calling `selinuxenabled`.

https://bugzilla.redhat.com/show_bug.cgi?id=2023579

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -521,7 +521,7 @@ fi
 
 %post ws
 %if 0%{?with_selinux}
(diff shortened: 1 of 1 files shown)


martinpitt:
I validated that this works in https://src.osci.redhat.com/rpms/cockpit/pull-request/86 (sorry, RH internal only)

marusak (approved):
Thank you!

PR #16629, opened 2021-11-22
start merging flatpak

allisonkarlitskaya:
Let's start discussing underlying changes to the build system, login page, and cockpit-ws to support cockpit-client.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -6,6 +6,7 @@ check_SCRIPTS =
 check_PROGRAMS =
 noinst_SCRIPTS =
 libexec_PROGRAMS =
+libexec_SCRIPTS =
 noinst_PROGRAMS =
 sbin_PROGRAMS =
 noinst_LIBRARIES =
--- a/configure.ac
+++ b/configure.ac
@@ -154,7 +154,6 @@ AC_CHECK_HEADER([security/pam_appl.h], ,
 )
 PAM_LIBS="-lpam"
 COCKPIT_SESSION_LIBS="$COCKPIT_SESSION_LIBS $PAM_LIBS"
-COCKPIT_WS_LIBS="$COCKPIT_WS_LIBS $PAM_LIBS"
 
 # pam module directory
 AC_ARG_WITH([pamdir],
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -558,7 +558,7 @@
         id("login-button-text").textContent = (form == "hostkey") ? _("Accept key and log in") : _("Log in");
         id("login-password-input").value = '';
 
-        if (need_host()) {
+        if (environment.page.require_host) {
             hide("#option-group");
             expanded = true;
         } else {
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -350,6 +350,8 @@
             }
         } else if (logout_intent) {
             show_login(logout_reason);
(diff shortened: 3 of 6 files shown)


src/ws/cockpithandlers.c
```diff
@@ -302,9 +302,11 @@ add_page_to_environment (JsonObject *object)
     }
 
   require_host = cockpit_conf_bool ("WebService", "RequireHost", FALSE);
+  gboolean hide_login = cockpit_conf_bool ("WebService", "HideLogin", FALSE);
```
martinpitt, line 305: Sorry for raising a naming discussion again, but as this introduces config/API: In essence, this enables "client mode" on the login page, right? Which particularly means "only give me an SSH address/host target". So it's still a log in, just not user/password.

So this also conceptually collides with `RemoteHost=`, as you already wrote in the commit message. In https://bugzilla.redhat.com/show_bug.cgi?id=2018741 there was some discussion about fixing that option to also allow the user to entirely disable remote hosts. Do you think we could make this compatible with that approach and have e.g.

    RemoteHost=require|allow|no|only

with "only" meaning that the login page only shows the host field, and nothing else? I'm of course happy to change the values here, but I think it makes sense to unify mutually exclusive config keys into such an enum.

Of course, for this PR we'd only implement the `=only` value.

Also, this needs documentation.

src/branding/default/branding.css
```diff
@@ -28,6 +28,10 @@ body.login-pf {
     font-weight: bold;
 }
 
+#brand:before {
+    content: "Cockpit";
```
martinpitt, line 32: Where and when does this actually appear? On the standard web login page we actually don't want the product name to appear. I see that in default mode the page has class `hide-before`.

martinpitt (changes requested):
Thanks! 6 commits are fine/obvious, but some discussion needed for the config option.

src/ws/cockpithandlers.c
allisonkarlitskaya: I like this suggestion.  I'll look into it.

src/branding/default/branding.css
allisonkarlitskaya: This is the big banner that appears right above the login form.  I'll drop it again if that was a design decision.  It just seemed like a bug, when we have the OS name for all the other cases.

allisonkarlitskaya:
Removed the controversial bits.  Let's do follow-ups.

src/branding/default/branding.css
martinpitt: Right, that was sort of my question -- I'm not saying that it's wrong, I'd just like to understand it (I'm not very familiar with the login CSS). *If* there is an OS name, then showing "Ubuntu 20.04" or "Fedora Linux" in #brand -- without a "Cockpit" prefix -- is desired. If there is no OS (such as in a bastion container or flatpak?) then showing "Cockpit" as a fallback makes sense, so that it's not completely empty.

src/ws/cockpithandlers.c
martinpitt: FTR, for both here and PR #16630 -- if this is too entangled, I'd be happy to temporarily use an undocumented [WebService] `_Internal_Client = yes`  flag, and merge this into RequireHost= later.

martinpitt (approved):
Thanks!

PR #16630, opened 2021-11-22
ws: add option to report cockpit-ssh stderr

allisonkarlitskaya:
Add a "ReportStderr" option to the [Ssh-Login] session which sends the
stderr output from cockpit-ssh to the client as part of the error
message.

This is only intended for use with the Cockpit Client flatpak.  Sending
logging information to random remote users might be a security problem.

On the login page we reuse the fatal error message and the "Try again"
link.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -108,6 +108,23 @@
         hideToggle(arguments, true);
     }
 
+    function show_captured_stderr(msg) {
+        if (window.console)
+            console.warn("stderr:", msg);
+
+        hide("#login-wait-validating");
+
+        hide("#login", "#login-details");
+        show("#login-fatal");
+
+        id("login-again").onclick = () => { hide('#login-fatal'); show_login() };
+        show("#login-again");
+
+        const el = id("login-fatal-message");
+        el.textContent = "";
+        el.appendChild(document.createTextNode(msg));
+    }
+
     function fatal(msg) {
         if (window.console)
             console.warn("fatal:", msg);
@@ -785,7 +802,9 @@
                 } else {
                     if (window.console)
                         console.log(xhr.statusText);
-                    if (xhr.statusText.indexOf("authentication-not-supported") > -1) {
+                    if (xhr.statusText.startsWith("captured-stderr:")) {
+                        show_captured_stderr(decodeURIComponent(xhr.statusText.replace(/^captured-stderr:/, '')));
+                    } else if (xhr.statusText.indexOf("authentication-not-supported") > -1) {
                         const user = trim(id("login-user-input").value);
                         fatal(format(_("The server refused to authenticate '$0' using password authentication, and no other supported authentication methods are available."), user));
                     } else if (xhr.statusText.indexOf("terminated") > -1) {
--- a/src/ws/cockpitauth.c
+++ b/src/ws/cockpitauth.c
@@ -525,7 +525,8 @@ session_child_setup (gpointer data)
 
 static CockpitTransport *
 session_start_process (const gchar **argv,
-                       const gchar **env)
+                       const gchar **env,
+                       gboolean capture_stderr)
 {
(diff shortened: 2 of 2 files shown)


src/ws/cockpitauth.c
```diff
@@ -1147,6 +1158,7 @@ cockpit_session_launch (CockpitAuth *self,
       if (!host)
         host = type_option (COCKPIT_CONF_SSH_SECTION, "host", "127.0.0.1");
 
+      capture_stderr = cockpit_conf_bool (COCKPIT_CONF_SSH_SECTION, "reportstderr", FALSE);
```
martinpitt, line 1161: TBH I'm not happy for this to be a config option, at least at first sight. Is there any harm in just always capturing stderr? That seems useful for "regular" bastion host/remote host usage as well. I suppose with cockpit-ssh we have tight control about stderr and there's usually nothing interesting on it (as it speaks the json protocol), but then it's at most "harmless".

Then the login page can decide for itself in which mode it is, and show stderr if it is in client mode. I realize that we might not want to do that unconditionally for web mode, in order to not give away too much information on login failures -- but as the login page is already different in client mode, it already knows whether or not its appropriate to show ssh stderr?

src/ws/cockpitauth.c
```diff
@@ -892,13 +895,20 @@ on_transport_closed (CockpitTransport *transport,
...
+          g_set_error (&error, COCKPIT_ERROR, COCKPIT_ERROR_AUTHENTICATION_FAILED,
+                       "captured-stderr:%s", captured_error);
+        }
       /* we get "access-denied" both if cockpit-session cannot execute cockpit-bridge (common case)
        * and if cockpit-session itself is not executable (corner case, messed up install) */
-      if (problem && (!session->authorize || g_strcmp0 (problem, "access-denied") != 0))
+      else if (problem && (!session->authorize || g_strcmp0 (problem, "access-denied") != 0))
         {
           g_set_error (&error, COCKPIT_ERROR, COCKPIT_ERROR_FAILED,
                        g_strcmp0 (problem, "no-cockpit") == 0
```
martinpitt, line 914: IMHO it should first look at known error codes from the protocol (such as `no-cockpit`), and *then* fall back to showing raw stderr. I.e. the ":" branch here with "internal error" should show stderr (if present), and this could even happen unconditionally.

pkg/static/login.js
```diff
@@ -785,7 +802,9 @@
                 } else {
                     if (window.console)
                         console.log(xhr.statusText);
-                    if (xhr.statusText.indexOf("authentication-not-supported") > -1) {
+                    if (xhr.statusText.startsWith("captured-stderr:")) {
+                        show_captured_stderr(decodeURIComponent(xhr.statusText.replace(/^captured-stderr:/, '')));
+                    } else if (xhr.statusText.indexOf("authentication-not-supported") > -1) {
```
martinpitt, line 807: Similarly here -- known error codes should be handled first, and captured stderr should be the fallback.

martinpitt (changes requested):
Thanks!

src/ws/cockpitauth.c
martinpitt: As I said in https://github.com/cockpit-project/cockpit/pull/16629#discussion_r754859348, I'd be happy to temporarily use an undocumented [WebService] _Internal_Client = yes flag, and merge this into RequireHost= later. But at first sight, reporting ssh stderr with `RequireHost=!no` (i.e. "we are doing something with ssh") might not be too wrong?

But either way, to me this doesn't feel right for a config option, as it's not really a static decision. auth.c should know which auth mode it is doing, and whether it's appropriate to expose stderr. 

As this is a bit hairy (give enough information vs. avoid information leaks), maybe putting this into `_Internal_Client` for now is not the worst start?

allisonkarlitskaya:
→ #16632

PR #16632, opened 2021-11-23
X-For-CockpitClient

allisonkarlitskaya:
super secret internal config option to enable lots of unrelated things!

Would also be possible as an envvar.  I don't mind either way.


martinpitt (approved):
Thanks! I like this better -- at least it keeps the provisional stuff/temporary bolted-on features greppable, and behind a single switch.

PR #16636, opened 2021-11-23
separate cockpit-client from cockpit-desktop, adding many improvements

allisonkarlitskaya:


src/client/cockpit-client
```diff
@@ -75,6 +82,9 @@ class CockpitClient(Gtk.Application):
 
         self.uri = self.ws.start()
 
+    def new_window(self, *args):
+        CockpitClientWindow(self, self.uri).present()
+
     def do_activate(self):
         CockpitClientWindow(self, self.uri).present()
```
martinpitt: These two do exactly the same, could new-window just be bound to do_activate()?

src/client/cockpit-client
```diff
@@ -48,6 +49,20 @@ class CockpitClientWindow(Gtk.ApplicationWindow):
...
 
 
+    def decide_policy(self, view, decision, decision_type):
+        if decision_type == WebKit2.PolicyDecisionType.NEW_WINDOW_ACTION:
+            uri = decision.get_navigation_action().get_request().get_uri()
+            if uri.startswith('http://127'):
+                logging.error('warning: no support for pop-ups')
+            else:
+                # We can't get the timestamp from the request, so use Gdk.CURRENT_TIME (== 0)
+                Gtk.show_uri_on_window(self, uri, 0)
```
martinpitt, line 73: Nice! That's half of the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1882677 . I suppose this still won't work in cockpit-desktop as that most likely won't reach out through the ns namespace/flatpak?

src/client/cockpit-client
```diff
@@ -19,35 +19,49 @@
...
 
-        self.webview = WebKit2.WebView(is_ephemeral=True)
         self.webview.get_settings().set_enable_developer_extras(True)
         self.webview.connect('decide-policy', self.decide_policy)
-        self.add(self.webview)
-        self.webview.show()
-
         self.webview.load_uri(uri)
-        self.set_title("Loading...")  # like login.html
-        self.webview.bind_property("title", self, "title")
```
martinpitt: For cockpit-desktop we need this back and hide the GtkHeaderBar, so that it continues to look like a normal desktop app (not half a web browser), at least if we are showing a particular frame (`cockpit-desktop subscriptions`) instead of the whole shell (`/`). You can keep the accelerators, I think.

So these two should hopefully not be too intrusive to conditionalize?

martinpitt (changes requested):

src/client/cockpit-client
allisonkarlitskaya: no.  the signatures are different.

martinpitt: `*args` is compatible to "arguments"

martinpitt:
This [coreos failure](https://logs.cockpit-project.org/logs/pull-16636-20211123-163257-21d98bbb-fedora-coreos/log.html#13-2) is real -- the rpm picked up an automatic python3 dependency, and that's not available in Fedora CoreOS. Can we prevent this, by adding an explicit `Suggests: python3`?

martinpitt:
I think I tamed the over-aggressive python3 rpm dependency now. https://docs.fedoraproject.org/en-US/packaging-guidelines/AutoProvidesAndRequiresFiltering/

martinpitt (approved):
Thanks! If this goes green, good to go.

allisonkarlitskaya:
The rhel test failures look related to 0b206dabefb111ea8a866370993a6e2ccf7aefea (and the fact that they are running later in the day).

Issue #13091, opened 2019-11-05
Navigating back on Logs page is broken


marusak:
Often going back one page with browser back button (or any browser supported shortcut) does not work as expected.
Examples: https://github.com/cockpit-project/cockpit/issues/1043 https://github.com/cockpit-project/cockpit/issues/3809

martinpitt:
I actually think the page specific issues are better, unless they all have the same root cause?

marusak:
> I actually think the page specific issues are better, unless they all have the same root cause?

Ah, I thought this is directly related to using `cockpit.navigation.go`

martinpitt:
My gut feeling is that for this to work, every "state" of a page (looking at a particular service, log entry, etc.) needs to build a unique URL. I may have misunderstood, of course.

garrett:
It's also related to jump.

@martinpitt: That's certainly part of the solution. Every page should have its own URL. We usually do this, but not always.

But even when we do this, going back doesn't always work quite right. For example, have a bunch of services on the services page, scroll down some, click on a service, then click back. You lose your place. It's already horrible for everyone, but it's even worse for screen readers.

marusak:
Well I don't think it is only related to having different urls.
1. Go to services - url: `https://localhost:9090/system/services`
2. Go to logs - url: `https://localhost:9090/system/logs`
3. Go to specific log - url: `https://localhost:9090/system/logs#/s=<some_long_identfier>`
4. Button back - url: `https://localhost:9090/system/logs` OK
5. Button back - url:  `https://localhost:9090/system/logs#/s=<some_long_identfier>` WRONG
6. Button back - url: `https://localhost:9090/system/logs`
... and it is infinite loop between these two.

martinpitt:
Ack, thanks for the heads-up!

martinpitt:
This issue is too broad and too generic. Needs to be reported for particular pages/situations.

garrett:
It seemed a lot of the issues are stemming from `cockpit.navigation.go` and `cockpit.jump` and fixing those would fix most, if not all, of the back button issues?

martinpitt:
OK, making this issue specific to the Logs page described above.

marusak:
Ok, so I look into this and the issue is as follows:
We have `hashchanged` event listener [here](https://github.com/cockpit-project/cockpit/blob/master/pkg/shell/base_index.js#L345) that adds the new url (with hash) into history. That means the following:
[...]

PR #16637, opened 2021-11-23
add a standalone mode for cockpit-client

allisonkarlitskaya:
Turn cockpit-client into a standalone app which is capable of spawning `cockpit-ws`for itself.  Add desktop file and icons, conditionally installed based on a new `--enable-cockpit-client` config option.

 * [x] #16636


allisonkarlitskaya:
Not sure about the installation of `cockpit-client-ssh`: should we do that unconditionally or only in `--enable-cockpit-client` mode.

 - standalone cockpit-client (which is in libexec regardless) is useless without it
 - but on the other hand, this is very much flatpak-oriented.  it'll also work in a toolbox if you have flatpak-spawn installed there, though...

src/client/cockpit-client-ssh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/python3
```
martinpitt, line 1: This would be the new Python script -- it needs to be added to the `__requires_exclude_from` in tools/cockpit.spec.

src/client/cockpit-client
```diff
@@ -163,6 +184,45 @@ class ExternalCockpitWs:
...
+                                   preexec_fn=self.preexec_fn, pass_fds=[3])
+
+        host, port = self.socket.getsockname()
+        return f'http://{host}:{port}/'
+
+    def preexec_fn(self):
+        os.environ['XDG_CONFIG_DIRS'] = self.config_dir.name
+        os.environ['LISTEN_PID'] = str(os.getpid())
+        os.environ['LISTEN_FDS'] = str(1)
+        prctl(prctl.SET_PDEATHSIG, signal.SIGKILL)
```
martinpitt, line 226: OOI, why this big hammer? TERM not good enough?

src/client/cockpit-client
```diff
@@ -163,6 +184,45 @@ class ExternalCockpitWs:
...
+    def stop(self):
+        self.ws.kill()
+        self.config_dir.cleanup()
+
+    def write_config(self):
+        self.config_dir = tempfile.TemporaryDirectory(prefix='cockpit-client-', suffix='-etc')
+        os.mkdir(f'{self.config_dir.name}/cockpit')
+        with open(f'{self.config_dir.name}/cockpit/cockpit.conf', 'x') as cockpit_conf:
+            cockpit_conf.write(f'''# dynamically generated by cockpit-client
+                               [WebService]
```
martinpitt, line 238: Looks a little odd to indent the config file so much (but should be harmless). There's [textwrap.dedent()](https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent) for such cases.

(Feel free to entirely ignore)

src/client/cockpit-client-ssh
```diff
@@ -0,0 +1,83 @@
...
+    # time, and then it's our turn to say something before the next one
+    # will be sent.  therefore, it's impossible to over-read.
+    fp = sys.stdin.buffer
+    length = int(fp.readline())
+    # take advantage of empty channel name being json whitespace
+    return json.loads(fp.read(length))
+
+
+def do_authorize(challenge='*', **kwargs):
+    send_control(command='authorize', challenge=challenge, cookie='-', **kwargs)
```
martinpitt, line 30: Will this ever be used for external ws? If so, then this should really generate a random cookie and validate it in the response.

tools/cockpit.spec
```diff
@@ -459,6 +459,7 @@ Suggests: python3
 
 # prevent hard python3 dependency for cockpit-desktop, it falls back to other browsers
 %global __requires_exclude_from ^%{_libexecdir}/cockpit-client$
+%global __requires_exclude_from ^%{_libexecdir}/cockpit-client-ssh$
```
martinpitt, line 462: You can only set this once, so the original regexp needs to be extended instead. I suggest to do `client.*$` -- while it's equivalent to just `client`, it makes it explicit that we expect suffixes there.

martinpitt (changes requested):
I'm a bit on the fence about the client-ssh installation into deb/rpm; I currently lean towards "no", as this is still heavily in flux, completely untested and undocumented at this stage, and there's no obvious benefit for it right now.

src/client/cockpit-client-ssh
allisonkarlitskaya: I avoided unconditionally installing this.

src/client/cockpit-client
allisonkarlitskaya: To turn it around: would TERM help anything?  This is a stateless process and we want it to simply go away.  Using TERM seems like a good way to end up with a weird edge case where we have a cockpit-ws hanging around after the fact.

src/client/cockpit-client
allisonkarlitskaya: I had it left-aligned first, but the config parser seems happy enough...

I'll add dedent if you like

src/client/cockpit-client-ssh
allisonkarlitskaya: No.  It won't.  But I'll generate the random cookie anyway.

tools/cockpit.spec
allisonkarlitskaya: I just conditionalised the install to dodge the problem.

src/client/cockpit-client
martinpitt: Good point.

garrett:
As our back button doesn't work (#13091 for example, but also when navigating back to the login screen), I don't feel confident in having it in the desktop app yet.

We should prioritize back button fixes in general and then reconsider adding it to the app.

martinpitt (approved):
I have like a dozen different bugs with that, but I think all of these are better done as follow-ups. Thanks!

garrett (changes requested):
Three things:

1. In addition to removing the back/forward buttons (for now at least)

2. could we also show the description that we use from the Flatpak information (or something similar) on the login screen itself?

Currently:

Screenshot from 2021-11-24 12-16-09

In the guidelines, # 12:
https://www.patternfly.org/v4/components/login-page/design-guidelines

image

> 12. **Description (optional)**: Provides a description of the website or application.

I think it would be useful to explain what's going on, as we just have a "connect to" and an empty input.

3. We might want to have helper text on the input to explain what the field expects. https://www.patternfly.org/v4/components/form/design-guidelines#helper-text

I'd consider # 1 pretty important, and we should probably do # 2 and # 3, but those could be follow-ups. (Although they would be good to have before we publish on Flathub.)

allisonkarlitskaya:
I will do a PR to remove the back/forward buttons.  For the other stuff, a mockup would be very useful :)

PR #16641, opened 2021-11-24
static: show login form onpageshow

allisonkarlitskaya:
Currently, the login form is shown from the onload event, from a
function which immediately removes itself from running again.

This means that when we navigate backwards to the login page, we'll find
it in exactly the state we left it: with the spinner active and all
controls effectively dead.

Move the code for showing the login page to the pageshow event.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -301,8 +301,6 @@
     }
 
     function boot() {
-        window.onload = null;
-
         translate();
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Turns out this only helps cockpit-client.  No other browser (including Epiphany) has this problem: they all have a back button which goes directly back to the page before the login screen (eg: browser new tab page).

PR #16644, opened 2021-11-24
storaged: Change VDO package installation alert to a helper text

martinpitt:
Leftover from https://github.com/cockpit-project/cockpit/pull/16623#issuecomment-976407637

Looks like this now:

shot

### DIFF
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -837,7 +837,7 @@ export class VGroup extends React.Component {
                                   value: "block",
                                   choices: purposes
                               }),
-                    Message("info", cockpit.format(_("The $0 package will be installed to create VDO devices."), vdo_package),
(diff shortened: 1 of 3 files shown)


mvollmer (approved):

garrett (changes requested):
Better... but isn't the (i) icon supposed to be blue?

martinpitt:
@garrett : Maybe.. but that's what I meant earlier -- [HelperText](https://www.patternfly.org/v4/components/helper-text/#static-with-custom-icons) does not have an "info" variant, just "default" or "warning". I'll try for a bit to customize the color.

martinpitt:
As discussed, I switched that to the non-circled i icon, and kept it at the default color. Screenshot updated.

garrett (approved):
OK, with @martinpitt's help and a bunch of effort, I managed to test this live in a VM and it looks good to me! :+1: 

Thanks!

PR #16647, opened 2021-11-25
cockpit-client: add a reload button

allisonkarlitskaya:
..and the usual <Ctrl>R accelerator.  <Ctrl><Shift>R does a hard reload,
bypassing caches.


martinpitt (approved):
Thanks!

PR #16648, opened 2021-11-25
webpack-jumpstart: look past merge commit for --rebase

allisonkarlitskaya:
If we are rebasing and the HEAD commit is a merge, use the merged commit
ID to find the cached data, instead of the merge itself.  The subsequent
rebase will result in the merge commit itself being removed.

### DIFF
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -49,7 +49,14 @@ if ! git diff --quiet -- ':^test' ':^packit.yaml' ':^.github'; then
     exit 1
 fi
 
-tag="sha-$(git rev-parse HEAD)"
+# If we are rebasing and the HEAD commit is a merge, use the merged
+# commit ID to find the cached data, instead of the merge itself.
+if test -n "${rebase}" && git rev-parse --verify HEAD^2 >/dev/null; then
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Note: I proposed this PR with its head based on a not-up-to-date main (and even across a commit which updates `node_modules`)  So if everything goes green here, it proves that the idea is sound.

allisonkarlitskaya:
This might theoretically fail in the case where additional commits on main/ have been added after the PR was last pushed, resulting in an updated merge commit.  There's a very small window where packit might pick up that extra commit in its merge, and webpack-jumpstart not.  In the case, rebasing would likely include that extra commit which wasn't present at the time of the jumpstart build, resulting in a different tree checksum, and a failed jumpstart.

We could maybe fix that by doing a hard reset to the feature branch head commit before performing the rebase, but ... *shrug*

martinpitt (approved):
Cool, it works! That's a lot more elegant and less involved than I feared.

martinpitt:
This [f35 failure](http://artifacts.dev.testing-farm.io/54daa638-14e6-4030-84d0-bd97f2a9b066/) is an "honest" test failure now, everything else green. Niiice!

PR #16650, opened 2021-11-25
Remove TLS support from cockpit-ws

allisonkarlitskaya:
...let's do some pair programming with the bots.

### DIFF
--- a/containers/bastion/run.sh
+++ b/containers/bastion/run.sh
@@ -2,6 +2,12 @@
 
 set -ex
 
-/usr/libexec/cockpit-certificate-ensure
+wssockdir=/run/cockpit/wsinstance
+sha256nil=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+wssock=$wssockdir/${sha256nil}
 
-exec /usr/libexec/cockpit-ws
+mkdir -p "$wssockdir"
+/usr/libexec/cockpit-ws -U ${wssock} &
+
+/usr/libexec/cockpit-certificate-ensure --for-cockpit-tls
+exec /usr/libexec/cockpit-tls
--- a/src/bridge/cockpitpackages.c
+++ b/src/bridge/cockpitpackages.c
@@ -1214,12 +1214,7 @@ cockpit_packages_new (void)
 
   packages = g_new0 (CockpitPackages, 1);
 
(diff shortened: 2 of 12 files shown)


allisonkarlitskaya:
Some notes: we still use the certificate-finding code from cockpit-ws in order to serve up the ca.crt, if we have it.  That feels increasingly anachronistic: we may not even have access to scan for the certificate from cockpit-ws at this point, and with self-signed certs only valid for 1 year, installing the CA cert is less useful.  Also: did anyone ever use this feature?

Also: the redirect-to-HTTPS stuff being handled in cockpit-ws is seemingly increasingly ... odd.

If we do decide to drop that feature, then we could have some nice cleanup by moving the certificate finding code to live (only) in cockpit-certificate-ensure.  If not (and as a general observation) I'm not actually sure how helpful it is to remove TLS support from cockpit-ws.  It's not a whole lot of code, since most of the implementation is in cockpitwebserver.

Note: removing certificates from cockpitwebserver is not currently possible because we use cockpitwebserver from the tests for quite a lot of testing that cockpit-bridge's https connector is working.  We could remove that as well, of course, ...

Lots of questions raised by this PR.

In any case: I really like the restructure of CockpitWebService here to separate the port listening from the constructor, and also think the `-U` option might be useful, so we should land those in any case.

martinpitt:
> installing the CA cert is less useful

AFAIK it's still the only possible way to make Safari work. 

> we may not even have access to scan for the certificate from cockpit-ws at this point

That shouldn't change for this PR, though? The CA should remain world-readable, and we could make `locate_selfsign_ca()`  independent from any TLS support code, in particular cockpit_certificate_locate(). Could it literally call `/usr/libexec/cockpit-certificate-ensure --check` to avoid copying code? For this route I think it's okay to be a little slow. But right now that doesn't work as unpriv user as it fails on EPERM on the .key file (although we only really need to know the cert/ca). This might need another (undocumented) option/mode?

> we use cockpitwebserver from the tests for quite a lot of testing that cockpit-bridge's https connector is working

FTR, of course I'm fine to remove that -- we could use [python](https://www.piware.de/2011/01/creating-an-https-server-in-python/) or `gnutls serve` or `openssl s_client` or something similar?

allisonkarlitskaya:
I think we're sort of increasingly concluding that `cockpit-ws`, as it is, might be on the way out.  No need to change this now...

PR #16657, opened 2021-11-26
workflows: use merge instead of head for flatpak test

allisonkarlitskaya:
flatpak-test.yml checks out the head commit instead of the merge commit.
That was done to dodge a bug in webpack-jumpstart which caused it to
fail on merge commits.  That's been fixed, meanwhile, and we're also not
using jumpstart anymore anyway, and now this workaround is causing new
issues, like the failure in #16624.

TL;DR: let's just use the default and checkout the merge commit, since
that's what we're supposed to be testing anyway.


martinpitt (approved):
Danke!

PR #16659, opened 2021-11-27
testlib: Correctly deal with older shell versions

marusak:
In #16522 a new shell was introduced. This break on new images that
already have cockpit-system-258 installed.

We will need to change then test/common checkouts in all subprojects to include this fix - as now seen for example in https://github.com/cockpit-project/bots/pull/2669 or failing rawhide packit in machines or podman.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -109,7 +109,7 @@ def attach(filename, move=False):
 
 
 class Browser:
-    def __init__(self, address, label, pixels_label=None, port=None):
+    def __init__(self, address, label, machine, pixels_label=None, port=None):
         if ":" in address:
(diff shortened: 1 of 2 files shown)


martinpitt:
AttributeError: 'Browser' object has no attribute 'system_before'

martinpitt:
This is a bit awkward indeed - the Browser class does not know about the MachineCase instance. Moving the method and changing all `b.logout()` to `self.logout()`  seems a bit intrusive -- how about we pass the Machine object to Browser on instantiation? That may be useful for other cases, I think I ran into a similar problem in the past already. And there's only two places which instantiate Browser.

martinpitt (approved):
Thanks! Re-triggering flake

PR #16662, opened 2021-11-28
test: Add Arch support to MachineCase.system_before()

martinpitt:
This unblocks https://github.com/cockpit-project/cockpit-podman/pull/838 and similar ones.

I tested this locally against cockpit-podman.


martinpitt:
I just saw that @jelly did almost the same in https://github.com/cockpit-project/cockpit/pull/16547/commits/7d83045b65bb402324df94c6ff626c2fa69191b9#r757888474 - so I think using `pacman -Q cockpit` is correct.

marusak (approved):

PR #16665, opened 2021-11-29
lib: remove listingActions property from ListingPanel component

KKoukiou:
This was serving a design that cockpit was using heavily till now, and
namely the actions of table rows being placed next to the expanded row tabs selector.

However, the suggested design from patternfly is to put actions behind a
kebab in table rows.

So, let's remove this option, so that we don't promote it's future usage.

Note: Last consumer is cockpit-machines and I am working on the porting right now.


martinpitt (approved):
Thanks! Do you want to block landing this until the c-machines port lands? So that we don't get blocked on updating pkg/lib?

KKoukiou:
> Thanks! Do you want to block landing this until the c-machines port lands? So that we don't get blocked on updating pkg/lib?

No, I worked on it on friday, the commit is almost ready.

PR #16666, opened 2021-11-29
tests: autoconvert all python string formatting to fstring

KKoukiou:
Let's make this string formatting method the only one to use from now
on.

These files were autoconverted with flynt.


martinpitt (approved):
Thanks! I wonder if we can teach pycodestyle to refuse % and .format() from now on?

Let's land this ASAP as it's very intrusive.

KKoukiou:
> Thanks! I wonder if we can teach pycodestyle to refuse % and .format() from now on?
> 
> Let's land this ASAP as it's very intrusive.

I tried to find such an option without success unfortunately.

PR #16667, opened 2021-11-29
shell: remove patternfly 3 *last* import

KKoukiou:


garrett (approved):
This looks great! Thanks!

KKoukiou:
This is a bit blocked on some stupid pixel tests almost impossible to notice differences.

marusak (approved):
Awesome, thank you!

martinpitt (approved):
Thanks! The pixel diffs are very subtle indeed, I only saw a slight shift in the terminal one.

martinpitt:
Note that there is one place left [here](https://github.com/KKoukiou/cockpit/blob/7c8c45c0008cf1b64485ce43d76799e0629e640e/pkg/shell/shell.html#L12):
```html
        <div class="blank-slate-pf-icon">
            <span class="fa fa-meh-o"></span>
        </div>
```
This icon will be broken now. It's very obscure, and I'm not even sure it's still relevant, and not having an icon there is certainly not a big deal. So it can certainly become a follow-up, without blocking the pixel tests.

martinpitt:
I sent PR #16690 to clean up the above leftover.

PR #16668, opened 2021-11-30
bridge: remove "ssh-agent" internal address

allisonkarlitskaya:
This used to be used when cockpit-ws originated ssh connections to other
hosts on behalf of the session, but that hasn't happened since
082da24a5a2a27d521e7b6d1beb2bbe094ffdc54.  Nowhere else in the code
refers to this anymore, so let's drop it.


martinpitt (approved):
Thanks!

PR #16669, opened 2021-11-30
clean up "internal" endpoints

allisonkarlitskaya:
Major changes:
 * cockpit-bridge no longer opens a TCP port for serving packages
 * we no longer have the `internal` connection option
 * `internal: "packages"` is preserved as a one-off exception on HTTP streams for reasons of cockpit-ws/cockpit-bridge compatibility
 * cleanup of listening options on cockpitwebserver, ~and cockpit-ws gains a -U option to listen on a unix socket~ (from #16650)

### DIFF
--- a/src/common/cockpitwebserver.c
+++ b/src/common/cockpitwebserver.c
@@ -22,9 +22,10 @@
 #include "cockpitwebserver.h"
 
 #include "cockpithash.h"
-#include "cockpitmemory.h"
-#include "cockpitwebresponse.h"
 #include "cockpitmemfdread.h"
+#include "cockpitmemory.h"
+#include "cockpitsocket.h"
+#include "cockpitwebresponse.h"
 
 #include "websocket/websocket.h"
 
@@ -1293,6 +1294,21 @@ cockpit_web_server_start (CockpitWebServer *self)
 {
   g_return_if_fail (COCKPIT_IS_WEB_SERVER (self));
   g_socket_service_start (self->socket_service);
+}
+
+GIOStream *
+cockpit_web_server_connect (CockpitWebServer *self)
+{
+  g_return_val_if_fail (COCKPIT_IS_WEB_SERVER (self), NULL);
+
+  g_autoptr(GIOStream) server = NULL;
+  g_autoptr(GIOStream) client = NULL;
+
+  cockpit_socket_streampair (&client, &server);
+
+  cockpit_request_start (self, server, TRUE);
+
+  return g_steal_pointer (&client);
 }
 
 /* ---------------------------------------------------------------------------------------------------- */
--- a/src/common/cockpitwebserver.c
+++ b/src/common/cockpitwebserver.c
@@ -35,8 +35,6 @@
 #include <string.h>
 #include <unistd.h>
 
-#include <systemd/sd-daemon.h>
-
 /* Used during testing */
 gboolean cockpit_webserver_want_certificate = FALSE;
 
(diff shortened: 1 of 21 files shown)


src/common/cockpitwebserver.c
```diff
@@ -1296,116 +1222,96 @@ on_incoming (GSocketService *service,
...
+                                      const gchar *address,
+                                      guint16 port,
+                                      GError **error)
+{
+  if (address != NULL)
     {
-      /* We got file descriptors passed in, use those. */
-
-      for (fd = SD_LISTEN_FDS_START; fd < SD_LISTEN_FDS_START + n; fd++)
+      g_autoptr(GSocketAddress) socket_address = g_inet_socket_address_new_from_string (address, port);
```
martinpitt, line 1244: It's legit to call that with `0`? https://docs.gtk.org/gio/ctor.InetSocketAddress.new_from_string.html does not mention that case.

src/ws/main.c
```diff
@@ -53,6 +54,7 @@ static gboolean  opt_version      = FALSE;
 static GOptionEntry cmd_entries[] = {
   {"port", 'p', 0, G_OPTION_ARG_INT, &opt_port, "Local port to bind to (9090 if unset)", NULL},
   {"address", 'a', 0, G_OPTION_ARG_STRING, &opt_address, "Address to bind to (binds on all addresses if unset)", "ADDRESS"},
+  {"unix", 'U', 0, G_OPTION_ARG_STRING, &opt_unix, "Listen on this unix address", "PATH"},
```
martinpitt, line 57: Also needs to go into doc/man/cockpit-ws.xml . It's a tiny bit sad to see features being added to ws, while we wanted to get rid of it, but it's also nice to clean up the TCP parts.

martinpitt (changes requested):
This is amazing, thanks! I mostly just have two questions.

"request changes" for the unit test and compile errors and such.

src/ws/main.c
martinpitt: What is that for, OOI? the current patch series does not seem to actually use it at all, the "get socketpair connection" seems much more elegant anyway?

src/common/cockpitwebserver.c
allisonkarlitskaya: It must be.  `0` is a completely legit port number to have in an inet socket address structure, with a well-defined meaning: assign a port for me.  Also: it works :)

Note that the old code used to call

```
socket_address = g_inet_socket_address_new (server->address, server->port);
```

with a potentially-zero port.

src/ws/main.c
allisonkarlitskaya: This is eventually meant to enable cockpit-ws on a unix socket inside of the container in order to enable it to be used with cockpit-tls.  We could defer this patch until we actually deal with #16650 if you prefer.

src/common/cockpitwebserver.c
martinpitt: Tought as much, but good to confirm. Thanks!

src/ws/main.c
martinpitt: It looks fine, I was just curious. As it's currently completely untested, I slightly prefer postponing it. But if it's in any way useful for you already, then let's just add it to the manpage and land it.

allisonkarlitskaya: I dropped it for now.  Let's handle it in #16650.

martinpitt (approved):
Thanks!

PR #16670, opened 2021-11-30
enable dbus activation for cockpit-client

allisonkarlitskaya:
...I thought I'd never see the day.  :rofl:


martinpitt (approved):
Wohoo! I can't test flatpak activation here (but that's not your fault, does not work for gnome clocks either), but I trust the integration test.

PR #16671, opened 2021-11-30
client: add win.pwn action

allisonkarlitskaya:
Add a "pwn" action on the window that will execute arbitrary code and
set the state of the action to the result.

You can use it like this:

```
  gdbus monitor --session --dest=org.cockpit_project.CockpitClient &

  gdbus call --session --dest=org.cockpit_project.CockpitClient \
        -o /org/cockpit_project/CockpitClient/window/1
        -m org.gtk.Actions.Activate \
        '"pwn"' '[<"2+2">]' '[]'
```

and see

```
  /org/cockpit_project/CockpitClient/window/1: org.gtk.Actions.Changed (@as [], @a{sb} {}, {'pwn': <'4'>}, @a{s(bgav)} {})
```

Right now it runs Python, but obviously it should run Javascript in the
future.

### DIFF
--- a/src/client/cockpit-client
+++ b/src/client/cockpit-client
@@ -62,6 +62,7 @@ class CockpitClientWindow(Gtk.ApplicationWindow):
             ('go-forward', self.go_forward),
             ('zoom', self.zoom, 's'),
             ('open-inspector', self.open_inspector),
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
this wasn't meant to land, so let's look at #16676 instead.

PR #16676, opened 2021-12-01
flatpak: Add browser integration smoke test

martinpitt:
First draft of how that could look like.

 - [x] Fix tests to run in xvfb/dbus-test-runner (in GitHub workflow)
 - [x] Fix the `sleep 3` and properly wait for page load
 - [x] Introspect the /system frame in the browser test (simple frame API)
 - [x] Do some action in the overview
 - [x] move test stuff into subdirectory
 - [x] Use GApplication action states instead of writing .result files, as in PR #16671 ; requires rewriting test in Python
 - [x] Use [webkit user content manager](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitUserContentManager.html#webkit-user-content-manager-register-script-message-handler) to pass result


martinpitt:
@allisonkarlitskaya: I am reasonably happy about this initial structure and scope of the new tests; see the todo list.

Using an action state like in #16671 is indeed more elegant than returning the result in a file. However, it is also a lot harder to use -- Neither `gapplication` nor `busctl` nor `dbus-monitor` nor `gdbus` have a way to wait for a particular D-Bus signal. Doing that polling and parsing the result from `busctl monitor` in shell is way too unwieldy IMHO.

It's certainly possible and halfway decent in Python. I'll see how that looks like. We could then also sensibly pass the JS code as string instead of a file, without running into quoting issues.

allisonkarlitskaya:
> It's certainly possible and halfway decent in Python. I'll see how that looks like. We could then also sensibly pass the JS code as string instead of a file, without running into quoting issues.

https://docs.gtk.org/gio/class.DBusActionGroup.html

martinpitt:
@allisonkarlitskaya : Thanks! I used the GApplication PyGI wrappers after all, they are a bit more comfortable than raw D-Bus. I find this quite pleasant now. I added the action state/Python rewrite as a separate FIXUP as I wanted to see how it fared on CI, but it works just fine. It's also fairly pleasant to run locally now, and we have plenty of extension points for things like `--sit` or making screenshots and such.

Do you want to have a first coarse review to see if you are d'accord with the general approach? Then I'll rebase this properly, let all tests run, etc.

allisonkarlitskaya (changes requested):
I don't like the polling approach.

I'd suggest either:
 - removing it to outside of the process by having the javascript synchronous result immediately posted to the action state, and using polling from outside to query the result variable
 - `webkit_web_view_get_user_content_manager()` → `webkit_user_content_manager_register_script_message_handler()` to give a way for JS world to send results to us, without polling.  https://webkitgtk.org/reference/webkit2gtk/stable/WebKitUserContentManager.html#webkit-user-content-manager-register-script-message-handler has an example of that.

martinpitt:
> using polling from outside to query the result variable

I don't like this honestly. You also have to expose the page-load events as signal then, and I don't see how to do "result polling OR pick up an async load signal" from outside in a race-free manner. The current version is simpler and much more robust.

> webkit_web_view_get_user_content_manager

That OTOH sounds like a very promising idea, thanks for the pointer! I'll look into this.

martinpitt:
I reworked this as above. There are still some beautifications to do (in particular, providing a helper to post the result). On the face of it, this is much more elegant indeed, but somehow it makes the entire test horribly flaky. In

    env -u WAYLAND_DISPLAY xvfb-run dbus-run-session containers/flatpak/test/test-browser

this works fine (mostly), but when I run it interactively, I often get failures like "timed out waiting for .system-information". With adding a few console.log()s, I see that the overview frame document is essentially empty. It feels like installing the message handler somehow messes up the JS state.

Let's see what CI thinks about this, but not being able to run the test interactively is a big no-go for me.

martinpitt:
I am happy with the structure now. The [old FIXUP head](https://github.com/cockpit-project/cockpit/commit/812a167632fc745a10b03ef9818b9cb5c861d571) is still in git's brain for archaeology, but I rebased this now to clean up.

I also dropped the commit to disable unit and packit tests.

.github/workflows/flatpak-test.yml
```diff
@@ -31,4 +31,4 @@ jobs:
       - name: Smoke-test the installed flatpak
         run: |
           . /etc/profile.d/flatpak.sh
-          xvfb-run dbus-run-session containers/flatpak/test-ssh
+          xvfb-run dbus-run-session containers/flatpak/test/test-ssh
```
allisonkarlitskaya: Can you just combine these two changes into a single commit?

src/client/cockpit-client
```diff
@@ -114,6 +115,55 @@ class CockpitClientWindow(Gtk.ApplicationWindow):
     def open_inspector(self, *unused):
         self.webview.get_inspector().show()
 
+    def run_js(self, action, parameter, *unused):
```
allisonkarlitskaya, line 118: So here's my counter-proposal:

 - add a `--enable-run-js` commandline argument which sets `self.enable_run_js = True` on the app
 - check `app.enable_run_js` when constructing the window and in case it's set:
   - install the action
   - register the script message handler → it will just always set the action state
   - register the page-load handler → it will just always set the action state
   - never unregister anything

or is it actually required to register the handler each time separately because the JS context got destroyed in between?

allisonkarlitskaya (changes requested):

.github/workflows/flatpak-test.yml
martinpitt: Done.

src/client/cockpit-client
martinpitt: I added the `--enable-run-js` hidden option. I'll look into keeping the handler next week. I am rather sure the context gets destroyed between page loads (that's why I have to split up the test so heavily in the first place), but I'll experiment with that -- would certainly make things easier.

martinpitt: I pushed a fixup which implements your suggestion. Honestly I don't like it that much, as it smears out the run-js functionality across the code, and it is overall an extra three lines of code. Also, it does not really simplify things other than the two lines of signal handler cleanup -- but that also makes the action state a lot more muddy and potentially racy.

It does work to leave the handlers permanently, though.

martinpitt:
I have no end of trouble right now with running this  locally:
```
  File "/var/home/martin/upstream/cockpit/containers/flatpak/test/test-browser", line 39, in setUp
    self.assertIn("run-js", actions)
AssertionError: 'run-js' not found in {'go-back', 'reload', 'go-forward', 'open-inspector', 'reload-bypass-cache', 'zoom'}
```

It seems the `--enable-run-js` command line argument gets ignored entirely (I don't see it in `ps`), and it's somehow running the D-Bus activated command line instead? If I replace the `if app.enable_run_js` with `if True`, it works fine. I already tried to add `--disable-uniqueness`, but that does not help.

Curiously the same thing [happens in CI now](https://github.com/cockpit-project/cockpit/actions/runs/1543352135).

src/client/cockpit-client
allisonkarlitskaya: Okay.  After looking at the diff, I agree that it's not awesome.  Let's do it your way.

But, one detail that might make this seem a bit nicer to me:

install the `run-js` action unconditionally at the same time as all of the others, but directly under the call to `add_action_entries()` put something like:

`self.lookup_action('run-js').set_enabled(app.enable_run_js)`

martinpitt: Done, thanks for the hint!

allisonkarlitskaya (approved):
This is getting pretty cool.  Thanks :)

PR #16677, opened 2021-12-01
storage: Don't use Patternfly table actions

mvollmer:
They have a broken layout in mobile mode, see https://github.com/patternfly/patternfly/issues/4547

Instead, move them to their own unlabeled row at the bottom.


mvollmer:
Looks like this now:

image

garrett:
Hmm, yeah, that's not ideal either. I guess I might need to adjust things a bit. But it's getting late today, so I might have to do that on Monday.

mvollmer:
See #16681 for an alternative.

mvollmer:
I like #16681 so much more, I close this one.

PR #16679, opened 2021-12-01
handle http→https redirect from cockpit-tls

allisonkarlitskaya:
This is a temporary increase in code size, since we still need to keep http→https redirection in `cockpit-ws` for the container case, but it's a precondition to removing it.

 * [ ] read the parser code very very very carefully

### DIFF
--- a/src/tls/httpredirect.c
+++ b/src/tls/httpredirect.c
@@ -1,0 +1,176 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2019 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "config.h"
+
+#include "httpredirect.h"
+
+#include <sys/socket.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <string.h>
+#include <unistd.h>
+
+static char *
+read_line (FILE *stream,
+           char *buffer,
+           size_t sizeof_buffer,
+           size_t *offset)
+{
+  if (*offset >= sizeof_buffer)
+    return NULL;
+
+  char *line = fgets (buffer + *offset, sizeof_buffer - *offset, stream);
+
+  if (line == NULL)
+    return NULL;
+
+  /* Make sure we're terminated with \r\n or \n */
+  size_t line_length = strcspn (line, "\r\n");
+  const char *ending = line + line_length;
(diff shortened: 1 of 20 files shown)


allisonkarlitskaya:
I made this a thread, for efficiency reasons, but we have a few options to improve isolation:
 - `fork()` instead of creating a thread
 - continue to use a systemd unit to respond to these requests
 - either of the above with lots of isolation and/or unshare and/or seccomp sprinkled on top

src/tls/httpredirect.c
```diff
@@ -0,0 +1,176 @@
...
+               FILE *output)
+{
+  char buffer[10000];
+  size_t offset = 0;
+
+  char *request_line = read_line (input, buffer, sizeof buffer, &offset);
+  if (request_line == NULL)
+    return write_error (output);
+
+  char *path = strchr (request_line, ' ');
```
martinpitt, line 80: This entirely ignores the first word, i.e. the method (GET/PUT/POST etc.). Indeed that's not actually relevant for the behaviour here, and I think it's ok to send a redirect even for unsupported/bogus methods. I can't poke any holes in this, so let's keep it simple indeed.

src/tls/httpredirect.c
```diff
@@ -0,0 +1,176 @@
...
+        return write_error (output);
+
+#define HOST_HEADER "Host:"
+      if (strncmp (header, HOST_HEADER, strlen (HOST_HEADER)) == 0)
+        {
+          if (host != NULL)
+            return write_error (output);
+
+          host = header + strlen (HOST_HEADER);
+          host += strspn (host, " \t");
```
martinpitt, line 105: `host` can be an empty string here. Again, that's just throwing garbage back to the sender if they send us garbage, and I don't think it'd break anything.

src/tls/httpredirect.c
```diff
@@ -0,0 +1,176 @@
...
+  char *line = fgets (buffer + *offset, sizeof_buffer - *offset, stream);
+
+  if (line == NULL)
+    return NULL;
+
+  /* Make sure we're terminated with \r\n or \n */
+  size_t line_length = strcspn (line, "\r\n");
+  const char *ending = line + line_length;
+  if (strcmp (ending, "\r\n") != 0 && strcmp (ending, "\n") != 0)
+    return NULL;
```
martinpitt, line 49: Initially I thought "the `\r` check is completely insignificant here". Reading it again I realized it's not -- it makes sure that there is no `\r` in the *middle* of the line. Can you please amend the comment to say ".. and that there are no stray \r"?

Then again I wonder why we specifically handle `\r` but not any of the other ASCII control characters -- if they don't break the logic, why would `\r`? Sure, that could freak our clients when we throw their garbage back and them, but that's not our problem? Also, it feels like that doesn't belong into a generic function like `read_line()`.

IOW,  what would/could break if we changed this to
```C
if (line [strlen (line) - 1] != '\n')
    return NULL;
```

allisonkarlitskaya: `\r` is special here because it's not just whitespace.  the http spec says that you should send `\r\n` to terminate lines.  I tolerate bare '\n` as a courtesy to those who connect manually, but by spec, we actually don't need to.

Having `ending` set to the logical end of the data is important here: we need to trim the `\r` following `Host: foo`, for example, before we paste it together to make our redirection location and end up with `https://foo\r/path`.

allisonkarlitskaya: On the other hand, it's possible also to imagine that the field could theoretically contain other trailing whitespace, and we would then want to trim that as well, I guess.  Nobody ever actually sends that in practice, of course...

martinpitt (approved):
Thanks! Still not a big fan of the \r handling, but this isn't a general-purpose library, and it's not worth stalling over this. I retried the r85-distropkg test, almost surely a flake.

PR #16681, opened 2021-12-02
storage: Hide table buttons in narrow layouts

mvollmer:
The main goal is to avoid https://github.com/patternfly/patternfly/issues/4547 by not having buttons in the mobile mode in the top right corner. We (almost) always have an expander on the row, which already takes space, and thus having a menu in mobile mode is free for us. And I like it in the top right corner. :-)

Demo: https://www.youtube.com/watch?v=-BSB-DN4yBM

### DIFF
--- a/pkg/storaged/storage.scss
+++ b/pkg/storaged/storage.scss
@@ -425,3 +425,15 @@ td.ct-text-align-right {
 td button.pf-m-link {
     padding: 0px;
 }
+
+@media ( max-width: $pf-global--breakpoint--md - 1 ) {
+    .show-only-when-wide {
+        display: none;
+    }
+}
+
+@media ( min-width: $pf-global--breakpoint--md ) {
+    .show-only-when-narrow {
+        display: none;
+    }
+}
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -128,8 +128,29 @@ function create_tabs(client, target, is_partition, is_extended) {
     if (content_block)
         warnings = warnings.concat(client.path_warnings[content_block.path] || []);
 
+    const tab_actions = [];
+    const tab_menu_actions = [];
+    const tab_menu_danger_actions = [];
+
+    function add_action(title, func) {
+        tab_actions.push(<StorageButton onlyWide key={title} onClick={func}>{title}</StorageButton>);
+        tab_menu_actions.push({ title: title, func: func, only_narrow: true });
+    }
+
+    function add_danger_action(title, func) {
+        tab_actions.push(<StorageButton onlyWide key={title} onClick={func}>{title}</StorageButton>);
+        tab_menu_danger_actions.push({ title: title, func: func, only_narrow: true });
+    }
+
+    function add_menu_action(title, func) {
+        tab_menu_actions.push({ title: title, func: func });
+    }
+
+    function add_menu_danger_action(title, func) {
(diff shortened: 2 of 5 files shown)


mvollmer:
This is an alternative to #16677.

mvollmer:
Looks like this:

image

mvollmer:
The second commit also hides the "Create partition" button and gives that row a permanent menu:

image

image

pkg/storaged/storage.scss
```diff
@@ -439,3 +439,12 @@ td.ct-text-align-right {
     // Overlay at 1/3  to mix the light and dark against the trough color
     opacity: 0.333;
 }
+
+$phone: 767px;
```
KKoukiou, line 443: the variabls from pkg/lib/_global-variables.scss should be used here.

mvollmer: I got this from pkg/shell/shell.scss, so let's update that as well.

mvollmer: Done.  But I think the storage page does not include pkg/lib/__global-variables.scss, since the breakpoint is still at 768px.

KKoukiou: yeap you can include it with: :@use "_global-variables.scss" as *;

mvollmer: storage.scss already has `@import "global-variables";` and your `@use` doesn't seem to have any effect, the breakpoint stays at 768px. So *shrug*, let's leave things as they are.

garrett:
> The second commit also hides the "Create partition" button and gives that row a permanent menu:
>
> image

It's weird to have the action duplicated like this, isn't it?

mvollmer:
> It's weird to have the action duplicated like this, isn't it?

Yes, I agree, especially since there is nothing else in the menu.

garrett:
> Yes, I agree, especially since there is nothing else in the menu.

In this case, it would be better to not show the item in the menu, if there's more than 1 item, and if there's just the same item, don't show the menu kebab icon (just the button).

mvollmer:
Let's go with the third option, actions move from a button into the menu.  I have made a demo for that: https://www.youtube.com/watch?v=-BSB-DN4yBM

pkg/storaged/storage.scss
```diff
@@ -439,3 +439,15 @@ td.ct-text-align-right {
     // Overlay at 1/3  to mix the light and dark against the trough color
     opacity: 0.333;
 }
+
+@media ( max-width: $pf-global--breakpoint--md - 1 ) {
```
KKoukiou, line 443: Seems like Pf4 offers already classes for this https://github.com/patternfly/patternfly/blob/main/src/patternfly/sass-utilities/placeholders.scss#L70
pf-m-hidden-on, pf-m-visible-on.

mvollmer: Can you show how to use them?

KKoukiou: From the examples I see, `pf-m-hidden pf-m-visible-on-md` if you want to show the icon for more than md breakpoint. (Did not try it)

mvollmer: Thanks, I'll try that!

mvollmer: I can't get this to work, sorry.  And since it is not documented, let's not use this.

KKoukiou: Ok no worries about it.

KKoukiou (approved):

PR #16682, opened 2021-12-02
users: use superuser for `useradd -D`

jelly:
On Arch Linux the useradd file `/etc/default/useradd` is only readable
by root, so any SHELL override is not picked up by `useradd -D`.


martinpitt (approved):
Thanks!

PR #16684, opened 2021-12-05
Translation: Simplified Chinese

WhymustIhaveaname:
cockpit's Chinese translation is much better than 2 years ago when I first knew it. However, it can be better.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -15,6 +15,7 @@
 # Pany <geekpany@gmail.com>, 2020.
 # Matej Marusak <mmarusak@redhat.com>, 2020.
 # Charles Lee <lchopn@gmail.com>, 2020.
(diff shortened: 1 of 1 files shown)


martinpitt:
Thanks @WhymustIhaveaname ! Can you please make these changes here: https://translate.fedoraproject.org/projects/cockpit/main/zh_CN/ 

I'm afraid we can't process translations as pull requests. I tried hard to upload your updated zh_CN.po file to weblate, but that just doesn't want to work.

martinpitt:
I copied all these translations into the weblate UI now. Thanks! (Next time, please do it there right away, to save both you and us some trouble).

WhymustIhaveaname:
Sorry, I was too busy for the last two weeks. I want to do it today XD.

PR #16687, opened 2021-12-06
ws: add --accept-session argument

allisonkarlitskaya:
This causes cockpit-ws to listen on the given port for an incoming
connection, which it will use in place of spawning cockpit-session.

1.
```
[root@fedora-34-127-0-0-2-2201 ~]# /usr/libexec/cockpit-ws --accept-session 1234 --no-tls
```

2.
```
[root@fedora-34-127-0-0-2-2201 ~]# socat tcp:localhost:1234 exec:/usr/libexec/cockpit-session
```

### DIFF
--- a/src/ws/cockpitauth.c
+++ b/src/ws/cockpitauth.c
@@ -254,6 +254,11 @@ cockpit_auth_finalize (GObject *object)
   CockpitAuth *self = COCKPIT_AUTH (object);
   if (self->timeout_tag)
     g_source_remove (self->timeout_tag);
+  if (self->session_service)
+    {
+      g_socket_service_stop (self->session_service);
+      g_object_unref (self->session_service);
+    }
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
Many many many TODOs here:

- if there's no pending connection, we should wait for it instead of just failing the auth
- we need some kind of a multiple-connection model with a control channel and payload channels where we can instruct the other side to start a new cockpit-session instance and connect it to us.
- closing the socket at the end of the session doesn't seem to work properly at the moment
- authenticating the incoming connection (although maybe we can get openshift to deal with that for us as part of the edge termination) to avoid sending the root password to whoever randomly connected

allisonkarlitskaya:
Dropping this in favour of less hacks: https://github.com/allisonkarlitskaya/cockpit-cloud

PR #16689, opened 2021-12-07
test: Robustify TestCurrentMetrics.testMemory

martinpitt:
In the "have swap" code path of this test it often happens that the
memory usage after killing both memhogs does not go back quite to the
same number as at the start of the test. In our noisy test VM this could
have any number of reasons. So make the check less strict by measuring
against the usage before starting mem-hog2 (which is higher, as memhog
is already running). The point of the check is that current memory usage
goes down after stopping the services, but we can't pinpoint an exact
number.

----

Fixes [this failure](https://logs.cockpit-project.org/logs/pull-16686-20211206-123627-ca37ceaf-fedora-35/log.html#291-1), which pretty much always happens on Fedora 35 in tests which touch metrics (and thus have to survive 3x affected retries)

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -784,6 +784,8 @@ class TestCurrentMetrics(MachineCase):
         b.wait_visible("table[aria-label='Top 5 memory services']")
 
         if have_swap:
+            usage_hog1 = progressValue(self, "#current-memory-usage")
+
             # use even more memory to trigger swap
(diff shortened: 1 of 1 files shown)


marusak (approved):
Hmm yeah, but it can go down for numerous reasons, not just because you killed the hog. But this is always somewhat flaky, so good enough, thanks!

PR #16695, opened 2021-12-07
rhel-8-5: Utilize PF4 components for the superuser dialog

### LINKED ISSUE #16688: Administrative Access MFA Prompt Window Broken
Cockpit version: 251
OS: Red Hat  8.5
Page: Administrative Access

Our Red Hat servers are configured to use Duo for multifactor authentication.  In prior versions, we were able to log in as a non-root user and click "Turn on administrative access" and be greeted by a dialog box where we can press a number to select what method we would like Duo to use to authenticate us (push notification to phone #1, push notification to phone #2, call to phone #1, call to phone #2, etc.).  

In the 251 version, which appears to be the latest version available via Red Hat repositories, the dialog box that would allow us to enter a number and initiate the Duo request is not visible.  In addition, the dialog text is cut off on the screen.  

This is how it looks currently.  As you can see, the dialog info is cutoff and no dialog box is there to enter the number.  As a result, it is impossible to authenticate properly and elevate to admin permissions as a non-root user.  This was tested and worked in prior versions.
 
image

mvollmer:
Fixes #16688

### DIFF
--- a/pkg/shell/superuser.jsx
+++ b/pkg/shell/superuser.jsx
@@ -19,12 +19,10 @@
 
 import cockpit from "cockpit";
 import React from "react";
-import { Alert, Button, Modal } from '@patternfly/react-core';
+import { Alert, Button, Form, FormGroup, Modal, TextInput } from '@patternfly/react-core';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import { host_superuser_storage_key } from './machines/machines';
 import { LockIcon } from '@patternfly/react-icons';
-
-import "form-layout.scss";
 
 const _ = cockpit.gettext;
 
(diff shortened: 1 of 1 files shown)


garrett:
The fix exists, is in production in newer versions of Cockpit already, and it's affecting our customers as-is in RHEL right now. We should land this ASAP. I was thinking it makes sense as a release-blocker... but our release already does have this code, so it's a release for RHEL, not a general release. (Hence me adding release-blocker and removing it.)

martinpitt:
This was backported from commit 038ecc8fb0441adf4026558766a5c3483e2da98a (sans the pixel test), so adding backport label.

I think we can safely ignore the packit fedora-33 failure, this is only interesting for rhel anyway (i.e. we should drop packit.yaml from this branch).

Thanks! I'm not sure if we will actually be allowed to land this in 8.5.z, but it doesn't hurt to be prepared.

martinpitt (approved):

garrett:
> Thanks! I'm not sure if we will actually be allowed to land this in 8.5.z, but it doesn't hurt to be prepared.

Right now, someone using RHEL can't sign in, so I'd consider it to be urgent. It could be argued for, IMO.

PR #16699, opened 2021-12-07
test: Close port 8765 to update the pixel tests

jelly:
The tests no longer require port 8765 to be opened as the dbus tests now
run locally in the virtual machine. The port is closed in cockpit, so
the pixel tests can be updated and not opening the ports is an easier
change in bots.

### DIFF
--- a/test/verify/check-networkmanager-firewall
+++ b/test/verify/check-networkmanager-firewall
@@ -64,6 +64,14 @@ class TestFirewall(NetworkCase):
         if m.image not in ["ubuntu-2004"]:
             m.execute("if virsh net-list --name | grep -q default; then until firewall-cmd --get-active-zones | grep -q libvirt; do sleep 1; done; fi")
 
+        # Temporary remove the test Dbus port 8765, this workaround can be
+        # removed once the pixel tests are merged and the image updated to not
+        # open port 8765
+        if m.image.startswith("fedora"):
+            m.execute("firewall-cmd --permanent --remove-port=8765/tcp")
+            m.execute("firewall-cmd --remove-port=8765/tcp")
+            m.execute("firewall-cmd --reload")
+
     def testNetworkingPage(self):
         b = self.browser
         m = self.machine


test/verify/check-networkmanager-firewall
```diff
@@ -55,6 +55,10 @@ class TestFirewall(NetworkCase):
         m = self.machine
         self.restore_dir("/etc/firewalld", "systemctl try-restart firewalld")
         m.execute("systemctl start firewalld")
+        # Temporary remove the test Dbus port 8765, this workaround can be
+        # removed once the pixel tests are merged and the image updated to not
+        # open port 8765
+        m.execute("firewall-cmd --remove-port=8765/tcp")
```
martinpitt, line 61: My first guess is that you also need to do this with `--persistent`. Supposedly the page shows the persistent configuration?

test/verify/check-networkmanager-firewall
```diff
@@ -64,6 +64,13 @@ class TestFirewall(NetworkCase):
         if m.image not in ["ubuntu-2004"]:
             m.execute("if virsh net-list --name | grep -q default; then until firewall-cmd --get-active-zones | grep -q libvirt; do sleep 1; done; fi")
 
+        # Temporary remove the test Dbus port 8765, this workaround can be
+        # removed once the pixel tests are merged and the image updated to not
+        # open port 8765
+        m.execute("firewall-cmd --permanent --remove-port=8765/tcp")
+        m.execute("firewall-cmd --remove-port=8765/tcp")
```
martinpitt, line 71: Does this work if the port is already not opened? i.e. is it idempotent?

I'll trigger a rhel and debian test to make sure, and also trigger the mobile test run.

martinpitt (approved):
Thanks! Looks good if this goes green.

test/verify/check-networkmanager-firewall
jelly: --permanent does not seem to close the port when it's open. so we have to remove it explicitly

martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-16699-20211208-111950-fc22b983-debian-testing/log.html#301) might be a flake, retrying for comparison.

martinpitt (approved):

PR #16700, opened 2021-12-08
c-client: Show recent hosts

marusak:
## Client: Show previously used hosts

Cockpit client now shows list of hosts that were used before so you can log in with one click. It is also possible to remove unwanted hosts from this list.

Screenshot from 2021-12-10 08-09-06


martinpitt:
Would be great to extend containers/flatpak/test/test-browser to actually test this functionality. The new quality will be that the flatpak gets run more than once -- so we need to enforce some ordering of the tests so that the "pick recent host" and "delete recent host" ones run after the initial login. This is not hard to do with naming them `test_1_foo` and `test_2_bar` etc..

I am happy to work on the tests if you want me to.

martinpitt:
In general, I don't actually see a reason to do this for client only -- "standard" web server setups for bastion host would equally benefit from that. But I do understand that we may want to introduces this in a cautious way?

Either way, please add a "normal" integration test to make sure that recent hosts don't appear there; or (if we decide to show it everywhere) that it works there as well.

marusak:
> In general, I don't actually see a reason to do this for client only -- "standard" web server setups for bastion host would equally benefit from that. But I do understand that we may want to introduces this in a cautious way?

The issue that I have with this is that we require name and password for bastion hosts. Now when you click on a host, it will immediately log you in.  We could make this work also for bastion hosts - but when username and password would be empty, we would just fill in the "Connect to" field with the selected value?

martinpitt:
Good point, I forgot about the merging of username and host. So indeed, let's start with only showing this for -client.

garrett:
I had some issues with building flatpaks again, then had issues with the network when rebuilding my toolbox to make things work. It's finally resolved, but I need to go AFK for a little bit. I'll continue after lunch.

Meanwhile: How do I make the hosts show up? I built from the branch, but I don't see the hosts UI.

garrett:
Aha. I was in an old branch called "flatpak". Rebuilding now... but I still need to run a few errands and have lunch. :wink:

marusak:
When you visit the login page for the first time (or if you remove all hosts) then you won't see this new panel. Any successful login then will add it into list. 
If you want to fake more hosts, you can just open browser console and execute `localStorage.setItem("cockpit-client-sessions", JSON.stringify(["localhost", "127.0.0.2", "mmarusak@tnuro"]))`

marusak:
Oh and yes, sometimes old .css files are kept, then `ctrl+shift+R` helps :)

garrett:
My big problem is I was building a (now) very old version of the flatpak app. :rofl: I thought it looked off, as the header bar was the slim one... that was the giveaway, and then I realized both branches were called "flatpak"... and I was on the old one. Whoops!

But these tips are good to know, definitely. Thanks! :+1:

pkg/static/login.js
```diff
@@ -638,8 +638,10 @@
         if (form == "login")
             id("login-button").addEventListener("click", call_login);
 
-        if (environment.is_cockpit_client)
+        if (environment.is_cockpit_client) {
             render_recent_hosts();
+            document.getElementsByTagName('body')[0].classList.add("cockpit-client");
```
marusak, line 643: cannot one just do `document.body`?

garrett: :facepalm: Right. Good point.

marusak:
@garrett thank you for the improvements! I've cleaned up commits and added one more to remember cookies. I also updated screenshot and will write release note. Now I'll take a look at tests.

marusak:
hmm, `flatpak-test` in CI is not happy about something, worked locally

garrett:
I can no longer build or install the flatpak. I'm not sure why. I switched to main and it builds and installs fine... but no longer on this branch.

garrett:
Should this have `--user`?

```diff
-flatpak install --user --assumeyes "${FLATPAK_ID}".flatpak
+flatpak-builder \
+    "$@" \
+    --disable-rofiles-fuse \
+    --disable-cache \
+    --force-clean \
+    --install flatpak-build-dir "${FLATPAK_ID}".yml
```

Adding `--user` in this block fixes it for me.

marusak:
`error: Environment variable must be given in the form VARIABLE=VALUE, not j`

The error message is cut off for some reason - and also I have no idea where it comes from. @martinpitt maybe you have some idea about this?

containers/flatpak/test/test-browser
```diff
@@ -11,8 +11,23 @@ TESTLIB = os.path.join(TEST_DIR, "test-lib.js")
 FLATPAK_ID = "org.cockpit_project.CockpitClient"
 JS_TIMEOUT = 15
 
+STORAGE = os.path.join(os.path.expanduser("~"), ".var/app/org.cockpit_project.CockpitClient/data/webkitgtk/localstorage/")
```
martinpitt, line 14: Simplify to

    os.path.expanduser("~/.var/app/org.cockpit_project.CockpitClient/data/webkitgtk/localstorage/")

containers/flatpak/test/test-browser
```diff
@@ -11,8 +11,23 @@ TESTLIB = os.path.join(TEST_DIR, "test-lib.js")
...
+TMP_STORAGE = "/tmp/localstorage"
+
 
 class BrowserTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        # Hide localstorage to start browser from clean state
+        subprocess.check_call(["mkdir", "-p", TMP_STORAGE])
+        subprocess.check_call(["cp", "-a", STORAGE, TMP_STORAGE])
+        subprocess.check_call(["rm", "-rf", STORAGE])
```
martinpitt, line 24: Please avoid a hardcoded /tmp/localstorage path. That is likely to exist already from previous aborted runs (and then would wreak havoc), it changes time stamps and other attributes. Please just rename the whole existing `~/.var/app/org.cockpit_project.CockpitClient` directory to `*.cockpittest`, clean up the new one, and rename it back.

This will also fail if the directory does not exist at all -- I'm surprised that it doesn't crash and burn on the `cp` in the GitHub workflow test.

But this is still a nasty hack -- @allisonkarlitskaya , isn't there some way that we could use some `$XDG_*` variable here?

martinpitt (changes requested):

containers/flatpak/test/test-browser
```diff
@@ -83,5 +98,18 @@ class BrowserTest(unittest.TestCase):
         self.assertEqual(self.run_js("test-browser-login.js"), "page-load")
         self.assertEqual(self.run_js("test-browser-overview.js"), "PASS")
 
+    def testRecentHosts(self):
+        self.assertIn(self.run_js("test-browser-wait-init.js"), ["found", "page-load"])
+
+        # try to log in by clicking on `localhost`
+        self.assertEqual(self.run_js("test-browser-login-recent.js"), "page-load")
```
martinpitt, line 105: You forgot to git add the new test files here, so I can't reproduce this locally. This is probably also why the test fails on GitHub:

> FileNotFoundError: [Errno 2] No such file or directory: '/var/home/martin/upstream/cockpit/containers/flatpak/test/test-browser-login-recent.js'

martinpitt:
@marusak : `git status` is your friend!

   FileNotFoundError: [Errno 2] No such file or directory: '/var/home/martin/upstream/cockpit/containers/flatpak/test/test-browser-logout.js'

marusak:
> git status is your friend!

It is :) sorry - should be fine now, I just noticed

containers/flatpak/test/test-browser
martinpitt: I started an ubuntu-2004 standard bots VM and ran the steps in .github/workflows/flatpak-test.yml, but only invoked the browser test directly (not test-ssh):
```
ERROR: setUpClass (__main__.BrowserTest)
  File "containers/flatpak/test/test-browser", line 23, in setUpClass
    subprocess.check_call(["cp", "-a", STORAGE, TMP_STORAGE])
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['cp', '-a', '/home/admin/.var/app/org.cockpit_project.CockpitClient/data/webkitgtk/localstorage/', '/tmp/localstorage']' returned non-zero exit status 1.
```

So it's test-ssh which avoids the crash. Doing a `mkdir -p` works around it.

containers/flatpak/test/test-browser
```diff
@@ -83,5 +98,18 @@ class BrowserTest(unittest.TestCase):
...
+        self.assertEqual(self.run_js("test-browser-logout.js"), "PASS")
```
martinpitt, line 106: This is potentially racy -- it could be that the connection gets closed before it gets to `PASS`, so accepting "page-load" in addition here would make sense. But that's not what breaks the test.

martinpitt:
In my local run in an ubuntu-2004 machine, it fails differently:
```
======================================================================
FAIL: testRecentHosts (__main__.BrowserTest)
  File "containers/flatpak/test/test-browser", line 106, in testRecentHosts
    self.assertEqual(self.run_js("test-browser-logout.js"), "PASS")
AssertionError: 'Error: timed out waiting for #toggle-menu' != 'PASS'
- Error: timed out waiting for #toggle-menu
+ PASS
```

I added a (local for now) commit to create screenshots after tests, and that shows me a normal cockpit session. I certainly see the toggle-menu button. Of course on GitHub it fails even earlier, but that might be related.

I also tried to merge the test cases, in case our tearDown/setUp is not properly isolating multiple flatpak runs:
```diff
--- containers/flatpak/test/test-browser
+++ containers/flatpak/test/test-browser
@@ -100,13 +100,13 @@ class BrowserTest(unittest.TestCase):
 
         self.assertEqual(self.run_js("test-browser-login.js"), "page-load")
         self.assertEqual(self.run_js("test-browser-overview.js"), "PASS")
+        self.assertIn(self.run_js("test-browser-logout.js"), ["PASS", "page-load"])
 
-    def testRecentHosts(self):
         self.assertIn(self.run_js("test-browser-wait-init.js"), ["found", "page-load"])
 
         # try to log in by clicking on `localhost`
         self.assertEqual(self.run_js("test-browser-login-recent.js"), "page-load")
-        self.assertEqual(self.run_js("test-browser-logout.js"), "PASS")
+        self.assertIn(self.run_js("test-browser-logout.js"), ["PASS", "page-load"])
 
         # try to log in to unreachable host
         self.assertEqual(self.run_js("test-browser-login-invalid.js"), "page-load")

```
(this also fixes the page-load race condition on logout)

This works locally and fails the same way in the ubuntu VM, but interestingly, when I comment out the wait-init.js, it fails the same way locally. But that's only a red herring, adding more wait-inits does not fix the VM run.

martinpitt:
I added a test commit on top which works both locally and in my Ubuntu VM. Let's see how it fares on CI.

martinpitt:
The "error: Environment variable must be given in the form VARIABLE=VALUE, not j" error comes from [flatpak](https://github.com/flatpak/flatpak/blob/d4b681ec6cfb869efdaada4c10a991fc8e17f079/common/flatpak-context.c#L1146).

The `"not %.*s", (int) len, p)` syntax is not familiar to me, but it's printf(3) syntax and means "read the precision length from the next int argument", so this is correct. So, somewhere in the flatpak env there is some trailing "j" somewhere which cannot be interpreted. At this point it needs interactive debugging on a github instance.

containers/flatpak/test/test-browser
allisonkarlitskaya: If you have the PID then you could mine the environment out of the running process using something like:

```
xargs -0n1 < /proc/172619/environ | grep XDG_DATA_HOME
```

containers/flatpak/test/test-browser
martinpitt: done in a fixup

containers/flatpak/test/test-browser
martinpitt: done in a fixup

containers/flatpak/test/test-browser
martinpitt: done in fixup

martinpitt:
@marusak: From all the intermediate noise, there are three remaining fixups: https://github.com/cockpit-project/cockpit/pull/16700/commits/e831c8401c77560f002ff9772d33ba68e48d06c2 https://github.com/cockpit-project/cockpit/pull/16700/commits/414b1552a061b78d118076607b2544be11772da0 https://github.com/cockpit-project/cockpit/pull/16700/commits/ea79aca2409b17dba9db79b95e51eb93ba443e2d . This will make it easier for you to cross-review.

I'll now squash them together and re-enable all the tests, so that we can get some good CI.

martinpitt:
Succeeded. I pushed a clean commit series now.

martinpitt (dismissed):
This is fine for me now. I'd still appreciate another review from @allisonkarlitskaya , and @mmarusak please cross-check my fixups. Thanks!

marusak:
Thank you @martinpitt ! Looks good to me :)

garrett (dismissed):
This works, looks good enough, and and remembers the logged in language now.

(It doesn't remember the logged in state, so admin escalation is needed each time. But this can be fixed in a follow-up. This PR fixes so much already and the admin stuff is out of scope for this PR.)

src/client/cockpit-client
```diff
@@ -276,7 +277,8 @@ class ExternalCockpitWs:
 class InternalCockpitWs:
     def start(self):
         self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        self.socket.bind(('127.0.90.90', 0))
+        # Use static port to convince WebKit it is the same domain
```
allisonkarlitskaya, line 284: A bit more explicit like "static port is needed for WebKit local storage to persist across runs" or so?

pkg/static/login.js
```diff
@@ -563,6 +563,39 @@
         }
     }
 
+    function render_recent_hosts() {
+        const hosts = JSON.parse(localStorage.getItem("cockpit-client-sessions") || "[]");
```
allisonkarlitskaya, line 567: Is it possible that this variable gets corrupted and then `JSON.parse()` will throw, and prevent the login page from showing properly?

If not, then probably it makes sense to move the default `[]` outside of being parsed.

containers/flatpak/test/test-browser
```diff
@@ -57,6 +58,8 @@ class BrowserTest(unittest.TestCase):
         self.assertIn("run-js", actions)
 
     def tearDown(self):
+        if shutil.which("import"):
+            subprocess.call(["import", "-window", "root", f"test-browser-{ self.id().split('.')[-1] }.png"])
```
allisonkarlitskaya, line 62: Is this going to break if we try to run the tests under Wayland?  I do that sometimes locally.

allisonkarlitskaya:
This is really really cool work, thanks :heart: :blush:

containers/flatpak/test/test-browser
martinpitt: So do I :grin:  It's certainly got going to succeed, but as it's (deliberately) not `check_call()` it will not fail the test.

marusak: It seems to work under wayland

pkg/static/login.js
marusak: I wrapped it into try/catch block to prevent breaking of the whole login page if this variable would be corrupted. Good catch!

src/client/cockpit-client
marusak: Done

pkg/static/login.js
```diff
@@ -922,6 +972,15 @@
         let wanted = window.sessionStorage.getItem('login-wanted');
         const machine = id("server-field").value;
 
+        /* When using cockpit client remember all the addresses being used */
+        if (machine && environment.is_cockpit_client) {
+            const hosts = JSON.parse(localStorage.getItem("cockpit-client-sessions") || "[]");
```
allisonkarlitskaya, line 977: ...and this one too :/

Maybe there ought to be a function?

allisonkarlitskaya (changes requested):

pkg/static/login.js
marusak: oops, fixed

allisonkarlitskaya (approved):
thanks!

and thanks for doing this

PR #16701, opened 2021-12-08
test: Better handle selinux alerts produced during login

mvollmer:
The next rhel-8-6 image will produce selinux alerts for rhsm.service
when logging into Cockpit.  We want them to dismiss them reliably
before starting the test.

In order for them to show up, auditd needs to be running already
during login, and we need to give the machinery copious time to pick
up the alert after the journal message has been written.

### DIFF
--- a/test/verify/check-selinux
+++ b/test/verify/check-selinux
@@ -21,6 +21,7 @@ import parent
 from testlib import *
 from machine_core.constants import TEST_OS_DEFAULT
 
+import time
 
 FIX_AUDITD = """
 set -e
@@ -69,6 +70,9 @@ class TestSelinux(MachineCase):
         b = self.browser
         m = self.machine
 
+        # fix audit events first
+        self.machine.execute(script=FIX_AUDITD)
+
         # Do some local modifications
         m.execute("semanage fcontext --add -t samba_share_t /var/tmp/")
         m.execute("semanage boolean --modify --on zebra_write_config")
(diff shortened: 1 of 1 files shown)


mvollmer:
This is unsatisfactory... maybe we should just fix the selinux policy issue in the rhel-8-6 image.

test/verify/check-selinux
```diff
@@ -77,6 +81,12 @@ class TestSelinux(MachineCase):
 
         self.login_and_go("/selinux/setroubleshoot")
 
+        # Starting Cockpit might produce some alerts (e.g. from
+        # starting rhsm.service), so let's give them some time to
+        # appear before cleaning it all out.
+        b.wait_js_cond("ph_in_text('body', 'No SELinux alerts.') || ph_is_present('tbody .pf-c-table__toggle')")
+        time.sleep(30)
```
martinpitt, line 88: The sleep here is annoying indeed, but I don't have a better idea either. We could at least restrict it to rhel-8-6/centos-8-stream with a "HACK:" comment prefix, so that at some point we'll clean it up again.

But why the extra wait_js_cond state? The "while" loop below already does that, and it shouldn't matter much if the test sleeps first and then does the UI query, or the other way around?

mvollmer: The idea was that we wait for the page to finish connecting to setroubleshoot first, since that seems to take a long time as well. But since thesubsequent wait needs to be really long anyway, *shrug*

mvollmer:
I am investigating the remaining subman failures first before going forward with this, but it already looks like rhsm is broken in more ways than one and the selinux denial has nothing to do with the other failures.

mvollmer:
This doesn't seem to be super necessary anymore... this has not flaked since we merged the rhel-8-6 image.

martinpitt:
I saw it again today: https://logs.cockpit-project.org/logs/pull-16734-20211216-063357-9b673908-rhel-8-6/log.html#126-2

So it's still necessary.

martinpitt (approved):
Let's land this, thanks!

PR #16702, opened 2021-12-08
flatpak: clean up build scripts

allisonkarlitskaya:
This is a massive reworking of our container/flatpak/ scripts.  All of
the scripts take different arguments than they used to.  Some
highlights:

 - no script takes the app id as a commandline argument anymore, but the
   FLATPAK_ID environment variable can still be used to override it.
   This will currently produce broken builds, however, due to the
   incorrectly named desktop and D-Bus service files (which is no change
   from the existing situation).

 - the prepare script now takes more reasonable arguments: the tarball
   to build from (or `make dist` if not specified) and the url where we
   can download the tarball from (or the current directory, if not
   specified).

 - update-flathub now produces output that's actually suitable for
   uploading to flathub, including knowledge of where our tarballs
   are released.

 - the install script now assumes that you want to install from running
   `make dist` on the working tree (but lets you override via the
   TARBALL environment variable).  Its commandline arguments are now
   passed directly to flatpak-builder.  It also no longer assumes to
   install as the user, so if you want that you need to give --user.  It
   also no longer uses the build script to install into a repository,
   build a bundle, and install the bundle, favouring the --install
   argument of flatpak-builder as more direct.

 - consequently, the build script has been dropped.

 - documentation added to all of the scripts

The flatpak-test workflow has been updated accordingly.


allisonkarlitskaya:
About the failure:

TIL:

```
. script args go here
```
is a bashism.

in POSIX shell you need to write that like
```
set -- args go here
. script
```

martinpitt (approved):
Thank you!

martinpitt:
[centos-stream-9 TF failure](https://copr-be.cloud.fedoraproject.org/results/packit/cockpit-project-cockpit-16702/centos-stream-9-x86_64/03012973-cockpit/builder-live.log.gz) is independent (gcc is uninstallable, ugh!), we can ignore that.

PR #16703, opened 2021-12-08
session: Validate client certificates against sssd's trusted CA

martinpitt:
sssd 2.6.1 introduces a new FindByValidCertificate() method which
validates the given certificate against the configured CA. Use this API
if available, and fall back to FindByCertificate() to keep compatibility
with older distributions.

This enables OCSP/CRL checks, and enables certmap rules in a safe way.
Update TestLogin.testClientCertAuthentication() to adjust the scary
comments, and make sure that certificates without a valid CA are
rejected if the new API is available.

Similarly, adjust check-system-realms to tell sssd about a valid CA, and
test valid/invalid cases.

Update the client certificate docs to show how to set up the CA in sssd,
and adjust the warning: certmap rules are now actually okay with sssd
2.6.1.

https://bugzilla.redhat.com/show_bug.cgi?id=1992149
CVE-2021-3698

-----

TODO:
 - [x] update FreeIPA cert login docs to use `ipa cert-request`
 - [x] add integration test for matching cert with invalid sig (works on older OSes, fails with sssd 2.6.1)
 - [x] update `TestLogin.testClientCertAuthentication` for CA validation, certmap rules are safe with this new API
 - [x] update documentation: Preview (without CSS) at https://piware.de/tmp/cert-authentication.html
 - [x] add release note: in particular, point out sssd CA installation requirement on updates

## Certificate login validation: Action required on updates

Earlier Cockpit/sssd versions did not check trust or revocation status of a presented client certificate, and thus certificate/smart card login was secure and supported only when matching the entire binary certificate against the Identity Management's database. With sssd 2.6.1 and Cockpit 260, the certificate signature and revocation status is now validated against the CA [configured in sssd](https://www.mankier.com/5/sssd.conf#Services_Sections-PAM_configuration_options), and any non-trusted certificate is rejected. This includes the case when the local sssd has *no* configured CA, which may break certificate logins after updating cockpit and sssd.

Thus if you use certificate login, you need to set up the trusted CA in sssd. Please see the [certificate authentication documentation](https://cockpit-project.org/guide/latest/cert-authentication.html#certauth-server-cockpitconf) for details.

This issue was assigned [CVE-2021-3698](https://access.redhat.com/security/cve/CVE-2021-3698).

### DIFF
--- a/doc/guide/cert-authentication.xml
+++ b/doc/guide/cert-authentication.xml
@@ -35,12 +35,11 @@ openssl req -new -newkey rsa:2048 -days 365 \
 
     <para>Now get this certificate request signed by the Certificate Authority of your Identity
     Management domain, to get a PEM certificate. Browsers and smart cart utilities accept PKCS#12 format
-    for importing/transfer, so convert the certificate/key pair, and protect it with a transfer
-    password:</para>
+    for importing/transfer, so convert the certificate/key pair; it will ask for and protect it
+    with a transfer password:</para>
 
 <programlisting>
-openssl pkcs12 -export -password pass:somepassword \
-    -in alice.pem -inkey alice.key -out alice.p12
+openssl pkcs12 -export -in alice.pem -inkey alice.key -out alice.p12
 </programlisting>
 
     <para>Don't forget to clean up the key file when you do not need it any more:</para>
--- a/doc/guide/cert-authentication.xml
+++ b/doc/guide/cert-authentication.xml
@@ -29,7 +29,7 @@
       This command generates a simple key and certificate request for the "alice" user:</para>
 
 <programlisting>
-openssl req -new -newkey rsa:2048 -days 365 -nodes \
+openssl req -new -newkey rsa:2048 -days 365 \
     -keyout alice.key -out alice.csr -subj "/CN=alice"
 </programlisting>
 
@@ -41,6 +41,12 @@ openssl req -new -newkey rsa:2048 -days 365 -nodes \
 <programlisting>
 openssl pkcs12 -export -password pass:somepassword \
     -in alice.pem -inkey alice.key -out alice.p12
+</programlisting>
+
+    <para>Don't forget to clean up the key file when you do not need it any more:</para>
+
+<programlisting>
+shred -u alice.key
 </programlisting>
 
   <para>You can now import <code>alice.p12</code> directly into your browser,
--- a/doc/guide/cert-authentication.xml
+++ b/doc/guide/cert-authentication.xml
@@ -144,6 +144,16 @@ maprule = LDAP:(userCertificate={cert!base64})
   <section id="certauth-server-cockpitconf">
     <title>Cockpit web server configuration</title>
 
+    <para>Set the trusted Certificate Authority of your user certificates in <command>sssd</command>,
+    either by copying the CA PEM file to <code>/etc/sssd/pki/sssd_auth_ca_db.pem</code> or setting the
+    <ulink url="https://www.mankier.com/5/sssd.conf#Services_Sections-PAM_configuration_options">
+    <command>pam_cert_db_path</command></ulink> configuration option to the path of the CA.
+    If you use FreeIPA and its CA:</para>
+
+<programlisting>
+cp /etc/ipa/ca.crt /etc/sssd/pki/sssd_auth_ca_db.pem
+</programlisting>
+
     <para>Certificate authentication needs to be enabled in
       <ulink url="./cockpit.conf.5.html">cockpit.conf</ulink> explicitly:</para>
 
@@ -155,12 +165,17 @@ ClientCertAuthentication = yes
   <warning>
     <para>Any client can generate certificates with arbitrary information, thus
       the client certificates must be checked for validity. Cockpit currently
-      accepts any client certificate and relies on sssd to verify their validity.
-      This is secure only when the entire certificate gets matched, i.e. when
-      importing the complete certificates into Identity Management as shown above.
-      <emphasis>Do not use this with local sssd <code>certmap</code> rules</emphasis>
-      which only match on Subject/Issuer properties!
-  </para>
+      accepts any client certificate and relies on sssd to verify their validity.</para>
+
+    <para>Check if your <command>sssd</command> version is at least 2.6.1, as earlier
+      versions did not validate the user certificate against the configured CA with
+      Cockpit logins. With earlier versions, this is secure only when the
+      entire certificate gets matched, i.e. when importing the complete
+      certificates into Identity Management as shown above. Also, certificate
+      revocations are <emphasis>not checked with earlier sssd versions</emphasis>.</para>
+
+    <para><emphasis>Do not use this with local sssd <code>certmap</code> rules</emphasis>
+      which only match on Subject/Issuer properties, if you run sssd earlier than 2.6.1!</para>
  </warning>
 
   <para>When enabling this mode,
--- a/doc/guide/cert-authentication.xml
+++ b/doc/guide/cert-authentication.xml
@@ -26,19 +26,19 @@
     <para>Generating the certificates for users is usually done with a certificate management system like
       <ulink url="https://pagure.io/certmonger">certmonger</ulink> or
       <ulink url="https://www.freeipa.org/page/PKI">FreeIPA</ulink>, which are not documented here.
-      For testing purposes, these commands will generate a self-signed certificate/key for the "alice" user:</para>
+      This command generates a simple key and certificate request for the "alice" user:</para>
 
 <programlisting>
-# create self-signed certificate and key
-# some browsers insist on specifying key usage, so it needs a config file
-printf '[req]\ndistinguished_name=dn\nextensions=v3_req\n[dn]\n
-    [v3_req]\nkeyUsage=digitalSignature,keyEncipherment,keyAgreement\n' > /tmp/openssl.cnf
+openssl req -new -newkey rsa:2048 -days 365 -nodes \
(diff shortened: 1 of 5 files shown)


martinpitt:
Took me two hours to track down [this failure](https://logs.cockpit-project.org/logs/pull-16703-20211209-061647-c0bea28d-debian-testing/log.html#278) -- turns out it is a rather subtle bug in Debian's libpam-sss package, I reported it at https://bugs.debian.org/1001377. There is a simple workaround, though.

doc/guide/cert-authentication.xml
```diff
@@ -26,19 +26,19 @@
...
-    -out alice.pem -subj "/CN=alice" -config /tmp/openssl.cnf -extensions v3_req
+    <para>Now get this certificate request signed by the Certificate Authority of your Identity
+    Management domain, to get a PEM certificate. Browsers and smart cart utilities accept PKCS#12 format
+    for importing/transfer, so convert the certificate/key pair, and protect it with a transfer
+    password:</para>
 
-# browsers and smart cart utilities accept PKCS#12 format, convert it
-# this needs to set a transfer/import password
+<programlisting>
 openssl pkcs12 -export -password pass:somepassword \
```
allisonkarlitskaya, line 42: Can you please do something about this anti-pattern of using a password on the commandline?

doc/guide/cert-authentication.xml
```diff
@@ -26,19 +26,19 @@
...
       <ulink url="https://www.freeipa.org/page/PKI">FreeIPA</ulink>, which are not documented here.
-      For testing purposes, these commands will generate a self-signed certificate/key for the "alice" user:</para>
+      This command generates a simple key and certificate request for the "alice" user:</para>
 
 <programlisting>
-# create self-signed certificate and key
-# some browsers insist on specifying key usage, so it needs a config file
-printf '[req]\ndistinguished_name=dn\nextensions=v3_req\n[dn]\n
-    [v3_req]\nkeyUsage=digitalSignature,keyEncipherment,keyAgreement\n' > /tmp/openssl.cnf
+openssl req -new -newkey rsa:2048 -days 365 -nodes \
```
allisonkarlitskaya, line 32: `-nodes`?  Raw key on disk? :(

src/session/client-certificate.c
```diff
@@ -248,16 +248,35 @@ sssd_map_certificate (const char *certificate, char** username)
...
                           "/org/freedesktop/sssd/infopipe/Users",
                           "org.freedesktop.sssd.infopipe.Users",
-                          "FindByCertificate",
+                          "FindByValidCertificate",
                           &err,
                           &user_obj_msg,
                           "s",
                           certificate);
 
+  /* fallback for sssd < 2.6.1; this is only secure for matching the full certificate, not mapping rules */
```
allisonkarlitskaya, line 262: I don't like how we fall back to this automatically if we're on older sssd and meanwhile the documentation has been updated to say "this works now!" with only a version warning at the bottom, after the instructions, that probably nobody will get around to reading.

I feel like we should either find a way to match only complete known certificates on older sssd (ie: disable ones that are merely signed by a CA), or not update the docs until this is widely supported and secure.

Another option might be a config flag to enable falling back to this API, but of course, that would break existing users.

Let's chat about how we manage this.

allisonkarlitskaya (changes requested):
I'm not a big fan of the false perception that this gives that "this is secure now".

allisonkarlitskaya:
Another random idea: could we package cockpit in such a way that it is marked as broken by old sssd versions, and then drop the fallback?  Might force distros to update their sssds faster?

martinpitt:
> Another random idea: could we package cockpit in such a way that it is marked as broken by old sssd versions, and then drop the fallback?

Only at the price of immediately stopping backportability in Debian/Ubuntu, and Fedora 34. I think at least RHEL 8.6 will get the new sssd, but it's not there yet. So this could only be conditional. But also, why? It's not like the new version makes anything worse?

doc/guide/cert-authentication.xml
martinpitt: I added a new commit that drops the `-password`, it will just ask interactively. An admin can figure out how to use `-password file:...` if they need to do this noninteractively. Not pushed yet, let's fist discuss the other comments.

doc/guide/cert-authentication.xml
martinpitt: Nothing new, but I (locally) added a commit that adds a `shred -u alice.key` after the p12 conversion.

src/session/client-certificate.c
martinpitt: I updated the release note in this PR to drop the "certmap works now" paragraph. Indeed, let's only announce this once sssd 2.6.1 is available more broadly. The patches to the guide only became more strict -- of course I'm happy to adjust the wording in the warning, but it is still true that you can safely ignore the warning completely -- just as before, the guide *still* only documents how to set up complete certificate matching. For that, the only things that changed is (1) that you *must* set the CA in sssd now, and (2) it will now start to check CRL. Neither of this is "this works now", it just makes the whole thing more restrictive.

Due to that, I find that the automatic fallback is acceptable for now. I am not really worried about that at all, I am worried that using the validation API by default now is too much of a regression on upgrades. It's acceptable for Fedora and even RHEL 9, but not really for RHEL 8.6. I would like to refine that, but I need https://github.com/SSSD/sssd/issues/5911 fixed for that first -- then I can send a follow-up to fall back to the unvalidated API also on "sssd does not have a CA".

doc/guide/cert-authentication.xml
martinpitt: Pushed now.

doc/guide/cert-authentication.xml
martinpitt: I updated the commit to also drop `-nodes`. You don't need the .key for signing the CSR, and p12 conversion will happily ask about the key. This is now safe, but highly interactive -- but as this is not a course in automated key management, but just a demo how to integrate this with IdM, I think that's ok.

martinpitt:
I pushed the documentation updates. You still had some qualms about the updated warning text, so I kept that thread open. Happy to discuss further, of course.

test/verify/check-static-login
```diff
@@ -714,27 +710,47 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
...
                              'org.freedesktop.sssd.infopipe.Users FindByCertificate s -- '
                              '''"$(cat %s)" | sed 's/^o "//; s/"$//' ''' % alice_cert)
         self.assertEqual(m.execute('busctl get-property org.freedesktop.sssd.infopipe ' + user_obj.strip() +
                                    ' org.freedesktop.sssd.infopipe.Users.User name').strip(),
                          's "alice"')
 
+        # sssd 2.6.1 introduces FindByValidCertificate() method which validates the given certificate
+        # against the configured CA. Without this, certmap rules are completely unsafe!
+        api = m.execute("busctl introspect org.freedesktop.sssd.infopipe /org/freedesktop/sssd/infopipe/Users")
+        has_validate_api = 'FindByValidCertificate' in api
```
martinpitt, line 737: I did this instead of our usual static OS list to avoid breaking packit tests and image refreshes, and forcing lockstep updates between bots and cockpit. However, let's make sure that it does not regress: Let's assert that this is True for fedora-35 and debian-testing. Then we can add more OSes as they come in, and we'll find that place when grepping for "fedora-35" (as we usually do with new OS releases).

test/verify/check-static-login
```diff
@@ -740,6 +740,9 @@ matchrule = <SUBJECT>^DC=LAN,DC=COCKPIT,CN=alice$
         # it into the browser.
 
         def do_test(authopts, expected, not_expected=[], session_leader=None):
+            # clear sssd logs; this makes debugging individual failures easier, and also works around https://bugs.debian.org/1001377
+            m.execute("rm -f /var/log/sssd/*")
```
martinpitt, line 760: Let's only do this on debian-testing; the bug is distro specific for now, and I want to at least know when it starts to happen in Fedora as well.

martinpitt:
Unrelated side issue: This [rhel-9-0 check](https://logs.cockpit-project.org/logs/pull-16703-20211210-102326-3d1e0983-rhel-9-0/log.html#289-1) repeatedly crashes polkit in the 3rd test re-run. Let's investigate and report/naughty.

allisonkarlitskaya (dismissed):
I think this can go now, but I'd like to squabble over the release note a bit more.  In particular, the "with this version and sssd version..." text is a bit too easy to scan over.

test/verify/check-static-login
martinpitt: Done.

test/verify/check-static-login
martinpitt: Done.

martinpitt:
@allisonkarlitskaya : I addressed the two points of my self-review. I also adjusted the release note to say the specific cockpit version (before your review I wasn't sure if it would make it into 260).

allisonkarlitskaya (dismissed):

martinpitt:
The remaining issue is [this arch failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16703-20211210-102326-3d1e0983-arch/log.html#285-1). It's unrelated to the changes here, but it does hit me as the login tests all run three times due to being affected.

martinpitt:
I added another commit on top to fix the [arch failures](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16703-20211210-134029-6d4bf4c2-arch/log.html). @jelly, can I ask you to review that one, please? @allisonkarlitskaya already approved everything below that.

jelly (approved):

PR #16704, opened 2021-12-08
shell: Add pacman compatibility to "About web console"

jelly:
List the installed cockpit packages on Arch Linux.

### DIFF
--- a/pkg/shell/shell-modals.jsx
+++ b/pkg/shell/shell-modals.jsx
@@ -45,28 +45,15 @@ export class AboutCockpitModal extends React.Component {
 
     componentDidMount() {
         const packages = [];
-
-        cockpit.spawn(["bash", "-ec", "command -v zypper dnf apt | head -n1 | xargs --no-run-if-empty basename"], [], { err: "message" })
-                .then(output => {
-                    output = output.trim();
-                    let cmd = null;
-                    if (output === "dnf" || output === "zypper")
-                        cmd = "rpm -qa --qf '%{NAME} %{VERSION}\\n' | grep cockpit | sort";
-                    else if (output === "apt")
-                        cmd = "dpkg-query -f '${Package} ${Version}\n' --show | grep cockpit | sort";
-                    else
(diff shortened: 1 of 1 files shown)


pkg/shell/shell-modals.jsx
```diff
@@ -46,14 +46,16 @@ export class AboutCockpitModal extends React.Component {
     componentDidMount() {
         const packages = [];
 
-        cockpit.spawn(["bash", "-ec", "command -v zypper dnf apt | head -n1 | xargs --no-run-if-empty basename"], [], { err: "message" })
+        cockpit.spawn(["bash", "-ec", "command -v zypper dnf apt pacman | head -n1 | xargs --no-run-if-empty basename"], [], { err: "message" })
```
martinpitt, line 49: How about we completely ditch this `command` thing, and just run `(set +e; rpm -qa ... ; dpkg-query ..; pacman -Q ...) | grep cockpit | sort` ? This avoids the "zypper" vs. "yum" vs. "dnf" knowledge and has fewer moving parts.

With a lot of effort you can probably construct a system where you install cockpit-system.deb and cockpit-podman.rpm, but (1) it's actually correct to show both packages in the list, and (2) who actually does that :grin:

martinpitt: Or rather, drop the `-e` shell option and skip the `set +e` of course.

jelly: Makes sense, applied!

martinpitt (changes requested):
Ach, [our tests are so picky](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16704-20211208-145538-469c0c5a-ubuntu-stable/log.html#135), but they are right! Please add a `2>/dev/null` there to avoid the error messages. They are absolutely expected.

I tested this on rpm and deb, and it works fine otherwise. Thanks!

martinpitt (approved):
Thanks!

PR #16706, opened 2021-12-08
new workflow for updating flathub

allisonkarlitskaya:
I did this for the 259 release and it resulted in:

 - https://github.com/cockpit-project/cockpit/runs/4458153711?check_suite_focus=true
 - https://github.com/cockpit-project/org.cockpit_project.CockpitClient/commit/deb285ddee68ea0a2f42cf9cc9984dce98d0abec

I'm going to do a manual fix-up before I PR.

### DIFF
--- a/.github/deploy-keys.sh
+++ b/.github/deploy-keys.sh
@@ -1,21 +1,63 @@
-#!/bin/sh
-# (Re-)generate all deploy keys on https://github.com/cockpit-project/cockpit/settings/environments
+#!/bin/sh -x
 
-set -eux
+# (Re-)generate all deploy keys on
+#   https://github.com/cockpit-project/cockpit/settings/environments
+#
+# You might want this first:
+#   dnf install python3-pynacl
+#
+# Helpful command to check/update the lists below:
+#   git grep -l 'environment: cockpit-dist' (or whatever)
+
+set -eu
+cd "$(realpath -m "$0"/../..)"
 
 ORG=cockpit-project
 THIS=cockpit
 
-[ -e bots ] || make bots
+DRY_RUN="-v"
+if test -n "${1:-}"; then
+    if test "$1" = "--dry-run" -o "$1" = "-n"; then
+        DRY_RUN="-n"
+    else
+        echo "Unrecognised argument"
+        exit 1
+    fi
+fi
 
-# for workflows pushing to our own repo: npm-update.yml and weblate-sync-po.yml
-bots/github-upload-secrets --receiver "${ORG}/${THIS}" --env self --ssh-keygen DEPLOY_KEY --deploy-to "${ORG}/${THIS}"
+deploy_env() {
+    ENVIRONMENT="$1"
+    if [ -n "${2:-}" ]; then
+        DEPLOY_TO="${ORG}/$2"
+    else
+        DEPLOY_TO="${ORG}/${ENVIRONMENT}"
+    fi
 
-# for weblate-sync-pot.yml: push to https://github.com/cockpit-project/cockpit-weblate/settings/keys
-bots/github-upload-secrets --receiver "${ORG}/${THIS}" --env "${THIS}-weblate" --ssh-keygen DEPLOY_KEY --deploy-to "${ORG}/${THIS}-weblate"
+    bots/github-upload-secrets $DRY_RUN \
+        --receiver "${ORG}/${THIS}" \
+        --env "${ENVIRONMENT}" \
+        --ssh-keygen DEPLOY_KEY \
+        --deploy-to "${DEPLOY_TO}"
+}
 
-# for webpack-jumpstart.yml/prune-dist.yml: push to https://github.com/cockpit-project/cockpit-dist/settings/keys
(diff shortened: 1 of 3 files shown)


.github/workflows/update-flathub.yml
```diff
@@ -0,0 +1,43 @@
+name: update-flathub
+on:
+  workflow_dispatch:
```
martinpitt, line 3: This is fine for now of course. For the future, we'd move this as a job into release.yml and let that run after the `cockpituous` job?

.github/workflows/update-flathub.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          path: src
+          # we need tags for what comes next...
+          fetch-depth: 0
+
+      - name: Clone flathub repository
+        uses: actions/checkout@v2
+        with:
+          path: flathub
+          repository: cockpit-project/org.cockpit_project.CockpitClient
```
martinpitt, line 24: Conceptually, why does this clone our fork instead of the actual flathub repo? Can we assume that they are always in sync? Or do we need that hack because the default github token can't create PRs on https://github.com/flathub/org.cockpit_project.CockpitClient ?

.github/workflows/update-flathub.yml
```diff
@@ -0,0 +1,43 @@
...
+          wget "${URLBASE}/${VERSION}/${TARBALL}"
+          test -f "${TARBALL}"
+
+      - name: Update flathub repository
+        run: |
+          set -x
+          git config --global user.name "GitHub Workflow"
+          git config --global user.email "cockpituous@cockpit-project.org"
+          src/containers/flatpak/update-flathub "$(realpath *.tar.xz)" "$(realpath flathub)"
+          git -C flathub push origin HEAD
```
martinpitt, line 43: So this doesn't create a PR yet, it just updates our fork. Thus so far we'll let the workflow do that, and propose the PR manually? (that's fine of course, just want to make sure I understand it)

.github/deploy-keys.sh
```diff
@@ -61,3 +61,7 @@ deploy_env "${THIS}-dist"
 #   - npm-install.yml
 #   - release.yml
 deploy_env node-cache
+
+# https://github.com/cockpit-project/org.cockpit_project.CockpitClient
+#   - update-flathub.yml
+deploy_env flathub org.cockpit_project.CockpitClient
```
martinpitt, line 67: Can we have a deploy key for https://github.com/flathub/org.cockpit_project.CockpitClient ? I suppose that still won't help us with creating a PR (at least not with the default github token), but it would at least avoid having to keep this fork on cockpit-project -- we could just directly create a branch on flathub/org.cockpit_project.CockpitClient and then propose that.

martinpitt (approved):
Thanks! A few questions, but of course this is Fairly Harmless™ so far.

.github/workflows/update-flathub.yml
allisonkarlitskaya: Correct.  That's the plan.  But I'd like to do this manually for the next couple of releases to be honest.  My next step (for next release) would be to have the PR automatically opened from the workflow.

.github/workflows/update-flathub.yml
allisonkarlitskaya: It's not about the PRs: it's about the ability to push.  We need admin access to install deploy keys, and we don't have admin access on the repo in the flathub project.  So we push to our own fork (which we do have a deploy key for) and then propose the PR from there.

.github/workflows/update-flathub.yml
allisonkarlitskaya: Yes.  See above: that's the next thing I'd like to improve.

.github/deploy-keys.sh
allisonkarlitskaya: The short answer is "no".  But I can ask.

.github/workflows/update-flathub.yml
martinpitt: Right, that was my idea -- check out main from the upstream repo, then create a "version" branch for the release, and push that to our fork. So that we don't have to keep our fork's "main" branch up to date, and it'd be basically irrelevant (we could even remove it)

.github/deploy-keys.sh
martinpitt: Ack, thanks. It would alleviate the need for keeping a fork, but it's not a big deal. I'd just like to sort out the syncing, see above.

.github/workflows/update-flathub.yml
allisonkarlitskaya: Ah!  This is a good point.  I didn't think of that.  And indeed, I think your proposal is the right one.  I'd probably push a main branch with a README.md (since GitHub gets grumpy when there are no branches).

I'd prefer to make this change as a follow-up though, mostly due to how difficult it is to test the branch in its current state (due to this annoying github limitation about triggering workflows from branches when a workflow with the same name doesn't exist on the main branch).

martinpitt: Ack, makes sense -- deal!

martinpitt (approved):

PR #16707, opened 2021-12-08
flatpak: update desktop file in metainfo

allisonkarlitskaya:
We renamed the desktop file but forgot to update it in the metainfo.
This was flagged by @flathubbot when we attempted to push the update.

We won't have to re-do the release or anything.  I've just manually patched it into the bot-written commit for the time being.  This kind of thing is part of the reason why we keep the metainfo outside of the tarball.


martinpitt (approved):
Oops, thanks! I noticed the corresponding update in the flathub PR.

PR #16709, opened 2021-12-09
test: Call `tuned-adm recommend` with running tuned

martinpitt:
This avoids the "Cannot talk to TuneD daemon via DBus. Is TuneD daemon
running?" warning, and gives more accurate results. In particular,
tuned in CentOS 9 Stream started to recommend "atomic-guest" on the
testing farm.

----

[c9s packit](http://artifacts.dev.testing-farm.io/07c9b635-19fc-4802-90c7-1cf1985773bb/) spontaneously broke two days ago, let's see if that fixes it.

First version still has some debugging stuff.

### DIFF
--- a/test/run-verify-host-user.sh
+++ b/test/run-verify-host-user.sh
@@ -89,13 +89,15 @@ if [ -n "$test_basic" ]; then
         "
 fi
 
+TESTS=TestTuned.testBasic
+
 exclude_options=""
 for t in $EXCLUDES; do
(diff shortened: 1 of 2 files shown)


martinpitt:
Still [fails](https://github.com/cockpit-project/cockpit/pull/16709/checks?check_run_id=4467666606). This is a genuine regression, I filed it as https://bugzilla.redhat.com/show_bug.cgi?id=2030580 and will send a naughty now.

PR #16710, opened 2021-12-09
test: Add Arch Linux compatibility for `About Web Console`

jelly:
On Arch Linux the package cockpit also provides cockpit-bridge while on
other distributions it's seperate.


martinpitt (approved):
Thanks! Please trigger an arch test here, and also add it to our regular testmap -- let's not break it again! :wink:

PR #16714, opened 2021-12-10
test: Wait for animations to finish before pixel testing

mvollmer:


mvollmer:
Draft because I want to remove the explicit sleeps that we might have to deal with animations.

mvollmer:
> Draft because I want to remove the explicit sleeps that we might have to deal with animations.

Only one... :-)

mvollmer:
I have tested this with `TestMachinesNetworks.testNetworkState` and it works great there.

martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-16714-20211210-112518-79874aab-fedora-34/log.html#127-2) seems relevant? Exactly the same on [mobile](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16714-20211210-112518-79874aab-fedora-34-mobile/log.html#127-2)

Can you please rebase your branch (it's 25 commits behind main) to fix packit?

Cheers!

test/common/test-functions.js
```diff
@@ -309,3 +309,7 @@ function ph_element_clip(sel) {
     var f = currentFrameAbsolutePosition();
     return { x: r.x + f.x, y: r.y + f.y, width: r.width, height: r.height, scale: 1 };
 }
+
+function ph_wait_for_animations_to_finish(sel) {
+    return Promise.all(ph_find(sel).getAnimations({subtree:true}).map(a => a.finish()));
```
martinpitt, line 314: Cool about `subtree`! I was already afraid that we had to walk the tree ourselves.

[docs](https://developer.mozilla.org/en-US/docs/Web/API/element/getAnimations)

test/common/testlib.py
```diff
@@ -755,6 +755,7 @@ def assert_pixels(self, selector, key, ignore=[]):
         if not (Image and self.pixels_label and self.cdp and self.cdp.valid):
             return
 
+        self.call_js_func('ph_wait_for_animations_to_finish', selector)
         self.call_js_func('ph_scrollIntoViewIfNeeded', selector)
```
martinpitt, line 759: Should we swap these lines? Intuitively it could happen that scrolling to make an element visible will trigger further animations?

martinpitt (commented):
Awesome! Thanks for figuring this out! (almost..)

mvollmer:
> [This failure](https://logs.cockpit-project.org/logs/pull-16714-20211210-112518-79874aab-fedora-34/log.html#127-2) seems relevant? Exactly the same on [mobile](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16714-20211210-112518-79874aab-fedora-34-mobile/log.html#127-2)

Interesting.  I had seen that locally, but it was so weird that I dismissed it as a broken local CI setup...

mvollmer:
> > [This failure](https://logs.cockpit-project.org/logs/pull-16714-20211210-112518-79874aab-fedora-34/log.html#127-2) seems relevant? Exactly the same on [mobile](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16714-20211210-112518-79874aab-fedora-34-mobile/log.html#127-2)
> 
> Interesting. I had seen that locally, but it was so weird that I dismissed it as a broken local CI setup...

Oh, the tracer output has nothing to do with this.  The error is "Cannot finish Animation with an infinite target effect end." which must come from the "Softwatre updated" spinner.  Let's just count the animations and wait until there aren't any.

test/common/testlib.py
mvollmer: Ok, done.

martinpitt:
Ah, that spinner, I see -- and that does not count for `.length()` then? it's a bit hard to see how it doesn't, given that the old code mapped the (demonstrably non-empty) list to the .finish() call. My gut feeling is that we need another selector argument for "allow animation", or just disable the animation waiting completely for that particular call. But let's see what the tests say.

mvollmer:
> Ah, that spinner, I see -- and that does not count for `.length()` then?

No, I expect that it does, but it will eventually be removed from the DOM, and then it doesn't count anymore.

martinpitt:
New [arch test failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16714-20211210-153801-6bb6de60-arch/log.html#283-2), but totally unrelated to pixel tests. Retrying for comparison.

martinpitt:
Dang -- the [stratis test failed on mobile](https://logs.cockpit-project.org/logs/pull-16714-20211210-153801-6bb6de60-fedora-34-mobile/pixeldiff.html#TestStorageStratis-testBasic-copy-fsys-mobile) because the tooltip animation is not done any more. Seems like the previous .length fix actually made it worse? Or was the previous pass just "good"/bad luck?

We could go back to the .finish() approach and explicitly allow animations for the overview health card with the "forever" spinner?

mvollmer:
> Dang -- the [stratis test failed on mobile](https://logs.cockpit-project.org/logs/pull-16714-20211210-153801-6bb6de60-fedora-34-mobile/pixeldiff.html#TestStorageStratis-testBasic-copy-fsys-mobile) because the tooltip animation is not done any more. Seems like the previous .length fix actually made it worse? Or was the previous pass just "good"/bad luck?

I must have been luck.  The tooltip is part of the DOM, but not a child of `#dialog`, so we miss it completely.

mvollmer:
> The tooltip is part of the DOM, but not a child of `#dialog`, so we miss it completely.

Unfortunately, it comes into existence very asynchronously and I don't think we want to reliably predict whether any tooltips will show up (that depends on the length of the dialog title and the size of the screen), so there is not better way than to sleep a bit I am afraid.

mvollmer:
Ok, I have added a generic `blur()` to assert_pixels...

martinpitt (changes requested):
Thanks! I agree about the general blurring; we could even revert the `blur=False` hacks from commit 9b86016ce13b4c638b47599a940a2a25b7121305 after this.

This broke quite a few ref images due to the dialog close field or cancel button not being focused any more. They look right, but need to be updated.

martinpitt (approved):
Thanks! Pixel ref changes look nice now.

PR #16716, opened 2021-12-10
test: Update pixel test references for Fedora 35

mvollmer:


martinpitt (changes requested):
Most pixel test updates seem reasonable to me. Showing zram0 on the Storage page as available drive looks like a dangerous bug to me (well, it would hopefully fail as "busy"?), but it is what it is.

However, I do object to cementing the service failure in the Health card. Let's please just reset failed services (`systemctl reset-failed`) before taking the screenshot, and wait for the "service has failed" indicator to go away. Otherwise this will plague us forever.

Thanks!

mvollmer:
> However, I do object to cementing the service failure in the Health card. Let's please just reset failed services (`systemctl reset-failed`) before taking the screenshot, and wait for the "service has failed" indicator to go away. Otherwise this will plague us forever.

Yeah.. the whole list in the health card should be ignored, but that doesn't work when the current list is smaller than the reference.

Let's not have the failed service in the reference, and also ignore the whole card instead of just the list, ok?

martinpitt (approved):
Makes sense, thanks!

martinpitt:
The fedora 34 failures are expected -- once https://github.com/cockpit-project/bots/pull/2705 lands, they will start working again as they will stop doing pixel tests. Setting "blocked" for that, as these need to land in lockstep.

rhel-9-0 is a flake, retrying. I can't for the life of me reproduce that polkit crash locally.

mvollmer:
> rhel-9-0 is a flake, retrying.

I retried as well, still crashing, but I'd say it is unrelated.

allisonkarlitskaya (commented):
So I'm not a big fan of this change because of how it needs to land at the same time as cockpit-projects/bots#2705.  What we really ought to be doing is to have this check in `test/common/testlib.py`:

```
        if machine.image == testvm.TEST_OS_DEFAULT and os.environ.get("TEST_BROWSER", "chromium") == "chromium" and not self.is_devel_build():
            pixels_label = self.label()
```

be based on a file that lives either in the tree of cockpit or in the pixel tests reference repository itself.  Like `test/reference/cockpit-uses-fedora-35` or something like that.

That way we could land changes like this without syncing with bots.

PR #16719, opened 2021-12-13
Move unit-tests container to testing

allisonkarlitskaya:


allisonkarlitskaya:
So uh, no chromium in bookworm.  That's a bit of a drag...

martinpitt:
`TEST_BROWSER=firefox make check TESTS=dist/storaged/test-util.html` is unhappy locally. I'll have a look in a bit.

containers/unit-tests/setup.sh
```diff
@@ -37,7 +37,7 @@ dependencies="\
     libssh-dev \
     libsystemd-dev \
     npm \
-    nodejs:amd64 \
+    nodejs \
```
martinpitt, line 40: Curious that this works now -- not sure about the background, I guess node wasn't supported on 32 bit in at least Debian 11?

martinpitt (approved):
Thanks! This will make the weekly auto-refreshes quite a bit more interesting.

Issue #2323, opened 2015-05-21
systemd_manager.GetUnitFileState is not a function


stefwalter:
See this on RHEL on the main server page.

systemd version is: systemd-208-20.el7_1.3.x86_64

Code location is: cockpit-server.js: 232

stefwalter:
Same thing happens on the Services page.

This is due to insufficient privileges.

Likely we'll have to implement #1595 to fix this properly.

stefwalter:
This doesn't happen on a cleanly installed RHEL 7.1 system.

stefwalter:
Should be fixed in 0.58.

PR #16720, opened 2021-12-13
test: Skip TestUpdatesSubscriptions on fedora-35

mvollmer:
This was only enabled by accident and doesn't work very well, see
#2323.  Also, at least testNoSubOsRepo is broken on any "*/devel" test
context since packagekit is not preloaded there.

Thus, let's disable these tests before making fedora-35 the default.


mvollmer:
This should make `fedora-35/devel` pass.

martinpitt:
This [c9s packit failure](http://artifacts.dev.testing-farm.io/bf14db99-dbf3-4d53-8796-a4fe00b8dc91/) is known issue https://github.com/cockpit-project/bots/issues/2703 , somehow it doesn't match. Unrelated problem.

martinpitt (approved):

mvollmer:
Green enough!

PR #16721, opened 2021-12-13
cockpit-client: add support for :port

allisonkarlitskaya:
Connecting to an address like
```
root@127.0.0.2:2201
```
doesn't currently work because ssh(1) doesn't under the port number. Break it out to a -p parameter, if it's present.

### DIFF
--- a/src/client/cockpit-client-ssh
+++ b/src/client/cockpit-client-ssh
@@ -52,14 +52,20 @@ def cockpit_client_ssh():
     askpass = os.path.join(xdg_data_home, 'cockpit-client-askpass')
     shutil.copy(__file__, askpass, follow_symlinks=False)
 
-    host = sys.argv[1]
-
-    if host == 'localhost':
+    connect_to = sys.argv[1]
+    if connect_to == 'localhost':
         cmd = ['flatpak-spawn', '--host', 'cockpit-bridge']
     else:
+        host, _, port = connect_to.rpartition(':')
+        # catch cases like `host:123` but not cases like `[2001:abcd::1]`
+        if port.isdigit():
+            host_args = ['-p', port, host]
+        else:
+            host_args = [connect_to]
+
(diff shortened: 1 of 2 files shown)


src/client/cockpit-client-ssh
```diff
@@ -54,12 +54,19 @@ def cockpit_client_ssh():
 
     host = sys.argv[1]
 
+    # ssh(1) doesn't support ':port' syntax
+    if ':' in host:
+        host, _, port = host.rpartition(':')
```
allisonkarlitskaya, line 59: Uh ya, IPv6 `[2001::1]` style notation yada yada

martinpitt:
flatpak test timed out, that's new -- I haven't seen it fail/time out before. Retrying once.

src/client/cockpit-client-ssh
```diff
@@ -52,14 +52,20 @@ def cockpit_client_ssh():
     askpass = os.path.join(xdg_data_home, 'cockpit-client-askpass')
     shutil.copy(__file__, askpass, follow_symlinks=False)
 
-    host = sys.argv[1]
+    connect_to = sys.argv[0]
```
martinpitt, line 55: This is wrong -- argv[0] is the program name. argv[1] was right.

martinpitt (changes requested):

src/client/cockpit-client-ssh
allisonkarlitskaya: oops.

martinpitt:
Ugh, [another one](https://github.com/cockpit-project/cockpit/runs/4517621520?check_suite_focus=true#step:4:3391):
```
> log: not ok 154 - Expected 11 assertions, but 14 were run, expected: undefined, got: null, test: filter message in, module: , source: ge@http://localhost:35995/dist/base1/test-chan.js:2:33537
```
(failed three times)

martinpitt (approved):
Danke!

PR #16723, opened 2021-12-13
test: enable testTally on Arch Linux

jelly:
Arch Linux has faillock default enabled with an deny default of three,
so the configuration file has to be updated to enable four deny's.


martinpitt (approved):
Thanks! Retrying the flakes. Two were "stupid" timeouts, but [this one](https://logs.cockpit-project.org/logs/pull-16723-20211213-173633-6d5453f9-rhel-8-6/log.html#125) deserves a closer look.

**Updated**: Fixed the latter in https://github.com/cockpit-project/bots/pull/2719

PR #16724, opened 2021-12-13
test: Enable pam_ssh_add for Arch Linux tests

jelly:
Ideally the package would enable pam_ssh_add but as it does not, enable
it for the private keys test.


martinpitt (approved):
Thanks! Retrying the flakes.

PR #16726, opened 2021-12-14
unit-tests: remove amd64 arch from i386 container

allisonkarlitskaya:
This was originally used to install amd64 nodejs because there was no
binary version of node-sass for i386.  That stopped being a problem a
long time ago and we've since dropped installing nodejs:amd64, but we
forgot to remove adding the architecture for it.

 - [ ] Rebuild here: https://github.com/cockpit-project/cockpit/actions/runs/1576746588

### DIFF
--- a/pkg/base1/test-dbus-common.js
+++ b/pkg/base1/test-dbus-common.js
@@ -932,12 +932,12 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                             "com.redhat.Cockpit.DBusTests.Hidden": { Name: name }
                         }, "got data before signal");
                         dbus.removeEventListener("notify", onnotify);
-                        done();
                     });
                     dbus.call("/otree/frobber", "com.redhat.Cockpit.DBusTests.Frobber",
                               "EmitHidden", [name])
                             .always(function() {
                                 assert.equal(this.state(), "resolved", "method called");
+                                done();
                             });
                 });
     });
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -34,5 +34,5 @@ jobs:
           fi
 
       - name: Run unit-tests container
-        timeout-minutes: 20
+        timeout-minutes: 30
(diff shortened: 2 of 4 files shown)


martinpitt (dismissed):

martinpitt (dismissed):

pkg/base1/test-dbus-common.js
```diff
@@ -932,12 +932,12 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                             "com.redhat.Cockpit.DBusTests.Hidden": { Name: name }
                         }, "got data before signal");
                         dbus.removeEventListener("notify", onnotify);
-                        done();
                     });
                     dbus.call("/otree/frobber", "com.redhat.Cockpit.DBusTests.Frobber",
                               "EmitHidden", [name])
                             .always(function() {
                                 assert.equal(this.state(), "resolved", "method called");
+                                done();
```
martinpitt, line 940: This looks wrong. See src/ws/mock-service.c, the EmitHidden() call is what *causes* the DBusTests.Hidden signal to be sent, so it's correct to wait for it and only succeed the test in the signal handler above.

If this fails, it means that the signal wasn't sent, or not received.

martinpitt: from IRC: The EmitHidden function is implemented in a way that emits the signal before returning. The subscription would get a callback before the promise on the call came back

martinpitt (approved):
Danke!

pkg/base1/test-dbus-common.js
allisonkarlitskaya: Looking at `emit_hidden()` in `mock-service.c` it's pretty clear that

```
  g_dbus_connection_emit_signal (connection, NULL, path,
                                 "com.redhat.Cockpit.DBusTests.Hidden", "Yooohooo",
                                 g_variant_new ("()"), &error);
```

gets called before

```
  test_frobber_complete_emit_hidden (frobber, invocation);
```

which means we'd indeed expect to see the signal before the method returns.

PR #16727, opened 2021-12-14
networking: Fix typo perserve -> preserve

marusak:


martinpitt (approved):
Hah, thanks -- I had this locally, then a meeting got in between :grin:

garrett:
This makes me wonder how many spelling issues we have in the codebase. I ran [codespell](https://github.com/codespell-project/codespell) against Cockpit a couple years ago and sent PRs. It's probably time to do so again.

I wonder if we could/should have automated tests for this. (It wouldn't catch misspellings that are actual words and it wouldn't help for the translations.) We could have a wordlist for non-dictionary words and deal with it kind of like the "pixel" tests where the sidecars might need updates sometimes.

Or we could just run codespell from time to time and fix the issues it ~fines~ finds.

garrett:
OK, I ran Codespell on Cockpit again, and it did find a few things we would not want to bother with automating, I think. There's a new PR @ #16729. We probably want to do this on our various sub-projects too.

PR #16729, opened 2021-12-14
Fix typos and spelling issues found by codespell

garrett:
Running codespell again; inspired by #16727.

### DIFF
--- a/pkg/lib/long-running-process.js
+++ b/pkg/lib/long-running-process.js
@@ -82,7 +82,7 @@ export class LongRunningProcess {
     /*  Stop long-running process while it is RUNNING, or reset a FAILED one */
     terminate() {
         if (this.state !== ProcessState.RUNNING && this.state !== ProcessState.FAILED)
-            throw new Error(`cannot terminate LongRunningProcess in state ${this.sate}`);
+            throw new Error(`cannot terminate LongRunningProcess in state ${this.state}`);
 
         /* This sends a SIGTERM to the unit, causing it to go into "failed" state. This would not
          * happen with `systemd-run -p SuccessExitStatus=0`, but that does not yet work on older
--- a/pkg/networkmanager/dialogs-common.jsx
+++ b/pkg/networkmanager/dialogs-common.jsx
@@ -66,7 +66,7 @@ export const MacMenu = ({ idPrefix, model, mac, setMAC }) => {
         });
         optionsMapInit.push(
             { toString: () => _("Permanent"), value: "permanent" },
-            { toString: () => _("Perserve"), value: "preserve" },
+            { toString: () => _("Preserve"), value: "preserve" },
             { toString: () => _("Random"), value: "random" },
             { toString: () => _("Stable"), value: "stable" },
         );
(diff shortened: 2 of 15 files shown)


pkg/lib/long-running-process.js
```diff
@@ -82,7 +82,7 @@ export class LongRunningProcess {
     /*  Stop long-running process while it is RUNNING, or reset a FAILED one */
     terminate() {
         if (this.state !== ProcessState.RUNNING && this.state !== ProcessState.FAILED)
-            throw new Error(`cannot terminate LongRunningProcess in state ${this.sate}`);
+            throw new Error(`cannot terminate LongRunningProcess in state ${this.state}`);
```
garrett, line 85: This is an actual code change. I left it in as it looks like it fixes a bug. Please review this to make sure it is actually supposed to be state and not "sate".

I think it is supposed to be state, as it checks for `this.state`, so I'm pretty sure codespell spotted and fixed a bug here.

pkg/systemd/services/services.jsx
```diff
@@ -189,7 +189,7 @@ class ServicesPageBody extends React.Component {
             failed: _("Failed to start"),
             activating: _("Running"),
             deactivating: _("Not running"),
-            maintainance: _("Maintanance"),
+            maintenance: _("Maintenance"),
```
garrett, line 192: This is a change in code. It should be reviewed. Usually when codespell finds these changes, it changes the code where it is referenced. But I don't see it elsewhere in this file, so it might be used elsewhere. Or perhaps it was unused or is a bug elsewhere?

garrett:
Oh, I realized I didn't run codespell on our docs. Fixing now; I will force-push.

pkg/lib/long-running-process.js
martinpitt: That's correct, thanks!

pkg/networkmanager/dialogs-common.jsx
```diff
@@ -66,7 +66,7 @@ export const MacMenu = ({ idPrefix, model, mac, setMAC }) => {
         });
         optionsMapInit.push(
             { toString: () => _("Permanent"), value: "permanent" },
-            { toString: () => _("Perserve"), value: "preserve" },
+            { toString: () => _("Preserve"), value: "preserve" },
```
martinpitt, line 69: Collides with #16727, so this will need rebasing.

pkg/systemd/services/services.jsx
martinpitt: It's a very rare corner case of a systemd unit state, we don't use/test it anywhere. That table is just for completeness. So this is correct.

martinpitt (approved):
Thanks!

pkg/networkmanager/dialogs-common.jsx
marusak: Or we can just close #16727 if this has chance to land reasonably quickly as well

martinpitt: I landed #16727, and git+github are clever enough to rebase over it.

PR #16730, opened 2021-12-14
unit-tests: Enable -fanalyzer for gcc

martinpitt:
Commits b3e3b10fdae8 and 4193c1d96 fixed gcc -fanalyzer errors a while
ago. Now that we have a new enough gcc in the unit-tests container,
enable the analyzer permanently.


martinpitt:
packit fails due to missing rebasing, but I think we can safely ignore this here.

allisonkarlitskaya (approved):
Thanks :)

PR #16731, opened 2021-12-14
tests: return true (not undefined) from a filter

allisonkarlitskaya:
The filter function should always return a boolean, and in one case we
were missing a return statement.  That seems to have worked out okay on
Chromium, but sometimes Firefox doesn't like it, leading to fails like

> log: not ok 154 - Expected 11 assertions, but 14 were run...

Make it explicit.

I have no idea if this is actually the problem from https://github.com/cockpit-project/cockpit/runs/4517621520?check_suite_focus=true#step:4:3391 but this fix certainly won't make things worse...


martinpitt (approved):
Nice catch!

PR #16735, opened 2021-12-16
workflows: Add missing apt update to flatpak-test

martinpitt:
Some package versions in the GitHub workflow machine's old package index
got cleaned up and are not available any more.

----

Fixes [this failure](https://github.com/cockpit-project/cockpit/runs/4543993599?check_suite_focus=true)


allisonkarlitskaya (approved):
huh.  thanks.

of course, they'll rebuild the image soon enough and this problem will go away on its own... but then, of course, Debian will update some packages and it'll be back again...

PR #16736, opened 2021-12-16
session: Close shell's stdin in user_has_valid_login_shell()

martinpitt:
tlog in RHEL/CentOS 9 now tries to read from its stdin even with `-c`.
While this is arguably a bug, it does not happen during VT or SSH logins
nor with `runas`. Let's just be robust against that and set stdin to
/dev/null.

----

This fixes the recent [c9s packit regression](http://artifacts.dev.testing-farm.io/96eec411-b0b0-49de-ae33-9778e853e7ad/). See https://github.com/cockpit-project/cockpit/pull/16718#issuecomment-995601497 for debugging notes.

### DIFF
--- a/src/session/session.c
+++ b/src/session/session.c
@@ -838,10 +838,17 @@ user_has_valid_login_shell (const char **envp)
    * <pitti> https://xkcd.com/221/
    */
   const char *argv[] = { pwd->pw_shell, "-c", "exit 71;", NULL };
-  const int remap_fds[] = { -1, 2, -1 }; /* send stdout to stderr */
+
+  int devnull = open ("/dev/null", O_RDONLY);
+  if (devnull < 0)
+      err (EX, "couldn't open /dev/null");
+
+  const int remap_fds[] = { devnull, 2, -1 }; /* send stdout to stderr */
   int wstatus;
 
   wstatus = spawn_and_wait (argv, envp, remap_fds, 3, pwd->pw_uid, pwd->pw_gid);
+  debug ("user_has_valid_login_shell: exited with status %x", wstatus);
+  close (devnull);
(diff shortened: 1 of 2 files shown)


martinpitt:
FTR, in https://github.com/cockpit-project/bots/pull/2729 I'll add tlog to our RHEL images, then we can also test this on our own CI.

src/session/session.c
```diff
@@ -838,10 +838,13 @@ user_has_valid_login_shell (const char **envp)
    * <pitti> https://xkcd.com/221/
    */
   const char *argv[] = { pwd->pw_shell, "-c", "exit 71;", NULL };
-  const int remap_fds[] = { -1, 2, -1 }; /* send stdout to stderr */
+  int devnull = open ("/dev/null", O_RDONLY);
```
allisonkarlitskaya, line 841: hate to do this to you, particularly if the bots are down, but this deserves at least an assert and maybe a proper ` if(...) abort();` or `return false;`

allisonkarlitskaya (changes requested):

src/session/session.c
martinpitt: I agree that this is a bit sloppy -- but if it fails, devnull will be -1, and thus fall back to the current behaviour. If /dev/null does not exist, there's little that we can do. But of course I'm happy to log a warning.

allisonkarlitskaya: But the premise of this patch is that the current behaviour is wrong...

If we can't open `/dev/null` then likely we have some fairly serious issues...

martinpitt: I added an error check. I just err() out on failure, indeed without /dev/null a lot of things will go wrong anyway.

allisonkarlitskaya (approved):
Looks okay :)

shame there's no err variant that aborts.  that's what we actually want here...

PR #16737, opened 2021-12-16
tls: Fix potential out-of-bounds buffer access

martinpitt:
Fix areadlinkat() to not write to `buffer[-1]` if `readlinkat()` fails
with ENOENT.

Spotted by Coverity.

### DIFF
--- a/src/tls/test-cockpit-certificate-ensure.c
+++ b/src/tls/test-cockpit-certificate-ensure.c
@@ -182,7 +182,7 @@ areadlinkat (int         dirfd,
   ssize_t s = readlinkat (dirfd, filename, buffer, sizeof buffer);
   g_assert_cmpint (s, <, sizeof buffer);
   if (s == -1 && errno == ENOENT)
-    buffer[0] = '\0';
+    s = 0; /* results in returning empty string */
   else
     g_assert_no_errno (s);
   buffer[s] = '\0';


src/tls/test-cockpit-certificate-ensure.c
```diff
@@ -182,10 +182,14 @@ areadlinkat (int         dirfd,
...
-    buffer[0] = '\0';
+    {
+      buffer[0] = '\0';
+    }
   else
-    g_assert_no_errno (s);
-  buffer[s] = '\0';
+    {
+      g_assert_no_errno (s);
+      buffer[s] = '\0';
```
martinpitt, line 191: This should not make any practical difference, Coverity doesn't know/assume that the program will always terminate on the assert in this case. But let's just quiesce this.

allisonkarlitskaya: This is actually a completely legitimate bug.  In the `ENOENT` case we set `buffer[0] = '\0';` and then proceed to set `buffer[-1] = '\0';`  If this was about Coverity not trusting `g_assert_no_errno()` to exit then this patch wouldn't help anything (since it still does the assignment right after the assert).

I think my preferred fix here is to simply replace `buffer[0] = '\0';` with `s = 0;` or even `return g_strdup("");`.

allisonkarlitskaya (changes requested):

allisonkarlitskaya (approved):

PR #16738, opened 2021-12-16
build: build deterministic tarballs

allisonkarlitskaya:
The goal is for a given git commit ID to always give the exact same `.tar.xz` output.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,7 +39,7 @@ AC_CONFIG_MACRO_DIR([tools])
 AM_INIT_AUTOMAKE([1.11 foreign dist-xz no-dist-gzip subdir-objects])
 # we want tar-ustar to avoid introducing extra metadata (ctime, atime) which
 # only adds useless non-determinism to the result.  we also want to sort.
-am__tar='tar --format=ustar --sort=name -chf - "$$tardir"'
+am__tar='tar --format=ustar --sort=name --owner=root:0 --group=root:0 -chf - "$$tardir"'
 
 AC_USE_SYSTEM_EXTENSIONS
 AC_PROG_RANLIB
--- a/configure.ac
+++ b/configure.ac
@@ -37,7 +37,9 @@ AC_CONFIG_AUX_DIR([tools])
 AC_CONFIG_MACRO_DIR([tools])
 
 AM_INIT_AUTOMAKE([1.11 foreign dist-xz no-dist-gzip subdir-objects])
-am__tar='tools/tar-create "$$tardir"'
+# we want tar-ustar to avoid introducing extra metadata (ctime, atime) which
+# only adds useless non-determinism to the result.  we also want to sort.
+am__tar='tar --format=ustar --sort=name -chf - "$$tardir"'
 
 AC_USE_SYSTEM_EXTENSIONS
 AC_PROG_RANLIB
--- a/configure.ac
+++ b/configure.ac
@@ -36,7 +36,8 @@ AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_AUX_DIR([tools])
 AC_CONFIG_MACRO_DIR([tools])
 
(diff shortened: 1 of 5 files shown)


allisonkarlitskaya:
In local testing on Fedora 35:
 - xz-5.2.5-7.fc35.x86_64
 - tar-1.34-2.fc35.x86_64
 - autoconf-2.69-37.fc35.noarch
 - automake-1.16.2-5.fc35.noarch

building this commit (2741a8f25ff9d4f65516b1bb21cd629381179c9e) gets me a consistent sha256sum of

```
d2a467f7561c7d94951bf8155e5236be159f079d4658546b7d7444c400532fd6  cockpit-259.76.g2741a8f25.tar.xz
```

allisonkarlitskaya:
```
tar: value 1000830000 out of uid_t range 0..2097151
tar: Exiting with failure status due to previous errors
```

Interesting fail.  I guess that's a 32bit container uid and the ustar format doesn't like that.  It's a nice suggestion for more determinism, though: `--user` and `--group`.

configure.ac
```diff
@@ -39,7 +39,7 @@ AC_CONFIG_MACRO_DIR([tools])
 AM_INIT_AUTOMAKE([1.11 foreign dist-xz no-dist-gzip subdir-objects])
 # we want tar-ustar to avoid introducing extra metadata (ctime, atime) which
 # only adds useless non-determinism to the result.  we also want to sort.
-am__tar='tar --format=ustar --sort=name -chf - "$$tardir"'
+am__tar='tar --format=ustar --sort=name --owner=root:0 --group=root:0 -chf - "$$tardir"'
```
martinpitt, line 42: This is a fixup for "build: change timestamp mangling approach", so it'd be nicer to squash it instead of landing a broken commit.

martinpitt (commented):
Thanks! Just one comment, but not a blocker (no need to repush just for this). I'm generally not a fan of storing abandoned/broken stuff in main commits, as it breaks bisection, backporting, and is prone to copying wrong stuff to other places.

But the main question I have about this is "why?". I do get why we want reproducibility for binary packages (rpm/deb), but who cares about the tarball? The only long-lived ones are the ones which we release. Surely we ought to start signing them, but that's independent.

(Honest question, this is not a "no")

allisonkarlitskaya:
So at least one concrete usecase: running the flatpak install script after only changing the manifest or metadata is now much faster.

But just generally: it bothers me that `make dist` is non-deterministic.  And for such silly reasons.  Like, whoever decided that the UID and username (and group) along with nanosecond accuracy of the last access time of the file belongs in a release?  And the whole thing should be sorted according to whatever filesystem was in use where the script was run.  This is weird.

I really like the idea that a given commit ID, given the same version of the autotools, tar and xz, will always get the same hash.

It would be really cool if we knew what the hash of our released tarball was before we released it.  I even wanted to use that property a couple of weeks ago when I was working on the flatpak packaging but the upload to GitHub kept failing...  That's a one time thing, but like, on the other hand: why not?  Do we gain anything from the non-deterministism?

allisonkarlitskaya:
Also: I'm genuinely unsure about if I prefer the approach of using a custom tar command or if I prefer to have a timestamp-updating dist hook followed by a custom tar command anyway (since we still need `--sort`, `--owner`, etc.)  That's part of why I did the two separate commits: which one do you prefer?

martinpitt:
I slightly prefer the separate script, as it is easier to debug/develop separately. It also feels like "separation of concerns". I.e. the current PR as a whole is fine for me. I won't bother too much about the intermediate "broken" commits though. Thanks!

martinpitt (approved):

allisonkarlitskaya:
> I slightly prefer the separate script, as it is easier to debug/develop separately. It also feels like "separation of concerns". I.e. the current PR as a whole is fine for me. I won't bother too much about the intermediate "broken" commits though. Thanks!

For what it's worth, it's only broken if you have a ridiculously large UID.  Normal humans doing a bisect aren't gonna get tripped up here.

PR #16741, opened 2021-12-17
test: sudo with cert auth now works on RHEL 8.6/C8S

martinpitt:
The sudo fix made it into RHEL 8.6 nightly and CentOS 8 Stream. Adjust
the test special case accordingly.

----

This needs to land in lockstep with https://github.com/cockpit-project/bots/pull/2729


marusak (approved):
Danke!

PR #16744, opened 2021-12-17
containers: Stop using deprecated imp module

jelly:
The imp module will be dropped in Python 3.12, it's replacement is
importlib.


martinpitt (approved):
Thank you! This does the same thing as test/common/run-tests now.

PR #16748, opened 2021-12-21
test: change browser locale to C.UTF-8

allisonkarlitskaya:
We used to use en_US.UTF-8 as our "always available" English locale, but
it's not always available.  Use C.UTF-8 instead.

This fixes the many warnings from firefox that can be seen when running
unit-tests:

(firefox-esr:16949): Gtk-WARNING **: 14:46:06.423: Locale not supported by C library.
	Using the fallback 'C' locale.


martinpitt (approved):
Thanks!

allisonkarlitskaya:
@jelly said that the arch fail is a known flake.

PR #16750, opened 2021-12-21
a few lgtm fixups

allisonkarlitskaya:
...all of them for issues I introduced myself :facepalm:

### DIFF
--- a/src/client/cockpit-client-ssh
+++ b/src/client/cockpit-client-ssh
@@ -11,7 +11,6 @@ import uuid
 
 def send_control(**msg):
     logging.debug(msg)
-    fp = sys.stdout.buffer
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
From the flatpak test:

```
error: Environment variable must be given in the form VARIABLE=VALUE, not j
F
```

cool.

martinpitt (approved):
Thanks! I just wonder, if LGTM found these, why didn't that status fail? I don't think I've *ever* seen it fail..

allisonkarlitskaya:
> Thanks! I just wonder, if LGTM found these, why didn't that status fail? I don't think I've *ever* seen it fail..

I was thinking exactly the same...

PR #16751, opened 2021-12-21
Add a cockpit rsync webpack plugin for cockpit plugins

jelly:
Currently it is not possible to easily run `npm run watch` and upload
artifacts using rsync as `./tools/webpack-make -r c` does. A webpack
plugin has some limitations no custom cli arguments can be passed,
either use env variables or --env.foo=1. Apart from that the plugin is
fairly simply but requires the `webpack.config.js` file to provide a
valid `packagename` for example `cockpit-machines` for the machines
plugin to the instantiated plugin.

### DIFF
--- a/pkg/lib/cockpit-rsync-plugin.js
+++ b/pkg/lib/cockpit-rsync-plugin.js
@@ -1,0 +1,28 @@
+const child_process = require("child_process");
+
+module.exports = class {
+    constructor(options) {
+        if (!options)
+            options = {};
+        this.dest = options.dest || "";
+        this.source = options.source || "dist/";
+    }
+
+    apply(compiler) {
+        compiler.hooks.afterEmit.tapAsync('WebpackHookPlugin', (compilation, callback) => {
+            if (process.env.RSYNC) {
+                const proc = child_process.spawn("rsync", ["--recursive", "--info=PROGRESS2", "--delete",
+                    this.source, process.env.RSYNC + ":/usr/share/cockpit/" + this.dest], { stdio: "inherit" });
+                proc.on('close', (code) => {
+                    if (code !== 0) {
+                        process.exit(1);
+                    } else {
+                        callback();
+                    }
+                });
+            } else {
+                callback();
+            }
+        });
+    }
+};
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -10,6 +10,7 @@ const path = require("path");
 const argparse = require("argparse");
 const fs = require("fs");
 const child_process = require("child_process");
+const CockpitRsyncPlugin = require("../pkg/lib/cockpit-rsync-plugin");
 
 // argv0 is node
 const webpack_watch = process.argv[1].includes('webpack-watch');
@@ -38,6 +39,11 @@ const srcdir = (process.env.SRCDIR || ".").replace(/\/$/, '');
 const cwd = process.cwd();
 const config_path = path.resolve(cwd, args.config);
 const config = require(config_path);
+
+if (args.rsync) {
+  process.env["RSYNC"] = args.rsync;
+  config.plugins.push(new CockpitRsyncPlugin({source: path.dirname(makefile)}));
+}
 
 const compiler = webpack(config);
 
@@ -75,9 +81,6 @@ function process_result(err, stats) {
             process.exit(1);
         return;
     }
-
-    if (args.rsync)
-        rsync();
 
     generateDeps(makefile, stats);
 }
@@ -184,12 +187,4 @@ function generateDeps(makefile, stats) {
     lines.push(prefix + ": " + stampfile);
 
     fs.writeFileSync(makefile, lines.join("\n") + "\n");
-}
-
-function rsync() {
-    const ret = child_process.spawnSync("rsync", ["--recursive", "--info=PROGRESS2", "--delete",
-                                        path.dirname(makefile), args.rsync + ":/usr/share/cockpit/"],
-                                        { stdio: "inherit" });
-    if (ret.status !== 0)
-        process.exit(1);
 }


jelly:
For `cockpit-machines` this would require:

```
diff --git a/webpack.config.js b/webpack.config.js
index 9c80a91..a72b7df 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const path = require("path");

 const copy = require("copy-webpack-plugin");
@@ -7,6 +8,7 @@ const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CompressionPlugin = require("compression-webpack-plugin");
 const ESLintPlugin = require('eslint-webpack-plugin');
 const CockpitPoPlugin = require("./src/lib/cockpit-po-plugin");
+const CockpitRsyncPlugin = require("./src/lib/cockpit-rsync-plugin");

 const webpack = require("webpack");

@@ -22,14 +24,19 @@ const copy_files = [
     "./src/manifest.json",
 ];

+// Obtain package name from package.json
+const data = fs.readFileSync('./package.json', { encoding:'utf8', flag:'r' });
+const packageJson = JSON.parse(data);

 const plugins = [
     new copy({ patterns: copy_files }),
-    new extract({filename: "[name].css"}),
+    new extract({ filename: "[name].css" }),
     new ESLintPlugin({
         extensions: ["js", "jsx"],
         failOnWarning: true,
     }),
     new CockpitPoPlugin(),
+    new CockpitRsyncPlugin({ packagename: packageJson.name }),
 ];
```

As changes, an alternative is to provide `tools/webpack-make` to other projects but then we need to work out the custom cockpit Makefile dependency logic.

martinpitt:
Nice, I like that idea! That annoyed me several times as well already. I suggest to change tools/webpack-watch to use this new plugin, i.e. drop its rsync() function and replace it with setting the env var. Also, can you please fix the various missing commas and typos in the commit message?

pkg/lib/cockpit-rsync-plugin.js
```diff
@@ -0,0 +1,31 @@
...
+module.exports = class {
+    constructor(options) {
+        if (!options)
+            options = {};
+        this.dest = options.dest;
+        if (options.source) {
+            this.source = options.source;
+        } else {
+            this.source = "dist/";
+        }
```
martinpitt, line 12: Usual JS syntax for defaults: `this.source = options.source || "dist/";`

pkg/lib/cockpit-rsync-plugin.js
```diff
@@ -0,0 +1,31 @@
+const child_process = require("child_process");
+
+module.exports = class {
+    constructor(options) {
+        if (!options)
+            options = {};
+        this.dest = options.dest;
```
martinpitt, line 7: If this is not given, it looks like the code would just happily rsync into root's home directory. There should be a check that it's not empty.

Also, it would be kind of nice to not *have* to specify it -- it's always `/usr/share/cockpit/<manifestname>`, but indeed at this point that is hard to figure out. So let's keep it explicit for now.

pkg/lib/cockpit-rsync-plugin.js
```diff
@@ -0,0 +1,31 @@
...
+            this.source = options.source;
+        } else {
+            this.source = "dist/";
+        }
+    }
+
+    apply(compiler) {
+        compiler.hooks.afterEmit.tapAsync('WebpackHookPlugin', (compilation, callback) => {
+            if (process.env.RSYNC) {
+                rsync(this.source, this.dest, process.env.RSYNC);
```
martinpitt, line 18: rsync() is blocking. Is that right for tapAsync()? It feels to me it should rather use `child_process.spawn()` and hook up its `close` event to calling `callback`? I'm not very familiar with the webpack plugin API, though.

nodejs also offeers [promisify()](https://nodejs.org/api/util.html#utilpromisifyoriginal) to use it together with `tapPromise()`, which might be more convenient.

pkg/lib/cockpit-rsync-plugin.js
```diff
@@ -0,0 +1,31 @@
...
+            }
+            callback();
+        });
+    }
+};
+
+function rsync(source, dest, hostname) {
+    const ret = child_process.spawnSync("rsync", ["--recursive", "--info=PROGRESS2", "--delete",
+        source, hostname + ":" + dest],
+                                        { stdio: "inherit" });
```
martinpitt, line 28: ugly indentation

martinpitt (changes requested):
Nice, thanks! Looking forward to using this! Some more questions/nitpicks. Please also fix the commit message a bit -- "build" → "built", the duplicated "rsync", and the line-broken verbatim (backticks).

Please feel free to use no-test and only run f-35, that's enough for this purpose -- this functionality isn't covered by CI. We are currently a bit low on CI power.

pkg/lib/cockpit-rsync-plugin.js
jelly: Hmm we could always specify "/usr/share/cockpit" as base path in the plugin, as everything should live under it correct?

Then we can go back to what I originally did with `packagename` and append that, if not set it should work for cockpit, what do you think?

pkg/lib/cockpit-rsync-plugin.js
jelly: :+1:

pkg/lib/cockpit-rsync-plugin.js
martinpitt: Not sure what you originally did, but in cockpit itself we have `section`, and in c-podman and friends I suppose we can just hardcode the name.

jelly: I mean we can change the code to `this.dest = "/usr/share/cockpit/" + dest`. That could in theory avoid hardcoding it in cockpit-machines as we can parse the package.json:
```
cat package.json | jq .name
"machines"
```

martinpitt: That's a nice idea -- and parsing JSON from JS is simple. Then we'd only need to override it for cockpit itself, but it would DTRT for starter-kit and friends.

martinpitt (approved):
Thanks! This works well enough, even better than before -- the rsync is now part of the webpack, so the "build succeeded" message now appears after rsync is done.

We can do the automagic dest handling as a follow-up if we want. Let's land this, and you can go ahead and implant it into starter-kit. Cheers!

PR #16754, opened 2021-12-22
Make configuring a custom shutdown/reboot datetime possible again and apply some UX improvements.

jelly:


martinpitt:
Thanks for the fixes! Looks good in general, but one test fails everywhere.

jelly:
> Thanks for the fixes! Looks good in general, but one test fails everywhere.

Yeah, it tested the blur functionality which now changed.

martinpitt (approved):
Dankje wel!

PR #16756, opened 2021-12-24
test: Enable skipped timesyncd tests

jelly:
The testTime test works without skipping, for the testTimeServers test
enable the by default disabled ntp service in timesyncd so the test is
able to set specific time servers

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -553,12 +553,16 @@ class TestHistoryMetrics(MachineCase):
         self.assertEqual(m.execute("systemctl is-enabled pmlogger").strip(), "enabled")
 
     @nondestructive
-    @skipImage("TODO: pm proxy alert doesn't show on Arch Linux", "arch")
     @skipImage("no PCP support", "fedora-coreos")
     @todoPybridge()
     def testPmProxySettings(self):
         b = self.browser
         m = self.machine
+
+        # Arch Linux has no active zone by default which the firewalld port alert test requires.
+        if m.image == "arch":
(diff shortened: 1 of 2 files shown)


jelly:
Ok, further debugging shows that `TestSystemInfo-testTimeServers` does not require any modifications but that `timedate.cal('SetNTP', [val, true])` either fails or times out, the console logs:

```
re {problem: null, name: 'org.freedesktop.DBus.Error.UnknownInterface', message: "Unknown interface 'org.freedesktop.systemd1.Service'.", toString: ƒ}message: "Unknown interface 'org.freedesktop.systemd1.Service'."name: "org.freedesktop.DBus.Error.UnknownInterface"problem: nulltoString: ƒ ()[[Prototype]]: Object```

martinpitt:
CLI reproducer:

 - `timedatectl set-ntp on` is fast, `systemctl status systemd-timesyncd` shows that it starts quickly.
 -  `timedatectl set-ntp off` → timesyncd stuck in "stopping" for a long time

This can also be done more directly with `systemctl start systemd-timesyncd` and `systemctl stop systemd-timesyncd` - the latter hangs.

jelly:
Let's also retest it when systemd 250 has landed in the image refresh https://github.com/cockpit-project/bots/pull/2796

KKoukiou:
Just rebased and let the arch tests rerun to check the current state.

jelly:
As testPmProxySettings works, let's get that merged in a separate PR.

jelly:
Done in https://github.com/cockpit-project/cockpit/pull/18071

jelly:
This needs more work, closing.

Issue #16764, opened 2021-12-27
Disconnected after login


hamdyaea:
Cockpit version: cockpit-251.1-1.el8.x86_64
OS: Rocky Linux 8.5

I connect to my cockpit with the right pasword and user then I get disconnected after a few seconds.

Relevant parts of the system log are also useful:
`journalctl -f`

```
Dec 27 21:02:44 Daylight systemd[1]: Starting Socket for Cockpit Web Service http-redirect instance.
Dec 27 21:02:44 Daylight systemd[1]: Starting Socket for Cockpit Web Service https instance factory.
Dec 27 21:02:44 Daylight systemd[1]: Starting Socket for Cockpit Web Service http instance.
Dec 27 21:02:44 Daylight systemd[1]: Listening on Socket for Cockpit Web Service http-redirect instance.
Dec 27 21:02:44 Daylight systemd[1]: Listening on Socket for Cockpit Web Service https instance factory.
Dec 27 21:02:44 Daylight systemd[1]: Listening on Socket for Cockpit Web Service http instance.
Dec 27 21:02:44 Daylight systemd[1]: Starting Cockpit Web Service...
Dec 27 21:02:45 Daylight systemd[1]: Started Cockpit Web Service.
Dec 27 21:02:45 Daylight systemd[1]: Started Cockpit Web Service https instance factory (PID 49513/UID 994).
Dec 27 21:02:45 Daylight systemd[1]: Starting Socket for Cockpit Web Service https instance e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.
Dec 27 21:02:45 Daylight systemd[1]: Listening on Socket for Cockpit Web Service https instance e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.
Dec 27 21:02:45 Daylight systemd[1]: Started Cockpit Web Service https instance e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.
Dec 27 21:02:45 Daylight systemd[1]: cockpit-wsinstance-https-factory@12-49513-994.service: Succeeded.
Dec 27 21:02:45 Daylight cockpit-ws[49517]: json_object_get_string_member: assertion 'node != NULL' failed
Dec 27 21:03:01 Daylight systemd[1]: Started Session 3217 of user root.
Dec 27 21:03:24 Daylight systemd[1]: session-3206.scope: Succeeded.
[...]

PR #16765, opened 2021-12-28
ws: Avoid regular json_object_get_string_member assertion, catch unexpected messages

martinpitt:
Whenever the login page was opened without a client certificate (i.e. in
the vast majority of cases), ws would log an error

    json_object_get_string_member: assertion 'node != NULL' failed

due to the sloppy way of checking for the existence of a client
certificate. Do this properly, and fail hard if tls sends a wrong data
type.

---

Spotted in #16764

### DIFF
--- a/src/ws/cockpitauth.c
+++ b/src/ws/cockpitauth.c
@@ -1502,9 +1502,9 @@ cockpit_auth_login_async (CockpitAuth *self,
 
   /* If the client sends a TLS certificate to cockpit-tls, treat this as a
    * definitive login type, and don't just silently fall back to other types */
-  const gchar *client_certificate;
+  const gchar *client_certificate = NULL;
   JsonObject *metadata = get_connection_metadata (connection);
-  if (metadata && (client_certificate = json_object_get_string_member (metadata, "client-certificate")))
+  if (metadata && cockpit_json_get_string (metadata, "client-certificate", NULL, &client_certificate) && client_certificate)
     {
       g_debug ("TLS connection has peer certificate, using tls-cert auth type");
       type = g_strdup ("tls-cert");
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1338,6 +1338,7 @@ class MachineCase(unittest.TestCase):
 
     def allow_restart_journal_messages(self):
         self.allow_journal_messages(".*Connection reset by peer.*",
+                                    "connection unexpectedly closed by peer",
                                     ".*Broken pipe.*",
                                     "g_dbus_connection_real_closed: Remote peer vanished with error: Underlying GIOStream returned 0 bytes on an async read \\(g-io-error-quark, 0\\). Exiting.",
                                     "connection unexpectedly closed by peer",
@@ -1380,11 +1381,17 @@ class MachineCase(unittest.TestCase):
             "SYSLOG_IDENTIFIER=cockpit-ws",
             "SYSLOG_IDENTIFIER=cockpit-bridge",
             "SYSLOG_IDENTIFIER=cockpit/ssh",
+            # also catch GLIB_DOMAIN=<library> which apply to cockpit-ws (but not to -bridge, too much random noise)
+            "_COMM=cockpit-ws",
             "GLIB_DOMAIN=cockpit-ws",
             "GLIB_DOMAIN=cockpit-bridge",
             "GLIB_DOMAIN=cockpit-ssh",
             "GLIB_DOMAIN=cockpit-pcp"
         ]
+
+        # older c-ws versions always log an assertion, fixed in PR #16765
+        if self.image == "rhel-8-5-distropkg":
+            self.allowed_messages.append("json_object_get_string_member: assertion 'node != NULL' failed")
 
         if not self.allow_core_dumps:
             matches += ["SYSLOG_IDENTIFIER=systemd-coredump"]
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -437,6 +437,7 @@ class TestConnection(MachineCase):
             ".*Peer failed to perform TLS handshake",
             ".*Peer sent fatal TLS alert:.*",
             ".*invalid base64 data in Basic header",
+            "Received unexpected TLS connection and no certificate was configured",
             ".*Error performing TLS handshake: No supported cipher suites have been found.",
             ".*Error performing TLS handshake: Could not negotiate a supported cipher suite.")
 
@@ -852,6 +853,7 @@ until pgrep -f cockpit-bridge.*--privileged; do sleep 1; done
         self.allow_journal_messages(".*No authentication agent found.*")
         self.allow_journal_messages(".*Peer failed to perform TLS handshake.*")
         self.allow_journal_messages(r".*cannot reauthorize identity\(s\): unix-user:.*")
+        self.allow_journal_messages("admin: Executing command .*COMMAND=/usr/bin/cockpit-bridge --privileged.*")
 
     @skipBrowser("Firefox needs proper cert and CA", "firefox")
     def testReverseProxy(self):
--- a/test/verify/check-kdump
+++ b/test/verify/check-kdump
@@ -28,6 +28,7 @@ class KdumpHelpers(MachineCase):
         self.machine.execute("grubby --args=crashkernel=256M --update-kernel=ALL")
         self.machine.execute("mkdir -p /var/crash")
         self.machine.spawn("sync && sync && sync && sleep 0.1 && reboot", "reboot")
+        self.allow_restart_journal_messages()
         self.machine.wait_reboot()
         self.machine.start_cockpit()
         self.browser.switch_to_top()
@@ -64,8 +65,6 @@ class TestKdump(KdumpHelpers):
         m = self.machine
 
         b.wait_timeout(120)
-        self.allow_restart_journal_messages()
-
         m.execute("systemctl enable kdump && systemctl start kdump || true")
 
         self.login_and_go("/kdump")
--- a/test/verify/check-storage-luks
+++ b/test/verify/check-storage-luks
@@ -530,6 +530,7 @@ class TestStorageLuks(StorageCase):
 
         self.setup_systemd_password_agent("vainu-reku-toma-rolle-kaja")
         m.spawn("sync && sync && sync && sleep 0.1 && reboot", "reboot")
+        self.allow_restart_journal_messages()
         m.wait_reboot()
         m.start_cockpit()
         b.relogin()
--- a/test/verify/check-storage-stratis
+++ b/test/verify/check-storage-stratis
@@ -361,6 +361,7 @@ class TestStorageStratis(StorageCase):
 
         # Reboot
         m.spawn("sync && sync && sync && sleep 0.1 && reboot", "reboot")
+        self.allow_restart_journal_messages()
         m.wait_reboot()
         m.start_cockpit()
         b.relogin()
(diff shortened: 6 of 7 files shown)


allisonkarlitskaya:
why didn't testing catch this as an unexpected log message?

martinpitt:
Most of our tests disable https, and then `metadata` itself is NULL. But we do have a few https tests, at least `TestConnection.testTls`. This *does* trigger the assertion, but indeed the test case does not complain. There's something wrong with our filter:

```
# journalctl -b | grep assertion.*node
Dec 31 11:30:26 fedora-35-127-0-0-2-2201 cockpit-ws[2075]: json_object_get_string_member: assertion 'node != NULL' failed
Dec 31 11:30:52 fedora-35-127-0-0-2-2201 cockpit-ws[2303]: json_object_get_string_member: assertion 'node != NULL' failed
Dec 31 11:30:52 fedora-35-127-0-0-2-2201 cockpit-ws[2303]: json_object_get_string_member: assertion 'node != NULL' failed

# journalctl -b SYSLOG_IDENTIFIER=cockpit-ws
-- Journal begins at Fri 2021-12-31 06:46:21 UTC, ends at Fri 2021-12-31 11:32:31 UTC. --
-- No entries --
```

However, `SYSLOG_IDENTIFIER=cockpit-tls` works fine. For some reason, the message does not have an identifier at all:

```Fri 2021-12-31 11:30:26.424466 UTC [s=4b73b116c09042f1a6237330bd941683;i=1563;b=a426bb71858d42ecad373a8909b37718;m=af8421;t=5d46f7f57168f;x=605ec48e054c1974]
    _MACHINE_ID=d8fe00a2bd274eb9b7d0de65bb78c262
    _HOSTNAME=fedora-35-127-0-0-2-2201
    _TRANSPORT=journal
    PRIORITY=4
    _BOOT_ID=a426bb71858d42ecad373a8909b37718
    _CAP_EFFECTIVE=0
    GLIB_OLD_LOG_API=1
    MESSAGE=json_object_get_string_member: assertion 'node != NULL' failed
    GLIB_DOMAIN=Json
    _PID=2075
    _UID=981
    _GID=977
    _COMM=cockpit-ws
    _EXE=/usr/libexec/cockpit-ws
    _CMDLINE=/usr/libexec/cockpit-ws --for-tls-proxy --port=0
    _SELINUX_CONTEXT=system_u:system_r:cockpit_ws_t:s0
    _SYSTEMD_CGROUP=/system.slice/system-cockpithttps.slice/cockpit-wsinstance-https@e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.service
    _SYSTEMD_UNIT=cockpit-wsinstance-https@e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855.service
    _SYSTEMD_SLICE=system-cockpithttps.slice
    _SYSTEMD_INVOCATION_ID=156e077f901f45a089a7c766ad58e1d3
    _SOURCE_REALTIME_TIMESTAMP=1640950226424466
```

This works:
```
# journalctl -b _COMM=cockpit-ws
-- Journal begins at Fri 2021-12-31 06:46:21 UTC, ends at Fri 2021-12-31 11:36:00 UTC. --
Dec 31 11:30:26 fedora-35-127-0-0-2-2201 cockpit-ws[2075]: json_object_get_string_member: assertion 'node != NULL' failed
Dec 31 11:30:52 fedora-35-127-0-0-2-2201 cockpit-ws[2303]: json_object_get_string_member: assertion 'node != NULL' failed
Dec 31 11:30:52 fedora-35-127-0-0-2-2201 cockpit-ws[2303]: json_object_get_string_member: assertion 'node != NULL' failed
```

martinpitt:
I added a commit which fixes the unexpected messages for cockpit-ws. Without the previous fix, it now fails:
```
  File "/var/home/martin/upstream/cockpit/test/common/testlib.py", line 1228, in tearDown
    self.check_journal_messages()
  File "/var/home/martin/upstream/cockpit/test/common/testlib.py", line 1432, in check_journal_messages
    raise Error(UNEXPECTED_MESSAGE + "journal messages:\n" + first)
json_object_get_string_member: assertion 'node != NULL' failed

# Result testTls (__main__.TestConnection) failed
# 1 TEST FAILED [20s on toolbox]
```

This might result in more failures, let the bots tell us about them (at least on non-RHEL)

martinpitt:
Of course -- the message comes from `GLIB_DOMAIN=Json`, so this silently broke with some recent glib which directly logs to journal. I updated the commit accordingly.

martinpitt:
There is now [quite some fallout](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20211231-114420-f7600b58-fedora-35/log.html), most of which on Fedora 35. But ` TestSuperuserOldDashboard.test` fails everywhere with an old bridge message.

Fixed.

martinpitt:
[One more](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20211231-150510-568e1af0-ubuntu-2004/log.html#15)

martinpitt:
Some more failures thanks to the "affected 3x retry":

 - [TestConnection.testTls](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20211231-162826-ca4cb10b-fedora-35/log.html#273-2) unexpected message says

       cockpit-ws[2099]: Received unexpected TLS connection and no certificate was configured

   which is rather straightforward fallout from this change. This reproduces locally perfectly well.

- [TestSuperuserOldDashboard.test](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20211231-162826-ca4cb10b-ubuntu-stable/log.html#288-1) has a polkit message 

       pkexec[1641]: pam_unix(polkit-1:session): session opened for user root by (uid=1001)

    The old cockpit indeed uses `pkexec`, but it's not quite clear why a polkit message appears in our filtered logs at all.

martinpitt:
Meh, that's a lot more than I bargained for.. [yet another](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20220101-115405-5fb6a441-fedora-35/log.html#40) unexpected message:

    cockpit-bridge[44408]: 1:2!3: Could not connect: Connection refused

This is rather difficult to pin down -- This is almost certainly not specific to that test case, and random things like that will be an eternal source of such flakes. I'm going to back out the `_COMM=cockpit-bridge` bit and only keep it for ws.

martinpitt:
[one more](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20220101-155008-4f1580a8-centos-8-stream/log.html#167), fixed

martinpitt:
Fixed [this one](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16765-20220102-072039-bc40537e-centos-8-stream/log.html#194) as well.

martinpitt:
I brought up aws task runners, and now get RHEL results. The [distropkg](https://logs.cockpit-project.org/logs/pull-16765-20220102-100138-1fcf18f8-rhel-8-5-distropkg/log.html) run has tons of failures due to this assertion, so @allisonkarlitskaya  here is your proof that the fix works and the tests catch it. I'll push another update to ignore this error message for r85-distropkg.

martinpitt:
[very close](https://logs.cockpit-project.org/logs/pull-16765-20220102-192134-f58a4da6-centos-8-stream/log.html#31)! one more flake fix.

martinpitt:
@allisonkarlitskaya : This is ready for review now, at last.

test/common/testlib.py
```diff
@@ -1380,12 +1381,18 @@ def check_journal_messages(self, machine=None):
...
+            "_COMM=cockpit-ws",
             "GLIB_DOMAIN=cockpit-ws",
             "GLIB_DOMAIN=cockpit-bridge",
             "GLIB_DOMAIN=cockpit-ssh",
             "GLIB_DOMAIN=cockpit-pcp"
         ]
 
+        # older c-ws versions always log an assertion, fixed in PR #16765
+        if self.image == "rhel-8-5-distropkg":
+            self.allowed_messages.append("json_object_get_string_member: assertion 'node != NULL' failed")
```
allisonkarlitskaya, line 1394: glad i get to carry this brown paper bag around for a while.

is this the kind of thing that it makes sense to distro-patch?

src/ws/cockpitauth.c
```diff
@@ -1502,11 +1502,20 @@ cockpit_auth_login_async (CockpitAuth *self,
 
   /* If the client sends a TLS certificate to cockpit-tls, treat this as a
    * definitive login type, and don't just silently fall back to other types */
-  const gchar *client_certificate;
   JsonObject *metadata = get_connection_metadata (connection);
-  if (metadata && (client_certificate = json_object_get_string_member (metadata, "client-certificate")))
+  if (metadata && json_object_has_member (metadata, "client-certificate"))
```
allisonkarlitskaya, line 1506: I feel like this code is trying to basically duplicate the cockpit json helpers here...

src/ws/cockpitauth.c
```diff
@@ -1502,11 +1502,20 @@ cockpit_auth_login_async (CockpitAuth *self,
...
     {
       g_debug ("TLS connection has peer certificate, using tls-cert auth type");
+
+      const gchar *client_certificate = json_object_get_string_member (metadata, "client-certificate");
```
allisonkarlitskaya, line 1510: I'm not sure it makes sense to do a GError style fail here.  I'd prefer to see a much smaller footprint: maybe just pretend it doesn't exist in that case?

Also: won't json_object_get_string_member throw an assert as well if we get the wrong type?

allisonkarlitskaya (changes requested):

test/common/testlib.py
martinpitt: Not for z-stream. Only security and important (data loss etc.) issues there, and for 8.6 it'll be fixed. We can drop this in a few months when we move to 8-6-distropkg.

src/ws/cockpitauth.c
martinpitt: Right, this could use cockpit_json_get_string().

src/ws/cockpitauth.c
martinpitt: Sure, we could silently ignore type errors, but it's really an unexpected error condition, so I thought it would make sense to fail very verbosely about it, as it's an internal programming error on our side -- possibly when we change the protocol in the future and forget to keep backwards compatibility.

If we get the wrong type, then json_object_get_string_member() [logs a critical](https://gitlab.gnome.org/GNOME/json-glib/-/blob/master/json-glib/json-object.c#L611) (via `g_return_val_if_fail()`), so it'll be similar to what we have in current main. I.e. that's exactly the case when we want to actually fail with "we screwed up the protocol"

allisonkarlitskaya: `cockpit-ws` and `cockpit-tls` are in the same package.  This won't become a compatibility problem in the future.

fwiw, this code is going to get dropped in #16715 anyway, so I don't feel the need to bikeshed too much about it.  If you'd prefer to merge it as-is, I'd be OK with that.  It's definitely an improvement over the current situation.

src/ws/cockpitauth.c
martinpitt: Done.

src/ws/cockpitauth.c
martinpitt: Ack, done. It's simpler that way, and if we are going to change it soon anyway, then indeed no need to be overly picky about it.

allisonkarlitskaya (approved):

PR #16770, opened 2022-01-03
test: Give content iframes a fixed size

mvollmer:
This isolates pixel tests for the pages from changes in the shell
layout.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -122,11 +122,12 @@ class Browser:
         self.pixels_label = pixels_label
         self.machine = machine
         path = os.path.dirname(__file__)
-        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
+        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace, fixed_content_size=self.pixels_label,
                            inject_helpers=[os.path.join(path, "test-functions.js"), os.path.join(path, "sizzle.js")])
         self.password = "foobar"
         self.timeout_factor = int(os.getenv("TEST_TIMEOUT_FACTOR", "1"))
         self.failed_pixel_tests = 0
+        self.body_clip = None
 
     def title(self):
         return self.cdp.eval('document.title')
@@ -594,6 +595,16 @@ class Browser:
             self.eval_js('window.localStorage.setItem("superuser:%s", "%s");' % (user, "any" if superuser else "none"))
         self.click('#login-button')
 
+    def adjust_window_for_fixed_content_size(self):
+        # When doing pixel tests, try to give the content iframe the
+        # expected fixed size so that pixel tests don't have to adapt
+        # to changes in the shell layout too much.  But don't bother
+        # when the browser is visible since we don't control its
+        # window size and it will likely be too small.
+        if self.pixels_label and not self.cdp.show_browser:
+            self.call_js_func("ph_adjust_content_size", self.cdp.content_width, self.cdp.content_height)
+            self.body_clip = self.call_js_func('ph_element_clip', 'body')
+
     def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False, password=None,
                      legacy_authorized=None):
         href = path
@@ -610,6 +621,7 @@ class Browser:
         self.expect_load()
         self._wait_present('#content')
         self.wait_visible('#content')
+        self.adjust_window_for_fixed_content_size()
         if path:
             self.enter_page(path.split("#")[0], host=host)
 
@@ -644,6 +656,7 @@ class Browser:
         self.expect_load()
         self._wait_present('#content')
         self.wait_visible('#content')
+        self.adjust_window_for_fixed_content_size()
         if path:
             if path.startswith("/@"):
                 host = path[2:].split("/")[0]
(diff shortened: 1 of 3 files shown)


mvollmer:
This should work, but the code is in the wrong place.  cdp.py shouldn't be the one that determines the window/content size.

mvollmer:
We should probably also restrict Browser.snapshot to snapshot `body` only.

mvollmer:
> This should work, but the code is in the wrong place.

Ah well, this is all over the place anyway, and cdp.py already takes care of other browser config stuff, like TEST_SHOW_BROWSER, and I am afraid things like self.browser.cdp.mobile might be API now.

martinpitt:
This still needs some work at least on [firefox](https://logs.cockpit-project.org/logs/pull-16770-20220104-084404-0fa86890-fedora-35-firefox/log.html#1-2):

    testlib.Error: eval_js(ph_adjust_content_size(1680,1130)): Can't adjust content iframe size.  Browser window is too small.

martinpitt:
Interestingly there is also a single [failure with chromium](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16770-20220104-084404-0fa86890-fedora-35-mobile/log.html#12-2) in the embed tests.

test/common/testlib.py
```diff
@@ -732,7 +745,10 @@ def snapshot(self, title, label=None):
             self.cdp.command("clearExceptions()")
 
             filename = "{0}-{1}.png".format(label or self.label, title)
-            ret = self.cdp.invoke("Page.captureScreenshot", no_trace=True)
+            if self.body_clip:
+                ret = self.cdp.invoke("Page.captureScreenshot", clip=self.body_clip, no_trace=True)
```
martinpitt, line 749: Why is that important? This just takes screenshots for human consumption, these are not involved in pixel tests.

test/common/testlib.py
```diff
@@ -610,6 +621,7 @@ def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=
         self.expect_load()
         self._wait_present('#content')
         self.wait_visible('#content')
+        self.adjust_window_for_fixed_content_size()
```
martinpitt, line 624: We have a few more tests which do custom logins. These would only need to call this if they do pixel tests, right?

martinpitt (commented):
Honestly I can't make up my mind yet whether this is an improvement overall. As shown by the test failures it makes the browser handling more brittle due to the rather sensitive and precise dimension calculations. I think the idea itself is valid and good, but the "price" of the implementation needs to be worth the very rare shell layout changes. But thanks for investigating this! Let's see how it looks like in a working version.

test/common/testlib.py
mvollmer: Yes, and only if they call assert_pixels on their `body` or something else that is affected by the iframe size.

test/common/testlib.py
mvollmer: It's not important. Without this, the snapshots would contain some white margin because the window is bigger than the top-level `body`.

mvollmer:
Firefox headless doesn't respect the `--window-size` command line option. One has to set MOZ_HEADLESS_WIDTH and `MOZ_HEADLESS_HEIGHT` in the environment instead: https://github.com/mozilla/geckodriver/issues/1354

This works for the width, but only up to a maximum of 1280.  (The default without setting MOZ_HEADLESS_WIDTH is 1366, which is curiously larger than the max.)

I had no luck setting the height. Firefox seems to tie width and height together somehow... E.g. requesting 1200x1220 gives you 1151x926, requesting 1280x800 gives you 1280x686.  It doesn't really make sense.

So let's not try to fix the iframes on Firefox.

(And note that because we don't currently set MOZ_HEADLESS_*, Firefox always runs at 1366x768, even for "mobile")

mvollmer:
> Honestly I can't make up my mind yet whether this is an improvement overall.

Yeah, it's a lot of hairy code that doesn't change anything about the status quo and we can only hope that it will do its job when it is needed, and we haven't accidentally broken it by then.

mvollmer:
> > Honestly I can't make up my mind yet whether this is an improvement overall.
> 
> Yeah, it's a lot of hairy code that doesn't change anything about the status quo and we can only hope that it will do its job when it is needed, and we haven't accidentally broken it by then.

On the other hand, there is a good chance it will just keep working and its good to at least record in the code what we want to achieve.

If you are bothered with how the logic is spread out over cdp.py and testlib.py, I can try to make it cleaner.

test/common/testlib.py
```diff
@@ -122,11 +122,12 @@ def __init__(self, address, label, machine, pixels_label=None, port=None):
         self.pixels_label = pixels_label
         self.machine = machine
         path = os.path.dirname(__file__)
-        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
+        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace, fixed_content_size=self.pixels_label,
```
martinpitt, line 125: So we indirectly rely on running pixel tests only on chromium? I think at least in cockpit-machines we run them on firefox also. But we can adjust that later on if it gets in the way.

martinpitt (approved):
Thanks!

test/common/testlib.py
mvollmer: Well, the code only runs the hairy adjustment stuff when pixel tests are active.  If we activate pixel tests on Firefox, we will run into the "can't request specific window size" bug of Firefox and we will get the "can't adjust iframe" warning, but then the tests continue.

So if pixel tests are green on Firefox right now, they should remain green.  (No adjustement to the iframe have been made, and the window is still the same size because Firefox ignores --window-size.)

PR #16771, opened 2022-01-03
urls-check: Ignore patternfly redirects

marusak:
PF maintainers often transfer issues between their repositories. This is
nothing we need to worry about.

Such reports can be seen in:
https://github.com/cockpit-project/cockpit/issues/16607
https://github.com/cockpit-project/cockpit/issues/16221
https://github.com/cockpit-project/cockpit/issues/14752


martinpitt (approved):
Dakujem!

PR #16776, opened 2022-01-04
shell: Fix locale provided to DatePicker

marusak:
When using locale that contains country code (such as pt_BR or zh_TW) and
clicking on calendar icon oopses the page.

[1] mentions that hyphens should be used as separators. With this fix it
behaves as expected, that `pt_BR` has starting day Sunday but `pt` Monday.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation


KKoukiou (commented):
You can use dateFormatLang from pkg/lib/timeformat.js

marusak:
> You can use dateFormatLang from pkg/lib/timeformat.js

good call! Changed it

martinpitt (approved):
Danke!

PR #16779, opened 2022-01-05
Consistently include config.h, enable largefile support

### LINKED ISSUE #16778: scanning for certificates fails with "Value too large for defined data type" on armhf-qemu guest on amd64 host
Today I ran into an interesting little bug:
I wanted to run cockpit inside a armhf-qemu guest on an amd64 host, but ran into the following error message:

`cockpit-tls: Could not locate server certificate: Error loading certificates from /etc/cockpit/ws-certs.d: Value too large for defined data type`

After investigating a bit, I tracked it down to the following line:
https://github.com/cockpit-project/cockpit/blob/ef9002e32e85e88cccd99f1e49996e7891018218/src/common/cockpitwebcertificate.c#L55

Would it be possible to replace the call with
`n = scandir64 (dir_name, &certs, filter_cert, alphasort64);`
and adjust https://github.com/cockpit-project/cockpit/blob/ef9002e32e85e88cccd99f1e49996e7891018218/src/common/cockpitwebcertificate.c#L38 to use `dirent64`.

Thanks a lot for providing the world with this excellent piece of software!

martinpitt:
Fixes #16778

### DIFF
--- a/src/common/Makefile-common.am
+++ b/src/common/Makefile-common.am
@@ -243,7 +243,7 @@ nodist_noinst_DATA += libpreload-temp-home.so
 EXTRA_DIST += src/common/preload-temp-home.c
 
 libpreload-temp-home.so: src/common/preload-temp-home.c
-	$(AM_V_CCLD) $(CC) $(CFLAGS) -fPIC -shared -o $@ $^
+	$(AM_V_CCLD) $(CC) $(AM_CPPFLAGS) $(CFLAGS) -fPIC -shared -o $@ $^
 
 CLEANFILES += \
 	$(test_locale_MO) \
--- a/src/common/preload-temp-home.c
+++ b/src/common/preload-temp-home.c
@@ -17,8 +17,8 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* for RTLD_NEXT */
-#define _GNU_SOURCE
+#include "config.h"
+
 #include <dlfcn.h>
 #include <pwd.h>
 #include <stdio.h>
--- a/src/tls/server.c
+++ b/src/tls/server.c
@@ -17,12 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-/* for secure_getenv () */
-#define _GNU_SOURCE
-
-#ifdef HAVE_CONFIG_H
-#include <config.h>
-#endif
+#include "config.h"
 
 #include "server.h"
 
--- a/tools/test-static-code
+++ b/tools/test-static-code
@@ -2,7 +2,7 @@
 # run static code checks like pyflakes and pycodestyle
 set -eu
 
-echo "1..8"
+echo "1..9"
 
 cd "${srcdir:-.}"
 fail=0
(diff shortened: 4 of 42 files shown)


src/common/preload-temp-home.c
```diff
@@ -18,7 +18,6 @@
  */
 
 /* for RTLD_NEXT */
```
martinpitt, line 20: This comment should be dropped along with the define. I'll wait for the current test run and re-push.

src/tls/server.c
```diff
@@ -18,7 +18,6 @@
  */
 
 /* for secure_getenv () */
```
martinpitt, line 20: same here

allisonkarlitskaya (changes requested):
We already have `_GNU_SOURCE` defined, by the "official" mechanism of including `AC_USE_SYSTEM_EXTENSIONS` in our `configure.ac`.  That makes sure `_GNU_SOURCE` is defined inside of any file which has `#include "config.h"` as its first line.  Unfortunately, we have an awful lot of files that *don't* do that.

Here's a suggestion for `test-static-code`:

```
# for every C file, if there is any line starting with '#', then the first one
# must be exactly: #include "config.h"
if find -name '*.c' -not -exec sed -n '/^#include "config.h"$/q0;/^\s*#/q1' {} ';' -print | grep ''; then
    echo 'not ok 9 config.h properly included from all files'
    fail=1
else
    echo 'ok 9 config.h properly included'
fi
```

martinpitt:
Thanks @allisonkarlitskaya , great idea! Changed the PR accordingly.

src/common/Makefile-common.am
```diff
@@ -243,7 +243,7 @@ nodist_noinst_DATA += libpreload-temp-home.so
 EXTRA_DIST += src/common/preload-temp-home.c
 
 libpreload-temp-home.so: src/common/preload-temp-home.c
-	$(AM_V_CCLD) $(CC) $(CFLAGS) -fPIC -shared -o $@ $^
+	$(AM_V_CCLD) $(CC) $(AM_CPPFLAGS) $(CFLAGS) -fPIC -shared -o $@ $^
```
allisonkarlitskaya, line 246: The commit message doesn't explain what this is about.

martinpitt: Right, this can be dropped now; it was necessary before so that this custom .c file build rule would actually use the `_GNU_SOURCE` from `AM_CPPFLAGS`, but that's not necessary any more. I'll drop it. It also fails to build now, will investigate (not obvious why at first sight)

martinpitt: We need this after all, but I moved it to a separate commit and consistently fixed it for all custom compiler invocations.

martinpitt:
Tests fail because the rpm build stumbles over the autogenerated ./src/ws/mock-dbus-tests.c which breaks the rule:
```
#ifdef HAVE_CONFIG_H
#  include "config.h"
#endif
```

I'll move to `git ls-tree` like for the json test, that's more robust after all. No need to do this during an RPM build and for autogenerated files.

martinpitt:
I made a [scratch build](https://koji.fedoraproject.org/koji/taskinfo?taskID=80859545) to test this on all architectures.

tools/test-static-code
```diff
@@ -145,4 +145,18 @@ else
     fail=1
 fi
 
+# for every C file, if there is any line starting with '#', then the first one
```
allisonkarlitskaya, line 148: We should strengthen this requirement, now that we are only inspecting checked in files: they should all start with `#include "config.h"` unconditionally.

Follow up please, so this can make it into the release.

allisonkarlitskaya (approved):

allisonkarlitskaya:
Please let testing farm finish.  This is exactly the sort of change that we want to be checked out as much as possible.

allisonkarlitskaya (approved):

tools/test-static-code
martinpitt: Done in PR #16786

Issue #16775, opened 2022-01-04
"Not found" screen needs to be more helpful


allisonkarlitskaya:
I've had two experiences in the past couple of weeks (one on IRC, one filed as flathub/org.cockpit_project.CockpitClient#2) where I've witnessed users getting confused about the "Not Found" screen due to `cockpit-system` not being installed.

We need to detect if the user is trying to access a path that requires this package to be installed (like `/`) and throw a better error message . During stand-up yesterday, @garrett also mentioned the possibility to offer one-click installation of the required package, but this will require some technical feasibility investigation due to the limited abilities of our error pages.

martinpitt:
Thanks for pointing this out! Indeed, let's fix this.

One-click installation is too complex IMHO -- that requires lots of JS code, interfacing packagekit and the like, which needs to be in a webpack. We could ship that page in cockpit-bridge, but I believe that's overkill. If the error message would say "You need to install the cockpit-system package", this is good enough for almost everyone. The point is not to teach people how to run GNOME software or `apt`/`dnf` on the command line, but to explain what the root cause is.

If we just show that message for path == /, this should work for the majority of cases. One error report had it for http://localhost:9090/cockpit/@localhost/podman/index.html# , but that URL was not specified by a human, but some desktop applet (podman CRC) -- but that demo was just missing the installation of cockpit-podman. Again it's too complex to map out every possible path to its package on a static error page, but I dont' think we have to.

garrett:
What do you think of using an [empty state](https://www.patternfly.org/v4/components/empty-state/) that looks like this?

![image](https://user-images.githubusercontent.com/10246/148079150-fce8a45f-1515-4fd4-a064-ec74bef2651a.png)

Text:

> ### Missing system package
> Please install the `cockpit-system` package with your package manager.

If we could use the one-click installation, we'd have a button to install it as the default action. Since we can't, then we can use some text like in the screenshot.

garrett:
It would be lovely if we could actually give the command based on the OS, but that gets into having all the dnf/yum/apt variants and:

> The point is not to teach people [...] the command line

allisonkarlitskaya:
> Please install the `cockpit-system` package with your package manager.

I wonder if we need to somehow emphasize that this needs to be done on the server you're trying to connect to.  This might otherwise be confusing for Cockpit Client users, for example.

I can't think of any elegant way to phrase that, though.

allisonkarlitskaya:
Also, a small technical note:

[...]

PR #16782, opened 2022-01-05
bridge: give hints for missing cockpit-system package

allisonkarlitskaya:
We've had a few users get confused about the generic "Not found" error
that we throw when trying to login to a system that's missing the
cockpit-system package.

Issue a better error message for that case.

# bridge: Warning on missing cockpit-system package

Previously, when attempting to connect to a system which had `cockpit-bridge` installed, but not `cockpit-system`, you'd get a generic "Not found" error, with no hints about how to fix that.

cockpit-bridge now detects this problem and issues a more helpful message.

image

### DIFF
--- a/src/bridge/cockpitpackages.c
+++ b/src/bridge/cockpitpackages.c
@@ -990,7 +990,17 @@ package_content (CockpitPackages *packages,
         {
           /* On the first round */
           if (l == names)
-            cockpit_web_response_error (response, 404, NULL, NULL);
+            {
+              /* cockpit_packages_resolve() only fails if the entire
+               * package is missing.  Check if that's a package that
+               * ought to have been available and issue a more helpful
+               * message.
+               */
+              if (g_str_equal (name, "shell") || g_str_equal (name, "systemd"))
+                cockpit_web_response_error (response, 404, NULL, _("Server is missing the cockpit-system package"));
+              else
+                cockpit_web_response_error (response, 404, NULL, NULL);
+            }
           else
             cockpit_web_response_abort (response);
           goto out;


src/bridge/cockpitpackages.c
```diff
@@ -990,7 +990,17 @@ package_content (CockpitPackages *packages,
...
           if (l == names)
-            cockpit_web_response_error (response, 404, NULL, NULL);
+            {
+              /* cockpit_packages_resolve() only fails if the entire
+               * package is missing.  Check if that's a package that
+               * ought to have been available and issue a more helpful
+               * message.
+               */
+              if (g_str_equal (name, "shell") || g_str_equal (name, "systemd"))
+                cockpit_web_response_error (response, 404, NULL, "Server is missing the cockpit-system package");
```
martinpitt, line 1000: Nice and simple! But let's please wrap this in `_()` for translation.

martinpitt (changes requested):

allisonkarlitskaya:
Manual testing confirms that this works, but we should probably add a proper integration test for it.

We don't have a lot of prior art for uninstalling part of the thing that we're trying to test from inside a testcase.  Do you want to think about that, @martinpitt ?

allisonkarlitskaya:
(mental note: merge via squash and include mention of #16775 in the commit message)

src/bridge/cockpitpackages.c
allisonkarlitskaya: done, thanks.

I assume cockpit-bridge will be running in a session with the locale correctly set as per the browser locale?

martinpitt:
> We don't have a lot of prior art for uninstalling part of the thing that we're trying to test from inside a testcase.

We actually do, for example [here](https://github.com/cockpit-project/cockpit/blob/4e50211a11d834bc557c45bd80a2348f353fef6d/test/verify/check-metrics#L979) or [here](https://github.com/cockpit-project/cockpit/blob/4e50211a11d834bc557c45bd80a2348f353fef6d/test/verify/check-connection#L606). Note that `cockpit-system` isn't a thing on Arch (it's just "cockpit"), so this will need a special case either way.

This must happen in a destructive test, of course. check-connection sounds quite adequate?

martinpitt (approved):
Thanks! Test would be nice, but it's also nice to add this to today's release, to clean up after our wrong documentation.

allisonkarlitskaya:
> Thanks! Test would be nice, but it's also nice to add this to today's release, to clean up after our wrong documentation.

Agreed.

I consider this as only a partial fix for #16775 (which I intend to keep open).  We can add the tests when we do the proper fix, with @garrett's design suggestions.

martinpitt:
Testing Farm queue is too long, this should be covered well enough (we don't test this scenario in CI anyway), and this is a bit urgent.

PR #16783, opened 2022-01-05
cockpit-bridge: add Recommends and Suggests

allisonkarlitskaya:
We've seen users getting confused by trying to use cockpit-client to log
in to servers where they've installed cockpit-bridge, and not being able
to due to missing cockpit-system.

Add cockpit-system as a recommend from the cockpit-bridge package for
both dpkg and rpm.

Additionally, list networkmanager, packagekit, pcp, sosreport, and
storaged as suggested packages on both distros, and additionally
kdump and selinux for rpm.

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -366,6 +366,14 @@ troubleshooting, interactive command-line sessions, and more.
 %package bridge
 Summary: Cockpit bridge server-side component
 Requires: glib-networking
+Recommends: cockpit-system
+Suggests: cockpit-kdump
+Suggests: cockpit-networkmanager
+Suggests: cockpit-packagekit
+Suggests: cockpit-pcp
+Suggests: cockpit-selinux
+Suggests: cockpit-sosreport
+Suggests: cockpit-storaged
 Provides: cockpit-ssh = %{version}-%{release}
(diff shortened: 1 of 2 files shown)


martinpitt (changes requested):
This is completely the wrong way around IMHO. We must fix the Cockpit-client documentation/home page to not suggest to install bridge. Users should not even be aware of -bridge, that's just an internal library-like implementation detail. Users should install "cockpit" metapackage, or "cockpit-system" or a minimal remote target. We also have legit use cases for installing only cockpit-podman (CRC) or subscription-manager-cockpit (on RHEL desktop), which both pull in -bridge as a dependency.

allisonkarlitskaya:
What do you suggest we do?  Tell users to install the cockpit-system package?

I start to wonder if maybe we should add a metapackage for "full server-side cockpit install" (ie: basically everything except -ws).

martinpitt:
> Tell users to install the cockpit-system package?

That's the minimum, the "cockpit" metapackage would give a more complete experience. In particular, cockpit-packagekit provides the Applications tab, which enables more packages in a graphical way (such as c-podman)

allisonkarlitskaya:
eh.  Let's just go with #16784 and call it a day, then.

PR #16784, opened 2022-01-05
flatpak: stop suggesting to install cockpit-bridge

allisonkarlitskaya:
This language ends up being extremely visible on our flathub page and in
GNOME Software, and it's possibly leading to people running `dnf install
cockpit-bridge` and expecting that to work.

Drop the language and just say that the server needs to have Cockpit
installed, but not enabled.

See also #16775


martinpitt (approved):
Thanks!

PR #16785, opened 2022-01-05
flatpak workflow: validate the generated metainfo

allisonkarlitskaya:
flathub suggests running `appstream-util validate` on each upstream
commit.  Add it to the flatpak testing workflow.

### DIFF
--- a/.github/workflows/flatpak-test.yml
+++ b/.github/workflows/flatpak-test.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Install required build and test dependencies
         run: |
           sudo apt update
-          sudo apt install -y --no-install-recommends autoconf automake elfutils libglib2.0-dev libsystemd-dev xsltproc xmlto flatpak-builder xvfb cockpit-system
+          sudo apt install -y --no-install-recommends autoconf automake elfutils libglib2.0-dev libsystemd-dev xsltproc xmlto flatpak-builder xvfb cockpit-system appstream-util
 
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
```
appstream-util: command not found
```

naja

.github/workflows/flatpak-test.yml
```diff
@@ -30,6 +30,9 @@ jobs:
       - name: Build and install flatpak
         run: sh -x containers/flatpak/install --user --install-deps-from=flathub
 
+      - name: Validate metainfo
+        run: appstream-util validate org.cockpit_project.CockpitClient.metainfo.xml
```
martinpitt, line 34: It's in the [appstream-util](https://packages.debian.org/search?searchon=contents&keywords=appstream-util&mode=path&suite=stable&arch=any) package, so needs to be added to apt above.

martinpitt: We already do this in tools/check-dist, maybe let's just add it there?

allisonkarlitskaya: The problem is that we don't generate this file, except as part of the `containers/flatpak/prepare` script.  Doing it from this workflow is definitely the right place for it, and only adds one more item that we need, when we're already installing stuff anyway.

martinpitt: Makes sense, thanks!

martinpitt (approved):

PR #16786, opened 2022-01-05
tools: Fix comment for config.h check

martinpitt:
The current logic already requires all files to contain the include, as
the sed would exit zero if there is no `^#` at all. Update the comment
accordingly to avoid confusion.

### DIFF
--- a/tools/test-static-code
+++ b/tools/test-static-code
@@ -145,8 +145,7 @@ else
     fail=1
 fi
 
-# for every C file, if there is any line starting with '#', then the first one
-# must be exactly: #include "config.h"
+# every C file should #include "config.h" at the top
 if [ -e .git ]; then
     for f in $(git ls-tree --name-only -r --full-name HEAD | grep '\.c$'); do
         if sed -n '/^#include "config.h"$/ q1; /^\s*#/ q0' "$f"; then


tools/test-static-code
```diff
@@ -145,11 +145,10 @@ else
     fail=1
 fi
 
-# for every C file, if there is any line starting with '#', then the first one
-# must be exactly: #include "config.h"
+# every C file should #include "config.h" at the top
 if [ -e .git ]; then
     for f in $(git ls-tree --name-only -r --full-name HEAD | grep '\.c$'); do
-        if sed -n '/^#include "config.h"$/ q1; /^\s*#/ q0' "$f"; then
```
allisonkarlitskaya, line 152: ... but I think there's actually no change in logic here, is there?  It was already performing as-expected, due to the inversion of the return statuses (which i missed).  In case we see no '#' at all then we'll return 0 (true), which will enter this block and "fail".

All things being equal, I sort I prefer how this was written before, for clarity.

martinpitt: Ah right, I only tested the three cases with the new code, but not with the old one. Indeed if there is no `^#` at all, the old code would also exit zero.

martinpitt: Done.

martinpitt:
Now this is utterly trivial, but :shrug:

allisonkarlitskaya (approved):
Hey.  On the bright side, it removes twice as many lines as it adds!  :partying_face:

PR #16794, opened 2022-01-07
Coverity fixes

martinpitt:
These are pretty much all "harmless", but let's reduce the noise.

### DIFF
--- a/src/pam-ssh-add/pam-ssh-add.c
+++ b/src/pam-ssh-add/pam-ssh-add.c
@@ -46,6 +46,7 @@
 #include "pam-ssh-add.h"
 
 #include "../common/cockpitcloserange.h"
+#include "../common/cockpitmemory.h"
 
 /* programs that can be overwidden in tests */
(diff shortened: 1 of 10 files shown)


src/pam-ssh-add/pam-ssh-add.c
```diff
@@ -809,9 +809,11 @@ static int
 stash_password_for_session (pam_handle_t *pamh,
                             const char *password)
 {
-  if (pam_set_data (pamh, STORED_AUTHTOK, strdup (password),
+  char *password_copy = strdup (password);
```
allisonkarlitskaya, line 812: check for null or use the "x" variant, please

allisonkarlitskaya (changes requested):
everything here looks okay except for one nitpick

src/pam-ssh-add/pam-ssh-add.c
martinpitt: Good catch, done.

allisonkarlitskaya (approved):

PR #16799, opened 2022-01-10
ssh: Drop unused variable in on_channel_data()

martinpitt:
This has existed since the initial commit b3a463baa2fa. Newer gcc
versions now spot this:

    src/ssh/cockpitsshrelay.c:1688:11: error: variable 'size' set but not used [-Werror,-Wunused-but-set-variable]
      guint32 size, i;
              ^
    1 error generated.

 Also make the `i` and `bytes` variables more local while we are at it.

----

Spotted in last night's [failed unit-test container refresh](https://github.com/cockpit-project/cockpit/runs/4755294331?check_suite_focus=true)

### DIFF
--- a/src/ssh/cockpitsshrelay.c
+++ b/src/ssh/cockpitsshrelay.c
@@ -1685,21 +1685,18 @@ on_channel_data (ssh_session session,
                  void *userdata)
 {
   CockpitSshRelay *self = userdata;
-  guint32 size, i;
   gint ret = 0;
-  GBytes *bytes = NULL;
   guint8 *bdata = data;
 
   if (!self->received_frame && !is_stderr)
     {
-      size = 0;
+      guint32 i;
+
       for (i = 0; i < len; i++)
         {
           /* Check invalid characters, prevent integer overflow, limit max length */
           if (i > 7 || bdata[i] < '0' || bdata[i] > '9')
             break;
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (changes requested):
This entire loop is really very pointless.  How about just `g_ascii_strtoll()` and check that `end` is `'\n'` as expected?  As far as I can tell, that's all that's happening here, and we could be doing it with ~20 fewer lines.

martinpitt:
I first tried to use `size_t frame_size_len = strspn (bdata, "0123456789")` instead of the loop, but both that and g_ascii_strtoll() both don't work -- they expect to work on null-terminated strings, but this is (potentially) binary data. I didn't find something like `memspn()` which would take a data size, that'd be really nice. So ATM I'm not sure how to significantly simplify this loop..

allisonkarlitskaya (approved):
The guint32 thing is awkward, but this is obviously an improvement vs. what it was before.

martinpitt:
> The guint32 thing is awkward

You mean because it ought to be `uint32_t`? Hopefully they are the same, but it looks inconsistent indeed.

Anyway, I'm a bit keen on landing this to unbreak the automatic unit-tests container refresh. Happy to do a follow-up to clean this up further.

PR #16801, opened 2022-01-10
DBG: test/verify/check-shell-multi-machine TestMultiMachine.testSshKe…

marusak:
…ySetup

File "/work/bots/make-checkout-workdir/test/verify/check-shell-multi-machine", line 843, in testSshKeySetup
    b.wait_visible("#machine-troubleshoot")

Example: https://logs.cockpit-project.org/logs/pull-16793-20220110-065937-630ab4f4-fedora-coreos/log.html#139-2


marusak:
Notes:
`machine.problem` is `undefined`. It is race, if I reload the page then it works correctly.

PR #16802, opened 2022-01-10
common: start using closefrom(), keeping our hack

allisonkarlitskaya:
4cfc28a09dcf042b8091ca7739407a36f223ec6d introduced our own version of
close_range(), but this turned out to be the wrong API.  closefrom() is
meanwhile available on modern versions of glibc.  It features a
simplified API, with no chance of failure.

Rename cockpitcloserange.c to cockpitclosefrom.c and move the definition
into cockpithacks.h.  Introduce a configure check for closefrom() and
port all users to the closefrom() API.  In case that's not available, we
redirect those calls to our own cockpit_closefrom() (which is now only
conditionally-built).

Since neither the glibc API nor our own code have any chance of failure,
we can drop all the error checking we were doing before.

Incidentally, this should fix warnings we were occasionally seeing from
Valgrind that we were finding and attempting to close its log fd.
Modern Valgrind wraps closefrom() and prevents this from happening.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,10 @@ AC_USE_SYSTEM_EXTENSIONS
 AC_SYS_LARGEFILE
 AC_PROG_RANLIB
 
+AC_CHECK_FUNCS(
+    closefrom
+)
+
 AM_SILENT_RULES([yes])
 
 AC_MSG_CHECKING([whether to install to prefix only])
--- a/src/bridge/bridge.c
+++ b/src/bridge/bridge.c
@@ -38,7 +38,6 @@
 #include "cockpitwebsocketstream.h"
 
 #include "common/cockpitchannel.h"
-#include "common/cockpitcloserange.h"
 #include "common/cockpitfdpassing.h"
 #include "common/cockpithacks-glib.h"
 #include "common/cockpitjson.h"
--- a/src/common/Makefile-common.am
+++ b/src/common/Makefile-common.am
@@ -39,8 +39,7 @@ noinst_LIBRARIES += libcockpit-common.a
 libcockpit_common_a_SOURCES = \
 	src/common/cockpitchannel.c \
 	src/common/cockpitchannel.h \
-	src/common/cockpitcloserange.c \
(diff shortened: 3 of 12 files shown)


martinpitt (commented):
Thanks, this looks nice! It just worries me a bit that there is no manpage for it (unlike `man close_range`, which now works on Fedora 35). The [glibc documentation](https://www.gnu.org/software/libc/manual/html_node/Opening-and-Closing-Files.html) mentions both. It is a bit suspicious that it does not mention any compatibility for `closefrom()` -- does glibc fall back to an internal implementation just like ours if the kernel does not support it? And the other way around, does it actually *use* the kernel's `close_range` syscall if available? I.e. is this safe and also efficient across kernel versions?

allisonkarlitskaya:
```
lis@fedora-toolbox-35 ~> cat x.c 
#define _GNU_SOURCE
#include <unistd.h>

int
main (void)
{
  closefrom(3);
}
lis@fedora-toolbox-35 ~> gcc -o x x.c
lis@fedora-toolbox-35 ~> strace ./x
[... snip ...]
close_range(3, 4294967295, 0)           = 0
[... snip ...]
```

allisonkarlitskaya:
Also, from `close_range(2)`:

> Note: There is no glibc wrapper for this system call; see NOTES.
> Glibc does not provide a wrapper for this system call; call it using syscall(2).

and in particular, if we do use `syscall()` we'd need to implement the fallback for the `ENOSYS` case.  If we use `closefrom()` then that's glibc's business.  All in all, it's just much cleaner to use the libc API for this, I think.

martinpitt:
> There is no glibc wrapper for this system call

That's obsolete -- [glibc added it](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=286286283e9bdc) a while ago, which was the thing that triggered me to do commit 4cfc28a09dc in the first place (I think I saw it on LWN?). But either way, fully agreed that closefrom() is nicer then, thanks!

martinpitt (approved):

PR #16803, opened 2022-01-10
two changes for webpack-jumpstart

allisonkarlitskaya:
- `--force` will delete `dist/` if it already exists (should be uncontroversial)
 - in `--partial` mode, only touch related files (this appeared in an earlier iteration of jumpstart work, but @martinpitt expressed misgivings about how it was duplicating the logic of the build system: let's discuss).

### DIFF
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -17,11 +17,14 @@ set -eu
 cd "$(realpath -m "$0"/../..)"
 . tools/git-utils.sh
 
+force=''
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
OK -- indeed this does duplicate build system logic, and I'm not generally happy about that. None of these commits by themselves are "too complex", but collectively they are. I think by now it is really hard to understand the whole jumpstart/caching setup and the scripts behind it..

PR #16804, opened 2022-01-10
.gitignore: clean up

allisonkarlitskaya:
Remove many ancient and unused patterns, and sort the remaining items.

There may be some things that need to be added back, but let's find those as they pop up...

PR #16805, opened 2022-01-10
src/systemd: new directory

allisonkarlitskaya:
Our increasingly large web of systemd integration mostly lives in src/ws/, for mostly historical reasons.  These days, cockpit-tls has at least as much integration with systemd as does cockpit-ws, and cockpit-session is about to gain some.

Move everything to a new src/systemd/ directory.

 - [x] #16804


configure.ac
```diff
@@ -348,16 +348,6 @@ AC_ARG_WITH([admin-group],
...
 
-
-AC_ARG_WITH(selinux_config_type,
-	    AS_HELP_STRING([--with-selinux-config-type=<type>],
-			   [SELinux context type for cockpit config files]
-                          )
-           )
-
-COCKPIT_SELINUX_CONFIG_TYPE=$with_selinux_config_type
-AC_SUBST(COCKPIT_SELINUX_CONFIG_TYPE)
```
martinpitt, line 359: Please drop it from the textual summary at the end of the script as well.

Makefile.am
```diff
@@ -232,8 +232,13 @@ multi-word: po/cockpit.pot
 title2sentence: multi-word
 	tools/title2sentence.py -i titles -o t2s.sh
 
-include po/Makefile.am
+include containers/Makefile.am
+include containers/flatpak/Makefile.am
```
martinpitt, line 236: This has nothing to do with moving src/systemd, but then I realized you probably just sorted it. Any chance of doing the sorting as a separate commit?

configure.ac
```diff
@@ -277,13 +277,13 @@ AC_MSG_RESULT($asan_status)
 # User and group for running cockpit web server (cockpit-tls or -ws in customized setups)
 
 AC_ARG_WITH(cockpit_user,
-	    AS_HELP_STRING([--with-cockpit-user=<user>],
-			   [User for running cockpit (root)]
+           AS_HELP_STRING([--with-cockpit-user=<user>],
```
martinpitt, line 280: Same here -- consistent indentation is good, but should be a separate commit

martinpitt (changes requested):
Thanks! Just a little cleanup to do, but looks good in general.

Makefile.am
allisonkarlitskaya: ok.

configure.ac
allisonkarlitskaya: this was an accident, i think

allisonkarlitskaya: I know what happened, because it just happened again: I have a habit of using `git show` + copy + `patch -R` + `git commit --amend` to get rid of things that I don't want in commits, and sometimes it turns tabs into spaces.

allisonkarlitskaya:
I had some more stuff already queued up in a follow-up commit, but since we're gonna do another round here, let's include it.

martinpitt (approved):
Thanks!

PR #16806, opened 2022-01-10
test: Don't retry failed pixel tests

mvollmer:
We expect pixel test failures to be deterministic, expecially since we
wait for all animations to have stopped.  Often, failed pixel tests
are even expected since we update the references only late in the life
of a pull request.


mvollmer:
Oh man, I broke a unused reference image...

mvollmer:
Proof that this works: https://logs.cockpit-project.org/logs/pull-16806-20220110-133700-5faba08d-fedora-35/log.html#171

The failed pixel test has not been retried.

PR #16809, opened 2022-01-10
test: sudo and socketpair() now work properly on fedora-testing

skobyda:
The fix for https://bugzilla.redhat.com/show_bug.cgi?id=1814569 is now
present in fedora-testing
This PR should unblock https://github.com/cockpit-project/bots/pull/2785

### DIFF
--- a/test/verify/check-static-login
+++ b/test/verify/check-static-login
@@ -493,7 +493,7 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
 
         # This should work, but it will not because of
         # https://bugzilla.redhat.com/show_bug.cgi?id=1814569
-        working_images = []  # none yet :-(
+        working_images = ["fedora-testing"]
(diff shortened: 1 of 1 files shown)


skobyda:
This is weird, trying to trigger it with bots from https://github.com/cockpit-project/bots/pull/2785, test-triggers says "Pull request author 'skobyda' isn't allowed.", yet I'm amoung contributors https://github.com/orgs/cockpit-project/teams/contributors/members

martinpitt (approved):
Nice, thanks! Ugh, this will give us some fun in the next days, when this fix gets into F34/35. But at least now it's easy enough to fix.

I pushed a clean rebase and triggered the test run against the image refresh.

martinpitt:
There is another SELinux problem, but let's handle that in the image refresh and land this anyway -- we don't run the f-testing image anywhere except image refreshes, after all.

PR #16813, opened 2022-01-11
test: Skip non-pixel-tests for the mobile scenario

mvollmer:
The point is to reduce the load on the CI machinery, and in turn allow
us to introduce more layout scenarios, such as a "old-laptop" size
between "mobile" and "desktop".

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -37,6 +37,7 @@ import time
 import unittest
 import gzip
 import inspect
+import glob
 
 import testvm
 import cdp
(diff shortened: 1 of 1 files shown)


mvollmer:
Works: https://logs.cockpit-project.org/logs/pull-16813-20220111-093123-e3d1f72f-fedora-35-mobile/log.html

martinpitt (changes requested):
Thanks! Just some confusion about the documentation in the comment. Also, I'm not sure if you actually want to land this as-is, as you and Lis came up with an alternate strategy?

PR #16815, opened 2022-01-11
test: Use MOZ_HEADLESS_WIDTH et al to control Firefox window size

mvollmer:
The --window-size parameter has no effect.

### DIFF
--- a/test/common/cdp.py
+++ b/test/common/cdp.py
@@ -139,6 +139,8 @@ class Firefox(Browser):
 
         cmd = [exe, "-P", "blank", f"--window-size={window_width},{window_height}", f"--remote-debugging-port={cdp_port}", "--no-remote", "localhost"]
         if not show_browser:
+            env["MOZ_HEADLESS_WIDTH"] = str(window_width)
+            env["MOZ_HEADLESS_HEIGHT"] = str(window_height)
             cmd.insert(3, "--headless")
         return cmd
 


test/common/cdp.py
```diff
@@ -137,7 +137,10 @@ def cmd(self, cdp_port, env, show_browser, window_width, window_height,
         with open(os.path.join(profile, "handlers.json"), "w") as f:
             f.write('{"defaultHandlersVersion":{"en-US":4},"mimeTypes":{"application/xz":{"action":0,"extensions":["xz"]}}}')
 
-        cmd = [exe, "-P", "blank", f"--window-size={window_width},{window_height}", f"--remote-debugging-port={cdp_port}", "--no-remote", "localhost"]
+        env["MOZ_HEADLESS_WIDTH"] = str(window_width)
+        env["MOZ_HEADLESS_HEIGHT"] = str(window_height)
```
martinpitt, line 141: This needs to be conditional, as I'm fairly sure that `--window-size` *does* apply to an interactive browser with `TEST_SHOW_BROWSER=1`. Can you please move it underneath the `if not show_browser:` in line 144?

martinpitt (changes requested):
Thanks, well spotted! (But also, ugh!)

test/common/cdp.py
mvollmer: Ok, done!

martinpitt (approved):
Danke!

PR #16816, opened 2022-01-11
Quench warnings in console

marusak:
I have been looking a lot into browser console lately and these nothing-saying warnings were always bothering me. So here I am getting rid of those I saw.

### DIFF
--- a/pkg/base1/cockpit.js
+++ b/pkg/base1/cockpit.js
@@ -798,7 +798,7 @@ function Channel(options) {
         self.valid = false;
         if (transport && id)
             transport.unregister(id);
-        if (closed.message)
+        if (closed.message && !options.err)
             console.warn(closed.message);
         self.dispatchEvent("close", closed);
         if (waiting)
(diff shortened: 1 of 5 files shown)


pkg/base1/cockpit.js
```diff
@@ -798,7 +798,7 @@ function Channel(options) {
         self.valid = false;
         if (transport && id)
             transport.unregister(id);
-        if (closed.message)
+        if (closed.message && options.err !== "message")
```
marusak, line 801: This one I am not super sure about - can it possibly hide some messages we might need?

martinpitt: That's a good point -- it's [documented](https://cockpit-project.org/guide/latest/cockpit-spawn.html) to return the error through the API and *not* the console, and we specifically use `message` in many cases to *avoid* it going to the journal or console, and just ignore them; the latter should really use `err: "ignore"`, but I figure we are not terribly consistent about it.

But either way, this would still mean that the console would get closed messages for `err` == `"ignore"` or `"out"`, which we also don't want. So it almost seems to me that this should be `&& !options.err` ?

martinpitt (changes requested):
Nice cleanup, thank you!

Nitpick: "Quench" feels like a strange word, did you mean "quiesce"? (But that's not part of the commit messages, so nevermind!)

martinpitt (approved):
Dakujem!

PR #16819, opened 2022-01-11
ws: add support for connecting to unix sockets

allisonkarlitskaya:
cockpit.conf can now specify a unix socket address in the ConnectTo= field of a given authentication method.  If this is provided, and no Command= is present, then a connection will be opened to that address instead of spawning a command.                        

This is the "easy stuff" split out from #16808, and is already enough to help us with cockpit-cloud-connector.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -1045,6 +1045,26 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
         b.wait_visible("#login")
         b.assert_pixels("body", "login-screen")
 
+    @skipImage("no cockpit-ws package", "fedora-coreos")
+    def testAuthUnixPath(self):
+        '''test UnixPath for auth method in cockpit.conf'''
+        m = self.machine
+
+        m.execute(['systemctl', 'start', 'cockpit-session.socket'])
+        m.write('/etc/cockpit/cockpit.conf', '''
+[Negotiate]
+Action=none
+
+[Basic]
+UnixPath=/run/cockpit/session
+''')
+
+        # make sure this isn't being run via spawning
+        m.execute(['chmod', '700', f'{self.libexecdir}/cockpit-session'])
+
+        m.start_cockpit()
+        self.login_and_go("/system")
+
 
 @skipDistroPackage()
 class TestReverseProxy(MachineCase):
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -257,9 +257,8 @@ done
 for data in doc man pixmaps polkit-1; do
     rm -r %{buildroot}/%{_datadir}/$data
 done
-for lib in systemd tmpfiles.d; do
-    rm -r %{buildroot}/%{_prefix}/%{__lib}/$lib
-done
+rm -r %{buildroot}/%{_prefix}/%{__lib}/tmpfiles.d
+find %{buildroot}/%{_unitdir}/ -type f ! -name 'cockpit-session*' -delete
 for libexec in cockpit-askpass cockpit-session cockpit-ws cockpit-tls cockpit-wsinstance-factory cockpit-client cockpit-client.ui cockpit-desktop cockpit-certificate-helper cockpit-certificate-ensure; do
     rm %{buildroot}/%{_libexecdir}/$libexec
 done
@@ -277,6 +276,8 @@ for pkg in apps packagekit pcp playground storaged; do
 done
 # files from -tests
 rm -f %{buildroot}/%{pamdir}/mock-pam-conv-mod.so
+rm -f %{buildroot}/%{_unitdir}/cockpit-session.socket
+rm -f %{buildroot}/%{_unitdir}/cockpit-session@.service
 # files from -pcp
 rm -r %{buildroot}/%{_libexecdir}/cockpit-pcp %{buildroot}/%{_localstatedir}/lib/pcp/
 # files from -storaged
@@ -636,6 +637,8 @@ These files are not required for running Cockpit.
 
 %files -n cockpit-tests -f tests.list
 %{pamdir}/mock-pam-conv-mod.so
+%{_unitdir}/cockpit-session.socket
+%{_unitdir}/cockpit-session@.service
 
 %package -n cockpit-pcp
 Summary: Cockpit PCP integration
--- a/doc/authentication.md
+++ b/doc/authentication.md
@@ -29,6 +29,9 @@ Authentication commands are called with a single argument which is the host that
 is connecting to. They communicate with their parent process using the cockpit protocol on
 stdin and stdout.
 
+Instead of `Command`, it's also possible to specify `UnixPath` to connect to a unix socket at a
+given path.  The protocol is the same.
+
 Credentials can then be retrieved by issuing a authorize command with a challenge. The challenge
 should correspond to the authorization type in header (ei: Basic or Bearer). For example:
 
--- a/selinux/cockpit.te
+++ b/selinux/cockpit.te
@@ -46,6 +46,9 @@ kernel_read_system_state(cockpit_ws_t)
 # cockpit-tls can execute cockpit-ws
 can_exec(cockpit_ws_t,cockpit_ws_exec_t)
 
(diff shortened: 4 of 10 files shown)


test/verify/check-connection
```diff
@@ -965,6 +965,51 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
         m.write("/etc/cockpit/ws-certs.d/0-self-signed-ca.pem", "FAKE CERT FOR TESTING\n")
         self.assertEqual(m.execute("curl -sfS http://localhost:9090/ca.cer"), "FAKE CERT FOR TESTING\n")
 
+    def testAuthUnixPath(self):
+        '''test UnixPath for auth method in cockpit.conf'''
+        m = self.machine
+
+        # selinux doesn't like this...
+        if m.image in ['centos-8-stream', 'fedora-34', 'fedora-35', 'fedora-coreos', 'rhel-8-5', 'rhel-8-6', 'rhel-9-0']:
```
martinpitt, line 973: I propose to check `selinuxenabled` on the image, which will be less error prone.

martinpitt (changes requested):
Not quite obvious why this fails *only* on Debian, and not on Ubuntu or anywhere else..

```
Jan 14 18:00:26 debian cockpit-ws[2003]: /run/cockpit/session: incorrect protocol: received invalid length prefix
Jan 14 18:00:26 debian cockpit-ws[2003]: /shell/index.html: external channel failed: protocol-error
```

The code changes LGTM, this just needs some debugging of this test. Thanks!

test/verify/check-connection
allisonkarlitskaya: I was looking for something like this!  Thanks!

allisonkarlitskaya: I thought this was an attribute on the image.... This seems annoying to use.  What would you propose exactly?

```
    if m.execute("selinuxenabled && echo 'enabled' && setenforce 0 || true"):
         self.allow_journal_messages()...
```
?

martinpitt: Yes, something like this. The existing cases (`git grep selinuxenabled test` all do an `if selinuxenabled; then action..` kind of thing. (Please use `if`, it feels a little cleaner than alternative truths; but no strong opinion)

Also, does this actually work anywhere? I have a déjà vu that I worked on some patch to our SELinux policy to fix that, but that's not in this series? I.e. it feels like we should pull that into this series and drop all these special cases?

allisonkarlitskaya: The problem is that I want to conditionalise ignoring this journal message.  Or do you think I should just do that unconditionally?

martinpitt: I am mostly concerned about how we get rid of this.

 * Is this something we can fix in our own SELinux policy? If so, then let's do that right away (I'm happy to help with this, of course!), and we shouldn't need the test quirk at all.
 *  If this is outside of our own policy, then we need to have a plan how to enable this: A bugzilla against selinux-policy, reference it in a comment, and *then* the OS list actually makes sense -- we can then drop fixed OSes to make sure it works and doesn't regress.

So all in all, I retract my proposal to use `selinuxenabled`, it doesn't make sense.

allisonkarlitskaya: Take a look at 
https://github.com/cockpit-project/cockpit/commit/98e1ef7f6dd238239380ef9bbf66f778f952d590 for what the fix would look like.  I didn't include it here because I'm not sure we should carry that just for a test case...

martinpitt: Ah, I remember this :wink: I wouldn't mind adding that selinux fix right now to spare us the cleanup in the future. But this confirms that the fix will not be OS specific, so the long and detailed OS list does not make much sense after all, and checking `selinuxenabled` seems right. Sorry for having forgotten all the context here in the last three weeks!

So if you prefer to withhold that policy, let's check selinuxenabled and otherwise keep the current quirk.

allisonkarlitskaya: I think maybe we could try to merge everything from the other PR except the two last commits that actually do the switchover.  That would provide a semi-supported mechanism to use this stuff without it being "the way".  In that case I'd pick the selinux fix.

test/common/testlib.py
```diff
@@ -2419,6 +2419,8 @@ def test_main(options=None, suite=None, attachments=None, **kwargs):
                         default=None, help="Run this test against an already running machine")
     parser.add_argument('--browser', metavar="hostname[:port]", dest="browser",
                         default=None, help="When using --machine, use this cockpit web address")
+    parser.add_argument('-c', action='store_true',
+                        help="Equivalent to '--machine 127.0.0.2:2201 --browser 127.0.0.2:9091'")
```
martinpitt, line 2423: Unrelated, and most probably debugging leftover. Let's rather keep this local (I have a `$RUNC` variable for that)

tools/cockpit.spec
```diff
@@ -277,6 +277,8 @@ for pkg in apps packagekit pcp playground storaged; do
 done
 # files from -tests
 rm -f %{buildroot}/%{pamdir}/mock-pam-conv-mod.so
+rm -f ${buildroot}/%{_unitdir}/cockpit-session.socket
+rm -f ${buildroot}/%{_unitdir}/cockpit-session@.service
```
martinpitt, line 281: Wrong currency, it's not dollars, but you only get a percentage :grin: 

This fixes the [unpackaged files](https://cockpit-logs.us-east-1.linodeobjects.com/pull-16819-20230104-101153-3466f0a9-rhel-8-8/log.html) failure.

The [opposite](https://cockpit-logs.us-east-1.linodeobjects.com/pull-16819-20230104-101153-3466f0a9-rhel-8-8-distropkg/log.html) happens because "when not building basic packages, remove their files" removes the entirety of %{buildroot}/%{_prefix}/%{__lib}/systemd. I'll get that fixed as well.

test/common/testlib.py
martinpitt: Removed.

tools/cockpit.spec
martinpitt: Fixed both.

martinpitt:
Bah, accidentally  pushed the wrong worktree, sorry for the confusing diff above!

allisonkarlitskaya:
:+1: from me, but I need another reviewer :)

martinpitt (approved):
Thank you! LGTM, I was just waiting for green.

test/verify/check-connection
```diff
@@ -1045,6 +1045,26 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
         b.wait_visible("#login")
         b.assert_pixels("body", "login-screen")
 
+    def test_auth_unix_path(self):
```
martinpitt, line 1048: Our test names are in camelCaps.

test/verify/check-connection
```diff
@@ -1045,6 +1045,26 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
...
+        '''test UnixPath for auth method in cockpit.conf'''
+        m = self.machine
+
+        m.execute(['systemctl', 'start', 'cockpit-session.socket'])
```
martinpitt, line 1052: This [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-16819-20230104-122301-a9528565-fedora-coreos/log.html#331) on coreos:
```
systemd[1]: Starting cockpit-session.socket - Initiator socket for Cockpit sessions...
systemd[1014]: cockpit-session.socket: Failed to resolve group cockpit-wsinstance: No such process
systemd[1]: cockpit-session.socket: Control process exited, code=exited, status=216/GROUP
```
That doesn't exist because the cockpit-ws package isn't installed on coreos. We need to skip that test here.

martinpitt (changes requested):

test/verify/check-connection
martinpitt: Updated.

test/verify/check-connection
martinpitt: Added a skip.

martinpitt (approved):

PR #16821, opened 2022-01-12
Fix TestConnection.testCockpitDesktop arch test, and race condition

martinpitt:
Fixes [this top flake](https://logs.cockpit-project.org/logs/pull-16795-20220107-134435-a1131daf-arch/log.html#15-2)

On [arch it fails most of the time](https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&days=14&test=test%2Fverify%2Fcheck-connection+TestConnection.testCockpitDesktop#)

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -842,7 +842,7 @@ ResultActive=yes""")
         m.write(r"/tmp/browser.sh", """#!/bin/sh -e
 curl --silent --compressed -o /tmp/out.html "$@"
 # wait until privileged bridge starts
-until pgrep -f cockpit-bridge.*--privileged; do sleep 1; done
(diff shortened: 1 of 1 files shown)


jelly (approved):

jelly:
Nice! Shouldn't we start defaulting to the JS polkit rules it seems starting polkit 0.106 the JS rules are the default.

martinpitt:
IMNSHO these JS polkit rules were a terrible idea from the start (not auditible, JS interpreter in critical security path, mozjs versions are not supportable long-term). That's also why Debian/Ubuntu have rejected to ever package these versions, and later polkit brought back the shim support for the *.pkla shim rules. So from my POV pkla is the primary format that everyone (and for sure all the distros) should keep :grin:

PR #16823, opened 2022-01-12
test: Rework pixel testing and mobile layout logic

mvollmer:
test/common/testlib.py has a set of layouts and test code can switch a
Browser into one via Browser.set_layout.  Right now there is only
"desktop" and "mobile", but more can be easily added.
    
Browser.assert_pixels cycles through all the defined layouts and does
a pixel test for each.

 - [ ] merge change to testmap at the same time, https://github.com/cockpit-project/bots/pull/2848

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -93,6 +93,42 @@ opts.address = None
 opts.jobs = 1
 opts.fetch = True
 
+# Browser layouts
+#
+# A browser can be switched into a number of different layouts, such
+# as "desktop" and "mobile".  A default set of layouts is defined
+# here, but projects can override this with a file called
+# "test/browser-layouts.json".
+#
+# Each layout defines the size of the shell (where the main navigation
+# is) and also the size of the content iframe (where the actual page
+# like "Networking" or "Overview" is displayed).
+#
+# When the browser layout is switched (by calling Browset.set_layout),
+# this will either set the shell size or the content size, depending
+# on which frame is current (as set by Browser.enter_page or
+# Browser.leave_page).
+#
+# This makes sure that pixel tests for the whole content iframe are
+# always the exact size as specified in the layout definition, and
+# don't change size when the navigation stuff in the shell changes.
+#
+# The browser starts out in the first layout of this list, which is
+# "desktop" by default.
+
+default_layouts = [
+    {
+        "name": "desktop",
+        "shell_size": [1920, 1200],
+        "content_size": [1680, 1130]
+    },
+    {
+        "name": "mobile",
+        "shell_size": [414, 1920],
+        "content_size": [414, 1856]
+    }
+]
+
 
 def attach(filename, move=False):
     '''Put a file into the attachments directory
@@ -124,12 +160,20 @@ class Browser:
         self.pixels_label = pixels_label
         self.machine = machine
         path = os.path.dirname(__file__)
-        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace, fixed_content_size=self.pixels_label,
+        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
                            inject_helpers=[os.path.join(path, "test-functions.js"), os.path.join(path, "sizzle.js")])
         self.password = "foobar"
         self.timeout_factor = int(os.getenv("TEST_TIMEOUT_FACTOR", "1"))
         self.failed_pixel_tests = 0
         self.body_clip = None
+        try:
+            with open(f'{TEST_DIR}/browser-layouts.json') as fp:
+                self.layouts = json.load(fp)
+        except FileNotFoundError:
+            self.layouts = default_layouts
+        self.current_layout = self.layouts[0]
+        size = self.current_layout["shell_size"]
+        self._set_window_size(size[0], size[1])
 
     def title(self):
         return self.cdp.eval('document.title')
@@ -597,16 +641,6 @@ class Browser:
             self.eval_js('window.localStorage.setItem("superuser:%s", "%s");' % (user, "any" if superuser else "none"))
         self.click('#login-button')
 
-    def adjust_window_for_fixed_content_size(self):
-        # When doing pixel tests, try to give the content iframe the
-        # expected fixed size so that pixel tests don't have to adapt
-        # to changes in the shell layout too much.  But don't bother
-        # when the browser is visible since we don't control its
-        # window size and it will likely be too small.
-        if self.pixels_label and not self.cdp.show_browser:
-            self.call_js_func("ph_adjust_content_size", self.cdp.content_width, self.cdp.content_height)
-            self.body_clip = self.call_js_func('ph_element_clip', 'body')
-
     def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False, password=None,
                      legacy_authorized=None):
         href = path
@@ -623,7 +657,6 @@ class Browser:
         self.expect_load()
         self._wait_present('#content')
         self.wait_visible('#content')
-        self.adjust_window_for_fixed_content_size()
         if path:
             self.enter_page(path.split("#")[0], host=host)
 
@@ -658,7 +691,6 @@ class Browser:
         self.expect_load()
         self._wait_present('#content')
         self.wait_visible('#content')
-        self.adjust_window_for_fixed_content_size()
         if path:
             if path.startswith("/@"):
                 host = path[2:].split("/")[0]
@@ -671,15 +703,18 @@ class Browser:
         if (self.attr("#toggle-menu", "aria-expanded") != "true"):
             self.click("#toggle-menu")
 
+    def layout_is_mobile(self):
+        return self.current_layout["shell_size"][0] < 420
+
     def open_superuser_dialog(self):
-        if self.cdp.mobile:
+        if self.layout_is_mobile():
             self.open_session_menu()
             self.click("#super-user-indicator-mobile button")
         else:
             self.click("#super-user-indicator button")
 
     def check_superuser_indicator(self, expected):
-        if self.cdp.mobile:
+        if self.layout_is_mobile():
             self.open_session_menu()
             self.wait_text("#super-user-indicator-mobile", expected)
             self.click("#toggle-menu")
@@ -717,8 +752,6 @@ class Browser:
         (useful for remote hosts).
         '''
         self.switch_to_top()
-        if self.cdp.mobile:
-            self.click("#nav-system-item")
         self.click(f"#host-apps a[href='{path}']")
         if enter:
             # strip off parameters after hash
@@ -767,12 +800,42 @@ class Browser:
             attach(filename, move=True)
             print("Wrote HTML dump to " + filename)
 
-    def assert_pixels(self, selector, key, ignore=[]):
-        """Compare the given element with its reference"""
+    def _set_window_size(self, width, height):
+        self.cdp.invoke("Emulation.setDeviceMetricsOverride",
+                        width=width, height=height,
+                        deviceScaleFactor=0, mobile=False)
 
-        if not (Image and self.pixels_label and self.cdp and self.cdp.valid):
+    def set_layout(self, name):
+        layout = [lo for lo in self.layouts if lo["name"] == name][0]
+        if layout != self.current_layout:
+            self.current_layout = layout
+            size = layout["shell_size"]
+            self._set_window_size(size[0], size[1])
+            self._adjust_window_for_fixed_content_size()
+
+    def _adjust_window_for_fixed_content_size(self):
+        if self.eval_js("window.name").startswith("cockpit1:"):
+            # Adjust the window size further so that the content is
+            # exactly the expected size.  This will make sure that
+            # pixel tests of the content will not be affected by
+            # changes in shell navigation elements around it.  It is
+            # important that we do this only after getting the shell
+            # into about the right size so that it switches into the
+            # right layout mode.
+            shell_size = self.current_layout["shell_size"]
+            want_size = self.current_layout["content_size"]
+            have_size = self.eval_js("[ document.body.offsetWidth, document.body.offsetHeight ]")
+            delta = (want_size[0] - have_size[0], want_size[1] - have_size[1])
+            if delta[0] != 0 or delta[1] != 0:
+                self._set_window_size(shell_size[0] + delta[0], shell_size[1] + delta[1])
+
+    def assert_pixels_in_current_layout(self, selector, key, ignore=[]):
+        """Compare the given element with its reference in the current layout"""
+
+        if not (Image and self.pixels_label):
             return
 
+        self._adjust_window_for_fixed_content_size()
         self.call_js_func('ph_scrollIntoViewIfNeeded', selector)
         self.call_js_func('ph_blur_active')
 
@@ -814,8 +877,8 @@ class Browser:
 
         ignore_rects = relative_clips(list(map(lambda item: selector + " " + item, ignore)))
         base = self.pixels_label + "-" + key
-        if self.cdp.mobile:
-            base += "-mobile"
+        if self.current_layout != self.layouts[0]:
+            base += "-" + self.current_layout["name"]
         filename = base + "-pixels.png"
         ref_filename = os.path.join(reference_dir, filename)
         ret = self.cdp.invoke("Page.captureScreenshot", clip=rect, no_trace=True)
@@ -898,6 +961,19 @@ class Browser:
                 attach(delta_filename, move=True)
                 print("Differences in pixel test " + base)
                 self.failed_pixel_tests += 1
+
+    def assert_pixels(self, selector, key, ignore=[], skip_layouts=[]):
+        """Compare the given element with its reference in all layouts"""
+
+        if not (Image and self.pixels_label):
+            return
+
+        previous_layout = self.current_layout["name"]
+        for layout in self.layouts:
+            if layout["name"] not in skip_layouts:
+                self.set_layout(layout["name"])
+                self.assert_pixels_in_current_layout(selector, key, ignore=ignore)
+        self.set_layout(previous_layout)
 
     def get_js_log(self):
         """Return the current javascript log"""
--- a/test/run
+++ b/test/run
@@ -11,10 +11,9 @@ TEST_SCENARIO=${TEST_SCENARIO:-verify}
 TEST_OS_DEFAULT=$(PYTHONPATH=bots python3 -c 'from machine.machine_core.constants import TEST_OS_DEFAULT; print(TEST_OS_DEFAULT)')
 
 case $TEST_SCENARIO in
-    verify|mobile|devel|firefox|firefox-devel)
+    verify|devel|firefox|firefox-devel)
         [ "${TEST_SCENARIO}" = "${TEST_SCENARIO%%devel}" ] || export NODE_ENV=development
         [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##firefox}" ] || export TEST_BROWSER=firefox
-        [ "${TEST_SCENARIO}" = "${TEST_SCENARIO##mobile}" ] || export TEST_MOBILE=true
         test/image-prepare --verbose $TEST_OS
         test/common/run-tests --jobs ${TEST_JOBS:-1} --test-dir test/verify
         ;;
--- a/test/verify/check-networkmanager-vlan
+++ b/test/verify/check-networkmanager-vlan
@@ -45,9 +45,7 @@ class TestNetworkingVLAN(NetworkCase):
         # Remove focus ring around the close button, which causes pixel tests retries.
         b.blur("#network-vlan-settings-dialog > button[aria-label=Close]")
         b.wait_visible("#network-vlan-settings-interface-name-input")
-        # Ignore flaky pixel issues with the select triangle
-        ignoredClasses = ["#network-vlan-settings-parent-select"] if self.browser.cdp.mobile else []
-        b.assert_pixels("#network-vlan-settings-dialog", "networking-vlan-settings-dialog", ignore=ignoredClasses)
+        b.assert_pixels("#network-vlan-settings-dialog", "networking-vlan-settings-dialog")
 
         b.select_from_dropdown("#network-vlan-settings-parent-select", iface)
         b.set_input_text("#network-vlan-settings-interface-name-input", "tvlan")
--- a/test/verify/check-packages
+++ b/test/verify/check-packages
@@ -61,12 +61,7 @@ class TestPackages(MachineCase):
         self.login_and_go("/playground/pkgs")
 
         b.switch_to_top()
-        if b.cdp.mobile:
-            b.click("#nav-system-item")
         b.wait_visible("#nav-system li:contains(Terminal)")
-        if b.cdp.mobile:
-            # close menu again
-            b.click("#nav-system-item")
 
         m.execute("mkdir /usr/share/cockpit/test")
         m.write("/usr/share/cockpit/test/manifest.json", test_manifest)
@@ -75,8 +70,6 @@ class TestPackages(MachineCase):
         b.click("#reload")
 
         b.switch_to_top()
-        if b.cdp.mobile:
-            b.click("#nav-system-item")
         b.click("#nav-system a:contains(Test)")
 
         b.enter_page("/test/test")
@@ -88,12 +81,7 @@ class TestPackages(MachineCase):
         b.click("#reload")
 
         b.switch_to_top()
-        if b.cdp.mobile:
-            b.click("#nav-system-item")
         b.wait_not_present("#nav-system a:contains(Test)")
-        if b.cdp.mobile:
-            # close menu again
-            b.click("#nav-system-item")
 
 
 if __name__ == '__main__':
--- a/test/verify/check-pages
+++ b/test/verify/check-pages
@@ -54,17 +54,10 @@ class TestPages(MachineCase):
 
     def check_system_menu(self, label, present):
         b = self.browser
-        if b.cdp.mobile:
-            b.click("#nav-system-item")
-
         if present:
             b.wait_visible(f"#host-apps li a:contains('{label}')")
         else:
             b.wait_not_present(f"#host-apps li a:contains('{label}')")
-
-        if b.cdp.mobile:
-            # close menu again
-            b.click("#nav-system-item")
 
     def open_lang_modal(self):
         self.browser.switch_to_top()
@@ -432,8 +425,6 @@ OnCalendar=daily
         self.allow_journal_messages("invalid or unusable locale: de_DE.UTF-8")
 
         self.login_and_go()
-        if b.cdp.mobile:
-            b.click("#nav-system-item")
 
         filter_sel = ".pf-c-search-input__text-input"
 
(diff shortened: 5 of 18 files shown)


mvollmer:
TODO:

- [ ] skip non-pixel tests as in #16813
- [ ] run the fedora-35/pixels scenario

mvollmer:
Getting rid of references to `self.browser.cdp.mobile` and the `*/mobile` context comes in a separate PR.

mvollmer:
Hmm, tests not derived from MachineCase will still be running in the */pixels scenario (since the skip logic is in the setUp of MachineCase).  Have to find a better place...

mvollmer:
Not bad!

The tooltips for dialog titles are not triggered anymore, very likely because the dialog is opened in the "desktop" layout and we switch to "mobile" while it is already open. I think that is okay and we can just update the pixel test references.

There are a couple of attempts to take a screenshot of a 0x0 element, which probably means that the element is invisible.  Needs investigating.

martinpitt:
> tests not derived from MachineCase will still be running in the */pixels scenario

That should only be a few trivial ones, though? Do they hurt?

mvollmer:
What about this idea: We don't skip any tests with TEST_PIXELS, but we only run `fedora-35/pixels` and no plain `fedora-35`?

martinpitt:
@mvollmer : That is pretty much the status quo, isn't it? Then we can just as well keep that

mvollmer:
> @mvollmer : That is pretty much the status quo, isn't it? Then we can just as well keep that

We would still get rid of the */mobile scenario and fedora-35/pixels would test all layouts, including the upcoming "medium" one.

martinpitt:
Right, I just meant that the only benefit would be to make it more explicit which OS runs the pixel tests, right? Locally that doesn't make much of a difference.

mvollmer:
> Right, I just meant that the only benefit would be to make it more explicit which OS runs the pixel tests, right?

If we are only talking about how to enable pixel tests, yes.  (The main change here is of course that assert_pixel now does both desktop and mobile tests in a `for` loop.)

> Locally that doesn't make much of a difference.

Locally, the change with this PR is that now one needs to set TEST_PIXELS explicitly in order to run pixel tests.

mvollmer:
I have removed changes to the "control layer" again.  We keep test/reference-image to control when to run pixel tests, there is no TEXT_PIXELS or a */pixels scenario.

This means: and we don't need the code to skip tests that don't call assert_pixels, and we don't need to worry about bootstrapping new pixel tests.

mvollmer:
Oh, Firefox can't do layout changes, have to account for that, hmm.

mvollmer:
> Oh, Firefox can't do layout changes, have to account for that, hmm.

Firefox supports Emulate.setDeviceMetricsOverride, so we can just use that instead of the method on Page.  This finally gives us a way to reliably set the size of Firefox windows. Hurray!

test/common/testlib.py
```diff
@@ -120,14 +131,18 @@ def __init__(self, address, label, machine, pixels_label=None, port=None):
...
-        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace, fixed_content_size=self.pixels_label,
+        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
+                           mobile=(self.current_layout == "mobile"),
                            inject_helpers=[os.path.join(path, "test-functions.js"), os.path.join(path, "sizzle.js")])
         self.password = "foobar"
         self.timeout_factor = int(os.getenv("TEST_TIMEOUT_FACTOR", "1"))
         self.failed_pixel_tests = 0
         self.body_clip = None
+        size = layout_scenarios[self.current_layout]["shell_size"]
+        self._set_window_size(size[0], size[1])
```
mvollmer, line 145: We should just call set_layout("desktop" or "mobile") here.

test/common/testlib.py
```diff
@@ -797,6 +830,15 @@ def assert_pixels(self, selector, key, ignore=[]):
         time.sleep(0.5)
         self.wait_js_cond('ph_count_animations(%s) == 0' % jsquote(selector))
 
+        reference_dir = os.path.join(TEST_DIR, 'reference')
```
mvollmer, line 833: This code block doesn't need to be moved here, let's leave it where it was.

test/verify/check-shell-host-switching
```diff
@@ -136,25 +136,25 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
         m2 = self.machines["machine2"]
         m3 = self.machines["machine3"]
 
+        if b.cdp.mobile:
```
mvollmer, line 139: Remove this.

test/verify/check-shell-host-switching
```diff
@@ -136,25 +136,25 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
...
+        b.set_layout("mobile")
+        b.click("#nav-system-item")
+        b.wait_visible("#nav-system.interact")
+        b.assert_pixels_in_current_layout("#nav-system", "nav-system")
+        b.click("#nav-system-item")
+        b.wait_not_present("#nav-system.interact")
+        b.set_layout("desktop")
 
-        b.assert_pixels("#hosts-sel", "hosts-sel-closed")
+        b.assert_pixels("#hosts-sel", "hosts-sel-closed")  # XXX - this is now the same as "nav-system", no?
```
mvollmer, line 157: Check this.

test/common/testlib.py
```diff
@@ -93,6 +93,17 @@
...
+layout_scenarios = {
+    "desktop": {
+        "shell_size": (1920, 1200),
+        "content_size": (1680, 1130)
+    },
+    "mobile": {
+        "shell_size": (414, 1920),
+        "content_size": (414, 1856)
+    }
+}
```
allisonkarlitskaya, line 105: I'm not sure this belongs in `common/`.  The choice of which sizes to test is probably reasonably per-project, no?

I'd expect this to live in a file in the specific repository, or in the reference images repository.

test/common/testlib.py
```diff
@@ -122,14 +133,18 @@ def __init__(self, address, label, machine, pixels_label=None, port=None):
         self.default_user = "admin"
         self.label = label
         self.pixels_label = pixels_label
+        self.current_layout = "mobile" if bool(os.getenv("TEST_MOBILE", "")) else "desktop"
         self.machine = machine
         path = os.path.dirname(__file__)
-        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace, fixed_content_size=self.pixels_label,
+        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
+                           mobile=(self.current_layout == "mobile"),
```
allisonkarlitskaya, line 140: How about defining this as a boolean `is_mobile` property on the layout scenarios instead of hardcoding it here?  We might imagine several mobile scenarios with difference sizes, for example.

test/common/testlib.py
```diff
@@ -122,14 +133,18 @@ def __init__(self, address, label, machine, pixels_label=None, port=None):
         self.default_user = "admin"
         self.label = label
         self.pixels_label = pixels_label
+        self.current_layout = "mobile" if bool(os.getenv("TEST_MOBILE", "")) else "desktop"
```
allisonkarlitskaya, line 136: I'd store the scenario dict here, rather than the name.  It'll make things easier everywhere else, particularly if you store the `is_mobile` flag in the dict (see below).

test/common/testlib.py
```diff
@@ -672,14 +675,14 @@ def open_session_menu(self):
...
-        if self.cdp.mobile:
+        if self.current_layout == "mobile":
             self.open_session_menu()
             self.click("#super-user-indicator-mobile button")
         else:
             self.click("#super-user-indicator button")
 
     def check_superuser_indicator(self, expected):
-        if self.cdp.mobile:
+        if self.current_layout == "mobile":
```
allisonkarlitskaya, line 685: ...then, for example

```suggestion
        if self.current_layout.is_mobile:
```

test/run
```diff
@@ -27,6 +26,9 @@ case $TEST_SCENARIO in
         test/image-prepare --verbose --overlay $TEST_OS
         test/common/run-tests --jobs ${TEST_JOBS:-1} --test-dir test/verify
         ;;
+    mobile)
+        echo "Deprecated test scenario: $TEST_SCENARIO"
```
allisonkarlitskaya, line 30: may as well `exit 1` here.  Nobody has this status set as required.

test/verify/check-pages
```diff
@@ -462,7 +453,12 @@ OnCalendar=daily
         b.wait_visible("#host-apps li a:contains('Services')")
         b.wait_text("#host-apps li a:contains('Services')", "ServicesContains: systemd")
         b.wait_text("#host-apps li a:contains('Services') mark", "systemd")
-        b.assert_pixels("#nav-system", "menu-search")
+        b.assert_pixels("#nav-system", "menu-search", ignore=[".nav-item-status"], skip_layouts=["mobile"])
+        b.set_layout("mobile")
```
allisonkarlitskaya, line 457: huh.  This is not what I expected.

I thought the idea was to run all the asserts in all the layouts, automatically, instead of having each test case drive this for itself...

This probably needs some discussion.

allisonkarlitskaya (changes requested):
Some minor suggestions, but also one outstanding big question.

test/common/testlib.py
mvollmer: Hmm, for this specific line, I think I would actually go back to the original code, which was setting `cdp.mobile` based on the TEST_MOBILE environment variable. `cdp.mobile` and TEST_MOBILE are now deprecated and I want to get rid of them eventually, once c-machines etc stop using them.

So keeping the old code would make most sense, I'd say.

But in general, I say it's a good idea to have more attributes in the layout definitions. If we move the definitions out of common/, then each project can put whatever stuff they want in there and check them in the tests.

But `is_mobile` will be a standard attribute, since the shell helpers in teslib.py itself need to know what layout the shell is in.

test/common/testlib.py
mvollmer: Yes, good idea!

test/run
mvollmer: Yeah, this is just here to make the transition slightly more correct. If we remove fedora-35/mobile from testmap immediately when (or before) merging this, we probably don't need to be that correct.

test/verify/check-pages
mvollmer: This is just an exception. Normally all calls to assert_pixels run through all the layouts, but in this case the mobile layout needs some special treatment (clicking on `#nav-system-item`).  Note the `skip_layouts=["mobile"]` in the call to assert_pixels.

test/common/testlib.py
mvollmer: > `cdp.mobile` and TEST_MOBILE are now deprecated and I want to get rid of them eventually, once c-machines etc stop using them.

Ah, let's just remove them altogether.  We don't need to keep testlib.py backwards compatible...

test/common/testlib.py
mvollmer: Done, it's in `test/browser-layouts`.

test/verify/check-pages
allisonkarlitskaya: Gotcha.  Thanks.

allisonkarlitskaya (approved):
No more comments from me.  It all seems pretty reasonable now, thanks :)

Please make sure to file a PR to drop the mobile scenario from the testmap before merging this.

As we discussed: this has the unfortunate side-effect of increasing the length of time we run for the "default" OS, which is also the one which we require to pass for merging even trivial changes.  Do we have any concrete plans going forward for splitting the pixel testing into a separate scenario which only runs tests which have `assert_pixels()` calls?

allisonkarlitskaya (changes requested):
oops.  hihi vulture:

```
# /tmp/source/test/common/cdp.py:59: unused variable 'window_height' (100% confidence)
# /tmp/source/test/common/cdp.py:59: unused variable 'window_width' (100% confidence)
```

mvollmer:
`TestServices-testBasic-text-filter-test` has trouble, the tab headers are not in a consistent place.  There is code in `main` to take care of that, I guess we have to put it back here.

mvollmer:
`TestTerminal.testBasic` should be taken care of by #16877.

test/common/testlib.py
```diff
@@ -93,6 +93,21 @@
 opts.jobs = 1
 opts.fetch = True
 
+default_layouts = [
```
martinpitt, line 96: Let's document this format a bit, and also mention test/browser-layouts.json here.

test/common/testlib.py
```diff
@@ -124,12 +139,20 @@ def __init__(self, address, label, machine, pixels_label=None, port=None):
...
         path = os.path.dirname(__file__)
-        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace, fixed_content_size=self.pixels_label,
+        self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
                            inject_helpers=[os.path.join(path, "test-functions.js"), os.path.join(path, "sizzle.js")])
         self.password = "foobar"
         self.timeout_factor = int(os.getenv("TEST_TIMEOUT_FACTOR", "1"))
         self.failed_pixel_tests = 0
         self.body_clip = None
+        try:
+            with open(f'{TEST_DIR}/browser-layouts') as fp:
```
martinpitt, line 149: Nitpickk: can this be named `browser-layouts.json`?

test/common/testlib.py
```diff
@@ -93,6 +93,21 @@
 opts.jobs = 1
 opts.fetch = True
 
+default_layouts = [
+    {
+        "name": "desktop",
+        "is_mobile": False,
```
martinpitt, line 99: This just describes whether cockpit's shell switches to desktop or mobile mode in the given resolution, right? So this is not an independent variable, but depends on the window width. Is this too complicated to determine automatically? Please add a comment.

But see below, I'm fairly confident that we can lose this.

test/common/testlib.py
```diff
@@ -93,6 +93,21 @@
...
+        "shell_size": [1920, 1200],
+        "content_size": [1680, 1130]
```
martinpitt, line 101: Again, these two don't seem independent from each other -- the delta should be pretty much fixed, and depend on our shell/menu CSS. If we can't determine these automatically (e.g. by waiting for overview to render and then querying the content size), let's at least add a comment how to compute these.

test/common/testlib.py
```diff
@@ -93,6 +93,21 @@
...
+    },
+    {
+        "name": "mobile",
+        "is_mobile": True,
+        "shell_size": [414, 1920],
+        "content_size": [414, 1856]
```
martinpitt, line 107: Here it's even more obvious -- for such narrow widths we expect both widths to be equal, there is no freedom here.

test/common/testlib.py
```diff
@@ -672,14 +683,14 @@ def open_session_menu(self):
             self.click("#toggle-menu")
 
     def open_superuser_dialog(self):
-        if self.cdp.mobile:
+        if self.current_layout["is_mobile"]:
```
martinpitt, line 686: There is only three places which neeed `is_mobile` now. Could we replace this with a `self.is_mobile()` function which checks the DOM and returns the current mode?

test/common/testlib.py
```diff
@@ -717,7 +728,7 @@ def click_system_menu(self, path, enter=True):
         (useful for remote hosts).
         '''
         self.switch_to_top()
-        if self.cdp.mobile:
+        if self.current_layout["is_mobile"] == "mobile":
```
martinpitt, line 731: Typo, this is a bool.

test/common/testlib.py
```diff
@@ -744,6 +755,7 @@ def snapshot(self, title, label=None):
             title: Used for the filename.
         """
         if self.cdp and self.cdp.valid:
+            self._adjust_window_for_fixed_content_size()
```
martinpitt, line 758: This seems strange, and like it would destroy evidence. We want the screenshot of *exactly* the thing that the test fails on, not mutate it.

test/common/testlib.py
```diff
@@ -767,12 +779,42 @@ def snapshot(self, title, label=None):
...
 
-    def assert_pixels(self, selector, key, ignore=[]):
-        """Compare the given element with its reference"""
+    def _set_window_size(self, width, height):
+        self.cdp.invoke("Emulation.setDeviceMetricsOverride",
+                        width=width, height=height,
+                        deviceScaleFactor=0, mobile=False)
+
+    def set_layout(self, name):
+        layout = next(filter(lambda l: l["name"] == name, self.layouts))
```
martinpitt, line 788: The next() here looks really strange.. It's python syntax for getting the first element of an iterator, but it *looks* like an off-by-one.

Perhaps this is a bit clearer:

    [layout for layout in self.layouts if layout["name"] == name][0]

The failure mode is a bit different -- your's would just throw a `StopIteration`, mine would throw an `IndexError`, so in either case this should be caught and replaced with something more readable. Or you drop the `[0]` and check for an empty list.

But what you should *really* be doing is to convert this array to a dict, indexed by `name` :grin:

test/common/testlib.py
```diff
@@ -767,12 +779,42 @@ def snapshot(self, title, label=None):
...
+        if layout and layout != self.current_layout:
```
martinpitt, line 789: `if layout` is always true, as the lookup throws an exception. (But probably moot -- let's just use a proper dict)

test/common/testlib.py
```diff
@@ -767,12 +779,42 @@ def snapshot(self, title, label=None):
...
+            self.current_layout = layout
+            size = layout["shell_size"]
+            self._set_window_size(size[0], size[1])
+            self._adjust_window_for_fixed_content_size()
+
+    def _adjust_window_for_fixed_content_size(self):
+        if self.eval_js("window.name").startswith("cockpit1:"):
+            # Adjust thw window size further so that the content is
```
martinpitt, line 797: "thw"

test/common/testlib.py
```diff
@@ -899,6 +941,19 @@ def img_eq(ref, now, delta):
                 print("Differences in pixel test " + base)
                 self.failed_pixel_tests += 1
 
+    def assert_pixels(self, selector, key, ignore=[], skip_layouts=[]):
+        """Compare the given element with its reference in all layouts"""
+
+        if not (Image and self.pixels_label and self.cdp and self.cdp.valid):
```
martinpitt, line 947: I know it's not new code, we already have the same condition in assert_pixels_in_current_layout() -- but why the `cdp` and `cdp.valid`? That just seems to hide broken tests, where we try to do an assert_pixels() before initializing/logging inot the browser. This could lead to pixel tests which never run, silently ignoring them. Let's please error out in that case.

test/verify/check-networkmanager-vlan
```diff
@@ -45,9 +45,7 @@ class TestNetworkingVLAN(NetworkCase):
         # Remove focus ring around the close button, which causes pixel tests retries.
         b.blur("#network-vlan-settings-dialog > button[aria-label=Close]")
         b.wait_visible("#network-vlan-settings-interface-name-input")
-        # Ignore flaky pixel issues with the select triangle
-        ignoredClasses = ["#network-vlan-settings-parent-select"] if self.browser.cdp.mobile else []
-        b.assert_pixels("#network-vlan-settings-dialog", "networking-vlan-settings-dialog", ignore=ignoredClasses)
+        b.assert_pixels("#network-vlan-settings-dialog", "networking-vlan-settings-dialog")
```
martinpitt, line 48: This is a welcome change of course, obsolete due to waiting for animations now? But should be a separate commit.

test/verify/check-packages
```diff
@@ -61,12 +61,7 @@ class TestPackages(MachineCase):
         self.login_and_go("/playground/pkgs")
 
         b.switch_to_top()
-        if b.cdp.mobile:
-            b.click("#nav-system-item")
         b.wait_visible("#nav-system li:contains(Terminal)")
```
martinpitt, line 64: Note, I factorized this a bit in #16884, so these PRs will collide. Let's see who is first :grin:

test/run
```diff
@@ -11,10 +11,9 @@ TEST_SCENARIO=${TEST_SCENARIO:-verify}
 TEST_OS_DEFAULT=$(PYTHONPATH=bots python3 -c 'from machine.machine_core.constants import TEST_OS_DEFAULT; print(TEST_OS_DEFAULT)')
 
 case $TEST_SCENARIO in
-    verify|mobile|devel|firefox|firefox-devel)
+    verify|devel|firefox|firefox-devel)
```
martinpitt, line 14: This "Remove support for mobile scenario" commit changes the mobile pixel tests, they now [dropped the tooltips](https://github.com/cockpit-project/pixel-test-reference/compare/dba18567f57691d3e0e4db3f3c1d5789f7a7853c..5e461db94d7f188eb42885e9c8ee583ee64735d2#diff-0ed3412b32e0734685f9414341d2c7dcad89d5873eb12c241015c8e60a5dcace). I wonder how that happened? Does changing the window size somehow change the mouse position?

(It's fine, I am just curious)

martinpitt (changes requested):
Sorry for the delay in reviewing this!

I like the general approach, it's more flexible and greatly reduces the (overall) amount of CI resources that we need. I just have a bunch of smaller bugs/improvements.

Thanks!

test/common/testlib.py
martinpitt: There is a reasonable default now, so that projects don't *have* to define their own scenarios -- and I expect most dont', and we should not even recommend it. There is a reason for why we pick these scenarios, and they are not exactly trivial or arbitrary. I like the current approach with offering the possibility of overrides, but not making too much fuss about it.

test/common/testlib.py
mvollmer: Yes, that should hopefully answer your questions as well.

test/common/testlib.py
mvollmer: Sure.

test/common/testlib.py
mvollmer: > This just describes whether cockpit's shell switches to desktop or mobile mode in the given resolution, right?

Yes.  It's for the benefit of a few utility functions only.  But since tests now know at every point in which layout they are in, I think we don't need this at all anymore.  It's buggy right now, as you noticed, and the tests are fine.

So I'll remove `is_mobile`.

test/common/testlib.py
mvollmer: These sizes are a "contract" with the reference images.  We guarantee that the content iframe is this exact size whenever assert_pixels takes a screenshot, and thus the reference image for a full frame is this exact size.  We can not dynamically compute the content size, but we do dynamically compute the window size that is necessary to get this exact content size.

The content size here has been determined by looking at the current sizes of reference images for the whole content iframe.

test/common/testlib.py
mvollmer: I think we can remove `is_mobile` completely.  If necessary, it can be a optional argument to `open_superuser_dialog`.  Tests do know exactly in which layout they are.

test/common/testlib.py
mvollmer: Ouch, and it doesn't seem to make any difference... evidence that this condition never needs to be true.

test/common/testlib.py
mvollmer: Yes, good point. I'll remove it.

test/common/testlib.py
martinpitt: Ah, I see -- so shell tests would resize to shell_size, and page tests to content_size? That makes sense. (Let's document it, but we already have a conversation about it, so resolving)

test/common/testlib.py
mvollmer: It started out as a dict... :-)  I turned it into an array because:

- `self.current_layout` now points to the actual dict (instead of containing the name of a layout)
- which means the dict needs a "name" or "label" field for forming the filenames
- which would mean repeating the name: `{ "mobile": { "label": "mobile", ... } }`
- also, with an array, the initial layout can just be the first. With a dict, we would have to hard-code "desktop" as the default, or add some field like "is_default".

Personally, *shrug*  I was surprised that there is no `find` in Python and that `next` is considered canonical, but *shrug*

test/common/testlib.py
mvollmer: Good point.

test/common/testlib.py
mvollmer: Good point.  This comes originally from `Browser.snapshot`, which needs to be much more robust and not blow up when a test fails without a working browser.  `assert_pixel` can assume a working browser.

test/verify/check-networkmanager-vlan
mvollmer: Ok.

test/run
mvollmer: My guess is that the tooltip appears only when the dialog is opened in a too narrow size.  Now the tests open the dialof in "desktop" layout and switch to "mobile" layout while it is already open.

My hunch is that this is buggy behavior of Patternfly, one way or the other.

mvollmer:
fedora-35/mobile is expected to fail, rhel-9 seems unrelated.

allisonkarlitskaya (dismissed):
Still failing on `tests/static-code`.  *cough* commit and pre-push hooks *cough*

test/common/testlib.py
```diff
@@ -672,19 +704,10 @@ def open_session_menu(self):
             self.click("#toggle-menu")
 
     def open_superuser_dialog(self):
-        if self.cdp.mobile:
-            self.open_session_menu()
-            self.click("#super-user-indicator-mobile button")
```
martinpitt, line 677: This means that with this PR we completely lost testing the superuser indicator in mobile mode? I don't see any replacement for it.

test/common/testlib.py
```diff
@@ -717,8 +740,6 @@ def click_system_menu(self, path, enter=True):
         (useful for remote hosts).
         '''
         self.switch_to_top()
-        if self.cdp.mobile:
-            self.click("#nav-system-item")
```
martinpitt, line 721: This OTOH does have a replacement (although only two commits later, but that's fine).

martinpitt (changes requested):
Thanks for the fixes! Much better. Just needs the code style fix, and I have one question (which could be a follow-up).

mvollmer:
> Still failing on `tests/static-code`. _cough_ commit and pre-push hooks _cough_

_cough_ someone changed the name and broke my hooks _cough_ :-)

allisonkarlitskaya:
> _cough_ someone changed the name and broke my hooks _cough_ :-)

What?  You don't wake up every morning and open the `README.md` and `HACKING.md` files to see what's new? :sweat_smile:

test/common/testlib.py
mvollmer: True.  I add one.

mvollmer: Done.  OPening the dialog in "mobile" brings back the title tooltip. Oh well.

allisonkarlitskaya:
The failing pixels tests on fedora-35 look probably-relevant...

mvollmer:
(Hmm, for some reason ref

> The failing pixels tests on fedora-35 look probably-relevant...

Yes, there is code to make sure the tab headers are scrolled to the expected place, but it doesn't seem to be effective. I try harder to reproduce this locally.

martinpitt:
I sent https://github.com/cockpit-project/bots/pull/2867 to [fix rhel-9-0](https://logs.cockpit-project.org/logs/pull-16823-20220128-084727-487dcffa-rhel-9-0/log.html#47-2)

martinpitt (approved):
Argh sleep(), but it's well justified. Thanks! :clap:

allisonkarlitskaya (approved):

PR #16824, opened 2022-01-12
test: Remove support for the */mobile test scenario 

mvollmer:
Instead, switch to the "mobile" layout explicitly in the approriate
places in the tests.

This only affects cockpit itself, not the support for TEST_MOBILE in
test/common/.  Cockpit-machines can still run their "mobile" scenarios
as before.

- [ ] #16823 
- [ ] Some more "mobile" tests for the "smart table actions" in storage, at least


mvollmer:
Merged into #16823.

PR #16825, opened 2022-01-12
selinux: Allow cockpit-session to access LDAP port

martinpitt:
This is now required on RHEL 8.6 for authenticating identity domain
managed users:

    avc:  denied  { name_connect } for comm="cockpit-session"
      dest=389 scontext=system_u:system_r:cockpit_session_t:s0
      tcontext=system_u:object_r:ldap_port_t:s0 tclass=tcp_socket

----

Will fix two of the issues in https://github.com/cockpit-project/bots/pull/2794

 - [ ] Decide if this is actually desired: https://bugzilla.redhat.com/show_bug.cgi?id=2039892

### DIFF
--- a/selinux/cockpit.te
+++ b/selinux/cockpit.te
@@ -155,13 +155,15 @@ kernel_read_network_state(cockpit_session_t)
 
 # cockpit-session runs a full pam stack, including pam_selinux.so
 auth_login_pgm_domain(cockpit_session_t)
-# cockpit-session resseting expired passwords
+# cockpit-session resetting expired passwords
 auth_manage_passwd(cockpit_session_t)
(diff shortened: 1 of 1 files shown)


martinpitt:
I validated this locally against the new rhel-8-6 image. But I don't want to trigger it here, as there are more issues, and I don't want to land a red PR.

martinpitt:
Please hold off on this -- it's a little dubious, and might just be more fallout from https://bugzilla.redhat.com/show_bug.cgi?id=2039892

martinpitt:
Let's not do this, confirmed fallout from the sssd update. https://bugzilla.redhat.com/show_bug.cgi?id=2039892#c5

PR #16826, opened 2022-01-12
cleanup tools/test-static-code

allisonkarlitskaya:
...and fix a bunch of Python flakes that we were missing before.

### DIFF
--- a/tools/test-static-code
+++ b/tools/test-static-code
@@ -1,62 +1,41 @@
 #!/bin/bash
 # run static code checks like pyflakes and pycodestyle
+
 set -eu
 
-echo "1..9"
-
-cd "${srcdir:-.}"
-fail=0
+# we consider any function named test_* to be a test case
+# each test is considered to succeed if it exits with no output
+# exit with status 77 is a skip, with the message in the output
+# otherwise, any output is a failure, even if exit status is 0
 
 #
 # pyflakes
 #
-PYFLAKES="python3 -m pyflakes"
-
-if ! $PYFLAKES /dev/null >/dev/null 2>&1; then
-    echo "ok 1 pyflakes pkg tools # SKIP pyflakes3 not installed"
-
-elif $PYFLAKES tools/build-debian-copyright tools/npm-release-time >&2; then
-    echo "ok 1 pyflakes pkg tools"
-
-else
-    echo "not ok 1 pyflakes pkg tools"
-    fail=1
-fi
+test_pyflakes_pkg_tools() {
+    python3 -m pyflakes /dev/null || skip
+    python3 -m pyflakes tools/build-debian-copyright tools/npm-release-time
+}
 
 # TODO: there are currently a lot of pyflakes errors like
 #   'parent' imported but unused
 #   'from testlib import *' used; unable to detect undefined names
 # Filter these out until these get fixed properly.
-if ! $PYFLAKES /dev/null >/dev/null 2>&1; then
-    echo "ok 2 pyflakes test # SKIP pyflakes3 not installed"
-
-else
-    out=$($PYFLAKES test/ 2>&1 | grep -Ev "(unable to detect undefined names|defined from star imports|'parent' imported but unused)") || true
-    if [ -n "$out" ]; then
-        echo "$out" >&2
-        echo "not ok 2 pyflakes test"
-        fail=1
-    else
-        echo "ok 2 pyflakes test"
-    fi
-fi
+test_pyflakes_test() {
+    python3 -m pyflakes /dev/null || skip
+    filter="(unable to detect undefined names|defined from star imports|'parent' imported but unused)"
+    ! python3 -m pyflakes test/ 2>&1 | grep -Ev "${filter}"
+}
 
 #
 # wrongly marked translatable strings
 #
-if out=$(find src/ pkg/ -name '*.js' -o -name '*.jsx' | xargs grep -n -E "(gettext|_)\(['\`]"); then
-    echo 'ERROR: translatable strings must be marked with _("")' >&2
-    echo "$out" >&2
-    echo "not ok 3 js-translatable-strings"
-    fail=1
-elif out=$(grep -r 'translatable=["'\'']yes' pkg/ doc/); then
-    echo "ERROR: 'translatable' HTML attribute is invalid, use 'translate' instead"
-    echo "$out" >&2
-    echo "not ok 3 js-translatable-strings"
-    fail=1
-else
-    echo "ok 3 js-translatable-strings"
-fi
+test_js_translatable_strings() {
+    ! git grep -n -E "(gettext|_)\(['\`]" -- {src,pkg}/'*'.{js,jsx}
+}
+
+test_no_translatable_attr() {
+    ! git grep -n 'translatable=["'\'']yes' -- pkg doc
+}
 
 #
 # Unsafe content-security-policy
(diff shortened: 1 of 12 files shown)


tools/pre-push
```diff
@@ -0,0 +1,83 @@
...
+check_commit() {
+    tmpdir="tmp/pre-push/${commit}"
+    rm -rf "tmp/pre-push/${commit}"
+
+    # unpack all of the changed files, plus the test-static-code from that commit
+    files="$(git show --format='' --name-only "${commit}" --) tools/test-static-code"
+    git archive --prefix="${tmpdir}/" "${commit}" -- ${files} | tar x
+
+    # check that node_modules always gets updated with package.json
+    if [ -e "${tmpdir}/package.json" -a ! -e "${tmpdir}/node_modules" ]; then
```
martinpitt, line 27: I checked out your branch, in a local `tss` branch of mine, adjusted the hook:

    lrwxrwxrwx. 1 martin martin   20 Jan 13 08:59 pre-push -> ../../tools/pre-push

and tried to push to my fork with `git push my`.
```
❱❱❱ git push my 
Performing pre-push checks for refs/heads/tss:
 - a01793db3 tools/pre-push: introduce a proper git pre-push hook: pass
 - 2b91fa724 test-static-code: clean up python checks:
    not ok 5 /static-code/test-node-modules-freshness
    # tools/test-static-code: line 85: tools/node-modules: No such file or directory
 - 8736283a8 storage: clean up vdo python imports:
    not ok 5 /static-code/test-node-modules-freshness
    # tools/test-static-code: line 77: tools/node-modules: No such file or directory
 - 6850624cd check-bastion test: don't 'import *':
    not ok 5 /static-code/test-node-modules-freshness
    # tools/test-static-code: line 77: tools/node-modules: No such file or directory
 - e87ea3ef2 tools/escape-to-c: drop a superfluous f-string:
    not ok 5 /static-code/test-node-modules-freshness
    # tools/test-static-code: line 77: tools/node-modules: No such file or directory
 - 076521182 pkg/lib: add 'import sys' to inotify.py:
    not ok 5 /static-code/test-node-modules-freshness
    # tools/test-static-code: line 77: tools/node-modules: No such file or directory
 - cb3e3aac6 tools/test-static-code: major overhaul:
    not ok 5 /static-code/test-node-modules-freshness
    # tools/test-static-code: line 77: tools/node-modules: No such file or directory
error: failed to push some refs to 'github.com:martinpitt/cockpit.git'
```

`tools/node_modules` obviously exists in my tree, but it didn't get copied into tmp/pre-push/sha/. The top-level commit passed, so this might only work for future once we land this. But of course it'd still break testing of stable branches.

Honestly I don't like this copying/temp git tree magic very much. It is brittle, hard to debug, and imposes huge assumptions on test-static-code: it has to work on any random combinations of changed files that we imagine, plus it doesn't check the corresponding dist/ anyway. It also seems to work against your goal of making the hook faster. The goal of checking every commit *is* nice, of course. So not a veto, but if you'd cut that back I wouldn't complain at all :grin:

tools/pre-push
```diff
@@ -0,0 +1,83 @@
...
+
+        if [ "${local_object_name}" = "${nothing}" ]; then
+            # deleting remote ref: no checks
+            continue
+        fi
+
+        if [ "${remote_object_name}" != "${nothing}" ]; then
+            reference="${remote_object_name}"
+        else
+            reference="origin/main"
```
martinpitt, line 62: Having to hardcode a branch name here feels wrong. If git can't figure it out by itself somehow, it's almost certainly going to be broken for stable branches like origin/rhel-8.5

martinpitt (changes requested):
Thanks! I generally like where this is going, sans the too magic pre-push hook.

tools/test-static-code
```diff
@@ -9,21 +9,36 @@ set -eu
...
+python_files() {
+    (
+        git grep --cached -lIz '^#!.*python3'
+        git ls-files -z '*.py'
+    ) | sort -z | uniq -z
+}
+
+# files not in test/verify/
+test_pyflakes() {
+    command -v pyflakes >/dev/null || skip 'no pyflakes'
```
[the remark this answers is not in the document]
allisonkarlitskaya: My tests show it to be about 3x faster on average, so let's take this as a compromise.

tools/pre-push
allisonkarlitskaya: > tools/node_modules obviously exists in my tree, but it didn't get copied into tmp/pre-push/sha/. The top-level commit passed, so this might only work for future once we land this. But of course it'd still break testing of stable branches.

Most of the changes in this branch are specifically about preparing `tools/test-static-code` to work with this new model.  It makes quite a lot of assumptions (simple stuff like "there will be at least one python file to test") which are broken under this model, and that brokenness is demonstrated by the failing commits here.

> Honestly I don't like this copying/temp git tree magic very much.

So the rationale is something like this:
 - every single commit should have passing tests, as far as we can reasonably manage (think: our future selves bisecting stuff)
 - doing this all out of the git object store is problematic for a lot of reasons.  Specifically: the tools we want to run expect to inspect files on the filesystem.  Also, I've checked, and in a lot of cases, the git version of things is actually slower than the filesystem version, presumably related to decompression and lack of caching.  We want to run this on the filesystem.
 - considering that we may be pushing something that's not the currently checked out branch, or might be pushing a branch when we have changes in the working tree, there's actually absolutely no guarantee that the thing we want to test exists anywhere on the filesystem at all.
 - the only solution i can think of is that we need to check out the thing we actually want to test.  That's the entire basis of this approach, and I think it's pretty reasonable.

tools/pre-push
```diff
@@ -0,0 +1,83 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
allisonkarlitskaya: OK.

tools/test-static-code
```diff
@@ -1,161 +1,133 @@
...
-    echo "ok 6 json-verify # SKIP not a git tree"
-fi
+}
 
 # pycodestyle python syntax check
-if python3 -m pycodestyle /dev/null >/dev/null 2>&1; then
-    PYTHON_STYLE_CHECKER="pycodestyle"
-fi
+test_pycodestyle() {
+    python3 -m pycodestyle /dev/null || exit 77
```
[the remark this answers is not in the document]
allisonkarlitskaya: sure.  That's what it's there for.

tools/test-static-code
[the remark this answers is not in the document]
allisonkarlitskaya: I had another idea for an approach here: just run the thing, and check the output if it matches some kind of a pattern for complaining about "No module named x".  That's a bit magic, though, so let's set it aside for now.

tools/pre-push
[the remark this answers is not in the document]
allisonkarlitskaya: I assume you mean `git` and not `github`.  Will adjust accordingly.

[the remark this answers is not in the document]
martinpitt: Err yes, of course *blush*

tools/pre-push
martinpitt: Right, as I said I agree to the goal, I just don't like the complexity of it. (Kinda similar to the jumpstart discussion). But as I said it's not a veto.

tools/pre-push
martinpitt: Another idea: Once we submit a PR, then the origin target is known. Could this "check all commits" move into a workflow instead, and as the pre-push hook we just keep the "check the current tree" behaviour that we have now? It's not like the pre-push hook is going to find all problems with your PRs anyway, it's just a tool to tell you the most glaring problems.

allisonkarlitskaya: This doesn't help much, though: the main benefit here is that we only check changed files, but without being able to determine what the "starting point is" we also can't determine the list of files changed by the current branch...

I remembered that `git merge-base` has a `--fork-point` option which somehow does some magic using the reflog to try to guess where you "started work".  I'll take a look at that next...

Otherwise, it would be really awesome if we had some way to be able to tell commits that were created by GitHub (during "rebase and merge") and commits that were created by human users of `git`.  I don't know any, though.

martinpitt: It helps with being thorough (every commit passes individually). It of course does not help with testing only changed files locally, but then again, if you push your PRs so often that 7.5s matter, we have bigger problems with our CI capacity :wink: I can't say I'm very thrilled about starting guesswork and heuristics here..

martinpitt: origin == cockpit-project/cockpit works for me, and might work for you, but ISTR that @mvollmer has a different workflow where origin == his fork. I think that does *not* invalidate your reasoning, as these pushes would also have been pre-verified. I just wanted to point it out here. It's not that important, but the one thing we need to avoid is to erroneously check hundreds of commits.

martinpitt (approved):
Thanks! It's rather hard to actually test the pre-push hook now, as it doesn't work outside of git, and doesn't work on current origin. But it looks okay now, and it's fungible without endangering anything.

PR #16827, opened 2022-01-13
test: Add 1280x768 "medium" layout for pixel tests

mvollmer:
- [x] #16823 
- [x] Add all the new reference images

New references: https://github.com/cockpit-project/pixel-test-reference/compare/a292f9e2799749fe0b99bd662099611d019997f2..bed9ed36da1a31965361ad19c110cb2aef50d95f

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -124,6 +124,12 @@ default_layouts = [
         "content_size": [1680, 1130]
     },
     {
+        "name": "medium",
+        "is_mobile": False,
+        "shell_size": [1280, 768],
+        "content_size": [1040, 698]
+    },
+    {
         "name": "mobile",
         "shell_size": [414, 1920],
         "content_size": [414, 1856]
@@ -831,14 +837,14 @@ class Browser:
             if delta[0] != 0 or delta[1] != 0:
                 self._set_window_size(shell_size[0] + delta[0], shell_size[1] + delta[1])
 
-    def assert_pixels_in_current_layout(self, selector, key, ignore=[]):
+    def assert_pixels_in_current_layout(self, selector, key, ignore=[], scroll_into_view=None):
         """Compare the given element with its reference in the current layout"""
 
         if not (Image and self.pixels_label):
             return
 
         self._adjust_window_for_fixed_content_size()
-        self.call_js_func('ph_scrollIntoViewIfNeeded', selector)
+        self.call_js_func('ph_scrollIntoViewIfNeeded', scroll_into_view or selector)
         self.call_js_func('ph_blur_active')
 
         # Wait for all animations to be over.  This is done by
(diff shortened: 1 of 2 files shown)


martinpitt:
@mvollmer : This is still missing the login page that landed this morning. Do we have to have each test in all three scenarios? It rather feels like we should, and fail if we miss one. We could say that we don't care about some scenarios for some tests, but that makes it a bit prone to merely forgetting. What's the plan?

martinpitt:
@mvollmer: Nevermind, the link in the description was out of date. I updated it. LGTM, many thanks!

martinpitt (changes requested):
[TestStorageUsed-testUsed-popover-medium-pixels.png](https://github.com/cockpit-project/pixel-test-reference/compare/780cc3c8055a77321f175cc8c86cb6fddb35ac09..f6083c6c2ccf7b75b7c4cde00d80b08991d24fa1#diff-ac1bbd6c8b19c3b24f289c6b0a63ad8e1c2688cb7b1f4bd4bb8e771cc8b15c89) is the only one which looks *really* odd. That screenshot looks fine in other variants, e.g. [desktop variant](https://github.com/cockpit-project/pixel-test-reference/blob/f6083c6c2ccf7b75b7c4cde00d80b08991d24fa1/TestStorageUsed-testUsed-popover-pixels.png) (mobile looks almost the same).

The other medium screenshots look good.

mvollmer:
> [TestStorageUsed-testUsed-popover-medium-pixels.png](https://github.com/cockpit-project/pixel-test-reference/compare/780cc3c8055a77321f175cc8c86cb6fddb35ac09..f6083c6c2ccf7b75b7c4cde00d80b08991d24fa1#diff-ac1bbd6c8b19c3b24f289c6b0a63ad8e1c2688cb7b1f4bd4bb8e771cc8b15c89) is the only one which looks _really_ odd.

Weird.  Thanks for catching this, I'd have missed it.

mvollmer:
> > [TestStorageUsed-testUsed-popover-medium-pixels.png](https://github.com/cockpit-project/pixel-test-reference/compare/780cc3c8055a77321f175cc8c86cb6fddb35ac09..f6083c6c2ccf7b75b7c4cde00d80b08991d24fa1#diff-ac1bbd6c8b19c3b24f289c6b0a63ad8e1c2688cb7b1f4bd4bb8e771cc8b15c89) is the only one which looks _really_ odd.
> 
> Weird. Thanks for catching this, I'd have missed it.

Interesting.  The body of the dialog is too tall and the element that triggers the popover is out of view. In that case, the popover is still in the DOM at the expected position and would be in view, but is kept invisible because its trigger element is out of view.

So we need to scroll the dialog body up, or something like that.

mvollmer:
> So we need to scroll the dialog body up, or something like that.

I have made it so that the trigger element is scrolled into view by assert_pixels, instead of the popup itself.  Let's trust that the popup is fully visible when its trigger is.  It is placed intelligently, after all.

mvollmer:
(Argh, forgot to remove the now unused terminal reference image. Sorry bots, one more go.)

mvollmer:
Fixed the popover thing, updated link to new reference images in description.

New reference image for the popover:

https://github.com/cockpit-project/pixel-test-reference/compare/a292f9e2799749fe0b99bd662099611d019997f2..bed9ed36da1a31965361ad19c110cb2aef50d95f#diff-ac1bbd6c8b19c3b24f289c6b0a63ad8e1c2688cb7b1f4bd4bb8e771cc8b15c89

martinpitt (approved):
Danke! \o/

PR #16830, opened 2022-01-13
packagekit: Rename "Kernel patch" to "Kernel live patch"

martinpitt:
This describes more precisely what this is about, and makes our UI
consistent with the RHEL documentation:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_monitoring_and_updating_the_kernel/applying-patches-with-kernel-live-patching_managing-monitoring-and-updating-the-kernel


martinpitt:
Nothing surprising, but here we are:

image

marusak (approved):
Danke!

PR #16831, opened 2022-01-13
packagekit: Show a spinner while checking for kpatch support on RHEL

martinpitt:
Testing for kpatch support can take quite long (UX review reported 8
seconds), and it is apparently confusing/not discoverable for users to
wait that long until they see the status of kpatch.

During detection of kpatch support on RHEL (the only OS where we
actually expect this to succeed), show a spinner as status. Also show a
disabled "Edit" button to avoid the layout jumping around once it gets
enabled.

Don't show the spinner on other OSes, as we expect it to fail there, and 
it is both confusing and also changing the page layout when the "Kernel
live patching" entry entirely disappears from the Settings card.

------
 - Depends on PR #16833

On RHEL while loading:
image

On RHEL when loading is done (no change compared to main):

image

On RHEL when kpatch package is not available (done synthetically, this is rather unlikely):

image
image

On Fedora nothing ever gets displayed.

### DIFF
--- a/pkg/packagekit/kpatch.jsx
+++ b/pkg/packagekit/kpatch.jsx
@@ -23,16 +23,19 @@ import {
     Alert, Button, Checkbox,
     Flex, FlexItem,
     Modal, Radio,
+    Popover,
+    Spinner,
     Split, SplitItem,
     Stack,
     Text, TextVariants
 } from "@patternfly/react-core";
-import { InfoIcon } from "@patternfly/react-icons";
+import { InfoIcon, InfoCircleIcon } from "@patternfly/react-icons";
 
 import cockpit from "cockpit";
 import { proxy as serviceProxy } from "service";
 import { check_missing_packages } from "packagekit.js";
 import { install_dialog } from "cockpit-components-install-dialog.jsx";
+import { read_os_release } from "os-release.js";
 
 const _ = cockpit.gettext;
 
@@ -42,6 +45,7 @@ export class KpatchSettings extends React.Component {
 
         this.state = {
             loaded: false,
+            showLoading: null, // true: show spinner during initialization; false: hide
             auto: null, // `dnf kpatch` is set to `auto`
             enabled: null, // kpatch.service is enabled
             missing: [], // missing packages from `kpatch`, `kpatch-dnf`
@@ -66,6 +70,9 @@ export class KpatchSettings extends React.Component {
         this.handleChange = this.handleChange.bind(this);
         this.onClose = this.onClose.bind(this);
         this.handleInstall = this.handleInstall.bind(this);
+
+        // only show a spinner during loading on RHEL (the only place where we expect this to work)
+        read_os_release().then(os_release => this.setState({ showLoading: os_release.ID == 'rhel' }));
     }
 
     // Only current patches or also future ones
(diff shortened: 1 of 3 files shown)


martinpitt:
I am not entirely happy with this yet. This now has the inverse problem, that the status goes away entirely after the check, so once again the page jumps around. But this time it hits Fedora, Debian, etc. I think I'd like to add a really fast check at the beginning that can see if kpatch is *not* supported on this platform -- like, `ID!="rhel"` in /etc/os-release?

@marusak, any opinion/suggestion?

martinpitt:
I reworked this now using os-release (and thus blocking on #16833). It feels too complicated, but I don't have a good idea how to simplify it

marusak:
Yeah, always showing spinner is no-go, showing it on rhel is fine. But if we show spinner, we should not go from spinner to nothing. I guess we should then show something like:
```
Kernel live patching                      Not supported(?)
```
and `(?)` would have tooltip explaining which packages are unavailable.

martinpitt:
It's not actually that easy to make it fail on RHEL, but with this hack I can simulate it:

```diff
--- pkg/packagekit/kpatch.jsx
+++ pkg/packagekit/kpatch.jsx
@@ -87,7 +87,7 @@ export class KpatchSettings extends React.Component {
                     this.checkSetup().then(() =>
                         this.setState({
                             loaded: true,
-                            unavailable: d.unavailable_names || [],
+                            unavailable: /* d.unavailable_names || [] */ ["kpatch-dnf"],
                             missing: d.missing_names || [],
                         })
```

martinpitt:
It's not actually that easy to make it fail on RHEL, but with this hack I can simulate it:

```diff
--- pkg/packagekit/kpatch.jsx
+++ pkg/packagekit/kpatch.jsx
@@ -87,7 +87,7 @@ export class KpatchSettings extends React.Component {
                     this.checkSetup().then(() =>
                         this.setState({
                             loaded: true,
-                            unavailable: d.unavailable_names || [],
+                            unavailable: /* d.unavailable_names || [] */ ["kpatch-dnf"],
                             missing: d.missing_names || [],
                         })
```

martinpitt:
@marusak: The flakes here are due to 3x affected from the blocking PR, and will go away once #16833 lands. WDYT about the current design now? I updated the screenshots in the description.

pkg/packagekit/kpatch.jsx
```diff
@@ -211,25 +218,37 @@ export class KpatchSettings extends React.Component {
...
+        let state;
         let actionText = _("Edit");
         let action = () => this.setState({ showModal: true });
 
-        if (this.state.missing.length > 0) {
+        if (this.state.loaded === false || this.state.patchName === null) {
+            // Not yet recognized
+            state = <Spinner isSVG size="md" />;
+        } else if (this.state.unavailable.length > 0) {
+            state = <Popover headerContent={ _("Missing packages") } bodyContent={ this.state.unavailable.join(", ") }>
```
marusak, line 236: Not sure if I would say "missing" as that suggest they are not installed. If they are missing, we ofter installing them, they are "unavailable".

marusak:
> WDYT about the current design now?

I like the current design and we can go with this.

pkg/packagekit/kpatch.jsx
martinpitt: Good point -- I'll adjust this after the os-release dependency PR lands.

martinpitt: Changed to "Unavailable packages".

martinpitt:
The depending PR landed. Fixed the wording, rebased.

marusak (approved):
Nice, thank you!

PR #16833, opened 2022-01-14
Fix /etc/os-release handling

martinpitt:
The most obvious confusion is with remote logins:

image

It's more subtle with the Applications page, the refresh would simply fail if you e.g. logging into a remote Debian machine from a Fedora host.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -345,10 +345,6 @@
             el.focus();
         });
 
-        const os_release = environment["os-release"];
-        if (os_release)
-            localStorage.setItem('os-release', JSON.stringify(os_release));
-
         const logout_intent = window.sessionStorage.getItem("logout-intent") == "explicit";
         if (logout_intent)
             window.sessionStorage.removeItem("logout-intent");
--- a/pkg/storaged/client.js
+++ b/pkg/storaged/client.js
@@ -24,6 +24,7 @@ import { superuser } from 'superuser';
 import * as utils from './utils';
 
 import * as python from "python.js";
+import { read_os_release } from "os-release.js";
 import inotify_py from "raw-loader!inotify.py";
 import nfs_mounts_py from "raw-loader!./nfs-mounts.py";
 import vdo_monitor_py from "raw-loader!./vdo-monitor.py";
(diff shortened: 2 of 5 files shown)


martinpitt:
Oops, [this failure](https://logs.cockpit-project.org/logs/pull-16833-20220114-090009-8cde0735-fedora-coreos/log.html#239-2) is relevant.

pkg/storaged/client.js
```diff
@@ -593,15 +572,15 @@ function init_model(callback) {
...
+            client.os_release = os_release;
+
+            enable_features().then(() => {
+                query_fsys_info().then((fsys_info) => {
                     client.fsys_info = fsys_info;
 
-                    client.storaged_client.addEventListener('notify', function () {
-                        client.update();
-                    });
+                    client.storaged_client.addEventListener('notify', client.update);
```
mvollmer, line 583: This is not a valid transformation in general. `client.update` is now called with different arguments and a different `this`. It will work in this specific case, but it might break in the future and if applied elsewhere.  Let's not do this.  (Also, it has nothing to do with the topic of this PR, no?)

pkg/static/login.js
```diff
@@ -345,10 +345,6 @@
             el.focus();
         });
 
-        const os_release = environment["os-release"];
```
mvollmer, line 348: Removing this is the right idea, but it _might_ break external packages, no?

mvollmer (changes requested):

pkg/storaged/client.js
martinpitt: No, it's unrelated, I just didn't want to  mix arrows and old-style full-function calls, so took the opportunity. I put back the extra function wrapper now, it's not important.

pkg/static/login.js
martinpitt: Theoretically yes, but it was never documented and is hard to just find by accident. I checked all of our own projects, session-recording, subscription-manager, and all the other 3rd party apps on https://cockpit-project.org/applications.html , and nothing is using that.

martinpitt: Would you like me to drop this commit, or was this just a question how thoroughly I checked?

marusak (approved):
Thanks! Good catch with the docs mainly

martinpitt:
Landing now, to avoid bitrot. I'm fairly sure this is sane, I'm happy to put it back if necessary when @mvolmer returns and we can discuss this.

PR #16834, opened 2022-01-14
test: Don't allow unused pixel test references

mvollmer:
- [x] #16823

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -37,6 +37,7 @@ import time
 import unittest
 import gzip
 import inspect
+import glob
 
 import testvm
 import cdp
@@ -158,6 +159,7 @@ class Browser:
         self.default_user = "admin"
         self.label = label
         self.pixels_label = pixels_label
+        self.used_pixel_references = set()
         self.machine = machine
         path = os.path.dirname(__file__)
         self.cdp = cdp.CDP("C.utf8", verbose=opts.trace, trace=opts.trace,
@@ -881,6 +883,7 @@ class Browser:
             base += "-" + self.current_layout["name"]
         filename = base + "-pixels.png"
         ref_filename = os.path.join(reference_dir, filename)
+        self.used_pixel_references.add(ref_filename)
         ret = self.cdp.invoke("Page.captureScreenshot", clip=rect, no_trace=True)
         png_now = base64.standard_b64decode(ret["data"])
         png_ref = os.path.exists(ref_filename) and open(ref_filename, "rb").read()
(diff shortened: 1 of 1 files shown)


test/run
```diff
@@ -26,6 +26,9 @@ case $TEST_SCENARIO in
         test/image-prepare --verbose --overlay $TEST_OS
         test/common/run-tests --jobs ${TEST_JOBS:-1} --test-dir test/verify
         ;;
+    mobile)
+        echo "Deprecated test scenario: $TEST_SCENARIO"
```
martinpitt, line 30: We don't need this any more, it was just dropped from bots.

mvollmer: Oh, rebase mistake.

test/common/testlib.py
```diff
@@ -1550,8 +1565,10 @@ def check_browser_errors(self):
                 raise Error(UNEXPECTED_MESSAGE + "browser errors:\n" + log)
 
     def check_pixel_tests(self):
-        if self.browser and self.browser.failed_pixel_tests > 0:
-            raise Error(PIXEL_TEST_MESSAGE)
+        if self.browser:
+            self.browser.assert_no_unused_pixel_test_references()
+            if self.browser.failed_pixel_tests > 0:
+                raise Error(PIXEL_TEST_MESSAGE)
```
martinpitt, line 1571: Nitpick: If there are failures, let's fail on these, and only then complain about unused refs. That makes typos in file names more obvious?

martinpitt (changes requested):
Little thing only, but this is nice and simple, thanks!

martinpitt:
Also, after repushing, please trigger fedora-35 and the required fedora-coreos.

mvollmer:
Hmm, this should find some unused images...

mvollmer:
> Hmm, this should find some unused images...

Ah, I know!

Calling `assert_no_unused_blah` will only increase n_failed_pixel_tests, so my fix was wrong.

test/common/testlib.py
mvollmer: Calling `assert_no_unused_pixel_test_references()` will not raise an exception itself, it will just increase `failed_pixel_tests`, in the same way as `assert_pixels`.  So we will always see all failures, including unused references.

No need to change anything here, IMO.

martinpitt: I see, thanks!

martinpitt (approved):
Thanks! So that means we don't actually have any unused images right now?

mvollmer:
> Thanks! So that means we don't actually have any unused images right now?

No, we have some.  But I broke the logic, and tests would not actually fail if unused images were found.  Now it is fixed again: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16834-20220128-161618-830c7ea3-fedora-35/log.html#284

martinpitt (changes requested):
Ah good, thanks! So let's remove these, and then turn this to green. :-1: 'ing so that we don't accidentally hit the button (as fedora-35 is not a required test any more)

martinpitt:
Ugh, [another weird flake](https://logs.cockpit-project.org/logs/pull-16834-20220128-174232-48925c00-rhel-9-0/log.html#3), will have a look on Monday.

martinpitt (approved):
Danke!

PR #16837, opened 2022-01-17
flatpak: Wait for application to stop during test

martinpitt:
`test-ssh` send a `quit` command to Client, but did not previously wait
for the application to actually quit. Let's make sure this actually
works properly.

This also fixes a race condition with the flatpak-test workflow:
Starting test-browser immediately after test-ssh often failed with funny
errors like

    error: Environment variable must be given in the form VARIABLE=VALUE, not ��

as the test tried to talk to the dying instance which then went away.

-----

Failed test run [example 1](https://github.com/cockpit-project/cockpit/runs/4760838344?check_suite_focus=true), [example 2](https://github.com/cockpit-project/cockpit/runs/4805942961?check_suite_focus=true)

These always go along with

    error: Environment variable must be given in the form VARIABLE=VALUE, not �

or some other broken string, like `not j`. 

This is coming from flatpak, [here](https://github.com/flatpak/flatpak/blob/master/common/flatpak-context.c#L1208) or [here](https://github.com/flatpak/flatpak-xdg-utils/blob/master/src/flatpak-spawn.c#L730) and smells like a race condition in flatpak itself.

### DIFF
--- a/containers/flatpak/test/test-ssh
+++ b/containers/flatpak/test/test-ssh
@@ -54,4 +54,5 @@ gapplication action "${FLATPAK_ID}" open-path '"=cockpit-client-test"'
 until test -f cockpit-client-test-"${USER}"-cockpit-client-test-22; do sleep 0.5; done
 
 gapplication action "${FLATPAK_ID}" quit
+while flatpak ps | grep org.cockpit_project.CockpitClient; do sleep 1; done
 rm cockpit-client-test-*-*-*


martinpitt:
I tried this in a GitHub interactive ssh shell, and this succeeds:

    for i in `seq 50`; do xvfb-run dbus-run-session containers/flatpak/test/test-browser || break; sleep 1; done

(the sleep 1 is to  avoid some Xvfb shutdown/startup race, it's unrelated)

There is no permanently running "flatpak"-like service that I can see. User process are just the gh worker, bash, and tmate. `ps aux|grep -i flat` does not have anything at all. Also, the previous flatpak-install etc. should already have "warmed up the cache".

However, I *can* reproduce this with running the SSH test before:

    rm -rf ~/.ssh; xvfb-run dbus-run-session sh -exc 'containers/flatpak/test/test-ssh; containers/flatpak/test/test-browser'

After test-ssh, the Client flatpak application is still (barely) running and shutting down. test-browser then catches this and tries to to talk to the dying process.

PR #16838, opened 2022-01-17
test/static-code: add vulture to the list of checks

allisonkarlitskaya:
- add vulture to test/static-code
 - add it to the unit-tests container
 - fix up a bunch of issues that it found

### DIFF
--- a/containers/flatpak/test/test-browser
+++ b/containers/flatpak/test/test-browser
@@ -50,7 +50,7 @@ class BrowserTest(unittest.TestCase):
 
         # wait for window actions to populate
         actions = set(self.win_actions.list_actions())
-        handler = self.win_actions.connect('action-added', lambda group, name: actions.add(name))
+        handler = self.win_actions.connect('action-added', lambda _group, name: actions.add(name))
         main = GLib.MainContext.default()
         while not actions:
             main.iteration(True)
@@ -69,7 +69,7 @@ class BrowserTest(unittest.TestCase):
 
         result = None
 
-        def on_done(group, name, value):
+        def on_done(_group, name, value):
             # cockpit-client resets the value to '' initially, to guarantee that a Changed signal happens
             # even when two consecutive run-js calls have the same result
             v = value.get_string()
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1249,7 +1249,7 @@ class MachineCase(unittest.TestCase):
         shutil.rmtree(self.tmpdir, ignore_errors=True)
 
     def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False):
-        self.machine.start_cockpit(host, tls=tls)
+        self.machine.start_cockpit(tls=tls)
         self.browser.login_and_go(path, user=user, host=host, superuser=superuser, urlroot=urlroot, tls=tls)
 
     # List of allowed journal messages during tests; these need to match the *entire* message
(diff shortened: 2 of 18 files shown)


containers/flatpak/test/test-browser
```diff
@@ -50,7 +50,7 @@ class BrowserTest(unittest.TestCase):
 
         # wait for window actions to populate
         actions = set(self.win_actions.list_actions())
-        handler = self.win_actions.connect('action-added', lambda group, name: actions.add(name))
+        handler = self.win_actions.connect('action-added', lambda _group, name: actions.add(name))
```
martinpitt, line 53: Big :+1: on that -- this makes it much clearer for a developer.

test/common/testlib.py
```diff
@@ -113,7 +113,7 @@ def attach(filename, move=False):
 class Browser:
     def __init__(self, address, label, machine, pixels_label=None, port=None):
         if ":" in address:
-            (self.address, unused, self.port) = address.rpartition(":")
+            self.address, _, self.port = address.rpartition(":")
```
martinpitt, line 116: It feels a little silly to use the API that returns the separator, and then ignoring it. That's pretty much exactly what `rsplit(':', 1) is used for.

test/verify/check-shell-keys
```diff
@@ -39,7 +39,7 @@ class TestKeys(MachineCase):
 
         m.start_cockpit()
 
-        def login(user, password, user_name):
+        def login(user, password):
```
martinpitt, line 42: That commit message has "more shell keys", which probably refers to the other unused variables. Not a biggie, just nice to clean up if you have to force-push again for some other reason.

test/static-code
```diff
@@ -48,6 +48,16 @@ test_pycodestyle() {
     find_python_files | xargs -r -0 python3 -m pycodestyle --max-line-length=415 --ignore W504
 }
 
+test_pyvulture() {
+    # vulture to find unused variables/functions
+
+    python3 -c 'import vulture' 2>/dev/null || skip 'no python3-vulture'
+    find_python_files | xargs -r -0 python3 -m vulture \
+        --min-confidence "${VULTURE_CONFIDENCE:-100}" \
+        --ignore-names 'do_*,test[A-Z0-9]*,__*__' \
```
martinpitt, line 57: Are the `do_*()` methods still something to clean up after that PR, just too hard to do it right away? Or is there a reason why they can't be fixed?

martinpitt (approved):
Only trivial/follow-up stuff, thanks!

test/common/testlib.py
allisonkarlitskaya: I think that's a matter of taste.  I always use partition for this stuff...

test/verify/check-shell-keys
allisonkarlitskaya: Oops.  That's from squashing the fixup to the last round of test failures.  I'd prefer not to merge that, so I'll fix the message and try again.  I'm not in a hurry.

test/static-code
allisonkarlitskaya: Note that none of these ignore flags affect anything at the 100% confidence level: you need to turn it down to 60% confidence before you'd ever see something affected by that . As for the specific case of `do_`: this is a semi-convention for virtual functions involved by another framework.  This is in use by at least gobject-introspection (for the GApplication vfuncs implemented in CockpitClient) as well as by the `do_GET`, `do_POST`, etc. methods called by the Python http server code.

The convention is strong enough that vulture even lists it as an example:

```
  --ignore-names PATTERNS
                        Comma-separated list of names to ignore (e.g., "visit_*,do_*").
```

but doesn't enable it by default, unfortunately.

martinpitt (approved):

PR #16841, opened 2022-01-17
image-prepare --build-in-place

allisonkarlitskaya:
Add a `--build-in-place` option to `image-prepare`.

Uses the `--build-in-place` option of `tools/make-rpms` added in #16839.

 - [ ] #16839


allisonkarlitskaya:
Let's close this for now.  The direction we want to go with this stuff is more like #16871 + rsync.

PR #16842, opened 2022-01-18
.gitignore: ignore *.log and *.trs

allisonkarlitskaya:
These are generated by `make check`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@
 .deps/
 Makefile
 Makefile.in
+test-*.log
+test-*.trs
 
 # toplevel (/...)
 /.flatpak-builder


.gitignore
```diff
@@ -8,11 +8,13 @@
 *.a
 *.cache
 *.dirstamp
+*.log
```
martinpitt, line 11: These got dropped in the recent commit 293a928e62, but they were more specific:

    test-*.log
    test-*.trs

I think that is safer, just in case we ever get "precious" files (unlikely, but still..)

martinpitt (approved):
Not a strong opinion,  I'm ok with both.

.gitignore
allisonkarlitskaya: I agree, and I actually prefer the more specific pattern.  I'll change it back.

martinpitt (approved):

PR #16843, opened 2022-01-18
tools: add tool to create compile_commands.json

allisonkarlitskaya:
Add a new script, tools/make-compile-commands, which can generate the
compile_commands.json script used by many static analysis tools.

### DIFF
--- a/tools/make-compile-commands
+++ b/tools/make-compile-commands
@@ -1,0 +1,59 @@
+#!/usr/bin/python3
+
+import argparse
+import json
+import os
+import re
+import shlex
+import subprocess
+import tempfile
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--output', '-o', default='compile_commands.json',
+                    help='The output file [default: compile_commands.json]')
+args = parser.parse_args()
+
+directory = os.getcwd()
+
+with tempfile.NamedTemporaryFile() as makefile_copy:
+    # rebuilding the makefile is an exception to --dry-run, and we don't want
+    # to run the whole configure machinery, so copy the Makefile to a temporary
+    # path to short-circuit the exception.
+    with open('Makefile', 'rb') as original_makefile:
+        makefile_copy.write(original_makefile.read())
+
+    # get the build commands
+    output = subprocess.check_output(['make',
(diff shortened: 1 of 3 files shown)


jelly:
As noted on IRC, this only lacks the `command` key as noted in the [docs](https://clang.llvm.org/docs/JSONCompilationDatabase.html#format). Although it also mentions arguments is generallyy preferred.

martinpitt:
Moving to draft. Jelle had some comments, and ATM this is neither being used, documented, nor smoke-tested.

allisonkarlitskaya:
so, I'm already using this locally to help [ALE](https://github.com/dense-analysis/ale) produce correct warnings for editing C files.

allisonkarlitskaya:
hihi rhel8
```
  GEN      compile_commands.json
  File "./tools/make-compile-commands", line 39
    if match := cc_pattern.fullmatch(line):
              ^
SyntaxError: invalid syntax
```

martinpitt:
```
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'text'
```

tools/make-compile-commands
```diff
@@ -0,0 +1,59 @@
...
+    with open('Makefile', 'rb') as original_makefile:
+        makefile_copy.write(original_makefile.read())
+
+    # get the build commands
+    output = subprocess.check_output(['make',
+                                      '-f', makefile_copy.name,
+                                      'MAKE=false',   # block recursive make
+                                      'V=1',          # simplify the output
+                                      '--always-make', '--dry-run', 'all-am'],
+                                     text=True)
```
martinpitt, line 31: Should be `universal_newlines=True` for backwards compatibility

allisonkarlitskaya:
Interesting thing to note: this introduces a build dependency on Python.  I don't think we have one of those yet, for the tarball: the closest we come is `termschutz`, and this is only used during `make dist`.

martinpitt (approved):
Thanks! Looking forward to testing it. (I didn't yet, as this is "mostly harmless", and if it works for you, that's great)

PR #16848, opened 2022-01-19
test: sudo and socketpair() now work properly on fedora-35

martinpitt:
The fix made it to Fedora 35 stable as well. Similar to what commit
d9eda6d263 did for Fedora 34.

---

Blocks and needs to land in lockstep with a fedora-35 image refresh that includes [selinux-policy-35.11-1.fc35 ](https://bodhi.fedoraproject.org/updates/FEDORA-2022-87a0b7e8d0): https://github.com/cockpit-project/bots/pull/2839

This has caused [f35 packit failures](http://artifacts.dev.testing-farm.io/6ef24b96-630c-4044-b24a-d4509c34b4f4/) a few days ago when the new SELinux landed. This PR should make them green again.

PR #16849, opened 2022-01-19
[rhel-8.5] test: Don't expect only one item on dashboard 

marusak:
This page consists of one `table` element with more `tr` elements, where
each has a button. Previous selector was not unique if there were more
than one item in the table.

Cherry-picked from main branch commit 73497a629


martinpitt (approved):

martinpitt:
TF failures are "ok" -- c8s is just infra error, and F34 is due to #16848

martinpitt:
[This is unrelated](https://logs.cockpit-project.org/logs/pull-16849-20220119-090844-ce19f66c-centos-8-stream/log.html#170-2), apparently the SELinux fix now reached c8s and we need to update the tests. But  let's not block the image refresh on this.

PR #16850, opened 2022-01-19
test: Add known journal message to logouts

marusak:
Right now we have this pattern in `allow_restart_journal_messages` but
it is documented in `src/bridge/cockpitpolkitagent.c` as:
```
This can happen if there's a race between the polkit request and closing of
Cockpit. So it's not unheard of. We can complain, but not too loudly.
```

Logout causes closing of Cockpit as well.

As seen [here (twice)](https://logs.cockpit-project.org/logs/pull-2825-20220119-022917-6b8f29bd-rhel-8-5-cockpit-project-cockpit/log.html#128)

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -650,6 +650,9 @@ class Browser:
                 self.click('#logout')
         self.expect_load()
 
+        # Log out can cause race with polkit request
+        self.allow_journal_messages(".*couldn't create polkit session subject: No session for pid.*")
+
     def relogin(self, path=None, user=None, superuser=None, wait_remote_session_machine=None):
         self.logout()
         if wait_remote_session_machine:


test/common/testlib.py
```diff
@@ -650,6 +650,9 @@ def logout(self):
                 self.click('#logout')
         self.expect_load()
 
+        # Log out can cause race with polkit request
+        self.allow_journal_messages(".*couldn't create polkit session subject: No session for pid.*")
```
martinpitt, line 654: Not sure if that wouldn't better belong into allow_restart_messages()`, but it's fine. One of these days we need to clean this up anyway.

martinpitt (approved):
Dakujem!

marusak:
It seemed like a fluke. I'll watch out for this

PR #16851, opened 2022-01-19
debian/control: pin pages to the same bridge version

allisonkarlitskaya:
At Debian package build time, we run a Python script which scans all of our manifest.json files and returns a single number representing the highest version of the bridge that any of the packages depends on, and then we add this as a dependency for all of the pages packages.                                       

It's a bit arbitrary that we take the highest version and apply it as a dependency for every single pages package, when we could scan the individual manifest for the page in question and generate the dependency more precisely.
                                                                           
On the other hand, however, we gain very little value from any of this: the new versions of the binary packages are all released at the same time.                                                  
                                                                       
Let's drop this and just pin the pages to the same cockpit-bridge version.  This also lets us drop our `override_dh_gencontrol:` rule from debian/rules.

### DIFF
--- a/tools/debian/control
+++ b/tools/debian/control
@@ -78,7 +78,7 @@ Package: cockpit-networkmanager
 Architecture: all
 Multi-Arch: foreign
 Depends: ${misc:Depends},
-         cockpit-bridge (>= ${bridge:minversion}),
+         cockpit-bridge (>= ${source:Version}),
          network-manager (>= 1.6)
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks. This is much stricter than necessary, but as you said it doesn't matter in practice for Debian/Ubuntu, and makes things simpler. (It *does* matter for rpm/RHEL still due to the cockpit/cockpit-appstream split, though!)

PR #16852, opened 2022-01-19
[r8.5] test: sudo with cert auth now works on RHEL 8.6/C8S

martinpitt:
The sudo fix made it into RHEL 8.6 nightly and CentOS 8 Stream. Adjust
the test special case accordingly.

Cherry-picked from main commit fd9e6a09fd20c.

----

Should fix [this failure](https://logs.cockpit-project.org/logs/pull-16849-20220119-090844-ce19f66c-centos-8-stream/log.html#170-2) which I ignored in #16849.


marusak (approved):
Thanks! Assuming green tests

PR #16855, opened 2022-01-19
lib: cockpit-po-plugin: add support for file prefixes in pattern creation

KKoukiou:
This is necessary in the case that translations are not generated in the same directory
where webpack lives.

This happens in the new anaconda webui cockpit based plugin.

See more https://github.com/rhinstaller/anaconda/pull/3785

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -11,9 +11,9 @@ module.exports = class {
     constructor(options) {
         if (!options)
             options = {};
-        this.filenamePrefixPattern = options.filenamePrefixPattern || '';
+        this.filenamePrefix = options.filenamePrefix || '';
         this.subdir = options.subdir || '';
-        this.reference_patterns = options.reference_patterns;
+        this.reference_paths = options.reference_paths;
         this.wrapper = options.wrapper || 'cockpit.locale(PO_DATA);';
     }
 
@@ -65,8 +65,8 @@ module.exports = class {
         return expr;
     }
 
-    build_patterns(compilation, extras) {
-        const patterns = [
+    build_paths(compilation, extras) {
+        const paths = [
             // all translations for that page, including manifest.json and *.html
             `pkg/${this.subdir}.*`,
             // FIXME: https://github.com/cockpit-project/cockpit/issues/13906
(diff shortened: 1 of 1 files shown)


pkg/lib/cockpit-po-plugin.js
```diff
@@ -80,7 +81,7 @@ module.exports = class {
 
         Array.prototype.push.apply(patterns, extras);
 
-        return patterns.map((p) => new RegExp(`^${p}:[0-9]+$`));
+        return patterns.map((p) => new RegExp(`^${this.filePrefix}${p}:[0-9]+$`));
```
allisonkarlitskaya, line 84: I guess we should be escaping the path to guard against regexp-y characters that we'd find in there, no?

That was probably already true for the `${p}` (and this would be a welcome cleanup) but now that we include other paths, we might imagine seeing some more 'exotic' things there...

allisonkarlitskaya (changes requested):
I wonder if it might be nicer to rather provide a configuration option to specify the root directory from which the filenames in the po files were calculated and compute the difference between that, and the directory that webpack is running in.  I guess that's too much pain for not really very much improvement, though.

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: We might also just check that the string has the expected prefix, to be honest.  I'm not entirely 100% sure why I did it like this...

allisonkarlitskaya (changes requested):
Definitely a lot better, thank you.

One suggestion to make this nicer, though.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -64,8 +65,8 @@ module.exports = class {
         return expr;
     }
 
-    build_patterns(compilation, extras) {
-        const patterns = [
+    build_paths(compilation, extras) {
+        const paths = [
             // all translations for that page, including manifest.json and *.html
             `pkg/${this.subdir}.*`,
```
allisonkarlitskaya, line 71: Ugh.  This is why we were using a regexp, actually.  This actually is *a pattern*.  :(

allisonkarlitskaya (changes requested):

KKoukiou:
Using the exising reference_patterns exposed option works as well. Thanks @lis for the pointers

PR #16858, opened 2022-01-20
test: Update lists of supported OSes

marusak:
I need this to run some tests first :) Please don't land


martinpitt (approved):
Thank you! Failed packit tests are because your fork is out of date and the commits since then completely changed how packit srpm works. But that's fine in this case.

Issue #8519, opened 2018-01-30
Provide GUI for crypto policies


martinpitt:
Fedora has had a CLI and config API for setting global [crypto policies](https://fedoraproject.org/wiki/Changes/CryptoPolicy) for a while. For the most part this is a setting which large institutions want to roll out globally through configuration management systems (mostly changing `/etc/crypto-policies/config`).

However, @nmav pointed out that in some cases this makes sense in "troubleshooting" mode. E. g. your server fails to connect to your mail server as your current crypto policy forbids the usage of the auth algorithm that they use. So one could try a different mode, see if it works, and then take actions to either roll out the different policy or file a bug, etc. Once you found a setting that works, the page then should show you how to apply it to other systems, e. g. the command line or that being wrapped in an Ansible playbook.

It's a global setting, so it doesn't make sense on the subsystem pages. For the global /system page it's a too specialized use case (in particular, we don't want users to mess around with that unless they really know what they are doing). Thus this would make a nice start for a standalone Cockpit App.

Notes:
 * `update-crypto-policy` command lives in a separate package; preferred over changing the config file
 * Enabling/disabling FIPS requires a reboot
 * Policies are abstract names, thus they apply to different OS versions; but might not mean the same thing on them.
 * UI page needs to describe what the policy does; this needs some help from the `crypto-policies` package, either through exporting that in machine readable format (JSON?) from which the manpages could be generated as well; or some kind of `update-crypto-policy describe`.

martinpitt:
@nmav, could you please look through this whether I captured this correctly?

nmav:
Your description is accurate, thank you.

PR #16860, opened 2022-01-20
systemd: Implement crypto policies support

jelly:
This implements the ability to set and show crypto policies from within
the cockpit system overview. The current implementation is limited to
three hardcoded policies and `update-security-policies` does not have an
option to list all existing policies. When a crypto policy is applied
but a user chooses to not reboot a notification is shown in the health
card overview.

---

# Crypto policies support

Cockpit now has basic support for displaying and changing a subset of crypto policies (default, future, legacy). The current configured crypto policy is shown in the `configuration` card.

image

Crypto policy changes can be applied directly or later, when later is selected the health card shows that a reboot is required to fully apply the changed crypto policy.

image

### DIFF
--- a/pkg/lib/cockpit-components-privileged.jsx
+++ b/pkg/lib/cockpit-components-privileged.jsx
@@ -16,12 +16,14 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
 import { Button, Tooltip, TooltipPosition } from '@patternfly/react-core';
 
 import cockpit from "cockpit";
+import { superuser } from 'superuser';
+import { useEvent } from "hooks";
 
 /**
  * UI element wrapper for something that requires privilege. When access is not
@@ -47,42 +49,23 @@ export function Privileged({ excuse, allowed, placement, tooltipId, children }) 
 /**
  * Convenience element for a Privilege wrapped Button
  */
-export class PrivilegedButton extends React.Component {
-    constructor(props) {
-        super(props);
-        this.state = { allowed: !!this.props.permission.allowed };
-        this.onPermissionChanged = this.onPermissionChanged.bind(this);
-    }
+export const PrivilegedButton = ({ tooltipId, placement, excuse, buttonId, onClick, ariaLabel, variant, children }) => {
+    const [user, setUser] = useState(null);
+    useEvent(superuser, "changed");
+    useEffect(() => cockpit.user().then(user => setUser(user)));
 
-    componentDidMount() {
-        this.props.permission.addEventListener("changed", this.onPermissionChanged);
-        this.onPermissionChanged();
-    }
-
-    componentWillUnmount() {
-        this.props.permission.removeEventListener("changed", this.onPermissionChanged);
-    }
-
-    onPermissionChanged() {
-        // default to allowed while not yet initialized
-        this.setState({ allowed: !!this.props.permission.allowed });
-    }
-
-    render() {
-        return (
-            <Privileged allowed={ this.state.allowed } tooltipId={ this.props.tooltipId } placement={ this.props.placement }
-                        excuse={ cockpit.format(this.props.excuse, this.props.permission.user ? this.props.permission.user.name : '') }>
-                <Button id={ this.props.buttonId } variant={ this.props.variant } onClick={ this.props.onClick }
-                        isInline isDisabled={ !this.state.allowed } aria-label={ this.props.ariaLabel }>
-                    { this.props.children }
-                </Button>
-            </Privileged>
-        );
-    }
-}
+    return (
+        <Privileged allowed={ superuser.allowed } tooltipId={ tooltipId } placement={ placement }
+                    excuse={ cockpit.format(excuse, user ? user.name : '') }>
+            <Button id={ buttonId } variant={ variant } onClick={ onClick }
+                    isInline isDisabled={ !superuser.allowed } aria-label={ ariaLabel }>
+                { children }
+            </Button>
+        </Privileged>
+    );
+};
 
 PrivilegedButton.propTypes = {
-    permission: PropTypes.object.isRequired,
     excuse: PropTypes.string.isRequired, // must contain a $0, replaced with user name
     onClick: PropTypes.func,
     variant: PropTypes.string,
--- a/pkg/systemd/overview-cards/configurationCard.jsx
+++ b/pkg/systemd/overview-cards/configurationCard.jsx
@@ -30,6 +30,7 @@ import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
 import { ServerTimeConfig } from 'serverTime.js';
 import { RealmdClient, RealmButton } from "./realmd.jsx";
 import { TunedPerformanceProfile } from '../../tuned/dialog.jsx';
+import { CryptoPolicyRow } from './cryptoPolicies.jsx';
 
 import "./configurationCard.scss";
 
@@ -82,6 +83,8 @@ export const ConfigurationCard = ({ hostname }) => {
                                 <th scope="row">{_("Performance profile")}</th>
                                 <td><TunedPerformanceProfile /></td>
                             </tr>
+
+                            <CryptoPolicyRow />
 
                             <tr>
                                 <th scope="row">{_("Secure shell keys")}</th>
--- a/pkg/systemd/overview-cards/cryptoPolicies.jsx
+++ b/pkg/systemd/overview-cards/cryptoPolicies.jsx
@@ -1,0 +1,210 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import cockpit from "cockpit";
+import React, { useState, useEffect } from 'react';
+import { Button, Flex, Modal, Popover } from "@patternfly/react-core";
+import { ExternalLinkSquareAltIcon, HelpIcon, InProgressIcon } from '@patternfly/react-icons';
+
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
+import { PrivilegedButton } from "cockpit-components-privileged.jsx";
+import { TunedDialogBody } from '../../../pkg/tuned/change-profile.jsx';
+
+import "./cryptoPolicies.scss";
+
+const _ = cockpit.gettext;
+
+// Found in /usr/share/crypto-policies/policies/
+const cryptopolicies = {
+    DEFAULT: _("Recommended, secure settings for current threat models."),
+    FUTURE: _("Protects from anticipated near-term future attacks at the expense of interoperability."),
+    LEGACY: _("Higher interoperability at the cost of an increased attack surface."),
+};
+const applyNeedRebootStamp = "/run/cockpit/crypto-policies-reboot-stamp";
+const displayProfileText = profile => profile.charAt(0) + profile.slice(1, profile.length).toLowerCase();
+
+export const CryptoPolicyRow = () => {
+    const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState(null);
+    const [isOpen, setIsOpen] = useState(false);
+
+    useEffect(() => {
+        // Avoid cockpit.spawn every re-render from parent
+        if (currentCryptoPolicy === null) {
+            cockpit.spawn(["update-crypto-policies", "--show"], { err: "message" })
+                    .then(output => setCurrentCryptoPolicy(output.trim()))
+                    .catch(err => {
+                        console.debug("no crypto policies support: ", err.toString());
+                        setCurrentCryptoPolicy(undefined);
+                    });
+            // When users run `update-crypto-policies` manually.
+            cockpit.file("/etc/crypto-policies/state/current")
+                    .watch(content => setCurrentCryptoPolicy(content ? content.trim() : null));
+        }
+    }, [currentCryptoPolicy]);
+
+    if (!currentCryptoPolicy) {
+        return null;
+    }
+
+    return (
+        <tr>
+            <th scope="row">{_("Crypto policy")}</th>
+            <td>
+                <PrivilegedButton variant="link" buttonId="crypto-policy-button" tooltipId="tip-crypto-policy"
+                                  excuse={ _("The user $0 is not permitted to change crypto policies") }
+                                  onClick={() => setIsOpen(true)}>
+                    {displayProfileText(currentCryptoPolicy)}
+                </PrivilegedButton>
+                {isOpen &&
+                <CryptoPolicyDialog close={() => setIsOpen(false)}
+                                    currentCryptoPolicy={currentCryptoPolicy}
+                                    setCurrentCryptoPolicy={setCurrentCryptoPolicy}
+                />
+                }
+            </td>
+        </tr>
+    );
+};
+
+const CryptoPolicyDialog = ({
+    close,
+    currentCryptoPolicy,
+    setCurrentCryptoPolicy,
+}) => {
+    const [error, setError] = useState();
+    const [selected, setSelected] = useState(currentCryptoPolicy);
+
+    const setPolicy = (reboot) => {
+        cockpit.spawn(["update-crypto-policies", "--set", selected], { superuser: "require", err: "message" })
+                .then(() => {
+                    setCurrentCryptoPolicy(selected);
+                    if (reboot) {
+                        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
+                                .catch(error => setError(error));
+                    } else {
+                        cockpit.file(applyNeedRebootStamp, { superuser: "require" }).replace("\n")
+                                .then(() => close());
+                    }
+                })
+                .catch(error => setError(error));
+    };
+
+    const policies = Object.keys(cryptopolicies).map(policy => ({
+        name: policy,
+        title: displayProfileText(policy),
+        description: cryptopolicies[policy],
+        active: policy === currentCryptoPolicy,
+        recommended: false,
+    }));
+
+    // Custom profile
+    if (!(currentCryptoPolicy in cryptopolicies)) {
+        policies.push({
+            name: currentCryptoPolicy,
+            title: displayProfileText(currentCryptoPolicy),
+            description: _("Custom crypto policy"),
+            active: true,
+            recommended: false,
+        });
+    }
+
+    const help = (
+        <Popover
+            id="crypto-policies-help"
+            bodyContent={
+                <div>
+                    {_("Crypto Policies is a system component that configures the core cryptographic subsystems, covering the TLS, IPSec, SSH, DNSSec, and Kerberos protocols.")}
+                </div>
+            }
+            footerContent={
+                <Button component='a'
+                        rel="noopener noreferrer" target="_blank"
+                        variant='link'
+                        isInline
+                        icon={<ExternalLinkSquareAltIcon />} iconPosition="right"
+                        href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening">
+                    {_("Learn more")}
+                </Button>
+            }
+        >
+            <Button variant="plain" aria-label={_("Help")}>
+                <HelpIcon />
+            </Button>
+        </Popover>
+    );
+
+    return (
+        <Modal position="top" variant="medium"
+               isOpen
+               help={help}
+               onClose={close}
+               id="crypto-policy-dialog"
+               title={_("Change crypto policy")}
+               footer={
+                   <>
+                       {error && <ModalError dialogError={typeof error == 'string' ? error : error.message} />}
+                       <Button id="crypto-policy-apply-reboot" variant='primary' onClick={() => setPolicy(true)}>
+                           {_("Apply and reboot")}
+                       </Button>
+                       <Button id="crypto-policy-apply-reboot-later" variant='secondary' onClick={() => setPolicy(false)}>
+                           {_("Apply only")}
+                       </Button>
+                       <Button variant='link' onClick={close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }
+        >
+            {currentCryptoPolicy && <TunedDialogBody active_profile={currentCryptoPolicy}
+                                                     change_selected={setSelected}
+                                                     profiles={policies} />}
+        </Modal>
+    );
+};
+
+export const CryptoPolicyStatus = () => {
+    const [requiresReboot, setRequiresReboot] = useState(false);
+    const [showShutdownModal, setShowShutDownModal] = useState(false);
+
+    useEffect(() => {
+        if (!requiresReboot) {
+            // For when we change the setting
+            cockpit.file(applyNeedRebootStamp).watch(content => setRequiresReboot(content !== null));
+        }
+    }, [requiresReboot]);
+
+    if (!requiresReboot) {
+        return null;
+    }
+
+    return (
+        <li className="system-health-crypto-policies">
+            <Flex flexWrap={{ default: 'nowrap' }} spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
+                <InProgressIcon size="sm" className="crypto-policies-health-card-icon" />
+                <Button isInline variant="link" onClick={() => setShowShutDownModal(true)}>
+                    {_("Reboot to apply new crypto policy")}
+                </Button>
+            </Flex>
+            {showShutdownModal &&
+            <ShutdownModal onClose={() => setShowShutDownModal(false)} />
+            }
+        </li>
+    );
+};
--- a/pkg/systemd/overview-cards/cryptoPolicies.jsx
+++ b/pkg/systemd/overview-cards/cryptoPolicies.jsx
@@ -25,7 +25,7 @@ import { ExternalLinkSquareAltIcon, HelpIcon, InProgressIcon } from '@patternfly
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
 import { PrivilegedButton } from "cockpit-components-privileged.jsx";
-import { TunedDialogBody } from '../../../pkg/tuned/change-profile.jsx';
+import { ProfilesMenuDialogBody } from "./profiles-menu-dialog-body.jsx";
 
 import "./cryptoPolicies.scss";
 
@@ -172,7 +172,7 @@ const CryptoPolicyDialog = ({
                    </>
                }
         >
-            {currentCryptoPolicy && <TunedDialogBody active_profile={currentCryptoPolicy}
+            {currentCryptoPolicy && <ProfilesMenuDialogBody active_profile={currentCryptoPolicy}
                                                      change_selected={setSelected}
                                                      profiles={policies} />}
         </Modal>
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -835,6 +835,66 @@ password=foobar
         dbus_call = 'busctl get-property org.freedesktop.login1 /org/freedesktop/login1  org.freedesktop.login1.Manager ScheduledShutdown'
         self.assertEqual('(st) "" 0', m.execute(dbus_call).strip())
 
+    @skipImage("crypto-policies not avaiable", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable", "arch")
+    def testCryptoPolicies(self):
+        def shown_profile_text(profile):
+            return profile[0] + profile[1:].lower()
+
+        def change_profile(profile, new_profile, reboot=False):
+            b.click("#crypto-policy-button")
+            b.wait_in_text(".pf-c-menu__item.pf-m-selected", shown_profile_text(profile))
+            profile_button_name = shown_profile_text(new_profile)
+            b.click(f".pf-c-menu__item-main .pf-c-menu__item-text:contains('{profile_button_name}')")
+            if reboot:
+                b.click("#crypto-policy-apply-reboot")
+            else:
+                b.click("#crypto-policy-apply-reboot-later")
+                b.wait_text("#crypto-policy-button", shown_profile_text(new_profile))
+                profile = m.execute(cmd + " --show").strip()
+                self.assertEqual(new_profile, profile)
+
+        m = self.machine
+        b = self.browser
+        cmd = "update-crypto-policies"
+
+        self.login_and_go("/system")
+
+        profile = m.execute(cmd + " --show").strip()
+        b.wait_text("#crypto-policy-button", shown_profile_text(profile))
+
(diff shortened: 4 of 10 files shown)


pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+    FUTURE: _("A level that will provide security on a conservative level that is believed to withstand any near-term future attacks. And also provide some (not complete) preparation for post quantum encryption support in form of 256 bit symmetric encryption requirement. It provides at least an 128-bit security. This level may prevent communication with many used systems that provide weaker security levels (e.g., systems that use SHA-1 as signature algorithm)."),
+    LEGACY: _("Provides settings for ensuring maximum compatibility with legacy systems. This policy is less secure and intended to be a easy way to switch system to be compatible with older systems. It should provide at least 80-bit security and excludes 3DES and RC4."),
+};
+const defaultpolicy = "DEFAULT";
+const stampfile = "/run/cockpit/crypto-policies-reboot-stamp";
+
+export const CryptoPolicyRow = () => {
+    const [showCryptoPolicy, setShowCryptoPolicy] = useState(false);
+    const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState("");
+    const [isOpen, setIsOpen] = useState();
```
martinpitt, line 45: init to `false` to point out that  it's a bool?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+                setCurrentCryptoPolicy(output.trim());
+                setShowCryptoPolicy(true);
+            });
+        }
+    }, [showCryptoPolicy]);
+
+    if (!showCryptoPolicy) {
+        return null;
+    }
+    const permission = cockpit.permission({ admin: true });
```
martinpitt, line 60: This is old-fashioned. It doesn't react to superuser changes on the page. Please use pkg/lib/superuser.js instead.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+    const [selected, setSelected] = useState();
+
+    const setPolicy = (reboot) => {
+        cockpit.spawn(["update-crypto-policies", "--set", selected], { superuser: "require", err: "message" }).then(() => {
+            setCurrentCryptoPolicy(selected);
+            if (reboot) {
+                cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
+                        .catch(error => setError(error));
+            } else {
+                cockpit.spawn(["touch", stampfile], { superuser: "require", err: "message" }).then(() => {
```
martinpitt, line 97: cockpit.file.replace() avoids calling a program.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+                    close();
+                });
+            }
+        })
+                .catch(error => {
```
martinpitt, line 102: This is either mis-indented or misplaced.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+                        rel="noopener noreferrer" target="_blank"
+                        variant='link'
+                        isInline
+                        icon={<ExternalLinkSquareAltIcon />} iconPosition="right"
+                        href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening">
+                    {_("Learn more")}
+                </Button>
+            }
+        >
+            <Button variant="plain" aria-label={_("Help")}>
```
martinpitt, line 136: I know the popover provides the functionality, and this should be focusable. It just looks weird to have a button without an action or link, is that legit?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+    );
+};
+
+export const CryptoPolicyStatus = () => {
+    const [requiresReboot, setRequiresReboot] = useState(false);
+
+    useEffect(() => {
+        if (!requiresReboot) {
+            cockpit.file(stampfile).read()
+                    .then((content, tag) => {
```
martinpitt, line 176: Doesn't use tag, just use `content =>` here.

martinpitt (changes requested):
Thanks! A few remarks about the code. I'll look at the TODOs separately.

martinpitt:
> Reboot notification is not shown when clicking "restart later" only after a page reload, as the file is not being "watched".

You mean when you reload the page while the dialog is open? Then any change wouldn't get applied either? The health card independently looks at the stamp file, why wouldn't it pick up an applied change?

> Should the "reboot notification" have an action to reboot the machine?

I think yes. It is currently a link, but without any target/action. There's the reboot dialog in pkg/lib/cockpit-components-shutdown.jsx, or perhaps it could even use the overview's reboot dialog with passing around some handlers as properties.

> Are the current crypto policies acceptable?

Seems so, these are the three documented ones. However, it should disable itself if there is any customization of the standard policies, or even a custom one.

jelly:
> > Reboot notification is not shown when clicking "restart later" only after a page reload, as the file is not being "watched".
> 
> You mean when you reload the page while the dialog is open? Then any change wouldn't get applied either? The health card independently looks at the stamp file, why wouldn't it pick up an applied change?

After clicking `reboot later` the healthcard entry does not show up, most likely as the component is not re-rendered / no prop changes.
 
> > Should the "reboot notification" have an action to reboot the machine?
> 
> I think yes. It is currently a link, but without any target/action. There's the reboot dialog in pkg/lib/cockpit-components-shutdown.jsx, or perhaps it could even use the overview's reboot dialog with passing around some handlers as properties.

Hmmmm, yeah I'm not sure if an immediate reboot would be expected or if we at least should provide a warning.

> > Are the current crypto policies acceptable?
> 
> Seems so, these are the three documented ones. However, it should disable itself if there is any customization of the standard policies, or even a custom one.

Good point, I haven't tested it yet with non-default options in the [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2038841#c3) issue there is a suggestion to simplify the description.

garrett:
1. Are the upper case item titles really necessary? They should be title case and bold, not uppercase.
2. Recommended and active badges should probably be different colors.
3. What's the difference between active and the checkmark?
4. Should the badges be next to the item titles?
5. Wow, that text is verbose. Is it supposed to be light grey? That makes it harder to read. But I guess if the PF widget intends for it light grey, then that's OK (as long as contrast is at least AA).

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+                    {currentCryptoPolicy}
+                </PrivilegedButton>
+                {isOpen &&
+                <CryptoPolicyDialog close={() => setIsOpen(false)}
+                          currentCryptoPolicy={currentCryptoPolicy}
+                          setCurrentCryptoPolicy={setCurrentCryptoPolicy}
+                />
+                }
+            </td>
+        </tr>
```
garrett, line 78: Why are we using tables here? Why aren't we using the PF menu item, like in languages (at least I think languages is now using it)?

I think that a plain, scrollable menu is probably what's wanted? And probably with a border, so it's obvious it's a widget it's scrollable.

https://www.patternfly.org/v4/components/menu#with-description
https://www.patternfly.org/v4/components/menu/html#plain (`isPlain` and `isScrollable` in React, although there's no demo for it in React)

I don't see a helper to add a border, oddly. That seems like an omission in PF. I suppose we could do something in custom CSS like:

```css
border: var(--pf-global--BorderWidth--sm) solid var(--pf-global--BorderColor--100)
```

jelly: This is the overview card which always used a table and we can't really change all the overview cards in this PR but we can make an issue to convert it in a separate PR.

jelly:
Reported upstream https://github.com/systemd/systemd/issues/22237

pkg/systemd/overview-cards/cryptoPolicies.jsx
jelly: Hmm, it's something eslint seems to format for me.

martinpitt: Oh, I see -- please move the cockpit.spawn()'s `.then` on the next line, then it's properly aligned.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,199 @@
...
+    const [showCryptoPolicy, setShowCryptoPolicy] = useState(false);
+    const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState("");
+    const [isOpen, setIsOpen] = useState();
+
+    useEffect(() => {
+        // Avoid cockpit.spawn every re-render from parent
+        if (!showCryptoPolicy) {
+            cockpit.spawn(["update-crypto-policies", "--show"]).then(output => {
+                setCurrentCryptoPolicy(output.trim());
+                setShowCryptoPolicy(true);
```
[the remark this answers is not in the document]
jelly: Hmm it starts with an empty string in setState. So where would it be set to `undefined`?

pkg/systemd/overview-cards/cryptoPolicies.jsx
jelly: I think it looks alright, I'm not sure where else we would put the extra information.

jelly:
> > > Reboot notification is not shown when clicking "restart later" only after a page reload, as the file is not being "watched".
> > 
> > 
> > You mean when you reload the page while the dialog is open? Then any change wouldn't get applied either? The health card independently looks at the stamp file, why wouldn't it pick up an applied change?

The file is now watched, so it's no longer an issue
 
> After clicking `reboot later` the healthcard entry does not show up, most likely as the component is not re-rendered / no prop changes.
> 
> > > Should the "reboot notification" have an action to reboot the machine?
> > 

We have decided to re-use current reboot dialog, as we shouldn't allow users to "accidently" reboot.
 
> > I think yes. It is currently a link, but without any target/action. There's the reboot dialog in pkg/lib/cockpit-components-shutdown.jsx, or perhaps it could even use the overview's reboot dialog with passing around some handlers as properties.
> 
> Hmmmm, yeah I'm not sure if an immediate reboot would be expected or if we at least should provide a warning.
> 
> > > Are the current crypto policies acceptable?
> > 
> > 
> > Seems so, these are the three documented ones. However, it should disable itself if there is any customization of the standard policies, or even a custom one.
> 
> Good point, I haven't tested it yet with non-default options in the [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2038841#c3) issue there is a suggestion to simplify the description.

garrett:
IIRC: s/restart/reboot/ when talking about the system; restart is used for services, whereas reboot is unambiguous.

garrett:
Also, "Apply and reboot" instead of "Apply and Restart" (note the sentence case)

pkg/systemd/overview-cards/cryptoPolicies.jsx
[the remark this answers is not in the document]
martinpitt: `undefined` was just a proposal, but it feels right if calling u-c-p fails (like on Debian, for example). That's the missing `.catch()` handler I mentioned. You can also set it back to `null`, but you somehow need an indication whether the initialization just didn't happen yet (null) or whether it ran and failed (something else which is not `null` but also not a valid policy)

jelly:
> Also, "Apply and reboot" instead of "Apply and Restart" (note the sentence case)

100% agree, applied.

jelly:
Use the `InProgressIcon` icon in the healthcard.

jelly:
> Use the `InProgressIcon` icon in the healthcard.

fixed.

jelly:
> > 1. Are the upper case item titles really necessary? They should be title case and bold, not uppercase.
> 
> That's how the profiles are also set on the command line so it makes sense to keep it the same?
> 
> > 2. Recommended and active badges should probably be different colors.
> 
> This is a "shared" component from the tuned profiles, so if this design is chosen we should

I can't find a good recommendation for the active/recommended [colors](https://www.patternfly.org/v4/guidelines/colors/#primary-colors) in PF4.
> 
> > 3. What's the difference between active and the checkmark?
> 
> Check mark is selected, active is currently used.
> 
> > 4. Should the badges be next to the item titles?
> 
> Not sure
> 
> > 5. Wow, that text is verbose. Is it supposed to be light grey? That makes it harder to read. But I guess if the PF widget intends for it light grey, then that's OK (as long as contrast is at least AA).
> 
> In the bugzilla ticket there is a simplified version of the text we can use.

garrett:
> @garrett I can't find a good recommendation for the active/recommended [colors](https://www.patternfly.org/v4/guidelines/colors/#primary-colors) in PF4. Do you have some ideas/input?

During the transition of PF3 to PF4 (which took a long time), I had a lot of `--ct-color-` variables, based on the PF palette. I have since removed most of them, except for the ones in use for various widgets. (It's probably a good idea to remove more soon.)

The relevant ones to your query are:

```css
    --ct-color-list-selected-text      : var(--ct-color-link);
    --ct-color-list-selected-bg        : var(--pf-global--BackgroundColor--150);
    --ct-color-list-selected-border    : var(--pf-global--BackgroundColor--150);
    --ct-color-list-active-text        : var(--pf-global--palette--blue-500);
    --ct-color-list-active-bg          : var(--ct-color-bg);
    --ct-color-list-active-border      : var(--ct-color-list-border);
```

But these might need to be revised, as they're built with a different selection concept in mind.

However, the color values used are probably what we want, especially as PF uses blue for selection and active color (which is pretty typical across operating systems and UI pattern libraries).

garrett:
Mockup with some adjustments.

Main changes:

0. No uppercase!
1. Compact "label" (bubble) next to the text.
2. Shortended text strings, based on the bugzilla, but reworked a little
3. Moved "Recommended" into the summary, as a modifier for settings "Recommended, secure settings..."
4. Action buttons are "Apply and restart" and "Apply only" (this is subject for discussion; it could be just "Apply" or "Apply without a restart")

Crypto policy modal

garrett:
Hmm. Glancing at this again, I wonder if we need something like this, probably as text right under the heading, or sandwitched between the selector and the row of action buttons:

"Policy changes will be effective on next boot."

Otherwise the "Apply only" "Apply without a reboot" or whatever we call it will seem like it will take effect.

Also, "Apply" means that something will take effect normally, unless otherwise stated, right? In this case, we probably even might want to consider "Save" instead. This is the whole Apply versus Save thing (which still hasn't landed). #16475 

WDYT?

jelly:
> Hmm. Glancing at this again, I wonder if we need something like this, probably as text right under the heading, or sandwitched between the selector and the row of action buttons:
> 
> "Policy changes will be effective on next boot."
> 
> Otherwise the "Apply only" "Apply without a reboot" or whatever we call it will seem like it will take effect.
> 
> Also, "Apply" means that something will take effect normally, unless otherwise stated, right? In this case, we probably even might want to consider "Save" instead. This is the whole Apply versus Save thing (which still hasn't landed). #16475
> 
> WDYT?

For reference the tool outputs:

```
[root@rhel-9-0-127-0-0-2-2201 ~]# update-crypto-policies --set FUTURE
Setting system policy to FUTURE
Note: System-wide crypto policies are applied on application start-up.
It is recommended to restart the system for the change of policies
to fully take place.
```

jelly:
> 

So what happens is that the policy is **applied** but it is recommended to restart services as the policies apply once a service (ie. nginx) is started. So apply is technically true but comes with a caveat.

garrett:
> @garrett I am leaving out the active change as this also applies to tuned (they share the same dialog)

We also solved this in the call, where they're grouped for proper spacing, aligned to the left now, and use different colors. So that works for me! :+1: 

Looking forward to reviewing this again soon. We're almost there!

jelly:
The three debian-stable/testing failures look like flakes? Fedora-35 needs pixel test updates, plus one small fix for https://logs.cockpit-project.org/logs/pull-16860-20220223-110816-c1e23cbd-fedora-35/log.html#298

martinpitt:
Systematic and relevant test failures in [TestSuperuserOldDashboard.test](https://logs.cockpit-project.org/logs/pull-16860-20220304-082649-b3d796c8-centos-8-stream/log.html) and [testTime](https://logs.cockpit-project.org/logs/pull-16860-20220304-082653-b3d796c8-fedora-35-firefox/log.html#306) -- seems this somehow breaks the superuser dialog?

martinpitt:
I set a customized policy with `update-crypto-policies --set DEFAULT:NO-SHA1`. It is shown (although in a bit weird way, "Default:no-sha1") and it remains editable:

image

Iniitally I thought this should be read-only, but this actually seems fine -- I can switch back to a default policy, and then the custom policy disappears.

What's a little weird is that I can switch from default to future, apply (w/o reboot), then switch back to default, and it still says "Reboot to apply new crypto policy". But that's not too bad. We could put the original policy into /run/cockpit/crypto-policies-reboot-stamp, but that's more house-keeping than necessary.

Otherwise this works really nicely!

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
+import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
+import { PrivilegedButton } from "cockpit-components-privileged.jsx";
+import { TunedDialogBody } from '../../../pkg/tuned/change-profile.jsx';
+
+import "./cryptoPolicies.scss";
+
+const _ = cockpit.gettext;
+
+const cryptopolicies = {
```
martinpitt, line 34: Please add a comment where you took the descriptions from, so that we can update them occasionally.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+    DEFAULT: _("Recommended, secure settings for current threat models."),
+    FUTURE: _("Protects from anticipated near-term future attacks at the expense of interoperability."),
+    LEGACY: _("Higher interoperability at the cost of an increased attack surface"),
```
martinpitt, line 37: Please add a period for consistency.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+};
+const stampfile = "/run/cockpit/crypto-policies-reboot-stamp";
```
martinpitt, line 39: something more descriptive, perhaps `applyNeedRebootStamp`?

pkg/systemd/overview-cards/cryptoPolicies.jsx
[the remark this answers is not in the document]
martinpitt: This comment is still fully valid.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+        return null;
+    }
+
+    return (
+        <tr>
+            <th scope="row">{_("Crypto policy")}</th>
+            <td>
+                <PrivilegedButton variant="link" buttonId="crypto-policy-button" tooltipId="tip-crypto-policy"
+                                excuse={ _("The user $0 is not permitted to change crypto policies") }
+                                permission={ permission } onClick={() => setIsOpen(true)}>
```
martinpitt, line 76: Please align the properties to the same column (i.e. indent two more spaces)

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+                    {displayProfileText(currentCryptoPolicy)}
+                </PrivilegedButton>
+                {isOpen &&
+                <CryptoPolicyDialog close={() => setIsOpen(false)}
+                          currentCryptoPolicy={currentCryptoPolicy}
+                          setCurrentCryptoPolicy={setCurrentCryptoPolicy}
```
martinpitt, line 82: property indentation

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+                .then(() => {
+                    setCurrentCryptoPolicy(selected);
+                    if (reboot) {
+                        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
+                                .catch(error => setError(error));
+                    } else {
+                        cockpit.file(stampfile, { superuser: "require" }).replace("\n")
+                                .then(() => {
+                                    close();
+                                });
```
martinpitt, line 109: Please drop the unnecessary block wrapper.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+                    }
+                })
+                .catch(error => setError(error));
+    };
+
+    const policies = Object.keys(cryptopolicies).map(policy => {
+        return {
```
martinpitt, line 116: Please replace `{ return ... }` code block with `({ ... })` to get a pure data mapping

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+                       </Button>
+                       <Button variant='link' onClick={close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }
+        >
+            {currentCryptoPolicy && <TunedDialogBody active_profile={currentCryptoPolicy}
+                                               change_selected={setSelected}
+                                               profiles={policies} />}
```
martinpitt, line 185: Property indentation

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+};
+
+export const CryptoPolicyStatus = () => {
+    const [requiresReboot, setRequiresReboot] = useState(false);
+    const [showShutdownModal, setShowShutDownModal] = useState(false);
+
+    useEffect(() => {
+        if (!requiresReboot) {
+            // For when we we re-render
+            cockpit.file(stampfile).read()
```
martinpitt, line 197: This is redundant. `.watch()` always fires the callback with the initial read, as [per documentation](https://cockpit-project.org/guide/latest/cockpit-file.html)

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,230 @@
...
+                    .then(content => {
+                        // cockpit.file does not error when a file does not exists instead returns null.
+                        if (content !== null) {
+                            setRequiresReboot(true);
+                        }
+                    });
+            // For when we change the setting
+            cockpit.file(stampfile).watch(content => {
+                if (content) {
```
martinpitt, line 206: `!== null` for correctness. It should work to have an empty, but existing file (and not write `\n` into it, but just set contents to an empty string)

test/verify/check-system-info
```diff
@@ -870,6 +870,66 @@ password=foobar
...
+                b.click("#crypto-policy-apply-reboot")
+            else:
+                b.click("#crypto-policy-apply-reboot-later")
+                b.wait_text("#crypto-policy-button", shown_profile_text(new_profile))
+                profile = m.execute(cmd + " --show").strip()
+                self.assertEqual(new_profile, profile)
+
+        m = self.machine
+        b = self.browser
+        cmd = "/usr/bin/update-crypto-policies"
```
martinpitt, line 893: Let's not hardcode the path. This should live in `$PATH` somewhere.

martinpitt (changes requested):
Thanks!

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+                        console.warn("no crypto policies support: ", err.toString());
+                        setCurrentCryptoPolicy(undefined);
+                    });
+            // When users run `update-crypto-policies` manually.
+            cockpit.file("/etc/crypto-policies/state/current").watch(content => {
+                if (content) {
+                    setCurrentCryptoPolicy(content.trim());
+                }
+            });
+            setPermission(cockpit.permission({ admin: true }));
```
mvollmer, line 64: We can't use `cockpit.permission` on the overview page. It will cause the page to reload when the superuser level changes.  Instead, copy what `realmd.jsx` does: Use `Privileged` with a superuser object.

mvollmer (changes requested):

jelly:
The superuserOldDashboard fails, as due to the usage of `cockpit.permission` the page reloads which closes the dialog which was currently open.

jelly:
> The superuserOldDashboard fails, as due to the usage of `cockpit.permission` the page reloads which closes the dialog which was currently open.

This is now resolved by changing the PrivilegedButton to use `superuser`.

jelly:
> 

Technically if you apply and then restart a daemon it runs in the new profile, so if you then switch back you have to reboot. So I actually think it's better to always show that you have to reboot.

pkg/lib/cockpit-components-privileged.jsx
```diff
@@ -47,42 +49,23 @@ export function Privileged({ excuse, allowed, placement, tooltipId, children })
...
-
-    componentDidMount() {
-        this.props.permission.addEventListener("changed", this.onPermissionChanged);
-        this.onPermissionChanged();
-    }
-
-    componentWillUnmount() {
-        this.props.permission.removeEventListener("changed", this.onPermissionChanged);
-    }
+export const PrivilegedButton = ({ tooltipId, placement, excuse, buttonId, onClick, ariaLabel, variant, children, permission }) => {
```
martinpitt, line 52: The `permission` property is not passed, declared, or used any more. Please drop from the list.

Would also be nice to move this commit first, so that the previous commit doesn't introduce something broken.

pkg/systemd/overview-cards/configurationCard.jsx
```diff
@@ -30,6 +30,7 @@ import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
 import { ServerTimeConfig } from 'serverTime.js';
 import { RealmdClient, RealmButton } from "./realmd.jsx";
 import { TunedPerformanceProfile } from '../../tuned/dialog.jsx';
+import { CryptoPolicyRow } from './cryptoPolicies.jsx';
```
martinpitt, line 33: Please fix the syntax in the commit message: `Close #8519` → `Fixes #8519`. (Traditionally we used "Close*s*" for closing PRs, and "Fixes" for fixing bugs)

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+
+export const CryptoPolicyRow = () => {
+    const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState(null);
+    const [isOpen, setIsOpen] = useState(false);
+    const [permission, setPermission] = useState(null);
+
+    useEffect(() => {
+        // Avoid cockpit.spawn every re-render from parent
+        if (currentCryptoPolicy === null) {
+            cockpit.spawn(["update-crypto-policies", "--show"], { err: "message" }).then(output => {
```
martinpitt, line 50: Please move the `.then` to next line, so that it aligns with the `.catch`. (Odd, normally ESLint complains about multiple chained function calls on one line..).

Also, not necessary to introduce a { } block.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+                setCurrentCryptoPolicy(output.trim());
+            })
+                    .catch(err => {
+                        console.warn("no crypto policies support: ", err.toString());
```
martinpitt, line 54: Let's tone this down to console.debug() -- this is not an error, and on many OSes it's not even available.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+                        setCurrentCryptoPolicy(undefined);
+                    });
+            // When users run `update-crypto-policies` manually.
+            cockpit.file("/etc/crypto-policies/state/current").watch(content => {
+                if (content) {
+                    setCurrentCryptoPolicy(content.trim());
+                }
```
martinpitt, line 61: It feels like this should handle `null` for robustness -- i.e. the file going away. Seems easier to do `content => setCurrentCryptoPolicy(content ? content.trim() : null)`

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+            });
+            setPermission(cockpit.permission({ admin: true }));
+        }
+    }, [currentCryptoPolicy]);
+
+    if (currentCryptoPolicy === undefined || currentCryptoPolicy === null) {
```
martinpitt, line 66: an empty string is just as good for ignoring this, so let's just check `!currentCryptoPolicy`

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+                       <Button id="crypto-policy-apply-reboot-later" variant='secondary' onClick={() => setPolicy(false)}>
+                           {_("Apply only")}
+                       </Button>
+                       <Button variant='link' onClick={close}>
+                           {_("Cancel")}
+                       </Button>
+                   </>
+               }
+        >
+            {currentCryptoPolicy && <TunedDialogBody active_profile={currentCryptoPolicy}
```
martinpitt, line 179: It feels rather odd (and looks like a copy&paste error) to use `TunedDialogBody` here. Could we split that out into a separate component and use it from both places? That's fine for me as a follow-up, though.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -0,0 +1,220 @@
...
+
+export const CryptoPolicyStatus = () => {
+    const [requiresReboot, setRequiresReboot] = useState(false);
+    const [showShutdownModal, setShowShutDownModal] = useState(false);
+
+    useEffect(() => {
+        if (!requiresReboot) {
+            // For when we change the setting
+            cockpit.file(applyNeedRebootStamp).watch(content => {
+                if (content !== null) {
```
martinpitt, line 194: Again, let's also handle `null` here, i.e. the file being absent or going away. Not a big difference in practice, but it's more robust and also makes the code easier (avoiding the `if`).

martinpitt (changes requested):
Thanks! some small nitpicks still, but works really nicely now!

martinpitt (approved):
Cheers! Good now once green. @mvollmer 's requested change got applied, so I'll drop this nack.

martinpitt:
The other failures are unrelated, but [this one](https://logs.cockpit-project.org/logs/pull-16860-20220314-151025-5bdd6c52-fedora-35/log.html#300-1) seems to show a race condition

jelly:
> The other failures are unrelated, but [this one](https://logs.cockpit-project.org/logs/pull-16860-20220314-151025-5bdd6c52-fedora-35/log.html#300-1) seems to show a race condition

Tried reproducing it here locally, ran the test 10 times on repeat but it did not flake or timeout. In the screenshot the button is shown and not greyed out so it's enabled and not disabled.

martinpitt:
In the second retry there are [two pixel failures](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16860-20220314-161814-5bdd6c52-fedora-35/log.html) but somehow our log.html forgot how to link to them?

jelly:
> In the second retry there are [two pixel failures](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16860-20220314-161814-5bdd6c52-fedora-35/log.html) but somehow our log.html forgot how to link to them?

Did someone break by trying to get rid of more jQuery? Probably it's https://github.com/cockpit-project/bots/pull/3065 which fixes that.

jelly:
> > In the second retry there are [two pixel failures](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16860-20220314-161814-5bdd6c52-fedora-35/log.html) but somehow our log.html forgot how to link to them?
> 
> Did someone break by trying to get rid of more jQuery? Probably it's [cockpit-project/bots#3065](https://github.com/cockpit-project/bots/pull/3065) which fixes that.

Made a seperate PR https://github.com/cockpit-project/cockpit/pull/17138 with no-test to just trigger the pixel test related images.

martinpitt (approved):
Thanks @jelly for cleaning up after my tasks update! I returned the favor with rebasing yours, as pixel refs conflicted.

PR #16861, opened 2022-01-20
branding: fix symlink installation

allisonkarlitskaya:
We need to add an extra / after the $(DESTDIR) and before /usr in order
to get the correct path.

...but the real scandal here is that not a single test failed :(

### DIFF
--- a/src/branding/arch/Makefile.am
+++ b/src/branding/arch/Makefile.am
@@ -7,6 +7,6 @@ archbranding_DATA = \
 EXTRA_DIST += $(archbranding_DATA)
 
 install-data-hook::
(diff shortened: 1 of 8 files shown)


martinpitt (approved):
Eww, thanks -- marking as release blocker. But yes, we should do a pixel test for the login page? That should have the Fedora logo. I suppose that's one of the things which would be broken by this?

martinpitt:
Indeed, current main image-prepare login page looks like this:

image

It should have the Fedora logo. After this lands, do you want to do a login page pixel test, or want me to?

allisonkarlitskaya:
> After this lands, do you want to do a login page pixel test, or want me to?

I'll give it a try :)

PR #16863, opened 2022-01-21
some build reproducibility fixes

allisonkarlitskaya:
Let's try to do our part to get reproducible builds working.

See https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/cockpit.html
See https://bugzilla.redhat.com/show_bug.cgi?id=2043295 for why this won't work on Fedora (at least not with gcc).


allisonkarlitskaya:
CentOS/RHEL 8 are failing due to an inability to create the -debugsource package, because find-debuginfo.sh filters by files being present in the source directory (which we now strip away).  Added a workaround to disable the prefix mapping on RHEL8.

src/bridge/askpass.c
```diff
@@ -27,7 +27,7 @@
 #include "common/cockpitmemory.h"
 #include "common/cockpittransport.h"
 #if WITH_DEBUG
-#include "common/cockpittest.h"
+#include "common/cockpitstacktrace.h"
```
martinpitt, line 30: I wonder if that is still actually the right thing to do. systemd-coredump does that these days, in a much nicer way; and Apport on the Ubuntu side. I wouldn't mind if we just ripped out all that stuff entirely -- I can't remember a single situation in the last 5 years where this would have actually helped us. I haven't even *seen* these in bug reports and journals.

martinpitt (changes requested):
Nice! I don't veto the current approach, but I'd be in favor of more cleanup.

src/bridge/askpass.c
allisonkarlitskaya: sehr gerne.

martinpitt (approved):
Thanks! This all looks nice. All green, but this is still marked as Draft. If you just forgot to flip it to ready, please ago ahead.

PR #16867, opened 2022-01-24
metrics: Improve load average splitting for narrow layouts

martinpitt:
Use Flex for a better layout, to avoid line breaks inside of the 1/5/15
minute fields.

---

## Current main:

Wide:
image

Medium narrow:
image

Minimum width:
image

## This PR:
Wide:
image

Medium narrow:
image

Minimum width:
image

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -216,7 +216,7 @@ class CurrentMetrics extends React.Component {
             swapUsed: null, // GiB
             cpuUsed: 0, // percentage
             cpuCoresUsed: [], // [ percentage ]
-            loadAvg: null, // string
+            loadAvg: null, // [ 1min, 5min, 15min ]
             disksRead: 0, // B/s
             disksWritten: 0, // B/s
             mounts: [], // [{ target (string), use (percent), avail (bytes) }]
@@ -322,8 +322,7 @@ class CurrentMetrics extends React.Component {
         cockpit.file("/proc/loadavg").read()
                 .then(content => {
                     // format: three load averages, then process counters; e.g.: 0.67 1.00 0.78 2/725 87151
-                    const load = content.split(' ').slice(0, 3);
-                    this.setState({ loadAvg: cockpit.format("$0: $1, $2: $3, $4: $5", _("1 min"), load[0], _("5 min"), load[1], _("15 min"), load[2]) });
+                    this.setState({ loadAvg: content.split(' ').slice(0, 3) });
                     // update it again regularly
                     window.setTimeout(this.updateLoad, 5000);
                 })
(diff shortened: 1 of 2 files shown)


martinpitt:
Spin-off from https://github.com/cockpit-project/cockpit/pull/16686#issuecomment-1014680922 . @garrett (may have) proposed that the fields completely switch to row layout once the cards become too narrow to put them all side by side, but I don't know how to do this. ([docs](https://www.patternfly.org/v4/layouts/flex)).

pkg/metrics/metrics.jsx
```diff
@@ -479,7 +478,13 @@ class CurrentMetrics extends React.Component {
                             <DescriptionList className="pf-m-horizontal-on-sm">
                                 <DescriptionListGroup>
                                     <DescriptionListTerm>{ _("Load") }</DescriptionListTerm>
-                                    <DescriptionListDescription id="load-avg">{this.state.loadAvg}</DescriptionListDescription>
+                                    <DescriptionListDescription id="load-avg">
+                                        <Flex>
+                                            <FlexItem spacer={{ default: 'spacerSm' }}>{ _("1 min") }:&nbsp;{ this.state.loadAvg[0] },</FlexItem>
```
martinpitt, line 483: It's a bit stupid having to repeat the spacer= attribute, instead of setting `<Flex spaceItems={{ modifier: 'spaceItemsSm' }}>` as the documentation suggests. But the latter just simply doesn't work. @KKoukiou , @marusak , do you happen to have any idea why?

KKoukiou: I think you just use it incorrectly.
`spaceItems={ default: 'spaceItemsXs' }` is the correct usage.

martinpitt:
Needs test adjustment, but waiting for initial feedback first.

pkg/metrics/metrics.jsx
KKoukiou: Right:
``` diff
diff --git pkg/metrics/metrics.jsx pkg/metrics/metrics.jsx
index 376ed0e27..5bb0c6e35 100644
--- pkg/metrics/metrics.jsx
+++ pkg/metrics/metrics.jsx
@@ -479,10 +479,10 @@ class CurrentMetrics extends React.Component {
                                 <DescriptionListGroup>
                                     <DescriptionListTerm>{ _("Load") }</DescriptionListTerm>
                                     <DescriptionListDescription id="load-avg">
-                                        <Flex>
-                                            <FlexItem spacer={{ default: 'spacerSm' }}>{ _("1 min") }:&nbsp;{ this.state.loadAvg[0] },</FlexItem>
-                                            <FlexItem spacer={{ default: 'spacerSm' }}>{ _("5 min") }:&nbsp;{ this.state.loadAvg[1] },</FlexItem>
-                                            <FlexItem spacer={{ default: 'spacerSm' }}>{ _("15 min") }:&nbsp;{ this.state.loadAvg[2] }</FlexItem>
+                                        <Flex spaceItems={{ default: 'spaceItemsXs' }}>
+                                            <FlexItem>{ _("1 min") }:&nbsp;{ this.state.loadAvg[0] },</FlexItem>
+                                            <FlexItem>{ _("5 min") }:&nbsp;{ this.state.loadAvg[1] },</FlexItem>
+                                            <FlexItem>{ _("15 min") }:&nbsp;{ this.state.loadAvg[2] }</FlexItem>
                                         </Flex>
                                     </DescriptionListDescription>
                                 </DescriptionListGroup>
```

martinpitt: Hah, thank you! That was a bit confusing in the docs.

garrett:
It's possible to do that with flex or probably even just applying `display: inline-block` if each time (plus comma) is in an element.

I'll take a look at this, but it'll have to wait until tomorrow.

martinpitt:
@garrett : Do you think we should block this on the "full" solution, or is that an incremental improvement worth landing? (I think it looks a bit better with this, so at least not split the number and unit in the middle)

garrett (approved):
@martinpitt: Let's land this, as it improves things in our release. We can further enhance it later.

Normally, I wouldn't like much churn in the UI, but this is rather minor and basically changes wrapping behavior in smaller sizes. It isn't rewording strings or really changing the UI in a significant manner, so I think this is fine to improve incrementally.

PR #16869, opened 2022-01-24
test: In the year 2037...

marusak:
This is some very odd problem with new systemd that it does not list
boots after 14. January 2037.

See https://github.com/cockpit-project/bots/pull/2828#issuecomment-1019929426


martinpitt (approved):
Thanks. 2038-01-01 very much ought to work (and on 64 bit systems, way beyond that of course), but indeed let's not waste time on this.

PR #16871, opened 2022-01-24
autogen.sh: add 'rpm' option

allisonkarlitskaya:
When run as `./autogen.sh rpm` then configure the package in exactly the same way as it would be configured by RPM for a build on this system.

@martinpitt mentioned wanting to have something like this.  To be honest, I'm not crazy about putting this in `./autogen.sh` and would like to find a better place for it.

 - [x] #16870

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -8,6 +8,18 @@ srcdir="$(realpath -m "$0"/..)"
 [ -n "${NOCONFIGURE:-}" ] && exit
 
 case "${1:-}" in
+    mock)
+        # broken configuration with as few deps as possible
+        # mostly only useful for running `make dep`
+        exec ./autogen.sh \
+            CPPFLAGS=-Itools/mock-build-env \
+            PKG_CONFIG_PATH=tools/mock-build-env \
+            --enable-prefix-only \
+            --disable-pcp \
+            --disable-polkit \
+            --disable-ssh \
+        ;;
+
     rpm)
         # configure with the same flags as when building an RPM
         mkdir -p tmp/rpmbuild/SPECS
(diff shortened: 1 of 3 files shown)


martinpitt:
Happy to discuss this further, it's not urgent. It just seemed we want something like this if your goal is to do incremental in-tree builds "exactly like Fedora builds"? We could also put this into tools/autogen-fedora or so?

Why does this block on the SELinux build changes?

allisonkarlitskaya:
> Happy to discuss this further, it's not urgent. It just seemed we want something like this if your goal is to do incremental in-tree builds "exactly like Fedora builds"? We could also put this into tools/autogen-fedora or so?

I think I'd prefer a separate script, yes.  I think I'd prefer to keep the name based on "rpm" though since it's theoretically possible that you could also do this on a rhel host.


> Why does this block on the SELinux build changes?

This depended on getting the extra stuff out of the %build rule so that we could short-circuit %make_build in order to get "just configure".

todo: add tests to make sure it stays that way :)

allisonkarlitskaya:
eh, I decided that it's actually an OK fit in `autogen.sh` and decided to add another long-overdue feature there.

autogen.sh
```diff
@@ -5,5 +5,16 @@ set -eu
 srcdir="$(realpath -m "$0"/..)"
 
 (cd "${srcdir}" && autoreconf -is --warnings obsolete)
+[ -n "${NOCONFIGURE:-}" ] && exit
```
martinpitt, line 8: This syntax still gives me the creeps.. But it does seem to work with `set -e`. Yay shell..

(and if it does fail, we could invert to `[ -z ] || `)

martinpitt (approved):
Nice and simple! I like the extra tweak of factorizing the "minimal build" hack.

PR #16872, opened 2022-01-24
Revert "build: add -ffile-prefix-map to CFLAGS"

martinpitt:
Commit c66a65feacf4 caused Fedora koji builds to break on finding the 
debug symbols. Revert it for now. Let's figure this out after the 
release: Debian builds probably already supply the flag as part of their
reproducible build efforts, and we can put it back into the .spec file
when/where needed.

Also revert the corresponding unit test from commit 59e173100.

---

Supposed to fix [these koji failures](https://kojipkgs.fedoraproject.org//work/tasks/5249/81755249/build.log) which broke the release. I did a [scratch build with a clean revert](https://koji.fedoraproject.org/koji/taskinfo?taskID=81758410) which succeeded.

We don't see this upstream because this only fails on all arches in [Rawhide](https://koji.fedoraproject.org/koji/taskinfo?taskID=81754804) but in [Fedora 34](https://koji.fedoraproject.org/koji/taskinfo?taskID=81755150) it works on x86_64, i686, and aarch64, but fails on the three others.

### DIFF
--- a/containers/unit-tests/scenario/distcheck
+++ b/containers/unit-tests/scenario/distcheck
@@ -27,21 +27,7 @@ cd _distcleancheck/cockpit-*
 ./configure
 make distclean
 ./configure
-make -j$(nproc)
-make install DESTDIR=destdir
 make check 2>&1 || {
     find -name test-suite.log | xargs cat
     exit 1
 }
-cd ../..
-
-# do another build in a different directory and ensure that it produces the
-# same output (in the name of reproducible builds)
-mkdir _otherdir
-tar -C _otherdir -xf cockpit-[0-9]*.tar.xz
-cd _otherdir/cockpit-*
-./configure
-make -j$(nproc)
-make install DESTDIR=destdir
-cd ../..
-diff --no-dereference -ur _distcleancheck/*/destdir _otherdir/*/destdir
--- a/configure.ac
+++ b/configure.ac
@@ -247,14 +247,6 @@ AC_DEFINE_UNQUOTED([PATH_SSH_AGENT], ["$SSH_AGENT"], [Location of ssh-agent bina
 
 changequote(,)dnl
 if test "x$GCC" = "xyes"; then
-  if  grep -q platform:el8 /usr/lib/os-release; then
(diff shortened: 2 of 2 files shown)


martinpitt:
I talked this over with @allisonkarlitskaya , and moved to a clean revert.

allisonkarlitskaya (dismissed):
Thanks.

Let's see how this impacts the debian reproducibility report after the next release.  If that goes green, then I think we can take a hint and add something similar in our spec.

martinpitt:
Ah, we [need to revert the corresponding unit test](https://github.com/cockpit-project/cockpit/runs/4922220097?check_suite_focus=true) as well.

allisonkarlitskaya:
> Ah, we [need to revert the corresponding unit test](https://github.com/cockpit-project/cockpit/runs/4922220097?check_suite_focus=true) as well.

Oh.  Right.  boo :(

containers/unit-tests/scenario/distcheck
```diff
@@ -27,21 +27,7 @@ cd _distcleancheck/cockpit-*
...
-# do another build in a different directory and ensure that it produces the
-# same output (in the name of reproducible builds)
-mkdir _otherdir
-tar -C _otherdir -xf cockpit-[0-9]*.tar.xz
-cd _otherdir/cockpit-*
-./configure
-make -j$(nproc)
-make install DESTDIR=destdir
-cd ../..
-diff --no-dereference -ur _distcleancheck/*/destdir _otherdir/*/destdir
```
allisonkarlitskaya, line 47: Sad to see this go.  Let's try and figure out another way to bring it back later, possibly by passing the necessary CFLAGS as part of the `./configure` invocation.

allisonkarlitskaya (approved):

containers/unit-tests/scenario/distcheck
martinpitt: Right, that's a good idea. Once we figure out how Debian/Fedora builds already handle this, we can put that back here.

PR #16875, opened 2022-01-25
packit: Run package build and unit tests on i386 and s390x

martinpitt:
This introduces some architecture variety; so far we have only tested on
x86_64, and thus were missing issues about `int` size, endianess, or RPM
builds on different architectures.

----

This should hopefully prevent issues like yesterday's release failure -- see PPR #16872

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -18,3 +18,13 @@ jobs:
       - fedora-35
       - centos-stream-8-x86_64
       - centos-stream-9-x86_64
+
+  # run build/unit tests on some interesting architectures
+  - job: copr_build
+    trigger: pull_request
+    metadata:
+      targets:
+        # 32 bit
+        - fedora-development-i386
+        # big-endian
+        - fedora-development-s390x


packit.yaml
```diff
@@ -18,3 +18,13 @@ jobs:
...
       - centos-stream-8-x86_64
       - centos-stream-9-x86_64
+
+  # run build/unit tests on some interesting architectures
+  - job: copr_build
+    trigger: pull_request
+    metadata:
+      targets:
+        # 32 bit
+        - fedora-development-i386
```
martinpitt, line 28: :thinking: Hmm, should we use fedora-stable-* rather? Rawhide is more likely to fail and thus catch relevant problems earlier -- at the same time exactly that can also be annoying. I just wouldn't want to waste *too* many cycles on building everywhere.

martinpitt: Perhaps, let's start with rawhide, and once it gets in the way we have a look and consider switching to stable?

marusak: Sure, it is fine to use rawhide unless it becomes too broken too often.
Don't we want to also run integration tests?

martinpitt: s390x (and 32 bit arm) are emulated architectures, running browser tests is way too slow there. We could in principle run them on i386, not sure how well browsers are actually supported there. Fedora doesn't support 32 bit desktop any more, but let's try in #16876 just to get the data.

Also, I think unit tests should be enough -- JavaScript isn't really architecture sensitive. We could of course discover 32 bit bugs in the Linux stack.

martinpitt: Running testing-farm tests on i386 [is not a thing](https://github.com/cockpit-project/cockpit/pull/16876/checks?check_run_id=4934762122), so I closed #16876.

marusak (approved):

allisonkarlitskaya:
I love this!  I was considering to set up a nightly Koji scratch build off of main to catch issues before they become release day headaches, but this is way cooler.  Thanks!

PR #16877, opened 2022-01-25
system: Improve terminal toolbar on mobile

mvollmer:
The existing test ignores almost everything and is thus not very
useful.  The ".xterm-cursor-layer" is probably intended to catch the
cursor only, but does in fact covers the whole terminal area.

Also, the exact scrollbar position seems to vary, which has started to
cause random failures.

Thus, let's test the header and actual text area (without scrollbar)
separately.

### DIFF
--- a/pkg/systemd/terminal.jsx
+++ b/pkg/systemd/terminal.jsx
@@ -5,7 +5,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import {
     FormSelect, FormSelectOption, NumberInput,
-    Toolbar, ToolbarContent, ToolbarItem
+    Toolbar, ToolbarContent, ToolbarItem, ToolbarGroup
 } from "@patternfly/react-core";
 
 import { Terminal } from "cockpit-components-terminal.jsx";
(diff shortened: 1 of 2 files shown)


mvollmer:
New reference images: https://github.com/cockpit-project/pixel-test-reference/compare/dba18567f57691d3e0e4db3f3c1d5789f7a7853c..befb7a1dfeefa8d781328fba6ac2144633b0a006

marusak:
This is probably one of the places where we should notice that the screenshot looks bad and fix it before we commit to testing it? :) (mobile header)

mvollmer:
> This is probably one of the places where we should notice that the screenshot looks bad and fix it before we commit to testing it? :) (mobile header)

No objections :-)

marusak:
> Group terminal toolbar items

And also maybe space between lines when it wraps?

mvollmer:
> > Group terminal toolbar items
> 
> And also maybe space between lines when it wraps?

That would be a Patternfly bug, I guess: https://www.patternfly.org/v4/components/toolbar

marusak (approved):
Thanks!

PR #16878, opened 2022-01-25
test: Give starting kdump.service more time in TestKDumpNFS

martinpitt:
Enabling kdump for the first time rebuilds the initrd, which is
inherently slow. On busy CI machines it sometimes takes more than a
minute.

The `TestKDump` class already does this by bumping the global timeout.
Be more specific here, to avoid hiding UI issues for operations that
should be fast.

----

[example 1](https://logs.cockpit-project.org/logs/pull-16686-20220125-102735-bf8b3591-rhel-8-6/log.html), [example 2](https://logs.cockpit-project.org/logs/pull-16826-20220113-155723-e5f39086-rhel-8-6/log.html#48), [example 3](https://logs.cockpit-project.org/logs/pull-16779-20220105-090740-28f20c62-centos-8-stream/log.html#48). I found some more on our logs server, but these should be sufficient to prove the issue. :grinning:


marusak (approved):
Thanks!

PR #16879, opened 2022-01-25
 test: add tests for working branding 

allisonkarlitskaya:
Try to check if we're being served the expected branding (CSS, logo,
favicon) correctly.  Make sure the variable substitutions are working
properly.

Add a pixel test for the login page to make sure it's visually correct.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -966,6 +966,54 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
         m.write("/etc/cockpit/ws-certs.d/0-self-signed-ca.pem", "FAKE CERT FOR TESTING\n")
         self.assertEqual(m.execute("curl -sfS http://localhost:9090/ca.cer"), "FAKE CERT FOR TESTING\n")
 
+    def test_branding(self):
+        m = self.machine
+        m.start_cockpit()
+
+        # for all of our CI images, the part before the dash is the name of the
+        # subdirectory in /usr/share/cockpit/branding.
+        brand = m.image.split('-')[0]
+        branddir = f'/usr/share/cockpit/branding/{brand}'
+
+        # We don't have access to the stuff on the filesystem if it's living in
+        # the container.
+        if not m.ostree_image:
+            # make sure that there are no broken links in "our" directory
+            self.assertEqual(m.execute(f'find {branddir} -xtype l'), '')
+
+            # branding.css undergoes variable substitution based on the content of
+            # /usr/lib/os-release.  Perform the substitution for ourselves for
+            # validation.  envsubst comes from gettext.
+            os_release_vars = m.execute("cat /usr/lib/os-release").replace('\n', ' ')
+            m.execute(f'{os_release_vars} envsubst < {branddir}/branding.css > /tmp/branding.ref')
+
+        # fetch some files and make sure they match what we expect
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Mobile is broken for missing screenshot.

The `fedora-coreos` is more interesting: it's failing to find the branding stuff on the VM system because it's in the container.  Checking in the container fails because `find` isn't available there, but pushing past that (using `tar ch > /dev/null`) hits a bigger problem: the container is missing the branding assets for Fedora, despite announcing itself as being Fedora-based (ie: we get "Fedora" in the title, but the default cockpit icon).

I guess we should install the icon in the container.... maybe also `find` :wink:

test/verify/check-connection
```diff
@@ -966,6 +966,51 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
...
+    def test_branding(self):
+        m = self.machine
+
+        # for all of our CI images, the part before the dash is the name of the
+        # subdirectory in /usr/share/cockpit/branding.
+        brand = m.image.split('-')[0]
+        branddir = f'/usr/share/cockpit/branding/{brand}'
+
+        # make sure that there are no broken links in "our" directory
+        self.assertFalse(m.execute(f'find {branddir} -xtype l'))
```
martinpitt, line 978: Please append a `.strip()` and use `assertEqual` with `""`, so that we get useful assertion failures.

test/verify/check-connection
```diff
@@ -966,6 +966,51 @@ until pgrep -f '^/usr/bin/cockpit-bridge.*--privileged'; do sleep 1; done
...
+            # Check that we can fetch the file
+            m.execute(f'curl --fail -o /tmp/{name} {url}')
+
+            # compare that it matches what we expected
+            m.execute(f'cmp /tmp/{name} {reference}')
+
+        m.start_cockpit()
+
+        # some brands miss the images, but the OSes we CI have them all
+        curl_and_compare('branding.css', 'text/css', reference='/tmp/branding.ref')
```
martinpitt, line 1004: I'm honestly not a fan about this. It's prodding in internal implementation details, and integration tests ought to test things that are visible/relevant to the user. Like the pixel test, which already makes sure that the logo works.

No veto, but unnecessary complexity IMHO.

martinpitt (changes requested):
Thanks!

I don't think the container itself is supposed to have Fedora branding. I mean, it sort of does, we install the cockpit-ws package and thus all the /usr/share/cockpit/branding/ bits, but they wouldn't directly apply as the non-default symlink targets are not present in the container. containers/bastion/install.sh makes some effort to redirect the brand detection/os-release to the host (which is mounted at /host), but apparently that is not working.

test/verify/check-connection
allisonkarlitskaya: i originally had assertEqual.  Will put it back :)

test/verify/check-connection
allisonkarlitskaya: The main reason I like these "other" tests is that they run on other OSes, whereas pixel testing only makes sure we didn't break Fedora.

For something that's very specifically designed to be different on different OSes, I think this is worth testing.  We might find out on image udpate that some png that we were using for the logo has disappeared from the OS, for example, which is useful information.

martinpitt: Right, fair enough.

martinpitt (approved):
Thank you!

PR #16880, opened 2022-01-25
Quiesce coverity warning

marusak:


src/ws/test-webservice.c
```diff
@@ -449,7 +449,7 @@ close_client_and_stop_web_service (TestCase *test,
   timeout = g_timeout_add_seconds (20, on_timeout_fail, "closing web service");
   g_object_add_weak_pointer (G_OBJECT (test->service), (gpointer *)&test->service);
   g_object_unref (test->service);
-  while (test->service != NULL)
+  while (test->service != NULL) // check_after_deref: Null-checking "test->service" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
```
marusak, line 452: @allisonkarlitskaya WDYT of this report? I am not sure what it means so I just wanted to raise it with someone who understands this.

src/pam-ssh-add/pam-ssh-add.c
```diff
@@ -815,6 +815,7 @@ stash_password_for_session (pam_handle_t *pamh,
       return PAM_AUTHTOK_RECOVER_ERR;
     }
 
+  free_password (password_copy);
```
martinpitt, line 818: It feels like I just touched that code.. commit ce1c6f91ecd1


I think this is wrong -- `pam_set_data` "consumes" the copy on success (i.e. it holds a reference to it and defines its lifetime), hence the need for passing the free function. Like this it's access to freed memory. See [manpage](https://linux.die.net/man/3/pam_set_data)

If that wasn't so, we would not need to bother with copying `password` in the first place. But `pam_set_data()` isn't in a position to copy the pointer itself, as it doesn't know its size or structure.

So this needs an annotation instead.

src/ws/test-webservice.c
martinpitt: This coverity warning is wrong. Of course test->service is not NULL *initially*, but the point is that the expected callbacks eventually unref it and thus cause it to get set *to* NULL. This happens asynchronously (well, not really, but you can think about it that way). If that were an `if`, coverity would be right, but apparnetly it doesn't understand this `while` loop and that test->service is mutated by the main loop.

martinpitt (changes requested):

src/pam-ssh-add/pam-ssh-add.c
marusak: > So this needs an annotation instead.

Annotation?

martinpitt: Like in commit 9d48655bc3 , see e.g. [stackoverflow](https://stackoverflow.com/questions/69596188/how-can-i-disable-coverity-checking-using-code-annotation).

martinpitt (approved):
Thanks!

marusak:
centos-stream 8 failed on `couldn't create runtime dir: /run/user/1001: Permission denied` and centos-stream-9 on `Command 'tuned-adm auto_profile' returned non-zero exit status 1.` Both unrelated, but both should be looked at.

martinpitt:
@marusak : Whoops! the c9s atomic thing is fallout from https://github.com/cockpit-project/bots/pull/2852 , I'll put it back.

PR #16881, opened 2022-01-26
argv robustifications

martinpitt:
In the light of [pwnkit](https://blog.qualys.com/vulnerabilities-threat-research/2022/01/25/pwnkit-local-privilege-escalation-vulnerability-discovered-in-polkits-pkexec-cve-2021-4034) I audited all our usage of `argv`. Fortunately I didn't find anything important, but a few cleanups and robustifications can't hurt.

### DIFF
--- a/src/bridge/bridge.c
+++ b/src/bridge/bridge.c
@@ -175,6 +175,8 @@ start_helper_process (const gchar * const  *argv,
                       const gchar          *socket_pattern,
                       const gchar          *socket_envvar)
 {
+  g_return_val_if_fail (argv[0] != NULL, NULL);
+
   {
     const gchar *env = g_getenv (socket_envvar);
     if (env && env[0])
--- a/src/ws/cockpitauth.c
+++ b/src/ws/cockpitauth.c
@@ -487,6 +487,8 @@ session_start_process (const gchar **argv,
   GPid pid = 0;
   int fds[2];
 
+  g_return_val_if_fail (argv[0] != NULL, NULL);
+
   g_debug ("spawning %s", argv[0]);
 
   /* The main stdin/stdout for the socket ... both are read/writable */
--- a/src/bridge/cockpitpipechannel.c
+++ b/src/bridge/cockpitpipechannel.c
@@ -440,6 +440,12 @@ cockpit_pipe_channel_prepare (CockpitChannel *channel)
         flags = COCKPIT_PIPE_STDERR_TO_MEMORY;
 
       self->name = g_strdup (argv[0]);
+      if (!self->name)
+        {
+          cockpit_channel_fail (channel, "protocol-error",
+                                "\"spawn\" array must not be empty");
+          goto out;
+        }
       if (!cockpit_json_get_string (options, "directory", NULL, &dir))
         {
           cockpit_channel_fail (channel, "protocol-error",
--- a/src/bridge/test-pipe-channel.c
+++ b/src/bridge/test-pipe-channel.c
@@ -599,6 +599,38 @@ test_spawn_pty (void)
   g_object_unref (transport);
 }
 
+static void
+test_spawn_errors (void)
+{
+  MockTransport *transport;
+  g_autofree gchar *problem = NULL;
+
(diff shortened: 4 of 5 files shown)


allisonkarlitskaya (approved):
These all look good, thanks.

One minor comment is that, *rein theoretisch*, we shouldn't technically rely on `assert` for security critical checks...

martinpitt:
> we shouldn't technically rely on assert for security critical checks

Fully agreed. I only added these as second-level defence, but tried to do "proper" error returns in all places where the argv initially gets passed/processed. Did I miss a spot?

[c9s packit failure](https://github.com/cockpit-project/cockpit/pull/16881/checks?check_run_id=4948830903) is due to https://github.com/cockpit-project/bots/pull/2853 , [c8s packit failure](https://github.com/cockpit-project/cockpit/pull/16881/checks?check_run_id=4948827530) occasionally happens -- needs to be tracked down (I can't reproduce it locally with tmt), but it has happened for a while already, so unrelated.

allisonkarlitskaya:
> Fully agreed. I only added these as second-level defence, but tried to do "proper" error returns in all places where the argv initially gets passed/processed. Did I miss a spot?

If you're making a belt and suspenders argument, then you should definitely harden these to real checks.

In fact, I might argue that someone hacking this code in the future might see the assert and then make assumptions about something being safe.  On the other hand, any system integrator who is building security-critical components with assertions disabled is playing with fire and might well expect to get burnt...

But here's a tie-breaker: assertion fails and SIGABRT in general should be reserved for situations where there is a programming fault in the program itself (or some library, or the kernel).  It should not be possible to trigger assertion failures from outside.  From that standpoint I'd say: change it.

martinpitt:
> in general should be reserved for situations where there is a programming fault in the program itself

Fully agreed, that was exactly my intention. For example, the one in start_helper_process() is such a case: all invocations are right inside the file itself, with static argv vectors. Whereas the ones in cockpitpipechannel or the shell in session are user-controlled and are proper errors. Hence my question whether you saw anything that I missed.

src/bridge/bridge.c
```diff
@@ -181,6 +181,8 @@ start_helper_process (const gchar * const  *argv,
       return NULL;
   }
 
+  g_assert (argv[0] != NULL);
```
allisonkarlitskaya, line 184: This reads better as a `g_return_val_if_fail()` at the top.

src/ws/cockpitauth.c
```diff
@@ -487,6 +487,8 @@ session_start_process (const gchar **argv,
   GPid pid = 0;
   int fds[2];
 
+  g_assert (argv[0] != NULL);
```
allisonkarlitskaya, line 490: ditto

allisonkarlitskaya (changes requested):
I think I misunderstood what this patch was doing.

The most obvious target, `cockpit-session`, has this for example:

```
    program_name = basename (argv[0]);
```

before any checking.  That definitely needs to be fixed.

Most other things are using `g_option_context_parse()`, which should hopefully be fine.

allisonkarlitskaya (dismissed):

src/bridge/bridge.c
martinpitt: That's a trade-off then. AFAIUI `g_return_val_if_fail()` would still apply even with building with `NDEBUG` (which would filter out the asserts). However, it specifically does *not* crash the program, but continues with just a critical warning. I actually *do* want to crash the program, so that our tests have a very specific failure in the sense of "you called start_helper_process() with a bogus argument". Our tests also fail on criticals, but you'd have to look a lot harder to find the root cause.

allisonkarlitskaya: GLib standard approach to this says to use g_return_val_if_fail() to check preconditions on function calls, and turn on abort-on-critical from your tests.

martinpitt: Updated.

src/ws/cockpitauth.c
martinpitt: Dito :grinning:

allisonkarlitskaya (approved):

PR #16883, opened 2022-01-26
Fix certificate filenames for debug calls

leomoty:
While trying to capture those `authorize` messages, I enabled `DEBUG 1` in `utils.h` and got greeted by compilation errors.


martinpitt (approved):
Thank you! That was some fallout from shifting the code around, sorry.

PR #16884, opened 2022-01-26
bridge: Search for override.json in all XDG_DATA_DIRS

### LINKED ISSUE #9299: User manifest override.json for system packages
I'd like to disable the terminal.  Possibly just for certain users, but even completely would be fine.  At the moment I have to edit /usr/share/cockpit/systemd/manifest.json to remove it.

martinpitt:
So far, bridge has only looked for override.json right next to
manifest.json. But in practically all real-world cases these are going
to override rpm/deb packaged manifests in /usr/share/cockpit. But we
should not recommend that admins place files there, and with e.g. OSTree
that's not even physically possible.

So make overrides behave like desktop or systemd unit files, and search
for them in all previously visited and the currently visited
XDG_USER_DIR/XDG_DATA_DIRS directories. This ensures that e.g.
~/.share/cockpit can override things in /usr/share/cockpit, but not vice
versa.

Adjust the documentation to stop recommending override files in
/usr/share, but use /usr/local/ or the user dir instead.

Fixes #9299


martinpitt:
TF/C9S is a [failure](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-16884/centos-stream-9-x86_64/03255697-cockpit/builder-live.log.gz) outside of our control:
```
  - nothing provides libgomp = 11.2.1-7.6.el9 needed by gcc-11.2.1-7.6.el9.x86_64
  - nothing provides libgcc >= 11.2.1-7.6.el9 needed by gcc-11.2.1-7.6.el9.x86_64
```
Ignoring.

allisonkarlitskaya (changes requested):
This is actually config.  Let's take a page from systemd and look for overrides in /etc.

martinpitt:
@allisonkarlitskaya : That's a nice idea. We have to continue to support the data dirs, as that functionality exists today and it's also conceivable that packages ship overrides. But of course it could check /etc/cockpit/page/override.json and ~/.config/cockpit/page/override.json in addition. This would be a new commit and separate functionality -- do you want to block this PR on that, or follow-up?

allisonkarlitskaya:
I'm not sure the functionality here is correct in terms of allowing overrides from different data directories.  I think it should be the same data directory or a "higher" config directory.

jelly:
Unrelated should `./tools/webpack-make -r c` learn about XDG_DATA_DIRS?

martinpitt:
Huh, [how did I break that](https://logs.cockpit-project.org/logs/pull-16884-20220128-095734-b87b01b4-fedora-35/log.html#264), but it's definitively related. Investigating

src/bridge/cockpitpackages.c
```diff
@@ -374,12 +374,30 @@ expand_libexec (const gchar *variable,
...
 
+static void
+apply_override (JsonObject *manifest,
+                const char *directory,
+                const char *filename)
+{
+  g_autoptr(GError) error = NULL;
+
+  g_autoptr(JsonObject) override = read_json_file (directory, filename, &error);
+  if (error)
```
allisonkarlitskaya, line 385: I know you're just copying this, but the normal paradigm is to check the return value.

src/bridge/cockpitpackages.c
```diff
@@ -374,12 +374,30 @@ expand_libexec (const gchar *variable,
...
+    {
+      if (g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        g_debug ("no override found in %s/%s", directory, filename);
+      else
+        g_message ("couldn't read %s/%s: %s", directory, filename, error->message);
+    }
+  if (override)
+    cockpit_json_patch (manifest, override);
```
allisonkarlitskaya, line 393: *shudders*

The one sole case of mutable JSON in the entire cockpit C codebase, and you're making it worse!

src/bridge/cockpitpackages.c
```diff
@@ -400,20 +418,14 @@ read_package_manifest (const gchar *directory,
...
-      if (override)
-        {
-          cockpit_json_patch (manifest, override);
-          json_object_unref (override);
-        }
+      /* possible override locations, in ascending priority */
+      apply_override (manifest, directory, "override.json");
+
+      g_autofree gchar *package_override_name = g_strconcat (package, ".override.json", NULL);
+      apply_override (manifest, PACKAGE_SYSCONF_DIR "/cockpit", package_override_name);
```
allisonkarlitskaya, line 425: We need to talk about this a bit, and maybe do a big cleanup: cockpitconf tries also to check the XDG_CONFIG_DIRS variable.  I'm okay with killing that off, for what it's worth, but we should at least be consistent.  There's even a function in there that returns the list of config directories.

src/bridge/cockpitpackages.c
```diff
@@ -400,20 +418,14 @@ read_package_manifest (const gchar *directory,
...
+
+      g_autofree gchar *cockpit_user_config = g_build_filename (g_get_user_config_dir (), "cockpit", NULL);
```
allisonkarlitskaya, line 427: I find the multiple layers of path pasting here to be a bit distasteful.  Let's modify `json_read_file()` to rather take a full pathname and perform all path computations here, or modify it to take a `const gchar **` and use compound literals (which would let us keep everything here `const`).  I think `...` will be too much effort, because there's no variant of the glib path building (or even `g_strjoin()`) which takes a va_args.

allisonkarlitskaya (changes requested):

src/bridge/cockpitpackages.c
martinpitt: How does this make it worse? It's the exact same call, just moved into a helper function?

martinpitt: I.e. how would you like to change it? This looks perfectly reasonable to me,  `JsonObject*` is mutable.

src/bridge/cockpitpackages.c
martinpitt: There is [g_build_filename_valist()](https://docs.gtk.org/glib/func.build_filename_valist.html). This has more runtime penalty and increases the amount of code, and IMHO this is not worth it for this case. Likewise, [g_build_filenamev()](https://docs.gtk.org/glib/func.build_filenamev.html) is overkill here -- single-time construction and iteration of whole arrays.

So I'll pull the g_build_filename() out of read_json_file(). It will make the read_package_manifest() a little bigger, but still readable and avoid unnecessary effort or unwieldy syntax.

src/bridge/cockpitpackages.c
martinpitt: Good point, fixed. As repentance I cleaned up read_json_file() some more.

src/bridge/cockpitpackages.c
martinpitt: Good point, thanks! Changed to use `cockpit_conf_get_dirs()`.

src/bridge/cockpitpackages.c
```diff
@@ -338,30 +338,16 @@ package_walk_directory (GChecksum *own_checksum,
...
 {
-  JsonObject *object;
-  GMappedFile *mapped;
-  gchar *filename;
-  GBytes *bytes;
-
-  filename = g_build_filename (directory, name, NULL);
-  mapped = g_mapped_file_new (filename, FALSE, error);
-  g_free (filename);
+  g_autoptr(GMappedFile) mapped = g_mapped_file_new (path, FALSE, error);
```
allisonkarlitskaya, line 345: "This temporarily makes the code more complicated" +9 -22 lol.

doc/guide/packages.xml
```diff
@@ -252,9 +252,33 @@ $ cockpit-bridge --packages
...
-    <para>A file called <filename>override.json</filename> may be placed next to the manifest.
-      containing overrides to the information in the manifest. These override files are in the
-      simple <ulink url="https://tools.ietf.org/html/rfc7386">JSON Merge Patch</ulink> format.</para>
+  </section>
+
+  <section id="package-manifest-override">
+    <title>Manifest overrides</title>
+
+    <para>A file called <filename>override.json</filename> may be placed next to the manifest, or a file
+      <filename>&lt;name&gt;.override.json</filename> may be placed into <filename>/etc/cockpit/</filename>
```
allisonkarlitskaya, line 261: One thing that could maybe use some clarification (since we complicate this): is this according to the package name (as listed in the manifest.json) or per the name of the containing directory?

doc/guide/packages.xml
```diff
@@ -252,9 +252,33 @@ $ cockpit-bridge --packages
...
+      or <filename>~/.config/cockpit</filename>.
+      These contain overrides to the information in the manifest in the simple
+      <ulink url="https://tools.ietf.org/html/rfc7386">JSON Merge Patch</ulink> format.</para>
+
+    <para>This can be used to hide or modify menu items of an existing package.
+      For example <filename>/etc/cockpit/systemd.override.json</filename> or
```
allisonkarlitskaya, line 267: . o O ( Ah okay.  So then it's the directory name. )

src/bridge/cockpitpackages.c
```diff
@@ -360,12 +361,31 @@ expand_libexec (const gchar *variable,
...
+  if (override)
+    {
+      cockpit_json_patch (manifest, override);
+    }
+  else
+    {
+      if (g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        g_debug ("no override found in %s", path);
+      else
+        g_message ("couldn't read %s: %s", path, error->message);
```
allisonkarlitskaya, line 380: I think that's a `g_warning()`, no?  Almost an `errx()` to be honest...

src/bridge/cockpitpackages.c
```diff
@@ -386,22 +406,21 @@ read_package_manifest (const gchar *directory,
...
-          g_clear_error (&error);
-        }
-      if (override)
+      /* possible override locations, in ascending priority */
+      g_autofree gchar *pkgdir_override = g_build_filename (directory, "override.json", NULL);
+      apply_override (manifest, pkgdir_override);
+
+      g_autofree gchar *package_override_name = g_strconcat (package, ".override.json", NULL);
+
+      for (const char * const *dir = cockpit_conf_get_dirs (); *dir; ++dir)
```
allisonkarlitskaya, line 415: This is non-idiomatic.

Something more like

```
const char * const *dirs = cockpit_conf_get_dirs ();
for (gint i = 0; dirs[i]; i++)
  ... dirs[i] ...
```
plz.

allisonkarlitskaya (changes requested):
"not even physically possible".  wow :)

Look for a package.override.json in in /etc/cockpit/,
*"in in"


This looks fantastic, thanks.

One outstanding question: what's the story for changing these files?  Now that they live in /etc, it's clear that this is more than just a static mechanism for distributions to use during packaging time, so we should up our game here a bit, or at least our documentation.  We support reloading the packages, right?And we'd reload the overrides at the time the packages are reloaded.  How is the checksum mechanism impacted by this?

src/bridge/cockpitpackages.c
martinpitt: But that's due to g_auto() -- I meant the duplication of g_build_filename() . But I figure that was just a joke anyway :grin:

martinpitt:
> We support reloading the packages, right? And we'd reload the overrides at the time the packages are reloaded.

Yes, that works. Tested manually, and the integration test uses that as well.

> How is the checksum mechanism impacted by this?

Honestly I don't know -- the code does differ between "system" and "user" packages, but manifest overrides  live somewhere in between. I haven't looked at the checksum design/logic at all so far.

src/bridge/cockpitpackages.c
martinpitt: Changed accordingly.

src/bridge/cockpitpackages.c
allisonkarlitskaya: I agree.  Path-pasting in a post-`g_autofree` world is pretty reasonable.

src/bridge/cockpitpackages.c
martinpitt: Agreed. Changed to g_warning() in a separate commit, as that was existing behaviour. errx() is too strong, it's not an unrecoverable or logic error, just user misconfiguration (most probably, inaccessible files)

doc/guide/packages.xml
martinpitt: I clarified the documentation for now to point out that it is the directory name, so that the PR is at least self-consistent. Overall this seems a little less confusing to me, but I'm open to the approaches above.

allisonkarlitskaya: I think the directory name is the correct approach.  I just think that it should be documented.  Saying "name" is a bit confusing, particularly because the property in the manifest is also called "name".

Incidentally: override files could actually rename packages, which is weird... but hey... you get to keep both pieces. :)

martinpitt:
> "not even physically possible". wow :)

Heh, that's too strong -- toned down the language.

> Look for a package.override.json in in /etc/cockpit/, *"in in"

Fixed.

There is still some outstanding discussion above, but I pushed the other cleanups.

doc/guide/packages.xml
martinpitt: From IRC discussion: yes, we want package directory here. The go-to use case for `name:` is "packagekit" and "ostree" both providing "name: updates", and the latter has a higher priority. We need to be clear which packae directory gets overridden.

martinpitt:
I did some investigations about the checksums.

## Bridge
Checksums are handled in [cockpitpackages.c](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/src/bridge/cockpitpackages.c). That has a lengthy comment about it.

In summary, these are a caching mechanism for the shell to decide when it has to reload the packages, mostly their actual contents (.js, .html and such). The `bundle_checksum` is a SHA across all files encountered in *system* (XDG_DATA_DIRS, i.e. /usr/share/cockpit and /usr/local/share/cockpit/):

```
# cockpit-bridge --packages
system               Overview, Logs, Services, Terminal       /usr/share/cockpit/systemd
[...]
checksum = b52848666e6626567d8150e23574238f164fd1c48987ec2c4073c1cb9e075644
```

With a "classical" override, the checksum changes, because the override is part of the system package directory:
```
# echo '{"menu": {"services": {"label": "Hackices"}}}' > /usr/share/cockpit/systemd/override.json
#  cockpit-bridge --packages
system               Overview, Logs, Hackices, Terminal       /usr/share/cockpit/systemd
[...]
checksum = eb89bbd9b603e57c9047c7884cc43b6ad4571ae06b70eb4be513d9f9b355455e
```

With an override outside of the package directory:
```
# echo '{"menu": {"services": {"label": "Hackices"}}}' > /etc/cockpit/systemd.override.json
```
the checksum does *not* change, you get the same "b52.." as in the first "pristine" output.

OTOH, with user packages (install examples/pinger into ~admin/.local/share/cockpit/), maybe_add_package() gets called with `system==FALSE` and [this code](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/src/bridge/cockpitpackages.c#L629) does not update the bundle checksum. Supposedly the idea is that user packages are always reloaded on refresh, without caching (I can't say why, this pre-dated me).

As soon as you have a single user package, the bridge does not report a bundle checksum any more:

```
$ cockpit-bridge --packages
[...]
pinger               Pinger                                   /home/admin/.local/share/cockpit/pinger
system               Overview, Logs, Services, Terminal       /usr/share/cockpit/systemd
##### ← *NO* checksum= here
```

This is a [conscious decision](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/src/bridge/cockpitpackages.c#L705), not a bug.

If a user merely has an override, the checksum also does not change, as the files in the system dir are not changed:
```
$ echo '{"menu": {"services": {"label": "Hackices"}}}' > .config/cockpit/systemd.override.json
$ cockpit-bridge --packages
[...]
system               Overview, Logs, Hackices, Terminal       /usr/share/cockpit/systemd
checksum = b52848666e6626567d8150e23574238f164fd1c48987ec2c4073c1cb9e075644
```
(again, same as the first "pristine" output).

The bridge [sends the checksum as part of its init message](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/src/bridge/bridge.c#L132) and  in an [ETag header](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/src/bridge/cockpitpackages.c#L859) when requesting manifests.json.

From there it lands in the `cockpit.manifests` object, both as a global property and a per-package property:

```
cockpit.manifests
Object { ".checksum": "b52848666e6626567d8150e23574238f164fd1c48987ec2c4073c1cb9e075644", users: {…}, static: {…}, system: {…}, ...}
> system: Object { name: "system", libexecdir: "/usr/libexec", ".checksum": "b52848666e6626567d8150e23574238f164fd1c48987ec2c4073c1cb9e075644", … }
```

So this clarifies how the checksums are computed and what they are from the bridge's POV: In summary, it does not do anything with it by itself, just sends this to the shell.

## Shell

The checksums are used in [lookup()](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/pkg/shell/base_index.js#L155) -- it constructs a per-machine URL which includes the checksum (for non-localhost anyway) *if* it has a checksum, i.e. there are no user packages. However, this *only* applies to the package *contents* at that point -- at this point it already knows about the manifests (and overrides).

The loading of the manifests themselves happens in [Loader.connect()](https://github.com/cockpit-project/cockpit/blob/9cbe71b3f3b25c17a66548c3610257ec007949a7/pkg/shell/machines/machines.js#L552). In particular, this shows that this also uses the checksum'ed URL if the manifests contain a checksum.

So all of this explains how it works for "localhost", as this is special-cased. But it does indicate that not updating the checksum might affect not picking up *remote* override changes.  I'll need to test that, and then update this comment.

martinpitt:
@allisonkarlitskaya : I added what I could figure out about the checksums in the previous commit. If you think it's better/more robust, I can certainly include the overrides into the checksum calculations?

allisonkarlitskaya:
> @allisonkarlitskaya : I added what I could figure out about the checksums in the previous commit. If you think it's better/more robust, I can certainly include the overrides into the checksum calculations?

I think my concern is that cockpit-ws tries to decide where to serve from based on the checksum given in the URL.  Consider this scenario:

 - we have machines a and b, running a given distro
 - base install of cockpit on this distro has checksum 123456.
 - we do an override on `a`, but the checksum doesn't change
 - we then connect also to `b`, but because `b` claims to have the same package checksum, `cockpit-ws` just asks `a` for the files to use with `b`, meaning `a`'s override also ends up being used with `b`, which might be unexpected.

I'm not sure if this is actually how it works, but I suspect it is, and that's what I'm worried about.

martinpitt:
@allisonkarlitskaya : Right, that's how I understood the code as well, and I tried to produce that scenario.

But, with the investigations above I am now relatively well convinced that it is at least not wrong to include the overrides into the checksum. It conceptually matches what the checksums are trying to do, it's not *really* strictly about the directory contents, but a checksum over "what is this package". So I'll add that.

martinpitt:
@allisonkarlitskaya : Reworked to include the overrides into the checksum.
```
❱❱❱ ./cockpit-bridge --packages
base1                                                         /usr/share/cockpit/base1
metrics                                                       /usr/share/cockpit/metrics
performance                                                   /usr/share/cockpit/tuned
shell                                                         /usr/share/cockpit/shell
ssh                                                           /usr/share/cockpit/ssh
system               Overview, Logs, Services, Terminal       /usr/share/cockpit/systemd
users                Accounts                                 /usr/share/cockpit/users
checksum = 4216bec4b77e196c82175fb7872640032a4b69fa40be928f79347de023e77744

❱❱❱  echo '{"menu": {"services": {"label": "Hackices"}}}' > ~/.config/cockpit/systemd.override.json
❱❱❱ ./cockpit-bridge --packages
base1                                                         /usr/share/cockpit/base1
metrics                                                       /usr/share/cockpit/metrics
performance                                                   /usr/share/cockpit/tuned
shell                                                         /usr/share/cockpit/shell
ssh                                                           /usr/share/cockpit/ssh
system               Overview, Logs, Hackices, Terminal       /usr/share/cockpit/systemd
users                Accounts                                 /usr/share/cockpit/users
checksum = 1e91604ca14db481a1f1918e6060041af4f1ad277fe716b6f38830b5a97ef483
```

src/bridge/cockpitpackages.c
```diff
@@ -600,7 +620,7 @@ maybe_add_package (GHashTable *listing,
 
   path = g_build_filename (parent, name, NULL);
 
-  manifest = read_package_manifest (path, name);
+  manifest = read_package_manifest (path, name, bundle_checksum);
```
allisonkarlitskaya, line 623: I think this isn't enough to get the desired effect: ~25-50 lines down from here is a section of code that calculates the `own_checksum` over the files in the directory (and only those files) and uses the existing bundle checksum in case nothing in that directory changed.  I'm pretty sure this would specifically prevent the checksum being updated in the case an override was added.

src/bridge/cockpitpackages.c
```diff
@@ -338,30 +339,26 @@ package_walk_directory (GChecksum *own_checksum,
...
+
+  g_autoptr(GBytes) bytes = g_mapped_file_get_bytes (mapped);
+
+  if (checksum)
     {
-      return NULL;
+      gsize contents_size;
+      const void *contents = g_bytes_get_data (bytes, &contents_size);
+      if (contents)
+        g_checksum_update (checksum, contents, contents_size);
```
allisonkarlitskaya, line 358: That's a bit different than the existing code does it.  It also hashes the filename, along with the content, and uses nuls to separate things.  I have trouble thinking about a case where this would make a difference, however.  On the other hand, it's also a bit uncouth (imho) to add +1 to the size you get out of a GBytes and assume that it's definitely nul terminated...

allisonkarlitskaya (commented):
I had a thought that maybe the easier thing to do is just to print and digest the final patched manifest into both checksums, somewhere around the call to `package_walk_directory()` in `maybe_add_package()`.  If you use `cockpit_json_write()` you also know that you definitely have a nul, and can include that in the digest for consistency.

While you're there, you could swap the hilarious (but irrelevant) incorrect order of the parameters to `package_walk_directory()`.

src/bridge/cockpitpackages.c
martinpitt: Well spotted, thanks! Changed to what you suggested.

src/bridge/cockpitpackages.c
martinpitt: I liked this better back then, but with the new approach it's back to the string checksum.

martinpitt:
Thanks @allisonkarlitskaya for your guidance! Reworked accordingly, it's indeed a bit easier now.

> swap the hilarious (but irrelevant) incorrect order of the parameters to package_walk_directory()

Done in a separate commit.

martinpitt:
@jelly :  do these [journal](https://logs.cockpit-project.org/logs/pull-16884-20220228-092006-bbfe53a5-debian-testing/log.html) [failures](https://logs.cockpit-project.org/logs/pull-16884-20220228-092006-bbfe53a5-arch/log.html) ring a bell? Seems related to the "slowly.." unit log changes in commit check-system-journal

marusak:
> do these [journal](https://logs.cockpit-project.org/logs/pull-16884-20220228-092006-bbfe53a5-debian-testing/log.html) [failures](https://logs.cockpit-project.org/logs/pull-16884-20220228-092006-bbfe53a5-arch/log.html) ring a bell? Seems related to the "slowly.." unit log changes in commit https://github.com/cockpit-project/cockpit/commit/a5a801a9a3965b895423dff3fd9d7735255339aa

I see exactly the same 3 OSes failing the same way - see https://github.com/cockpit-project/cockpit/pull/17049#issuecomment-1050669668

martinpitt:
@allisonkarlitskaya : not ready for landing yet (need to sort out that unrelated journal test), but good for re-review.

src/bridge/cockpitpackages.c
```diff
@@ -621,7 +621,7 @@ maybe_add_package (GHashTable *listing,
 
   if (bundle_checksum || paths)
     {
-      if (!package_walk_directory (bundle_checksum, own_checksum, paths, directory, NULL))
+      if (!package_walk_directory (directory, NULL, paths, bundle_checksum, own_checksum))
```
allisonkarlitskaya, line 645: Still wrong :rofl:

src/bridge/cockpitpackages.c
```diff
@@ -630,7 +651,15 @@ maybe_add_package (GHashTable *listing,
   directory = NULL;
 
   if (own_checksum)
-    package->own_checksum = g_strdup (g_checksum_get_string (own_checksum));
+    {
+       /* digest the whole final manifest, which may have overrides from external directories */
+      gsize manifest_len;
+      g_autofree gchar *manifest_str = cockpit_json_write_object (manifest, &manifest_len);
+      g_checksum_update (own_checksum, (guchar *) manifest_str, manifest_len);
```
allisonkarlitskaya, line 658: I'd still prefer to see this done a little bit more like the other files: possibly add `"manifest.json\0"` before, and definitely include the nul byte.

allisonkarlitskaya (changes requested):
Looks a lot more robust, thanks.  Now I need to update the Python bridge :)

Two small comments.

src/bridge/cockpitpackages.c
martinpitt: OK, now I need a little more input which argument order you would actually prefer :grin:

src/bridge/cockpitpackages.c
martinpitt: Does that make any practical difference? Adding constant strings into a checksum just feels like a waste of time. No strong opinion though, I can update this once we unbreak journal tests on main.

src/bridge/cockpitpackages.c
allisonkarlitskaya: The problem is (and was) the order of the "own" and "bundle" checksum arguments.

src/bridge/cockpitpackages.c
allisonkarlitskaya: Ya.  Fair enough.  this is always at the end, and we always do it exactly once.

src/bridge/cockpitpackages.c
martinpitt: In what sense is that anything but arbitrary? I.e. why does one should æsthetically come first?

Do you agree to the root/directory/paths change, or do you have a different order in mind there?

allisonkarlitskaya: The order in the declaration is `own_checksum` then `bundle_checksum`.

```
static gboolean package_walk_directory (const gchar *root,
                                        const gchar *directory,
                                        GHashTable *paths,
                                        GChecksum *own_checksum,
                                        GChecksum *bundle_checksum);
```

But at the call site, `bundle_checksum` comes first.

```
       if (!package_walk_directory (directory, NULL, paths, bundle_checksum, own_checksum)) 
```

allisonkarlitskaya: For what it's worth, in `bridge.py` I just pass an array here :)

martinpitt: *slaps forehead* of course :grin: Sorry for the misunderstanding. Fixed locally, will push after #17049 lands.

allisonkarlitskaya (approved):
Thanks :)

martinpitt:
"Let me just fix this quickly", and other lies developer will tell you... :grinning:

PR #16885, opened 2022-01-26
systemd: display the scheduled reboot/shutdown time

jelly:
In the health card in the system overview page we now show a scheduled
reboot/poweroff with the option to cancel it. The implementation does
not watch dbus property changes on org.freedesktop.login1.Manager as
they are not emitted when using `shutdown --reboot $timestamp`. For now
the file (/run/systemd/shutdown/poweroff) systemd creates when
requesting a shutdown is watched to detect scheduled shutdowns. The
systemd issue for dbus property changes is tracked here:

https://github.com/systemd/systemd/issues/22244

---

# Overview: Show scheduled shutdowns

The Health card shows a scheduled reboot or poweroff. This can also be cancelled.

image

### DIFF
--- a/pkg/systemd/overview-cards/shutdownStatus.jsx
+++ b/pkg/systemd/overview-cards/shutdownStatus.jsx
@@ -1,0 +1,111 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+import React, { useState, useEffect } from 'react';
+import { Button, Flex, FlexItem } from '@patternfly/react-core';
+import { PowerOffIcon, RedoIcon } from "@patternfly/react-icons";
+
+import * as timeformat from "timeformat";
+
+import cockpit from "cockpit";
+
+const _ = cockpit.gettext;
+
+const getScheduledShutdown = (setShutdownTime, setShutdownType) => {
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.DBus.Properties", "Get",
+                       ["org.freedesktop.login1.Manager", "ScheduledShutdown"], { type: "ss" })
+            .then(([result]) => {
+                setShutdownType(result.v[0]);
+                setShutdownTime(result.v[1]);
+            })
+            .catch(err => console.warn("Failed to get ScheduledShutdown property", err.toString()));
+};
+
+const cancelShutdownAction = () => {
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.login1.Manager", "CancelScheduledShutdown")
+            .then(([cancelled]) => {
+                if (!cancelled) {
+                    console.warn("Unable to cancel shutdown");
+                }
+            })
+            .catch(err => console.warn("Failed to cancel shutdown", err.toString()));
+};
+
+export const ShutDownStatus = () => {
+    const [shutdownType, setShutdownType] = useState(null);
+    const [shutdownTime, setShutdownTime] = useState(0);
+
+    useEffect(() => {
+        getScheduledShutdown(setShutdownTime, setShutdownType);
+        // logind does not have a propertieschanged mechanism https://github.com/systemd/systemd/issues/22244
+        cockpit.file("/run/systemd/shutdown/scheduled").watch(() => {
+            getScheduledShutdown(setShutdownTime, setShutdownType);
+        });
+    }, []);
+
+    // We only care about these two types
+    if (shutdownType !== "poweroff" && shutdownType !== "reboot") {
+        // don't log undefined
+        if (shutdownType !== null && shutdownType !== "") {
+            console.log(`unsupported shutdown type ${shutdownType}`);
+        }
+        return null;
+    }
+
+    const date = new Date(shutdownTime / 1000);
+    const now = new Date();
+    let displayDate = null;
+    if (date.getFullYear() == now.getFullYear()) {
+        displayDate = timeformat.dateTimeNoYear(date);
+    } else {
+        displayDate = timeformat.dateTime(date);
+    }
+
+    let text;
+    let cancelText;
+    let icon;
+    if (shutdownType === "poweroff") {
+        icon = <PowerOffIcon />;
+        text = _("Scheduled poweroff at $0");
+        cancelText = _("Cancel poweroff");
+    } else {
+        icon = <RedoIcon />;
+        text = _("Scheduled reboot at $0");
+        cancelText = _("Cancel reboot");
+    }
+
+    return (
+        <li id="system-health-shutdown-status">
+            <Flex flexWrap={{ default: 'nowrap' }}>
+                <FlexItem spacer={{ default: 'spacerSm' }}>{icon}</FlexItem>
+                <Flex id="system-health-shutdown-status-text" direction={{ default: 'column' }}>
+                    {cockpit.format(text, displayDate)}
+                    <FlexItem>
+                        <Button variant="link" isInline
+                                id="system-health-shutdown-status-cancel-btn"
+                                className="pf-u-font-size-sm"
+                                onClick={cancelShutdownAction}>
+                            {cancelText}
+                        </Button>
+                    </FlexItem>
+                </Flex>
+            </Flex>
+        </li>);
+};
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -822,6 +822,38 @@ password=foobar
         wait(lambda: progressValue(2) <= hog_usage - 15)
         self.assertGreater(progressValue(2), 10)
 
+    def testShutdownStatus(self):
+        m = self.machine
+        b = self.browser
+
+        self.login_and_go("/system")
+        b.wait_not_present("#system-health-shutdown-status")
+
+        # Schedule a reboot
+        m.execute("shutdown --reboot +10")
+        self.addCleanup(m.execute, "shutdown -c")
+
+        b.wait_in_text('#system-health-shutdown-status-text', "Scheduled reboot")
+
+        # Check that reloading still shows the reboot text
+        b.reload()
+        b.enter_page("/system")
+        b.wait_in_text('#system-health-shutdown-status-text', "Scheduled reboot")
+
+        # Cancel
+        b.click("#system-health-shutdown-status-cancel-btn")
+        b.wait_not_present('#system-health-shutdown-status')
+
+        # Schedule a poweroff
+        m.execute("shutdown --poweroff +10")
+        b.wait_in_text('#system-health-shutdown-status-text', "Scheduled poweroff")
+
+        # Cancel
+        b.click("#system-health-shutdown-status-cancel-btn")
+        b.wait_not_present('#system-health-shutdown-status')
+        dbus_call = 'busctl get-property org.freedesktop.login1 /org/freedesktop/login1  org.freedesktop.login1.Manager ScheduledShutdown'
+        self.assertEqual('(st) "" 0', m.execute(dbus_call).strip())
+
 
 if __name__ == '__main__':
     test_main()
--- a/pkg/systemd/overview-cards/healthCard.jsx
+++ b/pkg/systemd/overview-cards/healthCard.jsx
@@ -23,6 +23,7 @@ import { Card, CardBody, CardFooter, CardTitle } from '@patternfly/react-core';
 import cockpit from "cockpit";
 import { PageStatusNotifications } from "../page-status.jsx";
 import { InsightsStatus } from "./insights.jsx";
+import { ShutDownStatus } from "./shutdownStatus.jsx";
 import LastLogin from "./lastLogin.jsx";
 
 import "./healthCard.scss";
@@ -38,6 +39,7 @@ export class HealthCard extends React.Component {
                     <ul className="system-health-events">
                         <PageStatusNotifications />
                         <InsightsStatus />
+                        <ShutDownStatus />
                         <LastLogin />
                     </ul>
                 </CardBody>


garrett:
This looks great! :+1: 

As a follow-up, we might want to consider having relative times. Reboots and shutdowns are usually X minutes / X seconds from the current time, not calendar events. Plus, the time on the server can be different from the local time, so we really should do the math for our users in this case. 

But, as @jelly pointed out on IRC yesterday, we probably would want a relative time widget that handles the refreshing using [window.requestAnimationFrame()](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). [CSS-Tricks has a nice writeup on the function](https://css-tricks.com/using-requestanimationframe/) too. Additionally, we probably only want it updating only every unit; that is: if it's X minutes, then update every minute; if it's X seconds, update every second. Updating on the minute should probably be aligned to the computer's minute (but perhaps a second or so after, to prevent rounding errors).

I mentioned this all to @jelly on IRC yesterday, but am writing it here as to capture what we talked about.

---

Aside: We should do a follow-up to add a little more space between items and fix the extra space between the last successful login and the history link, so everything's consistent. But that's not for this PR; just commenting on the context as seen in this screenshot. (Previously, we didn't have so many items in the health card, especially ones with actions. This PR makes it obvious that we need to adjust things a little. :wink:)

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+export const ShutDownStatus = () => {
+    const [shutdownType, setShutdownType] = useState(null);
+    const [shutdownTime, setShutdownTime] = useState(0);
+
+    useEffect(() => {
+        if (shutdownType === null) {
+            getScheduledShutdown(setShutdownTime, setShutdownType);
+
+            // logind does not have a propertieschanged mechanism https://github.com/systemd/systemd/issues/22244
+            cockpit.file(systemd_file).watch(content => {
```
martinpitt, line 59: The constant is just being used once here, and has an overly generic name. Let's just put the actual value right here, so that the hack  is localized in one place.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+                getScheduledShutdown(setShutdownTime, setShutdownType);
+            });
+        }
+    }, [shutdownType]);
```
martinpitt, line 63: What happens if you change the scheduled time, but keep the type? Would that still treat this as a change and re-run the effect, even if it's getting set to the same value? It may be safer to not make that assumption, and re-run this on changes to `shutdownTime` as well.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+import * as timeformat from "timeformat";
+
+import cockpit from "cockpit";
+
+const _ = cockpit.gettext;
+const systemd_file = "/run/systemd/shutdown/scheduled";
+
+const getScheduledShutdown = (setShutdownTime, setShutdownType) => {
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.DBus.Properties", "Get", ["org.freedesktop.login1.Manager", "ScheduledShutdown"], { type: "ss" }).then(reply => {
```
martinpitt, line 32: Please break up the long line, and split the `.then()` into a new line in either case. This should also handle failures (console.warn about them) instead of crashing the page.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+        const result = reply[0];
```
martinpitt, line 33: This can be done in a bit less ugly way with structured decomposition: `.then(([type_v, time_v]) => { ... }`

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+        if (result) {
+            setShutdownType(result.v[0]);
+            setShutdownTime(result.v[1]);
+        }
+    });
+};
+
+const cancelShutdownAction = () => {
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.login1.Manager", "CancelScheduledShutdown").then(reply => {
```
martinpitt, line 43: Same thing -- `.then` on new line, structured decomposition, and handling failed calls.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+    });
+};
+
+export const ShutDownStatus = () => {
+    const [shutdownType, setShutdownType] = useState(null);
+    const [shutdownTime, setShutdownTime] = useState(0);
+
+    useEffect(() => {
+        if (shutdownType === null) {
```
martinpitt, line 55: Why do we need this condition?

pkg/systemd/overview-cards/shutdownStatus.jsx
martinpitt, line 59: This means that each re-run of the effect sets  up its own new file watch?

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+            // logind does not have a propertieschanged mechanism https://github.com/systemd/systemd/issues/22244
+            cockpit.file(systemd_file).watch(content => {
+                getScheduledShutdown(setShutdownTime, setShutdownType);
+            });
+        }
+    }, [shutdownType]);
+
+    // We only care about these two types
+    if (shutdownType !== "poweroff" && shutdownType !== "reboot") {
+        return null;
```
martinpitt, line 67: We don't really expect any others, do we? Please make this a bit stronger and console.warn() about unknown ones, so that this is easier to debug. Also "care about" → "know of"?

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    const date = new Date(shutdownTime / 1000);
+    const now = new Date();
+    let displayDate = null;
+    if (date.getFullYear() == now.getFullYear()) {
+        displayDate = timeformat.dateTimeNoYear(date);
+    } else {
+        displayDate = timeformat.dateTime(date);
+    }
```
martinpitt, line 77: You may want timeformat.distanceToNow() (but of course then you'd have to regularly refresh that -- let's perhaps rather do this in a follow-up).

test/verify/check-system-info
```diff
@@ -828,6 +828,36 @@ password=foobar
...
+        b.wait_in_text('#system-health-shutdown-status-text', "Scheduled reboot")
+
+        # Check that reloading still shows the reboot text
+        b.reload()
+        b.enter_page("/system")
+        b.wait_in_text('#system-health-shutdown-status-text', "Scheduled reboot")
+
+        # Cancel
+        b.click("#system-health-shutdown-status-cancel-btn")
+        b.wait_not_present('#system-health-shutdown-status')
```
martinpitt, line 851: Let's also validate with a busctl call that this really made it into systemd, and `ScheduledShutdown` property is back to 0.

martinpitt (changes requested):
Really nice feature, thank you! Some nitpicks, though.

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: There are others, such as "halt" which we don't support but maybe should show?

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: Let's do it in a follow up, I was considering if we should make that a component in cockpit.

pkg/systemd/overview-cards/shutdownStatus.jsx
martinpitt: Showing the additional ones can be a follow-up for sure. So maybe console.log() for now?

test/verify/check-system-info
jelly: Sure!

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: Yes, hmm that's also why I added the shutdownType guard so every re-render doesn't call useEffects.

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: Not sure how that can be done, reply is an array of which contains a one object. so we can't destrcuct it easily

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: useEffects is called a lot in react, so it seems worthwhile to not always call cockpit.file().watch() and call dbus.

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: Good point that's wrong, so moved it out of the if statement.

pkg/systemd/overview-cards/shutdownStatus.jsx
martinpitt: You can at least deconstruct the array, so so `.then(([result]) => {...}` instead of your `const`. You can also decompose objects, but indeed that's not too useful here.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,116 @@
...
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.DBus.Properties", "Get",
+                       ["org.freedesktop.login1.Manager", "ScheduledShutdown"], { type: "ss" })
+            .then(([result]) => {
+                if (result) {
+                    setShutdownType(result.v[0]);
+                    setShutdownTime(result.v[1]);
+                }
+            })
+            .catch(err => console.warn("Failed to get ScheduledShutdown property", err));
```
martinpitt, line 39: `err.toString()` please; the standard JS console or our test logs can't print objects, they just get rendered as `[Object]`.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,116 @@
...
+import cockpit from "cockpit";
+
+const _ = cockpit.gettext;
+
+const getScheduledShutdown = (setShutdownTime, setShutdownType) => {
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.DBus.Properties", "Get",
+                       ["org.freedesktop.login1.Manager", "ScheduledShutdown"], { type: "ss" })
+            .then(([result]) => {
+                if (result) {
```
martinpitt, line 34: Would this ever be false if the call succeeds? It's a struct, after all. If there's any doubt, let's please log it in `else`.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,116 @@
...
+                    setShutdownTime(result.v[1]);
+                }
+            })
+            .catch(err => console.warn("Failed to get ScheduledShutdown property", err));
+};
+
+const cancelShutdownAction = () => {
+    const client = cockpit.dbus("org.freedesktop.login1");
+    return client.call("/org/freedesktop/login1", "org.freedesktop.login1.Manager", "CancelScheduledShutdown")
+            .then(reply => {
```
martinpitt, line 45: replace `reply` with `([cancelled])` to make the return values more readable?

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,116 @@
...
+                if (!reply[0]) {
+                    console.error("Unable to cancel shutdown");
```
martinpitt, line 47: Warning is enough -- it's not a bug in our own code, "just" something unexpected from the OS which we can handle.

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,116 @@
...
+export const ShutDownStatus = () => {
+    const [shutdownType, setShutdownType] = useState(null);
+    const [shutdownTime, setShutdownTime] = useState(0);
+
+    useEffect(() => {
+        getScheduledShutdown(setShutdownTime, setShutdownType);
+
+        if (shutdownType === null) {
+            // logind does not have a propertieschanged mechanism https://github.com/systemd/systemd/issues/22244
+            cockpit.file("/run/systemd/shutdown/scheduled").watch(content => {
```
martinpitt, line 62: content is unused, `() =>` pelase; also, it's just a single call, so it can just be `=> handler()` without the {}

pkg/systemd/overview-cards/shutdownStatus.jsx
martinpitt: Still looks odd and asymmetric -- why not make the effect trigger list empty (so that it's effectively like `componentDidMount()`) and drop the `if`?

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,116 @@
...
+    let displayDate = null;
+    if (date.getFullYear() == now.getFullYear()) {
+        displayDate = timeformat.dateTimeNoYear(date);
+    } else {
+        displayDate = timeformat.dateTime(date);
+    }
+
+    let text = "";
+    let cancelText = "";
+    let icon = null;
```
martinpitt, line 88: No need to initialize these.

martinpitt (changes requested):
Thanks! a few more nitpicks, but very close!

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: I'm not sure what our api dictates here? So I guess the if is useless and can be removed.

```
The returned value is a Promise that will complete successfully when the method returns, or fail if an error occurs.
```

pkg/systemd/overview-cards/shutdownStatus.jsx
jelly: Dropped the if, react requires an "dependency" to be passed so I've passed [] and this is only called once so it seems fine.

martinpitt: Yes, `[]` is *significantly* different from `undefined` -- the former means "just run once", the latter "run on every render"

pkg/systemd/overview-cards/shutdownStatus.jsx
```diff
@@ -0,0 +1,111 @@
...
+        // logind does not have a propertieschanged mechanism https://github.com/systemd/systemd/issues/22244
+        cockpit.file("/run/systemd/shutdown/scheduled").watch(() => {
+            getScheduledShutdown(setShutdownTime, setShutdownType);
+        });
+    }, []);
+
+    // We only care about these two types
+    if (shutdownType !== "poweroff" && shutdownType !== "reboot") {
+        // don't log undefined
+        if (shutdownType !== null && shutdownType !== "") {
```
martinpitt, line 66: At this point you could just as well say `if (shutdownType)`. However, not important enough to re-push just for this.

martinpitt (approved):
Thank you! Happy to see this in the next release!

pkg/systemd/overview-cards/shutdownStatus.jsx
martinpitt: Exactly, I don't see how the response in a successful call would ever not be a valid struct (of course its *members* could be null, but the structure itself must be present.)

garrett:
We need to update the screenshot once https://github.com/cockpit-project/cockpit/pull/16900 lands.

PR #16890, opened 2022-01-27
Firewall page improvements

jelly:
Some UXResearch/general improvements for the firewall page.

---

# Add firewall service description

Custom firewall ports can now include a small description for the system administrator.

image


martinpitt (approved):
Nice, thank you! The tests are a bit weak: They only test the dialog, but not that the values *actually* made it to the system (`firewall-cmd` or reload the page, open the dialog again and check that the expected values are still there). However, the current tests all have that problem, I don't want to block this PR on that.

garrett:
Description and ID are backwards. ID should really be the first thing. Descriptions should always come after the names/iDs of things (in general).

PR #16891, opened 2022-01-28
Firewall improvements II

jelly:
Moves away from one of the deprecated end points and some test/code cleanups.

### DIFF
--- a/pkg/networkmanager/firewall-client.js
+++ b/pkg/networkmanager/firewall-client.js
@@ -253,15 +253,15 @@ function fetchServiceInfos(services) {
         let info;
         return firewalld_dbus.call('/org/fedoraproject/FirewallD1',
                                    'org.fedoraproject.FirewallD1',
-                                   'getServiceSettings', [service])
-                .then(reply => {
-                    const name = reply[0][1];
-                    const description = reply[0][2];
-                    const ports = reply[0][3];
+                                   'getServiceSettings2', [service])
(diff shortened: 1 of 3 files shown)


pkg/networkmanager/firewall-client.js
```diff
@@ -253,15 +253,16 @@ function fetchServiceInfos(services) {
...
-                                   'getServiceSettings', [service])
+                                   'getServiceSettings2', [service])
                 .then(reply => {
-                    const name = reply[0][1];
-                    const description = reply[0][2];
-                    const ports = reply[0][3];
+                    const obj = reply[0];
+                    const name = obj.short ? obj.short.v : "";
+                    const description = obj.description ? obj.description.v : "";
+                    const ports = obj.ports ? obj.ports.v : [];
```
martinpitt, line 261: This would be another excellent use case for structured decomposition?

    .then(([{ short, description, ports }]) => { ... }

?

martinpitt (dismissed):
Thanks! I'll leave it up to you if you want to try and clean up the reply handler, but good enough to land as-is.

pkg/networkmanager/firewall-client.js
jelly: I can squeeze that in.

jelly:
Handled https://github.com/cockpit-project/cockpit/pull/16890#issuecomment-1026860796 in this PR.

martinpitt (approved):
Thank you!

PR #16898, opened 2022-01-31
storage: Document interface revision policy of Stratis

mvollmer:
So that we don't forget to bump them all together at the same time
when the time comes.


martinpitt (approved):
Danke!

PR #16899, opened 2022-01-31
lib: Let webpack rsync plugin rsync to /usr/local/share/cockpit/

martinpitt:
Scribbling over /usr/share/cockpit/ is evil: This is package manager
territory, and it might not even be writable (such as on Fedora CoreOS).
Sync to /usr/local/share/cockpit/ instead, which is meant for custom
built code. This also makes it possible to quickly go back to the distro
version.

### DIFF
--- a/pkg/lib/cockpit-rsync-plugin.js
+++ b/pkg/lib/cockpit-rsync-plugin.js
@@ -6,13 +6,17 @@ module.exports = class {
             options = {};
         this.dest = options.dest || "";
         this.source = options.source || "dist/";
+
+        // ensure the target directory exists
+        if (process.env.RSYNC)
+            child_process.spawnSync("ssh", [process.env.RSYNC, "mkdir", "-p", "/usr/local/share/cockpit/"], { stdio: "inherit" });
     }
 
     apply(compiler) {
         compiler.hooks.afterEmit.tapAsync('WebpackHookPlugin', (compilation, callback) => {
             if (process.env.RSYNC) {
                 const proc = child_process.spawn("rsync", ["--recursive", "--info=PROGRESS2", "--delete",
-                    this.source, process.env.RSYNC + ":/usr/share/cockpit/" + this.dest], { stdio: "inherit" });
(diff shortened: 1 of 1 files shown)


pkg/lib/cockpit-rsync-plugin.js
```diff
@@ -6,13 +6,17 @@ module.exports = class {
             options = {};
         this.dest = options.dest || "";
         this.source = options.source || "dist/";
+
+        // ensure the target directory exists
+        if (process.env.RSYNC)
+            child_process.spawnSync("ssh", [process.env.RSYNC, "mkdir", "-p", "/usr/local/share/cockpit/"], { stdio: "inherit" });
```
martinpitt, line 12: This is ugly, but I didn't find an rsync option which would auto-create parent directories.

martinpitt:
I tested this with normal `image-prepare fedora-35` and `vm-run -s cockpit.socket`, then logged into cockpit, mutilated the kdump page, `tools/webpack-watch kdump -r c`, press Ctrl+Shift+R (reload) and I see the new page. So c-bridge behaves as expected, even for a running session.

pkg/lib/cockpit-rsync-plugin.js
allisonkarlitskaya: There have been a few other times where I thought it might be useful to have `dist/` be called `cockpit/` or at least to make a symlink from `cockpit -> dist`, for exactly this reason.

Do you think that might help you here?

allisonkarlitskaya (approved):
The change is fine, and I agree with the rational (both parts of it).

But: why does this plugin exist?  It seems like it would be some pretty trivial stuff to add to webpack-make, no?

martinpitt:
@allisonkarlitskaya : We want to use it in c-machines, -podman etc. as well, and we don't have webpack-make there.

allisonkarlitskaya:
> @allisonkarlitskaya : We want to use it in c-machines, -podman etc. as well, and we don't have webpack-make there.

Thanks for the explanation.

PR #16903, opened 2022-02-01
tools: simplify cockpit-appstream trickery

allisonkarlitskaya:
Instead of extracting the spec file from the tarball and patching it
before manually passing it to rpmbuild (and ensuring that we set
_sourcedir properly so that it can still find the tarball), let's just
add a macro which changes the package name if it's defined.

### DIFF
--- a/tools/make-srpm
+++ b/tools/make-srpm
@@ -51,12 +51,7 @@ rpmbuild "${rpmbuild_args[@]}" -ts ${source} | sed -n 's@^Wrote:.*/@@p'
 # "cockpit" with build_basic=1, and "cockpit-appstream" with build_optional=1
 # the spec has proper build_* defaults depending on the Name:
 if grep -q 'platform:el8' /usr/lib/os-release; then
-    appstream_spec="${tmpdir}/cockpit-appstream.spec"
-    tar xf "$source" -O '*/tools/cockpit.spec' |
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
I had the thought that this might fail to work because the flag won't be properly defined at binary package build time...

tools/make-srpm
```diff
@@ -51,12 +51,7 @@ rpmbuild "${rpmbuild_args[@]}" -ts ${source} | sed -n 's@^Wrote:.*/@@p'
 # "cockpit" with build_basic=1, and "cockpit-appstream" with build_optional=1
 # the spec has proper build_* defaults depending on the Name:
 if grep -q 'platform:el8' /usr/lib/os-release; then
-    appstream_spec="${tmpdir}/cockpit-appstream.spec"
-    tar xf "$source" -O '*/tools/cockpit.spec' |
-        sed '/^Name:/ s/$/-appstream/' > "${appstream_spec}"
```
martinpitt, line 56: FTR: I did it that way because it's pretty much exactly the process which we do downstream. I'm a bit nervous about this breaking our release scripts and places which sed/grep/scan the spec, but no veto if it works.

allisonkarlitskaya:
Ya, so this broke all over the place, and pretty much for exactly the reason I suspected it would.  Nice idea, but it's not gonna fly.

PR #16904, opened 2022-02-01
services: Fix user relationships

### LINKED ISSUE #16889: Details page for user-level systemd timer links to _system_-level service
Cockpit version: 260-1~bpo21.10.1+1
OS: Pop!_OS 21.10
Page: "Services"

The page that shows the details of a user-level timer links to a system-level service. Looking at the URL of the link, it seems to need the query parameter `owner=user`.

Steps to reproduce

1. Visit **Services**, then select the **User** option.
2. Select the **Timer** tab.
3. Select any timer to view its details.
4. Follow a link listed for **Triggers**.

Result: The error "Unit not found"

Note: If I add the query parameter `owner=user` to the URL in the address bar, then Cockpit shows the expected user-level service unit details.

tomasmatus:
Fixes #16889

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -473,7 +473,7 @@ export class ServiceDetails extends React.Component {
                             <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
                             <DescriptionListDescription id={rel.Name.split(" ").join("")}>
                                 <ul className="comma-list">
-                                    {rel.Units.map(unit => <li key={unit}><Button isInline variant="link" component="a" href={"#/" + unit} isDisabled={!this.props.isValid(unit)}>{unit}</Button></li>)}
+                                    {rel.Units.map(unit => <li key={unit}><Button isInline variant="link" component="a" href={"#/" + unit + (this.props.owner === "user" ? "?owner=user" : "")} isDisabled={!this.props.isValid(unit)}>{unit}</Button></li>)}
                                 </ul>
                             </DescriptionListDescription>
                         </DescriptionListGroup>
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -630,10 +630,18 @@ WantedBy=multi-user.target
         b.wait_not_present("#condition")
 
     def testRelationships(self):
+        self._testRelationships()
+
+    def testRelationshipsUser(self):
+        self._testRelationships(True)
+
+    def _testRelationships(self, user=False):
         m = self.machine
         b = self.browser
 
-        self.write_file("/etc/systemd/system/test-a.service",
+        systemd_path = "/etc/systemd/user" if user else "/etc/systemd/system"
+
+        self.write_file(f"{systemd_path}/test-a.service",
                         """
 [Unit]
 Description=A Service
@@ -644,7 +652,7 @@ Conflicts=test-c.service
 ExecStart=/bin/sh -c "while true; do sleep 5; done"
 """)
 
-        self.write_file("/etc/systemd/system/test-b.service",
+        self.write_file(f"{systemd_path}/test-b.service",
                         """
 [Unit]
 Description=B Service
@@ -654,7 +662,7 @@ After=test-a.service
 ExecStart=/bin/sh -c "while true; do sleep 5; done"
 """)
 
-        self.write_file("/etc/systemd/system/test-c.service",
+        self.write_file(f"{systemd_path}/test-c.service",
                         """
 [Unit]
 Description=C Service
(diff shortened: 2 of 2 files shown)


test/verify/check-system-services
```diff
@@ -630,10 +630,18 @@ WantedBy=multi-user.target
         b.wait_not_present("#condition")
 
     def testRelationships(self):
+        self._testRelationships()
+
+    def testRelationshipsSession(self):
```
martinpitt, line 635: Nitpick: Please let's call this "RelationshipsUser" for consistency -- nothing here applies to a "session", neither in the logind nor in the cockpit sense.

test/verify/check-system-services
```diff
@@ -630,10 +630,18 @@ WantedBy=multi-user.target
...
+        self._testRelationships(True)
+
+    def _testRelationships(self, user=False):
         m = self.machine
         b = self.browser
 
-        self.write_file("/etc/systemd/system/test-a.service",
+        path = "/etc/systemd/user" if user else "/etc/systemd/system"
```
martinpitt, line 642: s/path/unit_dir/ ? The scope of "path" is quite large, so the variable name should be self-describing.

test/verify/check-system-services
```diff
@@ -665,46 +673,52 @@ PartOf=test-b.service
...
         # services page
-        self.login_and_go("/system/services")
+        url = "/system/services#/?owner=user" if user else "/system/services"
+        self.login_and_go(url, user="admin")
         self.wait_page_load()
+
         self.wait_service_present("test-a.service")
         self.goto_service("test-a.service")
 
+        usr_sel = "?owner=user" if user else ""
```
martinpitt, line 692: It's not a selector, but a query string parameter.

pkg/systemd/services/service-details.jsx
```diff
@@ -473,7 +473,7 @@ export class ServiceDetails extends React.Component {
                             <DescriptionListTerm>{rel.Name}</DescriptionListTerm>
                             <DescriptionListDescription id={rel.Name.split(" ").join("")}>
                                 <ul className="comma-list">
-                                    {rel.Units.map(unit => <li key={unit}><Button isInline variant="link" component="a" href={"#/" + unit} isDisabled={!this.props.isValid(unit)}>{unit}</Button></li>)}
+                                    {rel.Units.map(unit => <li key={unit}><Button isInline variant="link" component="a" href={"#/" + unit + (this.props.owner === "user" ? "?owner=user" : "")} isDisabled={!this.props.isValid(unit)}>{unit}</Button></li>)}
```
martinpitt, line 476: This looks the wrong way around -- every [documentation](https://datatracker.ietf.org/doc/html/rfc3986/#page-16) [that I](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax) [know of](https://www.geeksforgeeks.org/components-of-a-url/) has the query *first* and then the fragment. Apparently cockpit-ws is cool with that order, but maybe let's stick to the common format? Or do we use this order somewhere else already?

martinpitt (changes requested):
Very nice, congratulations @Booooozer to your first cockpit PR! :tada: :star: 

This is really good, I just have some "nice to have" nitpicks.

pkg/systemd/services/service-details.jsx
marusak: this is how we do it everywhere. For example just to open user service: `https://localhost:9090/system/services#/colord-session.service?owner=user` but even logs `https://localhost:9090/system/logs#/s=19a35472f0f24e9cba7fd302491e93ab;i=1b0d2dd;b=6996acaedd1841efb0202962dcd37c38;m=436ac49485;t=5d6f0f1ae5216;x=9de9d67f897c7222?parent_options={%22priority%22:%22err%22}`

martinpitt:
Next up: @Booooozer , please have a look at how to set up the [git pre-push hook](https://github.com/cockpit-project/cockpit/blob/main/HACKING.md#building), so that this unit-tests failure doesn't happen:

    # test/verify/check-system-services:693:1: W293 blank line contains whitespace

The [centos-8-stream failure](https://logs.cockpit-project.org/logs/pull-16904-20220201-154009-119705c0-centos-8-stream/log.html#294-2) seems relevant?

pkg/systemd/services/service-details.jsx
martinpitt: Fair enough.

marusak:
> The [centos-8-stream failure](https://logs.cockpit-project.org/logs/pull-16904-20220201-154009-119705c0-centos-8-stream/log.html#294-2) seems relevant?

I don't think so. Different test, different page (this only touched details) and failed only on centos. I think it is just flake as all services tests had to succeed 3 times in this PR. We will see on re-push.

martinpitt (approved):
Thank you, this looks great now! I triggered the tests -- they have the last word before landing.

martinpitt:
@mvollmer: [single-pixel failure](https://logs.cockpit-project.org/logs/pull-16904-20220202-163953-7852b004-fedora-35/pixeldiff.html#TestStorageStratis-testBasic-create-pool-medium) here on [f35](https://logs.cockpit-project.org/logs/pull-16904-20220202-163953-7852b004-fedora-35/log.html#215)? I can't spot it even with magnifying.

martinpitt:
This `couldn't create runtime dir: /run/user/1001: Permission denied` flake is *super* annoying and non-obvious in 
[c8s](http://artifacts.dev.testing-farm.io/a7837a17-19ca-4e7e-9c3f-ecb056fe3fd2/). I've tried in vain to understand/reproduce it, this needs some deeper investigation. Ignoring for this PR, but it's getting on the flake list

PR #16905, opened 2022-02-01
systemd: style the shutdown/reboot icon grey

jelly:
Apply the same change as was applied to the "bug fix" icon, so the
scheduled reboot icon does not stand out.

This makes everything look the same:

image

Required for https://github.com/cockpit-project/cockpit/pull/16885

### DIFF
--- a/pkg/systemd/page-status.jsx
+++ b/pkg/systemd/page-status.jsx
@@ -55,7 +55,7 @@ function get_pficon(name) {
     if (name == "check")
         return <CheckIcon color="green" data-pficon={name} />;
     if (name == "spinner")
-        return <Spinner isSVG size="sm" data-pficon={name} />;
+        return <Spinner isSVG size="md" data-pficon={name} />;
 
     throw new Error(`get_pficon(): unknown icon name ${name}`);
 }
--- a/pkg/systemd/page-status.jsx
+++ b/pkg/systemd/page-status.jsx
@@ -28,7 +28,7 @@ import {
     ExclamationCircleIcon,
     InfoCircleIcon,
     SecurityIcon,
-    WarningTriangleIcon,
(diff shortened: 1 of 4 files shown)


garrett:
So I'm working on swapping out the warning icon to be the correct one. I have that done locally. That was easy. I'm also trying to track down why the spinner is the wrong size. It was sm, which was way too small. But md is too big... it's 18px whereas the icons are 16px. As the text is aligned with icon + space, it bumps the text over 2 pixels.

Screenshot from 2022-02-01 17-13-52

Zoomed, with guides:

Screenshot from 2022-02-01 17-36-53

I think this is a PatternFly issue. But they probably won't want to change it anytime soon, as it will be considered a "breaking change"... if it is even a problem in PF.

Should I add these fixes in this PR too, as they're icon related followups? Or keep the fixes separate?

pkg/systemd/page-status.jsx
```diff
@@ -55,7 +55,7 @@ function get_pficon(name) {
     if (name == "check")
         return <CheckIcon color="green" data-pficon={name} />;
     if (name == "spinner")
-        return <Spinner isSVG size="sm" data-pficon={name} />;
+        return <Spinner isSVG size="md" data-pficon={name} />;
```
garrett, line 58: This is kind of unnecessary, as we redefine the size in the CSS, but md is closer than sm.

garrett:
After my changes:

https://user-images.githubusercontent.com/10246/152018908-0e88a57c-8b0c-4d5f-85c0-1aeb97e899b3.mp4

Screenshot 2022-02-01 at 18-23-13 Overview - garrett Bolt

martinpitt (approved):
Thanks @jelly and @garrett , much appreciated! Let's include this in today's release, nice polish.

PR #16909, opened 2022-02-02
test: Documentation updates

martinpitt:
- Add more section, so that things are easier to find and link to

 - Start with the usual human concern of running a single test. Move the
   "run all tests" below and add a caveat -- most developers really
   don't want to run the full suite locally.

 - Document `TEST_SHOW_BROWSER` instead of the manual approach

 - Remove some redundancy.

 - Expand the section about iterating fast, in particular with a permanently
   running VM.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -13,45 +13,67 @@ with nested virtualization, refer to
 [this Fedora guide](https://docs.fedoraproject.org/en-US/quick-docs/using-nested-virtualization-in-kvm/index.html)
 for more details and recommendations on ensuring it is enabled correctly.
 
-## Introduction
+## Preparation and general invocation
 
-Before running the tests, ensure Cockpit has been built where the test suite
-expects to find it (do NOT run the build step as root):
+*Warning*: Never run the build, test, or any other command here as root!
 
-    $ ./test/image-prepare
+You first need to build cockpit, and install it into a VM:
 
-To run the integration tests run the following (do NOT run the integration tests
-as root):
+    $ test/image-prepare
 
-    $ ./test/common/run-tests
+This uses the default OS image, which is currently Fedora 35. See `$TEST_OS`
+below how to select a different one.
+
+In most cases you want to run an individual test in a suite, for example:
+
+    $ test/verify/check-metrics TestCurrentMetrics.testCPU
+
+You can get a list of tests by inspecting the `def test*` in the source, or by
+running the suite with `-l`/`--list`:
+
+    $ test/verify/check-metrics -l
+
+Sometimes you may also want to run all tests in a test file suite:
+
+    $ test/verify/check-session
+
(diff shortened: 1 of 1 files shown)


martinpitt:
@tomasmatus this may be interesting to you -- I hear you fell into the trap of wanting to run all tests. :grin:

test/README.md
```diff
@@ -13,45 +13,62 @@ with nested virtualization, refer to
...
+This uses the default OS image, which is currently Fedora 35. See `$TEST_OS`
+below how to select a different one.
 
-The tests will automatically download the VM images they need, so expect
-that the initial run may take a couple of hours (there are quite a few
-images to retrieve for different scenario tests).
+In most cases you want to run an individual test:
+
+    $ test/verify/check-metrics TestCurrentMetrics.testCPU
+
```
marusak, line 30: I would update this slightly.
I would but there `... for example test/verify...` now it sounds like there is only this one. 

Also would be worth mentioning that you can use `--list` to see the list of all available tests in given file. (Tricky now to see where you got `TestCurrentMetrics.testCPU`)

test/README.md
```diff
@@ -13,45 +13,62 @@ with nested virtualization, refer to
...
+If you specify `-s`/`--sit` in addition, then the test will wait on failure and
+allow you to log into cockpit and/or the test instance and diagnose the issue.
+The cockpit and SSH addresses of the test instance will be printed:
 
-Alternatively you can run an individual test like this:
+    $ test/verify/check-session -st
 
-    $ ./test/image-prepare
-    $ ./test/verify/check-session
+You can also run *all* the tests, with some parallelism:
```
marusak, line 45: well, you can also run one test with `test/common/run-tests` or any subset you add as arguments. But it is fine, normally users would not use this, so just saying it exists is fine.

test/README.md
```diff
@@ -13,45 +13,62 @@ with nested virtualization, refer to
...
+
+    $ TEST_SHOW_BROWSER=1 test/verify/check-session --trace
+
+You can also run a test against Firefox instead of Chromium:
+
+    $ TEST_BROWSER=firefox test/verify/check-session --trace
 
-    $ chromium-browser --remote-debugging-port=9222 about:blank
-    $ TEST_CDP_PORT=9222 ./test/verify/check-session --trace
+See below for detail.s
```
marusak, line 67: /s/detail.s/details

test/README.md
martinpitt: Thanks. Updated for both.

test/README.md
martinpitt: Right; I added that, but it's really not a tool that most developers should even know about.

test/README.md
martinpitt: Fixed.

marusak (approved):
Awesome, thanks!

PR #16912, opened 2022-02-03
test: Allow 5 differences before failing a pixel test

mvollmer:
Once in a hundred times or so, anti-aliasing of glyphs, rounded
corners, or similar things produces one or two pixel differences among
tens of thousands of perfectly rendered ones.  They are really hard to
spot in a delta image and not worth doing anything about.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -919,6 +919,8 @@ class Browser:
             #
             # - The RGB values of pixels can differ by up to 2.
             #
+            # - There can be up to 5 different pixels
+            #
             # Pixels that are different but have been ignored are
             # marked in the delta image in green.
 
@@ -943,6 +945,7 @@ class Browser:
                 data_now = now.load()
                 data_delta = delta.load()
                 result = True
+                count = 0
                 width, height = ref.size
                 for y in range(height):
                     for x in range(width):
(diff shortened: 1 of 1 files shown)


mvollmer:
I don't think we need to make this a parameter of `assert_pixels`. We don't know when and where these noisy pixels turn up, could be anywhere.

mvollmer:
As an alternative, we could tweak the threshold for when a pixel counts as different.  Currently it needs to differ in more than 2 in each of R, G, and B.  We could raise that level, or do something more clever.

marusak (approved):
Thanks! I fully agree with this approach. I cannot imagine when less then 5 changed pixels would actually make user visible changes.

> I don't think we need to make this a parameter of assert_pixels. We don't know when and where these noisy pixels turn up, could be anywhere.

Agree as well. It would be always just guesswork, we would put for some random test `4` then in later day there would actually be 5 different pixels so we would send PR to bump it to 6... If we will have issues with more then 5 pixels, then we need to think of more general fixes than just bumping this number up.

test/common/testlib.py
```diff
@@ -943,6 +945,7 @@ def img_eq(ref, now, delta):
                 data_now = now.load()
                 data_delta = delta.load()
                 result = True
+                count = 0
```
martinpitt, line 948: I'd call this "diff_count" or so.

martinpitt (approved):
Thanks! Agreed, if it's really about mostly random noise, and not something specific to the test, this approach makes much more sense.

Retrying the (unrelated) failing debian test. Ignoring packit -- the infra is busted right now, the team knows about it.

PR #16922, opened 2022-02-04
fmf: Don't install weak firefox dependencies

martinpitt:
firefox pulls in half a desktop, we don't need any of these for a
headless test. This reduces the number packages from 133 (459 MB) to 66
(365 MB).


marusak (approved):
Danke :)

PR #16923, opened 2022-02-04
test: Reduce packagekit.service stop time

martinpitt:
packagekit.service tends to hang on shutdown. This bug [1] is ancient,
and we kept working around it in bots' *.install files; but we keep
losing that hack with newer images, and it's also missing on
rhel-9-0.install.

So centralize the hack here for all PackageKit related tests.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1717185

----

This ought to help a bit with the ridiculous [testTracer](https://logs.cockpit-project.org/logs/pull-16891-20220201-165621-7b1e6dcf-rhel-9-0/log.html) runtime of > 5 minutes on our infra. Locally it was ~ 170 seconds on main, and now around 90.

### DIFF
--- a/test/verify/packagelib.py
+++ b/test/verify/packagelib.py
@@ -52,6 +52,10 @@ class PackageCase(MachineCase):
             self.machine.execute("nmcli con add type dummy con-name fake ifname fake0 ip4 1.2.3.4/24 gw4 1.2.3.1")
             self.addCleanup(self.machine.execute, "nmcli con delete fake")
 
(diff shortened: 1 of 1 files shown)


martinpitt:
In [this run](https://logs.cockpit-project.org/logs/pull-16923-20220204-083730-57963962-rhel-9-0/log.html), `testTracer` now took 148s, compared to the 300 it used before. Overall, the test ran for 2854s (47 min), which is quite a nice improvement. It also likely  helps other OSes (I didn't check them)

jelly (approved):

PR #16924, opened 2022-02-04
test: Stop marking rebooting TestUpdates tests as nondestructive

martinpitt:
testInfoSecurity and testTracer reboot, so they can't run on the Testing
Farm. Mark all other tests as nondestructive, and drop the special
treatment in the FMF tests. This will run a few more tests in FMF.

----

Also do some FMF test adjustments. The first round is experimental to get some details if the excluded tests work now, and if not, why exactly.

### DIFF
--- a/test/run-verify-host-user.sh
+++ b/test/run-verify-host-user.sh
@@ -49,12 +49,10 @@ TESTS=""
 EXCLUDES=""
 RC=0
 if [ -n "$test_optional" ]; then
-    # TestUpdates: we can't run rebooting tests
     TESTS="$TESTS
+         TestUpdates
          TestAutoUpdates
-         TestStorage
-         TestUpdates.testBasic
-         TestUpdates.testSecurityOnly"
+         TestStorage"
 fi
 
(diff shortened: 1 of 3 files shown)


martinpitt:
[c8s](https://logs.cockpit-project.org/logs/pull-16924-20220204-085125-77643b4c-centos-8-stream/log.html) passed.

There are some FMF test failures on [c9s](http://artifacts.dev.testing-farm.io/b19ba198-41fa-45e7-890a-3c9e5e237922/) and [f34](http://artifacts.dev.testing-farm.io/0948cf1f-2e0c-4e13-956c-93f111f5e441/) in the newly enabled tests, but curiously *not* in the old/removed exclusions. But the failures seem fairly obvious -- I pushed the fixes here, if it still fails I'll split them out into a separate PR.

[f35](http://artifacts.dev.testing-farm.io/29badaf9-ff89-410a-8b60-b32e6a47d33e/) has the exact same failures as F34.

jelly (approved):

PR #5898, opened 2017-02-15
po: Put translations into individual packages

mvollmer:

PR #16930, opened 2022-02-07
pkg: stop including cross-package translations

### LINKED ISSUE #13906: Include translations right into the "compiled" cockpit.js
Include translations right into the "compiled" cockpit.js, so that it works for third-party pages like podman as well and not bundle it into every po.lang.js file.

See #13901

allisonkarlitskaya:
A great number of our pages contain this:

```
  <script type="text/javascript" src="../*/po.js"></script>
```

in their .html files.  The effect is that the page pulls in the
translations for every package that the bridge knows about, including
those that may have come from a different source package.

This was introduced in #5898, but unfortunately the PR is a bit thin on
rationale for why this was done.

Each page really only needs it own translations, plus the translations
for base1.  Adjust accordingly.

See also discussion in #13906.

 - Blocked on fixing #8160

Fixes #13906

### DIFF
--- a/pkg/apps/index.html
+++ b/pkg/apps/index.html
@@ -24,7 +24,8 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <meta charset="utf-8">
     <link href="apps.css" type="text/css" rel="stylesheet">
     <script type="text/javascript" src="../base1/cockpit.js"></script>
-    <script type="text/javascript" src="../*/po.js"></script>
+    <script type="text/javascript" src="../base1/po.js"></script>
+    <script type="text/javascript" src="po.js"></script>
     <script type="text/javascript" src="../manifests.js"></script>
     <script type="text/javascript" src="apps.js"></script>
   </head>
--- a/pkg/shell/index.html
+++ b/pkg/shell/index.html
@@ -7,8 +7,9 @@
     <link href="index.css" rel="stylesheet">
     <link href="../../static/branding.css" rel="stylesheet">
     <script src="../base1/cockpit.js"></script>
+    <script src="../base1/po.js"></script>
     <script src="../manifests.js"></script>
-    <script src="../*/po.js"></script>
+    <script src="po.js"></script>
   </head>
   <body class="pf-m-redhat-font" hidden>
     <div id="main" class="page">
--- a/pkg/kdump/index.html
+++ b/pkg/kdump/index.html
@@ -27,7 +27,8 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <link rel="stylesheet" href="kdump.css">
 
     <script type="text/javascript" src="../base1/cockpit.js"></script>
-    <script type="text/javascript" src="../*/po.js"></script>
+    <script type="text/javascript" src="../base/po.js"></script>
+    <script type="text/javascript" src="po.js"></script>
     <script type="text/javascript" src="kdump.js"></script>
 </head>
 
--- a/pkg/metrics/index.html
+++ b/pkg/metrics/index.html
@@ -26,7 +26,8 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
 
     <script type="text/javascript" src="../base1/cockpit.js"></script>
     <script type="text/javascript" src="../manifests.js"></script>
-    <script type="text/javascript" src="../*/po.js"></script>
+    <script type="text/javascript" src="../base1/po.js"></script>
+    <script type="text/javascript" src="po.js"></script>
     <script type="text/javascript" src="index.js"></script>
 </head>
 
--- a/pkg/networkmanager/firewall.html
+++ b/pkg/networkmanager/firewall.html
@@ -25,7 +25,8 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <link href="firewall.css" type="text/css" rel="stylesheet">
 
     <script src="../base1/cockpit.js"></script>
-    <script src="../*/po.js"></script>
+    <script src="../base1/po.js"></script>
+    <script src="po.js"></script>
     <script src="firewall.js"></script>
   </head>
 
--- a/pkg/networkmanager/index.html
+++ b/pkg/networkmanager/index.html
@@ -26,7 +26,8 @@
   <link href="network.css" type="text/css" rel="stylesheet">
   <script src="../base1/cockpit.js"></script>
   <script src="../manifests.js"></script>
-  <script src="../*/po.js"></script>
+  <script src="../base1/po.js"></script>
+  <script src="po.js"></script>
   <script src="network.js"></script>
 </head>
 <body class="pf-m-redhat-font">
--- a/pkg/packagekit/index.html
+++ b/pkg/packagekit/index.html
@@ -26,8 +26,9 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
   <link href="updates.css" rel="stylesheet">
 
   <script src="../base1/cockpit.js"></script>
+  <script src="../base1/po.js"></script>
   <script src="updates.js"></script>
-  <script src="../*/po.js"></script>
+  <script src="po.js"></script>
 </head>
 <body class="pf-m-redhat-font">
     <div class="ct-page-fill" id="app"></div>
--- a/pkg/playground/translate.html
+++ b/pkg/playground/translate.html
@@ -130,7 +130,8 @@
     </div>
     <script src="translate.js"></script>
     <!-- Bring in initial translations -->
-    <script src="../*/po.js"></script>
+    <script src="../base1/po.js"></script>
+    <script src="po.js"></script>
     <!-- Override translations from here -->
     <script src="po.extra.js"></script>
 </body>
--- a/pkg/selinux/setroubleshoot.html
+++ b/pkg/selinux/setroubleshoot.html
@@ -27,7 +27,8 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <link rel="stylesheet" href="selinux.css">
 
     <script type="text/javascript" src="../base1/cockpit.js"></script>
-    <script type="text/javascript" src="../*/po.js"></script>
+    <script type="text/javascript" src="../base1/po.js"></script>
+    <script type="text/javascript" src="po.js"></script>
     <script type="text/javascript" src="selinux.js"></script>
 </head>
 
--- a/pkg/sosreport/index.html
+++ b/pkg/sosreport/index.html
@@ -23,7 +23,8 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <meta charset="utf-8">
     <link href="sosreport.css" rel="stylesheet">
     <script type="text/javascript" src="../base1/cockpit.js"></script>
-    <script type="text/javascript" src="../*/po.js"></script>
+    <script type="text/javascript" src="../base1/po.js"></script>
+    <script type="text/javascript" src="po.js"></script>
     <script type="text/javascript" src="sosreport.js"></script>
   </head>
   <body class="pf-m-redhat-font">
--- a/pkg/storaged/index.html
+++ b/pkg/storaged/index.html
@@ -25,8 +25,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link href="storage.css" type="text/css" rel="stylesheet">
     <script src="../base1/cockpit.js"></script>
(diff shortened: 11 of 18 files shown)


martinpitt (dismissed):
Thanks! starter-kit and c-podman already do this correctly. [c-machines needs fixing](https://github.com/cockpit-project/cockpit-machines/blob/main/src/index.html#L12). I didn't look at our [other third-party projects](https://cockpit-project.org/applications.html) yet, I suppose some of them inherited this as well.

allisonkarlitskaya:
I wanna wait on specifically @mvollmer to look at this PR before we merge it.  He might have some more information about why this was done in the first place.

allisonkarlitskaya:
Somewhat unsurprisingly, 
```
    3: test/verify/check-apps TestApps.testL10N
```

martinpitt:
@allisonkarlitskaya : I don't think that test fails on missing "foreign page" translations -- apps translations are supposed to come out of their XML description, with the `xml:lang="de"` thing. There is some subtle reason why this test is broken now, but we surely don't translate `DESCRIPTION:none` into German or Japanese in the official po files.

mvollmer:
> This was introduced in #5898, but unfortunately the PR is a bit thin on
> rationale for why this was done.

As far as I can see, #5898 did not change which translations a page would load, just how they were stored on disk. Before and after, a page would load all translations. (After, it might load those translations multiple times that are shared between pages, no?)

mvollmer:
> Each page really only needs it own translations, plus the translations
> for base1.

What about translations for code in `pkg/lib/`, do we handle those correctly?

A page _might_ also import stuff from other pages during build, and then might need translations from that other page, no?  I check if I can find an instance.

mvollmer:
> > Each page really only needs it own translations, plus the translations
> > for base1.
> 
> What about translations for code in `pkg/lib/`, do we handle those correctly?

Yes, cockpit-po-plugin looks into dependencies as well.  For example, `dist/systemd/po.de.js` contains translations for `pkg/lib/serverTime.js`.
 
> A page _might_ also import stuff from other pages during build, and then might need translations from that other page, no? I check if I can find an instance.

This should also be handled correctly by cockpit-po-plugin, and indeed: `pkg/systemd/overview.jsx` has 

```
```

and `dist/systemd/po.de.js` has translations for that.

So all is good, dependency-wise.  I guess this was missing with #5898, and that's why we still needed to load all translations into every page.

martinpitt:
> What about translations for code in pkg/lib/, do we handle those correctly?

However we handle it, it can't happen through the *.po globbing approach: pkg/lib is not a webpack, so it doesn't ship its own translations. This needs to happen at build time. It also seems to work -- e.g.

    pkg/lib/cockpit-components-install-dialog.jsx: title: options.title || _("Install software"),

We use this component in multiple pages, and it does appear in all of them:

```
❱❱❱ find dist -name 'po.de.js'  | xargs grep -l 'Install software'
dist/storaged/po.de.js
dist/packagekit/po.de.js
dist/systemd/po.de.js
dist/metrics/po.de.js
```

mvollmer:
> However we handle it, it can't happen through the *.po globbing approach

Good point!

mvollmer (approved):
Looks good to me.

allisonkarlitskaya:
I guess this has something to do with the weird side-effect-y call to `cockpit_locale_set_language()` in `handle_packages()`.

martinpitt:
Debugging notes:

 - This *does* work correctly with `TEST_BROWSER=firefox`, so this smells a bit like our old friend #8160. However, no amount of reloading, killing the page, logging out and back in, or restarting the browser has any effect, it just keeps showing the German translation. *Even* when I switch the language back to English.
 - There is nothing special about Japanese. It happens with French as well.
 - As soon as there is *any* language cookie, it only ever shows the German translation.
 - Adding a French (pseudo-)translation to the test XML file doesn't change things, it still gets stuck on "de". It feels like the first-ever language you select sticks.
 - This isn't limited to the "Apps" page. When I switch language, only the shell changes, but the Overview and whole Apps page are still in German.
 - `cockpit.language` gets changed correctly in the shell (top-level document, `fr`), but it's still wrong in the individual frames (`de`)
 
So for now it looks like this just heavily aggravates #8160. Unfortunately fixing this for frames is where @marusak got stuck at in PR #15633

allisonkarlitskaya:
My theory is that everything is working exactly correctly.  Chromium notices that it has an etag-tagged version of the page in the cache and does the request conditional on `If-None-Match` and gets told that it already has the correct content.  This is demonstrated by the fact that we actually have the correct translation visible on the page.  In that case, the request would never reach `cockpit-bridge` (and doesn't need to: this is the entire point of caching).

The really broken thing here is the assumption in cockpit-bridge that the client will definitely request at least one file from the packages component when the language changes.  This is an awful place to change the language of the entire bridge process.

I think what we want to do here is to add a `locale` field to the `init` message to the bridge, and re-send the `init` when the language changes.  This would be a far more appropriate place to change the locale of the process than as a side-effect of serving some file from the webserver.

allisonkarlitskaya:
Actually, I think it's our bug.  We ought to be setting `Vary: Accept-Language` if that's the case.

martinpitt:
>  We ought to be setting Vary: Accept-Language if that's the case.

I tried that:
```diff
--- src/common/cockpitwebresponse.c
+++ src/common/cockpitwebresponse.c
@@ -856,10 +856,12 @@ finish_headers (CockpitWebResponse *self,
         g_string_append (string, "Cache-Control: max-age=86400, private\r\n");
     }
 
-  if ((seen & HEADER_VARY) == 0 && status >= 200 && status <= 299 &&
-      self->cache_type == COCKPIT_WEB_RESPONSE_CACHE_PRIVATE)
+  if ((seen & HEADER_VARY) == 0 && status >= 200 && status <= 299)
     {
-      g_string_append (string, "Vary: Cookie\r\n");
+      if (self->cache_type == COCKPIT_WEB_RESPONSE_CACHE_PRIVATE)
+        g_string_append (string, "Vary: Cookie\r\n");
+      else if (self->cache_type == COCKPIT_WEB_RESPONSE_CACHE_FOREVER)
+        g_string_append (string, "Vary: Accept-Language\r\n");
     }
 
   if (!self->keep_alive)
```
(that's too broad of course, just for testing).

I do see the header in the responses in the network inspector, but as the file is cached, the browser doesn't re-request it. Same chicken-and-egg problem. I.e. it doesn't "see" that Accept-Language would change *if* it did a request.

allisonkarlitskaya:
Okay.  There's no depth to the insanity here.  It turns out that not listing `Accept-Language` in `Vary` is actually appropriate: we don't use this (at least not in a way that's visible to the browser).  channelresponse does this instead:

```
    /* Parse the language out of the CockpitLang cookie and set Accept-Language */
    language = cockpit_web_server_parse_cookie (headers, "CockpitLang");
    if (language)
      g_hash_table_replace (headers, g_strdup ("Accept-Language"), language);
```

which is, indeed, variance based solely on `Cookie` (as we currently advertise).

Sigh.

allisonkarlitskaya:
I did some measuring with the caching changes from #17092.

I emulated latency by doing this:

```
ssh c tc qdisc add dev eth0 root netem delay 250ms
```

note: I used debian-testing, because the Fedora images don't have netem in the kernel.

In each case, I logged in, then changed to the logs page, then storage, then users.

Here's before this change (ie: serving `*/po.js`):

Screenshot from 2022-03-10 16-14-04

and after (ie: serving individual `page/po.js` files):
Screenshot from 2022-03-10 16-08-54

Notice in both cases the `.po` file gets loaded in parallel with the main `.js` for the page, and always takes significantly less time.

I think I'm comfortable with saying that this shouldn't make too big of a difference in practice...

martinpitt (dismissed):
Thanks! Assuming that this works now.

The last two rounds had some horrible failures, but CI usage should be lower in the evening now. I'll tune the parallelism tomorrow morning.

martinpitt:
[this failure](https://logs.cockpit-project.org/logs/pull-16930-20220310-174823-1f4a208e-rhel-8-6-distropkg/log.html#3) is significant. I have a quirk in that test to *do* reload the page for distropkg, but it doesn't seem to bite here (but did work in #17092). Happy to take a look tomorrow.

martinpitt:
Oh, this is hard to get green because it literally affects *all* pages, so *all* tests run three times. That's going to be "fun" :grin: Note to self: Temporarily go back to TEST_JOBS=5 (old setup) and see if it works any better.

In the worst case we'll have to split this up a bit.

martinpitt:
I rolled back the increased $TEST_JOBS (see https://github.com/cockpit-project/cockpituous/pull/459#issuecomment-1064400522) and retried the tests. Let's see if they look any better, i.e. if the reason is the increased parallelization or the many hundred "affected" retries.

martinpitt:
This is *muuuch* greener, so the different parallelization was to blame. I adjusted https://github.com/cockpit-project/cockpituous/pull/459 to hopefully worked a bit better now.

I updated this PR to deal with the remaining distropkg failure. Let's give this a try with the new bots layout, and if it still fails that badly, I'll revert cockpituous, we let that PR land, and I'll create a synthetic "CI infra stress test" PR.

martinpitt:
The updated increased parallelization [is still messy](https://logs.cockpit-project.org/logs/pull-16930-20220311-064509-e9f51076-centos-8-stream/log.html#138). I reverted it, and force-pushed to re-run all tests again.

martinpitt (approved):

martinpitt:
The multi-machine/host-switching tests are [so demanding](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16930-20220311-072722-d0832932-debian-stable/log.html#128-2) that running them multiple times still messes up. I am working on a PR which may alleviate that.

martinpitt:
[this failure](https://logs.cockpit-project.org/logs/pull-16930-20220311-085325-d0832932-ubuntu-2004/log.html#297) does not look like fallout from too much CPU/IO/RAM on the infra -- but more like some systematic cleanup error when running some tests multiple times. Retrying for comparison, but I think it may already have failed the same way before (but I didn't keep the log)

pkg/apps/index.html
```diff
@@ -24,7 +24,8 @@
     <meta charset="utf-8">
     <link href="apps.css" type="text/css" rel="stylesheet">
     <script type="text/javascript" src="../base1/cockpit.js"></script>
-    <script type="text/javascript" src="../*/po.js"></script>
+    <script type="text/javascript" src="../base1/po.js"></script>
```
martinpitt, line 27: You made a good point in https://github.com/cockpit-project/cockpit/pull/16982/commits/b2453bab40a8ad3fdb87f2dcd9929c9fad9388e4 -- maybe we should keep the current cockpit-po-plugin hack that includes all base1 translations into each .po file, and drop this separate inclusion. That means slightly bigger files, but saves the extra roundtrip for this additional file. Especially as we don't cache it any more?

allisonkarlitskaya: I disagree with this.

 - these files get loaded in parallel, so even if base1/po.js isn't cached, it's not going to hurt us much
 - the real fix for this problem is to fix caching, which we can do by making language switching more explicit, if that's what we want

martinpitt: Ack.

allisonkarlitskaya: One more point: this solution is only available to modules which live in the `cockpit` source tree.  External modules will have to include `base1/` separately, as there's no way for them to inspect the .po files for `base1/` at their build time.

The other way to fix this, if you really hate the separate include, is as was suggested in the original issue: integrate the po into base1 itself.  But that means we'd need to pick two options:
 - have per-language variants of `cockpit.js`, which, after the experiences of the last few days seems like a very very bad idea
 - include translations for all languages

I don't consider either of those options to be good.

pkg/shell/index.html
```diff
@@ -7,8 +7,9 @@
     <link href="index.css" rel="stylesheet">
     <link href="../../static/branding.css" rel="stylesheet">
     <script src="../base1/cockpit.js"></script>
+    <script src="../base1/po.js"></script>
     <script src="../manifests.js"></script>
-    <script src="../*/po.js"></script>
+    <script src="po.js"></script>
```
martinpitt, line 12: Meh, this caused a regression, see https://github.com/cockpit-project/starter-kit/pull/542#issuecomment-1070472816

This stops showing translated menu entries. I wonder why we don't see that for cockpit's own pages? They are all translated for me in German and French. But not for starter-kit (and presumably other third-party pages)

martinpitt: I sent #17145 to fix this

PR #16932, opened 2022-02-07
metrics: Use `setUTC*` functions to work across all timezones

### LINKED ISSUE #16492: metrics: History view: Embedded logs aren't aligned with the time
As seen in the historical view in the UX Study: Metrics can say spike events happened at 5:35, 5:36, and 5:37, but when the logs are expanded, they say 5:06.

This prevented people from completing the last task successfully in an easier manner.

From the screen of participant 5 around 30 minutes in, where it was specifically called out:

Screenshot from 2021-10-19 14-03-33_ L3

marusak:
If timezone is shifted by 30 minutes compared to UTC, then `setMinutes`
does not behave as expected.

Example:
`time` in UTC is "10:34" which in some timezone is "3:04". We then want
to round the time by one minute down, so we do `setMinutes(33)` which
works for all timezones that are shifted by full hours but it completely
breaks some timezones as they now would not be "3:03" but "3:33" which
in UTC would be "11:03".

Fixes #16492

Before:
Same timezone:
Screenshot from 2022-02-07 12-42-24

Different timezone:
Screenshot from 2022-02-07 15-36-18

With this PR different timezone:
newminutes

@ testing https://github.com/cockpit-project/cockpit/pull/15626

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -665,12 +665,12 @@ class MetricsMinute extends React.Component {
         const end_second = (end - (end_minute * SAMPLES_PER_MIN)) * (60 / SAMPLES_PER_MIN);
 
         const time = new Date(timestamp);
-        time.setMinutes(start_minute);
-        time.setSeconds(start_second);
+        time.setUTCMinutes(start_minute);
+        time.setUTCSeconds(start_second);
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Many thanks!

PR #16933, opened 2022-02-07
build: add `make rsync` target

allisonkarlitskaya:
This builds the tree, runs `make install` into a temporary DESTDIR, and then rsyncs the result to another host (default: `c`).

This is intended to be used with the recently-added ./autogen.sh rpm feature in order to live-apply changes to a running VM image.

An incremental `make rsync` after no changes takes less than one second.

### How to use this / manual testing:

 - check out the branch
 - `./autogen.sh rpm`
 - `test/image-prepare -q fedora-35`
 - `vm-run -q fedora-35&`
 - `make -j8 rsync`
 - `ssh c systemctl enable --now cockpit.socket`
 - open web browser, login
 - break some stuff
 - `ssh c systemctl stop cockpit.socket`   # needed because of "Text file busy"
 - `make rsync`
 - `ssh c systemctl start cockpit.socket`
 - open web browser, watch stuff fail

### Future directions
 - automatic cockpit restart to avoid the `ETXTBUSY` thing
 - ability to install directly into a downloaded image, without first preparing it

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,6 +2,7 @@ NULL =
 man_MANS =
 BUILT_SOURCES =
 bin_PROGRAMS =
+check_DATA =
 check_SCRIPTS =
 check_PROGRAMS =
 noinst_SCRIPTS =
--- a/Makefile.am
+++ b/Makefile.am
@@ -63,6 +63,7 @@ install-data-local:: $(WEBPACK_INSTALL) $(MANIFESTS)
 install-data-local:: $(WEBPACK_GZ_INSTALL)
 	@for file in $^; do \
 		target="$(DESTDIR)$(pkgdatadir)/$${file##*dist/}.gz"; \
+		[ -n "$(DESTDIR)" -a "$${target}" -nt "$${file}" ] && continue; \
 		mkdir -p "$${target%/*}"; \
 		echo "$${file} → $${target}"; \
 		gzip -9 < "$${file}" > "$${target}"; \
--- a/Makefile.am
+++ b/Makefile.am
@@ -209,6 +209,14 @@ multi-word: po/cockpit.pot
 title2sentence: multi-word
 	tools/title2sentence.py -i titles -o t2s.sh
 
+.PHONY: rsync
+RSYNC_HOST ?= c
+RSYNC_DEST ?= $(RSYNC_HOST):/
+rsync:
+	$(MAKE)
+	$(MAKE) install DESTDIR=tmp/rsync >/dev/null
+	rsync --recursive --links --checksum --verbose --inplace tmp/rsync/ $(RSYNC_DEST)
+
 include containers/Makefile.am
 include containers/flatpak/Makefile.am
 include doc/Makefile-doc.am
--- a/src/bridge/Makefile.am
+++ b/src/bridge/Makefile.am
@@ -238,7 +238,7 @@ test_websocketstream_SOURCES = \
 test_websocketstream_CFLAGS = $(libcockpit_bridge_a_CFLAGS)
 test_websocketstream_LDADD = $(libcockpit_bridge_LIBS) -ldl
 
-noinst_PROGRAMS += $(BRIDGE_CHECKS) mock-bridge
+check_PROGRAMS += $(BRIDGE_CHECKS) mock-bridge
 TESTS += $(BRIDGE_CHECKS)
 
 EXTRA_DIST += \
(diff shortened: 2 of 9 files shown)


allisonkarlitskaya:
Note: I consider the change to remove building of the tests from `make all` to be somewhat controversial.  I'm not 100% sure this is a good thing, for two reasons:

 - the tests being able to link serves as some kind of a useful check at `make` time to make sure you didn't mess something up
 - it's common to write `make -j$(nproc) all && make check`, which would result in the tests build happening non-parallel now.

It's also worth noting that I found another way to approach this: we can just pass `noinst_DATA= noinst_PROGRAMS= noinst_SCRIPTS=` as arguments to the `make ` invocations from the `rsync` rule.  Reviewer decides :)

allisonkarlitskaya:
> Note: I consider the change to remove building of the tests from `make all` to be somewhat controversial.

I start to consider a third way forward on this one: I rather enjoy the separation of the tests stuff into the explicit `check_` variables, so we could consider keeping that part, but then adding at the top level something like:

```
# Build all test programs and resources by default
noinst_DATA += $(check_DATA)
noinst_PROGRAMS += $(check_PROGRAMS)
noinst_SCRIPTS += $(check_SCRIPTS)
```

in the toplevel Makefile.am.

That might also be a good place to introduce a single simple conditional flag to disable building the tests as part of the main build, which we could then use from the `make rsync` rule.

It's all a bit academic to be honest, though:  most of our uses of `noinst_*` are precisely for "we use this for tests", with some exceptions:

 - `noinst_LIBRARIES` for all of our internal static libraries
 - a single (questionable) `noinst_SCRIPTS` to drive the building of the webpacks
 - a single (questionable) `noinst_DATA` to drive the docs build
 - a very strange `noinst_SCRIPTS` in `src/ssh/` for two files that are already in version control...

src/bridge/Makefile.am
```diff
@@ -238,7 +238,7 @@ test_websocketstream_SOURCES = \
 test_websocketstream_CFLAGS = $(libcockpit_bridge_a_CFLAGS)
 test_websocketstream_LDADD = $(libcockpit_bridge_LIBS) -ldl
 
-noinst_PROGRAMS += $(BRIDGE_CHECKS) mock-bridge
+check_PROGRAMS += $(BRIDGE_CHECKS) mock-bridge
```
martinpitt, line 241: FTR, I really like this change! That's how it's supposed to be, IMHO.

Makefile.am
```diff
@@ -209,6 +209,13 @@ multi-word: po/cockpit.pot
 title2sentence: multi-word
 	tools/title2sentence.py -i titles -o t2s.sh
 
+RSYNC_HOST ?= c
+RSYNC_DEST ?= $(RSYNC_HOST):/
+rsync:
+	$(MAKE)
+	$(MAKE) install DESTDIR=tmp/rsync >/dev/null
+	rsync -avcO tmp/rsync/ $(RSYNC_DEST)
```
martinpitt, line 217: Why the `-c`, isn't that more expensive? Are you worried about different timezones on the target or so? (I would have thought that it would compare seconds-since-epoch Unix timestamps)

I had to look up `-c` and `-O`. For scripts, let's please use long options.

Makefile.am
```diff
@@ -63,6 +63,7 @@ install-data-local:: $(WEBPACK_INSTALL) $(MANIFESTS)
 install-data-local:: $(WEBPACK_GZ_INSTALL)
 	@for file in $^; do \
 		target="$(DESTDIR)$(pkgdatadir)/$${file##*dist/}.gz"; \
+		[ -n "$(DESTDIR)" -a "$${target}" -nt "$${file}" ] && continue; \
```
martinpitt, line 66: I had to think about this a bit -- "newer" because gzip creates a new file, so if a file was not touched/rebuilt locally, the dest file.gz will generally be newer than the local file.

The commit message's subject is hard to read.

Makefile.am
martinpitt: `rsync` needs to be added to `.PHONY`.

martinpitt (changes requested):
Thanks! This looks great to me, just two tiny nitpicks.

Makefile.am
allisonkarlitskaya: If we don't use `--checksum` then we end up copying every file every time, because the version in the `DESTDIR` was just copied there a moment ago and is therefore newer than the target.

I adjusted to long options and added `.PHONY`.

Makefile.am
allisonkarlitskaya: I've reworked the subject line to be slightly less accurate but significantly easier to parse.

Sometimes staying under 50 characters has its cost...

allisonkarlitskaya: Btw: the fact that this adds a (serial) 1.85s to `make install` suggests to me that maybe we just want to do the compression as part of the webpack build... There are plugins for that...

That's another PR, for sure, though.

PR #16935, opened 2022-02-07
test: Rewrite image-prepare using image-customize

martinpitt:
Use `image-customize --build` for building and installing the packages on all
operating systems. Move the remaining setup and OS specific hacks to
test/vm.install.

This brings cockpit in line with starter-kit and derivatives, and gets rid of a
lot of redundancy. In particular, this gets rid of the last cockpit specific
*.install files in the bots repository.

Drop the following CLI options:
 - `--overlay`, `--install-only`: Nobody ever uses this; we have faster
   tools to hack existing images.
 - `--build-image`: The test image determines the build image, there is only
   one correct value anyway
 - `--build-only`, `--sit`, not currently supported by image-customize

-----

Dependencies/To do:
 - [x] PR #16939
 - [x] https://github.com/cockpit-project/bots/pull/2909
 - [x] Use the above option in [starter-kit](https://github.com/cockpit-project/starter-kit/pull/526), [c-podman](https://github.com/cockpit-project/cockpit-podman/pull/910), [c-machines](https://github.com/cockpit-project/cockpit-machines/pull/562)
 - [x] Use the above option in image-prepare
 - [x] add distropkg: as a replacement for [don't upgrade existing packages filter](https://github.com/cockpit-project/bots/blob/3b4c20a2fd51f51c7e0679b700d90fbad0f32d92/images/scripts/lib/fedora.install#L5), only build/install c-appstream
 - [x] add custom rpm build/copy/install for fedora-coreos
 - [x] add containers/bastion preparation
 - [x] replace `FEDORA_COREOS_BUILD_IMAGE` with bots testmap query
 - [x] clean up make-srpm
 - [x] add rpm duplicate files check
 - [x] factorize image-prepare's main() function


martinpitt:
Hey, look at that, almost all green -- just a [single flake](https://logs.cockpit-project.org/logs/pull-16935-20220212-063305-dd3ff86b-rhel-9-0/log.html#97-2). Retrying for comparison.

This is mostly feature complete (see TODO list), but still needs some code cleanup.

martinpitt:
This is ready, working (arch is known issue, see https://github.com/cockpit-project/bots/pull/2944), and I am now reasonably happy with it. @allisonkarlitskaya , I'd appreciate your review.

@marusak, I'd also appreciate your feedback, if you care (if not, that's fine -- happy to banter over the details with Lis, but it'd be good to at least get an opinion about the general approach)

martinpitt:
Just a clean rebase, mostly to get green tests.

martinpitt:
Something here [does not like arch](https://logs.cockpit-project.org/logs/pull-16935-20220220-073518-591b1907-arch/log.html). I think I need to cherry-pick that [glibc version pin-down hack](https://github.com/cockpit-project/bots/commit/50d159acd1ef78da07a8dc219d6c4fe2a752f9a0).

martinpitt:
@marusak : Lis is still out for the next two weeks, and I wouldn't want to let this bitrot for so long. She was okay (in fact, really happy) about the general approach, and as this does not go into a product anywhere, we can do some Python beautifications as follow-ups. Could you please review?

martinpitt:
FTR: After this lands, I'll apply it to the rhel-7.9 branch as well, and then drop all these `*.install` files from bots.

marusak (approved):
Awesome!
I am very much looking forward to dropping of `.install` scripts!

PR #16936, opened 2022-02-07
logs: Don't navigate to log details when selecting details

### LINKED ISSUE #14327: logs: Selecting text to copy from the log viewer is difficult
Page: Logs

I haven't seen the corresponding video, but I have seen it in the notes that copying from the logs is difficult. In testing it myself, I found that Cockpit wants to eagerly visit the details page on a mouse up. 

When selecting text on the logs page, a mouseUp activates the link and takes someone to the details page for that log entry.

A possible workaround to make clicking on a line work as intended and still allow for someone to select the text to copy and paste might be something like the following pseudocode:

- mouseDown
   - start a timer
   - measure mouse position 
- mouseUp
   - check position
   - check timer
- if ((distanceOfMouseMovement > somePixels) || clickTimer > someTime)
   - // allow selection and let someone copy the text
   - stayOnPage
- else
   - // this is what we normally do
   - navigate to details page

I'm not sure what distance or time we should use that would feel comfortable for most people and still be accessible for those with tremor or other coordination-related accessibility issues.

_Alternatively_, we could [check for a selection](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection) on mouseUp. This may have the drawback of having selected text elsewhere preventing the navigation effort. To be certain, we might want to check if the selection as a substring against the log line.

(This second approach is probably better of the two, as we wouldn't have to use arbitrary values based on time or distance.)

marusak:
Users might want to select log message or identifier to use it for
searching. We should not navigate in such case.

Fixes #14327

I tried to write tests but I am having no luck...
```
b.mouse(".cockpit-logline:nth-child(2) .cockpit-log-message", "mousedown", 0, 10)
b.mouse(".cockpit-logline:nth-child(2) .cockpit-log-message", "mousemove", 140, 10)
b.mouse(".cockpit-logline:nth-child(2) .cockpit-log-message", "mouseup", 140, 10)

```

### DIFF
--- a/pkg/lib/cockpit-components-logs-panel.jsx
+++ b/pkg/lib/cockpit-components-logs-panel.jsx
@@ -40,10 +40,17 @@ export class JournalOutput {
         this.search_options = search_options || {};
     }
 
-    onEvent(ev, cursor) {
+    onEvent(ev, cursor, full_content) {
         // only consider primary mouse button for clicks
-        if (ev.type === 'click' && ev.button !== 0)
-            return;
+        if (ev.type === 'click') {
+            if (ev.button !== 0)
+                return;
+
+            // Ignore if text is being selected - less than 3 characters most likely means misclick
+            const selection = window.getSelection().toString();
+            if (selection && selection.length > 2 && full_content.indexOf(selection) >= 0)
+                return;
+        }
 
         // only consider enter button for keyboard events
         if (ev.type === 'keypress' && ev.key !== "Enter")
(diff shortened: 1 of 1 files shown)


pkg/lib/cockpit-components-logs-panel.jsx
```diff
@@ -63,11 +70,13 @@ export class JournalOutput {
...
 
+        const full_content = [time, message, ident].join("\n");
+
         return (
             <div className="cockpit-logline" role="row" tabIndex="0" key={entry.__CURSOR}
                 data-cursor={entry.__CURSOR}
-                onClick={ev => this.onEvent(ev, entry.__CURSOR)}
-                onKeyPress={ev => this.onEvent(ev, entry.__CURSOR)}>
+                onClick={ev => this.onEvent(ev, entry.__CURSOR, full_content)}
+                onKeyPress={ev => this.onEvent(ev, entry.__CURSOR, full_content)}>
```
jelly, line 79: Can't we make the onclick handler more specific? or is this done for mobile?

jelly:
Noticed this while clicking on a log.
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in JournalBox (created by LogsPage)
    in section (created by PageSection)
    in PageSection (created by LogsPage)
    in main (created by Page)
    in div (created by Page)
    in Page (created by LogsPage)
    ```

**Comment by jelly** [comment]
Unrelated but selecting and copy pasting something which includes time it is pasted as:

```
9:01 PM
error: kex_exchange_identification: Connection closed by remote host```

jelly (approved):

pkg/lib/cockpit-components-logs-panel.jsx
jelly: Yes, but as discussed on irc, we won't change that.

PR #16938, opened 2022-02-07
shell: Fix browser history

mvollmer:
The browser creates an entry in the history whenever an iframe changes
its "window.location" (which is how our pages implement
in-page-navigation).  The shell did not expect this and would add
another entry whenever the location of the current iframe changes.

---

### The back and forward buttons in your browser now work as expected

Cockpit used to interfere with the browser history, and hitting the back button would frequently get into a loop, for example. This has been fixed.

### DIFF
--- a/pkg/users/account-details.js
+++ b/pkg/users/account-details.js
@@ -226,7 +226,7 @@ export function AccountDetails({ accounts, groups, shadow, current_user, user })
                 </Title>
                 <EmptyStateSecondaryActions>
                     <Breadcrumb>
-                        <BreadcrumbItem onClick={() => cockpit.location.go("/")} to="#">{_("Back to accounts")}</BreadcrumbItem>
+                        <BreadcrumbItem to="#/">{_("Back to accounts")}</BreadcrumbItem>
                     </Breadcrumb>
                 </EmptyStateSecondaryActions>
             </EmptyState>
@@ -258,7 +258,7 @@ export function AccountDetails({ accounts, groups, shadow, current_user, user })
               id="account"
               breadcrumb={
                   <Breadcrumb>
-                      <BreadcrumbItem onClick={() => cockpit.location.go("/")} to="#">{_("Accounts")}</BreadcrumbItem>
+                      <BreadcrumbItem to="#/">{_("Accounts")}</BreadcrumbItem>
                       <BreadcrumbItem isActive>{title_name}</BreadcrumbItem>
                   </Breadcrumb>}>
             <PageSection>
--- a/pkg/apps/application.jsx
+++ b/pkg/apps/application.jsx
@@ -137,10 +137,6 @@ export const Application = ({ metainfo_db, id, progress, progress_title, action 
         );
     }
 
-    function navigate_up() {
-        cockpit.location.go("/");
-    }
-
     return (
         <Page groupProps={{ sticky: 'top' }}
               id="app-page"
@@ -148,7 +144,7 @@ export const Application = ({ metainfo_db, id, progress, progress_title, action 
               isBreadcrumbGrouped
               breadcrumb={
                   <Breadcrumb>
-                      <BreadcrumbItem className="pf-c-breadcrumb__link" onClick={navigate_up} to="#">{_("Applications")}</BreadcrumbItem>
+                      <BreadcrumbItem to="#/">{_("Applications")}</BreadcrumbItem>
                       <BreadcrumbItem isActive>{comp ? comp.name : id}</BreadcrumbItem>
                   </Breadcrumb>
               }>
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -394,7 +394,7 @@ class CurrentMetrics extends React.Component {
         }
 
         function serviceRow(name, value) {
-            const name_text = <Button variant="link" isInline component="a" key={name} href="#" onClick={ e => cockpit.jump("/system/services#/" + name + ".service") }><TableText wrapModifier="truncate">{name}</TableText></Button>;
+            const name_text = <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service") }><TableText wrapModifier="truncate">{name}</TableText></Button>;
             const value_text = <TableText wrapModifier="nowrap">{value}</TableText>;
             return {
                 cells: [{ title: name_text }, { title: value_text }]
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1426,7 +1426,7 @@ export const Application = () => {
                   <Flex>
                       <FlexItem>
                           <Breadcrumb>
-                              <BreadcrumbItem onClick={() => cockpit.jump("/system")} to="#">{_("Overview")}</BreadcrumbItem>
+                              <BreadcrumbItem onClick={() => cockpit.jump("/system")} className="pf-c-breadcrumb__link">{_("Overview")}</BreadcrumbItem>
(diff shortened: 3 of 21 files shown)


mvollmer:
Looks like this broke `cockpit.jump` somehow...

mvollmer:
> Looks like this broke `cockpit.jump` somehow...

No, tracking the hash needs to call `navigate` to get all the internal state updated.  `jump(..., false) would not do that, but `navigate` will call `jump`, so let's call `navigate` directly.

mvollmer:
I'll add a test.

garrett (approved):
Wow! This is already _so much better!_

It's not perfect, and there are some issues across some pages, but this feels like it's 99% solved.

Example I found where there's an issue:
- Navigate to accounts, then user, then back to accounts and try the back/forward buttons there.

But, provided the code and tests are good, I think this should probably be merged. The account issue is probably something else, as would be anything else that might be currently unknown.

We might even want to mention this fix in the release notes as it does impact user experience quite a bit and this was a longstanding bug. WDYT?

garrett (changes requested):
Please look into the issues I mentioned above. Perhaps they're outliers. If so, then I'll re-approve this. Otherwise, if they do affect this PR, then I'll be happy to review again later.

mvollmer:
>     1. Navigate to accounts, 
>     2. then a user,
>     3. then back to the accounts list

This means clicking the "Accounts" link in the bread crumbs, right?  This does indeed put two entries into the history instead of the expected single one.

mvollmer:
> This means clicking the "Accounts" link in the bread crumbs, right? This does indeed put two entries into the history instead of the expected single one.

All of our breadcrumbs put two entries into the history, even when not embedded into the shell, because of how they are implemented.

```
<BreadcrumbItem onClick={() => cockpit.location.go("/")} to="#">{_("Accounts")}</BreadcrumbItem>
```

This turns into

```
<li onClick=...>
  <a href="#">...</a>
</li>
```

Both the `a` and `li` will cause the browser to add a history entry. But this is not the whole story...

mvollmer:
> Both the `a` and `li` will cause the browser to add a history entry. But this is not the whole story...

We can use this:

```
 <BreadcrumbItem to="#/">{_("Accounts")}</BreadcrumbItem>
```

This will cause the browser to only add one entry.  Using `#/` instead of `#` is important:

- The shell will normalize a empty hash to "#/" (but only for the "src" attribute of the iframe)
- So when the iframe itself changes its hash to `#`, the shell will set it's `src` attribute to end in `#/`.
- This counts as another navigation event and the browser adds one more entry to the history.

mvollmer:
> Both the `a` and `li` will cause the browser to add a history entry.

And they only do that because they navigate to *different* locations, the `li` goes to "...#/" and the `a` goes to "...#".  So this also works:

```
<BreadcrumbItem onClick={() => cockpit.location.go("/")} to="#/">{_("Accounts")}</BreadcrumbItem>
```

but the onClick really shouldn't be necessary.

mvollmer:
> 2\. Click a service under CPU or Memory

This would two entries because this was also a link with both a click handler and a href="#".

mvollmer:
Another piece of the puzzle was that cockpit.jump with a hash would add two entries because the hash change alone would cause the browser to add one, and then the jump command would add another.

mvollmer:
> Another piece of the puzzle was that cockpit.jump with a hash would add two entries because the hash change alone would cause the browser to add one, and then the jump command would add another.

Unfortunately, the history that is added by the browser does not create a "popstate" event when removed from the history stack, the browser just undoes the hash change. Thus the frame change doesn't happen and things get confusing.

Hmm....

mvollmer:
> Unfortunately, the history that is added by the browser does not create a "popstate" event when removed from the history stack, the browser just undoes the hash change. Thus the frame change doesn't happen and things get confusing.

We can avoid the extra history entry from setting the "src" attribute by forcing the window.location to the right value beforehand, with window.location.replace.

garrett (approved):
I've been running the current version of this PR across browsers (Firefox, Chrome, GNOME Web, & Safari iOS) and haven't found an issue.

Congrats! I think this is solved! If there's anything not quite working, it would be a corner case — and I haven't found it in my latest round of random testing on various pages on various browsers.

The only one small "gotcha" is that the scroll state doesn't seem to be preserved. That is, scroll down a page, navigate to a new one, and hit back. The page should be at the same location as where you left it. (Same for going forward.) Browsers do this by default and it should be our behavior here too. It seems that something in the code has the pages jump to the top when it shouldn't.

However, I think this is a _HUGE_, dramatic improve over what we've had for several years, so I'm excited for this to go in ASAP. Well done! I think we can investigate the scroll issue as a follow-up.

garrett:
The History.scrollRestoration might need to be set:

- https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration
- https://html.spec.whatwg.org/multipage/history.html#she-scroll-restoration-mode

It's supposed to be set at the top of the document. (I wonder if we set it to manual somewhere.)

garrett:
I've looked through Cockpit code and don't see any obvious setting of history.scrollRestoration or scrollTo / scrollTop (except in updates.js, which wouldn't affect everything).

However, when grepping through dist, scrollTo / scrollTop shows up a LOT. (This was just a quick look.) I think it might be somewhere in React or PatternFly. It's all the more reason to land this PR and _then_ look into the scroll history stuff... it'll probably take some debugging and possibly messing with our stack. :cry: 

Meanwhile, what's in this PR is a 10000% improvement. :+1:

mvollmer:
> I've been running the current version of this PR across browsers (Firefox, Chrome, GNOME Web, & Safari iOS) and haven't found an issue.

YES! :-) Thanks a lot!

mvollmer:
> The only one small "gotcha" is that the scroll state doesn't seem to be preserved. 

Right, let's look at this in a separate PR.

pkg/users/account-details.js
```diff
@@ -226,7 +226,7 @@ export function AccountDetails({ accounts, groups, shadow, current_user, user })
                 </Title>
                 <EmptyStateSecondaryActions>
                     <Breadcrumb>
-                        <BreadcrumbItem onClick={() => cockpit.location.go("/")} to="#">{_("Back to accounts")}</BreadcrumbItem>
```
marusak, line 229: Do we need to fix this in other projects as well? If they don't fix it, does it break them or just the history is gonna be broken as before?

marusak (approved):
Works perfectly, thanks!

pkg/users/account-details.js
mvollmer: Just the history will be broken. We should check for this pattern also in the external projects.

garrett:
I think this is a sizable enough UX improvement to merit a mention in the release notes. Yes, it's a bug fix, but it's a very important one. Plus, we should have an update with Cockpit Client, which will have a visible UI change (addition of back/forward buttons). Both can be paired together in the final release notes.

pkg/users/account-details.js
mvollmer: I have checked c-machines, c-podman, and starter-kit. None of them uses this pattern.

mvollmer:
testHistory failure on coreos is not a flake, investigating...

mvollmer:
> testHistory failure on coreos is not a flake, investigating...

CoreOS does not seem to keep login history, so that part of the test simply needs to be skipped.

mvollmer:
@garrett, @marusak, the only changes since your last review were to get the test pass on CoreOS. No changes to Cockpit code.

garrett (approved):
Re-approving, as only tests were changed.

I'm looking forward to this landing!

martinpitt:
Let's land PR #16987 first to fix the fedora-coreos woes, and then retry.

PR #16940, opened 2022-02-08
users: Use `autocomplete` property for password inputs

marusak:
Mark fields that need new passwords so that browser does not try to
input remembered passwords but rather offers to generate a new one.
Also mark old password so in that case browser tries to offer remembered one.

Before when I tried to create new user it auto filled my current password and username:
Screenshot from 2022-02-08 09-25-52

Now it does not do that:
Screenshot from 2022-02-08 09-47-02

And it offers to create a new one that is then filled in both fields:
Screenshot from 2022-02-08 09-47-22

And when old one is needed it is still correctly auto set:
Screenshot from 2022-02-08 09-46-38

### DIFF
--- a/pkg/lib/cockpit-components-password.jsx
+++ b/pkg/lib/cockpit-components-password.jsx
@@ -78,7 +78,7 @@ export const PasswordFormFields = ({
                        validated={error_password ? "error" : "default"}
                        fieldId={idPrefix + "-pw1"}>
                 <TextInput className="check-passwords" type="password" id={idPrefix + "-pw1"}
-                           value={password} onChange={value => change("password", value)} />
+                           autocomplete="new-password" value={password} onChange={value => change("password", value)} />
                 <div>
                     <Progress id={idPrefix + "-meter"}
                               className={"ct-password-strength-meter " + variant}
@@ -95,7 +95,7 @@ export const PasswordFormFields = ({
                        helperTextInvalid={error_password_confirm}
                        validated={error_password_confirm ? "error" : "default"}
                        fieldId={idPrefix + "-pw2"}>
-                <TextInput type="password" id={idPrefix + "-pw2"}
+                <TextInput type="password" id={idPrefix + "-pw2"} autocomplete="new-password"
                            value={password_confirm} onChange={value => change("password_confirm", value)} />
             </FormGroup>}
         </>
--- a/pkg/users/password-dialogs.js
+++ b/pkg/users/password-dialogs.js
@@ -130,19 +130,22 @@ export function passwd_change(user, new_pass) {
 function SetPasswordDialogBody({ state, errors, change }) {
     const {
         need_old, password_old, password, password_confirm,
-        password_strength, password_message
+        password_strength, password_message, current_user,
     } = state;
 
     return (
         <Form isHorizontal onSubmit={apply_modal_dialog}>
             { need_old &&
-            <FormGroup label={_("Old password")}
-                       helperTextInvalid={errors && errors.password_old}
-                       validated={(errors && errors.password_old) ? "error" : "default"}
-                       fieldId="account-set-password-old">
-                <TextInput className="check-passwords" type="password" id="account-set-password-old"
-                           autocomplete="current-password" value={password_old} onChange={value => change("password_old", value)} />
-            </FormGroup> }
+            <>
+                <input hidden disabled value={current_user} />
+                <FormGroup label={_("Old password")}
+                           helperTextInvalid={errors && errors.password_old}
+                           validated={(errors && errors.password_old) ? "error" : "default"}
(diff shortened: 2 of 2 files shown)


marusak:
> And when old one is needed it is still correctly auto set:

To make it even more useful, we might want to add hidden user name input with prefilled username so that browser knows which password to put there when more than one login for cockpit are remembered.

marusak:
Ok, added also hidden input and now it auto completes the right password.

garrett:
This is tricky. Chrome doesn't follow the autocomplete spec.

- https://stackoverflow.com/questions/48150635/autocomplete-new-password-ignored-by-chrome-63-in-windows/49204211#49204211
- https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off/16130452#16130452
- https://stackoverflow.com/questions/25823448/ng-form-and-autocomplete-off/39689037#39689037
- https://stackoverflow.com/questions/53179114/unexpected-chrome-autofill-behaviour-disable-chrome-autofill/53193537#53193537

They've gone back and forth and back again on not supporting it / supporting it at various times. Right now, they're out of spec again.

A choice quote from the last link says: "As of recent Chrome (definitely version 70) autocomplete="off" is now respected, as long as your inputs do not look like user profile, address or credit card data."

And ours looks like user profile data, naturally, as we're letting users edit not just their own user profile data, but others. Sigh.

What we _can_ do is set a unique string for this. It's actually what I've recommended to PatternFly for a related issue for their autocomplete forms. (Not password related.)

So it would be possible to do something like this JS:

```js
autocomplete = "user-" + userName + "-" + crypto.getRandomValues(new Uint32Array(1))[0];
```

That would prevent hitting autocomplete by polluting the autocomplete value. I went a step further for PF and suggest that only be enabled for Chromium, like this:

```js
// Turn off autocomplete, with a workaround for Chrome and Chromium-based browsers which do not follow the spec
autocomplete = navigator.userAgent.match("Chrome") ? "user-" + userName + "-" + crypto.getRandomValues(new Uint32Array(1))[0] : "off";
```

Normally I'm completely against browser sniffing. But this is one of the rare cases where it's actually necessary. This is a side effect of Google being the 2000 kg gorilla in the room. Google basically has a monopoly on browser engines, and broken stuff like this happens when they decide to not follow the spec. (This happens a LOT, especially on new, incomplete, broken features they're trying to push. But autocomplete's a great example of something simple where they deviate from agreed upon specs and end up breaking the web.)

If we do want autocomplete, then the solution would be to set the autocomplete value to something unique but repeatable so it's stored by the browser and recalled, but partitioned based on the user of the system.

```js
// Set autocomplete to a string so the autocomplete only applies to the same user on the same system
autocomplete = "cockpit-account-" + hostName + "-" + userName 
```

The above examples include workarounds that work on all browsers. And they should be resilient, as they depend on the value of autocomplete being an unreserved string, which is valid.

marusak:
So I just tested this in chrome and it works all fine.

- The main issue was, that when you open 'Create account' dialog now in either chrome or firefox it autofills username and password - which is never correct. With this PR neither browser does that anymore. :heavy_check_mark: 

- The other issue was that if you have more remembered passwords browser would not know which password to fill in - now both browsers fill in the correct password. :heavy_check_mark: 

The link you provided talk about disabling autocomplete. I don't want to disable it, I just don't want browser to autofill my fields when I am creating a new user. That works nicely now. One thing I noticed, is that when I click on "new password" field it still offers remembered logins alongside with "generate new" option - but that is the same behavior in both chrome and firefox.

So from my point of view, this PR improves it equally on both browsers.

garrett:
Hey, look! Chrome suggested I sign in as 'bob':

Screenshot from 2022-02-10 17-05-43

I think this is a problem.

garrett (changes requested):
This is obviously an improvement over browsers suggesting your current password. However, it's problematic after entering passwords for other users.

We could either change it to be proper in this PR or submit this with as-is (as it's better than what we have) and consider a follow up to fix it in a way that the browser does not attempt to save the password.

Marking as request changes for now, unless we decide to handle this via a follow-up.

garrett:
Note: I actively did save the password on Chrome, so it might not be quite as bad as I point out. I need to double check what happens on other browsers though. Some always ask with a prompt, from what I remember, which could be quite annoying. Double checking now.

garrett:
Yep, Edge pops this up when you change pages, and it's probably too easy to accidentally click the big blue button to make the prompt go away:

Screenshot from 2022-02-10 17-11-34

Firefox:

image

garrett (approved):
Have you tried visiting another page after saving the password? Some browsers do auto popup in that scenario.

But, again, I think this PR is already a lot better than what we have and we should merge this in as-is.

PR #16942, opened 2022-02-08
run-tests dynamic serial test scheduling, drop static --batches, show number of parallel tests

martinpitt:
This reworks the run-tests parallel scheduler for serial tests in a very fundamental way.

Computing the batches in advance by just splitting the list into equal parts does not work well: It does not take retries and wildy different durations into account, and thus often created a wild imbalance beween the total runtime on the serial runner machines:
    
        4592s on [host], 118 serial tests: 0: 4589s, 1: 2078s
    
In many cases this even led to a serial test batch dominate the total test time.

About half of the current weather report's "> 1 hour" test runs are due to this problem. [example 1](https://logs.cockpit-project.org/logs/pull-16904-20220202-163951-7852b004-fedora-35-firefox/log), [example 2](https://logs.cockpit-project.org/logs/pull-16904-20220202-163953-7852b004-fedora-35/log)

 - builds on top of: #16952 
 - requires: #16955
 - requires: #16962

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -11,6 +11,7 @@ import time
 import socket
 import tempfile
 import binascii
+import logging
 
 import importlib.machinery
 import importlib.util
@@ -183,6 +184,11 @@ class GlobalMachine:
         if not os.path.exists(self.machine.image_file):
             self.machine.pull(self.machine.image_file)
         self.machine.start()
+        self.start_time = time.time()
+        self.duration = None
+        self.ssh_address = f"{self.machine.ssh_address}:{self.machine.ssh_port}"
+        self.web_address = f"{self.machine.web_address}:{self.machine.web_port}"
+        self.running_test = None
 
     def reset(self):
         # It is important to re-use self.networking here, so that the
@@ -192,8 +198,16 @@ class GlobalMachine:
         self.machine.start()
 
     def kill(self):
+        assert self.running_test is None, "can't kill global machine with running test"
         self.machine.kill()
         self.network.kill()
+        self.duration = round(time.time() - self.start_time)
+        self.machine = None
+        self.ssh_address = None
+        self.web_address = None
+
+    def is_available(self):
+        return self.machine and self.running_test is None
 
 
 def check_valid(filename):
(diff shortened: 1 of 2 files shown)


test/common/run-tests
```diff
@@ -292,147 +306,134 @@ def detect_tests(test_files, image, opts):
...
+
+    # create appropriate number of serial machines; assume a coarse approximation that a typical destructive (parallel)
+    # test takes twice the time of a typical serial test; the goal is to not over-parallelize serial tests if there
+    # are lots of parallel ones (to detect inter-test dependencies), but use maximum parallelization if the test suite
+    # only has a few or even no parallel tests
+    if serial_tests_len > 0:
+        num_global = max(1, round(serial_tests_len / (serial_tests_len + 2 * parallel_tests_len) * opts.jobs))
+    else:
+        num_global = 0
+    logging.debug(f"allocating {num_global} global machines for {serial_tests_len} serial and {parallel_tests_len} parallel tests")
```
allisonkarlitskaya, line 345: I kinda disagree with the logic here.  I'm not sure if semi-reliable detection of inter-test dependencies is a desirable characteristic or not, but I feel like we could accomplish that by randomising the queue.

How about this (rather simple) approach:
 - ignore parallel tests at first
 - start `n` serial job runners for `n = min(len(serial_tests), opts.jobs)`, and have them consume `@nd` tests from the queue
 - as each serial runner finishes its last job, have it switch over to running parallel tests

This feels like a better way to keep all cores as busy as possible, without making any generalized guess about the relative required running time of the parallel vs. serial tests.

allisonkarlitskaya: One more point: the names "serial" and "parallel" are awful here.  I wonder if we could get rid of them?

martinpitt:
Also [needs fixing](http://artifacts.dev.testing-farm.io/913e7f5f-156e-40e7-b59a-bb919bf07b8e/) with `--machine`.

test/common/run-tests
martinpitt: FTR: I considered this, and discarded it as the original rationale still seems to make sense. But indeed, let's do this simple approach, check our statistics in a few weeks, and let the data talk. This is super-easy to tweak now.

martinpitt:
On the plus side, tests are quite a bit faster now. E.g. [rhel-9-0](https://logs.cockpit-project.org/logs/pull-16942-20220208-132850-836165a4-rhel-9-0/log):

    [2727s on 3-cockpit-9, 187 parallel tests, 116 serial tests: 0: 628s, 1: 579s, 2: 580s, 3: 581s, 4: 574s]

so the nondestructive tests are nicely balanced, and 45 mins is reasonably fast (including 15 retries).

However, this also triggered a few new flakes, due to the denser test packing.

martinpitt:
[testPmProxySettings on debian-stable](https://logs.cockpit-project.org/logs/pull-16942-20220208-132850-836165a4-debian-stable/log.html#61-2): The [journal](https://logs.cockpit-project.org/logs/pull-16942-20220208-132850-836165a4-debian-stable/TestHistoryMetrics-testPmProxySettings-debian-stable-127.0.0.2-2201-FAIL.log.gz) does not have any specifics about why it fails with "protocol-error". What could probably help is to stop/reset-fail all units at the start of the test, so that it runs in a defined state, and a retry works?

test/verify/check-networkmanager-basic
```diff
@@ -42,6 +42,12 @@ class TestNetworkingBasic(NetworkCase):
         b.wait_in_text("#networking-graphs", "Transmitting")
         b.wait_in_text("#networking-graphs", "Receiving")
 
+        if not m.ostree_image:
+            # screenshot assumes running firewalld and absent virbr0 (in particular, *3* interfaces)
+            m.execute("systemctl start firewalld; virsh net-destroy default 2>/dev/null || true")
+            b.wait_not_present("tr[data-interface=virbr0]")
+            self.wait_onoff("#networking-firewall-summary", True)
```
allisonkarlitskaya: I'm not sure how I feel about this...

If we take extra steps just for the purpose of making sure the pixels look good, then it seems like those ought to be conditionalized on if we're actually running the pixels tests.

But I think a better way to look at it is that it's wrong that we rely on the state of a previously-running test because it's wrong that that test left that state: it's a case of non-nondestructiveness, which ought to be fixed.

test/common/run-tests
```diff
@@ -344,8 +344,11 @@ def run(opts, image):
     global_machines = []
 
     if not opts.machine:
-        # create serial machines; adjust for few/no serial tests
-        for i in range(min(opts.batches, len(serial_tests))):
+        # Create appropriate number of serial machines; prioritize the nondestructive (serial) tests, to get
+        # them out of the way as fast as possible, then let the destructive (parallel) ones "float around" them
```
allisonkarlitskaya, line 348: "float around" is confusing.  I think you mean "start running as soon as a given serial runner is finished" or so?

test/common/run-tests
```diff
@@ -463,8 +466,6 @@ def main():
                         help="Directory in which to glob check-* files; default: %(default)s")
     parser.add_argument('--exclude', action="append", default=[], metavar="TestClass.testName",
                         help="Exclude test (exact match only); can be specified multiple times")
-    parser.add_argument('-b', '--batches', type=int,
-                        help="Number of concurrent batches of nondestructive tests")
```
allisonkarlitskaya, line 456: :+1:

test/common/run-tests
```diff
@@ -301,139 +315,135 @@ def detect_tests(test_files, image, opts):
     return (serial_tests, parallel_tests)
 
 
-def run(opts, image):
-    result = 0
-    jobs = 1 if opts.list else opts.jobs
-    start_time = time.time()
+def list_tests(opts):
+    test_files = glob.glob(os.path.join(opts.test_dir, "check-*"))
+    (serial_tests, parallel_tests) = detect_tests(test_files, "dummy", opts)
```
allisonkarlitskaya, line 320: it's more idiomatic without the parens

test/common/run-tests
```diff
@@ -301,139 +315,135 @@ def detect_tests(test_files, image, opts):
...
+    names = [t.command[-1] for t in serial_tests] + [t.command[-1] for t in parallel_tests]
```
allisonkarlitskaya, line 321: `serial_tests` and `parallel_tests` are both lists, so you can DRY a bit here.  Also: what you're building is conceptually a set, so:

`names = {t.command[-1] for t in serial_tests + parallel_tests}` or so

test/common/run-tests
```diff
@@ -301,139 +315,135 @@ def detect_tests(test_files, image, opts):
...
+    names.sort()
+    for n in names:
```
allisonkarlitskaya, line 323: `for n in sorted(names):`
is nice

test/common/run-tests
```diff
@@ -301,139 +315,135 @@ def detect_tests(test_files, image, opts):
...
+            if not running_tests and serial_tests:
+                test = serial_tests.pop(0)
+                logging.debug(f"Static machine is free, assigning next test {test}")
+                test.assign_machine(-1, opts.machine, opts.browser)
+                test.start()
+                running_tests.append(test)
+                made_progress = True
+        else:
+            # find free global machines, and either assign a new serial test, or kill them to free resources
+            for (idx, machine) in enumerate(global_machines):
```
allisonkarlitskaya, line 400: parens

martinpitt:
The two mega-PRs with fixing three handfuls of tests landed, and I think/hope this will get fairly green. I just pushed another fix (to retry tests right away instead of putting them back at the end of the queue), so this is good for a first round of review. (Mind you, it may still expose more test bugs, hence keeping draft for now)

test/common/run-tests
```diff
@@ -301,139 +315,135 @@ def detect_tests(test_files, image, opts):
...
+                        test.start()
+                        running_tests.append(test)
+                    else:
+                        logging.debug(f"Global machine {idx} is free, and no more serial tests; killing")
+                        machine.kill()
+
+                    made_progress = True
+
+        # fill the remaining available job slots with parallel tests
+        while parallel_tests and len(running_tests) < opts.jobs:
```
allisonkarlitskaya, line 416: The logic here is funny.  For some reason I kinda thought that we'd rather look at the number of running serial machines before going to start parallel jobs, but since the code above pretty much ensures that there is the exact number of tests running as there are serial machines, then this is equivalent...

So fine, but not what I was expecting to see.  Probably better, though, since we don't need to keep track of the number of running "parallel" tests separately.

allisonkarlitskaya (commented):
This is very very very cool work.  I have a few suggestions for syntax tweaks and fixing up commit messages, but otherwise it's ready.

test/verify/check-networkmanager-basic
martinpitt: Right, the line between "leave the VM broken" and "don't make overly harsh assumptions about the VM" are blurry. However, here I feel that it's the test's responsibility to set up the conditions that it needs. Especially for virbr0, as that's such a dynamic thing -- it's equally valid and plausible for it to be up or down (a simple `virsh net-list` will bring it up, due to socket activations and daemon autostart!). So in this case my gut feeling went for this approach, which will also help us with gating (where the condition of the testbed is not strictly defined either)

martinpitt:
Some more flakes:

 - [missing pixel test ref image](https://logs.cockpit-project.org/logs/pull-16942-20220209-124540-22d91db9-fedora-35/log.html#89) - it's a WTH??? :see_no_evil: case; @mvollmer says he has seen it once so far, but too dubious to track down now
 - [ksmtuned SELinux violation](https://logs.cockpit-project.org/logs/pull-16942-20220209-123107-22d91db9-rhel-8-6/log.html#106); I filed a [downstream bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2052554) and a [PR to add a naughty](https://github.com/cockpit-project/bots/pull/2920)
 - [timeout loading the Services page](https://logs.cockpit-project.org/logs/pull-16942-20220209-123107-22d91db9-rhel-8-6/log.html#291); this does take very long, and on an overloaded machine it may very well exceed the standard one minute. I'll bump the timeout.
 - [TestNetworkingBasic.testBasic](http://artifacts.dev.testing-farm.io/c3d6699f-6a6a-42be-a7c6-4471594e6228/) failure on TF fedora-34; trouble with shutting down virbr0?

test/common/run-tests
martinpitt: Updated.

test/common/run-tests
martinpitt: FTR: I actually thought in-place sorting of an list was more efficient than sorting an iterator; but it hardly matters here.

martinpitt:
@allisonkarlitskaya : I addressed your comments, added fixes for three more test flakes, let's give it another round of bots and review. Thanks!

martinpitt:
This is starting to look good.. The only failures were an ages old [dirty tasks container problem](https://logs.cockpit-project.org/logs/pull-16942-20220210-061520-ae80e711-rhel-8-6-distropkg/log.html) and [this flake](https://logs.cockpit-project.org/logs/pull-16942-20220210-061520-ae80e711-ubuntu-stable/log.html#131-2) which we had had forever. [c8s packit](http://artifacts.dev.testing-farm.io/66ec30f1-38df-49ef-9275-0ac34828e780/) is also well known, the "couldn't create runtime dir: /run/user/1001: Permission denied" error (been around for a long time, I have absolutely no idea about it)

martinpitt:
This run's flakes:

 - fcos](https://logs.cockpit-project.org/logs/pull-16942-20220210-091454-a2b3c82a-fedora-coreos/log.html#74) is known, happens on main, and tracked in issue #16970.
 - The green tests have one or two retries at most, within our "usual" range
 - [debian-testing](https://logs.cockpit-project.org/logs/pull-16942-20220210-091454-a2b3c82a-debian-stable/log.html#59-2) is still the pcp "protocol error" from above; this will need a naughty, it's a distro bug
 - [this virbr0 networking failure](https://logs.cockpit-project.org/logs/pull-16942-20220210-091454-a2b3c82a-fedora-35/log.html#68) is something which I thought I fixed yesterday -- but apparently not hard enough

I'll look at these, but I think these don't invalidate the general approach here, they might just a bit more apparent now.

martinpitt:
I have an idea about these [missing pixel test files](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16942-20220210-104937-a2b3c82a-fedora-35/log.html#267). The tests run in parallel, one is calling `pixel-tests pull` while the other sees that `test/reference/.git` exists, but the checkout isn't complete yet. There needs to be some locking involved.

allisonkarlitskaya:
This is a really really nice improvement.  It's gonna shake out a lot of existing bugs, too, which is actually a good thing.  Let's hope the next days aren't too bad ;)

PR #16943, opened 2022-02-08
Moar nondestructive tests

jelly:


test/verify/check-system-info
```diff
@@ -391,10 +391,12 @@ class TestSystemInfo(MachineCase):
         # restarts timesyncd to pick up the new config
         wait(lambda: get_timesyncd_start() > prev_timesyncd_start, delay=0.2)
 
+    @nondestructive
```
martinpitt, line 394: commit message subject has an excess "to"

martinpitt (changes requested):
Looks nice, thanks! However, please drop the no-test before the final run.

jelly:
> Looks nice, thanks! However, please drop the no-test before the final run.

Ofcourse, I have some more WIP stuff locally, just wanted to run a single distro before running it on all as one distro should already shown some breakages.

test/verify/check-users-roles
```diff
@@ -103,10 +103,12 @@ class TestRoles(MachineCase):
 
         self.allow_restart_journal_messages()
 
+    @nondestructive
     def testDynamic(self):
         m = self.machine
         b = self.browser
 
+        self.addCleanup(m.execute, "groupdel docker || true")
```
martinpitt, line 112: We shouldn't need this, testlib.py already restores passwd/group and so on. But of course it doesn't really hurt.

martinpitt (approved):
Nice, thanks!

PR #16944, opened 2022-02-08
decrease our Python line length limit to 300

allisonkarlitskaya:
I'd like to get this to ~118 or so, eventually, but here's some extra-bad cases to start the motion in that direction....

 - [x] #16950


test/verify/check-shell-keys
```diff
@@ -21,9 +21,9 @@ import parent
 from testlib import *
 
 
-FP_MD5 = "93:40:9e:67:82:78:a8:99:89:39:d5:ba:e0:50:70:e1"
-FP_SHA256 = "SHA256:SRvBhCmkCEVnJ6ascVH0AoVEbS3nPbowZkNevJnXtgw"
-KEY = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDG4iipTovcMg0xn+089QLNKVGpP2Pgq2duxHgAXre2XgA3dZL+kooioGFwBQSEjbWssKy82hKIN/W82/lQtL6krf7JQWnT3LZwD5DPsvHFKhOLghbiFzSI0uEL4NFFcZOMo5tGLrM5LsZsaIkkv5QkAE0tHIyeYinK6dQ2d8ZsfmgqxHDUQUWnz1T75X9fWQsUugSWI+8xAe0cfa4qZRz/IC+K7DEB3x4Ot5pl8FBuydJj/gb+Lwo2Vs27/d87W/0KHCqOHNwaVC8RBb1WcmXRDDetLGH1A9m5x7Ip/KU/cyvWWxw8S4VKZkTIcrGUhFYJDnjtE3Axz+D7agtps41t test-name"
```
martinpitt, line 26: I think I saw a test failure around that yesterday -- if this leads to non-trivial problems, I recommend to just break this up into a multi-line string for this PR. (It seems useful to test RSA keys somewhere, after all)

allisonkarlitskaya: Honestly, this is a place I'd rather use #noqa.

...or generate the key at run time...

allisonkarlitskaya: There are no fails related to this in the tests.  Just the curl thing...

Do you still prefer me to put back RSA?

allisonkarlitskaya: eh, I take it back?  What is this failing in *only* Firefox?  That needs to be looked into...

martinpitt: I'm fine with moving to EC.

The [firefox failure](https://logs.cockpit-project.org/logs/pull-16944-20220209-055643-eee3a6c4-fedora-35-firefox/log.html#122) is "just" an unexpected message, looks like this test is missing `allow_restart_journal_messages()`? I suppose it does not actually *re*start, but that function enables the messages for "sudden" disconnects.

allisonkarlitskaya: Hilarious.  Trying to reproduce that locally gets me another fun fail: `testlib.Error: FAIL: Test completed, but found unexpected journal messages:
audit: type=1400 audit(1644413645.103:237): avc:  denied  { execute } for  pid=2126 comm="chpasswd" name="sss_cache" dev="vda5" ino=39569 scontext=unconfined_u:unconfined_r:passwd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:sssd_exec_t:s0 tclass=file permissive=0`

martinpitt: @allisonkarlitskaya : That is just our latest OS regression: https://github.com/cockpit-project/bots/issues/2898

martinpitt:
Fixed the trivial conflict and rebased. Let's see where we are here, as this looks good, and unblocks sharing test/static-code with c-machines.

pkg/storaged/luksmeta-monitor-hack.py
```diff
@@ -93,18 +93,15 @@ def info(dev):
...
-                        config = json.dumps(get_clevis_config_from_protected_header(token_object["jwe"]["protected"]))
-                        for slot_str in token_object.get("keyslots", []):
-                            slot = int(slot_str)
-                            slots[slot] = {"Index": {"v": slot},
-                                           "ClevisConfig": {"v": config}}
-                except subprocess.CalledProcessError:
-                    pass
+                result = subprocess.run(["cryptsetup", "token", "export", dev, "--token-id", match.group(1)],
+                                        check=False, capture_output=True, universal_newlines=True)
+                token_object = json.loads(result.stdout)
```
martinpitt, line 98: This looks broken, or at least dubious. With `check=False`, we need to validate whether calling the command was successful before processing its output. This may or may not be the cause of [this regression](https://logs.cockpit-project.org/logs/pull-16944-20220420-122609-856b7ebc-rhel-8-6-distropkg/log.html#184-2).

However, I kinda disagree that this was out of style -- catching exceptions and assuming silent success is Pythonic.

martinpitt:
[r86](https://logs.cockpit-project.org/logs/pull-16944-20220420-122609-856b7ebc-rhel-8-6/log.html) and [c8s](https://logs.cockpit-project.org/logs/pull-16944-20220420-122609-856b7ebc-centos-8-stream/log.html) failed as well -- I figure its Python 3 is too old to support `subprocess.run`. I backed out the commit, and locally it's happy again.

pkg/storaged/luksmeta-monitor-hack.py
martinpitt: See below, backed out.

martinpitt:
[testInconsistentCryptoPolicy](https://logs.cockpit-project.org/logs/pull-16944-20220420-152532-e0814f83-centos-8-stream/log.html#249) failed twice in a row on c8s. This is oddly specific, and this test doesn't touch any related file, but it happened twice in a row. I'll try this locally.

allisonkarlitskaya:
Thanks for cleaning this up.

...but I can't review it :)

martinpitt (approved):

PR #16951, opened 2022-02-08
check-connection: use m.write_conf()

allisonkarlitskaya:
Replace a couple of cases of us manually writing to cockpit.conf with
the new API for this from the bots.

 - [ ] cockpit-project/bots#2915


martinpitt:
We already have `m.write_file()`. Perhaps it would rather make more sense to make that clever enough to do `mkdir -p` first? Possibly as an option? That seems more generally useful to me than this special case?

allisonkarlitskaya:
Can't do this without the bots API, and the bots API got nack'd, so let's set this one aside.

PR #16955, opened 2022-02-09
fix networking and services pixel tests

martinpitt:
The previous networking pixel tests were wildly inconsistent in their
assumed network interfaces: The references for basic and team had no
"virbr0", while the references for bond and bridge did. This was not a
conscious decision, just an arbitrary result of in whichever order
the tests happened to run when generating the references.

Stop the libvirt default network (i.e. virbr0) at the beginning of all
of these tests. In addition, the basic test assumed that firewalld was
running, so start it to actually make sure of that.

This makes these tests independent from whether
check-networkmanager-firewall runs before or after them.

----

This addresses the first four failures of [this mess](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16942-20220208-185207-b1fb7771-fedora-35/log.html).

This was reproducible by running

    test/verify/check-networkmanager-firewall TestFirewall.testFirewallPage $RUNC

before the network tests. I now ran all four networking tests before and after the firewall tests, and everything is happy.

[changed pixels](https://github.com/cockpit-project/pixel-test-reference/compare/bed9ed36da1a31965361ad19c110cb2aef50d95f..bf328a02249b3c5334c6f7983ba941d4f56fb302#diff-afc19c4b64926669ecc01eac7cf35559b3760cf8dc45fbb89fe9da7234f7e719)

### DIFF
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -328,7 +328,10 @@ Unit=test.service
         self.toggle_onoff()
         self.check_service_details(["Failed to start", "Automatically starts"], ["Start", "Clear 'Failed to start'", "Disallow running (mask)"], True)
         if not user:
-            b.assert_pixels("#service-details", "details-test-fail", ignore=[".cockpit-log-panel .pf-c-card__body", ".pf-c-switch__toggle"])
+            b.assert_pixels("#service-details", "details-test-fail",
+                            # in medium layout we sometimes get a scrollbar depending on how may test-fail logs exist
+                            skip_layouts=["medium"],
+                            ignore=["#service-details-logs", ".pf-c-switch__toggle"])
         b.click(".action-button:contains('Start service')")
         b.go(f'/system/services#/{suffix}')
         self.wait_service_present("test-fail.service")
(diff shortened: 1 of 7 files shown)


martinpitt:
OK, this [repeats](https://logs.cockpit-project.org/logs/pull-16955-20220209-101203-ca9d3982-fedora-35/log.html#67) -- I really *really* need to get out of my current PR dependency hell, so for now I'll just disable the medium scenario.

test/verify/check-system-services
```diff
@@ -328,7 +328,10 @@ Unit=test.service
         self.toggle_onoff()
         self.check_service_details(["Failed to start", "Automatically starts"], ["Start", "Clear 'Failed to start'", "Disallow running (mask)"], True)
         if not user:
-            b.assert_pixels("#service-details", "details-test-fail", ignore=[".cockpit-log-panel .pf-c-card__body", ".pf-c-switch__toggle"])
+            b.assert_pixels("#service-details", "details-test-fail",
+                            # in medium layout we sometimes get a scrollbar depending on how may test-fail logs exist
```
mvollmer, line 332: "many"

martinpitt: Ugh -- I'll fix the typo in a follow-up commit in #16942, ok?

PR #16958, opened 2022-02-09
webserver: add support for X-Forwarded-For

### LINKED ISSUE #16927: Is there any way to let cockpit accept XFF header?
It's possible that let cockpit accepts the X-FORWARDED-FROM header from localhost? this [document](https://cockpit-project.org/guide/latest/cockpit.conf.5.html) just says make it accept X-Forwarded-Proto, but if without XFF header the cockpit can log correct IP address,and without Nginx i can't use ssl_client_certificate

allisonkarlitskaya:
Add a "forwarded for" header option to cockpitwebserver, and wire it through from the config file in cockpit-ws.  Add docs and unit tests.  Also verify that this is working as part of our existing nginx reverse proxy test.

Fixes #16927


src/common/test-webserver.c
```diff
@@ -34,22 +34,22 @@ typedef struct {
...
-} TestCase;
+} TestFixture;
 
 typedef struct {
   const gchar *cert_file;
   gboolean local_only;
   gboolean inet_only;
   CockpitWebServerFlags server_flags;
-} TestFixture;
+} TestCase;
```
martinpitt, line 53: Hah, thanks. This affects pretty much every unit test, so one of these days you could perhaps apply that seddery to the others. :grinning:

src/common/test-webserver.c
```diff
@@ -34,17 +34,38 @@ typedef struct {
   CockpitWebServer *web_server;
   gchar *localport;
   gchar *hostport;
+
+  const gchar *expected_protocol;
+  gchar *expected_remote;
```
martinpitt, line 39: It feels wrong to put expectations into the fixture. This should be the actual state, i. e. just `protocol` and `remote`? Expectiations should live in the TestCase only.

src/common/test-webserver.c
```diff
@@ -82,6 +103,17 @@ setup (TestFixture *fixture,
                                            test_case ? test_case->server_flags : COCKPIT_WEB_SERVER_NONE);
   g_clear_object (&cert);
 
+  /* We want to check all incoming requests to ensure that they match
+   * our expectations about remote hostname and protocol.  Add a
+   * "handler" that does that, but never claims to handle anything.
+   */
+  fixture->expected_remote = g_strdup (address);
+  if (test_case && test_case->expected_protocol)
+    fixture->expected_protocol = test_case->expected_protocol;
```
martinpitt, line 125: That reads really strange. If the test case overwrites the current state (fixture), what is there to check? It'd feel cleaner to not initialize it in the fixture at all, and move this check to verify_request().

src/common/test-webserver.c
```diff
@@ -1113,5 +1135,38 @@ main (int argc,
   g_test_add ("/web-server/for-tls-proxy", TestFixture, &test_case_for_tls_proxy,
               setup, test_webserver_for_tls_proxy, teardown);
 
+  /* Header is enabled, but not passed.  Default to "http" */
+  static const TestCase test_case_proto_header_empty = {
```
martinpitt, line 1146: I find it quite nice to declare the TestCase right here; but for consistency, please move the existing global ones as well (in a preceeding commit), or move the new ones to the global scope as well.

src/common/cockpitwebserver.c
```diff
@@ -1371,38 +1371,32 @@ cockpit_web_request_get_protocol (CockpitWebRequest *self)
 gchar *
 cockpit_web_request_get_remote_address (CockpitWebRequest *self)
 {
-  gchar *result = NULL;
```
martinpitt, line 1374: Hah, good one. Supposedly this was a leftover from before we had the g_auto* magic :grin:

doc/man/cockpit.conf.xml
```diff
@@ -76,6 +76,20 @@ Origins = https://somedomain1.com https://somedomain2.com:9090
...
 [WebService]
 ProtocolHeader = X-Forwarded-Proto
+</programlisting>
+          </informalexample>
+        </listitem>
+      </varlistentry>
+        <varlistentry>
+          <term><option>ForwardedForHeader</option></term>
+          <listitem>
+            <para>Configure cockpit to look at the contents of this header to determine the real origin of a
```
martinpitt, line 86: Is `X-F-F` some reasonably widely accepted standard? Does nginx/Apache etc. have some recommendations or builtin defaults around that? Then the documentation should recommend to use X-F-F.

doc/man/cockpit.conf.xml
```diff
@@ -76,6 +76,20 @@ Origins = https://somedomain1.com https://somedomain2.com:9090
...
+              connection.  This should only be used when cockpit is behind a reverse proxy, and care
+              should be taken to make sure that incoming requests cannot set this header.</para>
+          <informalexample>
+<programlisting language="js">
```
martinpitt, line 90: Certainly not "js", more like "ini"? I know, "just" copy&pasta from the previous example, but let's fix it while we are at it.

src/common/cockpitwebserver.c
```diff
@@ -1371,6 +1381,23 @@ cockpit_web_request_get_protocol (CockpitWebRequest *self)
...
+      const gchar *forwarded_header = g_hash_table_lookup (self->headers, self->web_server->forwarded_for_header);
+      if (forwarded_header)
+        {
+          /* This isn't really standardised, but in practice, it's a
+           * space separated list and the last item is from the
+           * immediately upstream server.
+           */
+          const gchar *last_space = strrchr (forwarded_header, ' ');
+          if (last_space)
+            return g_strdup (last_space + 1);
```
martinpitt, line 1395: Is there any chance that trailing spaces could happen? (I hope not..) But not sure if this should prevent returning an empty string.

src/common/test-webserver.c
```diff
@@ -106,14 +108,19 @@ setup (TestFixture *fixture,
...
   if (test_case && test_case->protocol_header)
     cockpit_web_server_set_protocol_header (fixture->web_server, test_case->protocol_header);
 
   /* We want to check all incoming requests to ensure that they match
    * our expectations about remote hostname and protocol.  Add a
    * "handler" that does that, but never claims to handle anything.
    */
-  fixture->expected_remote = g_strdup (address);
+  if (test_case && test_case->expected_remote)
+    fixture->expected_remote = g_strdup (test_case->expected_remote);
```
martinpitt, line 121: Same thing as in the first commit; this is weird IMHO

src/common/test-webserver.c
```diff
@@ -1168,5 +1175,66 @@ main (int argc,
...
+    .extra_headers = "X-Forwarded-For: 6.6.6.6 2.2.2.2 1.2.3.4\r\n",
+    .expected_remote = "1.2.3.4",
+  };
+  g_test_add ("/web-server/forwarded-for/multiple", TestFixture, &test_case_forwarded_for_multiple,
+              setup, test_webserver_with_headers, teardown);
+
+  /* Header enabled, and passed multiple IPs, and junk */
+  static const TestCase test_case_forwarded_for_junk = {
+    .forwarded_for_header = "X-Forwarded-For",
+    .extra_headers = "X-Forwarded-For: !@{}\"#%^&*()<>?`~\\|'$\t $whatever;   ;; ,,,  1.2.3.4\r\n",
```
martinpitt, line 1226: Would also be good to test the other edge cases: empty string and a lone space.

martinpitt (changes requested):
Nice one, thanks! Nothing that you couldn't talk me into accepting, just a few nitpicks.

jelly:
After merging this we should update our [nginx proxy](https://github.com/cockpit-project/cockpit/wiki/Proxying-Cockpit-over-nginx) article and explain that this header can be set.

src/common/test-webserver.c
allisonkarlitskaya: The issue here is that (particularly in the case of `expected_remote`) there is quite a lot of logic in the fixture setup function to decide what these values will eventually become.  And in the end, we need to have a single pointer to pass through to the verification function.

I could indeed essentially perform all of that logic, statically, and move it to the testcase structs and pass those through instead.  Thoughts?

doc/man/cockpit.conf.xml
allisonkarlitskaya: So the story here is that there's a standardised "Forwarded" header that we all ought to be using since 8 years ago, but nobody supports it and everyone is still using X-Forwarded-For in the wild.  I looked for examples of how to get nginx to add the Forwarded header for example, and I got some really complicated suggestions, whereas nginx has already-built-in support for X-Forwarded-For.  For other servers that I checked I did searches like "squid reverse proxy forwarded header" and landed on pages that were just talking about X-Forwarded-For, seemingly oblivious to the existence of the standard.

tl;dr: "Forwarded" is the official standard, but nobody cares.  X-Forwarded-For is the (absolutely pervasive) de facto standard for this.  If there were to be a second place finisher, it would be X-Real-IP.

doc/man/cockpit.conf.xml
allisonkarlitskaya: OK.

src/common/cockpitwebserver.c
allisonkarlitskaya: I ask myself what a trailing space would mean in that case: I feel like from a practical standpoint that would most likely be the result of a proxy server attempting, and failing, to insert the IP address.

In any case, HTTP is clear about this: if there's trailing whitespace on a header line then it doesn't count as part of the value, and indeed, in our parser code:

```
            g_autofree gchar *value = g_strndup (colon + 1, line - (colon + 1));
            g_strstrip (value);
```

So it's actually impossible for there to be trailing whitespace there.  We can add a test for it, of course.

src/common/test-webserver.c
allisonkarlitskaya: Ya, I'll clean that up.

src/common/test-webserver.c
allisonkarlitskaya: OK.

src/common/test-webserver.c
allisonkarlitskaya: This seems like an awful lot of churn...

allisonkarlitskaya: I did some macro crack to make this feel a lot nicer, and changed all cases (at least in this one file) over to use the new stuff.

doc/man/cockpit.conf.xml
allisonkarlitskaya: Done.  Thanks for the catch.  Funny: "js" isn't even the right thing.  It's "javascript".

src/common/test-webserver.c
allisonkarlitskaya: I tried to do this and it's really not possible without a significant rethink here, or duplication of some code: there's no way to know the remote address in advance, because for some test cases it's the IP address of the machine that the test is running on.

src/common/test-webserver.c
allisonkarlitskaya: This actually shook out some bad behaviour: the reported remote host was `""` which is kinda useless.  I've added a check to ignore the header if the value is empty, so in that case we'll get the fallback to the IP of the proxy itself.

src/common/test-webserver.c
martinpitt: Ack, nevermind then. Thanks for checking!

src/common/test-webserver.c
allisonkarlitskaya: this is in `fixture_setup()`.  It should be read as initializing the value of `expected_protocol` according to the test case.

doc/man/cockpit.conf.xml
martinpitt: Ack, thanks. So let's add something like

> Most web servers use the `X-Forwarded-For` header name for this purpose.

to this documentation?

martinpitt (approved):
Nice, thank you! This looks really nice to me (kudos for the test macro). I'd like to expand documentation a bit, but unless you have some reason to push again for other reasons, this can very well be a no-test follow-up.

martinpitt:
This [arch failure](https://logs.cockpit-project.org/logs/pull-16958-20220224-132808-6264f87d-arch/log.html#299-1) seems suspicious?

allisonkarlitskaya:
> This [arch failure](https://logs.cockpit-project.org/logs/pull-16958-20220224-132808-6264f87d-arch/log.html#299-1) seems suspicious?

At first look it does to me too, but this is looking at the content of the motd file, which is just about the furthest possible piece of code from anything to do with `X-Forwarded-For`...

Let's run it again and see what happens :)

PR #16957, opened 2022-02-09
users: Add override button to confirm weak password

tomasmatus:
Fixes: #16612

Issue #16612, opened 2021-11-16
accounts: Creating password with weak accounts uses a double-tap button, but should use an explicit override button instead


garrett:
When creating or changing the password, there's a weak password override. That's great.

However, it re-uses the same button. This is problematic as _everyone_ can accidentally click a second time and some people even have motor issues which cause them to press keys and click multiple times.

Current screenshot:

![Screenshot 2021-11-16 at 11-44-53 Accounts - garrett Bolt](https://user-images.githubusercontent.com/10246/141971415-76e50910-4e05-4163-b8c0-ac9de5b422b2.png)

garrett:
### New create account password modal design

![create-account](https://user-images.githubusercontent.com/10246/141973961-9b2e4eaf-4bb8-4c4f-a154-dbcf89bccb67.png)

### With weak password error

![create-account-weak](https://user-images.githubusercontent.com/10246/141974017-6e5e641b-5e22-474e-9e08-c73820969bfa.png)

### With weak password error, but with the legacy design

![create-account-weak-legacy(1)](https://user-images.githubusercontent.com/10246/141974279-4afde41d-9b82-43c3-81c6-f4b25279019f.png)

### Source

- [create-account.excalidraw.zip](https://github.com/cockpit-project/cockpit/files/7545736/create-account.zip)
- https://excalidraw.com/#json=46P_Cp_fGDWB2sRk0N59D,OgvInW-3GJ51EGPdK_YoGg

---

### Notes

I've included the legacy design so you can focus on just making a minor change if you'd like. Otherwise, if you feel like overhauling the design in one go, the complete new design is above.

The change password needs similar adjustments. I can create mockups for that later, if you wish.

garrett:
Design needs updating:

- Access should be Options
- Options should have Lock account as in the mockup
- Options should also have "Force password change on first use"
- Perhaps Options should also have "Require password change every [] days"?

garrett:
But the design above works enough for this issue of having a confirmation button.

@jelle is creating additional separate issues for the other proposals.

PR #16959, opened 2022-02-09
Password component cleanups

marusak:
I am trying to add password hints but for that I need some preparation work. Only visible change is dropping of password confirmation, now looks like this:
Screenshot from 2022-02-11 07-55-32

One disadvantage is, that now the value of password input is held by two components separately. That is not ideal, but once component needs it for setting it eventually up and other to run checks on it.

### DIFF
--- a/pkg/lib/cockpit-components-password.jsx
+++ b/pkg/lib/cockpit-components-password.jsx
@@ -47,20 +47,40 @@ export function password_quality(password, force) {
 
 export const PasswordFormFields = ({
     password_label, password_confirm_label,
-    password_strength, password_message,
     password_label_info,
     error_password, error_password_confirm,
     idPrefix, change
 }) => {
     const [password, setPassword] = useState(undefined);
     const [passwordConfirm, setConfirmPassword] = useState(undefined);
+    const [passwordStrength, setPasswordStrength] = useState("");
+    const [passwordMessage, setPasswordMessage] = useState("");
+
+    function onPasswordChanged(value) {
+        setPassword(value);
+        change("password", value);
+
+        if (value) {
+            password_quality(value)
+                    .catch(ex => {
+                        return { value: 0 };
+                    })
+                    .then(strength => {
+                        setPasswordStrength(strength.value);
+                        setPasswordMessage(strength.message);
+                    });
+        } else {
+            setPasswordStrength("");
+            setPasswordMessage("");
+        }
+    }
 
     let variant;
-    if (password_strength === "")
+    if (passwordStrength === "")
         variant = "default";
-    else if (password_strength > 66)
+    else if (passwordStrength > 66)
         variant = "success";
-    else if (password_strength > 33)
+    else if (passwordStrength > 33)
         variant = "warning";
     else
         variant = "danger";
@@ -80,7 +100,7 @@ export const PasswordFormFields = ({
                        validated={error_password ? "error" : "default"}
                        fieldId={idPrefix + "-pw1"}>
                 <TextInput className="check-passwords" type="password" id={idPrefix + "-pw1"}
-                           autocomplete="new-password" value={password} onChange={value => { setPassword(value); change("password", value) }} />
+                           autocomplete="new-password" value={password} onChange={onPasswordChanged} />
                 <div>
                     <Progress id={idPrefix + "-meter"}
                               className={"ct-password-strength-meter " + variant}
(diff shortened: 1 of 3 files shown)


jelly:
Should this be a `release-note`?

jelly (changes requested):

jelly:
So moving the validation inside the component would be nice I think but does it preserve the option to implement the required validation in cockpit-machines using `error_password`?

marusak:
> Should this be labelled with release-note?

Not necessary. It is just small change and hopefully soon there will be more interesting changes around this input :crossed_fingers:

marusak:
> The eye icon does not change to a slashed eye when toggling states. It's always an eye currently (and it should change).

Good catch, thanks! Did not realize at all that we need to update it. Fixed

martinpitt:
Two [TestAccounts](https://logs.cockpit-project.org/logs/pull-16959-20220212-070803-163abf03-debian-stable/log.html#300) [failures](https://logs.cockpit-project.org/logs/pull-16959-20220212-070809-163abf03-fedora-coreos/log.html#300-2), retrying for comparison.

marusak:
It seems that coreos is busted on second retry for `TestAccounts.testBasic`. The pattern is always the same in any PR that forces this test to be retried.

Happend in this PR - [link](https://logs.cockpit-project.org/logs/pull-16959-20220212-070809-163abf03-fedora-coreos/log.html#300-2)
In #16957 (retried 3 times already) [link](https://logs.cockpit-project.org/logs/pull-16957-20220212-093355-c5c8f01e-fedora-coreos/log.html)
In #16981 - [link](https://logs.cockpit-project.org/logs/pull-16981-20220212-092125-57e3bb8a-fedora-coreos/log.html#300-1)

It was fine 4 days ago, see #16940.

It always says:
```
> warning: fatal: Authentication failed: Timeout
```
and then CDP is sad is no page is loaded.

Time wise it seems like https://github.com/cockpit-project/bots/pull/2907 might break this.

pkg/lib/cockpit-components-password.jsx
```diff
@@ -112,14 +125,6 @@ export const PasswordFormFields = ({
                     <div id={idPrefix + "-password-meter-message"} className="pf-c-form__helper-text" aria-live="polite">{passwordMessage}</div>
                 </div>
             </FormGroup>
-
-            {password_confirm_label && <FormGroup label={password_confirm_label}
```
martinpitt, line 94: This commit is stronger than "do not require" -- to me it looks like "completely drop confirmation", which is reflected in the test changes. This seems rather dangerous to me -- typing a hidden password just once? There at least needs to be some big justification in the commit message, but I still don't quite like this. It's of course okay to not require confirmation *if* the pssword gets shown, but I don't see this conditional logic here?

martinpitt (changes requested):
Thanks! The first two commits are really nice centralization (also wrt. reimplementing the same thing in c-machines). The third one seems dubious to me, let's discuss next week.

pkg/lib/cockpit-components-password.jsx
marusak: Indeed, it drops it. It follows up design from https://github.com/cockpit-project/cockpit/issues/16612#issuecomment-970163787

garrett: The idea is that it only shows up when a weak password is made. It's a visual change (with new UI) with a warning color that isn't the default action (which would be disabled in this case).

garrett: The old method is effectively double-clicking the same button, which is easy enough to do, especially for people who have a tremor (but also for _everyone_ accidentally misclicking once in a while), so having to select a secondary button in cases like these would be a better confirmation than clicking the same thing twice.

marusak: @garrett we are talking here about dropping of "Confirm password" and adding he eye icon, not about the yellow button.

garrett: Right, I was explaining how the button works in the context of a confirmation.

garrett: Let's defer the second entry confirmation discussion and get the rest landed. The rest of the changes are well agreed upon; it would be a shame to not land them sooner than later.

martinpitt (approved):
Thanks! As I [mentioned 11 days ago](https://github.com/cockpit-project/cockpit/pull/16959/commits/c389baf8c920e8b756cf00fad95e4f5f213e3ede#r805169538), I'm not happy with the error handling here. But as it's just moved code, we can leave that for another day.

marusak:
> I'm not happy with the error handling here

Right, I just moved the code and don't want to go down rabbit hole in this PR if there is more to it than just `pwscore` is missing.

PR #16961, opened 2022-02-09
Revert "cockpit-client: disable and hide forward/back"

garrett:
This reverts commit 6b716b1b87927ed4428a5514c27ed69242ce2fc4.

This is intended to land after #16938 and undo the temporary removal of back/forward in #16643.

 - [x] #16938

### DIFF
--- a/src/client/cockpit-client
+++ b/src/client/cockpit-client
@@ -70,8 +70,8 @@ class CockpitClientWindow(Gtk.ApplicationWindow):
         self.webview.connect('decide-policy', self.decide_policy)
         self.webview.load_uri(uri)
 
-        # history = self.webview.get_back_forward_list()
-        # history.connect('changed', self.history_changed)
+        history = self.webview.get_back_forward_list()
+        history.connect('changed', self.history_changed)
         self.history_changed()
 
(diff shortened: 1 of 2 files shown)


garrett:
I figured I'd push up a PR for reverting the back/forward button since I was the one to ask for it to be removed originally. :wink:

garrett:
Oh, forward isn't working. I'll relaunch and reconnect just to be sure. (I had an old branch compiled when logging in.)

garrett:
I tried Client again without even recompiling. I must've had an old branch without the back/forward fix or had a cached version or something. Back _and forward_ are both working today for me.

So never mind on that forward issue above, and this should be good to merge.

It would be nice if someone else could confirm though.

jelly:
Back button works fine, I've only found an issue with moving back when selecting "Paths" in the "Services" page but that's unrelated.

martinpitt (approved):
It still feels a little odd -- the first time I have to click "Back" twice for it to have any effect. But still, it generally works and we don't take anything away, so let's land this. Thanks! (Just retriggered the rhel-8-6 test)

PR #16966, opened 2022-02-09
tools: Drop title2sentence script

### LINKED ISSUE #16965: title2sentence is broken
The `title2sentence` script is broken (it throws exceptions) and even when you fix it, it gives some pretty wildly incorrect suggestions.

There's some good stuff in there too, though.

Someone needs to clean up the good stuff.  After that, we have a few options:

 - and then add rules for the wrong stuff (and possibly fix up the matching engine to be a bit more flexible)
 - delete it as being unmaintained and apparently unused
 - turn it into a testcase which runs on every commit and rejects incorrect strings (this way it won't rot again)

marusak:
This script was used for migration we have done. Now it is unused.

@allisonkarlitskaya I think we can just drop it. It would be nice, if we could always check that we don't introduce such new strings, but there is waaay to many special cases and exceptions for it to reasonably work. I think when everyone in the team is aware it is fine.

Fixes #16965


marusak:
Meh, forgot `no-test`

Makefile.am
```diff
@@ -206,9 +206,6 @@ bots:
 multi-word: po/cockpit.pot
```
allisonkarlitskaya, line 206: This one too, I think?

allisonkarlitskaya (dismissed):

Makefile.am
marusak: I am not going to pretend that I left it on purpose - but I found it useful now that I could easily grab all multi world strings and then check them for weirdness. But I can drop it indeed.

marusak: Dropped it and fixed few strings

pkg/storaged/drive-details.jsx
```diff
@@ -60,7 +60,7 @@ export class DriveDetails extends React.Component {
                     <DescriptionListTerm>{_("storage", "Assessment")}</DescriptionListTerm>
                     <DescriptionListDescription>
                         { drive_ata.SmartFailing
-                            ? <span className="cockpit-disk-failing">{_("DISK IS FAILING")}</span>
+                            ? <span className="cockpit-disk-failing">{_("Disk is failing")}</span>
```
marusak, line 63: So I changed this. Seemed odd to have it like that. There is also `Disk is OK` right underneath it - the `OK` seems fine though, or?

I am not sure about this change. What do you think @garrett ?

garrett: We shouldn't have strings in uppercase within Cockpit, unless we're quoting 100% some output and cannot change it for whatever reason (which isn't the case here).

However, "OK" is actually a word that's all uppercase in English. 

https://www.dictionary.com/browse/ok

Deep dive:

It used to be debated if it comes from German ("ohne Korrektur" (meaning: "without correction")) or was originally made as a joke for a short version of "orl korrect" and "oll korrect" ("all correct") in the 1830s in New England (north-east US). The latter is now generally accepted as the original etymology, with a bunch of evidence for it (including old newspapers).

Originally, "oll korrect" was shortened to O.K., but was quickly so popular periods were dropped, as is the case with often-used initialisms (not to be confused with acronyms, which are spoken instead of saying the letters, but also have periods dropped when they become commonly used).

The spelled out form of "okay" was thought to be derived from the sounds of "O K" as kind of a reverse-bacronym — but not really related to an acronym, as then it'd just be "O". (A bacronym being a portmanteu (blend) of "back and acronym", which gives a fake definition to an existing arconym.) Some people felt using a word like OK was wrong and had to have an origin of "okay", but was spelled out properly... but they were wrong and invented that backstory. But "okay" became OK to use somewhat informally, even if not as correct as OK.

Oh, I'm not kidding about this either: https://www.history.com/this-day-in-history/ok-enters-national-vernacular

So "OK" is OK, "okay" is only kind of OK, and "ok" is never OK, except in extremely informal digital messaging (IMs, DMs, SMSes, IRC, etc.) where case, grammar, punctuation, and fragments are all kind of acceptable.

It's infectious too... most languages have adopted both "OK" and "Okay" as they're so short to say. It's normal in German, for example.

(I needed to look this up years ago for a similar question, as OK shows up in UI quite a bit, especially with OK/Cancel dialogs, which aren't great. There are almost always better contextual words to use than that pairing, such as "Save", "Apply", "Create", "Commit", etc. "Cancel" is still acceptable, though.)

As for other words in its place: "Normal", "fine", or even the phrase "no errors detected" can all work. There's also "nominal", which actually means "related to a name" or "insignificantly small" in typical usage... but is commonly used in aeronautical lingo (such as in cockpits, helms, and bridges in air, sea, and space) to mean "we can't detect errors; everything appears fine": https://english.stackexchange.com/questions/184876/how-did-nominal-come-to-mean-within-acceptable-tolerances/189686#189686 — however, this is probably too jargony for us to use, even though it the word someone would actually use in a "real" cockpit when describing system status.

---

**Summary**: OK is OK.

marusak: Thank you @garrett for the explanation!
I've rebased this PR (as there were some unrelated test failures that got fixed in main in the meantime) so this is ready.

martinpitt (approved):
Cheers!

PR #16968, opened 2022-02-10
fmf: Ignore runtime dir error message on centos-8-stream

martinpitt:
This breaks quite a few of our c8s Testing Farm runs, makes little
sense, and is pretty well impossible to debug. As long as it does not
affect c9s or Fedora, ignore the message, at least until we have time to
spend a day or two on tracking this down.

----

I've [seen this a lot](http://artifacts.dev.testing-farm.io/66ec30f1-38df-49ef-9275-0ac34828e780/) recently. I already spent some time to understand it, but no idea so far. I realize that this is just throwing hands into the air, but "pick your battles".. :man_shrugging:


marusak (approved):
Thanks!

PR #16973, opened 2022-02-10
test: Fix race condition with checking out pixel tests

martinpitt:
With parallel tests it could happen that two pixel tests were running
around the same time. One would start the git clone, the other would see
the existing .git; but if the cloning was not done yet, the actual files
were missing, and the test failed with "New pixel test reference".

Let's avoid that by going back to pulling the reference images from
test/run; this reverts commit c20822b980b. In exchange, replace the 
on-demand checkout in testlib.py with an instructive error message.

----

[example 1](https://logs.cockpit-project.org/logs/pull-16942-20220209-124540-22d91db9-fedora-35/log.html#89), [example 2](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16942-20220210-104937-a2b3c82a-fedora-35/log.html#267)

Now that we run 8 `@nd` tests in parallel instead of 2, this became a lot more apparent.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -885,7 +885,7 @@ class Browser:
 
         reference_dir = os.path.join(TEST_DIR, 'reference')
         if not os.path.exists(os.path.join(reference_dir, '.git')):
-            subprocess.check_call([f'{TEST_DIR}/common/pixel-tests', 'pull'])
+            raise SystemError("Pixel test references are missing, please run: test/common/pixel-tests pull")
 
         ignore_rects = relative_clips(list(map(lambda item: selector + " " + item, ignore)))
         base = self.pixels_label + "-" + key
--- a/test/run
+++ b/test/run
@@ -6,6 +6,7 @@
 set -eu
 
 tools/make-bots
+test/common/pixel-tests pull
 
 TEST_SCENARIO=${TEST_SCENARIO:-verify}
 TEST_OS_DEFAULT=$(PYTHONPATH=bots python3 -c 'from machine.machine_core.constants import TEST_OS_DEFAULT; print(TEST_OS_DEFAULT)')


allisonkarlitskaya (commented):
I feel like this is needlessly complex.  I'd strongly favour a solution that failed if the reference images weren't already there and an explicit step to check them out before starting the tests.

mvollmer:
> I feel like this is needlessly complex. I'd strongly favour a solution that failed if the reference images weren't already there and an explicit step to check them out before starting the tests.

I agree.

mvollmer:
> We could also just split the difference: leave the code here alone (for the convenience of people running individual tests), but check out the pixel tests from `run-tests` or the toplevel `test/run` script, in cases where it's appropriate.

Yes, let's do that.  c-machines etc run "test/common/pixel-tests pull" from their Makefile (via a dependency of "make check"), we could do the same in `test/run`.

martinpitt:
Now a local run fails like this:
```
❱❱❱ test/verify/check-networkmanager-team TestTeam.testBasic $RUNC

# testBasic (__main__.TestTeam)
[...]
  File "/var/home/martin/upstream/c2/test/verify/check-networkmanager-team", line 58, in testBasic
    b.assert_pixels("#network-team-settings-dialog", "networking-team-settings-dialog")
  File "/var/home/martin/upstream/c2/test/common/testlib.py", line 989, in assert_pixels
  File "/var/home/martin/upstream/c2/test/common/testlib.py", line 884, in assert_pixels_in_current_layout
    raise SystemError("Pixel test references are missing, please run: test/common/pixel-tests pull")
SystemError: Pixel test references are missing, please run: test/common/pixel-tests pull

# Result testBasic (__main__.TestTeam) failed
# 1 TEST FAILED [27s on toolbox]
```

allisonkarlitskaya (approved):
Looks good to me!

allisonkarlitskaya:
One note (which is probably completely irrelevant).

The previous code would happily use an "old" copy of the reference images if that's what happened to be checked out.  This version will too (since the condition is unchanged).

One thing that is changed, however, is that we now always pull the images from `test/run`, so we'll automatically refresh.  That might impact the workflow of people who are trying to work on pixel tests and have outstanding/uncommitted changes in the reference directory.

That said, I can't imagine that someone who is in the middle of outstanding changes to reference images is going to be opting for `test/run`, rather than running a specific testcase directly.

martinpitt:
@allisonkarlitskaya : Right, test/run isn't really a human thing at all. But I'd hope that `git submodule update` would fail if the tree has modified files?

Issue #16974, opened 2022-02-10
check-metrics often fails on debian-stable


martinpitt:
### Explain what happens

`pmlogger.service` often gets stuck, so that stopping/restarting the service does not help. It stays in state "activating", times out, and eventually fails because it restarted too often.

[example 1](https://logs.cockpit-project.org/logs/pull-16942-20220210-091454-a2b3c82a-debian-stable/log.html#59), [example 2](https://logs.cockpit-project.org/logs/pull-16942-20220208-132850-836165a4-debian-stable/log.html#61-2)

I tried to robustify this with commit de74d2e0c91, but that didn't help, or was not sufficient.

### Version of Cockpit

main

martinpitt:
The root cause is that the pmlogger permissions get totally whacked up:
```
# ls -ld /var/log/pcp/pmlogger
drwxrwxr-x 3 1000 wheel 4096 Sep 19 22:13 /var/log/pcp/pmlogger
```
Once I fix this (`chown -R pcp:pcp /var/log/pmlogger`), pmlogger.service happily starts again.

This is triggering a very strong déjà-vu, I saw this bug already a few months ago. After some digging I found it, it's https://github.com/cockpit-project/bots/issues/2518 aka https://bugzilla.redhat.com/show_bug.cgi?id=2013937, and debian-stable is still affected. So let's add a workaround on debian-stable only, as this bug manifests in lots of different ways.

Issue #16970, opened 2022-02-10
fedora-coreos test flakes with authentication timeouts


martinpitt:
It's still suspicious that [so many tests are flaking](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16962-20220210-080844-c18704c6-fedora-coreos/log.html) on the initial login ([another example](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16968-20220210-084828-58f992d1-fedora-coreos/log.html)) . That does not seem to happen in yesterday's PRs, fcos is fairly clean there.

I can even reproduce that locally, and it even seems to happen *at the very first test* right after booting, i.e. it can't just be related to the "terminate session" cleanup fix. In fact none of the changes here affect the container startup.

```
# ----------------------------------------------------------------------
# testBasic (__main__.TestSession)
Starting machine1 fedora-coreos-127.0.0.2-2201
Assuming machine is already running
+ ! test -f /run/nologin && cat /proc/sys/kernel/random/boot_id
1abef374-30b2-4d44-b780-6f2d907746c1
+ journalctl --show-cursor -n0 -o cat | sed 's/^.*cursor: *//'
s=f35fbca747ae41e6ae4efdc663ac30af;i=1629;b=1abef37430b24d44b7806f2d907746c1;m=5d214ae;t=5d7a612dc52ce;x=41d054f8431962e2
-> Emulation.setDeviceMetricsOverride({"width": 1920, "height": 1200, "deviceScaleFactor": 0, "mobile": false})
Started /usr/lib64/chromium-browser/headless_shell (pid 35232) on port 9307

DevTools listening on ws://127.0.0.1:9307/devtools/browser/2e5efa0b-c197-4d33-a115-a712144ca7e3
[0210/095244.570959:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is swiftshader, ANGLE is 
CDP: executionContextCreated {"context":{"id":1,"origin":"://","name":"","uniqueId":"-4823293432612784198.-5009512227711019971","auxData":{"isDefault":true,"type":"default","frameId":"B7D4FA269922BAFB9A47BB4A8C144B97"}}}
<- {}
+ cat > '/etc/cockpit/cockpit.conf'
+ logger -p user.info 'COCKPITTEST: start TestSession.testBasic'
+ if test -e /etc/passwd; then echo yes; fi
yes
[...]

PR #16975, opened 2022-02-10
Test flake fixes

martinpitt:
This addresses more of the discovered flakes from PR #16942. The big remaining one is #16970, but that's a "tomorrow" thing.

[example for the multi-machine failure](https://logs.cockpit-project.org/logs/pull-16973-20220210-151712-c7a966fe-fedora-35-firefox/log.html#142)

Issue #16974 has the details for the second commit.

### DIFF
--- a/test/verify/check-shell-multi-machine
+++ b/test/verify/check-shell-multi-machine
@@ -818,6 +818,7 @@ class TestMultiMachine(MachineCase):
 
         # Relogin.  This should now work seamlessly.
         b.relogin(None, wait_remote_session_machine=m1)
+        self.allow_restart_journal_messages()
         b.enter_page("/system", host="fred@10.111.113.2")
 
         # De-authorize key and relogin, then re-authorize.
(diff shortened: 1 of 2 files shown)


test/verify/check-shell-multi-machine
```diff
@@ -818,6 +818,7 @@ class TestMultiMachine(MachineCase):
 
         # Relogin.  This should now work seamlessly.
         b.relogin(None, wait_remote_session_machine=m1)
+        self.allow_restart_journal_messages()
```
marusak, line 821: Just an idea - why don't we add this into definition of relogin/logout/reload...?

marusak (approved):
Thanks!

test/verify/check-shell-multi-machine
marusak: Since c3309b7a56 `Browser` has reference to `machine`. I might take a look at this.

martinpitt: I stand corrected -- last time I checked this wasn't the case, and it felt a little dirty (layering violation). But so much the better, then we can fix this wholesale :partying_face: thanks!

PR #16976, opened 2022-02-10
tests: Unify usage of `allow_restart_journal_messages()`

marusak:
When it is used due to `relogin()`, `reload()` or `logout()` don't
always specify it but let these methods to do it.

There are still 11 cases where `allow_restart_journal_messages()` is used - these are due to different reasons than us logging out or relaoding (like killing cockpit, restarting the machine, removing host...)

Lets see if there are some flaky messages

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -217,6 +217,8 @@ class Browser:
         self.cdp.invoke("Page.reload", ignoreCache=ignore_cache)
         self.expect_load()
 
+        self.machine.allow_restart_journal_messages()
+
     def expect_load(self):
         if opts.trace:
             print("-> expect_load")
@@ -690,6 +692,8 @@ class Browser:
                 self.open_session_menu()
                 self.click('#logout')
         self.expect_load()
+
+        self.machine.allow_restart_journal_messages()
 
     def relogin(self, path=None, user=None, superuser=None, wait_remote_session_machine=None):
         self.logout()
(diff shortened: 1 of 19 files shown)


martinpitt:
Nice! This fixes these pesky issues in a more general way.

FYI, I just landed #16975 (as fixing these flakes is urgent), which introduces another one of these.

marusak:
> FYI, I just landed https://github.com/cockpit-project/cockpit/pull/16975 (as fixing these flakes is urgent), which introduces another one of these.

Fixes and pushed

martinpitt:
Flake review:

 - [pcp failure on debian-stable](https://logs.cockpit-project.org/logs/pull-16976-20220210-194929-5ee45c85-debian-stable/log.html#37-2): That's the bit which I just fixed in commit b9f03a07d47. You rebased after that, so apparently there's still some case where it happens. Unrelated.
 - [debian-testing storage failures](https://logs.cockpit-project.org/logs/pull-16976-20220210-194929-5ee45c85-debian-testing/log.html): unrelated
 - [session page timeout on rhel-8-6](https://logs.cockpit-project.org/logs/pull-16976-20220210-194929-5ee45c85-rhel-8-6/log.html#133-2): I tried to address that in commit 45008f6096246, but apparently 5 mins is *still* not enough? That starts to smell like some actual page bug (we really REALLY need to make this page load less).

Either way, all unrelated to this change, retrying once.

martinpitt (approved):

martinpitt:
I sent PR #16979 for the first flake mentioned above.

PR #16977, opened 2022-02-10
Added a link under "Unit not found" to view all services

### LINKED ISSUE #16946: services: Unit not found pages have no action
Generally, you probably won't hit a unit not found page. But in bugs like #16945, you may. And there may be some edge cases where you could hit it, especially for user services which might be spawned just for one session and not for another.

Right now, the page looks like this:

image

We should use an [empty state pattern](https://www.patternfly.org/v4/components/empty-state/) with an action. The action would just be a link back to the main page. (It should be styled like a link, not a button, as it goes somewhere instead of doing something.)

---

**TL;DR**: Add a link under "Unit not found" that says "View all services" which goes back to the services list.

Zach-Clay:
Screenshot from 2022-02-21 09-53-02

Fixes #16946

### DIFF
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -729,9 +729,17 @@ class ServicesPageBody extends React.Component {
             const unit_path = get_unit_path(unit_id);
 
             if (unit_path === undefined) {
+                const path = "/system/services" + (this.props.owner === "user" ? "#/?owner=user" : "");
                 return <EmptyStatePanel
                             icon={ExclamationCircleIcon}
-                            title={_("Unit not found")} />;
+                            title={_("Unit not found")}
+                            paragraph={
+                                <Button variant="link"
+                                        component="a"
+                                        onClick={() => cockpit.jump(path, cockpit.transport.host)}>
+                                    {_("View all services")}
+                                </Button>
+                            } />;
             }
 
             const unit = this.state.unit_by_path[unit_path];
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -523,6 +523,15 @@ ExecStart=/bin/awk -f /tmp/mem-hog{params}.awk
         # Check that listen is not shown for .service units
         b.wait_not_present("#listen")
 
+        # Check 'Unit not found' pattern
+        b.go(f'/system/services#/nonexistent{suffix}')
+        b.wait_in_text(".pf-c-empty-state", "Unit not found")
+        b.click("a:contains('View all services')")
+        b.switch_to_top()
+        b.wait_js_cond('window.location.pathname == "/system/services"')
+        if user:
+            b.wait_js_cond(f"window.location.hash === '#/{suffix}'")
+
     def testLogs(self):
         b = self.browser
 


pkg/systemd/services/services.jsx
```diff
@@ -729,9 +729,17 @@ class ServicesPageBody extends React.Component {
             const unit_path = get_unit_path(unit_id);
 
             if (unit_path === undefined) {
-                return <EmptyStatePanel
+                return (<EmptyStatePanel
```
marusak, line 732: Are these `()` necessary here?

pkg/systemd/services/services.jsx
```diff
@@ -729,9 +729,17 @@ class ServicesPageBody extends React.Component {
...
                             icon={ExclamationCircleIcon}
-                            title={_("Unit not found")} />;
+                            title={_("Unit not found")}
+                            secondary={
```
marusak, line 735: Using `secondary` seems like the most logical solution. However it feels like there is too much space now.
`secondary`:
Screenshot from 2022-02-11 07-29-24

`paragraph`:
paragraph

WDYT?

marusak (dismissed):
Awesome job, thank you!
It would be great if you could add test for it. You can read about how to run tests locally [here](https://github.com/cockpit-project/cockpit/blob/main/test/README.md). You most likely want to add test into `test/verify/check-system-services`. Just going to nonexisting service, clicking the button and making sure it leads to services page would be enough. Thanks!

While on re-pushing, please also `--amend` commit message to use [imperative mood](https://cbea.ms/git-commit/#imperative), add `services:` prefix and add [Fixes #...](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) line into commit message.

Zach-Clay:
@marusak Hi! I have removed the parentheses that were not needed also also made the link a `paragraph` instead of `secondary`. I think this removes some of the unwanted space. 

In terms of adding a test, I took a look at the file and I am not really sure how to go about doing this...

marusak:
> While trying to run the test I am getting multiple errors on top of my VM crashing

If you want to, please post your issues here or join us on IRC (#cockpit on libera) and we can help you

> how important is it that I write a test in order for my PR to be merged?

It won't land without tests. But if you prefer, I am happy to write tests and push them here.

pkg/systemd/services/services.jsx
```diff
@@ -731,7 +731,14 @@ class ServicesPageBody extends React.Component {
             if (unit_path === undefined) {
                 return <EmptyStatePanel
                             icon={ExclamationCircleIcon}
-                            title={_("Unit not found")} />;
+                            title={_("Unit not found")}
+                            paragraph={
+                                <Button variant="link"
+                                        component="a"
+                                        onClick={() => cockpit.jump("/system/services", cockpit.transport.host)}>
```
marusak, line 738: I just realized, that when you would see this for user unit, you should go to all user units and not to system ones

Zach-Clay:
@marusak 
> It won't land without tests. But if you prefer, I am happy to write tests and push them here.

That would be awesome given my current workload!

marusak:
Rebased, cleaned up commits, added tests and fixed redirection link to work correctly for system and user. Running tests now.

jelly (approved):

PR #16982, opened 2022-02-11
webpack: replace our gzip tricks with compression-webpack-plugin

allisonkarlitskaya:
Our build system does quite some tricks to compress our assets on the
fly during the `make install` phase.  Let's just do the compression as
part of the webpack.

 - [x] wait for #17105 
 - [x] #17120

### DIFF
--- a/pkg/Makefile.am
+++ b/pkg/Makefile.am
@@ -25,22 +25,19 @@ EXTRA_DIST += \
 	$(pixmaps_DATA) \
 	$(NULL)
 
+pcpmanifestdir = $(datadir)/cockpit/pcp
+dist_pcpmanifest_DATA = pkg/pcp/manifest.json
+
+sshmanifestdir = $(datadir)/cockpit/ssh
+dist_sshmanifest_DATA = pkg/ssh/manifest.json
+
 include pkg/build
 
-INSTALL_DATA_LOCAL_TARGETS += install-webpack
-install-webpack: $(WEBPACK_INSTALL) $(MANIFESTS)
-	mkdir -p $(DESTDIR)$(pkgdatadir)
-	$(V_TAR) tar --format=posix -cf - $^ | tar --no-same-owner -C $(DESTDIR)$(pkgdatadir) --strip-components=1 -xvf -
+# automake doesn't support stripping initial path components (ie: `dist`) with
+# `nobase_`, so we create a cockpit/ symlink and use that instead.
+# That way, the files land in $(datadir)/cockpit/ and `dist` is avoided.
+all-local:: cockpit
+cockpit:
+	$(AM_V_GEN) ln -sf $(srcdir)/dist $@
 
-INSTALL_DATA_LOCAL_TARGETS += install-webpack-map
-install-webpack-map: $(wildcard $(top_srcdir)/dist/*/*.map)
-	$(V_TAR) tar --format=posix -c --transform="s@.*dist/@$(debugdir)$(pkgdatadir)/@" -T/dev/null $^ | tar --no-same-owner -C $(DESTDIR)/ -xv
-
-UNINSTALL_LOCAL_TARGETS += uninstall-webpack
-uninstall-webpack:
-	test ! -d $(DESTDIR)$(pkgdatadir) || \
-	  (find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type f -exec rm -f {} \; && \
-	   find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete)
-	test ! -d $(DESTDIR)$(debugdir)$(pkgdatadir) || \
-	  (find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type f -delete && \
-	   find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete)
+nobase_data_DATA = $(patsubst dist/%,cockpit/%,$(WEBPACK_INSTALL))
--- a/pkg/Makefile.am
+++ b/pkg/Makefile.am
@@ -32,16 +32,6 @@ install-webpack: $(WEBPACK_INSTALL) $(MANIFESTS)
 	mkdir -p $(DESTDIR)$(pkgdatadir)
 	$(V_TAR) tar --format=posix -cf - $^ | tar --no-same-owner -C $(DESTDIR)$(pkgdatadir) --strip-components=1 -xvf -
 
-INSTALL_DATA_LOCAL_TARGETS += install-webpack-gz
-install-webpack-gz: $(WEBPACK_GZ_INSTALL)
-	@for file in $^; do \
-		target="$(DESTDIR)$(pkgdatadir)/$${file##*dist/}.gz"; \
-		[ -n "$(DESTDIR)" -a "$${target}" -nt "$${file}" ] && continue; \
-		mkdir -p "$${target%/*}"; \
-		echo "$${file} → $${target}"; \
-		gzip -9 < "$${file}" > "$${target}"; \
-	done
-
 INSTALL_DATA_LOCAL_TARGETS += install-webpack-map
 install-webpack-map: $(wildcard $(top_srcdir)/dist/*/*.map)
 	$(V_TAR) tar --format=posix -c --transform="s@.*dist/@$(debugdir)$(pkgdatadir)/@" -T/dev/null $^ | tar --no-same-owner -C $(DESTDIR)/ -xv
--- a/pkg/build
+++ b/pkg/build
@@ -23,11 +23,7 @@ WEBPACK_PACKAGES = \
 	users \
 	$(NULL)
 
-MANIFESTS = \
-	$(WEBPACK_PACKAGES:%=$(srcdir)/dist/%/manifest.json) \
-	pkg/pcp/manifest.json \
-	pkg/ssh/manifest.json \
-	$(NULL)
+MANIFESTS = $(WEBPACK_PACKAGES:%=$(srcdir)/dist/%/manifest.json)
 
 .PHONY: all-webpack
 all-webpack: $(MANIFESTS)
--- a/src/common/cockpitwebresponse.c
+++ b/src/common/cockpitwebresponse.c
@@ -1264,13 +1264,6 @@ web_response_file (CockpitWebResponse *response,
                    CockpitTemplateFunc template_func,
                    gpointer user_data)
 {
-  const gchar *default_policy = "default-src 'self' 'unsafe-inline';";
-  const gchar *headers[5] = { NULL };
-  g_autofree gchar *alloc = NULL;
-  GList *l = NULL;
-  gint content_length = -1;
-  gint at = 0;
-
   g_return_if_fail (COCKPIT_IS_WEB_RESPONSE (response));
 
   if (!escaped)
@@ -1334,23 +1327,24 @@ web_response_file (CockpitWebResponse *response,
     }
 
   g_autoptr(GBytes) body = g_mapped_file_get_bytes (file);
-  GList *output = NULL;
+
+  GList *output;
+  gint content_length = -1;
   if (template_func)
     {
       output = cockpit_template_expand (body, "${", "}", template_func, user_data);
     }
   else
     {
-      output = g_list_prepend (output, g_bytes_ref (body));
+      output = g_list_prepend (NULL, g_bytes_ref (body));
       content_length = g_bytes_get_size (body);
     }
-  g_bytes_unref (body);
+
+  GString *string = begin_headers (response, 200, "OK");
+  guint seen = 0;
 
   if (response->origin)
-    {
-      headers[at++] = "Access-Control-Allow-Origin";
-      headers[at++] = response->origin;
-    }
+    seen |= append_header (string, "Access-Control-Allow-Origin", response->origin);
 
   /*
    * The default Content-Security-Policy for .html files allows
@@ -1359,13 +1353,15 @@ web_response_file (CockpitWebResponse *response,
    */
   if (g_str_has_suffix (unescaped, ".html"))
     {
-      headers[at++] = "Content-Security-Policy";
-      headers[at++] = alloc = cockpit_web_response_security_policy (default_policy, response->origin);
+      const gchar *default_policy = "default-src 'self' 'unsafe-inline';";
+      g_autofree gchar *policy = cockpit_web_response_security_policy (default_policy, response->origin);
+      seen |= append_header (string, "Content-Security-Policy", policy);
     }
 
-  cockpit_web_response_headers (response, 200, "OK", content_length,
-                                headers[0], headers[1], headers[2], headers[3], NULL);
+  g_autoptr(GBytes) headers_block = finish_headers (response, string, content_length, 200, seen);
+  queue_bytes (response, headers_block);
 
+  GList *l;
   for (l = output; l != NULL; l = g_list_next (l))
     {
       if (!cockpit_web_response_queue (response, l->data))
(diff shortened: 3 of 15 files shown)


allisonkarlitskaya:
The compression seems to interfere with the creation of `.LICENSE.txt` files...

allisonkarlitskaya:
Also: the unit tests are broken because the test server doesn't know how to serve `cockpit.js.gz`.

allisonkarlitskaya:
@martinpitt Maybe you could help me figure out this LICENSE.txt business.  I have no idea how the combination of Terser and the compression plugin is supposed to work.  In starter-kit, for example, I was quite surprised to see that a `LICENSE.txt.gz` is generated, despite the compression plugin apparently only targetting js/html/css files...

allisonkarlitskaya:
or @KKoukiou :)

KKoukiou:
> or @KKoukiou :)

This is weird, it seems to be applying the test on the base (index.js) and not the actual file.
With this change the bahavior changes to the expected:
```
[kkoukiou@sequioa starter-kit]$ git diff
diff --git webpack.config.js webpack.config.js
index d7eebca9..b71c3300 100644
--- webpack.config.js
+++ webpack.config.js
@@ -37,7 +37,11 @@ const plugins = [
 if (production) {
     plugins.unshift(new CompressionPlugin({
         test: /\.(js|html|css)$/,
-        deleteOriginalAssets: true
+        deleteOriginalAssets: true,
+        filename(input) {
+            // return "[path][base].gz"; <--- This is the code that runs by default
+            return input.filename + ".gz";
+        }
     }));
 }
 
[kkoukiou@sequioa starter-kit]$ ls dist/
index.css.gz  index.html.gz  index.js.gz  index.js.LICENSE.txt  manifest.json  po.de.js
````

To understand this I read https://webpack.js.org/plugins/compression-webpack-plugin/#filename

**Comment by martinpitt** [comment]
@allisonkarlitskaya : I'm still trying to make sense of this. Aside from the two flakes, the main issue here is the [/devel scenario](https://logs.cockpit-project.org/logs/pull-16982-20220211-161712-b549123a-fedora-35-devel/log.html#1-2) which seems totally broken. Loading `*.gz` files apparently works fine, but loading uncompressed files doesn't.

But what does that have to do with the auto-built LICENSE.txt file? We don't even use terser's license building cockpit -- I've tried that in the past, but it just doesn't work (can't find the reference now, sorry).

This crashes here:
```
Uncaught ReferenceError: cockpit is not defined
    <anonymous> https://127.0.0.2:9091/cockpit/$e9ad480aa861e7c50e340f26ee46521d1c3e48d0a768f593f8ffb5d374562d8e/*/po.js:1
```

I.e. the commits to cockpitwebserver are somehow not happy about the globbing. The generated webpacks *do* work with main's cockpit-ws (i.e. if I just point ~/.local/share/cockpit/ to this branch), but fail with cockpit-ws from image-prepare.

So perhaps we need #16930 first? But no, I added that commit, and get the same crash, just without the glob:
```
Uncaught ReferenceError: cockpit is not defined
    <anonymous> https://127.0.0.2:9091/cockpit/$ee48016ff63817bccecdbc6ac41ddbed216261dd174319e400f14af591276a66/shell/po.js:1
```

I then tested your first 5 commits in isolation (copying cockpit-{bridge,ws} into the VM), but they don't reproduce the failure.

So, I currently don't understand what's going wrong, but I am curious what led you to believe that this LICENSE file is somehow the cause of this? Just so that I avoid re-doing that odyssey again.

**Comment by martinpitt** [comment]
FTR, fastest way to iterate:

    rm -rf dist/sosreport; NODE_ENV=production tools/webpack-make -e sosreport && diff -u <(ls ../cockpit/dist/sosreport/) <(ls dist/sosreport)

This is identical to current main when completely disabling the CompressionPlugin. I think it's the combination of `test` (which does not cover LICENSE.txt) and `deleteOriginalAssets` which causes this. The latter is *really* evil -- it seems to delete anything which is not explicitly compressed.

I.e. I tried this, instead of the hack commit:

```diff
--- webpack.config.js
+++ webpack.config.js
@@ -353,13 +353,14 @@ const plugins = [
 
 if (production) {
     plugins.push(new CompressionPlugin({
-        test: /\.(css|html|js)$/,
+        test: /\.(css|html|js|LICENSE.txt)$/,
         deleteOriginalAssets: true,
         exclude: [
             '/test-[^/]+.$', // don't compress test cases
             '^static/[^/]+$', // cockpit-ws cannot currently serve compressed login page
             '^shell/index.html$', // COMPAT: Support older cockpit-ws binaries. See #14673
-        ].map((r) => new RegExp(r)),
+            '\.LICENSE\.txt$', // build-debian-copyright does not read compressed files
+        ].map(r => new RegExp(r)),
     }));
 }
```
but it *still* deletes the license file.

The only thing that actually works is to give in and let it compress the license file as well, and teach the copyright builder to read gzip'ed files:

```diff
--- tools/build-debian-copyright
+++ tools/build-debian-copyright
@@ -4,6 +4,7 @@
 #         Allison Karlitskaya <allison.karlitskaya@redhat.com>
 
 import argparse
+import gzip
 import os
 import re
 import sys
@@ -80,10 +81,10 @@ dist_licenses = {}  # Files: dirglob → set(licenses)
 
 for directory, _subdirs, files in os.walk('dist'):
     for file in files:
-        if not file.endswith('.LICENSE.txt'):
+        if not file.endswith('.LICENSE.txt.gz'):
             continue
 
-        with open(os.path.join(directory, file)) as license_file:
+        with gzip.open(os.path.join(directory, file), 'rt') as license_file:
             comments = license_file.read().split('\n\n')
 
         for comment in comments:
diff --git webpack.config.js webpack.config.js
index 0d796e454..d78c45ee4 100644
--- webpack.config.js
+++ webpack.config.js
@@ -353,13 +353,13 @@ const plugins = [
 
 if (production) {
     plugins.push(new CompressionPlugin({
-        test: /\.(css|html|js)$/,
+        test: /\.(css|html|js|LICENSE.txt)$/,
         deleteOriginalAssets: true,
         exclude: [
             '/test-[^/]+.$', // don't compress test cases
             '^static/[^/]+$', // cockpit-ws cannot currently serve compressed login page
             '^shell/index.html$', // COMPAT: Support older cockpit-ws binaries. See #14673
-        ].map((r) => new RegExp(r)),
+        ].map(r => new RegExp(r)),
     }));
 }

```

I'm not pushing this, as it's awkward due to the node_modules rebuild.

My recommendation is to PR the first five commits separately, to test them in isolation, and add a unit test for serving *.gz files. Hopefully with your recent test-webserver fixes this should be a lot more pleasant? Then have a separate commit to add the compression webpack plugin, so that we get this out of the way.

martinpitt:
#17105 landed, so this is conflicty and  needs a rebase. But unblocked.

src/common/cockpitwebresponse.c
```diff
@@ -1299,6 +1302,18 @@ web_response_file (CockpitWebResponse *response,
 
       g_autoptr(GError) error = NULL;
       file = g_mapped_file_new (path, FALSE, &error);
+
+      if (file == NULL && search_gzip &&
+          g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        {
+          g_debug ("%s: file not found in root: %s", escaped, root);
```
martinpitt, line 1309: Maybe "..., looking for *.gz"?

martinpitt (commented):
Changes LGTM; not approving yet due to conflict and one small nitpick

allisonkarlitskaya:
hm.  Dealing with the static manifests (`ssh`, `pcp`) is going to be annoying :(

src/common/cockpitwebresponse.c
allisonkarlitskaya: Done, thanks.

allisonkarlitskaya:
The `devel` fail is pretty easy to understand: for development builds, we don't engage the compression plugin, but the old machinery (from combined `pkg/Makefile.am` and `webpack-make`) compressed the files on install, unconditionally.

The packaging adds a compatibility symlink (and I think the check there is broken: it seems to intend only to install it on RHEL, but it installs it also on Fedora, clearly), and in making this symlink it assumes that the file will be called `cockpit.css.gz`.

Then: when the cockpitpackages code is iterating the directory to calculate the checksums and build the table of files, its inability to open this broken symlink results in it aborting the iteration, and never finding `cockpit.js`.

Then, well, 404.


There are a lot of ways to fix this.  First, I think we should have a discussion about if a broken symlink in a directory is enough to cause the entire package loading process to abort.  Then probably we should remove this symlink from Fedora and/or conditionalise its install for only "production" builds.

allisonkarlitskaya:
Need to wait for #17120.

martinpitt:
[This error](https://logs.cockpit-project.org/logs/pull-16982-20220311-092505-8f4464a1-fedora-35-devel/log.html#141-2) is not a flake -- terminal.html.gz does not exist, as it's not compressed at all any more. This needs fixing.

KKoukiou:
> [This error](https://logs.cockpit-project.org/logs/pull-16982-20220311-092505-8f4464a1-fedora-35-devel/log.html#141-2) is not a flake -- terminal.html.gz does not exist, as it's not compressed at all any more. This needs fixing.

Could be the same reason as https://github.com/rhinstaller/anaconda/pull/3949 ..

src/ws/test-server.c
```diff
@@ -632,7 +632,7 @@ on_handle_source (CockpitWebServer *server,
       inject_address (response, "bus_address", bus_address);
       inject_address (response, "direct_address", direct_address);
     }
-  cockpit_web_response_file (response, path, (const gchar **)server_roots);
+  cockpit_web_response_file_or_gz (response, TRUE, path, (const gchar **)server_roots);
```
martinpitt, line 635: Could there be two unit test that make sure that serving gzip files actually works? (first one for accepting/receiving gz, the second for transparent decompression)

pkg/lib/cockpit-po-plugin.js
```diff
@@ -68,8 +68,6 @@ module.exports = class {
         const patterns = [
             // all translations for that page, including manifest.json and *.html
             `pkg/${this.subdir}.*`,
-            // FIXME: https://github.com/cockpit-project/cockpit/issues/13906
-            'pkg/base1/cockpit.js'
```
martinpitt, line 72: Hmm.. now I actually wonder whether we should keep this hack, and not include the .po separately. That would avoid another browser ←→ ws roundtrip. The timing tests have shown that it makes no measurable difference how big the file is (4 kB or 150 KB took almost the same time)

See https://github.com/cockpit-project/cockpit/pull/16930#discussion_r824575488

martinpitt:
@KKoukiou I'm afraid not -- it's 100% totally clear what happens here -- because we *told it to*. :grin:

pkg/build
```diff
@@ -12,22 +12,20 @@ WEBPACK_PACKAGES = \
...
 	systemd \
 	tuned \
 	users \
 	$(NULL)
 
-MANIFESTS = \
-	$(WEBPACK_PACKAGES:%=$(srcdir)/dist/%/manifest.json) \
-	pkg/pcp/manifest.json \
-	pkg/ssh/manifest.json \
-	$(NULL)
```
martinpitt, line 30: TBH like it better the current way -- pcp and ssh really aren't webpacks. They have manifests, yes, but that's not a webpack concept. It adds unnecessary build overhead, too.

pkg/Makefile.am
```diff
@@ -27,20 +27,11 @@ EXTRA_DIST += \
...
-	   find $(DESTDIR)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete)
-	test ! -d $(DESTDIR)$(debugdir)$(pkgdatadir) || \
-	  (find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type f -delete && \
-	   find $(DESTDIR)$(debugdir)$(pkgdatadir) -ignore_readdir_race -type d -empty -delete)
+# automake doesn't support stripping initial path components (ie: `dist`) with
+# `nobase_`, so we create a cockpit/ symlink and use that instead.
+# That way, the files land in $(datadir)/cockpit/ and `dist` is avoided.
+all-local:: cockpit
+cockpit:
+	$(AM_V_GEN) ln -sf $(srcdir)/dist $@
```
martinpitt, line 35: No veto, but it would seem easier and cleaner to just include the symlink into git. This would also make your `$XDG_DATA_DIRS` scenario actually work -- we don't bother with `make` any more when developing in pkg/, it's all webpack-watch and friends now.

martinpitt (changes requested):
Thanks! Some minor comments, aside from the test failure.

pkg/lib/cockpit-po-plugin.js
martinpitt: We resolved/nack'ed the referenced comment, so this comment is obsolete as well.

pkg/build
allisonkarlitskaya: The problem with this is the `cockpit/` link goes to `dist/` which means that if we want to install these files with the `nobase_` rule, we need to figure out a way to get the manifests from `pkg/` to `dist/`.

Alternatives are:
 - copy.  This is what we used to do.
 - add two separate/simple rules (with corresponding `dir` variables) for installing these two files and completely remove them from the `MANIFESTS` list.

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: If we stop running ssh and pcp through webpack, then I certainly care less about this, but I still think that this change is correct. Let's have a call?

pkg/Makefile.am
allisonkarlitskaya: Sure.

src/ws/test-server.c
allisonkarlitskaya: There's actually three potential cases here:

 - request the .gz file explicitly, and want to receive .gz data
 - request the base file, with `Accept-Encoding: gzip` and expect to get the compressed stream
 - request the base file, with `Accept-Encoding: none` and make sure the transparent decompression happens

Do I understand correctly that you want to test the second two?

pkg/lib/cockpit-po-plugin.js
martinpitt: Yes, it's correct (when assuming that we *do* want to load base1 po separately) -- I already closed the conversation

pkg/build
martinpitt: Ah, I didn't realize that this commit was a prerequisite for the subsequent ones, or just standalone -- as by itself, the little clenaup isn't worth the overhead. If it helps to enable the compression-webpack plugin and such, it's ok for me.

src/ws/test-server.c
martinpitt: Well, all cases should be covered/used somewhere -- otherwise why add the functionality in the first place? Or does this get implicitly covered by some later commit? (I didn't see any which would use that)

martinpitt: FTR: as webpack now spits out .gz files, test-server does use them. So it's a preprequisite, and covered implicitly.

pkg/build
allisonkarlitskaya: Dropped this after our chat: let's just add the 4 lines to the build system to deal with these manually.

pkg/Makefile.am
allisonkarlitskaya: Not doing this, because of the automake tar issue.  For the record: creating a tar of the dist *must* dereference symlinks because of the symlinks for `tools/missing` and friends.

martinpitt (approved):
Cheers! Looks nice to me now. Thanks for bearing with me :grinning:

allisonkarlitskaya:
> Cheers! Looks nice to me now. Thanks for bearing with me grinning

As usual, thanks for pushing me to make it better :)

martinpitt:
[cockpit-pcp crash](https://logs.cockpit-project.org/logs/pull-16982-20220311-124341-bd412312-fedora-35-devel/log.html#209), retrying

PR #16984, opened 2022-02-11
test: Fix use of f-string

marusak:
As seen [here](https://logs.cockpit-project.org/logs/pull-16981-20220211-141347-7844ce1d-fedora-34/log.html#275):
```
Failed to issue method call: No session '%s' known
```


martinpitt:
Argh, thanks @marusak !

PR #16993, opened 2022-02-14
Fix user test races

marusak:
1. Enter long password (we enter 300 characters long)
2. Click on 'Set password' button - that triggers `validate()`
3. `validate()` sets `errors.password` to "Password is longer than 256 characters" and it is shown in the UI
4. Then `validate()` calls `password_quality()`. This takes longer because it checks 300 characters long password
5. But since the UI is updated with the error, bots type in new correct password
6. `change()` function that is invoked when inputs change clears the `errors` object
7. Now that `password_quality()` finally resolved, it checks if there are errors
   in `errors` object, but that was already cleared. So it succeeds.
8. `validate()` succeeded, we get back to step 2, the promise resolved,
   so it calls `passwd_change(account.name, state.password)`
9. `state.password` is the already new typed in password, so it also
   succeeds as it is not the long one anymore.
10. We set up password from step 5 while checking password from step 1

### DIFF
--- a/pkg/users/account-create-dialog.js
+++ b/pkg/users/account-create-dialog.js
@@ -192,29 +192,32 @@ export function account_create_dialog(accounts) {
         update();
     }
 
-    function validate(force) {
-        errors = { };
+    function validate(force, real_name, user_name, password, password_confirm) {
+        const errs = { };
 
-        if (!state.real_name)
+        if (!real_name)
             errors.real_name = _("No real name specified");
 
-        if (state.password != state.password_confirm)
-            errors.password_confirm = _("The passwords do not match");
+        if (password != password_confirm)
+            errs.password_confirm = _("The passwords do not match");
 
-        if (state.password.length > 256)
-            errors.password = _("Password is longer than 256 characters");
+        if (password.length > 256)
+            errs.password = _("Password is longer than 256 characters");
 
-        errors.user_name = validate_username(state.user_name, accounts);
+        errs.user_name = validate_username(user_name, accounts);
 
-        return password_quality(state.password, force)
+        return password_quality(password, force)
                 .catch(ex => {
-                    errors.password = (ex.message || ex.toString()).replace("\n", " ");
-                    errors.password += "\n" + cockpit.format(_("Click $0 again to use the password anyway."), _("Create"));
+                    errs.password = (ex.message || ex.toString()).replace("\n", " ");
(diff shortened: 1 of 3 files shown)


marusak:
One more user flake (much less frequent) is this: 
```
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 1351, in terminate_sessions
    m.execute(f"while loginctl show-session {s}; do sleep 1; done")
  File "/work/bots/make-checkout-workdir/bots/machine/machine_core/ssh_connection.py", line 334, in execute
    ret = select.select(rset, wset, [], 10)
  File "/work/bots/make-checkout-workdir/bots/machine/machine_core/timeout.py", line 41, in handle_timeout
    raise RuntimeError(self.error_message)
RuntimeError: Timed out on 'while loginctl show-session 176; do sleep 1; done'
```
[example](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16938-20220214-145222-37e948c0-fedora-34/log.html#39)

martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-16993-20220214-162516-f665c18d-debian-stable/log.html#301) is yet another unrelated flake, right?

martinpitt:
Current flakes:
 - [c8s](https://logs.cockpit-project.org/logs/pull-16993-20220214-162516-f665c18d-centos-8-stream/log.html#108): Just fixed in PR #16994, retry should work.
 - [f34](https://logs.cockpit-project.org/logs/pull-16993-20220214-162516-f665c18d-fedora-34/log.html#300): Investigating in #16996 , need to land that to know more
 - [f35](https://logs.cockpit-project.org/logs/pull-16993-20220214-162516-f665c18d-fedora-35/log.html#55): My recent fix to stop virbr0 was apparently not sufficient, queueing for tomorrow
 - [rhel-9-0](https://logs.cockpit-project.org/logs/pull-16993-20220214-162516-f665c18d-rhel-9-0/log.html#296-2): services page stuck on rhel 9, seen before, unrelated

I.e. all unrelated. One more round, then I'll merge *throws hands into the air*

marusak:
I have seen `b.wait_in_text('#accounts-list', "anton")` many times before.
We watch `/etc/passwd` and the page just does not update. I have some sneaky suspision that our `cockpit.file.watch` dies or so, similar issue is also reoccurring [flake](https://logs.cockpit-project.org/logs/pull-16981-20220211-141347-7844ce1d-rhel-9-0/log.html#177-2) where we watch a file and expect it to update. Every now and then it does not.

PR #16994, opened 2022-02-14
test: Ignore message on missing /run/systemd/coredump

martinpitt:
This can happen when something crashes during shutdown/reboot, when the
socket is already gone.

----

[example](https://logs.cockpit-project.org/logs/pull-16987-20220214-130456-74c31fa5-centos-8-stream/log.html#124)

This is rather rare, so it most likely won't happen in this PR.


marusak (approved):
Thanks!

martinpitt:
[This c8s packit failure](http://artifacts.dev.testing-farm.io/323043d2-b07f-4814-b5ae-f889afcb8991/) was introduced today, most likely from PR #16938. I'll adjust it tomorrow.

Issue #16972, opened 2022-02-10
cockpit hardware information is wrong on a Raspberry Pi 3


jelly:
### Explain what happens

On my RPI 3 cockpit does not show the model 

![image](https://user-images.githubusercontent.com/67428/153428487-03a7e730-69e5-4543-94cc-18b5432e15f6.png)

![image](https://user-images.githubusercontent.com/67428/153428509-bba25d6b-4fd0-4750-929a-1d8cf723c38b.png)

Device tree has the correct information:
```
[jelle@fedora ~]$ sudo cat /proc/device-tree/model
Raspberry Pi 3 Model B Rev 1.2

[jelle@fedora ~]$ cat /proc/device-tree/serial-number
00000000895f62e7
```

dmidecode
```
Getting SMBIOS data from sysfs.
SMBIOS 3.0 present.
7 structures occupying 265 bytes.
Table at 0x3CB3F020.

Handle 0x0000, DMI type 0, 24 bytes
BIOS Information
	Vendor: U-Boot
	Version: 2021.10
	Release Date: 10/01/2021
	ROM Size: 64 kB
	Characteristics:
		PCI is supported
		BIOS is upgradeable
		Selectable boot is supported
		Targeted content distribution is supported
		UEFI is supported
	BIOS Revision: 21.10

Handle 0x0001, DMI type 1, 27 bytes
System Information
	Manufacturer: Unknown
	Product Name: Unknown Product
	Version: Not Specified
	Serial Number: 00000000895f62e7
	UUID: 30303030-3030-3030-3839-356636326537
	Wake-up Type: Reserved
	SKU Number: Not Specified
	Family: Not Specified

Handle 0x0002, DMI type 2, 14 bytes
Base Board Information
	Manufacturer: Unknown
	Product Name: Unknown Product
	Version: Not Specified
	Serial Number: Not Specified
	Asset Tag: Not Specified
	Features:
		Board is a hosting board
	Location In Chassis: Not Specified
	Chassis Handle: 0x0000
	Type: Motherboard

Handle 0x0003, DMI type 3, 21 bytes
Chassis Information
	Manufacturer: Unknown
	Type: Desktop
	Lock: Not Present
	Version: Not Specified
	Serial Number: Not Specified
	Asset Tag: Not Specified
	Boot-up State: Safe
	Power Supply State: Safe
	Thermal State: Safe
	Security Status: None
	OEM Information: 0x00000000
	Height: Unspecified
[...]

PR #16995, opened 2022-02-14
Hardware information improvements

jelly:
This PR improves the hardware information shown on ARM devices such as the Raspberry Pi and shows why there is no memory information shown for normal users.

image

### DIFF
--- a/pkg/systemd/hwinfo.jsx
+++ b/pkg/systemd/hwinfo.jsx
@@ -36,6 +36,7 @@ import {
     DataListItemCells,
     DataListAction,
     DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
+    EmptyState,
     Flex, FlexItem,
     Gallery,
     Page, PageSection,
@@ -50,6 +51,7 @@ import { ListingTable } from "cockpit-components-table.jsx";
 import kernelopt_sh from "raw-loader!./kernelopt.sh";
 import detect from "./hw-detect.js";
 
+import { superuser } from "superuser";
 import { PrivilegedButton } from "cockpit-components-privileged.jsx";
 
 const _ = cockpit.gettext;
@@ -311,6 +313,10 @@ class HardwareInfo extends React.Component {
                         columns: [dimm.locator, dimm.technology, dimm.type, dimm.size, dimm.state, dimm.rank, dimm.speed]
                     })) } />
             );
+        } else if (!superuser.allowed) {
+            memory = (<EmptyState>
+                {_("Viewing memory information requires administrative access.")}
+            </EmptyState>);
         }
 
         return (
--- a/pkg/systemd/hw-detect.js
+++ b/pkg/systemd/hw-detect.js
@@ -126,7 +126,7 @@ export default function detect() {
 
     tasks.push(machine_info.cpu_ram_info()
             .then(result => {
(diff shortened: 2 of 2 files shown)


pkg/systemd/hwinfo.jsx
```diff
@@ -118,7 +118,7 @@ class SystemInfo extends React.Component {
                         { info.nproc !== undefined && <>
                             <DescriptionListGroup>
                                 <DescriptionListTerm>{ _("CPU") }</DescriptionListTerm>
-                                <DescriptionListDescription>{ (info.nproc > 1) ? `${info.nproc}x ${info.cpu_model}` : info.cpu_model }</DescriptionListDescription>
+                                <DescriptionListDescription>{ (info.nproc > 1) ? `${info.nproc}x ${info.cpu_model || _("unknown")}` : info.cpu_model }</DescriptionListDescription>
```
martinpitt, line 123: Please fix this for the single-CPU case as well. I.e. pull this out into a new constant to avoid repetition.

pkg/systemd/overview-cards/systemInformationCard.jsx
```diff
@@ -83,6 +83,18 @@ export class SystemInfomationCard extends React.Component {
...
                     self.setState({ assetTagText: fields.product_serial || fields.chassis_serial || undefined });
+
+                    // The Raspberry Pi 3 Model B Rev 1.2 has /sys/class/dmi/id/* but it's filled with "Unknown" values when using
+                    // U-boot with an EFI payload as is the case with Fedora Server / IoT.
+                    machine_info.devicetree_info()
+                            .then(dev_fields => {
+                                if (!fields.product_serial) {
+                                    self.setState({ assetTagText: dev_fields.serial });
+                                }
+                                if ((!vendor && !name && dev_fields.model) || (dev_fields.model && vendor && vendor.includes("Unknown"))) {
```
martinpitt, line 94: Let's please rather filter this out in pkg/lib/machine-info.js. That already does a similar hack for the dreaded "To be filled by O.E.M." value. Then you can factorize this into some `is_invalid_value()` helper and use it in DMI and DT.

pkg/systemd/hwinfo.jsx
```diff
@@ -316,6 +318,10 @@ class HardwareInfo extends React.Component {
                         columns: [dimm.locator, dimm.technology, dimm.type, dimm.size, dimm.state, dimm.rank, dimm.speed]
                     })) } />
             );
+        } else if (!superuser.allowed) {
+            memory = (<EmptyState>
+                {_("Viewing memory information requires admin privileges")}
```
martinpitt, line 323: We call it "administrative access". Please also end the sentence with a period.

martinpitt (changes requested):
Thanks! Sorry that this slipped through the cracks for so long :crying_cat_face:

pkg/systemd/overview-cards/systemInformationCard.jsx
jelly: Dropping this out because of this comment https://github.com/cockpit-project/cockpit/issues/16972#issuecomment-1041455088

As a nice exercise I've build my own u-boot with a patch from #u-boot to see if I can easily fix it in upstream.

martinpitt (approved):
Thanks!

PR #16999, opened 2022-02-14
test: Go back to fewer global machines

martinpitt:
Commit 4999a18a increased the number of parallel nondestructive tests
(global machines) to the full number of `$TEST_JOBS`. Since around that
time we have no end of test flakes and browser timeouts. While a lot of
the failures are real, they are a bit much to keep up with all at once.
The increased number of random white screens and timeouts are likely due 
to a lot more browsers being started and stopped in parallel, which
happens much faster with nondestructive tests. This clogs our infra's
tasks containers too much.

So for now, go back to the old distribution of destructive vs. 
nondestructive tests. This still retains the better balancing between
the nondestructive runners and the more efficient packing of the 
destructive tests, thus does not actually increase the wallclock test
runtime.

---

This is an emergency measure. But first I want to see some results here whether this actually helps.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -345,8 +345,9 @@ def run(opts, image):
     if not opts.machine:
         # Create appropriate number of serial machines; prioritize the nondestructive (serial) tests, to get
         # them out of the way as fast as possible, then let the destructive (parallel) ones start as soon as
-        # a given serial runner is done.
-        num_global = min(serial_tests_len, opts.jobs)
+        # a given serial runner is done. However, limit the number of parallel nondestructive tests, as they
+        # are currently too flaky to run with full parallelism.
+        num_global = min(serial_tests_len, opts.jobs // 2)
 
         for i in range(num_global):
             global_machines.append(GlobalMachine(restrict=not opts.enable_network, cpus=opts.nondestructive_cpus,


test/common/run-tests
```diff
@@ -346,7 +346,7 @@ def run(opts, image):
         # Create appropriate number of serial machines; prioritize the nondestructive (serial) tests, to get
         # them out of the way as fast as possible, then let the destructive (parallel) ones start as soon as
```
martinpitt, line 347: This comment needs adjusting if we actually go ahead with this.

martinpitt:
After first run: This doesn't destroy our recent speedups -- e.g. [rhel-9-0](https://logs.cockpit-project.org/logs/pull-16999-20220214-184732-b47b7d00-rhel-9-0/log) still takes ~ 45 mins. We still retain the balanced distribution of nondestructive tests and more efficient packing of destructive ones.

The failures were [f34](https://logs.cockpit-project.org/logs/pull-16999-20220214-184732-b47b7d00-fedora-34/log.html#299-2), [f35](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16999-20220214-184732-b47b7d00-fedora-35/log.html), [f35-devel](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16999-20220214-184732-b47b7d00-fedora-35-devel/log.html) and [f35-devel](https://logs.cockpit-project.org/logs/pull-16999-20220214-184732-b47b7d00-fedora-35-firefox/log.html#253-2). TestAccounts just got fixed in another PR, and it seems `TestJournal.testAbrtReportCancel` is *really* screwed. We must have broken that recently with some other fixes.

But all in all this smells a lot better. I'll clean up the comment/commit message and let this run another round.

martinpitt:
About [TestJournal.testAbrtReportCancel](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16999-20220214-194134-4b331678-fedora-35/log.html). This reproduces locally. After this happens, I can even log into cockpit with a fresh browser, and it is *still* wrong ("Report to Fedora" instead of "Report to Cockpit").

```
# busctl call org.freedesktop.problems /org/freedesktop/Problems2 org.freedesktop.Problems2 GetProblems 'ia{sv}' 0 0
ao 1 "/org/freedesktop/Problems2/Entry/820002045f3c85f4723c4c00471b980c0532c835"

# busctl call org.freedesktop.reportd /org/freedesktop/reportd/Service org.freedesktop.reportd.Service GetWorkflows o /org/freedesktop/Problems2/Entry/820002045f3c85f4723c4c00471b980c0532c835
a(sss) 1 "workflow_FedoraCCpp" "Report to Fedora" "Process the C/C++ crash using the Fedora infrastructure"
```

Restarting `reportd.service` seems to help. I suppose initially it gets auto-activated, and then the next time it does not re-read its new configuration?

martinpitt:
Interestingly, the [Services page times out](https://logs.cockpit-project.org/logs/pull-16999-20220214-194134-4b331678-rhel-8-6/log.html#273-2) even with this reduced parallelism, and [other common flakes](https://logs.cockpit-project.org/logs/pull-16999-20220214-194134-4b331678-ubuntu-stable/log.html) also still appear. Maybe that's not *that* helpful after all?

martinpitt:
I found out that half of our e2e machines don't have a tmpfs on /tmp, i.e. qemu overlays are running on spinning rust. This would totally explain the long runtimes and timeouts!

martinpitt:
I sent https://github.com/cockpit-project/cockpituous/pull/456 . That sounded like a much more plausible explanation for why tests were so sh*tty recently. Let's give that a few rounds.

PR #17000, opened 2022-02-14
test: Factorize and strengthen disabling of virbr0

martinpitt:
Commit e3802c02407 was not strong enough -- sometimes, virbr0 still came
online in the middle of a test. This also sometimes disrupts the
Firewall tests when they count the expected zones at the beginning, but
a second one (for virbr0) suddenly springs into existence.

Stop this harder by also disabling autostart. Also centralized it in
NetworkCase.setUp().

Avoid clobbering the `m` machine alias in setUp with a regexp match,
rename the variable.

---

[example failure](https://logs.cockpit-project.org/logs/pull-16993-20220214-162516-f665c18d-fedora-35/log.html#55)


marusak (approved):
Thanks!

PR #17001, opened 2022-02-15
test: Restart reportd after installing custom configuration

martinpitt:
If the daemon is already running when installing the mock
config/workflow, it will not pick it up.

---

[example](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-16999-20220214-194134-4b331678-fedora-35/log.html#253-2), [example](https://logs.cockpit-project.org/logs/pull-16999-20220214-194134-4b331678-fedora-34/log.html#253-2)

[debugging notes](https://github.com/cockpit-project/cockpit/pull/16999#issuecomment-1039858186)


marusak (approved):
Hmm, how was this supposed to be working before?
[this abrt crash](https://logs.cockpit-project.org/logs/pull-17001-20220215-061845-0d61751f-fedora-34/log.html#296) I believe is due to reportd crashing. We can confirm when we have journal - there is naughty for it already.

martinpitt:
> Hmm, how was this supposed to be working before?

Sheer luck with test order,  I think -- it was relying on reportd.service to not run at the start of the test?

PR #17004, opened 2022-02-15
test: Fix race condition in TestFirewall.testNetworkingPage

martinpitt:
This test often failed because starting firewalld asynchronously started
the libvirt default network, and then added a new "libvirt" zone for it.
This then broken the zone count between the test's initial
`active_zones` count and what the UI actually displayed later on.

There was a special case for ubuntu-2004 which fixed that, but it did
not run on any other OS (the same thing happens on ubuntu-stable and
both debian-*).

Commit 7fabac288b6 was supposed to fix this for all networking tests,
but it did not apply to TestFirewall as that just skipped NetworkCase's
constructor. Call the proper parent constructor to include that, and
drop the now obsolete ubuntu-2004 special case.

---

See [weather report](https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&days=7&test=test%2Fverify%2Fcheck-networkmanager-firewall+TestFirewall.testNetworkingPage) and [example](https://logs.cockpit-project.org/logs/pull-17003-20220215-072756-28540dbb-debian-testing/log.html#70)


marusak (approved):
Danke!

martinpitt:
Urgh, this [totally breaks arch](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17004-20220215-094144-43273621-arch/log.html)

martinpitt:
But wait, what? The breakage happens right at the very first test (`TestAccounts.testAccountLogs`), it didn't even get near the firewall test yet. Some problem on CentOS CI? Retrying.

martinpitt:
arch now fails everywhere, due to:

    cockpit-certificate-ensure[750]: /usr/lib/cockpit/cockpit-certificate-ensure: /usr/lib/libc.so.6: version `GLIBC_2.34' not found (required by /usr/lib/cockpit/cockpit-certificate-ensure)

so it now gets mis-built. That's what we get for not fixing the build env on our image, but always download it off the internet..

@jelly , any idea?

jelly:
> arch now fails everywhere, due to:
> 
> ```
> cockpit-certificate-ensure[750]: /usr/lib/cockpit/cockpit-certificate-ensure: /usr/lib/libc.so.6: version `GLIBC_2.34' not found (required by /usr/lib/cockpit/cockpit-certificate-ensure)
> ```
> 
> so it now gets mis-built. That's what we get for not fixing the build env on our image, but always download it off the internet..
> 
> @jelly , any idea?

The Image has an old glibc and `extra-x86_64-build` build with a (just out) new glibc. This PR addresses it with a hack https://github.com/cockpit-project/bots/pull/2944

PR #17005, opened 2022-02-15
Moar nondestructive tests

jelly:
This seems to be the last "low hanging fruit".

### DIFF
--- a/test/run-verify-host-user.sh
+++ b/test/run-verify-host-user.sh
@@ -167,6 +167,15 @@ if [ -n "$test_basic" ]; then
     if [ "$TEST_OS" = "fedora-36" ]; then
         EXCLUDES="$EXCLUDES TestSOS.testBasic"
     fi
+
+    # Exclude NFS tests due to a timing issue https://github.com/cockpit-project/cockpit/issues/17142
+    EXCLUDES="$EXCLUDES
+            TestStorageNfs.testNfsBusy
+            TestStorageNfs.testNfsClient
+            TestStorageNfs.testNfsListExports
+            TestStorageNfs.testNfsMountWithoutDiscovery
+            TestStoragePackagesNFS.testNfsMissingPackages
+            "
 fi
 
 exclude_options=""
--- a/test/verify/check-storage-nfs
+++ b/test/verify/check-storage-nfs
@@ -23,6 +23,7 @@ from storagelib import *
 from testlib import *
 
 
+@nondestructive
 class TestStorageNfs(StorageCase):
 
     def testNfsClient(self):
@@ -31,8 +32,11 @@ class TestStorageNfs(StorageCase):
 
         self.login_and_go("/storage")
 
-        m.execute("mkdir /home/foo /home/bar")
-        m.write("/etc/exports", "/home/foo 127.0.0.0/24(rw)\n/home/bar 127.0.0.0/24(rw)\n")
+        m.execute("mkdir /home/foo /home/bar /mnt/test")
+        # Removing the nfs mount removes the target dir.
+        self.addCleanup(m.execute, "! test -e /mnt/test || rm -r /mnt/test")
+        self.write_file("/etc/exports", "/home/foo 127.0.0.0/24(rw)\n/home/bar 127.0.0.0/24(rw)\n",
+                        post_restore_action="systemctl restart nfs-server")
         m.execute("systemctl restart nfs-server")
 
         # Nothing there in the beginnging
@@ -43,17 +47,17 @@ class TestStorageNfs(StorageCase):
         self.dialog_wait_open()
         self.dialog_set_val("server", "127.0.0.1")
         self.dialog_set_val("remote", "/home/foo")
-        self.dialog_set_val("dir", "/mnt")
+        self.dialog_set_val("dir", "/mnt/test")
         self.dialog_apply()
         self.dialog_wait_close()
 
         b.wait_visible("#nfs-mounts td:contains(/home/foo)")
-        b.wait_visible("#nfs-mounts td:contains(/mnt)")
-        b.wait_text_not("#nfs-mounts tr:contains(/mnt) .pf-c-progress__status", "")
+        b.wait_visible("#nfs-mounts td:contains(/mnt/test)")
+        b.wait_text_not("#nfs-mounts tr:contains(/mnt/test) .pf-c-progress__status", "")
 
         # Should be saved to fstab
         self.assertEqual(m.execute("grep -w nfs /etc/fstab").strip(),
-                         "127.0.0.1:/home/foo /mnt nfs defaults")
+                         "127.0.0.1:/home/foo /mnt/test nfs defaults")
 
         # Try to add some non-exported directory
         b.click("#nfs-mounts button[aria-label=Add]")
@@ -93,6 +97,7 @@ class TestStorageNfs(StorageCase):
         self.dialog_set_val("dir", "/mounts/barbar")
         self.dialog_apply()
         self.dialog_wait_close()
+        self.addCleanup(m.execute, "set -e; umount /mounts/barbar; rmdir /mounts/barbar")
 
         b.wait_text('#detail-header dt:contains("Mount point") + dd', "/mounts/barbar")
         m.execute("! test -e /mounts/bar")
@@ -129,7 +134,7 @@ class TestStorageNfs(StorageCase):
         b.click("#nfs-mounts tr:contains(/home/foo)")
         b.wait_visible('#storage-detail')
         b.wait_text('#detail-header dt:contains("Server") + dd', "127.0.0.1:/home/foo")
-        b.wait_text('#detail-header dt:contains("Mount point") + dd', "/mnt")
+        b.wait_text('#detail-header dt:contains("Mount point") + dd', "/mnt/test")
 
         # Unmount and remount /home/foo
         b.click("#detail-header button:contains(Unmount)")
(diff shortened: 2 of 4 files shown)


martinpitt (changes requested):
Cheers! Let's clean this up a little, though.

test/verify/check-superuser
```diff
@@ -161,6 +165,7 @@ session include system-auth
         # "admin" as its primary group.
         #
         if "ubuntu" in m.image:
+            self.restore_file("/etc/sudoers")
```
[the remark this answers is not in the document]
jelly: ah sed_file it is!

test/verify/check-superuser
```diff
@@ -77,6 +77,7 @@ class TestSuperuser(MachineCase):
         b.leave_page()
         b.check_superuser_indicator("Administrative access")
 
+    @nondestructive
```
martinpitt, line 80: OOI, at first sight I don't see anything in testBasic which would be destructive? Possibly you could make the whole class `@nond`? (Follow-up material)

test/verify/check-superuser
```diff
@@ -86,7 +87,7 @@ class TestSuperuser(MachineCase):
         b.check_superuser_indicator("Limited access")
 
         # Give us password-less sudo and use it
-        m.write("/etc/sudoers.d/admin", "admin ALL=(ALL) NOPASSWD:ALL")
+        m.write_file("/etc/sudoers.d/admin", "admin ALL=(ALL) NOPASSWD:ALL")
```
martinpitt, line 90: `self`, not `m`. It has to be on teh MachineCase class so that it can hook into `.addCleanup()`.

test/verify/check-storage-nfs
```diff
@@ -32,6 +33,8 @@ class TestStorageNfs(StorageCase):
         self.login_and_go("/storage")
 
         m.execute("mkdir /home/foo /home/bar")
+        self.addCleanup(m.execute, "systemctl restart nfs-server")
+        self.restore_file("/etc/exports")
```
martinpitt, line 37: restore_file() has a `post_restore_action` for that purpose; reads a little nicer to keep things together which belong together.

test/verify/check-storage-nfs
```diff
@@ -189,13 +195,16 @@ class TestStorageNfs(StorageCase):
         self.login_and_go("/storage")
 
         m.execute("mkdir /home/foo /home/bar")
+        self.addCleanup(m.execute, "systemctl restart nfs-server")
         m.write("/etc/exports", "/home/foo 127.0.0.0/24(rw)\n/home/bar 127.0.0.0/24(rw)\n")
+        self.restore_file("/etc/exports")
```
martinpitt, line 200: Again, write_file(). Please feel free to add and plumb through a new `post_restore_action` argument to write_file() to simplify/unify these cases.

martinpitt (changes requested):

test/verify/check-superuser
jelly: What about,
```
          m.execute("rm -rf /var/{db,lib}/sudo/lectured/admin")
```

martinpitt: It's not super important to restore it; but in principle, you could use `self.restore_dir()` or `restore_file()` and replace the braces with a conditional or two calls.

jelly: Okido fixed.

jelly:
Arch image seems to flake a lot and I'm not sure what's up with packit.

test/verify/check-storage-nfs
```diff
@@ -93,6 +95,7 @@ class TestStorageNfs(StorageCase):
         self.dialog_set_val("dir", "/mounts/barbar")
         self.dialog_apply()
         self.dialog_wait_close()
+        self.addCleanup(m.execute, "umount /mounts/barbar && rmdir /mounts/barbar")
```
martinpitt, line 98: Using && here is dangerous, as (annoyingly in this context), `false && true` is true (i.e. doesn't fail); use `;` and add `set -e`, or have two separate commands. We want to know when the umount fails, as it leaves the VM in an unclean state.

martinpitt (changes requested):
Thanks! 

The [wrong username or password](https://logs.cockpit-project.org/logs/pull-17005-20220223-095055-789cd51b-arch/log.html) arch test sounds very suspicious, though -- some leftover from a test that run before on the same ND@n machine?

test/verify/check-storage-nfs
jelly: Correct.

jelly:
gt

> > Thanks!
> > The [wrong username or password](https://logs.cockpit-project.org/logs/pull-17005-20220223-095055-789cd51b-arch/log.html) arch test sounds very suspicious, though -- some leftover from a test that run before on the same ND@n machine?
> 
> Ok, will try to reproduce it locally.

Ok, reproduced:

TestSuperuser.testWrongPasswd

Then run TestTerminal.testOnlyTermnal

jelly:
So, after some debugging with TMT a few times I manged to get screenshots and some output:

```
14:17:42                 out: -> wait: ph_has_val("#dialog [data-field=\"remote\"]","/home/foo")
14:17:42                 out: <- {'exceptionDetails': {'exception': {'type': 'string', 'value': 'timeout'}}}
14:17:42                 out: Traceback (most recent call last):
14:17:42                 out:   File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
14:17:42                 out:     method()
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/verify/check-storage-nfs", line 47, in testNfsClient
14:17:42                 out:     self.dialog_set_val("remote", "/home/foo")
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/verify/storagelib.py", line 269, in dialog_set_val
14:17:42                 out:     self.browser.set_input_text(sel, val)
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/common/testlib.py", line 402, in set_input_text
14:17:42                 out:     self.wait_val(selector, val)
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/common/testlib.py", line 487, in wait_val
14:17:42                 out:     self.wait_js_func('ph_has_val', selector, val)
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/common/testlib.py", line 467, in wait_js_func
14:17:42                 out:     self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/common/testlib.py", line 457, in wait_js_cond
14:17:42                 out:     self.raise_cdp_exception("timeout\nwait_js_cond", cond, result["exceptionDetails"], trailer)
14:17:42                 out:   File "/var/tmp/tmt/run-012/plans/all/discover/default/tests/test/common/testlib.py", line 241, in raise_cdp_exception
14:17:42                 out:     raise Error("%s(%s): %s" % (func, arg, msg))
14:17:42                 out: testlib.Error: timeout
```

image

Locally I can at least reproduce the empty input field, but the test passes.

So in the test it goes into:
```
          elif ftype == "text-input":
              self.browser.set_input_text(sel, val)
```

While the shown component is:

```
<div data-field="remote" data-field-type="combobox"><div class="pf-c-select" data-ouia-component-type="PF4/Select" data-ouia-safe="true" data-ouia-component-id="OUIA-Generated-Select-typeahead-1"><div class="pf-c-select__toggle pf-m-typeahead"><div class="pf-c-select__toggle-wrapper"><input class="pf-c-form-control pf-c-select__toggle-typeahead" id="pf-select-toggle-id-0-select-typeahead" aria-label="" placeholder="" type="text" autocomplete="off" value=""></div><button id="pf-select-toggle-id-0" aria-labelledby=" pf-select-toggle-id-0" aria-expanded="false" aria-haspopup="listbox" type="button" class="pf-c-button pf-c-select__toggle-button pf-m-plain" aria-label="Options menu" tabindex="-1"><svg style="vertical-align: -0.125em;" fill="currentColor" height="1em" width="1em" viewBox="0 0 320 512" aria-hidden="true" role="img" class="pf-c-select__toggle-arrow"><path d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"></path></svg></button></div></div></div>
```

martinpitt:
@jelly : `TestStorageNfs.testNfsBusy` is still running on packit, I thought you wanted to exclude it for the time being?

test/run-verify-host-user.sh
```diff
@@ -167,6 +167,9 @@ if [ -n "$test_basic" ]; then
     if [ "$TEST_OS" = "fedora-36" ]; then
         EXCLUDES="$EXCLUDES TestSOS.testBasic"
     fi
+
+    # Exclude NFS tests due to a timing issue https://github.com/cockpit-project/cockpit/issues/17142
+    EXCLUDES="$EXCLUDES TestStorageNfs"
```
martinpitt, line 172: Oh, you need to specify the full name. Unlike the opt-in "tests" list, this needs to be precise and not substrings (this would be too error prone otherwise).

martinpitt (approved):
Cheers! "Low hanging fruit" and other lies developers are telling you :sweat_smile:

martinpitt:
This now causes [regressions](https://logs.cockpit-project.org/logs/pull-17135-20220317-081127-f0ffb5db-fedora-35-firefox/log.html#189-2) when running multiple times. Odd, the tests ran here three times as well, perhaps different order? I'll fix it in PR #17135

PR #17007, opened 2022-02-15
test: Allow exceptions while waiting

mvollmer:
ph_wait now allows all exceptions and keeps trying.  If the last try
has an exception, it will be reported.  This means that we don't lose
any exceptions that happen consistently, but we always have to wait
for the timeout.

Browser.wait_js_cond retries when the execution context is not found
or disappears in the middle.  To support this, getFrameExecId never
raises an exception itself but returns -1 so that Runtime.evaluate
raises an error that causes Browser.wait_js_cond to retry.

Together these changes mean that calls to functions like
Browser.wait_in_text can now survive a page reload in their middle.

Consequently, all calls to expect_load and expect_load_frame have been
removed.

Specifically, this fixes a race in the TestKeys-testAuthorizedKeys
test case.  That test case experiences a page reload that is triggered
asynchronously and can therefore not be reliably waited for with
expect_load_frame.  (The call to enter_page was essentially a no-op.)


mvollmer:
I have reproduced the original race by delaying the call to `window.location.reload` in pkg/lib/superuser.js by 5 seconds.

mvollmer:
> They would just need to look up the execution context in each iteration instead of just once at the beginning, I think.

And ignore the exceptions, of course.

martinpitt:
Hmm, this [fails](https://logs.cockpit-project.org/logs/pull-17007-20220215-113751-c9dffc75-ubuntu-2004/log.html) [everywhere](https://logs.cockpit-project.org/logs/pull-17007-20220215-113751-c9dffc75-rhel-9-0/log.html#277) :cry:

mvollmer:
> Hmm, this [fails](https://logs.cockpit-project.org/logs/pull-17007-20220215-113751-c9dffc75-ubuntu-2004/log.html) [everywhere](https://logs.cockpit-project.org/logs/pull-17007-20220215-113751-c9dffc75-rhel-9-0/log.html#277) cry

Yeah, that's the race "in the other direction" mentioned in the description.

mvollmer:
> But I think the best option here is to make all `wait`s robust against reloads in their middle.

Hmm, that turns out to be hairy.  If this typical sequence `wait(ph_is_present), wait(ph_is_visible), wait(ph_in_text)` is interrupted by a reload in `wait(ph_in_text)`, we need to go back to `wait(ph_is_present)` to make this all work.

So I try the deracing of expect_load_frame.

mvollmer:
> If this typical sequence `wait(ph_is_present), wait(ph_is_visible), wait(ph_in_text)` is interrupted by a reload in `wait(ph_in_text)`, we need to go back to `wait(ph_is_present)` to make this all work.

Or! We just allow any kind of exception in ph_in_text and keep waiting.  This is what the global `wait` function in testlib does, for example.

martinpitt:
@mvollmer : Should not be too hard to detect the "execution context destroyed" error, we even log it. If that happens during a wait, it seems fine to me to automatically continue to wait. With that we could probably drop all our expect_load()s. I would not ignore *all* errors, that's too confusing.

mvollmer:
> I would not ignore _all_ errors, that's too confusing.

I would actually, or rather I would allow them all temporarily and only raise the last one when the timeout triggers.

test/common/chromium-cdp-driver.js
```diff
@@ -241,10 +241,10 @@ function getFrameExecId(frame) {
         frame = "cockpit1";
     var frameId = frameNameToFrameId[frame];
     if (!frameId)
-        throw Error(`Frame ${frame} is unknown`);
+        return -1;
```
martinpitt, line 244: Needs the same thing on firefox-cdp-driver.js? Or more cleanly, leave this alone and instead catch the exception in the wait loop only, but not anywhere else?

martinpitt (commented):
I didn't review this in detail yet, but cool general approach!

mvollmer:
First test run looks pretty good! I'll check all removed expect_load calls to make sure there is a wait as the next call that will now do the expected waiting.

mvollmer:
`expect_load` needs to be removed from testlib.py as well.  I let the bots finish the current run.

mvollmer:
At least b.reload() really needs to wait until the page has been loaded.  But this can be done with pretty simple local code.

mvollmer:
There is a race in how the PR synthesizes the "context destroyed" error, and we might get an additional error from Runtime.evaluate, sometimes.  I have pushed a complicated sequence machinery that I had also used for debugging to prevent this, so the firefox tests should be greener now. I try to find a simpler solution. (Also, there is still a sleep missing in testlib, which will reduce the stress when installed. But let's keep the stress up a bit for now.)

mvollmer:
Note to self:

```
RETRY !ph_is_visible(".curtains-ct") Execution context was destroyed.
OUT OF SEQUENCE (success)
> log: CONTINUE WAIT .curtains-ct not found
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/work/bots/make-checkout-workdir/test/verify/check-pages", line 188, in testBasic
    b.enter_page("/system")
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 595, in enter_page
    self.wait_visible("iframe.container-frame[name='%s']" % frame)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 484, in wait_visible
    self.wait_js_func('ph_is_visible', selector)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 468, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 457, in wait_js_cond
    self.raise_cdp_exception("timeout\nwait_js_cond", cond, result["exceptionDetails"], trailer)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 241, in raise_cdp_exception
    raise Error("%s(%s): %s" % (func, arg, msg))
testlib.Error: timeout
```

test/verify/check-system-info
```diff
@@ -516,8 +516,8 @@ class TestSystemInfo(MachineCase):
         m.write("/sys/class/dmi/id/product_name", "To Be Filled By O.E.M.")
         m.write("/sys/class/dmi/id/board_vendor", "brdven")
         m.write("/sys/class/dmi/id/board_name", "brdnam")
-        b.reload()
         b.go("/system")
+        b.reload()
```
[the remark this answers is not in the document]
mvollmer: Good question, probably not. I try and maybe revert.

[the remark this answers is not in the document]
mvollmer: Yes, let's not do any any changes around reloads, since they still wait for the actual reload to have happened.  There were only two.

test/verify/check-metrics
```diff
@@ -1152,7 +1153,7 @@ class TestGrafanaClient(MachineCase):
             # expect a "Load average" panel with a sensible number
             max_load = bg.text("div:contains('Load average') .graph-legend-series:contains('1 minute') .max")
             self.assertGreater(float(max_load), 0)
-        except Error:
+        except Exception:
```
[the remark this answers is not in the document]
mvollmer: What the linters don't like is just `except:` without any type, but `except Exception` is fine, even without raise. And why wouldn't it? :-)

test/verify/check-metrics
```diff
@@ -1132,6 +1127,7 @@ class TestGrafanaClient(MachineCase):
             # Select PCP redis, HTTP URL http://10.111.112.1:44322
             redis_url = 'http://10.111.112.1:44322'
             bg.open("/datasources/new")
+            bg.wait_visible("[aria-label='Data source plugin item PCP Redis']")
```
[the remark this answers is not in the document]
mvollmer: Yep, done.

test/verify/check-metrics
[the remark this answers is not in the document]
martinpitt: They probably accept it because it gets re-thrown. without an unconditional `raise` this should never be done, as it's the exact same blanket exception as a bare `except:` (as *every* exception is derived from class `Exception`).

Anyway, this was just thinking aloud.

test/common/firefox-cdp-driver.js
```diff
@@ -47,13 +47,29 @@ function fatal() {
...
 }
 
-function fail(err) {
+// We keep sequence numbers so that we never get the protocol out of
+// synch with re-ordered or duplicate replies.  This only matters for
+// duplicate replies due to destroyed contexts, but that is already so
+// hairy that this big hammer seems necessary.
+
+var cur_cmd_seq = 0;
+var next_reply_seq = 1;
```
martinpitt, line 56: Could these both start at 0 or 1, and both use the same increment? Right now, `cur_cmd_seq` uses pre-increment and `next_reply_seq` post-increment; but they both track the same thing, so let's please make this consistent.

test/common/firefox-cdp-driver.js
```diff
@@ -262,52 +261,52 @@ function setupFrameTracking(client) {
...
+        // we would send the reply here, and would also send the
+        // error. This would drive the protocol out of synch. Also, our driver
+        // might immediately send more commands after seeing the first reply,
+        // and the unwanted second reply might be triggered in the middle of one
+        // of the next commands.  To reliably suppress the second reply we have
+        // the pretty general sequence number checks.
+        //
+        if (info.executionContextId == currentExecId) {
+            currentExecId = null;
+            fail(cur_cmd_seq, { "response": { "message": "Execution context was destroyed." } });
```
martinpitt, line 282: I'm sure you thought this through, I just find it rather hard to understand. This is an event handler that tells you "an execution context got destroyed", which happens all the time. Raising a test failure here, plus this `currentExecId` is quite a big layering violation and hard to keep track of in one's head. See below for a proposal.

test/common/firefox-cdp-driver.js
```diff
@@ -373,37 +374,34 @@ CDP(options)
...
                     // HACKS: See description of related functions
                     if (command.startsWith("client.Page.addScriptToEvaluateOnNewDocument"))
                         command = command.substring(12);
 
                     // run the command
+                    let seq = ++cur_cmd_seq;
                     eval(command).then(reply => {
+                        currentExecId = null;
                         if (unhandledExceptions.length === 0) {
-                            success(reply);
```
martinpitt, line 389: So if I understood the  above "Firefox does not report an error when the execution context.. got destroyed" correctly: Could we instead move the hack here? Then the `currentExecId` hack would at least stay localized to this command loop.

Instead, there could be another global variable, a set `validContexts` which gets updated in `executionContext{Created,Destroyed}`, in addition to `frameIdToContextId`. Then we could check for the validity of `currentExecId` here?

This would avoid the `currentExecId` global variable, which only has a well-defined meaning under the assumption that only a single command runs at a time. This seems currently true with our cdp.py, but is still error prone IMHO -- e.g. it's not unthinkable that some execution context from an earlier operation gets destroyed later on, and then `currentExecId` would be wrong.

test/verify/check-pages
```diff
@@ -178,11 +178,11 @@ OnCalendar=daily
         self.open_lang_modal()
         b.click('#display-language-modal li[data-value=de-de] button')
         b.click("#display-language-modal footer button.pf-m-primary")
-        b.expect_load()
         if b.cdp.browser.name == "chromium":
             # HACK: work around language switching in Chrome not working in current session (issue #8160)
             b.reload(ignore_cache=True)
         b.wait_visible("#content")
+        b.wait_visible("#host-apps li a:contains('Dienste')")
```
martinpitt: This does not belong into the "Allow exceptions while waiting" commit.

test/verify/check-pages
```diff
@@ -523,11 +520,11 @@ OnCalendar=daily
 
         b.click('#display-language-modal li[data-value=de-de] button')
         b.click("#display-language-modal footer button.pf-m-primary")
-        b.expect_load()
         if b.cdp.browser.name == "chromium":
             # HACK: work around language switching in Chrome not working in current session (issue #8160)
             b.reload(ignore_cache=True)
         b.wait_visible("#content")
+        b.wait_visible("#host-apps li a:contains('Dienste')")
```
martinpitt: This does not belong into the "Allow exceptions while waiting" commit.

martinpitt (changes requested):
Thanks! Sad to see all this sequence number stuff, thanks for figuring this out!

test/verify/check-pages
mvollmer: This was added just for synchronization, which later got replaced with a better version using wait_language, but I didn't clean up the history properly. I'll do that. (The actual checks for correct translations are later in this test.)

test/common/firefox-cdp-driver.js
mvollmer: > Could we instead move the hack here? 

No, I think that wont work.  The `eval.command()` promise will never settle, and the whole machinery just hangs. We need to send a reply out from the contextDestroyed event handler since that is the only place that will run.

mvollmer: If we want to support multiple concurrent commands (ugh!), we can do that by having a global seq -> contextId mapping. However, if we support concurrent commands, we need some protocol anyway to associate replies with requests (seqno maintained by cdp.py, say), and we should hook into that with the context tracking. But before have that, I'd rather not code up something complicated that is not used and will very likely need to be replaced anyway.

martinpitt: Sorry, I did not mean to suggest that we should support multiple concurrent commands. My concern is that we can't be sure that all pending contexts get destroyed before a command ends. It could be that a context destroy event happens that applies to the *previous* command, and then would fail the current one. That logic is really hard to validate.

test/common/chromium-cdp-driver.js
```diff
@@ -234,50 +216,28 @@ function setupFrameTracking(client) {
     });
 }
 
+function setupLocalFunctions(client) {
+    client.Page.reloadAndWait = (args) => {
```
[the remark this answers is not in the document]
mvollmer: > This is not [CDP Page API](https://chromedevtools.github.io/devtools-protocol/1-3/Page/). Is there a reason to shoehorn it into the chrome-remote-interface module?

Isn't that what JavaScript is all about? :-) I'll put it into client directly.

test/common/firefox-cdp-driver.js
mvollmer: It's a complicated workaround for a bug in Firefox's CDP.  But as long as it stays in the driver here, I wouldn't complain.

test/common/firefox-cdp-driver.js
martinpitt: OK, thanks; I hope we'll be able to debug that in a few months :grin:

test/common/chromium-cdp-driver.js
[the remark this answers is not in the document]
mvollmer: Done.

test/common/firefox-cdp-driver.js
mvollmer: They could, but the executionContextDestroyed would have code like `fail(next_cmd_seq-1, ...)` which is also awkward.  I have changed the code a bit to always use `... += 1` for the counters.

test/verify/check-pages
mvollmer: Removed.

test/verify/check-pages
```diff
@@ -181,8 +181,7 @@ OnCalendar=daily
         if b.cdp.browser.name == "chromium":
             # HACK: work around language switching in Chrome not working in current session (issue #8160)
             b.reload(ignore_cache=True)
-        b.wait_visible("#content")
-        b.wait_visible("#host-apps li a:contains('Dienste')")
```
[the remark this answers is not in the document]
mvollmer: Cleaned up.

test/verify/check-pages
```diff
@@ -523,8 +527,7 @@ OnCalendar=daily
...
```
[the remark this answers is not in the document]
mvollmer: Cleaned up.

test/common/firefox-cdp-driver.js
martinpitt: You didn't actually [change the code](https://github.com/cockpit-project/cockpit/compare/edec865986b2e0e37f7d34be77965185da16596b..1b521e01ba7046424ebe3e6d2db739c76f3437a2), but ok; agreed we don't want the -1 either. That's not what I had in mind, but bikeshedding.

martinpitt (approved):
Danke!

martinpitt:
[This flake](https://logs.cockpit-project.org/logs/pull-17007-20220221-123000-1b521e01-centos-8-stream/log.html) looks familiar, retrying.

Issue #16887, opened 2022-01-27
Add ability to "Force password change on first use" in account creationg dialogue


jelly:
Page: Accounts

Related to https://github.com/cockpit-project/cockpit/issues/16886 this issue adds a new checkbox `Force password change on first use` which if enabled forces the new user to change the password on login. This can be implemented using `passwd --expire`.

As this new option contradicts with the "Lock" or "Disable password access" option selecting this option should deselect/disable the other option.

PR #17010, opened 2022-02-15
users: Add Force password change on first login

tomasmatus:
When one checkbox is checked the other is disabled.

Fixes: #16887

force_passwd_change
force_passwd_change_disabledL
force_passwd_change_disabledR


pkg/users/account-create-dialog.js
```diff
@@ -250,6 +263,12 @@ export function account_create_dialog(accounts) {
                                         user_name,
                                         "--lock"
                                     ], { superuser: "require", err: "message" });
+                                if (state.change_passw_force)
+                                    return cockpit.spawn([
+                                        "/usr/bin/passwd",
+                                        "-e",
+                                        state.user_name
```
marusak, line 270: The code updated not to use these global objects, please don't use `state.user_name`.

marusak (changes requested):
Thank you!

It will need tests - both that they disable each other as well as that the account is really requiring a new password.

Also @garrett WDYT design wise that they are on the same line? We usually put then underneath each other, but here you can have zero or maximum one of them (so they are exclusive).

garrett:
> Also @garrett WDYT design wise that they are on the same line? We usually put then underneath each other, but here you can have zero or maximum one of them (so they are exclusive).

I think they should go on separate lines, not next to each other. They're not really related, except that they're mutually exclusive.

What happens when one is checked and you select the other? Does it uncheck and disable it? Should we even disable it? (You can select the other and it would uncheck the first one.)

Something more understandable might be:

Change the label to "Authentication"

Then the options could be:

(\_) Use a password
       [\_] Require password change on login
(*) Disallow password authentication (?)

Something like this would probably have less of a surprise. If the disallow radio is selected and someone clicks on the checkbox for the require password change, it'd select the "Use a password" radio option.

It's in this order (instead of radios first with the check option below the later one) because I'd expect more people are using passwords for authentication. As it's more common, it's first.

WDYT?

garrett (dismissed):
Just to be clear: I'm waiting on the suggested changes, mainly to reform the option from two checkboxes to radio buttons with a checkbox under one of them before I review again.

tomasmatus:
> Just to be clear: I'm waiting on the suggested changes, mainly to reform the option from two checkboxes to radio buttons with a checkbox under one of them before I review again.

Yes, I started working on this but I was mostly busy with my other PR.

garrett:
> I was mostly busy with my other PR

I just wanted to make sure you didn't think I was blocking you on this. I'm juggling so much with everyone and I want it to be clear when I'm blocking something (and also when I'm not). :grin:

tomasmatus:
Redesigned to use radios as discussed above.
Updated screenshots in the first comment.

marusak:
Idea - now we require password even when selecting `Require password change on first login`. Does that make sense? Will the password be ever used?

If not, then something like this maybe?
Screenshot from 2022-03-07 10-23-24

and hiding it:
Screenshot from 2022-03-07 10-23-44

I just moved `PasswordFormFields` into `body` of `Checkbox`:
```
                        description={
                            <Checkbox id="accounts-create-force-password-change"
                                      label={_("Require password change on first login")}
+                                     body={change_passw_force ? null : <PasswordFormFields password_label={_("Password")}
+                                            password_confirm_label={_("Confirm")}
+                                            error_password={errors && errors.password}
+                                            error_password_confirm={errors && errors.password_confirm}
+                                            idPrefix="accounts-create-password"
+                                            change={change} /> }
                                      isChecked={change_passw_force} onChange={checked => change("change_passw_force", checked)} />

```

It would need some styling and some logic about not checking and setting of the password in this case.
WDYT? But also possible follow up?

@tomasmatus also coreos needs some special case it seems as the test is failing. Are you able to reproduce this locally?

tomasmatus:
> Does that make sense? Will the password be ever used?

Yes, it makes sense. The password set here is used for first login. Then before the user is allowed to access cockpit he's required to change his password.
I think what you wanted to say is that we shouldn't ask for password when `Disallow password authentication` is checked.

> coreos

I'm going to look at it right now.

garrett:
> If not, then something like this maybe?
> Screenshot from 2022-03-07 10-23-24

:facepalm: I totally overlooked the password box above; I was only looking at the options when considering this.

The UI isn't right here, but I guess we should change things around regardless... but not quite like this.

garrett:
If we did the UI nested like this, then we should have the labels above the entries for the sub-elements. And require password change should go under the password entry. (And the options are back to being on the same line. :grin:)

force-password

Additionally, we could consider inlining the messaging like this, so it shows up as a message under the selection instead of relying on a popover:

force-password-disallow

Source:
https://excalidraw.com/#room=3d069c94243d59305eb9,mzWbvYOtAV8d7BVGeOsr6Q

**[🠟 force-password-disallow.excalidraw.zip](https://github.com/cockpit-project/cockpit/files/8198951/force-password-disallow.excalidraw.zip)**

jelly (changes requested):

marusak:
Rebased to get in https://github.com/cockpit-project/cockpit/commit/9877d5163352fe2a0a72859c1b80ad7d37ff5fa3 to make c8s tests working.

pkg/users/users.scss
```diff
@@ -133,6 +133,10 @@
     }
 }
 
+#accounts-create-dialog .pf-c-check {
+    margin-bottom: var(--pf-global--spacer--xs);
+}
+
```
marusak, line 139: So I am not huge fan of this. I tried removing it and even experimented with using `body` property instead of `description`:
Screenshot from 2022-03-10 12-34-57

but the spacing on top is too much and on bottom too little :man_shrugging: 

I think we can just use utilities then - pushing changes

marusak (approved):
Nice, thank you!

PR #17012, opened 2022-02-15
Show when error happens on services page

marusak:
Screenshot from 2022-02-16 18-12-44

### DIFF
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -158,6 +158,7 @@ class ServicesPageBody extends React.Component {
             isFullyLoaded: false,
 
             error: null,
+            currentStatus: null,
         };
 
         this.onCurrentTextFilterChanged = (currentTextFilter) => {
@@ -373,7 +374,7 @@ class ServicesPageBody extends React.Component {
             return this.listFailedUnits();
 
         // Reinitialize the state variables for the units
-        this.setState({ loadingUnits: true });
+        this.setState({ loadingUnits: true, currentStatus: _("Listing units") });
 
         this.seenPaths = new Set();
 
@@ -381,6 +382,7 @@ class ServicesPageBody extends React.Component {
 
         // Run ListUnits before LIstUnitFiles so that we avoid the extra LoadUnit calls
         // Now we call LoadUnit only for those that ListUnits didn't tell us about
+
         systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, "ListUnits", null)
                 .then(([results]) => {
                     results.forEach(result => {
@@ -404,6 +406,7 @@ class ServicesPageBody extends React.Component {
                         );
                     });
 
+                    this.setState({ currentStatus: _("Listing unit files") });
                     systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, "ListUnitFiles", null)
                             .then(([results]) => {
                                 results.forEach(result => {
@@ -436,7 +439,7 @@ class ServicesPageBody extends React.Component {
                                         this.seenPaths.add(unit_path);
 
                                         return this.getUnitByPath(unit_path);
-                                    }, ex => this.setState({ error: cockpit.format(_("Listing of unit failed: $0"), ex.toString()), loadingUnits: false })));
+                                    }, ex => this.setState({ error: cockpit.format(_("Loading unit failed: $0"), ex.toString()), loadingUnits: false })));
                                 });
 
                                 Promise.all(promisesLoad)
@@ -724,7 +727,7 @@ class ServicesPageBody extends React.Component {
         if (this.state.error)
             return <EmptyStatePanel title={_("Loading of units failed")} icon={ExclamationCircleIcon} paragraph={this.state.error} />;
         if (!this.state.isFullyLoaded)
-            return <EmptyStatePanel loading title={_("Loading...")} />;
+            return <EmptyStatePanel loading title={_("Loading...")} paragraph={this.state.currentStatus} />;
 
         /* Perform navigation */
         if (path.length == 1) {
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -156,6 +156,8 @@ class ServicesPageBody extends React.Component {
             loadingUnits: false,
             path: cockpit.location.path,
             isFullyLoaded: false,
+
+            error: null,
         };
 
         this.onCurrentTextFilterChanged = (currentTextFilter) => {
@@ -434,7 +436,7 @@ class ServicesPageBody extends React.Component {
                                         this.seenPaths.add(unit_path);
 
                                         return this.getUnitByPath(unit_path);
-                                    }, ex => console.warn(ex)));
+                                    }, ex => this.setState({ error: cockpit.format(_("Listing of unit failed: $0"), ex.toString()), loadingUnits: false })));
                                 });
 
                                 Promise.all(promisesLoad)
@@ -463,8 +465,8 @@ class ServicesPageBody extends React.Component {
                                             this.setState(newState);
                                             this.processFailedUnits();
                                         });
-                            }, ex => console.warn('ListUnitFiles failed: ', ex.toString()));
-                }, ex => console.warn('ListUnits failed: ', ex.toString()));
+                            }, ex => this.setState({ error: cockpit.format(_("Listing unit files failed: $0"), ex.toString()), loadingUnits: false }));
+                }, ex => this.setState({ error: cockpit.format(_("Listing units failed: $0"), ex.toString()), loadingUnits: false }));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pkg/systemd/services/services.jsx
```diff
@@ -719,6 +721,8 @@ class ServicesPageBody extends React.Component {
         const { unit_by_path } = this.state;
         const path = this.props.path;
 
+        if (this.state.error)
+            return <EmptyStatePanel title={_("Loading of units has failed")} icon={ExclamationCircleIcon} paragraph={this.state.error} />;
```
martinpitt, line 725: maybe "has failed" → "failed" for consistency?

test/verify/check-system-services
```diff
@@ -34,6 +34,7 @@ class TestServices(MachineCase):
         # when we add additional services to this target below
         # we don't race with the boot process
         self.machine.execute("systemctl start default.target")
+        self.machine.execute("systemctl --user start default.target")
```
martinpitt, line 37: I'm quite surprised that this is necessary. Do we have some evidence that it is not already running after logging into cockpit? Unlike the system instance, this should do pretty much nothing. I'm a bit afraid of hiding errors here if it's not at least in "activating". And I suppose we should change the other command from an explicit "start" (which hides bugs) to a waiting loop on `is-active`.

test/verify/check-system-services
```diff
@@ -34,6 +34,7 @@ class TestServices(MachineCase):
         # when we add additional services to this target below
         # we don't race with the boot process
         self.machine.execute("systemctl start default.target")
+        self.machine.execute("systemctl --user is-active default.target")
```
martinpitt: This is good, but let's please give it a proper commit message.

test/verify/check-system-services
```diff
@@ -34,7 +34,6 @@ class TestServices(MachineCase):
         # when we add additional services to this target below
         # we don't race with the boot process
         self.machine.execute("systemctl start default.target")
-        self.machine.execute("systemctl --user is-active default.target")
```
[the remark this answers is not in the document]
marusak: Bad rebase, I tried to drop that commit. The issue is that at this stage there is no user session - which also points to those bogus `systemctl --user` calls. But I'll check the system one to do waiting instead of starting.

martinpitt:
The commits look nice now, thanks! However, it seems the [testRelationshipsUser](https://logs.cockpit-project.org/logs/pull-17012-20220217-194521-4af3a3de-centos-8-stream/log.html#301) failure on c8s still exists, with the never-endling "loading" (no [js log](https://logs.cockpit-project.org/logs/pull-17012-20220217-194521-4af3a3de-centos-8-stream/TestServices-testRelationshipsUser-centos-8-stream-127.0.0.2-2501-FAIL.js.log) errors either), dn this now breaks all `TestServices.testBasic{,Session}` on the [testing farm](http://artifacts.dev.testing-farm.io/9bb6d1f4-b8c3-49ae-872d-2bf0beed4f95/).

marusak:
Wait, what? [here](https://logs.cockpit-project.org/logs/pull-17012-20220218-063639-c029d1f1-rhel-8-6/TestServices-testRelationshipsUser-rhel-8-6-127.0.0.2-2301-FAIL.png) it only shows `Loading...` but now I added there to always have some details what is loading. So there is some different path as well that I missed that happens when this issue occurs.

marusak:
@martinpitt I think I need to pick up your brains here.

The issue why it sometimes is stuck on `Loading...` page is due to [Subcribe](https://github.com/cockpit-project/cockpit/blob/main/pkg/systemd/services/services.jsx#L233) (translates to `busctl call --user org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager Subscribe`) being stuck. It never resolves nor fails in cockpit. Calling it on cmdline it resolves with `Connection timed out`. Right now I see it happening mostly on [centos-8-stream](https://logs.cockpit-project.org/logs/pull-17012-20220221-141343-38f61066-centos-8-stream/log.html) (around every second run). I think I might have seen it on other OSes as well, but right now it predominantly affects centos-8.

In journal I can see `dbus-daemon[32654]: avc:  denied  { send_msg } for msgtype=method_return dest=:1.10 spid=32628 tpid=53688 scontext=sysadm_u:sysadm_r:sysadm_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=dbus permissive=0` which seems to correspond to callings on `Subscribe`. I added `busctl` call when test is stuck on `Loading...` and in that case there are two cases of this avc. Might be red herring, but suspicious as some readings on the internet suggest that avcs similar to this cause very long boots. E.g. https://bugzilla.redhat.com/show_bug.cgi?id=1647920

When I just run centos-8-stream tests on my system then there is no AVC and the `busctl` works just fine.

Notice, that I cannot reproduce this locally and CI also hits this only every now and then. So I am not sure how this AVC works (if that is the issue anyway) and I cannot get much data easily, it is just about putting something into tests and running tests with `tests-trigger -f` until it hits it.

Do you have some ideas about this? Or should I just report BZ and create naughty - but the pattern would be rather generic and might hide other issues. Possibly dump journal when it is stuck on Loading and write pattern to match this avc?

marusak:
@martinpitt had an idea that `test/verify/check-static-login:        m.execute("semanage login -a -s sysadm_u admin"` might be causing this and indeed now it is easily reproducible:
```
TEST_OS=centos-8-stream ./test/common/run-tests --test-dir test/verify TestLogin.testSELinuxRestrictedUser  TestServices.testRelationshipsUser

```

martinpitt:
Yay, glad to hear that my hunch was right. I propose to not change the `admin` user in that test, but create a new one, add it to the `wheel` group, and change it to `sysadm_u` there. Then the test can make fewer assumptions about working cleanup.

test/verify/check-system-services
```diff
@@ -33,7 +33,7 @@ class TestServices(MachineCase):
         # Make sure the system finishes "booting" so that
         # when we add additional services to this target below
         # we don't race with the boot process
-        self.machine.execute("systemctl start default.target")
+        self.machine.execute("while ! systemctl is-active default.target; do sleep 1; done")
```
martinpitt, line 37: BTW, the commit message "Actually wait until default.target is active" is a bit misleading. `systemctl start` *does* wait for default.target to become active. The "problem" is that it also starts it if it wasn't activating yet. That's the bit that we don't want, as it hides bugs.

marusak: I changed the message

pkg/systemd/services/services.jsx
```diff
@@ -439,6 +442,14 @@ class ServicesPageBody extends React.Component {
                                     }, ex => this.setState({ error: cockpit.format(_("Listing of unit failed: $0"), ex.toString()), loadingUnits: false })));
                                 });
 
+                                let counter = 0;
+                                this.setState({ currentStatus: cockpit.format(_("Loading unit $0/$1"), counter, promisesLoad.length) });
+                                for (const p of promisesLoad) {
+                                    counter++;
+                                    p.then(() => this.setState({ currentStatus: cockpit.format(_("Loading unit $0/$1"), counter, promisesLoad.length) }))
```
martinpitt, line 452: The services page already does too much on startup (we really shouldn't be doing hundreds of D-Bus calls!), and now this does hundreds of gettext calls and state updates in addition. This should at least use `throttle-debounce` or some simple `if (counter % 10 == 0)` or similar.

pkg/systemd/services/services.jsx
```diff
@@ -439,6 +442,14 @@ class ServicesPageBody extends React.Component {
...
+                                            .catch(ex => this.setState({ error: cockpit.format(_("Loading unit failed: $0"), ex.toString()), loadingUnits: false }));
```
martinpitt, line 453: If a single unit fails to load, we will never see this. The next successful unit load will immediately overwrite the state again. This makes errors much harder to debug, it'd be better to log the failures (in addition/instead).

martinpitt (changes requested):
Dakujem! Most of this looks good, I am just not happy about the new progress commit. Possibly postpone this to a follow-up, as the rest is really desirable?

pkg/systemd/services/services.jsx
marusak: But rendering first checks if `error` is set. So once a single unit fails, we will switch from `Loading...` to `Loading unit failed` state. Sure, it will continue to load rest of them, but the page will stay on error.

pkg/systemd/services/services.jsx
marusak: ok, I can drop this. It was really just me playing around seeing if it gets stuck on some specific unit

pkg/systemd/services/services.jsx
martinpitt: I see, makes sense. However, could we still please log this in addition? That's much easier to see both from test logs and from user's JS console than having them to do screenshots and such.

marusak: but indeed, this is already handled by the content of `promisesLoad`.

martinpitt (approved):
Cheers!

PR #17015, opened 2022-02-16
login: Check selector support & :is/:where

### LINKED ISSUE #16990: login: Block on lack of :where() and :is() support.
Browsers have added support for `:is()` and `:where()` a bit over 2 (Firefox, Safari) to 1 (Chrome, et al) years ago. PatternFly, in the latest release 4.171.1 started using `:where()` (in 2 places).

As using `:is()` or `:where()` _will_ definitely break the UI where it is used, we need to block the old, out of date, insecure browsers that do not support this CSS.

This will need to be added as a supports check on the login page — and the website needs to check for this too, with its browser compatibility pass/fail and browser version estimates.

garrett:
Fixes #16990


garrett:
Currently a WIP. I need to test it across browsers... but it's getting late.

garrett:
I tested this in Fedora 32 and Fedora 33 (both very old versions) to get an idea of what it'd take for our users to not be able to sign in to Cockpit anymore. They'd have to be running unpatched, years old distros.

Here's Chromium 85 from Fedora 33 (first version shipped in F33):

Screenshot from 2022-02-17 16-50-42

I tried Chromium 80, but we have a replaceAll check that comes first, so I can't check for select(), but that's OK anyway, as proved above, select() can work before the is/where check.

Screenshot from 2022-02-17 16-54-37

---

Current browsers (to demonstrate nothing is broken):

Here's a screenshot of updated Chrome (no issues):

Screenshot from 2022-02-17 16-54-57

WebKit:

Screenshot from 2022-02-17 16-55-36

Firefox:

Screenshot from 2022-02-17 16-55-45

garrett:
This blocks upgrading PatternFly @ https://github.com/cockpit-project/cockpit/pull/16967

garrett:
Added description to the commit and rebased on current main.

martinpitt (approved):
Many thanks for the super thorough checking!

garrett:
@haomi123: What browser are you using? 

No browser should be so old that it shouldn't pass our feature tests. If your browser cannot pass our feature tests, then it's nearly a year and a half or older at this point and wildly insecure. You _need_ to upgrade your browser.

Here's support for the feature check: https://caniuse.com/mdn-css_selectors_where:

- Firefox: June 2020 (1 year 11 months ago)
- Edge: January 2021 (1 year 4 months 7 days ago)
- Chrome: January 2021 (1 year 4 months 11 days ago)
- Safari : September 2020 (1 year 8 months 14 days ago)

It's important to note that Cockpit does not do browser checking, nor does it exclude any browsers specifically, either by browser agent string or version. Cockpit checks browser support for various things that are used within Cockpit. If your browser cannot pass the tests, then it cannot use Cockpit properly.

Cockpit only ever updates the browser check after all users have been given ample time to upgrade their browsers. Generally, we don't add anything new until years have passed.

There have been _several_ critical security bugs for all browsers that have been fixed in the past year. If your browser doesn't have support for `:is()` and `:where()`, then you have a giant security problem.

Examples of recent security issues that have been fixed in just the past few months:

- Chrome: https://www.forbes.com/sites/gordonkelly/2022/05/25/google-chrome-security-new-vulnerabilities-reported-update-chrome-now/
- Firefox: https://www.forbes.com/sites/daveywinder/2022/05/22/firefox-browser-hacked-in-8-seconds-using-2-critical-security-flaws/
- Safari: https://vpnoverview.com/news/apple-safari-browser-vulnerability-exploited-in-the-wild/
- Brave is affected by the Chrome CVEs and has its own security issues too: https://www.opencve.io/cve?vendor=brave

The Chrome CVEs also affect all Chrome-based browsers, such as Brave and Edge.

All browsers are very complex and all have CVEs that are constantly found and fixed. You really want to make sure you have the latest browser version (and browser rendering engine, of which are just 3: WebKit (Safari, GNOME Web), Servo (Firefox), and Chromium (Chrome, Edge, Brave, and all the rest)).

haomi123:
I use cockpit-264.1 and Firefox79，There was an error：
image
but firefox79 can pass  feature test,
image

When I use Edge 95,cockpit can work properly.
image

garrett:
Why are you using such out of date versions of both Firefox and Edge?

[Firefox 79](https://www.mozilla.org/en-US/firefox/79.0/releasenotes/) is almost 2 years old... released on July 28, 2020. While it is weird that it's shown as supporting `:is()` and `:where()` in that version, even though it apparently doesn't. However, it's nearly 2 years old and you should've upgraded long ago, and I can confirm that Firefox has supported both `:is()` and `:where()` for a very long time. (Firefox is my main browser and I do development work for Cockpit in Firefox, while checking in other browsers afterward.)

Like every other browser available, [Firefox has had a lot of big security problems](https://www.mozilla.org/en-US/security/known-vulnerabilities/firefox/) in the browser version that you're running — that have since been fixed, thankfully.

And Microsoft Edge 95 came out 9 months ago. While it's new enough to work, there have been [dozens of high-profile security bugs that have been patched since](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-relnotes-security).

 If you don't want to upgrade as quickly, there is the [Firefox Extended Support Release](https://www.mozilla.org/en-US/firefox/enterprise/) which has a much slower 6 month cycle. It's mainly intended for enterprises, but works fine everywhere. It's at version 91.9.1 currently (wheras [the standard release of Firefox is at 100.0.2](https://www.mozilla.org/en-US/firefox/100.0.2/releasenotes/)). Both editions get security updates.

Edge has an "extended stable" channel that doubles the release cycle window from every 4 weeks to every 8 weeks. It's much shorter than Firefox's every six months cycle. https://docs.microsoft.com/en-us/deployedge/microsoft-edge-channels But it does mean you'd have to update a tiny bit less and still have security features.

haomi123:
@garrett  So I can't fix this by modifying the cockpit code ，t he best way to solve this problem is to upgrade the browser.

garrett:
@haomi123: This change was due to PatternFly starting to use `:is()` and `:where()`. As Cockpit depend son PatternFly, we inherited that change too. And due to the nature of how `:is()` and `:where()` work, if they're used in CSS and a browser doesn't support them, then the browser _will_ have visual glitches, which could even mean functionality might be broken in several places.

Therefore, I made the decision to check for `:is()` and `:where()` at the login screen, as all of the current browsers at the time of writing this code had supported it for well over a year.

In other words: Even if you bypass the check manually by modifying files on your installation, that means that your browser still will not render Cockpit correctly if it does lack support for `:is()` and `:where()`.

In other other words: Just upgrade your browsers. The security issues for your older browser versions (especially on Windows) are extremely scary ones, so you should upgrade your browsers for other reasons too.

haomi123:
Thanks for your reply.

PR #17019, opened 2022-02-17
package.json: eslint plugins cleanup

KKoukiou:
- Blocked on #17190


martinpitt:
This fell through the cracks, but this is some nice cleanup. I adopt this PR. Thanks @KKoukiou!

martinpitt:
Failures are due to #17190, we need to land that ASAP.

martinpitt:
As this affects most pages and thus has hundreds of affected retries, it runs into all kinds of weird general flakes now. [This one](https://logs.cockpit-project.org/logs/pull-17019-20220329-063749-0866f573-fedora-coreos/log.html#135-2) might be something which we should fix in general, though -- these "infrastructure retries" should not count in the "affected" logic.

martinpitt (approved):
Thanks!

PR #17017, opened 2022-02-17
packit: Start testing on Fedora 36

martinpitt:
Corresponding to cockpituous release targets in commit c9eadef1a0

----

Yesterday's [fedora-36 release](https://bodhi.fedoraproject.org/updates/FEDORA-2022-7b7b2119dc) [failed several tests](https://osci-jenkins-1.ci.fedoraproject.org/job/fedora-ci/job/dist-git-pipeline/job/master/112990/testReport/(root)/tests/_test_verify/). I waived it, let's walk through them here.

 - Builds on top of PR #17021
 - Needs https://github.com/cockpit-project/bots/pull/2956
 - Needs https://github.com/cockpit-project/bots/pull/2970

PR #17021, opened 2022-02-17
fmf fixes

martinpitt:
This will fix the [fedora 35 failure](http://artifacts.dev.testing-farm.io/bf9e81a9-e3f7-4ceb-89ba-4ffe8de5dd8e/) which started yesterday or today. It will also bring us artifacts, finally!

Broken out from PR #17017, adding F36 will take some more work.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -14,7 +14,6 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      - fedora-34
       - fedora-35
       - centos-stream-8-x86_64
       - centos-stream-9-x86_64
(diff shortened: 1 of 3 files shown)


martinpitt:
See [here](https://github.com/cockpit-project/cockpit/pull/17017#issuecomment-1042645974) for a more detailed analysis of the new test skip

martinpitt:
[rawide i386 build](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-17021/fedora-rawhide-i386/03525030-cockpit/builder-live.log.gz) is a broken koji, let's ignore here.

PR #17022, opened 2022-02-17
test: workaround purge not cleaning up pmlogger requires

jelly:
When purging pcp on debian testing with 5.3.6 we are left with a broken
dependency so pmlogger does not start:
/etc/systemd/system/pmlogger.service.requires/pmlogger_farm.service:
broken symbolic link to /lib/systemd/system/pmlogger_farm.service

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986074


martinpitt (approved):
Thanks!

PR #17023, opened 2022-02-17
test: Fix Firefox CDP driver

marusak:
Firefox 96 enables Fission [1] by default to improve site isolation.
This broke DevTools Protocol's `Page` events [2]. Disable Fission to
work with current nightlies as well.

[1] https://wiki.mozilla.org/Project_Fission
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1746154


martinpitt (approved):
Thanks! Triggering /firefox would have been enough :grin:

marusak:
> Triggering /firefox would have been enough 

I was so excited I could not constrain myself to think about labels :smile:

PR #17030, opened 2022-02-22
users: Call `date` instead of using Date() to parse `lastlog` output

martinpitt:
Date() has trouble parsing some dates/times; e.g. Firefox 91 and 97:

    > new Date('Jan 4 11:52:37 -0500 2022')
    Invalid Date

    > new Date('Jan 4 11:52:37 +0100 2022')
    Date Tue Jan 04 2022 11:52:37 GMT+0100 (Central European Standard Time)

The Date documentation [1] also explicitly discourages using it for
parsing dates if they are not in ISO format (which the above is very
much not).

As `lastlog` has no way of outputting timestamps (or machine readable
output, for that matter), call `date` to convert the date/time string
into a time stamp. However, still keep the JavaScript dissection of the
lastlog string, as cutting out the string does not work at all with
`cut` and `sed`, and is way too complicated with `awk` even.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

### DIFF
--- a/pkg/users/account-details.js
+++ b/pkg/users/account-details.js
@@ -72,17 +72,17 @@ function get_last_login(name) {
 
         // line looks like this: admin            web cons ::ffff:172.27.0. Tue Mar 23 14:49:04 +0000 2021
         // or like this:         admin            web cons ::ffff:172.27.0. Thu Apr  1 08:58:51 +0000 2021
+        // this is impossible to parse with Date() (e.g. Firefox does not work with all time zones), so call `date` to parse it
         const date_fields = line.split(/ +/).slice(-5);
-        const d = new Date(date_fields.join(' '));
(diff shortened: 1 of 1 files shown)


martinpitt:
This got caught during [RHEL 9 gating](https://src.osci.redhat.com/rpms/cockpit/pull-request/97) (sorry, RH internal link only), that PR has lots of debugging details. This has been broken for a looooong time, but for some reason it only surfaced in integration tests now. I don't know the precise pattern which strings Firefox can parse and which it can't. But I tested all the ones I found with `date`, and that works much more robustly.

marusak (approved):
ugh, that is nasty from firefox. Thanks!

PR #17031, opened 2022-02-22
test: Reinstate offline builds for rpms/debs

martinpitt:
This ensures that we stay independent from any changes that go on in
online repositories, and our mock/pbuilders are sufficiently pre-cached.

This got lost in the move to image-customize in commit e060058027fecc83.


martinpitt:
Ah, [fcos also requires networking](https://logs.cockpit-project.org/logs/pull-17031-20220222-135822-0a8bc7f6-fedora-coreos/log.html) at the moment. I'll check if that can be avoided, but in the worst case I'll re-enable networking. (We didn't run offline in the past)

jelly (approved):

PR #17035, opened 2022-02-23
abrt: Only navigate once problem is deleted

marusak:
While this should be pretty much immediate, every now and then bots are
too fast to navigate back to this item before the problem has been
deleted resulting in still seeing the report.

Note: I am not super sure if this waits for the report to be deleted or not. If it does not, then I will add direct wait into test.

Example: https://logs.cockpit-project.org/logs/pull-17012-20220223-065011-732ceefd-fedora-35/log.html#232

### DIFF
--- a/pkg/systemd/abrtLog.jsx
+++ b/pkg/systemd/abrtLog.jsx
@@ -225,8 +225,7 @@ export class AbrtLogDetails extends React.Component {
     }
 
     onDelete(ev) {
-        this.props.service.DeleteProblems([this.props.problem.path]);
-        this.props.reloadProblems(ev);
+        this.props.service.DeleteProblems([this.props.problem.path]).then(() => this.props.reloadProblems(ev));
     }
 
     renderBacktrace(val) {


martinpitt (approved):
Even if that's not sufficient for the tests (possibly because DeleteProblems() is async), it's still a good change IMHO.

I saw [an unrelated flake](https://logs.cockpit-project.org/logs/pull-17035-20220223-084238-42d9c17e-centos-8-stream/log.html#80-2), that would need a retry.

martinpitt:
Ugh, what a [horrible React spam](https://logs.cockpit-project.org/logs/pull-17035-20220223-084238-42d9c17e-fedora-35-devel/log.html#252) about duplicate keys. Unrelated, but perhaps you could look into this while you are in that part of the code?

marusak:
> Ugh, what a [horrible React spam](https://logs.cockpit-project.org/logs/pull-17035-20220223-084238-42d9c17e-fedora-35-devel/log.html#252) about duplicate keys. Unrelated, but perhaps you could look into this while you are in that part of the code?

Yes, I'll fix it here as well. And might as well get into more abrt issues that I will see while running tests in this PR.

martinpitt (approved):
second commit> OMG, how did that *ever* work? :cry:

marusak:
> how did that ever work?

not correctly :man_shrugging:

PR #17036, opened 2022-02-23
networking: Fix return value of CheckpointCreate method - FIX COMMIT MESSAGE ON MERGING

mvollmer:
The call_object_method function used to pass all return values as
individual arguments to the "then" function, but this got accidentally
changed in e4093f5e6ab3125de96aca0db403313a5962ccde to pass them all
in an array.

Thus, checkpoint_create would return an array containing the
checkpoint object path instead of the object path directly.  When this
got passed to checkpoint_destroy later on, this would fail with an
"invalid arguments" error.  Cockpit doesn't look closely at the exact
error, interpretes a failed checkpoint_destroy call always as "we got
disconnected", assumes that the checkpoint has been rolled back, and
brings up the "Connection will be lost" dialog.

This was missed by the tests because we switch off checkpoints during
those tests that dont expect them.

This is fixed here just for CheckpointCreate.  Other calls to
call_object_method need to be checked and corrected accordingly as
well.

### DIFF
--- a/pkg/networkmanager/interfaces.js
+++ b/pkg/networkmanager/interfaces.js
@@ -1169,6 +1169,7 @@ export function NetworkManagerModel() {
                                           devices.map(objpath),
                                           timeout,
                                           0)
+                        .then(results => results[0])
                         .catch(function (error) {
                             if (error.name != "org.freedesktop.DBus.Error.UnknownMethod")
                                 console.warn(error.message || error);
--- a/test/verify/check-networkmanager-checkpoints
+++ b/test/verify/check-networkmanager-checkpoints
@@ -134,6 +134,44 @@ class TestNetworkingCheckpoints(NetworkCase):
             b.click("#confirm-breaking-change-popup button:contains('Keep connection')")
             b.wait_not_present("#confirm-breaking-change-popup")
 
+    def testNoRollback(self):
+        b = self.browser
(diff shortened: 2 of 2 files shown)


mvollmer:
I'll add a test that will catch this.  We had switched off checkpoints because they inherently rely on real time timeouts, and those cause flakes in our CI runs. But let's see if we can get one test to reliably use checkpoints without the need for rollback.

mvollmer:
Next I'll go over all other calls to call_object_method and make sure they use return values correctly, in a separate PR. I don't think there are any additional bugs from what I have seen so far, but let's see.

martinpitt (approved):
Heh, the test is really crude (sifting through logs like that), but it seems to work everywhere, and at least if it fails, it fails in the "right" way for a test.

If you push again, please add references to https://bugzilla.redhat.com/show_bug.cgi?id=2056362 and https://bugzilla.redhat.com/show_bug.cgi?id=2056386 to the commit message; otherwise we can do it with squash&merge.

Thank you!

martinpitt:
[Annoying flake](https://logs.cockpit-project.org/logs/pull-17036-20220223-141216-a0d0af14-ubuntu-2004/log.html#69-1), but this is unrelated (I'm fairly sure), just due to the 3x affected retry.

Issue #16945, opened 2022-02-08
metrics: CPU and Memory services show user services but link to system services (which do not exist if it's just a user service)


garrett:
### Explain what happens

1. Visit the metrics page.
2. Have a user service running. On Fedora, that can be `user@1000`, `org.gnome.Shell@wayland.service`, `pipewire-pulse`, etc.
3. Click on the user service.
4. See a "Unit not found" page when the services page tries to load the system version of the unit (which in most cases won't exist... but even if it does also exist, it would be unrelated).

This is probably an issue with the way links are constructed on the metrics page.

What I get:
https://localhost:9090/system/services#/pipewire-pulse.service

What it should be:
https://localhost:9090/system/services#/pipewire-pulse.service?owner=user

(The `owner=user` is weird; it really should be `/username/pipewire-pulse.service`. But that's another issue.)

### Version of Cockpit

main

### Where is the problem in Cockpit?

Metrics

### Server operating system

Fedora

### Server operating system version

35

### What browsers are you using?

Firefox

martinpitt:
That sounds very close to what @tomasmatus fixed the other day in PR #16889, except that was for timers. @tomasmatus , would that be something for you  as a finger exercise?

tomasmatus:
> That sounds very close to what @tomasmatus fixed the other day in PR #16889, except that was for timers. @tomasmatus , would that be something for you as a finger exercise?

Yes, I can have a look!

PR #17039, opened 2022-02-23
metrics: Fix link construction for user services

tomasmatus:
Fixes: #16945

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -379,24 +379,31 @@ class CurrentMetrics extends React.Component {
             }
         }
 
-        // return [ { [key, value] } ] list of the biggest n values
+        // return [ { [key, value, is_user] } ] list of the biggest n values
         function n_biggest(names, values, n) {
             const merged = [];
             names.forEach((k, i) => {
                 const v = values[i];
                 // filter out invalid values, the empty (root) cgroup, non-services
                 if (k.endsWith('.service') && typeof v === 'number' && v != 0) {
+                    const is_user = k.match(/^user.*user@\d+\.service.+/);
                     const label = k.replace(/.*\//, '').replace(/\.service$/, '');
                     // only keep cgroup basenames, and drop redundant .service suffix
-                    merged.push([label, v]);
+                    merged.push([label, v, is_user]);
                 }
             });
             merged.sort((a, b) => b[1] - a[1]);
             return merged.slice(0, n);
         }
 
-        function serviceRow(name, value) {
-            const name_text = <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service") }><TableText wrapModifier="truncate">{name}</TableText></Button>;
+        function serviceRow(name, value, is_user) {
+            const name_text = (
+                <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : "")) }>
+                    <TableText wrapModifier="truncate">
+                        {name}
+                    </TableText>
+                </Button>
+            );
             const value_text = <TableText wrapModifier="nowrap">{value}</TableText>;
             return {
                 cells: [{ title: name_text }, { title: value_text }]
@@ -405,11 +412,11 @@ class CurrentMetrics extends React.Component {
 
         // top 5 CPU and memory consuming systemd units
         newState.topServicesCPU = n_biggest(this.cgroupCPUNames, this.samples[9], 5).map(
-            x => serviceRow(x[0], Number(x[1] / 10 / numCpu).toFixed(1)) // usec/s → percent
+            ([key, value, is_user]) => serviceRow(key, Number(value / 10 / numCpu).toFixed(1), is_user) // usec/s → percent
         );
 
         newState.topServicesMemory = n_biggest(this.cgroupMemoryNames, this.samples[10], 5).map(
-            x => serviceRow(x[0], cockpit.format_bytes(x[1], 1000))
+            ([key, value, is_user]) => serviceRow(key, cockpit.format_bytes(value, 1000), is_user)
         );
 
         this.setState(newState);
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -700,6 +700,8 @@ class TestCurrentMetrics(MachineCase):
         self.addCleanup(self.machine.execute, "systemctl unmask packagekit")
         # make sure to clean up our test resource consumers on failures
         self.addCleanup(self.machine.execute, "systemctl stop cockpittest.slice 2>/dev/null || true")
+        self.addCleanup(self.machine.execute, "su - admin -c 'XDG_RUNTIME_DIR=/run/user/$(id -u admin) "
+                                              "systemctl --user stop cockpittest.slice 2>/dev/null || true'")
         login(self)
 
     def testCPU(self):
@@ -755,6 +757,16 @@ class TestCurrentMetrics(MachineCase):
         # this settles down slowly, don't wait for becoming really quiet
         with b.wait_timeout(180):
             b.wait(lambda: float(b.text("#load-avg .pf-l-flex div:first-child").split()[-1].rstrip(',')) < 10)
+
+        # Test link to user services
+        # older releases don't have CPU accounting enabled for user services
+        if m.image not in ["ubuntu-2004", "rhel-8-6", "centos-8-stream"]:
+            m.execute("su - admin -c 'XDG_RUNTIME_DIR=/run/user/$(id -u admin) systemd-run --user --collect --slice cockpittest -p CPUQuota=60% --unit cpu-userhog dd if=/dev/urandom of=/dev/null'")
+            # user services are always running underneath user@1000.service, so these two will compete for row 1 or 2
+            b.wait_in_text("table[aria-label='Top 5 CPU services'] tbody", "cpu-userhog")
+            b.click("table[aria-label='Top 5 CPU services'] tbody tr:contains(cpu-userhog) td[data-label='Service'] a span")
+            b.enter_page("/system/services")
+            b.wait_in_text(".service-name", "/usr/bin/dd if=/dev/urandom of=/dev/null")
 
     def testMemory(self):
         b = self.browser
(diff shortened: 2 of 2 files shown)


tomasmatus:
It appears that `user@1000` does not require the `?owner=user` in link.

pkg/metrics/metrics.jsx
```diff
@@ -382,19 +382,25 @@ class CurrentMetrics extends React.Component {
...
             names.forEach((k, i) => {
                 const v = values[i];
+                let user;
                 // filter out invalid values, the empty (root) cgroup, non-services
                 if (k.endsWith('.service') && typeof v === 'number' && v != 0) {
+                    if (k.match(/^user.*user@\d+\.service.+/)) {
+                        user = true;
+                    } else {
+                        user = false;
+                    }
```
martinpitt, line 392: Please prefer writing code to (1) localize variables as much as possible (so that they don't leak into a bigger scope than necessary -- easier to follow in developer's head, and easier to avoid conflicts), and (2) making them `const` whenever possible.  Also, as this is not an user name, but a boolean flag, rename it a bit:

So please write this as

    const is_user = k.match(/^user.*user@\d+\.service.+/);

pkg/metrics/metrics.jsx
```diff
@@ -382,19 +382,25 @@ class CurrentMetrics extends React.Component {
...
+                    merged.push([label, v, user]);
                 }
             });
             merged.sort((a, b) => b[1] - a[1]);
             return merged.slice(0, n);
         }
 
-        function serviceRow(name, value) {
-            const name_text = <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service") }><TableText wrapModifier="truncate">{name}</TableText></Button>;
+        function serviceRow(name, value, isUser) {
```
martinpitt, line 402: Right, this then matches nicely with the `is_user` renaming above.

pkg/metrics/metrics.jsx
```diff
@@ -382,19 +382,25 @@ class CurrentMetrics extends React.Component {
...
+            const name_text = <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service" + (isUser ? "?owner=user" : "")) }><TableText wrapModifier="truncate">{name}</TableText></Button>;
```
martinpitt, line 403: Please avoid such long lines. Break it up into the usual HTML multi-line structuring. Usually with wrapping in () to pacify the ELSlint checks. I know the original line was already quite long, but let's clean this up while touching the code anyway.

pkg/metrics/metrics.jsx
```diff
@@ -382,19 +382,25 @@ class CurrentMetrics extends React.Component {
...
                 if (k.endsWith('.service') && typeof v === 'number' && v != 0) {
+                    if (k.match(/^user.*user@\d+\.service.+/)) {
+                        user = true;
+                    } else {
+                        user = false;
+                    }
                     const label = k.replace(/.*\//, '').replace(/\.service$/, '');
                     // only keep cgroup basenames, and drop redundant .service suffix
-                    merged.push([label, v]);
+                    merged.push([label, v, user]);
```
martinpitt, line 395: You also need to adjust the comment at the top of the function to document the new third item in the returned list.

pkg/metrics/metrics.jsx
```diff
@@ -403,11 +409,11 @@ class CurrentMetrics extends React.Component {
 
         // top 5 CPU and memory consuming systemd units
         newState.topServicesCPU = n_biggest(this.cgroupCPUNames, this.samples[9], 5).map(
-            x => serviceRow(x[0], Number(x[1] / 10 / numCpu).toFixed(1)) // usec/s → percent
+            x => serviceRow(x[0], Number(x[1] / 10 / numCpu).toFixed(1), x[2]) // usec/s → percent
```
martinpitt, line 412: This is also a good opportunity to replace the non-descriptive `x` list with destructuring and descriptive names. Like

    ([key, value, is_user]) => serviceRow(key, ..., is_user)

This makes it easier for a developer to follow the code what happens here.

martinpitt (changes requested):
Thanks for fixing this! Can you please add tests for a user unit to `TestCurrentMetrics.test{CPU,Memory}()`, similar to the system cpu-hog etc. services that it starts?

This is a bit tricky (not sure how much you already know about systemd), so I add some hints here. check-system-services already does this a lot to test user units, it has a nice `def run_cmd()` helper which you could copy.

The test can start a user cpu hog through `m.execute()` like this:

    su - admin -c 'export XDG_RUNTIME_DIR=/run/user/$(id -u admin); systemd-run --user --collect --slice cockpittest -p CPUQuota=60% --unit cpu-hog dd if=/dev/urandom of=/dev/null'

The `XDG_RUNTIME_DIR` setting is necessary as `su` does not "enter" the existing admin user login session and systemd user instance, it literally just drops privileges of the root shell. I'm happy to explain in more detail if you are interested, of course.

Thanks!

test/verify/check-metrics
```diff
@@ -748,6 +748,13 @@ class TestCurrentMetrics(MachineCase):
         with b.wait_timeout(180):
             b.wait(lambda: float(b.text("#load-avg .pf-l-flex div:first-child").split()[-1].rstrip(',')) < 10)
 
+        # Test link to user services
+        m.execute("su - admin -c 'export XDG_RUNTIME_DIR=/run/user/$(id -u admin); systemd-run --user --collect --slice cockpittest -p CPUQuota=60% --unit cpu-hog dd if=/dev/urandom of=/dev/null'")
```
martinpitt, line 752: Please make this more robust by naming the unit "cpu-userhog" or similar, to ensure that this won't collide with the system instances from above. (Same for the memory check)

test/verify/check-metrics
```diff
@@ -748,6 +748,13 @@ class TestCurrentMetrics(MachineCase):
...
+        b.wait_text("table[aria-label='Top 5 CPU services'] tbody tr:nth-of-type(2) td[data-label='Service']", "cpu-hog")
+        b.click("table[aria-label='Top 5 CPU services'] tbody tr:nth-of-type(2) td[data-label='Service'] a span")
```
martinpitt, line 754: Let's wait until cpu-userhog makes it to the top, i.e. `tr:nth-of-type(1)` like above; the "cpu-piglet" uses less CPU, and thus we expected that in the second row.

Right now the test fails everywhere, e.g. on [fedora-35](https://logs.cockpit-project.org/logs/pull-17039-20220228-185055-ffceafe7-fedora-35-devel/log.html#293-1). The [screenshot](https://logs.cockpit-project.org/logs/pull-17039-20220228-185055-ffceafe7-fedora-35-devel/TestCurrentMetrics-testCPU-fedora-35-127.0.0.2-2301-FAIL.png) looks rather messy - there are two cpu-hog services, probably the old ones from above which already got stopped, but the UI didn't catch up yet. (That's the reason for giving them a new name). Also, user@1000 is still very busy, possibly the test case failed to stop some process in the session?

test/verify/check-metrics
```diff
@@ -811,6 +818,15 @@ class TestCurrentMetrics(MachineCase):
             b.wait_in_text(".pf-c-tooltip", "GiB total")
             b.mouse("#current-swap-usage", "mouseleave")
 
+        # Test link to user services
+        m.execute("su - admin -c \"export XDG_RUNTIME_DIR=/run/user/$(id -u admin); systemd-run --user --collect --slice cockpittest --unit mem-hog3 awk "
+                  """\'BEGIN { x = sprintf(\\"%""" + str(100) + """000000s\\",\\"\\"); system(\\"touch /tmp/hogged; sleep infinity\\") }\'\"""")
+        m.execute("while [ ! -e /tmp/hogged ]; do sleep 1; done")
+        b.wait_text("table[aria-label='Top 5 memory services'] tbody tr:nth-of-type(2) td[data-label='Service']", "mem-hog3")
```
martinpitt, line 825: Same here, we should really expect it at the top of the list, and make sure we clean up previous memory allocations from the session. Most test failures happened because mem-hog3 is at the top of the list.

martinpitt (changes requested):
Thanks! You accidentally committed a test/reference change (reverting back to an older version). Please drop this from the commit, and consider installing tools/git-hook-pre-rebase (see HACKING.md) to automatically deal with the git submodules (test/reference) on rebases.

marusak:
@tomasmatus I triggered tests and they are not  green yet. They will need some more tweaking. Happy to discuss them!

martinpitt:
Oops, when writing a proper test for this I ran into a crash on the Services page: When you stop a transient service, it goes away, and the page crashes. I added a workaround for now, but this is on my todo list to fix. Unrelated to this PR.

I pushed the following changes:
 - Drop the spurious change of test/reference
 - Clean up the admin user's test services after a test, so that the tests work repeatedly. Before they left the {cpu,mem}-userhog running, which broke the test when running a second time.
 - Account for the fact that all user services run underneath user@1000.service, and the latter will also "see" the child CPU/memory usage. So we can't pin this down to the first row, the hog.service and user@1000.service will always have similar numbers and thus compete for row 1 and 2.

I am not happy with the super-complicated quoting for mem-userhog, still looking at this. But let's give this a round of tests in between.

martinpitt:
More changes:
 - Factorized the memhog script, to avoid the ridicuously complicated quoting
 - Re-dropped the hack for the services page crash; it does appear in the log, but it's the last browser command anyway
 - Clean up the /tmp/hogged stamp file from the system service, so that the wait loop actually has something to wait for

test/verify/check-metrics
```diff
@@ -772,8 +772,13 @@ class TestCurrentMetrics(MachineCase):
         # allocate a chunk of memory; this may cause other stuff to get unmapped,
         # thus not exact addition, but usage should go up
         size = 300 if have_swap else 200
-        m.execute("systemd-run --collect --slice cockpittest --unit mem-hog awk "
-                  """'BEGIN { x = sprintf("%""" + str(size) + """000000s",""); system("touch /tmp/hogged; sleep infinity") }'""")
+        m.write("/run/memhog.sh", f"""#!/usr/bin/awk -f
```
martinpitt, line 785: FTR: This uses /run instead of /tmp, as otherwise SELinux throws a spanner into the works again, and doesn't let systemd-run execute the script.

martinpitt: Move to /usr/local/bin/, as /run doesn't work for Debians: that's mounted `noexec` there.  Also fixed to clean up after itself.

martinpitt:
Next round, fixed the test failures.

marusak (approved):
Nice, thank you!

martinpitt:
Backported to rhel-9.0 branch in https://github.com/cockpit-project/cockpit/pull/17074 , dropping label.

PR #17044, opened 2022-02-24
networking: Clarify return values of call_object_method

mvollmer:
This is a continuation of 5b4ed2fce2b495b464a09358b7467f7938e5915d.

All callers of call_object_method now explicitly convert the array of
return values into a single return value, as appropriate for the
specific method.

This still uses Cockpit's promises, but we don't go back to the old
multi-argument "then" functions because standard promises don't
support them. Thus, the code is now correct again and also ready to be
moved to standard promises.


martinpitt:
Pre-push hook *cough*

pkg/networkmanager/interfaces.js
```diff
@@ -871,7 +871,7 @@ export function NetworkManagerModel() {
                     return call_object_method(self,
                                               "org.freedesktop.NetworkManager.Settings.Connection", "Update",
                                               settings_to_nm(settings, priv(self).orig))
-                            .then(function () {
+                            .then(results => {
                                 set_settings(self, settings);
                             });
```
martinpitt, line 876: Does this actually do anything? `results` looks unused here. Should just be `.then(() => set_settings(self, settings))`?

pkg/networkmanager/interfaces.js
```diff
@@ -957,7 +959,8 @@ export function NetworkManagerModel() {
             deactivate: function() {
                 return call_object_method(get_object("/org/freedesktop/NetworkManager", type_Manager),
                                           "org.freedesktop.NetworkManager", "DeactivateConnection",
-                                          objpath(this));
+                                          objpath(this))
+                    .then(results => undefined);
```
martinpitt, line 963: This is to signify that we don't expect any return value, and don't parse the empty list?

pkg/networkmanager/interfaces.js
```diff
@@ -1109,7 +1115,7 @@ export function NetworkManagerModel() {
                                           'org.freedesktop.NetworkManager.Settings',
                                           'AddConnection',
                                           settings_to_nm(conf, { }))
-                        .then(path => get_object(path, type_Connection));
+                        .then(results => get_object(results[0], type_Connection));
```
martinpitt, line 1118: For such a case it's nice to use destructuring, and give the output arguments some descriptive name. E.g.

    .then(([con_path]) => get_object(con_path, ...))

pkg/networkmanager/interfaces.js
mvollmer: It doesn't change anything, but it follows the (new) convention that call_object_method is always used like `call_object_method(...).then(results => ...)`. That is supposed to make it clear when the "then" function gets.

pkg/networkmanager/interfaces.js
mvollmer: Exactly.

pkg/networkmanager/interfaces.js
mvollmer: Good point! We could use destructuring everywhere for even better clarity.

mvollmer:
> Pre-push hook _cough_

I am using it... hmm.

mvollmer:
> > Pre-push hook _cough_
> 
> I am using it... hmm.

Ah, I never compiled this with linting...

pkg/networkmanager/interfaces.js
mvollmer: Done!  Unfortunately, eslint doesn't like empty array patterns like `.then(([]) => undefined)`, oh well...

martinpitt:
Moving to draft to clean up review list, as long as this has test failures

pkg/networkmanager/interfaces.js
```diff
@@ -882,11 +882,13 @@ export function NetworkManagerModel() {
             activate: function (dev, specific_object) {
                 return call_object_method(get_object("/org/freedesktop/NetworkManager", type_Manager),
                                           "org.freedesktop.NetworkManager", "ActivateConnection",
-                                          objpath(this), objpath(dev), objpath(specific_object));
+                                          objpath(this), objpath(dev), objpath(specific_object))
+                        .then(([active_connection]) => active_connection);
```
martinpitt, line 886: So I suppose this return value is not actually being used anywhere? At least right now? Otherwise it would be completely broken right now.

I.e. this PR is just to give us correct return values in case we ever start using them?

martinpitt (approved):

PR #17048, opened 2022-02-25
test: Put back image-prepare --overlay

martinpitt:
This was removed in commit e060058027f, but we are using it for the
dnf-copr scenario.

---

See https://github.com/cockpit-project/bots/pull/2993 , where the [dnf-copr scenario failed](https://logs.cockpit-project.org/logs/pull-2993-20220224-025251-50c295d9-fedora-testing-dnf-copr-cockpit-project-cockpit/log.html)


martinpitt:
Note: The test is still expected to fail due to the bugzilla API change in ABRT, like in https://github.com/cockpit-project/bots/pull/2993. That will be handled separately. But it should at least prepare and run all other test. I did that locally and double-checked that both cockpit and dnf in the VM have the expected version (COPR and local build)

marusak (approved):

PR #17049, opened 2022-02-25
test: New libreport uses bugzilla API key

marusak:

PR #17053, opened 2022-02-25
misc build system swamp draining

allisonkarlitskaya:
See individual commit messages.


po/.gitignore
```diff
@@ -1,3 +0,0 @@
-Makefile.in.in
```
martinpitt, line 1: Ugh, admittedly I wasn't even aware that we had gitignore files in subdirectories. By all means let's please kill them. Thanks!

Makefile.am
```diff
@@ -98,6 +98,17 @@ AM_CPPFLAGS = \
 	-DPACKAGE_SYSCONF_DIR=\""$(sysconfdir)"\" \
 	$(NULL)
 
+AM_CFLAGS = \
+	-Wall \
+	-Werror=strict-prototypes \
```
martinpitt, line 103: Note to self: Apparently all of them are valid for clang as well, the unit test is happy.

martinpitt (commented):
Nice stuff, thanks! Not formally approving yet, as still draft/no-test.

Makefile.am
allisonkarlitskaya: This patch is actually broken: it's overridden via the various `*_CFLAGS` variables that we define per-target.  I've dropped this for now, and will bring it back as part of a more comprehensive cleanup that I've been preparing.

martinpitt (approved):
Thanks!

allisonkarlitskaya:
The journal fails are #17049

PR #17055, opened 2022-02-27
fmf: Skip TestPages.testBasic on RHEL 9.0

martinpitt:
This test repeatedly fails on the RHEL testing farm without any error
message, and then corrupts the VM. Skip it for now, until we can
investigate what's going on. Leave it running on the public Testing Farm
to ensure it does not get broken further.

---

This test works fine in packit (public TF) and even [CentOS 9 Stream](https://gitlab.com/redhat/centos-stream/rpms/cockpit/-/merge_requests/19) (also public TF), but failed in [RHEL downstream packaging](https://src.osci.redhat.com/rpms/cockpit/pull-request/99)

### DIFF
--- a/test/run-verify-host-user.sh
+++ b/test/run-verify-host-user.sh
@@ -105,6 +105,11 @@ if [ -n "$test_basic" ]; then
     if [ "$TEST_OS" = "fedora-36" ]; then
         TESTS="${TESTS/TestSOS/}"
     fi
+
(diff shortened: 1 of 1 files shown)


marusak (approved):

mvollmer (approved):

mvollmer:
Might be a timeout.  TestPages.testBasic is now properly testing all locales, which has added some minutes to its run-time.

PR #17058, opened 2022-02-28
remove deprecated GLib use

allisonkarlitskaya:
- port `CockpitAuth` to `GTask` (finally)
 - sed-patch some old instance-private-data users to the new stuff
 - drop the "min version" workaround we were using to avoid deprecation warnings


martinpitt:
LGTM, thanks! However, [test-handlers](https://github.com/cockpit-project/cockpit/runs/5356427398?check_suite_focus=true#step:4:62280) has a reported leak under valgrind, that needs to be fixed first.

allisonkarlitskaya:
> LGTM, thanks! However, [test-handlers](https://github.com/cockpit-project/cockpit/runs/5356427398?check_suite_focus=true#step:4:62280) has a reported leak under valgrind, that needs to be fixed first.

Should be fixed now.

PR #17059, opened 2022-02-28
test: Be verbose when a test times out

mvollmer:
So that this event can be spotted easily when looking at logs.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -58,7 +58,7 @@ class Test:
         if self.nondestructive:
             assert self.machine_id is not None, f"need to assign nondestructive test {self} {self.command} to a machine"
         self.outfile = tempfile.TemporaryFile()
-        self.process = subprocess.Popen(["timeout", str(self.timeout)] + self.command,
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! It's not super-obvious:
```
❱❱❱ timeout -v 3 sleep 10
timeout: sending signal TERM to command ‘sleep’
```
but at least better than nothing at all.

mvollmer:
This might help shed light on the issue described in #17055, but run-tests would reset the global machine after a time out, and since that doesn't happen, it's probably something else...

mvollmer:
> It's not super-obvious:

It should be the last line right before the "not ok", so I guess we will spot it easily enough.

martinpitt:
@mvollmer : In gating, the test runs with `--machine 127.0.0.1`, there is no way to reset it as we can't launch VMs there. We just have the single existing instance that we are given.

martinpitt (approved):
Pushed a clean rebase to clean up after the journal test breakage.

PR #17063, opened 2022-02-28
TestMenu improvements

martinpitt:
I'm trying to get to the bottom of the [TestMenu.testBasic failure](https://logs.cockpit-project.org/logs/pull-17043-20220224-094403-fd5a366e-rhel-8-6/log.html#138-2) which happens very often. It's our current [top flake](https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&days=7) -- it fails in [2 out of 3 cases on centos-8-stream](https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit&days=7&test=test%2Fverify%2Fcheck-shell-menu+TestMenu.testBasic)!

[example 2](https://logs.cockpit-project.org/logs/pull-17043-20220224-094403-fd5a366e-centos-8-stream/log.html#138-2), [example 3](https://logs.cockpit-project.org/logs/pull-17059-20220228-131339-7b5d1203-centos-8-stream/log.html)

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1307,8 +1307,7 @@ class MachineCase(unittest.TestCase):
         self.addCleanup(cleanup_home_dirs)
 
         # cockpit configuration
-        self.addCleanup(m.execute, "rm -f /etc/cockpit/cockpit.conf")
-        self.addCleanup(m.execute, "rm -f /etc/cockpit/machines.d/*")
+        self.addCleanup(m.execute, "rm -f /etc/cockpit/cockpit.conf /etc/cockpit/machines.d/* /etc/cockpit/*.override.json")
 
         if not m.ostree_image:
             # for storage tests
--- a/test/verify/check-shell-host-switching
+++ b/test/verify/check-shell-host-switching
@@ -117,7 +117,7 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
         if self.is_devel_build():
             for machine in ["machine2", "machine3"]:
                 for pkg in ["packagekit", "systemd", "playground"]:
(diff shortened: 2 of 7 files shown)


marusak (approved):
Thanks!
The 75s check was there to validate that the 60s behavior was not happening always. But I guess if we would happen to break it, then we would notice random log outs in different tests.

martinpitt:
@marusak : Right, I know -- but what makes 60s special if the option isn't set at all? (We used to default to 15 minutes, for example)

mvollmer:
Maybe "test: Avoid override files in /usr where possible" should be its own PR?  It seems rather general and unrelated.

mvollmer:
Which change makes TestMenu more reliable?  Splitting out the session timeout test?

martinpitt:
> Maybe "test: Avoid override files in /usr where possible" should be its own PR?

I initially thought that I could use it for the check-menu setup of the changed manifests. But turns out I can't yet. So it is unrelated indeed, but then it didn't hurt to just piggy-back it here.

> Which change makes TestMenu more reliable? Splitting out the session timeout test?

Mostly yes, at least that's my theory. Showing the "About" dialog takes some considerable time on CI due to all the package querying that happens. My guess is that this already took a significant portion of the idle timeout and thus broke the rather carefully tuned sleeps.

mvollmer:
> > Which change makes TestMenu more reliable? Splitting out the session timeout test?
> 
> Mostly yes, at least that's my theory. Showing the "About" dialog takes some considerable time on CI due to all the package querying that happens. My guess is that this already took a significant portion of the idle timeout and thus broke the rather carefully tuned sleeps.

Makes sense, thanks!

PR #17064, opened 2022-02-28
create libcockpit-test.a, many other cleanups

allisonkarlitskaya:
The goals here are, roughly:
 - split the testing code out from the main library
   - possibly add a static code check to make sure testlib/* is only included from files that have names starting with `test-`?
 - compile each file only once
 - visually cleaner
 - reduce dependencies between different parts of Makefiles: for example, adding a new testcase should involve only adding lines in one place
 - removing cruft


martinpitt (commented):
I did an initial coarse read through this, and this generally LGTM. Could use some commit splitting, but if it's too much effort, it's ok as-is, too.

allisonkarlitskaya:
Some of the groundwork here landed in #17088.  Let's rebase on that and finish this up before it rots...

martinpitt:
The [arch failure](https://logs.cockpit-project.org/logs/pull-17064-20220309-044246-05d882a8-arch/log.html) is unrelated, I just saw it in https://github.com/cockpit-project/bots/issues/3046 as well. That arch mirror is down at the moment.

src/bridge/Makefile.am
```diff
@@ -3,10 +3,11 @@
...
 noinst_LIBRARIES += libcockpit-metrics.a
 
-libcockpit_metrics_a_CFLAGS = \
+libcockpit_metrics_a_CPPFLAGS = \
 	-DG_LOG_DOMAIN=\"cockpit-bridge\" \
-	$(COCKPIT_BRIDGE_CFLAGS) \
-	$(NULL)
+	$(glib_CFLAGS) \
+	$(json_glib_CFLAGS) \
+	$(AM_CPPFLAGS)
```
martinpitt, line 10: This is more philosophical, but does it make sense to include `*_CFLAGS` into `CPPFLAGS`? Also, do we need to give `AM_CFLAGS` explicitly? If not, why is that implied, but `AM_CPPFLAGS` is not? Should this be split into two _CPPFLAGS and _CFLAGS rules perhaps?

It's the same for other paragraphs below, but let's discuss it here first.

src/bridge/Makefile.am
```diff
@@ -277,24 +239,35 @@ polkitdir        = $(datadir)/polkit-1/actions
...
 
-# make sure this ends up in the tarball, even with --disable-pcp
-pmlogconf_files = src/bridge/pmlogconf/cockpit
-EXTRA_DIST += $(pmlogconf_files)
-
 if ENABLE_PCP
 
+pmlogconfdir = $(localstatedir)/lib/pcp/config/pmlogconf/tools
+dist_pmlogconf_DATA = src/bridge/pmlogconf/cockpit
+
```
martinpitt, line 254: I still think we want the file in tarballs even with `--disable-pcp`, as the deleted comment says. Does the `_DATA` rule work differently?

martinpitt (changes requested):
Thanks! Mostly LGTM, but I find the mixing of preprocessor and compiler flags rather confusing. This looks really unclean/wrong.

src/bridge/Makefile.am
allisonkarlitskaya: The reason for this is that the `--cflags` parameter to `pkg-config` isn't as specific about the distinction between `CFLAGS` and `CPPFLAGS` as automake is.  As a result, most-to-all of what ends up in `glib_CFLAGS` is actually flags for the preprocessor.  That's mostly `-I` paths, but also `-pthread` which is also a flag for the preprocessor (which results in definition of some additional macros).

At the end of the day it's pretty academic: no part of our build process calls the preprocessor separately from the comiler and these two things always end up being pasted together.  In my mind though, I'd prefer to reserve the `CFLAGS` for things like debugging, optimisation, and warnings flags (which are common to all of our targets).

allisonkarlitskaya: I just realised though that the way I define the `CFLAGS` for `cockpit-tls` is technically incorrect as per the above definitions.  I'm going to rework that.  I think I actually want to add `-pthread` to our top-level project-wide `CFLAGS` (yes, I know I just said that it's actually a preprocessor flag, but .. eh... it's a weird case).  It doesn't actually do anything anymore as far as I can tell (since all code is reentrant by default now) but, strictly speaking, it's correct.

src/bridge/Makefile.am
allisonkarlitskaya: Thanks for raising this.  I thought it did, but by reading the generated code in `Makefile.in`, I think that maybe I was wrong.  I'll change this back.

allisonkarlitskaya: Ah, no.  I was right in the first place.  We get this emitted:

```
@ENABLE_PCP_TRUE@dist_pmlogconf_DATA = src/bridge/pmlogconf/cockpit
```

but also this:

```
am__dist_pmlogconf_DATA_DIST = src/bridge/pmlogconf/cockpit
...
DIST_COMMON = ... $(am__dist_pmlogconf_DATA_DIST) ...
```

so, indeed, this is exactly the point of using `dist_*` instead of `EXTRA_DIST`.

martinpitt: Ack, thanks for checking!

src/bridge/Makefile.am
martinpitt: Thanks, pkg-config not separating these makes sense.

There's `AM_CPPFLAGS` for project-wide flags already anyway, so `-pthread` might be appropriate there?

So that leaves my question about why `$(AM_CPPFLAGS)` needs to be specified explicitly, but `$(AM_CFLAGS)` doesn't?

allisonkarlitskaya: the rule is that if you specify a per-target `foo_CFLAGS` then it uses that, otherwise it uses `AM_CFLAGS`.  ditto `foo_CPPFLAGS` and `AM_CPPFLAGS`.  That's why I'm always careful to include `AM_CPPFLAGS` in the `foo_CPPFLAGS` overrides.  Since we generally *don't* override `foo_CFLAGS` we get the default `AM_CFLAGS` automatically.

That said, I just found a few places where that isn't true (specifically, when we add `-fPIC`).  I'll fix those up in the next round :)

martinpitt (dismissed):
Thanks for the explanations! Test failures are accounted for, so LGTM and ready to land.

martinpitt (approved):

PR #17066, opened 2022-03-01
Fix ESLint errors in test/common/

martinpitt:
Now it's almost clean:
```
❱❱❱ node_modules/.bin/eslint test/common/

/var/home/martin/upstream/c3/test/common/chromium-cdp-driver.js
  221:72  error  Extra semicolon  semi

/var/home/martin/upstream/c3/test/common/firefox-cdp-driver.js
  290:72  error  Extra semicolon  semi

✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```

That remaining error is fixed in PR #17062, and I want to avoid colliding.

Once these two land, I'll add it to the `npm run eslint` rules in package.json. But that will also incur a node_modules/ refresh, so let's do that separately.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 node_modules/*
+test/common/sizzle.js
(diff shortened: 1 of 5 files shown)


marusak (approved):
Thank you!

PR #17067, opened 2022-03-01
packit: build SRPM in Copr

TomasTomecek:
...and be able to specify the precise list of deps needed to create the SRPM

TODO:
* [x] resolve the problem that specfile_path is needed

The SRPM is also created in a fresh VM instead of a locked-down openshift pod.

Please bear in mind this feature made it to production just today :)

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -9,6 +9,17 @@ actions:
     - tools/webpack-jumpstart --wait --rebase
     - make XZ_OPT=-0 dist
     - sh -c 'echo cockpit-*.tar.xz'
+srpm_build_deps:
+  - npm
+  - selinux-policy
+  - autogen
+  - autoconf
+  - automake
+  - make
+  - gcc
+  - glib2-devel
+  - systemd-devel
+  - xmlto
 jobs:
   - job: tests
     trigger: pull_request


packit.yaml
```diff
@@ -9,6 +9,18 @@ actions:
     - tools/webpack-jumpstart --wait --rebase
     - make XZ_OPT=-0 dist
     - sh -c 'echo cockpit-*.tar.xz'
+specfile_path: cockpit.spec
```
martinpitt, line 12: Why is this necessary? As per the [docs](https://packit.dev/docs/configuration/), the implicit default should be downstream_package_name`.spec`, and `downstream_package_name` defaults to the github repo name ("cockpit").

packit.yaml
```diff
@@ -9,6 +9,18 @@ actions:
...
+srpm_build_deps:
```
martinpitt, line 13: Cool! That may also allow us to drop some hacks which avoid some build dependencies.

martinpitt (commented):
Cheers! One question, and the TF seems laggy/broken (tests never start) but the [srpm build worked](https://copr-be.cloud.fedoraproject.org/results/packit/cockpit-project-cockpit-17067/srpm-builds/03588470/builder-live.log.gz).

The only drawback is that building the srpm alone (*not* the rpms) takes 5 mins now, mostly due to having to install all the build deps all the time; it just took a few seconds in the sandcastle container. But in general it makes sense to lose custom infra and move workloads to well-maintained common ones.

Once that lands, I'm happy to apply the same thing to starter-kit, c-podman and friends.

packit.yaml
TomasTomecek: yup, exactly!

Although I personally dislike that a lot of these deps are a duplicate of BuildRequirements from the spec file. But, in order to extract them from the spec, we would need to process the spec, which means triggering action commands, which means running them in our sandbox... Or create a custom solution in collaboration with Copr.

packit.yaml
TomasTomecek: whoops, so it's a bug on our side then; it did not populate it correctly unless I explicitly specified the path

TomasTomecek:
> Cheers! One question, and the TF seems laggy/broken (tests never start) but the [srpm build worked](https://copr-be.cloud.fedoraproject.org/results/packit/cockpit-project-cockpit-17067/srpm-builds/03588470/builder-live.log.gz).
> 
> The only drawback is that building the srpm alone (_not_ the rpms) takes 5 mins now, mostly due to having to install all the build deps all the time; it just took a few seconds in the sandcastle container. But in general it makes sense to lose custom infra and move workloads to well-maintained common ones.
> 
> Once that lands, I'm happy to apply the same thing to starter-kit, c-podman and friends.

Martin, you're right. We are concerned about this as well. I hope that we'll be able to implement optimizations in the future to speed this process up. We already discussed this with the Copr team.

martinpitt:
@TomasTomecek: The night-and-day change for debian chroots is [eatmydata](https://packages.debian.org/sid/eatmydata), a preload library which nop-ifys all fsync()/sync() calls. package managers tend to do that a lot, which makes sense on production machines, but is an utter nuisance for CI environments. This can easily speed up installing lots of packages from minutes to seconds. Unfortunately it's not packaged for Fedora/RHEL, but it's also a very simple project.

packit.yaml
TomasTomecek: I cannot reproduce locally :| so let's drop it and see

TomasTomecek:
@martinpitt thank you, I'll bring this up! (the library is hilarious :) https://github.com/stewartsmith/libeatmydata/blob/master/libeatmydata/libeatmydata.c#L139 )

packit.yaml
TomasTomecek: reproduced, fixed; will rebase this PR once available in production

martinpitt:
@TomasTomecek : Thanks! Could you please put it back with a comment that points to the fix, so that we can drop it later? It doesn't hurt to have it, as long as we know why. Then we can land this and don't have to block it.

martinpitt:
@TomasTomecek : Oh, and please rebase on your next push, so that packit runs the latest code.

TomasTomecek:
These two are the fixes:
https://github.com/packit/packit/pull/1498
https://github.com/packit/packit/pull/1499

Unfortunately, we reverted the latest deployment (due to that problem with TF status reporting you ran into) and this feature is no longer available.

I'll rebase and update this PR (and will create even more :) once we have it back in production and functional :)

TomasTomecek:
I can see a few timeouts, errors and `bash: virsh: command not found` in the test output.

TomasTomecek:
hm, the errors look the same :/

martinpitt:
It still worked at least 15 hours ago in #17091. So the [c8s failure](https://artifacts.dev.testing-farm.io/c5b2556a-adef-4870-bf6a-a9e579ac9c30/) could be a very recent regression in c8s proper, or some subtle fallout from the package build (but it really doesn't look that way). #17092 is running tests, so we can compare in half an hour.

martinpitt:
PR #17092 [failed the same way](https://artifacts.dev.testing-farm.io/41fd8f77-f445-4a74-9035-4702875b4dbc/), so unrelated. We'll investigate that separately. Thanks!

martinpitt (approved):
I'll remove the S-o-b from the commit message on merging. We don't use it (and frankly, I find it pointless)

TomasTomecek:
@martinpitt I do that by default for all my commits since one of the projects I contributed to required it `¯\_(ツ)_/¯`

I'll drop it from my other cockpit PRs :)

PR #17070, opened 2022-03-02
The prompt text "Permission denied" should display internationalized text when the login failed

ReganHe93:
Our browser is in Chinese. When a user login to cockpit-ws and uses a wrong password, it always displays "Permission denied", but the user actually wants to see the prompt text in Chinese.

I found the prompt text passed by a .c file is always fetched first in login.js, and the internationalized "Permission denied" is used only when the decoding fails.
english_permission_denied

When we can confirm that the returned text is "Permission denied", using internationalized text directly will meet the needs of users.
chinese_permission_denied

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -888,7 +888,7 @@
                     }
                 }
             } else if (xhr.status == 403) {
-                login_failure(decodeURIComponent(xhr.statusText) || _("Permission denied"));
+                login_failure(_(decodeURIComponent(xhr.statusText)) || _("Permission denied"));
             } else if (xhr.statusText) {
                 fatal(decodeURIComponent(xhr.statusText));
             } else {


martinpitt (approved):
Thanks! Mind you, this is just one of several possible strings, but it certainly helps.

pkg/static/login.js
```diff
@@ -888,7 +888,11 @@
                     }
                 }
             } else if (xhr.status == 403) {
-                login_failure(decodeURIComponent(xhr.statusText) || _("Permission denied"));
+                if (!xhr.statusText || xhr.statusText.indexOf("Permission denied") > -1) {
```
marusak, line 891: We could just wrap it into `_()`. In this case `Permission denied` is known string so it would work equally and it might also be able to translate other strings.

```
                login_failure(_(decodeURIComponent(xhr.statusText)) || _("Permission denied"));
```

martinpitt: Updated accordingly.

marusak (approved):
I am not sure how to reproduce the `Permission denied` error. Entering wrong username or password leads to different error message.

But in any case, I believe this works just fine.

PR #17073, opened 2022-03-02
pf: Add xs to flex gap override

garrett:


martinpitt:
On main:

image

this PR:

image

I can't really see a difference? Or does that address something else?

martinpitt (approved):
Thanks!

martinpitt:
Backported to rhel-9.0 branch in https://github.com/cockpit-project/cockpit/pull/17074 , dropping label.

PR #17074, opened 2022-03-02
[rhel-9.0] Backport some fixes

martinpitt:
These are good and low-risk polish for the post dev-freeze RHEL update. This is also the first-ever PR against the new rhel-9.0 branch, so using this to fix tests (if necessary).

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -14,17 +14,4 @@ jobs:
     trigger: pull_request
     metadata:
       targets:
-      - fedora-35
-      - fedora-36
-      - centos-stream-8-x86_64
       - centos-stream-9-x86_64
-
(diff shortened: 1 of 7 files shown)


martinpitt:
[One failure](https://logs.cockpit-project.org/logs/pull-17074-20220302-114731-323f182c-rhel-9-0/log.html#291-2) which rings a bell: commit a5a801a9a3965 and 7c4b8bc3fa6b69a6

Also, it seems that centos-stream-9-{i386,s390x} packit targets are not a thing unfortunately. Well, not a biggie, we are not going to do intrusive C code changes to that branch anyway.

marusak (approved):
Thanks!

PR #17075, opened 2022-03-02
metrics: Compress space a bit

garrett:
I adjusted the extra spacing between the CPU bars & load and then decided to adjust all the spacing in the current metrics area.

The extra spacing between the CPU bars & load _definitely_ needs to be adjusted, but the rest are mostly subjective changes and I'm still not fully sold on it. I'd like feedback.

Before:

Screenshot 2022-03-02 at 13-19-52 Overview - garrett Bolt

After:

Screenshot 2022-03-02 at 13-19-40 Overview - garrett Bolt

### DIFF
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -123,7 +123,7 @@
   // Shrink progress bars and their gap a little
   .pf-c-progress {
       grid-gap: 0;
-      --pf-c-progress__bar--Height: var(--pf-global--spacer--sm);
+      --pf-c-progress__bar--Height: var(--pf-global--spacer--xs);
       // PF4 uses end; it probably should arguably use last baseline.
       // (perhaps end may work better for non-text content?)
       align-items: last baseline;
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -66,21 +66,21 @@
 
       .progress-stack {
           display: grid;
-          padding-bottom: var(--pf-global--spacer--md);
-          grid-gap: var(--pf-global--spacer--md) var(--pf-global--spacer--lg);
+          padding-bottom: var(--pf-global--spacer--sm);
+          grid-gap: var(--pf-global--spacer--sm) var(--pf-global--spacer--md);
 
           &:not(:first-child) {
-            padding-top: var(--pf-global--spacer--md);
+            padding-top: var(--pf-global--spacer--sm);
           }
       }
 
       .progress-stack-no-space {
           display: grid;
-          padding-bottom: var(--pf-global--spacer--lg);
+          padding-bottom: var(--pf-global--spacer--sm);
           grid-gap: 0;
 
           &:not(:first-child) {
-            padding-top: var(--pf-global--spacer--lg);
+            padding-top: var(--pf-global--spacer--sm);
           }
 
           // avoid the "x" icon on high CPU usage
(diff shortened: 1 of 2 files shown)


martinpitt:
I like this! I'm not a big fan of putting "paragraphs" everywhere anyway. This makes better use of space to not lose sight of the history, for example. Thanks!

garrett:
OK, another small change:

What if we make the progress bars half their already reduced height?

Screenshot:

Screenshot 2022-03-02 at 14-13-45 Overview - garrett Bolt

Commit: https://github.com/cockpit-project/cockpit/pull/17075/commits/85c49c73b6b41fbc4e6f77dfc67c64731d52f356

The idea is that this further de-emphasizes the bars. (With less space around items, the bars became more dominant relatively speaking.)

martinpitt:
@garrett: I like them for disks, and fine for RAM/Swap as well; I must say, I don't like them that much for the combined CPU graph. It's rather hard to tell them apart now. Can we use the normal height for the stacked graph?

martinpitt:
Or perhaps we should leave a little bit of space between the two CPU progress bars?

garrett:
> Can we use the normal height for the stacked graph?

Everything was already reduced in half. This changed reduced it by half again.

I have a patch locally that makes the CPU bars thicker. It looks like this:

Screenshot 2022-03-02 at 14-13-45 Overview - garrett Bolt

```scss
  // Shrink progress bars and their gap a little
  .pf-c-progress {
      grid-gap: 0;
      --pf-c-progress__bar--Height: var(--pf-global--spacer--xs);
      // PF4 uses end; it probably should arguably use last baseline.
      // (perhaps end may work better for non-text content?)
      align-items: last baseline;

      @at-root #current-metrics-card-cpu .pf-c-progress {
          // Make CPU usage bars thinner than default, but thicker than others
          --pf-c-progress__bar--Height: var(--pf-global--spacer--sm);
      }
  }
```
(It's the `@at-root` part.)

garrett:
Here's a gap, but reduced to 1px:

Screenshot 2022-03-02 at 15-32-42 Overview - garrett Bolt

martinpitt:
Thanks @garrett : Personally, I prefer the last one from these three -- it's nice to have all progress bars have the same thickness, and the 1px gap feels quite enough to tell the two apart.

pkg/metrics/metrics.scss
```diff
@@ -127,6 +127,15 @@
...
       align-items: last baseline;
+
+      + .pf-c-progress {
+          // Add a tiny gap for sandwiched usage bars (as in CPU)
+          padding-top: 1px;
+      }
+  }
+
+  #current-metrics-card-cpu .pf-c-progress {
+    // --pf-c-progress__bar--Height: var(--pf-global--spacer--sm);
```
martinpitt, line 138: This shouldn't stay for the commit that lands, IMHO.

garrett:
Should we land this? WDYT?

I would guess it needs some pixel test updating for the metrics page.

martinpitt (dismissed):
Thanks! I was just waiting for tests, but they are almost done.

We don't have pixel tests for the metrics page, as there is by and large nothing there which does *not* jiggle all the time :grin:

martinpitt:
Again, failed [four times](https://logs.cockpit-project.org/logs/pull-17075-20220302-175048-e2740d2a-fedora-36/log.html#51). I'll have a closer look tomorrow, I don't want to land more red PRs.

martinpitt:
That test failure reproduces straight away locally. I added a trivial fix.

martinpitt:
@jelly, @marusak : As I added a commit here, do you mind reviewing/landing this?

martinpitt (approved):

marusak (approved):
Thanks!

PR #17076, opened 2022-03-02
More nondestructive tests (split out from 17005 without nfs)

jelly:


test/verify/check-superuser
```diff
@@ -77,16 +80,16 @@ class TestSuperuser(MachineCase):
         b.leave_page()
         b.check_superuser_indicator("Administrative access")
 
+    @nondestructive
```
martinpitt, line 83: Is there some particular reason why you left testTwoPasswds as destructive? This doesn't seem difficult either, and you could make the whole class nd. But, totally fine for a follow-up of course.

test/verify/check-superuser
```diff
@@ -106,14 +107,14 @@ class TestSuperuser(MachineCase):
 
         # Configure the sudo PAM stack to make two prompts
         if "debian" in m.image or "ubuntu" in m.image:
-            m.write("/etc/pam.d/sudo", """
+            self.write_file("/etc/pam.d/sudo", """
```
martinpitt, line 110: Ah, so the next commit addresses the remaining test :grin: This could just be squashed into the first commit.

test/verify/check-superuser
```diff
@@ -134,11 +135,15 @@ session include system-auth
...
     def testWrongPasswd(self):
         b = self.browser
+        m = self.machine
 
         # Log in with limited access
         self.login_and_go(superuser=False)
         b.check_superuser_indicator("Limited access")
 
+        if m.image == "arch":
+            self.addCleanup(m.execute, "rm -f /var/run/faillock/admin")
```
martinpitt, line 145: This should be in the previous commit. If you squash them, that'll go into the right place.

martinpitt (changes requested):
Looks almost fine (just one squashing) and seems to work. Thanks!

test/verify/check-superuser
martinpitt: And please move the `@nd` to the class, instead of on each individual test.

jelly:
> Looks almost fine (just one squashing) and seems to work. Thanks!

Squashed, green let's go :)

martinpitt (approved):
/me hands @jelly the peace award for so much less destructivism

PR #17078, opened 2022-03-02
test: Code coverage for JavaScript

mvollmer:


mvollmer:
This should collect coverage data into `TestFoo.testBar.info.gz` files, one for each test.  They can be turned into a summary report via `genhtml *.info.gz --output-dir coverage-report`.

mvollmer:
Yay, this works!

mvollmer:
check-users is not writing coverage data, hmm...

mvollmer:
> > check-users is not writing coverage data, hmm...
> 
> It had a tearDown menthod but super().tearDown() was missing.

Also, calling Page.Navigate resets the coverage data, and thus we were missing a whole lot of it. Now we take a sample also before each such call.

mvollmer:
More refinements:

 - Now ranges are sorted by size so that smaller ranges can reliably punch holes into bigger ranges, which is important for nested functions.  (This mostly worked before because count == 0 was given priority over count > 0, but it that wasn't general enough.)

 - source files that are included in more than one dist file have tgheir hits merged correctly now.

test/common/run-tests
```diff
@@ -430,6 +436,17 @@ def run(opts, image):
...
 
+    # Create coverage report
+    if opts.coverage:
+        base_dir = os.path.realpath(f'{__file__}/../../..')
+        output = os.environ.get("TEST_ATTACHMENTS", base_dir)
+        lcov_files = glob.glob(f"{base_dir}/lcov/*.info.gz")
+        if len(lcov_files) > 0:
+            try:
+                subprocess.check_call(["genhtml", "--no-function-coverage", "--output-dir", f"{output}/Coverage"] + lcov_files)
+            except Exception as e:
```
martinpitt, line 447: `subprocess.CalledProcessError` -- we want to crash on other errors.

test/common/lcov.py
```diff
@@ -0,0 +1,302 @@
...
+        for line in open(path, newline='').readlines():
+            line_starts.append(line_starts[-1] + len(line))
+        self.smap = parse_sourcemap(open(path + ".map"), line_starts)
+
+    def find_sources_slow(self, start, end):
+        res = []
+        for m in self.smap:
+            if m[0] >= start and m[0] < end:
+                res.append(m)
+        return res
```
martinpitt, line 126: This can go, it's not called anywhere. Or does `find_sources()` have some disadvantage?

test/common/link-patterns.json
```diff
@@ -20,5 +20,10 @@
         "label": "changed pixels",
         "pattern": "Differences in pixel test ([A-Za-z0-9\\-\\.]+)$",
         "url": "pixeldiff.html#$1"
+    },
+    {
+        "label": "coverage",
+        "pattern": "Code coverage report in ([A-Za-z0-9\\-\\.]+)$",
```
martinpitt, line 26: The matched file name will only ever be "Coverage", right?

martinpitt (changes requested):
Thank you! This is amazing. Some little cleanup, but that's trivial.

What's the next step here? I feel like this is/will be stuck at the same place like our a11y review (the aXe report) -- we create them, but then don't do anything without it. ISTM that we need to come  up with a mechanism to compare a coverage or aXe report, and fail a PR on regressions.

We somehow need a place to store information (most usefully, the URL about the /devel test run) about the most recently landed PR; we could figure this out from the GitHub API by mapping commits to PRs. Alternatively, we could put this into the "update test-results.db" workflow, which runs on every landed PR anyway. That could update some running stats in a new file in https://logs-https-frontdoor.apps.ocp.ci.centos.org/ . WDYT?

test/common/lcov.py
mvollmer: It's the reference implementation, for making sure that `find_sources` works correctly.  I'll make that clear.

mvollmer:
> What's the next step here? 

Good question.  We could make it a requirement that a PR must not lower the coverage ratio (and implement that somehow via test-results.db or otherwise), but that could already be problematic. One might write a hole lot of new code and do ones best to cover it, but the totals might still be lower than they were before.

What about inventing a mechanism to "waive" specific uncovered line, i.e., we need to write a small comment about every uncovered line (maybe in the source itself?), and a PR is blocked until all uncovered lines are explained.

martinpitt:
> What about inventing a mechanism to "waive" specific uncovered line

I think we need to collect some experience with this first, before we make our lives hard right from the start. For example, we could teach the /devel scenario to post a comment to the PR about coverage delta, like [composer](https://github.com/osbuild/cockpit-composer/pull/113#issuecomment-311463333) used to do. Then we can treat it as an advise during PR review until we get some data. WDYT?

mvollmer:
> > What about inventing a mechanism to "waive" specific uncovered line
> 
> I think we need to collect some experience with this first, before we make our lives hard right from the start.

Yeah, also implementing this isn't trivial either.

> For example, we could teach the /devel scenario to post a comment to the PR about coverage delta, like [composer](https://github.com/osbuild/cockpit-composer/pull/113#issuecomment-311463333) used to do. Then we can treat it as an advise during PR review until we get some data. WDYT?

Yes, sounds good.  Why did composer stop doing this? :-)

test/common/lcov.py
mvollmer: Done

test/common/run-tests
mvollmer: Done.

martinpitt:
Thanks! LGTM now. It's a bit suspicious that `TestStorageLvm2.testLvm` failed the same way on both [f35](https://artifacts.dev.testing-farm.io/6acabcf8-9b8f-4df8-96af-022eb7a2244e/) and [f36](https://artifacts.dev.testing-farm.io/7d1851e3-29a3-4548-adbf-ab1477590d94/) -- but that sounds like unrelated spontaneous breakage. I'll re-run the tests to compare.

This is still draft, but I'm happy if you want to land this as-is, and we sort out the usage of this data as a follow-up. Thanks!

martinpitt (dismissed):

mvollmer:
> * [ ]  Do we really need to write multiple lcov files?  I think the coverage data might actually be complete at the end, but it will contain multiple entries for the same dist file, and since we merged multiple files wrong, this looked like it was incomplete. Now with the correct merging, we are probably good.

We don't it! Yay!

mvollmer:
Another reference, without snapshots at "navigate": https://logs.cockpit-project.org/logs/pull-17078-20220331-102416-cd9b4054-fedora-35-devel/Coverage/

This is _almost_ the same as the earlier reference, but the code itself has changed enough to maybe explain the differences. Let's make another report with the snapshots.

mvollmer:
> Another reference, without snapshots at "navigate": https://logs.cockpit-project.org/logs/pull-17078-20220331-102416-cd9b4054-fedora-35-devel/Coverage/
> 
> This is _almost_ the same as the earlier reference, but the code itself has changed enough to maybe explain the differences. Let's make another report with the snapshots.

Here: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17078-20220331-123059-4b374737-fedora-35-devel/Coverage/

That's close enough.

martinpitt (approved):
Danke!

PR #17079, opened 2022-03-02
Show memory (RAM) information for unprivileged users

jelly:
Systemd now exposes physical memory information via udev which which means we don't need dmidecode anymore and normal users can view memory information.


jelly:
dmidecode fix.
```
Handle 0x0004, DMI type 17, 84 bytes
Memory Device
        Array Handle: 0x0002
        Error Information Handle: Not Provided
        Total Width: 64 bits
        Data Width: 64 bits
        Size: 16 GB
        Form Factor: SODIMM
        Set: None
        Locator: ChannelB-DIMM0
        Bank Locator: BANK 2
        Type: DDR4
        Type Detail: Synchronous
        Speed: 3200 MT/s
        Manufacturer: Samsung
        Serial Number: 00000000
        Asset Tag: None
        Part Number: M471A2G44AM0-CWE
        Rank: 1
        Configured Memory Speed: 2667 MT/s
        Minimum Voltage: Unknown
        Maximum Voltage: Unknown
        Configured Voltage: 1.2 V
        Memory Technology: DRAM
        Memory Operating Mode Capability: Volatile memory
```

pkg/lib/machine-info.js
```diff
@@ -248,7 +248,7 @@ function processMemory(info) {
             memorySize = cockpit.format(_("$0 GiB"), memorySizeValue / 1024);
         }
 
-        let memoryTechnology = memoryProperty["Memory technology"];
+        let memoryTechnology = memoryProperty["Memory Technology"];
```
martinpitt, line 251: We didn't dream this up -- back then we collected the actual output and added it to the mock in the test suite. I suppose it was changed in some newer version, but let's please not break older dmidecode versions then - can you please make this case independent?

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,58 @@ function findPCI(udevdb, info) {
...
+    }
+
+    const props = udevdb[dmipath];
+    // Systemd now exposes memory information in udev, introduced in systemd => 248
+    // https://github.com/systemd/systemd/blob/main/NEWS#L1713
+    if ('MEMORY_ARRAY_NUM_DEVICES' in props) {
+        for (let slot = 0; slot < props.MEMORY_ARRAY_NUM_DEVICES; slot++) {
+            let memorySize = props[`MEMORY_DEVICE_${slot}_SIZE`];
+            if (memorySize) {
+                memorySize = cockpit.format(_("$0 GB"), memorySize / 1024 / 1024 / 1024);
```
martinpitt, line 80: unit confusion, this is GiB. Please use our standard [cockpit.format_bytes()](https://cockpit-project.org/guide/latest/cockpit-util.html) API for that, with an explicit `factor=1024` (as this is one of the very few cases where we actually *do* want base-2 units)

test/verify/check-system-info
```diff
@@ -566,26 +566,28 @@ revision\t: 2.3 (pvr 004e 1203)
...
-            b.wait_text('#memory-listing tr:nth-of-type(1) td[data-label=Rank]', "Unknown")
-            b.wait_in_text('#memory-listing tr:nth-of-type(1) td[data-label=Speed]', "1600 MT/s")
-            # second DIMM has known technology and rank
-            b.wait_in_text('#memory-listing tr:nth-of-type(2) td[data-label=ID]', "BANK 0: ChannelB-DIMM0")
-            b.wait_text('#memory-listing tr:nth-of-type(2) td[data-label="Memory technology"]', "Awesome")
-            b.wait_text('#memory-listing tr:nth-of-type(2) td[data-label=Rank]', "Single rank")
-            # Locator can be the same on multi-channel memory platforms - bank locator should differentiate them
-            b.wait_in_text('#memory-listing tr:nth-of-type(3) td[data-label=ID]', "BANK 1: ChannelB-DIMM0")
+            systemd_version = int(m.execute("systemctl --version | awk '{print $2; exit}'").strip())
+            if systemd_version < 248:
```
martinpitt, line 570: Can you please add an equivalent `udevadm` mock, to have the parsing of that in CI? Ideally the assertions should be exactly the same, just the mock setup should differ.

martinpitt (changes requested):
Thanks!

pkg/lib/machine-info.js
jelly: I looked into dmidecode.git and I couldn't find it, as QEMU vm's don't show memory technology I can't really reproduce it here.

jelly:
```
[root@rhel-8-6-127-0-0-2-2201 ~]# strings /usr/sbin/dmidecode  | grep -i techno
Memory Technology
```
Hmmm maybe debian-ism?

But that also reports: (debian-stable)

```
root@debian:~# strings /usr/sbin/dmidecode  | grep -i techno
Memory Technology
```

jelly: I still think it's fine, but dropped the patch for now.

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,60 @@ function findPCI(udevdb, info) {
...
+    }
+
+    const props = udevdb[dmipath];
+    // Systemd now exposes memory information in udev, introduced in systemd => 248
+    // https://github.com/systemd/systemd/blob/main/NEWS#L1713
+    if (!('MEMORY_ARRAY_NUM_DEVICES' in props)) {
+        return;
+    }
+
+    for (let slot = 0; slot < props.MEMORY_ARRAY_NUM_DEVICES; slot++) {
```
martinpitt, line 80: Ugh, is that JS automagically turning a string into a number? (Tested in `node`, seems to work). An explicit `parseInt()` would be a bit cleaner, though.

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,60 @@ function findPCI(udevdb, info) {
...
+        let memorySize = props[`MEMORY_DEVICE_${slot}_SIZE`];
+        if (memorySize) {
+            memorySize = cockpit.format(_("$0 GiB"), memorySize / 1024 / 1024 / 1024);
```
martinpitt, line 83: Please use standard `cockpit.format_bytes()` for this.

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,60 @@ function findPCI(udevdb, info) {
     }
 }
 
+function findMemoryDevices(udevdb, info) {
+    const memoryArray = [];
+    const dmipath = '/devices/virtual/dmi/id';
+    if (!(dmipath in udevdb)) {
```
martinpitt, line 69: Style issue: You don't need { } wrappers for a single line. The code further below doesn't do it either.

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,60 @@ function findPCI(udevdb, info) {
...
+            memorySize = cockpit.format(_("$0 GiB"), memorySize / 1024 / 1024 / 1024);
+        } else {
+            memorySize = _("Unknown");
+        }
+
+        let memoryRank = props[`MEMORY_DEVICE_${slot}_RANK`] || _("Unknown");
+        if (memoryRank == 1)
+            memoryRank = _("Single rank");
+        if (memoryRank == 2)
+            memoryRank = _("Dual rank");
```
martinpitt, line 92: Please do this as an `else if` and add an `else` just in case there's ever an "Unknown" or > 2.

martinpitt (changes requested):
Initial review (need to run now, sorry). Thanks!

test/verify/check-system-info
```diff
@@ -477,7 +424,8 @@ class TestSystemInfo(MachineCase):
         # now pretend this is a system without DMI
         b.logout()
         m.execute("mount -t tmpfs none /sys/class/dmi/id")
-        self.addCleanup(m.execute, "umount /sys/class/dmi/id")
+        # check if it's mounted as the memory tests umount it.
+        self.addCleanup(m.execute, "! findmnt /sys/class/dmi/id || umount /sys/class/dmi/id")
```
martinpitt, line 428: `mountpoint -q` is more specific

test/verify/check-system-info
```diff
@@ -559,33 +507,49 @@ revision\t: 2.3 (pvr 004e 1203)
         b.enter_page('/system/hwinfo')
         b.wait_in_text('#hwinfo-system-info-list .hwinfo-system-info-list-item:nth-of-type(2) .pf-c-description-list__group:nth-of-type(1) dd', "2x POWER9 (architected), altivec supported")
 
-        # Memory details should be shown for generic QEMU memory
-        # demidecode not available on certain images
+        # Memory details should be shown from our mocked DMI information from systemd's test files.
         if not m.ostree_image:
```
martinpitt, line 511: fedora-coreos is a recent Fedora 35, that should now actually work?

test/verify/check-system-info
```diff
@@ -559,33 +507,49 @@ revision\t: 2.3 (pvr 004e 1203)
...
             b.wait_in_text('#hwinfo #memory-listing' + heading_selector, "Memory")
             b.wait_in_text('#hwinfo #memory-listing table', "DIMM")
             b.wait_in_text('#hwinfo #memory-listing table', "RAM")
 
-            # Test more specific memory data with a fake dmidecode
-            m.write('/tmp/dmidecode', dmidecode)
-            m.execute('chmod +x /tmp/dmidecode; mount -o bind /tmp/dmidecode $(which dmidecode)')
-            self.addCleanup(m.execute, 'until umount $(which dmidecode); do sleep 1; done')
+            m.upload(["verify/files/dmi/smbios_entry_point", "verify/files/dmi/Lenovo-ThinkPad-X280.bin"], "/tmp/")
+            m.execute("mount -o bind /tmp/smbios_entry_point /sys/firmware/dmi/tables/smbios_entry_point")
```
martinpitt, line 517: How about mounting /sys/firmware/dmi/tables/ directory, let m.upload() copy directly into that, and unmount it just once at cleanup?

test/verify/check-system-info
```diff
@@ -559,33 +507,49 @@ revision\t: 2.3 (pvr 004e 1203)
...
-            # first DIMM has unknown technology and rank
+            systemd_version = int(m.execute("systemctl --version | awk '{print $2; exit}'").strip())
+
+            # Test more specific memory data with a fake dmidecode
             b.wait_in_text('#memory-listing tr:nth-of-type(1) td[data-label=ID]', "BANK 0: ChannelA-DIMM0")
-            b.wait_in_text('#memory-listing tr:nth-of-type(1) td[data-label=Type]', "DDR3")
-            b.wait_in_text('#memory-listing tr:nth-of-type(1) td[data-label=Size]', "8 GiB")
+            b.wait_in_text('#memory-listing tr:nth-of-type(1) td[data-label=Type]', "DDR4")
+            # dmidecode shows GB, ubuntu backports changes to systemd 245
+            if systemd_version < 248 and m.image != "ubuntu-2004":
```
martinpitt, line 533: Such a mix is ugly IMHO -- it's nicer to just have a list of "old" OSes that shrinks to zero over time, and an OS list is much easier to grep for. E.g. once we drop testing rhel-8-6 or ubuntu-2004, we grep for these OS image names.

test/verify/check-system-info
```diff
@@ -559,33 +507,49 @@ revision\t: 2.3 (pvr 004e 1203)
...
-            # second DIMM has known technology and rank
-            b.wait_in_text('#memory-listing tr:nth-of-type(2) td[data-label=ID]', "BANK 0: ChannelB-DIMM0")
-            b.wait_text('#memory-listing tr:nth-of-type(2) td[data-label="Memory technology"]', "Awesome")
+            b.wait_text('#memory-listing tr:nth-of-type(1) td[data-label=Rank]', "Single rank")
+            b.wait_in_text('#memory-listing tr:nth-of-type(1) td[data-label=Speed]', "2400 MT/s")
+
+            b.wait_in_text('#memory-listing tr:nth-of-type(2) td[data-label=ID]', "BANK 2: ChannelB-DIMM0")
+            b.wait_in_text('#memory-listing tr:nth-of-type(2) td[data-label=Type]', "DDR4")
+            # dmidecode shows GB, ubuntu backports changes to systemd 245
+            if systemd_version < 248 and m.image != "ubuntu-2004":
```
martinpitt, line 545: DRY, please set some `supports_udev_meminfo` boolean once.

martinpitt (changes requested):
Please mention the origin of the two DMI dumps in the commit message. Cheers!

test/verify/check-system-info
jelly: Hmm this was already conditionality before, I guess fedora-coreos didn't have dmidecode?

martinpitt: Exactly. But now with this PR it doesn't need to any more, as it surely does have udev. I mean we should be able to entirely drop the `if not m.ostree_image:` condition and run this on coreos as well.

martinpitt:
still [fails on debian-stable](https://logs.cockpit-project.org/logs/pull-17079-20220307-112802-141971a1-debian-stable/log.html#300)

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,62 @@ function findPCI(udevdb, info) {
...
+        return;
+
+    const props = udevdb[dmipath];
+    // Systemd now exposes memory information in udev, introduced in systemd => 248
+    // https://github.com/systemd/systemd/blob/main/NEWS#L1713
+    if (!('MEMORY_ARRAY_NUM_DEVICES' in props)) {
+        return;
+    }
+
+    for (let slot = 0; slot < parseInt(props.MEMORY_ARRAY_NUM_DEVICES, 10); slot++) {
```
martinpitt, line 79: Please compute this as a constant once, instead of in each loop iteration.

pkg/systemd/hw-detect.js
```diff
@@ -61,6 +63,62 @@ function findPCI(udevdb, info) {
...
+        let memorySize = parseInt(props[`MEMORY_DEVICE_${slot}_SIZE`], 10);
+        if (memorySize) {
+            memorySize = cockpit.format_bytes(memorySize);
+        } else {
+            memorySize = _("Unknown");
+        }
+
+        let memoryRank = props[`MEMORY_DEVICE_${slot}_RANK`] || _("Unknown");
```
martinpitt, line 87: DRY: Drop the `||` here, as the `else` now sets this "Unknown" string anyway. Avoids the strings going out of sync, and calling gettext twice.

test/verify/check-system-info
```diff
@@ -559,35 +507,50 @@ revision\t: 2.3 (pvr 004e 1203)
         b.enter_page('/system/hwinfo')
         b.wait_in_text('#hwinfo-system-info-list .hwinfo-system-info-list-item:nth-of-type(2) .pf-c-description-list__group:nth-of-type(1) dd', "2x POWER9 (architected), altivec supported")
 
-        # Memory details should be shown for generic QEMU memory
-        # demidecode not available on certain images
-        if not m.ostree_image:
+        # Memory details should be shown from our mocked DMI information from systemd's test files.
+        # fedora-coreos does not have dmidecode
+        if m.image != "fedora-coreos":
```
martinpitt, line 512: NB that `m.ostree_image` is currently exactly the same as `m.image == "fedora-coreos"`. We used to have 3 OSTree based OSes, but right now just one.

Why don't these three checks work on FCOS? isn't that the same code path as for the technology/rank? I originally really meant to *drop* the `if` completely, not rewriting the if condition.

test/verify/check-system-info
```diff
@@ -631,7 +594,7 @@ revision\t: 2.3 (pvr 004e 1203)
         spoof_threads(2, False, cmdline='param1 nosmt=someunknown param3=value3')
         spoof_threads(2, True, self.expect_smt_default, None)
 
-    @skipImage("TODO: add Arch Linux grub entry support", "arch")
+    @ skipImage("TODO: add Arch Linux grub entry support", "arch")
```
martinpitt, line 597: spurious space

martinpitt (changes requested):
Thanks! Some remaining tiny nitpicks, but really close!

test/verify/check-system-info
jelly: This is indeed true, and now the if is removed \o/

martinpitt (approved):
Beautiful, dank je wel!

PR #17082, opened 2022-03-03
test: Set browser window size when first loading a page

mvollmer:
And not immediately when creating the Browser object.  This avoids
having a "self.cdp" that looks valid but can nevertheless not take
snapshots.


martinpitt (approved):
Nice and clean, thank you!

PR #17083, opened 2022-03-04
common: simplify closefrom() emulation hack

allisonkarlitskaya:
All of our CI images now have either closefrom() in the libc or
close_range() in the kernel (without libc wrapping).

Instead of shipping our own implementation of closefrom() in a separate
C file for the case where it's not available in the libc, move a very
small static inline wrapper into the cockpithacks header itself.

This prevents compiling cockpitclosefrom.c 3 times (for each place it's
used from), even in the case where it's unused.


martinpitt (dismissed):
If that goes green, niiice!

allisonkarlitskaya:
Looks like we're about seven months too early :(

martinpitt:
[does not work on ubuntu-2004](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17083-20220304-114414-ea4b99bd-ubuntu-2004/log.html). 22.04 LTS will be released in about six weeks, then we can stop supporting 20.04.

martinpitt:
Also fails valgrind checks. Supposedly these could already be sorted out, then this would only block on dropping ubuntu-2004? Moving to draft for now.

allisonkarlitskaya:
The valgrind check fails are because we have `_NR_closefrom` in the container we use for unit tests, but the kernel on the runner doesn't support it, so we get ENOSYS at runtime, and bomb out with memory still allocated.

So in any case, we need to wait until github updates its runners to Ubuntu 22.04 (hopefully within a month or so).

The main goal of this patch is to get this extra `.c` file out of the build for `pam_ssh_add`.  I see three ways that we could proceed on that front, without waiting for new Ubuntu:

 - drop `pam_ssh_add`.  I have no idea why we ship this.  It seems pretty dubious.  There are also third-party pam modules like [`pam_ssh`](http://pam-ssh.sourceforge.net/) which address this use case independently.
 - inline the `/proc/self/fd`-iteration code into the hacks header.  Note: I think we can kill the rlimit alternative here: `closefrom()` apparently lacks this fallback (according to OpenSSH release notes I recently read), and nobody seems to have cared for Cockpit.
 - use pam's own library version of this functionality: `pam_modutil_sanitize_helper_fds()`: this sets stdin/out/err to known state (but can skip this part with `PAM_MODUTIL_IGNORE_FD`) and then closes all fds 3 and above.

allisonkarlitskaya:
@mvollmer @martinpitt feedback appreciated

martinpitt:
I didn't realize that this was actually blocking some further work, instead of just being some "nice to have" cleanup.

> wait until github updates its runners to Ubuntu 22.04 (hopefully within a month or so).

Don't hold your breath for it. If the time it took for 20.04 to be available in gh is any indication, it's more like a year.

> use pam's own library version of this functionality

That may work for the close_range() usage in cockpit-session and pam-ss-add, but I wouldn't want to link e.g. cockpit-ws and cockpit-bridge (cockpitpipe.c user?) to PAM just for that.

> inline the /proc/self/fd-iteration code into the hacks header

Acceptable IMHO

> drop pam_ssh_add

I don't know the history of this either, but doc/key-auth.md documents it and check-shell-keys/check-shell-multi-machine-key test it. I wouldn't want to kill it en passant without some more historic research.

martinpitt:
> use pam's own library version of this functionality

Nevermind, I see PR #17087 -- that is totally fine, of course. I misunderstood before.

allisonkarlitskaya:
So we're probably going to want to do something like this at some point, but this PR is gonna conflict pretty badly with what's going to come next (#17087 and following), so let's close it and try again once Ubuntu 20.04 is gone from our CI and from GitHub.

Issue #9047, opened 2018-04-26
firewall: not clear that firewall is open to outgoing traffic


andreasn:
This came up during the [usability testing](https://github.com/cockpit-project/cockpit/wiki/Usability-test-Firewall).
It's not clear that the firewall allows all outgoing traffic, and that the ports listed in the table are only for incoming traffic. This was voiced as an issue by 50% of the participants (3 out of 6)

garrett:
Original design had a message about incoming requests being blocked by default. We could bring that back and expand upon it.

![original firewall design](https://user-images.githubusercontent.com/10246/151836287-35d8078f-2540-4735-85bf-cb5b9e21ece0.png)

It would look like this (quick mockup where only the title is changed from what's in the cockpit-design repo):

![recent firewall design, modified for the text at the top](https://user-images.githubusercontent.com/10246/151837240-10abaabc-d56b-4938-b288-7b7a03263757.png)

That would be a "quick fix" (relative to now, not how long this bug has been open).

We could, eventually, add support for outgoing. Since it'd be tied to zones, I think the only* way where it'd make sense would be to display it like this (revised to dump the trash icons and move dangerous and secondary functionality to kebab menus)]:

![recent firewall design adjusted a bit further, for the future, to include incoming/outgoing per zone](https://user-images.githubusercontent.com/10246/151839044-f22bc1b8-0436-4a8c-a59a-cf9ba1be9408.png)

\* (I've considered other methods, but they all require mode switching or grouping incoming and outgoing. Especially when zones are concerned, it's important to be able to get an overview of the firewall, so the above mockup is probably the only way to show the concepts that make sense.)

garrett:
If we did decide to handle outgoing traffic, we might want to build upon the concept above (last of the mockups) we might want to:

1. Change the text at the top.
2. Make outgoing not blocked by default actually, unless that's a setting in the zone dialog. (I guess it should be a block by default setting.)
3. Change messaging to be in the incoming/outgoing perhaps? It'd be text to the right of the sub-sub-headings.
4. Start with non-editable, already-included services and just message what's included in the zone.

I had a mockup that illustrated this, but Inkscape crashed and it's already getting late here today. So we'll have to discuss it... and if it's a good idea, I'm happy to mock it up again.

Meanwhile, we probably want to consider mockup # 2 above, with the addition of the text at the top.

jelly:
[...]

PR #17084, opened 2022-03-04
Firewalld redesign

jelly:
Firewall page improvements from https://github.com/cockpit-project/cockpit/issues/9047#issuecomment-1026005844

### DIFF
--- a/pkg/networkmanager/firewall-switch.jsx
+++ b/pkg/networkmanager/firewall-switch.jsx
@@ -18,7 +18,7 @@
  */
 import cockpit from "cockpit";
 import React from "react";
-import { Switch, Tooltip } from "@patternfly/react-core";
+import { HelperText, HelperTextItem, Switch, Tooltip } from "@patternfly/react-core";
 
 const _ = cockpit.gettext;
 
@@ -71,6 +71,13 @@ export class FirewallSwitch extends React.Component {
                                     onChange={this.onSwitchChanged}
                                     aria-label={enabled ? _("Disable the firewall") : _("Enable the firewall")} />;
         }
-        return firewallOnOff;
+        return (
+            <>
+                {firewallOnOff}
+                <HelperText>
+                    <HelperTextItem variant="indeterminate">{enabled ? _("Enabled") : _("Disabled")}</HelperTextItem>
+                </HelperText>
+            </>
+        );
     }
 }
--- a/pkg/networkmanager/firewall.jsx
+++ b/pkg/networkmanager/firewall.jsx
@@ -170,18 +170,20 @@ function ZoneSection(props) {
     return <Card className="zone-section" data-id={props.zone.id}>
         <CardHeader className="zone-section-heading">
             <CardTitle>
-                <Flex alignItems={{ default: 'alignItemsCenter' }}>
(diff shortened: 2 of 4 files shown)


garrett:
Is this mainly for converting the delete icon to the ... kebab menus? (That's what it looks like here when I build it locally.)

Will it also include
- the message about incoming traffic at the top?
- changing the toggle to include "enabled"

Basically this:

image
...or should that be a separate PR?

garrett:
Mobile needs work:

image

After thinking about it for a bit, can we have the incoming message at the top and the enabled switch in a separate PR to fast track those changes? It'd allow us to make sure we get this PR right while having simpler changes go through more quickly.

garrett:
The text should probably be aligned to the baseline.

Screenshot from 2022-03-10 14-23-37

The padding between items feels wrong and it should really be gap to fix the issues in desktop and mobile wrapping. I think I'll probably have to dig into the code to adjust this.

jelly:
> Mobile needs work:
> 
> image
> 
> I see if we do add "Enabled", it'd cause the button to wrap. Perhaps we should add the incoming/outgoing string and not "Enabled" / "Disabled".
> 
> Horizontal alignment is off. It's inconsistent between the header and table, but more importantly, the interface is way over to the right. And there's probably too much space between the Zone header and Interface. We should align these with gaps, not margin.
> 
> Screenshot from 2022-03-10 14-25-28
> 
> After thinking about it for a bit, can we have the incoming message at the top and the enabled switch in a separate PR to fast track those changes? It'd allow us to make sure we get this PR right while having simpler changes go through more quickly.
> 
> The incoming message at the top is mean to be a placeholder until we have incoming and outgoing traffic listed. This PR has incoming traffic and doesn't handle outgoing (as far as I can see).
> 
> But I think we'll want to work on this PR and it might not make it in for the next release. So I'm fine with adding some text in a quick PR and adjusting the UI later. (I don't consider adding the text as much UI churn, as it's just adding a string and not really changing the UI.)
> 
> For desktop:
> 
> The text should probably be aligned to the baseline.
> 
> Screenshot from 2022-03-10 14-23-37
> 
> The padding between items feels wrong and it should really be gap to fix the issues in desktop and mobile wrapping. I think I'll probably have to dig into the code to adjust this.

Switched this to use PF's Flex so we can get rid of all css classes and no gap seems to be required.

image

garrett:
> should this also appear on the network page, it also has a firewall switch? 

Yes.

The issue was originally with the services page in the metrics study, which needs this as well. (However, that also needs some additional improvements too.) People generally did not see the Firewall page there (but the same principle of discoverability of the switch applies).

> And what element is the small text again? It's not a as far as I know.

Helper text?

- https://www.patternfly.org/v4/components/helper-text/

If helper text doesn't work for whatever reason, it could just be text with a small font size helper applied:

- https://www.patternfly.org/v4/components/text/
- https://www.patternfly.org/v4/utilities/text/#font-size

garrett (approved):
Looks great! :+1: 

I did notice it's still not quite right in smaller modes, but it's still better than what we have in main at the moment, so we can optimize things in a follow-up.

Screen Shot 2022-03-16 at 11 40 34


Screen Shot 2022-03-16 at 11 41 01

PR #17085, opened 2022-03-06
overview: Use standard PF list

marusak:
Before:
before

After:
Screenshot from 2022-03-07 09-59-26


pkg/systemd/overview-cards/configurationCard.jsx
```diff
@@ -186,14 +187,19 @@ class SystemInformationSshKeys extends React.Component {
...
-                        <h4>{key.title}</h4>
-                        {key.fps.map((fp, i) => <div key={i}><small>{fp}</small></div>)}
-                    </div>
-                )}
-            </div>;
+            body = <ListingTable aria-label={_("SSH key fingerprints")}
+                    variant="compact"
+                    columns={[""]}
+                    rows={this.state.keys.map(key => {
+                        return {
```
martinpitt, line 194: If you have to force-push for some reason (pixel tests?), please `k => { return v}` → `k => v`. (Otherwise, not important enough to re-run everything_

martinpitt (approved):
Nice cleanup, thanks!

pkg/systemd/overview-cards/configurationCard.jsx
marusak: I did this because I need to return object. I guess there is way how to do it, but I always used `return` when I was returning object. Any help with the syntax as doing `k => {foo: "bar"}` does not work.

martinpitt:
I already did a round of retries this morning, but [more](https://logs.cockpit-project.org/logs/pull-17085-20220307-060841-8a401082-debian-stable/log.html) [flakes](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17085-20220307-060838-8a401082-fedora-coreos/log.html)

marusak:
> I already did a round of retries this morning, but [more](https://logs.cockpit-project.org/logs/pull-17085-20220307-060841-8a401082-debian-stable/log.html) [flakes](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17085-20220307-060838-8a401082-fedora-coreos/log.html)

And broken pixel tests - I am not sure how is that related to these changes, but looking

pkg/systemd/overview-cards/configurationCard.jsx
martinpitt: `k => ({ foo: "bar" })` works. Similar to destructuring in arrow arguments.

martinpitt (changes requested):
Indeed this [pixel test failure](https://logs.cockpit-project.org/logs/pull-17085-20220307-060840-8a401082-fedora-35/pixeldiff.html#TestSystemInfo-testBasic-overview-mobile) shows that it breaks the Usage card.

pkg/systemd/overview-cards/configurationCard.jsx
```diff
@@ -27,6 +27,7 @@ import cockpit from "cockpit";
 import { superuser } from "superuser";
 import { useObject, useEvent } from "hooks.js";
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
+import { ListingTable } from "cockpit-components-table.jsx";
```
marusak, line 30: This breaks the pixel text. It pulls in some table styles and the usage cards also use table but not with this styling. I'll see what I can do -hopefully just import it later? :crossed_fingers:

marusak: Ok, so it was very tricky to fix and would lead down a rabbit hole, so I used `<List>` instead.

martinpitt (dismissed):
Cheers! LGTM (assuming green)

marusak:
every TF failed with `configure: error: the xmlto command was not found (perhaps --disable-doc)` :thinking:

martinpitt:
@marusak : packit was just fixed in 6231d1c675ab065b5a249b91ce1f9b5d16afde63

martinpitt:
Twe two failures are flakes from 3x affected retries, so retrying.

martinpitt (approved):

garrett (approved):
This is a lot nicer. :+1:

PR #17086, opened 2022-03-07
systemd: Fix Service details crash on disappearing units

martinpitt:
These happen after `systemctl daemon-reload` when changing unit files,
or when stopping transient units (systemd-run). Then the page would
crash with

    Cannot read properties of undefined (reading 'Names')

Intercept this condition and show the "Unit not found" empty state.

---

This can be seen in current logs, e.g. [here](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17082-20220303-091332-1070fb01-fedora-35/log) in `TestCurrentMetrics.testCPU`.


marusak (approved):
Thanks!

PR #17087, opened 2022-03-07
pam-ssh-add: drop two uses of libcockpit-common APIs

allisonkarlitskaya:
This prevents us from recompiling libcockpit-common files as -fPIC.

### DIFF
--- a/src/pam-ssh-add/Makefile.am
+++ b/src/pam-ssh-add/Makefile.am
@@ -1,8 +1,6 @@
 noinst_LIBRARIES += libpam_ssh_add.a
 
 libpam_ssh_add_a_SOURCES = \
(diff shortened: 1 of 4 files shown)


allisonkarlitskaya:
This is the heavier-hitting version of #17083 (which is stuck because Ubuntu 20.04's kernel doesn't have `close_range()`).

martinpitt (approved):
Cheers! Nice to make this PAM module completely independent.

martinpitt:
[very common flake](https://logs.cockpit-project.org/logs/pull-17087-20220307-095055-49c7fcef-centos-8-stream/log.html#51-2) these days, I'll look at that next. Retrying.

PR #17091, opened 2022-03-07
test: Increase timeout for load to drop

martinpitt:
On some busy CI machines, 3 minutes is not enough for the load to drop
below 10, so wait a little more.

---

I've seen this [fail a lot](https://logs.cockpit-project.org/logs/pull-17087-20220307-095055-49c7fcef-centos-8-stream/log.html#51-2) today, and it's getting annoying. I can't reproduce the long wait times locally, it's usually done after ~ 1.5 minutes. But our CI machines are a lot more busy.


martinpitt:
Cool, passed 3x everywhere :partying_face:

marusak (approved):

PR #17092, opened 2022-03-08
ws: Don't cache po.js files forever

### LINKED ISSUE #8160: language change has no effect in chromium on our test images

martinpitt:
This causes at least Chromium to not ever re-request the file from the
webserver, so that switching languages does not have any effect.

This is a very heavy hammer, as it generally disables caching po.js even
when not switching languages (the common case). So this should be
replaced with a better design at some point.

Note that caching is already disabled if any Cockpit package is present
in the user's home directory, so this is not something completely new.

Fixes #8160

----

Draft for now. I want a full test run, and possibly also use it to test #16930

### DIFF
--- a/src/bridge/cockpitpackages.c
+++ b/src/bridge/cockpitpackages.c
@@ -965,6 +965,7 @@ package_content (CockpitPackages *packages,
   GBytes *bytes = NULL;
   gboolean globbing;
   gboolean gzipped = FALSE;
+  gboolean is_language_specific = FALSE;
   gboolean allow_gzipped = FALSE;
   const gchar *type;
   gchar *policy;
@@ -1029,7 +1030,7 @@ package_content (CockpitPackages *packages,
 
       g_clear_error (&error);
 
-      bytes = cockpit_web_response_negotiation (filename, package ? package->paths : NULL, language, NULL, &gzipped, &error);
+      bytes = cockpit_web_response_negotiation (filename, package ? package->paths : NULL, language, &is_language_specific, &gzipped, &error);
 
       /* When globbing most errors result in a zero length block */
       if (globbing)
@@ -1039,6 +1040,7 @@ package_content (CockpitPackages *packages,
               g_message ("%s", error->message);
               bytes = g_bytes_new_static ("", 0);
               gzipped = FALSE;
+              is_language_specific = FALSE;
             }
         }
       else
@@ -1069,6 +1071,12 @@ package_content (CockpitPackages *packages,
               goto out;
             }
         }
+
+      /* If the response is language specific, don't cache the file. Caching "po.js" breaks
+       * changing the language in Chromium, as that does not respect `Vary: Cookie` properly.
+       * See https://github.com/cockpit-project/cockpit/issues/8160 */
+      if (is_language_specific || globbing)
+        cockpit_web_response_set_cache_type (response, COCKPIT_WEB_RESPONSE_NO_CACHE);
 
       /* Do we need to decompress this content? */
       if (gzipped && !allow_gzipped)
--- a/src/bridge/cockpitpackages.c
+++ b/src/bridge/cockpitpackages.c
@@ -963,9 +963,8 @@ package_content (CockpitPackages *packages,
   gchar *filename = NULL;
   GError *error = NULL;
   GBytes *bytes = NULL;
-  gchar *chosen = NULL;
   gboolean globbing;
-  gboolean gzipped;
+  gboolean gzipped = FALSE;
   gboolean allow_gzipped = FALSE;
   const gchar *type;
   gchar *policy;
@@ -1029,10 +1028,8 @@ package_content (CockpitPackages *packages,
         g_bytes_unref (bytes);
 
       g_clear_error (&error);
-      g_free (chosen);
-      chosen = NULL;
 
-      bytes = cockpit_web_response_negotiation (filename, package ? package->paths : NULL, language, &chosen, &error);
+      bytes = cockpit_web_response_negotiation (filename, package ? package->paths : NULL, language, NULL, &gzipped, &error);
 
       /* When globbing most errors result in a zero length block */
       if (globbing)
@@ -1040,8 +1037,8 @@ package_content (CockpitPackages *packages,
           if (error)
             {
               g_message ("%s", error->message);
-              chosen = g_strdup ("");
               bytes = g_bytes_new_static ("", 0);
+              gzipped = FALSE;
             }
         }
       else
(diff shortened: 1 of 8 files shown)


martinpitt:
Also, I really think we should lower `max-age=31556926` to 86400. If browsers would actually heed this absurdly long timeout, we would never be able to recover after messing up the caching. @marusak , @allisonkarlitskaya , WDYT? (Happy to send this as a separate PR)

src/ws/cockpithandlers.c
```diff
@@ -567,6 +567,9 @@ handle_resource (CockpitHandlerData *data,
         {
           cockpit_channel_response_serve (service, headers, response, where,
                                           cockpit_web_response_get_path (response));
+          /* po.js is magic, it loads po.LL.js depending on current client language */
+          if (g_str_has_suffix (path, "/po.js"))
```
allisonkarlitskaya, line 571: I don't disagree with the logic, but this is the wrong place for it.

Let's rather put it in the bridge, somewhere near where cockpitpackages calls into `cockpit_web_response_negotiation()`, conditional on it actually having used the language name to find the file.

allisonkarlitskaya (changes requested):

src/ws/cockpithandlers.c
martinpitt: Ack, will do, thanks for the hint! I disagree with the logic as well, as it's just a horrible hack around this caching maze + weird magic around loading "po.js". I hope we find a better solution still, but if we get stuck again (like we did for the last 5 years), we may have to bite the bullet.

I'd still like to hammer a bit on [changing the package checksum for a different locale](https://github.com/cockpit-project/cockpit/issues/8160#issuecomment-1061646262). This fails, and I don't yet completely understand why.

martinpitt: Reworked to be in package_content(). That required quite a bit more work, but the initial hack at least showed that this does work in principle.

martinpitt:
Ugh, this worked fine in my interactive testing, but some failures. I can understand the [distropkg failure](https://logs.cockpit-project.org/logs/pull-17092-20220308-143056-01bb053d-rhel-8-6-distropkg/log.html), we need a quirk in the test for that.

But this also fails on Debian [stable](https://logs.cockpit-project.org/logs/pull-17092-20220308-143056-01bb053d-debian-stable/log.html#295) and [testing](https://logs.cockpit-project.org/logs/pull-17092-20220308-143056-01bb053d-debian-testing/log.html). Back to draft.

martinpitt:
Rebased to pick up packit fix.

martinpitt:
@marusak : I think @allisonkarlitskaya wants to review this, but FYI as you worked on this topic as well. WDYT about this compromise?

marusak:
I will leave the full review to Lis, just a question - So we won't cache any po.js ever. Does this have any significant impact on traffic and how much data is send over the wire?

martinpitt:
> Does this have any significant impact on traffic and how much data is send over the wire?

A little bit, yes -- that's my concern. #16930 will significantly reduce the overall data (and also allow us to simplify the bridge), but it's blocked on fixing this issue somehow. The net result is a bit unclear -- currently every page loads ~ 10 po files instead of 2, but at least they get cached.

allisonkarlitskaya (commented):
Does this definitely fix #16930?

If so, then it's a bitter pill to swallow, but I think it's worth it for the time being...

Ultimately, this piece of code needs to die and be replaced with something much more explicit:

```
    /*
     * This is how we find out about the frontends cockpitlang
     * environment. We tell this process to update its locale
     * if it has changed.
     */
    cockpit_locale_set_language (languages[0]);
```

src/common/cockpitwebresponse.c
```diff
@@ -1623,6 +1626,7 @@ cockpit_web_response_negotiation (const gchar *path,
   gchar *shorter = NULL;
   gchar *lang = NULL;
   gchar *lang_region = NULL;
+  gboolean _is_lang, _is_compressed;
```
allisonkarlitskaya, line 1629: ugh.

If you want to do this, the usual convention is rather to use `out_is_language_specific` as the function parameter and call the variable `is_language_specific`.

allisonkarlitskaya (changes requested):
One change please.  This `_` thing is a bit of a no-go for me.

Strictly speaking, as a non-external identifier, it's not a reserved name, but that's a pretty thin technicality.

src/bridge/cockpitpackages.c
```diff
@@ -1073,15 +1072,19 @@ package_content (CockpitPackages *packages,
             }
         }
 
+      /* If the response is language specific, don't cache the file. The only known case is "po.js",
+       * caching breaks changing the language in the browser. */
```
allisonkarlitskaya, line 1076: This comment could also benefit from a bit more exposition on the topic.  At least a link to the relevant issue would be nice.

allisonkarlitskaya (commented):
And please please please rebase this :)

martinpitt:
On second thought, the impact of this isn't actually that bad. The po files only get loaded the first time a page gets opened -- i.e. when cockpit loads the page's index.html. After that, the page (frame) stays loaded, and all the action happens through JS DOM manipulation. I.e. the `<script po.js>` only runs at most once per cockpit session, which is okay-ish IMHO.

I measured this in the networking inspector. Lis and I  *really* want PR #16930 to clean this up, so I measured this and that PR together, against main. I logged in, switched to German, and then opened the Networking page and hwinfo.

On main it basically loads the combined translations once at login, and all other pages re-use the same cache:

main

And with this PR every new page loads its own po file:

pr

After that you can switch between pages or click around without new loads.

So there *is* an impact, but it feels like it's bearable to get a more correct behaviour?

src/bridge/cockpitpackages.c
martinpitt: I changed it, WDYT?

allisonkarlitskaya (approved):
:+1: thanks!

Issue #8610, opened 2018-02-13
Update stale known issues


cockpituous:
Check for any stale known issues and open a pull request to close them
 * [ ] FAIL: naughty-prune

cockpituous:
naughty-prune in progress on cockpit-tasks-h8lh3.
Log: http://fedorapeople.org/groups/cockpit/logs/naughty-prune-8610-20180213-142727/

cockpituous:
naughty-prune in progress on cockpit-tasks-1tg4w.
Log: http://fedorapeople.org/groups/cockpit/logs/naughty-prune-8610-20180213-142727/

cockpituous:
Command '('git', 'push', 'https://github.com/cockpituous/cockpit', '+HEADNone/heads/naughty-prune-8389-20180213-142727')' returned non-zero exit status 1.: http://fedorapeople.org/groups/cockpit/logs/naughty-prune-8610-20180213-142727/

cockpituous:
naughty-prune in progress on cockpit-tasks-1tg4w.
Log: http://fedorapeople.org/groups/cockpit/logs/naughty-prune-8610-20180213-142727/

cockpituous:
Task threw an exception: http://fedorapeople.org/groups/cockpit/logs/naughty-prune-8610-20180213-142727/

PR #17093, opened 2022-03-08
ws: Lower cache time-to-live to one day

martinpitt:
Treat `COCKPIT_WEB_RESPONSE_CACHE_FOREVER` as "one day". Otherwise, if
browsers actually would take the value literally (which they don't,
fortunately), we would never be able to clean up after responses with
wrong cache parameters. The most prominent case for this is #8610
(serving `po.js`).


martinpitt:
The [packit/c8s failure](https://artifacts.dev.testing-farm.io/fe0e566a-ce5e-4bd0-b991-718a0ff430f9/) is unrelated, it started to happen this morning (broken c8s update?). We see it in other PRs as well, e.g. #17067

allisonkarlitskaya (commented):
imho, our use of "public" here is questionable.  I'd rather we fold the number of possibilities to two (yes or no) and use "24h private" for the yes case.

martinpitt:
@allisonkarlitskaya : Happy to do that. Follow-up, or do it in one commit?

martinpitt:
Replaced by PR #17100. Agreed that it makes more sense to clean this up more thoroughy.

Issue #10059, opened 2018-09-12
storage: hard to notice a new volume group (or generally another resource, see lvolumes) that was just created


andreasn:
During a usability test on the storage page, where the participants were asked to create a volume, a lot of participants were able to create a volume group, but then failed to find the item they had created in the list.

garrett:
Here's an HTML mockup of a possible animation to help with this issue: https://garrett.github.io/cockpit-mockweb/animation/new-row/

It uses opacity, grows in, has a highlight color and the highlight color eventually fades out. It uses squash & stretch principles (from traditional animation) and eased timing.

All one needs to do is add a class of "new" to the tbody. (I could make it work for a `<tbody>` or `<tr>`.)

garrett:
Note: This should apply to all of Cockpit — not just volume groups. We should probably have an animations.css file that gets pulled into the base CSS.

andreasn:
Yeah, there was an issue with participants having a hard time locating newly created logical volumes, so I think that this fix could apply there as well.

andreasn:
I get a 404 on https://garrett.github.io/cockpit-mockweb/animation/new-row/ now it seems

garrett:
Odd. It works @ https://garrett.github.io/cockpit-mockweb/animation/new-row (without the slash)... did GitHub change how Jekyll sites work?

garrett:
I need to get the CSS into a library file that's used across the site, perhaps in an `animations.less` or `animations.css` file or something like that. And then the JS across the site needs to be (slightly*) adapted to use the new CSS.

_\* slightly: New rows would need to have the class of `new` (or `ct-new` or `animate-new` or whatever we decide._

JeffMoyer:
I think the color is a bit too subtle.  Other than that, I like it, and would find it useful.

KKoukiou:
@garrett would it be easy to paste here the ct-new class content, and then I can work on the JS side to implement it.

KKoukiou:
so @garrett just let me know that PF is working on the design for it https://github.com/patternfly/patternfly-design/issues/985

garrett:
I moved the repo from master to main a week or two ago and told GitHub to that the new branch to build from is main. However, GitHub never rebuilt it, so the entire repo was 404'd.

In the hopes that GitHub would rebuild it, I added a few minor commits. It finally did, so the links work again. Sorry about the breakage.

[...]

PR #17094, opened 2022-03-08
lib: Animate new items in ListingTable

marusak:
## Animate new rows in lists

When new items appear on a page, they are highlighted for a short while to spot them more easily.

Peek 2022-03-15 07-47

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -17,7 +17,7 @@
  * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import {
     ExpandableRowContent,
     TableComposable, Thead, Tbody, Tr, Th, Td,
@@ -75,8 +75,36 @@ export const ListingTable = ({
     let rows = tableRows;
 
     const [expanded, setExpanded] = useState({});
+    const [newItems, setNewItems] = useState([]);
+    const [currentRowsKeys, setCurrentRowsKeys] = useState([]);
     const [activeSortIndex, setActiveSortIndex] = useState(sortBy ? sortBy.index : 0);
     const [activeSortDirection, setActiveSortDirection] = useState(sortBy ? sortBy.direction : SortByDirection.asc);
+
+    useEffect(() => {
+        const getRowKeys = rows => {
+            const keys = [];
+
+            rows.forEach(row => {
+                if (row.props && row.props.key)
+                    keys.push(row.props.key);
+            });
+
+            return keys;
+        };
+
+        const current_keys = getRowKeys(rows);
+        if (JSON.stringify(current_keys) === JSON.stringify(currentRowsKeys))
+            return;
+
+        // Don't highlight all when the list gets loaded
+        if (currentRowsKeys.length !== 0) {
+            const new_keys = current_keys.filter(key => currentRowsKeys.indexOf(key) === -1);
+            if (new_keys.length)
+                setNewItems([...newItems, ...new_keys]);
+        }
+
+        setCurrentRowsKeys(current_keys);
+    }, [rows, currentRowsKeys, newItems]);
 
     const isSortable = cells.some(col => col.sortable);
     const isExpandable = rows.some(row => row.expandedContent);
@@ -147,6 +175,9 @@ export const ListingTable = ({
             rowProps.isHoverable = true;
             rowProps.onRowClick = (event) => onRowClick(event, row);
         }
+
+        if (rowProps.key && newItems.indexOf(rowProps.key) >= 0)
+            rowProps.className = (rowProps.className || "") + " new-item-ct";
 
         const rowKey = rowProps.key || rowIndex;
         const isExpanded = expanded[rowKey] === undefined ? !!row.initiallyExpanded : expanded[rowKey];
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -75,6 +75,50 @@ a.disabled:hover {
     width: 100%;
 }
 
+/* Animation of new items */
+.new-item-ct {
+    animation: newRowCt 4s ease-in;
+}
+
+:root {
+    --ct-animation-new-background: #fdf4dd;
+}
+
+/* Simple fade animation by default (mainly for reduced motion) */
+@keyframes newRowCt {
+  0% {
+    background-color: var(--ct-animation-new-background);
+  }
+  50% {
+    background-color: var(--ct-animation-new-background);
+  }
+}
+
+/* Redefine to a fancy animation when reduced-motion is not requested */
+@media (prefers-reduced-motion: no-preference) {
+  @keyframes newRowCt {
+    0% {
+      height: 0;
+      background-color: var(--ct-animation-new-background);
+      transform: scaleX(0) scaleY(0);
+      opacity: 0;
+    }
+    10% {
+      transform: scaleX(1) scaleY(1);
+    }
+    12% {
+      transform: scaleX(1) scaleY(1.25);
+      opacity: 1;
+    }
+    15% {
+      transform: scaleX(1) scaleY(1);
+    }
+    70% {
+      background-color: var(--ct-animation-new-background);
+    }
+  }
+}
+
 /* Dialog patterns */
 
 .dialog-wait-ct {


pkg/lib/cockpit-components-table.scss
```diff
@@ -106,3 +106,28 @@
         border-bottom: 1px solid var(--ct-color-list-critical-border);
     }
 }
+
```
marusak, line 109: This might then go to some generic css file like page.css or so and then we can add the `.new-item` logic to other components and it would "just work" :tm:

pkg/lib/cockpit-components-table.jsx
```diff
@@ -75,9 +75,36 @@ export const ListingTable = ({
...
+
+            rows.forEach(row => {
+                if (row.props && row.props.key)
+                    keys.push(row.props.key);
+            });
+
+            return keys;
+        };
+
+        const new_keys = getRowKeys(rows);
```
marusak, line 95: This needs to be checked if it does not misbehave on any place where we use `ListingTable`. And it only works if the component specifies `props.keys` - but they should as it is a good practice if nothing else.
But I think that we might condition this by some property like `animateNewItems` so that we don't enable it by default everywhere.

marusak:
Without any other changes parts of storage page are already working:
lvolumes

The side items ("Physical volumes", "Devices". "Drives"...) use `<SidePanel>` from `pkg/storaged/side-panel.jsx`. It should be possible to implement the same logic for it as well.

marusak:
> When visiting the SELinux page, the entries on screen are all marked as new.

Yeah, what I was worried about. As I mentioned above, we most likely want to enable this page by page and always make sure it works correctly. I briefly looked at selinux page and don't see any obvious issue (like using the same key), but I think it might be related how we fetch those alerts in the first place.

marusak:
Ah, the selinux is bit annoying - it reads all alerts without any details, renders then and then reads more info about them one by one and updates them. In this update a new key is generated. This whole seems a bit odd - both that we render it once and then update it item by item and that we generate it with different keys. If we decide to go with this PR I would just disable it on selinux page for now and try to fix this behaviour outside of this PR.

marusak:
@garrett I have fixed the issue on selinux page and also found one place that I also adjusted. In the first comment I dropped down list of all `ListingTable` occurrences and went through all of them.

I also moved the css parts into `pkg/lib/page.scss` so that we can re-use them in other components as well.

pkg/lib/page.scss
```diff
@@ -76,29 +76,46 @@ a.disabled:hover {
...
-  12% {
-    transform: scaleX(1) scaleY(1.25);
-    opacity: 1;
-  }
-  15% {
-    transform: scaleX(1) scaleY(1);
-  }
-  70% {
-    background: #fdf4dd;
+}
```
garrett, line 95: So it starts with the background color, then doesn't do anything for half the time then fades out over 2 seconds.

The one below does a bunch of other stuff, but uses the same technique.

I adjusted the timing because it doesn't need to accommodate the moving animation time and then it also makes the animation slightly more gradual.

(Prefers reduced motion is not meant to completely remove all animations. It's meant to have little to no movement and not have drastic blinking. Mild, slow movements are even OK, although I went with just a simple fade for reduced motion.)

All major OSes have means to remove motion, usually under accessibility. I tested this on Firefox, which acknowledges GNOME's animation setting (in accessibility).

garrett: Just explaining how this all works, especially on the default reduced motion, with more animation on the allows animation ('no preference") part.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -75,9 +75,36 @@ export const ListingTable = ({
...
+        if (JSON.stringify(new_keys) === JSON.stringify(currentRowsKeys))
+            return;
+        // Don't highlight all when the list gets loaded
+        if (currentRowsKeys.length !== 0) {
+            new_keys.forEach(key => {
+                if (currentRowsKeys.indexOf(key) === -1)
+                    setNewItems([...newItems, key]);
```
martinpitt, line 102: Is it wise to call a state set function that often? Shouldn't it rather do this once after the loop?

(And then use filtering on new_keys instead of forEach)

pkg/lib/cockpit-components-table.jsx
```diff
@@ -148,6 +175,13 @@ export const ListingTable = ({
             rowProps.onRowClick = (event) => onRowClick(event, row);
         }
 
+        if (rowProps.key && newItems.indexOf(rowProps.key) >= 0) {
+            if (rowProps.className)
+                rowProps.className = rowProps.className + " new-item";
```
martinpitt, line 180: This is a globally visible class name, and thus should have a name which is less prone to collisions. But see below.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -148,6 +175,13 @@ export const ListingTable = ({
...
+            else
+                rowProps.className = "new-item";
```
martinpitt, line 182: Please drop the `if` and use `(rowProps.className || "")` for simplicity.

pkg/lib/page.scss
```diff
@@ -75,6 +75,33 @@ a.disabled:hover {
     width: 100%;
 }
 
+/* Animation of new items */
+
+.new-item {
```
martinpitt, line 79: Does this really have to go into the global page.css? I'd much rather like to see this in pkg/lib/cockpit-components-table.scss, to avoid namespace collisions. Otherwise please add a `ct-table` to the name. Same for `newRow`.

martinpitt (changes requested):
Thanks, this is really cool! Animated transitions make it so much easier to see what's going on.

pkg/lib/page.scss
marusak: I originally had it in `pkg/lib/cockpit-components-table.scss` but moved it to `pkg/lib/page.scss` as we want to use the same pattern on different components as well. As a good example see https://github.com/cockpit-project/cockpit/issues/10059 where we should also add animation but it does not use `ListingTable`. I purposefully want to create this as a global item so that we can just add this class into any page into any element and it just works the same way.

pkg/lib/cockpit-components-table.jsx
marusak: Any suggestions? `new-item-animation` or you want to go one step further and add `cockpit` into mix as well e.g. `cockpit-item-animation`?

pkg/lib/page.scss
martinpitt: I see, thanks! So let's please at least add a `-ct` suffix

pkg/lib/cockpit-components-table.jsx
martinpitt: `new-item-ct` seems good enough for now?

marusak:
@garrett thank you! looks good and I learned about the `prefers-reduced-motion` option as well

@martinpitt I have updated all your comments.

I also updated the gif in the first comment for the current changes. Not sure if we can have gif in release notes though.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -75,9 +75,37 @@ export const ListingTable = ({
...
+
+        const current_keys = getRowKeys(rows);
+        if (JSON.stringify(current_keys) === JSON.stringify(currentRowsKeys))
+            return;
+
+        // Don't highlight all when the list gets loaded
+        if (currentRowsKeys.length !== 0) {
+            const new_keys = current_keys.filter(key => currentRowsKeys.indexOf(key) === -1);
+            if (new_keys.length)
+                setNewItems([...newItems, ...new_keys]);
```
martinpitt, line 103: This only ever gets added to, instead of just showing the actually new keys once rows changes. I know the animation will eventually fade out, but over time it would mean that all rows are "new"? Why not `setNewItems(new_keys)` here?

martinpitt (commented):
Thanks! One more question, but this is likely just my ignorance. (And will flip to +1 after answer)

pkg/lib/cockpit-components-table.jsx
marusak: Mostly I am worried that when two items would be added in quick succession that it would remove this new class from the freshly added item. I haven't tested what happens when you remove class mid animation but leaving the class there has no negative impact.

martinpitt (approved):
Ack, thank you! Can you still prepare the release-notes bits, please?

martinpitt (approved):

PR #17097, opened 2022-03-08
tls: Add hostname to certificate alt subject also for openssl

mvollmer:
This makes the OpenSSL created certificates more similar to the sscg
created ones.

### DIFF
--- a/src/tls/cockpit-certificate-helper.in
+++ b/src/tls/cockpit-certificate-helper.in
@@ -59,7 +59,7 @@ selfsign_openssl() {
     distinguished_name = req_distinguished_name
     [ req_distinguished_name ]
     [ v3_req ]
-    subjectAltName=IP:127.0.0.1,DNS:localhost
+    subjectAltName=IP:127.0.0.1,DNS:localhost,DNS:${HOSTNAME}
     basicConstraints = critical, CA:TRUE
     keyUsage = critical, digitalSignature,cRLSign,keyCertSign,keyEncipherment,keyAgreement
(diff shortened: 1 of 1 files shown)


mvollmer:
So, I don't know if this is a good idea or not, but here is the story:

- sscg with OpenSSL 3 is broken for us: https://bugzilla.redhat.com/show_bug.cgi?id=2061923
- this means we create the self signed certs with OpenSSL
- but they don't work with the mock-insights server since they don't have a 0-self-signed-ca.pem file
- but one might use 0-self-signed.cert as the CA file
- except it doesn't have the real hostname in its alt subject (while sscg puts it there)

So all will be well again once sscg starts working with OpenSSL 3 and we don't need this at all.

Or I make a dedicated set of certs for mock-insights and avoid this issue that way. That would be a good idea in general, I guess.

mvollmer:
Let's make this a draft until I know whether I really want this.

martinpitt:
Ugh -- that reminds me that we don't have any test which validates that sscg works. We just test that "something" can generate a working certificate. sscg is also in Debian testing and will be in the next Ubuntu LTS, so at some point we actually wanted to rely on it and drop our own openssl bits.

mvollmer:
> Or I make a dedicated set of certs for mock-insights and avoid this issue that way. That would be a good idea in general, I guess.

This has been done: https://github.com/candlepin/subscription-manager/pull/2998

mvollmer:
I am abandoning this. I don't want to have any opinion about our self signed certificates, if possible. :-)

PR #17099, opened 2022-03-09
packit.yaml: reduce the srpm depends

allisonkarlitskaya:
The list here contains very many dependencies that are either no longer
required by this project to build an SRPM (xmlto, selinux-policy, npm)
or never were (the unrelated "autogen" package).

automake pulls in autoconf as a hard dependency central to its
functionality, but gcc's dependency on make seems a bit less central, so
list it explicitly.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -10,16 +10,11 @@ actions:
     - make XZ_OPT=-0 dist
     - sh -c 'echo cockpit-*.tar.xz'
 srpm_build_deps:
-  - npm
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Future nice-to-have: ability to `make dist` without:
 - glib or systemd headers (more mock env `.pc` hacking?)
 - without a C compiler?

martinpitt (approved):
Danke!

PR #17100, opened 2022-03-09
ws: Eliminate eternal caching

martinpitt:
Drop `COCKPIT_WEB_RESPONSE_CACHE_FOREVER`. If browsers actually would
take the value literally (which they don't, fortunately), we would never
be able to clean up after responses with wrong cache parameters. The
most prominent case for this is #8610 (serving po.js).

Likewise, branding images also change occasionally, as they are symlinks
to OS provided files.

Thus move branding and ETag resources to "24 hour private" caching, and
drop the "forever" mode.


allisonkarlitskaya (commented):
I agree with the change.

Knowing your personal sensibilities, I don't dare ask you to change this into a boolean.

That said, maybe we could make the "private" name more generic, since it doesn't stand in contrast to "public" anymore?  Something like `_WEB_RESPONSE_CACHED` vs `_WEB_RESPONSE_UNCACHED` or so?

martinpitt:
@allisonkarlitskaya : Agreed, good idea. Renamed it to just `_CACHE`. And yes, I like self-describing enum names more than booleans :-)

allisonkarlitskaya (approved):
:+1:

PR #17101, opened 2022-03-09
packit.yaml: stop jumpstarting

allisonkarlitskaya:
Now that we build the SRPM on COPR instead of in the sandcastle, the
previous resource limitations no longer apply.

Add nodejs to our build dependencies and run the webpack build in-place.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -4,9 +4,6 @@ actions:
     - tools/create-spec --version 0 --build-all -o cockpit.spec tools/cockpit.spec.in
   create-archive:
     - ./autogen.sh mock
-    # The sandcastle doesn't have enough ram to run webpack, so wait
-    # until the webpack-jumpstart workflow has run and grab the result.
-    - tools/webpack-jumpstart --wait --rebase
     - make XZ_OPT=-0 dist
     - sh -c 'echo cockpit-*.tar.xz'
 srpm_build_deps:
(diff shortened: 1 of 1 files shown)


martinpitt:
What's the motivation here? I'd actually like to get us to a place where we build the webpacks *less* often, to improve reproducibility and reduce the amount of useless CPU cycle wasting.  I.e. trigger tests after the webpack-jumpstart run. We can't do that with the packit runs, but we can at least in principle do it with our own CI (and thus reduce capacity usage)

allisonkarlitskaya:
Basically: our motivation for using the jumpstart before is gone, and using it introduces an extra moving part.  Also: it seems like building the webpack on copr is actually *faster* than waiting for GitHub.  Finally: it might be nice to kill `--rebase` mode in jumpstart... I think this is the only thing that uses it.

I still don't see jumpstarting as the primary way that we ought to be building the project — more like an optimisation.

If we wanted to make this the "official" way, then we might imagine something exotic like turning `dist/` into a gitlink and having every commit that changes `pkg/` also include an updated build of `dist/`.  I'm pretty sure we don't want that, though.

allisonkarlitskaya:
One note about the COPR thing that weighs against this: from my (very limited) observations there is a short wait in the queue before we get picked up.  Also, there's the time taken to install the dependencies.  Having the webpack running on GitHub in parallel with that happening *is* a win...

Since you seem to be against this and I'm kinda ambivalent, let's just close this.

PR #17105, opened 2022-03-09
Add and use tools/make-dist

allisonkarlitskaya:
A new approach to get a tarball, with minimal dependencies, without impacting the configuration of the working directory.


martinpitt (approved):
Wow, nice cleanups! Lots of changes, but I can't really find anything wrong with it.

PR #17106, opened 2022-03-09
sos report test (slow)

jelly:
See:

```
test/verify/check-sosreport TestSOS.testBasic
9.23 min
test/verify/check-sosreport TestSOS.testWithUrlRoot
9.23 min
```
https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci?orgId=1&from=now-2d&to=now&viewPanel=34

### DIFF
--- a/test/verify/check-sosreport
+++ b/test/verify/check-sosreport
@@ -35,6 +35,9 @@ class TestSOS(MachineCase):
         b = self.browser
         m = self.machine
 
+        # HACK: sos ships wrong sos.conf, we need to mkdir it, see https://bugzilla.redhat.com/show_bug.cgi?id=1903281
+        if m.image == "centos-8-stream":
+            m.execute("mkdir -p /etc/sos")
         self.write_file("/etc/sos/sos.conf",
                         """
 [global]
@@ -52,9 +55,8 @@ only-plugins=release,date,host,cgroups,networking
         b.click("button:contains('Create report')")
         b.wait_visible("#sos-dialog")
 
-        with b.wait_timeout(540):
-            b.wait(lambda: b.eval_js('ph_find("#sos-progress .pf-c-progress__bar").offsetWidth') > 0)
-            b.wait_visible("#sos-download")
+        b.wait(lambda: b.eval_js('ph_find("#sos-progress .pf-c-progress__bar").offsetWidth') > 0)
+        b.wait_visible("#sos-download")
 
         if b.cdp.browser.name == "chromium":
             b.cdp.invoke("Page.setDownloadBehavior", behavior="allow", downloadPath=b.cdp.download_dir)
--- a/test/verify/check-sosreport
+++ b/test/verify/check-sosreport
@@ -35,12 +35,8 @@ class TestSOS(MachineCase):
         b = self.browser
         m = self.machine
 
-        # HACK: sos ships wrong sos.conf, we need to mkdir it, see https://bugzilla.redhat.com/show_bug.cgi?id=1903281
-        m.execute("mkdir -p /etc/sos")
-
-        self.restore_file("/etc/sos/sos.conf")
-        m.write("/etc/sos/sos.conf",
-                """
+        self.write_file("/etc/sos/sos.conf",
+                        """
 [global]
 threads=1
 


test/verify/check-sosreport
```diff
@@ -52,9 +52,8 @@ only-plugins=release,date,host,cgroups,networking
         b.click("button:contains('Create report')")
         b.wait_visible("#sos-dialog")
 
-        with b.wait_timeout(540):
```
martinpitt, line 59: Note that as long as everything is as expected, this will *not* wait for 9 minutes regularly. It will wait *up to* 9 minutes, just like a normal b.wait_* will wait *up to* 60s, but of course immediately return once the condition gets true.

This is a correct change -- we expect to write a config which is fast, and apparently we violate this somewhere. So let's find out where/how. Thanks!

jelly: > Note that as long as everything is as expected, this will _not_ wait for 9 minutes regularly. It will wait _up to_ 9 minutes, just like a normal b.wait_* will wait _up to_ 60s, but of course immediately return once the condition gets true.

Yes, that's what I mentioned in the commit on Fedora-36 sos report is broken so we wait 9 minutes (2x)

> This is a correct change -- we expect to write a config which is fast, and apparently we violate this somewhere. So let's find out where/how. Thanks!

I am not following what you mean here? Writing the config file was not fast or made it slower?

martinpitt: I meant that we *may* write the wrong config contents, something that the current version of sos can't understand any more. sosreport taking very long (minutes instead of seconds) would be a symptom of that. This already happened for sos 3 → 4.

martinpitt:
Seems to work everywhere. Now I understand what *could* go wrong -- if TestSOS was broken, then the test would take 9 minutes -- and then be SKIPed through a naughty pattern.

But [f36](https://logs.cockpit-project.org/logs/pull-17106-20220309-171407-d0004471-fedora-36/log.html) seems to work just fine?

martinpitt (approved):
Thanks!

test/verify/check-sosreport
```diff
@@ -35,12 +35,8 @@ class TestSOS(MachineCase):
         b = self.browser
         m = self.machine
 
-        # HACK: sos ships wrong sos.conf, we need to mkdir it, see https://bugzilla.redhat.com/show_bug.cgi?id=1903281
-        m.execute("mkdir -p /etc/sos")
```
martinpitt, line 39: This [still happens on c8s](https://artifacts.dev.testing-farm.io/43902888-1adf-4c1c-82e5-c855afa5b497/), so the mkdir needs to be put back. The `write_file()` can stay, of course.

martinpitt (changes requested):

jelly:
> Seems to work everywhere. Now I understand what _could_ go wrong -- if TestSOS was broken, then the test would take 9 minutes -- and then be SKIPed through a naughty pattern.
> 
> But [f36](https://logs.cockpit-project.org/logs/pull-17106-20220309-171407-d0004471-fedora-36/log.html) seems to work just fine?

Ah, it was broken under my pilot duty, so I guess the statistics are just that old as we collect statistics from the last 7 days.

```
sqlite> select run,seconds,(select context from testruns where id=run),(select DATETIME(time, 'unixepoch') from testruns where id=run) from tests where testname = 'test/verify/check-sosreport TestSOS.testBasic' and seconds > 500;
17823|552|fedora-36|2022-02-24 10:30:49
17823|552|fedora-36|2022-02-24 10:30:49
17823|551|fedora-36|2022-02-24 10:30:49
17824|549|fedora-36|2022-02-24 14:54:35
18074|549|fedora-36|2022-03-01 13:13:29
18123|550|fedora-36|2022-03-02 07:48:06
18138|553|fedora-36|2022-03-02 09:22:22
18161|549|fedora-36|2022-03-02 10:09:35
18183|550|fedora-36|2022-03-02 06:25:42
18184|549|fedora-36|2022-03-02 07:48:28
18186|551|fedora-36|2022-03-02 08:36:01
18187|549|fedora-36|2022-03-02 09:32:20
18203|551|fedora-36|2022-03-02 10:30:42
18233|550|fedora-36|2022-03-02 16:55:19
18248|550|fedora-36|2022-03-02 23:01:08
18272|550|fedora-36|2022-03-03 08:59:26
18287|549|fedora-36|2022-03-03 11:04:49
18347|554|fedora-36|2022-03-03 08:57:07
```

test/verify/check-sosreport
jelly: Do we need to report this somewhere? It worked on rhel-8-6 so I'm kinda suprised.

test/verify/check-sosreport
jelly: Ok, is there a proposal to do it different, we could use sed instead but that feels a lot more brittle as not all configuration options are available or commented out by default:

```
[root@rhel-8-6-127-0-0-2-2201 ~]# grep threads /etc/sos/sos.conf
[root@rhel-8-6-127-0-0-2-2201 ~]# grep only-plugins /etc/sos/sos.conf
[root@rhel-8-6-127-0-0-2-2201 ~]#
```

test/verify/check-sosreport
martinpitt: Not sure if it's to be considered a bug, but if you think so, bugzilla away :grin:

martinpitt (approved):
Dank je wel!

PR #17111, opened 2022-03-10
fmf: Trim tests to only exercise external APIs and be faster

martinpitt:
packit tests currently take ~ 1.5 hours, which is way too long for our
"max. 1 hour" goal. Many of them are also not that useful in downstream
gating, as they only exercise cockpit-internal functionality -- it's
enough to do that in upstream CI where we have a lot more performance
(due to parallelization).

So trim the list to those tests which exercise external APIs, and are
thus useful for reverse dependency gating.

This also eliminates all the excludes. Still keep the mechanism in
place, as we are surely going to have more of them in the future
(especially for development releases).

Adjust the test timeout to 1 hour to match our goal.

---

See the [log of a current test](https://artifacts.dev.testing-farm.io/bca39cdc-ee44-40bd-96a4-8d6b3afcea1f//work-allJ0Tz_d/log.txt) for current timings:
```
07:08:14 /var/ARTIFACTS/work-allJ0Tz_d
[...]
08:31:31 total: 1 test passed
```

### DIFF
--- a/test/run-verify-host-user.sh
+++ b/test/run-verify-host-user.sh
@@ -62,12 +62,12 @@ if [ -n "$test_optional" ]; then
          TestUpdates
          TestAutoUpdates
          TestStorage"
+
+    # Testing Farm machines often have pending restarts/reboot
+    EXCLUDES="$EXCLUDES TestUpdates.testBasic TestUpdates.testFailServiceRestart"
 fi
 
 if [ -n "$test_basic" ]; then
-    # Testing Farm machines often have pending restarts/reboot
-    EXCLUDES="$EXCLUDES TestUpdates.testBasic TestUpdates.testFailServiceRestart"
-
     # PCI devices list is not predictable
     EXCLUDES="$EXCLUDES TestSystemInfo.testHardwareInfo"
 
--- a/test/run-verify-host-user.sh
+++ b/test/run-verify-host-user.sh
@@ -53,35 +53,36 @@ fi
 
 export TEST_ALLOW_JOURNAL_MESSAGES
 
-# select tests
+# We only have one VM and tests should take at most one hour. So run those tests which exercise external API
+# (and thus are useful for reverse dependency testing and gating), and exclude those which test cockpit-internal
+# functionality to upstream CI. We also need to leave out some which make too strict assumptions about the testbed.
 TESTS=""
 EXCLUDES=""
 RC=0
 if [ -n "$test_optional" ]; then
     TESTS="$TESTS
-         TestUpdates
-         TestAutoUpdates
-         TestStorage"
+           TestAutoUpdates
+           TestUpdates
+           TestStorage
+           "
 
     # Testing Farm machines often have pending restarts/reboot
     EXCLUDES="$EXCLUDES TestUpdates.testBasic TestUpdates.testFailServiceRestart"
+
+    # These don't test more external APIs
+    EXCLUDES="$EXCLUDES
+              TestAutoUpdates.testBasic
+              TestAutoUpdates.testPrivilegeChange
+
+              TestUpdates.testUnprivileged
+              TestUpdates.testPackageKitCrash
+              TestUpdates.testNoPackageKit
+              TestUpdates.testInfoTruncation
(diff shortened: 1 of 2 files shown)


martinpitt:
That's better -- [f35](https://artifacts.dev.testing-farm.io/1b79f828-f836-449e-b41d-986f87d6a994/work-alldMCpTp/log.txt) took about 30 minutes now. Of course the previous "overhead" of building the srpm and then rpm on COPR goes on top of that.

test/run-verify-host-user.sh
```diff
@@ -53,63 +53,58 @@ fi
...
         TestTeam
-        TestTerminal
-        TestTuned
-        "
+        TestNetworkingVLAN
 
-    # HACK: check-sos fails 100% on Testing Farm Fedora 36 without any error message; works in local tmt VM
-    if [ "$TEST_OS" = "fedora-36" ]; then
-        TESTS="${TESTS/TestSOS/}"
-    fi
```
martinpitt, line 107: I was hoping that this would be fixed by the recent check-sos PR and the naughty, but it [still fails](https://artifacts.dev.testing-farm.io/a10eee44-0474-42d5-a3f3-834eda9c1276/) on F36. Back to draft.

martinpitt:
Meh, [new run](https://artifacts.dev.testing-farm.io/7d3c64c0-e15d-4820-a73c-bd609f1e4c82/) says "nothing to do" for SOS upgrade. I'll go back to the old hack, let's not get bogged up with this issue for now.

test/run-verify-host-user.sh
```diff
@@ -53,62 +53,62 @@ fi
...
-    # HACK: check-sos fails 100% on Testing Farm Fedora 36 without any error message; works in local tmt VM
+    # HACK: https://bugzilla.redhat.com/show_bug.cgi?id=2058142
     if [ "$TEST_OS" = "fedora-36" ]; then
-        TESTS="${TESTS/TestSOS/}"
-    fi
-
-    # HACK: repeatedly fails on RHEL Testing Farm without error message, then corrupts VM
-    if [ "$TEST_OS" = "rhel-9-0" ]; then
-        EXCLUDES="$EXCLUDES TestPages.testBasic"
+        TESTS="${TESTS/TestSOS.testBasic/}"
```
allisonkarlitskaya, line 111: I don't really understand this: if we keep the excludes mechanism, why not use it for this?  Or dose it break if the test is explicitly listed in TESTS?

martinpitt: No, we could use it. I am not entirely sure why we changed `$TESTS` for this back then (I kept the old code). But I'm happy to clean this up and use EXCLUDES, it's more consistent.

martinpitt: Ah, now I know: We originally did this as we would otherwise have to mention all four TestSOS tests individually (--exclude specifices a single test by full name, by design). But now that we only run one, exclude is more appropriate.

martinpitt:
@allisonkarlitskaya , @marusak : I reworked this to be mostly opt-out now (i.e. via long EXCLUDES list). That makes it much harder to see which tests actually run (which is why I originally favored the opt-in approach), but as I was outvoted I'm also fine with this. I do agree that it is more robust to maintain -- new tests will automatically run and increase the test time, but with the 1 hour limit we will notice when it grows too much, and it's time for another pruning round.

martinpitt:
Tests still pass. [Fedora 35](https://artifacts.dev.testing-farm.io/fd55d315-fd02-48bb-8f24-0f3e99c77dab/work-allQZL2Y5/log.txt) took 34 mins, and [c8s](https://artifacts.dev.testing-farm.io/48ba6f5f-362e-4677-9740-cb8d05614005/work-allA6opMs/log.txt) took 25 mins. That gives us enough breathing space for adding some more nondestructive tests, and I think the current set of tests is a good balance.

allisonkarlitskaya (commented):
I don't really have a strong opinion on this other than to say thanks for doing it :)

marusak (approved):
Thanks!

PR #17113, opened 2022-03-10
test: drop DEFAULT_TEST_OS from locale tests

jelly:
Only test locales on RHEL as our default image also does pixel tests
which take longer and the locale tests take up a significant time (~ 5 min.)


martinpitt (approved):
Thanks!

PR #17114, opened 2022-03-10
test: drop bugzilla workaround for realmd

jelly:


test/verify/check-system-realms
```diff
@@ -272,10 +272,6 @@ class CommonTests:
 
         # delete the local admin user, going to use the domain one instead
         m.execute("userdel --remove admin; systemctl try-restart sssd")
-
-        # Tell realmd to not enable domain-qualified logins
-        # (https://bugzilla.redhat.com/show_bug.cgi?id=1575538)
-        m.write("/etc/realmd.conf", "[cockpit.lan]\nfully-qualified-names = no\n", append=True)
         m.execute(f"echo {self.admin_password} | realm join -vU {self.admin_user} cockpit.lan", timeout=300)
```
martinpitt, line 275: It's still necessary for AD, so we can at least move that to the TestAD class. I remember reporting this (but can't find where), and getting a "wontfix".

martinpitt (changes requested):
see CI failures. Thanks for the cleanup!

test/verify/check-system-realms
jelly: Ok, then let's drop this PR?

martinpitt: Yes, perhaps -- I think it doesn't hurt for the tests to explicitly say whether they want qualified or unqualified names, and leave the default to IPA/AD.

PR #17119, opened 2022-03-11
doc: Clarify unit of cockpit.file's max_read_size option

martinpitt:
See #17118


martinpitt:
@elcarn23 do you think that this is better? I appreciate hints how to express the unit more clearly.

elcarn23:
Yes, I think that's great.  I appreciate the clarification on the parameter.  Thanks again.

marusak (approved):

PR #17120, opened 2022-03-11
cockpit.spec: fix RHEL 8 check

allisonkarlitskaya:
We currently install a symlink for compatibility purposes with the
intention of only installing it on RHEL <= 8, but the way the test is
written:

```
%if 0%{?rhel} <= 8
```

We install it on RHEL 8 plus every non-RHEL system, because if %{rhel}
is not defined, then it's equivalent to `%if 0 <= 8`, which is true.

Since we don't care about RHEL 7 on this branch anymore, change the
comparison to exactly 8.


martinpitt (approved):
Thanks for spotting!

PR #17122, opened 2022-03-11
test: Reduce multi-VM test memory

martinpitt:
Our test VMs get 1.1 GiB of RAM by default, which is enough to fit more
complex tests. However, this demands a lot of unnecessary RAM in tests
which provision multiple VMs.

Explicitly specify the RAM for these tests; strive for making the sum of
all VMs not exceed 1.1 GiB. For example, the networking tests only start
a second machine to provide a DHCP server -- that its completely happy
with just 256 MiB. This isn't always possible -- e.g. candlepin or
FreeIPA on the services image need a lot of RAM, and the host
switching/multi-machine tests run three cockpit machines, which can only
be made that small. But at least severely reduce their size, so that
they don't exceed 2 GiB in total.

This makes it easier to plan capacity for our infrastructure (for
calculating maximum allowable parallelism), and also makes it easier for
developers to run these tests on laptops which don't have tons of RAM.

----

This should allow us to go ahead with https://github.com/cockpit-project/cockpituous/pull/459 to pack our tests on machines in a more predictable/less inefficient manner. It may also help with PRs like #16930 -- as soon as the multi-host tests are "affected" and thus retried, they fail rather reliably.

### DIFF
--- a/test/verify/check-kdump
+++ b/test/verify/check-kdump
@@ -249,8 +249,8 @@ class TestKdump(KdumpHelpers):
 @skipDistroPackage()
 class TestKdumpNFS(KdumpHelpers):
     provision = {
-        "0": {"address": "10.111.113.1/24"},
-        "nfs": {"image": TEST_OS_DEFAULT, "address": "10.111.113.2/24"}
+        "0": {"address": "10.111.113.1/24", "memory_mb": 1024},
+        "nfs": {"image": TEST_OS_DEFAULT, "address": "10.111.113.2/24", "memory_mb": 512}
     }
 
     def testBasic(self):
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -1117,9 +1117,9 @@ class TestMultiCPU(MachineCase):
 class TestGrafanaClient(MachineCase):
 
     provision = {
-        "0": {"address": "10.111.112.1/20", "dns": "10.111.112.1"},
+        "0": {"address": "10.111.112.1/20", "dns": "10.111.112.1", "memory_mb": 512},
         # forward Grafana port, so that a developer can connect to it with local browser
-        "services": {"image": "services", "forward": {"3000": 3000}}
+        "services": {"image": "services", "forward": {"3000": 3000}, "memory_mb": 512}
     }
 
     def testBasic(self):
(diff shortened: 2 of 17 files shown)


martinpitt:
[This testSudo arch failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17122-20220311-095551-5bc76534-arch/log.html#76-2) is curious -- I didn't even touch this, and by itself it passes. I have a feeling this gets broken by an earlier test due to shuffling the test cases around. I'll investigate.

martinpitt:
[TestMultiMachine fails a lot](https://logs.cockpit-project.org/logs/pull-17122-20220311-095551-5bc76534-rhel-8-6/log.html#246-2) now on some OSes. 400 MB is enough for fedora, but not for RHEL/CentOS 8/9.

There's also an [IPA flake](https://logs.cockpit-project.org/logs/pull-17122-20220311-095551-5bc76534-rhel-9-0/log.html#273), due [low memory on the client machine](https://logs.cockpit-project.org/logs/pull-17122-20220311-095551-5bc76534-rhel-9-0/TestIPA-testClientCertAuthentication-rhel-9-0-127.0.0.2-2201-FAIL.log.gz), so I'll bump this a bit.

The rest of the tests look good enough.

martinpitt:
[c9s testSessionRecordingShell](https://artifacts.dev.testing-farm.io/d1af5de5-1473-4044-b6a1-2c13c73511a3/) failure. Retrying to see if that's a flake or it broke in C9S. But either way, unrelated to this PR.

martinpitt:
FTR: The failed webpack-jumpstart is because I accidentally closed and reopened this PR, which re-runs all the workflows.

allisonkarlitskaya (commented):
Could we maybe investigate ballooning?  I know this kinda sucks, but it seems like it may be a better approach than hardcoding...

According to https://libvirt.org/formatdomain.html#memory-balloon-device libvirt even has the device node enabled by default...

On the other hand, this could increase non-determinism and lead to more strange bugs... just wanted to throw it out there.

martinpitt:
[sessionRecordingShell failed again](https://artifacts.dev.testing-farm.io/c2bbcad3-d46a-45b3-b476-f1581912439c/),  I'll investigate this on Monday. Unrelated.

@allisonkarlitskaya : Actually, I *prefer* the hardcoding. It makes it much easier to calculate how much memory we need to allocate for each container and a given `$TEST_JOBS`. That said, I don't have a strong opinion about it, but overcommitting and unpredictability is why a lot of tests start to fail once we pack them too densely..

allisonkarlitskaya (approved):
I kinda agree with static is better...

PR #17123, opened 2022-03-11
selinux: Don't re-render after each item

marusak:
This page does too many re-renders while initializing the page. First it
adds item by item and re-render after each one. Then it loads details one
by one and re-renders after each item is updated. With this patch it
does render only once.

Now works nicely with https://github.com/cockpit-project/cockpit/pull/17094


martinpitt (approved):
Thanks! Yay for less flickering.

martinpitt:
[F35/packit](https://artifacts.dev.testing-farm.io/8051a164-f955-47c6-81e2-573d9d2722bd/) failed `TestServices.testBasic`, obviously unrelated.

PR #17124, opened 2022-03-11
cockpit-po-plugin: stop including base1/ in all packages

### LINKED ISSUE #13906: Include translations right into the "compiled" cockpit.js

allisonkarlitskaya:
Each package already includes base1/po.js, so including the base1/
translations in each webpack is redundant.

Fixes #13906


martinpitt (approved):
Ack. Right now this happens through the globbing, and #16930 makes that more explicit. Should be good either way.

PR #17125, opened 2022-03-11
updates: Use better key

marusak:
When new item would get added, the existing ones would get re-keyed.
Timestamp is unique enough.


martinpitt (approved):
Dakujem!

PR #17128, opened 2022-03-11
test: Robustify TestReauthorize.testSudo

martinpitt:
Create a new user `barney` instead of our over-used `user` name.
Apparently some tests leave behind a valid sudo token in the right
circumstances, which makes the test unexpectedly get sudo privileges at
the beginning.

----

[example 1](https://logs.cockpit-project.org/logs/pull-17122-20220311-095551-5bc76534-fedora-coreos/log.html#76), [example 2](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17122-20220311-095551-5bc76534-arch/log.html#76-2), [example 3](https://logs.cockpit-project.org/logs/pull-17122-20220311-131019-99f17013-rhel-8-6/log.html#76-2)

### DIFF
--- a/test/verify/check-reauthorize
+++ b/test/verify/check-reauthorize
@@ -90,10 +90,10 @@ class TestReauthorize(MachineCase):
         m = self.machine
         b = self.browser
 
-        m.execute("useradd user -s /bin/bash -c 'Barney' || true")
-        m.execute("echo user:foobar | chpasswd")
+        m.execute("useradd barney -s /bin/bash -c 'Barney' || true")
+        m.execute("echo barney:foobar | chpasswd")
 
-        b.default_user = "user"
+        b.default_user = "barney"
         self.login_and_go("/playground/test")
         b.click(".super-channel button")
         b.wait_in_text(".super-channel span", 'result: ')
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Okay, but I feel like we keep running into a philosophical difference here.  From my side, leftover sudo tokens are failures to properly clean up...

martinpitt:
@allisonkarlitskaya : They are -- but we already clean up users between tests, and it's a timing issue -- I know it's the easy way out, I just don't have the time to dig through it properly admittedly..

martinpitt:
I recently saw a [similar failure](https://artifacts.dev.testing-farm.io/c2bbcad3-d46a-45b3-b476-f1581912439c/) in C9S in `TestLogin.testSessionRecordingShell`. This seems to have been "fixed" as well with this PR. But this is still unclean.

I tested this behaviour like this. As root (on our test VM, I started with rhel-8-6):

    useradd user -s /bin/bash ; echo user:foobar | chpasswd ; usermod -aG wheel user

Then `ssh user@c` and run `sudo whoami` (asks for password). Log out, back in, and `sudo whoami` again asks for a password, even though logging out keeps `/run/sudo/ts/user` intact. The same is true for a VT login (which ought to keep the same TTY).

`pkill -u user` does the same. So I'm not sure how exactly the sudo perms stay around, but I'll send a PR to fix this more cleanly.

martinpitt:
Sent #17129 with a better fix.

PR #17129, opened 2022-03-12
test: Clean up sudo privileges for test user in TestLogin.testNoAdminGroup, Drop alternative truths on user creation

martinpitt:
Better fix for #17128, plus some cleanup as repentence.

This ought to fix [this failure](https://artifacts.dev.testing-farm.io/c2bbcad3-d46a-45b3-b476-f1581912439c/), and similar ones that we had before: [example 1](https://logs.cockpit-project.org/logs/pull-17122-20220311-095551-5bc76534-fedora-coreos/log.html#76), [example 2](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17122-20220311-095551-5bc76534-arch/log.html#76-2), [example 3](https://logs.cockpit-project.org/logs/pull-17122-20220311-131019-99f17013-rhel-8-6/log.html#76-2)

### DIFF
--- a/test/verify/check-pages
+++ b/test/verify/check-pages
@@ -299,7 +299,7 @@ OnCalendar=daily
         m = self.machine
         b = self.browser
 
-        m.execute('useradd scruffy -s /bin/bash -c \'Scruffy\' || true')
+        m.execute('useradd scruffy -s /bin/bash -c Scruffy')
         m.execute('echo scruffy:foobar | chpasswd')
 
         if "debian" in m.image:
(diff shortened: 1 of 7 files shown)


martinpitt:
Didn't help for C9s, need to try and reproduce.

martinpitt:
I split the cleanup to PR #17130, but we can also land it right here.

martinpitt:
Found the culprit, *headdesk*.

allisonkarlitskaya (approved):
I feel *much* better about this now, thanks.

PR #17130, opened 2022-03-12
test: Drop alternative truths on user creation

martinpitt:
We don't expect this to fail -- if it does, it points out a bug in our
testbed setup/cleanup, and we very much want to know.


martinpitt:
Closing in favor of #17129

PR #17132, opened 2022-03-13
IGNORE: Stress-test CI

martinpitt:
Touch all pages to cause an affected test avalanche, for testing
https://github.com/cockpit-project/cockpituous/pull/459


martinpitt:
New round, after today's test fixes, and with a busy CI (other PRs are running)

PR #17133, opened 2022-03-13
HACKING: Drop obsolete Edge documentation

martinpitt:
We dropped the windows-10 image a year ago (bots commmit ea9543f8818c1).

----

This is legit, but also part of stress-testing CI with https://github.com/cockpit-project/cockpituous/pull/459


marusak (approved):
Thanks! Just the other day I happened to scroll down to this section and though that I should drop it, but forgot to do it :)

marusak:
I will let you land this if you want to use it for more test runs.

PR #17135, opened 2022-03-14
test: Call super().tearDown() as appropriate, test fixes

mvollmer:
So that the usual tearDown actions will run, such as journal checking
and code coverage processing.

 - Requires https://github.com/cockpit-project/bots/pull/3097
 - Requires https://github.com/cockpit-project/bots/pull/3098

PR #17137, opened 2022-03-14
test: Restore complete /etc/ for each test

martinpitt:
Instead of hand-picking config files in /etc, let's just backup/restore
all files there. It's cheap enough, and avoids bugs like the one fixed
in commit ec24d029076b4.

Drop some redundant restoration of files/directories inside /etc.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1293,13 +1293,9 @@ class MachineCase(unittest.TestCase):
         # temporary directory in the VM
         self.addCleanup(m.execute, "if [ -d {0} ]; then findmnt --list --noheadings --output TARGET | grep ^{0} | xargs -r umount && rm -r {0}; fi".format(self.vm_tmpdir))
 
-        # users/groups/home dirs
-        self.restore_file("/etc/passwd")
-        self.restore_file("/etc/group")
-        self.restore_file("/etc/shadow")
-        self.restore_file("/etc/gshadow")
-        self.restore_file("/etc/subuid")
-        self.restore_file("/etc/subgid")
+        # configuration (including users) and home dirs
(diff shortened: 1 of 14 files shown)


martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-17137-20220314-231202-5b464996-fedora-35-firefox/log.html#47-2) needs fixing, done in a separate commit.

martinpitt:
[f35/packit](https://artifacts.dev.testing-farm.io/1ce432c7-0da2-4cf7-a9db-072d4b90e23b/work-allh1Vvl7/log.txt) still failed this in the bond test:
```
subprocess.CalledProcessError: Command '! test -d /etc/sysconfig/network-scripts || test -f /etc/sysconfig/network-scripts/ifcfg-tbond' returned non-zero exit status 1.
```

martinpitt:
I can [reproduce this failure](https://logs.cockpit-project.org/logs/pull-17137-20220315-122653-70db55eb-ubuntu-2004/log.html#136-2), it's because tests now restore /etc/fstab and udisks doesn't notice the unmount (inotify doesn't do that).

The [fedora-36 failures](https://artifacts.dev.testing-farm.io/ef1d36c0-389b-47ac-a0b7-9c603c237500/) are most likely because of missing tearDown() handlers, see #17135. Blocking on that, I'll fix that first.

martinpitt:
I don't like this udisks hack, and the bind-mounting on /etc might break other things. I sent #17139 to avoid the bind mount (at the expense of taking longer for test cleanup), but let's get two parallel runs and compare.

martinpitt:
I don't like this enough to land it. Breaking inotify watches of running services on /etc is a big enough red flag. So I call this a failed experiment.

PR #17139, opened 2022-03-15
test: Restore complete /etc/ for each test (version 2)

martinpitt:
Instead of hand-picking config files in /etc, let's just backup/restore
all files there. It's cheap enough, and avoids bugs like the one fixed
in commit ec24d029076b4.
    
Drop some redundant restoration of files/directories inside /etc.

-----

Alternate version of PR #17137. It should be more robust, e.g. it doesn't need the udisks reloading hack (which would introduce new ways to shoot ourselves into the foot). Let's get me a parallel test run.


martinpitt:
OK, this is very clearly the wrong approach.

PR #17143, opened 2022-03-16
storage: Highlight side panel new items

### LINKED ISSUE #10059: storage: hard to notice a new volume group (or generally another resource, see lvolumes) that was just created

marusak:
Peek 2022-03-18 08-02

Fixes #10059

### DIFF
--- a/pkg/storaged/side-panel.jsx
+++ b/pkg/storaged/side-panel.jsx
@@ -32,12 +32,27 @@ export class SidePanel extends React.Component {
     constructor() {
         super();
         this.state = { collapsed: true };
+        this.current_rows_keys = [];
+        this.new_keys = [];
     }
 
     render() {
         let show_all_button = null;
-        let rows = this.props.rows;
+        let rows = this.props.rows.filter(row => !!row);
 
+        // Find new items for animations
+        const current_keys = rows.map(row => row.key);
+
+        if (JSON.stringify(this.current_rows_keys) !== JSON.stringify(current_keys)) {
+            if (this.current_rows_keys.length !== 0) {
+                const new_keys = current_keys.filter(key => this.current_rows_keys.indexOf(key) === -1);
+                if (new_keys.length)
+                    this.new_keys.push(...new_keys);
+            }
+            this.current_rows_keys = current_keys;
+        }
+
+        // Collapse items by default if more than 20
         if (this.state.collapsed && rows.length > 20) {
             show_all_button = (
                 <FlexItem alignSelf={{ default: 'alignSelfCenter' }}>
@@ -49,6 +64,11 @@ export class SidePanel extends React.Component {
                 </FlexItem>);
             rows = rows.slice(0, 20);
         }
+
+        rows.forEach(row => {
+            if (row.key && this.new_keys.indexOf(row.key) !== -1)
+                row.className = (row.className || "") + " new-item-ct";
+        });
 
         const children = rows.map(row => row.block ? <SidePanelBlockRow {...row} /> : <SidePanelRow {...row} />);
 
@@ -117,7 +137,7 @@ class SidePanelRow extends React.Component {
 
         return (
             <FlexItem data-testkey={this.props.testkey}
-                      className="sidepanel-row"
+                      className={"sidepanel-row " + (this.props.className || "")}
                       role="link" tabIndex="0"
                       onKeyPress={this.props.go ? go : null}
                       onClick={this.props.go ? go : null}>
@@ -148,6 +168,7 @@ class SidePanelBlockRow extends React.Component {
                              detail={detail}
                              go={() => { cockpit.location.go(parts.location) }}
                              actions={actions}
+                             className={this.props.className}
         />;
     }
 }
--- a/pkg/storaged/side-panel.jsx
+++ b/pkg/storaged/side-panel.jsx
@@ -67,7 +67,7 @@ export class SidePanel extends React.Component {
 
         rows.forEach(row => {
             if (row.key && this.new_keys.indexOf(row.key) !== -1)
-                row.className = (row.className || "") + " new-item-ct";
+                row.className = (row.className || "") + " ct-new-item";
         });
 
         const children = rows.map(row => row.block ? <SidePanelBlockRow {...row} /> : <SidePanelRow {...row} />);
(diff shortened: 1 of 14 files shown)


garrett:
IIRC, the sidepanel rows are implemented with flex, whereas the table rows are implemented with tables?

The animation was intended for table rows. If it adjusts display in the animation, then it would have an affect like in the video/GIF above, as the display type would change.

(I can look into this tomorrow. It's a bit too late for me to do anything other than speculate in a comment like this. 😉)

marusak:
> IIRC, the sidepanel rows are implemented with flex, whereas the table rows are implemented with tables?

correct, both statements

Do you have a good idea how this could work also with flex?

garrett:
I'm working on this right now. I think I know what's up, but I'm building it in VMs for testing, as I don't want to accidentally nuke anything on my work laptop. :sweat_smile: (Therefore, it's a bit slower... but it is safer.)

I'll push up a fix as a commit in this PR if/when I have one.

marusak:
Thank you @garrett! Now it works superbly - I have updated the gif in the first comment.

marusak:
oops, I cause some oops :)

pkg/storaged/side-panel.jsx
```diff
@@ -32,12 +32,28 @@ export class SidePanel extends React.Component {
...
     }
 
     render() {
         let show_all_button = null;
-        let rows = this.props.rows;
+        let rows = this.props.rows.filter(!!row);
+
+        // Find new items for animations
+        const current_keys = [];
+        rows.forEach(row => { current_keys.push(row.key) });
```
martinpitt, line 45: In case you have to push again: This is much nicer as a map operation:

    const current_keys = rows.map(row => row.key);

pkg/storaged/side-panel.jsx
```diff
@@ -32,12 +32,28 @@ export class SidePanel extends React.Component {
...
+
+        if (JSON.stringify(this.current_rows_keys) !== JSON.stringify(current_keys)) {
+            if (this.current_rows_keys.length !== 0) {
+                const new_keys = current_keys.filter(key => this.current_rows_keys.indexOf(key) === -1);
+                if (new_keys.length)
+                    this.new_keys.push(...new_keys);
```
martinpitt, line 51: This looks strange -- pushing a copy of an array onto itself? That gets dangerously close to [Russel's pararox](https://en.wikipedia.org/wiki/Russell%27s_paradox) and hence making the universe explode (or implode.. I always forget which :grinning: )

Joke aside, you mean something like "concat" here (except that this is not an inline modification). Can't you just do `this.new_keys = new_keys` here?

pkg/storaged/side-panel.jsx
```diff
@@ -117,7 +138,7 @@ class SidePanelRow extends React.Component {
 
         return (
             <FlexItem data-testkey={this.props.testkey}
-                      className="sidepanel-row"
+                      className={"sidepanel-row" + (this.props.className ? " " + this.props.className : "")}
```
martinpitt, line 141: It doesn't hurt to move the space to the initial `"side-panel-row "` string, and simplify this.

martinpitt (changes requested):
Thanks!

pkg/storaged/side-panel.jsx
martinpitt: Illustration, in case it's not clear:

```node
> a = []
[]
> b = [1,2]
[ 1, 2 ]
> a.push(b)
1
> a
[ [ 1, 2 ] ]
> 
```

pkg/storaged/side-panel.jsx
marusak: Of yeah, I think I had there also some condition before and ha would lead o `.reduce()` but now this surely can be done with `map()`. I'll update

pkg/storaged/side-panel.jsx
marusak: > Joke aside, you mean something like "concat" here (except that this is not an inline modification). Can't you just do this.new_keys = new_keys here?

Same as for ListingTable: https://github.com/cockpit-project/cockpit/pull/17094#discussion_r826726870

I want to keep list of all items that were added after initial load. For that I keep global state of new items and only append into it. Otherwise I could just have local array of new keys for each render, but that might cause issues when more items are added one by one in quick succession

marusak: > Illustration, in case it's not clear:

Not sure what it demonstrates. I need the spread operator exactly to avoid creating array of arrays as that is hell then to find element in.

martinpitt: Oh, I wasn't aware of *that* usage of the spread operator. Your linked comment used `[...var]`, which is by and large "a copy of `var`", but this doesn't have the `[]` and thus is more like calling .push() with lots of arguments. That works fine (I just found it hard to read, but now I know it).

marusak:
I need to refresh two pixel tests it seems :man_shrugging:

pkg/storaged/side-panel.jsx
marusak: done

pkg/storaged/side-panel.jsx
marusak: done

martinpitt:
[unrelated failure](https://logs.cockpit-project.org/logs/pull-17143-20220321-062402-200e1966-arch/log.html#302-2), I sent PR #17156 to try and fix that.

martinpitt (approved):
I feel like the [stratis pixel test update](https://github.com/cockpit-project/pixel-test-reference/compare/6579d45fef970177293dc1c666530fb150476109..ac6d9ad037ca8939d6795834abf82b35c4039836) here is both unrelated and an ongoing flake. You yourself just updated its pixels a week ago in commit cb38ee82e59caff7da.

Nevertheless, LGTM if/once green. Thank you!

marusak:
> I feel like the [stratis pixel test update](https://github.com/cockpit-project/pixel-test-reference/compare/6579d45fef970177293dc1c666530fb150476109..ac6d9ad037ca8939d6795834abf82b35c4039836) here is both unrelated and an ongoing flake. You yourself just updated its pixels a week ago in commit https://github.com/cockpit-project/cockpit/commit/cb38ee82e59caff7da0cebd62feaaf7c874c2bcf.

Yeah, I realized as well. It seems a bit odd that it fails on these two PRs which are very similar and not in any other

PR #17144, opened 2022-03-16
selinux: Fix policy for pam_faillock

martinpitt:
pam_faillock tries to chown its state files to the target user. That's
legit, so allow this in our policy. Drop the test hack that we've been
dragging around for a while.

### DIFF
--- a/selinux/cockpit.te
+++ b/selinux/cockpit.te
@@ -133,7 +133,8 @@ optional_policy(`
 #
 
 # cockpit-session changes to the actual logged in user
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):
It feels deeply weird to have this here, but since we limit the scope to cockpit-session, I guess it's fine.

As for everyone else: do you have any idea why they closed the downstream bug?

martinpitt:
> do you have any idea why they closed the downstream bug?

Why wouldn't they? We maintain the policy in cockpit now, nowhere else.

PR #17145, opened 2022-03-17
shell: Re-include cross-package translations

martinpitt:
Revert the shell part of commit af5678b0be5. This broke translation of
menu entries for external pages like starter-kit.

----

See https://github.com/cockpit-project/starter-kit/pull/542#issuecomment-1070472816 for details. I am actually not sure how this worked for our internal pages -- do we somehow bundle all manifest translations into shell's po file?

I built this locally, installed it into a VM, and validated that this makes starter-kit menu translation and tests happy again. We obviously need a better solution for this, but for now let's fix the regression.

### DIFF
--- a/pkg/shell/index.html
+++ b/pkg/shell/index.html
@@ -7,9 +7,8 @@
     <link href="index.css" rel="stylesheet">
     <link href="../../static/branding.css" rel="stylesheet">
     <script src="../base1/cockpit.js"></script>
-    <script src="../base1/po.js"></script>
     <script src="../manifests.js"></script>
-    <script src="po.js"></script>
+    <script src="../*/po.js"></script>
   </head>
   <body class="pf-m-redhat-font" hidden>
     <div id="main" class="page">


martinpitt:
Worked around in https://github.com/cockpit-project/starter-kit/pull/543 for now, but only for version 265. Hence release-blocker.

martinpitt:
The only alternative that comes to my mind is to include translations into our manifests. Similar to desktop/appstream/polkit files. There had actually been a large push to *drop* translations from these latter files and use gettext() at runtime instead (much easier to update distro-wide translations that way), but this doesn't work for cockpit pages anyway.

allisonkarlitskaya (commented):
This should be handled by this:

https://github.com/cockpit-project/cockpit/blob/main/webpack.config.js#L335

It's apparently broken.

Edit: nevermind.  *external* pages.  Right.  That actually makes sense.

allisonkarlitskaya:
So, for example, the string "Software Updates" appears in exactly one spot:

```
> git grep "Software Updates"
pkg/packagekit/manifest.json:            "label": "Software Updates",
```

And, as expected:

```
> zgrep -l '"Software Updates"' dist/*/po.*.js.gz
dist/packagekit/po.cs.js.gz
dist/packagekit/po.de.js.gz
dist/packagekit/po.es.js.gz
dist/packagekit/po.fi.js.gz
dist/packagekit/po.fr.js.gz
dist/packagekit/po.he.js.gz
dist/packagekit/po.it.js.gz
dist/packagekit/po.ja.js.gz
dist/packagekit/po.ka.js.gz
dist/packagekit/po.ko.js.gz
dist/packagekit/po.nb_NO.js.gz
dist/packagekit/po.nl.js.gz
dist/packagekit/po.pl.js.gz
dist/packagekit/po.pt_BR.js.gz
dist/packagekit/po.sk.js.gz
dist/packagekit/po.sv.js.gz
dist/packagekit/po.tr.js.gz
dist/packagekit/po.uk.js.gz
dist/packagekit/po.zh_CN.js.gz
dist/shell/po.cs.js.gz
dist/shell/po.de.js.gz
dist/shell/po.es.js.gz
dist/shell/po.fi.js.gz
dist/shell/po.fr.js.gz
dist/shell/po.he.js.gz
dist/shell/po.it.js.gz
dist/shell/po.ja.js.gz
dist/shell/po.ka.js.gz
dist/shell/po.ko.js.gz
dist/shell/po.nb_NO.js.gz
dist/shell/po.nl.js.gz
dist/shell/po.pl.js.gz
dist/shell/po.pt_BR.js.gz
dist/shell/po.sk.js.gz
dist/shell/po.sv.js.gz
dist/shell/po.tr.js.gz
dist/shell/po.uk.js.gz
dist/shell/po.zh_CN.js.gz
```

Something else weird is going on here.  I'd prefer to understand it in more detail first.

allisonkarlitskaya:
This problem kinda sorta lends itself to be solved by injecting the translations into the already-dynamically-generated `/manifests.js` blob that we're sending to the shell.  The "network" side of this is pretty clear.  That would even let us remove these translations from the shell `.po` file (removing the exception I linked above).

How to organise this on disk is a lot harder to think about.  We could have the bridge open the .po file for the current language and send the relevant translations along with the manifest.  That would be a fair amount of work, at least in C.  It doesn't help that the `.po.js` files are not JSON, but rather have a small header of JS code.

We could also have webpack pack all of the translations into the manifest at build time, so we have like

```
   "label": "Software Updates"
   "label:de": "Software-Aktualisierungen"
```

or so, and ditto for the search terms.

The `*` is pretty overkill, though.

allisonkarlitskaya:
Note also: trying to do this dynamically per-language risks running into additional cache troubles, of course.

PR #17146, opened 2022-03-17
test: fips-mode-setup should be avaiable everywhere

jelly:
Test CI run, to check if it's really available everywhere.


martinpitt:
If this works, good to land with adjusting the commit message with "squash & merge".

martinpitt:
Looks good! Just some flakes due to retrying expensive tests.

martinpitt (approved):

PR #17149, opened 2022-03-18
workflows: Fix race between flatpak tests

martinpitt:
Even after commit 91854c8cc7b492, test-browser still sometimes tries to
talk to an allegedly existing instance, and then fails on talking into
the void. There are no related processes any more, apparently the
session D-Bus does not recognize the disappeared process fast enough.

Let's avoid that by running each test in its own D-Bus session.

Note: Keep sharing a single Xvfb server, as xvfb-run does not like being
run multiple times in a row without waiting.

-----

The flatpak test [fails](https://github.com/cockpit-project/cockpit/runs/5596465518?check_suite_focus=true) annoyingly often again:
```
+ flatpak ps
+ grep org.cockpit_project.CockpitClient
429308595	9871	org.cockpit_project.CockpitClient	org.gnome.Platform
12201561	9891	org.cockpit_project.CockpitClient	org.gnome.Platform
+ sleep 1
+ flatpak ps
+ grep org.cockpit_project.CockpitClient
+ rm cockpit-client-test-runner-cockpit-client-test-22

(cockpit-client:2): dbind-WARNING **: 06:10:29.601: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown
error: Environment variable must be given in the form VARIABLE=VALUE, not �
F
======================================================================
FAIL: testBasic (__main__.BrowserTest)
  File "containers/flatpak/test/test-browser", line 100, in testBasic
    self.assertIn(self.run_js("test-browser-wait-init.js"), ["found", "page-load"])
AssertionError: 'timed out waiting for JavaScript result' not found in ['found', 'page-load']
```

This is clearly test-browser talking to an old instance of the client from test-ssh still. I fixed that previously in commit 91854c8cc7b492fa96b , but something changed some two weeks ago.

### DIFF
--- a/.github/workflows/flatpak-test.yml
+++ b/.github/workflows/flatpak-test.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Smoke-test the installed flatpak
         run: |
           . /etc/profile.d/flatpak.sh
-          xvfb-run dbus-run-session sh -ec '
-            containers/flatpak/test/test-ssh
-            containers/flatpak/test/test-browser
+          xvfb-run sh -ec '
+            dbus-run-session containers/flatpak/test/test-ssh
+            dbus-run-session containers/flatpak/test/test-browser
           '
--- a/containers/flatpak/test/test-ssh
+++ b/containers/flatpak/test/test-ssh
@@ -12,6 +12,7 @@ FLATPAK_ID="${1:-org.cockpit_project.CockpitClient}"
 if mkdir ~/.ssh; then
     # ~/.ssh didn't exist yet.  Assume that we're safe to configure this.
     echo 'Include cockpit-client-test.config' > ~/.ssh/config
+    trap 'rm -rf ~/.ssh' EXIT INT QUIT PIPE
 else
     # There's already a ~/.ssh directory.  Don't modify that, but check it.
     if ! grep -q 'Include cockpit-client-test.config' ~/.ssh/config; then


martinpitt:
I fixed test-ssh to clean up properly, and now [this log](https://github.com/cockpit-project/cockpit/runs/5597220784?check_suite_focus=true) reproduces the failure. However, the only extra processes that are running after test-ssh are

```
runner      9274  0.1  0.1 262980 10296 ?        Sl   07:52   0:00 /usr/libexec/flatpak-session-helper
runner      9278  0.0  0.0   3984   208 ?        Ss   07:52   0:00 server --sh -n /home/runner/.cache/.flatpak-helper/pkcs11-flatpak-9274 --provider p11-kit-trust.so pkcs11:model=p11-kit-trust?write-protected=yes
runner      9305  0.2  0.1 306284 12868 ?        Sl   07:52   0:00 /usr/libexec/flatpak-portal
```
which don't seem to be related to our Client app. Perhaps the D-Bus session is asynchronous, and it still thinks that the died process is somehow running and wants to talk to it?

I'll try to run the tests in separate D-Bus contexts, to avoid this.

martinpitt:
There -- with [this version](https://github.com/cockpit-project/cockpit/commit/1064a8de6f319e76e1ffba3f4a764668db91f5f5) it survived 6 pairs of tests in a row. [passed test log](https://github.com/cockpit-project/cockpit/runs/5597433236?check_suite_focus=true)

So let's go with that, shall we?

containers/flatpak/test/test-ssh
```diff
@@ -12,6 +12,7 @@ FLATPAK_ID="${1:-org.cockpit_project.CockpitClient}"
 if mkdir ~/.ssh; then
     # ~/.ssh didn't exist yet.  Assume that we're safe to configure this.
     echo 'Include cockpit-client-test.config' > ~/.ssh/config
+    trap 'rm -rf ~/.ssh' EXIT INT QUIT PIPE
```
allisonkarlitskaya, line 15: Why doesn't this work with the second branch which checks that this has been configured?

.github/workflows/flatpak-test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Smoke-test the installed flatpak
         run: |
           . /etc/profile.d/flatpak.sh
-          xvfb-run dbus-run-session sh -ec '
-            containers/flatpak/test/test-ssh
-            containers/flatpak/test/test-browser
+          xvfb-run sh -ec '
+            dbus-run-session containers/flatpak/test/test-ssh
+            dbus-run-session containers/flatpak/test/test-browser
```
allisonkarlitskaya, line 37: I expected a bit more compartmentalisation here: we run several browser tests in `test-browser`, no?  And the compartmentalisation between them is working properly, apparently.

Maybe the problem is that we don't do the waiting before the first test in test-browser (or after the test in test-ssh)?  It seems a bit arbitrary to split it up like this...

On the other hand, if it works...

containers/flatpak/test/test-ssh
martinpitt: Before I added that, [tests failed](https://github.com/cockpit-project/cockpit/runs/5596989327?check_suite_focus=true) the second time. That might be coincidental (I didn't investigate too much). But mostly, I find it cleaner in general: If the test creates a file, it should remove it again. For running this locally, I need to `mv ~/.ssh ~/.ssh.bak`, and renaming it back doesn't work as the directory already exists.

.github/workflows/flatpak-test.yml
martinpitt: It does work, see the [stress test above](https://github.com/cockpit-project/cockpit/pull/17149#issuecomment-1072131094).

I wouldn't know what exactly to wait for at the start of test-browser (or the end of test-ssh, doesn't matter much). We already added the `flatpak ps` wait loop, and process wait loop. We could probably introspect D-Bus for a bit, but that's more time than I'd like to spend on this.

An alternative might be to simply swap the order of test-browser and test-ssh, and pretend that it's nicer to sort them alphabetically :grin: If you'd like me to test that, I can do that.

containers/flatpak/test/test-ssh
allisonkarlitskaya: You could also just add the line to your ssh config :)

.github/workflows/flatpak-test.yml
allisonkarlitskaya: I think I may have misunderstood what the browser tests are doing.  You're using one instance for all of the tests after all.

In that case, let's just go with your idea here.

allisonkarlitskaya (approved):

PR #17151, opened 2022-03-18
test: test Arch Linux offline

jelly:
Run cockpit tests offline on Arch Linux.

### DIFF
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -180,17 +180,13 @@ def main():
 
     dist_tar = subprocess.check_output([f'{BASE_DIR}/tools/make-dist'], text=True).strip()
 
-    customize = [os.path.join(BOTS_DIR, "image-customize")]
+    customize = [os.path.join(BOTS_DIR, "image-customize"), "--no-network"]
     if not args.overlay:
         customize.append("--fresh")
     if args.verbose:
         customize.append("--verbose")
     if args.quick:
         customize.append("--quick")
-
-    # rpms/debs can be built offline; arch can't yet
-    if args.image != 'arch':
-        customize.append("--no-network")
 
     if args.image == "fedora-coreos":
         customize += build_install_coreos(dist_tar, args.image, args.verbose, args.quick)


test/image-prepare
```diff
@@ -188,10 +188,6 @@ def main():
     if args.quick:
         customize.append("--quick")
 
-    # rpms/debs can be built offline; arch can't yet
-    if args.image != 'arch':
-        customize.append("--no-network")
```
martinpitt, line 193: Inverted logic -- you need to *always* specify `--no-network` now. This switches back all other OSes to online.

I was puzzled how the [arch test here](https://logs.cockpit-project.org/logs/pull-17151-20220318-154439-ecffe463-arch/log.html) was able to build.

jelly: That was silly, do we potentially want to allow an online build via a flag in image-customize?

martinpitt: The default already is "online build". `--no-network` disables that.

jelly:
So now tests started to fail due to `--no-network`?

```
# test (__main__.TestSuperuserDashboardOldMachine)
 x https://cockpit-images.eu-central-1.linodeobjects.com/
 x https://cockpit-images.us-east-1.linodeobjects.com/
 x https://cockpit-11.apps.cnfdb2.e2e.bos.redhat.com:8493/
image-download: image-download: couldn't find file anywhere: centos-7-0cc8c869540a7451538553248ab30402c94a784ee8c5cb811c43ac6184a73c74.qcow2
```

jelly:
> I'd suggest to first file the sudo fix, then land the image refresh, then land this PR. otherwise they both depend on each other and are impossible to get green.

Done in https://github.com/cockpit-project/cockpit/pull/17158

martinpitt:
Your PR still includes the sudo message commit (which is already on main), but hopefully the rebase button can deal with that? Please also trigger the required fedora-coreos test.

test/image-prepare
martinpitt: You dropped the conditional `--no-network`, but you need  to add it unconditionally somewhere. Preferably right to its initial value.

martinpitt (changes requested):
As you need to re-push anyway, please rebase and drop the sudo commit.

test/image-prepare
jelly: I've added it directly to `customize` now, is that ok?

martinpitt: Yes, looks fine. Thanks!

martinpitt (approved):
Cheers! There is some hiccup on GitHub -- the "build" workflow failed (without error message); I retried it, but it's stuck in "queued". So that and all packit builds will fail. But we don't need them for this PR.

PR #17152, opened 2022-03-19
tests: Stronger user/home directory cleanup between nondestructive tests

martinpitt:
This should fix failures like [this](https://logs.cockpit-project.org/logs/pull-16475-20220318-172950-02d7d82c-arch/log.html#265) and [this](https://logs.cockpit-project.org/logs/pull-16475-20220318-172950-02d7d82c-arch/log.html#265). I've also seen these countless times on c8s packit tests.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1344,11 +1344,13 @@ class MachineCase(unittest.TestCase):
         def terminate_sessions():
             # on OSTree we don't get "web console" sessions with the cockpit/ws container; just SSH; but also, some tests start
             # admin sessions without Cockpit
-            self.machine.execute("""for u in $(loginctl --no-legend list-users  | awk '{ if ($2 != "root") print $2 }'); do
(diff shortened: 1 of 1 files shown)


martinpitt:
There are  [two](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17152-20220319-055252-424695ad-fedora-35/log.html#245) [failures](https://logs.cockpit-project.org/logs/pull-17152-20220319-055252-424695ad-fedora-36/log.html#53):
```
umount: /run/user/1000: target is busy.
rm: cannot remove '/run/user/1000': Device or resource busy
```
Retries were fine, but it's not nice to replace one problem with another. Moving to draft for now.

martinpitt:
On second thought, it does incrementally improve things. It fixes some cases which would have previously failed anyway due to "permission denied" and such, and for the forever-busy runtime directories it at least points out the source of the problem more precisely.

marusak (approved):
Thank you!

PR #17156, opened 2022-03-21
test: Retry unmounting user runtime directory

martinpitt:
This should fix situations where the directory is still busy, so that it
cannot be removed.

---

This should hopefully fix [this kind of failure](https://logs.cockpit-project.org/logs/pull-17143-20220321-062402-200e1966-arch/log.html#302-2). I also added a cleanup to check-users to make its tests "affected", so that we get a more thorough run.


marusak (approved):
:crossed_fingers:

Issue #9467, opened 2018-06-21
services: No ability to edit services


andreasn:
This came up in the usability testing. There is apparently no way to edit a timer. Either if you did something wrong during the creation of it, or if you want to change the time it runs.

mvollmer:
The details page also doesn't show timer specific things like when it will trigger next.

kaushalyap:
Ability to delete timers also appreciated!

garrett:
I wonder if it's possible to add a custom field to the `[Timer]` section, such as `CockpitEditable=true`, which would allow us to find out which ones were custom made by Cockpit and are allowed to be edited in Cockpit. (I'm not sure if this would be fine with systemd, but I imagine it would be? We'd need to make sure.)

We probably don't want to edit system timers, right? Or if we "edit" system ones, we probably would want to create a custom timer based on the settings in the system-installed one and disable the original when enabling our custom one.

martinpitt:
> possible to add a custom field to the `[Timer]` section, such as `CockpitEditable=true`.

Yes, that's possible with the common `X-` prefix convention, i.e.  `X-Cockpit-Editable`.

garrett:
@jelly: Ah cool! Then we could use that instead. Even better!

jelly:
Another issue is that timers currently overwrite existing timers, so ideally it would not allow that too.

garrett:
> Another issue is that timers currently overwrite existing timers, so ideally it would not allow that too.

Yeah, we should definitely check for that. We could use a force overwrite pattern of sorts (perhaps like the account's force weak password in the account password modal) where the main button is disabled (to prevent accidental misclicks) and a secondary button shows up to force overwrite it. I suppose we'd want to have an error message of sorts show up, like an inline alert at the top of the modal showing that the timer's filename already exists.

jelly:
> @garrett, @jelly, @martinpitt Another feature that would be a great add (maybe there is already a way im not seeing) is the ability for the timer to repeat at a certain frequency. more specifically every x seconds, minutes, hours from creation. something like ![image](https://user-images.githubusercontent.com/56134647/179573373-66ac4cee-04ca-4b29-a27c-54f974a51e94.png)
> 
> thoughts?

I wonder how is this different from adding `Minutely` next to `Hourly. Currently hourly with 5 minutes generates:

```
[jelle@t14s][~]%systemd-analyze calendar '*-*-* *:5' --iterations=5
  Original form: *-*-* *:5
[...]

PR #17157, opened 2022-03-21
Check if command exists when creating a systemd timer

jelly:
image

### DIFF
--- a/pkg/systemd/services/timer-dialog.jsx
+++ b/pkg/systemd/services/timer-dialog.jsx
@@ -75,13 +75,14 @@ const CreateTimerDialogBody = ({ setIsOpen, owner }) => {
     const [repeatPatterns, setRepeatPatterns] = useState([]);
     const [specificTime, setSpecificTime] = useState("00:00");
     const [submitted, setSubmitted] = useState(false);
+    const [commandNotFound, setCommandNotFound] = useState(false);
     const validationFailed = {};
 
     if (!name.trim().length || !/^[a-zA-Z0-9:_.@-]+$/.test(name))
         validationFailed.name = true;
     if (!description.trim().length)
         validationFailed.description = true;
-    if (!command.trim().length)
+    if (!command.trim().length || commandNotFound)
         validationFailed.command = true;
     if (!/^[0-9]+$/.test(delayNumber))
         validationFailed.delayNumber = true;
@@ -109,9 +110,19 @@ const CreateTimerDialogBody = ({ setIsOpen, owner }) => {
             return false;
 
         setInProgress(true);
-        create_timer({ name, description, command, delay, delayUnit, delayNumber, repeat, specificTime, repeatPatterns, owner })
-                .then(() => setIsOpen(false), exc => {
-                    setDialogError(exc.message);
+
+        // Verify if the command exists
+        const command_parts = command.split(" ");
+        cockpit.spawn(["test", "-f", command_parts[0]], { err: "ignore" })
+                .then(() => {
+                    create_timer({ name, description, command, delay, delayUnit, delayNumber, repeat, specificTime, repeatPatterns, owner })
+                            .then(setIsOpen(false), exc => {
+                                setDialogError(exc.message);
+                                setInProgress(false);
+                            });
+                })
+                .fail(() => {
+                    setCommandNotFound(true);
                     setInProgress(false);
                 });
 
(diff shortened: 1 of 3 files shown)


garrett:
What improvements are there in this PR? All I see is checking if a command exists. Is there more?

Other things that would be nice:

- Editing timers (probably remove and re-add with adjustments)
- Sorting the manually added (probably Cockpit-added timers, really) at the top
- Timer deletion (for Cockpit-added timers)

...but all probably as follow-ups. 

Related: 

- https://github.com/cockpit-project/cockpit/issues/9467

I'll add the list over on that issue.

jelly:
> What improvements are there in this PR? All I see is checking if a command exists. Is there more?
> 
> Other things that would be nice:
> 
> * Editing timers (probably remove and re-add with adjustments)
> * Sorting the manually added (probably Cockpit-added timers, really) at the top
> * Timer deletion (for Cockpit-added timers)
> 
> ...but all probably as follow-ups.
> 
> Related:
> 
> * [services: No ability to edit timers #9467](https://github.com/cockpit-project/cockpit/issues/9467)
> 
> I'll add the list over on that issue.

For now it's just checking if the command is available, I have a local branch for deletion which I plan to create a new PR for.

garrett:
> For now it's just checking if the command is available, I have a local branch for deletion which I plan to create a new PR for.

Minor nitpick. Can we rename the PR's title to help explain what the PR is for? :grin:

jelly:
> > For now it's just checking if the command is available, I have a local branch for deletion which I plan to create a new PR for.
> 
> Minor nitpick. Can we rename the PR's title to help explain what the PR is for? grin

Sure!

garrett (changes requested):
Looks great!

However: Shouldn't the save button go disabled until the entry is changed?

jelly:
> Looks great!
> 
> However: Shouldn't the save button go disabled until the entry is changed?

Good point, shouldn't that apply for all? Read https://www.patternfly.org/v4/components/form/design-guidelines#buttons so I'll see what we can do here.

garrett:
> Good point, shouldn't that apply for all?

We already do this elsewhere in Cockpit. (For example: in Cockpit-Machines create VM, which is where I checked when suggesting this.) Some places _probably_ don't (as software written over time by multiple people can have differences), but should.

I'm suggesting resetting the disabled button state when the form is changed, which is literally "form submit buttons should be disabled until a user starts filling out the form or edits the form in any way", but after a submission.

jelly:
> 



> Looks great!
> 
> However: Shouldn't the save button go disabled until the entry is changed?

Done, please re-check

garrett:
> However: Shouldn't the save button go disabled until the entry is changed?

It doesn't seem to have worked?

image

jelly:
> > > However: Shouldn't the save button go disabled until the entry is changed?
> > 
> > 
> > Done, please re-check
> 
> It doesn't seem to have worked?
> 
> image

Hmmm

image

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -109,9 +110,19 @@ const CreateTimerDialogBody = ({ setIsOpen, owner }) => {
             return false;
 
         setInProgress(true);
-        create_timer({ name, description, command, delay, delayUnit, delayNumber, repeat, specificTime, repeatPatterns, owner })
-                .then(() => setIsOpen(false), exc => {
-                    setDialogError(exc.message);
+
+        // Verify if the command exists
+        const command_parts = command.split(" ");
+        cockpit.spawn(["test", "-f", command_parts[0]], { err: "ignore" })
```
marusak, line 116: Hmm, this assumes absolute path, right?

jelly: Yes, but I don't think ExecStart supports an relative paths? Reading the man page

https://www.man7.org/linux/man-pages/man5/systemd.service.5.html at ExecStart=
```
         For each of the specified commands, the first argument must
           be either an absolute path to an executable or a simple file
           name without any slashes.
           ```
But as far as I know it does not work with an relative path?

```
[jelle@t14s][/usr/bin]%sudo systemd-run -t ExecStart="uname"
Failed to find executable ExecStart=uname: No such file or directory
```

marusak (approved):
Thank you!

pkg/systemd/services/timer-dialog.jsx
marusak: Ah right, good to know!

PR #17159, opened 2022-03-21
Cleanup cockpit.format_*() API, make it more flexible, and use it to drop some hacks

martinpitt:
In our test VMs, the Overview page looks the same on main and on this PR:

vm-main

On my laptop the output is slightly different, as we moved from "one decimal digit" to "2 digits of precision". main:
laptop-main

this PR:
laptop-pr


martinpitt:
Ugh, [distropkg](https://logs.cockpit-project.org/logs/pull-17159-20220321-125026-769eae34-rhel-8-6-distropkg/log.html) -- looks like we may have to back out the storaged cleanup until after this hits a RHEL 8 release.

martinpitt:
Meh, seems with the dashed version it now works in packit, but not on [our CI in all cases](https://logs.cockpit-project.org/logs/pull-17159-20220321-134656-6079be23-fedora-35-container-bastion/log.html)

marusak (approved):
Heh, I was wondering where did 0.4 GiB went :) But good from my side, thanks!

Issue #11364, opened 2019-03-11
lm_sensors


jacksnodgrass:
Page: System 

would be nice to see CPU temp and Fan Speeds graphs from lm_sensor data on the system page.

JamesGKent:
temperatures are also mentioned in issue [5454](https://github.com/cockpit-project/cockpit/issues/11364)

and I think the method mentioned there has more merit, currently cockpit does not depend on lm-sensors, and this information is all exposed through the /sys/ interface. if anything I think the preferred option would be PCP if present, thermald if present, then the sys interface? that way it should work on the most basic of installs as well as others without dragging in more dependencies?

JamesGKent:
I have a rough prototype using lm-sensors [here](https://github.com/JamesGKent/cockpit-sensors-plugin)
however when i can work out how best to access the data i would like to rework this to make use of the graphs other stats use.
Unfortunately the machine I have available does not give me access to fan speeds through lm-sensors so I cannot write code to handle that.

waikontse:
@JamesGKent I think i can help with the code for reading the fan speed through lm_sensors.
@martinpitt The issue seems a bit sleeping at the moment. What are the plans for the current issue? I think I have some time to spare at the moment. Is the plan to still support cpu temperature with the details described in #5454 ?

martinpitt:
There are no plans from our team's side to do this. If this information is accessible through /sys, pcp, or some D-Bus interface, then if someone wants to extend our hardware information page with that, this would be great. Otherwise, if lm_sensors just has a shared library API, then it's not going to happen in cockpit itself, and this should be a standalone project then.

maltegrosse:
Another more generic approach could be to use the openbmc dbus "scheme" for reading data from a bmc. And then extend existing bmc data on dbus with data from hwmon. ( https://github.com/openbmc/phosphor-hwmon ). Afterwards cockpit can read (and even send) data from (to) the dbus and provide a fitting UI. (some UI inspiration: https://github.com/openbmc/phosphor-webui )

marusak:
We introduced new separate page for seeing current and historical performance stats. That would be a good place for this.

See design on https://garrett.github.io/cockpit-mockweb/metrics/current/

https://www.mjmwired.net/kernel/Documentation/thermal/sysfs-api.txt

e.g.
```
cat /sys/devices/virtual/thermal/thermal_zone0/temp 
61000
```
means 61 degrees Celsius.

<hr>

[...]

PR #17160, opened 2022-03-21
Metrics: Add CPU temperature to metrics page

tomasmatus:
Fixes: #11364

cpu_temperature_normal

cpu_temperature_warning

cpu_temperature_critical

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -23,7 +23,7 @@ import {
     Alert,
     Breadcrumb, BreadcrumbItem,
     Button,
-    Card, CardTitle, CardBody, Gallery,
+    Card, CardTitle, CardBody, CardHeader, Gallery,
     DescriptionList, DescriptionListGroup, DescriptionListTerm, DescriptionListDescription,
     Flex, FlexItem,
     Grid, GridItem,
@@ -37,7 +37,7 @@ import {
     Tooltip,
 } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody, TableGridBreakpoint, TableVariant, TableText, RowWrapper, cellWidth } from '@patternfly/react-table';
-import { ExclamationCircleIcon, CogIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { ExclamationTriangleIcon, ExclamationCircleIcon, CogIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 import cockpit from 'cockpit';
 import * as machine_info from "../lib/machine-info.js";
@@ -161,6 +161,10 @@ const CURRENT_METRICS = [
     { name: "cpu.core.nice", derive: "rate" },
 ];
 
+const CPU_TEMPERATURE_METRICS = [
+    { name: "cpu.temperature" },
+];
+
 const HISTORY_METRICS = [
     // CPU utilization
     { name: "kernel.all.cpu.nice", derive: "rate" },
@@ -213,10 +217,17 @@ class CurrentMetrics extends React.Component {
         super(props);
 
         this.metrics_channel = null;
+        this.temperature_channel = null;
         this.samples = [];
+        this.temperatureSamples = [];
         this.netInterfacesNames = [];
         this.cgroupCPUNames = [];
         this.cgroupMemoryNames = [];
+        this.cpuTemperatureColors = {
+            textColor: "",
+            iconColor: "",
+            icon: null,
+        };
 
         this.state = {
             userid: null,
@@ -224,6 +235,7 @@ class CurrentMetrics extends React.Component {
             swapUsed: null, // bytes
             cpuUsed: 0, // percentage
             cpuCoresUsed: [], // [ percentage ]
+            cpuTemperature: NaN, // degree Celsius
             loadAvg: null, // [ 1min, 5min, 15min ]
             disksRead: 0, // B/s
             disksWritten: 0, // B/s
@@ -237,6 +249,7 @@ class CurrentMetrics extends React.Component {
 
         this.onVisibilityChange = this.onVisibilityChange.bind(this);
         this.onMetricsUpdate = this.onMetricsUpdate.bind(this);
+        this.onTemperatureUpdate = this.onTemperatureUpdate.bind(this);
         this.updateMounts = this.updateMounts.bind(this);
         this.updateLoad = this.updateLoad.bind(this);
 
@@ -256,11 +269,23 @@ class CurrentMetrics extends React.Component {
     }
 
     onVisibilityChange() {
+        if (cockpit.hidden && this.temperature_channel !== null) {
+            this.temperature_channel.removeEventListener("message", this.onTemperatureUpdate);
+            this.temperature_channel.close();
+            this.temperature_channel = null;
+        }
+
         if (cockpit.hidden && this.metrics_channel !== null) {
             this.metrics_channel.removeEventListener("message", this.onMetricsUpdate);
             this.metrics_channel.close();
             this.metrics_channel = null;
             return;
+        }
+
+        if (!cockpit.hidden && (this.temperature_channel === null)) {
+            this.temperature_channel = cockpit.channel({ payload: "metrics1", source: "internal", interval: 3000, metrics: CPU_TEMPERATURE_METRICS });
+            this.temperature_channel.addEventListener("close", (ev, error) => console.error("CPU temperature metric closed:", error));
+            this.temperature_channel.addEventListener("message", this.onTemperatureUpdate);
         }
 
         if (!cockpit.hidden && this.metrics_channel === null) {
@@ -343,6 +368,33 @@ class CurrentMetrics extends React.Component {
                     console.warn("Failed to read /proc/loadavg:", ex.toString());
                     this.setState({ loadAvg: null });
                 });
+    }
+
+    onTemperatureUpdate(event, message) {
+        debug("current CPU temperature  message", message);
+        const data = JSON.parse(message);
+
+        if (!Array.isArray(data)) {
+            return;
+        }
+
+        data.forEach(temperatureSamples => decompress_samples(temperatureSamples, this.temperatureSamples));
+
+        this.cpuTemperature = parseInt(Math.max(...this.temperatureSamples[0]));
+
+        if (this.cpuTemperature <= 80) {
+            this.cpuTemperatureColors.textColor = "";
+            this.cpuTemperatureColors.iconColor = "";
+            this.cpuTemperatureColors.icon = null;
+        } else if (this.cpuTemperature < 95) {
+            this.cpuTemperatureColors.textColor = "text-color-warning";
+            this.cpuTemperatureColors.iconColor = "icon-color-warning";
+            this.cpuTemperatureColors.icon = <ExclamationTriangleIcon />;
+        } else {
+            this.cpuTemperatureColors.textColor = "text-color-critical";
+            this.cpuTemperatureColors.iconColor = "icon-color-critical";
+            this.cpuTemperatureColors.icon = <ExclamationCircleIcon />;
+        }
     }
 
     onMetricsUpdate(event, message) {
@@ -592,7 +644,19 @@ class CurrentMetrics extends React.Component {
         return (
             <Gallery className="current-metrics" hasGutter>
                 <Card id="current-metrics-card-cpu">
-                    <CardTitle>{ _("CPU") }</CardTitle>
+                    <CardHeader className='align-baseline'>
+                        <CardTitle>{ _("CPU") }</CardTitle>
+                        { !isNaN(this.cpuTemperature) &&
+                        <span className="temperature">
+                            <span className={this.cpuTemperatureColors.iconColor}>
+                                {this.cpuTemperatureColors.icon}
+                            </span>
+                            &nbsp;
+                            <span className={this.cpuTemperatureColors.textColor}>
+                                { cockpit.format("$0 °C", this.cpuTemperature) }
+                            </span>
+                        </span> }
+                    </CardHeader>
                     <CardBody>
                         <div className="progress-stack-no-space">
                             <Progress
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -66,6 +66,26 @@
           }
       }
 
+      .align-baseline {
+          align-items: baseline;
+      }
+
+      .text-color-warning {
+          color: var(--pf-global--warning-color--200);
+      }
+
+      .icon-color-warning {
+          color: var(--pf-global--warning-color--100);
+      }
+
+      .text-color-critical {
+          color: var(--pf-global--danger-color--200);
+      }
+
+      .icon-color-critical {
+          color: var(--pf-global--danger-color--100);
+      }
+
       .progress-stack {
           display: grid;
           padding-bottom: var(--pf-global--spacer--sm);
--- a/src/bridge/cockpitcpusamples.c
+++ b/src/bridge/cockpitcpusamples.c
@@ -112,3 +112,88 @@ out:
   g_strfreev (lines);
   g_free (contents);
 }
+
+static gchar*
+read_file (const gchar *path)
+{
+  g_autoptr(GError) error = NULL;
+  gchar *content = NULL;
+
+  if (!g_file_get_contents (path, &content, NULL, &error))
+    {
+      // ENOENT is used to break loops, do not log it
+      if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+          g_warning ("error reading file: %s", error->message);
+
+      return NULL;
+    }
+
+  return content;
+}
+
+static void
+sample_cpu_sensors (CockpitSamples *samples,
+                    int hwmonID)
+{
+  for (int i = 1; TRUE; i++)
+    {
+      g_autofree gchar *temp = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_input", hwmonID, i);
+      g_autofree gchar *temp_content = read_file (temp);
+
+      // end loop on error
+      if (temp_content == NULL)
+          break;
+
+      gint64 temperature = g_ascii_strtoll (temp_content, NULL, 10);
+      // overflow or invalid
+      if (temperature == G_MAXINT64 || temperature == G_MININT64 || temperature == 0)
+        {
+          g_debug("Invalid number in %s: %m", temp);
+          continue;
+        }
+
+      g_autofree gchar *label = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_label", hwmonID, i);
+      g_autofree gchar *label_content = read_file (label);
+
+      if (label_content == NULL)
+        {
+          // labels aren't used on ARM
+          label_content = g_strdup_printf ("Core %d", i);
+        }
+
+      g_strchomp (label_content);
+
+      // ignore Tctl on AMD devices
+      if (g_str_equal (label_content, "Tctl"))
+          continue;
+
+      g_autofree gchar *instance = g_strdup_printf ("hwmon%d %s", hwmonID, label_content);
+      cockpit_samples_sample (samples, "cpu.temperature", instance, temperature/1000);
+    }
+}
+
+void
+cockpit_cpu_temperature (CockpitSamples *samples)
+{
+  // iterate through all hwmon folders to find CPU sensors
+  for (int i = 0; TRUE; i++)
+    {
+      g_autofree gchar *path = g_strdup_printf ("/sys/class/hwmon/hwmon%d/name", i);
+      g_autofree gchar *name = read_file (path);
+
+      // end loop on error
+      if (name == NULL)
+          break;
+
+      g_strchomp (name);
+
+      // compare device name with CPU names
+      // Intel: coretemp, AMD: k8temp or k10temp, ARM: cpu_thermal
+      if (g_str_equal (name, "coretemp") || g_str_equal (name, "cpu_thermal") ||
+          g_str_equal (name, "k8temp") || g_str_equal (name, "k10temp"))
+        {
+          // hwmon contains CPU info
+          sample_cpu_sensors (samples, i);
+        }
+    }
+}
--- a/src/bridge/cockpitcpusamples.h
+++ b/src/bridge/cockpitcpusamples.h
@@ -26,6 +26,7 @@ G_BEGIN_DECLS
 
 void            cockpit_cpu_samples         (CockpitSamples *samples);
 
+void            cockpit_cpu_temperature     (CockpitSamples *samples);
 
 G_END_DECLS
 
--- a/src/bridge/cockpitinternalmetrics.c
+++ b/src/bridge/cockpitinternalmetrics.c
@@ -53,7 +53,8 @@ typedef enum {
   NETWORK_SAMPLER = 1 << 3,
   MOUNT_SAMPLER = 1 << 4,
   CGROUP_SAMPLER = 1 << 5,
-  DISK_SAMPLER = 1 << 6
+  DISK_SAMPLER = 1 << 6,
+  THERMAL_SAMPLER = 1 << 7,
 } SamplerSet;
 
 typedef struct {
@@ -99,6 +100,8 @@ static MetricDescription metric_descriptions[] = {
   { "cgroup.memory.sw-limit", "bytes",    "instant", TRUE, CGROUP_SAMPLER },
   { "cgroup.cpu.usage",       "millisec", "counter", TRUE, CGROUP_SAMPLER },
   { "cgroup.cpu.shares",      "count",    "instant", TRUE, CGROUP_SAMPLER },
+
+  { "cpu.temperature",        "celsius",  "instant", TRUE, THERMAL_SAMPLER },
 
   { NULL }
 };
@@ -334,6 +337,8 @@ cockpit_internal_metrics_tick (CockpitMetrics *metrics,
     cockpit_cgroup_samples (COCKPIT_SAMPLES (self));
   if (self->samplers & DISK_SAMPLER)
     cockpit_disk_samples (COCKPIT_SAMPLES (self));
+  if (self->samplers & THERMAL_SAMPLER)
+    cockpit_cpu_temperature (COCKPIT_SAMPLES (self));
 
   /* Check for disappeared instances
    */
--- a/src/bridge/test-metrics.c
+++ b/src/bridge/test-metrics.c
@@ -850,6 +850,58 @@ test_cpu_cores (void)
   g_object_unref (transport);
 }
 
+static void
+test_cpu_temperature (void)
+{
+  MockTransport *transport = mock_transport_new ();
+
+  g_signal_connect (transport, "closed", G_CALLBACK (on_transport_closed), NULL);
+
+  g_autoptr(JsonObject) options = json_obj ("{ 'metrics': [ { 'name': 'cpu.temperature' } ],"
+                                  "  'interval': 1000"
+                                  "}");
+
+  CockpitChannel *channel = g_object_new (cockpit_internal_metrics_get_type (),
+                          "transport", transport,
+                          "id", "1234",
+                          "options", options,
+                          NULL);
+
+  cockpit_channel_prepare (channel);
+
+  g_autoptr(GBytes) msg = NULL;
+  /* receive meta information */
+  while ((msg = mock_transport_pop_channel (transport, "1234")) == NULL)
+    g_main_context_iteration (NULL, TRUE);
+
+  g_autoptr(GError) error = NULL;
+  g_autoptr(JsonObject) res = cockpit_json_parse_bytes (msg, &error);
+  g_assert_no_error (error);
+  g_assert (res != NULL);
+  g_assert (json_object_has_member (res, "metrics"));
+
+  JsonArray *metrics = json_object_get_array_member (res, "metrics");
+  g_assert_cmpint (json_array_get_length (metrics), ==, 1);
+
+  JsonObject *description = json_array_get_object_element (metrics, 0);
+  g_assert (description);
+  g_assert_cmpstr (json_object_get_string_member (description, "name"), ==, "cpu.temperature");
+  g_assert_cmpstr (json_object_get_string_member (description, "units"), ==, "celsius");
+
+  g_autoptr(JsonArray) samples = recv_array (transport);
+  g_assert_cmpint (json_array_get_length (samples), ==, 1);
+  JsonArray *core = json_array_get_array_element (samples, 0);
+  g_assert_cmpint (json_array_get_length (core), ==, 1);
+  JsonArray *temperature = json_array_get_array_element (core, 0);
+
+  // file does not exist in virtual machines, skip value check
+  if (json_array_get_length (temperature) >= 1)
+    {
+      g_assert_cmpint (json_array_get_int_element (temperature, 0), >, 0);
+      g_assert_cmpint (json_array_get_int_element (temperature, 0), <, 150);
+    }
+}
+
 int
 main (int argc,
       char *argv[])
@@ -879,6 +931,7 @@ main (int argc,
   g_test_add_func ("/metrics/cgroup-memory", test_cgroup);
 
   g_test_add_func ("/metrics/cpu-cores", test_cpu_cores);
+  g_test_add_func ("/metrics/cpu-temperature", test_cpu_temperature);
 
   return g_test_run ();
 }
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -835,6 +835,70 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
         with b.wait_timeout(300):
             b.wait(lambda: float(b.text("#load-avg .pf-l-flex div:first-child").split()[-1].rstrip(',')) < 10)
 
+        # Files with CPU temperature do not exist, nothing is displayed
+        b.wait_not_present("#current-metrics-card-cpu .temperature")
+
+        # No matching type
+        m.execute("mkdir -p /tmp/hwmon/hwmon0")
+        m.execute("echo BAT0 > /tmp/hwmon/hwmon0/name")
+        m.execute("mount -o bind /tmp /sys/class")
+
+        b.wait_not_present("#current-metrics-card-cpu .temperature")
+
+        # create files that contain CPU temperature
+        m.execute("mkdir -p /tmp/hwmon/hwmon1")
+
+        # Testing on ARM devices
+        m.execute("echo cpu_thermal > /tmp/hwmon/hwmon1/name")
+        m.execute("echo 60000 > /tmp/hwmon/hwmon1/temp1_input")
+
+        b.wait_in_text("#current-metrics-card-cpu", "60 °C")
+
+        # Testing on AMD devices
+        m.execute("echo k10temp > /tmp/hwmon/hwmon1/name")
+        m.execute("echo Tctl > /tmp/hwmon/hwmon1/temp1_label && echo Tccd1 > /tmp/hwmon/hwmon1/temp2_label && echo Tccd3 > /tmp/hwmon/hwmon1/temp3_label")
+        # setting temp1_input to the highest value as Tctl should be ignored
+        m.execute("echo 150000 > /tmp/hwmon/hwmon1/temp1_input && echo 45000 > /tmp/hwmon/hwmon1/temp2_input && echo 43000 > /tmp/hwmon/hwmon1/temp3_input")
+
+        b.wait_in_text("#current-metrics-card-cpu", "45 °C")
+
+        m.execute("echo 90000 > /tmp/hwmon/hwmon1/temp3_input")
+        b.wait_visible("#current-metrics-card-cpu .text-color-warning")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-warning", "90 °C")
+
+        # Testing on Intel devices
+        m.execute("echo coretemp > /tmp/hwmon/hwmon1/name")
+        m.execute("echo Package id 0 > /tmp/hwmon/hwmon1/temp1_label && echo Core 0 > /tmp/hwmon/hwmon1/temp2_label && echo Core 1 > /tmp/hwmon/hwmon1/temp3_label")
+        m.execute("echo 45000 > /tmp/hwmon/hwmon1/temp1_input && echo 50000 > /tmp/hwmon/hwmon1/temp2_input && echo 43000 > /tmp/hwmon/hwmon1/temp3_input")
+
+        b.wait_in_text("#current-metrics-card-cpu", "50 °C")
+
+        # change temperature, check color and icon
+        m.execute("echo 85000 > /tmp/hwmon/hwmon1/temp2_input")
+        b.wait_visible("#current-metrics-card-cpu .text-color-warning")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-warning", "85 °C")
+
+        # change temperature, check color and icon
+        m.execute("echo 100000 > /tmp/hwmon/hwmon1/temp3_input")
+        b.wait_visible("#current-metrics-card-cpu .text-color-critical")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-critical", "100 °C")
+
+        # test multiple CPUs
+        m.execute("mkdir -p /tmp/hwmon/hwmon2")
+        m.execute("echo coretemp > /tmp/hwmon/hwmon2/name")
+        m.execute("echo Package id 0 > /tmp/hwmon/hwmon2/temp1_label && echo Core 0 > /tmp/hwmon/hwmon2/temp2_label && echo Core 1 > /tmp/hwmon/hwmon2/temp3_label")
+        m.execute("echo 45000 > /tmp/hwmon/hwmon2/temp1_input && echo 50000 > /tmp/hwmon/hwmon2/temp2_input && echo 110000 > /tmp/hwmon/hwmon2/temp3_input")
+
+        b.wait_visible("#current-metrics-card-cpu .text-color-critical")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-critical", "110 °C")
+
+        m.execute("echo 120000 > /tmp/hwmon/hwmon1/temp2_input")
+        b.wait_visible("#current-metrics-card-cpu .text-color-critical")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-critical", "120 °C")
+
+        m.execute("umount /sys/class")
+        m.execute("rm -rf /tmp/hwmon")
+
         # Test link to user services
         # older releases don't have CPU accounting enabled for user services
         if m.image not in ["rhel-8-7", "centos-8-stream"]:


pkg/metrics/metrics.jsx
```diff
@@ -492,7 +518,10 @@ class CurrentMetrics extends React.Component {
         return (
             <Gallery className="current-metrics" hasGutter>
                 <Card id="current-metrics-card-cpu">
-                    <CardTitle>{ _("CPU") }</CardTitle>
+                    <CardHeader>
+                        { isNaN(this.cpuTemperature) ? null : <CardActions>{ this.cpuTemperature + " °C"}</CardActions> }
```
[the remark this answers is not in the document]
martinpitt: I remember rather angry UK users who (rightfully to a large degree) complain about US-isms like "broken date format" in our UI. Supporting en-GB separately is not a drive-by fix, though. Possibly we should just show both Celsius and Fahrenheit for the time being?

Also, please don't combine strings that way. You need to use `cockpit.format(_("$0 °C"), temp)` so that this can be localized. This is even more important if you make the string more complex.

jelly:
I'm wondering if we want to extend `check-metrics` with a mocked thermal_zone0, this can be done as following:

```
cd /tmp
mkdir -p thermal/thermal_zone0
echo '99000' > thermal/thermal_zone0/temp
mount -o bind /tmp/thermal /sys/devices/virtual/thermal
```

martinpitt:
Thanks @tomasmatus ! Please add an integration test for it. Schema for mocking it:
```sh
mkdir -p /tmp/thermal/thermal_zone0
echo 42000 > /tmp/thermal/thermal_zone0/temp
mount -o bind /tmp/thermal /sys/devices/virtual/thermal
```

and clean up with `umount /sys/devices/virtual/thermal`. It could be that our VMs don't have /sys/devices/virtual/thermal/ already, then you'll have to mount it on the parent dir.

garrett:
I can't test this locally, but the temperature seems a bit too large? I thought I had it a smaller size in the mockups.

garrett:
> Do we want to show the temperature in red when it reaches a certain threshold say 90C?

We cannot rely on color alone. We should probably have a warning icon when it gets too high.

Is 90 the cutoff for too hot? I thought it was > 80?

---

Let's look around:

https://techreviewteam.com/hardware/what-cpu-temp-is-too-high-how-to-fix-it/

> The highest temperature we recommend would be 80º Celsius, anything above 80º C becomes extremely dangerous to your system & processor since 100º C is the boiling point. Even running at 80º C is already pushing and we recommend you to try and keep it lower than that at all times, even during gaming.
> 
> If your processor temp is constantly above or at 80º Celsius, there may be something wrong with your PC hardware or its configuration settings.
> 
> We recommend that you play a very resource-intensive game while monitoring the temperatures, if it reaches 80ºC or goes past it then it’s a problem you need to solve, if it stays around 70ºC or less it should be ok and you don’t need to worry about it.

https://www.wepc.com/cpu/how-to/reduce-cpu-temperatures/

> A ‘standard’ gaming PC (if you can even standardize such a thing) should be seeing **temperatures between 158°F (70°C) and 176°F (80°C)** while under heavy CPU load.

https://www.maketecheasier.com/troubleshoot-high-cpu-temperature/

> How Hot Should My CPU Be?
>
> This is where things get a little complicated. Different CPUs are built in different ways; as such, they each have their limits as to how far you can push them. A temperature of 80°C, for instance, can be shrugged off by some processors and seriously damage others.

https://www.pcgamer.com/cpu-temperature-overheat/

> If you’re running stock and your CPU temperatures are hitting 80°C or higher, that’s a warning sign something isn’t working properly. Maybe a fan is dead, or thermal paste wasn’t applied properly, or your heatsink is super dusty.

https://www.tomshardware.com/how-to/how-to-monitor-cpu-temp-temperature

> Under load, you want your CPU to ideally stay under 80 degrees Celsius (176 degrees Fahrenheit), though some CPUs may run hotter when they're in ultrabooks, gaming laptops, or small-form-factor (SFF) computers. You have some wiggle room to creep past 80 degrees Celsius, but anything above 95 degrees Celsius (203 degrees Fahrenheit) is critical. At this point, some CPUs will begin throttling, meaning the clock speed will slow down to ensure it doesn't overheat, and your PC may turn off.

---

Looking at some of the top sites on this topic all are in agreement that:
- >80 is bad
- temperatures are all in °C

Some mention that 95 is critical. Some state that modern CPUs will throttle (but it's still important to check).

I'd say we show a warning icon and color (orange) when temperature is > 80 and an danger icon and color (red) at >= 95.

jelly:
> > Do we want to show the temperature in red when it reaches a certain threshold say 90C?
> 
> We cannot rely on color alone. We should probably have a warning icon when it gets too high.
> 
> Is 90 the cutoff for too hot? I thought it was > 80?
> 
> Let's look around:
> 
> https://techreviewteam.com/hardware/what-cpu-temp-is-too-high-how-to-fix-it/
> 
> > The highest temperature we recommend would be 80º Celsius, anything above 80º C becomes extremely dangerous to your system & processor since 100º C is the boiling point. Even running at 80º C is already pushing and we recommend you to try and keep it lower than that at all times, even during gaming.
> > If your processor temp is constantly above or at 80º Celsius, there may be something wrong with your PC hardware or its configuration settings.
> > We recommend that you play a very resource-intensive game while monitoring the temperatures, if it reaches 80ºC or goes past it then it’s a problem you need to solve, if it stays around 70ºC or less it should be ok and you don’t need to worry about it.
> 
> https://www.wepc.com/cpu/how-to/reduce-cpu-temperatures/
> 
> > A ‘standard’ gaming PC (if you can even standardize such a thing) should be seeing **temperatures between 158°F (70°C) and 176°F (80°C)** while under heavy CPU load.
> 
> https://www.maketecheasier.com/troubleshoot-high-cpu-temperature/
> 
> > How Hot Should My CPU Be?
> > This is where things get a little complicated. Different CPUs are built in different ways; as such, they each have their limits as to how far you can push them. A temperature of 80°C, for instance, can be shrugged off by some processors and seriously damage others.
> 
> https://www.pcgamer.com/cpu-temperature-overheat/
> 
> > If you’re running stock and your CPU temperatures are hitting 80°C or higher, that’s a warning sign something isn’t working properly. Maybe a fan is dead, or thermal paste wasn’t applied properly, or your heatsink is super dusty.
> 
> https://www.tomshardware.com/how-to/how-to-monitor-cpu-temp-temperature
> 
> > Under load, you want your CPU to ideally stay under 80 degrees Celsius (176 degrees Fahrenheit), though some CPUs may run hotter when they're in ultrabooks, gaming laptops, or small-form-factor (SFF) computers. You have some wiggle room to creep past 80 degrees Celsius, but anything above 95 degrees Celsius (203 degrees Fahrenheit) is critical. At this point, some CPUs will begin throttling, meaning the clock speed will slow down to ensure it doesn't overheat, and your PC may turn off.
> 
> Looking at some of the top sites on this topic all are in agreement that:
> 
> * > 80 is bad
> * temperatures are all in °C
> 
> Some mention that 95 is critical. Some state that modern CPUs will throttle (but it's still important to check).
> 
> I'd say we show a warning icon and color (orange) when temperature is > 80 and an danger icon and color (red) at >= 95.

We do support other devices as well (RPI, etc.), so I think it's a bit tricky to come up with something. Sysfs does have a `trip point` which for me is 128 C?!!

https://www.kernel.org/doc/html/latest/driver-api/thermal/sysfs-api.html#sysfs-attributes-structure

And sensors is fun as well:

```
Package id 0:  +47.0°C  (high = +100.0°C, crit = +100.0°C)
Core 0:        +47.0°C  (high = +100.0°C, crit = +100.0°C)
Core 1:        +46.0°C  (high = +100.0°C, crit = +100.0°C)
Core 2:        +47.0°C  (high = +100.0°C, crit = +100.0°C)
Core 3:        +47.0°C  (high = +100.0°C, crit = +100.0°C)
```

tomasmatus:
> I can't test this locally, but the temperature seems a bit too large? I thought I had it a smaller size in the mockups.

You are right. I didn't notice/realize it's smaller in this mockup.
https://github.com/cockpit-project/cockpit/issues/11364#issuecomment-716819011

Should I make it smaller or do you think it looks better when it's the same size as "CPU" text on the left?

garrett:
Here's the md size, but alignment is incorrect

image

pkg/metrics/metrics.jsx
```diff
@@ -493,7 +526,19 @@ class CurrentMetrics extends React.Component {
...
+                    <CardHeader>
+                        { !isNaN(this.cpuTemperature) &&
+                        <CardActions>
+                            <HelperText>
+                                <HelperTextItem hasIcon variant={
+                                    this.cpuTemperature < 80 ? "warning" : (this.cpuTemperature < 95 ? "warning" : "error")
+                                }>
+                                    { cockpit.format("$0 °C", this.cpuTemperature) }
+                                </HelperTextItem>
+                            </HelperText>
```
jelly, line 538: This should not use a HelperText as it is not part of a form. Most likely you want to put the Icon and text in a Flex, and then `align-items: baseline` should work. 

baseline now does not work as CSS aligns the outer html element and not the text within the `<CardActions>`

jelly (changes requested):

test/verify/check-metrics
```diff
@@ -758,6 +758,28 @@ class TestCurrentMetrics(MachineCase):
         with b.wait_timeout(300):
             b.wait(lambda: float(b.text("#load-avg .pf-l-flex div:first-child").split()[-1].rstrip(',')) < 10)
 
+        # File with CPU temperature does not exist, nothnig is displayed
+        b.wait_not_in_text("#current-metrics-card-cpu", "42 °C")
```
martinpitt, line 762: This is too specific for testing a negative. This should be more precise, such as `wait_text(..., "")`.

pkg/metrics/metrics.jsx
```diff
@@ -333,6 +358,32 @@ class CurrentMetrics extends React.Component {
...
+        this.cpuTemperature = parseInt(this.temperatureSamples[0]);
+
+        if (this.cpuTemperature <= 80) {
+            this.cpuTemperatureColors.textColor = "";
+            this.cpuTemperatureColors.iconColor = "";
+            this.cpuTemperatureColors.icon = null;
+        } else if (this.cpuTemperature < 95) {
+            this.cpuTemperatureColors.textColor = "text-color-warning";
+            this.cpuTemperatureColors.iconColor = "icon-color-warning";
+            this.cpuTemperatureColors.icon = <ExclamationTriangleIcon className="fa-exclamation-triangle" />;
```
martinpitt, line 379: These class names are confusing. `fa-*` were the old-style FontAwesome icon names. But this has nothing to do with the icon shape, I suppose it's just for the tests to validate the current warning level?  Could you name these `temperature-{warning,critical}`?

src/bridge/cockpitinternalmetrics.c
```diff
@@ -73,6 +73,7 @@ static MetricDescription metric_descriptions[] = {
   { "cpu.core.user",   "millisec", "counter", TRUE, CPU_SAMPLER },
   { "cpu.core.system", "millisec", "counter", TRUE, CPU_SAMPLER },
   { "cpu.core.iowait", "millisec", "counter", TRUE, CPU_SAMPLER },
+  { "cpu.temperature", "celsius", "counter", FALSE, CPU_SAMPLER },
```
allisonkarlitskaya, line 76: This is not a `counter` value.

I also feel like this is an awkward fit.

All of the existing CPU samples are counters in milliseconds, and are available per-core and "basic" summary values.

I'd prefer to see a new sampler created for sensor information.  We might imagine reporting other temperatures in the future...

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,29 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+void
+cockpit_cpu_temperature (CockpitSamples *samples)
+{
+  gchar *contents = NULL;
+  GError *error = NULL;
+  gsize len;
+  guint32 temperature;
+
+  // file does not exist in virtual machines
+  if (g_file_test ("/sys/devices/virtual/thermal/thermal_zone0/temp", G_FILE_TEST_EXISTS))
```
allisonkarlitskaya, line 125: This is also a bit of a bad pattern — rather just try to open the file.  If you get a `ENOENT` (see `g_error_matches()`) then ignore it, otherwise report it.

allisonkarlitskaya: And: how do we know for sure that zone0 is definitely the CPU?

garrett:
```sh
~ ❱ tail /sys/devices/virtual/thermal/thermal_zone*/type
==> /sys/devices/virtual/thermal/thermal_zone0/type <==
acpitz

==> /sys/devices/virtual/thermal/thermal_zone1/type <==
INT3400 Thermal

==> /sys/devices/virtual/thermal/thermal_zone2/type <==
SEN1

==> /sys/devices/virtual/thermal/thermal_zone3/type <==
pch_cannonlake

==> /sys/devices/virtual/thermal/thermal_zone4/type <==
B0D4

==> /sys/devices/virtual/thermal/thermal_zone5/type <==
iwlwifi_1

==> /sys/devices/virtual/thermal/thermal_zone6/type <==
x86_pkg_temp
```

0 is `acpitz` ("ACPI thermal zone")

At least here, my CPU "package" (`x86_pkg_temp`: CPU, iGPU, etc.) is zone6.

image


Temperatures differ quite a bit too:

```
~ ❱ tail /sys/devices/virtual/thermal/thermal_zone*/temp
==> /sys/devices/virtual/thermal/thermal_zone0/temp <==
52000

==> /sys/devices/virtual/thermal/thermal_zone1/temp <==
20000

==> /sys/devices/virtual/thermal/thermal_zone2/temp <==
40050

==> /sys/devices/virtual/thermal/thermal_zone3/temp <==
51000

==> /sys/devices/virtual/thermal/thermal_zone4/temp <==
52050

==> /sys/devices/virtual/thermal/thermal_zone5/temp <==
42000

==> /sys/devices/virtual/thermal/thermal_zone6/temp <==
51000
```

However, right now, zone 0 is 52 and zone 6 is 51.

(I also wonder what Windows and macOS show for temperatures.)

jelly:
If we only want to get the CPU temperature there is one way to achieve this to sysfs and it is not thermal zones.

From my understanding thermal zones are zones inside a computer which are defined by ACPI (bios) your bios uses this to monitor the systems temperature zones such as CPU, southbridge, northbridge, etc. and acts on high temperatures by spinning up a fan etc. For cockpit there isn't much to act on if these temperatures are too high.

Another sysfs interface is hwmon for which different manufactures have different types of sensors to measure their specific hardware sensors. Hwmon also exports thermal_zones, For CPU temperature this is:

* For intel devices coretemp `/sys/class/hwmon/hwmon8/name` => coretemp (note that the location is not predictable)
```
[jelle@t14s][~]%cat --style=plain /sys/class/hwmon/hwmon8/temp?_label
Package id 0
Core 0
Core 1
Core 2
Core 3
```
This shows the 4 core temperatures and `Package id 0` is the package temperature, this is driver specific.
https://www.kernel.org/doc/html/latest/hwmon/coretemp.html

* On AMD the driver is either k8temp or k10temp (Example on AMD EPYC 7502P 32-Core Processor)

```
[jelle@build ~]# cat /sys/class/hwmon/hwmon1/temp?_label
Tctl
Tccd1
Tccd3
Tccd5
Tccd7
[jelle@build ~]# cat /sys/class/hwmon/hwmon1/name
k10temp
```

Here Tctl is a very special temperature control sensorand not useful for us, see more details in the driver's docs:

https://www.kernel.org/doc/html/latest/hwmon/k10temp.html

* On my RPI zero W

```
pi@raspberrypi:~ $ cat /sys/class/hwmon/hwmon0/name
cpu_thermal
pi@raspberrypi:~ $ cat /sys/class/hwmon/hwmon0/temp1_input
35242
```

It seems that these on ARM (at least rpi) are thermal_zones exported in hwmon as cpu_thermal

```
/sys/class/thermal/thermal_zone0/hwmon0/name
cpu_thermal
```

The sensor to use seems to be defined in device tree which means it's arbitrary per board.

https://www.kernel.org/doc/html/latest/hwmon/sysfs-interface.html

garrett:
On my Intel laptop, it's different:

```sh
~ ❱ tail /sys/class/hwmon/hwmon*/name
==> /sys/class/hwmon/hwmon0/name <==
AC

==> /sys/class/hwmon/hwmon1/name <==
acpitz

==> /sys/class/hwmon/hwmon2/name <==
BAT0

==> /sys/class/hwmon/hwmon3/name <==
thinkpad

==> /sys/class/hwmon/hwmon4/name <==
nvme

==> /sys/class/hwmon/hwmon5/name <==
ucsi_source_psy_USBC000:001

==> /sys/class/hwmon/hwmon6/name <==
ucsi_source_psy_USBC000:002

==> /sys/class/hwmon/hwmon7/name <==
hid_CC2901200PKJ5R9AX_battery

==> /sys/class/hwmon/hwmon8/name <==
pch_cannonlake

==> /sys/class/hwmon/hwmon9/name <==
coretemp

==> /sys/class/hwmon/hwmon10/name <==
iwlwifi_1

==> /sys/class/hwmon/hwmon11/name <==
hidpp_battery_2
```

(Note: There's no `temp?_label` for `hwmon8` here. It's `/sys/class/hwmon/hwmon9/temp1_label`)

```sh
~ ❱ tail /sys/class/hwmon/hwmon*/temp*label
==> /sys/class/hwmon/hwmon3/temp1_label <==
CPU

==> /sys/class/hwmon/hwmon3/temp2_label <==
GPU

==> /sys/class/hwmon/hwmon4/temp1_label <==
Composite

==> /sys/class/hwmon/hwmon4/temp2_label <==
Sensor 1

==> /sys/class/hwmon/hwmon9/temp1_label <==
Package id 0

==> /sys/class/hwmon/hwmon9/temp2_label <==
Core 0

==> /sys/class/hwmon/hwmon9/temp3_label <==
Core 1

==> /sys/class/hwmon/hwmon9/temp4_label <==
Core 2

==> /sys/class/hwmon/hwmon9/temp5_label <==
Core 3
```

There are a bunch of `temp?_inputs` here:

```sh
~ ❱ tail /sys/class/hwmon/hwmon*/temp*input

==> /sys/class/hwmon/hwmon1/temp1_input <==
47000

==> /sys/class/hwmon/hwmon3/temp1_input <==
47000

==> /sys/class/hwmon/hwmon3/temp2_input <==
tail: error reading '/sys/class/hwmon/hwmon3/temp2_input': No such device or address

==> /sys/class/hwmon/hwmon3/temp3_input <==
38000

==> /sys/class/hwmon/hwmon3/temp4_input <==
0

==> /sys/class/hwmon/hwmon3/temp5_input <==
47000

==> /sys/class/hwmon/hwmon3/temp6_input <==
47000

==> /sys/class/hwmon/hwmon3/temp7_input <==
47000

==> /sys/class/hwmon/hwmon3/temp8_input <==
tail: error reading '/sys/class/hwmon/hwmon3/temp8_input': No such device or address

==> /sys/class/hwmon/hwmon4/temp1_input <==
38850

==> /sys/class/hwmon/hwmon4/temp2_input <==
38850

==> /sys/class/hwmon/hwmon8/temp1_input <==
47000

==> /sys/class/hwmon/hwmon9/temp1_input <==
47000

==> /sys/class/hwmon/hwmon9/temp2_input <==
46000

==> /sys/class/hwmon/hwmon9/temp3_input <==
46000

==> /sys/class/hwmon/hwmon9/temp4_input <==
46000

==> /sys/class/hwmon/hwmon9/temp5_input <==
47000

==> /sys/class/hwmon/hwmon10/temp1_input <==
44000
```

Oddly, I get an error in the above list! Extracted here for emphasis:

```sh
tail: error reading '/sys/class/hwmon/hwmon3/temp2_input': No such device or address
```

Temp1 looks like this (basically a filtered version of the list directly above):

```sh
~ ❱ tail /sys/class/hwmon/hwmon*/temp1*input

==> /sys/class/hwmon/hwmon1/temp1_input <==
49000

==> /sys/class/hwmon/hwmon3/temp1_input <==
49000

==> /sys/class/hwmon/hwmon4/temp1_input <==
36850

==> /sys/class/hwmon/hwmon8/temp1_input <==
49000

==> /sys/class/hwmon/hwmon9/temp1_input <==
50000

==> /sys/class/hwmon/hwmon10/temp1_input <==
41000
```

garrett:
In my Raspberry Pi 4 running Fedora, it's device 1, not 0:

```
~> tail /sys/class/hwmon/hwmon*/name

==> /sys/class/hwmon/hwmon0/name <==
rpi_volt

==> /sys/class/hwmon/hwmon1/name <==
cpu_thermal
```

But on the Pi 1, it's flipped. But it's Debian:

```
~> tail /sys/class/hwmon/hwmon*/name

==> /sys/class/hwmon/hwmon0/name <==
cpu_thermal

==> /sys/class/hwmon/hwmon1/name <==
rpi_volt
```

We definitely need to parse and match; this just reinforces it. At least the names seem consistent on each architecture.

garrett:
At least for Intel, I don't think we can reliably use the max, crit, and crit_alarm values:

==> /sys/class/hwmon/hwmon9/temp1_crit <==
100000

==> /sys/class/hwmon/hwmon9/temp1_crit_alarm <==
0

==> /sys/class/hwmon/hwmon9/temp1_input <==
51000

==> /sys/class/hwmon/hwmon9/temp1_label <==
Package id 0

==> /sys/class/hwmon/hwmon9/temp1_max <==
100000

That's 100, which is a bit too late, isn't it? And crit_alarm seems either like a boolean (I guess 1 for audible beep or 0 for not?) or a bug.

garrett:
> SandyBridge

This means Intel >= 2011. Most people using Cockpit on Intel based CPUs are probably doing so on machines made in the past decade, right? But I'm sure not everyone is.

I do have a very old MacBook Pro that's a Core 2 Duo from mid-2009. I suppose I should dust it off and see what happens as a corner case. I don't know what version of Linux it has installed (as it's been a while), but I can always do a live boot of Fedora. :wink:

garrett:
I've booted my [MacBook Pro Mid-2009 (5,3)](https://support.apple.com/kb/SP544?locale=en_US) (an "Intel(R) Core(TM)2 Duo CPU     T9600  @ 2.80GHz") into Fedora 36 beta from a liveUSB stick and checked.

Here are the hwmon names:

```
==> /sys/class/hwmon/hwmon0/name <==
ADP1

==> /sys/class/hwmon/hwmon1/name <==
BAT0

==> /sys/class/hwmon/hwmon2/name <==
BAT0

==> /sys/class/hwmon/hwmon3/name <==
nouveau

==> /sys/class/hwmon/hwmon4/name <==
nouveau

==> /sys/class/hwmon/hwmon5/name <==
coretemp
```

Here are the temperature labels:

```
==> /sys/class/hwmon/hwmon2/temp1_label <==
temp

==> /sys/class/hwmon/hwmon3/in0_label <==
GPU core

==> /sys/class/hwmon/hwmon5/temp2_label <==
Core 0

==> /sys/class/hwmon/hwmon5/temp3_label <==
Core 1
```
Here are the current values:

```
==> /sys/class/hwmon/hwmon1/temp1_input <==
30400

==> /sys/class/hwmon/hwmon2/curr1_input <==
1162

==> /sys/class/hwmon/hwmon2/in0_input <==
12556

==> /sys/class/hwmon/hwmon2/temp1_input <==
30400

==> /sys/class/hwmon/hwmon3/in0_input <==
900

==> /sys/class/hwmon/hwmon3/temp1_input <==
61000

==> /sys/class/hwmon/hwmon4/temp1_input <==
59000

==> /sys/class/hwmon/hwmon5/temp2_input <==
53000

==> /sys/class/hwmon/hwmon5/temp3_input <==
57000
```

In this case, it looks like it's hwmon5, aka `coretemp`. It doesn't have an aggregate temperature; it has a per-core temperature of 53 and 57. Meanwhile, the first NVidia GPU (yes, it has 2 NVidia GPUs) is running the hottest @ 61.

Since it had a super old installation of Linux, I told Fedora 36 to install on it. One of the fans is bad... and during installation both fans kicked into high gear (including the bad one, which makes frightening noises) and I had CPU readings of 86 and 98 (!!!). Whoops.

I also have an atom netbook that has both an `acpitz` and `coretemp`. It only has one core.

jelly:
> I've booted my [MacBook Pro Mid-2009 (5,3)](https://support.apple.com/kb/SP544?locale=en_US) (an `Intel(R) Core(TM)2 Duo CPU T9600 @ 2.80GHz` ×2) into Fedora 36 beta from a liveUSB stick and checked.
> 
> Here are the hwmon names:
> 
> ```
> ==> /sys/class/hwmon/hwmon0/name <==
> ADP1
> 
> ==> /sys/class/hwmon/hwmon1/name <==
> BAT0
> 
> ==> /sys/class/hwmon/hwmon2/name <==
> BAT0
> 
> ==> /sys/class/hwmon/hwmon3/name <==
> nouveau
> 
> ==> /sys/class/hwmon/hwmon4/name <==
> nouveau
> 
> ==> /sys/class/hwmon/hwmon5/name <==
> coretemp
> ```
> 
> Here are the temperature labels:
> 
> ```
> ==> /sys/class/hwmon/hwmon2/temp1_label <==
> temp
> 
> ==> /sys/class/hwmon/hwmon3/in0_label <==
> GPU core
> 
> ==> /sys/class/hwmon/hwmon5/temp2_label <==
> Core 0
> 
> ==> /sys/class/hwmon/hwmon5/temp3_label <==
> Core 1
> ```
> 
> Here are the current values:
> 
> ```
> ==> /sys/class/hwmon/hwmon1/temp1_input <==
> 30400
> 
> ==> /sys/class/hwmon/hwmon2/curr1_input <==
> 1162
> 
> ==> /sys/class/hwmon/hwmon2/in0_input <==
> 12556
> 
> ==> /sys/class/hwmon/hwmon2/temp1_input <==
> 30400
> 
> ==> /sys/class/hwmon/hwmon3/in0_input <==
> 900
> 
> ==> /sys/class/hwmon/hwmon3/temp1_input <==
> 61000
> 
> ==> /sys/class/hwmon/hwmon4/temp1_input <==
> 59000
> 
> ==> /sys/class/hwmon/hwmon5/temp2_input <==
> 53000
> 
> ==> /sys/class/hwmon/hwmon5/temp3_input <==
> 57000
> ```
> 
> In this case, it looks like it's hwmon5, aka `coretemp`. It doesn't have an aggregate temperature; it has a per-core temperature of 53 and 57. Meanwhile, the first NVidia GPU (yes, it has 2 NVidia GPUs) is running the hottest @ 61.
> 
> Since it had a super old installation of Linux, I told Fedora 36 to install on it. One of the fans is bad... and during installation both fans kicked into high gear (including the bad one, which makes frightening noises) and I had CPU readings of 86 and 98 (!!!). Whoops.
> 
> I also have an atom netbook that has both an `acpitz` and `coretemp`. It only has one core.

This is to be expected, the driver mentions that all pre-skylake models don't have a "package temperature sensor".

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +113,84 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+  gchar *temp_content = NULL;
+  gchar *label_content = NULL;
+  GError *error = NULL;
+  gsize len;
+  gchar path[CPU_STR_MAXLEN];
+  gchar label[CPU_STR_MAXLEN];
+  gchar name[CPU_STR_MAXLEN];
+  guint32 temperature;
+
+  for (int i = 0; error == NULL; i++)
```
allisonkarlitskaya, line 130: I feel like maybe `for` is the wrong choice of loop here, and certainly using the condition like this makes things strange.  Some kind of an explicitly infinite-looping construct with well-placed `break;` on errors would probably read a lot nicer.

src/bridge/cockpitinternalmetrics.c
```diff
@@ -53,7 +53,8 @@ typedef enum {
   NETWORK_SAMPLER = 1 << 3,
   MOUNT_SAMPLER = 1 << 4,
   CGROUP_SAMPLER = 1 << 5,
-  DISK_SAMPLER = 1 << 6
+  DISK_SAMPLER = 1 << 6,
+  THERMAL_SAMPLER = 1 << 7
```
allisonkarlitskaya, line 57: Can you add a trailing `,` here?

src/bridge/cockpitinternalmetrics.c
```diff
@@ -100,6 +101,8 @@ static MetricDescription metric_descriptions[] = {
   { "cgroup.cpu.usage",       "millisec", "counter", TRUE, CGROUP_SAMPLER },
   { "cgroup.cpu.shares",      "count",    "instant", TRUE, CGROUP_SAMPLER },
 
+  { "cpu.temperature", "celsius", "counter", TRUE, THERMAL_SAMPLER },
```
allisonkarlitskaya, line 104: It's still not a counter :)

+ please try to line it up with the ones above...

src/bridge/cockpitinternalmetrics.c
```diff
@@ -321,7 +324,7 @@ cockpit_internal_metrics_tick (CockpitMetrics *metrics,
   /* Sample
    */
   if (self->samplers & CPU_SAMPLER)
-    cockpit_cpu_samples (COCKPIT_SAMPLES (self));
+      cockpit_cpu_samples (COCKPIT_SAMPLES (self));
```
allisonkarlitskaya, line 327: This looks like an accidental edit.

src/bridge/cockpitinternalmetrics.c
```diff
@@ -334,6 +337,8 @@ cockpit_internal_metrics_tick (CockpitMetrics *metrics,
     cockpit_cgroup_samples (COCKPIT_SAMPLES (self));
   if (self->samplers & DISK_SAMPLER)
     cockpit_disk_samples (COCKPIT_SAMPLES (self));
+  if (self->samplers & THERMAL_SAMPLER)
+    cockpit_cpu_temperature(COCKPIT_SAMPLES (self));
```
allisonkarlitskaya, line 341: Space after function name and before `(` please.

src/bridge/test-metrics.c
```diff
@@ -850,6 +850,66 @@ test_cpu_cores (void)
...
+  CockpitChannel *channel;
+
+  JsonObject *options = json_obj ("{ 'metrics': [ { 'name': 'cpu.temperature' } ],"
+                                  "  'interval': 1000"
+                                  "}");
+
+  GBytes *msg;
+  JsonObject *res, *description;
+  JsonArray *metrics, *samples, *temperature, *core;
+  GError *error = NULL;
```
allisonkarlitskaya, line 866: more `g_autoptr()` and declare-at-site-of-first-use please.

allisonkarlitskaya (changes requested):
Sorry for taking so long.  For some reason GitHub didn't notify me that you re-requested.

Initial round of C comments.  Most of these are in the category of "this isn't how we do things in G-world".  Happy to talk about that on IRC if you like :)

src/bridge/test-metrics.c
```diff
@@ -850,6 +850,58 @@ test_cpu_cores (void)
...
+static void
+test_cpu_temperature (void)
+{
+  MockTransport *transport = mock_transport_new ();
+
+  g_autoptr (JsonObject) options = json_obj ("{ 'metrics': [ { 'name': 'cpu.temperature' } ],"
+                                  "  'interval': 1000"
+                                  "}");
+
+  g_signal_connect (transport, "closed", G_CALLBACK (on_transport_closed), NULL);
```
allisonkarlitskaya, line 862: It's very strange that this signal handler is between the site where the JSON is parsed and where it's used.  Please move it up or down.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,72 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+        {
+          if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+              g_message ("error loading name: %s", error->message);
+
+          break;
+        }
+
+      g_strchomp(name);
+
+      if (!g_str_equal (name, "coretemp") && !g_str_equal (name, "cpu_thermal") &&
```
allisonkarlitskaya, line 136: Comment explaining this, please :)

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,72 @@ cockpit_cpu_samples (CockpitSamples *samples)
   g_strfreev (lines);
   g_free (contents);
 }
+
+void
+cockpit_cpu_temperature (CockpitSamples *samples)
+{
+  for (int i = 0; TRUE; i++)
+    {
+      gsize len;
```
allisonkarlitskaya, line 121: The length field is optional, and you're not using it, so just pass NULL instead.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,72 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+      g_autoptr (GError) error = NULL;
+      g_autofree gchar *path = g_strdup_printf ("/sys/class/hwmon/hwmon%d/name", i);
+      g_autofree gchar *name = NULL;
+
+      if (!g_file_get_contents (path, &name, &len, &error))
```
allisonkarlitskaya, line 126: This pattern of "read the file, and ignore it if it doesn't exist, otherwise abort with an error" is repeated quite a lot.  It probably deserves a helper function of some kind.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,72 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+      g_strchomp(name);
+
+      if (!g_str_equal (name, "coretemp") && !g_str_equal (name, "cpu_thermal") &&
+          !g_str_equal (name, "k8temp") && !g_str_equal (name, "k10temp"))
+        {
+          // doesn't match any supported CPU
+          continue;
+        }
+
+      for (int j = 1; TRUE; j++)
```
allisonkarlitskaya, line 143: I'm not a fan of this nested loop structure, for a couple of reasons.

One reason is that it makes it makes it complicated to understand what is supposed to happen in case we hit an error when reading the temperature file of a particular hwmon device.  I think the intent here is that we log the error, clear it, and then go back to searching for more temperature devices?  Is it expected that there are multiple matching temperature devices on the system?

That brings to the second reason: in the event that there are multiple matching temperature devices on the system, we'll set the temperature property on the samples object twice.

This function is also just quite long.

Depending on if it's expected that we might have multiple matching devices, I recommend a structure where you first determine the device to use, and then query it.

You might split that into 3 or 4 functions, as so:
 - the read-a-possibly-non-existing-file helper function
 - a function which finds the hwmon device that we want to use
 - a function to read the temperature out of that device
 - a main function that ties it all together

Otherwise if we do intend to query multiple devices, in case one of them fails, it might be nice to express that logic plainly in the main function, and rather have a simple helper function for deciding "is this a valid hwmon device?"

src/bridge/test-metrics.c
```diff
@@ -850,6 +850,58 @@ test_cpu_cores (void)
   g_object_unref (transport);
 }
 
+static void
+test_cpu_temperature (void)
+{
+  MockTransport *transport = mock_transport_new ();
+
+  g_autoptr (JsonObject) options = json_obj ("{ 'metrics': [ { 'name': 'cpu.temperature' } ],"
```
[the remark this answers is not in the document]
tomasmatus: ACK. Why is it the case with `g_autoptr` but not with, say, `g_assert`?

src/bridge/test-metrics.c
```diff
@@ -850,6 +850,58 @@ test_cpu_cores (void)
...
+  while ((msg = mock_transport_pop_channel (transport, "1234")) == NULL)
+    g_main_context_iteration (NULL, TRUE);
+
+  g_autoptr (GError) error = NULL;
+  g_autoptr (JsonObject) res = cockpit_json_parse_bytes (msg, &error);
+  g_assert_no_error (error);
+  g_assert (res != NULL);
+  g_assert (json_object_has_member (res, "metrics"));
+
+  JsonArray *metrics = json_object_get_array_member (res, "metrics");
```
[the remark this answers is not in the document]
tomasmatus: To avoid double (or more) frees on same memory. Here `metrics` points to a part of memory which is freed with `g_autoptr(JsonObject) res` at line 878.

src/bridge/cockpitcpusamples.c
tomasmatus: > That brings to the second reason: in the event that there are multiple matching temperature devices on the system, we'll set the temperature property on the samples object twice.

Yes, multiple matching devices are expected and sample object is always unique.

Sample for intel and amd looks like this:
`hwmon(NUM) (label)`
where `NUM` is number of current hwmon directory and `label` is unique label in current directory.
As arm doesn't use labels I create a unique one using the counter from second loop.

> This function is also just quite long.

This is true, I'll try to refactor it and split into multiple functions as you suggest.

Thanks for the review!

src/bridge/test-metrics.c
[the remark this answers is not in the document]
allisonkarlitskaya: `g_assert` is something that's vaguely function-like whereas `g_autoptr(type)` is meant to be treated as part of the type name

src/bridge/cockpitcpusamples.c
jelly: Or what about doing the "filtering" in JavaScript and pass all temperature values to the client?

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,76 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+
+gchar*
+read_hwmon_file (gchar *path, GError **error)
+{
+  gchar *content = NULL;
+
+  if (!g_file_get_contents (path, &content, NULL, error))
+    {
+      if (!g_error_matches (*error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+          g_message ("error loading name: %s", (*error)->message);
```
allisonkarlitskaya, line 124: So more GLib conventions time.

It's customary that `error` here might be `NULL`, and even if we know that it never is, you should write the function in a way that it might be.

The way that usually works is by using a `g_autoptr(GError) my_error` and then checking the condition that you want to check.  In case you want to re-raise the error.  That looks like `g_propagate_error (error, g_steal_pointer (&my_error));`.  Otherwise, just do nothing, and the error gets freed automatically.

Which brings to the next point: what do you intend to do differently if you get `NOENT` here?  There are a couple of options.  Although it's slightly non-idiomatic, I feel like it would be okayish to return `NULL` here in case of `ENOENT`, but not propagate the error.  In case of "hard" errors, you would propagate.

If you want to be hardcore idiomatic (ie: this is "proper GLib-style API"), you would probably write a function with a signature like

```
gboolean
read_hwmon_file (const gchar *path,
                 gchar **contents,
                 GError **error);
```

which would perform as follows:

 - successfully read file: set `*contents` and return `TRUE`
 - `ENOENT`: set `*contents = NULL` and return `TRUE`
 - other error: propagate error, `*contents` undefined, return `FALSE`.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,76 @@ cockpit_cpu_samples (CockpitSamples *samples)
   g_strfreev (lines);
   g_free (contents);
 }
+
+gchar*
+read_hwmon_file (gchar *path, GError **error)
```
allisonkarlitskaya, line 117: `path` should be `const` here.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,76 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+void
+sample_cpu_sensors (CockpitSamples *samples, int hwmonID)
+{
+  for (int i = 1; TRUE; i++)
+    {
+      g_autoptr(GError) error = NULL;
+      g_autofree gchar *temp = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_input", hwmonID, i);
+      g_autofree gchar *temp_content = read_hwmon_file (temp, &error);
+
+      if (error != NULL)
```
allisonkarlitskaya, line 139: The way this is currently written, this will match `ENOENT` as well. See monster comment above for some options.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,76 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+        }
+
+      g_strchomp (label_content);
+
+      // ignore Tctl on AMD devices
+      if (!g_str_equal (label_content, "Tctl"))
+        {
+          guint32 temperature;
+          g_autofree gchar *instance = g_strdup_printf ("hwmon%d %s", hwmonID, label_content);
+          sscanf (temp_content, "%"G_GUINT32_FORMAT, &temperature);
```
allisonkarlitskaya, line 159: I guess this is also more usually written like `g_ascii_stroull()` or friends...

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,76 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+          break;
+
+      g_strchomp(name);
+
+      // compare device name with CPU names
+      if (g_str_equal (name, "coretemp") || g_str_equal (name, "cpu_thermal") ||
+          g_str_equal (name, "k8temp") || g_str_equal (name, "k10temp"))
+        {
+          // hwmon contains CPU info
+          sample_cpu_sensors(samples, i);
```
allisonkarlitskaya, line 184: space before `(`

allisonkarlitskaya (changes requested):
I think my overall impression here is that the logic is a *lot* easier to understand since the last one, but:

 - it's still not clear why you consider `ENOENT` to be different from the other error cases.  Do you want to treat that as a hard error, or not?  Should it effect the outcome of the function, or is it really just about logging?
 - in general, I feel like you'd benefit from not passing `GError` around at all.  Just log it from the read_file helper and `return FALSE` or something...

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,81 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+        {
+          // labels aren't used on ARM
+          label_content = g_strdup_printf ("Core %d", i);
+        }
+
+      g_strchomp (label_content);
+
+      // ignore Tctl on AMD devices
+      if (!g_str_equal (label_content, "Tctl"))
+        {
```
allisonkarlitskaya, line 160: I'd reverse the logic here.

```
if (g_str_equal (...))
   continue;
```

allisonkarlitskaya (changes requested):
Thanks for all the changes.  This code is super easy to understand now.

One very small nag, but otherwise this looks great.

src/bridge/cockpitcpusamples.c
tomasmatus: Done

allisonkarlitskaya (dismissed):
I'm happy with the C stuff here.

Thanks for all the fixes!

pkg/metrics/metrics.jsx
```diff
@@ -592,7 +644,18 @@ class CurrentMetrics extends React.Component {
         return (
             <Gallery className="current-metrics" hasGutter>
                 <Card id="current-metrics-card-cpu">
-                    <CardTitle>{ _("CPU") }</CardTitle>
+                    <CardHeader className='align-baseline'>
+                        <CardTitle>{ _("CPU") }</CardTitle>
+                        { !isNaN(this.cpuTemperature) &&
+                        <Flex className="temperature-flex" spaceItems={{ default: 'spaceItemsXs' }}>
+                            <FlexItem className={this.cpuTemperatureColors.iconColor}>
```
martinpitt, line 651: Should we really change the text color here? Red is still okay, but yellow-on-white is getting into poor contrast territory. Coloring the icon seems enough to me? @garrett

pkg/metrics/metrics.jsx
```diff
@@ -592,7 +644,18 @@ class CurrentMetrics extends React.Component {
         return (
             <Gallery className="current-metrics" hasGutter>
                 <Card id="current-metrics-card-cpu">
-                    <CardTitle>{ _("CPU") }</CardTitle>
+                    <CardHeader className='align-baseline'>
+                        <CardTitle>{ _("CPU") }</CardTitle>
+                        { !isNaN(this.cpuTemperature) &&
+                        <Flex className="temperature-flex" spaceItems={{ default: 'spaceItemsXs' }}>
```
martinpitt, line 650: Using a flex here feels overkill - an `&nbsp` between icon and temperature feels enough? Moreover, we really don't want to break up the icon and the temp, as then you would have some "detached" warning icon floating around somewhere.

pkg/metrics/metrics.scss
```diff
@@ -66,6 +66,31 @@
           }
       }
 
+      .pf-c-helper-text {
+          font-size: var(--pf-global--FontSize--md);
+          font-weight: var(--pf-global--FontWeight--normal);
+      }
```
martinpitt, line 72: Is this used anywhere indirectly? Not in this patch. Might be a leftover?

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,81 @@ cockpit_cpu_samples (CockpitSamples *samples)
   g_strfreev (lines);
   g_free (contents);
 }
+
+gchar*
+read_hwmon_file (const gchar *path)
```
martinpitt, line 117: This function has nothing to do with "hwmon", this could just be `read_file()`.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,81 @@ cockpit_cpu_samples (CockpitSamples *samples)
   g_strfreev (lines);
   g_free (contents);
 }
+
+gchar*
```
martinpitt, line 116: All new helper functions (all new ones except `cockpit_cpu_temperature`)in this file should be `static`, these are not API but just internal implementation details.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,81 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+          label_content = g_strdup_printf ("Core %d", i);
+        }
+
+      g_strchomp (label_content);
+
+      // ignore Tctl on AMD devices
+      if (g_str_equal (label_content, "Tctl"))
+          continue;
+
+      gint temperature = g_ascii_strtoll (temp_content, NULL, 10);
```
martinpitt, line 162: This can fail on invalid strings. See the [docs](https://docs.gtk.org/glib/func.ascii_strtoll.html). Not crucial, but passing "0 degress" forever when the file is e.g. empty or contains some error message instead isn't nice.

Please add an `endptr`, validate it, and check `errno` on a zero return value. You can also keep it simple and just generally ignore "0" values -- it's relatively safe to say that zero Celsius is rather unrealistic, unless we are talking space telescopes or mars rovers :grin:

src/bridge/cockpitcpusamples.h
```diff
@@ -26,6 +26,11 @@ G_BEGIN_DECLS
 
 void            cockpit_cpu_samples         (CockpitSamples *samples);
 
+gchar*          read_hwmon_file             (const gchar *path);
+
+void            sample_cpu_sensors          (CockpitSamples *samples, int hwmonID);
```
martinpitt, line 31: Please drop these -- this isn't API for other parts of cockpit. Make the functions static instead (see above).

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,81 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+{
+  // iterate through all hwmon folders to find CPU sensors
+  for (int i = 0; TRUE; i++)
+    {
+      g_autofree gchar *path = g_strdup_printf ("/sys/class/hwmon/hwmon%d/name", i);
+      g_autofree gchar *name = read_hwmon_file (path);
+
+      // end loop on error
+      if (name == NULL)
+          break;
```
martinpitt, line 179: This is done a *lot*. Could this remember the highest valid `i` in a `static` variable and then  stop trying to read non-existing files over and over again?

This is fine for a follow-up though, it's not that important.

src/bridge/cockpitcpusamples.c
```diff
@@ -112,3 +112,81 @@ cockpit_cpu_samples (CockpitSamples *samples)
...
+      g_strchomp (name);
+
+      // compare device name with CPU names
+      // Intel: coretemp, AMD: k8temp or k10temp, ARM: cpu_thermal
+      if (g_str_equal (name, "coretemp") || g_str_equal (name, "cpu_thermal") ||
+          g_str_equal (name, "k8temp") || g_str_equal (name, "k10temp"))
+        {
+          // hwmon contains CPU info
+          sample_cpu_sensors (samples, i);
+        }
```
martinpitt, line 190: Likewise that. I think it would be better to do the "which hwmon%i do we want" detection *once*, store the paths in an array, and then only ever read these in each iteration.

src/bridge/test-metrics.c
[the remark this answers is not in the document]
martinpitt: These should ideally be `const` then. Sometimes this isn't possible as APIs are not appropriately `const`ified, but whenever you can, please do.

test/verify/check-metrics
```diff
@@ -834,6 +834,70 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
         with b.wait_timeout(300):
             b.wait(lambda: float(b.text("#load-avg .pf-l-flex div:first-child").split()[-1].rstrip(',')) < 10)
 
+        # Files with CPU temperature do not exist, nothing is displayed
+        b.wait_not_present("#current-metrics-card-cpu .temperature-flex")
+
+        # No matching type
+        m.execute("mkdir -p /tmp/hwmon/hwmon0")
+        m.execute("echo BAT0 > /tmp/hwmon/hwmon0/name")
+        m.execute("mount -o bind /tmp /sys/class")
```
martinpitt, line 843: This is a nondestructive test, so any failure would leave this mounted and thus break the next test. Do this:

    self.addCleanup(m.execute, "umount /sys/class")

which will then run this cleanup on failures as well. Same for creating /tmp/hwmon/.

test/verify/check-metrics
```diff
@@ -834,6 +834,70 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
+
+        b.wait_not_present("#current-metrics-card-cpu .temperature-flex")
+
+        # create files that contain CPU temperature
+        m.execute("mkdir -p /tmp/hwmon/hwmon1")
+
+        # Testing on ARM devices
+        m.execute("echo cpu_thermal > /tmp/hwmon/hwmon1/name")
+        m.execute("echo 60000 > /tmp/hwmon/hwmon1/temp1_input")
```
martinpitt, line 852: Note that there is m.write_file(), which looks a little nicer.

test/verify/check-metrics
```diff
@@ -834,6 +834,70 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
+
+        b.wait_visible("#current-metrics-card-cpu .text-color-critical")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-critical", "110 °C")
+
+        m.execute("echo 120000 > /tmp/hwmon/hwmon1/temp2_input")
+        b.wait_visible("#current-metrics-card-cpu .text-color-critical")
+        b.wait_in_text("#current-metrics-card-cpu .text-color-critical", "120 °C")
+
+        m.execute("umount /sys/class")
+        m.execute("rm -rf /tmp/hwmon")
```
martinpitt, line 899: See above, these need to be addCleanup() handlers to also run on failures.

martinpitt (changes requested):
Thank you, really nice work! Still a few things to fix, and one or two items should get @garrett 's feedback. As I said, the optimizations (remember valid file paths) are fine for a follow-up, but some bits (like the test cleanup) need to be fixed before landing.

garrett:
I'm on the metrics page with the branch compiled, yet I don't see CPU temperature. How do I make sure this works? Do I need something specific installed?

garrett:
> Did you try this on your own machine or on a VM?

My own machine.

I see the temperature in other places, like in [bottom](https://github.com/clementtsang/bottom) (rust-based top-like utility which has graphs, processes, temperatures, etc., much like our metrics page but in a terminal). So my work laptop does have temperatures shown in places... just not in the metrics page for whatever reason.

marusak:
@garrett since this changes bridge, just make and symlink won't work. You need to run new bridge as well. Easiest way is to prepare VM.

garrett:
> Easiest way is to prepare VM.

Is it supposed to show up in a VM? (It isn't here.)

garrett (approved):
LGTM! :+1: 

---

Solution to test locally: Start a VM, ssh to it, and run:

```
mount -o bind /tmp /sys/class 
mkdir -p /tmp/hwmon/hwmon1                                        
mkdir -p /tmp/hwmon/hwmon0 
echo BAT0 > /tmp/hwmon/hwmon0/name   
echo 60000 > /tmp/hwmon/hwmon1/temp1_input  
```

image


Then try higher temperatures:

```
echo 90000 > /tmp/hwmon/hwmon1/temp1_input  
```

image

```
echo 100000 > /tmp/hwmon/hwmon1/temp1_input
```

image

Of course, you can do silly stuff too :wink::

image

tomasmatus:
@martinpitt sounds good! I can try to add a static variable which would remember all hwmon%i folders that contain CPU info so the code to find it doesn't run every single time.

@garrett to test this on your local machine you need to stop `cockpit.service`, then in your cockpit directory run 
```# mount -o bind ./cockpit-pcp /usr/libexec/cockpit-pcp```
and start cockpit again.

**Comment by martinpitt** [comment]
@garrett: No, this isn't in the cockpit-pcp bridge, but the main one. You'll need

    mount -o bind ./cockpit-bridge /usr/bin/cockpit-bridge

**Comment by martinpitt** [comment]
FWIW, I did test it on my real system a few weeks ago, and it looked alright. I'll do a final review tomorrow. Most of my earlier review comments are still outstanding though, and not "outdated" -- but I'll close them during review as appropriate.

**Comment by martinpitt** [comment]
Argh, ignore the previous comment, this was for another PR.

File: src/bridge/test-metrics.c (hunk 19 above)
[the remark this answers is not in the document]
Comment (martinpitt) [reply src/bridge/test-metrics.c:883]: json-glib is not constified at all, so nevermind.

File: src/bridge/cockpitcpusamples.c (hunk 31 above)
Comment (martinpitt) [reply src/bridge/cockpitcpusamples.c:162]: I added some error handling.

File: src/bridge/cockpitcpusamples.c (hunk 29 above)
Comment (martinpitt) [reply src/bridge/cockpitcpusamples.c:117]: Renamed and made static.

File: src/bridge/cockpitcpusamples.c (hunk 30 above)
Comment (martinpitt) [reply src/bridge/cockpitcpusamples.c:116]: Done.

File: pkg/metrics/metrics.scss (hunk 28 above)
Comment (martinpitt) [reply pkg/metrics/metrics.scss:72]: Dropped.

File: pkg/metrics/metrics.jsx (hunk 27 above)
Comment (martinpitt) [reply pkg/metrics/metrics.jsx:650]: Changed to `&nbsp;` which is more correct IMHO. Renamed "temperature-flex" to just "temperature", we only use that class in the test and it should convey the meaning, not the implementation.

File: src/bridge/cockpitcpusamples.h (hunk 32 above)
Comment (martinpitt) [reply src/bridge/cockpitcpusamples.h:31]: Done.

**Review by martinpitt — APPROVED** [review]
Thanks! I made a few small fixes as mentioned above (resolving threads). This leaves the "excessive polling/reading files" issues, but as I said I'm happy for these to be a follow-up. Are you okay with that?

I also rebased as your branch was 64 commits behind and this is breaking the packit tests.

**Review by jelly — APPROVED** [review]

**Comment by jelly** [comment]
> @jelly : You still have a nack on this, want to re-review? (I think your comments got addressed)

LGTM!

**Comment by garrett** [comment]
Edited screenshot; merged several from this thread to make a hybrid example that shows off not just the CPU temperature, but one in a warning state.

**Comment by dkragen** [comment]
My apologies if this is not the right way to do this, but there seems to be an error in Cockpit's interface with lm-sensors:

image

image

**Comment by dkragen** [comment]
Just looking at the above it would appear that cockpit is reporting the critical temperature rather than the actual temperature.

**Comment by martinpitt** [comment]
@dkragen : This was a bug in cockpit 285. Please update to a newer (or older, if you must) version, it got fixed in 286.

**Comment by fekberg** [comment]
I'm on version 287-1 (Ubutu) and can't see any temperature data at all in the web interface.

`sensors` report temperatures correctly:

```
coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +36.0°C  (high = +100.0°C, crit = +100.0°C)
Core 0:        +35.0°C  (high = +100.0°C, crit = +100.0°C)
Core 1:        +35.0°C  (high = +100.0°C, crit = +100.0°C)
Core 2:        +34.0°C  (high = +100.0°C, crit = +100.0°C)
Core 3:        +34.0°C  (high = +100.0°C, crit = +100.0°C)
```

Am I missing something that I need to install?

martinpitt:
@fekberg Please file a new issue with the output of `grep -r . /sys/class/hwmon/hwmon*/name` and `grep -r . /sys/class/hwmon/hwmon*/temp*`. Thanks!

BAProductions:
Allow user to switch from C to F as well

garrett:
> Allow user to switch from C to F as well

Computers are _always_ measured in °C, however, even in the US. (I'll explain why a little further on in this comment.)

It's also a great scale, too: as you generally want to be close to 40°C ideally, and almost always < 80°C (but especially < 95°C) and _never_ want to be >= 100°C.

https://blog.pcbuilder.net/ideal-cpu-temperature/

Temperature depends on several things, however:

- Specific specifications of the component. Some hardware can withstand slightly higher temperatures than others.
- Cooling system, both active and passive. This includes fans, liquid coolers, and airflow in general.
- Load of the machine. Something with higher CPU and/or GPU usage will have higher temperatures than an idle machine.
- Ambient temperature. Examples: Summer days can make most computers run hotter. Server rooms often have air conditioning to deal with the excess heat.

But, generally, due to physics, the warning at levels > 80 to > 95 and emergency levels at > 95 are generally quite correct for almost all hardware, with levels > 100 causing components to get fried if they don't shutdown first.

The point of measuring temperatures within a computer is to determine if something needs to be optimized or fixed, not to compare to weather in your region. Perhaps the load is too high; this can be an indicator. Perhaps the environment is too warm. Perhaps there's been a failure in the cooling system (or it was the wrong one chosen for the device). And so on. Having a scale from 0 to 100 works really well for this.

Now, lets compare those reference temperatures:

°C | °F
-- | --
40 | 104
75 | 167
80 | 176
95 | 203
100 | 212

Which ones can you easily remember and understand? The Fahrenheit numbers are [arbitrary numbers on an arbitrary measuring system](https://www.britannica.com/science/Fahrenheit-temperature-scale) (some unknown mixture of water, ice, and ammonium chloride vs. an incorrect estimate of body temperature), whereas the Celsius ones are standard, easy to remember, and based on the properties of water (freezing and boiling at sea level), and the scale actually applies well enough for silicon too.

_This_ is exactly why everyone uses Celsius worldwide for computers (in addition to most of the world for everything else too). And this is why there shouldn't be a setting for changing computer heat values to Fahrenheit.

PR #17161, opened 2022-03-21
pf4: Properly handled nested PF Flex

garrett:
Fix for https://github.com/cockpit-project/cockpit-podman/pull/943, where there's a PF Flex nested immediately inside another PF Flex.


marusak (approved):
Nice, thank you!

PR #17163, opened 2022-03-22
base1: Default to base-10 units in cockpit.format_bytes*()

martinpitt:
This makes more sense as a default now. Storage and network devices,
network transmission speeds etc. have all been base-10 for decades. The
only place which is inherently/physically base-2 is total RAM size, and
reserved kdump memory because that's just its API. But these special
cases have all been explicitly moved to format as base-2 recently.

https://bugzilla.redhat.com/show_bug.cgi?id=1970119

---

[pixel diff review](https://github.com/cockpit-project/pixel-test-reference/compare/ac6d9ad037ca8939d6795834abf82b35c4039836..fa3e8b66637ce0815d01915779b5906b98245f4d)

This is most apparent on the Metrics, Storage, and Networking pages:
image
image
image

## Show disk/memory/network sizes and rates in decimal units

All sizes and rates are now shown in decimal units, such as "MB" or "kB/s". Physical storage, network devices,
and network transmission speeds have all been sold and used in decimal units for a long time. This also avoids some inconsistencies, such as the Metrics page showing available/used RAM in decimal units, but the memory usage of individual services in binary units.

Only total RAM size on the Overview, the size of individual RAM modules on the Hardware Details page, and the size of KDump memory are shown in binary units such as MiB or GiB, as these are all natively based on these.

image

### DIFF
--- a/test/verify/check-storage-stratis
+++ b/test/verify/check-storage-stratis
@@ -33,20 +33,25 @@ class TestStorageStratis(StorageCase):
 
         self.login_and_go("/storage")
 
+        if m.image == "rhel-8-6-distropkg":
+            SIZE_4GB = 4 * 1024 * 1024 * 1024
+        else:
+            SIZE_4GB = 4 * 1000000000
+
         dev_1 = "/dev/sda"
-        m.add_disk("4G", serial="DISK1")
+        m.add_disk(SIZE_4GB, serial="DISK1")
         b.wait_in_text("#drives", dev_1)
 
         dev_2 = "/dev/sdb"
-        m.add_disk("4G", serial="DISK2")
+        m.add_disk(SIZE_4GB, serial="DISK2")
         b.wait_in_text("#drives", dev_2)
 
         dev_3 = "/dev/sdc"
-        m.add_disk("4G", serial="DISK3")
+        m.add_disk(SIZE_4GB, serial="DISK3")
         b.wait_in_text("#drives", dev_3)
 
         dev_4 = "/dev/sdd"
-        m.add_disk("4G", serial="DISK4")
+        m.add_disk(SIZE_4GB, serial="DISK4")
         b.wait_in_text("#drives", dev_4)
 
         # Create a pool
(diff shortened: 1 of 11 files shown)


martinpitt:
OK, this works in general. There are a [few expected pixel failures](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17163-20220322-100044-02321839-fedora-35/log.html) which I'll update after this is otherwise ready. @marusak , @mvollmer , could you please do an initial review here? The test changes do show some impedance mismatch between the CLI tools like `lvcreate` (which *only* support base-2 units) and our new UI consistency aim for base-10 everywhere. I don't think it's too bad, but I'd welcome some more eyes on this. Thanks!

test/verify/check-storage-stratis
```diff
@@ -33,20 +33,22 @@ class TestStorageStratis(StorageCase):
 
         self.login_and_go("/storage")
 
+        SIZE_4GB = 4 * 1000000000
+
         dev_1 = "/dev/sda"
-        m.add_disk("4G", serial="DISK1")
```
marusak, line 37: 4G meant 4GiB?

martinpitt: Yes, the QEMU API works in base-2 units

mvollmer:
"Usage" on the overview still shows "GiB".

image

mvollmer:
I wonder if we should change the order here to "GB", "GiB", "bytes", so that the familiar unit comes first.

image

mvollmer (changes requested):
Overview "Usage" should use base10 as well.

martinpitt:
> "Usage" on the overview still shows "GiB".

That's the one I'm unsure about -- we discussed in the meeting that total memory size is "special". I don't have a machine where it is actually a round number (my laptop uses some of its 16 GiB for video memory, resulting in 15.4 GiB, and my other stuff is just VMs). I wouldn't mind too much with switching that over to base-10 as well, but there's a tension here -- being consistent wiht other pages vs. reflecting what your computer says on the tin.

> I wonder if we should change the order here to "GB", "GiB", "bytes", so that the familiar unit comes first.

Good idea, I'll do that.

mvollmer:
> > "Usage" on the overview still shows "GiB".
> 
> That's the one I'm unsure about -- we discussed in the meeting that total memory size is "special".

Ahh.  If it's not a bug then I am fine with it.

martinpitt:
I added another commit for swapping the binary/decimal units on the storage page, as above. Rebased, and re-running to see if that broke any test, and get a current pixel diff.

martinpitt:
Bah, right, [distropkg](https://logs.cockpit-project.org/logs/pull-17163-20220325-165806-84561a78-rhel-8-6-distropkg/log.html) -- I'll have to add back these special cases. These are not nearly as important as they used to be, but still a thing for RHEL 8.

martinpitt:
@mvollmer : This is ready for review now.

mvollmer (approved):

PR #17165, opened 2022-03-22
metrics: fix missing aria-label on progress element

jelly:
Without specifying an aria-label the console logs the following message
on every metrics update:
"One of aria-label or aria-labelledby properties should be passed when
using the progress component without a title"


martinpitt (approved):
Argh, that was my fault I think. Thanks! I triggered the required test, and two fedora ones for good measure (especially the /devel one is interesting here)

PR #17167, opened 2022-03-22
HACKING: Document ./configure paths for cockpit-bridge

jelly:
Testing cockpit-bridge changes here locally did not work due to the
paths defaulting to /usr/local.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -233,6 +233,14 @@ You need to disable SELinux with
 for this to work, as your local build tree does not otherwise have the expected
 SELinux type.
 
+Some cockpit binaries rely on specific paths in /usr/share or libexecdir to be
+set correctly, by default they are set to `/usr/local`.
+
+On RPM based systems this can be set using an autogen.sh argument, and
+afterwards you need to rebuild:
+
+    $ ./autogen.sh rpm
+
 ## Installation from upstream sources
 
     $ make


jelly:
Not 100% convinced about my own changes, do you have a better idea @martinpitt ?

HACKING.md
```diff
@@ -233,6 +233,13 @@ You need to disable SELinux with
 for this to work, as your local build tree does not otherwise have the expected
 SELinux type.
 
+Note that `./configure` defaults might not represent the running system's,
+cockpit-bridge might look in the wrong directory for libexecdir executables.
```
martinpitt, line 237: cockpit-bridge shouldn't call any libexecdir, but I suppose it will not look into /usr/share/cockpit/ for finding the package-installed cockpit manifests. Check with `cockpit-bridge --packages`. It should run, though? Or is it failing to link, etc.?

This is true for cockpit-ws, though -- that will look for /usr/libexec/cockpit-session at least. So perhaps

> `./configure` defaults don't work with files installed from packages, so `cockpit-bridge` and `cockpit-ws` will look in the wrong directories for libexecdir executables or manifests.

HACKING.md
```diff
@@ -233,6 +233,13 @@ You need to disable SELinux with
...
+
+On RPM based systems this can set correctly with:
```
martinpitt, line 239: can "be" set, and afterwards you need to rebuild.

martinpitt (changes requested):
Thanks!

HACKING.md
jelly: Ah, I went with:

```
[jelle@t14s][~/projects/cockpit]%strings /usr/bin/cockpit-bridge| grep libexec
libexecdir
/usr/lib/cockpit
```

So it at least records it, not sure if it does anything with it. Maybe we should just generalise the text?

"Some cockpit binaries rely on specific paths in usr/share or libexecdir requiring the paths to be resolved correctly."

martinpitt (approved):
Thanks! Still triggering required test.

PR #17170, opened 2022-03-22
systemd: add FIPS as crypto policy

jelly:
FIPS is a restricted set of cryptographic algorithms which requires a
special kernel boot parameter to be enabled. To enable FIPS mode on
RHEL/Fedora fips-mode-setup --enable has to be executed and the machine
succesfully rebooted.

### DIFF
--- a/pkg/systemd/overview-cards/cryptoPolicies.jsx
+++ b/pkg/systemd/overview-cards/cryptoPolicies.jsx
@@ -19,11 +19,10 @@
 
 import cockpit from "cockpit";
 import React, { useState, useEffect } from 'react';
-import { Button, Flex, Modal, Popover } from "@patternfly/react-core";
-import { ExternalLinkSquareAltIcon, HelpIcon, InProgressIcon } from '@patternfly/react-icons';
+import { Button, Flex, FlexItem, Modal, Popover } from "@patternfly/react-core";
+import { ExclamationTriangleIcon, ExternalLinkSquareAltIcon, HelpIcon } from '@patternfly/react-icons';
 
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
-import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
 import { PrivilegedButton } from "cockpit-components-privileged.jsx";
 import { ProfilesMenuDialogBody } from "./profiles-menu-dialog-body.jsx";
 
@@ -36,28 +35,33 @@ const cryptopolicies = {
     DEFAULT: _("Recommended, secure settings for current threat models."),
     FUTURE: _("Protects from anticipated near-term future attacks at the expense of interoperability."),
     LEGACY: _("Higher interoperability at the cost of an increased attack surface."),
+    FIPS: (<Flex alignItems={{ default: 'alignItemsCenter' }}>
+        {_("Only use approved and allowed algorithms when booting in FIPS mode.")}
+        <Button component='a'
+                rel="noopener noreferrer" target="_blank"
+                variant='link'
+                isInline
+                icon={<ExternalLinkSquareAltIcon />} iconPosition="right"
+                href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/assembly_installing-a-rhel-8-system-with-fips-mode-enabled_security-hardening">
+            {_("Learn more")}
+        </Button>
+    </Flex>),
 };
-const applyNeedRebootStamp = "/run/cockpit/crypto-policies-reboot-stamp";
-const displayProfileText = profile => profile.charAt(0) + profile.slice(1, profile.length).toLowerCase();
+
+const displayProfileText = profile => profile === "FIPS" ? profile : profile.charAt(0) + profile.slice(1, profile.length).toLowerCase();
+const isInconsistentPolicy = (policy, fipsEnabled) => policy === "FIPS" !== fipsEnabled;
 
 export const CryptoPolicyRow = () => {
     const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState(null);
     const [isOpen, setIsOpen] = useState(false);
+    const [fipsEnabled, setFipsEnabled] = useState(null);
 
     useEffect(() => {
-        // Avoid cockpit.spawn every re-render from parent
-        if (currentCryptoPolicy === null) {
-            cockpit.spawn(["update-crypto-policies", "--show"], { err: "message" })
-                    .then(output => setCurrentCryptoPolicy(output.trim()))
-                    .catch(err => {
-                        console.debug("no crypto policies support: ", err.toString());
-                        setCurrentCryptoPolicy(undefined);
-                    });
-            // When users run `update-crypto-policies` manually.
-            cockpit.file("/etc/crypto-policies/state/current")
-                    .watch(content => setCurrentCryptoPolicy(content ? content.trim() : null));
-        }
-    }, [currentCryptoPolicy]);
+        cockpit.file("/proc/sys/crypto/fips_enabled").read()
+                .then(content => setFipsEnabled(content ? content.trim() === "1" : false));
+        cockpit.file("/etc/crypto-policies/state/current")
+                .watch(content => setCurrentCryptoPolicy(content ? content.trim() : null));
+    }, []);
 
     if (!currentCryptoPolicy) {
         return null;
@@ -76,6 +80,7 @@ export const CryptoPolicyRow = () => {
                 <CryptoPolicyDialog close={() => setIsOpen(false)}
                                     currentCryptoPolicy={currentCryptoPolicy}
                                     setCurrentCryptoPolicy={setCurrentCryptoPolicy}
+                                    fipsEnabled={fipsEnabled}
                 />
                 }
             </td>
@@ -83,34 +88,38 @@ export const CryptoPolicyRow = () => {
     );
 };
 
+const setPolicy = (policy, setError, setInProgress) => {
+    setInProgress(true);
+
+    let promise;
+    if (policy === "FIPS") {
+        promise = cockpit.spawn(["fips-mode-setup", "--enable"], { superuser: "require", err: "message" });
+    } else {
+        promise = cockpit.spawn(["fips-mode-setup", "--disable"], { superuser: "require", err: "message" }).then(() =>
+            cockpit.spawn(["update-crypto-policies", "--set", policy], { superuser: "require", err: "message" }));
+    }
+
+    promise.then(() => cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" }))
+            .catch(error => setError(error))
+            .finally(() => setInProgress(false));
+};
+
 const CryptoPolicyDialog = ({
     close,
     currentCryptoPolicy,
-    setCurrentCryptoPolicy,
+    fipsEnabled,
+    reApply,
 }) => {
     const [error, setError] = useState();
+    const [inProgress, setInProgress] = useState(false);
     const [selected, setSelected] = useState(currentCryptoPolicy);
-
-    const setPolicy = (reboot) => {
-        cockpit.spawn(["update-crypto-policies", "--set", selected], { superuser: "require", err: "message" })
-                .then(() => {
-                    setCurrentCryptoPolicy(selected);
-                    if (reboot) {
-                        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
-                                .catch(error => setError(error));
-                    } else {
-                        cockpit.file(applyNeedRebootStamp, { superuser: "require" }).replace("\n")
-                                .then(() => close());
-                    }
-                })
-                .catch(error => setError(error));
-    };
 
     const policies = Object.keys(cryptopolicies).map(policy => ({
         name: policy,
         title: displayProfileText(policy),
         description: cryptopolicies[policy],
-        active: policy === currentCryptoPolicy,
+        active: !isInconsistentPolicy(policy, fipsEnabled) && policy === currentCryptoPolicy,
+        inconsistent: isInconsistentPolicy(policy, fipsEnabled) && policy === currentCryptoPolicy,
         recommended: false,
     }));
 
@@ -120,7 +129,8 @@ const CryptoPolicyDialog = ({
             name: currentCryptoPolicy,
             title: displayProfileText(currentCryptoPolicy),
             description: _("Custom crypto policy"),
-            active: true,
+            active: !isInconsistentPolicy(currentCryptoPolicy, fipsEnabled),
+            inconsistent: isInconsistentPolicy(currentCryptoPolicy, fipsEnabled),
             recommended: false,
         });
     }
@@ -160,13 +170,16 @@ const CryptoPolicyDialog = ({
                footer={
                    <>
                        {error && <ModalError dialogError={typeof error == 'string' ? error : error.message} />}
-                       <Button id="crypto-policy-save-reboot" variant='primary' onClick={() => setPolicy(true)}>
-                           {_("Save and reboot")}
+                       {inProgress &&
+                       <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
+                           {_("Applying new policy... This may take a few minutes.")}
+                       </Flex>}
+                       <Button id="crypto-policy-save-reboot" variant='primary' onClick={() => setPolicy(selected, setError, setInProgress)}
+                               isDisabled={inProgress} isLoading={inProgress}
+                       >
+                           {reApply ? _("Reapply and reboot") : _("Apply and reboot")}
                        </Button>
-                       <Button id="crypto-policy-save-reboot-later" variant='secondary' onClick={() => setPolicy(false)}>
-                           {_("Save only")}
-                       </Button>
-                       <Button variant='link' onClick={close}>
+                       <Button variant='link' onClick={close} isDisabled={inProgress}>
                            {_("Cancel")}
                        </Button>
                    </>
@@ -174,37 +187,50 @@ const CryptoPolicyDialog = ({
         >
             {currentCryptoPolicy && <ProfilesMenuDialogBody active_profile={currentCryptoPolicy}
                                                      change_selected={setSelected}
+                                                     isDisabled={inProgress}
                                                      profiles={policies} />}
         </Modal>
     );
 };
 
 export const CryptoPolicyStatus = () => {
-    const [requiresReboot, setRequiresReboot] = useState(false);
-    const [showShutdownModal, setShowShutDownModal] = useState(false);
+    const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState(null);
+    const [fipsEnabled, setFipsEnabled] = useState(null);
+    const [showReApplyCryptoPolicy, setShowReApplyCryptoPolicy] = useState(false);
 
     useEffect(() => {
-        if (!requiresReboot) {
-            // For when we change the setting
-            cockpit.file(applyNeedRebootStamp).watch(content => setRequiresReboot(content !== null));
+        if (currentCryptoPolicy === null) {
+            cockpit.file("/etc/crypto-policies/state/current")
+                    .watch(content => setCurrentCryptoPolicy(content ? content.trim() : undefined));
         }
-    }, [requiresReboot]);
 
-    if (!requiresReboot) {
-        return null;
+        cockpit.file("/proc/sys/crypto/fips_enabled").read()
+                .then(content => setFipsEnabled(content ? content.trim() === "1" : false));
+    }, [currentCryptoPolicy]);
+
+    if (isInconsistentPolicy(currentCryptoPolicy, fipsEnabled)) {
+        return (
+            <li className="system-health-crypto-policies">
+                <Flex spacer={{ default: 'spaceItemsSm' }} flexWrap={{ default: 'nowrap' }}>
+                    <FlexItem><ExclamationTriangleIcon size="sm" className="crypto-policies-health-card-icon" /></FlexItem>
+                    <div>
+                        <div id="inconsistent_crypto_policy">
+                            {currentCryptoPolicy === "FIPS" ? _("FIPS is not properly enabled") : _("Crypto policy is inconsistent")}
+                        </div>
+                        <Button isInline variant="link" className="pf-u-font-size-sm" onClick={() => setShowReApplyCryptoPolicy(true)}>
+                            {_("Review crypto policy")}
+                        </Button>
+                    </div>
+                </Flex>
+                {showReApplyCryptoPolicy &&
+                <CryptoPolicyDialog currentCryptoPolicy={currentCryptoPolicy}
+                                    close={() => setShowReApplyCryptoPolicy(false)}
+                                    fipsEnabled={fipsEnabled}
+                                    reApply />
+                }
+            </li>
+        );
     }
 
-    return (
-        <li className="system-health-crypto-policies">
-            <Flex flexWrap={{ default: 'nowrap' }} spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
-                <InProgressIcon size="sm" className="crypto-policies-health-card-icon" />
-                <Button isInline variant="link" onClick={() => setShowShutDownModal(true)}>
-                    {_("Reboot to apply new crypto policy")}
-                </Button>
-            </Flex>
-            {showShutdownModal &&
-            <ShutdownModal onClose={() => setShowShutDownModal(false)} />
-            }
-        </li>
-    );
+    return null;
 };
--- a/pkg/systemd/overview-cards/profiles-menu-dialog-body.jsx
+++ b/pkg/systemd/overview-cards/profiles-menu-dialog-body.jsx
@@ -48,16 +48,17 @@ export class ProfilesMenuDialogBody extends React.Component {
 
     render() {
         const profiles = this.props.profiles.map((itm) => {
-            const active = this.props.active_profile == itm.name;
-
             return (
-                <MenuItem itemId={itm.name} key={itm.name} data-value={itm.name} description={itm.description}>
+                <MenuItem itemId={itm.name} key={itm.name} data-value={itm.name}
+                          description={itm.description} isDisabled={this.props.isDisabled}
+                          isActive={itm.active}>
                     <Flex alignItems={{ default: 'alignItemsCenter' }}>
                         <p>{ itm.title }</p>
                         <FlexItem>
                             <LabelGroup>
                                 {itm.recommended && <Label color="green" variant='filled'>{_("recommended")}</Label>}
-                                {active && <Label color="blue" variant='filled'>{_("active")}</Label>}
+                                {itm.active && <Label color="blue" variant='filled'>{_("active")}</Label>}
+                                {itm.inconsistent && <Label color="orange" variant='filled'>{_("inconsistent")}</Label>}
                             </LabelGroup>
                         </FlexItem>
                     </Flex>
@@ -70,7 +71,6 @@ export class ProfilesMenuDialogBody extends React.Component {
                       this.setState({ selected_profile: selected });
                       this.props.change_selected(selected);
                   }}
-                  activeItemId={this.state.selected_profile}
                   selected={this.state.selected_profile}>
                 <MenuContent>
                     <MenuList>
@@ -85,4 +85,5 @@ ProfilesMenuDialogBody.propTypes = {
     active_profile: PropTypes.string.isRequired,
     change_selected: PropTypes.func.isRequired,
     profiles: PropTypes.array.isRequired,
+    isDisabled: PropTypes.bool,
 };
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -871,26 +871,30 @@ password=foobar
         dbus_call = 'busctl get-property org.freedesktop.login1 /org/freedesktop/login1  org.freedesktop.login1.Manager ScheduledShutdown'
         self.assertEqual('(st) "" 0', m.execute(dbus_call).strip())
 
-    @skipImage("crypto-policies not avaiable", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-2204", "ubuntu-stable", "arch")
+    @skipImage("crypto-policies not available", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-2204", "ubuntu-stable", "arch")
     def testCryptoPolicies(self):
-        def shown_profile_text(profile):
-            return profile[0] + profile[1:].lower()
+        m = self.machine
+        b = self.browser
+        self.allow_restart_journal_messages()
 
-        def change_profile(profile, new_profile, reboot=False):
+        def shown_profile_text(profile):
+            return profile if profile == "FIPS" else profile[0] + profile[1:].lower()
+
+        def change_profile(profile, new_profile):
             b.click("#crypto-policy-button")
             b.wait_in_text(".pf-c-menu__item.pf-m-selected", shown_profile_text(profile))
             profile_button_name = shown_profile_text(new_profile)
             b.click(f".pf-c-menu__item-main .pf-c-menu__item-text:contains('{profile_button_name}')")
-            if reboot:
-                b.click("#crypto-policy-save-reboot")
+            b.click("#crypto-policy-save-reboot")
+            if new_profile == "FIPS":
+                # Initramfs re-generation takes a while
+                m.wait_reboot(timeout_sec=600)
             else:
-                b.click("#crypto-policy-save-reboot-later")
-                b.wait_text("#crypto-policy-button", shown_profile_text(new_profile))
-                profile = m.execute(cmd + " --show").strip()
-                self.assertEqual(new_profile, profile)
+                m.wait_reboot()
+            m.start_cockpit()
+            self.login_and_go("/system")
+            b.wait_text("#crypto-policy-button", shown_profile_text(new_profile))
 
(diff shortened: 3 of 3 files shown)


garrett:
Mockups:

crypto-policy-fips-modal

Here's what happens when the machine is out of sync, in case someone switched crypto policies outside of Cockpit and didn't enable the attribute for the initrd and reboot:

crypto-policy-fips-out_of_sync

martinpitt:
Still several test failures

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -89,18 +101,34 @@ const CryptoPolicyDialog = ({
...
+        };
         cockpit.spawn(["update-crypto-policies", "--set", selected], { superuser: "require", err: "message" })
                 .then(() => {
                     setCurrentCryptoPolicy(selected);
-                    if (reboot) {
-                        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
-                                .catch(error => setError(error));
+                    if (selected === "FIPS") {
+                        setRebootRequired(reboot);
+                        setInProgress(true);
```
martinpitt, line 123: Let's make this a bit more robust -- move the setInProgress(true) in front of the `update-crypto-policies` call, just in case this takes a non-trivial amount of time (it might change implementation in the future).

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -89,18 +101,34 @@ const CryptoPolicyDialog = ({
...
+                        cockpit.spawn(["fips-mode-setup", "--enable"], { superuser: "require", err: "message" })
+                                .then(() => {
+                                    setInProgress(false);
+                                    applyPolicy(reboot);
+                                })
+                                .catch(error => { setInProgress(false); setError(error) });
```
martinpitt, line 129: DRY -- Use `.finally()` for setInProgress(false), this then also simplifies the then/catch handlers to be functional instead of a code block.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +188,132 @@ const CryptoPolicyDialog = ({
                footer={
                    <>
                        {error && <ModalError dialogError={typeof error == 'string' ? error : error.message} />}
-                       <Button id="crypto-policy-save-reboot" variant='primary' onClick={() => setPolicy(true)}>
+                       {inProgress &&
+                           <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }} className="fips-mode-progress">
+                               <Spinner isSVG size="md" />
```
martinpitt, line 193: Hmm, I thought an `isLoading` button already showed a spinner.. we don't want two of them.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -89,18 +101,34 @@ const CryptoPolicyDialog = ({
...
         cockpit.spawn(["update-crypto-policies", "--set", selected], { superuser: "require", err: "message" })
                 .then(() => {
                     setCurrentCryptoPolicy(selected);
-                    if (reboot) {
-                        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
-                                .catch(error => setError(error));
+                    if (selected === "FIPS") {
+                        setRebootRequired(reboot);
+                        setInProgress(true);
+                        cockpit.spawn(["fips-mode-setup", "--enable"], { superuser: "require", err: "message" })
```
martinpitt, line 124: To make double sure: just calling `fips-mode-setup --enable` will *not* automatically switch the crypto policy to "FIPS"? This still has to happen explicitly? The [documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening#switching-the-system-to-fips-mode_using-the-system-wide-cryptographic-policies) specifically documents "FIPS" as

> A policy level that conforms with the FIPS 140-2 requirements. This is used internally by the fips-mode-setup tool, which switches the RHEL system into FIPS mode. 

So ISTM we should call one or the other, but not both.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +188,132 @@ const CryptoPolicyDialog = ({
...
-    }, [requiresReboot]);
 
-    if (!requiresReboot) {
-        return null;
+        // Avoid cockpit.spawn every re-render from parent
+        if (currentCryptoPolicy === null) {
+            cockpit.spawn(["update-crypto-policies", "--show"], { err: "message" })
+                    .then(output => setCurrentCryptoPolicy(output.trim()))
+                    .catch(err => {
+                        console.debug("no crypto policies support: ", err.toString());
```
martinpitt, line 238: Let's please take this out. I am now seeing this message in each test log (e.g. open any test in [a recent log](https://logs.cockpit-project.org/logs/pull-17176-20220324-101200-bb8921de-debian-stable/log.html)). There must be another similar message around in the code, as this one is new, and this PR does not just move it -- this is in addition.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +188,132 @@ const CryptoPolicyDialog = ({
...
+        return (
+            <li className="system-health-crypto-policies">
+                <Flex spacer={{ default: 'spaceItemsSm' }} flexWrap={{ default: 'nowrap' }}>
+                    <FlexItem><ExclamationTriangleIcon size="sm" className="crypto-policies-health-card-icon" /></FlexItem>
+                    <div>
+                        <div id="fips_mode_disabled_text">
+                            {_("FIPS crypto policy selected, but not active")}
+                        </div>
+                        <Button isInline variant="link" className="pf-u-font-size-sm" onClick={() => setShowEnableFIPSModal(true)}>
+                            {_("Enable FIPS mode and reboot")}
```
martinpitt, line 277: This sounds a bit scary, like it would immediately do this. Can we take out the "and reboot" here? (Also see below)

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +188,132 @@ const CryptoPolicyDialog = ({
...
-            <ShutdownModal onClose={() => setShowShutDownModal(false)} />
-            }
-        </li>
+        <Modal isOpen position="top" variant="medium"
+                   onClose={onClose}
+                   id="enable-fips-mode-dialog"
+                   title={_("Enable FIPS mode")}
+                   footer={<>
+                       {error && <ModalError dialogError={typeof error == 'string' ? error : error.message} />}
+                       <Button variant='primary' isDisabled={inProgress} isLoading={inProgress} onClick={onSubmit}>{_("Enable FIPS mode and reboot")}</Button>
```
martinpitt, line 312: Could we do a similar deferral of rebooting as in the crypto policy dialog? The time when spotting this error is not necessarily a good time to reboot your machine.

test/verify/check-system-info
```diff
@@ -895,6 +895,36 @@ password=foobar
         b.wait_in_text(".pf-c-menu__item.pf-m-selected", "Custom crypto policy")
         b.click("#crypto-policy-dialog button.pf-c-button.pf-m-link")
 
+    @skipImage("crypto-policies not avaiable", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable", "arch")
```
martinpitt, line 898: Please add "ubuntu-2204" as well, to preempt a collision with #17182

test/verify/check-system-info
```diff
@@ -895,6 +895,36 @@ password=foobar
...
+
+        # Admin sets FIPS in terminal
+        m.execute(cmd + " --set FIPS")
+        self.login_and_go("/system")
+        b.wait_text("#fips_mode_disabled_text", "FIPS crypto policy selected, but not active")
+        b.click(".system-health-crypto-policies button.pf-c-button.pf-m-link")
+        b.click("#enable-fips-mode-dialog button.pf-c-button.pf-m-primary")
+
+        # fips-mode-setup takes a while to complete.
+        m.wait_reboot(timeout_sec=180 * 3)
```
martinpitt, line 912: heh -- just make it a round "600", or at least "60 * 9" (But really, 5 mins, aka. 300s ought to suffice? It should take about a minute)

test/verify/check-system-info
```diff
@@ -895,6 +895,36 @@ password=foobar
         b.wait_in_text(".pf-c-menu__item.pf-m-selected", "Custom crypto policy")
         b.click("#crypto-policy-dialog button.pf-c-button.pf-m-link")
 
+    @skipImage("crypto-policies not avaiable", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2004", "ubuntu-stable", "arch")
+    def testFIPSPolicy(self):
```
martinpitt, line 899: This specifically tests the "clean up broken state" case, so perhaps `testInconsistentFIPSPolicy`?

test/verify/check-system-info
```diff
@@ -895,6 +895,36 @@ password=foobar
...
+        b.click("#enable-fips-mode-dialog button.pf-c-button.pf-m-primary")
+
+        # fips-mode-setup takes a while to complete.
+        m.wait_reboot(timeout_sec=180 * 3)
+        m.start_cockpit()
+        self.login_and_go("/system")
+        b.wait_text("#crypto-policy-button", "FIPS")
+
+        # Setting via dialog
+        m.execute(cmd + " --set DEFAULT")
```
martinpitt, line 918: At that point, the kernel still uses FIPS mode, no? So this is again inconsistent -- we should check this both ways. This is also worth a bug report against `update-crypto-policies` perhaps -- trying to switch away from FIPS while the kernel still has it enabled should fail, and point you to `fips-mode-setup`.

test/verify/check-system-info
```diff
@@ -895,6 +895,36 @@ password=foobar
...
+        b.click("#crypto-policy-button")
+        b.click(".pf-c-menu__item-main .pf-c-menu__item-text:contains('FIPS')")
+        b.click("#crypto-policy-save-reboot")
+
+        m.wait_reboot()
+        m.start_cockpit()
+        self.login_and_go("/system")
+        b.wait_text("#crypto-policy-button", "FIPS")
```
martinpitt, line 926: Let's also check `/proc/sys/crypto/fips_enabled` here. The way the test is currently written, it's pretty much guaranteed to be 1, but after the above comment it's not such a given any more.

test/verify/check-system-info
```diff
@@ -895,6 +895,36 @@ password=foobar
...
+        self.login_and_go("/system")
+        b.wait_text("#fips_mode_disabled_text", "FIPS crypto policy selected, but not active")
+        b.click(".system-health-crypto-policies button.pf-c-button.pf-m-link")
+        b.click("#enable-fips-mode-dialog button.pf-c-button.pf-m-primary")
+
+        # fips-mode-setup takes a while to complete.
+        m.wait_reboot(timeout_sec=180 * 3)
+        m.start_cockpit()
+        self.login_and_go("/system")
+        b.wait_text("#crypto-policy-button", "FIPS")
```
martinpitt, line 915: Please also check /proc/sys/crypto/fips_enabled that it really fully worked. Also check that the health card does not show the "but not active" message.

martinpitt (changes requested):
Thanks! First review round. I didn't test interactively yet, will do that on the next round.

test/verify/check-system-info
jelly: Yes, but we don't show a message when you have enabled FIPS mode and you don't use fips policy. Maybe that was an misunderstanding from my part?

test/verify/check-system-info
jelly: Or should it start with enabling FIPS the supported way, then putting the VM in an incosistent state and then testing that?

pkg/systemd/overview-cards/cryptoPolicies.jsx
martinpitt: We do, but the text does not make this clear. We also don't use the "..." convention to indicate that some action opens a dialog first.

pkg/systemd/overview-cards/cryptoPolicies.jsx
martinpitt: Right, but that's exactly the same as with changing the crypto policy. It's oddly inconsistent, and rather than forcing a reboot now, it will more likely lead to ignoring the issue. We wouldn't *have* to have two health card statuses -- both kinds of reboot (after update-crypto-policy and after fips-mode-setup) are the same kind of root cause, so there should be just one.

test/verify/check-system-info
martinpitt: That works as well, sure!

test/verify/check-system-info
martinpitt: I'm fine with postponing that direction to a follow-up (jira card). It may also make sense to ask the crypto-policies devs what should happen in such a case.

pkg/systemd/overview-cards/cryptoPolicies.jsx
jelly: I think the difference is that with FIPS mode you really have to mandatory reboot for it to work, while other policies can work somewhat without rebooting. It is indeed inconsistent but for crypto-policies we don't handle this case (when an admin changes a crypto policy without a reboot we can't detect it afaik.

garrett: So are you saying we should remove the inconsistent defer for the normal one too? 

Sure; I guess that would mean less code, less UI, and fewer things to go wrong for our users.

The inconsistent state would only be for when someone changes things outside of Cockpit then?

jelly: Conclusion from our meeting:

* Change the crypto policies dialog to always reboot
* Switching away from FIPS crypto policy disables FIPS mode

Inconsistent state can happen with FIPS mode:

* Admin changes to FIPS crypto policy with enabling fips mode
* Admin has fips mode enabled but uses a default policy
* Admin enables fips mode setup but didn't reboot

In this case we show a healthcard entry and let the user re-apply the current crypto policy and reboot so it's in a consistent state again.

jelly:
@garrett 
image

image

FIPS policy selected, but FIPS mode is not active.

garrett (dismissed):
I'm not sure about "selected"; I think it works without "selected", and they didn't select it to be inconsistent.

Re-apply is a little clunky, perhaps, but I guess it does feel better than "Apply FIPS policy again and reboot"?

MW says "reapply" doesn't need a hypen: https://www.merriam-webster.com/dictionary/reapply — but it does seem a little odd like reap-ply. :shrug: 

---

So:

1. Drop "selected"
2. Let's ask @allisonkarlitskaya about: re-apply, reapply, apply ... again

garrett:
@allisonkarlitskaya had a great suggestion:

> Crypto policy is inconsistent
> [Review crypto policy](#)

or, if it would only happen with FIPS (which I am pretty sure is the case), we could be more explicit:

> FIPS is not properly enabled
> [Review cryto policy](#)

And then we can just re-use the main switching modal.

And, since the modal mockup is old, here's a refreshed version:

crypto-policy

garrett:
I don't know if Save and reboot really works here. Should we, in this case, use "apply" and then "re-apply" if it's using FIPS? (We can't say "Saving FIPS policy..." we do say "Applying" right above, so it's not really saving what someone did, but applying a pre-used setting... so in this case, it'd look like this:

crypto-policy-FIPS

WDYT?

test/verify/check-system-info
```diff
@@ -931,6 +915,37 @@ password=foobar
         b.wait_in_text(".pf-c-menu__item.pf-m-selected", "Custom crypto policy")
         b.click("#crypto-policy-dialog button.pf-c-button.pf-m-link")
 
+    @skipImage("crypto-policies not avaiable", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2204", "ubuntu-2004", "ubuntu-stable", "arch")
+    def testInconsistentCryptoPolicy(self):
+        m = self.machine
+        b = self.browser
+        self.allow_journal_messages()
```
martinpitt, line 922: Heh, you want allow_*restart*_journal_messages() :grin:

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +182,69 @@ const CryptoPolicyDialog = ({
                footer={
                    <>
                        {error && <ModalError dialogError={typeof error == 'string' ? error : error.message} />}
-                       <Button id="crypto-policy-save-reboot" variant='primary' onClick={() => setPolicy(true)}>
-                           {_("Save and reboot")}
+                       {inProgress && selected === "FIPS" &&
+                           <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }} className="fips-mode-progress">
+                               <Spinner isSVG size="md" />
+                               <p>{_("Applying FIPS policy... This may take a few minutes.")}</p>
```
martinpitt, line 188: Can we show this for the other policies? In princile this might also take some time, no? Or would that create flickering?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +182,69 @@ const CryptoPolicyDialog = ({
...
-        if (!requiresReboot) {
-            // For when we change the setting
-            cockpit.file(applyNeedRebootStamp).watch(content => setRequiresReboot(content !== null));
+        if (currentCryptoPolicy === null) {
+            cockpit.spawn(["update-crypto-policies", "--show"], { err: "message" })
+                    .then(output => setCurrentCryptoPolicy(output.trim()))
+                    .catch(setCurrentCryptoPolicy(undefined));
+            // When users run `update-crypto-policies` manually.
+            cockpit.file("/etc/crypto-policies/state/current")
+                    .watch(content => setCurrentCryptoPolicy(content ? content.trim() : undefined));
```
martinpitt, line 221: These two race against each other. Chances are that `update-crypto-policies --show` will win, as it is slower. But ISTM that we should pick a Single Source of Truth.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -36,28 +35,36 @@ const cryptopolicies = {
...
-                    });
-            // When users run `update-crypto-policies` manually.
-            cockpit.file("/etc/crypto-policies/state/current")
-                    .watch(content => setCurrentCryptoPolicy(content ? content.trim() : null));
-        }
-    }, [currentCryptoPolicy]);
+        cockpit.spawn(["update-crypto-policies", "--show"], { err: "message" })
+                .then(output => setCurrentCryptoPolicy(output.trim()))
+                .catch(err => {
+                    console.debug("no crypto policies support: ", err.toString());
```
martinpitt, line 61: Still, let's please take this out, it's too noisy.

pkg/systemd/overview-cards/cryptoPolicies.jsx
martinpitt: It's still in the current version, but opening a new thread.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +182,69 @@ const CryptoPolicyDialog = ({
...
         }
-    }, [requiresReboot]);
 
-    if (!requiresReboot) {
-        return null;
+        cockpit.file("/proc/sys/crypto/fips_enabled").read()
+                .then(content => setFipsEnabled(content ? content.trim() === "1" : false));
+    }, [currentCryptoPolicy]);
+
+    if ((currentCryptoPolicy === "FIPS" && !fipsEnabled) || (fipsEnabled && currentCryptoPolicy !== "FIPS")) {
```
martinpitt, line 228: `(currentCryptoPolicy === "FIPS") !== fipsEnabled`?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -160,51 +182,69 @@ const CryptoPolicyDialog = ({
...
+        return (
+            <li className="system-health-crypto-policies">
+                <Flex spacer={{ default: 'spaceItemsSm' }} flexWrap={{ default: 'nowrap' }}>
+                    <FlexItem><ExclamationTriangleIcon size="sm" className="crypto-policies-health-card-icon" /></FlexItem>
+                    <div>
+                        <div id="inconsistent_crypto_policy">
+                            {currentCryptoPolicy === "FIPS" ? _("FIPS is not properly enabled") : _("Crypto policy is inconsistent")}
+                        </div>
+                        <Button isInline variant="link" className="pf-u-font-size-sm" onClick={() => setShowReApplyCryptoPolicy(true)}>
+                            {_("Review crypto policy")}
```
martinpitt, line 238: :+1: This is much less scary than the previous "fix and reboot", thanks!

pkg/systemd/overview-cards/profiles-menu-dialog-body.jsx
```diff
@@ -51,7 +51,7 @@ export class ProfilesMenuDialogBody extends React.Component {
             const active = this.props.active_profile == itm.name;
 
             return (
-                <MenuItem itemId={itm.name} key={itm.name} data-value={itm.name} description={itm.description}>
+                <MenuItem itemId={itm.name} key={itm.name} data-value={itm.name} description={itm.description} isDisabled={this.props.isDisabled || false}>
```
martinpitt, line 54: Why the `|| false`? Does `isDisabled={undefined}` result in the element to be disabled?

test/verify/check-system-info
```diff
@@ -873,54 +873,38 @@ password=foobar
...
             b.click("#crypto-policy-button")
             b.wait_in_text(".pf-c-menu__item.pf-m-selected", shown_profile_text(profile))
             profile_button_name = shown_profile_text(new_profile)
             b.click(f".pf-c-menu__item-main .pf-c-menu__item-text:contains('{profile_button_name}')")
-            if reboot:
-                b.click("#crypto-policy-save-reboot")
+            b.click("#crypto-policy-save-reboot")
+            if new_profile == "FIPS":
+                # Initramfs re-generation takes a while
+                m.wait_reboot(timeout_sec=600)
```
martinpitt, line 891: When disabling FIPS it certainly also has to regenerate the initrd?

test/verify/check-system-info
```diff
@@ -931,6 +915,37 @@ password=foobar
         b.wait_in_text(".pf-c-menu__item.pf-m-selected", "Custom crypto policy")
         b.click("#crypto-policy-dialog button.pf-c-button.pf-m-link")
 
+    @skipImage("crypto-policies not avaiable", "fedora-coreos", "debian-stable", "debian-testing", "ubuntu-2204", "ubuntu-2004", "ubuntu-stable", "arch")
```
martinpitt, line 918: avai*l*able; it's also wrong in the existing skipImage already, can you please fix while you are at it?

martinpitt (changes requested):
Cheers!

pkg/systemd/overview-cards/cryptoPolicies.jsx
martinpitt: Current version still has two spinners:

image

This is broken -- they even rotate in opposite directions! :fearful:

pkg/systemd/overview-cards/cryptoPolicies.jsx
martinpitt: I tried to switch away from FIPS. It's much faster than enabling FIPS, but it still takes some 5s. Would be nice to make this consistent.

I.e. drop the extra `<Spinner>` (it looks really weird, see screenshot above), always says "Applying new policy...", and always do this.

martinpitt:
Aside from the spinner/progress issue from above, the only oddity that I found was the inconsitent FIPS state after `update-crypto-policies --set FIPS`:

image

The "active" label there is misleading IMHO. Could we perhaps not show anything as active then? Or maybe a red "not properly enabled"?

pkg/systemd/overview-cards/cryptoPolicies.jsx
martinpitt: I ran `update-crypto-policies --show` and `--set` with a few values, and `--show` and the config file always coincide for me. So just reading the config seems ok?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -83,29 +90,44 @@ export const CryptoPolicyRow = () => {
...
 
+const setPolicy = (policy, setError, setInProgress) => {
+    setInProgress(true);
+
+    const promises = [];
+    if (policy === "FIPS") {
+        promises.push(cockpit.spawn(["fips-mode-setup", "--enable"], { superuser: "require", err: "message" }));
+    } else {
+        promises.push(
+            new Promise(function(resolve, reject) {
```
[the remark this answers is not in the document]
jelly: The issue I faced was the outer promise (cockpit.file) already completing before fips-mode-setup was done leading to fips not being applied.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -83,29 +90,44 @@ export const CryptoPolicyRow = () => {
...
+                cockpit.file("/proc/sys/crypto/fips_enabled").read()
```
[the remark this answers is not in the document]
jelly: I thought we shouldn't run commands when it's not required. But I also found that fips-mode-setup should run fine if you run it multiple times in a row.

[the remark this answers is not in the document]
jelly: So I'll adjust the code so it doesn't check anymore if it needs to disable fips mode.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -83,29 +90,44 @@ export const CryptoPolicyRow = () => {
...
+                        });
+            })
+        );
+        promises.push(cockpit.spawn(["update-crypto-policies", "--set", policy], { superuser: "require", err: "message" }));
+    }
+
+    Promise.all(promises).then(() => {
+        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" })
+                .catch(error => setError(error));
+    })
```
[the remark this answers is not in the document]
jelly: Seems we need smarter computers:

Moving the indentation around causes:
```
ERROR in
/home/jelle/projects/cockpit/pkg/systemd/overview-cards/cryptoPolicies.jsx
  107:1  error  Expected indentation of 12 spaces but found 11  indent
  108:1  error  Expected indentation of 12 spaces but found 11  indent

✖ 2 problems (2 errors, 0 warnings)
```

test/verify/check-system-info
jelly: It doesn't seem to be the case, what I think is happening is that some "FIPS" module has to be loaded in initramfs when enabled and a boot flag set and disabling only switches the boot flag? I'd have to do some more digging to figure that out.

garrett:
> The "active" label there is misleading IMHO. Could we perhaps not show anything as active then? 

Agreed.

> Or maybe a red "not properly enabled"?

It could be "Pending".

pkg/systemd/overview-cards/cryptoPolicies.jsx
jelly: Dropping the spinner sounds fine, made the text more general as also switching policies takes a "while"

jelly:
> As mentioned on IRC; the inconsistent mode would say "inconsistent" in the label bubble and use the light orange color, reminiscent of a warning:
> 
> FIPS-inconsistent
> 
> https://www.patternfly.org/v4/components/label/design-guidelines:
> 
> image

Agreed

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -76,41 +80,49 @@ export const CryptoPolicyRow = () => {
...
+const setPolicy = (policy, setError, setInProgress) => {
+    setInProgress(true);
+
+    const promises = [];
+    if (policy === "FIPS") {
+        promises.push(cockpit.spawn(["fips-mode-setup", "--enable"], { superuser: "require", err: "message" }));
+    } else {
+        promises.push(cockpit.spawn(["fips-mode-setup", "--disable"], { superuser: "require", err: "message" }).then(() =>
+            cockpit.spawn(["update-crypto-policies", "--set", policy], { superuser: "require", err: "message" })));
+    }
```
martinpitt, line 100: This is always exactly one promise, so turn this into a `const promise` with a ternary.

Or maybe even better, don't chain the f-m-s and u-c-p promises, but run them in parallel. I think u-c-p doesn't mind, right? Then you can also factorize the f-m-s call and just conditionalize the `--enable` vs. `--disable`, not the whole spawn.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -76,41 +80,49 @@ export const CryptoPolicyRow = () => {
...
+
+    Promise.all(promises).then(() => {
+        cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" }).catch(error => setError(error));
+    })
```
martinpitt, line 104: I know it's not relevant here, as setPolicy() doesn't return a promise. But this is a rather subtle anti-pattern to get used to, also in case we'll ever use this in a different context. This code block as a whole will *not* wait for the inner spawn. Please drop the `{}` in the then arrow, so this is properly functional and chained.

(If this makes no sense to you, please do speak up! Happy to explain.)

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -76,41 +80,49 @@ export const CryptoPolicyRow = () => {
...
-                                .catch(error => setError(error));
-                    } else {
-                        cockpit.file(applyNeedRebootStamp, { superuser: "require" }).replace("\n")
-                                .then(() => close());
-                    }
-                })
-                .catch(error => setError(error));
-    };
-
+    console.log(isInconsistentPolicy(currentCryptoPolicy, fipsEnabled));
```
martinpitt, line 119: Debugging leftover.

martinpitt (changes requested):
Thanks! Some tiny issues and some bikeshedding (feel free to ignore that one if you don't care). I re-tested this interactively, and can't find anything wrong any more. Nice work! :clap:

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -76,41 +80,46 @@ export const CryptoPolicyRow = () => {
...
+    if (policy === "FIPS") {
+        promises.push(cockpit.spawn(["fips-mode-setup", "--enable"], { superuser: "require", err: "message" }));
+    } else {
+        promises.push(cockpit.spawn(["fips-mode-setup", "--disable"], { superuser: "require", err: "message" }).then(() =>
+            cockpit.spawn(["update-crypto-policies", "--set", policy], { superuser: "require", err: "message" })));
+    }
+
+    Promise.all(promises).then(() => cockpit.spawn(["shutdown", "--reboot", "now"], { superuser: "require", err: "message" }).catch(error => setError(error)))
+            .catch(error => setError(error))
+            .finally(() => setInProgress(false));
```
martinpitt, line 104: Sorry, I missed that before -- .then and .catch belong on their own line, and be aligned. Now this has *two* .catch handlers.

martinpitt (changes requested):

pkg/systemd/overview-cards/cryptoPolicies.jsx
jelly: Aligned, need to figure out why eslint in my editor misindents this :|

jelly:
```
/tmp/source/pkg/systemd/overview-cards/cryptoPolicies.jsx
  103:1  error  Expected indentation of 12 spaces but found 11  indent
  104:1  error  Expected indentation of 12 spaces but found 11  indent
```

GRMBBBL!!

martinpitt: Haha -- I told you, you really need to align `.then()` and `.catch` :grin: . But let's not beat this any more now, it's good enough.

martinpitt: I meant, put `.then()` on a line of its own.

martinpitt (approved):
Cheers! (Waiting for green now)

martinpitt:
Whoa, what [happened here](https://logs.cockpit-project.org/logs/pull-17170-20220405-161026-c47479eb-arch/log.html#271-2)? Several failures like `ValueError: invalid literal for int() with base 10: ''` when it tries to call `self.machine.system_before(258)`. Could this have something to do with it?

    error: failed to initialize alpm library:
    (root: /, dbpath: /var/lib/pacman/)

Retrying once to compare, but this smells weird -- the [last arch refresh](https://github.com/cockpit-project/bots/pull/3166) was 5 days ago already.

jelly:
> Whoa, what [happened here](https://logs.cockpit-project.org/logs/pull-17170-20220405-161026-c47479eb-arch/log.html#271-2)? Several failures like `ValueError: invalid literal for int() with base 10: ''` when it tries to call `self.machine.system_before(258)`. Could this have something to do with it?
> 
> ```
> error: failed to initialize alpm library:
> (root: /, dbpath: /var/lib/pacman/)
> ```
> 
> Retrying once to compare, but this smells weird -- the [last arch refresh](https://github.com/cockpit-project/bots/pull/3166) was 5 days ago already.

I see packagekit dumps core first so I wonder, did it somehow fubar /var/lib/pacman?! (Which sounds weird)

martinpitt:
@jelly , @allisonkarlitskaya : I added a screenshot and a release note. Can you please proofread? Please edit in-place.

PR #17172, opened 2022-03-23
Configuration card improvements

jelly:
Made the configuration card a little bit more consistent with the change/edit options and drop the information bubble when it's not required (when ntp is already successfully synchronised)

### DIFF
--- a/pkg/lib/serverTime.js
+++ b/pkg/lib/serverTime.js
@@ -26,7 +26,7 @@ import {
     Select, SelectOption, SelectVariant,
     Spinner, TimePicker
 } from '@patternfly/react-core';
-import { CloseIcon, ExclamationCircleIcon, InfoCircleIcon, PlusIcon } from "@patternfly/react-icons";
+import { CloseIcon, ExclamationCircleIcon, PlusIcon } from "@patternfly/react-icons";
 import { show_modal_dialog } from "cockpit-components-dialog.jsx";
 import { useObject, useEvent } from "hooks.js";
 
@@ -374,12 +374,7 @@ export function ServerTimeConfig() {
     let ntp_status = null;
     if (ntp && ntp.active) {
         let icon; let header; let body = ""; let footer = null;
-        if (ntp.synch) {
-            icon = <InfoCircleIcon className="ct-info-circle" />;
-            header = _("Synchronized");
-            if (ntp.server)
-                body = <div>{cockpit.format(_("Synchronized with $0"), ntp.server)}</div>;
-        } else {
+        if (!ntp.synch) {
             if (ntp.server) {
                 icon = <Spinner size="md" isSVG />;
                 header = _("Synchronizing");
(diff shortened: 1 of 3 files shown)


pkg/lib/serverTime.js
```diff
@@ -374,12 +374,7 @@ export function ServerTimeConfig() {
     let ntp_status = null;
     if (ntp && ntp.active) {
         let icon; let header; let body = ""; let footer = null;
-        if (ntp.synch) {
-            icon = <InfoCircleIcon className="ct-info-circle" />;
-            header = _("Synchronized");
-            if (ntp.server)
-                body = <div>{cockpit.format(_("Synchronized with $0"), ntp.server)}</div>;
```
martinpitt, line 381: Hmm, I kinda like that warm fuzzy feeling about "yes, your clock is okay". Does it get in the way somehow?

martinpitt (changes requested):

pkg/lib/serverTime.js
jelly: It doesn't get in the way, but it does feel like it's signalling something while everything is ok.

pkg/tuned/dialog.jsx
```diff
@@ -102,20 +102,24 @@ export const TunedPerformanceProfile = () => {
...
+                <span>{btnText}</span>
+            </Tooltip>
             <Button id="tuned-status-button"
+                    className="tuned-status-button"
                     isAriaDisabled={btnText == "error" || state == "not-installed" || !superuser.allowed}
                     isInline
                     onClick={() => setIsOpen(true)}
                     variant='link'>
-                {btnText}
+                {_("change")}
```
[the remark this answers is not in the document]
jelly: I agree, that it makes it more inconsistent but the actions in the `Configuration` card are different, the hostname has an `edit` button and the rest does not.

garrett (changes requested):
I'm not sure what the goal is for this PR.

Is it to make the configuration card more consistent? Because it isn't in this PR's current state.

In our call, I talked about having each item have an action. RIght now, that's not the case, unless I'm not looking at this PR:

Screenshot 2022-03-28 at 11-55-11 Overview - garrett@Bolt

We have some with none, some with change, some with edit, some without an action, some are lowercase, etc.

- **Edit**: Actually edit a value (that is, you're actively editing the value of something)
- **Change**: Change to something else (you're not editing the crypto policy or performance profile; you're changing to a different pre-generated one)

However:

https://www.patternfly.org/v4/ux-writing/terminology

> Edit: (v.) Use to describe making changes to an object (like a file, configuration, or policy).
> - Notes: Modify and Change are not recommended for this use case.
>
> Change: Do not use. See Edit.
> Modify: Do not use. See Edit.

So I guess we should make everything "Edit", according to PF?

What about something like this? (I don't like the misalignment)
overview-configuration-misaligned

To further change things, here's what it looks like with the actions aligned (use this mockup instead, not the one above):
overview-configuration(1)

As the layout is already a table, these actions could be in each row.

We could make it so that if you click on the value, then the action happens too.

PR #17174, opened 2022-03-24
test: Enable FreeIPA tests on debian-testing

martinpitt:
The package is finally available again, and got added to the image in [1].

Adjust the mock chrony.service hack a bit -- this bug is still around,
and I filed [2][3] about it. chrony.service is now masked by default, so
unmask it. And systemd does not accept symlinks to other units any more,
so replace that with a copy.

[1] https://github.com/cockpit-project/bots/pull/3132
[2] https://launchpad.net/bugs/1966181
[3] https://bugs.debian.org/1008195

 - Requires https://github.com/cockpit-project/bots/pull/3139

### DIFF
--- a/test/verify/check-system-realms
+++ b/test/verify/check-system-realms
@@ -619,8 +619,11 @@ class TestIPA(TestRealms, CommonTests):
         super().testUnqualifiedUsers()
 
         # now try 2FA with OTP
-        # This does not yet work with sssd < 2.2.2-1 on Ubuntu
-        if m.image in ["ubuntu-2004", "ubuntu-stable"]:
+        if m.image == "ubuntu-2004":
+            # This does not yet work with sssd < 2.2.2-1
+            return
+        if "debian" in m.image or "ubuntu" in m.image:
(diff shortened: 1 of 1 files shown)


martinpitt:
Hmm, [that sucks a bit](https://logs.cockpit-project.org/logs/pull-17174-20220324-100258-e1d10564-debian-testing-bots-3132/log.html#305) -- all TestIPA tests work fine for me locally. The `Process 2423 (nslookup) of user 0 dumped core.` breaks two tests, this will require a naughty, but there's no current stack trace right now. I'll check if the new version in unstable fixes things, as that already fixes [another bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008194).

martinpitt:
The `test{Unq,Q}ualifiedUsers` failure are both due to an `ipa-client-install --uninstall` crash. Reported as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008209 and created a known issue https://github.com/cockpit-project/bots/issues/3138

marusak (approved):
Thank you!

PR #17175, opened 2022-03-24
test: Fix nslookup call

martinpitt:
`-type` is not a valid option for nslookup; the documented way to look
up a particular DNS entry type is `type=value`. It happened to work for
older versions, but with 9.18.1 (in Debian testing) it now gives funny
and broken results.

See [nslookup(1)](https://linux.die.net/man/1/nslookup)


martinpitt:
This was bogus after all. Argh about bad documentation and highly non-standard CLI..

PR #17176, opened 2022-03-24
test: Simplify ABRT tests to only use the API key

martinpitt:
Now all Fedoras have a libreport version which only accepts an API key.
Revert the dynamic bits of commit 2b20d62b2e68, and only send the API
key.

### DIFF
--- a/test/verify.fmf
+++ b/test/verify.fmf
@@ -16,7 +16,6 @@ require:
   - libvirt-python3
   - nfs-utils
   - python3-tracer
-  - python3-packaging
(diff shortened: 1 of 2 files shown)


martinpitt:
[unrelated flake](https://logs.cockpit-project.org/logs/pull-17176-20220324-111239-bb8921de-fedora-35/log.html#303-2) (third affected retry failed), rolling the dice again :game_die:

marusak (approved):
Thanks!

PR #17177, opened 2022-03-24
Require the user to fully opt into sudo

### LINKED ISSUE #17169: remote host tries sudo three times
### Explain what happens

I am having an issue where due to the policies that we have on our machines cockpit is causing users accounts to be locked due to 3 failed password attempts with sudo. 

I have reproduced this bug with stock CentOS Stream virtual machines fully up to date as of March 22, 2022. On both machines (CentOSStream1 and CentOSStream2) all I have done is install the ISO and enable cockpit, and set the hostname through cockpit. Then from cockpit on  CentOSStream1 I added CentOSStream2 under the hosts menu. Then I logged out. 

This appears to be a caching issue since under normal usage we do not consistently see the locking. However it can be consistently reproduced with the following steps:

1. Restart web browser
2. Open private tab
3. Navigate to cockpit
4. Login
5. Navigate to a different host through the cockpit menu
6. Watch /var/log/secure to see that sudo had 3 failed login attempts

(EDIT - Video recording is a little choppy my machine seems to be having some slowness that is hampering the ability to record smoothly but this still should show the issue being reproduced. If a new less choppy video would help I can get a better one.)

https://user-images.githubusercontent.com/3435186/159513851-f5b8daef-8b96-4ee3-b296-b90be440a31f.mov

### Version of Cockpit

264.1

Navigation & Shell

### Server operating system

CentOS

CentOS Stream Up to date as of March 22 2022

Firefox, Chrome, Edge

### System log

```shell
/var/log/secure:
Mar 22 08:56:09 centosstream2 sshd[3555]: Accepted publickey for admin from 172.16.255.130 port 45142 ssh2: RSA SHA256:f84j2jRjR76q8GBdFYmGPYcxpPKNPHjxipCbfbhvIQk
Mar 22 08:56:09 centosstream2 sshd[3555]: pam_unix(sshd:session): session opened for user admin by (uid=0)
Mar 22 08:56:09 centosstream2 polkitd[969]: Registered Authentication Agent for unix-session:4 (system bus name :1.114 [cockpit-bridge], object path /org/freedesktop/PolicyKit1/AuthenticationAgent, locale en_US.UTF-8)
Mar 22 08:56:09 centosstream2 sudo[3583]: pam_unix(sudo:auth): authentication failure; logname= uid=1000 euid=0 tty= ruser=admin rhost=  user=admin
Mar 22 08:56:16 centosstream2 sudo[3583]:   admin : 3 incorrect password attempts ; TTY=unknown ; PWD=/run/user/1000 ; USER=root ; COMMAND=/bin/cockpit-bridge --privileged
Mar 22 08:56:16 centosstream2 polkitd[969]: Operator of unix-session:4 FAILED to authenticate to gain authorization for action org.freedesktop.packagekit.system-sources-refresh for system-bus-name::1.121 [cockpit-bridge] (owned by unix-user:admin)
Mar 22 08:56:40 centosstream2 polkitd[969]: Operator of unix-session:4 FAILED to authenticate to gain authorization for action org.freedesktop.packagekit.system-sources-refresh for system-bus-name::1.205 [cockpit-bridge] (owned by unix-user:admin)
```

mvollmer:
- [x] Add checkbox to "Administrative Access" dialog to control whether Cockpit will try on next connection.
- [x] Update tests


pkg/shell/superuser.jsx
```diff
@@ -73,6 +73,11 @@ class UnlockDialog extends React.Component {
                                 value={state.prompt.value}
                             />
                         </FormGroup>
+                        <FormGroup>
+                            <Checkbox onChange={this.props.change_remember_access_level_on_unlock}
+                                      isChecked={this.props.remember_access_level_on_unlock}
+                                      label={ _("Automatically switch to administrativ access on next login") } />
```
martinpitt, line 79: administrativ*e*

pkg/shell/superuser.jsx
```diff
@@ -88,7 +93,14 @@ class UnlockDialog extends React.Component {
...
         } else if (state.message) {
             title = _("Administrative access");
-            body = <p>{state.message}</p>;
+            body = (
+                <>
+                    <p>{state.message}</p>
+                    <br />
+                    <Checkbox onChange={this.props.change_remember_access_level_on_unlock}
+                              isChecked={this.props.remember_access_level_on_unlock}
+                              label={ _("Automatically gain administrativ access on next login") } />
```
martinpitt, line 102: same "administrativ" typo.  However, as this is quite a complex translatable string, could it use the exact same one as above? the difference is marginal, and does not seem too relevant?

martinpitt (changes requested):

pkg/shell/superuser.jsx
mvollmer: Yes, they should be the same, sloppy coding.  Thanks for catching!

pkg/shell/superuser.jsx
mvollmer: Fixed.

pkg/shell/superuser.jsx
mvollmer: Fixed.

mvollmer:
I think the TestMultiMachine.testSshKeySetup failure is a known flake.  The test has to pass three times here, which is a bit hard for it.

mvollmer:
`fedora-35` shows the expected pixel test changes.

image

image

pkg/shell/superuser.jsx
```diff
@@ -73,6 +73,11 @@ class UnlockDialog extends React.Component {
                                 value={state.prompt.value}
                             />
                         </FormGroup>
+                        <FormGroup>
+                            <Checkbox onChange={this.props.change_remember_access_level_on_unlock}
+                                      isChecked={this.props.remember_access_level_on_unlock}
+                                      label={ _("Automatically switch to administrative access on next login") } />
```
martinpitt, line 79: This may not always work, right? Especially when you use SSH keys (which we heavily promote now) or 2FA. So ISTM that this promises too much for SSH. Can we restrict this to localhost only, and always make it explicit for SSH hosts?

mvollmer: Could you elaborate?  Should we have no checkbox at all for connections via SSH?

martinpitt: I haven't thought this through, and I don't have a consistent proposal yet -- I just think that like this it's too confusing. It sounds like Cockpit would do long-term credential storage somehow, or that this is a promise. In many cases it's not technically possible to "automatically switch to admin access". It will sometimes fail locally as well, but much more often with SSH.

Maybe a wording change like "Automatically attempt this on next login" would suffice already. As it's off by default, admins then at least have a choice.

I could also live with not showing the checkbox at all for remote hosts (and never attempting sudo automatically, ever). However, I think we should keep that possibility for localhost at least, both for convenience and backwards compatibility. This would then be a pretty arbitrary design choice.

martinpitt: Right, in that spirit. I was aiming for something shorter, as this is both rather long, and also needs to be translated. But maybe @garrett has a better idea here.

mvollmer: I think we can be even more exact (and verbose): "Try to automatically switch to administrative access on next login by providing the login password also to sudo".

garrett: Uh... if it's just next time someone logs in, then it's kind of pointless, especially if someone has to check that each time.

What about something simpler?

"Use administration mode when logging in."

---

Also, where can someone undo this? Would we require them to drop their admin privs and then try to raise them again? That seems a bit odd, if so. The right place to undo something like, _after being set_ in the admin prompt, this would be under the session menu, as a toggle.

(Simply put: Every "remember" checkbox in a modal needs to have a setting somewhere outside that modal to undo it.)

garrett: Perhaps even toss in an always:

☑ Always use administration mode when logging in.

mvollmer: > Also, where can someone undo this? Would we require them to drop their admin privs and then try to raise them again?

The way it works right now in this PR is that the "Switch to limited access" dialog will do this.  The dialog also explains this: "Your browser will remember your access level across sessions."

And, yeah, this is diffcult for the user to put all this together and understand what is really happening...

mvollmer:
> Additionally, "Password for admin:" shouldn't have a colon. And I think the label is problematic, as it's not admin's password, but your account password, right?

The prompt comes directly from "sudo".  It is indeed the account password of the "admin" user that needs to be entered in this case.  (The account in this example is called literally "admin".) We can remove the trailing colon, however.

mvollmer:
What about this: Instead of letting the user control remotely what will happen at the next login, we can put a checkbox right into the login dialogs that controls whether sudo will be executed during login. This is the most direct way and easiest to explain.  Our current machinery for remembering past choices will provide the default for that checkbox.

Going further, we can try to fold the sudo dialog into the login sequence.  The first prompt will be silently answered with the login password (or the text input for it will be prefilled with the login password). Any other prompt will be shown as part of login in.

garrett:
> What about this: Instead of letting the user control remotely what will happen at the next login, we can put a checkbox right into the login dialogs that controls whether sudo will be executed during login. This is the most direct way and easiest to explain. Our current machinery for remembering past choices will provide the default for that checkbox.

Do you mean a remember checkbox on login? So we'd be back to a check on the login page? :rofl:

> Going further, we can try to fold the sudo dialog into the login sequence. The first prompt will be silently answered with the login password (or the text input for it will be prefilled with the login password). Any other prompt will be shown as part of login in.

Wouldn't this cause people to enter their password twice? When logging in and then again when becoming sudo? And in many cases, it's the same exact password, right? Not a great UX.

mvollmer:
> Do you mean a remember/reuse checkbox on login? So we'd be back to a check on the login page?

Yes, exactly.  IIRC, we didn't remove it because it was a bad idea in itself, but we thought it is no longer necessary.  Also, it did too much and was also reusing the password for SSH logins.

But giving the user an opportunity to decide during login whether or not sudo will also be started during that process makes sense to me.

But then again, find a default for that checkbox is not easy either, I guess.

I'll try to write some user stories.

garrett:
> we didn't remove it because it was a bad idea in itself, but we thought it is no longer necessary

Also, basically every other website in existence with a sign in form has a checkbox to remember the user so they don't have to sign in again and again. Having a checkbox there that does anything else whatsoever is not a great idea, as people will generally not read it (no matter what we put there) and will assume it's a "remember me" checkbox.

mvollmer:
I think the changed default of the bridge is uncontroversial: #17196.

mvollmer:
I wrote up something... I don't think there are any surprises in here...

### Sudo scenarios

Let's play through some scenarios.

Sudo configs:

- "standard sudo": sudo asks for account password

- "custom sudo": sudo asks for something that is not the account
  password, maybe a second secret

We always assume that "something bad" happens when sudo fails, like
accounts getting locked or security dashboards getting blinking red
alerts.

Login types:

- "password login": login on the main login screen or in a "connect to
  host" dialog, by entering the account password.

- "key login": login to a remote machine over SSH with keys. The key
  might need to be unlocked with a password, or the login might happen
  without any interaction.


Checkbox UX variants:

- "no checkboxes": current UX, Cockpit remembers the current access
  level and tries to get it back on next login to the same machine on
  the same broweser

- "checkbox at login": at each login, Cockpit asks whether to run sudo
  with the login password, default is always off.

  The label of the checkbox needs to explain this: During the login
  process that is started when you submit this dialog, sudo will be
  executed and the password from the login dialog will be used to
  answer sudo's prompt. This should only be done if you expect that
  password to work with sudo.

- "checkbox at login with memory": at each login, Cockpit asks whether
  to run sudo with the login password, default is whatever the access
  level was in the last session.

- "checkbox at sudo": in each "Admin access" dialog, Cockpit asks
  whether to run sudo on the next login to this machine on this
  browser.  That checkbox will always be "off" initially.

  The label of the checkbox needs to explain this: During the login
  process for the next session that is started in the same browser for
  the same user and same host as the current session, sudo will be
  executed and the password from the login dialog will be used to
  answer sudo's prompt. This should only be done if you expect that
  password to work with sudo.

  The way to stop sudo from being executed on the next login is to
  switch back to "limited access" in the current session.

Scenarios:

- standard sudo / password login / no checkboxes

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access.  User logs out and in again.
Session is in admin access mode.  "Something bad" has not happened.

This is the main scenario that we wanted to support.  People with a
standard sudo setup should be able to be admin in every session
without any extra work.

- standard sudo / key login / no checkboxes

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access.  User logs out and in again.
Session is in limited access mode and user needs to switch to it
explicitly again.  "Something bad" has not happened because Cockpit
doesn't try to run sudo during login when there is no password.

Since "key login" never runs sudo, we can ignore it below.  "Somethin
bad" should never happen with key login.  If it does, it's a bug.  We
should make sure.

- custom sudo / no checkboxes

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access by providing the custom sudo
credentials.  User logs out and in again.  Session is in limited
access mode.  "Something bad" has happened because Cockpit did run
sudo and gave it the account password, but sudo failed with that.

This is the scenario we want to fix and prevent "something bad" from
happening.  This is important, "something bad" happening is a bug that
needs fixing.

The easy fix would be to never run sudo during login. (This would also
allow us to remove a lot of code and complexity.) However, this would
kill convenience for the "standard sudo" scenario.  Let's see if we
can keep the convenience by letting the user opt into the convenience.

- standard sudo / checkbox at login

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access.  User logs out and in again.
Session is in limited access mode again.  User switches to admin
access explicitly again, like they are used to from the command line.

Eventually, they might notice the checkbox (which is probably hidden
behind the "Other options" expander on the login page, but readily
visible in the "connect to host" dialog) and figure out that they can
tick it instead of clicking the access level indicator and then typing
their password again.  They have to tick the checkbox on every login
since it will default to "off".  So it's a way to trade two clicks
(expand, tick) for one click and typing a password.


- standard sudo / checkbox at login with memory

Like above, but once user has discovered the checkbox, they get admin
access in each session automatically.  To switch that off, they need
to untick the checkbox, or switch to limited access during the
session.

The default for the checkbox depends on the user and host names, so it
might change while the user is typing into the dialog.

- standard sudo / checkbox at sudo

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access.  User logs out and in again.
Session is in limited access mode again.  User switches to admin
access explicitly again, like they are used to from the command line.

They will very likely notice the checkbox in the dialog since its
right there.  They tick it and from that point every session starts
with admin access until they explicitly switch back to limited access.

The challenge here, IMO, is that the checkbox controls stuff that
happens at a very different time in a different place than the dialog
it appears in.  This makes it harder to explain what it does, and
harder for the user to understand whether it has worked or not.

- custom sudo / checkbox at login

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access by providing the custom sudo
credentials.  User logs out and in again.  Session is in limited
access mode, as expected by the user, and user switches every time
explicitly by entering the custom sudo secrets, as they are used to
from the command line.

User might never find the checkbox in the login dialogs, but if they
do they hopefully realize that it will not help them.  If user ticks
the checkbox anyway, "something bad" will happen immediately, but only
once.

- custom sudo / checkbox at login with memory

Exactly like above.  Cockpit will notice the failed sudo attempt and
set the sudo default to "off".

- custom sudo / checkbox at sudo

User logs in for the first time ever.  Session is in limited access
mode.  User switches to admin access by providing the custom sudo
credentials.  User logs out and in again.  Session is in limited
access mode, as expected by the user, and user switches every time
explicitly by entering the custom sudo secrets, as they are used to
from the command line.

They will very likely notice the checkbox in the dialog since its
right there.  From the context and label they hopefully understand
that it wont help them.  If user ticks the checkbox anyway, "something
bad" will happen on the next login, but only once.

mvollmer:
> * We are getting further and further away from showing the login page (remote host switcher, flatpak, bastion host, smart card login, Foreman/Insights integration), so any effort there would have to be duplicated into the session anyway.

Do we have the account password in these cases?  If not, these cases are equivalent to the "key login" case above, and will never get a convenient sudo during login. (They can get inconvenient sudo with the "Admin access" dialog, of course.)
If they do have a password, we would need to add the checkbox to the place where the password is entered, one by one, which is a good argument against "checkbox at login", of course.

Remote host switcher: has a password, and it would get a checkbox
Flatpak: gets admin via desktop polkit prompt, no?
Bastion host: password, our own login page, so it gets the checkbox
Smart card login: no password, no checkbox possible
Foreman: Might be configured with password for "remote execution", but it's not accessible to cockpit, so no checkbox possible
Insights: I don't know

So I think "external login" is not a good argument against the checkbox in our own login dialogs, since external logins wont be able to run sudo during login anyway.  Running sudo only makes sense during "external login" if sudo grants access without any prompting, and then getting admin in the session is only two clicks anyway, so gaining convenience is not very important.

mvollmer:
> * this would have to use the exact same data store and logic. → combinatorial explosion and extra maintenance

I was thinking it doesn't require any changes to what we have, just some code to let the user override the stored value from local storage via a checkbox.  I don't see the explosion I am afraid.

mvollmer:
> We need to be very careful about ensuring that this and the checkbox in the session's sudo dialog always coincide.

I don't think we should have both a checkbox during login, and a checkbox during sudo.

mvollmer:
> How do we make sure about the "only once"? By resetting the "always" memory if the automatic attempt fails?

Yes, exactly.

martinpitt:
> > so any effort there would have to be duplicated into the session anyway.

> Do we have the account password in these cases?

Sorry, I was confused. You were of course not talking about removing the privilege indicator dialog, just about not adding the checkbox there. (Your case analysis for "these cases" looks correct)

> combinatorial explosion and extra maintenance

I meant that we'd have to make sure that the login page's check box and a possible checkbox in the privilege dialog coincide. If we only get one, that's not an issue, of course.

> I don't think we should have both a checkbox during login, and a checkbox during sudo.

Yes, agreed.

I must say, despite the above I personally still prefer adding the checkbox to the priv dialog instead of the login page. The former has better context, and there is no mental disconnect between the two different places. The checkbox on the login page sounds hard to explain, hard to discover, and hard to have a sensible default -- e.g. in the priv dialog we could at least *try* to check if it asks for exactly one secret (and if it is "password"). On the login page we don't even know yet whether the user even is an admin, let alone what kind of secrets sudo wants to know.

mvollmer:
> I must say, despite the above I personally still prefer adding the checkbox to the priv dialog instead of the login page. The former has better context, and there is no mental disconnect between the two different places.

Ok, let's do this!

> The checkbox on the login page sounds hard to explain, hard to discover, and hard to have a sensible default

The default would be whatever it was the last time.  The initial default would be "off".

> -- e.g. in the priv dialog we could at least _try_ to check if it asks for exactly one secret (and if it is "password"). On the login page we don't even know yet whether the user even is an admin, let alone what kind of secrets sudo wants to know.

Yeah, but the user knows, no?

Anyway, I'll polish this up according to @garrett's mockup.

martinpitt:
[eslint error](https://github.com/cockpit-project/cockpit/runs/6270941382?check_suite_focus=true#step:4:299)

pkg/shell/superuser.jsx
```diff
@@ -30,8 +30,12 @@ function sudo_polish(msg) {
     if (!msg)
         return msg;
 
+    msg = msg.trim();
     msg = msg.replace(/^\[sudo] /, "");
     msg = msg[0].toUpperCase() + msg.slice(1);
+    console.log("P", msg, msg[msg.length-1]);
```
martinpitt, line 36: debugging leftover

pkg/shell/superuser.jsx
```diff
@@ -30,8 +30,12 @@ function sudo_polish(msg) {
...
+    if (msg.length >= 1 && msg[msg.length - 1] == ":")
+        msg = msg.slice(0, -1);
```
martinpitt, line 38: Nitpicking, but it feels to me that just appending another `.replace(/:$/, "")` to line 34 would be a bit cleaner.

pkg/shell/superuser.jsx
```diff
@@ -46,6 +50,12 @@ class UnlockDialog extends React.Component {
         let body = null;
         let footer = null;
 
+        const auto_checkbox = (
+            <Checkbox onChange={this.props.change_remember_access_level_on_unlock}
+                      isChecked={this.props.remember_access_level_on_unlock}
+                      label={_("Use administrative mode when logging in")}
+                      description={_("Automatically switching to administrative mode may not work as expected, especially on remote hosts and when using 2FA")} />);
```
martinpitt, line 57: s/and/or/ ? "Remote hosts" is a bit confusing, as the machine you are connecting to (to port 9090) is already likely "remote", but that's not what we mean here. Perhaps "especially with SSH keys or when using two-factor authentication"?

test/common/testlib.py
```diff
@@ -747,7 +747,7 @@ def become_superuser(self, user=None, password=None):
 
         self.open_superuser_dialog()
         self.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", f"Password for {user or 'admin'}:")
-        self.set_input_text(".pf-c-modal-box:contains('Switch to administrative access') input", password or "foobar")
+        self.set_input_text(".pf-c-modal-box:contains('Switch to administrative access') input[type=password]", password or "foobar")
```
martinpitt, line 750: This should be a separate commit (there are more instances of this below).

test/verify/check-shell-keys
```diff
@@ -172,7 +172,7 @@ class TestKeys(MachineCase):
                 b.wait_not_present(f"tr[data-name='{identifier}']")
 
         def waitKeyLoaded(identifier, enabled):
-            b.wait_visible(f"tr[data-name='{identifier}'] input[type=checkbox]" + (":checked" if enabled else ":not(checked)"))
+            b.wait_visible(f"tr[data-name='{identifier}'] input[type=checkbox]" + (":checked" if enabled else ":not(:checked)"))
```
martinpitt, line 175: Ouch! Separate commit, please.

martinpitt (changes requested):
Thanks! I find the general approach very clean and appropriate now, and reasonably well understandable. I'm interested in @garrett 's opinion as well.

In the meantime, some small technical comments.

garrett (changes requested):
You're not using a label for the checkbox. Please fix. Here's a video clip of me trying to click the text to enable the checkbox... :wink: 

https://user-images.githubusercontent.com/10246/166441261-c0db82e1-17d5-49b4-9e9a-7e64c02d5376.mp4

Also: We currently remember things. Should we make the option checked by default? It's the status quo at the moment and has the least amount of friction. (If so, we'd want to remember that they opted out of it.)

test/common/testlib.py
mvollmer: Why separate?  This qualification is now necessary because there are twi `input` elements in the dialog now.

test/verify/check-shell-keys
mvollmer: Roger.

test/common/testlib.py
martinpitt: Oh, of course -- I misread that.

mvollmer:
> You're not using a label for the checkbox.

The bug was that I wasn't using `id="..."` for the checkbox. How annoying of PF to require it...

pkg/shell/superuser.jsx
mvollmer: Ouch, thanks.

pkg/shell/superuser.jsx
mvollmer: Yes, good idea.

mvollmer:
> Also: We currently remember things. Should we make the option checked by default? It's the status quo at the moment and has the least amount of friction. (If so, we'd want to remember that they opted out of it.)

Hmm, going into this, I specifically wanted people to opt _into_ automatic sudo during login, so that running sudo is only done because the user has requested it explicitly.  Running sudo might have bad consequences if it fails.

Also, if automatic sudo works for you, you only need to tick that box once, and then you can enjoy very many sessions with admin access without ever having to worry again about sudo.

But if automatic sudo does not work for you (because of 2FA, say), you have to use that dialog in every session and you have to remember every time to untick that checkbox, or else sudo will noisily fail the next time you log in.

However, we can try to detect a good default for this checkbox: If there is only one prompt, and it is "[sudo] password for <account>: ", we can default it to "on", and otherwise "off".  What about that?

martinpitt:
The above comment about the description is still current, I closed the other threads as resolved. I feel this is more between @garrett and @mvollmer now.

pkg/shell/superuser.jsx
mvollmer: What about:

"When logging in with a password, that password is automatically used to gain administrative access. This will only work if the login password is the same as the password that unlocks administrative access. This is often the case, but not when using 2FA, for example."

mvollmer:
> > If there is only one prompt
> 
> You can't really know that when you get the first prompt, can you? Honestly I'd keep the behaviour of this PR for now, and if we get some complaints we can add heuristics in a follow-up. My gut feeling is that this will be too brittle, though.

I think it can work, but it's indeed a bit magical. If the first prompt comes in and is the password prompt, we set the box to "on". When the second prompt comes in, we set it to "off". The user can override the box before hitting "Authenticate".

In fact, this scheme for predicting whether it makes sense to remember the access level might be accurate enough that it might not even need a checkbox in the dialog... :-)

test/common/testlib.py
```diff
@@ -746,8 +746,8 @@ def become_superuser(self, user=None, password=None):
         self.switch_to_top()
 
         self.open_superuser_dialog()
-        self.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", f"Password for {user or 'admin'}:")
-        self.set_input_text(".pf-c-modal-box:contains('Switch to administrative access') input", password or "foobar")
+        self.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", f"Password for {user or 'admin'}")
+        self.set_input_text(".pf-c-modal-box:contains('Switch to administrative access') input[type=password]", password or "foobar")
```
martinpitt, line 750: Adding the `[type=password]` looks like a rebase error, this shouldn't be in the "Drop trailing : from sudo prompts commit (as most other places with this change are in the next one). So if you have to re-push at all for any reason, it'd be nice to clean this up. Otherwise, not important enough.

pkg/shell/superuser.jsx
martinpitt: Honestly I'd rather make it shorter than longer.. Somebody think of the chil^Wtranslators! :grinning: Maybe:

> Using your login password to automatically switch to administrative mode using may not work as expected, especially with SSH keys or 2FA.

test/verify/check-superuser
```diff
@@ -121,11 +144,11 @@ session include system-auth
 """)
 
         b.open_superuser_dialog()
-        b.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", "Password for admin:")
```
martinpitt, line 124: Rebase error in the other direction, this belongs in the previous commit.

test/verify/check-superuser
```diff
@@ -235,8 +258,8 @@ session include system-auth
         self.login_and_go("/system", superuser=False)
         b.wait_visible(".pf-c-alert:contains('Web console is running in limited access mode.')")
         b.click(".pf-c-alert:contains('Web console is running in limited access mode.') button:contains('Turn on')")
-        b.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", "Password for admin:")
```
martinpitt, line 238: .. and so does this, and a few other places below.

martinpitt (changes requested):
Thanks!

test/common/testlib.py
mvollmer: > Adding the [type=password] looks like a rebase error, 

Yes, fixed.

test/verify/check-superuser
mvollmer: Fixed them all, I think.

garrett:
> The bug was that I wasn't using id="..." for the checkbox. How annoying of PF to require it...

PatternFly itself doesn't require ID. HTML does, therefore PatternFly has to as well. The only other way around it is by using `<label>` to wrap an element, which PatternFly doesn't support in this case. 

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label:

> To associate the `<label>` with an `<input>` element, you need to give the `<input>` an id attribute. The `<label>` then needs a for attribute whose value is the same as the input's id.

And also https://html.spec.whatwg.org/multipage/forms.html#the-label-element:

> The [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) a caption in a user interface. The caption can be associated with a specific form control, known as the [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element's labeled control, either using the [for](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for) attribute, or by putting the form control inside the [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element itself.

---

We do have a "[[action]] [action and other] cancel" pattern used on other modals now, so we _could_ "[[Authenticate]] [Authenticate and remember] Cancel". However, then it won't have a message about how it might not actually work.

pkg/shell/superuser.jsx
mvollmer: > Using your login password to automatically switch to administrative mode using may not work as expected, especially with SSH keys or 2FA.

Ok, I put that in.

pkg/shell/superuser.jsx
```diff
@@ -78,21 +81,28 @@ const UnlockDialog = ({ proxy, host }) => {
             did_prompt = true;
         };
 
+        const maybe_remember = () => {
+            if (rememberAccessLevelOnUnlock) {
+                if (key)
+                    window.localStorage.setItem(key, method);
+            }
```
cockpituous, line 88: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17177-20220927-070244-b0e468e8-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
Let's shelve this, I left a link in our secret backlog.

PR #17179, opened 2022-03-24
storage: Ignore anything mounted like a Canonical "snap"

### LINKED ISSUE #14008: Provide a way to skip /snap in storage module
Page: 
Storage

On my Ubuntu 20.04 machine, the storage module is dominated by snaps that are mounted at /snap/*. These are generally useless and get in the way of looking at actual filesystem information.

It would be useful to skip these by default or have a way to hide them.

### LINKED ISSUE #14424: On the Storage page, filter out /dev/loop*
Page:  "Storage"

Add a way to filter out /dev/loop* mounts.  These are not additional storage; they are already counted against /var/lib/snapd/snaps/*.snap and their presence takes up valuable real estate on  the Filesystems section.

Filtering them out would also remove them from the Other Devices section and eliminate the bogus warnings; "Inconsistent filesystem mount"
"The filesystem is currently mounted on /snap/canonical-livepatch/95 but will not be mounted after the next boot."

### LINKED ISSUE #17173: "The filesystem has no permanent mount point" on snap partitions
### Explain what happens

hi.. i'm not sure if this is a bug or a problem with snap , ubuntu  or cockpit , but i'm seeing a lot of warnings in the storage section about inconsistent mounts which all come from snap, and i was wondering if those can be ignored, or avoided somehow

incosistent_snaps

this is my layout

aeroscout@aessrv-561-0712:~$ lsblk
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
loop0    7:0    0  61.9M  1 loop /snap/core20/1328
loop1    7:1    0  61.9M  1 loop /snap/core20/1376
loop2    7:2    0  43.6M  1 loop /snap/snapd/14978
loop3    7:3    0  67.2M  1 loop /snap/lxd/21835
loop4    7:4    0  67.9M  1 loop /snap/lxd/22526
loop5    7:5    0  43.6M  1 loop /snap/snapd/15177
sda      8:0    0   200G  0 disk 
├─sda1   8:1    0     1M  0 part 
├─sda2   8:2    0   512M  0 part /boot
├─sda3   8:3    0     8G  0 part [SWAP]
└─sda4   8:4    0 191.5G  0 part /
sr0     11:0    1  1024M  0 rom  

$ snap list
Name    Version   Rev    Tracking       Publisher   Notes
core20  20220304  1376   latest/stable  canonical✓  base
lxd     4.0.9     22526  4.0/stable/…   canonical✓  -
snapd   2.54.4    15177  latest/stable  canonical✓  snapd

### Version of Cockpit

264-1~bpo20.04.1

Storage

### Server operating system

_No response_

20.04.04

Chrome

### System log

_No response_

mvollmer:
Fixes #17173
Fixes #14424
Fixes #14008

### DIFF
--- a/test/verify/check-storage-hidden
+++ b/test/verify/check-storage-hidden
@@ -21,6 +21,9 @@ import parent
 from storagelib import *
 from testlib import *
 
+import unittest
+import testvm
+
 
 class TestStorageHidden(StorageCase):
 
@@ -122,6 +125,32 @@ class TestStorageHidden(StorageCase):
         self.confirm()
         b.wait_visible("#storage")
         self.assertEqual(m.execute(f"grep {mount_point_2} /etc/fstab || true"), "")
+
+    @unittest.skipUnless("ubuntu" in testvm.DEFAULT_IMAGE, "Only test snaps on Ubuntu")
+    def testHiddenSnap(self):
+        m = self.machine
+        b = self.browser
+
+        self.login_and_go("/storage")
+
+        # We expect there to be at least one loop back device mounted
+        # somewhere below /snap.
+        snap_loops = []
+        devices = json.loads(m.execute("lsblk --json --list --output NAME,TYPE,MOUNTPOINT"))
+        for d in devices["blockdevices"]:
+            if d["type"] == "loop" and d["mountpoint"].startswith("/snap/"):
+                snap_loops.append(d["name"])
+        self.assertTrue(len(snap_loops) > 0)
+
+        # Make one more loopback device that we expect to see in the UI.
+        dev = self.add_loopback_disk()
+
+        # Now we wait until the regular loopback device is shown.  The
+        # snaps should not be shown.
+        b.wait_in_text("#others", dev)
+        for sl in snap_loops:
+            b.wait_not_in_text("#others", sl)
+            b.wait_not_in_text("#mounts", sl)
 
 
 if __name__ == '__main__':
--- a/pkg/storaged/fsys-panel.jsx
+++ b/pkg/storaged/fsys-panel.jsx
@@ -23,7 +23,7 @@ import { SortByDirection } from '@patternfly/react-table';
 
 import { ListingTable } from "cockpit-components-table.jsx";
 import { StorageUsageBar } from "./storage-controls.jsx";
-import { block_name, fmt_size, go_to_block, flatten } from "./utils.js";
+import { block_name, fmt_size, go_to_block, flatten, is_snap } from "./utils.js";
 import { OptionalPanel } from "./optional-panel.jsx";
 import { get_fstab_config } from "./fsys-tab.jsx";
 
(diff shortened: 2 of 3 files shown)


mvollmer:
Let's have a test, if possible.

test/verify/check-storage-hidden
```diff
@@ -123,6 +126,31 @@ class TestStorageHidden(StorageCase):
...
+
+        self.login_and_go("/storage")
+
+        # We expect there to be at least one loop back device mounted
+        # somewhere below /snap.
+        snap_loops = []
+        devices = json.loads(m.execute("lsblk --json --list --output NAME,TYPE,MOUNTPOINT"))
+        for d in devices["blockdevices"]:
+            if d["type"] == "loop" and d["mountpoint"].startswith("/snap/"):
+                snap_loops.append(d["name"])
```
martinpitt, line 142: There should be either an `else:` for the `for` loop or an assertion that `snap_loops` is not empty.

test/verify/check-storage-hidden
```diff
@@ -123,6 +126,31 @@ class TestStorageHidden(StorageCase):
...
+        print(snap_loops)
+
+        # Make one more loopback device that we expect to see in the UI.
+        dev = self.add_loopback_disk()
+
+        # Now we wait until the regular loopback device is shown.  The
+        # snaps should not be shown.
+        b.wait_in_text("#others", dev)
+        for sl in snap_loops:
```
martinpitt, line 151: ... otherwise this will be an empty loop, and no actual check.

test/verify/check-storage-hidden
mvollmer: Right, good point!

test/verify/check-storage-hidden
```diff
@@ -123,6 +126,31 @@ class TestStorageHidden(StorageCase):
...
+            b.wait_not_in_text("#others", sl)
```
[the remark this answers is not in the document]
mvollmer: Yes, very good point as well. Thanks!

test/verify/check-storage-hidden
mvollmer: Done.

test/verify/check-storage-hidden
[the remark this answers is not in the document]
mvollmer: Done.

test/verify/check-storage-hidden
```diff
@@ -123,6 +126,32 @@ class TestStorageHidden(StorageCase):
...
+            b.wait_not_in_text("#mounts", sl)
```
martinpitt, line 153: This is a good check as well. I originally meant that the string `/snap` should not appear anywhere on the page. But good enough, if you are bored enough about this by now :grin:

martinpitt (approved):
Thanks!

test/verify/check-storage-hidden
mvollmer: Cockpit wont unfortunately show the "/snap" mount point, since it is not in fstab but in a systemd mount unit...

PR #17181, opened 2022-03-24
overview: Adjust modals & select lists

garrett:
- Using updated PF menus lets us drop a little bit of overrides
- Moving the scrollbar from the modal to the menu makes sense here, but it needs some fixups

### DIFF
--- a/pkg/lib/menu-select-widget.scss
+++ b/pkg/lib/menu-select-widget.scss
@@ -4,19 +4,19 @@
 // Therefore, we're changing the visuals here locally.
 // PF4 upstream request for multi-select @ https://github.com/patternfly/patternfly/issues/4027
 .ct-menu-select-widget.pf-c-menu {
-  // It's  not really a menu, so it shouldn't have a shadow
-  box-shadow: none;
-
   // Divider is silly between the widgets in this context
   .pf-c-divider {
     display: none;
+
+    + .pf-c-menu__content {
+      // There should be minimal space between the widgets (replacing the divider)
+      margin-top: var(--pf-global--spacer--sm);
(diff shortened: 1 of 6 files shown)


garrett:
I'm wondering if we should move some of the CSS from overview to somewhere else. Marking this as a draft right now.

garrett:
Rebased on main. It would be nice if someone could finally look at this please.

This affects the selection modals in the overview page, where you can change the profiles for security and performance.

I've taken screenshots at a absurdly resized window to make it clearer what's going on.

The entire contents of the modal should not be scrollable; only the list inside should be the scrollable part. Look at where the scrollbar is and how the list's boundaries are not there in the before (the list is scrolled off).

Here's the before:

Screenshot from 2022-11-29 16-38-31

Here's after:

Screenshot from 2022-11-29 16-38-23

KKoukiou:
@garrett the language selection menu has with this a weird state with scroll for the menu and the modal. Is that expected?
Screen Shot 2022-11-29 at 17 22 46

garrett:
> @garrett the language selection menu has with this a weird state with scroll for the menu and the modal. Is that expected?

This is for the overview modals, not the language selector. The language selector is from the shell. This PR is specifically for the overview. There was supposed to be a follow-up for the shell after this was supposed to be merged.

garrett:
FWIW: We did have it working properly on the shell, but what you're seeing is a regression, when we changed the PF widgets a while ago.

But once we confirm this is the way we want to do it (or something else with the same effect), we can make the shell do things properly too.

KKoukiou (approved):

PR #17186, opened 2022-03-28
metrics: History metrics react to visibility changes

martinpitt:
Immediately refresh the history metrics when the Metrics page becomes
visible. This gives a better UX than having to wait for the next
one-minute auto-refresh timer.

Conversely, don't reload/refresh the history metrics at all if the page
is not visible, as that's just a waste of CPU and bandwidth.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1152,6 +1152,12 @@ class MetricsHistory extends React.Component {
 
         // FIXME: load less up-front, load more when scrolling
         machine_info_promise.then(() => this.initialLoadData());
+
+        cockpit.addEventListener("visibilitychange", () => {
+            // update history metrics when in auto-update mode
+            if (!cockpit.hidden && this.history_refresh_timer)
+                this.load_data(this.most_recent);
+        });
     }
 
     // load and render the last 24 hours (plus current one) initially; this needs numCpu initialized for correct scaling
@@ -1322,9 +1328,13 @@ class MetricsHistory extends React.Component {
                 // re-render
                 this.setState({ hours, loading: false });
 
-                // trigger automatic update every minute
-                if (!limit)
-                    this.history_refresh_timer = window.setTimeout(() => this.load_data(this.most_recent), 60000, false);
+                // trigger automatic update every minute when visible
+                if (!limit) {
+                    this.history_refresh_timer = window.setTimeout(() => {
+                        if (!cockpit.hidden)
+                            this.load_data(this.most_recent);
+                    }, 60000);
+                }
             }
 
             metrics.close();


martinpitt:
[Metrics failure on retry](https://logs.cockpit-project.org/logs/pull-17186-20220328-124453-9a0c0009-ubuntu-2004/log.html), retrying once to compare.

martinpitt:
@garrett: Open the metrics page, remember how the current history looks at the top. Switch to e.g. the Overview or any other cockpit page, wait for a few minutes. Go back to Metrics, then the history should be refreshed very quickly.

garrett (approved):
It looks good.

I dug around with timings using console.time() and console.timeEnd(), but I guess I'd have to put that into the channel to get the time it was requested and the time it had the data. Most of the timings on the page in metrics.jsx showed times of 1ms. But there were a few up to 24 and 52ms, which is still tiny.

(I was trying to see what the load and render time overhead might be... it feels fine really, from what I can tell too. So it's probably not a big deal, really.)

Additionally, I looked into requestTimeFrame(), but browsers usually throttle setTimeout() these days anyway. Using requestTimeFrame() (most likely in tandem with setTimeout() in this case) may improve things here too, but it's probably not really needed. We might want to consider it for some of the updates on the page, however... but I think that'd be outside of the scope of this PR.

pkg/metrics/metrics.jsx
```diff
@@ -1322,9 +1328,13 @@ class MetricsHistory extends React.Component {
                 // re-render
                 this.setState({ hours, loading: false });
 
-                // trigger automatic update every minute
-                if (!limit)
-                    this.history_refresh_timer = window.setTimeout(() => this.load_data(this.most_recent), 60000, false);
```
marusak, line 1327: What was this last argument `false`?

martinpitt:
Thanks for the benchmarking and testing, @garrett ! That coincides with my visual testing, I can't really see the update or a flicker -- it's by and large there when I switch the page.

Leaving the code review to Matej.

pkg/metrics/metrics.jsx
martinpitt: Good question -- the [API](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) allows additional arguments, but they are just passed to the called function. That lambda does not take any arguments, so that `false` was entirely ignored. Hence I dropped it. I could have mentioned that in the commit message -- I can do that at squash&merge time (unless there's some code change needed)

marusak: It is fine, don't worry about it. I also figured out it was ignored and was most likely some copy paste relic but wanted to double check.

marusak (approved):
Thank you!

PR #17187, opened 2022-03-28
eslint tweaks

allisonkarlitskaya:
- add eslint to `test/static-code`
 - make sure to catch our `#!/usr/bin/env node` scripts which don't necessarily end in `.js`
 - fix up a whole lot of problems
 - remove `eslint` from the webpack in production mode.  This is a substantial speedup.


martinpitt:
Our HTML log parser somehow doesn't like this, but the raw log shows the failure:
```
# testRetry (__main__.TestRunTest)
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/work/bots/make-checkout-workdir/test/verify/check-testlib", line 212, in testRetry
    self.assertNotRegex(stdout, b"RETRY")
  File "/usr/lib64/python3.10/unittest/case.py", line 1336, in assertNotRegex
    raise self.failureException(msg)
AssertionError: Regex matched: b'RETRY' matches b'RETRY' in b'1..1\n\nok 1 /work/bots/make-checkout-workdir/test/verify/check-system-testlib TestSystemd.testBasic # RETRY 1 (test affected tests 3 times)\n\nok 1 /work/bots/make-checkout-workdir/test/verify/check-system-testlib TestSystemd.testBasic # RETRY 2 (test affected tests 3 times)\n\nok 1 /work/bots/make-checkout-workdir/test/verify/check-system-testlib TestSystemd.testBasic\n\n# TESTS PASSED [1s on 2-cockpit-9, 1 parallel tests, 0 serial tests: ]\n'
Journal extracted to TestRunTest-testRetry-debian-testing-127.0.0.2-2901-FAIL.log.gz
  File "/work/bots/make-checkout-workdir/test/verify/check-testlib", line 212, in testRetry
    self.assertNotRegex(stdout, b"RETRY")
AssertionError: Regex matched: b'RETRY' matches b'RETRY' in b'1..1\n\nok 1 /work/bots/make-checkout-workdir/test/verify/check-system-testlib TestSystemd.testBasic # RETRY 1 (test affected tests 3 times)\n\nok 1 /work/bots/make-checkout-workdir/test/verify/check-system-testlib TestSystemd.testBasic # RETRY 2 (test affected tests 3 times)\n\nok 1 /work/bots/make-checkout-workdir/test/verify/check-system-testlib TestSystemd.testBasic\n\n# TESTS PASSED [1s on 2-cockpit-9, 1 parallel tests, 0 serial tests: ]\n'

# Result testRetry (__main__.TestRunTest) failed
# 1 TEST FAILED [118s on 2-cockpit-9]
not ok 296 test/verify/check-testlib TestRunTest.testRetry
```

.eslintrc.json
```diff
@@ -22,7 +22,7 @@
                 "MemberExpression": 2,
                 "ignoredNodes": [ "JSXAttribute" ]
             }],
-        "newline-per-chained-call": ["error", { "ignoreChainWithDepth": 2 }],
+        "newline-per-chained-call": ["error", { "ignoreChainWithDepth": 3 }],
```
martinpitt, line 25: What would be a legit use case for that? In a lot of cases even two is too much. In particular a `.then()` should never be chained if the same statement also has a `.catch()` or `.finally()`.

examples/long-running-process/index.js
```diff
@@ -27,26 +27,26 @@ function update(process) {
...
-            break;
-        case ProcessState.FAILED:
-            run_button.setAttribute("disabled", "");
-            run_button.innerHTML = "Start";
-            // Show the whole journal of this boot
-            showJournal(process.serviceName, "--boot");
-            break;
-        default:
-            throw new Error("unexpected process.state: " + process.state);
+    case ProcessState.INIT:
```
martinpitt, line 30: Really, `switch` and `case` have the same indentation level? That looks broken

martinpitt (changes requested):
Two questions, and of course it needs to fix tests. Thanks!

FWIW, I consider not running eslint during the commit hook a feature. It would be too annoying probably?

.eslintrc.json
allisonkarlitskaya: See the commit message about that one: I agree that for the case of chaining promise callbacks this is a bad idea, but there are other cases (like `.strip.split.join.etc()` type things) where this isn't so bad.

examples/long-running-process/index.js
allisonkarlitskaya: It's the same way in C...

allisonkarlitskaya:
> FWIW, I consider not running eslint during the commit hook a feature. It would be too annoying probably?

It takes a long time to run, but the performance is pretty reasonable when you only run it against changed files.  Even the "change all the things" commit here only takes a couple of seconds to check...

martinpitt (approved):
Cheers! I retried the tests, they looked unrelated (sigh)

PR #17192, opened 2022-03-29
test: Lower default CDP timeout to 15s

martinpitt:
One full minute is way too long for our UI to react for our UI
standards. In general, our UI should react very fast even on loaded CI
machines. The main exception are waiting for OS APIs that genuinely take
a lot of time (such as formatting a drive or joining a FreeIPA domain),
but these should (and often already do) have an explicit wait_timeout().

This also makes development of tests less annoying, and speeds up
known-failing tests (with naughty matching).

Lower the default timeout to 15s, drop the redundant low timeout in
TestConnection, and add explicit timeouts for legitimately long-running
steps. This makes it easy for us to grep the tests for things that are
slow, and worth looking into.

The biggest offender is the Services page, which is known to be dog
slow. Until we fix this properly, liberally and widely sprinkle
increased timeouts over the tests.

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -81,6 +81,11 @@ def redisService(image):
     return "redis"
 
 
+def applySettings(browser):
+    browser.click("#pcp-settings-modal button.pf-m-primary")
+    browser.wait_not_present("#pcp-settings-modal")
+
+
 def login(self):
     # HACK: Ubuntu and Debian need some time until metrics channel is available
     # Really no idea what it needs to wait for, so let's just try channel until it succeeds
@@ -443,8 +448,7 @@ class TestHistoryMetrics(MachineCase):
         b.wait_visible("#switch-pmlogger:not(:checked)")
         b.click("#switch-pmlogger")
         b.wait_visible("#switch-pmlogger:checked")
-        b.click("#pcp-settings-modal button.pf-m-primary")
-        b.wait_not_present("#pcp-settings-modal")
+        applySettings(b)
 
         m.execute("until systemctl is-active pmlogger; do sleep 1; done")
 
@@ -501,8 +505,7 @@ class TestHistoryMetrics(MachineCase):
         b.wait_visible("#switch-pmlogger:checked")
         b.click("#switch-pmlogger")
         b.wait_visible("#switch-pmlogger:not(:checked)")
-        b.click("#pcp-settings-modal button.pf-m-primary")
-        b.wait_not_present("#pcp-settings-modal")
+        applySettings(b)
 
         self.assertEqual(m.execute("systemctl is-active pmlogger || true").strip(), "inactive")
         self.assertEqual(m.execute("systemctl is-enabled pmlogger || true").strip(), "disabled")
@@ -513,8 +516,7 @@ class TestHistoryMetrics(MachineCase):
         b.wait_visible("#switch-pmlogger:not(:checked)")
         b.click("#switch-pmlogger")
         b.wait_visible("#switch-pmlogger:checked")
-        b.click("#pcp-settings-modal button.pf-m-primary")
-        b.wait_not_present("#pcp-settings-modal")
+        applySettings(b)
 
         m.execute("until systemctl is-active pmlogger; do sleep 1; done")
         self.assertEqual(m.execute("systemctl is-enabled pmlogger").strip(), "enabled")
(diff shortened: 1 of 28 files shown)


martinpitt:
Lowering to 5 s was clearly [too big a bite](https://github.com/cockpit-project/cockpit/commit/76e484a5f04d710e9469fb933f7faf2f2d965877), there were ~ 50 test failures. Re-running with 10s, let's do this in smaller steps. The final PR should double the value, of course.

martinpitt:
Very close! Just a [c8s packit](https://artifacts.dev.testing-farm.io/f3014611-daf5-4245-9ceb-22413b40f9ad/) failure of `TestServices.testBasic`, plus [three flakes](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17192-20220421-153830-8a424a43-fedora-35/log.html#246) where retries worked. Next round, full OS coverage.

martinpitt:
Still too many issues, let's move to 15 seconds and fix the tests for 10s.

martinpitt:
Wohoo, commit d4c3b795c31e0e2401bb1989efbce41b3fcaaa9c is fully green :partying_face: Most runs had zero retries, some had one or two. I added fixes for them.

martinpitt:
871e462d5e9cd9314255df673840d227ca777f92 is all green now; only 3 OSes had retries:

 - [debian-stable](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17192-20220424-063605-871e462d-debian-stable/log.html): infra failures/global test timeout
 - [f35/devel](https://logs.cockpit-project.org/logs/pull-17192-20220424-063605-871e462d-fedora-35-devel/log.html#144): This could be a too low timeout, I'll bump it
 - [ubuntu-2204](https://logs.cockpit-project.org/logs/pull-17192-20220424-063605-871e462d-ubuntu-2204/log.html): TestServices could be slow loading of Targets tab, I'll bump the timeout; NM-unmanaged failure is slow loading of the page, I'll also bump timeout

martinpitt:
"Interesting" (and super-annoying) [testPamAccess f35/TF failure](https://artifacts.dev.testing-farm.io/934dc629-eab7-4806-9d32-17524a2f8d71/). Apparently after logout(), the login page takes more than 10s to appear? The [screenshot](https://artifacts.dev.testing-farm.io/934dc629-eab7-4806-9d32-17524a2f8d71/work-alljtUTWe/plans/all/execute/data/test/verify/data/TestLogin-testPamAccess-fedora-35-localhost-22-FAIL.png) is just fine though, so I rather suspect some other reason. Indeed in the past we already had some trouble with `login_and_go()` immediately being followed by `logout()`, and worked around that by doing something in between. I added a separate commit for that.

Plus one more related timing flake.

martinpitt:
This looks really good now. A single related flake left on the metrics page, bumping timeout from 30 to 60s to match what we have now.

martinpitt:
OK, this now ran like 15 times, and I'm happy with the stability now. The remaining few (auto-retried) flakes have other causes.

@mvollmer , as our "do global test changes" specialist, do you have an opinion about that?

@marusak, I'd appreciate your opinion as well.

marusak (approved):
Looks good to me, thanks!
I am a bit worried about how many new flakes will pop up, but in a long run this is a good direction as it does not make sense to wait for too long - if something takes really long, then we should usually enforce that there is some spinner, disabled submit button and wait for action to finish up...

test/verify/check-metrics
```diff
@@ -201,6 +207,8 @@ class TestHistoryMetrics(MachineCase):
         b = self.browser
         m = self.machine
 
+        b.wait_timeout(60)
```
mvollmer, line 210: Oh, does this work just like that without `with`?

martinpitt:
> I am a bit worried about how many new flakes will pop up,

Me too -- I had all green on first try in the most recent 5+ force pushes, i.e. I always went through all the green ones and checked all the auto-retried tests. I'm sure there's a few left, but they don't block PR landing any more. I'd say it's certainly well within the noise range of flakes that happen for other reasons (such as fork() failure or machines failing to boot, or Firefox failing on "event context destroyed" and such)

test/verify/check-metrics
martinpitt: Yes, `with` just restores the timeout at the end of the context. Like this it's effectively equivalent to `b.cdp.timeout = 60`, but looks a bit less hacky/rude.

mvollmer:
My usual reply to decreasing thje timeout was "Doesn't make tests faster, just more flaky, and can be changed locally very easily if annoying", but I guess we have so many known failures now that decreasing the timeout does indeed make CI runs faster...

So, yeah, looks good to me. :-)

mvollmer (approved):

martinpitt:
@mvollmer : Making naughtied tests faster is not my primary goal (although it's a nice side effect). I primarily want to identify and spot places where the UI is too slow, and make these places greppable. It's most useful for developing new code or updating PF and such, to spot UI operations which are slow. Of course we have a lot of old stuff which needs fixing (*cough* services..)

PR #17193, opened 2022-03-29
systemd: Parse /proc/cpuinfo on s390x

martinpitt:
On this architecture, cpuinfo does not start processor blocks with
`processor:`, but it looks quite different. It also does not have a
per-cpu name, just a "vendor_id" in the header. Adjust the parsing
accordingly.

Test this with a real-life cpuinfo from a bug report.

https://bugzilla.redhat.com/show_bug.cgi?id=2069500
https://bugzilla.redhat.com/show_bug.cgi?id=2069552

### DIFF
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -512,6 +512,36 @@ revision\t: 2.3 (pvr 004e 1203)
         b.enter_page("/system")
         b.wait_in_text("#system-usage-cpu-progress + td", "of 2 CPUs")
 
+        # /proc/cpuinfo on s390x (reduced)
+        m.write("/tmp/cpuinfo", """vendor_id       : IBM/S390
+# processors    : 2
+bogomips per cpu: 3241.00
+max thread id   : 0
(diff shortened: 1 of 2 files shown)


jelly:
While investigating ARM inconsistencies I was wondering if we should move to using `lscpu` which is part of `util-linux`. That should be a fairly common distro dependency.

test/verify/check-system-info
```diff
@@ -512,6 +512,36 @@ revision\t: 2.3 (pvr 004e 1203)
         b.enter_page("/system")
         b.wait_in_text("#system-usage-cpu-progress + td", "of 2 CPUs")
 
+        # /proc/cpuinfo on s390x (reduced)
+        m.write("/tmp/cpuinfo", """vendor_id       : IBM/S390
+# processors    : 2
```
jelly, line 517: and here I thought the pound symbol was a typo ;)

jelly (approved):

test/verify/check-system-info
martinpitt: Hehe, no -- I copied this verbatim from the `cat /proc/cpuinfo` output from the bug report.

PR #17196, opened 2022-03-30
bridge: Default "init-superuser" to "none"

### LINKED ISSUE #17169: remote host tries sudo three times

mvollmer:
This makes SSH logins use the same default as HTTP logins.  Running
"sudo" without explicit request from the user is problematic.

Fixes #17169

### DIFF
--- a/src/bridge/cockpitpeer.c
+++ b/src/bridge/cockpitpeer.c
@@ -319,7 +319,7 @@ on_other_control (CockpitTransport *transport,
 
               if (explicit_superuser_capability)
                 {
-                  const gchar *superuser = "any";
+                  const gchar *superuser = "none";
                   if (self->init_superuser && *self->init_superuser)
                     superuser = self->init_superuser;
 
--- a/test/verify/check-superuser
+++ b/test/verify/check-superuser
@@ -403,6 +403,12 @@ class TestSuperuserDashboard(MachineCase):
 
         # The superuser indicator in the Shell should apply to machine2
 
+        b.check_superuser_indicator("Limited access")
+        b.open_superuser_dialog()
+        b.wait_in_text(".pf-c-modal-box:contains('Switch to administrative access')", "Password for admin:")
+        b.set_input_text(".pf-c-modal-box:contains('Switch to administrative access') input[type=password]", "foobar")
+        b.click(".pf-c-modal-box button:contains('Authenticate')")
+        b.wait_not_present(".pf-c-modal-box:contains('Switch to administrative access')")
         b.check_superuser_indicator("Administrative access")
         b.go("/@10.111.113.2/playground/test")
         b.enter_page("/playground/test", host="10.111.113.2")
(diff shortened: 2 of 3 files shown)


mvollmer:
I'll have to port some test changes from #17177 I guess.

martinpitt (approved):
Cheers!

PR #17199, opened 2022-03-30
services: Fix status wrapping and optimize layout

garrett:
This started as a task to fix the status messages wrapping and invading the space of the name and details for services.

But working on that was frustrating as performance was extremely poor, especially on mobile sizes. PatternFly adjusts the tables from tables to nested grids and we even had a nested flex... so laying out everything is incredibly intertwined and CPU expensive, especially on lower power devices (such as most phones) or when doing development work. I'm talking minutes to re-render resizes or poke at the DOM when developer tools are open, with nothing else happening on my system.

As we're already using tables a bit differently than intended for PatternFly, instead of rewriting the JSX, I found ways to optimize the display by changing some of the more advanced layout into standard block layout (which is much easier to for the browser figure out). Since this is specific to our own use of the widget, I don't think it can cleanly apply to upstream PatternFly, so this is a local (to the page in Cockpit) workaround, not a PatternFly-level override.

We _could_ see if adjusting at least a level or two of layout might be an optimization that could work for PatternFly, but I'm not sure if it could be.

As the next version of Safari will have subgrid support and it's also supposed to land in Chrome 101, this means we could consider subgrid. Firefox has had it for years. I started by trying to see if introducing subgrid could actually give us performance wins. It seemed like it might, but I couldn't get it quite how I wanted... and realized I should try to make things simpler rather than more complex, hence dropping to block display where possible.

This isn't a finalized solution, but it's getting late here and I wanted to share what I have so far. It still needs work. And I would like to see how much we could push back up into PatternFly (although I don't think that much, as we are doing things differently).

The performance problems happen when there are a few hundred rows. (I think there are around 300?) Each row has a dozen or so elements, some nested, with different layouts, so it's something like a few thousand intertwined smaller layouts to figure out to make the final overall layout. Browsers are fast, but that does add up.

I still have some ideas to simplify this and I do intend on removing the `!important`s and commented out code before I'm done. (And probably add in more / modify comments to explain what's happening.) This is a work in progress, after all.

### DIFF
--- a/pkg/systemd/services/services.scss
+++ b/pkg/systemd/services/services.scss
@@ -10,12 +10,6 @@
     border-bottom: 1px solid var(--pf-global--BorderColor--100);
 }
 
-// Magic pixie dust to make the list a bit faster and more furious
-.services-list tbody {
-    // Skip a lot of computations
-    contain: content;
-}
-
 .service-unit-failed {
     background-color: var(--ct-color-list-critical-bg);
 
@@ -30,7 +24,6 @@
 
     &-failed {
         color: red;
-        white-space: break-spaces;
     }
 
     &-flex-container {
@@ -94,4 +87,46 @@
 }
 .pf-m-grid-md.pf-c-table [data-label] {
     display: revert;
+}
+
+// Magic pixie dust to make the list a bit faster and more furious
+.services-list {
+    // Prevent reflows; only applies to desktop (mobile changes display)
+    // table-layout: fixed;
+
+    tr {
+        // Skip a lot of computations
+        contain: content;
+        // Only render what's needed
+        // content-visibility: auto;
+    }
+}
+
+// Optimize layout, mainly using block, so we don't nest multiple levels of
+// grids and flex on thousands of elements.
+.services-list {
+    @media screen and (max-width: $pf-global--breakpoint--md) {
+        display: block !important;
+
+        tbody {
+            display: block !important;
+        }
+
+        th, td {
+            // display: grid;
+            display: block !important;
+        }
+    }
+
+    @media screen and (max-width: $pf-global--breakpoint--sm) {
+        td {
+            > .service-unit-status-flex-container {
+                // display: grid !important;
+                display: flex !important;
+                flex-direction: column-reverse;
+                align-items: end;
+                gap: var(--pf-global--spacer--sm);
+            }
+        }
+    }
 }
--- a/pkg/systemd/services/services.scss
+++ b/pkg/systemd/services/services.scss
@@ -91,40 +91,51 @@
 
 // Magic pixie dust to make the list a bit faster and more furious
 .services-list {
-    // Prevent reflows; only applies to desktop (mobile changes display)
-    // table-layout: fixed;
-
     tr {
         // Skip a lot of computations
         contain: content;
-        // Only render what's needed
-        // content-visibility: auto;
     }
 }
 
-// Optimize layout, mainly using block, so we don't nest multiple levels of
-// grids and flex on thousands of elements.
-.services-list {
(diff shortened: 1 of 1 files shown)


garrett:
Here are screenshots that demonstrate the original problem that I was attempting to fix:

Screen Shot 2022-03-30 at 18 32 06

Screen Shot 2022-03-30 at 18 31 57

...and then the poor performance caused me to try to figure out what was going on and to find solutions to fix it.

(We could also consider _other_ PF widgets too. Perhaps.)

pkg/systemd/services/services.scss
```diff
@@ -95,3 +88,45 @@
 .pf-m-grid-md.pf-c-table [data-label] {
     display: revert;
 }
+
+// Magic pixie dust to make the list a bit faster and more furious
+.services-list {
+    // Prevent reflows; only applies to desktop (mobile changes display)
+    // table-layout: fixed;
```
KKoukiou, line 95: Let's not keep in comments unused CSS

pkg/systemd/services/services.scss
```diff
@@ -95,3 +88,45 @@
...
+
+    tr {
+        // Skip a lot of computations
+        contain: content;
```
KKoukiou, line 99: Is this something that can be applied directly upstream to https://github.com/patternfly/patternfly/blob/main/src/patternfly/components/Table/table.scss ? 

I read little bit about these properties [here](https://css-tricks.com/almanac/properties/c/contain/) and they seem something generic enough to be applied upstream.

KKoukiou (commented):
Thanks for working on this. Honestly however, this is a quite complicated piece of custom CSS, which the team won't be able to maintain easily over time. I am obviously happy to add it it, but only in the case it has direct benefits for the production code, that are explained thoroughly so that we know why this actually exist.

So, please adjust 'optimize etc etc' comments to more specific reproducible problems,

pkg/systemd/services/services.scss
garrett: I'm not sure, but it's possible... and it's a good idea; telling the browser to take some shortcuts when it can is always beneficial.

It can cause some rendering issues in some cases (depending on how others are using the lists), but at least _some_ setting of `contain` might be able to be a default and then could be set to something else for more or less optimization, depending on what's needed.

The primary win for this would be the tables, but any kinds of list or repeated and/or complex layout can see an improvement. (But it's usually negligible until there's a _lot_, so we probably shouldn't try to get this sprinkled everywhere.)

pkg/systemd/services/services.scss
garrett: Yep, I mentioned removing this above. :wink: This PR was to save my work off of my computer and show something that is working even if it's in a rough state.

I wanted to re-visit this line of CSS, as it's a huge performance win, but I'll probably just remove it as we're not really using tables as tables here (where everything needs to line up).

In the long run, we should probably use another widget for the lists rather than our own listing widget which uses composable tables.

 (It could be something like a grid (PF Grid even) which has a flex inside (PF Flex even) and that's it. It'd simplify the DOM and layout code quite a lot. But it also would mean an overhaul to the JSX versus a few lines of CSS to override the display.)

garrett:
I've pushed some fixups for mobile. Desktop mode still needs some work, but it wasn't as bad off anyway (and usually desktops have beefier CPUs too). Still, it would be nice to improve desktop as well.

I haven't researched to see if any of these changes can be pushed upstream. A lot of these changes are definitely related to the services page.

It's still a WIP; I need to work on this a bit more next week to finish it up.

KKoukiou:
Closed in favor of https://github.com/cockpit-project/cockpit/pull/18738

PR #17200, opened 2022-03-30
test: add firefox-developer to firefox binaries

jelly:
On Arch Linux there is a seperate package for the firefox developer
edition which can be installed alongside firefox. Add this as useable
firefox binary.


martinpitt (approved):
Thanks! Triggering required fcos test.

PR #17202, opened 2022-03-31
metrics: Consistent page title and Overview link

martinpitt:
We got a report about a confusing mental disconnection between the Usage
card's link ("View details and history") and the metrics page title
("Performance Metrics"). There is no overlap, and additionally
"Performance profile" on the Overview refers to tuned (which is
completely different).

Rename both to "metrics and history" for consistency, and avoiding the
"performance" overlap. Also move to sentence case for the breadcrumb, to
comply to [1] and be consistent with e.g. the "Kernel crash dump" page
tile.

[1] https://www.patternfly.org/v4/ux-writing/capitalization/#capitalization-across-patternfly

---

[pixel diff link](https://github.com/cockpit-project/pixel-test-reference/compare/fa3e8b66637ce0815d01915779b5906b98245f4d..54401adf76a823142cd2a1061fcb2d545342051b)

### DIFF
--- a/doc/guide/feature-pcp.xml
+++ b/doc/guide/feature-pcp.xml
@@ -7,7 +7,7 @@
   <para>If available, Cockpit uses the
     <ulink url="https://pcp.io/">Performance Co-Pilot</ulink> framework to
     gather metrics data about the system. This data is used to display the history
-    graphs on the "Performance Metrics" page.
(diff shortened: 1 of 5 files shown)


garrett (approved):
This looks great to me! Thanks for driving this change; it's definitely a nice improvement.

PR #17204, opened 2022-03-31
client: install metainfo file

allisonkarlitskaya:
Create a tool to turn the point form notes in our tags into a proper set
of `<release/>` tags for metainfo and fill this information in when
building a tarball.

Install the metainfo if building with --enable-client.

This will simplify the process of releasing the flatpak.

### DIFF
--- a/.github/workflows/flatpak-test.yml
+++ b/.github/workflows/flatpak-test.yml
@@ -24,10 +24,7 @@ jobs:
           fetch-depth: 0
 
       - name: Build and install flatpak
-        run: sh -x containers/flatpak/install --user --install-deps-from=flathub
-
-      - name: Validate metainfo
-        run: appstream-util validate org.cockpit_project.CockpitClient.metainfo.xml
+        run: ELEMENT_TREE_NO_INDENT=1 sh -x containers/flatpak/install --user --install-deps-from=flathub
 
       - name: Smoke-test the installed flatpak
         run: |
--- a/containers/flatpak/prepare
+++ b/containers/flatpak/prepare
@@ -18,30 +18,10 @@
 #
 # If the FLATPAK_ID environment variable is set, it overrides the
 # default value of org.cockpit_project.CockpitClient.
-#
-# If the PREVIOUS_RELEASES environment variable is set, it is written verbatim
-# into the metainfo xml after the current release info.
 
 set -eu
 
 FLATPAK_ID="${FLATPAK_ID:-org.cockpit_project.CockpitClient}"
-
-subst() {
-    ext="$1"
-
-    sed \
-        -e "s|@FLATPAK_ID@|${FLATPAK_ID}|" \
-        -e "s|@VERSION@|${VERSION}|" \
-        -e "s|@TODAY@|${TODAY}|" \
-        -e "s|@FLATPAK_BRANCH@|${FLATPAK_BRANCH}|" \
-        -e "s|@RELEASE_TYPE@|${RELEASE_TYPE}|" \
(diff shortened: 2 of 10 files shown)


containers/flatpak/prepare
```diff
@@ -78,8 +58,14 @@ prepare() {
 
     TODAY="$(date +%F)"
 
-    subst metainfo.xml
-    subst yml
+    sed \
```
martinpitt, line 61: I kinda liked the separate `subst()` function -- shortens the function a bit, and we might need it for other files, too. But bikeshedding, and very weak opinion.

src/client/add-metainfo-releases
```diff
@@ -0,0 +1,60 @@
...
+    tree = ET.parse(args.xml)
+    releases = tree.find('releases')
+
+    # Only update the file if it doesn't already contain releases.
+    if not releases.find('release'):
+        if 'g' in args.version:
+            today = time.strftime('%Y-%m-%d')
+            releases.append(element('release', version=args.version, date=today, type='development'))
+        releases.extend(format_release(r) for r in get_releases())
+        ET.indent(releases, space="  ", level=1)
```
martinpitt, line 55: This is not present in Ubuntu 20.04 yet, see [failed workflow](https://github.com/cockpit-project/cockpit/runs/5773904226?check_suite_focus=true). As 20.04 LTS is a very popular platform for all kinds of infrastructure, let's find an API that works there. I'm sure ElementTree has *some* pretty-printer?

src/client/add-metainfo-releases
```diff
@@ -0,0 +1,60 @@
...
+    url = element('url', f'https://cockpit-project.org/blog/cockpit-{version}.html', type='details')
+    ul = element('ul', children=points_from_body(body))
+    description = element('description', children=(ul,))
+    return element('release', version=version, date=date, children=(url, description))
+
+
+def get_releases():
+    format = '%(if)%(taggerdate)%(then)%(refname:short)|%(taggerdate:short)|%(contents:body)%(end)'
+    output = subprocess.check_output(['git', 'for-each-ref', '--shell', '--sort=-taggerdate',
+                                      '--contains', '258', '--format', format, 'refs/tags'], text=True)
```
martinpitt, line 36: What's the magic "258" here? Shouldn't that be the `--version` value?

martinpitt (changes requested):
Nice!

src/client/add-metainfo-releases
allisonkarlitskaya: This is when we started shipping the flatpak.

allisonkarlitskaya: Added a comment about it.

martinpitt: Oh, I see -- so for each release, we have to regenerate the entire release history. I forgot that, but makes sense now.

.github/workflows/flatpak-test.yml
```diff
@@ -24,10 +24,7 @@ jobs:
           fetch-depth: 0
 
       - name: Build and install flatpak
-        run: sh -x containers/flatpak/install --user --install-deps-from=flathub
-
-      - name: Validate metainfo
-        run: appstream-util validate org.cockpit_project.CockpitClient.metainfo.xml
+        run: ELEMENT_TREE_NO_INDENT=1 sh -x containers/flatpak/install --user --install-deps-from=flathub
```
martinpitt, line 27: That could have used a comment, but we can always get it from git blame.

martinpitt (approved):
Cheers!

PR #17207, opened 2022-04-01
lib: Provide cockpit.js as ES6 module

### LINKED ISSUE #17203: Make cockpit.js importable as an ES6 module
Page: base1

OpenSUSE is [moving to a Cockpit based installer](https://yast.opensuse.org/blog/2022-03-31/d-installer-first-public-release), and they currently [patch cockpit.js](https://github.com/yast/d-installer/blob/master/web/src/lib/cockpit.js#L4522) to make it importable as an ES6 module, instead of importing via `<script>`.

That's actually an interesting idea -- we could bundle cockpit.js into projects like cockpit-podman etc. (or at least should be able to) and import it similar to pkg/lib. It would save one more separate resource download.

We still need to keep `<script>` import support -- it's API, and we very much want to keep the possibility of [quick and dirty cockpit pages](https://cockpit-project.org/blog/creating-plugins-for-the-cockpit-user-interface.html) without mandating webpack and a build system.

martinpitt:
Move cockpit.js' code to pkg/lib and provide its object as default
export. With that, external projects like d-installer [1] can import it
through webpack/ES6 instead of `<script>`.

Only keep the cockpit.css build in base1/cockpit.js, as that's not 
relevant for using it as a module.

It would also be elegant to keep the `window.cockpit` global definition
in base1/cockpit.js, but that would break translations: Importing
`po.js` through `<script>` must have access to the `cockpit` global
already, and po.js *has* to be loaded that way due to the dynamic locale
specific path lookup.

Fixes #17203

[1] https://github.com/yast/d-installer

---

This also converts kdump to use cockpit.js as ES6 module, as a PoC that it works.

 - [x] get ack from @imobachgs that this makes sense for d-installer


martinpitt:
@imobachgs: Can you please have a look at this and see whether this makes sense to you?

This is a bit tricky wrt. translations -- I see that in your [main html](https://github.com/yast/d-installer/blob/master/web/index.html) you don't include translations (`po.js`) at all yet, so I figure you don't have set up an i18n infrastructure yet. I was careful here to *not* break that, but the `<script>` order matters.

martinpitt:
@larskarlitski : AFAIR you had some plan/PoC to do this years ago. Do you still remember why it wasn't done after all? Did that break something subtle?

martinpitt:
Oui, ça marche -- other than the unit test. Let's fix that as well.

larskarlitski:
> @larskarlitski : AFAIR you had some plan/PoC to do this years ago. Do you still remember why it wasn't done after all? Did that break something subtle?

I think we only didn't do it because nobody had the time back then. I still think it's a great idea.

imobachgs:
Hi @martinpitt,

Yes, it makes sense to me. Actually, I gave it a try in D-Installer and it seems to work perfectly. About translations, well, they are on our roadmap but not yet. But we plan to start working on them sooner than later. 

Thanks a lot!

martinpitt:
@mvollmer , @marusak : Friendly review ping? *bats eyelashes*

PR #17208, opened 2022-04-01
package.json: Update PF to prerelease

marusak:


martinpitt:
Yay, the [time dialog looks ok again](https://logs.cockpit-project.org/logs/pull-17208-20220401-074357-10d616b6-debian-testing/TestSystemInfo-testTime-debian-testing-127.0.0.2-3001-FAIL.png). So I figure this now merely needs the fixes from #16967 to adjust to the new API?

marusak:
I fixed two tests, but there are also pixel differnces, some of them seem wrong:

- spacing in the menu: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17208-20220401-104228-820f7b5b-fedora-35/log.html

- differences in buttons in top panel: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-17208-20220401-104228-820f7b5b-fedora-35/pixeldiff.html#TestSuperuser-testBasic-topnav-privileged-medium

garrett:
Yes, these two differences are clearly wrong. I wonder if it's because we're doing some special stuff in the shell which might break some PatternFly upstream assumptions, or if there's some bug in PatternFly's prerelease.

marusak:
@garrett I might need some helping hand. I cannot figure out why this overflows :man_shrugging: Do you have some ideas?

Screenshot from 2022-04-04 20-26-56

garrett:
@marusak: I'm looking into this now.

garrett:
I found the problem, figured out a fix, and realized there are a few minor regressions that I should address. I'll push something up soon.

garrett:
Oooooof. The hover states on the masthead are broken too.

I noticed this when PF showed off their new heading (in the community call today) that looks like this, which we'll want to adapt to, I guess:

image

image

I'll push what I have now, but will have to fix the hover state tomorrow... either to be like what we have now or adapt to the new look (which isn't a huge deal, but it is different).

marusak:
Thank you @garrett ! There are now pixel changes, but it is just that items are spaced a bit differently but nothing is broken as before :tada: Can you please check current changes and let me know that you are also fine with these changes and that I can update the reference images? Thanks!

https://logs.cockpit-project.org/logs/pull-17208-20220406-064826-e63ae384-fedora-35/log.html

garrett:
Pixel tests look fine.

I did realign the search area so the text inside alignes with the rest of the text in desktop mode. And in mobile, as the menu can stretch across multiple columns, I left it aligning to the block instead.

I guess there were some font updates too, as there are some offsets with fonts, especially the monospaced variant.

And I suppose PF changed the time picker, perhaps? And PF slightly adjusted stepper controls?

---

However...

We're still lacking hovers. I've been debugging this. At first, I didn't realize that your workaround to hide borders was causing nothing at all to happen on hover. :facepalm: (I'm out of coffee and it's getting late now.) Then, when working around your workaround (in the debugging process), I realized that the fix isn't quite as simple.

What are we doing that's so radically different from upstream PF4? Perhaps we should rewrite this part in this PR? Or should I just stack on hacks in the CSS for now?

We can't merge with no visual feedback whatsoever. This is the one thing that's blocking this PR as-is... (I think.)

garrett:
Scrolling back up at the screenshots of the current UI in PatternFly, I think if we just stretch it, it'd probably fix the problem. It's too late to take this on tonight, but I think I have a way to fix it without too much effort tomorrow. I hope.

garrett:
I filed this masthead select bug upstream @

- https://github.com/patternfly/patternfly/issues/4775

garrett:
For reference, here's what the header looks like now for me:

image

There was a gap between the selects, so I removed that and even nudged the two together to combine the line as to have a shared 1px border instead of a 2px border (1px x 2).

PatternFly's demos of this only ever have one select and their help is a link that goes offsite instead of a menu, hence only using an icon in their case and us doing a bit more.

garrett:
I tried using an InputGroup instead of my custom CSS and got this:

image

Whoops. I guess that won't work.

marusak:
Rebased and updated pixels. Looking for a green run :crossed_fingers:

marusak:
green \o/

martinpitt (approved):
Thanks! I reviewed the pixel diffs in detail, and the remaining differences are fine.

Great work @marusak and @garrett !

PR #17209, opened 2022-04-01
lib: shell: remove unused variables.scss

jelly:
Everyone using listing-ct-* has been ported to the React listing table.


martinpitt (approved):
Nice cleanup! Thanks @KKoukiou for spotting, too!

garrett:
Doesn't this need variables.scss?

https://github.com/jelly/cockpit/blob/b8f978a6c031c56e1857a06fbb5d099f1d4cb828/pkg/lib/patternfly/patternfly-overrides.scss#L529

It's needed to bump up PF3 font sizes to PF4 sizes. If we're using PF3 anywhere, it's still needed, therefore, we'd still need to import variables for this.

(It could be hardcoded, but if PF4 could possibly change global font sizes. However, I doubt they'll change it anytime soon. And the context here is pretty legacy.)

KKoukiou:
@garrett we don't use Patternfly 3 in cockpit or any of the plugins the team maintains since a long time.

KKoukiou:
@garrett plus this specific variable comes from https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/patternfly/patternfly-overrides-variables.scss#L80 , which is not deleted.

PR #17210, opened 2022-04-01
test: remove already handled restart messages

jelly:
In testlib reload(), logout() already calls
allow_restart_journal_messages().


martinpitt (approved):
Dankje wel!

PR #17214, opened 2022-04-04
kdump: Show current journal when (re)starting service fails

martinpitt:
This is often due to some mis-configuration, such as an NFS remote
directory not existing on the server. There is currently no better
machine-readable API for showing the root cause of the error, so
show the journal of the recently failed kdump.service run.

Use the same `<CodeBlockCode>` formatting and size as the realmd
overview card.

If the dialog is able to read the journal, then don't show the original
exception. That often reads "systemd job RestartUnit
["kdump.service","replace"] failed with result failed" which is not very
useful. Only show that if reading the journal failed (which "should not
happen™").

Assert the error formatting more precisely in the integration tests.

https://bugzilla.redhat.com/show_bug.cgi?id=2062297

----
Wide window, NFS error:
image

Narrow window, same NFS error:
image

The error message is still a bit confusing, I reported this in https://bugzilla.redhat.com/show_bug.cgi?id=2071529. After that gets fixed, it'll be okay. But showing the journal is still better than just "it failed".

Another example failure when having wrong SSH parameters:
image

### DIFF
--- a/pkg/kdump/kdump-view.jsx
+++ b/pkg/kdump/kdump-view.jsx
@@ -24,6 +24,7 @@ import React from "react";
 import {
     Button, Checkbox,
     Card, CardBody,
+    CodeBlockCode,
     Flex,
     Form, FormGroup, FormSection,
     FormSelect, FormSelectOption,
@@ -281,7 +282,18 @@ export class KdumpPage extends React.Component {
 
     handleSaveClick() {
         return this.props.onSaveSettings(this.state.dialogSettings)
-                .catch(error => Promise.reject(cockpit.format(_("Unable to save settings: $0"), String(error))));
+                .catch(error => {
+                    if (error.details) {
+                        // avoid bad summary like "systemd job RestartUnit ["kdump.service","replace"] failed with result failed"
+                        // if we have a more concrete journal
+                        error.message = _("Unable to save settings");
+                        error.details = <CodeBlockCode>{ error.details }</CodeBlockCode>;
+                    } else {
+                        // without a journal, show the error as-is
+                        error = new Error(cockpit.format(_("Unable to save settings: $0"), String(error)));
+                    }
+                    return Promise.reject(error);
+                });
     }
 
     handleTestSettingsClick(e) {
--- a/pkg/kdump/kdump-view.jsx
+++ b/pkg/kdump/kdump-view.jsx
@@ -280,11 +280,8 @@ export class KdumpPage extends React.Component {
     }
 
     handleSaveClick() {
-        return new Promise((resolve, reject) => {
-            this.props.onSaveSettings(this.state.dialogSettings)
-                    .then(resolve)
-                    .catch(error => reject(cockpit.format(_("Unable to save settings: $0"), String(error))));
-        });
+        return this.props.onSaveSettings(this.state.dialogSettings)
+                .catch(error => Promise.reject(cockpit.format(_("Unable to save settings: $0"), String(error))));
     }
 
     handleTestSettingsClick(e) {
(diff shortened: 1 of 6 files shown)


pkg/kdump/kdump-view.jsx
```diff
@@ -281,7 +282,17 @@ export class KdumpPage extends React.Component {
 
     handleSaveClick() {
         return this.props.onSaveSettings(this.state.dialogSettings)
-                .catch(error => Promise.reject(cockpit.format(_("Unable to save settings: $0"), String(error))));
+                .catch(error => {
+                    if (error.message) {
+                        // don't destroy error.details, stack trace, etc.
+                        error.message = cockpit.format(_("Unable to save settings: $0"), error.message);
+                        if (error.details)
+                            error.details = <CodeBlockCode>{ error.details }</CodeBlockCode>;
```
martinpitt, line 290: FTR: I first tried to add a `className="kdump-error-details"` here, but that completely destroy's PF's own CSS. This component isn't made for custom CSS classes, that's why I wrote the CSS in terms of its default `.pf-c-code-block__pre` class.

martinpitt:
Argh, the [c8s repo instability](https://copr-be.cloud.fedoraproject.org/results/packit/cockpit-project-cockpit-17214/centos-stream-8-x86_64/03942181-cockpit/builder-live.log.gz) is a bit unnerving.. Retrying, I just saw it fail the same way in a c-machines PR and now it succeeds.

garrett (changes requested):
Inline alerts related to form submission are supposed to be at the top:

https://www.patternfly.org/v4/components/form/design-guidelines#error-validation-on-submission

That is a lot of text, so the details are supposed to be hidden by default and expandable. PatternFly's implementation looks like this:

https://www.patternfly.org/v4/components/alert#expandable

We currently don't use PF's expandable alerts (AFAIK), as they're relatively new. I think we have used expandable sections in alerts before for this:

https://www.patternfly.org/v4/components/expandable-section

We should probably start to use expandable alerts, however. The details link could also expand the ">" part, to make it more discoverable.

martinpitt:
With an expandable alert it would look like this:

image

and once expanded:

image

However, I'm not very happy about this. The "service failed" error doesn't help with the "what do I need to fix", so unless the user is making the same error over and over and can guess it, it will have to be expanded. This is the original bug which this is aiming to fix.

However, the `<Alert isExpandable` is not a controlled component -- it means that we can neither expand it by default (which I'd like to do) nor can we add a "View Details.." link which could open it, as the expandedness is simply not controllable by code. This can probably be hacked somehow, with rebuilding the Alert ourselves, or putting an [ExpandableSection](https://www.patternfly.org/v4/components/expandable-section) inside the alert. But first I'd like to negotiate a bit whether we should really hide this by default -- I rather strongly believe that we shouldn't.

> Inline alerts related to form submission are supposed to be at the top.

OK - this is a widely used component (cockpit-components-dialog.jsx), so if I move it here, it'll also move on all the other pages. I can do this first, and block on that PR.

martinpitt:
@garrett : I sent #17221 to move dialog errors to the top. IMHO we should do this in one go to remain consistent, and decouple that from this PR. So what remains here is to discuss the expander stuff above. Thanks!

martinpitt:
@garrett: I reworked this a bit. I cleaned up the systemd unit noise from the error message, and also dropped that silly "RestartUnit failed with result failed" part of the title. Screenshots updated, they look a lot cleaner now. I also showed a different error for SSH (which can also fail for any number of reasons). I am convinced that we should not hide these messages by default.

garrett:
> I reworked this a bit. I cleaned up the systemd unit noise from the error message, and also dropped that silly "RestartUnit failed with result failed" part of the title. Screenshots updated, they look a lot cleaner now. I also showed a different error for SSH (which can also fail for any number of reasons). I am convinced that we should not hide these messages by default.

OK, better! I'm looking forward to the move to the top and the expanded version.

Thanks for adjusting this all!

martinpitt:
@marusak , @jelly : Friendly review ping? Thanks!

marusak (approved):
Nice, thank you!

PR #17216, opened 2022-04-04
tests: allow having the test/reference directory in another subdir

KKoukiou:
Useful for external projects, like anaconda-webui, where the cockpit
code is not in the root of the repo.


martinpitt (approved):
Cheers!

PR #17217, opened 2022-04-05
test: Don't expect lecture on debian-testing

marusak:
Since version of `sudo` 1.9.10-1 lecture is not enabled by default.
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006273


martinpitt (approved):
Thanks! Next time this happens, we should make this more generic. But let's unblock the image refresh quickly.

PR #17219, opened 2022-04-05
test: Work around race in Udisks2 LVM2 scanning

mvollmer:


mvollmer:
This doesn't work.  The "wrong" update is triggered somewhere during creation vgroup0, and not while deleting TEST1.

PR #17220, opened 2022-04-05
systemd: Show timer creation errors

martinpitt:
Fix the setOpen() call to actually happen *after* `create_timer()`
promise finishes and succeeds. Otherwise the dialog closes immediately.

Fix the promise chaining in create_timer_file() to actually return
errors from creating the .service file, by dropping the parallel catch
handler (which was just logging). Instead, chain it up to creating the
.timer file as well.

Fix chaining of the Reload call.

There is not a lot of opportunity for this to actually fail, but it can
happen on inaccessible or even non-existing directories. Do that in the
test.


martinpitt:
image

I noticed that while moving all dialog errors to the top, see https://github.com/cockpit-project/cockpit/pull/17214#pullrequestreview-930307611

jelly (approved):

PR #17222, opened 2022-04-05
Install kpatches only

jelly:
A new option to only install kpatches.

---

# Software Updates: Install kpatches only

It is now possible to only install kpatch updates. This is useful at times when a reboot or service restarts are not acceptable, to at least fix kernel vulnerabilities.

image

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -74,6 +74,12 @@ const PK_STATUS_STRINGS = {};
 const PK_STATUS_LOG_STRINGS = {};
 const packageSummaries = {};
 
+const UPDATES = {
+    ALL: 0,
+    SECURITY: 1,
+    KPATCHES: 2,
+};
+
 function init() {
     STATE_HEADINGS.loading = _("Loading available updates, please wait...");
     STATE_HEADINGS.locked = _("Some other program is currently using the package manager, please wait...");
@@ -156,6 +162,18 @@ function count_security_updates(updates) {
         if (updates[u].severity === PK.Enum.INFO_SECURITY)
             ++num_security;
     return num_security;
+}
+
+function isKpatchPackage(name) {
+    return name.startsWith("kpatch-patch");
+}
+
+function count_kpatch_updates(updates) {
+    let num_kpatches = 0;
+    for (const u in updates)
+        if (isKpatchPackage(updates[u].name))
+            ++num_kpatches;
+    return num_kpatches;
 }
 
 function find_highest_severity(updates) {
@@ -285,7 +303,7 @@ function updateItem(info, pkgNames, key) {
     if (pkgList.length > 4)
         pkgsTruncated.push(<span key="more">…</span>);
 
-    if (pkgNames.some(p => p.name.startsWith("kpatch-patch")))
+    if (pkgNames.some(pkg => isKpatchPackage(pkg.name)))
         pkgsTruncated.push(
             <LabelGroup key={`${key}-kpatches-labelgroup`} className="kpatches-labelgroup">
                 {" "}<Badge color="blue" variant="filled">{_("patches")}</Badge>
@@ -907,6 +925,7 @@ class CardsPage extends React.Component {
                             </FlexItem>
                         </Flex>}
                     {this.props.applySecurity}
+                    {this.props.applyKpatches}
                     {this.props.applyAll}
                 </div>),
                 containsList: true,
@@ -1275,10 +1294,13 @@ class OsUpdates extends React.Component {
                 });
     }
 
-    applyUpdates(securityOnly) {
+    applyUpdates(type) {
         let ids = Object.keys(this.state.updates);
-        if (securityOnly)
+        if (type === UPDATES.SECURITY)
             ids = ids.filter(id => this.state.updates[id].severity === PK.Enum.INFO_SECURITY);
+        if (type === UPDATES.KPATCHES) {
+            ids = ids.filter(id => isKpatchPackage(this.state.updates[id].name));
+        }
 
         PK.transaction()
                 .then(transactionPath => {
@@ -1301,7 +1323,7 @@ class OsUpdates extends React.Component {
     }
 
     renderContent() {
-        let applySecurity, applyAll;
+        let applySecurity, applyKpatches, applyAll;
 
         /* On unregistered RHEL systems we need some heuristics: If the "main" OS repos (which provide coreutils) require
          * a subscription, then point this out and don't show available updates, even if there are some auxiliary
@@ -1353,10 +1375,11 @@ class OsUpdates extends React.Component {
         {
             const num_updates = Object.keys(this.state.updates).length;
             const num_security_updates = count_security_updates(this.state.updates);
+            const num_kpatches = count_kpatch_updates(this.state.updates);
             const highest_severity = find_highest_severity(this.state.updates);
 
             applyAll = (
-                <Button id={num_updates == num_security_updates ? "install-security" : "install-all"} variant="primary" onClick={ () => this.applyUpdates(false) }>
+                <Button id={num_updates == num_security_updates ? "install-security" : "install-all"} variant="primary" onClick={ () => this.applyUpdates(UPDATES.ALL) }>
                     { num_updates == num_security_updates
                         ? _("Install security updates")
                         : _("Install all updates") }
@@ -1364,8 +1387,15 @@ class OsUpdates extends React.Component {
 
             if (num_security_updates > 0 && num_updates > num_security_updates) {
                 applySecurity = (
-                    <Button id="install-security" variant="secondary" onClick={ () => this.applyUpdates(true) }>
+                    <Button id="install-security" variant="secondary" onClick={ () => this.applyUpdates(UPDATES.SECURITY) }>
                         {_("Install security updates")}
+                    </Button>);
+            }
+
+            if (num_kpatches > 0) {
+                applyKpatches = (
+                    <Button id="install-kpatches" variant="secondary" onClick={ () => this.applyUpdates(UPDATES.KPATCHES) }>
+                        {_("Install kpatch updates")}
                     </Button>);
             }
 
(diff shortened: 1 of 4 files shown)


pkg/packagekit/updates.jsx
```diff
@@ -284,7 +284,7 @@ function updateItem(info, pkgNames, key) {
         pkgsTruncated.push(<span key="more">…</span>);
 
     if (pkgNames.some(p => p.name.startsWith("kpatch-patch")))
-        pkgsTruncated.push(<>{" "}<Badge>{_("patches")}</Badge></>);
+        pkgsTruncated.push(<React.Fragment key={`${key}-kpatches-spacer`}>{" "}<Badge key={key + "kpatches"}>{_("patches")}</Badge></React.Fragment>);
```
marusak, line 301: Why doing the same thing with two different syntaxes? Once `${key}` and then `key +`. But also the `Badge` does not necessarily need a key

pkg/packagekit/updates.scss
```diff
@@ -6,6 +6,13 @@
     @extend .ct-card;
 }
 
+@media screen and (max-width: 1200px) {
```
marusak, line 9: Do we do this somewhere else as well? Seems like it should be some generic pattern?

marusak (changes requested):
Can you please also post screenshot? But looks good in general, thanks!

pkg/packagekit/updates.scss
jelly: This is copied from cockpit-podman, I am not sure if we use this somewhere else, if so we should maybe consider putting it in patternfly-cockpit.scss?

martinpitt: AFAICS, pkg/lib/patternfly/patternfly-4-overrides.scss if it can be considered a bug in PF itself -- then it should be reported there, and added with a reference to the upstream issue. Otherwise, if it's cockpit specific, then pkg/lib/page.scss.

jelly: Both of these cases occur in our own ListingTable, so maybe the fix should be there.

garrett: We do have media queries all over the place... but do you mean if we should use variables instead of hardcoding the pixel* value?

(* pixels in CSS != pixels on your screen... sometimes they do, but it's different for high DPI devices, browser scaling, etc.)

For badge stretching, yeah, I agree with @martinpitt: This should probably be in our overrides and fixed upstream.

pkg/packagekit/updates.scss
```diff
@@ -6,6 +6,13 @@
...
+        // Badges should not stretch in mobile/tablet mode
+       .pf-c-table [data-label] > .pf-c-badge {
+           justify-self: start;
+       }
```
garrett, line 13: If this is moved more upstream, then it might need to handle other types of widgets too, other than just badges. For example, the badly-named ["label" component](https://www.patternfly.org/v4/components/label) is basically the same thing as a badge, but has a different name.

There are _probably_ other components that shouldn't be stretched too, but both badge/label are obvious ones that shouldn't be stretched in mobile mode.

HOWEVER, we might want to consider a so-called ["label" in a basic label group](https://www.patternfly.org/v4/components/label-group#basic) (of one), where the group is stretched but the labels aren't.

Then we probably wouldn't need any custom CSS, at the expense of the DOM becoming a little bit more needlessly complex.

jelly:
> Can you please also post screenshot? But looks good in general, thanks!

Done.

pkg/packagekit/updates.jsx
jelly: Hmmm, I later realised I could just use `${key}`, dropped the badge key as React is happy about the fragment having a key.

pkg/packagekit/updates.scss
jelly: A LabelGroup does not allow stretching of it's contents, so I went with that as solution and this CSS rule can be dropped.

pkg/packagekit/updates.scss
```diff
@@ -15,6 +15,12 @@
     }
 }
 
+.kpatches-labelgroup ul.pf-c-label-group__list,
```
marusak, line 18: Same question as before :) Do we use this elsewhere or should we?

marusak (approved):
Thank you!

pkg/packagekit/updates.jsx
martinpitt: The later "label group" commit cleans this up, but please rebase this for a clean history. Indeed let's just drop the Badge key here.

pkg/packagekit/updates.jsx
```diff
@@ -284,7 +286,8 @@ function updateItem(info, pkgNames, key) {
         pkgsTruncated.push(<span key="more">…</span>);
 
     if (pkgNames.some(p => p.name.startsWith("kpatch-patch")))
-        pkgsTruncated.push(<React.Fragment key={`${key}-kpatches-spacer`}>{" "}<Badge key={key + "kpatches"}>{_("patches")}</Badge></React.Fragment>);
+        pkgsTruncated.push(<LabelGroup key={`${key}-kpatches-spacer`} className="kpatches-labelgroup">
+            {" "}<Badge color="blue" variant="filled" isCompact>{_("patches")}</Badge></LabelGroup>);
```
martinpitt, line 304: Not a veto, but perhaps you can expand this to have proper jsx indentation/line breaks.

pkg/packagekit/updates.jsx
```diff
@@ -158,6 +164,14 @@ function count_security_updates(updates) {
     return num_security;
 }
 
+function count_kpatch_updates(updates) {
+    let num_kpatches = 0;
+    for (const u in updates)
+        if (updates[u].name.startsWith("kpatch-patch"))
+            ++num_kpatches;
+    return num_kpatches;
+}
```
martinpitt, line 173: I don't particularly like duplicating the "detect kpatch package name" logic here. updateItem() already has this:

    if (pkgNames.some(p => p.name.startsWith("kpatch-patch")))

and below you have another instance:

    ids = ids.filter(id => this.state.updates[id].name.startsWith("kpatch-patch"));

Can you please move this out into a `filterKpatchPackages()` helper function, that does the latter, and use it in all three places? I'm not sure how robust/reliable that prefix is, or whether e.g. other OSes (Ubuntu also supports kpatches, but in a slightly different form) also name them that way.

pkg/packagekit/updates.jsx
```diff
@@ -1349,22 +1366,30 @@ class OsUpdates extends React.Component {
         {
             const num_updates = Object.keys(this.state.updates).length;
             const num_security_updates = count_security_updates(this.state.updates);
+            const num_kpatches = count_kpatch_updates(this.state.updates);
             const highest_severity = find_highest_severity(this.state.updates);
 
             applyAll = (
-                <Button id={num_updates == num_security_updates ? "install-security" : "install-all"} variant="primary" onClick={ () => this.applyUpdates(false) }>
+                <Button id={num_updates == num_security_updates ? "install-security" : "install-all"} variant="primary" onClick={ () => this.applyUpdates(UPDATES.ALL) }>
```
martinpitt, line 1373: Yay for replacing booleans with descriptive enums! :+1:

test/verify/check-packagekit
```diff
@@ -196,14 +196,24 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
...
         b.click("#status .pf-c-card__actions button")
         b.wait_visible(".pf-c-badge:contains('patches')")
-        b.click("button:contains('Install all updates')")
+        b.click("button:contains('Install kpatch updates')")
         b.click("button:contains('Continue')")
         b.wait_in_text("#status", "Kernel live patch kpatch_3_10_0_1062_1_1 is active")
+        # Other updates are not installed
+        self.check_nth_update(1, "secdeclare", "3-4.b1", "security", 1,
+                              desc_matches=["Will crash your data center"], cves=["CVE-2014-123456"])
+        self.check_nth_update(2, "vanilla", "1.0-2")
```
martinpitt, line 216: You can additionally check that the installed version is still 1.0-1.

martinpitt (changes requested):
Thanks! Some minor style nits, but this looks fine to me. Can you please make the release note screenshot more narrow? (Do it from a half or at most ¾/ width browser window) It's too wide for the blog post.

pkg/packagekit/updates.jsx
jelly: I tried to find a better way to identify kpatches packages but as far as I can tell there is no constant defined in either PackageKit or DNF to identify kpatches only security/normal updates.

martinpitt: Right, that's fine. I'd just like to have a single central function for it, instead of three places.

jelly: The data is not entirely the same however, pkgNames is a list of dict's [{pkgname: "lala"}], this.state.updates is a dict.

test/verify/check-packagekit
jelly: Done.

pkg/packagekit/updates.jsx
jelly: Not entirely sure I like my cooked up solution as it involves creating a dict from a list of dictionaries, so I don't have to deal with different data types in `filterKpatchPackages`.

martinpitt: Right, that looks a bit ugly. Perhaps a simpler `isKpatchPackage(name) → bool` would be better?

martinpitt (changes requested):
Thanks for the cleanups! This looks good to me now. Please fix the page crash (see failed tests), then I'll :+1:

martinpitt (approved):
/me chuckles at `state.upates`, I'm glad our tests caught that :grin: Thanks!

martinpitt:
@jelly : Please still make a new screenshot for the release notes, the current one is much too wide.

garrett:
The screenshot was scaled... I'm not sure why. It was also extremely wide, so I fixed that. But I can't fix the blurriness.

Also: looking at this, I think the security and kpatch buttons are in the wrong order. It really should be kpatch, security, all, going from fewer updates to more updates, with all at the right. This also would have the same rightmost buttons on systems which are not kpatch capable. I guess we should do a quick follow-up PR to swap them before the release? (And update the screenshot.)

jelly:
image

garrett:
@jelly: I've updated the screenshot above based on your latest screenshot here and regenerated the notes.

PR #17224, opened 2022-04-06
test: add `--track-naughties` option for `run-tests`

ptoscano:
Reporting/updating the status of naughties requires proper permissions
(e.g. a GitHub token to update comments of Cockpituous), which is not
something common outside of the cockpit project members. Also, local
runs ought to not change the status of naughties, as the runs may be
local testing.

Hence, slightly change the default behaviour of `run-tests` to not update
the status of naughties anymore; introduce a new `--track-naughties` that
restores the old behaviour, and which will need to be used in CI
invocations of `run-tests`.

Use `--track-naughties` in existing `run-tests` invocation, to retain the
old behaviour.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -73,7 +73,7 @@ class Test:
 
         return poll_result
 
-    def finish(self, affected_tests, print_tap=True, thorough=False):
+    def finish(self, affected_tests, print_tap=True, thorough=False, track_naughties=False):
         """Returns if a test should retry or not
 
         Call test-failure-policy on the test's output, print if needed.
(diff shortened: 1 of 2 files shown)


test/run-verify-host-user.sh
```diff
@@ -185,7 +185,8 @@ done
 
 # execute run-tests
 test/common/run-tests --test-dir test/verify --nondestructive $exclude_options \
-    --machine localhost:22 --browser localhost:9090 $TESTS || RC=$?
+    --machine localhost:22 --browser localhost:9090 --track-naughties \
```
martinpitt, line 188: This one we should drop, as that doesn't work in packit and downstream tests. It doesn't work there because these don't have a GitHub token. That will implicitly disable naughty updating, but it's confusing to explicitly enable it here (in vain).

martinpitt (changes requested):
Thanks!

test/run-verify-host-user.sh
ptoscano: Ah thanks, I was not too sure about it, so went to the safest option. I will drop it.

martinpitt (approved):

PR #17228, opened 2022-04-07
storaged: add spacing between smart data

jelly:
From:

image

To:

image

(With mocked data, my nvme is not that chill :cold_face: :snowflake: )


martinpitt (approved):
Heh, thanks! I wonder if a simple text space wouldn't have done, but looks fine. :grin:

PR #17229, opened 2022-04-08
test: Check pam_access denied logins

martinpitt:
These should produce a proper error message. This was apparently broken
in the past: https://bugs.debian.org/859648

Skip this on OSTree, as root logins don't work there.

### DIFF
--- a/test/verify/check-static-login
+++ b/test/verify/check-static-login
@@ -162,12 +162,14 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
 
             # test login with tcsh
             if not m.image.startswith("rhel") and not m.image.startswith("centos") and not m.image == "arch":  # no tcsh in RHEL and in Arch Linux (TODO: available in [community])
-                m.execute("sed -r -i.bak '/^admin:/ s_:[^:]+$_:/bin/tcsh_' /etc/passwd")
-                b.login_and_go()
-                b.enter_page('/system')
-                b.wait_visible('.system-information')
-                b.logout()
-                m.execute("mv /etc/passwd.bak /etc/passwd")
+                try:
+                    m.execute("sed -r -i.bak '/^admin:/ s_:[^:]+$_:/bin/tcsh_' /etc/passwd")
+                    b.login_and_go()
+                    b.enter_page('/system')
+                    b.wait_visible('.system-information')
+                    b.logout()
+                finally:
+                    m.execute("mv /etc/passwd.bak /etc/passwd")
 
             # login with user shell that prints some stdout/err noise
             # having stdout output in ~/.bashrc confuses docker, so don't run on OSTree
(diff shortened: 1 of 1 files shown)


test/verify/check-static-login
```diff
@@ -162,12 +162,14 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
...
-                b.logout()
-                m.execute("mv /etc/passwd.bak /etc/passwd")
+                try:
+                    m.execute("sed -r -i.bak '/^admin:/ s_:[^:]+$_:/bin/tcsh_' /etc/passwd")
+                    b.login_and_go()
+                    b.enter_page('/system')
+                    b.wait_visible('.system-information')
+                    b.logout()
+                finally:
+                    m.execute("mv /etc/passwd.bak /etc/passwd")
```
allisonkarlitskaya, line 172: What happened with the idea to save/overlay all of `/etc` for non-destructive tests?

allisonkarlitskaya (approved):
This looks good of course, but one comment/question:

This test is getting really quite long.  I wonder if it makes sense to break it up some?  I guess even though it's long (in terms of number of lines) it's probably short in time, but since it's non-destructive it would also probably be very little additional overhead to have more tests...

I don't mean to suggest a change here.. more like, I wonder what your opinion is for best practices for similar patches in the future...

martinpitt:
> This test is getting really quite long

Indeed. I started with some assumption that check-static-login was destructive, so I didn't want to go through a new machine boot cycle for this. But later I realized that it is actually nd already. So I'll split this out into its own test indeed. That will also simplify the cleanup.

> What happened with the idea to save/overlay all of /etc for non-destructive tests?

I gave up on that, see PR #17137 and #17139 . It broke too much :cry:

martinpitt:
Split into a new test case, thanks for pointing out!

test/verify/check-static-login
```diff
@@ -719,6 +721,22 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
...
+    def testPamAccess(self):
+        b = self.browser
+
+        # root login works by default
+        self.login_and_go(user="root")
+        b.logout()
+
+        # disable root login with pam_access
+        self.write_file("/etc/security/access.conf", "- : root : ALL\n", append=True)
+        self.sed_file("1 aaccount required pam_access.so", "/etc/pam.d/cockpit")
```
allisonkarlitskaya, line 734: `a` :)

allisonkarlitskaya (approved):
Thanks!

test/verify/check-static-login
martinpitt: How do you mean? Is there a way to make this more readable and not look like a typo?

allisonkarlitskaya: No. I'm just happy to see you using `a` here, after our discussion the other day :)

martinpitt:
c8s repos once again are broken.. but this was green before the split, so should be fine.

PR #17230, opened 2022-04-08
node_modules: a new approach

allisonkarlitskaya:
Heavily WIP.

Rename the `tools/node-modules rebuild` command to `install` and change
how it works: it will now download the `node:alpine` container and use
it to build the contents of the node_modules/ directory.

This has two major advantages:

 - this improves the security of the workstation of the developer
   running this command, since `npm` is run in a container and doesn't
   have access to the system anymore — in fact, `npm` need not even be
   installed on the developer's workstation anymore.

 - by using a given container image with a set version of `npm` inside,
   we can ensure that we get deterministic builds of `node_modules`, no
   matter who performs them.

The last step enables an even larger change: move away from our approach
of having the bots build `node_modules/` commits for us (and drop the
workflow), preferring instead to have those commits created by
individual developers proposing PRs.  The commits must be pushed (new
command: `tools/node-modules push`) before opening a PR.

The approach to security now involves repeating the build of the
node_modules/ commit from inside of the reposchutz check.  Since this
should all be deterministic now, we should get the same tree checksum in
both places.

One drawback of this approach is that it's not as accessible to external
contributors: previously a team member would need to invoke `/npm
install` on behalf of the contributor, but this is no longer possible.
Nobody ever remembers if we actually ever used this ability, though.

Other outstanding points:

 - we need to figure out how we can test this

 - we need to figure out how this will impact the npm-update workflow.


tools/node-modules
```diff
@@ -56,35 +56,96 @@ EOF
...
-            exit 1
-        fi
-    done
+cmd_npm_install() {
+    if [ -f /run/.toolboxenv ]; then
+        exec flatpak-spawn --host -- "$0" npm_install "$@"
+        exit 1
+    fi
+
+    CACHE="cockpit-project-npm-cache-volume"
```
martinpitt, line 65: Hmm, this feels like clutter to me. Developers don't do this very often. If you want to cache it, could we use ./tmp/npm-cache/ ?

tools/node-modules
```diff
@@ -56,35 +56,96 @@ EOF
...
+    IMAGE="docker.io/library/node:alpine"
+
+    # make sure the node user can write to the cache volume
+    podman run \
+        --rm \
+        --volume "${CACHE}":/home/node/.npm:U \
+        "${IMAGE}" chown -R node:node /home/node >&2
+
+    # do the actual work
+    podman run \
```
martinpitt, line 75: I suggest `-it` here, so that you see proper output and can stop the container with Control-C. It's otherwise a bit annoying.

tools/node-modules
```diff
@@ -56,35 +56,96 @@ EOF
...
+        --user node \
+        --workdir /home/node \
+        --interactive \
+        --attach stdin \
+        --attach stdout \
+        --attach stderr \
+        --volume "${CACHE}":/home/node/.npm \
+        "${IMAGE}" /bin/sh -c '
+            set -eux
+            tee package.json >/dev/null
```
martinpitt, line 87: I had to stare at this for a bit. `cat > package.json` would be a bit clearer to me.

tools/node-modules
```diff
@@ -56,35 +56,96 @@ EOF
...
-    npm install
+    cmd_npm_install < package.json | tar_to_git node_modules
 
-    # try to keep this vaguely in sync with what .github/workflows/node_modules-* do
-    cp package.json node_modules/.package.json
-    cp package-lock.json node_modules/.package-lock.json
-    echo /.cache > node_modules/.gitignore
+    # consistency check
+    cmp package.json node_modules/.package.json
+    < node_modules/.package-lock.json
```
martinpitt, line 117: Can this be one line, please?

martinpitt (commented):
I know it's still early, but I had to look, I'm curious :grin: I like the direction!

.github/workflows/reposchutz.yml
```diff
@@ -7,20 +7,20 @@ jobs:
...
-          set -x
-          git init -b main
-          # git fetch --shallow-exclude="${BASE_SHA}" origin "${BASE_SHA}"  # GitHub doesn't support this
-          git remote add origin "${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}"
-          git fetch origin "${BASE_SHA}" "${HEAD_SHA}"
+      - name: Clone main branch
+        uses: actions/checkout@v2
+
+      - name: Fetch PR commits
+        run: git fetch origin "${BASE_SHA}" "${HEAD_SHA}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: No.

Because this workflow runs in `pull_request_target`, the branch that gets checked out is the target branch (ie: `main`).  At this point we don't have the PR, so we need to explicitly fetch it.

The key point here is that we run the scripts that are on the `main` branch, not the (possibly modified) modified ones from the proposed branch.  That's for security reasons (although this workflow has zero permissions aside from reading from a public repository and deciding a +1/-1 check result).  Of course, this means that we can't ever make changes to the scripts and update node modules at the same time, but I don't think we ever do that...

.github/workflows/reposchutz.yml
```diff
@@ -59,15 +59,21 @@ jobs:
...
               fi
-              [ "$(git rev-parse FETCH_HEAD)" = "${node_modules}" ]
-              node_modules_package_json="$(git rev-parse "${node_modules}:.package.json")"
-              if [ "${package_json}" != "${node_modules_package_json}" ]; then
+
+              # Get the tree checksum from the node_modules commit
+              tree_hash="$(git rev-parse "${node_modules}"'^{tree}')"
+
+              # Calculate correct tree checksum for the package.json file from this commit
+              expected_tree_hash="$(tools/node-modules tree_hash "${commit}")"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Oops.

My intention was to do exactly this.  I thought that the outer for loop that this was sitting inside of only visits the commits that changed `node_modules`.  I'll revisit.  Thanks for the catch.

tools/node-modules
allisonkarlitskaya: It certainly came in handy in the last days of testing from my not-so-fast home internet connection :)

I agree that it would be quite nice if there was a way to expire unused volumes...

Hard no on bindmounting part of the hose filesystem, though.  This is about isolation.

tools/node-modules
allisonkarlitskaya: This can't work as suggested.  `--interactive` is `-i` and we're already doing that, but `-t` has some side effects that make it unsuitable for piping binary data, and (particularly) for having separate stdout (`tar`) and stderr (logs) streams.

.github/workflows/reposchutz.yml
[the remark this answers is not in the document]
martinpitt: Right, sorry for the noise!

tools/node-modules
allisonkarlitskaya: fwiw, I agree that it's a bit annoying.  I consider wrapping the entire thing in `timeout` to make it killable.  I'll put a bit more research into our possibilities here.

tools/node-modules
martinpitt: Okay -- let's keep it for now then -- we can still drop the caching once this stabilized a bit.

tools/node-modules
allisonkarlitskaya: This is what I had at first.  This is about the `set -x` and the output we see.  `tee package.json` is more instructive than `cat` (which is what you see with your suggestion).  I had also tried `dd of=package.json` but then I got the annoying transfer statistics line... and turning it off would have made it even more ugly...

Other suggestions?

tools/node-modules
allisonkarlitskaya: This is one line, although apparently entirely unobvious what I'm trying to do here.  Those are two entirely separate commands.

Again, this is about the side effects, but for `set -e` in this case.  I want the script to fail if the `.package-lock.json` file is missing, and I want it to fail with a helpful error message.  Doing the usual `test -f ...` doesn't issue an error if it doesn't exist, unfortunately...

Again, open to better suggestions.  I guess I was trying to avoid open-coding the entire `if test ...; then echo ... 2>&1; exit 1; fi` ordeal...

tools/node-modules
martinpitt: Ack, seeing the file name in `set -x` makes sense, thanks!

tools/node-modules
martinpitt: Oh right, I mis-looked. It felt like calling `cmp` against a file and stdin, but that's of course not what happens.

How about `ls -l node_modules/.package-lock.json`? For anyone looking it might be a nice spot-check (file size and such), and for non-existing files it gives an appropriate error message.

allisonkarlitskaya: Thanks for the suggestion.  `>&2` of course.

tools/node-modules
allisonkarlitskaya: After a gigantic amount of research I got to the bottom of this.  There are two issues:

 - first, you should more or less always give `--init` to containers.  Otherwise, your process ends up running as pid 1, with strange implications for signal reception.  Even a simple `podman run ... sleep 100` won't die with `^C` without `--init`.
 - (more complicated): the signal forwarding code in podman works by delivering any received signal to the `podman` command to the main command inside of the container (in our case, the shell).  In case of a long-running subcommand, `SIGINT` gets either blocked (busybox `ash`) or ignored (`bash`) by the shell, which means that pressing `^C` effectively doesn't work.  I found a neat trick to work around this: `npm install & wait`: that waits until `npm` is done, but does it in a context where the shell is able to receive the forwarded `SIGINT`.  In the event that the signal gets delivered, the shell will terminate, causing the runtime to take the entire container with it.

allisonkarlitskaya:
Some initial testing:

A run where the changes are rejected (in this case, due to a legitimate update):

https://github.com/cockpit-project/cockpit/runs/5989757050?check_suite_focus=true
```
+ tee package.json
+ wait -n
+ npm install --ignore-scripts
npm WARN deprecated popper.js@1.16.1: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1
npm WARN deprecated core-js@2.6.12: core-js@<3.4 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js.

added 768 packages, and audited 769 packages in 50s

121 packages are looking for funding
  run `npm fund` for details

3 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
npm notice 
npm notice New minor version of npm available! 8.5.5 -> 8.6.0
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v8.6.0>
npm notice Run `npm install -g npm@8.6.0` to update!
npm notice 
+ cp package.json node_modules/.package.json
+ tar '--directory=node_modules' --create .
  INDEX    node_modules

Commit 892bbe466f4208d2f40c79ea78e62534c05c174c package.json and node_modules aren't in sync!

commit 892bbe466f4208d2f40c79ea78e62534c05c174c
Author: Allison Karlitskaya <allison.karlitskaya@redhat.com>
Date:   Tue Apr 12 10:27:31 2022 +0200

    Clean up package.json a bit

 node_modules | 2 +-
 package.json | 3 +--
 2 files changed, 2 insertions(+), 3 deletions(-)

The above commit refers to the following node_modules commit:

commit 5a484be3f4b7e77af0c4ef957343bf8c1e3dc035
Author: Allison Karlitskaya <allison.karlitskaya@redhat.com>
Date:   Tue Apr 12 11:07:17 2022 +0200

    7abc91987ff4ccaf6d7a97117de088de5d14e1407475883dd6e95374e29df608  package.json

Our attempt to recreate that commit differs as follows:

 .package-lock.json                | 6 +++---
 loader-runner/lib/LoaderRunner.js | 7 ++++---
 loader-runner/package.json        | 2 +-
 3 files changed, 8 insertions(+), 7 deletions(-)
diff --git a/.package-lock.json b/.package-lock.json
index cc5e5128..d7520064 100644
--- a/.package-lock.json
+++ b/.package-lock.json
@@ -6104,9 +6104,9 @@
       }
     },
     "node_modules/loader-runner": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/loader-runner/-/loader-runner-4.2.0.tgz",
-      "integrity": "sha512-92+huvxMvYlMzMt0iIOukcwYBFpkYJdpl2xsZ7LrlayO7E8SOv+JJUEK17B/dJIHAOLMfh2dZZ/Y18WgmGtYNw==",
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/loader-runner/-/loader-runner-4.3.0.tgz",
+      "integrity": "sha512-3R/1M+yS3j5ou80Me59j7F9IMs4PXs3VqRrm0TU3AbKPxlmpoY1TNscJV/oGJXo8qCatFGTfDbY6W6ipGOYXfg==",
       "dev": true,
       "engines": {
         "node": ">=6.11.5"
Error: Process completed with exit code 1.
```

allisonkarlitskaya:
Note cockpit-project/bots#3232.

We can land this PR before we land the bots PR because in the event of an npm-update, the situation won't be worse than the current experience: someone will need to force-push an update, just as now (although the procedure for that will be different).

On the other hand, this PR absolutely needs to land before the bots one does, because we call API introduced here from the bots.

tools/git-utils.sh
```diff
@@ -116,3 +116,26 @@ unpack_from_cache() {
...
+
+# This stores a .tar file from stdin into the cache as a tree object.
+# Returns the ID.  Opposite of `git archive`, basically.
+tar_to_cache() {
+    init_cache
+
+    # Use a sub-shell to enable cleanup of the temporary directory
+    (
+        tmpdir="$(mktemp --tmpdir --directory cockpit-tar-to-git.XXXXXX)"
+        trap "rm -r '${tmpdir}'" EXIT
```
martinpitt, line 129: I'd recommend to add `INT QUITE PIPE` as well (these four are my standard invocation)

tools/node-modules
```diff
@@ -109,6 +109,29 @@ Next steps:
...
+    # Verifies that the package.json and node_modules of the given commit match.
+    commit="$(git rev-parse "$1:node_modules")"
+    fetch_sha_to_cache "${commit}"
+
+    committed_tree="$(git_cache rev-parse "${commit}^{tree}")"
+    expected_tree="$(git cat-file blob "$1:package.json" | cmd_npm_install | tar_to_cache)"
+
+    if [ "${committed_tree}" != "${expected_tree}" ]; then
+        exec >&2
+        printf "\nCommit %s package.json and node_modules aren't in sync!\n\n" "$1"
```
martinpitt, line 130: What does this parameter mean? I naïvely ran
```
❱❱❱ tools/node-modules verify
tools/node-modules: line 122: $1: unbound variable
```
It's already referenced in line 114, but I suppose the `$()` subshell doesn't inherit `set -u`.

Running it with `HEAD` argument works fine. Would that be a sensible default?

martinpitt (approved):
Three minor things, but I'm fine with ignoring/postponing them.

I changed package.json to downgrade qunit-tap from 1.5.1 to 1.5.0 (didn't commit), and ran `tools/node-modules verify HEAD`. This did not fail, which feels a bit odd?

I then ran `tools/node-modules install`, this worked, gave me a proper node_modules ref diff, and

    git diff 860b6967a9e146872b50575caf5d569661bc66f9..18893dfc3332fd0b55cbb3f3d98d8a1d7708a118

looks as expected. Thanks!

tools/git-utils.sh
allisonkarlitskaya: So I tested this by (for example) pressing `^C` and the `EXIT` hook still ran...

I guess that's because the sigint was actually being delivered to the subcommand and not the shell itself...

allisonkarlitskaya:
> I changed package.json to downgrade qunit-tap from 1.5.1 to 1.5.0 (didn't commit), and ran `tools/node-modules verify HEAD`. This did not fail, which feels a bit odd?

I don't think this is odd.  This takes the package.json from HEAD and the node_modules from HEAD, neither of which you modified...

PR #17236, opened 2022-04-11
packagekit: handle timeSinceRefresh unitialized

jelly:
When the packagekit service has not refreshed yet PackageKit returns
G_MAXUINT which on Linux corresponds to 2^32-1 and shows as last checked
136 years ago.

PackageKit code https://github.com/PackageKit/PackageKit/blob/dadea5f987d7f86baff9abf42f9462d7f76a0e0e/src/pk-transaction-db.c#L219

image

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -738,7 +738,8 @@ const UpdatesStatus = ({ updates, highestSeverity, timeSinceRefresh, tracerPacka
     const numManualSoftware = tracerPackages.manual.length;
     const numRebootPackages = tracerPackages.reboot.length;
     let lastChecked;
-    if (timeSinceRefresh !== null)
+    // PackageKit returns G_MAXUINT if the db was never checked.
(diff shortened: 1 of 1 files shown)


pkg/packagekit/updates.jsx
```diff
@@ -738,7 +738,8 @@ const UpdatesStatus = ({ updates, highestSeverity, timeSinceRefresh, tracerPacka
     const numManualSoftware = tracerPackages.manual.length;
     const numRebootPackages = tracerPackages.reboot.length;
     let lastChecked;
-    if (timeSinceRefresh !== null)
+    // PackageKit returns G_MAXUINT if the db was never checked.
+    if (timeSinceRefresh !== null || timeSinceRefresh === 2 ** 32 - 1)
```
martinpitt, line 742: ITYM `&& timeSinceRefresh !== ..` here?

jelly: That actually makes sense..

PR #17238, opened 2022-04-12
test CI

martinpitt:
Test refreshed tasks container in https://github.com/cockpit-project/cockpituous/pull/477


martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-17238-20220412-074702-b15b3810-fedora-35/log.html#247) is curious, but not related to the container refresh. In particular, firefox/chromium and all pixel tests work.

PR #17240, opened 2022-04-12
sosreport: Reimplement report downloading

mvollmer:


garrett (changes requested):
vlcsnap-2022-04-20-10h19m28s061

"You need to" and "in order" are both redundant and unnecessary. Perhaps it should be this:

"Administrative access is needed to create and access reports."

And there should be an action, such as a button that says "Turn on administrative access", similar in functionality and wording inside the banner on the overview. Except, this would be a button in the standard empty state pattern. https://www.patternfly.org/v4/components/empty-state/

It'd look like this:

image

---


vlcsnap-2022-04-20-10h19m37s866

Since we're using the empty state widget for switching to admin access, we can't really use it again in the same place.

So we'd have to use the empty state pattern for the table instead. That is, make a table without rows and have a message like "No system reports."

Move the SOS text to the top of the modal modal and fix the capitalization of SOS, instead of "All settings are optional.":

"SOS reporting collects system information to help with diagnosing problems.
This information is only stored locally on the system"

---

image

Obfuscation should be checked on by default.

image

Options should be correctly aligned to the baseline of the first item. They are currently wrong.

They'd look like this:

image

image

IIRC, I think this basically can be resolved by adding `hasNoPaddingTop` to the `FormGroup`?

---

image

Run report should not change to a different button. It should have a disabled state and show the spinner. If we want to stop the report, then that should be a secondary button to the right of the "Run report" button (with a spinner).

What's 50%? I might assume that it's the current status of the report generation, but I might be wrong. We need to tell our users what that 50% is. It should be left aligned, not right aligned.

It'd look like this:

sos-modal(2)

---

image

This table inside a card does not look correct. It has gaps, a bottom border, etc.

It should look like the Machines page's list instead:

Screenshot 2022-04-20 at 11-04-35 Virtual Machines - garrett@Bolt

I think it needs to have the `pf-c-card` class (make sure the CSS is available for it to do something).

---

image

Mobile mode has issues.

There should be padding on the left of the heading. (You might need to use some breakpoint in the heading, although, I thought that happened by default?)

The list has problems too, but adding the pf-c-card would probably fix most of them.

The "label" widget (not to be confused with a form label) doesn't have proper spacing between items and it should not be stretched. It needs to be wrapped in a label group, which should fix both of these issues. https://www.patternfly.org/v4/components/label-group

garrett:
> Is there a way to distinguish a case id from a label? For example, is a case id only digits? Is there any way to look at a sosreport tarball and tell that it was generated with "--case-id 1234" or "--label 1234"? What about only supporting "label" and suggesting to provide a case number? I think "--case-id" is only really meaningful when uploading a report via "--upload" so unless we fully support uploading, we might omit "Case number" from the dialog.

These are probably questions for @TurboTurtle and possibly @martinpitt.

My gut feeling: You should be able to match up the download with the item in the row. (I don't know the specifics however.)


> When removing a report, which files should be removed exactly? Does it maybe even make sense to leave the checksum behind?

I'm unsure about this. Probably needs to be answered by @TurboTurtle and/or @martinpitt.


> Why is the "Report label" blue in the mockup? Are those links?

Yeah, I _think_ I when I start making these, I had the idea that they'd go to a sub-page for viewing the reports on the system... but then didn't mock that up, apparently. I guess I forgot mid-way through.

For a first PR, we don't even need to address that; it can be a follow-up. Or we could just decide that there's no way to view them in Cockpit, only to download. AFAIK, we don't have viewing right now; so a minimum of basic feature parity (and then some) is enough for this PR.

mvollmer:
@garrett, thanks for the review. I think I got it all addressed.

I am not happy with how tricky the "Turn on administrative access" button is to implement. I think this should be a common pattern across Cockpit, and I'll try to make this easier in a followup PR.

mvollmer:
> * Is there a way to distinguish a case id from a label?

One has to extract the manifest.json file from the report and inspect the "cmdline" field in it.  This is too expensive in general and not possible for encrypted reports.

mvollmer:
> Does it maybe even make sense to leave the checksum behind?

No.

mvollmer:
Updated pixels here: https://github.com/cockpit-project/pixel-test-reference/compare/6f87113bb19684428bbeeeee1fdbc8e042aa7a5d..6f7d3824c8963aa1c24ce3afdb41f94f40329f97

mvollmer:
> Obfuscation should be checked on by default.

Upps, I missed this.

garrett (changes requested):
Firefox thinks this is a user/password dialog... (It's probably the same in other browsers too.)

image

To fix this, you'd have to turn off autocomplete for the entries. But it's tricky in browsers (particularly Chrome, as it doesn't respect the spec anymore).

This is pretty important as it not only suggests a username & password based on the currently filled one, it also will want to save this SOS report label and encryption passphrase for the standard login too. :fearful: 

---

Progress was stuck on 99% for a very, very long time (several minutes). It _felt_ like the thing failed and didn't finish. The progress needs to be rebalanced.

Speaking of progress, it's not in an element; it's just freeform text. It needs to be in an element for proper spacing. Compare:

Screenshot 2022-04-25 at 10-21-58 Diagnostic Reports - admin@fedora-36-127-0-0-2-2201

And here's what happens even if I put it in an non-classed span (but any element at all should work):

Screenshot 2022-04-25 at 10-23-43 Diagnostic Reports - admin@fedora-36-127-0-0-2-2201

To emphasize what's changed, I did an overlay of one screenshot over the other and adjusted transparency (it was better than a diff in this particular case):

image

Of course, taking forever is bad.

If we can't improve the performance, to get it under a second, then I suggest moving the report generation out of the modal and into the table. I guess this can be a follow-up (as it's not a regression, I think), but it should be one that's worked on as soon as this lands, as waiting on a blocked modal for many minutes at 99% is a horrible experience as-is.

---

### Summary
1. Fix the autocomplete bug (_especially_ in Chrome)
2. Fix the progress spacing (by having the text in an element, instead of just freely floating)
3. Fix the report time (forever 99% is not good)
4. Move the sync in the modal to async in the table, so: the report generation starts, the modal goes away, and the progress is in the table

3 & 4 might be tricky and could be follow-ups, as this PR with 1 & 2 fixed is definitely a huge step in the right direction and better what we have on main.

garrett:
Also: All the screenshots in the pixel tests look good, and aside from what I mentioned above, everything seems good here. So it's _relatively_ close.

mvollmer:
> To fix this, you'd have to turn off autocomplete for the entries. But it's tricky in browsers (particularly Chrome, as it doesn't respect the spec anymore).

So how do we do it, then? :-) Is this the first place in Cockpit that this is releveant?

mvollmer:
> > To fix this, you'd have to turn off autocomplete for the entries. But it's tricky in browsers (particularly Chrome, as it doesn't respect the spec anymore).
> 
> So how do we do it, then? :-) Is this the first place in Cockpit that this is releveant?

I think we are supposed to use `autocomplete="new-password"`, but that doesn't seem to have any effect.
Regardless of that, I'll put it in and the rest is browser bugs.

mvollmer:
> 3\. Fix the report time (forever 99% is not good)

We need improvements in `sos report` for that, I am afraid. I'll try to get that going.

TurboTurtle:
> > Obfuscation should be checked on by default.
> 
> Upps, I missed this.

I disagree. Obfuscation is for security-minded end users, and the vast majority of interactions between support engineers and customers do not require obfuscation.


> We need improvements in sos report for that, I am afraid. I'll try to get that going.

What would you need here?

garrett:
> I think we are supposed to use autocomplete="new-password", but that doesn't seem to have any effect.

Yeah, welcome to trying to control autocomplete.

Well, I don't consider it "browser bugs" if we know the browsers still won't work as intended. It's annoying, but we need to work around it.

I've been trying to stop autocomplete across browsers and just couldn't get it working as expected here in this PR. I guess let's defer it?

Here are various attempts I tried to make Firefox behave. Some things in it are working in Chrome, apparently:

```patch
diff --git pkg/sosreport/index.jsx pkg/sosreport/index.jsx
index b5210d41a..5c8aee1ee 100644
--- pkg/sosreport/index.jsx
+++ pkg/sosreport/index.jsx
@@ -356,12 +356,14 @@ const SOSDialog = ({ onClose }) => {
         <br />
         <Form isHorizontal>
             <FormGroup label={_("Report label")}>
-                <TextInput value={label} onChange={setLabel} />
+                <TextInput value={label} onChange={setLabel} name="sos-label" autocomplete="off" />
             </FormGroup>
             <FormGroup label={_("Encryption passphrase")}
                        helperText="Leave empty to skip encryption">
                 <InputGroup>
+                    <TextInput type="hidden" value="stop-autocomplete" />
                     <TextInput type={showPassphrase ? "text" : "password"} value={passphrase} onChange={setPassphrase}
+                               name="sos-encryption"
                                autocomplete="new-password" />
                     <Button variant="control" onClick={() => setShowPassphrase(!showPassphrase)}>
                         { showPassphrase ? <EyeSlashIcon /> : <EyeIcon /> }
```

---

> I disagree. Obfuscation is for security-minded end users, and the vast majority of interactions between support engineers and customers do not require obfuscation.

Oh. Ok. 

Generally, two of the biggest things software should do with respect to data:

- Try to protect privacy
- Try to not lose data (not relevant to this point, but it's relevant below)

I guess you're saying that personal information can be trusted with the support engineers? And that SOS reporting is mainly for sharing the data with Red Hat support engineers?

---

Speaking about losing data (above), the deletion action is called "Remove", but it should be "Delete". We probably want to have a confirmation dialog.

mvollmer:
> > We need improvements in sos report for that, I am afraid. I'll try to get that going.
> 
> What would you need here?

Progress reporting for the obfuscation phase (I think), preferably in machine readable form. Our progress reporting gets stuck on "99%" for a very long time, and I think it's the obfuscation phase, but I need to look again to be sure.

It would be best if `sos report` would output machine readable progress for the whole operation. Right now, Cockpit is guessing by looking at how many plugins have been started.

garrett:
Perhaps we should specify what gets obfuscated?

"Obfuscate network addresses, hostnames, and usernames"

Right now, "Obfuscate potentially sensitive data" sounds like there are passwords in the output and that it strips out the passwords... not network addresses, hostnames, and usernames.

mvollmer:
> Perhaps we should specify what gets obfuscated?
> 
> "Obfuscate network addresses, hostnames, and usernames"

Yes, I'll do this.

mvollmer:
New pixels: https://github.com/cockpit-project/pixel-test-reference/compare/6f87113bb19684428bbeeeee1fdbc8e042aa7a5d..964e02b6a6acd5f7c37b29e25fcce226e5b4e2fd

garrett (approved):
:+1: I like it! Thanks for the changes.

I'd suggest we try to fix the autofill in a follow-up. (It's really not as straightforward as it should be.)

marusak:
When there is no label, the row is too short - the `Download` button barely fits:
Screenshot from 2022-04-28 09-16-48

pkg/sosreport/index.jsx
```diff
@@ -117,94 +267,357 @@ function sosCreate(setProgress, setError, setErrorDetail) {
...
+    return (
+        <Modal id="sos-remove-dialog"
+               position="top"
+               variant="medium"
+               isOpen
+               onClose={D.close}
+               title={<>
+                   <ExclamationTriangleIcon className="ct-icon-exclamation-triangle" />
+                   { "\n" }
+                   {_("Delete report permanently?")}
```
marusak, line 434: This seem odd with the `\n`.
You should be able to do:
```
<Modal title={_("Delete report permanently?")} titleIconVariant="warning" />
```
for the same effect

pkg/sosreport/index.jsx
```diff
@@ -48,7 +197,8 @@ function sosCreate(setProgress, setError, setErrorDetail) {
     const archive_regex = /Your sosreport has been generated and saved in:\s+(\/[^\r\n]+)/;
```
mvollmer, line 197: This isn't used any more.

pkg/sosreport/index.jsx
mvollmer: Oh, good point, I didn't know about titleIconVariant.

pkg/sosreport/index.jsx
mvollmer: Removed.

mvollmer:
> When there is no label, the row is too short - the `Download` button barely fits:
> Screenshot from 2022-04-28 09-16-48

Yes, I think this is a general problem with those lists. On the Storage page we made them non-compact for that reason, I think, so I'll do that here as well.

mvollmer:
@garrett, only visual change is to use the non-compact ListingTable variant.

mvollmer:
There seems to be trouble with TestSOS on  testing-farm:centos-stream-8-x86_64 and testing-farm:centos-stream-9-x86_64... hmm.

mvollmer:
> My only suggestion is that if parsing the label fails (or one isn't present), default to showing the full archive name rather than `-`.

Yeah, but does the archive name give more information if it doesn't contain a label? We only get the random identifier, which people wont recognize. I was thinking we could show a more precise date (hours and minutes), and also sort by that.

mvollmer:
> There seems to be trouble with TestSOS on testing-farm:centos-stream-8-x86_64.

Those tests use Firefox 91, and apparently downloading files ending in ".gpg" doesn't work with that during a test. Firefox 98 on fedora-35 in TF works as expected.

garrett:
It looks great overall... I do wonder why the report label is implied inconsistently.

I did have a "hi there" that worked without adding the numbers in front yet the others do add numbers (is it trying to use the hostname somewhat?)

And then when there is a label, it puts "cockpit" in there. And then spaces and some other characters aren't supported. (I understand stripping some things, but can we make spaces use underscores instead?)

Screenshot 2022-05-05 at 14-00-56 Diagnostic Reports - admin@fedora-36-127-0-0-2-2201

Here's what I typed for that long string:
 
image

mvollmer:
> I did have a "hi there" that worked without adding the numbers in front yet the others do add numbers (is it trying to use the hostname somewhat?)

That looks like a bug.  It looks like you are running this via bots/vm-run, and those machines have hostnames like `fedora-36-127-0-0-2-2022`.  These end up in the archive filename, and Cockpit is supposed to strip them out again (so that only the label remains).  This doesn't seem to work for you. Only the "fedora" prefix is stripped. And one of the reports only seem to have "fedora" in the archive filename.

I think I have seen something similar in the very beginning of this PR... maybe our CI images don't always have the long hostname?

What if we reduce the magic and only parse the archive filename minimally. (And call the table column "Report".)  How `sos` constructs a archive filename can actually be configured, and thus Cockpit parsing it is inherently unreliable anyway.

So, proposal: We always show the full variable part of the archive filename (which includes hostname, label, case id, date, and some random stuff), plus a relative timestamp like "10 hours ago", and we sort from newest to oldest?

mvollmer:
> How `sos` constructs a archive filename can actually be configured,

Well, not "configured", but it is OS specific.

mvollmer:
> Looks like this:

Upps, that was the wrong sort order, I think we should show the youngest at the top. Fixed now.

garrett:
When you download the archives, should we have a prefix like `sosreport-`*? I'll often have a cluttered `~/Downloads/` folder with a lot of things; I imagine others are similar. Having a way to know what it is by filename makes a lot of sense.

Meanwhile, I tried to build this PR as-is and it refuses to compile. Here's what happens on an image-prepare:

```
~/P/cockpit/cockpit sos-overhaul ❱ test/image-prepare -v -q fedora-36
  WEBPACK  sosreport
  WEBPACK  storaged
  WEBPACK  systemd
assets by path sosreport/*.js 3.78 MiB
  asset sosreport/sosreport.js 3.72 MiB [compared for emit] (name: sosreport/sosreport) 1 related asset
  asset sosreport/po.ka.js 4.59 KiB [compared for emit]
  asset sosreport/po.uk.js 3.86 KiB [compared for emit]
  asset sosreport/po.ja.js 3.06 KiB [compared for emit]
  asset sosreport/po.pl.js 2.98 KiB [compared for emit]
  asset sosreport/po.fr.js 2.97 KiB [compared for emit]
  asset sosreport/po.cs.js 2.93 KiB [compared for emit]
  asset sosreport/po.nl.js 2.86 KiB [compared for emit]
  asset sosreport/po.he.js 2.85 KiB [compared for emit]
  + 13 assets
asset sosreport/sosreport.css 643 KiB [compared for emit] (name: sosreport/sosreport) 1 related asset
asset sosreport/sosreport.png 5.68 KiB [compared for emit] [from: pkg/sosreport/sosreport.png] [copied]
asset sosreport/index.html 1.19 KiB [compared for emit] [from: pkg/sosreport/index.html] [copied]
asset sosreport/manifest.json 256 bytes [compared for emit] [from: pkg/sosreport/manifest.json] [copied]
Entrypoint sosreport/sosreport 4.35 MiB (4.08 MiB) = sosreport/sosreport.css 643 KiB sosreport/sosreport.js 3.72 MiB 2 auxiliary assets
orphan modules 3.04 MiB (javascript) 746 KiB (css/mini-extract) [orphan] 2627 modules
runtime modules 1.25 KiB 6 modules
built modules 2.69 MiB (javascript) 643 KiB (css/mini-extract) [built]
  javascript modules 2.69 MiB 982 modules
  css modules 643 KiB
    modules by path ./node_modules/@patternfly/react-styles/css/components/ 284 KiB
      modules by path ./node_modules/@patternfly/react-styles/css/components/Table/ 0 bytes 4 modules
      + 24 modules
    modules by path ./pkg/ 312 KiB
      modules by path ./pkg/lib/*.css 191 bytes 2 modules
      + 3 modules
    modules by path ./node_modules/@patternfly/react-styles/css/layouts/ 44.9 KiB
      css ./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[5].use[1]!./node_modules/sass-loader/dist/cjs.js??ruleSet[1].rules[5].use[2]!./node_modules/@patternfly/react-styles/css/layouts/Flex/flex.css 44.7 KiB [built] [code generated]
      css ./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[5].use[1]!./node_modules/sass-loader/dist/cjs.js??ruleSet[1].rules[5].use[2]!./node_modules/@patternfly/react-styles/css/layouts/Bullseye/bullseye.css 187 bytes [built] [code generated]
    css ./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[5].use[1]!./node_modules/sass-loader/dist/cjs.js??ruleSet[1].rules[5].use[2]!./node_modules/@patternfly/react-styles/css/utilities/Accessibility/accessibility.css 2.68 KiB [built] [code generated]

ERROR in /var/home/garrett/Project/cockpit/cockpit/pkg/sosreport/index.jsx
  540:1  error  Expected indentation of 42 spaces but found 40  indent
  541:1  error  Expected indentation of 42 spaces but found 40  indent
  542:1  error  Expected indentation of 42 spaces but found 40  indent

✖ 3 problems (3 errors, 0 warnings)
  3 errors and 0 warnings potentially fixable with the `--fix` option.

webpack 5.72.0 compiled with 1 error in 44642 ms
make[1]: *** [Makefile:8899: dist/sosreport/manifest.json] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:8019: dist] Error 2
  File "/var/home/garrett/Project/cockpit/cockpit/test/image-prepare", line 216, in <module>
    sys.exit(main())
  File "/var/home/garrett/Project/cockpit/cockpit/test/image-prepare", line 181, in main
    dist_tar = subprocess.check_output([f'{BASE_DIR}/tools/make-dist'], text=True).strip()
  File "/usr/lib64/python3.10/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib64/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/var/home/garrett/Project/cockpit/cockpit/tools/make-dist']' returned non-zero exit status 2.
```

mvollmer:
> I tried to build this PR as-is and it refuses to compile.

Yep, sorry, should be fixed now.

mvollmer:
> When you download the archives, should we have a prefix like sosreport-*? 

Yes, the downloaded file will have the full archive file name, which always includes `sosreport`. Should we show this full file name in the table?  Parts of it would be redundant with the "Attributes" column. Looks like this:

image

For comparison, the same reports with stripped names:

image

mvollmer:
> And then when there is a label, it puts "cockpit" in there

I have removed this again. It was meant to mark reports that have been produced by Cockpit for some statistics collection inside Redhat. It was not meant to be shown, but that logic was apparently buggy and probably responsible for your inconsistent labels.

garrett:
> > When you download the archives, should we have a prefix like sosreport-*?
> 
> Yes, the downloaded file will have the full archive file name, which always includes `sosreport`.

Yeah, my concern was just the files getting lost with names that aren't descriptive in someone's ~/Downloads/ folder.

> Should we show this full file name in the table? Parts of it would be redundant with the "Attributes" column. 

No, that's fine.

garrett (dismissed):
Looks good here! Thanks! :+1: 

---

My one issue is how autocomplete is broken... but I looked into it _yet again_ today and I still don't see anything we can do to handle this better. Even still, I'd consider that a follow-up.

Screenshots to show what I mean:

You're greeted with the autocomplete for the user who logged in by default in Firefox:

Screenshot from 2022-05-18 11-23-47

If you select it, it fills in your account's user/pass:

Screenshot from 2022-05-18 11-23-58

Even if you don't fill in your user, it still will suggest a passphrase. (This might actually be OK though, I guess?)

Screenshot from 2022-05-18 11-24-16

Anyway, there seems to be nothing other than clumsy workarounds (that no longer work) that we can do, and it varies per browser, and just generally doesn't work as expected anywhere.

So I guess we just have to (sigh) ignore the problem for now. You and I have spent a lot of time trying to find solutions, as have others across the Internet (and none of the old "solutions" work today).

---

**TL;DR**: :+1: Let's land this!

pkg/sosreport/index.jsx
```diff
@@ -117,94 +196,348 @@ function sosCreate(setProgress, setError, setErrorDetail) {
...
+            : null }
+        <p>{ _("SOS reporting collects system information to help with diagnosing problems.") }</p>
+        <p>{ _("This information is stored only on the system.") }</p>
+        <br />
+        <Form isHorizontal>
+            <FormGroup label={_("Report label")}>
+                <TextInput id="sos-dialog-ti-1" value={label} onChange={setLabel} />
+            </FormGroup>
+            <FormGroup label={_("Encryption passphrase")}
+                       helperText="Leave empty to skip encryption">
```
marusak, line 319: This should be localized as well.

marusak:
Look good, just one forgotten translation!

marusak (approved):
Thanks!

PR #17241, opened 2022-04-12
workflows: Split npm-update to PF and non PF tasks

marusak:
Run PF update every Monday and anything else try to update on Tue, Thu and Sat.


martinpitt (approved):
At first I thought "let's rather do this with `[ $(date +%w) -eq 2 ]`, but indeed for running the workflow manually having a copy is more convenient.

PR #17243, opened 2022-04-12
workflows/npm-update(-pf).yml: change the environment

allisonkarlitskaya:
Instead of running the npm-update workflow in the `self` environment and    
using SSH to push, push via HTTPS (with the `contents: write` permission    
now granted).    
    
This frees us to use the `node-cache` environment, which lets use use    
the `DEPLOY_KEY` secret from there with `tools/node-modules push`.    
    
Eventually we may want to create an environment with both deploy keys    
inside of it so that we can force-push the commit after we create the    
PR, in order to trigger the testing workflows to run.  Since we're not    
actually doing this force-push at the moment anyway, let's take this    
approach as the easier one.

Take a look at the branch here: 

https://github.com/cockpit-project/cockpit/commits/npm-update-workflow

It has the changes from this PR, plus a bogus commit to revert to old PF versions.  It had this workflow run:

https://github.com/cockpit-project/cockpit/runs/5993739199?check_suite_focus=true

which filed this PR:

https://github.com/cockpit-project/cockpit/pull/17242

### DIFF
--- a/.github/deploy-keys.sh
+++ b/.github/deploy-keys.sh
@@ -58,7 +58,6 @@ deploy_env "${THIS}-weblate"
 deploy_env "${THIS}-dist"
 
 # https://github.com/cockpit-project/node-cache
-#   - npm-install.yml
 #   - release.yml
 deploy_env node-cache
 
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Danke!

PR #17244, opened 2022-04-12
systemd: allow removal of cockpit created timers

jelly:
Cockpit allows for the creation of simple systemd system timers without
the option for deletion. This commit implements deletion for all timers
created in /etc/systemd/system/ by cockpit or by hand. The unit which is
triggered by the timer is also removed. Masked timers are not allowed to
be deleted as these are system timers in /etc/systemd/system symlinked
to /dev/null.

---

# Allow removal of user created timers

User created systemd timers in /etc/systemd/system can now be removed in cockpit.
 
image

image

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -23,9 +23,10 @@ import {
     Alert, Button,
     DescriptionList, DescriptionListTerm, DescriptionListGroup, DescriptionListDescription,
     Dropdown, DropdownItem, DropdownSeparator, KebabToggle,
-    ExpandableSection,
+    Flex, ExpandableSection,
     Tooltip, TooltipPosition,
     Card, CardBody, CardTitle, Text, TextVariants,
+    List, ListItem,
     Modal, Spinner, Switch
 } from "@patternfly/react-core";
 import {
@@ -38,6 +39,7 @@ import cockpit from "cockpit";
 import { systemd_client, SD_MANAGER, SD_OBJ } from "./services.jsx";
 import * as timeformat from "timeformat";
 import { useDialogs, DialogsContext } from "dialogs.jsx";
+import { ModalError } from 'cockpit-components-inline-notification.jsx';
 
 import './service-details.scss';
 
@@ -98,10 +100,12 @@ const ServiceConfirmDialog = ({ id, title, message, confirmText, confirmAction }
  *      Method for calling unit methods like `UnitStart`
  *  - fileActionCallback
  *      Method for calling unit file methods like `EnableUnitFiles`
+ *  - deleteActionCallback
+ *      Method for calling deleting the systemd unit
  *  - disabled
  *      Button is disabled
  */
-const ServiceActions = ({ masked, active, failed, canReload, actionCallback, fileActionCallback, disabled }) => {
+const ServiceActions = ({ masked, active, failed, canReload, actionCallback, deleteActionCallback, fileActionCallback, disabled }) => {
     const Dialogs = useDialogs();
     const [isActionOpen, setIsActionOpen] = useState(false);
 
@@ -128,6 +132,13 @@ const ServiceActions = ({ masked, active, failed, canReload, actionCallback, fil
         } else {
             actions.push(
                 <DropdownItem key="start" onClick={() => actionCallback("StartUnit")}>{ _("Start") }</DropdownItem>
+            );
+        }
+
+        if (deleteActionCallback) {
+            actions.push(<DropdownSeparator key="delete-divider" />);
+            actions.push(
+                <DropdownItem key="delete" className="pf-m-danger" onClick={() => deleteActionCallback()}>{ _("Delete") }</DropdownItem>
             );
         }
 
@@ -196,11 +207,15 @@ export class ServiceDetails extends React.Component {
             waitsAction: false,
             waitsFileAction: false,
             unit_properties: {},
+            showDeleteDialog: false,
+            unitPaths: [],
         };
 
         this.onOnOffSwitch = this.onOnOffSwitch.bind(this);
         this.unitAction = this.unitAction.bind(this);
         this.unitFileAction = this.unitFileAction.bind(this);
+        this.deleteAction = this.deleteAction.bind(this);
+        this.deleteTimer = this.deleteTimer.bind(this);
 
         this.unitType = props.unit.Id.split('.').slice(-1)[0];
         this.unitTypeCapitalized = this.unitType.charAt(0).toUpperCase() + this.unitType.slice(1);
@@ -278,23 +293,27 @@ export class ServiceDetails extends React.Component {
         }
     }
 
-    unitAction(method, extra_args) {
+    unitAction(method, extra_args, catchExc = true) {
         if (extra_args === undefined)
             extra_args = ["fail"];
         this.setState({ waitsAction: true });
-        systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, method, [this.props.unit.Names[0]].concat(extra_args))
-                .catch(error => {
-                    this.show_error(error.toString());
-                    this.setState({ waitsAction: false });
-                });
+        const promise = systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, method, [this.props.unit.Names[0]].concat(extra_args));
+        if (catchExc) {
+            return promise.catch(error => {
+                this.show_error(error.toString());
+                this.setState({ waitsAction: false });
+            });
+        } else {
+            return promise;
+        }
     }
 
-    unitFileAction(method, force) {
+    unitFileAction(method, force, catchExc = true) {
         this.setState({ waitsFileAction: true });
         const args = [[this.props.unit.Names[0]], false];
         if (force !== undefined)
             args.push(force == "true");
-        systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, method, args)
+        const promise = systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, method, args)
                 .then(([results]) => {
                     if (results.length == 2 && !results[0])
                         this.show_note(_("This unit is not designed to be enabled explicitly."));
@@ -302,11 +321,61 @@ export class ServiceDetails extends React.Component {
                      * see https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl.c [enable_unit function]
                      */
                     systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, "Reload", null);
-                })
-                .catch(error => {
-                    this.show_error(error.toString());
-                    this.setState({ waitsFileAction: false });
                 });
+        if (catchExc) {
+            return promise.catch(error => {
+                this.show_error(error.toString());
+                this.setState({ waitsFileAction: false });
+            });
+        } else {
+            return promise;
+        }
+    }
+
+    deleteAction() {
+        this.getUnitPaths().then(unitPaths => {
+            this.setState({ showDeleteDialog: true, unitPaths });
+        });
+    }
+
+    async getUnitPaths() {
+        const paths = [this.props.unit.FragmentPath];
+
+        await Promise.all(this.props.unit.Triggers.map(async trigger => {
+            // Getting dbus properties from a non-loaded unit is not possible so resort to systemctl show
+            const unitPath = await cockpit.spawn(["systemctl", "show", "--value",
+                "--property", "FragmentPath", trigger]);
+            paths.push(unitPath.trim());
+        })).catch(err => console.error("failed to look up unit details:", err.toString()));
+
+        return paths;
+    }
+
+    deleteTimer() {
+        // Stop timer so we don't get race conditions when the unit is gone.
+        if (this.interval) {
+            clearInterval(this.interval);
+            this.interval = null;
+        }
+
+        const promises = [];
+        if (this.props.unit.ActiveState === "active" || this.props.unit.ActiveState === "activating")
+            promises.push(this.unitAction("StopUnit", undefined, false));
+        if (this.props.unit.ActiveState === "failed")
+            promises.push(this.unitAction("ResetFailedUnit", undefined, false));
+        if (this.props.unit.UnitFileState === "enabled")
+            promises.push(this.unitFileAction("DisableUnitFiles", undefined, false));
+
+        return Promise.all(promises).then(() => {
+            const deletions = this.state.unitPaths.filter(path => path.startsWith("/etc/systemd/system"))
+                    .map(path => cockpit.file(path, { superuser: "required" }).replace(null));
+
+            // Reload after unit/timer removal
+            return Promise.all(deletions).then(() =>
+                systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, "Reload", null)
+                        .then(() => cockpit.jump("/system/services#/?type=timer"))
+            );
+        });
     }
 
     render() {
@@ -317,6 +386,8 @@ export class ServiceDetails extends React.Component {
         const masked = this.props.unit.LoadState === "masked";
         const unit = this.state.unit_properties;
         const showAction = this.props.permitted || this.props.owner == "user";
+        const isCustom = this.props.unit.FragmentPath.startsWith("/etc/systemd/system") && !masked;
+        const isTimer = this.props.unit.is_timer;
 
         let status = [];
 
@@ -345,7 +416,7 @@ export class ServiceDetails extends React.Component {
                     <ErrorCircleOIcon className="status-icon" />
                     <span className="status">{ _("Failed to start") }</span>
                     { showAction &&
-                        <Button variant="secondary" className="action-button" onClick={() => this.unitAction("StartUnit") }>{ _("Start service") }</Button>
+                    <Button variant="secondary" className="action-button" onClick={() => this.unitAction("StartUnit") }>{ _("Start service") }</Button>
                     }
                 </div>
             );
@@ -484,6 +555,20 @@ export class ServiceDetails extends React.Component {
 
         return (
             <Card>
+                { this.state.showDeleteDialog &&
+                <DeleteModal
+                    name={this.props.unit.Description}
+                    handleCancel={() => this.setState({ showDeleteDialog: false })}
+                    handleDelete={this.deleteTimer}
+                    reason={<Flex flex={{ default: 'column' }}>
+                        <p>{_("Deletion will remove the following files:")}</p>
+                        <List>
+                            {this.state.unitPaths.map(item => <ListItem key={item}>{item}</ListItem>)}
+                        </List>
+                    </Flex>
+                    }
+                />
+                }
                 { (hasLoadError && this.props.unit.LoadState)
                     ? <Alert variant="danger" isInline title={this.props.unit.LoadState}>
                         {loadError}
@@ -503,7 +588,10 @@ export class ServiceDetails extends React.Component {
                                             </span>
                                         </Tooltip>
                                     }
-                                    <ServiceActions { ...{ active, failed, enabled, masked } } canReload={this.props.unit.CanReload} actionCallback={this.unitAction} fileActionCallback={this.unitFileAction} disabled={this.state.waitsAction || this.state.waitsFileAction} />
+                                    <ServiceActions { ...{ active, failed, enabled, masked } } canReload={this.props.unit.CanReload}
+                                                    actionCallback={this.unitAction} fileActionCallback={this.unitFileAction}
+                                                    deleteActionCallback={isCustom && isTimer ? this.deleteAction : null}
+                                                    disabled={this.state.waitsAction || this.state.waitsFileAction} />
                                 </>
                             }
                         </CardTitle>
@@ -559,4 +647,29 @@ ServiceDetails.propTypes = {
     unit: PropTypes.object.isRequired,
     permitted: PropTypes.bool.isRequired,
     isValid: PropTypes.func.isRequired,
+};
+
+const DeleteModal = ({ reason, name, handleCancel, handleDelete }) => {
+    const [inProgress, setInProgress] = useState(false);
+    const [dialogError, setDialogError] = useState(undefined);
+    return (
+        <Modal isOpen
+               showClose={false}
+               position="top" variant="medium"
+               onClose={handleCancel}
+               title={cockpit.format(_("Confirm deletion of $0"), name)}
+               titleIconVariant="warning"
+               footer={<>
+                   <Button id="delete-timer-modal-btn" variant="danger" isDisabled={inProgress} isLoading={inProgress}
+                           onClick={() => { setInProgress(true); handleDelete().catch(exc => { setDialogError(exc.message); setInProgress(false) }) }}
+                   >
+                       {_("Delete")}
+                   </Button>
+                   <Button variant="link" isDisabled={inProgress} onClick={handleCancel}>{_("Cancel")}</Button>
+               </>}
+        >
+            {dialogError && <ModalError dialogError={_("Timer deletion failed")} dialogErrorDetail={dialogError} />}
+            {reason}
+        </Modal>
+    );
 };
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1829,6 +1829,8 @@ class MachineCase(unittest.TestCase):
         '''
         m = self.machine
         self.restore_file(path, post_restore_action=post_restore_action)
+        dirname = os.path.dirname(path)
+        m.execute(f"mkdir -p {dirname}")
         m.write(path, content, append=append, owner=owner, perm=perm)
 
 
--- a/test/verify/check-system-services
+++ b/test/verify/check-system-services
@@ -212,6 +212,33 @@ Description=Test OnBoot Timer
 OnBootSec=200min
 Unit=test.service
 """)
+        self.write_file("/usr/local/lib/systemd/system/cockpit-system.timer",
+                        """
+[Unit]
+Description=Not deleteable Timer
+
+[Timer]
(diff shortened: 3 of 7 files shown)


pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,44 @@ export class ServiceDetails extends React.Component {
...
+                                                          "org.freedesktop.DBus.Properties", "Get",
+                                                          ["org.freedesktop.systemd1.Unit", 'FragmentPath']).then(([dbus_obj]) => {
+                        if (dbus_obj.v.startsWith("/etc/systemd/system")) {
+                            cockpit.file(dbus_obj.v, { superuser: "required" }).replace(null);
+                        }
+                    }));
+            });
+
+            // Reload after unit/timer removal
+            Promise.all(deletions).then(results => {
```
martinpitt, line 367: This call can fail, and needs to be handled.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,44 @@ export class ServiceDetails extends React.Component {
...
+                systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, "Reload", null).then(() => cockpit.jump("/system/services#/?type=timer"));
```
martinpitt, line 368: Long line -- please break on .then(), that's good practice anyway.

pkg/systemd/services/service-details.jsx
```diff
@@ -438,6 +492,8 @@ export class ServiceDetails extends React.Component {
         const tooltipMessage = enabled ? _("Stop and disable") : _("Start and enable");
         const hasLoadError = this.props.unit.LoadState !== "loaded" && this.props.unit.LoadState !== "masked";
         const loadError = this.props.unit.LoadError ? this.props.unit.LoadError[1] : null;
+        console.log(hasLoadError, loadError);
+        console.log(this.state.error);
```
martinpitt, line 496: Please remember to clean these up once the PR works.

test/verify/check-system-services
```diff
@@ -203,6 +203,22 @@ Description=Test OnBoot Timer
 [Timer]
 OnBootSec=200min
 Unit=test.service
+""")
+        self.write_file("/usr/lib/systemd/system/cockpit-system.timer",
```
martinpitt, line 207: How about /run or /usr/local/? Scribbling into /usr is nasty, and won't work on CoreOS with read-only /usr.

martinpitt (changes requested):
Tests still fail anyway, but first cursory review. Thanks for working on this feature!

test/verify/check-system-services
jelly: /usr/local sounds good, I found that `write_file` does not handle non-existant directories but `restore_file` does! So I've expanded `write_file` with an option to mkdir -p after calling `restore_file` and everything is cleaned up nicely.

pkg/systemd/services/service-details.jsx
jelly: Yes, sorry :(

test/verify/check-system-services
```diff
@@ -1070,6 +1097,7 @@ class TestTimers(MachineCase):
         # checks if yearly timer repeats yearly on 2037-01-01 01:22
         wait_systemctl_timer("2037-01-01 01:22")
         self.assertIn("2037-01-01 01:22", m.execute("systemctl list-timers"))
+
```
jelly, line 1100: Extra newline.

test/common/testlib.py
```diff
@@ -1829,6 +1829,10 @@ def write_file(self, path, content, append=False, owner=None, perm=None, post_re
         '''
         m = self.machine
         self.restore_file(path, post_restore_action=post_restore_action)
+        dirname = os.path.dirname(path)
+        exists = m.execute(f"if test -d {dirname}; then echo yes; fi").strip() != ""
+        if not exists:
+            m.execute(f"mkdir -p {dirname}")
```
martinpitt, line 1835: `mkdir -p` already DTRT, you can drop the `test -d`. Thanks for that change, with that we can remove a lot of `m.execute("mkdir ...")` from tests.

Please change the commit message to "create directory", as the path includes the file name.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,45 @@ export class ServiceDetails extends React.Component {
                 });
     }
 
+    deleteAction() {
+        // Stop timer so we don't get race conditions when the unit is gone.
+        if (this.interval)
+            clearInterval(this.interval);
+
+        const promises = [];
+        if (this.props.unit.UnitFileState === "enabled") {
```
martinpitt, line 341: IMHO it should be possible to delete disabled custom timers as wel. In fact, that seems like the more plausible use case even -- first disable it, then see if you are happy with this, and then either re-enable or delete.

I.e. please split this into disable and a separate runtime (stop/reset) block.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,45 @@ export class ServiceDetails extends React.Component {
                 });
     }
 
+    deleteAction() {
+        // Stop timer so we don't get race conditions when the unit is gone.
+        if (this.interval)
+            clearInterval(this.interval);
```
martinpitt, line 338: componentWillUnmount() calls clearInterval() as well. Either clearInterval() is idempotent, then the `if` condition here and in componentWillUnmount() are unnecessary. Or it's not, then both places should set `this.interval = null` afterwards.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,45 @@ export class ServiceDetails extends React.Component {
...
+            promises.push(this.unitFileAction("DisableUnitFiles", undefined));
+        }
+
+        // Remove unit files
+        Promise.all(promises).then(() => {
+            const deletions = [];
+
+            deletions.push(cockpit.file(this.props.unit.FragmentPath, { superuser: "required" }).replace(null));
+            this.props.unit.Triggers.forEach(trigger => {
+                trigger = trigger.replace('.', '_2e');
```
martinpitt, line 355: Hmm, is that really the only char that systemd escapes? Would IMHO be safer to query this from the timer unit object.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,45 @@ export class ServiceDetails extends React.Component {
...
+                deletions.push(
+                    systemd_client[this.props.owner].call(`/org/freedesktop/systemd1/unit/${trigger}`,
+                                                          "org.freedesktop.DBus.Properties", "Get",
+                                                          ["org.freedesktop.systemd1.Unit", 'FragmentPath']).then(([dbus_obj]) => {
+                        if (dbus_obj.v.startsWith("/etc/systemd/system")) {
+                            cockpit.file(dbus_obj.v, { superuser: "required" }).replace(null);
```
martinpitt, line 361: This Should Not Fail™, but let's at least console.error() if it does, to avoid a page crash and make this debuggable.

pkg/systemd/services/service-details.jsx
```diff
@@ -105,8 +104,14 @@ ServiceConfirmDialog.propTypes = {
...
  *  - fileActionCallback
  *      Method for calling unit file methods like `EnableUnitFiles`
+ *  - deleteActionCallback
+ *      Method for calling deleting the systemd timer
  *  - disabled
  *      Button is disabled
+ *  - isCustom
+ *      Unit is installed in /etc/systemd/system
+ *  - isTimer
+ *      Unit is a timer
```
martinpitt, line 114: Proposal for generalization: Leave the policy decision to the caller, and offer the "Delete" action iff `deleteActionCallback` is defined. We may want to delete other units in the future, and this avoids some awkward inconsistencies wrt. "this gets passed as property" vs. "the deleteAction() code implicitly assumes a timer anyway".

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,45 @@ export class ServiceDetails extends React.Component {
...
+                        if (dbus_obj.v.startsWith("/etc/systemd/system")) {
+                            cockpit.file(dbus_obj.v, { superuser: "required" }).replace(null);
+                        }
+                    }));
+            });
+
+            // Reload after unit/timer removal
+            Promise.all(deletions).then(results => {
+                systemd_client[this.props.owner].call(SD_OBJ, SD_MANAGER, "Reload", null)
+                        .then(() => cockpit.jump("/system/services#/?type=timer"));
```
martinpitt, line 369: This comment and the jump are currently timer specific, but the "remove triggered units" logic above should work equally well for e.g. path or socket units. I wonder if we can make the jump generic enough to just chop off the unit name. This doesn't need to happen in *this* PR, just thinking aloud (fine to generalize later).

pkg/systemd/services/service-details.jsx
```diff
@@ -327,6 +379,9 @@ export class ServiceDetails extends React.Component {
         const masked = this.props.unit.LoadState === "masked";
         const unit = this.state.unit_properties;
         const showAction = this.props.permitted || this.props.owner == "user";
+        // These timers are added by cockpit via the UI
+        const isCustom = this.props.unit.FragmentPath.startsWith("/etc/systemd/system") && !masked;
```
martinpitt, line 383: These are not cockpit specific. Any custom timer should be placed there. It's fine if cockpit can delete them, but we shouldn't pretend that they were created by Cockpit (not unless we add some `X-Created-By: Cockpit` to the unit or similar -- but IMHO we shouldn't).

test/verify/check-system-services
```diff
@@ -208,6 +208,22 @@ Description=Test OnBoot Timer
...
+""")
+        self.write_file("/usr/local/lib/systemd/system/cockpit-system.timer",
+                        """
+[Unit]
+Description=Not deleteable Timer
+
+[Timer]
+OnCalendar=*:1/2
+""")
+        self.write_file(f"{path}/deleteme.timer",
```
martinpitt, line 220: This should also write a corresponding deleteme.service, and ensure that it gets cleaned up.

test/verify/check-system-services
```diff
@@ -285,8 +301,19 @@ Unit=test.service
         b.wait_in_text(self.svc_sel('test-onboot.timer') + ' .service-unit-last-trigger', "unknown")  # last trigger
         self.run_systemctl(user, "stop test-onboot.timer")
 
-        # Selects Paths tab - the test VM does not have any user path units
         if not user:
+            self.goto_service("deleteme.timer")
+            self.do_action("Delete")
+            self.wait_page_load()
+            self.assertNotIn("deleteme.timer", m.execute("systemctl list-timers"))
```
martinpitt, line 308: Please drop the `.timer` -- it *could* be that `list-timers` just omits the suffix. Also check that both files are gone from disk.

martinpitt (changes requested):
Thanks!

pkg/systemd/services/service-details.jsx
jelly: Looking at  the `systemd-esacpe` man page, more can be escaped.. Not sure how we can query this from the timer object?

jelly: Actually systemd-escape does not .. escape the . and the triggers are a list of services so "foo.service", so I assume `hallöchen.service` should be escaped.

jelly: So hmm as we only have the service name I'm not sure how we can obtain the FragmentPath differently.

martinpitt: I played around a little bit with this, but I can't make it work:
```sh
cat <<EOF > /etc/systemd/system/foo.timer
[Timer]
OnCalendar=*:1/2
Unit=$(systemd-escape hallöchen.service)
EOF

cat <<EOF /etc/systemd/system/hallöchen.service
[Service]
ExecStart=/bin/true
EOF
```

With `Unit=hallöchen.service`, systemd complains about "Unit type not valid, ignoring: hallöchen.service", so I escaped it, it ends up as `Unit=hall\xc3\xb6chen.service`. But systemd does not accept that either:

    foo.timer: Refusing to start, unit hall\xc3\xb6chen.service to trigger not loaded

But nevertheless, I figure this will work somehow, and that interaction with shell is somehow broken.

But anyway, the proper way to translate an unit name into an object path is [GetUnit()](https://www.freedesktop.org/wiki/Software/systemd/dbus/):
```
❱❱❱ busctl call org.freedesktop.systemd1 /org/freedesktop/systemd1 org.freedesktop.systemd1.Manager GetUnit s udisks2.service
o "/org/freedesktop/systemd1/unit/udisks2_2eservice"
```
Please use that instead of the manual escaping and hardcoding the `/org/freedesktop/systemd1/unit/...` object path.

martinpitt: BTW, just *if* you want to explore something new: This deleteAction() function would be an excellent candidate for `async`/`await`. This should get rid of all the Promise.all(), duplicated .catch(), etc.

garrett:
I'm glad to see this!

There's only one little concern: 
- There should be a confirmation modal (with information on what you're deleting in the modal), as there's no way to undo the action.

pkg/systemd/services/service-details.jsx
jelly: We could async/await it, but we need to remember that we need to use try/catch then for error handling iirc.

jelly:
Found a bug with our way of looking up the corresponding systemd service, when the timer is disabled we can't query the corresponding service. So we only remove the timer.

jelly:
I'll squash everything after review, otherwise the tests are triggered again which seems wasteful.

test/common/testlib.py
```diff
@@ -1829,6 +1829,8 @@ def write_file(self, path, content, append=False, owner=None, perm=None, post_re
         '''
         m = self.machine
         self.restore_file(path, post_restore_action=post_restore_action)
+        dirname = os.path.dirname(path)
+        m.execute(f"mkdir -p {dirname}")
```
martinpitt, line 1833: This commit could drop the `mkdir` call from test/verify/check-packagekit (`mkdir -p /etc/tracer`), packagelib.py, and check-sosreport. This would clean up the tests, and prove that this works properly.

pkg/systemd/services/service-details.jsx
```diff
@@ -105,6 +105,8 @@ ServiceConfirmDialog.propTypes = {
  *      Method for calling unit methods like `UnitStart`
  *  - fileActionCallback
  *      Method for calling unit file methods like `EnableUnitFiles`
+ *  - deleteActionCallback
+ *      Method for calling deleting the systemd timer
```
martinpitt, line 109: s/systemd timer/unit/, this is generic now.

pkg/systemd/services/service-details.jsx
```diff
@@ -217,11 +226,15 @@ export class ServiceDetails extends React.Component {
             note: "",
             error: "",
             unit_properties: {},
+            showDeleteDialog: false,
```
martinpitt, line 229: This should use the new dialog protocol introduced in PR #17298 -- sorry for the collision, but let's avoid introducing more instances now.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,53 @@ export class ServiceDetails extends React.Component {
...
+
+    async getUnitPaths() {
+        const paths = [this.props.unit.FragmentPath];
+
+        await Promise.all(this.props.unit.Triggers.map(async trigger => {
+            // Getting dbus properties from a non-loaded unit is not possible so resort to systemctl show
+            const unitPath = await cockpit.spawn(["systemctl", "show", "--value",
+                "--property", "FragmentPath", trigger]);
+            paths.push(unitPath.trim());
+        })).catch(err => console.error(_("failed to look up unit details:"), err.toString()));
```
martinpitt, line 349: Let's not i18n console errors, it's a waste of translator time.

pkg/systemd/services/service-details.jsx
```diff
@@ -319,6 +332,53 @@ export class ServiceDetails extends React.Component {
...
+        });
+    }
+
+    async getUnitPaths() {
+        const paths = [this.props.unit.FragmentPath];
+
+        await Promise.all(this.props.unit.Triggers.map(async trigger => {
+            // Getting dbus properties from a non-loaded unit is not possible so resort to systemctl show
+            const unitPath = await cockpit.spawn(["systemctl", "show", "--value",
+                "--property", "FragmentPath", trigger]);
```
martinpitt, line 347: Does an error here actually trigger the `.catch()` below? That seems to happen in a different scope. Can you try to intentionally break this command (locally) and make sure it doesn't cause a page oops due to an unhandled promise rejection or (more likely) unhandled exception? As this is `await`ed, it seems it would need a `catch() {} ` block.

pkg/systemd/services/service-details.jsx
```diff
@@ -494,6 +556,20 @@ export class ServiceDetails extends React.Component {
...
             <Card>
+                { this.state.showDeleteDialog &&
+                <DeleteModal
+                    name={this.props.unit.Description}
+                    handleCancel={() => this.setState({ showDeleteDialog: false })}
+                    handleDelete={() => this.deleteTimer()}
+                    reason={<Flex flex={{ default: 'column' }}>
+                        <p>{_("Deletion will remove the following files:")}</p>
+                        <List>
+                            {this.state.unitPaths.map(item => (<ListItem key={item}>{item}</ListItem>))}
```
martinpitt, line 567: No need for the () around the ListItem.

pkg/systemd/services/service-details.jsx
```diff
@@ -576,3 +655,26 @@ ServiceDetails.propTypes = {
     permitted: PropTypes.bool.isRequired,
     isValid: PropTypes.func.isRequired,
 };
+
+const DeleteModal = (props) => {
```
martinpitt, line 659: Please expand `props` here, i.e. `({ handleCancel, handleDelete })`. That makes it easier to see which props a component expects, and enables compile-time type/variable checks.

pkg/systemd/services/service-details.jsx
```diff
@@ -576,3 +655,26 @@ ServiceDetails.propTypes = {
...
+    return (
+        <Modal isOpen
+               showClose={false}
+               position="top" variant="medium"
+               onClose={props.handleCancel}
+               title={cockpit.format(_("Confirm deletion of $0"), props.name)}
+               titleIconVariant="warning"
+               footer={<>
+                   <Button id="delete-timer-modal-btn" variant="danger" isDisabled={inProgress} isLoading={inProgress}
+                           onClick={() => { setInProgress(true); props.handleDelete().catch(() => setInProgress(false)) }}
```
martinpitt, line 670: Hmm.. so on errors we keep the dialog open, but don't show the error? That's a bit awkward.

test/verify/check-system-services
```diff
@@ -288,8 +315,21 @@ Unit=test.service
...
+            self.do_action("Delete")
+            b.click("#delete-timer-modal-btn")
+            self.wait_page_load()
+            self.assertNotIn("deleteme", m.execute("systemctl list-timers"))
+            self.assertEqual(m.execute(f"! test -f {path}/deleteme.service && echo yes").strip(), "yes")
+
+            # system timers are not allowed to be deleted
+            self.goto_service("cockpit-system.timer")
+            b.click(".service-top-panel .pf-c-dropdown button")
+            b.wait_not_in_text(".service-top-panel .pf-c-dropdown__menu", "Delete")
```
martinpitt, line 329: Please first wait for an expected action, to make sure the menu has actually opened. Otherwise this is a race condition.

martinpitt (changes requested):
Thanks!

pkg/systemd/services/service-details.jsx
jelly: It did get catched:

```
failed to look up unit details: systemctl exited with code 1
```

pkg/systemd/services/service-details.jsx
jelly: Yeah, and hmm it doesn't really work...

pkg/systemd/services/service-details.jsx
jelly: Ok, let's do this in a follow up.

pkg/systemd/services/service-details.jsx
jelly: The promises in `deleteTimer` already catch their own errors so calling `unitAction`, `unitFileAction` will show an error in a new dialog quickly and then quickly navigate away as the promises succeed. (This is super annoying)

test/common/testlib.py
jelly: Done.

pkg/systemd/services/service-details.jsx
```diff
@@ -278,35 +293,87 @@ export class ServiceDetails extends React.Component {
...
+            promises.push(this.unitAction("StopUnit", [], false));
+        if (this.props.unit.ActiveState === "failed")
+            promises.push(this.unitAction("ResetFailedUnit", [], false));
+        if (this.props.unit.UnitFileState === "enabled")
+            promises.push(this.unitFileAction("DisableUnitFiles", undefined, false));
+
+        promises.push(this.state.unitPaths.filter(path => path.startsWith("/etc/systemd/system"))
+                .map(path => cockpit.file(path, { superuser: "required" }).replace(null)));
+
+        return Promise.all(promises).then(() => {
```
jelly, line 372: Ok, this does not fully work yet, as for some reason the promises are not executed in order while they should? So an error in unitAction still removes service/timer's.

pkg/systemd/services/service-details.jsx
mvollmer: https://github.com/cockpit-project/cockpit/pull/17300 has landed, I'd say we should use `Dialogs.show` etc right away.

pkg/systemd/services/service-details.jsx
martinpitt: What do you mean with "in order"? The whole point of Promise.all() is to run all of them in parallel, and collect the results. If you need ordering, you need to run them as `first.then(() => second)`.

jelly: I think I was confused about the order of output being the order of input which might not be execution order..

martinpitt:
Nitpick: Could we have a screenshot without such a bad spelling and grammar? If that is coming from podman, let's file a bug; if that is your demo unit, could you please use a shorter title?

jelly:
> Nitpick: Could we have a screenshot without such a bad spelling and grammar? If that is coming from podman, let's file a bug; if that is your demo unit, could you please use a shorter title?

Woops, that was made up. Updated the screenshot.

pkg/systemd/services/service-details.jsx
```diff
@@ -278,35 +293,89 @@ export class ServiceDetails extends React.Component {
...
+        this.getUnitPaths().then(unitPaths => {
+            this.setState({ showDeleteDialog: true, unitPaths });
+        });
+    }
+
+    async getUnitPaths() {
+        const paths = [this.props.unit.FragmentPath];
+
+        await Promise.all(this.props.unit.Triggers.map(async trigger => {
+            // Getting dbus properties from a non-loaded unit is not possible so resort to systemctl show
```
martinpitt, line 345: FTR, that's not really true -- systemd has a [LoadUnit()](https://www.freedesktop.org/wiki/Software/systemd/dbus/) for exactly that (after all, systemctl also just uses the D-Bus protocol).

test/verify/check-system-services
```diff
@@ -288,8 +315,22 @@ Unit=test.service
...
         self.run_systemctl(user, "stop test-onboot.timer")
 
-        # Selects Paths tab - the test VM does not have any user path units
         if not user:
+            self.goto_service("deleteme.timer")
+            self.do_action("Delete")
+            b.click("#delete-timer-modal-btn")
+            self.wait_page_load()
+            self.assertNotIn("deleteme", m.execute("systemctl list-timers"))
+            self.assertEqual(m.execute(f"! test -f {path}/deleteme.service && echo yes").strip(), "yes")
```
martinpitt, line 324: Should also check that deleteme.timer does not exist any more. Also, the `yes` bit and assertEqual() shouldn't be necessary -- `m.execute()` already fails if the command fails.

martinpitt (approved):
Cheers! I'd really like to get that additional test, so if you have another reason to force-push, that's easy to add. Otherwise (if you want to land that quickly for the release) I'm happy for it to be a follow-up PR, assuming that it actually deletes the file.

pkg/systemd/services/service-details.jsx
martinpitt: .. but, calling systemctl here is acceptable here. It's much more expensive, but it's not going to be that many calls.

garrett (changes requested):
As an admin, I created a test timer and tried to delete it.

Everything looked as expected, but when I attempted to delete it:

image

The alert needs padding between it and the content... but more importantly, deletion should work. I shouldn't get a "Too few values in tuple/struct" error like this.

jelly:
> 

Both fixed.

test/verify/check-system-services
```diff
@@ -288,8 +315,23 @@ Unit=test.service
...
 
-        # Selects Paths tab - the test VM does not have any user path units
         if not user:
+            self.goto_service("deleteme.timer")
+            self.do_action("Delete")
+            b.click("#delete-timer-modal-btn")
+            self.wait_page_load()
+            self.assertNotIn("deleteme", m.execute("systemctl list-timers"))
+            self.assertEqual(m.execute(f"! test -f {path}/deleteme.service"))
+            self.assertEqual(m.execute(f"! test -f {path}/deleteme.timer"))
```
martinpitt, line 325: equal to what? :grin:  Please just drop the assertEqual() calls.

martinpitt (approved):
Thanks! LGTM to me as a puny human, bots will have the last word.

PR #17246, opened 2022-04-13
workflows: Restore issues write permissions to reposchutz

martinpitt:
Commit 0ce651fc7a dropped all write permissions. But deleting the label
requires writing to the issues API.

----

See [current workflow failure](https://github.com/cockpit-project/cockpit/runs/6001588717?check_suite_focus=true). *This* PR will fail also in any case, as reposchutz runs from main.

I temporarily removed reposchutz as a required check.

### DIFF
--- a/.github/workflows/reposchutz.yml
+++ b/.github/workflows/reposchutz.yml
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      issues: write
     timeout-minutes: 5
     env:
       HEAD_SHA: ${{ github.event.pull_request.head.sha }}


.github/workflows/reposchutz.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+      issues: write
```
martinpitt, line 12: This *could* also be `pull-requests:`. The [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) does not make it clear whether this applies to the type of the API or target object (a PR is also an issue in that regard).

martinpitt:
My proposal: Let's land this, test it, and if it really requires the `pull-requests:` permission instead, do a follow-up. The regression risk is zero, and this is easier than setting up an entirely parallel deployment.

allisonkarlitskaya (approved):

PR #17250, opened 2022-04-13
Fully automatic npm-update workflow

allisonkarlitskaya:
Create a new `npm-update` environment that allows pushing to both the `cockpit` and `node-cache` repo.  Use the ssh key to force-push after opening the PR, hopefully triggering the tests.

Here's what the deploy keys on cockpit repo look like after running the modified `deploy-keys` script from this PR:

image


allisonkarlitskaya:
So: turns out that you can't use the same deploy key in more than one repository.

Back to the drawing board on that one... :(

PR #17253, opened 2022-04-13
Fully automatic npm-update

allisonkarlitskaya:
Do an extra force-push via SSH after opening the PR to trigger the tests to run.

See a test workflow run here:

https://github.com/cockpit-project/cockpit/actions/runs/2160854103

which opened this PR here:

https://github.com/cockpit-project/cockpit/pull/17252

and which reposchutz was able to successfully reproduce `npm install` for here:

https://github.com/cockpit-project/cockpit/runs/6006502062?check_suite_focus=true

### DIFF
--- a/.github/deploy-keys.sh
+++ b/.github/deploy-keys.sh
@@ -64,3 +64,11 @@ deploy_env node-cache
 # https://github.com/cockpit-project/org.cockpit_project.CockpitClient
 #   - update-flathub.yml
 deploy_env flathub cockpit-project/org.cockpit_project.CockpitClient
+
+# https://github.com/cockpit-project/cockpit (COCKPIT_DEPLOY_KEY)
+# https://github.com/cockpit-project/node-cache (NODE_CACHE_DEPLOY_KEY)
+#   - npm-update.yml
+#   - npm-update-pf.yml
+# Installed additionally to the above keys (+).
+deploy_env npm-update +cockpit-project/cockpit COCKPIT_DEPLOY_KEY
+deploy_env npm-update +cockpit-project/node-cache NODE_CACHE_DEPLOY_KEY
--- a/.github/deploy-keys.sh
+++ b/.github/deploy-keys.sh
@@ -31,11 +31,12 @@ fi
 deploy_env() {
     ENVIRONMENT="$1"
     DEPLOY_TO="${2:-${ORG}/${ENVIRONMENT}}"
+    SECRET_NAME="${3:-DEPLOY_KEY}"
 
     bots/github-upload-secrets $DRY_RUN \
         --receiver "${ORG}/${THIS}" \
         --env "${ENVIRONMENT}" \
-        --ssh-keygen DEPLOY_KEY \
+        --ssh-keygen "${SECRET_NAME}" \
         --deploy-to "${DEPLOY_TO}"
 }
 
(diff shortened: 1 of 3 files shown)


.github/deploy-keys.sh
```diff
@@ -64,3 +64,11 @@ deploy_env node-cache
 # https://github.com/cockpit-project/org.cockpit_project.CockpitClient
 #   - update-flathub.yml
 deploy_env flathub cockpit-project/org.cockpit_project.CockpitClient
+
+# https://github.com/cockpit-project/cockpit (COCKPIT_DEPLOY_KEY)
+# https://github.com/cockpit-project/node-cache (NODE_CACHE_DEPLOY_KEY)
+#   - npm-update.yml
+#   - npm-update-pf.yml
+# Installed additionally to the above keys (+).
+deploy_env npm-update +cockpit-project/cockpit COCKPIT_DEPLOY_KEY
```
martinpitt, line 73: ISTM that the first key should be generated/added without +, and thus clean/reset the env. Only the second one should use +. Or did I misunderstand this? It's probably not such a big deal, as I suppose old secrets with the same name will be replaced.

.github/workflows/npm-update-pf.yml
```diff
@@ -28,8 +28,17 @@ jobs:
...
           ssh-add -D
           ssh-agent -k
 
+      - name: Force push the change to trigger testing workflows
+        run: |
+          sleep 1 # make sure the timestamp changes
+          git commit --amend --no-edit
+          eval $(ssh-agent)
+          ssh-add - <<< '${{ secrets.COCKPIT_DEPLOY_KEY }}'
+          git push --force 'git@github.com:${{ github.repository }}' HEAD
```
martinpitt, line 42: There's a wart here: The initial PR creation from bots/npm-update will already queue/start our own tests, as we explicitly allowed the `github-actions` user into our contributors team. Our test machinery is clever enough to abort tests for obsolete PR SHAs, but it does create some noise. We could prevent that by creating the initial PR with no-test and drop the label here.

But this is not a blocker at all -- our current workflow has the same problem (except that tests fail very quickly due to the package.json vs. lock mismatch), so we can do this as an optimization later on if we care enough.

.github/workflows/npm-update-pf.yml
```diff
@@ -28,8 +28,17 @@ jobs:
...
           eval $(ssh-agent)
-          ssh-add - <<< '${{ secrets.DEPLOY_KEY }}'
+          ssh-add - <<< '${{ secrets.NODE_CACHE_DEPLOY_KEY }}'
           bots/npm-update @patternfly >&2
           ssh-add -D
           ssh-agent -k
 
+      - name: Force push the change to trigger testing workflows
+        run: |
+          sleep 1 # make sure the timestamp changes
```
martinpitt, line 38: Which timestamp? Of the git commit? I.e. to ensure a new SHA?

martinpitt (approved):
Amazing, thank you! With that, this whole mechanism is now comfortable enough to port it to our other projects. I think c-machines would be the next biggest beneficiary. As you asked, I think we should hold off on starter-kit, though -- setting all this up is still more overhead than small projects may be willing to spend, and it raises the barrier of entry quite a bit.

.github/workflows/npm-update-pf.yml
allisonkarlitskaya: Exactly.

.github/workflows/npm-update-pf.yml
allisonkarlitskaya: OK.

PR #17254, opened 2022-04-14
storage: Don't complain about missing fstab entry for "/"

### LINKED ISSUE #17231: Cockpit Doesn't Recognize Root FS Mounted Via Kernel Command Line Only
### Explain what happens

1. Make sure the `root*` kernel command line parameters are used to mount the rootfs.
2. Remove rootfs mount in `/etc/fstab` if present.
3. Reboot if changes were made.

The kernel command line in use on my Pi from `/proc/cmdline`:
```
root=PARTUUID=642e064b-02 rootfstype=btrfs rootflags=noatime,compress-force=zstd rootwait rw loglevel=4 cosnoleblank=0 net.ifnames=0 ipv6.disable=1 smsc95xx.macaddr=[redacted]
```

This results in the :warning: symbol showing up for the storage device.  Drilling down shows the issue Cockpit thinks is at play:

Screenshot_20220408_155633-1

IIRC systemd recommends using kernel command line parameters to mount the rootfs and not to do so in `/etc/fstab` since doing both results in the root filesystem being mounted twice.

Side note:  Offering to unmount root is probably A Bad Idea:tm: and should probably be avoided.

### Version of Cockpit

266-1

Storage

### Server operating system

Arch Linux

rolling

Chrome

### System log

_No response_

mvollmer:
Fixes #17231


pkg/storaged/fsys-tab.jsx
```diff
@@ -159,7 +159,10 @@ export function check_mismounted_fsys(client, path, enter_warning) {
         else if (is_mounted && opt_noauto && !opt_noauto_intent && !opt_systemd_automount)
             type = "no-mount-on-boot";
     } else if (other_mounts.length > 0) {
-        type = "mounted-no-config";
+        // We don't complain about the rootfs, it's probably
+        // configured somewhere else, like in the bootloader.
+        if (other_mounts[0] != "/")
```
martinpitt, line 164: Is it not possible that there are multiple entries, and / is only the second or later entry?

mvollmer: Yes maybe, but almost all of the current Cockpit storage page only looks at the first entry, since the UDisks2 API can only unmount that one anyway, and the world is just a simpler place if we close our eyes to multiple mount points... :-)

martinpitt: Ack, thanks! Can you please rebase this PR, so that we get over the "protection checks" failure? I'll approve then.

mvollmer: Done!

martinpitt (approved):
Danke!

martinpitt:
The [f35 packit failure](https://artifacts.dev.testing-farm.io/84e86bd2-2938-469e-add1-ee06cc45c38d/work-all5SDroa/log.txt) was just fixed in #17260.

PR #17255, opened 2022-04-14
test: make testNfsClient clean up properly

jelly:
When the testNfsClient flakes and a nfs share is mounted to /mnt/test
the cleanup first restores /etc/exports and then restarts nfs-server.
This causes issues with umount'ing the share and gives permission errors
when trying to remove the share. Making sure the share is umount'ed and
directory removed before restoring /etc/exports and restarting
nfs-server.

### DIFF
--- a/test/verify/check-storage-nfs
+++ b/test/verify/check-storage-nfs
@@ -33,11 +33,13 @@ class TestStorageNfs(StorageCase):
         self.login_and_go("/storage")
 
         m.execute("mkdir /home/foo /home/bar /mnt/test")
-        # Removing the nfs mount removes the target dir.
-        self.addCleanup(m.execute, "if [ -e /mnt/test ]; then umount /mnt/test 2>/dev/null || true; rm -r /mnt/test; fi")
         self.write_file("/etc/exports", "/home/foo 127.0.0.0/24(rw)\n/home/bar 127.0.0.0/24(rw)\n",
                         post_restore_action="systemctl restart nfs-server")
         m.execute("systemctl restart nfs-server")
+        # Removing the nfs mount removes the target dir, if the test fails it
+        # won't. It's important to umount before restoring /etc/exports as
+        # otherwise nfs is confused and we can't umount the share.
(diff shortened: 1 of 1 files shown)


martinpitt:
[two](https://logs.cockpit-project.org/logs/pull-17255-20220414-153337-8c102764-fedora-35-firefox/log.html#307) 
[NFS test failures](https://logs.cockpit-project.org/logs/pull-17255-20220414-153337-8c102764-ubuntu-2204/log.html#307), retrying to compare. The fix makes sense. Thank you!

martinpitt:
Failed again, sorry. I don't think it's related to this cleanup fix, but just uncovers the existing flake as the test isn't retried any more?

martinpitt:
I see the same failure in the 267 release gating test: https://bodhi.fedoraproject.org/updates/FEDORA-2022-70c1d3d182 . It's not a regression, but we still need to address it.

martinpitt:
In #17258 it also failed on [c8s](https://logs.cockpit-project.org/logs/pull-17258-20220419-033103-417fbe13-centos-8-stream/log.html#206) and [r90](https://logs.cockpit-project.org/logs/pull-17258-20220419-033103-417fbe13-rhel-9-0/log.html#206), so it's not specific to Firefox either. 

I can't reproduce a different failure locally by only running that test, but it does fail with these two:

    TEST_OS=fedora-35 test/verify/check-storage-nfs TestStorageNfs.testNfs{Busy,Client} $RUNC -tv

The dialog action seems to happen, as the journal says

    rpc.mountd[10213]: refused mount request from 127.0.0.1 for /usr/share (/): not exported

and the test log says

    > warning: mount.nfs: mounting 127.0.0.1:/usr/share failed, reason given by server: No such file or directory

So this is an actual bug in that dialog, it does not always pick up the error message.

As this is in main, and breaks everything, I land this with red (with a heavy heart). I'll split this out into a new issue.

martinpitt (approved):

PR #17267, opened 2022-04-20
systemd: Updates for the Insights status

mvollmer:
The mock-insights server and the client config for it have been
updated from here:

   https://github.com/candlepin/subscription-manager/pull/2998
   https://github.com/candlepin/subscription-manager/pull/2996

The /var/lib/insights directory needs to be read as superuser, so the
InsightsStatus component needs to re-open its file watches
accordingly.

### DIFF
--- a/test/verify/files/mock-insights
+++ b/test/verify/files/mock-insights
@@ -10,17 +10,19 @@
 #
 # You need these in your insights-client.conf:
 #
-# gpg=False
 # auto_config=False
-# base_url=127.0.0.1:8888/r/insights
-# insecure_connection=True
+# auto_update=False
+# base_url=127.0.0.1:8443/r/insights
+# cert_verify=/var/lib/insights/mock-certs/ca.crt
 # username=admin
 # password=foobar
 
 from http.server import *
 import json
+import os
 import re
 import ssl
+import subprocess
 
 systems = {}
 
@@ -36,12 +38,6 @@ class handler(BaseHTTPRequestHandler):
             self.send_response(200)
             self.end_headers()
             self.wfile.write(b"lub-dup")
-            return
-
-        m = self.match("/r/insights/v1/static/core/insights-core.egg")
-        if m:
-            self.send_response(304)
-            self.end_headers()
             return
 
         m = self.match("/r/insights/v1/static/uploader.v2.json")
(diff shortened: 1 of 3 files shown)


mvollmer:
This still needs the naughty to get over the SELinux denials.

martinpitt:
Did you try this (locally) with `setenforce 0`? I did yesterday with sub-man, and it was still failing, just differently. As I mentioned, SELinux breaks this, but it's not the only reason. Somehow our mock setup is still wrong.

test/verify/files/mock-insights
```diff
@@ -138,9 +134,22 @@ class handler(BaseHTTPRequestHandler):
 
 
 def insights_server(port):
+    # Let's put the certs into /var/lib/insights so that SELinux
+    # allows insights-client to actually read ca.crt when running as a
+    # systemd service.
+    certdir = "/var/lib/insights/mock-certs"
+    if not os.path.exists(certdir):
+        os.makedirs(certdir)
+        subprocess.check_call(["sscg"], cwd=certdir)
```
martinpitt, line 143: This might possibly require a `restorecon` call afterwards, to make them readable for insights?

mvollmer:
> [...] and it was still failing, just differently.

The test passes now on rhel-8-6, except for the unexpected journal messages.

The /var/lib/insights directory is `rwxr-x---` now, which makes the files in it inaccessible for no-root, independent of selinux. This might be a bug in the insights-client rpm, or not. *shrug*

test/verify/files/mock-insights
mvollmer: It doesn't seem to be necessary.

martinpitt (approved):
Nice, thank you! This fails on rhel-9 due to the unexpected message for https://github.com/cockpit-project/bots/issues/3042. We need to relax the pattern (it is currently written for sub-man).

mvollmer:
> Nice, thank you! This fails on rhel-9 due to the unexpected message for [cockpit-project/bots#3042](https://github.com/cockpit-project/bots/issues/3042). We need to relax the pattern (it is currently written for sub-man).

I have a look at subman, it probably runs into the same problem with reading the files in /var/lib/insights/, and we can fix the test failure there as well.

martinpitt (approved):
Whee, thank you!

PR #17268, opened 2022-04-20
test: Reduce maximum line length to 200

martinpitt:
With that, our test/static-code is getting the same maximum length as cockpit-machines, and we can share the script.

 - Builds on top of #16944

### DIFF
--- a/test/verify/check-pages
+++ b/test/verify/check-pages
@@ -23,6 +23,8 @@ import os
 import parent
 from testlib import *
 
+RHEL_DOC_BASE = "https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console"
+
 
 @nondestructive
 @skipDistroPackage()
@@ -111,12 +113,14 @@ OnCalendar=daily
         b.switch_to_top()
         self.checkDocs(["Managing services"])
         b.click("#toggle-docs")
-        b.wait_visible('#toggle-docs + ul a:contains("Managing services")[href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/managing-services-in-the-web-console_system-management-using-the-rhel-8-web-console"]')
-        b.wait_visible('#toggle-docs + ul a:contains("Web Console")[href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/index"]')
+        b.wait_visible(f'#toggle-docs + ul a:contains("Managing services")[href="{RHEL_DOC_BASE}/'
+                       'managing-services-in-the-web-console_system-management-using-the-rhel-8-web-console"]')
+        b.wait_visible(f'#toggle-docs + ul a:contains("Web Console")[href="{RHEL_DOC_BASE}/index"]')
         b.click("#toggle-docs")
         b.wait_not_present("#toggle-docs + ul")
         b.go("/network")
-        self.checkDocs(["Managing networking bonds", "Managing networking teams", "Managing networking bridges", "Managing VLANs", "Managing firewall"])
+        self.checkDocs(["Managing networking bonds", "Managing networking teams",
+                        "Managing networking bridges", "Managing VLANs", "Managing firewall"])
         b.go("/system/services")
 
         m.restart_cockpit()
--- a/test/verify/packagelib.py
+++ b/test/verify/packagelib.py
@@ -16,6 +16,7 @@
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import textwrap
 import warnings
 
 from testlib import *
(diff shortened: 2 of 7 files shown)


test/verify/check-pages
```diff
@@ -23,6 +23,9 @@ import os
 import parent
 from testlib import *
 
+RHEL_DOC_BASE = ("https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/"
```
allisonkarlitskaya, line 26: Maybe we can better just `#noqa` this?

test/verify/packagelib.py
```diff
@@ -156,13 +158,15 @@ def createDeb(self, name, version, depends, postinst, install, content, arch, pr
...
-        cmd = '''mkdir -p /tmp/b/DEBIAN {repo}
-                 printf "Package: {name}\nVersion: {ver}\nPriority: optional\nSection: test\nMaintainer: foo\nDepends: {deps}\nArchitecture: {arch}\nDescription: dummy {name}\n{provides}\n" > /tmp/b/DEBIAN/control
-                 {post}
-                 touch /tmp/b/stamp-{name}-{ver}
-                 dpkg -b /tmp/b {deb}
-                 rm -r /tmp/b
-                 '''.format(name=name, ver=version, deps=depends, deb=deb, post=postinstcode, repo=self.repo_dir, arch=arch, provides=provides)
+        cmd = f'''mkdir -p /tmp/b/DEBIAN {self.repo_dir}
+                  printf "Package: {name}\nVersion: {version}\nPriority: optional\nSection: test
+Maintainer: foo\nDepends: {depends}\nArchitecture: {arch}
```
allisonkarlitskaya, line 163: I had earlier rewritten this in a very nice way.  Something like:

```
    printf "%s: %s\n" \
        Maintainer  'foo'
        Depends '...' \
        .... \
    > /tmp/...
```

which allows keeping everything neatly intended.

test/verify/storagelib.py
```diff
@@ -445,7 +445,14 @@ def dialog_with_retry(self, trigger, values, expect):
         self.dialog_wait_close()
 
     def udisks_objects(self):
-        return json.loads(self.machine.execute("""python3 -c 'import dbus, json; print(json.dumps(dbus.SystemBus().call_blocking("org.freedesktop.UDisks2", "/org/freedesktop/UDisks2", "org.freedesktop.DBus.ObjectManager", "GetManagedObjects", "", [])))'"""))
+        return json.loads(self.machine.execute(
+            """python3 -c '
+import dbus, json
```
allisonkarlitskaya, line 450: Can you indent that?

test/verify/storagelib.py
```diff
@@ -445,7 +445,14 @@ def dialog_with_retry(self, trigger, values, expect):
         self.dialog_wait_close()
 
     def udisks_objects(self):
-        return json.loads(self.machine.execute("""python3 -c 'import dbus, json; print(json.dumps(dbus.SystemBus().call_blocking("org.freedesktop.UDisks2", "/org/freedesktop/UDisks2", "org.freedesktop.DBus.ObjectManager", "GetManagedObjects", "", [])))'"""))
+        return json.loads(self.machine.execute(
+            """python3 -c '
```
allisonkarlitskaya, line 449: imho this belongs on the previous line

test/verify/check-pages
martinpitt: I'm not a fan of `#noqa` to second-guess our policy.  However, without the line break the line is 141 characters, so it's actually not critical for *this* PR. And this conflicts with "breaking up URLs is evil", so I'll just re-join this for now.

test/verify/storagelib.py
martinpitt: Does this work for you?
```py
    def udisks_objects(self):
        return json.loads(self.machine.execute(["python3", "-c", textwrap.dedent("""
            import dbus, json
            print(json.dumps(dbus.SystemBus().call_blocking(
                "org.freedesktop.UDisks2",
                "/org/freedesktop/UDisks2",
                "org.freedesktop.DBus.ObjectManager",
                "GetManagedObjects", "", [])))""")]))
```

We could alternatively also stop calling python completely, and use `busctl -j call` instead.

martinpitt: Pro: This looks much nicer:
```py
    def udisks_objects(self):
        reply = self.machine.execute(["busctl", "call", "--json=pretty",
                                      "org.freedesktop.UDisks2", "/org/freedesktop/UDisks2",
                                      "org.freedesktop.DBus.ObjectManager", "GetManagedObjects"])
        return json.loads(reply)["data"][0]
```

Con: RHEL 8's systemd is too old to have a `--json` option :crying_cat_face:

test/verify/packagelib.py
martinpitt: I don't like this that much, as it requires a long list of `%s: %s` templates, and keep their number in sync with the fields. I used a slightly different form now with dedent() and m.write().

allisonkarlitskaya:
```
$ printf "%s: %s\n" a b c d e f g h
a: b
c: d
e: f
g: h
```

martinpitt: TIL! However, I honestly don't feel like cycling this again, unless you insist. IMHO dedent() is quite decent as well.

allisonkarlitskaya (approved):

test/verify/packagelib.py
allisonkarlitskaya: Sure.  I agree that it looks fine now with dedent.  Maybe even better.

PR #17270, opened 2022-04-21
workflows/npm-update{,-pf}.yml: don't force push main

allisonkarlitskaya:
In case no update happened, we really ought to be skipping the
force-push.  Check the current branch name, and don't do a force push if
it's main.


allisonkarlitskaya:
eh, this will work, but I actually want to do better.

PR #17271, opened 2022-04-21
workflows/npm-update{,-pf}.yml: don't force push main

allisonkarlitskaya:
In case no update happened, we really ought to be skipping the
force-push.  Check the current branch name, and don't do a force push if
it's main.


allisonkarlitskaya:
Working skip: https://github.com/cockpit-project/cockpit/runs/6111407795?check_suite_focus=true

martinpitt (approved):
Danke!

PR #17272, opened 2022-04-21
tools/node-modules: use bots/npm install

allisonkarlitskaya:
Drop our internal copy of the code for launching npm inside of the
container and use the version of it that now lives in the bots.

 - [x] cockpit-project/bots#3270

### DIFF
--- a/.github/workflows/npm-update-pf.yml
+++ b/.github/workflows/npm-update-pf.yml
@@ -12,11 +12,6 @@ jobs:
       contents: write
     runs-on: ubuntu-20.04
     steps:
-      - name: Set up dependencies
-        run: |
-          sudo apt update
-          sudo apt install -y npm
-
       - name: Clone repository
         uses: actions/checkout@v2
 
--- a/.github/workflows/npm-update.yml
+++ b/.github/workflows/npm-update.yml
@@ -12,11 +12,6 @@ jobs:
       contents: write
     runs-on: ubuntu-20.04
     steps:
-      - name: Set up dependencies
-        run: |
-          sudo apt update
-          sudo apt install -y npm
-
       - name: Clone repository
         uses: actions/checkout@v2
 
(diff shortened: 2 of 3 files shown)


martinpitt:
Looks good, but the protection checks are unhappy:

> fatal: path 'node_modules' exists on disk, but not in '0081e0ad4b63ceba86067cdd1433fba7c71d90ee'

Quoi ?

allisonkarlitskaya:
> Looks good, but the protection checks are unhappy:
> 
> > fatal: path 'node_modules' exists on disk, but not in '0081e0ad4b63ceba86067cdd1433fba7c71d90ee'
> 
> Quoi ?

`--depth 1` damage.

In short: we use `--not` on the base sha to limit the commits we check for consistency.  But if the base sha is not a parent of the head sha (ie: PR not fully rebased), we won't have its full history so we won't see the mege point, and we'll continue forever.

0081e0ad4b63ceba86067cdd1433fba7c71d90ee happens to be

> tools: Move package.json and node_modules to top level

from Stef, in 2015.

I'll fix this in a separate commit by dropping `--depth 1`.

allisonkarlitskaya:
I did #17275 to fix this.  I'll rebase this and push it again, which should work.

martinpitt (approved):
Thanks! This is a PR from your fork, so we can't easily trigger an npm-update workflow against this branch. But let's do it after landing, to ensure it works.

allisonkarlitskaya:
> Thanks! This is a PR from your fork, so we can't easily trigger an npm-update workflow against this branch. But let's do it after landing, to ensure it works.

https://github.com/cockpit-project/cockpit/runs/6126079573?check_suite_focus=true

allisonkarlitskaya:
> https://github.com/cockpit-project/cockpit/runs/6126079573?check_suite_focus=true

https://github.com/cockpit-project/cockpit/pull/17276

PR #17277, opened 2022-04-22
configure.ac: use AC_CHECK_TOOL to find ar

allisonkarlitskaya:
This makes sure we get the executable with the correct `--host` prefix,
allowing us to successfully cross-compile.  Without this, `AR` gets
hard-coded to `ar`.


martinpitt (approved):
Danke!

PR #17279, opened 2022-04-25
nosmt code and test cleanup, work around kernel-install bugs

martinpitt:
Removing a kernel leaves behind its BLS entry [1], and `kernel-install remove`
leaves GRUB's default kernel setting to the just removed kernel [2].
Both bugs together cause a boot failure when removing old kernels [3].
Work around that by restoring grub's defaults after the fake kernel
install/removal.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2078379
[2] https://bugzilla.redhat.com/show_bug.cgi?id=2078359
[3] https://github.com/cockpit-project/bots/pull/3293

 - Requires https://github.com/cockpit-project/bots/pull/3296

### DIFF
--- a/pkg/systemd/kernelopt.sh
+++ b/pkg/systemd/kernelopt.sh
@@ -29,21 +29,8 @@ grub() {
     if type grubby >/dev/null 2>&1; then
         if [ "$1" = set ]; then
             grubby --args="$2" --update-kernel=ALL
-            # HACK: grubby on RHEL 8.0 does not change default kernel args (https://bugzilla.redhat.com/show_bug.cgi?id=1690765)
-            envopts=$(grub2-editenv - list | grep ^kernelopts) || envopts=""
-            if [ -n "$envopts" ]; then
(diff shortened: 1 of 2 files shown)


martinpitt:
Triggered an extra fedora-35 run against the refreshed image in https://github.com/cockpit-project/bots/pull/3293

martinpitt:
Darn -- the [grubby bug](https://bugzilla.redhat.com/show_bug.cgi?id=1690765)  that got fixed in RHEL 8 still affects [9.0](https://logs.cockpit-project.org/logs/pull-17279-20220425-071154-b56d753a-rhel-9-1/log.html#300) and [9.1](https://logs.cockpit-project.org/logs/pull-17279-20220425-071154-b56d753a-rhel-9-1/log.html#300). It's even worse, as our current workaround does not apply any more (so we can still drop it). The test now fails because it does not accept the /etc/default/grub option any more.

I reported this as https://bugzilla.redhat.com/show_bug.cgi?id=2078397. I'll create a naughty, and adjust the commit description. Blocked until then.

marusak (approved):
Thanks!

PR #17282, opened 2022-04-26
metrics: show podman container cpu usage

jelly:
## Metrics: Show Podman containers in top CPU and memory lists

Current metrics cards for CPU and memory now show [Podman](https://podman.io/) containers, in addition to systemd services. 

container memory usage

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -67,6 +67,9 @@ const _ = cockpit.gettext;
 // format Date as YYYY-MM-DD HH:mm:ss UTC which is human friendly and systemd compatible
 const formatUTC_ISO = t => `${t.getUTCFullYear()}-${t.getUTCMonth() + 1}-${t.getUTCDate()} ${t.getUTCHours()}:${t.getUTCMinutes()}:${t.getUTCSeconds()} UTC`;
 
+// podman's containers cgroup
+const podmanCgroupRe = /libpod-(?<containerid>[a-z|0-9]{64})\.scope$/;
+
 // keep track of maximum values for unbounded data, so that we can normalize it properly
 // pre-init them to avoid inflating noise
 let scaleSatCPU = 4;
@@ -377,7 +380,7 @@ class CurrentMetrics extends React.Component {
             }
         }
 
-        // return [ { [key, value, is_user] } ] list of the biggest n values
+        // return [ { [key, value, is_user, is_container] } ] list of the biggest n values
         function n_biggest(names, values, n) {
             const merged = [];
             names.forEach((k, i) => {
@@ -387,21 +390,48 @@ class CurrentMetrics extends React.Component {
                     const is_user = k.match(/^user.*user@\d+\.service.+/);
                     const label = k.replace(/.*\//, '').replace(/\.service$/, '');
                     // only keep cgroup basenames, and drop redundant .service suffix
-                    merged.push([label, v, is_user]);
+                    merged.push([label, v, is_user, false]);
+                }
+                // filter out podman containers
+                const matches = k.match(podmanCgroupRe);
+                if (matches && v) {
+                    // truncate to 12 chars like the podman output
+                    const containerid = matches.groups.containerid.substr(0, 12);
+                    const is_user = k.match(/^user.slice/);
+                    merged.push([containerid, v, is_user, true]);
                 }
             });
             merged.sort((a, b) => b[1] - a[1]);
             return merged.slice(0, n);
         }
 
-        function serviceRow(name, value, is_user) {
-            const name_text = (
-                <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : "")) }>
+        function cgroupClickHandler(name, is_user, is_container) {
+            if (is_container) {
+                cockpit.jump("/podman");
+            } else {
+                cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : ""));
+            }
+        }
+
+        function cgroupRow(name, value, is_user, is_container) {
+            const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+            let name_text = (
+                <Button variant="link" isInline component="a" key={name}
+                        onClick={() => cgroupClickHandler(name, is_user, is_container)}
+                        isDisabled={is_container && !podman_installed}>
                     <TableText wrapModifier="truncate">
-                        {name}
+                        {is_container ? _("pod") + " " + name : name}
                     </TableText>
                 </Button>
             );
+            if (is_container && !podman_installed) {
+                name_text = (
+                    <Tooltip content={_("cockpit-podman is not installed")} key={name + "_tooltip"}>
+                        <div>
+                            {name_text}
+                        </div>
+                    </Tooltip>);
+            }
             const value_text = <TableText wrapModifier="nowrap">{value}</TableText>;
             return {
                 cells: [{ title: name_text }, { title: value_text }]
@@ -410,11 +440,11 @@ class CurrentMetrics extends React.Component {
 
         // top 5 CPU and memory consuming systemd units
         newState.topServicesCPU = n_biggest(this.cgroupCPUNames, this.samples[9], 5).map(
-            ([key, value, is_user]) => serviceRow(key, Number(value / 10 / numCpu).toFixed(1), is_user) // usec/s → percent
+            ([key, value, is_user, is_container]) => cgroupRow(key, Number(value / 10 / numCpu).toFixed(1), is_user, is_container) // usec/s → percent
         );
 
         newState.topServicesMemory = n_biggest(this.cgroupMemoryNames, this.samples[10], 5).map(
-            ([key, value, is_user]) => serviceRow(key, cockpit.format_bytes(value), is_user)
+            ([key, value, is_user, is_container]) => cgroupRow(key, cockpit.format_bytes(value), is_user, is_container)
         );
 
         this.setState(newState);
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -11,6 +11,7 @@ import parent
 import packagelib
 from testlib import *
 
+from machine_core import ssh_connection
 from machine_core.constants import TEST_OS_DEFAULT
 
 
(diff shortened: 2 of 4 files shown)


jelly:
So my assumption is not 100% correct, not all podman versions? crun? cgroups? have a /container

```
└─machine.slice (#74)
  ├─libpod-conmon-7ef2498ef1256a46a5392f23b74a35d13bc94c6c87180241acfac1523b1d0b11.scope (#36571)
  │ └─493955 /usr/bin/conmon --api-version 1 -c 7ef2498ef1256a46a5392f23b74a35d13bc94c6c87180241acfac1523b1d0b11 -u 7ef2498ef1256a46a5392f23b74a35d13bc94c6c87180241acfac1>
  └─libpod-7ef2498ef1256a46a5392f23b74a35d13bc94c6c87180241acfac1523b1d0b11.scope (#36637)
    └─container (#36703)
      └─493958 /bin/bash
```
So we need to adjust the regex for that.

jelly:
@martinpitt the coreos tests fail because we lack a busybox image (for podman tests usually) on fedora-coreos. Do we want to add that to our image and in the future also test podman on coreos? Or shall I skip it?

martinpitt:
@jelly Funny you should ask, I'm working on enabling fcos tests for c-podman as we speak. However, don't block on that, just skip that check on coreos for now.

pkg/metrics/metrics.jsx
```diff
@@ -387,16 +390,33 @@ class CurrentMetrics extends React.Component {
...
                     // only keep cgroup basenames, and drop redundant .service suffix
-                    merged.push([label, v, is_user]);
+                    merged.push([label, v, is_user, false]);
+                }
+                // filter out podman containers
+                const matches = k.match(podmanSystemRe);
+                if (matches && v) {
+                    // truncate to 12 chars like the podman output
+                    const containerid = matches.groups.containerid.substr(0, 12);
+                    const is_user = k.match(/^user.*user@\d+\.service.+/);
```
martinpitt, line 400: Is that a copy&paste error, or deliberate? user sessions should be handled independently from container scopes.

pkg/metrics/metrics.jsx
```diff
@@ -387,16 +390,33 @@ class CurrentMetrics extends React.Component {
...
+                    merged.push([containerid, v, is_user, true]);
                 }
             });
             merged.sort((a, b) => b[1] - a[1]);
             return merged.slice(0, n);
         }
 
-        function serviceRow(name, value, is_user) {
+        function serviceClickHandler(name, is_user, is_container) {
```
martinpitt, line 408: Let's rename "service" to "cgroup" then, as these are now not just systemd services any more.

test/verify/check-metrics
```diff
@@ -760,6 +761,40 @@ class TestCurrentMetrics(MachineCase):
...
+            container_sha = m.execute(f"podman inspect --format '{{{{.Id}}}}' {container_name}").strip()
+            shortid = container_sha[:12]
+
+            b.wait_in_text("#current-metrics-card-cpu", f"pod {shortid}")
+            b.wait(lambda: topServiceValue(self, "Top 5 CPU services", "%", 1) > 70)
+            m.execute(f"podman stop -t 0 {container_name}")
+
+            # copy images for user podman tests; podman insists on user session
+            m.execute("""
+loginctl enable-linger $(id -u admin)
```
martinpitt, line 778: You shouldn't need the linger here -- This happens during a running cockpit session for admin, i.e. admin's user bus should already be running.

pkg/metrics/metrics.jsx
```diff
@@ -387,16 +390,33 @@ class CurrentMetrics extends React.Component {
...
+            if (is_container && cockpit.manifests && cockpit.manifests.podman) {
+                cockpit.jump("/podman");
```
martinpitt, line 410: This needs to handle the "container && !manifests.podman" case separately -- it can't be a link to the services page then.

martinpitt (changes requested):
Thanks!

pkg/metrics/metrics.jsx
jelly: Ok

pkg/metrics/metrics.jsx
jelly: I want to be able to identify user containers so we could possibly link to the them later when cockpit-podman supports some sort of filtering via url query parameters:

User containers appear as the following:
```
'user.slice/user-1000.slice/user@1000.service/user.slice/libpod-30698d4aecca2fa80938b86d5df7b726fd6288e23cb49293eb0b63a9a4c3fdba.scope'
```

pkg/metrics/metrics.jsx
jelly: Yes, I think we don't want it to be a link then as that looks confusing, maybe we even want to show a tooltip that users should install cockpit-podman?

pkg/metrics/metrics.jsx
martinpitt: Makes sense, thanks!

pkg/metrics/metrics.jsx
jelly: Hmm turns out that is not too simple as there is some issue with the just adding a <Tooltip> around the button. It seems that react is being "smart" by drawing the <tooltip> once and then just changing the text which means tooltips initially show up but after the data changes it does not.

jelly:
RHEL-9-1 lacks an selinux naughty:

```
audit: type=1400 audit(1652094777.130:4): avc:  denied  { ioctl } for  pid=36256 comm="iptables" path="/var/lib/containers/storage/overlay/eb0ae146941864f0a374bfb57a3feeb94b82dc2ef989a386f8ed4e01fa311dd0/merged" dev="overlay" ino=50513020 scontext=unconfined_u:system_r:iptables_t:s0-s0:c0.c1023 tcontext=system_u:object_r:container_file_t:s0:c133,c341 tclass=dir permissive=0

```

jelly:
On rhel-8-6 (just like centos-8-stream) user containers don't show up as libpod-$containerid but:
```
│ └─user-1000.slice
│   ├─user@1000.service
│   │ ├─user.slice
│   │ │ └─podman-3679.scope
│   │ │   ├─3679 podman run --rm -ti busybox /bin/sh
│   │ │   ├─3702 /usr/bin/slirp4netns --disable-host-loopback --mtu=65520 --enable-sandbox --enable-seccomp --enable-ipv6 -c -e 3 -r 4 --netns-type=path /run/use>
│   │ │   ├─3708 /usr/bin/conmon --api-version 1 -c bf70e4ca2c315dd9bd8afe18dceecd9d681780503a32bf4abff6186e04b0ec7b -u bf70e4ca2c315dd9bd8afe18dceecd9d681780503>
│   │ │   └─bf70e4ca2c315dd9bd8afe18dceecd9d681780503a32bf4abff6186e04b0ec7b
│   │ │     └─3717 /bin/sh
```

@martinpitt is not supporting this on RHEL 8 a possibility? I'll have to do some thinking on what causes this and how we can support it if we want too.

martinpitt:
> is not supporting this on RHEL 8 a possibility

Yes, of course. It's fine to keep the status quo there and not show containers.

test/verify/check-metrics
jelly: linger was dropped.

pkg/metrics/metrics.jsx
```diff
@@ -67,6 +67,9 @@ const _ = cockpit.gettext;
 // format Date as YYYY-MM-DD HH:mm:ss UTC which is human friendly and systemd compatible
 const formatUTC_ISO = t => `${t.getUTCFullYear()}-${t.getUTCMonth() + 1}-${t.getUTCDate()} ${t.getUTCHours()}:${t.getUTCMinutes()}:${t.getUTCSeconds()} UTC`;
 
+// podman's system containers cgroup
+const podmanSystemRe = /libpod-(?<containerid>[a-z|0-9]{64})\.scope$/;
```
martinpitt, line 71: It's also being used for user containers, just inside `user.slice` instead of `system.slice`. So the "System" bit is a bit misleading.

pkg/metrics/metrics.jsx
```diff
@@ -387,21 +390,50 @@ class CurrentMetrics extends React.Component {
...
-                <Button variant="link" isInline component="a" key={name} onClick={ e => cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : "")) }>
+        function cgroupClickHandler(name, is_user, is_container) {
+            if (is_container) {
+                cockpit.jump("/podman");
+            } else {
+                cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : ""));
+            }
+        }
+
+        function serviceRow(name, value, is_user, is_container) {
```
martinpitt, line 416: Let's rename to cgroupRow for consistency.

pkg/metrics/metrics.jsx
```diff
@@ -387,21 +390,50 @@ class CurrentMetrics extends React.Component {
...
+            const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+            const original_name = name;
+            name = is_container ? _("pod") + " " + name : name;
```
martinpitt, line 419: Renaming `name` is a bit confusing when jumping around with `*`/`#` and such. Perhaps leave it alone, and do a `const label = ...` instead?

test/verify/check-metrics
```diff
@@ -760,6 +761,45 @@ class TestCurrentMetrics(MachineCase):
         b.wait(lambda: float(b.text("#load-avg .pf-l-flex div:first-child").split()[-1].rstrip(',')) > 15)
         m.execute("systemctl stop load-hog 2>/dev/null || true")  # ok to fail, as the command exits by itself
 
+        # Test podman containers
+        if m.image != "fedora-coreos":
```
martinpitt, line 765: Please add a `# HACK` comment -- this is just because of the missing busybox image on the fcos image, and offline test VM prep, right? It should generally work on FCOS. Hence, invert the condition and do `return`, so that we don't have to outdent the whole thing later on.

test/verify/check-metrics
```diff
@@ -760,6 +761,45 @@ class TestCurrentMetrics(MachineCase):
...
+
+            # On some test images the container takes a while to show up
+            with b.wait_timeout(300):
+                b.wait_in_text("#current-metrics-card-cpu", f"pod {shortid}")
+            b.wait(lambda: topServiceValue(self, "Top 5 CPU services", "%", 1) > 70)
+            m.execute(f"podman stop -t 0 {container_name}")
+
+            # RHEL-8 / CentOS-8's podman user containers do not show up as
+            # libpod-$containerid but as podman-3679.scope.
+            if m.image != "centos-8-stream" and m.image != "rhel-8-6":
```
martinpitt, line 780: `not m.image.startswith("rhel-8")`, so that we don't have to keep bumping this.

test/verify/check-metrics
```diff
@@ -760,6 +761,45 @@ class TestCurrentMetrics(MachineCase):
...
+                # copy images for user podman tests; podman insists on user session
+                m.execute("""
+    podman save quay.io/libpod/busybox | sudo -i -u admin podman load
```
martinpitt, line 783: Curious that this works without having to set `$XDG_RUNTIME_DIR`, but I guess it now just falls back to /run/user/<uid>.

martinpitt (changes requested):
Thanks! Some remaining simple and small nits, but this looks good. Let's figure out the test which causes the /home/admin permission failure (most likely an unrelated new commit).

test/verify/check-metrics
jelly: Yes, hmm maybe we should try the same in cockpit-podman now?

pkg/metrics/metrics.jsx
jelly: Excellent point, let's rename it

pkg/metrics/metrics.jsx
jelly: This is also better in the future as the cgroupClickHandler will get the container id instead of "pod $containerid"

test/verify/check-metrics
jelly: Yes, there is no busybox image yet (maybe after someone adds support in cockpit-podman :) )

If we return we miss the user services tests which run on fedora-coreos.

martinpitt: Ah right, misread the diff -- then let's just add the hack comment.

martinpitt:
I made a screenshot and added a stub release note.

garrett:
I guess this also makes containers show up in memory usage?

Screenshot 2022-05-12 at 10-29-47 Overview - garrett@Bolt

I don't have a pod though; just a container, and this information isn't useful yet, except to know that something (I know it's a container, and I know which container: `fedora-toolbox-36`, but our user's won't know) is taking up that amount of RAM.

---

Aside: The amount of RAM seems odd. It says my user's total (I'm assuming that's what user@1000 is) is 21.9 GB, whereas 22.2 GB is available... and my system's RAM is 32 GB. Here's `free -h`:

```
               total        used        free      shared  buff/cache   available
Mem:            31Gi       7.3Gi       4.7Gi       2.3Gi        19Gi        21Gi
Swap:          8.0Gi          0B       8.0Gi
```

Is the process ram usage also showing cache too? That's not so useful, if so, is it?

garrett (approved):
The issues I brought up in the previous comment are meant to be considered outside of this PR.

We know that the naming and direct linking to containers is not implemented yet, and we'll get that done in a follow-up PR.

The RAM sizing is also something outside the scope of this PR, but I thought was an interesting observation.

As for this PR, I think it makes sense to merge it in, as it's definitely better to have a hint that there's a container spiking CPU and/or memory than not. We'll make it better soon, but that'll take further effort.

In other words: :+1:

PR #17289, opened 2022-04-27
metrics: two small cleanups

martinpitt:


marusak (approved):
Thanks! Was wondering about the `.stream()/.then()` change and it seems safe in our case, as journalctl will always finish up in this case.

martinpitt:
Right, see https://cockpit-project.org/guide/latest/cockpit-spawn.html -- .then() will always be called last (or .catch()).

PR #17290, opened 2022-04-27
metrics: Show logs link for events without journal messages

### LINKED ISSUE #16496: metrics: "No logs found" when expanding logs from a spike
When expanding spikes on the side, sometimes it says "No logs found". I've seen this before locally, but it showed up a few times in our tests.

Here's a screenshot from a participant in the UX study of the metrics page:

Screenshot from 2021-10-19 14-52-43_ L3

Wild guess: I think this is related to the issue where the time in the logs is shifted versus the time in the history:

- https://github.com/cockpit-project/cockpit/issues/16492

martinpitt:
The originating event (such as starting a service) may have happened
earlier than in the current minute. So show a "View detailed logs"
action link for this case as well, but show an increased time range (15
minutes before the event to 1 minute after it) and verbosity (debug
instead of info).

Fixes #16496

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -720,13 +720,24 @@ class MetricsMinute extends React.Component {
         out.render_day_header = () => { return null };
         const render = journal.renderer(out);
 
-        const logsUrl = `/system/logs/#/?priority=info&since=${encodeURIComponent(since)}&until=${encodeURIComponent(until)}&follow=false`;
-
         journalctl.stream(entries => {
             entries.forEach(entry => render.prepend(entry));
             render.prepend_flush();
         })
-                .then(() => this.setState({ logs: out.logs, logsUrl }));
+                .then(() => {
+                    let logsUrl;
+                    if (out.logs.length === 0) {
+                        // without logs, increase verbosity and time range (-15 mins to + 1 min)
+                        const since = formatUTC_ISO(new Date(timestamp - 15 * 60000));
+                        const until = formatUTC_ISO(new Date(timestamp + 60000));
+                        logsUrl = `/system/logs/#/?priority=debug&since=${encodeURIComponent(since)}&until=${encodeURIComponent(until)}&follow=false`;
+                    } else {
+                        // with logs, show the exact minute and same log level as on the metrics page
+                        logsUrl = `/system/logs/#/?priority=info&since=${encodeURIComponent(since)}&until=${encodeURIComponent(until)}&follow=false`;
+                    }
+
+                    this.setState({ logs: out.logs, logsUrl });
+                });
     }
 
     render() {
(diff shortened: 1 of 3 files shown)


marusak (dismissed):
Good from my side, but I faintly remember that I was told by @garrett that we should not show it in this case. @garrett can you please also take a look?

martinpitt:
Note that this is just a part of the fix -- the more useful part will be to look back another minute, and only then show "no logs found".

garrett:
So I see these log entries in the metrics page:

Screenshot 2022-05-03 at 10-46-48 Overview - garrett@Bolt

And I click "View all logs", which shows me more:

Screenshot 2022-05-03 at 10-47-02 Logs - garrett@Bolt

Yet, there are still more hidden, as "all" is info and above. There are debug entries too:

Screenshot 2022-05-03 at 10-47-22 Logs - garrett@Bolt

Do we want to exclude debug from "all" when viewing the detailed time range?

Is "all" in "View all logs" on the metrics page misleading?

garrett:
I don't have an entry without logs yet. The closest is one with just one entry... still waiting (as I had to mess with my system to get logging working). Apologies for not addressing this PR directly yet. (Hopefully it'll happen soon.)

garrett:
So, back to my point about some logs being hidden in the view and "all" logs: When there are spikes and its empty in metrics, is there going to be something in the logs page? By default... or even with debug and such? Or are they going to expand, see nothing but the link to view all, then be greeted by yet another empty view?

martinpitt:
Summary from IRC:
 - Change label to "View detailed logs"
 - Change time range to "15 minutes before event", to increase the chance of catching something relevant
 - Change log level to "debug"

marusak (dismissed):
Awesome, thanks!
From my side +1

garrett (changes requested):
Oh, you still have "View all logs" instead of "View detailed logs in the corner for the logs with content. I thought we agreed to change all instances of it. You're not viewing _all_ logs in either case.

Screenshot from 2022-05-10 14-30-15

Additionally, the "No logs found" is very space constrained. I'd recommend adding 1 spacer unit vertically, which is 16px padding to the top and bottom.

It would then look like this:

image

You should be able to add it by using a utility class (first: make sure the CSS is imported at the topmost CSS file for the page) and then use `pf-u-py-md` — which stands for "PatternFly - utility - padding Y-axis - medium unit".

pkg/metrics/metrics.jsx
```diff
@@ -771,14 +782,22 @@ class MetricsMinute extends React.Component {
...
             </div>;
 
             let body = " "; // Cannot be false-y, otherwise table does not show '>'
             if (this.state.expanded) {
                 body = <div className="cockpit-log-panel">
-                    {this.state.logs === null ? _("Loading...") : this.state.logs.length === 0 ? _("No logs found") : this.state.logs}
+                    {this.state.logs === null
+                        ? _("Loading...")
+                        : this.state.logs.length === 0
+                            ? <span className="pf-u-py-md">{ _("No logs found") }</span>
```
garrett, line 798: Whoops. In looking at it in action here, I think we actually want `-sm`. I'm sorry for suggesting `md` instead of `sm`.

Aside from that, it looks good to me!

garrett (changes requested):

martinpitt:
Spacing with `sm` instead of `md`, as pushed recently:

image

garrett (approved):
LGTM!! :+1: 

image

PR #17293, opened 2022-04-28
packagekit: order kpatches before security updates

jelly:
Order the install updates buttons based on amount of updates with
kpatches being the fewest.

image


martinpitt (approved):
Makes sense, and requested by @garrett, so formal approval (pending green, of course). Let's use tihs screenshot for the relnotes. Thanks!

garrett (approved):
Thanks! Looks good. :+1:

PR #17294, opened 2022-04-28
test: make the `reference-image` file optional

ptoscano:
`reference-image` is used only for pixel tests, and currently also only
in some cases (e.g. only for chrom* browsers and non-devel builds).

Simplify its usage a bit: read it only when needed, and consider its
lack non-fatal (situation handled already).

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1150,10 +1150,17 @@ class MachineCase(unittest.TestCase):
             machine = self.machine
         label = self.label() + "-" + machine.label
         pixels_label = None
-        with open(f'{TEST_DIR}/reference-image') as fp:
-            reference_image = fp.read().strip()
-        if machine.image == reference_image and os.environ.get("TEST_BROWSER", "chromium") == "chromium" and not self.is_devel_build():
-            pixels_label = self.label()
+        if os.environ.get("TEST_BROWSER", "chromium") == "chromium" and not self.is_devel_build():
+            try:
+                with open(f'{TEST_DIR}/reference-image') as fp:
+                    reference_image = fp.read().strip()
+            except FileNotFoundError:
+                # no "reference-image" file available; this most likely means that
+                # there are no pixel tests to execute
+                pass
+            else:
+                if machine.image == reference_image:
+                    pixels_label = self.label()
         browser = Browser(machine.web_address,
                           label=label, pixels_label=pixels_label, coverage_label=self.label() if coverage else None,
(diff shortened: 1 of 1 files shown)


test/common/testlib.py
```diff
@@ -1150,10 +1150,17 @@ def new_browser(self, machine=None, coverage=False):
...
-            reference_image = fp.read().strip()
-        if machine.image == reference_image and os.environ.get("TEST_BROWSER", "chromium") == "chromium" and not self.is_devel_build():
-            pixels_label = self.label()
+        if os.environ.get("TEST_BROWSER", "chromium") == "chromium" and not self.is_devel_build():
+            try:
+                with open(f'{TEST_DIR}/reference-image') as fp:
+                    reference_image = fp.read().strip()
+            except FileNotFoundError:
+                # no "reference-image" file available; this most likely means that
+                # there are no pixel tests to execute
```
martinpitt, line 1159: Well, that, or someone is trying to add some and wonders why tests silently succeed even though the pixels are missing or don't match.. I don't like this blanket error ignoring. What would work better is to delay reading `reference-image` to the first time assert_pixels() is called -- then we get the correct behaviour for both use cases. Unfortunately the code structure is a bit awkward for that (at least it was when I tried to "quickly" do this).

ptoscano: What you say is correct, and I thought about that too.
On the other hand, the same can happen already know for various reasons:
- the browser is not chrom*
- the build is not a devel one
- the image is different than the one in `reference-image` (which also means a typo in that file kills the whole testing)
- python-pil is not installed/available during the testing

Unfortunately I don't have a better solution for this...

martinpitt: Would it help you already if you keep the outer part (moving the open into the `if:` condition) and just drop the try/except? That is a good and unanimous fix.

ptoscano: Hm not sure, as I think I would still hit the general conditions (chromium is the default `TEST_BROWSER`, and development builds). It can still be a small optimization, I can submit that on its own.

mvollmer: > On the other hand, the same can happen already now for various reasons:

I agree with this.  And a missing reference-image file can only mean "no pixel tests in this repo", so that looks fine to me.

If we want to double check that pixel tests are actually running for the expected context, we need to invent something else, like a "expected to fail" pixel test.

mvollmer (approved):

PR #17298, opened 2022-05-02
shell: Use new protocol for TopNav dialogs

mvollmer:
This puts the dialog specification close to where it is opened, and
allows the removal of some ceremonial code.

### DIFF
--- a/pkg/lib/dialogs.jsx
+++ b/pkg/lib/dialogs.jsx
@@ -1,0 +1,72 @@
+/*
+ * This file is part of Cockpit.
+ *
+ * Copyright (C) 2022 Red Hat, Inc.
+ *
+ * Cockpit is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * Cockpit is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* DIALOG PRESENTATION PROTOCOL
+ *
+ * Example:
+ *
+ * import { WithDialogs, useDialogs } from "dialogs.jsx";
+ *
+ * const App = () =>
+ *   <WithDialogs>
+ *     <Page>
+ *       <ExampleButton />
+ *     </Page>
+ *   </WithDialogs>;
+ *
+ * const ExampleButton = () => {
+ *   const D = useDialogs();
+ *   return <Button onClick={() => D.show(<MyDialog />)}>Open dialog</Button>;
+ * };
+ *
+ * const MyDialog = () => {
+ *   const Dialogs = useDialogs();
+ *   return (
+ *     <Modal title="My dialog"
+ *            isOpen
+ *            onClose={D.close}>
+ *       <p>Hello!</p>
+ *     </Modal>);
+ * };
+ *
+ * This does two things: It maintains the state of whether the dialog
+ * is open, and it does it high up in the DOM, in a stable place.
+ * Even if ExampleButton is no longer part of the DOM, the dialog will
+ * stay open and remain useable.
+ */
+
+import React, { useState, useContext } from "react";
+
+export const DialogsContext = React.createContext();
+export const useDialogs = () => useContext(DialogsContext);
+
+export const WithDialogs = ({ children }) => {
+    const [dialog, setDialog] = useState(null);
+
+    const Dialogs = {
+        show: setDialog,
+        close: () => setDialog(null)
+    };
+
+    return (
+        <DialogsContext.Provider value={Dialogs}>
+            {children}
+            {dialog}
+        </DialogsContext.Provider>);
+};
--- a/pkg/lib/hooks.js
+++ b/pkg/lib/hooks.js
@@ -39,6 +39,8 @@ import deep_equal from "deep-equal";
  *
  * - useEvent: For reacting to events emitted by arbitrary objects.
  *
+ * - useInit: For running a function once.
+ *
  * - useDeepEqualMemo: A utility hook that can help with things that
  * need deep equal comparisons in places where React only offers
  * Object identity comparisons, such as with useEffect.
@@ -298,4 +300,28 @@ export function useEvent(obj, event, handler) {
         obj && obj.addEventListener(event, update);
         return () => obj && obj.removeEventListener(event, update);
     }, [obj, event, handler]);
+}
+
+/* - useInit(func, deps, comps)
+ *
+ * function Component(arg) {
+ *   useInit(() => {
+ *     cockpit.spawn([ ..., arg ]);
+ *   }, [arg]);
+ *
+ *   ...
+ * }
+ *
+ * The function will be called once during the first render, and
+ * whenever "arg" changes.
+ *
+ * "useInit(func, deps, comps)" is the same as "useObject(func, null,
+ * deps, comps)" but if you want to emphasize that you just want to
+ * run a function (instead of creating a object), it is clearer to use
+ * the "useInit" name for that.  Also, "deps" are optional for
+ * "useInit" and default to "[]".
+ */
+
+export function useInit(func, deps, comps) {
+    return useObject(func, null, deps || [], comps);
 }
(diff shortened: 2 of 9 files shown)


mvollmer:
This doesn't fix any bugs, but it's a clean and simple example of where I would like us to go.

Some PRs with larger benefits are in preparation.

pkg/lib/dialogs.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+import React, { useState, useContext } from "react";
+
+export const DialogsContext = React.createContext();
+export const useDialogs = () => useContext(DialogsContext);
+
+export const WithDialogs = ({ children }) => {
+    const [dialog, setDialog] = useState(null);
+
+    const D = {
```
martinpitt, line 62: Could this be called `DialogContext` or `RCDialog` ("react context") or so? In this library the name doesn't matter much, but having almost-global `D` variables in our page code is hard to read and prone to collisions.

pkg/shell/indexes.js
```diff
@@ -213,12 +214,9 @@ function MachinesIndex(index_options, machines, loader) {
...
 
         ReactDOM.render(
-            React.createElement(TopNav, {
-                index: index,
-                state: state,
-                machine: machine,
-                compiled: compiled,
-            }),
+            <WithDialogs>
+                <TopNav index={index} state={state} machine={machine} compiled={compiled} />
```
martinpitt, line 218: The file should be renamed to indexes.jsx then. Not that our build chain would care, but for consistency, at least until we decide to just rename all *.jsx files to *.js.

pkg/shell/topnav.jsx
```diff
@@ -168,6 +171,8 @@ export class TopNav extends React.Component {
             </DropdownItem>,
         );
 
+        const D = this.context;
```
martinpitt, line 173: Can this be moved up in the file, so that a reverse search for the identifier finds the declaration? Also see above for giving this a better name, or perhaps we just use `this.context` as-is?

pkg/shell/active-pages-modal.jsx
```diff
@@ -57,14 +58,16 @@ export class ActivePagesDialog extends React.Component {
     }
 
     onRemove() {
+        const D = this.context;
         this.state.frames.forEach(element => {
             if (element.selected)
                 this.props.frames.remove(element.host, element.component);
         });
-        this.props.onClose();
+        D.close();
```
martinpitt, line 66: Especially for such single-use calls, just `this.context.close()` may be enough? And if we want to point out the "data type",  i.e. that we expect the context to be a DialogContext, it should have a descriptive name.

pkg/shell/active-pages-modal.jsx
```diff
@@ -110,3 +113,5 @@ export class ActivePagesDialog extends React.Component {
         );
     }
 }
+
+ActivePagesDialog.contextType = DialogsContext;
```
martinpitt, line 117: Oh, this is nice. Does ESLint or React warnings in our /devel scenario enforce this?

martinpitt (changes requested):
Thanks! I agree to the "PROPOSAL" commit -- if we go through the trouble of changing everything for this new "best practice", then let's do this more tightly.

pkg/lib/dialogs.jsx
mvollmer: Yes, I was thinking too that just `D` is too cute... :-)  It's just a pattern, every component can decide by itself what it wants to call this thing, but we should establish a proper pattern from the start.

What about `Dialogs`?  `const Dialogs = useDialogs();`

pkg/shell/topnav.jsx
mvollmer: Oh, I didn't realize it is declared after first use! Thanks for catching this!

pkg/shell/active-pages-modal.jsx
mvollmer: You mean `this.dialogs_context` instead of `this.context`?  I am afraid `this.context` is prescribed by React, or am I missing something?

pkg/shell/active-pages-modal.jsx
mvollmer: I don't think so. I missed it a couple of times, and then `this.context` remains `undefined`.

pkg/shell/active-pages-modal.jsx
martinpitt: I meant that we do this here:
```js
const D = this.context;
D.close();
```
and not use `D` anywhere else. So obviously it *is* `this.context`, or what am I missing? I don't see anything magic here?

martinpitt: Or do you *have* to assign `this.context` to a variable, and that's the magic that I'm missing?

mvollmer: There is no magic.  I did `const D = this.context` only to give it a name, so that we are reminded what the context actually is.  This is all gone now with the rewrite to function components.

pkg/shell/indexes.js
mvollmer: Done.

mvollmer: Done.

pkg/shell/topnav.jsx
mvollmer: Fixed by rewriting to function components.

pkg/lib/dialogs.jsx
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * const App = () =>
+ *   <WithDialogs>
+ *     <Page>
+ *       <ExampleButton />
+ *     </Page>
+ *   </WithDialogs>;
+ *
+ * const ExampleButton = () => {
+ *   const D = useDialogs();
```
martinpitt, line 34: IMHO the example/documentation should use the same "best practice" name as the code, i.e. `Dialogs`.

pkg/lib/hooks.js
```diff
@@ -299,3 +301,27 @@ export function useEvent(obj, event, handler) {
...
+ * "useInit(func, deps, comps)" is the same as "useObject(func, null,
+ * deps, comps)" but if you want to emphasize that you just want to
+ * run a function (instead of creating a object), it is clearer to use
+ * the "useInit" name for that.  Also, "deps" are optional for
+ * "useInit" and default to "[]".
+ */
+
+export function useInit(func, deps, comps) {
+    return useObject(func, null, deps || [], comps);
+}
```
martinpitt, line 327: As this is a library, it could make sense to preserve `this` by avoiding the `function`? I.e.

    export const useInit = (func, deps, comps) => useObject(...);

pkg/shell/credentials.jsx
```diff
@@ -42,12 +42,14 @@ import { ListingPanel } from 'cockpit-components-listing-panel.jsx';
...
 import { useEvent, useObject } from 'hooks';
+import { useDialogs } from "dialogs.jsx";
 
 import "./credentials.scss";
 
 const _ = cockpit.gettext;
 
-export const CredentialsModal = ({ onClose }) => {
+export const CredentialsModal = () => {
+    const D = useDialogs();
```
martinpitt, line 52: `D` renaming?

martinpitt (changes requested):
Cheers!

pkg/lib/dialogs.jsx
mvollmer: Of course, ouch.

pkg/lib/hooks.js
mvollmer: That wouldn't make any difference, no? The `this` of a function invocation is determined by how the function is defined (`function` vs `=>`) and by how it is called (`func()` vs `obj.func()`), but not by how the function is defined that contains the call.

mvollmer: (and in any case, the call is inside `useObject`, not `useInit`.)

pkg/lib/dialogs.jsx
martinpitt: Unresolving, this wasn't actually changed. Forgot `git add` or something?

pkg/shell/credentials.jsx
mvollmer: Ouch, yes.

pkg/lib/hooks.js
martinpitt: I just thought that in a call like

    const init_pages = useInit(() => { my init code }, [frames]);

`my init code` should be able to refer to the component as `this`. I understand that if `useInit()` is a `function`, it will change `this` to itself? But this has always confused me admittedly. As long as it doesn't get in the way, we can keep it as-is, we can change it later if you prefer.

mvollmer: Yes, let' leave it as it is. :-)

martinpitt (approved):
Cheers! Now let the conversion flood begin :grin:

PR #17299, opened 2022-05-02
shell: Rewrite superuser dialogs 

mvollmer:
Now they use function components and the UnlockDialog maintains its
own state.

Most importantly, the new SuperuserButton is no longer tied to the
life-cycle of the dialogs and the SuperuserAlert component and its
parents do not need to do anything special to keep it alive.

### DIFF
--- a/pkg/shell/superuser.jsx
+++ b/pkg/shell/superuser.jsx
@@ -18,7 +18,9 @@
  */
 
 import cockpit from "cockpit";
-import React from "react";
+import React, { useState } from "react";
+import { useObject, useEvent } from "hooks";
+import { useDialogs } from "dialogs.jsx";
 import { Alert, Button, Form, FormGroup, Modal, TextInput } from '@patternfly/react-core';
 import { ModalError } from 'cockpit-components-inline-notification.jsx';
 import { host_superuser_storage_key } from './machines/machines';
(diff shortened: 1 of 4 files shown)


pkg/shell/topnav.jsx
```diff
@@ -135,7 +132,8 @@ export class TopNav extends React.Component {
         });
 
         docItems.push(<DropdownSeparator key="separator1" />);
-        docItems.push(<DropdownItem key="about" component="button" onClick={() => this.setState({ aboutDialogOpened: true })}>
+        docItems.push(<DropdownItem key="about" component="button"
+                                    onClick={() => D.show(<AboutCockpitModal onClose={D.close} />)}>
```
martinpitt: This looks interesting, thanks! What I wonder here is what `D` is. It's declared nowhere in this file. It's obviously the `const D` from inside the `const WithDialogs`, I just wonder why that doesn't have to be qualified. Does merely importing `DialogsContext` magically declare the `D` global?

mvollmer: This commit is from #17298, let's discuss that first I'd say.

mvollmer: The D comes from line 173, `const D = this.context`, which is the value of the closests `DialogsContext.Provider` because of line 240, `TopNavInner.contextType = DialogsContext;`.  Using React contexts with class components is quite ugly, IMO. Maybe there is a better way that I didn't see.

With function components, one writes `const D = useDialogs();` at the top of the function.

mvollmer:
The "Authenticate" button got wider because it now has a consistent `isLoading={true|false}` attribute, as it should.  Previously we would have either `isLoading={true}` or no attribute at all. `isLoading={false}` makes the button slightly wider so that it doesn't need to grow when then spinner appears with `isLoading={true}`.  So this is a good change.

martinpitt (approved):
This looks great to me, thank you!

PR #17300, opened 2022-05-02
systemd: Use new protocol for dialogs

mvollmer:
- [x] #17298 
- [x] Dialogs should close when navigating away from a sub-page

### DIFF
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -27,12 +27,15 @@ import {
 
 import { ServerTime } from 'serverTime.js';
 import * as timeformat from "timeformat.js";
+import { DialogsContext } from "dialogs.jsx";
 
 import "cockpit-components-shutdown.scss";
 
 const _ = cockpit.gettext;
 
 export class ShutdownModal extends React.Component {
+    static contextType = DialogsContext;
+
     constructor(props) {
         super(props);
         this.date_spawn = null;
@@ -140,12 +143,13 @@ export class ShutdownModal extends React.Component {
     }
 
     onSubmit(event) {
+        const Dialogs = this.context;
         const arg = this.props.shutdown ? "--poweroff" : "--reboot";
         if (!this.props.shutdown)
             cockpit.hint("restart");
 
         cockpit.spawn(["shutdown", arg, this.state.when, this.state.message], { superuser: true, err: "message" })
-                .then(this.props.onClose)
+                .then(this.props.onClose || Dialogs.close)
                 .catch(e => this.setState({ error: e }));
 
         event.preventDefault();
@@ -160,6 +164,7 @@ export class ShutdownModal extends React.Component {
     }
 
     render() {
+        const Dialogs = this.context;
         const options = [
             <SelectOption value="0" key="0">{_("No delay")}</SelectOption>,
             <Divider key="divider" component="li" />,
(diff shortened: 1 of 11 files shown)


pkg/systemd/overview-cards/configurationCard.jsx
```diff
@@ -103,35 +101,14 @@ export const ConfigurationCard = ({ hostname }) => {
...
-    }
+const SystemInformationSshKeys = () => {
+    const Dialogs = useDialogs();
+    const [keys, setKeys] = useState([]);
+    const [error, setError] = useState("");
+    const [loading, setLoading] = useState(true);
 
-    keysUpdate() {
+    function keysUpdate() {
+        console.log("UPDATE");
```
mvollmer, line 111: Debug

mvollmer:
> I guess I broke services...

That was interesting. The test broke because a unexpected "Error" dialog was open. This dialog always happened, but it was closed when the test navigated away from the details sub-page. Now it stays open and breaks the test.

pkg/systemd/overview.jsx
```diff
@@ -39,10 +39,13 @@ import { UsageCard } from './overview-cards/usageCard.jsx';
 import { SuperuserAlert } from './superuser-alert.jsx';
 import { SuperuserIndicator } from "../shell/superuser.jsx";
 import { ShutdownModal } from 'cockpit-components-shutdown.jsx';
+import { WithDialogs, DialogsContext } from "dialogs.jsx";
 
 const _ = cockpit.gettext;
 
 class OverviewPage extends React.Component {
+    static contextType = DialogsContext;
```
martinpitt, line 47: Note to self: https://reactjs.org/docs/context.html#classcontexttype

This looks weird at first sight, as it's not referenced anywhere. This ought to go into the docs section of pkg/lib/dialogs.jsx.

pkg/systemd/services/service.jsx
```diff
@@ -72,26 +73,28 @@ export class Service extends React.Component {
...
-                        <GalleryItem id="service-details-unit">{serviceDetails}</GalleryItem>
-                        {((this.props.unit.LoadState === "loaded" || this.props.unit.LoadState === "masked") && this.props.owner == "system") &&
-                        <GalleryItem id="service-details-logs">
-                            <LogsPanel title={_("Service logs")} match={match} emptyMessage={_("No log entries")} max={10} goto_url={url} search_options={{ prio: "debug", service: cur_unit_id }} />
-                        </GalleryItem>}
-                    </Gallery>
-                </PageSection>
-            </Page>
+            <WithDialogs>
+                <Page groupProps={{ sticky: 'top' }}
```
martinpitt, line 77: This should be in the previous commit, not the "only reset units.." one.

martinpitt (changes requested):
Thanks!

pkg/systemd/services/service.jsx
mvollmer: Ouch, yes.

mvollmer: I have moved this into its own PR: #17327  It has nothing to do with this change here anymore.

pkg/lib/cockpit-components-shutdown.jsx
```diff
@@ -140,12 +143,13 @@ export class ShutdownModal extends React.Component {
...
 
     onSubmit(event) {
+        const Dialogs = this.context;
         const arg = this.props.shutdown ? "--poweroff" : "--reboot";
         if (!this.props.shutdown)
             cockpit.hint("restart");
 
         cockpit.spawn(["shutdown", arg, this.state.when, this.state.message], { superuser: true, err: "message" })
-                .then(this.props.onClose)
+                .then(this.props.onClose || Dialogs.close)
```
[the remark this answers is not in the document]
mvollmer: Yes, but personally I don't feel like being super strict about this right now. The advantages of not having to specify onClose are not that great, and tying the code to the dialog protocol has its own costs. But then again, everybody should be able to use WithDialogs, so yeah, ultimately we could get rid of all the onClose props.

PR #17302, opened 2022-05-03
packit: Enable tests on Fedora Rawhide, systemd: Cancel shutdown as root

martinpitt:
These should be treated as advisory, but they will help us to spot
issues like https://bugzilla.redhat.com/show_bug.cgi?id=2081039

----

systemd 251 started to guard this as a privileged operation with polkit [1].
So call this as root if possible.
    
This is an incomplete fix: We should not even show the cancel action if
the session is not root and not allowed via a polkit rule, but that is
more intrusive -- let's fix the exploding gating test first.
    
https://bugzilla.redhat.com/show_bug.cgi?id=2081039
    
[1] https://github.com/systemd/systemd/commit/ec14fba91c94f38f3

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -20,6 +20,7 @@ jobs:
       targets:
       - fedora-35
       - fedora-36
+      - fedora-development
       - centos-stream-8-x86_64
       - centos-stream-9-x86_64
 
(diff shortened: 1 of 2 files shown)


martinpitt:
I will add the fix next, but I first want to double-check that this spots the issue.

martinpitt:
Yup, that [exactly reproduces](https://artifacts.dev.testing-farm.io/80f605eb-1a20-4664-9d51-07e27f661c8d/work-allkjQZYk/log.txt) the bug.

martinpitt:
I filed issue #17325 for the refinement.

martinpitt:
@jelly : As this is a blocker for tomorrow's release, friendly review reminder? Thanks!

jelly (approved):

PR #17304, opened 2022-05-03
Take two at unit-tests container rework

allisonkarlitskaya:
Here's another attempt at the goals of #15210:
 - more flexibility in terms of how the unit-tests environment is built
 - capable of running with selinux enabled
 - capable of being used with git worktrees

This is definitely less of a hack, but the new `run-container` is potentially overkill, but I have a few other ideas for how I'd like to use that in the future. :)


tools/make-dist
```diff
@@ -1,18 +1,21 @@
 #!/bin/sh
 
 set -eu
-cd "$(realpath -m "$0"/../..)"
+cd "${0%/*}/.."
```
martinpitt, line 4: This should be in the first commit, not "stub out xgettext"

tools/make-dist
```diff
@@ -54,6 +57,7 @@ else
     ../../configure \
         CPPFLAGS=-I../../tools/mock-build-env \
         PKG_CONFIG_PATH=../../tools/mock-build-env \
+        XGETTEXT='/bin/true' \
```
martinpitt, line 60: Hmm.. I *think* we use xgettext and msg{cat,merge} etc. only for building/updating po/pot files (i.e. not relevant for the tarball), but this also controls `xsltproc` -- and that we *do* need to build the tarball. This has the potential to release tarballs with broken documentation. I assume the build wouldn't fail later on, as otherwise what's the point of stubbing this out.

tools/run-container
```diff
@@ -0,0 +1,199 @@
...
+        cmd = ['podman', *args]
+
+    return run(cmd, check=check)
+
+
+class PodmanTemporaryDirectory(tempfile.TemporaryDirectory):
+    """TemporaryDirectory subclass capable of removing files owned by subuids"""
+    @classmethod
+    def _rmtree(cls, name, ignore_errors=False):
+        podman('unshare', 'rm', '-r', name)
```
martinpitt, line 57: Unused -- supposedly this should be called on `__exit__()`?

containers/unit-tests/start
```diff
@@ -3,31 +3,17 @@ set -eu
 
 cd "${0%/*}/../.."
 image=ghcr.io/cockpit-project/unit-tests:latest
-flags=''
+cmd=''
 
 while test $# -gt 0; do
   case "$1" in
-    --docker) docker="docker";;
```
martinpitt, line 10: Feel free to eliminate this script entirely, as the commit message suggests. I doubt that anyone is "familiar" with that API, hardly noone except perhaps you actually runs this manually. There is already a gazillion scripts around this, and it's really hard to understand where the functionality actually lives -- so anything which can be removed is a little win.

martinpitt (changes requested):
This is still draft/red, but some initial remarks. In general I agree to the "overkill" -- honestly I think our current scripts are already too complicated, and it feels like over-generalization to me. This is a lot of extra/new code just to run `make check` in a container.. Honestly it rather feels to me that this whole "scenario" generalization etc. should be flattened back down to a single shell script..

tools/make-dist
allisonkarlitskaya: Thanks

tools/make-dist
allisonkarlitskaya: Now that I think about it I actually have no idea what the deal with this `xgettext`/`xsltproc` confusion is.  There two things have absolutely nothing to do with each other other than starting with `x`.  It smells like a copy/paste error.  I'll sort that out.

In any case:
 - xgettext is only used when building a `.pot` file, but since there is a rule for this in the `Makefile`, one might assume that this rule will function properly after `./configure` as run.  If we want to change that assumption, I'd add a `--enable-pot-rules` or so, but this seems like some serious overkill.  tl;dr: Status quo is OK here I think.
 - `xsltproc` is only used for building docs and our dependency on it ought to be disabled via `--disable-doc`.
 - `msgfmt` is how we get `.mo` files for `gettext()` in C, and this is part of the main build process.  Because we like pain, we also use `msggrep` as part of that process, in a very bizarre way that I'd like to burn down.  In short: this needs to be here, until we decide to kill it.

All in all, I think everything is right for this PR, but we certainly have a lot of work we could do in a separate PR.  Let's discuss that in the meeting we have in 6 minutes :)

tools/run-container
allisonkarlitskaya: This is not unused.  It's overriding a protected method declared in `TemporaryDirectory()` and is the entire point of this subclass: we need "more power" than the normal `rm -rf` approach to this, so we use `podman`.

containers/unit-tests/start
allisonkarlitskaya: Ya.  Sounds good.

Where would you like to see the new script live, btw?  `tools/run-container` seemed kinda ... I don't know.  Other ideas I had:
 - in `containers/unit-tests`
 - in `containers` itself
 - other ideas?

allisonkarlitskaya:
> This is still draft/red, but some initial remarks. In general I agree to the "overkill" -- honestly I think our current scripts are already too complicated, and it feels like over-generalization to me. This is a lot of extra/new code just to run `make check` in a container.. Honestly it rather feels to me that this whole "scenario" generalization etc. should be flattened back down to a single shell script..

It's not a bad idea.  This would be a good use of the `--make` option of the new script.  Then we could basically have like:

 - `--image=...:i386 --env=CC=clang --make check-memory`
 - `--image=...:latest --env=CC=gcc --make distcheck`

&c.

containers/unit-tests/start
martinpitt: containers/unit-tests/run or /start feels about right to me -- let's not generalize this until we actually need it (if at all), and we may be able to simplify it while it stays unit-tests specific?

.github/workflows/unit-tests.yml
```diff
@@ -24,15 +24,21 @@ jobs:
 
       - name: Build unit test container if it changed
         run: |
+          exit 0
```
martinpitt, line 27: Leftover

containers/unit-tests/start
```diff
@@ -1,33 +1,199 @@
...
+
+def logged(func):
+    def wrapper(args, **kwargs):
+        print('+', shlex.join(args))
+        return func(args, **kwargs)
+    return wrapper
+
+
+def git(*args):
+    run(['git', *args], check=True)
```
martinpitt, line 37: I was looking for a `def run` here for quite a while until I realized that this is `subprocess.run`. Do you mind if we just use this qualified? I feel it's a too generic name for the top-level namespace. (*Not* a strong opinion)

containers/unit-tests/start
```diff
@@ -1,33 +1,199 @@
...
+                       help='Set an environment variable in the container')
+    group.add_argument('--network', action="store_true",
+                       help="Enable network in the container (default: disabled)")
+    group.add_argument('--interactive', '-i', action="store_true",
+                       help="Interactive mode (implied by no command or script)")
+    group.add_argument('--tty', '-t', action="store_true",
+                       help="Allocate a pseudoterminal (implied by no command or script)")
+    group.add_argument('--user', help="Pass through the --user flag to podman")
+    group.add_argument('--entrypoint', metavar='CMD', help="Provide the --entrypoint flag to podman")
+    group.add_argument('--workdir', help="Provide the --workdir flag to podman")
```
martinpitt, line 119: This and the glue below is just duplicating `podman`s CLI and rather bulky. Having slight variations in there like `--network` also doesn't help. Couldn't this be moved wholesale to podman after `--`?

containers/unit-tests/start
```diff
@@ -1,33 +1,199 @@
...
+
+    group = parser.add_argument_group(title='What to build').add_mutually_exclusive_group()
+    group.add_argument('--head', action='store_true', help='Build the HEAD commit')
+    group.add_argument('-b', dest='branch', metavar='NAME', help='Build the named branch or tag')
+    group.add_argument('--work-tree', action='store_true',
+                       help='Build the HEAD commit, plus changes on the filesystem (default)')
```
martinpitt, line 126: This also feels like confusing duplication and overhead glue to me. There's a zillion way of testing a git tree, with local changes, local or remote branches or whatnot. Every other `make check` in the world just runs the current checkout.

martinpitt (changes requested):
Thanks! The bit I insist on is the `exit 0` (but I figure that's just an oversight). Other than that, I don't like replacing an already complicated structure with an even bigger script. This `start` script tries to do a million things, we use 4 scenarios (which are structurally pretty much the same), and it feels like this should all just be a rather simple `podman -v .:/source:ro,z .. make check` etc. invocation. But I know you want to get this off your desk, so I won't push back hard on this. I'm still interested in the "why oh why", see below :grin:

containers/unit-tests/start
allisonkarlitskaya: The issue is the clever hacking we do when `-v` is presented:

```python
def logged(func):
    def wrapper(args, **kwargs):
        print('+', shlex.join(args))
        return func(args, **kwargs)
    return wrapper

...later...

    if args.verbose:
        global run
        run = logged(run)
```

containers/unit-tests/start
allisonkarlitskaya: The problem is that we can't just use the current checkout because we can't get it inside of the container without relabelling it.  We could definitely trim down some of the options here, though.

containers/unit-tests/start
allisonkarlitskaya: That's an interesting suggestion.

martinpitt (approved):
Ack. As discussed, this goes into the wrong direction IMHO, but I don't want to block you on this. Next time I care, I'd like to drop that start script in lieu of shell single-liners that can run on gh or locally (and yes, it will break with worktrees or change your SELinux contexts, but this overhead is just too much a price for that).

allisonkarlitskaya:
From my side: if a script exists, I expect to be able to run it, from a worktree, with selinux enabled.

But I'd happily get rid of the script and replace it with a `container:` stanza in our workflow.

PR #17305, opened 2022-05-03
workflows: Prefix node-cache tags with project name

martinpitt:
This is a shared repository, and e.g. cockpit-machines is using that as
well now. We want to tag c-machines releases in this as well. To avoid
version number collisions, prefix the tag with the project name.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,8 @@ jobs:
       - name: Tag node-cache
         run: |
           set -eux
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Thanks.

Shall we rename the old ones?

martinpitt:
@allisonkarlitskaya : We don't have to and I wasn't planning to, as this is more of a lifeline than something we use regularly. (But of course no veto)

PR #17306, opened 2022-05-03
overview: Right alight reboot dropdown

marusak:
If `Shutdown` is much longer than `Reboot` (e.g. in Japanese) then the
`Shutdown` string overflows out of the page.

@garrett this is just very simple approach. Do you think we should do something different here, or is this good enough?

Before:
Screenshot from 2022-05-03 09-52-04

Now:
Screenshot from 2022-05-03 10-13-13

### DIFF
--- a/pkg/systemd/overview.jsx
+++ b/pkg/systemd/overview.jsx
@@ -26,7 +26,7 @@ import ReactDOM from 'react-dom';
 import {
     Page, PageSection, PageSectionVariants,
     Gallery,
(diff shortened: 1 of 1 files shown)


martinpitt (commented):
IMHO conceptually this makes sense anyway -- we know that this dropdown is on the right page border, so it should grow to the left. Still, leaving final yay/nay to Garrett. Thanks!

garrett (approved):
This makes sense, but may cause problems in RTL languages (which we currently do not support and probably has a ton of these kind of issues).

PR #17310, opened 2022-05-04
test: Ensure tests don't trigger Oops

martinpitt:
At the end of each test, ensure that the Oops indicator is not visible.
This validates commit 7308d59bbb7716 and similar situations.

This will not catch every situation, but most tests don't bother with
explicitly logging out, so this will still catch most potential
instances.

There is only one legitimate Oops in our entire test suite
(`TestMenu.testBasic`) which tests the Oops indicator. Log out to clear
it.


martinpitt:
Without the check-menu fix, it fails like this:
```
  File "/var/home/martin/upstream/cockpit/test/common/testlib.py", line 1390, in tearDown
    self.check_browser_errors()
  File "/var/home/martin/upstream/cockpit/test/common/testlib.py", line 1626, in check_browser_errors
    self.assertFalse(b.is_visible("#navbar-oops"))
AssertionError: True is not false

# Result testBasic (__main__.TestMenu) failed
```

martinpitt:
Cool, works! So, full round of tests.

test/common/testlib.py
```diff
@@ -1604,9 +1604,12 @@ def allow_browser_errors(self, *patterns):
             self.allowed_console_errors.append(p)
 
     def check_browser_errors(self):
-        if not self.browser:
+        b = self.browser
+
+        if not b:
```
allisonkarlitskaya, line 1609: I start to prefer assignment expressions for cases like these, but that's a matter of taste.

test/verify/check-shell-menu
```diff
@@ -75,6 +75,9 @@ class TestMenu(MachineCase):
         b.switch_to_top()
         b.wait_visible("#navbar-oops")
 
+        # clear Oops
```
allisonkarlitskaya, line 78: So that makes sense, but it's also kind of a lost opportunity.  Specifically: how do we know for sure that the check for oops is effective?  This seems very much like the sort of thing that might decay at some point in the future if we change the ID of the banner, for example.

It would be cool to have a (separate) test which we expect to fail which shows an oops banner and then exits directly, just to make sure the test works.

The fact that we can clear an oops banner merely by logging out is also somehow suboptimal.  It seems like a lot of tests would logout incidentally at the end, rendering the check ineffective.  It would be cool to be able to detect if there was an oops shown at any point during testing...

allisonkarlitskaya (changes requested):

test/common/testlib.py
martinpitt: These are fine if the scope of `b` is limited to the "then" part. But here it isn't -- it's our standard "local short alias for the whole function" pattern.

test/verify/check-shell-menu
martinpitt: I'll think about that -- I suppose `@unittest.expectedFailure` should do here, as there is no other way of intercepting a failure in tearDown().

> It would be cool to be able to detect if there was an oops shown at any point during testing...

It would, but it would also add an extra visibility check to *every* individual test step. That seems too expensive to me for such a coarse-grained opportunistic test. Sure, quite a lot of tests use `.logout()` in the middle, but they usually don't at the end.

So what I can do is to add the "no oops" check to `def logout()` as well, and then provide some `self.allow_oops()` helper which sets a flag that tearDown() looks at. WDYT?

test/common/testlib.py
martinpitt: This is gone now.

test/verify/check-testlib
```diff
@@ -213,8 +213,8 @@ class TestSystemd(unittest.TestCase):
 
 
 @skipDistroPackage()
+@nondestructive
```
martinpitt, line 216: Context: I originally started to write a new `TestTestlib.testOopsDetection`, but then discarded that -- it would do almost the same as `TestMenu.testBasic`, so I refined that instead. But this little detail is still legit IMHO, but doesn't warrant its own PR.

allisonkarlitskaya (approved):
I'd still like to see a simple test which throws an oops and is marked `@expectfail` or whatever...

PR #17311, opened 2022-05-04
shell: Translate documentation links

marusak:
We already pick these items from manifest and translate them. We just
never told Cockpit to show them in the target language.

### DIFF
--- a/pkg/shell/topnav.jsx
+++ b/pkg/shell/topnav.jsx
@@ -130,7 +130,7 @@ export class TopNav extends React.Component {
 
         docs.forEach(e => {
             docItems.push(<DropdownItem key={e.label} href={e.url} target="blank" rel="noopener noreferrer" icon={<ExternalLinkAltIcon />}>
-                {e.label}
+                {_(e.label)}
             </DropdownItem>);
         });
 


pkg/shell/topnav.jsx
```diff
@@ -130,7 +130,7 @@ export class TopNav extends React.Component {
 
         docs.forEach(e => {
             docItems.push(<DropdownItem key={e.label} href={e.url} target="blank" rel="noopener noreferrer" icon={<ExternalLinkAltIcon />}>
-                {e.label}
+                {_(e.label)}
```
martinpitt, line 133: This smells like it should trigger our static code checks, as it's not a valid xgettext-parseable string. I.e. it would be better to write this as `cockpit.gettext(e.label)`. But as there are no complaints, fine.

martinpitt (approved):
Dakujem!

martinpitt:
[testSshKeySetup](https://logs.cockpit-project.org/logs/pull-17311-20220509-065918-60111fbe-ubuntu-stable/log.html#144-1) is really hard to please these days, and unrelated. (3x affected retry)

PR #17312, opened 2022-05-04
shell: Only remember admin access when the prompts are right

mvollmer:
Sudo-on-login will only work when there is a single prompt and it is
for the account password.  If we see anything else, let's not remember
the access level since sudo will very likey fail on next login.

### DIFF
--- a/pkg/shell/superuser.jsx
+++ b/pkg/shell/superuser.jsx
@@ -263,7 +263,9 @@ export class SuperuserDialogs extends React.Component {
             cancel: cancel
         });
 
+        let user;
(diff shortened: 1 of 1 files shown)


mvollmer:
This has problems with l10n, obviously, but the sudo prompts don't seem to be localized in Cockpit anyway...

mvollmer:
This grew out of #17177.  Less maghic is better, so I think if we can agree on a good description for the checkbox in #17177, we should kill this PR here.

mvollmer:
Ach, let's close this. I have enough PR open... :-)

PR #17315, opened 2022-05-05
tls: avoid static port in test-tls-server

allisonkarlitskaya:
We've run into issues where port 9123 isn't available on the builder
we're using, so let's use a dynamically allocated port instead.

We accomplish this by passing 0 as the port number (which already
worked) but we need to have a way to figure out what port we actually
used.  We add a minimal accessor to the server code to let us get the
listener socket fd, which we can call getsockname() on to determine to
where we ought to connect.

### DIFF
--- a/src/tls/server.c
+++ b/src/tls/server.c
@@ -272,6 +272,13 @@ server_init (const char *wsinstance_sockdir,
     }
 }
 
+int
+server_get_listener (void)
+{
+  assert (server.first_listener == server.last_listener);
+  return server.first_listener;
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
Lol @`error: Unable to load summary from remote flathub: URI https://dl.flathub.org/repo/summary exceeded maximum size of 10485760 bytes`

Let's move that one out of the "required" set until they remove a few packages :rofl:

jelly (approved):

PR #17317, opened 2022-05-05
po: stop translating .c code

allisonkarlitskaya:
Our attempts to translate strings generated by C code in cockpit are
hopelessly complicated and currently ineffectual.

There are only three strings that have been marked for translation, and
they're all in the bridge.  Meanwhile, the code to compile the
translations only considers files found in `src/ws`, which skips the
files in the bridge, but includes the translations found in the
`cockpit-ws` appdata.  That might be helpful, except that we already use
`msgfmt` to merge these translations at build time.

Drop the rules for creating and installing the filtered .mo files.

I created `cockpit.pot` before and after this change and confirmed that
exactly the three expected strings disappeared, and that (except for
timestamps) nothing else changed.


martinpitt (approved):
Thanks! The most important string here is the "package cockpit-system is not installed", but as that's already not translated right now, this isn't a regression. We can translate that on the JS side without much effort, too.

PR #17320, opened 2022-05-05
test: document testlib functions (part one of X)

jelly:
@allisonkarlitskaya  @martinpitt @marusak what do you think of using sphinx for generating documentation of our testlib combined with some documentation on how to set up the tests. The documentation should be aimed for us but mostly for people new to writing tests.

Secondly which docstring format do we go for, the Rst style looks the best in sphinx.

https://realpython.com/documenting-python-code/#docstring-formats

Oh and we might want to make a cockpit theme for sphinx?

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -389,11 +389,11 @@ class Browser:
     # TODO: Unify them so we can have only one
     def key_press(self, keys, modifiers=0, use_ord=False):
         if self.cdp.browser.name == "chromium":
-            self.key_press_chromium(keys, modifiers, use_ord)
+            self._key_press_chromium(keys, modifiers, use_ord)
         else:
-            self.key_press_firefox(keys, modifiers, use_ord)
+            self._key_press_firefox(keys, modifiers, use_ord)
 
-    def key_press_chromium(self, keys, modifiers=0, use_ord=False):
+    def _key_press_chromium(self, keys, modifiers=0, use_ord=False):
         for key in keys:
             args = {"type": "keyDown", "modifiers": modifiers}
 
@@ -411,7 +411,7 @@ class Browser:
             args["type"] = "keyUp"
             self.cdp.invoke("Input.dispatchKeyEvent", **args)
 
-    def key_press_firefox(self, keys, modifiers=0, use_ord=False):
+    def _key_press_firefox(self, keys, modifiers=0, use_ord=False):
         # https://github.com/GoogleChrome/puppeteer/blob/master/lib/USKeyboardLayout.js
         keyMap = {
             8: "Backspace",   # Backspace key
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -141,11 +141,11 @@ default_layouts = [
 
 
 def attach(filename, move=False):
-    '''Put a file into the attachments directory
+    """Put a file into the attachments directory
 
     By default the file gets copied. You can set move=True for dynamically generated
     files which are not touched by parallel tests.
-    '''
+    """
     if not opts.attachments:
         return
     dest = os.path.join(opts.attachments, os.path.basename(filename))
@@ -770,11 +770,11 @@ class Browser:
         self.switch_to_frame(cur_frame)
 
     def click_system_menu(self, path, enter=True):
-        '''Click on a "System" menu entry with given URL path
+        """Click on a "System" menu entry with given URL path
 
         Enters the given target frame afterwards, unless enter=False is given
         (useful for remote hosts).
-        '''
+        """
         self.switch_to_top()
         self.click(f"#host-apps a[href='{path}']")
         if enter:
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
My first take is that I'm not a huge fan.  I'm all for writing better docstrings, but I'm not sure we should be aiming to generate HTML for this stuff...

marusak:
Can you generate the html and show how it would look like?

I am very much in favor of documenting our testlib - we expect everyone to use this to write tests yet we are fully aware it is very difficult for anyone new to understand how it works and what function they can use (I was asked so many times questions like "How do I check if this element is on the page", "How do I enter some text into input"...) so having some page (ideally in https://cockpit-project.org/documentation.html) would be awesome.

jelly:
> Can you generate the html and show how it would look like?

image


> I am very much in favor of documenting our testlib - we expect everyone to use this to write tests yet we are fully aware it is very difficult for anyone new to understand how it works and what function they can use (I was asked so many times questions like "How do I check if this element is on the page", "How do I enter some text into input"...) so having some page (ideally in https://cockpit-project.org/documentation.html) would be awesome.

Agree!

jelly:
> My first take is that I'm not a huge fan. I'm all for writing better docstrings, but I'm not sure we should be aiming to generate HTML for this stuff...

My plan is to first focus on adding the docstrings to everywhere it's missing, for that we need to agree on the docstrings format.

I still think having generated HTML is an added benefit, we document the [javascript API](https://cockpit-project.org/guide/latest/api-base1.html) so why not our browser API as well. The only downside is that it changes more and it's not a really strict API.

martinpitt:
I fully agree to adding docstrings to all of testlib's public API. That can then easily be viewed with `pydoc` or the usual `help()`, and it's even fine IMHO to just simply look at the file.

I'm less excited about doing the sphinx build -- we shouldn't publish that on our website. Every project uses its own tag/version of testlib, so this is prone to version confusion. It also adds complexity to our build system, and a new build dependency. I.e. this PR doesn't test the doc build *at all*.

Also, HTML is a bit "meh", Markdown would be nicer -- but that's just a little whim, as long as it's automated I don't mind much.

Thank you!

jelly:
- [ ] figure out an alternative easier html generation tool (with less makefile's n stuff) https://pypi.org/project/pdoc3/ maybe
- [ ] document
- [ ] is there a linter for pydoc strings? https://pypi.org/project/flake8-rst-docstrings/ ?

jelly:
> * [ ]  figure out an alternative easier html generation tool (with less makefile's n stuff) https://pypi.org/project/pdoc3/ maybe
> * [ ]  document
> * [ ]  is there a linter for pydoc strings? https://pypi.org/project/flake8-rst-docstrings/ ?

Good candidate seems to be https://github.com/PyCQA/pydocstyle

test/common/testlib.py
```diff
@@ -215,19 +216,39 @@ def open(self, href, cookie=None, tls=False):
...
         self.wait_js_cond("ph_select('iframe.container-frame').every(function (e) { return e.getAttribute('data-loaded'); })")
         self.cdp.invoke("reloadPageAndWait", ignoreCache=ignore_cache)
 
         self.machine.allow_restart_journal_messages()
 
     def switch_to_frame(self, name):
+        """Switch to frame in browser tab
+
+        Each page has a main frame and can have multiple subframes for example
+        iframes.
```
martinpitt, line 241: "subframes, usually iframes.

test/common/testlib.py
```diff
@@ -271,9 +299,18 @@ def eval_js(self, code, no_trace=False):
         return None
 
     def call_js_func(self, func, *args):
+        """Call a JavaScript function
+
+        :param func: JavaScript function to call
+        :param args: arguments for the JavaScript function
+        """
```
martinpitt, line 306: This is rather stub-y -- this does not really explain much beyond the signature. But we can build on this :-)

test/common/testlib.py
```diff
@@ -271,9 +299,18 @@ def eval_js(self, code, no_trace=False):
...
         return self.eval_js("%s(%s)" % (func, ','.join(map(jsquote, args))))
 
     def cookie(self, name):
+        """Retrieve the browsers cookie by name
```
martinpitt, line 310: "a browser cookie

test/common/testlib.py
```diff
@@ -143,6 +143,9 @@ def attach(filename, move=False):
 
     By default the file gets copied. You can set move=True for dynamically generated
     files which are not touched by parallel tests.
+
+    :param filename: file to put in attachments directory
+    :param move: move file to attachments directory (default False)
```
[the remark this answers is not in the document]
jelly: Okido. done

test/common/testlib.py
```diff
@@ -141,11 +139,12 @@
 
 
 def attach(filename, move=False):
-    '''Put a file into the attachments directory
+    """Put a file into the attachments directory.
```
martinpitt, line 142: For next time: No periods in short description.

martinpitt (approved):
Thanks!

PR #17322, opened 2022-05-06
bridge: Don't 404 on "po.js"

allisonkarlitskaya:
If we can't find a file called "po.js", return an empty document rather
than an error.  This matches what happens in the globbing case, and
avoids filling the debugging console with messages about incorrect
content type when requesting po.js in English (due to our HTML 404
page).


martinpitt (approved):
Cheers!

PR #17327, opened 2022-05-10
systemd: Only reset units that have actually failed

mvollmer:
Units that that have not failed might not be loaded and calling
ResetFailedUnit on them might return a "Unit is not loaded" error.


martinpitt:
```
/tmp/source/pkg/systemd/services/service-details.jsx
  171:51  error  'failed' is not defined  no-undef
```

martinpitt (approved):
Cheers!

PR #17328, opened 2022-05-10
sass: Reduce duplication in variables import

garrett:
Contents of `@patternfly/patternfly/base/patternfly-variables.scss` currently includes two imports:

```scss
@import "../sass-utilities/all";
@import "variables";
```

The `sass-utilties/all` import does not render into CSS; it only includes SASS-specific contents (that is: variables and mixins). However, the `variables` import _does_ render to CSS, which means that we wind up with CSS-rendered outputs of PatternFly variables every time our `_global-variables.scss` file is included anywhere... and this file of ours is included in `page.scss`, so this is a problem with every single page.

Changing the import in our `_global-variables.scss` to only use the sass-utilities file means that we're not accidentally redefining the CSS variables on the page. (The `@import "variables";` is included elsewhere already as a base part of PatternFly.)

### DIFF
--- a/pkg/lib/_global-variables.scss
+++ b/pkg/lib/_global-variables.scss
@@ -1,4 +1,4 @@
-@import "@patternfly/patternfly/base/patternfly-variables.scss";
+@import "@patternfly/patternfly/sass-utilities/all";
 
 /*
  * PatternFly 4 adapting the lists too early.


garrett:
I'm not sure if this was a change in PatternFly at some point or if it was an oversight on our side. (I would guess it was probably the latter, and I might even be at fault here.)

It's important to note that while this is just a 1 line change, it does affect every single page in Cockpit. It's a good thing we compare screenshots in our tests. :wink: 

If anything is somehow problematic after this change, we would want to change the imports locally to any affected page. We should keep this change, in other words.

As this is important for things like dark mode (so we're not accidentally rewriting the CSS after enabling a dark mode), but this change also shouldn't affect the current visuals of Cockpit (unless incorrect assumptions were made somewhere), I suggest that we do not merge this until after our upcoming release, so we're not accidentally breaking something without having some time to fix any possible issue.

martinpitt (approved):
Thanks for the detective work! This makes a lot of sense, we just need to watch out for strange effects as you said. I have quite some faith in our pixel tests these days, so if/once this goes green, we can land this.

garrett:
We might want to merge this on GitHub in a way so it pulls in the description I wrote instead of the commit (which doesn't have the description). I think the squash option does that?

martinpitt:
@garrett : Not automatically, but we can copy&paste it when doing squash&merge, yes.

garrett:
I copy/pasted and line-wrapped the text from the top comment into the commit and force-pushed. The contents of the commit are the same.

garrett:
Apologies for needing a review OK again and re-triggering the tests. Ah well; future us might thank current us. (Or not have to worry about it, hopefully.)

martinpitt (approved):
Thanks, tests look good so far! (crucially, pixel tests passed)

jelly (approved):

PR #17335, opened 2022-05-11
networking: Rewire dialogs to use the WithDialogs protocol

mvollmer:


pkg/networkmanager/dialogs-common.jsx
```diff
@@ -212,26 +215,43 @@ export const NetworkAction = ({ buttonText, iface, connectionSettings, type }) =
         if (type == 'bridge') settings = getBridgeGhostSettings({ newIfaceName });
     }
 
-    const properties = { connection: con, dev, setIsOpen, settings };
+    const properties = { connection: con, dev, settings };
+
+    function show() {
+        console.log("SHOW", type);
```
mvollmer, line 221: Debug

pkg/networkmanager/firewall.jsx
```diff
@@ -669,9 +676,24 @@ class ActivateZoneModal extends React.Component {
...
+    componentDidMount() {
+        firewall.addEventListener("changed", this.onFirewallChanged);
+    }
+
+    componentWillUnmount() {
+        firewall.removeEventListener("changed", this.onFirewallChanged);
+    }
+
+    onFirewallChanged() {
+        console.log("FIRE", JSON.stringify(Object.keys(firewall.zones)));
```
martinpitt, line 695: debugging leftover

pkg/networkmanager/firewall.jsx
```diff
@@ -669,9 +676,24 @@ class ActivateZoneModal extends React.Component {
             dialogError: null,
             dialogErrorDetail: null,
         };
+        this.onFirewallChanged = this.onFirewallChanged.bind(this);
         this.onInterfaceChange = this.onInterfaceChange.bind(this);
         this.onChange = this.onChange.bind(this);
         this.save = this.save.bind(this);
+        console.log("ZONES", JSON.stringify(Object.keys(firewall.zones)));
```
martinpitt, line 683: debugging leftover

martinpitt (changes requested):

pkg/networkmanager/firewall.jsx
```diff
@@ -669,9 +676,24 @@ class ActivateZoneModal extends React.Component {
...
+    }
+
+    componentDidMount() {
+        firewall.addEventListener("changed", this.onFirewallChanged);
+    }
+
+    componentWillUnmount() {
+        firewall.removeEventListener("changed", this.onFirewallChanged);
+    }
```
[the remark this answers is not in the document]
mvollmer: It avoids a race in the test.  The test opens the dialog before the firewall state has been fully loaded. I was debating with myself how to fix this:

1) Remove the race from the test by having it wait (or retry opening the dialog) until the expected zone is listed.
2) Fix the page to not allow opening the dialog before the state is fully loaded.
3) Keep the current behaviour of the page.

3) seemed to be the easiest. :-) With 1) and 2) we would also have to change the dialog to not access `firewall.zones` at all but a stable copy of it, so that a normal re-render doesn't change the list.

pkg/networkmanager/firewall.jsx
mvollmer: Ouch.

pkg/networkmanager/firewall.jsx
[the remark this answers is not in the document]
mvollmer: Also option 3 is not wrong, imo. :-)

pkg/networkmanager/firewall.jsx
mvollmer: Removed.

pkg/networkmanager/firewall.jsx
mvollmer: Removed.

pkg/networkmanager/firewall.jsx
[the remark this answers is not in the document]
martinpitt: Option 2 is our current approach on most pages, and IMHO the correct one. It's dangerous and  wrong to do anything on a page until it is fully loaded, and it's usually easy enough to tell -- there should be some unresolved promises. So it would be best to disable the button until some page component property becomes initialized; similar to `isDisabled={this.state.avail_services === null}`.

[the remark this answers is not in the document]
martinpitt: I don't see how (2) requires passing a copy of `firewall.zones`? It should just not be possible to open dialogs until firewall.zones is initialized in the first place.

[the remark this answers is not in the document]
mvollmer: If we want the dialog to not change when the firewall state changes, then the render code must not read `firewall.zones`.  So either we listen to changes on the firewall, or we don't access its state during render.

The old code did listen to changes (indirectly) and did access `firewall` state in its render method. So I kept that. (With the test race, I doubt I would have noticed...)

I'll push a FIXUP with option 2.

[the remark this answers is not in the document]
mvollmer: > I'll push a FIXUP with option 2.

Well, making a copy of the firewall state is more hair than I want to take on right now. :-) But showing a spinner until the state is loaded seems easy enough.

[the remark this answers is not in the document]
martinpitt: Yes, I understand -- I'm on the fence of whether the dialog should update while it's open (it feels bad, but then again not showing current state is bad as well). I was just saying that this question is independent from disabling action buttons during the initial page load -- the latter will not fix *all* race conditions, but the vast majority of practical ones.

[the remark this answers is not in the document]
mvollmer: > I was just saying that this question is independent from disabling action buttons during the initial page load

Right. I put a spinner there until the first getZones call has finished.

test/verify/check-networkmanager-firewall
```diff
@@ -138,12 +138,12 @@ class TestFirewall(NetworkCase):
 
         # Zones should be visible as unprivileged user
         self.login_and_go("/network/firewall", superuser=False)
-        self.wait_onoff("#firewall-heading-title-group", False)
+        self.wait_onoff("#firewall-heading-title-group", True)
```
martinpitt, line 141: ouch! this was a race condition all the time, wasn't it? Thanks for spotting!

pkg/networkmanager/app.jsx
```diff
@@ -114,11 +115,13 @@ const App = () => {
     if (path.length == 0) {
         return (
             <ModelContext.Provider value={model}>
-                <NetworkPage privileged={superuser.allowed}
-                             operationInProgress={model.operationInProgress}
-                             usage_monitor={usage_monitor}
-                             plot_state={plot_state_main}
-                             interfaces={interfaces} />
+                <WithDialogs key="1">
```
martinpitt, line 118: A more descriptive key would be nice, like `page` and `interface` below.

martinpitt (approved):
Danke!

test/verify/check-networkmanager-firewall
mvollmer: Yes.  Putting the spinner in there exposed it.

PR #17336, opened 2022-05-11
containers: Simplify package installation for bastion container

martinpitt:
The install-rpms.sh script is way more generic than it needs to be,
still from a time when we had a kubernetes container. It also does not
work on a standard fedora:36 container any more due to the `util-linux`
dependency.

Replace this with a direct rpm/dnf invocation in install.sh. Drop the
`--nodeps` mode, we don't need this and it's conceptually very
questionable.

---

The containers/bastion test will cover the "local rpms/offline install" case, and I did a `podman build` run in containers/bastion to ensure that the "online/dnf from repo" case works as well.


allisonkarlitskaya (approved):
Let's drop cockpit-project/bots#3368 then?

PR #17337, opened 2022-05-12
packit: Take over Fedora koji/bodhi releases

martinpitt:
This changes the release workflow quite a bit: Until now, releases were
fully automatic, but take a long time and often several retries mostly
due to the serialized and brittle koji builds.

With the packit workflow, this gets replaced with not directly pushing
to dist-git, but proposing dist-git PRs which run scratch builds and the
integration tests. These all run in parallel, and asynchronous to our
upstream release workflow. At least initially we will need to wait for
the PRs to go green and land them manually. This is automatable though
[1], and hopefully Fedora will also at some point support pipelines on
their dist-gits.

packit automatically creates koji builds whenever something gets
committed to dist-git, this part has worked reliably for several months
now.

It also can create automatic bodhi updates whenever a koji build
succeeds; however, this is currently often broken [2].  But this is
actually an improvement for us: for the last 5 releases, the automatic
bodhi updates have mysteriously and silently failed for cockpit, and we
have created them manually anyway. This works fine for c-machines and
our other projects, so let's start with moving to packit for cockpit and
collect experience with it.

[1] https://github.com/osbuild/fedora-bot
[2] https://github.com/packit/packit-service/issues/1413

---

I enabled this a while ago for [umockdev](https://github.com/martinpitt/umockdev/blob/master/packit.yaml) and [python-dbusmock](https://github.com/martinpitt/python-dbusmock/blob/master/packit.yaml), and I find the experience rather nice. It creates [PRs like this](https://src.fedoraproject.org/rpms/umockdev/pull-request/39#request_diff), also taking the `%changelog` bits from the upstream release tag just like our machinery. [koji builds](https://koji.fedoraproject.org/koji/packageinfo?packageID=20322) also are reliable.


martinpitt:
This is a bit hard to validate beforehand. I tried to run
```
❱❱❱ packit propose-downstream
Proposing update of the following branches: f35, f36, rawhide
2022-05-12 07:51:24.867 upstream.py       INFO   Version in upstream registries is '268.1'.
2022-05-12 07:51:24.917 local_project.py  INFO   Checking out upstream version 268.1.
2022-05-12 07:51:25.054 base_git.py       INFO   Using user-defined script for ActionName.post_upstream_clone: [['tools/create-spec', '--version', '0', '--build-all', '-o', 'cockpit.spec', 'tools/cockpit.spec.in']]
2022-05-12 07:51:25.145 upstream.py       INFO   Version in spec file is '0'.
2022-05-12 07:51:25.145 api.py            WARNING Version '0' in spec file is outdated.
2022-05-12 07:51:25.148 api.py            INFO   Using 'f35' dist-git branch.
100%[=============================>]    11.75M  eta 00:00:00 
2022-05-12 05:51:31.076 utils.py          ERROR  Failed to download file from URL https://github.com/cockpit-project/cockpit/releases/download/0/cockpit-0.tar.xz. Reason: 'Not Found'. 
```

This at least confirms that this does run the `post-upstream-clone` command to generate ./cockpit.spec, so that the `files_to_sync:` has a chance to work. However, this stumbles over the hardcoded "0" version. Commit b34acdfcb740fc introduced that refering to a "fix spec" stage, but I don't think that the local packit CLI does that?

So this is certainly risky. Whenever we try this, it's likely that we'd need to iterate a few rounds of "unpush/repush tag" (as now the remaining steps in tools/cockpitous-release don't produce anything any more which can't be updated), or just reverting this change, or manually running release-koji. So that's the contingency plan, but I'll leave it to the currently acting pilot (@allisonkarlitskaya at the moment) to decide if/when they want to do that. In the worst case, I'll wait until it's my turn again :grin:

packit.yaml
```diff
@@ -13,6 +13,9 @@ srpm_build_deps:
   - glib2-devel
   - make
   - systemd-devel
+files_to_sync:
+  - packit.yaml
+  - cockpit.spec
```
martinpitt, line 18: Note: This will sync everything *but* the `%changelog`, see [docs](https://packit.dev/docs/configuration/#sync_changelog). This is what we want, and what our own release machinery does as well. The changelog just gets a new entry based on the upstream release tag.

martinpitt:
@allisonkarlitskaya just brought up a good point about the release tarballs, and I confirmed that packit does *not* take the tarball from what the .spec's `Source:` URL does, but it rebuilds it on its own:

```
$ curl https://github.com/martinpitt/umockdev/releases/download/0.17.10/umockdev-0.17.10.tar.xz | sha512sum
cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e
```

but [dist-git sources](https://src.fedoraproject.org/rpms/umockdev/blob/288350b36237ea39e892dcabbfd11b33373b76b3/f/sources) has a rebuilt tarball with 9baad9a6b4ee010ad7..

@TomasTomecek , @LachmanFrantisek is there a trick to let it download the official tarball? Distros really should use the official upstream tarballs.

allisonkarlitskaya:
> @allisonkarlitskaya just brought up a good point about the release tarballs, and I confirmed that packit does _not_ take the tarball from what the .spec's `Source:` URL does, but it rebuilds it on its own:
> 
> ```
> $ curl https://github.com/martinpitt/umockdev/releases/download/0.17.10/umockdev-0.17.10.tar.xz | sha512sum
> cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e
> ```
> 
> but [dist-git sources](https://src.fedoraproject.org/rpms/umockdev/blob/288350b36237ea39e892dcabbfd11b33373b76b3/f/sources) has a rebuilt tarball with 9baad9a6b4ee010ad7..

I've been planning for a while to specify a container image (with checksum) somewhere in the git repository for cockpit which is the official container that we build our tarball against.  Given a particular version of `automake`, `tar` and `xz` our tarball build is then fully deterministic.

That would allow the builder to get a tarball with exactly the same checksum as the one we release upstream.

But: can we run a container as part of the build process in packit?

martinpitt:
The srpm build, which AFAIK includes building the tarball, now runs in COPR (i.e. we  have very little structural influence there, and we surely can't run a container there). With some big effort we may be able to make it bitwise reproducible, but conceptually I wouldn't want to rebuild the official upstream tarball, ever.

martinpitt:
@TomasTomecek , @lachmanfrantisek: in particular, is there some trick how to use `create-archive:` action for upstream PR builds (where this has to happen), but ignore it for releases? Then the `Source0:` URL of the spec should DTRT and download the tarball from the github release?

TomasTomecek:
> The srpm build, which AFAIK includes building the tarball, now runs in COPR (i.e. we have very little structural influence there, and we surely can't run a container there). With some big effort we may be able to make it bitwise reproducible, but conceptually I wouldn't want to rebuild the official upstream tarball, ever.

I haven't tried but the Copr environment is a full VM so it should be technically possible to run containers there.

martinpitt:
@TomasTomecek : I'd need to delete the key for `create-archive:`. But I figure for the `propose_downstream` job I could just replace it with a `curl -O` which downloads it from upstream? I'll try that (on umockdev first), thanks!

martinpitt:
With @TomasTomecek I sorted out the "use upstream official tarball" approach, and validated it. See https://github.com/cockpit-project/cockpit-podman/pull/978 , once that lands, I'll apply it here as well.

.github/workflows/release.yml
```diff
@@ -7,29 +7,14 @@ on:
 jobs:
   cockpituous:
     runs-on: ubuntu-latest
-    environment: release
     container:
```
allisonkarlitskaya, line 10: Explicit `permissions:` line for the `GITHUB_TOKEN` please (uploading the release I guess?)

packit.yaml
allisonkarlitskaya: I actually don't understand this part at all.  Let's talk.

allisonkarlitskaya (changes requested):

packit.yaml
martinpitt, line 18: @allisonkarlitskaya wants this gone. So let's file a packit issue and link it from here.

TomasTomecek: +1, please let us know what would be a better solution for you as we recently did an overhaul to the files_to_sync functionality.

martinpitt:
@allisonkarlitskaya : What about this?

 - we go back from 0 to 1 release secrets: generate a GPG key pair for signing releases
 - private key → `release` environment
 - public key → commit to Fedora dist-git
 - release workflow/cockpituous container: generate detached signature and add it as a release artifact (cockpit-N.tar.xz.asc)
 - packit `create-archive:`: download public key from Fedora dist-git, download signature, check sig

With that, an attacker would have to capture both Fedora's dist-git and our github repo. This isn't hard to implement, we retain "what do we trust" on the Fedora side, and it would have more or less the same trust to the tarball as we have today.

martinpitt:
So my question is, @allisonkarlitskaya which attack scenario do you want to protect against? I.e. why do you think this new approach is worse than our current one? In other words, why do you trust a release tarball download from github less than a git clone from github (given that we don't sign any commits and not every release tag even, and don't enforce signatures either). I'd like to understand that first before thinking more about how to solve this. Thanks!

(And FTR: This isn't urgent in any way; I admittedly have quite some eagerness to do this, but in practice there's no hurry).

martinpitt:
Unblocking. Turns out that for `propose_upstream:`, packit already DTRT. I incorrectly assumed that releases to COPR would be similar to releases to Fedora, that's not the case. So fixing COPR is independent (but also much less visible/urgent). I'll just finish up the metadata (see TODO), then we can land this if/when we want.

packit.yaml
```diff
@@ -46,3 +49,27 @@ jobs:
...
       - centos-stream-8-x86_64
       - centos-stream-9-x86_64
+
+  - job: propose_downstream
+    trigger: release
+    metadata:
+      dist_git_branches:
+        - fedora-development
+        - fedora-35
+        - fedora-36
```
lachmanfrantisek, line 59: @martinpitt Do you want to be explicit here for some reason? (I am just interested since we added two new [aliases](https://packit.dev/docs/configuration/#aliases) recently and I hope people can pick any subset without hardcoding.)

martinpitt:
As this is now independent from copr, I think this is ready for reconsideration again.

packit.yaml
martinpitt: When I added that to umockdev, I hit an unlucky time -- this was broken in packit back then, see https://github.com/packit/packit/issues/1525. But this got fixed in the meantime. I sent https://github.com/martinpitt/umockdev/pull/184 to drop this, and will test it there. If/when successful, I'll drop it here as well.

.github/workflows/release.yml
martinpitt: Done as a separate commit. I found https://docs.github.com/en/rest/overview/permissions-required-for-github-apps which is what I had been looking for for a long time, to document what these permissions do exactly.

packit.yaml
martinpitt: Not necessary any more, dropped.

martinpitt: Worked, dropped.

martinpitt:
@allisonkarlitskaya : Unfortunately I had to dial down the secrets dropping, as we still need the SSH key for release-dockerhub and release-guide. But these are easy to replace with deploy keys, I'll do that as the next step.

.github/workflows/release.yml
```diff
@@ -10,6 +10,11 @@ jobs:
     environment: release
     container:
       image: ghcr.io/cockpit-project/release
+    permissions:
+      # create GitHub release
+      contents: write
+      # pul the release container
```
allisonkarlitskaya, line 16: pul!

For what it's worth, it also works fine without this, since it's public.

allisonkarlitskaya (approved):
I mean, really, this is fine, if you want.

.github/workflows/release.yml
martinpitt: D'oh, thanks! I dropped this bit and adjusted the commit message.

allisonkarlitskaya (approved):

PR #17344, opened 2022-05-16
storaged: Fix VDO for RHEL 9

martinpitt:
RHEL 9 dropped the legacy API and the `vdo` command line tool. This
caused the "Create LV" dialog to falsely claim that the "vdo" package
needs to be installed, even though it was already -- and then fail the
creation as the `vdo` program was not available.

This was hidden by the obsolete test skips from the times when RHEL 9
images really didn't have a kmod-kvdo yet, but that has existed for a
long time already.

The `vdo` package is only necessary for creating a new volume (the rest
is built into the kernel and LVM itself), so define a new
`lvm_create_vdo` feature which checks the availability of `vdoformat`.
Check this in the "Create LV" dialog as well.

### DIFF
--- a/pkg/storaged/client.js
+++ b/pkg/storaged/client.js
@@ -463,6 +463,12 @@ function init_model(callback) {
             });
     }
 
+    function enable_lvm_create_vdo_feature() {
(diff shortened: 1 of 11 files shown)


mvollmer (approved):
LGTM!

PR #17345, opened 2022-05-16
test: Adjust for rhel-8-7

martinpitt:
This temporarily makes the base-2 vs. base-10 special cases even more
ugly. But we'll be able to drop all of that soon after landing the first
update in RHEL 8.7, and dropping support for testing on 8.6.


marusak (approved):
Thanks!

Issue #17352, opened 2022-05-18
translation error


zhangke-zk:
### Explain what happens

file  po/zh_CN.po translation error

### Where is the problem in Cockpit?

Documentation

### Server operating system

Red Hat Enterprise Linux

### What browsers are you using?

Firefox

PR #17353, opened 2022-05-18
翻译错误

zhangke-zk:
#17352


marusak:
Thank you!
This was actually fuzzy so it was not being used. But we need to translate this in weblate. I did this in your stead, see https://translate.fedoraproject.org/projects/cockpit/main/zh_CN/#history

If you are interested to do more tralsations you should use Weblate - which is also rather nice to use for translators :) Thank you!

PR #17365, opened 2022-05-20
New release workflow

allisonkarlitskaya:
Let's move away from cockpituous-release and use a workflow instead.

I installed deploy keys into my repository that would allow me to push to various cockpit repos from there, and this worked, and produced commits in the flathub and website repositories properly.

https://github.com/allisonkarlitskaya/cockpit/runs/6531613154?check_suite_focus=true

The release is on my page here: https://github.com/allisonkarlitskaya/cockpit/releases/tag/999

I force-pushed to remove the change to the website.

The branch created by the flathub process is there: https://github.com/cockpit-project/org.cockpit_project.CockpitClient/tree/999

The node-cache thing failed, but only because it tried to push to my fork instead of the cockpit one, and I set the keys up for the cockpit one.  I didn't really change anything major there anyway.

I feel reasonably comfortable merging this now.

 - [x] land, **and run** cockpit-project/bots#3413
 - [x] manually delete the release environment: https://github.com/cockpit-project/cockpit/settings/environments/


.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+
 jobs:
-  cockpituous:
+  source:
     runs-on: ubuntu-latest
-    environment: release
     container:
-      image: ghcr.io/cockpit-project/release
+      image: ghcr.io/cockpit-project/unit-tests
+      options: --user root
```
martinpitt, line 14: Why do we need root here? This is normally done to make it easier to share files between the host and the container, but that's not what happens here.

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
     permissions:
-      # create GitHub release
```
martinpitt, line 14: I find it nice to justify the permissions with comments. Did you drop that on purpose or by accident?

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
       contents: write
     steps:
-      - name: Set up configuration and secrets
+      - name: Clone repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Workaround for actions/checkout#697
```
martinpitt, line 23: Please specify the full URL https://github.com/actions/checkout/pull/697 here (took me a short while to find this).

I wonder why we didn't stumble over this before?

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+      - name: Get version number
+        run: git describe | tee version
+
+      - name: Extract tag body
+        run: git tag --points-at "${GITHUB_REF}" --format '%(contents:body)' | grep . | tee release-body
+
+      - name: Bootstrap automake
+        run: ./autogen.sh
+
+      - name: XXX JUMPSTART
```
martinpitt, line 35: XXX FIXME :grin:

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+            }
+
+            const update = await github.repos.updateRelease({
+              owner: context.repo.owner, repo: context.repo.repo, release_id: release.data.id,
+              draft: false});
+
+            core.setOutput('checksum', require('crypto').createHash('sha256').update(source).digest('hex'));
+            core.setOutput('download', update.data.assets[0].browser_download_url);
+            core.setOutput('filename', source_filename);
+            core.setOutput('version', version);
```
martinpitt, line 83: I was about to complain that the steps above like "Note source archive name" should set their outputs by itself. But indeed this looks better than the `echo "::set-output name` weirdness.

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+      download: ${{ steps.publish.outputs.download }}
+      filename: ${{ steps.publish.outputs.filename }}
+      version: ${{ steps.publish.outputs.version }}
+
+
+  guide:
+    needs: source
+    environment: website
+    permissions:
+      pull-requests: write
```
martinpitt, line 96: Why? This directly commits the update, no PR involved. This job does not even use `secrets.GITHUB_TOKEN`. As this uses a deploy key, `permissions: {}` ought to work?

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/cockpit-project/unit-tests
+      options: --user root
```
martinpitt, line 100: Same question as above.

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+          repository: cockpit-project/cockpit-project.github.io
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
+
+      - name: Download source release
+        run: curl -L -o '${{ needs.source.outputs.filename }}' '${{ needs.source.outputs.download }}'
+
+      - name: Verify checksum
+        run: echo '${{ needs.source.outputs.checksum }} ${{ needs.source.outputs.filename }}' | sha256sum -c
+
+      - name: Build guide
```
martinpitt, line 115: I'd prefer keeping tools/release-guide, and just calling that from the workflow with passing the `outputs` variables. It's so much easier to locally run and debug a shell script than a workflow..

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+          git commit --message='Update guide to version ${{ needs.source.outputs.version }}'
+          git show --stat
+          git push origin main
+
+
+  flathub:
+    needs: source
+    environment: flathub
+    permissions:
+      pull-requests: write
```
martinpitt, line 151: Dito

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+        uses: actions/checkout@v3
+        with:
+          path: src
 
-      - name: Run cockpituous
+      - name: Checkout flathub repository
+        uses: actions/checkout@v3
+        with:
+          path: flathub
+          repository: flathub/org.cockpit_project.CockpitClient
```
martinpitt, line 163: We can rely on actions/checkout providing this "flathub" alias for us? Still feels easier to understand and reproduce to just spell it out here.

.github/workflows/release.yml
```diff
@@ -4,46 +4,197 @@ on:
...
+          git checkout -b "${VERSION}"
+          git add "$(../src/containers/flatpak/prepare "${DOWNLOAD}" "${CHECKSUM}")"
+          git commit -m "Update to version ${VERSION}"
+          git show
+          git push git@github.com:cockpit-project/org.cockpit_project.CockpitClient HEAD
+
 
   release-node-cache:
+    # doesn't depend on these, but let's make sure the build passes before we do this
+    needs: [source, flathub, guide]
```
martinpitt, line 189: Bikeshedding, but IMHO this should only depend on `source`. Once the github release is done, the cache should be tagged as well. flathub and guide are follow-up actions, even if they fail, we should have a consistent github release.

martinpitt (changes requested):
Thanks! I'm looking forward to this!

.github/workflows/release.yml
allisonkarlitskaya: Same story here as cockpit-projects/bots#3365.  It's standard practice.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

says

> `/github/workspace` - **Note**: GitHub Actions must be run by the default Docker user (root). Ensure your Dockerfile does not set the `USER` instruction, otherwise you will not be able to access `GITHUB_WORKSPACE`.

.github/workflows/release.yml
allisonkarlitskaya: thx :)

.github/workflows/release.yml
allisonkarlitskaya: Done.

.github/workflows/release.yml
allisonkarlitskaya: Same answer as above.  :grin:

.github/workflows/release.yml
allisonkarlitskaya: Let's talk about this.

.github/workflows/release.yml
allisonkarlitskaya: I don't understand the question...

.github/workflows/release.yml
allisonkarlitskaya: Changed.

I actually have an idea to draft the release, do all of the other work, and then finalize the release once it all succeeds.  That would be a nice follow-up.

.github/workflows/release.yml
martinpitt: Err, nevermind; I thought "flathub" was just some alias for git.flathub.org/blah/..., but obviously it's just a github org.

.github/workflows/release.yml
allisonkarlitskaya: We're going to handle this as a follow-up, possibly using the new unit tests container script from #17304.

martinpitt (approved):
Ack, as I said the release-guide stuff is more like bikeshedding, let's not block on this.

Go, go, go! :rocket:

PR #17376, opened 2022-05-23
metrics: lookup podman container names

jelly:
As follow up of showing podman containers in the CPU usage overview, map
container ids to names. The cgroup holds no information on the name of
the podman container, so to look up the name we have to shell out to the
podman command line utility. As shelling out to podman is expensive we
only look up podman container mappings when a container makes it in top
cpu usage list. Multiple users can run a podman containers on a system
so we look up the userid based on the cgroup path and use sudo to
execute podman.


jelly:
There might be some UX issues left for when no mapping is found, we probably want to show something, maybe?

Or when you are in limited access and click on a root container you'll see nothing. We could map the curent user == containerid and then show something.

pkg/metrics/metrics.jsx
```diff
@@ -391,37 +399,62 @@ class CurrentMetrics extends React.Component {
...
+            } else {
+                return containerid.substr(0, 12);
+            }
+        }
+
+        function cgroupClickHandler(name, is_user, is_container, uid, podNameMapping) {
             if (is_container) {
-                cockpit.jump("/podman");
+                if (podNameInCache(podNameMapping, uid, name)) {
+                    cockpit.jump("/podman#/?name=" + podName(podNameMapping, uid, name));
```
jelly, line 441: This would show a link for users which can not be viewed in cockpit-podman for example:

I have a user named `jelle` who runs a container, with administrative access I can see their cgroups but cockpit-podman only shows your containers + administrative access containers, so the links don't really work.

As we know the userid, we could check if the uid is 0 or current logged and then show a link. Otherwise show for who the container owner is (that would require a username lookup) and ofcourse has to be done in cgroupRow.

garrett:
> There might be some UX issues left for when no mapping is found, we probably want to show something, maybe?

Yep. There should be some kind of fallback. What information is available in these cases? User & ID? Anything else?

> Or when you are in limited access and click on a root container you'll see nothing. We could map the curent user == containerid and then show something.

Yeah. I guess if there wouldn't be a way to see it in the Podman page due to limited access, we shouldn't link it? Otherwise it's just weird, right?

jelly:
> > There might be some UX issues left for when no mapping is found, we probably want to show something, maybe?
> 
> Yep. There should be some kind of fallback. What information is available in these cases? User & ID? Anything else?
> 
> > Or when you are in limited access and click on a root container you'll see nothing. We could map the curent user == containerid and then show something.
> 
> Yeah. I guess if there wouldn't be a way to see it in the Podman page due to limited access, we shouldn't link it? Otherwise it's just weird, right?

We could not link it yes, so if I log in as `jelle` we should only show containers for me *or* if I have privileged access `root` containers. But I think showing a tooltip with the container owner would be nice, but it does require another cockpit.spawn for `id -un -- "$cheruid"`.

martinpitt:
User containers all run inside the user session. I just verified that an user container which appears in the top list also appears as user@ID.service:

image

So perhaps that whole "su to different user" is just overkill? We don't have any UI to display containers other than the system and your own ones anyway. So if we'd restrict ourselves to just these two, and ignore the other ones, that might even suffice. Knowing that "it's user 1234's fault" is a good enough hint for the admin, and you can't control this kind of usage through system-level administration (like disabling or reconfiguring systemd units) anyway.

pkg/metrics/metrics.jsx
```diff
@@ -229,6 +232,7 @@ class CurrentMetrics extends React.Component {
             netInterfacesTx: [],
             topServicesCPU: [], // [ { name, percent } ]
             topServicesMemory: [], // [ { name, bytes } ]
+            podNameMapping: {}, // { cgroup, name }
```
martinpitt, line 235: This is a map, so the comment should be `cgroup → name`

pkg/metrics/metrics.jsx
```diff
@@ -391,37 +400,62 @@ class CurrentMetrics extends React.Component {
...
         }
 
-        function cgroupClickHandler(name, is_user, is_container) {
+        function podNameInCache(podNameMapping, uid, containerid) {
+            if (uid in podNameMapping) {
+                return containerid in podNameMapping[uid];
+            } else {
+                return false;
+            }
+        }
```
martinpitt, line 429: This should be `return podNameMapping[uid] && podNameMapping[uid][containerid];`. Then you'll get either `undefined` or the actual name back, which works just as well in a boolean context. And you can re-use this for `podName()` more easily, to avoid doing the lookup twice. I.e. I suggest to rename this to `getCachedPodName()`

Also, I don't see a reason to explicitly pass `podNameMapping` here -- it's always going to be `this.state.podNameMapping` anyway, isn't it? To use this, just turn this from a `function` into a `const` arrow.

pkg/metrics/metrics.jsx
```diff
@@ -391,37 +400,62 @@ class CurrentMetrics extends React.Component {
...
+
+        function podName(podNameMapping, uid, containerid) {
+            if (podNameInCache(podNameMapping, uid, containerid)) {
+                return podNameMapping[uid][containerid];
+            } else {
+                return containerid.substr(0, 12);
+            }
```
martinpitt, line 436: Then this is just `getCachedPodName(uid, containerid) || containerid.substr(0, 12)`.

pkg/metrics/metrics.jsx
martinpitt: I'd actually prefer to not show these at all. If you still want to, I don't veto this, though -- then for that I like your suggestion.

pkg/metrics/metrics.jsx
```diff
@@ -391,37 +400,62 @@ class CurrentMetrics extends React.Component {
...
+                } else {
+                    cockpit.jump("/podman");
+                }
             } else {
                 cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : ""));
             }
         }
 
-        function cgroupRow(name, value, is_user, is_container) {
+        function cgroupRow(name, value, is_user, is_container, uid, podNameMapping) {
```
martinpitt, line 451: Again, if you make this a const arrow, you don't need to pass podNameMapping here at all, but just access it on this.state.

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +474,68 @@ class CurrentMetrics extends React.Component {
...
-            ([key, value, is_user, is_container]) => cgroupRow(key, cockpit.format_bytes(value), is_user, is_container)
+        const topServicesMemory = n_biggest(this.cgroupMemoryNames, this.samples[10], 5);
+        newState.topServicesMemory = topServicesMemory.map(
+            ([key, value, is_user, is_container, userid]) => cgroupRow(key, cockpit.format_bytes(value), is_user, is_container, userid, this.state.podNameMapping)
         );
 
+        const notMappedContainers = topServicesMemory.concat(topServicesCPU).filter(([key, value, is_user, is_container, userid]) => is_container && !podNameInCache(this.state.podNameMapping, userid, key));
+        if (notMappedContainers.length !== 0) {
+            this.update_podman_name_mapping(notMappedContainers);
+        }
```
martinpitt, line 490: You need to take negative cache lookups into account here. I.e. if you previously failed to map an ID to a pod name (perhaps it's a docker/LXC/etc. container or custom cgroup), the cache must have an entry for that ID, mapping it to `null` (whereas "not attempted yet" would be `undefined` or even "key does not exist", which is by and large `undefined` as well). This should then *not* cause another cache refresh.

That should work fine wiht the cache building below, as that only ever updates podNameMapping instead of rewriting it.

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +474,68 @@ class CurrentMetrics extends React.Component {
...
+     * ps for that user.
+     */
+    update_podman_name_mapping = (cgroups) => {
+        const { privileged } = this.state;
+        const spawnPromise = (cmd, uid) => {
+            return new Promise((resolve, reject) => {
+                cockpit.spawn(cmd, { superuser: privileged ? "required" : null }).then(result => {
+                    resolve([result, uid]);
+                }).catch(err => reject(err));
+            });
```
martinpitt, line 506: Unnecessary Promise wrapper, just return cockpit.spawn(). (or more precisely, the chained promise after .then()/.catch().

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +474,68 @@ class CurrentMetrics extends React.Component {
...
+        };
+
+        const promises = cgroups.map(([key, value, is_user, is_container, userid]) => {
+            const cmd = ["podman", "ps", "--format", "json"];
+            if (userid === '0') {
+                return privileged ? spawnPromise(cmd, userid) : null;
+            } else {
+                if (privileged) {
+                    return spawnPromise(["sudo", "-u", "#" + userid].concat(cmd), userid);
```
martinpitt, line 515: This is actually rather clever, if we really want to go ahead with mapping arbitrary user IDs.

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +474,68 @@ class CurrentMetrics extends React.Component {
...
+            const podNameMapping = {};
+            for (const [output, uid] of results) {
+                podNameMapping[uid] = {};
+                try {
+                    const containers = JSON.parse(output);
+                    for (const container of containers) {
+                        podNameMapping[uid][container.Id] = container.Names[0];
+                    }
+                } catch (err) {
+                    console.error("podman ps outputs invalid JSON", err.toString());
```
martinpitt, line 532: In case that ever happens, I think this needs to set some flag (in the negative cache or separate) to entirely stop this polling, otherwise you'll poll in vain for each iteration.

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +474,68 @@ class CurrentMetrics extends React.Component {
...
+                }
+            }
+            this.setState({ podNameMapping: Object.assign(podNameMapping, this.state.podNameMapping) });
```
martinpitt, line 535: OK, just FYI: In ES6 you can use the spread operator for this: `{...podNameMapping, ...this.state.podNameMapping}`.

martinpitt (changes requested):
Thanks! Happy to gmeet if you want to discuss the approach with more bandwidth.

pkg/metrics/metrics.jsx
```diff
@@ -217,6 +219,7 @@ class CurrentMetrics extends React.Component {
         this.cgroupMemoryNames = [];
 
         this.state = {
+            privileged: false,
```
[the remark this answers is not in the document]
jelly: Ah, well we do need to keep track of it right or do we handle this cache by switching to superuser / dropping permissions reloading the page?

Took it from https://github.com/cockpit-project/cockpit/blob/main/pkg/shell/hosts.jsx#L73

pkg/metrics/metrics.jsx
jelly: Nice, that's a good cleanup!

pkg/metrics/metrics.jsx
jelly: This is now fixed by using `getCachedPodName` everywhere

pkg/metrics/metrics.jsx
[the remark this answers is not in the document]
martinpitt: hosts.jsx looks even weirder -- nothing actually uses that state, so dead code?

Sure you need to react on a change (that's the empty setState() below), but the code can look at `superuser.allowed` directly.

pkg/metrics/metrics.jsx
jelly: I'm ok with leaving it out, having a follow up which explains what `user@1000` is, so ideally it would be: `user session for jelle` but maybe that's too long already.

martinpitt: How about "user jelle", similar to "pod foo"? But yes, this is follow-up material, as it's unrelated to this PR.

jelly:
> User containers all run inside the user session. I just verified that an user container which appears in the top list also appears as [user@ID.service](mailto:user@ID.service):
> 
> image
> 
> So perhaps that whole "su to different user" is just overkill? We don't have any UI to display containers other than the system and your own ones anyway. So if we'd restrict ourselves to just these two, and ignore the other ones, that might even suffice. Knowing that "it's user 1234's fault" is a good enough hint for the admin, and you can't control this kind of usage through system-level administration (like disabling or reconfiguring systemd units) anyway.

I've updated the PR and simplified it a lot. We now only show the root and logged in users containers, for root we fallback to the podman id if the user is not a superuser.

This works pretty well, the only thing missing is to squash the commits and add a test. This would be easy as it just requires waiting one render longer for the name to be resolved.

pkg/metrics/metrics.jsx
```diff
@@ -391,37 +400,56 @@ class CurrentMetrics extends React.Component {
...
             return merged.slice(0, n);
-        }
+        };
+
+        const getCachedPodName = (uid, containerid) => {
+            return this.state.podNameMapping[uid] && this.state.podNameMapping[uid][containerid];
+        };
 
-        function cgroupClickHandler(name, is_user, is_container) {
+        function cgroupClickHandler(name, is_user, is_container, uid) {
```
martinpitt, line 432: I'm surprised that this works. Doesn't the `function` here introduce a new `this` context, and thus getCachedPodName() can't access this.state any more?

pkg/metrics/metrics.jsx
```diff
@@ -391,37 +400,56 @@ class CurrentMetrics extends React.Component {
...
+                    }
                 }
             });
             merged.sort((a, b) => b[1] - a[1]);
             return merged.slice(0, n);
-        }
+        };
+
+        const getCachedPodName = (uid, containerid) => {
+            return this.state.podNameMapping[uid] && this.state.podNameMapping[uid][containerid];
```
martinpitt, line 429: Can you please drop the `{ return }` wrapper and just directly return the value?

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +468,65 @@ class CurrentMetrics extends React.Component {
...
+        if (notMappedContainers.length !== 0) {
+            this.update_podman_name_mapping(notMappedContainers);
+        }
         this.setState(newState);
     }
 
+    /**
+     * Look up the container names using podman ps for the given cgroups. When
+     * the logged in user has administrative access we use sudo to call podman
+     * ps for that user.
```
martinpitt, line 491: Outdated comment

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +468,65 @@ class CurrentMetrics extends React.Component {
...
+            if (!(userid in podNameMapping)) {
+                podNameMapping[userid] = {};
+            }
+            // Always initialize the cache for when we hit an error.
+            podNameMapping[userid][containerid] = null;
+
+            if ((userid === 0 && !superuser.allowed) && userid !== this.state.userid) {
+                return null;
+            }
+            return cockpit.spawn(["podman", "ps", "--format", "json"], { superuser: userid === 0 ? "required" : null }).then(result => {
```
martinpitt, line 507: Please move the .then() to the next line, it's easy to miss. Please also drop the `{ return }` wrapper.

pkg/metrics/metrics.jsx
```diff
@@ -440,17 +468,65 @@ class CurrentMetrics extends React.Component {
...
+                    for (const container of containers) {
+                        podNameMapping[uid][container.Id] = container.Names[0];
+                    }
+                } catch (err) {
+                    console.error("podman ps outputs invalid JSON", err.toString());
+                }
+            }
+            this.setState({ podNameMapping: { ...podNameMapping, ...this.state.podNameMapping } });
+        }).catch(err => {
+            this.setState({ podNameMapping: { ...podNameMapping, ...this.state.podNameMapping } });
```
martinpitt, line 525: DRY -- can you use `.finally()` for that?

test/verify/check-metrics
```diff
@@ -796,6 +796,11 @@ class TestCurrentMetrics(MachineCase):
             with b.wait_timeout(300):
                 b.wait_in_text("#current-metrics-card-cpu", f"pod {shortid}")
             b.wait(lambda: topServiceValue(self, "Top 5 CPU services", "%", 1) > 70)
+
+            # It takes one re-render for the name lookup
+            with b.wait_timeout(300):
```
martinpitt, line 801: ... but that sholdn't take 5 minutes surely? This is updated every 5 seconds, isn't it?

martinpitt (changes requested):
Thanks! Sorry for letting that slip. The approach looks fine to me, I just have some smaller technical quibbles. The PR is conflicty now (as you enabled this test on FCOS).

test/verify/check-metrics
jelly: Yup, so let's reduce it.

martinpitt:
@jelly This is almost certainly related to https://github.com/cockpit-project/cockpit/commit/b83d91e929e350 . It might be a systemd regression, but my bet is that it's an SELinux issue. Worth sending a no-test PR with `setenforce 0` to run on packit.

pkg/metrics/metrics.jsx
```diff
@@ -229,6 +232,7 @@ class CurrentMetrics extends React.Component {
             netInterfacesTx: [],
             topServicesCPU: [], // [ { name, percent } ]
             topServicesMemory: [], // [ { name, bytes } ]
+            podNameMapping: {}, // { containerid -> name }
```
martinpitt, line 235: I admittedly liked this better as `cgroup`, but as it's now a two-level dict, the comment is still wrong. It's "uid → containerid → name", isn't it?

test/verify/check-metrics
```diff
@@ -793,6 +793,11 @@ class TestCurrentMetrics(MachineCase):
         with b.wait_timeout(300):
             b.wait_in_text("#current-metrics-card-cpu", f"pod {shortid}")
```
martinpitt, line 794: This is now a race condition. It could be that the name lookup is fast enough that we don't see the short ID on the page ever. But *shrug*, it seems to work in CI, so fine for now. Just something to watch out for later.

martinpitt (approved):
Thanks! I don't have the heart to nack this just because of the wrong comment, so please go ahead! Thanks for bearing with me :grin:

pkg/metrics/metrics.jsx
jelly: It is still wrong :|, it is indeed uid => containerid => name

martinpitt (approved):
Cheers!

martinpitt:
@jelly : Also, this c9s TF run has worked fine :man_shrugging:

PR #17379, opened 2022-05-24
test: Clean up rhel-8-6 and rhel-8-7-distropkg for storage tests

mvollmer:
rhel-8-6 is no longer supported, and rhel-8-7-distropkg now uses
base-10 units.


martinpitt (approved):
Cheers! 8-7-dp passed, I triggered the required fedora-coreos, rhel-8-7 and a fedora-35 for good measure.

PR #17380, opened 2022-05-24
tuned: Use WithDialogs protocol

mvollmer:


martinpitt (approved):
Thanks! I was going to ask about `<WithDialogs>`, but of course this is embedded into the Overview which already has that. Nice!

PR #17382, opened 2022-05-24
Work around WebKit text selection bug in flex and grid

garrett:
TODO:

- [ ] Report upstream to PatternFly
- [ ] Add link to upstream bug in workarounds file


garrett:
Closing this. It's fixed upstream and Cockpit Client will likely get the fix in days. Safari should have the fix. Distributions will probably upgrade WebKit to the latest versions soon too. (Arch has already upgraded; the rest are probably soon.)

PR #17386, opened 2022-05-25
test: Increase timeout for sos report building

martinpitt:
Despite only runnning 5 plugins, the sos report command now still takes
some non-trivial time due to the obfuscation and encryption. It
times out in RHEL 8.7 gating, so let's bump the timeout some more.


martinpitt:
Seen in https://src.osci.redhat.com/rpms/cockpit/pull-request/104

martinpitt:
Hmm no, even with `export TEST_TIMEOUT_FACTOR=2` it still fails downstream, this is something else.

PR #17389, opened 2022-05-25
test: output full path to coverage html

jelly:
With the move to s3 for logs.html we no longer can list directories so
coverage needs to link to an absolute path.


allisonkarlitskaya (approved):
Works for me.  I like this more than the other approach.

PR #17390, opened 2022-05-25
test: enable podman cpu metric tests for coreos

jelly:
The Fedora coreos image now has a busybox container, so these tests can
be enabled.


martinpitt (approved):
Nice, thank you!

PR #17394, opened 2022-05-27
test: Only assert type of cancelled shutdown

mvollmer:
The time doesn't matter when there is no scheduled shutdown, and it
used to be zero but seems to have changed to 18446744073709551615 with
systemd 251.

### DIFF
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -869,7 +869,7 @@ password=foobar
         b.click("#system-health-shutdown-status-cancel-btn")
         b.wait_not_present('#system-health-shutdown-status')
         dbus_call = 'busctl get-property org.freedesktop.login1 /org/freedesktop/login1  org.freedesktop.login1.Manager ScheduledShutdown'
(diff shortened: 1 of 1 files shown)


mvollmer:
systemd 251 comes with arch already, see https://cockpit-logs.us-east-1.linodeobjects.com/pull-3434-20220527-075748-12c461c1-arch-cockpit-project-cockpit/log.html#254-2

marusak (dismissed):
If tests agree, so do I, thanks! :)

mvollmer:
Bah, busctl on rhel 8 doesn't have "-j"...

marusak (approved):

mvollmer:
`fedora-36/container-bastion` runs into https://github.com/cockpit-project/bots/issues/2787 all the time now, but we don't check for naughties with the `container-*` scenarios.  This is being fixed in https://github.com/cockpit-project/cockpit/pull/17396.  Let's just ignore it here.

PR #17396, opened 2022-05-27
test: Get rid of test/containers/run-test

mvollmer:
Use test/common/run-test instead so that we can benefit from all its
features, especially naughty patterns.


mvollmer:
Nice, this also gives us a proper test summary back, there must have been something broken in how `test/containers/run-tests` was outputting the TAP markup.

mvollmer:
Nice, this also gives us a proper test summary back, there must have been something broken in how `test/containers/run-tests` was outputting the TAP markup.

martinpitt:
For some reason the [HTML log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17396-20220527-131312-aca62678-fedora-36-container-bastion/log.html) is empty -- but I think I've seen that in other occasions recently. The raw log looks fine.

Thanks @mvollmer! I ignored this so far as I am very close to getting rid of test/containers/ completely, by absorbing the bastion container into cockpit/ws. But I'm glad that this is relatively simple to fix!

Triggered the required fedora-coreos test.

martinpitt (approved):

PR #17398, opened 2022-05-30
containers/ws: Move os-release symlinking into run script

martinpitt:
This prepares the ws container for being able to run unprivileged.
Without /host, the branding will be completely broken, and trying to
run it with a customized /etc/os-release volume (aka bind mount) will
fail.

Only do the linking if /etc/os-release is not already mounted as a
volume.

---

The full support for unprivileged ws will come in a follow-up PR. But this particular bit is awkward to work around especially for our offline "update existing container" image-prepare, so I'd like to land this first, refresh the official container, and refresh our fedora-coreos image with that (although that step is blocked by https://github.com/cockpit-project/bots/pull/3437)

### DIFF
--- a/containers/ws/install.sh
+++ b/containers/ws/install.sh
@@ -38,6 +38,3 @@ rm -rf /container/rpms || true
 # And the stuff that starts the container
 ln -s /host/proc/1 /container/target-namespace
 chmod -v +x /container/label-install /container/label-uninstall /container/label-run
-
-# Make the container think it's the host OS version
-rm -f /etc/os-release /usr/lib/os-release && ln -sv /host/etc/os-release /etc/os-release && ln -sv /host/usr/lib/os-release /usr/lib/os-release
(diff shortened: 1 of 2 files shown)


martinpitt:
I locally built a "clean" container from this (`podman build -t cockpit/ws containers/ws`), copied the container to our fedora-coreos image (`podman save localhost/cockpit/ws | ssh c podman load`), ran our tests against it, and also did an interactive check that the branding was as expected.

```
# podman exec -it ws ls -l /etc/os-release 
lrwxrwxrwx. 1 root root 20 May 30 03:39 /etc/os-release -> /host/etc/os-release

# podman exec -it ws head -n5 /etc/os-release 
NAME="Fedora Linux"
VERSION="36.20220505.2.0 (CoreOS)"
ID=fedora
VERSION_ID=36
VERSION_CODENAME=""
```

marusak (approved):
Thanks!

PR #17399, opened 2022-05-30
git-utils.sh: pre-init the cache directory for untar

allisonkarlitskaya:
Trying to call `git init` with the GIT_WORK_TREE variable set gives the
following error:

  fatal: GIT_WORK_TREE (or --work-tree=<directory>) not allowed without specifying GIT_DIR (or --git-dir=<directory>)

So call init_cache before we set the variables to make sure the
directory already exists.

### DIFF
--- a/tools/git-utils.sh
+++ b/tools/git-utils.sh
@@ -120,6 +120,9 @@ unpack_from_cache() {
 # This stores a .tar file from stdin into the cache as a tree object.
 # Returns the ID.  Opposite of `git archive`, basically.
 tar_to_cache() {
+    # Need to do this before we set the GIT_* variables
+    init_cache
+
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
The bug can be seen here: https://github.com/cockpit-project/cockpit/runs/6648193909?check_suite_focus=true#step:3:68

Retrying the run from this branch works: https://github.com/cockpit-project/cockpit/actions/runs/2409122533

martinpitt:
Created https://github.com/cockpit-project/cockpit/pull/17400 , which now failed during package build (as expected, new PF release is broken). Thanks!

martinpitt (approved):
Danke!

PR #17403, opened 2022-05-30
workflow: Move release to centralized action

martinpitt:
Creating a GitHub release is complex enough to factorize this into a
central location: https://github.com/cockpit-project/action-release
This is not even specific to cockpit projects.

----

I tested this on my fork with an [extra debug commit](https://github.com/martinpitt/cockpit/commit/913df23ca0f878684bbd8ab529a1aebb2850112c) that disables flathub, node-cache, and packit. It only keeps the upstream github and the guide. 

 - [270.1 release](https://github.com/martinpitt/cockpit/releases/tag/270.1)
 - [workflow run](https://github.com/martinpitt/cockpit/actions/runs/2409887114)

The guide step fails on the expected place (not being able to push to the website repo), but tarball download and checksum validation succeeded.


.github/workflows/release.yml
```diff
@@ -24,68 +24,27 @@ jobs:
...
-
       - name: Bootstrap automake
         run: ./autogen.sh
 
       - name: Build release
         run: make dist -j$(nproc)
 
       - name: Note source archive name
-        run: basename "$(make dump-dist)" | tee source-filename
+        id: source_filename
```
allisonkarlitskaya, line 34: I wonder if you'd like to drop this step and use something like `archive: cockpit-${{ github.ref }}.tar.xz` or so?

martinpitt: `github.ref` is something like `refs/tag/123.4`. So this needs some extra string mangling using jinja2 API. I haven't done this before, let me check if there is something.

martinpitt: I dug a bit, and I didn't find anything.

martinpitt: Done -- it's using `github.ref_name` now.

martinpitt:
@allisonkarlitskaya : Updated as per our discussion, re-tested, links in description updated.

allisonkarlitskaya (approved):

Issue #16780, opened 2022-01-05
Allow for favoriting services


queengooborg:
First off, I must say: great job, Cockpit project team!  I really love how straightforward and easy to use the UI is -- quite a refreshing experience after struggling with the complexity of Zabbix!  Keep up the great work, everyone!  💜 

I had a feature suggestion that I was hoping you guys would be interested in.  I'd love to have the ability to favorite or pin services to the top of the list, so I can quickly see them all without having to scroll through the system-defined services.  I grabbed Cockpit as a way to quickly manage the various gaming servers I am hosting (which has been great, now that I've tied them all into `systemctl` controlled services), but I typically keep two or three on at a time.  Having them at the top of the list, or easily filterable, would be great for quick access and monitoring of them!

queengooborg:
Thanks for the quick response!  For my use case, I am commonly starting and stopping these particular services manually, and so I'd love to have quick access to trigger them at the top of the page.  I think that it makes sense to place the "pin" action in the menu on the service's page -- users are already expecting to control their services from there!

martinpitt:
Honestly this sounds more like a per-user preference, and should go into localstore -- like language selection or the list of recent hosts. There isn't anything "systemic" about these which would justify to create lots of little drop-in files.

PR #17406, opened 2022-05-31
services: Add pinning services

tomasmatus:
Fixes: #16780

This uses localStorage to store paths of every pinned unit.
There is no way to tell appart unit aliases because stored objects
are identical so as a side effect all aliases refering to the same
path end up being pinned.

Pinning unit:
pin_unit_pin

Unpinning unit:
pin_unit_unpin

Pinned unit is on the top of the list:
pin_unit

Failed units are above pinned units:
pin_unit_with_failed

Tooltip:
pin_unit_tooltip

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -32,7 +32,7 @@ import {
 import {
     AsleepIcon,
     BanIcon, ErrorCircleOIcon, OnRunningIcon, OffIcon,
-    OkIcon, UserIcon,
+    OkIcon, UserIcon, ThumbtackIcon,
 } from "@patternfly/react-icons";
 
 import cockpit from "cockpit";
@@ -89,11 +89,13 @@ const ServiceConfirmDialog = ({ id, title, message, confirmText, confirmAction }
  * React template for showing possible service action (in a kebab menu)
  * Required props:
  *  - masked
- *     Unit is masked
+ *      Unit is masked
  *  - active
- *     Unit is active (running)
+ *      Unit is active (running)
  *  - failed
- *     Unit has failed
+ *      Unit has failed
+ *  - isPinned
+ *      Unit is pinned
  *  - canReload
  *      Unit can be reloaded
  *  - actionCallback
@@ -102,10 +104,12 @@ const ServiceConfirmDialog = ({ id, title, message, confirmText, confirmAction }
  *      Method for calling unit file methods like `EnableUnitFiles`
  *  - deleteActionCallback
  *      Method for calling deleting the systemd unit
+ *  - pinUnitCallback
+ *      Method to pin unit
  *  - disabled
  *      Button is disabled
  */
-const ServiceActions = ({ masked, active, failed, canReload, actionCallback, deleteActionCallback, fileActionCallback, disabled }) => {
+const ServiceActions = ({ masked, active, failed, canReload, actionCallback, deleteActionCallback, fileActionCallback, disabled, isPinned, pinUnitCallback }) => {
     const Dialogs = useDialogs();
     const [isActionOpen, setIsActionOpen] = useState(false);
 
@@ -169,6 +173,11 @@ const ServiceActions = ({ masked, active, failed, canReload, actionCallback, del
         actions.push(
             <DropdownItem key="mask" onClick={confirm}>{ _("Disallow running (mask)") }</DropdownItem>
         );
+
+        actions.push(<DropdownSeparator key="pin-divider" />);
+        actions.push(
+            <DropdownItem key="pin" onClick={() => pinUnitCallback() }>{isPinned ? _("Unpin unit") : _("Pin unit")}</DropdownItem>
+        );
     }
 
     return (
@@ -216,6 +225,7 @@ export class ServiceDetails extends React.Component {
         this.unitFileAction = this.unitFileAction.bind(this);
         this.deleteAction = this.deleteAction.bind(this);
         this.deleteTimer = this.deleteTimer.bind(this);
+        this.pinUnit = this.pinUnit.bind(this);
 
         this.unitType = props.unit.Id.split('.').slice(-1)[0];
         this.unitTypeCapitalized = this.unitType.charAt(0).toUpperCase() + this.unitType.slice(1);
@@ -306,6 +316,24 @@ export class ServiceDetails extends React.Component {
         } else {
             return promise;
         }
+    }
+
+    pinUnit() {
+        let pinnedUnits = [];
+        try {
+            pinnedUnits = JSON.parse(localStorage.getItem('systemd:pinnedUnits')) || [];
+        } catch (err) {
+            console.warn("exception while parsing systemd:pinnedUnits", err);
+        }
+
+        if (this.props.isPinned) {
+            pinnedUnits = pinnedUnits.filter(unitId => unitId != this.props.unit.path);
+        } else {
+            pinnedUnits.push(this.props.unit.path);
+        }
+
+        localStorage.setItem('systemd:pinnedUnits', JSON.stringify(pinnedUnits));
+        dispatchEvent(new Event('storage'));
     }
 
     unitFileAction(method, force, catchExc = true) {
@@ -576,6 +604,10 @@ export class ServiceDetails extends React.Component {
                     : <>
                         <CardTitle className="service-top-panel">
                             <Text component={TextVariants.h2} className="service-name">{this.props.unit.Description}</Text>
+                            {this.props.isPinned &&
+                                <Tooltip content={_("Pinned unit")}>
+                                    <ThumbtackIcon className='service-thumbtack-icon' />
+                                </Tooltip>}
                             { showAction &&
                                 <>
                                     { !masked && !isStatic &&
@@ -591,7 +623,8 @@ export class ServiceDetails extends React.Component {
                                     <ServiceActions { ...{ active, failed, enabled, masked } } canReload={this.props.unit.CanReload}
                                                     actionCallback={this.unitAction} fileActionCallback={this.unitFileAction}
                                                     deleteActionCallback={isCustom && isTimer ? this.deleteAction : null}
-                                                    disabled={this.state.waitsAction || this.state.waitsFileAction} />
+                                                    disabled={this.state.waitsAction || this.state.waitsFileAction}
+                                                    isPinned={this.props.isPinned} pinUnitCallback={this.pinUnit} />
                                 </>
                             }
                         </CardTitle>
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -162,6 +162,13 @@ class ServicesPageBody extends React.Component {
             currentStatus: null,
         };
 
+        try {
+            this.state.pinnedUnits = JSON.parse(localStorage.getItem('systemd:pinnedUnits')) || [];
+        } catch (err) {
+            console.warn("exception while parsing systemd:pinnedUnits", err);
+            this.state.pinnedUnits = [];
+        }
+
         this.onCurrentTextFilterChanged = (currentTextFilter) => {
             this.setState({ currentTextFilter });
         };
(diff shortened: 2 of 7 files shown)


pkg/systemd/services/services.jsx
```diff
@@ -727,7 +736,6 @@ class ServicesPageBody extends React.Component {
     render() {
         const { unit_by_path } = this.state;
         const path = this.props.path;
-
```
skobyda, line 730: I think this line removal got here by accident

pkg/systemd/services/service-details.jsx
```diff
@@ -308,6 +324,21 @@ export class ServiceDetails extends React.Component {
         }
     }
 
+    pinUnit(method) {
+        let pinnedUnits = JSON.parse(localStorage.getItem('pinnedUnits')) || [];
+
+        this.setState({ waitsAction: true });
+        if (method === "Pin") {
+            pinnedUnits.push(this.props.unit.path);
+        } else if (method === "Unpin") {
```
skobyda, line 333: I think passing "method" argument to the `pinUnit` is unnecessary. It can be a boolean, or alternatively, since this method is only called when the user wants to negate the "pinned" status of a service, this function doesn't need any argument at all, you can just look at the "pinned" status of the service and negate it.

skobyda (changes requested):

pkg/systemd/services/service-details.jsx
tomasmatus: good point, fixed

garrett:
The pin icon on the right messes up the alignment of the service status. It's also not a service status, so I think it shouldn't go there.

It should probably be to the left of the name. Since we don't have a separate column for it, it should probably just indent the name a bit. (If we did have a separate column, it would be empty by default and would make all services have extra space to the left of the names — and we wouldn't want that.)

Therefore, I think it would be best like this instead:

image

Alternatively, it'd look like this, if the icon is to the right of the text:

image

(The icon itself should not be part of the linked text.)

I have the icon color simulating `pf-u-icon-color-light` (a text utility class). https://www.patternfly.org/v4/utilities/text/#icon-colors

A third option would be to have the icon to the left of the status, like this:

image

(I can't decide what's best out of these three at the moment.)

The icon, of course, should also have a tooltip on hover, regardless of where the icon is located.

jelly:
> Oh, this is great! Thanks for implementing it!
> 
> I have a bit of feedback regarding the pin's layout:
> 
> The pin icon on the right messes up the alignment of the service status. It's also not a service status, so I think it shouldn't go there.
> 
> It should probably be to the left of the name. Since we don't have a separate column for it, it should probably just indent the name a bit. (If we did have a separate column, it would be empty by default and would make all services have extra space to the left of the names — and we wouldn't want that.)
> 
> Therefore, I think it would be best like this instead:
> 
> image
> 
> Alternatively, it'd look like this, if the icon is to the right of the text:
> 
> image
> 

I prefer 1 or 2, leaning towards 2 as then you can quickly scan service names as they are all aligned correctly.

tomasmatus:
I've decided to go with 1, I've updated screenshots above.

garrett:
> I prefer 1 or 2, leaning towards 2 as then you can quickly scan service names as they are all aligned correctly.

Yeah, looking at this again, I think I also prefer 2.

skobyda (changes requested):

pkg/systemd/services/service-details.jsx
```diff
@@ -308,6 +324,22 @@ export class ServiceDetails extends React.Component {
...
+        if (this.props.unit.IsPinned) {
+            pinnedUnits = pinnedUnits.filter(unitId => unitId != this.props.unit.path);
+        } else {
+            pinnedUnits.push(this.props.unit.path);
+        }
+
+        localStorage.setItem('pinnedUnits', JSON.stringify(pinnedUnits));
+        dispatchEvent(new Event('storage'));
+        this.setState({ waitsAction: false });
+    }
```
skobyda, line 341: We set `waitsAction: true` only in cases when we don't want user to do any other action with the service, until the previous actions is finished, otherwise those 2 actions could trigger 2 conflicting API calls. I think setting `waitsAction` here is unecessary because:
1. It makes sense to set it only with asynchronous methods, which might take a long time. localStorage.setItem is however a synchronous method, so it resolves immediatelly and there is no time for user to do another action which would trigger some other API call.
2. Even if localStorage.setItem was an asynchronous method, it doesn't call systemd, it doesn't change configuration of the unit file, so there is no way `localStorage.setItem` would be conflicting with some other action/API call.

skobyda (changes requested):

pkg/systemd/services/service-details.jsx
skobyda: So I think this code can be like this:
```suggestion
        if (this.props.unit.IsPinned) {
            pinnedUnits = pinnedUnits.filter(unitId => unitId != this.props.unit.path);
        } else {
            pinnedUnits.push(this.props.unit.path);
        }

        localStorage.setItem('pinnedUnits', JSON.stringify(pinnedUnits));
        dispatchEvent(new Event('storage'));
    }
```

pkg/systemd/services/services.jsx
```diff
@@ -738,7 +747,8 @@ class ServicesPageBody extends React.Component {
             const unit_id = path[0];
             const get_unit_path = (unit_id) => this.path_by_id[unit_id];
             const unit_path = get_unit_path(unit_id);
-            const unit = this.state.unit_by_path[unit_path];
+            const unit = { ...this.state.unit_by_path[unit_path] };
+            unit.IsPinned = this.state.pinnedUnits.includes(JSON.stringify(unit.path));
```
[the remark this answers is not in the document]
tomasmatus: I remember having some problems with escaping characters but after changing the code it is no longer needed here. Good catch.

pkg/systemd/services/service-details.jsx
tomasmatus: I see, thank you for explaining when `waitsAction` should be used!

tomasmatus:
> [Arch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17406-20220608-134231-cd4d5e92-arch/log.html#298) and [debian-testing](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17406-20220608-134231-cd4d5e92-debian-testing/log.html) tests failed, since there are other failed service in the list before the pinned service.

yeah I can see that. Looks like this test isn't robust enough.

test/verify/check-system-services
```diff
@@ -397,7 +397,23 @@ OnCalendar=*:1/2
...
+        self.check_service_details(["Static", "Not running"], ["Start", "Disallow running (mask)", "Pin unit"], True, onoff=False)
+
+        # Pin unit
+        b.go(f'/system/services#/{suffix}')
+        b.wait_not_present('#test.service > svg')
+
+        self.goto_service("test.service")
+        self.do_action("Pin unit")
+        b.go(f'/system/services#/{suffix}')
+        b.is_present('#test.service > svg')
```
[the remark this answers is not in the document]
tomasmatus: added

pkg/systemd/services/services.jsx
```diff
@@ -160,6 +160,7 @@ class ServicesPageBody extends React.Component {
 
             error: null,
             currentStatus: null,
+            pinnedUnits: localStorage.getItem('pinnedUnits') || [],
```
jelly, line 163: This now returns a string || [] so if we want to load the state here it should use `JSON.parse`.

pkg/systemd/services/services.jsx
```diff
@@ -307,6 +308,10 @@ class ServicesPageBody extends React.Component {
                 this.listUnits();
         });
 
+        addEventListener('storage', () => {
+            this.setState({ pinnedUnits: localStorage.getItem('pinnedUnits') || [] });
```
jelly, line 312: Same as above

pkg/systemd/services/services.jsx
```diff
@@ -484,12 +489,16 @@ class ServicesPageBody extends React.Component {
         const unit_b = unit_b_t[1];
         const failed_a = unit_a.HasFailed ? 1 : 0;
         const failed_b = unit_b.HasFailed ? 1 : 0;
+        const pinned_a = this.state.pinnedUnits.includes(JSON.stringify(unit_a.path)) ? 1 : 0;
+        const pinned_b = this.state.pinnedUnits.includes(JSON.stringify(unit_b.path)) ? 1 : 0;
```
jelly, line 493: This would break if the changes above are applied.

pkg/systemd/services/services.jsx
```diff
@@ -811,6 +821,7 @@ class ServicesPageBody extends React.Component {
                         !filters.activeState.includes(this.activeState[unit.ActiveState]))
                         return false;
 
+                    unit.IsPinned = this.state.pinnedUnits.includes(unit.path);
```
jelly, line 824: This is interesting above includes uses `JSON.parse`, so how does this work?

jelly (changes requested):
The current localStorage reading/writing code seems a bit confusing, so would be good to re-check it.

I think it might be good to maybe prefix the localStorage item with `systemd:pinnedUnits` like `superuser`.

pkg/systemd/services/service-details.jsx
```diff
@@ -308,6 +318,19 @@ export class ServiceDetails extends React.Component {
         }
     }
 
+    pinUnit() {
+        let pinnedUnits = JSON.parse(localStorage.getItem('pinnedUnits')) || [];
```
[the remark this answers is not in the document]
tomasmatus: I cleared up the confusion where I was sometimes using strings and sometimes array and I've added try-catch every time JSON.parse is used for interacting with localStorage.

garrett:
@tomasmatus: That looks great! Thanks.

(FWIW: I don't like the jump of the widgets, but there's no good way to deal with it, aside from reserving space or changing the direction... and it's pretty minor. I think this is fine, really. :+1: )

pkg/systemd/services/services.jsx
```diff
@@ -307,6 +314,15 @@ class ServicesPageBody extends React.Component {
                 this.listUnits();
         });
 
+        addEventListener('storage', () => {
+            try {
+                this.setState({ pinnedUnits: JSON.parse(localStorage.getItem('systemd:pinnedUnits')) || [] });
+            } catch (err) {
+                console.warn(err);
```
jelly, line 321: Same as above, should have a message from where the error comes

pkg/systemd/services/services.jsx
```diff
@@ -738,7 +758,8 @@ class ServicesPageBody extends React.Component {
             const unit_id = path[0];
             const get_unit_path = (unit_id) => this.path_by_id[unit_id];
             const unit_path = get_unit_path(unit_id);
-            const unit = this.state.unit_by_path[unit_path];
+            const unit = { ...this.state.unit_by_path[unit_path] };
```
jelly, line 761: This seems unrequired, the previous line should do the same no?

jelly (changes requested):

pkg/systemd/services/services.jsx
tomasmatus: Ok, passing it as a prop.

pkg/systemd/services/services.jsx
```diff
@@ -811,6 +832,7 @@ class ServicesPageBody extends React.Component {
                         !filters.activeState.includes(this.activeState[unit.ActiveState]))
                         return false;
 
+                    unit.IsPinned = this.state.pinnedUnits.includes(unit.path);
```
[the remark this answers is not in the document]
tomasmatus: This is still required because it goes to `<ServicesList>`

jelly:
The packit failure is unrelated to your pull request, something is up with copr's infra.

jelly:
Triggered the failing tests again, but not sure if they are in fact flakes?

jelly (approved):

garrett (approved):
To be clear, I still think this is great! :+1:

Looking forward to it! Thanks again!

PR #17407, opened 2022-05-31
containers/ws: Support running as unprivileged user

martinpitt:
This provides the basic functionality of the ./containers/bastion demo
container. You can already customize cockpit.conf and the branding, but
it does not support SSH key logins yet. This will be added in a
follow-up.

## Unprivileged cockpit/ws container mode

The [cockpit/ws container](https://quay.io/repository/cockpit/ws) can now also be run in unprivileged mode. It then presents an unbranded login page. The user always has to specify a host name, and it connects to that host with SSH.

This mode is suitable for deploying to e.g. Kubernetes or similar environments where you cannot have or want privileged containers. In this "bastion host mode", you can get a Cockpit for servers in your data center without opening an extra port for cockpit-ws on them.

For now this only supports password authentication. Support for SSH key authentication will be added soon.

### DIFF
--- a/test/verify/check-ws-bastion
+++ b/test/verify/check-ws-bastion
@@ -1,0 +1,133 @@
+#!/usr/bin/python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2022 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import unittest
+
+import parent
+from testlib import *
+from testvm import DEFAULT_IMAGE
+
+HOST = "host.containers.internal"
+
+
+@unittest.skipUnless(DEFAULT_IMAGE == "fedora-coreos", "no cockpit/ws container on this image")
+@nondestructive
+class TestWsBastionContainer(MachineCase):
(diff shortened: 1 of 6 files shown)


martinpitt:
I can't for the life of me reproduce [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17407-20220607-171246-4708f543-fedora-coreos/log.html#310), not even with `run-tests` and the precise list of tests. Increasing the timeout didn't work. Added some further debugging.

martinpitt:
@marusak , @allisonkarlitskaya: Friendly review ping?

test/verify/check-ws-bastion
```diff
@@ -0,0 +1,133 @@
...
+        b.click("#login-button")
+        self.approve_key(b, HOST)
+
+        b.wait_visible('#content')
+        b.wait_text('#current-username', 'admin')
+        b.logout()
+
+        # remembers the last host via URL, server field should be pre-filled
+        self.assertEqual(b.eval_js("window.location.pathname"), f"/={HOST}/system")
+        # FIXME: login page does not really set this in the DOM? DOM has empty value, but browser shows the value
```
marusak, line 83: What? It is not in dom, but it is visible? How does that work?

test/verify/check-ws-bastion
```diff
@@ -0,0 +1,133 @@
...
+        b.wait_visible('#content')
+        b.logout()
+
+    def testCustomConf(self):
+        m = self.machine
+        b = self.browser
+
+        # custom cockpit.conf and pretend we are Fedora CoreOS
+        self.write_file("/root/cockpit.conf", """[WebService]
+            LoginTitle = My Walden
```
marusak, line 100: Is that reference to nightwish? :)

marusak (approved):
Looks good, thanks!

test/verify/check-ws-bastion
martinpitt: I have no idea.. I stared at this for 30 mins, then gave up (as our bastion test doesn't cover this either), and thought let's not block on that..

test/verify/check-ws-bastion
martinpitt: It's a [famous book](https://en.wikipedia.org/wiki/Walden), but of course I couldn't resist :grin: :musical_note:

PR #17414, opened 2022-06-01
build: change approach to versioning

allisonkarlitskaya:
Currently, we choose the version number that we use at the time that ./autogen.sh is first run.  That means that if you configure your tree and hack on it for a while, making commits as you go, the overly-specific version number will soon be out of date.  This version number is the number used when building tarballs for testing.  Even switching to old branches on previous versions can result in the version number not being updated.

Let's move to automake's suggested approach[1] to dynamic versioning by introducing a `version.m4` file.  We write that file from two places:

- `./autogen.sh` writes an approximate version number (the last release) appended by +git (eg. "270+git")

- from `make dist` we write the value we find in `$(VERSION)` and call autoreconf to update the configure script accordingly

The expectation is now that anyone who calls `make dist` should specify the VERSION explicitly.  That's two places:

- from `tools/make-dist` we specify the full `git describe` style version

- from the release workflow we specify the tag name

Since the only thing that uses `make dump-dist` anymore is `tools/make-dist`, and since it now provides its own version number, we can drop the rule and print the filename directly.

Since we move to `version.m4`-style versioning, our tarballs will no longer contain the `.tarball` file.

Finally, since we're going to end up having more tarballs created, with various version numbers, add `cockpit-*.tar.xz` to `CLEANFILES` to help keep things under control.

[1] https://www.gnu.org/software/automake/manual/html_node/Rebuilding.html

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
         run: ./autogen.sh
 
       - name: Build release
-        run: make dist -j$(nproc)
+        run: make dist -j$(nproc) VERSION='${{ github.ref_name }}'
 
       - id: publish
         name: Publish GitHub release
--- a/.gitignore
+++ b/.gitignore
@@ -14,14 +14,15 @@
 *.pyc
 *.so
 *.trs
-.dirstamp
 .deps/
+.dirstamp
 .extra_dist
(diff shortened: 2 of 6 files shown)


martinpitt (approved):
Danke!

PR #17417, opened 2022-06-01
Make Try Again button clickable

jelly:


pkg/static/login.css
```diff
@@ -86,6 +86,10 @@ button {
   opacity: .7;
 }
 
+.try-again {
```
marusak, line 89: This can be done by giving the element `href="#"`. Or ideally using `<button>` and styling it as link.

pkg/static/login.js
```diff
@@ -888,7 +888,9 @@
                     }
                 }
             } else if (xhr.status == 403) {
-                login_failure(_(decodeURIComponent(xhr.statusText)) || _("Permission denied"));
+                host_failure(_(decodeURIComponent(xhr.statusText)) || _("Permission denied"));
```
marusak, line 891: commit message does not explain why you moved this from `login_failure` to `host_failure`.

pkg/static/login.css
jelly: I opted for using `href="#"` styling as a button would introduce more custom css.

pkg/static/login.js
jelly: Yes hmm and neither do I... switched it back.

pkg/static/login.js
```diff
@@ -889,6 +889,8 @@
                 }
             } else if (xhr.status == 403) {
                 login_failure(_(decodeURIComponent(xhr.statusText)) || _("Permission denied"));
+            } else if (xhr.status == 500) {
```
marusak, line 892: in all other cases we `decodeURIComponent()`. Don't we need it here as well?

jelly: So this change broke a unit test and just fixes one issue out of multiple issues with our login page in cockpit client. I've dropped this change as I think we need a better approach here.

marusak: why have you dropped this commit?

marusak (dismissed):

marusak (approved):

PR #17418, opened 2022-06-01
test: Code coverage for the PR diff

mvollmer:
https://cockpit-logs.us-east-1.linodeobjects.com/pull-17418-20220602-161656-1dd16e0c-fedora-36-devel/Coverage/work/bots/make-checkout-workdir/github-pr.diff.gcov.html

### DIFF
--- a/test/common/lcov.py
+++ b/test/common/lcov.py
@@ -217,6 +217,56 @@ def print_file_coverage(path, line_hits, base_dir, out):
     out.write("end_of_record\n")
 
 
+class DiffMap:
+    # Parse a unified diff and make a index for the added lines
+    def __init__(self, diff):
+        self.map = {}
+        plus_name = None
+        diff_line = 0
+        with open(diff) as f:
+            for line in f.readlines():
+                diff_line += 1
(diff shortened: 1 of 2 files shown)


mvollmer:
Diff coverage for the diff is here, for example: https://cockpit-logs.us-east-1.linodeobjects.com/pull-17418-20220601-172147-f32c2ad9-fedora-36-devel/Coverage/work/bots/make-checkout-workdir/github-pr.diff.gcov.html

martinpitt:
This is nice! An important step for providing this as direct feedback on PRs. This is harmless to land without too much logic review, but not with the "DBG" commit. Moving to draft for now, please move back to "ready" and ask for review once this gets dropped.

marusak:
Looks awesome! Now it is rather hidden but I know that we want to promote this in annotations or so. Agree with pitti that it is landable as soon as you drop the DBG commit.

martinpitt:
As @mvollmer is on PTO, I dropped the DBG commit.

martinpitt (approved):

PR #17419, opened 2022-06-02
test: Use --track-naughties also for container scenarios

mvollmer:
Now that they use run-tests, they should also use --track-naughties.

### DIFF
--- a/test/run
+++ b/test/run
@@ -26,7 +26,7 @@ case $TEST_SCENARIO in
             exit 1
         fi
         test/image-prepare --quick --containers --verbose $TEST_OS
-        test/common/run-tests --test-dir test/containers --test-glob check-"$container"'*'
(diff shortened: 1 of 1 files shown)


test/run
```diff
@@ -26,7 +26,7 @@ case $TEST_SCENARIO in
             exit 1
         fi
         test/image-prepare --quick --containers --verbose $TEST_OS
-        test/common/run-tests --test-dir test/containers --test-glob check-"$container"'*'
+        test/common/run-tests--test-dir test/containers --test-glob check-"$container"'*' --track-naughties
```
martinpitt, line 29: You dropped the first space accidentally, that's not going to work.

martinpitt (changes requested):

test/run
mvollmer: Ouch, hacking too fast...

martinpitt (approved):

PR #17420, opened 2022-06-02
test: verify state before pushing pixel-tests

jelly:
Before pushing new pixel tests verify if the current pixel test
reference is not out of date. This is a bit tricky as we do not know how
a user has configured his git remotes to look up the latest test
reference. Therefore we fetch the latest main by hardcoding the
github.com and looking the project name from the configured git remotes.

### DIFF
--- a/test/common/pixel-tests
+++ b/test/common/pixel-tests
@@ -14,6 +14,11 @@ message() {
     [ "${V-}" != 0 ] || printf "  %-8s %s\n" "$1" "$2"
 }
 
+get_pixel_reference() {
+    local ref=$1
+    git ls-tree "${ref}" "${TEST_REFERENCE_SUBDIR}"
+}
+
 cmd_init() {
     git submodule add -b empty "$CLONE_REMOTE" "$TEST_REFERENCE_SUBDIR"
 }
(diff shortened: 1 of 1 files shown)


test/common/pixel-tests
```diff
@@ -33,6 +38,20 @@ cmd_status() {
 }
 
 cmd_push() {
+    # Check if the current branch is behind the project's main and has newer pixel-tests, in this case
+    # the user has to rebase to not run into pixel-tests conflicts. There is no
+    # elegant way figuring out the real "origin" of the cockpit project so we
+    # discover the project's repository name and hardcode the upstream url.
+    project=$(git remote -v | awk '/^origin/ { print $2; exit }' | xargs basename)
```
mvollmer, line 45: Does `git remote get-url origin` work? That would be less magic.

We have some code [here](https://github.com/cockpit-project/bots/blob/main/task/github.py#L170)
 to figure out the repo. We should probably use it in this PR and also in #17422.

test/common/pixel-tests
```diff
@@ -33,6 +38,20 @@ cmd_status() {
...
+    git fetch https://github.com/cockpit-project/${project} main
+
+    cur_sha=$(get_pixel_reference "FETCH_HEAD")
+    main_sha=$(get_pixel_reference "HEAD")
```
mvollmer, line 49: The names are mixed up, I think. `FETCH_HEAD` is the main branch, and `HEAD` is the current branch.

test/common/pixel-tests
```diff
@@ -33,6 +38,20 @@ cmd_status() {
...
+    if [[ "$cur_sha" != "$main_sha" ]]; then
```
mvollmer, line 50: This is too restrictive, I think. This also prevents running `pixel-tests push` a second time in the same PR.

We need to compare the SHA on main with the one on the common ancestor of main and the current branch.  If they are different, then changing test/reference on the current branch will lead to a conflict.

mvollmer: `git merge-base` can find the common ancestor.

mvollmer (changes requested):

test/common/pixel-tests
jelly: Yes, indeed

martinpitt:
@jelly I forgot the context here -- do we still need this?

jelly:
> @jelly I forgot the context here -- do we still need this?

I haven't ran into this lately, but basically we wanted to warn if your checkout is out of date and your old branch has an old pixel tests reference while main has a newer one.

subhoghoshX:
I've run into this couple of times, like recently I messed up [here](https://github.com/cockpit-project/cockpit/commit/f9e711f3f3cdc75a626b3ed45e1fbb17d8e178fe#diff-4d2a8b82625276fd3a0d33119371675547e384f46cd46eeee47a3f444394efab) and had to go through the trouble of fixing the conflict and pushing another new pixel commit. If I understand it correctly, it does `test/common/pixel-tests pull` or reminds me of it, before pushing new pixels. So I guess it's good to have :)

martinpitt:
This is stale, broken, and apparently it hasn't hurt us much in the last years that it is a pressing problem. If you ever feel like working on this again, please just reopen (GitHub doesn't forget anything). Thanks

PR #17421, opened 2022-06-02
pixels: Cause some pixel failures

mvollmer:
For testing "pixel-tests update".


mvollmer:
Doesn't need to be open.

PR #17423, opened 2022-06-03
systemd: Use WithDialogs protocol for mitigations dialog

mvollmer:


martinpitt (approved):
Thanks, this looks nice! Still draft, but seems to work alright. If you want to land this as-is, please go ahead.

PR #17424, opened 2022-06-03
firewall: Use new firewalld methods

marusak:
Right now checking if it is available everywhere. Then I would migrate few more calls that have `...2` version

### DIFF
--- a/pkg/lib/cockpit-components-firewalld-request.jsx
+++ b/pkg/lib/cockpit-components-firewalld-request.jsx
@@ -63,10 +63,9 @@ export const FirewalldRequest = ({ service, title, pageSection }) => {
         firewalld.call("/org/fedoraproject/FirewallD1", "org.fedoraproject.FirewallD1.zone", "getActiveZones")
                 .then(([info]) => {
                     const names = Object.keys(info);
-                    // TODO: superseded by org.fedoraproject.FirewallD1.zone.getZoneSettings2, but not available on Ubuntu 20.04 yet
-                    Promise.all(names.map(name => firewalld.call("/org/fedoraproject/FirewallD1", "org.fedoraproject.FirewallD1", "getZoneSettings", [name])))
+                    Promise.all(names.map(name => firewalld.call("/org/fedoraproject/FirewallD1", "org.fedoraproject.FirewallD1.zone", "getZoneSettings2", [name])))
                             .then(zoneInfos => {
-                                setEnabledAnywhere(!!zoneInfos.find(zoneInfo => zoneInfo[0][5].indexOf(service) >= 0));
+                                setEnabledAnywhere(!!zoneInfos.find(zoneInfo => ((zoneInfo[0].services || {}).v || []).indexOf(service) >= 0));
                                 setZones(names);
                             })
                             .catch(ex => {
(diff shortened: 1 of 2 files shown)


martinpitt (dismissed):
That does look a little nicer indeed!

marusak:
Converted rest of the cases

pkg/lib/cockpit-components-firewalld-request.jsx
```diff
@@ -63,10 +63,10 @@ export const FirewalldRequest = ({ service, title, pageSection }) => {
         firewalld.call("/org/fedoraproject/FirewallD1", "org.fedoraproject.FirewallD1.zone", "getActiveZones")
                 .then(([info]) => {
                     const names = Object.keys(info);
-                    // TODO: superseded by org.fedoraproject.FirewallD1.zone.getZoneSettings2, but not available on Ubuntu 20.04 yet
-                    Promise.all(names.map(name => firewalld.call("/org/fedoraproject/FirewallD1", "org.fedoraproject.FirewallD1", "getZoneSettings", [name])))
+                    Promise.all(names.map(name => firewalld.call("/org/fedoraproject/FirewallD1", "org.fedoraproject.FirewallD1.zone", "getZoneSettings2", [name])))
                             .then(zoneInfos => {
-                                setEnabledAnywhere(!!zoneInfos.find(zoneInfo => zoneInfo[0][5].indexOf(service) >= 0));
+                                console.log(zoneInfos);
```
martinpitt, line 68: okayokay we have a web console, but let's not go overboard with that. :grin:

pkg/networkmanager/firewall-client.js
```diff
@@ -309,19 +309,19 @@ function fetchZoneInfos(zones) {
             return info;
         }
         return firewalld_dbus.call('/org/fedoraproject/FirewallD1',
-                                   'org.fedoraproject.FirewallD1',
-                                   'getZoneSettings', [zone])
+                                   'org.fedoraproject.FirewallD1.zone',
+                                   'getZoneSettings2', [zone])
                 .then(reply => {
-                    const [, name, description, , target, services, ports, , , , interfaces, source] = reply[0];
+                    const zoneInfo = reply[0];
```
martinpitt, line 315: While at it, would you mind cleaning this up to `.then(([zoneInfo]) => ...`? (I.e.structured decomposition). In principle you can do this even one step further with `([{ description, target, services, ... }])`, but it doesn't help that much due to the required `.v` and fallbacks.

martinpitt (changes requested):
Nice!

martinpitt:
Uh-oh! [flatpak failure](https://github.com/cockpit-project/cockpit/runs/6730082189?check_suite_focus=true):
```
error: Signature made Thu Jan  1 00:00:00 1970 using [unknown name] key ID 562702E9E3ED7EE8
BAD signature from "Flathub Repo Signing Key <flathub@flathub.org>"
Primary key ID 4184DD4D907A7CAE
Primary key expires Mon Jun 14 13:18:56 2027
```

Wut? Retrying.

martinpitt:
Rawhide got Firefox 101, which breaks CDP. That affects all projects, so will become urgent rather soon. But let's not block existing PRs on that.

PR #17425, opened 2022-06-03
systemd: Fix link in BreadcrumbItem

tomasmatus:
Fixes issue on service details page where 'Services' link doesn't lead back
to the services overview page when service Id containts special characters.

### DIFF
--- a/pkg/systemd/services/service.jsx
+++ b/pkg/systemd/services/service.jsx
@@ -79,7 +79,7 @@ export class Service extends React.Component {
                       id="service-details"
                       breadcrumb={
                           <Breadcrumb>
-                              <BreadcrumbItem to={"#" + cockpit.location.href.replace(cockpit.location.path[0], '')}>{_("Services")}</BreadcrumbItem>
+                              <BreadcrumbItem to={"#" + cockpit.location.href.replace(/\/[^?]*/, '')}>{_("Services")}</BreadcrumbItem>
                               <BreadcrumbItem isActive>
                                   {this.props.unit.Id}
                               </BreadcrumbItem>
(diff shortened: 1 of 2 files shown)


pkg/systemd/services/service.jsx
```diff
@@ -79,7 +79,7 @@ export class Service extends React.Component {
                       id="service-details"
                       breadcrumb={
                           <Breadcrumb>
-                              <BreadcrumbItem to={"#" + cockpit.location.href.replace(cockpit.location.path[0], '')}>{_("Services")}</BreadcrumbItem>
+                              <BreadcrumbItem to={"#/" + cockpit.location.href.replace(cockpit.location.encode(cockpit.location.path[0]), '')}>{_("Services")}</BreadcrumbItem>
```
martinpitt, line 82: I feel like this is making an unnecessarily strong assumption about how the page encodes the URL. IMHO it would be simpler and more robust to just replace the entire `#.*(?|$)` with `#\1`, i. e. dropping the entire fragment for system services (no `?` in the URL), or dropping the unit name (keeping the `?`).

martinpitt (changes requested):
Thanks for fixing this! Indeed this is annoying. Do we have a test case that checks this breadcrumb for a user unit? Could we rename that unit to contain a "funny" character to reproduce this? (Happy to help with this, of course!)

tomasmatus:
> Thanks for fixing this! Indeed this is annoying. Do we have a test case that checks this breadcrumb for a user unit? Could we rename that unit to contain a "funny" character to reproduce this? (Happy to help with this, of course!)

I added new unit with "funny" name and a test to check if "Service" link works correctly.

martinpitt (approved):
Very nice, thank you! :100:

PR #17426, opened 2022-06-03
tools: Adjust RHEL dependencies

martinpitt:
Avoid some "Recommends", as they are discouraged:

- Lift subscription-manager-cockpit to Requires, it's an essential part
  on RHEL.
- Require sudo, that's our main UI for getting administrative
  privileges.
- Demote NetworkManager-team to Suggests.

Shuffle the Requires:, Provides:, and Suggests: lines a bit to keep the
same kind together in a single block.

https://bugzilla.redhat.com/show_bug.cgi?id=2093231

---

This also fixes an [unexpected error message](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17426-20220603-183752-19c4b4ea-fedora-35/log.html#304), which seems really unlucky, but let's just deal with it.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -607,8 +607,8 @@ class TestConnection(MachineCase):
             # clean up debug symbols, they get in the way of upgrading
             m.execute("dpkg --purge cockpit-ws-dbgsym")
         elif m.image.startswith("rhel"):
-            # subscription-manager-cockpit depends on cockpit-ws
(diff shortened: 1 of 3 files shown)


ptoscano:
subscription-manager developer here -- the bump of `subscription-manager-cockpit` to a Requires seems OK for me.

marusak (approved):
danke :)

PR #17429, opened 2022-06-04
test: Wait for CDP targets during browser/CDP startup

martinpitt:
With Firefox 101 the CDP target initialization takes a tad longer, so
that almost all tests fail with "Error: No inspectable targets".

Change the waiting loop to not just check that we can connect to the CDP
port, but that CDP has at least one debug target.

----

This fixes the Firefox test failures in packit on [F36](https://artifacts.dev.testing-farm.io/8e8179a4-068e-4137-9907-ea0650d6d5bc/) and rawhide, as 101 recently landed. This will soon affect Fedora 35 (i.e. our tasks container) as well.

### DIFF
--- a/test/common/cdp.py
+++ b/test/common/cdp.py
@@ -6,12 +6,13 @@ import os
 import random
 import resource
 import shutil
-import socket
 import subprocess
 import sys
 import tempfile
 import time
 import typing
+import urllib.request
+from urllib.error import URLError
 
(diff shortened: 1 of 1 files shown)


marusak (approved):
Nice, thanks!

martinpitt:
We now need to pull this into all of our projects. Unfortunately c-machines [needs some non-trivial work](https://github.com/cockpit-project/cockpit-machines/pull/725).

PR #17431, opened 2022-06-06
firewall: Prefer zone name when given

marusak:
Zone name may be defined so we should use that when possible.

Before:
Screenshot from 2022-06-06 14-48-06

After:
Screenshot from 2022-06-06 14-52-08

### DIFF
--- a/pkg/networkmanager/firewall.jsx
+++ b/pkg/networkmanager/firewall.jsx
@@ -175,7 +175,7 @@ function ZoneSection(props) {
             <CardTitle>
                 <Flex alignItems={{ default: 'alignSelfBaseline' }} spaceItems={{ default: 'spaceItemsXl' }}>
                     <Title headingLevel="h2" size="xl">
-                        { cockpit.format(_("$0 Zone"), upperCaseFirstLetter(props.zone.id)) }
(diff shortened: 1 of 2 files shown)


pkg/networkmanager/firewall.jsx
```diff
@@ -175,7 +175,7 @@ function ZoneSection(props) {
             <CardTitle>
                 <Flex alignItems={{ default: 'alignSelfBaseline' }} spaceItems={{ default: 'spaceItemsXl' }}>
                     <Title headingLevel="h2" size="xl">
-                        { cockpit.format(_("$0 Zone"), upperCaseFirstLetter(props.zone.id)) }
+                        { cockpit.format(_("$0 zone"), props.zone.name ? props.zone.name : upperCaseFirstLetter(props.zone.id)) }
```
martinpitt, line 178: `props.zone.name || upperCaseFirstLetter(props.zone.id)` is the more useful notation for this. Unfortunately JS does not understand `val ?: fallback`, like C.

martinpitt (changes requested):
Nice fix! Tests need adjustment, though.

martinpitt (approved):
Dakujem!

PR #17434, opened 2022-06-07
storaged: Set cockpit.format_bytes() precision instead of string hacking

martinpitt:
That string hacking was only working in locales with `.` as decimal
separator, but e.g. not German or Portugese.

Use the new precision option from commit 083f889c0a47 (released in
Cockpit 266). Adjust the VDO tests accordingly.


marusak (approved):
Thanks, much better than checking if it starts with `0.` :)

Issue #16359, opened 2021-09-16
Identity customization within Cockpit itself (of colors, mainly)


garrett:
### Discussed in https://github.com/cockpit-project/cockpit/discussions/16256

<div type='discussions-op-text'>

<sup>Originally posted by **gabrielefronze** August 24, 2021</sup>
I am on the way to customise it in order to (graphically) integrate it with an ecosystem I am working on and I would like to get some tips on how to override the default UI colours (mainly the blue/dark blue ones).
</div>

---

**Goal**: Recolor Cockpit to fit in when embedded with other applications which might not use our particular shade of blue.

**Related**: Distribution branding, except identity customization actually affects the look of Cockpit when logged in as well.

_Somewhat_ related: Dark mode (#10189), as it would redefine colors as well. However, it's not necessary to implement basic identity branding tweaks. It would be bad if we had switchable dark/light modes with color overrides intended for light mode only, so custom colors should be wrapped in a `@media (prefers-color-scheme: light) { }` from the beginning (and extended support would include a separate block for `prefers-color-scheme: dark` in the future, once dark mode exists).

PatternFly has a mention of customization, but then links to the CSS variables at https://www.patternfly.org/v4/developer-resources/global-css-variables. We'd need a place to expose modifications to these variables with a linked in stylesheet that's loaded after the variables are defined.

@gabrielefronze made an attempt by modifying `page.scss` @ https://github.com/cockpit-project/cockpit/discussions/16256#discussioncomment-1323821, but the variables probably need to come later (as to not need all the `!important`s, which may cause problems if some of them are redefined locally in elements).

---

I think we'd need to have:
- a file that's loaded after everything else
  - with some example customizations which are commented out
  - examples should be within a `@media (prefers-color-scheme: light) { }` for future-proofing
  - a link to the PF variable customization page
- documentation (possibly by expanding the branding documentation a little)
  - to let people know that this exists
  - very light basics on how to use it

garrett:
TODO:

1. Implement dark mode, as anything that's specified in identity customization would _most likely_ clash with dark mode support.
2. Specify which parts of Cockpit would like be themed and at what priority.
3. Come up with API via CSS Custom Properties ("CSS variables"), possibly with some "toggles" too.

Items 2 and 3 on the above list can be defined in parallel with the implementation of dark mode. Item # 2 can be suggested by anyone (so provide feedback here, please). Item # 3 would probably be drafted based on item # 2.

[...]

PR #17437, opened 2022-06-07
RFC: add theming support (via CSS overrides)

dgdavid:
This PoC is just a proposal to help with https://github.com/cockpit-project/cockpit/issues/16359. There might be a lot of details that need to be considered, which is why I am opening the PR as a RFC. Regardless, it's the cleanest method I've found after trying [ others](https://gist.github.com/dgdavid/ec2648c61f63f49a8aa8d9871d2b70f9). Read more at https://github.com/cockpit-project/cockpit/discussions/16256#discussioncomment-2834640

Basically, _the trick_ is to load a new _theme-overrides.css_ file in every HTML page. For overrides to take effect, this file must be the last to be applied.

Since the load of this file mimic how [_branding.css_ is included](https://github.com/cockpit-project/cockpit/blob/75c33b347ae80115241f058c196ec25ffbc9d274/pkg/shell/index.html#L8), Cockpit look&feel could be customized without having to patch it (once changes are merged, of course). Just using an external package to provide needed resources. That package must include those files following the same rules as [branding](https://github.com/cockpit-project/cockpit/blob/27cb665b5c135481f900dafac0c1b754ab91b5a0/doc/branding.md#how-cockpit-selects-branding).

For illustrative purposes, I have [created a simple package](https://build.opensuse.org/package/show/home:dgdavid/carbon-cockpit-theme) using a few bits (colors, spacing, typeface) from [Carbon Design System](https://carbondesignsystem.com) and patched Cockpit before using it. See how it looks:

|| Before | After |
|-|-|-|
| Overview | overview_before_theming | overview_after_theming |
| Logs | logs_before_theming | logs_after_theming |
| Storage | storage_before_theming | storage_after_theming |
| Network | network_before_theming | network_after_theming |

:point_right: Note: the package uses the _/usr/share/cockpit/branding/default_ directory just for the sake of simplicity.

## Caveats

With this approach, you can never be sure that every Cockpit module, like [cockpit-podman](https://github.com/cockpit-project/cockpit-podman) or [cockpit-wicked](https://github.com/openSUSE/cockpit-wicked), will be themed as you wish. They can be adapted to load the file as well, and even patching the [starter-kit](https://github.com/cockpit-project/starter-kit) can help with upcoming modules, but would still be subject to oversights.

So having the option of always _injecting_ or _applying_ the overrides after each iframe has been loaded seems like a better idea. Yet I haven't begun looking into this (I would appreciate any guidance).

## Pending stuff

_Only if this approach will be adopted_

* to choose a better name for that file
* to think about the convenience of adding an empty one by default (or with an explanatory comment pointed to the documentation).
* to check if worth reusing _branding.css_ instead (?) (maybe there could some _CSS collisions_ (??))

---

CC: @gabrielefronze

### DIFF
--- a/pkg/apps/index.html
+++ b/pkg/apps/index.html
@@ -23,6 +23,7 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <title translate="yes">Applications</title>
     <meta charset="utf-8">
     <link href="apps.css" type="text/css" rel="stylesheet">
+    <link href="../../static/theme-overrides.css" type="text/css" rel="stylesheet">
     <script type="text/javascript" src="../base1/cockpit.js"></script>
     <script type="text/javascript" src="../base1/po.js"></script>
     <script type="text/javascript" src="po.js"></script>
--- a/pkg/kdump/index.html
+++ b/pkg/kdump/index.html
@@ -25,6 +25,7 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <link rel="stylesheet" href="kdump.css">
+    <link href="../../static/theme-overrides.css" type="text/css" rel="stylesheet">
 
     <script type="text/javascript" src="kdump.js"></script>
     <script type="text/javascript" src="../base/po.js"></script>
--- a/pkg/metrics/index.html
+++ b/pkg/metrics/index.html
@@ -23,6 +23,7 @@ along with this package; If not, see <http://www.gnu.org/licenses/>.
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <link rel="stylesheet" href="index.css">
+    <link href="../../static/theme-overrides.css" type="text/css" rel="stylesheet">
 
     <script type="text/javascript" src="../base1/cockpit.js"></script>
     <script type="text/javascript" src="../manifests.js"></script>
--- a/pkg/networkmanager/firewall.html
+++ b/pkg/networkmanager/firewall.html
@@ -23,6 +23,7 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <meta charset="utf-8">
 
     <link href="firewall.css" type="text/css" rel="stylesheet">
+    <link href="../../static/theme-overrides.css" type="text/css" rel="stylesheet">
 
     <script src="../base1/cockpit.js"></script>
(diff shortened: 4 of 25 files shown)


pkg/apps/index.html
```diff
@@ -23,6 +23,7 @@
     <title translate="yes">Applications</title>
     <meta charset="utf-8">
     <link href="apps.css" type="text/css" rel="stylesheet">
+    <link href="../../static/theme-overrides.css" type="text/css" rel="stylesheet">
```
dgdavid, line 26: While writing the PR's description I just realized that using

```suggestion
    <link href="cockpit/static/theme-overrides.css" type="text/css" rel="stylesheet">
```
might work too. Not sure, just speculating because of https://github.com/cockpit-project/cockpit/blob/388c2f423a6d844ca2d1d9a55ded7086a4245ff1/pkg/static/login.html#L18

garrett:
There are some issues with the basic concept:

1. We should dynamically check if the file exists. 
   - If we don't, then it'll be a 404 with anyone who does not have customizations. 
   - If we ship a blank file (or one with comments) that people could customize, then customizations would be overridden on each Cockpit update.

2. We're working on a dark mode for Cockpit.
   - When dark mode lands, it will 100% conflict with almost all customization someone would do with arbitrary custom CSS.

3. Updates to Cockpit (and PatternFly) in general will eventually conflict with custom CSS, causing Cockpit's UI to eventually break for anyone who does customizations.
   - CSS is not an API and our CSS is not stable, as it's subject to change not just at the level of Cockpit, but also (more critically) at the level of PatternFly.

garrett:
We _could_ define some custom variables for CSS (aka: "CSS custom properties") and use those variables with fallbacks.

It would be for a custom color. However, then there's the issue with contrast. If the color is dark, then the text needs to be light, and vice versa. There's no way to detect that properly in CSS.

(If we did this, then we'd still need to dynamically detect if a custom file exists and load it if so.)

garrett:
There's another problem, as well: Cockpit add-ons, like Cockpit-Podman, Cockpit-Machines, Cockpit-ostree, and all the 3rd party Cockpit add-ons like [everything 45Drives develops](https://github.com/45Drives?q=cockpit).

dgdavid:
Thanks @garrett for the feedback :+1:  Let's close the PR and follow the conversation in https://github.com/cockpit-project/cockpit/issues/16359.

PR #17445, opened 2022-06-08
static: drop Internet Explorer workaround for baseUri

jelly:
Since Internet Explorer is no longer support and Microsoft Edge does
support document.baseUri this can be dropped.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -235,15 +235,7 @@
     /* Sets values for application, url_root and login_path */
     function setup_path_globals (path) {
         const parser = document.createElement('a');
-        let base = document.baseURI;
-        /* Some IEs don't support baseURI */
-        if (!base) {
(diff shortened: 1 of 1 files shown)


garrett (changes requested):
No, the login page is special.

We still need to support Internet Explorer to the extent to say that it is not supported.

Removing this code will break the message of not being supported on IE.

garrett:
Believe it or not, people still use IE. We need to tell them they're not supported instead of showing a broken page.

jelly:
Actually this does not break the page at all, it breaks the login functionality and I believe users should already see an message on the login page then that their "browser is not supported".

martinpitt:
@jelly : If you want to go ahead with this, can you please rebase? Note that my PR #17469 also deletes this code en passant; I just don't know how fast that will land.

martinpitt (approved):
Thanks!

PR #17446, opened 2022-06-13
test: Disable chromium SHM usage with small /dev/shm

martinpitt:
In containers, /dev/shm usually defaults to 64 MiB, which is too small
for Chromium. In our own CI environments we mount a bigger tmpfs, but
this is not always possible, e. g. on CirrOS CI.

Detect this, and tell chromium to not use shared memory in this case.

---

See https://github.com/cockpit-project/starter-kit/pull/585 for details and the crash this fixes.

### DIFF
--- a/test/common/cdp.py
+++ b/test/common/cdp.py
@@ -92,14 +92,22 @@ class Chromium(Browser):
         return None
 
     def cmd(self, cdp_port, env, show_browser, browser_home, download_dir):
-        exe = self.path(show_browser)
-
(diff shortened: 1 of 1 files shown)


martinpitt:
This initial approach [does not help](https://cirrus-ci.com/task/6231675553185792), i.e. the `--disable-dev-shm-usage` switch does not seem to do anything. Despite being documented: https://peter.sh/experiments/chromium-command-line-switches/

:man_shrugging:

PR #17449, opened 2022-06-13
base1: Compute url-root fallback in cockpit.js

### LINKED ISSUE #17444: cockpit.js does not respect URL root without login page
### Explain what happens

We want to run cockpit-ws in a cloud instance with external authentication for [cloud.redhat.com support](https://issues.redhat.com/browse/COCKPIT-795). I.e. we don't want to show a login page. As a demo:

```
$ cat cfg/cockpit/cockpit.conf 
[Webservice]
UrlRoot=/api/cockpit-9090/
Origins = https://localhost:8443

# in built cockpit tree
$ XDG_CONFIG_DIRS=cfg XDG_DATA_DIRS=$(pwd) ./cockpit-ws --port 9999 --local-session=./cockpit-bridge --for-tls-proxy
```

Plus running https://github.com/jelly/console.dot `make run` which sets up a TLS proxy from https://localhost:8443 to http://localhost:9999.

This works well enough with static resources, like

> GET https://localhost:8443/api/cockpit-9090/system
> GET | https://localhost:8443/api/cockpit-9090/cockpit/$c88cf5a57d8b77710ce5b436c63c45b4de2abed06f3b448c1af0256cb6eada06/manifests.js

But it mis-handles the websocket:

> GET wss://localhost:8443/cockpit/socket     418

i.e. it's missing the UrlRoot.

This is because cockpit.js looks at

    url_root = window.localStorage.getItem('url-root');

which is only set by login.js. But in this scenario we don't have a login page. We do test non-password logins in our tests, and we do test `UrlRoot=`, but we don't test both together (like for Kerberos or SSH login).

ATM I'm not *entirely* sure if we only skip login.js for the `--local-session=` trick/hack, or also for other noninteractive logins. My guess is the latter, but let's confirm.

Plan:
 - Add integration tests for UrlRoot= with a noninteractive login; both `--local-session` and kerberos preferably
 - Check whether this also skips login.js (I assume yes)
 - Fix it by adding the URL root detection to cockpit.js itself, instead of relying on login.js to do it. This might be a bit tricky as we must only do that for "primary" sessions, not remote ones from cockpit. I'm still sure we can tell by `window.location.path` only.

### Version of Cockpit

270

Navigation & Shell

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

martinpitt:
So far we have relied on pkg/static/login.js to compute the URL root
(`UrlRoot=` setting in cockpit-ws), and put it into localStorage as
`url-root`. However, these days we don't actually invoke the login page
in many scenarios: Kerberos/Smart card authentication, SSH auth with the 
bastion container, custom setups with Bearer token/OAuth, etc.

Provide a fallback for these situations: Look at the current URL path,
strip off all the known "special" paths (see doc/urls.md), and strip off 
leading and trailing slashes (like login.js does).

Keep the localStorage property, as older Cockpit versions (remote hosts)
rely on it. 

Fixes #17444

----

I am not really happy with this admittedly. It's a heuristic, and will fail in some cases like /myroot/system. But let's do a round of testing, while I think about alternatives.

### DIFF
--- a/pkg/lib/cockpit.js
+++ b/pkg/lib/cockpit.js
@@ -25,6 +25,14 @@ try {
     // Sometimes this throws a SecurityError such as during testing
     url_root = window.localStorage.getItem('url-root');
 } catch (e) { }
+if (url_root === null) {
+    // when not going through login page, derive it from URL; See doc/urls.md: /cockpit or /cockpit+,
+    // as well as /@host or /=host are special, and never part of the UrlRoot.
+    url_root = window.location.pathname
+        .replace(/\/cockpit[/+].*$/, '').replace(/\/[@=].*$/, '')
+        // url root must not start or end with slash
+        .replace(/^\/+|\/+$/g, '');
+}
 
 /* injected by tests */
 var mock = mock || { }; // eslint-disable-line no-use-before-define, no-var
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -1050,7 +1050,7 @@ ProtocolHeader = X-Forwarded-Proto
         self.allow_journal_messages("audit.*bool=httpd_can_network_connect.*val=1.*")
 
     def checkCockpitOnProxy(self, urlroot=""):
-        b = self.browser
+        b = self.new_browser()
 
         # should use nginx' certificate, not cockpit's; use --resolve so that SNI matches the certificate's CN
         (https_host, https_port) = self.machine.forward["443"].split(':')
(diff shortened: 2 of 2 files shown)


martinpitt:
Ah right, this approach is fundamentally flawed. Requesting an initial path of `/system/` would now consider that as an URL prefix. Pushing this detection to cockpit.js is the wrong place. However, cockpit-ws reading its UrlRoot= and exporting it via a bridge init also sounds wrong -- we have no idea how the proxy in between mangles the paths. So I suppose I drop this for now, and rather think about a shim on the ws container side.

PR #17450, opened 2022-06-13
Eliminate tools/create-spec, keep real .spec file in git, simplify packit

martinpitt:
See individual commits for details.

Fixes https://github.com/packit/packit-service/issues/1515

---

I tested this on my fork with an [extra debug commit](https://github.com/martinpitt/cockpit/commit/6ee32fcab3c7055b6020fa8c4939c86d47be8eb2). [release workflows](https://github.com/martinpitt/cockpit/runs/6881204681), [release](https://github.com/martinpitt/cockpit/releases/tag/271.1). [propose-downstream packit job](https://dashboard.packit.dev/results/propose-downstream/808) succeeded -- the propose-downstream one was where we had the [original problem](https://dashboard.packit.dev/results/propose-downstream/597). [dist-git PR](https://src.fedoraproject.org/rpms/cockpit/pull-request/68#) looks as expected (tests fail because I didn't change the fmf ref). [copr build](https://copr.fedorainfracloud.org/coprs/martinpitt/test-fixes/build/4533376/) is also :+1:


martinpitt:
WTH goes wrong [on RHEL 8](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17450-20220613-125102-ae858d18-rhel-8-7/log.html)..

martinpitt:
[rawhide run](https://artifacts.dev.testing-farm.io/fe379852-3215-4e49-b26e-596448a75b4c/) had a complete browser failure. Retrying.

martinpitt:
OK, new firefox in rawhide is busted, but that's unrelated to this PR.

allisonkarlitskaya (changes requested):
Summary from IRC discussion: no.

This is just trading one poison for another.

The main benefit of a project like packit is that it can define "The Way To Do It™" and we can just happily follow along.  We're holding too much to our old pre-packit ways of managing this.  Let's stop fighting the way that packit seems to want to opperate, kill `create-spec`, and commit `cockpit.spec` to git.

We can figure out a way to deal with this from `image-prepare`.  Perhaps `-Dversion=` flag to `rpmbuild` or so...

martinpitt:
I reported that problem to https://github.com/packit/packit/issues/1621 and added a workaround.

martinpitt:
I re-tested the release, it went fine. Description updated. Just [one test case](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17450-20220614-123200-927bf8a8-rhel-8-7/log.html#26-2) needs to be adjusted now.

Makefile.am
```diff
@@ -28,6 +28,7 @@ distdir: $(DISTFILES)
 		mv .extra_dist.tmp '$(srcdir)/.extra_dist'; fi
 	$(MAKE) $(AM_MAKEFLAGS) distdir-am EXTRA_FILES="$$(tr '\n' ' ' < $(srcdir)/.extra_dist) .extra_dist"
 	test ! -x $(srcdir)/src/client/add-metainfo-releases || $(srcdir)/src/client/add-metainfo-releases --version '$(VERSION)' '$(distdir)'/src/client/*.metainfo.xml
+	sed -i "/^Version:/ s/0/$(VERSION)/" "$(distdir)/tools/cockpit.spec"
```
allisonkarlitskaya, line 31: ugh.

martinpitt: Hmm, not sure what to respond.. :man_shrugging: maybe? :grin: This was more or less the status quo until commit b34acdfcb740. If we want this file to be a real spec, we have to replace "0", not a "more magic" placeholder.

allisonkarlitskaya: I guess I don't like the asymmetry here vis-a-vis the other packaging files, now.

Could we turn our other packaging assets back into non-`.in` files and put version 0 in them and `sed` them up in the same way?  That would bring us full circle more or less back to the place we started...

Let's maybe have a chat...

martinpitt: I added an extra commit which moves to the same approach for Debian and Arch, so it's again symmetric.

allisonkarlitskaya: :+1:

allisonkarlitskaya (approved):

PR #17451, opened 2022-06-13
hack metainfo releases a bit more

allisonkarlitskaya:
I wanted to get all of this stuff out of the process of building a tarball, but
 - managing this externally is (really) hard.  At some point you find yourself writing XSL and ask yourself "...what am I doing?".
 - ximion/appstream#354 will hopefully provide a proper fix for this eventually
 - the current approach is working well enough

Instead of trying to create some new thing which we'll just have to replace a little bit later anyway, let's just make the current stuff work.

This fixes `make distcheck`.


martinpitt (approved):
Thanks! This is not at all bad indeed.

Can we adjust containers/unit-tests/scenario/distcheck now to actually do `distcheck` again? ATM it just does a subset of that.

allisonkarlitskaya:
> Thanks! This is not at all bad indeed.
> 
> Can we adjust containers/unit-tests/scenario/distcheck now to actually do `distcheck` again? ATM it just does a subset of that.

Indeed, that's exactly the point.

See #17304.

PR #17452, opened 2022-06-14
kdump: show tooltip when crashkernel is not configured

jelly:
Disable the service switch and show a tooltip when the kernel's command
line lacks the crashkernel option. Without this change a user can try to
enable the service which will instantly fail without an indication that
crashkernel has to be configured upfront.

### DIFF
--- a/pkg/kdump/kdump.js
+++ b/pkg/kdump/kdump.js
@@ -63,11 +63,20 @@ const initStore = function(rootElement) {
             stateChanging: dataStore.stateChanging,
             reservedMemory: dataStore.kdumpMemory,
             kdumpStatus: dataStore.kdumpStatus,
+            kdumpCmdlineEnabled: dataStore.crashkernel || false,
             onSaveSettings: dataStore.saveSettings,
             onCrashKernel: dataStore.kdumpClient.crashKernel,
         }), rootElement);
     };
     dataStore.render = render;
+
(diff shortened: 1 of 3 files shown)


jelly:
rawhide failure is related to an update to firefox last night.

pkg/kdump/kdump.js
```diff
@@ -63,12 +63,18 @@ const initStore = function(rootElement) {
...
             onSaveSettings: dataStore.saveSettings,
             onCrashKernel: dataStore.kdumpClient.crashKernel,
         }), rootElement);
     };
     dataStore.render = render;
 
+    cockpit.file("/proc/cmdline").read()
+            .then(content => {
+                dataStore.crashkernel = content.indexOf('crashkernel=') !== -1;
+            });
```
martinpitt, line 76: Is it guaranteed to be enabled in /proc/cmdline always? I.e. there's no way to set that in the kernel config?

Please handle failures. read() returns `null` if the file does not exist (yes, strange API), so calling `null.indexOf()` will crash. Permission errors will cause a rejection as usual. I know it's unlikely, but in principle we can connect to LXC or nspawn containers or have an environment where an unprivileged user can't read the file.

martinpitt (changes requested):
Thanks, this is a nice idea! Let's just fix the error handling.

martinpitt (approved):
Dank je wel!

PR #17456, opened 2022-06-15
kdump: verify the nfs server and part

jelly:
Use showmount to verify if the nfs server is reachable and the nfs
directory is a valid mount point.


jelly:
Soooo `showmount` is not really a great tool for validation as it requires `mountd` to be open and as just opening the nfs port doesn't do that. So I think this shouldn't be used as validation.

PR #17457, opened 2022-06-15
containers: Add SSH key login support to ws, drop bastion

martinpitt:
Copy the custom auth script from containers/bastion. Without
`$COCKPIT_SSH_KEY_PATH` this transparently falls back to regular
cockpit-ssh, thus does not change behaviour.


martinpitt:
Argh, now the [coverage stuff](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17457-20220615-082125-3b1f5988-fedora-36-devel) goes completely nuts, and no @mvollmer here to fix it :cry: I'll look into this.

martinpitt:
That is even worse -- all tests failed due to the coverage code bug, but the test is shown as successful overall. So this PR is not blocked after all, but I wonder how many  other test runs this actually broke.

martinpitt:
Added fixes for that coverage bug.

marusak (approved):
Looks good! Maybe @allisonkarlitskaya would like to take a look as well

martinpitt:
I still have a follow-up to do (modern SSH key support), and modernize the Python code a bit (In this PR it's a no-change move). I'll ask Lis to review the follow-up, IMHO that makes more sense.

PR #17461, opened 2022-06-17
package.json: Update chrome-remote-interface

jelly:
This release contains a fix for nodejs to prefer ipv4 resolving which
is needed as Firefox the CDP runs only on ipv4 on Fedora rawhide on
packit.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -24,6 +24,9 @@ git clone --depth=1 https://github.com/cockpit-project/bots
 # support running from clean git tree
 if [ ! -d node_modules/chrome-remote-interface ]; then
     npm install chrome-remote-interface sizzle
+else
+    # make sure we use the correct node_modules
+    ./tools/node-modules checkout
 fi
 
 export TEST_OS="${ID}-${VERSION_ID/./-}"
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "argparse": "^2.0.1",
     "axe-core": "^3.5.2",
     "babel-loader": "^8.1.0",
-    "chrome-remote-interface": "^0.28.1",
+    "chrome-remote-interface": "^0.31.2",
     "compression-webpack-plugin": "^9.2.0",
     "copy-webpack-plugin": "^6.1.0",
     "css-loader": "^5.2.0",


jelly:
Ok, so the issue seems to be that packit's testing farm builds don't update the node_modules
```
+ git rev-parse --short HEAD
da4c1cee2
+ cd node_modules
+ git rev-parse --short HEAD
cdf431b45
```

```
-Subproject commit cdf431b45bde3f1fabe2a1dc9cc7684130a7e858
+Subproject commit 56d88f356c64144d205564be2b0d09eee2a28c32
```
```

**Comment by jelly** [comment]
Ok, so this is interesting, adding `./tools/node-module checkout` makes packit happy so for some reason the initial clone does not update node_modules?

Maybe packit clones main and then checkouts out the PR branch and doesn't pull node_modules.

**Comment by jelly** [comment]
Seems packit indeed clones my main branch and then checks the target:

```
git clone https://github.com/jelly/cockpit testcode
git -C testcode checkout -b testbranch 5e0a374516ed40b0ae3b3bc99eb1b1ddcd3c376c
```

**Review by marusak — APPROVED** [review]
Still as draft, but looks good, thanks!

File: test/browser/run-test.sh [hunk 1]
```diff
@@ -24,6 +24,9 @@ git clone --depth=1 https://github.com/cockpit-project/bots
 # support running from clean git tree
 if [ ! -d node_modules/chrome-remote-interface ]; then
     npm install chrome-remote-interface sizzle
+else
+    # make sure we use the correct node_modules
+    ./tools/node-modules checkout
```
martinpitt, line 29: I don't like this if/else hack much, but let's fix the Firefox breakage ASAP -- we can clean this up in a follow-up. Thank you!

martinpitt (approved):
Thanks for figuring this out!

test/browser/run-test.sh
jelly: Ok, I think it's worse if it was always pulling node modules right? How do you propose we make it better, just always pull node modules?

martinpitt: I meant that it shuold be like this:
```
[ -d node_modules/chrome-remote-interface ] ||  ./tools/node-modules checkout
```

PR #17468, opened 2022-06-20
shell: Fix rendering of access level indicator

### LINKED ISSUE #17467: shell: Access level indicator does not always update when switching hosts
See "Example 2" in https://bugzilla.redhat.com/show_bug.cgi?id=2097010

There are two "host" entries in the shell: the original "localhost" one for the original session with user "root", and a second one for "127.0.0.1" (thus, same actual machine, but Cockpit doesn't know that) for a non-root user.

The access level indicator should not be shown for "root@localhost", and it should be shown for "user@127.0.0.1" when switching between these two entries in the Shell.

This doesn't happen.

mvollmer:
The "proxy" property of the SuperuserIndicator was not updated in time
with a change to the "host" property since "componentDidUpdate" runs
after render.  Let's just keep the proxy up-to-date in "render"
itself.

Fixes #17467

### DIFF
--- a/pkg/shell/topnav.jsx
+++ b/pkg/shell/topnav.jsx
@@ -61,8 +61,8 @@ export class TopNav extends React.Component {
             osRelease: {},
         };
 
-        this.superuser_connection = cockpit.dbus(null, { bus: "internal", host: props.machine.connection_string });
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks!

PR #17469, opened 2022-06-20
ws: Pass UrlRoot to cockpit.js

### LINKED ISSUE #17444: cockpit.js does not respect URL root without login page

martinpitt:
So far we have relied on pkg/static/login.js to compute the URL root
(`UrlRoot=` setting in cockpit-ws), and put it into localStorage as
`url-root`. However, these days we don't actually invoke the login page
in many scenarios: Kerberos/Smart card authentication, SSH auth with the
bastion container, custom setups with Bearer token/OAuth, etc.

To make this more reliable, have cockpit-ws pass the UrlRoot as a
meta header, and read that in cockpit.js.
While at it, rewrite cockpit_channel_inject_perform() to use automatic
memory management.

We still need to keep the `url-root` localStorage approach for a while,
as we support old ws connecting to new bridge, and vice versa.

Fixes #17444


martinpitt:
@jelly : This WFM with [console.dot](https://github.com/jelly/console.dot).

src/ws/cockpitchannelresponse.c
```diff
@@ -82,49 +82,47 @@ cockpit_channel_inject_perform (CockpitChannelInject *inject,
...
 
-  if (!inject->base_path)
+  const gchar *url_root = cockpit_web_response_get_url_root (response);
+
+  if (!url_root && !inject->base_path)
     return;
 
-  str = g_string_new ("");
+  GString *str = g_string_new ("");
   creds = cockpit_web_service_get_creds (inject->service);
```
allisonkarlitskaya, line 94: Move this into the else branch, please, and bring the variable declaration with it.

src/ws/cockpitchannelresponse.c
```diff
@@ -82,49 +82,47 @@ cockpit_channel_inject_perform (CockpitChannelInject *inject,
...
+  g_autofree gchar *prefixed_application = NULL;
 
-  if (!inject->base_path)
+  const gchar *url_root = cockpit_web_response_get_url_root (response);
+
+  if (!url_root && !inject->base_path)
     return;
 
-  str = g_string_new ("");
+  GString *str = g_string_new ("");
```
allisonkarlitskaya, line 93: This ought to be `g_autoptr(GString)`.

src/ws/cockpitchannelresponse.c
```diff
@@ -82,49 +82,47 @@ cockpit_channel_inject_perform (CockpitChannelInject *inject,
...
+          g_string_append_printf (str, "\n    <base href=\"%s/@%s%s\">",
+                                  prefixed_application, inject->host, inject->base_path);
+        }
     }
 
-  base = g_string_free_to_bytes (str);
-  filter = cockpit_web_inject_new (marker, base, 1);
-  g_bytes_unref (base);
-
+  g_autoptr(GBytes) content = g_string_free_to_bytes (str);
```
allisonkarlitskaya, line 123: ...and here, `g_steal_pointer(&str)`

allisonkarlitskaya (changes requested):

src/ws/cockpitchannelresponse.c
martinpitt: It's also used in the `then` branch, so I didn't move the initialization. But I did move the declaration.

src/ws/cockpitchannelresponse.c
martinpitt: Nice trick, thanks! Done.

martinpitt:
@allisonkarlitskaya Friendly review ping? I think you mentioned something about the commit message, I'm happy to adjust it.

allisonkarlitskaya (approved):

PR #17408, opened 2022-05-31
Enable CSS linting through stylelint

jelly:

PR #17471, opened 2022-06-20
Add stylelint as CSS/SCSS linter

jelly:
This PR adds stylelint as a css/scss linter for our stylesheets. Currently none of the warnings it emits are fixed but all offending rules are disabled. Stylelint can automatically fix some rules but I opted to keep this PR and land the `node_modules` changes first then we can incrementally enable rules and fix issues. As the PR that fixes these things has +3000/-2800 changes.

https://github.com/cockpit-project/cockpit/pull/17408/commits

Performance implications:
```
[jelle@t14s][~/projects/cockpit/add_stylelint]%hyperfine --warmup 3 'STYLELINT=0 ./tools/webpack-make systemd' && hyperfine --warmup 3 'STYLELINT=1 ./tools/webpack-make systemd'
Benchmark 1: STYLELINT=0 ./tools/webpack-make systemd
  Time (mean ± σ):     27.338 s ±  0.301 s    [User: 46.508 s, System: 1.162 s]
  Range (min … max):   26.845 s … 27.787 s    10 runs

Benchmark 1: STYLELINT=1 ./tools/webpack-make systemd
  Time (mean ± σ):     27.938 s ±  0.445 s    [User: 47.845 s, System: 1.192 s]
  Range (min … max):   27.326 s … 28.763 s    10 runs
```


martinpitt (approved):
Very good approach to land the dependency first. Thanks!

PR #17473, opened 2022-06-21
PoC: All the changes we need to support 3scale/clouddot

martinpitt:
We want to support running cockpit-ws behind https://www.3scale.net/. This PR subsumes all the changes that we need for this on the cockpit side, so that we can easily build a container and packages out of them. We will land these in individual PRs, though.

Dependencies:
 - PR #17469
 - PR #17640 

With the extra commit for hacking the container, I published this as https://quay.io/repository/rhn_engineering_mpitt/ws with:

    podman build -t quay.io/rhn_engineering_mpitt/ws containers/ws/
    podman push quay.io/rhn_engineering_mpitt/ws

 - [x] Finally decide whether we can do everything through /wss on 3scale and ignore the /app route -- then drop the WebsocketRoot= commit

Superseded by https://github.com/jelly/console.dot/pull/47

PR #17474, opened 2022-06-21
Clean up tools/ and test/ directory

martinpitt:
This is preparation for moving test/verify/* to just test/*.

Note that this requires updating the path to git hooks in all existing checkouts. Sorry about that..


martinpitt:
@jelly : This [LGTM failure](https://lgtm.com/projects/g/cockpit-project/cockpit/logs/rev/pr-802a80a70d8470500a18e44f1ac02f5ed1b09a17/lang:cpp/stage:Build%20merge_eef6a8189f58700f39286328688f539b30173ee5) doesn't LGTM :cry: it already failed the same way in #17473. 

```
[2022-06-21 09:47:38] [build-stderr] /opt/src/node_modules/stylelint-webpack-plugin/node_modules/jest-worker/build/index.js:110
[2022-06-21 09:47:38] [build-stderr]   _ending;
[2022-06-21 09:47:38] [build-stderr]          ^
[2022-06-21 09:47:38] [build-stderr] SyntaxError: Unexpected token ;
```
Any idea about this?

jelly:
> @jelly : This [LGTM failure](https://lgtm.com/projects/g/cockpit-project/cockpit/logs/rev/pr-802a80a70d8470500a18e44f1ac02f5ed1b09a17/lang:cpp/stage:Build%20merge_eef6a8189f58700f39286328688f539b30173ee5) doesn't LGTM cry it already failed the same way in #17473.
> 
> ```
> [2022-06-21 09:47:38] [build-stderr] /opt/src/node_modules/stylelint-webpack-plugin/node_modules/jest-worker/build/index.js:110
> [2022-06-21 09:47:38] [build-stderr]   _ending;
> [2022-06-21 09:47:38] [build-stderr]          ^
> [2022-06-21 09:47:38] [build-stderr] SyntaxError: Unexpected token ;
> ```
> 
> Any idea about this?

```
  class Worker {
    _ending;
    _farm;
    _options;
    _workerPool;

    constructor(workerPath, options) {
      var _this$_options$enable,
```

So that's ES6 classes? Which seems valid syntax https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Classes_in_JavaScript

allisonkarlitskaya (commented):
I'm in favour of everything here except moving `test/static-code` to `test/common/`.

I think it doesn't make sense because this script isn't intended to be used from other projects, right?  `test/static-code` is also "API" in a sense, since the git hooks expect this script to be present in all commits of cockpit.

allisonkarlitskaya:
> > I think it doesn't make sense because this script isn't intended to be used from other projects, right?
> 
> It _is_ intended to be used, and that's what we [do in cockpit-machines already](https://github.com/cockpit-project/cockpit-machines/blob/main/Makefile#L147). I also want to use it in podman, which currently duplicates this functionality (in an inferior way). That's my main rationale why it belongs into test/common, similar to run-tests.

Honestly, I think this is a really bad idea.  We still do things like where we want to decrease the maximum line length a bit further and do a bunch of commits doing that, and then lower the number in this script.  Having to worry about syncing that cross-project (or risking blocking other important updates) sounds like it could turn into a big problem, real fast.

Maybe it would also be worth thinking about a way to deal with the "API break" problem.  Symlink?  A future list of files that we must always check out, in order to prevent this from being an issue going forward?

I think I'd really prefer to drop this one commit for now...

martinpitt:
Okay, *shrug*, dropped that.

allisonkarlitskaya (approved):

martinpitt:
> Honestly, I think this is a really bad idea. We still do things like where we want to decrease the maximum line length a bit further and do a bunch of commits doing that, and then lower the number in this script. Having to worry about syncing that cross-project (or risking blocking other important updates) sounds like it could turn into a big problem, real fast.

This is a bounded problem, though: once we are down at an acceptable line length, and all projects bumped to that cockpit API, it won't change any more. If needed, we can parameterize this with an option until then, but so far this hasn't bitten us in cockpit-machines.

PR #17476, opened 2022-06-21
Fix lgtm

jelly:
More details of the failure are described here https://github.com/cockpit-project/cockpit/pull/17474#issuecomment-1161542371


src/common/cockpitauthorize.c
```diff
@@ -121,6 +121,7 @@ cockpit_authorize_nonce (size_t length)
   do
     {
       errno = 0;
+      printf("%d\n", errno);
```
martinpitt, line 124: This is talking to stdout and thus to the json protocol. That will break it :-)

martinpitt (approved):

PR #17481, opened 2022-06-22
Fix LGTM warnings/errors

jelly:
This PR fixes a bunch of warnings generated by LGTM and accumulated over the years https://lgtm.com/projects/g/cockpit-project/cockpit/alerts/?mode=list&severity=warning&lang=javascript

Ideally LGTM would complain instead of turn green in PR's.

### DIFF
--- a/pkg/lib/cockpit.js
+++ b/pkg/lib/cockpit.js
@@ -2524,11 +2524,7 @@ function factory() {
                 href = href.substr(1);
 
             const pos = href.indexOf('?');
-            let first = href;
-            if (pos === -1)
-                first = href;
-            else
-                first = href.substr(0, pos);
+            const first = (pos === -1) ? href : href.substr(0, pos);
             const path = decode_path(first);
             if (pos !== -1 && options) {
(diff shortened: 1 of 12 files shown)


pkg/lib/cockpit.js
```diff
@@ -2524,7 +2524,7 @@ function factory() {
                 href = href.substr(1);
 
             const pos = href.indexOf('?');
-            let first = href;
+            let first;
             if (pos === -1)
                 first = href;
```
martinpitt, line 2529: This is better written as

    const first = (pos === -1) ? href : href.substr(0, pos);

martinpitt (approved):
Thanks! Still some test failures, not sure if they are related or just need a retry. Small issue in case you need to re-push.

jelly:
> Thanks! Still some test failures, not sure if they are related or just need a retry. Small issue in case you need to re-push.

Fixed, re-pushed. Only packit rawhide fails due to python 3.11 which is expected.

martinpitt (approved):
LGTM! :wink:

Issue #1091, opened 2014-08-28
ip settings: automatically suggest a gateway based on IP


andreasn:
Correct me if I'm wrong, but in most situations, this is true:
If I set my static IP address to 192.168.11.5, the gateway is most likely 192.168.11.1
Filling this automatically would be a nice touch (with the ability to change the gateway of course)

mvollmer:
Good idea.  I want this every time I use the dialog.

gil-obradors:
I'm in!

https://github.com/cockpit-project/cockpit/blob/69a93dc68fae67e5f9b8e31693d1244f8a6d08e5/pkg/networkmanager/ip-settings.jsx#L67

Need to write a function that encapsulates setAddresses by
- Every time that address field is changed
  - Set netmask according address class
  - Set gateway with the same address ending with 1

gil-obradors:
Could be something like this?

![Peek 2022-06-10 22-30](https://user-images.githubusercontent.com/6016929/173147264-3186961e-198e-4576-9ff5-2c4e09ec5809.gif)


https://github.com/cockpit-project/cockpit/compare/main...gil-obradors:ip-helper

jelly:
Without taking a too detailed look this looks great! You should be able to insert an address and then later be able to change the prefix/gateway to something you prefer.

gil-obradors:
...And tests: https://github.com/gil-obradors/cockpit/blob/3cc547b071267ff1e38beadb670cac45012ada3a/test/verify/check-networkmanager-ip-helper

PR #17484, opened 2022-06-23
networking: ipv4 helper for interface configuration

gil-obradors:
Will write the gateway and netmask in function of address class.  Will propuse gateway ended with .1 or .254. 
Allows later free edit.

https://github.com/cockpit-project/cockpit/issues/1091


test/verify/check-networkmanager-settings
```diff
@@ -186,6 +186,44 @@ class TestNetworkingSettings(NetworkCase):
         b.click('#network-ip-settings-cancel')
         b.wait_not_present("#network-ip-settings-dialog")
 
+    def testIpHelper(self):
```
martinpitt, line 189: The `TestNetworkingSettings` class is rather "expensive": It starts *two* new VMs for each new test case, which is disproportionate to the "harmless" things that happen in this test. This should either be merged into `testOtherSettings`, or be moved into a new `@nondestructive` test class.

It's probably the easiest way to move this test into test/verify/check-networkmanager-basic.

As this does not check external APIs, please add `TestNetworkingBasic.testIpHelper` (the name after the move) to test/browser/run-test.sh before the `TestNetworkingBasic.testNoService` line. This will avoid running this test in packit (we need to not make this take too long).

test/verify/check-networkmanager-settings
```diff
@@ -186,6 +186,44 @@ class TestNetworkingSettings(NetworkCase):
...
+        b.wait_val("#network-ip-settings-netmask-0", "255.255.255.0")
+        b.wait_val("#network-ip-settings-gateway-0", "192.168.1.254")
+        # free manual edit
+        b.set_input_text("#network-ip-settings-netmask-0", "255.255.255.128")
+        b.set_input_text("#network-ip-settings-gateway-0", "192.168.1.126")
+
+        b.wait_val("#network-ip-settings-address-0", "192.168.1.1")
+        b.wait_val("#network-ip-settings-netmask-0", "255.255.255.128")
+        b.wait_val("#network-ip-settings-gateway-0", "192.168.1.126")
+
```
martinpitt, line 226: It's a bit unclear what this does -- as such it just checks that the values are exactly the ones that you just set. So this doesn't effectively test anything. I suppose you want to ensure that it does not change the values automatically once an input line got a customized value, but that is a race condition.

I suggest to actually confirm/close the dialog, and validate the setting with either `nmcli` or inspecting the properties in the UI, as you prefer. Then you know that they *really* got applied.

pkg/networkmanager/ip-settings.jsx
```diff
@@ -130,7 +130,25 @@ export const IpSettingsDialog = ({ topic, connection, dev, settings }) => {
 
         return false;
     };
-
+    const addressToMask = (address) => {
+        const addSplited = address.split('.');
```
martinpitt, line 134: "splited" is not a word (and "splitted" isn't either, English is weird). I suggest to just rename it to "split".

pkg/networkmanager/ip-settings.jsx
```diff
@@ -130,7 +130,25 @@ export const IpSettingsDialog = ({ topic, connection, dev, settings }) => {
...
+        if (addSplited.length !== 4)
+            return "";
```
martinpitt, line 136: Please return `null` for "not a valid value".

pkg/networkmanager/ip-settings.jsx
```diff
@@ -130,7 +130,25 @@ export const IpSettingsDialog = ({ topic, connection, dev, settings }) => {
...
+        if (addSplited[0] >= 0 && addSplited[0] <= 127) {
+            return "255.0.0.0";
+        } else if (addSplited[0] >= 128 && addSplited[0] <= 191) {
+            return "255.255.0.0";
+        } else if (addSplited[0] <= 192 && addSplited[0] <= 223) {
+            return "255.255.255.0";
+        }
```
martinpitt, line 143: This will return `undefined` for addresses >= 224. React is cool with that, but it looks asymmetric here. Perhaps an `else { return null }` to clarify this? (Also see above)

pkg/networkmanager/ip-settings.jsx
```diff
@@ -130,7 +130,25 @@ export const IpSettingsDialog = ({ topic, connection, dev, settings }) => {
...
+    };
+    const addressToGateway = (address, gateway) => {
+        const addSplited = address.split('.');
+        if (addSplited.length !== 4)
+            return "";
+        addSplited[3] = (address === gateway) ? (addSplited[3] === "1") ? "254" : "1" : (addSplited[3] === "1") ? "254" : "1";
```
martinpitt, line 149: The outer ? and : branches compute exactly the same, so the `address === gateway` check does not make any difference. It's also hard to read the nested ?: , please break this up into three lines and enclose the inner ones with ().

How can address == gateway even happen? By definition the gateway is *not* your current machine, and you can't assign the same IP. This case should just be invalid, i.e. the helper should just return `null` then.

martinpitt (changes requested):
Thank you for working on this! Please add "Fixes #1091" to the commit message in the next round.

test/verify/check-networkmanager-settings
gil-obradors: ack

test/verify/check-networkmanager-settings
gil-obradors: ack.
I needed to activate the interface with dhcp_cidr parameter defined `self.add_veth(iface, dhcp_cidr="10.111.113.2/20")` because if `dhcp_cidr` is not set interface isn't activated.
Maybe this line can go out of _if scope_, but I don't know if it has more involvement
https://github.com/cockpit-project/cockpit/blob/8898cefcf66fa05c588cd6b760917f42f88bfc6a/test/verify/netlib.py#L44

pkg/networkmanager/ip-settings.jsx
gil-obradors: ack

pkg/networkmanager/ip-settings.jsx
gil-obradors: ack

gil-obradors:
Oh, don't review yet. Something is wrong. I delete addressToGateway, but is here again... !

gil-obradors:
I'm going to reduce `addressToMask` and `addressToGateway` to one function `addressIpv4Helper`, which is going to return ```{netmask:xxx,gateway:xxx}```
Simple, short and easier to understand

martinpitt:
Thanks @gil-obradors : I'm AFK next week, and cleaning up my request lists. If you need a review next week, please ask @jelly (he can also help with triggering the tests), otherwise I'm happy to re-review once I'm back.

Also, please always rebase to current origin/main on each push. This will ensure that the "testing-farm:*" tests run properly. Thanks!

jelly:
Triggered our tests, from a quick glance the code looks alright to me.

gil-obradors:
@martinpitt have nice week!

@jelly ready for me!

garrett:
I think this would be nice in the release notes for the next release. (Added label.)

martinpitt (approved):
This looks excellent now, thank you! I pushed a rebase to let the tests run properly, and if this goes green, this can land. :clap:

martinpitt:
Hmm, the test [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17484-20220712-043241-fa89b4fb-fedora-36/log.html#313):

   Error: Failed to modify connection 'cockpit1': ipv4.gateway: gateway cannot be set if there are no addresses configured

martinpitt (changes requested):
test failures

gil-obradors:
Oh!

https://github.com/cockpit-project/cockpit/blob/307615b48e39473dfa498c941628ce411ce0ca59/test/verify/check-networkmanager-basic#L113
This command with single quotes inside works fine with sh console. 

https://github.com/cockpit-project/cockpit/blob/307615b48e39473dfa498c941628ce411ce0ca59/test/verify/check-networkmanager-firewall#L69
looks like a distribution/python problem.

These tests have leaked from other changes, right?
We have to wait for them to be solved and launch them again in a few days?

jelly:
> Oh!
> 
> https://github.com/cockpit-project/cockpit/blob/307615b48e39473dfa498c941628ce411ce0ca59/test/verify/check-networkmanager-basic#L113
> 
> 
> This command with single quotes inside works fine with sh console.
> https://github.com/cockpit-project/cockpit/blob/307615b48e39473dfa498c941628ce411ce0ca59/test/verify/check-networkmanager-firewall#L69
> 
> 
> looks like a distribution/python problem.
> These tests have leaked from other changes, right? We have to wait for them to be solved and launch them again in a few days?

I can reproduce `TestNetworkingBasic.testBasic` here locally, it is due to the change you made:

Previously the gateway would have been empty and not prefilled:

image

So this would fix that:
```
         b.set_input_text('#network-ip-settings-netmask-0', "255.255.192.0")
+        b.set_input_text('#network-ip-settings-gateway-0', "")
```

However I am not sure if the change is now correct.

gil-obradors:
Short summary

We are asking for delete a ipv4.addresses information without deleting ipv4.gateway. NM don't want to store a gateway with empty addresses on interface. Totally understandable.

I push the solution for this problem, but this comes from past. How Can We know the last time that this test pass?



Large explanation

I have set a breakpoint 
https://github.com/cockpit-project/cockpit/blob/307615b48e39473dfa498c941628ce411ce0ca59/test/verify/check-networkmanager-basic#L113
and execute test with python debug. I have done all with last commit from main. I have rebuilt VM. We are running this nmcli version
```
print(m.execute("nmcli --version"))
nmcli tool, version 1.38.0-2.fc36
```
Initial config
```
print(m.execute("nmcli connection show cockpit1 | sed -e '/ipv4\.[ag]/!d' "))
ipv4.addresses:                         1.2.3.4/18
ipv4.gateway:                           1.2.3.1
```
I try to delete ipv4.addresses without delete ipv4.gateway. Like testBasic do.
```
print(m.execute("nmcli connection modify cockpit1 ipv4.addresses ''"))
Error: Failed to modify connection 'cockpit1': ipv4.gateway: gateway cannot be set if there are no addresses configured
  File "/snap/pycharm-professional/290/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "/home/gil/cockpit/bots/machine/machine_core/ssh_connection.py", line 368, in execute
    raise subprocess.CalledProcessError(proc.returncode, command, output=output)
subprocess.CalledProcessError: Command 'nmcli connection modify cockpit1 ipv4.addresses ''' returned non-zero exit status 1.
```
Ok, that is the knowed error.
I delete first gateway, then addresses
```
print(m.execute("nmcli connection modify cockpit1 ipv4.gateway ''"))
print(m.execute("nmcli connection modify cockpit1 ipv4.addresses ''"))
print(m.execute("nmcli connection show cockpit1 | sed -e '/ipv4\.[ag]/!d' "))
ipv4.addresses:                         --
ipv4.gateway:                           --
```


Now works fine.
I review `test/verify/check-networkmanager-firewall TestFirewall.testAddCustomServices` but pass without problems. I don't know way not in past


Comming new-force-push

martinpitt (approved):
Thanks @gil-obradors ! This LGTM now, happy to see this land in the next release! This still needs a release note, I'll craft one tomorrow unless you want to.

garrett:
I've edited the animation to crop out the browser UI, add a line to the bottom (so the white of the modal doesn't hit the white of the page), removed the mistake in typing, cut out the tab switching, and removed the mouse moving around to stop the recording.

PR #17486, opened 2022-06-23
Drop PatternFly 3, OpenSans fonts, clean up examples

martinpitt:
We will commit patternfly.css (aka cockpit.css) to RHEL 8's dist-git instead. There is little sense in rebuilding the same file over and over again, PF3 does not change any more, and we don't even test this. So it's actually safer to ship a known-good version in dist-git. If needed, we can rebuild it from an old rhel-8.y branch.

This is how Cockpit looks with OpenSans fonts on CentOS 7:

image

image

And this is how it looks with the browser default fonts:

image

image

The font is noticeably different, but IMHO this is acceptable for a version jump across two RHEL releases.


garrett:
> The font is noticeably different, but IMHO this is acceptable for a version jump across two RHEL releases.

No, this isn't really acceptable; we should have a sans-serif in the font stack so it at least kind of looks correct-ish.

martinpitt:
> I guess this would have to be changed in ancient Cockpit? 

Yes, unfortunately; and RHEL 7 is "important security fixes only" at this point. So we mostly need to decide whether to live with this, or keep pulling in and building PF3 and shipping OpenSans for several more years. Even our RedHat fonts are outdated AFAICS, but OpenSans is even more so..

garrett:
> So we mostly need to decide whether to live with this

"**this**", defined here: Someone connecting to a RHEL 7 machine also running Cockpit from a RHEL 8 or RHEL 9 machine running Cockpit, getting wrong but still legible fonts; most of which nobody would notice. (The serif fonts are quite noticeable.)

(If I understand this correctly.)

martinpitt:
@garrett: Exactly, thanks!

marusak (approved):
Nice cleanups! Thanks!

PR #17487, opened 2022-06-23
tap-cdp: return non-zero if we see 'not ok' lines

allisonkarlitskaya:
This used to be handled by the tap driver, but since we don't use that
anymore, we need to add a check for it here.

Fortunately, we didn't break anything in the meantime.


martinpitt (approved):
Whoops! Thanks for spotting!

PR #17488, opened 2022-06-23
base1: fix some test-http asserts

allisonkarlitskaya:
Using `assert.async()` is a bit dangerous with QUnit: you need to make
sure the `done()` method always gets called.

We have a few cases in our tests that fail when run against the Python
bridge, but don't call `done()`.  Clean those up.

In general, converting the tests to async/await syntax is easier than
trying to sort out the state machines involved here.

### DIFF
--- a/pkg/base1/test-http.js
+++ b/pkg/base1/test-http.js
@@ -119,7 +119,7 @@ QUnit.test("not found", function (assert) {
                 assert.strictEqual(ex.problem, null, "mapped to cockpit code");
                 assert.strictEqual(ex.status, 404, "has status code");
                 assert.equal(ex.message, "Not Found", "has reason");
-                assert.true(data.includes('<h1>Not Found</h1>'), "got body");
+                assert.true(data !== undefined && data.includes('<h1>Not Found</h1>'), "got body");
             })
             .always(function() {
                 assert.equal(this.state(), "rejected", "should fail");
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
This is the set of fixes required to get "http" out of the "hanging" column into the "failing" column on the python branch.

pkg/base1/test-http.js
```diff
@@ -284,102 +284,47 @@ QUnit.test("http promise recursive", function (assert) {
...
-                    ex.problem = "not-found";
-                assert.strictEqual(ex.problem, "not-found", "can't resolve is not found");
-            })
-            .always(function() {
-                assert.equal(this.state(), "rejected", "should fail");
-                done();
-            });
+    assert.rejects(cockpit.http({ port: 8080, address: "the-other-host.example.com" }).get("/"),
+                   /* Unfortunately we can see either of these errors when running unit tests */
+                   ex => { return ex.problem === "timeout" || ex.problem === "not-found" });
```
martinpitt, line 327: Style guide: Drop useless `{ return .. }` wrappers for arrows, to stay functional. I.e. `ex => ex.problem === "timeout" || ex.problem === "not-found"`.

Not a reason to block this, though.

martinpitt (approved):
Thanks, this looks really nice!

Also, for future PRs: no-test please.

pkg/base1/test-http.js
allisonkarlitskaya: Oof.  I tried to translate it more literally first, which is why I did that.  Oh well.

PR #17489, opened 2022-06-23
lib/page: Simplify new row animation

### LINKED ISSUE #17482: Kebab menus are under new items, due to z-index stacking order
### Explain what happens

1. Open a kebab menu before or during new items show up.
2. See that it's behind the new item

### Version of Cockpit

_No response_

_No response_

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

garrett:
Both transform and opacity adjustment change the stacking order. This causes issues like popups displaying under the rows during animation.

Fixes #17482

### DIFF
--- a/pkg/lib/page.scss
+++ b/pkg/lib/page.scss
@@ -84,37 +84,13 @@ a.disabled:hover {
     --ct-animation-new-background: #fdf4dd;
 }
 
-/* Simple fade animation by default (mainly for reduced motion) */
+/* Animation background is instantly yellow and fades out halfway through */
 @keyframes ctNewRow {
   0% {
     background-color: var(--ct-animation-new-background);
   }
   50% {
     background-color: var(--ct-animation-new-background);
-  }
-}
-
-/* Redefine to a fancy animation when reduced-motion is not requested */
-@media (prefers-reduced-motion: no-preference) {
-  @keyframes ctNewRow {
-    0% {
-      background-color: var(--ct-animation-new-background);
-      transform: scaleX(0) scaleY(0);
-      opacity: 0;
-    }
-    10% {
-      transform: scaleX(1) scaleY(1);
-    }
-    12% {
-      transform: scaleX(1) scaleY(1.25);
-      opacity: 1;
-    }
-    15% {
-      transform: scaleX(1) scaleY(1);
-    }
-    70% {
-      background-color: var(--ct-animation-new-background);
-    }
   }
 }
 


garrett:
Before this change, we have 2 animations: a simplified version for prefers reduced motion and the one usually used with additional animation. I removed the one with more animation so we only use the simplified version. Meaning: This change is 100% code removal and a comment tweak.

garrett:
This would have to be merged and then pulled into Cockpit-Podman and other `cockpit-*` places it's used.

martinpitt:
Thank you! There are [two stratis pixel tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17489-20220623-115347-2e6bfa68-fedora-36/log.html) with very subtle difference. Retrying to see if they are a glitch or systematic from this change.

martinpitt:
Happened again. The difference is on the "Create" button, its upper and lower right corners have some rounding error (in the geometrical sense :grin: ). I have no clue why a table row animation would affect this, but :man_shrugging: let's just update them?

martinpitt:
I wanted to try `test/common/pixel-tests fetch 17489`, but too slow in the train. Someone?

marusak:
> I wanted to try test/common/pixel-tests fetch 17489, but too slow in the train. Someone?

```
$ test/common/pixel-tests fetch 17489

  File "<string>", line 5, in <module>
KeyError: 'head'
  File "<string>", line 7, in <module>
TypeError: string indices must be integers

```

:) I will update pixel tests and also look at the script

marusak (approved):

garrett:
> Happened again. The difference is on the "Create" button, its upper and lower right corners have some rounding error (in the geometrical sense grin ). I have no clue why a table row animation would affect this, but man_shrugging let's just update them?

Wait. What? Which create button and how? Can you record a video? (If it's recorded in MP4, you could upload it here to GitHub.)

martinpitt:
@garrett : See the [pixel test failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17489-20220623-115347-2e6bfa68-fedora-36/log.html) above.

garrett:
I see literally no difference whatsoever with the sliders. (It could be broken? Or just no difference?) And none of the screenshots have anything to do with new row animation. None of these screenshots should have any animation at all. And the change in this PR is completely unrelated to these modal screenshots... so I'm really confused about this.

This PR literally just removed a block of CSS that specified a `@keyframe`. That's it.

Issue #17329, opened 2022-05-10
lint css via webpack


garrett:
We should probably add in some sort of linting for our CSS.

This would warn people when they try to use bad CSS practices, which would solve some headaches and me manually pointing out some of the issues. :wink: 

This post has details on the concept and suggests stylelint (it should be pointed out that it's a little biased, as it's from the author of stylelint):
https://css-tricks.com/stylelint/

But stylelint _does_ look good and it does have webpack integration and does also support SASS (SCSS):

- https://stylelint.io/
- https://webpack.js.org/plugins/stylelint-webpack-plugin/
- https://www.npmjs.com/package/stylelint-webpack-plugin

We'll probably have to introduce it with some relaxed rules at first and then tighten it up over time as we make fixes across the codebases.

garrett:
@martinpitt: As this is both webpack and CSS related, it's probably a matter of you and me working on this together, most likely?

(And to anyone else who might be reading this who wants to take it on instead: You're more than welcome to do so!)

martinpitt:
PR #17471 laid the foundation for this (just landed). Now we can resurrect PR #17408 and land this piece by piece.

jelly:
Auto fixing rules to appy in future pull requests:

- [ ] 
- [ ] I found all the autofixing rules according to the documentation, made a template file that disabled all the rules (after importing the base ones) and wrote a 1-liner loop for fish shell (the shell I use) that iterated through the listing of the commands, excluded the line from the template file, ran stylelint with the custom config with --fix, and did a git commit with a basic summary.
- [ ] 
- [ ] Fish loop:
- [ ] 
- [ ] for i in (cat autorules); echo $i; grep -v \"$i\" stylelint-template.json > rules.json; npx stylelint --config rules.json --fix "pkg/**/*.*css"; git commit -a -m "stylelint: $i"; end
- [ ] Template to turn off things:
- [ ] 
- [ ] {
- [ ]   "extends": "stylelint-config-standard-scss",
- [ ]   "rules": {
- [ ] 	"alpha-value-notation": null,
- [ ] 	"at-rule-empty-line-before": null,
- [ ] 	"at-rule-name-case": null,
- [ ] 	"at-rule-name-space-after": null,
- [ ] 	"at-rule-no-vendor-prefix": null,
- [ ] 	"at-rule-semicolon-newline-after": null,
- [ ] 	"block-closing-brace-empty-line-before": null,
[...]

PR #17493, opened 2022-06-23
Enabling stylelint rules take 1

jelly:
This is the first take of the [auto fixeable](https://github.com/cockpit-project/cockpit/issues/17329#issuecomment-1164754748), should include the most non-controversial rules.


martinpitt (approved):
Cheers! These are very reviewable.

PR #17494, opened 2022-06-23
test: document more testlib functions

jelly:


test/common/testlib.py
```diff
@@ -351,38 +351,84 @@ def mousedown(self, selector):
         self.mouse(selector + ":not([disabled]):not([aria-disabled=true])", "mousedown", 0, 0, 0)
 
     def val(self, selector):
+        """Get the value attribute of a selector
```
martinpitt, line 354: Please consistently add a period to short description (or consistently drop it).

Also, this description does not really add any extra information. Does that somehow look weird in pydoc if there is no docstring?

Also, can we do the `:type` with proper type annotations instead: `def val(self, selector: str)` is more declarative and useful (mypy!)

test/common/testlib.py
```diff
@@ -351,38 +351,84 @@ def mousedown(self, selector):
...
+        This also emits a change DOM change event.
+
+        :param selector: the selector to set the value of
+        :param val: the value to set
+        """
         self.wait_visible(selector + ':not([disabled]):not([aria-disabled=true])')
         self.call_js_func('ph_set_val', selector, val)
 
     def text(self, selector):
+        """Get an elements textContent value
```
martinpitt, line 374: "element's"

test/common/testlib.py
jelly: Typing exists since Python `3.5`, it seems RHEL 8 has `3.6` so that should be possible. It also looks way nicer at least in my editor, which is my primary use case.

image

Sadly I realised neovim's LSP cannot determine in a TestClass what self.browser is as it's initially `None` 

image

Not sure if typing will help here.

pydoc out of the box is broken and annoying to fix:
```
cd test/common
pydoc -b
```

```
pydoc.ErrorDuringImport: problem in testlib - ModuleNotFoundError: No module named 'testvm'
```
So either you need some import magic and fiddle every time we change `testvm` which no one will test :)

jelly: Ok scratch that, I want `typing` now, we just need to teach `MachineCase` that `self.browser` can be a `Browser` instance:

image

```
@@ -1357,7 +1356,9 @@ class MachineCase(unittest.TestCase):
         if opts.address and self.provision is not None:
             raise unittest.SkipTest("Cannot provision multiple machines if a specific machine address is specified")

-        self.browser = None
+        from typing import Type, Optional
+
+        self.browser: Optional[Browser] = None
```

jelly: This is partial success, as my LSP does not know where `testlib` is, in `test/verify/check-connection` which we ought to really fix:

image

jelly: So it seems that pyright (my LSP) does not resolve our `sys.path` paths?! annoying!

jelly: Creating a new file next to `testlib.py` and import `MachineCase` with a simple class that inherits `MachineCase` then `self.browser` is Unknown | None, so do I not understand Python inheritance or why does it work when I inherit `MachineCase` in `testlib.py`.

jelly: And nevermind, this works now with the `Optional` type in testlib.py and making your editor able to include the correct git worktree. Even in cockpit-podman!

image


image

martinpitt:
@jelle: My other concerns from above are still current: inconsistent ending of the summary with a period (or not), and adding `:type` docstrings. I thought you wanted to go ahead with adding proper type annotations now? even if we don't use them for anything just yet, I think they are much nicer for starting to document the API.

martinpitt (approved):
Thanks! Let's go with this for now. I'm triggering the required fedora-coreos test, please land when green.

PR #17498, opened 2022-06-24
firewall: Update docs link

### LINKED ISSUE #17496: Validate all URLs
Checked 154 URLs out of which 2 is/are invalid:
    https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/using-the-web-console-for-managing-firewall_system-management-using-the-rhel-8-web-console
    https://github.com/patternfly/patternfly-react/issues/5993

marusak:
Fixes #17496

I will get in touch with rhel docs team to ask about this. We got promised that they will not break these urls :man_shrugging:

### DIFF
--- a/pkg/networkmanager/manifest.json
+++ b/pkg/networkmanager/manifest.json
@@ -27,7 +27,7 @@
                 },
                 {
                     "label": "Managing firewall",
-                    "url": "https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_systems_using_the_rhel_8_web_console/using-the-web-console-for-managing-firewall_system-management-using-the-rhel-8-web-console"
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! Note that the URL checker also complained about https://github.com/patternfly/patternfly-react/issues/5993 . That is a redirect to https://github.com/patternfly/patternfly/issues/4754 , but we should update that as well.

marusak:
> Note that the URL checker also complained about https://github.com/patternfly/patternfly-react/issues/5993 . That is a redirect to https://github.com/patternfly/patternfly/issues/4754 , but we should update that as well.

It should not complain about redirect in PF, see https://github.com/cockpit-project/cockpit/blob/main/tools/urls-check#L41

I ran the check on this branch and it reports all urls as valid. I think there was just outage on GitHub or so.

PR #17499, opened 2022-06-24
Stylelint Part Two

jelly:


martinpitt (approved):
Hah, that was easier than I feared :-) thanks!

PR #17500, opened 2022-06-24
packagekit: Add "Reboot after completion" switch to update progress page

martinpitt:
This allows users to start a manual update and queue a reboot without
having to babysit the progress.

Rework the progress page layout a bit:

 - Use a Grid layout, so that the update log has the same alignment than
   the progress bar.

 - Stop centering the "Cancel" button, which was a bit dubious wrt. PF
   recommendations anyway. Show it to the right of the progress bar
   instead.

 - Hide the cancel button instead of disabling it, once cancellation is
   not allowed any more.

 - Rename "Update log" to "View update log" to clarify what the action
   means.

 - Increase the spinner size, to make it easier to see.

https://bugzilla.redhat.com/show_bug.cgi?id=2056786

-----

 - Builds on top of PR #17517

## Software Updates: Reboot after applying updates

The update progress page offers you to reboot the machine after applying updates succeeded. If the machine can be rebooted without careful planning, this avoids having to wait and do another interactive step.

Screen Shot 2022-07-13 at 13 33 36

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -24,15 +24,18 @@ import React, { useState, useEffect } from "react";
 import ReactDOM from 'react-dom';
 
 import {
-    Alert, Badge, Button, Gallery, Modal, Progress, Popover, Tooltip,
+    Alert, Badge, Button, Gallery, Modal, Popover, Tooltip,
     Card, CardTitle, CardActions, CardHeader, CardBody,
     DescriptionList, DescriptionListTerm, DescriptionListGroup, DescriptionListDescription,
     ExpandableSection,
     Flex, FlexItem,
+    Grid, GridItem,
     LabelGroup,
+    Page, PageSection, PageSectionVariants,
+    Progress, ProgressSize,
     Spinner,
     Stack, StackItem,
-    Page, PageSection, PageSectionVariants,
+    Switch,
     Text, TextContent, TextListItem, TextList, TextVariants,
 } from '@patternfly/react-core';
 
@@ -513,7 +516,7 @@ class RestartServices extends React.Component {
     }
 }
 
-const ApplyUpdates = ({ transaction, onCancel, allowCancel }) => {
+const ApplyUpdates = ({ transaction, onCancel, allowCancel, rebootAfter, setRebootAfter }) => {
     const [percentage, setPercentage] = useState(0);
     const [timeRemaining, setTimeRemaining] = useState(null);
     // chronological list of { status: PK_STATUS_*, package: "name version" } events that happen during applying updates
@@ -560,14 +563,10 @@ const ApplyUpdates = ({ transaction, onCancel, allowCancel }) => {
         });
     }, []); // eslint-disable-line react-hooks/exhaustive-deps
 
-    const cancelButton = (
-        <Button className={actions.length !== 0 && "progress-cancel"}
-                variant="secondary"
-                onClick={onCancel}
-                isDisabled={!allowCancel}>
-            {_("Cancel")}
-        </Button>
-    );
+    const cancelButton = allowCancel
+        ? <Button variant="secondary" onClick={onCancel} isSmall>{_("Cancel")}</Button>
+        : null;
+
     if (actions.length === 0) {
         return <EmptyStatePanel title={ _("Initializing...") }
                                 headingLevel="h5"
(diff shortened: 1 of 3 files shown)


martinpitt:
I fixed [this one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17500-20220626-175709-6cee657b-centos-8-stream/log.html#306-1) locally, just didn't push yet -- let's wait for the CSS fix.

garrett:
Mockup:

updates-reboot

garrett:
The button could still be set to disabled and we could have something like

```scss
.the-update-ui-thing-here .pf-c-button.pf-m-disabled {
   visibility: hidden;
}
```

(The names of the classes would probably be different, especially that first one. :wink:)

martinpitt:
Meh, this is really painful. I'm trying to use a Flex to put the progress bar and cancel button side by side, and the closest I got was this mess:

image

(yes, other ways of nesting them looked even *more* broken). ProgressBar isn't really designed to contain actions associated to it. If we want to do that, I think we'll have to dissect the React component and reimplement it ourselves.

garrett:
Either use a flex for just the progress bar and the cancel button, with `align-items: baseline` (there's a PF way of doing this too) and keep upgrade log out of it completely.

Or use grid for all the elements, but then the update log would have to span across the bottom cell.

I'd suggest using flex; it's simpler.

garrett:
I tried this, but haven't been able to test it yet:

```jsx
        <>
            <Flex className="progress-main-view">
                <FlexItem>
                    <div className="progress-description">
                        <Spinner isSVG size="sm" />
                        <strong>{PK_STATUS_STRINGS[lastAction.status] || PK_STATUS_STRINGS[PK.Enum.STATUS_UPDATE]}</strong>
                        &nbsp;{lastAction.package}
                    </div>
                    <Progress title={formatTimeRemaining} value={percentage} />
                </FlexItem>
                <FlexItem>{cancelButton}</FlexItem>
            </Flex>

            <Switch id="reboot-after" isChecked={rebootAfter}
                    label={ _("Reboot after completion") }
                    onChange={setRebootAfter} />

            <div className="update-log">
...
```

garrett:
Something like this:

```jsx
  <Flex flexWrap={{ default: 'nowrap' }} alignItems={{ default: 'alignItemsFlexEnd'}}>
    <FlexItem grow={{ default: 'grow' }}>Progress</FlexItem>
    <FlexItem>Cancel</FlexItem>
  </Flex>
```

martinpitt:
@garrett : I updated your two comments from above. Is that helpful in any way?

garrett:
The first suggestion is exactly the same as the second suggestion, BTW. The second one is abstracted and has some alignment and size attributes. The first drops the wrapper though, whereas the second one needs to be within it, I guess and is why it's going too wide in your screenshot?

martinpitt:
@garrett: The main difference is the `grow={{ default: 'grow' }}` bit, which causes the progress bar to be reasonably wide again, instead of being squeezed so narrow.

martinpitt:
This is an absolute nuisance to develop on right now, due to the broken page reload. That gets stuck forever on the "Initializing.." empty state, instead of properly re-attaching to a running upgrade. That's also quite a major regression, so I'm going to look at that one first.

martinpitt:
I sent PR #17517 for the issue above, and also included the general code cleanups, so that this one will become less noisy. I'd like to block on that, so that developing this change gets easy. Right now you'll have to restart the VM and re-upload the changed code for each iteration.

martinpitt:
@garrett : I tweaked the CSS for the cancel button and reboot switch to be center-aligned, and updated the screenshot.

martinpitt:
@garrett : Right; I've fought with the cancel button for two hours to try and get it to the right of the progress bar in a sensible manner. I got all kinds of "ridiculously broken" and then gave up and left it where it is.

Agreed about the update log -- that's the status quo (I didn't touch that), I am happy to look at that in a separate PR. It's not super-easy to do (that whole page is a big soup of flex and custom layout..).

garrett:
> Agreed about the update log -- that's the status quo (I didn't touch that), I am happy to look at that in a separate PR. It's not super-easy to do (that whole page is a big soup of flex and custom layout..).

This makes it sound like it's using nested flexes and should really be using a grid overall. (We'd need a grid object toplevel and then all major components at the immediate next-level.) With a lot of items, especially more than just in a line, grid is usually less fiddly to get what's wanted.

The order is really weird, wrong, and confusing with the cancel and the toggle button like this. :cry: Otherwise I'd say add it in an do some follow-ups. But it's just so wrong as-is that I think we shouldn't split it apart.

martinpitt:
With a grid, it looks like this now while cancelling is allowed:

image

and once it's not allowed, the button disappears:

image

Sorry for the imprecise image capturing -- the progress bar does *not* resize/move when the button disappears, this is just me not being able to select a constant snapshot frame.

martinpitt:
@garrett : I updated the commit message and "official" (relnote) screenshot in the description, and the three screenshots above to the current grid-based design. WDYT?

martinpitt:
[oh, right](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17500-20220712-103054-a3091b21-debian-testing/log.html#300) -- this is a race condition now. We would need to somehow postpone the downloading, or drop that check.

garrett:
> @garrett : I updated the commit message and "official" (relnote) screenshot in the description, and the three screenshots above to the current grid-based design. WDYT?

Oh, did you try what I suggested on IRC?

Use smaller progress bar, add sm margin to the bottom. This would not just thin the bar, but replace some of its thickness with space. This means that the cancel button would probably line up a little better.

It'd look like this:

image

It's something like this (making sure you have the sizing utility classes loaded):
```
 <Progress size={ProgressSize.sm} className="pf-u-mb-sm" />
```

https://www.patternfly.org/v4/utilities/spacing

martinpitt:
@garrett : Re-pushed with the implementation for the last screenshot (xs border, small progress and button). Thank you for your help!

marusak (approved):
Not doing design review, I see you went through a few iteration already :) Thanks!

PR #17501, opened 2022-06-24
test: Stop resizing the browser by default with TEST_SHOW_BROWSER

martinpitt:
Iterating page code and test development or doing screenshots  with
TEST_SHOW_BROWSER is rather unnerving, as most of the page isn't
visible/scrollable due to the fixed sizing. Stop this, and go back to
the old behavior where the browser uses its actual window size.

The only corner case is to watch the browser while doing assert_pixels()
development. Support that with `TEST_SHOW_BROWSER=pixels`.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -204,7 +204,7 @@ class Browser:
             schema = tls and "https" or "http"
             href = "%s://%s:%s%s" % (schema, self.address, self.port, href)
 
-        if not self.current_layout:
+        if not self.current_layout and os.environ.get("TEST_SHOW_BROWSER") in [None, "pixels"]:
             self.current_layout = self.layouts[0]
             size = self.current_layout["shell_size"]
             self._set_window_size(size[0], size[1])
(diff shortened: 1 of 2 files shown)


martinpitt:
I don't often touch the UI, but when I do, this drives me nuts...

test/common/testlib.py
```diff
@@ -204,7 +204,7 @@ def open(self, href, cookie=None, tls=False):
             schema = tls and "https" or "http"
             href = "%s://%s:%s%s" % (schema, self.address, self.port, href)
 
-        if not self.current_layout:
+        if not self.current_layout and os.environ.get("TEST_SHOW_BROWSER") in [None, "pixels"]:
```
mvollmer, line 207: This is just the initial size, calls to assert_pixels and set_layout will still resize the browser.  Is that acceptable?

mvollmer (approved):

test/common/testlib.py
martinpitt: That's debatable, I guess -- between YAFIYGI or just ignoring the assert_pixels() calls with TEST_SHOW_BROWSER, but I currently tend towards the former. At least it doesn't get in the way so much now.

PR #17506, opened 2022-06-27
containers/ws: Support all OpenSSH key types

martinpitt:
Load a provided SSH key into ssh-agent, and let cockpit-ssh use it from
there. This is the same approach that we use with "regular" cockpit and
connecting to remote hosts in the shell.

This drops the restriction that keys must be in the old RSA PEM format. SSH has
moved to the proprietary "OpenSSH" format a few years ago, but openssl (or
anything else other than SSH itself) cannot decrypt these any more.

Unfortunately ssh-add does not allow reading the password from stdin, so
construct and use an askpass agent (similar to Client).

Restrict the lifetime of keys in the agent to 30 seconds. That gives the login
process enough time to connect, but avoids keeping the unencrypted key in
memory indefinitely.

----

  - Builds on top of #17504
  - [x] Refresh official container to pick up the changes here
  - [x] Refresh fedora-coreos image: https://github.com/cockpit-project/bots/issues/3564
  - [x] Rebase, drop the XXX hack commit

## cockpit/ws container: Support for modern SSH keys

[Cockpit 272](https://cockpit-project.org/blog/cockpit-272.html) introduced [cockpit/ws container](https://quay.io/repository/cockpit/ws) support for authenticating with classic RSA PEM type SSH keys. This was improved to support all key types, in particular the "OpenSSH" format or EC types.

### DIFF
--- a/containers/ws/cockpit-auth-ssh-key
+++ b/containers/ws/cockpit-auth-ssh-key
@@ -134,16 +134,36 @@ def decode_basic_header(response):
     return user, password
 
 
-def send_decrypted_key(fname, password):
-    p = subprocess.run(["openssl", "rsa", "-in", fname, "-passin", "stdin"],
-                       check=False, capture_output=True, encoding="UTF-8",
-                       input=password)
+def load_key(fname, password):
+    # ssh-add has the annoying behavior that it re-asks without any limit if the password is wrong
+    # to mitigate, self-destruct to only allow one iteration
+    with open("/run/askpass", "w") as fd:
+        fd.write("""#!/bin/sh
+rm -f $0
+cat /run/password""")
+
+        os.fchmod(fd.fileno(), 0o755)
+
+    env = os.environ.copy()
+    env["SSH_ASKPASS_REQUIRE"] = "force"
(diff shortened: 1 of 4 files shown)


containers/ws/cockpit-auth-ssh-key
```diff
@@ -134,16 +134,36 @@ def decode_basic_header(response):
...
+rm -f $0
+cat /run/password""")
+
+        os.fchmod(fd.fileno(), 0o755)
+
+    env = os.environ.copy()
+    env["SSH_ASKPASS_REQUIRE"] = "force"
+    env["SSH_ASKPASS"] = "/run/askpass"
+
+    pass_fd = os.open("/run/password", os.O_CREAT | os.O_EXCL | os.O_WRONLY | os.O_CLOEXEC, mode=0o600)
```
martinpitt, line 151: I can't say I like this... I just don't have a better idea :cry:

martinpitt:
@allisonkarlitskaya : Friendly review ping? Thanks!

marusak (approved):
I cannot say that I have any strong options about this. For me look good and don't want to block this anymore, so thanks!

PR #17507, opened 2022-06-27
containers/ws: Fix pulling in packages from COPR

martinpitt:
Commit 76da7b7fc29 broke building the official container against COPR:
`dnf --installroot` only considers the repos in the target tree, not on
the host. So create cockpit.repo in /build.

---

I noticed that when I [synced the recent changes](https://github.com/cockpit-project/cockpit-container/commit/ff6df9118190b4aca436591d37fdea274c38c58e) to cockpit-container, and bumped the release to 272. I [added the fix](https://github.com/cockpit-project/cockpit-container/commit/92a0ec596c10ed186c1ce1b139a3b404c2e409ef)  to cockpit-containers directly, built the container, validated it, and pushed it to quay.

I didn't notice that in my earlier testing as I was building the container against local rpms. Sorry for the blunder!


mvollmer (approved):

mvollmer:
Makes sense to me, but I am only a tourist in this part of town.

PR #17509, opened 2022-06-28
containers/ws: Accept $container == podman, test ws install script, pycodestyle fix

martinpitt:
This changed on the most recent Fedora CoreOS refresh [1] and thus breaks the install script.
    
[1] https://github.com/cockpit-project/bots/pull/3562

After this lands, refresh the official container, then refresh the image.

### DIFF
--- a/containers/ws/label-install
+++ b/containers/ws/label-install
@@ -44,7 +44,7 @@ chown root:root /host/etc/cockpit/ws-certs.d /host/etc/cockpit/machines.d
 mkdir -p /etc/ssh
 
 # For podman, generate a systemd unit for starting on boot
-if [ "${container:-}" = oci ] && [ -n "$IMAGE" ] && [ ! -e /host/etc/systemd/system/cockpit.service ]; then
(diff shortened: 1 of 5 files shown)


containers/ws/label-install
```diff
@@ -44,7 +44,7 @@ chown root:root /host/etc/cockpit/ws-certs.d /host/etc/cockpit/machines.d
 mkdir -p /etc/ssh
 
 # For podman, generate a systemd unit for starting on boot
-if [ "${container:-}" = oci ] && [ -n "$IMAGE" ] && [ ! -e /host/etc/systemd/system/cockpit.service ]; then
+if [ "${container:-}" = oci -o "${container:-}" = podman ] && [ -n "$IMAGE" ] && [ ! -e /host/etc/systemd/system/cockpit.service ]; then
```
mvollmer, line 47: The old code uses `&&` for logic, should the new code maybe use `||`?

mvollmer (approved):

containers/ws/label-install
martinpitt: I prefer `||` as well, but we would need to use some () grouping then, which is syntactically awkard (or maybe even impossible -- this might then need some nested `if`s rather?)

PR #17513, opened 2022-06-29
Stylelint squash

garrett:
Exactly the same as #17408, except with related rules squashed into one commit each for easier reading. Rules are listed either as the title or description (for combined rules).

There are now less than half of the number of original commits. It should be easier to review.


pkg/static/login.css
```diff
@@ -28,8 +26,6 @@
 
 html {
   font-family: sans-serif;
-  -ms-text-size-adjust: 100%;
-  -webkit-text-size-adjust: 100%;
```
jelly, line 32: This does not seem to be supported yet? https://caniuse.com/mdn-css_properties_text-size-adjust_percentages

Never worked on Firefox either which looks weird.

.stylelintrc.json
```diff
@@ -1,97 +1,38 @@
 {
   "extends": "stylelint-config-standard-scss",
   "rules": {
```
jelly, line 3: Should we include a comment/description somewhere of the three blocks, I think that makes sense.

jelly (changes requested):

pkg/static/login.css
garrett: https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust:

> The text-size-adjust [CSS](https://developer.mozilla.org/en-US/docs/Web/API/CSS) property controls the text inflation algorithm used on some smartphones and tablets. Other browsers will ignore this property.

So MS would be IE. WebKit would be Safari and/or older Chrome.

The next line is unprefixed and preserved:

>  text-size-adjust: 100%;

But it only looks like it works in Edge from 2015+ and Chrome 2016+.

So I think it's fine. There would be no changes as this is mobile only and Firefox and Safari don't care. Only Chrome does and even ancient Android phones get newer browsers... according to caniuse.

HOWEVER, that's wrong.

MDN states that Firefox on Android and Safari on iOS also support this, but are both prefixed only.

https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust#browser_compatibility

(Click the legend. The -X- icon means it's prefixed still.)

Firefox on Android uses the -moz- prefix but also supports the -webkit- prefix.

Therefore, I think we should probably partially revert this and include the -webkit- prefix and add an ignore for prefixes.

This is one of those rare places where a prefix still matters.

.stylelintrc.json
garrett: No, we can't. I tried. I'd like it. But I tried several hacky ways. It's impossible.

JSON does not support comments. And Stylelint doesn't support arbitrary bogus rules.

pkg/static/login.css
jelly: Ack, -ms can go everywhere, my comment was aimed at Safari/Firefox

garrett: OK, after trying to get phones connected to the VM for quite some time today, I finally was able to debug it all and test it out... and...

It looks like this CSS does absolutely nothing for the front page, so I guess we should just drop it completely.

garrett: OK, I completely removed it in the CSS, as it didn't even do anything on any browser, including mobile ones.

garrett:
@jelle: I addressed your concerns with da3d3b3, which removes some unnecessary CSS in the login page that both you and stylelint (with some stricter rules) didn't like.

jelly:
Looks good! I'll make a note to also add stylelint to cockpit-podman,machines and starterkit.

PR #17516, opened 2022-07-01
packagekit: Fix "Restart services" button priority

### LINKED ISSUE #17514: software updates: Services button should be secondary, not primary
### Explain what happens

You're not supposed to have two primary buttons for the same related section

image

The services button should be secondary.

(Reboot being primary is correct; the ignore being a link styled button is also correct.)

martinpitt:
You're not supposed to have two primary buttons for the same related
section.  panel was missing the case when the reboot button was shown
not because of packages requesting a reboot, but because of services
that need a manual restart.

Fixes #17514

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -728,6 +728,8 @@ const UpdateSuccess = ({ onIgnore, openServiceRestartDialog, openRebootDialog, r
         });
     }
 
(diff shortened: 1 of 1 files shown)


garrett:
Thanks for fixing this! (I can't seem to test it now though. I've tried. Happy to see the fix incoming though!)

jelly (approved):

PR #17526, opened 2022-07-05
Add optional static typing for testlib

jelly:
This makes LSP work nicely in our tests:

image

Downside is that we have some work todo:

image

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1385,7 +1385,6 @@ class MachineCase(unittest.TestCase):
         if opts.address and self.provision is not None:
             raise unittest.SkipTest("Cannot provision multiple machines if a specific machine address is specified")
 
-        self.browser: Optional[Browser] = None
         self.machines = {}
         provision = self.provision or {'machine1': {}}
         self.tmpdir = tempfile.mkdtemp()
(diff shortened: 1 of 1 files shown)


test/common/testlib.py
```diff
@@ -300,15 +302,15 @@ def eval_js(self, code, no_trace=False):
             print("eval_js(%s): cannot interpret return value %s" % (code, result))
         return None
 
-    def call_js_func(self, func, *args):
+    def call_js_func(self, func: str, *args: Any) -> Optional[Any]:
```
martinpitt, line 305: Interesting, "Any" does not include `None`, i.e. this still needs to be wrapped into "Optional"?

test/common/testlib.py
```diff
@@ -1231,10 +1231,10 @@ class MachineCase(unittest.TestCase):
     image = testvm.DEFAULT_IMAGE
     libexecdir = None
     runner = None
-    machine = None
+    machine: testvm.Machine
```
martinpitt, line 1234: This is a curious syntax -- so it exists purely to declare the type now, but does not actually "create" the member?

martinpitt (approved):
Nice, thank you! This cries for adding mypy or pyright to CI now, do you want to work on that?

test/common/testlib.py
jelly: Good point! It seems I was wrong! https://docs.python.org/3/library/typing.html#the-any-type

> Nice, thank you! This cries for adding mypy or pyright to CI now, do you want to work on that?

I want to add mypy yes, but it's a bit tricky to run:

```
[jelle@t14s][~/projects/cockpit/statictyping]%MYPYPATH=bots/machine:test/common mypy --config-file mypy.ini ./test/verify/check-shell-menu
test/verify/check-shell-menu:26: error: Name "nondestructive" is not defined
test/verify/check-shell-menu:27: error: Name "skipDistroPackage" is not defined
test/verify/check-shell-menu:28: error: Name "MachineCase" is not defined
test/verify/check-shell-menu:30: error: Name "enableAxe" is not defined
test/verify/check-shell-menu:125: error: Name "test_main" is not defined
Found 5 errors in 1 file (checked 1 source file)
```

Fails pretty hard still, all star imports need to be fixed.

```
[jelle@t14s][~/projects/cockpit/statictyping]%MYPYPATH=bots/machine:test/common mypy --config-file mypy.ini ./test/common/testlib.py
test/common/testlib.py:1124: error: Argument "scroll_into_view" to "assert_pixels_in_current_layout" of "Browser" has incompatible type "Optional[str]"; expected "Optional[bool]"
test/common/testlib.py:1882: error: Missing positional argument "self" in call to "items" of "dict"
test/common/testlib.py:1883: error: "_VT" has no attribute "ssh_reachable"
test/common/testlib.py:1884: error: "_VT" has no attribute "label"
test/common/testlib.py:1889: error: "_VT" has no attribute "download_dir"
```

Seems I have to fix the first one, as it was introduced in the PR.

martinpitt (approved):
Cheers!

PR #17535, opened 2022-07-08
storaged: add Arch Linux nfs-utils/stratisd packages

jelly:
Make it possible to install optional NFS/Stratis support using
PackageKit.

### DIFF
--- a/test/verify/packagelib.py
+++ b/test/verify/packagelib.py
@@ -323,7 +323,7 @@ post_upgrade() {{
         if postinst:
             cmd += f"rm /tmp/{name}.install"
         self.machine.execute(cmd)
-        self.addCleanup(self.machine.execute, f"pacman -R --noconfirm {name} 2>/dev/null || true")
+        self.addCleanup(self.machine.execute, f"pacman -Rdd --noconfirm {name} 2>/dev/null || true")
 
     def createAptChangelogs(self):
         # apt metadata has no formal field for bugs/CVEs, they are parsed from the changelog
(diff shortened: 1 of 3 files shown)


mvollmer:
The TestStoragePackagesNFS.testNfsMissingPackages is skipped in arch: "TODO: bug in refreshing removed/installed state on Arch Linux".  So does this work at all?  Would be great to get the test working as well.

jelly:
> The TestStoragePackagesNFS.testNfsMissingPackages is skipped in arch: "TODO: bug in refreshing removed/installed state on Arch Linux". So does this work at all? Would be great to get the test working as well.

Excellent idea, enabled the test :eye:

jelly:
This worked fine when running just the test, so it seems that something is getting in the way

jelly:
So pkcon can install the package fine, so it seems to be something with the dbus api which is broked?

test/verify/packagelib.py
```diff
@@ -71,6 +71,7 @@ def setUp(self):
             self.restore_file("/etc/pacman.d/mirrorlist")
             self.restore_file("/usr/share/libalpm/hooks/90-packagekit-refresh.hook")
             self.machine.execute("rm /etc/pacman.conf /etc/pacman.d/mirrorlist /var/lib/pacman/sync/* /usr/share/libalpm/hooks/90-packagekit-refresh.hook")
+            self.machine.execute("test -d /var/lib/PackageKit/alpm && rm -r /var/lib/PackageKit/alpm || true")  # Drop alpm state directory as it interferes with running offline
```
marusak, line 74: Do we need to test for it when you do `|| true` anyway?

jelly: As when it does not exists test returns "false" which makes execute fail.

marusak (approved):
Thanks!

PR #17537, opened 2022-07-08
.github: run static code checks

jelly:
Shockingly ./test/static-code does not run succesfully at the moment. In
b746f5aef367144695 the way we run tests changed from a bash script to an
Python script which does not all ./test/static-code anymore.


martinpitt:
The reason is that `make distcheck` runs the tests from a tarball, not from a git clone any more. Thus `find_scripts()` in test/static-code does not work and that whole thing is silently skipped. And `make check-memory` skips the static code tests (no point in running them under valgrind). So indeed test/static-code should be run as a separate scenario, or perhaps `make check` (which takes a bit longer, but no worries..)

martinpitt (approved):
Thanks!

PR #17546, opened 2022-07-12
shell: update PF link to link to it's redirect

### LINKED ISSUE #17545: Validate all URLs
Checked 154 URLs out of which 1 is/are invalid:
    https://github.com/patternfly/patternfly/issues/4541

jelly:
Closes #17545

### DIFF
--- a/pkg/shell/nav.scss
+++ b/pkg/shell/nav.scss
@@ -31,7 +31,7 @@ $desktop: $phone + 1px;
         background: var(--pf-global--BackgroundColor--dark-400);
     }
 
-    // https://github.com/patternfly/patternfly/issues/4541
+    // https://github.com/patternfly/patternfly-design/issues/1112
     .search {
         background: transparent !important;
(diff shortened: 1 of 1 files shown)


marusak (approved):
Thanks!
Although should not be necessary as per https://github.com/cockpit-project/cockpit/blob/main/tools/urls-check#L41

Also I think it was just github issue as otherwise the report would be "Following url is redirected", see https://github.com/cockpit-project/cockpit/blob/main/tools/urls-check#L127

But this change is correct in any case :)

jelly:
> Thanks! Although should not be necessary as per https://github.com/cockpit-project/cockpit/blob/main/tools/urls-check#L41
> 
> Also I think it was just github issue as otherwise the report would be "Following url is redirected", see https://github.com/cockpit-project/cockpit/blob/main/tools/urls-check#L127
> 
> But this change is correct in any case :)

Isn't this the second tme the script was broken?

PR #17544, opened 2022-07-11
kdump: Refactor config handling

skazi0:
Config handling was modified to decouple UI from config file format.
All of the platform-specific code was moved to config-client.js and
"abstract" settings model was added as an interface.
In addition, NFS UI page was modified to not require NFS mounts in
specific format but as separate "server" and "export" fields.

PR #17547, opened 2022-07-12
fmf: Fix syntax error in node_modules checkout

martinpitt:
Introduced in commit 70de12208c4

----

Should fix [this failure](https://artifacts.dev.testing-farm.io/8c35e8cb-73e0-4894-b642-96f153b322bc/), seen in PR #17544.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -22,7 +22,7 @@ fi
 git clone --depth=1 https://github.com/cockpit-project/bots
 
 # support running from clean git tree; this doesn't work from release tarballs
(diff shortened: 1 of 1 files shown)


jelly (approved):

martinpitt:
Meh, there is *still* CDP errors on [rawhide](https://artifacts.dev.testing-farm.io/fa11be8a-c0ca-4b0e-bc62-07533d508a9a/). There's this pre-existing node_modules/ directory which is totally unaccounted for. Well, I'll look at that in a follow-up.

PR #17549, opened 2022-07-12
test: replace pycodestyle and pyflakes with flake8

jelly:
flake8 is a wrapper over pyflakes and pycodestyle, and most importantly
supports ignoring specific lines with `noqa`. This allows us to drop our
custom grep filters. The last remaning grep is for inotify which is a
special case as it used by combined with other scripts and therefore
can't be imported.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -42,7 +42,7 @@ For running integration tests, the following dependencies are required:
 
     $ sudo dnf install curl expect xz rpm-build chromium-headless \
         libvirt-daemon-driver-storage-core libvirt-daemon-driver-qemu libvirt-client python3-libvirt \
-        python3-pyflakes python3-pycodestyle python3-pyyaml
+        python3-flake8 python3-pyyaml
 
 Creating VM images locally (not necessary for running tests) needs the
 following:
--- a/containers/unit-tests/setup.sh
+++ b/containers/unit-tests/setup.sh
@@ -11,6 +11,7 @@ dependencies="\
     curl \
     dbus \
     firefox-esr \
+    flake8 \
     gcc-multilib \
     gdb \
     git \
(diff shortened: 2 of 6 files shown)


martinpitt:
This looks great, thank you! I [triggered a container refresh](https://github.com/cockpit-project/cockpit/actions/runs/2657032344) against this branch. Let's please wait for that to succeed, as the most recent automatic refresh [failed](https://github.com/cockpit-project/cockpit/actions/workflows/unit-tests-refresh.yml).

martinpitt (approved):

martinpitt:
Dang, it's [hanging again](https://github.com/cockpit-project/cockpit/runs/7303271689?check_suite_focus=true). We need to fix that first before we can land this. Apparently something changed the unit tests last week without updating the refresh workflow? As the container builds *did* work here, I don't believe it's a regression in Debian itself.

jelly:
> Dang, it's [hanging again](https://github.com/cockpit-project/cockpit/runs/7303271689?check_suite_focus=true). We need to fix that first before we can land this. Apparently something changed the unit tests last week without updating the refresh workflow? As the container builds _did_ work here, I don't believe it's a regression in Debian itself.

Strange locally it worked.

martinpitt:
Notes:
 - refresh runs on "ubuntu-latest", while unit-tests.yml runs on "ubuntu-20.04"; however, "latest" == "20.04" at the moment, so that's not ot.
 - both workflows use exactly the same commands to build the container and run the scenarios, including podman (not docker) and rootless.

So ATM I don't have an idea why one works but not the other. :crying_cat_face: 

Unrelated:
 - Let's add per-step `timeout-minutes:` to avoid the 3 hour hang
 - Let's add the two new scenarios (pycheck and check)

martinpitt:
I'm doing some experiments on https://github.com/cockpit-project/cockpit/compare/fix-unit-test-refresh , see [action runs](https://github.com/cockpit-project/cockpit/actions/workflows/unit-tests-refresh.yml)

PR #17551, opened 2022-07-13
Fix unit-test container refresh, fix valgrind invocation for HTML tests

martinpitt:
Refreshing our unit-tests container has recently [started to fail](https://github.com/cockpit-project/cockpit/actions/runs/2646144001) both on the scheduled runs on main, and in PR #17549. See [this comment](https://github.com/cockpit-project/cockpit/pull/17549#issuecomment-1182741018) for details of the debugging.

I [triggered a run](https://github.com/cockpit-project/cockpit/actions/runs/2661693623) against this branch, but with an extra commit that avoids pushing the container to the registry -- for now I just want to see if the tests work. And they did! :partying_face:

### DIFF
--- a/.github/workflows/unit-tests-refresh.yml
+++ b/.github/workflows/unit-tests-refresh.yml
@@ -12,7 +12,7 @@ jobs:
     permissions:
       contents: read
       packages: write
-    timeout-minutes: 180
+    timeout-minutes: 60
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/unit-tests-refresh.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   # we do both builds and all tests in a single run, so that we only upload the containers on success
   refresh:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
martinpitt, line 11: Note that this is not actually necessary to fix the container refresh. But as it works, it can't hurt anyway to jump to 22.04 now -- it'll be around for a while.

jelly (approved):

PR #17552, opened 2022-07-13
test: enable more packagekit tests on Arch Linux

jelly:
After the commit which drops the alpm directory state in
/var/lib/PackageKit these issues should be resolved.


jelly:
After a few runs, we get:
```
checking package integrity...
loading package files...
checking for file conflicts...
error: failed to commit transaction (conflicting files)
slow: /stamp-slow-1-1 exists in filesystem
Errors occurred, no packages were upgraded.
```

jelly:
So this only enables one more test..... but I think the makepkg cleanups are good to have.

martinpitt (approved):
Thank you!

PR #17553, opened 2022-07-13
client: Fix crash on absent Handy.StyleManager

martinpitt:
Current RHEL 9 has libhandy 1.2.3, which does not yet have the
`StyleManager` API. This causes client to crash with an AttributeError
since commit 9d9b1f8e5c04.

https://bugzilla.redhat.com/show_bug.cgi?id=2106096

---

See bz for how I tested this.


marusak (approved):
Thanks!

PR #17557, opened 2022-07-13
CSS: Rework custom breakpoint CSS

### LINKED ISSUE #17556: _variables.scss is imported multiple times (5 times!)
### Explain what happens

1. Browser to https://127.0.0.1:9090/users
2. Open developer tools
3. Click inspect element and select the `body` of the users page
4. Look at the `styles` imported and notice that `:root { }` is imported multiple times from `_variables.scss`

What seems to happen is:

`pkg/lib/users.scss` imports `@import "global-variables.scss";` which imports `@import "@patternfly/patternfly/base/patternfly-variables.scss";` to override the breakpoints.

variables.scss is imported multiple times, probably the cause of our issue.
```
[jelle@t14s][~/projects/cockpit/dark-theme]%git grep variables.scss
pkg/Makefile.am:        pkg/lib/patternfly/patternfly-overrides-variables.scss \
pkg/lib/_global-variables.scss:@import "@patternfly/patternfly/base/patternfly-variables.scss";
pkg/lib/cockpit-components-table.scss:@use "@patternfly/patternfly/base/patternfly-variables.scss";
pkg/lib/ct-card.scss:@use "_global-variables.scss" as *;
pkg/systemd/logs.scss:@import "global-variables.scss";
pkg/systemd/services/service-details.scss:@use "_global-variables.scss";
pkg/users/users.scss:@import "global-variables.scss";
pkg/users/users.scss:// @import "@patternfly/patternfly/sass-utilities/scss-variables.scss";
webpack.config.js:                // [Redefine grid breakpoints] section in pkg/lib/_global-variables.scss for more details
```

```
[jelle@t14s][~/projects/cockpit/dark-theme]%git grep patternfly-variables
pkg/lib/_global-variables.scss:@import "@patternfly/patternfly/base/patternfly-variables.scss";
pkg/lib/cockpit-components-table.scss:@use "@patternfly/patternfly/base/patternfly-variables.scss";
```

### Version of Cockpit

_No response_

_No response_

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

garrett:
This might fix #17556. Or at least address it somewhat?


garrett:
I tried removing the breakpoint inclusions everywhere, but it turns out whenever a breakpoint is used, the file needs to be included. And whenever a PF component is imported in the CSS, it _also_ needs the breakpoints file included.

If we do move the PF component includes into the JSX, I think we'll need to also include the breakpoints overriding before the component inclusion.

pkg/lib/cockpit-components-table.scss
```diff
@@ -1,5 +1,4 @@
-@use "@patternfly/patternfly/base/patternfly-variables.scss";
-@import "global-variables";
+@use "global-variables" as *;
```
jelly: So the solution here is that `@use` namespaces the import to this file right?

jelly (approved):
Looks good to me, haven't verified the branch yet locally

pkg/lib/cockpit-components-table.scss
garrett: No, the solution is to not import patternfly variables. (Removing the first line.)

Use is more modern and generally better than import for SASS, but it _probably_ doesn't really matter in this PR.

jelly:
With my current main:

image

With this PR:
image

garrett:
All the places where we import from PatternFly in our CSS files:

image

(The selected one happens to be different in the file with an unecessary `../../node_modules/` in it that I removed locally.)

garrett:
Screenshot from 2022-07-14 11-45-55

garrett:
webpack's null-loader (which is deprecated, BTW), prevents loading those files... so they can be loaded by the SCSS build stack instead of webpack, to prevent duplication.

What would be ideal is to get webpack to load the overridden breakpoints while it is telling SASS to compile the SCSS for the components, so we don't have to jump through hoops and can just use PF React components in JSX directly, with correct breakpoints as well.

However, that still won't address the static PF CSS.

garrett:
We have 2 main problems: 
1. React and SASS compilation no nothing about each other.  
2. Both of these compile steps, `import` (JSX) and `@import` (SASS) copy the CSS, regardless of whether import has been used before (even in other parts of included SCSS files elsewhere on the page). WebPack _might_ de-duplicate JSX `import` (this needs to be tested), but it definitely does not touch SASS `@import`.

Both of these cause multiple copies of CSS to happen.

Redundant CSS is:
- bad for bloat (takes longer to send, parse, etc.),
- hurts debugging
- messes up specificity (meaning CSS won't be applied as expected, and this is especially a problem for dark mode)

garrett:
Rebased. Although, I think we need to rework our build system and rewrite the breakpoints at that level (by changing the magic pixel values.... and drop our custom stuff).

garrett:
I think this attempt is no longer needed, thanks to the esbuild refactor: https://github.com/cockpit-project/cockpit/pull/18447

PR #17561, opened 2022-07-15
test: add pixel tests for our custom progress implementation

jelly:
For stratis if you have multiple filesystems we show the space one
filesystem takes up relative to the other. As this uses our own written
usagebar let's make sure PF doesn't break it on updates.


jelly:
> I wonder how predictable the free space is going to be, but we can keep it as-is for now. We can still add an ignore selector if it gets in the way at some point (most likely during an image refresh when kernel/stratis change). Thanks, good idea!

We already have a pixel test for the case where there is one filesystem mounted, so I hope/think this should be fine with two filesystems.

martinpitt (approved):
Cheers!

PR #17564, opened 2022-07-16
ws: Fix test-channelresponse for glib dropping global child watch

### LINKED ISSUE #17563: Rawhide builds hang in test-channelresponse
### Explain what happens

A few days ago, our builds in rawhide started to hang and time out after 5 hours, e.g. in [this run](https://copr.fedorainfracloud.org/coprs/packit/cockpit-project-cockpit-17441/build/4635836/).

This is reproducible in `toolbox create -r 37` and our standard
```
tools/webpack-jumpstart
./autogen.sh rpm
make
make check
```

In particular, it is `./test-channelresponse` which hangs in `/web-channel/resource/failure`; everything else works.

### Version of Cockpit

272 / main

_No response_

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

martinpitt:
glib 2.73.2 dropped the global SIGCHLD handler [1]. That causes the 
killed bridge process in test_resource_failure() to end up as a zombie,
so that `kill (pid, 0)` succeeds forever and the test hangs. So with
that version we need to wait() the bridge process ourself.

However, the previous comment is still true for older glib versions:
wait()ing the process there still does interfere with the global
handler:

    GLib-WARNING: GChildWatchSource: Exit status of a child process was requested but ECHILD was received by waitpid().

To avoid the (fatal) warning, conditionalize the waiting on new glib
versions only.

Fixes #17563

[1] https://gitlab.gnome.org/GNOME/glib/-/commit/f615eef4bafaa2f

### DIFF
--- a/src/ws/test-channelresponse.c
+++ b/src/ws/test-channelresponse.c
@@ -520,7 +520,7 @@ test_resource_failure (TestResourceCase *tc,
   gconstpointer str;
   GPid pid;
   const gchar *expected = "HTTP/1.1 500 terminated\r\nContent-Type: text/html; charset=utf8\r\nTransfer-Encoding: chunked\r\n" STATIC_HEADERS "\r\n13\r\n<html><head><title>\r\na\r\nterminated\r\n15\r\n</title></head><body>\r\na\r\nterminated\r\nf\r\n</body></html>\n\r\n0\r\n\r\n";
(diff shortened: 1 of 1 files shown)


src/ws/test-channelresponse.c
```diff
@@ -529,9 +529,15 @@ test_resource_failure (TestResourceCase *tc,
   g_assert (cockpit_pipe_get_pid (tc->pipe, &pid));
   g_assert_cmpint (pid, >, 0);
   g_assert_cmpint (kill (pid, SIGTERM), ==, 0);
-  /* Wait until it's gone; we can't use waitpid(), it interferes with GChildWatch */
+  /* Wait until it's gone */
+#if GLIB_CHECK_VERSION(2,73,2)
```
martinpitt, line 533: FTR: For a production binary that would not be good enough, as it's possible to build a binary against the old glib, and then upgrade at runtime to the new one. But for a unit test it's good enough IMHO. Maybe @allisonkarlitskaya has an idea how to make this work with both versions at runtime, but this is urgent enough to land a quick fix now.

jelly (approved):

PR #17572, opened 2022-07-18
test: Avoid uploading to ~user directories

martinpitt:
scp is deprecated, and in Fedora 36 you can't scp something to a ~user
directory any more:

    scp: server expand-path extension is required for ~user paths in SFTP mode

Specify an absolute path instead.

### DIFF
--- a/test/verify/check-examples
+++ b/test/verify/check-examples
@@ -20,7 +20,7 @@ class TestPinger(MachineCase):
         super().setUp()
         self.restore_dir("/home/admin")
         self.machine.execute("mkdir -p ~admin/.local/share/cockpit")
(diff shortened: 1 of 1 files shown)


martinpitt:
After that, I'll look into replacing `scp` with `sftp` or possibly `rsync`, but this is a good first step to fix our tests when running on F36.

ptoscano:
Note that, by default (even when `scp` used the scp protocol instead of sftp), CWD is the user's home, you can use paths relative to it, e.g.:
```
$ scp *.html example.com:public_html/
```
(I've been doing that for many years)

martinpitt:
@ptoscano I know, but we ssh to root@, so that doesn't help us here.

jelly (approved):

PR #17573, opened 2022-07-19
test: Make venvs happy

skazi0:
When running `test/image-prepare` using python virtualenv, some scripts were still using system-wide python3.


martinpitt (approved):
Heh, okay! Be warned, nobody else in the team uses venv (we prefer toolbox and real distro packages), so we probably break this again at some point..

PR #17575, opened 2022-07-19
lib: Set id of PasswordFormFields FormGroup

skobyda:
Setting id of a whole FormGroup will ease referencing it in the tests.


skobyda:
Example of how I plan to use id of this Formgroup in test can be found here: https://github.com/cockpit-project/cockpit/pull/17575

martinpitt (approved):
Thanks!

PR #17578, opened 2022-07-20
Flake8 improvements

jelly:


.flake8
```diff
@@ -0,0 +1,14 @@
...
+    # line break after binary operator
+    W504
+builtins =
+    # inotify.py get pasted together with other files
+    Inotify,
+    IN_CREATE, IN_MOVED_TO,
+    IN_MOVED_FROM, IN_MOVE_SELF,
+    IN_DELETE, IN_DELETE_SELF,
+    IN_CLOSE_WRITE,
+    IN_IGNORED,
```
jelly, line 12: Ideally this would be file specific, but not sure if it can be done.

martinpitt (approved):
Whee, this is nice, thank you! Tests are still running, but good if/when green.

PR #17579, opened 2022-07-20
lib: drop duplicate patternfly-variables include

jelly:
We already include the breakpoints via "global-variables" so due to the
nature of SASS we end up with duplicated --pf-global--font-path
declaration in :root {}. Before this commit:

$ grep -cr "\-\-pf-global--font-path" dist/networkmanager/*.css
dist/networkmanager/firewall.css:2

After:
$ grep -cr "\-\-pf-global--font-path" dist/networkmanager/*.css
dist/networkmanager/firewall.css:1

See also for more context:
https://github.com/cockpit-project/cockpit/issues/17556#issuecomment-1184518556

Let's hope the pixel tests agree this is a good idea.

### DIFF
--- a/pkg/lib/cockpit-components-table.scss
+++ b/pkg/lib/cockpit-components-table.scss
@@ -1,4 +1,3 @@
-@use "@patternfly/patternfly/base/patternfly-variables.scss";
 @import "global-variables";
 @import "@patternfly/patternfly/components/Table/table.scss";
 @import "@patternfly/patternfly/components/Table/table-grid.scss";


martinpitt (approved):
LGTM (and straightforward) from my end, but leaving the final review to @garrett.

garrett (approved):
@martinpitt: Finding and removing this line actually came from me originally, so I'm a bit biased. :wink: 

But it seemed to remove a lot of duplication in :root {} for the various CSS variables PF uses. We need to remove the duplication for dark mode to work (and it's generally a good idea to remove CSS duplication anyway, so we're not wrestling with them in general).

PR #17582, opened 2022-07-21
test: Adjust for ubuntu-stable moving to 22.10 "kinetic"

martinpitt:
Ubuntu 21.10 was the last supported image which still had pam_tally2, so
simplify testFaillock() to only support pam_faillock.

Drop the obsolete tang workaround.

----

Needs to land in lockstep with https://github.com/cockpit-project/bots/pull/3641


martinpitt:
I can't make hands or tail of that, I filed https://bugs.launchpad.net/ubuntu/+source/openssh/+bug/1982482 for now. Let's see if I get a response soon.

martinpitt:
D'oh, got it! Fixed. Wash, rinse, repeat.

marusak (approved):
Thanks!

PR #17583, opened 2022-07-21
packit: Drop obsolete FMF plan bumping

martinpitt:
tmt got fixed to be able to run the tests from the dist-git tarball
instead of checking it out from git, and this PR enabled that:
https://src.fedoraproject.org/rpms/cockpit/pull-request/76


marusak (approved):
great!

PR #17584, opened 2022-07-21
systemd: Don't call updateTime() for every property change ever

mvollmer:
A D-Bus signal subscription will get all signals from all services,
not just from the service that the client was created for.

### DIFF
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -322,6 +322,7 @@ class ServicesPageBody extends React.Component {
         });
 
         this.timedated_subscription = timedate_client.subscribe({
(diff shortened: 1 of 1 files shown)


mvollmer:
I found this while hacking on LVM2 RIAD support.  Running "lvcreate --test ..." unfortunately triggers a bunch of udev events, which cause a lot of pointless systemd activity, which caused our Services page to spawn "cat /proc/uptime" and "date +%s" a couple hundred times.

martinpitt (approved):
Ouch! Well spotted.

mvollmer:
I have tested this by running "timedatectl set-ntp true".  This still triggers updateTime().  Other signals don't anymore.

PR #17585, opened 2022-07-22
bridge: Assert valid user in cockpit_peer_handle()

martinpitt:
main() sets $USER and asserts that it succeeds, so we can rely on $USER
here.

This quiesces Coverity:

    dereference: Dereferencing a pointer that might be "NULL" "user" when calling "cockpit_hex_encode".


marusak (approved):
Thanks!

PR #17592, opened 2022-07-25
Add Minutely and Custom options to specific time trigger for creating timers

dphaas2004:
Added minutely and Custom option to the specific time trigger in the CreateTimer Dialog per conversation on issue #9467 

Also fixed error from UserInputs.js TextInput not having proper id or aria-labels.


marusak:
Nice, thank you for this PR!

A few comments/ideas that I have noticed when I tried this one out.
- When I open the dialog and any time I switch to 'Don't repeat' the dropdown is opened. That is broken. Very likely not related to this PR, I will try current main and if it happens there as well I will send separate fix. Just wanted to mention it in case it got broken here:
Screenshot from 2022-07-26 09-42-32

- `Minutely` is all good! It needs test though. See for example here: https://github.com/cockpit-project/cockpit/blob/main/test/verify/check-system-services#L1149

- `Custom` is not clear to me what is expected as input. Can we have there some (?) icon with tooltip explaining what is expected as input? And also, it would need tests the same way.

marusak:
> When I open the dialog and any time I switch to 'Don't repeat' the dropdown is opened. 

Happens on main, don't worry about this comment then.

marusak:
> > When I open the dialog and any time I switch to 'Don't repeat' the dropdown is opened.

> Happens on main, don't worry about this comment then.


Sent https://github.com/cockpit-project/cockpit/pull/17595

dphaas2004:
Hi.  I noticed that issue too.  kind of annoying.  Anyway I have added the integration tests and a popover as suggested.  not sure if its alright to have a link in the popover or not but I think it helps clarify the custom field.  thoughts?

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -233,27 +248,58 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                                <FormSelectOption value="custom" label={_("Custom")} />
                             </FormSelect>
                         </FormGroup>
                         {repeat == "no" &&
                         <FormGroup label={_("Run at")}>
                             <TimePicker className="create-timer-time-picker specific-no-repeat"
                                         menuAppendTo={() => document.body} time={specificTime} is24Hour onChange={setSpecificTime} />
                         </FormGroup>}
+                        {repeat == "custom" &&
+                        <FormGroup
```
marusak, line 266: You can use `labelIcon` property here, see for example https://www.patternfly.org/v4/components/form

Right now it somehow breaks:
Screenshot from 2022-07-27 10-19-03

garrett (approved):
Changes look good.

---

I also see the dropdown issue. I'll be glad when that's fixed.

I think this is outside the scope of this PR as well, but, since we're here: if specific time is more commonly used _and_ is default, why is it second? We should reorder.

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -233,27 +248,58 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                            label={_("Custom")}
+                            validated={submitted && validateCustom ? "error" : "default"}
+                            helperTextInvalid={_("Invalid OnCalendar Event")}>
```
marusak, line 269: `Invalid OnCalendar event`. No title case please

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -233,27 +248,58 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                            <Flex>
+                                <TextInput
+                                    id="custom"
+                                    className='custom-group'
+                                    value={custom}
+                                    onChange={setCustom}
+                                    validated={submitted && validateCustom ? "error" : "default"}
+                                />
+                                <Popover key="popover-unix-group-custom" id="popover-unix-group-custom" position={ PopoverPosition.right }
+                                        headerContent={<div>Custom Timer Details</div>}
```
marusak, line 279: `Custom timer details`, no sentance case. Also it needs to be localized, so writing it as `headerContent={_("Custom timer details")}.
Also not sure if that is the best title or if the title is actually required.
Screenshot from 2022-07-27 10-23-16

@garrett WDYT of the text?

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -233,27 +248,58 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                                        bodyContent={
+                                            <div>
+                                                <p>Any systemd calendar event.</p>
+                                                <p>IE: *:0/15</p>
```
marusak, line 283: This needs to be localized

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -233,27 +248,58 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                                                <a href='https://man.archlinux.org/man/systemd.time.7#CALENDAR_EVENTS' target='_blank' rel="noreferrer">More Details</a>
```
marusak, line 284: This needs to be localized as well as it should have icon, for example see:
Screenshot from 2022-07-27 10-24-52

pkg/systemd/services/timers.scss
```diff
@@ -12,4 +12,8 @@
     .specific-repeat-group, .delay-group {
         row-gap: var(--pf-global--spacer--sm);
     }
+
+    .custom-group {
+        width: 50rem;
+    }
```
marusak, line 18: why? This seems fragile on different screen sizes and also I think it broke it for me right now

test/verify/check-system-services
```diff
@@ -1309,6 +1309,67 @@ class TestTimers(MachineCase):
         m.execute("timedatectl set-time '2036-04-10 04:10:00'")
         wait_systemctl_timer("2036-04-10 04:26")
         self.assertIn("2036-04-10 04:26", m.execute("systemctl list-timers"))
+        # creates a new minutely timer that runs at *:*:05 and at *:*:26
+        b.wait_visible("#create-timer")
+        b.click('#create-timer')
```
marusak, line 1314: `b.click()` already does `b.wait_visible()` so no need to do it like this. Feel free to drop the `wait_visible()` call

marusak (changes requested):
Nice, thank you for the fixes! Some comments and also for @garrett about design

garrett (changes requested):
Whoops! I missed the "custom" part of this PR. I'll write what needs to change inline.

pkg/systemd/services/timer-dialog.jsx
garrett: - I missed the custom options. Whoops. (Blame the morning, coffee not kicking in yet, etc.) This isn't acceptable (yet).
- I don't understand what this is supposed to be saying. Is it supposed to be something like this? "Use systemd onCalendar formatting."
- Why does it say "IE"? Is that part of the string? If you're trying to say "i.e." (lowercase, periods needed); that's wrong as it means "that is", it's supposed to be "e.g.", which means "for example"… but please write out "Example:" instead.
- However, we normally DO NOT use popovers for this. It would be better as helper text here, so people can actually refer to it while typing and do not need to click on something extra to see what to do.

Mockup:

create-timer-custom excalidraw(2)

pkg/systemd/services/timers.scss
garrett: Yeah, we shouldn't have hardcoded sizes like this. What is "custom-group"? The naming doesn't even say what it's for.

pkg/systemd/services/timer-dialog.jsx
marusak: Would also link fit into the description?

garrett: Helper text guidelines: https://www.patternfly.org/v4/components/helper-text/design-guidelines/#when-to-use-helper-text

When to use helper text:
> Use helper text to provide imperative information regarding an input, such as character limits, password expectations, and more.

When to use a popover instead of helper text:
> Do not use helper text to explain what a form field is or to provide additional form context. Instead, use a popover to convey any additional, non-imperative information to a user.

When to use placeholder text instead of helper text:
> Do not use helper text to provide formatting examples within a text box. Instead, use placeholder text to temporarily showcase what content should appear in a field before the user inputs information. 

This kind of information falls under the first point, to use helper text, as it's important, necessary information to understand what "custom" means and has an example that needs to be referred to while typing. Also, the placeholder text would go away, so having an example for a specific format versus something simple (like a name) needs to stay put to be referenced.

And, as I mentioned above, showing that information overlapping with where you need it is a problem... additionally, it's an accessibility concern as it changes the context, which makes it tricky for screen readers and extremely difficult for multiple cognitive disabilities (such as anyone with attention issues, "brain fog", short term memory problems, and so on).

I'm just trying to clarify why it's better to use this style instead of the popover. 

Additionally, we also have the space here, so let's use it.

garrett: > Would also link fit into the description?

https://www.patternfly.org/v4/components/helper-text/design-guidelines/#when-to-use-helper-text

> Include hyperlinks to additional information if necessary

If it's offsite (which I assume a link would be), then we need an offsite icon.

We do use offsite links to more information in other places in Cockpit, including helperText areas on forms like this.

So: Yes.

pkg/systemd/services/timer-dialog.jsx
marusak: Also please use https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events rather then arch wiki

pkg/systemd/services/timers.scss
dphaas2004: @garrett and @marusak Thanks for the suggestions.  First time working on cockpit so still trying to figure out your standard practices.  I added the css as I didn't like the icon location trailing underneath the input.  but since we'll switch to helper text that issue goes away.  however it still exists for the delay-group and will need to addressed at a later date to stay away from hard coded sizes.

pkg/systemd/services/timer-dialog.jsx
dphaas2004: Remove the icon as we are switching to helper text

pkg/systemd/services/timer-dialog.jsx
dphaas2004: i changed this to Invalid custom event

pkg/systemd/services/timer-dialog.jsx
dphaas2004: Removed due to change to helpertext

pkg/systemd/services/timer-dialog.jsx
dphaas2004: switched to helper text as suggested.

pkg/systemd/services/timer-dialog.jsx
dphaas2004: added external link icon and switched syntax to match other external links in the system.  updated link address

test/verify/check-system-services
dphaas2004: Removed from Minutely and custom.  this persists on the rest of the timer tests.

dphaas2004 (commented):
This should all be cleaned up now with adding helpertext and removing the custom css

marusak:
Current screenshot for reference:
Screenshot from 2022-08-04 09-34-36

@garrett can you please ack this? Thanks!

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -233,27 +248,56 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                            helperTextInvalid={_("Invalid custom event")}>
+                            <Flex>
+                                <TextInput
+                                    id="custom"
+                                    value={custom}
+                                    onChange={setCustom}
+                                    validated={submitted && validateCustom ? "error" : "default"}
+                                />
+                                <HelperText>
+                                    <HelperTextItem>Use systemd oncalendar format.  Example *:0/5</HelperTextItem>
```
marusak, line 278: This needs to be localized. `<HelperTextItem>{_("Use systemd oncalendar format.  Example *:0/5")}</HelperTextItem>`

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -84,6 +87,15 @@ const CreateTimerDialogBody = ({ owner }) => {
...
         validationFailed.delayNumber = true;
+    if (custom) {
+        cockpit.spawn(["systemd-analyze", "calendar", custom])
+                .then(() => {
+                    setValidateCustom(false);
+                })
+                .catch(() => {
+                    setValidateCustom(true);
+                });
+    }
```
marusak, line 98: This could be written just like:
```
        cockpit.spawn(["systemd-analyze", "calendar", custom])                  
                .then(() => setValidateCustom(false))                           
                .catch(() => setValidateCustom(true));          
```

marusak:
Thank you! Just two tiny nitpicks. I also triggered tests to see how they stand. Also could you please squash all the commits and rebase to current main, thanks!

garrett (changes requested):
Helper text on a form is supposed to be closer to the entry.

https://www.patternfly.org/v4/components/form#horizontal

The link isn't supposed to be on its own line. The link should be modifying the text, not the example. I tried a version with the example in the same line, but it's a bit much, so I guess having the example as a placeholder (as guidance) and linking to the reference is probably more correct in this case, like this:

timer-mod2

garrett:
The problem with the spacing is because of this:

image

It should be using helperText on the input, not the helperText component. They're different things, often used similarly. This is one of PatternFly's confusing naming issues, so it's easy to make this mistake.

Additionally, you can see that the link is outside the text area, so it's on its own line, instead of being inline in the same area.

For an example of the helperText attribute (instead of component), look at https://www.patternfly.org/v4/components/form#horizontal, which has the following snippet:

```jsx
<FormGroup label="Full name" isRequired fieldId="horizontal-form-name" helperText="Include your middle name if you have one.">
  <TextInput value={name} isRequired type="text" id="horizontal-form-name" aria-describedby="horizontal-form-name-helper" name="horizontal-form-name" onChange={handleNameChange} />
</FormGroup>
```

I _think_ you might be able to construct a React.Fragment (`<> … </>`) with an inline link inside and pass that to the helperText attribute.

pkg/systemd/services/timer-dialog.jsx
dphaas2004: I cleaned this up.  Thanks for the suggestion.

pkg/systemd/services/timer-dialog.jsx
dphaas2004: oops.  missed that one.  ill get this cleaned up with Garrets suggestions.

dphaas2004:
@garrett I was able to use a React.Fragment for this so that was nice.  I also added the placeholder text as I too believe that is the correct way to do this.  @marusak should be all Squashed and rebased!  Thx guys!

marusak:
Nice thank you!
Looks like this now:
Screenshot from 2022-08-08 10-33-15

I rebased once more as there was conflict with submodules. Running tests now

marusak:
The test often fails on:
```
wait_js_cond(ph_is_present("#drop-repeat option[value='minutely']")): Uncaught (in promise) Error: condition did not become true
```

I think this might be some race condition, looking

marusak:
`Failed to parse calendar specification '*:': Invalid argument`

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -253,15 +265,46 @@ const CreateTimerDialogBody = ({ owner }) => {
                                         isOpen={isSpecificTimeOpen} setIsOpen={setSpecificTimeOpen}
                                         menuAppendTo={() => document.body} time={specificTime} is24Hour onChange={setSpecificTime} />
                         </FormGroup>}
+                        {repeat == "custom" &&
+                        <FormGroup
+                            label={_("Custom")}
+                            validated={submitted && validateCustom ? "error" : "default"}
```
cockpituous, line 271: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17592-20220809-115752-7fb44acf-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -253,15 +265,46 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                                    </HelperTextItem>
+                                </>
+                            }>
+                            <Flex>
+                                <TextInput
+                                    id="custom"
+                                    value={custom}
+                                    onChange={setCustom}
+                                    placeholder={_("Example *:0/5")}
+                                    validated={submitted && validateCustom ? "error" : "default"}
```
cockpituous, line 293: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17592-20220809-115752-7fb44acf-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -271,14 +314,33 @@ const CreateTimerDialogBody = ({ owner }) => {
                                 validated = "error";
                                 helperTextInvalid = _("Minute needs to be a number between 0-59");
                             }
+                            const sec = repeatPatterns[idx].second;
+                            const validationFailedSecond = !(/^[0-9]+$/.test(sec) && sec <= 59 && sec >= 0);
+
+                            if (submitted && repeat == 'minutely' && validationFailedSecond) {
+                                validated = "error";
+                                helperTextInvalid = _("Second needs to be a number between 0-59");
+                            }
```
cockpituous, line 323: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17592-20220809-115752-7fb44acf-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -271,14 +314,33 @@ const CreateTimerDialogBody = ({ owner }) => {
...
+                                        {repeat == "minutely" && <>
+                                            <TextInput className='delay-number'
+                                                       id={repeat}
+                                                       value={repeatPatterns[idx].second}
+                                                       onChange={second => {
+                                                           const arr = [...repeatPatterns];
+                                                           arr[idx].second = second;
+                                                           setRepeatPatterns(arr);
+                                                       }}
+                                                       validated={submitted && validationFailedSecond ? "error" : "default"} />
```
cockpituous, line 339: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17592-20220809-115752-7fb44acf-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -85,6 +87,11 @@ const CreateTimerDialogBody = ({ owner }) => {
         validationFailed.command = true;
     if (!/^[0-9]+$/.test(delayNumber))
         validationFailed.delayNumber = true;
+    if (custom) {
+        cockpit.spawn(["systemd-analyze", "calendar", custom])
+                .then(() => setValidateCustom(false))
+                .catch(() => setValidateCustom(true));
+    }
```
marusak, line 94: we run this waay to often but show it only when you click on submit button. Let's dial that down then

dphaas2004: I did find it interesting that the validation happens on the fields in the background but only become displayed after the initial submit.  After that all fields are validated "realtime".  should we change this for all inputs?  I kind of liked the ability to validate the custom input as I was writing it but it does become resource heavy.

marusak: Right, I think that is more bug than feature :smile: 
I think what we want is to validate changed fields on focus out and then do validation on submit (to catch missing empty fields).

Do you want to take a look at that? Just please pull the changes I pushed here and feel free to hack away. If not I am happy to take a look at it.

dphaas2004: I can take a look but it wont likely be until Friday with my schedule.  if you get to it before then so be it.  ill check back here before I start anything.

dphaas2004: @marusak I have made numerous updates on how and when inputs are validated.  Please take a look and let me know.  I did leave some comments in the code for now just to assist in checking.  I can remove later if all is good.

marusak:
It works very nicely!
Two issues I found:
Screenshot from 2022-08-18 08-01-06

The button is disabled without explanation why. We might want to add * next to those required fields, see [isRequired prop](https://www.patternfly.org/v4/components/text-input#props)

Screenshot from 2022-08-18 08-01-55

You are able to submit without entering anything in the custom fields.

garrett:
> The button is disabled without explanation why. We might want to add * next to those required fields, see [isRequired prop](https://www.patternfly.org/v4/components/text-input#props)

We haven't done this in Cockpit yet, but really probably should.

Design guidelines on user guidance, starting with required: https://www.patternfly.org/v4/components/form/design-guidelines/#user-guidance

dphaas2004:
@garrett and @marusak 
Little bit of a mess on the existing "specific time" validation routines.  They validate on change currently but that only displays the helpertext.  nothing prevents the submission so it creates bad timers.  tested on current version.  Dont know if you want a separate ticket for that or not?  I have everything else ready but this should be taken care of i would think.  please advise.

dphaas2004:
Fixed repeat validation and rebased.

marusak:
> Little bit of a mess on the existing "specific time" validation routines. They validate on change currently but that only displays the helpertext. nothing prevents the submission so it creates bad timers. tested on current version. Dont know if you want a separate ticket for that or not? I have everything else ready but this should be taken care of i would think. please advise.

Do you have estimate how much work would it take? If it happens currently as well, then we can open issue for it and then fix it separately. I tried it out and the submit button is not disabled but it does nothing when clicking on submit.
Also with empty description the submit button is already enabled, but clicking on it shows that the description cannot be empty.

dphaas2004:
@marusak I did fix the validation issues for the repeating entries which atleast prevents a bad timer from getting generated like the current version.  As for submission with a bad value right now the submit function just returns which is why you dont see anything.  however, the bad fields should be invalidated showing the helperText.  since this type of form validation doesnt yet exist in cockpit im unsure on how we want to display error messages related to this type of failure.  The helperText does show the invalid fields but maybe thats not enough?  It also disables the submit button if the form is not validated.  As for the save button disabling I have been struggling on how best to implement this. The issue you observed where you enter a name and the save button becomes enabled is because I have no default state in the validationFailed hook.  if I set default states then all the fields are invalidated immediately which causes helper text to display right away which I would think is not what we want.

I guess my thought on it was even if the save button is enabled the submit function invalidates the submission and disables the button if the form is not complete.  Since this does prevent bad form submission I guess my thought was lets get the custom field added to the main and fix the now known bug related to the repeating values and then open another issue for the UI and form submissions in general for cockpit and develop a standard on how/when we should validate all form submissions platform wide and start working on implementing that.  thoughts?

Also just a side note I did search patternfly pretty extensively and I did not see anything related to how/when to display error messages back to the user related to form submission.  Could have easily missed it though there is quite a lot going on there. @garrett  may be better at pointing me in the right direction.  Further the form validation section in patternfly is based on state changes which if they get out of sync with the current DOM or are undefined allows for bad form submission.  that's why I opted to use traditional html/javascript form validation checks in combination with state hooks.

garrett:
@dphaas2004: PatternFly has a section on error messages on the form component design page @ https://www.patternfly.org/v4/components/form/design-guidelines/#errors-and-validation:

> When a form field submission results in an error, let users know as soon as possible and as close as possible to the error. Default to presenting error states on a form using field level errors whenever possible.
>
> In some use cases, you may choose to use _inline errors at the top of a form_ to emphasize errors within it. These inline errors should always be a supplemental component used in addition to field level errors.

(Screenshot above shows an inline error at the bottom of the form. It should be at the top, and this should be fixed... either in this PR or in a follow-up. Both are fine, as long as it's addressed at some point soon.)

Then it has additional text, including multiple sub-sections, such as:

- [Error validation on loss of focus](https://www.patternfly.org/v4/components/form/design-guidelines/#error-validation-on-loss-of-focus) 
- [Error validation on submission](https://www.patternfly.org/v4/components/form/design-guidelines/#error-validation-on-submission) 

For in-modal forms in Cockpit, they're checked on the fly during entering information (sometimes), on focus blur (usually, when possible), and on submission (always). In other words, we usually try to check things as soon as possible and _always_ check it during the submission process.

dphaas2004:
@marusak just getting some time to jump back into this.  I believe I have most all cleaned up.  @garrett Thanks for the links.  I have moved the error messages to a new header in the modal and added a statehook for the form to track its validation.  the submit button validation is conditional on the form being valid now.  Please take a look when you get a moment.

only lingering issue I am working on is related to the datepicker.  I see this is a beta feature in patternfly.  There appears to be an open bug on the datepicker validators.  any reason why we cant use a standard date input field for this?

garrett:
@dphaas2004: The native date input has issues, sadly. From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#handling_browser_support:

> The second problem is the more serious one; with date input supported, the value is normalized to the format yyyy-mm-dd. But with a text input, the browser has no recognition of what format the date should be in, and there are many different formats in which people write dates.

...

> At the moment, the best way to deal with dates in forms in a cross-browser way is to have the user enter the day, month, and year in separate controls, or to use a JavaScript library such as [jQuery date picker](https://jqueryui.com/datepicker/).

It's a bummer, but that's how it is. This is why PatternFly has their own widget for it.

garrett (dismissed):
Overall: looks good. :+1: 

I spotted a problem with yearly:

image

I don't think it's related to this PR, however. I'll open up a separate issue for it.

garrett:
If PatternFly's isn't working, then we can use the native one, try to be aware of its shortcoming(s), and test cross-browser (in latest Firefox, Chrome, WebKit). And then we can file bugs upstream to PF.

Generally, I'm a fan of native widgets when possible anyway.

jelly:
Regression with yearly:

image

jelly:
> @jelly yes there was (and still are some) some issues specifically with validation in the repeating loop section of the form. I have made many improvements to this but its still now quite right. ill push my current code that should take care of your concern. Im open to any suggestions on how to get the btn disabled state to work properly with the repeating validation. that is the last missing piece.

If you can wait, I'm going to push a rebased PR and take a further look.

dphaas2004:
yep I can wait.  Also just a side note im pretty positive in the current version the form submission will crash.  I have also fixed this.

jelly:
> yep I can wait. Also just a side note im pretty positive in the current version the form submission will crash. I have also fixed this.

It indeed crashes, locally you should be able to `git pull --rebase` now.

dphaas2004:
@KKoukiou, If someone at cockpit could take a look at the validation routine for the repeating loops as I suggested to @jelly that would be great.  I'm tied up with another project implementation so wont be able to get back on this for a while.

jelly:
Minutely was implemented separately. The custom option is still left to be implemented.

martinpitt:
@dphaas2004 This has become very stale. I'm afraid our team became smaller and we don't have the capacity to fix this up ourselves -- this is rather non-trivial. Are you still interested in finishing this?

martinpitt:
Closing, sorry. Let's reopen if/when someone finds time to work on this.

PR #17595, opened 2022-07-26
systemd: Control isOpen in TimePicker

marusak:
When using `menuAppendTo` the default state is open. That is annoying
as since you open the dialog or switch to any option that contains
TimePicker the menu is opened by default.

### DIFF
--- a/pkg/systemd/services/timer-dialog.jsx
+++ b/pkg/systemd/services/timer-dialog.jsx
@@ -92,13 +92,13 @@ const CreateTimerDialogBody = ({ owner }) => {
                     is24Hour
                     isOpen={repeatPatterns[idx].isOpen || false}
                     setIsOpen={isOpen => {
-                        const arr = [...repeatPatterns];
+                        const arr = JSON.parse(JSON.stringify(repeatPatterns));
                         arr[idx].isOpen = isOpen;
                         setRepeatPatterns(arr);
                     }}
                     menuAppendTo={() => document.body}
                     onChange={time => {
-                        const arr = [...repeatPatterns];
+                        const arr = JSON.parse(JSON.stringify(repeatPatterns));
                         arr[idx].time = time;
                         setRepeatPatterns(arr);
                     }}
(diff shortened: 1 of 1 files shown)


pkg/systemd/services/timer-dialog.jsx
```diff
@@ -89,6 +90,12 @@ const CreateTimerDialogBody = ({ owner }) => {
         <TimePicker className="create-timer-time-picker"
                     time={repeatPatterns[idx].time || "00:00"}
                     is24Hour
+                    isOpen={repeatPatterns[idx].isOpen || false}
+                    setIsOpen={isOpen => {
+                        const arr = [...repeatPatterns];
+                        arr[idx].isOpen = isOpen;
```
martinpitt, line 96: This does not do what you intend. `...` only creates a copy of the list, but not a deep copy of the elements inside it:
```node
> e1 = {open: false}; e2 = {open: true}; l = [e1, e2]
[ { open: false }, { open: true } ]
> l[0]
{ open: false }
> l2 = [...l]
[ { open: false }, { open: true } ]
> l2[0].open = true
true
> l
[ { open: true }, { open: true } ]
```
So you either need to deep-copy (better), or spare the ceremony and directly change `repeatPatterns[idx].isOpen` (it's bad behaviour, but I don't know how it matters in practice).

martinpitt (changes requested):

pkg/systemd/services/timer-dialog.jsx
marusak: It is the same pattern as in `onChange` 2 lines down.  Let me take a look at the deep copy for both cases.

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -92,13 +92,13 @@ const CreateTimerDialogBody = ({ owner }) => {
                     is24Hour
                     isOpen={repeatPatterns[idx].isOpen || false}
                     setIsOpen={isOpen => {
-                        const arr = [...repeatPatterns];
+                        const arr = JSON.parse(JSON.stringify(repeatPatterns));
```
martinpitt, line 95: Heh, this must be the most expensive way ever to create a deep copy :grin: (But it doesn't matter here, convenience and robustness beats performance here)

martinpitt (approved):
Dakujem!

pkg/systemd/services/timer-dialog.jsx
marusak: Yeah, but since it is maximum two elements object it will be just fast enough

PR #17594, opened 2022-07-26
cmmon: Fix memory allocation in cockpit_hex_decode()

kklionz:
Dear Maintainer,
        There is a coding err at ```src/common/cockpithex.c:cockpit_hex_decode```, which leads a over-size-malloc issue.
         Could U please review this PR?

PR #17597, opened 2022-07-26
test: Increase scheduling cost of provisioned VMs, speed up build

martinpitt:
We are still getting a lot of VM boot failures on PSI, so let's reduce
the amount of parallel VMs more.

---

See e.g. [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17594-20220726-100137-69b3f677-ubuntu-stable/log.html) or [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17594-20220726-130123-69b3f677-ubuntu-stable/log.html) run (I cancelled the latter)

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -318,8 +318,8 @@ def detect_tests(test_files, image, opts):
                 nd = getattr(test_method, "_testlib__non_destructive", False)
                 rwa = getattr(test_method, "_testlib__retry_when_affected", True)
                 if getattr(test.__class__, "provision", None):
-                    # each additionally provisioned VM costs half a parallel test capacity
-                    cost = 1 + (len(test.__class__.provision) - 1) / 2
+                    # each additionally provisioned VM costs parallel test capacity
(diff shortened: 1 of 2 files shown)


martinpitt:
@marusak : I added the go-faster stripes as I was annoyed with waiting for all the webpack builds -- but they are quite unrelated. Would you rather want me to split the PR?

martinpitt:
It's pretty near impossible to get ubuntu-stable to green in #17594 (I retried *four* times already). The [run here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17597-20220726-140024-6b004e67-ubuntu-stable/log.html) was successful on the first try (there's still a bunch of other flakes, but not related to VM startup).

martinpitt:
FTR, I added it to my todo list to investigate if ubuntu-stable takes too long to boot, and how we can detect that in our image refreshes.

marusak (approved):
Thanks!

PR #17598, opened 2022-07-27
services: Swap timer order: specific & after boot

garrett:
Change order to prefer the more-commonly-used and default choice "at specific time" instead of "after system boot".

As spotted in #17592.


marusak (approved):
Thanks!

PR #17599, opened 2022-07-27
pybridge: Add basic interactive mode

martinpitt:
Provide a simple replacement for the C bridge's `--interact=separator`
option. This makes it possible for humans to speak the protocol without
having to count the bytes.

Wiring this through main() would require a parallel Router class
hierarchy with multiple inheritance, as asyncio's
connect_accepted_socket() does not accept/pass on any constructor
arguments. So just check if stdin is on a terminal, and automatically
use the interactive mode then. Hardcode the `---\n` separator as that is
reasonably safe and the one used in our blog posts.

----

I'm quite used to develop/debug low-level bridge issues that way, this is also quite useful for metrics. You can paste in some sequence of channel open commands and then watch what happens.

```

{ "command": "open", "channel": "c", "payload": "echo" }
---
DEBUG:cockpit.bridge:Received control message open for channel c: {'command': 'open', 'channel': 'c', 'payload': 'echo'}
DEBUG:cockpit.protocol:sending message  {'channel': 'c', 'command': 'ready'}
DEBUG:cockpit.protocol:writing to transport <_SelectorSocketTransport fd=7 read=polling write=<idle, bufsize=0>>
44

{
  "channel": "c",
  "command": "ready"
}
```

This is an ugly hack, and I don't expect it to actually get merged. But I'd like to place it somewhere in case it's useful for others.

### DIFF
--- a/src/cockpit/protocol.py
+++ b/src/cockpit/protocol.py
@@ -18,6 +18,7 @@
 import asyncio
 import json
 import logging
+import os
 
 logger = logging.getLogger('cockpit.protocol')
 
@@ -35,6 +36,7 @@ class CockpitProtocol(asyncio.Protocol):
     buffered protocols.
     '''
     transport = None
+    is_interactive = os.isatty(0)
     buffer = b''
 
     def do_ready(self):
@@ -77,8 +79,17 @@ class CockpitProtocol(asyncio.Protocol):
         # Nothing to look at?  Save ourselves the trouble...
         if not view:
             return 0
-
         view = bytes(view)
+
+        if self.is_interactive:
+            # we read until we get a '---' separator
+            try:
+                sep = view.index(b'---\n')
+                self.do_frame(view[0:sep])
+                return sep + 4
+            except ValueError:
+                return 0
+
         # We know the length + newline is never more than 10 bytes, so just
         # slice that out and deal with it directly.  We don't have .index() on
         # a memoryview, for example.


martinpitt:
Let's dust this off when we have time and need it.

allisonkarlitskaya:
Minor brain-dump, since I've been thinking about this for a while:

My preferred approach here would be to take something like the `MockTransport` class we have in the unit tests and instantiate that, alongside the bridge, and then drop the user at a Python REPL which makes it easy to invoke methods on the transport.  That would provide a nice high-level API for things like opening channels, pushing data, etc, complete with tab completion and not completely destroying your session on simple typos.  It would also allow poking at the internal state of the bridge using Python, which is likely to be extremely useful as well.

The main blocker there is that it's difficult to mix Python's REPL with asyncio.  I have a couple of ideas for how I'd approach that:
 - just force the user to manually run the loop for 1 second when they await results
 - use two separate threads for REPL input and handling the output

For what it's worth, I did a bit of looking into how this can all be implemented and the `readline`, `rlcompleter` and `code.InteractiveInterpreter` class are interesting starting points, if we want to get fancy.  Honestly — and particularly for a first pass — I think we might just want to run the basic default interpreter...

martinpitt:
@allisonkarlitskaya FWIW, that's not how I usually tested and developed the C bridge. I was usually building canned commands, such as

    (printf '\n{ "command": "open", "channel": "m", "payload": "metrics1", "source": "internal", "interval": 1000, "metrics": [ { "name": "cpu.basic.user", "derive": "rate" } ] }\n---\n'; sleep 5) | cockpit-bridge --interact=---

and then re-ran this after each edit/compile cycle, possibly with `G_MESSAGES_DEBUG` and such. An interactive REPL isn't useful for iterating code, but of course it may be very useful for debugging.

Issue #17580, opened 2022-07-20
No CPU temp shown in metrics for AMD AM2/3 CPUs


g6avk:
### Explain what happens

Go to overview > Metrics and History and no CPU temp is shown.

![Screenshot_20220720_230339](https://user-images.githubusercontent.com/4700941/180090611-43ae5616-1ea3-4595-9f93-7e18cd119abf.jpeg)

### Version of Cockpit

273

### Where is the problem in Cockpit?

Metrics

### Server operating system

Fedora

### Server operating system version

36

### What browsers are you using?

Brave

martinpitt:
On that machine, can you please copy&paste the output of these commands?

    grep -r . /sys/class/hwmon/hwmon*/name
    grep -r . /sys/class/hwmon/hwmon*/temp*

g6avk:
$ grep -r . /sys/class/hwmon/hwmon*/name

/sys/class/hwmon/hwmon0/name:atk0110
/sys/class/hwmon/hwmon1/name:k10temp
/sys/class/hwmon/hwmon2/name:hidpp_battery_0
/sys/class/hwmon/hwmon3/name:hidpp_battery_1

 grep -r . /sys/class/hwmon/hwmon*/temp*

/sys/class/hwmon/hwmon0/temp1_crit:95000
/sys/class/hwmon/hwmon0/temp1_input:42000
/sys/class/hwmon/hwmon0/temp1_label:CPU Temperature
/sys/class/hwmon/hwmon0/temp1_max:60000
/sys/class/hwmon/hwmon0/temp2_crit:75000
/sys/class/hwmon/hwmon0/temp2_input:37000
/sys/class/hwmon/hwmon0/temp2_label:MB Temperature
/sys/class/hwmon/hwmon0/temp2_max:45000
/sys/class/hwmon/hwmon1/temp1_crit:99500
/sys/class/hwmon/hwmon1/temp1_crit_hyst:97500
/sys/class/hwmon/hwmon1/temp1_input:0
/sys/class/hwmon/hwmon1/temp1_max:70000

tomasmatus:
> /sys/class/hwmon/hwmon1/temp1_input:0

Well this is weird. The `k10temp` driver seems to export 0C.
What CPU model is this? Are other programs such as `lm-sensors` able to read CPU Temp?

g6avk:
The CPU is  AMD Phenom II X4 965

[...]

PR #17601, opened 2022-07-28
Metrics: Improve CPU temperature sensors detection

tomasmatus:
Sensors are detected once and are stored in a static list.

This also adds `atk0110` sensor mentioned in #17580

### DIFF
--- a/src/bridge/cockpitcpusamples.c
+++ b/src/bridge/cockpitcpusamples.c
@@ -132,25 +132,37 @@ read_file (const gchar *path)
 }
 
 static void
-sample_cpu_sensors (CockpitSamples *samples,
+sample_cpu_sensors (gchar *sensor_path,
+                    CockpitSamples *samples)
+{
+    g_autofree gchar *temp_content = read_file (sensor_path);
+
+    if (temp_content == NULL)
+        return;
+
+    gint64 temperature = g_ascii_strtoll (temp_content, NULL, 10);
+    // overflow or invalid
+    if (temperature == G_MAXINT64 || temperature == G_MININT64 || temperature == 0)
+      {
+        g_debug("Invalid number in %s: %m", sensor_path);
+        return;
+      }
+
+    cockpit_samples_sample (samples, "cpu.temperature", sensor_path, temperature/1000);
+}
+
+static void
+detect_cpu_sensors (GList **devices,
+                    const gchar *name,
                     int hwmonID)
 {
   for (int i = 1; TRUE; i++)
     {
-      g_autofree gchar *temp = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_input", hwmonID, i);
-      g_autofree gchar *temp_content = read_file (temp);
-
-      // end loop on error
-      if (temp_content == NULL)
+      g_assert (devices != NULL);
+      // break loop if input file does not exist
+      g_autofree gchar *sensor_path = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_input", hwmonID, i);
+      if (!g_file_test (sensor_path, G_FILE_TEST_EXISTS))
           break;
-
-      gint64 temperature = g_ascii_strtoll (temp_content, NULL, 10);
-      // overflow or invalid
-      if (temperature == G_MAXINT64 || temperature == G_MININT64 || temperature == 0)
-        {
-          g_debug("Invalid number in %s: %m", temp);
-          continue;
-        }
 
       g_autofree gchar *label = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_label", hwmonID, i);
       g_autofree gchar *label_content = read_file (label);
@@ -158,22 +170,27 @@ sample_cpu_sensors (CockpitSamples *samples,
       if (label_content == NULL)
         {
           // labels aren't used on ARM
-          label_content = g_strdup_printf ("Core %d", i);
+          if (!g_str_equal (name, "cpu_thermal"))
+              continue;
+        }
+      else
+        {
+          g_strchomp (label_content);
+
+          // only sample CPU Temperature in atk0110
+          if (!g_str_equal (label_content, "CPU Temperature") && g_str_equal (name, "atk0110"))
+              continue;
+          // ignore Tctl on AMD devices
+          if (g_str_equal (label_content, "Tctl"))
+              continue;
         }
 
-      g_strchomp (label_content);
-
-      // ignore Tctl on AMD devices
-      if (g_str_equal (label_content, "Tctl"))
-          continue;
-
-      g_autofree gchar *instance = g_strdup_printf ("hwmon%d %s", hwmonID, label_content);
-      cockpit_samples_sample (samples, "cpu.temperature", instance, temperature/1000);
+      *devices = g_list_append (*devices, g_steal_pointer (&sensor_path));
     }
 }
 
-void
-cockpit_cpu_temperature (CockpitSamples *samples)
+static void
+detect_hwmon_device (GList **devices)
 {
   // iterate through all hwmon folders to find CPU sensors
   for (int i = 0; TRUE; i++)
@@ -188,12 +205,26 @@ cockpit_cpu_temperature (CockpitSamples *samples)
       g_strchomp (name);
 
       // compare device name with CPU names
-      // Intel: coretemp, AMD: k8temp or k10temp, ARM: cpu_thermal
+      // Intel: coretemp, AMD: k8temp or k10temp, ARM: cpu_thermal, Asus motherboard: atk0110
       if (g_str_equal (name, "coretemp") || g_str_equal (name, "cpu_thermal") ||
-          g_str_equal (name, "k8temp") || g_str_equal (name, "k10temp"))
+          g_str_equal (name, "k8temp") || g_str_equal (name, "k10temp") || g_str_equal (name, "atk0110"))
         {
           // hwmon contains CPU info
-          sample_cpu_sensors (samples, i);
+          detect_cpu_sensors (devices, name, i);
         }
     }
+}
+
+void
+cockpit_cpu_temperature (CockpitSamples *samples)
+{
+  static GList *devices = NULL;
+  static gboolean initialized = FALSE;
+  if (!initialized)
+    {
+      detect_hwmon_device (&devices);
+      initialized = TRUE;
+    }
+
+  g_list_foreach (devices, (GFunc)sample_cpu_sensors, samples);
 }
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -839,65 +839,113 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
         b.wait_not_present("#current-metrics-card-cpu .temperature")
 
         # No matching type
-        m.execute("mkdir -p /tmp/hwmon/hwmon0")
-        m.execute("echo BAT0 > /tmp/hwmon/hwmon0/name")
-        m.execute("mount -o bind /tmp /sys/class")
+        m.execute("mkdir -p /tmp/sensor-sys-class/hwmon/hwmon0")
+        self.addCleanup(m.execute, "rm -rf /tmp/sensor-sys-class")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon0/name", "BAT0")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon0/temp1_input", "40000")
+        m.execute("mount -o bind /tmp/sensor-sys-class /sys/class")
+        self.addCleanup(m.execute, "umount /sys/class")
+        b.logout()
+        self.login_and_go("/metrics")
 
         b.wait_not_present("#current-metrics-card-cpu .temperature")
 
         # create files that contain CPU temperature
-        m.execute("mkdir -p /tmp/hwmon/hwmon1")
+        # ARM
+        m.execute("mkdir -p /tmp/sensor-sys-class/hwmon/hwmon1")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/name", "cpu_thermal")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp1_input", "30000")
 
-        # Testing on ARM devices
-        m.execute("echo cpu_thermal > /tmp/hwmon/hwmon1/name")
-        m.execute("echo 60000 > /tmp/hwmon/hwmon1/temp1_input")
+        b.logout()
+        self.login_and_go("/metrics")
 
-        b.wait_in_text("#current-metrics-card-cpu", "60 °C")
-
-        # Testing on AMD devices
-        m.execute("echo k10temp > /tmp/hwmon/hwmon1/name")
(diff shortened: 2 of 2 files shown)


src/bridge/cockpitcpusamples.c
```diff
@@ -132,48 +132,65 @@ read_file (const gchar *path)
...
+        g_debug("Invalid number in %s: %m", sensor_path);
+        return;
+      }
+
+    cockpit_samples_sample (samples, "cpu.temperature", sensor_path, temperature/1000);
+}
+
+static void
+add_cpu_sensor_files (GList **devices,
+                      gchar *name,
```
martinpitt, line 156: This needs to be `const`, as the caller retains ownership and this function does not, and should not, modify it.

src/bridge/cockpitcpusamples.c
```diff
@@ -132,48 +132,65 @@ read_file (const gchar *path)
...
 
-      g_strchomp (label_content);
-
-      // ignore Tctl on AMD devices
-      if (g_str_equal (label_content, "Tctl"))
-          continue;
-
-      g_autofree gchar *instance = g_strdup_printf ("hwmon%d %s", hwmonID, label_content);
-      cockpit_samples_sample (samples, "cpu.temperature", instance, temperature/1000);
+      if (devices != NULL)
```
martinpitt, line 187: This shouldn't be a runtime check -- if it's NULL, it's a programming error. Please make this an assertion instead at the top of the function.

You can also simplify this by just creating a devices list from scratch and returning it, instead of getting passed a pointer to the list pointer. (That's a matter of taste, I'm fine with either.)

src/bridge/cockpitcpusamples.c
```diff
@@ -132,48 +132,65 @@ read_file (const gchar *path)
...
+      {
+        g_debug("Invalid number in %s: %m", sensor_path);
+        return;
+      }
+
+    cockpit_samples_sample (samples, "cpu.temperature", sensor_path, temperature/1000);
+}
+
+static void
+add_cpu_sensor_files (GList **devices,
```
martinpitt, line 155: Perhaps detect_cpu_sensor() for more symmetry, or detect_one_cpu_sensor() if you want to make it a bit more distinct? "_files" sounds a bit weird.

test/verify/check-metrics
```diff
@@ -840,64 +840,64 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
 
         # No matching type
         m.execute("mkdir -p /tmp/hwmon/hwmon0")
-        m.execute("echo BAT0 > /tmp/hwmon/hwmon0/name")
+        self.addCleanup(m.execute, "rm -rf /tmp/hwmon")
+        m.write("/tmp/hwmon/hwmon0/name", "BAT0")
+        m.write("/tmp/hwmon/hwmon0/temp1_input", "40000")
         m.execute("mount -o bind /tmp /sys/class")
```
martinpitt, line 846: This is a bit brittle -- there are often other files in /tmp which may disturb the layout. I recommend creating a new /tmp/sensor-sys-class/ or so directory for that, and only bind that, instead of the whole /sys. Sorry for missing that in the original PR.

test/verify/check-metrics
```diff
@@ -840,64 +840,64 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
+        # atk0110 motherboard
+        m.execute("mkdir -p /tmp/hwmon/hwmon4")
+        m.write("/tmp/hwmon/hwmon4/name", "atk0110")
+        # MB Temperature will be ignored
+        m.write("/tmp/hwmon/hwmon4/temp1_label", "MB Temperature")
+        m.write("/tmp/hwmon/hwmon4/temp2_label", "CPU Temperature")
+        m.write("/tmp/hwmon/hwmon4/temp1_input", "40000")
+        m.write("/tmp/hwmon/hwmon4/temp2_input", "30000")
+
+        b.wait_in_text("#current-metrics-card-cpu", "30 °C")
```
martinpitt, line 882: You cannot tell from this whether Tctl's temp2_input was actually ignored or considered, or whether  "CPU Temperature" was detected, as they all use 30 degrees. You need to set up and test each scenario individually, and reload the page (or log out and back in, to restart the bridge).

martinpitt (changes requested):
Thank you! The code looks great, just needs some minor tweaks. The tests are failing, and need some work anyway.

src/bridge/cockpitcpusamples.c
tomasmatus: This doesn't really detect only one sensor. `detect_cpu_sensors` finds correct hwmon and `add_cpu_sensor_files` finds all CPU related `tempY_input` files in said hwmonX. I'll think of a better naming for both of these functions.

zeroZshadow:
Am I understanding that this will IGNORE use the `Tctl` labeled sensor on AMD chips? This is the only label that has the cpu temperature on my setup.

src/bridge/cockpitcpusamples.c
martinpitt: The new names are nicer, thank you!

test/verify/check-metrics
```diff
@@ -839,65 +839,89 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
         b.wait_not_present("#current-metrics-card-cpu .temperature")
 
         # No matching type
-        m.execute("mkdir -p /tmp/hwmon/hwmon0")
-        m.execute("echo BAT0 > /tmp/hwmon/hwmon0/name")
-        m.execute("mount -o bind /tmp /sys/class")
+        m.execute("mkdir -p /tmp/sensor-sys-class/hwmon/hwmon0")
+        self.addCleanup(m.execute, "rm -rf /tmp/sensor-sys-class/hwmon")
```
martinpitt, line 843: Please clean up the whole /tmp/sensor-sys-class directory.

test/verify/check-metrics
```diff
@@ -839,65 +839,89 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
-        # create files that contain CPU temperature
-        m.execute("mkdir -p /tmp/hwmon/hwmon1")
-
-        # Testing on ARM devices
-        m.execute("echo cpu_thermal > /tmp/hwmon/hwmon1/name")
-        m.execute("echo 60000 > /tmp/hwmon/hwmon1/temp1_input")
-
-        b.wait_in_text("#current-metrics-card-cpu", "60 °C")
+        def clean_and_relog():
+            m.execute("rm -rf /sys/class/hwmon/hwmon1/*")
```
martinpitt, line 852: In your test you only ever check hwmon1/, not a second one. This should be done, but can be a follow-up.

test/verify/check-metrics
```diff
@@ -839,65 +839,89 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
+            b.relogin()
+            b.go("/metrics")
+            b.enter_page("/metrics")
```
martinpitt, line 855: This looks like a race condition to me -- you already open the metrics page *without any sensors set up*. As the bridge only detects them once, ever, you currently hope that the page opens the channel slower than the test will create the fake files. Do your tests still work if you add a `time.sleep(3)` after this line here? If they do, then I don't understand how this can work at all.

You should better split the cleanup/logout and the relogin, and put the new /sys setup in  between for each scenario.

test/verify/check-metrics
```diff
@@ -839,65 +839,89 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
+
+        # AMD
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/name", "k10temp")
+        # Tctl (temp1_input) will be ignored
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp1_label", "Tctl")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp2_label", "Tccd1")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp3_label", "Tccd3")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp1_input", "40000")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp2_input", "35000")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp3_input", "30000")
```
martinpitt, line 877: This would be a bit easier to read if you keep the _label and _input of a particular sensor together; i.e. the "40000" will be ignored.

test/verify/check-metrics
```diff
@@ -839,65 +839,89 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
 
-        # create files that contain CPU temperature
-        m.execute("mkdir -p /tmp/hwmon/hwmon1")
-
-        # Testing on ARM devices
-        m.execute("echo cpu_thermal > /tmp/hwmon/hwmon1/name")
-        m.execute("echo 60000 > /tmp/hwmon/hwmon1/temp1_input")
-
-        b.wait_in_text("#current-metrics-card-cpu", "60 °C")
+        def clean_and_relog():
```
martinpitt, line 851: Please "relogin"; "log" mean something else.

martinpitt (changes requested):
Thanks! I retried some [weird test failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17601-20220729-050131-728e75ad-ubuntu-stable/log.html) which looked like infra trouble. But still some details to fix.

test/verify/check-metrics
martinpitt: [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17601-20220729-050759-728e75ad-centos-8-stream/log.html#305-1) could very well be due to this race condition?

src/bridge/cockpitcpusamples.c
```diff
@@ -188,12 +205,22 @@ cockpit_cpu_temperature (CockpitSamples *samples)
...
         }
     }
 }
+
+void
+cockpit_cpu_temperature (CockpitSamples *samples)
+{
+  static GList *devices = NULL;
+  if (devices == NULL)
+      detect_hwmon_device (&devices);
```
martinpitt, line 223: This isn't sufficient -- if this does not find any sensors (like in a VM or embedded device), this will run the initialization over and over again. Please introduce a separate `static gboolean initialized = FALSE` and use that to run (or not) the detection. I.e. you need to tell apart "never run" from "nothing found".

This will also "break" the tests with the sleep.

src/bridge/cockpitcpusamples.c
```diff
@@ -132,48 +132,65 @@ read_file (const gchar *path)
...
+        return;
+      }
+
+    cockpit_samples_sample (samples, "cpu.temperature", sensor_path, temperature/1000);
+}
+
+static void
+detect_cpu_sensors (GList **devices,
+                      const gchar *name,
+                      int hwmonID)
```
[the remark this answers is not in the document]
tomasmatus: How does this even happen...

test/verify/check-metrics
tomasmatus: I added a test where it checks hwmon1 and hwmon2 at the same time and shows the hottest sensor.

src/bridge/cockpitcpusamples.c
tomasmatus: Thanks for pointing this out!

src/bridge/cockpitcpusamples.c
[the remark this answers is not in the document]
martinpitt: You renamed the function, the old name was two characters longer. Supposedly you just forgot to adjust the indentation of the other lines. Happens all the time to all of us, don't worry :grin:

src/bridge/cockpitcpusamples.c
```diff
@@ -188,12 +205,26 @@ cockpit_cpu_temperature (CockpitSamples *samples)
...
         }
     }
 }
+
+void
+cockpit_cpu_temperature (CockpitSamples *samples)
+{
+  static GList *devices = NULL;
+  static gboolean initialized = FALSE;
+  if (initialized == FALSE)
```
martinpitt, line 223: These are booleans, so `if (!initialized)`. If nothing else requires changes any more, I'll let that slide, though.

test/verify/check-metrics
```diff
@@ -839,65 +839,107 @@ podman save quay.io/libpod/busybox | sudo -i -u admin podman load
...
-        m.execute("mkdir -p /tmp/hwmon/hwmon1")
+        # ARM
+        m.execute("mkdir -p /tmp/sensor-sys-class/hwmon/hwmon1")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/name", "cpu_thermal")
+        m.write("/tmp/sensor-sys-class/hwmon/hwmon1/temp1_input", "30000")
 
-        # Testing on ARM devices
-        m.execute("echo cpu_thermal > /tmp/hwmon/hwmon1/name")
-        m.execute("echo 60000 > /tmp/hwmon/hwmon1/temp1_input")
+        self.login_and_go("/metrics")
```
martinpitt, line 858: Curious that this works without a previous `b.logout()`. I don't see anything in `login_and_go()` which would log you out first.

tomasmatus:
@zeroZshadow thank you for pointing this out.
[Kernel documentation](https://docs.kernel.org/hwmon/k10temp.html) says:
> It does _not_ represent an actual physical temperature like die or case temperature.

But it seems we'll have to reinvestigate what this sensor actually tells us.

martinpitt (approved):
Well done, thank you!

zeroZshadow:
@tomasmatus To give an indication of what seems to be available on my setup.

```text
$ grep -r . /sys/class/hwmon/hwmon*/temp*
/sys/class/hwmon/hwmon0/temp1_crit:20800
/sys/class/hwmon/hwmon0/temp1_input:16800
/sys/class/hwmon/hwmon0/temp2_crit:20800
/sys/class/hwmon/hwmon0/temp2_input:16800
/sys/class/hwmon/hwmon1/temp1_alarm:0
/sys/class/hwmon/hwmon1/temp1_crit:84850
/sys/class/hwmon/hwmon1/temp1_input:38850
/sys/class/hwmon/hwmon1/temp1_label:Composite
/sys/class/hwmon/hwmon1/temp1_max:81850
/sys/class/hwmon/hwmon1/temp1_min:-273150
/sys/class/hwmon/hwmon1/temp2_input:38850
/sys/class/hwmon/hwmon1/temp2_label:Sensor 1
/sys/class/hwmon/hwmon1/temp2_max:65261850
/sys/class/hwmon/hwmon1/temp2_min:-273150
/sys/class/hwmon/hwmon1/temp3_input:42850
/sys/class/hwmon/hwmon1/temp3_label:Sensor 2
/sys/class/hwmon/hwmon1/temp3_max:65261850
/sys/class/hwmon/hwmon1/temp3_min:-273150
/sys/class/hwmon/hwmon2/temp1_input:37000
/sys/class/hwmon/hwmon2/temp1_label:Tctl
grep: /sys/class/hwmon/hwmon3/temp1_input: No data available

$ grep -r . /sys/class/hwmon/hwmon*/name
/sys/class/hwmon/hwmon0/name:acpitz
/sys/class/hwmon/hwmon1/name:nvme
/sys/class/hwmon/hwmon2/name:k10temp
/sys/class/hwmon/hwmon3/name:iwlwifi_1
```

This is on Debian 11 (Bullseye) using the 5.18 backport kernel with an `AMD 5600G` cpu in a `B550I AORUS PRO AX` motherboard.
The newer kernel was installed as otherwise this cpu wasn't supported at all, and k10temp didnt even show up.

PR #17611, opened 2022-08-01
test: Adjust pixel tests for changed NM IPv6 behaviour

martinpitt:
NetworkManager 1:1.38.0-2.fc36 → 1:1.38.2-1.fc36 changed its IPv6
address calculation that ends up with a wider address. This causes a
line wrap in the mobile view.

---

See [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3672-20220730-120638-40b8e106-fedora-36-cockpit-project-cockpit/pixeldiff.html#TestNetworkingBasic-testBasic-network-main-mobile) in https://github.com/cockpit-project/bots/pull/3672 . Needs to be landed in lockstep.

[pixel diff review](https://github.com/cockpit-project/pixel-test-reference/compare/a740b3b4e8c7a97f7ac1e28aaf03cea54b16e5e5..c3336e7b3877046ae02f52c5aec21e2b704270fa)


garrett (approved):
Wow, this is unfortunate, as there's way more than enough space to not need this wrapping. Anyway, that's best fixed elsewhere.

Here's a visualization of the unused space in the label gap:
image

PF is way too generous with the fixed width label space. If only all browsers supported subgrid already. (Firefox + Safari do. We're waiting on Chrome and its siblings.)

We could do a custom override too, I guess.

PR #17616, opened 2022-08-02
Firewall: add services dialog fix

garrett:


garrett:
This includes the fix for the CSS removal and also improves the scrollable services list so it's a better height (for both desktop and mobile) and has a border. It will cause the pixel tests to break, as the border and height will be different from the reference image.

garrett:
Default:

image

Empty:

image

Partial:

image

Extremely small mobile (iPhone SE):

Screen Shot 2022-08-02 at 09 55 29

pkg/networkmanager/networking.scss
```diff
@@ -1,6 +1,8 @@
 @use "ct-card";
 @use "page";
 @import "global-variables";
+// Needed for Firewall's add services dialog's filter layout
```
garrett, line 4: Note that I've added a comment as to why we're importing this CSS here, to prevent this mistaken removal in the future.

martinpitt:
Meeh, this is [still broken](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17616-20220802-075257-49000c4e-fedora-36/pixeldiff.html#TestFirewall-testAddServices-firewall-add-services-to-zone-modal). Could this be one of these "production vs. devel build" issues again?

garrett:
> Meeh, this is [still broken](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17616-20220802-075257-49000c4e-fedora-36/pixeldiff.html#TestFirewall-testAddServices-firewall-add-services-to-zone-modal). Could this be one of these "production vs. devel build" issues again?

Only if our CSS "optimizer" is breaking things again.

Again, I really don't like the fact that we "optimize" our CSS, for random breakages such as these (and others). Even when we test what things look like, it still manages to mess things up. 

As long as the CSS is sent over the wire compressed, the optimizer probably doesn't really do much (except cause headaches once in a while, randomly).

garrett:
It looks as expected in Chrome (Version 103.0.5060.134 (Official Build) unknown (64-bit)):

Screenshot from 2022-08-02 15-16-51

garrett:
We should either land this before the next release, fixing #17616, or revert #17612 (and then land this PR or something similar to it).

martinpitt:
When I run this PR locally, I get an expected/sensible pixel diff. The new screenshot is this:
TestFirewall-testAddServices-firewall-add-services-to-zone-modal-pixels

... which I believe is right. The one [spat out on our CI](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17616-20220802-075257-49000c4e-fedora-36/TestFirewall-testAddServices-firewall-add-services-to-zone-modal-pixels.png) is wrong.

However, I currently ran this on the *previous* cockpit/tasks version, with chromium-102.0.5005.115-1.fc36.x86_64 . I'm going to retry this on the current tasks container with 103.

garrett:
> ... which I believe is right.

Correct! :+1: :tada: 

It has:

✔ the correct filter layout
✔ a line around the scrollable list
✔ a taller list

martinpitt:
I pushed the updated pixels that I get locally (the ones that look as intended), and I also updated the storage one. Let's see how this behaves on CI, I'll watch it while it's running.

martinpitt:
I replaced the test-only hack commit with something that we can (grudgingly) land. As I now meddled with this too much to review myself -- @marusak , can you please have a look?

marusak (approved):
Thank!

martinpitt:
The firewall pixel failure still happens, e.g. [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17614-20220803-052318-bdd91803-fedora-36/log.html#80) or [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17617-20220803-052202-98986fd1-fedora-36/log.html).

I am now examining this on rhos-01-6. Preparation:
```
git clone https://github.com/cockpit-project/cockpit
cd cockpit
test/common/pixel-tests pull
tools/webpack-jumpstart --partial
export TEST_OS=fedora-36
test/image-prepare -q
```

Scenarios:

* `test/common/run-tests --test-dir test/verify TestFirewall.testAddServices`: passed 3 times
* `test/common/run-tests --test-dir test/verify TestFirewall TestNetworkingBasic` (8 parallel tests, for maximum load):  passed
* `for i in $(seq 5); do test/common/run-tests --test-dir test/verify TestFirewall TestNetworkingBasic || break; done`: passed
* same with this wait commit reverted: passed

So, I'm none the wiser.. I started a complete test run now, just to see if it's "even more load" which triggers it.

PR #17620, opened 2022-08-03
lib: allow setting initial value in FileAutoComplete

tomasmatus:
Allows setting initial value in FileAutoComplete component. This use case
is needed in c-podman to prefill ImageRunModal when editing/recreating containers.


martinpitt (approved):
Thanks!

PR #17621, opened 2022-08-04
http-stream2 unit tests and doc improvements

martinpitt:
Noticed while I was working on the pybridge http channel. The JS unit tests don't run in our VMs, thus no-test.

### DIFF
--- a/pkg/base1/test-http.js
+++ b/pkg/base1/test-http.js
@@ -115,21 +115,20 @@ QUnit.test("not found", assert => {
 
 QUnit.test("streaming", assert => {
     const done = assert.async();
-    assert.expect(1);
+    assert.expect(3);
 
-    let at = 0;
+    let num_chunks = 0;
     let got = "";
     cockpit.http(test_server)
             .get("/mock/stream")
             .stream(resp => {
                 got += resp;
-                at++;
+                num_chunks++;
             })
             .finally(() => {
-                let expected = "";
-                for (let i = 0; i < at; i++)
-                    expected += String(i) + " ";
-                assert.equal(got, expected, "stream got right data");
+                assert.true(num_chunks > 1, "got at least two chunks");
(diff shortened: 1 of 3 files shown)


martinpitt:
I tried for a bit to find a cheap way to connect to a Unix socket, but this is too annoyingly involved in C. Let's test that through integration tests instead.

pkg/base1/test-http.js
```diff
@@ -25,12 +25,12 @@ const test_server = {
     port: parseInt(window.location.port, 10)
 };
 
-QUnit.test("simple request", function (assert) {
+QUnit.test("simple request", assert => {
     const done = assert.async();
-    assert.expect(2);
+    assert.expect(1);
```
marusak, line 30: hmm, what does 1 and 2 mean?

pkg/base1/test-http.js
```diff
@@ -308,39 +280,31 @@ QUnit.test("http connection different", async function (assert) {
...
     assert.equal(first, second, "same connection");
 });
 
-QUnit.test("no dns address", function (assert) {
+QUnit.test("no dns address", assert => {
     assert.expect(1);
 
     assert.rejects(cockpit.http({ port: 8080, address: "the-other-host.example.com" }).get("/"),
-                   /* Unfortunately we can see either of these errors when running unit tests */
+                   //
```
marusak, line 295: hm? Empty comment?

pkg/base1/test-http.js
martinpitt: It's the number of assertions that a test expects to run. As I dropped the `assert.equal(this.state(), ...`, I had to reduce it by one.

pkg/base1/test-http.js
martinpitt: Whoops! Thanks for spotting, restored.

marusak (approved):
Thank you!

PR #17622, opened 2022-08-05
client: Support WebKit 4.1 API

martinpitt:
Fedora 37 dropped webkit2gtk3 from their default desktop installation,
and only ships with webkit2gtk4.1 now. Confusingly, this is *not* GTK4,
but the GTK3 binding linked against libsoup 3.

-----

@M4rtinK, this should fix your woes. Quite possibly it'll also fix https://github.com/cockpit-project/bots/pull/3696

I tested that on a current Fedora rawhide workstation live system. Both with the default webkit2gtk4.1, and after installing webkit2gtk3 and damaging the gtk4.1 import code path.

### DIFF
--- a/src/client/cockpit-client
+++ b/src/client/cockpit-client
@@ -29,10 +29,15 @@ import sys
 import tempfile
 import textwrap
 
-gi.require_version("Gtk", "3.0")  # NOQA
-gi.require_version("WebKit2", "4.0")  # NOQA
+gi.require_version("Gtk", "3.0")
+from gi.repository import GLib, Gio, Gtk  # NOQA
 
-from gi.repository import GLib, Gio, Gtk, WebKit2  # NOQA: E402
+try:
+    gi.require_version("WebKit2", "4.1")
+    from gi.repository import WebKit2
+except (ValueError, ImportError):
+    gi.require_version("WebKit2", "4.0")
+    from gi.repository import WebKit2
 
 try:
     gi.require_version("Handy", "1")


kalev:
A quick drive by comment:

The change looks good to me, but the commit message is wrong. This is not at all about GTK3 vs GTK4, it's about libsoup 2.4 vs 3.0.

Due to symbol conflicts it's not possible to link to both libsoup 2.4 and 3.0 in the same process. To make it possible for apps to choose which libsoup they want to use, webkitgtk has two parallel installable versions available, webkit2gtk 4.0 and 4.1, the only difference being that 4.0 is linked with libsoup 2.4 and 4.1 is linked with libsoup 3.0. Otherwise it's the same API and both are using GTK3.

Fedora 37 has switched everything in the Workstation default install over to webkit2gtk 4.1 ABI. 4.0 is available in the repos for apps that are still linking with libsoup 2.4 and need to use the libsoup 2.4 version of webkitgtk because of that.

Since I guess cockpit doesn't link to libsoup and doesn't link to any other libraries that link to libsoup it should be fine to choose either version of webkitgtk :)

martinpitt:
@kalev : Thanks for pointing out! And ugh for really confusing package names (webkit2gtk3 vs. webkit2gtk4..). I'll also change the code a bit then, as the GTK import should look differently then. I accidentally dropped the "3.0" version requirement anyway, that was just an experiment.

kalev:
Yep, the naming is a mess :) There's a bit of transition going on in F37 right now and the plan is to end up with the following naming:

F36:
webkit2gtk3 (gtk3 version, 4.0 API)

F37:
webkit2gtk4.0 (gtk3 version, 4.0 API, libsoup 2.4)
webkit2gtk4.1 (gtk3 version, 4.1 API, libsoup 3.0)
webkit2gtk5.0 (gtk4 version, 5.0 API)

The last one might still change, depending on how upstream ends up naming the API. See also https://fedoraproject.org/wiki/Changes/libsoup_3:_Part_One and https://bugzilla.redhat.com/show_bug.cgi?id=2108905

martinpitt:
Thanks @kalev . I cleaned this up and re-tested on rawhide live system.

src/client/cockpit-client
```diff
@@ -30,9 +30,14 @@ import tempfile
 import textwrap
 
 gi.require_version("Gtk", "3.0")  # NOQA
-gi.require_version("WebKit2", "4.0")  # NOQA
+from gi.repository import GLib, Gio, Gtk  # NOQA
 
-from gi.repository import GLib, Gio, Gtk, WebKit2  # NOQA: E402
+try:
+    gi.require_version("WebKit2", "4.1")  # NOQA
```
allisonkarlitskaya, line 36: I think you should be able to drop the `# NOQA` since this is no longer a top-level import, no?

allisonkarlitskaya (commented):
Awful that we need to do this for something that's really an implementation detail (at least for us), but I guess if it needs to be, then it needs to be...

Just out of curiosity: would we get the big awful warning if we just omitted the `require_version()` call entirely?

src/client/cockpit-client
martinpitt: Indeed, thanks! Fixed.

martinpitt:
> would we get the big awful warning if we just omitted the require_version() call entirely

Just tested, and we do. Also, I find that kind of okay -- once we have, say, a WebKit2 "5.0" API (for GTK 4?), we'll most likely have to do some porting anyway. I find it better to explicitly specify the compatible API(s), even though this currently looks awful with the try/ImportError.

allisonkarlitskaya (approved):

PR #17624, opened 2022-08-08
tests: Fix for Machine.execute() enabling `set -e`

martinpitt:
This actually spotted two bugs. Blocks https://github.com/cockpit-project/bots/pull/3706

 - https://github.com/cockpit-project/cockpit/pull/17625

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -772,7 +772,7 @@ class TestConnection(MachineCase):
         self.assertIn('id="overview"', out)
 
         # shut it down, wait until it is gone
-        m.execute("pkill -ef cockpit-ws")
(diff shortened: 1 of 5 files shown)


martinpitt:
Was green, but rebased to pick up the pixel testing FUBAR fix.

allisonkarlitskaya (commented):
Anything wrong with `-9` and then we don't have to wait?

martinpitt:
> Anything wrong with -9 and then we don't have to wait?

signal delivery is not synchronous either way, so I'd still want to wait to prevent races. It was obviously working without the wait "most of the time", but this doesn't hurt IMHO.

allisonkarlitskaya (approved):

PR #17628, opened 2022-08-08
kdump: Use close with cockpit.file, not remove

mvollmer:
There is no remove, and the intention is to close the watch channel.

### DIFF
--- a/pkg/kdump/config-client.js
+++ b/pkg/kdump/config-client.js
@@ -47,7 +47,7 @@ export class ConfigFile {
 
     close() {
         if (this._fileHandle) {
-            this._fileHandle.remove();
+            this._fileHandle.close();
             this._fileHandle = undefined;
         }
     }


skazi0 (approved):
lgtm

pkg/kdump/config-client.js
```diff
@@ -47,7 +47,7 @@ export class ConfigFile {
 
     close() {
         if (this._fileHandle) {
-            this._fileHandle.remove();
+            this._fileHandle.close();
```
cockpituous, line 50: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17628-20220808-115321-9be51b56-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly, line 50: Looking at the [docs](https://cockpit-project.org/guide/latest/cockpit-file) cockpit.file("foo").watch() has a `remove` function.

As for example in our test `pkg/base1/test-file.js`:

```
      const file = cockpit.file(dir + "/foobar");
      const watch = file.watch(changed);
      watch.remove();
```

So is our code above wrong here, or does cockpit.file().close() also remove the watchers?

jelly: As in, should we assign `this._watchhandle = file.watch()` and `remove()` that?

skazi0: I thought it might have come from the `watch.remove()` but there's no history in git to prove it wrong/right.
I think `close()` has enough logic to take care of the watches: https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/cockpit.js#L3800

jelly (approved):

PR #17631, opened 2022-08-09
pybridge: Add http-stream2 channel

martinpitt:
This is sufficient to pass pkg/base1/test-http.js, but is still missing a lot:

 - [x] needs to not block other channels:
     *  [aiohttp](https://github.com/aio-libs/aiohttp/) would be nice, but it's an external module and we only want to use included batteries
     * we might be able to poll() the `response.fileno()` for at least asynchronously reading the response data, but that still leaves the connection handshake and sending the request; all of that can block as well
     * we might spawn `curl`, but that's ugly and not always installed either
     * :heavy_check_mark: our best bet might be to run the whole thing in a thread
  - [x] add smoke test with lots of parallel channels, to validate correct serialization in threading: also split out to PR #17639
  - [x] add support for Unix sockets (not covered by unit tests)
  - [x] correct problem key for SSL certificate validation errors ("unknown-hostkey")
  - [x] add support for TLS options (not covered by unit tests)
  - [x] ~check if C bridge supports TLS on Unix socket~ we'd rather not support this weird case, so forbid + unit test
  - [ ] ~Understand C bridge's `Keep-Alive:` header handling, write a test for it, implement in py bridge~ (see below)
  - [ ] ~Understand C bridge's connection sharing with serial requests and with server-side timeouts/closes; write a test for it, implement in py bridge~ (disabled the feature, see below)
  - [x] Fix POST data; validate with cockpit-podman
  - [x] Goes on top of PR #17702

## manual tests for async stream

These require the hack from PR #17599 to run the py bridge interactively.

Interleaved spawn and http channel with C bridge:

```
(printf '\n{ "command": "open", "channel": "h", "payload": "http-stream2", "address": "piware.de", "port": 443, "method": "GET", "path": "/tmp/x.txt", "tls": {} }\n---\n\n{ "command": "done", "channel": "h" }\n---\n\n{ "command": "open", "channel": "s", "payload": "stream", "spawn": ["ping", "-c3", "piware.de"] }\n---\n'; sleep 5;) | cockpit-bridge --interact=---
```
This works fine. If you change the port to "444", the `h` channel hangs forever (my server doesn't send REJECT, the firewall just blocks it). But the `s` channel responds fine.

A similar command with the py bridge:

```
make cockpit-bridge.pyz && (printf '\n{ "command": "open", "channel": "h", "payload": "http-stream2", "address": "piware.de", "port": 443, "method": "GET", "path": "/tmp/x.txt", "tls": {} }\n---\n\n{ "command": "done", "channel": "h" }\n---\n\n{ "command": "open", "channel": "s", "payload": "stream", "spawn": ["ping", "-c3", "piware.de"] }\n---\n'; sleep 5;) | ./cockpit-bridge.pyz
```
~That works reasonably as well, and the blocking isn't visible here. But if you change port to 444, it currently hangs in "creating new connection for None to piware.de:444; tls: {}", as that never returns and has no timeout either.~ This works similar to the C bridge now.

## manual test for Unix socket

Server:
```
python3 -m http.server & socat unix-listen:/tmp/sock,,reuseaddr,fork tcp-connect:localhost:8000
```
C bridge:
```
(printf '\n{ "command": "open", "channel": "m", "payload": "http-stream2", "unix": "/tmp/sock", "method": "GET", "path": "/.bashrc" }\n---\n\n{ "command": "done", "channel": "m" }\n---\n'; sleep 2;) | cockpit-bridge --interact=---
```
py bridge:
```
make cockpit-bridge.pyz && (printf '\n{ "command": "open", "channel": "m", "payload": "http-stream2", "unix": "/tmp/sock", "method": "GET", "path": "/.bashrc" }\n---\n\n{ "command": "done", "channel": "m" }\n---\n'; sleep 2;) | ./cockpit-bridge.pyz
```

## manual tests for TLS

Simple, no options (using publicly known certificate):
```
(printf '\n{ "command": "open", "channel": "h", "payload": "http-stream2", "address": "piware.de", "port": 443, "method": "GET", "path": "/tmp/x.txt", "tls": {} }\n---\n'; sleep 2;) | ...
```

Custom certificate, failing:
```
(printf '\n{ "command": "open", "channel": "h", "payload": "http-stream2", "address": "cockpit-11.apps.cnfdb2.e2e.bos.redhat.com", "port": 443, "method": "GET", "path": "/images/", "tls": {} }\n---\n'; sleep 2;) | ...
```
C bridge fails with `"problem":"unknown-hostkey"`, current py bridge ~with "not-found"~ fixed, same behaviour now.

Custom certificate with CA:
```
(printf '\n{ "command": "open", "channel": "h", "payload": "http-stream2", "address": "cockpit-11.apps.cnfdb2.e2e.bos.redhat.com", "port": 443, "method": "GET", "path": "/images/", "tls": { "authority": { "file": "/home/martin/upstream/bots/images/files/ca.pem" } } }\n---\n\n{ "command": "done", "channel": "m" }\n---\n'; sleep 5;)
```

Likewise, disabling CA validation:
```
(printf '\n{ "command": "open", "channel": "h", "payload": "http-stream2", "address": "cockpit-11.apps.cnfdb2.e2e.bos.redhat.com", "port": 443, "method": "GET", "path": "/images/", "tls": { "validate": false } }\n---\n\n{ "command": "done", "channel": "m" }\n---\n'; sleep 5;)
```
These two fail on the C bridge, it never responds with anything. These work on the py bridge (but no automatic test.)

### DIFF
--- a/pkg/lib/qunit-tests.js
+++ b/pkg/lib/qunit-tests.js
@@ -30,4 +30,9 @@ require("./qunit-config.js");
 
 require("qunit/qunit/qunit.css");
 
+QUnit.mock_info = async key => {
+    const response = await fetch(`http://${window.location.hostname}:${window.location.port}/mock/info`);
+    return (await response.json())[key];
+};
+
 export default QUnit;
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -20,6 +20,7 @@ pyTESTS_PASSING = \
 	dist/base1/test-format.html \
 	dist/base1/test-framed-cache.html \
 	dist/base1/test-framed.html \
+	dist/base1/test-http.html \
 	dist/base1/test-journal-renderer.html \
 	dist/base1/test-locale.html \
 	dist/base1/test-location.html \
@@ -38,7 +39,6 @@ pyTESTS_PASSING = \
 	$(NULL)
 
 pyTESTS_FAILING = \
-	dist/base1/test-http.html \
 	dist/base1/test-websocket.html \
 	$(NULL)
 
@@ -62,6 +62,7 @@ PYTHON_BRIDGE_FILES = \
 	src/cockpit/channels/dbus.py \
 	src/cockpit/channels/dbus_internal.py \
 	src/cockpit/channels/filesystem.py \
+	src/cockpit/channels/http.py \
 	src/cockpit/channels/metrics.py \
 	src/cockpit/channels/packages.py \
 	src/cockpit/channels/stream.py \
--- a/src/cockpit/channels/__init__.py
+++ b/src/cockpit/channels/__init__.py
@@ -18,6 +18,7 @@
 from .dbus import DBusChannel
 from .dbus_internal import DBusInternalChannel
 from .filesystem import FsListChannel, FsReadChannel, FsReplaceChannel, FsWatchChannel
+from .http import HttpChannel
 from .metrics import MetricsChannel
 from .packages import PackagesChannel
 from .stream import StreamChannel
@@ -32,6 +33,7 @@ CHANNEL_TYPES = [
     FsReadChannel,
     FsReplaceChannel,
     FsWatchChannel,
+    HttpChannel,
     MetricsChannel,
     NullChannel,
     PackagesChannel,
--- a/src/cockpit/channels/http.py
+++ b/src/cockpit/channels/http.py
@@ -1,0 +1,154 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2022 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import asyncio
+import http.client
+import logging
+import ssl
+import socket
+import threading
+
+from ..channel import Channel
+
+logger = logging.getLogger(__name__)
+
+
+class HttpChannel(Channel):
+    payload = 'http-stream2'
+
+    def create_connection(self):
+        opt_address = self.options.get('address') or 'localhost'
+        opt_port = self.options.get('port')
+        opt_unix = self.options.get('unix')
+        opt_tls = self.options.get('tls')
+        logger.debug('connecting to %s:%s; tls: %s', opt_address, opt_port or opt_unix, opt_tls)
+
+        if opt_tls is not None:
+            if 'authority' in opt_tls:
+                if 'data' in opt_tls['authority']:
+                    context = ssl.create_default_context(cadata=opt_tls['authority']['data'])
+                else:
+                    context = ssl.create_default_context(cafile=opt_tls['authority']['file'])
+            else:
+                context = ssl.create_default_context()
+
+            if 'validate' in opt_tls and not opt_tls['validate']:
+                context.check_hostname = False
+                context.verify_mode = ssl.VerifyMode.CERT_NONE
+
+            connection = http.client.HTTPSConnection(opt_address, opt_port, context=context)
+        else:
+            connection = http.client.HTTPConnection(opt_address, opt_port)
+
+        try:
+            if opt_unix:
+                # create the connection's socket so that it won't call .connect() internally (which only supports TCP)
+                connection.sock = socket.socket(socket.AF_UNIX)
+                connection.sock.connect(opt_unix)
+            else:
+                # explicitly call connect(), so that we can do proper error handling
+                connection.connect()
+        except (OSError, IOError) as e:
+            logger.error('Failed to open %s:%s: %s %s', opt_address, opt_port or opt_unix, type(e), e)
+            problem = 'unknown-hostkey' if isinstance(e, ssl.SSLCertVerificationError) else 'not-found'
+            self.close(problem=problem, message=str(e))
+            return None
+
+        return connection
+
+    def read_send_response(self, response):
+        '''Completely read the response and send it to the channel'''
+
+        while True:
+            # we want to stream data blocks as soon as they come in
+            block = response.read1(4096)
+            if not block:
+                logger.debug('reading response done')
+                # this returns immediately and does not read anything more, but updates the http.client's
+                # internal state machine to "response done"
+                block = response.read()
+                assert block == b''
+                break
+            logger.debug('read block of size %i', len(block))
+            self.loop.call_soon_threadsafe(self.send_data, block)
+
+    def parse_headers(self, http_msg):
+        headers = dict(http_msg)
+        remove = ['Connection', 'Transfer-Encoding']
+        if self.options.get('binary'):
+            remove = ['Content-Length', 'Range']
+        for h in remove:
+            try:
+                del headers[h]
+            except KeyError:
+                pass
+        return headers
+
+    def request(self):
+        connection = self.create_connection()
+        if not connection:
+            # make_connection does the error reporting
+            return
+
+        connection.request(self.options.get('method'),
+                           self.options.get('path'),
+                           headers=self.options.get('headers') or {},
+                           body=self.body)
+        response = connection.getresponse()
+        self.loop.call_soon_threadsafe(lambda: self.send_control(
+            command='response', status=response.status, reason=response.reason,
+            headers=self.parse_headers(response.headers)))
+        self.read_send_response(response)
+        connection.close()
+
+        self.loop.call_soon_threadsafe(self.done)
+        self.loop.call_soon_threadsafe(self.close)
+        logger.debug('closed')
+
+    def do_open(self, options):
+        logger.debug('open %s', options)
+        if not options.get('method'):
+            self.close(problem='protocol-error', message='missing or empty "method" field in HTTP stream request')
+            return
+        if options.get('path') is None:
+            self.close(problem='protocol-error', message='missing "path" field in HTTP stream request')
+            return
+        if options.get('tls') is not None and options.get('unix'):
+            self.close(problem='protocol-error', message='TLS on Unix socket is not supported')
+            return
+        if options.get('connection') is not None:
+            self.close(problem='protocol-error', message='connection sharing is not implemented on this bridge')
+            return
+
+        opt_port = options.get('port')
+        opt_unix = options.get('unix')
+        if opt_port is None and opt_unix is None:
+            self.close(problem='protocol-error', message='no "port" or "unix" option for channel')
+            return
+        if opt_port is not None and opt_unix is not None:
+            self.close(problem='protocol-error', message='cannot specify both "port" and "unix" options')
+            return
+
+        self.options = options
+        self.body = b''
+
+    def do_data(self, data):
+        self.body += data
+
+    def do_done(self):
+        self.loop = asyncio.get_event_loop()
+        threading.Thread(target=self.request, daemon=True).start()
--- a/pkg/base1/test-echo.js
+++ b/pkg/base1/test-echo.js
@@ -82,10 +82,8 @@ QUnit.test("binary", function (assert) {
 QUnit.test("fence", async assert => {
     const done = assert.async();
 
-    const response = await fetch(`http://${window.location.hostname}:${window.location.port}/mock/info`);
-    const info = await response.json();
     // This is implemented in the C bridge, but not in Python.
-    if (info.bridge == "cockpit-bridge.pyz") {
+    if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz") {
         assert.ok(true, "skipping on python bridge, not implemented");
         done();
         return;
--- a/pkg/base1/test-http.js
+++ b/pkg/base1/test-http.js
@@ -260,6 +260,15 @@ QUnit.test("http promise recursive", assert => {
 QUnit.test("http keep alive", async assert => {
     assert.expect(1);
 
+    // connection sharing is not implemented in the pybridge
+    if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz") {
+        assert.rejects(
+            cockpit.http({ port: test_server.port, connection: "one" }).get("/mock/connection"),
+            ex => ex.problem == "protocol-error" && ex.status == undefined,
+            "rejects connection option on python bridge");
+        return;
+    }
+
     /*
      * The /mock/connection handler returns an identifier that changes if
      * a different connection is used.
@@ -272,6 +281,12 @@ QUnit.test("http keep alive", async assert => {
 QUnit.test("http connection different", async assert => {
     assert.expect(1);
 
+    // connection sharing is not implemented in the pybridge
+    if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz") {
+        assert.ok(true);
+        return;
+    }
+
     /*
      * The /mock/connection handler returns an identifier that changes if
      * a different connection is used.
@@ -283,6 +298,12 @@ QUnit.test("http connection different", async assert => {
 
 QUnit.test("http connection without address", async assert => {
     assert.expect(1);
+
+    // connection sharing is not implemented in the pybridge
+    if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz") {
+        assert.ok(true);
+        return;
+    }
 
     // Able to reuse connection client info and not specify address again.
     const first = await cockpit.http({ port: test_server.port, connection: "one" }).get("/mock/connection");
@@ -333,7 +354,7 @@ QUnit.test("HEAD method", assert => {
             .finally(done);
 });
 
-QUnit.test("wrong options", assert => {
+QUnit.test("wrong options", async assert => {
     assert.rejects(
         cockpit.http({}).get("/"),
         // unfortunately cockpit.js does not propagate the detailed error message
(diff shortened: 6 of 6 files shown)


martinpitt:
@allisonkarlitskaya : This is getting to a state where I'm reasonably happy with the functionality and robustness. I split out the bits that affect the C bridge to PR #17639; after that lands, I'll rebase and squash this one, but so far I found it nice to keep the evolution in separate commits. I think it makes sense for you to have a first look. I'm sure we can beautify the code here and there, too! :wink:

src/Makefile.am
```diff
@@ -62,6 +62,7 @@ PYTHON_BRIDGE_FILES = \
 	src/cockpit/channels/dbus.py \
 	src/cockpit/channels/dbus_internal.py \
 	src/cockpit/channels/filesystem.py \
+	src/cockpit/channels/http_stream2.py \
```
allisonkarlitskaya, line 65: My impulse would have been to call this `http.py`.  Compare that it's not `dbus_json3` or so.

src/cockpit/channels/__init__.py
```diff
@@ -18,6 +18,7 @@
 from .dbus import DBusChannel
 from .dbus_internal import DBusInternalChannel
 from .filesystem import FsListChannel, FsReadChannel, FsReplaceChannel, FsWatchChannel
+from .http_stream2 import HttpStream2Channel
```
allisonkarlitskaya, line 21: ditto, why not just `HttpChannel`?  `HttpStreamChannel` would probably also be fine too, but this number thing is just silly cruft.

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class HttpStream2Channel(Channel):
+    payload = 'http-stream2'
+    connections = {}
+
+    def get_connection(self, name):
+        if name and name in HttpStream2Channel.connections:
```
allisonkarlitskaya, line 35: In general, this "check and then access" pattern is typically considered unpythonic.

Either (my preference) use `.get()` (possibly with a walrus) or just do the lookup directly and catch the exception.

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+            logger.debug('reusing connection %s', name)
+            return HttpStream2Channel.connections[name]
+        return None
+
+    def create_connection(self):
+        connection = self.options.get('connection')
+        address = self.options.get('address') or "localhost"
```
allisonkarlitskaya, line 42: There are several schools of thoughts on when to use `'` and `"`, but I don't know of any which would inform this particular combination :)

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+        port = self.options.get('port')
+        tls = self.options.get('tls')
+        logger.debug('creating new connection for %s to %s:%s; tls: %s', connection, address, port, tls)
+        _cls = http.client.HTTPSConnection if tls is not None else http.client.HTTPConnection
+        con = _cls(address, port)
```
allisonkarlitskaya: I think I'd prefer to see this as an `if:`/`else:` block with direct calls to the respective constructors.

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+        try:
+            # explicitly call connect(), so that we can do proper error handling
+            con.connect()
+        except Exception as e:
```
allisonkarlitskaya, line 77: Please try to list the exceptions that you expect might happen here.  This wide net would catch (for example) even the case where `con` was `None` or the `connect()` method was not available.

If it would be totally unwieldy to do so, then at least add a comment about why you need to catch `Exception`.

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+            logger.error("Failed to open %s:%s [connection: %s]: %s", address, port, connection, e)
+            self.close(problem='not-found', message=str(e))
+            return None
+
+        if connection:
```
allisonkarlitskaya, line 82: `... is not None`.

allisonkarlitskaya: Also: too bad about the variable names.  I get that you want this to be the same name as what was given in the JSON options object, but I wonder if it might make things a bit clearer to rename it to `connection_name` or even `persistent_id` or so (or some other variation) and use the `connection` variable name for the actual connection object...

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+        self.close()
+        logger.debug('closed')
+
+    def do_open(self, options):
+        logger.debug('open %s', options)
+        self.is_binary = options.get('binary')
+        if not options.get('method'):
+            self.close(problem='protocol-error', message='missing "method" field in HTTP stream request')
+            return
+        if not options.get('path'):
```
allisonkarlitskaya, line 138: `...is None`

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+            self.close(problem='protocol-error', message='missing "path" field in HTTP stream request')
+            return
+
+        self.conn = self.get_connection(options.get('connection'))
+        if not self.conn:
+            # these are not required/ignored for shared connections
+            port = options.get('port')
+            unix = options.get('unix')
+            if not port and not unix:
```
allisonkarlitskaya, line 150: `if port is None and unix is None`

src/cockpit/channels/http_stream2.py
```diff
@@ -0,0 +1,130 @@
...
+                self.close(problem='protocol-error', message='no "port" or "unix" option for channel')
+                return
+            if port and unix:
```
allisonkarlitskaya, line 153: `if port is not None and unix is not None`

etc etc :)

allisonkarlitskaya (changes requested):
Really cool first pass!

The way this is written, it looks like two separate instances of the channel could both take reference on the same httpclient object.  Is that intended?  If not, it might make sense to check for the case of the channel being in use and throw an error.

You might also handle it by removing the object from the cache and depositing it back again when you're done, but this seems potentially messy.  You could end up with multiple connection objects with the same name, in this case.

Also missing, of course, is some kind of expiry for objects in the cache.

src/cockpit/channels/http_stream2.py
martinpitt: It's the "developer is torn between using `"` everywhere else, and trying to stick to your `'` convention in the vicinity of channels/*.py.

Current main uses `'` 74 times and `"` 357 times -- indeed let's make this consistent.

martinpitt:
@allisonkarlitskaya : FWIW, I agree to all your comments, thanks! Just for expectation management, I'll work on #17639 first.

src/cockpit/channels/http_stream2.py
martinpitt: I renamed the options.get aliases consistently to `opt_<name>`, and renamed `con` to `connection`. But I don't understand why `if connection is not None:` is any better than `if connection:`? This is an object. I do get it for "port" and "unix" (see below), and have changed these.

src/Makefile.am
martinpitt: Renamed to http.py.

src/cockpit/channels/__init__.py
martinpitt: Renamed.

src/cockpit/channels/http_stream2.py
martinpitt: Used try/except KeyError.

martinpitt: Remember, no walruses yet -- RHEL 8!

src/cockpit/channels/http_stream2.py
martinpitt: Changed everything to '

src/cockpit/channels/http_stream2.py
martinpitt: Indeed, that was just a leftover from sloppy initial experiments. Done.

src/cockpit/channels/http_stream2.py
martinpitt: Done.

src/cockpit/channels/http_stream2.py
martinpitt: Done.

src/cockpit/channels/http_stream2.py
martinpitt: Done

src/cockpit/channels/http_stream2.py
martinpitt: This was already rewritten in a fixup. I squashed everything now, so this thing is gone.

martinpitt:
> The way this is written, it looks like two separate instances of the channel could both take reference on the same httpclient object. Is that intended?

Yes, very much -- every cockpit.http() channel is its own separate thing, but the connection sharing happens between all of them. We have an explicit QUnit test "http keep alive" for this. That's why I made the map a class member instead of a per-instance member. This corresponds to the global `static GHashTable *clients` in src/bridge/cockpithttpstream.c.

> Also missing, of course, is some kind of expiry for objects in the cache.

Right, that's the major missing bit. The C bridge handles the clients in a much more elaborate way, with parsing the Keep-Alive: header, and (I think) listening to "close" signals from the HTTP streams. I just don't quite see where it invalidates the cache entries. TBH I don't understand the whole connection sharing in the C bridge enough yet -- this requires some more unit tests that check what happens if you share a connection, use it multiple times *serially* (not in parallel), and what happens if the server-side of a connection times out/closes/fails on a request.

My current gut feeling about this is: *if* a channel opening has a `connection` name, *and* the connection was taken from the cache, *and* it fails, *then* delete the cached object and re-create it from scratch.

I also don't understand the role of the Keep-Alive: header. I think that's unrelated to the connection name caching, but also somehow needs to be handled.

martinpitt:
I am trying to make sense of the connection sharing stuff. I tried this:
```
(printf '\n{ "command": "open", "channel": "h1", "payload": "http-stream2", "connection": "c1", "address": "piware.de", "port": 443, "method": "GET", "path": "/tmp/x.txt", "tls": {} }\n---\n\n{ "command": "done", "channel": "h1" }\n---\n'; sleep 1; printf '\n{ "command": "open", "channel": "h2", "payload": "http-stream2", "connection": "c1", "method": "GET", "path": "/tmp/hello.txt"}\n---\n\n{ "command": "done", "channel": "h2" }\n---\n'; sleep 2) | ./cockpit-bridge --interact=---
```

The first x.txt gets retrieved and streamed fine, but the second fails with
```
{"message":"no \"port\" or \"unix\" or other address option for channel","command":"close","channel":"h2","problem":"protocol-error"}
```
I thought that was the whole *point* of this? that should correspond to the "http connection without address" unit test, but apparently that does something slightly different.

However, I wonder if it's even worth the effort: this connection sharing makes everything a lot more complicated (see above), and *none* of the cockpit* projects actually use it. c-composer and c-podman both have a *single* cockpit.http() call to the local weldr/podman Unix socket, and that's it.

So I propose to not implement this for now, raise a protocol-error when specifying the connection argument, and adjust the tests accordingly.

martinpitt:
@allisonkarlitskaya: Ready for another review round. Of course I'll squash the fixup eventually, but for discussion it seems better to keep it separate for now.

martinpitt:
I tested c-podman against the pybridge and this branch. Most things work fine (to my surprise), like detecting containers, images, and event streaming. The only thing that fails is creating a new container.

On the C bridge I see:
```
XXX connection.call {"method":"GET","path":"/v1.12/libpod/images/registry.fedoraproject.org/fedora:latest/exists","body":"","params":{}} rest.js:51:20
XXX connection.call success <empty string> rest.js:54:32
XXX connection.call {"method":"POST","path":"/v1.12/libpod/containers/create","body":"{\"image\":\"registry.fedoraproject.org/fedora:latest\",\"name\":\"pedantic_engelbart\",\"command\":[\"/bin/bash\"],\"terminal\":true}","params":{}} rest.js:51:20
XXX connection.call success {"Id":"190181d7f5c80a02b035e59c7966c202022444ecffcbe5a0394fc03521c94a9f","Warnings":[]}
rest.js:54:32
XXX connection.call {"method":"POST","path":"/v1.12/libpod/containers/190181d7f5c80a02b035e59c7966c202022444ecffcbe5a0394fc03521c94a9f/start","body":"","params":{}} rest.js:51:20
XXX connection.call success <empty string> rest.js:54:32
```

while on the py bridge, the /create call returns with an empty string instead of the `{Id, Warnings}` JSON reply. Plausibly because that's an operation which just takes a while, but most others are fast; or because it's a POST instead of a GET. But what really makes me wonder are the `body` and `params` attributes to the http call -- these are not documented, nor do I see any mention of it in the C bridge. Let's replicate this with a direct bridge call and investigate.

src/cockpit/channels/http.py
```diff
@@ -0,0 +1,164 @@
...
+import threading
+
+from ..channel import Channel
+
+logger = logging.getLogger(__name__)
+
+
+class HttpChannel(Channel):
+    payload = 'http-stream2'
+    connections = {}
```
allisonkarlitskaya: Didn't you say you wanted to drop connection caching?

src/cockpit/channels/http.py
```diff
@@ -0,0 +1,164 @@
...
+            remove = ['Content-Length', 'Range']
+        for h in remove:
+            try:
+                del headers[h]
+            except KeyError:
+                pass
+        return headers
+
+    def request(self):
+        if not self.connection:
```
allisonkarlitskaya: This is happening in a thread?  This looks extremely unsafe.

allisonkarlitskaya: This is happening in a thread?  This looks extremely unsafe.

allisonkarlitskaya (changes requested):
https://pypi.org/project/aiohttp/ . We wat to only use the included

s/wat/want/


I'd like to discuss the big conceptual question here before continuing.

src/cockpit/channels/http.py
martinpitt: I fixed the "wat" typo. The connection sharing was dropped in a FIXUP, I squashed it now as we are in agreement about not implementing it.

src/cockpit/channels/http.py
martinpitt: Dropped.

pkg/lib/qunit-tests.js
```diff
@@ -30,4 +30,9 @@ require("./qunit-config.js");
 
 require("qunit/qunit/qunit.css");
 
+QUnit.mock_info = async () => {
```
allisonkarlitskaya, line 33: Geschmackssache: why not have it take the name of the attribute you want to check as the first argument and do the dereference for you.  That way you can do like QUnit.mock_info('slow-tests') or whatever.

src/cockpit/channels/http.py
```diff
@@ -0,0 +1,153 @@
...
+        if not connection:
+            # make_connection does the error reporting
+            return
+
+        connection.request(self.options.get('method'),
+                           self.options.get('path'),
+                           headers=self.options.get('headers') or {},
+                           body=self.body)
+        response = connection.getresponse()
+        self.loop.call_soon_threadsafe(lambda: self.send_control(
```
allisonkarlitskaya, line 112: Are `.call_soon_threadsafe()` calls guaranteed to be well-ordered when they arrive?

src/cockpit/channels/http.py
```diff
@@ -0,0 +1,153 @@
...
+            block = response.read1(4096)
+            if not block:
+                logger.debug('reading response done')
+                # this returns immediately and does not read anything more, but updates the http.client's
+                # internal state machine to "response done"
+                block = response.read()
+                assert block == b''
+                break
+            logger.debug('read block of size %i', len(block))
+            self.loop.call_soon_threadsafe(self.send_data, block)
```
allisonkarlitskaya, line 87: So funny, enough, I *think* that calling `self.send_data()` is probably already currently threadsafe.  It basically ends up boiling down to a single write call on the underlying transport.

I don't want to guarantee that it would always be that way, but this shunting-over-to-the-main-thread on the main data path here is probably quite wasteful...

This needs a bit of a think.  I've already been contemplating moving away from Python's internal protocol/transport abstraction to a self-rolled thing that we have more control over.  This would remove the asyncstdio hackery and give us a chance to increase performance, and perhaps also make guarantees about threadsafety in places like this...

So for now: please keep it like it is :)

src/cockpit/channels/http.py
```diff
@@ -0,0 +1,153 @@
...
+
+        connection.request(self.options.get('method'),
+                           self.options.get('path'),
+                           headers=self.options.get('headers') or {},
+                           body=self.body)
+        response = connection.getresponse()
+        self.loop.call_soon_threadsafe(lambda: self.send_control(
+            command='response', status=response.status, reason=response.reason,
+            headers=self.parse_headers(response.headers)))
+        self.read_send_response(response)
```
allisonkarlitskaya, line 115: Do we need to close the response object or something in order to make sure the connection goes away properly?  Or is that implicit in reading it to EOF?

allisonkarlitskaya (approved):
A cool 120 lines.  Not bad.

src/cockpit/channels/http.py
martinpitt: I indeed started with .send_data(), with a big "XXX TODO" item that this didn't smell thread-safe. That's when I wrote the parallel torture unit test -- I fully expected that to break, but I wasn't actually able to. But that only raced against other http channel writes, and if that send_data() call is a single write that would explain why that worked. However, I feel very nervous to assume that *every* channel does "atomic" writes -- that smells like hard-to-debug problems later one.

So agreed that for now I feel better with keeping it as-is.

src/cockpit/channels/http.py
martinpitt: The [call_soon()](https://docs.python.org/3/library/asyncio-eventloop.html?highlight=call%20soon%20threadsafe#asyncio.loop.call_soon) docs say "Callbacks are called in the order in which they are registered", and c_s_t() says "A thread-safe variant of call_soon()", which I interpret as "the "output order is the same as the input order". If that weren't so, this API would be useless at least for our cases indeed. But it at least survived the "highly parallel torture test" as well as c-podman.

src/cockpit/channels/http.py
martinpitt: [HTTPResponse](https://docs.python.org/3/library/http.client.html#httpresponse-objects) does not have any kind of .close()/.cleanup() etc. In Python I'd expect the `connection` object to close itself when it goes out of scope, but I added an explicit `connection.close()` now just to be on the safe side.

pkg/lib/qunit-tests.js
martinpitt: I have no strong opinion either way. It's slighly nicer because it avoids the extra pair of () around the invocation. At the price of having to turn a .bridge identifier into a string. But that's fine.

allisonkarlitskaya (approved):
Big +1 :+1:

Thanks!!

PR #17632, opened 2022-08-09
storage: Remove some pixel tests for complex tables

mvollmer:
The tables can change layout completely "at random".  When the width
of a PID changes, for example, the Command column might wrap
differently and increase the height of the whole dialog.


martinpitt (approved):
It's with a sad eye, but I think it's the right thing to do. pixel tests should be a tool, not a nuisance. Danke!

mvollmer:
> pixel tests should be a tool, not a nuisance.

Yeah... I can't think of any good options.  We can try to fix the table layout, but that would mean dictating UI decisions just to make testing work. Or we could insert fake data into a table just before pixel testing, but that seems brittle and racy.

PR #17633, opened 2022-08-09
pf: Fix breadcrumb link cursor

garrett:
Breadcrumb links should have the correct pointing hand cursor.

PatternFly requires a "to" attribute for an actual link, but we use some
funky onClick JS for navigating and override it with a className.

Therefore, instead of having a proper `<a href="...">` being rendered, we need
to override the link. This is a problem with a (correct) assumption in PF
and our (incorrect) way of not using links (but using JavaScript) for
linking.

Nevertheless, Cockpit needs to be adapted for this to work as expected.

---

Screenshot of the issue:

image

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -227,3 +227,18 @@ svg {
     position: absolute;
     width: 100%;
  }
+
+// Breadcrumb links should have the correct pointing hand cursor.
+//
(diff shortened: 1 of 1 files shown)


garrett:
I didn't open and link a PF upstream issue, as the problem is ours using the component incorrect, not PatternFly, which makes a (correct) assumption. However, we need to override the PF widget to work as expected, so this is still the correct place. It's a rare exception where we don't need an upstream PF link.

jelly (approved):

PR #17639, opened 2022-08-10
bridge: Disallow TLS connections to Unix socket, http parallel stress test

martinpitt:
We don't test or use TLS connections in channels, and a lot of TLS modes
are even just plain broken with the C bridge.

It's hard to imagine for what we'd ever need TLS to connect to a local
Unix socket (the main use cases are REST calls to docker, podman, or
weldr). Explicitly disallow this, so that we don't have to implement all
these corner cases in the Python bridge.

----

Also add a parallel stress test for http-stream2. This will be useful for validating the py bridge.

For the record: We use cockpit.http() exactly twice in all our projects: in [podman](https://github.com/cockpit-project/cockpit-podman/blob/main/src/rest.js#L18) (connect to podman Unix socket) and [composer](https://github.com/osbuild/cockpit-composer/blob/main/core/composer.js#L4) (connect to weldr Unix socket).

### DIFF
--- a/src/bridge/cockpitconnect.c
+++ b/src/bridge/cockpitconnect.c
@@ -220,6 +220,13 @@ parse_address (CockpitChannel *channel,
       goto out;
     }
 
+  if (unix_path && json_object_has_member (options, "tls"))
+    {
+      cockpit_channel_fail (channel, "protocol-error",
+                            "TLS on Unix socket is not supported");
+      goto out;
+    }
+
   if (port != G_MAXINT64 && unix_path)
     {
       cockpit_channel_fail (channel, "protocol-error", "cannot specify both \"port\" and \"unix\" options");
@@ -654,4 +661,3 @@ cockpit_connect_parse_stream (CockpitChannel *channel)
 
   return connectable;
 }
-
--- a/pkg/base1/test-http.js
+++ b/pkg/base1/test-http.js
@@ -342,6 +342,11 @@ QUnit.test("wrong options", assert => {
         cockpit.http({ port: 1234, unix: "/nonexisting/socket" }).get("/"),
         ex => ex.problem == "protocol-error" && ex.status == undefined,
         "rejects request with both port and unix option");
+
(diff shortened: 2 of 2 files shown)


martinpitt:
> a lot of TLS modes are even just plain broken with the C bridge.

See the manual tests in PR #17631

martinpitt:
Back to draft.. The [unit test](https://github.com/cockpit-project/cockpit/runs/7760767009?check_suite_focus=true) does not work under valgrind, it's just too slow and eventually crashes/times out. I didn't find or see an obvious way how to check in a JS unit test whether we are running under valgrind. This code is running in the browser, thus it can't look at `$COCKPIT_SKIP_SLOW_TESTS` or check LD_PRELOAD. Do we need to export this in the test-server, so that the unit test can ask it via a http request?

src/bridge/cockpitconnect.c
```diff
@@ -220,6 +220,13 @@ parse_address (CockpitChannel *channel,
       goto out;
     }
 
+  if (unix_path && json_object_has_member (options, "tls"))
```
allisonkarlitskaya, line 223: can we think of a way to do this without modifying the C?

martinpitt: We could if we could make JS unit tests specific to the bridge. My current idea: add some /info endpoint to test-server which gives you a JSON object like `{ "slow_tests": true, "bridge": "cockpit-bridge.pyz" }`. Then unit tests can query that and skip tests under valgrind or conditionalize expectations on the bridge. This would also allow us to re-enable the barrier test which you disabled when landing the pybridge.

allisonkarlitskaya: This is perfect.

martinpitt: I implemented that. But I can't land it separately, as querying the /mock/info path requires... *drumroll* cockpit.http() :grin: So I'll land it all in #17631.

PR #17640, opened 2022-08-10
cockpit.js: Support different URL root for the websocket

martinpitt:
We eventually want to support running cockpit-ws behind https://www.3scale.net/.
This requires two different API paths: `/api/<service>-<port>/` for HTTP
traffic, and `/wss/<service>-<port>/` for websocket connections.

cockpit.js needs to know which URL/path to contact for the websocket, so this
cannot be done with path rewriting on the server side only, or even with
multiple UrlRoot= values.

Thus introduce support for reading a `websocket-root` meta tag from the 
webserver, similar to the existing `url-root` one. Nothing sets that tag 
right now, as we are still not entirely sure how we want to configure
this. See https://github.com/cockpit-project/cockpit/pull/17473 how all 
of this fits together.

This is the only part required on the client side (which is much harder
to change and get into distros than the ws container), so land this now.
It is harmless, and easy enough to revert if we won't use it after all.

 - [x] Re-test scenarios with COPR from this PR, to ensure that it suffices on the client side

### DIFF
--- a/doc/man/cockpit.conf.xml
+++ b/doc/man/cockpit.conf.xml
@@ -49,50 +49,50 @@
   </refsect1>
 
   <refsect1 id="cockpit-conf-webservice">
-	  <title>WebService</title>
-	  <variablelist>
-	    <varlistentry>
-	      <term><option>Origins</option></term>
-	      <listitem>
-                <para>By default cockpit will not accept crossdomain websocket connections. Use this
-                  setting to allow access from alternate domains. Origins should include scheme, host
-                  and port, if necessary.</para>
+  <title>WebService</title>
+  <variablelist>
+    <varlistentry>
+      <term><option>Origins</option></term>
+      <listitem>
+        <para>By default cockpit will not accept crossdomain websocket connections. Use this
+            setting to allow access from alternate domains. Origins should include scheme, host
+            and port, if necessary.</para>
 
-          <informalexample>
+        <informalexample>
 <programlisting language="ini">
 [WebService]
 Origins = https://somedomain1.com https://somedomain2.com:9090
 </programlisting>
-          </informalexample>
-        </listitem>
-      </varlistentry>
-	    <varlistentry>
-	      <term><option>ProtocolHeader</option></term>
-	      <listitem>
-                <para>Configure cockpit to look at the contents of this header to determine if a connection
-                  is using tls. This should only be used when cockpit is behind a reverse proxy, and care
-                  should be taken to make sure that incoming requests cannot set this header.</para>
-          <informalexample>
+        </informalexample>
+      </listitem>
+    </varlistentry>
+    <varlistentry>
+      <term><option>ProtocolHeader</option></term>
+      <listitem>
+        <para>Configure cockpit to look at the contents of this header to determine if a connection
+          is using tls. This should only be used when cockpit is behind a reverse proxy, and care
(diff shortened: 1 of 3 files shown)


doc/man/cockpit.conf.xml
```diff
@@ -149,6 +149,14 @@ ForwardedForHeader = X-Forwarded-For
           <code>/cockpit/</code> and <code>/cockpit+new/</code> are not.</para>
         </listitem>
       </varlistentry>
+      <varlistentry>
+        <term><option>WebsocketRoot</option></term>
```
marusak, line 153: I would put here `EXPERIMENTAL: Might change or be dropped in future releases`

marusak:
I am fine with this approach. I think we want to be prepared and have something working. Unless someone is strictly against, let's go with this.

doc/man/cockpit.conf.xml
martinpitt: Then maybe let's just not document it at all yet?

pkg/lib/cockpit.js
```diff
@@ -316,7 +316,8 @@ function calculate_url(suffix) {
     if (!suffix)
         suffix = "socket";
     const window_loc = window.location.toString();
-    let _url_root = url_root;
+    const meta_websocket_root = document.head.querySelector("meta[name='websocket-root']");
```
allisonkarlitskaya, line 319: Summary from IRC and meeting: we're gonna split these two lines into a separate commit which we merge today.  The rest of it will get compiled into our hacked up version of `cockpit-ws` that we used when testing the cloud side of this all.

Once we're sure that the entire solution is viable, we'll merge the rest.  If not, we can trivially back out these two lines, without any compatibility concerns.

martinpitt: Done. Review is fine and appreciated, but please don't merge this yet before I re-run my PoCs against the COPR of *this* PR. I want to b 100% sure that I wasn't missing anything that was necessary on the client side.

martinpitt:
I also included two auxiliary commits from PR #17641 which seem rather obvious and harmless, and will make future rebasing easier.

doc/man/cockpit.conf.xml
```diff
@@ -49,50 +49,50 @@
...
-	      <listitem>
-                <para>By default cockpit will not accept crossdomain websocket connections. Use this
-                  setting to allow access from alternate domains. Origins should include scheme, host
-                  and port, if necessary.</para>
+  <title>WebService</title>
+  <variablelist>
+    <varlistentry>
+      <term><option>Origins</option></term>
+      <listitem>
+      <para>By default cockpit will not accept crossdomain websocket connections. Use this
```
allisonkarlitskaya, line 57: Shouldn't `<para>` be indented here?  It is in the item you add below under `ProtocolHeader`.

Normally, I wouldn't nit about this, but since this commit is supposed to be about fixing indentation... ;)

martinpitt: You are right, and so should the `<informalexample>`. Fixed, thank you!

pkg/lib/cockpit.js
martinpitt: I did that testing now, and it's all :+1: Unblocked.

allisonkarlitskaya (approved):

PR #17641, opened 2022-08-10
ws: Support non-standard Authorization: header

martinpitt:
In some scenarios, the normal `Authorization:` header is already being
used for something else, like authenticating to console.redhat.com's
3scale proxy. So we can't use that for authentication between the login
page and the managed machine.

Introduce a new `AuthorizationHeader` cockpit.config option to use
another header instead. Recommended value is `X-Cockpit-Authorization`.

---

See PR #17473, which pulls this into a bigger context.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -21,6 +21,8 @@
             oauth.ErrorParam = "error_description";
     }
 
+    const authorization_header = environment.authorization_header || "Authorization";
+
     const fmt_re = /\$\{([^}]+)\}|\$([a-zA-Z0-9_]+)/g;
     function format(fmt /* ... */) {
         const args = Array.prototype.slice.call(arguments, 1);
@@ -422,7 +424,7 @@
             show("#login-wait-validating");
             const xhr = new XMLHttpRequest();
             xhr.open("GET", login_path, true);
-            xhr.setRequestHeader("Authorization", "Bearer " + token_val);
+            xhr.setRequestHeader(authorization_header, "Bearer " + token_val);
             xhr.onreadystatechange = function () {
                 if (xhr.readyState == 4) {
                     if (xhr.status == 200) {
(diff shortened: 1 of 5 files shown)


pkg/static/login.js
```diff
@@ -422,7 +424,7 @@
             show("#login-wait-validating");
             const xhr = new XMLHttpRequest();
             xhr.open("GET", login_path, true);
-            xhr.setRequestHeader("Authorization", "Bearer " + token_val);
+            xhr.setRequestHeader(authorization_header, "Bearer " + token_val);
```
cockpituous, line 427: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17641-20220810-135247-a1e6fd8c-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

allisonkarlitskaya (commented):
I'd actually prefer to drop our (questionable) use of this header entirely.  There isn't any compatibility issue here (since the `static` pages always ship with the `-ws`).

Why not just straight rename?  Am I missing something?

martinpitt:
> (questionable) use of this header

Why is it questionable? You authenticate to a web service, and the standard way to do that is the "Authorize:" header. Browsers, proxies, kerberos etc. usually know about and use that.

> Why not just straight rename?

I'm just very afraid of breaking valid setups. Right now you can drive the authentication with e.g. curl or CDP, not all authentication setups go via the login page, etc. If @mvollmer or you are convinced that it's safe to always use X-Cockpit-Authorize:, then so much the better of course. But it does feel like an API break.

I'd be happy to not advertise this much for now, and drop the documentation. Then we can use it as an internal/experimental feature, and only use it for the clouddot setup. This might give us some more flexibility?

martinpitt:
@marusak, @allisonkarlitskaya : I just realized that this PR doesn't affect the managed server side, only the ws pod. I updated [the PoC](https://issues.redhat.com/browse/COCKPIT-795) accordingly, that the managed server *only* pulls in the change from PR #17640, and validated that it still works as expected. So this PR is *not* urgent, and if there's any concern we can easily let it sit here for a while.

martinpitt:
Let's not land this for now, it's not time critical. We only need #17640. I added the first two commits from this PR there, and the third to #17473.

PR #17644, opened 2022-08-11
Run mypy on testlib.py

jelly:
This is just a PR to add mypy for testlib.py and make it pass. The typing in testlib.py is not complete yet.

In the future we can lint src/cockpit but that is actually quite trivial.


jelly:
@allisonkarlitskaya argggh so the issue is:

```
drwxr-xr-x 2 root root 4.0K Aug 16 15:46 builder
```

And builder tries to write `config.log` in the previous image it was:

```
builder@175cc8c22bf2:~$ ls -lh /home/
total 0
drwxr-xr-x 1 builder nogroup 0 Jul 24 22:06 builder
```

So yay debian changed!

jelly:
Proof mypy works https://github.com/cockpit-project/cockpit/runs/7878382609?check_suite_focus=true

containers/unit-tests/setup.sh
```diff
@@ -56,7 +56,8 @@ apt-get update
 apt-get install -y --no-install-recommends eatmydata
 DEBIAN_FRONTEND=noninteractive eatmydata apt-get install -y --no-install-recommends ${dependencies}
 
-adduser --system --gecos "Builder" builder
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=679746
```
allisonkarlitskaya, line 60: Thanks for tracking that down.  The bug link probably belongs in the commit message rather than in a comment, though.

test/static-code
```diff
@@ -39,6 +39,11 @@ test_pyvulture() {
         --ignore-decorators '@*'
 }
 
+test_mypy() {
+    command -v mypy >/dev/null || skip 'no mypy'
+    mypy --no-error-summary test/common/testlib.py
```
allisonkarlitskaya: Is there any danger we'll get py2 `mypy` here?  Does `python3 -m mypy` work, like for the other things?

src/cockpit/channel.py
```diff
@@ -30,12 +31,12 @@ def do_channel_data(self, channel, data):
 
 
 class Channel(Endpoint):
-    payload = None
-    restrictions = ()
+    payload = ""
+    restrictions: Dict[str, str] = {}
```
allisonkarlitskaya, line 35: The problem with using a dict here is that a dict is mutable.  It's considered bad form to use mutable values for defaults because if you modify it from one instance, it may not be obvious that you're effectively modifying all instances...

allisonkarlitskaya (changes requested):
I wonder if you want to first try getting this working with `systemd_ctypes` as a practice run?

test/static-code
jelly: Hmmmm it seems Debian does not provide python2 mypy at all https://packages.debian.org/buster/python3-mypy

test/static-code
```diff
@@ -39,6 +39,11 @@ test_pyvulture() {
         --ignore-decorators '@*'
 }
 
+test_mypy() {
```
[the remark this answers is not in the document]
jelly: I want to introduce mypy in steps if we need to lint all Python files, I need to do some more work.

src/cockpit/channel.py
jelly: Ah, but mypy is not happy about switching types so hmm a solution should be found or we make it `Optional[Dict[str, str]] = None`

Or of course `Set | Dict[str, str] = ()`

jelly:
> I wonder if you want to first try getting this working with `systemd_ctypes` as a practice run?

Sounds good, also an excellent time to figure out how the library stub files should be shipped :-)

test/static-code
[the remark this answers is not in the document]
jelly: As alternative we can grep and only include things we are interested in but maybe that's just not worth the effort...

src/cockpit/channel.py
supakeen: You can set only the type in the class:
```python
payload: str
restrictions: dict[str, str]
```

test/static-code
martinpitt: In python-dbusmock I call mypy through `python3 -m mypy`, which is generally the recommended and more robust way of invoking these tools. The only known exception that I know is flake8.

PR #17645, opened 2022-08-12
test: Robustify executed commands

martinpitt:
Most of the `a && b` constructs were wrong for our intent, as it will
silently succeed if `a` fails, and never run `b`. This can often cause
follow-up failures which are harder to investigate.

This was solved in in https://github.com/cockpit-project/bots/pull/3706
by running all Machine.execute() commands with `set -e`.

Replace `&&` with `;` where appropriate. There are still four places in
opportunistic cleanup commands which currently use `&&` in a broken way;
but keep them until we can get an agreement how to fix them.

Use Machine.write() to write files, instead of echo commands.

Drop the now redundant `set -e` from check-connection.

---

Similar to https://github.com/cockpit-project/cockpit-machines/pull/793 and https://github.com/cockpit-project/cockpit-podman/pull/1069

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1479,7 +1479,7 @@ class MachineCase(unittest.TestCase):
         self.addCleanup(m.execute, "find /var/lib/systemd/coredump -type f -delete")
 
         # temporary directory in the VM
-        self.addCleanup(m.execute, "if [ -d {0} ]; then findmnt --list --noheadings --output TARGET | grep ^{0} | xargs -r umount && rm -r {0}; fi".format(self.vm_tmpdir))
+        self.addCleanup(m.execute, "if [ -d {0} ]; then findmnt --list --noheadings --output TARGET | grep ^{0} | xargs -r umount; rm -r {0}; fi".format(self.vm_tmpdir))
 
         # users/groups/home dirs
         self.restore_file("/etc/passwd")
@@ -1960,7 +1960,7 @@ class MachineCase(unittest.TestCase):
             return
 
         backup = os.path.join(self.vm_tmpdir, path.replace('/', '_'))
-        self.machine.execute("mkdir -p %(vm_tmpdir)s && cp -a %(path)s/ %(backup)s/" % {
+        self.machine.execute("mkdir -p %(vm_tmpdir)s; cp -a %(path)s/ %(backup)s/" % {
             "vm_tmpdir": self.vm_tmpdir, "path": path, "backup": backup})
 
         if not reboot_safe:
@@ -1971,7 +1971,7 @@ class MachineCase(unittest.TestCase):
             self.addCleanup(self.machine.execute, post_restore_action)
 
         if reboot_safe:
-            self.addCleanup(self.machine.execute, "rm -rf {0} && mv {1} {0}".format(path, backup))
+            self.addCleanup(self.machine.execute, "rm -rf {0}; mv {1} {0}".format(path, backup))
         else:
             self.addCleanup(self.machine.execute, "umount -lf " + path)
 
@@ -1988,7 +1988,7 @@ class MachineCase(unittest.TestCase):
         exists = self.machine.execute("if test -e %s; then echo yes; fi" % path).strip() != ""
         if exists:
             backup = os.path.join(self.vm_tmpdir, path.replace('/', '_'))
-            self.machine.execute("mkdir -p %(vm_tmpdir)s && cp -a %(path)s %(backup)s" % {
+            self.machine.execute("mkdir -p %(vm_tmpdir)s; cp -a %(path)s %(backup)s" % {
                 "vm_tmpdir": self.vm_tmpdir, "path": path, "backup": backup})
             if post_restore_action:
                 self.addCleanup(self.machine.execute, post_restore_action)
--- a/test/verify/check-apps
+++ b/test/verify/check-apps
@@ -74,7 +74,7 @@ class TestApps(PackageCase):
             """,
             "/usr/share/app-info/icons/test/64x64/test.png": {"path": "/var/tmp/test.png"}})
         self.enableRepo()
-        self.machine.execute("systemctl stop packagekit && pkcon refresh force")
+        self.machine.execute("systemctl stop packagekit; pkcon refresh force")
         # ignore the corresponding journal entry
         self.allow_journal_messages("org.freedesktop.PackageKit.*org.freedesktop.DBus.Error.NoReply.*")
 
@@ -94,7 +94,8 @@ class TestApps(PackageCase):
                 '{ "config": { "appstream_data_packages": [ "appstream-data-test" ] } }')
 
         if urlroot != "":
-            m.execute(f'mkdir -p /etc/cockpit/ && echo "[WebService]\nUrlRoot={urlroot}" > /etc/cockpit/cockpit.conf')
+            m.execute("mkdir -p /etc/cockpit")
+            m.write("/etc/cockpit/cockpit.conf", f"[WebService]\nUrlRoot={urlroot}")
 
         m.start_cockpit()
         b.login_and_go("/apps", urlroot=urlroot)
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -58,8 +58,8 @@ class TestConnection(MachineCase):
         # on OSTree, also check with overlaid cockpit-ws rpm
         if m.ostree_image:
             def ws_start():
-                m.execute(r"""set -e;
-                    mkdir -p /etc/systemd/system/cockpit.service.d/ &&
+                m.execute(r"""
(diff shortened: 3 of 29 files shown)


test/verify/check-examples
```diff
@@ -101,7 +101,7 @@ class TestLongRunning(MachineCase):
         m.execute("mkdir -p ~admin/.local/share/cockpit")
         m.upload([os.path.join(EXAMPLES_DIR, "long-running-process")], "/home/admin/.local/share/cockpit/")
         # clean up after test failures
-        self.addCleanup(m.execute, "systemctl stop cockpit-longrunning.service 2>/dev/null && systemctl reset-failed cockpit-longrunning.service || true")
+        self.addCleanup(m.execute, "systemctl stop cockpit-longrunning.service 2>/dev/null || true; systemctl reset-failed cockpit-longrunning.service || true")
```
allisonkarlitskaya, line 104: I feel like this is a good example of something that's not really improved by this change.

Splitting it into two separate `addCleanup()` would be my first suggestion, but then I don't know if that function is run-first-first or run-last-first, and the fact that I'd need to check for it is already an argument against doing that.

I almost find myself wanting a version of m.execute() that accepts an array — but instead of being an argument vector, it would be an array of commands to run.  But that's gross too.

Why not make better use of `check=False` for the cases we don't care about failures for?  For that matter, what if we caused `check=False` to have the side effect of not running `set -e`?

martinpitt: I disagree about "not an improvement" -- by now I have found several instances where the second command was silently skipped in case of the first one failing. It's not fully clear whether it would break this particular case -- the stop command *definitively* fails in many cases (that was one of the force-pushes). It was never the intention to skip the reset-failed in that case.

I mostly care about having these opportunistic cleanups be simply greppable, `|| true` in this case. A `check=False` would work as well, it's just less flexible in the case of multiple commands. Not my prefered solution (unnecessary API complexity), but I don't have a strong opinion about it, i.e. no veto from me.

martinpitt:
@allisonkarlitskaya: Friendly re-review ping?

martinpitt:
Needs rebasing now, but let's first discuss with @allisonkarlitskaya what she wants to change.

test/verify/check-examples
martinpitt: I looked at this again with fresh eyes, considered alternatives like `check=True` or prepending `set +e`, and I still prefer `|| true` as the most explicit and most robust variant. So I just resolved the conflict for now.

martinpitt: I dislike `check=True` in particular because it's much less obvious: Does it ignore non-zero exit codes of the whole Popen command (which would be my first instinct on a Python API)? That's not what it does, it's equivalent to `set +e` -- but that isn't reflected in the name. I really like the simplicity and clarity of `.execute()` being "a strict `set -e` script, period". `check=False` is harder to grep and visually scan for (as it often is on a separate line), and not self-descriptive enough IMHO.

allisonkarlitskaya (changes requested):
Review is generally the same as before.

Anywhere where I see `|| true"`, I'd (strongly) prefer to see `, check=False` instead.

martinpitt:
@allisonkarlitskaya : As we are at an impasse here, I reverted the `|| true` bits. The rest is still good, and big enough to get out of sync soon.

test/common/testlib.py
```diff
@@ -1479,7 +1479,7 @@ def nonDestructiveSetup(self):
         self.addCleanup(m.execute, "find /var/lib/systemd/coredump -type f -delete")
 
         # temporary directory in the VM
-        self.addCleanup(m.execute, "if [ -d {0} ]; then findmnt --list --noheadings --output TARGET | grep ^{0} | xargs -r umount && rm -r {0}; fi".format(self.vm_tmpdir))
+        self.addCleanup(m.execute, "if [ -d {0} ]; then findmnt --list --noheadings --output TARGET | grep ^{0} | xargs -r umount; rm -r {0}; fi".format(self.vm_tmpdir))
```
allisonkarlitskaya, line 1482: What do you think about explicitly using bash and also setting `pipefail`?

test/verify/check-apps
```diff
@@ -94,7 +94,8 @@ class TestApps(PackageCase):
                 '{ "config": { "appstream_data_packages": [ "appstream-data-test" ] } }')
 
         if urlroot != "":
-            m.execute(f'mkdir -p /etc/cockpit/ && echo "[WebService]\nUrlRoot={urlroot}" > /etc/cockpit/cockpit.conf')
+            m.execute("mkdir -p /etc/cockpit")
+            m.write("/etc/cockpit/cockpit.conf", f"[WebService]\nUrlRoot={urlroot}")
```
allisonkarlitskaya, line 98: `m.write()` should imply `mkdir -p` of the parent directory.

test/verify/check-connection
```diff
@@ -467,8 +468,8 @@ class TestConnection(MachineCase):
     @nondestructive
     def testConfigOrigins(self):
         m = self.machine
-        m.execute(
-            'mkdir -p /etc/cockpit/ && echo "[WebService]\nOrigins = http://other-origin:9090 http://localhost:9090" > /etc/cockpit/cockpit.conf')
+        m.execute("mkdir -p /etc/cockpit")
+        m.write("/etc/cockpit/cockpit.conf", "[WebService]\nOrigins = http://other-origin:9090 http://localhost:9090")
```
allisonkarlitskaya, line 472: I also wonder if we shouldn't add a `dedent` call on the content for `write()`.

test/verify/check-examples
```diff
@@ -61,9 +61,9 @@ class TestXHRProxy(MachineCase):
 
         # set up served directory
         httpdir = self.vm_tmpdir + "/root"
-        m.execute("mkdir {0} && echo world > {0}/hello.txt".format(httpdir))
+        m.execute("mkdir {0}; echo world > {0}/hello.txt".format(httpdir))
```
allisonkarlitskaya, line 64: why not change this to an f-string?

test/verify/check-pages
```diff
@@ -312,11 +312,11 @@ OnCalendar=daily
         m.execute('echo scruffy:foobar | chpasswd')
 
         if "debian" in m.image:
-            m.execute('echo \'pt_BR.UTF-8 UTF-8\' >> /etc/locale.gen && locale-gen && update-locale')
+            m.execute('echo \'pt_BR.UTF-8 UTF-8\' >> /etc/locale.gen; locale-gen; update-locale')
```
allisonkarlitskaya, line 315: does `write()` have an append argument?  You split similar things like this out before, although I could understand if you don't want to do that here for reasons of symmetry.

allisonkarlitskaya (approved):
This is a substantial robustification and is ready to go as-is as far as I'm concerned.  All review comments are suggestions and/or possible follow-ups.

test/verify/check-examples
allisonkarlitskaya: ....and possibly also `m.write()`

test/common/testlib.py
martinpitt: Good idea!

test/verify/check-apps
martinpitt: I thought about that, I can't quite make up my mind between "too much magic" and making things simpler. But as you suggest it as well, I'll send a bots PR to imply mkdir, and then do another pass over tests to clean up redundant mkdirs.

test/verify/check-examples
martinpitt: It would mean to repeat `httpdir`, but it's not a very big deal. But let's clean this up once m.write() implies mkdir, then this duplication goes away automatically and f-string is the natural choice.

test/verify/check-pages
martinpitt: Indeed it does! I'll pick this up in the next round after we get that mkdir behaviour.

martinpitt:
I made TODO items for the leftover issues above, will send bots and cockpit PR follow-ups. Thanks for the review!

test/verify/check-apps
martinpitt: See https://github.com/cockpit-project/bots/pull/3806 .

test/common/testlib.py
martinpitt: Ah no -- this cleans up any bind mounts from within vm_tmpdir, but most tests wouldn't have any. I.e. the `grep` is expected to fail, that's also why xargs has to use the -r option (--no-run-if-empty).

test/verify/check-examples
martinpitt: Fixed in PR #17698

test/verify/check-pages
martinpitt: Fixed in PR #17698

PR #17646, opened 2022-08-12
test: Move domain specific libraries to test/common/

martinpitt:
cockpit-machines and subscription-manager-cockpit currently have
outdated code copies of these. It makes more sense to move them to
test/common, so that they can be shared and versioned similar to
testlib.py.

Right now it's awkward to do that, as these can only be checked out to
test/verify/ with our current mechanism, and that's not where every
other project keeps their tests.


martinpitt:
This touches the same files as PR #17645 -- I hope that they won't collide (as git is fairly bad with tracking renames), otherwise it's an easy enough rebase.

martinpitt:
@skobyda , @ptoscano : Once this lands, you can add *lib.py to Makefile's `COCKPIT_REPO_FILES` and drop the code copies.

ptoscano:
> @ptoscano : Once this lands, you can add *lib.py to Makefile's `COCKPIT_REPO_FILES` and drop the code copies.

sub-man-cockpit does not use `COCKPIT_REPO_FILES`; that said, as long as it is in `test/common/`, it is enough for sub-man-cockpit.

martinpitt:
@ptoscano : Ah right, that's another commit to backport :grin:

ptoscano:
> @ptoscano : Ah right, that's another commit to backport :grin:

Possibly, although it requires first updating the cockpit library to a newer version, which most likely means adapting to newer APIs to cockpit components (and I know for sure at least one change needed). Even if it's something I'd like to, it is definitely a bigger task.

martinpitt:
@allisonkarlitskaya : Bah, [this c8s failure](https://artifacts.dev.testing-farm.io/6a0a49ac-073f-4bc9-90e7-a22bff18a349/) is fallout from https://github.com/cockpit-project/bots/pull/3709 , sorry about that. It's not immediately clear to me what happens -- .write() definitively uses input=, and the value is `"[Log]\nFatal = criticals\n"`, i. e. not even an empty string or similar corner case. There is a certain asymmetry:
```
stdin=None if input is not None else subprocess.DEVNULL,
input=input.encode("UTF-8") if input else None,
```
In the second line it should be `if input is not None`, but that only matters for input=="" and such. In this case I'd expect `stdin=None. I don't have time any more to debug this, smells like some Python 3.6 bug in RHEL/CentOS 8?

jelly:
> 



> @allisonkarlitskaya : Bah, [this c8s failure](https://artifacts.dev.testing-farm.io/6a0a49ac-073f-4bc9-90e7-a22bff18a349/) is fallout from [cockpit-project/bots#3709](https://github.com/cockpit-project/bots/pull/3709) , sorry about that. It's not immediately clear to me what happens -- .write() definitively uses input=, and the value is `"[Log]\nFatal = criticals\n"`, i. e. not even an empty string or similar corner case. There is a certain asymmetry:
> 
> ```
> stdin=None if input is not None else subprocess.DEVNULL,
> input=input.encode("UTF-8") if input else None,
> ```
> 
> In the second line it should be `if input is not None`, but that only matters for input=="" and such. In this case I'd expect `stdin=None. I don't have time any more to debug this, smells like some Python 3.6 bug in RHEL/CentOS 8?

This looks like https://github.com/python/cpython/issues/79067 to me.

https://github.com/python/cpython/pull/11727/files

allisonkarlitskaya:
Fix is on the way in cockpit-project/bots#3734

martinpitt:
@allisonkarlitskaya Friendly review ping? I can rebase, but it seems packit tests are now broken on rawhide, so would just trade one known red against another?

allisonkarlitskaya (approved):
wfm

PR #17651, opened 2022-08-15
storaged: fix spelling issue found by codespell

jelly:


garrett:
This isn't a spelling error. It's talking about the trough of the scrollbar. It isn't something going "through" the scrollbar.

https://www.thefreedictionary.com/trough

> trough  (trôf, trŏf)
> n.
> 1.
>    a. A long, narrow, generally shallow receptacle for holding water or feed for animals.
>    b. Any of various similar containers for domestic or industrial use, such as kneading or washing.
> 2. A gutter under the edge of a roof for carrying off rainwater.
> 3. A long, narrow depression, as between waves or ridges.
> 4. A low point in a business cycle or on a statistical graph.
> 5. Meteorology An elongated region of relatively low atmospheric pressure, often associated with a front.
> 6. Physics A minimum point in a wave or an alternating signal.

It's usually associated with a pig trough:

image

Which a scrollbar's trough looks quite a lot like. (A long box, really.)

It's mentioned in Wikipedia's article too: https://en.wikipedia.org/wiki/Scrollbar

> Although scrollbar designs differ throughout their history, they usually appear on one or two sides of the viewing area as long rectangular areas containing a bar (or thumb) that can be [dragged](https://en.wikipedia.org/wiki/Mouse_gesture) along a trough (or track) to move the body of the document.

> A proportional thumb that completely fills the trough indicates that the entire document is being viewed, at which point the scrollbar may temporarily become hidden.

https://en.wikipedia.org/wiki/Scrollbar#Clicking_the_trough

> The trough above or below the thumb can be clicked in order to immediately jump to that point on the page, or page-by-page for multipage content.[[15]](https://en.wikipedia.org/wiki/Scrollbar#cite_note-BatesPat-15) After clicking the trough, scrolling automatically begins and then stop once the thumb has reached the position of the mouse pointer. This scrolling technique is faster than others and is best used when the user has to scroll through a lot of content at once, or when the user knows where exactly on a page they need to scroll to.

garrett:
For clarification: In this particular case, it's not a scrollbar but a progress bar, but progress bars still have a trough. :wink:

PR #17654, opened 2022-08-16
storage: Use mock values during some pixel tests

mvollmer:
Instead of ignoring parts of the DOM that are unpredictable, we
replace them with some constant non-sense.  This gets rid of the
unpredictable content and also produces a predictable layout by
avoiding variable sized content.

### DIFF
--- a/pkg/lib/hooks.js
+++ b/pkg/lib/hooks.js
@@ -44,6 +44,8 @@ import deep_equal from "deep-equal";
  * - useDeepEqualMemo: A utility hook that can help with things that
  * need deep equal comparisons in places where React only offers
  * Object identity comparisons, such as with useEffect.
+ *
+ * - useMock: For retrieving values injected during testing.
  */
 
 /* - usePageLocation()
@@ -325,4 +327,32 @@ export function useEvent(obj, event, handler) {
 
 export function useInit(func, deps, comps, destroy = null) {
     return useObject(func, destroy, deps || [], comps);
+}
+
+/* - useMock()
+ *
+ * function Component() {
+ *   const mock = useMock();
+ *   const now = Date.now();
+ *
+ *   return <div id="date">{mock?.date ?? now.toString()}</div>;
+ * }
+ *
+ * In the test:
+ *
+ *   b.assert_pixels('#date', mock={'date': "12. January"})
+ *
+ * Normally, "useMock()" will return undefined or null, but a test can
+ * cause it to return some other arbitrary value temporarily.  This is
+ * mostly useful for producing a stable layout for a pixel test.
+ *
+ * The value returned by useMock is typically a object with a few
+ * members. You don't need to worry about the member names; it is
+ * usually not a problem when two unrelated pieces of code happen to
+ * use the same names (like "date" in the example above).
+ */
+
+export function useMock() {
+    useEvent(window, "test_mock_changed");
+    return window.test_mock;
 }
--- a/pkg/storaged/dialog.jsx
+++ b/pkg/storaged/dialog.jsx
@@ -240,6 +240,8 @@ import { ListingTable } from "cockpit-components-table.jsx";
 
 import { fmt_size, block_name, format_size_and_text, format_delay, for_each_async } from "./utils.js";
 import { fmt_to_fragments } from "utils.jsx";
+import { useMock } from "hooks.js";
+
 import client from "./client.js";
 
 const _ = cockpit.gettext;
@@ -1068,6 +1070,7 @@ export const BlockingMessage = (usage) => {
 };
 
 const UsersPopover = ({ users }) => {
+    const mock = useMock()?.busy;
     const max = 10;
     const services = users.filter(u => u.unit);
     const processes = users.filter(u => u.pid);
@@ -1094,7 +1097,7 @@ const UsersPopover = ({ users }) => {
                         ? <p>
                             <b>{_("Processes using the location")}</b>
                             <List>
-                                { processes.slice(0, max).map((u, i) => <ListItem key={i}>{u.comm} (user: {u.user}, pid: {u.pid})</ListItem>) }
+                                { processes.slice(0, max).map((u, i) => <ListItem key={i}>{u.comm} (user: {u.user}, pid: {mock?.pid ?? u.pid})</ListItem>) }
                                 { processes.length > max ? <ListItem key={max}>...</ListItem> : null }
                             </List>
                         </p>
@@ -1175,13 +1178,15 @@ export function init_active_usage_processes(client, usage) {
 }
 
 export const StopProcessesMessage = ({ mount_point, users }) => {
+    const mock = useMock()?.busy;
+
     const process_rows = users.filter(u => u.pid).map(u => {
         return {
             columns: [
-                u.pid,
+                mock?.pid ?? u.pid,
                 { title: u.cmd.substr(0, 100), props: { modifier: "breakWord" } },
                 u.user || "-",
-                { title: format_delay(-u.since * 1000), props: { modifier: "nowrap" } }
+                { title: mock?.since ?? format_delay(-u.since * 1000), props: { modifier: "nowrap" } }
             ]
         };
     });
--- a/pkg/storaged/part-tab.jsx
+++ b/pkg/storaged/part-tab.jsx
@@ -22,35 +22,35 @@ import { DescriptionList, DescriptionListDescription, DescriptionListGroup, Desc
 
 import cockpit from "cockpit";
 import * as utils from "./utils.js";
+import { useMock } from "hooks.js";
 
 const _ = cockpit.gettext;
 
-export class PartitionTab extends React.Component {
-    render() {
-        const block_part = this.props.client.blocks_part[this.props.block.path];
+export const PartitionTab = ({ client, block }) => {
+    const mock_UUID = useMock()?.part_UUID;
+    const block_part = client.blocks_part[block.path];
 
-        return (
-            <DescriptionList className="pf-m-horizontal-on-sm">
-                <DescriptionListGroup>
-                    <DescriptionListTerm>{_("Name")}</DescriptionListTerm>
-                    <DescriptionListDescription>{block_part.Name || "-"}</DescriptionListDescription>
-                </DescriptionListGroup>
+    return (
+        <DescriptionList className="pf-m-horizontal-on-sm">
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Name")}</DescriptionListTerm>
+                <DescriptionListDescription>{block_part.Name || "-"}</DescriptionListDescription>
+            </DescriptionListGroup>
 
-                <DescriptionListGroup>
-                    <DescriptionListTerm>{_("Size")}</DescriptionListTerm>
-                    <DescriptionListDescription>{utils.fmt_size(block_part.Size)}</DescriptionListDescription>
-                </DescriptionListGroup>
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Size")}</DescriptionListTerm>
+                <DescriptionListDescription>{utils.fmt_size(block_part.Size)}</DescriptionListDescription>
+            </DescriptionListGroup>
 
-                <DescriptionListGroup>
-                    <DescriptionListTerm>{_("UUID")}</DescriptionListTerm>
-                    <DescriptionListDescription>{block_part.UUID}</DescriptionListDescription>
-                </DescriptionListGroup>
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("UUID")}</DescriptionListTerm>
+                <DescriptionListDescription>{mock_UUID ?? block_part.UUID}</DescriptionListDescription>
+            </DescriptionListGroup>
 
-                <DescriptionListGroup>
-                    <DescriptionListTerm>{_("Type")}</DescriptionListTerm>
-                    <DescriptionListDescription>{block_part.Type}</DescriptionListDescription>
-                </DescriptionListGroup>
-            </DescriptionList>
-        );
-    }
-}
+            <DescriptionListGroup>
+                <DescriptionListTerm>{_("Type")}</DescriptionListTerm>
+                <DescriptionListDescription>{block_part.Type}</DescriptionListDescription>
+            </DescriptionListGroup>
+        </DescriptionList>
+    );
+};
--- a/test/verify/check-storage-mounting
+++ b/test/verify/check-storage-mounting
@@ -77,6 +77,7 @@ ExecStart=/usr/bin/sleep infinity
         b.wait_in_text("#dialog", "Test Service")
         b.wait_in_text("#dialog", "/usr/bin/sleep infinity")
         b.wait_in_text("#dialog", "The listed processes and services will be forcefully stopped.")
+        b.assert_pixels("#dialog", "busy-unmount", mock={'busy': {'pid': 1234, 'since': "since 1m"}})
         self.confirm()
         self.content_tab_wait_in_info(1, 1, "Mount point", "The filesystem is not mounted")
 
--- a/test/common/test-functions.js
+++ b/test/common/test-functions.js
@@ -315,3 +315,8 @@ function ph_element_clip(sel) {
 function ph_count_animations(sel) {
     return ph_find(sel).getAnimations({ subtree: true }).length;
 }
+
+function ph_set_test_mock(config) {
+    window.test_mock = config;
+    window.dispatchEvent(new Event("test_mock_changed"));
+}
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -551,6 +551,9 @@ class Browser:
 
         self.wait_val("{0} > div input[type=text]".format(group_identifier), location)
 
+    def set_test_mock(self, config):
+        self.call_js_func("ph_set_test_mock", config)
+
     def wait_timeout(self, timeout: int):
         browser = self
 
@@ -1022,12 +1025,16 @@ class Browser:
 
     def assert_pixels_in_current_layout(self, selector: str, key: str,
                                         ignore: List[str] = [],
+                                        mock=None,
                                         scroll_into_view: Optional[str] = None,
                                         wait_animations: bool = True):
         """Compare the given element with its reference in the current layout"""
 
         if not (Image and self.pixels_label):
             return
+
+        if mock:
+            self.set_test_mock(mock)
 
         self._adjust_window_for_fixed_content_size()
         self.call_js_func('ph_scrollIntoViewIfNeeded', scroll_into_view or selector)
@@ -1070,6 +1077,7 @@ class Browser:
             raise SystemError("Pixel test references are missing, please run: test/common/pixel-tests pull")
 
         ignore_rects = relative_clips(list(map(lambda item: selector + " " + item, ignore)))
+
         base = self.pixels_label + "-" + key
         if self.current_layout != self.layouts[0]:
             base += "-" + self.current_layout["name"]
@@ -1162,8 +1170,12 @@ class Browser:
                 print("Differences in pixel test " + base)
                 self.failed_pixel_tests += 1
 
+        if mock:
+            self.set_test_mock(None)
+
     def assert_pixels(self, selector: str, key: str,
                       ignore: List[str] = [],
+                      mock=None,
                       skip_layouts: List[str] = [],
                       scroll_into_view: Optional[str] = None,
                       wait_animations: bool = True,
@@ -1181,7 +1193,8 @@ class Browser:
                     time.sleep(wait_delay)
                     if "rtl" in self.current_layout["name"]:
                         self._set_direction("rtl")
-                    self.assert_pixels_in_current_layout(selector, key, ignore=ignore,
+                    self.assert_pixels_in_current_layout(selector, key,
+                                                         ignore=ignore, mock=mock,
                                                          scroll_into_view=scroll_into_view,
                                                          wait_animations=wait_animations)
 
--- a/test/verify/check-storage-basic
+++ b/test/verify/check-storage-basic
@@ -61,12 +61,7 @@ class TestStorageBasic(StorageCase):
         self.content_row_wait_in_col(1, 2, "ext2 filesystem")
 
         self.content_tab_expand(1, 1)
-        b.assert_pixels("#detail-content", "partition", ignore=[
-            # UUID is not stable
-            "dt:contains(UUID) + dd",
-            # browsers/harfbuzz don't render "ext2 filesystem" label reliably
-            "td[data-label='Type']",
-        ])
+        b.assert_pixels("#detail-content", "partition", mock={'part_UUID': "1234"})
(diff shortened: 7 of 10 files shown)


mvollmer:
We don't need to restore the old values; the idea is that React will do that on the next render.  But there is a race between setting the mock text content and taking the snapshot.  If React decides to render in that window, things will break. Let's see.

garrett:
> We don't need to restore the old values; the idea is that React will do that on the next render. But there is a race between setting the mock text content and taking the snapshot. If React decides to render in that window, things will break. Let's see.

No, I don't think this is the right solution, as we're still setting up a race condition.

The data itself has to change before the DOM.

mvollmer:
> No, I don't think this is the right solution, as we're still setting up a race condition.

We could have a global toggle that puts each page into "mock mode" where it will replace certain elements with fake data.  This means changes to our code, but that's totally acceptable.

Or, we could try to stop the pages JavaScript via the CDP Debugger.pause() function... but maybe that will also block our own eval_js and we need to use CDP also to poke in the fake data... sounds a bit too involved to me.

garrett:
> We could have a global toggle that puts each page into "mock mode" where it will replace certain elements with fake data. This means changes to our code, but that's totally acceptable.

I was wondering about this and was going to suggest it, but the ramifications could be immense, as so much can change everywhere.

> Or, we could try to stop the pages JavaScript via the CDP Debugger.pause() function... but maybe that will also block our own eval_js and we need to use CDP also to poke in the fake data... sounds a bit too involved to me.

I have been looking for ways to pause React's DOM updating instead. The problem isn't with browser DOM updates, but with React updating the DOM, so we might as well try to attack it at that level.

If we _can_ do something like `React.RenderUpdates.Pause()` (this is totally made up), swap out the string(s) in the DOM, then test, then `React.RenderUpdates.Resume()` (also made up), then this simple string swap approach would work fine.

Surely, we're not the first to hit this kind of problem. Right?

garrett:
`shouldComponentUpdate()` might almost be what we want.

https://reactjs.org/docs/react-component.html#shouldcomponentupdate

Use shouldComponentUpdate() to let React know if a component’s output is not affected by the current change in state or props.

_HOWEVER:_ 

> This method only exists as a [performance optimization](https://reactjs.org/docs/optimizing-performance.html). Do not rely on it to “prevent” a rendering, as this can lead to bugs.

and

> In the future React may treat shouldComponentUpdate() as a hint rather than a strict directive, and returning false may still result in a re-rendering of the component.

So I guess we should keep looking.

It would also be nice to flip the theoretical pause switch on a React level, not on a component level.

garrett:
There's `ReactDOM.flushSync()` (https://reactjs.org/docs/react-dom.html#flushsync) which causes React to render immediately. If we called that in the test code (_important: not in production code_) before monkeying with the DOM, then it's a way to prevent some of the race conditions. It's a bad hack though; it would be better to pase/resume React's rendering somehow.

mvollmer:
`Debugger.pause` works well enough, but changing the DOM via the CDP API is cumbersome and apparently also unreliable.  Hmm.

mvollmer:
> and apparently also unreliable

No, i's just that querySelectorAll doesn't understand the`:contains` syntax.

mvollmer:
> I have been looking for ways to pause React's DOM updating instead.

Yeah, me too, but I didn't find anything.

React renders should not be totally random, there needs to be an event of some kind. During testing, we control all input events.  That leaves the WebSocket events.  (Idle and timeout events hopefully don't matter...)

mvollmer:
So I am reasonably happy with the Debugger.pause approach.  Code could be cleaned up, of course.

mvollmer:
Arg, now I get this:

```
RuntimeError: DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

This originates from React, so I am guessing React is unhappy with our changes to the DOM when it tries to re-render.  This would be a show-stopper for the "mock without React involvement" approach.

garrett:
> React renders should not be totally random, there needs to be an event of some kind. During testing, we control all input events. That leaves the WebSocket events. (Idle and timeout events hopefully don't matter...)

No, afaik, React updates the DOM during render calls, and there could be a different render call on the page which could trigger it being updated. It's not necessarily tied to input events.

Again: if we cannot fix the data (static IP), then having mock data is probably the more correct approach here. Otherwise we'll constantly be fighting React and having race conditions.

mvollmer:
Alright, what about the global "mock mode", where the tests instruct the page to behave differenty just for the benefits of the tests.  I think that is a useful idea in general, and allows us to test things that are otherwise hard to set up "for real".

I have pushed a minimal version of this idea that just replaces certain texts, via a new hook.

garrett:
> Alright, what about the global "mock mode", where the tests instruct the page to behave differenty just for the benefits of the tests. I think that is a useful idea in general, and allows us to test things that are otherwise hard to set up "for real".

Yes, that makes sense.

The big issue with the IPv6 addresses is that we need to be able to swap out data, not just add it, so this should be taken into consideration.

garrett:
Alternative is to be able to not show stuff when in mock mode, so things can be hidden and alternatives can be shown, instead of changing it.

pkg/lib/hooks.js
```diff
@@ -325,3 +325,25 @@ export function useEvent(obj, event, handler) {
...
+    let config = useMockConfig();
+
+    return (val, tag) => {
+        if (!config)
+            return val;
+        if (topic)
+            config = config[topic];
+        if (tag && config)
+            config = config[tag];
+        return config || "mock";
```
cockpituous, line 347: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220819-071652-20323b89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/networkmanager/network-main.jsx
```diff
@@ -92,8 +93,8 @@ export const NetworkPage = ({ privileged, operationInProgress, usage_monitor, pl
 
         if (show_traffic) {
             const samples = usage_monitor.samples[iface.Name];
-            row.columns.push({ title: samples ? cockpit.format_bits_per_sec(samples[1][0] * 8) : "" });
-            row.columns.push({ title: samples ? cockpit.format_bits_per_sec(samples[0][0] * 8) : "" });
+            row.columns.push({ title: samples ? mock(cockpit.format_bits_per_sec(samples[1][0] * 8), "tx") : "" });
+            row.columns.push({ title: samples ? mock(cockpit.format_bits_per_sec(samples[0][0] * 8), "rx") : "" });
```
cockpituous, line 97: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220819-071652-20323b89-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/hooks.js
```diff
@@ -325,3 +325,29 @@ export function useEvent(obj, event, handler) {
...
+    return (val, tag) => {
+        let mock = config;
+
+        if (!mock)
+            return val;
+
+        if (topic)
+            mock = mock[topic];
+        if (tag && mock)
+            mock = mock[tag];
```
cockpituous, line 349: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220819-083129-c0e512fa-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/hooks.js
```diff
@@ -325,3 +325,29 @@
...
+
+        return mock || "mock";
```
cockpituous, line 351: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220819-083129-c0e512fa-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/networkmanager/network-main.jsx
cockpituous, line 97: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220819-083129-c0e512fa-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/hooks.js
```diff
@@ -325,3 +325,21 @@ export function useEvent(obj, event, handler) {
...
+    useEvent(window, "test_mock_changed");
+    return window.test_mock;
+}
+
+export function useMock(topic) {
+    const config = useMockConfig();
+
+    return (val, mock_val) => {
+        if (config === true || (topic && config === topic))
+            return mock_val || "mock";
```
cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220822-064711-e2f09928-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/networkmanager/network-main.jsx
```diff
@@ -92,8 +93,8 @@ export const NetworkPage = ({ privileged, operationInProgress, usage_monitor, pl
 
         if (show_traffic) {
             const samples = usage_monitor.samples[iface.Name];
-            row.columns.push({ title: samples ? cockpit.format_bits_per_sec(samples[1][0] * 8) : "" });
-            row.columns.push({ title: samples ? cockpit.format_bits_per_sec(samples[0][0] * 8) : "" });
+            row.columns.push({ title: samples ? mock(cockpit.format_bits_per_sec(samples[1][0] * 8)) : "" });
+            row.columns.push({ title: samples ? mock(cockpit.format_bits_per_sec(samples[0][0] * 8)) : "" });
```
cockpituous, line 97: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220822-064711-e2f09928-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/hooks.js
cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220822-092047-5ac70a0f-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/networkmanager/network-main.jsx
cockpituous, line 97: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220822-092047-5ac70a0f-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

test/verify/check-storage-mounting
```diff
@@ -72,11 +72,8 @@ ExecStart=/usr/bin/sleep infinity
 
         self.content_dropdown_action(1, "Unmount")
         b.wait_in_text("#dialog", str(sleep_pid))
-        b.wait_in_text("#dialog", "sleep infinity")
```
mvollmer, line 75: These assertions should not be removed.

pkg/lib/hooks.js
cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220823-070904-893c0860-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/networkmanager/network-main.jsx
cockpituous, line 97: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220823-070904-893c0860-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/hooks.js
cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220823-091359-727e4c6c-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220823-130157-33c3a71e-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
In defense of the general idea of putting fake data into a page while doing pixel tests:

We already have a mechanism to ignore certain DOM elements.  This is necessary if there are any elements that have unpredictable and hard to control content, such as UUIDs of newly created things. However, if the unpredictable content varies enough, it will influence the layout of elements around it.  Sometimes this influence is a bug; no matter what an element shows, we don't want it to change the layout.  Sometimes, it is not a bug, such as with table columns.

The way tables work with Patternfly is that even the smallest change in one column (such as replacing a "9" with a narrower "1") will change the position of all columns, also by a very small amount.  This will shift all texts by half a pixel, say, and will look like unreliable text rendering.

In those cases, it helps to make sure that all columns always have a constant, predictable content.

In the case where a layout change counts as a bug, we can use the mocking mechanism to test extreme cases that might be hard to produce in other ways.

pkg/lib/hooks.js
cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220824-103107-617c38e7-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
Pixel tests are supposed to fail for the time being.  Let's fix them immediately before merging.

pkg/lib/hooks.js
cockpituous, line 342: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20220927-081450-953815f3-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/hooks.js
```diff
@@ -326,3 +326,21 @@ export function useEvent(obj, event, handler) {
...
```
cockpituous, line 343: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-084754-aa03fba9-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1161,13 +1163,15 @@ export function init_active_usage_processes(client, usage) {
 }
 
 export const StopProcessesMessage = ({ mount_point, users }) => {
+    const mock = useMock()?.busy;
```
cockpituous, line 1166: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-101354-40692ee9-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1161,13 +1163,15 @@
...
+
     const process_rows = users.filter(u => u.pid).map(u => {
         return {
             columns: [
-                u.pid,
+                mock?.pid ?? u.pid,
```
cockpituous, line 1171: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-101354-40692ee9-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1161,13 +1163,15 @@
...
                 { title: u.cmd.substr(0, 100), props: { modifier: "breakWord" } },
                 u.user || "-",
-                { title: format_delay(-u.since * 1000), props: { modifier: "nowrap" } }
+                { title: mock?.since ?? format_delay(-u.since * 1000), props: { modifier: "nowrap" } }
```
cockpituous, line 1174: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-101354-40692ee9-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/part-tab.jsx
```diff
@@ -22,35 +22,35 @@ import { DescriptionList, DescriptionListDescription, DescriptionListGroup, Desc
...
-                <DescriptionListGroup>
-                    <DescriptionListTerm>{_("Type")}</DescriptionListTerm>
-                    <DescriptionListDescription>{block_part.Type}</DescriptionListDescription>
-                </DescriptionListGroup>
-            </DescriptionList>
-        );
-    }
-}
+export const PartitionTab = ({ client, block }) => {
+    const mock_UUID = useMock()?.part_UUID;
```
cockpituous, line 30: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-101354-40692ee9-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
cockpituous, line 1166: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-122128-2c63e908-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
cockpituous, line 1171: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-122128-2c63e908-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
cockpituous, line 1174: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-122128-2c63e908-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/part-tab.jsx
cockpituous, line 30: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230302-122128-2c63e908-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1068,6 +1070,7 @@ export const BlockingMessage = (usage) => {
 };
 
 const UsersPopover = ({ users }) => {
+    const mock = useMock()?.busy;
```
cockpituous, line 1073: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230322-115306-eab6c63e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1094,7 +1097,7 @@
                         ? <p>
                             <b>{_("Processes using the location")}</b>
                             <List>
-                                { processes.slice(0, max).map((u, i) => <ListItem key={i}>{u.comm} (user: {u.user}, pid: {u.pid})</ListItem>) }
+                                { processes.slice(0, max).map((u, i) => <ListItem key={i}>{u.comm} (user: {u.user}, pid: {mock?.pid ?? u.pid})</ListItem>) }
```
cockpituous, line 1100: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230322-115306-eab6c63e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1175,13 +1178,15 @@
 }
 
 export const StopProcessesMessage = ({ mount_point, users }) => {
+    const mock = useMock()?.busy;
```
cockpituous, line 1181: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230322-115306-eab6c63e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1175,13 +1178,15 @@
...
+
     const process_rows = users.filter(u => u.pid).map(u => {
         return {
             columns: [
-                u.pid,
+                mock?.pid ?? u.pid,
```
cockpituous, line 1186: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230322-115306-eab6c63e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/dialog.jsx
```diff
@@ -1175,13 +1178,15 @@
...
                 { title: u.cmd.substr(0, 100), props: { modifier: "breakWord" } },
                 u.user || "-",
-                { title: format_delay(-u.since * 1000), props: { modifier: "nowrap" } }
+                { title: mock?.since ?? format_delay(-u.since * 1000), props: { modifier: "nowrap" } }
```
cockpituous, line 1189: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230322-115306-eab6c63e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/storaged/part-tab.jsx
cockpituous, line 30: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17654-20230322-115306-eab6c63e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

mvollmer:
Alright, nobody likes to change the production code, including me and the Coverage report.

Summary of past approaches:

 - Changing the DOM without Reacts involvement might be racy, but we can get that done with CDP and Debugger.pause.
  - Forcefully changing the DOM and then hoping that React puts its back how it was doesn't work because React will stop working when it finds the DOM in an unexpected state.
 
So, I'll try to modify the DOM with Debugger.pause and put it back how it was before unfreezing the page. To make this simply, I'll try to get the desired mocking effect by just adding things that can be cleanly taken away again, like `display: none`, and using `:after` .

mvollmer:
> This means changes to our code, but that's totally acceptable.

Nope, that wasn't acceptable. :-)

KKoukiou:
@mvollmer shall we close this one then as the implementation with the dom manipulation from testlib was prefered?

mvollmer:
> @mvollmer shall we close this one then as the implementation with the dom manipulation from testlib was prefered?

Yes, let's.

New story in #18772.

PR #17657, opened 2022-08-17
Metrics: Add read/write usage per disk

tomasmatus:
Adds popover with read/write stats for all disks. Fixes bug where paritions on
NVMe are not skipped which results in incorrect read/write stats.

image

The button is hidden when only 1 disk is found.
image


tomasmatus:
Needs robust tests, I found a solution.

garrett:
"View all disks" is misleading. We'd need to say something like "View per-disk throughput" or "View individual data rates".

pkg/metrics/metrics.jsx
```diff
@@ -641,6 +646,40 @@ class CurrentMetrics extends React.Component {
             cpu_label = this.state.cpuUsed + '%';
         }
 
+        let disksUsage = [];
+        // Clueless what the problem is, just testing
+        if (this.samples[14] && this.samples[15]) {
+            disksUsage = this.disksNames.map((name, i) => [
+                name,
+                this.samples[14][i] >= 1 ? cockpit.format_bytes_per_sec(this.samples[14][i]) : "0",
```
tomasmatus, line 654: ok the problem is here so I just need to rewrite this part

pkg/metrics/metrics.scss
```diff
@@ -233,6 +233,24 @@
...
+      width: fit-content;
+    }
+  }
+}
+
+.pf-c-popover {
+  .disks-nowrap-white-space {
+    td[data-label="Device"],
+    td[data-label="Read"],
+    td[data-label="Write"] {
```
marusak, line 250: Would this work on translated page?

tomasmatus: Nope, it wouldn't. The css should be now fixed here and in the network card.

pkg/metrics/metrics.scss
```diff
@@ -184,18 +184,14 @@
...
-    white-space: nowrap;
-  }
-
   // Shrink network cards and data a little more
   // (as space is costly and they use a lot)
   // FIXME: Let's fix the content (split off units) and drop this when we can
-  td[data-label="In"],
-  td[data-label="Out"],
-  td[data-label="Interface"] {
-    font-size: var(--pf-global--FontSize--xs);
```
marusak, line 198: I guess this should have been a separate commit. It is fine, just for the next time

src/bridge/cockpitdisksamples.c
```diff
@@ -135,9 +135,15 @@ cockpit_disk_samples (CockpitSamples *samples)
           && g_ascii_isdigit (dev_name[strlen (dev_name) - 1]))
         continue;
 
+      // ignore nvme partitions
+      if (g_str_has_prefix (dev_name, "nvme") && g_strrstr (dev_name, "p"))
+        continue;
+
```
marusak, line 141: also somewhat unrelated, separate commit would be probably nicer

marusak (approved):
Thank you! Can you please write release note?

garrett (changes requested):
I'm trying to review this PR and it just breaks metrics collection on my computer.

Here's this PR. Note how there's a lack of realtime metrics:

Screenshot from 2022-09-06 11-20-18

Here's how it looks on main:

Screenshot from 2022-09-06 11-17-58

tomasmatus:
@garrett in order to test this locally you need to mount cockpit-pcp:
```
# mount -o bind ./cockpit-pcp /usr/libexec/cockpit-pcp
```

I am not sure how this translates to f-silverblue but this is what I have to do in order to see changes locally.
Another option is to use a VM.

garrett (changes requested):
Bind mount worked. Thanks!

---

Alignment is wrong. Units should be right-aligned.

Additionally, the units are also different. One uses MB/s for read, kB/s for write... and the other has them swapped. They should all be using the same units, consistent with themselves and consistent with others. Also, I thought we were using MiB/s now, not MB/s?

So, can you:

- Unify units; probably MiB/s for everything
- Right-align
- Use tabular numbers (_ideally_, although I'm not sure if Cockpit is _still_ not set up for this? PF got it several months back and we desperately need it)

tomasmatus:
> They should all be using the same units, consistent with themselves and consistent with others.

Why? This is how `cockpit.format_bytes_per_sec()` works. IO can range from a couple kilo Bytes to hundreds of MB or even GB on SSDs. I don't think there's a reason to "unify" unit for all values.
Also units aren't unified in the overall read/write on the top and in the Network card they also differ.
image

> - Right-align
> - Use tabular numbers (ideally, although I'm not sure if Cockpit is still not set up for this? PF got it several months back and we desperately need it)

I'll look into it.

garrett:
> Why?

1. To make it  possible to compare
2. So the meaning doesn't change every second (like when when 300 Kb/s jumps to 2 MB/s and back)

> Also units aren't unified in the overall read/write on the top and in the Network card they also differ.

Yep, also a problem, for the same reasons.

tomasmatus:
> Yep, also a problem, for the same reasons.

I honestly don't think it's a problem and I think it's much better this way.

tomasmatus:
Comparing megabytes and kilobytes is certainly not a problem. Using the same unit every time is pointless (why do we even have unit prefixes then?)
If it were to use MB every time, then small read/write would be displayed as 0.002 MB/s and on fast devices I'd get something like 1200 MB/s... This is why I believe it's unreasonable.

garrett:
> Comparing megabytes and kilobytes is certainly not a problem. 

It certainly is a problem when units change and you're comparing numbers. I don't know why we're arguing this. It's a problem.

> Using the same unit every time is pointless (why do we even have unit prefixes then?)

I never said this. I said that we should make it more stable and make it the same for everything. Disk access of kb/s or less is minimal.

> If it were to use MB every time, then small read/write would be displayed as 0.002 MB/s and on fast devices I'd get something like 1200 MB/s... This is why I believe it's unreasonable.

You could compare those numbers though!

20 KB/s versus 20 MB/s is quite different! But they look the same at a glance, especially on an always-updating area where it could bounce around a bit (both with the number and the units).

The whole point is to see what's using data, which is using it the most, and then how much.

garrett:
I'm not making an arbitrary rule. This is a general UI design guideline.

Here's what PatternFly says:

https://www.patternfly.org/v4/ux-writing/units-and-symbols/#writing-units-of-measure

> Use consistent units, and avoid mixing them.
> Example: Write 10 to 75 seconds instead of 10 seconds to 1.25 minutes.

Here's an article talking about better table design, where you can see them recommending the same thing:

https://pencilandpaper.io/articles/ux-pattern-analysis-enterprise-data-tables/

Further down the page, it does specifically have a part about units:

image

Text:

> Speaking of units of measures, you should make sure to always use the same unit within columns. If, for example, the entries for a measurement field span from millimeters to meters, you should use centimeters as a middle ground and have the values reflect that with decimals.

This is their example table (in that context) after modifying it:

Screenshot 2022-09-08 at 14-31-46 UX Data Table Design Patterns   Best Practices



https://medium.com/mission-log/design-better-data-tables-430a30a00d8c

pkg/metrics/metrics.scss
garrett: I found and fixed a few others of these @ https://github.com/cockpit-project/cockpit/pull/17721. (I purposefully skipped fixing these there, so the PRs should not conflict.)

marusak:
I am on this one with @tomasmatus. The range we show is too big for using one unit. Examples you linked make sense, but they are always numbers that are too close. We talk here from bytes to gigabytes, that is 10^9 difference.

> Example: Write 10 to 75 seconds instead of 10 seconds to 1.25 minutes.

But you would not write 5259487 seconds to express two months, would you? Or 0.16 years to express two months.

> If, for example, the entries for a measurement field span from millimeters to meters, you should use centimeters as a middle ground and have the values reflect that with decimals

In our case it spans from micrometers to kilometers and I don't want my kilometers to be expressed in cm. Surely I can understand that 0.5km is much more that 50cm without it being in the same unit. What we want to show users is what multiple they are in - whether the device is using bytes, or kilobytes or gigabytes. None is gonna care if it is 1.25Kb or 1.26Kb.

garrett:
> But you would not write 5259487 seconds to express two months, would you? Or 0.16 years to express two months.

No one is suggesting this whatsoever. It's a total [straw-man argument](https://en.wikipedia.org/wiki/Straw_man).

> Agreed - in most cases, the only really useful thing here is the unit (i.e. the magnitude), not the actual number. I.e. "idle", "trickle", or "firehose".

Incorrect. The actual magnitude is the concern, not the units. Having consistent units _with a reasonable unit that covers the devices_ as I suggested above, would allow someone to compare the throughput of devices and understand the magnitude. 

It's like comparing heights of humans. You could say 188 cm or 1.88 m. Nobody would say 0.00188 km, like @marusak is suggesting. Similarly, disk speed generally is in MB/s. If something gets higher or lower, it's still reasonable. Theoretical maximum speeds at the top of the line NVMe devices today, according to this article with the "fastest" as of right now, is 7000 MB/s... it still works; it's not absurd — and this article is even using it by default. If speeds do hit multiple GB/s for a sustained time, we could even adjust all the units (and values) consistently too.

Having random units and numbers that are all changing every second is useless. There needs to be some stability to compare the data being pushed around... and comparing the speed being read and written across devices is the entire point of this popup.

If we don't make it consistent then this feature is pointless.

tomasmatus:
Gnome system monitor does it in the same way with "mixing" units.

per process disk usage:
image

and the network usage:
image

It really is not a problem you make it out to be.

garrett (changes requested):
It is definitely a problem. I don't care that GNOME does it wrong.

Can we please stop arguing about this extremely basic thing and just fix it? I've pointed out several design references, including PatternFly, that indicate that mixing units is an actual problem. **Please fix this bug.**

martinpitt:
Note that mixing units is the status quo on main (and has been for some time), this isn't anything new from this PR. If we want to change it, then the existing stuff needs to be changed first. However, the difficulty is to get some agreement and figure out a generic way to sync up units between different fields -- once we have that, it's straightforward to apply to the disks as well, so IMHO we don't have to block this PR on that.

> Nobody would say 0.00188 km, like @marusak is suggesting. Similarly, disk speed generally is in MB/s. If something gets higher or lower, it's still reasonable.

In fact @marusak and @tomasmatus were suggesting exactly the *opposite*, and want to *avoid* numbers like 0.000188 (and instead use a proper unit) -- and FTR, I agree. I fully appreciate that I might be biased here, as I've been reading and using physical measurements for decades, and of course not everyone is used to that.

So if a disk or network interface A writes at 9 B/s, and another B at 9 MB/s what would we show for A? 
 - appropriate units: You vetoed "9 B/s"  (it would be my prefered option)
 - fixed units: You vetoed "0.000009 MB/s", and I veto that too)
 - rounding: I veto "0 MB/s", as it is important to tell the difference between "idle" and "lightly busy", especially in terms of power management or figuring out if there is any active connection.
 - So the only compromise that I can see is something like "< 1 MB/s" or "> 0 MB/s", or some similar notation of "a little bit, but not much". I admittedly don't like that, but I could live with that. Do you have a proposal what we should show there?

martinpitt:
@garrett : The PF examples or body heights don't apply here. Both humans and pencils all have the same magnitude, and of course in such a situation you can get away with picking a fixed units. But network and disk speeds vary over 12 magnitudes, so it's really not the same situation.

garrett (dismissed):
> - rounding: I veto "0 MB/s", as it is important to tell the difference between "idle" and "lightly busy", especially in terms of power management or figuring out if there is any active connection.

Rounding could work, as long as it's never 0 when there's actual traffic. In this case, bytes/s is basically nothing, so 0.1 MB/s is close enough in the comparison. It would, of course, require special case handling, but it's worth it, I think.

So, pseudocode would be:

```js
rounded_value = round(value, stable_unit)
if rounded == 0 then { rounded_value = round_up(value, stable_unit) }
```

Actual JS implementation, adapted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed:

```js
// Ensure that values are never 0 (unless they're supposed to be).
function round_to_string(x, decimals = 2) {
  let float_string = Number.parseFloat(x).toFixed(decimals);

  // If we get the equivalent to 0, then re-round using a ceiling
  if (!parseFloat(float_string)) {
    float_string = Number.parseFloat(
      Math.ceil(x * 10 ** decimals) / 10 ** decimals
    ).toFixed(decimals);
  }

  return float_string;
}
```

And here's the function in action:

```js
console.log(round_to_string(123.456));
// expected output: "123.46"

console.log(round_to_string('1.23e+5'));
// expected output: "123000.00"

console.log(round_to_string(0.004, 1));
// expected output: "0.1"

console.log(round_to_string(0.004, 2));
// expected output: "0.01"

console.log(round_to_string(0.004, 3));
// expected output: "0.004"

console.log(round_to_string(0.004, 0));
// expected output: "1"

console.log(round_to_string(0.000));
// expected output: "0.00"

console.log(round_to_string(0));
// expected output: "0.00"
```

Someone on the team could probably write this more elegantly. I know it could use recursion or some other tricksy thing, but I wrote it to be pretty straightforward.

> - So the only compromise that I can see is something like "< 1 MB/s" or "> 0 MB/s", or some similar notation of "a little bit, but not much". I admittedly don't like that, but I could live with that. Do you have a proposal what we should show there?

I'm also fine with this (for the less than, not for the greater than). But if we're going fractional already, then we'd still need to do special case handling. We might as well go with approximate rounding up (when rounding down returns 0) in this case and save the space (and possible confusion) of "< ".

Therefore, I prefer the rounding approach more.

> @garrett : The PF examples or body heights don't apply here. Both humans and pencils all have the same magnitude, and of course in such a situation you can get away with picking a fixed units. But network and disk speeds vary over 12 magnitudes, so it's really not the same situation.

My point is that nobody cares if it's a few bytes for one disk when something else is doing MB/s.

Of course, it's definitely also not 0 (and we shouldn't show 0 unless there really is no current transfer), but the exact number doesn't matter if it's so small and updating every second.

> If we want to change it, then the existing stuff needs to be changed

Yes, let's change it.

Sure, a compromise could be to not let it block this PR, but we'd need to do an immediate follow-up to make this PR useful. (And improve Cockpit elsewhere, especially on the metrics page.)

As a result, **I'm OK with the design** of this now, **provided we do work on making units stable in another PR**.

Note: If we're going to mention this in the release notes, the screenshot is out of date. And we'll probably want to re-take it once the stable units PR lands too.

martinpitt:
> In this case, bytes/s is basically nothing, so 0.1 MB/s is close enough in the comparison. 

It would be a lie though, as it's 5 magnitudes more than the actual value. "< 0.1 MB/s" would be true, and signal "a little". Nevertheless, I can live with that, and it at least has the advantage of not requiring some new formatting (which is probably also an issue for translations). Thanks for the discussion!

garrett:
Similar function, slightly modified for less-than reporting:

```js
function round_to_string(x, decimals = 2) {
  let float_string = Number.parseFloat(x).toFixed(decimals);

  // If we get a float_string equivalent to 0, then re-round using a ceiling
  // Also check to make sure x is != 0.
  if (!parseFloat(float_string) && parseFloat(x)) {
    float_string = "< " + Number.parseFloat(
      Math.ceil(x * 10 ** decimals) / 10 ** decimals
    ).toFixed(decimals);
  }

  return float_string;
}
```

Changes:

Add a check for a value in the if. Also concat a "< " in front.

pkg/metrics/metrics.jsx
```diff
@@ -641,6 +646,39 @@ class CurrentMetrics extends React.Component {
...
         }
 
+        let disksUsage = [];
+        if (this.disksNames.length > 0 && this.samples[14] && this.samples[15]) {
+            disksUsage = this.disksNames.map((name, i) => [
+                name,
+                this.samples[14][i] >= 1 ? cockpit.format_bytes_per_sec(this.samples[14][i]) : "0",
+                this.samples[15][i] >= 1 ? cockpit.format_bytes_per_sec(this.samples[15][i]) : "0",
+            ]);
+            disksUsage = disksUsage.filter(disk => !disk[0].match(/^zram[0-9]+$/));
```
martinpitt, line 656: Why not do the filtering first, and then the .map? saves some work. You can also chain these, i.e. no need to re-assign.

pkg/metrics/metrics.jsx
```diff
@@ -641,6 +646,39 @@ class CurrentMetrics extends React.Component {
...
+                    cells={ [{ title: _("Device"), transforms: [fitContent] }, _("Read"), _("Write")] }
+                    rows={disksUsage}
+                    rowWrapper={ props => <RowWrapper device-name={ props.row[0] } {...props} /> }>
+                    <TableHeader />
+                    <TableBody className="pf-m-tabular-nums disks-nowrap" />
+                </Table>
+            );
+
+            allDisks = (
+                <Popover minWidth={0} aria-label={ _("View all disks") } bodyContent={disksTableContent}>
```
martinpitt, line 676: Do we have any chance of only creating `disksTableContent` when it's actually visible? Most of the time this would be generated in vain. I didn't see anything in https://www.patternfly.org/v4/components/popover/ which could help here; bodyContent *can* take a function, but only to pass a "hide" callback, so that doesn't help us here.

Just thinking aloud, probably not actionable.

pkg/metrics/metrics.scss
```diff
@@ -184,18 +184,14 @@
   }
 
   // Don't wrap network in / out data
-  td[data-label="In"],
```
martinpitt, line 187: As already requested by Matej and Garrett, please split this out into a separate commit.

pkg/metrics/metrics.scss
```diff
@@ -184,18 +184,14 @@
...
-  }
-
   // Shrink network cards and data a little more
   // (as space is costly and they use a lot)
   // FIXME: Let's fix the content (split off units) and drop this when we can
-  td[data-label="In"],
-  td[data-label="Out"],
-  td[data-label="Interface"] {
-    font-size: var(--pf-global--FontSize--xs);
+  .network-nowrap-shrink {
```
martinpitt, line 190: This is also an independent change, isn't it? Please split out into a separate commit.

src/bridge/cockpitdisksamples.c
martinpitt: Agreed. This also needs some justification -- why do you want to do that? I quite like seeing my partitions on the metrics page..
```
/dev/nvme0n1p3 on /sysroot type ext4 (rw,relatime,seclabel)
/dev/nvme0n1p3 on / type ext4 (rw,relatime,seclabel)
/dev/nvme0n1p3 on /usr type ext4 (ro,relatime,seclabel)
/dev/nvme0n1p3 on /var type ext4 (rw,relatime,seclabel)
/dev/nvme0n1p2 on /boot type ext4 (rw,relatime,seclabel)
/dev/nvme0n1p4 on /var/home type ext4 (rw,relatime,seclabel)
/dev/nvme0n1p1 on /boot/efi type vfat (rw,relatime,fmask=0077,dmask=0077,codepage=437,iocharset=ascii,shortname=winnt,errors=remount-ro)
```

pkg/metrics/metrics.jsx
martinpitt, line 656: This "zram" special-case seems odd. I agree that we don't want to show disk usage on swap partitions, LVM PVs and similar "not quite a file system" partitions. But this should be done more generally -- perhaps a positive check that you only show mounted partitions? Or at least filter out all swaps?

test/verify/check-metrics
```diff
@@ -1094,6 +1094,31 @@ BEGIN {{
         m.execute("dd if=/dev/zero of=/var/cockpittest/blob bs=1M count=40")
         b.wait(lambda: progressValue(self, ".pf-c-progress[data-disk-usage-target='/var/cockpittest']") >= 90)
 
+        # read/write in popover
+        b.click("#current-metrics-card-disks .all-disks-no-gap button")
+        b.wait_visible(".pf-c-popover .disks-nowrap")
+        # read
+        m.execute("systemd-run --collect --slice cockpittest --unit disk-read-hog sh -ec 'while true; do echo 3 > /proc/sys/vm/drop_caches; grep -r . /usr >/dev/null; done'")
```
martinpitt, line 1101: This is very taxing. Could you please move the checks here into the already existing "reading lots of data" section above? It would fit very well there, and avoid testing the same scenario twice.

martinpitt (changes requested):
Thank you, nice work!

tomasmatus:
@martinpitt 
> Agreed. This also needs some justification -- why do you want to do that? I quite like seeing my partitions on the metrics page..

As it is now bridge already ignores partitions on `sdX`, `hdX` and `vdX`. However current logic does not cover nvme. So what happens is that it samples read/write of whole nvme + partition1 + partition2 + ... which results in incorrect values.

martinpitt:
@tomasmatus : Ah, so you only want to show per-device, not per file system, I see. That's fine then, thanks! Please still make it a separate commit.

pkg/metrics/metrics.jsx
tomasmatus: I'm doing map first because I'd have to remove the respective value from arrays in `this.samples[14]` and `this.samples[15]` if I filtered first. Not sure which approach is better but I can use splice and remove these indexes.

martinpitt: Ah right, of course you wouldn't get correct `i` indexes any more. Nevermind then.

Still, you can chain these, looks nicer than reassignment -- plus, you can make it `const`.

martinpitt: Sorry, I walked into the same trap as above -- we *are* showing devices here, *not* file systems. So this comment was mostly bogus. It's still not obvious to me what we actually want to show here -- only physical disks? So not ram disks, scsi_debug, or things like iSCSI and NBD? That would require some definitions first.

Given that RAM disk usage is still usage (in terms of CPU and "why is my system busy" at least), what makes them special?

But I agree, we don't need to pinpoint this right now -- we can start with showing mostly everything, and cut down once we have some more in-the-field experience. So I'd actually prefer to drop this filter entirely for now, but I don't veto the zram special case either.

tomasmatus: I agree, let's not filter at all. Thanks for the input!

martinpitt (approved):
Looks great now, thank you! Retrying the flaking fedora-37 test, and the one on packit is unrelated.

PR #17667, opened 2022-08-24
test: make Fedora-37 Fedora-36

jelly:
Use the same skip/naughties as Fedora-36 for Fedora-37.

Maybe we should stop doing this and explicitly add fedora-37 to our naughties/skips for packit and rawhide.

I think we might also want to stop using `fedora-development` as it's an alias to rawhide OR the beta when available which means your PR will randomly start failing as Fedora-37 is added.


marusak (approved):

marusak:
rawhide in TF fails, I am looking into it in https://github.com/cockpit-project/cockpit/pull/17666

PR #17671, opened 2022-08-25
kdump: use correct base for translations

jelly:
Introduced in af5678b0be55474ccf523 as a typo most likely.

This breaks the new Python bridge with KeyError 'base', do we need to fix it in there?

The old bridge handles it here:

```
src/bridge/cockpitpackages.c:  if (!cockpit_json_get_string (manifest, "base", NULL, &base))
```


allisonkarlitskaya (approved):
ya, obviously.  Thanks.

Issue #17672, opened 2022-08-25
storage error messages from mkfs.xfs are too verbose


allisonkarlitskaya:
### Explain what happens

![image](https://user-images.githubusercontent.com/36541154/186694618-bf4081df-1849-424f-b3b5-bd929f64e676.png)

mvollmer:
Yeah.  The strategy so far is to avoid triggering errors.  For example, we have code to prevent passing too long labels when formatting XFS: https://github.com/cockpit-project/cockpit/blob/main/pkg/storaged/utils.js#L144

In addition, we can trim everything after "Usage: ". *shrug*

jelly:
![image](https://user-images.githubusercontent.com/67428/186855576-d058509c-3bb4-48a3-85c4-609b3f9c0f15.png)

garrett:
The bold part of the alert is the title. The contents shouldn't be the title.

https://www.patternfly.org/v4/components/alert#inline-variations

![image](https://user-images.githubusercontent.com/10246/188599432-e3c365a9-36d4-4306-a2d9-76373008aee9.png)

We have been using our own custom solution for expandable errors (I think with the view details link that expands?), but PatternFly "recently" added expandable alerts:

https://www.patternfly.org/v4/components/alert#expandable

![image](https://user-images.githubusercontent.com/10246/188601608-301cd929-e528-4ad5-9548-b21c98500c86.png)

![image](https://user-images.githubusercontent.com/10246/188601652-2de87f4d-bab0-4f58-b083-80ef8ea8edd9.png)

These are all "success" alerts in the examples. We'd use errors/warnings. (Alerting on success isn't generally a good pattern, but PF uses them in the examples.)

PR #17673, opened 2022-08-25
test: change minimum filesystem sizes to 320MB

allisonkarlitskaya:
Newer versions of xfsprogs (5.19) refuse to create XFS filesystems on
devices that are smaller than 300MB.

See cockpit-project/bots#3769.

### DIFF
--- a/test/verify/check-storage-format
+++ b/test/verify/check-storage-format
@@ -52,7 +52,7 @@ class TestStorageFormat(StorageCase):
 
         self.login_and_go("/storage")
 
-        m.add_disk("50M", serial="DISK1")
+        m.add_disk("320M", serial="DISK1")  # xfs minimum size is ~300MB.
         b.wait_in_text("#drives", "DISK1")
         b.click('#drives .sidepanel-row:contains("DISK1")')
         b.wait_visible('#storage-detail')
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
Bots aren't having a good day, so I tested this manually with the arch image from cockpit-bots/bots#3769.  It works.

jelly:
A follow up should adjust our formatting UI to set a minimum of 300 MB for XFS right?

jelly:
> > A follow up should adjust our formatting UI to set a minimum of 300 MB for XFS right?
> 
> Formatting or resizing? You don't usually get to choose what size to use, when formatting...

I guess a better error when formatting and something for resizing?

mvollmer (approved):

mvollmer:
> > > A follow up should adjust our formatting UI to set a minimum of 300 MB for XFS right?
> > 
> > 
> > Formatting or resizing? You don't usually get to choose what size to use, when formatting...
> 
> I guess a better error when formatting and something for resizing?

Cockpit creates and formats partitions in one step, so we can stop people from creating too small XFS partitions.

XFS can not be made smaller, so there should be no issue with resizing.

When formatting a existing block device (such as a logical volume), we can still check the size and give a nice validation error before running mkfs, to avoid the aweful error message from it.

allisonkarlitskaya:
> When formatting a existing block device (such as a logical volume), we can still check the size and give a nice validation error before running mkfs, to avoid the aweful error message from it.

→ #17672 for all of that?

PR #17674, opened 2022-08-25
python: fix loading pages from /usr and XDG_*

jelly:
When loading from a Python zipfile we used to get str instances, but
when loading from the file system we get Path like objects. This breaks
negotiate_file as it will never find files from your system as it
compares a string to a Path like object.


allisonkarlitskaya (approved):
Nice catch, thanks!

PR #17678, opened 2022-08-26
python: metrics1 channel

jelly:
This only implements the overview page metrics, so let's make incremental changes for the other metric types.

### DIFF
--- a/src/cockpit/channels/metrics.py
+++ b/src/cockpit/channels/metrics.py
@@ -15,23 +15,169 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import asyncio
+import json
+import sys
+import time
 import logging
+from typing import Any, Dict, List, Set, NamedTuple, Optional, Tuple
+from collections import defaultdict
 
-from ..channel import Channel
+from ..channel import AsyncChannel, ChannelError
+from ..samples import SAMPLERS, Sampler, SampleDescription
 
 logger = logging.getLogger(__name__)
 
 
-class MetricsChannel(Channel):
-    payload = 'metrics1'
+class MetricInfo(NamedTuple):
+    derive: Optional[str]
+    desc: SampleDescription
 
-    def do_open(self, options):
-        assert options['source'] == 'internal'
-        assert options['interval'] == 3000
-        assert 'omit-instances' not in options
-        assert options['metrics'] == [
-            {"name": "cpu.basic.user", "derive": "rate"},
-            {"name": "cpu.basic.system", "derive": "rate"},
-            {"name": "cpu.basic.nice", "derive": "rate"},
-            {"name": "memory.used"},
-        ]
+
+class InternalMetricsChannel(AsyncChannel):
+    payload = 'metrics1'
+    restrictions = {'source': 'internal'}
+
+    metrics: List[MetricInfo]
+    samplers: Set
+    samplers_cache: Optional[Dict[str, Tuple[Sampler, SampleDescription]]] = None
+
+    interval: int = 1000
+    need_meta: bool = True
+    last_timestamp: float = 0
+    next_timestamp: float = 0
+
+    @classmethod
+    def ensure_samplers(cls):
+        if cls.samplers_cache is None:
+            cls.samplers_cache = {desc.name: (sampler, desc) for sampler in SAMPLERS for desc in sampler.descriptions}
+
+    def parse_options(self, options):
+        logger.debug('metrics internal open: %s, channel: %s', options, self.channel)
+
+        interval = options.get('interval', self.interval)
+        if not isinstance(interval, int) or interval <= 0 or interval > sys.maxsize:
+            raise ChannelError('protocol-error', message='invalid interval option')
+
+        self.interval = interval
+
+        metrics = options.get('metrics')
+        if not isinstance(metrics, list) or len(metrics) == 0:
+            logger.error('invalid "metrics" value: %s', metrics)
+            raise ChannelError('protocol-error', message='metrics is not an list')
+
+        sampler_classes = set()
+        for metric in metrics:
+            # validate it's an object
+            name = metric.get('name')
+            units = metric.get('units')
+            derive = metric.get('derive')
+
+            try:
+                sampler, desc = self.samplers_cache[name]
+            except KeyError:
+                logger.error('unsupported metric: %s', name)
+                raise ChannelError('protocol-error', message=f'unsupported metric: {name}')
+
+            if units and units != desc.units:
+                raise ChannelError('protocol-error', message=f'{name} has units {desc.units}, not {units}')
+
+            sampler_classes.add(sampler)
+            self.metrics.append(MetricInfo(derive=derive, desc=desc))
+
+        self.samplers = {cls() for cls in sampler_classes}
+
+    def send_meta(self, samples: Dict[str, Any], timestamp: float):
+        metrics = []
+        for metricinfo in self.metrics:
+            if metricinfo.desc.instanced:
+                metrics.append({
+                    'name': metricinfo.desc.name,
+                    'units': metricinfo.desc.units,
+                    'instances': list(samples[metricinfo.desc.name].keys()),
+                    'semantics': metricinfo.desc.semantics
+                })
+            else:
+                metrics.append({
+                    'name': metricinfo.desc.name,
+                    'derive': metricinfo.derive,
+                    'units': metricinfo.desc.units,
+                    'semantics': metricinfo.desc.semantics
+                })
+
+        meta = {
+            'timestamp': timestamp * 1000,
+            'interval': self.interval,
+            'source': 'internal',
+            'metrics': metrics
+        }
+        self.send_message(**meta)
+        self.need_meta = False
+
+    def sample(self):
+        samples = defaultdict(dict)
+        for sampler in self.samplers:
+            sampler.sample(samples)
+        return samples
+
+    def calculate_sample_rate(self, value: Any, old_value: Optional[Any]):
+        if old_value and self.last_timestamp:
+            return (value - old_value) / (self.next_timestamp - self.last_timestamp)
+        else:
+            return False
+
+    def send_updates(self, samples: Dict[str, Any], last_samples: Dict[str, Any]):
+        data = []
+        timestamp = time.time()
+        self.next_timestamp = timestamp
+
+        for metricinfo in self.metrics:
+            value = samples[metricinfo.desc.name]
+            old_value = last_samples[metricinfo.desc.name]
+
+            if metricinfo.desc.instanced:
+                # If we have less or more keys the data changed, send a meta message.
+                if value.keys() != old_value.keys():
+                    self.need_meta = True
+
+                if metricinfo.derive == 'rate':
+                    instances = []
+                    for key, val in value.items():
+                        instances.append(self.calculate_sample_rate(val, old_value.get(key)))
+
+                    data.append(instances)
+                else:
+                    data.append([val for val in value.values()])
+            else:
+                if metricinfo.derive == 'rate':
+                    data.append(self.calculate_sample_rate(value, old_value))
+                else:
+                    data.append(value)
+
+        if self.need_meta:
+            self.send_meta(samples, timestamp)
+
+        self.last_timestamp = self.next_timestamp
+        self.send_data(json.dumps([data]).encode())
+
+    async def run(self, options):
+        self.metrics = []
+        self.samplers = set()
+
+        InternalMetricsChannel.ensure_samplers()
+
+        self.parse_options(options)
+        self.ready()
+
+        last_samples = defaultdict(dict)
+        while True:
+            samples = self.sample()
+            self.send_updates(samples, last_samples)
+            last_samples = samples
+
+            try:
+                await asyncio.wait_for(self.read(), self.interval / 1000)
+                return
+            except asyncio.TimeoutError:
+                # Continue the while loop, we use wait_for as an interval timer.
+                continue
--- a/src/cockpit/samples.py
+++ b/src/cockpit/samples.py
@@ -1,65 +1,307 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2022 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
 import os
-
-CPU_SAMPLER = 1 << 0
-MEMORY_SAMPLER = 1 << 1
-
-descriptions = {
-    'cpu.basic.nice': ('millisec', 'counter', False, CPU_SAMPLER),
-    'cpu.basic.user': ('millisec', 'counter', False, CPU_SAMPLER),
-    'cpu.basic.system': ('millisec', 'counter', False, CPU_SAMPLER),
-    'cpu.basic.iowait': ('millisec', 'counter', False, CPU_SAMPLER),
-
-    'cpu.core.nice': ('millisec', 'counter', True, CPU_SAMPLER),
-    'cpu.core.user': ('millisec', 'counter', True, CPU_SAMPLER),
-    'cpu.core.system': ('millisec', 'counter', True, CPU_SAMPLER),
-    'cpu.core.iowait': ('millisec', 'counter', True, CPU_SAMPLER),
-
-    'memory.free': ('bytes', 'instant', False, MEMORY_SAMPLER),
-    'memory.used': ('bytes', 'instant', False, MEMORY_SAMPLER),
-    'memory.cached': ('bytes', 'instant', False, MEMORY_SAMPLER),
-    'memory.swap-used': ('bytes', 'instant', False, MEMORY_SAMPLER),
-}
-
-
-# Friendly utility class to allow dynamically building the tree
-class Samples(dict):
-    def __getitem__(self, k):
-        if k not in self:
-            self[k] = Samples()
-        return self.get(k)
+import os.path
+import sys
+from typing import Any, Dict, List, NamedTuple, Optional
 
 
 USER_HZ = os.sysconf(os.sysconf_names['SC_CLK_TCK'])
 MS_PER_JIFFY = 1000 / (USER_HZ if (USER_HZ > 0) else 100)
 
 
-def cockpit_memory_samples(samples):
-    with open('/proc/meminfo') as meminfo:
-        items = {k: int(v.strip(' kB\n')) for line in meminfo for k, v in [line.split(':', 1)]}
-
-    samples['memory.free'] = 1024 * items['MemFree']
-    samples['memory.used'] = 1024 * (items['MemTotal'] - items['MemAvailable'])
-    samples['memory.cached'] = 1024 * (items['Buffers'] + items['Cached'])
-    samples['memory.swap-used'] = 1024 * (items['SwapTotal'] - items['SwapFree'])
+class SampleDescription(NamedTuple):
+    name: str
+    units: str
+    semantics: str
+    instanced: bool
 
 
-def cockpit_cpu_samples(samples):
-    with open('/proc/stat') as stat:
-        for line in stat:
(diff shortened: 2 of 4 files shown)


src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,148 @@
...
-        assert options['interval'] == 3000
-        assert 'omit-instances' not in options
-        assert options['metrics'] == [
-            {"name": "cpu.basic.user", "derive": "rate"},
-            {"name": "cpu.basic.system", "derive": "rate"},
-            {"name": "cpu.basic.nice", "derive": "rate"},
-            {"name": "memory.used"},
-        ]
+            try:
+                ret = await asyncio.wait_for(self.read(), self.interval / 1000)
```
jelly, line 158: @allisonkarlitskaya this never returns anything when the channel is closed so we keep sending metrics at the moment. You said something about `self.read` returning EOF but this doesn't seem to happen here.

jelly:
@martinpitt tagged you in the review, as you probably know how the current metrics work :)

jelly:
Moved the small fixes into https://github.com/cockpit-project/cockpit/pull/17694

src/cockpit/channels/metrics.py
allisonkarlitskaya: https://github.com/cockpit-project/cockpit/blob/main/src/cockpit/channel.py#L187

```python
    def do_done(self):
        self.receive_queue.put_nowait(b'')
```

the EOF gets written only if the other side says `done`.  We could probably modify the asyncchannel class to also send it on close.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,147 @@
...
+    desc: MetricDescription
+    derive: str
+    instances: List
+    value: Any
+
+
+class InternalMetricsChannel(AsyncChannel):
     payload = 'metrics1'
+    restrictions = {'source': 'internal'}
+    metrics: List[MetricInfo] = []
```
allisonkarlitskaya, line 50: I'd put a newline before this to separate it from the above data.

allisonkarlitskaya: Possibly also use `_` to show they're locals.

And in any case, don't assign `[]` there, please.  This is a dangerous value.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,147 @@
...
+            # Keep track of what samplers are requested
+            self.samplers |= sampler.value
+
+            desc = MetricDescription(name, unit, semantics, [], sampler.value)
+            self.metrics.append(MetricInfo(desc, derive, [], None))
+
+        self.ready()
+
+    async def run(self, options):
+        self.parse_options(options)
```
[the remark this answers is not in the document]
allisonkarlitskaya: ...I wonder if we should have some kind of a ChannelError that can be thrown as an exception from any channel function in order to close the channel, similar to how we have ProtocolError for the entire thing.  That would probably be pretty convenient.

allisonkarlitskaya (changes requested):

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,147 @@
...
+    # If any metric "subscriptions" change, we need to inform the client.
+    need_meta: bool = False
+    last_timestamp: int
+    next_timestamp: int
+
+    def parse_options(self, options):
+        logging.debug("metrics internal open: %s", options)
+        interval = options.get('interval', self.interval)
+        if interval <= 0 or interval > sys.maxsize:
+            logging.error(f'invalid "interval" value: {interval}')
```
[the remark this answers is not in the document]
jelly: Also the logging is a non no, logging should always use logging.info("thing", *kwargs)

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,147 @@
...
+import asyncio
+import json
+import sys
+import time
 import logging
+from typing import Any, List, NamedTuple
 
-from ..channel import Channel
+from ..channel import AsyncChannel
+from ..samples import Sampler, SAMPLE_DESCRIPTIONS
```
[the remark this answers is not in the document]
jelly: Should also imo rename the file then :)

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,192 @@
...
+            # Keep track of what samplers are requested
+            self.samplers |= sampler.value
+
+            desc = MetricDescription(name, unit, semantics, instanced, sampler.value)
+            self.metrics.append(MetricInfo(desc, derive, {}, None, None))
+
+        return True
+
+    def cockpit_internal_metrics_sample(self, metric, instance=None, value=None):
+        for index, metricinfo in enumerate(self.metrics):
```
allisonkarlitskaya, line 109: This loop seems inefficient.  I think you want some kind of a lookup table here, no?

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,192 @@
...
+                                                 value=value, old_value=metricinfo.value)
+
+    def send_updates(self):
+        data = []
+        metrics = []
+        timestamp = int(time.time())
+        self.next_timestamp = timestamp
+
+        if self.samplers & Sampler.CPU_SAMPLER.value:
+            cockpit_cpu_samples(self.cockpit_internal_metrics_sample)
```
allisonkarlitskaya, line 136: I'm not a fan of passing in a bound callback to the samples function.  Why not either:

 - add the sampler functions directly on the metrics class itself; or
 - define some sort of an interface for setting samples data and pass an object which implements that interface (possibly the samples object itself)

again: the naming here is a bit too "C" and not enough "py"

src/cockpit/channels/metrics.py
jelly: Lookup sounds like the way to go yes, I had this on my mind, forgot...

src/cockpit/channels/metrics.py
jelly: 1. I would love to keep the file small, honestly rename it to `internalmetrics.py`, but hmm seems doable, it does mean the class will be filled with:
* cpu
* memory
* cgroups
* disk

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,192 @@
...
+
+            # Keep track of what samplers are requested
+            self.samplers |= sampler.value
+
+            desc = MetricDescription(name, unit, semantics, instanced, sampler.value)
+            self.metrics.append(MetricInfo(desc, derive, {}, None, None))
+
+        return True
+
+    def cockpit_internal_metrics_sample(self, metric, instance=None, value=None):
```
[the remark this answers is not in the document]
jelly: True.... this is a bit of copy pasta from C

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,192 @@
...
-from ..channel import Channel
+from ..channel import AsyncChannel
+from .samples import Sampler, SAMPLE_DESCRIPTIONS
+from .samples import (cockpit_cpu_samples, cockpit_memory_samples, cockpit_cpu_temperature_samples)
 
 logger = logging.getLogger(__name__)
 
 
-class MetricsChannel(Channel):
+class Instance(NamedTuple):
```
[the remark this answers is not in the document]
jelly: In cockpit we have normal metrics which are:

`memory.usage` => 8

But there are also metrics which are "multi-valued" (maybe a better word). Such as `cpu.nice` which has multiple values as: `{ 0: 1, 1: 2,... }` So that's an "instance". The protocol also calls them `instances`

src/cockpit/channels/samples.py
```diff
@@ -0,0 +1,120 @@
...
+
+class Sampler(enum.Enum):
+    CPU_SAMPLER = 1 << 0
+    MEMORY_SAMPLER = 1 << 1
+    BLOCK_SAMPLER = 1 << 2
+    NETWORK_SAMPLER = 1 << 3
+    MOUNT_SAMPLER = 1 << 4
+    CGROUP_SAMPLER = 1 << 5
+    DISK_SAMPLER = 1 << 6
+    THERMAL_SAMPLER = 1 << 7
```
jelly, line 14: Use Enum.Flag or we make `self.samplers` a list of functions to call.

src/cockpit/channels/metrics.py
jelly: Ok, we just keep two dicts: one for the old values and one for new.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+                    cls.descriptions_cache[name] = desc
+
+    def parse_options(self, options):
+        sampler_classes = set()
+        logger.debug("metrics internal open: %s, channel: %s", options, self.channel)
+
+        interval = options.get('interval', self.interval)
+        if interval <= 0 or interval > sys.maxsize:
+            logger.error('invalid "interval" value: %s', interval)
+            self.close(problem="protocol-error", message='invalid interval option')
```
allisonkarlitskaya, line 76: . o O ( TODO: extend #17753 to work from `AsyncChannel.run()` )

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+            return False
+
+        self.interval = interval
+
+        metrics = options.get('metrics')
+        if not metrics or not isinstance(metrics, list):
```
allisonkarlitskaya, line 82: What is the logic that you're trying to implement here?  Because I read this backwards from what you probably intend.  Specifically:

 - I think you're trying to check for `None` left of the `and`, but you're also rejecting empty lists there (ie: rewrite to `is None`).
 - If it's really specifically `None` that you want to check, that could be handled with the `isinstance` check, since `None` is obviously not a `list`.

ie: if your intention is to check that it's a list *and* not empty, I'd rather more intuitively write:

`if not instance(metrics, list) or len(metrics) == 0:` or so

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+            cls.samplers_cache = {}
+            cls.descriptions_cache = {}
+            for sampler in SAMPLERS:
+                print(sampler)
+                for name, desc in sampler.descriptions.items():
+                    cls.samplers_cache[name] = sampler
+                    cls.descriptions_cache[name] = desc
+
+    def parse_options(self, options):
+        sampler_classes = set()
```
allisonkarlitskaya, line 70: Can you move this just above the loop that fills it in?

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+        logger.debug("metrics internal open: %s, channel: %s", options, self.channel)
+
+        interval = options.get('interval', self.interval)
+        if interval <= 0 or interval > sys.maxsize:
```
allisonkarlitskaya, line 74: That's gonna crash if `interval` isn't an int, right?

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+                self.close(problem='not-supported',
+                           message='invalid metric')
+                return False
+
+            if units and units != desc.units:
+                self.close(problem='protocol-error',
+                           message=f'{name} has units {desc.units}, not {units}')
+                return False
+
+            sampler_classes.add(self.samplers_cache.get(name))
```
allisonkarlitskaya, line 105: Your logic seems to be "if we found the description then surely the sampler also exists", but then you should just use the `[]` subscript notation instead of `.get()` (which would confusingly add `None` to the list of classes).

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+            self.close(problem='metrics is not an list')
+            return False
+
+        for metric in metrics:
+            # validate it's an object
+            name = metric.get('name')
+            units = metric.get('units')
+            derive = metric.get('derive')
+
+            desc = self.descriptions_cache.get(name)
```
allisonkarlitskaya, line 93: Since you're doing the explicit loop to deal with this, I think maybe it would make sense to have a single cache, which maps names to tuples of (class, description).

Also: it's more pythonic to use `[]` and catch the `KeyError`.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,179 @@
...
+                           message='invalid metric')
+                return False
+
+            if units and units != desc.units:
+                self.close(problem='protocol-error',
+                           message=f'{name} has units {desc.units}, not {units}')
+                return False
+
+            sampler_classes.add(self.samplers_cache.get(name))
+            self.metrics.append(MetricInfo(name=name, derive=derive, semantics=desc.semantics, units=desc.units, instanced=desc.instanced))
```
allisonkarlitskaya, line 106: I wonder if it wouldn't make sense to have a `MetricInfo` constructor which took a `SampleDescription` and a `derive` parameter (since the `derive` parameter seems to be the only thing here which isn't part of the description).

You could also make good use of some syntax like

```python3
MetricInfo(**desc._asdict(), derive=derive)
```

See below for comments about how to deal with the `name` field.

One more note: `MetricInfo` seems to be more or less the same as `SampleDescription` with some extra fields.  Does a subtype relationship make sense here?  Or maybe a containment relationship?  I played a bit with namedtuple subtyping, and I'd lean towards containment based on my findings...

I don't like the duplication of the 3 or 4 fields, basically...

src/cockpit/samples.py
```diff
@@ -1,65 +1,180 @@
...
-            samples[f'{prefix}.iowait'][core] = int(iowait) * MS_PER_JIFFY
-
-
-s = Samples()
-cockpit_memory_samples(s)
-cockpit_cpu_samples(s)
-
-print(s)
+class SampleDescription(NamedTuple):
+    units: str
```
allisonkarlitskaya, line 28: Now that we're just doing a linear scan of all of the descriptions anyway, why not make the name part of the description, and turn the whole thing into a list?

ie, not:
```python
{
    'a.b.c': Description('ms', ...),
    'd.e.f': Description('ms', ...),
}
```

but

```python
[
    Description('a.b.c', 'ms', ...),
    Description('d.e.f', 'ms', ...),
]
```

This reads a bit nicer to me, and having the name built into the description will better match `MetricInfo` and help you construct those (see comment above).

allisonkarlitskaya (changes requested):
This is already starting to look a whole lot better.  Some more ideas/suggestions below.

jelly:
image

Maybe counters are wrong or we don't convert units, but there is data...

jelly:
> image
> 
> Maybe counters are wrong or we don't convert units, but there is data...

We don't handle "rate" metrics for instanced metrics, so that's the bo boo. Wrote the code doesn't work, so let's postpone that for later....

allisonkarlitskaya (changes requested):
Can you please try rebasing on top of #17757?

jelly:
> Can you please try rebasing on top of #17757?

Done, applied ChannelErrors everywhere and it's much nicer then returning a bool and `self.close`. However I do not see any close message in my browser console, so it seems something is broken. I'd like to fix that in a follow up and write some unit-tests with a real python bridge for the metrics channel.

jelly:
Uhhh *facepalm*:

```
2514   │     if isinstance(interval, int) and interval <= 0 or interval > sys.maxsize:
2515   │ TypeError: '>' not supported between instances of 'str' and 'int'
```

src/cockpit/samples.py
```diff
@@ -1,65 +1,300 @@
...
+class NetworkSampler(BaseSampler):
+    descriptions = [
+        SampleDescription('network.interface.tx', 'bytes', 'counter', True),
+        SampleDescription('network.interface.rx', 'bytes', 'counter', True),
+    ]
+
+    def sample(self, samples):
+        with open("/proc/net/dev") as network_samples:
+            for line in network_samples:
+                lines = line.split()
```
allisonkarlitskaya, line 249: huh?

You iterate the file line-by-line, but then you split each line up and store it into a variable called "lines".  I think either this isn't doing what you think, or I missed something.

allisonkarlitskaya (changes requested):

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,184 @@
...
+            'timestamp': timestamp * 1000,
+            'interval': self.interval,
+            'source': 'internal',
+            'metrics': metrics
+        }
+        self.send_message(**meta)
+        self.need_meta = False
+
+    def sample(self):
+        self.old_samples = copy.deepcopy(self.samples)
```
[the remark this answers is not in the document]
jelly: We need a deepcopy to compare old and new values, with `.copy()` we would end up with a shallow copy.

src/cockpit/samples.py
```diff
@@ -1,65 +1,300 @@
...
+            self.cgroups_v2 = os.path.exists('/sys/fs/cgroup/cgroup.controllers')
+
+        if self.cgroups_v2_path:
+            for path, _, _ in os.walk(self.cgroups_v2_path):
+                cgroup = path.replace(self.cgroups_v2_path, '')  # TODO: Pathlib?
+
+                if not cgroup:
+                    continue
+
+                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.current'), cgroup, 'cgroup.memory.usage')
```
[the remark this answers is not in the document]
jelly: Can we do this in a follow up?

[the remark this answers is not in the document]
allisonkarlitskaya: Yes.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,176 @@
...
+
+    interval: int = 1000
+    need_meta: bool = True
+    last_timestamp: float = 0
+    next_timestamp: float = 0
+
+    @classmethod
+    def ensure_samplers(cls):
+        if cls.samplers_cache is None:
+            cls.samplers_cache = {}
```
allisonkarlitskaya, line 58: This could totally be a dictionary comprehension, but I'll let it slide :wink:

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,176 @@
...
+
+        sampler_classes = set()
+        for metric in metrics:
+            # validate it's an object
+            name = metric.get('name')
+            units = metric.get('units')
+            derive = metric.get('derive')
+
+            try:
+                (sampler, desc) = self.samplers_cache[name]
```
allisonkarlitskaya, line 85: Without the parens is more usual.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,184 @@
...
+        for metricinfo in self.metrics:
+            value = self.samples[metricinfo.desc.name]
+            old_value = self.old_samples[metricinfo.desc.name]
+
+            if metricinfo.desc.instanced:
+                if metricinfo.derive == "rate":
+                    samples = []
+                    for key, val in value.items():
+                        if key not in old_value:
+                            self.need_meta = True
```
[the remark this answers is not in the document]
allisonkarlitskaya: Sure.   But then you should surely send the meta for the cgroup metrics, and not absolutely everything, no?

src/cockpit/samples.py
```diff
@@ -1,65 +1,308 @@
...
+    ]
+
+    def sample(self, samples):
+        with open("/proc/net/dev") as network_samples:
+            for line in network_samples:
+                if len(line) <= 2:
+                    continue
+
+                # Skip header line
+                if line[0][-1] != ':':
```
allisonkarlitskaya, line 253: This still doesn't make sense to me.  `line` is a string.  You're taking the first character of that string, and then taking the last character of that first character...

src/cockpit/samples.py
```diff
@@ -1,65 +1,308 @@
...
-print(s)
+SAMPLERS = [
+    BlockSampler,
+    CGroupSampler,
+    CPUSampler,
+    CPUTemperatureSampler,
+    DiskSampler,
+    MemorySampler,
+    MountSampler,
+    NetworkSampler
```
allisonkarlitskaya, line 307: always trailing comma in lists like this, please

allisonkarlitskaya (changes requested):

src/cockpit/channels/metrics.py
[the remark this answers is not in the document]
jelly: You need to send everything again.

src/cockpit/channels/metrics.py
```diff
@@ -15,23 +15,173 @@
...
+        self.last_samples = defaultdict(dict)
+
+        InternalMetricsChannel.ensure_samplers()
+
+        self.parse_options(options)
+        self.ready()
+
+        while True:
+            self.sample()
+            self.send_updates()
```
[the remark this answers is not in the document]
jelly: I'd say None is an antipattern, let's just assign it to a dict.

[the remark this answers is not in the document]
allisonkarlitskaya: Works for me.

allisonkarlitskaya (approved):
I tried it out and it's working nicely with Cockpit Client on my laptop.

Nice work!

PR #17683, opened 2022-08-29
test: allow root user to be named "Super User"

allisonkarlitskaya:
fedora-coreos recently changed the "real name" of the root user to
"Super User".  Allow that as a valid value.


mvollmer:
Oh, should we maybe use `getent` to get the expected value?

mvollmer (approved):

mvollmer (approved):

PR #17692, opened 2022-08-31
Metrics: Add cgroup disk IO usage

tomasmatus:
## Metrics: Add per-cgroup disk IO usage

output

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -171,6 +171,11 @@ const CPU_TEMPERATURE_METRICS = [
     { name: "cpu.temperature" },
 ];
 
+const PRIVILEGED_METRICS = [
+    { name: "disk.cgroup.read", units: "bytes", derive: "rate" },
+    { name: "disk.cgroup.written", units: "bytes", derive: "rate" },
+];
+
 const HISTORY_METRICS = [
     // CPU utilization
     { name: "kernel.all.cpu.nice", derive: "rate" },
@@ -243,11 +248,14 @@ class CurrentMetrics extends React.Component {
 
         this.metrics_channel = null;
         this.temperature_channel = null;
+        this.privileged_channel = null;
         this.samples = [];
         this.temperatureSamples = [];
+        this.privilegedSamples = [];
         this.netInterfacesNames = [];
         this.cgroupCPUNames = [];
         this.cgroupMemoryNames = [];
+        this.cgroupDiskNames = [];
         this.disksNames = [];
         this.cpuTemperatureColors = {
             textColor: "",
@@ -269,12 +277,14 @@ class CurrentMetrics extends React.Component {
             netInterfacesTx: [],
             topServicesCPU: [], // [ { name, percent } ]
             topServicesMemory: [], // [ { name, bytes } ]
+            topServicesDiskIO: [], // [ [ name, read, write ] ]
             podNameMapping: {}, // { uid -> containerid -> name }
         };
 
         this.onVisibilityChange = this.onVisibilityChange.bind(this);
         this.onMetricsUpdate = this.onMetricsUpdate.bind(this);
         this.onTemperatureUpdate = this.onTemperatureUpdate.bind(this);
+        this.onPrivilegedMetricsUpdate = this.onPrivilegedMetricsUpdate.bind(this);
         this.updateMounts = this.updateMounts.bind(this);
         this.updateLoad = this.updateLoad.bind(this);
 
@@ -300,6 +310,12 @@ class CurrentMetrics extends React.Component {
             this.temperature_channel = null;
         }
 
+        if (cockpit.hidden && this.privileged_channel !== null) {
+            this.privileged_channel.removeEventListener("message", this.onPrivilegedMetricsUpdate);
+            this.privileged_channel.close();
+            this.privileged_channel = null;
+        }
+
         if (cockpit.hidden && this.metrics_channel !== null) {
             this.metrics_channel.removeEventListener("message", this.onMetricsUpdate);
             this.metrics_channel.close();
@@ -311,6 +327,13 @@ class CurrentMetrics extends React.Component {
             this.temperature_channel = cockpit.channel({ payload: "metrics1", source: "internal", interval: INTERVAL, metrics: CPU_TEMPERATURE_METRICS });
             this.temperature_channel.addEventListener("close", (ev, error) => console.error("CPU temperature metric closed:", error));
             this.temperature_channel.addEventListener("message", this.onTemperatureUpdate);
+        }
+
+        // requires sudo access in order to sample every cgroup
+        // limited access only allows sampling of current user's cgroups
+        if (!cockpit.hidden && (this.privileged_channel === null)) {
+            this.privileged_channel = cockpit.channel({ superuser: "try", payload: "metrics1", source: "internal", interval: INTERVAL, metrics: PRIVILEGED_METRICS });
+            this.privileged_channel.addEventListener("message", this.onPrivilegedMetricsUpdate);
         }
 
         if (!cockpit.hidden && this.metrics_channel === null) {
@@ -507,61 +530,139 @@ class CurrentMetrics extends React.Component {
             return merged.slice(0, n);
         };
 
-        const getCachedPodName = (uid, containerid) => this.state.podNameMapping[uid] && this.state.podNameMapping[uid][containerid];
-
-        function cgroupClickHandler(name, is_user, is_container, uid) {
-            if (is_container) {
-                const container_name = getCachedPodName(uid, name);
-                if (container_name) {
-                    cockpit.jump("/podman#/?name=" + container_name);
-                } else {
-                    cockpit.jump("/podman");
-                }
-            } else {
-                cockpit.jump("/system/services#/" + name + ".service" + (is_user ? "?owner=user" : ""));
-            }
-        }
-
-        function cgroupRow(name, value, is_user, is_container, uid) {
-            const podman_installed = cockpit.manifests && cockpit.manifests.podman;
-            let name_text = (
-                <Button variant="link" isInline component="a" key={name}
-                        onClick={() => cgroupClickHandler(name, is_user, is_container, uid)}
-                        isDisabled={is_container && !podman_installed}>
-                    <TableText wrapModifier="truncate">
-                        {is_container ? _("pod") + " " + (getCachedPodName(uid, name) || name.substr(0, 12)) : name}
-                    </TableText>
-                </Button>
-            );
-            if (is_container && !podman_installed) {
-                name_text = (
-                    <Tooltip content={_("cockpit-podman is not installed")} key={name + "_tooltip"}>
-                        <div>
-                            {name_text}
-                        </div>
-                    </Tooltip>);
-            }
-            const value_text = <TableText wrapModifier="nowrap">{value}</TableText>;
-            return [name_text, value_text];
-        }
-
         // top 5 CPU and memory consuming systemd units
         const topServicesCPU = n_biggest(this.cgroupCPUNames, this.samples[9], 5);
         newState.topServicesCPU = topServicesCPU.map(
-            ([key, value, is_user, is_container, userid]) => cgroupRow(key, Number(value / 10 / numCpu).toFixed(1), is_user, is_container, userid) // usec/s → percent
+            ([key, value, is_user, is_container, userid]) => this.cgroupRow(key, is_user, is_container, userid, Number(value / 10 / numCpu).toFixed(1)) // usec/s → percent
         );
 
         const topServicesMemory = n_biggest(this.cgroupMemoryNames, this.samples[10], 5);
         newState.topServicesMemory = topServicesMemory.map(
-            ([key, value, is_user, is_container, userid]) => cgroupRow(key, cockpit.format_bytes(value), is_user, is_container, userid)
+            ([key, value, is_user, is_container, userid]) => this.cgroupRow(key, is_user, is_container, userid, cockpit.format_bytes(value))
         );
 
-        const notMappedContainers = topServicesMemory.concat(topServicesCPU).filter(([key, value, is_user, is_container, userid]) => is_container && getCachedPodName(userid, key) === undefined);
+        const notMappedContainers = topServicesMemory.concat(topServicesCPU).filter(([key, value, is_user, is_container, userid]) => is_container && this.getCachedPodName(userid, key) === undefined);
         if (notMappedContainers.length !== 0) {
             this.update_podman_name_mapping(notMappedContainers);
         }
         this.setState(newState);
     }
+
+    onPrivilegedMetricsUpdate(event, message) {
+        debug("process metrics message", message);
+        const data = JSON.parse(message);
+
+        if (!Array.isArray(data)) {
+            this.cgroupDiskNames = data.metrics[0].instances.slice();
+            return;
+        }
+
+        data.forEach(privilegedSamples => decompress_samples(privilegedSamples, this.privilegedSamples));
+
+        // return [ name, read, write, isUser, isContainer, uid | cgroup ]
+        const n_biggest = (n, names, valuesA, valuesB) => {
+            const merged = [];
+            const userSlices = {};
+            names.forEach((name, i) => {
+                // filter out invalid values, the empty (root) cgroup, non-services
+                if (name.endsWith('.service')) {
+                    // only keep cgroup basenames, and drop redundant .service suffix
+                    const label = name.replace(/.*\//, '').replace(/\.service$/, '');
+                    const isUser = name.match(useridCgroupRe);
+                    merged.push([label, valuesA[i], valuesB[i], isUser, false, name]);
+                    return;
+                }
+
+                // filter out podman containers, but only for the logged in
+                // user or root user if the user is privileged. Other users
+                // containers will show up under the user@uid cgroup
+                const matches = name.match(/libpod-(?<containerid>[a-z|0-9]{64})\.scope/);
+                if (matches && valuesA[i] !== undefined && valuesB[i] !== undefined) {
+                    const containerid = matches.groups.containerid;
+                    const umatches = name.match(useridCgroupRe);
+                    const isUser = !!umatches;
+                    const uid = parseInt(umatches?.groups.userid) || 0;
+
+                    if (uid === 0 || this.state.userid == uid) {
+                        merged.push([containerid, valuesA[i], valuesB[i], isUser, true, uid]);
+                        return;
+                    }
+                }
+
+                // combine user slices into user@ID
+                // { name: [ read, write ] }
+                const umatches = name.match(useridCgroupRe);
+                if (umatches) {
+                    if (userSlices[umatches.groups.userid] === undefined) {
+                        userSlices[umatches.groups.userid] = [valuesA[i], valuesB[i]];
+                    } else {
+                        userSlices[umatches.groups.userid][0] += valuesA[i];
+                        userSlices[umatches.groups.userid][1] += valuesB[i];
+                    }
+                }
+            });
+
+            Object.keys(userSlices).forEach((key) => {
+                merged.push(["user@" + key, userSlices[key][0], userSlices[key][1], false, false]);
+            });
+
+            // sort by overall (read + write) disk usage
+            merged.sort((a, b) => (b[1] + b[2]) - (a[1] + a[2]));
+            return merged.slice(0, n);
+        };
+
+        const newState = {};
+
+        const topServicesDiskIO = n_biggest(5, this.cgroupDiskNames, this.privilegedSamples[0], this.privilegedSamples[1]);
+        newState.topServicesDiskIO = topServicesDiskIO.filter(([_, read, write, ..._rest]) => read !== 0 || write !== 0).map(([name, read, write, isUser, isContainer, uid]) => {
+            return this.cgroupRow(name, isUser, isContainer, uid, read > 1 ? cockpit.format_bytes_per_sec(read) : 0, write > 1 ? cockpit.format_bytes_per_sec(write) : 0);
+        });
+
+        this.setState(newState);
+    }
+
+    getCachedPodName = (uid, containerid) => this.state.podNameMapping[uid] && this.state.podNameMapping[uid][containerid];
+
+    cgroupRow = (name, is_user, is_container, uid, ...values) => {
+        const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
+                }
+            } else {
+                cockpit.jump("/system/services#/" + name + ".service" + (isUser ? "?owner=user" : ""));
+            }
+        };
+
+        let name_text = (
+            <Button variant="link" isInline component="a" key={name}
+                    onClick={() => cgroupClickHandler(name, is_user, is_container, uid)}
+                    isDisabled={is_container && !podman_installed}>
+                <TableText wrapModifier="truncate">
+                    {is_container ? _("pod") + " " + (this.getCachedPodName(uid, name) || name.substr(0, 12)) : name}
+                </TableText>
+            </Button>
+        );
+        if (is_container && !podman_installed) {
+            name_text = (
+                <Tooltip content={_("cockpit-podman is not installed")} key={name + "_tooltip"}>
+                    <div>
+                        {name_text}
+                    </div>
+                </Tooltip>);
+        }
+
+        const values_text = values.map((value, idx) => {
+            return <TableText key={idx} wrapModifier="nowrap">{value}</TableText>;
+        });
+
+        return [name_text, ...values_text];
+    };
 
     /**
      * Look up the container names using podman ps for the given cgroups.
@@ -706,6 +807,7 @@ class CurrentMetrics extends React.Component {
         }
 
         const rowWrapperIface = row => ({ 'data-interface': row[0] });
+        const rowWrapperDiskIO = row => ({ 'cgroup-name': row[0] });
         const topServicesCPUColumns = [_("Service"), "%"];
         const topServicesMemoryColumns = [_("Service"), _("Used")];
         const ifaceColumns = [_("Interface"), _("In"), _("Out")];
@@ -852,6 +954,23 @@ class CurrentMetrics extends React.Component {
                             })
                         }
                         </div>
+                        { this.state.topServicesDiskIO.length > 0 &&
+                            <Table
+                                variant={TableVariant.compact}
+                                gridBreakPoint={TableGridBreakpoint.none}
+                                borders={false}
+                                aria-label={ _("Top 5 disk usage services") }>
+                                <Thead>
+                                    <Tr>
+                                        <Th width={50}>{_("Service")}</Th>
+                                        <Th>{_("Read")}</Th>
+                                        <Th>{_("Write")}</Th>
+                                    </Tr>
+                                </Thead>
+                                <Tbody className="pf-m-tabular-nums">
+                                    {make_rows(this.state.topServicesDiskIO, rowWrapperDiskIO, [_("Service"), _("Read"), _("Write")])}
+                                </Tbody>
+                            </Table> }
                     </CardBody>
                 </Card>
 
--- a/src/bridge/cockpitdisksamples.c
+++ b/src/bridge/cockpitdisksamples.c
@@ -21,9 +21,16 @@
 
 #include "cockpitdisksamples.h"
 
+#include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
+
+typedef struct {
+  guint64 disk_read;
+  guint64 disk_write;
+} cgroup_values_t;
 
 /* TODO: this should be optimized so we don't allocate network and call open()/close() all the time */
 
@@ -152,4 +159,132 @@ cockpit_disk_samples (CockpitSamples *samples)
 out:
   g_strfreev (lines);
   g_free (contents);
+}
+
+static FILE *
+open_file (int dirfd,
+           const char *name)
+{
+  const int fd = openat (dirfd, name, O_RDONLY);
+  if (fd < 0)
+    {
+      if (errno != EACCES && errno != ESRCH && errno != ENOENT)
+        g_message ("error opening file descriptor: %s: %m", name);
+      return NULL;
+    }
+
+  FILE *fp = fdopen (fd, "r");
+  if (!fp)
+    {
+      if (errno != ESRCH && errno != ENOENT)
+        g_message ("error opening file %s: %m", name);
+
+      close (fd);
+      return NULL;
+    }
+
+  return fp;
+}
+
+static void
+table_add_values(GHashTable *table,
+                 const gchar *cgroup,
+                 guint64 disk_read,
+                 guint64 disk_write)
+{
+  cgroup_values_t *values = g_hash_table_lookup (table, cgroup);
+  if (values == NULL)
+    {
+      values = g_new0(cgroup_values_t, 1);
+      g_hash_table_insert (table, g_strdup (cgroup), values);
+    }
+
+  values->disk_read += disk_read;
+  values->disk_write += disk_write;
+}
+
+static void
+get_process_io (const int dirfd,
+                GHashTable *table)
+{
+  FILE *io_fp = open_file (dirfd, "io");
+  if (!io_fp)
+    return;
+
+  gchar *key;
+  guint64 disk_read = 0, disk_write = 0, value = 0;
+  while (fscanf (io_fp, "%m[^: ]: %" G_GUINT64_FORMAT "\n", &key, &value) == 2)
+    {
+      if (g_str_equal (key, "read_bytes"))
+        disk_read = value;
+      else if (g_str_equal (key, "write_bytes"))
+        disk_write = value;
+
+      free (key);
+    }
+
+  fclose (io_fp);
+
+  // get process cgroup
+  FILE *cgroup_fp = open_file (dirfd, "cgroup");
+  if (!cgroup_fp)
+    return;
+
+  gchar *cgroup = NULL;
+  if (fscanf (cgroup_fp, "%ms\n", &cgroup) != 1)
+    g_debug ("Failed to read cgroup name: %m");
+  else
+    table_add_values (table, cgroup, disk_read, disk_write);
+
+  g_free (cgroup);
+  fclose (cgroup_fp);
+}
+
+void
+cockpit_cgroup_disk_usage (CockpitSamples *samples)
+{
+  DIR *d = opendir ("/proc");
+  if (!d)
+    {
+      g_warning ("Error when opening /proc, %m");
+      return;
+    }
+  int proc_fd = dirfd (d);
+
+  GHashTable *table = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+  struct dirent *de;
+  while ((de = readdir (d)))
+    {
+      // non-pid entries in proc are guaranteed to start with a character a-z
+      if (de->d_name[0] < '0' || '9' < de->d_name[0])
+        continue;
+
+      int dfd = openat(proc_fd, de->d_name, O_PATH | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);
+      if (dfd < 0)
+      {
+        if (errno != ENOENT)
+          g_message ("failed to open /proc/%s, %m", de->d_name);
+        continue;
+      }
+      get_process_io (dfd, table);
+      close (dfd);
+    }
+
+  closedir (d);
+
+  GHashTableIter iter;
+  gpointer key, value;
+  g_hash_table_iter_init (&iter, table);
+  while (g_hash_table_iter_next (&iter, &key, &value))
+    {
+      const gchar *cgroup_name = key;
+      const cgroup_values_t *values = value;
+
+      // Skip ::0/
+      cgroup_name += 4;
+
+      cockpit_samples_sample ((CockpitSamples *)samples, "disk.cgroup.read", cgroup_name, values->disk_read);
+      cockpit_samples_sample ((CockpitSamples *)samples, "disk.cgroup.written", cgroup_name, values->disk_write);
+    }
+  g_hash_table_unref (table);
 }
--- a/src/cockpit/samples.py
+++ b/src/cockpit/samples.py
@@ -16,9 +16,10 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import os
+import re
 from ._vendor.systemd_ctypes import Handle
 
-from typing import Any, DefaultDict, Iterable, List, NamedTuple, Optional
+from typing import Any, DefaultDict, Iterable, List, NamedTuple, Optional, Tuple
 
 
 USER_HZ = os.sysconf(os.sysconf_names['SC_CLK_TCK'])
@@ -301,6 +302,56 @@ class CGroupSampler(Sampler):
                     samples['cgroup.cpu.usage'][cgroup] = usage_nsec / 1000000
 
 
+class CGroupDiskIO(Sampler):
+    IO_RE = re.compile(rb'\bread_bytes: (?P<read>\d+).*\nwrite_bytes: (?P<write>\d+)', flags=re.S)
+    descriptions = [
+        SampleDescription('disk.cgroup.read', 'bytes', 'counter', True),
+        SampleDescription('disk.cgroup.written', 'bytes', 'counter', True),
+    ]
+
+    @staticmethod
+    def get_cgroup_name(fd: int) -> str:
+        with Handle.open('cgroup', os.O_RDONLY, dir_fd=fd) as cgroup_fd:
+            cgroup_name = os.read(cgroup_fd, 2048).decode().strip()
+
+            # Skip leadin ::0/
+            return cgroup_name[4:]
+
+    @staticmethod
+    def get_proc_io(fd: int) -> Tuple[int, int]:
+        with Handle.open('io', os.O_RDONLY, dir_fd=fd) as io_fd:
+            data = os.read(io_fd, 4096)
+
+            match = re.search(CGroupDiskIO.IO_RE, data)
+            if match:
+                proc_read = int(match.group('read'))
+                proc_write = int(match.group('write'))
+
+                return proc_read, proc_write
+
+            return 0, 0
+
+    def sample(self, samples: Samples):
+        with Handle.open('/proc', os.O_RDONLY | os.O_DIRECTORY) as proc_fd:
+            reads = samples['disk.cgroup.read']
+            writes = samples['disk.cgroup.written']
+
+            for path in os.listdir(proc_fd):
+                # non-pid entries in proc are guaranteed to start with a character a-z
+                if path[0] < '0' or path[0] > '9':
+                    continue
+
+                try:
+                    with Handle.open(path, os.O_PATH, dir_fd=proc_fd) as pid_fd:
+                        cgroup_name = self.get_cgroup_name(pid_fd)
+                        proc_read, proc_write = self.get_proc_io(pid_fd)
+                except (FileNotFoundError, PermissionError):
+                    continue
+
+                reads[cgroup_name] = reads.get(cgroup_name, 0) + proc_read
+                writes[cgroup_name] = writes.get(cgroup_name, 0) + proc_write
+
+
 class NetworkSampler(Sampler):
     descriptions = [
         SampleDescription('network.interface.tx', 'bytes', 'counter', True),
@@ -362,6 +413,7 @@ class BlockSampler(Sampler):
 SAMPLERS = [
     BlockSampler,
     CGroupSampler,
+    CGroupDiskIO,
     CPUSampler,
     CPUTemperatureSampler,
     DiskSampler,
--- a/src/bridge/cockpitdisksamples.h
+++ b/src/bridge/cockpitdisksamples.h
@@ -26,6 +26,7 @@ G_BEGIN_DECLS
 
 void            cockpit_disk_samples         (CockpitSamples *samples);
 
+void            cockpit_cgroup_disk_usage    (CockpitSamples *samples);
 
 G_END_DECLS
 
--- a/src/bridge/cockpitinternalmetrics.c
+++ b/src/bridge/cockpitinternalmetrics.c
@@ -55,6 +55,7 @@ typedef enum {
   CGROUP_SAMPLER = 1 << 5,
   DISK_SAMPLER = 1 << 6,
   THERMAL_SAMPLER = 1 << 7,
+  CGROUP_IO_SAMPLER = 1 << 8,
 } SamplerSet;
 
 typedef struct {
@@ -104,6 +105,9 @@ static MetricDescription metric_descriptions[] = {
   { "cgroup.cpu.shares",      "count",    "instant", TRUE, CGROUP_SAMPLER },
 
   { "cpu.temperature",        "celsius",  "instant", TRUE, THERMAL_SAMPLER },
+
+  { "disk.cgroup.read",    "bytes", "counter", TRUE, CGROUP_IO_SAMPLER },
+  { "disk.cgroup.written", "bytes", "counter", TRUE, CGROUP_IO_SAMPLER },
 
   { NULL }
 };
@@ -341,6 +345,8 @@ cockpit_internal_metrics_tick (CockpitMetrics *metrics,
     cockpit_disk_samples (COCKPIT_SAMPLES (self));
   if (self->samplers & THERMAL_SAMPLER)
     cockpit_cpu_temperature (COCKPIT_SAMPLES (self));
+  if (self->samplers & CGROUP_IO_SAMPLER)
+    cockpit_cgroup_disk_usage (COCKPIT_SAMPLES (self));
 
   /* Check for disappeared instances
    */
--- a/src/bridge/test-metrics.c
+++ b/src/bridge/test-metrics.c
@@ -902,6 +902,65 @@ test_cpu_temperature (void)
     }
 }
 
+static void
+test_cgroup_disk_io (void)
+{
+  MockTransport *transport = mock_transport_new ();
+
+  g_signal_connect (transport, "closed", G_CALLBACK (on_transport_closed), NULL);
+
+  g_autoptr(JsonObject) options = json_obj ("{ 'metrics': [ { 'name': 'disk.cgroup.read' }, { 'name': 'disk.cgroup.written' } ],"
+                                  "  'interval': 1000"
+                                  "}");
+
+  CockpitChannel *channel = g_object_new (cockpit_internal_metrics_get_type (),
+                          "transport", transport,
+                          "id", "1234",
+                          "options", options,
+                          NULL);
+
+  cockpit_channel_prepare (channel);
+
+  g_autoptr(GBytes) msg = NULL;
+  /* receive meta information */
+  while ((msg = mock_transport_pop_channel (transport, "1234")) == NULL)
+    g_main_context_iteration (NULL, TRUE);
+
+  g_autoptr(GError) error = NULL;
+  g_autoptr(JsonObject) res = cockpit_json_parse_bytes (msg, &error);
+  g_assert_no_error (error);
+  g_assert (res != NULL);
+  g_assert (json_object_has_member (res, "metrics"));
+
+  JsonArray *metrics = json_object_get_array_member (res, "metrics");
+  g_assert_cmpint (json_array_get_length (metrics), ==, 2);
+
+  JsonObject *description = json_array_get_object_element (metrics, 0);
+  g_assert (description);
+  g_assert_cmpstr (json_object_get_string_member (description, "name"), ==, "disk.cgroup.read");
+  g_assert_cmpstr (json_object_get_string_member (description, "units"), ==, "bytes");
+
+  description = json_array_get_object_element (metrics, 1);
+  g_assert (description);
+  g_assert_cmpstr (json_object_get_string_member (description, "name"), ==, "disk.cgroup.written");
+  g_assert_cmpstr (json_object_get_string_member (description, "units"), ==, "bytes");
+
+  g_autoptr(JsonArray) samples = recv_array (transport);
+  g_assert_cmpint (json_array_get_length (samples), ==, 1);
+  JsonArray *core = json_array_get_array_element (samples, 0);
+  g_assert_cmpint (json_array_get_length (core), ==, 2);
+  JsonArray *read = json_array_get_array_element (core, 0);
+  JsonArray *write = json_array_get_array_element (core, 1);
+  guint length = json_array_get_length (read);
+  g_assert_cmpint (length, ==, json_array_get_length (write));
+  g_assert_cmpint (length, >, 0);
+  for (guint i = 0; i < length; i++)
+    {
+      g_assert_cmpint (json_array_get_int_element (read, i), >=, 0);
+      g_assert_cmpint (json_array_get_int_element (write, i), >=, 0);
+    }
+}
+
 int
 main (int argc,
       char *argv[])
@@ -932,6 +991,8 @@ main (int argc,
 
   g_test_add_func ("/metrics/cpu-cores", test_cpu_cores);
   g_test_add_func ("/metrics/cpu-temperature", test_cpu_temperature);
+
+  g_test_add_func ("/metrics/cgroup-disk-io", test_cgroup_disk_io);
 
   return g_test_run ();
 }
--- a/test/pytest/test_samples.py
+++ b/test/pytest/test_samples.py
@@ -152,3 +152,13 @@ class TestSamples(unittest.TestCase):
             expected = ['hwmon4/temp4_input', 'hwmon4/temp3_input', 'hwmon4/temp2_input', 'hwmon4/temp1_input', 'hwmon3/temp3_input', 'hwmon2/temp1_input']
             sensors = [os.path.join(os.path.basename(os.path.dirname(p)), os.path.basename(p)) for p in samples['cpu.temperature']]
             assert sorted(sensors) == sorted(expected)
+
+    def test_cgroup_disk_io(self):
+        samples = collections.defaultdict(dict)
+        cockpit.samples.CGroupDiskIO().sample(samples)
+        samples = self.get_checked_samples(cockpit.samples.CGroupDiskIO())
+
+        assert len(samples['disk.cgroup.read']) == len(samples['disk.cgroup.written'])
+        for cgroup in samples['disk.cgroup.read']:
+            assert samples['disk.cgroup.read'][cgroup] >= 0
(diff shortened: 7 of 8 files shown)


martinpitt:
Hmm, this is tricky. [cgroupv2](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html) certainly has an "io" controller which can do this per-service, but it's not enabled by default at least in Fedora. The approach here is *very* expensive, thus things like cgroup default config or PCP really shouldn't collect this all the time. On our metrics page we can afford to do more expensive operations, as that isn't running all the time.

But you are right that this makes things inconsistent -- showing services (cgroups) in two cards, and processes here. Showing the process names is not actionable -- we can't link them anywhere (like the services page), and the current version does not even show a pid number so that the admin can kill or investigate.

However, it is reasonably simple to map a pid to its containing cgroup by reading `/proc/pid/cgroup`. This may also reduce the amount of data a bit -- currently it sends this very large array of processes. I suppose compression takes place (null values for processes which are quiet and thus don't have any I/O), but it's still quite taxing on the bridge/websocket I/O?

tomasmatus:
> current version does not even show a pid number so that the admin can kill or investigate

pid is there in the code but it's not visible in the UI yet

> However, it is reasonably simple to map a pid to its containing cgroup by reading /proc/pid/cgroup. This may also reduce the amount of data a bit

So what you're saying is: read IO of processes and sum it for its cgroup, then send it?

martinpitt:
> So what you're saying is: read IO of processes and sum it for its cgroup, then send it?

Yes, I think that would make sense. Note that all of this would be significantly easier in the Python bridge. However, that is probably still 6 months out..

tomasmatus:
> all of this would be significantly easier in the Python bridge. However, that is probably still 6 months out..

Oh it sure would! And I hope it's sooner than that 😃

pkg/metrics/metrics.jsx
```diff
@@ -370,6 +391,81 @@ class CurrentMetrics extends React.Component {
...
+                const matches = name.match(/libpod-(?<containerid>[a-z|0-9]{64})\.scope/);
+                if (matches && valuesA[i] && valuesB[i]) {
+                    let isUser = false;
+                    let uid = 0;
+                    const containerid = matches.groups.containerid;
+                    const umatches = name.match(useridCgroupRe);
+                    if (umatches) {
+                        isUser = true;
+                        uid = parseInt(umatches.groups.userid);
+                    }
```
cockpituous, line 431: These 9 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220907-112243-c768a0f4-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -370,6 +391,81 @@ class CurrentMetrics extends React.Component {
...
+
+                    if (uid === 0 || this.state.userid == uid) {
+                        merged.push([containerid, valuesA[i], valuesB[i], isUser, true, uid]);
+                        return;
+                    }
+                }
```
cockpituous, line 437: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220907-112243-c768a0f4-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -535,6 +593,56 @@ class CurrentMetrics extends React.Component {
...
+
+        function cgroupClickHandler(name, isUser, isContainer, uid) {
+            if (isContainer) {
+                const containerName = getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
+                }
+            } else {
```
cockpituous, line 609: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220907-112243-c768a0f4-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/bridge/cockpitdisksamples.c
```diff
@@ -148,3 +153,128 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+           gchar **contents)
+{
+  g_autoptr(GError) error = NULL;
+  if (!g_file_get_contents (path, contents, NULL, &error))
+    {
+      // do not log following erros:
+      // file does not exist: process died and can no longer be sampled
+      // access denied: bridge is not privileged and can't sample all processes
+      if (!(g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT) ||
+          g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_ACCES)))
```
tomasmatus, line 168: Looks like sometimes [G_FILE_ERROR_FAILED](https://docs.gtk.org/glib/error.FileError.html?q=#failed) is returned...

pkg/metrics/metrics.jsx
cockpituous, line 609: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220912-093730-e35fe7c5-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 431: These 9 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220912-104552-88be796a-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 437: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220912-104552-88be796a-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 609: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220912-104552-88be796a-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 431: These 9 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220913-080738-76591a4a-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 437: These 5 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220913-080738-76591a4a-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 609: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220913-080738-76591a4a-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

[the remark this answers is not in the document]
tomasmatus: Can't test this without c-podman

pkg/metrics/metrics.jsx
```diff
@@ -477,64 +573,75 @@ class CurrentMetrics extends React.Component {
...
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
+                }
+            } else {
```
cockpituous, line 608: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20220919-091511-388f64dc-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/bridge/cockpitdisksamples.c
```diff
@@ -148,3 +153,124 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+          g_warning ("Domain: %s, code: %d, message: %s", g_quark_to_string (error->domain), error->code, error->message);
+        }
+
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static void
```
allisonkarlitskaya, line 181: I'm not a fan of how you read the key (the cgroup name) from this function, but the values (the read/write figures) from the other function.  It also feels weird that you pass the allocated struct across the function boundary, expecting it to be consumed — that's highly irregular.  It also feels weird that, if you find the item already in the hash table, you modify the new item and replace the old item, instead of modifying the existing item, in-place.

I feel like it would be more natural to have some lower-level functions exclusively to deal with the hash table, something like:

```C
void
table_add_values (GHashTable *table,
                  const gchar *cgroup,
                  guint64 read,
                  guint64 write)
{
  // create or update the hash table entry for this cgroup
}
```

and then you could get the key and values from a single function and call this function.

allisonkarlitskaya (changes requested):
Iterating over every single process seems like it could get expensive, depending on the system... Is there a better way to collect these statistics from the cgroup machinery itself?

tomasmatus:
> Is there a better way to collect these statistics from the cgroup machinery itself?

There is a [cgroup io controller](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#io) but as far as I understand it is disabled by default.

pkg/metrics/metrics.jsx
```diff
@@ -486,64 +583,75 @@ class CurrentMetrics extends React.Component {
...
+                cockpit.jump("/system/services#/" + name + ".service" + (isUser ? "?owner=user" : ""));
```
cockpituous, line 619: These 9 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230130-084654-b6542dcc-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -486,64 +584,69 @@ class CurrentMetrics extends React.Component {
...
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
+                }
+            } else {
```
cockpituous, line 618: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230130-115547-84294f61-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

cockpituous, line 618: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230130-143643-3bb2e1e7-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly:
As alternative we can read `/sys/fs/cgroup/system.slice/libvirtd.service/cgroup.procs` and then only read the `/proc/$pid/io` files. Still feels a bit meh

garrett:
> It is an expensive task to perform but since it is only going to run when the metrics page is visible I believe we can afford to do this.

How "expensive" is it? Can you put it in terms of IO, CPU load, etc.? Does it cause any spikes? Is it noticeable by the user? Or is it just "more expensive than just asking cgroups for the data"?

> not enabled by default at least in Fedora

Can we petition for them to enable it in future versions? Is there anywhere we know of (like our supported distros) that have it enabled? Does it make sense to try to use cgroups when possible and then use the "expensive" fallback otherwise? (If it's not enabled anywhere, then it feels like that extra work to do it the right way is pointless for now... just future proofing otherwise.)

tomasmatus:
It's mainly this:
> Or is it just "more expensive than just asking cgroups for the data"?

and it doesn't cause any spikes even after I spawned multiple hundreds of processes that it needs to sample.

garrett:
> and it doesn't cause any spikes even after I spawned multiple hundreds of processes that it needs to sample.

Ah cool.

I was doing to suggest only updating half as often as other updates (or something like this) otherwise. So the information is still there, still updated, and still useful... just not refreshed as much.

I'm glad "expensive" is only _relatively_ expensive.

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,134 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+    {
+      // directory name is PID of a process, PID is an unsigned int starting from 1
+      // when g_ascii_strtoull returns 0 it is not a PID directory
+      if (g_ascii_strtoull (dir_name, NULL, 10))
+        {
+          get_process_io (dir_name, table);
+        }
+    }
+
+  g_hash_table_foreach (table, sample_table_values, samples);
```
allisonkarlitskaya, line 289: It would be a lot clearer if you open-coded this with an iter:

```c
GHashTableIter iter;
gpointer key, value;
g_hash_table_iter_init (&iter, sample_table_values));
while (g_hash_table_iter_next (&iter, &key, &value))
  {
    const gchar *cgroup_name = key;
    const cgroup_values_t *values = value;

    // Skip ::0/
    cgroup_name += 4;

    ....etc
  }
```

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,134 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+      g_warning ("Domain: %s, code: %d, message: %s", g_quark_to_string (error->domain), error->code, error->message);
+      return NULL;
+    }
+
+  return contents;
+}
+
+static void
+table_add_values(GHashTable *table,
+                   gchar *cgroup,
```
allisonkarlitskaya, line 190: Having this consume the string is a bit gross.  I'd rather take a copy of it in the case you add to the thing.

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,134 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+    }
+  else
+    {
+      // create new entry
+      cgroup_values_t *new = g_malloc0 (sizeof (cgroup_values_t));
+      new->read = read;
+      new->write = write;
+      g_hash_table_insert (table, cgroup, new);
+    }
+}
```
allisonkarlitskaya, line 210: I guess that more usually looks something like

```c
values = g_hash_table_lookup (table, cgroup);
if (values == NULL)
  {
    values = g_new0(cgroup_values_t, 1);
    g_hash_table_insert (table, g_strdup (cgroup), values);
  }

values->read += read;
values->write += write;
```

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,134 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+  gchar **lines = g_strsplit (contents, "\n", -1);
+  for (guint n = 0; lines != NULL && lines[n] != NULL; n++)
+    {
+      gchar description[32] = {'\0'};
+      guint value;
+      const gchar *line = lines[n];
+      if (strlen (line) == 0)
+        continue;
+
+      if (sscanf (line, "%s %d", description, &value) != 2)
```
allisonkarlitskaya, line 231: I guess `%d` isn't going to be enough here.  You store the totals in 64bit fields, so probably this needs to be a 64bit read as well.

allisonkarlitskaya (changes requested):
I still have huge reservations about the massive amount of work that this is going to involve.  I wonder if we rather do this with the cgroups approach, but only if that controller is enabled?

Otherwise, a very large amount of the code here is essentially replicating `read_keyed_int64()` which already exists in `cockpitcgroupsamples.c`.  I wonder if we could find a way to share that code rather than adding a new copy of the same thing.

tomasmatus:
> I still have huge reservations about the massive amount of work that this is going to involve. I wonder if we rather do this with the cgroups approach, but only if that controller is enabled?

With the controller often being disabled or not fully enabled in the whole cgroup structure I think it's more accurate to collect this from `/proc`.

> Otherwise, a very large amount of the code here is essentially replicating `read_keyed_int64()` which already exists in `cockpitcgroupsamples.c`. I wonder if we could find a way to share that code rather than adding a new copy of the same thing.

You have a point, actually. The way @jelly suggested and how this could fit into the `cockpitcgroupsamples.c` is reading PIDs from `cgroup.procs`, then getting relevant io statistics from `/proc/PID/io` and sample it. It's still a lot of processes to check but on the upside it avoids the complexity of using hash table and I could reuse the existing code.

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,130 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+get_process_io (const gchar *dir_name,
+                GHashTable *table)
+{
+  g_autofree gchar *path = g_strdup_printf ("/proc/%s/io", dir_name);
+  g_autofree gchar *contents = read_file (path);
+  if (!contents)
+    return;
+
+  guint64 read = 0, write = 0;
+  gchar **lines = g_strsplit (contents, "\n", -1);
```
allisonkarlitskaya, line 215: It seems you don't intend to integrate this with the cgroups stuff and use the file-reading functions there, so let's talk about the approach here.  I feel like reading the file, splitting it into lines and then iterating over the lines, only to free everything at the end is probably a little bit overkill.

Maybe try a bit of `stdio`.  Consider this sample program as potential inspiration:

```c
#include <stdio.h>  
#include <stdlib.h>    
  
int  
main (int argc, char **argv)  
{  
  FILE *fp = fopen("/proc/self/io", "r");  
    
  char *field;  
  unsigned long long value;  
  while (fscanf (fp, "%m[^: ]: %llu\n", &field, &value) == 2)   
    {  
      printf ("field %s value %llu\n", field, value); 
      free (field);   
    }

  fclose (fp);

  return 0;  
}
```

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,130 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+      return;
+    }
+
+  GHashTable *table = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+  g_autofree const gchar *dir_name = NULL;
+  while ((dir_name = g_dir_read_name (dir)) != NULL)
+    {
+      // directory name is PID of a process, PID is an unsigned int starting from 1
+      // when g_ascii_strtoull returns 0 it is not a PID directory
+      if (g_ascii_strtoull (dir_name, NULL, 10))
```
allisonkarlitskaya, line 266: as per `proc(5)`, non-pid entries in proc are guaranteed to start with a character a-z, so you could just check:

```C
  if ('0' <= dir_name[0] && dir_name[0] <= '9')
```

with a comment mentioning the above info.

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,130 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+
+  values->read = read;
+  values->write = write;
+}
+
+static void
+get_process_io (const gchar *dir_name,
+                GHashTable *table)
+{
+  g_autofree gchar *path = g_strdup_printf ("/proc/%s/io", dir_name);
```
allisonkarlitskaya, line 209: The extremely-theoretical race between reading `/io` and `/cgroup` tells me that you ought to be using dirfds, but I think we could let it slide...

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,130 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+        read = value;
+      else if (g_str_equal (description, "write_bytes:"))
+        write = value;
+    }
+
+  g_strfreev (lines);
+
+  // get process cgroup
+  g_autofree gchar *cgroup_path = g_strdup_printf ("/proc/%s/cgroup", dir_name);
+  g_autofree gchar *cgroup = read_file (cgroup_path);
```
allisonkarlitskaya, line 240: You could use an `fopen` wrapper here which implements:

 - dirfd opening + fdopen handling
 - all the error handling you want to do

and then, `FILE *` in hand, reading the cgroup is trivial.

Then you could modify this entire function to take:

1) dirfd to the process subdir
2) pointer to hash table

allisonkarlitskaya (changes requested):
In general, I try to avoid `stdio` file handling, and particularly `scanf`, and particularly particularly `scanf` `%s`, but we're dealing with the kernel here — it's not likely to try to attack us.  And even if it does, cockpit-bridge isn't in any way a security-critical component.  Using stdio here would be a significant reduction in effort.  See comments.

I like dirfd.  Maybe a bit too much.  I always use it whenever possible, and in this case it would even close an extremely theoretical race whereby the PID got recycled between the time we read the `/io` and `/cgroup` files.  I don't think that's *ever* going to happen in practice, but in theory, doing this correctly requires the use of dirfd.  You might disagree that it's worthwhile here.

If you do use `dirfd`, I'd basically expect to see an `open("/proc", O_PATH | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);` at the toplevel and then a `openat(..., O_PATH | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC)` for each numerical subdirectory, with the resulting fd passed to the `get_process_io()` function.

src/bridge/cockpitdisksamples.c
```diff
@@ -153,3 +158,130 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+                   guint64 write)
+{
+  cgroup_values_t *values = g_hash_table_lookup (table, cgroup);
+  if (values == NULL)
+    {
+      values = g_new0(cgroup_values_t, 1);
+      g_hash_table_insert (table, g_strdup (cgroup), values);
+    }
+
+  values->read = read;
```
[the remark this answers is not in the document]
tomasmatus: yes, it should be... how did I even...?

pkg/metrics/metrics.jsx
```diff
@@ -486,64 +507,146 @@ class CurrentMetrics extends React.Component {
...
```
cockpituous, line 618: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230215-105425-d1cc4359-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

allisonkarlitskaya:
As mentioned in IRC: this entire approach might not be very robust against processes changing cgroups.  In that case, the amount of cumulative IO in the old cgroup is going to have a sudden negative change, and in the new group, a sudden positive change...

pkg/metrics/metrics.jsx
```diff
@@ -486,64 +508,146 @@ class CurrentMetrics extends React.Component {
...
```
cockpituous, line 619: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230219-233351-4565a66d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -292,6 +308,12 @@ class CurrentMetrics extends React.Component {
             this.temperature_channel.addEventListener("message", this.onTemperatureUpdate);
         }
 
+        // try privileged channel so system administrator can see per-service IO usage of all users
+        if (!cockpit.hidden && (this.privileged_channel === null)) {
+            this.privileged_channel = cockpit.channel({ superuser: "try", payload: "metrics1", source: "internal", interval: 3000, metrics: PRIVILEGED_METRICS });
```
tomasmatus, line 313: @martinpitt Since I need privileged access to be able to sample every process in /proc/<pid> should I open new channel with `"superuser": "try"` or should I reuse the existing metrics channel and add superuser to it?

I assume limiting what runs with escalated permissions is the safer call here and opening one extra channel won't add too much overhead.

martinpitt: Is this question related to this particular `metrics1` channel? I suppose not, as you want to read files from /proc directly, just making sure we talk about the same thing.

Please do the same thing for reading /proc, with superuser=try. Then admins can scan all processes, and non-admins can at least scan their own. But please don't open hundreds of channels to read all /proc files directly, run a helper shell or python script that scrapes all the info and outputs a JSON stream (so that it keeps running).

pkg/metrics/metrics.jsx
cockpituous, line 619: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230220-094243-ab5b7938-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
tomasmatus: Yes, it's about this one channel. c-bridge is who reads from /proc and then sends data through this channel as a JSON stream.

pkg/metrics/metrics.jsx
```diff
@@ -486,64 +509,146 @@ class CurrentMetrics extends React.Component {
...
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
+                }
+            } else {
```
cockpituous, line 620: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230220-112942-982465b3-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

cockpituous, line 620: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230220-131652-98c00acd-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/bridge/cockpitdisksamples.c
```diff
@@ -159,37 +161,34 @@ cockpit_disk_samples (CockpitSamples *samples)
...
+        g_message ("error opening file descriptor: %s: %m", name);
+      return NULL;
+    }
 
-      g_warning ("Domain: %s, code: %d, message: %s", g_quark_to_string (error->domain), error->code, error->message);
+  FILE *fp = fdopen (fd, "r");
+  if (!fp)
+    {
+      if (errno != EACCES)
+        g_message ("error opening file: %s: %m", name);
```
allisonkarlitskaya: I think this can't fail except with `ENOMEM`.  You'd get the permission error only from the `openat()` call.

src/bridge/cockpitdisksamples.c
```diff
@@ -159,37 +161,34 @@ cockpit_disk_samples (CockpitSamples *samples)
   g_free (contents);
 }
 
-static gchar*
-read_file (const gchar *path)
+static FILE*
```
allisonkarlitskaya, line 164: Space here: `FILE *`.

src/bridge/cockpitdisksamples.c
allisonkarlitskaya: Honestly, if this does fail, I'd `g_error()` to abort the program.  If you don't want to do that, then you need to `close (fd);`.

src/bridge/cockpitdisksamples.c
```diff
@@ -203,55 +202,49 @@ table_add_values(GHashTable *table,
...
-  g_autofree gchar *cgroup_path = g_strdup_printf ("/proc/%s/cgroup", dir_name);
-  g_autofree gchar *cgroup = read_file (cgroup_path);
-  if (!cgroup)
+  FILE *cgroup_fp = open_file (dirfd, "cgroup");
+  if (!cgroup_fp)
     return;
-  g_strchomp (cgroup);
 
-  table_add_values (table, cgroup, disk_read, disk_write);
+  g_autofree gchar *cgroup = NULL;
```
allisonkarlitskaya, line 231: Call me oldschool, but mixing autofree with stdio rubs me the wrong way.  I think these days it's officially safe, but it seems weird...

src/bridge/cockpitdisksamples.c
```diff
@@ -203,55 +202,49 @@ table_add_values(GHashTable *table,
...
+    table_add_values (table, cgroup + 4, disk_read, disk_write);
+
+  fclose (cgroup_fp);
 }
 
 void
 cockpit_cgroup_disk_usage (CockpitSamples *samples)
 {
   g_autoptr(GError) error = NULL;
   g_autoptr(GDir) dir = g_dir_open ("/proc", 0, &error);
```
allisonkarlitskaya, line 244: I think I'd just trade out `GDir` for `opendir` and `readdir`, since then you get to use `dirfd()` and don't need to open the directory twice.

It also lets you check for the items being directories (although I think it's pretty pointless if we're dealing with procfs -- we already know all numeric entries are directories).

src/bridge/cockpitdisksamples.c
```diff
@@ -264,9 +257,17 @@ cockpit_cgroup_disk_usage (CockpitSamples *samples)
       // non-pid entries in proc are guaranteed to start with a character a-z
       if ('0' <= dir_name[0] && dir_name[0] <= '9')
         {
-          get_process_io (dir_name, table);
+          int dfd = openat(proc_fd, dir_name, O_PATH | O_DIRECTORY);
+          if (dfd < 0)
+          {
+            g_message ("failed to open /proc/%s", dir_name);
```
allisonkarlitskaya: A `%m` might be nice there.

allisonkarlitskaya (dismissed):
This is starting to look really cool.  Needs squashing, obviously, but only very minor tweaks left now...

pkg/metrics/metrics.jsx
cockpituous, line 620: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230221-100738-475c62d2-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/bridge/cockpitdisksamples.c
tomasmatus: Haha, okay. Never expected to hear I shouldn't use autofree :)

src/bridge/cockpitdisksamples.c
tomasmatus: Done, and I think it looks nicer

pkg/metrics/metrics.jsx
cockpituous, line 620: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230221-192144-3bd1e918-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

cockpituous, line 620: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230222-002235-1aef9c77-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

cockpituous, line 620: These 8 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230222-084226-1aef9c77-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -485,64 +508,146 @@ class CurrentMetrics extends React.Component {
...
+    cgroupRow = (name, is_user, is_container, uid, ...values) => {
+        const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
```
cockpituous, line 617: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230309-092834-eabc9108-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -485,64 +508,146 @@
...
+                }
+            } else {
```
cockpituous, line 619: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230309-092834-eabc9108-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

allisonkarlitskaya (commented):
For the record: I don't have any further comments, but am waiting to review until you push a version that fixes the tests.

pkg/metrics/metrics.jsx
```diff
@@ -487,64 +510,146 @@
...
+    cgroupRow = (name, is_user, is_container, uid, ...values) => {
+        const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
+                } else {
+                    cockpit.jump("/podman");
```
cockpituous, line 619: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230310-085943-18a5fb70-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+
+        # Skip leadin ::0/
+        return cgroup_name[4:]
+
+    @staticmethod
+    def get_proc_io(fd: int) -> Optional[List[int]]:
+        data = ""
+        try:
+            with Handle.open('io', os.O_RDONLY, dir_fd=fd) as io_fd:
+                data = os.read(io_fd, 1024).decode()
```
allisonkarlitskaya, line 327: For cgroups 1024 made me think "...hopefully this is enough?".

For the "io" file I feel like it's definitely not enough.  This is a very small file, but the format suggests that it might be extended at some point.  Although probably at that point the items we're looking for will still be at the front....

I feel like there's little benefit here to being so conservative with this number.

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
 
 
+class CGroupDiskIO(Sampler):
+    descriptions = [
+        SampleDescription('disk.cgroup.read', 'bytes', 'counter', True),
+        SampleDescription('disk.cgroup.written', 'bytes', 'counter', True),
+    ]
+
+    @staticmethod
+    def get_cgroup_name(fd: int) -> Optional[str]:
```
allisonkarlitskaya, line 311: Make this rather return `str`...

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+        cgroup_name = ""
+        try:
+            with Handle.open('cgroup', os.O_RDONLY, dir_fd=fd) as cgroup_fd:
+                cgroup_name = os.read(cgroup_fd, 1024).decode().strip()
+        except FileNotFoundError:
```
allisonkarlitskaya, line 316: ...and stop catching exceptions.  They can bubble out in case the file isn't there.  We already catch exceptions in the toplevel function and do the right thing in that case (skip to the next iteration).

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
 
+class CGroupDiskIO(Sampler):
+    descriptions = [
+        SampleDescription('disk.cgroup.read', 'bytes', 'counter', True),
+        SampleDescription('disk.cgroup.written', 'bytes', 'counter', True),
+    ]
+
+    @staticmethod
+    def get_cgroup_name(fd: int) -> Optional[str]:
+        cgroup_name = ""
```
allisonkarlitskaya, line 312: ...drop this initializer...

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+            with Handle.open('cgroup', os.O_RDONLY, dir_fd=fd) as cgroup_fd:
+                cgroup_name = os.read(cgroup_fd, 1024).decode().strip()
+        except FileNotFoundError:
+            return None
+
+        # Skip leadin ::0/
+        return cgroup_name[4:]
+
+    @staticmethod
+    def get_proc_io(fd: int) -> Optional[List[int]]:
```
allisonkarlitskaya, line 323: You should try to make this function return `Tuple[int, int]`.  Remove the `Optional` as per above.

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+        data = ""
```
allisonkarlitskaya, line 324: ...and the initializer

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+        try:
+            with Handle.open('io', os.O_RDONLY, dir_fd=fd) as io_fd:
+                data = os.read(io_fd, 1024).decode()
+        except (FileNotFoundError, PermissionError):
```
allisonkarlitskaya, line 328: ...and the except (but in this case, you'll need to add PermissionError to the main function's except list)

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+            return None
+
+        values = []
+        for key in ['read_bytes:', 'write_bytes:']:
+            start = data.index(key) + len(key)
+            end = data.index('\n', start)
+            values.append(int(data[start:end]))
```
allisonkarlitskaya, line 335: Although I think the general strategy here is fairly OK (albeit possibly buggy in presence of keys with names like `other_read_bytes:` or something) I really don't like the list.  I wonder if maybe you could do this more reliably with two regexps instead of all of this slicing and dicing.  Maybe even one, if you use named groups.

Something like this works, for example:

```python
>>> int(re.match(rb'\bread: (\d+)', b'read: 5').group(1))
5
```

...and you can also save the `.decode()` by working on bytes.

I'm not sure that this is the best of all possible solutions, but I definitely dislike it a lot less than the list.  Maybe you can think of something better.

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+        with Handle.open('/proc', os.O_RDONLY | os.O_DIRECTORY) as proc_fd:
+            for path in os.listdir(proc_fd):
+                # non-pid entries in proc are guaranteed to start with a character a-z
+                if path[0] < '0' or path[0] > '9':
+                    continue
+
+                try:
+                    with Handle.open(path, os.O_RDONLY | os.O_DIRECTORY, dir_fd=proc_fd) as pid_fd:
+                        cgroup_name = self.get_cgroup_name(pid_fd)
+                        proc_io = self.get_proc_io(pid_fd)
```
allisonkarlitskaya, line 349:
```suggestion
                        read, written = self.get_proc_io(pid_fd)
```

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+
+                        if proc_io is not None:
+                            samples['disk.cgroup.read'][cgroup_name] = samples['disk.cgroup.read'].get(cgroup_name, 0) + proc_io[0]
```
allisonkarlitskaya, line 352: I think the best option here is to capture `samples['disk.cgroup.read']` and `...written...` at the top of the loop and use them here to avoid having to say `samples['disk.cgroup.read']` every time, twice.

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+                            samples['disk.cgroup.written'][cgroup_name] = samples['disk.cgroup.written'].get(cgroup_name, 0) + proc_io[1]
+                except FileNotFoundError:
```
allisonkarlitskaya, line 354: Probably pull this up a bit.  This should cover only the part that's actually doing IO.  Take the dictionary stuff out of the context manager and `try:` clause.

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+                # non-pid entries in proc are guaranteed to start with a character a-z
+                if path[0] < '0' or path[0] > '9':
+                    continue
+
+                try:
+                    with Handle.open(path, os.O_RDONLY | os.O_DIRECTORY, dir_fd=proc_fd) as pid_fd:
+                        cgroup_name = self.get_cgroup_name(pid_fd)
+                        proc_io = self.get_proc_io(pid_fd)
+
+                        if proc_io is not None:
```
allisonkarlitskaya, line 351: You never check `cgroup_name` for being `None`, but both checks can be avoided if you allow the exceptions to escape the functions to be handled by this `try:` clause.

src/cockpit/samples.py
```diff
@@ -301,6 +301,60 @@ def sample(self, samples: Samples) -> None:
...
+
+    def sample(self, samples: Samples):
+        with Handle.open('/proc', os.O_RDONLY | os.O_DIRECTORY) as proc_fd:
+            for path in os.listdir(proc_fd):
+                # non-pid entries in proc are guaranteed to start with a character a-z
+                if path[0] < '0' or path[0] > '9':
+                    continue
+
+                try:
+                    with Handle.open(path, os.O_RDONLY | os.O_DIRECTORY, dir_fd=proc_fd) as pid_fd:
```
allisonkarlitskaya, line 347: This could probably be `O_PATH`

allisonkarlitskaya (changes requested):
I really like what I see!  The approach here is quite nice and I don't think any restructuring is needed — only local changes.  See suggestions below.

src/cockpit/samples.py
allisonkarlitskaya: ...it's pretty frustrating that stock `open()` doesn't take `dir_fd`.  Grr.

pkg/metrics/metrics.jsx
```diff
@@ -487,64 +510,146 @@ class CurrentMetrics extends React.Component {
...
+    getCachedPodName = (uid, containerid) => this.state.podNameMapping[uid] && this.state.podNameMapping[uid][containerid];
+
+    cgroupRow = (name, is_user, is_container, uid, ...values) => {
+        const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
```
cockpituous, line 617: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230313-094330-e8bd3b58-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 619: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230313-094330-e8bd3b58-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 617: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230314-094406-227d8926-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 619: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230314-094406-227d8926-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/cockpit/samples.py
```diff
@@ -301,6 +302,55 @@ def sample(self, samples: Samples) -> None:
...
+                    continue
+
+                try:
+                    with Handle.open(path, os.O_PATH, dir_fd=proc_fd) as pid_fd:
+                        cgroup_name = self.get_cgroup_name(pid_fd)
+                        proc_read, proc_write = self.get_proc_io(pid_fd)
+                except (FileNotFoundError, PermissionError):
+                    continue
+
+                reads[cgroup_name] = reads.get(cgroup_name, 0) + proc_read
```
allisonkarlitskaya, line 350: :+1:

src/cockpit/samples.py
```diff
@@ -301,6 +302,55 @@ def sample(self, samples: Samples) -> None:
...
+
+            # Skip leadin ::0/
+            return cgroup_name[4:]
+
+    @staticmethod
+    def get_proc_io(fd: int) -> Tuple[int, int]:
+        with Handle.open('io', os.O_RDONLY, dir_fd=fd) as io_fd:
+            data = os.read(io_fd, 4096)
+
+            match = re.search(rb'\bread_bytes: (?P<read>\d+)\nwrite_bytes: (?P<write>\d+)', data)
```
allisonkarlitskaya, line 324: Is it guaranteed that the read bytes will come before the write bytes?

Can you please pre-compile the RE in static scope?  We do this once a second for every single process on the system.

allisonkarlitskaya (dismissed):
I reviewed the Python parts and they're great.  Also: no comments about the C parts since they've been stable a while already.

See below for a suggested improvement and a question.  Feel free to deal with those in follow-ups.

pkg/metrics/metrics.jsx
cockpituous, line 617: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230315-110937-e91d2403-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 619: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230315-110937-e91d2403-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 617: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230316-090626-49f50ac0-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 619: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230316-090626-49f50ac0-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -489,64 +512,146 @@
...
+                } else {
+                    cockpit.jump("/podman");
```
cockpituous, line 621: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230317-132211-67e5d819-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -489,64 +512,146 @@ class CurrentMetrics extends React.Component {
...
+                // filter out podman containers, but only for the logged in
+                // user or root user if the user is privileged. Other users
+                // containers will show up under the user@uid cgroup
+                const matches = name.match(/libpod-(?<containerid>[a-z|0-9]{64})\.scope/);
+                if (matches && valuesA[i] !== undefined && valuesB[i] !== undefined) {
+                    let isUser = false;
+                    let uid = 0;
+                    const containerid = matches.groups.containerid;
+                    const umatches = name.match(useridCgroupRe);
+                    if (umatches) {
```
KKoukiou, line 567: This could be nicer riten like:
const isUser = umatches;
const uid = parseInt(umtches?.groups.userid) || 0;

KKoukiou (commented):
Is this blocked on something?

KKoukiou (changes requested):
When not consuming disk IO services should not be displayed on the list.

Screenshot from 2023-04-27 14-12-39

pkg/metrics/metrics.jsx
```diff
@@ -507,62 +530,140 @@ class CurrentMetrics extends React.Component {
...
+    getCachedPodName = (uid, containerid) => this.state.podNameMapping[uid] && this.state.podNameMapping[uid][containerid];
+
+    cgroupRow = (name, is_user, is_container, uid, ...values) => {
+        const podman_installed = cockpit.manifests && cockpit.manifests.podman;
+
+        const cgroupClickHandler = (name, isUser, isContainer, uid) => {
+            if (isContainer) {
+                const containerName = this.getCachedPodName(uid, name);
+                if (containerName) {
+                    cockpit.jump("/podman#/?name=" + containerName);
```
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230427-140842-f3cf22b1-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -507,62 +530,140 @@
...
+                } else {
+                    cockpit.jump("/podman");
```
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230427-140842-f3cf22b1-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230428-144554-199a5aac-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230428-144554-199a5aac-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-072403-a2b59b9b-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-072403-a2b59b9b-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-130819-e103b751-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-130819-e103b751-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-141142-9d208265-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-141142-9d208265-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

KKoukiou (dismissed):

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-155255-42a47b6e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230502-155255-42a47b6e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230503-115756-df104754-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

KKoukiou (dismissed):

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230503-134737-62c9b9bd-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230503-134737-62c9b9bd-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt (dismissed):
Thanks! (Assuming green, as usual)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230503-162135-79159ad2-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230503-162135-79159ad2-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
Test [fails on rhel-8-9](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230503-162135-79159ad2-rhel-8-9/log.html#337)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-071225-f305d649-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-071225-f305d649-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-083157-f305d649-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-083157-f305d649-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

KKoukiou:
Metrics page pixel testing became very unstable with this PR https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-101125-f305d649-fedora-37/log.html

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-144128-83b0249d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
cockpituous, line 635: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-144128-83b0249d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

KKoukiou:
@martinpitt are you fine with the removal of the pixel tests? I can try to re-add them - but i don't want to block this anymore.

pkg/metrics/metrics.jsx
cockpituous, line 633: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17692-20230504-164506-83b0249d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt (approved):
Yes, I am -- the metrics cards are so laden with unpredictable info that IMHO this is better tested textually. Thanks!

PR #17694, opened 2022-08-31
Small bridge fixes

jelly:


allisonkarlitskaya (changes requested):
Thanks for splitting those out.

The patches are both fine, but both commit messages are too short.  You should mention at the very least that the dbus channel is involved, and write a sentence about the change, like "If foo fails to quux, then we bazz instead, but then we need to bar instead of zomg".  You also misspelled send_message as sends_message in the summary.

allisonkarlitskaya (approved):
thanks 🙂👍

PR #17695, opened 2022-09-01
python: send EOF on AsyncChannel close

allisonkarlitskaya:
If an AsyncChannel gets `close` from the other side without having first
received `done` then the EOF will never get delivered to the reader in
the `run()` function.

Add an extra EOF on `close`, as well.


jelly (approved):

jelly:
Works for, the metrics channel.

PR #17698, opened 2022-09-01
test: Drop obsolete mkdir -p calls

martinpitt:
Since https://github.com/cockpit-project/bots/commit/e84a589781a3fa ,
SSHConnection.write() creates the target directory by itself.

---

Follow-ups from https://github.com/cockpit-project/cockpit/pull/17645#pullrequestreview-1092152421

### DIFF
--- a/test/verify/check-pages
+++ b/test/verify/check-pages
@@ -97,9 +97,11 @@ OnCalendar=daily
 
         # On Debian and Ubuntu we have to generate the other locales
         if "debian" in m.image:
-            m.execute("echo 'de_DE.UTF-8 UTF-8' >> /etc/locale.gen; locale-gen; update-locale")
+            m.write("/etc/locale.gen", "de_DE.UTF-8 UTF-8\n", append=True)
+            m.execute("locale-gen; update-locale")
         elif "arch" == m.image:
-            m.execute("echo 'de_DE.UTF-8 UTF-8' >> /etc/locale.gen; locale-gen")
+            m.write("/etc/locale.gen", "de_DE.UTF-8 UTF-8\n", append=True)
+            m.execute("locale-gen")
         elif "ubuntu" in m.image:
             m.execute("locale-gen de_DE; locale-gen de_DE.UTF-8; update-locale")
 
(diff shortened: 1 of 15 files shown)


test/verify/check-pages
```diff
@@ -563,7 +564,8 @@ OnCalendar=daily
 
         stuff = os.path.join(self.vm_tmpdir, "stuff")
         # prepare a directory for testing file autocomplete widget
-        m.execute("mkdir -p {0}/dir1; touch {0}/file1.txt".format(stuff))
+        m.execute(f"mkdir -p {stuff}/dir1")
```
allisonkarlitskaya, line 567: Confusing, but correct.

Could also be re-written as `mkdir` (without `-p`) *after* the `write()`.

test/verify/check-shell-multi-machine-key
```diff
@@ -239,15 +239,14 @@ class TestMultiMachineKeyAuth(MachineCase):
...
-        m1.execute("mkdir -p /home/admin/.ssh")
-        m1.upload(["verify/files/ssh/id_ed25519", "verify/files/ssh/id_ed25519.pub",
-                   "verify/files/ssh/id_rsa", "verify/files/ssh/id_rsa.pub"],
-                  "/home/admin/.ssh/")
         m1.write("/home/admin/.ssh/config", """
 Host 10.111.113.2
     User user
     IdentityFile /home/admin/.ssh/id_ed25519
 """)
+        m1.upload(["verify/files/ssh/id_ed25519", "verify/files/ssh/id_ed25519.pub",
```
allisonkarlitskaya, line 247: This reordering is begging for `mkdir -p` to be added to `Machine.upload()` as well.

allisonkarlitskaya (approved):
Good cleanup, thanks.  Ready to go, as-is.

A nice follow-up would be the dedenting of the contents, as we were discussing.

test/verify/check-shell-multi-machine-key
martinpitt: Yes, I thought the same. It's on my TODO list now, next to the dedent() thing.

test/verify/check-pages
martinpitt: It was indeed confusing before, as I initially mis-read it as "make the dir for the touched file", but that's not what it did. It's at least less confusing now.

The `-p` is indeed unnecessary, but TBH I don't feel like running this again just for this.

PR #17699, opened 2022-09-01
lib: Allow setting initial password for PasswordFormFields

skobyda:
This allows a user of PasswordFormFields to set the password value
from the outside, instead of the value being just initialized and stored
internally inside the component.

The motivation behind this change can be found at https://github.com/cockpit-project/cockpit-machines/pull/803 , which is blocked on this PR.


martinpitt (approved):
Thanks!

PR #17702, opened 2022-09-02
Various http stream test improvements, doc update

martinpitt:
Splitting this out from PR #17631 as the /mock/info stuff might be useful for @jelly and @mvollmer to adjust their parts of the unit test for py vs. C bridge differences.

### DIFF
--- a/pkg/base1/test-echo.js
+++ b/pkg/base1/test-echo.js
@@ -79,10 +79,18 @@ QUnit.test("binary", function (assert) {
     channel.send(buffer);
 });
 
-// This is implemented in the C bridge, but not in Python.
-// Nobody uses it, so skip the test for now.
-QUnit.test.skip("fence", function (assert) {
+QUnit.test("fence", async assert => {
     const done = assert.async();
+
+    const response = await fetch(`http://${window.location.hostname}:${window.location.port}/mock/info`);
+    const info = await response.json();
+    // This is implemented in the C bridge, but not in Python.
+    if (info.bridge == "cockpit-bridge.pyz") {
+        assert.ok(true, "skipping on python bridge, not implemented");
+        done();
+        return;
+    }
+
     assert.expect(2);
 
     const before = cockpit.channel({ payload: "echo" });
--- a/doc/protocol.md
+++ b/doc/protocol.md
@@ -717,12 +717,12 @@ payload type:
    host name. To use this option you must also specify a port.
    This option should be used to communicate with external
    apis and not as a synonym for the host parameter.
+ * "method": "GET", "POST", or other valid HTTP method
 
 You may also specify these options:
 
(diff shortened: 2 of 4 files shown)


pkg/base1/test-echo.js
```diff
@@ -79,41 +79,55 @@ QUnit.test("binary", function (assert) {
...
-    }
-
-    /* We send messages in this order, but they should echoed in numeric order */
-    before.send("1");
-    after.send("4");
-    after.send("5");
-    fence.send("2");
-    fence.send("3");
+    // This is implemented in the C bridge, but not in Python.
+    cockpit.http(test_server).get("/mock/info")
```
allisonkarlitskaya, line 91: Less chaining and arrows.  More `await`/`async`, please.

allisonkarlitskaya (changes requested):
Looks good except for one major issue.  I'm gonna start getting more and more annoying on this point, going forward... ;)

martinpitt:
Oops, this also needs work -- `make pycheck` hangs. cockpit.http() on /mock/info can't work on the pybridge yet in this PR, so this is bogus. I'll drop this and re-put everything into #17631

pkg/base1/test-echo.js
martinpitt: Done.

pkg/base1/test-echo.js
```diff
@@ -79,10 +79,18 @@ QUnit.test("binary", function (assert) {
     channel.send(buffer);
 });
 
-// This is implemented in the C bridge, but not in Python.
-// Nobody uses it, so skip the test for now.
-QUnit.test.skip("fence", function (assert) {
+QUnit.test("fence", async assert => {
     const done = assert.async();
+
+    const response = await fetch(`http://${window.location.hostname}:${window.location.port}/mock/info`);
```
allisonkarlitskaya, line 85: Could we integrate that into `pkg/lib/qunit-tests.js` somehow?

allisonkarlitskaya (approved):
Looks good, but please see suggestion: if we want to use this more often then the current approach is a bit too anti-DRY for my tastes.

pkg/base1/test-echo.js
martinpitt: Indeed in the http channel PR I [have a helper function](https://github.com/martinpitt/cockpit/commit/85f54d839bd22b36de7bfc4d725494a22401603e#diff-8f24a1fc445b04e36587c1207a30cec18e7973665db3543ef9e8854530758e95R18) already. Moving to qunit-tests.js sounds good, I'll experiment with that there, do a rebase, and centralize it.

martinpitt: Done in https://github.com/cockpit-project/cockpit/pull/17631/commits/ab40c130da35a517076e0adefafcccf607c158e0

PR #17703, opened 2022-09-02
ci(lint): Add Shell linter - Differential Shellcheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get reports only about newly added defects. 

Since this repository has few shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,26 @@
+# doc: https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme
+---
+
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [main, rhel-*]
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          severity: warning
+          token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,26 @@
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches: [main]
```
martinpitt, line 4: This would equally apply to stable RHEL branches, so this should be dropped.

.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+
+permissions:
+  contents: read
```
martinpitt, line 7: There is already a permissions: on the `jobs/lint` level, so this one seems redundant and unnecessary? I also think `contents` only applies to private repos, but all of cockpit is public. So it seems this could just be dropped?

.github/workflows/differential-shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+      pull-requests: write
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
```
martinpitt, line 24: For other third-party code that gets a token we usually pin down an exact SHA instead of a (fungible) tag. However, token here does not have privileges which would allow secret infiltration, in the worst case it can cause spamming. So I think it's okay here. @allisonkarlitskaya do you agree?

martinpitt (changes requested):
Thanks @jamacku, this is very interesting! I have some minor questions/remarks before landing.

.github/workflows/differential-shellcheck.yml
jamacku: I have updated the branches array to: `branches: [main, rhel-*]`. It should also cover RHEL branches. I think it's better to try to specify branches rather than run a check on all branches and possibly consume sources.

.github/workflows/differential-shellcheck.yml
jamacku: Pinning to SHA is good practice, but it is usually complemented by some tool (e.g. Dependabot) that looks for updates and notifies you when a new version of GitHub Action has been released.

.github/workflows/differential-shellcheck.yml
jamacku: > There is already a permissions: on the `jobs/lint` level, so this one seems redundant and unnecessary?

I usually set permission `contents: read` on top level, in case there would be added another job. But I can merge it with job-level permissions if you would like to.

> I also think `contents` only applies to private repos, but all of cockpit is public. So it seems this could just be dropped?

I don't have much experience with private repos, but I think this also applies to public repos. From [GitHub docs](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#assigning-permissions-to-a-specific-job):

* > If you specify the access for any of these scopes, all of those that are not specified are set to none.

So setting permissions `contents: read` indirectly sets everything else to `none`. And then, in each job, you specify precisely which permissions the job requires.

jamacku: @martinpitt Would you prefer the removal of `permissions`?

.github/workflows/differential-shellcheck.yml
martinpitt: ack, thank you! That's a good compromise.

.github/workflows/differential-shellcheck.yml
martinpitt: Yes, please. It's not intuitive what happens when there's two `permission:`s. Does the more specific one replace or augment the global one?

.github/workflows/differential-shellcheck.yml
martinpitt: OK, let's keep it at `@v3` for now. Please also revert the SHA pinning of `actions/checkout@v2` -- that one is from GitHub itself. If they want to 0wn us, they have more discreet ways to do that :grin:

martinpitt:
Sorry @jamacku for the terribly long delay, this slipped through the cracks. Unfortunately this never ran on https://github.com/jamacku/cockpit/actions/workflows/differential-shellcheck.yml , so one can't see the results of that before merging? I wouldn't want to merge this if it would immediately be red.

What would be useful is to add an extra `workflow_dispatch:` trigger to the workflow, so that it can be run manually without a PR; at least temporarily (but it really can't hurt).

I added the proposed changes above [as a commit on my fork](https://github.com/martinpitt/cockpit/commit/0c40a59276e4ee7e15d09baba7fb33224b7e1f96), and [did an initial run](https://github.com/martinpitt/cockpit/actions/runs/3157949323). That indeed didn't find anything. I then [added a commit which breaks a shell script](https://github.com/martinpitt/cockpit/commit/252c3035be6e04b768f0aff08527b06269d2d0b4), and [ran shellcheck again](https://github.com/martinpitt/cockpit/actions/runs/3157964961) -- but to my surprise it is [still happy](https://github.com/martinpitt/cockpit/actions/runs/3157964961) and did not spot the error at all. Shouldn't it?

This branch also needs a rebase, to fix the unit and flatpak tests.

.github/workflows/differential-shellcheck.yml
jamacku: I have reduced permissions to just:

```yml
permissions:
  contents: read
```

It should be sufficient enough.

jamacku:
> Unfortunately this never ran on https://github.com/jamacku/cockpit/actions/workflows/differential-shellcheck.yml , so one can't see the results of that before merging?

Differential ShellCheck workflow never ran because I didn't create any PR so far. To demonstrate functionality, I have submitted testing PR -  https://github.com/jamacku/cockpit/pull/1, and the run is available at - https://github.com/jamacku/cockpit/actions/runs/3172839723

>  I wouldn't want to merge this if it would immediately be red.

Differential ShellCheck reports only issues introduced via PR. When PR doesn't contain any changes to shell scripts, `differential-shellcheck` will be green. If you would like to check how it behaves in production, I would suggest you check `systemd` repo:

- https://github.com/systemd/systemd/actions/workflows/differential-shellcheck.yml

> What would be useful is to add an extra `workflow_dispatch:` trigger to the workflow, so that it can be run manually without a PR; at least temporarily (but it really can't hurt).
>
> I added the proposed changes above [as a commit on my fork](https://github.com/martinpitt/cockpit/commit/0c40a59276e4ee7e15d09baba7fb33224b7e1f96), and [did an initial run](https://github.com/martinpitt/cockpit/actions/runs/3157949323). That indeed didn't find anything. I then [added a commit which breaks a shell script](https://github.com/martinpitt/cockpit/commit/252c3035be6e04b768f0aff08527b06269d2d0b4), and [ran shellcheck again](https://github.com/martinpitt/cockpit/actions/runs/3157964961) -- but to my surprise it is [still happy](https://github.com/martinpitt/cockpit/actions/runs/3157964961) and did not spot the error at all. Shouldn't it?

I think `differential-shellcheck` behaves correctly since you have pushed breaking changes directly without going through PR. There is opened RFE for running Differential ShellCheck also on `push` events, but it would require more work since we are currently using information from PR about changed files. 

- https://github.com/redhat-plumbers-in-action/differential-shellcheck/issues/62

.github/workflows/differential-shellcheck.yml
martinpitt: I am actually surprised that this is actually sufficient -- I would have expected something like `issues: write` to be able to post comments to the PR. But as it worked in [your PR](https://github.com/jamacku/cockpit/pull/1), that' seems to be fine.

martinpitt (approved):
Thank you! Let's see this in action then. @marusak , @allisonkarlitskaya please be aware of that new workflow.

jamacku:
Thank you for merging. Feel free to ping me in case of any questions or issues.

PR #17707, opened 2022-09-02
webpack: Fix regexp for unit test files

### LINKED ISSUE #16829: webpack build is broken, depending on where you build
1. Make a checkout of cockpit into a directory with a name like "test-more-static-code".
2. Try to build it, and watch it fail.

```
make[1]: Entering directory '/var/home/lis/src/cockpit/test-more-static-code'
  WEBPACK  apps
assets by path apps/*.js 63.4 KiB 21 assets
asset apps/default.png 2.48 KiB [compared for emit] [from: pkg/apps/default.png] [copied]
asset apps/index.html 1.21 KiB [compared for emit] [from: pkg/apps/index.html] [copied]
asset apps/manifest.json 600 bytes [compared for emit] [from: pkg/apps/manifest.json] [copied]
./pkg/apps/index.jsx 2.7 KiB [built] [code generated] [1 error]

ERROR in ./pkg/apps/index.jsx 52:15
Module parse failed: Unexpected token (52:15)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|     if (!metainfo_db.ready)
>         return <EmptyStatePanel loading />;
| 
|     if (path.length === 0) {

webpack 5.65.0 compiled with 1 error in 1476 ms
make[1]: *** [Makefile:9875: dist/apps/manifest.json] Error 1
make[1]: Leaving directory '/var/home/lis/src/cockpit/test-more-static-code'
make: *** [Makefile:3088: all] Error 2
```

That comes down to a couple of places in the code which check for filenames containing `test-` and behave differently.  There's one of those in `webpack-make`:

```
        if (output.includes("/test-")) {
            if (output.endsWith(".html")) {   
                tests.add(output);
            }
            continue;   
        }
```

but the one that causes the most immediate breakage is this one from `webpack.config.js`:

```
            {
                test: /\.(js|jsx)$/,
                // exclude external dependencies; it's too slow, and they are already plain JS except the above
                // also exclude unit tests, we don't need it for them, just a waste and makes failures harder to read
                exclude: /\/node_modules|\/test-/,
                use: "babel-loader"   
            },
```

We need to figure out a way to tighten these checks up a bit.

I can fix the `webpack-make` stuff easily enough on my own, but it would probably be faster/better for someone like @KKoukiou or @marusak to look at the webpack config.

martinpitt:
`exclude:` matches on the complete absolute path. This broke when
cockpit was checked out into a directory (e.g. a branch worktree) that
started with "test-".

To avoid that, tighten the regular expression to only match on test-*.js
files.

Fixes #16829

### DIFF
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -125,17 +125,17 @@ function generateDeps(makefile, stats) {
         const output = path.join(dir, asset);
         fs.utimesSync(output, now, now);
 
-        if (!output.endsWith("/manifest.json") && !output.endsWith(".map"))
+        if (!asset.endsWith("/manifest.json") && !asset.endsWith(".map"))
             outputs.add(output);
 
-        if (output.includes("/test-")) {
-            if (output.endsWith(".html")) {
+        if (asset.includes("/test-")) {
+            if (asset.endsWith(".html")) {
                 tests.add(output);
             }
             continue;
         }
 
-        if (output.endsWith(".map") || output.endsWith(".LICENSE.txt"))
+        if (asset.endsWith(".map") || asset.endsWith(".LICENSE.txt"))
             continue;
 
         installs.add(output);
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):
Thanks!  👍

PR #17711, opened 2022-09-05
unit-tests: explicitly create the home directory

allisonkarlitskaya:
`adduser` recently changed its behaviour for system users, not to create
the home directory.  We expect the home directory to be created, though
(since we build in it), so explicitly specify `--create-home`.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -39,5 +39,6 @@ jobs:
               --verbose \
               ${{ matrix.startarg.network }} \
               --env=CC='${{ matrix.startarg.cc }}' \
+              --env=CFLAGS='-O2 -gdwarf-4' \
               --image-tag='${{ matrix.startarg.tag }}' \
               --make '${{ matrix.startarg.make }}'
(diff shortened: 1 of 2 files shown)


jelly:
FYI: comes from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=679746

jelly (dismissed):

allisonkarlitskaya:
The failing unit tests are because of clang, not i386 (I tested gcc/i386 and clang/amd64 locally).

That points to https://bugs.kde.org/show_bug.cgi?id=452758  (thanks @jelly).

.github/workflows/unit-tests.yml
```diff
@@ -6,7 +6,9 @@ jobs:
     strategy:
       matrix:
         startarg:
-          - { make: 'check-memory', cc: 'clang', tag: 'i386' }
+          # new Clang breaks Valgrind, so use gcc for all `check-memory`
+          # scenarios for now: https://bugs.kde.org/show_bug.cgi?id=452758
```
jelly, line 10: Alternatively:
```
            # HACK: valgrind does not work with DWARF5 symbols generated by Clang 14 https://bugs.kde.org/show_bug.cgi?id=452758
            command = ['sh', '-c', 'CFLAGS="-O2 -gdwarf-4" /source/autogen.sh; exec make -j$(nproc) ' + shlex.quote(args.make)]
```

allisonkarlitskaya: ...thus also avoiding the test name changing.  Sounds good :)

martinpitt:
Hmmm.. the tests succeeded here, but they still fail in the refresh run: https://github.com/cockpit-project/cockpit/runs/8189281079?check_suite_focus=true

PR #17712, opened 2022-09-05
test: Don't let our dnsmasq instances read /etc/resolv.conf

mvollmer:
They are not supposed to resolve DNS queries and thus don't need any
upstream servers.  Also, NetworkManager puts the interfaces added via
add_veth into /etc/resolv.conf on Debian, and if there are two of
these, they will forward queries to each other endlessly.  This seems
to slow down the test enough to sometimes cause timeouts elsewhere.


martinpitt:
[d-testing failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17712-20220905-085347-9cc14efc-debian-testing/log.html) looks unrelated, retrying.

martinpitt (approved):
Danke!

PR #17720, opened 2022-09-08
Fixes for ubuntu-stable

mvollmer:


mvollmer:
Alright, that should be it!

test/verify/check-system-info
```diff
@@ -827,7 +827,12 @@ password=foobar
         self.assertLess(initial_usage, 80)
         # allocate an extra 300 MB; this may cause other stuff to get unmapped,
         # thus not exact addition, but usage should go up
-        mem_hog = m.spawn("MEMBLOB=$(yes | dd bs=1M count=300 iflag=fullblock); touch /tmp/hogged; sleep infinity", "mem_hog.log")
+        #
+        # The "true" after "sleep" is there to prevent bash from
+        # replacing it's own process with the sleep (as a "tail call
+        # optimization") and thereby dropping the memory blob too early.
+        #
+        mem_hog = m.spawn("MEMBLOB=$(yes | dd bs=1M count=300 iflag=fullblock); touch /tmp/hogged; sleep infinity; true", "mem_hog.log")
```
martinpitt, line 835: TIL! Thanks for figuring this out. In another place I'm invoking awk as I couldn't get bash to work reliably. Now I know why!

test/verify/check-shell-multi-machine
```diff
@@ -107,15 +107,20 @@ def change_ssh_port(machine, address, port=None, timeout_sec=120):
...
-    # We stop the sshd.socket unit and just go with a regular
-    # daemon.  This is more portable and reloading/restarting the
-    # socket doesn't seem to work well.
-    #
-    machine.execute("( ! systemctl is-active sshd.socket || systemctl stop sshd.socket) && systemctl restart sshd.service")
+    if machine.image in ["ubuntu-stable"]: # always socket activated
+        machine.write("/etc/systemd/system/ssh.socket.d/override.conf",
+                      f"[Socket]\nListenStream=\nListenStream=127.27.0.15:22\nListenStream={address}:{port}")
+        machine.execute("systemctl daemon-reload")
+        machine.execute("systemctl restart ssh.socket")
```
martinpitt, line 114: In fact Fedora and RHEL also have sshd.socket -- I am just never sure which one is the  "intended" one to start/enable. Possibly we could use this approach on more, or even all OSes? Anyway, something for later.

However, is it really ssh.socket, or is that just an alias to sshd.socket? It's the latter in Fedora & Co. Wouldn't it be cool if distros could agree on such things?

test/verify/check-shell-multi-machine
```diff
@@ -706,25 +706,7 @@ class TestMultiMachine(MachineCase):
...
-        else:
-            m2.execute("! selinuxenabled || semanage port -a -t ssh_port_t -p tcp 2222")
-        m2.execute("sed -i 's/.*Port .*/#\\0/' /etc/ssh/sshd_config")
-        m2.execute("printf 'ListenAddress 127.27.0.15:22\nListenAddress 10.111.113.2:2222\n' >> /etc/ssh/sshd_config")
-
-        # We stop the sshd.socket unit and just go with a regular
-        # daemon.  This is more portable and reloading/restarting the
-        # socket doesn't seem to work well.
-        m2.execute("( ! systemctl is-active sshd.socket || systemctl stop sshd.socket) && systemctl restart sshd.service")
+        change_ssh_port(m2, "10.111.113.2", 2222)
```
martinpitt, line 709: This change needs to run on all OSes, so please drop no-test.

test/verify/check-shell-multi-machine
```diff
@@ -706,25 +706,7 @@ class TestMultiMachine(MachineCase):
         b.logout()
         b.wait_visible("#login")
 
-        if m1.image in ["fedora-coreos"]:  # no semanage
-            m1.execute("setenforce 0")
-        else:
-            m1.execute("! selinuxenabled || semanage port -a -t ssh_port_t -p tcp 2222")
```
martinpitt, line 707: Yep, this was surely a copy&paste error.

martinpitt (changes requested):
Thanks! However, unit tests are unhappy:

    # test/verify/check-shell-multi-machine:110:43: E261 at least two spaces before inline comment

I think after that you can un-draft. It still needs to land in lockstep with the bots PR, thus adding blocked.

test/verify/check-shell-multi-machine
mvollmer: There is no alias for the socket unit, unfortunately, only for the service.

test/verify/check-shell-multi-machine
mvollmer: Yes, thanks.

test/verify/check-shell-multi-machine
mvollmer: Yeah, thanks for confirming.

martinpitt (approved):
Thanks! I expect the ubuntu-stable test to fail (only succeed against the refreshed image).

PR #17721, opened 2022-09-08
Don't use `data-label` for selectors in CSS

garrett:
As spotted at https://github.com/cockpit-project/cockpit/pull/17657#pullrequestreview-1086711740, we shouldn't use data-label in CSS for selectors. Doing so ties it to the language specified.


garrett:
There's still a use of "%", but I guess this won't change across languages? Even still, we should probably remove that as well. (It's not included in this PR.)

martinpitt (approved):
Oops! Thanks for spotting.

PR #17723, opened 2022-09-09
python: remove "dbus_internal" hack

allisonkarlitskaya:
Remove the special-casing of the "internal" dbus channel and implement it as a peer-to-peer connection inside of the bridge.

The other side of this connection lives inside of a new file, `internal_endpoints.py`, which contains the objects that we want to appear on the internal D-Bus.

Most of these are still stubbed out, but the parts that were implemented before have been copied over.

Bump our dependency on systemd_ctypes for the required APIs for peer-to-peer connections and object exporting.


src/cockpit/internal_endpoints.py
```diff
@@ -0,0 +1,138 @@
...
+@bus.Object.interface('cockpit.Packages')
+class PackagesEndpoint(bus.Object):
+    ...
+
+
+@bus.Object.interface('cockpit.Superuser')
+class SuperuserEndpoint(bus.Object):
+    @bus.Object.method(in_types=['s'])
+    def start(self, _bridge):
+        ...
```
martinpitt, line 62: Hmm, why doesn't this make the pycheck scenario burst up in flames? This file is full of `...`

martinpitt: `<class 'ellipsis'>` what on earth is that now? Sometimes it feels like Python is getting worse..

martinpitt (approved):
Thanks!

src/cockpit/internal_endpoints.py
allisonkarlitskaya: `...` is a conventional alternative for "I'll implement this later"

It's basically equivalent to a docstring, or `pass`.

PR #17726, opened 2022-09-11
login: Use valid selectors when testing for :is() / :where() support.

### LINKED ISSUE #17724: This web browser is too old to run the Web Console (missing selector(:is():where()))
### Explain what happens

Firefox Nightly now reports unsupported CSS selectors (see [issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1789248)).
Since that fix was implemented, Firefox Nightly users can no longer log in into Cockpit. This change might eventually reach upstream stable branch as Firefox follow W3C guidelines on how things should be reported or implemented on the engine.

I must say, **the UI was working fine before that change**, therefore I'm not sure if it is essential for Cockpit to check for `selector(:is():where())` on the login page and prevent users from logging in if it isn't supported.

Maybe the FF change broke something else, or maybe it was simply unsupported from the start. I never encountered issues with Cockpit before that change. The opened issue is currently awaiting triage on FF end. Details on if a commit broke something else, or if this behaviour is 100% intended, is still unclear.

https://bugzilla.mozilla.org/show_bug.cgi?id=1790259

I wanted to post this here for:
- Maybe Cockpit remove that check;
- Other fellow users that might eventually encounter such issue.

### Version of Cockpit

264.1

Unknown or not applicable

### Server operating system

CentOS

8.6

Firefox

### System log

```shell
Not relevant.
```

emilio:
Rejecting CSS.supports(":is()") is the correct thing to do per https://github.com/w3c/csswg-drafts/issues/7280

Fixes #17724.

### DIFF
--- a/pkg/static/login.js
+++ b/pkg/static/login.js
@@ -225,7 +225,7 @@
                css("display", "flex") &&
                css("display", "grid") &&
                css("selector(test)") &&
-               css("selector(:is():where())");
+               css("selector(:is(*):where(*))");
     }
 
     function trim(s) {


garrett (approved):
Thanks for spotting this! Confirmed working here.

This is a critical fix that should go into all backports, as it prevents upcoming browsers from signing into Cockpit.

martinpitt:
@TomasTomecek, @lachmanfrantisek : According to https://copr.fedorainfracloud.org/coprs/packit/cockpit-project-cockpit-17726/build/4832346/ the COPR builds succeeded, but somehow packit and the PR don't pick that up? Is it worth retrying?

mvollmer:
I have tested this with Firefox 106.0a1 and everything checks out.  I can reproduce the issue exactly, and this change makes eveyrthing work. Thanks!

mvollmer:
> Is it worth retrying?

No. :-)

mgol:
@emilio it'd be good to get WPTs for no-parameter `:is()` / `:has()`, both in `qSA` and in `CSS.supports("selector(...)")`.

0xSheepdog:
I am seeing this error again on RHEL8 (cockpit 264.2-1.el8_6) from Firefox 106.0.3 64-bit on Windows 10 Enterprise (ver 21H2, OS build 19044.2130).
Did it come back, or am I maybe missing an update or piece of code somewhere?

Verld:
> Rejecting CSS.supports(":is()") is the correct thing to do per [w3c/csswg-drafts#7280](https://github.com/w3c/csswg-drafts/issues/7280)
> 
> Fixes #17724.
> 
> ## login: Adjust selectors when checking for `:is()` and `:where()` support
> Newer, upcoming browser versions have improved next-level support for `:is()` and `:where()` selectors. Cockpit was checking for support an empty usage which passed on browsers in the earlier (current as of writing) implementation. However, browsers have recently updated their parsing support for "[Forgiving Selector Parsing](https://developer.mozilla.org/en-US/docs/Web/CSS/:is#forgiving_selector_parsing)", which caused the newer development versions of Firefox, Chrome, and WebKit to fail this check, preventing the browsers from logging into Cockpit.
> 
> The check has now been adjusted so current and upcoming browser versions all pass.
> 
> Additionally, hotfixes for older supported affected versions of Cockpit have been published for CentOS, Debian, and Ubuntu. If you have an error while trying to log in with a new browser on an older Cockpit version, please upgrade your version of Cockpit.
> 
> _Huge thanks to [Emilio Cobos Álvarez](https://github.com/emilio) for bringing this to our attention and sending a PR with a fix!_

Is there any browser extension that adds support for :is() and :where()? Don't Edge and Chromium support extensons for third-party/custom JavaScript procedures?

garrett:
> Is there any browser extension that adds support for :is() and :where()? Don't Edge and Chromium support extensons for third-party/custom JavaScript procedures?

(No, there's no extension for this, nor should there be. It's simply not needed. Browsers have had this functionality for years already. Only extremely insecure browser versions don't support it.)

Chromium and Edge have both supported `:is()` and `:where() ` for a very, very long time in web browser time (31 major versions ago for Chromium and 42 major versions ago for Firefox). This bug was also fixed [a while ago](https://cockpit-project.org/blog/cockpit-277) in Cockpit (27 major releases ago), and it shouldn't affect anyone, as all distributions have since been fixed (most at the time of publication of Cockpit release 277, in September of 2022). People were only somewhat affected in the timespan of a week or two (depending on your browser version and distro) of the middle of 2022, with newer browser versions at the time and older Cockpit versions also at the time.

https://caniuse.com/mdn-css_selectors_where (Jan 2021 for Blink based browsers, which includes Chromium and Edge. Even earlier for Firefox and WebKit.)

This was fixed in Cockpit, along with backports in the middle of last year. (See dates above.) There was only a short week or two window when browsers updated their usage of `:is()` and `:where()` and didn't support Cockpit's usage. Earlier browsers didn't have a problem, and Cockpit was quickly updated for later browser versions too. We even published a workaround for an immediate fix (as some distributions took a week or two longer than expected for a hotfix).

Make sure you **update your software**: Update both your system that has Cockpit (including the Cockpit packages) and _especially_ your browsers. If you're seeing this problem, then you're woefully out of date with Cockpit _and_ your browser, and probably have security problems on your system (assuming your browser and/or your system are out of date — there have been many CVEs in various parts of all OSes and browsers since this timeframe).

**TL;DR (summary)**: Update your version of Cockpit (and your system too, as other packages are probably out of date as well) and make sure your browser is also up to date.

PR #17728, opened 2022-09-12
ws: Go back to assuming GChildWatch on mips in test-channelresponse

martinpitt:
Commit 0c7793da868cc adjusted the "wait for killed bridge" behaviour for glib 2.73.2: that dropped the global GChildWatch waiting in favor of pidfds. But on mips{,64}el, glib now went back to the old behaviour, so adjust the condition.

See https://buildd.debian.org/status/fetch.php?pkg=cockpit&arch=mipsel&ver=276-1&stamp=1662541707&file=log

----

I can't say I like this hack, but I don't really have a good idea how to predict the behaviour otherwise.

I tested this on a Debian mips porter box, and it makes the test happy. I keep the build environment around for a day or so.

See https://buildd.debian.org/status/package.php?p=cockpit for an overview of all build logs (current and older)

### DIFF
--- a/src/ws/test-channelresponse.c
+++ b/src/ws/test-channelresponse.c
@@ -530,7 +530,7 @@ test_resource_failure (TestResourceCase *tc,
   g_assert_cmpint (pid, >, 0);
   g_assert_cmpint (kill (pid, SIGTERM), ==, 0);
   /* Wait until it's gone */
-#if GLIB_CHECK_VERSION(2,73,2)
+#if GLIB_CHECK_VERSION(2,73,2) && !defined(__MIPSEL)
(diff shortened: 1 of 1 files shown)


martinpitt:
FTR, the official Debian buildd and the porter box are both running Linux 4.19, so that wouldn't support pidfds yet, and fail the runtime check in gmain.c. So instead of this, let's maybe gate this on the running kernel version? Or copy glib's `HAVE_PIDFD` configure test? It's rather annoying that glib changed its runtime behaviour in such an unpredictable way. Or should we do this in some entirely different way?

allisonkarlitskaya (approved):
honestly, since it's just a test and it's a known and understood issue, I wouldn't waste any more time here

PR #17729, opened 2022-09-12
tools: replace usage of deprecated egrep

jelly:
Since grep 3.8, there is now a warning about egrep being deprecated and users should switch to `grep -E`.

See https://lists.gnu.org/archive/html/info-gnu/2022-09/msg00001.html


martinpitt (approved):
Dankjewel!

PR #17730, opened 2022-09-12
login: Fix username/password autofill

skazi0:
Some browsers get confused when there is some type="password" field before the username field. Re-ordering mutually exclusive parts of the login page should not have any visible impact but fixes automatic filling of credentials saved in browser's password manager.

The password field was also explicitly cleared in all auth scenarios while it should be only for conversations.

### DIFF
--- a/pkg/static/login.html
+++ b/pkg/static/login.html
@@ -50,12 +50,6 @@
     <div id="login" class="login-area" hidden>
       <form onsubmit="return false">
 
-        <div id="conversation-group" class="form-group" hidden>
-          <div id="conversation-message"></div>
-          <label id="conversation-prompt" for="conversation-input"></label>
-          <input type="password" class="form-control" id="conversation-input">
-        </div>
-
         <div id="hostkey-group" class="form-group" hidden>
           <h1 id="hostkey-title"></h1>
           <div id="hostkey-warning-group" class="pf-c-alert pf-m-warning pf-m-inline dialog-error" aria-label="inline warning alert" hidden>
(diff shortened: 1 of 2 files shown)


garrett (changes requested):
Interesting idea.

Which browsers though? I tried Firefox, Chrome, and GNOME Web (WebKit), and couldn't get password management to work as expected (prefilling saved password), except for the username.

skazi0:
@garrett I tried applying those changes in place it it looks like after moving the `conversation-input` down, it *does* fill the form correctly *but* one piece of JS cleans the password explicitly. It was added as part of https://github.com/cockpit-project/cockpit/commit/74e58fc8c12a4f3fa0033bbdd3f9be79519d416f so I assume it was meant to be done only for conversation auth. I don't have a test env for checking how this works in those cases. Can you test this somehow?

garrett (approved):
I've tried this latest iteration locally and also in a VM and found it working in Firefox and Chrome.

Thanks!

garrett:
> It was added as part of https://github.com/cockpit-project/cockpit/commit/74e58fc8c12a4f3fa0033bbdd3f9be79519d416f so I assume it was meant to be done only for conversation auth. I don't have a test env for checking how this works in those cases. Can you test this somehow?

We might want to have @mvollmer look at the auth stuff. He's the one who worked the most on it recently. @martinpitt could also be good too. We don't want to break the login page accidentally... :sweat_smile:

marusak (approved):
I have tested the change in .js and could not find any case where this would be an issue, seems fine!

PR #17738, opened 2022-09-15
Give more time for candlepin

skobyda:
In recent rhel-7-9 image refresh[1], it seems candlepin began to timeout:
+ curl -s https://services.cockpit.lan:8443/candlepin --verbose
*   Trying 10.111.112.100...
[timeout]

[1] https://github.com/cockpit-project/bots/pull/3866

### DIFF
--- a/test/verify/check-subscriptions
+++ b/test/verify/check-subscriptions
@@ -109,7 +109,7 @@ class SubscriptionsCase(MachineCase):
 
         # Wait for the web service to be accessible
         args = {"addr": "services.cockpit.lan"}
-        m.execute(WAIT_SCRIPT % args)
+        m.execute(WAIT_SCRIPT % args, timeout=360)
 
         if m.image != "rhel-atomic":  # insights-client not installed there
             m.write("/etc/insights-client/insights-client.conf",


skobyda:
Similar situation as https://github.com/cockpit-project/cockpit/commit/33c002e642b11685793a7eb1a102b12360c3c3f2

skobyda:
For folks from the future looking at this PR, this is the test failure:

```
# testInsights (check_subscriptions.TestSubscriptions)
#
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
++ seq 1 200
+ for x in '$(seq 1 200)'
+ curl -s https://services.cockpit.lan:8443/candlepin
+ sleep 1
+ for x in '$(seq 1 200)'
+ curl -s https://services.cockpit.lan:8443/candlepin
Journal extracted to TestSubscriptions-testInsights-services-127.0.0.2-2402-FAIL.log
  File "/work/bots/make-checkout-workdir/test/verify/check-subscriptions", line 112, in setUp
    m.execute(WAIT_SCRIPT % args)
  File "/work/bots/make-checkout-workdir/bots/machine/machine_core/ssh_connection.py", line 307, in execute
    res = subprocess.run(command_line,
  File "/usr/lib64/python3.10/subprocess.py", line 503, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib64/python3.10/subprocess.py", line 1152, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib64/python3.10/subprocess.py", line 2003, in _communicate
    ready = selector.select(timeout)
  File "/usr/lib64/python3.10/selectors.py", line 416, in select
    fd_event_list = self._selector.poll(timeout)
  File "/work/bots/make-checkout-workdir/bots/machine/machine_core/timeout.py", line 41, in handle_timeout
    raise RuntimeError(self.error_message)
RuntimeError: Timed out on '
set -x
for x in $(seq 1 200); do
    if curl -s https://services.cockpit.lan:8443/candlepin; then
        break
    else
        sleep 1
    fi
done
'

not ok 239 testInsights (check_subscriptions.TestSubscriptions) # duration: 150s
```

skobyda:
`rhel-7-9@bots#3866` passes

martinpitt (approved):
It feels really dubious that candlepin would take more than 2.5 **minutes** to start on the already booted image. This rather smells like something is actually broken there, and the pass/fail is just some sheer luck, regardless of the timeout? Then the only effect of this would be to make failed tests even slower.

But oh well, this is our "old cruft" branch, ok to not waste too much time on this for now. Thanks!

PR #17736, opened 2022-09-14
storage: Assorted improvements

mvollmer:

PR #17739, opened 2022-09-16
python: ObjectManager support for the dbus payload

mvollmer:
This implements "add_watch" with a path_namespace that points at a ObjectManager.  This also takes care to deliver "notify" and "meta" message before method call replies and to not reorder signals.  This is expected by the DbusProxy and DbusProxies implemented in cockpit.js.

It is an error when there is no ObjectManager at path_namespace.  The C bridge falls back to emulating an object manager via recursive introspection in this case, but we might not ever want to implement that in the Python bridge.

- [x] run the unit tests that should be green now
- [x] get the "proxies" unit test green
- [x] #17752 
- [x] Add tests from @allisonkarlitskaya and get them green

### DIFF
--- a/pkg/base1/test-dbus-common.js
+++ b/pkg/base1/test-dbus-common.js
@@ -262,29 +262,6 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad variants", function (assert) {
-        const done = assert.async();
-        assert.expect(3);
-
-        const dbus = cockpit.dbus(bus_name, channel_options);
-        dbus.call("/otree/frobber", "com.redhat.Cockpit.DBusTests.Frobber",
-                  "TestAsv", [{
-                      one: "foo",
-                      two: "/bar",
-                      three: "assgit",
-                      four: 42,
-                      five: 1000.0
-                  }])
-                .fail(function(ex) {
-                    assert.equal(ex.name, "org.freedesktop.DBus.Error.InvalidArgs", "error name");
-                    assert.equal(ex.message, "Unexpected type 'string' in argument", "error message");
-                })
-                .always(function() {
-                    assert.equal(this.state(), "rejected", "should fail");
-                    done();
-                });
-    });
-
     QUnit.test("get all", function (assert) {
         const done = assert.async();
         assert.expect(2);
--- a/pkg/base1/test-dbus-common.js
+++ b/pkg/base1/test-dbus-common.js
@@ -50,7 +50,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("call method with timeout", function (assert) {
+    QUnit.test.skipWithPybridge("call method with timeout", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -142,7 +142,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("integer bounds", function (assert) {
+    QUnit.test.skipWithPybridge("integer bounds", function (assert) {
         assert.expect(35);
 
         const dbus = cockpit.dbus(bus_name, channel_options);
@@ -215,7 +215,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("variants", function (assert) {
+    QUnit.test.skipWithPybridge("variants", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -239,7 +239,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad variants", function (assert) {
+    QUnit.test.skipWithPybridge("bad variants", function (assert) {
         const done = assert.async();
         assert.expect(3);
 
@@ -315,7 +315,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("call bad base64", function (assert) {
+    QUnit.test.skipWithPybridge("call bad base64", function (assert) {
         const done = assert.async();
         assert.expect(3);
 
@@ -333,7 +333,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("call unknown", function (assert) {
+    QUnit.test.skipWithPybridge("call unknown", function (assert) {
         const done = assert.async();
         assert.expect(3);
 
@@ -413,7 +413,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("with types", function (assert) {
+    QUnit.test.skipWithPybridge("with types", function (assert) {
         const done = assert.async();
         assert.expect(3);
 
@@ -431,7 +431,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("with meta", function (assert) {
+    QUnit.test.skipWithPybridge("with meta", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -466,7 +466,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("empty base64", function (assert) {
+    QUnit.test.skipWithPybridge("empty base64", function (assert) {
         const done = assert.async();
         assert.expect(3);
 
@@ -484,7 +484,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad object path", function (assert) {
+    QUnit.test.skipWithPybridge("bad object path", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -499,7 +499,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad interface name", function (assert) {
+    QUnit.test.skipWithPybridge("bad interface name", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -514,7 +514,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad method name", function (assert) {
+    QUnit.test.skipWithPybridge("bad method name", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -529,7 +529,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad flags", function (assert) {
+    QUnit.test.skipWithPybridge("bad flags", function (assert) {
         const done = assert.async();
         assert.expect(2);
 
@@ -544,7 +544,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad types", function (assert) {
+    QUnit.test.skipWithPybridge("bad types", function (assert) {
         const done = assert.async();
         assert.expect(3);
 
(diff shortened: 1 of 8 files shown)


mvollmer:
Some helper function that allows async functions as handlers for add_match would be nice.

mvollmer:
> Some helper function that allows coroutines as handlers for add_match would be nice.

Done.

martinpitt:
Does this fix the hanging test-dbus* unit tests by any chance? Right now there is zero coverage of this in the python bridge, as these are in `pyTESTS_HANGING`. There's the /pycheck scenario soon (see PR #17705), but that still works very poorly -- it'd be much better to let the unit tests work on the py bridge for now. Note that you can now do py vs. C bridge specific checks/behaviour there; see `git grep mock_info`.

mvollmer:
> Does this fix the hanging test-dbus* unit tests by any chance?

Some of them.  I will try to get test-dbus.html green here (by skipping most tests for the Python bridge, but not all).

mvollmer:
I have pushed a couple of bug fixes.

mvollmer:
rhel-9-2 failure is a flake that is hopefully fixed in #17736

src/cockpit/channels/dbus.py
```diff
@@ -135,28 +152,110 @@ async def do_call(self, call, message):
...
+
+        self.add_match(f"type='signal',sender='{self.name}',path='{path}',interface='org.freedesktop.DBus.ObjectManager'", handler)
+        objects, = await self.bus.call_method_async(self.name, path, 'org.freedesktop.DBus.ObjectManager', 'GetManagedObjects')
+        for p, ifaces in objects.items():
+            for iface, props in ifaces.items():
+                if interface_name is None or iface == interface_name:
+                    mm = await self.cache.get_interface_if_new(iface, self.bus, self.name, p)
+                    if mm:
+                        meta.update({iface: mm})
+                        notify.update({p: {iface: {k: v['v'] for k, v in props.items()}}})
```
mvollmer, line 220: This needs to be one level de-indented.  I though I had fixed that...

mvollmer: Fixed.

src/cockpit/channels/dbus.py
```diff
@@ -135,9 +135,9 @@ async def do_call(self, call, message):
         # We have to figure out the signature of the call.  Either we got told it:
         signature = message.get('type')
 
-        # ... or all of our arguments are string (and we can guess):
```
allisonkarlitskaya, line 121: This was meant to be a reasonable default in case the signature was not explicitly specified.

pkg/base1/test-dbus-common.js
```diff
@@ -239,30 +239,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 });
     });
 
-    QUnit.test("bad variants", function (assert) {
```
allisonkarlitskaya, line 242: Can you move the deduplication to a separate commit?

allisonkarlitskaya (commented):
Some comments on the non-main commits.

pkg/base1/test-dbus-common.js
mvollmer: Done.

pkg/base1/test-dbus.js
```diff
@@ -1,824 +1,21 @@
...
-};
-common_dbus_tests(options, "com.redhat.Cockpit.DBusTests.Test");
-dbus_track_tests(options, "com.redhat.Cockpit.DBusTests.Test");
+QUnit.mock_info("bridge").then(bridge => {
+    if (bridge == "cockpit-bridge.pyz")
+        QUnit.test.butNotForPy = QUnit.test.skip;
+    else
+        QUnit.test.butNotForPy = QUnit.test;
 
-QUnit.test("proxy no stutter", function (assert) {
```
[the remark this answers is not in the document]
mvollmer: I found a better way that requires less code movement: top-level await.  (Moving it all into dbus-common was a bad idea no matter what, and not necessaty, I guess.)

pkg/base1/test-dbus.js
```diff
@@ -782,17 +787,17 @@ Empty=
                 proxy = dbus.proxy("cockpit.Config", "/config");
                 return proxy.wait();
             })
-            // the above changes the config after bridge startup, so reload
+    // the above changes the config after bridge startup, so reload
```
mvollmer, line 790: This was done by eslint:fix.  I will remove this change.

src/cockpit/channels/dbus.py
```diff
@@ -135,28 +152,110 @@ async def do_call(self, call, message):
...
+
+        self.add_match(f"type='signal',sender='{self.name}',path='{path}',interface='org.freedesktop.DBus.ObjectManager'", handler)
+        objects, = await self.bus.call_method_async(self.name, path, 'org.freedesktop.DBus.ObjectManager', 'GetManagedObjects')
+        for p, ifaces in objects.items():
+            for iface, props in ifaces.items():
+                if interface_name is None or iface == interface_name:
+                    mm = await self.cache.get_interface_if_new(iface, self.bus, self.name, p)
+                    if mm:
+                        meta.update({iface: mm})
+                    notify.update({p: {iface: {k: v['v'] for k, v in props.items()}}})
```
mvollmer, line 220: This also needs to use `notify_udpate`.

mvollmer: Done.

mvollmer:
I will add some more documentation here, especially about message ordering and how the Lock realizes that.

mvollmer:
> I will add some more documentation here, especially about message ordering and how the Lock realizes that.

Done.

src/cockpit/channels/dbus.py
```diff
@@ -67,6 +159,14 @@ async def get_signature(self, interface_name, method, bus=None, destination=None
         return ''.join(interface['methods'][method]['in'])
 
 
+def notify_update(notify, path, interface_name, props):
+    data = {k: v['v'] for k, v in props.items()}
+    if path in notify:
+        notify[path][interface_name] = data
+    else:
+        notify[path] = {interface_name: data}
```
martinpitt, line 167: The standard Python idiom for this is:

    notify.setdefault(path, {})[interface_name] = data

pkg/base1/test-dbus-address.js
```diff
@@ -3,6 +3,11 @@ import { common_dbus_tests, dbus_track_tests } from "./test-dbus-common.js";
 
 import QUnit from "qunit-tests";
 
+if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz")
```
martinpitt, line 6: This is now spelled `await QUnit.mock_info("pybridge")`

pkg/base1/test-dbus-address.js
```diff
@@ -3,6 +3,11 @@ import { common_dbus_tests, dbus_track_tests } from "./test-dbus-common.js";
...
+    QUnit.test.butNotForPy = QUnit.test.skip;
```
martinpitt, line 7: Cool trick! Could you move this to pkg/lib/qunit-tests.js instead of duplicating it here and in test-dbus.js? We can then also use that in pkg/base1/test-http.js, two of the tests are also skipped on the pybridge.

I'd call it `.skipPybridge` to be a little more consistent with `.skip` and "pybridge", but bikeshedding :grin:

pkg/base1/test-dbus-common.js
```diff
@@ -1050,7 +1050,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
                 .always(function() {
                     assert.equal(this.state(), "resolved", "deleted stray objects");
 
-                    const proxies = dbus.proxies("com.redhat.Cockpit.DBusTests.Frobber");
+                    const proxies = dbus.proxies("com.redhat.Cockpit.DBusTests.Frobber", "/otree");
```
martinpitt, line 1053: Can we drop any conditional `.skip` with that change? ("test: Use ObjectManager in the D-Bus "proxies" test ")

martinpitt (changes requested):
Thanks! There are still a lot of TODOs here, but then again the whole pybridge is still that :wink: This looks fairly reasonable to me, and a nice incremental improvement -- anything that tests more things on the pybridge is good. This also blocks a fair share of follow-up work (you have at least two more PRs building on top of that), so I'd be in favor of landing that quickly.

I have a few small comments and this needs rebasing/deconflicting.

pkg/base1/test-dbus-address.js
mvollmer: Changed.

pkg/base1/test-dbus-address.js
mvollmer: Done!

pkg/base1/test-dbus-common.js
mvollmer: No, I think not. The other tests really want to test non-ObjectManager things.  We might never want to support that, but we shouldn't change the tests.

src/cockpit/channels/dbus.py
mvollmer: Done, thanks!

pkg/base1/test-locale.js
```diff
@@ -227,4 +227,9 @@ document.addEventListener("DOMContentLoaded", event => {
 
     /* Ready to go */
     QUnit.start();
-});
+}
+
+if (document.readyState == "loading")
+    document.addEventListener("DOMContentLoaded", init);
+else
+    init();
```
martinpitt, line 235: Hmm, is the `DOMContentLoaded` handler not called if the document is already "ready"? We use the old pattern a lot, if that is broken it should be replaced everywhere?

martinpitt (approved):
Thanks! My previous issues are resolved. The objectmanager handling code is a bit of a mouthful and maybe could use some splitting. Nevertheless, I'm good with this, and I'd rather not block it longer. @allisonkarlitskaya , do you still want to look at this, too?

pkg/base1/test-locale.js
mvollmer: This is a consequence of the new top-level "await" in qunit-tests.js.  It delays the execution of the rest of the script.

allisonkarlitskaya:
Here's a couple of test cases.  It would be cool to see those passing before we merge this.  (or tell me why my tests are dumb :rofl:)

[0001-python-add-two-failing-dbus-test-cases.txt](https://github.com/cockpit-project/cockpit/files/10106461/0001-python-add-two-failing-dbus-test-cases.txt)

mvollmer:
> Here's a couple of test cases. It would be cool to see those passing before we merge this. (or tell me why my tests are dumb rofl)

Yes, thanks!

src/cockpit/channels/dbus.py
```diff
@@ -220,9 +220,9 @@ async def do_call(self, call, message):
         # We have to figure out the signature of the call.  Either we got told it:
         signature = message.get('type')
 
-        # ... or all of our arguments are string (and we can guess):
-        if signature is None and all(isinstance(arg, str) for arg in args):
```
allisonkarlitskaya, line 126: Pretty sure it's caused by this change.  Before, `GetAll` was a method where all parameters were strings, so we skipped introspection.

I suspect this is a case of uncovering a bug rather than making a new one, but it's still a regression...

allisonkarlitskaya (changes requested):

allisonkarlitskaya:
Here's the merge, in case you want to play with it: https://github.com/cockpit-project/cockpit/tree/all-the-things

allisonkarlitskaya (changes requested):
Thanks for working on this, and sorry that it's taken so long.

Here's my high-level take, after our discussion today:

 - we should merge this, more or less as it is
 - please squash the final commit (python: ObjectManager support without a service name) into the first one, since it's basically a fixup
 - please either drop the second commit (python: Don't assume strings are always 's' in D-Bus method calls) or fix the regression uncovered by it (see comment above)

In terms of sequencing, I think this should (and will) go in before the peer bridge stuff, so don't worry about the conflicts in the test cases — I'll resolve them from my side.

allisonkarlitskaya:
The other issues (like deduplication of signal stuff) can afford to wait.  It's not really blocking anything, as far as I know, and it might be totally unserious.  I also think we could investigate if we find a better mechanism for dealing with it.

mvollmer:
> * please either drop the second commit (python: Don't assume strings are always 's' in D-Bus method calls) or fix the regression uncovered by it (see comment above)

Dropping this commit will break method calls with a signature like 'o', so this is not an option.

mvollmer:
> * fix the regression uncovered by it

See the "Python: Support D-Bus introspection without a service name" commit.

allisonkarlitskaya (commented):
The commits are all backwards :rofl: 

I tried rebasing this to the order I'd have liked to have seen it, but it's just a giant ball of conflicts.  Let's not waste more time.  If everything commit here passes `pytest` (checking now) then I'm happy to land it like it is.

allisonkarlitskaya (approved):
:rocket: los!

PR #17741, opened 2022-09-20
python: Handle early errors during cockpit.spawn

mvollmer:
- [x] #17758 
- [x] land #17763 and rebase

### DIFF
--- a/pkg/base1/test-spawn-proc.js
+++ b/pkg/base1/test-spawn-proc.js
@@ -91,6 +91,16 @@ QUnit.test("error message fail", function (assert) {
             });
 });
 
+QUnit.test("nonexisting executable", assert => {
+    assert.rejects(cockpit.spawn(["/bin/nonexistent"]),
+                   ex => ex.problem == "not-found");
+});
+
+QUnit.test("permission denied", assert => {
+    assert.rejects(cockpit.spawn(["/etc/hostname"]),
+                   ex => ex.problem == "access-denied");
+});
+
 QUnit.test("write eof read", function (assert) {
     const done = assert.async();
     assert.expect(2);
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -19,7 +19,7 @@ import asyncio
 import logging
 import os
 
-from ..channel import AsyncChannel
+from ..channel import AsyncChannel, ChannelError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


pkg/base1/test-spawn-proc.js
```diff
@@ -91,6 +91,32 @@ QUnit.test("error message fail", function (assert) {
             });
 });
 
+QUnit.test("nonexisting executable", function (assert) {
+    const done = assert.async();
+    assert.expect(2);
+    cockpit.spawn(["/bin/nonexistent"])
```
allisonkarlitskaya, line 97: Please rewrite this using async/await syntax.

pkg/base1/test-spawn-proc.js
```diff
@@ -91,6 +91,32 @@ QUnit.test("error message fail", function (assert) {
...
+                assert.equal(this.state(), "rejected", "failed");
+                done();
+            });
+});
+
+QUnit.test("permission denied", function (assert) {
+    const done = assert.async();
+    assert.expect(2);
+    cockpit.spawn(["/etc/hostname"])
+            .fail(function(error) {
```
allisonkarlitskaya, line 111: ditto.

src/cockpit/channels/stream.py
```diff
@@ -94,13 +94,26 @@ async def stream_process(self, args, options):
...
+        try:
+            process = await asyncio.create_subprocess_exec(
+                *args,
+                cwd=cwd,
+                env=env,
+                stdin=asyncio.subprocess.PIPE,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=stderr)
+        except FileNotFoundError:
+            problem = "not-found"
```
allisonkarlitskaya, line 107: I do not like it.  Let me propose something that might help.

pkg/base1/test-spawn-proc.js
mvollmer: I think this is actually best expressed with `assert.rejects`. WDYT?

mvollmer:
> test-spawn-proc unfortunately hangs on "stream large output". Let's try to fix that before merging this.

https://github.com/cockpit-project/cockpit/pull/17758

pkg/base1/test-spawn-proc.js
allisonkarlitskaya: I agree.

src/cockpit/channels/stream.py
```diff
@@ -94,13 +94,21 @@ async def stream_process(self, args, options):
...
+                cwd=cwd,
+                env=env,
+                stdin=asyncio.subprocess.PIPE,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=stderr)
+        except FileNotFoundError as error:
+            raise ChannelError('not-found') from error
+        except PermissionError as error:
+            raise ChannelError('access-denied') from error
+        except Exception as error:
```
allisonkarlitskaya, line 109: The only question I have: I wonder if this should rather be OSError.

allisonkarlitskaya (approved):
Can land like this, but see one suggestion/comment below.

src/cockpit/channels/stream.py
mvollmer: Right, I think "yes".

allisonkarlitskaya (approved):
:+1: thanks!

allisonkarlitskaya (approved):
[check, check, checkity check, check](https://www.youtube.com/watch?v=_lYdxQ6vDGQ&t=17s)

mvollmer:
Let's ignore the rhel-8-8 flake.

PR #17750, opened 2022-09-21
python: Track name owner for dbus channel

mvollmer:
- [x] #17739

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -17,7 +17,6 @@
 
 # Missing stuff compared to the C bridge that we should probably add:
 #
-# - name tracking
 # - connecting to given address instead of bus
 # - some more ways to connect to the internal bus (like { bus: "none", address: "internal" })
 # - removing matches
@@ -172,10 +171,46 @@ class DBusChannel(Channel):
     matches = None
     name = None
     bus = None
+    owner = None
 
     # This needs to be a fair mutex so that outgoing messages don't
     # get re-ordered.  asyncio.Lock is fair.
     watch_processing_lock = asyncio.Lock()
+
+    async def setup_name_owner_tracking(self):
+        def send_owner(owner):
+            # We must be careful not to send duplicate owner
+            # notifications. cockpit.js relies on that.
+            if self.owner != owner:
+                self.owner = owner
+                self.send_message(owner=owner)
+
+        def handler(message):
+            name, old, new = message.get_body()
+            send_owner(owner=new if new != "" else None)
+        rule = f"type='signal',sender='org.freedesktop.DBus',path='/org/freedesktop/DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',arg0='{self.name}'"
+        self.matches.append(self.bus.add_match(rule, handler))
+        try:
+            unique_name, = await self.bus.call_method_async("org.freedesktop.DBus",
+                                                            "/org/freedesktop/DBus",
+                                                            "org.freedesktop.DBus",
+                                                            "GetNameOwner", "s", self.name)
+        except BusError as error:
+            if error.name == "org.freedesktop.DBus.Error.NameHasNoOwner":
+                # Try to start it.  If it starts successfully, we will
+                # get a NameOwnerChanged signal (which will set
+                # self.owner) before StartServiceByName returns.
+                try:
+                    await self.bus.call_method_async("org.freedesktop.DBus",
+                                                     "/org/freedesktop/DBus",
+                                                     "org.freedesktop.DBus",
+                                                     "StartServiceByName", "su", self.name, 0)
+                except BusError as error:
+                    logger.debug("Failed to start service '%s': %s", self.name, error.message)
+            else:
+                logger.debug("Failed to get owner of service '%s': %s", self.name, error.message)
+        else:
+            send_owner(unique_name)
 
     def do_open(self, options):
         self.cache = InterfaceCache()
@@ -205,7 +240,19 @@ class DBusChannel(Channel):
             if err.errno != errno.EBUSY:
                 raise
 
-        self.ready()
+        if self.name is not None:
+            async def get_ready():
+                async with self.watch_processing_lock:
+                    await self.setup_name_owner_tracking()
+                    if self.owner:
+                        self.ready(unique_name=self.owner)
+                    else:
+                        self.close(problem="not-found")
+            task = asyncio.create_task(get_ready())
+            self.tasks.add(task)
+            task.add_done_callback(self.tasks.discard)
+        else:
+            self.ready()
 
     def add_match(self, rule, handler):
         def sync_handler(message):
--- a/pkg/base1/test-dbus-common.js
+++ b/pkg/base1/test-dbus-common.js
@@ -1201,7 +1201,7 @@ export function common_dbus_tests(channel_options, bus_name) { // eslint-disable
 }
 
 export function dbus_track_tests(channel_options, bus_name) { // eslint-disable-line no-unused-vars
-    QUnit.test.skipWithPybridge("track name", function (assert) {
+    QUnit.test("track name", function (assert) {
         const done = assert.async();
         assert.expect(4);
 
--- a/pkg/base1/test-dbus.js
+++ b/pkg/base1/test-dbus.js
@@ -153,7 +153,7 @@ QUnit.test.skipWithPybridge("bad dbus bus", function (assert) {
     });
 });
 
-QUnit.test.skipWithPybridge("wait ready", function (assert) {
+QUnit.test("wait ready", function (assert) {
     const done = assert.async();
     assert.expect(1);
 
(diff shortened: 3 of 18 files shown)


allisonkarlitskaya (commented):
I feel like maybe we should add https://www.freedesktop.org/software/systemd/man/sd_bus_track_add_name.html to systemd_ctypes ?

mvollmer:
> I feel like maybe we should add https://www.freedesktop.org/software/systemd/man/sd_bus_track_add_name.html to systemd_ctypes ?

Yeah... I frist dismissed it as being designed for a service to track its clients, but i guess that doesn't matter and we can still use it in a client to track a service.  But then I didn't see any way to get the unique name for a well known name, or notifications when a name owner changes... How is this supposed to work?

mvollmer:
> > I feel like maybe we should add https://www.freedesktop.org/software/systemd/man/sd_bus_track_add_name.html to systemd_ctypes ?
> 
> Yeah... I first dismissed it as being designed for a service to track its clients, but i guess that doesn't matter and we can still use it in a client to track a service. But then I didn't see any way to get the unique name for a well known name, or notifications when a name owner changes... How is this supposed to work?

Ah, there is a handler in sd_bus_track_new...

mvollmer:
> > > I feel like maybe we should add https://www.freedesktop.org/software/systemd/man/sd_bus_track_add_name.html to systemd_ctypes ?
> > 
> > 
> > Yeah... I first dismissed it as being designed for a service to track its clients, but i guess that doesn't matter and we can still use it in a client to track a service. But then I didn't see any way to get the unique name for a well known name, or notifications when a name owner changes... How is this supposed to work?
> 
> Ah, there is a handler in sd_bus_track_new...

And then there is https://www.csslayer.info/wordpress/linux/one-bad-api-in-sd-bus/ ...

And will still need to call StartServiceByName explicitly.  So I am on the fence right now, leaning away from sd_bus_track.

mvollmer:
> And will still need to call StartServiceByName explicitly. So I am on the fence right now, leaning away from sd_bus_track.

Of course, we might want to drop that functionality, our JS shouldn't really rely on this anyway.

mvollmer:
> > And will still need to call StartServiceByName explicitly. So I am on the fence right now, leaning away from sd_bus_track.
> 
> Cool. My first blush of that API, as well, is that it's more trouble than it's worth. I just wanted to mention it. If you're leaning against it, then I'll lean with you ;)

We can't use it to implement the `owner` messages faithfully since it wont give us the actual unique name for a well known name, nor does it seem to be able to notify us when a owner appears, only when it disappears.  So those are other reason.  The sd_bus_track API is made for a service to track a client while the servcie is handling a request of that client, so that the service can cleanly cancel the processing when the client disappears in the middle of that, and nothing else.

mvollmer:
@martinpitt, this makes a lot of the NetworkManager integration tests green.

src/cockpit/channels/dbus.py
```diff
@@ -172,11 +171,54 @@ class DBusChannel(Channel):
...
+            # notifications. cockpit.js relies on that.
+            if self.owner != owner:
+                self.owner = owner
+                self.send_message(owner=owner)
+        async def handler(message):
+            if message.get_member() == "NameOwnerChanged":
+                name, old, new = message.get_body()
+                if name == self.name:
+                    send_owner(owner=new if new != "" else None)
+        self.add_match("type='signal',sender='org.freedesktop.DBus',path='/org/freedesktop/DBus',interface='org.freedesktop.DBus'", handler)
```
allisonkarlitskaya, line 199: Why not add `member='NameOwnerChanged'` and an `arg0=''` for the well known name to the match rule?

src/cockpit/channels/dbus.py
```diff
@@ -172,11 +171,54 @@ class DBusChannel(Channel):
...
+                    await self.bus.call_method_async("org.freedesktop.DBus",
+                                                     "/org/freedesktop/DBus",
+                                                     "org.freedesktop.DBus",
+                                                     "StartServiceByName", "su", self.name, 0)
+                except BusError as error:
+                    logger.debug("Failed to start service '%s': %s", self.name, error.message)
+            else:
+                logger.debug("Failed to get owner of service '%s': %s", self.name, error.message)
+        else:
+            send_owner(unique_name)
```
allisonkarlitskaya, line 220: Can't we just send a `Ping` on the well-known name and check the sender field of the reply message?

allisonkarlitskaya: ...but, do we really want to implicitly start it?  This stuff always seemed so nasty to me... :(

src/cockpit/channels/dbus.py
```diff
@@ -172,11 +171,54 @@ class DBusChannel(Channel):
...
+        self.matches.append(self.bus.add_match(rule, sync_handler))
+
+    async def setup_name_owner_tracking(self):
+        def send_owner(owner):
+            # We must be careful not to send duplicate owner
+            # notifications. cockpit.js relies on that.
+            if self.owner != owner:
+                self.owner = owner
+                self.send_message(owner=owner)
+        async def handler(message):
```
allisonkarlitskaya, line 194: I don't understand why this needs to be async.  It doesn't await anything.

src/cockpit/channels/dbus.py
```diff
@@ -172,11 +171,54 @@ class DBusChannel(Channel):
...
     name = None
     bus = None
+    owner = None
 
     # This needs to be a fair mutex so that outgoing messages don't
     # get re-ordered.  asyncio.Lock is fair.
     watch_processing_lock = asyncio.Lock()
 
+    def add_match(self, rule, handler):
+        def sync_handler(message):
```
allisonkarlitskaya, line 181: I don't like this at all.  Handlers should be simple callbacks.

allisonkarlitskaya (changes requested):

src/cockpit/channels/dbus.py
martinpitt: Does that even have a chance to work? Normally, D-Bus autostart service files should do this automatically ("bus activation"), for services where it's intended and declared. Unless you explicitly connect with the NO_AUTOSTART flag. But some services like NetworkManager really shouldn't get auto-activated; why would a manual StartServiceByName suddenly work then, if it didn't work the first time around?

mvollmer: I have copied this from GDBus name tracking, which is used by the C bridge (IIRC, it's been a while).

src/cockpit/channels/dbus.py
mvollmer: It's async because self.add_match below expects a async function. I did that because it is slightly more convenient for other users of self.add_match, but it's not clean, agreed.

src/cockpit/channels/dbus.py
mvollmer: Good idea! Done.

src/cockpit/channels/dbus.py
mvollmer: We have some code that relies on this to figure out whether a service is available or not.  For example, the Storage page relies on the channel being closed with "not-found" to figure out that stratis2 isn't available and we should be using stratis3.

src/cockpit/channels/dbus.py
mvollmer: This is old code that just got moved around. I'll try to improve this in another PR.

src/cockpit/channels/dbus.py
mvollmer: I have changed this to use self.bus.add_match directly.

src/cockpit/channels/dbus.py
```diff
@@ -172,11 +171,47 @@ class DBusChannel(Channel):
...
+            # We must be careful not to send duplicate owner
+            # notifications. cockpit.js relies on that.
+            if self.owner != owner:
+                self.owner = owner
+                self.send_message(owner=owner)
+
+        def handler(message):
+            name, old, new = message.get_body()
+            send_owner(owner=new if new != "" else None)
+        rule = f"type='signal',sender='org.freedesktop.DBus',path='/org/freedesktop/DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',arg0='{self.name}'"
```
allisonkarlitskaya, line 191: You seem to be putting `rule` on its own line to avoid excessive line length, but I don't think it's helping much.

I think it might be nice to do something like:

```python
   rule = {
       'type': 'signal',
       'sender': 'org.freedesktop.DBus',
       ...
    }
    rule_text = ','.join(f"{key}='{val}'" for key, val in rule.items())
    self.matches.append(self.bus.add_match(rule_text, handler))
```

or so, and even to have a proper internal API for that.  IIRC, we need to do that kind of thing in response to match rules from the user, already.

You could even add a `**kwargs` on your internal API and then call it like

```python
    self.my_magic_matchy_thing(type='signal', sender='org.freedesktop.DBus', ...)
```

src/cockpit/channels/dbus.py
```diff
@@ -205,7 +240,19 @@ def do_open(self, options):
...
+            async def get_ready():
+                async with self.watch_processing_lock:
+                    await self.setup_name_owner_tracking()
+                    if self.owner:
+                        self.ready(unique_name=self.owner)
+                    else:
+                        self.close(problem="not-found")
+            task = asyncio.create_task(get_ready())
+            self.tasks.add(task)
+            task.add_done_callback(self.tasks.discard)
```
allisonkarlitskaya, line 253: ...at some point i almost start to prefer the explicit callback style...

It would be cool if Python had a syntax that let you use await from non-async functions with the meaning of "turn this into a task now, returning from the original function immediately"...

Short of that I think I might prefer if you handled the `self.name is None` case as an early-exit and unintended this code, providing a large amount of commenting about what's going on here.

allisonkarlitskaya (commented):
Two "...could this be nicer?" comments, but no hard blockers.  Let's talk about it in standup.

src/cockpit/channels/dbus.py
mvollmer: This is done better in #18076.

allisonkarlitskaya (approved):
My main concern is addressed in follow-up #18076, so let's land this.

PR #17733, opened 2022-09-13
containers: add mypy to the unit-test container

jelly:
Required for my bots PR https://github.com/cockpit-project/bots/pull/3735

Blocked on PR #17785

PR #17755, opened 2022-09-22
test-channelresponse: use pidfd to watch for child exit

allisonkarlitskaya:
Sending signals to old pids is unsafe (due to recycling) and ineffective (due to zombies).  The waitpid() approach messes too much with GLib's internals.

Let's avoid both of these and use a pidfd to do our watching in a way that's guaranteed to be race-free and effective, without mangling the process state in a way that confuses GLib.


allisonkarlitskaya:
. o O ( I wonder if `<sys/pidfd.h>` exists on RHEL 8.  Let's find out! )

martinpitt:
@TomasTomecek, @lachmanfrantisek: packit tests here failed with "Could not find copr build id '4868126' for chroot 'fedora-37-x86_64' ". However, the actual COPR builds *succeeded*, e.g. here: https://dashboard.packit.dev/results/copr-builds/470175. Infra problem or some API impedance mismatch? The same happened in e.g. https://github.com/cockpit-project/cockpit-machines/pull/809 , it's not specific to this PR.

martinpitt:
Meh, [doesn't work on RHEL 8](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-17755/centos-stream-8-x86_64/04868126-cockpit/builder-live.log.gz):
```
src/ws/test-channelresponse.c:541:24: error: 'SYS_pidfd_open' undeclared (first use in this function); did you mean 'SYS_fsopen'?
   int pid_fd = syscall(SYS_pidfd_open, pid, 0);
                        ^~~~~~~~~~~~~~
                        SYS_fsopen
src/ws/test-channelresponse.c:541:24: note: each undeclared identifier is reported only once for each function it appears in
```

But that sounds fixable? As long as the kernel provides it, we could do some existence test and declare it ourself? systemd does that all the time.

allisonkarlitskaya:
**\*hammer\*** **\*hammer\*** damn **\*hammer\*** ancient **\*hammer\*** libc **\*hammer\*** lemme just use the kernel headers directly

src/ws/test-channelresponse.c
```diff
@@ -525,19 +527,24 @@ test_resource_failure (TestResourceCase *tc,
...
+
+  /* The SIGTERM gets delivered to the bridge via a glib unix signal
+   * handler, and it is theoretically possible that the request that we
+   * send below could get delivered before the SIGTERM.  For that
+   * reason, we need to make sure that the process actually properly
+   * exited before sending the request.
+   */
+  struct pollfd pid_pfd = { .fd = pid_fd, .events = POLLIN };
+  while (poll (&pid_pfd, 1, -1) != 1)
+    ;
```
martinpitt, line 546: Could this do the `g_usleep (1000)` again to not spin in a tight loop (which would only make the process shut down more slowly)?

allisonkarlitskaya: This isn't spinning a tight loop.  `-1` as the last parameter to `poll()` means "sleep indefinitely".

martinpitt: Nevermind -- poll timeout does that.

allisonkarlitskaya: The loop is more about dealing with the (in this case extremely likely) `EINTR` case when the `SIGCHLD` gets delivered to us and restarting the call.  I could have made that more explicit, but this is also just a test...

martinpitt:
FTR: [copr build on c8s](https://copr.fedorainfracloud.org/coprs/packit/cockpit-project-cockpit-17755/build/4868147/) failed due to:
`__NR_pidfd_open' undeclared`. The [package list](https://download.copr.fedorainfracloud.org/results/packit/cockpit-project-cockpit-17755/centos-stream-8-x86_64/04868147-cockpit/root.log.gz) says kernel-headers-4.18.0-408.el8.x86_64 and glibc-devel-2.28-211.el8.x86_64


These are *exactly the same* versions that we have in our centos-8-stream image mock. That image is running that very kernel version, too.

martinpitt (approved):
[builds look good](https://copr.fedorainfracloud.org/coprs/packit/cockpit-project-cockpit-17755/build/4868230/)! phew, thank you so much for figuring this out!

martinpitt:
[This f36 failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17755-20220922-110217-f135cb5f-fedora-36-devel/log.html) is a bit suspicious, it smells like a race condition that makes the test hang. Unfortunately we disabled verbose mock build logs, as they are annoyingly verbose even in the success case. If this repeats, we may have to temporarily turn on verbosity in this PR?

martinpitt (approved):
:ram: :razor: Thanks!

allisonkarlitskaya:
> :ram: :razor: Thanks!

gern :hammer:

jelly:
Please re-trigger https://github.com/cockpit-project/cockpit/pull/17733 if this is merged if the github workflow rebases.

martinpitt:
I just [triggered a unit-tests refresh](https://github.com/cockpit-project/cockpit/actions/runs/3105709620). :crossed_fingers:

PR #17757, opened 2022-09-22
python: support ChannelError from AsyncChannel.run()

allisonkarlitskaya:
We add a function called run_wrapper() to catch ChannelError from the run() function and turn it into a close() message.

It would be more elegant to handle this by setting a done callback on the task and calling .result() from it in order to re-raise and catch the exception.  Unfortunately, there are some non-ChannelError exceptions being raised in our code right now, and they're currently being ignored, and this change would result in them being re-raised.  We can address that in a future PR.

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -169,6 +169,13 @@ class AsyncChannel(Channel):
     async def run(self, options):
         raise NotImplementedError
 
+    async def run_wrapper(self, options):
+        try:
+            await self.run(options)
+            self.close()
+        except ChannelError as exc:
+            self.close(**exc.kwargs)
+
     async def read(self):
         while not isinstance(item := await self.receive_queue.get(), bytes):
             self.send_pong(item)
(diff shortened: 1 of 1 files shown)


src/cockpit/channel.py
```diff
@@ -194,7 +201,7 @@ def do_pong(self, message):
     def do_open(self, options):
         self.receive_queue = asyncio.Queue()
         self.flow_control = options.get('flow-control') is True
-        asyncio.create_task(self.run(options), name='f{self.__class__.__name__}.run({options})')
+        asyncio.create_task(self.run(options), name='f{self.__class__.__name__}.run_wrapper({options})')
```
mvollmer, line 204: Also `self.run_wrapper(options)` in the first argument?

mvollmer (changes requested):

src/cockpit/channel.py
allisonkarlitskaya: oops :)

mvollmer: And maybe just `run` in the name?

mvollmer (approved):
Works great now!

PR #17758, opened 2022-09-22
python: Send expected 'ping' during flow control

mvollmer:
Otherwise we never get the 'pong' that will allow us to write more data.

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -165,7 +165,11 @@ class AsyncChannel(Channel):
     async def write(self, data):
         if self.flow_control:
             assert len(data) <= AsyncChannel.CHANNEL_FLOW_WINDOW
-            self.out_sequence += len(data)
+
+            out_sequence = self.out_sequence + len(data)
+            if self.out_sequence // AsyncChannel.CHANNEL_FLOW_PING != out_sequence // AsyncChannel.CHANNEL_FLOW_PING:
+                self.send_control(command='ping', sequence=out_sequence)
+            self.out_sequence = out_sequence
 
             while self.out_window < self.out_sequence:
                 self.write_waiter = asyncio.get_running_loop().create_future()
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -31,6 +31,7 @@ pyTESTS_PASSING = \
 	dist/base1/test-protocol.html \
 	dist/base1/test-series.html \
 	dist/base1/test-spawn.html \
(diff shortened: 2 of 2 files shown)


mvollmer:
This still doesn't use COCKPIT_FLOW_PING, though.  Are we maybe supposed to send a ping this many bytes before reaching the end of the current window?

mvollmer:
@martinpitt, does this help with your http channel woes?

src/cockpit/channel.py
```diff
@@ -168,6 +168,7 @@ async def write(self, data):
             self.out_sequence += len(data)
 
             while self.out_window < self.out_sequence:
+                self.send_control(command='ping', sequence=self.out_sequence)
```
allisonkarlitskaya, line 171: I think this isn't right.  Waiting until we're already going to block is going to result in unnecessary latency.

We rather want to send a ping, unconditionally, for every `n` bytes we send.

allisonkarlitskaya: the C bridge has these constants:

```C
/* Every 16K Send a ping */
#define  CHANNEL_FLOW_PING        (16L * 1024L)

/* Allow up to 1MB of data to be sent without ack */
#define  CHANNEL_FLOW_WINDOW       (2L * 1024L * 1024L)
```

and implements the following algorithm for deciding when to send:

```C
      /* Every CHANNEL_FLOW_PING bytes we send a ping; also when applying back
       * pressure as there is otherwise nothing more to send and generate pings for */
      if ((out_sequence / CHANNEL_FLOW_PING != priv->out_sequence / CHANNEL_FLOW_PING) || trigger_pressure)
        {
```

allisonkarlitskaya (changes requested):
Thanks for looking into this!!  This is very clearly "oops, I forgot" territory.

src/cockpit/channel.py
mvollmer: Alright!

mvollmer: I don't think we need something like `trigger_pressure`.

martinpitt:
@mvollmer : Sorry, which woes? http works fine for me. In test-http.js I only disable features which we chose to not implement.

mvollmer:
> @mvollmer : Sorry, which woes? 

I heard you were running into races or dead locks or something.  Never mind if I misunderstood.

allisonkarlitskaya (approved):
:heart:

PR #17761, opened 2022-09-22
metrics: ignore flatpak revokefs-fuse

jelly:
I believe these mounts are for flatpak portals and are not interesting to show in our metrics page.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -307,6 +307,9 @@ class CurrentMetrics extends React.Component {
                      * which are always at 100% capacity, but are uninteresting for disk usage alerts */
                     if ((fields[0].indexOf("/loop") >= 0 && options.indexOf('ro') >= 0))
                         result.add(fields[1]);
+                    /* hide flatpaks */
+                    if ((fields[0].indexOf('revokefs-fuse') >= 0 && fields[1].indexOf('flatpak') >= 0))
+                        result.add(fields[1]);
                 });
         return result;
     }


pkg/metrics/metrics.jsx
```diff
@@ -307,6 +307,9 @@ class CurrentMetrics extends React.Component {
                      * which are always at 100% capacity, but are uninteresting for disk usage alerts */
                     if ((fields[0].indexOf("/loop") >= 0 && options.indexOf('ro') >= 0))
                         result.add(fields[1]);
+                    /* hide flatpaks */
+                    if ((fields[0].indexOf('revokefs-fuse') >= 0 && fields[1].indexOf('flatpak') >= 0))
+                        result.add(fields[1]);
```
cockpituous, line 312: These 2 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17761-20220922-163244-044cdcae-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt: Can you please add a check for this to testDiskIO, similar to the "does not show read-only devices"?
```py
        # read-only loop devices are not shown
        self.assertFalse(b.is_present(".pf-c-progress[data-disk-usage-target='/var/cockpit-ro-test']"))
```

i.e. mount something which looks like a flatpak.

jelly: This would require flatpak to be installed on the test image, as revokefs-fuse is part of the flatpak package. So we would either lax the check to include "fuse" or skip it.

martinpitt: Oh, because that "revokefs-fuse" file system isn't available on our VMs, I see. I do have flatpak installed, and even running (Signal app), and even my system doesn't have that file system.

martinpitt (approved):
Thanks! Re-running infra-failed tests

pkg/metrics/metrics.jsx
jelly: I think it's due to some portal handling which fails or isn't cleaned up, which makes me reconsider how useful this change is...

jelly: Altho.. I have it now, seems updating stuff also causes these files

image

PR #17762, opened 2022-09-23
tools: Recommend cockpit-ostree on OSTree systems

martinpitt:
Similarly how we recommend cockpit-packagekit if dnf is installed, recommend cockpit-ostree if rpm-ostree is installed. That will make the package more discoverable, and provide an OOTB "Software Updates" page.

cockpit-ostree is not in RHEL 8/9 at the moment, so only generate this
dependency on non-RHEL (i.e. Fedora) for the time being.

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -147,6 +147,8 @@ Suggests: cockpit-pcp
 
 %if 0%{?rhel} == 0
 Recommends: (cockpit-networkmanager if NetworkManager)
+# c-ostree is not in RHEL 8/9
+Recommends: (cockpit-ostree if rpm-ostree)
 Suggests: cockpit-selinux
 %endif
 %if 0%{?rhel} && 0%{?centos} == 0


marusak (dismissed):
Thank you! This is exactly what I had in mind when I talked with Brian but was not entirely sure on what to depend

tools/cockpit.spec
```diff
@@ -143,6 +143,7 @@ Requires: cockpit-system
 # Optional components
 Recommends: (cockpit-storaged if udisks2)
 Recommends: (cockpit-packagekit if dnf)
+Recommends: (cockpit-ostree if rpm-ostree)
```
martinpitt, line 146: Hmm, wait -- as long as the package is not in RHEL, I think we should move this to the section below. We can move it back if/when it gets into RHEL.

marusak: right, but should not hurt, right? Surely dnf can deal with recommends that do not exists in any repo.

martinpitt: well, https://bugzilla.redhat.com/show_bug.cgi?id=2093231 …

martinpitt:
@marusak : PTAL?

marusak (approved):
Thanks!

PR #17763, opened 2022-09-23
build: rip out automated version history

allisonkarlitskaya:
Automatically checking all of the tags in order to construct our version history has always been a bit gross.  It means that what lands in the release tarball is determined by more than just the current commit.

On top of that, the script to do this nasty bit of work just produced an output which failed to validate, owing to the recent out-of-sequence release of 264.2.

Version history really ought to be tracked downstream, so let's rip all of this out and replace it with a very simple approach to writing a single release line into the file.  This is the minimum required for the file to validate.  Any additional history can be handled downstream (manually, for the time being).


martinpitt (approved):
Danke!

PR #17764, opened 2022-09-23
[rhel-9-0] Update rhel 9.0.0 branch

marusak:


martinpitt (approved):
Looks good, thank you! Note that our CI cannot handle this any more. You can test the RHEL downstream update from your branch, and if that works, land it.

test/verify/check-pages
```diff
@@ -27,6 +27,7 @@ from machine_core.constants import TEST_OS_DEFAULT
 
 @nondestructive
 @skipDistroPackage()
+@timeout(1200)
```
martinpitt, line 30: This should be a separate commit, and also be applied to main.

martinpitt (changes requested):

test/verify/check-pages
martinpitt: actually, it doesn't need to go to main -- in FMF we don't run TestPages at all, this is only for the old STI tests in RHEL 8.

martinpitt (approved):
Dakujem!

PR #17765, opened 2022-09-23
build: add `pcc` (Python Cockpit Client) rule

allisonkarlitskaya:
`make pcc` will attempt to install a symlink to cockpit-bridge.pyz in ~/.local/bin/cockpit-bridge and launch Cockpit Client on it.


martinpitt (approved):
Haha - indeed let's better not document this in README, this smells like a "lis testing target". Alternatively, we could make client read a `$COCKPIT_BRIDGE` env var, so that this works without permanent changes to the file system?

I just wonder what you'll do about the pages themselves? Should that grow some `XDG_DATA_DIRS=...`?

Anyway, fine for me if it helps you.

PR #17766, opened 2022-09-23
python: use dir_fd for fetching cgroups

jelly:
This speeds up the fetching of samples in cgroups by doing less syscalls.

# Test program

```
#!/bin/python

import collections
import sys
import setproctitle

sys.path.append("./src/cockpit")

from samples import SAMPLERS

setproctitle.setproctitle("sampling1")

for i in range(0, 1):
    for sample in SAMPLERS:
        data = collections.defaultdict(dict)
        sample().sample(data)
        # print(data)
```

# Testing syscalls

```
sudo bpftrace -f json -e 'tracepoint:raw_syscalls:sys_enter { @[comm] = count(); } interval:s:5 { print(@); clear(@); }'
```

Or using setproctitle:

```
sudo bpftrace -f json -e 'tracepoint:raw_syscalls:sys_enter /comm == "sampling1"/ { @[comm] = count(); } interval:s:5 { print(@); clear(@); }'
```

comm is the process name, this is 'sampling`' in the test program.

The change is 8285 => 4252

# Profiling with hyperfine

Change the `range` to `range(0, 100)`.

Old:
```
[jelle@t14s][~/projects/cockpit/metrics_improvements]%hyperfine --warmup 3 'python3 bench_sampler.py'
Benchmark 1: python3 bench_sampler.py
  Time (mean ± σ):      1.620 s ±  0.041 s    [User: 1.021 s, System: 0.586 s]
  Range (min … max):    1.551 s …  1.679 s    10 runs
```

New:
```
[jelle@t14s][~/projects/cockpit/metrics_improvements]%hyperfine --warmup 3 'python3 bench_sampler.py'
Benchmark 1: python3 bench_sampler.py
  Time (mean ± σ):     919.4 ms ±  30.2 ms    [User: 558.7 ms, System: 348.7 ms]
  Range (min … max):   854.9 ms … 964.6 ms    10 runs
```


src/cockpit/samples.py
```diff
@@ -25,6 +26,15 @@
...
 
+@contextlib.contextmanager
+def os_open(*args, **kwargs):
+    fd = os.open(*args, **kwargs)
+    try:
+        yield fd
+    finally:
+        os.close(fd)
+
+
```
allisonkarlitskaya, line 37: In my experience, generators can be a bit slow.

The only thing you ever do with this is to call os.read() on it, so why not just make the function return the bytes?

src/cockpit/samples.py
```diff
@@ -192,23 +202,25 @@ class CGroupSampler(Sampler):
     cgroups_v2: Optional[bool] = None
     cgroups_v2_path = '/sys/fs/cgroup/'
 
-    def read_cgroup_keyed_stat(self, samples, path, cgroup, name, key):
-        with open(path) as stat:
-            for line in stat:
+    def read_cgroup_keyed_stat(self, samples, rootfd, statfile, cgroup, name, key):
+        with os_open(statfile, os.O_RDONLY, dir_fd=rootfd) as fd:
+            stat = os.read(fd, 1024)
+            for line in stat.decode().splitlines():
```
allisonkarlitskaya, line 208: Why decode?  I think this is actually kinda bogus.  The kernel interface is definitely expressed in locale-independent bytes.  You should rather change all of the key values to bytestrings, and then just work in bytes.

allisonkarlitskaya: Also: I wonder if we could do better than the splitlines approach.  Maybe directly scan for the key, using `.index()`?

src/cockpit/samples.py
```diff
@@ -223,18 +235,18 @@ def sample(self, samples):
...
                 if not cgroup:
                     continue
 
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.current'), cgroup, 'cgroup.memory.usage')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.max'), cgroup, 'cgroup.memory.limit')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.swap.current'), cgroup, 'cgroup.memory.sw-usage')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.swap.max'), cgroup, 'cgroup.memory.sw-limit')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'cpu.weight'), cgroup, 'cgroup.cpu.shares')
-                self.read_cgroup_keyed_stat(samples, os.path.join(path, 'cpu.stat'), cgroup, 'cgroup.cpu.usage', 'usage_usec')
+                self.read_cgroup_integer_stat(samples, rootfd, 'memory.current', cgroup, 'cgroup.memory.usage')
```
allisonkarlitskaya, line 244: I wonder why not something like

```python
samples['cgroup.memory.usage'][cgroup] = read_cgroup_int(rootfd, 'memory.current')
samples['cgroup.cpu.usage'][cgroup] = read_cgroup_keyed(rootfd, 'cpu.stat', 'usage_usec')
...etc
```

with `read_cgroup_int()` and `read_cgroup_keyed()` being a single functions that open, read, (key search), converts-to-int and returns the result

I wouldn't worry too much about the case of an out-of-bounds value... This is the kernel.  If that happens you can just write in a 0 or just let `int()` throw.

allisonkarlitskaya (changes requested):
MOAR!!

src/cockpit/samples.py
allisonkarlitskaya: Try to think in terms of memory allocations to get a rough idea of what takes time and not — converting bytes to a unicode object is a memory allocation.  Splitting a file into multiple lines is a memory allocation for each line, plus at least one for the array.  Also: working on unicode objects is potentially slower than working on raw bytes.

jelly: Yeah, the decode() was a quick hack from late Friday afternoon. I'll keep it at bytes. Seems to be slightly faster:

Old
```
In [41]: !hyperfine --warmup 3 'python3 bench_sampler.py'
Benchmark 1: python3 bench_sampler.py
  Time (mean ± σ):      2.407 s ±  0.440 s    [User: 1.470 s, System: 0.906 s]
  Range (min … max):    1.973 s …  3.257 s    10 runs
```

New
```
In [44]: !hyperfine --warmup 3 'python3 bench_sampler.py'
Benchmark 1: python3 bench_sampler.py
  Time (mean ± σ):      2.095 s ±  0.128 s    [User: 1.258 s, System: 0.809 s]
  Range (min … max):    1.964 s …  2.331 s    10 runs
```

src/cockpit/samples.py
jelly: Well, the generator does a bunch more, we catch file found exceptions and value errors and it is the most pythonic way to write it I'd say.

We could measure it for sure.

src/cockpit/samples.py
jelly: Uh, I don't think we want to assign 0 when the file is not found or can't be read, the C code just errors out and doesn't set a sample so I think we want to replicate that behaviour

allisonkarlitskaya: None?

jelly:
Updated benchmarks: https://gist.github.com/jelly/67dfc8aecff41425cb065524f41b95a3

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
-                    value = int(stat.read().strip())
-                except ValueError:
-                    return
-
-                if sys.maxsize > value > 0:
-                    samples[name][cgroup] = value
-        except FileNotFoundError:
-            pass
+            fd = os.open(statfile, os.O_RDONLY, dir_fd=rootfd)
+        except Exception:
```
allisonkarlitskaya, line 200: I think you want `FileNotFoundError`, no?

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
+            return
+
+        data = os.read(fd, 1024)
```
allisonkarlitskaya, line 203: If this fails for whatever reason, you'll leak `fd`.  Please put the `close()` in a `finally:` block.

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
+            # cpu.stat are in usecs
+            value = int(data[start:end]) / 1000
+        else:
+            # Some samples such as "memory.max" contains "max" when there is a no limit
+            try:
+                value = int(data)
+            except ValueError:
+                return
+
+        if value > sys.maxsize:
```
allisonkarlitskaya, line 218: Have you ever seen this happen?  It seems like a strange thing to check for, to be honest.

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
+            start = data.index(key) + len(key)
+            end = data.index(b'\n', start)
+            # cpu.stat are in usecs
+            value = int(data[start:end]) / 1000
+        else:
+            # Some samples such as "memory.max" contains "max" when there is a no limit
+            try:
+                value = int(data)
+            except ValueError:
+                return
```
allisonkarlitskaya, line 216: Please explicitly return `None` if that's your intent.  Mixing non-value returns with value returns is uncool.

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
+
+        if value > sys.maxsize:
+            return
+
+        if value > 0:
+            return value
+        elif include_zero:
```
allisonkarlitskaya, line 223: It makes sense to combine this into a single `if include_zero or value > 0:`

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
+            return value
```
allisonkarlitskaya, line 224: Please explicitly return `None` here.

src/cockpit/samples.py
allisonkarlitskaya: The `include_zero` thing seems hacky, btw.

src/cockpit/samples.py
```diff
@@ -192,49 +192,55 @@ class CGroupSampler(Sampler):
...
                 if not cgroup:
                     continue
 
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.current'), cgroup, 'cgroup.memory.usage')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.max'), cgroup, 'cgroup.memory.limit')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.swap.current'), cgroup, 'cgroup.memory.sw-usage')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'memory.swap.max'), cgroup, 'cgroup.memory.sw-limit')
-                self.read_cgroup_integer_stat(samples, os.path.join(path, 'cpu.weight'), cgroup, 'cgroup.cpu.shares')
-                self.read_cgroup_keyed_stat(samples, os.path.join(path, 'cpu.stat'), cgroup, 'cgroup.cpu.usage', 'usage_usec')
+                samples['cgroup.memory.usage'][cgroup] = self.read_cgroup_integer_stat(rootfd, 'memory.current', True)
```
allisonkarlitskaya, line 238: This reads a whole lot nicer.  But what do you do with all of those `None` values?

allisonkarlitskaya (changes requested):

src/cockpit/samples.py
jelly: Hmm ideally the channel would return False, however everything now works by coincidence more or less...

We request:
```
      { name: "cgroup.cpu.usage", derive: "rate" },
      { name: "cgroup.memory.usage" },
```

As it's an `instanced` metric and a `rate` we will need an `old_value` and `value`. As None is not True the metrics won't be send.

src/cockpit/samples.py
jelly: Yes, it's a bit iffy

src/cockpit/samples.py
jelly: Sure, but that doesn't fix the potential exception right? Although that seems pretty unlikely.

src/cockpit/samples.py
```diff
@@ -192,49 +192,54 @@ class CGroupSampler(Sampler):
...
+        try:
+            data = os.read(fd, 1024)
+        finally:
+            os.close(fd)
+
+        if key:
+            start = data.index(key) + len(key)
+            end = data.index(b'\n', start)
+            # cpu.stat are in usecs
+            value = int(data[start:end]) / 1000
```
allisonkarlitskaya, line 212: I don't like how this `/ 1000` is hardcoded.  It makes the `key` path function fundamentally differently from the unkeyed path, in an unexpected way.  Please move the `/ 1000` to the call site before assigning the value to the dictionary.  Unless: is there a chance that the result might be None?  I wonder if it might make sense here to do NaN.

allisonkarlitskaya (approved):
Honestly, it looks pretty OK now.  There are some rough edges that I'd like you to consider smoothing out, but it could also go in the way it is... It's an improvement over what was there, in any case.

src/cockpit/samples.py
jelly: The result can be None for sure, so I'll break it up and add a check.

allisonkarlitskaya (approved):
good enough :)

PR #17770, opened 2022-09-26
metric1 unit tests

jelly:
After this PR, the cockpit-bridge and Python bridge should be somewhat consistent in error handling

### DIFF
--- a/pkg/base1/test-metrics.js
+++ b/pkg/base1/test-metrics.js
@@ -104,7 +104,17 @@ function MockSink(expected, callback) {
     return self;
 }
 
-QUnit.test("non-instanced decompression", function (assert) {
+let old_channel;
+QUnit.module("Mock tests", {
+    beforeEach: () => {
+        old_channel = cockpit.channel;
+    },
+    afterEach: () => {
+        cockpit.channel = old_channel;
+    }
+});
+
+QUnit.test("non-instanced decompression", assert => {
     assert.expect(1);
 
     const peer = new MockPeer();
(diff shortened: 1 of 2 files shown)


pkg/base1/test-metrics.js
```diff
@@ -104,6 +104,16 @@ function MockSink(expected, callback) {
     return self;
 }
 
+let old_channel;
+QUnit.module("Mock tests", {
+    beforeEach: () => {
+        old_channel = cockpit.channel;
+    },
+    afterEach: () => {
+        cockpit.channel = old_channel;
```
martinpitt, line 113: Why is that necessary? AFAIK it does not get diverted anywhere?

pkg/base1/test-metrics.js
```diff
@@ -127,4 +137,60 @@ QUnit.test("non-instanced decompression", function (assert) {
     assert.deepEqual(sink.samples, [[10], [10]], "got correct samples");
 });
 
+QUnit.test("interval validation", function (assert) {
```
martinpitt, line 140: I'd rather we'd used ES6 `assert => {... }` syntax, but not a biggie.

pkg/base1/test-metrics.js
```diff
@@ -127,4 +137,60 @@ QUnit.test("non-instanced decompression", function (assert) {
...
+    const metrics_channel = cockpit.channel({
+        payload: "metrics1",
+        source: "internal",
+        interval: -1,
+    });
+
+    metrics_channel.addEventListener("close", (_, error) => {
+        assert.equal(error.problem, "protocol-error");
+        assert.equal(error.message, 'invalid "interval" value: -1');
+        assert.ok(true, "Connection was closed for invalid interval");
```
martinpitt, line 152: This assertion feels superfluous? We only need these if there's nothing more specific to assert, no?

martinpitt (changes requested):
Thanks! Some cleanups (perhaps), but no strong "no". In case you push again, please add "no-test", as these unit tests don't run in the integration tests anyway.

pkg/base1/test-metrics.js
jelly: It does in mockChannel.

pkg/base1/test-metrics.js
jelly: Yeah, can be dropped.

pkg/base1/test-metrics.js
jelly: Can do, as I'm pushing anyway.

pkg/base1/test-metrics.js
martinpitt: Oh, you mean an existing test damages this, and for the new ones we want to use the real cockpit.channel.

martinpitt (approved):
Danke!

jelly:
```
# Result testBasic (__main__.TestTeam.testBasic) failed
# 1 TEST FAILED [81s on ip-172-31-21-69]
```

Seems this one fails in all PR's

martinpitt:
The failing test is `TestStorageUsed.testUsedAsPV`, that's obviously independent from this PR, but we need to investigate. It's very likely to come up at our next fedora-37 image refresh, too.

Issue #17669, opened 2022-08-24
Permissions check for Firewall rely on `pkcheck`; fall back to root check


ancorgs:
### Explain what happens

If the firewalld configuration is considered to be read-only, then the interface does not allow the authenticated user to start/stop the Firewall or to modify the zones. That's good, but I think the check for read-only needs to be improved.

It's done by executing the command `pkcheck --action-id org.fedoraproject.FirewallD1.all --process $$ --allow-user-interaction` (see [here](https://github.com/cockpit-project/cockpit/blob/c606641c83bffb49c0ac68c8d8f25d34e01e9786/pkg/networkmanager/firewall-client.js#L333)), which needs `polkit` to be installed. So no PolicyKit -> no `pkcheck` -> no firewall configuration.

In a system in which `polkit` is not installed, it should still be possible to manage the firewall if you log into Cockpit as root. I did a test in such a system just hacking Cockpit to remove the `pkcheck` call and it certainly worked. So PolicyKit is not really a requirement to manage Firewalld from Cockpit (at least with a privileged user).

I just saw https://github.com/cockpit-project/cockpit/issues/11033 which states the check done by the Firewall page should be adopted everywhere. Please, improve the Firewall check first. Adopting it everywhere in its current form would make Cockpit totally dependent on `polkit`.

### Version of Cockpit

271

### Where is the problem in Cockpit?

Firewall

### Server operating system

openSUSE

### Server operating system version

ALP

### What browsers are you using?

Firefox

PR #17776, opened 2022-09-27
networkmanager: Improved firewall permission check

lslezak:
## The Problem

See issue https://github.com/cockpit-project/cockpit/issues/17669.

The firewall configuration uses the `pkcheck` tool for checking the user permissions. But that tool might not be available in a minimal system.

## The Solution

If the `pkgcheck` test fails fallback to the usual root permission check.

## Testing

Tested manually with the patched plugin.

## Screenshots

Before applying the patch the firewall configuration could not be changed:

cockpit_fw_no_permission

After applying the patch the switch is enabled and changing the value works correctly.

cockpit_fw_fixed

### DIFF
--- a/pkg/networkmanager/firewall-client.js
+++ b/pkg/networkmanager/firewall-client.js
@@ -335,6 +335,19 @@ cockpit.spawn(['sh', '-c', 'pkcheck --action-id org.fedoraproject.FirewallD1.all
             firewall.readonly = false;
             firewall.debouncedEvent('changed');
             firewall.debouncedGetZones();
+        })
+        .catch(error => {
+            console.log("pkcheck failed", error);
+
+            // Fall back to cockpit.permissions, "pkcheck" might not be available,
+            // always allow edits by admins
+            const permission = cockpit.permission({ admin: true });
+            const update_permissions = () => {
+                firewall.readonly = !permission.allowed;
+                firewall.debouncedEvent('changed');
+                firewall.debouncedGetZones();
+            };
+            permission.addEventListener("changed", update_permissions);
         });
 
 firewall.enable = () => firewalld_service.enable().then(() => firewalld_service.start());
--- a/test/verify/check-networkmanager-firewall
+++ b/test/verify/check-networkmanager-firewall
@@ -122,6 +122,27 @@ class TestFirewall(NetworkCase):
             ".*The name org.fedoraproject.FirewallD1 was not provided by any .service files.*",
             ".*org.fedoraproject.FirewallD1: .*: GDBus.Error:org.freedesktop.DBus.Error.NoReply.*")
 
+    # test missing "pkcheck" binary, in that case fallback to the admin check
+    def testPkcheckMissing(self):
+        b = self.browser
+        m = self.machine
+
+        # Path to the "pkcheck" PolicyKit utility
+        pkcheck = m.execute("which pkcheck").strip()
+        # "Hide" pkcheck
+        m.execute(f"mount --bind /dev/null '{pkcheck}'")
+
+        try:
+            # Regular user is not allowed to change the firewall switch
+            self.login_and_go("/network", superuser=False)
+            b.wait_visible("#networking-firewall-switch:disabled")
+
+            # Super user is allowed to change the firewall switch
+            b.relogin("/network", superuser=True)
+            b.wait_visible("#networking-firewall-switch:not([disabled])")
+        finally:
+            m.execute(f"umount '{pkcheck}'")
+
     def testFirewallPage(self):
         b = self.browser
         m = self.machine


pkg/networkmanager/firewall-client.js
```diff
@@ -335,6 +335,22 @@ cockpit.spawn(['sh', '-c', 'pkcheck --action-id org.fedoraproject.FirewallD1.all
             firewall.readonly = false;
             firewall.debouncedEvent('changed');
             firewall.debouncedGetZones();
+        })
+        .catch(error => {
+            console.warn("pkcheck failed", error);
```
martinpitt, line 340: This is too broad -- the command also fails if `pkcheck` is installed, but the user does not have the permission. This needs to inspect the error and only fall back on "command not found". Also, this should not be a `warn`, I think.

pkg/networkmanager/firewall-client.js
```diff
@@ -335,6 +335,22 @@ cockpit.spawn(['sh', '-c', 'pkcheck --action-id org.fedoraproject.FirewallD1.all
...
+            // Fall back to cockpit.permissions, "pkcheck" might not be available
+            const permission = cockpit.permission({ admin: true });
+            const update_permissions = () => {
+                console.log("permissions.allowed: ", permission.allowed);
+
+                if (permission.allowed) {
+                    firewall.readonly = false;
+                    firewall.debouncedEvent('changed');
+                    firewall.debouncedGetZones();
+                }
```
martinpitt, line 351: The events should also be sent if the permission is not allowed, i.e. when dropping admin privileges again.  I. e. set `firewall.readonly = !permission.allowed`, and drop the `if`?

pkg/networkmanager/firewall-client.js
lslezak: That's actually a question how it should behave. It depends on priorities. You do not have the PolicyKit permission but you are admin (`root`). What should have a higher priority?

I was already thinking about this and decided to use the same analogy as with the file system permissions. If you remove all permissions from a file (`chmod a-rw`) you can still read (`cat`) or write (`echo >>`) to the file if you are `root`.

On the other hand I'm not a PolicyKit expert and I do not know how this problem is solved elsewhere, it should be consistent...

martinpitt: root has all the polkit permissions anyway, at least for default rules that are shipped by projects. I was mostly worried about doing (essentially) the same check twice, as this further slows down the page for non-admins. But you are right about the consistency argument. OK, then let's please just tone it down to console.log(), and see how the PR turns out otherwise. Thanks!

garrett:
Great to see the check improved. I think the UI as seen in this screenshot can be improved as well:

screenshot when you're not an admin

Perhaps something like removing the tooltip and toggle and put the tooltip's content inline there instead, perhaps with a button to become admin?

...But this should be solved in a separate follow-up PR. 

I like what I see here and am looking forward to this PR landing!

lslezak:
I have added an integration test for this. The test hides the `pkcheck` via bind mount as suggested.

However, there is one important thing to discuss for the future. That bind mount might potentially cause a race condition if another tests use it as well and if those tests run in parallel. So for there is no such potentially conflicting test. But the question is: is it OK? Could we do that somehow better? @martinpitt ?

lslezak:
Just a friendly ping. Any feedback about this change? How should we proceed?
Thank you!

martinpitt:
@lslezak Sorry for the delay. We are currently a bit understaffed in the cockpit team, and I'm myself working in a different team this quarter. I haven't forgotten about this, but it will still take some time to get to this. I'll try to give this some love next week.

lslezak:
OK, np.
Thank you for the info!

test/verify/check-networkmanager-firewall
```diff
@@ -127,6 +127,32 @@ class TestFirewall(NetworkCase):
             ".*The name org.fedoraproject.FirewallD1 was not provided by any .service files.*",
             ".*org.fedoraproject.FirewallD1: .*: GDBus.Error:org.freedesktop.DBus.Error.NoReply.*")
 
+    # test missing "pkcheck" binary, in that case fallback to the admin check
+    def testPkcheckMissing(self):
+        b = self.browser
+        m = self.machine
+
+        # Path to the "pkcheck" PolicyKit utility
+        pkcheck = m.execute("which pkcheck").rstrip("\n")
```
martinpitt, line 136: Just `.strip()` will DTRT.

martinpitt (changes requested):
This looks really nice, thank you and sorry for dragging this out for so long! I'll do a rebase with a few small cleanups, and then move this to the finish line.

test/verify/check-networkmanager-firewall
```diff
@@ -127,6 +127,32 @@ class TestFirewall(NetworkCase):
...
+
+        if pkcheck:
```
martinpitt, line 138: Let's drop this. First, it's a no-op -- `which` will fail if pkcheck is not installed. Second, `if` in tests is a code smell in general, tests should be predictable. I'll try a round of tests without the `if`, and see where it's missing (possibly on Fedora CoreOS?)

martinpitt:
Huh, now `test/verify/check-users TestAccounts.testBasic` fails on *all* testing-farm (packit) tests. This PR does not touch the accounts page at all, and firewall and users tests even run in independent testbeds. We just got a new Testing Farm release rolled out, so this is 99% unrelated. I will wait for #17952 , I fully expect tests to fail in the same way there.

Meanwhile, our own CI tests look good so far (they still take a while)

martinpitt (approved):

PR #17782, opened 2022-09-28
python: add CGroups v1 support

jelly:
For RHEL 8 we still need to support CGroups v1, this was tested seperate of the cockpit-bridge as the bridge currently does not run on Python 3.6 due to the usage of walrus operators.

### DIFF
--- a/src/cockpit/samples.py
+++ b/src/cockpit/samples.py
@@ -189,7 +189,6 @@ class CGroupSampler(Sampler):
     ]
 
     cgroups_v2: Optional[bool] = None
-    cgroups_v2_path = '/sys/fs/cgroup/'
 
     @staticmethod
     def read_cgroup_integer_stat(rootfd, statfile, include_zero=False, key=b''):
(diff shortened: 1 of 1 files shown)


martinpitt (commented):
Nice, thank you! :+1:  I'm glad it wasn't that bad after all. But I'll leave the final review to @allisonkarlitskaya

src/cockpit/samples.py
```diff
@@ -241,6 +240,30 @@ def sample(self, samples):
...
+                if not cgroup:
+                    continue
+
+                samples['cgroup.memory.usage'][cgroup] = self.read_cgroup_integer_stat(rootfd, 'memory.usage_in_bytes', True)
+                samples['cgroup.memory.limit'][cgroup] = self.read_cgroup_integer_stat(rootfd, 'memory.limit_in_bytes')
+                samples['cgroup.memory.sw-usage'][cgroup] = self.read_cgroup_integer_stat(rootfd, 'memory.memsw.usage_in_bytes', True)
+                samples['cgroup.memory.sw-limit'][cgroup] = self.read_cgroup_integer_stat(rootfd, 'memory.memsw.limit_in_bytes')
+
+            cpu_path = '/sys/fs/cgroup/cpu/'
+            for path, _, _, rootfd in os.fwalk(cpu_path):
```
allisonkarlitskaya, line 257: Fun fact: `fwalk()` itself has a `dir_fd` parameter.

Less fun fact: the pathnames that it returns are prefixed by `./`, so you'd need to trim them anyway :(

allisonkarlitskaya (approved):

src/cockpit/samples.py
jelly: Interesting, that would also mean wrapping it in a `try`/`finally` to handle the `os.close`.

PR #17784, opened 2022-09-30
containers: Update unit-tests README to current reality

martinpitt:
The invocation of the start script has changed a lot in commit b746f5aef3671.

---

This is primarily meant to investigate the current [unit-test refresh failure](https://github.com/cockpit-project/cockpit/actions/runs/3149441424/jobs/5121063633). That only runs the tests serially, but I want to know what other combinations of clang/i386/valgrind fail.

### DIFF
--- a/containers/unit-tests/README.md
+++ b/containers/unit-tests/README.md
@@ -30,40 +30,39 @@ Tests in that container get started with the `start` script.  By default, this
 script runs the unit tests on amd64.  The script accepts a number of arguments
 to modify its behaviour:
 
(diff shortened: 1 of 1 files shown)


martinpitt:
All unit tests succeeded here, despite unit-tests.yml and unit-tests-refresh.yml running on the same OS (ubuntu-22.04) with the same backend (podman). **However**, the unit-test-refresh.yml ./start invocation is missing `--env=CFLAGS='-O2 -gdwarf-4'`, which explains the clang failure!

I can reproduce that locally as well. I'll close this PR and re-propose it from origin, so that I can run a refresh against it.

PR #17798, opened 2022-10-10
test: Run partprobe on new loop devices

mvollmer:
This should get rid of non-destructive flakes like this: https://cockpit-logs.us-east-1.linodeobjects.com/pull-3947-20221007-232750-12dac395-centos-8-stream-cockpit-project-cockpit/log.html#231-2


marusak (approved):
Thank you!!

PR #17799, opened 2022-10-10
test: Run TestStorageLuks.testLuks1Slots also on Arch

mvollmer:
It doesn't depend on clevis/tang so it should be fine on Arch.


jelly (approved):
Nice, and it passes!

PR #17803, opened 2022-10-10
python: add options control command

jelly:
The Cockpit terminal uses options to set the rows and cols.

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -70,6 +70,8 @@ class Channel(Endpoint):
             self.do_ping(message)
         elif command == 'pong':
             self.do_pong(message)
+        elif command == 'options':
+            self.do_options(message)
 
     def do_channel_control(self, command, message):
         # Catch errors and turn them into close messages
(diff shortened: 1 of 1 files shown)


src/cockpit/channel.py
```diff
@@ -95,6 +97,9 @@ def do_close(self):
     def do_pong(self, message):
         pass
 
+    def do_options(self, message):
+        pass
```
allisonkarlitskaya, line 101: how about `raise ChannelError("this channel don't do options")`?

jelly: Sounds fine!

src/cockpit/channel.py
```diff
@@ -95,6 +97,9 @@ def do_close(self):
     def do_pong(self, message):
         pass
 
+    def do_options(self, message):
+        raise ChannelError('This channel does not implement "options"')
```
allisonkarlitskaya, line 101: It needs a problem code as the first field.  Probably `protocol-error`?  What does the C bridge do here, anyway?

jelly: Not, `not-supported`?

allisonkarlitskaya (approved):

PR #17805, opened 2022-10-11
test: drop Debian networkmanager hack

jelly:
After adding Arch Linux testing support for checkpoints, I discovered the workaround is not required anymore for Debian as well. This was supposedly fixed in NetworkManager master in 2016.


mvollmer (approved):
Thanks!

PR #17808, opened 2022-10-11
python: support 'disk.dev.{read,written}' metrics

jelly:
This brings the Python bridge internal metrics support up to date with the C bridge.

### DIFF
--- a/src/cockpit/samples.py
+++ b/src/cockpit/samples.py
@@ -145,12 +145,14 @@ class DiskSampler(Sampler):
     descriptions = [
         SampleDescription('disk.all.read', 'bytes', 'counter', False),
         SampleDescription('disk.all.written', 'bytes', 'counter', False),
+        SampleDescription('disk.dev.read', 'bytes', 'counter', True),
+        SampleDescription('disk.dev.written', 'bytes', 'counter', True),
     ]
 
     def sample(self, samples: Dict[str, Any]):
         with open('/proc/diskstats') as diskstats:
-            bytes_read = 0
-            bytes_written = 0
+            all_read_bytes = 0
+            all_written_bytes = 0
             num_ops = 0
 
             for line in diskstats:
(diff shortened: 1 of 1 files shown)


src/cockpit/samples.py
```diff
@@ -145,6 +145,8 @@ class DiskSampler(Sampler):
     descriptions = [
         SampleDescription('disk.all.read', 'bytes', 'counter', False),
         SampleDescription('disk.all.written', 'bytes', 'counter', False),
+        SampleDescription('disk.dev.read', 'bytes', 'counter', True),
```
allisonkarlitskaya, line 148: Ya.  This is awesome.

src/cockpit/samples.py
```diff
@@ -169,10 +171,16 @@ def sample(self, samples: Dict[str, Any]):
                 if dev_name.startswith('nvme') and 'p' in dev_name:
                     continue
 
-                bytes_read += int(num_sectors_read) * 512
-                bytes_written += int(num_sectors_written) * 512
+                read_bytes = int(num_sectors_read) * 512
+                written_bytes = int(num_sectors_written) * 512
+
+                bytes_read += read_bytes
+                bytes_written += written_bytes
```
allisonkarlitskaya, line 178: I don't like how you're avoiding the naming conflict by basically changing the order and tense of the words.

I'd take the opportunity to change the name here to something more like `all_read_bytes` and `all_written_bytes`, and then use the `read_bytes` form for the per-instance value.

allisonkarlitskaya (changes requested):

src/cockpit/samples.py
jelly: Ah thanks for the suggestion! Yeah I didn't want to duplicate `(int(num_sectors_read) * 512)`.

I also wonder if this is wrong btw, we always multiple with 512, but are sectors always this size?

https://developer.ibm.com/tutorials/l-linux-on-4kb-sector-disks/ For example mentions 4kb sizes:

```
[jelle@t14s][~/projects/cockpit/main]%cat /sys/block/nvme0n1/queue/hw_sector_size
512
```

Now my ssd can't be formatted with 4 kb bock sizes, but still an interesting question.

allisonkarlitskaya (approved):

PR #17810, opened 2022-10-11
test: drop removed Dbus port workaround

jelly:
The pixel tests do not contain the 8765 port anymore, so this workaround can be dropped.


jelly:
The fedora rawhide flake is unrelated, merging!

PR #17813, opened 2022-10-12
bridge: python: drop disk.all.ops

jelly:
This was added in 2015 but it seems the client never used it, as there are no plans to use it drop it.

Was added in af2d488a27b7beb7e410a673923442ead7d02ac2 
The overhead of keeping it is minimal to be fair, but seems a bit wasteful to keep things which are unused since 2015 :)

### DIFF
--- a/src/bridge/cockpitdisksamples.c
+++ b/src/bridge/cockpitdisksamples.c
@@ -148,7 +148,6 @@ cockpit_disk_samples (CockpitSamples *samples)
 
   cockpit_samples_sample (samples, "disk.all.read", NULL, bytes_read);
   cockpit_samples_sample (samples, "disk.all.written", NULL, bytes_written);
-  cockpit_samples_sample (samples, "disk.all.ops", NULL, num_ops);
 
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
Are you sure that this isn't used from any (possibly 3rd party) plugins?

allisonkarlitskaya (commented):
In general, I hesitate to remove functionality from the C bridge, even if we think that nobody is using it...

jelly:
> Are you sure that this isn't used from any (possibly 3rd party) plugins?

For it to be exposed it needs to be enabled in `cockpitinternalmetrics`

Like for example `disk.all.written`
```
src/bridge/cockpitdisksamples.c:  cockpit_samples_sample (samples, "disk.all.written", NULL, bytes_written);
src/bridge/cockpitinternalmetrics.c:  { "disk.all.written", "bytes", "counter", FALSE, DISK_SAMPLER },
```

allisonkarlitskaya (approved):

PR #17823, opened 2022-10-18
test: Prevent bash TCO when keeping a mount point busy

mvollmer:
Different versions of bash do tail call optimization in different scenarios, and the "Currently in use" pop-up would either show the bash process or not (depending on whether TCO was done). So let's prevent bash from doing TCO of the "sleep" call.

### DIFF
--- a/test/verify/check-storage-used
+++ b/test/verify/check-storage-used
@@ -42,8 +42,10 @@ class TestStorageUsed(StorageCase):
         m.execute("mke2fs -q -L TEST /dev/mapper/dm-test")
         m.execute("mount /dev/mapper/dm-test /mnt")
 
-        # Keep the mount point busy
-        sleep_pid = m.spawn("cd /mnt; sleep infinity", "sleep")
+        # Keep the mount point busy.  The extra "true" is here to
+        # prevent bash from applying tail call optimization to the
+        # "sleep" invocation.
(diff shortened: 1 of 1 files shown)


mvollmer:
The alternative would be to expect TCO always and update the pixel
test references.  That's probably also feasible and less tricky.

mvollmer:
This is needed for the fedora-36 refresh in https://github.com/cockpit-project/bots/pull/3955

test/verify/check-storage-used
```diff
@@ -63,6 +65,7 @@ ExecStart=/usr/bin/sleep infinity
 
         self.content_dropdown_action(1, "Format")
         self.dialog_wait_open()
+        sit()
```
jelly, line 68: This should probably not be in the commit :)

jelly (changes requested):

test/verify/check-storage-used
mvollmer: Ouch!

mvollmer: Fixed.

jelly (approved):

mvollmer:
TestNetworkingCheckpoints.testNoRollback is fixed in https://github.com/cockpit-project/cockpit/pull/17829

PR #17826, opened 2022-10-18
Datepicker popup fixes

jelly:


jelly:
I think this ideally should go into the next release as it's an annoying visual bug.

pkg/lib/serverTime.js
```diff
@@ -531,7 +531,8 @@ function ChangeSystimeBody({ state, errors, change }) {
                                         weekStart={timeformat.firstDayOfWeek()}
                                         placeholder={timeformat.dateShortFormat()}
                                         onChange={d => change("manual_date", d)}
-                                        value={manual_date} />
+                                        value={manual_date}
+                                        appendTo={() => document.body} />
```
cockpituous, line 535: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17826-20221018-195623-2ba05daf-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/timer-dialog.jsx
```diff
@@ -335,7 +335,8 @@ const CreateTimerDialogBody = ({ owner }) => {
                                                             const arr = [...repeatPatterns];
                                                             arr[idx].date = str;
                                                             setRepeatPatterns(arr);
-                                                        }} />
+                                                        }}
+                                                        appendTo={() => document.body} />
```
cockpituous, line 339: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17826-20221018-195623-2ba05daf-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt (approved):
Thank you! This triggers a strong déjà vu, I remember fixing that already. But apparently we don't use appendTo= anywhere. Maybe that got lost at some point.

jelly:
> Thank you! This triggers a strong déjà vu, I remember fixing that already. But apparently we don't use appendTo= anywhere. Maybe that got lost at some point.

I made one similar PR, but this feels a bit strange yes. I guess this might be a Patternfly regression? I don't think we make pixel tests for the datepicker so that would be something to investigate.

garrett:
> an annoying visual bug

It's actually functional, making it hard to interact with, so it's much worse than just a "visual" glitch.

Thanks for this PR, and I agree, we should get it in ASAP.

And this did used to work, so we either broke something or PatternFly did. And our tests didn't catch it.

garrett (approved):
:+1: Looks good! I tested the creation of a new timer and rebooting the system at a specific time.

jelly:
Blocked by https://github.com/cockpit-project/cockpit/pull/17829

PR #17829, opened 2022-10-19
test: Don't use "journalctl -n -u"

mvollmer:
This seems to be bugged when systemd 251 rotates journal files, see https://github.com/systemd/systemd/issues/25063.

The NetworkManager journal is shorter than 1000 lines in this test anyway, so dropping "-n 1000" has no effect on performance.

### DIFF
--- a/test/verify/check-networkmanager-checkpoints
+++ b/test/verify/check-networkmanager-checkpoints
@@ -161,7 +161,7 @@ class TestNetworkingCheckpoints(NetworkCase):
         # back.
 
         def checkpoint_was_destroyed():
-            lines = m.execute("journalctl -n 1000 -u NetworkManager | grep op=").split("\n")
+            lines = m.execute("journalctl -u NetworkManager | grep op=").split("\n")
             last_checkpoint = None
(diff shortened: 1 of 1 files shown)


jelly:
RHEL seems to have flaked with a network issue...

```
 > https://cockpit-images.us-east-1.linodeobjects.com/ (authenticated)
image-download: argument of type 'NoneType' is not iterable
```
Retrying.

test/verify/check-networkmanager-checkpoints
```diff
@@ -161,7 +161,7 @@ class TestNetworkingCheckpoints(NetworkCase):
         # back.
 
         def checkpoint_was_destroyed():
-            lines = m.execute("journalctl -n 1000 -u NetworkManager | grep op=").split("\n")
+            lines = m.execute("journalctl -u NetworkManager | grep op=").split("\n")
```
jelly, line 164: A few days ago I learned that `journalctl` has a `-g` option so we could drop `| grep op=` in the future.

jelly (approved):

PR #17834, opened 2022-10-24
ssh: Copy session before using it to check known hosts

mvollmer:


martinpitt (approved):
Nice work, @mvollmer ! This looks like a fine workaround.

jelly:
F37 now randomly fails with a corrupt systemd journal, going to let this through to land the image refreshes.

martinpitt:
@jelly right, that's some very recent regression in F37. I saw this in the most recent [F37 bodhi update](https://bodhi.fedoraproject.org/updates/FEDORA-2022-957c7eaa0d) as well.

PR #17835, opened 2022-10-24
test: Avoid multi-arch conflict on mock-pam-conv-mod.so

martinpitt:
In Fedora you can install multiple architectures of a package
simultaneously. This causes a file conflict when trying to install both
the i686 and x86_64 architecture of "cockpit-tests", as they both ship
/usr/lib/cockpit-test-assets/mock-pam-conv-mod.so (and this is not a
magic multi-arch directory).

This is of no practical importance other than that rpminspect complains,
which causes a nasty red stain in gating.

Install the file into the proper PAM dir right away. This is properly
multi-arched in both Debian and Fedora, and tests then don't need to
know the exact directory any more. For people installing from upstream
this is still sufficiently guarded, as `make install` does not install
that module (only `make test-install` does).

---

This is the [rpmdeplint failure](https://artifacts.dev.testing-farm.io/0e7f4544-b665-47ac-8811-6f813efceab8/) of cockpit's [Fedora updates](https://bodhi.fedoraproject.org/updates/FEDORA-2022-4f1a38ada0). 

(I am also handling the rpminspect failure, but that needs to be addressed in rpmdeplint-data-fedora -- [fix](https://github.com/rpminspect/rpminspect-data-fedora/pull/29))

### DIFF
--- a/test/verify/check-superuser
+++ b/test/verify/check-superuser
@@ -107,14 +107,14 @@ class TestSuperuser(MachineCase):
         if "debian" in m.image or "ubuntu" in m.image:
             self.write_file("/etc/pam.d/sudo", """
 auth required pam_unix.so
-auth required /usr/lib/cockpit-test-assets/mock-pam-conv-mod.so
+auth required mock-pam-conv-mod.so
 @include common-account
 @include common-session-noninteractive
 """)
         else:
             self.write_file("/etc/pam.d/sudo", """
 auth required pam_unix.so
-auth required /usr/lib/cockpit-test-assets/mock-pam-conv-mod.so
+auth required mock-pam-conv-mod.so
 account include system-auth
 password include system-auth
 session include system-auth
--- a/src/ws/Makefile-ws.am
+++ b/src/ws/Makefile-ws.am
@@ -162,12 +162,11 @@ mock_pam_conv_mod_so_CFLAGS = -fPIC $(AM_CFLAGS)
 mock_pam_conv_mod_so_LDFLAGS = -shared
 mock_pam_conv_mod_so_LDADD = $(PAM_LIBS)
 
-testassetsdir = $(prefix)/lib/cockpit-test-assets
 testserviceddir = $(systemdunitdir)/cockpit.service.d
 
 install-tests::
-	mkdir -p $(DESTDIR)$(testassetsdir) $(DESTDIR)$(testserviceddir) $(DESTDIR)/etc/cockpit
-	$(INSTALL_DATA) mock-pam-conv-mod.so $(DESTDIR)$(testassetsdir)
+	mkdir -p $(DESTDIR)$(pamdir) $(DESTDIR)$(testserviceddir) $(DESTDIR)/etc/cockpit
+	$(INSTALL_DATA) mock-pam-conv-mod.so $(DESTDIR)$(pamdir)/
 
 install-exec-hook::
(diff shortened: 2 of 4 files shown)


martinpitt:
I can't say that I like this, it's a hack. But I can't think of a better way. If possible, we could add an explicit `Conflicts:` declaration to "any architecture other than my own", but I don't know how to express that. I spent some 30 minutes googling for Fedora's multilib specification and this error, but the [packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/) say nothing about this, and https://fedoraproject.org/wiki/PackagingDrafts/MultilibTricks does not cover this either.

test/verify/check-superuser
```diff
@@ -104,17 +104,18 @@ class TestSuperuser(MachineCase):
...
-            self.write_file("/etc/pam.d/sudo", """
+            self.write_file("/etc/pam.d/sudo", f"""
 auth required pam_unix.so
-auth required /usr/lib/cockpit-test-assets/mock-pam-conv-mod.so
+auth required /usr/lib/cockpit-test-assets/mock-pam-conv-mod.{arch}.so
 @include common-account
 @include common-session-noninteractive
 """)
         else:
             self.write_file("/etc/pam.d/sudo", """
```
mvollmer, line 116: `f"""`?

martinpitt: Ach, ja klar :grin:  So much for "throw this against the wall^W^WCI and see what sticks. Fixed! (The other place, too)

mvollmer:
> (and this is not a magic multi-arch directory).

I know nothing about multilib, but this makes me think: If we put the file into a magic multi-arch directory, will the conflict go away as well?

How does this work for regular pam modules, like our own pam_ssh_add.so?  They are in `/lib64/security/`.  Maybe we can put mock-pam-conv-mod.so into '/lib64/cockpit-test/`?  I think I would prefer that.

martinpitt:
@mvollmer : Then we'd need to put it into e.g. /usr/lib/x86_64-linux-gnu/ in Debian/Ubuntu (which have a more general multi-**arch** implementation, not just multi-**lib**), /usr/lib64 in Fedora/RHEL, etc. I wanted to avoid that as it makes it much harder for the test to predict where the file would live.

mvollmer:
> @mvollmer : Then we'd need to put it into e.g. /usr/lib/x86_64-linux-gnu/ in Debian/Ubuntu (which have a more general multi-**arch** implementation, not just multi-**lib**), /usr/lib64 in Fedora/RHEL, etc. I wanted to avoid that as it makes it much harder for the test to predict where the file would live.

Right, hmm. 

But if we install mock-pam-conv-mod.so next to pam_ssh_add.so (in /lib64/security on Fedora), then we can just refer to it without a explicit path, no?

```
self.sed_file('1 a auth       required    mock-pam-conv-mod.so')
```

But, yes, maybe it's too intrusive to have test assets outside of `/usr/lib/cockpit-test-assets`.

mvollmer (dismissed):

martinpitt:
> But if we install mock-pam-conv-mod.so next to pam_ssh_add.so

That's an interesting idea! This isn't urgent, so I'll keep this open for now and play around with this when I have the next "nothing else to do" time slot.

martinpitt:
@mvollmer : Thanks for the suggestion! This looks a lot nicer now indeed!

martinpitt:
Need to [fix spec for RHEL](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17835-20221025-132258-f243b3d5-rhel-8-7/log.html). But waiting for the Debians to get a result there.

mvollmer (approved):

PR #17840, opened 2022-10-25
build: patch <releases> in metainfo on install

allisonkarlitskaya:
If a DOWNSTREAM_RELEASES_XML environment variable is set while calling `make install` then patch the installed metainfo for Cockpit Client to use the release history from that external file.

This allows downstream packaging of Cockpit Client (particularly the Flatpak) to maintain its own version history, now that we don't ship one for ourselves.

### DIFF
--- a/tools/patch-metainfo
+++ b/tools/patch-metainfo
@@ -1,0 +1,35 @@
+#!/usr/bin/python3
+
+# Merges a downstream releases.xml file with the upstream metainfo.xml.
+#
+# Opens the metainfo file, finds the <releases> tag, and replaces its contents
+# with the contents of the <releases> tag found in the releases file.
+#
+# This facilitates keeping a downstream changelog, not shipped in tarballs.
+#
+# This substitution is meant to be done at `make install` time.  Set the
+# `DOWNSTREAM_RELEASES_XML` environment variable while calling `make install`.
+
+import argparse
+import xml.etree.ElementTree as ET
+
+parser = argparse.ArgumentParser()
+parser.add_argument('metainfo')
+parser.add_argument('releases')
+args = parser.parse_args()
+
+releases = ET.parse(args.releases)
+releases_releases = releases.find('releases')
+assert releases_releases is not None
+
+metainfo = ET.parse(args.metainfo)
+metainfo_releases = metainfo.find('releases')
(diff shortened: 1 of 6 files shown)


tools/patch-metainfo
```diff
@@ -0,0 +1,35 @@
...
+
+metainfo = ET.parse(args.metainfo)
+metainfo_releases = metainfo.find('releases')
+assert metainfo_releases is not None
+
+stub_release = metainfo_releases.find('release')
+if stub_release is not None:
+    metainfo_releases.remove(stub_release)
+
+metainfo_releases.extend(releases_releases)
```
martinpitt, line 33: This looks a little funny, like it'd completely remove the `<releases>` subtree, and then try to extend it?

martinpitt (commented):
This looks mostly sane (just one question), but what [happened here](https://github.com/cockpit-project/cockpit/actions/runs/3320303122/jobs/5486578506)?

tools/patch-metainfo
allisonkarlitskaya: I actually tried this — there's a clear() method that looked promising, but that also resulted in the formatting *around* the tag being broken.

This preserves it.

allisonkarlitskaya:
> but what [happened here](https://github.com/cockpit-project/cockpit/actions/runs/3320303122/jobs/5486578506)?

I asked in the Flatpak channel.  They say they're aware of the issue, but don't know how to fix it.  :shrug:

tools/patch-metainfo
martinpitt: OK -- as long as it works :-) it just really *looks* wrong.

martinpitt:
Oh, flatpak test succeeded now! I re-triggered the broken integration tests (infra/DNS argh argh). I'm fine with this now, will +1 once you are happy with your test release, and undraft this.

allisonkarlitskaya:
Tested here: https://github.com/allisonkarlitskaya/cockpit/actions/runs/3323115895/jobs/5493138357

You can see that it works and produces a reasonable looking diff.  It fails to push, of course, but that's to be expected.

I think this can go in now.

martinpitt (approved):
Really looking forward to the "blip" and "bloop" features! :+1:

PR #17842, opened 2022-10-25
.github: update webpack-jumpstart actions

jelly:
actions/{download,upload}-artifact@v3 uses nodejs 16 as runtime instead of the deprecated nodejs 12 runtime.

Seems I forgot some as can be seen [here](https://github.com/cockpit-project/cockpit/actions/runs/3319238959), not sure about the last warning we never use `set-output` so I guess that might be the action we use doing it?

### DIFF
--- a/.github/workflows/webpack-jumpstart.yml
+++ b/.github/workflows/webpack-jumpstart.yml
@@ -32,7 +32,7 @@ jobs:
           docker container rm -f "${id}"
 
       - name: Create artifact
-        uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Okay, https://github.com/actions/upload-artifact does document a @v3. The workflow here ran with `@v2` still as it runs from main, but (1) we can revert if it breaks, and (2) this will at most break the packit tests, not everything.

I triggered the required fcos test.

allisonkarlitskaya (approved):
:+1:

PR #17844, opened 2022-10-25
test: enable clevs tang test on Arch Linux

jelly:
Works locally, let's verify it here after the refresh:

https://github.com/cockpit-project/bots/pull/3998


mvollmer (approved):

jelly:
Woops, forgot to remove the now unrequired import, repushing.

PR #17847, opened 2022-10-26
test: Relax test skip expected output

martinpitt:
In Fedora 37's Python, the skip message now contains the test name:

    # SKIP testSkip (__main__.TestExample.testSkip)\n

Relax the regular expression, we don't need to pin it down *that* precisely.

----

This got exposed by https://github.com/cockpit-project/cockpituous/pull/531 , so that [this test now fails everywhere](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17839-20221026-054748-c5ecd1ca-arch/log.html#307-2). I think this is harmless/obvious enough to fix without rolling back the tasks container. This would also happen for developers locally, so we need to fix it either way.


marusak (approved):
Danke!

PR #17848, opened 2022-10-26
test: allowed systemd user journal corruption messages

jelly:
This happens on packit's Fedora 37.

### DIFF
--- a/test/browser/run-test.sh
+++ b/test/browser/run-test.sh
@@ -56,6 +56,11 @@ if [ "${TEST_OS#centos-8}" != "$TEST_OS" ]; then
     TEST_ALLOW_JOURNAL_MESSAGES="${TEST_ALLOW_JOURNAL_MESSAGES},couldn't create runtime dir: /run/user/1001: Permission denied"
 fi
 
+# HACK: https://github.com/systemd/systemd/issues/24150
+if [ "$ID" = "fedora" ]; then
+       TEST_ALLOW_JOURNAL_MESSAGES="${TEST_ALLOW_JOURNAL_MESSAGES},Journal file /var/log/journal/*/user-1000@*.journal corrupted, ignoring file .*"
+fi
+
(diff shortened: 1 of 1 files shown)


test/browser/run-test.sh
```diff
@@ -56,6 +56,11 @@ if [ "${TEST_OS#centos-8}" != "$TEST_OS" ]; then
     TEST_ALLOW_JOURNAL_MESSAGES="${TEST_ALLOW_JOURNAL_MESSAGES},couldn't create runtime dir: /run/user/1001: Permission denied"
 fi
 
+# HACK: https://github.com/systemd/systemd/issues/24150
+if [ "$ID" = "fedora" ]; then
+       TEST_ALLOW_JOURNAL_MESSAGES="Journal file /var/log/journal/*/user-1000@*.journal corrupted, ignoring file .*"
```
martinpitt, line 61: You need to append, like the other cases do. Please add the missing `${TEST_ALLOW_JOURNAL_MESSAGES},`

martinpitt (changes requested):
Ah, that's an even nicer way to do that than adding a nasty to bots/machine_core!

test/browser/run-test.sh
jelly: Oh I am so silly, yes!

martinpitt (approved):
Cheers! Please trigger the required test.

jelly:
```
# Result testBasic (__main__.TestTeam.testBasic) failed
```
So this now also fails in F37 or flakes...

jelly:
```
Error: Connection activation failed: IP configuration could not be reserved (no available address, timeout, etc.)
```

Is the issue.

martinpitt:
@jelly The Team test has been failing for ages (and a retry works), that's not new.

jelly:
> @jelly The Team test has been failing for ages (and a retry works), that's not new.

Yup, but I still want to get the flake away ;)

PR #17850, opened 2022-10-26
test: Python 3.11 compatibility

jelly:
Python 3.11 changed it's unittest internals for the Outcome class and no longer calls testPartExecutor with isTest as it used too. The Outcome class now has a TestResult attribute which records failures, errors and successful tests different.


mvollmer (approved):
LGTM!

PR #17851, opened 2022-10-27
Injector proof of concept

allisonkarlitskaya:
Here's a proof of concept for how the Python bridge can be used in the future.

We create a separate `injector.pyz` bundle, which contains:
 - the `cockpit/` and `systemd_ctypes/` code
 - the packages (`dist/`)
 - a "bootloader" script
 - another small bundle, containing:
   - the `cockpit/` and `systemd_ctypes/` code
   - no packages

When you run `injector.pyz` it runs a Python interpreter and plays the bootloader into it.  If the bootloader discovers that we already have the correct version of the bridge (in `~/.cache`) then it will run it.  Otherwise, it requests it from the injector, and runs it.

Since there's no packages on the remote side, the injector serves those from its own bundle, by diverting incoming requests to the `internal`:`packages` endpoint.

This all seems to work fairly OK, although (for a currently unknown reason) several unit tests are failing.

Here's a test:

```sh
$ cat ~/.local/bin/cockpit-bridge 
#!/bin/sh

~/src/cockpit/injector/injector.pyz ssh cockpit-7
```

and then I accessed `localhost` from Cockpit Client (ie: run the above script) and saw:

image

Note that `cockpit-7` is a Fedora IoT machine with no extra packages installed.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -53,9 +53,14 @@ pyTESTS = $(pyTESTS_PASSING) $(pyTESTS_FAILING)
 pycheck: cockpit-bridge.pyz
 	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE=./cockpit-bridge.pyz
 
+injcheck: injector.pyz
+	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE=./injector.pyz
+
 PYTHON_BRIDGE_FILES = \
 	src/cockpit/__init__.py \
 	src/cockpit/bridge.py \
+	src/cockpit/bootloader.py \
+	src/cockpit/injector.py \
 	src/cockpit/channel.py \
 	src/cockpit/channels/__init__.py \
 	src/cockpit/channels/dbus.py \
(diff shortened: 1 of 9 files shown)


garrett:
This is super cool. :sparkles:

allisonkarlitskaya:
Note: in the future, I'm not sure it will make sense for the injector to be a separate piece of software.

Out of the two major roles it performs:
 - the local package serving can be better addressed in cockpit-ws, or its future replacement, or even at a level closer to the web browser (like as a separate static webserver)
 - the sending of the bootloader and the bundle could also be incorporated in a lower-level part of `cockpit-ws`, or into `cockpit-ssh` or similar.  It would need to be in-band between the websocket and the eventual bridge, in any case.

martinpitt:
@allisonkarlitskaya : I'm on PTO, so no time to review/discuss in detail today (I have some initial questions, we should have a gmeet). However, the first four commits LGTM and could already land right away, so feel free to split out these cleanups. Thanks! Exciting to see this progress here!

src/Makefile.am
```diff
@@ -94,6 +99,23 @@ cockpit-bridge.pyz: $(MANIFESTS) $(PYTHON_BRIDGE_FILES) $(SYSTEMD_CTYPES_STAMP)
 	$(AM_V_GEN) python3 -m zipapp --python /usr/bin/python3 --output $@ --main cockpit.bridge:main tmp/pyz
 	@chmod a+x $@
 
+CLEANFILES += injector.pyz
+injector.pyz: $(MANIFESTS) $(PYTHON_BRIDGE_FILES) $(SYSTEMD_CTYPES_STAMP)
+	@rm -rf tmp/bundle-pyz
+	@mkdir -p tmp/bundle-pyz tmp/bundle-pyz/dist/ssh tmp/bundle-pyz/dist/shell
+	@cp -r $(srcdir)/src/cockpit src/systemd_ctypes/ tmp/bundle-pyz
+	@cp -r $(srcdir)/pkg/ssh/manifest.json tmp/bundle-pyz/dist/ssh
+	@cp -r $(srcdir)/pkg/shell/manifest.json tmp/bundle-pyz/dist/shell
```
martinpitt, line 108: I'm afraid I need some more background/explanation around what this does. As I see it, we currently have a ~ 12 MB cockpit-bridge.pyz (which bundles all the pages), and then an additional injector.pyz which contains another copy of the bridge code *and* another copy of the shell.

As a draft this is fine of course, but I'd like to understand the eventual plan here. cockpit-bridge.pyz  should not contain any webpack, right? These should be served either from XDG_DATA_DIRS on the target system, or from cockpit-ws. So bridge.pyz is going to be small-ish. Could it not just include the bootloader code? Or can the injector *only* contain the boot loader and fetch bridge.pyz from ws then? But if so, why does it make sense to have two zipapps and that additional logic, instead of just one?

allisonkarlitskaya: I think my idea going forward is something like this:

 - `cockpit-bridge.pyz`, as it is, should go away
 - installing the `cockpit-bridge` package on a server, as we do today, should look more like a traditional python application, with files installed into `site-packages` or whatever other distro-specific place is appropriate
 - running the `pycheck` tests in-tree should also run against the files directly.  Something like this works:
```
PYTHONPATH=src ./test-server python3 -m cockpit.bridge
```
 - we will want to produce the injector bundle for some specific use-cases:
   - the -ws container might want to inject this to machines that it connects to over SSH
   - ditto Cockpit Client

PR #17852, opened 2022-10-27
a few minor fixes

allisonkarlitskaya:
These are the trivial/non-controversial bits split out from #17851.

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -20,9 +20,14 @@ import asyncio
 
 from typing import Any, Dict, Iterable, Optional, Tuple
 
+from .protocol import CockpitProtocol
+
 
 class Endpoint:
-    router = None
+    router: CockpitProtocol
+
+    def __init__(self, router):
+        self.router = router
 
     def do_channel_control(self, command, message):
         raise NotImplementedError
--- a/src/cockpit/transports.py
+++ b/src/cockpit/transports.py
@@ -113,6 +113,7 @@ class _Transport(asyncio.Transport):
             self._is_reading = True
 
     def abort(self) -> None:
+        self._closing = True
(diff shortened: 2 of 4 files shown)


src/cockpit/channel.py
```diff
@@ -20,9 +20,14 @@
 
 from typing import Any, Dict, Iterable, Optional, Tuple
 
+from .protocol import CockpitProtocol
+
 
 class Endpoint:
-    router = None
+    router: CockpitProtocol
```
jelly, line 27: This is also mypy approved ^_^

src/cockpit/transports.py
```diff
@@ -143,7 +144,12 @@ def _write_eof_now(self) -> None:
 
     def _write_ready(self):
         assert self._queue is not None
-        n_bytes = os.writev(self._queue)
+
+        try:
+            n_bytes = os.writev(self._queue)
+        except BrokenPipeError:
+            self.abort()
+            return
```
jelly, line 152: On line 166 we have:

```
            if self._closing:
                self.abort()
```

As `self.abort` now sets `self._closing` but I guess this needs to be conditional.

jelly: Note, that `self._consider_closing()` seems to not be implemented so that sounds like a bug!

jelly (changes requested):

src/cockpit/transports.py
allisonkarlitskaya: It is a bug!  I think we don't fail the tests because we never test a situation where the write queue backs up enough to trigger this issue.

On the other hand: this might be exactly why a few random testcases fail for the injector.  I'll try to figure out how to test this...

jelly (approved):

PR #17854, opened 2022-10-28
test: enable dark theme pixel tests

jelly:
This introduces a new "layout" for the dark theme. The theme switching is implemented using the media emulation api of CDP and we always set it on set_layout. The reason the dark theme pixel tests are a layout, is that we don't want to implement a new (costly) scenario.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -124,19 +124,28 @@ opts.coverage = False
 default_layouts = [
     {
         "name": "desktop",
+        "theme": "light",
         "shell_size": [1920, 1200],
         "content_size": [1680, 1130]
     },
     {
         "name": "medium",
+        "theme": "light",
         "is_mobile": False,
         "shell_size": [1280, 768],
         "content_size": [1040, 698]
     },
     {
         "name": "mobile",
+        "theme": "light",
         "shell_size": [414, 1920],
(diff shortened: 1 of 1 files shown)


jelly:
`ImportError: cannot import name 'Literal'` in TF, so we can't use fancy features :)

jelly:
Firefox CDP does not seem to support `Emulation.setEmulatedMedia` which means we most likely we need to skip it.

garrett:
According to https://bugzilla.mozilla.org/show_bug.cgi?id=1549434, there's `emulateMedium` and `stopEmulatingMedium`. Perhaps it might work?

jelly:
> According to https://bugzilla.mozilla.org/show_bug.cgi?id=1549434, there's `emulateMedium` and `stopEmulatingMedium`. Perhaps it might work?

Hmmm, I tried Page.setEmulatedMedia which is what [playright](https://github.com/microsoft/playwright/blob/cb1dcccbe6def06a6d55c5e28e7223daf68befdb/packages/playwright-core/src/server/firefox/ffPage.ts#L377) does, but calling that in our CDP code returns function not found...

Firefox documents absolutely nothing about their CDP API, so I'll leave this as is.

jelly:
This was part of the main dark theme PR, can be closed.

PR #17319, opened 2022-05-05
pkg: enable dark theme for PF

jelly:
To do after merge:

- [ ] bump COCKPIT_REPO_COMMIT in machines and create a PR
- [ ] bump COCKPIT_REPO_COMMIT in podman and create a PR

---

# Dark mode support

Cockpit now switches to a dark theme when system dark mode is enabled.

PR #17855, opened 2022-10-28
overview: Fix split button variant

jelly:
Change separated out of https://github.com/cockpit-project/cockpit/pull/17319 so we only have one pixel test update in the PR


mvollmer (approved):
Code looks good, can't comment on the UX change.

garrett:
> Code looks good, can't comment on the UX change.

Reason: It looks more like a button (which it is), and, while good for light mode, it is super important for dark mode as it was too low contrast before.

Menu toggle looks like a menu by default, but we want something that's a split button, which menu toggle does provide — but it's neither a menu nor a toggle. :laughing: The split part does have a menu, but the main part should look more like an action button.
https://www.patternfly.org/v4/components/menu-toggle/html/#split-button-secondary-action

PR #17856, opened 2022-10-28
python: Fix transport buffered writing, add python unit tests

allisonkarlitskaya:


allisonkarlitskaya:
Interesting caveat:

If the toplevel `cockpit -> ./dist` symlink exists, it breaks coverage reporting for some reason.  Probably related to something being confused that the cockpit package lives in `./cockpit` instead of in `src/cockpit`.  Needs debugging.

allisonkarlitskaya:
A bit of clarification about goals here, since the topic has been raised: It's my intention that we hit (and remain at) 100% coverage with unit tests alone.  In general, this sort of exhaustive coverage testing is a lot easier to do with unit testing anyway.

jelly (commented):
Would be nice to have the unit-tests container run these tests. Guess we only have to add tox? As new dependency or do we simply use the standard Python container with a matrix of Python versions which we need to support.

allisonkarlitskaya:
Here's how it looks like in [`systemd_ctypes`](https://github.com/allisonkarlitskaya/systemd_ctypes/blob/main/.github/workflows/tests.yaml#L79).

Basically, [`actions/setup-python`](https://github.com/actions/setup-python) + `pip install tox` and off you go.

allisonkarlitskaya:
@jelle: I've added a workflow (unit-tests / pytest).

src/cockpit/transports.py
```diff
@@ -394,6 +404,7 @@ class StdioTransport(_Transport):
         - character devices (including terminals)
         - sockets
     """
+
```
martinpitt, line 406: Given how the previous PR made it past CI, this can only mean that we don't cover src/python/ in our tests/static-code?

pyproject.toml
```diff
@@ -0,0 +1,66 @@
+[build-system]
```
martinpitt, line 1: This triggers a big "tech debt/duplication/combinatorial explosion/confusion" red flag for me. This duplicates ./flake8 and test/static-code, and introduces all kinds of new dependencies, like `tox` and whatnot. Aside from the fact that they need to be documented and added to our unit-tests container, I must say I dislike wrappers like `tox` -- it is often rather hard to peel them off to debug a single test case, and invoke the plain test .py file directly. It also isn't integrated into `make check`, and instead starts to build a parallel toolset. I don't see anything in test_transport.py which couldn't/shouldn't work with e.g. `PYTHONPATH=src python3 -m unittest discover`?

Also, this isn't really a Python *module* that we'd ever want to publish or support -- the fact that the bridge is written in Python ought to be an implementation detail. Do you introduce this file only for putting in the parameters for static code checkers? That I can agree with, but then we don't need most of the other metadata and setuptools config here.

.github/workflows/unit-tests.yml
```diff
@@ -43,3 +43,25 @@ jobs:
...
+    permissions: {}
+    timeout-minutes: 10
+    env:
+      TOX_TESTENV_PASSENV: FORCE_COLOR
+      FORCE_COLOR: 1
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
```
martinpitt, line 58: Strong veto. Please test the *real* thing (python and installed libraries) from at least one of our supported OSes. This can be the unittest container.

Such artificial environments are irrelevant for telling us whether our code actually works (e.g. on CentOS 8 or RHEL 9 or current Python in Fedora 36), and it's also not reproducible locally for development.

pyproject.toml
```diff
@@ -40,7 +40,7 @@ log_cli = true
 source = ["src", "*/site-packages"]
 
 [tool.coverage.run]
-source = ["cockpit"]
+source_pkgs = ["cockpit"]
```
martinpitt, line 45: Please squash this commit ("use source_pkgs for coverage")

.gitignore
```diff
@@ -6,6 +6,7 @@
 
 # general patterns
 *.a
+*.egg-info/
```
martinpitt, line 9: Please squash

martinpitt (changes requested):
Thanks! I'm excited to see the bug fixes and unit test coverage!

This PR is a bit over-split, the fixups should be squashed.

I'm unhappy about introducing pyproject/tox/virtualenv here, though. That may require some more in-depth/interactive discussion, and possibly a third person as arbiter :balance_scale:

pyproject.toml
allisonkarlitskaya: This mostly exists as the one central location for config settings for:
 - mypy
 - pylint
 - pytest
 - coverage
 - tox
 - probably something I forgot

but also, the project metadata is needed for building the package, which is required for tox to construct the virtual environment for testing.  I agree that we're probably not going to want to ship this package as such (ie: on PyPI) any time soon.  That's why, for example, the version is hardcoded to 0.

allisonkarlitskaya: Note also: assuming you have `pytest` installed, you can run `pytest` directly, and that will work, out of the source tree.  And indeed: that's because of the configuration in `pyproject.toml`.

allisonkarlitskaya: One more note: I'd like to be able to do the python testing without having configured the project.  In particular, I find the way that we currently import `systemd_ctypes` (which I've intentionally avoided testing, until now) to be a bit of a hack.  Being able to specify that, python-style in a `requirements.txt`, for example, seems a lot nicer to me, and it's precisely I'm planning to do as a next step.

We could, of course, try to figure out a way to launch `pytest` via our existing build system, but ... eh...

I think our 1:1 today will be interesting :)

.github/workflows/unit-tests.yml
allisonkarlitskaya: OK.  I think one of our containers would be fine, in that case.  I'll look into that.  The Python setup in the `ubuntu-latest` VM on GitHub actions is positively ancient.

pyproject.toml
allisonkarlitskaya: Yup.

.gitignore
allisonkarlitskaya: Yup.

pyproject.toml
martinpitt: > This duplicates ./flake8 and test/static-code

A more careful inspection: It doesn't actually; I mis-read e.g. the `max-line-length`, but that's for pylint. It'd be nice to move .flake8 into pyproject.toml, but that's another story, difficult, and not a blocker.

pyproject.toml
```diff
@@ -0,0 +1,71 @@
...
+
+[tool.setuptools.packages.find]
+where=["src"]
+include=["cockpit"]
+
+[tool.mypy]
+mypy_path = 'src'
+
+[tool.pylint]
+max-line-length = 118
```
martinpitt, line 25: I'd vote for just entirely disabling line length checking in pylint. We already do that in flake8, and then we don't need to keep the number in sync.

pyproject.toml
jelly: vulture has pyproject support https://github.com/jendrikseipp/vulture#configuration

allisonkarlitskaya:
This is looking very promising: https://github.com/cockpit-project/cockpit/actions/runs/3376426202/jobs/5604134752

src/cockpit/transports.py
allisonkarlitskaya: I can't get any of the tools to complain about this.  I think it'll accept it one way or the other.

In any case, PEP 257 is fairly clear about what's correct:

> Insert a blank line after all docstrings (one-line or multi-line) that document a class – generally speaking, the class’s methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.

martinpitt: Ah, I looked wrong. I thought this was a missing empty line before a function declaration. But of course it's just bewetween docstring and body.

allisonkarlitskaya:
So we need to figure out what to do about the cgroup sampler finding no cgroups in the unit tests container.

We also maybe want to think about the fact that `make pycheck` vs. `make pytest` is possibly a bit silly.  At a minimum, it would probably make sense to have one target that ran both, and then we could reduce our number of workflow scenarios.

I think I'd like to start thinking about getting rid of the `cockpit-bridge.pyz` soon, and running the qunit tests against the unpacked files.  Something like this works:

```
PYTHONPATH=src ./test-server /usr/bin/python3 -m cockpit.bridge
```

martinpitt (dismissed):
Thanks! I'm much happier about this now. One new small bug, and there's still two outstanding threads from the previous review round: Disabling line length check in pylint, and if we really need all that build-related project metadata in pyproject.toml (as opposed to just adding the linter config there).

tools/systemd_ctypes
```diff
@@ -0,0 +1,26 @@
...
+# we default to check out the node_modules corresponding to the gitlink in the index
+sha="${1-$(get_index_gitlink "${SUBDIR}")}"
+
+# fetch by sha to prevent us from downloading something we don't want
+fetch_sha_to_cache "${sha}"
+
+# we're actually going to do this; let's remove the old one
+rm -rf "${SUBDIR}"
+git submodule deinit "${SUBDIR}"
+rm -rf -- "$(git rev-parse --absolute-git-dir)/${SUBDIR}"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Funny.  I remember having exactly this thought when I wrote it in the first place, for `node_modules`.  Let's go back to hardcoding the values.

src/Makefile.am
```diff
@@ -53,6 +53,12 @@ pyTESTS = $(pyTESTS_PASSING) $(pyTESTS_FAILING)
 pycheck: cockpit-bridge.pyz
 	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE=./cockpit-bridge.pyz
 
+pytest:
+	cd '$(srcdir)' && pytest
+
+pytest-cov:
```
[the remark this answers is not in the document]
allisonkarlitskaya: Thanks

allisonkarlitskaya:
> Thanks! I'm much happier about this now. One new small bug, and there's still two outstanding threads from the previous review round: Disabling line length check in pylint, and if we really need all that build-related project metadata in pyproject.toml (as opposed to just adding the linter config there).

I actually don't agree with disabling the line-length check.  I use pylint as part of my ALE config and this is helpful for me.

allisonkarlitskaya:
Also note: getting rid of the build system metadata in the pyproject.toml will kill tox support.  I guess I can live with that...

martinpitt:
> I actually don't agree with disabling the line-length check. I use pylint as part of my ALE config and this is helpful for me.

Hmm, but ALE already uses .flake8. That usually works well enough for me, and complains about line lengths that way. My main concern is to not having to specify "our" max line length in two different places.

> getting rid of the build system metadata in the pyproject.toml will kill tox support

Ah, I wasn't aware of that. This will actually try to build/install the project into the venv, I suppose? Then keep it (if you care), I mostly wonered why we need it.

src/cockpit/samples.py
```diff
@@ -163,7 +163,7 @@ def sample(self, samples: Samples) -> None:
                 [dev_major, _, dev_name, _, num_reads_merged, num_sectors_read, _, _, num_writes_merged, num_sectors_written, *_] = line.strip().split()
 
                 # ignore device-mapper and md
-                if (dev_major == 253 or dev_major == 9):
+                if dev_major in ['9', '253']:
```
jelly, line 166: Thanks, that was a dumb mistake from my part :)

martinpitt:
The two TF failures are due to your branch being out of date. But we can ignore them here.

allisonkarlitskaya:
> The two TF failures are due to your branch being out of date. But we can ignore them here.

Ah, thanks!  I was doing a `fedora-37 `run locally to double-check, and it seems fine, indeed.

martinpitt (commented):
Thanks! Code-wise this looks good to me now. The duplicate pylint line length check isn't so important, not a reason to stall this mega-PR further.

Just the "add a framework for unit testing" commit message is severely wrong now:
"with enough information to package Cockpit" and "You can run the tests (with coverage reporting) via `tox`.". Instead, it'd be nice to document the new `pytest` and coverage targets in HACKING.md.

jelly (approved):

martinpitt (approved):
Cheers!

PR #17862, opened 2022-10-31
containers/ws: Clean up and re-automate release

martinpitt:
This brings back automatic releases. As we now build from pure Fedora, we cannot predict when they actually make it through bodi. But refreshing them every week is good enough in most cases (and if not, we can still trigger a manual refresh).

It's also desirable to refresh the container image regularly even if there is no new cockpit upstream release -- it can then still pick up Fedora security fixes in other packages.

### DIFF
--- a/.github/workflows/build-ws-container.yml
+++ b/.github/workflows/build-ws-container.yml
@@ -9,28 +9,15 @@ jobs:
     environment: quay.io
     permissions: {}
     timeout-minutes: 20
+    env:
+      RUNC: docker
 
     steps:
       - name: Clone repository
         uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


martinpitt:
@marusak had an initial look and "LGTM". I [triggered a run](https://github.com/cockpit-project/cockpit/actions/runs/3377115670). It [built the 278 tag](https://quay.io/repository/cockpit/ws?tab=tags) (and latest). I tested it with

    podman run -it --rm --pull always -p 9999:9090 quay.io/cockpit/ws

At https://localhost:9999/ I get the login page and can log into my host on `host.containers.internal`.

marusak (approved):
Danke!

PR #17863, opened 2022-11-01
tools: disallow root login by default

jelly:
## Disallow root login by default
On all operating systems, logging in as root with a username/password via SSH has been disallowed for a long time. Cockpit now also disallows it by default on new installations, but still allows it on upgrades for backwards compatibility. You can configure this in `/etc/cockpit/disallowed-users`.

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -42,6 +42,11 @@ if grep -q 'ID=.*rhel' /etc/os-release; then
     dnf install -y kpatch kpatch-dnf
 fi
 
+# On CentOS Stream 8 the cockpit package is upgraded so the file isn't touched.
+if [ ! -f /etc/cockpit/disallowed-users ]; then
+    echo 'root' > /etc/cockpit/disallowed-users
+fi
+
 #HACK: unbreak RHEL 9's default choice of 999999999 rounds, see https://bugzilla.redhat.com/show_bug.cgi?id=1993919
 sed -ie 's/#SHA_CRYPT_MAX_ROUNDS 5000/SHA_CRYPT_MAX_ROUNDS 5000/' /etc/login.defs
 
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1587,7 +1587,9 @@ class MachineCase(unittest.TestCase):
                 self.check_pixel_tests()
         shutil.rmtree(self.tmpdir, ignore_errors=True)
 
-    def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False):
+    def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False, enable_root_login=False):
+        if enable_root_login:
+            self.enable_root_login()
         self.machine.start_cockpit(tls=tls)
         self.browser.login_and_go(path, user=user, host=host, superuser=superuser, urlroot=urlroot, tls=tls)
 
@@ -2029,6 +2031,16 @@ class MachineCase(unittest.TestCase):
         m = self.machine
         self.restore_file(path, post_restore_action=post_restore_action)
         m.write(path, content, append=append, owner=owner, perm=perm)
+
+    def enable_root_login(self):
+        """Enable root login
+
+        By default root login is disabled in cockpit, removing the root entry of /etc/cockpit/disallowed-users allows root to login.
+        """
+
+        # fedora-coreos runs cockpit-ws in a containter so does not install cockpit-ws on the host
+        if not self.machine.ostree_image:
+            self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
 
 
 def jsquote(js: str) -> str:
--- a/test/image-prepare
+++ b/test/image-prepare
@@ -119,6 +119,10 @@ def build_install_rhel8(dist_tar, image, verbose, quick):
     # in distropkg, keep basic OS packages, otherwise build/install cockpit
     if 'distropkg' not in image:
         args += [
+            # remove already installed packages
+            "--run-command",
+            "if rpm -q cockpit-ws >/dev/null 2>&1; then rpm --erase --nodeps --verbose cockpit cockpit-ws cockpit-bridge cockpit-system; fi",
+
             # create cockpit.spec
             "--run-command",
             f"tar xf '{vm_dist_tar}' -O '*/tools/cockpit.spec' | sed '/%define build_all/d' > /var/tmp/cockpit.spec",
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -631,14 +631,20 @@ class TestConnection(MachineCase):
             else:
                 m.execute("rpm --erase cockpit cockpit-ws")
 
-        # upgrade from distro version (our images have cockpit-ws preinstalled) sets up dynamic motd/issue symlink
+        # clean install sets up dynamic motd/issue symlink and pam disallowed users.
         self.assertIn('Activate the web console', m.execute("cat /etc/motd.d/cockpit"))
         self.assertIn('Activate the web console', m.execute("cat /etc/issue.d/cockpit.issue"))
+        self.assertIn('root', m.execute("cat /etc/cockpit/disallowed-users"))
+        # remove disallowed-users to simulate an upgrade where we did not have this file yet.
+        m.execute("rm /etc/cockpit/disallowed-users")
 
         # package upgrade keeps them
         install()
         self.assertIn('Activate the web console', m.execute("cat /etc/motd.d/cockpit"))
         self.assertIn('Activate the web console', m.execute("cat /etc/issue.d/cockpit.issue"))
+        # disallowed-users should not exists now as this is an upgrade
+        m.execute("test ! -e /etc/cockpit/disallowed-users")
+        m.execute("echo 'root' > /etc/cockpit/disallowed-users")
 
         # HACK: On Arch Linux the symlink is overwritten, bug?
         if m.image != "arch":
@@ -652,12 +658,14 @@ class TestConnection(MachineCase):
         remove()
         m.execute("test ! -e /etc/motd.d/cockpit")
         m.execute("test ! -e /etc/issue.d/cockpit.issue")
+        m.execute("test ! -e /etc/cockpit/disallowed-users")
 
         # fresh install (most of our test images have cockpit-ws preinstalled, so the first test above does not cover that)
         install()
         # verify that we installed relative links in the new package
         self.assertEqual(m.execute("readlink /etc/motd.d/cockpit").strip(), "../../run/cockpit/motd")
         self.assertEqual(m.execute("readlink /etc/issue.d/cockpit.issue").strip(), "../../run/cockpit/motd")
+        self.assertIn('root', m.execute("cat /etc/cockpit/disallowed-users"))
 
     @skipImage("OSTree doesn't have cockpit-ws", "fedora-coreos")
     @nondestructive
@@ -737,7 +745,7 @@ class TestConnection(MachineCase):
         m = self.machine
         b = self.browser
 
-        self.login_and_go("/playground/speed", user="root")
+        self.login_and_go("/playground/speed", user="root", enable_root_login=True)
 
         # Check the speed playground page
         b.switch_to_top()
--- a/test/verify/check-static-login
+++ b/test/verify/check-static-login
@@ -639,6 +639,8 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
             expect_fail_count(0, user=user)
             b.logout()
 
+        self.enable_root_login()
+
         # ensure we have no module in our pam config already
         # arch has faillock enabled by default
         if m.image != "arch":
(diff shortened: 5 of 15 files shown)


jelly:
A bunch of test failures as expected:

TestConnection.testWsPackage

And tests which try to login as root.

jelly:
This needs a new test which tests our upgrade behaviour and the default behaviour of not allowing a root login.

jelly:
To test upgrading we have the current installed packages in our prepared virtual machine in `/var/tmp/build/`, so to test if an upgrade does not add a the new file:

1. verify the file is there
1. remove the file
2. upgrade using --force
3. verify the file is not there

test/common/testlib.py
```diff
@@ -1567,7 +1567,9 @@ def tearDown(self):
                 self.check_pixel_tests()
         shutil.rmtree(self.tmpdir, ignore_errors=True)
 
-    def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False):
+    def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False, enable_root_login=False):
+        if enable_root_login:
+            self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
```
jelly, line 1572: This should just call `self.enable_root_login`

tools/cockpit.spec
```diff
@@ -481,6 +482,7 @@ if [ "$1" = 1 ]; then
     mkdir -p /etc/motd.d /etc/issue.d
     ln -s ../../run/cockpit/motd /etc/motd.d/cockpit
     ln -s ../../run/cockpit/motd /etc/issue.d/cockpit.issue
+    echo -e "# List of users which are not allowed to login to Cockpit\nroot" > /etc/cockpit/disallowed-users
```
jelly, line 485: Also update line 480

jelly:
Ok great... centos-8-stream on packit upgrades cockpit so no file is created. We need to work around this in tests..

test/common/testlib.py
```diff
@@ -1587,7 +1587,9 @@ def tearDown(self):
...
```
martinpitt: DRY. Please call self.enable_root_login()

test/verify/check-connection
```diff
@@ -634,11 +634,17 @@ class TestConnection(MachineCase):
         # upgrade from distro version (our images have cockpit-ws preinstalled) sets up dynamic motd/issue symlink
         self.assertIn('Activate the web console', m.execute("cat /etc/motd.d/cockpit"))
         self.assertIn('Activate the web console', m.execute("cat /etc/issue.d/cockpit.issue"))
+        self.assertIn('root', m.execute("cat /etc/cockpit/disallowed-users"))
```
martinpitt, line 637: The comment at the top of this paragraph talks about upgrade; but this tests the "clean install" scenario, can you please add a comment? Also, I think that existing comment is actually wrong, no? "cockpit-ws preinstalled" only applies to the -distropkg scenario, in all other scenarios we rpm -e all cockpit packages first.

tools/arch/cockpit.install
```diff
@@ -0,0 +1,7 @@
+post_install() {
+  printf "# List of users which are not allowed to login to Cockpit\nroot\n" > /etc/cockpit/disallowed-users
```
martinpitt, line 2: So that bit will only run on fresh installs, not on upgrades?

Shuold get same chmod as for rpm.

tools/arch/cockpit.install
```diff
@@ -0,0 +1,7 @@
...
+}
+
+post_remove() {
+  [ -f /etc/cockpit/disallowed-users ] && rm /etc/cockpit/disallowed-users || true
```
martinpitt, line 6: Feels a bit easier to just do `rm -f`?

tools/cockpit.debian.pam
```diff
@@ -20,3 +20,6 @@ session    required     pam_env.so # [1]
 # In Debian 4.0 (etch), locale-related environment variables were moved to
 # /etc/default/locale, so read that as well.
 session    required     pam_env.so user_readenv=1 envfile=/etc/default/locale
+
+# List of users to deny access to Cockpit, by default root is included.
+auth       required     pam_listfile.so item=user sense=deny file=/etc/cockpit/disallowed-users onerr=succeed
```
martinpitt, line 25: Please move this into the "auth" block, to keep everything together.

tools/cockpit.pam
```diff
@@ -15,3 +15,5 @@ session    optional     pam_keyinit.so force revoke
 session    optional     pam_ssh_add.so
 session    include      password-auth
 session    include      postlogin
+# List of users to deny access to Cockpit, by default root is included.
+auth       required     pam_listfile.so item=user sense=deny file=/etc/cockpit/disallowed-users onerr=succeed
```
martinpitt, line 19: Dito

tools/cockpit.spec
```diff
@@ -481,6 +482,8 @@ if [ "$1" = 1 ]; then
     mkdir -p /etc/motd.d /etc/issue.d
     ln -s ../../run/cockpit/motd /etc/motd.d/cockpit
     ln -s ../../run/cockpit/motd /etc/issue.d/cockpit.issue
+    printf "# List of users which are not allowed to login to Cockpit\nroot\n" > /etc/cockpit/disallowed-users
+    # chmod 600 /etc/cockpit/disallowed-users
```
martinpitt: Hmm -- I have a feeling 644 is ok, but in the face of dynamic `umasks` I think it's a good idea to explicitly set permissions either way. 600 would be more cautious, also fine for me. Please pick one, and make it consistent everywhere (including the %ghost)

tools/debian/cockpit-ws.postinst
```diff
@@ -26,6 +26,7 @@ if [ "$1" = "configure" ] && [ -z "$2" ]; then
     mkdir -p /etc/motd.d /etc/issue.d
     ln -s ../../run/cockpit/motd /etc/motd.d/cockpit
     ln -s ../../run/cockpit/motd /etc/issue.d/cockpit.issue
+    printf "# List of users which are not allowed to login to Cockpit\nroot\n" > /etc/cockpit/disallowed-users
```
martinpitt, line 29: same chmod as for rpm

test/common/testlib.py
```diff
@@ -1589,7 +1589,7 @@ def tearDown(self):
 
     def login_and_go(self, path=None, user=None, host=None, superuser=True, urlroot=None, tls=False, enable_root_login=False):
         if enable_root_login:
-            self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
+            self.enable_root_login()
```
martinpitt, line 1592: Rebase error, this should be in the first commit (and resolves a comment)

test/common/testlib.py
```diff
@@ -2037,7 +2037,10 @@ def enable_root_login(self):
 
         By default root login is disabled in cockpit, removing the root entry of /etc/cockpit/disallowed-users allows root to login.
         """
-        self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
+
+        # fedora-coreos runs cockpit-ws in a containter so does not install cockpit-ws on the host
+        if not self.machine.ostree_image:
+            self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
```
martinpitt, line 2043: same, please move into first commit. Unrelated to "erase rpms before installing on rhel8"

tools/arch/cockpit.pam
```diff
@@ -3,3 +3,6 @@ auth      include   system-remote-login
 account   include   system-remote-login
 password  include   system-remote-login
 session   include   system-remote-login
+
+# List of users to deny access to Cockpit, by default root is included.
+auth       required     pam_listfile.so item=user sense=deny file=/etc/cockpit/disallowed-users onerr=succeed
```
martinpitt, line 8: again, this should be in the first commit.

tools/cockpit.spec
```diff
@@ -478,12 +478,12 @@ if [ -x %{_sbindir}/selinuxenabled ]; then
 fi
 
 # set up dynamic motd/issue symlinks on first-time install; don't bring them back on upgrades if admin removed them
+# disable root login on first-time install; so existing installations aren't changed
```
martinpitt, line 481: that, too

test/image-prepare
```diff
@@ -119,6 +119,10 @@ def build_install_rhel8(dist_tar, image, verbose, quick):
     # in distropkg, keep basic OS packages, otherwise build/install cockpit
     if 'distropkg' not in image:
         args += [
+            # remove already installed packages
+            "--run-command",
+            "if rpm -q cockpit-ws >/dev/null 2>&1; then rpm --erase --nodeps --verbose cockpit cockpit-ws cockpit-bridge cockpit-system; fi",
```
martinpitt, line 124: Please move this commit to the first in the series, so that logically each commit can pass integration tests (for bisecting)

test/browser/browser.sh
```diff
@@ -37,6 +37,11 @@ if grep -q 'ID=.*fedora' /etc/os-release; then
     dnf install -y tcsh
 fi
 
+# On CentOS Stream 8 the cockpit package is upgraded so the file isn't touched.
+if grep -q 'PRETTY_NAME="CentOS Stream 8"' /etc/os-release; then
```
martinpitt, line 41: Hmm, this isn't really specific to CentOS, just the way images are built. At this point we can't tell any more if the test preparation installed or upgraded cockpit. I'd say, just check if the file doesn't exist

martinpitt (approved):
Dankjewel! Needs some cleanup and rebasing, but I agree to the approach. Please add a release note.

martinpitt (changes requested):
Err, I meant to say "request changes"..

test/browser/browser.sh
jelly: Right, good point!

test/verify/check-connection
jelly: Right, this test is not executed on any `distropkg` scenario, the whole class has a `@skipDistroPackage` decorator.

A prepared image has erased the original image's cockpit and installed our build version, so yes, this is a clean install and I'll update the comment.
```
[root@fedora-37-127-0-0-2-2201 ~]# rpm -qa | grep cockpit
cockpit-bridge-278.33.g16ad2b4d6.dirty-1.fc37.x86_64
cockpit-system-278.33.g16ad2b4d6.dirty-1.fc37.noarch
cockpit-storaged-278.33.g16ad2b4d6.dirty-1.fc37.noarch
cockpit-networkmanager-278.33.g16ad2b4d6.dirty-1.fc37.noarch
cockpit-selinux-278.33.g16ad2b4d6.dirty-1.fc37.noarch
cockpit-packagekit-278.33.g16ad2b4d6.dirty-1.fc37.noarch
cockpit-pcp-278.33.g16ad2b4d6.dirty-1.fc37.x86_64
cockpit-ws-278.33.g16ad2b4d6.dirty-1.fc37.x86_64
```

tools/arch/cockpit.install
jelly: Yes, this only runs on clean installs, for upgrades there is `post_upgrade`.

test/verify/check-static-login
```diff
@@ -720,13 +722,17 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
     @skipImage("root logins disabled by default with ssh", "fedora-coreos")
     def testPamAccess(self):
         b = self.browser
+        m = self.machine
 
-        # root login works by default
-        self.login_and_go("/system", user="root")
-        b.wait_visible('.system-information')
-        b.logout()
+        # root login is disabled by default default
```
martinpitt, line 727: Are you sure sure? :grin:

tools/cockpit.spec
```diff
@@ -477,10 +478,12 @@ if [ -x %{_sbindir}/selinuxenabled ]; then
...
 if [ "$1" = 1 ]; then
     mkdir -p /etc/motd.d /etc/issue.d
     ln -s ../../run/cockpit/motd /etc/motd.d/cockpit
     ln -s ../../run/cockpit/motd /etc/issue.d/cockpit.issue
+    printf "# List of users which are not allowed to login to Cockpit\nroot\n" > /etc/cockpit/disallowed-users
```
martinpitt, line 486: missing `chmod 644`, to be robust against umask.

martinpitt (changes requested):
Thanks! Very close now, only a missing chmod. That arch failure doesn't look like a flake at first sight?

jelly:
> [arch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17863-20221114-125518-de60e1bc-arch/log.html) failed with
> 
> ```
> ==> Validating source files with sha256sums...
>     cockpit-279.15.ge35c9668e.tar.xz ... Skipped
>     cockpit.pam ... FAILED
>     cockpit-ws.sysuser.conf ... Passed
>     cockpit-wsinstance.sysuser.conf ... Passed
> ==> ERROR: One or more files did not pass the validity check!
> ```
> 
> That's related to that magic SHA sum update? Does that require an image refresh?

Nope, that's validating the file I changed, I guess we should skip checksums for those files.

martinpitt (approved):
Dankjewel!

(FTR, it's really cool that we cover upgrades in our test suite!)

martinpitt:
@jelly : I tweaked the release note a bit, WDYT?

jelly:
> @jelly : I tweaked the release note a bit, WDYT?

LGTM!

PR #17865, opened 2022-11-02
RHEL 8 host test fixes

jelly:


test/image-prepare
```diff
@@ -131,7 +131,7 @@ def build_install_rhel8(dist_tar, image, verbose):
             # build rpms in mock
             "--run-command",
             """su builder -c 'mock --no-clean --no-cleanup-after --disablerepo=* --offline --resultdir /var/tmp/build """
-            f"""--nocheck {"--verbose" if verbose else ""} --rebuild /var/tmp/build/SRPMS/*.src.rpm' """,
+            f"""{"--nocheck" if quick else ""} {"--verbose" if verbose else ""} --rebuild /var/tmp/build/SRPMS/*.src.rpm' """,
```
martinpitt, line 134: Thanks. Sorry for that blunder!

src/ws/test-channelresponse.c
```diff
@@ -541,6 +541,15 @@ test_resource_failure (TestResourceCase *tc,
   g_assert (cockpit_pipe_get_pid (tc->pipe, &pid));
   g_assert_cmpint (pid, >, 0);
   int pid_fd = syscall(SYS_pidfd_open, pid, 0);
+  if (pid_fd < 0)
+    {
+      if (errno == ENOSYS)
+	g_test_skip ("no pidfd_open support, skipping");
+      else
+        g_test_skip (strcat("pidfd_open call failed: ", g_strerror (errno)));
```
martinpitt, line 549: We shouldn't silently skip the test if that fails for any other reason, as then we have screwed up the code. Also, the strcat() here is both invalid (you can't append to a const string), and if it *would* copy, it'd be a memory leak. To fix all of these, use this:

    g_error ("pidfd_open call failed: %m");

martinpitt (changes requested):

test/image-prepare
jelly: No worries!

src/ws/test-channelresponse.c
jelly: Ah yes, I was not 100% certain how we would like to handle it. `g_error` makes sense.

src/ws/test-channelresponse.c
```diff
@@ -541,6 +541,15 @@ test_resource_failure (TestResourceCase *tc,
...
   g_assert_cmpint (pid, >, 0);
   int pid_fd = syscall(SYS_pidfd_open, pid, 0);
+  if (pid_fd < 0)
+    {
+      if (errno == ENOSYS)
+	g_test_skip ("no pidfd_open support, skipping");
+      else
+        g_error ("pidfd_open call failed: %m");
+
+      return;
```
martinpitt, line 551: This is slightly "meh"/confusing, as g_error() aborts. It'd be a bit cleaner to move this into the "then" branch. But it doesn't matter in practice (not sure how strongly @allisonkarlitskaya feels about that). Thanks!

martinpitt (approved):

src/ws/test-channelresponse.c
jelly: Ah, I guess it's only logical when you don't know the glib api

src/ws/test-channelresponse.c
```diff
@@ -541,6 +541,15 @@ test_resource_failure (TestResourceCase *tc,
...
+    }
```
allisonkarlitskaya, line 552: I think I'd write it

```suggestion
  if (pid_fd < 0)
    {
      if (errno != ENOSYS)
        g_error ("pidfd_open call failed: %m");
        
      g_test_skip ("no pidfd_open support, skipping");
      return;
    }
```

But I don't think it's important.  I'd be fine with yours.

test/image-prepare
allisonkarlitskaya, line 134: Oof.  Thanks.

allisonkarlitskaya (approved):

martinpitt:
Ack, let's land this then, as it's a release blocker.

PR #17874, opened 2022-11-03
pf4: Fix dark mode dropmarker in form controls

garrett:
I'm specifically not using `:where()` here to bump up the specificity so that this will always take precedence over PF4's broken version. (So we don't have to use `!important`, in other words.)


jelly (dismissed):

jelly (approved):

jelly:
Packit tests fail, but as this is a visual issue which was caught by our pixel tests, I'm letting it through.

PR #17877, opened 2022-11-04
Move to gnu18 C standard, fail on old C constructs

martinpitt:
See https://fedoraproject.org/wiki/Changes/PortingToModernC

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -56,12 +56,17 @@ AM_CPPFLAGS = \
 	$(NULL)
 
 AM_CFLAGS = \
+	-std=gnu18 \
 	-pthread \
 	-Wall \
 	-Werror=strict-prototypes \
 	-Werror=missing-prototypes \
 	-Werror=implicit-function-declaration \
-	-Werror=pointer-arith -Werror=init-self \
+	-Werror=implicit-int \
+	-Werror=int-conversion \
+	-Werror=old-style-definition \
+	-Werror=pointer-arith \
+	-Werror=init-self \
 	-Werror=format=2 \
 	-Werror=return-type \
 	-Werror=missing-include-dirs \


src/Makefile.am
```diff
@@ -111,12 +111,17 @@ AM_CPPFLAGS = \
 	$(NULL)
 
 AM_CFLAGS = \
+	-std=gnu99 \
```
martinpitt, line 114: `gnu11` also works. Maybe we want to jump to that right away? That's what I [did in umockdev](https://github.com/martinpitt/umockdev/pull/201), and I think we don't even support ancient platforms which can't do C11. Opinions?

allisonkarlitskaya: I can't believe that we're debating between a 1-decade-old and a 2-decade-old vintage of a 5-decade-old language :rofl: 

C11 is fine with me, but that actually seems like a step back.  The GCC manpage mentions:

> gnu17
> gnu18
> GNU dialect of ISO C17.  This is the default for C code.


...but it's not like there's a huge difference between C17 and C11 anyway...

Meanwhile, the next highest option is `gnu2x` which is still marked experimental.

allisonkarlitskaya (commented):
I'm not crazy about the "let's clean this up later" sentiment...

martinpitt:
OK -- I had a look at the deprecated SSH functions, and it doesn't seem to be a simple function exchange; they say something about "now uses a callback" and such. I'll look into it in the next days..

src/Makefile.am
martinpitt: Ah, I thought the default was still c89 or so. Let's use gnu18 then!

martinpitt:
FTR, I've thrown about 2 hours on rewriting mock-sshd.c so far. The libssh API changed completely. We'll have to do it at some point, but this isn't a good usage of my time right now. I'll keep this in draft for now.

src/Makefile.am
martinpitt: Moved to gnu18.

martinpitt:
Ah, that mock-sshd.c was a red herring. I always build locally with `--enable-strict`, that's what turned these deprecation warnings (which we've had for a while already) into errors. So that mock-sshd.c thing is completely independent, and I took it out from this PR.

allisonkarlitskaya (approved):
This looks fine to me, but:

What's the point of all of the extra warnings?  If they get enabled by gnu18, then we ought not to require them, and if they don't, then what does it have to do with this PR?

allisonkarlitskaya:
Guess, after thinking about it for a moment: is it that these things are officially no longer part of the spec, but GCC supports them anyway, and only warns about that, and we want to turn those warns into outright errors?

If so, some mention of that in the commit message might be nice.

mvollmer:
mock-sshd rewrite is here: https://github.com/cockpit-project/cockpit/pull/17932

I think I fixed a bug or two in the process.

martinpitt:
@allisonkarlitskaya : On main when I do

```diff
--- src/common/cockpitauthorize.c
+++ src/common/cockpitauthorize.c
@@ -46,7 +46,7 @@
   } while (0)
 #endif
 
-static int logger_verbose = 0;
+static logger_verbose = 0;
 static void (* logger) (const char *data);
 
 #ifndef message
```

I only get a warning:
```
src/common/cockpitauthorize.c:49:8: warning: type defaults to ‘int’ in declaration of ‘logger_verbose’ [-Wimplicit-int]
   49 | static logger_verbose = 0;
      |        ^~~~~~~~~~~~~~
```

The point of this is to make them fatal. Adding `-std=gnu18` does not change that it only comes out as a warning, as it's merely making the current gcc default explicit (which is a good thing -- C standard is a property of the compiled source, not the compiler!).

Indeed the wiki page makes it sound like that would only be necessary for pre-C99, but that is apparently misleading.

So I believe the change itself is correct. So how about I add this to the commit message on squash&merge:

```
The C code standard is a property of the compiled source, not the 
compiler. So make that explicit (it's the default standard of current
gcc anyway).

C99 and later dropped a few old C constructs, but still compiles them --
unfortunately only with a warning. Make these fatal, to ensure we
actually comply with gnu18.
```

allisonkarlitskaya:
```
$ gcc -pedantic-errors -c test.c
test.c:1:8: error: type defaults to ‘int’ in declaration of ‘foo’ [-Wimplicit-int]
    1 | static foo;
      |        ^~~
```

allisonkarlitskaya:
By the docs:

> -pedantic-errors

>    Give an error whenever the base standard (see -Wpedantic) requires a diagnostic, in some cases where there is undefined behavior at compile-time and in some other cases that do not prevent compilation of programs that are valid according to the standard. This is not equivalent to -Werror=pedantic, since there are errors enabled by this option and not enabled by the latter and vice versa.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html


I feel like maybe this is what we're looking for.

martinpitt:
Hmm, seems GitHub ate my comment a few days ago: ` -pedantic-errors` sounds good, but triggers a lot more (and non-trivial) errors, so this will require more work. In particular, this one:

```
src/tls/connection.c: In function ‘buffer_write_to_fd’:
src/tls/connection.c:234:22: error: invalid use of structure with flexible array member [-Wpedantic]
  234 |       struct cmsghdr cmsg[2];
      |                      ^~~~
```
That actually is true, the current code is a major hack. There's also a handful of new errors in `make check`, e.g.

```
src/bridge/test-bridge.c: In function ‘test_bridge_init’:
/usr/include/glib-2.0/gobject/gsignal.h:582:55: error: ISO C forbids passing argument 6 of ‘g_signal_handlers_disconnect_matched’ between function pointer and ‘void *’ [-Wpedantic]
  582 |                                           0, 0, NULL, (func), (data))
      |                                                       ^~~~~~
src/bridge/test-bridge.c:79:3: note: in expansion of macro ‘g_signal_handlers_disconnect_by_func’
```

@allisonkarlitskaya , does your approval here stand, or do you want to block this on sorting out all the pedantic errors?

martinpitt:
> @allisonkarlitskaya , does your approval here stand, or do you want to block this on sorting out all the pedantic errors?

Friendly ping?

allisonkarlitskaya (approved):
Oh.  That's still here.  Sure :)

PR #17879, opened 2022-11-04
Don't set max-width for Patternfly's select

skobyda:
Since new patternfly update, [this PF commit](https://github.com/patternfly/patternfly/commit/5c0c776fbe056a7829c47cdb81d59a8a190e4a84) changed the width of our Selectors all across Cockpit and [broke pixel tests.](https://github.com/cockpit-project/cockpit-machines/pull/840)
Screenshot from 2022-11-04 13-17-37
Screenshot from 2022-11-04 13-17-44

However after digging into it, it seems that there is nothing wrong with [that PF commit](https://github.com/patternfly/patternfly/commit/5c0c776fbe056a7829c47cdb81d59a8a190e4a84), but rather it only managed for our custom patternfly override to take effect, which was setting Selector's menu to `20rem`.
To me, that override doesn't seem right, as we have selectors wider that. I guess it was introduced by accident or is no longer viable.

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -39,8 +39,11 @@ to wrap around when there isn't enough space.
   flex-wrap: wrap;
 }
 
+/* Fix select menu rendering */
 ul.pf-c-select__menu {
-  max-height: 20rem;
+  /* Don't get too tall */
+  max-height: min(20rem, 50vh);
+  /* Don't have a horizontal scrollbar */
   overflow-y: auto;
 }
 
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -40,7 +40,6 @@ to wrap around when there isn't enough space.
 }
 
 ul.pf-c-select__menu {
-  max-width: 20rem;
   max-height: 20rem;
   overflow-y: auto;
 }


garrett:
You're right that this is visually wrong here.

I'm looking at git blame to try to figure out where this came from and why. This might give a hint:

Screenshot of https://github.com/cockpit-project/cockpit/blame/8eb58dbbfdee6c82c91b3dd1743a24237dbfc16d/pkg/lib/patternfly/patternfly-4-overrides.scss#L42-L46:
image

It's @ https://github.com/cockpit-project/cockpit/commit/7d44ac7b93223bcd86ea748ec548948fc0f397f7#diff-325ba8d07b7191456e22e2785c61be383372356c4e38b9a7cbfe35021be2af96R24

Which means this was thought to be important for the Cockpit AutoComplete. When I look for it in our codebase (`FileAutoComplete`), I only see it used in `credentials.jsx`. _However_, it's also used in a few places in Cockpit Machines, in `createVmDialog.jsx`, `createStoragePoolDialog.jsx`, `diskAdd.jsx`, `vmFilesystemsCard.jsx`.

Oddly, the widget doesn't seem to have its own classing, which is what we _should_ be using instead of messing with ALL of PF4 in overrides (which causes this issue).

So now I'm digging into this further to see if it's actually needed for file autocomplete (the commit suggests it is) and if it can be split out appropriately from the selects without messing up autocompletes.

jelly:
As this is a pure CSS change, I'll ignore the rawhide packit test.

PR #17881, opened 2022-11-07
containers: fix unit-tests with no submodules

allisonkarlitskaya:
The 'append' action of ArgumentParser will set an empty list as the value of the argument in the return namespace, but only the first time an argument is encountered.

Put another way: in case the argument isn't given, we get `None`, instead of `[]`.  We then go on to try to iterate over this to build the arguments to `git clone`, which doesn't work very well.

Use `default=` to make sure we always have a list.

 - [x] test refresh: https://github.com/cockpit-project/cockpit/actions/runs/3408639784

### DIFF
--- a/containers/unit-tests/start
+++ b/containers/unit-tests/start
@@ -124,7 +124,7 @@ def main():
                        help='Build the HEAD commit, plus changes on the filesystem (default)')
 
     group = parser.add_argument_group(title='Preparation').add_mutually_exclusive_group()
-    group.add_argument('--submodule', action='append', help='Check out this submodule')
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers! The other thing that seems missing is to run the new pycheck scenarios in the candidate container. Can you please fix as well? (Right here seems easiest)

allisonkarlitskaya:
> Cheers! The other thing that seems missing is to run the new pycheck scenarios in the candidate container. Can you please fix as well? (Right here seems easiest)

I didn't want to do that because I have another set of commits queued up that changes stuff around there.  I'll include it with that.

PR #17882, opened 2022-11-07
tests: fix periodic test-chan.html failure

allisonkarlitskaya:
The "filter message in" qunit test in test-chan.html fails periodically due to a race.  The source of this race is a workaround that was added in 66e283fd78f4e05c24ce803050eafac676b861d6:

  // HACK: send the reply asynchronously, so that mock_peer addEventListener() gets to see it before our own onrecv()
  window.setTimeout(() => self.send(null, JSON.stringify(reply), 0));

Unfortunately, this hack was only applied to the control messages, and not data.  This meant that for our "filter message in" test case, for example, we'd usually see the "one" "two" "three" data messages echoed back to us immediately — and terminate the test — before receiving the "ready" control message.

Every now and then, due to timing, we'd get the extra "ready" control message, though, and get an error looking like:

  not ok 154 - Expected 11 assertions, but 14 were run, expected: undefined, got: null, test: filter message in, module: , source: at http://localhost:32847/dist/base1/test-chan.js:2:148292

The race seemed more likely to happen when the python bridge was in use, and particularly more likely to happen while running tests under pytest with coverage enabled (which the next commit will introduce).

Since window.setTimeout() makes no guarantees about delivery order, let's move to a full-powered ordered queue for sending replies.  Now that "ready" is always guaranteed to be received before the data messages, bump the expected assert count to 14.

### DIFF
--- a/pkg/base1/test-chan.js
+++ b/pkg/base1/test-chan.js
@@ -16,6 +16,9 @@ function MockPeer() {
     const self = this;
     const echos = { };
     const nulls = { };
+    const queue = [];
+    let pending = false;
+
     cockpit.event_target(this);
 
     /* These are events */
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
(Not adding no-test, because although the integration tests are useless here, I want to run `make check` in as many different configurations as possible).

martinpitt:
@allisonkarlitskaya : `make check` in image-prepare doesn't run the browser unit tests (only the C ones), as there is no browser on the images.

martinpitt (approved):
Ooh, thanks for fixing that! :heart:

allisonkarlitskaya:
> @allisonkarlitskaya : `make check` in image-prepare doesn't run the browser unit tests (only the C ones), as there is no browser on the images.

Ah, of course!  Oops.

martinpitt:
/bin/bash: line 1: pytest: command not found

Argh, we forgot to actually refresh the unit-test container after landing the previous pytest branch. And now it doesn't refresh any more until #17881 lands. So this blocked for a bit.

martinpitt:
#17881 landed, with refreshing the real container. So retrying.

PR #17883, opened 2022-11-07
Unify python testing, get qunit tests under coveragepy

allisonkarlitskaya:
...the latest crack.

 - [x] #17882

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -10,7 +10,6 @@ jobs:
           - { make: 'check-memory', cc: 'gcc', tag: 'latest' }
           - { make: 'distcheck', cc: 'clang', tag: 'latest' }
           - { make: 'distcheck', cc: 'gcc', tag: 'i386' }
-          - { make: 'pycheck', cc: 'gcc', tag: 'latest' }
           - { make: 'pytest-cov', cc: 'gcc', tag: 'latest' }
           - { make: 'check', cc: 'gcc', tag: 'latest' }
       fail-fast: false
--- a/pkg/base1/test-echo.js
+++ b/pkg/base1/test-echo.js
@@ -83,7 +83,7 @@ QUnit.test("fence", async assert => {
     const done = assert.async();
 
     // This is implemented in the C bridge, but not in Python.
-    if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz") {
+    if (await QUnit.mock_info("pybridge")) {
         assert.ok(true, "skipping on python bridge, not implemented");
         done();
         return;
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -45,20 +45,6 @@ SYSTEMD_CTYPES_STAMP = src/systemd_ctypes/__init__.py
 $(SYSTEMD_CTYPES_STAMP):
 	$(srcdir)/tools/systemd_ctypes
 
-CLEANFILES += cockpit-bridge.pyz
-cockpit-bridge.pyz: $(MANIFESTS) $(PYTHON_BRIDGE_FILES) $(SYSTEMD_CTYPES_STAMP)
-	@rm -rf tmp/pyz
-	@mkdir -p tmp/pyz
-	@cp -r $(srcdir)/dist $(srcdir)/src/cockpit $(srcdir)/src/systemd_ctypes/ tmp/pyz
-	$(AM_V_GEN) python3 -m zipapp --python /usr/bin/python3 --output $@ --main cockpit.bridge:main tmp/pyz
-	@chmod a+x $@
-
-.PHONY: pcc
-# Python Cockpit Client testing target
-pcc: cockpit-bridge.pyz
-	ln -vsf "$(realpath $<)" ~/.local/bin/cockpit-bridge
-	gapplication action org.cockpit_project.CockpitClient open-path '"=localhost"'
-
 # -----------------------------------------------------------------------------
 #  C
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -51,8 +51,8 @@ pyTESTS_HANGING = \
 pyTESTS = $(pyTESTS_PASSING) $(pyTESTS_FAILING)
 
 .PHONY: pycheck
-pycheck: cockpit-bridge.pyz
-	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE=./cockpit-bridge.pyz
+pycheck: $(SYSTEMD_CTYPES_STAMP)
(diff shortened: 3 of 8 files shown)


pkg/base1/test-echo.js
```diff
@@ -83,7 +83,7 @@ QUnit.test("fence", async assert => {
     const done = assert.async();
 
     // This is implemented in the C bridge, but not in Python.
-    if (await QUnit.mock_info("bridge") == "cockpit-bridge.pyz") {
+    if (await QUnit.mock_info("bridge") == "python3") {
```
martinpitt, line 86: Can we keep the unit tests work with either? Maybe in the simplest case with `.includes("py")`?

src/Makefile.am
```diff
@@ -51,8 +51,8 @@ pyTESTS_HANGING = \
 pyTESTS = $(pyTESTS_PASSING) $(pyTESTS_FAILING)
 
 .PHONY: pycheck
-pycheck: cockpit-bridge.pyz
-	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE=./cockpit-bridge.pyz
+pycheck: $(SYSTEMD_CTYPES_STAMP)
+	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE='python3 -m cockpit.bridge'
```
martinpitt: With that, we have no tests any more for the built .pyz. As long as we build that, there should IMHO still be at least one simple test that runs it. (I have some faint memory that you wanted to stop building it at some point, but don't we still need it to send over the wire in the injector?)

martinpitt: I.e. how about doing what you do, and additionally running `dist/base1/test-echo.html` with the .pyz?

pkg/base1/test-echo.js
allisonkarlitskaya: Sure.  I actually think maybe we should add a helper function for this, then.

Another option would be to export that as a boolean flag and move the logic to `./test-server` itself, closer to the implementation.

src/Makefile.am
allisonkarlitskaya: I actually want to drop the `.pyz`.

tools/webpack-jumpstart
```diff
@@ -170,9 +170,9 @@ unpack_from_cache "${tag}"
 
 # "${changed_files}" is empty unless --partial was given
 for file in ${changed_files}; do
-    # don't touch files that have nothing to do with webpack
-    if grep -q -F ${file} dist/*/Makefile.deps; then
-        message TOUCH "${file}"
-        touch "${file}"
-    fi
+    # remove any webpacks that mention this file
+    grep -l "${file}" dist/*/Makefile.deps | cut -f2 -d/ | while read -r webpack; do
```
martinpitt, line 174: srsly, selinux relabeling updates mtimes? that sounds wrong..

src/Makefile.am
```diff
@@ -7,63 +7,18 @@ sbin_PROGRAMS =
...
-	$(MAKE) check XFAIL_TESTS='$(pyTESTS_FAILING)' TESTS='$(pyTESTS)' COCKPIT_BRIDGE='python3 -m cockpit.bridge'
-
 # Will only be honoured if pytest-timeout plugin is installed
 export PYTEST_TIMEOUT = 120
 
 .PHONY: pytest
-pytest: $(SYSTEMD_CTYPES_STAMP)
-	cd '$(srcdir)' && pytest
+pytest: $(BUILT_SOURCES) $(MANIFESTS) $(SYSTEMD_CTYPES_STAMP)
+	$(MAKE) test-server
```
martinpitt, line 15: Why is this not just a standard rule dependency? If we need to mess with make's normal logic, this should get a comment why it's necessary.

src/Makefile.am
martinpitt: OK, but this PR doesn't, and HACKING even still advertises testing it. We also still have the pybridge scenario in test/run. Hence, I'd either like this PR to add a test-echo smoketest for the .pyz,  or drop it all in one go.

martinpitt (commented):
Thanks for figuring out the unit test failure! How subtle..

tools/webpack-jumpstart
allisonkarlitskaya: Honestly, this was my best guess.  But why wouldn't it?

src/Makefile.am
allisonkarlitskaya: Because `$(BUILT_SOURCES)` needs to be done strictly before we build the `test-server` due to the need to generate the mock dbus interfaces.  This is the same reason that automake is full of stanadard targets like:

```
all:  $(BUILT_SOURCES)
    $(MAKE) all-am
```

We might also seek to eliminate our use of `BUILT_SOURCES` by introducing manual dependency information for the places that use that header.  I kinda hate `BUILT_SOURCES`...

tools/webpack-jumpstart
martinpitt: Because it doesn't change the file contents,  just the inode. If I call `chmod` or `chown` on a file that doesn't change the mtime either (it *does* change ctime, but I thought make would look at mtime?)

src/Makefile.am
martinpitt: Ah, evil. Thanks for the explanation!

pkg/base1/test-echo.js
allisonkarlitskaya: Done.

tools/webpack-jumpstart
allisonkarlitskaya: I split this out to #17913 anyway.  We can discuss if it makes sense to keep it.

martinpitt:
The [unit test refresh](https://github.com/cockpit-project/cockpit/actions/runs/3462516623/jobs/5781508133) finished. Was that the last blocker here? Or something else? (Curiously the unit tests here didn't fail, even though they ran against the old container)

.github/workflows/unit-tests.yml
```diff
@@ -10,7 +10,6 @@ jobs:
           - { make: 'check-memory', cc: 'gcc', tag: 'latest' }
           - { make: 'distcheck', cc: 'clang', tag: 'latest' }
           - { make: 'distcheck', cc: 'gcc', tag: 'i386' }
-          - { make: 'pycheck', cc: 'gcc', tag: 'latest' }
```
martinpitt, line 13: I was going to complain "unit-tests-refresh.yml needs the same change", but turns out we never added it there in the first place. It should call pytest-cov, however (separate commit/PR).

martinpitt (approved):
That's quite nice, thanks! Note that test/run's `pybridge` scenario also still refers to the pyz, but I'm ok with cleaning that up in a follow-up. It's not really being used yet anyway.

PR #17892, opened 2022-11-08
test: extend timeout for CPU usage going back to idle

jelly:
In a recent PCP update pmie_check consumes a lot of CPU which interferes with waiting for idle CPU usage.

### DIFF
--- a/test/verify/check-system-info
+++ b/test/verify/check-system-info
@@ -818,7 +818,8 @@ password=foobar
         wait(lambda: progressValue(1) > 75)
         m.execute("pkill -e -f [c]at.*urandom")
         # should go back to idle usage
-        wait(lambda: progressValue(1) < 20)
+        # HACK: work around pmie CPU usage https://bugzilla.redhat.com/show_bug.cgi?id=2140572
+        wait(lambda: progressValue(1) < 20, tries=200)
 
         # memory: our test machines should use a reasonable chunk of available memory; MiB or GiB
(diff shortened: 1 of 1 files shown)


martinpitt:
@jelly: Bah, we apparently still have some [time zone confusion](https://artifacts.dev.testing-farm.io/d417d9c0-437b-46bf-be5b-0b6b2bfd510c/) in testServices... I thought you fixed that harder the other day. Anyway, retrying.

test/verify/check-system-info
```diff
@@ -818,7 +818,8 @@ password=foobar
         wait(lambda: progressValue(1) > 75)
         m.execute("pkill -e -f [c]at.*urandom")
         # should go back to idle usage
-        wait(lambda: progressValue(1) < 20)
+        # HACK: work around pmie CPU usage https://bugzilla.redhat.com/show_bug.cgi?id=2140572
+        wait(lambda: progressValue(1) < 20, delay=200)
```
martinpitt, line 822: Ugh, no -- that severely penalizes all other distros where this works. I think you mean `tries=200`.

martinpitt (changes requested):

test/verify/check-system-info
jelly: Woops, that was me thinking about raising the delay or tries...

martinpitt (approved):
Cheers!

PR #17895, opened 2022-11-09
workflows: Fix date format for Flathub release

martinpitt:
See build in https://github.com/flathub/org.cockpit_project.CockpitClient/pull/16 . This isn't sufficient to fix the release, but it was an obvious bug in the generated diff.


allisonkarlitskaya (approved):
oops, thanks!

PR #17896, opened 2022-11-09
flatpak: preserve xml declaration in installed metainfo

allisonkarlitskaya:
The flathub appdata validation requires an <?xml?> line.

If we want one of those, we need to pass the `xml_declaration=True` kwarg to `ElementTree.write()`.

To make sure we catch issues like this for ourselves, validate the in-tree metainfo before patching it, and validate the installed copy after patching.


martinpitt (approved):
Cheers! Not sure that we actually need to call `appstream-util validate` twice, but :man_shrugging: it should be fast enough.

PR #17897, opened 2022-11-09
test: enable stratis tests on Arch Linux

jelly:
On Arch Linux services are not automatically started so we have to start them in tests, as the tests reboot enable them as well.

### DIFF
--- a/test/verify/check-storage-stratis
+++ b/test/verify/check-storage-stratis
@@ -23,9 +23,15 @@ from storagelib import StorageCase
 from testlib import skipImage, test_main
 
 
-# TODO: Arch Linux has Stratis packages, tests need adjustments
-@skipImage("No Stratis", "debian-stable", "debian-testing", "ubuntu-stable", "ubuntu-2204", "arch")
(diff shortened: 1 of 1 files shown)


mvollmer (approved):

test/verify/check-storage-stratis
```diff
@@ -23,9 +23,13 @@ from storagelib import StorageCase
 from testlib import skipImage, test_main
 
 
-# TODO: Arch Linux has Stratis packages, tests need adjustments
-@skipImage("No Stratis", "debian-stable", "debian-testing", "ubuntu-stable", "ubuntu-2204", "arch")
+@skipImage("No Stratis", "debian-stable", "debian-testing", "ubuntu-stable", "ubuntu-2204")
 class TestStorageStratis(StorageCase):
+    def setUp(self):
+        if self.image == "arch":
+            super().setUp()
```
mvollmer, line 30: This needs to be called always, not just on Arch. :-)

mvollmer (changes requested):

mvollmer (approved):

PR #17898, opened 2022-11-09
FMF: RHEL fixes

martinpitt:
279 failed in RHEL 9 gating, due to some TF machine specific differences and a missing test dependency.

I tried to test this with a draft PR against RHEL, but tests don't run there. I asked, but I'm fairly sure about these fixes.


marusak (approved):
Danke!

PR #17910, opened 2022-11-14
login: Define autocomplete attribute for password field

marusak:


garrett:
This doesn't fix it. Chrome is broken. Changing this will make it autocomplete still. You won't see it on the first try, but it will show up in subsequent tries.

garrett:
https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off

marusak:
Yes, I did not set it to "off" as we want browsers to autocomplete. Also [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) states that:
```
In most modern browsers, setting autocomplete to "off" will not prevent a password manager from asking the user if they would like to save username and password information, or from automatically filling in those values in a site's login form. 
```

garrett:
Yes, I know it's not off; I'm talking about it being broken in general. And MDN isn't quite right when Chrome keeps changing things.

garrett:
I've built a VM and tested it in various browsers:

- Firefox: works as expected
- Chrome: does not work
- Epiphany: does not work

(Chrome works on my localhost instance w/o the PR)

Are these the results you're seeing?

marusak:
Closing in favor of https://github.com/cockpit-project/cockpit/pull/18330

PR #17911, opened 2022-11-14
webpack-make: include $(SRCDIR) in stampfile name

### LINKED ISSUE #17462: webpacks not being rebuilt during out-of-tree `make dist`
### Explain what happens

1. `tools/make-dist`
2. `touch pkg/static/login.css`
3. `tools/make-dist`

Notice that the `static` webpack doesn't get rebuilt.

I looked at this briefly, but I couldn't find the cause.

allisonkarlitskaya:
When writing the Makefile.deps, make sure we consistently refer to the fact that the manifest.json is created in the SRCDIR.  Otherwise, when doing a builddir != srcdir build, we won't consider the dependencies at all.

Fixes #17462

This was the real cause of the weird timestamp issues in #17883

### DIFF
--- a/tools/webpack-make
+++ b/tools/webpack-make
@@ -93,7 +93,7 @@ function process_result(err, stats) {
 }
 
 function generateDeps(makefile, stats) {
-    const stampfile = path.dirname(makefile) + '/manifest.json';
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thank you!

PR #17913, opened 2022-11-14
webpack-jumpstart --partial: delete changed webpacks instead of touching changed files

allisonkarlitskaya:
This was part of #17883 but the real issue there was #17462 (fixed in #17911).

I still think the changes to `webpack-jumpstart` are worthwhile, but now that they have absolutely nothing to do with the coveragepy branch anymore, we can discuss them here, instead.

### DIFF
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -40,7 +40,7 @@ done
 
 [ -n "${quiet}" ] || set -x
 
-if [ -e dist -a -z "${force}" ]; then
+if [ -e dist ] && [ -z "${force}" ]; then
     echo "jumpstart: dist/ already exists, skipping" >&2
     exit 1
 fi
@@ -100,7 +100,7 @@ if test -n "${partial}"; then
     # fetch the available commits, then iterate our history looking for the first match
     available="$(git_cache ls-remote origin 'refs/tags/sha-*' | cut -f2 -d-)"
     for commit in $(git log -n 100 --format=%H); do
-        if [ "${available%${commit}*}" != "${available}" ]; then
+        if [ "${available%"${commit}"*}" != "${available}" ]; then
             tag="sha-${commit}"
             break
         fi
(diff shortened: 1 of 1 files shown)


tools/webpack-jumpstart
```diff
@@ -170,9 +170,9 @@ unpack_from_cache "${tag}"
 
 # "${changed_files}" is empty unless --partial was given
 for file in ${changed_files}; do
-    # don't touch files that have nothing to do with webpack
-    if grep -q -F ${file} dist/*/Makefile.deps; then
-        message TOUCH "${file}"
-        touch "${file}"
-    fi
+    # remove any webpacks that mention this file
+    grep -l "${file}" dist/*/Makefile.deps | cut -f2 -d/ | while read -r webpack; do
```
martinpitt, line 174: I don't have a strong opinion either way. It feels a tad more robust to remove the whole webpack. I just think this should probably grow a `sort -u` to avoid showing the "REMOVE" message for the same webpack multiple times?

allisonkarlitskaya: I think that shouldn't be possible: The `grep -l` will display each filename at most once, and we won't hit the same file more than once because on the next iteration (to check the next changed file) the `Makefile.deps` that mentioned it will also have disappeared.

But it raises an interesting point: in theory, if all of the webpacks got deleted (like if some babelrc changed, or so) then, for the next file changed after that, our glob would fail to match, and we'd end up with errors about that.

martinpitt: I meant that if you change e.g. pkg/systemd/terminal.{jsx,scss} (which is quite common), it would go through that loop twice. But you are right that this concerns the outer loop, not the inner one, so we'd have the duplication either way.

allisonkarlitskaya: The point is that, on the second run through the loop, we run `grep` again.  Since we deleted the entire webpack in the previous iteration, we won't find it on the list of things that need to be deleted, the second time around, therefore no double message.

allisonkarlitskaya:
I added `grep -s` to suppress the message about the missing file in case the wildcard matches nothing:

https://github.com/cockpit-project/cockpit/compare/25ffae9210c3c77d55789863224cf53715e4923a..cb991b027dc12d3641b2e513291a8de87bb92c3e

tools/webpack-jumpstart
martinpitt: Oh, of course, sorry. Reviewing PRs during meetings, argh :brain:

PR #17919, opened 2022-11-15
autogen: Copy automake tools

martinpitt:
autogen: Copy automake tools

Building the dist tarball in the current Fedora 37 cockpit/tasks
container now fails when running as user podman container, when using
systemd-homed:

    cp: failed to preserve ownership for 'cockpit-279.39.gb4b41375f/tools/compile': Value too large for defined data type

This was due to a change in systemd 251 with mapped user IDs[1][2].

Work around this by copying the files instead of symlinking them. That
is the default behaviour anyway, and does not hurt much: git trees are 
regularly cleaned by developers anyway, and dist tarballs contain a copy
either way.

[1] https://github.com/systemd/systemd/blob/1d679b208d982bd5b8ba893981774cac5959b4b4/NEWS#L789
[2] https://bugzilla.redhat.com/show_bug.cgi?id=2144558

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -7,7 +7,7 @@ srcdir="${0%/*}"
 (
     cd "${srcdir}"
     echo "m4_define(VERSION_NUMBER, [$(git describe --tags --abbrev=0)+git])" > version.m4
-    autoreconf -is --warnings obsolete
+    autoreconf -i --warnings obsolete
 )
 
 [ -n "${NOCONFIGURE:-}" ] && exit


allisonkarlitskaya (commented):
This sounds like some kind of a bug in the container setup.  From the standpoint of the `tar` command running inside the container, this file should be owned by `root:root`, no?

Is there something else unusual going on here?

martinpitt:
It is owned by root, and I don't see anything unusual about it:
```
❱❱❱ stat /usr/share/automake-1.16/compile
  File: /usr/share/automake-1.16/compile
  Size: 7400      	Blocks: 24         IO Block: 4096   regular file
Device: 0,55	Inode: 6593784     Links: 1
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
```

I'm not sure how to debug this. I tried `make --debug` and `make distdir-am`, this is being done by something generated by automake; that `cp` does not come from our own code.

martinpitt:
FTR, I refreshed my toolbox to the latest tasks container from https://github.com/cockpit-project/cockpituous/pull/532 and this issue persists.

martinpitt:
I drilled down into this:
```
tools/make-dist

unset MAKEFLAGS
cd tmp/build-dist
make --debug dist
# Must remake target 'distdir-am'.
# if test -d "cockpit-280+git"; then find "cockpit-280+git" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "cockpit-280+git" || { sleep 5 && rm -rf "cockpit-280+git"; }; else :; fi
# test -d "cockpit-280+git" || mkdir "cockpit-280+git"
# cp: failed to preserve ownership for 'cockpit-280+git/tools/compile': Value too large for defined data type

make --debug distdir-am

strace -fvvs1024 -o /tmp/d make --debug distdir-am
```

The strace shows:
```
53838 clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fb4437cea10) = 54043
54043 execve("/usr/bin/cp", ["cp", "-p", "../../tools/compile", "cockpit-280+git/tools/compile"], [...]) = 0

54043 openat(AT_FDCWD, "cockpit-280+git/tools/compile", O_WRONLY|O_CREAT|O_EXCL, 0700) = 4
54043 ioctl(4, BTRFS_IOC_CLONE or FICLONE, 3) = -1 EXDEV (Invalid cross-device link)
54043 copy_file_range(3, NULL, 4, NULL, 9223372035781033984, 0) = -1 EXDEV (Invalid cross-device link)
54043 utimensat(4, NULL, [{tv_sec=1626057673, tv_nsec=0} /* 2021-07-12T04:41:13+0200 */, {tv_sec=1626057673, tv_nsec=0} /* 2021-07-12T04:41:13+0200 */], 0) = 0
54043 fchown(4, 0, 0)                   = -1 EOVERFLOW (Value too large for defined data type)

54043 write(2, ": Value too large for defined data type", 39) = 39
```

Indeed `cp -p tools/compile x` reproduces this. So this is a bug somewhere between `cp` not *actually* ignoring the error (as the manpage says "if possible"), glibc reporting a bogus error `EOVERFLOW` instead of, say, `EPERM` for the chown, and podman/toolbox doing something funky.

Outside of cockpit and automake stuff, the simplest reproducer is this:

```
❱❱❱ cp -p /usr/bin/cat x; echo $?
cp: failed to preserve ownership for 'x': Value too large for defined data type
1
```

Interestingly it works if the destination directory is not bind-mounted into the container:
```
❱❱❱ cp -p /usr/bin/cat /var/tmp/x; echo $?
0
```

Trying with podman directly:
```
mkdir ~/test-tmp
podman run -it -v $HOME/test-tmp:/h:z --rm registry.fedoraproject.org/fedora cp -p /usr/bin/cat /h/x
```

That works. But it fails in a similar way like this:
```
rm -f ~/test-tmp/x
podman run -it -v $HOME/test-tmp:/h:z -u $(id -u) --rm registry.fedoraproject.org/fedora cp -p /usr/bin/cat /h/x
# cp: cannot create regular file '/h/x': Value too large for defined data type
```

With toolbox it's a bit more straightforward:
```
toolbox create -r 37
toolbox run -r 37 -- cp -p /usr/bin/cat ~/test-tmp/x
# cp: failed to preserve ownership for '/home/martin/test-tmp/x': Value too large for defined data type
```

At this point I give up -- this is a bug somewhere between podman, coreutils, glibc, or the kernel.

I reported this to https://bugzilla.redhat.com/show_bug.cgi?id=2144558

allisonkarlitskaya:
Honestly, this feels like it's maybe vaguely kinda sorta expected.  You're trying to copy a file that belongs to the container into a directory on the host, and also requesting that its permissions be preserved.

On my system this works: on the outside the file ends up owned by uid 165536, which is the first number in my subuid range for my user in `/etc/subuid`.  That makes sense: as viewed from outside of the container, the `uid_map` of processes inside of the container looks like:

```
⬢[root@toolbox lis]# sleep inf  # we need to view an "inside" process from "outside"
```

```
[lis@x1 ~]$ cat /proc/"$(pgrep -f 'sleep inf')"/uid_map
         0     165536       1001
      1001       1001          1
      1002     166537      64535
```


Can you check that your uid map shows something similar, and confirm that manually creating a file with that UID on the outside of any container actually works?

Either as root, on the outside:
```
# chown 165536:165536 ~/cat
```

or using `podman unshare` on the outside (which has a different default UID mapping scheme):

```
$ podman unshare chown 1:1 cat
```

Both of those are working for me, and producing the expected result (ie: file owned by 165536:165536, as viewed from the host).

All of the volume mounts you're trying to use seem to be inside of your home directory, and I have a sneaking suspicion that the real culprit here is some uid mapping issue with systemd-homed, preventing podman from creating files that it ought to be able to create.

allisonkarlitskaya:
> Why would it, though? UID "100000" isn't really a thing on my host system, especially not on-disk. It's just the uid that uid 1000 processes in podman would have on the host, but that shouldn't apply to fsuids?

UID 100000 is absolutely a thing on your host system, and I bet it works in any directory that's not your home directory...

I recall Lennart ranting at some point about how subuid-based user containers were a bad idea, so I wouldn't be surprised to find out that systemd-home simply doesn't support the subuid range.

In any case, I'm very much leaning towards: this is not our bug.

allisonkarlitskaya:
I'd be interested in seeing the mount options for your home directory, btw.  Maybe it says something about uid mapping in `/proc/mounts`.  That might make this all very clear, very quickly.

I know that systemd-homed was doing, effectively `chmod -R` on the home directory at login, at some point, and that they had plans to move over to a uid mapping approach.

allisonkarlitskaya:
Ah, indeed: https://github.com/systemd/systemd/blob/main/NEWS


>   CHANGES WITH 251:
>   ...
>         Changes in systemd-homed:
> 
>         * Starting with v250 systemd-homed uses UID/GID mapping on the mounts
>           of activated home directories it manages (if the kernel and selected
>           file systems support it). So far it mapped three UID ranges: the
>           range from 0…60000, the user's own UID, and the range 60514…65534,
>           leaving everything else unmapped (in other words, the 16bit UID range
>           is mapped almost fully, with the exception of the UID subrange used
>           for systemd-homed users, with one exception: the user's own UID).
>           Unmapped UIDs may not be used for file ownership in the home
>           directory — any chown() attempts with them will fail. With this
>           release a fourth range is added to these mappings:
>           524288…1879048191. This range is the UID range intended for container
>           uses, see:
> 
>                   https://systemd.io/UIDS-GIDS
> 
>           This range may be used for container managers that place container OS
>           trees in the home directory (which is a questionable approach, for
>           quota, permission, SUID handling and network file system
>           compatibility reasons, but nonetheless apparently commonplace). Note
>           that this mapping is mapped 1:1 in a pass-through fashion, i.e. the
>           UID assignments from the range are not managed or mapped by
>           `systemd-homed`, and must be managed with other mechanisms, in the
>           context of the local system.
> 
>           Typically, a better approach to user namespacing in relevant
>           container managers would be to leave container OS trees on disk at
>           UID offset 0, but then map them to a dynamically allocated runtime
>           UID range via another UID mount map at container invocation
>           time. That way user namespace UID ranges become strictly a runtime
>           concept, and do not leak into persistent file systems, persistent
>           user databases or persistent configuration, thus greatly simplifying
>           handling, and improving compatibility with home directories intended
>           to be portable like the ones managed by systemd-homed.

martinpitt:
> UID 100000 is absolutely a thing on your host system, and I bet it works in any directory that's not your home directory...

Ah, I meant that it doesn't have any particular meaning. Of course one can sudo chown a file to any arbitrary uid. But yes, this works:

    touch /home/martin-cache/somefile; sudo chown 100000:100000 /home/martin-cache/somefile

/home/martin-cache/ is outside of homed/LUKS.

> mount options for your home directory

on the host:

    /dev/mapper/home-martin on /var/home/martin type ext4 (rw,nosuid,nodev,relatime,idmapped,seclabel)

More or less the same in the container:

   /dev/mapper/home-martin on /var/home/martin type ext4 (rw,nosuid,nodev,relatime,idmapped,seclabel)

(/home is just a symlink to /var/home on ostree).

> this is not our bug.

Oh, I never thought it was. This is just a workaround that reduces the assumptions on the file system, and I don't see a disadvantage in copying the files instead of symlinking (that's the default behaviour anyway).

Thank you for your investigations!

allisonkarlitskaya (approved):
as promised: :+1:

martinpitt:
Ack, I updated the commit message. Also, you pretty much talked me into going back to unencrypted home directories and full disk encryption -- I still prefer per-user encrypted home dirs, but given the "PoC" quality of systemd-homed in Fedora in the last years it's probably time to let it go. :cry:

martinpitt:
@allisonkarlitskaya To conclude the saga, I changed `/etc/sub[ug]id` to `martin:524288:65536`, and things work well again. I also updated the bugzilla accordingly. Do you want me to revert this?

PR #17922, opened 2022-11-16
test: build a wheel for the pybridge scenario

allisonkarlitskaya:
Since we dropped the Makefile rule for creating cockpit-bridge.pyz, the pybridge scenario has been broken.

Add some minimal pyproject.toml metadata required to build a wheel of the Python package.  Build this wheel on the outside, and inject it into the VM via image-customize.

Note: we no longer bother with building the C parts of the package anymore — we rely on the distro packages already present in the image.

 - [x] cockpit-project/cockpituous#532
 - [x] cockpit-project/bots#4068  (one way or the other)

### DIFF
--- a/test/run
+++ b/test/run
@@ -25,9 +25,10 @@ case $TEST_SCENARIO in
         test/common/run-tests --jobs ${TEST_JOBS:-1} --test-dir test/verify --track-naughties
         ;;
     pybridge)
+        tools/systemd_ctypes
         test/image-prepare --verbose $TEST_OS
-        make -C tmp/build-dist/ cockpit-bridge.pyz
-        bots/image-customize -v --upload tmp/build-dist/cockpit-bridge.pyz:/usr/bin/cockpit-bridge $TEST_OS
+        python3 -m build --no-isolation --wheel --outdir tmp/wheel
+        bots/image-customize --verbose --no-network --install tmp/wheel/*.whl "${TEST_OS}"
         test/common/run-tests --jobs ${TEST_JOBS:-1} --test-dir test/verify --track-naughties
         ;;
     *)
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ Makefile.in
 /aclocal.m4
 /autom4te.cache
 /bots
+/build/
 /cockpit
 /cockpit-*.tar*
 /cockpit-*.xz
@@ -71,6 +72,7 @@ Makefile.in
 /pkg/*/*.metainfo.xml
 /po/*.pot
 /src/bridge/org.cockpit-project.cockpit-bridge.policy
+/src/cockpit.egg-info/
 /src/common/fail.html.c
 /src/systemd/cockpit*.service
 /src/systemd/cockpit*.socket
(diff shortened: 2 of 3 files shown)


test/run
```diff
@@ -25,9 +25,15 @@ case $TEST_SCENARIO in
...
-        make -C tmp/build-dist/ cockpit-bridge.pyz
-        bots/image-customize -v --upload tmp/build-dist/cockpit-bridge.pyz:/usr/bin/cockpit-bridge $TEST_OS
+        rm -rf tmp/wheel
+        tools/systemd_ctypes
+        python3 -m build --no-isolation --wheel --outdir tmp/wheel
+        bots/image-customize --verbose \
+            --fresh \
+            --install python3-pip \
+            --upload tmp/wheel:/var/tmp \
+            --run-command 'pip install --prefix=/usr /var/tmp/wheel/*' \
```
martinpitt, line 35: Why do we have to call pip install inside the image? Isn't it enough to just copy it to some place in `$PYTHONPATH`? Also, this needs some extra commands to ensure that this does not access the internet. In fact `image-customize` option` --no-network` is missing in a lot of places here in test/run (we added it to e.g. podman and machines, but apparently forgot it in cockpit?)

martinpitt: See discussion in https://github.com/cockpit-project/bots/pull/4068 . I'm happy enough with `--no-network` after landing that. Until then, good enough.

allisonkarlitskaya:
Successfully builds the image for `/pybridge` scenario again (and let's us get to the sea of red).

martinpitt (approved):
Hmm.. Admittedly I liked the .pyz a lot more than all this stuff, but :man_shrugging:

allisonkarlitskaya:
> Hmm.. Admittedly I liked the .pyz a lot more than all this stuff, but man_shrugging

The .pyz is pretty far removed from the concept of "let's test things as close as possible to how they'll be used in production".  Nobody was ever expecting that installing a giant .pyz in /usr/bin would be how we handle a complex Python program.  I also feel like "all this stuff" is a bit unfair — it's actually a very small amount of code.

In terms of pushing towards how this will be done eventually, I feel like this is definitely the right thing to do.

martinpitt:
> Nobody was ever expecting that installing a giant .pyz in /usr/bin

Of course not in /usr/bin -- our rpm/deb should just install stuff into the default OS python path, using the packaging helpers. I meant for "cockpit-less" target machines where the bridge gets sent through the wire. I understand that you have a different method than .pyz for that (the clever importlib/dict thing). But this cannot use pip either. That's what I'm grumbling about -- this setup here is a special case for CI which doesn't reflect either production mode.

Aber nichts für ungut, we can always change this stuff once it matures a bit :grin:

allisonkarlitskaya:
My intent with this PR is to very closely model what the system will look like once we properly package a python `cockpit-bridge` RPM.  Specifically:

 - python files installed in the standard python path
 - `cockpit-bridge` wrapper in `/usr/bin`

As we know, the RPM macros are using `pip` internally to accomplish this, as well.  I only do the direct invocation myself to:
 - avoid having to create a separate .spec file for creating a `cockpit-bridge` rpm containing the python bridge
 - avoiding modifying the existing .spec before we're "ready"

Also, for the injector case, well.... let's just say I actually intended to use the `.pyz` until I got stuck on an extra long train trip this weekend... the importlib/dict approach is suddenly looking better, indeed ;)

Issue #17596, opened 2022-07-26
Service kebab menu is cut off by menu


LunarTwilight:
### Explain what happens

1. go to a systemd service (eg `/system/services#/test.service`)
2. click on the kebab menu next to the name
3. observe the kebab menu is cut off due to it being left aligned (changing it in my dev tools to be right aligned fixes it)

![image](https://user-images.githubusercontent.com/13520392/181012902-1833db24-2b27-4b46-802d-542098e7f465.png)

### Version of Cockpit

264-1~bpo20.04.1

### Where is the problem in Cockpit?

Services

### Server operating system

Ubuntu

### Server operating system version

20.04.4 LTS

### What browsers are you using?

Chrome

garrett:
Thanks for reporting! 

I was able to reproduce this in both Chrome and Firefox with short systemd service names. It's important to note that at least on Fedora, none of the default services have names this short to see this problem.

The menu would be cut off on the right for long names and/or mobile mode if we changed the alignment.

Here's a screenshot in mobile, with a standard name:

![image](https://user-images.githubusercontent.com/10246/181038431-1d8d9d01-2664-4cfb-ab46-369cbf63cba6.png)

I do have a test service and it does happen in mobile also, as the name is so short:

![image](https://user-images.githubusercontent.com/10246/181038872-8eb4c0d8-e859-49d0-8d44-4b9ffa309777.png)

We could fix this for mobile, by always aligning the menu icon to the right. For desktop, it's trickier. A simple, but not correct, approach would be to add a min-width for the text, but then it's going to break like the top screenshot in various languages.

If we combine this with breakpoint alignment in the `<Dropdown>` component, we could have the position the same on desktop modes and right-aligned in mobile, then have       

```jsx
<Dropdown
  alignments={{
    default: 'right',
    sm: 'left'
  }}
  ...
>
```

To align it, we would add some CSS, like this, for a mobile-only breakpoint:

```css
.service-top-panel .pf-c-dropdown {
  flex: auto;
  justify-content: end;
}
```

I think this can be done with PatternFly utilities:

`pf-u-flex-grow-on-sm pf-u-justify-content-flex-end-on-sm`

End result would look like this in mobile:

![image](https://user-images.githubusercontent.com/10246/181042567-8b0a9479-c7f3-43e0-8b0f-ca086b68e4f1.png)

[...]

PR #17926, opened 2022-11-16
systemd: remove position="right" for service dropdown

jelly:
This causes issue on mobile and with very tiny service names which makes the dropdown disappear under the sidebar menu on mobile and small screens.

See https://github.com/cockpit-project/cockpit/issues/17596#issuecomment-1317365314 for screenshots

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -187,7 +187,6 @@ const ServiceActions = ({ masked, active, failed, canReload, actionCallback, del
                   isOpen={isActionOpen}
                   isPlain
                   onSelect={() => setIsActionOpen(!isActionOpen)}
-                  position='right'
                   dropdownItems={actions} />
     );
 };


garrett (changes requested):
No, this isn't the right fix. I explained it in the bug @ https://github.com/cockpit-project/cockpit/issues/17596#issuecomment-1195615379, with a workaround for us, based on brreakpoints.

The proper fix is for PatternFly to fix their auto layout. They've been working on that.

Meanwhile, we need one alignment for desktop and another for mobile.

JSX:
```
<Dropdown
  alignments={{
    default: 'right',
    sm: 'left'
  }}
  ...
>
```

But then a little bit of CSS to improve things too:

```scss
.service-top-panel .pf-c-dropdown {
  flex: auto;
  justify-content: end;
}
```

But this CSS could be in the className in JSX with helper classes:
`pf-u-flex-grow-on-sm pf-u-justify-content-flex-end-on-sm`

pkg/systemd/services/service-details.jsx
```diff
@@ -187,7 +187,6 @@ const ServiceActions = ({ masked, active, failed, canReload, actionCallback, del
                   isOpen={isActionOpen}
                   isPlain
                   onSelect={() => setIsActionOpen(!isActionOpen)}
-                  position='right'
                   dropdownItems={actions} />
```
garrett, line 190:
```suggestion
                  className='pf-u-flex-grow-on-sm pf-u-justify-content-flex-end-on-sm'
                  alignments={{ default: 'right', sm: 'left' }}
                  dropdownItems={actions} />
```

garrett: It should probably be something like this for a proper fix. Screenshots in the comment I posted on the issue.

garrett:
Oh, I forgot that PF changed auto placement. Whoops! Checking out this PR to test it locally. A simpler fix is almost always better, right? :grin:

garrett (changes requested):
Still broken for me on mobile...

image

jelly:
> Still broken for me on mobile...
> 
> image

Right, I only tested it with a small unit name.. sorry!

KKoukiou:
This solution wont' work as is. Closing to clean up PR list, and you can always reopen when you work on this.

PR #17927, opened 2022-11-17
test: Update pixel references to current Chromium

martinpitt:
Fedora got updated in
https://github.com/cockpit-project/cockpituous/pull/532 , bringing along
updates to font rendering. This caused a lot of jitter, requiring all 
pixel tests to be re-done.

Unfortunately the jitter is completely unpredictable/unreproducible in
the Test{AD,IPA}.testQualifiedUsers join/leave dialogs, each run ends up
with a slightly different image. Disable these for now.

----

[review diff](https://github.com/cockpit-project/pixel-test-reference/compare/19b97eb573a874ef8f15b73a85e6636f7b648f07..77b25f0b21eb5082d51802180db9676c8a3b7afc)

### DIFF
--- a/test/verify/check-system-realms
+++ b/test/verify/check-system-realms
@@ -128,7 +128,8 @@ class CommonTests:
         # admin gets auto-detected
         b.wait_val(self.op_admin, self.admin_user)
         b.set_input_text(self.op_admin_password, self.admin_password)
-        b.assert_pixels("#realms-join-dialog", "realm-join")
+        # FIXME: there is tons of subpixel noise in the fonts, impossible to match with our naïve algorithm
+        # b.assert_pixels("#realms-join-dialog", "realm-join")
         b.click(f"#realms-join-dialog button{self.primary_btn_class}")
         # running operation cannot be cancelled any more
         b.wait_visible("#realms-join-dialog button.pf-m-link:disabled")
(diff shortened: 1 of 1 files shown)


martinpitt:
Uh, [two more](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17927-20221117-034107-4020f0b8-fedora-36/log.html#277) which didn't happen before

martinpitt:
Meh, [some jitter here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17927-20221117-062625-d5f03930-fedora-36/log.html). Re-running to compare.

martinpitt:
I ran `TestAD.testQualifiedUsers` a few times locally, and there is just no pleasing this -- *every* run has different jitter. This needs a bigger hammer, or we need to disable this pixel test for now.

martinpitt:
[argh argh argh another one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17927-20221117-114034-214da023-fedora-36/log.html#277)

martinpitt:
NOOooooo....  this just hates me, it [failed again](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17927-20221117-114034-214da023-fedora-36/log.html#315). There is some unpredictability in Fedora's font rendering now, and our pixel matching is too rigid for that.

martinpitt:
I experimented for a while with making the fuzz factors (color and spatial) larger, but that doesn't help -- there are many hundred pixels which are "wrong", even though it's not noticeable with the bare eye (only on zoom). I disabled the two assert_pixels() in that test case for now, this needs a scientific/mathematical solution, not a cowboy one.

@mvollmer, perhaps you have some idea? Some library that can do this kind of fuzzy matching? Or perhaps reducing to a 16 bit palette, or some edge detection or so?

KKoukiou (approved):

PR #17932, opened 2022-11-18
ssh: Update mock-sshd to new example server code

mvollmer:


mvollmer:
This starts to work.  The large echo test hangs for me, but so does it with the old mock-sshd.  Will debug later.

The basic approach right now is to change test-sshbridge to use the existing arguments of the new example server, in order to have a smaller diff from that example code.

mvollmer:
The biggest changes to the example code:

 - interactive auth needs to be done from a generic message callback, but pubkey auth needs to be done from a specific auth_pubkey callback. Thus we need to install both.
 - Writing to the child stdin might block, so we need to poll for writability

martinpitt:
Nice, thanks @mvollmer! This pretty much validates itself. Apart from that silly

    # src/ssh/mock-sshd.c: #include "config.h" is not the first line

static-code failure, it seems to be happy! :bow:

martinpitt (approved):
Many thanks for this! That now means that we can strengthen our unit test runs to build with `--enable-strict`.

PR #17934, opened 2022-11-18
Enable mounting of /etc/cockpit as volume

realk1ko:
A rather small change that would allow either `/etc/cockpit` or `/etc/cockpit/cockpit.conf` to be mounted via volume options for the WS container.
This should not break existing container deployments, since the link creation for the `default-bastion.conf` still is done if both the directory and the file are not mounted on start.


martinpitt (approved):
Thank you! On landing I will prefix the commit message with "containers/ws:".

PR #17935, opened 2022-11-19
Multi-arch builds for WS container and other improvements

realk1ko:
This PR changes the image tags for the WS container and adds ARM64 builds via Docker Buildx or podman buildx. I've tried to keep changes as minimal as possible, since I don't have a good enough overview to fully understand the workflows on this repository and want to avoid any collateral damage.

I'm not fully satisfied with this solution and I have a few open topics I'd like to take care of in the next few days or weeks, hence the draft status:

- [ ] I've not been able testing to push to quay.io itself, only to my private packages on the GHCR. *pitti*: That's fine, I don't expect any trouble there, and if there is I'm happy to deal with it once this is ready to land.
- [ ] The smoke test currently only tests the first built image. *pitti* See review comment thread below.
- [ ] I've not tested the local RPM installation due to lack of time, I'd appreciate help on that as well. *pitti* our "fedora-coreos" integration test does that.
- [x] I'm not fully aware of how the packages are published to the Fedora repositories, but I suppose that in rare cases the available x86 version of Cockpit is different than the arm64 version when a new Cockpit package was just released. *pitti*: No, a new version always enters Fedora for all architectures at the same time. This is fine.
- [ ] The image itself could be a little "optimized" by reducing the files added to the container. For instance it's not necessary to include the release scripts as well as the README in the container. *pitti* Let's do this in a separate PR, to not make this unnecessarily complex.
- [x] The README for the container (also displayed on quay.io) is missing a few things I'd like it to point out, e. g. an overview of the available image tags and some info on the SSH key authentication.
- [x] The [Docker Hub page for the container](https://hub.docker.com/r/cockpit/ws) should be updated to inform of the deprecation of the images there and link to quay.io *pitti*: Done, thanks for pointing out!

The release script itself can be run via Docker or podman, that's the main reason this has been rather cumbersome to implement. **If that's not a requirement I'd be happy to simplify the workflow.** It should be possible to replace most if not all of the image tagging/multi-arch stuff currently in the script via GitHub actions provided by Docker. *pitti*: I'm a bit wary of dropping docker; we used to have problems in the past that old docker versions could not read podman-generated images (IIRC because docker was not able to read the new manifest format or something such). So for now, let's keep building the official images with docker.

Also a side note: Docker Buildx is pretty slow when building an ARM image on x86. It took about 30 minutes for ARM image to be built when I tested this. *pitti*: Yup, that's QEMU full system emulation for you :grin:

### DIFF
--- a/containers/ws/README.md
+++ b/containers/ws/README.md
@@ -1,99 +1,153 @@
-# Cockpit on Fedora CoreOS or other container hosts
+# Cockpit WS Container
 
-The standard Fedora and Red Hat Enterprise Linux CoreOS images does not contain
-Cockpit packages.
+This container is suitable for scenarios where you can not or wish not to deploy the Cockpit Web Service and browser
+application natively.
 
-1. Install Cockpit packages as overlay RPMs:
+## Tags
+
+The following tags are provided for the image:
+
+- `:latest`, `:latest-amd64`, `:latest-arm64` point to the latest images and their respective architecture
+- `:{Cockpit WS Version}`, `:{Cockpit WS Version}-amd64`, `:{Cockpit WS Version}-arm64` (e. g. `:280-amd64`) point to
+  images with the specified Cockpit WS version and their respective platform
+
+## Preparing Managed Host
+
+The hosts you want to manage via this container need to have both `cockpit-system` installed and SSH access enabled.
+
+Depending on your configuration, you may want to install
+[additional extensions](https://packages.fedoraproject.org/search?query=cockpit-) as well, such as `cockpit-kdump` or
+`cockpit-networkmanager`.
+
+The standard Fedora and Red Hat Enterprise Linux CoreOS images do not provide Cockpit packages. If you wish to install
+the container on CoreOS or simply manage such a system, you need to follow these
+steps to setup Cockpit:
+
+1. Install Cockpit core packages as overlay RPMs:
    ```
    rpm-ostree install cockpit-system cockpit-ostree cockpit-podman
    ```
-
-   Depending on your configuration, you may want to use
-   [other extensions](https://apps.fedoraproject.org/packages/s/cockpit-) as
-   well, such as `cockpit-kdump` or `cockpit-networkmanager`.
 
    If you have a custom-built OSTree, simply include the same packages in your build.
 
 2. Reboot
 
-Steps 1 and 2 are enough when the CoreOS machine is only connected to through another host running Cockpit.
+3. Continue with setting up the Cockpit WS container below
 
-If you want to also run a web server to log in directly on the CoreOS host, you
-can use this container in two modes.
+## Running the Cockpit WS Container
 
-## Privileged ws container
+There are two modes in which this container can operate.
 
-Privileged mode provides tight integration with the host: You can use the
-host's users and their passwords to log in, and Cockpit's branding will adjust
-to the host operating system.
+### Privileged Mode
 
-1. Enable password based SSH logins, unless you only use [SSO logins](https://cockpit-project.org/guide/latest/sso.html):
-   ```
-   echo 'PasswordAuthentication yes' | sudo tee /etc/ssh/sshd_config.d/02-enable-passwords.conf
-   sudo systemctl try-restart sshd
-   ```
+Privileged mode provides tight integration with the host. You can use the host's users and their passwords to log in,
+and Cockpit's branding will adjust to the host operating system.
 
-2. Run the Cockpit web service with a privileged container (as root):
+1. Run the Cockpit web service with a privileged container (as root):
    ```
    podman container runlabel --name cockpit-ws RUN quay.io/cockpit/ws
    ```
 
-3. Make Cockpit start on boot:
+2. Make Cockpit start on boot:
    ```
    podman container runlabel INSTALL quay.io/cockpit/ws
    systemctl enable cockpit.service
    ```
 
-Afterward, use a web browser to log into port `9090` on your host IP address as usual.
+### Unprivileged Mode (Bastion Container)
 
-## Unprivileged bastion container
-
-When you run the container without `--privileged`, it presents an unbranded
-login page. The user always has to specify a host name, and it connects to that
-host with SSH.
+When you run the container without `--privileged`, it presents an unbranded login page. The user always has to specify a
+host name, and it connects to that host with SSH.
 
 ```
-podman run -d --name cockpit-bastion -p 9090:9090 quay.io/cockpit/ws
+podman run -d \
+   --name cockpit \
+   -p 9090:9090 \
+   quay.io/cockpit/ws
 ```
 
-This mode is suitable for deploying to e.g. Kubernetes or similar environments
-where you cannot have or want privileged containers. In this "bastion host
-mode", you can get a Cockpit for servers in your data center without opening an
-extra port for cockpit-ws on them.
+You can still log into the container's host if you specify its **public** address or alternatively
+`host.containers.internal` if you're running the container with [podman](https://podman.io).
 
-You can still log into the ws container's host if you specify its *public*
-address. [podman](https://podman.io/) provides the special name
-`host.containers.internal` for that.
+This mode is suitable for deploying to e.g. Kubernetes or similar environments where you can not have or want privileged
+containers. In this "bastion" mode you can run Cockpit without ensuring you can connect to the managed hosts directly.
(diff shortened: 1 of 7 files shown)


containers/ws/release.sh
```diff
@@ -8,24 +8,52 @@ if [ -z "${RUNC:-}" ]; then
...
-$RUNC rm -f $name
+version_arch_tags=""
+for arch in $ARCHS; do
+    release_arch_tag="quay.io/cockpit/ws:release-${arch}"
+    version_arch_tag="quay.io/cockpit/ws:${version}-${arch}"
+    latest_arch_tag="quay.io/cockpit/ws:latest-${arch}"
+
+    version_arch_tags="${version_arch_tags} ${version_arch_tag}"
+
+    $RUNC tag "${release_arch_tag}" "${version_arch_tag}"
```
martinpitt, line 46: Hmm, on dockerhub there was some convention to put the architecture in the repo name, like `docker.io/arm64v8/ubuntu:latest`. But indeed that doesn't seem to work on quay.io; we could create a new `quay.io/cockpit/ws-arm46` repo,  but including it into the tag seems okay. </thinking_aloud>

containers/ws/release.sh
```diff
@@ -8,24 +8,52 @@ if [ -z "${RUNC:-}" ]; then
...
+done
 
 # smoke test
-name=ws-release
-$RUNC run --name $name -p 19999:9090 -d quay.io/cockpit/ws:release
-until curl --fail --show-error -k --head https://localhost:19999; do
-    sleep 1
+first_arch=$(echo "${ARCHS}" | cut -d " " -f 1)
+test_container=ws-release
+$RUNC run --name "${test_container}" -p 19999:9090 -d "quay.io/cockpit/ws:release-${first_arch}"
```
martinpitt, line 25: This `first_arch` feels too arbitrary to me. At least on my laptop I cannot actually run arm64 binaries -- perhaps podman has some integration with qemu-system-arm or so, but I don't have that installed. Either the GitHub runners can do that, then the smoke test should run for all images. Or this should hardcode `release-amd64` as that's then the only one which we can test anyway.

containers/ws/release.sh
```diff
@@ -8,24 +8,52 @@ if [ -z "${RUNC:-}" ]; then
...
+
+for i in $(seq 1 30); do
+    curl --fail --show-error -k --head https://localhost:19999 && break
+
+    [ "${i}" -eq 30 ] && $RUNC rm -f "${test_container}" && exit 1 || sleep 1
```
martinpitt, line 30: This chaining of && and || is rather brittle, it duplicates the `rm` command and the retry count. If this times out, then I deliberately did *not* remove the container, as that makes interactive debugging a lot simpler.  And we don't have to clean it up on a GitHub runner.

I would just keep the previous indefinite loop and let the workflow timeout take care of it, but if you want to make this explicit, please either consider counting down (`seq 30 -1 0`) or set a `success=1` flag or so.

containers/ws/release.sh
```diff
@@ -8,24 +8,52 @@ if [ -z "${RUNC:-}" ]; then
...
+    $RUNC push "${version_arch_tag}"
+    $RUNC push "${latest_arch_tag}"
+
+    $RUNC rmi "${release_arch_tag}"
+done
 
-$RUNC tag quay.io/cockpit/ws:release quay.io/cockpit/ws:$TAG
-$RUNC tag quay.io/cockpit/ws:release quay.io/cockpit/ws:latest
-$RUNC rmi quay.io/cockpit/ws:release
+version_tag="quay.io/cockpit/ws:${version}"
```
martinpitt, line 55: I don't see where this tag gets defined? I.e. there's something missing here like

    $RUNC tag "quay.io/cockpit/ws/release-amd64" "$version_tag"

? Same thing for the `:latest` tag

containers/ws/label-install
```diff
@@ -1,4 +1,5 @@
-#!/bin/sh -eu
+#!/bin/sh
+set -eu
```
martinpitt, line 2: These kind of changes are good, but unrelated -- would be nice to split into a separate commit. Please prefix them all with "cockpit/ws:".

containers/ws/README.md
```diff
@@ -1,99 +1,153 @@
-# Cockpit on Fedora CoreOS or other container hosts
+# Cockpit WS Container
 
-The standard Fedora and Red Hat Enterprise Linux CoreOS images does not contain
-Cockpit packages.
+This container is suitable for scenarios where you can not or wish not to deploy the Cockpit Web Service and browser
+application natively.
```
martinpitt, line 4: "you cannot or do not want to deploy..". Let's please also drop the "browser application" -- we don't use this term anywhere, and it's not even clear what that should mean. The browser talks to the cockpit web server.

"natively" also sounds a bit vague -- maybe "through distribution packages (rpm/deb)"?

containers/ws/README.md
```diff
@@ -1,99 +1,153 @@
...
+- `:{Cockpit WS Version}`, `:{Cockpit WS Version}-amd64`, `:{Cockpit WS Version}-arm64` (e. g. `:280-amd64`) point to
+  images with the specified Cockpit WS version and their respective platform
+
+## Preparing Managed Host
+
+The hosts you want to manage via this container need to have both `cockpit-system` installed and SSH access enabled.
+
+Depending on your configuration, you may want to install
+[additional extensions](https://packages.fedoraproject.org/search?query=cockpit-) as well, such as `cockpit-kdump` or
+`cockpit-networkmanager`.
```
martinpitt, line 20: Please move this back down after the "install cockpit core packages" step, it IMHO makes more logical sense that way.

containers/ws/README.md
```diff
@@ -1,99 +1,153 @@
...
 
-   Depending on your configuration, you may want to use
-   [other extensions](https://apps.fedoraproject.org/packages/s/cockpit-) as
-   well, such as `cockpit-kdump` or `cockpit-networkmanager`.
-
    If you have a custom-built OSTree, simply include the same packages in your build.
 
 2. Reboot
 
-Steps 1 and 2 are enough when the CoreOS machine is only connected to through another host running Cockpit.
```
martinpitt, line 19: This piece of information should stay. In fact, we actively promote to *not* install webservers willy-nilly, and prefer SSH mode whenever possible.

containers/ws/README.md
```diff
@@ -1,99 +1,153 @@
...
-Privileged mode provides tight integration with the host: You can use the
-host's users and their passwords to log in, and Cockpit's branding will adjust
-to the host operating system.
+### Privileged Mode
 
-1. Enable password based SSH logins, unless you only use [SSO logins](https://cockpit-project.org/guide/latest/sso.html):
-   ```
-   echo 'PasswordAuthentication yes' | sudo tee /etc/ssh/sshd_config.d/02-enable-passwords.conf
-   sudo systemctl try-restart sshd
-   ```
```
martinpitt, line 34: This is crucial and must stay.

containers/ws/default-bastion.conf
```diff
@@ -1,6 +1,6 @@
 [WebService]
 ProtocolHeader = X-Forwarded-Proto
-LoginTitle = Cockpit Bastion
+LoginTitle = Cockpit
```
martinpitt, line 3: Okay -- but please split this out into a separate commit "cockpit/ws: Change bastion login title", as it's independent from the README updates.

containers/ws/README.md
```diff
@@ -1,99 +1,153 @@
-# Cockpit on Fedora CoreOS or other container hosts
+# Cockpit WS Container
 
-The standard Fedora and Red Hat Enterprise Linux CoreOS images does not contain
-Cockpit packages.
+This container is suitable for scenarios where you can not or wish not to deploy the Cockpit Web Service and browser
+application natively.
 
-1. Install Cockpit packages as overlay RPMs:
+## Tags
```
martinpitt, line 6: Adding this section should happen in the previous commit which adds these builds/tags, as these belong together. The other README changes should indeed stay a separate commit.

martinpitt (changes requested):
Thank you, much appreciated! I did a first review round and also updated the description to reply to your outstanding issues.

martinpitt:
This would indeed be nice to do, but requires a lot more work and is stale -- like actually testing the ARM image. This would be possible on TF, but let's be honest: With our ever-shrinking team we won't get to it anytime soon. If this ever comes up, we have this as an archive. Thanks, and sorry!

PR #17938, opened 2022-11-21
pkg: correct indexOf function call in cockpit.location.encoding

jelly:
The originally introduced cockpit.location.encode function incorrectly does the 0 comparison in the indexOf call, so with url_root set the encode function would always prepend the url_root. As "foo".indexOf(true) returns -1 which is truthy in JavaScript.

### DIFF
--- a/pkg/base1/test-location.js
+++ b/pkg/base1/test-location.js
@@ -298,4 +298,22 @@ QUnit.test("test", function (assert) {
     window.location.hash = "#/other";
 });
 
+/* Set url_root in beforeEach so the test does not flake */
+QUnit.module("url_root tests", {
+    beforeEach: () => {
+        window.mock = { url_root: "cockpit" };
+    },
+    afterEach: () => {
+        window.mock = null;
+    }
+});
+
+QUnit.test("encode", assert => {
+    let path = cockpit.location.encode("path", "", true);
+    assert.equal(path, "/cockpit/path", "path is correct");
+    /* an existing url_root should not be prepended */
+    path = cockpit.location.encode("/cockpit/path", "", true);
+    assert.equal(path, "/cockpit/path", "path is correct");
+});
+
 QUnit.start();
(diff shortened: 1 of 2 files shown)


pkg/base1/test-location.js
```diff
@@ -298,4 +298,22 @@ QUnit.test("test", function (assert) {
...
+QUnit.module("url_root tests", {
+    beforeEach: () => {
+        window.mock = { url_root: "cockpit" };
+    },
+    afterEach: () => {
+        window.mock = null;
+    }
+});
+
+QUnit.test("encode", function (assert) {
```
martinpitt, line 311: Can you please use an arrow, like above?

pkg/base1/test-location.js
```diff
@@ -298,4 +298,22 @@ QUnit.test("test", function (assert) {
...
+    let path = cockpit.location.encode("path", "", true);
```
martinpitt, line 312: Hm, eslint ought to complain that this can be `const`; apparently it doesn't?

pkg/base1/test-location.js
```diff
@@ -298,4 +298,22 @@ QUnit.test("test", function (assert) {
...
+    assert.deepEqual(path, "/cockpit/path", "path is correct");
```
martinpitt, line 313: A string isn't very deep, is it? :grin:  (Should work fine that way, of course).

martinpitt (dismissed):
Thanks! What an embarassing typo..

OK to land as-is, I proposed a few small cleanups.

pkg/base1/test-location.js
martinpitt: ah - these should be two independent `const`s, but nevermind.

jelly:
> Thanks! What an embarassing typo..
> 
> OK to land as-is, I proposed a few small cleanups.

Fixed it, as it's too small for a follow up

martinpitt (approved):
Dankjewel!

PR #17940, opened 2022-11-21
test: drop pcp workaround for bugz#2140572

jelly:


jelly:
Clearly still broken on fedora-36, fedora-37 even with a refresh.

jelly:
pcp-1.0.1-2 should contain a related fix, but needs an image refresh.

PR #17941, opened 2022-11-22
Drop obsolete versioned dependencies

martinpitt:
Drop cockpit version requirements for versions which are > 2 years old. None of the distributions that we support on the main branch have that ancient versions even in their backport repositories. This does not change anything in practice, as the rpm/deb dependencies are a *lot* stricter already (rpm has ≥ 266, deb has ≥ source version).

Also update the "238.1.1" requirement to "239". Back then that meant "the next version after 238.1".

Fewer magic numbers!

----

https://salsa.debian.org/debian/cockpit-podman/-/merge_requests/4 triggered me to have a closer look.

### DIFF
--- a/tools/cockpit.spec
+++ b/tools/cockpit.spec
@@ -322,8 +322,6 @@ troubleshooting, interactive command-line sessions, and more.
 Summary: Cockpit bridge server-side component
 Requires: glib-networking
 Provides: cockpit-ssh = %{version}-%{release}
-# PR #10430 dropped workaround for ws' inability to understand x-host-key challenge
-Conflicts: cockpit-ws < 181.x
 # 233 dropped jquery.js, pages started to bundle it (commit 049e8b8dce)
 Conflicts: cockpit-dashboard < 233
 Conflicts: cockpit-networkmanager < 233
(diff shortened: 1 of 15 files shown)


tools/cockpit.spec
```diff
@@ -617,8 +615,8 @@ The Cockpit component for managing storage.  This package uses udisks.
 
 %package -n cockpit-tests
 Summary: Tests for Cockpit
-Requires: cockpit-bridge >= 138
-Requires: cockpit-system >= 138
+Requires: cockpit-bridge >= %{required_base}
+Requires: cockpit-system >= %{required_base}
```
jelly, line 619: Interesting, that raises base from 138 -> 266, but that should be fine for `cockpit-tests`

jelly (approved):

tools/cockpit.spec
martinpitt: Right, it mostly just needs to be satisfied in the RHEL cockpit/appstream split, but that's been the case for some time. This is mostly to get rid of magic numbers.

martinpitt:
Meh, I fell into the "3x affected" trap here. [TestSelinux.testTroubleshootAlerts()](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17941-20221122-100526-02eb2ea7-fedora-36/log.html#132) really doesn't like being exercised too strongly.. I don't currently have time to investigate this, but this PR isn't urgent.

PR #17942, opened 2022-11-22
Disallowed users check

jelly:
This is required for cockpit-podman not check itself if /etc/cockpit/disallowed-users exists and if so sed the file.

https://github.com/cockpit-project/cockpit-podman/pull/1145

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -2040,8 +2040,9 @@ class MachineCase(unittest.TestCase):
         """
 
         # fedora-coreos runs cockpit-ws in a containter so does not install cockpit-ws on the host
-        if not self.machine.ostree_image:
-            self.sed_file('/root/d', '/etc/cockpit/disallowed-users')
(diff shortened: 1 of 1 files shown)


test/common/testlib.py
```diff
@@ -1370,15 +1370,15 @@ def is_devel_build(self):
     def disable_preload(self, *packages):
         for pkg in packages:
             path = "/usr/share/cockpit/%s" % pkg
-            if self.machine.execute("if test -e %s; then echo yes; fi" % path):
+            if self.machine.file_exists(path):
```
martinpitt, line 1373: I don't understand the `self.machine` here. This helper is defined on our own class, so shouldn't it be `self.file_exists`?

martinpitt (approved):

test/common/testlib.py
jelly: Duh!

PR #17947, opened 2022-11-23
dummy change to test CI

martinpitt:
Validate container refresh in https://github.com/cockpit-project/cockpituous/pull/534


martinpitt:
Good enough!

PR #17948, opened 2022-11-23
test: exclude Kpatch on tmt

jelly:
The gating test environment set's up and installs testing packages. This leaves the system in a state where there are services to restart/reboot making the kpatch test fail (as there is no continue button).


martinpitt:
Triggering the two pixel fedoras.

martinpitt (approved):
Erk, sorry, pixels was that other PR :grin: Thanks!

PR #17949, opened 2022-11-23
pf, dark: Fix disabled buttons in modals

garrett:
To fix disabled buttons in dark mode in general, we had to change the background and contrast. To fix the buttons in modals, we also need to change the background, as otherwise the background is the same color.


garrett:
Without the fix:

Screenshot from 2022-11-23 16-53-19

With this PR's fix:

Screenshot from 2022-11-23 16-52-56

It also affects https://github.com/cockpit-project/cockpit-machines/pull/646#discussion_r1024649142 and likely other parts of Cockpit (whenever there's a disabled button on a modal).

martinpitt (approved):
Danke!

Issue #17822, opened 2022-10-17
Colon in Full Name on account screen has an obscure message.


AsocPro:
### Explain what happens

1. On the accounts extension press the create new account button.
2. Fill out the form and include a colon in the Full name field
3. Press create and see the error message: useradd: invalid comment 'UserName:'

![Screen Shot 2022-10-17 at 10 54 00 AM](https://user-images.githubusercontent.com/3435186/196238140-5babfd1c-fa62-4539-abe9-b301c0d50961.png)

This is minor for people who would realize this probably means that the useradd application does not like the colon but the error message could be more clear. It should be pretty simple to check for a colon in the full name field and throw a specific error if a colon is included.

### Version of Cockpit

276.1

### Where is the problem in Cockpit?

Accounts

### Server operating system

CentOS

### Server operating system version

CentOS 8

### What browsers are you using?

Firefox

marusak:
Interesting, man pages for `useradd` define:
```
       -c, --comment COMMENT
           Any text string. It is generally a short description of the account, and is currently used as the field for the user's full name.
```

From `any text string` I would not guess there is limit for characters that one can use. Does anyone know if there are more restrictions? I assume colon is restricted so it does not break `/etc/passwd`.

jelly:
The [useradd](https://man.archlinux.org/man/useradd.8#CAVEATS) man page explains what a valid username is and the regex to verify. So sounds like something we can easily validate.

marusak:
This is not the user name, this is the full name or for the matter of fact any description for the account

jelly:
Aah, woops. Interestingly I can't find anything about "Full name" restrictions, reading the shadow-utils code gives me:

https://github.com/shadow-maint/shadow/blob/master/src/useradd.c#L1258
https://github.com/shadow-maint/shadow/blob/master/src/useradd.c#L93

so yeah a colon is not allowed.

AsocPro:
Yeah I couldn't find anything specifically either on restrictions. I did try a bunch of different special characters and none of them had issues besides colon. which makes sense based on /etc/passwd being comma delimited and the code you linked to.

ianbrown78:
Hi All, I don't appear to have permission to assign this to me, but I have created a PR above with changes to validate the username anyway.

PR #17950, opened 2022-11-23
#17822 - Validation of real name field

ianbrown78:
Created a validation method for real_name field with char set of a-z A-Z 0-9 ._-
Replaced the if statement in validate function with a call to validate_real_name which includes the empty real_name check.

Results in the following:
image

### DIFF
--- a/pkg/users/account-create-dialog.js
+++ b/pkg/users/account-create-dialog.js
@@ -116,6 +116,23 @@ function validate_username(username, accounts) {
     return null;
 }
 
+function is_valid_char_real_name(c) {
+    return (c >= 'a' && c <= 'z') ||
+        (c >= 'A' && c <= 'Z') ||
+        (c >= '0' && c <= '9') ||
+        c == '.' || c == '_' || c == '-' || c == ' ';
+}
+
+function validate_real_name(real_name) {
+    if (!real_name)
+        return _("No real name specified");
+
+    for (let i = 0; i < real_name.length; i++) {
+        if (!is_valid_char_real_name(real_name[i]))
+            return _("The full name can only consist of letters from a-z, digits, dots, dashes, underscores and spaces.");
+    }
+}
+
 function suggest_username(realname) {
     function remove_diacritics(str) {
         const translate_table = {
@@ -203,8 +220,7 @@ export function account_create_dialog(accounts) {
     function validate(force, real_name, user_name, password, password_confirm) {
         const errs = { };
 
-        if (!real_name)
-            errors.real_name = _("No real name specified");
+        errs.real_name = validate_real_name(real_name)
 
         if (password != password_confirm)
             errs.password_confirm = _("The passwords do not match");
--- a/pkg/users/account-create-dialog.js
+++ b/pkg/users/account-create-dialog.js
@@ -116,21 +116,13 @@ function validate_username(username, accounts) {
     return null;
 }
 
-function is_valid_char_real_name(c) {
-    return (c >= 'a' && c <= 'z') ||
-        (c >= 'A' && c <= 'Z') ||
-        (c >= '0' && c <= '9') ||
-        c == '.' || c == '_' || c == '-' || c == ' ';
-}
-
 function validate_real_name(real_name) {
     if (!real_name)
         return _("No real name specified");
 
-    for (let i = 0; i < real_name.length; i++) {
-        if (!is_valid_char_real_name(real_name[i]))
-            return _("The full name can only consist of letters from a-z, digits, dots, dashes, underscores and spaces.");
-    }
+    const real_name_chars = Array.from(real_name);
+    if (!real_name_chars.every(character => character != ':'))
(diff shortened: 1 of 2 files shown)


pkg/users/account-create-dialog.js
```diff
@@ -116,6 +116,23 @@ function validate_username(username, accounts) {
...
+        (c >= 'A' && c <= 'Z') ||
+        (c >= '0' && c <= '9') ||
+        c == '.' || c == '_' || c == '-' || c == ' ';
+}
+
+function validate_real_name(real_name) {
+    if (!real_name)
+        return _("No real name specified");
+
+    for (let i = 0; i < real_name.length; i++) {
```
KKoukiou, line 130: Javascript has a nicer way to do it:
real_name.every(is_valid_char_real_name)

KKoukiou (commented):
How will this function behave for languages which don't have latin characters? Example greek/russian/chinese etc?

jelly:
> How will this function behave for languages which don't have latin characters? Example greek/russian/chinese etc?

The validation was discussed in https://github.com/cockpit-project/cockpit/issues/17822 `useradd` only disallows a colon in a username as it's a colon delimited field. So the check can be vastly simplified.

pkg/users/account-create-dialog.js
ianbrown78: Oh that is much nicer! Will update my PR and only check for the delimiter. Thanks!

ianbrown78:
Validation has been simplified to check for no real_name and chars matching ':'.
Works using the Array.every() method as suggested by @KKoukiou.
image

pkg/users/account-create-dialog.js
```diff
@@ -116,6 +116,15 @@ function validate_username(username, accounts) {
     return null;
 }
 
+function validate_real_name(real_name) {
+    if (!real_name)
+        return _("No real name specified");
+
+    const real_name_chars = Array.from(real_name);
+    if (!real_name_chars.every(character => character != ':'))
```
martinpitt, line 124: Heh, that's a rather complicated and slow way of saying ` if (real_name_chars.includes(':'))`

pkg/users/account-create-dialog.js
```diff
@@ -203,8 +212,7 @@ export function account_create_dialog(accounts) {
     function validate(force, real_name, user_name, password, password_confirm) {
         const errs = { };
 
-        if (!real_name)
-            errors.real_name = _("No real name specified");
```
martinpitt, line 207: Uh -- so this was a typo/dead variable before? Well spotted.

martinpitt (commented):
Thank you! This should come with a little integration test in test/verify/check-users. Do you want to have a go at this, or should we push that to the PR? IOW, did you already figure out how to run integration tests locally?

pkg/users/account-create-dialog.js
ianbrown78: Thanks, yeah I noticed that as well, but I thought it was just my computer being funky again.
I am going to change it to use .includes instead of the .every()

I will have a go at the integration test. Never done one in Python, so this will be fun.

martinpitt: Please read test/README.md first, that should hopefully get you started (it refers you to HACKING.md for installing the test dependencies). If you have troubles, please yell here or on IRC (libera.chat) in `#cockpit`, that's usually the fastest way to sort this out.

martinpitt:
This fails for me:
```
  File "/var/home/martin/upstream/cockpit/main/test/verify/check-users", line 83, in testBasic
    b.set_input_text('#account-real-name', "Col:n Colon")  # This should fail
[...]
testlib.Error: timeout
wait_js_cond(ph_is_present("#account-real-name:not([disabled]):not([aria-disabled=true])")): Uncaught (in promise) Error: condition did not become true
```

Not immediately clear why -- I'm triggering a test on our CI so that you can compare. Does this work for you?

    TEST_OS=fedora-37 test/verify/check-users TestAccounts.testBasic -stv

you can also add `TEST_SHOW_BROWSER=1` to get an interactive browser, and due to the `-s` option it's comfortable to debug.

martinpitt:
Thanks! test works fine locally, I triggered them on CI. Please ignore the packit failures, that spontaneously broke this morning and is completely unrelated (unfortunately it hits the very same test).

pkg/users/account-create-dialog.js
```diff
@@ -116,6 +116,15 @@ function validate_username(username, accounts) {
     return null;
 }
 
+function validate_real_name(real_name) {
+    if (!real_name)
+        return _("No real name specified");
```
cockpituous, line 121: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17950-20221125-124129-c16b147c-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

Issue #13396, opened 2020-01-14
overview: Middle-click on links does not work


garrett:
I've tried to middle-click to open the graphs on a secondary page. It doesn't work.

I've also tried a few other links and they have the same issue, all on the overview page and also on the breadcrumbs in the graphs. 

Additionally, right clicking does not show link-related actions (such as open in new tab or window).

FWIW: Middle clicking does work in some other parts of Cockpit, such as the navigation.

KKoukiou:
I couldn't find any links in cockpit (apart from the navigation) that actually middle clicks work. So it's related to all pages, not just overview.

martinpitt:
We discussed this on IRC, but summary for posterity:

 * Conceptually, I'd rather continue to treat cockpit more like an "app" than a collection of static(ish) web pages.

 * This means that external links should always open in a new window/tab, and middle/right click for them should work normally. That's what already happens on at least the "Read more.." in the CPU mitigations dialog, or the Updates changelogs.

 * I am not a fan of internal links creating new tabs. It's losing the "integrated" feeling of cockpit, probably duplicates and triplicates our considerable amount of JavaScript/memory usage, and I'm sure it will break local state on some pages. But I see that this is a debatable point, and I won't veto the general idea, if other people think that it would be better to start supporting this.

 * I hard-veto the approach from #13402, though. First, we must not sidestep cockpit.jump that way -- I haven't tested this, but it's prone to break remote ssh sessions where you need to consider `cockpit.host` (which cockpit.jump does). Second, the solution can't be to add hundreds of preventDefault() to all of our click handlers -- there must be a better solution that does that centrally.

martinpitt:
> This means that external links should always open in a new window/tab

Forgot the justification: If it would open in the same tab, this would kill your cockpit session (after 10 seconds, anyway). I'd say that's a bigger surprise than opening in a new window. We also need this for the cases of cockpit-desktop and for [embedding Cockpit](https://cockpit-project.org/guide/latest/embedding.html).

garrett:
FWIW, navigation items do support middle click and it works as expected.

I agree that we shouldn't support middle-click on non-links, however.

If something goes somewhere (such as navigation or the link to the graphs page in the overview), it should work like a link (and support middle click), but if it performs an action or triggers a modal, it shouldn't (like the overview's table of configuration buttons-that-look-like-links).

In other words, it really should be an opt-in for special cases, not extended everywhere.

---

[...]

PR #17952, opened 2022-11-25
shell: make middle click work for nav items

### LINKED ISSUE #17725: Multiple terminal tabs
Page: Terminal

Couldn't introducing multiple terminal tabs (like in Windows Terminal) be a good idea?

jelly:
When Cockpit is reverse proxied the navigation links did not respect the set Urlroot breaking a middle click.

Related #13396
Fixes #17725


martinpitt (approved):
Nice, thank you! Note that there is a very high chance that `TestAccounts.testBasic` fails on the Testing Farm [like this](https://artifacts.dev.testing-farm.io/d258eb81-68a7-47ba-ba9f-ec31bdabd286/). This is unrelated, and most probably related to this morning's TF release.

martinpitt:
Hmm, [these two failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17952-20221125-093413-8933f638-arch/log.html) look relevant.

As predicted, TestAccounts failed on TF, but I recommend to handle that independently.

jelly:
> 



> Hmm, [these two failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17952-20221125-093413-8933f638-arch/log.html) look relevant.
> 
> As predicted, TestAccounts failed on TF, but I recommend to handle that independently.

Thanks, I was afraid I'd break something and I did! I didn't think machines would break as that should work fine reverse proxied, but I'll take a look.

jelly:
```
cockpit.location.encode("foo", {}, true)
'/=10.111.113.2/foo'
cockpit.location.encode("foo", {}, false)
'/foo'
```

So we wrongly add the `urlRoot` for machines hmmm!

jelly:
Seems it's not so easy!

```
<a class="pf-c-nav__link" aria-current="false" href="/system/%23/memory"><span class="nav-item-name">Memory</span></a>```

That breaks clicking on `Memory` in TestMenu.testBasic.

jelly:
I've adjusted the tests, so it works but I am not 100% confident about this change.

martinpitt:
@jelly: Breaks TestMultiMachine.testUrlRoot :cry:

jelly:
> @jelly: Breaks TestMultiMachine.testUrlRoot cry

I made a boo boo, it takes a root='/' argument so had to adjust the test for that

martinpitt (approved):
Thanks! (Assuming green)

PR #17954, opened 2022-11-25
python: Core classes rework

allisonkarlitskaya:
This is some initial work for landing the peer bridge stuff.  Let's start getting it checked out in advance.

### DIFF
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -21,18 +21,46 @@ import json
 import logging
 import pwd
 import os
+import shlex
 import sys
 
 from systemd_ctypes import EventLoopPolicy
 
+from .channel import Channel
+from .channels import CHANNEL_TYPES
 from .transports import StdioTransport
 from .packages import Packages
 from .router import Router
 
-logger = logging.getLogger('cockpit.bridge')
+logger = logging.getLogger(__name__)
 
 
-async def run():
+class Bridge(Router):
+    def __init__(self, args: argparse.Namespace):
+        self.packages = Packages()
+        self.args = args
+
+        super().__init__(Channel.create_routing_rules(CHANNEL_TYPES))
+
+    @staticmethod
+    def get_os_release():
+        try:
+            file = open('/etc/os-release', encoding='utf-8')
+        except FileNotFoundError:
+            file = open('/usr/lib/os-release', encoding='utf-8')
+
+        with file:
+            lexer = shlex.shlex(file, posix=True, punctuation_chars=True)
+            return dict(token.split('=', 1) for token in lexer)
+
+    def do_send_init(self) -> None:
+        self.send_control(command='init', version=1,
+                          checksum=self.packages.checksum,
+                          packages={p: None for p in self.packages.packages},
+                          os_release=self.get_os_release())
+
+
+async def run(args) -> None:
     logger.debug("Hi. How are you today?")
 
     # Unit tests require this
@@ -42,7 +70,7 @@ async def run():
     os.environ['USER'] = me.pw_name
 
     logger.debug('Starting the router.')
-    router = Router()
+    router = Bridge(args)
     StdioTransport(asyncio.get_running_loop(), router)
 
     logger.debug('Startup done.  Looping until connection closes.')
@@ -54,7 +82,7 @@ async def run():
         pass
 
 
-def main():
+def main() -> None:
     parser = argparse.ArgumentParser(description='cockpit-bridge is run automatically inside of a Cockpit session.')
     parser.add_argument('--privileged', action='store_true', help='Privileged copy of the bridge')
     parser.add_argument('--packages', action='store_true', help='Show Cockpit package information')
@@ -72,7 +100,7 @@ def main():
         print(json.dumps(Packages().get_bridges(), indent=2))
     else:
         asyncio.set_event_loop_policy(EventLoopPolicy())
-        asyncio.run(run(), debug=True)
+        asyncio.run(run(args), debug=True)
 
 
 if __name__ == '__main__':
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -46,7 +46,12 @@ async def run():
     StdioTransport(asyncio.get_running_loop(), router)
 
     logger.debug('Startup done.  Looping until connection closes.')
-    await router.communicate()
+
+    try:
+        await router.communicate()
+    except BrokenPipeError:
+        # not unexpected if the peer doesn't hang up cleanly
+        pass
 
 
 def main():
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -54,10 +54,10 @@ class Bridge(Router):
             return dict(token.split('=', 1) for token in lexer)
 
     def do_send_init(self) -> None:
-        self.send_control(command='init', version=1,
-                          checksum=self.packages.checksum,
-                          packages={p: None for p in self.packages.packages},
-                          os_release=self.get_os_release())
+        self.write_control(command='init', version=1,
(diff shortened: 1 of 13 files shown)


src/cockpit/channel.py
```diff
@@ -15,25 +15,61 @@
...
+        return True
+
+    def apply_rule(self, options: Dict[str, object]) -> Optional[Channel]:
+        assert self.router is not None
+
+        payload = options.get('payload')
+        if not isinstance(payload, str):
+            return None
+
+        for cls in self.table.get(payload, []):
```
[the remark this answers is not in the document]
allisonkarlitskaya: I hate this alert.  This is really and truly a matter of taste.  Can we turn it off?

test/pytest/test_bridge.py
```diff
@@ -0,0 +1,114 @@
...
+from cockpit.router import Router
+
+asyncio.set_event_loop_policy(systemd_ctypes.EventLoopPolicy())
+
+
+class MockTransport(asyncio.Transport):
+    queue: asyncio.Queue[Tuple[str, bytes]]
+
+    def send_json(self, _channel: str, **kwargs) -> None:
+        self.send_data(_channel, json.dumps(kwargs).encode('ascii'))
```
martinpitt, line 18: Totally unimportant style question: `_channel` is not an unused argument, why the _ prefix?

test/pytest/test_bridge.py
```diff
@@ -0,0 +1,114 @@
...
+    def send_data(self, channel: str, data: bytes) -> None:
+        msg = channel.encode('ascii') + b'\n' + data
+        msg = str(len(msg)).encode('ascii') + b'\n' + msg
+        self.protocol.data_received(msg)
+
+    def send_init(self, version=1, host='test', **kwargs):
+        self.send_json('', command='init', version=version, host=host, **kwargs)
+
+    def send_open(self, channel, payload, **kwargs):
+        self.send_json('', command='open', channel=channel, payload=payload, **kwargs)
```
martinpitt, line 29: This looks weird. New channels are always opened on the control channel AFAIK. Why does this need a channel name? Moreover, why does it specify channel twice? Once in the first positional arg (which looks like the correct one), and a different one as kwarg?

test/pytest/test_bridge.py
```diff
@@ -0,0 +1,114 @@
...
+
+        await transport.assert_msg('', command='init')
+        transport.send_init()
+
+        return bridge, transport
+
+    async def test_echo(self):
+        bridge, transport = await self.start()
+
+        transport.send_open('1', 'echo')
```
martinpitt, line 85: That `'1'` here is irritating. See my question above why `send_open()` even expects a channel name.

test/pytest/test_bridge.py
```diff
@@ -0,0 +1,114 @@
...
+        await transport.assert_msg('', command='done', channel='1')
+
+        transport.send_close('1')
+        await transport.assert_msg('', command='close', channel='1')
+
+    async def test_host(self):
+        bridge, transport = await self.start()
+
+        # try to open an null channel, explicitly naming our host
+        transport.send_open('1', 'null', host='test')
```
martinpitt, line 104: It took me a bit to figure out why/how `test` magic. Would you mind introducing a global constant for this, and using it here and in `send_init()`?

src/cockpit/channel.py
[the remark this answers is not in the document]
martinpitt: Fine for me. This is a tool for *us*, and personally I actually like the stronger `else:` form, as it conveys the developer's intent more clearly. See .github/codeql-config.yml you can disable it there.

src/cockpit/bridge.py
```diff
@@ -21,18 +21,39 @@
...
-async def run():
+class Bridge(Router):
+    def __init__(self, args: argparse.Namespace):
+        self.packages = Packages()
+        self.args = args
+
+        super().__init__(Channel.create_routing_rules(CHANNEL_TYPES))
+
+    def do_send_init(self) -> None:
+        with open('/usr/lib/os-release', encoding='utf-8') as file:
```
martinpitt, line 50: Side note: I know I recently moved some `/etc/os-release` over to this, but turns out this was wrong: os-release(5) explicitly wants applications to check the one in /etc.

src/cockpit/channel.py
```diff
@@ -71,7 +71,7 @@ def do_control(self, command, message):
         elif command == 'options':
             self.do_options(message)
 
-    def do_channel_control(self, command, message):
+    def do_channel_control(self, channel, command, message):
```
martinpitt, line 112: Hmm, shouldn't some flake/lint/whatever complain about the unused `channel` variable here and below? I.e. these would normally be named `_channel`.

martinpitt (changes requested):
Thanks! I'm by and large happy with this rework, and in fact excited to see it pave the way for proper routing! :tada: Just some minor remarks/questions, and tests are still complaining.

test/pytest/test_bridge.py
allisonkarlitskaya: Same as the reason we do it in protocol.py: this `_channel` is the channel in the sense of the first line of the message (ie: which is the empty string for control messages).  It needs to have a different name than `channel` which often appears within those messages.

Like, considering this message:

```
51

{"command":"open","channel":"1","payload":"echo"}
```

Then `_channel` is `''` and `channel` is `'1'`.

test/pytest/test_bridge.py
allisonkarlitskaya: See the comment above.  The "channel" in the sense of the framing protocol has to be the empty string for control messages.  Separately, the control message, itself, has a (non-empty) "channel" field.

test/pytest/test_bridge.py
allisonkarlitskaya: The client always decides the names of channels.  We really do need to pick a channel name and send it.

Perhaps we could keep an internal counter on the MockTransport object for this purpose, but then we'd need to return it and check it against subsequent fields...

test/pytest/test_bridge.py
allisonkarlitskaya: Sure.

src/cockpit/bridge.py
allisonkarlitskaya: I just re-read the manpage and it seems to suggest that we ought to check in both places, with precedence given to `/etc`.  The symlink is "to provide compatibility with applications only looking at /etc/."

src/cockpit/channel.py
allisonkarlitskaya: I think it doesn't complain because this is an implementation of a virtual method, and depending on the method, the individual variable names could be considered part of the API (ie: if you wanted to use kwargs with them).

test/pytest/test_bridge.py
```diff
@@ -0,0 +1,113 @@
...
+        await transport.assert_msg('', command='ready', channel='1')
+
+        # try to open an null channel, no host
+        transport.send_open('2', 'null')
+        await transport.assert_msg('', command='ready', channel='2')
```
martinpitt, line 109: What happens if you try to open channel "2" again? That should fail with something like "channel already exists". That seems worth checking?

src/cockpit/channel.py
martinpitt: Makes sense. Python is becoming complex.. :grin:

test/pytest/test_bridge.py
allisonkarlitskaya: Sure.  I wanted to add some tests for this stuff, indeed.  Also things like trying to do an open without specifying a payload or a channel name.  Follow-up? :)

martinpitt (approved):
Cheers!

PR #17958, opened 2022-11-28
test: check-users: expect disabled 'Terminate session' button for logged in users

KKoukiou:


KKoukiou (commented):
The commit message is wrong, should be 'enabled', please change on merge.

KKoukiou (commented):
@marusak for the root user the button should be always disabled according to the code. https://github.com/cockpit-project/cockpit/blob/main/pkg/users/account-details.js#L272

test/verify/check-users
```diff
@@ -84,10 +84,7 @@ class TestAccounts(MachineCase):
         b.wait_visible("#account-delete[disabled]")
         b.wait_visible("#account-real-name[disabled]")
         # when running against existing machines there might be an ssh root session
```
marusak, line 86: This comment is now pointless, isn't it?

a41aca6fded2caaba672769cc5c38a6dbb8bfb88 introduced this change which was released in version 216. Indeed code in this version did not do any checks for `uid === 0`, that was only introduced in 5df053cad5eb37919bbdc5ecaaa02911e5b3a980 which was released in version 221. So this all indeed makes sense, thank you!

marusak (approved):
Thank you!

KKoukiou:
The one test failing on centos stream on tmt fails also in other PRs https://github.com/cockpit-project/cockpit/pull/17959

PR #17962, opened 2022-11-29
python: superuser peer bridge (--privileged mode)

allisonkarlitskaya:
This is semi-working.  Let's start discussing it.

The new `test_superuser()` case in `test/pytest/test_bridge.py` is probably a nice place to look to get an idea of what's already working here.

 - [x] implement `kill` transport control command
 - [x] implement superuser-on-init
 - [x] #17964
 - [x] #17973
 - [x] #17977

### DIFF
--- a/src/cockpit/protocol.py
+++ b/src/cockpit/protocol.py
@@ -124,6 +124,7 @@ class CockpitProtocol(asyncio.Protocol):
     def connection_lost(self, exc):
         logger.debug('connection_lost')
         self.transport = None
+        self.do_closed(exc)
 
         if self._communication_done is not None:
             if exc is None:
@@ -188,10 +189,7 @@ class CockpitProtocol(asyncio.Protocol):
 # CockpitProtocolClient or CockpitProtocolServer.  The main difference here is
 # that the server should send its init message immediately upon the connection
 # being established, whereas the client shouldn't do anything until it sees the
-# init message from the server.  Additionally, the client needs to be able to
-# respond to authorize challenges (via `do_authorize()`), whereas on the server
-# side, we will never send those ourselves (and therefore not need to handle
-# the responses).
+# init message from the server.
 #
 # Both clients and servers need to implement `do_channel_control()` and
 # `do_channel_data()` as well as `do_init()`.
@@ -226,6 +224,9 @@ class CockpitProtocolServer(CockpitProtocol):
     def do_kill(self, host: Optional[str], group: Optional[str]) -> None:
         raise NotImplementedError
 
+    def do_authorize(self, message: Dict[str, object]) -> None:
+        raise NotImplementedError
+
     def transport_control_received(self, command, message):
         if command == 'init':
             try:
@@ -243,6 +244,8 @@ class CockpitProtocolServer(CockpitProtocol):
             self.do_init(message)
         elif command == 'kill':
             self.do_kill(message.get('host'), message.get('group'))
+        elif command == 'authorize':
+            self.do_authorize(message)
         else:
             raise CockpitProtocolError(f'unexpected control message {command} received')
 
--- a/src/cockpit/protocol.py
+++ b/src/cockpit/protocol.py
@@ -41,8 +41,11 @@ class CockpitProtocol(asyncio.Protocol):
     buffer = b''
     _communication_done: Optional[asyncio.Future] = None
 
-    def do_ready(self):
+    def do_ready(self) -> None:
         raise NotImplementedError
+
+    def do_closed(self, exc: Optional[Exception]) -> None:
+        pass
 
     def transport_control_received(self, command: str, message: Dict[str, object]) -> None:
         raise NotImplementedError
--- a/src/cockpit/protocol.py
+++ b/src/cockpit/protocol.py
@@ -220,6 +220,9 @@ class CockpitProtocolServer(CockpitProtocol):
     def do_init(self, message):
         pass
 
+    def do_kill(self, host: Optional[str], group: Optional[str]) -> None:
+        raise NotImplementedError
+
     def transport_control_received(self, command, message):
         if command == 'init':
             try:
@@ -235,6 +238,8 @@ class CockpitProtocolServer(CockpitProtocol):
             except KeyError as exc:
                 raise CockpitProtocolError('missing host field', 'protocol-error') from exc
             self.do_init(message)
+        elif command == 'kill':
+            self.do_kill(message.get('host'), message.get('group'))
         else:
             raise CockpitProtocolError(f'unexpected control message {command} received')
 
--- a/src/cockpit/superuser.py
+++ b/src/cockpit/superuser.py
@@ -1,0 +1,198 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2022 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import os
+import pwd
+
+from systemd_ctypes import bus
+
+from typing import Dict, Optional
+
+from .router import Router, RoutingError, RoutingRule
+from .peer import Peer, PeerStateListener
+
+logger = logging.getLogger(__name__)
+
+SUPERUSER_AUTH_COOKIE = 'supermarius'
+
+# Keep a static list, for the time being.
+SUPERUSER_BRIDGES = {
+    'sudo': (
+        ['sudo', '--askpass', 'cockpit-bridge', '--privileged'], {
+            'SUDO_ASKPASS': '/usr/libexec/cockpit-askpass'
+        }
+    ),
+}
+
+
+class SuperuserStartup:
+    peer: Peer  # the peer that's being started
+
+    def success(self, rule: SuperuserRoutingRule) -> None:
+        raise NotImplementedError
+
+    def failed(self, rule: SuperuserRoutingRule, exc: Exception) -> None:
+        raise NotImplementedError
+
+    def auth(self, rule: SuperuserRoutingRule, prompt: str, echo: bool) -> None:
+        raise NotImplementedError
+
+
+class ControlMessageStartup(SuperuserStartup):
+    def success(self, rule: SuperuserRoutingRule) -> None:
+        rule.router.write_control(command='superuser-init-done')
+
+    def failed(self, rule: SuperuserRoutingRule, exc: Exception) -> None:
+        rule.router.write_control(command='superuser-init-done')
+
+    def auth(self, rule: SuperuserRoutingRule, prompt: str, echo: bool) -> None:
+        username = pwd.getpwuid(os.getuid()).pw_name
+        hexuser = ''.join(f'{c:02x}' for c in username.encode('ascii'))
+        rule.router.write_control(command='authorize', cookie=SUPERUSER_AUTH_COOKIE, challenge=f'plain1:{hexuser}')
+
+
+class DBusStartup(SuperuserStartup):
+    future: asyncio.Future
+
+    def __init__(self):
+        self.future = asyncio.get_running_loop().create_future()
+
+    def success(self, rule: SuperuserRoutingRule) -> None:
+        self.future.set_result(None)
+
+    def failed(self, rule: SuperuserRoutingRule, exc: Exception) -> None:
+        self.future.set_exception(exc)
+
+    def auth(self, rule: SuperuserRoutingRule, prompt: str, echo: bool) -> None:
+        rule.prompt('', prompt, '', echo, '')
+
+    async def wait(self) -> None:
+        await self.future
+
+
+class SuperuserRoutingRule(PeerStateListener, RoutingRule, bus.Object, interface='cockpit.Superuser'):
+    startup: Optional[SuperuserStartup]
+    peer: Optional[Peer]
+
+    # D-Bus signals
+    prompt = bus.Interface.Signal('s', 's', 's', 'b', 's')  # message, prompt, default, echo, error
+
+    # D-Bus properties
+    bridges = bus.Interface.Property('as', value=[])
+    current = bus.Interface.Property('s', value='none')
+
+    # RoutingRule
+    def apply_rule(self, options: Dict[str, object]) -> Optional[Peer]:
+        superuser = options.get('superuser')
+
+        if not superuser or self.current == 'root':
+            # superuser not requested, or already superuser?  Next rule.
+            return None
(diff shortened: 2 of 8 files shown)


allisonkarlitskaya:
Part of the reason for opening this already is because it is a good workout for #17739, and exposes at least one regression there.

We start getting:
```
{'error': ['python.error', ["Introspection: 'Interface org.freedesktop.DBus.Properties is not found'"]], 'id': 'x'}
```

in reply to calling `GetAll` from `assert_props()` in our testcase.  That's probably because this is currently handled by the "all args are strings" case which is removed in #17739.

allisonkarlitskaya:
Cool:

```
DEBUG:cockpit.protocol:sending message  {'channel': 'root-internal', 'command': 'close', 'message': 'failed to connect to None bus: [Errno 2] sd_bus_default_system: No such file or directory', 'problem': 'protocol-error'}
```

Nothing here should be trying to connect to the system bus, so this failure is really helpful.  Will look into it locally.

test/pytest/test_bridge.py
```diff
@@ -113,3 +123,70 @@ async def test_host(self):
...
+    async def test_superuser(self):
+        cockpit.superuser.SUPERUSER_BRIDGES['mock'] = (
+            [sys.executable, '-m', 'cockpit.bridge', '--privileged'],
+            {'PYTHONPATH': ':'.join(sys.path)}
+        )
+
+        bridge, transport = await self.start()
+
+        # try to open root dbus — should fail
+        transport.send_open('root-internal', 'dbus-json3', dus='internal', superuser=True)
```
allisonkarlitskaya, line 147: `dus=` lol.

allisonkarlitskaya:
So, the pybridge scenario generates an insanely huge log, probably due to all the debugging output.  It probably makes sense to start thinking about turning the volume down a bit there...

allisonkarlitskaya:
Next big blocker: `MachineCase.login_and_go()` has a default kwarg `superuser=True` which (long story short) causes cockpit-ws to send a `"superuser"` argument as part of the `"init"` message to the bridge.  After that, the request/response is supposed to happen via authorize control messages.  No part of the above is yet implemented.  Getting that working is probably (hopefully?) going to unblock a huge number of testcases.

allisonkarlitskaya:
Nice baseline
```diff
+ 54 passing
! 19 skipped
- 237 failing
# 15 didn't run for some reason
```

allisonkarlitskaya:
This more than doubles the number of passing tests (well over 100 now) in the `/pybridge` scenario.

allisonkarlitskaya:
```
  File "/source/src/cockpit/superuser.py", line 66, in auth
    hexuser = ''.join(f'{c:02x}' for c in os.getlogin().encode('ascii'))
OSError: [Errno 6] No such device or address
```

...wtf?

I think `os.getlogin()` is opening and reading `/etc/passwd` so the error is probably coming from something there... but... how?

allisonkarlitskaya:
eh whatever

```diff
     def auth(self, rule: SuperuserRoutingRule, prompt: str, echo: bool) -> None:
-        hexuser = ''.join(f'{c:02x}' for c in os.getlogin().encode('ascii'))
+        hexuser = ''.join(f'{c:02x}' for c in os.environ['USER'].encode('ascii'))
         rule.router.write_control(command='authorize', cookie=SUPERUSER_AUTH_COOKIE, challenge=f'plain1:{hexuser}')
```

test/pytest/test_bridge.py
```diff
@@ -76,97 +94,156 @@ async def assert_msg(self, expected_channel, **kwargs) -> None:
...
+    # D-Bus helpers
+    internal_bus: str = ''
+
+    async def open_internal_bus(self):
+        if not self.internal_bus:
+            self.internal_bus = await self.check_open('dbus-json3', bus='internal')
+        assert self.protocol.open_channels[self.internal_bus].bus == self.protocol.internal_bus.client
+        return self.internal_bus
+
+    def send_bus_call(self, bus: str, path: str, iface: str, name: str, args: list) -> str:
```
mvollmer, line 115: Isn't the first argument the channel identifier?  `self.send_json(channel_id, ...)`.  It is called `bus` here.

test/pytest/test_bridge.py
```diff
@@ -76,97 +94,156 @@ async def assert_msg(self, expected_channel, **kwargs) -> None:
         msg = await self.next_msg(expected_channel)
         assert msg == msg | {k.replace('_', '-'): v for k, v in kwargs.items()}
 
+    # D-Bus helpers
+    internal_bus: str = ''
+
+    async def open_internal_bus(self):
```
mvollmer, line 109: Calling this `get_internal_bus` might be better. I was confused by the `if bus is None: bus = open_internal_bus()` pattern below, and thought it would open a new channel every time, which felt pretty useless.

test/pytest/test_bridge.py
allisonkarlitskaya: Yes.  It's the channel ID of the bus.  I could rename it.

test/pytest/test_bridge.py
allisonkarlitskaya: Well ,it opens it if it's not already.  Sure, makes sense, though.

allisonkarlitskaya:
Ahah!!

python says

```
Help on built-in function getlogin in module posix:

getlogin()
    Return the actual login name.
```

But the cpython source code says this is implemented via `getlogin()`  in libc.  The manpage for that tells quite a different story:

```
       getlogin()  returns  a pointer to a string containing the name of the user logged in
       on the controlling terminal of the process, or a null pointer  if  this  information
       cannot  be  determined.
```

and sure enough

```
       ENXIO  The calling process has no controlling terminal.
```

test/pytest/test_bridge.py
mvollmer: Ahh, it's the channel for the bus... got it.

src/cockpit/protocol.py
```diff
@@ -235,6 +235,8 @@ def transport_control_received(self, command, message):
             except KeyError as exc:
                 raise CockpitProtocolError('missing host field', 'protocol-error') from exc
             self.do_init(message)
+        elif command == 'kill':
+            self.do_kill(message.get('host'), message.get('group'))
```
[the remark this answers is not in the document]
allisonkarlitskaya: Yes.  I think we should, indeed.

src/cockpit/superuser.py
```diff
@@ -0,0 +1,215 @@
...
+    def auth(self, rule: SuperuserRoutingRule, prompt: str, echo: bool) -> None:
+        raise NotImplementedError
+
+
+class ControlMessageStartup(SuperuserStartup):
+    def success(self, rule: SuperuserRoutingRule) -> None:
+        rule.router.write_control(command='superuser-init-done')
+
+    def failed(self, rule: SuperuserRoutingRule, exc: Exception) -> None:
+        pass  # what do we do for this?
```
mvollmer, line 64: I think we also send `super-user-init-done`.  Cockpit-ws doesn't care whether superuser startup has worked or not.

mvollmer: The main point of super-user-init-done is to know that no more auth messages will come and the password can be removed from memory.

allisonkarlitskaya: Ah.  Thanks for flagging this.  I'll add it.

src/cockpit/superuser.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # RoutingRule
+    def apply_rule(self, options: Dict[str, object]) -> Optional[Peer]:
+        if not options.get('superuser') or self.current == 'root':
+            # superuser not requested, or already superuser?  Next rule.
+            return None
+        elif self.peer:
+            # superuser requested and active?  Return it.
+            return self.peer
+        else:
```
mvollmer, line 110: There is also `superuser: "try"`.

src/cockpit/superuser.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def go(self, startup: SuperuserStartup, name: str) -> None:
+        assert self.current == 'none'
+        assert self.peer is None
+        assert self.startup is None
+
+        args, env = SUPERUSER_BRIDGES[name]
+        startup.peer = Peer(self.router, name, self)
+        startup.peer.spawn(args, env)
+        self.startup = startup
```
mvollmer, line 157: Let's set `self.startup` before calling `spawn`, no? Just in case `spawn` triggers some PeerStateListener events.

allisonkarlitskaya: I actually specifically wanted to do it the other way around.  I nearly added a comment about this, in fact:

If spawn throws some error (which it might do) then I want us to abandon the process and *not* set `self.startup`.

src/cockpit/superuser.py
```diff
@@ -0,0 +1,215 @@
...
+        self.go(ControlMessageStartup(), name)
+
+    # D-Bus methods
+    @bus.Interface.Method(in_types=['s'])
+    async def start(self, name: str) -> None:
+        try:
+            startup = DBusStartup()
+            self.go(startup, name)
+            await startup.wait()
+        except AssertionError as exc:
```
mvollmer, line 172: I feel this is better done with a direct check: `if self.current != 'none':`

src/cockpit/superuser.py
```diff
@@ -0,0 +1,215 @@
...
+            raise bus.BusError('cockpit.Superuser.Error', 'Superuser bridge already running') from exc
+        except KeyError as exc:
+            raise bus.BusError('cockpit.Superuser.Error', 'Unknown superuser bridge type') from exc
+        except OSError as exc:
+            raise bus.BusError('cockpit.Superuser.Error', f'Failed to start peer bridge: {exc}') from exc
+
+    @bus.Interface.Method()
+    def stop(self) -> None:
+        if self.peer is not None:
+            self.peer.close()
```
mvollmer, line 182: We might need to wait here until the peer has really gone away before sending a reply. JS will call Stop() and then immediately Start() sometimes.

src/cockpit/superuser.py
allisonkarlitskaya: "Ask forgiveness, not permission"

allisonkarlitskaya: But indeed, for assertions, I could be convinced of the other way.

src/cockpit/superuser.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @bus.Interface.Method(in_types=['s'])
+    def answer(self, reply: str) -> None:
+        if self.startup is not None:
+            self.startup.peer.authorize_response(reply)
+
+
+'''
```
mvollmer, line 190: What is this block for? :)

allisonkarlitskaya: This is how the interface definition would look when a new experimental branch of `systemd_ctypes` lands.  It kinda doesn't belong here to be honest.  I'll remove it.

src/cockpit/superuser.py
allisonkarlitskaya: Done.

src/cockpit/superuser.py
allisonkarlitskaya: Done.

src/cockpit/superuser.py
allisonkarlitskaya: Done.

src/cockpit/superuser.py
allisonkarlitskaya: Added the comment.

src/cockpit/superuser.py
allisonkarlitskaya: I added this.  It's remarkably simple.  In case "try" is given, `return self.peer`.  then it will either be the root bridge, or None, which is what we want.

src/cockpit/superuser.py
allisonkarlitskaya: Done.

test/pytest/test_bridge.py
```diff
@@ -247,3 +265,103 @@ async def test_dbus_watch(self):
...
+        # watch for signals
+        await self.transport.add_bus_match('/superuser', 'cockpit.Superuser')
+        await self.transport.watch_bus('/superuser', 'cockpit.Superuser',
+                                       {'Bridges': ['sudo', 'pseudo'], 'Current': 'none'})
+
+        # start the bridge.  with a password this is more complicated
+        with unittest.mock.patch.dict(os.environ, {"PSEUDO_PASSWORD": "p4ssw0rd"}):
+            start = self.transport.send_bus_call(self.transport.internal_bus, '/superuser', 'cockpit.Superuser', 'Start', ['pseudo'])
+            # first, init state
+            await self.transport.assert_bus_notify('/superuser', 'cockpit.Superuser', {'Current': 'init'})
```
mvollmer, line 329: I am not sure what the "init" state is actually supposed to mean... I don't think the C bridge cycles through it when calling Start(), for example. I think Current is only "init" during bridge startup, to help reduce some banner flickering during Cockpit login.

allisonkarlitskaya: Confusingly, it's what happens after the bridge has been spawned, but before we see the "init" message.  This is the state we're in when we get authentication requests, for example.

In this state, we have a banner about "Limited privs", but the button is no longer present.

mvollmer: Some docs: https://github.com/cockpit-project/cockpit/blob/main/src/bridge/cockpitrouter.c#L1444

mvollmer:
Nice!  here is what I found:

 - `superuser: "try"` needs to be supported
 - `Current == "init"` needs to be looked at. If we get this wrong, pages might get into endless reloading.

Doesn't need to be done in this PR, however.

mvollmer (approved):

test/pytest/test_bridge.py
mvollmer: > In this state, we have a banner about "Limited privs", but the button is no longer present.

Hmm, sounds like a bug.  The whole banner should be hidden when Current == "init".

mvollmer:
So the rules for the Current property are, afaics:

The initial value for Current is:

- "root", when the bridge runs as root already.  This value will never change.
- "none", when a superuser bridge has not been requested in the init message. This is stable until Start() is called on the D-Bus API.
- "init", when a superuser bridge has been requested but it's not fully up yet. Authorization messages are being exchanged. This will change to "none" when the startup fails, or to the rule name when successful. Start() during this time should fail, and Stop() during this time should abort the peer startup and set Current to "none".

When Start() is called, the value of Current changes exactly once from "none" to the new value.

When Stop() is called, the value changes exactly once to "none".

allisonkarlitskaya:
I think we've decided that this discrepancy is harmless, and possibly might be a useful API change.

mvollmer:
> * `Current == "init"` needs to be looked at. If we get this wrong, pages might get into endless reloading.

Oh, no need to panic, supermarius!  There will be no endless reloading in any case, and with the current code here, nothing bad will happen at all. We might even document that Start() will set Current to "init" momentarily.

allisonkarlitskaya:
The bar has been raised.

image

allisonkarlitskaya:
Now all that's left is all the hard work... :sweat_smile:

martinpitt:
Yay, congrats @allisonkarlitskaya !

My favourite so far is that one:

> warning: failed to connect to session bus: [Errno 123] sd_bus_default_user: No medium found

Did we forget to put an actual SD card on the bus? :wink:

martinpitt:
More seriously, it seems that might be the next biggest step: That session bus error seems to break dozens of networking tests. Aside from the fact that NM does not really listen on the session bus anyway, so I wonder what's confused here.

So, as you say, "just" hard work now! :hammer_and_pick:

PR #17963, opened 2022-11-29
python: Deduplicate signal emission

mvollmer:
Each signal that arrives from D-Bus should only result in a single JSON message on the channel.

- [x] #17739 
- [ ] https://github.com/allisonkarlitskaya/systemd_ctypes/pull/38

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -43,7 +43,7 @@ import json
 import logging
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
I think we discussed it already, but I write it here for the record:

I'd prefer to wait on this until we know if:
 - the C bridge does this
 - overlapping signals on a single channel come up often
 - deduplicating them is important for anything in particular

mvollmer:
If this ever gets interesting again, we can dig it out again.

PR #17964, opened 2022-11-29
two small python tweaks

allisonkarlitskaya:
- change quoting to `"""` instead of `'''`
 - disable debug output by default


src/cockpit/buffered.pyfragment
```diff
@@ -1,6 +1,6 @@
 
 class CockpitBufferedProtocol(CockpitBaseProtocol, asyncio.BufferedProtocol):
-    '''A hopefully-reasonably-efficient implementation of the Cockpit frame protocol
+    """A hopefully-reasonably-efficient implementation of the Cockpit frame protocol
```
jelly, line 3: Seems black also prefers """, interesting! Although this is still inconsistent in our codebase I bet, so linting it some day would be nice

jelly (approved):

PR #17966, opened 2022-11-30
test: apply the same workaround for centos-9-stream

jelly:


jelly:
Should we report this somewhere? I am not 100% sure what the bugzilla is about.

martinpitt:
@jelly It's the opposite -- this bug was the reason why requesting the superuser channel did *not* work earlier. It was eventually fixed in sudo and Fedora, and apparently the fix is now making its way into C9S/RHEL 9 as well.

The problem is that this still doesn't work on our own image -- I trigger a centos-9-stream test here, but I'm 99% sure it  will fail. I think we'll need to land this in lockstep with an image refresh?

martinpitt:
> I trigger a centos-9-stream test here

Haha, we don't actually test our centos-9-stream image on cockpit, lol. So this is ready to go, thank you!

martinpitt (approved):

PR #17967, opened 2022-11-30
pkg: fix document class checking 

jelly:


jelly:
Thanks @martinpitt for finding this!

martinpitt (approved):
Thanks! c9s packit fix just landed, we can ignore that here.

jelly:
All green except centos-9-stream ofcourse, let's land it and rebase the dark theme switcher!

PR #17969, opened 2022-11-30
test: only capture failed services header

jelly:
Fix TestSystem.testOverview flake by only capturing the unit header as the log section can differ in layout causing flakes. We already take a full screenshot of the service overview page for succesful units.


jelly:
Only triggered fedora-coreos and fedora-36 as this is purely a pixel test related change.

KKoukiou (dismissed):

KKoukiou (approved):

PR #17970, opened 2022-11-30
ws: always take packages from the first host

allisonkarlitskaya:
When constructing requests to the packages webserver in the bridge, in response to incoming GET requests, always route the request to the bridge on the login host.

This prevents the additional hosts that we login to from sending us evil JS.

### DIFF
--- a/src/ws/cockpitchannelresponse.c
+++ b/src/ws/cockpitchannelresponse.c
@@ -617,7 +617,6 @@ cockpit_channel_response_serve (CockpitWebService *service,
                                          "payload", "http-stream1",
                                          "internal", "packages",
                                          "method", "GET",
-                                         "host", host,
                                          "path", path,
                                          "binary", "raw",
                                          NULL);


allisonkarlitskaya:
My opinion, for the record: I think we're inescapably moving toward a place where the pages available for use with cockpit are more a feature of the client side (or bastion host/container) and less a feature of the installed OS.  We're discussing — after all — using cockpit to admin machines that don't have *any* cockpit packages on them.  I think we can come up with all kinds of interesting ways to offer sufficient extensibility from the side of the client.

BUT, for the sake of this PR: this is not about packages coming from the client.  In this case, the packages are still coming from the server — but the first server.  I don't think that doing this is in any way dishonest or reduced functionality — in some ways it might even be preferable, since custom pages (for example) need only be installed one one host, and then can be used on all the others.

I certainly disagree that we're somehow ending up with a dysfunctional vestige of what we used to have here.  It's just different — in some ways better, but definitely different.

And I certainly disagree with ripping the feature out entirely as being the "honest" way to proceed.

martinpitt:
> I certainly disagree that we're somehow ending up with a dysfunctional vestige of what we used to have here. It's just  different — in some ways better, but definitely different.

Strongly disagree. I have my server (Debian stable) as a remote host. Of course for dogfooding, but also because I find some features genuinely interesting, e.g. rummaging in PCP's logs (the new metrics page was the reason why I installed PCP on my server). This is actually quite a common setup -- workstations and servers are necessarily very different, and servers also have different purposes, packages, and OSes. I built cockpit-ws with this change, connected to my server, and in the first two minutes see these issues:

 - The metrics page says "Package cockpit-pcp is missing for metrics history" instead of showing history, which is a lie -- cockpit-pcp is installed on my server, just not on my workstation.
 - The Updates page is completely kaputt. My workstation runs Fedora OSTree and has cockpit-ostree, my server has cockpit-packagekit (Debian).
 - My server now shows SELinux, kdump, and sosreport pages, which don't work.
 - cockpit-podman is a "nice" example of a cockpit-foo package co-evolving with foo. We've had [several](https://github.com/cockpit-project/cockpit-podman/commit/76393d457fcf05) [API](https://github.com/cockpit-project/cockpit-podman/commit/bde9b67d96f8442) [changes](https://github.com/containers/podman/issues/7078) in the past (and many more, like that stupid `Id` -> `ID` rename, can't find the commit quickly). Using incompatible cockpit-podman vs. podman versions is going to break the page.

All these are fixable *somehow* -- sometimes it's relatively easy, like adding some runtime detection to manifest parsing. Sometimes it's super hard, like keeping compatibility with changing APIs for a *lot* longer, and testing a lot more old OS releases (Ubuntu 20.04, Debian oldstable, RHEL 7, etc.). But we have to do *something* -- we can't possibly sell this broken remote session in good faith. 

FTR, it has never been my intention to only *ever* use pages from the server. I'd like to get some *usable* Cockpit with a few crucial pages (Overview, some metrics, Terminal), and most importantly the ability to discover and install more cockpit-* packages as needed (like podman, storaged, machines, 389-ds, file-sharing, and [more](https://cockpit-project.org/applications.html)). Supporting *all* of these in a static flatpak or container is neither doable nor desirable.

So, if your intention of this PR is something like "it'll be at the end of a long dependency list and a year or three of work", then sure. If you actually intended to land it soon, then I frankly don't see how to salvage shell remote hosts, and it'd be honest to just kill it.

allisonkarlitskaya:
See #19071 for the new way forward here...

PR #17972, opened 2022-11-30
Allow root login on RHEL 8

jelly:
With the disallowed-users feature we now disable root login by default in Cockpit. In RHEL 8 however, a default install can login as root over ssh with password authentication so allow this as well in Cockpit.

The /etc/cockpit/disallowed-users file is kept as then administrators can optionally easily disable root login when wanted.

@martinpitt @marusak do we want what I did? My reasoning is that it's then easier to disable if required.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -604,6 +604,10 @@ class TestConnection(MachineCase):
     def testWsPackage(self):
         m = self.machine
 
+        # On RHEL-8 SSH allows password root login by default, so Cockpit does too.
+        ROOT_LOGIN_ENABLED = ['rhel-8', 'centos-8']
+        root_login_disallowed = not any(m.image.startswith(img) for img in ROOT_LOGIN_ENABLED)
+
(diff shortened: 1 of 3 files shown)


martinpitt:
The approach here is fine, I like creating the file on RHEL 8 as well. However, this breaks [testPamAccess](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17972-20221130-195125-823feb40-rhel-8-8/log.html) in RHEL 8, I suppose that test also needs to be adjusted

jelly:
> 



> The approach here is fine, I like creating the file on RHEL 8 as well. However, this breaks [testPamAccess](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17972-20221130-195125-823feb40-rhel-8-8/log.html) in RHEL 8, I suppose that test also needs to be adjusted

Ah, I only tested the cockpitWs test last night, so I missed one! Will investigate, that can't be hard.

martinpitt (approved):

PR #17977, opened 2022-12-01
Some more peer prep patches

allisonkarlitskaya:
#17962 got too big, yet again.  More "non-controversial" stuff to review first.

### DIFF
--- a/src/cockpit/protocol.py
+++ b/src/cockpit/protocol.py
@@ -131,15 +131,19 @@ class CockpitProtocol(asyncio.Protocol):
     def write_frame(self, frame):
         frame_length = len(frame)
         header = f'{frame_length}\n'.encode('ascii')
-        self.transport.write(header + frame)
+        if self.transport is not None:
+            self.transport.write(header + frame)
 
     def write_channel_data(self, channel, payload):
         """Send a given payload (bytes) on channel (string)"""
         # Channel is certainly ascii (as enforced by .encode() below)
         frame_length = len(channel + '\n') + len(payload)
         header = f'{frame_length}\n{channel}\n'.encode('ascii')
-        logger.debug('writing to transport %s', self.transport)
-        self.transport.write(header + payload)
+        if self.transport is not None:
+            logger.debug('writing to transport %s', self.transport)
+            self.transport.write(header + payload)
+        else:
+            logger.debug('cannot write to closed transport')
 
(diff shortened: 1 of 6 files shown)


src/cockpit/protocol.py
```diff
@@ -131,15 +131,19 @@ def connection_lost(self, exc):
     def write_frame(self, frame):
         frame_length = len(frame)
         header = f'{frame_length}\n'.encode('ascii')
-        self.transport.write(header + frame)
+        if self.transport is not None:
```
martinpitt, line 134: Should this raise a warning or even assertion? This doesn't just look like an "remote peer unexpectedly closed its end" situation, as `self.transport` is local state -- i.e. this smells like a logic error in our code? I.e. why do we still call write_frame() after getting connection_lost()?

src/cockpit/protocol.py
```diff
@@ -131,15 +131,19 @@ def connection_lost(self, exc):
...
         # Channel is certainly ascii (as enforced by .encode() below)
         frame_length = len(channel + '\n') + len(payload)
         header = f'{frame_length}\n{channel}\n'.encode('ascii')
-        logger.debug('writing to transport %s', self.transport)
-        self.transport.write(header + payload)
+        if self.transport is not None:
+            logger.debug('writing to transport %s', self.transport)
+            self.transport.write(header + payload)
+        else:
+            logger.debug('cannot write to closed transport')
```
martinpitt, line 146: That's the direction I meant above, but debug() isn't very visible.

src/cockpit/protocol.py
allisonkarlitskaya: This isn't about the peer but rather about the main stdio transport.  If the connection closes suddenly due to a protocol error or something and we're in the middle of writing a bunch of stuff from some channel, this is the case where we try to write these frames.

It's annoying too, since the original error message gets lost in s bunch of unrelated tracebacks about "NoneType has no write method".

I think that asyncio.run() is partially responsible here too.  We return from the main() function in the bridge as soon as we see that the connection dropped, but it seems to want to run outstanding tasks to completion.   Or at least there are probably cases where some tasks have IO scheduled before our communicate() future callback gets run.

martinpitt (approved):
Thanks! Annoying, but let's not get stuck on that.

Issue #15090, opened 2021-01-04
Tabular numbers in Cockpit


garrett:
We don't have tabular number support in RedHatFont. This is a regression, as OpenSans (our old font for text) had it.

See https://github.com/RedHatOfficial/RedHatFont/issues/27 for the open issue.

garrett:
This is basically when you have monospaced numbers in a proportional font so numbers line up in columns, especially around any decimal point. Additionally, it ensures that times don't shift when they change ("13:00" to "13:01" for example, where the 0 is much wider than the 1) and progress percentages don't jump around when they update (30% to 31% to 32% etc.).

garrett:
I have a hacky solution in mind that might hold us over while we wait for proper support in the font. (We've been literally waiting for years.)

The idea is to use FontForge to make a supplimentary monospaced font with the glyphs from RedHatText. Delete everything but the numbers and separators (0123456789,.:) and then use CSS to patch the font with a subset override.

This should be super-tiny in file size and let us have tabular numbers.

I am pretty sure we can make it an opt-in by redefining RedHatText additionally as RedHatTextTabular _then_ patch it in. This would allow us to default to the proper font but switch to a tabular version in lists, progress bars, times, and everywhere else Cockpit would benefit from inlined monospaced numbers ("tabular numbers").

garrett:
The problem is illustrated very well by our recent SVG charting changes:

![fonts-chart](https://user-images.githubusercontent.com/10246/98678167-6e569c80-235e-11eb-93ad-b88c7fec65dd.png)

Look at the 1 and how it's badly aligned to the 0 and the 2 that surround it. (Other numbers are off too, but not quite as much.)

garrett:
I've hacked the font a bit to make all the characters be the same width and then aligned the glyphs to the center (with a manual offset for the 1) as a proof of concept: 
[RedHatText-Regular-subset-numbers.zip](https://github.com/cockpit-project/cockpit/files/5765997/RedHatText-Regular-subset-numbers.zip)

All other characters have been removed as to reduce the file size.

Resulting space on disk is 3.7 kB compressed (which is how it would be sent). I think we could inline it into CSS to not have an additional connection (where the overhead would be higher than the embedding).

We've done this font patching before with by adding a subset of newer PF3 icons into an older PF3 set for older RHEL support (and those were inlined). @martinpitt and I did that back then. It's the same technique here as a possible fix until _whenever_ we get the real fix from upstream.

[...]

PR #17979, opened 2022-12-01
Fonts PF4 enable updates and tabular fonts

KKoukiou:
* [x] blocked on https://github.com/patternfly/patternfly/issues/5273

Closes https://github.com/cockpit-project/cockpit/issues/15090

### DIFF
--- a/pkg/apps/index.html
+++ b/pkg/apps/index.html
@@ -29,7 +29,7 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <script type="text/javascript" src="../manifests.js"></script>
     <script type="text/javascript" src="apps.js"></script>
   </head>
-  <body class="pf-m-redhat-font">
+  <body class="pf-m-redhat-font pf-m-tabular-nums">
 
     <div class="ct-page-fill" id="apps-page">
     </div>
--- a/pkg/kdump/index.html
+++ b/pkg/kdump/index.html
@@ -31,7 +31,7 @@ along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
     <script type="text/javascript" src="po.js"></script>
 </head>
 
-<body class="pf-m-redhat-font">
+<body class="pf-m-redhat-font pf-m-tabular-nums">
     <div class="ct-page-fill" id="app"></div>
 </body>
 </html>
--- a/pkg/lib/patternfly/_fonts.scss
+++ b/pkg/lib/patternfly/_fonts.scss
@@ -30,3 +30,9 @@ $relative: true
 @include printRedHatFont(700, "Medium");
 @include printRedHatFont(700, "MediumItalic", $style: "italic");
 @include printRedHatFont(400, "Regular");
+@include printRedHatFont(700, "Bold", $familyName: "RedHatMono");
+@include printRedHatFont(700, "BoldItalic", $style: "italic", $familyName: "RedHatMono");
+@include printRedHatFont(300, "Italic", $style: "italic", $familyName: "RedHatMono");
+@include printRedHatFont(400, "Medium", $familyName: "RedHatMono");
+@include printRedHatFont(400, "MediumItalic", $style: "italic", $familyName: "RedHatMono");
+@include printRedHatFont(300, "Regular", $familyName: "RedHatMono");
--- a/pkg/lib/qunit-template.html
+++ b/pkg/lib/qunit-template.html
@@ -25,7 +25,7 @@
     <script src="<%= htmlWebpackPlugin.options.builddir %>dist/base1/cockpit.js"></script>
     <script src="<%= htmlWebpackPlugin.options.script %>"></script>
 </head>
-<body class="pf-m-redhat-font">
+<body class="pf-m-redhat-font pf-m-tabular-nums">
     <h1 id="qunit-header"><%= htmlWebpackPlugin.options.title %></h1>
     <h2 id="qunit-banner"></h2><div id="qunit-testrunner-toolbar"></div>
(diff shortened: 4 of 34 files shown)


doc/guide/static/style.css
```diff
@@ -2,17 +2,17 @@
...
   font-style: normal;
   font-weight: 400;
   text-rendering: optimizeLegibility;
 }
 
 @font-face {
-  font-family: "RedHatText";
-  src: url("./RedHatText-Medium.woff2") format("woff2");
+  font-family: "RedHatTextVF";
+  edf: url("./RedHatTextVFModified-updated.woff2") format("woff2");
```
KKoukiou, line 15: s/edf/src

KKoukiou:
There is something weird going on with the font-weights I feel.
Tabular fonts for numbers work as expected.

garrett:
The fonts are completely, absolutely incorrect locally. I'm seeing Overpass instead of Red Hat Display and Red Hat Text.

image

I'm building a VM now to see if it's the same case there. It could be one of the few things the symlink hack doesn't cover (such as the login page and ws stuff).

garrett:
OK. So it is totally different in a VM.

First impression: Oh, wow. It feels quite different.

image

Oddly, the font chosen for the graphs is wrong. That's Overpass...?!

Alignment feels off a bit. I guess ems are differently sized now, as they're based on font size. Hm. It's especially noticeable in the space above the headings.

garrett:
image

OK, in main: 
- alignment of the Filesystems table is a regression (I remember we did fix that at some point)
  - it's correct in storage logs, drives, other devices (look at where the text starts on the left)
- headings for the reading / writing are wrong here too

Aha! The alignment is a breakpoint thing; it's correct when wider:

image

These things are unrelated to the font. I'm just trying to figure out _why_ the pages feel different in real-time.

KKoukiou:
@marusak I want to discuss with you shortly the version compatibility issue here. As you can see I dropped the old fonts, from webpack, so the dist/static will no longer include the old fonts. The 'old' fonts are 'needed' when someone uses 'pf-m-redhat-font' class, which is what all external projects currently do.

Do we want to ship old and new fonts for a while?  Or keep it clean, drop the old fonts, and make sure to update all external projects to require the new fonts and also the spec files to require the new version of cockpit-ws which ships the new fonts?

garrett:
Font's aren't working properly... Here's a VM:

image

I do see tabular fonts, this is true.

However, look at the headings. That's clearly Overpass and not Red Hat Display. It might be that the fonts aren't being set or sent via the browser and that Overpass is in the fallback stack (and I have Overpass installed, as several old mockups use it, as it predated Red Hat Text and Red Hat Display), so you might see something different on your machine.

KKoukiou:
> Perhaps I guess this means should step back a little and use a more normal set of fonts instead of variable.
> 
> I'm suggesting that we still jump forward to the new fonts and the new features (especially tabular)... just not all the way to variable. The main downside would be that we would still have to send a few more files (like we currently do) versus variable fonts. But hopefully the weights would be correct across all browsers.

These are not variable fonts. These are normal updated fonts.

garrett:
I guess I didn't mention that I am seeing tabular numbers across Cockpit, and that's great.

I think we might be using them in too many places... and this is mainly a problem because when using tabular numbers, the 0 turns into a slashed variant. But the thing is: the font variants can be controlled via CSS, including a slashed zero. But the font doesn't seem to support that and always uses a slashed 0.

We need to file a bug both in the Red Hat Font repo and in PatternFly to have this fixed. I think we can work around it in Cockpit too, by subsetting the mono font (which we should also be loading) and splice in the 0 from it into places where the tabular numbers are used (via the class).

This subset workaround should work, as long as the mono font has a version without the slash. Otherwise, we'd need to either ship an entirely separate font file just for the 0. Or, I could inline it within the CSS (which makes a lot more sense here)... it wouldn't be that large as it would only contain a single character. We actually did this before for icons for an older version of RHEL, where I patched in support with a custom font (where I extracted just the relevant glyphs) and included via CSS.

...However, we shouldn't let the slashed zero block this. It should be a follow-up. But before it lands, we should probably be careful to not introduce the tabular numbers everywhere.

The slashed zero is used to differentiate the `0` from an `O`. But at small sizes, especially for those of us with astigmatism, it can sometimes look like an `8`. Since most of the places where we actually want/need tabular numbers is actually just numbers, it's more useful for us to have a non-slashed 0, especially for accessibility reasons. (We could still keep a slashed version for mixed alphanumeric strings like showing hashes. Perhaps by having the override include a `:not()` with a class.)

garrett:
> That's because fonts are in static and by default on local developer mode you want see changes there. You need to do this 

No, not for a VM. This was a VM. (Only the first try was on my local machine, which is why I started using a VM after that.)

But I'll build another VM.

garrett:
Everything's looking good in my test VM.

And you even fixed the headings for the graphs?! Awesome.

image

(Top is the heading selected; under that is all the fonts used on the page.)

image

We should also use Red Hat Mono instead of Noto Sans Mono, but that could be done as a follow-up too:

image

garrett (dismissed):
We might want to use `pf-m-tabular-nums` more locally (perhaps on tables and lists and things like that), but I think it's probably fine.

We can discuss it and change it here in this PR or even change the targets in a follow-up too.

Meanwhile approving.

marusak (dismissed):
Thank you!

KKoukiou:
Just added one comment.

KKoukiou:
There are two unrelated flakes. Since this pr retries all pages, I will do an expeption and merge this not completely green, because I already retried 3 times, and it's blocking test/reference updates.

Issue #17975, opened 2022-12-01
Test Flake menu


jelly:
### Explain what happens

```
ok 28 test/verify/check-examples TestXHRProxy.testBasic [ND@0]
ok 84 test/verify/check-packagekit TestUpdates.testBasic [ND@0]
ok 211 test/verify/check-superuser TestSuperuser.testRemoveBridgeConfig [ND@0]
ok 206 test/verify/check-superuser TestSuperuser.testBrokenBridgeConfig [ND@0]
ok 181 test/verify/check-storage-partitions TestStoragePartitions.testPartitions [ND@0]
ok 163 test/verify/check-storage-lvm2 TestStorageLvm2.testSnapshots [ND@0]
ok 261 test/verify/check-system-services TestServices.testRelationships [ND@0]
ok 256 test/verify/check-system-services TestServices.testConditions [ND@0]
ok 145 test/verify/check-sosreport TestSOS.testVerbose [ND@0]
ok 112 test/verify/check-reauthorize TestReauthorize.testSudo [ND@0]
not ok 104 test/verify/check-pages TestPages.testMenuSearch [ND@0]
```

jelly:
Running TestSOS.testVerbose multiple times gives me:

```
Dec 02 09:23:44 fedora-36-127-0-0-2-2201 systemd[28005]: Reached target default.target - Main User Target.
Dec 02 09:23:44 fedora-36-127-0-0-2-2201 systemd[28005]: Startup finished in 91ms.
Dec 02 09:23:46 fedora-36-127-0-0-2-2201 systemd[28005]: Created slice session.slice - User Core Session Slice.
Dec 02 09:23:46 fedora-36-127-0-0-2-2201 systemd[28005]: Starting dbus-broker.service - D-Bus User Message Bus...
Dec 02 09:23:46 fedora-36-127-0-0-2-2201 dbus-broker-launch[28032]: Service file '/usr/share/dbus-1/services/certmonger.service' is not named after the D-Bus name 'org.fed>
Dec 02 09:23:46 fedora-36-127-0-0-2-2201 systemd[28005]: Started dbus-broker.service - D-Bus User Message Bus.
Dec 02 09:23:46 fedora-36-127-0-0-2-2201 dbus-broker-launch[28032]: Policy to allow eavesdropping in /usr/share/dbus-1/session.conf +31: Eavesdropping is deprecated and ig>
[...]

PR #17981, opened 2022-12-02
test: reset failed units at the start of testMenuSearch

jelly:
As we reset all failed systemd units and then directly take a screenshot this test is bound to have race conditions as the browser might not have been re-rendered. Resetting all units before we login should be good enough as the test does not start/stop services.

Closes: #17975


allisonkarlitskaya (approved):

KKoukiou (commented):
Thanks!

KKoukiou:
This fixes a flake, let's not turture outselves with retrying this time.

PR #17986, opened 2022-12-04
shell: setDarkMode also when updating existing frames

### LINKED ISSUE #17978: In 281, Services page has black background, other pages are white
### Explain what happens

1. Run the Fedora openQA Cockpit tests on a cockpit-281 build
2. Note that the Services page appears to be using 'dark mode' (i.e. the background is black), but every other page (overview, Logs, Software updates...) is still 'light mode'

Overview: overview screenshot
Logs: logs
But, Services: services

Maybe this is intentional? But it looks pretty odd. Reporting it before I just adjust the test to expect it.

### Version of Cockpit

281

Services

### Server operating system

Fedora

any

Firefox

### System log

_No response_

KKoukiou:
Fixes #17978

### DIFF
--- a/pkg/shell/base_index.js
+++ b/pkg/shell/base_index.js
@@ -219,6 +219,8 @@ function Frames(index, setupIdleResetTimers) {
                     return;
                 setDarkMode(event.target.documentElement, index.dark_mode);
             }, true);
+        } else if (frame.contentDocument.documentElement) {
+            setDarkMode(frame.contentDocument.documentElement, index.dark_mode);
         }
         frame_ready(frame);
         return frame;
--- a/pkg/shell/base_index.js
+++ b/pkg/shell/base_index.js
@@ -33,6 +33,14 @@ function component_checksum(machine, component) {
         return "$" + machine.manifests[pkg][".checksum"];
 }
 
+function setDarkMode(documentElement, dark_mode) {
+    if (dark_mode) {
+        documentElement.classList.add('pf-theme-dark');
+    } else {
+        documentElement.classList.remove('pf-theme-dark');
+    }
+}
+
 function Frames(index, setupIdleResetTimers) {
     const self = this;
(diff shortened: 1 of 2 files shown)


KKoukiou:
This desperately needs a test, can check it tomorrow.

pkg/shell/base_index.js
```diff
@@ -33,6 +33,15 @@ function component_checksum(machine, component) {
         return "$" + machine.manifests[pkg][".checksum"];
 }
 
+function setDarkMode(documentElement, dark_mode) {
+    console.info("Set dark mode", dark_mode);
+    if (dark_mode) {
+        documentElement.classList.add('pf-theme-dark');
+    } else {
```
cockpituous, line 40: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17986-20221204-161200-7083ac82-fedora-36-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly (approved):

jelly:
ubuntu-stable :heavy_check_mark:  debian-testing :heavy_check_mark: now just fedora-36 :grimacing:

jelly:
We should probably merge this as TestServices.testTimers flakes on main as well, the issue here seems to be some sort of race condition:

Debugging this a bit, it seems to be dependent 

```
log: RELOADING
log: RELOADING
log: listUnits
log:
log: RELOADING
log: RELOADING
log: listUnits
log:
log: RELOADING
log: RELOADING
log: listUnits
log:
log:
log: RELOADING
log: RELOADING
```

Note that `listUnits` isn't called, which is exactly the issue shown in CI. So most likely `loadingUnits` is true and we don't call `listUnits`.

jelly:
Turns out my hunch was correct, the page is somehow in loading units state, interestingly for both `Reloading` calls.

jelly:
Let's see if this fixes it https://github.com/cockpit-project/cockpit/pull/18005 :crossed_fingers:

PR #17988, opened 2022-12-05
selinux: avoid AT_SECURE on exec from cockpit-session

allisonkarlitskaya:
Since cockpit-session has its own selinux label, executing other (differently-labelled) processes from it is considered to be a privilege boundary transition, similar to executing a setuid binary.

Because of this, `AT_SECURE` gets set in the auxiliary vector, causing the libc of the spawned process to enter "secure execution" mode.  Among other things, that causes `secure_getenv()` to return NULL, even if a particular variable is in the environment.

libsystemd uses `secure_getenv()` to look for the user's D-Bus.  If `AT_SECURE` is set, that will fail, causing `sd_bus_default_user()` to return `-ENOMEDIUM`.  This breaks the Python bridge.

The solution to this problem is to mark the state transition as `noatsecure`. That makes sense: there's no real scenario where cockpit-session attacks cockpit-bridge via its environment — cockpit-session is the (substantially) higher-privileged process here.

 - [x] Stop ruining our flake stats: https://github.com/cockpit-project/bots/pull/4129
 - [ ] Get this reviewed by SELinux team

### DIFF
--- a/selinux/cockpit.te
+++ b/selinux/cockpit.te
@@ -164,8 +164,9 @@ auth_write_login_records(cockpit_session_t)
 corenet_tcp_bind_ssh_port(cockpit_session_t)
 corenet_tcp_connect_ssh_port(cockpit_session_t)
 
-# cockpit-session can execute cockpit-agent as the user
+# cockpit-session can execute cockpit-bridge as the user, without setting AT_SECURE
 userdom_spec_domtrans_all_users(cockpit_session_t)
+userdom_noatsecure_login_userdomain(cockpit_session_t)
 usermanage_read_crack_db(cockpit_session_t)
 
 optional_policy(`


allisonkarlitskaya:
This needs a good look over from someone who actually understands this stuff.

martinpitt:
@vmojzis, @wrabcak , @bachradsusi, we'd really appreciate a look from you -- is this sane? Thanks!

martinpitt:
@bachradsusi : You can think of "cockpit-session" as the equivalent of /usr/bin/login, and "cockpit-bridge" as /bin/bash. I.e:

cockpit-session runs as root, starts/runs the PAM session, starts a user session, and runs cockpit-bridge as the primary process in it. -session is highly security critical, small, and we audit it very carefully.

cockpit-bridge is a huge complex program which can do anything, but does not have any privileges (i.e. just like a shell) and runs inside the user session.

We want to set up some environment variables for the session in `cockpit-session`, and want `secure_getenv()` to work on  them, as these are trustworthy. 

So the question is, would something like `login` or `sshd` or other PAM/session launchers also get `userdom_noatsecure_login_userdomain()`?

zpytela:
Right, we haven't found any substantial obstacle. Feel free to merge the commit if you are sure it's needed - the noatsecure permission is quite rare and there is a reason for it.

Although:

I don't quite understand the first sentence:
Since cockpit-session has its own selinux label, executing other (differently-labelled) processes from it is considered to be a privilege boundary transition, similar to executing a setuid binary.
Actually SELinux domains are allowed in the policy execute another confined executable either in the caller domain, or with a transition to a new one.
For setuid, a policy capability is needed.

and this one:
That makes sense: there's no real scenario where cockpit-session attacks cockpit-bridge via its environment — cockpit-session is the (substantially) higher-privileged process here.
So why noatsecure is needed? Is it only because of the getenv functions usage?

As I understand, SELinux is blocking the transition since not long ago, is this a new cockpit feature?

In current policy, the `userdom_noatsecure_login_userdomain()` interface is used only for `init_t`.
There are noatsecure-rules for local_login_t, remote_login_t, xdm_t, but they never target login_userdomain. Cockpit behaves different to general login process then?

FYI this is the kernel no_atsecure check implementation:
https://access.redhat.com/labs/rhcb/RHEL-9.0/kernel-5.14.0-70.13.1.el9/source/blob/security/selinux/hooks.c#L2351

allisonkarlitskaya:
selinux isn't blocking the transition.  it's just that the launched executable gets AT_SECURE in it's aux vector which causes problems.

the issue is basically that we were using gdbus before (which reads envvars via getenv, and is unaffected by AT_SECURE) and now we want to use sd-bus (which uses secure_getenv()).

we never wanted AT_SECURE set before, so in some sense, this is actually an extremely old bug -- we just never noticed it until now.

I feel like behavior similar to what xdm or so is doing would be appropriate.  how would that look, in terms of changes to our policy?

allisonkarlitskaya:
Okay.  Let's merge it like this, then.

PR #17989, opened 2022-12-05
More python bridge improvements

allisonkarlitskaya:


martinpitt:
@allisonkarlitskaya : This landed too early, it fails on c8s (see packit test):
```
  File "test/common/run-tests", line 43, in <module>
    class Test:
  File "test/common/run-tests", line 83, in Test
    def finish(self, affected_tests: list[str], opts: argparse.Namespace) -> Tuple[Optional[str], int]:
TypeError: 'type' object is not subscriptable
+ RC=1
+ echo 1
```

PR #17991, opened 2022-12-05
systemd_ctypes: sync with main

allisonkarlitskaya:
This prevents lots of "You forgot to close fd xx" messages from spamming the log on bridge shutdown, which is the leading cause of failures due to "Test passed, but unexpected journal messages".


martinpitt:
@allisonkarlitskaya : This [c8s failure](https://artifacts.dev.testing-farm.io/4f0a2f3f-274b-418c-8f4a-a55648796738/) smells like a regression from #17989:
```
  File "test/common/run-tests", line 43, in <module>
    class Test:
  File "test/common/run-tests", line 83, in Test
    def finish(self, affected_tests: list[str], opts: argparse.Namespace) -> Tuple[Optional[str], int]:
TypeError: 'type' object is not subscriptable
+ RC=1
+ echo 1
```

This regression is on main already, so OK for this PR.

martinpitt (approved):

PR #17993, opened 2022-12-05
run-tests: using typing.List[] instead of list[]

allisonkarlitskaya:
The later isn't supported with older Python version.


martinpitt (approved):
Danke!

PR #18000, opened 2022-12-06
tools: Update Debian's policykit-1 dependency for package split

martinpitt:
In Debian, policykit-1 0.120-4 got split into multiple smaller packages, in particular `pkexec` and `polkitd`. Cockpit can call `pkexec` and does not directly talk to `org.freedesktop.PolicyKit1`, so update the recommendation for `pkexec`. Still keep the old name as a fallback, to keep the package backportable to Debian 11 and Ubuntu.

https://bugs.debian.org/1025556


marusak (approved):
Danke!

PR #17536, opened 2022-07-08
shell: Support for alternatives to sudo

mvollmer:
Demo: https://www.youtube.com/watch?v=ET5w39M3HIw

------

### Shell: Support for alternatives to sudo

Cockpit always had the ability to run arbitrary programs to gain administrative access, by adding them to a package manifest.  However, the UI would always use "sudo". Now the UI lets you select which program to use when multiple are defined in the manifests.

By default, Cockpit will still only use "sudo", so unless you actually add more programs via a manifest, there will be no visible change in the UI.

PR #18002, opened 2022-12-06
add support for non-superuser peers from manifest.json files

allisonkarlitskaya:
This was easy and fun.   Let's hope it works :sweat_smile:

### DIFF
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -34,6 +34,7 @@ from .channels import CHANNEL_TYPES
 from .config import Config, Environment
 from .internal_endpoints import EXPORTS
 from .packages import Packages, PackagesListener
+from .peer import PeersRoutingRule
 from .remote import HostRoutingRule
 from .router import Router
 from .superuser import SUPERUSER_AUTH_COOKIE, SuperuserRoutingRule
@@ -73,12 +74,15 @@ class Bridge(Router, PackagesListener):
         self.internal_bus.export('/config', Config())
         self.internal_bus.export('/environment', Environment())
 
+        self.peers_rule = PeersRoutingRule(self)
+
         self.packages_loaded()
 
         super().__init__([
             HostRoutingRule(self),
             self.superuser_rule,
             ChannelRoutingRule(self, CHANNEL_TYPES),
+            self.peers_rule,
         ])
 
     @staticmethod
@@ -118,6 +122,7 @@ class Bridge(Router, PackagesListener):
         bridge_configs = self.packages.get_bridge_configs()
         if self.bridge_rules != bridge_configs:
             self.superuser_rule.set_configs(bridge_configs)
+            self.peers_rule.set_configs(bridge_configs)
             self.bridge_configs = bridge_configs
 
 
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -115,10 +115,10 @@ class Bridge(Router, PackagesListener):
 
     # PackagesListener interface
     def packages_loaded(self):
-        bridge_rules = self.packages.get_bridges()
-        if self.bridge_rules != bridge_rules:
-            self.superuser_rule.set_bridge_rules(bridge_rules)
-            self.bridge_rules = bridge_rules
+        bridge_configs = self.packages.get_bridge_configs()
+        if self.bridge_rules != bridge_configs:
+            self.superuser_rule.set_configs(bridge_configs)
+            self.bridge_configs = bridge_configs
 
 
 async def run(args) -> None:
@@ -208,7 +208,7 @@ def main() -> None:
     if args.packages:
         Packages().show()
     elif args.bridges:
-        print(json.dumps(Packages().get_bridges(), indent=2))
+        print(json.dumps(Packages().get_bridge_configs(), indent=2))
     else:
         asyncio.set_event_loop_policy(EventLoopPolicy())
         asyncio.run(run(args), debug=args.debug)
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 
 import asyncio
 
-from typing import Any, ClassVar, Dict, List, Optional, Sequence, Tuple, Type
+from typing import ClassVar, Dict, List, Optional, Sequence, Tuple, Type
 
 from .router import Endpoint, Router, RoutingRule
 
@@ -190,7 +190,7 @@ class ProtocolChannel(Channel, asyncio.Protocol):
     _transport: Optional[asyncio.Transport]
     _loop: Optional[asyncio.AbstractEventLoop]
     _send_pongs: bool = True
-    _last_ping: Optional[Dict[str, Any]]
+    _last_ping: Optional[Dict[str, object]]
 
     _send_pings: bool = False
     _out_sequence: int = 0
@@ -200,7 +200,7 @@ class ProtocolChannel(Channel, asyncio.Protocol):
     _close_on_eof: bool = False
     _eof: bool = False
 
-    def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, Any]) -> asyncio.Transport:
+    def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> asyncio.Transport:
         """Creates the transport for this channel, according to options.
 
         The event loop for the transport is passed to the function.  The
@@ -219,7 +219,7 @@ class ProtocolChannel(Channel, asyncio.Protocol):
         assert isinstance(transport, asyncio.Transport)
         self._transport = transport
 
-    def _close_args(self) -> Dict[str, Any]:
+    def _close_args(self) -> Dict[str, object]:
         return {}
 
     def connection_lost(self, exc: Optional[Exception]) -> None:
--- a/src/cockpit/peer.py
+++ b/src/cockpit/peer.py
@@ -21,7 +21,7 @@ import asyncio
 import logging
 import os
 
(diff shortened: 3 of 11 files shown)


test/pytest/test_bridge.py
```diff
@@ -419,3 +421,52 @@ async def test_freeze(self):
...
+        assert command == 'open'
+        assert message == {'command': 'open', 'channel': pseudo_echo, 'payload': 'echo', 'for-pseudo': True}
+
+        # wait for the open reply, which should allow us to receive init and thaw the queue
+        await self.transport.assert_msg('', command='ready', channel=pseudo_echo)
+        assert peer.queue is None
+
+        # and of course, it should echo
+        self.transport.send_data(pseudo_echo, b'hihi\n')
+        await self.transport.assert_data(pseudo_echo, b'hihi\n')
```
martinpitt, line 462: It's not a surprise that `pseudo_echo` replies with `hihi` -- it's a standard "echo" channel after all, i.e. at least that part of the test would succeed just equally well without the `for_pseudo`. Could this somehow test that "hihi" actually gets sent to and processed by the executed peer program? Or even better, could this change test/pytest/pseudo.py to do something different than just echoing, like returning the reverse of its input?

src/cockpit/peer.py
```diff
@@ -142,3 +142,41 @@ def do_channel_control(self, channel: str, command: str, message: Dict[str, obje
...
+        super().__init__(router)
+
+        self.args = args
+        self.env = env
+        self.predicate = predicate
+        self.peer = None
+
+    def apply_rule(self, options: Dict[str, object]) -> Optional[Peer]:
+        if not self.predicate(options):
+            return None
```
martinpitt, line 163: This needs some documenation what apply_rule() actually does and what it returns. Currently it *both* has a side effect (changing its internal state `self.peer`) *and* a return value (returning a "peer"). It's also not actually being called anywhere, so I can't infer how it's supposed to be used.

src/cockpit/peer.py
```diff
@@ -142,3 +142,41 @@ def do_channel_control(self, channel: str, command: str, message: Dict[str, obje
...
+
+        if self.peer is None:
```
martinpitt, line 165: Is this supposed to be idempotent?

src/cockpit/peer.py
```diff
@@ -142,3 +142,41 @@ def do_channel_control(self, channel: str, command: str, message: Dict[str, obje
...
+            try:
+                peer = Peer(self.router, self.args[0], self)
+                peer.spawn(self.args, self.env)
+                peer.freeze_endpoint()
+                self.peer = peer
+            except OSError:
+                raise RoutingError('blah-blah')
```
martinpitt, line 172: :grin:

src/cockpit/peer.py
```diff
@@ -142,3 +142,41 @@ def do_channel_control(self, channel: str, command: str, message: Dict[str, obje
...
+
+        return self.peer
+
+    def peer_state_changed(self, peer: Peer, event: str, exc: Optional[Exception] = None):
```
martinpitt, line 176: Note to self: This isn't called from anywhere in this commit; This is called from `Peer`.

src/cockpit/peer.py
```diff
@@ -142,3 +142,41 @@ def do_channel_control(self, channel: str, command: str, message: Dict[str, obje
...
+        logger.debug('%s got peer state event %s %s %s', self, peer, event, exc)
+        if event == 'init':
+            if self.peer is not None:
+                self.peer.thaw_endpoint()
+        elif event == 'closed':
+            self.peer = None
```
martinpitt, line 182: Should this handle an exception somehow? By closing the channel or so, to not leak spawned peer bridge processes?

src/cockpit/bridge.py
```diff
@@ -62,10 +63,15 @@ def __init__(self, args: argparse.Namespace):
         self.superuser_rule = SuperuserRoutingRule(self, args.privileged)
         self.internal_bus.export('/superuser', self.superuser_rule)
 
+        # we run this after the channel rules, so we don't need to check source="internal"
+        self.pcp_rule = PeerRoutingRule(self, ['/usr/libexec/cockpit-pcp'], {},
```
martinpitt, line 67: The path needs to come from the manifest. But see below.

src/cockpit/bridge.py
```diff
@@ -62,10 +63,15 @@ def __init__(self, args: argparse.Namespace):
...
+                                        lambda msg: msg.get('payload') == 'metrics')
+
         super().__init__([
             HostRoutingRule(self),
             self.superuser_rule,
             ChannelRoutingRule(self, CHANNEL_TYPES),
+            self.pcp_rule,
```
martinpitt, line 74: So to be clear: This does not use the routing rule in `pkg/pcp/manifest.json` at all, but instead hardcodes it. With the C bridge you could add third-party bridges without changing cockpit code, which was quite cool (I don't know how much this is actually being used in practice, though).

This already hardcodes the superuser rules, which breaks our recent feature to make them more flexible (PR #17536), on [customer demand](https://bugzilla.redhat.com/show_bug.cgi?id=2091034). But the machinery that we now have in place here seems powerful enough that it could just actually read the `match` rules and the `spawn` program vector from the manifest?

martinpitt (changes requested):
Thanks! This was quite a mouthful to understand and go through. Some things to be discussed, unsurprisingly :grin:

src/cockpit/peer.py
martinpitt: Ah, this overrides `RoutingRule.apply_rule`, that has a docstring.

src/cockpit/peer.py
martinpitt: Looking at src/cockpit/router.py, this is called for every channel open, so this acts like a cache and it should be idempotent.

src/cockpit/bridge.py
martinpitt: To clarify: I don't mind hardcoding this here as a provisional step towards building the bridge. I'd just like to discuss the eventual plan, i.e. that this is a temporary hack until we implement manifest reading.

src/cockpit/bridge.py
martinpitt: See PR #18168 for actually fixing the manifest expansion of that macro.

martinpitt:
#18177 cleaned up the freezing, and addressed some of my comments above. I resolved them. This PR now needs a rebase.

martinpitt (changes requested):
The first 6 commits are fine, but IMHO "python: add routing for bridges from manifests" commit is not ready to land. It looks mostly fine to me, but it's dead code, neither unit tested nor used at run time, and does not actually enable any new bridges. It also looks suspiciously similar to what superuser.py does. Could you please move this into a new draft PR, and do the unification with superuser or enablement of pcp there?

There is also one remaining comment ("Should this handle an exception somehow") from the previous review round. I resolved all the others.

src/cockpit/channel.py
```diff
@@ -190,7 +190,7 @@ class ProtocolChannel(Channel, asyncio.Protocol):
     _transport: Optional[asyncio.Transport]
     _loop: Optional[asyncio.AbstractEventLoop]
     _send_pongs: bool = True
-    _last_ping: Optional[Dict[str, Any]]
+    _last_ping: Optional[Dict[str, object]]
```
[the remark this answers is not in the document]
allisonkarlitskaya: `Any` means "don't attempt to reason about types"

`object` means `object`.


That means that if you treat `Any` as a `str`, for example, you won't get errors.  But if you try to treat `object` as a `str` without additional checks, you'll get errors.

src/cockpit/peer.py
```diff
@@ -170,3 +170,88 @@ def do_channel_control(self, channel: str, command: str, message: Dict[str, obje
...
+                return None
+
+        # Start the peer if it's not running already
+        if self.peer is None:
+            try:
+                args = self.config['spawn']
+                env = self.config.get('environ', [])
+                name = self.config.get('label', args[0])
+
+                peer = Peer(self.router, name, self)
```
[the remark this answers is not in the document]
allisonkarlitskaya: Note: this name is only used for debugging purposes...

src/cockpit/channel.py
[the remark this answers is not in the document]
martinpitt: Thanks!

src/cockpit/peer.py
allisonkarlitskaya: This happens when the peer has shutdown, either due to request (which can happen if peer rules are removed, for example) or because of the peer process exiting.  In either case, `Peer` is responsible for sending `close` messages for the outstanding channels.

src/cockpit/bridge.py
allisonkarlitskaya: This is taken from the manifest now.

allisonkarlitskaya:
> The first 6 commits are fine, but IMHO "python: add routing for bridges from manifests" commit is not ready to land. It looks mostly fine to me, but it's dead code, neither unit tested nor used at run time, and does not actually enable any new bridges.

I agree that it's not ready to land: it fixes 5 testcases that we need to remove `@todo` markers from.  It's very much not dead code, though.

> It also looks suspiciously similar to what superuser.py does.

I agree.  I intend to address this in the future.  I'm not sure how, yet.

src/cockpit/peer.py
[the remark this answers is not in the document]
martinpitt: Sorry, this is right. It already falls back to args[0].

allisonkarlitskaya:
Removed 5 `@todo`.

martinpitt (approved):
Hooray!

PR #4139, opened 2016-04-02
bridge: Return PEM data when TLS validation fails

petervo:
Return the PEM data of the certificate that failed validation as part of the close message for http-stream and websocket-stream channels.

PR #18004, opened 2022-12-07
test: Adjust testSELinuxRestrictedUser for fixed SELinux in RHEL 9.2

martinpitt:
Becoming superuser as SELinux restricted user now started to work in RHEL 9.2, right after C9S (commit 1589689abe7). Invert the condition, so that we now have a shrinking list of broken OSes -- that is easier to review and maintain.

---

This needs to land in lockstep with the rhel-9-2 image refresh in #4139


martinpitt:
Hmm, somehow GitHub missed to run its actions here, trying a force-push

jelly (approved):
The invert is much better, thanks!

PR #18006, opened 2022-12-08
Fix users highlight

marusak:


KKoukiou (commented):
Patternfly advices to keep the table headers when applying filters and the table is empty.

But we can take a look to fix it in a global way in cockpit.

PR #18015, opened 2022-12-09
users: Show loading state

marusak:
Current implementation would show blank page and wait until everything was loaded. Now the overview is shown with loading state inside groups and users views. Details page shows big spinner.

Screenshot from 2022-12-09 13-08-21
Screenshot from 2022-12-09 13-08-08

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -51,6 +51,7 @@ import './cockpit-components-table.scss';
  * - emptyCaptionDetail: extra details to show after emptyCaption if list is empty
  * - emptyComponent: Whole empty state component to show if the list is empty
  * - isEmptyStateInTable: if empty state is result of a filter function this should be set, otherwise false
+ * - loading: Set to string when the content is still loading. This string is shown.
  * - variant: For compact tables pass 'compact'
  * - gridBreakPoint: Specifies the grid breakpoints ('', 'grid' | 'grid-md' | 'grid-lg' | 'grid-xl' | 'grid-2xl')
  * - sortBy: { index: Number, direction: SortByDirection }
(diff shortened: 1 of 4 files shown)


pkg/lib/cockpit-components-table.jsx
```diff
@@ -133,6 +135,13 @@ export const ListingTable = ({
             : null
     );
 
+    if (loading)
+        return <EmptyState>
+            <EmptyStateBody>
+                <div>{loading}</div>
```
KKoukiou, line 141: I bet the div is not needed.

pkg/users/account-details.js
```diff
@@ -173,6 +174,17 @@ export function AccountDetails({ accounts, groups, shadow, current_user, user })
                 .catch(show_unexpected_error);
     }
 
+    if (!accounts.length) {
+        return (
+            <EmptyState variant={EmptyStateVariant.small}>
+                <Spinner className="ct-spinner" isSVG size="xl" />
```
KKoukiou, line 180: THis class is not defined not used.

tomasmatus (approved):
Looks like the tests are failing on the flaky check-account locked but you're looking into that on another branch. The loading state looks good!

garrett:
Technically, loading should look different:

https://www.patternfly.org/v4/components/table/html-demos/loading-state-demo/

image

But having some sort of loading indicator is definitely better than none. And we'd probably want to switch inline table/list loading indicators across Cockpit.

---

So, I'm fine with landing this as-is and doing some follow-ups. Or, we could have a fixup here and follow-ups for others.

KKoukiou (approved):

martinpitt:
[two](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18015-20221212-113219-7e3c68d1-fedora-36/log.html#317-1) [TestAccounts](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18015-20221212-113215-7e3c68d1-rhel-8-7/log.html#317) failures due to affected retry. Trying again.

PR #18022, opened 2022-12-12
python: run the static code linters from pytest

allisonkarlitskaya:
Similar to the browser tests, add some more adaptor code for running our static checkers from pytest.

### DIFF
--- a/test/pytest/test_linters.py
+++ b/test/pytest/test_linters.py
@@ -1,0 +1,15 @@
+import glob
+import os
+import subprocess
+import sys
+
+import pytest
+
+SRCDIR = os.path.realpath(f'{__file__}/../../..')
+
+
+@pytest.mark.parametrize('linter', ['flake8', 'pycodestyle', 'pylint', 'vulture'])
+def test_lint(linter):
+    pytest.importorskip(linter)
+    files = glob.glob(f'{SRCDIR}/src/cockpit/**/*.py', recursive=True)
+    subprocess.run([sys.executable, '-m', linter, *files], cwd=SRCDIR)


martinpitt:
[This](https://github.com/cockpit-project/cockpit/actions/runs/3675923494/jobs/6215941476#step:4:328) looks rather weird:
```
test/pytest/test_linters.py::test_lint[src/cockpit/packages.py-flake8] PASSED [ 27%]
test/pytest/test_linters.py::test_lint[src/cockpit/packages.py-pycodestyle] PASSED [ 27%]
test/pytest/test_linters.py::test_lint[src/cockpit/packages.py-pylint] SKIPPED [ 28%]
test/pytest/test_linters.py::test_lint[src/cockpit/packages.py-vulture] PASSED [ 29%]
test/pytest/test_linters.py::test_lint[src/cockpit/__init__.py-flake8] PASSED [ 29%]
test/pytest/test_linters.py::test_lint[src/cockpit/__init__.py-pycodestyle] PASSED [ 30%]
test/pytest/test_linters.py::test_lint[src/cockpit/__init__.py-pylint] SKIPPED [ 31%]
test/pytest/test_linters.py::test_lint[src/cockpit/__init__.py-vulture] PASSED [ 31%]
test/pytest/test_linters.py::test_lint[src/cockpit/bridge.py-flake8] PASSED [ 32%]
test/pytest/test_linters.py::test_lint[src/cockpit/bridge.py-pycodestyle] PASSED [ 32%]
test/pytest/test_linters.py::test_lint[src/cockpit/bridge.py-pylint] SKIPPED [ 33%]
test/pytest/test_linters.py::test_lint[src/cockpit/bridge.py-vulture] PASSED [ 34%]
```

It means it imports and runs pylint, etc. for every single file individually? That's very slow and redundant, isn't it? As it needs to import/parse imports for each file, and then check the common ones multiple times.

allisonkarlitskaya:
Is that why pylint is so slow? :sweat_smile:

test/pytest/test_linters.py
```diff
@@ -0,0 +1,15 @@
...
+import subprocess
+import sys
+
+import pytest
+
+SRCDIR = os.path.realpath(f'{__file__}/../../..')
+
+
+@pytest.mark.parametrize('linter', ['flake8', 'pycodestyle', 'pylint', 'vulture'])
+@pytest.mark.parametrize('filename', glob.glob('src/cockpit/**/*.py', root_dir=SRCDIR, recursive=True))
```
martinpitt, line 12: In other words, this glob shouldn't be a test parameter, but an argument to calling the linter -- or even better, they normally should detect the modules by themselves.

You explicitly don't want to run them on executables here?

martinpitt (changes requested):
See above

allisonkarlitskaya:
Ya — it's an interesting point: we go out of our way to do this differently for `test-static-code`, so let's make the effort here as well.

For what it's worth, I modelled this after what the pytest flake8 plugin trys to do — create a separate case for each file.

allisonkarlitskaya:
Yikes.  This was passing because it lacked `check=True`.  There's an awful lot of stuff to fix up if we actually want to do this.

allisonkarlitskaya:
Don't care enough.

PR #18030, opened 2022-12-13
metrics: fix pod lookup in top 5 memory users

jelly:
The userid was not passed to `cgroupRow` which meant that the lookup of podman names did not work for the top 5 memory services. We now also test if the containers show up, this is a bit different from our service test as we are limited by busybox sh.

### DIFF
--- a/test/verify/check-metrics
+++ b/test/verify/check-metrics
@@ -804,9 +804,7 @@ class TestCurrentMetrics(MachineCase):
         # libpod-$containerid but as podman-3679.scope.
         if m.image != "centos-8-stream" and not m.image.startswith("rhel-8"):
             # copy images for user podman tests; podman insists on user session
-            m.execute("""
-podman save quay.io/libpod/busybox | sudo -i -u admin podman load
-            """)
+            m.execute("podman save quay.io/libpod/busybox | sudo -i -u admin podman load")
 
             # Test user containers
             admin_s = ssh_connection.SSHConnection(user="admin",
@@ -1025,6 +1023,42 @@ BEGIN {{
             m.execute("systemctl stop mem-hog")
 
         m.execute("rm /tmp/hogged")
+
+        # Test Podman containers
+        container_name = "pod-mem-hog"
+        # pipe to tail to keep the data in memory
+        m.execute(f"""
+            podman run --rm -d --name {container_name} quay.io/libpod/busybox /bin/sh -c '
+            head -c 300m /dev/zero | tail | sleep infinity'""")
+
+        # It takes one re-render for the name lookup
+        with b.wait_timeout(30):
+            b.wait_text("table[aria-label='Top 5 memory services'] tbody tr:nth-of-type(1) td[data-label='Service']", f"pod {container_name}")
+
+        m.execute(f"podman stop -t 0 {container_name}")
+
+        # RHEL-8 / CentOS-8's podman user containers do not show up as
+        # libpod-$containerid but as podman-3679.scope.
+        if m.image != "centos-8-stream" and not m.image.startswith("rhel-8"):
+            # copy images for user podman tests; podman insists on user session
+            m.execute("podman save quay.io/libpod/busybox | sudo -i -u admin podman load")
+
+            # Test user containers
+            admin_s = ssh_connection.SSHConnection(user="admin",
+                                                   address=m.ssh_address,
+                                                   ssh_port=m.ssh_port,
+                                                   identity_file=m.identity_file)
+            user_container_name = "user-mem-hog"
+            admin_s.execute(f"""
+                podman run --rm -d --name {user_container_name} quay.io/libpod/busybox /bin/sh -c '
+                head -c 300m /dev/zero | tail | sleep infinity'
+            """)
+
+            # It takes one re-render for the name lookup
+            with b.wait_timeout(30):
+                b.wait_text("table[aria-label='Top 5 memory services'] tbody tr:nth-of-type(2) td[data-label='Service']", f"pod {user_container_name}")
+
+            admin_s.execute(f"podman stop -t 0 {user_container_name}")
 
         # Test link to user services
         # older releases don't have memory accounting enabled for user services
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -534,7 +534,7 @@ class CurrentMetrics extends React.Component {
 
         const topServicesMemory = n_biggest(this.cgroupMemoryNames, this.samples[10], 5);
         newState.topServicesMemory = topServicesMemory.map(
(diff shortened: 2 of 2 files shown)


martinpitt:
[this](https://github.com/cockpit-project/cockpit/pull/18030/checks?check_run_id=10069996027) is a "fun" COPR build failure: 

> Submit of the build failed: Cmd('git') failed due to: exit code(128) cmdline: git fetch -v origin +refs/pull/18030/head:refs/remotes/origin/pr-changes/18030 stderr: 'fatal: couldn't find remote ref refs/pull/18030/head'

Let's try that again..

test/verify/check-metrics
```diff
@@ -1026,6 +1026,52 @@ BEGIN {{
 
         m.execute("rm /tmp/hogged")
 
+        # Test Podman containers
+        container_name = "pod-mem-hog"
+        m.execute(f"podman run --rm -d --name {container_name} quay.io/libpod/busybox /bin/sh -c 'cat /dev/zero | head -c 300m | tail | sleep infinity'")
```
martinpitt, line 1031: useless use of cat :cat2: 

So that is the POSIX shell way of allocating 300 MB of RAM? l33t! Doesn't busybox have `dd`?

test/verify/check-metrics
```diff
@@ -1026,6 +1026,52 @@ BEGIN {{
...
+
+        # On some test images the container takes a while to show up
+        with b.wait_timeout(300):
+            b.wait_text("table[aria-label='Top 5 memory services'] tbody tr:nth-of-type(1) td[data-label='Service']", f"pod {shortid}")
+
+        # It takes one re-render for the name lookup
+        with b.wait_timeout(30):
+            b.wait_text("table[aria-label='Top 5 memory services'] tbody tr:nth-of-type(1) td[data-label='Service']", f"pod {container_name}")
+
+        m.execute(f"podman stop -t 0 {container_name}")
```
martinpitt, line 1044: Does `-t 0` exist on RHEL 8? ISTR that it doesn't, in console.dot I had to add a few hacks for that. Well, tests will tell!

test/verify/check-metrics
```diff
@@ -1026,6 +1026,52 @@ BEGIN {{
...
+podman save quay.io/libpod/busybox | sudo -i -u admin podman load
+            """)
+
+            # Test user containers
+            admin_s = ssh_connection.SSHConnection(user="admin",
+                                                   address=m.ssh_address,
+                                                   ssh_port=m.ssh_port,
+                                                   identity_file=m.identity_file)
+            user_container_name = "user-mem-hog"
+            admin_s.execute(f"podman run --rm -d --name {user_container_name} quay.io/libpod/busybox /bin/sh -c 'cat /dev/zero | head -c 300m | tail | sleep infinity'")
```
martinpitt, line 1060: Lines really get long, can you please split this into a multi-line string?

martinpitt (approved):
Nice, thank you! Good enough for a nice release-blocker fix, but if you have to push again for some reason, there are some minor cleanups.

martinpitt:
Hmm, do you mind trying to rebase this, to pacify the packit copr build? Happy to reapprove quickly.

test/verify/check-metrics
jelly: It has dd, what did you envision?

test/verify/check-metrics
jelly: We already had `-t 0` in the CPU tests so this should be fine.

test/verify/check-metrics
jelly: As `dd` is used for the cpu hog test so hmm? I stole the solution from [SO](https://unix.stackexchange.com/a/254976) and adjusted it to work with busybox sh. The `tail` is required to keep the data into memory which is quite interesting :)

jelly:
> Hmm, do you mind trying to rebase this, to pacify the packit copr build? Happy to reapprove quickly.

Voila

martinpitt:
Multiple failures of `TestCurrentMetrics.testMemory`

test/verify/check-metrics
martinpitt: I meant `dd if=/dev/zero bs=1m count=300` to read 300 MB of zeros. But indeed `head -c 300m < /dev/zero` is just as good, I wasn't aware that `head` can deal with prefixes as well. TIL!

(Nice trick with tail|sleep)

jelly:
> Multiple failures of `TestCurrentMetrics.testMemory`

There is a race in looking up the pod name, we first check if there is a pod id, but as the process directly consumes memory when it first appears it's name is already resolved. I've added a sleep, so this might still bit a bit brittle but locally this seems to pass 10 times in a row.

martinpitt (dismissed):
Meh @ sleep, and the lines are still fairly long, but good enough really -- let's get that in, it's a nice fix for today's releases. Cheers!

jelly:
> Meh @ sleep, and the lines are still fairly long, but good enough really -- let's get that in, it's a nice fix for today's releases. Cheers!

Yeah, maybe we should just only check for the end result, if the name is resolved. We don't necessarily care about the podid being shown.

test/verify/check-metrics
```diff
@@ -1026,6 +1026,41 @@ BEGIN {{
...
+
+        m.execute(f"podman stop -t 0 {container_name}")
+
+        # RHEL-8 / CentOS-8's podman user containers do not show up as
+        # libpod-$containerid but as podman-3679.scope.
+        if m.image != "centos-8-stream" and not m.image.startswith("rhel-8"):
+            # copy images for user podman tests; podman insists on user session
+            m.execute("""
+podman save quay.io/libpod/busybox | sudo -i -u admin podman load
+            """)
```
martinpitt, line 1046: This looks ugly, and a multi-line string is excessive here.

test/verify/check-metrics
```diff
@@ -1026,6 +1026,41 @@ BEGIN {{
...
+
+            # Test user containers
+            admin_s = ssh_connection.SSHConnection(user="admin",
+                                                   address=m.ssh_address,
+                                                   ssh_port=m.ssh_port,
+                                                   identity_file=m.identity_file)
+            user_container_name = "user-mem-hog"
+            admin_s.execute(f"""
+podman run --rm -d --name {user_container_name} quay.io/libpod/busybox /bin/sh -c 'head -c 300m /dev/zero | tail | sleep infinity'
```
martinpitt, line 1055: It doesn't hurt to properly indent the multi-line string, the shell doesn't care about space prefixes. Alternatively you can also do something like
```py
    execute('first half | '
            'second half')
```

which IMHO looks quite pleasant and also doesn't mess up the actual command with spaces (for proper verbose logging).

martinpitt (approved):
Thanks for figuring this out! Two style nitpicks, but I leave it up to you if you want to land as-is, or fix this (happy to re-approve quickly)

PR #18034, opened 2022-12-14
test: Speed up CI runs with our jumpstart cache

martinpitt:
Only build changed webpacks. This takes off some 5 minutes (and heavy CPU load) for every test from the bots.

This was already applied in commit a7dcaf554dc6 half a year ago, but reverted in commit 0b25285522 because `webpack-jumpstart --partial` did not handle node_module/ updates correctly. But commit 3a77525f8e4d6a6e fixed that, it now removes all webpacks which are affected by a node_modules/ update (usually all of them).

-----

I validated this with the most recent PF update:

```sh
# check out latest package.json change
git checkout c49fc4b1d544c10375f9040fb63813d0784d61ea
# change it to make it not have a jumpstart cache
git commit --amend --no-edit
# cherry-pick the commit following the above update (SELinux, does not affect webpacks)
git cherry-pick 755d372e50395256de6914fa6a72691c57d55160
```

Now jumpstart DTRT:
```
$ tools/webpack-jumpstart --partial
  FETCH    dist  [ref: tag sha-776460684e6bdf81542934525e0f9272b21c3e5a]
  COPY     package-lock.json
  UNPACK   dist  [ref: sha-776460684e6bdf81542934525e0f9272b21c3e5a]
  REMOVE   dist/apps
  REMOVE   dist/base1
  REMOVE   dist/kdump
  REMOVE   dist/metrics
  REMOVE   dist/networkmanager
  REMOVE   dist/packagekit
  REMOVE   dist/playground
  REMOVE   dist/selinux
  REMOVE   dist/shell
  REMOVE   dist/sosreport
  REMOVE   dist/static
  REMOVE   dist/storaged
  REMOVE   dist/systemd
  REMOVE   dist/tuned
  REMOVE   dist/users

$ ls -l dist/
# total 0
```

Earlier version didn't:
```
$ git clean -ffdx
$ git show 3a77525f8e4d6a6ee7c99341846a543a6dab0a66^:tools/webpack-jumpstart | sed '/^cd/d' > /tmp/wj
$ sh /tmp/wj --partial
  FETCH    dist  [ref: tag sha-776460684e6bdf81542934525e0f9272b21c3e5a]
  COPY     package-lock.json
  UNPACK   dist  [ref: sha-776460684e6bdf81542934525e0f9272b21c3e5a]
  TOUCH    package.json

# ... and all of dist/ is intact
```

### DIFF
--- a/tools/webpack-jumpstart
+++ b/tools/webpack-jumpstart
@@ -104,6 +104,10 @@ if test -n "${partial}"; then
             tag="sha-${commit}"
             break
         fi
+        if git show --oneline --name-only "$commit" | grep -q '^node_modules'; then
+            echo "commit $commit changes node_modules" >&2
+            break
+        fi
     done
     if [ -z "${tag}" ]; then
         echo "Unable to find any suitable commit" >&2
--- a/test/run
+++ b/test/run
@@ -7,6 +7,7 @@ set -eu
 
 tools/make-bots
 test/common/pixel-tests pull
+tools/webpack-jumpstart --partial || true
 
 TEST_SCENARIO=${TEST_SCENARIO:-verify}
 


martinpitt:
Now the [build log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18034-20221214-050359-0a416dbc-fedora-coreos/log.html) uses the jumpstart cache from current main HEAD:
```
  FETCH    dist  [ref: tag sha-a0d164579ee57ba41e994bd6d7277c86750202eb]
```

martinpitt:
This works fine now. We could add a second line of defense and change webpack-jumpstart to stop iterating over past commits once a commit touches `node_modules`. Looking into that.

martinpitt:
> We could add a second line of defense and change webpack-jumpstart to stop iterating over past commits once a commit touches `node_modules`.

Done. It now looks like this:
```
$ git clean -ffdx; tools/webpack-jumpstart --partial; echo $?
Removing dist/
Removing package-lock.json
commit aa13652b285d6f3b3ea0f7ca3b210c3eb10dc177 changes node_modules/, stopping
Unable to find any suitable commit
1
```

martinpitt:
And again the [build log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18034-20221214-052216-1defd25e-fedora-37/log.html) shows that it's still using the partial cache.

tools/webpack-jumpstart
```diff
@@ -100,6 +100,10 @@ if test -n "${partial}"; then
     # fetch the available commits, then iterate our history looking for the first match
     available="$(git_cache ls-remote origin 'refs/tags/sha-*' | cut -f2 -d-)"
     for commit in $(git log -n 100 --format=%H); do
+        if git show --oneline --name-only "$commit" | grep -q '^node_modules'; then
```
allisonkarlitskaya, line 103: I think there's an off-by-one error here: if we have a jumpstart build for this commit (ie: the commit that made the change, ie: directly after the change) then we should use it.

So put this check below the "if available ..." below, right?

allisonkarlitskaya (commented):
There's nothing wrong here, to be quite honest, but it still feels a bit like crossing the rubicon.  I don't want to say that I know of any bugs in the partial jumpstart code (and indeed, I don't know of any) but it makes me feel less sure that we're testing what we think we're testing.

#17462 is a perfect example of how another bug in the build system can lead to strange results when we're not building "fresh".

tools/webpack-jumpstart
martinpitt: Whoops, you are right of course. Fixed.

allisonkarlitskaya (approved):
Let's give it a go.  I actually think it's a good idea.

PR #18035, opened 2022-12-14
test: Fix TestNetworkingBasic.testBasic for stopped pcp

martinpitt:
The zoom and range selector on the Networking metrics is only visible if PCP is running. Make sure that is the case, as
https://github.com/cockpit-project/bots/pull/4159 will disable it by default.


martinpitt:
I triggered an extra test against the fedora-36 image from https://github.com/cockpit-project/bots/pull/4159

Issue #16693, opened 2021-12-07
shell: Port host switcher to PF4 dropdown


garrett:
Follow-up to:
- https://github.com/cockpit-project/cockpit/pull/16614

Related to: 
- https://github.com/cockpit-project/cockpit/issues/14304

Default style of host switcher will change, but that can be in a separate PR from the port. Or it can be in the same PR. But both should land in the same release.

Changing the dropdown means we'll need to change how host management is handled. Perhaps in a dialog? As hosts are already edited in a dialog, that means we'll have to figure out how to add, remove, and edit hosts from within in the same dialog... and this means we'll probably need a design for it.

garrett:
Here's the mockup from a few years ago:

![unified-sidebar-host-switcher-few-isolated](https://user-images.githubusercontent.com/10246/212878359-170b7d94-ef56-4585-9eee-1cc08046bf7e.png)

Longer version, with a scrollbar:

![unified-sidebar-host-switcher-many-isolated](https://user-images.githubusercontent.com/10246/212878665-8b963f45-b4e7-4ca1-b207-a6a563ef29ab.png)

This was from the PF3 era, although you can't tell it so much from the mockup. It does have dashboards listed and doesn't have a way to edit (aside from the dashboard, which we haven't had for a very long time).

In context, it looked like this:

![sidebar-double-host-switcher](https://user-images.githubusercontent.com/10246/212879518-a0c349f7-3c2b-426d-b30a-a32c6ed34e6e.png)

The overall layout of Cockpit changed, as we removed the buttons, shuffled around apps, and moved the host switching to be top-level and integrated with the topbar. Navigation became even more unified. (A very, very long time ago Cockpit used to have split navigation with disconnected navigation above and to the side. This mockup is intermediate.)

The reason why we went with the implementation we currently have is to make editing work. We'd need to figure out a way to have editing (and deletion) work from a popover. The issue is that it would go away once editing starts, so editing and switching would be cumbersome, especially when editing multiple hosts.

One possible way of addressing editing in a popover is to have inline edit functionality. We'd want to capture misclicks outside the popover when editing, so the modal doesn't accidentally go away in the middle of an edit.

I think I'd need to work on some mockups for this, but I'd need someone interested in working on implementing this with me too.

martinpitt:
Host switcher is deprecated.

PR #18036, opened 2022-12-14
shell: Use createRoot as mandated by React 18

mvollmer:


mvollmer:
Oops, I misunderstood what `root.unmount()` does.

I think we should also work on reducing the number of `render` calls, by rewriting this whole lot into straight React components.

jelly:
> Oops, I misunderstood what `root.unmount()` does.
> 
> I think we should also work on reducing the number of `render` calls, by rewriting this whole lot into straight React components.

I believe we have https://github.com/cockpit-project/cockpit/issues/16693 for that.

KKoukiou (approved):

PR #18037, opened 2022-12-14
Right-to-Left fixes, round 1

garrett:


garrett:
I tried to find the most obvious RTL problems and attempted to solve them at a generic level. Most of these are ones we inherit from PatternFly.

garrett:
This is the first step to identify and workaround PatternFly's RTL problems. Hopefully we can use this (and upcoming fixes) as a guideline to identify what PatternFly needs to fix, so we can more properly fix it upstream too.

PatternFly's RTL issue is at: https://github.com/patternfly/patternfly/issues/3812

(We do have a Cockpit-specific issue fixed in this PR as well, but most of these are PatternFly issues.)

garrett:
(Apologies for this coming in later than expected. There's just one of me. We are still lacking an Andreas.)

PR #18039, opened 2022-12-14
test: Skip failing tests for pybridge scenario

martinpitt:
Let's get the scenario to green, and re-enable tests as we fix them.

---

https://issues.redhat.com/browse/COCKPIT-882

 - [reference run on  current main](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221214-143444-aa8f6793-fedora-36-pybridge/log.html)
 - [non-current successful run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221216-132659-3b488e57-fedora-36-pybridge/log.html), [another one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221216-142401-b4036bbe-fedora-36-pybridge/log.html)

Depending PRs:
 - #18041
 - #18043 
 - #18044
 - #18046
 - #18049
 - #18045

Noteworthy test failures:
 - [x] [TestCurrentMetrics.testDiskIO](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-064658-7fd5dc02-fedora-36-pybridge/log.html#116): "wrong username or password". This causes a [JSONDecodeError bridge crash](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-165208-9ff13c4a-fedora-36-pybridge/TestCurrentMetrics-testDiskIO-fedora-36-127.0.0.2-2301-FAIL.log.gz). Fixed in a commit here and #18049

 - [x]  [TestSystemInfo.testBasic](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-165208-9ff13c4a-fedora-36-pybridge/log.html#274) causes a `FileNotFoundError` crash in the bridge. Fixed in a commit here and #18049

 - [x] There ABRT tests fail on "Cockpit shows an Oops". I can reproduce this on `TestJournal.testAbrtDelete` where it says:

        TypeError: Cannot read properties of undefined (reading 'ID')\n    at http://127.0.0.2:9091/cockpit/@localhost/system/logs.js:2:551352\n

    Fixed in #18044, and also included here.

 - [x] Lots of test crash with a JS exception "RuntimeError: RangeError: Invalid time value" in services.js:23203:66  at "Module.dateTime" → ServicesPageBody.addTimerProperties (useful trace with `NODE_ENV=development`). I can reproduce with `TestAutoUpdates.testBasic` and with `TestAD.testQualifiedUsers`; the problem is that it tries to call `Intl.DateTimeFormat('en-US').format(18446744073709552000/1000)` which is invalid and causes that bug. The issues is that `timer_unit.NextElapseUSecMonotonic == 18446744073709552000 == 0x10000000000000180` is clearly bogus. Fixed in #18041 and here.

 - [x] [TestUpdates.testInfoSecurity](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-165208-9ff13c4a-fedora-36-pybridge/log.html#163) is a flake, added a commit to fix it here and in #18049

- [x] [TestUpdates.testRebootAfterSuccess](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-165208-9ff13c4a-fedora-36-pybridge/log.html#168): attempted a fix in #18049 by waiting longer, but that's not enough; the "Initializing.." is really stuck. This reproduces locally in about 50% of cases. The only hint of failure is that "warning: python.error" in the JS log; packagekitd.service, `pkcon get-transactions` are fine. But interestingly, reloading the page gets into the same "stuck" state again, and `pkcon get-updates` is stuck as well, which makes this nice to iterate. Even after logging out from cockpit, `pkcon get-updates` is stuck, so something thoroughly killed PackageKit. Skipped for now.

- [x] [TestTimers.testCreate](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221216-090333-05d0b1aa-fedora-36-pybridge/log.html#319) is flaky, and sometimes crashes with a bridge PermissionError; not sure if that's the user or the root bridge. Skipped for now.

 - [x] [TestAccounts.testExpire](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221216-090333-05d0b1aa-fedora-36-pybridge/log.html#325) sometimes fails with "internal error", but there's no bridge crash in the logs. Needs deeper debugging, but this isn't just a matter of waiting longer. Skipped for now.

- [x] There are several flakes. They should be dealt with by re-enabling auto-retries: `TestKdump.testBasic`, `TestAutoUpdates.testBasic`,  `TestSelinux.testTroubleshootAlerts`, `TestTimers.testCreate`, `TestReverseProxy.testNginxNoTLS` . Re-enabling auto-retries of failures helps enough.

 - Calling methods on D-Bus proxies is broken.  Simplest reproducer is to go to /playground/pkgs and press the reload button. This also breaks ABRT. Tests are skipped, but this is pretty basic functionality. @mvollmer

 - several tests fail and just have "warning: python.error" in the journal. We need to show these errors properly

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -76,6 +76,7 @@ __all__ = (
     'skipBrowser',
     'skipPackage',
     'todo',
+    'todoPybridge',
     'enableAxe',
     'timeout',
     'Error',
@@ -1346,6 +1347,9 @@ class MachineCase(unittest.TestCase):
             else:
                 if machine.image == reference_image:
                     pixels_label = self.label()
+        # HACK: until @todoPybridge disappears from all pixel tests
+        if os.environ.get("TEST_SCENARIO") == "pybridge":
+            pixels_label = None
         browser = Browser(machine.web_address,
                           label=label, pixels_label=pixels_label, coverage_label=self.label() if coverage else None,
                           port=machine.web_port, machine=self)
@@ -1759,6 +1763,78 @@ class MachineCase(unittest.TestCase):
             self.allowed_messages.append("Failed to connect to coredump service: No such file or directory")
             self.allowed_messages.append("Failed to connect to coredump service: Connection refused")
 
+        # HACK: pybridge bugs
+        if os.environ.get("TEST_SCENARIO") == "pybridge":
+            self.allowed_messages += [
+                "ERROR:asyncio:Task was destroyed but it is pending!",
+                "task:.*Task pending.*cockpit/channels/dbus.py.*"]
+
+            self.allowed_messages += [
+                r"Exception ignored on calling ctypes callback function: <function Slot.__init__.<locals>.handler.*",
+                r"asyncio.exceptions.InvalidStateError: invalid state",
+                r"ERROR:asyncio:Exception in callback _Transport._read_ready.*",
+                r"handle: <Handle _Transport._read_ready\(\)>",
+                r"Traceback \(most recent call last\):",
+                r"File .*asyncio/events.py.*",
+                r"self._context.run\(self._callback, \*self._args\)",
+                r"File .*cockpit/transports.py.* _read_ready",
+                r"data = os.read\(self._in_fd, _Transport.BLOCK_SIZE\)"]
+
+            self.allowed_messages += [
+                r"File .*/systemd_ctypes/bus.py.* in handler",
+                r"return 1 if callback.*BusMessage.ref.* else 0",
+                r"File .*/systemd_ctypes/bus.py.* in done",
+                r"future.set_result\(message\)"]
+
+            self.allowed_messages += [
+                r"ERROR:asyncio:Exception in callback BusMessage._coroutine_task_complete.*",
+                r"handle: .*Handle BusMessage._coroutine_task_complete.*",
+                r"File .*/systemd_ctypes/bus.py.* in _coroutine_task_complete",
+                r"self.reply_method_function_return_value.*",
+                r"File .*/cockpit/superuser.py.*",
+                r"await startup.wait.*",
+                r"await self.future",
+                r"File .*/cockpit/transports.py.*",
+                r"n_bytes = os.write\(self._out_fd, data\)",
+                r"BrokenPipeError: \[Errno 32\] Broken pipe"]
+
+            self.allowed_messages += [
+                r"while .*result := self.consume_one_frame.*",
+                r"self._protocol.data_received\(data\)",
+                r"File .*/cockpit/protocol.py.*",
+                r"self.frame_received.*",
+                r"File .*/cockpit/channel.py.*",
+                r"self.channel_control_received.*",
+                r"endpoint.do_channel_control.*",
+                r"File .*/cockpit/router.py.*",
+                r"self.do_control.*",
+                r"self.do_done.*",
+                r"File .*/cockpit/channels/packages.py.*",
+                r"self.router.packages.serve_file.*",
+                r"File .*/cockpit/packages.py.*",
+                r"self.serve_package_file.*",
+                r"self.packages.*.serve_file.*",
+                r"KeyError: 'manifests.json'"]
+
+            self.allowed_messages.append(".* is not in the sudoers file.  This incident will be reported.")
+            self.allowed_messages.append("sudo: no valid sudoers sources found, quitting")
+
+            # TestSuperuser.testWrongPasswd, message should go to the caller, not the journal
+            self.allowed_messages.append("Sorry, try again.")
+            # likewise for TestAccounts.testBasic; should not go to journal
+            self.allowed_messages.append("sudo: no password was provided")
+            self.allowed_messages.append("sudo: .* incorrect password attempt")
+
+            # TestSuperuserOldWebserver.test{,NotAuth}
+            self.allowed_messages.append("session timed out")
+            self.allowed_messages.append("cockpit-ssh.*: refusing to connect to unknown host:.*")
+
+            # TestJournal.testAbrtSegv
+            self.allowed_messages.append("invalid non-UTF8 @data passed as text to web_socket_connection_send.*")
+
+            # TestLogin.testSessionRecordingShell
+            self.allowed_messages.append(r"future.set_exception\(error\)")
+
         messages = machine.journal_messages(matches, 6, cursor=cursor)
 
         if "TEST_AUDIT_NO_SELINUX" not in os.environ:
@@ -2134,6 +2210,13 @@ def todo(reason='', flaky=False):
         testEntity._testlib_todo = (reason, flaky)
         return testEntity
     return wrapper
+
+
+def todoPybridge(reason=None, flaky=False):
+    if os.getenv('TEST_SCENARIO') == 'pybridge':
+        return todo(reason or 'still fails with python bridge', flaky)
+    else:
+        return lambda testEntity: testEntity
 
 
 def checkRunAxe():
--- a/test/run
+++ b/test/run
@@ -30,7 +30,7 @@ case $TEST_SCENARIO in
         test/image-prepare --verbose $TEST_OS
         python3 -m build --no-isolation --wheel --outdir tmp/wheel
         bots/image-customize --verbose --no-network --install tmp/wheel/*.whl "${TEST_OS}"
-        test/common/run-tests --jobs ${TEST_JOBS:-1} --no-retry-fail --test-dir test/verify --track-naughties
+        test/common/run-tests --jobs ${TEST_JOBS:-1} --test-dir test/verify --track-naughties
         ;;
     *)
         echo "Unknown test scenario: $TEST_SCENARIO"
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -23,7 +23,7 @@ import time
 
 import parent  # noqa: F401
 from packagelib import PackageCase
-from testlib import nondestructive, skipDistroPackage, skipImage, test_main
+from testlib import nondestructive, skipDistroPackage, skipImage, todoPybridge, test_main
 
 
 WAIT_SCRIPT = """
@@ -282,6 +282,7 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
         m.execute("rpm -q kpatch-patch-" + sanitized_kernel_ver)
 
     @nondestructive
+    @todoPybridge()
     def testBasic(self):
         # no security updates, no changelogs
         b = self.browser
@@ -435,6 +436,7 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
 
     @skipImage("TODO: Packagekit on Arch does not detect the pear update", "arch")
     @skipImage("tracer not available", *OSesWithoutTracer)
+    @todoPybridge(".updates-success-table randomly appears or not with pybridge", flaky=True)
     def testTracer(self):
         b = self.browser
         m = self.machine
@@ -980,6 +982,7 @@ ExecStart=/usr/local/bin/{packageName}
 
         # seems we can't verify that the description has a scrollbar
 
+    @todoPybridge("sometimes locks up packagekit hard, warning: python.error in the JS log", flaky=True)
     def testRebootAfterSuccess(self):
         b = self.browser
         m = self.machine
@@ -1043,6 +1046,7 @@ ExecStart=/usr/local/bin/{packageName}
         self.assertFalse(b.is_present("#app button"))
 
     @nondestructive
+    @todoPybridge()
     def testPackageKitCrash(self):
         b = self.browser
         m = self.machine
@@ -1080,6 +1084,7 @@ ExecStart=/usr/local/bin/{packageName}
         b.wait_in_text("#app .pf-c-page .pf-c-empty-state__body", "PackageKit crashed")
 
     @nondestructive
+    @todoPybridge()
     def testNoPackageKit(self):
         b = self.browser
         m = self.machine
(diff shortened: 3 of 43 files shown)


martinpitt:
Good news: `@expectedFailure` DTRT for classes: With this
```diff
--- test/verify/check-testlib
+++ test/verify/check-testlib
@@ -268,6 +268,7 @@ class TestTestlib(MachineCase):
 
 
 @skipDistroPackage()
+@unittest.expectedFailure
 class NoTestRetryExample(unittest.TestCase):
 
     def testFail(self):
```

```
❱❱❱ TEST_FAILURES=1 test/verify/check-testlib  NoTestRetryExample
# ----------------------------------------------------------------------
# testBasic (__main__.NoTestRetryExample.testBasic)
# Result testBasic (__main__.NoTestRetryExample.testBasic) failed
# ----------------------------------------------------------------------
# testFail (__main__.NoTestRetryExample.testFail)
# Result testFail (__main__.NoTestRetryExample.testFail) succeeded
# ----------------------------------------------------------------------
# testNoRetry (__main__.NoTestRetryExample.testNoRetry)
# Result testNoRetry (__main__.NoTestRetryExample.testNoRetry) failed
# 2 TESTS FAILED [0s on toolbox]
```

and if I change them to all fail:
```diff
--- test/verify/check-testlib
+++ test/verify/check-testlib
@@ -268,6 +268,7 @@ class TestTestlib(MachineCase):
 
 
 @skipDistroPackage()
+@unittest.expectedFailure
 class NoTestRetryExample(unittest.TestCase):
 
     def testFail(self):
@@ -276,10 +277,10 @@ class NoTestRetryExample(unittest.TestCase):
 
     @no_retry_when_changed
     def testNoRetry(self):
-        self.assertTrue(True)
+        self.assertFalse(True)
 
     def testBasic(self):
-        self.assertTrue(True)
+        self.assertFalse(True)
 
 
 if __name__ == '__main__':
```

```
❱❱❱ TEST_FAILURES=1 test/verify/check-testlib  NoTestRetryExample
# ----------------------------------------------------------------------
# testBasic (__main__.NoTestRetryExample.testBasic)
# Result testBasic (__main__.NoTestRetryExample.testBasic) succeeded
# ----------------------------------------------------------------------
# testFail (__main__.NoTestRetryExample.testFail)
# Result testFail (__main__.NoTestRetryExample.testFail) succeeded
# ----------------------------------------------------------------------
# testNoRetry (__main__.NoTestRetryExample.testNoRetry)
# Result testNoRetry (__main__.NoTestRetryExample.testNoRetry) succeeded
# 3 TESTS PASSED [0s on toolbox]
```

Of course we still need to conditionalize this on the pybridge, but we can (hopefully) build on the existing decorator.

martinpitt:
[Next iteration](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221214-153452-f60436f5-fedora-36-pybridge/log.html#18), the approach generally works, but now it is not recognizable *at all* that a test was an expected failure, as opposed to a success. I added a commit that makes that a little more obvious, although not in log.html. But for identifying commonalities, looking at the raw `log` in vim is probably better anyway.

martinpitt:
@marusak, @allisonkarlitskaya , all: I'm afraid the pybridge isn't robust enough to reliably pass or fail tests. E.g. `TestEmbed.testBasic` passed in the [initial reference run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221214-143444-aa8f6793-fedora-36-pybridge/log.html), but now [fails in the next one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221214-162912-dd2d659c-fedora-36-pybridge/log.html#8). 18 test fail on

    umount: /var/lib/cockpittest/repo/chocolate-2.0-2.x86_64.rpm: target is busy.

or

   umount: /run/user/1000: target is busy.

which is some kind of leak from a previous test, I presume. 

Also, some tests now fail on "Cockpit shows an Oops", despite being marked as expected failure (as these checks happen after the actual `test..()` function in common teardown).

So we'll have some work to do for fixing the leaks first, and probably have to move to skips for the unstable tests which don't reliably pass or fail.

martinpitt:
We also need a different approach for the unexpected messages. In particular, these two randomly break about 40 tests:
```
ERROR:asyncio:Task was destroyed but it is pending!
task: <Task pending name='Task-12' coro=<DBusChannel.do_call() running at /usr/lib/python3.10/site-packages/cockpit/channels/dbus.py:248> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[set.discard()]>

Exception ignored on calling ctypes callback function: <function Slot.__init__.<locals>.handler at 0x7fb6edbfb6d0>
File "/usr/lib/python3.10/site-packages/systemd_ctypes/bus.py", line 280, in handler
return 1 if callback(BusMessage.ref(message)) else 0
File "/usr/lib/python3.10/site-packages/systemd_ctypes/bus.py", line 298, in done
future.set_result(message)
asyncio.exceptions.InvalidStateError: invalid state
ERROR:asyncio:Exception in callback _Transport._read_ready()
handle: <Handle _Transport._read_ready()>
```

I added message filters.

martinpitt:
- [ ] [TestLongRunning.testBasic](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-064658-7fd5dc02-fedora-36-pybridge/log.html#10) "Error: unexpected failure of GetUnit(cockpit-longrunning.service): Not permitted to perform this action.":  Not reproducible locally easily, this is either a flake or unclean VM
 - [ ] [TestCurrentMetrics.testDiskIO](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-064658-7fd5dc02-fedora-36-pybridge/log.html#116): "wrong username or password" on login makes no sense, must be a leak from previous test
 - [ ] There are a bunch of tests failing on "Cockpit shows an Oops", but the [JS log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-064658-7fd5dc02-fedora-36-pybridge/TestHistoryMetrics-testNoDataEnable-fedora-36-127.0.0.2-2401-FAIL.js.log) is utterly useless: it just says `warning: python.error`; I can reproduce this on `TestJournal.testAbrtDelete` where it says:

        TypeError: Cannot read properties of undefined (reading 'ID')\n    at http://127.0.0.2:9091/cockpit/@localhost/system/logs.js:2:551352\n

 - [ ] [TestRunTest.testExistingMachine](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221215-064658-7fd5dc02-fedora-36-pybridge/log.html#61): not reproducible locally

[many more, to be continued..]

mvollmer:
Would it help to disable the @nondestructive optimizations and run each test in a fresh VM?

martinpitt:
@mvollmer : Maybe a little bit, but I already fixed the leaked mount; most stuff that I've stared at in the last hour is unrelated to nondestructiveness. There's still tons of things to look at before, at least.

test/verify/check-packagekit
```diff
@@ -368,26 +368,29 @@ echo -e "Loaded patch modules:\nkpatch_3_10_0_1062_1_1 [enabled]\n\nInstalled pa
...
+            b.click(".progress-main-view button.pf-m-secondary")
+            # abort the current download, so that read calls don't hang indefinitely
+            m.spawn(f"echo > {self.vm_tmpdir}/fifo", "fifo")
+
+            # going back to overview, nothing happened just yet
+            b.wait_not_present(".progress-main-view")
+            self.assertEqual(b.text("#available-updates button#install-all"), "Install all updates")
+            self.assertFalse(b.is_present("table.updates-history"))
+            m.execute("test -f /stamp-vanilla-1.0-1; test -f /stamp-chocolate-2.0-1")
+        finally:
```
allisonkarlitskaya, line 389: I feel like these ought to be cleanups, but I guess the point is that you want to do them at this point already?

test/common/run-tests
```diff
@@ -165,7 +165,7 @@ class Test:
 
         # be quiet in TAP mode for successful tests
         lines = self.output.strip().splitlines(keepends=True)
-        if print_tap and self.returncode == 0 and len(lines) > 0:
+        if print_tap and self.returncode == 0 and len(lines) > 0 and b'# WARNING: expectedFailure' not in self.output:
```
allisonkarlitskaya, line 168: I was wondering if we could get some orange or purple or something for these on the logs page...

test/common/testlib.py
```diff
@@ -1339,6 +1340,9 @@ def new_browser(self, machine=None, coverage=False):
             else:
                 if machine.image == reference_image:
                     pixels_label = self.label()
+        # HACK: until @expectedFailurePybridge disappears from all pixel tests
+        if os.environ.get("TEST_SCENARIO") == "pybridge":
```
allisonkarlitskaya, line 1344: This is slightly incorrect.  It might be the case that we prepare the image with the bridge, but then run tests manually, without setting the scenario...

It's probably "good enough", though.  In fact, it's probably exactly the right sort of pragmatic solution.  After all, it's a hack.

test/run
```diff
@@ -9,7 +9,7 @@ tools/make-bots
 test/common/pixel-tests pull
 tools/webpack-jumpstart --partial || true
 
-TEST_SCENARIO=${TEST_SCENARIO:-verify}
+export TEST_SCENARIO=${TEST_SCENARIO:-verify}
```
allisonkarlitskaya, line 12: Why this?

allisonkarlitskaya (changes requested):
I have no substantial issues with this, and I think we ought to land it asap.

Outstanding TODOs (both against bots):
 - better display of xfail on logs pages
 - testmap update to switch this on

It also needs a rebase to remove the commit disabling the unit tests and deal with the fact that I landed the partial jumpstart PR already.  That's the only reason I'm marking this requested changes.

test/run
martinpitt: Because `run-tests` and test-foo need to be able to see this; it can't just be internal to `test/run`.

test/common/run-tests
martinpitt: Yes, that'd be nice -- we need to do that in logs.html indeed, as a new recognized state.

test/common/testlib.py
martinpitt: Agreed it's a hack. All of that is slated to go away entirely, so I'm not overly worried about it right now. Alternatively the test could also do something like`"python" in m.execute("file /usr/bin/cockpit-bridge")`, but that's quite expensive. Better ideas?

allisonkarlitskaya: I mean it — I think this is fine.  No need to change it.

test/run
allisonkarlitskaya: But in the case that it is set externally, it'll already get wired through, no?  And in the case that it's set internally (ie: the "verify" scenario) then we don't need it to be wired through since the `.get()` call will return `None`.

I guess it's probably more consistent this way, though.

martinpitt: Ah, I see what you mean -- right, we could drop it. Committed locally, will be in the the next round of rebasing/push.

allisonkarlitskaya: I actually don't mind to keep it, now that you explained it :)

allisonkarlitskaya (commented):
For what it's worth, I was just about to look at implementing this in the `log.html` parser, and I decided to take a look at the TAP spec: https://testanything.org/tap-specification.html

> TODO tests
>
> If the directive starts with # TODO, the test is counted as a todo test, and the text after TODO is the explanation.
>
> `not ok 13 # TODO bend space and time`
>
>Note that if the TODO has an explanation it must be separated from TODO by a space. These tests represent a feature to be implemented or a bug to be fixed and act as something of an executable “things to do” list. They are not expected to succeed. Should a todo test point begin succeeding, the harness should report it as a bonus. This indicates that whatever you were supposed to do has been done and you should promote this to a normal test point.

I wonder if that's a better way to implement this.

test/run
martinpitt: Pushed.

test/verify/check-system-realms
```diff
@@ -86,6 +87,11 @@
 @skipImage("No realmd available", "arch")
 @skipDistroPackage()
 class CommonTests:
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
martinpitt: Yesyes, sorry. Fixed.

martinpitt:
Wohoo, a [successful run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221216-132659-3b488e57-fedora-36-pybridge/log.html)! Pushed another unexpected message fix, and running again to get some data about how stable this is.

martinpitt:
[Second pass in a row](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18039-20221216-142401-b4036bbe-fedora-36-pybridge/log.html)! I'm beginning to be confident about this. Next round, let's re-enable unit and packit tests.

I'd like to discuss the expectedFailure vs. todo with @allisonkarlitskaya on Monday, and #18049 still needs to be reviewed and landed. But we could manage to land this by next Tuesday (my last workday)?

test/common/run-tests
martinpitt: See the combined efforts of #18045 and https://github.com/cockpit-project/bots/pull/4180 , this happens now.

martinpitt:
I rebased this on top of the `@todo` decorator from PR #18045.

allisonkarlitskaya (approved):
One note: due to the way `@todo` tests are implemented, those won't be retried, even for `flaky=True`.  But I think you didn't want them to be, anyway.

martinpitt:
@allisonkarlitskaya : I'm counting on that. Otherwise the tests would run *waaay* too long, that's why we had to use the `--no-retry-fail` hack.

martinpitt:
Wohoo, all green! :guitar:

PR #18043, opened 2022-12-15
python: better exception handling in Transport code

allisonkarlitskaya:
There are a lot of reasons why we might get an exception when trying to read or write on a transport (for example: ECONNRESET), and pretty much all of those should be treated as a disconnection of the transport.

Previously we handled EPIPE on write() — expand handling to all OSError types, and also on read.


martinpitt (approved):
Cheers!

PR #18045, opened 2022-12-15
testlib: add a @todo decorator

allisonkarlitskaya:
Decorated testcases will be marked as "TODO" and report as per the TAP specification.

The test will be counted as a fail, the individual test case run will return non-zero, and run-tests will report "not ok", but the failure will not be counted towards the total number of failures reported as part of the exit status.

This is my counter-proposal to the use of xfail in #18039

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -91,7 +91,6 @@ class Test:
         print_tap = not opts.list
         affected = any([self.command[0].endswith(t) for t in affected_tests])
         retry_reason = ""
-        skip_reason = ""
 
         # Try affected tests 3 times
         if self.returncode == 0 and affected and self.retry_when_affected and self.retries < 2:
@@ -105,9 +104,11 @@ class Test:
             lines = self.output.splitlines()
             skip_reason = lines[-1].strip().decode("utf-8")
             self.output = b"\n".join(lines[:-1])
-
-        if self.returncode in [0, 77]:
             self._print_test(print_tap, skip_reason=skip_reason)
+            return None, 0
+
+        if self.returncode == 0:
+            self._print_test(print_tap)
             return None, 0
 
         if not opts.thorough:
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -121,6 +121,7 @@ class Test:
                 reason = proc.communicate(self.output + ("not ok " + str(self)).encode())[0].strip()
 
                 if proc.returncode == 77:
+                    self.returncode = proc.returncode
                     self._print_test(skip_reason="# SKIP {0}".format(reason.decode("utf-8")))
                     return None, 0
 
@@ -183,12 +184,8 @@ class Test:
             return
 
         print()  # Tap needs to start on a separate line
-        if self.returncode == 0:
-            print(f"ok {self}{retry_reason}")
-        elif skip_reason:
-            print(f"ok {self}{skip_reason}")
-        else:
-            print(f"not ok {self}{retry_reason}")
+        status = 'ok' if self.returncode in [0, 77] else 'not ok'
+        print(f"{status} {self}{retry_reason}{skip_reason}")
         flush_stdout()
 
 
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -41,7 +41,7 @@ def flush_stdout():
 
 
 class Test:
-    def __init__(self, test_id, command, timeout, nondestructive, retry_when_affected, cost=1):
+    def __init__(self, test_id, command, timeout, nondestructive, retry_when_affected, todo, cost=1):
         self.process = None
         self.retries = 0
         self.test_id = test_id
@@ -50,6 +50,7 @@ class Test:
         self.nondestructive = nondestructive
         self.machine_id = None
         self.retry_when_affected = retry_when_affected
+        self.todo = todo
         self.cost = cost
         self.returncode = None
 
@@ -107,6 +108,21 @@ class Test:
             self._print_test(print_tap, skip_reason=skip_reason)
             return None, 0
 
+        # If the test was marked with @todo then...
+        if self.todo is not None:
+            todo_reason, flaky = self.todo
+            if self.returncode == 0 and not flaky:
+                # The test passed, but it shouldn't have.
+                self.returncode = 1  # that's a fail
+                self._print_test(print_tap, todo_reason=f'# expected failure: {todo_reason}')
+                return None, 1
+            else:
+                # The test either failed (always OK) or passed (and flaky=True)
+                # Outputs  'not ok 1 test # TODO ...'
+                #      or  'ok 1 test # TODO ...'
+                self._print_test(print_tap, todo_reason=f'# TODO {todo_reason}')
+                return None, 0
+
         if self.returncode == 0:
             self._print_test(print_tap)
             return None, 0
@@ -155,7 +171,7 @@ class Test:
         nd = f" [ND@{self.machine_id}]" if self.nondestructive else ""
         return f"{self.test_id} {self.command[0]} {self.command[-1]}{cost}{nd}"
 
-    def _print_test(self, print_tap=True, retry_reason="", skip_reason=""):
+    def _print_test(self, print_tap=True, retry_reason="", skip_reason="", todo_reason=""):
         def write_line(line):
             while line:
                 try:
@@ -177,15 +193,17 @@ class Test:
             retry_reason = " " + retry_reason
         if skip_reason:
             skip_reason = " " + skip_reason
+        if todo_reason:
+            todo_reason = " " + todo_reason
 
         if not print_tap:
-            print(retry_reason + skip_reason)
+            print(retry_reason + skip_reason + todo_reason)
             flush_stdout()
             return
 
         print()  # Tap needs to start on a separate line
         status = 'ok' if self.returncode in [0, 77] else 'not ok'
-        print(f"{status} {self}{retry_reason}{skip_reason}")
+        print(f"{status} {self}{retry_reason}{skip_reason}{todo_reason}")
         flush_stdout()
 
 
@@ -319,12 +337,14 @@ def detect_tests(test_files, image, opts):
                     continue
                 nd = getattr(test_method, "_testlib__non_destructive", False)
                 rwa = getattr(test_method, "_testlib__retry_when_affected", True)
+                todo = getattr(test_method, "_testlib_todo",
+                               getattr(test.__class__, "_testlib_todo", None))
                 if getattr(test.__class__, "provision", None):
                     # each additionally provisioned VM costs parallel test capacity
                     cost = len(test.__class__.provision)
                 else:
                     cost = 1
-                test = Test(test_id, build_command(filename, test_str, opts), test_timeout, nd, rwa, cost=cost)
+                test = Test(test_id, build_command(filename, test_str, opts), test_timeout, nd, rwa, todo, cost=cost)
                 if nd:
                     serial_tests.append(test)
                 else:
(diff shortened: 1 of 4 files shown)


martinpitt (approved):
Nice, thanks! So for #18039 we can wrap this into a conditional `@todoPyBridge` decorator. :+1:

martinpitt:
@allisonkarlitskaya : Hmm, your new `TestRunTest.testTodo` fails everywhere. OTOH, `TestSimple.testAddition` failed in the right way, nice! We should still color it a bit differently in log.html, but that's already much better than just claiming they'd be ok. But can you please rename it to `testTodoFailure()` or so, to make it less of a "huh?"

test/common/run-tests
```diff
@@ -110,6 +111,12 @@ class Test:
             self._print_test(print_tap, skip_reason=skip_reason)
             return None, 0
 
+        # If the test was marked with @todo then output a fail line, but treat
+        # the test as passed for purposes of the return status.
+        if self.todo is not None:
+            self._print_test(print_tap, todo_reason=f'# TODO {self.todo}')
+            return None, 0
```
martinpitt, line 118: AFAICS this doesn't retain the nice behaviour of `@expectedFailure` that it would consider passing tests as "unexpectedly failed". I think that's useful especially for the large-scale fixes for pybridge, as it's otherwise hard to figure exactly which tests get fixed.

martinpitt (changes requested):
The "request changes" is mainly for fixing and renaming the tests, but let's also discuss the behaviour.

allisonkarlitskaya:
About the example: my thought was to leave it in the failing state and colour it orange in the bar.  That means that every test run from now on would have one `# TODO` failure.  Any objections?

test/common/run-tests
allisonkarlitskaya: To be honest, I'm not sure.  I feel like we might also want to be able to express flakiness (which we still have some amount of).  Having a strict "this must fail" prevents us from capturing that.

On the other hand, we should absolutely make sure we have a way to track that we don't have TODOs that are accidentally fixed.  Probably the correct thing to do is to output the `# TODO` regardless of the test pass/fail status.  Then we can grep for `^ok .* # TODO` at least.

martinpitt: In #18039 I currently tally 4 skips (due to too heavy flakiness) and 91 expectedFailures (which often apply to a whole class, so in practice it's even more). Also, once I'm done with investigating the remaining 11 failures (on my list today), we can re-enable auto-retries of flakes.

Hence I'd make the common case easier to handle in the future, which I believe is the expectedFailure semantics.

martinpitt:
> every test run from now on would have one # TODO failure. 

You can guard it with `TEST_FAILURES` like `NoTestRetryExample.testFail` does. `TestRunTest.testRetry` exports that env variable and thus it only fails for the inner run-tests invocation inside the test, not the "outside" run-tests that runs the whole cockpit suite. Then we'd have an overall clean test result.

allisonkarlitskaya:
after another round of hacking and trying to stay true to the TAP specification while also respecting @martinpitt's  desire to have passing @todo tests count as fails makes me think that maybe this proposal is wrong.

martinpitt:
What I do like: The `# TODO` TAP representation of an "expected to fail" test. What I don't like: treating passed and failed `# TODO` tests as both successful, as this essentially means "ignore the test", and then we don't even need to run it at all (really nobody looks at green tests).

But nothing stops us from using that TAP representation and apply an "xfail" semantics to it, no?

- `@todo(flaky=False)`, fails -> "not ok testname # TODO reason", does not count as failure
- `@todo(flaky=False)`, passes -> "not ok testname # unexpected success", counts as failure
- `@todo(flaky=True)`, fails -> "not ok testname # TODO reason", does not count as failure
- `@todo(flaky=True)`, passes -> "ok testname"

The "flaky=True" is my concession, which essentially ignores the test result; but we can at least grep for these and fix them first.

martinpitt:
I filed https://github.com/cockpit-project/bots/pull/4180 to better handle `not ok # TODO` tests. Regardless of how exactly we want the semantics here, I think that'd be a good move.

test/common/run-tests
```diff
@@ -154,7 +170,7 @@ class Test:
         nd = f" [ND@{self.machine_id}]" if self.nondestructive else ""
         return f"{self.test_id} {self.command[0]} {self.command[-1]}{cost}{nd}"
 
-    def _print_test(self, print_tap=True, retry_reason="", skip_reason=""):
+    def _print_test(self, print_tap=True, retry_reason="", skip_reason="", todo_reason=None):
```
martinpitt, line 173: This needs to default to `""`, otherwise you'll be printing "None" in line 205. You could of course also print `todo_reason or ""` below, but then please do the same with skip_reason as well?

test/common/run-tests
```diff
@@ -107,6 +108,21 @@ class Test:
...
+            if self.returncode == 0 and not flaky:
+                # The test passed, but it shouldn't have.
+                self.returncode = 1  # that's a fail
+                self._print_test(print_tap, todo_reason=f'# expected failure: {todo_reason}')
+                return None, 1
+            else:
+                # The test either failed (always OK) or passed (and flaky=True)
+                # Outputs  'not ok 1 test # TODO ...'
+                #      or  'ok 1 test # TODO ...'
+                self._print_test(print_tap, todo_reason=f'# TODO {todo_reason}')
```
martinpitt, line 123: This keeps printing the `# TODO` even for successful (flaky) tests. The TAP spec doesn't mention this case, and my recent [log.html change](https://github.com/cockpit-project/bots/pull/4180) doesn't recognize that either, but AFAICS it *should* be fine. It will just appear in the HTML as a normal successful test.

martinpitt (changes requested):
Thanks! This looks nice. I'm not sure why the None vs. "" doesn't actually cause a test failure somewhere else, but supposedly it's just a bad printing.

test/common/testlib.py
```diff
@@ -80,6 +80,7 @@
     'Error',
 
     'sit',
+    'todo',
```
martinpitt, line 83: This would conceptually fit better next to the various `skip*` decorators in the block above.

test/common/run-tests
martinpitt: See [failures like this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18045-20221219-120825-54457608-ubuntu-stable/log.html#154):
image

This both illustrates the `None` issue, and also that this somehow breaks the handling of "SKIP" tests.

test/common/run-tests
```diff
@@ -318,12 +336,13 @@ def detect_tests(test_files, image, opts):
                     continue
                 nd = getattr(test_method, "_testlib__non_destructive", False)
                 rwa = getattr(test_method, "_testlib__retry_when_affected", True)
+                todo = getattr(test_method, "_testlib_todo", None)
```
martinpitt, line 339: While rebasing #18039 to this, I noticed that the decorator does not work against a class, just against an individual `def test_*` method. That's why I check both the class and the method [here](https://github.com/cockpit-project/cockpit/pull/18039/commits/f95244f06a1579c94d54fe19d8a6751733f439ac#diff-fdfce066f8a78233d7423f73a8ab6f0e060584f8623b8558f0b2a0f131b79319R1588). It's not such a big deal, I can rework the PR to add `@todoPybridge` to all individual methods, but if it's not too much hassle, it'd be nice.

martinpitt: Ah, this works for `@nondestructive` because of that `set_obj_attribute()` helper. The `no_retry_when_changed()` decorator iterates over all functions. Consistency R US!

I did a quick check that this works:
```diff
--- test/common/run-tests
+++ test/common/run-tests
@@ -336,7 +336,8 @@ def detect_tests(test_files, image, opts):
                     continue
                 nd = getattr(test_method, "_testlib__non_destructive", False)
                 rwa = getattr(test_method, "_testlib__retry_when_affected", True)
-                todo = getattr(test_method, "_testlib_todo", None)
+                todo = getattr(test_method, "_testlib_todo",
+                               getattr(test.__class__, "_testlib_todo", None))
                 if getattr(test.__class__, "provision", None):
                     # each additionally provisioned VM costs parallel test capacity
                     cost = len(test.__class__.provision)
--- test/common/testlib.py
+++ test/common/testlib.py
@@ -2213,12 +2206,12 @@ def todo(reason='', flaky=False):
     unexpected pass is considered to be an error in that case.  If flaky=True
     then a pass is not considered to be an error.
     """
-    def wrapper(func):
-        func._testlib_todo = (reason, flaky)
-        return func
+    def wrapper(testEntity):
+        testEntity._testlib_todo = (reason, flaky)
+        return testEntity
     return wrapper
 
```
(the second part is just renaming to clarify it could be a class or method).

martinpitt: I added this fix as a [hack commit](https://github.com/cockpit-project/cockpit/pull/18039/commits/758293affd78e4c4a5cedb26755696159f7f3d6d) to #18039.

test/common/run-tests
```diff
@@ -91,7 +91,6 @@ class Test:
         print_tap = not opts.list
         affected = any([self.command[0].endswith(t) for t in affected_tests])
         retry_reason = ""
-        skip_reason = ""
```
martinpitt, line 94: `skip_reason` is being used again in line 124. This is plausibly the cause of SKIPed tests now being broken?

test/common/run-tests
martinpitt: I [fixed that](https://github.com/cockpit-project/cockpit/pull/18039/commits/52d5b3074cc30e15759f82e59abc084f623dcd71) in #18039. But it still appears as failed, -ish:

```
$ TEST_OS=fedora-36 test/common/run-tests --no-retry-fail --test-dir test/verify TestStorageResize.testResizeLuks
# Result testResizeLuks (__main__.TestStorageResize.testResizeLuks) failed
# 1 TEST FAILED [87s on toolbox]
not ok 1 test/verify/check-storage-resize TestStorageResize.testResizeLuks # SKIP Known issue #1739

# TESTS PASSED [87s on toolbox, 1 parallel tests, 0 serial tests: ]
```
I think it's supposed to be `ok`? Running out of time now, will look tomorrow morning or possibly this evening.

test/common/run-tests
```diff
@@ -183,12 +183,8 @@ class Test:
             return
 
         print()  # Tap needs to start on a separate line
-        if self.returncode == 0:
-            print(f"ok {self}{retry_reason}")
-        elif skip_reason:
-            print(f"ok {self}{skip_reason}")
-        else:
-            print(f"not ok {self}{retry_reason}")
+        status = 'ok' if self.returncode in [0, 77] else 'not ok'
```
martinpitt, line 204: It's this change which breaks the "SKIP" case. The returncode is 1 in the failed case, because finish() calls `test-failure-policy` without adjusting the test's `self.returncode`. I added a fixup.

test/common/run-tests
martinpitt: Both fixed in a force-push.

test/common/testlib.py
martinpitt: Fixed.

test/common/run-tests
martinpitt: ... and applied here in force-push. I made an entry to my todo list to clean up the "method vs. class" approaches for our decorators, this is too wild.

test/common/run-tests
martinpitt: Pushed.

martinpitt (approved):
Assuming green tests, this works fine for me now. I'll also update #18039 with this, so that we have a complete picture of test results.

But I don't want to land this without your cross-review @allisonkarlitskaya , as I messed around with your PR.

allisonkarlitskaya (commented):
Thanks for cleaning this up.  It was a busy afternoon ;)

Looks good to me.

PR #18046, opened 2022-12-15
Various small fixes

martinpitt:
Broken out from #18039, these should be pretty harmless.

### DIFF
--- a/test/verify/check-static-login
+++ b/test/verify/check-static-login
@@ -467,7 +467,7 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
         self.allow_journal_messages('.*type=1400.*avc:  denied .* comm="sudo".*')
         self.allow_journal_messages('.*type=1400.*avc:  denied .* comm="systemd".*')
         self.allow_journal_messages('.*type=1400.*avc:  denied  { watch } .* comm="cockpit-bridge".*')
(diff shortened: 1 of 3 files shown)


jelly (approved):
Danke! Also nice t osee the pybridge catching some bugs!

test/verify/check-static-login
```diff
@@ -467,7 +467,7 @@ account    required     pam_succeed_if.so user ingroup %s""" % m.get_admin_group
         self.allow_journal_messages('.*type=1400.*avc:  denied .* comm="sudo".*')
         self.allow_journal_messages('.*type=1400.*avc:  denied .* comm="systemd".*')
         self.allow_journal_messages('.*type=1400.*avc:  denied  { watch } .* comm="cockpit-bridge".*')
-        self.allow_journal_messages('.*sudo: PERM_SUDOERS: setresuid(.*): Operation not permitted')
+        self.allow_journal_messages('.*sudo:.* setresuid(.*): Operation not permitted')
```
allisonkarlitskaya, line 470: Why is sudo throwing this error to begin with?

martinpitt: This is with a restricted SELinux `user_u` type, which forbids all kinds of things. We shouldn't be too picky here, as that also keeps changing.

PR #18066, opened 2022-12-19
all: eliminate ct-pagesection-mobile class

KKoukiou:
* [x] This needs pixel tests updates as the previous class had forgotten a few pages :disappointed: 

This class removes side padding in mobile view for the page section component. This is a clear patternfly override and should be treated as such, by being placed in the patternfly-overrides file.


martinpitt (approved):
Thank you!

PR #18067, opened 2022-12-19
lib: fix tables in modals in dark mode

### LINKED ISSUE #17976: dark mode: lists in modals have different backgrounds
Lists in modals, specifically used in the storage page, have a white background on white background in light mode. 

image

But in dark mode, the list background and the modal backgrounds both switch to different shades of grey.

image

The solution would be to either make the background transparent for the tables or align it to the same color.

This might actually come from PF. (I'm not sure if they intend on tables to be used like this in modals or not.)

KKoukiou:
* [x] update pixel tests

Fixes #17976


skobyda (approved):
LGTM, thanks

PR #18069, opened 2022-12-20
python: cockpit.Packages D-Bus API

mvollmer:


src/cockpit/packages.py
```diff
@@ -298,3 +307,14 @@ def get_bridges(self):
         for package in sorted(self.packages.values(), key=lambda package: -package.priority):
             bridges.extend(package.bridges)
         return bridges
+
+    @bus.Interface.Method(in_types=[])
+    def reload(self, *args):
```
mvollmer, line 312: I guess `*args` should be removed.

src/cockpit/packages.py
```diff
@@ -298,3 +307,14 @@ def get_bridges(self):
         for package in sorted(self.packages.values(), key=lambda package: -package.priority):
             bridges.extend(package.bridges)
         return bridges
+
+    @bus.Interface.Method(in_types=[])
```
allisonkarlitskaya, line 311: I think you can omit `in_types=`

src/cockpit/packages.py
```diff
@@ -192,11 +193,18 @@ def relative_to(self, path):
...
+
+class Packages(bus.Object, interface='cockpit.Packages'):
+    manifests = bus.Interface.Property('s', value="{}")
 
-class Packages:
     def __init__(self):
+        super().__init__()
         self.packages = {}
         self.load_packages()
+        self.ignore_next_reload_hint = True
```
allisonkarlitskaya, line 207: Why?

src/cockpit/internal_endpoints.py
```diff
@@ -76,10 +76,6 @@ def update(self, *args):
         ...
 
 
-class cockpit_Packages(bus.Object):
-    ...
-
-
```
allisonkarlitskaya, line 82: This makes me happy.  I think as we implement more and more of this functionality "for real" we might notice that this file disappears...

allisonkarlitskaya (commented):
Looks really nice!

Just one question about the "ignore" hint thing.

src/cockpit/packages.py
mvollmer: When the shell loads, it always calls "ReloadHint".  The idea is that we want to reload packages when the user reloads the browser, so that any changes done on the filesystem (maybe by installing more cockpit-foo packages) will show up in the menu.

However, we don't want to reload packages when Cockpit is opened for the first time.

I think the Shell could get smarter and call "Reload", but only when it is actually being reloaded by the user.  I think there are ways to detect that. I didn't know anything about that when implementing package reloading way back when...

And yeah, this needs documenting....

src/cockpit/packages.py
mvollmer: Done.

src/cockpit/packages.py
mvollmer: Documented in the code.

src/cockpit/packages.py
```diff
@@ -192,11 +193,18 @@ def relative_to(self, path):
         assert x.startswith(y)
         return Path(x[len(y):])
 
+# TODO: This doesn't yet implement the slighty complicated checksum
```
allisonkarlitskaya, line 196: This *does* implement the checksum scheme, and even does it in exactly the same way (which was needed to get unit tests passing).  But reloading probably does need more work, indeed.

src/cockpit/packages.py
```diff
@@ -298,3 +307,19 @@ def get_bridges(self):
...
+        self.load_packages()
+
+    @bus.Interface.Method()
+    def reload_hint(self):
+        # Reloading the Shell in the browser should reload the
+        # packages.  This is implemented by having the Shell call this
+        # method whenever it starts.  The first call of this method in
+        # each session is ignored so that packages are not loaded
+        # twice right after logging in.
+        if not self.ignore_next_reload_hint:
```
allisonkarlitskaya, line 323: I think the purpose of this variable would be a bit more clear if it was called `saw_first_reload` or so.  Something that clearly indicates that this is used once at the start and never again.  I'd also move the comment up to the top of the class — throw in a type annotation and comment that.  Or use a docstring?

src/cockpit/packages.py
```diff
@@ -298,3 +307,19 @@ def get_bridges(self):
         for package in sorted(self.packages.values(), key=lambda package: -package.priority):
             bridges.extend(package.bridges)
         return bridges
+
+    @bus.Interface.Method()
+    def reload(self):
+        self.packages = {}
+        self.load_packages()
```
allisonkarlitskaya, line 314: I guess this is the part that will have to get more intelligent in the future if we want to support the checksums properly, right?  We need to be able to detect the case if the scanned package is the same as the package we had before, on a package-by-package basis.

There's code in `load_packages()` that supports this — the meaning of passing the list of checksum objects was that multiple checksums (global and local) could be built.  What's missing is the part to record the individual checksum of each package...

I guess it's a nice chance for a follow-up, anyway.  I question if we want this stuff at all in the future, but at least implementing it shouldn't be too much more work at this point.

allisonkarlitskaya (approved):
```
not ok 2 /static-code/test-flake8
# src/cockpit/packages.py:200:1: E302 expected 2 blank lines, found 1
FAIL test/static-code (exit status: 1)
```

"Request changes" only for the above.  Everything else is stylistic and optional.

allisonkarlitskaya (changes requested):
Wrong button, sorry.

src/cockpit/packages.py
mvollmer: > I guess this is the part that will have to get more intelligent in the future if we want to support the checksums properly, right? We need to be able to detect the case if the scanned package is the same as the package we had before, on a package-by-package basis.

Yes, exactly.

> I guess it's a nice chance for a follow-up, anyway. I question if we want this stuff at all in the future, but at least implementing it shouldn't be too much more work at this point.

Yeah, it's not a lot of work, but it's a complicated scheme and it would simplify things in general if we could get rid of it.

src/cockpit/packages.py
mvollmer: I have renamed the flag and moved the comment to the top.

allisonkarlitskaya:
Both of those fails look unrelated, so retrying, but...

```
not ok 112 test/verify/check-packagekit TestUpdates.testPackageKitCrash [ND@2] # expected failure: still fails with python bridge
```

looks interesting.

allisonkarlitskaya (dismissed):
If the bots are happy, so am I.

mvollmer:
> Both of those fails look unrelated, so retrying, but...
> 
> ```
> not ok 112 test/verify/check-packagekit TestUpdates.testPackageKitCrash [ND@2] # expected failure: still fails with python bridge
> ```
> 
> looks interesting.

Yes, this should pass now.

allisonkarlitskaya (approved):
thanks very much!

PR #18071, opened 2022-12-20
test: make testPmProxySettings work on Arch Linux

jelly:
Arch Linux has no default zone which means the open port for pmproxy functionally does not show (it requires an active zone).


martinpitt (approved):

martinpitt:
. o O { no-test ! }

jelly:
Ignore the flake in fedora-36/pybridge — Tests failed with code 1

PR #18073, opened 2022-12-20
test: Clean up decorator implementation

martinpitt:
Our cockpit test specific decorators were using wildly different
implementations. Make them consistent by always adding a
`_testlib__decoratorname` attribute to the test method or class.

Introduce a `get_decorator()` helper which returns the decorator value on the 
test method or class. This avoids the custom code which iterates over the 
class'es methods and duplicates the value.

This also debundles testlib from the `fmf_metadata` library, which we
have never used so far, and don't really plan to either.

There are still two kinds of decorators now: These with arguments, like @todo,
or even the empty-argumented skipDistroPackage(); and the ones without
arguments, like @nondestructive. This is still a bit confusing, but converting
the latter is more intrusive.

### DIFF
--- a/test/common/run-tests
+++ b/test/common/run-tests
@@ -330,15 +330,14 @@ def detect_tests(test_files, image, opts):
                 # most tests should take much less than 10mins, so default to that;
                 # longer tests can be annotated with @timeout(seconds)
                 # check the test function first, fall back to the class'es timeout
-                test_timeout = getattr(test_method, "__timeout", getattr(test, "__timeout", 600))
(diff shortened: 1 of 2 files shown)


martinpitt:
This was another leftover from #18045, where Lis and I added a sloppy new implementation of `@todo`, and I promised to clean this up afterwards.

martinpitt:
@mvollmer , @allisonkarlitskaya : Happy new year! :partying_face:  Since the review request notification probably got lost over the holidays, this is another one.

mvollmer (approved):

PR #18076, opened 2022-12-22
python: Signal improvements

mvollmer:
- [x] #17750

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -188,8 +188,12 @@ class DBusChannel(Channel):
         def handler(message):
             name, old, new = message.get_body()
             send_owner(owner=new if new != "" else None)
-        rule = f"type='signal',sender='org.freedesktop.DBus',path='/org/freedesktop/DBus',interface='org.freedesktop.DBus',member='NameOwnerChanged',arg0='{self.name}'"
-        self.matches.append(self.bus.add_match(rule, handler))
+        self.add_signal_handler(handler,
+                                sender='org.freedesktop.DBus',
+                                path='/org/freedesktop/DBus',
+                                interface='org.freedesktop.DBus',
+                                member='NameOwnerChanged',
+                                arg0=self.name)
         try:
             unique_name, = await self.bus.call_method_async("org.freedesktop.DBus",
                                                             "/org/freedesktop/DBus",
@@ -254,12 +258,33 @@ class DBusChannel(Channel):
         else:
             self.ready()
 
-    def add_match(self, rule, handler):
+    def add_signal_handler(self, handler, **kwargs):
+        r = dict(**kwargs)
+        r['type'] = 'signal'
+        if 'sender' not in r and self.name is not None:
+            r['sender'] = self.name
+        # HACK - https://github.com/bus1/dbus-broker/issues/309
+        # path_namespace='/' in a rule does not work.
+        if r.get('path_namespace') == "/":
+            del r['path_namespace']
+
+        def filter_owner(message):
+            if self.owner is not None and self.owner == message.get_sender():
+                handler(message)
+
+        if self.name is not None and 'sender' in r and r['sender'] == self.name:
+            func = filter_owner
+        else:
+            func = handler
+        r_string = ','.join(f"{key}='{value}'" for key, value in r.items())
+        self.matches.append(self.bus.add_match(r_string, func))
+
+    def add_async_signal_handler(self, handler, **kwargs):
         def sync_handler(message):
             task = asyncio.create_task(handler(message))
             self.tasks.add(task)
             task.add_done_callback(self.tasks.discard)
-        self.matches.append(self.bus.add_match(rule, sync_handler))
+        self.add_signal_handler(sync_handler, **kwargs)
 
     async def do_call(self, message):
         path, iface, method, args = message['call']
@@ -321,9 +346,7 @@ class DBusChannel(Channel):
                     list(message.get_body())
                 ])
 
-        rule = ','.join(f"{key}='{value}'" for key, value in add_match.items())
-        self.add_match("type='signal'," + rule, match_hit)
-        self.send_message(reply=[], id=message.get('id'))
+        self.add_async_signal_handler(match_hit, **add_match)
 
     async def setup_objectmanager_watch(self, path, interface_name, meta, notify):
         # Watch the objects managed by the ObjectManager at "path".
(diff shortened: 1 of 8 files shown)


src/cockpit/channels/dbus.py
```diff
@@ -205,17 +244,44 @@ def do_open(self, options):
...
+                        self.ready(unique_name=self.owner)
+                    else:
+                        self.close(problem="not-found")
+            task = asyncio.create_task(get_ready())
+            self.tasks.add(task)
+            task.add_done_callback(self.tasks.discard)
+        else:
+            self.ready()
+
+    def add_signal_handler(self, rule, handler):
```
allisonkarlitskaya, line 261: I think I'd prefer to see this as a function which takes the handler as the first argument and the match rule as `**kwargs`.

src/cockpit/channels/dbus.py
```diff
@@ -205,17 +244,44 @@ def do_open(self, options):
...
+        r = rule.copy()
+        r['type'] = 'signal'
```
allisonkarlitskaya, line 263: This is better written `r = dict(rule, type='signal')`

src/cockpit/channels/dbus.py
```diff
@@ -205,17 +244,44 @@ def do_open(self, options):
...
+        if 'sender' not in r and self.name:
```
allisonkarlitskaya, line 264: `... is not None:`

src/cockpit/channels/dbus.py
```diff
@@ -205,17 +244,44 @@ def do_open(self, options):
...
+            r['sender'] = self.name
+        if r.get('path_namespace') == "/":
+            del r['path_namespace']
```
allisonkarlitskaya, line 267: That deserves a comment.  Did you ever get to the bottom of why this is a problem?  Is it a bug in the broker or something?

src/cockpit/channels/dbus.py
```diff
@@ -205,17 +244,44 @@ def do_open(self, options):
...
+        def filter_owner(message):
+            if r['sender'] == self.name:
+                if self.owner and self.owner != message.get_sender():
```
allisonkarlitskaya, line 270: Is this some kind of attempt to protect against races where we receive signals from a new name owner before we're aware that the name has a new owner?  Is that even possible?  We'll have seen the `NameOwnerChanged` signal first...

I'd lean towards thinking that this shouldn't be necessary, but the fact that you've included it sort of indicates that probably it *is* necessary for some reason.  This definitely needs at least a comment.

allisonkarlitskaya (changes requested):
Thanks!  Needs rebasing, then a few questions still to answer.

src/cockpit/channels/dbus.py
mvollmer: After looking at the dbus-broker code, I think it might be a bug.

https://github.com/bus1/dbus-broker/blob/92142f321fc80c76ab1cc973ce2defa1039cac4a/src/bus/match.c#L387

`match_string_prefix("/foo", "/", '/', false)` returns false, no?  I check that more carefully.

mvollmer: https://github.com/bus1/dbus-broker/issues/309

src/cockpit/channels/dbus.py
mvollmer: > Is this some kind of attempt to protect against races

No, this is to filter out signals that come from other senders.  Filtering by a well-known name doesn't seem to work, and we would receive generic signals like PropertiesChanged or InterfacesAdded from the wrong sender.

I think dbus-broker should be able to filter by well-known sender names, but sd-bus probably can not.

allisonkarlitskaya: This is strange.  I also had a memory that filtering by non-unique names didn't work, so I checked it.

The D-Bus spec certainly thinks that you can filter by well-known name.  It includes a couple of examples of doing exactly that.

src/cockpit/channels/dbus.py
```diff
@@ -254,15 +258,34 @@ async def get_ready():
...
+        r = dict(**kwargs)
+        r['type'] = 'signal'
+        if 'sender' not in r and self.name is not None:
+            r['sender'] = self.name
+        # HACK - https://github.com/bus1/dbus-broker/issues/309
+        # path_namespace='/' in a rule does not work.
+        if r.get('path_namespace') == "/":
+            del r['path_namespace']
+        def filter_owner(message):
+            if self.owner is None or self.owner == message.get_sender():
```
allisonkarlitskaya, line 271: is that right?  In case the name has no owner I'd rather think we want to drop the message...

allisonkarlitskaya (changes requested):

src/cockpit/channels/dbus.py
mvollmer: Yes, calling AddMatch on the bus daemon with a well-known name should work.  The daemon knows the well-known name for all unique names when it processes the match rules.

But sd-bus (or glib) also need to process the match rules when a signal comes in, and they only know the unique name of the sender.

Our bridge will call AddMatch multiple times on a single connection.  For example, the Networking page will call bus.add_match for sender=org.freedesktop.NetworkManager and sender=org.freedesktop.systemd (both with interface=org.freedesktop.DBus.Properties).

The handler given to the bus.add_match call for NetworkManager will then see the signals for systemd, because they will both come from the dbus daemon, bu sd-bus can't tell what the well-known name of the signal is.

dbus.py knows, so we can complete the filtering there.

(That's my theory, at least.)

src/cockpit/channels/dbus.py
mvollmer: Yes, I think you are right.  We don't want to filter at all when `self.name is None`, but this code only runs when `self.name is not None` and then we should insist on knowing the owner.

I guess NameOwnerChanged will be received before any signal from the new owner, right?

src/cockpit/channels/dbus.py
allisonkarlitskaya: I wonder if there is a sd-bus bug here.  Would it do a better job if (for example) we used its built-in name-tracking code?  Might be worth a report upstream.

allisonkarlitskaya (changes requested):
The logic looks better now, thanks.  But still failing:

```

not ok 2 /static-code/test-flake8
# src/cockpit/channels/dbus.py:270:9: E306 expected 1 blank line before a nested definition, found 0
# src/cockpit/channels/dbus.py:378:9: E303 too many blank lines (2)
FAIL test/static-code (exit status: 1)
```

src/cockpit/channels/dbus.py
mvollmer: Done.

src/cockpit/channels/dbus.py
mvollmer: Done.

allisonkarlitskaya:
Still tests trouble...

mvollmer:
> Still tests trouble...

A couple of NetworkManager tests now pass. They failed earlier because the empty match used [here](https://github.com/cockpit-project/cockpit/blob/0632543987d85bdb648d578a4ece29911c65a2c8/pkg/networkmanager/interfaces.js#L402) is no longer ignored.

allisonkarlitskaya (approved):
I keep looking at the testing status and thinking "Let's wait for `testing-farm:fedora-37-x86_64 In progress` to finish".  After ~22hours I guess that's not gonna happen, though :)

PR #18077, opened 2022-12-22
cc-table: fix infinite call of useEffect hook

KKoukiou:
Arrays as dependencies in hooks are very problematic and often result in too many uneccessary re-renders. This solution simply serializes the array of row keys into a JSON string object before passing them to the useEffect deps list.

This fixes some issues with the use of the ListingTable component among which the most prominent being the 'ct-new-item' class being added > 100 times in some cases.

This fixes warnings in all pages that use ListingTable component - error: Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.%s

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -78,41 +78,32 @@ export const ListingTable = ({
     sortMethod,
     ...extraProps
 }) => {
-    let rows = tableRows;
-
+    let rows = [...tableRows];
     const [expanded, setExpanded] = useState({});
     const [newItems, setNewItems] = useState([]);
     const [currentRowsKeys, setCurrentRowsKeys] = useState([]);
     const [activeSortIndex, setActiveSortIndex] = useState(sortBy ? sortBy.index : 0);
     const [activeSortDirection, setActiveSortDirection] = useState(sortBy ? sortBy.direction : SortByDirection.asc);
+    const rowKeys = rows.map(row => row.props ? row.props.key : undefined)
+            .filter(key => key !== undefined);
+    const rowKeysStr = JSON.stringify(rowKeys);
+    const currentRowsKeysStr = JSON.stringify(currentRowsKeys);
 
     useEffect(() => {
-        const getRowKeys = rows => {
-            const keys = [];
-
-            rows.forEach(row => {
-                if (row.props && row.props.key !== undefined)
-                    keys.push(row.props.key);
-            });
-
-            return keys;
-        };
-
-        const current_keys = getRowKeys(rows);
-        if (JSON.stringify(current_keys) === JSON.stringify(currentRowsKeys))
-            return;
-
         // Don't highlight all when the list gets loaded
-        if (currentRowsKeys.length !== 0) {
-            const new_keys = current_keys.filter(key => currentRowsKeys.indexOf(key) === -1);
+        const _currentRowsKeys = JSON.parse(currentRowsKeysStr);
+        const _rowKeys = JSON.parse(rowKeysStr);
+
+        if (_currentRowsKeys.length !== 0) {
+            const new_keys = _rowKeys.filter(key => _currentRowsKeys.indexOf(key) === -1);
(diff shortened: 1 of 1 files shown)


pkg/lib/cockpit-components-table.jsx
```diff
@@ -78,41 +78,32 @@ export const ListingTable = ({
...
-
-        const current_keys = getRowKeys(rows);
-        if (JSON.stringify(current_keys) === JSON.stringify(currentRowsKeys))
-            return;
-
         // Don't highlight all when the list gets loaded
-        if (currentRowsKeys.length !== 0) {
-            const new_keys = current_keys.filter(key => currentRowsKeys.indexOf(key) === -1);
+        const _currentRowsKeys = JSON.parse(currentRowsKeysStr);
+        const _rowKeys = JSON.parse(rowKeysStr);
```
skobyda, line 95: I don't get this. Few lines above, we stingify the object, and then here we turn the JSON string back into the same object? Cannot we use the same rowKeys and currentRowsKeys objects?

KKoukiou: that's the point of this PR, if we use the objects, react get's confused and endlessly re-renders causing warnings in the console. 
Check console in any page that contains a table and you will see tons of:
 
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.%s

when adding or removing elements from an array.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -78,41 +78,32 @@ export const ListingTable = ({
...
 }) => {
-    let rows = tableRows;
-
+    let rows = [...tableRows];
     const [expanded, setExpanded] = useState({});
     const [newItems, setNewItems] = useState([]);
     const [currentRowsKeys, setCurrentRowsKeys] = useState([]);
     const [activeSortIndex, setActiveSortIndex] = useState(sortBy ? sortBy.index : 0);
     const [activeSortDirection, setActiveSortDirection] = useState(sortBy ? sortBy.direction : SortByDirection.asc);
+    const rowKeys = rows.map(row => row.props ? row.props.key : undefined)
```
martinpitt, line 87: Hint: Use the null coalescing operator for that: `row.props?.key`

pkg/lib/cockpit-components-table.jsx
martinpitt: I get the stringification part for using as row key, but why the `.parse()` call? I. e. why do you need to create `_rowKeys` and `_currentRowKeys` instead of just using the original values?

skobyda: > that's the point of this PR, if we use the objects, react get's confused and endlessly re-renders causing warnings in the console.

Isn't that fixed in `setCurrentRowsKeys` where you pass strings instead of the whole objects? Passing objects there causes the endless re-renders no?
Therefore I understand how does using `rowKeysStr` and `currentRowsKeysStr` change rerender. 

But how does using `_rowKeys` and `_currentRowKeys` change anything instead of just using `rowKeys` and `currentRowKeys`? Because here you use `_currentRowsKeys` and `_rowKeys` only as part of some if condition few lines below, and do not pass them to a state, so nothing changed the way rerender works, no?

KKoukiou: @martinpitt @skobyda if I would use the rowKeys and currentRowKeys in the useEffect body I would need to add these to in the dependency array of this hook, that cancelling the whole point of this PR, as these variable are complicated objects, and would make the code inside the useEffect hook run without changes happening.

martinpitt: Sorry, I mis-read the patch. The JSON stringification happens in render, outside of the effect. And the parsing happens inside. So this makes sense.

martinpitt (approved):

PR #18082, opened 2022-12-23
networkmanager: fix network interface members actions

### LINKED ISSUE #17318: Broken layout on network interface details
### Explain what happens

1. Visit network interface
2. Add member if there is none
3. Resize the window size so member details starts to wrap. Notice how the minus button become too big and is not spaced with regards to the switch

Screenshot from 2022-05-05 15-38-07

### Version of Cockpit

main

Networking

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

KKoukiou:
Fix actions stretch in medium width and stop using colSpan as it breaks on mobile.

Fixes #17318

### DIFF
--- a/pkg/networkmanager/network-interface-members.jsx
+++ b/pkg/networkmanager/network-interface-members.jsx
@@ -105,7 +105,7 @@ export const NetworkInterfaceMembers = ({
                     // Will add traffic info right after
                     {
                         title: (
-                            <>
+                            <div className="btn-group">
                                 {onoff}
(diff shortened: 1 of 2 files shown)


KKoukiou:
After:
Screen Shot 2022-12-23 at 15 57 30
Screen Shot 2022-12-23 at 15 57 21

Before:
Screen Shot 2022-12-23 at 15 54 48
Screen Shot 2022-12-23 at 15 59 27

martinpitt (approved):
Danke!

PR #18087, opened 2022-12-26
Fix 'edit' mode for hosts

### LINKED ISSUE #17479: Editing the username of an existing host is not possible
### Explain what happens

1. Add a host
2. Edit the host and change the username

image

### Version of Cockpit

271

_No response_

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

KKoukiou:
When editing hosts, the dialog should reflect the 'edit' mode with the title and primary buttons correctly.

Also it should be allowed to just change the username.

Fixes #17479


martinpitt (approved):
Epharisto!

PR #18102, opened 2022-12-30
services: try out lazy loading of components that are not rendered in the default load

### LINKED ISSUE #17643: React code splitting
To improve our time-to-load experience (performance), we should consider code splitting. React + WebPack can do this.

React docs on code splitting: https://reactjs.org/docs/code-splitting.html

Using "Suspense" on web.dev: https://web.dev/code-splitting-suspense/

We might want to pick a slower-loading page to start and see what it's like. Or we could pick one of the simpler pages to experiment.

KKoukiou:
This reduces the bundled size of the services.js from ~5MB to ~4.5MB. By analyzing the time needed for my browser to load this file when looking at the 'Network' tab in developers tools, I noticed that it also decreased 500ms.

Do the same for the 'Shutdown' component code in overview page.

Fixes #17643

### DIFF
--- a/pkg/lib/cockpit-components-shutdown.jsx
+++ b/pkg/lib/cockpit-components-shutdown.jsx
@@ -237,3 +237,5 @@ export class ShutdownModal extends React.Component {
         );
     }
 }
+
+export default ShutdownModal;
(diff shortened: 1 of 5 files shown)


KKoukiou:
Without suspense:
services-without-suspense

With suspense:
Screenshot from 2022-12-30 13-23-11

KKoukiou:
Oh this crashes the lazy pages.

KKoukiou:
Can't make it work, it crashes in production with missing file chunks. Found tons of issues reported in different sites, but non of the sugested work arounds seemed to help. Not taking more time now, as this scale of improvement does not worth the effort.

PR #18108, opened 2023-01-02
test: Accept insights_client_t SELinux policy

martinpitt:
on RHEL, insights-client started to modify the default policy to make
itself permissive; accept that. We can't reset it with
`semanage permissive -D` as that OOMs.

---

This blocks the [rhel-8-8](https://github.com/cockpit-project/bots/pull/4178) and [rhel-9-2](https://github.com/cockpit-project/bots/pull/4226) image refreshes.


martinpitt:
Meh, what? This doesn't work, it gets killed as it runs out of memory. Let's do this differently.

marusak (approved):
Danke :)

PR #18109, opened 2023-01-02
Edit http to https in w3.org links

marusak:
See https://github.com/cockpit-project/cockpit/issues/17679


martinpitt:
I thought about that, but this isn't right -- that http:// is API. Instead, we need to teach the URL checker to ignore these xmlns= attributes.

martinpitt (changes requested):
See above, this is wrong.

martinpitt:
Closing, this isn't the right approach.

PR #18111, opened 2023-01-02
tests: when running checking failure policy rename f-rawhide to f-38

KKoukiou:


test/common/run-tests
```diff
@@ -131,7 +131,7 @@ class Test:
             cmd = ["test-failure-policy"]
             if not opts.track_naughties:
                 cmd.append("--offline")
-            cmd.append(testvm.DEFAULT_IMAGE)
+            cmd.append("fedora-38" if testvm.DEFAULT_IMAGE == "fedora-rawhide" else testvm.DEFAULT_IMAGE)
```
martinpitt, line 134: I find this really ugly here. Could we do this either in test-failure-policy, or in bots lib/testmap.py somewhere?

Actually, how about we add a naughty/fedora-rawhide → fedora-38 symlink? That's already a common pattern anyway. Sorry for not thinking of this before!

martinpitt (changes requested):

test/common/run-tests
KKoukiou: Indeed good idea

PR #18112, opened 2023-01-03
systemd: Use sysusers.d to create ws users

### LINKED ISSUE #15027: ws: please provide sysuser.d entries for system users/groups
It looks like most socket and service units for `cockpit-ws` support running under custom system users/groups (e.g. `@user@` / `@wsinstanceuser@`, and related groups) from [autoconf](https://github.com/cockpit-project/cockpit/blob/233.1/configure.ac#L409-L463).

It would be nice to provide separate [sysusers.d](https://www.freedesktop.org/software/systemd/man/sysusers.d.html) files with relevant entries for each of those, so that downstream distributions could automatically pick up the system user/group creation in a uniform way (see https://fedoraproject.org/wiki/Changes/Adopting_sysusers.d_format for an example from Fedora).

martinpitt:
Move to using systemd's sysusers declarative files [1] for creating our 
system users/groups. Arch already does that, Fedora moved to it since
Fedora 32 [2], and Debian supports it as well [3].

In debian/cockpit-ws.postinst, move the `#DEBHELPER#` block above the 
statoverride, as the former now generates the user, and the latter needs
it.

Unfortunately Fedora/rpm's `%attr` does not really work with sysusers
files shipped upstream yet. The conf files are not installed yet during
`%pre`, but creating the users in `%post` is too late for the file
unpack phase, so cockpit-session would get the wrong permissions. Thus
duplicate the two sysusers config lines verbatim in `%pre`, which is at
least marginally better than calling `useradd` etc. programmatically.

Extend TestConnection.testWsPackage to remove the system users, reboot,
and validate that cockpit still works. This ensures correct sysusers.d
packaging across all distributions, as our normal CI images already have
the system users.

Fixes #15027

[1] https://www.freedesktop.org/software/systemd/man/sysusers.d.html
[2] https://fedoraproject.org/wiki/Changes/Adopting_sysusers.d_format
[3] https://manpages.debian.org/dh_installsysusers

-----

 - [ ] Rework this after PR #16808

We originally tried `DynamicUser=` in #16811, but that has been blocked for too long. IMHO this is a nice improvement already, and does not block moving to DynamicUser= in the future.

### DIFF
--- a/src/systemd/Makefile.am
+++ b/src/systemd/Makefile.am
@@ -1,5 +1,5 @@
 # -----------------------------------------------------------------------------
-# all systemd units, tmpfiles, and related helpers
+# all systemd units, users, tmpfiles, and related helpers
 
 nodist_systemdunit_DATA = \
 	src/systemd/cockpit-motd.service \
@@ -31,6 +31,9 @@ install-exec-hook::
 tempconfdir = $(prefix)/lib/tmpfiles.d
 nodist_tempconf_DATA = src/systemd/cockpit-tempfiles.conf
 
+sysusersdir = $(prefix)/lib/sysusers.d
+nodist_sysusers_DATA = src/systemd/sysusers/cockpit.conf
+
 # we can't generate these with config.status because,
 # eg. it does "@libexecdir@" -> "${exec_prefix}/libexec"
 src/systemd/%: src/systemd/%.in
@@ -48,6 +51,7 @@ src/systemd/%: src/systemd/%.in
 systemdgenerated = \
 	$(nodist_systemdunit_DATA) \
 	$(nodist_tempconf_DATA) \
+	$(nodist_sysusers_DATA) \
 	$(NULL)
 systemdgenerated_in = $(patsubst %,%.in,$(systemdgenerated))
 
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,7 @@ Makefile.in
 /src/systemd/cockpit*.service
 /src/systemd/cockpit*.socket
 /src/systemd/cockpit-tempfiles.conf
+/src/systemd/sysusers/*.conf
 /src/tls/cockpit-certificate-helper
 /src/ws/cockpit-desktop
 /src/ws/cockpit.appdata.xml
--- a/src/systemd/sysusers/cockpit.conf.in
+++ b/src/systemd/sysusers/cockpit.conf.in
@@ -1,0 +1,2 @@
+u @user@ - "User for cockpit web service"
+u @wsinstanceuser@ - "User for cockpit-ws instances"
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -674,6 +674,16 @@ class TestConnection(MachineCase):
         if root_login_disallowed:
             self.assertIn('root', m.execute("cat /etc/cockpit/disallowed-users"))
 
+        # system users are regenerated at boot
+        m.execute("systemctl stop cockpit cockpit.socket")
+        m.execute("userdel cockpit-ws; userdel cockpit-wsinstance; ! grep cockpit /etc/passwd /etc/group")
+        # HACK: ConditionNeedsUpdate=/etc does not notice changed /etc/passwd file
+        m.execute("rm -f /etc/.updated")
+        m.reboot()
+        m.execute("systemctl start cockpit.socket")
+        headers = m.curl("--head", "--fail", "http://127.0.0.1:9090/cockpit/static/login.html")
+        self.assertIn("HTTP/1.1 200 OK\r\n", headers)
+
     @skipImage("OSTree doesn't have cockpit-ws", "fedora-coreos")
     @nondestructive
     def testCommandline(self):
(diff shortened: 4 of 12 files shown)


martinpitt:
The [arch failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18112-20230103-084117-21f21f5f-arch/log.html) is straightforward, already fixed/verified locally.

The two [coreos failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18112-20230103-084117-21f21f5f-fedora-coreos/log.html#317) are more serious: "error: While applying overrides for pkg cockpit-ws: Could not find group 'cockpit-wsinstance' in group file". It seems `rpm-ostree install` does not properly process the sysusers.d rpm script.

I moved it from `%post` to `%pre`, and that works. It's the right thing to do anyway, `%post` was an oversight. Yay tests!

This should also fix [all the TF failures](https://artifacts.dev.testing-farm.io/ace783e4-0451-497a-baee-e2b5395a821a/) which use a fresh install instead of an upgrade. Nice to cover that case cleanly as well!

martinpitt:
[TF failures](https://artifacts.dev.testing-farm.io/f0c93994-a8d2-45e6-993d-7da00b0bbb7e/) are still significant:
```
  Running scriptlet: cockpit-ws-282.1-1.20230103102400455339.pr18112.15   87/95 
Failed to open 'cockpit.conf', ignoring: No such file or directory
  Installing       : cockpit-ws-282.1-1.20230103102400455339.pr18112.15   87/95 
warning: group cockpit-wsinstance does not exist - using root
```

[This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18112-20230103-101829-677a9061-centos-8-stream/log.html#33-2) is an independent flake, I'll add a piggy-back commit here.

src/systemd/Makefile.am
```diff
@@ -48,6 +51,7 @@ src/systemd/%: src/systemd/%.in
 systemdgenerated = \
 	$(nodist_systemdunit_DATA) \
 	$(nodist_tempconf_DATA) \
+	$(nodist_sysusers_DATA) \
```
allisonkarlitskaya, line 54: I'd lean towards doing that in `configure.ac`.  The only reason we need to do these ones here is to get the recursive path expansion (see the comment just above), but that's not an issue for these files.

martinpitt:
@allisonkarlitskaya : Yes, I'm aware of this, I mentioned it in the description already. But I don't think this will make progress anytime soon.

allisonkarlitskaya (dismissed):
This looks good assuming everything passes.  One comment for your consideration.

allisonkarlitskaya:
> @allisonkarlitskaya : Yes, I'm aware of this, I mentioned it in the description already. But I don't think this will make progress anytime soon.

I totally agree.  We're basically waiting on support in systemd which nobody seems to be interested in working on.  I have no intention to block this PR in the meantime.

src/systemd/Makefile.am
martinpitt: I was actually vaguely aware of this, but I found it nice to keep all the systemd bits together in one file. I don't have a strong opinion about it, though.

martinpitt:
Hmm, the two IPA tests fail a little too often for my taste on [R8.8](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18112-20230103-170652-3cdade7b-rhel-8-8/log.html) and [C8S](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18112-20230103-170656-3cdade7b-centos-8-stream/log.html#312). A lot of it is noise due to the "3x affected" retries, but apparently not these. They reproduce well locally, I'll investigate them tomorrow.

allisonkarlitskaya:
could this be caused by a conflict with the existing users that we get from installing the package in the base image?  in a certain sense we're not really properly testing the creation of new users here...

martinpitt:
@allisonkarlitskaya : We test three cases now: 
 - Upgrades from the old `useradd` generated user, where the sysusers.d config should be a no-op (most tests in our CI). That's causing the subtle breakage of IPA certificate generation, which must be some weird side effect. I'm going to investigate it this morning.
 - Fresh package installs with sysusers generated users from RPM `%pre` script (Testing Farm)
 - sysusers generated users during boot after resetting /etc (`TestConnection.testWsPackage`)

martinpitt:
LOL :rofl: The reason was that the macro is broken in RHEL 8 (see edited comment above), and thus the remainder of `%pre` was skipped. As a result, it was using the selinux-policy cockpit policy instead of our own. Fixed!

allisonkarlitskaya:
I think the trouble we're seeing here is maybe a fundamental incompatibility between sysusers and having non-root-owned files installed by the package.

If the user gets deleted and later recreated, the uid on the filesystem might not be the same as the uid of the new user...

martinpitt:
Ack, indeed. So let's shelve this and reconsider after #16808 -- if that ever lands, that would unblock the sysusers approach. Thanks for catching!

martinpitt:
This is still very far out, and I don't want to stare at this on my /pulls page for that long. I sent the Debian packaging fix to #18339, and will close this one for now.

PR #18114, opened 2023-01-03
ssh: Increase GSource poll timeout

### LINKED ISSUE #17466: cockpit-ssh sits in a tight loop all the time
### Explain what happens

1. Log into remote machine via Cockpit
2. Look at cockpit-ssh process on that machine in "top" or similar, it's about 5% for me

### Version of Cockpit

269

Unknown or not applicable

### Server operating system

Fedora

36

Firefox

### System log

_No response_

martinpitt:
We don't want poll() to wake up every 1ms, as that causes cockpit-ssh to
burn a noticeable amount of CPU. We still need to handle the occasional
`status == SSH_READ_PENDING`, so we can't wait indefinitely; but at
least increase the timeout to 100ms.

Fixes #17466

### DIFF
--- a/src/ssh/cockpitsshrelay.c
+++ b/src/ssh/cockpitsshrelay.c
@@ -2065,7 +2065,7 @@ cockpit_ssh_source_prepare (GSource *source,
   CockpitSshRelay *self = cs->relay;
   gint status;
 
-  *timeout = 1;
+  *timeout = 100;
 
   status = ssh_get_status (self->session);
 


mvollmer (dismissed):
Looks good to me, but I wonder if there is anything in libssh that requires polling...

allisonkarlitskaya (changes requested):
I think this isn't good enough.  This will deadlock in the case that `status == SSH_READ_PENDING`, for example.  You need to check *at least* that case, and maybe others.

You also need to be quite sure that there is no way for the status to change without the fd being reported readable.  Does the client try to send keepalives, for example?

martinpitt:
@allisonkarlitskaya : Admittedly I don't really understand what you are saying (I didn't dive into the code very deeply). I didn't notice anything wrong at least when testing interactively, and tests are happy too. But we could also change it to only poll every 10 or 100 ms, which would noticeably lower the overhead. Would that work for you?

allisonkarlitskaya:
This is a tricky question.  Assuming that this start to fail for somebody at some point, a smaller value would make the failure less annoying, but a bigger value would make the failure easier to track down.

If you tested it and say it seems to work in the usual case, I'd maybe err towards a larger value (or indeed -1).  Someone should really check libssh to find out when various conditions get set, though, and in response to what events.

martinpitt:
My suspicion is that I didn't notice a problem because there are plenty of events (i.e. reads from ssh) coming in anyway, as our pages and the bridge is very chatty in general. I bumped it to 100ms to be on the safe side; that's Good Enough™, and at some point I feel like this is going to be replaced with a Python wrapper anyway :wink:

allisonkarlitskaya (changes requested):
Seriously, though: if `SSH_READ_PENDING` is set then this needs to return immediately ready.

It probably makes sense to do that from the `_check()` function, since we call that at the end of `_prepare()`.

martinpitt:
OK, nevermind then -- I don't really feel like deep-diving into this code, especially not when it's going to be scrapped soon.

PR #18116, opened 2023-01-03
test: add pixel tests for accounts page

KKoukiou:


marusak:
Can you please also add pixel test for opened groups view?

KKoukiou:
> Can you please also add pixel test for opened groups view?

I think this will be very problematic. There are 100s of groups available, and i expect these to change with operating system changes. I thought I could include a single row instead, of a known group. What do you think?

PR #18120, opened 2023-01-04
test: Warn when static-code skips a test

mvollmer:
Without "--tap", test/static-code would be silent about skipping tests.  This might cause people not to notice that they are not running the full suite.


martinpitt (approved):
Danke!

PR #18123, opened 2023-01-04
test: Use 'python bridge' when skipping tests

marusak:
We want to be able to quickly know how many tests are skipped due to python bridge so this makes it easy to just search for 'python bridge' in test results.

For example, if I search for `python bridge` in [this log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18121-20230104-110815-da847583-fedora-36-pybridge/log.html) it shows 81 skips. There are 4 occurrences of `pybridge` but one of them is name of the scenario. And separately there is this "wedges hard" message. so all together 85 skips. There are 107 skipped tests, so 22 are skipped independently from python bridge. With this change I can just search for `python bridge` and see 85 straight away.

### DIFF
--- a/test/verify/check-connection
+++ b/test/verify/check-connection
@@ -747,7 +747,7 @@ class TestConnection(MachineCase):
         self.assertIn("HTTP/1.1 404 Not Found\r\n", headers)
 
     @skipImage("ssh root login not allowed", "fedora-coreos")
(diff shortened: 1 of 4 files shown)


martinpitt (approved):
Thanks! A real manager PR :grin:

PR #18126, opened 2023-01-05
tests: pixel-testing: introduce a scenario for RTL

KKoukiou:
This commit introduces a new pixel test scenario for RTL (Right to Left). This scenario keeps using English for language but changes the direction of the language to be RTL.

The reasoning behind this choice is that pixel tests take screenshots in the middle of tests execution, by dynamically adopting the theme, browser size etc for the different scenarios.
However in order to change the UI language one needs to reload the page, which effectively means losing the application state. For this reason, we would need to set the language at the very beginning of each test if we wanted to properly test RTL with a language like Hebrew.

Since this means making a whole test run in another language, we decided that we will for the time being test RTL with English. This is obviously an unrealistic scenario but allows us to view spacing issues at least.

### DIFF
--- a/pkg/shell/nav.scss
+++ b/pkg/shell/nav.scss
@@ -166,6 +166,13 @@ $desktop: $phone + 1px;
   }
 }
 
+[dir="rtl"] {
+  // Not setting this breaks pixel tests
+  .screenreader-text {
+    right: -999px;
+  }
+}
+
 /* Desktop navigation */
 @media (min-width: $desktop) {
   // Adapt PF4's new header select for 2-up
--- a/test/common/test-functions.js
+++ b/test/common/test-functions.js
@@ -51,6 +51,14 @@ function ph_set_val (sel, val) {
     el.dispatchEvent(ev);
 }
 
+function ph_set_text_direction(sel, direction) {
+    const el = ph_find(sel)
+
+    if (el.value === undefined)
+        throw new Error(sel + " does not have a value");
+    el.dir = direction;
+}
+
 function ph_has_val (sel, val) {
     return ph_val(sel) == val;
 }
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -146,7 +146,13 @@ default_layouts = [
         "theme": "dark",
         "shell_size": [1920, 1200],
         "content_size": [1680, 1130]
-    }
+    },
+    {
+        "name": "rtl",
+        "theme": "light",
+        "shell_size": [1920, 1200],
+        "content_size": [1680, 1130]
+    },
 ]
 
 
@@ -227,6 +233,7 @@ class Browser:
             self._set_window_size(size[0], size[1])
         if cookie:
             self.cdp.invoke("Network.setCookie", **cookie)
+
         self.switch_to_top()
         self.cdp.invoke("Page.navigate", url=href)
 
@@ -971,6 +978,14 @@ class Browser:
         if self.cdp.browser.name == "chromium":
             self.cdp.invoke("Emulation.setEmulatedMedia", features=[{'name': 'prefers-color-scheme', 'value': name}])
 
+    def _set_direction(self, direction: str):
+        cur_frame = self.cdp.cur_frame
+        if self.is_present("#shell-page"):
+            self.switch_to_top()
+            self.set_attr("#shell-page", "dir", direction)
+        self.switch_to_frame(cur_frame)
+        self.set_attr("html", "dir", direction)
+
     def set_layout(self, name: str):
         layout = [lo for lo in self.layouts if lo["name"] == name][0]
         if layout != self.current_layout:
@@ -1155,10 +1170,15 @@ class Browser:
         for layout in self.layouts:
             if layout["name"] not in skip_layouts:
                 self.set_layout(layout["name"])
+                if "rtl" in self.current_layout["name"]:
+                    self._set_direction("rtl")
                 self.assert_pixels_in_current_layout(selector, key, ignore=ignore,
                                                      scroll_into_view=scroll_into_view,
                                                      wait_animations=wait_animations,
                                                      wait_delay=wait_delay)
+
+                if "rtl" in self.current_layout["name"]:
+                    self._set_direction("ltr")
         self.set_layout(previous_layout)
 
     def assert_no_unused_pixel_test_references(self):
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -409,7 +409,7 @@ class Browser:
         :param val: the value of the attribute
         """
         self._wait_present(selector + ':not([disabled]):not([aria-disabled=true])')
-        self.call_js_func('ph_set_attr', selector, attr, val and 'true' or 'false')
+        self.call_js_func('ph_set_attr', selector, attr, val)
 
     def get_checked(self, selector: str):
         """Get checked state of a given selector.
--- a/test/verify/check-networkmanager-basic
+++ b/test/verify/check-networkmanager-basic
@@ -49,6 +49,7 @@ class TestNetworkingBasic(NetworkCase):
         b.wait_in_text("#networking-graphs", "Transmitting")
         b.wait_in_text("#networking-graphs", "Receiving")
 
+        # FIXME: rtl test was flaky, debug it and remove the skip
         b.assert_pixels(
             "#networking", "network-main",
             [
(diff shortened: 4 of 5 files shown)


KKoukiou (commented):
@martinpitt in case you 're wondering why I did not just sed the po.js files to change the language-direction key there.
There are parsed on page load. But I need a mechanism to dynamically change that during the test execution. :/ 

This solution is uglier but serves this purpose.

marusak:
Login screen does not look rtl: https://github.com/cockpit-project/pixel-test-reference/compare/2ab940de7ba68a553e5244e634c1db77eff62138..676436b49df61130c96f0257a3767ea6388ecaad#diff-a4866e60b946fb584e7f5b28e90b69c0e14d18ed0e0928fc18c53eff3e690deb

Nor does anything in shell https://github.com/cockpit-project/pixel-test-reference/compare/2ab940de7ba68a553e5244e634c1db77eff62138..676436b49df61130c96f0257a3767ea6388ecaad#diff-3255874f79f71d95ef658599d3a2ef61773b263f44327904a8417c81b1d29638

I guess this is due to you setting rtl to iframes, but login page and shell are not iframes

KKoukiou:
https://cockpit-logs.us-east-1.linodeobjects.com/pull-18126-20230106-125306-c34f483e-fedora-36/log.html

KKoukiou:
There is still some flakiness with the groups RTL tests.

test/common/testlib.py
```diff
@@ -1006,6 +1032,7 @@ def assert_pixels_in_current_layout(self, selector: str, key: str,
         if not (Image and self.pixels_label):
             return
 
+        self._set_rtl("rtl" in self.current_layout["name"])
```
martinpitt, line 1035: I don't like that we do this for *every* scenario. It's expensive to do, but moreover it changes the pages, and thus we don't validate the default layout any more. I. e. after this we could accidentally enable RTL for English and still get passing tests.

Could this move into assert_pixels(), and only switch once it reaches the "rtl" scenario? Then of course the layout list needs a comment to say that all RTL scenarios must come at the end, and assert_pixels() could even grow an assertion for this (although it should be obvious enough if we screw this up, as all screenshots would yell at us). After that, set it back to LTR.

pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -342,6 +342,11 @@ svg {
     }
   }
 
+  // Not setting this breaks pixel tests
+  .screenreader-text {
+    right: -999px;
+  }
```
martinpitt, line 348: That feels like a weird place. All the other attributes are set in pkg/shell/nav.scss, including `left: -999px`. Shouldn't that rather move there? This would keep all settings together, and this isn't a PatternFly class either.

martinpitt (changes requested):
Thanks for setting this up! Indeed this exposes some very broken layouts like [this Services page unit state](https://raw.githubusercontent.com/cockpit-project/pixel-test-reference/6ffe4e1ee69af4c1cca8455a2263bf3e5b035292/TestServices-testBasic-details-test-fail-rtl-pixels.png), so that's useful to validate fixes (and find out about in the first place).

I have some minor comments about implementation details, but I agree that this is a sensible first test strategy. We can fix the obvious broken layouts ourselves, but after that we need someone who actually uses an RTL language to review the finer issues.

Cheers!

pkg/lib/patternfly/patternfly-4-overrides.scss
KKoukiou: Indeed. I thought that all dark theme overrides are here, but it's my mistake. Will move.

test/common/testlib.py
KKoukiou: > I don't like that we do this for _every_ scenario. It's expensive to do, but moreover it changes the pages, and thus we don't validate the default layout any more. I. e. after this we could accidentally enable RTL for English and still get passing tests.
> 
> Could this move into assert_pixels(), and only switch once it reaches the "rtl" scenario? Then of course the layout list needs a comment to say that all RTL scenarios must come at the end, and assert_pixels() could even grow an assertion for this (although it should be obvious enough if we screw this up, as all screenshots would yell at us). After that, set it back to LTR.

I don't see how this will work. Even if we have  the RTL scenario in the end of the list, we want to undo the RTL setting after the assert pixels for the RTL scenario. This is because test continues to run after the assert_pixels call , and there might be a followup assert_pixels in the same test or the test might fail, which generates a screenshot, and we want that to be the original screen, not RT.

So I adjust the code to call set_rtl only on the RTL scenario, but we definitely need to revert this after this scenario.

test/common/testlib.py
```diff
@@ -971,6 +978,25 @@ def _set_emulated_media_theme(self, name: str):
         if self.cdp.browser.name == "chromium":
             self.cdp.invoke("Emulation.setEmulatedMedia", features=[{'name': 'prefers-color-scheme', 'value': name}])
 
+    def _set_rtl(self, is_rtl: bool):
```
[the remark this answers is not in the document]
KKoukiou: test/common/test-functions do not have any cdp.invode calls however. Do you suggest that I move the code as is there or create smaller helper functions to be moved to test-function.js?

test/common/testlib.py
martinpitt: Yes, of course -- that's what I meant with "After that, set it back to LTR." There's already some reset code after the loop which iterates the scenarios. It's still DOM hacking (which is super ugly), but I don't think we can avoid it with our current language switching. This whole "RTL scenario" is a big hammer..

test/common/testlib.py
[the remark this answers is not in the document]
martinpitt: I mean to create a `ph_set_text_direction()` helper in test-functions.js, and call it through `b.eval_js()`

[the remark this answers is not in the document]
KKoukiou: indeed it makes more sense, thanks for the suggestion!

pkg/shell/nav.scss
```diff
@@ -166,6 +166,13 @@ $desktop: $phone + 1px;
   }
```
martinpitt, line 166: Thanks for the excellent commit message with documenting our approach!

test/common/test-functions.js
```diff
@@ -51,6 +51,14 @@ function ph_set_val (sel, val) {
     el.dispatchEvent(ev);
 }
 
+function ph_set_text_direction(sel, direction) {
```
martinpitt, line 54: I was going to complain that this function actually doesn't do enough to justify its existence. Sorry for the suggestion then, it looked a lot more involved originally  (with DOM searching and multiple commands). But indeed you actually don't call this anywhere, so this is just dead weight. Please remove -- this is fine in a follow-up if you want to land this quickly (due to blocking pixel tests).

test/common/testlib.py
```diff
@@ -971,6 +978,14 @@ def _set_emulated_media_theme(self, name: str):
         if self.cdp.browser.name == "chromium":
             self.cdp.invoke("Emulation.setEmulatedMedia", features=[{'name': 'prefers-color-scheme', 'value': name}])
 
+    def _set_direction(self, direction: str):
+        cur_frame = self.cdp.cur_frame
+        if self.is_present("#shell-page"):
+            self.switch_to_top()
+            self.set_attr("#shell-page", "dir", direction)
+        self.switch_to_frame(cur_frame)
+        self.set_attr("html", "dir", direction)
```
martinpitt, line 987: Is this safe to do when there is no current frame, i.e. the test currently looks at the shell? But anyway, tests are green, and we do have at least the [TestKeys](https://github.com/cockpit-project/pixel-test-reference/compare/2ab940de7ba68a553e5244e634c1db77eff62138..c4c71f2278662fe7c82d76c409f2fa3890c6501c#diff-600b1434b650c24d5e3fafd69e013d648f5dd87d032ac56b1e42905d832be72f) screenshot, which is shell.

test/common/testlib.py
```diff
@@ -1155,10 +1170,15 @@ def assert_pixels(self, selector: str, key: str,
...
                 self.set_layout(layout["name"])
+                if "rtl" in self.current_layout["name"]:
+                    self._set_direction("rtl")
                 self.assert_pixels_in_current_layout(selector, key, ignore=ignore,
                                                      scroll_into_view=scroll_into_view,
                                                      wait_animations=wait_animations,
                                                      wait_delay=wait_delay)
+
+                if "rtl" in self.current_layout["name"]:
+                    self._set_direction("ltr")
```
martinpitt, line 1181: We should do this only once at the end, but this only becomes relevant once we introduce a second RTL scenario. We can do this later.

martinpitt (approved):
Thanks!

PR #18127, opened 2023-01-05
bridge: AMD CPU temperature detection

jelly:
For AMD CPU's a Tctl and Tccd1 sensor exists, we want to show the values from Tccd1. HWmon devices do not show up how we expect them, we expect them to show up as 1,2,3 but for k10temp it reports as 1,3 so we never showed the value of Tccd1 as it's temp3_label for me.

I'd love to add a test for the Python bridge but unsure how easy it is. I also dislike the solution a lot, maybe we can count the times we can't open a sensor and if it's more then ~ 5 we break the loop?

### DIFF
--- a/src/bridge/cockpitcpusamples.c
+++ b/src/bridge/cockpitcpusamples.c
@@ -153,24 +153,31 @@ sample_cpu_sensors (gchar *sensor_path,
 
 static void
 detect_cpu_sensors (GList **devices,
-                    const gchar *name,
+                    const gchar *hwmon_name,
                     int hwmonID)
 {
-  for (int i = 1; TRUE; i++)
-    {
-      g_assert (devices != NULL);
-      // break loop if input file does not exist
-      g_autofree gchar *sensor_path = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_input", hwmonID, i);
-      if (!g_file_test (sensor_path, G_FILE_TEST_EXISTS))
-          break;
+  g_autofree gchar *path = g_strdup_printf ("/sys/class/hwmon/hwmon%d", hwmonID);
+  g_autoptr(GDir) dir = g_dir_open (path, 0, NULL);
+  if (dir == NULL)
+      return;
 
+
+  const char *name;
+  while ((name = g_dir_read_name (dir)))
+    {
+      uint i;
+      if (!g_str_has_suffix(name, "_input"))
+          continue;
+
+      g_autofree char *sensor_path = g_build_filename (path, name, NULL);
+      sscanf(name, "temp%d_input", &i);
       g_autofree gchar *label = g_strdup_printf ("/sys/class/hwmon/hwmon%d/temp%d_label", hwmonID, i);
       g_autofree gchar *label_content = read_file (label);
 
       if (label_content == NULL)
         {
           // labels aren't used on ARM
-          if (!g_str_equal (name, "cpu_thermal"))
+          if (!g_str_equal (hwmon_name, "cpu_thermal"))
               continue;
         }
       else
@@ -178,7 +185,7 @@ detect_cpu_sensors (GList **devices,
           g_strchomp (label_content);
 
           // only sample CPU Temperature in atk0110
-          if (!g_str_equal (label_content, "CPU Temperature") && g_str_equal (name, "atk0110"))
+          if (!g_str_equal (label_content, "CPU Temperature") && g_str_equal (hwmon_name, "atk0110"))
               continue;
           // ignore Tctl on AMD devices
           if (g_str_equal (label_content, "Tctl"))
--- a/src/cockpit/samples.py
+++ b/src/cockpit/samples.py
@@ -17,12 +17,14 @@
 
 import collections
 import os
+from systemd_ctypes import Handle
 
-from typing import Dict, List, NamedTuple, Optional, Union
+from typing import Dict, Iterable, List, NamedTuple, Optional, Union
 
 
 USER_HZ = os.sysconf(os.sysconf_names['SC_CLK_TCK'])
 MS_PER_JIFFY = 1000 / (USER_HZ if (USER_HZ > 0) else 100)
+HWMON_PATH = '/sys/class/hwmon'
 
 Samples = collections.defaultdict[str, Union[float, Dict[str, Union[float, None]]]]
 
@@ -95,47 +97,67 @@ class MemorySampler(Sampler):
 
 class CPUTemperatureSampler(Sampler):
     # Cache found sensors, as they can't be hotplugged.
-    sensors: List[str] = []
+    sensors: Optional[List[str]] = None
 
     descriptions = [
         SampleDescription('cpu.temperature', 'celsius', 'instant', True),
     ]
 
-    def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
-        for index in range(1, 2 ** 32):
-            sensor_path = f'/sys/class/hwmon/hwmon{hwmonid}/temp{index}_input'
-            if not os.path.exists(sensor_path):
-                break
+    @staticmethod
+    def detect_cpu_sensors(dir_fd: int) -> Iterable[str]:
+        # Read the name file to decide what to do with this directory
+        try:
+            with Handle.open('name', os.O_RDONLY, dir_fd=dir_fd) as fd:
+                name = os.read(fd, 1024).decode().strip()
+        except FileNotFoundError:
+            return
 
-            with open(f'/sys/class/hwmon/hwmon{hwmonid}/temp{index}_label') as f:
-                label = f.read().strip()
-            if label:
-                # only sample CPU Temperature in atk0110
-                if label != 'CPU Temperature' and name == 'atk0110':
-                    continue
-                # ignore Tctl on AMD devices
-                if label == 'Tctl':
-                    continue
-            else:
-                # labels are not used on ARM
-                if name != 'cpu_thermal':
+        if name == 'atk0110':
+            # only sample 'CPU Temperature' in atk0110
+            predicate = (lambda label: label == 'CPU Temperature')
+        elif name == 'cpu_thermal':
+            # labels are not used on ARM
+            predicate = None
+        elif name == 'coretemp':
+            # accept all labels on Intel
+            predicate = None
+        elif name in ['k8temp', 'k10temp']:
+            # ignore Tctl on AMD devices
+            predicate = (lambda label: label != 'Tctl')
+        else:
+            # Not a CPU sensor
+            return
+
+        # Now scan the directory for inputs
+        for input_filename in os.listdir(dir_fd):
+            if not input_filename.endswith('_input'):
+                continue
+
+            if predicate:
+                # We need to check the label
+                try:
+                    label_filename = input_filename.replace('_input', '_label')
+                    with Handle.open(label_filename, os.O_RDONLY, dir_fd=dir_fd) as fd:
+                        label = os.read(fd, 1024).decode().strip()
+                except FileNotFoundError:
                     continue
 
-            self.sensors.append(sensor_path)
+                if not predicate(label):
+                    continue
+
+            yield input_filename
+
+    @staticmethod
+    def scan_sensors() -> Iterable[str]:
+        with Handle.open(HWMON_PATH, os.O_RDONLY | os.O_DIRECTORY) as top_fd:
+            for hwmon_name in os.listdir(top_fd):
+                with Handle.open(hwmon_name, os.O_RDONLY | os.O_DIRECTORY, dir_fd=top_fd) as subdir_fd:
+                    for sensor in CPUTemperatureSampler.detect_cpu_sensors(subdir_fd):
+                        yield f'{HWMON_PATH}/{hwmon_name}/{sensor}'
 
     def sample(self, samples: Samples) -> None:
-        cpu_names = ['coretemp', 'cpu_thermal', 'k8temp', 'k10temp', 'atk0110']
-
-        if not self.sensors:
-            # TODO: 2 ** 32?
-            for index in range(0, 2 ** 32):
-                try:
-                    with open(f'/sys/class/hwmon/hwmon{index}/name') as f:
-                        name = f.read().strip()
-                    if name in cpu_names:
-                        self.detect_cpu_sensors(index, name)
-                except FileNotFoundError:
-                    break
+        if self.sensors is None:
+            self.sensors = list(CPUTemperatureSampler.scan_sensors())
 
         for sensor_path in self.sensors:
             with open(sensor_path) as sensor:
--- a/test/pytest/test_samples.py
+++ b/test/pytest/test_samples.py
@@ -20,12 +20,79 @@ import multiprocessing
 import numbers
 import os
 import unittest
+from unittest import mock
+from pathlib import Path
 
 import pytest
 
 import cockpit.samples
 
 
+@pytest.fixture(scope="session")
+def hwmon_mock(tmpdir_factory):
+    hwmon_dir = Path(tmpdir_factory.mktemp('hwmon'))
+
+    # hwmon1 - no name
+    Path(hwmon_dir, 'hwmon1').mkdir()
+
+    # hwmon2 - no label (on ARM)
+    hwmon2_dir = hwmon_dir / 'hwmon2'
+    hwmon2_dir.mkdir()
+    with open(hwmon2_dir / 'name', 'w') as fp:
+        fp.write('cpu_thermal')
+    with open(hwmon2_dir / 'temp1_input', 'w') as fp:
+        fp.write('32000')
+
+    # hwmon3 - AMD workaround #18098
+    hwmon3_dir = hwmon_dir / 'hwmon3'
+    hwmon3_dir.mkdir()
+    with open(hwmon3_dir / 'name', 'w') as fp:
+        fp.write('k10temp')
+    with open(hwmon3_dir / 'temp1_input', 'w') as fp:
+        fp.write('27500')
+    with open(hwmon3_dir / 'temp1_label', 'w') as fp:
+        fp.write('Tctl')
+    with open(hwmon3_dir / 'temp3_input', 'w') as fp:
+        fp.write('37000')
+    with open(hwmon3_dir / 'temp3_label', 'w') as fp:
+        fp.write('Tccd1')
+
+    # hwmon4 - Intel coretemp
+    hwmon4_dir = hwmon_dir / 'hwmon4'
+    hwmon4_dir.mkdir()
+
+    with open(hwmon4_dir / 'name', 'w') as fp:
+        fp.write('coretemp')
+    with open(hwmon4_dir / 'temp1_input', 'w') as fp:
+        fp.write('47000')
+    with open(hwmon4_dir / 'temp1_label', 'w') as fp:
+        fp.write('Package id 0')
+    with open(hwmon4_dir / 'temp2_input', 'w') as fp:
+        fp.write('46000')
+    with open(hwmon4_dir / 'temp2_label', 'w') as fp:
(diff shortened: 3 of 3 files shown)


src/bridge/cockpitcpusamples.c
```diff
@@ -156,13 +156,14 @@ detect_cpu_sensors (GList **devices,
                     const gchar *name,
                     int hwmonID)
 {
-  for (int i = 1; TRUE; i++)
+  // hwmon sensors can appear as temp1_input, temp2_input so we can't break if we don't find a sensor.
+  // As we don't want to scan 2**32 sensors we limit it to 10.
+  for (int i = 1; i < 10; i++)
```
tomasmatus, line 161: I'm wondering if 10 is good limit. According to docs for k10temp the number should stop at 10 but what about coretemp driver for intel? That uses coreID for this number which probably can go above 10 on a 16 core CPU.

I only gave this a quick look so please if you could doublecheck in case I'm wrong :)

jelly: Right, so locally (quad core) this is: So ugh no workey. I guess we do have to iterate everything?
```
[jelle@t14s][~/projects/cockpit-bots]%cat /sys/class/hwmon/hwmon9/temp*_label
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /sys/class/hwmon/hwmon9/temp1_label
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Package id 0
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /sys/class/hwmon/hwmon9/temp2_label
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Core 0
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /sys/class/hwmon/hwmon9/temp3_label
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Core 1
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /sys/class/hwmon/hwmon9/temp4_label
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Core 2
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /sys/class/hwmon/hwmon9/temp5_label
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Core 3
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

jelly:
Rewrote it to just read directories, I also added "real" pytest tests due to the usage of unittest.TestCase which is a bit weird if you use pytest I had to do some silly workarounds for fixtures.

If this approach is acceptable I'll also rewrite the C bridge handling.

jelly:
@allisonkarlitskaya I'd like to dump `unittest.Testcase` in favour of plan `def test_foo` functions, but we can do that in a new PR.

test/pytest/test_samples.py
```diff
@@ -81,3 +131,19 @@ def test_cpu_temperature(self):
...
+            # no name
+            assert 'hwmon1' not in name
+            # no label
+            assert 'hwmon2' not in name
+
+            assert 20 < temperature < 50
+
+        expected = ['hwmon4/temp4_input', 'hwmon4/temp3_input', 'hwmon4/temp2_input', 'hwmon4/temp1_input', 'hwmon3/temp3_input']
+        sensors = [os.path.join(os.path.basename(os.path.dirname(p)), os.path.basename(p)) for p in samples['cpu.temperature']]
+        assert sensors == expected
```
jelly, line 149: Well duh, list order is random silly. So unittests fails atm

src/cockpit/samples.py
```diff
@@ -121,21 +126,24 @@ def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
                 if name != 'cpu_thermal':
                     continue
 
-            self.sensors.append(sensor_path)
+            self.sensors.append(entry.path)
 
-    def sample(self, samples: Samples) -> None:
+    def sample(self, samples: Samples, hwmon_path='/sys/class/hwmon') -> None:
```
allisonkarlitskaya, line 131: No.  GitHub is rightly annoyed about this.

Use a constant defined at the toplevel and switch it out using `unittest.mock.patch`

src/cockpit/samples.py
```diff
@@ -121,21 +126,24 @@ def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
...
         cpu_names = ['coretemp', 'cpu_thermal', 'k8temp', 'k10temp', 'atk0110']
 
         if not self.sensors:
-            # TODO: 2 ** 32?
-            for index in range(0, 2 ** 32):
+            for path in os.listdir(hwmon_path):
```
allisonkarlitskaya, line 135: I'm gonna have to insist on some `*at()` variant stuff here, please.

src/cockpit/samples.py
```diff
@@ -121,21 +126,24 @@ def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
...
-                    if name in cpu_names:
-                        self.detect_cpu_sensors(index, name)
+
+                    if name not in cpu_names:
+                        continue
+
+                    self.detect_cpu_sensors(sensor_path, name)
                 except FileNotFoundError:
-                    break
+                    continue
```
allisonkarlitskaya, line 146: Is it still an expected case to see `FileNotFoundError` here?  We just iterated and found this, no?  Is this in case it hotswaps out in realtime?

test/pytest/test_samples.py
```diff
@@ -20,12 +20,62 @@
...
+def hwmon_mock(tmpdir_factory):
+    hwmon_dir = Path(tmpdir_factory.mktemp('hwmon'))
+
+    # hwmon1 - no name
+    Path(hwmon_dir, 'hwmon1').mkdir()
+
+    # hwmon2 - no label
+    hwmon2_dir = hwmon_dir / 'hwmon2'
+    hwmon2_dir.mkdir()
+    open(hwmon2_dir / 'name', 'w').write('coretemp')
```
allisonkarlitskaya, line 40: `with:` please.

allisonkarlitskaya (changes requested):

src/cockpit/samples.py
jelly: Hmm? What is this response too? Sure can do! Feels cleaner.

src/cockpit/samples.py
jelly: This is in case /sys/class/hwmon/hwmon1/name does not exists. Not sure if this can happen, but let's be safe.

`listdir` just iterates over /sys/class/hwmon/ so has no idea if `hwmon1/name` exists.

jelly:
C part done now :)

src/cockpit/samples.py
allisonkarlitskaya: I mean please use more dirfd and less path pasting.

src/cockpit/samples.py
allisonkarlitskaya: gotcha.

src/cockpit/samples.py
```diff
@@ -121,21 +129,24 @@ def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
...
     def sample(self, samples: Samples) -> None:
         cpu_names = ['coretemp', 'cpu_thermal', 'k8temp', 'k10temp', 'atk0110']
 
         if not self.sensors:
-            # TODO: 2 ** 32?
-            for index in range(0, 2 ** 32):
+            for path in os.listdir(HWMON_PATH):
                 try:
-                    with open(f'/sys/class/hwmon/hwmon{index}/name') as f:
+                    sensor_path = os.path.join(HWMON_PATH, path)
```
allisonkarlitskaya, line 140: still waiting to trade the path pasting for fds.  :)

src/cockpit/samples.py
jelly: Hmmmm `os.fwalk'?

```
>>> list(os.fwalk('/sys/class/hwmon'))
[('/sys/class/hwmon', ['hwmon8', 'hwmon6', 'hwmon10', 'hwmon4', 'hwmon2', 'hwmon0', 'hwmon9', 'hwmon7', 'hwmon5', 'hwmon3', 'hwmon1'], [], 3)]
```

doesn't give me much dirfd improvements tbh. The dirfd is from "/sys/class/hwmon" while I think you want it from "/sys/class/hwmon1".

```
>>> list(os.fwalk('/sys/class/hwmon/hwmon3/'))
[('/sys/class/hwmon/hwmon3/', ['power', 'device', 'subsystem'], ['uevent', 'temp2_min', 'temp2_max', 'temp3_input', 'temp3_min', 'temp1_label', 'temp1_crit', 'temp3_max', 'temp1_alarm', 'temp2_label', 'temp1_input', 'temp1_min', 'temp3_label', 'temp1_max', 'name', 'temp2_input'], 3), ('/sys/class/hwmon/hwmon3/power', [], ['runtime_active_time', 'runtime_status', 'autosuspend_delay_ms', 'runtime_suspended_time', 'control'], 4)]
```

And then loop over files (ignore the second loop as it's `hwmon3/power/`, I don't think fwalk can limit it's iteration.

allisonkarlitskaya:
Something like that:

```python
from systemd_ctypes import Handle
import os

TOPDIR = '/sys/class/hwmon'

with Handle.open(TOPDIR, os.O_RDONLY | os.O_DIRECTORY) as top:
    for hwmon_name in os.listdir(top):
        with Handle.open(hwmon_name, os.O_RDONLY | os.O_DIRECTORY, dir_fd=top) as sub:
            for input_name in os.listdir(sub):
                if not input_name.endswith('_input'):
                    continue
                print(hwmon_name, input_name)
```

martinpitt:
It still feels like [os.fwalk()](https://docs.python.org/3/library/os.html) would be more approrpriate here than funnelling the walking through our ctypes library? Can't Python do that with its stdlib?

jelly:
> 



> Something like that:
> 
> ```python
> from systemd_ctypes import Handle
> import os
> 
> TOPDIR = '/sys/class/hwmon'
> 
> with Handle.open(TOPDIR, os.O_RDONLY | os.O_DIRECTORY) as top:
>     for hwmon_name in os.listdir(top):
>         with Handle.open(hwmon_name, os.O_RDONLY | os.O_DIRECTORY, dir_fd=top) as sub:
>             for input_name in os.listdir(sub):
>                 if not input_name.endswith('_input'):
>                     continue
>                 print(hwmon_name, input_name)
> ```

If the goal here was to optimise for less syscalls, it sort of does.

(The whole loading of systemd_ctypes consumes ~ 1712 syscalls) but to fetch the initial sensors only takes 140 versus the 221 in the current implementation)

allisonkarlitskaya:
> If the goal here was to optimise for less syscalls, it sort of does.
>
> (The whole loading of systemd_ctypes consumes ~ 1712 syscalls) but to fetch the initial sensors only takes 140 versus the 221 in the current implementation)

It was about avoiding path pasting when possible.  Reducing syscalls is not the be-all-and-end-all.  I just consider path-pasting to be an anti-pattern when equivalent operations can be used.

I'd actually have been OK to do this with a simple call to glob.glob('.../*/*_input'), even though that's doing path pasting internally.  At least it wouldn't have been out fault :)

jelly:
> 



> > If the goal here was to optimise for less syscalls, it sort of does.
> > (The whole loading of systemd_ctypes consumes ~ 1712 syscalls) but to fetch the initial sensors only takes 140 versus the 221 in the current implementation)
> 
> It was about avoiding path pasting when possible. Reducing syscalls is not the be-all-and-end-all. I just consider path-pasting to be an anti-pattern when equivalent dirfd operations can be used.
> 
> I'd actually have been OK to do this with a simple call to `glob.glob('.../*/*_input')`, even though that's doing path pasting internally. At least it wouldn't have been out fault :)

Ok! But is the PR in good shape now? :)

src/cockpit/samples.py
```diff
@@ -101,14 +103,24 @@ class CPUTemperatureSampler(Sampler):
...
-            sensor_path = f'/sys/class/hwmon/hwmon{hwmonid}/temp{index}_input'
-            if not os.path.exists(sensor_path):
-                break
+    def detect_cpu_sensors(self, name: str, hwmon_name: str, dir_fd: int) -> None:
+        for sensor in os.listdir(dir_fd):
+            if not sensor.endswith('_input'):
+                continue
+
+            # Try opening the corresponding label
+            temp_id = sensor.split('_input')[0]
```
allisonkarlitskaya, line 112: This seems like a convoluted way of writing `.replace('_input', '_label')`.

allisonkarlitskaya (changes requested):
I'd like to see at least the split→replace change.

Also: I hate to do this to you, but I wonder if you'd want to try to do a glob-based version of this, just to see how it turns out.  The fd approach was not as elegant as I imagined...  We can discuss tomorrow, if you like.

src/cockpit/samples.py
```diff
@@ -121,21 +132,27 @@ def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
                 if name != 'cpu_thermal':
                     continue
 
-            self.sensors.append(sensor_path)
+            # need full path
+            self.sensors.append(f'{HWMON_PATH}/{hwmon_name}/{sensor}')
 
     def sample(self, samples: Samples) -> None:
         cpu_names = ['coretemp', 'cpu_thermal', 'k8temp', 'k10temp', 'atk0110']
```
allisonkarlitskaya, line 139: This probably deserves to be a top-level global constant called `CPU_NAMES`.  Or just open-coded at its single point-of-use.

src/cockpit/samples.py
```diff
@@ -121,21 +132,27 @@ def detect_cpu_sensors(self, hwmonid: int, name: str) -> None:
...
+                            fd = os.open('name', os.O_RDONLY, dir_fd=sub)
+                            try:
+                                name = os.read(fd, 1024)
+                                name = name.decode().strip()
+                                if name in cpu_names:
+                                    self.detect_cpu_sensors(name, hwmon_name, sub)
+                            finally:
+                                os.close(fd)
+                        except FileNotFoundError:
+                            continue
```
allisonkarlitskaya, line 155: We're once or twice too many times indented here.  I feel like this (highlighted section) should be a separate function.

src/cockpit/samples.py
```diff
@@ -101,14 +103,23 @@ class CPUTemperatureSampler(Sampler):
...
-                break
+    def detect_cpu_sensors(self, name: str, hwmon_name: str, dir_fd: int) -> None:
+        for sensor in os.listdir(dir_fd):
+            if not sensor.endswith('_input'):
+                continue
+
+            # Try opening the corresponding label
+            label = None
+            try:
+                fd = os.open(sensor.replace('_input', '_label'), os.O_RDONLY, dir_fd=dir_fd)
```
allisonkarlitskaya, line 114: This is a good use for 

```python
with Handle.open(...) as fd:
```

src/cockpit/samples.py
```diff
@@ -101,14 +103,23 @@ class CPUTemperatureSampler(Sampler):
...
+                    label = os.read(fd, 1024)
+                    label = label.decode().strip()
+                finally:
+                    os.close(fd)
+            except FileNotFoundError:
+                pass
 
-            with open(f'/sys/class/hwmon/hwmon{hwmonid}/temp{index}_label') as f:
-                label = f.read().strip()
             if label:
```
allisonkarlitskaya, line 123: I know this logic is pre-existing, but it seems wrong.  If labels are not used on arm, and we should accept all samplers if `name == 'cpu_thermal'`, then why do we even bother trying to open the label file?

In general, the logic here seems to be:
 - `name == 'cpu_thermal'`: accept all inputs which don't have a label file (but the comment suggests that in this case, that will be all inputs — so why check?)
 - `name == 'atk0110'`: accept results only if label is `'CPU Temperature'`
 - `name` is absolutely anything else?  accept everything except `label == 'Tctl'`?

I had an idea to (partially) lift this logic to the calling function, which is where we sample `name`, and could make the decision about what check to use just once, instead of repeatedly.  The exact nature of the checks being performed makes that inelegant, though.

I also dislike how so many separate parameters (`name`, hwmon_name`, `dirfd` and implicit access to the global `HWMON_PATH`) are inputs to this function.

It would be interesting if this function returned something more like a list of sensor input basenames and their labels, if applicable, else `None`.  If you wanted to get really fancy, you could add a `read_label` kwarg.  You could also pass in a predicate function to decide if a particular label is to be considered valid or not (or `None` to mean "all labels are valid, so don't bother reading it").

This is certainly some kind of over-engineering.  This problem well and truly drives me crazy.  Every time I look at this code, I hate it a bit more :)

allisonkarlitskaya (changes requested):
There are some minor obvious changes, as well as some rambling.

My offer to you, if you're getting sick of the iteration: I could do the next round and you review my work, if you like.

src/cockpit/samples.py
```diff
@@ -95,47 +97,67 @@ def sample(self, samples: Samples) -> None:
...
 
-    def sample(self, samples: Samples) -> None:
-        cpu_names = ['coretemp', 'cpu_thermal', 'k8temp', 'k10temp', 'atk0110']
+    @staticmethod
+    def scan_sensors() -> Iterable[str]:
+        with Handle.open(HWMON_PATH, os.O_RDONLY | os.O_DIRECTORY) as top_fd:
+            for hwmon_name in os.listdir(top_fd):
+                with Handle.open(hwmon_name, os.O_RDONLY | os.O_DIRECTORY, dir_fd=top_fd) as subdir_fd:
+                    for sensor in CPUTemperatureSampler.detect_cpu_sensors(subdir_fd):
+                        yield f'{HWMON_PATH}/{hwmon_name}/{sensor}'
```
jelly, line 156: This path mangling! :P

jelly (commented):
Cannot approve my own PR, but looks great!

src/cockpit/samples.py
allisonkarlitskaya: ya, what can you do? :)  We could store a list of fds to read from....

allisonkarlitskaya (approved):
I'm happy, obviously.  :)

jelly:
Follow up idea:

Keep the sensor fd's around so we don't have to do an open and don't have to do path mangling. Every iteration we just do `a bunch of pread() calls in a loop would be really fast`.

allisonkarlitskaya:
> Follow up idea:
> 
> Keep the sensor fd's around so we don't have to do an open and don't have to do path mangling. Every iteration we just do `a bunch of pread() calls in a loop would be really fast`.

I did a proof of concept of this — keeping the fds around as `Handle` objects yielded directly from detect_cpu_sensors()` is super easy, and allows us to replace the path-pasting loop with a `yield from`.  Where it falls over, though: the names of the instances are the filenames of the files that we read them from.  If we can replace that with something like static ints (like the fd number...) then it would be totally fine, but I think that this might be more trouble than its worth, unless we find that we really want to do this for some other reason.

PR #18137, opened 2023-01-07
tools: Fix motd/issue symlinks on upgrade from Debian stable

### LINKED ISSUE #17828: Debian: No motd for cockpit when upgrading from 239
### Explain what happens

I had cockpit 239 installed on a debian 11 system (arm64 on rpi3b+). This is the version that the debian repo provides for stable. Then I noticed that it also provides 276.1 from backports, so I installed that one because it's newer.

And a couple of days later I noticed that cockpit's info message, e.g. 
`Web console: https://rpi3b:9090/ or https://192.168.1.100:9090/`
which was there in 239, was missing from both the issue (the message above the login prompt) and from motd (the message after the login).
For the above behavior in 239, there was a cockpit-related file under both `/etc/issue.d` and /etc/motd.d/, which was actually a symlnk to `/run/cockpit/motd`.
Then I installed 277 on my main system which runs debian unstable x64 and I noticed that the file is missing there as well.

The file in /run is still there, but the symlinks are missing. So, I would like to ask if this is a change from upstream or something is broken on the packages themselves.

### Version of Cockpit

277 and 276.1

_No response_

### Server operating system

Debian

Stable (11) and Unstable

Other

### System log

```shell
(not applicable)
```

martinpitt:
Commit 5705709d4169 moved the motd/issue symlinks out of the deb towards
dynamic creation in the postinst. But this only happened on fresh
installs, not on upgrades. We still want to retain (i.e. create) the 
files when upgrading from < 244. We just want to avoid bringing them
back if the admin removed them and upgraded from a later version.

Fixes #17828

---

Plus another commit to drop autom4te.cache/ from the tarball, see https://github.com/cockpit-project/cockpit/issues/17828#issuecomment-1374526261


tools/debian/cockpit-ws.postinst
```diff
@@ -21,8 +21,8 @@ if [ -d /run/systemd/system ] && [ -n "$2" ]; then
     deb-systemd-invoke try-restart cockpit.service >/dev/null || true
 fi
 
-# set up dynamic motd/issue symlinks on first-time install; don't bring them back on upgrades if admin removed them
-if [ "$1" = "configure" ] && [ -z "$2" ]; then
+# set up dynamic motd/issue symlinks on first-time install or upgrades from < 244 (which moved them out of the .deb)
+if [ "$1" = "configure" ] && dpkg --compare-versions "$2" lt 244; then
```
KKoukiou, line 25: Should not the && be || ?

martinpitt: No, && is correct. This should only happen for the "configure" phase, not e.g. for "abort-remove" or "abort-upgrade".

KKoukiou: But with this change on fresh installs which are *not* upgrades from < 244 this file creation code will not run. 

From what I understand you want to do:
`if is_configure_phase || (is_configure_phase && dpkg --compare-versions "$2" lt 244)`

KKoukiou: OH I see, after version 244 this file was present. Sorry missed that.

KKoukiou (dismissed):

tools/debian/cockpit-ws.postinst
martinpitt: `--compare-versions` treats an empty version as earlier than any other, so `'' lt 244` is true, and the code runs as intended. FTR, this isn't always the case, for that there's also `lt-nl` operator ("none is later").

martinpitt:
@allisonkarlitskaya , do you have an opinion about the autom4te.cache fix? It could be put somewhere else as well.

allisonkarlitskaya:
My drive-by: this feels wrong.  Why does this end up there in the first place if it's not meant to be disted?

Is this a side-effect of rerunning autogen in the distdir to get the new version number?  If so, I'd move the `rm` to immediately follow that, and add a comment about it, and probably better explain it in the commit message.

I might also start thinking again about the hacks we do with the version... this certainly wasn't an intended side-effect...

martinpitt:
@allisonkarlitskaya : It feels like fallout from commit e2f2d830e1d15, but I didn't bisect (and not going to today).  Moving the rm into this rule sounds better indeed:
```
$(distdir)/configure: $(distdir)/version.m4
      @rm -f $(distdir)/configure
      $(AM_V_GEN) autoreconf $(distdir)
```
I'll do that.

martinpitt:
@allisonkarlitskaya Done, PTAL?

martinpitt:
[This accounts test](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18137-20230108-190112-10b3836c-fedora-36/log.html#324) is unrelated, but happens very often now. This cannot simply be reproduced by running the test twice in a row, as users are reset between tests. Also, /var/log/btmp is empty afterwards, and /var/log/wtmp does not contain `anton`.

I grepped the log for "ND@2". Running `TestAccounts.testUnprivileged` first does reproduce the bug, afterwards `grep -r anton /var/log` says "/var/log/wtmp: binary file matches". So let's save/restore that file.

allisonkarlitskaya (approved):
+1 from me for the `autom4te.cache` business.

PR #18142, opened 2023-01-09
test: add --python option to image-prepare

allisonkarlitskaya:
This will prepare an image with the Python bridge for manual testing, in the same way as was done for the /pybridge scenario in test/run.

Simplify test/run to reflect the fact that this is now just an option to image-prepare.


martinpitt (approved):
Yay, this is nice. Thanks!

PR #18144, opened 2023-01-09
systemd: ignore subpolicy for FIPS crypto policies

### LINKED ISSUE #18062: Crypto for NIST 800-171
### Explain what happens

According to the OSCAP NIST 800-171 security profile, FIPS mode must be enabled and crypto policy configured as FIPS:OSPP
That is to say:
- sudo fips-mode-setup --enable: http://static.open-scap.org/ssg-guides/ssg-rhel8-guide-cui.html#xccdf_org.ssgproject.content_rule_enable_dracut_fips_module
- sudo update-crypto-policies --set FIPS:OSPP http://static.open-scap.org/ssg-guides/ssg-rhel8-guide-cui.html#xccdf_org.ssgproject.content_rule_configure_crypto_policy

However, when it is configured like this, Cockpit shows "Fips:ospp" inconsistent.
The view is therefore not correct.

### Version of Cockpit

276.1

Overview

### Server operating system

Red Hat Enterprise Linux

RHEL 8.7

Chrome

### System log

_No response_

jelly:
Our FIPS check compared the whole policy including subpolicies, as FIPS:OCPP is a valid FIPS policy we should only check the policy.

Closes #18062


test/verify/check-system-info
```diff
@@ -947,6 +947,9 @@ password=foobar
         b.wait_text("#crypto-policy-button", "FIPS")
         self.assertEqual(m.execute("cat /proc/sys/crypto/fips_enabled").strip(), "1")
 
+        m.execute(cmd + " --set FIPS:OSPP")
+        b.wait_not_present("#inconsistent_crypto_policy")
```
martinpitt, line 951: This is pretty much a no-op. `wait_not_present` always needs to be preceeded by some "positive" change of that --set action. I suggest to reorder this after the `--set DEFAULT` below -- then do the `FIPS:OSPP` and wait for the UI to show "FIPS", and *then* the `self.assertFalse(b.is_present(#inconsistent))` makes sense.

martinpitt (changes requested):
Thanks! the fix is nice, but let's fix the test.

test/verify/check-system-info
```diff
@@ -947,6 +947,12 @@ password=foobar
         b.wait_text("#crypto-policy-button", "FIPS")
         self.assertEqual(m.execute("cat /proc/sys/crypto/fips_enabled").strip(), "1")
 
+        m.execute(cmd + " --set DEFAULT")
+        b.wait_text("#inconsistent_crypto_policy", "Crypto policy is inconsistent")
+        m.execute(cmd + " --set FIPS:OSPP")
+        b.wait_text("#crypto-policy-button", "Fips:ospp")
+        self.assertFalse(b.is_present("#inconsistent_crypto_policy"))
```
martinpitt, line 954: Ah, for this case a wait_present() would actually be correct, the `#inconsistent-crypto-policy` object actually was present before, and with react the state update would be async. In the previous version of the patch, the "inconsistent" was not checked to be present.

So, one more round with wait_not_present here, and it's good!

martinpitt (changes requested):

martinpitt (approved):
Cheers!

Issue #18125, opened 2023-01-05
Follow  some more fine grained approach for applying the tabular fonts class  from Patternfly


KKoukiou:
We now insert this class on the  top of all page, thus making anything numeric (there are exceptions ?!) use tabular fonts. We should rather focus  on adding this only to elements that need this. For example sliders values.

martinpitt:
Big +1. IMHO this looks super-ugly right now, as the monospaced text is being used overzealously everywhere. We have wild font inconsistencies like this now:

![image](https://user-images.githubusercontent.com/200109/210770096-cf09e1a3-5efd-4188-a175-dab6f4e8b29e.png)

![image](https://user-images.githubusercontent.com/200109/210770135-a31a959c-35d7-4b47-8e62-d7c0d30cd114.png)

Notice how we format zeros differently in the Configuration card, and  how ugly the dates are.

Monospaced numbers only make sense in some tabular environments where they *actually* need to be aligned. The only place where this is the case on the overview page is here:

![image](https://user-images.githubusercontent.com/200109/210770342-13c1c5bb-6394-4d3e-aa5e-4402091c81d8.png)

But ironically it doesn't even help, as the progress bars still have different lengths.

I went through all the pages, and the only other potential usages of tabular numbers that I found are on hwinfo:

![image](https://user-images.githubusercontent.com/200109/210770534-e795d661-81e2-4eaa-9282-599b7541a00f.png)

(Although it's not really necessary there)

and to ensure properly aligned timestamps on the Networking page's Log view:

![image](https://user-images.githubusercontent.com/200109/210770620-c282851d-6727-461f-bf65-310c30852e02.png)

But in all of these cases we'd actually be better off with using a proper monospace font.

In other words: These need to be opt-in and carefully/explicitly used, instead of "everywhere". They are called *tabular* for a reason? 

Also, I find these zeros super ugly, at first I thought my system was broken and somehow missed proper fonts -- but that's a matter of taste. But more objectively, paragraph text or most text which doesn't rely on alignment should use proportional fonts.

garrett:
The 0 with a slash is a bug in the font; it should have variants with and without the slash, and it should be controlled by a font feature (which is exposed in CSS).

Tabular numbers are needed for columns in tables that have numbers and anywhere where numbers are compared, including numbers compared with themselves over time (such as progress bars and sliders) and numbers compared with other numbers (which is why I mentioned columns in tables). It's also useful for things like IP addresses.

[...]

PR #18145, opened 2023-01-09
pf4: Unslash tabular zeros via subset & range

garrett:
Partially addresses #18125

This is the least intrusive change; fonts are embedded as to:
- Not require buildsystem changes to ship font files
- Not inflate number of connections from browser to Cockpit
- Be easily portable to cockpit-machines, cockpit-podman, etc.

It's messy when viewing the stylesheet, so they're at the bottom of the file.

Additionally, subsetting and mapping these custom versions of the fonts are much smaller in size than forking the font and using our own custom version, which would be worse.


garrett:
We also don't use tabular numbers everywhere that we should. Look at disks in this screenshot (the numbers on the usage bars, before "free" are all proportional):

image

KKoukiou (commented):
Whoa @garrett that's cool. 
Small question. I see in the pixel tests, even zeros without the slash changed enough to make pixel tests fail. Is that expected? 

https://cockpit-logs.us-east-1.linodeobjects.com/pull-18145-20230109-173929-c6864d3a-fedora-36/pixeldiff.html#TestBonding-testBasic-networking-bond-settings-dialog

garrett:
>  Small question. I see in the pixel tests, even zeros without the slash changed enough to make pixel tests fail. Is that expected?

Yes, see the pixel diff @ https://github.com/cockpit-project/cockpit/issues/18125#issuecomment-1375678940. I didn't get the hinting exactly right, as I had to recreate the inner part without the slash. It's all sub-pixel and nobody would notice except the pixel tests.

After a bit of digging around: The zeros for bold in this PR are actually using bold, not semi-bold. PatternFly is using different weight numbers than I had expected. Fixing now. (They purposefully push the semibold as bold and bold as heavy... and don't use the heavy. I guess because they thought the bold was a bit too much.)

garrett:
I've dropped the unused weights and recalibrated the weight values (as PF uses nonstandard weights to make "bold" show a medium weight instead, as the bold is pretty heavy for a UI). 

The adjusted weights are the screenshot was off last night. It has been fixed in the PR.

KKoukiou:
@garrett when you slide in this screenshot https://cockpit-logs.us-east-1.linodeobjects.com/pull-18145-20230110-084945-faa442a8-fedora-36/pixeldiff.html#TestStorageBasic-testBasic-partition
it's obvious that the new numbers are smaller width than the previous ones. I though we keep the width for the tabular nums as is. Any idea what's causing this issue?

garrett:
@KKoukiou: No, they're not smaller width. I made sure to preserve the widths. Perhaps tabular isn't being respected and it's dropping in the normal 0, which _is_ a different width? That's the only explanation I can see for differing widths. It seemed to work perfectly in the first try; I did try to optimize the default one a bit more to get the filesize down... I might've deleted too much metadata.

If that's the case, it's so minor that perhaps we should just standardize the 0 to the tabular width regardless. This would have the side effect of making the file size (and therefore base64 string) smaller too, since it would be just 1 glyph instead of 2.

Or I could try to make sure I don't delete metadata mapping.

Also, I think this is just wrong for hashes: Hashes should use probably monospace, not tabular data. (And therefore, they'd have the 0 with the slash, as I didn't touch monospace.)

KKoukiou:
The new zero width is obviously bigger than the tabular one. it's visible in most of the screenshots 
if you use the slider to move between the current and previous pixel test 
https://cockpit-logs.us-east-1.linodeobjects.com/pull-18145-20230110-084945-faa442a8-fedora-36/pixeldiff.html#TestAccounts-testBasic-users-page-medium

I am not trying to be picky, but this cancels the point of tabular - fixed width nums

garrett:
Original was correct; I did mess up trying to optimize the default. I swapped out the normal one to the original test and it works now.

(I didn't try to optimize the variants, so those should be fine.)

Slash from main:

Screenshot 2023-01-10 at 14-23-54 Networking - garrett@Bolt

This PR (now, fixed):

Screenshot 2023-01-10 at 14-30-06 Networking - garrett@Bolt

Diff:

image

KKoukiou (dismissed):

KKoukiou (approved):

PR #18143, opened 2023-01-09
systemd: Fix ObjectManager path for realmd, improve error logging

martinpitt:
realmd's ObjectManager is on the primary manager object (`/org/freedesktop/realmd` aka. `MANAGER`), not at proxies()'s default path_namespace `/`.

The C bridge papered over this with its ObjectManager emulation, but the Python bridge does not do that.

Unfortunately testQualifiedUsers is still broken with the Python bridge, it fails later on the Services page with "interactive authentication required". This is unrelated to realmd, though.

-----

[attempt 1](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18143-20230109-102959-5a0811fe-fedora-37-pybridge/log.html#32)

PR #18147, opened 2023-01-10
python: Fix "python.error" JS warning

martinpitt:
I am debugging the "python.error" in the log which we get on the Overview page, which is throughly unhelpful. The first commit turns this into both a JS log:
```

  File "/usr/lib/python3.11/site-packages/cockpit/channels/dbus.py", line 321, in do_call
    reply = await self.bus.call_method_async(self.name, path, iface, method, signature, *args,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 382, in call_method_async
    message = self.message_new_method_call(destination, path, interface, member, types, *args)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 349, in message_new_method_call
    message.append(types, *args)

  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 156, in append
    assert len(infos) == len(args)
           ^^^^^^^^^^^^^^^^^^^^^^^

AssertionError
index.js:2:637875
```
as well as a journal message:
```
cockpit-ws[5707]: WARNING:cockpit.channels.dbus:do_call({'call': ['/config', 'cockpit.Config', 'GetUInt', ['Session', 'IdleTimeout', 0, 240, 0]], 'id': '4'}): generic exception Traceback (most recent call last):
cockpit-ws[5707]:   File "/usr/lib/python3.11/site-packages/cockpit/channels/dbus.py", line 321, in do_call
cockpit-ws[5707]:     reply = await self.bus.call_method_async(self.name, path, iface, method, signature, *args,
cockpit-ws[5707]:             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cockpit-ws[5707]:   File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 382, in call_method_async
cockpit-ws[5707]:     message = self.message_new_method_call(destination, path, interface, member, types, *args)
cockpit-ws[5707]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cockpit-ws[5707]:   File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 349, in message_new_method_call
cockpit-ws[5707]:     message.append(types, *args)
cockpit-ws[5707]:   File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 156, in append
cockpit-ws[5707]:     assert len(infos) == len(args)
cockpit-ws[5707]:            ^^^^^^^^^^^^^^^^^^^^^^^
cockpit-ws[5707]: AssertionError
```

The second commit fixes the root cause of this, the cockpit.Config.GetUInt() stub was wrong. As the next step, I'm actually going to implement it, which will fix a few more tests.

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -40,6 +40,7 @@ import asyncio
 import errno
 import json
 import logging
+import traceback
 import xml.etree.ElementTree as ET
 
 from systemd_ctypes import Bus, BusError, introspection
@@ -330,7 +331,9 @@ class DBusChannel(Channel):
             # actually, should send the fields from the message body
             self.send_message(error=[error.name, [error.message]], id=cookie)
         except Exception as exc:
-            self.send_message(error=['python.error', [str(exc)]], id=cookie)
+            logger.exception("do_call(%s): generic exception", message)
+            printable = '\n'.join(traceback.format_exception(exc))
+            self.send_message(error=['python.error', [printable]], id=cookie)
 
     async def do_add_match(self, message):
         add_match = message['add-match']
--- a/src/cockpit/internal_endpoints.py
+++ b/src/cockpit/internal_endpoints.py
@@ -31,8 +31,8 @@ class cockpit_Config(bus.Object):
     def __init__(self):
         ...
 
-    @bus.Interface.Method(out_types='u', in_types='suuu')
-    def get_u_int(self, name, _minimum, default, _maximum):
+    @bus.Interface.Method(out_types='u', in_types='ssuuu')
+    def get_u_int(self, _section, _key, default, _minimum, _maximum):
         return default
 
 


src/cockpit/channels/dbus.py
```diff
@@ -330,7 +331,9 @@ async def do_call(self, message):
             # actually, should send the fields from the message body
             self.send_message(error=[error.name, [error.message]], id=cookie)
         except Exception as exc:
-            self.send_message(error=['python.error', [str(exc)]], id=cookie)
+            printable = '\n'.join(traceback.format_exception(exc))
+            logger.warning("do_call(%s): generic exception %s", message, printable)
```
jelly, line 335: Alternatively, use `logger.exception`? https://docs.python.org/3/library/logging.html#logging.Logger.exception

I don't think you need the `printable part them. You do ofcourse for `send_message` so gains seem minimal.

martinpitt: But that logs at level "error", which we cannot do right now -- if we do that, we need to fix the pybridge/cockpit.js/storage page to not trigger this error if stratis is not found. See https://github.com/cockpit-project/cockpit/pull/18143#issuecomment-1375638126

martinpitt: Wait, ignore me -- that was a different error. I'll look into that, thanks!

martinpitt: The journal output looks pretty much identical except for s/WARNING/ERROR/. Indeed the code diff is minimal, but at least a little more idiomatic:
```
--- src/cockpit/channels/dbus.py
+++ src/cockpit/channels/dbus.py
@@ -331,8 +331,8 @@ class DBusChannel(Channel):
             # actually, should send the fields from the message body
             self.send_message(error=[error.name, [error.message]], id=cookie)
         except Exception as exc:
+            logger.exception("do_call(%s): generic exception", message)
             printable = '\n'.join(traceback.format_exception(exc))
-            logger.warning("do_call(%s): generic exception %s", message, printable)
             self.send_message(error=['python.error', [printable]], id=cookie)
 
```

So I'll push this.

allisonkarlitskaya (changes requested):
The real issue here is that we're using a highlevel wrapper function (`.call_method_async()`) to do low-level work with arguments passed to us from someone else, without checking for validity.

We need a way to make sure that `*args` and `signature` actually fit to each other, and if that's not the case, throw a more meaningful error message.  The assertion from systemd_ctypes isn't good enough here.

One thing we could trivially do is check the number of items in the signature vs. the number of arguments.  That would have caught the issue that's been causing us trouble here, but it wouldn't catch another equally concerning issue: the case of someone trying to pass off 5 as a string (etc).  systemd_ctypes doesn't have a good API for that, unfortunately.

I think my preferred solution to this would be to attempt to manually construct the message, one argument at a time, and send it using the lower-level API.  If we get an exception while attempting to create the message, we can be pretty sure of what the issue was.

martinpitt:
@allisonkarlitskaya : Agree that this could become more robust (follow-up), but how does that make these two commits invalid? IMHO we still need a generic fallback handler for arbitrary exceptions and show them, instead of just "python.error". In other words, what do you want me to change in these two commits?

allisonkarlitskaya:
Discussed this in stand-up: Landing this now to unblock the config PR.  I'll look at doing a follow-up.

Issue #18047, opened 2022-12-15
Pull request preview environments to improve maintainer productivity


gadkins:
### Proposal

I would like to support cockpit by implementing [Uffizzi](https://github.com/UffizziCloud/uffizzi) preview environments.
Disclaimer: I work on the Uffizzi project.

Uffizzi is an open-source, full-stack previews engine. Our platform is available completely free for cockpit (and all open-source projects). This will provide maintainers with preview environments of every PR in the cloud, which enables faster iterations and reduces time to merge. You can see the open-source repos which are currently using Uffizzi over [here](https://uffizzi.notion.site/).

Uffizzi is purpose-built for the task of previewing PRs, and it integrates with your workflow to deploy preview environments in the background without any manual steps for maintainers or contributors. 

We can go ahead and create an initial PoC for you right away if you think there is value in this proposal.
cc @waveywaves

### To Do

- [ ] Initial PoC

jkonecny12:
Hi @gadkins, I'm working on [Anaconda](https://github.com/rhinstaller/anaconda) project where we are now working on Web UI frontend build on top of Cockpit. I wonder would it be possible to use your solution for us?
The biggest issue for us could be that we need to be able to boot the machine and then connect to the remote URL. Do you think that is doable?

gadkins:
@jkonecny12 Seems doable

jkonecny12:
In that case we will definitely take a look :). Thanks!

marusak:
We actually used to have something very much similar before - https://github.com/cockpit-project/cockpit/pull/15384
I don't remember why but we eventually dropped it. @KKoukiou might remember why

Not sure what the team thinks of this. For example @martinpitt @KKoukiou @jelly WDYT?

martinpitt:
Our /livetest had been broken for a while, required some questionable infrastructure, only worked on ubuntu 22.04 (what GitHub gives us), and nobody in our team actually cared enough to keep it alive. We made building a local VM for a PR so fast (< 1 min) that the benefits of an online test env weren't big enough for the additional overhead.

That said, if Uffizzi makes this work without custom infrastructure on our side, this is definitively interesting. Maybe even more for starter-kit and friends, so that it's easy for the community to help with additional plugins. So I'd much appreciate a PoC. Thanks @gadkins !

PR #18149, opened 2023-01-10
Integrate pull request preview environments 

waveywaves:
This PR adds a github workflow which will build the cockpit image and push it to the registry.uffizzi.com registry which is an ephemeral images registry. This image will be then used by Uffizzi to deploy the preview for the PR. Initially we are hoping this works as a bastion cockpit instance for you but in subsequent PRs we can work on standing up another instance of cockpit which will work as a normal server with cockpit installed. 

This contents of this PR have been tested against my own fork of cockpit https://github.com/waveywaves/cockpit/pull/1. 
After this PR is merged, you can expect all subsequent PRs to this repo to have a a comment like https://github.com/waveywaves/cockpit/pull/1#issuecomment-1374946004 pinned which will contain the link to the preview environment url like https://app.uffizzi.com//github.com/waveywaves/cockpit/pull/1 which was created by the github action workflow.

If you want more information on the github action workflow and why it has been written in two different files the way it is, I would suggest you to checkout a blogpost we wrote as an explainer over [here](https://www.uffizzi.com/preview-environments-guide/github-actions-two-stage-workflow).

related issue https://github.com/cockpit-project/cockpit/issues/18047

### DIFF
--- a/.github/workflows/uffizzi-build.yaml
+++ b/.github/workflows/uffizzi-build.yaml
@@ -1,0 +1,87 @@
+name: Build PR Image
+on:
+  pull_request:
+    types: [opened,synchronize,reopened,closed]
+
+jobs:
+
+  build-cockpit:
+    name: Build and Push `cockpit`
+    runs-on: ubuntu-latest
+    if: ${{ (github.event_name != 'pull_request' || github.event.action != 'closed') && contains(github.event.head_commit.message, '/livetest')}}
+    outputs:
+      tags: ${{ steps.meta.outputs.tags }}
+    steps:
+      - name: Checkout git repo
+        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Generate UUID image name
+        id: uuid
+        run: echo "UUID_TAG_APP=$(uuidgen)" >> $GITHUB_ENV
+      - name: Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: registry.uffizzi.com/${{ env.UUID_TAG_APP }}
+          tags: type=raw,value=60d
+      - name: Build and Push Image to registry.uffizzi.com ephemeral registry
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          context: ./containers/ws
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+
+  render-compose-file:
+    name: Render Docker Compose File
+    # Pass output of this workflow to another triggered by `workflow_run` event.
+    runs-on: ubuntu-latest
+    needs:
+      - build-cockpit
+    outputs:
+      compose-file-cache-key: ${{ steps.hash.outputs.hash }}
+    steps:
+      - name: Checkout git repo
+        uses: actions/checkout@v3
+      - name: Render Compose File
+        run: |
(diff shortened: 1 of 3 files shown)


KKoukiou (commented):
Hi :) First thanks for taking the time to implement this. 

I would say that making this default run with each PR is a bit overkill and unnecessary waste of uffizzis resources. I would like to talk about it with the rest of the cockpit team, but I feel on demand run of uffizzi workflow with listening for some comment is the direction I would prefer.

@martinpitt any feedback  on this?

.github/workflows/uffizzi-build.yaml
```diff
@@ -0,0 +1,87 @@
+name: Build PR Image
+on:
+  pull_request:
+    types: [opened,synchronize,reopened,closed]
```
martinpitt, line 4: I agree with @KKoukiou about not triggering this for every PR run. And this is especially not needed on "closed".

We used to trigger this with a magic comment `/livetest`, see the [old workflow](https://github.com/cockpit-project/cockpit/blob/266/.github/workflows/livetest.yml). Something similar could be done here.

martinpitt: Ah, nevermind, "closed" is used for cleaning up. But what about cleaning up old instances from force pushes? Don't they pile up?

.github/workflows/uffizzi-build.yaml
```diff
@@ -0,0 +1,87 @@
...
+    name: Build and Push `cockpit`
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' || github.event.action != 'closed' }}
+    outputs:
+      tags: ${{ steps.meta.outputs.tags }}
+    steps:
+      - name: Checkout git repo
+        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
```
martinpitt, line 18: I had a look at https://github.com/docker/setup-buildx-action , but this seems very generic. It wraps https://github.com/docker/buildx, which again wraps https://github.com/moby/buildkit, i.e. this very quickly gets to the point when this is completely non-understandable nor reproducible locally, and presumably will also take a lot longer than necessary. We already know how to build cockpit, and should re-use that here.

I'm afraid I don't yet know anything about the uffizi environment (not even whether it's deb, rpm, arch, or something else), so that will take some time to learn.

Did you test this on a fork? How do these environments look like?

Thanks!

martinpitt: How do you log in there? I.e. which user/password/host? I don't see this on https://github.com/UffizziCloud/uffizzi. So far it looks like it's running a cockpit/ws container in bastion mode, but that's usually not the interesting part. Thanks!

.github/workflows/uffizzi-build.yaml
martinpitt: Sure, you can/should use the default GITHUB_TOKEN for that. However, I'm happy to work on this part, I've done it before and it's not hard. So it's fine for you to leave the trigger alone in this PR.

.github/workflows/uffizzi-build.yaml
martinpitt: @waveywaves : I'm afraid I don't understand your answer. Right now, on https://app.uffizzi.com//github.com/waveywaves/cockpit/pull/1 I see the cockpit login page. In order to to see the actual cockpit session, one  needs to log into the machine which has the packages/pages from the PR installed. Otherwise, the only thing you could test was the login page itself (which doesn't change often, and isn't very interesting). So what user/password do I need to type there to log into the temporary server?

martinpitt:
This is quite stale, and has some unresolved important questions open. Do you still care about this?

PR #18150, opened 2023-01-10
python: Implement /config and /environment internal bus objects

martinpitt:
These behave like the C bridge, except that the bridge now logs a
warning if a config key is expected to be an integer, but isn't. That
will help admins to track down why their config change is not effective.

----

 - Builds on top of PR #18147

### DIFF
--- a/src/cockpit/config.py
+++ b/src/cockpit/config.py
@@ -15,10 +15,65 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import configparser
+import logging
 import os
 
 from pathlib import Path
 
+from systemd_ctypes import bus
+
+logger = logging.getLogger(__name__)
+
 ETC_COCKPIT = Path('/etc/cockpit')
 XDG_CONFIG_HOME = Path(os.getenv('XDG_CONFIG_HOME') or os.path.expanduser('~/.config'))
 DOT_CONFIG_COCKPIT = XDG_CONFIG_HOME / 'cockpit'
+
+
+class Config(bus.Object, interface='cockpit.Config'):
+    def __init__(self):
+        self.reload()
+
+    @bus.Interface.Method(out_types='s', in_types='ss')
+    def get_string(self, section, key):
+        try:
+            return self.config[section][key]
+        except KeyError:
+            raise bus.BusError('cockpit.Config.KeyError', f'key {key} in section {section} does not exist')
+
+    @bus.Interface.Method(out_types='u', in_types='ssuuu')
+    def get_u_int(self, section, key, default, maximum, minimum):
+        try:
+            value = self.config[section][key]
+        except KeyError:
+            return default
+
+        try:
+            int_val = int(value)
+        except ValueError:
+            logger.warning(f'cockpit.conf: [{section}] {key} is not an integer')
+            return default
+
+        return min(max(int_val, minimum), maximum)
+
+    @bus.Interface.Method()
+    def reload(self):
+        self.config = configparser.ConfigParser(interpolation=None)
+
+        config_dirs = os.environ.get('XDG_CONFIG_DIRS', '/etc').split(':')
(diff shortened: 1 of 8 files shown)


martinpitt:
[weird failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18150-20230110-113113-8520a473-fedora-37-pybridge/log.html#259) -- the test already allows the message, but apparently now it's prefixed with WARNING:cockpit-bridge:". I have no idea why -- None of the patches here change bridge.py's logger. This must be some strange side effect of the logger introduced into src/cockpit/config.py. But the same thing happens in all channels/*.py files..

allisonkarlitskaya (commented):
Honestly — for a first pass, this is really great work.  See the one comment below about an improvement.  It's not a hard-no (since the code you wrote looks *so* nice), but I think a few extra lines would get us a better result there.

src/cockpit/config.py
```diff
@@ -15,10 +15,61 @@
...
+            if os.path.exists(config_file):
+                logger.debug("cockpit.Config: loading %s", config_file)
+                self.config.read(config_file)
+                break
+
+        # it's ok to not have a config file and thus leave self.config empty
+
+
+class Environment(bus.Object, interface='cockpit.Environment'):
+    variables = bus.Interface.Property('a{ss}', value=os.environ)
```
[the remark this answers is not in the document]
martinpitt: Good point, done. If we accept the environment as malleable, then it would actually be cleaner to have a property getter function that returns the current os.environ at the time of calling Get() -- but AFAICS systemd_ctype's API does not have a concept of a property getter function, or I'm not seeing it?

Anyway, we only use this endpoint in two unit tests, and changing the environment of a running process is fragile at best in Linux, so I suppose right now we don't need to go overboard with that.

martinpitt:
Thanks @allisonkarlitskaya . Ready for re-reviewing.

martinpitt:
Hmm, need to make sense of [this](https://github.com/cockpit-project/cockpit/actions/runs/3885307132/jobs/6629327537):
```
log: not ok 65 - signal received, expected: true, got: false, test: proxy signal, module: , source: 347/</l/</<@http://localhost:44425/dist/base1/test-dbus.js:2:155440
```
I'm not getting this with `make pytest`, and `make pytest-cov` bombs out early for me:

    ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]

Puzzle for tomorrow!

martinpitt:
Ah, it was just a flake, all green now.

allisonkarlitskaya (dismissed):
I'm happy enough with how this is now, but: what happened to your getter idea?  I liked that.

martinpitt:
@allisonkarlitskaya If you tell me how to write a property getter, I'm happy to add it. I tried a naïve approach of assigning a lambda to the value,and that doesn't work.

martinpitt:
I sent a test case to https://github.com/allisonkarlitskaya/systemd_ctypes/pull/44

martinpitt:
@allisonkarlitskaya Changed Environment to a getter function.

src/cockpit/config.py
```diff
@@ -15,10 +15,65 @@
...
+
+        # it's ok to not have a config file and thus leave self.config empty
+
+
+class Environment(bus.Object, interface='cockpit.Environment'):
+    variables = bus.Interface.Property('a{ss}')
+
+    @variables.getter
+    def get_variables(self):
+        return os.environ.copy()
```
allisonkarlitskaya, line 79: I was thinking about this since yesterday, and how passing `os.environ` itself would be a weird choice, but that making a copy would also be strange.  It's annoying that Python doesn't really have value types, and that everything is a reference to mutable state... but that's Python.

In any case, I think the cautious approach is also completely fine.

allisonkarlitskaya: In any case, this is much much better.  Thanks!  And thanks for the tests.

allisonkarlitskaya (approved):

PR #18156, opened 2023-01-11
pybridge: Don't timeout D-Bus method calls by default

mvollmer:
This is what the C bridge does, and some parts of Cockpit rely on it.

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -289,7 +289,7 @@ class DBusChannel(Channel):
 
     async def do_call(self, message):
         path, iface, method, args = message['call']
-        timeout = message.get('timeout')
+        timeout = message.get('timeout', 2**64 - 1)
         cookie = message.get('id')
         flags = message.get('flags')
         type = message.get('type')
--- a/test/verify/check-superuser
+++ b/test/verify/check-superuser
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+import time
+
 import parent  # noqa: F401
 from testlib import MachineCase, nondestructive, skipDistroPackage, skipImage, todoPybridge, test_main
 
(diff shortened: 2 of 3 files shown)


src/cockpit/channels/dbus.py
```diff
@@ -289,7 +289,7 @@ def sync_handler(message):
 
     async def do_call(self, message):
         path, iface, method, args = message['call']
-        timeout = message.get('timeout')
+        timeout = message.get('timeout', 2**63)
```
martinpitt, line 292: This feels ugly. `timeout` is an `uint64_t`, so this *should* be safe on 32 bit systems.

I see that systemd_ctypes defaults to 0:

    super().call_async(pending, message, pending.callback, pending.userdata, timeout or 0)

[sd_bus_call_async()](https://man.archlinux.org/man/sd_bus_call_async.3.en) indeed does not have a mode for "indefinite timeout" -- that's sad, I would have liked "0" for "indefinite" and "-1" for "default".

So this is good enough for me. `2**64 -1` aka `MAX_UINT64` (like in pkg/systemd/services/services.jsx) would be a bit more standard, but as it's not a magic value, it doesn't make much difference either way.

martinpitt (approved):
Thank you! I left some thinking aloud below.

src/cockpit/channels/dbus.py
allisonkarlitskaya, line 292: `2**64 - 1`, yes?

Not that it matters.  Just looks weird.

test/verify/check-superuser
```diff
@@ -341,6 +345,23 @@ session include system-auth
         b.wait_not_present(".pf-c-modal-box:contains('Switch to administrative access')")
         b.wait_not_present(".pf-c-alert:contains('Web console is running in limited access mode.')")
 
+    @unittest.skipIf(DEFAULT_IMAGE != "fedora-coreos", "Slow test, runs only on fedora-coreos")
```
allisonkarlitskaya, line 348: Yikes.  Can you take a look at this @martinpitt ?  Is this the best way?

src/cockpit/channels/dbus.py
mvollmer: Yeah, `2**64-1` should work, but something made me want to play it safe... Let's just use `2**64-1`.

mvollmer: Yes, indeed.

test/verify/check-superuser
martinpitt: It could move into an existing test with a more standard `if m.image == ["fedora-coreos"]:` check, which would also save a VM start/boot.

mvollmer: Actually, we want to test this with the python bridge, no? So fedora-coreos is the wrong image to run this on.

mvollmer:
TestStorageFormat.testFormatCancel now passes. Could that be our test that checks whether D-Bus calls time out?

allisonkarlitskaya (approved):

PR #18157, opened 2023-01-11
terminal: update xterm dependency and start using an add on for the canvas

### LINKED ISSUE #18138: Cockpit terminal is nearly unusable for use on mobile phones
I was using cockpit for a bit and I'm still using it but only thing that bothers me and thing that makes me not use cockpit as much as I would want is extremely bad experience with terminal on mobile. First idea of starting using cockpit was to close ssh and use only cockpit but after I found out that when I want to delete 1 character I must delete the whole word once then I can remove the one character i started using it less and less. This wasn't too big of a problem normally but when u have a big command and need to edit something at the start I usually just copied it to some text editor. But fixing this would make a lot better experience to use cockpit on daily basis.

KKoukiou:
The 'xterm-canvas-addon' is now used because the new version of the original xterm renderer is using inline styles which we block with our strict CDP. https://github.com/xtermjs/xterm.js/issues/4133

The above add on was chosen against https://www.npmjs.com/package/xterm-addon-webgl because we can't rely on webgl being enabled on users browsers.

With the usage of the renderer addon we can now remove most of the custom CSS, which is anyway not applied as many of the classnames in xterm Terminal changed.

The new xterm release was a breaking changes one and thus some of the component properties changed - therefore the c-c-terminal was adopted to the new APIs.

Fixes #18138

### DIFF
--- a/pkg/lib/cockpit-components-terminal.jsx
+++ b/pkg/lib/cockpit-components-terminal.jsx
@@ -21,6 +21,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Modal, Button } from "@patternfly/react-core";
 import { Terminal as Term } from "xterm";
+import { CanvasAddon } from 'xterm-addon-canvas';
 
 import { ContextMenu } from "cockpit-components-context-menu.jsx";
 import cockpit from "cockpit";
@@ -136,11 +137,17 @@ export class Terminal extends React.Component {
         if (props.onTitleChanged)
             term.onTitleChange(props.onTitleChanged);
 
-        this.state = { terminal: term };
+        this.terminal = term;
+        this.state = {
+            showPastingModal: false,
+            cols: props.cols || 80,
+            rows: props.rows || 25
+        };
     }
 
     componentDidMount() {
-        this.state.terminal.open(this.terminalRef.current);
+        this.terminal.open(this.terminalRef.current);
+        this.terminal.loadAddon(new CanvasAddon());
         this.connectChannel();
 
         if (!this.props.rows) {
@@ -148,11 +155,11 @@ export class Terminal extends React.Component {
             this.onWindowResize();
         }
         this.setTerminalTheme(this.props.theme || 'black-theme');
-        this.state.terminal.focus();
+        this.terminal.focus();
     }
 
     resizeTerminal(cols, rows) {
-        this.state.terminal.resize(cols, rows);
+        this.terminal.resize(cols, rows);
         this.props.channel.control({
             window: {
                 rows: rows,
@@ -163,7 +170,7 @@ export class Terminal extends React.Component {
 
     componentDidUpdate(prevProps, prevState) {
         if (prevProps.fontSize !== this.props.fontSize) {
-            this.state.terminal.setOption("fontSize", this.props.fontSize);
+            this.terminal.options.fontSize = this.props.fontSize;
 
             // After font size is changed, resize needs to be triggered
             const dimensions = this.calculateDimensions();
@@ -182,7 +189,7 @@ export class Terminal extends React.Component {
             this.setTerminalTheme(this.props.theme);
 
         if (prevProps.channel !== this.props.channel) {
-            this.state.terminal.reset();
+            this.terminal.reset();
             this.disconnectChannel(prevProps.channel);
             this.connectChannel();
             this.props.channel.control({
@@ -192,7 +199,7 @@ export class Terminal extends React.Component {
                 }
             });
         }
-        this.state.terminal.focus();
+        this.terminal.focus();
     }
 
     render() {
@@ -211,7 +218,7 @@ export class Terminal extends React.Component {
                     {_("Your browser does not allow paste from the context menu. You can use Shift+Insert.")}
                 </Modal>
                 <div ref={this.terminalRef}
-                     key={this.state.terminal}
+                     key={this.terminal}
                      className="console-ct"
                      onFocus={this.onFocusIn}
                      onContextMenu={this.contextMenu}
@@ -223,7 +230,7 @@ export class Terminal extends React.Component {
 
     componentWillUnmount() {
         this.disconnectChannel();
-        this.state.terminal.dispose();
+        this.terminal.dispose();
         window.removeEventListener('resize', this.onWindowResize);
         this.onFocusOut();
     }
@@ -233,7 +240,7 @@ export class Terminal extends React.Component {
             navigator.clipboard.readText()
                     .then(text => this.props.channel.send(text))
                     .catch(e => this.setState({ showPastingModal: true }))
-                    .finally(() => this.state.terminal.focus());
+                    .finally(() => this.terminal.focus());
         } catch (error) {
             this.setState({ showPastingModal: true });
         }
@@ -241,20 +248,20 @@ export class Terminal extends React.Component {
 
     getText() {
         try {
-            navigator.clipboard.writeText(this.state.terminal.getSelection())
+            navigator.clipboard.writeText(this.terminal.getSelection())
                     .catch(e => console.error('Text could not be copied, use Ctrl+Insert ', e ? e.toString() : ""))
-                    .finally(() => this.state.terminal.focus());
+                    .finally(() => this.terminal.focus());
         } catch (error) {
             console.error('Text could not be copied, use Ctrl+Insert:', error.toString());
         }
     }
 
     onChannelMessage(event, data) {
-        this.state.terminal.write(data);
+        this.terminal.write(data);
     }
 
     onChannelClose(event, options) {
-        const term = this.state.terminal;
+        const term = this.terminal;
         term.write('\x1b[31m' + (options.problem || 'disconnected') + '\x1b[m\r\n');
         term.cursorHidden = true;
         term.refresh(term.rows, term.rows);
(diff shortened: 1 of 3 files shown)


KKoukiou:
@martinpitt can you please force push on my branch the node_modules for this package.json change :) ? sorry!!.

martinpitt:
@KKoukiou : Sure! Done. FTR:
```
git fetch git@github.com:KKoukiou/cockpit.git xterm-update
git checkout FETCH_HEAD
git rebase origin/main
tools/node-modules install
git commit -a --amend
git push -f git@github.com:KKoukiou/cockpit.git HEAD:xterm-update
```

garrett:
Firstly: I'm having problems building this PR; it's this (snippet):

```
  WEBPACK  kdump
node:internal/modules/cjs/loader:408
      throw err;
      ^

Error: Cannot find module '/var/home/garrett/Project/cockpit/cockpit/node_modules/copy-webpack-plugin/dist/cjs.js'. Please verify that the package.json has a valid "main" entry
```

But, anyway, the actual thing I want to comment on: xterm.js is fragile. Other renderer have been broken in WebKit, which is used in Safari on macOS and iOS, and we also use WebKit for Cockpit client.

Please make sure that this still works in all browsers: Firefox, Chrome, and WebKit (you can test GNOME Web / Epiphany or you can use Cockpit Client).

KKoukiou:
> Firstly: I'm having problems building this PR; it's this (snippet):
> 
> ```
>   WEBPACK  kdump
> node:internal/modules/cjs/loader:408
>       throw err;
>       ^
> 
> Error: Cannot find module '/var/home/garrett/Project/cockpit/cockpit/node_modules/copy-webpack-plugin/dist/cjs.js'. Please verify that the package.json has a valid "main" entry
> ```
> 
> But, anyway, the actual thing I want to comment on: xterm.js is fragile. Other renderer have been broken in WebKit, which is used in Safari on macOS and iOS, and we also use WebKit for Cockpit client.
> 
> Please make sure that this still works in all browsers: Firefox, Chrome, and WebKit (you can test GNOME Web / Epiphany or you can use Cockpit Client).

yes it's expected. When it's ready for review I will ask you.

martinpitt:
Some observations:
 * The generated node_modules/ that I pushed here, and that [CI agrees with](https://github.com/cockpit-project/cockpit/actions/runs/3900202016/jobs/6660595736), has *zero* /dist directories, that's why it's broken.
 * OTOH, CI [disagrees](https://github.com/cockpit-project/cockpit/actions/runs/3901936637/jobs/6664384179) with [Katerina's rebase](https://github.com/cockpit-project/cockpit/compare/7a8d52bd9fb9e28222e35deddbf47af59db4b295..e0b46d5e8c97478dfc972835bd6f1c4e25299be2), but that could also just be because something else changed package.json in the meantime? I get this locally as well -- merely checking this out fails `tools/node-modules checkout`, with "package.json - differ: byte 761, line 23".
 * I reset Katerina's branch: I rebased, reverted the n_m update with `git -C node_modules/ checkout 3062b3b024b2; git commit -a --amend`, `tools/node-modules install`; this **worked**. I pushed this here.
 * `bots/npm download < package.json | tar tv | grep copy-webpack-plugin` does contain dist/. This is the most basic command which runs *totally* isolated -- the only thing that goes in is package.json via stdin. This is also true if I run it from /tmp/ with `~/upstream/bots/npm download < /tmp/package.json | tar tv | grep copy-webpack-plugin`.
 * After `podman volume rm cockpit-project-npm-cache-volume` it still works.

@KKoukiou how excactly did you build [this commit](https://github.com/cockpit-project/cockpit/commit/e0b46d5e8c97478dfc972835bd6f1c4e25299be2)? It's internally inconsistent. Can you reproduce and note down the steps starting from current main?

martinpitt:
As for the actual change here, the chromium are firefox tests are past TestTerminal, so this is looking good for more extensive manual testing. Would be especially useful to test on mobile, Safari, etc.

pkg/lib/cockpit-components-terminal.jsx
```diff
@@ -136,23 +137,25 @@ export class Terminal extends React.Component {
         if (props.onTitleChanged)
             term.onTitleChange(props.onTitleChanged);
 
-        this.state = { terminal: term };
+        this.terminal = term;
+        this.state = {};
```
martinpitt, line 141: This is at least missing `showPastingModal`. Would be good to have complete state here.

KKoukiou: Fixed.

pkg/lib/cockpit-components-terminal.jsx
```diff
@@ -223,7 +230,7 @@ export class Terminal extends React.Component {
 
     componentWillUnmount() {
         this.disconnectChannel();
-        this.state.terminal.dispose();
+        this.terminal.dispose();
```
cockpituous, line 233: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18157-20230112-171828-95ce097c-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

KKoukiou:
Some npm modules not available any more? https://github.com/cockpit-project/cockpit/actions/runs/3906373458/jobs/6674480532
Rebuilding

pkg/lib/cockpit-components-terminal.jsx
cockpituous, line 233: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18157-20230112-221043-95b93d07-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
I tested all four themes, font size change, overflowing/scrolling, and emojis on desktop with Firefox 108, Chromium 107, and current Cockpit Client. I tested Desktop, Firefox' mobile emulation, and Fennec on my Sony phone (rather old, 2015, Android 7 with Swiftkey). Issues:

 -  On my phone and firefox' mobile tap emulation, the "System" menu keeps popping out whenever I click anywhere on the actual page frame. This is hugely annoying, as it makes using a page pretty much impossible. This is unrelated to this PR though, and happens on main.
 - Terminal page on my phone is still [completely unusable](https://github.com/cockpit-project/cockpit/issues/18138#issuecomment-1374700274) , but again not a regression from this PR. It behaves exactly as bad as before wrt. input, but rendering is fine.
 - Firefox' mobile emulation uses a ludicrously huge font by default. This doesn't happen on an actual phone, though:
Screen Shot 2023-01-13 at 07 54 40

   This is a regression -- on main it looks much more reasonable:
Screen Shot 2023-01-13 at 08 08 22

 - My ❱❱❱ prompt works fine in Firefox:
   image
   ... but not in Chromium, regardless of which theme I pick:
prompt-chromium. Other emojis like 👍 work when I type them as Ctrl+Shift+U 1f44d. Trying to type one through wofimoji sets bash into edit mode (it seems to send an Escape or so). These work fine in firefox. This is again unrelated to this PR, already happens on main.

So there's just one regression: font size in Firefox mobile mode.

@garrett , would you be able to test Safari?

garrett:
**Safari on iPhone**: Terminal works, but it is immediately covered up by the navigation when tapping on the terminal. (So it's unusable, even if it technically works.) I don't think it's related to this PR; I'll test it further to be sure. I wonder if it's like this on Android phones now too? (It's definitely a regression somewhere.)

**Safari on macOS Monterey 12.6.2**: Terminal does not work. It's blank and only has a block cursor. However, as Metal (the macOS 3d acceleration, similar to OpenGL+Vulkan, but Apple-only) accelerates hardware and there are a few glitches in the VM, it could be related to the VM. The xterm.js website example terminal works fine however. We probably want to test this on actual Mac hardware somehow.

image

image


**Epiphany** stable & canary (WebKit on Linux): Works as expected.

garrett:
Switching colors to "light" makes no change in most of the terminal. The outer part is yellow, as expected, for the Solarized light theme.

image

garrett:
In macOS, I typed "touch hellothere" and hit enter without seeing feedback. In a terminal, I did an ls and the file is there. So it's registering keyboard inputs... it's just not showing any updates. When reloading the page after selecting a different appearance, the background color is the expected one. It seems the main part of the console is only updating on page load.

martinpitt:
> Terminal works, but it is immediately covered up by the navigation when tapping on the terminal

@garrett: I saw the same during my testing, see [my comment above](https://github.com/cockpit-project/cockpit/pull/18157#issuecomment-1381404879). It happens with every browser and even mobile emulation. This is a critical bug, but independent from this PR. I found that after a bit of fiddling and tapping, the nav menu eventually stays closed.

garrett:
During the meeting @allisonkarlitskaya mentioned she has a Mac to test things out on.

It involves:

1. Building this branch on a computer
2. Using Safari on a Mac to visit Cockpit
3. Going to the terminal page
4. Confirming if the terminal works as expected or not

garrett:
@martinpitt: I've managed to recreate the menu issue outside of the terminal as well.

2. Switch pages to something that's not cluttered (so you're not clicking on something accidentally), such as SELinux or Diagnostic Reports.

It's worth spinning out into a separate issue and fixing it there, with a high priority, as it breaks mobile functionality everywhere. (Especially but not always the terminal.) It can also break desktop browsers too, but only when resized to be small and/or using large enough fonts to trigger the issue. (Both of these things can happen when used "in the wild".)

KKoukiou:
I tested in on safari on the mac seen in the specs bellow, and everything works the same way as on my fedora laptop with firefox.
photo_2023-01-24_11-38-10

PR #18159, opened 2023-01-11
Serve po.js in the Python Bridge

jelly:
Implement serving po.js files based on the Accept-Language header. So far this only implements it for normal pages, for the shell we need to rewrite the currently used `/*/po.js` implementation into something else (follow up).

Screenshot from 2023-01-11 18-38-59

### DIFF
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -24,6 +24,7 @@ import mimetypes
 import os
 
 from pathlib import Path
+from typing import List
 from systemd_ctypes import bus
 
 from . import config
@@ -68,6 +69,30 @@ def get_libexecdir() -> str:
             LIBEXECDIR = '/nonexistent/libexec'
 
     return LIBEXECDIR
+
+
+def parse_accept_language(accept_language: str) -> List[str]:
+    """Parse the Accept-Language header
+
+    Returns an ordered list of languages.
+
+    https://tools.ietf.org/html/rfc7231#section-5.3.5
+    """
+
+    accept_languages = accept_language.split(',')
+    locales = []
+    for language in accept_languages:
+        language = language.strip()
+        locale, _, weight = language.partition(';q=')
+        weight = float(weight or 1)
+
+        # Skip possible empty locales
+        if not locale:
+            continue
+
+        locales.append((locale, weight))
+
+    return [locale for locale, _ in sorted(locales, key=lambda k: k[1], reverse=True)]
 
 
 class Package:
@@ -169,18 +194,15 @@ class Package:
         return True
 
     @staticmethod
-    def filename_variants(filename, locale):
+    def filename_variants(filename, locales):
         base, _, ext = filename.rpartition('.')
 
         while base:
-            if locale:
+            for locale in locales:
+                # po files are generated as language_Variant
+                locale = locale.replace('-', '_')
                 yield f'{base}.{locale}.{ext}'
                 yield f'{base}.{locale}.{ext}.gz'
-
-                if '_' in locale:
-                    language, _, _ = locale.partition(' ')
-                    yield f'{base}.{language}.{ext}'
-                    yield f'{base}.{language}.{ext}.gz'
 
             yield f'{base}.{ext}'
             yield f'{base}.min.{ext}'
@@ -191,10 +213,20 @@ class Package:
 
     def negotiate_file(self, path, headers):
         dirname, sep, filename = path.rpartition('/')
-        locale = headers.get('Accept-Language', '').split(',')[0].strip()  # obviously fix this
+        locales = []
+        accept_language = headers.get('Accept-Language')
+        if accept_language is not None:
+            locales = parse_accept_language(accept_language)
+            # Stripped variants come after non-stripped variants
+            for locale in list(locales):
+                if '-' in locale:
+                    language = locale.split('-')[0]
+                    # Don't append the same language
+                    if language not in locales:
+                        locales.append(language)
 
-        for variant in self.filename_variants(filename, locale):
-            logger.debug('consider variant %s', variant)
+        for variant in self.filename_variants(filename, locales):
+            logger.debug('consider variant %s for filename %s', variant, filename)
             if Path(f'{dirname}{sep}{variant}') in self.files:
                 return f'{dirname}{sep}{variant}'
 
@@ -227,22 +259,31 @@ class Package:
     def serve_file(self, path, channel):
         filename = self.negotiate_file(path, channel.headers)
 
-        if filename is None:
+        if filename:
+            with (self.path / filename).open('rb') as file:
+                data = file.read()
+        # HACK: if a translation file is missing, just return empty
+        # content. This saves a whole lot of 404s in the developer
+        # console when trying to fetch po.js for English, for example.
+        elif path.endswith('po.js'):
+            data = b''
+            filename = 'po.js'
+        else:
+            logging.error("filename=%s, path=%s", filename, path)
             logger.debug('path %s not in %s', path, self.files)
             channel.http_error(404, 'Not found')
             return
 
         content_type, encoding = mimetypes.guess_type(filename)
 
-        with (self.path / filename).open('rb') as file:
-            headers = {
-                "Access-Control-Allow-Origin": channel.origin,
-                "Content-Encoding": encoding,
-            }
-            if content_type is not None and content_type.startswith('text/html'):
-                headers['Content-Security-Policy'] = self.get_content_security_policy(channel.origin)
-            channel.http_ok(content_type, headers)
-            channel.send_data(file.read())
+        headers = {
+            "Access-Control-Allow-Origin": channel.origin,
+            "Content-Encoding": encoding,
+        }
+        if content_type is not None and content_type.startswith('text/html'):
+            headers['Content-Security-Policy'] = self.get_content_security_policy(channel.origin)
+        channel.http_ok(content_type, headers)
+        channel.send_data(data)
 
 
 # TODO: This doesn't yet implement the slighty complicated checksum
@@ -343,7 +384,10 @@ class Packages(bus.Object, interface='cockpit.Packages'):
     def serve_file(self, path, channel):
         assert path[0] == '/'
 
-        if self.checksum is not None:
+        # HACK: If the response is language specific, don't cache the file. Caching "po.js" breaks
+        # changing the language in Chromium, as that does not respect `Vary: Cookie` properly.
+        # See https://github.com/cockpit-project/cockpit/issues/8160
+        if self.checksum is not None and not path.endswith('po.js'):
             channel.push_header('X-Cockpit-Pkg-Checksum', self.checksum)
         else:
             channel.push_header('Cache-Control', 'no-cache, no-store')
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -184,7 +184,6 @@ class Package:
 
             yield f'{base}.{ext}'
             yield f'{base}.min.{ext}'
-            yield f'{base}.{ext}'
             yield f'{base}.{ext}.gz'
             yield f'{base}.min.{ext}.gz'
 
--- a/test/pytest/test_packages.py
+++ b/test/pytest/test_packages.py
@@ -1,0 +1,28 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import pytest
+
+from cockpit.packages import parse_accept_language
+
+
+@pytest.mark.parametrize("test_input,expected", [
+                         ('de-at, zh-CH, en,', ['de-at', 'zh-CH', 'en']),
+                         ('es-es, nl;q=0.8, fr;q=0.9', ['es-es', 'fr', 'nl']),
+                         ('fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5', ['fr-CH', 'fr', 'en', 'de', '*'])])
+def test_parse_accept_language(test_input, expected):
+    assert parse_accept_language(test_input) == expected


jelly:
The shell is not translated as the translations are loaded through globbing "/*/po.js" which is something we might not want to support at all. Leaving this for a follow up: https://issues.redhat.com/browse/COCKPIT-852

allisonkarlitskaya:
Answers to your questions:
 - the yielding of the variants also covers gzipped vs. non-gzipped, but I agree that trying locale variants for anything not called `po.js` is silly.  On the other hand, it's not slow either: we keep a hash table of the files that are present, so it's not like we're doing 1000 syscalls.
 - as for unit testing, I see no problem with reaching directly into the code to directly call the method, even without an instance.
 - about globbing: I'd like to avoid that, ya.  This is something we need to figure out a better approach for, and is sort of the motivating factor in my desire to rethink how we approach translation.  Long story short: we should probably serve this information as part of manifests.json.  That might look like including the translations for the relevant strings directly into manifest.json of each package at build-time, but I'm not sure yet if we want to do that.  In short: we need to talk about this, and I agree that it's OK to leave it unaddressed for now.

jelly:
Sadly, this PR will not make any new tests pass as we check the language with `cockpit.language` which interestingly is set by `cockpit.locale` which is interestingly called in `po.js`. So as we don't have translations for the shell yet.

PS: How the hell does this globbing magic even work of loading "/po/*/po.js", does it really concatenate all po files together for all "applications"? I guess I rather not found out :ghost:

src/cockpit/packages.py
```diff
@@ -43,6 +44,29 @@ def directory_items(path):
...
+    Returns an ordered list of (language, weight) pairs.
+
+    https://tools.ietf.org/html/rfc7231#section-5.3.5
+    """
+
+    accept_languages = accept_language.split(',')
+    locales = []
+    for index, language in enumerate(accept_languages):
+        if ';' in language:
+            locale, weight = language.split(';')
```
allisonkarlitskaya, line 59: According to the spec, the weight defaults to 1 if it's not present.  Look under "[5.3.1](https://www.rfc-editor.org/rfc/rfc7231#section-5.3.1).  Quality Values"

Something like this seems sensible:

```python
locale, _, weight = accept_language.partition('?q=')
weight = float(weight or 1)
```

src/cockpit/packages.py
```diff
@@ -43,6 +44,29 @@ def directory_items(path):
...
+            locale = locale.strip()
+            weight = weight.split('=')[-1].strip()
+            locales.append((locale, weight))
+        else:
+            # Retain given locales order by assiging a fake weight
+            locales.append((language.strip(), str(1 - (index / 100))))
+
+    return list(reversed(sorted(locales, key=lambda k: k[1])))
```
allisonkarlitskaya, line 67: `sorted` takes a `reverse=` kwarg.  Your use of `reversed()` here reverses the original order, but `reverse=` would preserve it, removing the need to do the indexing trick above.

I also don't see the point of returning the tuples ­— you really just care about the ordered list.

src/cockpit/packages.py
```diff
@@ -159,12 +183,18 @@ def filename_variants(filename, locale):
...
     def negotiate_file(self, path, headers):
         dirname, sep, filename = path.rpartition('/')
-        locale = headers.get('Accept-Language', '').split(',')[0].strip()  # obviously fix this
-
-        for variant in self.filename_variants(filename, locale):
-            logger.debug('consider variant %s', variant)
-            if Path(f'{dirname}{sep}{variant}') in self.files:
-                return f'{dirname}{sep}{variant}'
+        locales = []
+        accept_language = headers.get('Accept-Language', '')
```
allisonkarlitskaya, line 187: I'm reasonably sure you'll want to omit the `''` here...

src/cockpit/packages.py
```diff
@@ -159,12 +183,18 @@ def filename_variants(filename, locale):
...
+        if accept_language:
```
allisonkarlitskaya, line 188: ...and change this to `accept_language is not None`.  No?  Or maybe even one better: 

```python
if isinstance(accept_language, str):
```

src/cockpit/packages.py
allisonkarlitskaya: This function could return empty strings as locales.  It might be worth trying to defend against that.

src/cockpit/packages.py
```diff
@@ -195,6 +225,17 @@ def get_content_security_policy(self, origin):
...
 
+        # HACK: if a translation file is missing, just return empty
+        # content. This saves a whole lot of 404s in the developer
+        # console when trying to fetch po.js for English, for example.
+        if filename is None and path.endswith('po.js'):
+            headers = {
+                "Access-Control-Allow-Origin": channel.origin,
+            }
+            channel.http_ok("application/javascript", headers)
+            channel.send_data(b"")
```
allisonkarlitskaya, line 236: Is that necessary?

src/cockpit/packages.py
```diff
@@ -311,7 +352,10 @@ def serve_checksum(self, channel):
     def serve_file(self, path, channel):
         assert path[0] == '/'
 
-        if self.checksum is not None:
+        # HACK: If the response is language specific, don't cache the file. Caching "po.js" breaks
```
allisonkarlitskaya, line 355: I guess we should also do this for the missing-file (ie: English) case above, no?  Otherwise changing to/from English might be broken.

test/pytest/test_packages.py
```diff
@@ -0,0 +1,32 @@
...
+import pytest
+
+from cockpit.packages import parse_accept_language
+
+
+TEST_ACCEPT_LANGUAGES = [
+    ('de-at, zh-CH, en', [('de-at', '1.0'), ('zh-CH', '0.99'), ('en', '0.98')]),
+    ('es-es, nl;q=0.8, fr;q=0.9', [('es-es', '1.0',), ('fr', '0.9'), ('nl', '0.8')]),
+    ('fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5', [('fr-CH', '1.0'), ('fr', '0.9'), ('en', '0.8'), ('de', '0.7'), ('*', '0.5')])
+]
```
allisonkarlitskaya, line 27: This seems less like "useful global constant" and more like data specific to a particular test case.  It belongs inside of that case, I'd say.

test/pytest/test_packages.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.parametrize("test_input,expected", TEST_ACCEPT_LANGUAGES)
```
allisonkarlitskaya, line 30: ...and I feel like this is probably over-parameterized.  Let's have @martinpitt referee on this one.

allisonkarlitskaya (changes requested):
Thanks for taking care of this.  In general, I think it's time we tried to move towards a slightly more defensive style of programming in the bridge, so see the comments below about checking some error cases.

martinpitt:
> PS: How the hell does this globbing magic even work of loading "/po/*/po.js", does it really concatenate all po files together for all "applications"? I guess I rather not found out

Right, that's exactly the reason why we want to get rid of it and not support globbing in the new bridge :grin:

src/cockpit/packages.py
```diff
@@ -146,7 +170,7 @@ def filename_variants(filename, locale):
                 yield f'{base}.{locale}.{ext}.gz'
 
                 if '_' in locale:
-                    language, _, _ = locale.partition(' ')
+                    language, _, _ = locale.partition('_')
```
martinpitt, line 173: Is that really right? It looks correct for most locales like `de`, but we need to retain the country for at least en, nb, pt, and zh. In general, we should prefer `ll_CC.po` if we have it, and if not fall back to to `ll.po`.

Also, personally I prefer `language = locale.split('_')[0]`, but no strong opinion.

test/pytest/test_packages.py
martinpitt: Not sure what you mean here, but three scenarios feels quite okay, and they do cover a lot of corner cases.

src/cockpit/packages.py
allisonkarlitskaya: Look up to the line just before the `if`.  We check the full locale name first.  The stripped down version comes next.

martinpitt: I see, thanks! So @jelly , please feel free to resolve this or replace the partition with a simpler split()[0].

src/cockpit/packages.py
jelly: This needs to be discussed, it seemed irrelevant as a default make generates translations. So maybe this is for cockpit-podman? I rather retain the behaviour and later rip it out.

src/cockpit/packages.py
jelly: Hmm good point! Btw noticed that almost everything is served with no-cache now.

test/pytest/test_packages.py
jelly: I can move it to parametrize() if you don't want the constant. Basically this is a pytest convenience feature

src/cockpit/packages.py
jelly: Brrr, I rather have None then.

src/cockpit/packages.py
jelly: Right that's the else case.

src/cockpit/packages.py
jelly: * Sure, can drop the tuples
* empty strings as locales, hmm requires a test?

src/cockpit/packages.py
allisonkarlitskaya: I meant only the one line — sending the empty data.  I think that's a no-op and can be excluded — but see above.

src/cockpit/packages.py
```diff
@@ -195,6 +226,18 @@ def get_content_security_policy(self, origin):
     def serve_file(self, path, channel):
         filename = self.negotiate_file(path, channel.headers)
 
+        # HACK: if a translation file is missing, just return empty
+        # content. This saves a whole lot of 404s in the developer
+        # console when trying to fetch po.js for English, for example.
+        if filename is None and path.endswith('po.js'):
+            headers = {
+                'Access-Control-Allow-Origin': channel.origin,
+                'Cache-Control': 'no-cache, no-store',
```
allisonkarlitskaya, line 235: Not a huge fan of the duplication here.

allisonkarlitskaya: In general I feel like it would be better to handle this case by using the same serving code as you use below, but figuring out a way to serve the empty string instead of the contents of a file.  Something like the moral equivalent of `filename = '/dev/null'`, but without actually reading the file.

src/cockpit/packages.py
```diff
@@ -159,12 +184,18 @@ def filename_variants(filename, locale):
...
-        for variant in self.filename_variants(filename, locale):
-            logger.debug('consider variant %s', variant)
-            if Path(f'{dirname}{sep}{variant}') in self.files:
-                return f'{dirname}{sep}{variant}'
+        locales = []
+        accept_language = headers.get('Accept-Language')
+        if accept_language is not None:
+            locales = parse_accept_language(accept_language)
+
+        for locale in locales:
```
allisonkarlitskaya, line 192: This logic is definitely some kind of wrong.

 - calling the variant function multiple times means that you consider the untranslated filename over and over again
 - doing a separate scan for each page means that we can end up with a mix of languages

The C bridge tries to set the language to the first supported locale it finds, then ignores all of the others.  I wonder if we should do something like that.

If not, and you really want to check multiple languages, then I'd rather modify the filename_variants function to take a list of locales and iterate over them internally before trying the untranslated variants.

allisonkarlitskaya (changes requested):

src/cockpit/packages.py
```diff
@@ -24,6 +24,7 @@
 import os
 
 from pathlib import Path
+from typing import List, Tuple
```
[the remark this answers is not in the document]
jelly: I should have a pre-push hook...

src/cockpit/packages.py
```diff
@@ -43,6 +44,30 @@ def directory_items(path):
...
+        weight = float(weight or 1)
+        locale = locale.strip()
+
+        # Skip possible empty locales
+        if not locale:
+            continue
+
+        locales.append((locale, weight))
+
+    return [loc[0] for loc in (sorted(locales, key=lambda k: k[1], reverse=True))]
```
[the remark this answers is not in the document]
jelly: And no I prefer _ over _weight

src/cockpit/packages.py
jelly: I did a thing to reduce duplucation.

allisonkarlitskaya (changes requested):
See comments below.

One more extremely annoying point (and I think C bridge also gets this wrong): according to RFC[^1] 

> Language tags and thus language ranges are to be treated as case-insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning.  Matching of language tags to language ranges MUST be done in a case-insensitive manner.

So it seems that if the user gives us `pt-br` or `EN`, we ought to still be able to find the files on our case-sensitive filesystem...  Ugh.

[^1]: https://www.rfc-editor.org/rfc/rfc4647#section-2

src/cockpit/packages.py
```diff
@@ -191,10 +213,20 @@ def filename_variants(filename, locale):
...
-        locale = headers.get('Accept-Language', '').split(',')[0].strip()  # obviously fix this
-
-        for variant in self.filename_variants(filename, locale):
-            logger.debug('consider variant %s', variant)
+        locales = []
+        accept_language = headers.get('Accept-Language')
+        if accept_language is not None:
+            locales = parse_accept_language(accept_language)
+            # Append variants
+            for locale in locales:
```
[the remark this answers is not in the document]
allisonkarlitskaya: I was curious, so I wrote that:

```python
seq = [0, 1]  
last = None  
for num in seq:  
    print(num)   
    if last is not None:  
        seq.append(num+last)  
    last = num                
```

and indeed, it loops forever.  Please fix this. :)

src/cockpit/packages.py
```diff
@@ -70,6 +71,30 @@ def get_libexecdir() -> str:
...
+
+    accept_languages = accept_language.split(',')
+    locales = []
+    for language in accept_languages:
+        language = language.strip()
+        locale, _, weight = language.partition(';q=')
+        weight = float(weight or 1)
+
+        # Skip possible empty locales
+        if not locale:
```
allisonkarlitskaya, line 90: :+1:  Thanks for this addition.

src/cockpit/packages.py
```diff
@@ -227,22 +259,31 @@ def get_content_security_policy(self, origin):
     def serve_file(self, path, channel):
         filename = self.negotiate_file(path, channel.headers)
 
-        if filename is None:
+        if filename:
```
allisonkarlitskaya, line 262: `... is not None` lol

I don't think it matters in this case, because there's no way we found an empty filename as a variant, but seriously :sweat_smile:

src/cockpit/packages.py
```diff
@@ -227,22 +259,31 @@ def get_content_security_policy(self, origin):
...
+            with (self.path / filename).open('rb') as file:
+                data = file.read()
+        # HACK: if a translation file is missing, just return empty
+        # content. This saves a whole lot of 404s in the developer
+        # console when trying to fetch po.js for English, for example.
+        elif path.endswith('po.js'):
+            data = b''
+            filename = 'po.js'
+        else:
+            logging.error("filename=%s, path=%s", filename, path)
```
allisonkarlitskaya, line 272: This arrangement is *so* much nicer.  Thanks!

src/cockpit/packages.py
```diff
@@ -227,22 +259,31 @@ def get_content_security_policy(self, origin):
     def serve_file(self, path, channel):
         filename = self.negotiate_file(path, channel.headers)
 
-        if filename is None:
+        if filename:
+            with (self.path / filename).open('rb') as file:
+                data = file.read()
+        # HACK: if a translation file is missing, just return empty
```
allisonkarlitskaya, line 265: I think it reads better if this is inside the elif block, but it's a matter of taste.

allisonkarlitskaya (approved):
This looks good now, thanks :)

PR #18155, opened 2023-01-11
packagekit: Avoid redundant property GetAll() call on update transaction

martinpitt:
This has always been racy due to the short 0.5s timeout. This often timed out    
with the Python bridge, as PackageKit isn't the fastest thing on earth. Butit's    
also not necessary at all: We already get told about property updates via the    
`notify` handler.    
    
So let's rework this to make the whole structure a lot simpler: Turn    
ApplyUpdates into a pure functional component, and drop its second    
watchTransaction() for packages. Let OsUpdates' watchTransaction() pick up    
`Packages` signals and all property values. This works because the initial    
`notify` event always includes the full set of properties.    
    
Also improve the initialization: At the beginning of an update we usually get    
`Percentage` properties from PackageKit before the first `Package` signal. If    
that happens, move from the "Initializing" empty state to the progress bar    
already. That provides better visual feedback, and it's also what    
TestUpdates.testBasic expects. With that, `lastAction` can now be `undefined`,    
so handle that in the JSX. 

---

This should fix *all* check-packagekit tests in the pybridge scenario.

PR #18163, opened 2023-01-12
python: Fix D-Bus timeout

martinpitt:
sd_bus_call()'s timeout is in microseconds, while our API is in milliseconds. Convert the timeout appropriately.

[1] https://www.freedesktop.org/software/systemd/man/sd_bus_call.html

----

This should "accidentally" fix most PackageKit tests, [letting this run once](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18163-20230112-111112-e9b5d701-fedora-37-pybridge/log.html) . But PR #18155 fixes them in a much more elegant way, so I'll wait for that to land to avoid collisions on removing `@todoPybridge`.

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -289,10 +289,17 @@ class DBusChannel(Channel):
 
     async def do_call(self, message):
         path, iface, method, args = message['call']
-        timeout = message.get('timeout', 2**64 - 1)
         cookie = message.get('id')
         flags = message.get('flags')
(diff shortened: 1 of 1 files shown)


mvollmer (approved):
Quite verbose, but that's good!

martinpitt:
@mvollmer Agreed about verbose. A more succinct way would be

    timeout = timeout * 1000 if timeout is not None else 2 ** 64 - 1

but that's not exactly sparkling with readability either IMHO..

PR #18164, opened 2023-01-12
python: allow filtering of log messages on module names

jelly:
In the Python bridge we setup a logger per module such as "cockpit.packages". By setting the COCKPIT_DEBUG environment variable we can now specify which modules we want to see logs for (comma separate values). If you want to see logging for all modules set COCKPIT_DEBUG to "all".

Setting COCKPIT_DEBUG also sets the log level to debug.

### DIFF
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -154,6 +154,23 @@ def try_to_receive_stderr():
             os.close(stderr_fd)
 
 
+def setup_logging(debug: bool):
+    """Setup our logger with optional filtering of modules if COCKPIT_DEBUG env is set"""
+
+    modules = os.getenv('COCKPIT_DEBUG', '')
+    logging.basicConfig(format='%(name)s-%(levelname)s: %(message)s')
+
+    if debug or modules == 'all':
+        logging.getLogger().setLevel(level=logging.DEBUG)
+    elif modules:
+        for module in modules.split(','):
+            module = module.strip()
+            if not module:
+                continue
+
+            logging.getLogger(module).setLevel(logging.DEBUG)
+
+
 def main() -> None:
     # The --privileged bridge gets spawned with its stderr being consumed by a
     # pipe used for reading authentication-related message from sudo.  The
@@ -171,8 +188,7 @@ def main() -> None:
     parser.add_argument('--version', action='store_true', help='Show Cockpit version information')
     args = parser.parse_args()
 
-    if args.debug:
-        logging.basicConfig(level=logging.DEBUG)
+    setup_logging(args.debug)
 
     if args.packages:
         Packages().show()
--- a/src/cockpit/protocol.py
+++ b/src/cockpit/protocol.py
@@ -74,7 +74,7 @@ class CockpitProtocol(asyncio.Protocol):
                 logger.debug('channel control received %s', message)
                 self.channel_control_received(channel, command, message)
             else:
-                logging.debug('transport control received %s', message)
+                logger.debug('transport control received %s', message)
                 self.transport_control_received(command, message)
 
     def consume_one_frame(self, view):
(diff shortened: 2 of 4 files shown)


src/cockpit/bridge.py
```diff
@@ -154,6 +154,26 @@ def try_to_receive_stderr():
...
+            return record
+
+    if debug or modules:
+        cfg['level'] = logging.DEBUG
+    if modules:
+        handler = logging.StreamHandler()
+        handler.addFilter(filter_record)
+        cfg['handlers'] = [handler]
+
+    logging.basicConfig(**cfg)
```
jelly, line 174: I'm trying not to hardcode our defaults, but just realised we can always initialise logging.StreamHandler() imo.

allisonkarlitskaya (commented):
Could we do this without the filter?  For the `--debug` or `all` case we could just keep the existing `.basicConfig()` call, and for the case of listing off the domains, we could `getLogger()` on those domains and call `.setLevel()`, no?

src/cockpit/bridge.py
```diff
@@ -154,6 +154,26 @@ def try_to_receive_stderr():
             os.close(stderr_fd)
 
 
+def setup_logging(debug: bool):
+    """Setup our logger with optional filtering of modules if COCKPIT_DEBUG env is set"""
+
+    cfg = {}
+    modules = set(os.getenv('COCKPIT_DEBUG', '').split(','))
```
martinpitt, line 161: You should filter out empty names here, as otherwise trailing or doubled commas would show all messages.

src/cockpit/bridge.py
```diff
@@ -154,6 +154,26 @@
...
+
+    def filter_record(record: logging.LogRecord):
```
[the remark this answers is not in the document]
martinpitt: Yes, please fix this with `return None` or `return False` or whatever the API demands.

src/cockpit/bridge.py
```diff
@@ -157,21 +157,18 @@ def try_to_receive_stderr():
...
-        handler.addFilter(filter_record)
-        cfg['handlers'] = [handler]
-
-    logging.basicConfig(**cfg)
+    modules = os.getenv('COCKPIT_DEBUG', '')
+    if debug or modules == 'all':
+        logging.basicConfig(level=logging.DEBUG)
+    elif modules:
+        # HACK: Required to make loggers have a handler
+        logging.basicConfig(format='%(name)s-%(levelname)s: %(message)s')
```
allisonkarlitskaya, line 165: I think I'd prefer to see the format set in any case, even if debugging is not enabled.

Then for the "all" case I think that could be better handled with `logging.root.setLevellogging.DEBUG)` to make it more symmetric with the per-domain case.

allisonkarlitskaya (changes requested):

src/cockpit/bridge.py
```diff
@@ -154,6 +154,23 @@ def try_to_receive_stderr():
...
 
 
+def setup_logging(debug: bool):
+    """Setup our logger with optional filtering of modules if COCKPIT_DEBUG env is set"""
+
+    modules = os.getenv('COCKPIT_DEBUG', '')
+    logging.basicConfig(format='%(name)s-%(levelname)s: %(message)s')
+
+    if debug or modules == 'all':
+        logging.basicConfig(level=logging.DEBUG)
```
allisonkarlitskaya: Any reason why not `logging.root.setLevel`?

jelly: Fixed.

src/cockpit/bridge.py
```diff
@@ -161,7 +161,7 @@ def setup_logging(debug: bool):
     logging.basicConfig(format='%(name)s-%(levelname)s: %(message)s')
 
     if debug or modules == 'all':
-        logging.basicConfig(level=logging.DEBUG)
+        logging.getLogger().setLevel(logging.DEBUG)
```
allisonkarlitskaya, line 164: This landed in the wrong commit.

allisonkarlitskaya (changes requested):

allisonkarlitskaya (approved):
Thanks

allisonkarlitskaya (approved):
Thanks!

PR #18168, opened 2023-01-13
python: Add temporary ${libexecdir} expansion in manifests

martinpitt:
This provides the moral equivalent of what commit 94394128e8b6 did for    
the C bridge. However, we want to avoid parameterizing the py bridge    
with ./configure arguments, to keep it buildable with pip. Detect the    
path at runtime instead, there are only a few plausible options anyway.    
This is a temporary hack until we eliminate all `${libexecdir}` from    
manifests.    
    
This fixes joining an IPA domain, as it now finds    
cockpit-certificate-helper. TestIPA.testQualifiedUsers now gets much    
further, unfortunately it still fails on the remote host part of the    
test, as cockpit-ssh does not work yet.    
    
It also fixes the cockpit-{ssh,pcp,askpass} paths in the manifests, but as we    
don't do that routing yet, it does not fix any of the ssh/pcp tests yet.    
But it is a necessary prerequisite for actually implementing these.    

-----

Unfortunately the visible effect to the pybridge scenario is underwhelming, but this still feels like a stepping stone for implementing further functionality.

### DIFF
--- a/pkg/systemd/overview-cards/realmd.jsx
+++ b/pkg/systemd/overview-cards/realmd.jsx
@@ -177,6 +177,7 @@ export class RealmdClient {
                     const proc = cockpit.spawn([helper, "ipa", "request", kerberos.RealmName, user],
                                                { superuser: "require", err: "message" });
                     proc.input(password);
+                    proc.catch(ex => console.warn("Failed to run", helper, "ipa request:", ex.toString()));
                     return proc;
                 })
                 .catch(() => true); // no Kerberos domain? nevermind then
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -43,12 +43,44 @@ def directory_items(path):
     return sorted(path.iterdir(), key=lambda item: item.name)
 
 
+# HACK: We eventually want to get rid of all ${libexecdir} in manifests:
+# - rewrite cockpit-{ssh,pcp} in Python and import them as module instead of exec'ing
+# - rewrite cockpit-askpass in Python and write it into a temporary file
+# - bundle helper shell scripts into their webpacks
+# Until then, we need this libexecdir detection hack.
+LIBEXECDIR = None
+
+
+def get_libexecdir() -> str:
+    '''Detect libexecdir on current machine
+
+    This only works for systems which have cockpit-ws installed.
+    '''
+    global LIBEXECDIR
+    if LIBEXECDIR is None:
+        for candidate in ['/usr/local/libexec', '/usr/libexec', '/usr/local/lib/cockpit', '/usr/lib/cockpit']:
+            if os.path.exists(os.path.join(candidate, 'cockpit-certificate-helper')):
+                LIBEXECDIR = candidate
+                break
+        else:
+            logger.warning('Could not detect libexecdir')
+            # give readable error messages
+            LIBEXECDIR = '/nonexistent/libexec'
+
+    return LIBEXECDIR
+
+
 class Package:
     def __init__(self, path):
         self.path = path
 
         with (self.path / 'manifest.json').open(encoding='utf-8') as manifest_file:
-            self.manifest = json.load(manifest_file)
+            manifest = manifest_file.read()
+
+        # HACK: drop this after getting rid of ${libexecdir}, see above
+        manifest = manifest.replace('${libexecdir}', get_libexecdir())
+
+        self.manifest = json.loads(manifest)
 
         self.try_override(self.path / 'override.json')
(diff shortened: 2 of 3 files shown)


pkg/systemd/overview-cards/realmd.jsx
```diff
@@ -177,6 +177,7 @@ export class RealmdClient {
                     const proc = cockpit.spawn([helper, "ipa", "request", kerberos.RealmName, user],
                                                { superuser: "require", err: "message" });
                     proc.input(password);
+                    proc.catch(ex => console.warn("Failed to run", helper, "ipa request:", ex.toString()));
```
cockpituous, line 180: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18168-20230113-062259-d168e9f7-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt: I sure hope so :wink:

src/cockpit/packages.py
```diff
@@ -43,12 +44,32 @@ def directory_items(path):
...
 
 
+def get_libexecdir() -> str:
+    '''Detect libexecdir on current machine
+
+    This only works for systems which have cockpit-ws installed.
+    '''
+    global LIBEXECDIR
+    if LIBEXECDIR is None:
+        for candidate in ['/usr/local/libexec', '/usr/libexec', '/usr/local/lib/cockpit', '/usr/lib/cockpit']:
```
allisonkarlitskaya, line 54: I'm not a fan.

martinpitt:
Notes from meeting: It seems we can eventually get rid of most `${libexecdir}` usages in manifests:
 * `cockpit-askpass` can be rewritten in Python, sent/copied into /tmp/, and ran from there. (there might be `noexec`, but we could also use `$XDG_RUNTIME_DIR` perhaps?)
 * `cockpit-ssh` is going to be absorbed by the pybridge itself, or runtime-importing a module ([ferny](https://github.com/allisonkarlitskaya/ferny)?)
 * `cockpit-certificate-helper` needs to stay as a file for `cockpit.socket`, but we *could* duplicate it at build time and bundle it into the bridge (Mechanism TBD). Then we could expose it to JS as `cockpit.data["cockpit-certificate-helper"]`

martinpitt:
Summary: Keep this as a provisional thing, and add a spicy comment to it. Send a separate PR that bundles the helper into the webpack, and eliminate that instance of `${libexecdir}` from systemd.manifest.

src/cockpit/packages.py
```diff
@@ -43,12 +43,44 @@ def directory_items(path):
...
 class Package:
     def __init__(self, path):
         self.path = path
 
         with (self.path / 'manifest.json').open(encoding='utf-8') as manifest_file:
-            self.manifest = json.load(manifest_file)
+            manifest = manifest_file.read()
+
+        # HACK: drop this after getting rid of ${libexecdir}, see above
+        manifest = manifest.replace('${libexecdir}', get_libexecdir())
```
allisonkarlitskaya, line 81: This is awful :rofl: 

I don't care :rofl: 

(seriously, though: doing the substitution at this point, before parsing, is deeply weird.  Theoretically, libexecdir needs to be escaped/quoted, but we don't even know what kind of quotes contain this variable expansion, so we don't know how we should escape it.  But: I don't think it's ever going to matter.  Let's not overdo it for something temporary.)

allisonkarlitskaya (approved):
I have to admit I don't like how you deal with the global variable here, but it's also just provisional, so whatever :shrug: 

I originally thought we'd do this from `bridge.py` and set a `libexec` variable on `Bridge` from `__init__` but to be honest I can't think of a single place where we'd use that.  That means that putting it in packages and using it the way you do is quite a bit nicer, indeed.

pkg/systemd/overview-cards/realmd.jsx
cockpituous, line 180: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18168-20230113-085018-b4a747a8-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

src/cockpit/packages.py
martinpitt: Actually, in this case we can be sure of the quoting, as it can only have one of our possible values.

It's *much* easier to do the substitution on the string layer, as otherwise we'd have to iterate through the whole manifest. And as you say, it doesn't matter much.

PR #18172, opened 2023-01-13
test: Rewrite chromium-cdp-driver.js using async/await

martinpitt:
This makes the main function much less nested and simplifies the error handling. Also move to the readline API [1] instead of doing the buffering ourselves.
    
[1] https://nodejs.org/api/readline.html

### DIFF
--- a/test/common/chromium-cdp-driver.js
+++ b/test/common/chromium-cdp-driver.js
@@ -28,6 +28,7 @@
  * frame/execution context debugging.
  */
 
+const readline = require('node:readline/promises');
 const CDP = require('chrome-remote-interface');
 
 let enable_debug = false;
@@ -40,11 +41,6 @@ function debug(msg) {
 /**
  * Format response to the client
  */
-
-function fatal() {
-    console.error.apply(console.error, arguments);
-    process.exit(1);
-}
 
 function fail(err) {
     if (typeof err === 'undefined')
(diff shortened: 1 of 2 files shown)


test/common/chromium-cdp-driver.js
```diff
@@ -279,60 +274,63 @@ function setupSSLCertHandling(client) {
...
+    const target = await CDP.New(options);
+    target.port = options.port;
+    const client = await CDP({ target });
+    setupLogging(client);
+    setupFrameTracking(client);
+    setupSSLCertHandling(client);
+    setupLocalFunctions(client);
+
+    let input_buf = '';
+    process.stdin
```
allisonkarlitskaya, line 301: Couldn't we do something more like `await process.stdin.read()` in a loop or so?

allisonkarlitskaya: Apparently that looks like `for await (const chunk of process.stdin) {`

test/common/chromium-cdp-driver.js
```diff
@@ -279,60 +274,59 @@ function setupSSLCertHandling(client) {
...
+    const target = await CDP.New(options);
+    target.port = options.port;
+    const client = await CDP({ target });
+    setupLogging(client);
+    setupFrameTracking(client);
+    setupSSLCertHandling(client);
+    setupLocalFunctions(client);
+
+    let input_buf = '';
+    for await (const chunk of process.stdin) {
```
martinpitt, line 301: @allisonkarlitskaya brought up the idea of a line iterator, which would get rid of that buffer processing. Need to check if that's possible.

allisonkarlitskaya: https://nodejs.org/api/readline.html

martinpitt: Oh yeah, much nicer!

martinpitt:
For some reason, the `eval(command).then(reply =>` refuses to be rewritten with `async` like in the Chromium driver. I didn't dive into this much, but at least rewriting main and the input loop with async works fine. But let's give this a spin.

martinpitt:
Hmm, the TF tests [fail](https://artifacts.dev.testing-farm.io/b81e4bc3-cba5-4ca9-9bde-610df212a1d4/) with

    Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:readline/promises

on CS8/9 and F36. F37 and rawhide are okay. Indeed that promises/async API only exists in Node 18, while [16.x](https://nodejs.org/docs/latest-v16.x/api/readline.html) only has the simpler callback interface (what we used before). That's a shame. OTOH that API isn't too bad either, I'll see how it looks like.

martinpitt:
No joy --  This works on F36 and C8s now, but [C8s](https://artifacts.dev.testing-farm.io/c9a20643-8fd3-4f5a-ad2a-13badc067f15/) apparently has an even older nodejs (unfortunately I don't see the version in the log).

   Error: Cannot find module 'node:readline'

```
podman run -it --rm quay.io/centos/centos:stream8
dnf install -y nodejs
```

... installs 1:10.23.1-1.module_el8.4.0+645+9ce14ba2, which is ancient. Curiously, in interactive `node` I can do `readline.createInterface(process.stdin)` and it works without a `require`. But when I drop it from firefox-cdp-driver.js, I get `ReferenceError: readline is not defined`. So apparently an older node needs a different import? [10.x docs](https://nodejs.org/docs/latest-v10.x/api/readline.html) say `require('readline')`, and that seems to work for 18.x as well. Let's see!

martinpitt:
Hmm, I give up on firefox. The old events-based API isn't very compatible with `await`, as it's hard to properly pass on exceptions. Every test currently has [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18172-20230121-211625-949131bb-fedora-37-firefox/log):
```
Error: <html>                    <head><title>404 Not Found</title></head>                    <body>                      <h1>404 Not Found</h1>                      <p>                        <span style='font-family: monospace;'>&#47;&#106;&#115;&#111;&#110;&#47;&#99;&#108;&#111;&#115;&#101;&#47;&#117;&#110;&#100;&#101;&#102;&#105;&#110;&#101;&#100;</span> was not found.                      </p>                    </body>                  </html>
    at IncomingMessage.<anonymous> (/var/home/martin/upstream/cockpit/async/node_modules/chrome-remote-interface/lib/external-request.js:33:26)
    at IncomingMessage.emit (node:events:525:35)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
```

Let's revisit that in a few years when we have node 0.16 everywhere.

martinpitt:
Ah, but let's at least use readline, that already looks a tad nicer.

martinpitt:
[f37/devel](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18172-20230122-103906-69df97bb-fedora-37-devel/log.html) got stuck on the very last test, argh! Retrying. Otherwise all green. This is about as much effort as I want to throw at this.

allisonkarlitskaya (commented):
No async iterator on readline?  :(

real review after I wake the kids up and get them ready :)

martinpitt:
> No async iterator on readline? :(

WDYM? 

    for await (const command of readline.createInterface(process.stdin)) {

is about as async/fancy as it gets :grin: However, we can't use that in the Firefox driver yet, see my odyssey above.

martinpitt:
@marusak , @allisonkarlitskaya : Friendly review ping?

martinpitt:
Heh, makes sense :grin:

martinpitt:
@mvollmer, @marusak : Friendly review ping

mvollmer (approved):
Thanks!  Even this old donkey starts to like async/await.

PR #18173, opened 2023-01-13
selinux: Allow cockpit-session to write user's home

martinpitt:
pam_google_authenticator creates a new ~/.google_authenticator.TMPSUFFIX and renames it.

https://bugzilla.redhat.com/show_bug.cgi?id=2157901

----

See https://bugzilla.redhat.com/show_bug.cgi?id=2157901#c6 for how I reproduced and tested this. We don't have the google-authenticator package on our images, and the OTP token computation also is quite complicated to automate.

### DIFF
--- a/selinux/cockpit.te
+++ b/selinux/cockpit.te
@@ -172,6 +172,9 @@ userdom_spec_domtrans_all_users(cockpit_session_t)
 userdom_noatsecure_login_userdomain(cockpit_session_t)
 usermanage_read_crack_db(cockpit_session_t)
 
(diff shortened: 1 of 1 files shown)


jelly (approved):
Would have been nicer if pam_google_authenticator created the required temp file in /tmp

martinpitt:
@jelly : That's not possible for atomic replacement, it must be in the same directory. Otherwise $HOME and /tmp (or any other place) could be on different mounts, then you can't `rename()`.

PR #18177, opened 2023-01-16
python: Add support for freezing endpoints

allisonkarlitskaya:
This is my best guess about why `TestLongRunning.testBasic` is failing.  Let's see if it helps.

### DIFF
--- a/src/cockpit/router.py
+++ b/src/cockpit/router.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import collections
 import logging
 
 from typing import Dict, List, Optional
@@ -26,11 +27,57 @@ from .protocol import CockpitProtocolServer, CockpitProtocolError
 logger = logging.getLogger(__name__)
 
 
+class ExecutionQueue:
+    """Temporarily delay calls to a given set of class methods.
+
+    Functions by replacing the named function at the instance __dict__
+    level, effectively providing an override for exactly one instance
+    of `method`'s object.
+    Queues the invocations.  Run them later with .run(), which also reverses
+    the redirection by deleting the named methods from the instance.
+    """
+    def __init__(self, methods):
+        self.queue = collections.deque()
+        self.methods = methods
+
+        for method in self.methods:
+            self._wrap(method)
+
+    def _wrap(self, method):
+        # NB: this function is stored in the instance dict and therefore
+        # doesn't function as a descriptor, isn't a method, doesn't get bound,
+        # and therefore doesn't receive a self parameter
+        setattr(method.__self__, method.__func__.__name__, lambda *args: self.queue.append((method, args)))
+
+    def run(self):
+        logger.debug('ExecutionQueue: Running %d queued method calls', len(self.queue))
+        for method, args in self.queue:
+            method(*args)
+
+        for method in self.methods:
+            delattr(method.__self__, method.__func__.__name__)
+
+
 class Endpoint:
     router: Router
+    __endpoint_frozen_queue: Optional[ExecutionQueue] = None
 
     def __init__(self, router: Router):
         self.router = router
+
+    def endpoint_is_frozen(self) -> bool:
+        return self.__endpoint_frozen_queue is not None
+
+    def freeze_endpoint(self):
+        assert self.__endpoint_frozen_queue is None
+        logger.debug('Freezing endpoint %s', self)
+        self.__endpoint_frozen_queue = ExecutionQueue({self.do_channel_control, self.do_channel_data})
+
+    def thaw_endpoint(self):
+        assert self.__endpoint_frozen_queue is not None
+        logger.debug('Thawing endpoint %s', self)
+        self.__endpoint_frozen_queue.run()
+        self.__endpoint_frozen_queue = None
 
     # interface for receiving messages
     def do_channel_control(self, channel: str, command: str, message: Dict[str, object]) -> None:
--- a/src/cockpit/peer.py
+++ b/src/cockpit/peer.py
@@ -63,6 +63,9 @@ class Peer(CockpitProtocolClient, SubprocessProtocol, Endpoint):
                  init_host: Optional[str] = None):
         super().__init__(router)
 
+        # All Peers start out frozen — we only unfreeze after we see the first 'init' message
+        self.freeze_endpoint()
+
         self.name = name
         self.state_listener = state_listener
 
@@ -87,10 +90,14 @@ class Peer(CockpitProtocolClient, SubprocessProtocol, Endpoint):
             transport.send_stderr_fd()
 
     def do_init(self, message: Dict[str, Any]) -> None:
-        logger.debug('Peer %s connection got init message', self.name)
-        if self.state_listener is not None:
-            self.state_listener.peer_state_changed(self, 'init')
-        self.write_control(command='init', version='1', host=self.init_host)
+        if self.endpoint_is_frozen():
+            logger.debug('Peer %s connection got init message', self.name)
+            if self.state_listener is not None:
+                self.state_listener.peer_state_changed(self, 'init')
+            self.write_control(command='init', version='1', host=self.init_host)
+            self.thaw_endpoint()
+        else:
+            logger.warning('Peer %s connection got duplicate init message', self.name)
 
     def do_closed(self, transport_was: asyncio.Transport, exc: Optional[Exception]) -> None:
         logger.debug('Peer %s connection lost %s', self.name, exc)
--- a/src/cockpit/superuser.py
+++ b/src/cockpit/superuser.py
@@ -123,9 +123,9 @@ class SuperuserRoutingRule(PeerStateListener, RoutingRule, bus.Object, interface
         logger.debug('Peer %s state changed -> %s', peer.name, event)
         if event == 'connected':
             self.current = 'init'
+            self.peer = peer
 
         elif event == 'init':
-            self.peer = peer
             self.current = peer.name
             if self.startup is not None:
                 self.startup.success(self)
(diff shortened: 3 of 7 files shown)


src/cockpit/router.py
```diff
@@ -26,12 +27,47 @@
 logger = logging.getLogger(__name__)
 
 
+class ExecutionQueue(collections.deque):
+    def run(self):
+        for method, args in self:
+            method(*args)
```
martinpitt, line 33: If you try to call run() on a still frozen endpoint, it would re-queue all method calls again. Can this grow some assertion that the call does not lead to calling .append(), to avoid programming errors?

src/cockpit/router.py
```diff
@@ -26,12 +27,47 @@
...
+
+    def wrap(self, method):
+        """Temporarily delay calls to a given class method.
+
+        Functions by replacing the named function at the instance __dict__
+        level, effectively providing an override for exactly one instance.
```
martinpitt, line 39: That's the object (i.e. class instance) that contains the wrapped function, not the ExecutionQueue instance, right?

```suggestion
        level, effectively providing an override for exactly one instance
        of `method`'s object.
```

src/cockpit/router.py
```diff
@@ -26,12 +27,47 @@
...
 
+    def freeze_endpoint(self):
+        self.queue = ExecutionQueue()
+        self.queue.wrap(self.do_channel_control)
+        self.queue.wrap(self.do_channel_data)
+
+    def thaw_endpoint(self):
+        queue = self.queue
+
+        if queue is not None:
```
martinpitt, line 64: This should become an assertion, at least until we find a situtation where we really can't avoid an unbalanced freeze/thaw.

src/cockpit/router.py
```diff
@@ -26,12 +27,47 @@
...
+            self.queue = None
+            queue.run()
+            # remove the enqueue wrappers — will be handled according to class default now
+            del self.do_channel_control
+            del self.do_channel_data
```
martinpitt, line 69: This is super ugly and error prone -- couldn't this be done by `ExecutionQueue.__del__` or `.run()`? If you can't figure out `self` from `self.method`, then we could additionally pass `self` to the `.wrap()` call and record it, so that run() can remove the wrappers by itself.

src/cockpit/router.py
martinpitt, line 64: Why the reassignment dance?

martinpitt (changes requested):
Thanks! I copied my review from PR #18002, with some additional suggestions. But let's finish this run to see whether it actually helps. I'll also give it a spin locally.

martinpitt:
Oh [nice](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18177-20230116-094155-8fa625e8-fedora-37-pybridge/log.html#53), we may be getting two fixes for the price of one :wink: 

I ran `TestLongRunning.testBasic` locally 30 times and it always succeeded. On main, it usually takes one to 5 attempts to get it to fail.

src/cockpit/router.py
allisonkarlitskaya: Correct.  Accepting your change.

src/cockpit/router.py
allisonkarlitskaya: Because we need to call `queue.run()` after setting `self.queue = None`.

allisonkarlitskaya: I think I agree, ya.

src/cockpit/router.py
allisonkarlitskaya: Ya.  I'm not sure I agree with this.  The queue object doesn't contain any direct references to the frozen endpoint object or the methods (other than via the queued calls, of which there may be none).  We could add that, but it would be added complexity.  Also: doing things as side-effect of `__del__` is definitely evil.

You might convince me to transform the queue into a context manager which takes an object and a list of methods and which we explicitly __exit__...

martinpitt: It seems `method` is sufficient -- `method.__self__` is the object of method, I tried that in a REPL.
```
>>> class Foo:
...   def __init__(self, val):
...     self.val = val
...   def say(self):
...     print("hi", hash(self), self.val)
... 

>>> a = Foo(1)
>>> b = Foo(2)

>>> a.say()
hi 8738851316177 1
>>> b.say()
hi 8738851200769 2

>>> a
<__main__.Foo object at 0x7f2ac20e1d10>
>>> b
<__main__.Foo object at 0x7f2ac1f1f010>

>>> a.say.__self__
<__main__.Foo object at 0x7f2ac20e1d10>
>>> b.say.__self__
<__main__.Foo object at 0x7f2ac1f1f010>
```

src/cockpit/router.py
martinpitt: Right, but why do we have to?

src/cockpit/router.py
martinpitt: I thought about changing def wrap() to do something like `self.wrapped.append(method)`, and then in run() it can iterate over `self.wrapped` and call del on all of them?

martinpitt: This works. I'll let you do the context manager thing and then we can see which we like better.

```diff
--- src/cockpit/router.py
+++ src/cockpit/router.py
@@ -28,9 +28,15 @@ logger = logging.getLogger(__name__)
 
 
 class ExecutionQueue(collections.deque):
+    def __init__(self):
+        super().__init__()
+        self.wrappers = []
+
     def run(self):
         for method, args in self:
             method(*args)
+        for wrapper in self.wrappers:
+            del wrapper.__self__.__dict__[wrapper.__func__.__name__]
 
     def wrap(self, method):
         """Temporarily delay calls to a given class method.
@@ -43,6 +49,7 @@ class ExecutionQueue(collections.deque):
         # NB: this function is stored in the instance dict and therefore
         # doesn't function as a descriptor, isn't a method, doesn't get bound,
         # and therefore doesn't receive a self parameter
+        self.wrappers.append(method)
         setattr(method.__self__, method.__func__.__name__, lambda *args: self.append((method, args)))
 
 
@@ -64,9 +71,6 @@ class Endpoint:
         if queue is not None:
             self.queue = None
             queue.run()
-            # remove the enqueue wrappers — will be handled according to class default now
-            del self.do_channel_control
-            del self.do_channel_data
 
     # interface for receiving messages
     def do_channel_control(self, channel: str, command: str, message: Dict[str, object]) -> None:
```

src/cockpit/router.py
```diff
@@ -26,12 +27,54 @@
...
 
 
+class ExecutionQueue(collections.deque):
+    """Temporarily delay calls to a given set of class methods.
+
+    Functions by replacing the named function at the instance __dict__
+    level, effectively providing an override for exactly one instance
+    of `method`'s object.
+    Queues the invocations.  Run them later with .run(). Reverse this call
+    by deleting the named methods from the instance.
```
martinpitt, line 37: That last sentence is not true any more.

src/cockpit/router.py
martinpitt: This works fine, and is a little more straightforward:
```diff
--- src/cockpit/router.py
+++ src/cockpit/router.py
@@ -69,11 +69,9 @@ class Endpoint:
         self.queue = ExecutionQueue({self.do_channel_control, self.do_channel_data})
 
     def thaw_endpoint(self):
-        queue = self.queue
-
-        if queue is not None:
+        if self.queue is not None:
+            self.queue.run()
             self.queue = None
-            queue.run()
 
     # interface for receiving messages
     def do_channel_control(self, channel: str, command: str, message: Dict[str, object]) -> None:
```

martinpitt: Curiously that fails: Adding `assert self.queue` at the top triggers the assertion, which means that the unit tests somehow run into the `init` event before the `connected` event? I.e. they try to thaw before a freeze?

src/cockpit/router.py
```diff
@@ -26,12 +27,54 @@
...
+
+
 class Endpoint:
     router: Router
+    queue: Optional[ExecutionQueue] = None
 
     def __init__(self, router: Router):
         self.router = router
 
+    def freeze_endpoint(self):
```
martinpitt, line 68: I tried the opposite here, `assert self.queue is None`, for preventing double freezing. That seems fine.

However, I wonder if `queue` should really be a class variable instead of an instance one? Doesn't that mean that all channels (which are Endpoints, right?) share a single queue?

src/cockpit/router.py
```diff
@@ -26,12 +27,54 @@
 logger = logging.getLogger(__name__)
 
 
+class ExecutionQueue(collections.deque):
```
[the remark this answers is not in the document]
allisonkarlitskaya: This is kinda something like the exact reason that I liked my previous approach a lot better.  Before, the queue object was just a queue.  Pure data.  Now it's this weird magical state tracking object glued to the side of a queue...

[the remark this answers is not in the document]
martinpitt: It just seemed weird to me to have an API to wrap(), but no API to unwrap. I think I'd be okay with adding an explicit `unwrap` to hide that gross `del`. It still leaves the housekeeping to the API caller then, which is kind of ugly, but with a bunch of extra assertions this would be good enough, I figure.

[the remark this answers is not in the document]
martinpitt: Also, I'd be fine with dismissing this alert, or adding a `NotImplementedError` stub.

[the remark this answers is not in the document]
allisonkarlitskaya: I took a different approach: ExecutionQueue is no longer a subclass of `deque`, but rather has it as an instance method.

src/cockpit/router.py
allisonkarlitskaya: Fixed.

src/cockpit/router.py
martinpitt: This is the only outstanding thread here. Adding an assertion here would be nice, but also not super-important. I'm happy with this PR once it gets green. Thanks!

martinpitt (dismissed):
Thanks for bearing with my complaints :grin: This looks really nice now! If/when the bots are happy, so am I.

allisonkarlitskaya:
```
not ok 256 test/verify/check-system-info TestSystemInfo.testTime # expected failure: still fails with python bridge
```

grumble grumble.  too much unbroken stuff!!

martinpitt (approved):

PR #18179, opened 2023-01-16
test: Don't swallow duplicate Browser messages by default

mvollmer:
But leave the code in place in case we need to switch this on again.

### DIFF
--- a/test/common/chromium-cdp-driver.js
+++ b/test/common/chromium-cdp-driver.js
@@ -66,7 +66,6 @@ const messages = [];
 let logPromiseResolver;
 let nReportedLogMessages = 0;
 const unhandledExceptions = [];
-const shownMessages = []; // Show every message just once, keep here seen messages
 
 function clearExceptions() {
     unhandledExceptions.length = 0;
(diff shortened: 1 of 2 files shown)


martinpitt:
@mvollmer : Something went wrong in GitHub with this PR. Can you please do a force-push to re-trigger the workflows?

test/common/chromium-cdp-driver.js
```diff
@@ -31,6 +31,7 @@
 const CDP = require('chrome-remote-interface');
 
 let enable_debug = false;
+const show_duplicate_messages = true;
```
martinpitt, line 34: Overengineered, IMHO. If this works and an inspection of the raw log doesn't show us a gazillion duplicates, then let's just drop it for good. For the "we need it again" case there is `git revert` :grin:

martinpitt (changes requested):

test/common/chromium-cdp-driver.js
mvollmer: Can't nobody find that in the commit history... :-) But yes, I am happy to rip it out completely.

martinpitt:
[raw chromium log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18179-20230116-153235-81dab67f-fedora-36/log) is 444 kB, [raw firefox log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18179-20230116-153235-81dab67f-fedora-37-firefox/log) is 453 kB, which is pretty much the same as on main.  Neither have undue repetition in it, so this looks good. Cheers!

martinpitt (approved):

PR #18183, opened 2023-01-16
python: apply user provided env variables

jelly:
Apply the user provided `environ` option, which makes the terminal page tests pass as now TERM is set to something sensible.

- [x] rebase on, and then land #18189
- [x] rebase on, and then land #18185

### DIFF
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -83,6 +83,7 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
         cwd: Optional[str] = options.get('directory')
         pty: bool = options.get('pty', False)
         window: Dict[str, int] = options.get('window')
+        environ = options.get('environ')
 
         if err == 'out':
             stderr = subprocess.STDOUT
@@ -92,7 +93,11 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
             stderr = subprocess.PIPE
 
         env: Dict[str, str] = dict(os.environ)
-        env.update(options.get('env') or [])
+        if environ is not None:
+            if not isinstance(environ, list) or not all(isinstance(e, str) and '=' in e for e in environ):
+                raise ChannelError('protocol-error', message='invalid "environ" option for stream channel')
+
+            env.update(dict(e.split('=', 1) for e in environ))
 
         try:
             logger.debug('Spawning process args=%s', args)
--- a/test/pytest/test_transport.py
+++ b/test/pytest/test_transport.py
@@ -457,3 +457,11 @@ class TestSubprocessTransport(unittest.IsolatedAsyncioTestCase):
         transport.set_window_size(44, 55)
         while b'44x55\r\n' not in protocol.get_output():
             await asyncio.sleep(0.1)
+
+    async def test_env(self) -> None:
+        protocol, transport = self.subprocess(['bash', '-ic', 'echo $HOME'],
+                                              pty=True,
+                                              env={'HOME': '/test'})
+        protocol.output = []
+        while b'/test\r\n' not in protocol.get_output():
+            await asyncio.sleep(0.1)
--- a/test/verify/check-system-terminal
+++ b/test/verify/check-system-terminal
@@ -18,7 +18,7 @@
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
 import parent  # noqa: F401
-from testlib import MachineCase, nondestructive, skipBrowser, skipDistroPackage, todoPybridge, test_main
+from testlib import MachineCase, nondestructive, skipBrowser, skipDistroPackage, test_main
 
 
 def line_sel(i):
@@ -42,7 +42,6 @@ PROMPT_COMMAND='printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/
         self.machine.execute("systemctl reset-failed")
 
     @skipBrowser("Firefox needs http to access clipboard", "firefox")
-    @todoPybridge()
     @nondestructive
     def testBasic(self):
         b = self.browser


jelly:
Holdup, also implementing `directory` support on top of this as the terminal test probably needs it too :)

```
pkg/systemd/terminal.jsx:     * Spawns the user's shell in the user's home directory.
pkg/systemd/terminal.jsx:                directory: user.home || "/",
```

jelly:
Actually testTerminal passes \o/

jelly:
Ideally the PR author would also add some tests in `test/pytest/test_transport.py` :)

src/cockpit/channels/stream.py
```diff
@@ -66,6 +66,7 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
         cwd: Optional[str] = options.get('directory')
         pty: bool = options.get('pty', False)
         window: Dict[str, int] = options.get('window')
+        environ: Optional[List[str]] = options.get('environ')
```
allisonkarlitskaya, line 69: I've actually grown to dislike these sorts of type annotations.  I'd prefer you exclude it and let mypy complain.  This is a legitimate error that we'll need to address some day.

src/cockpit/channels/stream.py
```diff
@@ -75,7 +76,11 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
             stderr = subprocess.PIPE
 
         env: Dict[str, str] = dict(os.environ)
-        env.update(options.get('env') or [])
+        if environ:
```
allisonkarlitskaya, line 80: `... is not None`

src/cockpit/channels/stream.py
```diff
@@ -75,7 +76,11 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
...
+            if not isinstance(environ, list):
```
allisonkarlitskaya, line 81: ...but this makes it redundant.  This will fail if `environ` is `None`, so this check is enough on its own.

src/cockpit/channels/stream.py
```diff
@@ -75,7 +76,11 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
...
+                raise ChannelError('protocol-error', message='invalid "environ" option for stream channel')
```
allisonkarlitskaya, line 82: But if you want to do that, then you also need to check if (a) every item is a string, and (b) every item contains an `=` sign.

allisonkarlitskaya (changes requested):
Thanks!  See suggested changes below.

src/cockpit/channels/stream.py
jelly: I suppose, the C bridge does `cockpit_json_get_strv` which does:
```
 * Gets a string array member from a JSON object. Validates
 * that the member is an array, and that all elements in the
 * array are strings. If these fail, then will return %FALSE.
```

src/cockpit/channels/stream.py
jelly: I kinda dislike mixing styles though, but I'm fine with it.

src/cockpit/channels/stream.py
```diff
@@ -75,7 +76,11 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
             stderr = subprocess.PIPE
 
         env: Dict[str, str] = dict(os.environ)
-        env.update(options.get('env') or [])
+        if environ is not None:
+            if not isinstance(environ, list) or isinstance(environ, list) and not all(isinstance(e, str) and '=' in e for e in environ):
+                raise ChannelError('protocol-error', message='invalid "environ" option for stream channel')
+
+            env.update({n: v for n, v in (envline.split('=') for envline in environ)})
```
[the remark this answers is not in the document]
jelly: You missed the `if 'PWD'` case which I introduced that won't make it possible to `env.update` with it.

src/cockpit/channels/stream.py
```diff
@@ -72,6 +72,7 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
         pty: bool = options.get('pty', False)
         window: Dict[str, int] = options.get('window')
         environ = options.get('environ')
+        directory = options.get('directory')
```
allisonkarlitskaya, line 75: This is already called `cwd`, 4 lines up . Possibly confusing.  Feel free to rename it.

src/cockpit/channels/stream.py
```diff
@@ -85,11 +86,17 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
             if not isinstance(environ, list) or not all(isinstance(e, str) and '=' in e for e in environ):
                 raise ChannelError('protocol-error', message='invalid "environ" option for stream channel')
 
-            env.update({n: v for n, v in (envline.split('=', 1) for envline in environ)})
+            userenv = {n: v for n, v in (envline.split('=', 1) for envline in environ)}
```
allisonkarlitskaya, line 89: Unnecessary comprehension.  This looks better like:

```python
userenv = dict(e.split('=', 1) for e in environ)
```

src/cockpit/channels/stream.py
```diff
@@ -80,11 +81,17 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, A
...
+            # The kernel only knows about the inode of the current directory.
+            # So when we spawn a shell, it won't know the directory it's
+            # meant to display. Pass it the path we care about in $PWD
+            if 'PWD' not in userenv and directory:
```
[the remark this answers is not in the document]
allisonkarlitskaya: I'm also not sure I love the logic here.  I don't think there should be any case where the user provides a `PWD` environment variable explicitly and it's not the same as what we set the `cwd=` to.  I'm not sure which way you want to resolve that conflict (`directory` option wins vs. `PWD` variable wins) but they shouldn't be different, in any case.

allisonkarlitskaya (dismissed):

martinpitt:
@jelly : Depending PR #18185 landed, please rebase.

src/cockpit/channels/stream.py
jelly: How did I read over that....

src/cockpit/channels/stream.py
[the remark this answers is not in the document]
jelly: Ok, I did some testing with our terminal. If I remove this logic, PWD is still set fine. So either subprocess does the right thing or hmmm it's set in `os.environ`

src/cockpit/channels/stream.py
jelly: Done!

martinpitt:
OK, I'm sick of staring at these flakes. I'll look at them during my pilot shift, but let's not stall this any longer.

PR #18184, opened 2023-01-17
python: read superuser rules from manifests

allisonkarlitskaya:


martinpitt:
I pushed a rebase to main with (trivial) conflict resolution. No other changes yet.

martinpitt:
Notes about [TestSuperuser.testSingleLabelBridgeConfig](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18184-20230123-153230-6079ec29-fedora-37-pybridge/log.html#329) failure:
 - `SuperuserRoutingRule.go()` gets called with "sudo", but we expect it to be the sole available method `hello`. That causes the `KeyError` and the error message. I locally extended the error message to show the requested name. So where does that `sudo` come from? Nothing in the Py bridge hardcodes it, and the bridge rules *only* have `hello`.
 - `SuperuserRoutingRule.init()` is **not** called. KeyError traceback shows that `go('sudo')` is called via `.start('sudo')` through D-Bus.
 - pkg/shell/superuser.jsx init() gets the `.Methods` D-Bus property, and only falls back to hardcoding "sudo" if that is empty.  Indeed that property does not exist in the py bridge. It's supposed to be an `a{sv}` mapping "id" to "label". 

@allisonkarlitskaya : I added (*not* force-pushed) a FIXUP which implements `.Methods`. This makes `__main__.TestSuperuser.testSingleLabelBridgeConfig` pass. The other two still need some more work, will look later.

martinpitt:
I fixed the Methods return value (it's a more complicated structure), and added a shell fix. Should be back to three failures now.

martinpitt:
Argh argh [testMultipleBridgeConfig](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18184-20230124-135736-0b4a56d0-fedora-37-pybridge/log.html#324) -- I already re-todo'ed that in the meantime, but seems I undid that accidentally. This will require more work (polkit agent) → separate PR.

martinpitt:
[This NM failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18184-20230124-142056-eacf5a40-fedora-37-pybridge/log.html#81-2) is a flake. We've had that for quite some time, same as  [this one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18225-20230123-165954-db139c5b-debian-testing/log.html#81-2). I tried hard to reproduce that locally, with running the same tests in a VM, running it in a loop, etc. No joy so far. But unrelated to this PR.

src/cockpit/superuser.py
```diff
@@ -146,7 +135,9 @@ def peer_authorization_request(self, peer: Peer, message: Optional[str], prompt:
     def __init__(self, router: Router, privileged: bool = False):
         super().__init__(router)
 
-        self.bridges = list(SUPERUSER_BRIDGES)
+        # name → (label, spawn, env)
+        self.superuser_rules: Dict[Optional[str], Tuple[str, List[str], Dict[str, str]]] = {}
```
allisonkarlitskaya, line 139: Honestly, I don't care for this — but I have some help coming in a new PR.

allisonkarlitskaya (commented):
Pretty much everything looks good to me.  Thanks for working on this!

src/cockpit/superuser.py
martinpitt: OK, sorry then -- I thought annotating non-trivial data structures like this would be helpful. Please let me know how you want this changed, I'm not very proficient in this yet.

PR #18187, opened 2023-01-17
python: metrics: fix rate calculation with 0 values

allisonkarlitskaya:
A conditional in the bridge aims to return a rate of change if it has a previous value from which to calculate a change, and the value "False" otherwise.  In case there is no previous value, it would be None.

Unfortunately that was coded like

```python
  if old_value:
```

instead of

```python
  if old_value is not None:
```

which means that if `old_value` happened to be `0`, we would keep reporting the rate of change as "false".

It just so happens that depending on which other tests ran before the `@nondestructive` `TestCurrentMetrics.testCPU`, the amount of "nice" CPU usage may or may not be zero, causing this bug to be triggered only sometimes.

When triggered, this was causing the CPU percentage to show up as NaN%.

### DIFF
--- a/src/cockpit/channels/metrics.py
+++ b/src/cockpit/channels/metrics.py
@@ -121,7 +121,7 @@ class InternalMetricsChannel(AsyncChannel):
         return samples
 
     def calculate_sample_rate(self, value: Any, old_value: Optional[Any]):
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice catch!

PR #18188, opened 2023-01-17
python: Implement closing of spawn channels

martinpitt:
This is being used to terminate a process, like on the sosreport page.

---

Goes on top of PR #18181, mostly to avoid test change conflicts.

### DIFF
--- a/pkg/base1/test-spawn-proc.js
+++ b/pkg/base1/test-spawn-proc.js
@@ -250,4 +250,17 @@ QUnit.test("stream large output", function (assert) {
             });
 });
 
+QUnit.test.skipWithPybridge("cancel process", async assert => {
+    const proc = cockpit.spawn(["sleep", "418"]);
+    await cockpit.script("until pgrep -af [s]leep.*418; do sleep 0.1; done");
+    proc.close("cancelled");
+    await cockpit.script("timeout 5 sh -ec 'while pgrep -af [s]leep.*418; do sleep 0.1; done'");
+    try {
+        await proc;
+        assert.ok(false, "proc should have failed");
+    } catch (ex) {
+        assert.equal(ex.problem, "cancelled", "proc failed with correct problem");
+    }
+});
+
 QUnit.start();
--- a/pkg/base1/test-spawn-proc.js
+++ b/pkg/base1/test-spawn-proc.js
@@ -250,7 +250,7 @@ QUnit.test("stream large output", function (assert) {
             });
 });
 
-QUnit.test.skipWithPybridge("cancel process", async assert => {
+QUnit.test("cancel process", async assert => {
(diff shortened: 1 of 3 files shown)


mvollmer (approved):
Nice! Thanks!

allisonkarlitskaya:
This regressed `TestLogin.testLogging`.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-18189-20230117-152118-88878800-fedora-37-pybridge/log.html#180

```
WARNING:asyncio:child process pid 22760 exit status already read: will report returncode 255
```

Calling `terminate()` on the subprocess triggers a piece of code deep inside the Python standard library that attempts to avoid the classical pid-reuse race by first checking if the process has exited.  That calls `waitpid()` which then intereferes with `asyncio.PidFdChildWatcher` which expects to be able to call `waitpid()` for itself.

The situation where we'll hit this is exactly the case that the code in question is trying to detect: when you're calling .terminate() on a process that already exited.  If the process is still running then the `waitpid()` call is harmless.

This doesn't break anything, since we clearly want the channel to go away at this point anyway, but it does result in extra junk in the journal.  And it's a bug anyway. 

Some ideas for a fix:

 - step up our attempts to prevent subprocess from reaping itself
 - avoid using the `.terminate()` function on the subprocess and do it via fd or pidfd instead
 - make sure we unregister the pidfd watch before we call `terminate()`

martinpitt:
Thanks @allisonkarlitskaya for the analysis.  This affects [other tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18185-20230117-164632-269e4108-fedora-37-pybridge/log.html#315) as well, seems fairly random.

Initially that looked like `do_close()` would be callled also for normal process exits, which is certainly not intended. A first check with the qunit tests does not confirm that, though. OTOH, the only place in pkg/users which explicitly calls `.close()` on a spawn channel is the password dialog after calling `passwd` times out. But this would also break the test, which does not happen, and this dialog isn't even tested in TestAccounts.testGroups.

I can reproduce this locally fairly well with TestAccounts.testGroups, the other two don't bite for me. With better logging, I get this:
```
cockpit.channels.stream:XXX created transport for pid 18049: ['/usr/bin/last', '--time-format', 'iso', '-n', '25', 'admin']
cockpit.channels.stream:XXX Received close signal from peer, terminating process 18049
asyncio:child process pid 18049 exit status already read:  will report returncode 255
```
I ran it several times, got 5 instances of that failure, and it's always the same `last` command, from pkg/users/account-logs-panel.jsx. Notably, this does **not call .close()** on the process, so this re-affirms my theory that `do_close()` gets called here for a regularly terminating process.

I'll investigate this further, and will update this comment.

martinpitt:
I sent PR #18202 with a first attempt.

PR #18189, opened 2023-01-17
webpack.config.js: don't gzip .html pages

allisonkarlitskaya:
The python bridge doesn't know how to decompress gzip, and -ws requires HTML to be decompressed so that it can patch it while serving it to the client.

All of our .html pages are small anyway, and since we're serving them uncompressed, we may as well also have them on disk uncompressed, saving ourselves the trouble.

 - [x] https://github.com/cockpit-project/cockpit/pull/18190 so that we don't try to jumpstart across a config change

### DIFF
--- a/test/verify/check-system-terminal
+++ b/test/verify/check-system-terminal
@@ -26,7 +26,6 @@ def line_sel(i):
 
 
 @skipDistroPackage()
-@todoPybridge()
 class TestTerminal(MachineCase):
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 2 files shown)


jelly:
While it makes sense, it doesn't seem to be that difficult to add `Accept-Encoding` handling in `cockpit.packages`, as I think this will break everyone who tries to embed their own plugin cockpit currently and they will end up with a "blank white screen" as nothing can be loaded.

allisonkarlitskaya:
Regardless of if we consider this a valid bugfix for the issue of the pybridge decompressing `.html.gz` files vis-a-vis third party packages, I think this makes sense anyway: why do we store these (small) files compressed on disk just to decompress them every time we serve them?  Let's just have them uncompressed.

allisonkarlitskaya:
This blocks #18183 because we both remove the "last" broken testcase in `test/verify/check-system-terminal`

allisonkarlitskaya:
https://cockpit-logs.us-east-1.linodeobjects.com/pull-18189-20230117-152118-88878800-fedora-37-pybridge/log.html#180 is fallout from #18188 and definitely unrelated to this PR.

martinpitt (approved):
Agree that this makes sense conceptually. We don't have to apply it to our external projects yet (we may still find a solution to transparently uncompress), but this is good either way. Thanks!

PR #18191, opened 2023-01-17
firewall: Show additional ports even when no service is defined 

marusak:


marusak:
Common flake, debugging it in https://github.com/cockpit-project/cockpit/pull/18201

test/common/test-functions.js
```diff
@@ -51,14 +51,6 @@ function ph_set_val (sel, val) {
     el.dispatchEvent(ev);
 }
 
-function ph_set_text_direction(sel, direction) {
```
martinpitt, line 54: Ah, I remember that leftover. Thanks for cleaning up! Should have been a separate commit, but :man_shrugging:

martinpitt (approved):
Danke!

PR #18192, opened 2023-01-17
DBG - Mark TestJournal.testAbrtReportCancel as stable

mvollmer:
I can't get this to fail locally.


martinpitt (approved):
It succeeded three times in a row, I didn't see it fail recently either. Presumably it was fixed by one of the countless recent PRs. That's why I didn't like `flaky=True` to begin with, so away with them!

PR #18201, opened 2023-01-18
tests: fix testFirewall.testServices pixel test flakiness on RTL screenshot

marusak:


KKoukiou (approved):

test/common/testlib.py
```diff
@@ -1171,12 +1169,12 @@ def assert_pixels(self, selector: str, key: str,
             for layout in self.layouts:
                 if layout["name"] not in skip_layouts:
                     self.set_layout(layout["name"])
+                    time.sleep(wait_delay)
```
jelly, line 1172: This means even layouts which have no pixel issues will have to wait 0.5 seconds for nothing. That feels a bit expensive for a RTL specific issue.

jelly: Ah, I read that wrong, we did already sleep in `assert_pixels_in_current_layout` called below.

But I wonder if this will introduce other flakes, seeing that `assert_pixels_in_current_layout` is also called (now without sleep).
```
test/verify/check-system-services:            b.assert_pixels_in_current_layout("#services-page", "text-filter-test")
test/verify/check-system-services:            b.assert_pixels_in_current_layout("#nav-system", pixel_label)
```

KKoukiou: Why do we use this assert_pixels_in_current_layout and not the assert_pixels? Do you maybe know?

KKoukiou: Also this wait_delay is not set in any of these two examples, so there are not affected.

test/common/testlib.py
```diff
@@ -1171,12 +1169,12 @@ def assert_pixels(self, selector: str, key: str,
...
                     if "rtl" in self.current_layout["name"]:
                         self._set_direction("rtl")
                     self.assert_pixels_in_current_layout(selector, key, ignore=ignore,
                                                          scroll_into_view=scroll_into_view,
-                                                         wait_animations=wait_animations,
-                                                         wait_delay=wait_delay)
+                                                         wait_animations=wait_animations)
 
                     if "rtl" in self.current_layout["name"]:
                         self._set_direction("ltr")
```
jelly, line 1180: What if we sleep here? As this is the issue right? We switch the layout back but it takes longer then expected.

KKoukiou: Hmm that's not the issue. The failing pixel test, is supposed to be RTL but it's showing the original UI. So the problem is not when resetting the direction to ltr.

test/common/testlib.py
jelly: `wait_delay: float = 0.5)` => wait_delay=0.5 so it's set by default.

test/common/testlib.py
jelly: Aha! But we sleep now before we switch rtl? So is the issue that something is still rendering before we switch rtl?

KKoukiou: That's exactly my assumption

test/common/testlib.py
KKoukiou: oops right, so now we still sleep but before we set the RTL.

jelly: Seems the test time is still around the same time: https://cockpit-logs.us-east-1.linodeobjects.com/pull-18172-20230122-103906-69df97bb-fedora-37/log.html

And 52 minutes here. So maybe I was bit over concerned?

martinpitt:
Triggering fedora-37, as that's the one doing pixel tests now.

PR #18202, opened 2023-01-18
python: Avoid waiting for the child in SubprocessStreamChannel.do_close()

martinpitt:
Commit e100e4d8b introduced a race condition: asyncio installs a "child    
watcher" (via pidfd) for all launched subprocesses, and expects it to be    
able to read the processes' exit code. But subprocess'es terminate()    
method *also* does that. If the process is still running, this is    
harmless (as it will just return None), but if the process terminates    
around the same time as cancelling it, it can happen that the process    
stopped in terminate(), it waits() the child and collects the exit code,    
and asyncio's child watcher then does not get it any more. This causes    
warnings like    
    
    asyncio:child process pid <PID> exit status already read:  will report returncode 255    
    
We actually want to retain that; terminate()'s wait call is just for    
avoiding PID reuse races, but that race is negligible for us.    
cockpit.spawn() already ensures that do_close() is only actually called    
for valid (running) processes, and the kernel usually makes sure that a    
PID does not get re-used immediately.    

----

This should fix [these failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18185-20230117-164632-269e4108-fedora-37-pybridge/log.html#315). I realize that this is somewhat hackish, but I'm afraid I'm too stupid to do better here. See https://github.com/cockpit-project/cockpit/pull/18188#issuecomment-1386470504 for my many failed attempts to catch this in a unit test. I understand fairly well *why* this happens, but I don't have a good idea how to improve it to be completely race free.

### DIFF
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -74,8 +74,16 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
 
     def do_close(self):
         assert isinstance(self._transport, SubprocessTransport)
-        logger.debug('Received close signal from peer, terminating process %i', self._transport.get_pid())
-        self._transport.terminate()
+        pid = self._transport.get_pid()
+        # avoid calling .terminate(), as that will try to read the process' exit code and race with
+        # asyncio's ChildWatcher (which also wants to wait() the process); the cockpit.spawn() API
+        # already ensures that the process is valid before even calling do_close.
+        try:
+            os.kill(pid, signal.SIGTERM)
+            logger.debug('Received close signal from peer, SIGTERMed process %i', pid)
+        except ProcessLookupError:
+            # already gone? fine!
+            logger.debug('Received close signal from peer, but process %i is already gone', pid)
 
     def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, Any]) -> SubprocessTransport:
         args: list[str] = options['spawn']
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -95,9 +95,10 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
         env.update(options.get('env') or [])
 
         try:
-            logger.debug('Spawning process args=%s', args)
-            return SubprocessTransport(loop, self, args, pty, window, env=env, cwd=cwd, stderr=stderr,
-                                       preexec_fn=lambda: prctl(SET_PDEATHSIG, signal.SIGHUP))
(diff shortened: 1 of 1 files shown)


martinpitt:
Yay, two green test runs in a row :tada:

martinpitt:
[This log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18204-20230118-221555-54b3003a-fedora-37-pybridge/TestLogin-testLogging-fedora-37-127.0.0.2-3001-FAIL.log.gz) *may* be a clue why these `close(disconnected)` calls happen:
```
asyncio-ERROR: Exception in callback _Transport._read_ready()    
handle: <Handle _Transport._read_ready()>    
  File "/usr/lib64/python3.11/asyncio/events.py", line 80, in _run    
    self._context.run(self._callback, *self._args)    
  File "/usr/lib/python3.11/site-packages/cockpit/transports.py", line 83, in _read_ready    
    data = os.read(self._in_fd, _Transport.BLOCK_SIZE)    
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^    
ConnectionResetError: [Errno 104] Connection reset by peer    
```
That's still one of the traces which testlib.py message filtering covers.

allisonkarlitskaya:
> [This log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18204-20230118-221555-54b3003a-fedora-37-pybridge/TestLogin-testLogging-fedora-37-127.0.0.2-3001-FAIL.log.gz) _may_ be a clue why these `close(disconnected)` calls happen:
> 
> ```
> asyncio-ERROR: Exception in callback _Transport._read_ready()    
> handle: <Handle _Transport._read_ready()>    
> Traceback (most recent call last):    
>   File "/usr/lib64/python3.11/asyncio/events.py", line 80, in _run    
>     self._context.run(self._callback, *self._args)    
>   File "/usr/lib/python3.11/site-packages/cockpit/transports.py", line 83, in _read_ready    
>     data = os.read(self._in_fd, _Transport.BLOCK_SIZE)    
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^    
> ConnectionResetError: [Errno 104] Connection reset by peer    
> ```
> 
> That's still one of the traces which testlib.py message filtering covers.

I've run into this a few times before.  This is coming from the privileged bridge.  It should be suppressed by the top-level `except ConnectionResetError` in `bridge.py` but for some reason it's not...

allisonkarlitskaya (approved):
I agree that this will fix the issue.

In the long-term I consider our use of subprocess to be a bit suspect, considering the lengths we have to go to to avoid it trying to reap itself.  `asyncio` in the standard library uses the same tricks we do, but it's in the standard library, so it's allowed to be "evil".  We have to worry about compatibility issues...

PR #18203, opened 2023-01-18
systemd: consider uint64(-1) as invalid for memory

jelly:
Similiar to 724b64a1b82b8cfa8abf6f10b8c8 only consider valid values between 0 and MAX_UINT64.


martinpitt (approved):
:guitar: cheers!

PR #18204, opened 2023-01-18
python: Respect umask in fsreplace1 (umask version)

martinpitt:
NamedTemporaryFile always creates files with 600 permissions, thus making e.g. /etc/motd unreadable for users (which breaks TestSystemInfo.testMotd). So lazily read the current umask (as we don't change it anywhere), and fix the file permissions right before renaming.

Detecting the current umask is a bit nasty, as one has to call umask(2) twice. But that's the recommended method, and if it's good enough for umask(1), it's good enough for us. Unfortunately there is no really good Linux API for atomically replacing a file as user, all the other options are much worse.

----

This is an alternative to PR #18193. I personally like it better.

### DIFF
--- a/src/cockpit/channels/filesystem.py
+++ b/src/cockpit/channels/filesystem.py
@@ -23,6 +23,7 @@ from systemd_ctypes import PathWatch
 from systemd_ctypes.inotify import Event as InotifyEvent
 
 from ..channel import Channel, ChannelError
+from ..config import get_umask
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


src/cockpit/channels/filesystem.py
```diff
@@ -115,10 +115,17 @@ def do_open(self, options):
...
     payload = 'fsreplace1'
 
+    _umask = None
     _path = None
     _tag = None
     _tempfile = None
 
+    def get_umask(self):
+        if self._umask is None:
+            self._umask = os.umask(0o077)
```
allisonkarlitskaya, line 125: Hard no.

If you want to do this, then it happens exactly once, and it happens as a call from the start of `main()` in `bridge.py`, please.

We don't have threads (that we know about, as of today) but burying thread-unsafe code like this to be run semi-regularly deep inside our code is a no-go for me.

allisonkarlitskaya (changes requested):

martinpitt:
Closing in favor of #18193

PR #18205, opened 2023-01-19
test: Adjust VDO tests for latest RHEL 9.2 image

martinpitt:
See https://github.com/cockpit-project/bots/pull/4291 . I tested this locally against thew new RHEL image, and also triggered a run against the refreshed image here.


martinpitt:
I tested this locally against thew new RHEL image. I didn't run it here as there are still two other failures to sort out.

mvollmer (approved):

PR #18206, opened 2023-01-19
test: Drop obsolete udevadm trigger hacks

martinpitt:
https://github.com/storaged-project/udisks/pull/631 landed four years ago in udisks 2.8.2. Even RHEL 8 has 2.9.0.

Likewise for check-storage-lvm2, tests run fine without the extra trigger.


mvollmer (approved):
If the tests are happy, so am I.

PR #18207, opened 2023-01-19
lib: Don't use "undefined" as column key with ListingTable

mvollmer:
React will complain about this.

### DIFF
--- a/pkg/lib/cockpit-components-table.jsx
+++ b/pkg/lib/cockpit-components-table.jsx
@@ -218,16 +218,16 @@ export const ListingTable = ({
                     {row.columns.map((cell, cellIndex) => {
                         const { key, ...cellProps } = cell.props || {};
                         const dataLabel = typeof cells[cellIndex] == 'object' ? cells[cellIndex].title : cells[cellIndex];
-
+                        const colKey = dataLabel || cellIndex;
                         if (cells[cellIndex] && cells[cellIndex].header)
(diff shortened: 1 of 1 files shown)


mvollmer:
Running TestSystemInfo.testAbrtReport should show the effect of this change.

pkg/lib/cockpit-components-table.jsx
```diff
@@ -218,16 +218,16 @@ export const ListingTable = ({
                     {row.columns.map((cell, cellIndex) => {
                         const { key, ...cellProps } = cell.props || {};
                         const dataLabel = typeof cells[cellIndex] == 'object' ? cells[cellIndex].title : cells[cellIndex];
-
+                        const colKey = dataLabel || cellIndex;
```
martinpitt, line 221: I suppose we don't need to be overly concerned about `""` labels? If we do, then this needs to be `??` instead of `||`.

martinpitt (approved):
Danke!

pkg/lib/cockpit-components-table.jsx
mvollmer: `??`?

martinpitt: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

PR #18219, opened 2023-01-20
systemd: support a subset of crypto subpolicies

jelly:
Support DEFAULT:SHA1, LEGACY:AD-SUPPORT and FIPS:OSPP cryptography polices.

### DIFF
--- a/pkg/systemd/overview-cards/cryptoPolicies.jsx
+++ b/pkg/systemd/overview-cards/cryptoPolicies.jsx
@@ -34,8 +34,10 @@ const _ = cockpit.gettext;
 // Found in /usr/share/crypto-policies/policies/
 const cryptopolicies = {
     DEFAULT: _("Recommended, secure settings for current threat models."),
+    "DEFAULT:SHA1": _("DEFAULT with SHA-1 signature verification allowed."),
     FUTURE: _("Protects from anticipated near-term future attacks at the expense of interoperability."),
     LEGACY: _("Higher interoperability at the cost of an increased attack surface."),
+    "LEGACY:AD-SUPPORT": _("LEGACY with Active Directory interoperability."),
     FIPS: (<Flex alignItems={{ default: 'alignItemsCenter' }}>
         {_("Only use approved and allowed algorithms when booting in FIPS mode.")}
         <Button component='a'
@@ -47,6 +49,7 @@ const cryptopolicies = {
             {_("Learn more")}
         </Button>
     </Flex>),
+    "FIPS:OSPP": _("FIPS with further Common Criteria restrictions."),
 };
 
 const displayProfileText = profile => profile === "FIPS" ? profile : profile.charAt(0) + profile.slice(1, profile.length).toLowerCase();
@@ -56,12 +59,16 @@ export const CryptoPolicyRow = () => {
     const Dialogs = useDialogs();
     const [currentCryptoPolicy, setCurrentCryptoPolicy] = useState(null);
     const [fipsEnabled, setFipsEnabled] = useState(null);
+    const [shaSubPolicyAvailable, setShaSubPolicyAvailable] = useState(null);
 
     useEffect(() => {
         cockpit.file("/proc/sys/crypto/fips_enabled").read()
                 .then(content => setFipsEnabled(content ? content.trim() === "1" : false));
         cockpit.file("/etc/crypto-policies/state/current")
                 .watch(content => setCurrentCryptoPolicy(content ? content.trim() : null));
+        // RHEL-8-8 has no SHA1 subpolicy
+        cockpit.file("/usr/share/crypto-policies/policies/modules/SHA1.pmod").read()
+                .then(content => setShaSubPolicyAvailable(content ? content.trim() : false));
     }, []);
 
     if (!currentCryptoPolicy) {
@@ -77,7 +84,8 @@ export const CryptoPolicyRow = () => {
                                   onClick={() => Dialogs.show(<CryptoPolicyDialog
                                                                   currentCryptoPolicy={currentCryptoPolicy}
                                                                   setCurrentCryptoPolicy={setCurrentCryptoPolicy}
-                                                                  fipsEnabled={fipsEnabled} />)}>
+                                                                  fipsEnabled={fipsEnabled}
+                                                                  shaSubPolicyAvailable={shaSubPolicyAvailable} />)}>
                     {displayProfileText(currentCryptoPolicy)}
                 </PrivilegedButton>
             </td>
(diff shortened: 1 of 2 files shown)


test/verify/check-system-info
```diff
@@ -911,6 +911,11 @@ password=foobar
         profile = m.execute(cmd + " --show").strip()
         b.wait_text("#crypto-policy-button", shown_profile_text(profile))
 
+        # Test if a new subpolicy can be set
+        new_profile = "DEFAULT:SHA1"
+        change_profile(profile, new_profile)
+
+        profile = m.execute(cmd + " --show").strip()
```
martinpitt, line 918: Missing assertion of `profile`?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -34,8 +34,10 @@ const _ = cockpit.gettext;
 // Found in /usr/share/crypto-policies/policies/
 const cryptopolicies = {
     DEFAULT: _("Recommended, secure settings for current threat models."),
+    "DEFAULT:SHA1": _("DEFAULT with SHA-1 signature verification allowed"),
     FUTURE: _("Protects from anticipated near-term future attacks at the expense of interoperability."),
     LEGACY: _("Higher interoperability at the cost of an increased attack surface."),
+    "LEGACY:AD-SUPPORT": _("LEGACY with Active Directory interoperability"),
```
martinpitt, line 40: Is there some way to read the available options/subpolicies from crypto-policies?

martinpitt (changes requested):
Thanks! The test fails on some OSes, though.

pkg/systemd/overview-cards/cryptoPolicies.jsx
marusak: There are many modifiers available and you can use multiple modifiers at once to change any existing policy. You can also write your own policies and modifiers and use those in any combinations. (If someone uses something different Cockpit will show it but won;t allow setting it).

This set is what was agreed on with crypto guys as sensible options that Cockpit should offer to customers as they make sense. Anything else is dangerous/weird/too unique that someone who wants to use that they surely know enough about this topic so setting it up on cmdline shouldn't be difficult for them.

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -47,6 +49,7 @@ const cryptopolicies = {
             {_("Learn more")}
         </Button>
     </Flex>),
+    "FIPS:OSPP": _("FIPS with further Common Criteria restrictions"),
```
marusak, line 52: Missing dot at the end of the sentence. Also I think we want to explain this better than just "further Common Criteria restrictions".

marusak: The dot is missing in all new added ones. Can you please add it everywhere?

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -34,8 +34,10 @@ const _ = cockpit.gettext;
 // Found in /usr/share/crypto-policies/policies/
 const cryptopolicies = {
     DEFAULT: _("Recommended, secure settings for current threat models."),
+    "DEFAULT:SHA1": _("DEFAULT with SHA-1 signature verification allowed"),
     FUTURE: _("Protects from anticipated near-term future attacks at the expense of interoperability."),
```
marusak, line 38: We talked that LEGACY should be last in the list and probably marked as "testing" or something similar as it is actually not advised to use in production but is meant for testing.

jelly: Wasn't that about `FUTURE`?

marusak: It was, sorry, meant FUTURE

jelly:
Updated it to 
image

jelly:
So we need to increase our timeout for rebooting and RHEL 8 has no `DEFAULT:SHA1` profile.

jelly:
@marusak RHEL-8-8 Does not support Default:SHA1 btw

pkg/systemd/overview-cards/cryptoPolicies.jsx
```diff
@@ -56,12 +59,16 @@ export const CryptoPolicyRow = () => {
...
+    const [shaSubPolicyAvailable, setShaSubPolicyAvailable] = useState(null);
 
     useEffect(() => {
         cockpit.file("/proc/sys/crypto/fips_enabled").read()
                 .then(content => setFipsEnabled(content ? content.trim() === "1" : false));
         cockpit.file("/etc/crypto-policies/state/current")
                 .watch(content => setCurrentCryptoPolicy(content ? content.trim() : null));
+        // RHEL-8-8 has no SHA1 subpolicy
+        cockpit.file("/usr/share/crypto-policies/policies/modules/SHA1.pmod").read()
+                .then(content => setShaSubPolicyAvailable(content ? content.trim() : false));
```
cockpituous, line 71: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18219-20230125-132112-3a64f51e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

jelly:
Screenshot from 2023-01-26 11-23-04

@garrett do you have any comments about the new profile strings for `Default:sha1` and `Fips:ospp`?

pkg/systemd/overview-cards/cryptoPolicies.jsx
cockpituous, line 71: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18219-20230126-101600-3ffc4e6f-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

test/verify/check-system-info
```diff
@@ -899,7 +899,7 @@ password=foobar
                 # Initramfs re-generation takes a while
                 m.wait_reboot(timeout_sec=600)
             else:
-                m.wait_reboot()
+                m.wait_reboot(timeout_sec=300)
```
martinpitt, line 902: So I take it this also needs to re-generate the initrd then? In that case, just get rid of the `if` and just always use the long timeout.

test/verify/check-system-info
```diff
@@ -911,6 +911,17 @@ password=foobar
         profile = m.execute(cmd + " --show").strip()
         b.wait_text("#crypto-policy-button", shown_profile_text(profile))
 
+        # RHEL 8 has no SHA1 policy, so do not show it.
+        b.click("#crypto-policy-button")
+        func = b.wait_not_present if m.image.startswith('rhel-8') or m.image.startswith('centos-8') else b.wait_visible
+        func(".pf-c-menu__item-main .pf-c-menu__item-text:contains('Default:sha1')")
```
martinpitt, line 917: That's a funky way to write this, instead of an if/else and a constant for the selector, but sure :grimacing:

test/verify/check-system-info
martinpitt: This is still current. The test only validates that the UI reflects `--show`, but it should assert that --show *actually* says `new_profile` now.

martinpitt (changes requested):
Thanks! I'm mostly fine with this, just one piece of code cleanup and a missing assertion. @marusak also had some outstanding comments, so needs a re-review from him as well.

garrett:
> @garrett do you have any comments about the new profile strings for Default:sha1 and Fips:ospp?

What are the further common criteria restrictions? Are there too many to mention? Is there a way of finding out?

test/verify/check-system-info
jelly: Okido, done!

jelly:
> > @garrett do you have any comments about the new profile strings for Default:sha1 and Fips:ospp?
> 
> What are the further common criteria restrictions? Are there too many to mention? Is there a way of finding out?

This is really hard to find, but I believe it is something along the lines of this link below, an industry standard for common criteria for security ceritication.


https://www.redhat.com/en/about/press-releases/red-hat-adds-common-criteria-security-certification-red-hat-enterprise-linux

martinpitt (changes requested):
@jelly: My [earlier review comment](https://github.com/cockpit-project/cockpit/pull/18219#discussion_r1087883317) still stands.

test/verify/check-system-info
jelly: There is an assertion below right now?

martinpitt: thanks!

martinpitt (approved):

jelly:
The rawhide failure also appears in other PR's and the tests don't run on Rawhide anyway.

PR #18223, opened 2023-01-23
python: (temporarily) send */po.js along with /manifests.js

allisonkarlitskaya:
Partial solution to #18224

### DIFF
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -17,6 +17,7 @@
 
 
 import collections
+import gzip
 import hashlib
 import json
 import logging
@@ -375,6 +376,18 @@ class Packages(bus.Object, interface='cockpit.Packages'):
 
     def serve_manifests_js(self, channel):
         channel.http_ok('text/javascript')
+
+        # Send the translations required for the manifest files, from each package
+        locales = parse_accept_language(channel.headers)
+        for name, package in self.packages.items():
+            if name in ['static', 'base1']:
+                continue
+            # find_translation will always find at least 'en'
+            data, (content_type, encoding) = find_translation(package.translations, locales)
+            if encoding == 'gzip':
+                data = gzip.decompress(data)
+            channel.send_data(data)
+
         channel.send_data(("""
             (function (root, data) {
                 if (typeof define === 'function' && define.amd) {
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -160,17 +160,22 @@ class Package:
         # 0-pad each numeric component.  Only supports numeric versions like 1.2.3.
         return '.'.join(part.zfill(8) for part in version.split('.'))
 
+    def add_file(self, item: Path, checksums: List['hashlib._Hash']) -> None:
+        rel = item.relative_to(self.path)
+
+        with item.open('rb') as file:
+            data = file.read()
+
+        sha = hashlib.sha256(data).hexdigest()
+        for context in checksums:
+            context.update(f'{rel}\0{sha}\0'.encode('ascii'))
+
     def walk(self, checksums, path):
         for item in directory_items(path):
             if item.is_dir():
                 self.walk(checksums, item)
             elif item.is_file():
-                rel = item.relative_to(self.path)
-                with item.open('rb') as file:
-                    data = file.read()
-                sha = hashlib.sha256(data).hexdigest()
-                for context in checksums:
-                    context.update(f'{rel}\0{sha}\0'.encode('ascii'))
+                self.add_file(item, checksums)
 
     def check(self, at_least_prio):
         if 'requires' in self.manifest:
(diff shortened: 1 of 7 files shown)


src/cockpit/packages.py
```diff
@@ -83,19 +84,27 @@ def parse_accept_language(accept_language: str) -> List[str]:
     locales = []
     for language in accept_languages:
         language = language.strip()
-        locale, _, weight = language.partition(';q=')
-        weight = float(weight or 1)
+        locale, _, weightstr = language.partition(';q=')
```
martinpitt, line 87: You really don't like variable aliasing, do you :grin:

src/cockpit/packages.py
```diff
@@ -159,12 +169,25 @@ def sortify_version(version: str) -> str:
...
     def add_file(self, item: Path, checksums: List['hashlib._Hash']) -> None:
-        rel = item.relative_to(self.path)
-        self.files.add(rel)
+        rel = str(item.relative_to(self.path))
+        guessed_type = mimetypes.guess_type(rel)
 
         with item.open('rb') as file:
             data = file.read()
 
+        # Keep the file in memory to serve it later: if this is a 'po.*.js'
```
martinpitt, line 178: FTR, I prefer re-reading it from disk. If there's enough RAM, the kernel would already cache that anyway, and with memory pressure it has more chances to drop caches. Of course building the `self.translations` map is a good idea, but it could perhaps just map to the computed file name, instead of the file contents?

Not a hard "no", just a tought. You mentioned "14 MB", which isn't small, but these days it's not unreasonable either.

src/cockpit/packages.py
```diff
@@ -257,26 +241,44 @@ def get_content_security_policy(self, origin):
...
+            translation = self.translations.get(language)
+            if translation:
+                return translation
+
+        # If nothing else worked, we always have English
+        return self.translations['en']
+
+    def find_file(self, path, channel):
+        if path == 'po.js':
+            # We do locale-dependent lookup, only for /po.js
```
martinpitt, line 268: Confusing, comma

src/cockpit/packages.py
```diff
@@ -257,26 +241,44 @@ def get_content_security_policy(self, origin):
...
+            return self.find_po_js(channel)
         else:
-            logging.error("filename=%s, path=%s", filename, path)
+            return self.files.get(path)
+
+    def serve_file(self, path, channel):
+        found = self.find_file(path, channel)
+        if found is None:
+            logging.error("path=%s", path)
             logger.debug('path %s not in %s', path, self.files)
```
martinpitt, line 277: Redundant logging

martinpitt (changes requested):
Nice cleanup, thanks! Some minor things. Next round, please run all tests (or at least fedora-37), as this affects the C bridge as well.

src/cockpit/packages.py
allisonkarlitskaya: mypy doesn't like a variable name within a given scope to have two different types.

src/cockpit/packages.py
allisonkarlitskaya: Some thoughts:

 - the main issue here is that I want the file serving code to always do the same thing.  "Open this filename and serve the bytes" doesn't work as that thing, since we need to serve an empty file for English po.js (unless we want to use `/dev/null` as a filename)
 - my ideal approach would be to `mmap()` this, which would let the kernel drop it if it needed to, but allow us to serve it directly, but this isn't the safest option: if someone `O_TRUNC`s the file, we get `SIGBUS`, and I don't really feel like trying to sort that out in Python
 - we could also keep the fd around, but see above notes about `/dev/null`

martinpitt: .. or we could just actually create a `po.en.js` to get rid of that special case.

Either way, feel free to resolve -- this feels like it's well into the "optimize later if ever necessary" category, especially since going from the C to Python bridge increased memory usage by a *lot* anyway. :grin:

src/cockpit/packages.py
allisonkarlitskaya: Dunno why this was an error.  404s don't belong in the journal.

allisonkarlitskaya:
In principle this is working now.

The shell translations are working with manual testing, and running TestPtBRLocale and manually reloading the page also works.  I'm not sure why the test gets stuck — needs debugging.

allisonkarlitskaya:
> In principle this is working now.
> 
> The shell translations are working with manual testing, and running TestPtBRLocale and manually reloading the page also works. I'm not sure why the test gets stuck — needs debugging.

Of course it was our usual friendly chromium language-cookie-related caching behaviour issues ruining the day.  Tweaked the 'no-cache' rules to include `/manifests.js` as well now.

martinpitt (approved):
Sweet, thank you!

PR #18226, opened 2023-01-23
nav: fix toggling of the system dropdown

### LINKED ISSUE #18182: mobile: When in responsive mode, the system navigation menu pops up again when unasked for
Steps to reproduce:

2. Switch pages to something that's not cluttered (so you're not clicking on something accidentally), such as SELinux or Diagnostic Reports.  This happens all the time on the terminal page as well.

This breaks mobile functionality everywhere. It can also break desktop browsers too, but only when resized to be small and/or using large enough fonts to trigger the issue. (Both of these things can happen when used "in the wild".)

_Originally posted by @garrett in https://github.com/cockpit-project/cockpit/issues/18157#issuecomment-1384277486_

KKoukiou:
* Don't assume 'active' state of dropdown when toggling the 'interact' class.
* Call the addEventListener only on component mount, and not each time the 'active' state changes.

Fixes #18182

### DIFF
--- a/pkg/shell/nav.jsx
+++ b/pkg/shell/nav.jsx
@@ -25,10 +25,7 @@ export const SidebarToggle = () => {
     }, []);
 
     useEffect(() => {
-        if (active)
-            document.getElementById("nav-system").classList.add("interact");
-        else
-            document.getElementById("nav-system").classList.remove("interact");
+        document.getElementById("nav-system").classList.toggle("interact", active);
     }, [active]);
 
     return (
--- a/pkg/shell/nav.jsx
+++ b/pkg/shell/nav.jsx
@@ -19,10 +19,16 @@ export const SidebarToggle = () => {
     useEffect(() => {
         const handleClickOutside = () => setActive(false);
 
-        document.getElementById("nav-system").classList.toggle("interact");
         window.addEventListener("blur", handleClickOutside);
 
         return () => window.removeEventListener("blur", handleClickOutside);
+    }, []);
+
+    useEffect(() => {
+        if (active)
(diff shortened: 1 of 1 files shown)


pkg/shell/nav.jsx
```diff
@@ -19,10 +19,16 @@ export const SidebarToggle = () => {
...
         window.addEventListener("blur", handleClickOutside);
 
         return () => window.removeEventListener("blur", handleClickOutside);
+    }, []);
+
+    useEffect(() => {
+        if (active)
+            document.getElementById("nav-system").classList.add("interact");
+        else
+            document.getElementById("nav-system").classList.remove("interact");
```
garrett, line 31: This does the same thing, but with 1 line. It either adds or removes, based on the boolean(ish) value of `active` passed as a second option to [classList.toggle()](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle).

```suggestion
        document.getElementById("nav-system").classList.toggle("interact", active);
```

...But it could probably be integrated in the above `useEffect` block as well, right? (There probably wouldn't need to be two `useEffect` blocks?)

KKoukiou: We need the two useEffect blocks, so that the first one runs only on component's first render, as it adds the event listeners.

garrett:
I've been playing around with this on my phone this morning and it's still not fixed. Now it sometimes has the opposite problem, where the menu doesn't show up when it should.

garrett:
A guess: I think when the menu is toggled with the button, it doesn't set the state? So you have to tap it again?

PR #18227, opened 2023-01-23
python: add cockpit.print helper module

allisonkarlitskaya:
This is useful for debugging the bridge (either C or Python).

Start exploring:

```
PYTHONPATH=src python3 -m cockpit.print help
```


martinpitt (dismissed):
I didn't test this nor reviewed in detail -- but this doesn't get shipped and is a new script, so zero chance for causing trouble/regressions (other than flake8 perhaps). So, go ahead!

PR #18231, opened 2023-01-24
Python Bridge enable test timers

jelly:


src/cockpit/channels/filesystem.py
```diff
@@ -232,7 +232,7 @@ def do_inotify_event(self, mask, _cookie, name):
             self.send_message(event=event, path=self._path, tag=self._tag, type=type_)
 
     def do_identity_changed(self, fd, err):
-        logger.debug("do_identity_changed(%s): fd %i, err %s", self._path, fd, err)
+        logger.debug("do_identity_changed(%s): fd %s, err %s", self._path, str(fd), err)
```
jelly, line 235: @allisonkarlitskaya soooo, not sure about this. But doing `fd or '-'` also feels weird.

martinpitt (approved):
OOh, nice! Does that mean we can also drop some more (un)expected messages from testlib.py now? Would you like to try, or want me to?

PR #18232, opened 2023-01-24
shell: unset 'active' state also when closing the menu by clicking on the same iframe

### LINKED ISSUE #18182: mobile: When in responsive mode, the system navigation menu pops up again when unasked for

KKoukiou:
See https://github.com/cockpit-project/cockpit/issues/18182#issuecomment-1401641783

Fixes #18182

### DIFF
--- a/pkg/shell/nav.jsx
+++ b/pkg/shell/nav.jsx
@@ -17,11 +17,23 @@ export const SidebarToggle = () => {
     const [active, setActive] = useState(false);
 
     useEffect(() => {
-        const handleClickOutside = () => setActive(false);
+        /* This is a HACK for catching lost clicks on the pages which live in iframes so as to close dropdown menus on the shell.
+         * Note: Clicks on an <iframe> element won't trigger document.documentElement listeners, because it's literally different page with different security domain.
+         * However, when clicking on an iframe moves focus to its content's window that triggers the main window.blur event.
+         * Addionally, when clicking on an element in the same iframe make sure to unset the 'active' state of the 'System' dropdown selector.
+         */
+        const handleClickOutside = (ev) => {
+            if (ev.target.id == "nav-system-item")
+                return;
 
-        window.addEventListener("blur", handleClickOutside);
+            setActive(false);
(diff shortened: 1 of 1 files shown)


garrett (changes requested):
Looks great, except for the missing semicolon. Fix that and I approve. Thanks for the comment in the code too!

garrett:
Oh, I tested in Chrome & Firefox on Android, Safari on iOS. Seemed to work as expected! Thanks again!

martinpitt (approved):
Thanks @garrett ! I applied your suggestion, will squash it during landing.

garrett (approved):
With the semicolon, this gets an approval! Thanks again! :tada:

pkg/shell/nav.jsx
```diff
@@ -17,11 +17,23 @@ export const SidebarToggle = () => {
...
+            setActive(false);
+        };
 
-        window.addEventListener("blur", handleClickOutside);
+        ["blur", "click"].map(ev_type => window.addEventListener(ev_type, handleClickOutside));
 
-        return () => window.removeEventListener("blur", handleClickOutside);
+        return () => {
+            ["blur", "click"].map(ev_type => window.removeEventListener(ev_type, handleClickOutside));
+        };
```
cockpituous, line 36: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18232-20230124-161236-853896e9-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

PR #18233, opened 2023-01-24
python: correctly log possible None value

jelly:
Or do you prefer to special case this? if `fd is None`: lala

### DIFF
--- a/src/cockpit/channels/filesystem.py
+++ b/src/cockpit/channels/filesystem.py
@@ -232,7 +232,7 @@ class FsWatchChannel(Channel):
             self.send_message(event=event, path=self._path, tag=self._tag, type=type_)
 
     def do_identity_changed(self, fd, err):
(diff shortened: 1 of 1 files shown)


src/cockpit/channels/filesystem.py
```diff
@@ -232,7 +232,7 @@ def do_inotify_event(self, mask, _cookie, name):
             self.send_message(event=event, path=self._path, tag=self._tag, type=type_)
 
     def do_identity_changed(self, fd, err):
-        logger.debug("do_identity_changed(%s): fd %i, err %s", self._path, fd, err)
+        logger.debug("do_identity_changed(%s): fd %s, err %s", self._path, str(fd), err)
```
martinpitt, line 235: I'd use `repr`, but for `None` and numbers this is really good enough.

martinpitt (approved):
Thanks! Please trigger pybridge and fcos tests.

PR #18236, opened 2023-01-24
python: add a new packages.BridgeRule NamedTuple

allisonkarlitskaya:
This is the data from a "bridges" section in a manifest.json file, in an immutable and type-safe form.

This is going to be helpful for creating bridge routing rules.

Also add `safely.py` which contains some experimental typesafe functions meant for deconstructing untrusted JSON data.  I'm not 100% sure I like the approach, but we're going to need something like this, and this is the best I can think of for the time being.

### DIFF
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -26,7 +26,7 @@ import os
 import re
 
 from pathlib import Path
-from typing import ClassVar, Dict, List, Optional, Pattern, Tuple
+from typing import ClassVar, Dict, List, Optional, NamedTuple, Pattern, Sequence, Tuple
 
 from systemd_ctypes import bus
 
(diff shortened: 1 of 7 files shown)


martinpitt:
Triggering a pybridge test run.

src/cockpit/packages.py
```diff
@@ -168,10 +171,11 @@ def __init__(self, path):
...
 
-        self.name = self.manifest.get('name', path.name)
-        self.content_security_policy = None
-        self.priority = self.manifest.get('priority', 1)
-        self.bridges = self.manifest.get('bridges', [])
+        with safely.catch(str(manifest_path), logger.warning):
+            self.content_security_policy = safely.get(self.manifest, 'content-security-policy', str, '')
+            self.name = safely.get(self.manifest, 'name', str, path.name)
+            self.priority = safely.get(self.manifest, 'priority', int, 1)
+            self.bridges = safely.get_list(self.manifest, 'bridges', dict, [])
```
martinpitt: safely.catch() does not really catch an exception, it just logs and re-throws it. So, shouldn't these still be handled here, and a broken manifest gets ignored instead of crashing the bridge?

src/cockpit/safely.py
```diff
@@ -0,0 +1,74 @@
...
+@contextlib.contextmanager
+def catch(description: str, on_error: Callable) -> Generator[None, None, None]:
+    try:
+        yield
+    except (TypeError, KeyError) as err:
+        on_error(f'{description}: {err}')
+        raise
+
+
+def get(mapping: Mapping[str, object], key: str, expected_type: Type[T], *default: D) -> Union[T, D]:
```
martinpitt, line 35: Admittedly I don't understand the syntactical difference between `Type[T]` and `D` here. But it seems dubious to me that the default can be of a different type than `expected_type`. We don't use that anywhere in packages.py.

I'd much prefer `default: Optional[T]` and return an `Optional[T]`, to support `None` values. That should give  better type checking and avoid bugs.

Same for get_{list,dict} below.

src/cockpit/safely.py
```diff
@@ -0,0 +1,74 @@
...
+    try:
+        value = mapping[key]
+    except KeyError:
+        if default:
+            return default[0]
```
martinpitt, line 40: `*default` means that you can add arbitrarily many extra arguments, which will just be ignored. This is reflected here in this rather ugly `[0]`. IMHO this should rather be declared as a kwarg `default: Optional[T]=None`?

Same for get_{list,dict} below.

martinpitt (changes requested):
Thanks, this is (or rather, will hopefully be) a nice type strengthening!

src/cockpit/safely.py
allisonkarlitskaya: It's true that the only usecase I had in mind here was for `None`, and indeed that could be better covered by `Optional[T]`, but: in the case that the default value is not given, or the default value is something other than `None`, then this would cause the return type of this function to be something like `Optional[str]` when it could be `str`.

I struggle with the right thing to do here...

src/cockpit/safely.py
allisonkarlitskaya: The issue is that I want to be able to detect the difference between no default value being given (in which case it's a hard error if the key lookup fails) and a default value of `None` being given.

I'd do two separate functions for this, but then I'd need to build two completely separate hierarchies...

The other option I considered was using a different sentinel value — maybe that would indeed be better.  That presents a problem with getting the types working nicely, though.

src/cockpit/safely.py
martinpitt: Hmm, but if no `default` is given, the return type would be `Union[T, D]` with `D` being entirely undefined. Unless typing/mypy is clever enough to realize that there is no `D` and it would collapse to a `Union[T]` then? But this is still murky for the "given default" case -- being able to call it with `(map, "key", int, "foo")` just feels broken.

If you care for a strict return type, perhaps have a `.get()` which would `KeyError`, and a `get_default()` with a mandatory `default` argument?

allisonkarlitskaya:
This is a mess.  Getting mypy to do what we want is going to be hard without a whole lot of `@overload` and I find the amount of boilerplate really distasteful.

I think the future is going to involve some sort of use of TypedDict with automated checking.  Let's close this one for now.

PR #18243, opened 2023-01-25
Pybridge unknown file error

jelly:


src/cockpit/responses.py
```diff
@@ -0,0 +1,75 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from typing import Optional
+
+TEMPLATE: str = """
```
martinpitt, line 20: FTR, we touch this file so rarely, I'm totally fine with just embedding it here. For the C bridge we have ./src/common/fail.html and stringify/compile it in at build time, which is a tad nicer -- but such build hackery really isn't something that I'd like to do, and I'm 99% sure @allisonkarlitskaya agrees with me.

So this is just a temporary duplication, fail.html can go once we kill the C bridge.

martinpitt, line 20: Since this file is called "responses.py", not "errorpage.py", perhaps rename this to `ERROR_TEMPLATE`?

martinpitt (changes requested):
Thanks! So, that doesn't fix any test? :cry: I was hoping for `TestConnection.testCockpitDesktop`..

martinpitt:
FTR, "needs changes" only for the flake8 thing above. Renaming the template is bikeshedding, I'd accept the current version, too.

not sure if [this TestReauthorize failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18243-20230125-161556-d7da4625-fedora-37-pybridge/log.html#128-2) is related

jelly:
> Thanks! So, that doesn't fix any test? cry I was hoping for `TestConnection.testCockpitDesktop`..

It does not :( testCOckpitDesktop now just seems to error out later.

test/verify/check-superuser
```diff
@@ -387,7 +387,6 @@ class TestSuperuserOldShell(MachineCase):
 
 
 @skipImage("TODO: broken on Arch Linux", "arch")
-@skipDistroPackage()
```
martinpitt, line 390: No, that one needs to stay :-)

jelly: Don't commit late at night :)

martinpitt:
Dang, [this failed](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18243-20230126-073146-7796a70c-fedora-37-pybridge/log.html#336) -- looks like we'll need to move the decorator from the class to individual test methods

jelly:
> Dang, [this failed](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18243-20230126-073146-7796a70c-fedora-37-pybridge/log.html#336) -- looks like we'll need to move the decorator from the class to individual test methods

Ah! At a quick glance I thought test was a helper method, didn't sound like a test method at first glance.

src/cockpit/packages.py
```diff
@@ -419,7 +419,10 @@ def serve_manifests_js(self, channel):
 
     def serve_package_file(self, path, channel):
         package, _, package_path = path[1:].partition('/')
-        self.packages[package].serve_file(package_path, channel)
+        try:
+            self.packages[package].serve_file(package_path, channel)
```
allisonkarlitskaya, line 423: Two possibilities:
 - there's too much covered by this `try:`: split the dictionary lookup from the function call
 - or: if it's your intention to also handle `KeyError` thrown by `serve_file()` then you could probably also change a use of `.get()` there into `[]` (in `find_find()`) and drop the `found is None` case in `serve_file()` — possibly simply merging those two functions, then.  This would actually be a nice cleanup.

src/cockpit/responses.py
```diff
@@ -0,0 +1,75 @@
...
+</html>
+""" # noqa
+
+
+def format_error_response(status: int, message: Optional[str]) -> bytes:
+    if message is None:
+        if status == 404:
+            message = 'Not Found'
+        else:
+            message = f'{status} Failed'
```
allisonkarlitskaya, line 73: I suspect this is dead code.  There are exactly 3 places that call `http_error()` and all of them are explicitly `(404, "Not Found")`.  Maybe add some static typing hints to that function to indicate that a non-`None` `message: str` is expected?

While we're at it: I notice that one of the call sites for sending 404 is in the PackagesChannel in a `except FileNotFoundError:` block.  I'm fairly sure we can drop that.  It shouldn't ever happen anymore.

allisonkarlitskaya (changes requested):
See comments below.

With a bit of restructuring we could move to a place where we only have a single code path that emits 404 errors.  That would be a nice cleanup.

src/cockpit/responses.py
```diff
@@ -0,0 +1,75 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from typing import Optional
+
+ERROR_TEMPLATE: str = """
```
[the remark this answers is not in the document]
allisonkarlitskaya: So `importlib.resources` is new in 3.7.  We probably can't use that.

I tested this:

```python
__loader__.get_data(os.path.dirname(__file__) + '/fail.html'))
```

and it seems to work with both files and beipack.  You end up with `bytes`.

martinpitt:
Closing in favor of #18253

PR #18248, opened 2023-01-26
test: Filter out "dark" layout after reading custom layouts

mvollmer:
This way external projects can still use a browser-layouts.json file that includes a "dark" themed variant.


martinpitt (approved):
Nice, thank you! Then we can revert these hacks in podman and machines.

PR #18249, opened 2023-01-26
improve static typing in cockpit.channels.metrics

allisonkarlitskaya:


src/cockpit/channels/metrics.py
```diff
@@ -149,6 +152,9 @@ def send_updates(self, samples: Dict[str, Any], last_samples: Dict[str, Any]):
                 else:
                     data.append([val for val in value.values()])
             else:
+                assert not isinstance(value, dict)
+                assert not isinstance(old_value, dict)
```
jelly, line 156: This breaks the test and this is SUPER fun,

So accessing `last_samples[metricinfo.desc.name]` causes it to be set to the default value a dict. Which means it is always set to `{}` if it's not initialised yet.

jelly:
So, I am ready for a review here @allisonkarlitskaya but you won't be able to approve it :)

src/cockpit/channels/metrics.py
```diff
@@ -120,35 +120,42 @@ def sample(self):
             sampler.sample(samples)
         return samples
 
-    def calculate_sample_rate(self, value: Any, old_value: Optional[Any]):
+    def calculate_sample_rate(self, value: float, old_value: Optional[float]) -> Union[float, bool]:
```
allisonkarlitskaya, line 123: That's actually `Literal[False]`, isn't it?

src/cockpit/channels/metrics.py
```diff
@@ -120,35 +120,42 @@ def sample(self):
...
         if old_value is not None and self.last_timestamp:
             return (value - old_value) / (self.next_timestamp - self.last_timestamp)
         else:
             return False
 
-    def send_updates(self, samples: Dict[str, Any], last_samples: Dict[str, Any]):
-        data = []
+    def send_updates(self, samples: Samples, last_samples: Samples):
+        data: List[Union[float, List[Optional[Union[float, bool]]]]] = []
```
allisonkarlitskaya, line 130: I think `Optional[Union[float, Literal[False]]` probably reads better as `Union[float, Literal[False], None]`, but that's taste.  Would you say "no change" exists at a semantically higher level than "failed to calculate"?  Maybe...

These types are getting big.  I wonder if it makes sense to define some symbolic names...

allisonkarlitskaya (commented):
I mean, if this drops `Any` and still keeps mypy happy, then it's already a win, in my books.

Some comments below for possible tweaks.

I wonder if we want to move to having a more explicit `Samples` type.  We've discussed it before... it could have separate interfaces for instanced and uninstanced values.  That said, this is actually not what you're dealing with here, but rather some of the more complicated value types that get shuffled around on the way to being sent to the client....

src/cockpit/channels/metrics.py
jelly: Yes, but Python 3.6

jelly:
Let's also fix
```
src/cockpit/samples.py:Samples = DefaultDict[str, Any]
```

jelly:
I think I want to merge it as is, then iterate later over it depending on if we support 3.6 or not. Then I can refactor it again and add the `Literal[false]`.

The other `Any` occurrence is some mypy issue which @martinpitt found. Triggered tests to see if it's all good.

allisonkarlitskaya (commented):
To be quite explicit: if I could approve this, I would.

jelly (approved):

jelly:
testing-farm:fedora-36-x86_64 flake is unrelated.

PR #18252, opened 2023-01-26
test: Fix session cleanup for rawhide

martinpitt:
In Fedora rawhide this started to fail often with

    Could not kill session: No such file or directory

This is okay -- it's an inherent TOCTOU race condition, sessions can get cleaned up during the loop. This is some late fallout from .execute() getting strict [1], it just happened to (mostly) work until now.

[1] https://github.com/cockpit-project/bots/commit/b7d0edba5d02

----

This avoids occasional failures like [this](https://artifacts.dev.testing-farm.io/b686f17b-c254-4e0a-a76f-d06304392c61/#artifacts-Fedora-Rawhide:x86_64:/plans/all/optional), or the [complete mess](https://artifacts.dev.testing-farm.io/01a177c2-f55c-4471-8102-9eb75cfef9d5/) that breaks the current [284 rawhide update](https://bodhi.fedoraproject.org/updates/FEDORA-2023-44a632e396). Recently, all our rawhide packit tests in cockpit became red due to this.

FTR, I can locally reproduce this with `tmt run`, but not with our own fedora-rawhide image.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1588,6 +1588,9 @@ class MachineCase(unittest.TestCase):
                                         rm -rf /run/user/$u
                                     done""")
 
+            # Restart logind to mop up empty "closing" sessions
+            self.machine.execute("systemctl restart systemd-logind")
+
             # Terminate all other Cockpit sessions
             sessions = self.machine.execute("loginctl --no-legend list-sessions | awk '/web console/ { print $1 }'").strip().split()
             for s in sessions:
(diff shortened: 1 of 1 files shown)


martinpitt:
[still an utter mess](https://artifacts.dev.testing-farm.io/95389af0-ff28-4aa0-abcc-06b9307100f0/). It seems the kill-session is now just broken in rawhide? I don't get this effect in a local tmt run, this needs a bigger debugging hammer..

martinpitt:
OK, [that](https://artifacts.dev.testing-farm.io/7e49d0f0-1095-42ca-adae-023cfb1394f9/) looks much better. Down to `TestAutoUpdates.testWithAvailableUpdates` which fails for an entirely different reason. However, that is a race condition, and I have a hard time reproducing it locally. I wanted to FP this anyway to run tests on all our OSes, to be sure.

jelly (approved):

PR #18254, opened 2023-01-27
python: Read files in chunks in a thread in fsread1

martinpitt:
The previous `open().read()` approach fails for very large files or endless streams, as it was reading the entire file into memory. This would also block the whole bridge if the open() takes a long time (think NFS!).

Therefore, move the whole file operation into a separate thread, and read/send the data in 1 MiB chunks.

----

This is a first step towards fixing `TestConnection.testFlowControl`. At least I can now manually run /playground/speed and read /dev/vda in a *normal* channel, at least for a while. This requires `chmod 666 /dev/vda` in the VM, or logging in as root. It still OOMs after maybe 15 seconds, but at least it's getting data. I suspect that we are missing flow control, and the bridge queues up too much data that it can't send out?

The good news is that I get about 50 MB/s with the C bridge, and at least as much with the Python bridge (between 50 and 60 MB/s). But the "sideband channel" still hangs and then OOMs.

While I was at it: "spawn test" also works. It's a tad slower than with the C bridge, but good enough. "echo" test is comparable with both.

### DIFF
--- a/src/cockpit/channels/filesystem.py
+++ b/src/cockpit/channels/filesystem.py
@@ -15,9 +15,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import asyncio
 import logging
 import os
 import random
+import threading
 
 from systemd_ctypes import PathWatch
 from systemd_ctypes.inotify import Event as InotifyEvent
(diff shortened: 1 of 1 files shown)


martinpitt:
@allisonkarlitskaya says that this isn't what she had in mind. I'm not very attached to this either -- for the most part that was a fun 20 minute experiment for me to let the py and C bridges race and compare speeds 🏎 🏁 . I was quite positively surprised how well the pybridge does!

PR #18256, opened 2023-01-27
test: Robustify encrypt_root()

martinpitt:
On current CentOS 8 Stream, this fails on all kinds of timeouts and 
ENOSPC. Make this more robust:

 * `cryptsetup luksFormat` uses oodles of RAM by default for its PBKDF
   function. Limit it to at most 300 kB. 

 * Increase the /new-boot size partition to 500 MB, as 300 MB is too 
   small for recent RHEL 8.

 * Exclude /new-root from tar to avoid recursive file inclusion.

 * Add `-set -x` for easier interactive debugging what exact command is
   hanging at the moment. Splitting the script is not easy, as it
   defines several variables which are being used throughout the script,
   but that already helps.

----

This fixes https://github.com/cockpit-project/bots/pull/4320.

### DIFF
--- a/test/common/storagelib.py
+++ b/test/common/storagelib.py
@@ -577,11 +577,12 @@ MakeDirectory=yes
         dev = "/dev/" + info["dev"]
         wait(lambda: m.execute(f"test -b {dev} && echo present").strip() == "present")
         m.execute(f"""
+set -x
 parted -s {dev} mktable msdos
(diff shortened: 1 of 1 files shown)


martinpitt:
Nice, test against the new c8s image worked as well :+1:

mvollmer (approved):

PR #18257, opened 2023-01-27
python: add flow control for fsread1

allisonkarlitskaya:


test/image-prepare
```diff
@@ -213,6 +213,7 @@ def main():
     parser = argparse.ArgumentParser(
         description='Prepare testing environment, download images and build and install cockpit',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('-n', '--no-c', action='store_true', help='Skip installing the C code')
```
martinpitt, line 216: Hmm, couldn't we just make `--python` do that? That will remove the /usr/bin/cockpit-bridge binary anyway, so the rpm won't get us anything.

test/image-prepare
```diff
@@ -228,20 +229,21 @@ def main():
...
 
-    if not args.quick:
-        customize += validate_packages()
+        if not args.quick:
+            customize += validate_packages()
 
-    # post build/install test preparation
-    customize += ["--script", os.path.join(TEST_DIR, "vm.install")]
+        # post build/install test preparation
+        customize += ["--script", os.path.join(TEST_DIR, "vm.install")]
```
martinpitt, line 246: This still needs to run for `--python`. Otherwise, if `--no-c` is always just meant to hack an existing image, then that's what `scp -r` is for, or `image-customize --upload`?

src/cockpit/channel.py
```diff
@@ -161,9 +170,43 @@ def done(self):
...
+            logger.warning("Got wild pong on channel %s", self.channel)
+            return
+
+        self._out_window = message['sequence'] + Channel.CHANNEL_FLOW_WINDOW
+        if self._out_sequence < self._out_window:
+            self.do_thaw_send(min(self._out_window - self._out_sequence, Channel.CHANNEL_FLOW_PING))
+
+    def do_thaw_send(self, n_bytes: int) -> None:
+        """Called to indicate that the channel may now send n_bytes more bytes."""
+        pass  # temporary.  should raise NotImplementedError
```
martinpitt, line 208: Forgotten to actually do that? The subclasses implement it already, apparently.

src/cockpit/channel.py
```diff
@@ -397,3 +397,35 @@ def do_data(self, data):
...
+class GeneratorChannel(Channel):
+    """A trivial Channel subclass for sending data from a generator with flow control.
+
+    Calls the .get_data() generator with the options from the open message and
+    sends the data which it yields.  If the generator returns a value it will
+    be used for the close message.
+    """
+    __generator: Generator[bytes, int, Optional[Dict[str, object]]]
+
+    def get_data(self, size_hint: int, options: dict[str, object]) -> Generator[bytes, int, Optional[Dict[str, object]]]:
```
martinpitt, line 411: Could the return type get its own type name? I think mypy/typing allow that. It's not quite trivial, and gets repeated in fsread1.

src/cockpit/channels/filesystem.py
```diff
@@ -77,39 +79,35 @@ def do_open(self, options):
...
-            except OSError:
-                raise ChannelError('internal-error')
+            with open(options['path'], 'rb') as filep:
+                buf = os.stat(filep.fileno())
+                if max_read_size is not None and buf.st_size > max_read_size:
+                    raise ChannelError('too-large')
 
-            if 'binary' not in options:
-                data = data.replace(b'\0', b'').decode('utf-8', errors='ignore').encode('utf-8')
+                while (data := filep.read(size_hint)):
```
martinpitt, line 97: Let's please avoid walruses. We still have a few, but let's not increase that problem.

martinpitt: Also, can we make that a `.read1()`. I don't think it makes much sense to read multiple chunks without sending  them out in between. cockpit.file() can read pipes, files from NFS, etc.

martinpitt, line 97: Finally, `size_hint` is documented to allow 0 or negative, and that shouldn't just be stuffed into .read{,1}, but needs to be handled separately.

martinpitt (changes requested):
Exciting, thanks! Looking foward to using that for http as well.

src/cockpit/channels/filesystem.py
martinpitt: Hmm, on second sight this is not true -- that documentation applied to `send_data()`. It also wouldn't make sense to even call get_data() with ≤ 0, because what could it even do with that information.. (sleep? yield None? all of that can be done by the parent class if necessary)

martinpitt: ... but I still don't understand this. What exactly happens in line 101 when it calls `size_hint  = yield data`? What kind of return values does that get, and when?

test/image-prepare
allisonkarlitskaya: By default we install the C bridge and webpack parts anyway.  That was done so that if we make some changes to `cockpit-ws` or (more likely) `pkg/` we get to test those changes against the Python bridge, instead of using the old distro package.  That seems like a reasonable default and I'd like to keep that, even with `--python`.

test/image-prepare
allisonkarlitskaya: Ya.  This is really for hacking an existing image (ie: with `-o`).  And indeed, that could be replaced by an `image-customize` invocation, but the logic for building the wheel is also in this script...

Maybe we need `tools/make-wheel` alongside our `tools/make-dist`.  But that's sort of what I was trying to avoid.

src/cockpit/channel.py
allisonkarlitskaya: No.  I wanted to leave it this way.  I originally had the `raise NotImplementedError` but it was crashing.  Consider that many of our channels still subclass directly `Channel` and don't implement this.

src/cockpit/channel.py
allisonkarlitskaya: It could, although I don't consider this particular type to be so complicated.  Generators are just like that :sweat_smile:

src/cockpit/channels/filesystem.py
allisonkarlitskaya: OK.

allisonkarlitskaya: This is what comes from the `__generator.send()` call in the `GeneratorChannel`.  We get the new size_hint here and will use it to make a read() and then yield the result, which becomes the return value of `.send()` back in the `GeneratorChannel`

allisonkarlitskaya: Note: the reason we do this is slightly non-obvious.  It's not expected that the size hint will ever change, so the generator could keep reading values of the original size.  But: if for some reason we have an incomplete read or the size changes (for example, through our character set conversion for non-binary channels) then we'll have sent a partial block.  We can use the size_hint (ie: up to the remainder of the block size) to get back on track again.

That said: I just notice that this value comes from the return value of `.send_data()`, and the required logic isn't actually implemented there.  Oops.

test/image-prepare
martinpitt: Hmm, I don't quite understand this TBH. We always want to install cockpit-ws and cockpit-session and such, I understood this option as "no C bridge", not "no C parts at all". Note that image-prepare isn't a tool for iterating (it's waay too slow for that)

test/image-prepare
martinpitt: OK, it seems you have a very different way of updating images for iterating than I (I really just use `scp -r src/cockpit ..`).

test/image-prepare
martinpitt: Anyway, if that were a permanent option I'd object harder, but as hopefully both --python and --no-c will go away in a few months, I can stomach this hack.

src/cockpit/channel.py
martinpitt: I see. Can you please update the comment?

src/cockpit/channel.py
martinpitt: Heh, ok -- not important either way, feel free to just resolve.

test/image-prepare
allisonkarlitskaya: I'm happy to drop the `--no-c` commit (although I think the other one still makes sense).

src/cockpit/channels/filesystem.py
allisonkarlitskaya: That said, I realized over the weekend that I remember this wrong.  That was the idea of my original approach, but I changed tack, and the new approach actually doesn't depend on this calculation at all.  It only depends on the sender sending "reasonable-sized" chunks, ideally as close to the "ping" size as possible.

allisonkarlitskaya: ...which also means that we can skip the complicated theatrics with the bidirectional generator, and ask the generator to always send blocks of the same size.  We could also replace the integer return value of `.send_data()` with a `bool` meaning "is it OK to send more".

allisonkarlitskaya: Did that.

src/cockpit/channel.py
allisonkarlitskaya: I added it.

src/cockpit/channel.py
allisonkarlitskaya: Done.

test/image-prepare
allisonkarlitskaya: This comment got dropped.

src/cockpit/channel.py
```diff
@@ -397,3 +397,35 @@ def do_data(self, data):
...
+        raise NotImplementedError
+
+    def do_open(self, options: Dict[str, object]) -> None:
+        self.__generator = self.get_data(Channel.CHANNEL_FLOW_WINDOW, options)
+        self.iterate(None)  # first Generator.send() call needs to be with None
+
+    def iterate(self, n_bytes: Optional[int]) -> None:
+        try:
+            while n_bytes is None or n_bytes > 0:
```
allisonkarlitskaya, line 420: This could infinite loop if the generator keeps yielding empty blocks.  That would happen is you tried to fsread `/dev/zero`, for example, since we drop `'\0'` characters.

martinpitt: We drop `\0`? That feels like quite a grave bug in itself? Why is that, python bytearrays should be fine with that?

```
❱❱❱ python3 -c 's = b"a\0b"; print(len(s), s)'
3 b'a\x00b'
```

(unrelated of course, but this should be tracked then)

src/cockpit/channels/filesystem.py
```diff
@@ -77,39 +79,38 @@ def do_open(self, options):
...
-            self.send_data(data)
         except FileNotFoundError:
-            logger.debug('  ...file not found!')
-        self.done()
-        self.close(tag=tag)
+            return {'tag': '-'}
+        except PermissionError:
+            raise ChannelError('access-denied')
+        except OSError:
+            raise ChannelError('internal-error')
```
martinpitt, line 107: While you are at it: Can we send the contents of the OSError as message in the channel error? or at least `logger` it? Ignoring it entirely makes things hard to debug. This is not a regression, thus can also be a follow-up.

src/cockpit/channels/filesystem.py
```diff
@@ -77,39 +79,38 @@ def do_open(self, options):
...
+                if max_read_size is not None and buf.st_size > max_read_size:
+                    raise ChannelError('too-large')
+
+                while True:
+                    data = filep.read1(size_hint)
+                    if data == b'':
+                        break
+                    logger.debug('  ...sending %d bytes', len(data))
+                    if not binary:
+                        data = data.replace(b'\0', b'').decode('utf-8', errors='ignore').encode('utf-8')
```
martinpitt, line 97: .. Oh, I suppose *this* is the place you meant which swallows `\0`? That seems okay -- how would it loop infinitely here? it still consumes reading the file blocks, after all.

src/cockpit/channels/filesystem.py
```diff
@@ -77,39 +79,38 @@ def do_open(self, options):
...
+        max_read_size = options.get('max_read_size')
+
+        logger.debug('Opening file "%s" for reading', options['path'])
 
-            if 'binary' not in options:
-                data = data.replace(b'\0', b'').decode('utf-8', errors='ignore').encode('utf-8')
+        try:
+            with open(options['path'], 'rb') as filep:
+                buf = os.stat(filep.fileno())
+                if max_read_size is not None and buf.st_size > max_read_size:
```
martinpitt, line 88: Note to self: This works ok for non-regular files. `os.stat('/dev/null').st_size == 0`.

src/cockpit/channel.py
```diff
@@ -394,3 +394,29 @@ def do_data(self, data):
...
+    """A trivial Channel subclass for sending data from a generator with flow control.
+
+    Calls the .do_yield_data() generator with the options from the open message
+    and sends the data which it yields.  If the generator returns a value it
+    will be used for the close message.
+    """
+    DataGenerator = Generator[bytes, None, Optional[Dict[str, object]]]
+    __generator: DataGenerator
+
+    def do_yield_data(self, size_hint: int, options: dict[str, object]) -> DataGenerator:
```
martinpitt, line 409: Do you actually plan to ever send something else than `Channel.CHANNEL_FLOW_WINDOW` as `size_hint`? If not, let's just drop this? When reading the implementation, I still keep asking myself "what kind of values do I need to handle for size_hint"?

martinpitt (changes requested):
Thanks! I'd still like to see some minor cleanups here, but this is much easier to understand now. Cheers!

src/cockpit/channel.py
allisonkarlitskaya: not `WINDOW`, but `PING`.  I think these constants are long overdue for renaming.

allisonkarlitskaya: I also just realized that the way `fslist1` is implemented is bonkers.  We're gonna send a ping after every single directory entry, which seems excessive.  This needs a bit of thinking.

martinpitt: The code really [calls it with FLOW_WINDOW](https://github.com/cockpit-project/cockpit/pull/18257/files#diff-50b095c0e61dea51dd6279554f093ec49005bb91f2e2078acc74a4d58eeba9b7R413), so I suppose you want to change tihs?

Do you just want to drop the fslist1 patch for now?

allisonkarlitskaya: Yes.

I have some vague plan in the future to support merging messages and supporting `str` and json as well.

martinpitt (approved):
Thank you! This looks rather nice now. I made a TODO for the outstanding [OSError details](https://github.com/cockpit-project/cockpit/pull/18257/commits/f128b1b3991481b4e0e1aede8c431a81fea6ddee#r1090499063).

src/cockpit/channels/filesystem.py
martinpitt: Sent to PR #18285

PR #18266, opened 2023-01-30
shell: python: load translations via manifest.js

### LINKED ISSUE #18224: Implement shell l10n in the python bridge
We're going to abandon the `*/po.js` globbing approach for collecting and serving the translations of all of the packages to the shell for several reasons:

 - globbing doesn't work well in the presence of gzipped data and requires the data to be decompressed by the bridge before it can be concatenated with non-compressed data.  gzip streams can be concatenated, but for simplicity, the C bridge decompresses even if all of the files are gzipped.  We want to avoid gzip decompression in the Python bridge

 - the globbing mechanism itself is a bit too big of a hammer for being implemented as a general purpose feature, when it's used for exactly one path `*/po.js` and from exactly one place (the shell)

 - the amount of data sent is very large: the shell doesn't need the full translation of every page — only the strings that appear in the manifest.

There are many possible solutions to this problem:
 - have the bridge parse the `po.js` files and extract the strings relevant to translating the manifest
   - can be done because the bridge reads the manifests and can build a list of those strings
   - could also be done using msgctxt to hint which strings are from manifests
   - could also be added to the `po.js` file in some other way by `cockpit-po-plugin` based on the file references informaiton (ie: is `manifest.json` mentioned?)
     - possibly per-string, in the unused `[0]` field of the destination of the translation
     - possibly by grouping the manifest-relevant and manifest-irrelevant strings onto separate lines in the `po.js` file — the bridge would only read the first line in order to get strings for the manifest.  This would greatly simplify the "parsing"
 - all translations could be merged into the built `manifest.json` file by webpack
   - this file is currently copied directly over from `pkg/` to `dist/` during the build.  Adding the translations might require creating a new webpack plugin.
   - it would also require changes on the client side unless the bridge converted the data in the manifest into `cockpit.locale()` calls
    - it would result in too much data being sent in another way: instead of the entire translation for all pages in the current language, we'd get only the manifest-relevant strings, but we'd get them in every language, which is a different kind of wastefulness
  - modify `cockpit-po-plugin` to write the manifest-relevant translations to separate files like `apps/po.manifest.de.js`/`.json`/`.gz` for each package and each language
    - this would double the number of translation files on the disk
    - if the standard says that all of these files need to be compressed, then we can compress them all and serve the concatenated compressed stream
    - if this is uncompressed JSON data, we can merge it into the `/manifests.js` that we already serve.  this would require client-side changes, or a bit of code added to the served manifests blob.
    - if this is uncompressed JS data we can simply concatenate it to the `/manifests.js` that we already serve

After a quick IRC conversation, we seem to be consolidating around the option to have `cockpit-po-plugin` split out the manifest-related translations into separate per-language files (ie: the last option).  It's not clear if those will be in JSON or JS form, but either way, they'll probably be stored uncompressed, minimally processed by the bridge (either merging of JSON dictionaries, as we already do for the `manifest.json` files, or straight-up file concatenation), and served as part of `/manifests.js`.

jelly:
Split the po files into po.$lang.js and po.manifest.$lang.js so the shell can only serve the translated manifests. As previously the shell loaded it's own translations via globbing for the Python bridge we need to explicitly load the po.js file.

Closes #18224

* [ ] Handle comment about inverse filter predicate
* [x] Drop duplication

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -44,10 +44,7 @@ function get_plural_expr(statement) {
     return expr;
 }
 
-function buildFile(po_file, subdir, webpack_module, webpack_compilation) {
-    if (webpack_compilation)
-        webpack_compilation.fileDependencies.add(po_file);
-
+function buildFile(po_file, subdir, webpack_module, webpack_compilation, out_path, filter) {
     return new Promise((resolve, reject) => {
         // Read the PO file, remove fuzzy/disabled lines to avoid tripping up the validator
         const po_data = fs.readFileSync(po_file, 'utf8')
@@ -81,6 +78,9 @@ function buildFile(po_file, subdir, webpack_module, webpack_compilation) {
                 if (translation.comments.flag?.match(/\bfuzzy\b/))
                     continue;
 
+                if (!references.some(filter))
+                    continue;
+
                 const key = JSON.stringify(context_prefix + msgid);
                 // cockpit.js always ignores the first item
                 chunks.push(`,\n ${key}: [\n  null`);
@@ -95,8 +95,6 @@ function buildFile(po_file, subdir, webpack_module, webpack_compilation) {
         const wrapper = config.wrapper?.(subdir) || DEFAULT_WRAPPER;
         const output = wrapper.replace('PO_DATA', chunks.join('')) + '\n';
 
-        const lang = path.basename(po_file).slice(0, -3);
-        const out_path = (subdir ? (subdir + '/') : '') + 'po.' + lang + '.js';
         if (webpack_compilation)
             webpack_compilation.emitAsset(out_path, new webpack_module.sources.RawSource(output));
         else
@@ -115,9 +113,20 @@ function init(options) {
 
 function run(webpack_module, webpack_compilation) {
     const promises = [];
-    config.subdirs.map(subdir =>
-        promises.push(...get_po_files().map(po_file =>
-            buildFile(po_file, subdir, webpack_module, webpack_compilation))));
+    for (const subdir of config.subdirs) {
+        for (const po_file of get_po_files()) {
+            if (webpack_compilation)
+                webpack_compilation.fileDependencies.add(po_file);
+            const lang = path.basename(po_file).slice(0, -3);
+            promises.push(Promise.all([
+                // Separate translations for the manifest.json file and normal pages
+                buildFile(po_file, subdir, webpack_module, webpack_compilation,
+                          `${subdir}/po.${lang}.js`, str => !str.includes('manifest.json')),
+                buildFile(po_file, subdir, webpack_module, webpack_compilation,
+                          `${subdir}/po.manifest.${lang}.js`, str => str.includes('manifest.json'))
+            ]));
+        }
+    }
     return Promise.all(promises);
 }
 
--- a/pkg/shell/index.html
+++ b/pkg/shell/index.html
@@ -8,7 +8,10 @@
     <link href="../../static/branding.css" rel="stylesheet" />
     <script src="../base1/cockpit.js"></script>
     <script src="../manifests.js"></script>
+    <!-- HACK: C bridge loads translations via glob and Python via manifest.js -->
+    <script src="../*/po.manifest.js"></script>
     <script src="../*/po.js"></script>
+    <script src="po.js"></script>
   </head>
   <body class="pf-v5-m-tabular-nums" hidden="true">
     <div id="main" class="page">
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -157,7 +157,8 @@ class Manifest(JsonObject):
 
 
 class Package:
-    PO_JS_RE: ClassVar[Pattern] = re.compile(r'po\.([^.]+)\.js(\.gz)?')
+    # For po{,.manifest}.js files, the interesting part is the locale name
+    PO_JS_RE: ClassVar[Pattern] = re.compile(r'(po|po\.manifest)\.([^.]+)\.js(\.gz)?')
 
     # immutable after __init__
     manifest: Manifest
@@ -166,7 +167,7 @@ class Package:
     priority: int
 
     # computed later
-    translations: Optional[Dict[str, str]] = None
+    translations: Optional[Dict[str, Dict[str, str]]] = None
     files: Optional[Dict[str, str]] = None
 
     def __init__(self, manifest: Manifest):
@@ -186,7 +187,7 @@ class Package:
             return
 
         self.files = {}
-        self.translations = {}
+        self.translations = {'po.js': {}, 'po.manifest.js': {}}
 
         for file in self.path.rglob('*'):
             name = str(file.relative_to(self.path))
@@ -195,13 +196,19 @@ class Package:
 
             po_match = Package.PO_JS_RE.fullmatch(name)
             if po_match:
-                locale = po_match.group(1)
+                basename = po_match.group(1)
+                locale = po_match.group(2)
                 # Accept-Language is case-insensitive and uses '-' to separate variants
                 lower_locale = locale.lower().replace('_', '-')
-                self.translations[lower_locale] = name
+
+                self.translations[f'{basename}.js'][lower_locale] = name
             else:
                 basename = name[:-3] if name.endswith('.gz') else name
                 self.files[basename] = name
+
+        # support old cockpit-po-plugin which didn't write po.manifest.??.js
+        if not self.translations['po.manifest.js']:
+            self.translations['po.manifest.js'] = self.translations['po.js']
 
     def get_content_security_policy(self) -> str:
         policy = {
@@ -236,21 +243,21 @@ class Package:
 
         return Document(path.open('rb'), content_type, content_encoding, content_security_policy)
 
-    def load_translation(self, locales: List[str]) -> Document:
+    def load_translation(self, path: str, locales: List[str]) -> Document:
         self.ensure_scanned()
         assert self.translations is not None
 
         # First check the locales that the user sent
         for locale in locales:
             with contextlib.suppress(KeyError):
-                return self.load_file(self.translations[locale])
+                return self.load_file(self.translations[path][locale])
 
         # Next, check the language-only versions of variant-specified locales
         for locale in locales:
             language, _, region = locale.partition('-')
             if region:
                 with contextlib.suppress(KeyError):
-                    return self.load_file(self.translations[language])
+                    return self.load_file(self.translations[path][language])
 
         # We prefer to return an empty document than 404 in order to avoid
         # errors in the console when a translation can't be found
@@ -261,9 +268,9 @@ class Package:
         assert self.files is not None
         assert self.translations is not None
 
-        if path == 'po.js':
+        if path in self.translations:
             locales = parse_accept_language(headers)
-            return self.load_translation(locales)
+            return self.load_translation(path, locales)
         else:
             return self.load_file(self.files[path])
 
@@ -496,13 +503,15 @@ class Packages(bus.Object, interface='cockpit.Packages'):
         for name, package in self.packages.items():
             if name in ['static', 'base1']:
                 continue
+
             # find_translation will always find at least 'en'
-            translation = package.load_translation(locales)
+            translation = package.load_translation('po.manifest.js', locales)
             with translation.data:
                 if translation.content_encoding == 'gzip':
                     data = gzip.decompress(translation.data.read())
                 else:
                     data = translation.data.read()
+
             chunks.append(data)
 
         chunks.append(b"""
--- a/test/pytest/test_packages.py
+++ b/test/pytest/test_packages.py
@@ -158,3 +158,33 @@ def test_condition_hides_priority(pkgdir):
     assert packages.packages['basic'].manifest['description'] == 'standard package'
     assert packages.packages['basic'].manifest['requires'] == {'cockpit': "42"}
     assert packages.packages['basic'].priority == 1
+
+
+def test_translation(pkgdir):
+    # old style: make sure po.de.js is served as fallback for manifest translations
+    make_package(pkgdir, 'one')
+    (pkgdir / 'one' / 'po.de.js').write_text('eins')
+
+    # new style: separated translations
+    make_package(pkgdir, 'two')
+    (pkgdir / 'two' / 'po.de.js').write_text('zwei')
+    (pkgdir / 'two' / 'po.manifest.de.js').write_text('zwo')
+
+    packages = Packages()
+
+    # make sure we can read a po.js file with language fallback
+    document = packages.load_path('/one/po.js', {'Accept-Language': 'es, de'})
+    assert '/javascript' in document.content_type
+    assert document.data.read() == b'eins'
+
+    # make sure we fall back cleanly to an empty file with correct mime
+    document = packages.load_path('/one/po.js', {'Accept-Language': 'es'})
+    assert '/javascript' in document.content_type
+    assert document.data.read() == b''
+
+    # make sure the manifest translations get sent along with manifests.js
+    document = packages.load_path('/manifests.js', {'Accept-Language': 'de'})
+    contents = document.data.read()
+    assert b'eins\n' in contents
+    assert b'zwo\n' in contents
+    assert b'zwei\n' not in contents


pkg/lib/cockpit-po-plugin.js
```diff
@@ -91,6 +89,9 @@ module.exports = class {
                     if (!references.some(str => str.startsWith(`pkg/${this.subdir}`)))
                         continue;
 
+                    if (references.every(filter))
```
martinpitt, line 92: I feel like this is the wrong way around. It's apparently consistent (the caller also specifies the "correct" filter), but it'd be more natural to understand if "filter" gets you the things you want, instead of the things you don't want.

I'm trying to validate this in my head what happens if a string has a reference in boh a .jsx and the .manifest file -- then it should appear in both, right? An `any(positive_filter)` would make this clearer.

martinpitt (changes requested):
This looks good, but I'd invert the meaning of "filter" to make this more readable. Also, can you please add an integration test for menu searching a translated keyword? Obviously we don't do this, but given how many moving parts are involved for that, it really should have one.

jelly:
> This looks good, but I'd invert the meaning of "filter" to make this more readable. Also, can you please add an integration test for menu searching a translated keyword? Obviously we don't do this, but given how many moving parts are involved for that, it really should have one.

This is tested in testMenuSearch, we switch to german and search for "Dien" and expect to see "Dienste" before this PR we just loaded /*/po.js.

martinpitt:
[testMenuSearch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18266-20230202-142654-4e13b10a-fedora-37-pybridge/log.html#127-2) still fails :cry:

jelly:
> [testMenuSearch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18266-20230202-142654-4e13b10a-fedora-37-pybridge/log.html#127-2) still fails cry

I suppose this is a race condition, locally I can trigger this. Locally nothing is translated so either the python-bridge returns the wrong po.js or switching translations is buggy.

jelly:
Ok this is hopeless, still fails in CI. No idea why the shell does not reload. I can add no-test I suppose and add a timeout.

src/cockpit/packages.py
```diff
@@ -223,11 +228,17 @@ def add_file(self, item: Path, checksums: List['hashlib._Hash']) -> None:
         # file then add it to the list of translations, by locale name.
         # Otherwise, add it to the normal files list (after stripping '.gz').
         po_match = Package.PO_JS_RE.fullmatch(rel)
+        po_manifest_match = Package.PO_MANIFEST_RE.fullmatch(rel)
```
allisonkarlitskaya, line 231: Why not do this as a single RE with the `.manifest` optional?  Then you could detect which of the two cases it is from the common `if:` block, after the match.

It might also be nice to call it `po-manifest` or so, to avoid any ambiguity about what follows `po.` (ie: the string `manifest`  can never be interpreted as a locale).

src/cockpit/packages.py
```diff
@@ -148,13 +148,17 @@ class Package:
...
     PO_JS_RE: ClassVar[Pattern] = re.compile(r'po\.([^.]+)\.js(\.gz)?')
 
+    PO_MANIFEST_RE: ClassVar[Pattern] = re.compile(r'po\.manifest\.([^.]+)\.js(\.gz)?')
+
     # A built in base set of "English" translations
     PO_EN_JS: ClassVar[Entity] = b'', mimetypes.guess_type('po.js')
     BASE_TRANSLATIONS: ClassVar[Entities] = {'en': PO_EN_JS, 'en-us': PO_EN_JS}
 
     files: Entities
     translations: Entities
```
allisonkarlitskaya, line 159: drop the newline here

allisonkarlitskaya (commented):
You're missing a fallback in case we find no manifest translations files — we should support packages using the old po-plugin by sending the full translations.

jelly:
So we load translations via manifests.js now on the Python bridge, but the C bridge doesn't do this. The easiest workaround is to also glob `/po/*/manifest.po.js`.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -71,6 +68,9 @@ function buildFile(po_file, webpack_module, webpack_compilation) {
                 if (!references.some(str => str.startsWith(`pkg/${config.subdir}`) || str.startsWith('src')))
                     continue;
 
+                if (references.every(filter))
```
allisonkarlitskaya, line 71: For the record, I also find this extremely confusing.  This should rather be a predicate which takes the translations which it matches.  ie:  please rewrite that as `if (!references.some(filter)) continue;` and adjust accordingly.

pkg/shell/index.html
```diff
@@ -8,7 +8,10 @@
     <link href="../../static/branding.css" rel="stylesheet">
     <script src="../base1/cockpit.js"></script>
     <script src="../manifests.js"></script>
+    <!-- HACK: C bridge loads translations via glob and Python via manifest.js -->
+    <script src="../*/po.manifest.js"></script>
     <script src="../*/po.js"></script>
+    <script src="po.js"></script>
```
allisonkarlitskaya, line 14: Of course!  Now that the shell no longer gets its own translations sent via manifests.js, you need to explicitly request them.  Subtle!

src/cockpit/packages.py
```diff
@@ -230,10 +232,14 @@ def add_file(self, item: Path, checksums: List['hashlib._Hash']) -> None:
...
-            locale = po_match.group(1)
+            manifest_po = po_match.group(1)
+            locale = po_match.group(2)
             # Accept-Language is case-insensitive and uses '-' to separate variants
             lower_locale = locale.lower().replace('_', '-')
-            self.translations[lower_locale] = data, guessed_type
+            if manifest_po is None:
+                self.translations[lower_locale] = data, guessed_type
+            else:
+                self.manifest_translations[lower_locale] = data, guessed_type
```
allisonkarlitskaya, line 242: Feels like it could use a bit more factoring out, but to be honest, I'm not sure I'd like the result...

allisonkarlitskaya (changes requested):
This is really starting to take shape!

1) I agree with Martin.  Please fix the filter thing.
2) There's still no apparent fallback for the Python bridge supporting packages built with the old plugin which doesn't split out manifest translations.

jelly:
Gonna let pybridge do a full test run first to see if this fixes the broken tests locally it was fine.

jelly:
> Gonna let pybridge do a full test run first to see if this fixes the broken tests locally it was fine.

Ok ignoring code reviews...

TestPages.testBasic fails, shell is partially translated https://cockpit-logs.us-east-1.linodeobjects.com/pull-18266-20230221-081454-53d304d8-fedora-37-pybridge/TestPages-testBasic-fedora-37-127.0.0.2-2401-FAIL.png

TestPages.testMenuSearch fails, same issue, shell is partially translated https://cockpit-logs.us-east-1.linodeobjects.com/pull-18266-20230221-081454-53d304d8-fedora-37-pybridge/TestPages-testMenuSearch-fedora-37-127.0.0.2-2801-FAIL.png

These should be loaded via manifests.js, it is logical that the shell is translated as it is loaded via /po.js. So as we load it via manifests.js is this some funky caching issue.

jelly:
```
cockpit.packages-ERROR: loading manifests js for packagename performance for locales: ['en-us'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename users for locales: ['en-us'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename apps for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename kdump for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename metrics for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename network for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename updates for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename pcp for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename playground for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename selinux for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename shell for locales: ['de-de'], size: 87
cockpit.packages-ERROR: loading manifests js for packagename sosreport for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename ssh for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename storage for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename system for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename performance for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename users for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename apps for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename kdump for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename metrics for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename network for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename updates for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename pcp for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename playground for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename selinux for locales: ['de-de'], size: 0
cockpit.packages-ERROR: loading manifests js for packagename shell for locales: ['de-de'], size: 87
cockpit.packages-ERROR: loading manifests js for packagename sosreport for locales: ['de-de'], size: 0
```

So the translations are empty...

jelly:
From the log:

```
DEBUG: Executing command: ['/usr/bin/systemd-nspawn', '-q', '-M', '76e1daa866f7479686fd568d170da89a', '-D', '/var/lib/mock/fedora-37-x86_64/root', '-a', '--console=pipe', '--setenv=TERM=vt100', '--setenv=SHELL=/bin/bash', '--setenv=HOME=/builddir', '--setenv=HOSTNAME=mock', '--setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin', '--setenv=PROMPT_COMMAND=printf "\\033]0;<mock-chroot>\\007"', '--setenv=PS1=<mock-chroot> \\s-\\v\\$ ', '--setenv=LANG=C.UTF-8', '--resolv-conf=off', 'rpmbuild', '--help'] with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8', 'SYSTEMD_NSPAWN_TMPFS_TMP': '0', 'SYSTEMD_SECCOMP': '0'} and shell False
DEBUG: Usage: rpmbuild [OPTION...]
DEBUG: Build options with [ <specfile> | <tarball> | <source package> ]:
DEBUG:       -bp                        build through %prep (unpack sources and apply
DEBUG:                                  patches) from <specfile>
DEBUG:       -bf                        build through %conf (%prep, then %conf) from
DEBUG:                                  <specfile>
DEBUG:       -bc                        build through %build (%prep, %conf, then
DEBUG:                                  compile) from <specfile>
DEBUG:       -bd                        check build dependencies <specfile>
DEBUG:       -bi                        build through %install (%prep, %conf, %build,
DEBUG:                                  then install) from <specfile>
DEBUG:       -bl                        verify %files section from <specfile>
DEBUG:       -ba                        build source and binary packages from
DEBUG:                                  <specfile>
DEBUG:       -bb                        build binary package only from <specfile>
DEBUG:       -bs                        build source package only from <specfile>
DEBUG:       -br                        build source package only from <specfile> -
DEBUG:                                  calculate dynamic build requires
DEBUG:       -rp                        build through %prep (unpack sources and apply
DEBUG:                                  patches) from <source package>
DEBUG:       -rf                        build through %conf (%prep, then %conf) from
DEBUG:                                  <specfile>
DEBUG:       -rc                        build through %build (%prep, %conf, then
DEBUG:                                  compile) from <source package>
DEBUG:       -rd                        check build dependencies <source package>
DEBUG:       -ri                        build through %install (%prep, %conf, %build,
DEBUG:                                  then install) from <source package>
DEBUG:       -rl                        verify %files section from <source package>
DEBUG:       -ra                        build source and binary packages from <source
DEBUG:                                  package>
DEBUG:       -rb                        build binary package only from <source
DEBUG:                                  package>
DEBUG:       -rs                        build source package only from <source
DEBUG:                                  package>
DEBUG:       -rr                        build source package only from <source
DEBUG:                                  package> - calculate dynamic build requires
DEBUG:       -tp                        build through %prep (unpack sources and apply
DEBUG:                                  patches) from <tarball>
DEBUG:       -tf                        build through %conf (%prep, then %conf) from
DEBUG:                                  <specfile>
DEBUG:       -tc                        build through %build (%prep, %conf, then
DEBUG:                                  compile) from <tarball>
DEBUG:       -td                        check build dependencies <tarball>
DEBUG:       -ti                        build through %install (%prep, %conf, %build,
DEBUG:                                  then install) from <tarball>
DEBUG:       -ta                        build source and binary packages from
DEBUG:                                  <tarball>
DEBUG:       -tb                        build binary package only from <tarball>
DEBUG:       -ts                        build source package only from <tarball>
DEBUG:       -tr                        build source package only from <tarball> -
DEBUG:                                  calculate dynamic build requires
DEBUG:       --rebuild                  build binary package from <source package>
DEBUG:       --recompile                build through %install (%prep, %build, then
DEBUG:                                  install) from <source package>
DEBUG:       --buildroot=DIRECTORY      override build root
DEBUG:       --build-in-place           run build in current directory
DEBUG:       --clean                    remove build tree when done
DEBUG:       --nobuild                  do not execute any stages of the build
DEBUG:       --nodeps                   do not verify build dependencies
DEBUG:       --nodirtokens              generate package header(s) compatible with
DEBUG:                                  (legacy) rpm v3 packaging
DEBUG:       --noclean                  do not execute %clean stage of the build
DEBUG:       --noprep                   do not execute %prep stage of the build
DEBUG:       --nocheck                  do not execute %check stage of the build
DEBUG:       --rmsource                 remove sources when done
DEBUG:       --rmspec                   remove specfile when done
DEBUG:       --short-circuit            skip straight to specified stage (only for
DEBUG:                                  c,i)
DEBUG:       --target=CPU-VENDOR-OS     override target platform
DEBUG: Common options for all rpm modes and executables:
DEBUG:   -D, --define='MACRO EXPR'      define MACRO with value EXPR
DEBUG:       --undefine=MACRO           undefine MACRO
DEBUG:   -E, --eval='EXPR'              print macro expansion of EXPR
DEBUG:       --target=CPU-VENDOR-OS     Specify target platform
DEBUG:       --macros=<FILE:...>        read <FILE:...> instead of default file(s)
DEBUG:       --load=<FILE>              load a single macro file
DEBUG:       --noplugins                don't enable any plugins
DEBUG:       --nodigest                 don't verify package digest(s)
DEBUG:       --nosignature              don't verify package signature(s)
DEBUG:       --rcfile=<FILE:...>        read <FILE:...> instead of default file(s)
DEBUG:   -r, --root=ROOT                use ROOT as top level directory (default: "/")
DEBUG:       --dbpath=DIRECTORY         use database in DIRECTORY
DEBUG:       --querytags                display known query tags
DEBUG:       --showrc                   display final rpmrc and macro configuration
DEBUG:       --quiet                    provide less detailed output
DEBUG:   -v, --verbose                  provide more detailed output
DEBUG:       --version                  print the version of rpm being used
DEBUG: Options implemented via popt alias/exec:
DEBUG:       --with=<option>            enable configure <option> for build
DEBUG:       --without=<option>         disable configure <option> for build
DEBUG:       --scm=<scm>                shortcut for '--define="__scm <scm>"'
DEBUG:       --buildpolicy=<policy>     set buildroot <policy> (e.g. compress man
DEBUG:                                  pages)
DEBUG:       --trace                    trace macro expansion
DEBUG:       --nodebuginfo              do not generate debuginfo for this package
DEBUG: Help options:
DEBUG:   -?, --help                     Show this help message
DEBUG:       --usage                    Display brief usage message
DEBUG: Child return code was: 0
INFO: ENTER ['do_with_status'](['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --noclean --target x86_64 --nodeps /builddir/build/SPECS/cockpit.spec'], chrootPath='/var/lib/mock/fedora-37-x86_64/root'env={'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'}shell=Falselogger=<mockbuild.trace_decorator.getLog object at 0x7f5d4e267610>timeout=0uid=1002gid=135user='mockbuild'nspawn_args=['--capability=cap_ipc_lock', '--bind=/tmp/mock-resolv.8kj_ym5i:/etc/resolv.conf', '--bind=/dev/btrfs-control', '--bind=/dev/mapper/control', '--bind=/dev/loop-control', '--bind=/dev/loop0', '--bind=/dev/loop1', '--bind=/dev/loop2', '--bind=/dev/loop3', '--bind=/dev/loop4', '--bind=/dev/loop5', '--bind=/dev/loop6', '--bind=/dev/loop7', '--bind=/dev/loop8', '--bind=/dev/loop9', '--bind=/dev/loop10', '--bind=/dev/loop11']unshare_net=TrueprintOutput=False)
```

Uhhh

jelly:
> From the log:
> 
> ```
> DEBUG: Executing command: ['/usr/bin/systemd-nspawn', '-q', '-M', '76e1daa866f7479686fd568d170da89a', '-D', '/var/lib/mock/fedora-37-x86_64/root', '-a', '--console=pipe', '--setenv=TERM=vt100', '--setenv=SHELL=/bin/bash', '--setenv=HOME=/builddir', '--setenv=HOSTNAME=mock', '--setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin', '--setenv=PROMPT_COMMAND=printf "\\033]0;<mock-chroot>\\007"', '--setenv=PS1=<mock-chroot> \\s-\\v\\$ ', '--setenv=LANG=C.UTF-8', '--resolv-conf=off', 'rpmbuild', '--help'] with env {'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8', 'SYSTEMD_NSPAWN_TMPFS_TMP': '0', 'SYSTEMD_SECCOMP': '0'} and shell False
> DEBUG: Usage: rpmbuild [OPTION...]
> DEBUG: Build options with [ <specfile> | <tarball> | <source package> ]:
> DEBUG:       -bp                        build through %prep (unpack sources and apply
> DEBUG:                                  patches) from <specfile>
> DEBUG:       -bf                        build through %conf (%prep, then %conf) from
> DEBUG:                                  <specfile>
> DEBUG:       -bc                        build through %build (%prep, %conf, then
> DEBUG:                                  compile) from <specfile>
> DEBUG:       -bd                        check build dependencies <specfile>
> DEBUG:       -bi                        build through %install (%prep, %conf, %build,
> DEBUG:                                  then install) from <specfile>
> DEBUG:       -bl                        verify %files section from <specfile>
> DEBUG:       -ba                        build source and binary packages from
> DEBUG:                                  <specfile>
> DEBUG:       -bb                        build binary package only from <specfile>
> DEBUG:       -bs                        build source package only from <specfile>
> DEBUG:       -br                        build source package only from <specfile> -
> DEBUG:                                  calculate dynamic build requires
> DEBUG:       -rp                        build through %prep (unpack sources and apply
> DEBUG:                                  patches) from <source package>
> DEBUG:       -rf                        build through %conf (%prep, then %conf) from
> DEBUG:                                  <specfile>
> DEBUG:       -rc                        build through %build (%prep, %conf, then
> DEBUG:                                  compile) from <source package>
> DEBUG:       -rd                        check build dependencies <source package>
> DEBUG:       -ri                        build through %install (%prep, %conf, %build,
> DEBUG:                                  then install) from <source package>
> DEBUG:       -rl                        verify %files section from <source package>
> DEBUG:       -ra                        build source and binary packages from <source
> DEBUG:                                  package>
> DEBUG:       -rb                        build binary package only from <source
> DEBUG:                                  package>
> DEBUG:       -rs                        build source package only from <source
> DEBUG:                                  package>
> DEBUG:       -rr                        build source package only from <source
> DEBUG:                                  package> - calculate dynamic build requires
> DEBUG:       -tp                        build through %prep (unpack sources and apply
> DEBUG:                                  patches) from <tarball>
> DEBUG:       -tf                        build through %conf (%prep, then %conf) from
> DEBUG:                                  <specfile>
> DEBUG:       -tc                        build through %build (%prep, %conf, then
> DEBUG:                                  compile) from <tarball>
> DEBUG:       -td                        check build dependencies <tarball>
> DEBUG:       -ti                        build through %install (%prep, %conf, %build,
> DEBUG:                                  then install) from <tarball>
> DEBUG:       -ta                        build source and binary packages from
> DEBUG:                                  <tarball>
> DEBUG:       -tb                        build binary package only from <tarball>
> DEBUG:       -ts                        build source package only from <tarball>
> DEBUG:       -tr                        build source package only from <tarball> -
> DEBUG:                                  calculate dynamic build requires
> DEBUG:       --rebuild                  build binary package from <source package>
> DEBUG:       --recompile                build through %install (%prep, %build, then
> DEBUG:                                  install) from <source package>
> DEBUG:       --buildroot=DIRECTORY      override build root
> DEBUG:       --build-in-place           run build in current directory
> DEBUG:       --clean                    remove build tree when done
> DEBUG:       --nobuild                  do not execute any stages of the build
> DEBUG:       --nodeps                   do not verify build dependencies
> DEBUG:       --nodirtokens              generate package header(s) compatible with
> DEBUG:                                  (legacy) rpm v3 packaging
> DEBUG:       --noclean                  do not execute %clean stage of the build
> DEBUG:       --noprep                   do not execute %prep stage of the build
> DEBUG:       --nocheck                  do not execute %check stage of the build
> DEBUG:       --rmsource                 remove sources when done
> DEBUG:       --rmspec                   remove specfile when done
> DEBUG:       --short-circuit            skip straight to specified stage (only for
> DEBUG:                                  c,i)
> DEBUG:       --target=CPU-VENDOR-OS     override target platform
> DEBUG: Common options for all rpm modes and executables:
> DEBUG:   -D, --define='MACRO EXPR'      define MACRO with value EXPR
> DEBUG:       --undefine=MACRO           undefine MACRO
> DEBUG:   -E, --eval='EXPR'              print macro expansion of EXPR
> DEBUG:       --target=CPU-VENDOR-OS     Specify target platform
> DEBUG:       --macros=<FILE:...>        read <FILE:...> instead of default file(s)
> DEBUG:       --load=<FILE>              load a single macro file
> DEBUG:       --noplugins                don't enable any plugins
> DEBUG:       --nodigest                 don't verify package digest(s)
> DEBUG:       --nosignature              don't verify package signature(s)
> DEBUG:       --rcfile=<FILE:...>        read <FILE:...> instead of default file(s)
> DEBUG:   -r, --root=ROOT                use ROOT as top level directory (default: "/")
> DEBUG:       --dbpath=DIRECTORY         use database in DIRECTORY
> DEBUG:       --querytags                display known query tags
> DEBUG:       --showrc                   display final rpmrc and macro configuration
> DEBUG:       --quiet                    provide less detailed output
> DEBUG:   -v, --verbose                  provide more detailed output
> DEBUG:       --version                  print the version of rpm being used
> DEBUG: Options implemented via popt alias/exec:
> DEBUG:       --with=<option>            enable configure <option> for build
> DEBUG:       --without=<option>         disable configure <option> for build
> DEBUG:       --scm=<scm>                shortcut for '--define="__scm <scm>"'
> DEBUG:       --buildpolicy=<policy>     set buildroot <policy> (e.g. compress man
> DEBUG:                                  pages)
> DEBUG:       --trace                    trace macro expansion
> DEBUG:       --nodebuginfo              do not generate debuginfo for this package
> DEBUG: Help options:
> DEBUG:   -?, --help                     Show this help message
> DEBUG:       --usage                    Display brief usage message
> DEBUG: Child return code was: 0
> INFO: ENTER ['do_with_status'](['bash', '--login', '-c', '/usr/bin/rpmbuild -bs --noclean --target x86_64 --nodeps /builddir/build/SPECS/cockpit.spec'], chrootPath='/var/lib/mock/fedora-37-x86_64/root'env={'TERM': 'vt100', 'SHELL': '/bin/bash', 'HOME': '/builddir', 'HOSTNAME': 'mock', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PROMPT_COMMAND': 'printf "\\033]0;<mock-chroot>\\007"', 'PS1': '<mock-chroot> \\s-\\v\\$ ', 'LANG': 'C.UTF-8'}shell=Falselogger=<mockbuild.trace_decorator.getLog object at 0x7f5d4e267610>timeout=0uid=1002gid=135user='mockbuild'nspawn_args=['--capability=cap_ipc_lock', '--bind=/tmp/mock-resolv.8kj_ym5i:/etc/resolv.conf', '--bind=/dev/btrfs-control', '--bind=/dev/mapper/control', '--bind=/dev/loop-control', '--bind=/dev/loop0', '--bind=/dev/loop1', '--bind=/dev/loop2', '--bind=/dev/loop3', '--bind=/dev/loop4', '--bind=/dev/loop5', '--bind=/dev/loop6', '--bind=/dev/loop7', '--bind=/dev/loop8', '--bind=/dev/loop9', '--bind=/dev/loop10', '--bind=/dev/loop11']unshare_net=TrueprintOutput=False)
> ```
> 
> Uhhh

This happens in other PR's which succeed so seems fine.

martinpitt:
Yay! I passed, and the [build log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18266-20230221-144133-50baa48f-fedora-37-pybridge/log.html) now shows that it `REMOVE`s all webpacks. Nice!

pkg/shell/index.html
jelly: Yup :)

martinpitt:
Also triggering rhel-* tests, as only these run checks for all languages. Dropping no-test label, in case you have to push again.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -105,7 +103,17 @@ function init(options) {
...
-    return Promise.all(get_po_files().map(f => buildFile(f, webpack_module, webpack_compilation)));
+    return Promise.all(get_po_files().map(f => {
+        if (webpack_compilation)
+            webpack_compilation.fileDependencies.add(f);
+        const lang = path.basename(f).slice(0, -3);
+        return Promise.all([
+            buildFile(f, webpack_module, webpack_compilation,
+                      `${config.subdir}po.${lang}.js`, str => !str.startsWith(`pkg/${config.subdir}manifest.json`)),
+            buildFile(f, webpack_module, webpack_compilation,
+                      `${config.subdir}po.manifest.${lang}.js`, str => str.startsWith(`pkg/${config.subdir}manifest.json`))
```
martinpitt, line 114: I initially did not understand why this is `.startsWith()` instead of `.includes()`. When I have a translation like
```
#: src/components/vm/vmDetailsPage.jsx:112    
#: src/components/vm/vmDetailsPage.jsx:265 src/manifest.json:0    
msgid "Virtual machines"    
msgstr "Virtuelle Maschinen"  
```
then I want *both* LL.po.js and LL.manifest.po.js to contain this string. However, buildFile() splits references into words.

But this is still wrong -- they are in pkg/subdir only for cockpit, but not for starter-kit and friends. Let's rather not make assumptions about the directory. How about `.endsWith('.manifest.json`)` instead? That entirely avoids making assumptions about the location.

martinpitt (changes requested):
Thanks! Needs one more fix,but otherwise LGTM

pkg/lib/cockpit-po-plugin.js
```diff
@@ -105,7 +103,17 @@ function init(options) {
...
 
 function run(webpack_module, webpack_compilation) {
-    return Promise.all(get_po_files().map(f => buildFile(f, webpack_module, webpack_compilation)));
+    return Promise.all(get_po_files().map(f => {
+        if (webpack_compilation)
+            webpack_compilation.fileDependencies.add(f);
+        const lang = path.basename(f).slice(0, -3);
+        return Promise.all([
+            buildFile(f, webpack_module, webpack_compilation,
+                      `${config.subdir}po.${lang}.js`, str => !str.includes('manifest.json') && !str.includes('manifest.json.in')),
```
martinpitt, line 112: The .in variants are redundant. I only brought that up as a possible scenario why .includes() is more correct than .endsWith(). Please drop them again.

martinpitt (changes requested):

pkg/lib/cockpit-po-plugin.js
jelly: *facepalm* yes ofcourse that makes sense.

martinpitt (approved):
Thanks! I'm happy now. However, also still needs re-review from @allisonkarlitskaya as she still has a pending -1.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -105,7 +103,17 @@ function init(options) {
 }
 
 function run(webpack_module, webpack_compilation) {
-    return Promise.all(get_po_files().map(f => buildFile(f, webpack_module, webpack_compilation)));
+    return Promise.all(get_po_files().map(f => {
+        if (webpack_compilation)
+            webpack_compilation.fileDependencies.add(f);
+        const lang = path.basename(f).slice(0, -3);
+        return Promise.all([
+            buildFile(f, webpack_module, webpack_compilation,
```
allisonkarlitskaya, line 111: Maybe a comment here about how these two conditions, although they appear opposite, are not, and how a string might end up in both files (and why we want that).

allisonkarlitskaya (changes requested):
This still seems to be missing the fallback to using `po.js` for the manifest translations if separate files are not present.

jelly:
> This still seems to be missing the fallback to using `po.js` for the manifest translations if separate files are not present.

This is now fixed, but I am not sure if you like it. I just used a naive approach of => no data, fallback to trying to po.js files. We can make this more specific with like by checking if something else then 'en' was requested but I am not sure if that's worth the effort.

Note that loading translations is borked in cockpit-macines as it uses globbing!
https://github.com/cockpit-project/cockpit-machines/pull/964

jelly:
Oh, we should maybe add a test for this fallback behaviour in `check-pages`. I'd need to figure out how we install a fake application...

src/cockpit/packages.py
```diff
@@ -234,6 +242,22 @@ def load_file(self, filename: str) -> Document:
 
         return Document(path.open('rb'), content_type, content_encoding, content_security_policy)
 
+    def load_manifest_translation(self, locales: List[str]) -> Optional[str]:
```
jelly, line 245: This is a useless duplicate, this needs a refactor.

pkg/lib/cockpit-po-plugin.js
jelly: But hmm we do filter out manifest.json for .po.js?

jelly:
image

Interesting, seems we combine gzipped JS and non-gzipped.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -111,8 +109,19 @@ function init(options) {
...
-        promises.push(...get_po_files().map(po_file =>
-            buildFile(po_file, subdir, webpack_module, webpack_compilation))));
+        promises.push(...get_po_files().map(f => {
+            if (webpack_compilation)
+                webpack_compilation.fileDependencies.add(f);
+            const lang = path.basename(f).slice(0, -3);
+            return Promise.all([
+                // Separate translations for the manifest.json file and normal pages
+                buildFile(f, subdir, webpack_module, webpack_compilation,
+                          `${subdir}/po.${lang}.js`, str => !str.includes('manifest.json')),
```
allisonkarlitskaya, line 119: `.endsWith('/manifest.json')` rather, no?

src/cockpit/packages.py
```diff
@@ -494,13 +505,22 @@ def load_manifests_js(self, headers: JsonObject) -> Document:
         for name, package in self.packages.items():
             if name in ['static', 'base1']:
                 continue
+
+            package.ensure_scanned()
+            # fallback for old packages without a po.manifest.$locale.js
+            if package.manifest_translations is None or not package.manifest_translations:
```
allisonkarlitskaya, line 511: This will never be `None` since you initialize it.

src/cockpit/packages.py
```diff
@@ -261,7 +272,7 @@ def load_path(self, path: str, headers: JsonObject) -> Document:
 
         if path == 'po.js':
             locales = parse_accept_language(headers)
-            return self.load_translation(locales)
+            return self.load_translation(self.translations, locales)
```
allisonkarlitskaya, line 275: This feels a little bit too implementation-detaily in that Packages reaches a bit too far inside of Package.  What if we handed the `path` to the function and had it decide internally?

allisonkarlitskaya (changes requested):

pkg/shell/index.html
```diff
@@ -8,7 +8,10 @@
     <link href="../../static/branding.css" rel="stylesheet" />
     <script src="../base1/cockpit.js"></script>
     <script src="../manifests.js"></script>
+    <!-- HACK: C bridge loads translations via glob and Python via manifest.js -->
+    <script src="../*/po.manifest.js"></script>
```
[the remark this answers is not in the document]
jelly: Alternatively we include the manifest translations in the `po.js` as well and drop this as they should then be loaded via `../*/po.js`?

src/cockpit/packages.py
jelly: I'd be up for that but up for that

src/cockpit/packages.py
```diff
@@ -220,6 +228,9 @@ def get_content_security_policy(self) -> str:
 
         return ' '.join(f'{k} {v};' for k, v in policy.items()) + ' block-all-mixed-content'
 
+    def __str__(self):
+        return f'Package(name={self.name})'
```
jelly, line 232: Hehe unrelated debug, might be nice to have however.

pkg/lib/cockpit-po-plugin.js
```diff
@@ -111,8 +109,19 @@ function init(options) {
 function run(webpack_module, webpack_compilation) {
     const promises = [];
     config.subdirs.map(subdir =>
```
allisonkarlitskaya, line 111: I'm getting a headache trying to read this.  Can we turn this into nested `for` loops or something instead?

src/cockpit/packages.py
```diff
@@ -234,21 +242,29 @@ def load_file(self, filename: str) -> Document:
 
         return Document(path.open('rb'), content_type, content_encoding, content_security_policy)
 
-    def load_translation(self, locales: List[str]) -> Document:
+    def load_translation(self, path: str, locales: List[str]) -> Document:
         self.ensure_scanned()
-        assert self.translations is not None
+        if path == 'po.js':
+            translations = self.translations
+        elif self.manifest_translations:
```
allisonkarlitskaya, line 249: That looks wrong.  Should rather have a `not` there, no?  But I'd actually change the bodies instead.

I still like the idea of adding a level of indirection on the translations by filename, so you could lookup

`self.translations['po.js']['de']` or `self.translations['po.manifest.js']['de']`

Then you could implement the fallback as a one-time check from `ensure_scanned()` — if `self.translations['po.manifest.js']` is empty then you set it equal to `self.translations['po.js']` and don't have to think about it anymore.

allisonkarlitskaya (changes requested):

allisonkarlitskaya:
Here's a rough outline of what I was trying to talk about...

https://github.com/cockpit-project/cockpit/pull/18266/commits/16cf39b438d9b030a10ee73c63161f4c40198880

maybe it even works :rofl:

pkg/lib/cockpit-po-plugin.js
jelly: Looking at the po files we do it correct:
```
po/cs.po:#: pkg/storaged/manifest.json:0
```

jelly (commented):
In general this PR LGTM @allisonkarlitskaya

pkg/lib/cockpit-po-plugin.js
allisonkarlitskaya: then maybe add `/` and `:` to the string we match on?

I guess this is not super important...

pkg/lib/cockpit-po-plugin.js
```diff
@@ -111,8 +109,19 @@ function init(options) {
...
-        promises.push(...get_po_files().map(po_file =>
-            buildFile(po_file, subdir, webpack_module, webpack_compilation))));
+        promises.push(...get_po_files().map(f => {
```
allisonkarlitskaya, line 112: Using a lambda in a map which does side effects via its closure is really double plus not good.  I'd love to see this turned one more layer into a for loop....

allisonkarlitskaya (changes requested):

pkg/lib/cockpit-po-plugin.js
```diff
@@ -109,15 +109,15 @@ function init(options) {
 function run(webpack_module, webpack_compilation) {
     const promises = [];
     for (const subdir of config.subdirs) {
-        for (const f of get_po_files()) {
+        for (const po_file of get_po_files()) {
```
allisonkarlitskaya, line 112: <3

jelly:
@allisonkarlitskaya re-triggered the CI

allisonkarlitskaya (approved):
If you can get the bots to green, I'm happy now.

Thanks for doing this!

jelly:
It went :green_circle:

PR #18268, opened 2023-01-31
lib: cockpit-components-file-autocomplete: fix moving back to previous path

KKoukiou:
Only update the file list if the directory exists in path. This approach solves the current issue where the menu list is not updating correctly, if pressing backspace to get some parent existing directory after a wrong path was given.

Also fix the error message; cockpit.message does not return a string if 'message' is not an attribute of the object parsed as this is now appearing wrongly.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1999945

### DIFF
--- a/pkg/lib/cockpit-components-file-autocomplete.jsx
+++ b/pkg/lib/cockpit-components-file-autocomplete.jsx
@@ -104,7 +104,7 @@ export class FileAutoComplete extends React.Component {
         });
 
         channel.addEventListener("close", (ev, data) => {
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks!

KKoukiou:
The new test lines were flaky on rhel-8-8, https://cockpit-logs.us-east-1.linodeobjects.com/pull-18268-20230131-123604-28d95670-rhel-8-8/log.html#337

martinpitt (approved):
Thanks!

PR #18275, opened 2023-01-31
test: add tests for the internal metrics channel

jelly:
This get's our coverage up to 90% for channels/metrics.py


martinpitt (approved):
So this is a translation of test-metrics.c test_cpu_cores . TBH this should be a QUnit test to cover both bridges, but fine of course. Thanks!

PR #18279, opened 2023-02-01
selinux: Allow cockpit-session to run gnome-keyring

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=2165996


martinpitt:
@zpytela does this make sense? Dekuji!

zpytela:
From selinux-policy PoV this seems correct, but it depends what you want.

From a long-term perspective it is good to keep the interfaces calls alphabetically ordered.

marusak (approved):
Danke

PR #18280, opened 2023-02-01
python: initial implementation of ssh support

allisonkarlitskaya:
Add support for connecting to additional hosts via ssh, using ferny.
    
In general, this is not working well on RHEL8 — ferny depends on the KnownHostsCommand feature of OpenSSH, which is not available on RHEL8.  Without this, we don't get access to the public key of the remote host that we're connecting to, in case the host isn't known yet, which means we can't add it to the file.

- [x] #18398 

https://issues.redhat.com/browse/COCKPIT-909

### DIFF
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -17,12 +17,15 @@
 
 import argparse
 import asyncio
+import ctypes
 import json
 import logging
 import pwd
 import os
 import shlex
+import signal
 import socket
+import subprocess
 
 from typing import Dict, Iterable, List, Tuple, Type
 
@@ -181,6 +184,34 @@ def setup_logging(debug: bool):
             logging.getLogger(module).setLevel(logging.DEBUG)
 
 
+def start_ssh_agent() -> None:
+    prctl = ctypes.CDLL(None).prctl
+    PR_SET_PDEATHSIG = 1
+
+    # NB: We prefer SIGTERM to SIGKILL here since the agent needs to have a
+    # chance to clean up its listener socket.
+    try:
+        proc = subprocess.Popen(['ssh-agent',
+                                 '-D',  # no-daemon
+                                 '-s',  # shell-style output
+                                 ], stdout=subprocess.PIPE, universal_newlines=True,
+                                preexec_fn=lambda: prctl(PR_SET_PDEATHSIG, signal.SIGTERM))
+        assert proc.stdout is not None
+
+        # Wait for the agent to write at least one line and look for the
+        # listener socket.  If we fail to find it, kill the agent — something
+        # went wrong.
+        for token in shlex.shlex(proc.stdout.readline(), punctuation_chars=True):
+            if token.startswith('SSH_AUTH_SOCK='):
+                os.environ['SSH_AUTH_SOCK'] = token.replace('SSH_AUTH_SOCK=', '', 1)
+                break
+        else:
+            proc.terminate()
+
+    except OSError as exc:
+        logger.warning("Could not start ssh-agent: %s", exc)
+
+
 def main() -> None:
     polyfills.install()
 
@@ -210,6 +241,10 @@ def main() -> None:
     elif args.bridges:
         print(json.dumps(Packages().get_bridge_configs(), indent=2))
         return
+
+    # The privileged bridge doesn't need ssh-agent, but the main one does
+    if 'SSH_AUTH_SOCK' not in os.environ and not args.privileged:
+        start_ssh_agent()
 
     # asyncio.run() shim for Python 3.6 support
     run_async(run(args), debug=args.debug)
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -23,7 +23,6 @@ import pwd
 import os
 import shlex
 import socket
-import sys
 
 from typing import Dict, Iterable, List, Tuple, Type
 
@@ -185,13 +184,6 @@ def setup_logging(debug: bool):
 def main() -> None:
     polyfills.install()
 
-    # The --privileged bridge gets spawned with its stderr being consumed by a
-    # pipe used for reading authentication-related message from sudo.  The
-    # absolute first thing we want to do is to recover the original stderr that
-    # we had.
-    if '--privileged' in sys.argv:
-        try_to_receive_stderr()
-
     parser = argparse.ArgumentParser(description='cockpit-bridge is run automatically inside of a Cockpit session.')
     parser.add_argument('--privileged', action='store_true', help='Privileged copy of the bridge')
     parser.add_argument('--packages', action='store_true', help='Show Cockpit package information')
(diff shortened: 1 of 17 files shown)


martinpitt:
Depending PR landed. Removing "blocked", needs rebasing now. Can't wait to see this in action!

martinpitt:
@allisonkarlitskaya It's quite amazing to see how "small" the "remote!" commit is now. :grin: (of course, "standing on the shoulder of giants", etc., etc.) Fixing all 24 tests in one go is maybe too ambitious -- As soon as this fixes a few, let's land and make sure they don't regress again, then we can spread out investigating the others to more people (I'm happy to help).

src/cockpit/remote.py
```diff
@@ -15,21 +15,185 @@
...
+
+    async def do_askpass(self, messages: str, prompt: str, hint: str) -> Optional[str]:
+        logger.debug('Got askpass(%s): %s', hint, prompt)
+
+        match = PasswordResponder.PASSPHRASE_RE.fullmatch(prompt)
+        if match is not None:
+            path = match.group(1)
+            logger.debug("This is a passphrase request for %s, but we don't do those.  Abort.", path)
+            self.error_message = f'locked identity: {path}'
+            return None
```
martinpitt, line 69: Fair enough -- Cockpit's UI does support it, but of course it's fine to implement it in a follow-up.

src/cockpit/remote.py
```diff
@@ -15,21 +15,185 @@
...
+
+        assert self.password is not None
+        assert self.attempts == 0
+        self.attempts += 1
+        return self.password
```
martinpitt, line 74: So this is the case where the askpass prompt is not for an encrypted key, but the remote login PAM password, right?

src/cockpit/remote.py
```diff
@@ -15,21 +15,185 @@
...
+        assert self.session is not None
+        logger.debug('Starting ssh session user=%s, host=%s, private=%s', self.user, self.host, self.private)
+
+        basename, colon, portstr = self.host.rpartition(':')
+        if colon and portstr.isdigit():
+            host = basename
+            port = int(portstr)
+        else:
+            host = self.host
+            port = None
```
martinpitt, line 94: This silently ignores cases like `self.host == 'foo:bar'`. Then `colon == ':` but the isdigit() test fails. This case should throw an error, instead of silently ignoring the `:bar` part.

test/pytest/test_bridge.py
```diff
@@ -78,6 +78,7 @@ async def test_echo(self):
         await self.transport.assert_msg('', command='done', channel=echo)
         await self.transport.assert_msg('', command='close', channel=echo)
 
+    @unittest.skip
```
martinpitt, line 81: `@todo` perhaps? Skipping without a reason or comment feels like we're going to forget about it.

martinpitt (changes requested):
Yay, I'm really happy to see some progress here, thanks! There's some things to discuss, though.

src/cockpit/remote.py
```diff
@@ -15,21 +15,185 @@
...
-from .router import RoutingError, RoutingRule
+from ._vendor import ferny
+from .router import RoutingRule, Router
+from .peer import Peer, PeerError
+
+logger = logging.getLogger(__name__)
+
+
+class PasswordResponder(ferny.InteractionResponder):
+    PASSPHRASE_RE = re.compile(r"Enter passphrase for key '(.*)': ")
```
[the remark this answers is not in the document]
allisonkarlitskaya: Ya.  I'm painfully aware of this.  I tried hard to avoid it, but I don't think that we have a lot of other options here, unless we want to start patching upstream ssh....

This is the sort of stuff that got us in trouble ~5 years ago...

src/cockpit/remote.py
```diff
@@ -15,21 +15,185 @@
...
+
+    def __init__(self, password: Optional[str]):
+        self.password = password
+        self.error_message = None
+        self.attempts = 0
+
+    async def do_hostkey(self, reason: str, host: str, algorithm: str, key: str, fingerprint: str) -> bool:
+        try:
+            logger.debug('checking ~/.ssh/known_hosts')
+            with open(os.path.expanduser('~/.ssh/known_hosts')) as file:
```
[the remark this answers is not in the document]
allisonkarlitskaya: This gets called, always, on every connect, sometimes multiple times, regardless of if the key is already known or not.

The manpage mentions

> Specifies a command to use to obtain a list of host keys, in addition to those listed in UserKnownHostsFile and GlobalKnownHostsFile.  This command is executed after the files have been read.

but ssh does absolutely nothing to inform us about if it found a match in those files or not, and calls the command unconditionally.

I'm glad you flagged this because I also feel a bit uncomfortable about it.  My "full strength" idea here was to check all of the files (which we can query from ssh with `-G`) and also any KnownHostsCommand that the user might have defined for themselves, but I'm not really thrilled about that either.

I think we might have to discuss the possibility of breaking compatibility with the behaviour the front-end expects from us and switch over to using an interaction that's based more on the "Do you accept this key?" prompt via askpass.  It would also let us avoid being the ones responsible for writing to the file.  More parsing, though... :/

In short, I'm not really sure what to tell you at this point other than "ya, this is gonna need to be better".

src/cockpit/remote.py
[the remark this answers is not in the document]
martinpitt: I think running ssh(1) through `LC_MESSAGES=C.UTF-8` would go a long way actually (but that's a ferny change, I think?)

src/cockpit/remote.py
allisonkarlitskaya: This is actually what's supposed to happen.  We don't have support for unlocking locked keys, and we don't have any way to receive the passphrases to do so.

src/cockpit/remote.py
allisonkarlitskaya: Yes.  And we want to make note about if we provided it or not to distinguish the report we send for the auth methods as part of the error message.

src/cockpit/remote.py
allisonkarlitskaya: I don't agree that the port will be dropped in this case.  It will hit the `else:` branch and set `host = self.host` (ie: the full original name) and `port = None`.

martinpitt: Oh, ok - that feels rather strange and unpredictable, but hmm..

src/cockpit/remote.py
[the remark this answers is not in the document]
allisonkarlitskaya: It was a dumb issue: I accidentally changed the `problem='unknown-hostkey'` to `problem='unknown-host'`, which apparently means something completely different to the frontend :)

src/cockpit/bridge.py
```diff
@@ -211,6 +242,10 @@ def main() -> None:
         print(json.dumps(Packages().get_bridge_configs(), indent=2))
         return
 
+    # The privileged bridge doesn't need ssh-agent, but the main one does
+    if 'SSH_AUTH_SOCK' not in os.environ and not args.privileged:
+        start_ssh_agent()
```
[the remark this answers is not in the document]
allisonkarlitskaya: I think it's definitely impossible: I'm not sure how the environment would leak out of the one bridge and into the other...

src/cockpit/remote.py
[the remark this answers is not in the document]
martinpitt: [current run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18280-20230418-072056-1e5ebd74-fedora-37-pybridge/log.html#307) didn't finish yet, but this is clearly superior. Thanks so much!

martinpitt:
[this run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18280-20230418-102746-de488724-centos-8-stream-pybridge/log.html#235-2) has some flakes:
 - TestServices.testBasic is issue #18465 
 - VDO failure is a too short timeout, I'll send a follow-up PR for that.

test/pytest/test_bridge.py
martinpitt: I filed https://issues.redhat.com/browse/COCKPIT-983 to track this.

martinpitt (approved):
Good enough! Let's land this, and investigate the remaining failures. Thanks!

PR #18283, opened 2023-02-02
python: Properly report HTTPExceptions

martinpitt:
If reading the response throws a http.client exception, close the channel with that error message, instead of crashing the thread and letting it log tracebacks. This does not change observable behaviour much, but cleans up unexpected error messages.

---

This will fix the last [podman failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1196-20230202-124820-1dc12492-fedora-37-pybridge/log.html) with the pybridge, see https://github.com/cockpit-project/cockpit-podman/pull/1196

### DIFF
--- a/src/cockpit/channels/http.py
+++ b/src/cockpit/channels/http.py
@@ -108,12 +108,19 @@ class HttpChannel(Channel):
                            self.options.get('path'),
                            headers=self.options.get('headers') or {},
                            body=self.body)
-        response = connection.getresponse()
-        self.loop.call_soon_threadsafe(lambda: self.send_control(
-            command='response', status=response.status, reason=response.reason,
-            headers=self.parse_headers(response.headers)))
-        self.read_send_response(response)
-        connection.close()
+        try:
+            response = connection.getresponse()
+            self.loop.call_soon_threadsafe(lambda: self.send_control(
+                command='response', status=response.status, reason=response.reason,
+                headers=self.parse_headers(response.headers)))
+            self.read_send_response(response)
+        except http.client.HTTPException as error:
+            msg = str(error)
+            logger.debug('HTTP reading response failed: %s', msg)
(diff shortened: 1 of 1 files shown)


src/cockpit/channels/http.py
```diff
@@ -108,11 +108,19 @@ def request(self):
...
-            headers=self.parse_headers(response.headers)))
-        self.read_send_response(response)
+        try:
+            response = connection.getresponse()
+            self.loop.call_soon_threadsafe(lambda: self.send_control(
+                command='response', status=response.status, reason=response.reason,
+                headers=self.parse_headers(response.headers)))
+            self.read_send_response(response)
+        except http.client.HTTPException as error:
+            logger.debug('HTTP %s reading response failed: %s', error)
```
jelly, line 118: This passes one argument while the format string expects two? Also shouldn't we call `str(error`)

jelly (dismissed):

src/cockpit/channels/http.py
```diff
@@ -108,11 +108,19 @@ def request(self):
...
+            response = connection.getresponse()
+            self.loop.call_soon_threadsafe(lambda: self.send_control(
+                command='response', status=response.status, reason=response.reason,
+                headers=self.parse_headers(response.headers)))
+            self.read_send_response(response)
+        except http.client.HTTPException as error:
+            msg = str(error)
+            logger.debug('HTTP reading response failed: %s', msg)
+            self.loop.call_soon_threadsafe(lambda: self.close(problem='terminated', message=msg))
+            connection.close()
```
allisonkarlitskaya, line 121: This goes in `finally:`

allisonkarlitskaya (changes requested):

src/cockpit/channels/http.py
martinpitt: Thanks! Done.

allisonkarlitskaya (approved):
:+1: thx

PR #18285, opened 2023-02-02
python: Report detailed error message when opening fsread1 channel fails

martinpitt:
Teach check_open() about validating the `message` part of `close`, when given.

----

Leftover from https://github.com/cockpit-project/cockpit/pull/18257/commits/f128b1b3991481b4e0e1aede8c431a81fea6ddee#r1090499063

### DIFF
--- a/test/pytest/test_bridge.py
+++ b/test/pytest/test_bridge.py
@@ -47,13 +47,13 @@ class MockTransport(asyncio.Transport):
         self.send_json('', command='open', channel=channel, payload=payload, **kwargs)
         return channel
 
-    async def check_open(self, payload, channel=None, problem=None, **kwargs):
+    async def check_open(self, payload, channel=None, problem=None, reply_keys: Optional[Dict[str, object]] = None, **kwargs):
         ch = self.send_open(payload, channel, **kwargs)
         if problem is None:
-            await self.assert_msg('', command='ready', channel=ch)
+            await self.assert_msg('', command='ready', channel=ch, **(reply_keys or {}))
             assert ch in self.protocol.open_channels
         else:
-            await self.assert_msg('', command='close', channel=ch, problem=problem)
+            await self.assert_msg('', command='close', channel=ch, problem=problem, **(reply_keys or {}))
             assert ch not in self.protocol.open_channels
         return ch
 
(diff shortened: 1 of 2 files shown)


martinpitt:
[Another exception](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18285-20230202-202039-3d34f6f9-fedora-37-pybridge/log.html#101) - unrelated, but needs to go onto the "look into" stack.

test/pytest/test_bridge.py
```diff
@@ -47,13 +47,16 @@ def send_open(self, payload, channel=None, **kwargs):
         self.send_json('', command='open', channel=channel, payload=payload, **kwargs)
         return channel
 
-    async def check_open(self, payload, channel=None, problem=None, **kwargs):
+    async def check_open(self, payload, channel=None, problem=None, message=None, **kwargs):
```
allisonkarlitskaya, line 50: This is too ad-hoc.  If you want to do this then it might make sense to either:
  1) change `problem` to `Optional[dict[str, object]]` or
  2) add `reply_keys: Optional[Dict[str, object]] = None` or so which is passed as the `kwargs` of `self.assert_msg()` for either `ready` or `close`

haven written the two out I think i prefer the second one.

allisonkarlitskaya (changes requested):

test/pytest/test_bridge.py
```diff
@@ -47,13 +47,13 @@ def send_open(self, payload, channel=None, **kwargs):
         self.send_json('', command='open', channel=channel, payload=payload, **kwargs)
         return channel
 
-    async def check_open(self, payload, channel=None, problem=None, **kwargs):
+    async def check_open(self, payload, channel=None, close_args: Optional[Dict[str, object]] = None, **kwargs):
```
allisonkarlitskaya, line 50: I feel like this approach makes it harder to use the function in the simple failure case while also not covering the case of checking for arguments in the `ready` message.  I checked, and not a lot of channel send arguments as part of ready, but at least dbus does.

martinpitt: Reworked. Third time's the charm?

allisonkarlitskaya (approved):
Thanks!  :+1:

PR #18287, opened 2023-02-03
Document and test TCP stream channel, implement it for the pybridge

martinpitt:
This will fix c-machine's [console test failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-935-20230203-091154-9b4701a9-fedora-37-pybridge/log.html), see https://github.com/cockpit-project/cockpit-machines/pull/935 . I locally ran all the console tests with this pybridge update, and they pass now.

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -231,12 +231,13 @@ class ProtocolChannel(Channel, asyncio.Protocol):
     _loop: Optional[asyncio.AbstractEventLoop]
     _send_pongs: bool = True
     _last_ping: Optional[Dict[str, object]]
+    _create_transport_task = None
 
     # read-side EOF handling
     _close_on_eof: bool = False
     _eof: bool = False
 
-    def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> asyncio.Transport:
+    async def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> asyncio.Transport:
         """Creates the transport for this channel, according to options.
 
         The event loop for the transport is passed to the function.  The
@@ -247,9 +248,22 @@ class ProtocolChannel(Channel, asyncio.Protocol):
         raise NotImplementedError
 
     def do_open(self, options):
+        self.freeze_endpoint()
         loop = asyncio.get_running_loop()
-        transport = self.create_transport(loop, options)
+        self._create_transport_task = asyncio.create_task(self.create_transport(loop, options))
+        self._create_transport_task.add_done_callback(self.create_transport_done)
+
+    def create_transport_done(self, task):
+        assert task is self._create_transport_task
+        self._create_transport_task = None
+        try:
+            transport = task.result()
+        except ChannelError as exc:
+            self.close(**exc.kwargs)
+            return
+
         self.connection_made(transport)
+        self.thaw_endpoint()
 
     def connection_made(self, transport: asyncio.BaseTransport):
         assert isinstance(transport, asyncio.Transport)
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -20,14 +20,13 @@ import ctypes
 import logging
 import os
 import signal
-import socket
 import subprocess
 
 from typing import Dict, Optional
 
 
 from ..channel import ProtocolChannel, ChannelError
-from ..transports import SocketTransport, SubprocessTransport, SubprocessProtocol
+from ..transports import SubprocessTransport, SubprocessProtocol
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


src/cockpit/channels/stream.py
```diff
@@ -54,6 +54,31 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, o
...
+
+    def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> SocketTransport:
+        try:
+            port: int = int(options['port'])
+        except ValueError:
+            raise ChannelError('protocol-error', message='invalid "port" option for stream channel')
+        address = (options.get('address', 'localhost'), port)
+
+        try:
+            connection = socket.create_connection(address)
```
allisonkarlitskaya, line 69: This will block.

martinpitt: Argh, yes -- let's use [asyncio's loop.create_connection()](https://docs.python.org/3/library/asyncio-eventloop.html?highlight=call%20soon%20threadsafe#opening-network-connections)

martinpitt: @allisonkarlitskaya : I think I'll need your help here. `loop.create_connection()` (which seems to fit rather well here), as well as `asyncio.open_connection()` (which fits a bit less well, as we'd have to manually treat the reader and writer objects separately) are both coroutines. But all methods in `ProtocolChannel` are sync, i.e. I can't just put either of these calls into `create_transport()` or more generally into `do_open()`.

Should I use `AsyncChannel` for that? I think I'll see how to do the reading in `async def run()`, but not yet where writes enter the picture here.

allisonkarlitskaya: Take a look at https://docs.python.org/3/library/asyncio-eventloop.html#executing-code-in-thread-or-process-pools for a possible alternative.  This is what I'd do.

martinpitt: That doesn't work for the same reason: One cannot call async functions in `do_open()` or `create_transport`. The asyncio and our Channel API don't seem to fit together very well.. I suppose I'll just start with a plain old thread, unless you have a better idea.

allisonkarlitskaya: I think we need to make `create_transport()` async, then.

martinpitt: Reworked. This is actually kind of nice! I probably got some details wrong :grin:

src/cockpit/channels/stream.py
```diff
@@ -46,7 +46,7 @@ class UnixStreamChannel(ProtocolChannel):
     payload = 'stream'
     restrictions = (('unix', None),)
 
-    def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> SocketTransport:
+    async def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> SocketTransport:
         path: str = options['unix']
         connection = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
```
allisonkarlitskaya, line 51: It's probably worth replacing this with the non-blocking equivalent.

src/cockpit/channels/stream.py
```diff
@@ -86,7 +86,7 @@ def do_close(self):
             # already gone? fine!
             logger.debug('Received close signal from peer, but process %i is already gone', pid)
 
-    def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> SubprocessTransport:
+    async def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, object]) -> SubprocessTransport:
```
allisonkarlitskaya, line 114: ...and maybe this one too, although it could be more work.

src/cockpit/channels/stream.py
allisonkarlitskaya: In fact, after you do that, you might find that the unix channel and the tcp channel end up looking fairly similar...

allisonkarlitskaya (commented):
Please see if you can merge unix and tcp channels.  If you specify no restrictions then it's guaranteed to be matched *after* the subprocess one (which has restrictions).

src/cockpit/channels/stream.py
martinpitt: *nod*, I thought the same, I just wanted to wait for a first round of CI. I merged both into a single class now. This has the advantage that we can give better error messages for wrong options.

src/cockpit/channels/stream.py
martinpitt: [subprocess.exec()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec) looks nice, but it doesn't support the `pty`/`windows` options. Perhaps this can be injected or subclassed, but this smells more like a follow-up?

src/cockpit/channel.py
```diff
@@ -247,9 +248,22 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, o
         raise NotImplementedError
 
     def do_open(self, options):
+        self.freeze_endpoint()
```
allisonkarlitskaya, line 251: I have to say I'm not super thrilled about this, for the non-async case...

that being said, I think our eventual plan ought to be that all Transport subclass creation is *async*, so I think it does make sense....

src/cockpit/channel.py
```diff
@@ -247,9 +248,22 @@ def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[str, o
...
+        self._create_transport_task.add_done_callback(self.create_transport_done)
+
+    def create_transport_done(self, task):
+        assert task is self._create_transport_task
+        self._create_transport_task = None
+        try:
+            transport = task.result()
+        except ChannelError as exc:
+            self.close(**exc.kwargs)
+            return
```
allisonkarlitskaya, line 263: This is frustrating, but definitely correct.

test/pytest/test_transport.py
```diff
@@ -94,109 +93,6 @@ async def eof_and_exited_with_code(self, returncode) -> None:
...
-        self.writer.write_a_lot()
-        assert self.writer.transport
-        self.writer.transport.close()
-        assert self.writer.transport
-        assert self.writer.transport.is_closing()
-        await self.read_to_end()
-        assert self.writer.transport is None
-        assert self.reader.transport is None
-
-    async def test_write_backlog_eof_and_close(self) -> None:
```
allisonkarlitskaya, line 189: These are the testcases I was hoping not to lose.  They test general behaviour of the Transport class.  It would be really nice if we could figure out a way to keep some of them alive...

allisonkarlitskaya (approved):
The drop in tests is a bit disappointing, but it's not the end of the world.  97% → 94% in `src/cockpit/transports.py`.  Feel free to ignore that and merge, if you like.

test/pytest/test_transport.py
martinpitt: Okay, I'll bring them back in a follow-up.

PR #18288, opened 2023-02-03
python: Fix fslist1 FileNotFoundError crash

martinpitt:
Trying to list a nonexisting directory crashed the bridge. Report a    
proper channel problem instead.    
    
Add unit tests.    

----

- [x] Builds on top of PR #18285

This fixes the last [cockpit-machines failure](https://github.com/cockpit-project/cockpit-machines/pull/935). I ran `TestMachinesFilesystems.testBasicSystemConnection` with this fix and it passes now.

### DIFF
--- a/src/cockpit/channels/filesystem.py
+++ b/src/cockpit/channels/filesystem.py
@@ -71,7 +71,17 @@ class FsListChannel(Channel):
         if watch:
             raise ChannelError('not-supported', message='watching is not implemented, use fswatch1')
 
-        for entry in os.scandir(path):
+        try:
+            scan_dir = os.scandir(path)
+        except OSError as error:
+            if isinstance(error, FileNotFoundError):
+                problem = 'not-found'
+            elif isinstance(error, PermissionError):
+                problem = 'access-denied'
+            else:
+                problem = 'internal-error'
+            raise ChannelError(problem, message=str(error)) from error
+        for entry in scan_dir:
             self.send_entry("present", entry)
 
         if not watch:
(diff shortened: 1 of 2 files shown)


KKoukiou (approved):
Looks good to me and that's straightforward but since it's the first time I review pbridge code let's wait for one more review.

src/cockpit/channels/filesystem.py
```diff
@@ -71,7 +71,17 @@ def do_open(self, options):
...
             raise ChannelError('not-supported', message='watching is not implemented, use fswatch1')
 
-        for entry in os.scandir(path):
+        try:
+            scan_dir = os.scandir(path)
+        except OSError as error:
+            if isinstance(error, FileNotFoundError):
+                problem = 'not-found'
+            elif isinstance(error, PermissionError):
+                problem = 'access-denied'
```
allisonkarlitskaya, line 80: I guess in these two cases, including the formatted strerror is somewhat redundant.

Also: in all cases, including the filename would probably be helpful.

allisonkarlitskaya (approved):
That said, this is an obvious improvement, so feel free to merge it.

src/cockpit/channels/filesystem.py
martinpitt: The file name is part of the standard exception message actually:

```
❱❱❱ cpf open fslist1 path=dir1/ watch=False : sleep 1 | cpy |cat

{
  "channel": "ch1",
  "command": "close",
  "message": "[Errno 2] No such file or directory: 'dir1/'",
  "problem": "not-found"
}
```
I think that's good enough?

PR #18290, opened 2023-02-06
lib: cockpit-po-plugin: allow also msgids from 'src/' directory

### LINKED ISSUE #18286: ef0ca5d66fdeff5ae11e2253d47be10c97813167 broke translations for external projects
### Explain what happens

The problematic check is:  `if (!references.some(str => str.startsWith(`pkg/${this.subdir}`)))`

which expects translated strings to only exist in pkg/* path, which is not true for any of the external projects. (the whole code lives in src/* usually)

This is already affecting cockpit-machines which upgraded pkg/lib with https://github.com/cockpit-project/cockpit-machines/pull/934

KKoukiou:
This is basically all translations of plugins that live in external repos.

Fixes #18286

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -88,7 +88,7 @@ module.exports = class {
                 for (const [msgid, translation] of Object.entries(context)) {
                     /* Only include msgids which appear in this source directory */
                     const references = translation.comments.reference.split(/\s/);
-                    if (!references.some(str => str.startsWith(`pkg/${this.subdir}`)))
+                    if (!references.some(str => str.startsWith(`pkg/${this.subdir}`) || str.startsWith('src')))
                         continue;
 
                     if (translation.comments.flag && translation.comments.flag.match(/\bfuzzy\b/))


pkg/lib/cockpit-po-plugin.js
```diff
@@ -88,7 +88,7 @@ module.exports = class {
                 for (const [msgid, translation] of Object.entries(context)) {
                     /* Only include msgids which appear in this source directory */
                     const references = translation.comments.reference.split(/\s/);
-                    if (!references.some(str => str.startsWith(`pkg/${this.subdir}`)))
+                    if (!references.some(str => str.startsWith(`pkg/${this.subdir}`) || str.startsWith('src')))
```
jelly, line 91: This seems wrong to me, we want to include references from `src` for plugins. With this change we end up skipping translations from src. I wonder how this used to work as it required less magic in the past?

jelly (changes requested):

pkg/lib/cockpit-po-plugin.js
KKoukiou: We don't skip. see the condition again: `if (!maching_pattern) continue`

martinpitt (approved):
This is too specialized for my taste. The pkg/* case is specific to cockpit.git (as it builds multiple webpacks), for any other project it should just be "everything". But this unblocks releases and updating test/common, so let's go with that. Thanks!

jelly:
> This is too specialized for my taste. The pkg/* case is specific to cockpit.git (as it builds multiple webpacks), for any other project it should just be "everything". But this unblocks releases and updating test/common, so let's go with that. Thanks!

I am more wondering if I we should revert back the old magic which figured out how to do the right way by parsing some build information.

KKoukiou:
> > This is too specialized for my taste. The pkg/* case is specific to cockpit.git (as it builds multiple webpacks), for any other project it should just be "everything". But this unblocks releases and updating test/common, so let's go with that. Thanks!
> 
> I am more wondering if I we should revert back the old magic which figured out how to do the right way by parsing some build information. This commit [ef0ca5d](https://github.com/cockpit-project/cockpit/commit/ef0ca5d66fdeff5ae11e2253d47be10c97813167)

Do you see any issues with unrelated translated string being present in the end po.js files?

jelly:
> 



> > > This is too specialized for my taste. The pkg/* case is specific to cockpit.git (as it builds multiple webpacks), for any other project it should just be "everything". But this unblocks releases and updating test/common, so let's go with that. Thanks!
> > 
> > 
> > I am more wondering if I we should revert back the old magic which figured out how to do the right way by parsing some build information. This commit [ef0ca5d](https://github.com/cockpit-project/cockpit/commit/ef0ca5d66fdeff5ae11e2253d47be10c97813167)
> 
> Do you see any issues with unrelated translated string being present in the end po.js files?

Huh? Yes if I bump COCKPIT_LIB in podman. What I mean to say is, we didn't hardcode `src` before but after we dropped `this.reference_patterns`. So this basically means we don't translate anything outside of `src` or `pkg`.

pkg/lib/cockpit-po-plugin.js
jelly: Right, I thought it was outside of the `.some`...

PR #18291, opened 2023-02-06
Support manifest conditions

martinpitt:
These pave the way for bundling packages together with the webserver, and only showing them if the target system actually supports the corresponding functionality. For example, CoreOS does not have libvirt, so we can bundle but    
hide the "Machines" page on such systems.    
        
This also allows us to better handle ostree vs. packagekit -- even if a user installs cockpit-ostree on a standard RPM system, they should still see the packagekit version of "Software Updates" (see https://github.com/cockpit-project/cockpit-ostree/issues/295). This is a more flexible approach than the old static `priority` field.        
        
Take inspiration from systemd unit files, which support e.g. `ConditionPathExists=`. The conditions should be declarative and cheap to evaluate. Support file exists/not exists tests for now, which should suffice for most use cases. The format and implementation is easy to extend in the future.                   

This also adds a condition (and integration) test to the packagekit "Software Updates" page.

- [x] ~Implement for C bridge, or decide that we don't actually need/want that~ -- let's skip that
- [x] Add documentation: disabled for now, but we'll get a reminder through a failing test once we switch over

### DIFF
--- a/pkg/packagekit/manifest.json
+++ b/pkg/packagekit/manifest.json
@@ -1,6 +1,11 @@
 {
     "name": "updates",
     "priority": 0,
+    "conditions": [
+        {"path-exists": "/lib/systemd/system/packagekit.service"},
+        {"path-not-exists": "/sysroot/ostree"}
+    ],
+
     "tools": {
         "index": {
             "label": "Software updates",
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -266,6 +266,30 @@ class Package:
             if self.manifest['priority'] <= at_least_prio:
                 return False
 
+        CONDITIONS = {
+            'path-exists': os.path.exists,
+            'path-not-exists': lambda p: not os.path.exists(p),
+        }
+
+        for condition in self.manifest.get('conditions', []):
+            try:
+                (predicate, value), = condition.items()
+            except (AttributeError, ValueError):
+                # ignore manifests with broken syntax
+                logger.warning('invalid condition in %s: %s', self.path, condition)
+                return False
+
+            try:
+                test_fn = CONDITIONS[predicate]
+            except KeyError:
+                # do *not* ignore manifests with unknown predicates, for forward compatibility
+                logger.warning('ignoring unknown predicate in %s: %s', self.path, predicate)
+                continue
+
+            if not test_fn(value):
+                logger.info('Hiding package %s as its %s condition is not met', self.path, condition)
+                return False
+
         return True
 
     def get_content_security_policy(self, origin):
--- a/test/verify/check-packagekit
+++ b/test/verify/check-packagekit
@@ -1097,16 +1097,32 @@ ExecStart=/usr/local/bin/{packageName}
                            systemctl daemon-reload'''.format(system_service))
 
         m.start_cockpit()
-        b.login_and_go("/updates")
+        if os.environ.get("TEST_SCENARIO") != "pybridge":
+            # TODO: conditions not implemented on C bridge; once we drop it, drop this special case and
+            # enable the manifest documentation for conditions in doc/guide/packages.xml
+            self.assertIn("\nupdates", m.execute("cockpit-bridge --packages"))
+            b.login_and_go("/updates")
 
-        # error message present
-        b.wait_in_text(".pf-c-page__main-section .pf-c-empty-state__body", "PackageKit is not installed")
+            # error message present
+            b.wait_in_text(".pf-c-page__main-section .pf-c-empty-state__body", "PackageKit is not installed")
 
-        # update status on front page should be invisible
-        b.go("/system")
-        b.enter_page("/system")
-        b.wait_text(self.update_text, "Loading available updates failed")
-        self.assertIn("exclamation", b.attr(self.update_icon, "class"))
+            # update status on front page should show error
+            b.go("/system")
+            b.enter_page("/system")
+            b.wait_text(self.update_text, "Loading available updates failed")
+            self.assertIn("exclamation", b.attr(self.update_icon, "class"))
+        else:
+            self.assertNotIn("\nupdates", m.execute("cockpit-bridge --packages"))
+            # should not appear in the menu at all
+            self.login_and_go(None)
+            b.wait_in_text("#host-apps .pf-m-current", "Overview")
+            self.assertNotIn("Updates", b.text("#host-apps .pf-m-current"))
+
+            # update status on front page should be invisible
+            b.go("/system")
+            b.enter_page("/system")
+            b.wait_visible(".system-health")
+            self.assertFalse(b.is_present(self.update_text))
 
     @skipDistroPackage()
     @nondestructive
--- a/doc/guide/packages.xml
+++ b/doc/guide/packages.xml
@@ -108,6 +108,16 @@ $ cockpit-bridge --packages
           where there are conflicts. For example given two packages with the same
           <code>name</code> a package is chosen based on its priority.</para></listitem>
       </varlistentry>
+      <!-- TODO: enable this once moving to the Python bridge by default; also update example below
+      <varlistentry>
+        <term>conditions</term>
+        <listitem><para>An optional list of <code>{"predicate": "value"}</code> objects. Cockpit will only
+          consider the package if <emphasis>all</emphasis> conditions are met. Currently supported predicates
+          are <code>path-exists</code> and <code>path-not-exists</code>. Unknown predicates are ignored.
+          This is preferable to using <code>priority</code>, but only available since Cockpit FIXME-RELEASE.</para>
+        </listitem>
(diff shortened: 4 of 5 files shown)


martinpitt:
@marusak , @allisonkarlitskaya : Can you please have a first look and share your opinion about the general design and syntax? I'd like to get an agreement before I start implementing this in the C bridge as well. Or maybe we say "screw it, we just don't support this for the C bridge, and wait until we release the py bridge"? That shouldn't actually be too far out any more.

test/verify/check-packagekit
```diff
@@ -1146,6 +1146,27 @@ ExecStart=/usr/local/bin/{packageName}
...
+            self.skipTest("conditions not implemented on C bridge yet")
+
+        m = self.machine
+        b = self.browser
+
+        self.assertIn("\nupdates", m.execute("cockpit-bridge --packages"))
+
+        # pretend that this is an OSTree image, which fails path-not-exists: /sysroot/ostree
+        self.write_file("/sysroot/ostree", "")
+        self.addCleanup(m.execute, "rm -rf /sysroot")
```
martinpitt, line 1162: actually, make this `rmdir || true` and run before the write_file(). The latter cleans up /sysroot/ostree, but not the /sysroot directory itself.

martinpitt:
Okay, I did some cleanups, added documentation, and a "trap" that will prevent us from forgetting the test case and doc update once we switch over.

pkg/packagekit/manifest.json
```diff
@@ -1,6 +1,11 @@
 {
     "name": "updates",
     "priority": 0,
+    "conditions": {
+        "path-exists": "/etc/PackageKit",
```
allisonkarlitskaya, line 5: This feels wrong.  I might be inclined to delete this path and still expect things to work (with the default config).  Even if that doesn't work today, it might change at some point in the future, possibly by way of alternate implementation of the same API.

src/cockpit/packages.py
```diff
@@ -262,6 +262,27 @@ def check(self, at_least_prio):
...
+        CONDITIONS = {
+            'path-exists': os.path.exists,
+            'path-not-exists': lambda p: not os.path.exists(p),
+        }
+
+        for condition, test_fn in CONDITIONS.items():
+            for value in get_condition_value(condition):
+                if not test_fn(value):
+                    logger.debug('Hiding package %s as its %s %s condition is not met', self.path, condition, value)
+                    return False
```
allisonkarlitskaya, line 284: This is awfully complicated, and I'm not sure I understand what the purpose of it all is.  Do you have a case in mind where all of the conditions of a given type would need to be met?  What about one or more?

I could imagine, for example, a condition would be met if either `/usr/bin/x` *or* `/usr/local/bin/x` are available.

If you'd like to support multiple items to check for then I suggest you drop the "single item" mode (it's complicated and will be annoying to express the type of in the future) and replace the key name with something more descriptive like `"all-paths-exist"` and `"no-paths-exist"`.  It wasn't clear from me from reading the documentation, for example, if a list to `"path-exists"` meant "all" or "any", and even less clear for the `not-exists` case.

In any case, I'd pull the content of the (vastly simplified) nested function into the loop body in the next iteration of this patch.  Or just write it like:

```python3
if not all(os.path.exists(path) for path in conditions.get('all-paths-exist', [])):
    return False
if any(os.path.exists(path) for path in conditions.get('no-paths-exist', [])):
    return False
return True
```

allisonkarlitskaya (changes requested):

pkg/packagekit/manifest.json
martinpitt: It wasn't my preferred choice either admittedly, but there actually aren't any other files with a predictable path between distros. Betwen /usr/libexec/ vs. /usr/lib/packagekit/ or /lib vs. /usr/lib there's an annoying amount of variation. However, `/lib/systemd/system/packagekit.service` actually ought to work, as even on /usr-merged systems there's always a /lib compat symlink anyway. Maybe we also eventually want `unit-exists:`? Asking systemd via d-bus is expensive, but we could just look into /lib/ vs. /usr/lib -- either way, it'd be a nice abstraction whose implementation we could change in the future. WDYT @allisonkarlitskaya ?

src/cockpit/packages.py
martinpitt: What I had in mind was something like "/dev/kvm and /usr/lib/systemd/libvirtd.service both exist" -- but at that point, it's all pretty much made up. As I said I took inspiration from systemd units, whose conditions are conjunctive.

However, agreed that ORed conditions can also make sense. But maybe both my `/usr/lib/systemd/ vs. /lib/systemd vs. /usr/local/lib/systemd` and your `/usr/bin/x or /usr/local/bin/x` are telling us that we should use different checks: `unit-exists` and `program-exists` (using `shutil.which()`)?

Adding "or" *could* be done with having lists instead of strings as entries, but then we'd again need to support both types:

```json
"paths-exist": [
    "/dev/kvm",
    ["/usr/local/lib/systemd/system/libvirtd.service", "/lib/systemd/system/libvirtd.service"]
]
```

but honestly this looks rather ridiculous and error prone. What I'd *really* like to do is to have a list of conditions, each of which have a predicate, like `[ ("path-exists", "/a"), ("unit-exists", "foo.service")]`, but that's not JSON -- requiring two-element lists could work.

So in summary, I don't yet have a brilliant idea about how to map propositional logic onto JSON..

martinpitt: How about this: `"conditions": [ "path:/good", "!path:/bad", "unit:foo.service"]` ? It's not JSON down to the primitives, but not hard to parse and a lot easier to read? That still doesn't address the "or" support, but at least the readability and "single vs. multiple" structure.

martinpitt: Once we go there, we could of course go one step further and `.split('|')` to support alternatives:

    "conditions": [ "path:/good | /also/good", "!path:/bad", "unit:packagekit.service | packagekitd.service"]

@marusak , @allisonkarlitskaya : Opinions?

(OMG, I'm gonna need  a lot of unit tests)

allisonkarlitskaya: I think this is getting more complicated, not less.

I'm also not fond of the pseudo-language embedded in JSON.  Just use JSON for that:

```json
"conditions": [
  "any": [{"path-exists": "foo"}, {"path-exists": "bar"}],
  "all": [ "..." ],
]
```

But really, this is getting extreme.  I'd start with:

```json
"conditions" [
  {"path-exists": "foo"},
  {"path-does-not-exist": "bar"},
]
```

and we can go Turing-complete if and when we find the actual need to do so.

src/cockpit/packages.py
```diff
@@ -201,7 +201,10 @@ def merge_patch(target: object, patch: object) -> object:
         result = dict(target if isinstance(target, dict) else {})
         for name, value in patch.items():
             if value is not None:
-                result[name] = Package.merge_patch(result[name], value)
+                if name in result:
+                    result[name] = Package.merge_patch(result[name], value)
```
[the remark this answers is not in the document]
martinpitt: Right, the former should work as `NoneType` fails the `if not isinstance(target, dict)` check at the top, and then inserts the key into a new `{}`. Sent to PR #18312 while I am reworking this PR.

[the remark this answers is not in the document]
martinpitt: Done in #18312

src/cockpit/packages.py
```diff
@@ -262,6 +262,27 @@ def check(self, at_least_prio):
             if self.manifest['priority'] <= at_least_prio:
                 return False
 
+        # this can be a single value, a list, or not exist at all
+        def get_condition_value(key):
```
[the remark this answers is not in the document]
martinpitt: this function is gone

pkg/packagekit/manifest.json
martinpitt: Changed to test for the systemd unit.

src/cockpit/packages.py
martinpitt: Implemented as above.

martinpitt:
[ah, of course](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18291-20230209-164547-f854af06-fedora-37-pybridge/log.html#329) -- that's fallout from testing for the unit, which now of course does not exist any more. That's the whole raison d'être of this patch.

martinpitt:
[insights failure on rhel4edge](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18291-20230210-044453-6a8fbefb-rhel4edge/log.html#234-2), this smells like a flake. Retrying.

martinpitt:
@allisonkarlitskaya Do you have some time to re-review this week? Thanks!

allisonkarlitskaya (approved):
This is a bit over-engineered still, but I think the semantics are right now.

Thanks!

PR #18292, opened 2023-02-06
python: simplify "kill" logic in router

### LINKED ISSUE #18244: TestActivePages.testBasic fails on Python bridge
### Explain what happens

The test disables preloading of the terminal page, loads the terminal page, spawns `basc -c 'exec -a kitten cat'` and then disables the active terminal page. The terminal iframe goes away, but the bridge does not kill the `kitten` process.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-18239-20230125-083800-1422b606-fedora-37-pybridge/log.html#323

allisonkarlitskaya:
Drop the logic for keeping track of which channels belong to which groups and just call .do_kill() on every endpoint that the router knows about, with the group and host details.

That slow, but 'kill' happens fairly infrequently.
    
Channel now keeps track of its own group on open and implements .do_kill() by calling the .do_close() handling if the group matches. Peer simply directly passes the message on to the peer.

This let's us drop one more `@todo`.
    
Fixes #18244

 - [x]  #18359

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -98,7 +98,9 @@ class Channel(Endpoint):
     payload: ClassVar[str]
     restrictions: ClassVar[Sequence[Tuple[str, object]]] = ()
 
+    # These get filled in from .do_open()
     channel = ''
+    group = ''
 
     # input
     def do_control(self, command, message):
@@ -110,6 +112,7 @@ class Channel(Endpoint):
             self.channel = message['channel']
             if message.get('flow-control'):
                 self._send_pings = True
+            self.group = message.get('group', 'default')
             self.freeze_endpoint()
             self.do_open(message)
         elif command == 'ready':
@@ -135,6 +138,13 @@ class Channel(Endpoint):
             self.do_control(command, message)
         except ChannelError as exc:
             self.close(**exc.kwargs)
+
+    def do_kill(self, host: Optional[str], group: Optional[str]) -> None:
+        if host is not None:
+            return
+        if group is not None and self.group != group:
+            return
+        self.do_close()
 
     # At least this one really ought to be implemented...
     def do_open(self, options):
--- a/src/cockpit/peer.py
+++ b/src/cockpit/peer.py
@@ -161,6 +161,9 @@ class Peer(CockpitProtocolClient, SubprocessProtocol, Endpoint):
     def do_channel_data(self, channel: str, data: bytes) -> None:
         self.write_channel_data(channel, data)
 
+    def do_kill(self, host: Optional[str], group: Optional[str]) -> None:
+        self.write_control(command='kill', host=host, group=group)
+
 
 class PeerRoutingRule(RoutingRule, PeerStateListener):
     config: Dict[str, object]
--- a/src/cockpit/router.py
+++ b/src/cockpit/router.py
@@ -69,7 +69,7 @@ class Endpoint:
     def freeze_endpoint(self):
         assert self.__endpoint_frozen_queue is None
         logger.debug('Freezing endpoint %s', self)
-        self.__endpoint_frozen_queue = ExecutionQueue({self.do_channel_control, self.do_channel_data})
+        self.__endpoint_frozen_queue = ExecutionQueue({self.do_channel_control, self.do_channel_data, self.do_kill})
 
     def thaw_endpoint(self):
         assert self.__endpoint_frozen_queue is not None
@@ -82,6 +82,9 @@ class Endpoint:
         raise NotImplementedError
 
     def do_channel_data(self, channel: str, data: bytes) -> None:
+        raise NotImplementedError
+
+    def do_kill(self, host: Optional[str], group: Optional[str]) -> None:
         raise NotImplementedError
 
     # interface for sending messages
@@ -127,7 +130,6 @@ class RoutingRule:
 class Router(CockpitProtocolServer):
     routing_rules: List[RoutingRule]
     open_channels: Dict[str, Endpoint]
-    groups: Dict[str, str]
     _eof: bool = False
 
     def __init__(self, routing_rules: List[RoutingRule]):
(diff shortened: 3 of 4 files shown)


src/cockpit/channels/dbus.py
```diff
@@ -254,7 +254,7 @@ async def get_ready():
                         self.ready(unique_name=self.owner)
                     else:
                         self.close(problem="not-found")
-            task = asyncio.create_task(get_ready())
+            task = self.create_task(get_ready())
             self.tasks.add(task)
```
mvollmer, line 258: Do we still need `self.tasks` at all?

allisonkarlitskaya: Indeed, we can drop this for a very nice cleanup!  Thanks!

mvollmer:
Here are some superficial thoughts after skimming the code.  I check this all in more detail with some unit tests and will come back to this.

This seems to reply to a `close` control message from JS with a `close` message from the bridge.  (Or is that filtered away somewhere?) This is not how the protocol works, I think.  After either JS or the bridge has sent `close`, there should be silence on a channel.

In particular, after JS has sent `close`, it doesn't expect any replies from D-Bus calls anymore.  AFAICS, the DBusClient.close() method in cockpit.js will artificially cancel all outstanding method replies and unregister the channel from the Transport object, which means that all further messages for that channel are dropped on the floor.  I am not sure if any part of Cockpit will actively complain when receiving a message for a already closed channel.

Given the above, there is not much difference between explicitly cancelling a task that waits for a D-Bus reply, and just letting it run until the reply comes and then preventing it from sending the `reply` message on the channel.  However, if we wait for the cleanup to be done before letting the bridge process exit completely, we might wait an arbitrary amount of time, especially since the default call timeout is infinite. So I guess I vote for cancelling the tasks when a channel closes. Other channel types might have different kinds of never-ending tasks.

mvollmer:
I am generally happy with this! Thanks! Do you want to handle slot cancellation also in this PR?

mvollmer:
> Do you want to handle slot cancellation also in this PR?

It's already here! Ouch.

mvollmer:
TestActivePages.testBasic now fails with the do_kill fix...

martinpitt:
Hmm, TestActivePages failure means that killing a channel did not actually kill the process. That was the very thing that set off this whole endeavour.

mvollmer:
> Hmm, TestActivePages failure means that killing a channel did not actually kill the process. That was the very thing that set off this whole endeavour.

I think this has never worked. The bridge correctly sends SIGTERM to the bash process, but that isn't enough to make it go away. The C bridge also closes stdin of the process.

mvollmer:
With [Close subprocess transport when closing stream channel](https://github.com/cockpit-project/cockpit/pull/18292/commits/d4f1b36bbd0c659cc2ae3ea4c82355b263106f95)"  TestActivePages seems green.

allisonkarlitskaya:
> With [Close subprocess transport when closing stream channel](https://github.com/cockpit-project/cockpit/pull/18292/commits/d4f1b36bbd0c659cc2ae3ea4c82355b263106f95) TestActivePages seems green.

I don't like this commit, and I think a proper fix to #18340 would be a better approach.  I split off #18356 to see if we can get the non-dependant bits landed and sort out the rest later.

src/cockpit/channel.py
```diff
@@ -98,6 +98,7 @@ class Channel(Endpoint):
     payload: ClassVar[str]
     restrictions: ClassVar[Sequence[Tuple[str, object]]] = ()
 
+    group = 'default'
```
martinpitt, line 101: This looks dangerous. I understand these as class variables, but it should be per-instance. It actually *seems* to work (with a little toy example), so maybe these are magically instantiated afresh for each new object instance?

But either way, I think this should go. The group name already gets assigned with a "default" default, so it's redundant.

martinpitt (changes requested):
Some nitpicks (or at least questions) about the code, but in general this looks nice to me. Thanks!

src/cockpit/channel.py
```diff
@@ -136,6 +138,13 @@ def do_channel_control(self, channel, command, message):
         except ChannelError as exc:
             self.close(**exc.kwargs)
 
+    def do_kill(self, host: Optional[str], group: Optional[str]) -> None:
+        if host is not None:
+            return
+        if self.group is not None and self.group != group:
```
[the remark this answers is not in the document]
martinpitt: Err, but I think we need to cover for `group is None`, and then always kill; see [protocol](https://github.com/cockpit-project/cockpit/blob/main/doc/protocol.md#command-kill). Perhaps that's what you actually intended to write?

src/cockpit/channel.py
allisonkarlitskaya: The way it works: each instance has a dictionary that contains all of its instance variables.  If you do a lookup of an instance variable, and it's not in the dictionary, it falls back to looking it up on the class.

I mostly assign the value I do here in order to make sure the typing is right.  I could also say `group: str` but then I should theoretically initialize it from `__init__()` instead of `do_open()`...

See also: channel is assigned a default value of `''` in the same way, and we do this with `= None` all over the place for variables that are meant as instance variables.

src/cockpit/channel.py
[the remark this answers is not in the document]
allisonkarlitskaya: This condition is false.  I had fixed that up in another commit, which apparently didn't make it in here.  This is meant to read `if group is not None ...` since not specifying a group on a kill message means "kill everything".

src/cockpit/channel.py
martinpitt: Thanks for the explanation wrt. instance dictionary! This makes total sense. Fix LGTM!

src/cockpit/channel.py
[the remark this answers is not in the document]
martinpitt: Ack, this looks a lot better now.

martinpitt (approved):
If the bots are happy, so am I!

PR #18295, opened 2023-02-06
python: Fix/ignore mypy errors, add mypy to CI

martinpitt:
This reduces `python3 -m mypy src/cockpit` from 95 to 7 errors.

 - [x] #18306

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,10 @@ include = [
 "cockpit.data" = ["*.html"]
 
 [tool.mypy]
-mypy_path = 'src'
+mypy_path = 'src/cockpit'
+[[tool.mypy.overrides]]
+module = ["systemd_ctypes.*"]
+follow_imports = 'silent'
 
 [tool.pylint]
 max-line-length = 118
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -308,6 +308,7 @@ class ProtocolChannel(Channel, asyncio.Protocol):
         """
         self._close_on_eof = True
         if self._eof:
+            assert self._transport is not None
             self._transport.close()
 
     def eof_received(self) -> bool:
--- a/src/cockpit/channels/filesystem.py
+++ b/src/cockpit/channels/filesystem.py
@@ -95,6 +95,11 @@ class FsReadChannel(GeneratorChannel):
     def do_yield_data(self, options: Dict[str, object]) -> GeneratorChannel.DataGenerator:
         binary = options.get('binary', False)
         max_read_size = options.get('max_read_size')
+        # TODO: generic JSON validation
+        if max_read_size is not None and not isinstance(max_read_size, int):
+            raise ChannelError('protocol-error', message='max_read_size must be an integer')
+        if not isinstance(options['path'], str):
+            raise ChannelError('protocol-error', message='path is not a string')
 
         logger.debug('Opening file "%s" for reading', options['path'])
 
--- a/src/cockpit/channels/http.py
+++ b/src/cockpit/channels/http.py
@@ -128,6 +128,7 @@ class HttpChannel(Channel):
 
     def do_open(self, options):
         logger.debug('open %s', options)
+        # TODO: generic JSON validation
         if not options.get('method'):
             self.close(problem='protocol-error', message='missing or empty "method" field in HTTP stream request')
             return
--- a/src/cockpit/channels/metrics.py
+++ b/src/cockpit/channels/metrics.py
@@ -100,7 +100,7 @@ class InternalMetricsChannel(AsyncChannel):
             else:
                 metrics.append({
                     'name': metricinfo.desc.name,
-                    'derive': metricinfo.derive,
+                    'derive': metricinfo.derive,  # type: ignore
                     'units': metricinfo.desc.units,
                     'semantics': metricinfo.desc.semantics
                 })
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -22,7 +22,7 @@ import os
 import signal
 import subprocess
 
-from typing import Dict, Optional
+from typing import Dict
 
 
 from ..channel import ProtocolChannel, ChannelError
@@ -52,16 +52,22 @@ class SocketStreamChannel(ProtocolChannel):
             # Unix
             if 'unix' in options:
                 path = options['unix']
+                # TODO: generic JSON validation
+                if not isinstance(path, str):
+                    raise ChannelError('protocol-error', message='unix option must be a string')
                 label = f'Unix socket {path}'
                 transport, _ = await loop.create_unix_connection(lambda: self, path)
 
             # TCP
             elif 'port' in options:
                 try:
-                    port: int = int(options['port'])
+                    port: int = int(options['port'])  # type: ignore
                 except ValueError:
                     raise ChannelError('protocol-error', message='invalid "port" option for stream channel')
                 host = options.get('address', 'localhost')
+                # TODO: generic JSON validation
+                if not isinstance(host, str):
+                    raise ChannelError('protocol-error', message='"address" option for stream channel must be a string')
                 label = f'TCP socket {host}:{port}'
 
                 transport, _ = await loop.create_connection(lambda: self, host, port)
@@ -77,6 +83,7 @@ class SocketStreamChannel(ProtocolChannel):
                 problem = 'terminated'
             raise ChannelError(problem, message=str(error)) from error
         self.close_on_eof()
+        assert isinstance(transport, asyncio.Transport)
         return transport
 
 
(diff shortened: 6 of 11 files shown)


pyproject.toml
```diff
@@ -22,7 +22,10 @@ include = [
 "cockpit.data" = ["*.html"]
 
 [tool.mypy]
-mypy_path = 'src'
+mypy_path = 'src/cockpit'
+[[tool.mypy.overrides]]
+module = ["systemd_ctypes.*"]
+follow_imports = 'silent'
```
jelly, line 28: I need to remind myself, that I have correctly generated PYI files then we can get rid of this. But I assume that will take some back n forth with lis about how to keep those definitions in sync.

martinpitt:
The next most annoying block are the many dozen lines

    src/cockpit/samples.py:68: error: Unsupported target for indexed assignment ("Union[float, Dict[str, Union[float, int, None]]]")

This is https://github.com/python/mypy/issues/2900 . I didn't find a good workaround, so I'll give up and use `Any` for now with a comment. This leaves 9 errors:
```
src/cockpit/peer.py:186: error: "object" has no attribute "items"
src/cockpit/peer.py:199: error: Value of type "object" is not indexable
src/cockpit/peer.py:201: error: Argument 2 to "Peer" has incompatible type "object"; expected "str"
src/cockpit/peer.py:202: error: Argument 1 to "spawn" of "Peer" has incompatible type "object"; expected "Sequence[str]"
src/cockpit/peer.py:202: error: Argument 2 to "spawn" of "Peer" has incompatible type "object"; expected "Sequence[str]"
src/cockpit/peer.py:241: error: Unsupported right operand type for in ("object")
src/cockpit/superuser.py:181: error: Argument 2 to "go" of "SuperuserRoutingRule" has incompatible type "object"; expected "str"
src/cockpit/superuser.py:200: error: Incompatible types in assignment (expression has type "Tuple[Optional[str], List[Any], List[Any]]", target has type "Tuple[str, Sequence[str], Sequence[str]]")
src/cockpit/channels/metrics.py:103: error: Dict entry 1 has incompatible type "str": "Optional[str]"; expected "str": "Sequence[Any]"
```
I'll have a quick look, but this set is small enough to compare manually or even `# type ignore` for the time being.

martinpitt:
@allisonkarlitskaya, @jelly : mypy behaves really unpredictable for me. Running `python3 -m mypy src`  always works, but still complains about src/systemd_ctypes despite the config change (first commit). OTOH, running `python3 -m mypy src/cockpit`  works the first time and now gives me 7 errors (roughly like in the comment above, I fixed the two in superuser.py); but when I re-run it again *without touching anything*, it crashes with

    AssertionError: Cannot find component 'inotify' for 'systemd_ctypes.inotify.inotify.Event@37'

So it seems we should really call it on the whole `src` folder? But then I still haven't figured out how to make it STFU about src/systemd_ctypes. :exploding_head:

pyproject.toml
martinpitt: But.. why? mypy has the entire source code readily at hand. It complains about things like 

    src/systemd_ctypes/inotify.py:48: error: "Type[inotify.Event@37]" has no attribute "CLOSE"

but how would PYI files fix any of these? Or are these hand-edited?

jelly:
> @allisonkarlitskaya, @jelly : mypy behaves really unpredictable for me. Running `python3 -m mypy src` always works, but still complains about src/systemd_ctypes despite the config change (first commit). OTOH, running `python3 -m mypy src/cockpit` works the first time and now gives me 7 errors (roughly like in the comment above, I fixed the two in superuser.py); but when I re-run it again _without touching anything_, it crashes with
> 
> ```
> AssertionError: Cannot find component 'inotify' for 'systemd_ctypes.inotify.inotify.Event@37'
> ```
> 
> So it seems we should really call it on the whole `src` folder? But then I still haven't figured out how to make it STFU about src/systemd_ctypes. exploding_head

Yes, from what I recall it's really meant to scan your whole project, this is why I had to do annoying tricks for mypy to see every script as a separate module. 

I'll take a look later today and see if I can silence mypy.

pyproject.toml
jelly: mypy does not succeed running over systemd_ctypes either, as it exports classes dynamically.

jelly:
> python3 -m mypy src/cockpit



> @allisonkarlitskaya, @jelly : mypy behaves really unpredictable for me. Running `python3 -m mypy src` always works, but still complains about src/systemd_ctypes despite the config change (first commit). OTOH, running `python3 -m mypy src/cockpit` works the first time and now gives me 7 errors (roughly like in the comment above, I fixed the two in superuser.py); but when I re-run it again _without touching anything_, it crashes with
> 
> ```
> AssertionError: Cannot find component 'inotify' for 'systemd_ctypes.inotify.inotify.Event@37'
> ```
> 
> So it seems we should really call it on the whole `src` folder? But then I still haven't figured out how to make it STFU about src/systemd_ctypes. exploding_head

Same for me:
```
[jelle@t14s][~/projects/cockpit/mypy-fixes]%python3 -m mypy src/cockpit
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/site-packages/mypy/__main__.py", line 37, in <module>
    console_entry()
  File "/usr/lib/python3.10/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "/usr/lib/python3.10/site-packages/mypy/main.py", line 95, in main
    res, messages, blockers = run_build(sources, options, fscache, t0, stdout, stderr)
  File "/usr/lib/python3.10/site-packages/mypy/main.py", line 174, in run_build
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/usr/lib/python3.10/site-packages/mypy/build.py", line 193, in build
    result = _build(
  File "/usr/lib/python3.10/site-packages/mypy/build.py", line 276, in _build
    graph = dispatch(sources, manager, stdout)
  File "/usr/lib/python3.10/site-packages/mypy/build.py", line 2903, in dispatch
    process_graph(graph, manager)
  File "/usr/lib/python3.10/site-packages/mypy/build.py", line 3284, in process_graph
    process_fresh_modules(graph, prev_scc, manager)
  File "/usr/lib/python3.10/site-packages/mypy/build.py", line 3365, in process_fresh_modules
    graph[id].fix_cross_refs()
  File "/usr/lib/python3.10/site-packages/mypy/build.py", line 2112, in fix_cross_refs
    fixup_module(self.tree, self.manager.modules, self.options.use_fine_grained_cache)
  File "/usr/lib/python3.10/site-packages/mypy/fixup.py", line 53, in fixup_module
    node_fixer.visit_symbol_table(tree.names, tree.fullname)
  File "/usr/lib/python3.10/site-packages/mypy/fixup.py", line 111, in visit_symbol_table
    stnode = lookup_fully_qualified(
  File "/usr/lib/python3.10/site-packages/mypy/lookup.py", line 49, in lookup_fully_qualified
    assert key in names, f"Cannot find component {key!r} for {name!r}"
AssertionError: Cannot find component 'inotify' for 'systemd_ctypes.inotify.inotify.Event@37'
```

jelly:
Not sure why everyone in the Cockpit team always uses `python -m $thing` but `mypy src` versus `mypy src/cockpit` also gives fun results. So my take is:

mypy src => fails with systemd_ctypes errors, as `mypy` requires an target module, packages, files. So it probably uses the last option for `files`.

mypy src/cockpit this is a module and it knows it from `pyproject.toml` so the rules apply.

This is all terribly confusing imo.

jelly:
> Not sure why everyone in the Cockpit team always uses `python -m $thing` but `mypy src` versus `mypy src/cockpit` also gives fun results. So my take is:
> 
> mypy src => fails with systemd_ctypes errors, as `mypy` requires an target module, packages, files. So it probably uses the last option for `files`.
> 
> mypy src/cockpit/ this is a module and it knows it from `pyproject.toml` so the rules apply. (Oh and not src/cockpit .... )
> 
> This is all terribly confusing imo and like other Python typing shenigans it feels like it is created in a vacuum where they expect all projects to be a perfect PEP517 thing.
> 
> (BTW this is also not going to work I assume with our `test/static-code` hacks.... which reminds me of https://github.com/cockpit-project/cockpit/pull/17644/files#r948049697
> 
> ```
> [jelle@t14s][~/projects/cockpit/mypy-fixes]%mypy  src/cockpit/bootloader.py src/cockpit/bridge.py src/cockpit/buffered.pyfragment src/cockpit/channel.py src/cockpit/channel
> s src/cockpit/config.py src/cockpit/__init__.py src/cockpit/internal_endpoints.py src/cockpit/packages.py src/cockpit/peer.py src/cockpit/print.py src/cockpit/protocol.py s
> rc/cockpit/remote.py src/cockpit/router.py src/cockpit/samples.py src/cockpit/superuser.py src/cockpit/transports.py
> Traceback (most recent call last):
>   File "/usr/bin/mypy", line 8, in <module>
>     sys.exit(console_entry())
> <snip>
> AssertionError: Cannot find component 'inotify' for 'systemd_ctypes.inotify.inotify.Event@37'
> ```
> 
> So this again ignores our pyproject.toml....

Thinking longer about it, this sort of makes sense, as systemd_ctypes lives in src, just specyfing src will parse both and ignores pyproject.toml

martinpitt:
@jelly: Thanks for your investigations! I poked the [docs](https://mypy.readthedocs.io/en/stable/config_file.html) again and played around with various combinations of `exclude`, `follow_imports = 'skip'` etc., but it's like talking to a wall. There is just no way to make `mypy src` respect the pyproject config or excludes. The only thing that works is `src/cockpit`, and for that the pyproject.toml changes work. So I'm afraid I have no better way to suppress systemd_ctypes at the moment.

Of course I'm happy to drop that commit and do `mypy | grep -v systemd_ctypes`, but I was hoping to get it clean so that we can put it into CI..

martinpitt:
Summary from IRC discussions:
 - Mark the `isinstance()` with something greppable and describe in commit message, so that we can replace them in the future with some generalized JSON validation
 - ignore the remaining 9 errors, so that we can put this into CI
 - include mypy into `pytest` (test/static-code must work on partial checkouts)

martinpitt:
@jelly , @allisonkarlitskaya : So there is one remaining and very important question to answer: [is this](https://github.com/cockpit-project/cockpit/actions/runs/4124779500/jobs/7124461230#step:4:331) my pie, or your pie, or do we have to share it? :pie:

allisonkarlitskaya (approved):
Ship it!!

PR #18297, opened 2023-02-07
python: Fixes for Python 3.6

martinpitt:
With that, I can  run `PYTHONPATH=src python3 -m cockpit.bridge --packages` in a centos 8 container without crashing any more:
```
# PYTHONPATH=src python3 -m cockpit.bridge --packages
cockpit.packages-WARNING: Could not detect libexecdir
base1                                                         /usr/share/cockpit/base1
kdump                                                         /usr/share/cockpit/kdump
metrics                                                       /usr/share/cockpit/metrics
network                                                       /usr/share/cockpit/networkmanager
performance                                                   /usr/share/cockpit/tuned
selinux                                                       /usr/share/cockpit/selinux
shell                                                         /usr/share/cockpit/shell
sosreport                                                     /usr/share/cockpit/sosreport
ssh                                                           /usr/share/cockpit/ssh
system                                                        /usr/share/cockpit/systemd
users                                                         /usr/share/cockpit/users
checksum = 3f17958463c66de080c7377ab561c9d5a4b527ca9563bf8601db411feccd47dc
```

I can also open channels:
```
# (cat <<EOF; sleep 2) |COCKPIT_DEBUG=all PYTHONPATH=src python3 -m cockpit.bridge | cat
64

{
  "command": "init",
  "host": "localhost",
  "version": 1
}
105

{
  "command": "open",
  "channel": "ch1",
  "payload": "fslist1",
  "path": "/etc",
  "watch": false
}
EOF
```
gets me lots of
```
ch1
{
  "event": "present",
  "path": "polkit-1",
  "type": "directory"
}
```

I also validated that I did not break mypy, I get the same 95 errors as on main. (Also see #18295).

Unfortunately running pytest is still very broken. There is no `unittest.IsolatedAsyncioTestCase`, `TypeError: glob() got an unexpected keyword argument 'root_dir'` and other goodies. But this should at least be good enough to run the integration tests.

 - [x] https://github.com/allisonkarlitskaya/systemd_ctypes/pull/49

### DIFF
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -27,7 +27,7 @@ import sys
 
 from typing import Dict, Iterable, List, Tuple, Type
 
-from systemd_ctypes import EventLoopPolicy, bus
+from systemd_ctypes import EventLoopPolicy, bus, run_async
 
 from .channel import ChannelRoutingRule
 from .channels import CHANNEL_TYPES
@@ -137,7 +137,9 @@ async def run(args) -> None:
 
     logger.debug('Starting the router.')
     router = Bridge(args)
-    StdioTransport(asyncio.get_running_loop(), router)
+    loop = asyncio.get_running_loop()
+    loop.set_debug(args.debug)
+    StdioTransport(loop, router)
 
     logger.debug('Startup done.  Looping until connection closes.')
 
@@ -216,7 +218,8 @@ def main() -> None:
         print(json.dumps(Packages().get_bridge_configs(), indent=2))
     else:
         asyncio.set_event_loop_policy(EventLoopPolicy())
-        asyncio.run(run(args), debug=args.debug)
+        # asyncio.run() shim for Python 3.6 support
+        run_async(run(args))
 
 
 if __name__ == '__main__':
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -44,7 +44,7 @@ logger = logging.getLogger(__name__)
 
 
 class InternalBus:
-    exportees: list[bus.Slot]
+    exportees: List[bus.Slot]
 
     def __init__(self, exports: Iterable[Tuple[str, Type[bus.BaseObject]]]):
         client_socket, server_socket = socket.socketpair()
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -372,9 +372,11 @@ class AsyncChannel(Channel):
             self.close(**exc.kwargs)
 
     async def read(self):
-        while not isinstance(item := await self.receive_queue.get(), bytes):
+        while True:
+            item = await self.receive_queue.get()
+            if isinstance(item, bytes):
+                return item
             self.send_pong(item)
-        return item
 
     async def write(self, data):
         if not self.send_data(data):
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -15,8 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from __future__ import annotations
-
 import asyncio
 import logging
 
@@ -29,9 +27,9 @@ logger = logging.getLogger(__name__)
 
 
 class ChannelRoutingRule(RoutingRule):
-    table: Dict[str, List[Type[Channel]]]
+    table: Dict[str, List[Type['Channel']]]
 
-    def __init__(self, router: Router, channel_types: List[Type[Channel]]):
+    def __init__(self, router: Router, channel_types: List[Type['Channel']]):
         super().__init__(router)
         self.table = {}
 
@@ -62,7 +60,7 @@ class ChannelRoutingRule(RoutingRule):
         # Everything checked out
         return True
 
-    def apply_rule(self, options: Dict[str, object]) -> Optional[Channel]:
+    def apply_rule(self, options: Dict[str, object]) -> Optional['Channel']:
         assert self.router is not None
 
         payload = options.get('payload')
@@ -421,7 +419,7 @@ class GeneratorChannel(Channel):
     DataGenerator = Generator[bytes, None, Optional[Dict[str, object]]]
     __generator: DataGenerator
 
-    def do_yield_data(self, options: dict[str, object]) -> DataGenerator:
+    def do_yield_data(self, options: Dict[str, object]) -> 'DataGenerator':
         raise NotImplementedError
 
     def do_open(self, options: Dict[str, object]) -> None:
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -103,7 +103,8 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
         return args
 
     def do_options(self, options):
-        if window := options.get('window'):
+        window = options.get('window')
+        if window is not None:
             self._transport.set_window_size(**window)
 
     def do_close(self):
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -149,7 +149,7 @@ class Package:
     PO_JS_RE: ClassVar[Pattern] = re.compile(r'po\.([^.]+)\.js(\.gz)?')
 
     # A built in base set of "English" translations
-    PO_EN_JS: ClassVar[Entity] = b'', mimetypes.guess_type('po.js')
+    PO_EN_JS: ClassVar[Entity] = (b'', mimetypes.guess_type('po.js'))
     BASE_TRANSLATIONS: ClassVar[Entities] = {'en': PO_EN_JS, 'en-us': PO_EN_JS}
 
     files: Entities
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -229,7 +229,8 @@ class Package:
             lower_locale = locale.lower().replace('_', '-')
             self.translations[lower_locale] = data, guessed_type
         else:
-            self.files[rel.removesuffix('.gz')] = data, guessed_type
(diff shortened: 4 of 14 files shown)


martinpitt:
I was getting bold and prepared a centos-8-stream image with that. When I log in, there is *some* life, I see stuff happening in `window.debugging = 'all'`. But blank page and no journal entries (in particular, no crashes). So this will require some more work. But these are good initial steps.

src/cockpit/packages.py
```diff
@@ -149,7 +149,7 @@ class Package:
     PO_JS_RE: ClassVar[Pattern] = re.compile(r'po\.([^.]+)\.js(\.gz)?')
 
     # A built in base set of "English" translations
-    PO_EN_JS: ClassVar[Entity] = b'', mimetypes.guess_type('po.js')
```
allisonkarlitskaya, line 152: What's the story with this change?  Is that invalid syntax in 3.6?

src/cockpit/superuser.py
```diff
@@ -38,24 +36,24 @@
 class SuperuserStartup:
     peer: Peer  # the peer that's being started
 
-    def success(self, rule: SuperuserRoutingRule) -> None:
+    def success(self, rule: 'SuperuserRoutingRule') -> None:
```
allisonkarlitskaya, line 39: I see lots of conflicts in my near future....

src/cockpit/packages.py
```diff
@@ -229,7 +229,8 @@ def add_file(self, item: Path, checksums: List['hashlib._Hash']) -> None:
             lower_locale = locale.lower().replace('_', '-')
             self.translations[lower_locale] = data, guessed_type
         else:
-            self.files[rel.removesuffix('.gz')] = data, guessed_type
+            index = rel[:-3] if rel.endswith('.gz') else rel
```
allisonkarlitskaya, line 232: `index` is a very confusing name.  I'd expect that to be an int used to subscript the string, but it's the resulting string to be used as the key.  Maybe call it `basename` or `key` instead?

src/cockpit/bridge.py
```diff
@@ -208,6 +208,24 @@ def main() -> None:
...
+        assert not hasattr(asyncio, 'create_task')
+        asyncio.create_task = loop.create_task
+
+        assert not hasattr(asyncio, 'run')
+
+        def asyncio_run(args, debug=False):
+            loop.run_until_complete(args)
+
+        asyncio.run = asyncio_run
+
```
allisonkarlitskaya, line 228: This shouldn't be necessary.  We should just call `systemd_ctypes.asyncio_run()`.  That's what it was intended for.

allisonkarlitskaya: Note: these shims have a bug in them.  But it's a systemd_ctypes issue.  Specifically, the `run()` replacement doesn't cancel (and wait on) outstanding asynchronous generators — it only waits on the main task.

src/cockpit/buffered.pyfragment
```diff
@@ -45,7 +45,10 @@ class CockpitBufferedProtocol(CockpitBaseProtocol, asyncio.BufferedProtocol):
 
         # Try to consume some frames
         consumed = 0
-        while (result := self.consume_one_frame(view)) > 0:
+        while True:
```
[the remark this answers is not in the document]
martinpitt: OK, I'll revert it. I'm not sure what this is good for TBH -- it seems to be dead code. This isn't referenced by anything.

src/cockpit/packages.py
martinpitt: Yes, it is:
```
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/src/py36/src/cockpit/bridge.py", line 36, in <module>
    from .packages import Packages, PackagesListener
  File "/src/py36/src/cockpit/packages.py", line 152
    PO_EN_JS: ClassVar[Entity] = b'', mimetypes.guess_type('po.js')
                                    ^
SyntaxError: invalid syntax
```

Also, I really wish that would continue to be the case in recent Python. I stared at this line for several minutes before I finally figured out what it's supposed to mean. The grouping of `=` and `,` is really very unhappy here, it wasn't clear at all that this is a tuple assignment.

src/cockpit/channel.py
```diff
@@ -371,7 +371,10 @@ async def run_wrapper(self, options):
             self.close(**exc.kwargs)
 
     async def read(self):
-        while not isinstance(item := await self.receive_queue.get(), bytes):
+        while True:
+            item = await self.receive_queue.get()
+            if isinstance(item, bytes):
+                break
```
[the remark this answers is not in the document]
martinpitt: Done.

src/cockpit/channels/stream.py
```diff
@@ -96,7 +96,8 @@ def _close_args(self) -> Dict[str, object]:
         return args
 
     def do_options(self, options):
-        if window := options.get('window'):
+        window = options.get('window')
+        if window:
```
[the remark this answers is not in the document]
martinpitt: Fixed.

src/cockpit/protocol.py
```diff
@@ -70,7 +70,8 @@ def frame_received(self, frame):
             except KeyError as exc:
                 raise CockpitProtocolError('control message is missing command field') from exc
 
-            if channel := message.get('channel'):
+            channel = message.get('channel')
+            if channel:
```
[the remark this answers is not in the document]
martinpitt: Fixed

src/cockpit/protocol.py
```diff
@@ -170,7 +171,10 @@ def write_control(self, **kwargs):
     def data_received(self, data):
         try:
             self.buffer += data
-            while (result := self.consume_one_frame(self.buffer)) > 0:
+            while True:
+                result = self.consume_one_frame(self.buffer)
+                if result <= 0:
+                    break
```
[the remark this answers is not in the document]
martinpitt: Done.

src/cockpit/packages.py
martinpitt: "basename" sounds good to me, changed.

martinpitt:
`cockpit.print` is still broken, and it's a bit tricky: The `/` argument separator isn't supported yet in Python 3.6:
```py
def data(self, channel: str, /, data: bytes) -> None:
    ...
def json(self, channel: str, /, **kwargs: object) -> None:
   ...
```

but it can't just be dropped, as it otherwise tries to submit `channel` twice, through kwargs:

     TypeError: Printer.json() got multiple values for argument 'channel'

Not sure yet how to fix this, I'll just avoid it in the smoke test for now.

src/cockpit/superuser.py
martinpitt: Yeah, that's why I'm aiming for putting this into CI. [almost there](https://github.com/cockpit-project/cockpit/actions/runs/4116286201/jobs/7106134717) :grin:

allisonkarlitskaya:
I'm okay with `cockpit.print` being broken on 3.6.  It's a developer tool.

src/cockpit/bridge.py
martinpitt: Done. Works fine, but this is now blocked on the systemd_ctypes PR, marked as such.

martinpitt:
@allisonkarlitskaya : I updated systemd_ctypes to the landed commit and reverted the printf changes. PTAL?

allisonkarlitskaya (dismissed):
Looks good, thanks!

martinpitt:
I messed up the rebase, the systemd_ctypes commit ought to be in the "python: Use systemd_ctype's asyncio.run() shim" commit. Also, this included https://github.com/allisonkarlitskaya/systemd_ctypes/commit/389c90b82996d1b2872e07caca9b7931b5d552e9 which now causes [these failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18297-20230208-093157-c36df50a-fedora-37-pybridge/log.html). Needs fixing in s-c with a unit test, and then updating again.

martinpitt:
I wanted to decouple this PR from the systemd_ctypes update due to the issue above. This will/would take me a few hours to understand and debug, and I have too many tasks in the air already. Setting the debug flag separately is not too evil IMHO. So hopefully this goes green again.

src/cockpit/bridge.py
```diff
@@ -216,7 +216,9 @@ def main() -> None:
         print(json.dumps(Packages().get_bridge_configs(), indent=2))
     else:
         asyncio.set_event_loop_policy(EventLoopPolicy())
-        asyncio.run(run(args), debug=args.debug)
+        asyncio.get_event_loop().set_debug(args.debug)
```
allisonkarlitskaya, line 219: This is pretty evil.  It might not be clear, but this actually creates the loop.  This usage is deprecated.  See https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

We pass the `args` into the run() function anyway so it's probably safer to call `get_running_loop()` from there and `.set_debug()` on that.

allisonkarlitskaya (changes requested):
Sorry for forcing yet another round on this :(

src/cockpit/bridge.py
martinpitt: Thanks for spotting! Reworked, how's this?

martinpitt: interdiff, before the rebase to mypy PR: https://github.com/cockpit-project/cockpit/compare/ef474e1ab55b133724461f3689fac492416bcd8c..6bd43c82e12d4354990ab1e0454586e52dcdf37a

allisonkarlitskaya (approved):
Not sure why, but the label needed to be reset.

allisonkarlitskaya:
This looks good now.  Thanks very much for working on this!

PR #18301, opened 2023-02-07
python3-cockpit RPM packaging

allisonkarlitskaya:
I'm not sure what we're going to do with this, but it seems like it might be useful, either now, or in the future.

Installing the created `python3-cockpit` binary package will uninstall (`Obsoletes`) `cockpit-bridge` and also `Provide` it so that `cockpit-pcp` can remain installed.


allisonkarlitskaya:
So one thing to note: actually installing the python bridge this way is going to break things.  We depend on the `cockpit-askpass` program from the bridge.  We should probably replace that with our own.  Makes me wonder about #18296, though...

allisonkarlitskaya:
> So one thing to note: actually installing the python bridge this way is going to break things. We depend on the `cockpit-askpass` program from the bridge. We should probably replace that with our own. Makes me wonder about #18296, though...

That's #18304.

martinpitt:
Note that cockpit-bridge.rpm also ships the base1 webpack -- pretty well nothing will work without that. This is fixable in principle with a lot of spec hacking (involving webpack-jumpstart, or adding a Source: for our release tarball, etc.), but I honestly think it's impractical to try and install cockpit-bridge.rpm entirely with Python. This will look a lot cleaner once the *real* cockpit-bridge.rpm contains the pybridge, and we integrate that into our top-level build system (autotools for now, but you already know that I laid my eyes on that much younger, more attractive, and much more python-y alternative :wink: )

allisonkarlitskaya:
I think we decided that this isn't going to land any time soon (or ever).  I'm going to close it.  The concept has been proven that our python stuff plays well with the `%pyproject_*` macros.  We can always come back to the closed PR to grab the spec file here to use as the basis for future work.

PR #18303, opened 2023-02-07
tools/venv: build a venv with cockpit installed

allisonkarlitskaya:
This creates a virtual environment in .venv/ and installs the Python package into it, in editable mode.  That lets you run .venv/bin/cockpit-bridge, for example, or activate the environment.

This is a fun toy.  I find it useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ Makefile.in
 # toplevel (/...)
 /*.list
 /.coverage
+/.venv/
 /Coverage
 /aclocal.m4
 /autom4te.cache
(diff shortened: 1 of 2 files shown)


martinpitt:
FTR: If you use it, sure. It's not what I will use nor recommend to anyone else (I'm in the "KISS" camp, and hate TIMTOWTDI and unnecessary dependencies and moving parts), I'd much rather just add a `src/cockpit-bridge` one-liner to run the bridge out of the source tree (I have an alias for that, but that's harder to set up for new developers).

PR #18304, opened 2023-02-07
python: cockpit-askpass

allisonkarlitskaya:
This was a fun little hack.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,7 @@ version = '0'
 
 [project.scripts]
 cockpit-bridge = "cockpit.bridge:main"
+cockpit-askpass = "cockpit.askpass:main"
 
 [tool.setuptools.packages.find]
 where = ["src"]
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -24,6 +24,7 @@ import logging
 import mimetypes
 import os
 import re
+import shutil
 
 from pathlib import Path
 from typing import ClassVar, Dict, List, Optional, Pattern, Tuple
@@ -81,6 +82,11 @@ def get_libexecdir() -> str:
 
 def patch_libexecdir(obj):
     if isinstance(obj, str):
+        if '${libexecdir}/cockpit-askpass' in obj:
+            # extra-special case: we handle this internally
+            abs_askpass = shutil.which('cockpit-askpass')
+            if abs_askpass is not None:
+                return obj.replace('${libexecdir}/cockpit-askpass', abs_askpass)
         return obj.replace('${libexecdir}', get_libexecdir())
     elif isinstance(obj, dict):
         return {key: patch_libexecdir(value) for key, value in obj.items()}
--- a/src/cockpit/askpass.py
+++ b/src/cockpit/askpass.py
@@ -1,0 +1,73 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import argparse
+import asyncio
+import getpass
+import logging
+import os
+import socket
+import sys
+import uuid
+
+from typing import Dict
+
+from .protocol import CockpitProtocol
+
+logger = logging.getLogger(__name__)
+
+
+class AuthorizeInteraction(CockpitProtocol):
+    def __init__(self, prompt: str):
+        self.prompt = prompt
+        self.cookie = str(uuid.uuid4())
+
+    def do_ready(self) -> None:
+        challenge = 'plain1:' + ''.join('%02x' % ord(c) for c in getpass.getuser())
+        self.write_control(command='authorize', challenge=challenge, cookie=self.cookie, prompt=self.prompt)
+
+    def transport_control_received(self, command: str, message: Dict[str, object]) -> None:
+        if command == 'authorize' and message.get('cookie') == self.cookie:
+            print(message.get('response'))
+        else:
+            logger.error('received invalid control message: %s', message)
+
+        assert self.transport is not None
+        self.transport.close()
+
+    async def run(self, connection: socket.socket) -> None:
+        loop = asyncio.get_running_loop()
+        await loop.create_connection(lambda: self, sock=connection)
+        await self.communicate()
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument('prompt')
+    args = parser.parse_args()
+
+    try:
+        connection = socket.socket(fileno=os.dup(0))
+    except OSError:
+        sys.exit('This command must be run with stdin connected to a socket.')
+
+    interaction = AuthorizeInteraction(args.prompt)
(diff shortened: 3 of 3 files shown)


pyproject.toml
```diff
@@ -8,6 +8,7 @@ version = '0'
 
 [project.scripts]
 cockpit-bridge = "cockpit.bridge:main"
+cockpit-askpass = "cockpit.askpass:main"
```
martinpitt, line 11: Can we tell this somehow to install into libexecdir? This doesn't belong into `$PATH`. I suppose not, but this seems easy enough to just do ourselves. This is just a single-line shell wrapper (`python3 -m cockpit.askpass`), or perhaps it can even be done with `#!/usr/bin/python3 ..`?

Perhaps the easiest way is to install it into libexecdir via the Debian/rpm packaging -- they know the destination path already, and an upstream `pip install` isn't something that any user would do, so it doesn't matter much where it goes there.

src/cockpit/packages.py
```diff
@@ -57,31 +57,24 @@ def directory_items(path):
...
-            # give readable error messages
-            LIBEXECDIR = '/nonexistent/libexec'
-
-    return LIBEXECDIR
+LIBEXEC_PATH = [
+    '/usr/local/libexec',
+    '/usr/libexec',
+    '/usr/local/lib/cockpit',
+    '/usr/lib/cockpit',
+    *os.environ['PATH'].split(':')
```
martinpitt, line 65: Not a fan of adding `$PATH` here, but then that whole `${libexecdir}` replacing is only temporary anyway. :grin:  But see the previous comment, we can fix this rather easily.

src/cockpit/packages.py
```diff
@@ -57,31 +57,24 @@ def directory_items(path):
...
 
 def patch_libexecdir(obj):
     if isinstance(obj, str):
-        return obj.replace('${libexecdir}', get_libexecdir())
+        before, expansion, after = obj.partition('${libexecdir}/')
+        if expansion:
+            for path in LIBEXEC_PATH:
+                if os.path.exists(f'{path}/{after}'):
+                    return f'{before}{path}/{after}'
+            logger.warning(f'Unable to expand {expansion}{after}')
```
martinpitt, line 76: This warning now appears in [two tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18304-20230207-220051-8dcd1eca-fedora-37-pybridge/log.html#184), somewhat surprisingly -- we don't allow the former "Could not detect libexecdir" message either.

martinpitt (commented):
Nice, more Python! I suggest to install that into /usr/local/lib/cockpit-askpass for `image-prepare --python`, so that it will win over /usr/lib/cockpit-askpass and doesn't overwrite the rpm (we should do the same for the bridge, but that still requires some test case adjustments).

src/cockpit/packages.py
allisonkarlitskaya: The warning is because of `cockpit-ssh` which is apparently not installed in those tests.  Maybe we need a way to evaluate this only at Peer launch time.  That wouldn't be unprecedented.

allisonkarlitskaya: Although I just realize now why we do it at manifest parse time: the "libexecdir" entry in `systemd/manifest.json` which we need to find the certificate helper script...

allisonkarlitskaya:
So here's the two constraints:

 - I am interested in using pure Python packaging, ideally with as few hacks as possible.  The bridge must function when installed via the "pure python" route.
 - `SUDO_ASKPASS` must refer to a single command which is an absolute path to an executable file on the filesystem

The Python equivalent of `libexecdir` is probably something like having a file present in the package directory and referring to it from there.  This would actually be quite nice, since we could do something like
```python
from . import askpass
ASKPASS=askpass.__file__
```
and just run it directly.  Unfortunately, putting a `#!/usr/bin/python3` line at the top and marking it executable isn't going to cut it, since the installation goes via a wheel file (which is a zip) and zip files don't have executable bits.

The only way to get an executable script installed seems to be via the entry points mechanism, and this seems to only put scripts in `bin/`.  Since this is the very thing we need for `SUDO_ASKPASS` and this is the only way to get it, it sort of ties our hands.


I think the way forward is indeed at the distro packaging level, and we could even provide support for it.  I already start to imagine a situation in which the pure "upstream" version of the bridge expects to find things in `$PATH` but distro packages:

 - divert the location of some files at installation time to `/usr/libexec` or `/usr/lib/cockpit`
 - modify the `cockpit-bridge` wrapper script to pass an extra `libexecdir=` kwarg to `main()` which causes the bridge to look in this directory instead of `$PATH`.

mvollmer (dismissed):
LGTM!

martinpitt (changes requested):
Officially adding my changes request for above's comments.

martinpitt:
I thought about this a little on the way home. I may have misunderstood your intent and the purpose of this. So, writing down my assumptions and what I learned from our discussion today:

 * My gut feeling is that the route of #18301 is too much throw-away work, and will be [rather hacky](https://github.com/cockpit-project/cockpit/pull/18301#issuecomment-1423272960) . → It's more practical to install the py bridge on top of an installed cockpit-bridge.{rpm,deb} rather than entirely replacing it.
 * Once the pybridge replaces the C bridge, cockpit-bridge.rpm will contain the python bridge/c-askpass and have Requires: python3. From the outside, it will provide the same functionality/API as earlier versions.
 * Actually doing this ⬆ is still a few weeks out (making c-ssh work, fixing c-ostree and sub-man, fixing for RHEL8, etc.), so until then we need a *temporary* way of overlaying the py bridge, which (1) is cheap to build, and (2) can be done from external projects (like c-podman).
 * We can't rely on autotools for our external projects. It's too expensive for image-prepare --python and our external Makefiles. (I was missing this part before)
 
 I agree that building/installing a wheel is a nice and elegant way of doing that overlay. It won't be how we *eventually* install the full cockpit-bridge (or other parts of cockpit) -- for that we need to have a stronger top-level build system which can deal with C, webpack, docbook, and whatnot. But it can certainly build-depend and use pip to install the modules and exe wrappers into the install DESTDIR, and then we can move the helpers cockpit-{ssh,askpass} into libexecdir.

So how about this:

 * We by and large use the approach here to install c-askpass into /usr/bin.
 * Ideally it would be /usr/local/bin (OSTree and not clobbering rpm files), but that will break a few integration tests which assume /usr/bin/cockpit-bridge (I'm happy to fix them, it's  not hard)
 * You change your libexecdir detection to check /usr/{,local/}bin *before* /usr/libexec/, so that it will prefer the py version over the C version from the RPM. Add a comment that this is temporary for `image-prepare --pybridge`.

This should help with not making your life unnecessarily hard (in fact, it requires very few changes in this PR),, and it's a temporary hack that won't ever appear in any official rpms/debs or even `make installs` -- that's really my main concern.

WDYT @allisonkarlitskaya ?

allisonkarlitskaya:
> WDYT @allisonkarlitskaya ?

:+1: mach ich

src/cockpit/packages.py
```diff
@@ -81,6 +82,11 @@ def get_libexecdir() -> str:
 
 def patch_libexecdir(obj):
     if isinstance(obj, str):
+        if '${libexecdir}/cockpit-askpass' in obj:
+            # extra-special case: we handle this internally
```
martinpitt, line 86: Heh, this should probably have a `# HACK:` comment to say that this only applies to image-prepare --python`. But then that whole libexecdir thing is (hopefully) temporary, so :man_shrugging:

src/cockpit/packages.py
```diff
@@ -81,6 +82,11 @@ def get_libexecdir() -> str:
...
+            abs_askpass = shutil.which('cockpit-askpass')
```
martinpitt, line 87: I'm rather surprised that this doesn't raise an exception when not found, but it doesn't.

martinpitt (approved):
Thanks!

src/cockpit/packages.py
allisonkarlitskaya: I was equally surprised when my tests were suddenly failing with a message about how .replace() doesn't like `None` as its second argument :)

PR #18305, opened 2023-02-08
python: freeze channels until they send "ready"

allisonkarlitskaya:
Channels should start out in a frozen state until they indicate that they're actually ready to start receiving data.

This fixes issues with messages arriving to channels with async startup processes (like the dbus channel finding its name owner).


src/cockpit/channel.py
```diff
@@ -248,7 +248,6 @@ async def create_transport(self, loop: asyncio.AbstractEventLoop, options: Dict[
         raise NotImplementedError
 
     def do_open(self, options):
-        self.freeze_endpoint()
```
allisonkarlitskaya, line 251: This landed in the wrong commit.  Let's let the bots run before moving it, though.

test/pytest/test_bridge.py
```diff
@@ -553,3 +556,108 @@ async def test_fslist1_notexist(self):
             'fslist1', path='/nonexisting', watch=False,
             problem='not-found',
             reply_keys={'message': "[Errno 2] No such file or directory: '/nonexisting'"})
+
+
+@pytest.mark.parametrize('channeltype', CHANNEL_TYPES)
+def test_channel(channeltype, tmp_path):
```
martinpitt, line 562: Really nice test! :100:

martinpitt (approved):

PR #18312, opened 2023-02-09
python: Fix KeyError on overriding nonexisting manifest keys

martinpitt:
Split out from PR #18291, it's unrelated and a good fix in its own right.


allisonkarlitskaya (approved):
Looks good.  Thanks!

PR #18313, opened 2023-02-09
Accounts page flakes

KKoukiou:


pkg/users/account-details.js
```diff
@@ -352,6 +352,7 @@ export const AccountGroupsSelect = ({ name, loggedIn, groups, setError }) => {
     const [isOpenGroup, setIsOpenGroup] = useState(false);
     const [selected, setSelected] = useState();
     const [primaryGroupName, setPrimaryGroupName] = useState();
+    const [loading, setLoading] = useState(true);
```
martinpitt, line 355: Just in case you need to re-push: The prefix of this commit message should be "users:", not "test:"

martinpitt (approved):
Nice work! There is still [one flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18313-20230209-134755-41b78373-centos-8-stream/log.html#333-2), but I'm happy for that to be a follow-up -- it does not look related to the changes here, and they should already help.

PR #18314, opened 2023-02-09
python: Connecting to specific D-Bus addresses

mvollmer:
This either needs https://github.com/systemd/systemd/pull/26375 or we need to stop using GDBus in test-server for the direct connection.

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -17,8 +17,6 @@
 
 # Missing stuff compared to the C bridge that we should probably add:
 #
-# - connecting to given address instead of bus
-# - some more ways to connect to the internal bus (like { bus: "none", address: "internal" })
 # - removing matches
 # - removing watches
 # - emitting of signals
@@ -221,8 +219,14 @@ class DBusChannel(Channel):
         self.tasks = set()
 
         bus = options.get('bus')
+        address = options.get('address')
 
-        if bus == 'internal':
+        if address is not None:
+            if bus is not None and bus != 'none':
+                raise ChannelError('protocol-error', message='only one of "bus" and "address" can be specified')
+            logger.debug('get bus with address %s for %s', address, self.name)
+            self.bus = Bus.new(address=address, bus_client=self.name is not None)
+        elif bus == 'internal':
             logger.debug('get internal bus for %s', self.name)
             self.bus = self.router.internal_bus.client
         else:
@@ -230,9 +234,11 @@ class DBusChannel(Channel):
                 if bus == 'session':
                     logger.debug('get session bus for %s', self.name)
                     self.bus = Bus.default_user()
-                else:
+                elif bus == 'system' or bus is None:
                     logger.debug('get system bus for %s', self.name)
                     self.bus = Bus.default_system()
+                else:
+                    raise ChannelError('protocol-error', message=f'invalid bus "{bus}"')
             except OSError as exc:
                 raise ChannelError('protocol-error', message=f'failed to connect to {bus} bus: {exc}') from exc
 


mvollmer:
With https://github.com/systemd/systemd/pull/26375, `pytest` is green for me, including test-dbus-address.html.

martinpitt:
@mvollmer : I started looking into rewriting the test dbus mock service as Python, but got stopped pretty early. We first need to teach systemd_ctypes to expose sd_bus_request_name() and friends, and sort out [this regression](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18297-20230208-093157-c36df50a-fedora-37-pybridge/log.html#25) in systemd_ctypes main.

So perhaps we can decouple this, and keep the unit tests skipped for the time being?

mvollmer:
> So perhaps we can decouple this, and keep the unit tests skipped for the time being?

Hmm, if this helps with c-ostree or subman, yes.

jelly:
Grab https://github.com/candlepin/subscription-manager/pull/3204

Apply
```
-COCKPIT_PYBRIDGE_REF = main
+COCKPIT_PYBRIDGE_REF = 1a19020d3cd6a3bbd2738ac17540f5948938fbcd
```


```
TEST_SCENARIO=pybridge TEST_OS=fedora-37 make -f integration-tests/run prepare
TEST_OS=fedora-37 TEST_SCENARIO=pybridge ./test/check-subscriptions -vst TestSubscriptionsPackages.testMissingPackages
````

Test now passes.

**Comment by mvollmer** [comment]
> Test now passes.

Yay!  Thanks!

**Comment by martinpitt** [comment]
Likewise, I tested this against cockpit-ostree in https://github.com/cockpit-project/cockpit-ostree/pull/323 . With current main it [fails pretty badly](https://cockpit-logs.us-east-1.linodeobjects.com/pull-323-20230207-104350-178167c3-fedora-coreos-pybridge/log.html), as any rpm-ostree transaction happens on a direct address. Against this PR it unfortunately still [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-323-20230210-085742-225bab5f-fedora-coreos-pybridge/log.html#3), not even "better", just "differently". I didn't look into this at all yet, I'm currently too deep in another yak shave.

**Comment by mvollmer** [comment]
> Against this PR it unfortunately still [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-323-20230210-085742-225bab5f-fedora-coreos-pybridge/log.html#3), not even "better", just "differently".

Thanks!  The subman progress is good enough to merge this, I'd say.

File: src/cockpit/channels/dbus.py [hunk 1]
```diff
@@ -225,18 +223,28 @@ def do_open(self, options):
...
 
         bus = options.get('bus')
+        address = options.get('address')
 
-        if bus == 'internal':
+        if bus == 'none':
+            if address is None or address == "internal":
+                logger.debug('get internal bus for %s', self.name)
+                self.bus = self.router.internal_bus.client
+            else:
```
allisonkarlitskaya, line 232: does anyone actually do that?

src/cockpit/channels/dbus.py
```diff
@@ -225,18 +223,28 @@ def do_open(self, options):
...
+                logger.debug('get bus with address %s for %s', address, self.name)
+                self.bus = Bus.new(address=address, bus_client=self.name is not None)
```
allisonkarlitskaya, line 234: What happens if both address and bus are given?

src/cockpit/channels/dbus.py
```diff
@@ -225,18 +223,28 @@ def do_open(self, options):
...
             try:
                 if bus == 'session':
                     logger.debug('get session bus for %s', self.name)
                     self.bus = Bus.default_user()
-                else:
+                elif bus == 'system' or bus is None:
                     logger.debug('get system bus for %s', self.name)
                     self.bus = Bus.default_system()
+                else:
+                    raise ChannelError('protocol-error', message=f'invalid bus "{bus}"')
```
allisonkarlitskaya, line 247: OK.

src/cockpit/channels/dbus.py
mvollmer: Good question. The C bridge supports it, but...

src/cockpit/channels/dbus.py
mvollmer: bus must be "none" (the string), otherwise "address" is ignored.

src/cockpit/channels/dbus.py
allisonkarlitskaya: I'd prefer to drop this part.  There's no sense in having two ways to do the same thing, particularly if we don't know that it's actually required.

src/cockpit/channels/dbus.py
allisonkarlitskaya: This seems annoying.  One would expect that having an unspecified bus but a specified address would result in the address being used, but I guess in that case, we should actually choose the system bus.

I think I'd prefer to see

 - a ChannelError in case address and bus are both specified and bus is anything other than "none"
 - connecting via address in case bus is unspecified (as well as the case where it's the string "none")

src/cockpit/channels/dbus.py
mvollmer: Agreed.

src/cockpit/channels/dbus.py
mvollmer: Yes, I like this.  Note also that `{ bus: "none", address: "...", name: "..." }` will actually make a connection to a bus, so insisting on `bus: "none"` in this case doesn't make sense.

martinpitt:
I tested cockpit-ostree against this pybridge PR and your systemd PR! I added the locally built libsystemd to /usr/local/lib, and edited /usr/local/bin/cockpit-bridge to add this to LD_LIBRARY_PATH (unfortunately on ostree,  dlopen() does not seem to consider /usr/local/lib{,64}/ by default). Now all but one tests succeed, and the last one fails on an unrelated issue.

mvollmer:
Fix for GDBus: https://github.com/GNOME/glib/pull/38. test-dbus-address.html is green for me with that as well.

src/cockpit/channels/dbus.py
```diff
@@ -225,18 +223,26 @@ def do_open(self, options):
         self.tasks = set()
 
         bus = options.get('bus')
-
-        if bus == 'internal':
+        address = options.get('address')
+
+        if address is not None:
+            if bus is not None and bus != 'none':
+                raise ChannelError('protocol-error', message=f'only one of "bus" and "address" can be specified')
```
allisonkarlitskaya, line 230: bzzt bzzt

```
 # src/cockpit/channels/dbus.py:226:62: F541 f-string is missing placeholders
```

allisonkarlitskaya (changes requested):
Looks perfect to me, but the bots disagree.

allisonkarlitskaya (approved):

PR #18274, opened 2023-01-31
python: Avoid pending dbus tasks

mvollmer:

PR #18318, opened 2023-02-10
pybridge: Back to the past!

martinpitt:
Once upon a time, there was a sad Python version, called by their peers as "3.6". Few features it had, little love it got from developers. Let that ugly duck grow into a beautiful swan, and learn how to run Cockpit!

Split out PRs:
 - #18297
 - #18320 
 - #18321
 - #18360
 - #18361 obsoleted by #18365
 - #18385
 - #18380 -- Skipped test instead, to avoid making this a blocker

### DIFF
--- a/src/cockpit/askpass.py
+++ b/src/cockpit/askpass.py
@@ -66,7 +66,13 @@ def main() -> None:
         sys.exit('This command must be run with stdin connected to a socket.')
 
     interaction = AuthorizeInteraction(args.prompt)
-    asyncio.run(interaction.run(connection))
+    loop = asyncio.get_event_loop()
+
+    # shim for Python < 3.7
+    if not hasattr(asyncio, 'get_running_loop'):
+        asyncio.get_running_loop = lambda: loop
+
+    loop.run_until_complete(interaction.run(connection))
 
 
 if __name__ == '__main__':
--- a/src/cockpit/bridge.py
+++ b/src/cockpit/bridge.py
@@ -29,6 +29,8 @@ from typing import Dict, Iterable, List, Tuple, Type
 
 from systemd_ctypes import bus, run_async
 
+from . import polyfills
+
 from .channel import ChannelRoutingRule
 from .channels import CHANNEL_TYPES
 from .config import Config, Environment
@@ -192,6 +194,8 @@ def setup_logging(debug: bool):
 
 
 def main() -> None:
+    polyfills.install()
+
     # The --privileged bridge gets spawned with its stderr being consumed by a
     # pipe used for reading authentication-related message from sudo.  The
     # absolute first thing we want to do is to recover the original stderr that
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -282,7 +282,15 @@ class DBusChannel(Channel):
             func = handler
         r_string = ','.join(f"{key}='{value}'" for key, value in r.items())
         if not self.is_closing():
-            self.matches.append(self.bus.add_match(r_string, func))
+            # this gets an EINTR very often especially on RHEL 8
+            while True:
+                try:
+                    match = self.bus.add_match(r_string, func)
+                    break
+                except InterruptedError:
+                    pass
+
+            self.matches.append(match)
 
     def add_async_signal_handler(self, handler, **kwargs):
         def sync_handler(message):
--- a/src/cockpit/polyfills.py
+++ b/src/cockpit/polyfills.py
@@ -1,0 +1,42 @@
+# This file is part of Cockpit.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import socket
+
+
+def install():
+    '''Add shims for older Python versions'''
+
+    # introduced in 3.9
+    if not hasattr(socket, 'recv_fds'):
+        import array
+        import _socket
+
+        def recv_fds(sock, bufsize, maxfds, flags=0):
+            fds = array.array("i")
+            msg, ancdata, flags, addr = sock.recvmsg(bufsize, _socket.CMSG_LEN(maxfds * fds.itemsize))
+            for cmsg_level, cmsg_type, cmsg_data in ancdata:
+                if (cmsg_level == _socket.SOL_SOCKET and cmsg_type == _socket.SCM_RIGHTS):
+                    fds.frombytes(cmsg_data[:len(cmsg_data) - (len(cmsg_data) % fds.itemsize)])
+            return msg, list(fds), flags, addr
+
+        socket.recv_fds = recv_fds
+
+        def send_fds(sock, buffers, fds, flags=0, address=None):
+            return sock.sendmsg(buffers, [(_socket.SOL_SOCKET, _socket.SCM_RIGHTS, array.array("i", fds))])
+
+        socket.send_fds = send_fds
--- a/test/verify/check-storage-stratis
+++ b/test/verify/check-storage-stratis
@@ -17,10 +17,14 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+import os
+import unittest
+
 import parent  # noqa: F401
 from packagelib import PackageCase
 from storagelib import StorageCase
 from testlib import skipImage, test_main
+from testvm import DEFAULT_IMAGE
 
 
 @skipImage("No Stratis", "debian-*", "ubuntu-*")
@@ -197,6 +201,10 @@ class TestStorageStratis(StorageCase):
         m.execute("! findmnt /run/fsys2-copy")
         m.execute("! findmnt /run/fsys3")
 
+    @unittest.skipIf(
+        os.environ.get("TEST_SCENARIO") == "pybridge" and
+        (DEFAULT_IMAGE.startswith("centos-8") or DEFAULT_IMAGE.startswith("rhel-8")),
+        "requires https://github.com/cockpit-project/cockpit/pull/18380")
     def testEncrypted(self):
         m = self.machine
         b = self.browser


martinpitt:
@mvollmer Can you please have a close look at the "Defer asyncio.Lock() creation for D-Bus" commit and check my assumption? Was this *indended* to be a single global lock across all channels, and does it have to be? Or is a per-channel instance sufficient? In the former case, I'd initialize the class variable to `None` and lazily create the lock in the first opened D-Bus channel.

@allisonkarlitskaya You may have an opinion about where to put the python stdlib backports?

I tried this interactively with a c8s `--python` image and also got some successful integration tests, so let's do a first run of [centos-8-stream/pybridge](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230210-075939-e2e439f3-centos-8-stream-pybridge/log.html) here. Note that this is still exempt from test statistics, so in theory we could land it red.

src/cockpit/channel.py
```diff
@@ -389,7 +390,10 @@ def do_resume_send(self) -> None:
 
     def do_open(self, options):
         self.receive_queue = asyncio.Queue()
-        asyncio.create_task(self.run_wrapper(options), name=f'{self.__class__.__name__}.run_wrapper({options})')
+        if sys.version_info >= (3, 8):
+            asyncio.create_task(self.run_wrapper(options), name=f'{self.__class__.__name__}.run_wrapper({options})')
+        else:
+            asyncio.create_task(self.run_wrapper(options))
```
martinpitt, line 396: I don't think cockpit-askpass uses that, so this could also get a polyfill wrapper.

allisonkarlitskaya:
I was having a think about this.

I think I'd like to do something like this:

`polyfills.py`:
```python
import socket as _socket
from typing import List

class socket(_socket.socket):
    def recv_fds(...):
        ....

class list(list):  # actually works(!!)
    def __class_getitem__(item_type):
        return List[item_type]
```

and then from other places:

```
from .polyfills import list, socket
```

This means that we can rely on unit tests working, for example, without a toplevel polyfill wrapper having run.  It also prevents us from mutating the global state.

martinpitt:
> This means that we can rely on unit tests working, for example, without a toplevel polyfill wrapper having run.

That's already the case, isn't it? The unit tests don't run bridge.py's main(). I specifically put the polyfills there for that very reason. That makes it a bit more awkward for askpass (as that has its own main), but I think that's worth it.

Importing "socket" from polyfills is bearable (albeit IMHO ugly -- that isn't a polyfill then, it's just a library) , there are only 6 imports. But this approach wouldn't work for e.g. `str.removesuffix()`. If we want to polyfill that, we'd have to modify the global str class.

We don't have to use a "one size fits it all" solution, though. This hybrid approach of avoiding removesuffix() and adding missing socket API seems fine to me -- and if you *don't* want polyfills (i.e. changing global state), and instead change the code to an explicit local library, then we don't even need to play dirty tricks, we can just give these proper names.

martinpitt:
That [first run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230210-075939-e2e439f3-centos-8-stream-pybridge/log.html#326-2) is shockingly successful.

 * `TestAccounts.testBasic` is an already [known flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18315-20230210-073706-1b1bf03d-fedora-37-pybridge/log.html#326-2) due to a leftover process from a deleted user. This isn't specific to RHEL 8 at all, and needs to be fixed in general.
 * `TestRoles.testDynamic` fails on `InterruptedError: [Errno 4] sd_bus_add_match: Interrupted system call`. Surprisingly, this reproduces locally, it seems to be very reliable. That's good, as usually `EINTR` is very racy.
 *

martinpitt:
I think the asyncio.Lock() fix in D-Bus is responsible for [these failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230210-093531-e2e439f3-fedora-37-pybridge/log.html#72):
```
asyncio-ERROR: Task exception was never retrieved
future: <Task finished name='Task-11408' coro=<DBusChannel.setup_path_watch.<locals>.handler() done, defined at /usr/lib/python3.11/site-packages/cockpit/channels/dbus.py:405> exception=BusError('org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.')>
  File "/usr/lib/python3.11/site-packages/cockpit/channels/dbus.py", line 411, in handler
    reply, = await self.bus.call_method_async(self.name, path,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 385, in call_method_async
    message = await self.call_async(message, timeout)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 379, in call_async
    return await pending.future
           ^^^^^^^^^^^^^^^^^^^^
systemd_ctypes.bus.BusError: org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.
```

I think these could have been previously hidden by running them on a wrong main loop? I think I want to split out this commit into a separate PR, to investigate it with less noise.

src/cockpit/channel.py
martinpitt: I tried that, but that's a bit difficult as the function does not exist at all in 3.6. Let's rather fix the systemd_ctypes regression and then bump it to latest main.

martinpitt:
There is a new problem, cockpit-askpass is broken again, and it's not due to shifting around the polyfills (reverting that still fails). Shell says "[Errno 104] Connection reset by peer",  and the journal does not say anything. Alternatively, shell says "or "recvmsg(stdin) failed: Resource temporarily unavailable", and journal
```
cockpit-ws[9670]:   File "/usr/local/lib64/python3.6/site-packages/cockpit/polyfills.py", line 40, in send_fds
cockpit-ws[9670]:     return sock.sendmsg(buffers, [(_socket.SOL_SOCKET, _socket.SCM_RIGHTS, array.array("i", fds))])
cockpit-ws[9670]: BrokenPipeError: [Errno 32] Broken pipe
```

I haven't yet figured out how to debug this, but I don't see anything obvious in `git diff  e2e439f..HEAD` which could have caused that. Even `git diff  e2e439f..HEAD src/cockpit|patch -Rp0` does not make it work again, and there is nothing else relevant in that diff. So either this is fallout from installing into /usr/local/, or a recent image refresh.

martinpitt:
Half of the failures of [this run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230216-065702-d382fce4-centos-8-stream-pybridge/log.html#69) are due to sd_bus_add_match() failing with EINTR. I added a commit to handle that, let's see how that works. If it does, we may want to move this to systemd_ctypes, but it's easier to test that way.

mvollmer:
@martinpitt , please remove "[storaged: Reimplement stratis2 pool rename hack with subscription" and reabse to main. That should fix it.

martinpitt:
@mvollmer says the stratis test now works with #18365, and this should also take care of the other d-bus errors. Let's try!

martinpitt:
The D-Bus [errors are still there](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230217-121320-2cbd5f4d-centos-8-stream-pybridge/log.html#135), I put back the message ignoring for now.

martinpitt:
Wow, #18361 worked wonders.. [Current run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230217-123049-e7ef14ef-centos-8-stream-pybridge/log.html#236-2) has exactly one failure (`TestStorageStratis.testEncrypted`) which @mvollmer is looking into, and otherwise *zero* flakes!

martinpitt:
@mvollmer mentioned that `TestStorageStratis.testEncrypted` is an SELinux issue. I added a `m.execute('setenforce 0')` to the start of the test, and that indeed gets a bit further: It fails in line 260 with "Error: The key description "pool0" is not currently set." (which is subtly different from "Error: Key with key description pool0 was not found", and that was already line 244).

Still, when the test fails there are *zero* AVC messages in the journal. I added `m.execute('semodule -DB')` to the start of the test, and that exposes a few issues:
```
avc:  denied  { siginh } for  pid=1778 comm="grep" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1780 comm="selinuxenabled" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1781 comm="grep" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1791 comm="ln" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1809 comm="cockpit-certifi" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0     
avc:  denied  { siginh } for  pid=1823 comm="sh" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:initrc_t:s0 tclass=process permissive=0     
avc:  denied  { noatsecure } for  pid=1834 comm="cockpit-ws" scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:system_r:cockpit_session_t:s0 tclass=process permissive=0     
avc:  denied  { rlimitinh } for  pid=1834 comm="cockpit-session" scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:system_r:cockpit_session_t:s0 tclass=process permissive=0     
avc:  denied  { read write } for  pid=756 comm="dbus-daemon" path="socket:[40508]" dev="sockfs" ino=40508 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=unix_stream_socket permissive=0     
avc:  denied  { noatsecure } for  pid=3164 comm="sshd" scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:chkpwd_t:s0-s0:c0.c1023 tclass=process permissive=0     
avc:  denied  { rlimitinh } for  pid=3164 comm="unix_chkpwd" scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:chkpwd_t:s0-s0:c0.c1023 tclass=process permissive=0     
avc:  denied  { siginh } for  pid=3164 comm="unix_chkpwd" scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:chkpwd_t:s0-s0:c0.c1023 tclass=process permissive=0     
```

Nothing here really stands out for stratis, though. I noticed that our python bridge in /usr/local has different SELinux users and contexts than the C bridge in /usr/bin, so I added this to the test:

    m.execute('chcon -u system_u -t shell_exec_t /usr/local/bin/cockpit-bridge')

but this does not help at all. SELinux normally does not care about the path of files, but I also tried

    m.execute('cp /usr/local/bin/cockpit-bridge /usr/bin/cockpit-bridge')

again no change.

So the best clue that I see so far is

    dbus.exceptions.DBusException: org.freedesktop.DBus.Error.Disconnected: Connection was disconnected before a reply was received

but I cannot match this with anything else in the journal, only to

    avc:  denied  { read write } for  pid=756 comm="dbus-daemon" path="socket:[40508]" dev="sockfs" ino=40508 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=unix_stream_socket permissive=0     

but I can't make sense of this yet. Let's confer with Marius next week.

martinpitt:
@mvollmer mentioned that `TestStorageStratis.testEncrypted` is an SELinux issue. I added a `m.execute('setenforce 0')` to the start of the test, and that indeed gets a bit further: It fails in line 260 with "Error: The key description "pool0" is not currently set." (which is subtly different from "Error: Key with key description pool0 was not found", and that was already line 244).

Still, when the test fails there are *zero* AVC messages in the journal. I added `m.execute('semodule -DB')` to the start of the test, and that exposes a few issues:
```
avc:  denied  { siginh } for  pid=1778 comm="grep" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1780 comm="selinuxenabled" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1781 comm="grep" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1791 comm="ln" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0    
avc:  denied  { siginh } for  pid=1809 comm="cockpit-certifi" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=process permissive=0     
avc:  denied  { siginh } for  pid=1823 comm="sh" scontext=system_u:system_r:init_t:s0 tcontext=system_u:system_r:initrc_t:s0 tclass=process permissive=0     
avc:  denied  { noatsecure } for  pid=1834 comm="cockpit-ws" scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:system_r:cockpit_session_t:s0 tclass=process permissive=0     
avc:  denied  { rlimitinh } for  pid=1834 comm="cockpit-session" scontext=system_u:system_r:cockpit_ws_t:s0 tcontext=system_u:system_r:cockpit_session_t:s0 tclass=process permissive=0     
avc:  denied  { read write } for  pid=756 comm="dbus-daemon" path="socket:[40508]" dev="sockfs" ino=40508 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=unix_stream_socket permissive=0     
avc:  denied  { noatsecure } for  pid=3164 comm="sshd" scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:chkpwd_t:s0-s0:c0.c1023 tclass=process permissive=0     
avc:  denied  { rlimitinh } for  pid=3164 comm="unix_chkpwd" scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:chkpwd_t:s0-s0:c0.c1023 tclass=process permissive=0     
avc:  denied  { siginh } for  pid=3164 comm="unix_chkpwd" scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:system_r:chkpwd_t:s0-s0:c0.c1023 tclass=process permissive=0     
```

Nothing here really stands out for stratis, though. I noticed that our python bridge in /usr/local has different SELinux users and contexts than the C bridge in /usr/bin, so I added this to the test:

    m.execute('chcon -u system_u -t shell_exec_t /usr/local/bin/cockpit-bridge')

but this does not help at all. SELinux normally does not care about the path of files, but I also tried

    m.execute('cp /usr/local/bin/cockpit-bridge /usr/bin/cockpit-bridge')

again no change.

So the best clue that I see so far is

    dbus.exceptions.DBusException: org.freedesktop.DBus.Error.Disconnected: Connection was disconnected before a reply was received

but I cannot match this with anything else in the journal, only to

    avc:  denied  { read write } for  pid=756 comm="dbus-daemon" path="socket:[40508]" dev="sockfs" ino=40508 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=unix_stream_socket permissive=0     

but I can't make sense of this yet. Let's confer with Marius next week.

martinpitt:
Trying to summarize what happens:
 * storage js calls CreatePool(), which succeeds.
 * storage js calls store_new_passphrase(), which runs stratis2-set-key.py, which calls `org.storage.stratis2.Manager.r1.SetKey()` and tries to pass fd 0 (stdin), where the JS writes the passphrase to.
 * The bridge strace contains `sendmsg(3, "…org.storage.stratis2.Manager.r1\0\3\1s\0\6\0\0\0SetKey…", so it's trying to do that call.
 * But that `SetKey` call is never actually seen in `busctl monitor`, so presumably that's the bit which SELinux rejects.
 * After delivering that SetKey call, stratis2-set-key.py only does one more `recvmsg(...) = 0` and then closes the D-Bus connection. After that it just reads some Python source files and prints "dbus.exceptions.DBusException: org.freedesktop.DBus.Error.Disconnected: Connection was disconnected before a reply was received". That suggests SELinux blocked the call.

My gut feeling is that stdin fd 0 is somehow labeled in an unexpected way (`unconfined_t`), so it cannot be passed to stratis. In the C bridge it's apparently labelled differently. But I already tried to relabel the cockpit-bridge binary, that didn't help.

martinpitt:
@allisonkarlitskaya I added a big hack for the last test case failure in stratis, but I'm afraid we need to talk about that. TL/DR: it would be better to do SubprocessTransport with a classic pipe pair rather than the socketpair(). There may be other situations where this leads to unexpected trouble.

martinpitt:
I dropped the stratis specific hack again, and instead made SubprocessTransport go back to pipes, with a flag to use a socketpair for Peers. I also sent that separately as PR #18380.

src/cockpit/askpass.py
```diff
@@ -66,7 +66,13 @@ def main() -> None:
         sys.exit('This command must be run with stdin connected to a socket.')
 
     interaction = AuthorizeInteraction(args.prompt)
-    asyncio.run(interaction.run(connection))
+    loop = asyncio.get_event_loop()
+
+    # shim for Python < 3.7
+    if not hasattr(asyncio, 'get_running_loop'):
+        asyncio.get_running_loop = lambda: loop
```
martinpitt, line 73: This still annoys me. I'll see to moving that to polyfills as well.

martinpitt: I tried that.. but it introduces a conflict with systemd_ctypes, is racy due to the global instantiation of an event loop, and does extra work. So I think it's less evil as it is here.

martinpitt:
@allisonkarlitskaya This is still blocked on #18380, but otherwise ready for the next round of review. Obviously the expected messages commit is a hack, but I'd rather like to add this to CI and not break it again and sort out these tasks messages separately (like in #18274)

martinpitt:
@allisonkarlitskaya , @mvollmer : Okay, removed the dependency on PR #18380, and replaced it with a test skip. This will still allow us to enable the scenario permanently, and stops us from breaking it again. PTAL?

src/cockpit/bridge.py
```diff
@@ -192,6 +194,8 @@ def setup_logging(debug: bool):
 
 
 def main() -> None:
+    polyfills.install()
```
martinpitt, line 197: Now that we know that {send,recv}_fds() are the only missing API that we still need, we actually could move them to bridge.py and transports.py, or call them as `polyfills.{send,recv}_fds()` Then we would not clutter global namespace. Let me know if you prefer that, I don't have a strong opinion either way.

src/cockpit/channels/dbus.py
```diff
@@ -282,7 +282,15 @@ def filter_owner(message):
             func = handler
         r_string = ','.join(f"{key}='{value}'" for key, value in r.items())
         if not self.is_closing():
-            self.matches.append(self.bus.add_match(r_string, func))
+            # this gets an EINTR very often especially on RHEL 8
```
allisonkarlitskaya, line 285: I wonder if we can do some root cause analysis here.  This might only be happening in a particular situation which we might avoid.  As far as I can tell, sd-bus never returns `EINTR` on its own, which means that this is most probably coming from some lower-level syscall returning `EINTR` which could then pop up as a return value in response to other calls....

An alternative might also be to try and deal with this at the binding level.  We already have a handler that inspects the return values of these functions and converts them into `OSError`.  Maybe we could redo the call in case we get EINTR?

allisonkarlitskaya: ps: relevant systemd commits look like b3d06b9226db96fddb6bb45a4708e2e8d413d91d and 3022916b4d2483452c3ddbbac9ee7c4372b1cb46.  It seems that they have an `ERRNO_IS_TRANSIENT()` macro to deal with fails like these and have been increasing its use recently.  I wonder if it makes sense to try and backport those patches into RHEL?

allisonkarlitskaya (approved):
With reservations, but approved.  Getting this under tests is important.

I wonder if you have a plan for SSH: we might have to skip a lot of tests there as well, once we start enabling them on other systems.  At that point, I imagine we'll introduce some sort of `@skipRHELpython` or so...

src/cockpit/askpass.py
```diff
@@ -66,7 +66,13 @@ def main() -> None:
         sys.exit('This command must be run with stdin connected to a socket.')
 
     interaction = AuthorizeInteraction(args.prompt)
-    asyncio.run(interaction.run(connection))
+    loop = asyncio.get_event_loop()
+
+    # shim for Python < 3.7
```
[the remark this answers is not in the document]
martinpitt: I thought about it, but I found it too heavy to pull in that entire library for just this one shim.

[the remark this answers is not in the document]
martinpitt: I [tried to move this to polyfill](https://github.com/cockpit-project/cockpit/pull/18318#pullrequestreview-1307181011), but it didn't end well. So in the end I found that the least ugly solution.

martinpitt:
> I wonder if you have a plan for SSH: we might have to skip a lot of tests there as well, once we start enabling them on other systems. At that point, I imagine we'll introduce some sort of @skipRHELpython or so...

My guess is that enabling ssh should un-fail the tests fairly consistently between Fedora and C8S, unless of course we *introduce* new 3.6 incompatibilities. But avoiding that is exactly the point of running the c8s/pybridge tests regularly. I am happy to move the `TestStratis.testEncrypted` skip into testlib to generalize it, and turn it from a skip into a `@todo`. I'm just not sure if it's worth doing that now, but indeed it would have been better to make testEncrypted a todo instead of a skip. I'll do that next, this is a good point. Thanks!

martinpitt:
> move the TestStratis.testEncrypted skip into testlib to generalize it, and turn it from a skip into a `@todo`.

I forgot: That test fails on "regular" C8S as well, due to a naughty. I'm not sure how well todo and naughty interact.. But that's why Marius and I eventually decided for "skip". That said, if we need this for other tests, I'm happy to add a `@todoPy36` or something such.

src/cockpit/channels/dbus.py
martinpitt: I actually ran into the EINTR with `journalctl` during my interactive tests. As it happens, this is already being backported in https://bugzilla.redhat.com/show_bug.cgi?id=2161683 . But this is completely independent from the sd-bus issue.

I reported this in https://bugzilla.redhat.com/show_bug.cgi?id=2172846 and asked for backporting this. If/once that happens, we can remove the workaround.

I looked into systemd_ctypes, and I didn't find a general wrapper around all bindings which would be able to run the function multiple times. There is `negative_errno()` (what you meant with "convert them into OSError"), but that does not have any context to be able to re-run the call. Possibly we could work this into `def instancemethod()`'s `wrapper()`? What would be the equivalent wrapper for `staticmethod`?

martinpitt: @allisonkarlitskaya : Forgot to tag you for my previous question :arrow_up:

PR #18320, opened 2023-02-10
Three small python fixes

martinpitt:
Split out from PR #18318, these should be the "harmless and obvious" ones.


mvollmer (approved):
Danke!

PR #18330, opened 2023-02-12
login: add autocomplete tags

marcusds:
Adding proper autocomplete to help password managers. Most importantly adding autocomplete="one-time-code" to stop browsers from thinking the 2fa field is the new password

https://bugzilla.redhat.com/show_bug.cgi?id=2102325

### DIFF
--- a/pkg/static/login.html
+++ b/pkg/static/login.html
@@ -99,13 +99,13 @@
 
         <div id="user-group" class="form-group">
           <label for="login-user-input" class="control-label" translate>User name</label>
-          <input type="text" class="form-control" id="login-user-input" autocorrect="off" autocapitalize="none" autofocus>
+          <input type="text" class="form-control" id="login-user-input" autocorrect="off" autocapitalize="none" autofocus autocomplete="username">
         </div>
 
         <div id="password-group" class="form-group">
           <label for="login-password-input" class="control-label" translate>Password</label>
           <div class="password-with-toggle">
-            <input type="password" class="form-control" id="login-password-input">
+            <input type="password" class="form-control" id="login-password-input" autocomplete="current-password">
             <button type="button" id="login-password-toggle" class="pf-c-button pf-m-control login-password-toggle" aria-label="Show password">
               <svg fill="currentColor" aria-hidden="true" viewBox="0 0 640 512">
                 <path class="password-show" d="M572.52 241.4C518.29 135.59 410.93 64 288 64S57.68 135.64 3.48 241.41a32.35 32.35 0 0 0 0 29.19C57.71 376.41 165.07 448 288 448s230.32-71.64 284.52-177.41a32.35 32.35 0 0 0 0-29.19zM288 400a144 144 0 1 1 144-144 143.93 143.93 0 0 1-144 144zm0-240a95.31 95.31 0 0 0-25.31 3.79 47.85 47.85 0 0 1-66.9 66.9A95.78 95.78 0 1 0 288 160z"/>
@@ -118,7 +118,7 @@
         <div id="conversation-group" class="form-group" hidden>
           <div id="conversation-message"></div>
           <label id="conversation-prompt" for="conversation-input"></label>
-          <input type="password" class="form-control" id="conversation-input">
+          <input type="password" class="form-control" id="conversation-input" autocomplete="one-time-code">
         </div>
 
         <div id="option-group">


martinpitt (approved):
Thank you! This matches https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete and https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields .

LGTM. @garrett , want to give this the final ack/nack?

marusak:
Similar to what I have done a while ago. See comments from @garrett

garrett:
Autocomplete is tricky.

Does this work in all browsers (Chrome, Edge, Firefox, WebKit) and multiple times in each browser? (Checking one time won't catch autocomplete issues, as you need to trigger the browser form saving and try again on a fresh reload.)

You _cannot_ rely on the specs. Browsers do not implement the specs, especially Chromium based ones, which have notoriously changed thing _several times_ over the past few years. And Edge does things differently from Chrome sometimes, when it comes to account/password autocomplete.

I'm trying to figure out how to test this all locally now (mostly the one time password part), currently waiting on the VM to build.

If others here could also please test it across the various browsers, that would be great.

garrett:
Oh, and you'll want to clear everything stored on the site when testing too (all cookies, site storage, password, etc.), followed by a shift-reload afterward. It's "easiest" by using a separate profile. Private browsing won't work, as it needs to remember all these things across sessions.

garrett (changes requested):
- Chrome: Doesn't work. (Remembers user, but not password.)
- Edge: Doesn't work. (Remembers user, but not password.)
- Firefox: Doesn't work at all. (Does not remember user or password.)
- GNOME Web (WebKit, like Safari): Doesn't work at all.  (Does not remember user or password.)

I only tested username and password, not OTP. But it doesn't work for user/pass across browsers, with the exception of Chrome & Edge remembering user only.

martinpitt:
Hmm, so back to `autocomplete="no"` for everything, as originally discussed in https://bugzilla.redhat.com/show_bug.cgi?id=2102325 ?

garrett:
> Hmm, so back to autocomplete="no" for everything

No. Setting autocomplete to no, sets it _on_, to a custom label called "no".

Note: Setting it to "off" is often ignored.

(Autocomplete is tricky, broken, and inconsistent across browsers. All thanks to the Google Chrome team mucking with it and breaking it further every few months.)

martinpitt:
@garrett : Sorry, I meant `autocomplete="off"` , that's what the bz was asking for. That's also what https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete documents (and even recommends for password managers)

garrett:
> and even recommends for password managers

Where does it say that?

All I see is:

> Note: In most modern browsers, setting autocomplete to "off" will not prevent a password manager from asking the user if they would like to save username and password information, or from automatically filling in those values in a site's login form.

Which is the opposite of what you're saying. It says the it won't prevent a password manager from handling it, not that it will hand it off to a password manager or that it is recommended.

martinpitt:
Well, right. I meant that apparently `autocomplete="off"` will *not* stop password managers from being able to fill out login forms. That's what we want, right?

garrett:
So this PR is _supposed to be_ doing the right thing.

It doesn't look like it from my testing (as browsers have probably changed it since the recommendation, according to things I've read). But I don't want to stall this further, if it is the right thing... could others please check to see what it does on all the various supported browsers? (**Firefox**, **Chrome**, **WebKit** (Epiphany / GNOME Web, or Safari if you have it), and **Edge**... which is basically Chrome at heart but does different stuff sometimes, like with password management).

Right now, I do have autocomplete in Firefox and Chrome, but don't with this PR. Which is why I'm concerned about it. But it is theoretically correct, even if it doesn't act like it here for me. Hence asking for verification by others on what's going on with their systems and browsers.

marcusds:
Sorry for the delayed response.

I was able to test in Chrome and Firefox and both functioned as I expected them to. Both saved the username and password and didn't try to save the one time code as the password. Firefox did offer to use the saved password in the one time code field, but that to me is acceptable since entering the otc didn't cause any issues.

marusak:
I checked this and it works on firefox and chrome. I tried with my current profile and also with blank profile. Firefox does not prompt but when you click on the key icon you can save the data and it is offered then for you. In Chrome it also pops up. In both cases it is the same behavior as on current main. 

So from my point of view this is good to merge. @garrett want to add something or can I land this?

garrett (approved):
As it's correct to the spec and it's working for you all, I supposed that means it's good to merge.

It's curious that I couldn't get it working with this PR in any browser, but as-is on main it works for me. :disappointed: (I hope this doesn't cause problems for anyone else.)

martinpitt:
Thanks @marcusds , @garrett , and @marusak for the extensive testing! Let's put this out into the wild and collect feedback then.

PR #18332, opened 2023-02-13
metrics: pass INTERVAL value when creating the metrics1 channel

KKoukiou:


martinpitt (approved):
Heh, good catch -- that means that we previously got "unaligned" data, as INTERVAL is 5s, while that was 3s. Thanks!

PR #18336, opened 2023-02-13
tls: Make test failures more useful

martinpitt:
test-tls-server often fails with

    recv_reply: assertion failed (len >= 100): (-1 >= 100)

which is not particularly enlightening. It could be something simple like EINTR, or the timeout loop actually running all the way to the end without ever getting an actual response. Make the failure of both of them more explicit and verbose, and also print errors.

---

I saw [this failure](https://download.eng.bos.redhat.com/brewroot/work/tasks/635/50720635/build.log) several times downstream by now. Let's figure out what this is!


allisonkarlitskaya (approved):
sure

PR #18339, opened 2023-02-14
tools: Bring back missing files check in Debian packaging

martinpitt:
`dh_install --fail-missing` has been deprecated for a while and got removed in dh compat level 12 (which we use). Use `dh_missing` instead. We can drop the override once we move to dh compat level 13. This fixes the warning during build:

    dh_install: warning: Please use dh_missing --list-missing/--fail-missing instead
    dh_install: warning: This feature will be removed in compat 12.

### DIFF
--- a/tools/debian/rules
+++ b/tools/debian/rules
@@ -43,3 +43,7 @@ override_dh_install:
 
 	dh_install -Xusr/src/debug
 	make install-tests DESTDIR=debian/cockpit-tests
+
+# will be the default in compat 13
+override_dh_missing:
+	dh_missing --fail-missing


tools/debian/rules
```diff
@@ -43,3 +43,7 @@ override_dh_install:
 
 	dh_install -Xusr/src/debug
 	make install-tests DESTDIR=debian/cockpit-tests
+
+# will be the default in compat 13
```
mvollmer, line 47: We already build-depend on `debhelper-compat (= 13)`, no? So aren't we already using compat level 13?

martinpitt: Argh, you are right. I fished this out from #18112, but indeed it's not needed any more.

mvollmer: Nice!

PR #18341, opened 2023-02-14
test: Factorize OS specific test skips

martinpitt:
Add globbing support in testlib.skipImage() and add testlib.skipOstree(). This makes it easier in the future to add/remove OSes.


jelly (approved):
I wonder why we never did this before, makes a lot of sense :+1:

PR #18342, opened 2023-02-14
Metrics zoom in/out

KKoukiou:


garrett:
How I feel about the current zoom level and different zoom level(s):

1. The current scale is definitely not optimal.
2. I've long thought that a better scale would be daily, with hours inside.
3. I'm not sure if controlling scaling up or down would be useful. (I'm not convinced yet.)

I'm including a redesign I've been working on in the past couple days based on what @jelly and I were talking about a while back. It's not a total redesign.

Meanwhile, I adjusted the colors, changed sizing and spacing, fixed alignment, and reworked the log view a bit.

The part that's most relevant to this PR:

- _If_ we do have a way to adjust the size, it could be something like a combined widget where it has zoom controls on the sides (with + and - icons specifically _not_ in a circle) and the current zoom in the middle. It would have to be fixed width to accommodate all views.

Additional notes:

- The log view now has icons for (relative) severity, versus the rest of the day. We probably want to have a cutoff point though, so we don't have red full events on otherwise "basically nothing" days.
- Spikes are counted separately as combined events, just showing the number. The types of spikes shows up in another column.
- Shutdown and Boot are now included. We'd probably want other major events as well. But these are rather major events that should show up here. They're especially useful on systems with a watchdog that does automatic shutdowns.

_**Important to note for this PR**_: If we do use zooming, having aggregated events and icons for relative severity (mentioned above, and in the mockup below) lets us group the events in a meaningful way, regardless of any zoom level.

Metrics page(4)

Please note: I don't have lines for segments (as that would be hard to add in here, with the way the mockup is made). I'm thinking the lines are a little bit too strong as-is and we want to soften them up (but not get rid of them completely). So I'm not removing the lines; they're just not in the mockup.

briansmith0:
One thing to keep in mind is that when you are zoomed out, you lose a lot of the detail, and things like a short 1 minute CPU spike might be smoothed out by the consolidation of something like 5 minutes of time in to a single pixel on the graph.  So in my opinion, it is important to be able to zoom out and see a multi-day view in order to easily track longer term trends (such as a slow memory leak that increases memory usage over several days), but then also be able to zoom in to specific times to see more details so that you can more easily identify these shorter term resource spikes (such as the CPU spiking to 100% for 1 minute every hour).

martinpitt:
I just discussed that with @KKoukiou in a meeting, and I proposed the same: when zooming out,  don't disregard events, but show a summary count of them. Yes, that stops the possibility of clicking on them to get a journal, but IMHO it makes more sense: When zooming out, the admin wants a summary view of resource usage, but not completely ignore relevant information like "less severe" spikes.

My main concern is with the computing overhead. At some point it becomes unbearable to do computations of millions of data points (pcp collects every 5 s or so?) on the browser/JS side, this has to happen on the bridge. I am currently not sure what *exactly* the `interval` option on the PCP channel does -- if it merely gives you data points at these intervals and disregards the ones in between, the option is useless. It should compute the average or perhaps maximum of *all* data points in the given interval and report a summary number. Perhaps it already does that, need to check. But if not, we need to teach it that first. Then the JS code can turn `INTERVAL` from a constant to a variable and ask the bridge to give it more coarse-grained information. Of course that approach means that we'd also have to move the event computation into the bridge.

`</brain dump>`, we probably need to discuss that in more detail.

marusak:
I talked about this to many people and here are some of my thoughts.

1. Default view should be "daily". Meaning one block is one day and on normal laptop screen you would fit around one day at a time. Pretty much what @garrett has in the [mockup](https://user-images.githubusercontent.com/10246/218791612-3f3d7472-d06b-40e4-9db0-ffba8b519de5.png)
2. I don't think we really need explicit zoom in/out. It can be dropdown where you select "Day view"/"Hour view" but maybe not even that just being able to go deeper from default view (which should be "daily").
3. Each day block should be split into 24 lines and these lines should be expandable(or switch to "hour view") to get view of one hour.
4. Expanded hour should be split to 60 lines. Pretty much what we have right now. This should contain events as it does right now. (is it good practice to expand hour and then expand minute in it?)
5. <maybe?> Full day view does not contain aggregated listing of events. This is both difficult to calculate, is likely to cause panic and is questionable if it provides valuable information. Point of having the full day view is to see how the server is doing over longer period of time. Is it loaded? Is it using too much swap?... This view is not about answering questions like What has happened at 12:34? Also having aggregated numbers of events can be dangerous as we say "28 spikes at 12", "18 spikes at 13", "29 spikes at 14"... for each and every hour. All hours would be red and it would look like there is so many problems happening while that can be normal behavior for the server. If then someone wants to see why the server started to be loaded after 12:20 they can expand the view and see what has been happening around that time. We also arbitrarily guess what is "spike" without having context so not showing them by default also seems like a little win and users can focus on what they are interested in and not what cockpit thinks is interesting.

garrett:
> I don't think we really need explicit zoom in/out. 

Yep, I agree. We do not need to redefine the view.

---

So this still needs a lot of work, but I'm sharing something that is not intended to be implemented... it's a heavy WIP and will definitely change.

With that said, I'm sharing this in-between mockup to help get the feel with hourly time per day and how high it might be. Right now, I do have shutdowns and boots in-between lines. They might just show up when hours are expanded. I'm still wondering what metadata we show at this level and how it should be organized.

Again: This is mainly for a feel about how tall it might be with 24 hours in a day like this. And _possibly_ with some additional rows. (But we might just make it 24 only? Not sure.)

And I made it so the red indicator is gone. There are just two states: standard and above threshold (but still 4 PF icons for very low, medium, high, very high).

Metrics page(7)

briansmith0:
It would be very helpful to have a view that shows a longer time period than one day.  For example, a weekly view in order to see things like slowly memory leaks (for example, where memory usage goes up 3% a day consistently, it is harder to spot this from a daily view).   Or with a weekly view you can see if there are resource spikes at certain times of the week (for example on Mondays at 8:00 AM there is a spike but you don't see that Tuesday - Friday).   

This is functionality that was previously available cockpit-pcp.  If you use cockpit-pcp-195 in RHEL 7, you have a drop down to select the time period for the graph:

- Go to now
- 5 minutes
- 1 hour
- 6 hours
- 1 day
- 1 week

From any of these graphs, I can highlight a time period on the graph and zoom in to that portion to see more detail.

garrett:
I did a quick, hacky in-dev-tools CSS mod to make it look like a popover with a selected item, to give the idea of how a popover could look. It should have an X button and both that and clicking on the origin line would dismiss the popover. (Perhaps clicking anywhere outside the popover would too.)

image

A popover wouldn't limit us to the size of the logs or the graphs and could be made to work better on mobile as well (especially with the X button).

I made the logs wrap with overflow-wrap, to ensure wacky non-breaking lines would be broken to wrap when needed.

The popover could have a title and even a filter bar to show different levels of logs and such and it could scroll if there are too many lines of logs to fit within.

(Most of these attributes are true of drawers too... but some of these things are advantages of popovers. Conceptually speaking, drawers are _basically_ kind of like anchored popovers.)

garrett:
Trying to re-wrap my head around this page, with all your changes too. I'm also trying to simplify layout a bit, where possible. I managed to change a bulk (not all yet) of the metrics from grid to flex; I think it helps with performance and it looks the same here.

Anything we can do with respect to layout performance should help when developing (and will eventually help anyone using the page, naturally), hence me spending some time poking at the page like this.

(I'll continue on with some other improvements, including visual, of course.)

pkg/metrics/metrics.scss
```diff
@@ -523,32 +523,26 @@ html {
 
     // compressed minutes without events are simple bars
     .compressed {
-      display: grid;
-      grid-template-areas: ". utilization .";
-
-      --util-pct: calc(100% * var(--utilization));
-      grid-template-columns: calc(100% - var(--util-pct)) var(--util-pct);
+      --graph-parts: 1;
+      --util-pct: calc(100% / var(--graph-parts) * var(--utilization));
```
KKoukiou, line 533: That's not correct. No the --saturation value is ignored. Please leave it as is, I had a whole commit on top that adds the red part of the graph and it's conflicty. I want to solve this first.

garrett: OK, we can revert this commit. It's probably worth looking into it though, de-grid-ifying where we can.

garrett:
My thoughts, suggested on IRC on how to proceed for mobile:

- Compress the graphs to be narrow
- Drop the width association for the labels, so that they're more of a "legend" with color dots and can be whatever width they need to be
- Either have the legend stacked vertically or have it horizontal and wrapping (let's see what works)

Since the graphs use 2 colors for the usage/load usage/swap, we could have dual indicators:

[color1][color2] CPU usage and load
[color3][color4] Memory usage and swap
[color5] Disk usage
[color6] Network usage

Now, if they're vertical, it we'd probably want to either right-align the indicator colors and left-align the text or "just" double the width of the colors for disk and network so everything aligns.

This is all to avoid having 6 labels (CPU usage, CPU load, Memory usage, Swap, ... and colors for each) for the legend instead of 4. Every little bit of space we can save somewhere in mobile will make it all work better.

If we have it horizontal, then we'd have it wrap. We'd want to make sure the color and text does not wrap independently, but together. This could be done with either inline-block or flex with flex-wrap. (It could be grid too, to make them align, but I don't think we want or need columns to align here.)

pkg/metrics/metrics.jsx
```diff
@@ -975,66 +979,47 @@ class MetricsMinute extends React.Component {
...
+                <span className="spikes_count" />
+                <Popover position="right" hasAutoWidth className="metrics-events-popover" bodyContent={logsPanel}>
+                    <Button
+                        variant="link" isInline
+                        className="spikes_info"
+                        onClick={() => this.expand(!this.state.expanded)}>
+                        <span className="type">
+                            { this.props.events.events.map(t => RESOURCES[t].event_description).join(", ") }
+                        </span>
+                    </Button>
```
martinpitt: I'm not sure how exactly this is buggy, but if there is more than one spike type in a second, then it's not a link:

image

I.e. I can click on the first "Disk I/O" and "CPU"  to get the journal popup for 7:02 and 7:08. But the "CPU, Swap, Network I/O" text is not a link, and not clickable to get the 7:07 journal.

This is repeatable: If I cause two spikes in the same minute, vs. causing a single spike.

```
nc -n -vv -l 2000 > /dev/null & sleep 1; nc -vv localhost 2000 </dev/zero
echo 3 > /proc/sys/vm/drop_caches; cat /dev/vda > /dev/null
dd if=/dev/urandom of=/dev/null
```

martinpitt (commented):
I don't want to bikeshed about the design  - this was already discussed between Matej, Brian, and Garrett. I tested this interactively for a while, and I like it! @garrett , you did a really classy job with making this work on mobile!

I only did a relatively coarse code review, but nothing  jumped out on me. I remember you wanted to change the polygons into boxes, but I suppose that's for the other PR with the "over-use color change"? (And I agree, it's a different topic)

pkg/metrics/metrics.jsx
martinpitt: Hmm, this feels a bit funny. After letting it run for a bit longer, my newer "not a link" suddenly turned out to be a link. I now saw the extra commit to not make it a link if there are no log messages. `journalctl -b | grep 06:07` indeed shows no logs, and the previously missing link for 07:15 now exists:

image

So this is just some journal loading lag, unrelated to this PR and most probably just a bit wonky for the current minute.

martinpitt:
@KKoukiou So TL/DR, no objections from me, great work! I figure you'll need to wrestle with the tests now.

marusak:
Could we load more than 24 hours on the initial load? Since default view is per day, loading just one day seems a bit too little. Maybe two?
24hours

The page seems a bit "messy"? Likely unrelated to this PR but the top part seems a bit unclean. @garrett can you review this PR and make sure with @KKoukiou that this PR does not regress any of it? Maybe some polish can be done in another PR

martinpitt:
@marusak It's a compromise -- loading takes quite a long time, and it blocks the initial page load. Try this with a remote server, it's quite considerable. I'd almost prefer the current behaviour of lazy loading on demand, but not a strong opinion.

garrett:
We should do everything we can to get to initial paint and then lazy load the rest incrementally.

I agree that it doesn't load enough (and I think I said that either in this PR somewhere above and/or IRC).

Also, I have been jumping between so many things, I haven't had the time I wanted to spend on this PR. I did spend a good bit already, but I meant to spend more for polishing. 🫤 I should jump back in today.

garrett:
I made a bunch of minor adjustments, mainly revolving around alignment, to try to "clean up" the feeling a bit.

garrett:
Also, I still think the hour is wrong; it should be the ending time and everything preceding it when expanded. It's weird to expand 1:00 and see times leading up to 2:00, right?

image

martinpitt:
> It's weird to expand 1:00 and see times leading up to 2:00, right?

Just my 2c: It's a bit confusing as it's expanding down, not up. But I'd still expect an event at 13:37 to be in the "13:00" line.

KKoukiou:
> > It's weird to expand 1:00 and see times leading up to 2:00, right?
> 
> Just my 2c: It's a bit confusing as it's expanding down, not up. But I'd still expect an event at 13:37 to be in the "13:00" line.

This is one thing, expecting the 13:37 to be at 13:00 line. one more problematic part is that when we want to show expanded logs from 23:00 to 23:59 if we show the 00:00 as header line, it's misleading as it belongs to next day and we group section per day.

martinpitt (changes requested):
Thanks! Code-wise this is fine for me, but this needs some better test coverage: expanding/collapsing an hour (i.e. the central part of this PR) isn't really covered. events_at() expands the hour, but it should at least check that events are not visible before expanding, and that it is collapsed by default. It should also test the new event counts, like "1 CPU, 2 Memory"

test/verify/check-metrics
```diff
@@ -426,16 +429,8 @@ class TestHistoryMetrics(MachineCase):
...
-        b.click("#metrics-hour-1615197600000 div.metrics-minute[data-minute=39] .metrics-events button")
-        b.wait_in_text("#metrics-hour-1615197600000 div.metrics-minute[data-minute=39] .cockpit-log-panel", "No logs found")
-        b.click("button:contains('View detailed logs')")
-        b.enter_page("/system/logs")
-        # jumps to the right time with larger window and debug level
-        url = b.eval_js('window.location.hash')
-        self.assertIn("priority=debug", url)
-        self.assertIn("since=2021-3-8%2010%3A24%3A0", url)
-        self.assertIn("until=2021-3-8%2010%3A40%3A0", url)
-        b.go("/metrics")
```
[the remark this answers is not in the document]
KKoukiou: The logs fetching is done automatically now for the expanded hours, and if there are no logs, there is not button. So all this is be now replaced with:

b.wait_in_text("#metrics-hour-1615197600000 div.metrics-minute[data-minute=39] .metrics-events span.spikes_info", "Load")
 
which is just the spike description span without being a button.
I could add a:

b.wait_not_present("#metrics-hour-1615197600000 div.metrics-minute[data-minute=39] .metrics-events button")

to make this button absense more explicit.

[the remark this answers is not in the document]
KKoukiou: @martinpitt I added the requested test coverage. Let me know if you still are worrying about the missing link to logs page, in case no logs are detected.

martinpitt (approved):
Excellent, thank you!

martinpitt:
@KKoukiou Added release-note label. Can you please prepare (or re-use) a representative screenshot and add it to the description?

KKoukiou:
There is some small regression - pixel tests cough it. Sending up a fix.

PR #18344, opened 2023-02-14
metrics: fix error message about missing data from PCP

KKoukiou:
The missing data are between the oldest_timestamp and the oldest item in the 'hours' list, which is the end of the list not the start.

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -1640,14 +1640,15 @@ class MetricsHistory extends React.Component {
             />;
 
         let nodata_alert = null;
-        if (!this.state.loading && this.state.hours.length > 0 && this.oldest_timestamp < this.state.hours[this.state.hours.length - 1]) {
+        const lastHourIndex = this.state.hours.length - 1;
+        if (!this.state.loading && this.state.hours.length > 0 && this.oldest_timestamp < this.state.hours[lastHourIndex]) {
             let t1, t2;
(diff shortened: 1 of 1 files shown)


martinpitt:
I stared at this for a few mins, but I'm afraid I don't understand this logic any more, I need to read more of the code to remember all the data structures. Review tomorrow?

KKoukiou:
> I stared at this for a few mins, but I'm afraid I don't understand this logic any more, I need to read more of the code to remember all the data structures. Review tomorrow?

Yeap not urgent at all. I am also happy to explain in call / irc.

martinpitt:
I've been staring at this for 20 minutes now, and I'm afraid I still can't make sense of it. `this.state.hours` is sorted in descending order, i.e. hours[0] is the earliest one which we have data available for, and `hours[-1]` the latest one. But when determining the "empty gap" between the oldest timestamp, this goes up to the *earliest* time for which we have available data. Moving to hours[0] would claim that there is no available data for the times covered by all the hours[..] indexes.

What frustrates me (not towards you, but towards me, as I wrote this stuff) is that we don't seem to cover this with an integration test. A change like this ought to cause a test failure, or need a test adjustment to demonstrate the effect. We only check for "No data available between" but without concrete values, presumably because this is very hard to keep reproducible?

martinpitt:
> hours[0] is the earliest one which we have data available for,

Argh, no. Speak after me: **descending** order. So this is correct, sorry for the confusion!

martinpitt (approved):

PR #18345, opened 2023-02-14
packit: add Fedora 38 testing and release, drop Fedora 36

jelly:


martinpitt (approved):
Awesome that works with no further changes.

PR #18350, opened 2023-02-15
Cockpit verify manifests

jelly:
@allisonkarlitskaya WDYT

### DIFF
--- a/src/cockpit/packages.py
+++ b/src/cockpit/packages.py
@@ -25,9 +25,10 @@ import mimetypes
 import os
 import re
 import shutil
+import typing
 
 from pathlib import Path
-from typing import ClassVar, Dict, List, Optional, Pattern, Tuple
+from typing import ClassVar, Dict, List, Optional, Pattern, Tuple, Union
 
 from systemd_ctypes import bus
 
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (commented):
In general I think I'd prefer something that does a little bit less "work" and looks a bit more like using `TypedDict`, possibly containing `Sequence`s of other `TypedDicts` (ie: for the bridges, which themselves, are complex types).

`TypedDict` in its subclass form (with type hints) isn't in 3.6, but we could just drop the validation if the old version doesn't support it.  I don't consider this to be a critical feature.

I'd also like to see about trying to pair this with some kind of a wrapper that ensures deep immutability of the data.  We store references to this data all over the place, and if someone thought that they are safe to append an item to their "copy" of the data, it could have extremely wide-ranging effects that are difficult to track down.  This lack of support for complex value types is sort of a weak spot in Python, in my opinion, and I wonder if we could find a way to deal with this issue at the same time.

I also had the idea that even if we don't deal with the read-only thing using actual runtime checks, we could at least get some kind of hinting around it by forcing the use of some kind of a readonly mapping type, `Sequence` instead of `list`, etc.  That would mean that at runtime everything is actually still a dict, but mypy would tell us if we're doing something wrong.  I'm not sure if it's possible to have this at the same time as `TypedDict` though, since there's no way (as far as I know) to denote a readonly typed mapping.

jelly:
This is better implemented in https://github.com/cockpit-project/cockpit/pull/19152

PR #18351, opened 2023-02-15
Delete swp file from git and gitignore it in the future

skobyda:
It seems pkg/users/.account-details.js.swp accidentally ended up in our git. Let's get rid of it.
Also add '*.swp' pattern into .gitignore


martinpitt (approved):
Whoops, thanks!

martinpitt:
No need to retry forever, this is a simple change.

PR #18352, opened 2023-02-15
test: Drop obsolete sssd hack

martinpitt:
The bug got fixed upstream in 2.8.0 which is in Debian testing, and the fix got backported to Ubuntu 22.04.


jelly (approved):
The bots agree!

PR #18321, opened 2023-02-10
python: Defer asyncio.Lock() creation for D-Bus

martinpitt:
Initializing a global lock at program start is too early: At that point, systemd_ctypes has not yet initialized its own main loop. In Python 3.6 this causes tons of crashes like

    RuntimeError: Task <Task pending coro=<DBusChannel.do_open.<locals>.get_ready() running at
    /usr/lib64/python3.6/site-packages/cockpit/channels/dbus.py:251>
    cb=[set.discard()]> got Future <Future pending> attached to a different loop

This also moves the lock from a class to an instance variable -- it's not necessary to globally order messages, doing it for a single channel is sufficient.

----

I split this out from PR #18318 , I think this causes some fallout, e.g. [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230210-093531-e2e439f3-fedora-37-pybridge/log.html#72) and [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230210-075939-e2e439f3-centos-8-stream-pybridge/log.html#333). 

These look like this:
```
asyncio-ERROR: Task exception was never retrieved
future: <Task finished name='Task-11408' coro=<DBusChannel.setup_path_watch.<locals>.handler() done, defined at /usr/lib/python3.11/site-packages/cockpit/channels/dbus.py:405> exception=BusError('org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.')>
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/cockpit/channels/dbus.py", line 411, in handler
    reply, = await self.bus.call_method_async(self.name, path,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 385, in call_method_async
    message = await self.call_async(message, timeout)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 379, in call_async
    return await pending.future
           ^^^^^^^^^^^^^^^^^^^^
systemd_ctypes.bus.BusError: org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.
```

or this:

```
[...]

PR #18356, opened 2023-02-16
python: close all tasks before exiting the bridge

allisonkarlitskaya:
Let's split this out from #18292 (which looks to have gained a dependency on #18340, which itself, needs work).

Let's make sure we wind down all tasks when closing the bridge.  This is implemented in two steps, basically:

 - no channel may send "close" until all of its tasks are complete
 - the bridge may not exit until all channels have sent "close"

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -18,7 +18,7 @@
 import asyncio
 import logging
 
-from typing import ClassVar, Dict, Generator, List, Optional, Sequence, Tuple, Type
+from typing import ClassVar, Dict, Generator, List, Optional, Sequence, Set, Tuple, Type
 
 from .router import Endpoint, Router, RoutingRule
 
@@ -90,6 +90,10 @@ class Channel(Endpoint):
     _out_sequence: int = 0
     _out_window: int = SEND_WINDOW
 
+    # Task management
+    _tasks: Set[asyncio.Task]
+    _close_args: Optional[Dict[str, object]] = None
+
     # Must be filled in by the channel implementation
     payload: ClassVar[str]
     restrictions: ClassVar[Sequence[Tuple[str, object]]] = ()
@@ -102,6 +106,7 @@ class Channel(Endpoint):
         # things that our subclass may be interested in handling.  We drop the
         # 'message' field for handlers that don't need it.
         if command == 'open':
+            self._tasks = set()
             self.channel = message['channel']
             if message.get('flow-control'):
                 self._send_pings = True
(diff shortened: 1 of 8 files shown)


src/cockpit/channel.py
```diff
@@ -167,8 +180,48 @@ def ready(self, **kwargs):
...
+        return self._close_args is not None
+
+    def _close_now(self):
+        self.send_control('close', **self._close_args)
+
+    def _task_done(self, task):
+        # Strictly speaking, we should read the result and check for exceptions but:
+        #   - exceptions bubbling out of the task are programming errors
+        #   - the only thing we'd do with it anyway, is to show it
+        #   - Python already does that with its "Task exception was never retrieved" messages
```
martinpitt, line 194: [This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18356-20230216-094601-9474f71d-fedora-37-pybridge/log.html#72) is an interesting case study for this comment/claim:
```
asyncio-ERROR: Task exception was never retrieved
systemd_ctypes.bus.BusError: org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.
```
This actually kind of *is* an error which should be caught and bubbled up through the protocol as `access-denied`.

This failure is *not* new -- since we landed #18321 we've been getting this on some test runs. It's not a regression really, #18321 just made these errors actually visible. In fact, in a run there we got the exact same test failure/message.

So TL/DR: I think we can ignore/retry the test run for this particular case, but adding this general claim as a comment here seems wrong?

mvollmer (approved):

src/cockpit/channel.py
martinpitt: In other words, can we actually collect the task error, close the channel with an appropriate problem code for `PermissionDeniedError`, D-Bus errors and similar run-time/user errors (what are these, `OSError`?), and log the crash otherwise?

martinpitt: That said, as (1) this isn't new, and (2) we have a constant reminder that we need to clean this up through failed tests, I don't veto this. Everything else looks good to me, and I'm happy to see this landing to untangle this web.

martinpitt: However, let's please retry and only land if green. It's very plausible that the effect of this PR is to turn this failure from "flaky" to "100% reliable", as we now wait for tasks to close. Otherwise, if it repeats, we at least have a reliable way for reproducing this scenario, which is hugely helpful as well.

allisonkarlitskaya: I consider the claim in this comment to be 100% correct to this case.  Specifically: exceptions bubbling out of the task are programming errors.  This particular failure is a programming error and we need to fix it.  The fact that Python has shown this to us helps us fix this error.

This is exactly what I had in mind when I wrote that comment.

allisonkarlitskaya: Bug for this → https://github.com/cockpit-project/cockpit/issues/18358

PR #18359, opened 2023-02-16
python: clean up subprocess signal handling, shutdown

allisonkarlitskaya:
This is my attempt at half of #18340, plus some bonus cleanups.

### DIFF
--- a/src/cockpit/channel.py
+++ b/src/cockpit/channel.py
@@ -336,6 +336,10 @@ class ProtocolChannel(Channel, asyncio.Protocol):
         if self._transport.can_write_eof():
             self._transport.write_eof()
 
+    def do_close(self) -> None:
+        if self._transport is not None:
+            self._transport.close()
+
     def data_received(self, data: bytes) -> None:
         assert self._transport is not None
         if not self.send_data(data):
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -107,19 +107,6 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
         if window is not None:
             self._transport.set_window_size(**window)
 
-    def do_close(self):
-        assert isinstance(self._transport, SubprocessTransport)
-        pid = self._transport.get_pid()
-        # avoid calling .terminate(), as that will try to read the process' exit code and race with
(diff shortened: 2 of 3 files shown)


martinpitt:
Ugh, a lot of failures:
```
os.kill(self._process.pid, sig)
^^^^^^^^^^^^^
AttributeError: 'SubprocessTransport' object has no attribute '_process'
```

Aside from that, this is really nice cleanup! Looking forward to bringing back the unit tests after this!

allisonkarlitskaya:
The test failures are really obvious: if the subprocess spawn throws then we end up with a half-initialized object that still gets `__del__()` called on it.  We need to deal with that by detecting the half-initialization.

It's a little bit annoying and I'm not clear what the "correct" way to deal with this is in terms of mypy.  It's happy if you unconditionally assign the values in `__init__()` but it doesn't consider exceptions.

martinpitt:
Thanks! So, this is the time and place where we can cheer for TestActivePages to go green, err todo-red? :tada:

martinpitt:
Hmm, sad, [TestActivePages still fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18359-20230216-133732-12b8e86c-fedora-37-pybridge/log.html)

allisonkarlitskaya:
> Hmm, sad, [TestActivePages still fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18359-20230216-133732-12b8e86c-fedora-37-pybridge/log.html)

This PR isn't trying to fix that.  See #18292.

allisonkarlitskaya:
I bet it's because we `SIGKILL` now rather than `SIGTERM`.  Probably something else would clean up those processes before, but it no longer gets the chance to.... hrmph.

allisonkarlitskaya:
> I bet it's because we `SIGKILL` now rather than `SIGTERM`. Probably something else would clean up those processes before, but it no longer gets the chance to.... hrmph.

Yup.

allisonkarlitskaya:
> I bet it's because we `SIGKILL` now rather than `SIGTERM`. Probably something else would clean up those processes before, but it no longer gets the chance to.... hrmph.

@mvollmer says "let's do SIGTERM", and I'm inclined to agree.

martinpitt (approved):
Unchanged opinion: This is still a great PR and nice cleanup! Crossing fingers for the last few meters of the tests.. :checkered_flag:

PR #18365, opened 2023-02-17
python: catch BusError on Properties.Get

### LINKED ISSUE #18358: python: dbus channel doesn't catch errors when fetching invalidated properties
### Explain what happens

See this trace: https://cockpit-logs.us-east-1.linodeobjects.com/pull-18356-20230216-094601-9474f71d-fedora-37-pybridge/log.html#72

Specifically:

```
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 systemd[1]: SELinux getfilecon_raw() on '/usr/lib/systemd/system/NetworkManager.service' failed (perm=status): No such file or directory
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 systemd[1]: Reloading.
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]: asyncio-ERROR: Task exception was never retrieved
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]: future: <Task finished name='Task-10962' coro=<DBusChannel.setup_path_watch.<locals>.handler() done, defined at /usr/local/lib/python3.11/site-packages/cockpit/channels/dbus.py:406> exception=BusError('org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.')>
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]: Traceback (most recent call last):
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:   File "/usr/local/lib/python3.11/site-packages/cockpit/channels/dbus.py", line 412, in handler
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:     reply, = await self.bus.call_method_async(self.name, path,
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:   File "/usr/local/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 385, in call_method_async
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:     message = await self.call_async(message, timeout)
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:   File "/usr/local/lib/python3.11/site-packages/systemd_ctypes/bus.py", line 379, in call_async
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:     return await pending.future
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]:            ^^^^^^^^^^^^^^^^^^^^
Feb 16 10:02:28 fedora-37-127-0-0-2-2201 cockpit-ws[16002]: systemd_ctypes.bus.BusError: org.freedesktop.DBus.Error.AccessDenied: Failed to get file context on /usr/lib/systemd/system/NetworkManager.service.
```

This is caused by this code:

```python3
    async def setup_path_watch(self, path, interface_name, recursive_props, meta, notify):
        # Watch a single object at "path", but maybe also watch for
        # property changes for all objects below "path".
        
        async def handler(message):
            async with self.watch_processing_lock:
                path = message.get_path()
                name, props, invalids = message.get_body()
                logger.debug('NOTIFY: %s %s %s %s', path, name, props, invalids)
                for inv in invalids:
                    reply, = await self.bus.call_method_async(self.name, path,
                                                              'org.freedesktop.DBus.Properties', 'Get',
                                                              'ss', name, inv)
```

which fails to catch errors from the `Get` call. Probably we need to ignore the error and move on to fetching the other properties, in this case.

It looks like a similar call to `GetManagedObjects` just above may also be similarly unchecked.

allisonkarlitskaya:
For DBus interfaces that notify property invalidations (without specifying the new value) we immediately try to "Get" the new value so that we can send it to JS.  The call to do this was missing a try: to catch the BusError which might result if that Get call fails.

Catch the failure and log a debug message.

Fixes #18358

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -409,9 +409,14 @@ class DBusChannel(Channel):
                 name, props, invalids = message.get_body()
                 logger.debug('NOTIFY: %s %s %s %s', path, name, props, invalids)
                 for inv in invalids:
-                    reply, = await self.bus.call_method_async(self.name, path,
-                                                              'org.freedesktop.DBus.Properties', 'Get',
-                                                              'ss', name, inv)
+                    try:
+                        reply, = await self.bus.call_method_async(self.name, path,
+                                                                  'org.freedesktop.DBus.Properties', 'Get',
+                                                                  'ss', name, inv)
+                    except BusError as exc:
+                        logger.debug('failed to fetch property %s.%s on %s %s: %s',
+                                     name, inv, self.name, path, str(exc))
+                        continue
                     props[inv] = reply
                 notify = {}
                 notify_update(notify, path, name, props)


src/cockpit/channels/dbus.py
```diff
@@ -409,9 +409,13 @@ async def handler(message):
...
-                    reply, = await self.bus.call_method_async(self.name, path,
-                                                              'org.freedesktop.DBus.Properties', 'Get',
-                                                              'ss', name, inv)
+                    try:
+                        reply, = await self.bus.call_method_async(self.name, path,
+                                                                  'org.freedesktop.DBus.Properties', 'Get',
+                                                                  'ss', name, inv)
+                    except BusError as exc:
+                        logger.debug('failed to fetch property %s.%s on %s %s: %s',
+                                     name, inv, self.name, path, str(exc))
```
martinpitt, line 418: This is missing at least a `continue`, as otherwise it would fall through the `props[inv] = reply` below. But presumably it should also something like `del props[inv]` in the error case?

martinpitt (changes requested):

src/cockpit/channels/dbus.py
allisonkarlitskaya: Ah, right!  I had the continue but I accidentally erased it when adding the logging.

From what I understand, `props` has just been unpacked from the message a few lines up.  That's the list of the properties that was just reported as changed, with their values.  It's theoretically possible that a property was reported in both places, I guess, but rather unlikely...

martinpitt: OK, and even in that unlikely case, if it's part of the reported props, it's presumably alright.

martinpitt (approved):
Thanks! LGTM with the fix now.

PR #18366, opened 2023-02-17
build: Move to a webpack module

martinpitt:
Declare ourselves as a "module" type. This allows (and also requires) us to    
move from the old require() and `module.exports` syntax to ES6 `import`s and    
`export`s.    
    
This paves the way for making our modules compatible with ES6 module plugins    
and bundlers like esbuild.    

----

Note that this is a breaking change -- all our starter-kit like derivatives which update to this new cockpit API will have to add `"type": "module"` as well. But I think it's worth it, both for keeping our stuff compatible with the rest of the world and also allowing other cockpit projects to build with esbuild, snowpack, or others.

 - [x] Fix unit tests
 - [x] Move pkg/lib/*2po to `import`s, they currently fail
 - [x] Built on top of #18368 (broken out for easier review)


pkg/lib/cockpit-po-plugin.js
```diff
@@ -5,7 +5,7 @@ const gettext_parser = require('gettext-parser');
 const Jed = require('jed');
 const webpack = require('webpack');
 
-const srcdir = process.env.SRCDIR || path.resolve(__dirname, '..', '..');
+const srcdir = process.env.SRCDIR || '.';
```
KKoukiou, line 8: Here you need ../../

martinpitt: Hmm, why? `__dirname` used to be 'pkg/lib', and the two .. just undid that.

martinpitt: It's supposed to be the top-level source directory. So it could actually be that this breaks for out-of-tree builds, but image-prepare does that, and it seems happy enough.

martinpitt:
[lol](https://github.com/cockpit-project/cockpit/actions/runs/4204055672/jobs/7294218010), terser got updated on npmjs in between me running `node-modules install` and the PR running. Timing is everything!

KKoukiou (approved):

martinpitt:
Landing.. There's over a hundred 3x affected tests, and they don't fail systematically. Good enough.

PR #18367, opened 2023-02-17
python: don't signal already-exited processes

allisonkarlitskaya:
In 95bffd540ce450792341b99d80ffc98762158c8e we implement our own "safe" signal sending logic.  In order to avoid sending signals to PIDs that may have been recycled, we carefully go out of our way to detect if the process has exited, log a message about it, and then ... send the signal anyway.

Add the obviously-missing `return` statement.


martinpitt (approved):
Sheesh. I spotted that once during review, but not there. Sorry :grin:

PR #18372, opened 2023-02-20
lib: Make PO and rsync bundler plugins work for webpack and esbuild

martinpitt:
- Pull out the common code into init() and run() functions, which leaves the webpack/esbuild glue relatively small.
 - Stop having a default export. The bundler config now explicitly needs to import the webpack or esbuild plugin.

---

This was developed in https://github.com/cockpit-project/cockpit-podman/pull/1210 , originally against cockpit-podman. I validated that generated po files are identical, and that watch mode with and without an rsync target work as expected, on all of cockpit, podman with esbuild, and podman with webpack.

### DIFF
--- a/pkg/lib/cockpit-po-plugin.js
+++ b/pkg/lib/cockpit-po-plugin.js
@@ -1,115 +1,136 @@
-import path from 'path';
-import glob from 'glob';
-import fs from 'fs';
-import gettext_parser from 'gettext-parser';
-import Jed from 'jed';
-import webpack from 'webpack';
+import fs from "fs";
+import glob from "glob";
+import path from "path";
 
-const srcdir = process.env.SRCDIR || '.';
+import Jed from "jed";
+import gettext_parser from "gettext-parser";
 
-export default class {
-    constructor(options) {
-        if (!options)
-            options = {};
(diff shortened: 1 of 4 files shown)


pkg/lib/cockpit-po-plugin.js
```diff
@@ -1,115 +1,136 @@
...
+        const dir = rtl_langs.includes(parsed.headers.language) ? "rtl" : "ltr";
+
+        // cockpit.js only looks at "plural-forms" and "language"
+        const chunks = [
+            '{\n',
+            ' "": {\n',
+            `  "plural-forms": ${get_plural_expr(parsed.headers['plural-forms'])},\n`,
+            `  "language": "${parsed.headers.language}",\n`,
+            `  "language-direction": "${dir}"\n`,
+            ' }'
```
jelly, line 63: Offtopic: Should this at one point become a properly serialized json object with JSON.stringify.

martinpitt: Yes, maybe -- I hope that nothing (or not much) depends on the precise formatting of this blob. Happy to do that as a follow-up!

KKoukiou (approved):

PR #18373, opened 2023-02-20
test: Prevent initrd regeneration after installing in OStree

mvollmer:


mvollmer:
This should make testInsightsStatus pass on rhel4edge.

martinpitt (changes requested):
Thanks for tracking this down! However, I'd rather like to have this in https://github.com/cockpit-project/bots/blob/main/images/scripts/ostree.setup#L63 . There we can control it at the right time (during image refreshes), and it will also unbreak other projects which use that image (like cockpit-{podman,ostree}).

mvollmer:
> and it will also unbreak other projects which use that image (like cockpit-{podman,ostree}).

Good point! I'll do that.

PR #18375, opened 2023-02-20
users: append's dropdown's menu to document body

KKoukiou:
The default is to render it inline. However, our tests showed that the clicks were often eaten (actions dialogs not opening causing flakes). I blame this on the Table component messing up with the events, so let's unnest the actions Dropdown from the Table as a fix.

### DIFF
--- a/pkg/users/accounts-list.js
+++ b/pkg/users/accounts-list.js
@@ -110,6 +110,8 @@ const UserActions = ({ account }) => {
                 isPlain
                 isOpen={isKebabOpen}
                 position="right"
+                id="accounts-actions"
+                menuAppendTo={document.body}
                 dropdownItems={actions} />
     );
     return kebab;
--- a/test/verify/check-users
+++ b/test/verify/check-users
@@ -30,7 +30,7 @@ def performUserAction(browser, user, action):
     browser.wait_visible(f"#accounts-list tbody tr:contains({user}) .pf-c-dropdown button")
     browser.click(f"#accounts-list tbody tr:contains({user}) .pf-c-dropdown button")
     browser.wait_visible(f"#accounts-list tbody tr:contains({user}) .pf-c-dropdown button[aria-expanded=true]")
-    browser.click(f"#accounts-list tbody tr:contains({user}) .pf-c-dropdown__menu li:contains({action})")
+    browser.click(f".pf-c-dropdown__menu-item:contains({action})")
 
 
 def createUser(browser, user_name, real_name, password, locked, force_password_change, verify_created=True):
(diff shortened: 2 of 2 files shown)


KKoukiou:
This fixes TestAccounts.testBasic.

pkg/users/accounts-list.js
```diff
@@ -110,6 +110,8 @@ const UserActions = ({ account }) => {
                 isPlain
                 isOpen={isKebabOpen}
                 position="right"
+                id="accounts-actions"
+                menuAppendTo={document.body}
```
martinpitt, line 114: Hmm, everywhere else we write this as `menuAppendTo={() => document.body}`. But quite obviously this works too, nice! https://www.patternfly.org/v4/components/select also documents both possibilities.

However, https://www.patternfly.org/v4/components/select#appending-document-body-vs-parent discourages document.body, and suggests `"parent"` instead, for accessibility. We need to do that for all cases, but test them individually whether it actually works. Let's do that as a follow-up.

martinpitt (approved):
Epharisto!

martinpitt:
Really hard to get tests to green, even with retries. I submitted PR #18377 to fix the rest of it, but it's too much hassle. Instead, I added my fix here, so that ideally everything should be green again, without annoying flakes.

martinpitt (approved):
Re-approving. However, @KKoukiou please land yourself after reviewing my second commit. It's obvious and trivial, but still..

In the meantime, feeding hamsters and crossing fingers! :crossed_fingers:

martinpitt:
The second commit should fix failures like [this](https://artifacts.dev.testing-farm.io/2126f39f-d12c-4261-900d-19678f3dcf3f/), [this](https://artifacts.dev.testing-farm.io/9760a0b6-ee14-4fe8-bba1-1d26e8cc2424/), [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18375-20230221-051235-e18f07a1-arch/log.html#173-2), or [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18375-20230221-051241-e18f07a1-rhel-9-2/log.html#173-2)

martinpitt:
Still a [few](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18375-20230221-064524-dbf64cec-rhel-8-8/log.html#333) [flakes](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18375-20230221-064524-dbf64cec-fedora-37-firefox/log.html#333-2) of the "userdel: user anton is currently used by process 92416" variety, due to 3x affected retries. I'll look at that next (added todo item), but that is independent and much older.

PR #18376, opened 2023-02-21
test: Use skipImage globs in check-packagekit

martinpitt:


jelly (dismissed):
Looks good, all failed tests seem to be flakes so requires some retries.

martinpitt:
Rebased after landing PR #18375, to get rid of these pesky failures.

jelly (approved):

PR #18379, opened 2023-02-21
test: use Machine.ostree_image where possible

jelly:
Prefer using `ostree_image` so we can easily add a new ostree variant for testing.


martinpitt (approved):

jelly:
rhel4edge is still a known failure, merging.

PR #18380, opened 2023-02-21
python: Go back to pipes for spawn channels

martinpitt:
Commit 8f1b75e5d6445a1ba moved SubprocessTransport from pipes to a socketpair(), as cockpit-askpass depends on that. However, we don't want to do that for spawn channels, as fd 0 being a writable Unix socket is e.g. rejected by SELinux on RHEL 8, and may also be unexpected in other scenarios.

So reintroduce support for pipes, make that the default, and add a `use_socket` flag for it which we enable for Peers only.

This changes the argument order, so explicitly pass `window` as a kwarg instead of positional arg in stream.py. But it still makes more sense to add the argument next to the `pty` flag rather than trying to keep the (internal-only) API.    

---

Broken out from PR #18318, see https://github.com/cockpit-project/cockpit/pull/18318#issuecomment-1436414420 and the previous comment for all the gory debug details. This matches what the C bridge does and [fixes running on RHEL 8](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18318-20230221-105343-68dc41d4-centos-8-stream-pybridge/log.html)

I changed #18318 to skip `TestStratis.testEncrypted` on RHEL/CentOS 8 for the time being. Once this (or a replacement) lands, the test should be unskipped again.

### DIFF
--- a/src/cockpit/channels/stream.py
+++ b/src/cockpit/channels/stream.py
@@ -149,7 +149,8 @@ class SubprocessStreamChannel(ProtocolChannel, SubprocessProtocol):
             env.update(dict(e.split('=', 1) for e in environ))
 
         try:
-            transport = SubprocessTransport(loop, self, args, pty, window, env=env, cwd=cwd, stderr=stderr,
+            transport = SubprocessTransport(loop, self, args, pty, window=window,
+                                            env=env, cwd=cwd, stderr=stderr,
                                             preexec_fn=lambda: prctl(SET_PDEATHSIG, signal.SIGHUP))
             logger.debug('Spawned process args=%s pid=%i', args, transport.get_pid())
             return transport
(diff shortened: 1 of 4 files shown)


martinpitt:
[fails one unit test](https://github.com/cockpit-project/cockpit/actions/runs/4231482864/jobs/7350066021), and I also still want to add some new unit tests to check the new flag. But integration tests can run in the meantime.

martinpitt:
argh [mypy](https://github.com/cockpit-project/cockpit/actions/runs/4231814997/jobs/7350790043), fixed.

martinpitt:
Will eventuallly be replaced with PR #18398, moving to draft.

mvollmer:
> as fd 0 being a writable Unix socket is e.g. rejected by SELinux on RHEL 8

Hmm, SELinux rejects passing a unix socket to dbus-daemon in a message. A unix socket as fd 0 is allowed.

mvollmer (commented):
Looks good to me.

mvollmer:
Let's run unskip the stratis test here and run centops-8-stream/pybridge.

martinpitt:
#18398 landed, this is obsolete. All tests now work on c8s/pybridge! :100:

Issue #18335, opened 2023-02-13
Overview/Metrics reports CPU Critical Temperature instead of Current


jwhaugustine:
### Explain what happens

Having updated cockpit from version 284 to 285, all of my physical machines running cockpit web interface are now reporting the CPU's critical temperature on the metrics page instead of their actual temperature.

These are all older PowerEdge systems, R210II R420 R620 on Ubuntu Jammy 22.04 server.

Running ```sensors``` confirms a display issue:
```
coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +46.0°C  (high = +87.0°C, crit = +97.0°C)
Core 0:        +44.0°C  (high = +87.0°C, crit = +97.0°C)
Core 1:        +45.0°C  (high = +87.0°C, crit = +97.0°C)
Core 2:        +46.0°C  (high = +87.0°C, crit = +97.0°C)
Core 3:        +43.0°C  (high = +87.0°C, crit = +97.0°C)
Core 4:        +45.0°C  (high = +87.0°C, crit = +97.0°C)
Core 5:        +40.0°C  (high = +87.0°C, crit = +97.0°C)

coretemp-isa-0001
Adapter: ISA adapter
Package id 1:  +43.0°C  (high = +87.0°C, crit = +97.0°C)
Core 0:        +42.0°C  (high = +87.0°C, crit = +97.0°C)
Core 1:        +40.0°C  (high = +87.0°C, crit = +97.0°C)
Core 2:        +43.0°C  (high = +87.0°C, crit = +97.0°C)
Core 3:        +40.0°C  (high = +87.0°C, crit = +97.0°C)
Core 4:        +41.0°C  (high = +87.0°C, crit = +97.0°C)
Core 5:        +38.0°C  (high = +87.0°C, crit = +97.0°C)
```

### Version of Cockpit

285

### Where is the problem in Cockpit?

Metrics

### Server operating system

Ubuntu

### Server operating system version

22.04.1

### What browsers are you using?

Chrome

tomasmatus:
Thank you for your report, I hope we didn't cause you any inconveniences by alarming that your servers are on fire.

@martinpitt looks like the sensor check was broken in abb9ed4 and `sscanf()` is satisfied with anything that begins with `temp%d` when it should only take `temp%d_input` as a valid temperature sensor.

sch-dmtr:
Same bug, Dell PowerEdge T130
[...]

PR #18381, opened 2023-02-21
Fix cpu temperature sensors detection - FIX COMMIT MESSAGE ON LANDING

tomasmatus:
Fixes: #18335
Also adds files containing critical cpu temperature so our tests catch if incorrect values are sampled.


garrett:
I guess since this is a C file, the symlink hack doesn't work and would require updating the installed version of Cockpit on my system, right?

I've built the branch and still see it, so I assume so. Reporting here, just in case, however.

image

martinpitt:
I tested it with the bind mount. I got a red "100 °C" on main, and "55 °C" with this fix, which matches exactly the value that is shown in my waybar.

PR #18382, opened 2023-02-21
test: Prevent grub2-mkconfig from clobbering BLS entries

mvollmer:
Running grub2-mkconfig will execute /etc/grub.d/10_linux, which will update /boot/loader/entries with the kernel command line from /etc/default/grub.  On at least rhel-9, /etc/default/grub contains a empty GRUB_CMDLINE_LINUX (while /etc/kernel/cmdline is correct), and grub2-mkconfig would thus clobber the BLS entries.

Concretely, "net.ifnames=0" would be lost, and the inter-VM network 10.111.112.0 would not be available anymore since it has been configured using the old name of the interface.

This has only recently become a actual problem since earlier versions of grub2-mkconfig would only look at BLS entries that match the current machine ID.  Our rhel images have entries like

    ffffffffffffffffffffffffffffffff-4.18.0-458.el8.x86_64.conf

that would be missed earlier.  Now all entries are processed.

The fundamental issue is probably that /etc/default/grub does not have the correct cmdline in it, but it is easier to stop this clobbering by moving the files away temporarily than to fix GRUB_CMDLINE_LINUX in /etc/default/grub.

### DIFF
--- a/test/common/storagelib.py
+++ b/test/common/storagelib.py
@@ -605,7 +605,13 @@ echo "UUID=$uuid / auto defaults 0 0" >/new-root/etc/fstab
 echo "UUID=$buuid /boot auto defaults 0 0" >>/new-root/etc/fstab
 dracut --regenerate-all --force
 grub2-install {dev}
-grub2-mkconfig -o /boot/grub2/grub.cfg
+( # HACK - grub2-mkconfig messes with /boot/loader/entries/ and /etc/kernel/cmdline
+  mv /boot/loader/entries /boot/loader/entries.stowed
+  ! test -f /etc/kernel/cmdline || mv /etc/kernel/cmdline /etc/kernel/cmdline.stowed
+  grub2-mkconfig -o /boot/grub2/grub.cfg
+  mv /boot/loader/entries.stowed /boot/loader/entries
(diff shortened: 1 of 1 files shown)


mvollmer:
I would file a bug but I don't know where.

martinpitt:
The approach is okay for me, but it does not work on [CentOS 8](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18382-20230221-121356-9a1f1955-centos-8-stream/log.html#181-2) and [RHEL 8](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18382-20230221-121356-9a1f1955-rhel-8-8/log.html#181-2)

martinpitt:
We already encountered similar bugs before, they were filed against [grub2](https://bugzilla.redhat.com/show_bug.cgi?id=1785460) or [distribution](https://bugzilla.redhat.com/show_bug.cgi?id=1646976) if the file isn't owned by a package. I'd start with grub2, and people there will know better where to reassign.

martinpitt:
@mvollmer : Can you please rebase in your next round to get the TF tests green?

martinpitt (approved):
Thanks, this ought to work now. bots have the last word, but LGTM!

martinpitt:
You didn't rebase, so TF failures are expected; at least for this PR I am very sure that it does not affect gating tests, so I'm willing to let it slide. (The last time I did that we got burned and actually broke the tests..)

PR #18383, opened 2023-02-21
tools: add po plugin as input for webpack-make

jelly:
When changing the po plugin our webpack's should be rebuild.


martinpitt (approved):
Thanks! The test run in #18266 proves that this DTRT.

PR #18388, opened 2023-02-22
test: Bump timeout for CDP response

martinpitt:
The short timeout caused test failures when running on RHEL 8.

----

This broke the downstream release earlier today, and I confirmed that this fixed it. For the Red Hatters out here, see https://gitlab.com/redhat/rhel/rpms/cockpit/-/merge_requests/11


martinpitt:
Hold off on that.. This may just have been a coincidence, the next run with 30s failed again.

PR #18393, opened 2023-02-22
Use whole path in patternfly imports

KKoukiou:
Otherwise we rely on treeshaking mechanisms to remove unused code, thus making our bundle size more vulnerable to bundler's ability to drop dead code.


martinpitt:
Wow, I sincerely hope that you did that with a script!

```
/source/pkg/shell/credentials.jsx
#   271:14  error  'ActionGroup' is not defined  react/jsx-no-undef
```
This also explains the integration test failure.

KKoukiou:
> Wow, I sincerely hope that you did that with a script!
> 
> ```
> /source/pkg/shell/credentials.jsx
> #   271:14  error  'ActionGroup' is not defined  react/jsx-no-undef
> ```
> 
> This also explains the integration test failure.

Of course it was script. https://paste.centos.org/view/78b0ee0e

martinpitt (approved):
Seems to work in general. Let's land this ASAP as it's very prone to bitrot. This has 3x affected retries of pretty much all tests, so will be hard to get green. Do one round of retries to benefit from the recent check-users test fixes, but otherwise I believe we'll need to look the other way :grin:

PR #18398, opened 2023-02-24
python: Peer startup async overhaul, via ferny

allisonkarlitskaya:
hihi bots

- [x] allisonkarlitskaya/ferny#2


test/pytest/test_bridge.py
```diff
@@ -161,6 +161,7 @@ async def assert_bus_notify(self, path: str, iface: str, expected: Dict[str, obj
         if bus is None:
             bus = await self.ensure_internal_bus()
         notify = await self.next_msg(bus)
+        assert 'notify' in notify, notify
```
martinpitt, line 165: I don't understand the second "notify". Doesn't pytest already show the notify value on a failure? If not, then `self.assertIn('notify', notify)` may be cleaner?

pyproject.toml
```diff
@@ -16,16 +16,18 @@ include = [
   "cockpit",
   "cockpit.channels",
   "cockpit.data",
-  "systemd_ctypes",
+  "cockpit._vendor",
```
martinpitt, line 19: :clap: This commit can land right away for my sake.

tools/ferny
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
martinpitt, line 1: Meh -- can we just change tools/systemd_ctypes to tools/install-submodules ?

src/cockpit/transports.py
```diff
@@ -330,31 +328,32 @@ def __init__(self,
...
                                              stdin=session_fd, stdout=session_fd,
                                              start_new_session=True, **kwargs)
-        finally:
             os.close(session_fd)
 
+            in_fd, out_fd = self._pty_fd, self._pty_fd
+            self._eio_is_eof = True
+
+        else:
+            self._process = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, **kwargs)
```
martinpitt, line 346: This "Revert "python: use sockets instead of pipes in SubprocessTransport"" commit should drop the test skip for TestStratis.testEncrypted, as it will fix it on c8s.

src/cockpit/peer.py
```diff
@@ -19,125 +19,99 @@
...
         self.init_host = init_host or router.init_host
 
-        self.authorize_pending = None
+    # Initialization
+    async def do_connect_transport(self) -> asyncio.Transport:
+        raise NotImplementedError
 
-    def spawn(self, argv: Sequence[str], env: Sequence[str], **kwargs) -> asyncio.Transport:
+    async def spawn(self, argv: Sequence[str], env: Sequence[str], **kwargs) -> asyncio.Transport:
+        # Not actually async...
```
martinpitt, line 59: So why is it declared async then? Do subclasses need to override this function? (not in this commit)

test/pytest/pseudo.py
```diff
@@ -1,15 +1,15 @@
+from cockpit._vendor.ferny import interaction_client
```
martinpitt, line 1: Please re-sort; stdlib modules first, then third-party ones.

test/pytest/pseudo.py
```diff
@@ -1,15 +1,15 @@
...
-import json
 import sys
 
 pw = os.environ.get('PSEUDO_PASSWORD')
 if pw:
-    challenge = '\n{"command":"authorize", "challenge":"plain1:", "cookie":"dontcare", "prompt": "can haz pw?"}\n'
-    sys.stdout.write(f'{len(challenge)}\n{challenge}')
-    sys.stdout.flush()
+    reader, writer = os.pipe()
+    interaction_client.interact(2, writer, ['-', 'can haz pw?'], {})
```
martinpitt, line 8: What does the single dash do?

martinpitt (changes requested):
Thanks! Admittedly I didn't read/understand the guts of "major peer init overhaul" in detail, but the approach in the commit message sounds good. However, I must say it feels rather detached to import the InteractionAgent from a separate module. I had initially thought ferny would just be used to reimplement cockpit-ssh, not being in the core part of the bridge peer logic (e.g. superuser has nothing to do with ssh). Perhaps we should just move the ferny InteractionAgent, or the whole module, into cockpit itself, but I suppose that collides with your desire to keep ferny as a standalone project. Well, I honestly haven't made up my mind about this. But I'm excited to see this stuff come together!

test/pytest/pseudo.py
allisonkarlitskaya: interact() is expecting to get `sys.argv` as that argument, so `'-'` is a stand-in for where the program name would be.

test/pytest/pseudo.py
allisonkarlitskaya: Done.

src/cockpit/peer.py
allisonkarlitskaya: It's complicated.

By all rights, this should probably properly be async.  Compare with `asyncio.create_subprocess_exec()` for example.  If we ever move to using that function, then this would definitely need to be made async.

The funny thing is, though: even though asyncio's own subprocesses APIs are marked async, they're also just calling `subprocess.Popen()` inside, in the same blocking way that we are....

So, it's a bit of a shoulder shrug.  This should be async, but it isn't, and even if it called another async API, it would still not be async.  :shrug:

src/cockpit/transports.py
allisonkarlitskaya: Done, thanks!

tools/ferny
allisonkarlitskaya: Meh.

Are you volunteering? :)

At some point I start to wonder if we can remove some of this scripting.  Things like `systemd_ctypes` and `ferny` are actually pretty small and might work fine via the normal git commands.

It's also interesting to note that this works:

```sh
git clone --recurse-submodules='modules/*' https://github.com/cockpit-project/cockpit
```

and avoids the heavy pixel references and node_modules repositories.

allisonkarlitskaya:
The main point of why we're using ferny for stuff that's far away from SSH: we need to perform a very similar askpass interaction for both SSH and `sudo`.  ferny implements that for SSH, so why not also use it for sudo, instead of having two separate code paths?

tools/ferny
martinpitt: Yes, it also works fairly well with pixel refs. The problem is that this --recurse-submodules is a magic "unbreak my setup", instead of just being a property of the actual repo (so that `git clone <url>` would just DTRT). Unfortunately git submodules leave a lot to be desired :cry: In other words, we need to get along with non-recursive clones anyway, so we can just as well always check them out ourselves.

I think we could even drop the fancy git-utils.sh machinery for modules/* if it gets in the way -- these are all tiny . If it helps, I can send a PR that renames/generalizes tools/systemd_ctypes first, but it's fairly obvious either way.

test/pytest/test_bridge.py
martinpitt: I don't understand the second "notify". Doesn't pytest already show the notify value on a failure? If not, then `self.assertIn('notify', notify)` may be cleaner?

tools/ferny
martinpitt: I'm willing to clean up this duplication in a follow-up. This is really rather trivial, but ok to not block this PR.

test/pytest/mocktransport.py
```diff
@@ -1,11 +1,55 @@
...
+    running = my_subprocesses()
+    if not running:
+        return
+
+    print('Some subprocesses still running', running)
+    # Clear it out for the sake of the other tests
+    subprocess.call(['pkill', '-9', '-P', f'{os.getpid()}'])
+    for _ in range(100):    # zombie vacuum
+        os.waitpid(-1, os.WNOHANG)  # will eventually throw, failing the test (which we want)
+    assert False  # more than 100?
```
martinpitt, line 29: Just to be clear, as soon as `running` is non-empty, this should *always* fail, right?

test/pytest/test_bridge.py
```diff
@@ -558,106 +406,106 @@
...
+        if channel == '':
+            control = json.loads(msg)
+            command = control['command']
+            if command == 'done':
+                continue
+            elif command == 'close':
+                assert 'problem' not in control
+                return
+
+    await settle_down()
```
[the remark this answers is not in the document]
martinpitt: It's right -- this is a `while true` loop with no `break` exit path, just a `return`. Presumably this should move into the `return` code path, but I'd actually find it cleaner to actually do `break` on `close`.

test/pytest/pseudo.py
```diff
@@ -1,15 +1,16 @@
...
 
+from cockpit._vendor.ferny import interaction_client
+
 pw = os.environ.get('PSEUDO_PASSWORD')
 if pw:
-    challenge = '\n{"command":"authorize", "challenge":"plain1:", "cookie":"dontcare", "prompt": "can haz pw?"}\n'
-    sys.stdout.write(f'{len(challenge)}\n{challenge}')
-    sys.stdout.flush()
+    reader, writer = os.pipe()
+    interaction_client.interact(2, writer, ['-', 'can haz pw?'], {})
```
martinpitt, line 9: > interact() is expecting to get sys.argv as that argument, so '-' is a stand-in for where the program name would be.

Do you mind adding a comment what the `-` does?

martinpitt (changes requested):
Thanks! I already reviewed a previous version of this, so fortunately this didn't take too long. I closed the old threads and copied one or two outstanding ones here, so that this is easier to follow.

I'm happy with the general approach, just some minor details. In particular,  let's get CodeQL green again and fix that unreachable statement (should be easy). Thanks!

test/pytest/test_bridge.py
allisonkarlitskaya: You're right.  I'll fix it.

test/pytest/pseudo.py
allisonkarlitskaya: OK

test/pytest/test_bridge.py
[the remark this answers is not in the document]
allisonkarlitskaya: There's another `return` path higher up, though.  I'll add the `await settle_down()` at both of them.

martinpitt (approved):
Thanks! Acking the .github-changes, will land then.

martinpitt:
I think this has broken the pybridge on c8s completely, like here: https://cockpit-logs.us-east-1.linodeobjects.com/pull-18519-20230317-125210-b369d0f7-centos-8-stream-pybridge/log.html#335

This PR didn't run on c8s/pybridge.

PR #18405, opened 2023-02-26
base1, lib: Avoid top-level await

martinpitt:
This is still incompatible with esbuild and also older nodejs versions.

----

@KKoukiou : This should help with esbuild-ifying cockpit itself. Any other trouble?

### DIFF
--- a/pkg/base1/test-file.js
+++ b/pkg/base1/test-file.js
@@ -1,6 +1,8 @@
 import cockpit from "cockpit";
 import QUnit from "qunit-tests";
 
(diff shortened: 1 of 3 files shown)


mvollmer (changes requested):
Let's also remove `topLevelAwait: true` from webpack.config.js, if possible.

martinpitt:
@mvollmer Indeed, well spotted/remembered! Done.

mvollmer (approved):

PR #18416, opened 2023-02-27
systemd/service: Track pinned units by unit name (ID), not object path; code cleanups

martinpitt:
Judging by the .filter() call, this had been the intention from the        
start. There is no advantage with tracking the object paths instead:        
        
 * They are not any more unique than IDs, they are even the same for        
   system and user units.        
        
 * They require an extra LoadUnit() to compute, which we want to stop        
   doing.        
        
 * They are an internal implementation detail of systemd, which can        
   change in principle.        
        
This is a breaking change, as it will disregard existing        
`systemd:pinnedUnits` settings from localstore. This has existed for        
less than a year (commit 065bf82d83), is not very easy to discover, and        
is not crucial, so we can hopefully get away with it.        

----

Also add some code cleanup.

## Services: Pinned units need to be re-done

The data storage for "pinned" systemd units has changed. After upgrading to this Cockpit version, you need to re-pin units.

### DIFF
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -222,7 +222,7 @@ export class ServiceDetails extends React.Component {
             unit_properties: {},
             showDeleteDialog: false,
             unitPaths: [],
-            isPinned: this.props.pinnedUnits.includes(this.props.unit.path),
+            isPinned: this.props.pinnedUnits.includes(this.props.unit.Id),
         };
 
         this.onOnOffSwitch = this.onOnOffSwitch.bind(this);
@@ -331,8 +331,8 @@ export class ServiceDetails extends React.Component {
 
     pinUnit() {
         const newPinned = this.state.isPinned
-            ? this.props.pinnedUnits.filter(unitId => unitId != this.props.unit.path)
-            : [...this.props.pinnedUnits, this.props.unit.path];
+            ? this.props.pinnedUnits.filter(unitId => unitId != this.props.unit.Id)
+            : [...this.props.pinnedUnits, this.props.unit.Id];
 
         localStorage.setItem('systemd:pinnedUnits', JSON.stringify(newPinned));
         this.setState(prevState => ({ isPinned: !prevState.isPinned }));
--- a/pkg/systemd/services/service-details.jsx
+++ b/pkg/systemd/services/service-details.jsx
@@ -222,6 +222,7 @@ export class ServiceDetails extends React.Component {
             unit_properties: {},
             showDeleteDialog: false,
             unitPaths: [],
+            isPinned: this.props.pinnedUnits.includes(this.props.unit.path),
         };
 
         this.onOnOffSwitch = this.onOnOffSwitch.bind(this);
@@ -329,20 +330,12 @@ export class ServiceDetails extends React.Component {
     }
 
     pinUnit() {
-        let pinnedUnits = [];
-        try {
-            pinnedUnits = JSON.parse(localStorage.getItem('systemd:pinnedUnits')) || [];
-        } catch (err) {
-            console.warn("exception while parsing systemd:pinnedUnits", err);
-        }
+        const newPinned = this.state.isPinned
+            ? this.props.pinnedUnits.filter(unitId => unitId != this.props.unit.path)
+            : [...this.props.pinnedUnits, this.props.unit.path];
 
-        if (this.props.isPinned) {
-            pinnedUnits = pinnedUnits.filter(unitId => unitId != this.props.unit.path);
-        } else {
-            pinnedUnits.push(this.props.unit.path);
-        }
-
-        localStorage.setItem('systemd:pinnedUnits', JSON.stringify(pinnedUnits));
+        localStorage.setItem('systemd:pinnedUnits', JSON.stringify(newPinned));
+        this.setState(prevState => ({ isPinned: !prevState.isPinned }));
         dispatchEvent(new Event('storage'));
     }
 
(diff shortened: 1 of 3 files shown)


pkg/systemd/services/service-details.jsx
```diff
@@ -536,7 +530,25 @@ export class ServiceDetails extends React.Component {
 
         const tooltipMessage = enabled ? _("Stop and disable") : _("Start and enable");
         const hasLoadError = this.props.unit.LoadState !== "loaded" && this.props.unit.LoadState !== "masked";
-        const loadError = this.props.unit.LoadError ? this.props.unit.LoadError[1] : null;
+
+        if (hasLoadError) {
+            const path = "/system/services" + (this.props.owner === "user" ? "#/?owner=user" : "");
+            const loadError = this.props.unit.LoadError ? this.props.unit.LoadError[1] : null;
+            const title = loadError || _("Failed to load unit");
```
cockpituous, line 537: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18416-20230227-161258-e3110a7d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/service-details.jsx
```diff
@@ -536,7 +530,25 @@
...
+
+            return <EmptyStatePanel
+                icon={ExclamationCircleIcon}
+                title={title}
+                paragraph={this.props.unitId}
+                action={
+                    <Button variant="link"
+                            component="a"
+                            onClick={() => cockpit.jump(path, cockpit.transport.host)}>
+                        {_("View all services")}
```
cockpituous, line 547: These 9 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18416-20230227-161258-e3110a7d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/service-details.jsx
```diff
@@ -536,7 +530,25 @@
...
+                    </Button>
+                }
+            />;
+        }
```
cockpituous, line 551: These 3 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18416-20230227-161258-e3110a7d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/services.jsx
```diff
@@ -343,6 +330,15 @@ class ServicesPageBody extends React.Component {
         return true;
     }
 
+    loadPinnedUnits() {
+        try {
+            this.setState({ pinnedUnits: JSON.parse(localStorage.getItem('systemd:pinnedUnits')) || [] });
+        } catch (err) {
+            console.warn("exception while parsing systemd:pinnedUnits", err);
+            this.setState({ pinnedUnits: [] });
+        }
```
cockpituous, line 339: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18416-20230227-161258-e3110a7d-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

martinpitt:
I took out the [third commit](https://github.com/cockpit-project/cockpit/commit/e3110a7d01c2f7ed014c7d1bd1b77a076d11d27d) which I believe is the [problematic one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18416-20230227-161258-e3110a7d-fedora-36/log.html). I'll investigate that separately.

pkg/systemd/services/services.jsx
```diff
@@ -343,6 +331,15 @@ class ServicesPageBody extends React.Component {
...
```
cockpituous, line 340: These 4 added lines are not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18416-20230227-182216-6d0ed185-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/systemd/services/service-details.jsx
```diff
@@ -222,7 +222,7 @@ export class ServiceDetails extends React.Component {
             unit_properties: {},
             showDeleteDialog: false,
             unitPaths: [],
-            isPinned: this.props.pinnedUnits.includes(this.props.unit.path),
```
KKoukiou: I did not touch this page since since a long time, but I recall very well, that we used to always work with object-paths instead of unit.Id so that we include unit aliases in the list. However this is a corner case, when these are also pinned, so I am ok with the change.

KKoukiou (approved):

pkg/systemd/services/service-details.jsx
martinpitt: There is no difference. When I pin "dbus-broker.service" on main, the aliased dbus.service appears as pinned as well:

image

```
> localStorage.getItem('systemd:pinnedUnits')
'["/org/freedesktop/systemd1/unit/dbus_2dbroker_2eservice"]'
```

they are *really* the same objects, with the same object paths.

martinpitt: Oh, wait, I see what you mean, sorry!  You meant the other way around. Right, with this branch only the pinned unit would be shown, not the aliased one. So that is a visible feature change indeed. It feels a bit more natural to me that way, but if you think its not, then I can rework this.

I am trying really hard to get rid of the object paths, as they are incredibly expensive.

KKoukiou: I don't have strong opinion. I totally agree with this change when thinking of the page load cost benefit as well.

tomasmatus (approved):
Thank you! You're right that the unit ID is more appropriate in this case

PR #18423, opened 2023-02-28
.github: scan JavaScript code with codeql

jelly:
This still has plenty of issues but I think it's time to start scanning

* Tons of false positives in test/ which are used from Python
* kdump needs some major re-factoring to make it more sane. I've fixed the state issues in a branch but I think it's more worthwhile to fix it in general
* Some interesting related issues which require investigation.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -19,8 +19,8 @@ jobs:
       matrix:
         language:
           - python
+          - javascript
           # Let's enable these one by one
-          #- javascript
           #- cpp
 
     steps:


martinpitt:
What's your goal here? One approach would be to ignore all current 70 issues to avoid introducing new ones, but if we are honest, we'll just never fix them then. Once we waive them away (which we need to to get green), they will be really hard to find again.

So I'd rather do a different approach: Go over them once, fix the trivial stuff, and then change the codeql scan to only look at / fail on the 20 security alerts -- then fix/waive these (I think you said there are a lot of false positives?). Then we can fix the 12 warnings and 38 notes in separate steps.

jelly:
> What's your goal here? One approach would be to ignore all current 70 issues to avoid introducing new ones, but if we are honest, we'll just never fix them then. Once we waive them away (which we need to to get green), they will be really hard to find again.

Yes, you are probably right here, once we ignore there is no way back.

> So I'd rather do a different approach: Go over them once, fix the trivial stuff, and then change the codeql scan to only look at / fail on the 20 security alerts -- then fix/waive these (I think you said there are a lot of false positives?). Then we can fix the 12 warnings and 38 notes in separate steps.

I think I figured most trivial stuff.

I've went through and marked all the test related things as non-applicable. The rest needs more investigation.

jelly:
@martinpitt ok, there are now ~ 8 issues left of which I am unsure how to fix. So shall we merge it and then sort those out?

martinpitt (approved):
Sure, sounds good! At least we'll prevent introducing new errors then.

PR #18425, opened 2023-02-28
test: add setup_provisioned_hosts helper for multi machine tests

jelly:
Introduce a new helper function which sets the hostname for all test machines, disables preload of pages when required and adds an Arch Linux systemd 253 / sssd related workaround. This writes all provisioned machines to /etc/hosts so DNS resolving is speedy.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -2094,6 +2094,22 @@ class MachineCase(unittest.TestCase):
         if not self.machine.ostree_image and self.file_exists(disallowed_conf):
             self.sed_file('/root/d', disallowed_conf)
 
+    def setup_provisioned_hosts(self, disable_preload: bool = False):
+        """Setup provisioned hosts for testing
+
+        This sets the hostname of all machines to the name given in the
+        provision dictionary and optionally disabled preload when it runs
+        on a development build.
+        """
+
+        for name, m in self.machines.items():
+            m.execute(f"hostnamectl set-hostname {name}")
+            # Disable preloading on all machines ("machine1" is done in testlib.py)
+            # Preloading on machines with debug build can overload the browser and cause slowness and browser crashes
+            # In these tests we actually switch between machines in quick succession which can make things even worse
+            if disable_preload and self.is_devel_build():
+                self.disable_preload("packagekit", "playground", "systemd", machine=m)
+
 
 ###########################
 # Global helper functions
--- a/test/verify/check-shell-host-switching
+++ b/test/verify/check-shell-host-switching
@@ -113,18 +113,10 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
     def setUp(self):
         super().setUp()
 
-        # Disable preloading on all machines ("machine1" is done in testlib.py)
-        # Preloading on machines with debug build can overload the browser and cause slowness and browser crashes
-        # In these tests we actually switch between machines in quick succession which can make things even worse
-        if self.is_devel_build():
-            for machine in ["machine2", "machine3"]:
-                self.disable_preload("packagekit", "playground", "systemd", machine=self.machines[machine])
-        # Also, quick logouts cause async preloads to run into "ReferenceError: cockpit is not defined"
-        self.disable_preload("packagekit", "playground", "systemd")
-
+        self.setup_provisioned_hosts(disable_preload=True)
+        # Override hostname from machine1 to localhost
         self.machines["machine1"].execute("hostnamectl set-hostname localhost")
-        self.machines["machine2"].execute("hostnamectl set-hostname machine2")
-        self.machines["machine3"].execute("hostnamectl set-hostname machine3")
+
         self.setup_ssh_auth()
 
         # removing machines interrupts channels
--- a/test/verify/check-shell-multi-machine
+++ b/test/verify/check-shell-multi-machine
@@ -160,16 +160,7 @@ class TestMultiMachineAdd(MachineCase):
         self.machine2 = self.machines['machine2']
         self.machine3 = self.machines['machine3']
 
-        for name in ['machine2', 'machine3']:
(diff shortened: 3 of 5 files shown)


jelly:
We should refer to https://github.com/systemd/systemd/issues/26620

test/verify/check-shell-host-switching
```diff
@@ -113,18 +113,12 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
...
-        # In these tests we actually switch between machines in quick succession which can make things even worse
-        if self.is_devel_build():
-            for machine in ["machine2", "machine3"]:
-                self.disable_preload("packagekit", "playground", "systemd", machine=self.machines[machine])
-        # Also, quick logouts cause async preloads to run into "ReferenceError: cockpit is not defined"
+        self.setup_provisioned_hosts(disable_preload=True)
+        # Override hostname from machine1 to localhost
+        self.machines["machine1"].execute("hostnamectl set-hostname localhost")
+        # Quick logouts cause async preloads to run into "ReferenceError: cockpit is not defined"
         self.disable_preload("packagekit", "playground", "systemd")
```
martinpitt, line 120: This is redundant now, as setup_provisioned_hosts also does it for  machine1.

test/verify/check-shell-host-switching
```diff
@@ -113,18 +113,12 @@ class TestHostSwitching(MachineCase, HostSwitcherHelpers):
...
-        # Disable preloading on all machines ("machine1" is done in testlib.py)
-        # Preloading on machines with debug build can overload the browser and cause slowness and browser crashes
-        # In these tests we actually switch between machines in quick succession which can make things even worse
-        if self.is_devel_build():
-            for machine in ["machine2", "machine3"]:
-                self.disable_preload("packagekit", "playground", "systemd", machine=self.machines[machine])
-        # Also, quick logouts cause async preloads to run into "ReferenceError: cockpit is not defined"
+        self.setup_provisioned_hosts(disable_preload=True)
+        # Override hostname from machine1 to localhost
+        self.machines["machine1"].execute("hostnamectl set-hostname localhost")
```
martinpitt, line 118: That's a  bit ugly now -- we could just change the test to assume "machine1" instead. But that can be a follow-up.

test/common/testlib.py
```diff
@@ -2092,6 +2092,40 @@ def enable_root_login(self):
...
+            m.execute(f"hostnamectl set-hostname {name}")
+            # Disable preloading on all machines ("machine1" is done in testlib.py)
+            # Preloading on machines with debug build can overload the browser and cause slowness and browser crashes
+            # In these tests we actually switch between machines in quick succession which can make things even worse
+            self.machine = m
+            if disable_preload and self.is_devel_build():
+                self.disable_preload("packagekit", "playground", "systemd")
+
+            # Workaround systemd DNS issue
+            if machine.image == "arch":
```
martinpitt, line 2125: This will soon hit us on other Fedora /debian-testing etc. images as well, so let's not make it arch specific. Instead, please add a `# HACK` comment pointing to the systemd issue.

test/verify/check-shell-multi-os
```diff
@@ -100,6 +100,9 @@ class TestMultiOS(MachineCase):
         dev_m = self.machine
         dev_b = self.browser
 
+        if self.machine.image == "arch":
+            self.write_file("/etc/hosts", "10.111.113.5 stock", append=True)
```
martinpitt, line 104: Why doesn't the common function work here? I suppose "stock" is set dynamically somewhere?

martinpitt (changes requested):
Thanks! Some minor cleanup to do and some questions. I also triggered a few tests to check this on other OSes, for some reason they weren't triggered.

test/verify/check-shell-host-switching
jelly: Agreed, wanted to see if this was stable enough first.

test/common/testlib.py
jelly: As Fedora does not use systemd-resolved I think, this should not happen there. This was worked around for Arch in bots by disabling mDNS. We should do the same for the other images.

martinpitt: Right, good point. Also, this is obsolete now.

test/common/testlib.py
```diff
@@ -2092,6 +2092,33 @@ def enable_root_login(self):
...
+    def setup_provisioned_hosts(self, disable_preload=False):
+        """Setup provisioned hosts for testing
+
+        This sets the hostname of all machines to the name given in the
+        provision dictionary and optionally disabled preload when it runs
+        on a development build.
+        """
+        host_entries = ""
+        for name, m in self.provision.items():
+            print(m)
```
martinpitt, line 2104: Debugging leftover

test/common/testlib.py
```diff
@@ -2092,6 +2092,33 @@ def enable_root_login(self):
...
+            except KeyError:
+                continue
+
+        machine = self.machine
+        for name, m in self.machines.items():
+            m.execute(f"hostnamectl set-hostname {name}")
+            # Disable preloading on all machines ("machine1" is done in testlib.py)
+            # Preloading on machines with debug build can overload the browser and cause slowness and browser crashes
+            # In these tests we actually switch between machines in quick succession which can make things even worse
+            self.machine = m
```
martinpitt, line 2117: This is ugly. disable_preload as a `machine=` kwarg, please use that instead.

martinpitt (changes requested):
Thanks! Some tests are red, I didn't check them yet.

test/common/testlib.py
jelly: That and `host_entries` is totally unused

martinpitt (approved):
Thanks! LGTM. Restarted one flaked test.

PR #18422, opened 2023-02-28
systemd/services: Avoid excessive D-Bus calls and simplify state management

martinpitt:
The services list does way too much: It loads all units, even the ones which are completely unused, and thus systemd intentionally ignores them. On top of that, it was getting all properties of all interfaces of all units plus installing proxies, so loading the page required an avalanche of (literally) thousands of D-Bus messages; and it was not even using most of the properties in the summary view.

This PR optimizes this, and drops all extended test timeouts related to the services page. Please see the individual commits for details.

---
    
TODO/Blockers:
 - [x] #18411
 - [x] #18416
 - [x] #18420
 - [x] #18445
 - [x] handle transient units which go away    
 - [x] loading sometimes triggers an expensive loop, causing systemd to spin 100% CPU for 30s    
 - [x] fix lagging failure notifications; difficult to fix on main, but easy after the state rework
 - [x] fix [TestServices.testConditions on c8s/pybridge](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18422-20230228-173303-e246369c-centos-8-stream-pybridge/log.html#328-2) (also happens with C bridge)
 - [x] update pixel tests for changed "unloaded test.service" behaviour in testBasic

Spotted flakes while working on this, which are independent:
 - [x] failure to switch to details page: [example 1](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18422-20230302-093755-0a6f967e-arch/log.html#339-2), [example 2](https://artifacts.dev.testing-farm.io/b2c3c498-912e-4511-966c-8c081eb9744b/), [example 3](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18422-20230302-135425-7d52084a-rhel-8-8/log.html#305-2). This also [happens on main](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18442-20230303-083420-6f3008ed-ubuntu-stable/log.html#297). Split out to issue #18443
 - [x] fix `testTransientUnits` [flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18422-20230303-201537-c95d73be-fedora-37-devel/log.html#339), [other flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18422-20230303-185421-c95d73be-ubuntu-stable/log.html#339)
[...]

PR #18426, opened 2023-02-28
systemd: Don't wind back time when adding timer properties

mvollmer:
By the time the timer proxy is ready, the unit state might have already been replaced with something new.

This was the reason for this failure: https://cockpit-logs.us-east-1.linodeobjects.com/pull-4465-20230227-232035-10d7d9c6-centos-8-stream-pybridge-cockpit-project-cockpit/log.html#291

### DIFF
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -685,15 +685,17 @@ class ServicesPageBody extends React.Component {
             if (unitNew.ActiveState == "active") {
                 const timer_unit = systemd_client[this.props.owner].proxy('org.freedesktop.systemd1.Timer', unitNew.path);
                 timer_unit.wait(() => {
-                    if (timer_unit.valid)
-                        if (this.addTimerProperties(timer_unit, unitNew)) {
+                    if (timer_unit.valid) {
+                        const unit = this.state.unit_by_path[path];
+                        if (unit && this.addTimerProperties(timer_unit, unit)) {
                             this.setState(prevState => ({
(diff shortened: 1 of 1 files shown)


martinpitt:
I don't understand how this is related to [this failure of test.service](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4465-20230227-232035-10d7d9c6-centos-8-stream-pybridge-cockpit-project-cockpit/log.html#300-2)? This isn't a timer. 

With this fix it [still fails the same way](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18426-20230228-144758-8b3eb125-centos-8-stream-pybridge/log.html#326-2).

But this code is all horrible anyway.. I'm cleaning it up in #18422, and I am getting very close.

mvollmer:
> I don't understand how this is related to [this failure of test.service](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4465-20230227-232035-10d7d9c6-centos-8-stream-pybridge-cockpit-project-cockpit/log.html#300-2)? This isn't a timer.

It's probably not.

martinpitt (changes requested):
This is a hack. setState() should always copy new state, not change the existing problem. I did that in one of my intermediate rewrites as well, but this shouldn't land. The current state management is broken in a lot of ways and causes lots of race conditions. #18422 now finally has a data model which both avoids the races and the extensive copying.

I think we can/should just close this, but please yell if you disagree. I'm really happy to discuss!

Issue #18424, opened 2023-02-28
testlib: add MachineCase.reboot and MachineCase.wait_reboot functions which allows restart messages


jelly:
Tests which call `m.reboot` or `m.wait_reboot` should automatically call `self.allow_restart_journal_mesages` so we don't have to add this for every test which uses these functions.

To implement this we will need to extend the `MachineCase` class and add two new functions `reboot` and `wait_reboot` which calls `self.allow_restart_journal_messages` and either `self.machine.reboot()` or `self.machine.wait_reboot()` taking the `timeout_sec` parameter into account.

gauravmishra2123:
Hey there! I would like to work on this issue

jelly:
Sure, there are no pre-assigned issues so please try to setup a Cockpit development environment and ask any questions if they come up.

gauravmishra2123:
> Sure, there are no pre-assigned issues so please try to setup a Cockpit development environment and ask any questions if they come up.

so I clone the project locally but I'm unable to understand how to implement changes please help me in this

gauravmishra2123:
> Sure, there are no pre-assigned issues so please try to setup a Cockpit development environment and ask any questions if they come up.

what exactly I have to do is extend the **MachineCase** class and add two functions **reboot** which calls self.machine.reboot().  and **wait_reboot** which calls self.machine.wait_reboot() 
the thing in which I'm confused is **timeout_sec** parameter please help me out in this

jelly:
> > Sure, there are no pre-assigned issues so please try to setup a Cockpit development environment and ask any questions if they come up.
> 
> what exactly I have to do is extend the **MachineCase** class and add two functions **reboot** which calls self.machine.reboot(). and **wait_reboot** which calls self.machine.wait_reboot() the thing in which I'm confused is **timeout_sec** parameter please help me out in this

The new function should accept a `timeout_sec` parameter and pass it through.

mrdetective:
Is it necessary to create a new subclass that inherits from the "MachineChase" class in order to make changes, or can modifications be made directly to the "MachineChase" class?

Deeppjp116:
Hello,

I'm Deep. I'm interested in working on this issue. Could you please explain the problem and guide me on setting up the repository on my computer? I'm eager to contribute and give my best effort to resolve this. This will be my first valuable contribution. Thank you for giving me this opportunity.

martinpitt:
[...]

PR #18431, opened 2023-03-01
Add two new functions reboot and wait_reboot

player-29:
Added two new functions reboot and wait_reboot which is calling self.allow_restart_journal_messages and self.machine.reboot(timeout_sec=timeout_sec)

PR #18432, opened 2023-03-01
added MachineCase.reboot and MachineCase.wait_reboot functions which allows restart messages

gauravmishra2123:
solved #18424 
To implement this feature,  I extend the MachineCase class and add the reboot and wait_reboot methods that automatically call self.allow_restart_journal_messages before calling self.machine.reboot() or self.machine.wait_reboot()
Now i can use ExtendedMachineCase as the base class for your test cases that require reboot or wait_reboot functionality,
also pass a timeout_sec parameter to reboot or wait_reboot

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -2097,8 +2097,6 @@ class MachineCase(unittest.TestCase):
 ###########################
 
 
-import unittest
-from datetime import timedelta
 
 class ExtendedMachineCase(unittest.TestCase):
     def reboot(self, timeout_sec=None):
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -37,6 +37,8 @@ import unittest
 import gzip
 import itertools
 import glob
+import unittest
+from datetime import timedelta
 
 from typing import Any, Callable, Dict, List, Optional, Union
 
@@ -2091,6 +2093,36 @@ class MachineCase(unittest.TestCase):
         disallowed_conf = '/etc/cockpit/disallowed-users'
         if not self.machine.ostree_image and self.file_exists(disallowed_conf):
             self.sed_file('/root/d', disallowed_conf)
+
+###########################
+
+
+import unittest
+from datetime import timedelta
+
+class ExtendedMachineCase(unittest.TestCase):
+    def reboot(self, timeout_sec=None):
+        self.allow_restart_journal_messages()
+        self.machine.reboot()
+        if timeout_sec is not None:
+            self.machine.wait_for_ssh(timeout=timedelta(seconds=timeout_sec))
+
+    def wait_reboot(self, timeout_sec=None):
+        self.allow_restart_journal_messages()
+        self.machine.wait_reboot()
+        if timeout_sec is not None:
+            self.machine.wait_for_ssh(timeout=timedelta(seconds=timeout_sec))
+
+
+
+class MyTestCase(ExtendedMachineCase):
+    def test_reboot_with_timeout(self):
+        self.reboot(timeout_sec=30)
+
+    def test_wait_reboot_with_timeout(self):
+        self.wait_reboot(timeout_sec=30)
+
+
 
 
 ###########################


garrett (changes requested):
What's this about?

You need to explain what a pull request actually does in the title and add a bit more information in the summary area too. Additionally, your commit messages should also explain what they do too.

(Numbers don't tell us anything.)

Thanks in advance for changing the commit messages (which will require a force push to the remote branch) and also this pull request's title and summary!

Also, in the future, please use a short branch name to describe what the purpose of the change is too. (You don't need to change it here.)

garrett (changes requested):
I still do not see descriptions. Please fix.

garrett:
Here's an example of other issues and pull requests, from my GitHub notifications view:

image

I saw this one in there and saw "_Resolved issue #18424_" and had no clue what this was about.

My notifications list often gets to 30 to 100 different issues or pull requests (depending on the day and how active everyone is and if I'm ignoring the list while concentrating on design tasks and such). Now imagine if they all said something like "_Resolved issue #18424_" and also didn't have a description.

It's likely that this specific PR is not for me to review (as I don't think it changes the UI (user interface) or impacts the UX (user experience) at all, but changes some lower level things)... but how am I supposed to know that from the (lack of description), especially between 30 to 100 other notifications?

Thanks for adding a descriptive title and summary! If these commits can be merged together, then it might be able to be squash merged (which would pick up the description), but otherwise those still need changing too.

gauravmishra2123:
> I still do not see descriptions. Please fix.

please have a look I just add the description

garrett:
Thanks for updating it! :sparkles: :+1: :sparkles: 

I'll add @jelly to the review requested area, since he opened the issue (and I don't know about this part of the codebase).

jelly:
Thanks for your PR, I just noticed this is a duplicate of https://github.com/cockpit-project/cockpit/pull/18431 which also tries to fix this issue. That PR has already been reviewed and is a bit further ahead, I'll still review this PR so you can get some feedback but I think we want to continue with https://github.com/cockpit-project/cockpit/pull/18431

test/common/testlib.py
```diff
@@ -2092,6 +2094,34 @@ def enable_root_login(self):
         if not self.machine.ostree_image and self.file_exists(disallowed_conf):
             self.sed_file('/root/d', disallowed_conf)
 
+###########################
+
+
+
+class ExtendedMachineCase(unittest.TestCase):
```
jelly, line 2101: This does not actually extend the MachineCase class, it just inherits the default Python TestCase class. What we actually want to do is add these two functions below to the `MachineCase` class.

test/common/testlib.py
```diff
@@ -2092,6 +2094,34 @@ def enable_root_login(self):
...
+    def reboot(self, timeout_sec=None):
+        self.allow_restart_journal_messages()
+        self.machine.reboot()
+        if timeout_sec is not None:
+            self.machine.wait_for_ssh(timeout=timedelta(seconds=timeout_sec))
```
jelly, line 2106: `wait_for_ssh` does not exist in `cockpit-project/bots`, so this can't work.

test/common/testlib.py
```diff
@@ -2092,6 +2094,34 @@ def enable_root_login(self):
...
+
+
+
+class MyTestCase(ExtendedMachineCase):
+    def test_reboot_with_timeout(self):
+        self.reboot(timeout_sec=30)
+
+    def test_wait_reboot_with_timeout(self):
+        self.wait_reboot(timeout_sec=30)
```
jelly, line 2121: If this worked, the test would reboot the test machine, that's not really what you want I think. It also does not really `assert` anything. So in essence this testcase is not testing anything, if you want to really test if the function rebooted, you could compare for example the `journalctl --list-boots` before and after.

martinpitt: Please drop the whole MyTestCase. Self-tests don't belong into testlib.py (there's a separate place for that, test/verify/check-testlib.py), and this also does not need an explicit test case. Rather just convert the existing test cases to use the new API, and if they pass, that'll be enough proof. E.g. change test/verify/check-system-shutdown-restart to use the new API.

KKoukiou:
Closing as it's duplicate of https://github.com/cockpit-project/cockpit/pull/18431 and inactive at this point.

PR #18434, opened 2023-03-01
Storage: Correctly highlight dangerous menu items in red 

### LINKED ISSUE #18415: storage: Filesystem delete entry in dropdown's should shown as dangerous
### Explain what happens

For example the stratis filesystem delete dropdown or a normal disk's delete dropdown entry is not shown as dangerous as required per [Patternfly guidelines](https://www.patternfly.org/v4/guidelines/colors/#status-and-state-colors).

image

They should be shown as the image below, with red text.

image

## Reproduce

```
./test/image-prepare fedora-37
./bots/vm-run -s cockpit.socket fedora-37
```

* Open a browser http://localhost:9090 login and open the storage page.
* Click on the `/dev/vda2` filesystem 
* Click on the kebab menu (three dots on the right)
* This should open a dropdown menu with a Delete entry which doesn't have a red text.

### Version of Cockpit

main

None

### Server operating system

None

_No response_

_No response_

### System log

_No response_

MahmoudHamdy02:
Closes #18415.

Dangerous menu items, such as 'Delete', were not highlighted in a red color correctly.

I added a 'danger' property to the `tab_menu_danger_actions` array, to identify the items in `StorageMenuItem`.

Here is a screenshot:
image

I think this makes the separate 'danger' array redundant, since they get merged into one array later. It could simply be one array with a 'danger' property set to false/true. I'm not sure if that would break the separator line placement, but I think placing it afterwards when the array is full won't be an issue.

Please let me know what changes I should make to this PR, I'm open to trying a different approach if needed. Thanks!

### DIFF
--- a/pkg/storaged/storage-controls.jsx
+++ b/pkg/storaged/storage-controls.jsx
@@ -215,8 +215,8 @@ export const StorageUsageBar = ({ stats, critical, block, offset, total, small }
         </div>);
 };
 
-export const StorageMenuItem = ({ onClick, onlyNarrow, children }) => (
-    <DropdownItem className={onlyNarrow ? "show-only-when-narrow" : null}
+export const StorageMenuItem = ({ onClick, onlyNarrow, danger, children }) => (
+    <DropdownItem className={(onlyNarrow ? "show-only-when-narrow" : null) + " " + (danger ? "delete-resource-red" : null)}
                   onKeyPress={checked(onClick)}
                   onClick={checked(onClick)}>
         {children}
--- a/pkg/storaged/storage-controls.jsx
+++ b/pkg/storaged/storage-controls.jsx
@@ -216,7 +216,7 @@ export const StorageUsageBar = ({ stats, critical, block, offset, total, small }
 };
 
 export const StorageMenuItem = ({ onClick, onlyNarrow, danger, children }) => (
-    <DropdownItem className={(onlyNarrow ? "show-only-when-narrow" : null) + " " + (danger ? "delete-resource-red" : null)}
+    <DropdownItem className={(onlyNarrow ? "show-only-when-narrow" : "") + (danger ? " delete-resource-red" : "")}
                   onKeyPress={checked(onClick)}
                   onClick={checked(onClick)}>
         {children}
--- a/pkg/storaged/content-views.jsx
+++ b/pkg/storaged/content-views.jsx
@@ -147,7 +147,7 @@ function create_tabs(client, target, is_partition, is_extended) {
     }
 
     function add_menu_danger_action(title, func) {
-        tab_menu_danger_actions.push({ title, func });
+        tab_menu_danger_actions.push({ title, func, danger: true });
     }
 
     const tabs = [];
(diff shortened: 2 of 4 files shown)


garrett (changes requested):
Why isn't this using PatternFly's `isDanger` menu attribute?

https://www.patternfly.org/v4/components/menu/#danger-menu-item

(We shouldn't be using custom code for things PatternFly provides for us. The CSS instantly stood out to me; we used to have to do things like this a long time ago, but PF added an `isDanger` attribute for us, at our request, a while back.)

KKoukiou:
> Why isn't this using PatternFly's `isDanger` menu attribute?
> 
> https://www.patternfly.org/v4/components/menu/#danger-menu-item
> 
> (We shouldn't be using custom code for things PatternFly provides for us. The CSS instantly stood out to me; we used to have to do things like this a long time ago, but PF added an `isDanger` attribute for us, at our request, a while back.)

Because this uses DropdownItem and not the Menu. This is the only available approach with this component.

KKoukiou:
@MahmoudHamdy02 can you please remove the merge commit?

pkg/storaged/storage-controls.jsx
```diff
@@ -215,8 +215,8 @@ export const StorageUsageBar = ({ stats, critical, block, offset, total, small }
         </div>);
 };
 
-export const StorageMenuItem = ({ onClick, onlyNarrow, children }) => (
-    <DropdownItem className={onlyNarrow ? "show-only-when-narrow" : null}
+export const StorageMenuItem = ({ onClick, onlyNarrow, danger, children }) => (
+    <DropdownItem className={(onlyNarrow ? "show-only-when-narrow" : null) + " " + (danger ? "delete-resource-red" : null)}
```
KKoukiou, line 219: Because javascript is weird this is not correct entiretly :)

KKoukiou: `"" + null` will result to "null"

For this reason you need to replace null with '' and  you can also remove the middle + " " + spacer, but just adding a space in the " delete-resource-red"

MahmoudHamdy02: Hello @KKoukiou!

I've removed the merge commit and changed the null condition, let me know if I should make any more changes, thanks :)

KKoukiou (commented):
Just realized we have missed one more occurance of the 'Delete' button;
These also need the new danger property;

pkg/storaged/stratis-details.jsx:        menuitems.push(<StorageMenuItem key="del" onClick={delete_fsys}>{_("Delete")}</StorageMenuItem>);

KKoukiou:
@garrett should the 'Format' action also be red? @MahmoudHamdy02 < let's take the designers opinion.

MahmoudHamdy02:
> pkg/storaged/stratis-details.jsx: menuitems.push({_("Delete")});

After some googling I found that stratis is a storage management library, but it's only for Arch and Fedora. I'm using Ubuntu 22.04. Is there a way to reach the stratis page without having the package installed? Thanks :P

MahmoudHamdy02:
Hi @KKoukiou 

I had some trouble running the test VM through cockpit, so I installed a full fedora 37 through Ubuntu's KVM.

I fixed the delete button color in the `stratis-details` page, let me know if I need to make more changes!

garrett:
> should the 'Format' action also be red? @MahmoudHamdy02 < let's take the designers opinion.

Thanks for asking! Formatting is "dangerous" as data loss will definitely happen, so: Yes.

KKoukiou (approved):
Commit nees squashing on merge. Otherwise all great. thanks!

martinpitt:
Thank you @MahmoudHamdy02 !

PR #18436, opened 2023-03-02
Esbuild pre-requisites

KKoukiou:


martinpitt (approved):

martinpitt (approved):

KKoukiou:
@martinpitt there is something wrong with two files and css https://cockpit-logs.us-east-1.linodeobjects.com/pull-18436-20230302-151917-395b1c7d-fedora-37/log.html

KKoukiou:
@martinpitt tests look now good. Can you please take onother look :) ?

pkg/systemd/overview.scss
```diff
@@ -1,3 +1,4 @@
+@use "page";
```
martinpitt, line 1: OOI, does that make a difference? the next line imports system-global.scss, which imports page.scss as the first thing. I'm worried about duplicating CSS with such duplicate imports.

martinpitt (commented):
Nice work figuring out the specificity bug! I have one question. I retriggered the queued tests which apparently got stuck.

pkg/systemd/overview.scss
KKoukiou: Oh probably it's not needed. Good catch.

martinpitt: I'm fine with leaving it in if the generated .css is the same in both cases. In other words, if dart-sass compiler de-duplicates imports.

martinpitt (dismissed):

martinpitt (approved):

martinpitt:
This is one of these "impossible to get green" PRs. I went through the failures, and I'm fairly convinced that they are unrelated and random. Most failures are only in the third retry, or due to some unexpected messages in the pybridge (which we already know about).

PR #18437, opened 2023-03-02
fmf: Enable ABRT tests on Fedora

martinpitt:
Run the two most useful/important ones, as they are quite expensive.

---

See https://bugzilla.redhat.com/show_bug.cgi?id=2172891#c7

 - [x] Land https://github.com/cockpit-project/bots/pull/4483 and re-run rawhide TMT test after that

### DIFF
--- a/test/browser/browser.sh
+++ b/test/browser/browser.sh
@@ -37,9 +37,12 @@ if rpm -q setroubleshoot-server; then
     dnf remove -y --noautoremove setroubleshoot-server
 fi
 
-if grep -q 'ID=.*fedora' /etc/os-release; then
-    # required by TestLogin.testBasic, but tcsh is not available in CentOS/RHEL
+if grep -q 'ID=.*fedora' /etc/os-release && [ "$PLAN" = "basic" ]; then
+    # Fedora-only packages which are not available in CentOS/RHEL
+    # required by TestLogin.testBasic
     dnf install -y tcsh
+    # required by TestJournal.testAbrt*
+    dnf install -y abrt abrt-addon-ccpp reportd libreport-plugin-bugzilla libreport-fedora
 fi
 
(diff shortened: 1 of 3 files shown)


martinpitt:
[Noooo](https://artifacts.dev.testing-farm.io/25f6a8f5-9da5-4981-b4cf-ba2a898bc8da/).. The journal looks okay, all services are running. For $deity's sake, this worked locally in `tmt run`.. Adding another debug commit.

martinpitt:
Aha! I *can* reproduce this locally. Something resets `/proc/sys/kernel/core_pattern` in between installing the packages + restarting systemd-sysctl, and starting the test. After the test failed the first time, core_pattern was back to `core`.

I wonder if that's somehow related to 
```
Mar 03 04:53:55 default-0 kernel: Unsafe core_pattern used with fs.suid_dumpable=2.
                                  Pipe handler or fully qualified core dump path required.
                                  Set kernel.core_pattern before fs.suid_dumpable.
```

Updated again, it now works locally in a `tmt run` without messing with it interactively.

martinpitt:
OK, this works now -- they both [passed on F37](https://artifacts.dev.testing-farm.io/29329487-59b5-4d43-ba60-9da57062af84/). On [F38](https://artifacts.dev.testing-farm.io/4eeca331-1934-40df-9052-70bd98c40678/) testAbrtReport failed due to https://bugzilla.redhat.com/show_bug.cgi?id=2172891#c7 and was caught by the naughty; on rawhide it [failed](https://artifacts.dev.testing-farm.io/f60ae1a8-ce74-4499-a67d-bba3f545f4c1/) as we are missing the naughty there; I sent https://github.com/cockpit-project/bots/pull/4483 to fix that.

mvollmer (approved):
Thank you!

PR #18447, opened 2023-03-03
Move from Webpack to Esbuild

KKoukiou:


pkg/build
```diff
@@ -22,43 +22,26 @@ WEBPACK_PACKAGES = \
 	tuned \
 	users \
 	$(NULL)
```
martinpitt, line 24: @KKoukiou : Wrt the "mega-glob": If you have trouble doing this with `$(wildcard)`, starter-kit and friends have done this with a simple `find` for many years without any trouble: https://github.com/cockpit-project/starter-kit/blob/main/Makefile#L83  This is a good way to start, and can be optimized later if anyone cares.

pkg/lib/cockpit-components-table.scss
```diff
@@ -1,6 +1,4 @@
 @import "global-variables";
-@import "@patternfly/patternfly/components/Table/table.scss";
-@import "@patternfly/patternfly/components/Table/table-grid.scss";
```
martinpitt, line 3: It's not clear to me how the "null loader" is related to the scss includes here. Could you please explain this a little better?

pkg/lib/esbuild-eslint-plugin.js
```diff
@@ -6,12 +6,14 @@ import { ESLint } from 'eslint';
 
 const NAME = 'eslintPlugin';
 
-export const eslintPlugin = {
+export const eslintPlugin = ({
+    filter = /src\/.*\.(jsx?|js?)$/,
```
martinpitt, line 10: Don't we need `(src|pkg)`?

packit.yaml
```diff
@@ -7,9 +7,9 @@ actions:
     - cp tools/cockpit.spec .
 
   create-archive:
-    # The sandcastle doesn't have enough ram to run webpack, so wait
-    # until the webpack-jumpstart workflow has run and grab the result.
-    - tools/webpack-jumpstart --wait --rebase
+    # The sandcastle doesn't have enough ram to run esbuild, so wait
+    # until the build-jumpstart workflow has run and grab the result.
+    - tools/build-jumpstart --wait --rebase
```
martinpitt, line 12: You said with esbuild, building all webpacks just takes a few seconds.. We can then just drop this. The sandcastle container is no more, and RAM limits are not an issue. I suppose we can drop the whole jumpstart machinery completely, but fine to do this as a follow-up.

tools/esbuild-make
```diff
@@ -0,0 +1,39 @@
...
+if args.watch:
+    os.environ["ESBUILD_WATCH"] = "1"
+if args.no_lint:
+    os.environ["LINT"] = "0"
+if args.rsync:
+    os.environ["RSYNC"] = args.rsync
+if args.prefix:
+    os.environ["ONLYDIR"] = args.prefix
+
+subprocess.check_call(["node", "build.js"])
```
martinpitt, line 39: It seems to me that we should just add the arg parsing to build.js directly, and then our entry point just becomes `./build.js`. Let's clean up all these wrappers. Again, this can be a follow-up.

martinpitt (commented):
Nice! I'm excited to see some progress here. Happy to help with the Debian copyright generation etc.

pkg/lib/cockpit-components-table.scss
KKoukiou: oh right. https://github.com/cockpit-project/cockpit/blob/main/webpack.config.js#L492..L504 

table, toolbar and page css files were passed to null-loader before, therefore not auto imported when these react components were used. No I changed the approach and I don't use null-loader., but I fix the breakpoints with inline replacement of the problematic numbers. So we don't need to manually import these scss file for Table, Tooltip and Page anymore.

pkg/lib/esbuild-eslint-plugin.js
KKoukiou: will do! This is overwritable so I was just passing the filter variable again when this was called from this repository. But your suggestion also works.

packit.yaml
KKoukiou: Right, I thought followup is better as this jumpstart script is not affected at all by the build changes

tools/esbuild-make
KKoukiou: I am fine with this, just thought that people have gotten used to the scripts being in ./tools/.  Plus there are still one there, like the jumpstart.

pkg/lib/esbuild-eslint-plugin.js
martinpitt: Hmm, could this be flipped around? Can we ignore node_modules/ and include all other *.js and *.jsx files?

martinpitt:
- [ ] i386 build

I tried to additionally install `esbuild-wasm`, for building on non-x86_64 platforms. That would be good enough -- slower, but at least working (and surely still a lot faster than with webpack). But alas, the node_modules/esbuild/bin/esbuild wrapper is too stupid to recognize the wasm build as a fallback :exploding_head: 

What kind of works is `ESBUILD_BINARY_PATH=node_modules/esbuild-wasm/bin/esbuild tools/esbuild-make` (in the unit-tests:i386 container). But it fails with tons of errors like

    ✘ [ERROR] Cannot read directory "pkg/lib/plot.js": Invalid argument

i.e. the wasm version is broken. Also, I don't quite know where to put this option -- it belongs into build.js, but setting `process.env` there is too late.

----

 - [ ] With the normal linux-64 version, it doesn't run any browser unit test, as it only builds dist/static, nothing else. I see

    cd . && NODE_ENV='production' tools/termschutz tools/esbuild-make static

but I'm not entirely sure where the "static" actually comes from.

----

 - [x] After that it "just" fails with three eslint errors.

----

 - [x]  when I run `tools/esbuild-make` locally, it fails with several errors:

```
✘ [ERROR] Could not resolve "raw-loader!./get-timesync-backend.py
    pkg/lib/serverTime.js:36:36:
      36 │ import get_timesync_backend_py from "raw-loader!./get-timesync-backend.py";

 [ERROR] ENOENT: no such file or directory, copyfile '/var/home/martin/upstream/cockpit/main/pkg/apps/index.html' -> '/var/home/martin/upstream/cockpit/main/dist/apps/index.html' [plugin plugin:copy]
```

I rebased this PR to fix the conflict, and added three FIXUPs for some of the above issues. Let's update this comment as we add more fixes.

martinpitt:
Hmm, I made matters worse with the [wasm attempt](https://github.com/cockpit-project/cockpit/commit/44c38b5cdec0271284222de09f58b52bd89b4eb3), one just can't update node_modules twice in the same PR? But even now everything goes red. I don't understand all these [useless errors](https://github.com/cockpit-project/cockpit/actions/runs/4423392489/jobs/7756085675)..

martinpitt:
I sense some impatience in the #cockpit channel, so here's the first rebased version which isn't completely broken at least. However, this still needs a lot of cleanups, and some split-outs.

martinpitt:
@KKoukiou : [this run](https://github.com/cockpit-project/cockpit/actions/runs/4481477852/jobs/7878238263?pr=18447) still ran out of memory, it needs to be split further? But cool to see that it worked in [this i386 run](https://github.com/cockpit-project/cockpit/actions/runs/4481477852/jobs/7878239074?pr=18447)! So it seems it's on the edge, sometimes works and sometimes doesn't.

The [64 bit check run](https://github.com/cockpit-project/cockpit/actions/runs/4481477852/jobs/7878239498?pr=18447) just failed on some little eslint issues, I'll push a fixup.

martinpitt:
@KKoukiou : I added a fixup to unbreak the login page; but I force-pushed again to fix the conflicts for the recently landed PRs. I cleared up the obsolete big comments/reviews, and added a new TODO list to the description.

martinpitt:
Yay, there's a new esbuild: https://github.com/evanw/esbuild/releases/tag/v0.17.13

I confirmed that wasm can now build this just fine, without the mega-commit https://github.com/cockpit-project/cockpit/pull/18447/commits/6a6d298d4f39accf2c3b81c3be1a18d82b20e3f4 (saving the link here in case we need it again)

@KKoukiou : I squashed this again, as it needed a force-push anyway. Looks *much* cleaner now!

KKoukiou:
> Yay, there's a new esbuild: https://github.com/evanw/esbuild/releases/tag/v0.17.13
> 
> I confirmed that wasm can now build this just fine, without the mega-commit [6a6d298](https://github.com/cockpit-project/cockpit/commit/6a6d298d4f39accf2c3b81c3be1a18d82b20e3f4) (saving the link here in case we need it again)
> 
> @KKoukiou : I squashed this again, as it needed a force-push anyway. Looks _much_ cleaner now!

Heh good timing, I see https://github.com/cockpit-project/cockpit/pull/18447/commits/8b2b5d18d21110d76c910bf30a69a71a358e155b was kept although we almost pushed the same time! Thanks :)

martinpitt:
@KKoukiou heh, yes. sorry for the unannounced force-push, I thought I'd be safe on a Saturday :grin: Looking forward to the results, now we shouldn't get 220 affected retries any more.

KKoukiou:
all the rest pixel tests failures from [This run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18447-20230325-100457-8b2b5d18-fedora-37/log.html)  don't look like regressions, rather are expected changes because of the breakpoints redefinition approach change.

martinpitt:
@KKoukiou Amazing! :100: So let's fix that trivial eslint error, land https://github.com/cockpit-project/cockpit-podman/pull/1243 , I'll adjust it for cockpit and add it here, we update pixel refs, and :rocket: ? Anything else?

martinpitt:
I implemented a proper watch mode like in https:[//github.com/cockpit-project/cockpit-podman/pull/1243 . Pushing it as a [separate commit](https://github.com/cockpit-project/cockpit/pull/18447/commits/f92ea3cb1aeb02c45fee2adbb326c2e033d25d4c) for review and history, but I'll squash it now, update pixel tests, and drop no-test. This will be the first push which has a chance to actually land.

martinpitt:
[something wrong with cockpit.locale()](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18447-20230326-231128-9b1cc4b2-fedora-37-pybridge/log.html), curiously only on the python bridge. Investigating..

build.js
```diff
@@ -1,88 +1,264 @@
...
+    outbase: './pkg',
+    outdir: "./dist",
+    target: ['es2020'],
+};
+
+// context options for qunit tests in qunit/
+const qunitOptions = {
+    bundle: true,
+    // show "build started/finished" messages in watch mode
+    logLevel: 'info',
```
martinpitt, line 42: This is a no-op, can be dropped.

build.js
```diff
@@ -1,88 +1,264 @@
...
+            // ignore errors in watch mode
+        }
+
+        if (args.watch) {
+            const on_change = async path => {
+                console.log("change detected:", path);
+                await Promise.all([pkgContext.cancel(), qunitContext.cancel()]);
+                try {
+                    await Promise.all([pkgContext.rebuild(), qunitContext.rebuild()]);
+                } catch (e) { console.warn(e) } // ignore in watch mode
```
martinpitt, line 259: Drop the `warn(e)`, debugging leftover.

martinpitt (changes requested):
I went through this again with my reviewer goggles on, and found a few small cleanups. But before I push again, I want to see a fully green test set.

build.js
```diff
@@ -1,88 +1,271 @@
...
+    },
+};
+
 const parser = (await import('argparse')).default.ArgumentParser();
 parser.add_argument('-c', '--config', { help: "Path to webpack.config.js", default: "webpack.config.js" });
 parser.add_argument('-r', '--rsync', { help: "rsync bundles to ssh target after build", metavar: "HOST" });
 parser.add_argument('-w', '--watch', { action: 'store_true', help: "Enable watch mode" });
-parser.add_argument('-e', '--no-eslint', { action: 'store_true', help: "Disable eslint linting" });
-parser.add_argument('-s', '--no-stylelint', { action: 'store_true', help: "Disable stylelint linting" });
+parser.add_argument('-e', '--no-eslint', { action: 'store_true', help: "Disable eslint linting", default: production });
```
KKoukiou, line 56: So when it's a production build the lint options are by default disabled now. So in CI we rely on 'test/static-code' for linting? This latter does not have stylelint.

KKoukiou: Right - that run for the devel package which is enough.

martinpitt (approved):
I'm happy enough with this now, but of course I messed around with this heavily. Anything from your side still, @KKoukiou ?

@mvollmer , do you want to take a quick look?

build.js
```diff
@@ -1,88 +1,269 @@
...
+    nodePaths,
+    outbase: './pkg',
+    outdir: "./qunit",
+    loader: {
+        ".sh": "text",
+    },
+};
+
 const parser = (await import('argparse')).default.ArgumentParser();
 parser.add_argument('-c', '--config', { help: "Path to webpack.config.js", default: "webpack.config.js" });
```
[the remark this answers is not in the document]
martinpitt: Let's wait for @mvollmer 's review, as this is otherwise green, and rather "expensive" (pixel tests, node_modules rebuild). I want to do a follow-up anyway to simplify `tests` in files.js, happy to clean it up then. Of course, when Marius finds other issues, we can fix this right away.

files.js
```diff
@@ -0,0 +1,194 @@
...
+        "systemd/overview.jsx",
+        "systemd/terminal.jsx",
+        "systemd/hwinfo.jsx",
+
+        "packagekit/updates.jsx",
+        "users/users.js",
+    ],
+
+    tests: [
+        "base1/test-base64",
```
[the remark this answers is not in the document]
martinpitt: .. except that node.js does not have an API for globbing, meh..

PR #18455, opened 2023-03-06
test: Copy SELinux attributes when encrypting the rootfs

mvollmer:
This allows us to get rid of the autorelabel and the need to switch off SELinux during relabeling (which was new for Fedora 38).


martinpitt (approved):
Oh, this is nice!

approve --selinux

PR #18457, opened 2023-03-06
Fix mobile padding for Applications and Kernel dump pages & make firewall header not sticky

### LINKED ISSUE #18230: mobile: Missing padding on pages: Applications, Kdump, Development (regression)
Left and right padding is missing in mobile on the following pages:

In the header:

- Applications
- Kernel crash dump

On the whole page:

- Development (this page doesn't matter too much, as nobody but us devs see it)

MahmoudHamdy02:
Closes #18230.

1- Removed a property in the firewall header that made it sticky.
2- Commented out some CSS lines that removed the padding for mobile.

The pages look good in the browser (mobile view) and on my phone. I also took a quick look through the rest of the website and it doesn't look like any padding is broken.

Let me know if I need to make any changes, thanks!


pkg/lib/patternfly/patternfly-4-overrides.scss
```diff
@@ -391,12 +391,12 @@ svg {
...
-    padding-left: 0;
-    padding-right: 0;
-  }
-}
+// @media screen and (max-width: $pf-global--breakpoint--sm) {
+//   .pf-c-page__main > section.pf-c-page__main-section:not(.pf-m-padding) {
+//     padding-left: 0;
+//     padding-right: 0;
+//   }
+// }
```
martinpitt, line 399: This was done deliberately in commit 3d876b610105fb, i.e. it's not an accidental bug. Let's please first discuss this in https://github.com/cockpit-project/cockpit/issues/18230#issuecomment-1457671305 .

@MahmoudHamdy02 : In the meantime: Please don't comment out things, just delete them. Also, commit messages should say the "where" (in this case, start Subject with "patternfly:") and the "why?", i.e. explain the reason for a change. It doesn't need to restate the "what". As the "why" isn't obvious at all here, I suggest to just remove the commit for now.

pkg/networkmanager/firewall.jsx
```diff
@@ -1039,7 +1039,7 @@ export class Firewall extends React.Component {
         const enabled = this.state.firewall.enabled;
 
         return (
-            <Page groupProps={{ sticky: 'top' }}
+            <Page
```
martinpitt, line 1042: This commit should again start with a "where?", i.e. "networkmanager: Make firewall header not sticky". However, this again needs a "why", and it introduces an inconsistency: A lot of other page headers are sticky as well:
```
pkg/apps/application.jsx:        <Page groupProps={{ sticky: 'top' }}
pkg/networkmanager/firewall.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/networkmanager/network-interface.jsx:        <Page groupProps={{ sticky: 'top' }}
pkg/storaged/details.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/systemd/hwinfo.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/systemd/logDetails.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/systemd/services/service.jsx:                <Page groupProps={{ sticky: 'top' }}
pkg/users/account-details.js:        <Page groupProps={{ sticky: 'top' }}
```
and then there are others which aren't:
```
pkg/apps/application-list.jsx:        <Page id="list-page">
pkg/storaged/overview.jsx:        <Page id="main-storage">
```

Can you please discuss this with @garrett first? This should be done consistently on all pages. Perhaps this should also only be dropped on mobile? Having the breadcrumbs visible is useful, but indeed it's quite expensive in terms of screen size usage.

martinpitt (changes requested):
Thank you!

pkg/networkmanager/firewall.jsx
MahmoudHamdy02: Hello @martinpitt

Thanks for the recommendations! :) 
When the design is decided I will edit the commits/messages and add the new changes.

garrett (changes requested):
This is not the correct approach.

The other pages are correct, and removing (or commenting out, which should normally be a removal) would break the other pages. This would also impact all other Cockpit projects, such as cockpit-podman, cockpit-machines, cockpit-ostree, etc. as they import the overrides file as well.

It's likely that the JSX in these two pages (Applications & Kernel dump) would need to change to match the other pages.

pkg/networkmanager/firewall.jsx
garrett: Sticky can be problematic when it gets too large. Sticky is useful when there's some critical information or action at the top and it doesn't need much space. Sticky is probably not really useful just for breadcrumbs.

Additionally, sticky might or might not work based on other elements, and some pages may want to use sticky for other parts.

Changing sticky should be done in another PR, not this one. It's not related to fixing the padding. It was mentioned in the same issue, but as a question later. But it's a different type of change, so it should really be in a different PR.

Sometimes, we do have some somewhat unrelated changes in a PR as a separate commit in the same PR, like here, but it's usually for an umbrella of changes, and those PRs often take longer to merge. (In these cases, it's usually X, Y, Z commits for a PR on page A. So while X, Y, and Z changes might not be related to one another, they're still all related to A.)

But, in this PR as-is, it changes code that affects all of Cockpit (even projects) instead of fixing Applications and Kernel dump specifically. And then it has a sticky removal for Firewall which is not even closely related to Applications or Kernel dump or padding at all, so these two concepts really should be separate PRs.

## Summary

- Fix the problem on the problem pages, do not change something that affects everything (breaking everything else).
- Drop the sticky change here. Move it to another PR. We can talk about what pages should or should not have sticky and what parts over there on that other PR and adjust it accordingly.

## Hint

See this line that you commented out? `.pf-c-page__main > section.pf-c-page__main-section:not(.pf-m-padding)` — that `:not()` means that it won't apply when there's `pf-m-padding`. Sections don't have padding by default (as the widgets within sometimes may provide that). These two pages probably had padding as an accident before, but they don't now because they don't have widgets within that also provide padding _and_ these pages also didn't set padding.

So that means the fix would be to actually use widgets that have padding inside these headings _or_ to set padding. Look at what other pages (that don't have this problem) are doing.

MahmoudHamdy02:
Hello @garrett!

I've reverted the old changes, and only changed the relevant JSX files as you mentioned.

I couldn't find the JSX for the 'development' page though, I think it only uses `html` and `js`. Let me know if I've missed something.

I've tried to use better/correct commit messages as well.

Thanks for all the help!

martinpitt:
@MahmoudHamdy02 Thanks! Wrt. the commit messages, it'd be better to rebase and squash most or all commits together with "f"ixup. As long as the whole PR can become just one commit, we can do this on landing with the GitHub "squash & merge" button. But if you want to e.g. keep the kdump and apps commits separate, it needs to be done on the actual branch. We don't want to have "revert", "do it the other way" etc. commits on main.

Also, don't worry about the playground page. There is just about zero design effort in them, they are mostly used in integration tests for covering the technical foundation of cockpit.

I also adjusted the title, as the "not sticky" part is now in PR #18476 .

The net diff here looks reasonable to me, but I'll let @garrett have the last word here. Thank you!

garrett (approved):
Thanks, @MahmoudHamdy02! It looks good to me. :+1:

pkg/kdump/kdump-view.jsx
```diff
@@ -480,7 +480,7 @@ export class KdumpPage extends React.Component {
         }
         return (
             <Page>
-                <PageSection variant={PageSectionVariants.light}>
+                <PageSection variant={PageSectionVariants.light} className="pf-m-padding">
```
KKoukiou, line 483: That works but it's kinda hacky. The correct way it to make it clear, that's a header for the page. You can do that like

```
g[kkoukiou@sequioa cockpit]$ git diff
diff --git pkg/kdump/kdump-view.jsx pkg/kdump/kdump-view.jsx
index b65548919..621238835 100644
--- pkg/kdump/kdump-view.jsx
+++ pkg/kdump/kdump-view.jsx
@@ -479,15 +479,18 @@ export class KdumpPage extends React.Component {
                 </Tooltip>);
         }
         return (
-            <Page>
-                <PageSection variant={PageSectionVariants.light}>
-                    <Flex alignItems={{ default: 'alignItemsCenter' }}>
-                        <Title headingLevel="h2" size="3xl">
-                            {_("Kernel crash dump")}
-                        </Title>
-                        {kdumpSwitch}
-                    </Flex>
-                </PageSection>
+            <Page
+                isBreadcrumbGrouped
+                additionalGroupedContent={
+                    <PageSection variant={PageSectionVariants.light}>
+                        <Flex alignItems={{ default: 'alignItemsCenter' }}>
+                            <Title headingLevel="h2" size="3xl">
+                                {_("Kernel crash dump")}
+                            </Title>
+                            {kdumpSwitch}
+                        </Flex>
+                    </PageSection>
+                }>
                 <PageSection>
                     <Card>
                         <CardBody>

```


Please do the same for the kdump page.

KKoukiou (changes requested):

KKoukiou (commented):
Also please remove all other wrong commits and have a clear PR with your next push.

MahmoudHamdy02:
@KKoukiou I've implemented the method you mentioned and cleaned up the unnecessary commits, hopefully should be good now!

KKoukiou:
Closing because of inactivity. Just reopen if you work on this still.

PR #18461, opened 2023-03-07
metrics: paint graph red when it reaches more than 0.8 load

KKoukiou:


garrett (changes requested):
Looks good!

The only thing I'd change is perhaps changing the suffix `-red` to something else. Perhaps `-high` or something else similar?

(Classes generally should not be named after an appearance.)

pkg/metrics/metrics.jsx
```diff
@@ -842,24 +843,57 @@
...
+        });
+    }
+    const dataPointsRed = key => (
+        Math.min(LOAD_THRESHOLD, dataFiltered[firstRed][key]).toString() + "," + firstRed + " " +
+        dataFiltered.map((samples, index) => samples[key] > LOAD_THRESHOLD ? samples[key].toString() + "," + index.toString() : "").join(" ") + " " +
+        Math.min(LOAD_THRESHOLD, dataFiltered[firstRed][key]).toString() + "," + lastRed // close polygon
+    );
+
+    if (isOverThreshold && minute == 12)
+        console.info(minute, dataPointsRed(key));
```
cockpituous, line 879: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18461-20230308-172459-d7e58a47-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -842,24 +843,57 @@
...
             />
             { have_sat && <polygon
                 transform={ "matrix(1,0,0,-1,1," + SVG_YMAX + ")" }
-                points={ dataPoints("sat_" + resource) }
+                points={ dataPoints(key) }
                 opacity="0.7"
             /> }
+            { isOverThreshold && <polygon
+                 className="graph-over"
+                 transform={ have_sat ? "matrix(-1,0,0,-1,1," + SVG_YMAX + ")" : "matrix(-2,0,0,-1,2," + SVG_YMAX + ")" }
```
cockpituous, line 894: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18461-20230308-172459-d7e58a47-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/metrics/metrics.jsx
```diff
@@ -842,25 +843,62 @@ class CurrentMetrics extends React.Component {
...
+    );
+};
+
+const Polygons = ({ data, type, resource, have_sat }) => {
+    const transform = type == "sat" ? "1,0,0,-1,1," : have_sat ? "-1,0,0,-1,1," : "-2,0,0,-1,2,";
+    const key = type + "_" + resource;
+    const dataFiltered = data.filter(d => d && typeof d[key] === 'number');
+
+    if (!dataFiltered.length)
+        return null;
```
cockpituous, line 861: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18461-20230309-121036-88868592-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

garrett (approved):
Looks great! :+1: Let's merge!

For reference, this what I'm seeing — it looks like what I expect:

image

I like the `-over` suffix too. (Much better than `-red`.)

pkg/metrics/metrics.jsx
```diff
@@ -842,25 +843,62 @@ class CurrentMetrics extends React.Component {
...
+
+    const isOverThreshold = dataFiltered.some(sample => sample[key] >= LOAD_THRESHOLD);
     const dataPoints = key => (
         "0,0 " + // start polygon at (0, 0)
-        data.map((samples, index) => (samples && typeof samples[key] === 'number') ? samples[key].toString() + "," + index.toString() : "").join(" ") +
-        " 0," + (data.length - 1) // close polygon
+        dataFiltered.map((samples, index) => Math.min(LOAD_THRESHOLD, samples[key]).toString() + "," + index.toString()).join(" ") +
```
martinpitt, line 866: In both you and Garrett's screenshots you notice that the normal part of the graph has a slope (lines between actual data points), but now they have a crease at the 0.8 critical point. This makes them look rather weird.

I.e. it's necessary to compute a proper trapezoid/polygon for both sides which keep the slope, instead of just "inventing" a data point at LOAD_THRESHOLD.

pkg/metrics/metrics.scss
```diff
@@ -498,13 +498,18 @@ html {
       --color: lightsalmon;
     }
 
+    .graph-over {
+      fill: red;
```
martinpitt, line 502: I find this really hurtful to the eyes. Could we change this to be a little less intense?

martinpitt (changes requested):
I must say, I don't like the looks of this at all. It looks spiky, odd, and hard to look at. We already have events for the over-usage, so it also feels redundant.

If we really need this, I'd actually prefer making the graph background light-red during the times of over-use. This would also be a lot cheaper to compute/render, and also avoid these creases.

That said, as you and Garrett already agreed on this, I'm fine with being outvoted. If you want to keep it, then please just fix the first three issues (that should be relatively easy), and we land this approach.

Thanks!

martinpitt:
Another proposal: Don't split a minute into two polygons, but change their color to something like 30% more saturation or luminosity, to make them stand out. This also avoids the crazy computation and creases. Of course this needs some experimentation with which color pairs look good.

garrett:
Another proposal: Use gradient backgrounds. It can be a sharp transition. Then we don't have to compute anything.

garrett (changes requested):
Martin makes some good points.

It is somewhat redundant (with the other PR) and is expensive to compute.

We should keep the PR open as we figure this out.

garrett:
As discussed on IRC, I suggested we completely drop SVGs altogether and use full-width elements with background color, clipped by CSS `clip-path: polygon(var(--points))` and specify the points via object properties like `style="--points: x y, x2 y2, x3 y3;"` — then CSS can handle the fill. If we want to have warning, it can be in the CSS as something like `background: linear-gradient(to var(--direction), var(--color) 0, var(--color) 80%, var(--color-extreme) 80%)`. (Or we could have a gap and make it like 85% or 90% or something and have it blend from one color to another instead of a sharp transition.)

Browser engines are heavily optimized for drawing rectangles. SVGs are much more complex; they're a different object (in this case embedded) and may contain all sorts of different types of objects. Switching to clip-path means we're back to rendering rectangles _only_, except parts are shown while other parts hidden (as specified by the geometry of clip-path polygon) and that should also be accelerated by the browser's compositor. Therefore, it _should_ _**theoretically**_ be much faster as the browser doesn't have to spin-up another type of rendering mode hundreds of times. But we'll see in practice.

I should've thought of this before... but I was only thinking about drawing shapes in solid colors prior to now, and SVG is an obvious choice for that. But clip-path is another idea well-suited for this task, and it would let us do a warning in a much better way.

Also, using `clip-path` polygons means we just need to worry about points relative in percentages and not have to do any calculations.  So I think it would allow us to simplify our code and CPU to process the metrics into visuals as well?

For porting, we could probably even use `clip-path: path()` which takes SVG path coordinates, so it's likely a drop-in port to start with. Ideally, we'd probably want to switch to `polygon` though, for simplicity, because we only need straight lines and not curves... and it might be faster as a result. (Unknown about that.)

Resources:

- https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path
- https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/polygon
- https://css-tricks.com/clipping-masking-css/#aa-the-new-clip-path

martinpitt:
@garrett I really like this proposal! :100:

We have 12 data points per minute (or more when zooming out). For rendering I suggest to draw their maximum instead of their average. That will "inflate"/exaggerate the graph, but IMHO that's more interesting for debugging performance problems and spotting spikes than spikes getting drowned by averaging.

KKoukiou:
@garrett can you please check the color-extreme variables? Widths changed in the dev tools to see the gradients till the end.
Screen Shot 2023-03-15 at 00 31 26

KKoukiou:
Closing as this contains more discussion which is confusing at this point. The whole graphs are now implemented with clip-paths and it was suggested use gradients. Using gradients is at this point a 2-3 lines of CSS changes which will require @garrett to choose the right colors. I don't see how I can contribute to that much.

PR #18469, opened 2023-03-08
test: Post fewer coverage comments

mvollmer:


mvollmer:
I expect one (1) comment from the bots about "foo = -1" not being executed by any test.

The details should show more added and not executed lines, but they should not be reported.

pkg/systemd/overview.jsx
```diff
@@ -60,10 +60,26 @@ class OverviewPage extends React.Component {
         this.superuser = cockpit.dbus(null, { bus: "internal" }).proxy("cockpit.Superuser", "/superuser");
     }
 
+    a_new_function(flag) {
+        let foo = 0;
+
+        if (flag) {
+            foo += 1;
+        } else
+            foo = -1; // will not be executed, and we want a report about this
```
cockpituous, line 69: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18469-20230308-121813-a074a05e-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

test/common/lcov.py
```diff
@@ -392,6 +392,17 @@ def get_review_comments(diff_info_file):
...
     cur_line = None
 
+    def is_interesting_line(text):
+        # Don't complain when being told to shut up
+        if "// not-covered: " in text:
+            return False
+        # Don't complain about lines that contain only punctuation, or
+        # nothing but "else".  We don't seem to get reliable
+        # information for them.
+        if not re.search('[a-zA-Z0-9]', text.replace("else", "")):
```
martinpitt, line 402: I'm a bit concerned about accidentally ignoring lines that have "else" as a sub-word, like "panelselector" or so, but I think it's okay to ignore that :grin:

martinpitt (commented):
Your demo commit worked perfectly! LGTM, thanks for adding this! Then we can start annotating the "should not happen" error handlers.

test/common/lcov.py
mvollmer: No, this will remove "else" from a line,. and will ignore it if there are no letters or digits left.  So, it wont ignore "panelselector" since "panlector" still has some letters. But it will ignore lines like "}" and "} else {".

martinpitt: Ah, right!

martinpitt:
@mvollmer : I see you triggered fcos, but remember to actually remove the demo commit first

martinpitt (approved):
Looking forward to using this!

PR #18234, opened 2023-01-24
users: Allow specifying home directory and shell during account creation

skobyda:
This add ability to specify custom home directory and custom shell for newly created users:

Screenshot from 2023-01-26 11-38-24

[Screencast from 2023-01-26 11-41-05.webm](https://user-images.githubusercontent.com/42733240/214817073-addde521-f013-4eb7-a4b0-54fb7fa5edcc.webm)

If user tries to use home directory which already exists:
Screenshot from 2023-01-26 14-36-46

I also added the listing of those user's home directory and shell into user's details page:
Screenshot from 2023-01-26 11-43-37

PR #18470, opened 2023-03-08
users: Show shell and home directory in user's details page

skobyda:
Split from https://github.com/cockpit-project/cockpit/pull/18234 as recommended

# Accounts: Show shell and home directory

The account's page now shows user's home directory and default shell.

223722047-2885b1d2-2bbb-4694-8135-398d51a79387

### DIFF
--- a/pkg/users/account-details.js
+++ b/pkg/users/account-details.js
@@ -342,6 +342,12 @@ export function AccountDetails({ accounts, groups, shadow, current_user, user })
                                     </div>
                                 </FormGroup>
                                 }
+                                { account.home && <FormGroup fieldId="account-home-dir" hasNoPaddingTop label={_("Home directory")}>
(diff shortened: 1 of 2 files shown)


martinpitt (changes requested):
Can you please add a corresponding integration test? Also, does this need a pixel ref update?

martinpitt (changes requested):
see above

skobyda:
In the end, I added getUserAddDetails helper function to tests which parses defaults of useradd, since that's the approach which I plan to use futher in https://github.com/cockpit-project/cockpit/pull/18234

skobyda:
Tests seems pretty green, apart from some flakes. I think the PR is ready

martinpitt (approved):
LGTM, thanks! Still fighting a bit with that i386 unit-test, previous two attempts failed with a podman error.

PR #18473, opened 2023-03-08
test: Install dnf-automatic into TMT images

mvollmer:
These images seem to have contained it until recently, but now we need to ask for it explicitly.

### DIFF
--- a/test/browser/main.fmf
+++ b/test/browser/main.fmf
@@ -8,6 +8,7 @@
     - python3
     - libvirt-python3
     # required by tests
+    - dnf-automatic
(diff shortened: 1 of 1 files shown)


mvollmer:
All the TF runs seem to be failing because of this, for example https://artifacts.dev.testing-farm.io/6beb9bff-5607-49c1-93b2-0cae9cd2e732/.

martinpitt (approved):
Hm, I haven't seen that in other PRs, but if it helps, sure! Tests are green here. I triggered the required fcos test.

PR #18476, opened 2023-03-08
patternfly: Adjust sticky property in header for specific pages

MahmoudHamdy02:
Originally from #18457, but moving this into a separate PR to continue the discussion.

I marked this as a draft PR as we haven't decided which pages will change yet.

### DIFF
--- a/pkg/networkmanager/firewall.jsx
+++ b/pkg/networkmanager/firewall.jsx
@@ -36,7 +36,7 @@ import { Stack } from "@patternfly/react-core/dist/esm/layouts/Stack/index.js";
 import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput/index.js";
 import { Title } from "@patternfly/react-core/dist/esm/components/Title/index.js";
 import { Toolbar, ToolbarContent, ToolbarItem } from "@patternfly/react-core/dist/esm/components/Toolbar/index.js";
-import { Page, PageSection, PageSectionVariants } from "@patternfly/react-core/dist/esm/components/Page/index.js";
+import { Page, PageBreadcrumb, PageGroup, PageSection, PageSectionVariants } from "@patternfly/react-core/dist/esm/components/Page/index.js";
 import { Modal } from "@patternfly/react-core/dist/esm/components/Modal/index.js";
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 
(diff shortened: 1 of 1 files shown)


KKoukiou:
@garrett so we want all headers in pages that have currently page title or breakcrumb to not be sticky? 
That's these:

````
[kkoukiou@sequioa cockpit]$ git grep "sticky: 'top'"
pkg/apps/application.jsx:        <Page groupProps={{ sticky: 'top' }}
pkg/networkmanager/firewall.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/networkmanager/network-interface.jsx:        <Page groupProps={{ sticky: 'top' }}
pkg/storaged/details.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/systemd/hwinfo.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/systemd/logDetails.jsx:            <Page groupProps={{ sticky: 'top' }}
pkg/systemd/services/service.jsx:                <Page groupProps={{ sticky: 'top' }}
pkg/users/account-details.js:        <Page groupProps={{ sticky: 'top' }}
```

garrett:
Summary: Let's not remove the sticky from the breadcrumbs. But let's remove it from the Firewall page.

Is it possible to have the breadcrumbs be sticky on their own without the heading being sticky? (And without having the shadow if the page is at the top, as there shouldn't be a shadow between the breadcrumbs and heading card.)

garrett (changes requested):
This is probably what we want... it's definitely better to not have the whole Firewall heading sticky... but if we could somehow have the breadcrumb sticky like the rest of the pages (reasoning stated above, after evaluating everything), then that would be great. If that's a problem, then please state so and we could then merge this as-is, as it's still better than what we have (since the whole heading is so huge).

MahmoudHamdy02:
Hello @garrett, this is what I have reached so far, let me know if this is the behavior you meant.

1.
image

2.
image

KKoukiou:
@garrett I honestly don't like the idea of the 'Firewall switcher' section going above the breadcrumbs just for even this short scroll. breadcrumbs are sticky and should always be visible. I would prefer to just remove the border from the breadcrumb section - that as you explain looks weird.

garrett:
I just thought of an alternate approach without the bugs and one that also doesn't need JS: Remove the dropshadow, but add it back on a pseudoelement that has a lower z-index. Then use a gradient mask that hides it when fully up and fades it in once the page is scrolled down (hence a gradient for opacity. This would be the most automatic way to handle it in pure CSS.

At this point, I think I would need to consider it as a PF override in a separate PR that we can propose to be upstreamed.

An additional (and related) issue is that there's double padding with the breadcrumb and header page area. This should be solved in the same area in the overrides file. That fix is a lot easier, with something like `breadcrumb + header { margin-top: none; }`. It's not implemented for scrollables as the shadow would get in the way... but with the shadow solved, then we can fix it globally too.

KKoukiou:
This is now obsolete.

PR #18479, opened 2023-03-09
doc: drop host option from our documentation

jelly:
We want to move away from running commands on a different connected host from JavaScript. In the future we will completely block the ability to specify a different host then the current iframe's host.


martinpitt (approved):
Dankjewel!

PR #18480, opened 2023-03-10
test: Disable preloading for release builds as well

martinpitt:
In TestHostSwitching.testBasic, the auxiliary machines often fail to load software updates. This breaks the pixel test, as there is an additional icon in the main menu which also changes the column sizes.

Change setup_provisioned_hosts() to always respect the `disable_preload` argument. If necessary, we can do the `self.is_devel_build` in the caller, but it's not really relevant: We already have several explicit integration tests to check the preload functionality (e.g. TestUpdates.testBasic).

The recent commit 7663244c8c2e also overzealously moved a comment. Move it back, and adjust it for the updated situation above.

---

[example 1](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18466-20230309-140241-2a80128f-fedora-37/log.html), [example 2](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18478-20230309-181427-20d2e50a-fedora-37-bots-4498/log.html)

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -2098,16 +2098,11 @@ class MachineCase(unittest.TestCase):
         """Setup provisioned hosts for testing
 
         This sets the hostname of all machines to the name given in the
-        provision dictionary and optionally disabled preload when it runs
-        on a development build.
+        provision dictionary and optionally disabled preload.
         """
-
         for name, m in self.machines.items():
             m.execute(f"hostnamectl set-hostname {name}")
-            # Disable preloading on all machines ("machine1" is done in testlib.py)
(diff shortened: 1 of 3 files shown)


martinpitt:
Nice, works! I'm also looking into [this common flake](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18466-20230309-140243-2a80128f-centos-8-stream-pybridge/log.html#185-2). With these rapid login/logout cycles, the UI sometimes hangs. It appears this is because of too many pending queries to packagekit and systemd, whose replies then take a long time to come back. I can reproduce this reasonably well locally, and it seems to be really robust (10 passes in a row) without preloads. So let's fix this as well.

jelly (approved):

Issue #18459, opened 2023-03-07
hacking: Improve Debian/Ubuntu development docs


garrett:
In the HACKING document, Debian and Ubuntu are mentioned for installing nodejs. However, the rest of the document is Fedora-specific.

We probably want to fix this.

I've seen that at least a few GSOC participants have been running into problems.

Alcadeus0:
Can i take up this issue :)

jelly:
Feel free!

garrett:
Note: The HACKING file is in the main repo, but it's regularly mirrored into the website. So edit the original file in the repo; don't worry about the version on the website.

https://github.com/cockpit-project/cockpit/blob/main/HACKING.md

martinpitt:
The current [HACKING](https://github.com/cockpit-project/cockpit/blob/main/HACKING.md) recommends using the cockpit/tasks container in toolbox, instead of installing development dependencies directly on your machine. This works on all distributions, is easier, cleaner, and more reliable.

PR #18484, opened 2023-03-10
Updated Hacking.MD

Alcadeus0:
Fix for #18459


HACKING.md
```diff
@@ -17,6 +17,30 @@ On Debian/Ubuntu:
 
     $ sudo apt install nodejs
 
+
+For running the browser unit tests, the following dependencies are required:
+
+    $ sudo apt-get install chromium-headless dbus
```
jelly, line 23: How did you get chromium-headless installed on Ubuntu 22.04 it does not exists:

```
[jelle@⬢ ubuntu-toolbox-22.04][~]%sudo apt-get install chromium-headless dbus
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package chromium-headless
```

Alcadeus0: Ah forgot to change it, we can run chromium using the --headless flag. Should I add that in or just remove it?

HACKING.md
```diff
@@ -17,6 +17,26 @@ On Debian/Ubuntu:
 
     $ sudo apt install nodejs
 
+
+Creating VM images locally (not necessary for running tests) needs the
```
KKoukiou, line 21: This is true, but it's quite out of place here. We have a separate repository which contains the image creation scripts https://github.com/cockpit-project/bots/

I would prefer the instruction on how to use these scripts  out of this document. Additionally building images locally is something we do seldom, as we have github actions that do it for us on a weekly basis.

HACKING.md
```diff
@@ -17,6 +17,26 @@ On Debian/Ubuntu:
...
+following:
+    
+    $ sudo apt-get install virtinst
+
+For running integration tests, the following dependencies are required:
```
KKoukiou, line 26: This list already exists, https://github.com/cockpit-project/cockpit/blame/main/HACKING.md#L41

Adjust there is some package is missing.

HACKING.md
```diff
@@ -17,6 +17,26 @@ On Debian/Ubuntu:
...
+
+    $ sudo apt-get install curl expect xz-utils rpm libvirt-daemon-system libvirt-daemon-driver-storage-core \
+        libvirt-daemon-driver-qemu libvirt-clients python3-libvirt
+        python3-pyyaml python3-flake8 \chromium-browser
+
+Updating the `node_modules` (in case you need to modify `package.json`)
+requires npm to be installed.
+
+    $ sudo apt-get install npm
```
KKoukiou, line 35: npm is a dependent package of nodejs which is already mentioned as dependency on the top of the document  https://github.com/cockpit-project/cockpit/blame/main/HACKING.md#L18

HACKING.md
```diff
@@ -17,6 +17,26 @@ On Debian/Ubuntu:
...
+
+Additionally to build cockpit from the `autogen.sh` command below, head to /tools/debian/control
```
KKoukiou, line 37: There are commands for both rpm and apt based distros to install build deps. For fedora it would be `sudo dnf builddep cockpit`

It's already mentioned here https://github.com/cockpit-project/cockpit/blame/main/HACKING.md#L35

You can exend that section to include the apt build-dep command.

KKoukiou (changes requested):
I have a few comments. Also please remove the configure~ backup file from the commit.

HACKING.md
martinpitt: This is also not true any more. `tools/node-modules install` runs in a podman container, not with local `npm`.

HACKING.md
Alcadeus0: I see, I added it because there's the command for installing virt-install for fedora in the document. Do we plan to remove that as well or keep it?

HACKING.md
Alcadeus0: I see, I'll have it removed from the document

HACKING.md
Alcadeus0: Should I include just an alternative apt-get command for debian then? Or just omit it out

HACKING.md
Alcadeus0: the `sudo apt-get builddep cockpit` does not work when i run it. What could be the issue?

HACKING.md
martinpitt: An `apt install` list would be nice indeed.

HACKING.md
martinpitt: That depends on what exactly it ways when you run it. A common cause is that you don't have any `deb-src` apt sources.

HACKING.md
Alcadeus0: An `apt install` list for running the integration tests right?

HACKING.md
Alcadeus0: Im trying to run it from the root of the repo, it just shows that its unable to find a source package for tools/cockpit.spec

Alcadeus0: Alright i got it running by enabling deb-src, should i mention about it?

martinpitt:
The current [HACKING](https://github.com/cockpit-project/cockpit/blob/main/HACKING.md) recommends using the cockpit/tasks container in toolbox, instead of installing development dependencies directly on your machine. This works on all distributions, is easier, cleaner, and more reliable.

This PR now has conflicts. If you still want to go ahead with this, please move this into the "Manually installing the development dependencies" section. Otherwise, let's perhaps just close this?

PR #18489, opened 2023-03-12
users: Automatically validate account expiry date on change

### LINKED ISSUE #17815: Account expiration
### Explain what happens

When inserting garbage in the date field, the validation is not reset once you have entered a correct date.

[Screencast from 13-10-22 14:22:09.webm](https://user-images.githubusercontent.com/67428/195595031-ffd31d89-d925-47ed-a25b-9201aee725d7.webm)

### Version of Cockpit

_No response_

_No response_

### Server operating system

_No response_

_No response_

_No response_

### System log

_No response_

MahmoudHamdy02:
Closes #17815.

Simple fix for the validation problem. Now the `validation()` function runs every time the date input changes, and outputs the correct message. 

I'm not sure if calling `validate()` and `update()` after each other is the most optimized solution, so let me know what you think about this.

Out of curiosity, why is entering a past date allowed? Setting the account to expire in 2021 doesn't make much sense, unless I'm missing something.

Thanks!


marusak:
@jelly can you please take a look?

KKoukiou (commented):
That's not nice behavior. Wit this change when the user tries to fill in a valid date they will see 'Invalid expiration date' message before they finish writing.


One approach would be to debounce the validate function in order to introduce some delay. And also validate first time on blur. If the validatation already happened, then typing should always call the debounced validate function.

jelly:
> That's not nice behavior. Wit this change when the user tries to fill in a valid date they will see 'Invalid expiration date' message before they finish writing.
> 
> One approach would be to debounce the validate function in order to introduce some delay. And also validate first time on blur. If the validatation already happened, then typing should always call the debounced validate function.

I think ideally it would use `blur` for validation and also reset when you start typing. See for example: https://www.patternfly.org/v4/components/form/react/validated/

MahmoudHamdy02:
I've just pushed an initial implementation of the debounce approach. The 'validating' message can be easily removed if it's not needed. Let me know if this isn't the best behavior!

MahmoudHamdy02:
I have updated the code to match the two points you listed. Let me know if this needs more changes!

In `onBlur`, I call both `validate` and `update`, because `validate` on its own changes the text but doesn't refresh the message in cockpit.

PR #18308, opened 2023-02-08
login: Change key verify command background

garrett:
Make the background transparent, so it works in light and dark mode.

Fixes #18307

Issue #18307, opened 2023-02-08
Unreadable command in dark mode


marusak:
### Explain what happens

There is something actually written in the white box :)

![Screenshot from 2023-02-08 14-16-57](https://user-images.githubusercontent.com/12330670/217540904-716894f0-a3be-411d-87fc-8781aa06e2be.png)

garrett:
Oh wow. That's an oversight. Looking into this now.

PR #18491, opened 2023-03-12
shell/scss: Fix dark theme code background for new ssh host modal

MahmoudHamdy02:
When adding a new SSH host for the first time, cockpit shows a modal with a line of code to verify the fingerprint. The background color in dark theme makes it almost impossible to read.

I added a condition for dark theme in the `scss`, let me know if this needs any changes!

## Before:

image

## After:

image

### DIFF
--- a/pkg/shell/hosts_dialog.jsx
+++ b/pkg/shell/hosts_dialog.jsx
@@ -31,6 +31,7 @@ import PropTypes from 'prop-types';
 import { Alert } from "@patternfly/react-core/dist/esm/components/Alert/index.js";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
 import { Checkbox } from "@patternfly/react-core/dist/esm/components/Checkbox/index.js";
+import { ClipboardCopy } from "@patternfly/react-core/dist/esm/components/ClipboardCopy/index.js";
 import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form/index.js";
 import { Modal } from "@patternfly/react-core/dist/esm/components/Modal/index.js";
 import { Radio } from "@patternfly/react-core/dist/esm/components/Radio/index.js";
@@ -463,10 +464,10 @@ class HostKey extends React.Component {
             body = <>
                 <Alert variant='danger' isInline title={_("Changed keys are often the result of an operating system reinstallation. However, an unexpected change may indicate a third-party attempt to intercept your connection.")} />
                 <p>{_("To ensure that your connection is not intercepted by a malicious third-party, please verify the host key fingerprint:")}</p>
-                <pre className="hostkey-fingerprint">{fp}</pre>
+                <ClipboardCopy isReadOnly hoverTip={_("Copy")} clickTip={_("Copied")} className="hostkey-fingerprint pf-u-font-family-monospace">{fp}</ClipboardCopy>
                 <p className="hostkey-type">({key_type})</p>
                 <p>{cockpit.format(_("To verify a fingerprint, run the following on $0 while physically sitting at the machine or through a trusted network:"), this.props.host)}</p>
-                <pre className="hostkey-verify-help-cmds">ssh-keyscan -t {key_type} localhost | ssh-keygen -lf -</pre>
+                <ClipboardCopy isReadOnly hoverTip={_("Copy")} clickTip={_("Copied")} className="hostkey-verify-help-cmds pf-u-font-family-monospace">ssh-keyscan -t {key_type} localhost | ssh-keygen -lf -</ClipboardCopy>
                 <p>{_("The resulting fingerprint is fine to share via public methods, including email.")}</p>
                 <p>{_("If the fingerprint matches, click 'Accept key and connect'. Otherwise, do not connect and contact your administrator.")}</p>
             </>;
@@ -474,10 +475,10 @@ class HostKey extends React.Component {
             body = <>
                 <p>{cockpit.format(_("You are connecting to $0 for the first time."), this.props.host)}</p>
                 <p>{_("To ensure that your connection is not intercepted by a malicious third-party, please verify the host key fingerprint:")}</p>
-                <pre className="hostkey-fingerprint">{fp}</pre>
+                <ClipboardCopy isReadOnly hoverTip={_("Copy")} clickTip={_("Copied")} className="hostkey-fingerprint pf-u-font-family-monospace">{fp}</ClipboardCopy>
                 <p className="hostkey-type">({key_type})</p>
                 <p>{cockpit.format(_("To verify a fingerprint, run the following on $0 while physically sitting at the machine or through a trusted network:"), this.props.host)}</p>
-                <pre className="hostkey-verify-help-cmds">ssh-keyscan -t {key_type} localhost | ssh-keygen -lf -</pre>
+                <ClipboardCopy isReadOnly hoverTip={_("Copy")} clickTip={_("Copied")} className="hostkey-verify-help-cmds pf-u-font-family-monospace">ssh-keyscan -t {key_type} localhost | ssh-keygen -lf -</ClipboardCopy>
                 <p>{_("The resulting fingerprint is fine to share via public methods, including email.")}</p>
                 <p>{_("If the fingerprint matches, click 'Accept key and connect'. Otherwise, do not connect and contact your administrator.")}</p>
             </>;
--- a/pkg/shell/shell.scss
+++ b/pkg/shell/shell.scss
@@ -23,6 +23,7 @@
 @use "table";
 @use "nav";
 @import "@patternfly/patternfly/components/Select/select.css";
+@import "@patternfly/patternfly/utilities/Text/text.css";
 
 :root {
   --ct-color-host-accent: var(--pf-global--active-color--100);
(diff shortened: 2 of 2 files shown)


marusak:
I wonder, wasn't this already fixed in https://github.com/cockpit-project/cockpit/pull/18308 ?
Pretty sure this is the same cause as https://github.com/cockpit-project/cockpit/pull/18307

garrett:
@marusak: I thought so too... but that was for the login page (which doesn't actually use PatternFly; it just looks like it), this is for when you're logged into Cockpit (which actually uses PatternFly).

However, this is a 1-off fix. It is probably a wider spread problem with code blocks in dark mode and should be investigated and fixed at a higher level (probably PF overrides) if this is the case.

KKoukiou (changes requested):
@MahmoudHamdy02 @marusak @garrett I think the correct way is to start using https://www.patternfly.org/v4/components/code-block/ instead of the pre. THe codeblock component has a black theme variant.

KKoukiou:
@MahmoudHamdy02 is this clear what is suggested here?

MahmoudHamdy02:
@KKoukiou Yup! I've converted one `<pre>` into PatternFly's `<CodeBlock>`. It looks correct in both light & dark modes automatically. Take a look and tell me what you think.

I've only kept the border radius from the old scss. Let me know what else should change in the design.

I will change the other occurrences of `<pre>` once the PF looks good

KKoukiou (changes requested):

pkg/shell/shell.scss
```diff
@@ -237,14 +237,6 @@ $desktop: $phone + 1px;
 
 .hostkey-verify-help-cmds {
   border-radius: 0.25rem;
```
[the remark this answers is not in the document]
KKoukiou: And please squash the commits.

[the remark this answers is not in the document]
MahmoudHamdy02: I've removed the old scss and replaced all the `<pre>`s with `<CodeBlocks>`. All commits should also be squashed now. 

However, I'm not sure where the MotD page is within cockpit, so I couldn't check if it works as it should. Let me know your thoughts.

pkg/shell/hosts_dialog.jsx
```diff
@@ -463,21 +464,29 @@ class HostKey extends React.Component {
             body = <>
                 <Alert variant='danger' isInline title={_("Changed keys are often the result of an operating system reinstallation. However, an unexpected change may indicate a third-party attempt to intercept your connection.")} />
                 <p>{_("To ensure that your connection is not intercepted by a malicious third-party, please verify the host key fingerprint:")}</p>
-                <pre className="hostkey-fingerprint">{fp}</pre>
```
KKoukiou, line 466: You removed this and some other classes, but I see these are used in the tests. 

test/verify/check-shell-multi-machine:                self.assertEqual(b.text('#hostkey-fingerprint'), fp)
test/verify/check-shell-multi-machine:                self.assertTrue(b.val('#hostkey-fingerprint'))


I suggest you keep the classes for the tests to work. 

If these is CSS that you want to drop thats' assosiated with the class, we need to check each case, and drop it if no longer relevant and adjust the classname like here, so that's affecting only the non-changed files:

pkg/static/login.css:#hostkey-fingerprint {
pkg/static/login.html:          <pre id="hostkey-fingerprint"></pre>

KKoukiou (changes requested):

pkg/systemd/overview-cards/motdCard.jsx
```diff
@@ -117,6 +118,9 @@ export const MotdCard = () => {
     </>;
 
     return <Alert id="motd-box" isInline variant="default" className="motd-box"
-                  title={<pre id="motd">{motdText}</pre>}
+                  title={
```
KKoukiou, line 121: same here, don't remove id, as we use it for the tests.

pkg/systemd/overview-cards/motdCard.scss
```diff
@@ -1,14 +1,3 @@
-#motd {
-  --motd--font-size: var(--pf-global--FontSize--sm);
-  background-color: transparent;
```
KKoukiou, line 3: This rule was actually intended

You can see the motd in http://localhost:9090/system after you add some  content in /etc/motd file.

KKoukiou: I just texted a bit interactively. THe idea is that you want o keep the looks of it exactly how it was. This heavily regressed. 
Please check this by yourself, if you manage to adjust the CSS to achieve the expected looks/ 

I can suggest that the only that can be removed is font-size setting as codeblock default font size is also small.

KKoukiou (changes requested):

pkg/shell/hosts_dialog.jsx
```diff
@@ -38,6 +38,7 @@ import { Stack } from "@patternfly/react-core/dist/esm/layouts/Stack/index.js";
 import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput/index.js";
 
 import { ModalError } from "cockpit-components-inline-notification.jsx";
+import { CodeBlock, CodeBlockCode } from "@patternfly/react-core";
```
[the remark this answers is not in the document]
MahmoudHamdy02: Will look out for this in the future :+1:

pkg/systemd/overview-cards/motdCard.scss
MahmoudHamdy02: When using `<CodeBlockCode>` directly, it looks identical to the old `<pre>`, let me know if this is OK.

MahmoudHamdy02:
I've kept the class names. As for motd, when using `<CodeBlockCode>` directly, it looks identical to the old `<pre>`. Let me know if this is OK!

pkg/shell/shell.scss
```diff
@@ -223,26 +223,10 @@ $desktop: $phone + 1px;
   }
 }
 
-.hostkey-fingerprint {
-  font-size: large;
```
KKoukiou, line 227: There is definitely some UI change here, so I will ask @garrett for review. Personally I think it's fine to drop the override. 

previous:
Screenshot from 2023-03-21 10-42-28

now:
Screenshot from 2023-03-21 10-50-42

KKoukiou (changes requested):

KKoukiou (approved):
From my side approved. Inline in the comments there are before after screenshots, for the relevant parts. Will need final ack from @garrett who is the designer.

garrett:
So this is definitely an improvement. However, it would ideally also have a copy icon for the command someone is supposed to run.

Code block is specifically for multi-line code blocks. Right at the top of the page, the PatternFly website states "_A code block is a component that contains 2 or more lines of read-only code. The code in a code block can be copied to the clipboard_." @ https://www.patternfly.org/v4/components/code-block

It's _probably_ better to use a read-only copy block with the PatternFly utility class `pf-u-font-family-monospace`

- https://www.patternfly.org/v4/components/clipboard-copy
- https://www.patternfly.org/v4/utilities/text/

image

And here's what that would look like with monospace (I did an override with `font-family: var(--pf-global--FontFamily--redhat--monospace);`):

image

garrett (changes requested):
Additionally, this should not be a danger alert, but a warning, as something COULD be wrong, but something isn't NECESSARILY wrong.

Here's the original mockup for it... Please note it was made for the login page and therefore isn't a modal.

login-key-warnings

Here's what everything else is cropped:

image

It would need more work to adapt it into what it has grown into (with the how-to-verify block (which is behind a popover in this mockup).

garrett:
So to make it clear:

I think we should address the main issue with the entry (no dark mode) by using a read-only copy block with the PatternFly utility class `pf-u-font-family-monospace`. It should be for the entry only, not the hash. (Neither are actually multiple lines of code, so they should not use the CodeBlock widget.)

(The rest was to show that there are some other problems with this modal, but this issue and PR are more focused at fixing the dark mode, which is fine. We can deal with other issues in follow-ups.)

KKoukiou (commented):
@MahmoudHamdy02 how about one more iteration where you change everything to use what @garrett suggested here? https://github.com/cockpit-project/cockpit/pull/18491#issuecomment-1477721135

it indeed looks much better like this.
Note: everything but for the motd.

garrett:
What about the utility class to make the font monospaced though?

Add `pf-u-font-family-monospace` to the `className` in React.

https://www.patternfly.org/v4/utilities/text

You'd need to make sure the text utility CSS is imported near the top (at the imports section) of the toplevel file, however, if it isn't already, like this:

```scss
@import "@patternfly/patternfly/utilities/Text/text.scss";
```

It's currently _not_ in `shell.scss`, so it should be included under the other import (for select).

Importing a utility means that you don't have to write custom CSS for it and can therefore depend on PatternFly instead. And it should be in the toplevel as we should only import things once per page (so sub-pages that import things can cause things to be imported multiple times, so it's best to always check the toplevel SCSS file for each page of Cockpit. In this case, the "page" is the special shell one.)

MahmoudHamdy02:
OK, switched to the utility class instead of the CSS, let me know if it's good now

KKoukiou (approved):

KKoukiou:
@MahmoudHamdy02 removed the motd change from this PR - as it was changing the UI and our designer is on PTO so he can't approve it today.

KKoukiou (approved):

PR #18498, opened 2023-03-13
node_modules: update chrome-remote-interface to 0.32.1

jelly:
Update to 0.32.1 for Chromium 111 compatibility. We need https://github.com/cyrus-and/chrome-remote-interface/commit/58ce8a3ad541709ab3883bc379e4a1dad5acb997 so the tests work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "argparse": "^2.0.1",
     "axe-core": "^3.5.2",
     "babel-loader": "^8.1.0",
-    "chrome-remote-interface": "^0.31.2",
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice! So that also gets rid of that pesky warning then?

I triggered firefox as well. The TF tests use firefox, but better safe than sorry :grin:  Thanks!

jelly:
> Nice! So that also gets rid of that pesky warning then?

It should, more important I can run the tests again :)
 
> I triggered firefox as well. The TF tests use firefox, but better safe than sorry grin Thanks!

Sure!

PR #18501, opened 2023-03-14
systemd/services: Fix initial loading of timers

martinpitt:
`this.timers` is indexed by object path, not unit ID. Fix the initialization in listUnits(). This regression was introduced in commit 0a19a39e9523 during the many rewrites.

This didn't get caught by integration tests as they only covered the `PropertiesChanged` update path, not the initial load. Add such a test, using the existing `deleteme.timer`.

----

release-blocker as this is a regression.

### DIFF
--- a/pkg/systemd/services/services.jsx
+++ b/pkg/systemd/services/services.jsx
@@ -485,7 +485,7 @@ class ServicesPageBody extends React.Component {
                         if (Id.endsWith(".timer")) {
                             timerPromises.push(dbus.call(ObjectPath, s_bus.I_PROPS, "GetAll",
                                                          [s_bus.I_TIMER]));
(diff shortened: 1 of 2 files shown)


martinpitt:
Got several failures [like this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18501-20230314-080254-854735bb-debian-stable/log.html#329), because the timer already fired. Let's not assume the "last run" then.

martinpitt:
Meh, [still fails on arch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18501-20230314-140324-36662bec-arch/log.html#332), taking a look.

KKoukiou (approved):

PR #18502, opened 2023-03-14
systemd/services: Lazily load timer information

martinpitt:
Take timer loading out of the critical path in listUnits(). We don't need them for the initial page load. Instead, load them if/when actually showing the Timers tab, and update the page once they are all loaded.

 - [x] Builds on top of PR #18501

https://issues.redhat.com/browse/COCKPIT-915


jelly (approved):
Great! I didn't realise we loaded so much on the services page load

PR #18505, opened 2023-03-14
base1: Drop cockpit.dbus.publish() and .meta()

martinpitt:
Nothing in cockpit or any related project publishes a D-Bus interface. This is only exercised in the unit tests. The Python bridge does not implement this either.

It's hard to imagine a good use case for calling .meta() without publishing an API. In theory, it could be used for calling D-Bus methods on services which don't have introspection data by themselves, but (1) this is exceedingly rare (not observed anywhere), and (2) better handled by explicitly giving type signatures with .call().

# API: Drop cockpit.dbus.publish() and .meta()

The `cockpit.js` library dropped these two functions, which were meant to create a D-Bus server object from inside a Cockpit page. No [known Cockpit project](https://cockpit-project.org/applications.html) has ever used this, and we are not aware of a good use case. If you are using this API, please [let us know](https://github.com/cockpit-project/cockpit/issues)!


martinpitt:
I wonder if we could actually just drop this from cockpit.js as well?

martinpitt:
I went all the way and dropped it. We'll have to deal with this sooner or later anyway (hello py bridge). If this fails review on "too bold", I'll go back to dropping the documentation only; otherwise I'll add a release note.

martinpitt:
> I went all the way and dropped it. 

That is, from doc and cockpit.js. I didn't drop it from the C bridge, not worth it IMHO.

allisonkarlitskaya (commented):
Only one question: what's the story with polkit?  If we ever want to do that properly, it might make sense to implement an agent from Javascript.

martinpitt:
I'm afraid I can't speak to this -- I just know that we haven't actually done this so far. If we ever do, we'll need to teach the py bridge about exporting objects (i.e. provide a D-Bus server). Doesn't seem unreasonable to me.

It's probably better to wait for @mvollmer next week.

mvollmer:
(`.meta` is implemented in the pybridge, no?)

mvollmer:
The only counter argument I can think of is that we might accidentally break existing code. Otherwise this looks like a very reasonable simplification.

If we need to publish something on the bus some day for some specific situation, what about implementing it that as a peer bridge then? Might avoid some round trips, even.

mvollmer (approved):

martinpitt:
@mvollmer Yes, I agree -- pretty much the only use case that we found so far is to implement a polkit responder -- but that could just be implemented in cockpit-bridge (py version) itself, no need to go via the page JS. Thanks for checking!

I added a proper release note.

PR #18298, opened 2023-02-07
lib: Support custom time servers with chronyd

mvollmer:
Release notes:

## Overview: Support for additional timeservers with chronyd

Cockpit can now configure additional time servers when chronyd is used to synchronize the clock. Previously, this was only supported with systemd-timesyncd.

image

PR #18508, opened 2023-03-15
lib: port the current manual NTP server dialog to PF

### LINKED ISSUE #18496: overview: NTP server needs PatternFly inputs and padding
image

1. Entries aren't using the correct element (they're not PatternFly styled)
2. There's absolutely no padding; there needs to be
3. I didn't even get asked about the UI pattern. It might need to be something else... I haven't had time to think about this.

_Originally posted by @garrett in https://github.com/cockpit-project/cockpit/issues/18298#issuecomment-1462320320_

jelly:
Use Patternfly elements for input elements and buttons, in a later follow up look into using FormFieldGroup's.

Closes #18496

image

### DIFF
--- a/pkg/lib/serverTime.js
+++ b/pkg/lib/serverTime.js
@@ -20,12 +20,13 @@ import cockpit from "cockpit";
 import React, { useState } from "react";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
 import { DatePicker } from "@patternfly/react-core/dist/esm/components/DatePicker/index.js";
-import { Flex } from "@patternfly/react-core/dist/esm/layouts/Flex/index.js";
+import { Flex, FlexItem } from "@patternfly/react-core/dist/esm/layouts/Flex/index.js";
 import { Form, FormGroup } from "@patternfly/react-core/dist/esm/components/Form/index.js";
 import { Popover } from "@patternfly/react-core/dist/esm/components/Popover/index.js";
 import { Select, SelectOption, SelectVariant } from "@patternfly/react-core/dist/esm/components/Select/index.js";
 import { Spinner } from "@patternfly/react-core/dist/esm/components/Spinner/index.js";
 import { TimePicker } from "@patternfly/react-core/dist/esm/components/TimePicker/index.js";
+import { TextInput } from "@patternfly/react-core/dist/esm/components/TextInput/index.js";
 import { CloseIcon, ExclamationCircleIcon, InfoCircleIcon, PlusIcon } from "@patternfly/react-icons";
 import { show_modal_dialog } from "cockpit-components-dialog.jsx";
 import { useObject, useEvent } from "hooks.js";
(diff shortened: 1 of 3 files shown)


pkg/lib/serverTime.js
```diff
@@ -562,26 +564,17 @@ function ChangeSystimeBody({ state, errors, change }) {
...
-                }
-            </tbody>
-        </table>);
+        custom_ntp.servers.map((s, i) => (
+            <InputGroup className="ntp-server-input-group" key={i}>
+                <TextInput value={s} placeholder={_("NTP server")} aria-label={_("NTP server")}
+                           onChange={(value, event) => change_server(event, i, value)} />
+                <Button variant="secondary" onClick={event => add_server(event, i)}
+                        icon={ <PlusIcon /> } />
+                <Button variant="secondary" onClick={event => remove_server(event, i)}
```
cockpituous, line 573: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18508-20230315-103007-c1b4a202-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

cockpituous, line 573: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18508-20230315-131044-159e27b8-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

garrett:
The action buttons should have space between themselves and the entry. They're not supposed to be connected, as they do not have anything to do with the entry itself. The remove, arguably does a bit (but it's not the default action, so therefore should not be connected), but the plus does not.

garrett (changes requested):
I should've requested changes.

Take them out of the InputGroup; they're separate actions (not linked to inputting the text) and therefore need spacing between each.

Sadly, PatternFly's guidelines are lacking on when to use them. They're surprisingly so sparse here. https://www.patternfly.org/v4/components/input-group/design-guidelines

jelly:
@garrett updated PR now looks like this:

image

pkg/lib/serverTime.js
```diff
@@ -562,26 +563,19 @@ function ChangeSystimeBody({ state, errors, change }) {
...
-        </table>);
+        custom_ntp.servers.map((s, i) => (
+            <Flex className="ntp-server-input-group" spaceItems={{ default: 'spaceItemsSm' }} key={i}>
+                <FlexItem grow={{ default: 'grow' }}>
+                    <TextInput value={s} placeholder={_("NTP server")} aria-label={_("NTP server")}
+                               onChange={(value, event) => change_server(event, i, value)} />
+                </FlexItem>
+                <Button variant="secondary" onClick={event => add_server(event, i)}
+                        icon={ <PlusIcon /> } />
+                <Button variant="secondary" onClick={event => remove_server(event, i)}
```
cockpituous, line 574: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18508-20230320-133044-9f586a3c-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

pkg/lib/serverTime.scss
```diff
@@ -1,3 +1,7 @@
 .ct-serverTime-time-picker {
   max-width: 7rem;
 }
+
+.ntp-server-input-group {
```
garrett, line 5: This should probably be:

```suggestion
.ntp-server-input-group:not(:last-child) {
```

...so it won't add the extra margin to the last item.

garrett (changes requested):
There's additional margin under the last item. (Replied inline.) This affects the way it's rendered.

Additionally, the DOM has an unnecessary wrapper, and the row wrapper is wrapping everything instead of individual rows. This doesn't affect how it looks, but it is incorrect in the code.

Here's what I mean in the rendered DOM:

image

Aside from these two issues, it _looks_ good.

pkg/lib/serverTime.js
cockpituous, line 574: This added line is not executed by any test.  [Details](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18508-20230320-192236-ceaa400f-fedora-37-devel//Coverage/lcov/github-pr.diff.gcov.html)

garrett (approved):
:+1: Thanks! This looks good now!

After merging this, please also take a screenshot and update https://github.com/cockpit-project/cockpit/pull/18298 too, so the release notes will have the fixed version.

pkg/lib/serverTime.js
```diff
@@ -562,26 +563,19 @@ function ChangeSystimeBody({ state, errors, change }) {
...
-                                    isDisabled={i == custom_ntp.servers.length - 1} />
-                        </td>
-                    </tr>))
-                }
-            </tbody>
-        </table>);
+        custom_ntp.servers.map((s, i) => (
+            <Flex className="ntp-server-input-group" spaceItems={{ default: 'spaceItemsSm' }} key={i}>
+                <FlexItem grow={{ default: 'grow' }}>
+                    <TextInput value={s} placeholder={_("NTP server")} aria-label={_("NTP server")}
```
KKoukiou, line 569: placeholders are used wrongly here. (and in many other places in cockpit) 
placeholders are supposed to contain examples of the input, not description of the input. Since it existed here, it's not a regression.

garrett: I agree with this! Thanks for pointing it out. :+1: :100: 

 I talked about this specific issue about the placeholder with @jelly in the call we had (and I think in the issue too). The idea is to handle this problem in a follow-up that redoes the UI anyway. We wanted to get in a fix for it not being PF first (as it's in a screenshot in the release notes and looks _really bad_ otherwise), then fix it properly after.

PR #18509, opened 2023-03-15
hacking: Improve used language

garrett:
I made a few changes in the previous PR and got a little carried away. :wink:

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -35,14 +35,15 @@ Enter the toolbox with
 
     toolbox enter cockpit
 
-Your home directory, user D-Bus, etc. are shared with the host, so you can e.g.
-edit files as you usually do. Building and running tests happens inside the
-toolbox. If desired, you can install additional packages with `sudo dnf install`.
+Your home directory, user D-Bus, etc. are shared with the host, so you can
+edit files as you normally would. Building and running tests happens inside the
+toolbox container. If desired, you can install additional packages with
+`sudo dnf install`.
 
 ## Working on Cockpit's session pages
 
-Most contributors want to work on the web (HTML/JavaScript) parts of Cockpit.
-First install Cockpit on your local machine as described in:
+Most contributors want to work on the web (HTML, JavaScript, CSS) parts of Cockpit.
+First, install Cockpit on your local machine as described in:
 
 <https://cockpit-project.org/running.html>
 
@@ -50,14 +51,15 @@ Cockpit uses a lot of build and runtime dependencies from [npm](https://www.npmj
 
     tools/node-modules checkout
 
-Next run this command from your top level Cockpit checkout directory, and make
+Next, run this command from your top level Cockpit checkout directory, and make
 sure to run it as the same user that you'll use to log into Cockpit below.
 
     mkdir -p ~/.local/share/
     ln -s $(pwd)/dist ~/.local/share/cockpit
 
-This will cause cockpit to read JavaScript and HTML files directly from the
-built package output directory instead of using the installed Cockpit UI files.
+This will cause cockpit to read JavaScript, HTML, and CSS files directly from the
+locally built package output directory instead of using the system-installed Cockpit
+files.
 
 Now you can log into Cockpit on your local Linux machine at the following
 address. Use the same user and password that you used to log into your Linux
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -35,67 +35,72 @@ Enter the toolbox with
...
+After every change to the source files, bundles need to be rebuilt. The
+recommended and fastest way is to do that is using the "watch" mode on 
+the page that you are working on. For example, if you want to work on 
+anything in [pkg/systemd](./pkg/systemd/), run:
 
     tools/webpack-watch systemd
 
-Note that this enables eslint by default -- if you want to disable it, run it
-with `-e`/`--no-eslint`.
+See [pkg/](https://github.com/cockpit-project/cockpit/tree/main/pkg)
```
martinpitt, line 77: Can you please keep the `(./pkg/)` relative link here? Then this will also work on branches and forks.

HACKING.md
```diff
@@ -35,67 +35,72 @@ Enter the toolbox with
...
 SSH `c` alias as described in [test/README.md](./test/README.md), you can use
 one of these commands:
 
     tools/webpack-make.js kdump -r c
     tools/webpack-watch kdump -r c
 
-To make Cockpit again use the installed code, rather than that from your git
-checkout directory, run the following, and log into Cockpit again:
+If you'd like Cockpit to use the system-installed package instead of your
+development checkout, run the following, and then log into Cockpit again:
```
martinpitt, line 96: This lost this "revert previous change" flavor that the previous "again" had. If you don't like that, perhaps "If you like to go back to using the system-installed pages.."?

HACKING.md
```diff
@@ -119,7 +124,7 @@ You can run unit tests of the current checkout:
 
     make check
 
-These should finish very quickly and it is good practice to do it often.
+These should finish very quickly. It is a "good practice" to do this often.
```
martinpitt, line 127: Why the quotes? This isn't meant in any ironic or indirect way.

HACKING.md
```diff
@@ -322,8 +329,8 @@ A local cache is maintained in `~/.cache/cockpit-dev`.
 
 ## Contributing a change
 
-Make a pull request on github.com with your change. All changes get reviewed,
-tested and iterated on before getting into Cockpit. The general workflow is
+Make a pull request on GitHub.com with your change. All changes get reviewed,
```
martinpitt, line 332: Not a strong opinion, but capitalizing domain names is rather unusual. (Feel free to ignore)

martinpitt (changes requested):
Thank you! Most changes are good of course, just some suggestions/corrections.

HACKING.md
garrett: But it wouldn't work on the website, where most people will see this document.

HACKING.md
garrett: It's not meant to be ironic or indirect. "Good practice" is a saying.

We can drop it, though, if people would think it's ironic. It's more correct with the quotes, but if people think it's ironic, that's worse than being proper.

HACKING.md
martinpitt: Not sure about "most people" (this addresses developers, not users). All other links in HACKING.md are relative at the moment, so if that is a problem then I suppose this should be done consistently.

garrett: I think I have functionality to rewrite URLs in the conversion script. We could leave it relative and have the script fix it for the website.

HACKING.md
garrett: I've made a third version. How's this?

https://github.com/cockpit-project/cockpit/pull/18509/files#diff-9769a3f5504af3cd735e08f078b94f827bdd51eacac2e9d80ac9761bc485171fL90-R96

> To make Cockpit use system packages again, instead of your checkout directory, remove the symlink with the following command and log back into Cockpit:

garrett:
@martinpitt: I went with your suggestions for everything except the comment above, where I made a third version. https://github.com/cockpit-project/cockpit/pull/18509#discussion_r1142131081... so please specifically look at that change.

I also have an additional idea of making the container numbered steps instead, with both Fedora/Ubuntu inside the create container step. (This way, it should be easier to follow, and look less complex, as you'd clearly see the steps instead of having the preformatted text jumping out at you. Additionally, we could more easily extend it and add Arch and openSUSE in the install toolbox step.)

HACKING.md
garrett, line 77: Oh, I guess I missed this:

```suggestion
See [pkg/](./pkg/)
```

garrett: I changed this and force-pushed after rebasing, but I don't know why GitHub isn't showing my changes — both here and as a force-push.

garrett: Aha! It just took _several minutes_. (Usually, it's pretty instant.)

martinpitt (approved):
Thank you!

PR #18510, opened 2023-03-16
Various CSS/eslint/build system cleanups

martinpitt:
Broken out from #18447. They are all from @KKoukiou and I approve (can't formally do it). Let's get them out of the way, and test separately.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -237,7 +237,6 @@ process.traceDeprecation = true;
 
 /* These can be overridden, typically from the Makefile.am */
 const srcdir = process.env.SRCDIR || '.';
-const builddir = process.env.BUILDDIR || '.';
 const libdir = path.resolve(srcdir, "pkg" + path.sep + "lib");
 const nodedir = path.relative(process.cwd(), path.resolve(srcdir, "node_modules"));
 const section = process.env.ONLYDIR || null;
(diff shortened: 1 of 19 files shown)


martinpitt:
The [no-webpack-loader syntax commit](https://github.com/cockpit-project/cockpit/actions/runs/4433552854/jobs/7778702102) is incomplete, we still do use that. I updated the commit to actually change the code and update the webpack config.

webpack.config.js
```diff
@@ -380,7 +380,7 @@ export default {
     entry: info.entries,
     // cockpit.js gets included via <script>, everything else should be bundled
     externals: (section === 'kdump/' || section === 'base1/') ? {} : { cockpit: "cockpit" },
-    plugins: plugins,
+    plugins,
```
martinpitt, line 383: Ah, this slipped through -- vim ALE automatically fixes eslint issues. Hopefully acceptable.

pkg/lib/cockpit-components-table.scss
```diff
@@ -1,6 +1,4 @@
 @import "global-variables";
-@import "@patternfly/patternfly/components/Table/table.scss";
-@import "@patternfly/patternfly/components/Table/table-grid.scss";
```
martinpitt, line 3: @KKoukiou Hmm, could it be that this is not actually safe yet with webpack? These were the first commits in your series, and they looked like general fixes. But there are [lots of pixel failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18510-20230316-051941-d8ca716c-fedora-37/log.html), and some of them are really broken, like the [top bar](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18510-20230316-051941-d8ca716c-fedora-37/pixeldiff.html#TestSuperuser-testBasic-topnav-privileged-medium)

martinpitt:
@KKoukiou : Should I just remove the "null loader" commit here, and we apply it as part of the esbuild conversion if it's too hard to fix? Or should this work, and is broken independently?

martinpitt:
I took out that commit, to get the other fixes out of the door. I also want/need to clean up the webpack entries list to make it more regular, as esbuild does not support renaming the entry points. But I think I'll do that in a separate PR.

pkg/lib/cockpit-components-table.scss
KKoukiou: This can be only introduce with esbuild. I kept the commit seperate for code review ease, but eventually they must be squashes with the commit that's porting to esbuild.

martinpitt: Thanks! I took it out.

martinpitt:
Closing this in favor of #18512, it does not really make sense to test/review them separately.

PR #18512, opened 2023-03-16
build system cleanups for esbuild preparation

martinpitt:
Let's stop renaming entry point file names in webpack. It makes them    
unintuitive for developers to find (e.g. "local.js" was the entry point    
for pkg/users), the reference in the HTML did not match, and their    
declaration in webpack.config.js was complicated.    
    
Change the info.entries map to be a simple list, as the source and target names    
now match (except for some .jsx → .js noise).    
    
This requires moving pkg/systemd/services/* up into pkg/systemd/. This was the    
odd one, no other page has entry points in a subdirectory. Eventually we want    
to move this into a completely separate webpack, but for now that will do.    
    
While we are at it, make their names more regular: The main entry point for    
pkg/page/ is now "page.js", and the HTML is "index.html".    
    
This prepares us for moving to esbuild, which does not support entry    
point renaming.    

---

Plus a few other commits broken out of #18447


martinpitt:
This is super hard to get green, due to the 175 (!) affected retries. But works well enough, so ready for review. I have another huge pile (and muuuuch more interesting, promised!) of changes which builds on top of this, so I'd appreciate a review today. Thanks!

KKoukiou (approved):
nice cleanups. Thanks for making the esbuild PR smaller :P

PR #18513, opened 2023-03-16
makefiles: use prefix to allow usage in non fhs distros with less patching

lucasew:
I am the maintainer, and daily driver user, of the cockpit package on NixOS.

Previously I used an approach based on patch files [1]  to fix some details in the code, so it would work with NixOS and some of these details would be convenient to be upstreamed.

The patch approach was replaced by a line substitution approach because it's resilient to line shifts because the patches can misalign and fail to be applied.

This patch basically instructs the makefiles to use the prefix because our builder uses $out (that is /nix/store/something) instead of /.

[1] https://github.com/NixOS/nixpkgs/blob/f7e045e14911989ce2222f3526941e1eb85cd3f0/pkgs/servers/monitoring/cockpit/default.nix

### DIFF
--- a/src/systemd/Makefile.am
+++ b/src/systemd/Makefile.am
@@ -25,10 +25,10 @@ dist_motd_SCRIPTS = src/systemd/update-motd
 
 # Automake: 'Variables using ... ‘sysconf’ ... are installed by install-exec.'
 install-exec-hook::
-	mkdir -p $(DESTDIR)$(sysconfdir)/motd.d
-	ln -sTfr $(DESTDIR)/run/cockpit/motd $(DESTDIR)$(sysconfdir)/motd.d/cockpit
-	mkdir -p $(DESTDIR)$(sysconfdir)/issue.d
-	ln -sTfr $(DESTDIR)/run/cockpit/motd $(DESTDIR)$(sysconfdir)/issue.d/cockpit.issue
+	mkdir -p $(DESTDIR)$(prefix)$(sysconfdir)/motd.d
+	ln -sTfr $(DESTDIR)$(prefix)/run/cockpit/motd $(DESTDIR)$(prefix)$(sysconfdir)/motd.d/cockpit
+	mkdir -p $(DESTDIR)$(prefix)$(sysconfdir)/issue.d
+	ln -sTfr $(DESTDIR)$(prefix)/run/cockpit/motd $(DESTDIR)$(prefix)$(sysconfdir)/issue.d/cockpit.issue
 
 tempconfdir = $(prefix)/lib/tmpfiles.d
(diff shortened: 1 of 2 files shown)


src/systemd/Makefile.am
```diff
@@ -25,10 +25,10 @@ dist_motd_SCRIPTS = src/systemd/update-motd
 
 # Automake: 'Variables using ... ‘sysconf’ ... are installed by install-exec.'
 install-exec-hook::
-	mkdir -p $(DESTDIR)$(sysconfdir)/motd.d
-	ln -sTfr $(DESTDIR)/run/cockpit/motd $(DESTDIR)$(sysconfdir)/motd.d/cockpit
-	mkdir -p $(DESTDIR)$(sysconfdir)/issue.d
-	ln -sTfr $(DESTDIR)/run/cockpit/motd $(DESTDIR)$(sysconfdir)/issue.d/cockpit.issue
+	mkdir -p $(DESTDIR)$(prefix)$(sysconfdir)/motd.d
```
jelly, line 28: The tests fail for CentOS/Fedora, as sysconfdir is already prefixed.

```
        sysconfdir:                 ${prefix}/etc
```

jelly (changes requested):

src/systemd/Makefile.am
lucasew: Weird 

```

               Cockpit 287
             ================

        prefix:                     /nix/store/nbdvnch6cwyk91x350xl80kvr8p82l6k-cockpit-287
        exec_prefix:                ${prefix}
        libdir:                     ${exec_prefix}/lib
        libexecdir:                 ${exec_prefix}/libexec
        bindir:                     ${exec_prefix}/bin
        sbindir:                    ${exec_prefix}/sbin
        datarootdir:                ${prefix}/share
        datadir:                    ${datarootdir}
        sysconfdir:                 /etc
        localstatedir:              ${prefix}/var
        pamdir:                     ${libdir}/security
        systemd unit dir:           ${prefix}/lib/systemd/system

        compiler:                   gcc
        cflags:                     -g -O2 -g
        cppflags:                   

        cockpit-ws user:            root
        cockpit-ws group:           
        cockpit-ws instance user:   root
        cockpit-ws instance group:  
```

lucasew: I found the problem. When the program runs it uses hardcoded $sysconfdir to look for stuff but $sysconfdir will be inside the nix store so it crashes.

lucasew:
I found out a different approach to deal with that etc problem. It looks for the config file inside the nix store and that makes features like allowed origins not to work. This PR is no longer relevant because of that. Thank you guys!

PR #18518, opened 2023-03-16
metrics: use polygon with clip path instead of svgs

KKoukiou:
Split out of https://github.com/cockpit-project/cockpit/pull/18461 to seperate the discussion about the gradient colors from porting to clip-paths.
Screen Shot 2023-03-16 at 20 30 20
Screen Shot 2023-03-16 at 20 31 45

### DIFF
--- a/pkg/metrics/metrics.jsx
+++ b/pkg/metrics/metrics.jsx
@@ -846,23 +846,24 @@ class CurrentMetrics extends React.Component {
 
 const SvgGraph = ({ data, resource, have_sat }) => {
     const dataPoints = key => (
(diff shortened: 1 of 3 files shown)


martinpitt (approved):
Thanks! This looks pretty good to me. Aside from the (minor) border changes, this PR by itself is more of an optimization than a new feature. I don't think *this* PR should be "release-note", but I'll leave it to you to which one you want to attach the "zoom" feature.

As per our recent discussion, asking @garrett for sign-off. This is a "small" change AFAIUI. Thanks!

PR #18519, opened 2023-03-16
build: Drop Makefile.deps mechanism and build dist/ in a single webpack run

martinpitt:
This gives a massive speedup compared to building them individually, as    
parsing/compiling `node_modules/` takes by far the most time:    
`node_modules/.bin/webpack` is able to build the whole dist/ in 2    
minutes (production with terser), instead of 27 linear minutes, or ~ 4    
minutes with 8 parallel processes.    
    
It's even faster in development mode without linters, then it just takes    
about 40s.    
    
`make` will now rebuild the whole dist/ if anything in pkg/ changes --    
the same approach that starter-kit and friends have used for years. That    
is already faster for a full "make dist" or jumpstart CI build. It is    
slower for incremental builds, but (1) the documented and recommended    
way to do that is with tools/webpack-watch and just building a single    
page; and (2) it will not matter any more once we switch to esbuild.    
    
We keep tools/webpack-{make.js,watch} for the time being. They are now a    
rather shallow wrapper around invoking webpack, which we can simplify    
further -- but that will require changing our habits and the    
documentation, let's keep this for a follow-up.    
    
This also paves the way for esbuild, as we don't have a mechanism to generate    
Makefile.deps there.    
    
----    
    
 - [x] clean up Makefile.deps computation in tools/webpack-make.js
 - [x] clean up the *.metainfo.xml hack
 - [x] adjust webpack-jumpstart --partial mode to not use Makefile.deps    
 - [x] adjust .github/workflows/codeql.yml and containers/flatpak/prepare which only want to build zero or one webpack


martinpitt:
@KKoukiou FYI -- still pretty much untested, but I did test the po plugin change quite thoroughly, if you already want to play around with it, and perhaps cherry-pick into the esbuild branch. Once this PR lands, most of the noise in esbuild should disappear.

martinpitt:
It's even more amazing with the recent sass compiler update (the previous version was slooooooooow). Check this out:

```
❱❱❱ rm -rf dist; time ESLINT=0 STYLELINT=0 node_modules/.bin/webpack --progress
webpack compiled successfully

real	0m41.412s
user	1m12.068s
sys	0m2.556s

❱❱❱ rm -rf dist; time NODE_ENV=production ESLINT=0 STYLELINT=0 node_modules/.bin/webpack --progress
webpack compiled successfully

real	2m3.307s
user	10m33.348s
sys	0m11.688s
```
I.e. terser accounts for the majority of work here, but still, 2 minutes for the whole thing even beats `-j8` on main (that took 3m 49s here)

martinpitt:
@allisonkarlitskaya This is nearing completion. There may be a corner case left somewhere, but the structure and scope should not change much any more. I'd appreciate a first round of review. Thanks!

tools/webpack-jumpstart
```diff
@@ -172,11 +172,18 @@
...
-    done
-done
+# for --partial, remove webpacks which changed since the jumpstart was built
+if [ -n "${partial}" ]; then
+    # assume that pkg/lib and build system changes affect all of them
+    if echo "$changed_files" | grep -q -E '(pkg/lib|Makefile|package.*.json|webpack.config.js)'; then
+        message REMOVE "dist"
+        rm -r dist
+    else
+        for webpack in $(ls dist); do
```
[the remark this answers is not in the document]
martinpitt: *shrug*, changed to a glob, but I think it's worse.

pkg/Makefile.am
```diff
@@ -4,19 +4,19 @@ pkg_TESTS = \
...
-%.metainfo.xml: %.metainfo.xml.in
+metainfo_in = \
+	$(srcdir)/pkg/sosreport/org.cockpit-project.cockpit-sosreport.metainfo.xml.in \
+	$(srcdir)/pkg/kdump/org.cockpit-project.cockpit-kdump.metainfo.xml.in \
+	$(srcdir)/pkg/selinux/org.cockpit-project.cockpit-selinux.metainfo.xml.in \
+	$(srcdir)/pkg/storaged/org.cockpit-project.cockpit-storaged.metainfo.xml.in \
+	$(srcdir)/pkg/networkmanager/org.cockpit-project.cockpit-networkmanager.metainfo.xml.in \
+	$(NULL)
+
+tmp/build/%.metainfo.xml: $(srcdir)/pkg/%.metainfo.xml.in
```
allisonkarlitskaya, line 15: I don't like using `tmp/` for this.  That's usually rather stuff that happens from our bots-type infra, not on `make`.

Is there a problem with having them in `pkg/`?  In out-of-tree builds that should end up in `pkg/` in the build directory, no?

pkg/Makefile.am
```diff
@@ -44,11 +44,10 @@ include pkg/build
...
-# `nobase_`, so we create a cockpit/ symlink and use that instead.
-# That way, the files land in $(datadir)/cockpit/ and `dist` is avoided.
-all-local:: cockpit
-cockpit:
-	$(AM_V_GEN) ln -sf $(srcdir)/dist $@
-
-nobase_data_DATA = $(patsubst dist/%,cockpit/%,$(WEBPACK_INSTALL))
+install-data-hook::
+	install -d $(DESTDIR)$(datadir)/cockpit
+	cp -r $(srcdir)/dist/* $(DESTDIR)$(datadir)/cockpit/
```
allisonkarlitskaya, line 49: That's gonna get you stuff like tests installed, isn't it?

pkg/Makefile.am
allisonkarlitskaya: Sorry.  I misunderstood that this is about dependencies, not what gets in the tarball (although it's probably also valid for this reason).

This is tricky.  Output belongs beside .in files, truly.  Why not move this all somewhere else like a new subdir in src/?

martinpitt: Done. PTAL?

martinpitt:
Let's move tools/webpack-make.js to ./build.js and let this be/become **the** entry point for bundle stuff. It's going to be called like that with esbuild as well. tools/webpack-watch can become a compat shim shell script which says "deprecated" and calls `./build.js --watch "$@"`. Let that do the `node-modules make_package_lock_json` by itself, and drop that and the `tree` stuff from pkg/build. Move the jumpstart tar build to the github workflow, it's going to be removed soon anyway. After that, pkg/build should cease to exist.

@allisonkarlitskaya , @KKoukiou ^ WDYT? I'll work on that on Monday.

pkg/Makefile.am
martinpitt: This is completely gone, it continues using nobase_data_DATA for the time being.

martinpitt:
@allisonkarlitskaya I did many of the changes that we talked about in this PR, but now this hits a wall: The jumpstart workflow is `on: pull_request_target`, so I can't change pkg/build here much. OTOH, changing that workflow to the new tools/webpack-make.js needs *this* PR first. So here's the plan:

 - [ ] Land #18522 (ready for review and landing)
 - [ ] Rebase this PR, review/land
 - [ ] Change jumpstart build workflow from pkg/build to tools/webpack-make.js; that's the [jumpstart-build branch](https://github.com/cockpit-project/cockpit/tree/jumpstart-build), which I tested in #18525. Make a PR, land it.

Then we are free to clean up pkg/build (which I'll do), perhaps move tools/webpack-make.js to ./build.js, or change the way we treat and install dist/ (needs some more discussion).

martinpitt:
I dropped #18522 to save another round of CI. I mostly wanted to test that in isolation, but it works fine. It's all in here now, and I dropped partial jumpstart mode.

martinpitt:
I took out the last commit ("tools: Teach webpack-make.js to get node_modules by itself"). It needs a different way of fixing the `FORCE:'` target, and the dist → pkg dependency also needs some thought: we want `make` to rebuild if merely the node_modules/ submodule changes. I'll do that in the next PR then, as this is big and hard enough to get green as it is.

tools/webpack-jumpstart
```diff
@@ -50,83 +47,18 @@ if [ "${NODE_ENV-}" = "development" ]; then
...
-#                       /
-#              a---b---H                     (PR head)
-#             /        |\
-#            /       --M \---N               (GitHub PR merge commits)
-#           /       /       /
-#   ---o---o---o---m---o---n                 (origin/main)
-#                   \
-#                    a'--b'--H'--L'--l'      (rebased result)
-#
-# We might be run from H, M, N, L or l.
```
allisonkarlitskaya, line 72: My beautiful artwork!!

pkg/Makefile.am
```diff
@@ -35,4 +35,4 @@ all-local:: cockpit
 cockpit:
 	$(AM_V_GEN) ln -sf $(srcdir)/dist $@
 
-nobase_data_DATA = $(patsubst dist/%,cockpit/%,$(WEBPACK_INSTALL))
+nobase_data_DATA = $(patsubst dist/%,cockpit/%,$(shell find $(srcdir)/cockpit/ -type f))
```
allisonkarlitskaya, line 38: I find this fairly objectionable.  If I understand it correctly, I think this `$(shell)` is gonna get run on every single invocation of `make`.

pkg/build
```diff
@@ -5,59 +5,32 @@ srcdir ?= .
...
-	systemd \
-	users \
-	$(NULL)
+# one built file in dist/ which we use as dependency
+DIST_STAMP = $(srcdir)/dist/static/manifest.json
 
-MANIFESTS = $(WEBPACK_PACKAGES:%=$(srcdir)/dist/%/manifest.json)
+# dynamic pkg → dist dependency, to rebuild the webpack if any web related file changes
+# exclude automake unit test log files
+PKG_INPUTS = $(shell find $(srcdir)/pkg -type f ! -name 'test*.trs' ! -name 'test*.log')
```
allisonkarlitskaya, line 13: More objections about this.  Do you want to land this, or should we wait for the changes where `webpack-make` will scan the timestamps itself?

martinpitt: I have a version where all of this goes away. But given how hard it is to get this PR to green (affected retries, and other details), I'd prefer to land this as a follow-up.

allisonkarlitskaya (approved):
I think we agree with where we're going, and this is a good step on that path, so go ahead.

PR #18353, opened 2023-02-15
Allow specifying UID during account creation

skobyda:
UID should be a integer bigger or same than 1000. There are use cases
when user would want multiple accounts to share the same UID. In such
case, a secondary confirmation button is shown to let user know that
the account will be created with a non-unique UID. The default UID is
prefilled with the next available based on the users at /etc/passwd.

[Screencast from 2023-02-15 16-23-31.webm](https://user-images.githubusercontent.com/42733240/219077484-a985e0d9-a4cf-495c-a668-cc853d4670fd.webm)

PR #18520, opened 2023-03-17
accounts: Redesign password process

### LINKED ISSUE #18391: user: Create new account modal has password-related UI issues
Follow-up to https://github.com/cockpit-project/cockpit/pull/18234#pullrequestreview-1309520648.

Screenshot of the create new user modal:

image

Issues:

1. "Confirm" should be "Confirm password".
2. Password strength meter is too far away from the entry, and there's no text to explain what's going on (unless you have an "Excellent" password.
3. Confirmation doesn't show if it matches or not. (In this screenshot, it clearly does not match and the focus is in another form.)
4. Authentication is about using a password, so it should be before the password entries.
5. Home and shell should probably come before authentication and password, especially since home changes when username does.

skobyda:
TODO:
- [x] add pixel tests

Fixes #18391 

# Accounts: Improve password process

There are several improvements to the process of setting passwords during account creation:
- The confirmation password is validated and doesn't match, validation is done after each keystroke. This improves user experience, as the user doesn't have to click "Create" whenever they want to know if passwords match.
- Password validation is now presented as warning, since a weak password is not actually forbidden, but discouraged nevertheless.

Screenshot from 2023-04-05 12-41-47

### DIFF
--- a/pkg/users/account-create-dialog.js
+++ b/pkg/users/account-create-dialog.js
@@ -131,7 +131,7 @@ function AccountCreateBody({ state, errors, change, shells }) {
             </FormGroup>
 
             <PasswordFormFields password_label={_("Password")}
-                                password_confirm_label={_("Confirm")}
+                                password_confirm_label={_("Confirm password")}
                                 error_password={errors?.password}
                                 error_password_confirm={errors?.password_confirm}
                                 idPrefix="accounts-create-password"
--- a/pkg/users/account-create-dialog.js
+++ b/pkg/users/account-create-dialog.js
@@ -202,6 +202,13 @@ function validate_home_dir(dir, directoryExpected) {
     return cockpit.spawn(["test", "!", directoryExpected ? "-d" : "-f", dir], { superuser: "require" });
 }
 
+function validate_password(password) {
+    if (!password)
+        return _("Empty password");
+
+    return null;
+}
+
 function validate_password_confirm(password_confirm, password) {
     if (password_confirm !== password)
         return _("The passwords do not match");
@@ -351,6 +358,7 @@ export function account_create_dialog(accounts, min_uid, max_uid, shells) {
         const errs = { };
 
         errs.real_name = validate_real_name(real_name);
+        errs.password = validate_password(password);
         errs.password_confirm = validate_password_confirm(password_confirm, password);
 
         if (password.length > 256)
@@ -358,12 +366,16 @@ export function account_create_dialog(accounts, min_uid, max_uid, shells) {
 
         errs.user_name = validate_username(user_name, accounts);
 
-        const promises = [
-            password_quality(password, force_weak)
-                    .catch(ex => {
-                        errs.password = (ex.message || ex.toString()).replaceAll("\n", " "); // not-covered: OS error
-                    })
-        ];
+        const promises = [];
+        // only evaluate password score if no other password error si present
+        if (!errs.password) {
+            promises.push(
+                password_quality(password, force_weak)
+                        .catch(ex => {
+                            errs.password = (ex.message || ex.toString()).replaceAll("\n", " "); // not-covered: OS error
+                        })
+            );
+        }
         if (!force_uid)
             errs.uid = validate_uid(uid, accounts, min_uid, max_uid, change);
 
--- a/pkg/users/account-create-dialog.js
+++ b/pkg/users/account-create-dialog.js
@@ -50,6 +50,18 @@ function AccountCreateBody({ state, errors, change, shells }) {
         shell, isShellSelectExpanded,
     } = state;
 
+    // We want to let user know that password and confirmation password do not match without them having to constantly click on "Create" to validate the form.
+    // But we also don't want to show an error message while they are typing the confirmation password, only when they are finished typing it.
+    // To solve the issue of telling that user is finished writing the confirmation password, let's do the following:
+    // The dialog does not validate if passwords match until:
+    //     1. they are at least the same length (which signals user has finished typing)
+    // OR
+    //     2. user submits the form (which also signals they are finished typing)
+    // Once that happens, and passwords do not match, the confirm password is validated after each keystroke.
+    let dynamic_password_confirm_error;
+    if (state.password_confirm_dirty)
+        dynamic_password_confirm_error = validate_password_confirm(state.password_confirm, state.password);
+
     return (
         <Form isHorizontal onSubmit={apply_modal_dialog}>
             <FormGroup label={_("Full name")}
@@ -133,7 +145,7 @@ function AccountCreateBody({ state, errors, change, shells }) {
             <PasswordFormFields password_label={_("Password")}
                                 password_confirm_label={_("Confirm password")}
                                 error_password={errors?.password}
-                                error_password_confirm={errors?.password_confirm}
+                                error_password_confirm={dynamic_password_confirm_error || errors?.password_confirm}
                                 idPrefix="accounts-create-password"
                                 change={change} />
         </Form>
@@ -188,6 +200,13 @@ function validate_uid(uid, accounts, min_uid, max_uid, change) {
 
 function validate_home_dir(dir, directoryExpected) {
     return cockpit.spawn(["test", "!", directoryExpected ? "-d" : "-f", dir], { superuser: "require" });
+}
+
+function validate_password_confirm(password_confirm, password) {
+    if (password_confirm !== password)
+        return _("The passwords do not match");
+
+    return null;
 }
 
 function suggest_username(realname) {
@@ -246,6 +265,7 @@ export function account_create_dialog(accounts, min_uid, max_uid, shells) {
         user_name: "",
         password: "",
         password_confirm: "",
+        password_confirm_dirty: false,
         locked: false,
         confirm_weak: false,
         change_passw_force: false,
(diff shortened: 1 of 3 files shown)


pkg/users/account-create-dialog.js
```diff
@@ -42,6 +42,16 @@ function AccountCreateBody({ state, errors, change }) {
...
     } = state;
 
+    // We want to let user know that password and confirmation password do not match without them having to constantly click on "Create" to validate the form.
+    // But we also don't want to show an error message while they are typing the confirmation password, only when they are finished typing it.
+    // To solve the issue of telling that user is finished writing the confirmation password, let's do the following:
+    // Since both passwords should be the same length, we should show error message while user is typing,
+    // but only if the confirm password is same length (or higher) than the original password (i.e. they are finished typing)
+    let dynamic_password_confirm_error;
+    if (state.password_confirm.length >= state.password.length)
+        dynamic_password_confirm_error = validate_password_confirm(state.password_confirm, state.password);
```
marusak, line 52: I would let @garrett to decide what we want here but does this mean that we don't show error if the passwords are different length? Which I think is the most usual case when passwords differ that you hit by accident one more key or forget some part of it. Or does this also validate after you blur? Which I think should be the primary method for validating fields.

pkg/users/account-create-dialog.js
```diff
@@ -42,6 +42,16 @@ function AccountCreateBody({ state, errors, change }) {
...
+    console.log(dynamic_password_confirm_error);
```
marusak, line 53: forgotten debugging output

pkg/users/account-create-dialog.js
marusak: Also this is validating one field under very specific circumstances. Why not validating all fields on blur?

skobyda: I agree onBlur is also a nice alternative, which I would like to see used more not only in this form but across the whole cockpit. 
But redesigning the whole form to work with onBlur is a bit bigger, since you have to redo the whole additional login related to "Create with weak password", etc. 
And I wanted to keep this PR small, since this PR is blocking #18470  and #18353 , I want to get it in quickly, so I saw this approach a nice fix without having to redo the whole validation for the whole dialog.

But again, I'm not against onBlur, I like that approach too, so if we agree to use it, in that case I would split the validation refactoring into a separate PR, so this can get merged quickly and not block the other 2 PRs.

pkg/users/account-create-dialog.js
skobyda: Ah thanks, I see it made the test output unreadable.

pkg/users/account-create-dialog.js
garrett: I thought I replied to this yesterday. I don't see it here though. (Or perhaps I'm thinking of a similar conversation?)

We should try to catch things like this on the fly, before submitting them. But I'm fine if that's a follow-up.

Related: @jelly and I were talking about improving validation in Cockpit. He was looking into it. It's probably worth having a larger conversation about this across Cockpit and improving everything, everywhere. (I was also talking with the PatternFly team about validation too. It's an odd omission that they just do not cover.) It's not necessarily related to this PR, but having guidelines and examples in place would make it clearer for PRs like this in the future. (And perhaps @skobyda could talk with @jelly if there's a good technique for this, even for this PR or a follow-up perhaps.)

But there are so many other changes here, I'm OK with improving this specifically in a follow-up.

garrett (dismissed):
I'm happy with the changes that are included here.

We do need to revisit the password part, and do something like show the error state of the input form as well, as shown at https://www.patternfly.org/v4/demos/password-strength/

(The outline and the icon)

If you can quickly add the error state to the entry in this PR, that would be _excellent_. (Probably toggling the state of the entry in the validate_password_confirm function.) Otherwise, it's good to have as a follow-up. I don't want to block the other changes on this.

skobyda:
> I'm happy with the changes that are included here.
> 
> We do need to revisit the password part, and do something like show the error state of the input form as well, as shown at https://www.patternfly.org/v4/demos/password-strength/
> 
> (The outline and the icon)
> 
> If you can quickly add the error state to the entry in this PR, that would be _excellent_. (Probably toggling the state of the entry in the validate_password_confirm function.) Otherwise, it's good to have as a follow-up. I don't want to block the other changes on this.

Since the reordering of the Account create dialog was merged already in #18353 , this PR now only consists of password-related stuff, so I think it makes sense to continue discussion here.
So let's fix our password process in this PR.

skobyda:
@garrett something like this?

Screenshot from 2023-03-28 11-57-40

pkg/lib/cockpit-components-password.scss
```diff
@@ -5,3 +5,7 @@
     display: none;
   }
 }
+
+#password-error-message .pf-c-helper-text__item-text {
+    color: var(--pf-c-form__helper-text--m-error--Color) !important;
```
garrett, line 10: Does this really need an `!important`? It is already using an ID (which you should also try to avoid, but less so than `!important`).

Also, why not just use the error variant?

For example, here's the error variant with an icon:

```jsx
    <HelperText>
      <HelperTextItem variant="error" hasIcon>
        This is error helper text
      </HelperTextItem>
    </HelperText>
```

https://www.patternfly.org/v4/components/helper-text

garrett (changes requested):
The error message for password is a bit weird when the password is empty, right?

image

Perhaps it should be something more like:

:warning: Empty password

Additionally, I'm able to create the password anyway, so it's not actually an error, but a warning. We should probably change it, even for the weak passwords, so it should be the warning gold color with warning icon :warning: instead of red error one. (This would be the case for all password warnings, unless weak passwords are forbidden by policy. And if that's the case, with the policy, there shouldn't be a bypass button.)

skobyda:
@garrett updated the PR so the validation is now in form of warnings:
Screenshot from 2023-04-03 14-24-33
Screenshot from 2023-04-03 14-24-14

pkg/users/account-create-dialog.js
skobyda: Solution discussed on Matrix:
```
Garrett
skobyda: if you're typing, then you're not done. but if if you trigger the check and they don't match, then you should be told that until they do match.
Garrett
skobyda: the delayed check is just to cover the part where you're typing and not done
Garrett
skobyda: but it shouldn't check and then not check and go back and forth on checking... especially when you trigger it and then they're obviously not matching (different lengths)

....

skobyda
Garrett: sooo, I should wait until both passwords are the same length, and if they are and they don't match, only then start checking after each keystroke?
Garrett
this would not be a problem if we did away with the password check and had a toggle visibility... ahem
Garrett
but anyway, yes
```

So not this PR implements it this way:

The dialog does not validate if passwords match until:
1. they are at least the same length (which signals user has finished typing)
OR
2. user clicks on "Create" (which also signals they are finished typing)
Once that happens, and passwords do not match, the confirm password is validated after each keystroke.

This still keeps the old form onSubmit validation, which @marusak was concerned would be lost, and also adds the new approach which tells user more dynamically if the passwords do not mathc.

skobyda:
@marusak mind giving it another code-review?

garrett:
> But it still keeps the old onSubmit validation. So if the user submits the form, and passwords do not match, since then it validates it after each keystroke.

:+1: Yep. Even when we do active validation, we still need to check it on submit. (Just like even when we do client-side validation, we still need to sanitize and validate on the server side.)

garrett (dismissed):
You've addressed my concerns, and the password validation now works as expected. Thanks!

skobyda:
Updated pixel tests.

@marusak @martinpitt Do you mind giving it code-review mind giving it another code-review? Since design and UX is approved, and if there will be no big blockers during code-review, I think this could land in release

pkg/users/account-create-dialog.js
```diff
@@ -50,6 +50,18 @@ function AccountCreateBody({ state, errors, change, shells }) {
...
+    // But we also don't want to show an error message while they are typing the confirmation password, only when they are finished typing it.
+    // To solve the issue of telling that user is finished writing the confirmation password, let's do the following:
+    // The dialog does not validate if passwords match until:
+    //     1. they are at least the same length (which signals user has finished typing)
+    // OR
+    //     2. user submits the form (which also signals they are finished typing)
+    // Once that happens, and passwords do not match, the confirm password is validated after each keystroke.
+    let dynamic_password_confirm_error;
+    if (state.password_confirm_dirty)
+        dynamic_password_confirm_error = validate_password_confirm(state.password_confirm, state.password);
```
martinpitt, line 63: Code style: Whenever possible, use `const`. Here this is quite straightforward:

    const dynamic_password_confirm_error = state.password_confirm_dirty && validate_password_confirm(..);

(with a line break)

pkg/users/account-create-dialog.js
```diff
@@ -246,6 +265,7 @@ export function account_create_dialog(accounts, min_uid, max_uid, shells) {
         user_name: "",
         password: "",
         password_confirm: "",
+        password_confirm_dirty: false,
```
martinpitt, line 275: I wouldn't know what "dirty" means in this context; Something like "_do_check" perhaps?

martinpitt (approved):
Thanks! I have some stylistic comments, but they aren't important to hold this up. I retriggered the test flakes.

PR #18521, opened 2023-03-17
Don't invoke our esbuild plugins when the build has errors

jelly:
This makes working on podman a lot better as when the build fails the dist dir is not there and rsync just kill's everything and I notice it minutes later :)


martinpitt (approved):
Thank you! :100:

PR #18525, opened 2023-03-20
TEST: jumpstart-build branch

martinpitt:
As the webpack-jumpstart workflow is `pull_request_target`, it cannot be tested with a normal PR. Let's test it here.


martinpitt:
The [jumpstart build](https://github.com/cockpit-project/cockpit/actions/runs/4464617625) succeeded, and from a completely clean checkout of this branch I can do `tools/webpack-jumpstart`, it fetched the tarball, and in Cockpit I can see the "Health" → " 'sup?" change. :heavy_check_mark:

PR #18527, opened 2023-03-20
build: simplify handling of systemd_ctypes and ferny

allisonkarlitskaya:
Drop our attempts to cache ferny and systemd_ctypes — it's just not worth it.  Both of them are tiny.

Instead, establish that we expect submodules in modules/ to be checked out, generally.  Add a (trivial) script to do that and call it from a bunch of "entry point" type places.

Note: we don't dist modules/checkout, and even if we did, it wouldn't work, since it needs to be run in a git checkout.  We therefore also carefully avoid adding calls to this script from anything that lands in the tarball.

<hr>

NB: the eventual goal here is also to lose the `git-utils.sh` script.  The only thing left using it is `node_modules` and this will eventually be made a part of the esbuild process.

 - [x] https://github.com/cockpit-project/cockpit-machines/pull/983
 - [x] https://github.com/cockpit-project/cockpit-podman/pull/1234
 - [x] https://github.com/cockpit-project/cockpit-certificates/pull/125
 - [x] https://github.com/osbuild/cockpit-composer/pull/1830
 - [x] https://github.com/cockpit-project/starter-kit/pull/631
 - [x] https://github.com/candlepin/subscription-manager-cockpit/pull/52


martinpitt:
@allisonkarlitskaya Are the test/reference changes accidental? This shouldn't change pixels.

allisonkarlitskaya:
> @allisonkarlitskaya Are the test/reference changes accidental? This shouldn't change pixels.

Yes.  Accidental.

martinpitt (dismissed):
All depending PRs landed, and this looks nice! Thanks for the cleanup. Feel free to undraft and land, I trigger the fcos test in the meantime.

allisonkarlitskaya:
This can't land yet — `test/image-prepare` calls `tools/make-wheel` still.  Which is pretty hilarious...

allisonkarlitskaya:
Let's get a full test run to make sure we don't have any unexpected "issues".

martinpitt (approved):
Tests look good so far -- if this breaks, I figure it should break very early on during build, and thoroughly. Thanks!

PR #18528, opened 2023-03-20
systemd: Improve realmd label

### LINKED ISSUE #9772: realmd: Better label when realmd is not installed, but machine is in a domain
Fallout from PR #9759: If realmd is not installed but the machine is actually joined to a domain, we transition from the "installing realmd" dialog to the "are you sure you want to leave domain" dialog.

We just can't detect that it's in a domain before installing realmd. Perhaps the label should be something more neutral then, and not imply that the machine isn't joined?

Alcadeus0:
fixes #9772 
Updated label from "Join Domain" to "Install Domain Support"

### DIFF
--- a/pkg/systemd/overview-cards/realmd.jsx
+++ b/pkg/systemd/overview-cards/realmd.jsx
@@ -445,7 +445,7 @@ export const RealmButton = ({ realmd_client }) => {
     useEvent(superuser, "changed");
 
     const buttonTooltip = superuser.allowed ? realmd_client.error : _("Not permitted to configure realms");
-    const buttonText = realmd_client.joined.length ? realmd_client.joined.map(r => r.Name).join(", ") : _("Join domain");
+    const buttonText = !realmd_client.install_realmd ? (realmd_client.joined.length ? realmd_client.joined.map(r => r.Name).join(", ") : _("Join domain")) : _("Install realmd support");
     const buttonDisabled = !superuser.allowed || (realmd_client.error && !realmd_client.install_realmd);
 
     const onClicked = () => {
@@ -467,7 +467,7 @@ export const RealmButton = ({ realmd_client }) => {
                     excuse={ buttonTooltip }>
             <Button id="system_information_domain_button" variant="link"
                     onClick={onClicked}
-                    isInline isDisabled={buttonDisabled} aria-label={_("Join domain")}>
+                    isInline isDisabled={buttonDisabled} aria-label={buttonText}>
                 { buttonText }
             </Button>
         </Privileged>);
--- a/test/verify/check-system-realms
+++ b/test/verify/check-system-realms
@@ -1169,7 +1169,7 @@ class TestPackageInstall(packagelib.PackageCase):
         m.execute("systemctl unmask packagekit")
         self.login_and_go("/system")
         # Joining a domain should bring up the install dialog
-        b.wait_text(self.domain_sel, "Join domain")
+        b.wait_text(self.domain_sel, "Install realmd support")
         self.waitTooltip("requires installation of realmd")
 
         b.click(self.domain_sel)
@@ -1188,7 +1188,7 @@ class TestPackageInstall(packagelib.PackageCase):
         self.login_and_go("/system")
 
         # Joining a domain should bring up the install dialog
-        b.wait_text(self.domain_sel, "Join domain")
+        b.wait_text(self.domain_sel, "Install realmd support")
         self.waitTooltip("requires installation of realmd")
 
         b.click(self.domain_sel)
@@ -1211,7 +1211,7 @@ class TestPackageInstall(packagelib.PackageCase):
         self.login_and_go("/system")
 
         # Joining a domain should bring up the install dialog
-        b.wait_text(self.domain_sel, "Join domain")
+        b.wait_text(self.domain_sel, "Install realmd support")
         self.waitTooltip("requires installation of realmd")
 
         b.click(self.domain_sel)


pkg/systemd/overview-cards/realmd.jsx
```diff
@@ -467,7 +467,7 @@ export const RealmButton = ({ realmd_client }) => {
                     excuse={ buttonTooltip }>
             <Button id="system_information_domain_button" variant="link"
                     onClick={onClicked}
-                    isInline isDisabled={buttonDisabled} aria-label={_("Join domain")}>
+                    isInline isDisabled={buttonDisabled} aria-label={_("Install Domain Support")}>
```
jelly, line 470: This change does not really change the Button text, it only changes the [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) so this wouldn't lead to any visible changes.

What should happen is that the `buttonText` is changed according to the state of `realmd_client.install_realmd`. If it's not installed we should show a `Install realmd support` button.


Note that this also might require tests to be changed

```
test/verify/check-system-realms:                b.wait_text(self.domain_sel, "Join domain")
test/verify/check-system-realms:                b.wait_text_not(self.domain_sel, "Join domain")
test/verify/check-system-realms:        b.wait_text(self.domain_sel, "Join domain")
test/verify/check-system-realms:        b.wait_text(self.domain_sel, "Join domain")
test/verify/check-system-realms:        b.wait_text(self.domain_sel, "Join domain")
test/verify/check-system-realms:        b.wait_text(self.domain_sel, "Join domain")
```

jelly (changes requested):

pkg/systemd/overview-cards/realmd.jsx
Alcadeus0: @jelly do all the specified tests need to be updated to "Install realmd support" ?

Alcadeus0:
@martinpitt  TestPackageInstall.testInstall gives an uncaught promise error as the button text does not change to "Install realmd support" from "Join domain" on reloading the page and entering "/system".
TestPackageInstall.testDialogTransition however succeeds in reading the updated text and passes the test. Both the tests follow the same procedure of logging in, reloading and entering yet one fails.

test/verify/check-system-realms
```diff
@@ -1160,16 +1160,20 @@ class TestPackageInstall(packagelib.PackageCase):
         # case 1: disable PackageKit
         m.execute("systemctl mask packagekit; systemctl stop packagekit.service || true")
         self.login_and_go("/system")
-        b.wait_text(self.domain_sel, "Join domain")
+        b.reload()
+        b.enter_page("/system")
```
martinpitt, line 1164: You don't need this -- the line before just logged into cockpit. This slows down the test in the best case, and in the worst it cases race conditions. Same thing for all other reloads that you added -- sorry for the confusing advice on IRC then, I didn't look at that actual test here.

test/verify/check-system-realms
```diff
@@ -1160,16 +1160,20 @@ class TestPackageInstall(packagelib.PackageCase):
...
+        b.wait_text(self.domain_sel, "Install realmd support")
```
martinpitt, line 1165: This can't work -- this case tests disabling packagekit, so `packagekit.detect()` will be false. As we don't have a state for "realmd does not exist and we can't install it", keeping the disabled "Join domain" is fine for now.

test/verify/check-system-realms
```diff
@@ -1160,16 +1160,20 @@ class TestPackageInstall(packagelib.PackageCase):
...
         b.logout()
 
         # case 2: enable PackageKit, but no realmd package available
         m.execute("systemctl unmask packagekit")
         self.login_and_go("/system")
+        b.reload()
+        b.enter_page("/system")
         # Joining a domain should bring up the install dialog
-        b.wait_text(self.domain_sel, "Join domain")
+        b.wait_text(self.domain_sel, "Install realmd support")
```
martinpitt, line 1176: This is the interesting and new case here. :+1:

martinpitt (changes requested):
Thanks! The tests need some work, but the code change looks good!

test/verify/check-system-realms
Alcadeus0: No worries, I assumed reloading would take us out of the system page and had kept it. But running the tests now without it still has them pass when they weren't initially.

martinpitt: No, there is nothign to "slow down". wait_text() already waits for up to 15 seconds -- if it does not happen by then, it's broken (the detection does not nearly take that long)

test/verify/check-system-realms
```diff
@@ -1211,7 +1211,7 @@ class TestPackageInstall(packagelib.PackageCase):
         self.login_and_go("/system")
 
         # Joining a domain should bring up the install dialog
-        b.wait_text(self.domain_sel, "Join domain")
+        b.wait_text(self.domain_sel, "Install realmd support")
```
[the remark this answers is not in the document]
Alcadeus0: I am not quite sure I understand, by label are you referring to the "Join Doman" text in the tests.
Because (after the changes we discussed) only one test needs to really be Updated for now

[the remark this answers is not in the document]
Alcadeus0: Gotcha, I've made the change in case 3 as well.

martinpitt:
There's [another test](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18528-20230330-175658-cf786c81-debian-stable/log.html#351) that needs to be adjusted -- see the"Error: actual text: Install realmd support" and the 
[screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18528-20230330-175658-cf786c81-debian-stable/TestPackageInstall-testInstall-debian-stable-127.0.0.2-2401-FAIL.png). @Alcadeus0 Please try to run the tests locally, these should be perfectly reproducible.

martinpitt (approved):
Perfect, all green now. Thanks, and congratulations for your first landed PR!

PR #18530, opened 2023-03-20
workflows: Build jumpstart tarball without pkg/build

martinpitt:
We only ever need the webpack-jumpstart.tar make target in the workflow, and it is significantly easier to build it right there than through pkg/build.

This is the first half of eliminating pkg/build entirely. Also, when moving to esbuild, we are going to drop the jumpstart mechanism entirely; that will be easier with fewer interaction points with the build system.


martinpitt:
Works!

.github/workflows/webpack-jumpstart.yml
```diff
@@ -25,7 +25,10 @@ jobs:
              git fetch --no-tags origin ${BASE_SHA}:base ${HEAD_SHA}:head
              git checkout --detach base
              git merge --no-edit head
-             pkg/build webpack-jumpstart.tar -j$(nproc)")"
+             git rev-parse HEAD^{tree} > tree
+             tools/node-modules make_package_lock_json
```
martinpitt, line 29: Note: This line is temporary, to make this workflow-changing PR as small as possible. My next PR will teach webpack-make.js to check out n_m/ by itself, and drop this again.

allisonkarlitskaya (approved):

PR #18466, opened 2023-03-08
systemd/services: Read CLOCK_BOOTTIME with Python

martinpitt:
Before it was bogus:

image

Now it's correct:

image

```
❱❱❱ systemctl --user list-timers
NEXT                        LEFT       LAST                        PASSED     UNIT                    ACTIVATES                
Wed 2023-03-08 10:34:19 CET 27min left Wed 2023-03-08 09:34:19 CET 32min ago  backup.timer            backup.service
n/a                         n/a        Sat 2023-03-04 07:28:53 CET 4 days ago grub-boot-success.timer grub-boot-success.service

❱❱❱ date
Wed Mar  8 10:06:30 AM CET 2023
```

Fixes #18418

PR #18532, opened 2023-03-20
modules: bump ferny version

allisonkarlitskaya:
This version of ferny avoids using socket.send_fds() in the interaction client (ie: askpass helper), unbreaking our tests on centos-8-stream.


allisonkarlitskaya:
So this is definitely better....

an awful lot of tests seem to be failing with

```
AttributeError: module 'time' has no attribute 'clock_gettime_ns'
```

in the log.  That's from https://github.com/cockpit-project/cockpit/pull/18466 I guess.

martinpitt:
@allisonkarlitskaya The clock_gettime_ns issue is just noise. Marius pointed it out today, it's now on my todo list. c8s/pybridge passed in #18466 , so I didn't see the warning because we don't show the output of successful tests. Indeed there's a handful of failures now which we need to track down. Some are also flakes, like this `asyncio-ERROR: Exception in callback None()` unexpected message that just keeps happening every now and then.

martinpitt:
Actually no, at least 4 tests fail on "asyncio-ERROR: Exception in callback None()", I think that's new. But that doesn't have to block this PR, this is already a big improvement.

martinpitt (approved):

PR #18533, opened 2023-03-20
systemd/services: Avoid time.clock_gettime_ns()

martinpitt:
This does not yet exist in Python 3.6 (on RHEL 8), thus causing a warning message and the root-only fallback. Use time.clock_gettime() instead, which was introduced in Python 3.3 already.

Change the Python code to return µs instead of ns, which simplifies the calculation.

---

This was a regression from #18466, but I didn't notice because we don't show errors from passed tests. In failed tests it [looks like this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18532-20230320-165624-d2a3bf0e-centos-8-stream-pybridge/log.html#302-2):

    AttributeError: module 'time' has no attribute 'clock_gettime_ns'

### DIFF
--- a/pkg/systemd/services.jsx
+++ b/pkg/systemd/services.jsx
@@ -72,12 +72,12 @@ function debug() {
 
 export function updateTime() {
     // To correctly interpret monotonic timers, we need to read CLOCK_MONOTONIC. This cannot be read with shell tools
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
happy bots, happy lis

PR #18534, opened 2023-03-20
pkg: Update dist/ install rule to avoid `find` in dependencies

martinpitt:
This has caused a lot of confusing noise during build:

    find: ‘../../cockpit/’: No such file or directory

Because the `find` command runs on every invocation of `make`, even when dist/ was not built yet. We only really need this for `make install`, so replace this with a procedural rule. automake would previously generate `install` calls for all files in `Makefile.deps`, or now all files in dist/, so this is still functionally equivalent.

This also avoids the hack with creating a `cockpit` symlink.

### DIFF
--- a/pkg/Makefile.am
+++ b/pkg/Makefile.am
@@ -28,11 +28,8 @@ include pkg/build
 # seeing them during ./autogen.sh, but need make to find them at compile time.
 -include $(wildcard pkg/Makefile.qunit*)
 
-# automake doesn't support stripping initial path components (ie: `dist`) with
-# `nobase_`, so we create a cockpit/ symlink and use that instead.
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
TIL:
 - `find -exec` can have `{}` twice
 - `{}` can appear as part of another string

I wonder if you'd like to replace the uninstall rule with a `find -exec rm ...` invocation... but since this is just to pacify distcheck, probably not necessary.

martinpitt:
> I wonder if you'd like to replace the uninstall rule with a find -exec rm ...

That's `-delete`. It would either restrict to `-type f`, or if it also deletes directories, do the same thing as the `rm -r`, so I thought let's keep it simple.

PR #18535, opened 2023-03-21
More build cleanups, rename tools/webpack-make.js to ./build.js

martinpitt:
The second part of my build cleanups, and convergence with esbuild. From the user's POV this changes the CLI entry point from tools/webpack-{make.js,watch} to ./build.js. This is less typing, but mostly anticipates the move to esbuild. cockpit-podman already uses ./build.js as top-level entry point, and eventually all our projects will.

@allisonkarlitskaya, I'm not actually sure about the "Move "needs build" detection from pkg/build into webpack-make.js" commit. We did talk about it, but it honestly does not help with anything; maybe when we talked about it we didn't take into account that `make` will still have to do the `package-lock.json` handling? I don't think we'd be any worse off with just keeping the `PKG_INPUTS` find command, it's very fast, and build.js otherwise has to do it for each `make` invocation as well.

@KKoukiou , can you please have a look at the `./build.js` renaming and CLI? I *think* this should be mostly/fully compatible with eslint, except that we'll most probably drop the possibility to build just a single page (as it's fast enough to build everything).

Blockers/TODO:
- [ ] Do not land before PR #18526


martinpitt:
Ach, [jumpstart can't work with this](https://github.com/cockpit-project/cockpit/actions/runs/4476243271/jobs/7866409767?pr=18535) as it gets built from main. I'll re-send this PR from origin, and test that separately.

PR #18536, opened 2023-03-21
More build cleanups, rename tools/webpack-make.js to ./build.js

martinpitt:
The second part of my build cleanups, and convergence with esbuild. From the user's POV this changes the CLI entry point from tools/webpack-{make.js,watch} to ./build.js. This is less typing, but mostly anticipates the move to esbuild. cockpit-podman already uses ./build.js as top-level entry point, and eventually all our projects will.

@allisonkarlitskaya, I'm not actually sure about the "Move "needs build" detection from pkg/build into webpack-make.js" commit. We did talk about it, but it honestly does not help with anything; maybe when we talked about it we didn't take into account that `make` will still have to do the `package-lock.json` handling? I don't think we'd be any worse off with just keeping the `PKG_INPUTS` find command, it's very fast, and build.js otherwise has to do it for each `make` invocation as well.

@KKoukiou , can you please have a look at the `./build.js` renaming and CLI? I *think* this should be mostly/fully compatible with eslint, except that we'll most probably drop the possibility to build just a single page (as it's fast enough to build everything).

Blockers/TODO:
- [x] Do not land before PR #18526
- [x] Ensure jumpstart build still works: PR #18537


martinpitt:
> I'm not actually sure about the "Move "needs build" detection from pkg/build into webpack-make.js" commit

It's actually significantly hairy to do that correctly for srcdir != builddir, it creates a second empty dist/static/manifest.json which is then empty. I take out [that commit](https://github.com/cockpit-project/cockpit/pull/18536/commits/a9c1a06e89346ac3a340bc4147387a974a7ebc79) -- I don't like it much anyway, and being that finnicky pushed me over the line.

allisonkarlitskaya:
my 2¢: rewrite from $(shell find) to some kind of recursive wildcard expansion and I'm happy.  Your choice about how you accomplish that — I can think of two good ways (which you also know about).

martinpitt:
With the current `$(shell find)` approach in main:
```
❱❱❱ hyperfine make
Benchmark 1: make
  Time (mean ± σ):     227.9 ms ±   9.6 ms    [User: 209.7 ms, System: 29.2 ms]
  Range (min … max):   223.5 ms … 256.6 ms    11 runs
```

with my local rewrite using a recursive wildcard there is no measurable difference:
```
❱❱❱ hyperfine make 
Benchmark 1: make
  Time (mean ± σ):     227.8 ms ±   4.2 ms    [User: 211.9 ms, System: 27.1 ms]
  Range (min … max):   224.4 ms … 240.7 ms    12 runs
```
So I'm torn.. It's more complicated now, but also more elegant as there's no shelling out. I'll add it as a separate commit here.

pkg/Makefile.am
```diff
@@ -25,9 +25,12 @@ dist_sshmanifest_DATA = pkg/ssh/manifest.json
 # one built file in dist/ which we use as dependency
 DIST_STAMP = $(srcdir)/dist/static/manifest.json
 
+# https://blog.jgc.org/2011/07/gnu-make-recursive-wildcard-function.html
+rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst *,%,$2),$d))
+
 # dynamic pkg → dist dependency, to rebuild the webpack if any web related file changes
 # exclude automake unit test log files
-PKG_INPUTS = $(shell find $(srcdir)/pkg -type f ! -name 'test*.trs' ! -name 'test*.log')
+PKG_INPUTS = $(filter-out %.log %.trs,$(call rwildcard,$(srcdir)/pkg,*))
```
allisonkarlitskaya, line 33: Why do we get test log files in pkg ?

martinpitt: It's from pkg/users/test-list-public-keys.sh. Unfortunately automake puts the corresponding *.log and *.trs next to the test itself (into the source dir).

martinpitt:
I backed out the recursive wildcard commit again. It doesn't help performance, is hard to debug, and most importantly, it includes directories, which seem to upset make -- it tries to rebuild the webpack during distcheck and in image-prepare VMs, which is wrong and doesn't work. I feel it's not worth the trouble.

tools/webpack-make.js
```diff
@@ -83,3 +64,29 @@ function process_result(err, stats) {
...
+            const time = new Date().toTimeString().split(' ')[0];
+            process.stdout.write(`${time} Build started\n`);
+        });
+        compiler.watch(config.watchOptions, process_result);
+    } else {
+        compiler.run(process_result);
+    }
+}
+
+build();
```
allisonkarlitskaya: No `await` or `.run()` or similar?  How does it know when to quit?

tools/webpack-make.js
```diff
@@ -1,15 +1,16 @@
...
 
-import webpack from 'webpack';
+import child_process from 'child_process';
 import path from 'path';
-import argparse from 'argparse';
 import process from 'process';
-import { CockpitRsyncWebpackPlugin } from '../pkg/lib/cockpit-rsync-plugin.js';
+
+// ensure node_modules is present and up to date
+child_process.spawnSync('tools/node-modules', ['make_package_lock_json'], { stdio: 'inherit' });
```
allisonkarlitskaya: This is an example of a thing that doesn't need `package-lock.json`...

On the other hand, `checkout` will always check out.

We need some kind of a new approach/target here, I think...  But let's not block on it.

allisonkarlitskaya (approved):
Looks good, thanks.

tools/webpack-make.js
martinpitt: Top-level await is always a bit problematic. Having an `async main()` and calling it (without await) is the suggested approach.

PR #18539, opened 2023-03-21
Update documentation and comments to not be webpack specific

martinpitt:
webpack is an implementation detail and soon change to esbuild. Make the verbiage more neutral.

---

Split from PR #18447 to reduce the noise there.

### DIFF
--- a/build.js
+++ b/build.js
@@ -9,8 +9,8 @@ child_process.spawnSync('tools/node-modules', ['make_package_lock_json'], { stdi
 
 const parser = (await import('argparse')).default.ArgumentParser();
 parser.add_argument('-c', '--config', { help: "Path to webpack.config.js", default: "webpack.config.js" });
-parser.add_argument('-r', '--rsync', { help: "rsync webpack to ssh target after build", metavar: "HOST" });
-parser.add_argument('-w', '--watch', { action: 'store_true', help: "Enable webpack watch mode" });
+parser.add_argument('-r', '--rsync', { help: "rsync bundles to ssh target after build", metavar: "HOST" });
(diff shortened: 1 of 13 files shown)


build.js
```diff
@@ -4,15 +4,17 @@ import child_process from 'child_process';
...
 // ensure node_modules is present and up to date
 child_process.spawnSync('tools/node-modules', ['make_package_lock_json'], { stdio: 'inherit' });
 
 const parser = (await import('argparse')).default.ArgumentParser();
 parser.add_argument('-c', '--config', { help: "Path to webpack.config.js", default: "webpack.config.js" });
 parser.add_argument('-r', '--rsync', { help: "rsync bundles to ssh target after build", metavar: "HOST" });
 parser.add_argument('-w', '--watch', { action: 'store_true', help: "Enable watch mode" });
-parser.add_argument('-e', '--no-eslint', { action: 'store_true', help: "Disable eslint linting" });
-parser.add_argument('-s', '--no-stylelint', { action: 'store_true', help: "Disable stylelint linting" });
+parser.add_argument('-e', '--no-eslint', { action: 'store_true', help: "Disable eslint linting", default: production });
```
allisonkarlitskaya, line 16: ...and what if we want to turn them on in production mode?

martinpitt: In current main, webpack.config.js defaults to "no linting" in production mode, so actually the visible behaviour is the same. However, you can turn it back on with an environment variable. That mechanism changed in the esbuild PR. But this PR is really meant as a "quickly get rid of uncontroversial noise", so I'll just back this out for now.

allisonkarlitskaya (approved):

PR #18542, opened 2023-03-22
pkg: When importing from lib don't use file extension

martinpitt:
Unfortunately, scss imports with non relative paths still don't work with esbuild-wasm if these are inside jsx files:
https://github.com/evanw/esbuild/issues/3001

Use relative paths for importing scss from JS. Same for .sh and .py files processed by the raw loader.

Replacement done with:
```
for file in $(ls pkg/lib); do echo 'Processing' $file; base=$(basename $file) find pkg/ -type f -exec sed -i "s/$base/${base%.*}/g" {} +; done
```

And fixed up the rest manually.

----

This is @KKoukiou ' s large commit from #18447, I just massaged the commit message a bit. Bet's get this out of the way. It doesn't matter IMHO to have this change in main, even though we won't need it any more once the next esbuild gets released.

@KKoukiou , I approve code-wise, but you have to do the formal approval.


martinpitt:
Meh, nevermind -- this doesn't seem to work with webpack.

PR #18543, opened 2023-03-22
lib: Actually render dialog static_error as alert

mvollmer:
The type of static_error is string, so not wrapping it in a Alert is a bug.  There is only one user of this: cockpit-components-install-dialog.


martinpitt (approved):
Thanks!

PR #18544, opened 2023-03-22
metrics: use gradient for marking graph area when load/saturation is over 80% limit

KKoukiou:
n the gradient we don't use 80% but 70% since it takes 10% to actually transcend to the new color.

### DIFF
--- a/pkg/metrics/metrics.scss
+++ b/pkg/metrics/metrics.scss
@@ -499,6 +499,8 @@ $graphs: cpu, memory, disks, network;
     // Make SVG and generated before/after line position relative to metrics-data
     position: relative;
     grid-template-areas: "utilization";
+    // Make sure rows expand (WebKit needs this)
+    grid-template-rows: 1fr;
 
     // vertical ruler line for areas with data, and horizontal tick mark for utilization half
     &.valid-data::before {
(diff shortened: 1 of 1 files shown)


KKoukiou:
The new looks will be something like:
Screenshot from 2023-03-22 11-08-40

garrett:
Excellent! Thanks for splitting this out.

I'll work on the colors in general (they should probably be aligned to PF chart colors and ideally also be more colorblind friendly — I think these are still based on some colors I chose in the mockup and never got back to, as they look pretty nice otherwise), and also specifically the extreme amount in this PR.

garrett:
Tons of fixes... I still haven't actually addressed what this PR is about though. :wink: But it's all necessary to actually get there (tomorrow). And now the metrics (in this PR) works cross-browser again. (It doesn't in main. :cry:)

garrett:
How's this?

It's now using PF Chart colors, and I think I optimized it for color blindness. (I uses a color blind emulator for Protanopia, Deuteranopia, and Tritanopia — and also in normal colors.)

image

Oh, and this is WebKit. I got both it (which wasn't working at all for the graphs) and Chrome (which wasn't displaying them correctly) working in this PR also.

What's debatable is the color selection and which parts are which colors. We could try different palettes (which would need more work checking) or remap these.

marusak:
@KKoukiou want to take a look?

garrett (approved):
Looks good, although it's kind of weird to be approving my own work. But I'm approving on behalf of @KKoukiou approving my changes. :wink:

PR #18546, opened 2023-03-22
test: Pixel test the full shell key dialog

mvollmer:
The flakiness might have disappeared...


mvollmer:
This is just a hunch. I haven't been around when this test was added in #16486, but we have so few tests that need `ignore`, that I really would like to have even less.

KKoukiou (approved):
Looks green - I guess not needed anymore.

martinpitt (approved):
Right, thanks! Let's try this for a while.

PR #18548, opened 2023-03-23
shell: Fix manifest loading race condition

martinpitt:
All `<script>` tags get loaded in parallel, in particular manifests.js and shell.js. This means that Loader cannot rely on `cockpit.manifests` being initialized already -- if it runs before `manifests.js` is done with setting `cockpit.manifests`, then the initial `overlay.manifests` gets set to `{}` (the uninitialized default in cockpit.js).

But `{}` is truthy in JS, which made Loader think that the manifests were already initialized, and thus never actually requested it.

Fix the condition to also load manifests for an empty object to fix that.

-----

This was quite a learning exercise for me since yesterday :grin:  This was the symptom:

no-results

We never got a bug report about this happening in the wild, presumably because shell.js's structure just happened to be "slow" enough for manifests.js to always finish before `Loader` started. But esbuild apparently structures the bundle a bit differently, so that this [fails very often](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18447-20230322-065259-f76d494f-fedora-37/log.html) like [this screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18447-20230322-065259-f76d494f-fedora-37/TestPages-testBasic-fedora-37-127.0.0.2-2301-FAIL.png).

This was perfectly reproducible locally -- the shell nav menu was always empty after logging in, and started to work after a page reload (when manifests.js was cached).

Curiously, I was even able to reproduce this on main, when running

    ln -s dist cockpit
    XDG_DATA_DIRS=`pwd` /usr/libexec/cockpit-ws --no-tls -p 9998 --local-session=cockpit-bridge

Firefox' performance inspector was useful to confirm my suspicion of parallel loading:

js-load-parallel

I also added this fix to #18447, and in the [current test run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18447-20230323-072347-1f624673-fedora-37/log.html) at least *this* problem does not occur any more.

### DIFF
--- a/pkg/shell/machines/machines.js
+++ b/pkg/shell/machines/machines.js
@@ -550,7 +550,7 @@ function Loader(machines, session_only) {
         let open = local;
 
         let url;
-        if (!machine.manifests) {
+        if (!machine.manifests || Object.keys(machine.manifests).length === 0) {
             if (machine.checksum)
                 url = "../../" + machine.checksum + "/manifests.json";
             else


mvollmer:
I think there might be more going on here.  I thought scripts are executed in order, unless this is relaxed via the "async" or "defer" attribute:  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#notes

(They might be loaded in parallel, but should execute serially, I would have said.)

martinpitt:
@mvollmer Maybe you have a better idea -- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script doesn't say much about parallelization; I saw the `async` flag too, but the inspector is rather clear about what happens.

mvollmer:
@martinpitt, also, I think unless web workers are involved, there is no concurrent execution of JavaScript ever, so manifests.js does not run at the same time as shell.js, for example.

mvollmer:
> Curiously, I was even able to reproduce this on main, when running

I can't reproduce this. Are you sure you didn't use esbuild on main by accident somehow?

martinpitt:
> I can't reproduce this. Are you sure you didn't use esbuild on main by accident somehow?

I was on the correct branch, but perhaps the ~/.local/share/cockpit symlink still pointed at the esbuild branch at that time, and XDG_DATA_DIRS does not override this properly? So I'm not 100% sure

martinpitt:
Put this on hold, I have a fairly good theory what *really* happens.

PR #18549, opened 2023-03-23
patternfly-overrides: Remove obsolete/wrong overrides 

KKoukiou:
Patternfly handles it differently we margins and I don't see the reason to override it.

This was introduced with the massive PF3 -> PF4 rewrite and is not relevant any more.

This is expected to reduce slightly the space between the Primary and Cancel buttons in modals, which aligns with how these looks on the Patternfly site.

### DIFF
--- a/pkg/lib/patternfly/patternfly-4-overrides.scss
+++ b/pkg/lib/patternfly/patternfly-4-overrides.scss
@@ -118,28 +118,6 @@ svg {
   }
 }
 
-// Alignment of checks is incorrect and inconsistent across browsers
-// Upstream issue: https://github.com/patternfly/patternfly/issues/3868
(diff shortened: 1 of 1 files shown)


martinpitt:
Nice! Does this by any chance absorb some of the pixel diffs from esbuild, or is this completely independent?

KKoukiou:
> Nice! Does this by any chance absorb some of the pixel diffs from esbuild, or is this completely independent?

Yes I think it will absorb failures like https://cockpit-logs.us-east-1.linodeobjects.com/pull-18447-20230323-072347-1f624673-fedora-37/pixeldiff.html#TestBridge-testBasic-networking-bridge-settings-dialog-mobile were dependent from the flex before - but now not anymore.

PR #18554, opened 2023-03-24
c8s/pybridge test: Skip three regressed tests and ignore new crash

martinpitt:
Unfortunately PR #18398 broke three tests and introduced a new crash as unexpected message. We didn't notice back then as that PR didn't run c8s/pybridge, only f-37/pybridge. Accept this as new baseline for now, so that we don't fail all PRs, and avoid accidentaly introducing further regressions.

Example failures: [one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18491-20230323-144756-5d101ba4-centos-8-stream-pybridge/log.html), [two](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18552-20230324-084110-5d219b81-centos-8-stream-pybridge/log.html)

I created https://issues.redhat.com/browse/COCKPIT-971 to track actually fixing this.

### DIFF
--- a/test/common/testlib.py
+++ b/test/common/testlib.py
@@ -1804,6 +1804,29 @@ class MachineCase(unittest.TestCase):
                 "asyncio-ERROR: Task was destroyed but it is pending!",
                 "task:.*Task pending.*cockpit/channels/dbus.py.*"]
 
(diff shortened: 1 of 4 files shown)


allisonkarlitskaya (approved):
I'm looking into this again, but I honestly find it pretty confusing.  Let's merge it for now.

PR #18563, opened 2023-03-27
tests: use globbing when possible in skipImage decorator

KKoukiou:


KKoukiou:
Oh I just blindly sedded - it's just lists - this was supposed to be only for skipImage.

PR #18568, opened 2023-03-27
packagekit: don't call formatPackageId on itself

jelly:
After the refactoring in 8b3c35aaf0c6415c071 we now call formatPackageId twice which ends up showing a lot of undefined ( undefined ).

This still has some issues where it shows undefined in the progress bar because the first events we get are DOWNLOADING and not updating so we end up calling formatPackageKid with an empty string. However this requires an update to retrigger the issue so needs a quicker development cycle? Maybe our tests? However as this fixes some uglyness let's get this in first.


martinpitt (approved):
Whoops, thanks!

PR #18569, opened 2023-03-27
build: Move stylesheet processing into pkg/lib/build-common.js to enable reusing - FIX COMMIT MESSAGE ON MERGING

KKoukiou:
There are some quite delibarate hacks there, it worths to export these so that other projects can keep in sync with cockpit more easily.


martinpitt (approved):
Excellent idea, thank you!

martinpitt:
Bah, of course I forgot to fix the commit message..

PR #18570, opened 2023-03-27
pkg: Use only full path imports as tree shaking is imperfect

KKoukiou:
This affects the bundle size noticeable. In particular this change reducer the production bundle size from 25.9 mb to 25.3 mb.


martinpitt (approved):
Thanks!

PR #18573, opened 2023-03-28
systemd: set a different text color in the dark theme

jelly:
The current version is barely readable in the dark theme.

Before:

image

After:

image

### DIFF
--- a/pkg/systemd/service-details.scss
+++ b/pkg/systemd/service-details.scss
@@ -88,6 +88,10 @@
   padding-left: 1em;
 }
 
+.pf-theme-dark .side-note {
+  color: var(--pf-global--Color--200);
+}
+
 .font-xs {
   font-size: var(--pf-global--FontSize--xs);
 }


martinpitt (dismissed):

KKoukiou (changes requested):
The proper solution is to use <Text component={TextVariants.small}> which work as expected in dark theme and then we also don't need the rest of custom CSS for font size, color and such.

jelly:
> The proper solution is to use which work as expected in dark theme and then we also don't need the rest of custom CSS for font size, color and such.

I think you mean using the correct PF component instead of our custom div/css's?

KKoukiou:
> > The proper solution is to use which work as expected in dark theme and then we also don't need the rest of custom CSS for font size, color and such.
> 
> I think you mean using the correct PF component instead of our custom div/css's?

Heh the text inside `<` `>` in markdown got hidden. Fixed.

jelly:
> The proper solution is to use `<Text component={TextVariants.small}>` which work as expected in dark theme and then we also don't need the rest of custom CSS for font size, color and such.

Our <span has `--pf-global--FontSize--xs` for which the Text component does not have, it's smallest is `small` which is bigger then `xs`. The HTML variant does have xs https://www.patternfly.org/v4/utilities/text/#font-size 

So I guess we first have to create an PF issue and then we can use the `Text` component. Aside from that, it should use a  different color then the default `Text` and Flex for layout.

image

KKoukiou:
> > The proper solution is to use `<Text component={TextVariants.small}>` which work as expected in dark theme and then we also don't need the rest of custom CSS for font size, color and such.
> 
> Our <span has `--pf-global--FontSize--xs` for which the Text component does not have, it's smallest is `small` which is bigger then `xs`. The HTML variant does have xs https://www.patternfly.org/v4/utilities/text/#font-size
> 
> So I guess we first have to create an PF issue and then we can use the `Text` component. Aside from that, it should use a different color then the default `Text` and Flex for layout.
> 
> image

We can keep the font-xs class, that's not a problem. I am just proposing to not add this extra custom CSS for background color - when PF supports that fine.

Issue #18463, opened 2023-03-07
Installation progress looks strange in the apps page


jelly:
### Explain what happens

The percentage should be next to the number (61) and maybe the progress bar should be bigger?

![image](https://user-images.githubusercontent.com/67428/223475014-7fe3bb96-03d0-48fd-b113-30826fb46456.png)

@garrett just for verification, how should it exactly look like?

Update: 
Discussion led to https://github.com/cockpit-project/cockpit/issues/18463#issuecomment-1459789817
The progress bar text should be put on the left of the bar, the bar should have inline style and limited width.

garrett:
Is it ever cancellable? If not, then we shouldn't show that action. And it should be a secondary button, not a link, if you can cancel it at some point (and a disabled button if it is supposed to be possible at some human-clickable point in time). So a disabled button should still look like a button.

I don't think it's realistically cancelable though. I suppose when it's downloading before installing... but that's such a small window of time, especially on a server. We're talking about downloading several megabytes at most. Which, in these days, on a server, is really absurdly quick. Meaning that it would be possible to cancel the download for maybe quarter of a second? Nobody's that fast to click the button, unless they're Superman or they had their mouse cursor over the area. As that's where the install button would've been, I think, then that brings up another problem of someone accidentally clicking the same area twice. (It's an accessibility problem, especially for some people. But everyone can accidentally double click.) So I think canceling is useless and could even be a misfeature here. It's probably better to remove it and rely on the remove button?

Otherwise, it's basically supposed to be a progress bar. And the percentage should be a fixed width with tabular numbers, else it'll make the bar bounce around.

So, kinda like this:

![progress](https://user-images.githubusercontent.com/10246/223480367-25c77e5b-a42f-4e49-9c7a-d4d843a7243a.png)

jelly:
Oh, for me the install process takes quite a while on a VM. I haven't tested cancellation, will have to see if that's broken or not.

jelly:
This also looks.. not 100% aligned:

![Screenshot from 2023-03-07 17-59-34](https://user-images.githubusercontent.com/67428/223494266-05567fa4-dcf6-4630-8917-271716db7063.png)

jelly:
Re-trying today and it all seems to work...

![image](https://user-images.githubusercontent.com/67428/223662122-fda44794-3e5b-49ad-b048-13a60f0548de.png)

garrett:
[...]

PR #18575, opened 2023-03-28
pkg/apps: Fix progress bar and cancel button

ahmedr2001:
Progress bar text is now on left of bar, bar has inline style and limited width.
Screenshot from 2023-03-28 14-48-25

This fixes issue #18463.

### DIFF
--- a/pkg/apps/application-list.jsx
+++ b/pkg/apps/application-list.jsx
@@ -47,7 +47,11 @@ const ApplicationRow = ({ comp, progress, progress_title, action }) => {
 
     let summary_or_progress;
     if (progress) {
-        summary_or_progress = <ProgressBar title={progress_title} data={progress} />;
+        summary_or_progress = (
+            <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
+                <span id={comp.name + "-progress"} className="progress-title-span">{progress_title}</span>
+                <ProgressBar title={progress_title} data={progress} ariaLabelledBy={comp.name + "-progress"} />
+            </Flex>);
     } else {
         if (error) {
             summary_or_progress = (
--- a/pkg/apps/application.scss
+++ b/pkg/apps/application.scss
@@ -22,6 +22,12 @@
   align-items: center;
 }
 
+.progress-bar {
+  display: inline-block;
+  width: 20%;
+  overflow: hidden;
+}
+
 /* Application */
 
 .application-details img {
--- a/pkg/apps/utils.jsx
+++ b/pkg/apps/utils.jsx
@@ -20,7 +20,7 @@
 import cockpit from "cockpit";
 import React from "react";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
-import { Progress } from "@patternfly/react-core/dist/esm/components/Progress/index.js";
+import { Progress, ProgressMeasureLocation } from "@patternfly/react-core/dist/esm/components/Progress/index.js";
 import { Split, SplitItem } from "@patternfly/react-core/dist/esm/layouts/Split/index.js";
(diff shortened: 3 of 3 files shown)


pkg/apps/application-list.jsx
```diff
@@ -47,7 +47,11 @@ const ApplicationRow = ({ comp, progress, progress_title, action }) => {
 
     let summary_or_progress;
     if (progress) {
-        summary_or_progress = <ProgressBar title={progress_title} data={progress} />;
+        summary_or_progress = (
+            <div className="sum-or-prog">
```
KKoukiou, line 51: This should be better replace with:

`<Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>`

pkg/apps/application.scss
```diff
@@ -22,6 +22,21 @@
   align-items: center;
 }
 
+.progress-bar {
+  display: inline-block;
+  width: 20%;
+  overflow: hidden;
+}
+
+.sum-or-prog {
```
KKoukiou, line 31: THis can now be removed.

pkg/apps/application.scss
```diff
@@ -22,6 +22,21 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.progress-title-span {
```
KKoukiou, line 36: Please remove this as well, also margin-right only works for LTR (Left to right) languages. For RTL languages like hebrew, this does not work, as the whole page is swaped in place.

See for example this margin missing in Hebrew.

Screenshot from 2023-03-28 21-29-57

KKoukiou: By using the suggested Flex this will not just be fixed, but it's about to be fixed upstream in patternfly for the Flex component by using `gap` instead of `margin`

pkg/apps/utils.jsx
```diff
@@ -60,12 +60,12 @@ export const ProgressBar = ({ size, title, data }) => {
             </SplitItem>
         </Split>);
     } else {
-        return <Progress value={data.percentage} size={size} title={title} />;
```
KKoukiou, line 63: I am not sure if you 're famiilar with the term of accesibility. It's essentially patterns that we use so that people with eyesight problems can still use our product. 

The problem with removing the title outside of the component is that you make it not accesible any more.

In order to make it accessible again while keeping the title outside is to use the 'aria-labelledby' as suggested in the documentation here https://www.patternfly.org/v4/components/progress

The aria-labeledby will have value the 'id' of the new 'title' element. So add a new property to the ProgressBar component here, called ariaLabeledBy where you pass the 'id' of the title element (create an id if that does not exist)

KKoukiou (changes requested):
Great job! Some feedback, but UI wise looks really great.

ahmedr2001:
> Great job! Some feedback, but UI wise looks really great.

Resolved

pkg/apps/application-list.jsx
```diff
@@ -47,7 +47,11 @@ const ApplicationRow = ({ comp, progress, progress_title, action }) => {
 
     let summary_or_progress;
     if (progress) {
-        summary_or_progress = <ProgressBar title={progress_title} data={progress} />;
+        summary_or_progress = (
+            <Flex spaceItems={{ default: 'spaceItemsSm' }} alignItems={{ default: 'alignItemsCenter' }}>
+                <span id={progress_title} className="progress-title-span">{progress_title}</span>
```
KKoukiou, line 52: id's needs to be unique per document. progress title is something like 'removing'. You need to make it more specific.

How about {comp.name + "-progress"}

KKoukiou (changes requested):

KKoukiou (approved):

ahmedr2001:
How's this PR going so far?

garrett:
:+1: Looks good! Thanks!

screenshot of progress

garrett (approved):

PR #18577, opened 2023-03-28
Kdump expendable ModalError

jelly:
Previous: (without expandable :) )
image

New:
Screenshot from 2023-03-28 21-19-42

### DIFF
--- a/pkg/kdump/kdump-view.jsx
+++ b/pkg/kdump/kdump-view.jsx
@@ -125,9 +125,9 @@ const KdumpSettingsModal = ({ settings, initialTarget, handleSave }) => {
                 .catch(error => {
                     if (error.details) {
                         // avoid bad summary like "systemd job RestartUnit ["kdump.service","replace"] failed with result failed"
-                        // if we have a more concrete journal
+                        // if we have a more concrete journal and trim journal's `kdump: ` prefix.
                         error.message = _("Unable to save settings");
-                        error.details = <CodeBlockCode>{ error.details }</CodeBlockCode>;
+                        error.details = <CodeBlockCode>{ error.details.replaceAll(/\nkdump: /g, "\n") }</CodeBlockCode>;
                         setError(error);
                     } else {
                         // without a journal, show the error as-is
--- a/pkg/kdump/kdump-view.jsx
+++ b/pkg/kdump/kdump-view.jsx
@@ -156,7 +156,8 @@ const KdumpSettingsModal = ({ settings, initialTarget, handleSave }) => {
                        </Button>
                    </>
                }>
-            {error && <ModalError dialogError={error.message || error}
+            {error && <ModalError isExpandable
+                                  dialogError={error.message || error}
                                   dialogErrorDetail={error?.details} />}
             <Form id="kdump-settings-form" isHorizontal>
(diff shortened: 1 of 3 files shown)


pkg/kdump/kdump-view.jsx
```diff
@@ -127,7 +127,7 @@ const KdumpSettingsModal = ({ settings, initialTarget, handleSave }) => {
                         // avoid bad summary like "systemd job RestartUnit ["kdump.service","replace"] failed with result failed"
                         // if we have a more concrete journal
                         error.message = _("Unable to save settings");
-                        error.details = <CodeBlockCode>{ error.details }</CodeBlockCode>;
+                        error.details = <CodeBlockCode>{ error.details.replaceAll("kdump: ", "") }</CodeBlockCode>;
```
garrett, line 130:
```suggestion
                        error.details = <CodeBlockCode>{ error.details.replaceAll(/^kdump: /, "") }</CodeBlockCode>;
```

Would this be better to make sure `kdump: ` a regular expression and have it match the beginning of lines? (It _probably_ doesn't matter, though. And I didn't check to see if this actually works. :sweat_smile:)

garrett (dismissed):
Looks good from the screenshot. It's probably fine once you address Katerina's comment. Also, perhaps consider my comment (if it even works here).

But I do like the idea (as I suggested it :wink:) and it looks like it works, according to the screenshot. :grin:

pkg/kdump/kdump-view.jsx
jelly: Yes, this is a bit sloppy as it removes one extraneous `kdump: `. in the middle of the line.

jelly: right, but in some more effort and now it doesn't strip extranous `kdump: `.

Screenshot from 2023-03-29 10-06-28

jelly:
> Looks good from the screenshot. It's probably fine once you address Katerina's comment. Also, perhaps consider my comment (if it even works here).
> 
> But I do like the idea (as I suggested it wink) and it looks like it works, according to the screenshot. grin

As a follow up we should probably look into other ModalError users where we can also use this new isExpandable property.

garrett:
> As a follow up we should probably look into other ModalError users where we can also use this new isExpandable property.

Definitely.

Once concern I have is if the `>` icon is enough. We could have some expander action link that also expands it (and an ignore one that does the same as the `x` icon, for balance, I suppose). It probably also depends on how important the additional supporting text is... in some cases, just hiding it, but making it _possible_ to show the extra text would be enough.

jelly:
> > As a follow up we should probably look into other ModalError users where we can also use this new isExpandable property.
> 
> Definitely.
> 
> Once concern I have is if the `>` icon is enough. We could have some expander action link that also expands it (and an ignore one that does the same as the `x` icon, for balance, I suppose). It probably also depends on how important the additional supporting text is... in some cases, just hiding it, but making it _possible_ to show the extra text would be enough.

I checked and the expanded state of the Alert is [internal](https://github.com/patternfly/patternfly-react/blob/v5/packages/react-core/src/components/Alert/Alert.tsx#L179) and can't be controlled from outside. So we can't add  a `view details` button which expands the alert.

PR #18579, opened 2023-03-29
various: replace mentions of IRC with Matrix

allisonkarlitskaya:
The Cockpit Project is now on Matrix at #cockpit:fedoraproject.org.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -350,7 +350,9 @@ unrelated to the commit message.
 Cockpit is a designed project. Anything that the user will see should have
 design done first. This is done on the wiki and mailing list.
 
-Bigger changes need to be discussed on the #cockpit IRC channel or our mailing list
+Bigger changes need to be discussed on the
+[#cockpit:fedoraproject.org](https://matrix.to/#/#cockpit:fedoraproject.org)
+Matrix channel or our mailing list
 [cockpit-devel@lists.fedoraproject.org](https://lists.fedorahosted.org/admin/lists/cockpit-devel.lists.fedorahosted.org/)
 before you invest too much time and energy.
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -350,7 +350,9 @@ unrelated to the commit message.
 Cockpit is a designed project. Anything that the user will see should have
 design done first. This is done on the wiki and mailing list.
 
-Bigger changes need to be discussed on the #cockpit IRC channel or our mailing list
+Bigger changes need to be discussed on the
+[#cockpit:fedoraproject.org](https://matrix.to/#/#cockpit:fedoraproject.org)
```
martinpitt, line 354: Is that really what we should use? Not https://chat.fedoraproject.org/#/room/#cockpit:fedoraproject.org ?

allisonkarlitskaya: As far as I can tell, this takes you to a login page which only works for people who are fedoraproject members.

The other one gives you a choice of which client and/or service you'd like to use.

allisonkarlitskaya: But there's an adjacent grammar error I should fix :wink:

allisonkarlitskaya: > But there's an adjacent grammar error I should fix wink

Ah no.  The line after it makes it correct.

martinpitt: I just tested this with my gh account, and it works. With "my" link you are kind of forced to have a Fedora account. So that seems better.

martinpitt:
So :+1 from me once we officially decide to do this.

garrett (approved):
:+1: LGTM. Good for a merge, once we make the decision.

PR #18580, opened 2023-03-29
Update stratis pixel tests

KKoukiou:
No obvious changes apart from the image dimension.


KKoukiou:
I can't really explain this change. I would make sence if this test did not run for a while, (some naughty) and the new image fixed it so it would execute the pixel test assestion code. However I was not able to find any naughty on f37 for stratis.

martinpitt (approved):
Needs to land in lockstep with https://github.com/cockpit-project/bots/pull/4578 . If it works, okay -- :man_shrugging: "pick your battles"..

mvollmer (changes requested):
Those `stratis-1-private` block devices should not appear in that dialog, or elsewhere in the Cockpit UI.

PR #18582, opened 2023-03-30
test: Stop hardcoding busybox container image name

martinpitt:
https://github.com/cockpit-project/bots/pull/4585 updates the repo for it, and renames it to "localhost/test-busybox". Stop hardcoding the full name, and check for available images instead.


martinpitt:
Triggering two extra runs against the refreshed RHEL images in https://github.com/cockpit-project/bots/pull/4585

jelly (approved):

PR #18583, opened 2023-03-30
tests: Introduce assertion for stratis private objects wrongly shown in block devices

KKoukiou:


jelly (approved):
Approved assuming F38 bots PR also turns green

PR #18584, opened 2023-03-30
test: Fix qunit test path and missing pytest-asyncio

martinpitt:
Commit 18576f7d18e moved them to /qunit. Adjust the pytest wrapper, documentation and example accordingly. This brings back the HTML tests in the pytest unit test run.

----

Oops! See a [recent pytest-cov run](https://github.com/cockpit-project/cockpit/actions/runs/4551701495/jobs/8026098929?pr=18577#step:4:280), there are no HTML tests.

### DIFF
--- a/containers/unit-tests/setup.sh
+++ b/containers/unit-tests/setup.sh
@@ -43,6 +43,7 @@ dependencies="\
     pkg-config \
     python3 \
     python3-mypy \
+    python3-pytest-asyncio \
     python3-pytest-cov \
     python3-pytest-timeout \
     ssh \
--- a/HACKING.md
+++ b/HACKING.md
@@ -42,7 +42,7 @@ edit files as you normally would. Building and running tests happens inside the
 toolbox container. If desired, you can install additional packages with
 `sudo dnf install`.
 
-The Cockpit team occasionally refreshes the `tasks` container image. 
+The Cockpit team occasionally refreshes the `tasks` container image.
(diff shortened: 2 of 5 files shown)


martinpitt:
Hmm.. the [asyncio warnings are gone](https://github.com/cockpit-project/cockpit/actions/runs/4564542499/jobs/8054447524?pr=18584), but still no sight of the qunit tests.

containers/unit-tests/setup.sh
```diff
@@ -43,6 +43,7 @@ dependencies="\
     pkg-config \
     python3 \
     python3-mypy \
+    python3-pytest-asyncio \
```
allisonkarlitskaya, line 46: Can we configure pytest to error out for stuff like this instead of silently passing?

allisonkarlitskaya (approved):
Looks good, thanks 👍

See one comment below.

containers/unit-tests/setup.sh
martinpitt: I'll add that to my todo list .. that, or file a bug that it should just happen. The current behaviour is really not okay.

allisonkarlitskaya: I've done a bit of research.

See https://github.com/pytest-dev/pytest/issues/5023 for the suggested solution (`--strict-markers`) as well as links to lots of other discussion about why it's not so easy.  I'll do a patch to add `--strict-markers`.

Issue #11876, opened 2019-05-20
File ownership changed when removing authorized pubkey from user


mvn23:
Cockpit version: 188, 194
OS: 188 on Debian Stretch, 194 on Arch Linux
Page: "Accounts"

When removing a key from the list of `Authorized Public SSH Keys` of a user, that user's `~/.ssh/authorized_keys` file becomes owned by `root:root`. If the file is not world-readable (or group-readable if the user is a member of group `root`), this breaks that user's ssh key login.

Steps to reproduce
1. Make sure the target user has a file `~/.ssh/authorized_keys` with more than 1 key in it.
2. Check that this file is owned by the target user and group.
3. Remove one of the keys through Cockpit>Accounts>[target user]>Authorized Public SSH Keys.
4. Check ownership of the target user's `~/.ssh/authorized_keys`.

Expected result:
File ownership remains unchanged.

Actual result:
File is now owned by `root:root`.

marusak:
Indeed, this is issue that is still happening. Verified with current master on Fedora 31

marusak:
Closing in favor of https://github.com/cockpit-project/cockpit/issues/18033 which is more generic and describes exactly this issue.

martinpitt:
I un-duplicate this, as the generic fix was not approved. I'll do a targeted but ugly fix for this specific issue, and then think about improving cockpit.file() after that.

PR #18585, opened 2023-03-30
bridge: Restore file owner/permissions in cockpit.file().replace()

### LINKED ISSUE #18033: cockpit.file().replace() doesn't preserve owner/group/mode
### Explain what happens

https://github.com/cockpit-project/cockpit/blob/27cb665b5c135481f900dafac0c1b754ab91b5a0/src/bridge/cockpitfsreplace.c#L270
replace create a new world readable file, this is a problem when using replace() to write any kind of sensitive data.
Developer can call chmod just after but this leave a small insecure window.
Developer could also miss the fact that replace just changed the permissions and leave everything open.
Idealy it would preserve:
- owner
- group
- mode
- selinux (bonus)

This already cases issues, see https://github.com/cockpit-project/cockpit/issues/11876 where cockpit changes owner on `~/.ssh/authorized_keys`. This component uses `file.modify()` which calls `file.replace()`. Or issue #19905 where editing motd makes it inaccessible.

martinpitt:
If the file already exists, restore its original owner, group, and    
permissions after renaming the temp file. This is still not perfect    
(e.g. SELinux context and xattrs), but at least a whole lot better.    
    
The reported example in issue #11876 was that Cockpit broke the file    
permissions of ~user/.ssh/authorized_keys after removing a key (to    
root:root 0644). Add an explicit integration check for that, as that    
also covers restoring the proper owner.    
    
Fixes #18033

### DIFF
--- a/src/cockpit/channels/filesystem.py
+++ b/src/cockpit/channels/filesystem.py
@@ -188,12 +188,22 @@ class FsReplaceChannel(Channel):
             if self._tag and self._tag != tag_from_path(self._path):
                 raise ChannelError('change-conflict')
 
+            # if the file exists already, keep its permissions
+            try:
+                stat = os.stat(self._path)
+            except FileNotFoundError:
+                pass
+            else:
+                os.chmod(self._temppath, stat.st_mode)
+                os.chown(self._temppath, stat.st_uid, stat.st_gid)
+
             try:
                 os.rename(self._temppath, self._path)
             except OSError:
                 # ensure to not leave the temp file behind
                 self.unlink_temppath()
                 raise
+
             self._tempfile.close()
             self._tempfile = None
 
--- a/pkg/base1/test-file.js
+++ b/pkg/base1/test-file.js
@@ -65,6 +65,17 @@ QUnit.test("simple replace", async assert => {
     await cockpit.file(dir + "/bar").replace("4321\n");
     const res = await cockpit.spawn(["cat", dir + "/bar"]);
     assert.equal(res, "4321\n", "correct content");
+    // keeps existing permissions; as we don't want to assume an umask, test with a narrow and a wide scenario
+    // we cannot test owner changes here, that's done in integration tests
+    for (const mode of ["600", "777"]) {
+        await cockpit.spawn(["chmod", mode, dir + "/bar"]);
+        const content = `"new-${mode}\n`;
+        await cockpit.file(dir + "/bar").replace(content);
+        const res = await cockpit.spawn(["cat", dir + "/bar"]);
+        assert.equal(res, content, "correct content after overwriting");
+        const stat = await cockpit.spawn(["stat", "-c", "%a", dir + "/bar"]);
+        assert.equal(stat.trim(), mode, "correct permissions");
+    }
 });
 
 QUnit.test("stringify replace", async assert => {
--- a/src/bridge/cockpitfsreplace.c
+++ b/src/bridge/cockpitfsreplace.c
@@ -177,11 +177,28 @@ cockpit_fsreplace_control (CockpitChannel *channel,
             {
               new_tag = cockpit_get_file_tag (self->tmp_path);
               json_object_set_string_member (options, "tag", new_tag);
+
+              // if the file exists already, keep its permissions
+              struct stat st;
+              if (stat (self->path, &st) == 0) {
+                if (chmod (self->tmp_path, st.st_mode) < 0)
+                  {
+                    close_with_errno (self, "couldn't chmod", errno);
+                    goto out;
+                  }
+                if (chown (self->tmp_path, st.st_uid, st.st_gid) < 0)
+                  {
+                    close_with_errno (self, "couldn't chown", errno);
+                    goto out;
+                  }
+              }
+
               if (rename (self->tmp_path, self->path) < 0)
                 {
                   close_with_errno (self, "couldn't rename", errno);
                   goto out;
                 }
+
             }
         }
     }
--- a/test/verify/check-shell-keys
+++ b/test/verify/check-shell-keys
@@ -97,17 +97,20 @@ class TestKeys(MachineCase):
 
         b.wait_in_text("#account-authorized-keys", "no authorized public keys")
 
+        def check_perms():
+            perms = m.execute("getfacl -a /home/user/.ssh")
+            self.assertIn("owner: user", perms)
+
+            perms = m.execute("getfacl -a /home/user/.ssh/authorized_keys")
+            self.assertIn("owner: user", perms)
+            self.assertIn("user::rw-", perms)
+            self.assertIn("group::---", perms)
+            self.assertIn("other::---", perms)
+
         # Adding keys sets permissions properly
         b.wait_text("#account-user-name", "user")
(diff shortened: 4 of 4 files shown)


src/cockpit/channels/filesystem.py
```diff
@@ -188,12 +188,23 @@ def do_done(self):
...
             try:
                 os.rename(self._temppath, self._path)
             except OSError:
                 # ensure to not leave the temp file behind
                 self.unlink_temppath()
                 raise
+            # restore original permissions
+            if stat is not None:
+                os.chmod(self._path, stat.st_mode)
+                os.chown(self._path, stat.st_uid, stat.st_gid)
```
martinpitt, line 206: Hmm, now that I look at this, I think we should do this on `temppath` instead, to get a better failure mode if the permission restoration fails.

martinpitt: Done.

allisonkarlitskaya (commented):
I'm sorry.  I simply don't agree with this.

If we want to support this then we need to either:

 a) find an option to clone an inode then trunc it, write to it, then rename it back in place; or
 b) let the caller choose their own poison: add a mode that enables trunc-based replacement

martinpitt:
> I'm sorry. I simply don't agree with this.

On the grounds that "much less broken" isn't better than "leaving it fully broken"? (I'm honestly interested in the "why")

> a) find an option to clone an inode then trunc it, write to it, then rename it back in place; or

If there's an API for that, that would be cool.

> b) let the caller choose their own poison: add a mode that enables trunc-based replacement

We can't always predict file permissions, especially not when updating files in /etc/ or in this case, HOME. These are under the control of the admin, umask, etc. This can only work through calling stat() first, which we don't even have an API for. I.e. `cockpit.system(["stat", ...])`. This is just awful, and surely not better than doing it in the bridge. So a strong veto on that option.

martinpitt:
Another option would be to not fix #18033 at all, and do a targetted fix for #11876 by not using cockpit.file() (as that's simply not up to the task), and instead update authorized_keys with a little shell helper similar to [ssh-add-public-key.sh](https://github.com/cockpit-project/cockpit/blob/main/pkg/users/ssh-add-public-key.sh). That feels somewhat unsatisfying, though.

allisonkarlitskaya:
I agree that leaving this unfixed is somehow unsatisfying, but on the other hand:
 - nobody thought this was a problem until now
 - this is a huge behaviour change in a widely-used API and may have unintended side-effects
 - omg the hax

I'd definitely prefer a more targeted solution.

allisonkarlitskaya:
Take a look at `ssh-copy-ids` from openssh, for reference/inspiration.  In particular:

```sh
# installkey_sh [target_path]
#    produce a one-liner to add the keys to remote authorized_keys file
#    optionally takes an alternative path for authorized_keys
installkeys_sh() {
  AUTH_KEY_FILE=${1:-.ssh/authorized_keys}
  AUTH_KEY_DIR=$(dirname "${AUTH_KEY_FILE}")

  # In setting INSTALLKEYS_SH:
  #    the tr puts it all on one line (to placate tcsh)
  #      (hence the excessive use of semi-colons (;) )
  # then in the command:
  #    cd to be at $HOME, just in case;
  #    the -z `tail ...` checks for a trailing newline. The echo adds one if was missing
  #    the cat adds the keys we're getting via STDIN
  #    and if available restorecon is used to restore the SELinux context
  INSTALLKEYS_SH=$(tr '\t\n' ' ' <<-EOF
        cd;
        umask 077;
        mkdir -p "${AUTH_KEY_DIR}" &&
                { [ -z \`tail -1c ${AUTH_KEY_FILE} 2>/dev/null\` ] ||
                        echo >> "${AUTH_KEY_FILE}" || exit 1; } &&
                cat >> "${AUTH_KEY_FILE}" || exit 1;
        if type restorecon >/dev/null 2>&1; then
                restorecon -F "${AUTH_KEY_DIR}" "${AUTH_KEY_FILE}";
        fi
        EOF
  )

  # to defend against quirky remote shells: use 'exec sh -c' to get POSIX;
  printf "exec sh -c '%s'" "${INSTALLKEYS_SH}"
}
```

allisonkarlitskaya:
I actually start to think that the problem is that we have a modify operation.  I wonder if we could drop that.  This would make this all conceptually much simpler.

allisonkarlitskaya:
> I actually start to think that the problem is that we have a modify operation. I wonder if we could drop that. This would make this all conceptually much simpler.

Specifically: POSIX has a modify operation.  It's called `ftruncate(); write();`.  We don't want to do that, so we try to model that with what is really a file creation operation.  Then we try and paste some extra stuff on to make it seem like it actually *was* a modify operation, after all.

I think, high-level view, I'd even go further than my previous statement and say:
 - either we implement modify via `ftruncate(); write();`; or
 - we don't have modify — it's up to the caller to read the old contents (if applicable) and write the new one.  All both users would be a bit more complicated, but we'd lose this problem.

martinpitt:
> it's up to the caller to read the old contents (if applicable) and write the new one

The API is designed around "tags" to prevent TOCTOU races -- i.e. reading the file, then it gets changed by something else, and writing it back based on the old contents. .modify() DTRT there. I'm not saying that this is very important for the above use cases, just what the original creators had in mind.

Note that .modify() only inherits the "bad permissions" problem -- it exists already at the level of .replace(). So even if we were to drop .modify(), it wouldn't fix the one place which is actually broken. But given that we don't know about all API users, we can't just simply drop the API IMHO. This was already questionable for cockpit.dbus().publish(), but the file API is much less obscure.

allisonkarlitskaya:
> The API is designed around "tags" to prevent TOCTOU races

This is what I meant by "more complicated".

> Note that .modify() only inherits the "bad permissions" problem -- it exists already at the level of .replace(). But given that we don't know about all API users, we can't just simply drop the API IMHO. This was already questionable for cockpit.dbus().publish(), but the file API is much less obscure.

We can declare it deprecated, however, and use the replace API.


As a general statement (and relevant to what I'm working on right now as well) I notice that we have an awful lot of shellscript fragments around Cockpit that related to listing, adding, or removing ssh public keys to `authorized_keys` and/or `known_hosts` files:

 - `pkg/lib/credentials-ssh-private-keys.sh`
 - `pkg/lib/credentials-ssh-remove-key.sh`
 - `pkg/shell/machines/ssh-add-key.sh` (used with both `known_hosts` and `authorized_keys`)
 - `pkg/shell/machines/ssh-show-default-key.sh`
 - `pkg/systemd/overview-cards/ssh-list-host-keys.sh`
 - `pkg/users/ssh-add-public-key.sh`
 - `pkg/users/ssh-list-public-keys.sh`

I wonder if it makes sense to unify some of these and/or use them for what you're trying to accomplish here.  Why does this one have to be via the file API when we have existing "better" mechanisms that we already use in other places?

martinpitt:
> Why does this one have to be via the file API when we have existing "better" mechanisms that we already use in other places?

It seems our idea of a good API differs somewhat :grin: In my eyes, these shell scripts are hacky workarounds *because* the cockpit.file() API is so bad... I'd much rather use the file API for all of these.

That said, I agree that it would make sense to unify these and move to pkg/lib/.

PR #18588, opened 2023-03-31
ci: trigger differential-shellcheck workflow on push

jamacku:
Also, update `differential-shellcheck` to the latest version (`v4.0.2`) - https://github.com/redhat-plumbers-in-action/differential-shellcheck/releases Push events are supported since `v4.0.0`.

Fixes: https://github.com/redhat-plumbers-in-action/differential-shellcheck/issues/215

This should get rid of annoying messages from the `github-code-scanning` bot.

Screenshot from 2023-03-22 07-00-32

Also, I have added the required permission for uploading SARIF reports to GitHub

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -3,6 +3,8 @@
 
 name: Differential ShellCheck
 on:
+  push:
+    branches: [main, rhel-*]
   pull_request:
     branches: [main, rhel-*]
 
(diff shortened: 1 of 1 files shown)


jamacku:
Yes, I understand the reason. I had the same configuration on many of my projects, but the new GitHub Feature assumes that every scanning tool reports results on `push`. I hope the behavior will change, but so far, I didn't notice any update.

martinpitt (approved):

PR #18589, opened 2023-03-31
lib: port cockpit.user to use standard promises

jelly:
If I understood it correctly, we completely want to drop `cockpit.defer`. This seems a bit tricky in the other cases like the DBusClient and cockpit.http, but seems doable for `cockpit.file` as a start :)

## Warning: this breaks API

We now return a proper Promise instead of an `function`. So this requires some code fixing as with `useEffect` in 1f5724d4e9a707b65279f49fa1bced2dae877f34 and ofcourse we can't use `.done()` or `.fail()`.

I want to land this tomorrow after the release and then on Thursday cockpit-lib-update runs and I can deal with the potential fallout in podman, machines, .etc.

### DIFF
--- a/pkg/lib/cockpit.js
+++ b/pkg/lib/cockpit.js
@@ -2391,29 +2391,25 @@ function factory() {
 
     let the_user = null;
     cockpit.user = function () {
-        const dfd = cockpit.defer();
-        if (!the_user) {
-            const dbus = cockpit.dbus(null, { bus: "internal" });
-            dbus.call("/user", "org.freedesktop.DBus.Properties", "GetAll",
-                      ["cockpit.User"], { type: "s" })
-                .then(([user]) => {
-                    the_user = {
-                        id: user.Id.v,
-                        name: user.Name.v,
-                        full_name: user.Full.v,
-                        groups: user.Groups.v,
-                        home: user.Home.v,
-                        shell: user.Shell.v
-                    };
-                    dfd.resolve(the_user);
-                })
-                .catch(ex => dfd.reject(ex))
-                .finally(() => dbus.close());
-        } else {
-            dfd.resolve(the_user);
-        }
-
-        return dfd.promise;
+            if (!the_user) {
+                const dbus = cockpit.dbus(null, { bus: "internal" });
+                return dbus.call("/user", "org.freedesktop.DBus.Properties", "GetAll",
+                          ["cockpit.User"], { type: "s" })
+                    .then(([user]) => {
+                        the_user = {
+                            id: user.Id.v,
+                            name: user.Name.v,
+                            full_name: user.Full.v,
+                            groups: user.Groups.v,
+                            home: user.Home.v,
+                            shell: user.Shell.v
+                        };
+                        return the_user;
+                    })
+                    .finally(() => dbus.close());
+            } else {
+                return Promise.resolve(the_user);
+            }
     };
 
     /* ------------------------------------------------------------------------
--- a/pkg/shell/hosts.jsx
+++ b/pkg/shell/hosts.jsx
@@ -70,7 +70,7 @@ export class CockpitHosts extends React.Component {
     componentDidMount() {
         cockpit.user().then(user => {
             this.setState({ current_user: user.name || "" });
-        });
+        }).catch(exc => console.log(exc));
     }
 
     static getDerivedStateFromProps(nextProps, prevState) {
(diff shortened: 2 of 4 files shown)


jelly:
Ok, this is annoying as it breaks our API. useEffect(() => cockpit.user()) now fails as it returns a promise? or at least not a function.

@martinpitt just to be clear, do we want to continue in this direction?

martinpitt:
@jelly Heh, I tried the same a few weeks ago, and ran into the exact same thing. I didn't manage to fully understand it, so I left it alone. But yes, I agree we should do this eventually, and get rid of such weird magic.

pkg/shell/hosts.jsx
```diff
@@ -70,7 +70,7 @@ export class CockpitHosts extends React.Component {
     componentDidMount() {
         cockpit.user().then(user => {
             this.setState({ current_user: user.name || "" });
-        });
+        }).catch(exc => console.log(exc));
```
cockpituous, line 73: This added line is not executed by any test.

pkg/lib/cockpit.js
```diff
@@ -2391,29 +2391,28 @@ function factory() {
...
-                    dfd.resolve(the_user);
-                })
-                .catch(ex => dfd.reject(ex))
-                .finally(() => dbus.close());
-        } else {
-            dfd.resolve(the_user);
-        }
-
-        return dfd.promise;
+        return new Promise((resolve, reject) => {
```
martinpitt, line 2394: We don't need this extra promise wrapper at all -- it could just directly return the `dbus.call()` promise (chained with .finaly for dbus.close()).

pkg/lib/cockpit.js
```diff
@@ -2391,29 +2391,28 @@ function factory() {
...
+                            id: user.Id.v,
+                            name: user.Name.v,
+                            full_name: user.Full.v,
+                            groups: user.Groups.v,
+                            home: user.Home.v,
+                            shell: user.Shell.v
+                        };
+                        resolve(the_user);
+                    })
+                    .catch(ex => reject(ex))
```
martinpitt, line 2410: ... then you can also drop this boilerplate.

pkg/lib/cockpit.js
```diff
@@ -2391,29 +2391,28 @@ function factory() {
...
+                    .finally(() => dbus.close());
+            } else {
+                resolve(the_user);
```
martinpitt, line 2413: ... and in this case just return `Promise.resolve(the_user)`.

martinpitt (dismissed):
Nice, thanks! I leave it up to you what you want to do with the further cleanup.

jelly:
> Nice, thanks! I leave it up to you what you want to do with the further cleanup.

Let's make it proper!

pkg/lib/cockpit.js
jelly: Right, but how would the caching work then? Or do we drop that, it doesn't seem too useful..

martinpitt: No, please keep it. I though just like this:
```js
if (the_user)
    return Promise.resolve(the_user);

return dbus.call(...).then([user] => {
    the_user = {...};
    return the_user;
});
```

jelly: Tried that, but it just explodes.

```Uncaught TypeError: import_cockpit23.default.user() is undefined
    LastLogin lastLogin.jsx:77
```

As `cockpit.user()` no longer returns a promise:

```
    let the_user = null;
    cockpit.user = function () {
            if (!the_user) {
                const dbus = cockpit.dbus(null, { bus: "internal" });
                return dbus.call("/user", "org.freedesktop.DBus.Properties", "GetAll",
                          ["cockpit.User"], { type: "s" })
                    .then(([user]) => {
                        the_user = {
                            id: user.Id.v,
                            name: user.Name.v,
                            full_name: user.Full.v,
                            groups: user.Groups.v,
                            home: user.Home.v,
                            shell: user.Shell.v
                        };
                        return the_user;
                    })
                    .finally(() => dbus.close());
            } else {
                Promise.resolve(the_user);
            }
    };
```

```
cockpit.user()
undefined 
```

So I'm gonna leave this out till the DBus API uses a proper promise API

martinpitt: Yes, you forgot the `return` in front of `Promise.resolve(the_user);` in the `else` branch :grin:

jelly: Facepalm, re-pushed! Hope it goes green.

pkg/shell/hosts.jsx
cockpituous, line 73: This added line is not executed by any test.

PR #18592, opened 2023-04-01
Software Update: Fix output logs printing in one line when failing to load updates in software update page

### LINKED ISSUE #18581: Software updates: Output when failing to load updates come in a single line
## Steps to reproduce
1. Disconnect the system from the internet (I used mobile hotspot with mobile data turned off)
2. Open cockpit web console on your linux machine and login to your local computer
3. Go to software updates to see that each line of the output joined or concatenated in a single line.

## Screenshots

Screenshot from 2023-03-29 20-38-48

image

## Implementation

- [X] I am interested in solving this issue

### Version of Cockpit

287-1~bpo22.10.1

Software updates

### Server operating system

Ubuntu

22.10

Chrome, Edge

### System log

_No response_

tripathics:
Resolves #18581 
Software updates: Output when failing to load updates come in a single line

The updated design follows the [`<ApplyUpdates>` component](https://github.com/cockpit-project/cockpit/blob/891fb8648cb732b4fdf9b93a6e35876eb81fdea0/pkg/packagekit/updates.jsx#L593)'s design to include error logs in a `<CodeBlockCode>` in an `<Expandable>`. 

## File changes
* pkg/packagekit/updates.jsx: 
    * Move logs outside of `<EmptyStatePanel>`'s paragraph to an `<Expandable>` component inside a `<CodeBlockCode>` to correctly print logs on separate lines (the bug), 
    * Move the *'Please reload the page after resolving the issue.'* message above the logs.
* pkg/packagekit/updates.scss:
    * Add class selectors following the styles in `progress-main-view`, `.update-log` and `.update-log-content` as well as include `text-align: left;` for the log output.
    
## Screenshots

Before|After
:---:|:---:
Single line log on updates error|Fixed

### DIFF
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -1477,27 +1477,18 @@ class OsUpdates extends React.Component {
                     <EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
                                      icon={ ExclamationCircleIcon }
                                      paragraph={
-                                         <TextContent>
-                                             <Text component={TextVariants.p}>
-                                                 {_("Please reload the page after resolving the issue.")}
-                                             </Text>
-                                         </TextContent>
+                                         <Text component={TextVariants.p}>
+                                             {_("Please reload the page after resolving the issue.")}
+                                         </Text>
                                      }
                     />
                     <Grid hasGutter>
                         <GridItem span="12" className="error-log">
-                            <ExpandableSection toggleText={_("View error log")} onToggle={() => {
-                                const log = document.getElementById("error-log");
-                                log.scrollTop = log.scrollHeight;
-                            }}>
-                                <div id="error-log" className="error-log-content">
-                                    <CodeBlockCode>
-                                        {this.state.errorMessages
-                                                .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
-                                                .map(m => <span key={m}>{m}</span>)}
-                                    </CodeBlockCode>
-                                </div>
-                            </ExpandableSection>
+                            <CodeBlockCode>
+                                {this.state.errorMessages
+                                        .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
+                                        .map(m => <span key={m}>{m}</span>)}
+                            </CodeBlockCode>
                         </GridItem>
                     </Grid>
                 </div>
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -27,7 +27,7 @@ import { createRoot } from 'react-dom/client';
 import { Alert } from "@patternfly/react-core/dist/esm/components/Alert/index.js";
 import { Badge } from "@patternfly/react-core/dist/esm/components/Badge/index.js";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
-import { CodeBlockCode, CodeBlock } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
+import { CodeBlock, CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
 import { Gallery } from "@patternfly/react-core/dist/esm/layouts/Gallery/index.js";
 import { Modal } from "@patternfly/react-core/dist/esm/components/Modal/index.js";
 import { Popover } from "@patternfly/react-core/dist/esm/components/Popover/index.js";
@@ -1483,7 +1483,7 @@ class OsUpdates extends React.Component {
                                         </TextContent>
                                     }
                     />
-                    <CodeBlock>
+                    <CodeBlock className='pf-u-mx-auto error-log'>
                         <CodeBlockCode>
                             {this.state.errorMessages
                                     .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -43,6 +43,7 @@ import { Spinner } from "@patternfly/react-core/dist/esm/components/Spinner/inde
 import { Stack, StackItem } from "@patternfly/react-core/dist/esm/layouts/Stack/index.js";
 import { Switch } from "@patternfly/react-core/dist/esm/components/Switch/index.js";
 import { Text, TextContent, TextList, TextListItem, TextVariants } from "@patternfly/react-core/dist/esm/components/Text/index.js";
+import { CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
 
 import {
     BugIcon,
@@ -76,7 +77,6 @@ import * as python from "python.js";
 import callTracerScript from './callTracer.py';
 
 import "./updates.scss";
-import { CodeBlockCode } from '@patternfly/react-core';
 
 const _ = cockpit.gettext;
 
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -27,6 +27,7 @@ import { createRoot } from 'react-dom/client';
 import { Alert } from "@patternfly/react-core/dist/esm/components/Alert/index.js";
 import { Badge } from "@patternfly/react-core/dist/esm/components/Badge/index.js";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
+import { CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
 import { Gallery } from "@patternfly/react-core/dist/esm/layouts/Gallery/index.js";
 import { Modal } from "@patternfly/react-core/dist/esm/components/Modal/index.js";
 import { Popover } from "@patternfly/react-core/dist/esm/components/Popover/index.js";
@@ -43,7 +44,6 @@ import { Spinner } from "@patternfly/react-core/dist/esm/components/Spinner/inde
 import { Stack, StackItem } from "@patternfly/react-core/dist/esm/layouts/Stack/index.js";
 import { Switch } from "@patternfly/react-core/dist/esm/components/Switch/index.js";
 import { Text, TextContent, TextList, TextListItem, TextVariants } from "@patternfly/react-core/dist/esm/components/Text/index.js";
-import { CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
 
 import {
     BugIcon,
@@ -1473,21 +1473,23 @@ class OsUpdates extends React.Component {
             });
 
             return (
-                <EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
-                                 icon={ ExclamationCircleIcon }
-                                 paragraph={
-                                     <TextContent>
-                                         <Text component={TextVariants.p}>
-                                             {_("Please resolve the issue and reload this page.")}
-                                         </Text>
-                                         <CodeBlockCode className='error-log'>
-                                             {this.state.errorMessages
-                                                     .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
-                                                     .map(m => <span key={m}>{m}</span>)}
-                                         </CodeBlockCode>
-                                     </TextContent>
-                                 }
-                />
+                <Stack hasGutter>
+                    <EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
+                                    icon={ ExclamationCircleIcon }
+                                    paragraph={
+                                        <TextContent>
+                                            <Text component={TextVariants.p}>
+                                                {_("Please resolve the issue and reload this page.")}
+                                            </Text>
+                                        </TextContent>
+                                    }
+                    />
+                    <CodeBlockCode className='error-log'>
+                        {this.state.errorMessages
+                                .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
+                                .map(m => <span key={m}>{m}</span>)}
+                    </CodeBlockCode>
+                </Stack>
             );
 
         case "applying":
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -1473,25 +1473,21 @@ class OsUpdates extends React.Component {
             });
 
             return (
-                <div className='update-error-main-view'>
-                    <EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
-                                     icon={ ExclamationCircleIcon }
-                                     paragraph={
+                <EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
+                                 icon={ ExclamationCircleIcon }
+                                 paragraph={
+                                     <TextContent>
                                          <Text component={TextVariants.p}>
-                                             {_("Please reload the page after resolving the issue.")}
+                                             {_("Please resolve the issue and reload this page.")}
                                          </Text>
-                                     }
-                    />
-                    <Grid hasGutter>
-                        <GridItem span="12" className="error-log">
-                            <CodeBlockCode>
-                                {this.state.errorMessages
-                                        .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
-                                        .map(m => <span key={m}>{m}</span>)}
-                            </CodeBlockCode>
-                        </GridItem>
-                    </Grid>
-                </div>
+                                         <CodeBlockCode className='error-log'>
+                                             {this.state.errorMessages
+                                                     .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
+                                                     .map(m => <span key={m}>{m}</span>)}
+                                         </CodeBlockCode>
+                                     </TextContent>
+                                 }
+                />
             );
 
         case "applying":
--- a/pkg/packagekit/updates.jsx
+++ b/pkg/packagekit/updates.jsx
@@ -76,6 +76,7 @@ import * as python from "python.js";
 import callTracerScript from './callTracer.py';
 
 import "./updates.scss";
+import { CodeBlockCode } from '@patternfly/react-core';
 
 const _ = cockpit.gettext;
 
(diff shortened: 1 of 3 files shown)


KKoukiou (commented):
thanks for the PR. I am having some concerns about hiding errors behind the expandable section. The issue is that if the users end atthis error screen they need to take action to solve it. This design was used in the update logs, where the user does not need to interact with that part of the UI, it's more informative. For this reason, hiding the error logs is imo wrong design wise. 


I would like that we keep the codeblock but drop the rest.

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,34 @@ class OsUpdates extends React.Component {
             });
 
             return (
-                <>
+                <div className='update-error-main-view'>
                     <EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
                                      icon={ ExclamationCircleIcon }
                                      paragraph={
                                          <TextContent>
```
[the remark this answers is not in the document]
tripathics: Removed TextContent

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,34 @@ class OsUpdates extends React.Component {
...
+                    <Grid hasGutter>
+                        <GridItem span="12" className="error-log">
+                            <ExpandableSection toggleText={_("View error log")} onToggle={() => {
+                                const log = document.getElementById("error-log");
+                                log.scrollTop = log.scrollHeight;
+                            }}>
+                                <div id="error-log" className="error-log-content">
+                                    <CodeBlockCode>
+                                        {this.state.errorMessages
+                                                .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
```
tripathics, line 1496: Remove `<ExpandableSection>`

pkg/packagekit/updates.scss
```diff
@@ -153,6 +153,20 @@ div.changelog {
...
+}
+
+.error-log {
+  text-align: left;
+
+  .error-log-content {
+    margin: 0 10ex;
+  }
+}
+
```
garrett: It's generally best to not add custom CSS unless it's needed (as it _can_ be brittle and would need to be supported, plus PF usually gives us ways to do things without adding custom CSS). In this case, I we could use PF widgets in the code itself and (probably) skip the CSS.

Some stuff, like alignment, also has PF utility classes now as well. https://www.patternfly.org/v4/utilities/alignment

Utility classes require 1 import on the top-level CSS (near the top of the file) for the page, if they're not already added in that file. (Do _not_ add them to imported files, else that causes a very high risk of CSS duplication.)

I think it'd be imported like this:

```scss
@import "@patternfly/patternfly/utilities/Alignment/alignment.css";
```

pkg/packagekit/updates.scss
```diff
@@ -153,6 +153,20 @@ div.changelog {
...
 .update-log {
   text-align: center;
```
garrett, line 169: Yeah, we have alignment in CSS here too.

Generally, we should use the PF utilities, unless it's part of a larger custom block that does multiple things.

To change this would require a separate commit and is probably best left for a separate PR. (So it's fine ignore it in this PR.)

tripathics:
@garrett I have updated the style of the output in CodeBlockCode to use the `--pf-global--FontSize--sm`. Is it ok?

Update error after removing Expandable and updated font size

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,25 @@ class OsUpdates extends React.Component {
...
-                                                         .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
-                                                         .map(m => <span key={m}>{m}</span>)}
-                                             </Text>
-                                             <Text component={TextVariants.p}>
-                                                 {_("Please reload the page after resolving the issue.")}
-                                             </Text>
-                                         </TextContent>
+                                         <Text component={TextVariants.p}>
+                                             {_("Please reload the page after resolving the issue.")}
+                                         </Text>
```
garrett, line 1482: This part should probably still come after the error messages.

The flow of information is:

1. Error title
2. Error details
3. How to resolve

Switching it around with error title, resolution, and then error details is a bit odd.

garrett: Additionally, we could change it to "Please resolve the issue and reload this page." This would put it in a more logical order.

(This isn't related to this PR, but if you're touching that part of the code anyway, we could improve it too.)

garrett (changes requested):
I agree with @KKoukiou. This shouldn't be in an expandable, as it's required to see what the problem is to fix it manually.

Generally in Cockpit, we try to have an automated way of fixing things from within Cockpit itself, but there are cases like this where we have to show the error and require manual intervention. (We can't automate _everything_.) In these cases, we need to show the error in a clear, concise manner and provide output. In most of these cases, it's not something that stops you from continuing completely, but this is one of those cases where the page cannot do anything else, so it's rather important to not hide it away.

I've added a bunch of details inline.

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,25 @@ class OsUpdates extends React.Component {
...
                                      }
                     />
-                </>
+                    <Grid hasGutter>
```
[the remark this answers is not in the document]
tripathics: Earlier I tried it inside the paragraph itself but it was applying `text-align: center;`. Guess I will use css there.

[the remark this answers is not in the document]
tripathics: Oh. I used the Textcontent to wrap both the CodeBlockCode and the Text. Is it fine?

pkg/packagekit/updates.jsx
tripathics: Yes. I have changed the How to resolve.

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,25 @@ class OsUpdates extends React.Component {
             });
 
             return (
-                <>
+                <div className='update-error-main-view'>
```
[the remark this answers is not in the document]
tripathics: Removed this div. Only have a single EmptyStatePanel component now.

```jsx
return (
<EmptyStatePanel title={ STATE_HEADINGS[this.state.state] }
                                 icon={ ExclamationCircleIcon }
                                 paragraph={
                                     <Stack hasGutter>
                                         <TextContent>
                                             <Text component={TextVariants.p}>
                                                 {_("Please resolve the issue and reload this page.")}
                                             </Text>
                                         </TextContent>
                                         <CodeBlockCode className='error-log'>
                                             {this.state.errorMessages
                                                     .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
                                                     .map(m => <span key={m}>{m}</span>)}
                                         </CodeBlockCode>
                                     </Stack>
                                 }
                />
)
```

KKoukiou (commented):
So I just tried and on mobile there is now an overflow, which makes this PR actually causing a regression for this page. The problem comes from the fact the Codeblock is using pre which does not allow lines to wrap and such.

Shall we better switch back to using text for the implementation? Why did you move to using codeblock anyway? 

For seperating the lines we can just do:

```
+                                         {this.state.errorMessages
+                                                 .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
+                                                 .map(m => <Text key={m}>{m}</Text>)}
```

So using Text elements instead of span for the messages.

pkg/packagekit/updates.scss
```diff
@@ -153,6 +153,12 @@ div.changelog {
   }
 }
 
+.error-log {
+  font-size: var(--pf-global--FontSize--sm);
```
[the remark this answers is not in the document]
tripathics: > So I just tried and on mobile there is now an overflow, which makes this PR actually causing a regression for this page. The problem comes from the fact the Codeblock is using pre which does not allow lines to wrap and such.
> 
> Shall we better switch back to using text for the implementation? Why did you move to using codeblock anyway?
> 
> For seperating the lines we can just do:
> 
> ```
> +                                         {this.state.errorMessages
> +                                                 .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
> +                                                 .map(m => <Text key={m}>{m}</Text>)}
> ```
> 
> So using Text elements instead of span for the messages.

Using Text wont help. Will have to use TextVariant pre or code but will be getting the same overflow issue. The overflow issue can be solved with `whitespace: pre-wrap`. Is there a PF component for that?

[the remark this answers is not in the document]
tripathics: The overflow issue would be solved if we keep error block outside EmptyStatePanel's paragraph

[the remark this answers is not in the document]
KKoukiou: I see. so this output is not in seperate entries or of errorMessages array, but in one. There is not, I guess you can put a custom css for that only then.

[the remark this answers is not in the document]
tripathics: @KKoukiou Should I move it outside the paragraph? Overflow is still there after applying custom CSS.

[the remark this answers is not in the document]
tripathics: @KKoukiou the default font-size for pre seemed too large for when the error log was large. Also I thought it looked better on small screens when I used the above font size. 

```suggestion
  font-size: var(--pf-global--FontSize--sm);
```

Should I keep this font size?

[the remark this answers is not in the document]
garrett: > pre which does not allow lines to wrap and such

Yeah, you need pre-wrap for that, which (AFAIK), isn't available in PF. So I guess it does need a sprinkling of custom CSS.

Also, does `paragraph={}` actually wrap it in `<p>` tags on the page? If so, that's wrong. If not, then it's most ideal to use it there (if it's just where detail text goes). 

In other words `<p>` tags cannot have block-level tags inside it.

Details:

- P can contain phrasing, but is flow content
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element
- PRE can contain phrasing, but is flow content
   - https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element
- phrasing cannot have flow content descendants
   - https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2

(Note: I'm not sure if PF's `paragraph` in that widget really renders to a paragraph tag `<p>` or not. Figuring this out — if it is rendered to `<p>` in the DOM — might help make a decision.)

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,23 @@ class OsUpdates extends React.Component {
...
+                                    paragraph={
+                                        <TextContent>
+                                            <Text component={TextVariants.p}>
+                                                {_("Please resolve the issue and reload this page.")}
+                                            </Text>
+                                        </TextContent>
+                                    }
                     />
-                </>
+                    <CodeBlockCode className='error-log'>
```
garrett, line 1487: You're using `CodeBlockCode`, but it isn't actually making this a code block. The font isn't correct.

I'm not sure why it's not pulling in the font correctly. A guess: Is the import not right? That would also explain why it's not styled at all.

Screenshot from 2023-04-04 12-54-57

pkg/packagekit/updates.scss
```diff
@@ -153,6 +153,13 @@ div.changelog {
   }
 }
 
+.error-log {
+  margin: 0 auto;
+  padding: 0 3ex;
+  max-width: fit-content;
+  width: 100%;
+}
```
garrett, line 161: Is there not a PF way to handle the alignment? (Perhaps you're adding this because the import doesn't seem to be done correctly somehow, so the PF CSS isn't being loaded.)

Spacing should use PF variables, so it's consistent (and would automatically update if PF ever did). I think `md` might be the right one here.

I'm not sure if max-width of fitting content and having the width at 100% makes sense here. The block is already being center aligned, so it will distribute the space on the sides (hence not needing fit-content). It might be better to let it auto-size by default and set the max-width to 100% so smaller sizes would wrap.

So, here's my suggestion to the changes:

```suggestion
.error-log {
  margin: 0 auto;
  padding: 0 var(--pf-globel--spacer--md);
  max-width: 100%;
}
```

(This might differ if the PF compoenent changes things when it's fully imported.)

pkg/packagekit/updates.jsx
```diff
@@ -1472,23 +1473,23 @@ class OsUpdates extends React.Component {
             });
 
             return (
-                <>
+                <Stack hasGutter>
```
garrett, line 1476: There's too much vertical space between items. I think the `hasGutter` is adding too much. It's either that, and/or one of the other elements adding vertical space between items as well.

pkg/packagekit/updates.jsx
tripathics: Yes this was concerning me as well.

pkg/packagekit/updates.scss
tripathics: @garrett any idea how to import in CSS

garrett (changes requested):
This is looking closer to what I'd expect! The result is definitely better than what's in main right now.

I've added a bunch of things inline, but I think the main issue stems from the PF component not being imported properly?

At a glance, it looks like it's imported in a correct manner from the code though? But it's definitely not pulling in any CSS that it needs. I think @KKoukiou will need to help you out with this.

My quick guess is that it should probably be:

```scss
import { CodeBlock, CodeBlockCode } from '@patternfly/react-core';
```

And you'd need to use `<CodeBlock>` below as well, I think.

pkg/packagekit/updates.scss
garrett: I meant for these commits to land as a part of my full review, not ahead of time. Whoops!

But I have a suggestion in the review. (Look below.) It changes the input and adds `<CodeBlock>` as a wrapper.

tripathics:
> This is looking closer to what I'd expect! The result is definitely better than what's in main right now.
> 
> I've added a bunch of things inline, but I think the main issue stems from the PF component not being imported properly?
> 
> At a glance, it looks like it's imported in a correct manner from the code though? But it's definitely not pulling in any CSS that it needs. I think @KKoukiou will need to help you out with this.
> 
> My quick guess is that it should probably be:
> 
> ```css-scss
> import { CodeBlock, CodeBlockCode } from '@patternfly/react-core';
> ```
> 
> And you'd need to use `<CodeBlock>` below as well, I think.

Where do I have to use CodeBlock?

pkg/packagekit/updates.jsx
tripathics: Removed `hasGutter`. Looks better now

pkg/packagekit/updates.jsx
tripathics: Even after using CodeBlock, the --pf-c-code-block__code--FontFamily is undefined

KKoukiou: The problem is that you 're using codeblockcode without wrapping it with codeblock. The styles are all from codeblock, not codeblockcode.

garrett: Oh? I guess you're testing it locally. PF often sets variables like that in a topmost element and then inherits the variable from children elements.

Looking at the React example on PatternFly's website, I see this:

image

And that shows me that `--pf-c-code-block__pre--FontFamily` is set in `pf-c-code-block`.

So I'm not sure why it's not working when you're using something like:

```jsx
<CodeBlock>
  <CodeBlockCode>
    ERRORS GO HERE
  </CodeBlockCode>
</CodeBlock>
```
(...With the import I suggested elsewhere in this PR that pulls in both CodeBlock and CodeBlockCode.)

@KKoukiou would probably know more.

tripathics: It's finally fixed. I didn't run test/image-prepare before testing.

tripathics:
I think it's finally fixed. @KKoukiou kindly review
I have removed all the css I previously added

image

pkg/packagekit/updates.scss
tripathics: Solved after wrapping in CodeBlockCode in CodeBlock

pkg/packagekit/updates.jsx
```diff
@@ -27,7 +27,7 @@ import { createRoot } from 'react-dom/client';
 import { Alert } from "@patternfly/react-core/dist/esm/components/Alert/index.js";
 import { Badge } from "@patternfly/react-core/dist/esm/components/Badge/index.js";
 import { Button } from "@patternfly/react-core/dist/esm/components/Button/index.js";
-import { CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
+import { CodeBlockCode, CodeBlock } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
```
garrett, line 30: It should probably be this instead:

```scss
import { CodeBlock, CodeBlockCode } from "@patternfly/react-core/dist/esm/components/CodeBlock/index.js";
```

tripathics: Will do this

pkg/packagekit/updates.jsx
```diff
@@ -1484,11 +1483,13 @@ class OsUpdates extends React.Component {
                                         </TextContent>
                                     }
                     />
-                    <CodeBlockCode className='error-log'>
-                        {this.state.errorMessages
-                                .filter((m, index) => index == 0 || m != this.state.errorMessages[index - 1])
-                                .map(m => <span key={m}>{m}</span>)}
-                    </CodeBlockCode>
+                    <CodeBlock>
```
garrett, line 1486: You should be able to add `className="pf-u-ml-auto pf-u-mr-auto"` to `CodeBlock` to make it align in the center. (We're already important the spacing helper classes on `updates.scss`, so adding the classes should be enough.)

garrett (dismissed):
So I do like this; it looks a lot better. I also like your suggestion about centering the block to the content's width.

However, all it needs is a margin on the left and right to be set to auto. All the rest just works. (No width or max-width needed.) And there's padding inside the widget, so that's fine too.

You could add classes to the element with `className="pf-u-ml-auto pf-u-mr-auto"`, which should add the CSS to add margin-left and margin-right set to auto... but that requires adding the PatternFly spacing utility CSS to the toplevel stylesheet (`updates.scss`) with an import line like this:

```
@import "@patternfly/patternfly/utilities/Spacing/spacing.css";
```

But that's already imported in the file! Which means we can just use the classes directly in the JSX.

The only other thing is the order of the imports in `updates.jsx`; not only do we have them in alphabetical order, we have the main widgets imported before the sub-widgets. I'll add a suggestion in an inline comment.

These are just minor changes. I think it's ready for a merge after making these slight adjustments and getting @KKoukiou's OK.

pkg/packagekit/updates.jsx
garrett: BTW: pf-u-ml-auto stands for "patternfly utility, margin-left: auto`

The utility classes are oddly named if you don't know what they mean, so I thought I should point it out. :grin: 

For reference, here's the spacing page: https://www.patternfly.org/v4/utilities/spacing

garrett: OH, this can actually be shortened to: `pf-u-mx-auto` for adding margin to the sides! TIL ("Today, I learned...")

martinpitt:
This looks good from a first quick look, triggering all tests. Thanks! Will do a proper review tomorrow morning.

tripathics:
> This looks good from a first quick look, triggering all tests. Thanks! Will do a proper review tomorrow morning.

@martinpitt 
Actually I was trying to rebase my repository with the updated cockpit/main but protection checks were failing. So I force pushed previous commit again :(

martinpitt:
@tripathics : There are three other places in the tests which need to be updated, see [failure log](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18592-20230404-190336-38f65a61-fedora-37/log.html#346). Should be straightforward, but please let me know if you don't understand any of them. Thanks!

martinpitt:
@garrett  This is nearing completion, and AFAICS [your earlier concerns](https://github.com/cockpit-project/cockpit/pull/18592#pullrequestreview-1371210957) should be moot now -- this was all addressed by using the proper element structure, and there are no "magic" classes any more. PTAL?

martinpitt:
Actually, unping. Garrett already reviewed this, is overloaded enough, and this looks good.

martinpitt (approved):
Thank you! I'll squash the commits and clean up the commit message on landing.

tripathics:
Thank you @martinpitt for your approval. It is showing here that merging can be performed automatically once requested changes are addressed. What does this mean?

martinpitt:
@tripathics : Fixed -- just waiting for the firefox test run now.

martinpitt:
@tripathics Congrats, and thank you for your contribution! :tada:

PR #18596, opened 2023-04-03
modules: update systemd_ctypes

allisonkarlitskaya:
The latest version of systemd_ctypes brings in a new approach to D-Bus serialization which is substantially more efficient and somewhat cleaner, but less Cockpit-compatible.

Fortunately, it also includes a new JSONEncoder subclass that knows how to represent the new types ('bytes' instead of base64 encoded text, and 'Variant' instances instead of magic dicts) in the expected JSON format.

On the sending side, the serializer code implicitly supports these magic forms if it sees them, so we don't need any changes there.

<hr/>

Doing this as a separate PR just in case the substantial changes in systemd_ctype have unanticipated side-effects.

### DIFF
--- a/src/cockpit/channels/dbus.py
+++ b/src/cockpit/channels/dbus.py
@@ -159,7 +159,7 @@ class InterfaceCache:
 
 
 def notify_update(notify, path, interface_name, props):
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Tests look promising so far. Triggering required fcos and undrafting. Thanks!

